PR #15568, opened 2024-10-01
Migrate fog volumes to required components

Jondolf:
# Objective

Another part of the migration to required components: fog volumes!

## Solution

Deprecate `FogVolumeBundle` and make `FogVolume` require `Transform` and `Visibility`, as per the [chosen proposal](https://hackmd.io/@bevy/required_components/%2FcO7JPSAQR5G0J_j5wNwtOQ).

---

## Migration Guide

Replace all insertions of `FogVolumeBundle` with the `Visibility` component. The other components required by it will now be inserted automatically.


kristoff3r (approved):
This PR looks fine, but I noticed some unrelated stuff while testing:
* Not having used these components before I didn't know we had both `FogVolume` and `VolumetricFog`. It makes total sense but it would be nice if the docs explained the difference
* The `volumetric_fog` example is currently broken on main due to the retained render world PR, so I couldn't actually test properly

NiseVoid (approved):

cart (approved):

Issue #15558, opened 2024-10-01
Reduce number of copies in the Mesh API


atlv24:
## What problem does this solve or what need does it fill?

Currently, using the Mesh API incurs a lot of unnecessary memcpy. If you have a custom mesh representation, it's even worse because you have to build a Mesh copying all your data in, and then go through all the normal Mesh overhead.

Current state of Mesh copies:

1. Mesh gets extracted which may involve a clone in `extract_render_asset`
[`extracted_assets.push((id, asset.clone()));`](https://github.com/bevyengine/bevy/blob/de888a373de666ff9d1bbdf8275e7fb625584394/crates/bevy_render/src/render_asset.rs#L296)

2. then it gets prepared into a GpuMesh, which interleave-copies the vertex data into a `Vec<u8>` in `Mesh::get_vertex_buffer_data`:
[`let mut attributes_interleaved_buffer = vec![0; vertex_count * vertex_size];`](https://github.com/bevyengine/bevy/blob/de888a373de666ff9d1bbdf8275e7fb625584394/crates/bevy_render/src/mesh/mesh/mod.rs#L469)

3. the interleave copy actually happens twice, the first time just to get the length and pre-allocate a slab 😬 
[`mesh.get_vertex_buffer_data().len() as u64,`](https://github.com/bevyengine/bevy/blob/de888a373de666ff9d1bbdf8275e7fb625584394/crates/bevy_render/src/mesh/allocator.rs#L430)

and then to fill in the buffer
[`let vertex_data = mesh.get_vertex_buffer_data();`](https://github.com/bevyengine/bevy/blob/de888a373de666ff9d1bbdf8275e7fb625584394/crates/bevy_render/src/mesh/allocator.rs#L477)

4. which then gets passed by reference to wgpu `create_buffer_with_data` which immediately copies it yet again in `DeviceExt::create_buffer_init`:
[`buffer.slice(..).get_mapped_range_mut()[..unpadded_size as usize].copy_from_slice(descriptor.contents);`](https://github.com/gfx-rs/wgpu/blob/6db097694ce121f9cd927ef6dbb5608a69906760/wgpu/src/util/device.rs#L95)
which then actually gets uploaded to the gpu

## What solution would you like?

To solve 1:
Retained render world will help here

To solve 2:
introduce an api which both Mesh and custom mesh representations uses, that permits writing directly into a u8 buffer thats already mapped. this can be via a builder style api potentially.

To solve 3:
[...]

PR #15569, opened 2024-10-01
Zero Copy Mesh

atlv24:
# Objective

- Another step towards #15558

## Solution

- Instead of allocating a Vec and then having wgpu copy it into a staging buffer, write directly into the staging buffer.
- gets rid of another hidden copy, in `pad_to_alignment`.

future work:
- why is there a gcd implementation in here (and its subpar, use binary_gcd. its in the hot path, run twice for every mesh, every frame i think?) make it better and put it in bevy_math
- zero-copy custom mesh api to avoid having to write out a Mesh from a custom rep

## Testing

- lighting and many_cubes run fine (and slightly faster. havent benchmarked though)

---

## Showcase

- look ma... no copies

at least when RenderAssetUsage is GPU only :3


crates/bevy_render/src/mesh/allocator.rs
```diff
@@ -557,17 +563,19 @@ impl MeshAllocator {
...
-                        label: Some(&format!(
-                            "large mesh slab {} ({}buffer)",
-                            slab_id,
-                            buffer_usages_to_str(buffer_usages)
-                        )),
-                        contents: data,
-                        usage: buffer_usages | BufferUsages::COPY_DST,
-                    },
-                ));
+                let buffer = render_device.create_buffer(&BufferDescriptor {
```
IceSentry, line 566: What happens with `large_object_slab.buffer`? Is it supposed to never be set?

atlv24: good catch, fixed

IceSentry (approved):
I haven't tested it yet, but the code LGTM

crates/bevy_render/src/mesh/mesh/mod.rs
```diff
@@ -385,6 +385,13 @@ impl Mesh {
             .sum()
     }
 
+    /// Returns the size of a vertex in bytes.
```
kristoff3r, line 388:
```suggestion
    /// Returns the size of all the vertices in bytes.
```

crates/bevy_render/src/mesh/allocator.rs
```diff
@@ -516,10 +516,11 @@ impl MeshAllocator {
 
     /// A generic function that copies either vertex or index data into a slab.
     #[allow(clippy::too_many_arguments)]
-    fn copy_element_data(
+    fn copy_element_data<T: Fn(&mut [u8])>(
```
kristoff3r, line 519: I prefer using `F` for function type parameters

crates/bevy_render/src/mesh/allocator.rs
```diff
@@ -540,12 +541,17 @@ impl MeshAllocator {
...
+                if let Some(size) = BufferSize::new((len as u64).next_multiple_of(slot_size)) {
+                    // Write the data in.
+                    if let Some(mut buffer) = render_queue.write_buffer_with(
+                        buffer,
+                        allocated_range.allocation.offset as u64 * slot_size,
+                        size,
+                    ) {
+                        let slice = &mut buffer.as_mut()[..len];
+                        fill_data(slice);
+                    }
```
kristoff3r, line 553: According to the documentation of `write_buffer_with` it only returns None if the buffer doesn't have enough bytes at the given offset. That seems like an invariant of the code, so we might want to panic or debug_assert instead of silently not copying the mesh?

kristoff3r (approved):
I tested a few examples and they still work.

crates/bevy_render/src/mesh/allocator.rs
atlv24: as is, it preserves the old behavior. the method it used to use for writing, `write_buffer`, says "This method fails if `data` overruns the size of `buffer` starting at `offset`."

I digged into the wgpu source, and neither of the two are silent about it, they both print errors to console:
```rs
match self.0.queue_write_buffer(queue_data.id, buffer_data.id, offset, data) {
    Ok(()) => (),
    Err(err) => {
        self.handle_error_nolabel(&queue_data.error_sink, err, "Queue::write_buffer")
    }
}
```

 I think its fine to leave it as is.

JMS55 (changes requested):
Code looks good minus the one part I don't understand yet. (didn't test perf)

crates/bevy_render/src/mesh/allocator.rs
```diff
@@ -557,17 +564,22 @@ impl MeshAllocator {
...
+                    )),
+                    size: len as u64,
+                    usage: buffer_usages | BufferUsages::COPY_DST,
+                    mapped_at_creation: true,
+                });
+                {
+                    let slice = &mut buffer.slice(..).get_mapped_range_mut()[..len];
+                    fill_data(slice);
+                    buffer.unmap();
+                }
```
kristoff3r, line 581: Tested locally, this fixes the wasm error for me

```suggestion
                }
                buffer.unmap();
```

PR #15500, opened 2024-09-28
Better warnings about invalid parameters

MiniaczQ:
# Objective

System param validation warnings should be configurable and default to "warn once" (per system).

Fixes: #15391

## Solution

`SystemMeta` is given a new `ParamWarnPolicy` field.
The policy decides whether warnings will be emitted by each system param when it fails validation.
The policy is updated by the system after param validation fails.

Example warning:
```
2024-09-30T18:10:04.740749Z  WARN bevy_ecs::system::function_system: System fallible_params::do_nothing_fail_validation will not run because it requested inaccessible system parameter Single<(), (With<Player>, With<Enemy>)>
```

Currently, only the first invalid parameter is displayed.

Warnings can be disabled on function systems using `.param_never_warn()`.
(there is also `.with_param_warn_policy(policy)`)

## Testing

Ran `fallible_params` example.

PR #15571, opened 2024-10-01
Optimize param validation through new `try_get_param`

### LINKED ISSUE #15505: Optimize `validate_param`

From discord discussion: https://discord.com/channels/691052431525675048/749335865876021248/1289681891611508809

As @cart pointed out, `validate_param` doubles lookup, which can have significant performance impact.

There are few potential ways of fixing that:
1. Merge `get_param` with `validate_param` into `try_get_param`.
1.1. In naive approach, we always create a system task (multithreaded executor), even if parameters can't be retrieved.
1.2. We can try to run `try_get_param` before spawning a task, but then we need to somehow pass the parameters into the system. (type erasure issues)
2. Keep `get_param` and `validate_param`, add `try_get_param` with default implementation that merges the previous 2.
Implement manually for performance gains.

## Additional context

Blocked by #15394, we want most machinery in place before we make this change.

MiniaczQ:
# Objective

Fixes #15505

Waiting for #15500, should be merged before this is even finished

## Solution

TODO

## Testing

TODO


MiniaczQ:
On hold.
It may be possible to use a better solution
https://discord.com/channels/691052431525675048/749335865876021248/1291089012039094303

MiniaczQ:
Closed in favor of #15606

PR #15572, opened 2024-10-01
Migrate motion blur, TAA, SSAO, and SSR to required components

Jondolf:
# Objective

Again, a step forward in the migration to required components: a bunch of camera rendering cormponents!

Note that this does not include the camera components themselves yet, because the naming and API for `Camera` hasn't been fully decided yet.

## Solution

As per the [selected proposals](https://hackmd.io/@bevy/required_components/%2FpiqD9GOdSFKZZGzzh3C7Uw):

- Deprecate `MotionBlurBundle` in favor of the `MotionBlur` component
- Deprecate `TemporalAntiAliasBundle` in favor of the `TemporalAntiAliasing` component
- Deprecate `ScreenSpaceAmbientOcclusionBundle` in favor of the `ScreenSpaceAmbientOcclusion` component
- Deprecate `ScreenSpaceReflectionsBundle` in favor of the `ScreenSpaceReflections` component

---

## Migration Guide

`MotionBlurBundle`, `TemporalAntiAliasBundle`, `ScreenSpaceAmbientOcclusionBundle`, and `ScreenSpaceReflectionsBundle` have been deprecated in favor of the `MotionBlur`, `TemporalAntiAliasing`, `ScreenSpaceAmbientOcclusion`, and `ScreenSpaceReflections` components instead. Inserting them will now also insert the other components required by them automatically.

### DIFF
--- a/examples/3d/anti_aliasing.rs
+++ b/examples/3d/anti_aliasing.rs
@@ -13,12 +13,12 @@ use bevy::{
     pbr::CascadeShadowConfigBuilder,
     prelude::*,
     render::{
+        camera::TemporalJitter,
         render_asset::RenderAssetUsages,
         render_resource::{Extent3d, TextureDimension, TextureFormat},
         texture::{ImageSampler, ImageSamplerDescriptor},
     },
 };
-use bevy_render::camera::TemporalJitter;
 
 fn main() {
     App::new()
--- a/examples/3d/anti_aliasing.rs
+++ b/examples/3d/anti_aliasing.rs
@@ -5,10 +5,9 @@ use std::{f32::consts::PI, fmt::Write};
 use bevy::{
     core_pipeline::{
         contrast_adaptive_sharpening::ContrastAdaptiveSharpening,
-        experimental::taa::{
-            TemporalAntiAliasBundle, TemporalAntiAliasPlugin, TemporalAntiAliasing,
-        },
+        experimental::taa::{TemporalAntiAliasPlugin, TemporalAntiAliasing},
         fxaa::{Fxaa, Sensitivity},
+        prepass::{DepthPrepass, MotionVectorPrepass},
         smaa::{Smaa, SmaaPreset},
     },
     pbr::CascadeShadowConfigBuilder,
@@ -19,6 +18,7 @@ use bevy::{
         texture::{ImageSampler, ImageSamplerDescriptor},
     },
 };
+use bevy_render::camera::TemporalJitter;
 
 fn main() {
     App::new()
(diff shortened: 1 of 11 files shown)


examples/3d/anti_aliasing.rs
```diff
@@ -28,6 +28,13 @@ fn main() {
         .run();
 }
 
+type TaaComponents = (
+    TemporalAntiAliasing,
+    TemporalJitter,
+    DepthPrepass,
+    MotionVectorPrepass,
+);
```
Jondolf, line 36: This is kinda scuffed, but component removal is a bit more annoying without bundles 🤔

(of course I could also make this an actual `Bundle` like `TemporalAntiAliasBundle` too, but the properties aren't needed in this case)

examples/3d/anti_aliasing.rs
```diff
@@ -141,7 +148,7 @@ fn modify_aa(
         camera
             .remove::<Fxaa>()
             .remove::<Smaa>()
-            .insert(TemporalAntiAliasBundle::default());
+            .insert(TaaComponents::default());
```
NiseVoid, line 151: The insert doesn't need to use the bundle since the components are required anyway. Also would be nice if these `removes` got batched but that's not really a new issue in this PR

examples/3d/ssao.rs
```diff
@@ -39,8 +36,8 @@ fn setup(
             msaa: Msaa::Off,
             ..default()
         })
-        .insert(ScreenSpaceAmbientOcclusionBundle::default())
-        .insert(TemporalAntiAliasBundle::default());
+        .insert(ScreenSpaceAmbientOcclusion::default())
```
NiseVoid, line 39: Merging these with the `spawn` also would've been nice. tho that might conflict with a PR for `Camera3dBundle` if anyone is working on one

NiseVoid (approved):

examples/3d/ssao.rs
Jondolf: Yeah, merged these now. I do have a camera migration basically ready, but I'm not sure if the `Camera` vs. `CameraSettings` vs. something else was fully decided yet.

examples/3d/anti_aliasing.rs
cart: It does seem like we're missing a `camera.remove_with_requires::<TemporalAntiAliasing>()` function in bevy_ecs

cart (approved):
We definitely needs some form of `remove_with_requires`, but that shouldn't hold up progress here.

examples/3d/anti_aliasing.rs
alice-i-cecile: Tracked in #15580.

alice-i-cecile (approved):

PR #15574, opened 2024-10-01
Derive `Clone` on `Parent` and `Children`

atornity:
# Objective

`Parent` and `Children` should be `Clone`ble.

## Solution

Derive `Clone` on `Parent` and `Children`.


MrGVSV:
Wouldn't this allow users to accidentally put these components into invalid states?

alice-i-cecile:
This enables UB via mem swap shenanigans breaking the invariants relied on during unsafe transform propagation. So we can't merge this.

PR #15581, opened 2024-10-01
Refactor TextPipeline::update_buffer to accept an interator

UkoeHB:
# Objective

- Prepare `TextPipeline` to work with multi-entity text blocks. See https://github.com/bevyengine/bevy/discussions/15014

## Solution

- Refactor `TextPipeline::update_buffer` to accept an iterator instead of slice. Adjust `update_buffer` implementation to only iterate spans once instead of three times (which would require iterating a hierarchy three times with multi-entity blocks).

## Testing

- Tested with `text_debug` example.

### DIFF
--- a/crates/bevy_text/src/pipeline.rs
+++ b/crates/bevy_text/src/pipeline.rs
@@ -147,7 +147,7 @@ impl TextPipeline {
         let spans_iter = spans.iter().map(|(span_index, span, style, font_info)| {
             (
                 *span,
-                get_attrs(*span_index, *style, font_info, scale_factor),
+                get_attrs(*span_index, style, font_info, scale_factor),
             )
         });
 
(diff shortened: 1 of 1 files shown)


crates/bevy_text/src/pipeline.rs
```diff
@@ -384,13 +401,13 @@ impl TextMeasureInfo {
...
     font_system: &mut cosmic_text::FontSystem,
-    map_handle_to_font_id: &mut HashMap<AssetId<Font>, (cosmic_text::fontdb::ID, String)>,
+    map_handle_to_font_id: &mut HashMap<AssetId<Font>, (cosmic_text::fontdb::ID, Arc<str>)>,
     fonts: &Assets<Font>,
-) {
-    let font_handle = section.style.font.clone();
-    map_handle_to_font_id
+) -> FontFaceInfo {
+    let font_handle = style.font.clone();
+    let (face_id, family_name) = map_handle_to_font_id
```
kristoff3r, line 410: You can return the face itself here as well, so you don't need to look it up again.

kristoff3r (approved):
Not super familiar with this code but the changes make sense to me. I tested the text example and it still works.

crates/bevy_text/src/pipeline.rs
kristoff3r: Nevermind, you don't actually want to store it.

alice-i-cecile (approved):

alice-i-cecile:
@UkoeHB some jerk created a merge conflict for your PR!

Issue #15570, opened 2024-10-01
Several examples appear broken following recent changes


alice-i-cecile:
The example runs shows several example regressions, per @rparrett [on Discord](https://discord.com/channels/691052431525675048/692572690833473578/1290741433870848103).

- [x] ui_texture_*,
- [x] custom_shader_instancing
- [x] asset_decompression
- [x] volumetric_fog
- [x] transparency_2d (changed, not sure if broken)

Retained rendering world #15320 or various required components changes are both plausible.

This is in addition to #15563.

ChristopherBiscardi:
I think deferred_rendering is broken too

```
2024-10-01T19:43:15.414646Z  INFO bevy_diagnostic::system_information_diagnostics_plugin::internal: SystemInfo { os: "Windows 11 Pro", kernel: "22631", cpu: "AMD Ryzen 9 7950X 16-Core Processor", core_count: "16", memory: "63.2 GiB" }
HookApp::findGameWindow, injectWindow: 0000000000000000
HookApp::findGameWindow, graphicsWindow: 0000000000000000
OverlayConnector::onMessage,overlay.init
2024-10-01T19:43:16.256784Z  INFO bevy_render::renderer: AdapterInfo { name: "NVIDIA GeForce RTX 2080 Ti", vendor: 4318, device: 7687, device_type: DiscreteGpu, driver: "NVIDIA", driver_info: "560.94", backend: Vulkan }
2024-10-01T19:43:16.750808Z  INFO bevy_winit::system: Creating new window "App" (0v1#4294967296)
HookApp::deferHook,
HookApp::findGameWindow, injectWindow: 0000000000000000
HookApp::findGameWindow, graphicsWindow: 0000000000480DA8
2024-10-01T19:43:16.862308Z  INFO bevy_input::gamepad: Gamepad 44v1#4294967340 connected.
2024-10-01T19:43:17.191075Z  WARN bevy_pbr::render::mesh_view_bindings: Too many textures in mesh pipeline view layout, this might cause us to hit `wgpu::Limits::max_sampled_textures_per_shader_stage` in some environments.
HookApp::deferHook,
2024-10-01T19:43:18.219489Z ERROR wgpu_core::device::global: Device::create_render_pipeline error: Error matching ShaderStages(FRAGMENT) shader requirements against the pipeline
2024-10-01T19:43:18.219683Z ERROR wgpu::backend::wgpu_core: Handling wgpu errors as fatal by default
[...]

PR #15582, opened 2024-10-01
Synchronize removed components with the render world

### LINKED ISSUE #15560: Directional light does not disappear after removing the `DirectionalLight` component
Removing the `DirectionalLight` component does not remove the directional light from the render world

Caused by #15320

Noticed in the `clearcoat` example

kristoff3r:
# Objective

Fixes #15560
Fixes (most of) #15570

Currently a lot of examples (and presumably some user code) depend on toggling certain render features by adding/removing a single component to an entity, e.g. `SpotLight` to toggle a light. Because of the retained render world this no longer works: Extract will add any new components, but when it is removed the entity persists unchanged in the render world.

## Solution

Add `SyncComponentPlugin<C: Component>` that registers `SyncToRenderWorld` as a required component for `C`, and adds a component hook that will clear all components from the render world entity when `C` is removed. We add this plugin to `ExtractComponentPlugin` which fixes most instances of the problem. For custom extraction logic we can manually add `SyncComponentPlugin` for that component.

We also rename `WorldSyncPlugin` to `SyncWorldPlugin` so we start a naming convention like all the `Extract` plugins.

In this PR I also fixed a bunch of breakage related to the retained render world, stemming from old code that assumed that `Entity` would be the same in both worlds.

I found that using the `RenderEntity` wrapper instead of `Entity` in data structures when referring to render world entities makes intent much clearer, so I propose we make this an official pattern.

## Testing

Run examples like

```
cargo run --features pbr_multi_layer_material_textures --example clearcoat
cargo run --example volumetric_fog
```

and see that they work, and that toggles work correctly. But really we should test every single example, as we might not even have caught all the breakage yet.

---

## Migration Guide

The retained render world notes should be updated to explain this edge case and `SyncComponentPlugin`


crates/bevy_pbr/src/lib.rs
```diff
@@ -308,7 +309,6 @@ impl Plugin for PbrPlugin {
             .register_type::<PointLight>()
             .register_type::<PointLightShadowMap>()
             .register_type::<SpotLight>()
-            .register_type::<DistanceFog>()
```
kristoff3r, line 311: This is already registered in `FogPlugin` further down

kristoff3r:
I'll add and update documentation if we decide this is the solution we want

crates/bevy_pbr/src/render/mesh.rs
```diff
@@ -1056,7 +1059,11 @@ pub fn extract_meshes_for_gpu_building(
                 previous_input_index,
             };
 
-            queue.push(entity, gpu_mesh_instance_builder, gpu_mesh_culling_data);
+            queue.push(
+                *render_entity,
```
Trashtalk217, line 1063: I take it this is the same as calling `render_entity.id()`?

crates/bevy_pbr/src/render/mesh.rs
```diff
@@ -549,11 +550,11 @@ pub enum RenderMeshInstanceGpuQueue {
     /// The version of [`RenderMeshInstanceGpuQueue`] that omits the
     /// [`MeshCullingData`], so that we don't waste space when GPU
     /// culling is disabled.
-    CpuCulling(Vec<(Entity, RenderMeshInstanceGpuBuilder)>),
+    CpuCulling(Vec<(RenderEntity, RenderMeshInstanceGpuBuilder)>),
     /// The version of [`RenderMeshInstanceGpuQueue`] that contains the
     /// [`MeshCullingData`], used when any view has GPU culling
     /// enabled.
-    GpuCulling(Vec<(Entity, RenderMeshInstanceGpuBuilder, MeshCullingData)>),
+    GpuCulling(Vec<(RenderEntity, RenderMeshInstanceGpuBuilder, MeshCullingData)>),
```
Trashtalk217, line 557: Is this strictly necessary? Entity always refers to an entity in the world that you are in (in this case the render world), so they are already render entities.

crates/bevy_render/src/extract_component.rs
```diff
@@ -160,15 +160,6 @@ fn prepare_uniform_components<C>(
...
 /// To do so, it sets up the [`ExtractSchedule`] step for the specified [`ExtractComponent`].
-///
-/// # Warning
-///
-/// Be careful when removing the [`ExtractComponent`] from an entity. When an [`ExtractComponent`]
-/// is added to an entity, that entity is automatically synced with the render world (see also
-/// [`WorldSyncPlugin`](crate::world_sync::WorldSyncPlugin)). When removing the entity in the main
-/// world, the synced entity also gets removed. However, if only the [`ExtractComponent`] is removed
-/// this *doesn't* happen, and the synced entity stays around with the old extracted data.
-/// We recommend despawning the entire entity, instead of only removing [`ExtractComponent`].
```
Trashtalk217, line 171: There is a new exception now, but I'll put that somewhere else.

crates/bevy_render/src/gpu_readback.rs
```diff
@@ -232,7 +233,7 @@ fn prepare_buffers(
     mut buffer_pool: ResMut<GpuReadbackBufferPool>,
     gpu_images: Res<RenderAssets<GpuImage>>,
     ssbos: Res<RenderAssets<GpuShaderStorageBuffer>>,
-    handles: Query<(Entity, &Readback)>,
+    handles: Query<(&MainEntity, &Readback)>,
```
Trashtalk217, line 236: Hey look! It's a use for MainEntity, love to see it.

crates/bevy_render/src/world_sync.rs
```diff
@@ -161,11 +165,24 @@ pub(crate) fn entity_sync_system(main_world: &mut World, render_world: &mut Worl
...
-                EntityRecord::Removed(e) => {
-                    if let Some(ec) = render_world.get_entity_mut(e) {
+                EntityRecord::Removed(render_entity) => {
+                    if let Some(ec) = render_world.get_entity_mut(render_entity.id()) {
                         ec.despawn();
                     };
                 }
+                EntityRecord::ComponentRemoved(main_entity) => {
+                    // It's difficult to remove only the relevant component because component ids aren't stable across worlds,
+                    // so we just clear the entire render world entity.
```
Trashtalk217, line 175: This is fun and has consequences, but we're gonna figure that out.

Trashtalk217 (commented):
Broadly in favour of this, but there's one big exception that would break this system. This is exceedingly rare, but it does need to be documented.

Say that we have two synced and extracted components `A` and `B`. That are located on the same entity. 

```
| Main World                   | Render World               |
| -----------------------------|----------------------------|
| e1: (A, B, RenderEntity(e2)) | e2: (A, B, MainEntity(e1)) |
```
If you now remove `A` from `e1`, this will remove both `A` and `B` in the render world.
```
| Main World                   | Render World              |
| -----------------------------|---------------------------|
| e1: (B, RenderEntity(e2))    | e2: (MainEntity(e1))      |
```
This may not be what the user expects. And while this scenario may be very rare, it should be documented. You could solve this by counting the number of synced components a entity has (and storing it with SyncToRenderWorld), but that might be a step to far.

Also some documentation is still required for the `sync_component` plugin.

crates/bevy_pbr/src/render/mesh.rs
kristoff3r: It's not necessary, but when you look at a datatype out of context it's really hard to figure out which world the entities are from if it's not documented. When you're doing extract and sync systems you have access to entities from both worlds, and you could reasonably need to store either. 

I know `RenderEntity` wasn't intended for this, but using it as a newtype wrapper communicates intent really clearly, and because it's used from an extract system you have the `RenderEntity` component at hand and can just use it directly.

crates/bevy_pbr/src/render/mesh.rs
kristoff3r: No, because I changed this function to take a `RenderEntity`. See the motivation in the comment below.

crates/bevy_render/src/world_sync.rs
kristoff3r: After thinking more I think it's what we need to do in any case. Components like `SpotLight` aren't extracted directly but generate derived values, e.g. `ExtractedPointLight`. The only reason `SyncComponentPlugin` works in this case is because we clear everything.

Trashtalk217: I'm deferring on your judgement on this as I don't have much insight in rendering internals. As long as it's correctly documented (and maybe also include I reason why we have to do this, besides it being difficult, because it's strictly not impossible), this should be good.

kristoff3r: @IceSentry or another SME should perhaps sanity check this logic, I'm not an expert in the internals either but it makes sense to me :D

crates/bevy_render/src/extract_component.rs
kristoff3r: Which exception is that? I can include it in this PR if you want.

crates/bevy_render/src/world_sync.rs
```diff
@@ -13,6 +13,12 @@ use bevy_reflect::Reflect;
 
 /// A plugin that synchronizes entities with [`SyncToRenderWorld`] between the main world and the render world.
 ///
+/// [`SyncToRenderWorld`] is automatically added as a required components by
```
Trashtalk217, line 16:
```suggestion
/// [`SyncToRenderWorld`] is automatically added as a required component by
```

crates/bevy_render/src/world_sync.rs
```diff
@@ -13,6 +13,12 @@ use bevy_reflect::Reflect;
...
+/// [`ExtractComponentPlugin`] and [`SyncComponentPlugin`], so it doesn't need be
```
Trashtalk217, line 17:
```suggestion
/// [`ExtractComponentPlugin`] and [`SyncComponentPlugin`], so it doesn't need to be
```

crates/bevy_render/src/sync_component.rs
```diff
@@ -5,6 +5,20 @@ use bevy_ecs::component::Component;
...
+/// Plugin that registers a component for automatic sync to the render world. See [`WorldSyncPlugin`] for more information.
+///
+/// This plugin is automatically added by [`ExtractComponentPlugin`], and only needs to be added for manual extraction implementations.
+///
+/// # Implementation details
+///
+/// It adds [`SyncToRenderWorld`] as a required component to make the [`WorldSyncPlugin`] aware of the component, and
+/// handles cleanup of the component in the render world when it is removed from an entity.
+///
+/// NOTE: When the component is removed from the main world entity, all components are removed from the entity in the render world.
```
Trashtalk217, line 17: You could probably make this a proper `# Warning`, otherwise it's enough.

Trashtalk217 (changes requested):
Last couple nits, after this it's ready to go.

kristoff3r:
There are still a few open questions:
* Do we do the `WorldSync` to `SyncWorld` rename?
* Do we want the `SyncComponentPlugin` to add the required component registration for `SyncToRenderWorld` component? If  yes we should remove all the ones that have been added in the required components migrations. If no we should re-add the ones removed here. In any case we should find some place to document it.

Also it looks like some of the examples broke when I merged main, I'll have to debug that later.

crates/bevy_render/src/sync_world.rs
```diff
@@ -161,11 +184,24 @@ pub(crate) fn entity_sync_system(main_world: &mut World, render_world: &mut Worl
...
                     };
                 }
+                EntityRecord::ComponentRemoved(main_entity) => {
+                    let Some(mut render_entity) = world.get_mut::<RenderEntity>(main_entity) else {
+                        continue;
+                    };
+                    if let Some(render_world_entity) = render_world.get_entity_mut(render_entity.id()) {
+                        // In order to handle components that extract to derived components, we clear the entity
+                        // and let the extraction system re-add the components.
+                        render_world_entity.despawn();
```
IceSentry, line 199: For now this is probably fine since we don't really take advantage of the retained render world yet, but that feels like a major footgun for when we start to do that.

kristoff3r:
@alice-i-cecile seems to work now, tested with a bunch of different examples. It has a tendency to break because of the required component migrations, so I hope we can get it in soon.

crates/bevy_render/src/sync_world.rs
IceSentry: Yeah, that's why I'm fine with it for now, but it is mildly scary

IceSentry (approved):

alice-i-cecile (approved):

PR #15586, opened 2024-10-02
Feature-gate all image formats

clarfonthey:
# Objective

Bevy supports feature gates for each format it supports, but several formats that it loads via the `image` crate do not have feature gates. Additionally, the QOI format is supported by the `image` crate and wasn't available at all. This fixes that.

## Solution

The following feature gates are added:

* `avif`
* `ff` (Farbfeld)
* `gif`
* `ico`
* `qoi`
* `tiff`

None of these formats are enabled by default, despite the fact that all these formats appeared to be enabled by default before. Since `default-features` was disabled for the `image` crate, it's likely that using any of these formats would have errored by default before this change, although this probably needs additional testing.

## Testing

The changes seemed minimal enough that a compile test would be sufficient.

## Migration guide

Image formats that previously weren't feature-gated are now feature-gated, meaning they will have to be enabled if you use them:

* `avif`
* `ff` (Farbfeld)
* `gif`
* `ico`
* `tiff`

Additionally, the `qoi` feature has been added to support loading QOI format images.

Previously, these formats appeared in the enum by default, but weren't actually enabled via the `image` crate, potentially resulting in weird bugs. Now, you should be able to add these features to your projects to support them properly.

----

If you were individually configuring the `bevy_render` crate, the feature flags for the general image formats were moved to `bevy_image` instead. For example, `bevy_render/png` no longer exists, and `bevy_image/png` is the new location for this. The texture formats are still available on `bevy_render`, e.g. `bevy_render/ktx2` is needed to fully enable `ktx2` support, and this will automatically enable `bevy_image/ktx2` for loading the textures.


Cargo.toml
```diff
@@ -100,37 +100,39 @@ unused_qualifications = "warn"
...
-  "tonemapping_luts",
+  "qoi",
   "smaa_luts",
-  "default_font",
-  "webgl2",
   "sysinfo_plugin",
+  "tonemapping_luts",
+  "vorbis",
+  "webgl2",
+  "x11",
```
clarfonthey, line 135: This was sorted since it was half-sorted, then in a random order at the end. I controversially decided to include bevy crates first, then everything else sorted alphabetically.

Only `gif` and `qoi` were added, nothing else was changed.

Cargo.toml
```diff
@@ -240,38 +242,53 @@ trace_tracy_memory = [
...
 
-# KTX2 compressed texture support
-ktx2 = ["bevy_internal/ktx2"]
+# TIFF image format support
+tiff = ["bevy_internal/tiff"]
 
-# PNM image format support, includes pam, pbm, pgm and ppm
-pnm = ["bevy_internal/pnm"]
+# WebP image format support
+webp = ["bevy_internal/webp"]
```
clarfonthey, line 291: These were sorted alphabetically since it seemed confusing otherwise. Like mentioned, `avif`, `ff`, `gif`, `ico`, `qoi`, and `tiff` are the new ones.

Cargo.toml
```diff
@@ -100,38 +100,40 @@ unused_qualifications = "warn"
...
+  "qoi",
   "smaa_luts",
-  "default_font",
-  "webgl2",
   "sysinfo_plugin",
-  "android-game-activity",
+  "tonemapping_luts",
+  "vorbis",
+  "webgl2",
+  "x11",
```
clarfonthey, line 136: This was sorted since it was half-sorted, then in a random order at the end. I controversially decided to include bevy crates first, then everything else sorted alphabetically.

Only `gif` and `qoi` were added, nothing else was changed.

Cargo.toml
```diff
@@ -241,38 +243,56 @@ trace_tracy_memory = [
...
 
-# KTX2 compressed texture support
-ktx2 = ["bevy_internal/ktx2"]
+# TIFF image format support
+tiff = ["bevy_internal/tiff"]
 
-# PNM image format support, includes pam, pbm, pgm and ppm
-pnm = ["bevy_internal/pnm"]
+# WebP image format support
+webp = ["bevy_internal/webp"]
```
clarfonthey, line 295: These were sorted alphabetically since it seemed confusing otherwise. Like mentioned, `avif`, `ff`, `gif`, `ico`, `qoi`, and `tiff` are the new ones.

docs/cargo_features.md
```diff
@@ -11,9 +11,6 @@ The default feature set enables most of the expected features of a game engine,
 
 |feature name|description|
 |-|-|
-|android-game-activity|Android GameActivity support. Default, choose between this and `android-native-activity`.|
-|android_shared_stdcxx|Enable using a shared stdlib for cxx on Android|
-|animation|Enable animation support, and glTF animation loading|
```
clarfonthey, line 16: Decided to do the same sorting as in `Cargo.toml` and move bevy crates to the top; more than happy to revert this change if preferred.

Cargo.toml
```diff
@@ -100,38 +100,40 @@ unused_qualifications = "warn"
...
+  "qoi",
   "smaa_luts",
-  "default_font",
-  "webgl2",
   "sysinfo_plugin",
-  "android-game-activity",
+  "tonemapping_luts",
+  "vorbis",
+  "webgl2",
+  "x11",
```
clarfonthey, line 136: This was sorted since it was half-sorted, then in a random order at the end. Only `gif` and `qoi` were added, nothing else was changed.

Cargo.toml
alice-i-cecile: I like sorting these, as long as they're grouped.

clarfonthey: Yup, this is only sorted inside the image formats group; everything else was left alone.

pcwalton:
My personal opinion is that QOI isn't popular enough to warrant being enabled by default.

JMS55:
Agreed. For GIF, do we have any gif playback support in the engine?

I also think we should push people towards not just common formats, but the best formats.

clarfonthey:
GIFs can be loaded as an array which feels somewhat nice, and QOI has a very small and simple spec which is nice. But I can remove both by default if that feels appropriate.

Kinda just would prefer the path of least friction and we can change the defaults later.

mockersf:
> For GIF, do we have any gif playback support in the engine?

Not built in, I have https://crates.io/crates/vleue_kinetoscope as a third party plugin to play animated gifs

alice-i-cecile:
Agreed on wanting both GIF and QOI off by default :)

Cargo.toml
```diff
@@ -100,38 +100,38 @@ unused_qualifications = "warn"
...
-  "tonemapping_luts",
   "smaa_luts",
-  "default_font",
-  "webgl2",
   "sysinfo_plugin",
-  "android-game-activity",
+  "tonemapping_luts",
+  "vorbis",
+  "webgl2",
+  "x11",
```
clarfonthey, line 134: Decided to still sort these despite this PR no longer changing the defaults. So, this literally just sorts without adding anything new.

clarfonthey:
Updated so now none of the newly-added gates are enabled by default. I think that in the future, we should leverage the animation loader from the `image` crate to load in all frames of images when loading (maybe adding as an image loader setting to disable) but that can be done in a future change.

alice-i-cecile:
I *think* this is a breaking change, but let me know if I'm mistaken please.

clarfonthey:
It's in the grey area of being a technically breaking change that probably won't affect anyone.

Like I said, because `image` is set with `default-features = false`, these features weren't actually included if you tried to load them anyway, and probably would have failed. But, technically, this does remove these enum variants by default, so, probably better safe than sorry.

If we get this in before 0.15 then I would say it's okay to include in the migration guide just to be safe, but if it doesn't, then I would say it's reasonable to include in 0.15.1 with just a warning.

alice-i-cecile (approved):

tychedelia (approved):

alice-i-cecile:
@clarfonthey can you please resolve merge conflicts?

clarfonthey:
So uh, would have kind have liked to merge this before the `bevy_image` split, or at least had that incorporate this change, since I kinda have to duplicate all of the feature changes onto `bevy_image` too now, and it's not clear exactly what the best way to do that is.

But I'll see what I can do.

clarfonthey:
Okay, I've made the changes and they also involved adding a bit more code to the PR, so, it should probably just have a new pass on review.

Essentially, I wanted `bevy_render` to no longer contain any of the feature flags for image formats, and it turns out that this was pretty easy by just moving the list of file extensions over from `bevy_render` to `bevy_image`, offering it as an associated constant there instead.

This adds a few new APIs and the consteval code is a *bit* ugly, but it works and feels like the best way to do it for now.

clarfonthey:
Right, okay, I may have been overzealous. Will work on that still for a bit and bug people when I've gotten it working.

Cargo.toml
```diff
@@ -100,39 +100,40 @@ unused_qualifications = "warn"
...
   "smaa_luts",
-  "default_font",
-  "webgl2",
   "sysinfo_plugin",
-  "android-game-activity",
-  "custom_cursor",
+  "tonemapping_luts",
+  "vorbis",
+  "webgl2",
+  "x11",
```
clarfonthey, line 136: Still just sorted, nothing special.

crates/bevy_image/src/image.rs
```diff
@@ -71,98 +78,341 @@ macro_rules! feature_gate {
...
+            ImageFormat::WebP => &["image/webp"],
+        }
+    }
+
     pub fn from_mime_type(mime_type: &str) -> Option<Self> {
         Some(match mime_type.to_ascii_lowercase().as_str() {
-            "image/avif" => ImageFormat::Avif,
+            // note: farbfeld does not have a MIME type
+            "image/avif" => feature_gate!("avif", Avif),
+            "image/basis" | "image/x-basis" => feature_gate!("basis-universal", Basis),
```
clarfonthey, line 313: This was omitted before; I tried looking and it appears that basis-universal is in the same state as QOI, where people kind of assume that a MIME type exists but there isn't actually an official one filed. So, that's why I included both here.

crates/bevy_render/Cargo.toml
```diff
@@ -9,31 +9,18 @@ license = "MIT OR Apache-2.0"
...
-ddsfile = ["bevy_image/ddsfile"]
-ktx2 = ["dep:ktx2", "bevy_image/ktx2"]
-flate2 = ["bevy_image/flate2"]
-ruzstd = ["bevy_image/ruzstd"]
+# Texture formats (require more than just image support)
 basis-universal = ["dep:basis-universal", "bevy_image/basis-universal"]
+dds = ["bevy_image/dds"]
+exr = ["bevy_image/exr"]
+hdr = ["bevy_image/hdr"]
+ktx2 = ["dep:ktx2", "bevy_image/ktx2"]
```
clarfonthey, line 17: I didn't explicitly check that all of these were actually used by `bevy_render` but I figured that it's good to just keep them all here in case any code gets added that does special-case them. We don't care about generic image storage formats like PNG, but the texture formats might have extra rendering info we need here.

atlv24 (approved):
I love this refactor, thank you so much

clarfonthey:
It has passed! Ready for review now.

PR #15341, opened 2024-09-20
Add UI `GhostNode`

villor:
# Objective

- Fixes #14826 
- For context, see #15238

## Solution
Add a `GhostNode` component to `bevy_ui` and update all the relevant systems to use it to traverse for UI children.

- [x] `ghost_hierarchy` module
  - [x] Add `GhostNode`
  - [x] Add `UiRootNodes` system param for iterating (ghost-aware) UI root nodes
  - [x] Add `UiChildren` system param for iterating (ghost-aware) UI children
- [x] Update `layout::ui_layout_system`
  - [x] Use ghost-aware root nodes for camera updates
  - [x] Update and remove children in taffy
    - [x] Initial spawn
    - [x] Detect changes on nested UI children
  - [x] Use ghost-aware children traversal in `update_uinode_geometry_recursive`
- [x] Update the rest of the UI systems to use the ghost hierarchy
  - [x] `stack::ui_stack_system`
  - [x] `update::`
    - [x] `update_clipping_system`
    - [x] `update_target_camera_system`
  - [x] `accessibility::calc_name`

## Testing
- [x] Added a new example `ghost_nodes` that can be used as a testbed.
- [x] Added unit tests for _some_ of the traversal utilities in `ghost_hierarchy`
- [x] Ensure this fulfills the needs for currently known use cases
  - [x] Reactivity libraries (test with `bevy_reactor`)
  - [ ] Text spans (mentioned by koe [on discord](https://discord.com/channels/691052431525675048/1285371432460881991/1285377442998915246))
  
---
## Performance
[See comment below](https://github.com/bevyengine/bevy/pull/15341#issuecomment-2385456820)

## Migration guide
Any code that previously relied on `Parent`/`Children` to iterate UI children may now want to use `bevy_ui::UiChildren` to ensure ghost nodes are skipped, and their first descendant Nodes included.

UI root nodes may now be children of ghost nodes, which means `Without<Parent>` might not query all root nodes. Use  `bevy_ui::UiRootNodes` where needed to iterate root nodes instead.

## Potential future work
- Benchmarking/optimizations of hierarchies containing lots of ghost nodes
- Further exploration of UI hierarchies and markers for root nodes/leaf nodes to create better ergonomics for things like `UiLayer` (world-space ui)

PR #15591, opened 2024-10-02
Text rework

UkoeHB:
**Ready for review. Examples migration progress: 100%.**

# Objective

- Implement https://github.com/bevyengine/bevy/discussions/15014

## Solution

This implements [cart's proposal](https://github.com/bevyengine/bevy/discussions/15014#discussioncomment-10574459) faithfully except for one change. I separated `TextSpan` from `TextSpan2d` because `TextSpan` needs to require the `GhostNode` component, which is a `bevy_ui` component only usable by UI.

Extra changes:
- Added `EntityCommands::commands_mut` that returns a mutable reference. This is a blocker for extension methods that return something other than `self`. Note that `sickle_ui`'s `UiBuilder::commands` returns a mutable reference for this reason.

## Testing

- [x] Text examples all work.

---

## Showcase

TODO: showcase-worthy

## Migration Guide

TODO: very breaking

### Accessing text spans by index

Text sections are now text sections on different entities in a hierarchy, Use the new `TextReader` and `TextWriter` system parameters to access spans by index.

Before:
```rust
fn refresh_text(mut query: Query<&mut Text, With<TimeText>>, time: Res<Time>) {
    let text = query.single_mut();
    text.sections[1].value = format_time(time.elapsed());
}
```

After:
```rust
fn refresh_text(
    query: Query<Entity, With<TimeText>>,
    mut writer: UiTextWriter,
    time: Res<Time>
) {
    let entity = query.single();
    *writer.text(entity, 1) = format_time(time.elapsed());
}
```

### Iterating text spans

Text spans are now entities in a hierarchy, so the new `UiTextReader` and `UiTextWriter` system parameters provide ways to iterate that hierarchy. The `UiTextReader::iter` method will give you a normal iterator over spans, and `UiTextWriter::for_each` lets you visit each of the spans.


viridia:
Note that my original conception of ghost nodes is that they would work in all rendering contexts, not just UI, however this is somewhat controversial.

crates/bevy_ui/src/widget/text.rs
```diff
@@ -1,49 +1,204 @@
...
+    Node,
+    Style, // TODO: Remove when Node uses required components.
+    ContentSize, // TODO: Remove when Node uses required components.
+    FocusPolicy, // TODO: Remove when Node uses required components.
+    ZIndex, // TODO: Remove when Node uses required components.
+    BackgroundColor, // TODO: Remove when Node uses required components.
+    Visibility, // TODO: Remove when Node uses required components.
+    Transform // TODO: Remove when Node uses required components.
+)]
+pub struct TextNEW(pub String);
```
ickshonpe, line 106: Why is  `NEW` capitalized, is this a naming convention we are using with required components or something?

crates/bevy_text/src/text.rs
```diff
@@ -23,160 +24,152 @@ impl Default for CosmicBuffer {
...
+/// to [`TextLayoutInfo`] for rendering.
+///
+/// See [`Text2d`] for the core component of 2d text, and `Text` in `bevy_ui` for UI text.
 #[derive(Component, Debug, Clone, Default, Reflect)]
 #[reflect(Component, Default, Debug)]
-pub struct Text {
-    /// The text's sections
-    pub sections: Vec<TextSection>,
+#[require(ComputedTextBlock, TextLayoutInfo)]
+pub struct TextBlock {
```
ickshonpe, line 105: The name TextBlock doesn't seem very intuitive, the name implies to me it would be the primary text component holding the string, not the arrangment settings?

crates/bevy_text/src/text.rs
```diff
@@ -23,160 +24,152 @@ impl Default for CosmicBuffer {
...
     /// The text's internal alignment.
     /// Should not affect its position within a container.
     pub justify: JustifyText,
-    /// How the text should linebreak when running out of the bounds determined by `max_size`
+    /// How the text should linebreak when running out of the bounds determined by `max_size`.
     pub linebreak: LineBreak,
     /// The antialiasing method to use when rendering text.
     pub font_smoothing: FontSmoothing,
```
ickshonpe, line 112: I think don't think the font smoothing setting belongs here, it doesn't affect the arrangement of the text. It would be better stored in an elidable component of its own or in `TextStyle`.

crates/bevy_text/src/text.rs
```diff
@@ -216,9 +209,10 @@ impl From<JustifyText> for cosmic_text::Align {
     }
 }
 
-#[derive(Clone, Debug, Reflect)]
-/// `TextStyle` determines the style of the text in a section, specifically
+/// `TextStyle` determines the style of a text span within a [`TextBlock`], specifically
 /// the font face, the font size, and the color.
+#[derive(Component, Clone, Debug, Reflect)]
+#[reflect(Component, Default, Debug)]
 pub struct TextStyle {
```
ickshonpe, line 216: The color should be in a separate component `TextColor`, so that changes to the color don't trigger change detection and a relayout of the text.

ickshonpe: Maybe `TextStyle` should be renamed as well to `TextFont` or something. It would be consistant maybe with the UI `Style` component if `TextBlock` was renamed to `TextStyle`, but `Style` should be renamed too.

crates/bevy_text/src/text.rs
```diff
@@ -23,160 +24,152 @@ impl Default for CosmicBuffer {
...
+}
+
+/// Component with text format settings for a block of text.
+///
+/// A block of text is composed of text spans, which each have a separate string value and [`TextStyle`]. Text
+/// spans associated with a text block are collected into [`ComputedTextBlock`] for layout, and then inserted
+/// to [`TextLayoutInfo`] for rendering.
+///
+/// See [`Text2d`] for the core component of 2d text, and `Text` in `bevy_ui` for UI text.
 #[derive(Component, Debug, Clone, Default, Reflect)]
```
ickshonpe, line 102: It's not very big, could be copy?

```suggestion
#[derive(Component, Debug, Copy, Clone, Default, Reflect)]
```

ickshonpe (commented):
Performance seems improved over main slightly overall, maybe ~5% in the stress tests on recomputations. Without recomputation it's slightly slower but that's to be expected because of the overhead of the extra nodes in the tree.

crates/bevy_text/src/text2d.rs
```diff
@@ -1,68 +1,177 @@
...
+        font: font_handle.into(),
+        font_size: 60.0,
+        color: RED.into(),
+    }
+));
+```
+*/
+#[derive(Component, Clone, Debug, Default, Deref, DerefMut, Reflect)]
+#[reflect(Component, Default, Debug)]
+#[require(TextStyle, Visibility(visibility_hidden), Transform)]
```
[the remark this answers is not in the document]
ickshonpe: Same thing with `Visibility`, I don't think it's useful for text.

crates/bevy_ui/src/widget/text.rs
```diff
@@ -1,49 +1,204 @@
...
+# use bevy_color::palettes::basic::BLUE;
+# use bevy_ecs::World;
+# use bevy_text::{Font, JustifyText, TextBlock, TextStyle};
+# use bevy_ui::Text;
+#
+# let font_handle: Handle<Font> = Default::default();
+# let mut world = World::default();
+#
+// Basic usage.
+world.spawn(Text::new("hello world!"));
```
Olle-Lukowski, line 72: I'm assuming that `TextNEW` is some sort of temporary name, since here you show it as `Text`?

Olle-Lukowski (approved):
Other than the fact that I don't think `TextNEW` is a good name (I don't think this was intended even), this LGTM!

crates/bevy_ui/src/widget/text.rs
ickshonpe, line 106: I really dislike how the first span is a different type. I'd prefer if it just required the first text span to be a child too.
Could we do something like put the `require` on a marker component and have a constructor that returns a tuple `(Marker, TextSpan)`?

ickshonpe, line 106: Another thing would be to call it `TextRootSpan`. it's uglier but less confusing.

crates/bevy_ui/src/widget/text.rs
```diff
@@ -1,49 +1,204 @@
...
+        font: font_handle.into(),
+        font_size: 60.0,
+        color: RED.into(),
+    }
+));
+```
+*/
+#[derive(Component, Debug, Default, Clone, Deref, DerefMut, Reflect)]
+#[reflect(Component, Default, Debug)]
+#[require(TextStyle, GhostNode, Visibility(hidden_visibility))]
```
ickshonpe, line 172: Like with `TextSpan2d` why do we require `Visibility` here I don't think it's very useful to set Visibility for individual text spans and it doesn't seem like that's supported anyway.

ickshonpe, line 172: I'd prefer if `TextStyle` were elidable and if missing it just uses the `TextStyle` of its parent or the root text span.

examples/ui/text_debug.rs
```diff
@@ -106,124 +106,121 @@ fn infotext_system(mut commands: Commands, asset_server: Res<AssetServer>) {
...
-                TextSection::new(
-                    "\nThis text changes in the bottom right - ",
-                    TextStyle {
-                        font: font.clone(),
-                        font_size: 21.0,
-                        color: RED.into(),
-                    },
-                ),
-                TextSection::from_style(TextStyle {
+            TextNEW::new(""),
```
ickshonpe, line 152: There really needs to be a text node constructor function that takes an iterator of strings as this stuff is torturous.
It's especially annoying if you have a list of items when you need to split off the head of the list and add it with `TextNEW`.

crates/bevy_ui/src/widget/text.rs
ickshonpe: Oh right, `GhostNode` already requires `Visibility` and `Transform`.

crates/bevy_ui/src/widget/text.rs
UkoeHB: Yes to force compile errors on the old `Text` to make sure I check their use (since 2d and UI are split now).

crates/bevy_text/src/text.rs
UkoeHB: It's not great but this is what we have in the [rework discussion](https://github.com/bevyengine/bevy/discussions/15014) so I will leave it as-is unless something better is proposed.

crates/bevy_text/src/text.rs
UkoeHB: Moved to TextStyle since smoothing is applied per-font. Can break it into a standalone component in a separate PR.

crates/bevy_text/src/text2d.rs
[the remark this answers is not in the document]
UkoeHB: You can't have entities in a spatial hierarchy without Visibility/Transform.

crates/bevy_ui/src/widget/text.rs
UkoeHB: The original proposal had a universal span component but cart wanted to consolidate: https://github.com/bevyengine/bevy/discussions/15014#discussioncomment-10574459

crates/bevy_text/src/text2d.rs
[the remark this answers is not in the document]
ickshonpe: I thought that this was changes so that they were just ignored now? After removing the requirements the examples still seem to work fine.

crates/bevy_text/src/text2d.rs
```diff
@@ -1,68 +1,182 @@
...
+pub struct TextSpan2d(pub String);
```
ickshonpe, line 159: Missing a `new` method.

crates/bevy_text/src/text2d.rs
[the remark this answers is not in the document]
UkoeHB: Oh was it? I must have missed that.

crates/bevy_text/src/text.rs
```diff
@@ -279,6 +279,21 @@ pub enum FontSmoothing {
     // SubpixelAntiAliased,
 }
 
+/// Provides convenience methods for constructing text blocks.
+pub trait TextBuilderExt {
+    /// Adds a flat list of spans as children of the current entity.
+    fn with_spans<S: Component>(&mut self, spans: Vec<(S, TextStyle)>) -> &mut Self;
```
ickshonpe, line 285: It might be more ergnomic if the `spans` parameter is `impl IntoIterator<Item = (String, TextStyle)>`, and the string gets wrapped automatically.

crates/bevy_text/src/text.rs
```diff
@@ -279,6 +279,21 @@ pub enum FontSmoothing {
...
+}
+
+impl TextBuilderExt for EntityCommands<'_> {
+    fn with_spans<S: Component>(&mut self, mut spans: Vec<(S, TextStyle)>) -> &mut Self {
```
ickshonpe, line 289: This is better but it still leaves the awkwardness where you have to spawn the first element of a list separately as a `TextNEW` entity.
I was thinking of a builder that spawns the root `TextNew` entity as well.

crates/bevy_text/src/text.rs
ickshonpe: `impl IntoIterator<Item = (impl Into<String>, TextStyle)>` even.

ickshonpe (commented):
There should be a warning or error if we encounter orphaned/misparented `TextSpan` entities. It seems like an easy mistake to make, like you create a text node, but add the child text_spawn to the wrong parent.

crates/bevy_text/src/text.rs
UkoeHB: Added `spawn_text_block`

crates/bevy_text/src/text.rs
UkoeHB: Fixed in https://github.com/bevyengine/bevy/pull/15591/commits/a12f0a6fac29e00b37fdd0299130fd3117b3333e

crates/bevy_ui/src/widget/text.rs
UkoeHB: I like this idea but want to see what @cart has to say.

crates/bevy_ui/src/widget/text.rs
```diff
@@ -1,49 +1,231 @@
...
```
cart, line 180: Can we discuss why `GhostNode` is here? My theory is that this is to support scenarios like `Text -> TextSpan -> Node -> TextSpan`.

1. It feels like `TextSpan` should always be a "leaf" from a UI perspective, other than having other child TextSpans. I can't think of a scenario where arbitrary hierarchy like this makes sense.
2. `GhostNode` is still a bit contentious / we're considering it experimental. I'd prefer it if we didn't take a dependency on it for "real things" yet.
3. By removing the GhostNode requirement, we can go back to having a single TextSpan type that works across contexts, which seems much nicer to me.

crates/bevy_dev_tools/src/fps_overlay.rs
```diff
@@ -95,35 +97,36 @@ fn setup(mut commands: Commands, overlay_config: Res<FpsOverlayConfig>) {
...
-        .with_children(|c| {
-            c.spawn((
-                TextBundle::from_sections([
-                    TextSection::new("FPS: ", overlay_config.text_config.clone()),
-                    TextSection::from_style(overlay_config.text_config.clone()),
-                ]),
-                FpsText,
-            ));
-        });
+        .spawn_text_block::<TextNEW>([
```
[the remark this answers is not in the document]
UkoeHB: > Given that one of the motivations behind this refactor is to make text easily representable as raw components, I find it a bit concerning that we're building new special case abstractions on top. I would prefer it if we removed this and used the normal spawn APIs to create the components directly.

The raw API makes spawning big lists of spans much more painful. I've used `spawn_text_block` multiple times while migrating examples and it is very helpful. If you want to aim at a different API I suggest grepping `spawn_text_block` in this branch and trying to migrate some of the heavy-user-examples to your vision of raw spawning (one example has 22 or more spans). If you can make it not painful, then I'll accept suggestions without resistance.

> I think this should be replaced with the original impl from this PR (directly spawning the components), and the helper method should be removed.

Doing that will make this PR a regression for existing users, rather than a step forward (or at least sideways). I would be fine with removing these ergonomic workarounds *after* viable alternatives are available.

crates/bevy_ecs/src/system/commands/mod.rs
```diff
@@ -1498,6 +1498,11 @@ impl EntityCommands<'_> {
         self.commands.reborrow()
     }
 
+    /// Returns a mutable reference to the underlying [`Commands`].
+    pub fn commands_mut(&mut self) -> &mut Commands<'a, 'a> {
```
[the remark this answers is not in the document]
UkoeHB: > This feels weirdly asymmetric with commands(). Given that this was added to facilitate the creation of spawn_text_block, and given that I think that should be removed, I think we should remove this as well. I see the need its solving, but I'd prefer to punt this problem unless we have to solve it here.

This API limitation is a headache downstream as well. I will open a separate PR if it ends up getting removed here.

crates/bevy_text/src/text.rs
UkoeHB: `TextBlock` is pretty awkward in practice, from my experience migrating examples. `TextLayout` would make a lot more sense. You don't actually add `TextBlock` to an entity and get a text block... you get a partial text block that needs a root text component and text style to function.

crates/bevy_ui/src/widget/text.rs
UkoeHB: > If we aren't already doing it, we could make text.0 for the default empty string behave as "no span", so the first child of Text::default() -> TextSpan is at index 0.

I disagree with this, it creates ambiguity and conceptual branching.

crates/bevy_text/src/text.rs
```diff
@@ -293,3 +278,163 @@ pub enum FontSmoothing {
...
+                warn_once!("found entity {:?} with a text span ({}) that is part of a broken hierarchy with a Parent \
+                    component that points at non-existent entity {:?}; this warning only prints once",
+                    entity, std::any::type_name::<Span>(), parent);
+                break;
+            };
+            if let Some(mut computed) = maybe_computed {
+                computed.needs_rerender = true;
+                break;
+            }
+            let Some(next_parent) = maybe_parent else {
```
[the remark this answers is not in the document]
UkoeHB: Yeah this is a bug thanks, will fix.

crates/bevy_text/src/text2d.rs
```diff
@@ -1,69 +1,209 @@
...
+
+impl TextSpan2d {
+    /// Makes a new 2d span text component.
+    pub fn new(text: impl Into<String>) -> Self {
+        Self(text.into())
+    }
+
+    /// Makes an empty 2d span text component.
+    pub fn empty() -> Self {
```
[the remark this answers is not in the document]
UkoeHB: Yep. It expresses intent and is shorter to type (5 vs 7 letters). Will remove if necessary.

crates/bevy_ui/src/widget/text.rs
UkoeHB: > Can we discuss why GhostNode is here?

We need text spans to not participate in layout. The alternative is `Display::None` but then you start lugging around useless `Node` and `Style` components. `GhostNode` is currently the most well-suited to the task.

> By removing the GhostNode requirement, we can go back to having a single TextSpan type that works across contexts, which seems much nicer to me.

Not without some serious design analysis, since we need a good way to skip entities during layout. See [my comment](https://github.com/bevyengine/bevy/pull/15341#issuecomment-2392344534) on the GhostNode PR.

> I can't think of a scenario where arbitrary hierarchy like this makes sense.

In the distant future we may have 'in-line UI' in text blocks. But there is no design around that now so I agree text spans can be considered simple leafs.

crates/bevy_text/src/text.rs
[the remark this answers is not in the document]
UkoeHB: Fixed in https://github.com/bevyengine/bevy/pull/15591/commits/7e1d3bcc31c0018885a3575648e8a242a5ad9610

crates/bevy_text/src/text.rs
UkoeHB: This is fine with me but can we do it in a separate PR? This one is already quite huge. The migration from TextStyle to TextFont/TextColor will by itself be a big chore and I call 'not it'.

crates/bevy_ui/src/widget/text.rs
eidloi: > We need text spans to not participate in layout. 

Why _wouldn't_ you want them to participate in UI layout? The main text node _defines_ the available / desired space for text, which communicates sizes both upwards and downwards the tree.

UkoeHB: > Why wouldn't you want them to participate in UI layout? The main text node defines the available / desired space for text, which communicates sizes both upwards and downwards the tree.

I lied a little, they *do* participate in layout, but not directly. Spans are collected into a `cosmic-text` buffer which knows how to manage text, and then the buffer is 'measured' when doing taffy layout (taffy needs to ignore spans).

UkoeHB: However, I'm leaning toward the idea that text spans as individual entities is not worth the trouble. Perhaps they can just be consolidated into a component and then the whole question of ghost nodes becomes moot.

UkoeHB: @eidloi one reason to use a hierarchy for spans is being able to reuse the hierarchy-based theming from `sickle_ui`. You can set `target` to a span entity to animate span color or use responsive font.

It looks like [this PR](https://github.com/bevyengine/bevy/pull/15635) introduces tools for making hierarchy structures other than Parent/Child. I think if that gets merged then there is a way to make text spans part of an 'ancillary hierarchy' instead of directly in-tree with UI.

crates/bevy_text/src/text.rs
cart: Yup thats fine by me.

crates/bevy_text/src/text.rs
cart: > you get a partial text block that needs a root text component and text style to function.

Not super important, but when I pitched this design I intended for TextBlock to be the "source of truth text span receiver component" that served its function independently of specific "root text" components. Of course I didn't explicitly call that out and I think the current impl is totally fine. But if we went that route, it would make TextBlock make sense (and would also probably make building future TextBlock types easier). Plenty of open questions there though in terms of implementation / I don't think we need to go down that road in this PR.

crates/bevy_ui/src/widget/text.rs
eidloi: yes, it is the reason I also researched and PoC-d a text create, and yes, that PR is the proper solution instead of ghost nodes. Except that is mile-stoned for 0.16.0. so maybe next year

Also, I think text measurement should be provided with a bounding box which should be coming from the node constraints. The only issue here is that the constraints don't exist before text is measured so to me the problem really is that calcs are done late and only once.

I would also like to point out that the entity per span is not the entire solution for text, as you cannot make a GPU renderer for it if its chopped up in entities (not without a render texture rendering an entire UI block - but that prevents splatting text on any part of the UV). However, if text is chopped up and has its own hierarchy, GPU rendering is much easier once you grab the entire text tree.

cart: > We need text spans to not participate in layout. The alternative is Display::None but then you start lugging around useless Node and Style components. GhostNode is currently the most well-suited to the task.

TextSpans aren't UI Nodes (or at least, have no reason to be as they are "metadata" / not rendered directly). Why would they participate in layout?

> In the distant future we may have 'in-line UI' in text blocks. But there is no design around that now (it may not be in-tree, or may use different relations) so I agree text spans can be considered simple leafs.

Even in those cases, I think it probably could serve as an independent UI root. Id prefer to punt that scenario down the road until we actually choose to implement it, in favor of the simpler design here.

I think for this PR we should:

1. Remove the GhostNode require
2. Remove the "arbitrary propagation of TextSpan" up the hierarchy. TextSpans only propagate up other TextSpans until they either hit a TextBlock (success) or hit a parent _without_ a TextSpan or TextBlock (invalid).
3. Remove TextSpan2d and embrace TextSpan as the "one context-free span type".

If you're burnt out on changes let me know. Happy to hop in.

crates/bevy_ui/src/widget/text.rs
cart: > I disagree with this, it creates ambiguity and conceptual branching. Note that it's common to set a default-empty span and then later write text to it.

I don't have strong opinions on that. Just pitching an alternative path that allows `Text::default()->TextSpan` to behave as if the first TextSpan child is actually the first span. I do agree that it makes the "empty Text to non-empty" change harder to reason about, as it would behave differently than an empty child TextSpan.

crates/bevy_ui/src/widget/text.rs
UkoeHB: > TextSpans aren't UI Nodes (or at least, have no reason to be as they are "metadata" / not rendered directly). Why would they participate in layout?

They are not, but they are *in-tree* and the layout algorithm walks the tree. Currently we issue a warning (or panic?) if any in-tree entity is missing `Node`. Whether or not to 'ignore' entities without Node is part of [the discussion](https://github.com/bevyengine/bevy/pull/15341#issuecomment-2392344534) around ghost nodes. I'm leaning toward EidLoi's position that hierarchies should be self-contained structures and that non-uniform entities should be attached with tailored relationships. So, text spans should be attached as an ancillary hierarchy that doesn't use Parent/Child (maybe `SpanParent`/`SpanChild`).

> Remove the "arbitrary propagation of TextSpan" up the hierarchy. 

This was already fixed, it was a bug.

> Remove TextSpan2d and embrace TextSpan as the "one context-free span type".

Only possible if changes are maybe to UI layout to make ghost nodes implicit instead of/in addition to explicit.

> If you're burnt out on changes let me know. Happy to hop in.

Not quite yet but I probably won't do any big changes that I'm not confident are improving the situation.

cart: > Currently we issue a warning (or panic?) if any in-tree entity is missing Node.

I think UI nodes having non UI children should probably be valid. Its when a UI node has non-UI children _who then have UI nodes_ where we have problems.

> Whether or not to 'ignore' entities without Node is part of the discussion around ghost nodes. I'm leaning toward EidLoi's position that hierarchies should be self-contained structures and that non-uniform entities should be attached with tailored relationships. So, text spans should be attached as an ancillary hierarchy that doesn't use Parent/Child (maybe ParentSpan/ChildSpan).

I agree that this is a path worth exploring. But it feels pretty non-viable without both relations and scene construction tools that make multi-hierarchy management (aka multi-relation construction) easier. The current Bevy paradigm is "one hierarchy" and we should design under that constraint until that changes.

> This was already fixed, it was a bug.

Excellent.

> Only possible if changes are maybe to UI layout to make ghost nodes implicit instead of/in addition to explicit. Or if we introduce a different relationship structure.

Having thought on this, I think the path forward is removing GhostNode and making replacing the `no UI->NonUI` layout constraint with a `no UI->NonUi->UI` constraint. Then we unify on a single TextSpan type as it can be a non-ui contextless component.

Not a fan of using GhostNode to enable these "leaf-like" cases because we're heavily considering reverting that changeset (it is still experimental), and I think these "leaf-like" cases should be valid.

UkoeHB: > Having thought on this, I think the path forward is removing GhostNode and making replacing the no UI->NonUI layout constraint with a no UI->NonUi->UI constraint. Then we unify on a single TextSpan type as it can be a non-ui contextless component.

This is fine with me, but changing the layout algorithm should not be part of this PR. Can you open a new PR supporting the changes you want? Then I will rebase once that's merged and consolidate the span types.

crates/bevy_dev_tools/src/fps_overlay.rs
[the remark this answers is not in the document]
cart: I've just pushed a commit that removes spawn_text_block. In many places it actually improves legibility / ergonomics because it doesn't force style to be specified in non-default cases. It also really helps legibility in cases where we need to add additional components to the root node.

I know you probably still disagree, but I'm making an executive decision for the reasons I've stated above (and the new ones discovered during the port).

For the duration of this PR, lets stick to normal spawning. We can pick up the ergonomics question later. But imo those changes were in general a lateral move ergonomically and they came at the cost of obfuscation.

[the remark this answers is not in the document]
ickshonpe: Yeah I found that just updating the stress test examples was really painful, I think without something that just leaves users with the burden of having to invent their own adhoc builders and these changes aren't immediately easy to understand anyway.

`spawn_text_block` isn't a great solution I realised though, it doesn't allow users to add marker components for instance.

crates/bevy_ecs/src/system/commands/mod.rs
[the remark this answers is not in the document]
cart: I'm down to leave this in for now.

crates/bevy_text/src/text2d.rs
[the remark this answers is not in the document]
cart: I've removed this for now in favor of `default()`

cart (approved):
I've made some adjustments (as discussed):

1. Removed TextSpan2d and switched everything to use TextSpan.
2. Removed GhostNode usage
3. Removed `X::empty()` variants in favor of `X::default()`
4. Renamed `TextNEW` to `Text`
5. Merged with main
6. Ported new examples added since last merge

I think this is now good to go.

cart:
(Still needs CI fixes)

ickshonpe (changes requested):
I’m sorry I know you want to get this merged and you’ve done a huge amount of work on it, I can’t help but keep feeling it’s not quite the right direction and there are better solutions. Just ignore me if this seems like nonsense but this is how I currently think things should work instead:

Have the top-level component be `Text`:
```rust
#[derive(Component, Debug, Default, Clone, Deref, DerefMut, Reflect)]
#[require(
    TextStyle,
    TextNodeFlags,
    Node,
    ContentSize, 
    FocusPolicy, 
    ZIndex, 
    Visibility, 
    Transform
)]
pub struct Text(pub String);
````

Remove the `TextBlock` and `TextSpan` components.

Instead of `TextBlock` add `TextJustification` and `LineBreak` to `Style`, then layout child text nodes according to their parent's `TextJustifcation` and `LineBreak` properties.

All text entities are `Text` nodes, no special casing of the head text node.  Blocks of text are implicitly created from any adjacent sibling text nodes. Each `Text` node in the block of text has an identical copy of the same `Node` component so they overlap. 
`Text` nodes don't have a `Style` component and aren't allowed children.

So an example using this api to spawn a column of text:: 
```rust
    commands.spawn(NodeBundle {
        style: Style {
            width: Val::Px(500.),
            ..default()
        },
        text_justify: JustifyText::Center,
        ..default()
    }).with_children(|commands| {
        commands.spawn(Text::new("first line is "));
        commands.spawn((Text::new("green"), TextStyle { color: GREEN.into(), ..default() }));
        commands.spawn(NodeBundle::default());
        commands.spawn(Text::new("second line is "));
        commands.spawn((Text::new("red"), TextStyle { color: RED.into(), ..default() }));
        commands.spawn(NodeBundle::default());
        commands.spawn(Text::new("third line is "));
        commands.spawn((Text::new("blue"), TextStyle { color: BLUE.into(), ..default() }));  
    });
```

(you could use newlines here but I’m using ui nodes to show how you could break up text into separate blocks with the api.)

UkoeHB:
> Instead of TextBlock add TextJustification and LineBreak to Style, then layout child text nodes according to their parent's TextJustifcation and LineBreak properties.

> Blocks of text are implicitly created from any adjacent sibling text nodes. Each Text node in the block of text has an identical copy of the same Node component so they overlap.

I'd prefer to keep the current design, which is much more explicit and does not further increase the complexity of `Style`.

PR #15597, opened 2024-10-02
Remove `Interaction`, and replace it with `bevy_picking` based fields on `Button`

### LINKED ISSUE #1239: [bevy_ui][focus] distinguish mouse buttons
**What problem does this solve or what need does it fill?**
`ui_focus_system` system currently handles only left button clicks/touches. It would be nice to be able to handle right button clicks too.

**What solution would you like?**
Well, probably `Interaction::Clicked` could be just replaced with `Interaction::Clicked(MouseButton)` or with `Interaction::Clicked(Option<MouseButton>)` to handle touches with `Option::None` 🤷🏻 .

### LINKED ISSUE #2322: Right click button interactions

Allowing right clicking on buttons would allow for unique behaviors on buttons depending on the type of click.

Instead of Clicked being a type of an enum, clicked itself uses the MouseButton enum so that a click can be left, right, middle, or otehr

Having a RightClicked type, though it feels like code bloat since MouseButton exists.

## Additional context

Any other information you would like to add such as related previous work,
screenshots, benchmarks, etc.

### LINKED ISSUE #5769: Add support for detecting release of a button

Currently we detect when a button is being hovered, or clicked, but not _released_.
```
pub enum Interaction {
    /// The node has been clicked
    Clicked,
    /// The node has been hovered over
    Hovered,
    /// Nothing has happened
    #[default]
    None,
}
```
In this case `Clicked` is true when the user presses down on the button.

Add a `Released` variant to the `Interaction` enum, then add whatever logic necessary to correctly detect it.

- Not supporting it is not really an option. Most applications (and games) only trigger the button's action on release.

### LINKED ISSUE #15550: Remove `bevy_ui::Interaction` in favor of a `bevy_picking` based solution
`bevy_ui`'s `Interaction` component is clunky and crude. Now that we have proper picking support, we should replace it completely in some form. See #8157, #10141, #9240, #7257, #420, #5769, #1239 and #2322 for prior related discussions.

I'm aiming to tackle this over the next week as my primary task in the lead-up to the Bevy 0.15 release candidate :)

alice-i-cecile:
# Objective

Fixes #15550. See that issue and the linked issues within it about why the existing `Interaction` component is inadequate.

## Solution

- [x] remove `Interaction` and all engine systems that use it
- [x] add `pressed` and `hovered` fields to `Button`
- [x] updated these fields using observers powered by `bevy_picking`
- [x] migrate all existing examples
- [x] figure out what's going wrong with hovering in the `button` example
   - pointer presses are getting propagated correctly, but pointer hovering seems to be blocked by Text children
- [x] ~~make Text entities not pickable by default~~
- [x] swap to HoverMap based detection for hovering to avoid blocking by children
- [x] remove `FocusPolicy` in favor of `Pickable`
- [x] remove `RelativeCursorPosition`. Widgets should compute this information themselves based on `bevy_picking` input.
- [ ] fix slider widgets in the `animation_graph` example
- [ ] consider how to improve the migration path for the existing dynamic-styling pattern

Note to reviewers: the previous solution has special-cased logic for hidden nodes to ensure that they can't be pressed. I would prefer not to implement that in this PR, as that's fundamentally a picking backend concern.

## Testing

For basic functionality, run the `button` example.

For something more holistic, try the `game_menu` example.

## Migration Guide

The `Interaction` component has been removed. Instead, use the new `pressed` and `hovered` fields on `Button`. Bear in mind that you can also use observers to listen for `bevy_picking` events such as `Pointer<Down>` and `Pointer<Over>` directly as well.

To migrate code that matches on `Interaction` to change some property, follow the example below:

Bevy 0.14:

```rust
outline.color = match *button {
    Interaction::Pressed => RED.into(),
    Interaction::Hovered => WHITE.into(),
    Interaction::None => Color::NONE,
};
```

Bevy 0.15:

```rust
outline.color = match (button.pressed, button.hovered) {
    (true, _) => RED.into(),
    (false, true) => WHITE.into(),
    (false, false) => Color::NONE,
};
```

`NodeBundle` (and their Bevy 0.15 equivalents) no longer track whether or not they are pressed / hovered by default. If you require this functionality, add the `Button` component to your entity.

The `FocusPolicy` component has been removed, and replaced with the universal `Pickable` component.

The `RelativeCursorPosition` component has been removed. If you need this information, you should compute it from `bevy_picking` events in your widget's logic. See the `animation_graph` example for a demonstration of how this can be done.


Leinnan:
From game maker perspective- how to most easily get the "is this button just released" info?

alice-i-cecile:
> From game maker perspective- how to most easily get the "is this button just released" info?

I think that the best strategy is to use an observer listening to `Pointer<Up>` event, and check if it matches the entities you care about. Can you say more about why you want this information? Things like dragging should use the specialized bevy_picking APIs for that purpose.

crates/bevy_ui/src/widget/button.rs
```diff
@@ -1,7 +1,58 @@
...
+    }
+}
+
+/// An observer that watches for [`Pointer<Out>`] events and sets the [`Button::hovered`] field to `false`.
+///
+/// [`Button::pressed`] is also set to `false`, to ensure that the button.
+pub fn button_out_observer(trigger: Trigger<Pointer<Out>>, mut button_query: Query<&mut Button>) {
+    if let Ok(mut button) = button_query.get_mut(trigger.entity()) {
+        button.hovered = false;
+        button.pressed = false;
```
MiniaczQ, line 42:
```suggestion
```
Apps can check for `hovered && pressed` but in web if you press and drag away off the button you will still see the "pressed down" effect.
It's also simpler, so easier to work with

MiniaczQ: But this will require `Pointer<Down>` to affect this even if it's unrelated to this entity.

MiniaczQ (commented):
I like the idea in general,
not super sure about the separation (button has pressed & hovered) / naming (button) here.
There can be label widgets that get highlighted when hovered over, does that make them a button?
In that case, should hovered be a separate component or should button have a better name

crates/bevy_ui/src/widget/button.rs
alice-i-cecile: Yeah, currently I'm encouraging a "handle button events as soon as buttons are pressed" design, which dodges this complexity IMO.

alice-i-cecile:
> There can be label widgets that get highlighted when hovered over, does that make them a button?

The idea from @NthTensor is that each widget type should track its own interaction state independently, as its rules may differ. I agree with this design. I would have showcased it here, but uh, we only have one widget type 😅

crates/bevy_ui/src/widget/button.rs
MiniaczQ: Sure, if we go with Nth's idea, people can pretty easily reimplement this behavior :)

MiniaczQ (approved):

NthTensor (approved):

crates/bevy_ui/src/node_bundles.rs
```diff
@@ -278,6 +271,10 @@ impl Default for ButtonBundle {
             inherited_visibility: Default::default(),
             view_visibility: Default::default(),
             z_index: Default::default(),
+            pickable: Pickable {
+                should_block_lower: true,
+                ..Default::default()
+            },
```
tim-blackbird, line 277:
```suggestion
            pickable: Pickable::default(),
```
`should_block_lower` is `true` by default

alice-i-cecile: Great; that means we can swap to a pure derive then.

NthTensor:
`Pickable::default()` is basically a no-op. We should discourage users from adding it imo, or make `Pickable` required by `Node`.

crates/bevy_ui/src/widget/button.rs
```diff
@@ -1,7 +1,94 @@
...
+        } else {
+            let mut hovering_over_descendant = false;
+
+            for descendant in children_query.iter_descendants(entity) {
+                if map.contains_key(&descendant) {
+                    hovering_over_descendant = true;
+                    break;
+                }
+            }
+            hovering_over_descendant
```
mnmaita, line 71: Isn't it possible to use `some()` with `DescendantsIter`?

crates/bevy_ui/src/focus.rs
```diff
@@ -1,350 +1,5 @@
...
-///
-/// The component is updated when it is in the same entity with [`Node`].
-#[derive(Component, Copy, Clone, Default, PartialEq, Debug, Reflect)]
-#[reflect(Component, Default, PartialEq, Debug)]
-#[cfg_attr(
-    feature = "serialize",
-    derive(serde::Serialize, serde::Deserialize),
-    reflect(Serialize, Deserialize)
-)]
-pub struct RelativeCursorPosition {
```
ickshonpe, line 85: We could keep the `RelativeCursorPosition` component for now and just update it in `update_hover_status` or something. It would make the migration simpler.

crates/bevy_ui/src/widget/button.rs
```diff
@@ -1,7 +1,94 @@
...
+    children_query: Query<&Children>,
+) {
+    let Some(map) = hover_map.get(&PointerId::Mouse) else {
+        // If no appropriate hover map data is found, we can assume that no buttons are hovered
+        for (_, mut button) in button_query.iter_mut() {
+            // Avoid triggering change detection spuriously
+            // We can't use `set_if_neq` here because we're only looking at a single component
+            if button.hovered {
+                button.hovered = false;
+            }
```
ickshonpe, line 51:
```suggestion
            button.set_if_neq(Button {
                pressed: button.pressed,
                hovered: false,
            });
```

ickshonpe: Wait, if no buttons are hovered  we don't need to preserve the `pressed` state either, so both fields should be set to false?

crates/bevy_ui/src/widget/button.rs
```diff
@@ -1,7 +1,94 @@
...
+                    hovering_over_descendant = true;
+                    break;
+                }
+            }
+            hovering_over_descendant
+        };
+
+        // Avoid triggering change detection spuriously
+        // We can't use `set_if_neq` here because we're only looking at a single component
+        if button.hovered != is_hovered {
```
ickshonpe, line 76:
```suggestion
             button.set_if_neq(Button {
                pressed: button.pressed,
                hovered: is_hovered,
            });
```

PR #15607, opened 2024-10-03
Track callsite for observers & hooks

### LINKED ISSUE #14708: Track callsites of where observers are initially triggered.

Similar idea to #14034. Via a feature flag, we should be able to track where observers are getting triggered from.

Storing the callsite location as part of the `Trigger`, so that it's accessible inside the triggered observer.

SpecificProtagonist:
# Objective

Fixes #14708

Also fixes some commands not updating tracked location.

## Solution

`ObserverTrigger` has a new `caller` field with the `track_change_detection` feature;
hooks take an additional caller parameter (which is `Some(…)` or `None` depending on the feature).

## Testing

See the new tests in `src/observer/mod.rs`

---

## Showcase

Observers now know from where they were triggered (if `track_change_detection` is enabled):
```rust
world.observe(move |trigger: Trigger<OnAdd, Foo>| {
    println!("Added Foo from {}", trigger.caller());
});
```

## Migration

- hooks now take an additional `Option<&'static Location>` argument


alice-i-cecile (commented):
I like this strategy, but I really want to see benchmarks before we remove the feature flag on `BundleInserter`.

NthTensor:
Personally, I'd rename `track_change_detection` and gate all tracking behind it.

For hooks, that's a bit harder. Ideally an optional `Option<&Location>` parameter, so that people don't have to litter their code with `cfg` attributes.

examples/ecs/component_hooks.rs
```diff
@@ -63,16 +63,22 @@ fn setup(world: &mut World) {
...
-        // A hook has 3 arguments:
+        // A hook has 4 arguments:
         // - a `DeferredWorld`, this allows access to resource and component data as well as `Commands`
         // - the entity that triggered the hook
         // - the component id of the triggering component, this is mostly used for dynamic components
+        // - the location of the code that caused the hook to trigger
         //
         // `on_add` will trigger when a component is inserted onto an entity without it
-        .on_add(|mut world, entity, component_id| {
+        .on_add(|mut world, entity, component_id, caller| {
```
bushrat011899, line 73: Follow-up PR: The signature for a hook is getting pretty unwieldy. The `component_id` was already an infrequently used parameter (IMO), and I imagine users ignoring `caller` (even if they shouldn't!). It might be worth refactoring `entity`, `component_id`, and `caller` into a `Context` struct. I've omitted `world` from that context since it's the only one out of these 4 that a user will mutate, so keeping it separate will avoid some partial-mutable-borow pains.

crates/bevy_ecs/src/bundle.rs
```diff
@@ -1058,12 +1058,16 @@ impl<'w> BundleInserter<'w> {
...
+                        #[cfg(feature = "track_location")]
+                        caller,
                     );
                 }
                 deferred_world.trigger_on_replace(
                     archetype,
                     entity,
                     archetype_after_insert.iter_existing(),
+                    #[cfg(feature = "track_location")]
+                    caller,
```
bushrat011899, line 1070: Follow-up PR: I think this pattern is pretty annoying to deal with. It's verbose and off-putting (IMO). I think we should investigate if obfuscating this behind a `Context` type which _internally_ contains this feature gate is easier to use without sacrificing performance.

As much as I'm a `no_std` advocate, another alternative could be to use thread-local storage to manage this information _without_ explicitly passing it as a function argument.

bushrat011899 (approved):
Excellent work. Trying to tie failures directly to _user_ actions rather than via engine internals in a stack trace is definitely worth pursuing. I do have a couple suggestions that should be tackled in a follow-up PR to address some of the ergonomics issues we're starting to collect with `track_caller`. But as stated, these are follow-up concerns.

NthTensor:
>  I do have a couple suggestions that should be tackled in a follow-up PR to address some of the ergonomics issues we're starting to collect with track_caller. But as stated, these are follow-up concerns.

I'll add these to https://github.com/bevyengine/bevy/issues/16775 when I get a sec.

chescock (approved):
Looks good!

PR #15611, opened 2024-10-03
Add "bevy_input_focus" crate.

viridia:
# Objective

Define a framework for handling keyboard focus and bubbled keyboard events, as discussed in #15374.

## Solution

Introduces a new crate, `bevy_input_focus`. This crate provides:

* A resource for tracking which entity has keyboard focus.
* Methods for getting and setting keyboard focus.
* Event definitions for triggering bubble-able keyboard input events to the focused entity.
* A system for dispatching keyboard input events to the focused entity.

This crate does *not* provide any integration with UI widgets, or provide functions for
tab navigation or gamepad-based focus navigation, as those are typically application-specific.

## Testing

Most of the code has been copied from a different project, one that has been well tested. However, most of what's in this module consists of type definitions, with relatively small amounts of executable code. That being said, I expect that there will be substantial bikeshedding on the design, and I would prefer to hold off writing tests until after things have settled.

I think that an example would be appropriate, however I'm waiting on a few other pending changes to Bevy before doing so. In particular, I can see a simple example with four buttons, with focus navigation between them, and which can be triggered by the keyboard.

@alice-i-cecile

### DIFF
--- a/crates/bevy_input_focus/src/lib.rs
+++ b/crates/bevy_input_focus/src/lib.rs
@@ -1,0 +1,131 @@
+#![cfg_attr(docsrs, feature(doc_auto_cfg))]
+#![forbid(unsafe_code)]
+#![doc(
+    html_logo_url = "https://bevyengine.org/assets/icon.png",
+    html_favicon_url = "https://bevyengine.org/assets/icon.png"
+)]
+
+//! Keyboard focus system for Bevy.
+//!
+//! This crate provides a system for managing keyboard focus in Bevy applications, including:
+//! * A resource for tracking which entity has keyboard focus.
+//! * Methods for getting and setting keyboard focus.
+//! * Event definitions for triggering bubble-able keyboard input events to the focused entity.
+//! * A system for dispatching keyboard input events to the focused entity.
+//!
+//! This crate does *not* provide any integration with UI widgets, or provide functions for
+//! tab navigation or gamepad-based focus navigation, as those are typically application-specific.
+
+use bevy_app::{App, Plugin, Update};
+use bevy_ecs::{
+    component::Component,
+    entity::Entity,
+    event::{Event, EventReader},
+    query::With,
+    system::{Commands, Query, Res, Resource},
+    world::{Command, DeferredWorld, World},
+};
+use bevy_hierarchy::Parent;
+use bevy_input::keyboard::KeyboardInput;
+use bevy_window::Window;
+
+/// Resource representing which entity has keyboard focus, if any.
+#[derive(Clone, Debug, Resource)]
+pub struct KeyboardFocus(pub Option<Entity>);
+
+/// Marker component for which entity should receive keyboard input if nothing is focused.
+#[derive(Clone, Debug, Component)]
+pub struct DefaultKeyHandler;
+
+/// Helper functions for [`World`] and [`DeferredWorld`] to set and clear keyboard focus.
+pub trait SetKeyboardFocus {
+    /// Set keyboard focus to the given entity.
+    fn set_keyboard_focus(&mut self, entity: Entity);
+    /// Clear keyboard focus.
+    fn clear_keyboard_focus(&mut self);
+}
+
+impl SetKeyboardFocus for World {
+    fn set_keyboard_focus(&mut self, entity: Entity) {
+        if let Some(mut focus) = self.get_resource_mut::<KeyboardFocus>() {
+            focus.0 = Some(entity);
+        }
+    }
+
+    fn clear_keyboard_focus(&mut self) {
+        if let Some(mut focus) = self.get_resource_mut::<KeyboardFocus>() {
+            focus.0 = None;
+        }
+    }
+}
+
+impl<'w> SetKeyboardFocus for DeferredWorld<'w> {
+    fn set_keyboard_focus(&mut self, entity: Entity) {
+        if let Some(mut focus) = self.get_resource_mut::<KeyboardFocus>() {
+            focus.0 = Some(entity);
+        }
+    }
+
+    fn clear_keyboard_focus(&mut self) {
+        if let Some(mut focus) = self.get_resource_mut::<KeyboardFocus>() {
+            focus.0 = None;
+        }
+    }
+}
+
+/// Command to set keyboard focus to the given entity.
+pub struct SetFocusCommand(Option<Entity>);
+
+impl Command for SetFocusCommand {
+    fn apply(self, world: &mut World) {
+        if let Some(mut focus) = world.get_resource_mut::<KeyboardFocus>() {
+            focus.0 = self.0;
+        }
+    }
+}
+
+/// A bubble-able event for keyboard input. This event is normally dispatched to the current
+/// keyboard focus entity, if any. If no entity has keyboard focus, then the event is dispatched to
+/// the main window.
+#[derive(Clone, Debug, Component)]
+pub struct FocusKeyboardInput(pub KeyboardInput);
+
+impl Event for FocusKeyboardInput {
+    type Traversal = &'static Parent;
+
+    const AUTO_PROPAGATE: bool = true;
+}
+
+/// Plugin which registers the system for dispatching keyboard events based on focus and
+/// hover state.
+pub struct InputDispatchPlugin;
+
+impl Plugin for InputDispatchPlugin {
+    fn build(&self, app: &mut App) {
+        app.insert_resource(KeyboardFocus(None))
+            .add_systems(Update, dispatch_keyboard_input);
+    }
+}
+
+fn dispatch_keyboard_input(
+    mut key_events: EventReader<KeyboardInput>,
+    focus: Res<KeyboardFocus>,
+    windows: Query<Entity, With<Window>>,
+    mut commands: Commands,
+) {
+    // If an element has keyboard focus, then dispatch the key event to that element.
+    if let Some(focus_elt) = focus.0 {
+        for ev in key_events.read() {
+            commands.trigger_targets(FocusKeyboardInput(ev.clone()), focus_elt);
+        }
+    } else {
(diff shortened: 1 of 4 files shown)


alice-i-cecile:
I really want this, and I like the fundamental design. At the start of the 0.16 cycle I'm going to come back to this, evaluate scope and then decide how to proceed. We may want a working group for all of the follow-up.

viridia:
One thing that has not been established is the relationship between this and bevy_a11y. I had originally assumed that this would depend on bevy_a11y, but it sounds like you were thinking that it would be the other way around.

alice-i-cecile:
Yeah, my expectation is that bevy_a11y is a consumer of the focus data produced here from user input, and so should depend on this crate.

tools/publish.sh
```diff
@@ -47,6 +47,7 @@ crates=(
     bevy_dylib
     bevy_color
     bevy_picking
+    bevy_input_focus
```
atlv24, line 50: this should not be at the end of the list, put it right after bevy_input imo

shanecelis:
I've got a [library waiting in the wings](https://github.com/shanecelis/bevy_asky) that needs keyboard focus before moving forward. I'll be happy to attempt integrating this as a means to exercise and review it.

viridia:
There are two immediate follow-ups that need to happen:
* Integration with bevy_a11y
* @NthTensor's proposed change to allow bubbled events to propagate to the window. This is important because it allows "global" key handlers to work regardless of whether focus is Some or None:
  * If focus is None, keyboard events are dispatched to the window directly
  * If focus is Some, keyboard events are first processed by the focus widget and its ancestors, and if none of them call propagate(false) then the window gets the event.

NthTensor:
I'll have time to publish that work tomorrow.

viridia:
I've updated this so that if the current `KeyboardFocus` entity is `None`, then keyboard events are dispatched to the `PrimaryWindow` instead of all windows. If there's no primary window, then the event is dropped.

One open issue is where this crate belongs in the dependency order. Right now it's last in the list, but the plan is for bevy_a11y to depend on it, so it will likely need to come before that. (The actual dependency will be added in a follow-up PR.)

alice-i-cecile:
@cart and I have approved this approach in principle [here](https://discord.com/channels/691052431525675048/789987800014651412/1312184208285892649), so we should nail down the preliminary details via the standard review process and get this merged.

viridia:
I've updated the PR as follows:
* Rebased to latest Bevy main
* Re-ordered the new crate to come after both `bevy_input` and `bevy_a11y`, but not at the end of the list.
* Added `KeyboardFocusVisible` resource based on our discussion on Discord.
* Added trait extensions for `World` and `DeferredWorld` for testing whether an element, or its descendants, have keyboard focus.

NthTensor:
Does this still need bubble-to-window? I have an open PR if so.

viridia:
> Does this still need bubble-to-window? I have an open PR if so.

Yes, this is assuming that bubble-to-window is available.

viridia:
@NthTensor Consider the following scenario: Let's say we have a blender-like app which has a lot of hotkeys. Say the user presses 'q' for 'quit'. There are three cases:

* No widget has focus (focus = None) - the keyboard event gets sent directly to the primary window, where it is handled and processed through the global hotkey map.
* A button has focus - the keyboard event is dispatched to the button entity, but since button doesn't recognize 'q' as a shortcut, it bubbles all the way to the top, eventually reaching the window, where it has handled as in the previous case.
* A text input field has focus - in this case, the key is sent to the text input entity, which *does* recognize 'q' - so it calls `propagate(false)`, and inserts the letter 'q' into the text buffer. The keyboard even never reaches the window.

The code in this PR handles the 'no-focus' case by detecting focus = None and sending the event directly to the primary window. But I still need your PR to handle the second case, where the event is sent to the focus entity, but bubbles to the top of the hierarchy.

crates/bevy_input_focus/src/lib.rs
```diff
@@ -0,0 +1,243 @@
...
+#[derive(Clone, Debug, Resource)]
+pub struct KeyboardFocusVisible(pub bool);
+
+/// Helper functions for [`World`] and [`DeferredWorld`] to set and clear keyboard focus.
+pub trait SetKeyboardFocus {
+    /// Set keyboard focus to the given entity.
+    fn set_keyboard_focus(&mut self, entity: Entity);
+    /// Clear keyboard focus.
+    fn clear_keyboard_focus(&mut self);
+}
```
doup, line 50: Should these names be input device agnostic? The keyboards is one of the ways to change focus.

viridia: Unfortunately, I need *some* sort of adjective here: Bevy uses the word "focus" to mean a couple of different things, for example there is a "focus" concept in picking which has nothing to do with keyboard devices. So we need to be able to distinguish between a navigational focus, and other uses of the word focus.

Slight digression: Bevy's use of the word "focus" is obsolescent - that is, UI frameworks in the early era of personal computers did in fact have a broader definition of the term, but in modern UIs the word "focus" generally means keyboard focus or something like it.

doup:
I was thinking in local multiplayer with the impression that it might need something more complex. But I think this PR design is also OK for local multiplayer. These are game specific considerations:
- in the menus, the game can decide that the focus can be changed by any player (input device)… or it could be locked to "Player 1" (or whoever opened the pause menu)
- a character selection screen is purely game code

crates/bevy_input_focus/src/lib.rs
doup: Sorry, I might not be understanding this correctly.

I was questioning the `Keyboard` part. Should we drop the `Keyboard` part in these names and leave it as `InputFocus`/`DeviceInputFocus`? It's not only keyboard events we want to forward to the focused widgets, it's any kind of event from an input device, no?

[From the GH discussion](https://github.com/bevyengine/bevy/discussions/15374):

> Simple widgets such as buttons and checkboxes can use the same strategy - the only difference is that they respond to fewer keys. A button will respond to SPACE or ENTER, which is equivalent to a button press. All other events are ignored.

These widgets would also handle a button press to toggle the checkbox or trigger the button. Or, for example, an slider could handle directional input to increase/decrease a value.

Also, if we leave it as `InputFocus` from a "focus navigation" pov the name makes more sense?

viridia: I agree with you that the naming is an issue, and that using "keyboard" is too restrictive by implication. The problem is that other words are also problematic. The "picking" crate also uses the word "focus" (to mean something different), and as picking also involves "input" and "devices", then `InputFocus` creates a risk of ambiguity.

viridia:
@doup 

> I was thinking in local multiplayer with the impression that it might need something more complex. But I think this PR design is also OK for local multiplayer.

I deliberately decided to punt on the issue of multiple game controllers. First, I don't know if there are any games which use multiple controllers which also have the concept of "navigable focus". Second, even if there were such games, I'm not sure that they would conform to our expectations - that is, we could easily end up building something that didn't match the needs of those games. So it seems like the best solution is for games like that to build their own solutions.

crates/bevy_input_focus/src/lib.rs
alice-i-cecile: I would like to take the undecorated focus terms here, and rename the picking concept to `PickingFocus` and so on. I think that's much less confusing than any of the alternatives, and this really shouldn't be specific to keyboards.

doup: Alice expressed this as "focus cursor" in discord. Maybe we could use `cursor` instead: `InputCursor`? Or a mix of `cursor` + `focus`?

doup:
@viridia, imo the approach of this PR is correct. In https://github.com/bevyengine/bevy/pull/15611#issuecomment-2511020639 I meant that **multi-focus is NOT something Bevy should solve**.

I checked Super Mario Wonder which has local multiplayer:
- Player1 controls the "focus" in general menus
- While playing if any of the players pauses the game they have control of the "focus" on the pause menu.
- The only instance where multiple focus is used in the player select screen. This is game code, **not Bevy's responsibility**.

I don't have any other local multiplayer game, but I suspect is similar in any another game. Although, it would be nice if someone could check this in a fighting game or in Mario Kart. :-)

crates/bevy_input_focus/src/lib.rs
NthTensor: Cursor is pretty tied to mice in my mind and sort of ambiguous with pointer. I'd rather we centralize around a specific meaning for focus, and remove the old use in picking. Everything I picking is more about hovering than focus.

doup (approved):

alice-i-cecile:
@viridia ping me when CI is green and I'll give this a final review before merging.

viridia:
@mockersf One of the reasons why the CI is failing is because I wrote the dependencies as "0.15.0", however Bevy mainline in still has its version numbers as "0.15.0-dev".

NthTensor:
I think main always uses the -dev suffix.

viridia:
@alice-i-cecile I fixed the problem with the deps. I don't know what the other CI failures are, or whether they have anything to do with this PR.

NthTensor:
All prs that touch the `Cargo.toml` are currently failing on that check-ban. It's flagging a newly unmaintained crate deep in our dependency tree (which incidentally happens to owned by a colleague). Don't think we can fix it unless the crate we depend on updates. Maybe there's a way to ignore it?

alice-i-cecile (approved):
I like the core direction here and the model is fine. I think we should expand this to be more than keyboards, but that's easily done in follow-up.  Merging.

alice-i-cecile:
Thank you to everyone involved with the authoring or reviewing of this PR! This work is relatively important and needs release notes! Head over to https://github.com/bevyengine/bevy-website/issues/1965 if you'd like to help out.

Issue #14717, opened 2024-08-12
Footgun when chaining observer registrations on World vs App


ItsDoot:
## Bevy version

`0.14`

## What you did

I wrote a bunch of observers and chain added them to a plugin, which worked just fine:

```rust
app.observe(foo)
    .observe(bar)
    .observe(baz)
```

Then I went to write some tests, and not thinking anything of it, chain-added observers on `World`:

```rust
world.observe(foo)
    .observe(bar)
    .observe(baz)
```

## What went wrong

Unfortunately what that actually does is spawn an observer on the previously spawned observer, which meant my `bar` and `baz` observers would not end up running.

PR #15616, opened 2024-10-03
Rename observe to observe_entity on EntityWorldMut

kristoff3r:
# Objective

The current observers have some unfortunate footguns where you can end up confused about what is actually being observed. For apps you can chain observe like `app.observe(..).observe(..)` which works like you would expect, but if you try the same with world the first `observe()` will return the `EntityWorldMut` for the created observer, and the second `observe()` will only observe on the observer entity. It took several hours for multiple people on discord to figure this out, which is not a great experience.

## Solution

Rename `observe` on entities to `observe_entity`. It's slightly more verbose when you know you have an entity, but it feels right to me that observers for specific things have more specific naming, and it prevents this issue completely.

Another possible solution would be to unify `observe` on `App` and `World` to have the same kind of return type, but I'm not sure exactly what that would look like.

## Testing

Simple name change, so only concern is docs really.

---

## Migration Guide

The `observe()` method on entities has been renamed to `observe_entity()` to prevent confusion about what is being observed in some cases.

### DIFF
--- a/benches/benches/bevy_ecs/observers/propagation.rs
+++ b/benches/benches/bevy_ecs/observers/propagation.rs
@@ -1,9 +1,5 @@
 use bevy_ecs::{
-    component::Component,
-    entity::Entity,
-    event::Event,
-    observer::Trigger,
-    world::World,
+    component::Component, entity::Entity, event::Event, observer::Trigger, world::World,
 };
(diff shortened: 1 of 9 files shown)


ItsDoot (approved):

ItsDoot:
This closes #14717 FYI

PR #15619, opened 2024-10-03
update stress tests

ickshonpe:
updates `many_buttons`, `many_glyphs` and `text_pipeline` to use the new text api


ickshonpe:
Oops, meant to be to UkoeHB's fork

PR #15622, opened 2024-10-03
Display the bounds of each text node in the `text_debug` ui example

ickshonpe:
# Objective

Add a background colour to each text node in the `text_debug` example to visualize their bounds.

## Showcase

deb

In the bottom right you can see the empty space at the bottom of the text node, making it much more obvious that there is a bug causing the size of the bounds to be calculated incorrectly.


kristoff3r (approved):

LiamGallagher737 (approved):

UkoeHB:
> bug causing the size of the bounds to be calculated incorrectly.

Looks like bounds are calculated using the `cosmic-text` `Metrics`, which is set using the largest font size in a text block.

PR #15624, opened 2024-10-03
Support using FilteredResources with ReflectResource.

chescock:
# Objective

Support accessing resources using reflection when using `FilteredResources` in a dynamic system.  This is similar to how components can be queried using reflection when using `FilteredEntityRef|Mut`.  

## Solution

Change `ReflectResource` from taking `&World` and `&mut World` to taking `impl Into<FilteredResources>` and `impl Into<FilteredResourcesMut>`, similar to how `ReflectComponent` takes `impl Into<FilteredEntityRef>` and `impl Into<FilteredEntityMut>`.  There are `From` impls that ensure code passing `&World` and `&mut World` continues to work as before.  

## Migration Guide

If you are manually creating a `ReflectComponentFns` struct, the `reflect` function now takes `FilteredResources` instead `&World`, and there is a new `reflect_mut` function that takes `FilteredResourcesMut`.


BenjaminBrienen (approved):

crates/bevy_ecs/src/reflect/resource.rs
```diff
@@ -52,7 +52,9 @@ pub struct ReflectResourceFns {
     /// Function pointer implementing [`ReflectResource::remove()`].
     pub remove: fn(&mut World),
     /// Function pointer implementing [`ReflectResource::reflect()`].
-    pub reflect: fn(&World) -> Option<&dyn Reflect>,
+    pub reflect: for<'w> fn(FilteredResources<'w, '_>) -> Option<&'w dyn Reflect>,
+    /// Function pointer implementing [`ReflectResource::reflect_mut()`].
```
BenjaminBrienen, line 56: It would be cool if this could be a static assertion

BenjaminBrienen (commented):
Looks good to me, but I'm not super confident in my ability to judge this code.

crates/bevy_ecs/src/reflect/resource.rs
chescock: I'm not quite sure what you want asserted?  I was trying to copy the style of the existing comments, and I think that's meant to be read as "`ReflectResource::reflect_mut()` invokes this function pointer".

crates/bevy_ecs/src/reflect/resource.rs
```diff
@@ -111,14 +113,19 @@ impl ReflectResource {
     }
 
     /// Gets the value of this [`Resource`] type from the world as a reflected reference.
-    pub fn reflect<'a>(&self, world: &'a World) -> Option<&'a dyn Reflect> {
-        (self.0.reflect)(world)
+    pub fn reflect<'w, 's>(
+        &self,
+        world: impl Into<FilteredResources<'w, 's>>,
```
cBournhonesque, line 118: nit: I don't know if this param should be renamed. The component reflection functions use `entity` to represent the `Into<FilteredEntityRef>`, maybe we should use `resource` or `resources`?

cBournhonesque (approved):
Looks great!

crates/bevy_ecs/src/reflect/resource.rs
chescock: I was waffling on this because I think most users will pass `&World` and I wanted to document that doing so was possible... but I eventually realized I should actually *document* that doing so was possible.  So I renamed the parameter to `resources` and also added a note to the documentation that `&World` is valid.

Issue #15609, opened 2024-10-03
Add more tools to conveniently work with hierarchies, especially in the presence of ghost nodes


alice-i-cecile:
## What problem does this solve or what need does it fill?

Following the introduction of https://github.com/bevyengine/bevy/pull/15341, some users have expressed concerns around difficulties migrating to account for ghost nodes in their hierarchies.

This is a great opportunity to add more tools for traversing hierarchies in the spirit of [`iter_descendants`](https://docs.rs/bevy/latest/bevy/hierarchy/trait.HierarchyQueryExt.html#tymethod.iter_descendants).

## What solution would you like?

There are two stages to this:

1. Add more methods to `HierarchyQueryExt` to capture other common tasks.
2. Add a ghost-node aware version of this trait, which lives in `bevy_ui` and takes a `Has<GhostNode>` argument as well. These methods should be named `iter_node_descendants` and so on.

In theory these can be done in either order, or in a single PR. Your choice!

In terms of additional tasks to cover:

1. `root_parent`: the entity that is highest up the hierarchy tree, if any.
2. `leaf_children`: an iterator over the children that have no children of their own.
3. `siblings`: other entities that are in direct children of the same parent.
4. `parent`: a convenience method that looks for the parent (much more useful when working with ghost nodes).
5. `children`: as above, but for children.

## What alternative(s) have you considered?

One day (🥺), this will get rolled into #3742. It's nice to ship helpful APIs to users sooner though, and these methods will be easier to migrate since they have semantic meaning.

alice-i-cecile:
The [`UiChildren`](https://dev-docs.bevyengine.org/bevy/ui/struct.UiChildren.html) `SystemParam` tackles a good chunk of task 2. I'm going to start on task 1, get that merged, and then beef up that API to match it.

villor:
With the first half of this issue is merged (#15627), we can focus on the second half: UI tree.

### The second half
For the second half of this issue, we would want to make `UIChildren` have feature parity with `HierarchyQueryExt`.

I propose we also change name of `UIChildren` to `UiTree` to better convey what it is. This could also serve as a good place to document the rules of the tree (e.g. has to be all Node/GhostNode up to the root, non-UI entities beneath the tree are allowed/ignored)

I looked over the APIs and summarized it the current state along with proposed `UiTree` APIs:
| `HierarchyQueryExt` (current)  | `UiChildren` (current) | `UiTree` (proposal) | Notes |
[...]

PR #15627, opened 2024-10-03
Add more tools for traversing hierarchies

alice-i-cecile:
# Objective

- Working with hierarchies in Bevy is far too tedious due to a lack of helper functions.
- This is the first half of #15609. 

## Solution

Extend [`HierarchyQueryExt`](https://docs.rs/bevy/latest/bevy/hierarchy/trait.HierarchyQueryExt) with the following methods:

- `parent`
- `children`
- `root_parent`
- `iter_leaves`
- `iter_siblings`
- `iter_descendants_depth_first`

I've opted to make both `iter_leaves` and `iter_siblings` collect the list of matching Entities for now, rather that operate by reference like the existing `iter_descendants`. This was simpler, and in the case of `iter_siblings` especially, the number of matching entities is likely to be much smaller.

I've kept the generics in the type signature however, so we can go back and optimize that freely without a breaking change whenever we want.

## Testing

I've added some basic testing, but they're currently failing. If you'd like to help, I'd welcome suggestions or a PR to my PR over the weekend <3

### DIFF
--- a/crates/bevy_hierarchy/src/query_extension.rs
+++ b/crates/bevy_hierarchy/src/query_extension.rs
@@ -10,11 +10,46 @@ use crate::{Children, Parent};
 
 /// An extension trait for [`Query`] that adds hierarchy related methods.
 pub trait HierarchyQueryExt<'w, 's, D: QueryData, F: QueryFilter> {
+    /// Returns the parent [`Entity`] of the given `entity`, if any.
+    fn parent(&'w self, entity: Entity) -> Option<Entity>
+    where
+        D::ReadOnly: WorldQuery<Item<'w> = &'w Parent>;
+
+    /// Returns a slice over the [`Children`] of the given `entity`.
+    ///
+    /// This may be empty if the `entity` has no children.
+    fn children(&'w self, entity: Entity) -> impl Iterator<Item = Entity> + 'w
+    where
+        D::ReadOnly: WorldQuery<Item<'w> = &'w Children>;
+
+    /// Returns the topmost ancestor of the given `entity`.
+    ///
+    /// This may be the entity itself if it has no parent.
+    fn root_parent(&'w self, entity: Entity) -> Entity
+    where
+        D::ReadOnly: WorldQuery<Item<'w> = &'w Parent>;
+
+    /// Returns an [`Iterator`] of [`Entity`]s over the leaves of the hierarchy that are underneath this `entity``.
+    ///
+    /// Only entities which have no children are considered leaves.
+    /// This will not include the entity itself, and will not include any entities which are not descendants of the entity,
+    /// even if they are leaves in the same hierarchical tree.
+    fn iter_leaves(&'w self, entity: Entity) -> LeafIter<'w, 's, D, F>
+    where
+        D::ReadOnly: WorldQuery<Item<'w> = &'w Children>;
+
+    /// Returns an [`Iterator`] of [`Entity`]s over the `entity`s immediate siblings, who share them same parent.
+    ///
+    /// The entity itself is not included in the iterator.
+    fn iter_siblings(&'w self, entity: Entity) -> SiblingIter<'w, 's, D, F>
+    where
+        D::ReadOnly: WorldQuery<Item<'w> = (&'w Parent, &'w Children)>;
+
     /// Returns an [`Iterator`] of [`Entity`]s over all of `entity`s descendants.
     ///
     /// Can only be called on a [`Query`] of [`Children`] (i.e. `Query<&Children>`).
     ///
-    /// Traverses the hierarchy breadth-first.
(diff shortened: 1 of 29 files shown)


villor:
Did a [PR](https://github.com/alice-i-cecile/bevy/pull/169) with fixed siblings iter.

The problem was that the query has to be `Query<(Option<&Parent>, Option<&Children>)>` since we are using the same query for the parent and children.

I wonder if it would be a good idea to convert `HierarchyQueryExt` into system params instead 🤔 what is the reason it is an extension in the first place?

alice-i-cecile:
> I wonder if it would be a good idea to convert HierarchyQueryExt into system params instead 🤔 what is the reason it is an extension in the first place?

I'm pretty sure it predates the `SystemParam` derive macro, and has grown organically from there. I really like the filtering capabilities, but I agree: I think that the SystemParam approach is a lot simpler as both a user and a maintainer. Adding a filter generic should be totally doable. Follow-up though!

villor:
@alice-i-cecile 

Pushed some more commits, [PR here](https://github.com/alice-i-cecile/bevy/pull/170)

- Cleaned up my previous `flat_map`->`into_iter` chaos. Much more readable now!
- Made sure children() returns a slice like the doc comment says, and cleaned up the implementation.
- Renamed root_parent -> root_ancestor
- Removed `LeafIter` and instead return the traversal iterator directly (reducing allocations)
- Added a doc comment about the `iter_leaves` traversal order. (Would it make sense to make this one traverse depth-first instead for the "correct" order?)

alice-i-cecile:
@villor you're the best. I really appreciate your help with this!

bushrat011899 (approved):
Like the options this provides, and helps justify the existence of what is otherwise a pretty small crate! I do think this should be refactored into a `SystemParam`, but that's a larger change that can easily be built on top of this work in 0.16. I also see a lot of refactoring in the near future anyway with #15635, so no real need to do that here and now.

villor:
> Like the options this provides, and helps justify the existence of what is otherwise a pretty small crate! I do think this should be refactored into a `SystemParam`, but that's a larger change that can easily be built on top of this work in 0.16. I also see a lot of refactoring in the near future anyway with #15635, so no real need to do that here and now.

SystemParam is going to be the bomb! I imagine we could add other APIs as well like change detection on different parts of the hierarchy. A lot more options open up once we can control the queries more 😁

And when we have #15635 we can make it generic over any one-to-many relation!

ItsDoot (approved):

crates/bevy_hierarchy/src/query_extension.rs
```diff
@@ -5,16 +5,54 @@ use bevy_ecs::{
...
+    /// Only entities which have no children are considered leaves.
+    /// This will not include the entity itself, and will not include any entities which are not descendants of the entity,
+    /// even if they are leaves in the same hierarchical tree.
+    ///
+    /// Traverses the hierarchy depth-first.
+    fn iter_leaves(&'w self, entity: Entity) -> impl Iterator<Item = Entity> + 'w
+    where
+        D::ReadOnly: WorldQuery<Item<'w> = &'w Children>;
+
+    /// Returns an [`Iterator`] of [`Entity`]s over the `entity`s immediate siblings, who share them same parent.
```
atornity, line 44:
```suggestion
/// Returns an [`Iterator`] of [`Entity`]s over the `entity`s immediate siblings, who share the same parent.
```

atornity (approved):

crates/bevy_hierarchy/src/query_extension.rs
```diff
@@ -5,16 +5,54 @@ use bevy_ecs::{
...
+    /// Only entities which have no children are considered leaves.
+    /// This will not include the entity itself, and will not include any entities which are not descendants of the entity,
+    /// even if they are leaves in the same hierarchical tree.
+    ///
+    /// Traverses the hierarchy depth-first.
+    fn iter_leaves(&'w self, entity: Entity) -> impl Iterator<Item = Entity> + 'w
+    where
+        D::ReadOnly: WorldQuery<Item<'w> = &'w Children>;
+
+/// Returns an [`Iterator`] of [`Entity`]s over the `entity`s immediate siblings, who share the same parent.
```
ItsDoot, line 44:
```suggestion
    /// Returns an [`Iterator`] of [`Entity`]s over the `entity`s immediate siblings, who share the same parent.
```

atornity: whops

atlv24 (approved):

alice-i-cecile:
Thank you to everyone involved with the authoring or reviewing of this PR! This work is relatively important and needs release notes! Head over to https://github.com/bevyengine/bevy-website/issues/1718 if you'd like to help out.

PR #15589, opened 2024-10-02
Impose a more sensible ordering for animation graph evaluation.

pcwalton:
This is an updated version of #15530.

This commit changes the animation graph evaluation to be operate in a more sensible order and updates the semantics of blend nodes to conform to [the animation composition RFC]. Prior to this patch, a node graph like this:

	    ┌─────┐
	    │     │
	    │  1  │
	    │     │
	    └──┬──┘
	       │
       ┌───────┴───────┐
       │               │
       ▼               ▼
    ┌─────┐         ┌─────┐
    │     │         │     │
    │  2  │         │  3  │
    │     │         │     │
    └──┬──┘         └──┬──┘
       │               │
   ┌───┴───┐       ┌───┴───┐
   │       │       │       │
   ▼       ▼       ▼       ▼
┌─────┐ ┌─────┐ ┌─────┐ ┌─────┐
│     │ │     │ │     │ │     │
│  4  │ │  6  │ │  5  │ │  7  │
│     │ │     │ │     │ │     │
└─────┘ └─────┘ └─────┘ └─────┘

Would be evaluated as (((4 ⊕ 5) ⊕ 6) ⊕ 7), with the blend (lerp/slerp) operation notated as ⊕. As quaternion multiplication isn't commutative, this is very counterintuitive and will especially lead to trouble with the forthcoming additive blending feature (#15198).

This patch fixes the issue by changing the evaluation order to postorder, with children of a node evaluated in ascending order by node index.

To do so, this patch revamps `AnimationCurve` to be based on an *evaluation stack* and a *blend register*. During target evaluation, the graph evaluator traverses the graph in postorder. When encountering a clip node, the evaluator pushes the possibly-interpolated value onto the evaluation stack. When encountering a blend node, the evaluator pops values off the stack into the blend register, accumulating weights as appropriate. When the graph is completely evaluated, the top element on the stack is *committed* to the property of the component.

A new system, the *graph threading* system, is added in order to cache the sorted postorder traversal to avoid the overhead of sorting children at animation evaluation time. Mask evaluation has been moved to this system so that the graph only has to be traversed at most once per frame. Unlike the `ActiveAnimation` list, the *threaded graph* is cached from frame to frame and only has to be regenerated when the animation graph asset changes.

[...]

PR #15631, opened 2024-10-03
Implement additive blending for animation graphs.

### LINKED ISSUE #14395: Animation: Layered Blend Per Bone / Additive Blending

 I want to be able to have apply a normal 'running' animation to my characters whole body,  but then also be able to apply a bone-layer-masked animation to just the torso of the character on TOP of that running animation at will.  

I would like for the bevy animation system to be able to support 'Layered Blend Per Bone' , the ability to apply an animation to a particular set of bones (  by bone [node] name ).   Ideally, I  also want the blend to apply in such a way that my layered animation is FULLY expressed,  not averaged down by a division of the total weights of all animations.    This is because I want the sword slash on the torso to look complete and not weird like it does when trying to blend it in 0.14 . 

To be honest, an example with a humanoid who is running and sometimes doing a sword slash would be preferable to the fox example because i dont think the fox example is powerful enough to full express and capture what is needed.

pcwalton:
*Additive blending* is an ubiquitous feature in game engines that allows animations to be concatenated instead of blended. The canonical use case is to allow a character to hold a weapon while performing arbitrary poses. For example, if you had a character that needed to be able to walk or run while attacking with a weapon, the typical workflow is to have an additive blend node that combines walking and running animation clips with an animation clip of one of the limbs performing a weapon attack animation.

This commit adds support for additive blending to Bevy. It builds on top of the flexible infrastructure in #15589 and introduces a new type of node, the *add node*. Like blend nodes, add nodes combine the animations of their children according to their weights. Unlike blend nodes, however, add nodes don't normalize the weights to 1.0.

The `animation_masks` example has been overhauled to demonstrate the use of additive blending in combination with masks. There are now controls to choose an animation clip for every limb of the fox individually.

This patch also fixes a bug whereby masks were incorrectly accumulated with `insert()` during the graph threading phase, which could cause corruption of computed masks in some cases.

Note that the `clip` field has been replaced with an `AnimationNodeType` enum, which breaks `animgraph.ron` files. The `Fox.animgraph.ron` asset has been updated to the new format.

Closes #14395.

## Showcase

https://github.com/user-attachments/assets/52dfe05f-fdb3-477a-9462-ec150f93df33

## Migration Guide

* The `animgraph.ron` format has changed to accommodate the new *additive blending* feature. You'll need to change `clip` fields to instances of the new `AnimationNodeType` enum.


crates/bevy_animation/src/animation_curves.rs
```diff
@@ -844,11 +885,20 @@ where
...
             Some((ref mut current_value, ref mut current_weight)) => {
                 *current_weight += weight_to_blend;
-                *current_value = A::interpolate(
-                    current_value,
-                    &value_to_blend,
-                    weight_to_blend / *current_weight,
-                );
+
+                if additive {
+                    *current_value = A::blend(iter::once(BlendInput {
```
mweatherley, line 890: This currently just ignores the existing value in the blend register, which doesn't seem right. I think that for the present implementation of `Animatable::blend`, the right thing to do would be to use an iterator of length two with two blend inputs, one for the current value and one for the incoming one, with both inputs set to additive.

(Probably we could also just change the definition of `Animatable::blend` to support the blend stacks more directly in the future as well; I don't think we use it for anything other than what's happening here.)

mweatherley (changes requested):
Looks great overall. Just the one thing.

crates/bevy_animation/src/animation_curves.rs
pcwalton: Oops! I feel silly now :) Fixed.

mweatherley (approved):

mweatherley:
It's probably worth mentioning in the 0.15 migration guide that the changes to weight normalization in #15589 can be counteracted by shoving each `AnimationClip` behind its own additive blend node if the previous behavior is desired. Please correct me if I'm wrong :) @pcwalton

pcwalton:
I never really considered that, but I *think* that's correct?

mockersf (approved):

alice-i-cecile:
Thank you to everyone involved with the authoring or reviewing of this PR! This work is relatively important and needs release notes! Head over to https://github.com/bevyengine/bevy-website/issues/1715 if you'd like to help out.

Issue #3742, opened 2022-01-22
Entity-entity relations 🌈 


alice-i-cecile:
##  Overview

Inspired by [flecs](https://www.flecs.dev/flecs/md_docs_2Relationships.html), relations are a first-class graph primitive for the ECS: useful for modelling generalized hierarchies, graphs and so on in a safe, fast and ergonomic way by working *with* the ECS, rather than fighting it.

These have three main categories of benefits:
- **Ergonomics:** query for entities with specific relation types and iterate over the various relations
- **Correctness:** enforce critical logic invariants (such as "each child must have a parent" or "each child can only have at most one parent")
- **Performance:** operate more quickly, especially compared to naive hand-rolled implementations

## Challenges

There are several critical challenges that need to be addressed by any design:
1. **Performance:** ECS is designed for random-order linear iteration
2. **Serialization:** entity ids are opaque and ephemeral
3. **Ergonomics:** there's a lot of interesting and important things you may want to do
4. **Graph structure:** how do you ensure that the rules of the graph that you want to represent are enforced upon creation, and stay correct as components and entities are added and removed. Some graphs may be directed, some may be symmetric, some may have edge weights, some may be acyclic, some may be trees, some may be forests...
5. **Multiple edges from / to the same entity:** this is an essential use case, but entities can (currently) only have one of each component type.

## Possible paths forward

Essential universal tasks:
- [ ] Thoroughly benchmark existing tools and strategies for working with graphs of entities

Optional universal tasks:
- [ ] #1527
- [ ] Investigate further optimizations for `query.get()`
- [x] #3022

Index-driven approaches:
- [ ] Create one or more reasonably performant graph-storing indexes
- [ ] Implement a complete, ergonomic indexes solution (see #1205)

Broadly speaking, there are two paths to this design:

1. **Index-based:** Store the graph in a resource, keep that resource up to date. Use that resource to ensure constraints aren't broken, and enable fast look-up.
2. **Baked-in:** Reframe components as relations with no targets. Reshape the ECS to support relations as the first-class primitives.

In both cases, these implementations details would be hidden to the end user.
Ultimately, we want to build an API that broadly follows the one outlined in https://github.com/bevyengine/rfcs/pull/18 if at all possible.

### Atomic indexes

- [x] Make the schedule less of a nightmare using https://github.com/bevyengine/rfcs/pull/45
- [ ] Implement atomic groups ala https://github.com/bevyengine/rfcs/pull/46
[...]

PR #15635, opened 2024-10-03
`ComponentHook` based `Parent`/`Children` Management

### LINKED ISSUE #12235: Use component lifecycle hooks to make `Parent` + `Children` hierarchy code simpler and more robust

`bevy_hierarchy`'s internals are complex, performance-critical, and have panicking unsafe code in them.
The core reason for this is that we *cannot* let the graph become invalid: this leads to wasted work, subtle bugs and huge amounts of user frustration. And because we use `unsafe` code internally that relies on the assumption of validity for soundness, this can become a soundness problem.

They're painful to maintain, and nearly impossible to replicate externally for other hierarchy-like code that wants to uphold these same variants.

#10756 adds component lifecycle hooks: operations that take effect when components are added or removed.

We should replace the existing complex code in `bevy_hierarchy` with this idiom, being sure to benchmark and optimize performance.

We could wait until the fabled day that relations (#3742) finally gets shipped.

## Additional context

This has been widely anticipated as one of the key benefits of that PR: I'm recording it here to make sure we don't forget (and can coordinate work).

While the user-facing API is *also* pretty gnarly and inconsistent, I'd like to leave that for future work to make sure that the correctness-sensitive internals get the review attention it needs.

### LINKED ISSUE #15270: Move methods on `Children` that mutate to `EntityCommands`

All of our hierarchy management controls happen through `Command`s, *except* methods of sorting children for a given entity. These methods should be changed to `Command`s like the rest of the hierarchy management. This might be required when changing `Children` to a relation, anyways.

Replace the following methods with commands:
- [`Children::swap`](https://docs.rs/bevy/0.14.0/bevy/hierarchy/struct.Children.html#method.swap)
- [`Children::sort_by`](https://docs.rs/bevy/0.14.0/bevy/hierarchy/struct.Children.html#method.sort_by)
- [`Children::sort_by_key`](https://docs.rs/bevy/0.14.0/bevy/hierarchy/struct.Children.html#method.sort_by_key)
- [`Children::sort_by_cached_key`](https://docs.rs/bevy/0.14.0/bevy/hierarchy/struct.Children.html#method.sort_by_cached_key)
- [`Children::sort_unstable_by`](https://docs.rs/bevy/0.14.0/bevy/hierarchy/struct.Children.html#method.sort_unstable_by)
- [`Children::sort_unstable_by_key`](https://docs.rs/bevy/0.14.0/bevy/hierarchy/struct.Children.html#method.sort_unstable_by_key)

Ideally:
```rust
impl EntityCommands {
    pub fn sort_children_by(&mut self, compare: impl FnMut(Entity, Entity) -> Ordering) -> &mut Self;
    pub fn sort_children_by_key<K: Ord>(&mut self, compare: impl FnMut(Entity) -> K) -> &mut Self;
    // etc...
}
```

### LINKED ISSUE #15575: Removing `Parent` / `Children` from the world and then reinserting them can break hierarchies
# Problem

Within [the internals of the transform propagation systems](https://github.com/bevyengine/bevy/blob/0fe17b8b303c35902de3d75681f971cd8d66712f/crates/bevy_transform/src/systems.rs#L74), we rely on hierarchies being well-formed (via https://github.com/bevyengine/bevy/pull/4197) for soundness, and panic if they are not in [this line](https://github.com/bevyengine/bevy/blob/0fe17b8b303c35902de3d75681f971cd8d66712f/crates/bevy_transform/src/systems.rs#L70). This was added in https://github.com/bevyengine/bevy/pull/4775.

This is why we haven't added simple constructors or mutable access to `Parent` or `Children` (like `Clone` or `Copy`), like in #15574. If we can ever swap or set the value of `Parent` or `Children` directly, it becomes very easy to break the hierarchy. Unfortunately, this isn't the *only* way to get that problem.

Path 1: get mutable access to two different components via Query::get_many_mut or the like, and then `mem_swap` them.

Path 2: remove `Parent` / `Children` from an entity via a method on `World` that returns the removed value

## Potential solutions

1. Wait for #3742.
2. Implement #12235 in an airtight way.
3. Make `bevy_transform` fully safe and eat the performance hit and correctness problems if people break their hierarchy by mistake.
4. Add more, niche, high complexity features to absolutely forbid mutable access to some component types. Somehow.

I prefer 2, and would be opposed to both 3 and 4.

## Additional context

Because this is actually guarded against by an assert, I think that merging #15574's successor is actually fine (cc @atornity, sorry).

bushrat011899:
# Objective

- Partially adresses #15575
- Fixes #12235 
- Indirectly fixes #15270
- Contributes to #15869

## Solution

- Made `Parent` and `Child` specific types of a One-to-Many relationship (relationship type `Family`)

## Testing

- Ran CI locally.

## Benchmarks

I ran the `despawn_world_recursive` benchmarks on my laptop and saw a 25% - 35% performance **improvement** moving to a `ComponentHooks` based relationship system in this PR.

despawn_world_recursive/10000_entities
_despawn_world_recursive/10000_entities_

I also used `tracy` to check that `alien_cake_addict` (my go-to example of a "real" game in Bevy) performed well.

image
_`alien_cake_addict` `tracy` fame-time plot_

While `alien_cake_addict` doesn't use the hierarchy extensively and explicitly (I think just for the score board at the end), I still think it's important to include.

---

## Showcase

Users can now directly insert `Parent` onto child entities without the need for the child-builder methods.

```rust
let parent = world.spawn_empty();
let child = world.spawn(Parent::new(parent));

// Automatically called at every sync point
world.flush();

assert!(world.get::<Children>(parent).is_some());
assert!(world.get::<Parent>(child).is_some());

world.entity_mut(parent).remove::<Children>();

world.flush();

assert!(world.get::<Children>(parent).is_none());
assert!(world.get::<Parent>(child).is_none());
```

Recursive despawn is now also offered in a generalised form operating on any `Component` which implements `VisitEntities`.

```rust
let bus = world.spawn_empty();
let passenger = world.spawn(Passenger::new(bus));

commands.despawn_recursive_with_options::<Passengers>(bus, true);
```

Note that currently this recursive despawn can only traverse a single relationship type.

### `HierarchyEvent`

The `HierarchyEvent` type has been replaced. The following mapping from old to new is required:
  - `ChildAdded` -> `Added`
  - `ChildRemoved` -> `Removed`
  - `ChildMoved` -> `Removed` and then `Added` in succession

### `DespawnRecursive` / `DespawnChildrenRecursive` / `despawn_with_children_recursive`

- `DespawnChildrenRecursive` has been removed and replaced with a parameter, `inclusive`, on the existing `DespawnRecursive`. The old `DespawnChildrenRecursive` is identical in functionality to the new `DespawnRecursive` with the parameter `inclusive: true`.
- The function `despawn_with_children_recursive` has been removed to reduce the public API surface area. Instead, construct a `DespawnRecursive` command and use `apply(...)` from the `Command` trait to immediately invoke it on a `World`.
- `DespawnRecursive`'s fields are now private. Use `new` and the appropriate `with_x` methods to change parameter defaults.
- `DespawnRecursive` now has a type parameter, `C`, which controls which relationship to follow recursively. Use `Children` to match the previous behaviour.

## Notes

In previous iterations of this PR, the possibility of generalised relationships was included, including one-to-one and many-to-many. These were removed during the review process based on feedback from subject matter experts. They may be re-added once MVP relationships lands at a TBD date. See #3742 for further details.


crates/bevy_hierarchy/src/hierarchy.rs
```diff
@@ -275,9 +271,9 @@ mod tests {
...
 
         {
-            let children = world.get::<Children>(grandparent_entity).unwrap();
+            let children = world.get::<Children>(grandparent_entity);
             assert!(
-                !children.iter().any(|&i| i == parent_entity),
+                children.is_none(),
                 "grandparent should no longer know about its child which has been removed"
             );
         }
```
bushrat011899, line 279: I want to explicitly call out that I believe this test was _wrong_ before this PR, since it left an empty `Children` component on the `grandparent` entity.

crates/bevy_hierarchy/src/one_to_one/component.rs
```diff
@@ -0,0 +1,188 @@
...
+        Component,
+        MapEntities,
+        VisitEntities,
+        VisitEntitiesMut,
+        PartialEq,
+        Debug,
+        FromWorld
+    )
+)]
+pub struct OneToOne<R> {
```
bushrat011899, line 41: This is a totally new feature and not strictly required to update `Parent` and `Children`. _However_, I found this to be a much simpler component to design, and I think it helps explain how the One-to-Many components work, while providing a new tool for users.

crates/bevy_hierarchy/src/one_to_one/event.rs
```diff
@@ -0,0 +1,46 @@
+use core::marker::PhantomData;
+
+use bevy_ecs::{entity::Entity, event::Event};
+
+/// An [`Event`] that is fired whenever there is a change in [`OneToOne`] relationship `R`.
+///
+/// [`Event`]: bevy_ecs::event::Event
+#[derive(Event)]
+pub enum OneToOneEvent<R> {
```
bushrat011899, line 9: I'm not very happy with way I've defined these `Event` types, since they expose the `PhantomData` field. I'm not sure of an alternative though and would love some feedback!

ItsDoot (commented):
Quick passover. Also, this PR indirectly addresses #15270

crates/bevy_hierarchy/src/hierarchy.rs
```diff
@@ -26,21 +26,14 @@ pub struct DespawnChildrenRecursive {
...
     despawn_with_children_recursive_inner(world, entity, warn);
 }
 
-// Should only be called by `despawn_with_children_recursive` and `try_despawn_with_children_recursive`!
+// Should only be called by `despawn_with_children_recursive`!
 fn despawn_with_children_recursive_inner(world: &mut World, entity: Entity, warn: bool) {
-    if let Some(mut children) = world.get_mut::<Children>(entity) {
-        for e in core::mem::take(&mut children.0) {
+    if let Some(children) = world.get::<Children>(entity) {
+        let children = children.as_slice().to_vec();
```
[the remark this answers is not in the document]
bushrat011899: I don't _think_ so. Only reason I did it this way was I hadn't exposed (either to the `crate` or `pub`) a method for taking out the inner `SmallVec`. I think I'll add a method to do that so we can avoid the allocation during despawn.

crates/bevy_hierarchy/src/one_to_many/many.rs
```diff
@@ -0,0 +1,218 @@
...
+                        moved.send(OneToManyEvent::<R>::Added(a_id, b_id, PhantomData));
+                    }
+                }
+            }
+        });
+    }
+
+    fn disassociate(mut world: DeferredWorld<'_>, a_id: Entity, _component: ComponentId) {
+        let Some(a_relationship) = world.get::<Self>(a_id) else {
+            unreachable!("component hook should only be called when component is available");
```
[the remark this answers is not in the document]
bushrat011899: Sadly not! `DebugCheckedUnwrap` is `pub(crate)` to `bevy_ecs`. But I do think it'd make sense to move that trait into `bevy_utils` to be honest.

crates/bevy_hierarchy/src/one_to_one/event.rs
eidloi: Maybe you could use a struct instead:
```rust
pub enum RelationEventType{
  Added,
  Removed,
}

pub struct OneToOneEvent<R>{
  phantom_data: PhantomData<fn(R)>,
  pub left: Entity,
  pub right: Entity,
  pub event_type: RelationEventType
}
```
Though I would probably not `pub` any field just provide constructors / accessors. They are the same size (24).

crates/bevy_hierarchy/src/one_to_many/many.rs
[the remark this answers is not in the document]
ItsDoot: We can leave this as-is then, and switch it when we pull hierarchy back into ecs.

crates/bevy_hierarchy/src/one_to_one/event.rs
bushrat011899: Yeah I agree with this. I was hoping to avoid a struct because it makes pattern matching more annoying, but it's the better option here.

andho:
> * Added two generalised relationship types, One-to-One, and One-to-Many, represented by the components `OneToOne`, and `OneToManyOne`/`OneToManyMany`

What's the reason behind calling it `OneToManyOne` and `OneToManyMany` instead of `OneToMany` and `ManyToOne`?

bushrat011899:
> What's the reason behind calling it `OneToManyOne` and `OneToManyMany` instead of `OneToMany` and `ManyToOne`?

Primary reason was I started with `OneToOne` went to create `OneToMany`, and then realised I needed two unique component types for that particular relationship. Honestly pretty reasonable name suggestion, thanks!

andho:
> Primary reason was I started with `OneToOne` went to create `OneToMany`, and then realised I needed two unique component types for that particular relationship. Honestly pretty reasonable name suggestion, thanks!

Actually...apologies if you've already changed it but, I'm coming around to `OneToManyOne` and `OneToManyMany`. The issue is `OneToMany` and `ManyToOne` works well if the RHS and LHS is clear, but in this case individually looking at the code, I can't figure out (referring to the example), if `Targetting` is many or `Targeted` is many (although it can be inferred from the context).

bushrat011899:
> Actually...apologies if you've already changed it but, I'm coming around to `OneToManyOne` and `OneToManyMany`. The issue is `OneToMany` and `ManyToOne` works well if the RHS and LHS is clear, but in this case individually looking at the code, I can't figure out (referring to the example), if `Targetting` is many or `Targeted` is many (although it can be inferred from the context).

I do agree it can be a little confusing, but I think that's fairly common for this kind of relational management. Speaking from personal experience (C# with Entity Framework), you have builder methods like `.WithOne()` and `.WithMany()` that you chain together to create these kinds of relationships. I still find myself getting confused with those methods from time to time even after several years.

I would say the solution to this problem is to encourage users to use `type` aliases (e.g., `Parent`, `Children`, `TargetedBy`, `Targeting`, etc.) to provide more meaningful names to these relationships. Like you experienced earlier, the name `OneToManyMany` is pretty off-putting, even if it _might_ make more sense _eventually_.

crates/bevy_hierarchy/src/one_to_many/many.rs
```diff
@@ -0,0 +1,308 @@
...
+use super::{ManyToOne, OneToManyEvent};
+
+/// Represents one half of a one-to-many relationship between an [`Entity`] and some number of other [entities](Entity).
+///
+/// The type of relationship is denoted by the parameter `R`.
+#[derive(Component)]
+#[component(
+    on_insert = Self::associate,
+    on_replace = Self::disassociate,
+    on_remove = Self::disassociate
```
nakedible, line 24: I may be absolutely, totally, wrong here as this is the very first time I find out that there even is an `on_replace` hook, and I've never even used hooks before, but...

Documentation says `on_replace` is also run when the component gets removed from the object, so why do we need `on_remove` as well?

I think it works because calling disassociate twice is safe. And I didn't add comments on all of the `on_remove` spots so all of them need to be changed if this is.

bushrat011899: Correct! I've explicitly made these methods safe for being run multiple times to account for this. When I originally wrote this I remembered finding an instance where `on_replace` was _not_ called after `on_remove`, but I'll double-check that now.

nakedible: If you do find it, it would probably be a good idea to add a test case which fails if `on_remove` is not given. Then if it is fixed, it's easy to see if the test case passes too.

crates/bevy_hierarchy/src/one_to_one/component.rs
```diff
@@ -0,0 +1,189 @@
...
+    entity::{Entity, VisitEntities, VisitEntitiesMut},
+    event::Events,
+    traversal::Traversal,
+    world::{DeferredWorld, FromWorld, World},
+};
+use core::{fmt::Debug, marker::PhantomData, ops::Deref};
+
+use super::OneToOneEvent;
+
+/// Represents one half of a one-to-one relationship between two [entities](Entity).
```
nakedible-p, line 17: It might be reasonable to mention that this is a symmetric or undirected one-to-one relationship. After adding the component, you can't tell which points to which.

There is also a need for `OneToOne` relationships that are directed, or not symmetric. Like:
```
struct Train;
type PreviousCar = OneToOneSource<Train>;
type NextCar = OneToOneTarget<Train>;
```
A single entity may be at the same time the source and the target of a relation towards different objects.

crates/bevy_hierarchy/src/one_to_one/event.rs
```diff
@@ -0,0 +1,123 @@
...
+    }
+
+    /// Create a new [`Removed`](OneToOneEvent::Removed) [`Event`]
+    pub const fn removed(primary: Entity, secondary: Entity) -> Self {
+        Self::Removed(OneToOneEventDetails::new(primary, secondary))
+    }
+
+    /// Get the primary [`Entity`] in this [`Event`].
+    /// The primary is the _cause_ of the event, while the secondary is the relation.
+    pub const fn primary(&self) -> Entity {
```
nakedible-p, line 27: Could consider `source` and `target`, even though the edge isn't directed. I'm not terribly fond of primary and secondary, although this is of course bikeshedding.

nakedible-p (commented):
Very cool work! I've spent way too long making ad-hoc versions of the exact same thing, so it's going to be extremely nice to convert all of those to generic versions that work better.

nakedible-p:
Whoops, accidentally was logged in with work account. @nakedible is my primary open source account.

crates/bevy_hierarchy/src/one_to_many/many.rs
bushrat011899: Ok found it [here](https://github.com/bevyengine/bevy/blob/d15d901cab763a1619f4dffbfd0418618daebe7f/crates/bevy_ecs/src/bundle.rs#L904). The `BundleInserter` `insert` method will trigger the `on_replace` hook without first calling any `on_remove` hooks (if applicable). Because of that, you _must_ have hooks for both `on_remove` _and_ `on_replace` to capture all cases of a `Component` being taken off of an `Entity`.

bushrat011899: Although your original question holds, since if `on_remove` is called,  `on_replace` is always called, so I _think_ you can omit the `on_remove` hook. I'll test that now...

bushrat011899: Yeah you're right, removing the `on_remove` for the sake of simplicity. Good catch!

nakedible: Yeah, I believe the way the hooks work is:
- on_insert gets called every time a component is added by any means
- on_replace gets called every time a component is removed by any means
- on_add gets called only if the component didn't exist yet on the entity
- on_remove gets called only if the component will no longer exist on the entity

So on_replace indeed catches all "removals".

Good to hear there weren't any bugs uncovered by this, as `on_replace` seems to be quite new.

nakedible:
I wonder if it would be good to document somewhere that
```rust
entity.insert(Parent::new(other));
```
will maintain the invariants but something like:
```rust
entity.get_mut::<Parent>().unwrap().clone_from(other_parent);
```
will not. So just calling out that parent should never be mutated, but instead must be replaced as a whole by `insert` for things to work.

crates/bevy_hierarchy/src/one_to_one/component.rs
pablo-lua: In the future, we can maybe add ManyToMany relationship, it probably can be useful overall, not sure of the use cases.

pablo-lua (commented):
Did some review. BTW, this PR says to close #15869, but this PR doesn't really add spawn_batch to said trait, but it does make trivial to impl this change.

crates/bevy_hierarchy/src/one_to_one/component.rs
bushrat011899: After refactoring how these components work, I've added `ManyToMany`, with directed and undirected forms for `OneToOne` and `ManyToMany`. I haven't added an undirected form of `OneToMany` as I'm not entirely sure what that actually represents, and I'd rather not invent it for the sake of completeness.

crates/bevy_hierarchy/src/one_to_one/component.rs
bushrat011899: I have added directed `OneToOne` relationships, along with `ManyToMany` as well.

crates/bevy_hierarchy/src/one_to_one/event.rs
bushrat011899: I've decided to adopt the terminology `primary` and `foreign`, as that maps well to the relational database terminology that these changes are most analogous to.

crates/bevy_hierarchy/src/despawn_recursive/command.rs
```diff
@@ -0,0 +1,139 @@
...
+    pub const fn with_warn(mut self, warn: bool) -> Self {
+        self.warn = warn;
+        self
+    }
+
+    /// Control whether this [`Command`] should also despawn the target [`Entity`] (`true`)
+    /// or on its descendants (`false`).
+    pub const fn with_inclusion(mut self, inclusive: bool) -> Self {
+        self.inclusive = inclusive;
+        self
```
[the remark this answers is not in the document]
bushrat011899: Good point, included!

crates/bevy_hierarchy/src/one_to_many/many.rs
```diff
@@ -0,0 +1,307 @@
...
+    pub fn without(mut self, other: Entity) -> Self {
+        self.entities.retain(|&mut e| e != other);
+        self
+    }
+
+    pub(super) fn entities_mut(&mut self) -> &mut SmallVec<[Entity; 8]> {
+        &mut self.entities
+    }
+
+    /// Swaps the entity at `a_index` with the entity at `b_index`.
```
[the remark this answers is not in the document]
bushrat011899: I've added a `#Panics` section to the documentation for this method.

crates/bevy_hierarchy/src/despawn_recursive/command.rs
```diff
@@ -0,0 +1,139 @@
...
+#[derive(Debug)]
+pub struct DespawnRecursive<C> {
+    /// Target entity
+    entity: Entity,
+    /// Whether or not this command should output a warning if the entity does not exist
+    warn: bool,
+    /// Whether this command will despawn the provided entity (`inclusive`) or just
+    /// its descendants (`exclusive`).
+    inclusive: bool,
+    /// Marker for the
```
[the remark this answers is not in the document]
bushrat011899: Whoops!

crates/bevy_hierarchy/src/one_to_many/mod.rs
```diff
@@ -0,0 +1,176 @@
...
+    use super::*;
+
+    /// A familial relationship
+    struct Family;
+
+    /// Shorthand for a Parent in a Family relationship
+    type Parent = ManyToOne<Family>;
+
+    /// Shorthand for a Parent in a Family relationship
+    type Children = OneToMany<Family>;
```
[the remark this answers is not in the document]
bushrat011899: My rationale here is I wanted to provide generic relationship types, which would mean `Parent` and `Children` would simply be a specific case of a general relationship. I believe I have completely replicated all functionality these types originally had in the new generic ones, so there shouldn't be any degradation in user experience.

The alternative was to either not add the generic relationships (seems like a shame to exclude), or duplicate the work for no real tangible benefit.

crates/bevy_hierarchy/src/one_to_many/one.rs
```diff
@@ -0,0 +1,200 @@
...
+pub struct ManyToOne<R> {
+    entity: Entity,
+    #[cfg_attr(feature = "reflect", reflect(ignore))]
+    _phantom: PhantomData<fn(&R)>,
+}
+
+impl<R: 'static> ManyToOne<R> {
+    fn associate(mut world: DeferredWorld<'_>, a_id: Entity, _component: ComponentId) {
+        world.commands().queue(move |world: &mut World| {
+            let b = world
```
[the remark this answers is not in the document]
bushrat011899: I do agree, I struggled with this as I was writing it. After some deliberation, I've decided on `primary` and `foreign` instead of `a` and `b`, and I've included some documentation explaining what those names mean, and their lineage (relational database terminology)

crates/bevy_hierarchy/src/one_to_many/one.rs
```diff
@@ -0,0 +1,200 @@
...
+                .map(|a_relationship| a_relationship.entity)
+                .and_then(|b_id| world.get_entity_mut(b_id).ok());
+
+            let Some(mut b) = b else { return };
+
+            let b_id = b.id();
+
+            let b_points_to_a = b
+                .get::<OneToMany<R>>()
+                .is_some_and(|b_relationship| b_relationship.contains(&a_id));
```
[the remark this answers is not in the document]
bushrat011899: For large relationships an `EntityHashSet` would be more appropriate, but `SmallVec` allows us to avoid allocation for small relationship graphs, which I believe is more common (or at least, it's the status-quo). Ideally, we would have some `SmallEntityHashSet` type which used a `SmallVec` optimisation for a small quantity of entities, and then switched to a `EntityHashSet` for larger quantities.

However, I think this would complicate an already challenging PR for an unknown performance win, so I'd rather save this for a future PR.

crates/bevy_hierarchy/src/many_to_many.rs
```diff
@@ -0,0 +1,334 @@
...
+    pub fn without(mut self, other: Entity) -> Self {
+        self.entities.retain(|&mut e| e != other);
+        self
+    }
+
+    /// Swaps the entity at `a_index` with the entity at `b_index`.
+    ///
+    /// # Panics
+    ///
+    /// Will panic of either index is out-of-bounds.
```
bushrat011899, line 184:
```suggestion
    /// Will panic if either index is out-of-bounds.
```

crates/bevy_hierarchy/src/one_to_many.rs
```diff
@@ -0,0 +1,438 @@
...
+    pub fn without(mut self, other: Entity) -> Self {
+        self.entities.retain(|&mut e| e != other);
+        self
+    }
+
+    /// Swaps the entity at `a_index` with the entity at `b_index`.
+    ///
+    /// # Panics
+    ///
+    /// Will panic of either index is out-of-bounds.
```
bushrat011899, line 181:
```suggestion
    /// Will panic if either index is out-of-bounds.
```

iiYese (changes requested):
This PR is like an [aery](https://github.com/iiYese/aery) upstream minus many features. I avoided upstreaming aery.
- There is no reason for the components that store the edge data `OneToOne`, `ManyToOne`, to have the exclusivity be expressed at a type level. This makes it so if you want to query for a relationship you need to know/provide this information as well which makes it awkward & verbose. Aery provides this configuration via a trait const enum it does not make it type level.
- The component types that store the edges are public. This means you can break invariant by modifying these components. The `Hosts` & `Targets` components in aery are private for this reason. They can only be changed by commands & querying for edge information is done via a `QueryData` struct.
-  Aery originally started out as a PR for bevy but when studying the subject more and specifically flecs's implementation I decided I didn't want it to be bevy's solution. The big difference between this approach (storing edges inside a component) & fragmenting relationships is the edge data is stored at an *archtype level* with fragmenting relationships. This means we can reuse many of our ECS tools for query operations as well as the optimizations that come with them. Not to say that it is an easy task but it is significantly easier to extend & optimize fragmenting relationships than it is to extend & optimize this. Aery was always meant to be deprecated when proper relationships were added to bevy. That is much harder for bevy to do if it is the one providing this tooling because that kind of breakage will bleed much deeper into the ecosystem. I think hierarchy as it currently is needs to be dropped & that is easier to do when it is still very limited in what it does. Extending it beyond that makes it harder. For that reason (especially given that mvp relationships is a lot closer now than it was when I first released aery) I think all of the generic relationship stuff in this PR should be dropped & it should focus on
  - hierarchy QOL
  - making cleanup use hooks
  - making hierarchy invariants unbreakable by making `Parent` & `Children` private

bushrat011899:
Thank you for such a detailed write-up! I'll write a proper response shortly, but I want to state immediately that making `Parent` and `Children` private does _not_ prevent breaking the invariants, since you can still access private components via component IDs. It would also require writing more supporting code to still allow for `Query` filtering and fetching. No more challenging than what I've written in this PR, but equally redundant once real relationships are added.

iiYese:
> since you can still access private components via component IDs
You still need a type to cast to which you can't do if the component itself it private. Unless there's something I'm missing.

crates/bevy_hierarchy/src/one_to_many/one.rs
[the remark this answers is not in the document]
nakedible: I think the current implementation is fine and good. But in case you do want to explore this further, please see https://docs.rs/tinyset/latest/tinyset/ . Basically, we'd want a `tinyset::Set64` that is optimized for `Entity` types in a way where small entity ids with small generation numbers would end up being small numbers, possibly meaning that we'd need to do Morton encoding on the entity id. But that's a very specific optimization so should be a separate pull with measurable speed benefits.

nakedible:
My two cents on this:

> This PR is like an [aery](https://github.com/iiYese/aery) upstream minus many features.

I agree 😉 

> * There is no reason for the components that store the edge data (`OneToOne`, `ManyToOne`, etc.) to have the exclusivity be expressed at a type level. This makes it so if you want to query for a relationship you need to know/provide this information as well which makes it awkward & verbose. Aery provides this configuration via a trait const it does not make it type level.

There is one important reason why, and that is of the interface for using these. A one-to-many relationship component derefs to `Entity`, where as as a many-to-one relationship component derefs to `[Entity]`. I believe it is nicer that the type has a different interface based on the exclusivity - it makes the difference explicit in code, and incorrect usage not compile,  and improves error messages. Unwrapping single element arrays, or having two different APIs (there's only on `Defer`) is not what I want to do.

> * The component types that store the edges are public. This means you can break invariants by modifying these components since they're held up by hooks which rely on archetypal changes. The `Hosts` & `Targets` components in aery are private for this reason. They can only be changed by commands & querying for edge information is done via an immutable `QueryData` struct.

While I agree it is safer, I personally think `entity.insert(Parent::new(other));` feels very Bevy, where as `entity.set::<Parent>(other);` does not. It's easier to reason about what is happening, and it leverages the component model fully. The magic happens in the component hooks to also set up respective components on the other side. There isn't a new concept to learn (relations), or wondering how it is implemented and where the data is stored - it's just a component with entities. It's also clear how this interfaces with scenes, reflection, etc. in a readable manner. You can't write `entity.set::<Parent>(other);` in a `ron` file if you are not even told what the underlying storage component is and it's private, but you can add a `ManyToOne<Family>` component, even if the type alias wouldn't work.

It's also closer to what `Parent` and `Children` are currently, so it's a much less invasive change for the user in the beginning, regardless of what the final interface for relations is going to look like. And naming the two halves is important, as it's much easier to reason about `Parent` and `Children` than some `Family` relationship.

The downside here is of course that if these are exposed, then they become a part of the public API, so then it's harder to change the implementation of relations while keeping the same API. Private components would be much easier to change.

> * I avoided upstreaming aery. Aery originally started out as a PR for bevy but when studying the subject more and specifically flecs's implementation I decided I didn't want it to be bevy's solution. The big difference between this approach (storing edges inside a component) & fragmenting relationships is the edge data is stored at an _archtype level_ with fragmenting relationships. This means we can reuse many of our ECS tools for query operations as well as the optimizations that come with them. Not to say that it is an easy task but it is significantly easier to extend & optimize fragmenting relationships than it is to extend & optimize this. Aery was always meant to be deprecated when proper relationships were added to bevy. That is much harder for bevy to do if it is the one providing this tooling because that kind of breakage will bleed much deeper into the ecosystem. I think hierarchy as it currently is needs to be dropped & that is easier to do when it is still very limited in what it does. Extending it beyond that makes it harder. For that reason (especially given that mvp relationships is a lot closer now than it was when I first released aery) I think all of the generic relationship stuff in this PR should be dropped & it should focus on:
>   
>   * hierarchy QOL
>   * making cleanup use hooks
>   * making hierarchy invariants unbreakable by making `Parent` & `Children` private

While I do understand where you are coming from, I'm not sure I agree on all points. This is a larger discussion, but quickly a couple of points:

- Fragmenting vs. non-fragmenting is a performance optimization issue, not an API issue. This pull takes the API in the opposite direction from what I've seen of the relations work so far. If this API is exposed to users, it will make it significantly harder to migrate to another kind of API, as compared to just having `Parent` and `Children` to migrate. Then again, Bevy is no stranger to changes which require a lot of work to upgrade from those who use the API, so it's a judgement call. A decision should be made on the API as well, or the relationship API discussion should be separated from the relationship implementation discussion.
- Fragmenting relationships are totally useless for my use cases, and will just lead to one archetype per entity. Non-fragmenting relationships based on relationship type (exactly like it is here) is what I need, regardless of the API. Even Sanders suggested me to use non-fragmenting relationships for my needs, so getting those exactly right is also really important. I'd be really happy if non-fragmenting relationships would not be gated behind all the work required to make fragmenting relationships work – but I do understand that getting the API right for both of them is probably easier if fragmenting relationships are a bit further along.

bushrat011899:
> This PR is like an [aery](https://github.com/iiYese/aery) upstream minus many features.

Full transparency: I haven't personally used Aery. I'm coming at this from the perspective of someone who works with relational databases (MSSQL, etc.). Looking at Aery, it's definitely _far_ more feature-complete than what I'm proposing here. What I will say is it's not my intention to match (nor exceed!) what Aery offers. And it is certainly not my intention to ship a half-baked dead-end feature that will only add to migration pains once MVP relations lands in Bevy.

In my opinion, the features that I am adding:

* One-to-One Directed and Undirected
* One-to-Many Directed
* Many-to-Many Directed and Undirected

Are all fundamental to any relations feature Bevy eventually ships (so much so that we already have `Parent` and `Children` in a worse state than what this PR provides). If MVP relations lands _without_ support for these fundamental relationships I don't think it's _really_ a viable MVP to be brutally honest. This PR doesn't include new syntax, `Query` tools, or any changes to Bevy's ECS model, so migration from this to true relations will be as simple as replacing a `Component` insertion with whatever syntax true relations will use.

>     * There is no reason for the components that store the edge data (`OneToOne`, `ManyToOne`, etc.) to have the exclusivity be expressed at a type level. This makes it so if you want to query for a relationship you need to know/provide this information as well which makes it awkward & verbose. Aery provides this configuration via a trait const it does not make it type level.

It _is_ possible to refactor my current PR to make the relationships have runtime validation of these invariants instead of compile-time, but I don't think I understand _why_ you would want this. If you _don't_ want these invariants managed, you can use `ManyToMany` and ignore that the other `Components` even exist. This PR is quite simplistic in that you _can_ use a `ManyToMany` as a one-to-one relationship, or as a one-to-many, if you really want to. All it does is gives the user the option of having the invariant managed at a type level if they wish to.

Looking at `Parent` and `Children` as an example, we already desire that invariant be codified in the type system, since it's required by many areas of Bevy that every `Entity` have at-most one `Parent`. I do think having a more loosely managed relationship type is important, but that role is provided by `ManyToMany` already.

>     * The component types that store the edges are public. This means you can break invariants by modifying these components since they're held up by hooks which rely on archetypal changes. The `Hosts` & `Targets` components in aery are private for this reason. They can only be changed by commands & querying for edge information is done via an immutable `QueryData` struct.

I _do_ agree with this idea, which is why all mutable methods provided for these edge types require owned (`self`) access to the component. This doesn't prevent users `swap`-ing data, but this is an existing issue improved by this PR (removing `Parent` and `Children` is a trivial example of breaking these invariants that this PR resolves). This PR makes all trivial `Component` operations invariant-preserving.

* Inserting a relationship onto an `Entity` will always create the appropriate partner component on the foreign `Entity`
* Removing will always dispose of dangling components
* Replacing will always update old and new entities to the new arrangement
* There are no mutable methods on any relationship components, you _must_ use an invariant preserving archetypical operation (insert, remove, replace), or a method like `core::mem::swap`.

If the above guarantees are insufficient, we _can_ adopt the techniques you've outlined (private components, `QueryData` struct, etc.) as well, they are not mutually exclusive to the techniques used in this PR (as can be seen by leaving `ChildBuilder`, etc. available).

>     * I avoided upstreaming aery. Aery originally started out as a PR for bevy but when studying the subject more and specifically flecs's implementation I decided I didn't want it to be bevy's solution. The big difference between this approach (storing edges inside a component) & fragmenting relationships is the edge data is stored at an _archtype level_ with fragmenting relationships. This means we can reuse many of our ECS tools for query operations as well as the optimizations that come with them.

Not a major point for or against, but I think it's self-evident that this PR does reuse existing ECS tooling, since it is exclusively confined to the `bevy_hierarchy` crate with no changes required in `bevy_ecs`. I do think fragmenting relationships is a highly important feature for Bevy to ship, but non-fragmenting ones are also an important feature to support.

> Not to say that it is an easy task but it is significantly easier to extend & optimize fragmenting relationships than it is to extend & optimize this. Aery was always meant to be deprecated when proper relationships were added to bevy. That is much harder for bevy to do if it is the one providing this tooling because that kind of breakage will bleed much deeper into the ecosystem.

I personally don't see how this PR creates a migration issue any worse than any other PR added to Bevy; it introduces 4 components which _should_ be trivially replaceable once Relations lands. Further, they don't need to be removed to land relations; they can be marked as depreciated and left as-is for as long as Bevy supports `Components` and `ComponentHooks`.

> I think hierarchy as it currently is needs to be dropped & that is easier to do when it is still very limited in what it does. Extending it beyond that makes it harder.

Again, I'm only introducing 4 components which could be moved to wherever the `ChildBuilder` methods get moved to. Alternatively, since these components exclusively rely on the public API of `bevy_ecs`, the worst-case-scenario migration guide would be to provide those components to users to include in their project themselves. Definitely not an option I would champion, but absolutely viable.

> For that reason (especially given that mvp relationships is a lot closer now than it was when I first released aery) I think all of the generic relationship stuff in this PR should be dropped

I personally disagree with this conclusion. I think at most, we should put this behind an `experimental` feature flag similar to `GhostNodes` in #15961. Again, I believe these components offer a very simple API which would be very straightforward to migrate from with the real relations MVP.

> it should focus on:
>       
>       * hierarchy QOL
>       * making cleanup use hooks

I agree with these points, and I do believe this PR addresses these concerns in ergonomics, reliability, and performance.

>       * making hierarchy invariants unbreakable by making `Parent` & `Children` private

Addressed above. I'm not against this, but as others have mentioned, this creates issues for BSN and `ron`, since you don't have a named component to use in those formats. I'm honestly not sure how we can integrate relations into those foundational parts of Bevy without named components, but I am not a subject matter expert like yourself so this may be something already considered and I am just unaware!

I really want to stress that I greatly appreciate you taking the time to review this PR. I disagree with some of your points but I am absolutely willing to change this PR if it's the general consensus that what you're proposing is the correct direction for Bevy. I am pushing back because I have seen several instances of people requesting this functionality as-presented, and I genuinely believe it is a positive direction to take Bevy, even if it's only for one or two releases.

iiYese:
@nakedible 
> There is one important reason why, and that is of the interface for using these. A one-to-many relationship component derefs to `Entity`, where as as a many-to-one relationship component derefs to `[Entity]`. I believe it is nicer that the type has a different interface based on the exclusivity - it makes the difference explicit in code, and incorrect usage not compile, and improves error messages. Unwrapping single element arrays, or having two different APIs (there's only on `Defer`) is not what I want to do.

The correctness here is actually worse because you can have both `OneToMany<R>` & `OneToOne<R>` on the same entity. This kind of API design has been misguiding bevy for a long time. Things being type level does not mean magically more correct. And besides that it makes the ergonomics terrible. You should be able to query for a relationship without knowing its edge rules.

> While I agree it is safer, I personally think entity.insert(Parent::new(other)); feels very Bevy, where as entity.set::<Parent>(other); does not... 

You can have a bundle API with a hook instead of a command if you really want. Either way the edge types should be private. Sacrificing upholding edge invariants for a slightly different method call is weird especially looking at your previous comment. The energy spent on upholding correctness is misplaced.

> Fragmenting vs. non-fragmenting is a performance optimization issue

No it's not. Archetype fragmentation is a different tool entirely. You cannot do many query operations when edge information is not exposed at an archetype level. 

> Fragmenting relationships are totally useless for my use cases, and will just lead to one archetype per entity

No it won't. You're either imagining only top down edge scenarios or you have the worst case scenario. If you have a relation like `Child` then you create a lot of needless fragmentation. If you instead have a relation like `Parent` or flecs's `ChildOf` you get perfect fragmentation. All children of an entity end up in the same table which gives you cache friendly dense iteration when traversing hierarchies.

If you instead have the actual worst case scenario and lots just of relations this is already the case in mainstream OO engines which perform more than fine.

Non-fragmenting relationships aren't mutually exclusive to fragmenting relationships either but they're best not implemented like this PR. The way flecs does them iirc is with sparse components.

iiYese:
@bushrat011899

> I don't think it's really a viable MVP to be brutally honest

I think you don't understand what am MVP is. An MVP is not complete functionality it is a minimal viable product. MVP relations will probably just be stuff like `Parent` to begin with. Enough to replace the current hierarchy. The tools extend them to do things like having x-to-x rules & cleanup are already there (hooks & observers) & shouldn't be a complicated follow up.

>  All it does is gives the user the option of having the invariant managed at a type level if they wish to.

Consistency is more important here than options. Why should users have to care about whether or not the information is erased & if it's not need to know that information every time they want to operate on edges? For what benefit? Being at a type level is not a benefit it is the option.

> Looking at Parent and Children as an example, we already desire that invariant be codified in the type system, since it's required by many areas of Bevy that every Entity have at-most one Parent

That is not the same. That is different sides of an edge not entirely different relationship types. 

> This PR makes all trivial Component operations invariant-preserving

The scenario that isn't covered was already mentioned
```rs
let e = world.entity(e0).get::<Parent>().get();
*world.entity_mut(e1).get_mut::<Parent>() = Parent::new(e);
```

> Not a major point for or against, but I think it's self-evident that this PR does reuse existing ECS tooling...

That's not what I'm referring to & as mentioned in my previous response to @nakedible non-fragmenting relationships are best not implemented like this. You will end up with entirely different APIs to do queries for fragmenting & non-fragmenting relations.

> I personally don't see how this PR creates a migration issue any worse than any other PR added to Bevy; it introduces 4 components which should be trivially replaceable once Relations lands.

It's not just replacing the component types it's also replacing *all* of the accompanying code doing any kind of operations on entities which will look very different. You will also have people relying on the fact that these are non-fragmenting which is just a complete break instead of a migration.

bushrat011899:
> > Looking at Parent and Children as an example, we already desire that invariant be codified in the type system, since it's required by many areas of Bevy that every Entity have at-most one Parent
> 
> That is not the same. That is different sides of an edge not entirely different relationship types. 

Sorry I don't think I made my point very clear here. What I was trying to convey is that a Bevy `Family` relationship (`Parent` and `Children`) already has a one-to-many invariant codified into the type system. `Parent` _must_ refer to a single entity, it cannot under any circumstances refer to multiple entities. Building relationships without support for those kinds of type-level constraints degrades the user experience and sacrifices correctness. Requiring a user to write:

```rust
// not valid code, representative only
let [parent] = e0.get::<Parent>() else { unreachable!() }; 
```
 Is a terrible user experience. As such, the MVP for relationships _must_ support defining a one-to-many relationship to be able to replace the current `Parent` and `Children` components. From personal experience, I can also say that implementing a one-to-one relationship is definitely simpler, so would likely be included in the MVP anyway.

As for including `OneToOne<R>` and `OneToMany<R>` on the same entity, I personally don't consider that a correctness issue, because the parameter `R` is a marker used in a relationship, the relationship itself is the component _and_ the marker combined. That's why in all the examples I've provided I've included type aliases for those components with their markers.

Knowing that a relationship is one-to-one, one-to-many, etc. is also a perfectly reasonable assumption to make of the user in my opinion. If the user doesn't know what the relationship is, what could they meaningfully write with them? Again, if they don't care, a many-to-many is always there.

iiYese:
> Bevy Family relationship (Parent and Children) already has a one-to-many invariant codified into the type system. Parent must refer to a single entity, it cannot under any circumstances refer to multiple entities.

That's more to do with the way hierarchy is implemented. When you have components for edges on entities of course the edges mean something different depending on if you're at the base or head of an edge. With proper relationships we will only have `Parent`. `Children` will be gone. The edge rules will still be "codified" but it will be a part of the relation itself. The only time you should be stating a relation type with its edge rules is when you're defining it. It's implicitly apart of the relation form there. It makes no sense for users to need to restate this definition every time they query for a relationship. 

> As for including OneToOne<R> and OneToMany<R> on the same entity, I personally don't consider that a correctness issue

I do. This is just erroneous & I would block on this if I had the power. When would it ever make sense to have both? It is not useful to just provide more ways to do things. You need to provide ways to *easily* do things that make sense that people actually want to do.  

> That's why in all the examples I've provided I've included type aliases for those components with their markers. If the user doesn't know what the relationship is, what could they meaningfully write with them?

You're contradicting yourself here. It can't both be that "they should always know this information" & "but I erased the information they should always know in the examples". It's not just needing to know it's also having to restate it every time. Redundant repetition is a big source of errors in programming.

nakedible:
@iiYese I'll answer to the rest a bit later, but:

> No it's not. Archetype fragmentation is a different tool entirely. You cannot do many query operations when edge information is not exposed at an archetype level.

Can you list the query operations that require and archetype based relations implementation, as opposed to component based?

iiYese:
> Can you list the query operations that require and archetype based relations

Some of the important ones off the top of my head:
- Wildcard queries `(*, e)`
- Named wildcard queries `&X, (Parent, $p), Ship($p)` (get the `X`s from all entities that have a parent that is a ship).
- Nested joins `* -R-> * -R-> *`
- Efficient up traversal

crates/bevy_hierarchy/src/one_to_one/event.rs
nakedible: Thanks, I hate it! 😉

bushrat011899:
Based on pushback on the functionality provided here I have removed all new functionality from this PR with the exception of hook-based management of `Parent` and `Children`, the generalisation of `DespawnRecursive`, and the ability to insert `Parent` and `Children` onto an entity directly. If someone else wishes to ship this functionality they are more than welcome to add them back in through a follow-up PR. I apologise to anyone who was excited for this functionality to come in this PR. I encourage you to look towards #3742 for true relationships coming in the not-too-distant future.

nakedible:
I'm sad that this is the outcome, but on the other hand, I do believe it's the right outcome for this pull, so that the non-controversial parts are not gated behind the controversial parts.

However, I would suggest that there is a follow-up pull which would reinstate the removed generic relations functionality behind an experimental flag. Then we can discuss those in isolation without combining it with the fixes to the current Parent/Children functionality.

Also, I don't think this pull is the right place to discuss @iiYese 's objections – so I'll take to answering to those elsewhere.

iiYese (changes requested):
So a lot of the relationship stuff is still there which becomes unused complexity that has to be maintained. The component types are still public which is the last thing that allows breaking invariants. I would suggest:
- Getting rid of all the remaining relationship stuff. Traits, generics etc. turning `OneToMany<R>` into `ChildrenInner` with an accompanying `ParentInner` that are *private*.
- Making immutable `QueryData` structs `Children` & `Parent` for easy migration. 

This PR got quite big & is trying to address multiple issues at once. Before this discussion was also trying to shoehorn new features in with the PR that were unrelated to the issues. In future keep the PR small & focused on a single issue & make different PRs for new features. Small PRs are easier to review & merge & will save you from these kinds of head aches.

bushrat011899:
Closing as I don't see any reason to continue changing this PR. Based on feedback, a more generally accepted solution would be to take the `associate` and `disassociate` hooks from this PR, add them to the existing `Parent`and `Children` components, rename them, privatise them, and create `QueryData` structs with the original names to allow read-only access. You'll _still_ be able to break invariants (e.g., by using the `ComponentId`), but the surface area would be reduced.

I personally disagree with this approach, so I would invite someone else to take over from here who has a resolute belief in this approach. Thanks again to everyone who reviewed this PR, and I apologise I couldn't see it through.

nakedible:
Okay, now it became the saddest outcome 😭

NthTensor:
Missed this initially. Disappointed this ended up getting closed. I'd like to float this by the ECS SME team. @alice-i-cecile @JoJoJet @maniwani do yall think you could give this some consideration after the 0.15 release?

alice-i-cecile:
To summarize: 

- hook based parent / child is good, which can be used to eliminate despawn_recursive
- move from current to ideal design with no detours
- smaller PRs with a clearer plan 
- read only components are a generally useful tool for further ensuring hierarchy validity

My broad position is that I'm fine with a very slow initial relations implementation, but we should try to target the ideal correctness/API design to avoid confusion. Moving incrementally towards that is welcome, and prototyping with just Parent/Children is going to be easiest :)

Issue #10455, opened 2023-11-08
0.12 - Method for us to obtain asset `root_path`


Testare:
## What problem does this solve or what need does it fill?

tldr; In 0.11 there was a way to obtain the root path for assets from code, but now it is not possible AFAICT.

I'm building a game that I require to be flexible for expansion. So in 0.11 I had code that searches the asset directory for all "registry files," so that mod/expansion content can simply be added as folders to the assets directory. This was done by getting the `root_path()` from `FileAssetIo` and then using the `glob` crate to search for `"**/*.reg.toml"` and load these files.

However, in 0.12 `AssetIo` was replaced with `AssetSource` and `AssetReader`. I'm excited for this change, but we also lost the ability to downcast `AssetReader`, so even though `FileAssetReader` has the `root_path()` method, I have no means of actually calling this function, and I am unable to migrate my code to 0.12.

## What solution would you like?

I would like to add the `Downcast` trait to `AssetReader`.  I'm not the only one to miss this functionality [\[1\]](https://discord.com/channels/691052431525675048/1171470656894947379), and I think it would be useful for many different plugins. The downside to this 

### Add `(prefix?_)root_path` to AssetReader, AssetSource, or AssetServer

For AssetReader: While not every implementation would return a path (for example, `embedded`), we could make AssetReader return an `Option<&Path>`. This way we could unblock me on 0.12 migration, and by making it return `None` by default we could make it not a breaking change and hopefully make this into a 0.12.X release so I don't have to wait until 0.13 to migrate. The downside is that other people might have different reasons for wanting to downcast AssetReaders, and this wouldn't really solve it for them.

For AssetServer/AssetSource: It would be nice if I didn't have to dig so deep in the type tree to get this information and was just able to get it from an AssetServer directly. It would also make it more likely that we don't lose this functionality in the future. However this does sort of feel like "exposing internals"

### `load_glob()` for AssetServer

This technically doesn't allow us to obtain a `root_path`, but it would solve my personal problem perhaps more elegantly than simply being able to obtain the root path again. It could behave similarly to `load_folder`, but allow us to specify things more dynamically. I can see this being a very useful tool for

[...]

PR #15639, opened 2024-10-04
Add `AssetPath::exists`

### LINKED ISSUE #14427: Add `AssetPath::exists`

Conditionally loading assets if a matching pattern exists is useful for data-driven behavior: you may only want to load a sound effect if a file with a matching name exists.

Checking whether the file is present `Path::exists` is the obvious solution, but is actually a footgun.
`Path` is an absolute path, relative to filesystem, while `AssetPath` is returned relative to the `assets` folder that Bevy is configured to use.

Add `AssetPath::exists`, which pre-parses the path.

We could also spin out something that returns a check further down the asset pipeline: whether an asset exists and whether a correct loader is found for example.

That would need to be a method on `AssetServer`.

RyanSpaker:
# Objective

- Add functionality to determine if an `AssetPath` references an actual file/directory or not.
- Fixes #14427

## Solution

- Added a `AssetPath::exists` method which returns whether the path references an existing file or directory.
- The method pre-parses the path to be relative to the default "assets" folder, returning false if the `AssetPath` uses a custom source.
- The existence check uses the std `Path::exists` method.

## Testing

- Tested using a small test app. `Assetpath`s that referenced real asset files returned true, and the rest returned false.
- I tested on linux, but since the method uses the standard library's `Path` struct, it should work on any system.

### DIFF
--- a/crates/bevy_asset/src/path.rs
+++ b/crates/bevy_asset/src/path.rs
@@ -474,7 +474,10 @@ impl<'a> AssetPath<'a> {
             return None;
         }
         let path = self.path();
-        let path = path.is_absolute().then(|| path.strip_prefix("/").ok()).flatten()?;
+        let path = path
+            .is_absolute()
+            .then(|| path.strip_prefix("/").ok())
+            .flatten()?;
         Some(Path::new(crate::AssetPlugin::DEFAULT_UNPROCESSED_FILE_PATH).join(path))
     }
 
--- a/crates/bevy_asset/src/path.rs
+++ b/crates/bevy_asset/src/path.rs
@@ -465,27 +465,36 @@ impl<'a> AssetPath<'a> {
         Some(extension)
     }
 
-    /// Returns whether the asset file exists using the std [`Path::exists`] method.
+    /// Returns the path of the [`AssetPath`] relative to the app root.
+    /// Returns None if the [`AssetPath`] uses a non-default Asset Source.
     ///
-    /// Returns false when the asset does not exist or if the [`AssetPath`] uses a custom [`AssetSourceId`].
-    pub fn exists(&self) -> bool {
+    /// Ex: an [`AssetPath`] referencing "/models/apple.gltf" would return a [`PathBuf`] for "assets/models/apple.gltf"
+    pub fn get_root_relative_path(&self) -> Option<PathBuf> {
         if let AssetSourceId::Name(_) = self.source {
-            return false;
+            return None;
         }
         let path = self.path();
         let path = if path.is_absolute() {
             match path.strip_prefix("/") {
                 Ok(path) => path,
                 Err(_) => {
-                    return false;
+                    return None;
                 }
             }
         } else {
             path
         };
-        Path::new(crate::AssetPlugin::DEFAULT_UNPROCESSED_FILE_PATH)
-            .join(path)
-            .exists()
+        Some(Path::new(crate::AssetPlugin::DEFAULT_UNPROCESSED_FILE_PATH).join(path))
+    }
+
+    /// Returns whether the asset file exists using the std [`Path::exists`] method.
+    ///
+    /// Returns false when the asset does not exist or if the [`AssetPath`] uses a custom [`AssetSourceId`].
+    pub fn exists(&self) -> bool {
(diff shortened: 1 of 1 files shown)


alice-i-cecile (commented):
Can we extract all of this logic up to the `.exists` part into a helper function that we use in multiple places?

RyanSpaker:
> Can we extract all of this logic up to the `.exists` part into a helper function that we use in multiple places?

I agree, I added a method to get a path relative to the app root. I'm not sold on the name I chose, but I couldn't think of a better one.

crates/bevy_asset/src/path.rs
```diff
@@ -465,6 +465,38 @@ impl<'a> AssetPath<'a> {
         Some(extension)
     }
 
+    /// Returns the path of the [`AssetPath`] relative to the app root.
+    /// Returns None if the [`AssetPath`] uses a non-default Asset Source.
+    ///
+    /// Ex: an [`AssetPath`] referencing "/models/apple.gltf" would return a [`PathBuf`] for "assets/models/apple.gltf"
+    pub fn get_root_relative_path(&self) -> Option<PathBuf> {
```
richchurcher, line 472: I see what you mean, it's a tricky name. On balance, I think you could get away with `get_relative_path` because the project root is implied, but not sure TBH.

crates/bevy_asset/src/path.rs
```diff
@@ -465,6 +465,38 @@ impl<'a> AssetPath<'a> {
...
+            match path.strip_prefix("/") {
+                Ok(path) => path,
+                Err(_) => {
+                    return None;
+                }
+            }
+        } else {
+            path
+        };
+        Some(Path::new(crate::AssetPlugin::DEFAULT_UNPROCESSED_FILE_PATH).join(path))
```
richchurcher, line 487: Something's bugging me about using the default here, because eventually I think we might end up needing to get the set value of `AssetPlugin::file_path` to cater to folks who don't want to use the default (see #10455 for example). I don't think it blocks merging this though.

richchurcher (approved):

crates/bevy_asset/src/path.rs
```diff
@@ -465,6 +465,38 @@ impl<'a> AssetPath<'a> {
...
+            return None;
+        }
+        let path = self.path();
+        let path = if path.is_absolute() {
+            match path.strip_prefix("/") {
+                Ok(path) => path,
+                Err(_) => {
+                    return None;
+                }
+            }
```
kristoff3r, line 483:
```suggestion
            path.strip_prefix("/").ok()?
```

crates/bevy_asset/src/path.rs
```diff
@@ -465,6 +465,38 @@ impl<'a> AssetPath<'a> {
...
+        } else {
+            path
+        };
```
kristoff3r, line 486: Or the code-golf version xD not sure which I prefer

```suggestion
        let path = self.path();
        let path = path.is_absolute().then(|| path.strip_prefix("/").ok()).flatten();
```

mockersf (changes requested):
The fact that the `AssetPath` is backed by a `Path` is pretty much an implementation detail, and shouldn't be relied on to check that the file exists.

to check if an asset exist, you should ask the `AssetSource` linked to it, which will ask its `AssetReader`, which is the only place accessing the filesystem (on platforms backed by the filesystem) actually happens

crates/bevy_asset/src/path.rs
mockersf, line 487: this is the root only for the default default asset source. it's possible to override the default asset source with something in another directory (and actually pretty much recommended if you use any kind of installer with standard path for an OS) and this would not match anymore

RyanSpaker:
`AssetPath::exists` only works on a small subset of `AssetPath`s, and the functionality can be easily replicated manually when needed, so I don't think it makes any sense to have this method, or `get_root_relative_path` for the same reasoning.

I think a better plan is to create a method on `AssetServer` to do this. Since `AssetServer` can actually access the Asset Sources, it would work on any `AssetPath`.

I'm closing this PR since I don't think it makes any sense.

PR #15643, opened 2024-10-04
Per-meshlet compressed vertex data

JMS55:
# Objective
- Prepare for streaming by storing vertex data per-meshlet, rather than per-mesh (this means duplicating vertices per-meshlet)
- Compress vertex data to reduce the cost of this

## Solution
The important parts are in from_mesh.rs, the changes to the Meshlet type in asset.rs, and the changes in meshlet_bindings.wgsl. Everything else is pretty secondary/boilerplate/straightforward changes.

- Positions are quantized in centimeters with a user-provided power of 2 factor (ideally auto-determined, but that's a TODO for the future), encoded as an offset relative to the minimum value within the meshlet, and then stored as a packed list of bits using the minimum number of bits needed for each vertex position channel for that meshlet
  - E.g. quantize positions (lossly, throws away precision that's not needed leading to using less bits in the bitstream encoding)
  - Get the min/max quantized value of each X/Y/Z channel of the quantized positions within a meshlet
  - Encode values relative to the min value of the meshlet. E.g. convert from [min, max] to [0, max - min]
  - The new max value in the meshlet is (max - min), which only takes N bits, so we only need N bits to store each channel within the meshlet (lossless)
  - We can store the min value and that it takes N bits per channel in the meshlet metadata, and reconstruct the position from the bitstream
- Normals are octahedral encoded and than snorm2x16 packed and stored as a single u32. 
  - Would be better to implement the precise variant of octhedral encoding for extra precision (no extra decode cost), but decided to keep it simple for now and leave that as a followup
  - Tried doing a quantizing and bitstream encoding scheme like I did for positions, but struggled to get it smaller. Decided to go with this for simplicity for now
- UVs are uncompressed and take a full 64bits per vertex which is expensive
  - In the future this should be improved
- Tangents, as of the previous PR, are not explicitly stored and are instead derived from screen space gradients
- While I'm here, split up MeshletMeshSaverLoader into two separate types

Other future changes include implementing a smaller encoding of triangle data (3 u8 indices = 24 bits per triangle currently), and more disk-oriented compression schemes.

References:
* "A Deep Dive into UE5's Nanite Virtualized Geometry" https://advances.realtimerendering.com/s2021/Karis_Nanite_SIGGRAPH_Advances_2021_final.pdf#page=128 (also available on youtube)
* "Towards Practical Meshlet Compression" https://arxiv.org/pdf/2404.06359
* "Vertex quantization in Omniforce Game Engine" https://daniilvinn.github.io/2024/05/04/omniforce-vertex-quantization.html

## Testing

  - Converted the stanford bunny, and rendered it with a debug material showing normals, and confirmed that it's identical to what's on main. EDIT: See additional testing in the comments below.
  - Could use some more size comparisons on various meshes, and testing different quantization factors. Not sure if 4 is a good default. EDIT: See additional testing in the comments below.
  - Also did not test runtime performance of the shaders. EDIT: See additional testing in the comments below.
  - Use my unholy script, replacing the meshlet example https://paste.rs/7xQHk.rs (must make MeshletMesh fields pub instead of pub crate, must add lz4_flex as a dev-dependency) (must compile with meshlet and meshlet_processor features, mesh must have only positions, normals, and UVs, no vertex colors or tangents)

---

## Migration Guide
- TBD by JMS55 at the end of the release


JMS55:
Before merging, we need to commit a new bunny to https://github.com/JMS55/bevy_meshlet_asset and update the commit hashes in this repo

atlv24 (approved):
looks good, i presume that delta encoding for positions normals and etc and all the triangle strip nonsense will come later as a stream unpacking step to keep it decompressed in vram?

JMS55:
Not sure what you mean by delta encoding. Triangle strip compression maybe, maybe not. Nanite's method they found too slow for in-memory access, but I haven't evaluated the other paper I linked yet. We'll see when I get to it.

JMS55:
Main (54.72mb):
image
image

This PR (52.35mb):
image
image

JMS55:
Perf seems pretty much the same.

```rust
for x in -5..=5 {
    for y in -5..5 {
        for z in -5..5 {
            commands.spawn(MaterialMeshletMeshBundle {
                meshlet_mesh: meshlet_mesh_handle.clone(),
                material: debug_material.clone(),
                transform: Transform::from_translation(Vec3::new(
                    x as f32 * 15.0,
                    y as f32 * 5.0,
                    z as f32 * 5.0,
                )),
                ..default()
            });
        }
    }
}
```

image

image

JMS55:
Tldr; pretty much identical quality, asset size, and perf, but there's room to improve asset size in the future and we now unlock streaming.

atlv24 (approved):

JMS55:
Actually I should also check size in memory, and not just disk which also has LZ4 applied. Not a blocker, but would be nice to know. I'll check when I get a chance.

JMS55:
Main: 109.972084 mb
This PR: 63.614636 mb

Big win on memory usage! (previous comparisons above was LZ4-compressed disk size for the asset)

IceSentry (approved):
Doc is good, I didn't find anything breaking and the example runs fine.

LGTM

PR #15644, opened 2024-10-04
Reduce compile time of bevy_ptr::OwnedPtr::make function

clarfonthey:
## Methodology

A good metric that correlates with compile time is the amount of code generated by the compiler itself; even if the end binary is exactly the same size, having more copies of the same code can really slow down compile time, since it has to figure out whether it needs to include them or not.

The measurement for this used was the [`cargo-llvm-lines` crate](https://docs.rs/crate/cargo-llvm-lines) which can measure which functions are generating the most lines of LLVM IR, which generally means more code compiled. The example compiled was the `breakout` game, to choose something that touches a decent portion of the engine.

## Solution

Based upon the measurements, `bevy_ptr::OwnedPtr::make` was taking up 4061 lines of LLVM IR in the example code. So, I separated part of this function into a less-monomorphised version to reduce the amount of generated code. This was by far the most lines emitted by any single function.

## Results

After this change, only 2560 lines are emitted, accounting for a 36% decrease. I tried timing the results and it seemed like it did decrease compile times a bit, but honestly, the data is really noisy and I can't be bothered to compile bevy for hours on end to get enough data points.

The tweak feels like an improvement, so, I'll offer it, however small.

### DIFF
--- a/crates/bevy_ptr/src/lib.rs
+++ b/crates/bevy_ptr/src/lib.rs
@@ -397,13 +397,23 @@ impl<'a, T: ?Sized> From<&'a mut T> for PtrMut<'a> {
 }
 
 impl<'a> OwningPtr<'a> {
+    /// This exists mostly to reduce compile times;
+    /// code is only duplicated per type, rather than per function called.
+    ///
+    /// # Safety
+    ///
+    /// Safety constraints of [`PtrMut::promote`] must be upheld.
+    unsafe fn make_internal<T>(temp: &mut ManuallyDrop<T>) -> OwningPtr<'_> {
+        // SAFETY: The constraints of `promote` are upheld by caller.
+        unsafe { PtrMut::from(&mut *temp).promote() }
+    }
+
     /// Consumes a value and creates an [`OwningPtr`] to it while ensuring a double drop does not happen.
     #[inline]
     pub fn make<T, F: FnOnce(OwningPtr<'_>) -> R, R>(val: T, f: F) -> R {
-        let mut temp = ManuallyDrop::new(val);
         // SAFETY: The value behind the pointer will not get dropped or observed later,
         // so it's safe to promote it to an owning pointer.
-        f(unsafe { PtrMut::from(&mut *temp).promote() })
+        f(unsafe { Self::make_internal(&mut ManuallyDrop::new(val)) })
     }
 }
 


bushrat011899 (approved):
It can be quite hard to measure these smaller changes, but I agree with the technique, and it's trivial for the compiler to inline the new function if that produces better results.

crates/bevy_ptr/src/lib.rs
```diff
@@ -397,13 +397,20 @@ impl<'a, T: ?Sized> From<&'a mut T> for PtrMut<'a> {
...
+        // so it's safe to promote it to an owning pointer.
+        unsafe { PtrMut::from(&mut *temp).promote() }
+    }
+
     /// Consumes a value and creates an [`OwningPtr`] to it while ensuring a double drop does not happen.
     #[inline]
     pub fn make<T, F: FnOnce(OwningPtr<'_>) -> R, R>(val: T, f: F) -> R {
-        let mut temp = ManuallyDrop::new(val);
         // SAFETY: The value behind the pointer will not get dropped or observed later,
         // so it's safe to promote it to an owning pointer.
```
MrGVSV, line 412: This can probably be removed now, right? Since the unsafe code was moved into `make_internal`?

crates/bevy_ptr/src/lib.rs
```diff
@@ -397,13 +397,20 @@ impl<'a, T: ?Sized> From<&'a mut T> for PtrMut<'a> {
 }
 
 impl<'a> OwningPtr<'a> {
+    /// This exists mostly to reduce compile times;
+    /// code is only duplicated per type, rather than per function called.
+    fn make_internal<T>(temp: &mut ManuallyDrop<T>) -> OwningPtr<'_> {
```
hymm, line 402: This should be an unsafe method and have the same safety requirements as the promote method.

crates/bevy_ptr/src/lib.rs
clarfonthey: Yes. I tried multiple ways to rearrange the code and forgot to remove the comment on the latest one.

crates/bevy_ptr/src/lib.rs
```diff
@@ -397,13 +397,20 @@ impl<'a, T: ?Sized> From<&'a mut T> for PtrMut<'a> {
...
+        // SAFETY: The value behind the pointer will not get dropped or observed later,
```
hymm, line 403:
```suggestion
        // SAFETY: Safety is unheld by the caller.
```

The following line should be deleted.

crates/bevy_ptr/src/lib.rs
hymm: make_internal should be an unsafe method. The safety is still unheld in this method.

crates/bevy_ptr/src/lib.rs
MrGVSV: I get making the function itself unsafe, but what value does this safety comment add? If safety is unheld, shouldn't we not be using unsafe code?

hymm: Oops typo. It should be upheld

MrGVSV: Oh okay that makes sense haha

clarfonthey: The caller lets go of all safety constraints and becomes free.

clarfonthey:
I fixed `make_internal` not being marked as unsafe (oops) and hopefully things should be okay now. Please let me know if this looks correct, since I don't really understand these types.

BenjaminBrienen (approved):

cart (approved):

PR #15649, opened 2024-10-04
Inverse bevy_render bevy_winit dependency and move cursor to bevy_winit

### LINKED ISSUE #15565: `bevy_render` should not depend on `bevy_winit`
## Problem

#14284 added a direct, non-optional dependency on `bevy_winit` from `bevy_render`.

This is messy for our crate structure, and makes it much harder to use `bevy_render` on exotic platforms.

It's also causing issues for contributors:

>  I'm having issues running tests locally on main. Trying tests like (e.g. cargo test --package bevy_ui --lib layout::ui_surface::tests) I'm getting The platform you're compiling for is not supported by winit

From @StrikeForceZero on [Discord](https://discord.com/channels/691052431525675048/692572690833473578/1290679083897851925).

## Proposed solution: revert the cursor icon PR.

Revert #14284 and reopen #14351.

Solves this problem very directly, at the expense of a valuable feature. Back to the drawing board, where *none* of the options seemed very good.

## Proposed solution: make the dependency optional by feature-flagging cursor icons

Still messy, but something that can be worked around. Doesn't lose us the nice feature.

## Proposed solution: bevy_cursors crate

Lift this code out into its *own* crate, which depends on `bevy_winit` and `bevy_render`. Probably the cleanest conceptually, but now we have yet another tiny crate to maintain.

KirmesBude:
# Objective

- `bevy_render` should not depend on `bevy_winit`
- Fixes #15565

## Solution

- `bevy_render` no longer depends on `bevy_winit`
- The following is behind the `custom_cursor` feature
  - Move custom cursor code from `bevy_render` to `bevy_winit` behind the `custom_cursor` feature
  - `bevy_winit` now depends on `bevy_render` (for `Image` and `TextureFormat`)
  - `bevy_winit` now depends on `bevy_asset` (for `Assets`, `Handle` and `AssetId`)
  - `bevy_winit` now depends on `bytemuck` (already in tree)
- Custom cursor code in `bevy_winit` reworked to use `AssetId` (other than that it is taken over 1:1)
- Rework `bevy_winit` custom cursor interface visibility now that the logic is all contained in `bevy_winit`

## Testing

- I ran the screenshot and window_settings examples
- Tested on linux wayland so far

---

## Migration Guide

`CursorIcon` and `CustomCursor`  previously provided by `bevy::render::view::cursor` is now available from `bevy::winit`.
A new feature `custom_cursor` enables this functionality (default feature).


JMS55:
I don't want to bevy_winit to unconditionally depend on bevy_render. That breaks anyone using bevy_winit, but with a custom non-wgpu renderer.

KirmesBude:
> I don't want to bevy_winit to unconditionally depend on bevy_render (and to a lesser extent bevy_asset). That breaks anyone using bevy_winit, but with a custom non-wgpu renderer.

That makes sense, yeah. With #15650 I should be able to remove bevy_render in favor of bevy_image.
bevy_asset will remain though, because we need to go from Handle<Image> to Image somehow.

JMS55:
Imo we should feature flag this and make bevy_render and bevy_assets optional dependencies flagged behind the feature. @alice-i-cecile thoughts?

alice-i-cecile:
Strongly agree on a feature flag here. On by default is preferred though.

KirmesBude:
Great point with the feature.
I have added a `custom_cursor` feature on the new bevy_winit dependencies. I would like some guidance on how to deal with that in the examples. Feels like it should not be a default feature, but I am not sure how to test it then. Any existing examples that deal with a non-default feature?

alice-i-cecile:
I think that this is fine to be on by default. Anyone who's sophisticated or picky enough to not pull in `bevy_render` will have default features off anyways.

KirmesBude:
Also `bevy_render` -> `bevy_image` does not seem worth it, because I still need TextureFormat. So I either need `bevy_render` or `wgpu` (wgpu-types?) anyways.

alice-i-cecile:
@atlv24 could you take a look at this PR since it relates to your rendering splitting work?

crates/bevy_winit/src/state.rs
```diff
@@ -134,23 +141,25 @@ impl<T: Event> WinitAppRunnerState<T> {
     }
 }
 
+#[cfg(feature = "custom_cursor")]
 /// Identifiers for custom cursors used in caching.
 #[derive(Debug, Clone, Hash, PartialEq, Eq)]
 pub enum CustomCursorCacheKey {
-    /// u64 is used instead of `AssetId`, because `bevy_asset` can't be imported here.
-    AssetIndex(u64),
-    /// u128 is used instead of `AssetId`, because `bevy_asset` can't be imported here.
```
atlv24, line 142: why couldnt bevy_asset be imported here before?

atlv24 (approved):
very cool thanks

alice-i-cecile (approved):

tychedelia (approved):
This makes sense with the feature flag, which we should have considered in the first place! Thanks.

crates/bevy_winit/src/state.rs
KirmesBude: The original custom cursor PR calls out a circular dependency if bevy_render depends on bevy_asset. Not really sure why/where, though. But now it works :D

PR #15650, opened 2024-10-04
Split off bevy_image from bevy_render

atlv24:
# Objective

- bevy_render is gargantuan

## Solution

- Split off bevy_image

## Testing

- Ran some examples


alice-i-cecile:
Adding this to the 0.15 milestone. I don't want this to wait around because of the merge conflict pain, and merging this just after a release will make point release fixes brutal.

crates/bevy_image/Cargo.toml
```diff
@@ -0,0 +1,61 @@
...
+] }
+bevy_utils = { path = "../bevy_utils", version = "0.15.0-dev" }
+
+# rendering
+image = { version = "0.25.2", default-features = false }
+
+# misc
+bitflags = { version = "2.3", features = ["serde"] }
+bytemuck = { version = "1.5" }
+wgpu = { version = "22", default-features = false }
```
JMS55, line 44: Do you need the full wgpu here, or will wgpu-types work?

JMS55 (approved):

IceSentry (approved):

mockersf:
I think I would prefer a more general crate than bevy_image, to be able to move other kinds of assets currently living in bevy_render like `Mesh`, rather than multiplying small crates like `bevy_image`, `bevy_mesh`, ...
maybe a `bevy_render_assets`?

alice-i-cecile (approved):

alice-i-cecile:
I'm neutral on the two approaches, but I'd like to decide relatively quickly. I do like that the current split means 2D games don't have to pull in 3D mesh formats, but that can be feature-flagged within this crate.

alice-i-cecile:
There was [more discussion on Discord](https://discord.com/channels/691052431525675048/692572690833473578/1291855507187040377), but we're moving forward with the `bevy_image` + `bevy_mesh` approach.

Issue #15459, opened 2024-09-26
Follow up on Retained Rendering


Trashtalk217:
Merging #15320 leaves behind a couple of gaps that can be easily filled in a follow-up PR(s).

### Implement `WorldQuery` for `RenderEntity` and `MainEntity`

When querying `&RenderEntity` currently from within the main world (or `Extract` systems in the render world), you constantly have to call `.id()` on every thing you query, because `RenderEntity` is a component that contains an entity. This indirection is annoying and makes the migration to 0.15 more cumbersome (more line changes), so we can implement a custom `WorldQuery` so that this happens automatically (think of querying `Entity`).

The same is true for `MainEntity`.

### Using required components

There are a couple of places, mainly in `bevy_render/src/extract_component.rs` where `SyncToRenderWorld` is automatically added to certain components. This is currently done with observers, but it's much better to do this with required components. This is being tackled in https://github.com/bevyengine/bevy/pull/15582.

### Deprecating `get_or_spawn`

`get_or_spawn` Is quite often used in extraction systems, given how things used to work. But with a retained rendering world, the entity should always already exist, so just `get` should suffice.

hymm:
We should probably also remove to in-engine uses of TemporaryRenderEntity if possible.

Trashtalk217:
> We should probably also remove the in-engine uses of TemporaryRenderEntity if possible.

Yes, definitely a good idea. Although I expect the solutions for that to be more specific to the rendering context in which they are used.

alice-i-cecile:
> Spawning a specific entity value is rarely the right choice. Most apps should favor [Commands::spawn](https://dev-docs.bevyengine.org/bevy/ecs/prelude/struct.Commands.html#method.spawn). This method should generally only be used for sharing entities across apps, and only when they have a scheme worked out to share an ID space (which doesn’t happen by default).

From `get_or_spawn`'s docs. `insert_or_spawn_batch` has the same warning. Since we're no longer using this internally, we should deprecate this pattern completely as a footgun.

JMS55:
I think we've covered everything but https://github.com/bevyengine/bevy/issues/15704 now?

alice-i-cecile:
> I have a strong motivation to use shared entity ids for multiple worlds

[...]

PR #15320, opened 2024-09-19
The Cooler 'Retain Rendering World'

Trashtalk217:
- Adopted from #14449
- Still fixes #12144.

## Migration Guide

The retained render world is a complex change: migrating might take one of a few different forms depending on the patterns you're using.

For every example, we specify in which world the code is run. Most of the changes affect render world code, so for the average Bevy user who's using Bevy's high-level rendering APIs, these changes are unlikely to affect your code.

### Spawning entities in the render world

Previously, if you spawned an entity with `world.spawn(...)`, `commands.spawn(...)` or some other method in the rendering world, it would be despawned at the end of each frame. In 0.15, this is no longer the case and so your old code could leak entities. This can be mitigated by either re-architecting your code to no longer continuously spawn entities (like you're used to in the main world), or by adding the `bevy_render::world_sync::TemporaryRenderEntity` component to the entity you're spawning. Entities tagged with `TemporaryRenderEntity` will be removed at the end of each frame (like before).

### Extract components with `ExtractComponentPlugin`

```
// main world
app.add_plugins(ExtractComponentPlugin::<ComponentToExtract>::default());
```

`ExtractComponentPlugin` has been changed to only work with synced entities. Entities are automatically synced if `ComponentToExtract` is added to them. However, entities are not "unsynced" if any given `ComponentToExtract` is removed, because an entity may have multiple components to extract. This would cause the other components to no longer get extracted because the entity is not synced. 

So be careful when only removing extracted components from entities in the render world, because it might leave an entity behind in the render world. The solution here is to avoid only removing extracted components and instead despawn the entire entity. 

### Manual extraction using `Extract<Query<(Entity, ...)>>`

```rust
// in render world, inspired by bevy_pbr/src/cluster/mod.rs
pub fn extract_clusters(
    mut commands: Commands,
    views: Extract<Query<(Entity, &Clusters, &Camera)>>,
) {
    for (entity, clusters, camera) in &views {
        // some code
        commands.get_or_spawn(entity).insert(...);
    }
}
```
One of the primary consequences of the retained rendering world is that there's no longer a one-to-one mapping from entity IDs in the main world to entity IDs in the render world. Unlike in Bevy 0.14, Entity 42 in the main world doesn't necessarily map to entity 42 in the render world. 

[...]

PR #15652, opened 2024-10-04
Deprecate `get_or_spawn`

Trashtalk217:
# Objective

After merging retained rendering world #15320, we now have a good way of creating a link between worlds (*HIYAA intensifies*). This means that `get_or_spawn` is no longer necessary for that function. Entity should be opaque as the warning above `get_or_spawn` says. This is also part of #15459.

I'm deprecating `get_or_spawn_batch` in a different PR in order to keep the PR small in size.  

## Solution

Deprecate `get_or_spawn` and replace it with `get_entity` in most contexts. If it's possible to query `&RenderEntity`, then the entity is synced and `render_entity.id()` is initialized in the render world.

## Migration Guide

If you are given an `Entity` and you want to do something with it, use `Commands.entity(...)` or `World.entity(...)`. If instead you want to spawn something use `Commands.spawn(...)` or `World.spawn(...)`. If you are not sure if an entity exists, you can always use `get_entity` and match on the `Option<...>` that is returned.


MrGVSV:
I have no clue how networking code works, so ignore this comment if it's completely irrelevant.

Does removing this affect networking crates at all? My assumption would be that they might desire to pass entities over the wire and sync them one-to-one using something like `get_or_spawn`. Or do they typically just use a map to correlate entities (e.g. `EntityMapper`)?

alice-i-cecile:
> Or do they typically just use a map to correlate entities (e.g. EntityMapper)?

This is the recommended pattern. Per @maniwani, since we don't have entity ID range reservation, relying on specific entity IDs is very fragile and ill-advised, because things like UI spawning will generating new entity IDs and clash with the networked entities.

NiseVoid (approved):
Changes look fine to me, a bit odd that the transform hierarchy and `prepare_cluster` ones don't have `expect` and all the other ones do, but not really blocking imo.

Trashtalk217:
> Changes look fine to me, a bit odd that the transform hierarchy and `prepare_cluster` ones don't have `expect` and all the other ones do, but not really blocking imo.

Those, I believe, don't use any of the sync machinery so I don't know what I would write in the expect clause.

Shatur:
> Or do they typically just use a map to correlate entities (e.g. EntityMapper)

That's what we do in bevy_replicon, so we won't be affected.

bushrat011899:
> > Or do they typically just use a map to correlate entities (e.g. EntityMapper)
> 
> That's what we do in bevy_replicon, so we won't be affected.

Can also confirm that's how `bevy_ggrs` works!

bushrat011899 (approved):

PR #15658, opened 2024-10-04
Deprecate `Events::oldest_id`

### LINKED ISSUE #15617: `Events::oldest_event_count` and `Events::oldest_id` don't make the difference to each other clear enough
I am not certain if this is a docs issue, but I found this by looking at the source, not by encountering some bug. That is why for now I focused on how the API appears while there are more issues.

The methods [`Events::oldest_event_count`](https://docs.rs/bevy/latest/bevy/ecs/event/struct.Events.html#method.oldest_event_count) and [`Events::oldest_id`](https://docs.rs/bevy/latest/bevy/ecs/event/struct.Events.html#method.oldest_id) seem to do the same thing when looked at the docs or the source.

Also the API itself is confusing. Instead of returning `usize` it should return an `Option<EventId<E>>` because no oldest event might be available currently if the sequences `events_b` and `events_a` are empty.

As a side note, the implementation of `oldest_event_count` looks like that:

```rs
/// Returns the index of the oldest event stored in the event buffer.
pub fn oldest_event_count(&self) -> usize {
    self.events_a
        .start_event_count
        .min(self.events_b.start_event_count)
}
```

When it should be clear that `self.events_b` as the more recent sequence cannot contain the oldest count.

- `oldest_event_count` was added by https://github.com/bevyengine/bevy/pull/5730
- `oldest_id` was added by https://github.com/bevyengine/bevy/pull/5735

urben1680:
# Objective

Fixes #15617 

## Solution

The original author confirmed it was not intentional that both these methods exist.

They do the same, one has the better implementation and the other the better name.

## Testing

I just ran the unit tests of the module.

---

## Migration Guide

- Change usages of `Events::oldest_id` to `Events::oldest_event_count`
- If `Events::oldest_id` was used to get the actual oldest `EventId::id`, note that the deprecated method never reliably did that in the first place as the buffers may contain no id currently.

### DIFF
--- a/crates/bevy_ecs/src/event/collections.rs
+++ b/crates/bevy_ecs/src/event/collections.rs
@@ -114,9 +114,7 @@ impl<E: Event> Default for Events<E> {
 impl<E: Event> Events<E> {
     /// Returns the index of the oldest event stored in the event buffer.
     pub fn oldest_event_count(&self) -> usize {
-        self.events_a
-            .start_event_count
-            .min(self.events_b.start_event_count)
+        self.events_a.start_event_count
     }
 
     /// "Sends" an `event` by writing it to the current event buffer.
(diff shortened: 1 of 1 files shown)


ItsDoot (approved):
Don't forget to update the PR title.

alice-i-cecile (approved):

PR #15662, opened 2024-10-05
Fix additive blending of quaternions

### LINKED ISSUE #13832: Additive blending of Quats looks suspect
I believe this was previously noticed by @djeedai in the review of the `Animatable` trait rollout.

I am not an expert in this subject, but here is my understanding.

## Context

In additive animation blending, a sample from an animation clip represents a relative displacement rather than an absolute quantity; this is applied on top of other animations by combining the displacement with the absolute quantity "underneath". Like non-additive animation blending (which interpolates between two absolute quantities instead), this can be used in concert with a weight which indicates the strength of the sample's influence in the blend. For additive blending, this means that, as the weight tends to zero, the displacement also attenuates; with a weight of zero, the blend should have no effect on the existing quantity.

For example, additive blending of translational components works something like this:
```rust
let (displacement, weight): (Vec3, f32) = incoming_sample;
translation += displacement * weight
```

The right-hand side of this is perhaps best conceptualized as:
```rust
Vec3::lerp(Vec3::ZERO, displacement, weight)
```

In other words, the additively blended sample is interpolated with zero and then composed with the existing value.

## The mismatch

Now, currently additive blending for Quat works like this:
```rust
let (rot_displacement, weight): (Quat, f32) = incoming_sample;
rotation = rotation.slerp(rot_displacement, weight);
```

What I would expect based on my understanding of additive blending is instead something like this:
```rust
let (rot_displacement, weight): (Quat, f32) = incoming_sample;
rotation = Quat::slerp(Quat::IDENTITY, rot_displacement, weight) * rotation;
```

Note that these are definitely not equivalent — when the incoming sample is `Quat::IDENTITY`, the latter never has any effect, while the former (with positive weight) will slerp the current rotation state towards the default orientation.

Of course, it's possible this was an intentional decision and I'm missing something; this is just my understanding of things.

mweatherley:
# Objective

Fixes #13832

## Solution

Additively blend quaternions like this:
```rust
rotation = Quat::slerp(Quat::IDENTITY, incoming_rotation, weight) * rotation;
```

## Testing

Ran `animation_masks`, which behaves the same as before. (In the context of an animation being blended only onto the base pose, there is no difference.)

We should create some examples that actually exercise more of the capabilities of the `AnimationGraph` so that issues like this can become more visible in general. (On the other hand, I'm quite certain this was wrong before.)

## Migration Guide

This PR changes the implementation of `Quat: Animatable`, which was not used internally by Bevy prior to this release version. If you relied on the old behavior of additive quaternion blending in manual applications, that code will have to be updated, as the old behavior was incorrect.

### DIFF
--- a/crates/bevy_animation/src/animatable.rs
+++ b/crates/bevy_animation/src/animatable.rs
@@ -149,7 +149,8 @@ impl Animatable for Transform {
             if input.additive {
                 translation += input.weight * Vec3A::from(input.value.translation);
                 scale += input.weight * Vec3A::from(input.value.scale);
-                rotation = rotation.slerp(input.value.rotation, input.weight);
+                rotation =
+                    Quat::slerp(Quat::IDENTITY, input.value.rotation, input.weight) * rotation;
             } else {
                 translation = Vec3A::interpolate(
                     &translation,
@@ -181,8 +182,17 @@ impl Animatable for Quat {
     #[inline]
     fn blend(inputs: impl Iterator<Item = BlendInput<Self>>) -> Self {
         let mut value = Self::IDENTITY;
-        for input in inputs {
-            value = Self::interpolate(&value, &input.value, input.weight);
+        for BlendInput {
+            weight,
+            value: incoming_value,
+            additive,
+        } in inputs
+        {
+            if additive {
+                value = Self::slerp(Self::IDENTITY, incoming_value, weight) * value;
+            } else {
+                value = Self::interpolate(&value, &incoming_value, weight);
+            }
         }
         value
     }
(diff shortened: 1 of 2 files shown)


IQuick143 (approved):
Not an animation expert, but this seems correct. (Are we sure about the order of the quaternion multiplication?)

mockersf (approved):

mweatherley:
> (Are we sure about the order of the quaternion multiplication?)

I'm pretty sure it's right; I think the last time I checked this was when we introduced `Isometry3d`.

atornity (approved):
multiplication order seems correct, according to answers on [stack exchange](https://math.stackexchange.com/questions/2124361/quaternions-multiplication-order-to-rotate-unrotate)

PR #15665, opened 2024-10-05
Unified `Transform` blend stack

mweatherley:
# Objective

Because curve-based animation (#15434) supports animating things based on arbitrary curves, it would be good to support `Curve<Transform>`. Unfortunately, the `TransformCurve` wrapper had to be killed in the migration to blend stacks (#15589, #15598) because `Transform` would fight with all of its parts over write access to the actual `Transform` component.

The goal of this PR is to reintroduce `TransformCurve` by making all the animated parts of `Transform` (`TranslationCurve`, `RotationCurve`, `ScaleCurve`) share the same blend stack. This allows `TransformCurve` to function by sharing the same blend stack with all of them, so they simply blend harmoniously instead of fighting over the same component data.

## Solution

A new `Animatable` type `TransformParts` has been introduced:
```rust
/// Basically `Transform`, but with every part optional. Note that this is the true
/// output of `Transform` animation, since individual parts may be unconstrained
/// if they lack an animation curve to control them. 
#[derive(Default, Debug, Clone, Reflect)]
pub(crate) struct TransformParts {
    translation: Option<Vec3>,
    rotation: Option<Quat>,
    scale: Option<Vec3>,
}
```

Subsequently, `TranslationCurve`, `RotationCurve`, `ScaleCurve`, and `TransformCurve` all share a single blend stack where the individual elements are `TransformParts`. For example, when `TranslationCurve` contributes an element to the blend stack, it sets the `rotation` and `scale` components to be `None`, and you can probably guess how everything else goes from there; the only wrinkle is that when multiple curves are added from the same animation clip, they append to the same element on top of the stack instead of each getting their own element — this is for compatibility with assumptions in graph evaluation.

An example `transform_curve_animation` has been added which shows a procedural animation defined by a `Curve<Transform>` with an animation defined by a `Curve<Vec3>` additively blended on top. 

## Testing

Tested on `many_foxes` and found little to no performance difference. Testing on `animation_graph` helps assure that nothing broke in the migration to the unified blend stack.

Screenshot 2024-10-07 at 8 18 47 AM

---

## Showcase

```rust
// Allocate an animation clip.
let mut clip = AnimationClip::default();

// This curve describes the position of the ship over time.
let wobbly_circle_curve =
    function_curve(Interval::new(0.0, std::f32::consts::TAU).unwrap(), |t| {
        vec3(t.sin() * 5.0, t.sin() * 1.5, t.cos() * 5.0)
    });

// This curve uses the position of the ship to make its inward wing point toward
// the center of the platform that we'll position.
let transform_curve = wobbly_circle_curve.map(|position| {
    Transform::from_translation(position).aligned_by(
        Dir3::NEG_X,
        vec3(0.0, -2.0, 0.0) - position,
        Dir3::Y,
        Dir3::Y,
    )
});

clip.add_curve_to_target(animation_target_id, TransformCurve(transform_curve));
```

https://github.com/user-attachments/assets/4181d705-3377-4323-9502-602836f4ac8d


mweatherley:
Closing because as of #16484 this would need to be completely reworked.

PR #15666, opened 2024-10-05
Split out bevy_mesh from bevy_render 

atlv24:
# Objective

- bevy_render is gargantuan

## Solution

- Split out bevy_mesh

## Testing

- Ran some examples, everything looks fine

## Migration Guide

`bevy_render::mesh::morph::inherit_weights` is now `bevy_render::mesh::inherit_weights`

if you were using `Mesh::compute_aabb`, you will need to `use bevy_render::mesh::MeshAabb;` now


BenjaminBrienen (approved):

crates/bevy_mesh/src/bounding.rs
```diff
@@ -13,21 +13,16 @@ use bevy_reflect::prelude::*;
 /// occupied by this entity, with faces orthogonal to its local axis.
 ///
 /// This component is notably used during "frustum culling", a process to determine
-/// if an entity should be rendered by a [`Camera`] if its bounding box intersects
```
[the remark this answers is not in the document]
atlv24: because Camera is not in the same crate anymore and thus cannot be linked in a doc comment

crates/bevy_mesh/Cargo.toml
```diff
@@ -0,0 +1,37 @@
+[package]
+name = "bevy_mesh"
+version = "0.15.0-dev"
+edition = "2021"
+description = "Provides image types for Bevy Engine"
```
Jondolf, line 5: I think this should be changed 😅
```suggestion
description = "Provides mesh types for Bevy Engine"
```

atlv24: lmaooooo good catch

crates/bevy_mesh/src/bounding.rs
```diff
@@ -13,21 +13,16 @@ use bevy_reflect::prelude::*;
...
-/// for example if the vertex positions of a [`Mesh3d`] are updated.
-///
-/// [`Camera`]: crate::camera::Camera
-/// [`NoFrustumCulling`]: crate::view::visibility::NoFrustumCulling
-/// [`CalculateBounds`]: crate::view::visibility::VisibilitySystems::CalculateBounds
-/// [`Mesh3d`]: crate::mesh::Mesh
+/// for example if the vertex positions of a `Mesh3d` are updated.
 #[derive(Component, Clone, Copy, Debug, Default, Reflect, PartialEq)]
 #[reflect(Component, Default, Debug, PartialEq)]
 pub struct Aabb {
```
Jondolf, line 28: All of these bounding primitives should be in bevy_render, not bevy_mesh, since they're a rendering implementation detail that meshes shouldn't need to really care about. It'd also make these docs referencing cameras and various other components make sense.

`Mesh::compute_aabb` is the only method I see here that uses `Aabb`, but that could be handled by an extension trait or just a function in `bevy_render`. Is there some other reason why these types are currently here?

atlv24: i felt they should eventually end up in bevy_math, so i popped them out of bevy_render to begin

crates/bevy_mesh/src/index.rs
```diff
@@ -0,0 +1,137 @@
+use bevy_reflect::Reflect;
+use core::iter::FusedIterator;
+use thiserror::Error;
+use wgpu::IndexFormat;
+
+/// A disjunction of four iterators. This is necessary to have a well-formed type for the output
+/// of [`super::Mesh::triangles`], which produces iterators of four different types depending on the
```
Jondolf, line 7: Would be nice if these doc links were changed not to display `super::` everywhere, like they were before. Either by specifying the link separately, or if there are *tons* of these links, just importing the relevant types at module-level and adding an `#[allow(unused_imports)]` (not sure how acceptable this is generally lol)

```suggestion
/// of [`Mesh::triangles`](super::Mesh::triangles), which produces iterators of four different types depending on the
```

Jondolf (approved):
Didn't go over this line-by-line to verify that absolutely everything looks correct, but I like the overall direction and module structure improvements. Would be nice to be able to put assets, morph weights, and skinning behind feature flags, and remove the dependence on bevy_ecs, but that's trickier and maybe controversial, so it's probably fine to attempt in follow-ups.

bushrat011899 (approved):
Confirming that this looks good, LGTM!

alice-i-cecile (approved):

Issue #15659, opened 2024-10-05
Bounds of text with mixed-size fonts is calculated incorrectly


alice-i-cecile:
Looks like bounds are calculated using the `cosmic-text` `Metrics`, which is set using the largest font size in a text block.

_Originally posted by @UkoeHB in https://github.com/bevyengine/bevy/issues/15622#issuecomment-2394752727_

alice-i-cecile:
I'm pretty sure the fix here requires changes upstream in cosmic-text. Link this issue in any issues / PRs created there.

PR #15669, opened 2024-10-05
Fix text measurement when multiple font sizes are present

UkoeHB:
# Objective

- Fixes https://github.com/bevyengine/bevy/issues/15659

## Solution

- Add up line heights to get text block height instead of using `Metrics`, which only records the largest line height.

## Testing

- [x] Fixed issue shown in https://github.com/bevyengine/bevy/pull/15622

### DIFF
--- a/crates/bevy_text/src/pipeline.rs
+++ b/crates/bevy_text/src/pipeline.rs
@@ -455,15 +455,13 @@ fn get_attrs<'a>(
 
 /// Calculate the size of the text area for the given buffer.
 fn buffer_dimensions(buffer: &Buffer) -> Vec2 {
-    let width = buffer
+    let (width, height) = buffer
         .layout_runs()
-        .map(|run| run.line_w)
-        .reduce(f32::max)
-        .unwrap_or(0.0);
-    let line_height = buffer.metrics().line_height.ceil();
-    let height = buffer.layout_runs().count() as f32 * line_height;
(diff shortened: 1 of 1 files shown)


alice-i-cecile (approved):

viridia:
This looks great, however I have one minor comment which I am not certain applies for Bevy.

Usually when fonts are laid out on a line, they are aligned on the font baseline, and each font has an ascent-height and a descent height. So the actual line height is calculated as max(all_ascents) + max(all_descents). So for example, if one font has a very large ascent height, and another font has a very larger descent height, you might end up with a line-height that is taller than either font is individually.

viridia (approved):

UkoeHB:
@viridia [These](https://docs.rs/cosmic-text/0.12.1/cosmic_text/struct.LayoutRun.html) are the docs for the layout info provided by `cosmic-text`. Do you think we need to incorporate `line_y` and `line_top`? The `line_height` claims to be distance to next line so my assumption is it encompasses all complexities.

viridia:
In any case, this PR is "strictly no worse" than what came before, so I'd commit it regardless.

ickshonpe (commented):
Looks good.

ickshonpe:
> If we had some fonts with large ascent and descent then it would be easy to test it.

I constructed a few naive examples like this and there didn't seem to be any problems.

PR #15674, opened 2024-10-06
Add the functions `start_drag_move` and `start_drag_resize` to `Window`

IsseW:
# Objective
Expose the `winit` functions [drag_window](https://docs.rs/winit/latest/winit/window/struct.Window.html#method.drag_window) and [resize_window](https://docs.rs/winit/latest/winit/window/struct.Window.html#method.drag_resize_window).

Which allows implementing move & resize for windows without decorations.

## Solution

Add the functions `start_drag_move` and `start_drag_resize` to `bevy_window::Window`, which are then assigned to fields in `InternalWindowState`, and propagated to `winit` in the `changed_windows` system.

## Testing

I've tested that both functions works on x11 and wayland. Not sure if someone needs to test on windows/mac?

---

## Showcase

[Screencast from 2024-10-06 11-49-58 (trimmed).webm](https://github.com/user-attachments/assets/1cdee7b1-22bd-41d3-8a0a-6872a6ebf62c)

(The flickering in the video is some issue with resizing without decorations on x11)

<details>

Not the same code used in the video, but simple way to test moving a window without decorations.
```rust
use bevy::prelude::*;

fn main() {
    App::new()
        .add_plugins(DefaultPlugins.set(WindowPlugin {
            primary_window: Some(Window {
                decorations: false,
                ..default()
            }),
            ..default()
        }))
        .add_systems(Update, move_windows)
        .run();
}

fn move_windows(mut windows: Query<&mut Window>, input: Res<ButtonInput<MouseButton>>) {
    if input.pressed(MouseButton::Left) {
        for mut window in windows.iter_mut() {
            window.start_drag_move();
        }
    }
}
```

</details>


alice-i-cecile (approved):

LiamGallagher737 (approved):
Tested on Linux with i3wm, for some reason letting go of the mouse does not stop the dragging. However this is because `input.pressed(MouseButton::Left)` is returning true even after I let go so I won't block as the issue seems unrelated.
Windows 11 works correctly.

alice-i-cecile:
Yeah, I'd like an example here for ease of testing :)

crates/bevy_winit/src/converters.rs
```diff
@@ -8,7 +8,7 @@ use bevy_input::{
 use bevy_math::Vec2;
 #[cfg(feature = "custom_cursor")]
 use bevy_window::SystemCursorIcon;
-use bevy_window::{EnabledButtons, WindowLevel, WindowTheme};
+use bevy_window::{EnabledButtons, ResizeDirection, SystemCursorIcon, WindowLevel, WindowTheme};
```
alice-i-cecile, line 11:
```suggestion
use bevy_window::{EnabledButtons, ResizeDirection, WindowLevel, WindowTheme};
```

shanecelis:
I'm on x86 macOS, and I was able to provoke a panic when using the above code after I changed it to look for `MouseButton::Right`. I know the docs suggest it should only work with `MouseButton::Left`.

```
thread 'main' panicked at /Users/shane/.cargo/registry/src/index.crates.io-6f17d22bba15001f/winit-0.30.5/src/platform_impl/macos/window_delegate.rs:1163:74:
called `Option::unwrap()` on a `None` value
Encountered a panic in system `bevy_winit::system::changed_windows`!
```

The code from winit that panicked looks like this:

```rust
    #[inline]
    pub fn drag_window(&self) -> Result<(), ExternalError> {
        let mtm = MainThreadMarker::from(self);
        let event = NSApplication::sharedApplication(mtm).currentEvent().unwrap();
        self.window().performWindowDragWithEvent(&event);
        Ok(())
    }
```

Perhaps these functions should be marked unsafe to encourage some caution.

I'll try to submit an example for this to issue #15734.

mockersf:
> Perhaps these functions should be marked unsafe to encourage some caution.
> 

Strictly speaking, if it panics it's not unsafe. Unsafe is for code that doesn't panic if its conditions are not met.

shanecelis:
I hear that and agree it's not undefined behavior. I'd counter with this from [std's docs on the unsafe keyword](https://doc.rust-lang.org/std/keyword.unsafe.html) 

> unsafe fn: calling this function means abiding by a contract the compiler cannot enforce.

I would submit that requiring a left mouse button event be pressed before calling these functions is such a contract.

extrawurst:
Why does this unwrap if it already returns a `Result` 🙈

shanecelis:
Who wants to submit an `.ok_or([ExternalError](https://docs.rs/winit/latest/winit/error/enum.ExternalError.html)::Ignored)?` PR to winit?

shanecelis:
As I'm making the example, I'm learning new things. The start drag resize isn't supported on macOS.

```
2024-10-09T09:56:14.055238Z  WARN bevy_winit::system: Winit returned an error while attempting to drag resize the window: the requested operation is not supported by Winit
```

This is fine. Good error message to my eyes.

PR #14788, opened 2024-08-16
add curve utilities to create curves interpolating/easing between two values

RobWalt:
# Objective

Citing @mweatherley 

> There is a lot of shortfall for simple cases— e.g., we should have library functions for making a curve connecting two points, eased versions of that, and so on.

## Solution

This PR implements

- a simple `Easing` trait which is implemented for all `impl Curve<f32>` types. We can't really guarantee that these curves have unit interval domain, which some people would probably expect, but it is documented that this isn't the case for these types and we redirect to `EasingCurve` which is used for that purpose
- an `EasingCurve` struct, which is used to interpolate between two values `start` and `end` using a `impl Easing` curve where the curve will be guaranteed to be reparametrized
- a `LinearCurve` which linearly interpolates between two values `start` and `end`
- a `CubicBezierCurve` which interpolates between `start` and `end` values using a `CubicSegment`
- a `StepCurve` which interpolates between `start` and `end` with an step-function with `n` steps
- an `ElasticCurve` which interpolates between `start` and `end` with spring like behavior where the elasticity of the spring is configurable
- some `FunctionCurve` easing curves for different popular functions including: `quadratic_ease_in`, `quadratic_ease_out`, `smoothstep`, `identity`

## Testing

- there are a few new tests for all of these in the main module

PR #15675, opened 2024-10-06
Add most common interpolations

mockersf:
# Objective

- Followup for #14788 
- Support most usual ease function

## Solution

- Use the crate [`interpolation`](https://docs.rs/interpolation/0.3.0/interpolation/trait.Ease.html) which has them all
- it's already used by bevy_easings, bevy_tweening, be_tween, bevy_tweening_captured, bevy_enoki, kayak_ui in the Bevy ecosystem for various easing/tweening/interpolation

### DIFF
--- a/crates/bevy_math/src/curve/easing.rs
+++ b/crates/bevy_math/src/curve/easing.rs
@@ -1,6 +1,8 @@
 //! Module containing different [`Easing`] curves to control the transition between two values and
 //! the [`EasingCurve`] struct to make use of them.
 
+use interpolation::{Ease, EaseFunction};
+
 use crate::{
     ops::{self, FloatPow},
     VectorSpace,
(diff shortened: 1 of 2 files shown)


NthTensor (changes requested):
I think we actually want to `impl Easing<f32>` for the easing functions?

alice-i-cecile (approved):

mockersf:
> I think we actually want to `impl Easing<f32>` for the easing functions? `Easing` is intended for general easing methods that interp between `(0,0)` and `(0,1)`. `EasingCurve` is parameterized by two values and one of the eazing methods.

trying with
```rust
impl Curve<f32> for EaseFunction {
    fn domain(&self) -> Interval {
        Interval::UNIT
    }

    fn sample_unchecked(&self, t: f32) -> f32 {
        t.calc(*self)
    }
}
```

fails with
```
conflicting implementations of trait `Curve<f32>` for type `EaseFunction`
upstream crates may add a new impl of trait `std::ops::Deref` for type `interpolation::EaseFunction` in future versions
```

because of the blanket impl
```rust
impl<T, C, D> Curve<T> for D
where
    C: Curve<T> + ?Sized,
    D: Deref<Target = C>,
{
    fn domain(&self) -> Interval {
        <C as Curve<T>>::domain(self)
    }

    fn sample_unchecked(&self, t: f32) -> T {
        <C as Curve<T>>::sample_unchecked(self, t)
    }
}
```

I copied the whole enum, that way I can add doc comments which the original crate is lacking of

@NthTensor is that better?

mweatherley:
Hmm... I think the thing most in line with the current philosophy would be something similar to`EasingCurve::new` that takes the desired start and end values (`T: VectorSpace`) together with an `EaseFunction` and then does the same thing as `new` after mapping the variants to their actual functions. The nice thing about how things are set up with `new` is that:
1. Targeting the start and end values makes explicit reparametrizations less necessary; if you want a cubically eased line segment between two `Vec3`s, you can just call it with those start and end values.
2. Since easing curves valued in `f32` are something you often reparametrize by, it's convenient even in the `f32` case to allow `start` and `end` to be specified as something other than `0` and `1` (e.g. the `start` and `end` of a given curve's domain interval), so that you can avoid an intermediate linear reparametrization.

NthTensor:
This api feels a tiny bit over designed. Could easing just be a function? I'm not sure using the full power of the curve trait is necessary.

crates/bevy_math/src/curve/easing.rs
```diff
@@ -296,3 +297,121 @@ impl ElasticCurve {
         Self { omega }
     }
 }
+
+/// Curve functions over the [unit interval], commonly used for easing transitions.
+///
+/// [unit interval]: `Interval::UNIT`
+#[derive(Debug, Copy, Clone, PartialEq)]
```
alice-i-cecile, line 304: I would really like this to be Reflect, so you can twiddle this in inspectors.

mweatherley:
> This api feels a tiny bit over designed. Could `Easing` just be a function? I'm not sure using the full power of the curve trait is necessary.

I don't really care about the `Easing` trait, but the high-level goal here is basically just to make it very easy to create a curve that links two things. If I recall correctly, I originally pitched this as an API that was basically like this:
```rust
let my_curve = curve_between(start_value, end_value, easing_fn);
```
The point was that it would produce a curve parametrized over the unit interval that started at `start_value` and ended at `end_value`, with how long it lingered at each end or whatever determined by `easing_fn`. We basically have that in `EasingCurve::new`, although I might still prefer if the API were just a library function.

From that angle, the wrinkle here is that `EaseFunction` provides a bunch of good values of `easing_fn` which we want to expose in a natural way, which is why I'm basically inclined towards something which still looks like `curve_between` but takes the canned enum as input instead of `easing_fn`.

NthTensor:
> Actually, if that were even the entire API, I would probably be fine with that, too.

Yep.

tychedelia:
> (EDIT: One other nice thing about using a canned enum is that, if the struct stores the enum value internally, it behaves much more nicely with respect to `Reflect` than an arbitrary `Fn` thing would, which is to say you'd actually be able to play with it in an editor or something.)

This is really important to support. My users will LOVE this feature, but it needs to be discoverable in the way that an enum provides. You can always add an enum variant that contains a custom fn too, if needed.

crates/bevy_math/src/curve/easing.rs
mockersf: added serialisation while I'm at it!

bushrat011899 (approved):
Nice to have these built-in!

alice-i-cecile:
Thank you to everyone involved with the authoring or reviewing of this PR! This work is relatively important and needs release notes! Head over to https://github.com/bevyengine/bevy-website/issues/1719 if you'd like to help out.

PR #15538, opened 2024-09-30
Add Support for Triggering Events via `AnimationEvent`s

atornity:
# Objective

Add support for events that can be triggered from animation clips. This is useful when you need something to happen at a specific time in an animation. For example, playing a sound every time a characters feet hits the ground when walking.

Closes #15494 

## Solution

Added a new field to `AnimationClip`: `events`, which contains a list of `AnimationEvent`s. These are automatically triggered in `animate_targets` and `trigger_untargeted_animation_events`. 

## Testing

Added a couple of tests and example (`animation_events.rs`) to make sure events are triggered when expected.

---

## Showcase

`Events` need to also implement `AnimationEvent` and `Reflect` to be used with animations.

```rust
#[derive(Event, AnimationEvent, Reflect)]
struct SomeEvent;
```

Events can be added to an `AnimationClip` by specifying a time and event.

```rust
// trigger an event after 1.0 second
animation_clip.add_event(1.0, SomeEvent);
```

And optionally, providing a target id.

```rust
let id = AnimationTargetId::from_iter(["shoulder", "arm", "hand"]);
animation_clip.add_event_to_target(id, 1.0, HandEvent);
```

I modified the `animated_fox` example to show off the feature.

CleanShot 2024-10-05 at 02 41 57

PR #15677, opened 2024-10-06
Don't trigger animation events when paused

atornity:
# Objective

Pausing the `animated_fox` example perfectly as one of the feet hits the ground causes the event to be triggered every frame. 

Context: #15538

## Solution

Don't trigger animation events if the animation is paused.

## Testing

Ran the example, I no longer see the issue.

### DIFF
--- a/crates/bevy_animation/src/lib.rs
+++ b/crates/bevy_animation/src/lib.rs
@@ -1144,7 +1144,9 @@ pub fn animate_targets(
                         };
 
                         // Trigger all animation events that occurred this tick, if any.
-                        if let Some(triggered_events) = TriggeredEvents::from_animation(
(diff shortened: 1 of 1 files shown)


alice-i-cecile (approved):
I think this would be clearer with an && !paused style, but I won't block on that.

atornity:
> I think this would be clearer with an && !paused style, but I won't block on that.

That's unstable for if `let` expressions. I could nest the `if let` inside a `if paused` if that's preferable.

alice-i-cecile:
Curse you unstable Rust features. Mild preference for nesting.

mweatherley (approved):

mockersf (approved):

PR #15678, opened 2024-10-06
Read back full screen state from window on macOS

coreh:
# Objective

I'm trying to display on a menu screen of my game the Windowed/Full screen state. However, on macOS it's possible for the user to trigger borderless fullscreen directly from the window controls, which results in a stale/wrong state displayed.

## Solution

- Add logic to read back full screen state and apply it to the window when the user directly makes it full screen via the window controls on macOS

## Testing

- Did you test these changes? If so, how? Yes, mostly on my game
- Are there any parts that need more testing? I'm not sure if this has performance implications, as the call might be blocking on the UI thread. However, unscientifically I didn't notice any degradation.
- How can other people (reviewers) test your changes? Is there anything specific they need to know? Adding log statements to the match branches should suffice.
- If relevant, what platforms did you test these changes on, and are there any important ones you can't test? Just macOS

### DIFF
--- a/crates/bevy_winit/src/system.rs
+++ b/crates/bevy_winit/src/system.rs
@@ -9,10 +9,12 @@ use bevy_ecs::{
 use bevy_input::keyboard::KeyboardFocusLost;
 use bevy_utils::tracing::{error, info, warn};
 use bevy_window::{
-    ClosingWindow, Monitor, MonitorSelection, PrimaryMonitor, RawHandleWrapper, VideoMode, Window,
-    WindowClosed, WindowClosing, WindowCreated, WindowFocused, WindowMode, WindowResized,
-    WindowWrapper,
+    ClosingWindow, Monitor, PrimaryMonitor, RawHandleWrapper, VideoMode, Window, WindowClosed,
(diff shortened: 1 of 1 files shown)


mockersf:
does winit sends a resolution changed event when the user changes the window controls? I think it would be better to add this when the window reacts to a corresponding event from winit

crates/bevy_winit/src/system.rs
```diff
@@ -327,6 +330,17 @@ pub(crate) fn changed_windows(
                     winit_window.set_fullscreen(new_mode);
                 }
             }
+        } else {
+            #[cfg(target_os = "macos")]
+            match (winit_window.fullscreen(), window.mode) {
+                (Some(winit::window::Fullscreen::Borderless(_)), WindowMode::Windowed) => {
+                    window.mode = WindowMode::BorderlessFullscreen(MonitorSelection::Current);
```
tychedelia, line 337: We should probably make sure we are copying the correct monitor returned from `winit` rather than just defaulting to `Current`.

alice-i-cecile:
Should we be upstreaming this fix to `winit` instead?

PR #15680, opened 2024-10-06
simplify adding headers and improve consistency for RemoteHttpPlugin

spacemen0:
# Objective

- a follow up pr for #15651 

## Solution

- rename add to insert cuz insert make more sense when using a HashMap and new value overrides previous value based on key (quote: https://github.com/bevyengine/bevy/pull/15651#discussion_r1788851778)
- use `TryInto<>` for add(insert) as well when constructing `Headers`. Doing so user won't need to interact with hyper APIs, and `with_headers` will align better with `with_header` (quote: https://github.com/bevyengine/bevy/pull/15651#discussion_r1788687251)
- move example usage of Headers to `with_headers` method (quote: https://github.com/bevyengine/bevy/pull/15651#discussion_r1788989500) 

## Testing

- the same as I tested my previous pr

### DIFF
--- a/crates/bevy_remote/src/http.rs
+++ b/crates/bevy_remote/src/http.rs
@@ -17,7 +17,7 @@ use bevy_ecs::system::{Res, Resource};
 use bevy_tasks::IoTaskPool;
 use core::net::{IpAddr, Ipv4Addr};
 use http_body_util::{BodyExt as _, Full};
-pub use hyper::header::{HeaderName, HeaderValue};
+use hyper::header::{HeaderName, HeaderValue};
 use hyper::{
     body::{Bytes, Incoming},
     server::conn::http1,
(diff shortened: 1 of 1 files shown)


crates/bevy_remote/src/http.rs
```diff
@@ -137,6 +129,26 @@ impl RemoteHttpPlugin {
         self
     }
     /// Set the extra headers that the response will include.
+    ///
+    /// ////// /// # Example
```
alice-i-cecile, line 133:
```suggestion
    /// # Example
```

alice-i-cecile (approved):

alice-i-cecile:
@LiamGallagher737 @villor reviews here please :)

mockersf (approved):

PR #15681, opened 2024-10-06
improve sub view example with dynamic viewports

### LINKED ISSUE #15576: macOS: camera with shear not working
> This is still broken on macOS and will minimally need a quick follow up for that.

_Originally posted by @tychedelia in https://github.com/bevyengine/bevy/pull/15537#issuecomment-2386389530_

m-edlund:
# Objective

- Adds better comments and includes an example where the aspect ratio between `size` and `full_size` differ
- Fixes #15576

## Solution

- Viewports are dynamically scaled to window size

## Testing

- Tested with moving the window around and by manually setting the window scaling factor
- Just to make sure nothing else is going on, someone on macOS should also test this 

## Showcase

Since calculating padding from window size is a hassle, the example now looks a bit more squished together:

image

### DIFF
--- a/examples/3d/camera_sub_view.rs
+++ b/examples/3d/camera_sub_view.rs
@@ -11,30 +11,11 @@ use bevy::{
     render::camera::{ScalingMode, SubCameraView, Viewport},
 };
 
-const PADDING: u32 = 10;
-const SMALL_SIZE: u32 = 100;
-const LARGE_SIZE: u32 = 450;
-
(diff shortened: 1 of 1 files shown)


LiamGallagher737 (changes requested):

tychedelia (approved):
Great, thanks for doing this. The comments are much improved and the `resize_viewports` is a great form of documentation for users.

LiamGallagher737 (approved):

PR #15685, opened 2024-10-06
Add some missing features from the gamepads-as-entities change that were needed to update `leafwing-input-manager`.

pcwalton:
The gamepads-as-entities change caused several regressions. This patch fixes each of them:

1. This PR introduces two new fields on `GamepadInfo`: `vendor_id`, and `product_id`, as well as associated methods. These fields are simply mirrored from the `gilrs` library.

2. That PR removed the methods that allowed iterating over all pressed and released buttons, as well as the method that allowed iterating over the axis values. (It was still technically possible to do so by using reflection to access the private fields of `Gamepad`.)

3. The `Gamepad` component wasn't marked reflectable. This PR fixes that problem.

These changes allowed me to forward port `leafwing-input-manager`.


alice-i-cecile:
@s-puig can I get your review here?

alice-i-cecile (approved):

crates/bevy_input/src/gamepad.rs
```diff
@@ -505,6 +531,34 @@ impl Gamepad {
             .into_iter()
             .all(|button_type| self.just_released(button_type))
     }
+
+    /// Returns an iterator over all digital [button]s that are pressed.
+    ///
+    /// [button]: GamepadButton
+    pub fn get_pressed(&self) -> impl Iterator<Item = &GamepadButton> {
```
bushrat011899, line 538: Non-blocking bikeshed question: Should these iterator methods instead be called `iter_pressed`, `iter_just_pressed` etc.?

bushrat011899 (approved):
Looks good to me! Just a mild bikeshed-style question around whether these methods should be named `iter_x` instead of `get_x`.

crates/bevy_input/src/gamepad.rs
alice-i-cecile: We should be consistent with the ButtonInput API: https://docs.rs/bevy/latest/bevy/input/struct.ButtonInput.html#method.get_pressed

atlv24 (approved):

crates/bevy_input/src/gamepad.rs
bushrat011899: Perfectly reasonable!

alice-i-cecile:
Uuid isn't Reflect, so we either need to ignore it for Reflection or do some remote reflect nonsense. I vote the former.

crates/bevy_input/src/axis.rs
```diff
@@ -70,6 +74,11 @@ where
     pub fn remove(&mut self, input_device: T) -> Option<f32> {
         self.axis_data.remove(&input_device)
     }
+
+    /// Returns an iterator over all axes.
+    pub fn all_axes(&self) -> impl Iterator<Item = &T> {
```
MrGVSV, line 79: Any reason we can't have an `iter` method that returns an iterator over `(&T, &f32)`?

crates/bevy_input/src/gamepad.rs
```diff
@@ -399,6 +404,27 @@ impl Gamepad {
         self.info.name.as_str()
     }
 
+    /// The UUID of the game pad.
+    ///
+    /// Bevy will try to reuse IDs for game controllers across invocations of
```
MrGVSV, line 409: Might be worth adding a note how it manages to do this

crates/bevy_input/src/lib.rs
```diff
@@ -134,6 +136,7 @@ impl Plugin for InputPlugin {
                 .register_type::<RawGamepadEvent>()
                 .register_type::<RawGamepadAxisChangedEvent>()
                 .register_type::<RawGamepadButtonChangedEvent>()
+                .register_type::<Gamepad>()
```
MrGVSV, line 139: We can also add `GamepadInput`, `GamepadAxis`, and `GamepadButton` now, right? Unless those are registered elsewhere?

MrGVSV:
> Uuid isn't Reflect, so we either need to ignore it for Reflection or do some remote reflect nonsense. I vote the former, but actually the latter may be needed for serialization.

It [should](https://github.com/bevyengine/bevy/blob/d016e52843ad62f8e635f34ad02994bdeb2b2559/crates/bevy_reflect/src/impls/uuid.rs) be?

pcwalton:
@s-puig Do you want me to not expose the UUID entirely or just stop recommending it?

crates/bevy_input/src/gamepad.rs
```diff
@@ -374,8 +376,11 @@ pub struct Gamepad {
 }
 
 impl Gamepad {
-    /// Creates a gamepad with the given metadata
-    fn new(info: GamepadInfo) -> Self {
+    /// Creates a gamepad with the given metadata.
+    ///
+    /// Ordinarily, you shouldn't call this, but it can be useful for test
+    /// mocking.
+    pub fn new(info: GamepadInfo) -> Self {
```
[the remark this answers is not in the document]
pcwalton: Using events to connect mock gamepads is not actually possible as it's broken.

I'm investigating.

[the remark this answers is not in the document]
pcwalton: The problem is that sending connection events via `GamepadEvent` or `RawGamepadEvent` doesn't work anymore. You have to send them to `Events<GamepadConnectionEvent>`. This was confusing.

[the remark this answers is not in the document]
pcwalton: I made it private.

crates/bevy_input/src/lib.rs
pcwalton: Added.

crates/bevy_input/src/gamepad.rs
pcwalton: I removed `uuid()` entirely per suggestion.

crates/bevy_input/src/axis.rs
pcwalton: Added.

PR #15687, opened 2024-10-07
Update how_to_test_systems.rs

Thewsthews:
Objective

- The PR introduces small adjustments and substitutes.

Solution

- A cleaner loop method is introduced, and one of the functions is optimised accordingly. The explicit for loop is replaced with the iter_mut() method which makes it more efficient.

### DIFF
--- a/tests/how_to_test_systems.rs
+++ b/tests/how_to_test_systems.rs
@@ -32,14 +32,12 @@ fn despawn_dead_enemies(
     }
 }
 
-fn hurt_enemies(mut enemies: Query<&mut Enemy>) {
-    for mut enemy in &mut enemies {
-        enemy.hit_points -= 1;
-    }
+fn hurt_enemies(mut enemies: Query<&mut Enemy>){
+    enemies.iter_mut().for_each(|mut enemy| enemy.hit_points -= 1);
 }
 
-fn spawn_enemy(mut commands: Commands, keyboard_input: Res<ButtonInput<KeyCode>>) {
-    if keyboard_input.just_pressed(KeyCode::Space) {
+fn spawn_enemy(mut commands: Commands, keyboard_input: Res<Input<KeyCode>>) {
+    if input.just_pressed(KeyCode::Space) {
         commands.spawn(Enemy {
             hit_points: 5,
             score_value: 3,


tests/how_to_test_systems.rs
```diff
@@ -32,14 +32,12 @@ fn despawn_dead_enemies(
...
-    for mut enemy in &mut enemies {
-        enemy.hit_points -= 1;
-    }
+fn hurt_enemies(mut enemies: Query<&mut Enemy>){
+    enemies.iter_mut().for_each(|mut enemy| enemy.hit_points -= 1);
 }
 
-fn spawn_enemy(mut commands: Commands, keyboard_input: Res<ButtonInput<KeyCode>>) {
-    if keyboard_input.just_pressed(KeyCode::Space) {
+fn spawn_enemy(mut commands: Commands, keyboard_input: Res<Input<KeyCode>>) {
```
alice-i-cecile, line 39: I don't think this change compiles.

tests/how_to_test_systems.rs
```diff
@@ -32,14 +32,12 @@ fn despawn_dead_enemies(
     }
 }
 
-fn hurt_enemies(mut enemies: Query<&mut Enemy>) {
-    for mut enemy in &mut enemies {
-        enemy.hit_points -= 1;
-    }
+fn hurt_enemies(mut enemies: Query<&mut Enemy>){
+    enemies.iter_mut().for_each(|mut enemy| enemy.hit_points -= 1);
```
alice-i-cecile, line 36: This is fine, but I don't think it's a clear improvement when the point of the example is to teach something else.

tests/how_to_test_systems.rs
Thewsthews: I was of the idea that the tweak being to the generic type parameter would not cause any compile errors

alice-i-cecile: Yeah, this was renamed from `Input` to `ButtonInput` in Bevy 0.13 or so :)

alice-i-cecile:
Closing this out, but do say hi on Discord if you'd like help finding issues to work on :) The [Contributing Guide](https://bevyengine.org/learn/contribute/introduction/) is also quite useful!

PR #15688, opened 2024-10-07
Use global clear color for camera driver node.

tychedelia:
When no cameras are configured, the `ClearColor` resource has no effect on the default window.

Fixes https://discord.com/channels/691052431525675048/866787577687310356/1292601838075379796

image
image


crates/bevy_render/src/camera/camera_driver_node.rs
```diff
@@ -53,6 +53,8 @@ impl Node for CameraDriverNode {
             }
         }
 
+        let clear_color_global = world.get_resource::<ClearColor>().unwrap();
```
IceSentry, line 56: If you're going to use get_resource() and unwrap(), might as well just use resource()

IceSentry (approved):

bushrat011899 (approved):

alice-i-cecile (approved):

PR #15689, opened 2024-10-07
Fixes to animation graph evaluation

mweatherley:
# Objective

Fix a couple of substantial errors found during the development of #15665:
- `AnimationCurveEvaluator::add` was secretly unreachable. In other words, additive blending never actually occurred.
- Weights from the animation graph nodes were ignored, and only `ActiveAnimation`'s weights were used.

## Solution

Made additive blending reachable and included the graph node weight in the weight of the stack elements appended in the curve application loop of `animate_targets`.

## Testing

Tested on existing examples and on the new example added in #15665.

### DIFF
--- a/crates/bevy_animation/src/lib.rs
+++ b/crates/bevy_animation/src/lib.rs
@@ -1098,7 +1098,7 @@ pub fn animate_targets(
                 };
 
                 match animation_graph_node.node_type {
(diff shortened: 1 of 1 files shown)


alice-i-cecile (approved):

pcwalton:
Looks good, oops!

pcwalton (approved):

PR #15691, opened 2024-10-07
bevy_ecs: Replace panics in `QueryData` derive compile errors

MrGVSV:
# Objective

The current `QueryData` derive panics when it encounters an error. Additionally, it doesn't provide the clearest error message:

```rust
#[derive(QueryData)]
#[query_data(mut)]
struct Foo {
    // ...
}
```

```
error: proc-macro derive panicked
  --> src/foo.rs:16:10
   |
16 | #[derive(QueryData)]
   |          ^^^^^^^^^
   |
   = help: message: Invalid `query_data` attribute format
```

## Solution

Updated the derive logic to not panic and gave a bit more detail in the error message.

This is makes the error message just a bit clearer and maintains the correct span:

```
error: invalid attribute, expected `mutable` or `derive`
  --> src/foo.rs:17:14
   |
17 | #[query_data(mut)]
   |              ^^^
```

## Testing

You can test locally by running the following in `crates/bevy_ecs/compile_fail`:

```
cargo test --target-dir ../../../target
```

### DIFF
--- a/crates/bevy_ecs/compile_fail/tests/ui/world_query_derive.rs
+++ b/crates/bevy_ecs/compile_fail/tests/ui/world_query_derive.rs
@@ -11,6 +11,27 @@ struct MutableUnmarked {
 }
 
 #[derive(QueryData)]
+#[query_data(mut)]
+//~^ ERROR: invalid attribute, expected `mutable` or `derive`
+struct MutableInvalidAttribute {
+    a: &'static mut Foo,
+}
+
+#[derive(QueryData)]
+#[query_data(mutable(foo))]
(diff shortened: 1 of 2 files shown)


alice-i-cecile (approved):

alice-i-cecile:
Does `QueryFilter` have the same problem? I suspect that was copy-pasted.

MrGVSV:
> Does `QueryFilter` have the same problem? I suspect that was copy-pasted.

I actually didn't check 😅

Looking at it though, I can't seem to see any usage of `panic!`.

bushrat011899 (approved):
Much cleaner!

PR #15693, opened 2024-10-07
Fix a system ordering issue with motion blur for skinned meshes.

pcwalton:
Currently, it's possible for the `collect_meshes_for_gpu_building` system to run after `set_mesh_motion_vector_flags`. This will cause those motion vector flags to be overwritten, which will cause the shader to ignore the motion vectors for skinned meshes, which will cause graphical artifacts.

This patch corrects the issue by forcing `set_mesh_motion_vector_flags` to run after `collect_meshes_for_gpu_building`.

### DIFF
--- a/crates/bevy_pbr/src/render/mesh.rs
+++ b/crates/bevy_pbr/src/render/mesh.rs
@@ -163,7 +163,7 @@ impl Plugin for MeshRenderPlugin {
                 .add_systems(
                     Render,
                     (
-                        set_mesh_motion_vector_flags.before(RenderSet::Queue),
+                        set_mesh_motion_vector_flags.in_set(RenderSet::PrepareAssets),
                         prepare_skins.in_set(RenderSet::PrepareResources),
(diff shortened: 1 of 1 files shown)


bushrat011899 (approved):
New ordering makes sense!

kristoff3r (approved):

crates/bevy_pbr/src/render/mesh.rs
```diff
@@ -163,7 +163,7 @@ impl Plugin for MeshRenderPlugin {
                 .add_systems(
                     Render,
                     (
-                        set_mesh_motion_vector_flags.before(RenderSet::Queue),
+                        set_mesh_motion_vector_flags.in_set(RenderSet::PrepareAssets),
```
superdump, line 166: I don’t think the systems have to be in the same set anymore to have edges between them? Or is there something else requiring moving between sets?

pcwalton: I just thought it was cleaner to put them in the same set.

PR #15698, opened 2024-10-07
Observed entity query data fetching: `Trigger<FooEvent, (), (&mut Foo, &Bar)>`

### LINKED ISSUE #15653: Provide target-relative entity refs for observers

When writing observers, it is frequently desirable to be able to access a component on the listener entity, that is, the entity in the trigger event.

Currently there are two ways to do this: either inject a query for all components of that type and pick the one you want; or inject a DeferredWorld, and then look up the component you need on the event's entity.

bevy_mod_picking had convenience methods for injecting components directly. I'd like to see something similar for Bevy.

I'm not sure about naming, but I was thinking of something like:

```rust
entity.observe(trigger: Trigger<Pointer<Down>>, component: ListenerRef<MyComponent>);
```
There would be a `ListenerMut` version as well. Alternate names might be `TargetRef`, `TriggerRef`, or `ObserverEntityRef`.

For now, I'm using DeferredWorld, as I am unsure about the overhead of injecting a Query.

## Additional context

bevy_mod_picking also provided a means to access the "target" entity and it's components, which was the entity upon which the event was triggered before bubbling, not the entity which handled the event. However, this feature was used rarely (at least by me anyway).

This is also somewhat related to a previous ticket filed by me a long time ago (#11048) - that was rejected (for good reason), but this ticket provides a much more narrowly-scoped variation of the same idea.

ItsDoot:
# Objective

- Closes #15653 

## Solution

Adds a new type parameter `D: QueryData` to `Trigger<E, B, D>`.

## Testing

---

## Showcase

<details>

```rust
```

</details>

## Migration Guide

### DIFF
--- a/crates/bevy_ecs/src/system/combinator.rs
+++ b/crates/bevy_ecs/src/system/combinator.rs
@@ -88,7 +88,7 @@ use super::{IntoSystem, ReadOnlySystem, System};
     label = "invalid system combination",
     note = "the inputs and outputs of `{A}` and `{B}` are not compatible with this combiner"
 )]
-pub trait Combine<A: System, B: System> {
+pub trait Combine<A: System<In = Self::In>, B: System<In = Self::In>> {
     /// The [input](System::In) type for a [`CombinatorSystem`].
     type In: SystemInput;
 
@@ -137,8 +137,8 @@ impl<Func, A, B> CombinatorSystem<Func, A, B> {
 impl<A, B, Func> System for CombinatorSystem<Func, A, B>
 where
     Func: Combine<A, B> + 'static,
-    A: System,
-    B: System,
+    A: System<In = Func::In>,
+    B: System<In = Func::In>,
 {
     type In = Func::In;
     type Out = Func::Out;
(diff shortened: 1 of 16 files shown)


crates/bevy_ecs/src/system/combinator.rs
```diff
@@ -431,13 +437,21 @@ where
...
-    unsafe fn validate_param_unsafe(&mut self, world: UnsafeWorldCell) -> bool {
+    unsafe fn validate_param_unsafe(
+        &mut self,
+        input: &SystemIn<'_, Self>,
+        world: UnsafeWorldCell,
+    ) -> bool {
         // SAFETY: Delegate to other `System` implementations.
-        unsafe { self.a.validate_param_unsafe(world) && self.b.validate_param_unsafe(world) }
+        unsafe {
+            // TODO
```
ItsDoot, line 447: TODO. Unsure how to handle validating the second system.

crates/bevy_ecs/src/system/combinator.rs
```diff
@@ -88,7 +88,7 @@ use super::{IntoSystem, ReadOnlySystem, System};
     label = "invalid system combination",
     note = "the inputs and outputs of `{A}` and `{B}` are not compatible with this combiner"
 )]
-pub trait Combine<A: System, B: System> {
+pub trait Combine<A: System<In = Self::In>, B: System<In = Self::In>> {
```
SkiFire13, line 91: Isn't this a breaking change? It will force the two systems to have the same input type, so e.g. piping would no longer be implementable this way (not that `bevy`'s implementation used `Combine`).

NthTensor:
Appears to implement https://github.com/bevyengine/bevy/issues/15653.

ItsDoot:
I don't like how this approach is written. We have two ways forward for solving the original issue (IMO):

- Input-aware `SystemParam`s: I have this written up here (it's finished with tests passing): https://github.com/ItsDoot/bevy/tree/experiment/system-param-input
    - This approach is nice because it's compile-time checked (i.e. you can't use `Target<D, F>` parameters without a `SystemInput` that can provide an `Entity`)
- Editing `SystemParam::State` is suggested by cart.
    - While this approach wouldn't be compile-time checked, it would be more flexible especially if we allow reflection-based editing.

PR #15700, opened 2024-10-07
Add new crates to publish.sh

atlv24:


mockersf:
could you add `bevy_remote` and `bevy_animation/derive` that are also missing?

mockersf:
also lines should not be sorted and the order is very important

alice-i-cecile (approved):

BD103 (approved):

mockersf (approved):

PR #8600, opened 2023-05-12
Add Command::insert_batch and World::insert_batch

Connor-McMillin:
# Objective

- Fixes #8384 

## Solution

- Instead of checking for Entities being pending via `alloc_with_no_replacement`, this function directly calls `world.entities.get(entity)` to find the current `EntityLocation` (which may be invalid). It panics if it runs into an invalid entity

---

## Changelog

> This section is optional. If this was a trivial fix, or has no externally-visible impact, you can delete this section.

- What changed as a result of this PR?
- If applicable, organize changes under "Added", "Changed", or "Fixed" sub-headings
- Stick to one or two sentences. If more detail is needed for a particular change, consider adding it to the "Solution" section
  - If you can't summarize the work, your change may be unreasonably large / unrelated. Consider splitting your PR to make it easier to review and merge!

## Migration Guide

> This section is optional. If there are no breaking changes, you can delete this section.

- If this PR is a breaking change (relative to the last release of Bevy), describe how a user might need to migrate their code to support these changes
- Simply adding new functionality is not a breaking change.
- Fixing behavior that was definitely a bug, rather than a questionable design choice is not a breaking change.

PR #15614, opened 2024-10-03
Allow `World::entity` family of functions to take multiple entities and get multiple references back

ItsDoot:
# Objective

Following the pattern established in #15593, we can reduce the API surface of `World` by providing a single function to grab both a singular entity reference, or multiple entity references.

## Solution

The following functions can now also take multiple entity IDs and will return multiple entity references back:
- `World::entity`
- `World::get_entity`
- `World::entity_mut`
- `World::get_entity_mut`
- `DeferredWorld::entity_mut`
- `DeferredWorld::get_entity_mut`

If you pass in X, you receive Y:
- give a single `Entity`, receive a single `EntityRef`/`EntityWorldMut` (matches current behavior)
- give a `[Entity; N]`/`&[Entity; N]` (array), receive an equally-sized `[EntityRef; N]`/`[EntityMut; N]`
- give a `&[Entity]` (slice), receive a `Vec<EntityRef>`/`Vec<EntityMut>`
- give a `&EntityHashSet`, receive a `EntityHashMap<EntityRef>`/`EntityHashMap<EntityMut>`

Note that `EntityWorldMut` is only returned in the single-entity case, because having multiple at the same time would lead to UB. Also, `DeferredWorld` receives an `EntityMut` in the single-entity case because it does not allow structural access.

## Testing

- Added doc-tests on `World::entity`, `World::entity_mut`, and `DeferredWorld::entity_mut`
- Added tests for aliased mutability and entity existence

---

## Showcase

<details>
  <summary>Click to view showcase</summary>

The APIs for fetching `EntityRef`s and `EntityMut`s from the `World` have been unified.

```rust
// This code will be referred to by subsequent code blocks.
let world = World::new();
let e1 = world.spawn_empty().id();
let e2 = world.spawn_empty().id();
let e3 = world.spawn_empty().id();
```

Querying for a single entity remains mostly the same:

```rust
// 0.14
let eref: EntityRef = world.entity(e1);
let emut: EntityWorldMut = world.entity_mut(e1);
let eref: Option<EntityRef> = world.get_entity(e1);
let emut: Option<EntityWorldMut> = world.get_entity_mut(e1);

// 0.15
let eref: EntityRef = world.entity(e1);
let emut: EntityWorldMut = world.entity_mut(e1);
let eref: Result<EntityRef, Entity> = world.get_entity(e1);
let emut: Result<EntityWorldMut, Entity> = world.get_entity_mut(e1);
```

Querying for multiple entities with an array has changed:

```rust
[...]

PR #15702, opened 2024-10-07
Add `insert_batch` and variations

### LINKED ISSUE #2693: Add Commands::insert_batch for faster component modification

Inserting components one at a time is slow.

Add `Commands::insert_batch` and `remove_batch` to modify components en-masse in a more efficient way.

As discussed in https://github.com/bevyengine/bevy/issues/3227#issuecomment-2376553464, this should be a safe, typed equivalent to `EntityWorldMut::insert_by_ids` that's exposed to commands.

I think that a builder API that iteratively modifies the underlying `Command` struct is probably the correct design.

We could use `insert_or_spawn_batch` for the insertion case, but that does not fail in the desired ways at all.

## Additional context

Follows up on the work from #2673, which added `insert_or_spawn_batch`. Ideally we can use the same strategy for this work, and perform similar benchmarks.

### LINKED ISSUE #8384: Add Commands::insert_batch
Creating multiple entities of the same prototype in the system is a very common requirement, and existing Commands::spawn_batch can create entities in bulk and achieve efficient insertion, but this interface cannot return the Enity I have already created. If I need these entities, I have adopted the following method:

```rust
let mut entity_ list = Vec::new();

let mut bundle_ list= Vec::new();
for _ in 0..100 {
  let entity = commands.spawn_ empty();
  entity_ list.push(entity);
  bundle_ list.push((entity, Bundle))
}
commands.insert_ or_ spawn_ batch(bundle_list.into_iter());
```

But I encountered a very serious performance issue, some entities may need to check if they are in pending , and this inspection will go through an iterative process, and the code snippet is:

```rust
else if let Some(index) = self.pending.iter().position(|item| *item == entity.index) {
  self.pending.swap_ remove(index);
  let new_ free_ cursor = self.pending.len() as IdCursor;
  *self.free_ cursor.get_ mut() = new_ free_ cursor;
  self.len += 1;
  AllocAtWithoutReplacement::DidNotExist
}
```

This approach is slower than inserting one at a time

Add Commands::insert_batch, but do not check if each entity is in pending, panic If the entity does not exist.

JaySpruce:
# Objective

`insert_or_spawn_batch` exists, but a version for just inserting doesn't
- Closes #2693 
- Closes #8384 
- Adopts/supersedes #8600 

## Solution

Add `insert_batch`, along with the most common `insert` variations:
- `World::insert_batch`
- `World::insert_batch_if_new`
- `World::try_insert_batch`
- `World::try_insert_batch_if_new`
- `Commands::insert_batch`
- `Commands::insert_batch_if_new`
- `Commands::try_insert_batch`
- `Commands::try_insert_batch_if_new`

## Testing

Added tests, and added a benchmark for `insert_batch`.
Performance is slightly better than `insert_or_spawn_batch` when only inserting:

Code_HPnUN0QeWe

<details>
<summary>old benchmark</summary>

This was before reworking it to remove the `UnsafeWorldCell`:

Code_QhXJb8sjlJ
</details>

---

## Showcase

Usage is the same as `insert_or_spawn_batch`:
```
use bevy_ecs::{entity::Entity, world::World, component::Component};
#[derive(Component)]
struct A(&'static str);
#[derive(Component, PartialEq, Debug)]
struct B(f32);

let mut world = World::new();
let entity_a = world.spawn_empty().id();
let entity_b = world.spawn_empty().id();
world.insert_batch([
    (entity_a, (A("a"), B(0.0))),
    (entity_b, (A("b"), B(1.0))),
]);

assert_eq!(world.get::<B>(entity_a), Some(&B(0.0)));

```


crates/bevy_ecs/src/world/mod.rs
```diff
@@ -2377,6 +2377,255 @@ impl World {
...
+            Location::caller(),
+        );
+    }
+
+    /// For a given batch of ([`Entity`], [`Bundle`]) pairs,
+    /// adds the `Bundle` of components to each `Entity` without overwriting.
+    /// This is faster than doing equivalent operations one-by-one.
+    ///
+    /// This is the same as [`World::insert_batch`], but in case of duplicate
+    /// components will leave the old values instead of replacing them with new
```
BD103, line 2412:
```suggestion
    /// components it will leave the old values instead of replacing them with new
```

(Nit) I think adding "it" to the sentence makes it flow better, but I don't feel strongly on it.

crates/bevy_ecs/src/world/mod.rs
```diff
@@ -2377,6 +2377,255 @@ impl World {
...
+        self.insert_batch_with_caller(
+            iter,
+            InsertMode::Keep,
+            #[cfg(feature = "track_change_detection")]
+            Location::caller(),
+        );
+    }
+
+    /// Split into a new function so we can pass the calling location into the function when using
+    /// as a command.
```
BD103, line 2436: For this and the other `try_insert_batch_with_caller()` method, can you link to the function that calls this?

For example:

```suggestion
    /// Split into a new function so we can pass the calling location into the function when using
    /// as a command.
    ///
    /// This is called by [`insert_batch()`] and [`insert_batch_if_new()`].
```

crates/bevy_ecs/src/world/mod.rs
```diff
@@ -2377,6 +2377,255 @@ impl World {
...
+    }
+
+    /// For a given batch of ([`Entity`], [`Bundle`]) pairs,
+    /// adds the `Bundle` of components to each `Entity`.
+    /// This is faster than doing equivalent operations one-by-one.
+    ///
+    /// This will overwrite any previous values of components shared by the `Bundle`.
+    /// See [`World::try_insert_batch_if_new`] to keep the old values instead.
+    ///
+    /// This function silently fails by ignoring any entities that do not exist.
```
BD103, line 2515: The `try_` prefix of this function makes me thing that it would return a `Result`, but instead it silently ignores errors.

How difficult is it to return an `Err` on failure, and is that desired at all?

crates/bevy_ecs/src/system/commands/mod.rs
```diff
@@ -614,6 +614,110 @@ impl<'w, 's> Commands<'w, 's> {
...
+    /// This command panics if any of the given entities do not exist.
+    ///
+    /// See [`try_insert_batch`](Self::try_insert_batch) for the non-panicking version.
+    #[track_caller]
+    pub fn insert_batch<I, B>(&mut self, bundles_iter: I)
+    where
+        I: IntoIterator<Item = (Entity, B)> + Send + Sync + 'static,
+        B: Bundle,
+    {
+        self.queue(insert_batch(bundles_iter));
```
BD103, line 641: Since the inner `insert_batch()` function is private, can it just be inlined? I'm not sure if it would break change detection tracking, but I would prefer this simpler form.

```suggestion
        #[cfg(feature = "track_change_detection")]
        let caller = Location::caller();
    
        self.queue(move |world: &mut World| {
            world.insert_batch_with_caller(
                bundles_iter,
                InsertMode::Replace,
                #[cfg(feature = "track_change_detection")]
                caller,
            );
        });
```

BD103 (approved):
Overall looks good! I have some nits and notes, but nothing blocking.

Do note that I'm unfamiliar with ECS internals, so I'd like someone with more experience to check over those parts of the PR.

crates/bevy_ecs/src/world/mod.rs
JaySpruce: The existing `try_insert` functions all fail silently I believe. Maybe that could change, but they should change together

BD103: Ok, I'm cool with keeping things consistent!

crates/bevy_ecs/src/system/commands/mod.rs
JaySpruce: That's how the other commands are set up, I just copied it lol. I guess it's something to make commands nicer on the user-side, but I dunno

BD103: Ok, then keep it as it is!

Trashtalk217:
This would be hugely helpful for deprecating `insert_or_spawn_batch`, which is part of #15459. I do however want to point to #15614 as an example of a way to add this functionality without doubling the API surface. Can you look if this may be possible for `insert` variants as well?

JaySpruce:
I'm far from an expert, but I'm not sure if it's doable in the same way.

The `World::entity` variations deal with `Single vs Plural`, `Ref vs Mut`, and `Static vs Dynamic`, all things that work nicely with Rust types. 
The `insert` variations are `Panic vs Silent` and `Replace vs Keep`, which I can only think of ugly solutions for.

I'd imagine that we would want to include all the other `insert` functions, to keep things consistent. I've seen ideas about chaining commands to achieve a similar nice API, which would work for the `insert` `Commands`, but I don't know if the functions on the `World` can do that.

The best I've got is passing in an enum or something, but that's obvious so there's probably a reason that's not how it works already

Trashtalk217 (approved):

JaySpruce:
Maybe this could be its own PR, but the original PR has some unhealthy usage of `UnsafeWorldCell`, so I think this is a worthwhile fix.

Here's the context:

The problem was that I needed access to both the `World`'s `Entities` storage and a `&mut World` to make a new `BundleInserter`. 

`insert_or_spawn_batch` snuck around this by making a `BundleSpawner` first, which internally made an `UnsafeWorldCell` that was then able to be used to access `Entities` while still using a `&mut World`.

`insert_batch`, of course, doesn't need a `BundleSpawner`. However, if you try to recreate the effect with a `BundleInserter`, you'll find that it needs an `ArchetypeId`, which you get from an `EntityLocation`... which you get from `Entities`.

I originally used an `UnsafeWorldCell` directly to get around this, which worked, but it turns out the documentation *specifically* says to *not* do that.

So, eventually I figured out a (hopefully) proper solution: deal with the first entity in the iterator on its own, so you can make a `BundleInserter` with its `ArchetypeId` without triggering the borrow checker, and without dealing in unsafe (well, *unsafe* unsafe) operations.

And as a bonus, it's slightly faster now!
Code_ta6N8YPp6M

crates/bevy_ecs/src/lib.rs
```diff
@@ -1699,6 +1699,114 @@ mod tests {
         );
     }
 
+    #[test]
+    fn insert_batch() {
```
hymm, line 1703: could we get a test where you're inserting onto multiple entities in the same archetype?

JaySpruce: Is this good? (it passes) Also, should I make more for the other variations, or is one enough?
```
#[test]
fn insert_batch_same_archetype() {
    let e0 = world.spawn((A(0), B(0))).id();
    let e1 = world.spawn((A(0), B(0))).id();
    let e2 = world.spawn(B(0)).id();

    let values = vec![(e0, (B(1), C)), (e1, (B(2), C)), (e2, (B(3), C))];

    world.insert_batch(values);

    assert_eq!(
        world.get::<A>(e0),
        Some(&A(0)),
        "first entity's A component should not have changed"
    );
    assert_eq!(
        world.get::<B>(e0),
        Some(&B(1)),
        "first entity's B component should have been replaced"
    );
    assert_eq!(
        world.get::<C>(e0),
        Some(&C),
        "first entity should have received C component"
    );
    assert_eq!(
        world.get::<A>(e1),
        Some(&A(0)),
        "second entity's A component should not have changed"
    );
    assert_eq!(
        world.get::<B>(e1),
        Some(&B(2)),
        "second entity's B component should have been replaced"
    );
    assert_eq!(
        world.get::<C>(e1),
        Some(&C),
        "second entity should have received C component"
    );
    assert_eq!(
        world.get::<A>(e2),
        None,
        "third entity should not have received A component"
    );
    assert_eq!(
        world.get::<B>(e2),
        Some(&B(3)),
        "third entity's B component should have been replaced"
    );
    assert_eq!(
        world.get::<C>(e2),
        Some(&C),
        "third entity should have received C component"
    );
}
```

JaySpruce:
Should I remove the `try_` variants from the `World`? I've realized that they're only really useful for `Commands`

hymm (commented):
Am I understanding correctly that this could potentially allocate twice if we're moving multiple entities into the same archetype? And also make multiple moves of entities in the old table as we're moving entities out of it? Probably some room for future optimizations if so.

JaySpruce:
I think keeping multiple inserters allocated would run into borrow problems, since they all need a &mut World

crates/bevy_ecs/src/lib.rs
hymm: I think just this one test is good enough. The asserts are getting a little long. Would it be cleaner to use a query and collect the results?

hymm:
> Should I remove the try_ variants from the World? I've realized that they're only really useful for Commands

I'm not sure I follow. We do have a `try_despawn` on world. I think it's more unlikely that an entity might not exist when working with World, but not impossible. You could be referencing an entity that is manually updated in a resource.

> I think keeping multiple inserters allocated would run into borrow problems, since they all need a &mut World

It would definitely require something different from BundleInserter and probably a bunch of unsafe.

crates/bevy_ecs/src/lib.rs
JaySpruce: Much shorter:
```
#[test]
fn insert_batch_same_archetype() {
    let e0 = world.spawn((A(0), B(0))).id();
    let e1 = world.spawn((A(0), B(0))).id();
    let e2 = world.spawn(B(0)).id();

    let values = vec![(e0, (B(1), C)), (e1, (B(2), C)), (e2, (B(3), C))];

    world.insert_batch(values);
    let mut query = world.query::<(Option<&A>, &B, &C)>();
    let component_values = query.get_many(&world, [e0, e1, e2]).unwrap();

    assert_eq!(
        component_values, 
        [(Some(&A(0)), &B(1), &C), (Some(&A(0)), &B(2), &C), (None, &B(3), &C)],
        "all entities should have had their B component replaced, received C component, and had their A component (or lack thereof) unchanged"
    );
}
```

JaySpruce:
I got the #15614 thing working, using this trait:
<details>
<summary>Trait implementation</summary>

```
use core::iter;
use crate::{bundle::Bundle, entity::Entity};

/// Types that can be used to insert [`Bundles`](Bundle) into [`Entities`](Entity) from a [`World`].
/// 
/// Provided implementations are:
/// - (`Entity`, `Bundle`): Insert into a single entity.
/// - [(`Entity`, `Bundle`); N]: Insert into a static array of entities.
/// - Vec<(`Entity`, `Bundle`)>: Insert into a vector of entities.
/// 
/// [`World`]: crate::world::World
pub trait WorldEntityInsert<B: Bundle>
{
    fn iterator(self) -> impl Iterator<Item = (Entity, B)>;
}

impl<B: Bundle> WorldEntityInsert<B> for (Entity, B)
{
    fn iterator(self) -> impl Iterator<Item = (Entity, B)> {
        iter::once(self).into_iter()
    }
}

impl<B: Bundle, const N: usize> WorldEntityInsert<B> for [(Entity, B); N]
{
    fn iterator(self) -> impl Iterator<Item = (Entity, B)> {
        self.into_iter()
    }
}

impl<B: Bundle> WorldEntityInsert<B> for Vec<(Entity, B)>
where
    B: Bundle
{
    fn iterator(self) -> impl Iterator<Item = (Entity, B)> {
        self.into_iter()
    }
}
```
</details>

It doesn't reduce the API, but it adds the ability to insert into a single entity or an array of entities, instead of just a `Vec`. Since `insert_batch` uses an iterator, it was pretty much a drop-in replacement, no changes to the actual function.

So, follow-up questions:
- Where should the trait live? I made an `entity_insert.rs` to write it, but I don't know if it needs a whole file
- What should the functions be called? `insert_batch` isn't quite accurate since you can insert just one, but it's still probably mostly used for batches
- Are there any other types that should be able to work? Maybe slices, but those are references and I'm not sure could/should work

PR #15703, opened 2024-10-07
add example for ease functions

mockersf:
# Objective

- Followup to #15675 
- Add an example showcasing the functions

## Solution

- Add an example showcasing the functions
- Some of the functions from the interpolation crate are messed up, fixed in #15706

ease

PR #15706, opened 2024-10-07
fix some of the ease functions from interpolation

mockersf:
# Objective

- Followup to #15675
- Some of the functions are wrong, noticed in #15703: `Sine`, `Elastic` and `Back`

## Solution

- Fix them and make them deterministic

ease-fixed-functions

### DIFF
--- a/crates/bevy_math/src/curve/easing.rs
+++ b/crates/bevy_math/src/curve/easing.rs
@@ -1,8 +1,6 @@
 //! Module containing different [`Easing`] curves to control the transition between two values and
 //! the [`EasingCurve`] struct to make use of them.
 
-use core::f32::consts::{FRAC_PI_2, FRAC_PI_3, PI};
-
 use crate::{
     ops::{self, FloatPow},
     VectorSpace,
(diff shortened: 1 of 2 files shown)


alice-i-cecile (approved):
I'd prefer to shed the whole `interpolation` dependency, but this is fine for now :)

crates/bevy_math/src/curve/easing.rs
```diff
@@ -84,6 +84,64 @@ where
...
+        1.0 - ops::cos(t * FRAC_PI_2)
+    }
+    #[inline]
+    pub(crate) fn sine_out(t: f32) -> f32 {
+        ops::sin(t * FRAC_PI_2)
+    }
+
+    #[inline]
+    pub(crate) fn back_in(t: f32) -> f32 {
+        let c = 1.70158;
```
IQuick143, line 103: What is this constant, is it some value (apparently not sqrt(3)) or is it just arbitrary?

mockersf: took it from https://easings.net/#easeInBack, I pretty much copied the "Math function" part

PR #15707, opened 2024-10-07
Rename the `Pickable` component and fix incorrect documentation

### LINKED ISSUE #15632: Consider a rename of the `Pickable` component
The `Pickable` component used to be required in `bevy_mod_picking` for an entity to be 'pickable', hence the name.
This requirement has since been lifted and now it is only used to override picking behavior.

The naming is now misleading, making users and [bevy contributors](https://github.com/bevyengine/bevy/blob/main/examples/picking/simple_picking.rs#L32) believe it's still required and so it's proposed to change it. Suggestions welcome

tim-blackbird:
# Objective

- Rename `Pickable` to `PickingBehavior` to counter the easily-made assumption that the component is required. It is optional
- Fix and clarify documentation
  - The docs in `crates/bevy_ui/src/picking_backend.rs` were incorrect about the necessity of `Pickable`
- Plus two minor code quality changes in this commit (7c2e75f48d20387ed27f38b3c9f2d28ae8adf84f)

Closes #15632

### DIFF
--- a/crates/bevy_picking/src/backend.rs
+++ b/crates/bevy_picking/src/backend.rs
@@ -20,7 +20,7 @@
 //! - The [`PointerHits`] events produced by a backend do **not** need to be sorted or filtered, all
 //!   that is needed is an unordered list of entities and their [`HitData`].
 //!
-//! - Backends do not need to consider the [`Pickable`](crate::Pickable) component, though they may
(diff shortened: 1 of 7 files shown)


Jondolf (approved):
`Pickable` confused me as well 😄 The new `PickingBehavior` name seems like a good replacement.

PR #10392, opened 2023-11-05
Add `Image` methods for easy access to a pixel's color

inodentry:
# Objective

If you want to draw / generate images from the CPU, such as:
 - to create procedurally-generated assets
 - for games whose artstyle is best implemented by poking pixels directly from the CPU, instead of using shaders

It is currently very unergonomic to do in Bevy, because you have to deal with the raw bytes inside `image.data`, take care of the pixel format, etc.

## Solution

This PR adds some helper methods to `Image` for pixel manipulation. These methods allow you to use Bevy's user-friendly `Color` struct to read and write the colors of pixels, at arbitrary coordinates (specified as `UVec3` to support any texture dimension). They handle encoding/decoding to the `Image`s `TextureFormat`, incl. any sRGB conversion.

While we are at it, also add methods to help with direct access to the raw bytes. It is now easy to compute the offset where the bytes of a specific pixel coordinate are found, or to just get a Rust slice to access them.

Caveat: `Color` roundtrips are obviously going to be lossy for non-float `TextureFormat`s. Using `set_color_at` followed by `get_color_at` will return a different value, due to the data conversions involved (such as `f32` -> `u8` -> `f32` for the common `Rgba8UnormSrgb` texture format). Be careful when comparing colors (such as checking for a color you wrote before)!

Also adding a new example: `cpu_draw` (under `2d`), to showcase these new APIs.

---

### Added

 - `Image` APIs for easy access to the colors of specific pixels.

PR #15709, opened 2024-10-07
Add methods for getting Image pixel rects

clarfonthey:
# Objective

In addition to modifying individual pixels' colors, it's very common to want to copy one texture onto another. In fact, the goal of this code is to eventually be used by `TextureAtlasBuilder`, which does exactly that, but I wanted to keep the change small for now.

## Solution

Two methods are added to `Image`: `rect_bytes` and `rect_bytes_mut`. Both of these return an iterator over the slices of bytes for the chosen rect.

The iterator types, `RectBytes` and `RectBytesMut`, both keep track of the texture format of the original image and offer a `format` method to get this. The goal here is so that functions which accept a rect can still verify that the texture formats are valid, and to additionally support the following methods on `RectBytesMut`:

* `copy_from(RectBytes)` will directly copy bytes from another rect, assuming that the sizes and formats match.
* `translate_from(RectBytes)` will also allow copying between rects with different formats, using the code added by #10392.

It also adds `TextureCopyError` as the error for these two methods, since `TextureAccessError` didn't quite fit.

Both methods make it clear that transparency is completely ignored, and that this is just copying from one rect to another. Maybe in the future we could add a version that supports transparency.

## Testing

I do plan to add proper unit tests for these, but I want to show the API first and potentially tweak that before I'm done, which is why I'm marking this PR as a draft for now.


crates/bevy_image/src/image.rs
```diff
@@ -1113,25 +1114,68 @@ impl Image {
...
+    /// Returns:
+    ///
+    /// 1. The absolute range in the data for the rect.
+    /// 2. The length of a row in the rect.
+    /// 3. The range within each row for the rect.
+    #[inline]
+    fn rect_bytes_inner(&self, rect: URect, z: u32) -> Option<(Range<usize>, usize, Range<usize>)> {
+        let width = self.texture_descriptor.size.width;
+        let start = self.pixel_data_offset(UVec3::new(0, rect.min.y, z))?;
+        let end = self.pixel_data_offset(UVec3::new(width, rect.max.y, z))?;
```
KirmesBude, line 1165: This seems to return past the given rect in my tests when trying to pass a rect that should describe the whole Image. Notice the -1 here https://github.com/KirmesBude/bevy/commit/7542e4e8acac1262d0438d437126da7f19863e79#diff-b2f6268dff5264f58b7bcbe294d170114d810b2d6bdf8c86f9c371bbbc9422d9R118-R137

On that note, I think a convenience method on Image to translate the whole Image would be nice.

crates/bevy_image/src/image.rs
```diff
@@ -1504,14 +1554,180 @@ impl Image {
...
+
+        for (dest, src) in self.zip(other.clone()) {
+            dest.copy_from_slice(src);
+        }
+        Ok(())
+    }
+
+    /// Copies the pixels from another rect to this one by performing color conversion.
+    ///
+    /// Will fail if the rects are different sizes or formats. Completely
```
KirmesBude, line 1668: This does not seem to be correct, conversion/translation does happen with different formats.

crates/bevy_image/src/image.rs
```diff
@@ -1504,14 +1554,180 @@ impl Image {
...
+    /// ignores transparency and will directly overwrite the bytes.
+    pub fn copy_from(self, other: RectBytes<'_>) -> Result<(), TextureCopyError> {
+        let dest = UVec2::new(self.range.len() as u32, self.data.len() as u32);
+        let src = UVec2::new(other.range.len() as u32, other.data.len() as u32);
+        if dest != src {
+            return Err(TextureCopyError::SizeMismatch { dest, src });
+        }
+
+        if self.format != other.format {
+            return Err(TextureCopyError::FormatMismatch {
```
KirmesBude, line 1654: There is no attempt to convert between formats here so this Error/message here does not quite fit.

crates/bevy_image/src/image.rs
```diff
@@ -1504,14 +1554,180 @@ impl Image {
...
+        );
+
+        if dest != src {
+            return Err(TextureCopyError::SizeMismatch { src, dest });
+        }
+
+        if !Image::supported_color_format(self.format)
+            || !Image::supported_color_format(other.format)
+        {
+            return Err(TextureCopyError::FormatMismatch {
```
KirmesBude, line 1691: I think this error does not quite fit for this case either. At this point in the code the formats will always "mismatch".

KirmesBude (commented):
I was just trying this out and wanted to leave some comments. Feel free to ignore considering it is still in draft.

crates/bevy_image/src/image.rs
clarfonthey: Yeah, I was thinking of adding a method to convert the entire image to a `RectBytes`, but hadn't gotten there yet.

crates/bevy_image/src/image.rs
clarfonthey: Yup, that was a typo; I copied the docs from the other funciton incorrectly.

crates/bevy_image/src/image.rs
clarfonthey: Maybe the name of the variant can be improved, but at least the error message simply says it couldn't convert from source to dest here, which is accurate based upon the method being called.

I wanted a version of the method that guarantees no fancy conversion was done for speed/accuracy (colour conversion is slow, and can cause rounding issues even between colours of the same format) but I didn't think it warranted a special error variant.

crates/bevy_image/src/image.rs
clarfonthey: They won't match, but it's only if they can't be converted that the conversion will fail. Unless you're referring to something else.

clarfonthey:
Also FWIW, I absolutely appreciate feedback despite this still being a draft! The reason why I posted it was so that folks could comment on it, since the goal is to get a feel whether this is a reasonable approach or not.

crates/bevy_image/src/image.rs
KirmesBude: I think it is ultimately fine, I just got confused by the name of the error variant and the error message.

clarfonthey:
I'm going to close this considering how it's definitely gotten stale, but someone is more than welcome to pick this up from where I left off!

PR #15710, opened 2024-10-07
Vary transforms for custom_skinned_mesh example

shanecelis:
# Objective

Enhance the [custom skinned mesh example](https://bevyengine.org/examples/animation/custom-skinned-mesh/) to show some variety and clarify what the transform does to the mesh. 

## Solution

https://github.com/user-attachments/assets/c919db74-6e77-4f33-ba43-0f40a88042b3

Add variety and clarity with the following changes:

- vary transform changes,
- use a UV texture,
- and show transform changes via gizmos.

(Maybe it'd be worth turning on wireframe rendering to show what happens to the mesh. I think it'd be nice visually but might make the code a little noisy.)

## Testing

I exercised it on my x86 macOS computer. It'd be good to have it validated on Windows, Linux, and WASM.

---

## Showcase

- Custom skinned mesh example varies the transforms changes and uses a UV test texture.


alice-i-cecile:
Wow those UVs really make this come to life :D

LiamGallagher737 (approved):
Looks just as awesome on Linux

alice-i-cecile (approved):

PR #15714, opened 2024-10-07
Curves: `FromReflect` boogaloo part 2

mweatherley:
# Objective

Allow curve adaptors to be reliably `Reflect` even if the curves they hold are not `FromReflect`. This allows them, for example, to be used in `bevy_animation`. I previously addressed this with the functional adaptors, but I forgot to address this in the case of fields that hold other curves and not arbitrary functions.

## Solution

Do the following on every curve adaptor that holds another curve:
```rust
// old:
#[derive(Reflect)]
```

```rust
// new:
#[derive(Reflect, FromReflect)]
#[reflect(from_reflect = false)]
```

This looks inane, but it's necessary because the default `#[derive(Reflect)]` macro places `FromReflect` bounds on everything. To avoid this, we opt out of deriving `FromReflect` with that macro by adding `#[reflect(from_reflect = false)]`, then separately derive `FromReflect`. (Of course, the latter still has the `FromReflect` bounds, which is fine.)

### DIFF
--- a/crates/bevy_math/src/curve/adaptors.rs
+++ b/crates/bevy_math/src/curve/adaptors.rs
@@ -9,7 +9,7 @@ use core::fmt::{self, Debug};
 use core::marker::PhantomData;
 
 #[cfg(feature = "bevy_reflect")]
-use bevy_reflect::{utility::GenericTypePathCell, Reflect, TypePath};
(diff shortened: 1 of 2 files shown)


mockersf (approved):

alice-i-cecile (approved):
That's *bad*. Merging.

PR #15728, opened 2024-10-08
Fix oit webgl

tychedelia:
The previous fixes were breaking pretty much everything on main due to naga-oil complaining about the OIT shader not being loaded, since apparently webgl is a default feature. This fix is a bit messier, but properly warns the user and is probably what we should have gone for in the first place.


crates/bevy_core_pipeline/src/oit/resolve/mod.rs
```diff
@@ -61,6 +61,17 @@ impl Plugin for OitResolvePlugin {
             return;
         };
 
+        if !render_app
```
JMS55, line 64: Is it ok to put this here? What about the call to add_systems() during build? Is it ok to add those?

We might be better off doing what I did with the meshlet visbuffer resolve, and insert a dummy shader when the plugin is not loaded.

IceSentry (approved):

crates/bevy_core_pipeline/src/oit/resolve/mod.rs
IceSentry: Why would it not be okay?

pcwalton:
Non-WebGL is broken too (e.g. `cargo run --example 3d_shapes` is broken), and I've confirmed that this PR fixes it.

crates/bevy_core_pipeline/src/oit/resolve/mod.rs
tychedelia: I wasn't sure exactly where to place this check, so decided to put it closest to the problem. The `OitResolvePipeline` layout that's created in it's `FromWorld` impl is the bad behavior here.

The missing resource means the pass can never be queued, so effectively solves the problem. The user can still try to add the component to their camera but it won't do anything.

This is all modeled after the checks we have in SSAO plugin, which requires similar feature checks.

tychedelia: Hmmm, I guess the SSAO plugin does move all the stuff into `finish`. I can do that here too for consistency, although it's mostly just the difference between a warning for the missing resource or not.

alice-i-cecile (approved):

PR #15733, opened 2024-10-08
Implement `From` translation and rotation for isometries

Jondolf:
# Objective

Several of our APIs (namely gizmos and bounding) use isometries on current Bevy main. This is nicer than separate properties in a lot of cases, but users have still expressed usability concerns.

One problem is that in a lot of cases, you only care about e.g. translation, so you end up with this:

```rust
gizmos.cross_2d(
    Isometry2d::from_translation(Vec2::new(-160.0, 120.0)),
    12.0,
    FUCHSIA,
);
```

The isometry adds quite a lot of length and verbosity, and isn't really that relevant since only the translation is important here.

It would be nice if you could use the translation directly, and only supply an isometry if both translation and rotation are needed. This would make the following possible:

```rust
gizmos.cross_2d(Vec2::new(-160.0, 120.0), 12.0, FUCHSIA);
```

removing a lot of verbosity.

## Solution

Implement `From<Vec2>` and `From<Rot2>` for `Isometry2d`, and `From<Vec3>`, `From<Vec3A>`, and `From<Quat>` for `Isometry3d`. These are lossless conversions that fit the semantics of `From`.

This makes the proposed API possible! The methods must now simply take an `impl Into<IsometryNd>`, and this works:

```rust
gizmos.cross_2d(Vec2::new(-160.0, 120.0), 12.0, FUCHSIA);
```


alice-i-cecile (approved):

tim-blackbird (approved):
Very nice, with this I'd say the gizmo API has been improved over what it was before the move to isometry :)

PR #15736, opened 2024-10-08
Remove warning for children in UI hierarchies without Style

alice-i-cecile:
# Objective

As discussed in #15591, this warning prevents us from storing leaf nodes without a `Style` component. Because text sections (as distinct entities) should not be laid out using `taffy`, this warning is incorrect.

Users may also have other uses for doing this, and this should generally increase flexibility without posing particularly serious correctness concerns.

## Solution

- removed warning about non-UI children with UI parents
- improved the warning about UI parents with non-UI parents
   - this warning should stay, for now, as it results in a genuine failure to perform `taffy` layout
   - that said, we should be clearer about the cause and potentially harmful results of this!
   
## Testing

I inserted an empty entity into the hierarchy in the `button` example as a leaf node, and it ran with no warnings.

### DIFF
--- a/crates/bevy_ui/src/layout/ui_surface.rs
+++ b/crates/bevy_ui/src/layout/ui_surface.rs
@@ -121,12 +121,6 @@ impl UiSurface {
         for child in children {
             if let Some(taffy_node) = self.entity_to_taffy.get(&child) {
                 self.taffy_children_scratch.push(*taffy_node);
-            } else {
-                warn!(
-                    "Unstyled child `{child}` in a UI entity hierarchy. You are using an entity \
-without UI components as a child of an entity with UI components, results may be unexpected. \
-If this is intentional, consider adding a GhostNode component to this entity."
-                );
             }
         }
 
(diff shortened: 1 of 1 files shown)


alice-i-cecile:
This fix was much simpler than the more specific UI -> non-UI -> UI pattern that @cart called out as problematic [here](https://github.com/bevyengine/bevy/pull/15591#discussion_r1788278534). I also think it's somewhat better, as non-UI -> UI still has not-particularly well defined behavior in general.

We *could* treat that case as fine if and only if there's no UI root above it by treating it as the start of a new tree, but really that feels quite a bit more complex for users to reason about and will involve a more complex fix and ongoing validity checks.

UkoeHB (approved):

tim-blackbird (approved):

cart (approved):
> This fix was much simpler than the more specific UI -> non-UI -> UI pattern that @cart called out as problematic

Ah yeah this is definitely the right path forward conceptually. It being simpler / just removing code is icing on the cake!

PR #15737, opened 2024-10-08
Migrate reflection probes to required components

Jondolf:
# Objective

Getting closer to the end! Another part of the required components migration: reflection probes.

## Solution

As per the [proposal added by Cart](https://hackmd.io/@bevy/required_components/%2FNmpIh0tGSiayGlswbfcEzw) (Proposal 2), make `LightProbe` require `Transform` and `Visibility`, and deprecate `ReflectionProbeBundle`.

Note that this proposal wasn't officially blessed yet, but it is the only existing one that really works, so I implemented it here for consideration.

## Testing

I ran the reflection probe example, and it appears to work.

---

## Migration Guide

`ReflectionProbeBundle` has been deprecated in favor of inserting the `LightProbe` and `EnvironmentMapLight` components directly. Inserting them will now automatically insert `Transform` and `Visibility` components.


crates/bevy_pbr/src/light_probe/mod.rs
```diff
@@ -104,6 +103,7 @@ pub struct LightProbePlugin;
 /// with other engines should be aware of this terminology difference.
 #[derive(Component, Debug, Clone, Copy, Default, Reflect)]
 #[reflect(Component, Default, Debug)]
+#[require(Transform, Visibility)]
 pub struct LightProbe;
```
Jondolf, line 107: Not directly related to this PR, but should this require `SyncToRenderWorld`? Lights require it, but this isn't really a light itself, and I'm not too sure which components should or shouldn't have the component. The bundle didn't have it, so I didn't add it yet.

crates/bevy_pbr/src/light_probe/environment_map.rs
```diff
@@ -137,6 +137,10 @@ pub struct EnvironmentMapIds {
 /// surrounding a region in space. For more information, see
 /// [`crate::environment_map`].
 #[derive(Bundle, Clone)]
+#[deprecated(
+    since = "0.15.0",
+    note = "Use the `LightProbe` and `EnvironmentMapLight` components instead. Inserting them will now also insert the other components required by them automatically."
+)]
```
Jondolf, line 143: Question: Do we actually want to deprecate this yet, considering you still need to add both components? Or should I only remove the `SpatialBundle` from here?

crates/bevy_pbr/src/light_probe/mod.rs
tim-blackbird: It doesn't look like it's ever used in the render world, so it doesn't need to be synced

crates/bevy_pbr/src/light_probe/mod.rs
```diff
@@ -66,9 +66,8 @@ pub struct LightProbePlugin;
...
 ///
 /// The light probe range is conceptually a unit cube (1×1×1) centered on the
-/// origin.  The [`bevy_transform::prelude::Transform`] applied to this entity
-/// can scale, rotate, or translate that cube so that it contains all fragments
-/// that should take this light probe into account.
+/// origin. The [`Transform`] applied to this entity can scale, rotate, or translate
+/// that cube so that it contains all fragments that should take this light probe into account.
 ///
 /// Note that a light probe will have no effect unless the entity contains some
 /// kind of illumination, which can either be an [`EnvironmentMapLight`] or an
```
tim-blackbird, line 73: Maybe we should bump this paragraph up by one?

crates/bevy_pbr/src/light_probe/environment_map.rs
tim-blackbird: Definitely unfortunate that the `EnvironmentMapLight` component is also used on camera entities. Looking at the examples it feels like that should be merged into the `Skybox` component. We could then require `LightProbe` for `EnvironmentMapLight`, but alas.

I think we should deprecate this, even though it's not perfect

tim-blackbird (approved):

bushrat011899 (approved):

crates/bevy_pbr/src/light_probe/environment_map.rs
IceSentry: What do you mean by "also used on camera entities".  Where else other than camera entities is it used?

Jondolf:
@alice-i-cecile attempt number two?

PR #15742, opened 2024-10-08
Replace `Handle<AnimationGraph>` component with a wrapper

### LINKED ISSUE #15717: Stop using `Handle<AnimationGraph>` as a component
See #15716

tim-blackbird:
# Objective

- Closes #15717 

## Solution

- Wrap the handle in a new wrapper component: `AnimationGraphHandle`.

## Testing

Searched for all instances of `AnimationGraph` in the examples and updated and tested those

## Migration Guide

`Handle<AnimationGraph>` is no longer a component. Instead, use the `AnimationGraphHandle` component which contains a `Handle<AnimationGraph>`.

### DIFF
--- a/crates/bevy_animation/src/graph.rs
+++ b/crates/bevy_animation/src/graph.rs
@@ -137,14 +137,14 @@ impl From<Handle<AnimationGraph>> for AnimationGraphHandle {
 }
 
 impl From<AnimationGraphHandle> for AssetId<AnimationGraph> {
-    fn from(material: AnimationGraphHandle) -> Self {
-        material.id()
+    fn from(handle: AnimationGraphHandle) -> Self {
+        handle.id()
     }
 }
 
 impl From<&AnimationGraphHandle> for AssetId<AnimationGraph> {
-    fn from(material: &AnimationGraphHandle) -> Self {
-        material.id()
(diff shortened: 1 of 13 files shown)


bushrat011899 (approved):

crates/bevy_animation/src/graph.rs
```diff
@@ -122,6 +125,29 @@ pub struct AnimationGraph {
...
+pub struct AnimationGraphHandle(pub Handle<AnimationGraph>);
+
+impl From<Handle<AnimationGraph>> for AnimationGraphHandle {
+    fn from(handle: Handle<AnimationGraph>) -> Self {
+        Self(handle)
+    }
+}
+
+impl From<AnimationGraphHandle> for AssetId<AnimationGraph> {
+    fn from(material: AnimationGraphHandle) -> Self {
```
mweatherley, line 140: It's not a material :)

mweatherley (approved):
Just one tiny nit

crates/bevy_animation/src/graph.rs
tim-blackbird: copy-pasta makes bad spaghetti :)

atornity (approved):

Issue #15716, opened 2024-10-07
Remove `Component` trait impl from `Handle<T>`


tim-blackbird:
# Objective
Remove the `Component` trait implementation from `Handle<T>` as its usage is now considered an anti-pattern.

The move to required components has notably made porting error prone both internally and for the code of early adopters. Due a lack of errors on the usage of high-profile but now no longer relevant `Handle` components like `Handle<Mesh>` and `Handle<StandardMaterial>`.

benfrankel:
This will also resolve https://github.com/bevyengine/bevy/issues/14124

tim-blackbird:
I thought I remembered there already being an issue related to this already :)

I haven't created an issue for `Handle<Image>` since the sprite migration to required components is well on it's way already

mgi388:
@tim-blackbird I think the new handle wrapper components need to be registered in the type registry. I tried locally myself changing one of my own `Handle<Foo>`s to `FooHandle<Foo>` and confirmed in egui inspector if we don't `app.register_type::<FooHandle>()`, then you cannot inspect the component. 

Separately, I was also wondering if the engine should have a macro for this boilerplate? https://github.com/bevyengine/bevy/blob/700123ec647d153b5d6cce70b94ea30efa5c1d86/crates/bevy_animation/src/graph.rs#L133-L149 Fine if not (or if it's not generic enough to be a macro generated for every component wrapper), or if it could come later, but I just wondered if it would help the ecosystem migrate.

tim-blackbird:
I always forget to add those calls to `register_type` 😔
- #15784

We could potentially make use of the `derive_more` crate to cut down on some boilerplate

smer44:
So how to query to get specific Mesh pls?

freddycansic:
> So how to query to get specific Mesh pls?

I managed to do it like this, bear in mind this is for `MeshMaterial2d`, I have not experimented with the regular `MeshMaterial`, but I would assume it works the same.
 
```
fn custom_material_system(
    mut custom_material: ResMut<Assets<CustomMaterial>>,
    specific_mesh: Single<&MeshMaterial2d<CustomMaterial>, With<SpecificMeshComponent>>,
) {
    if let Some(material) = custom_material.get_mut(specific_mesh.0.id()) {
        // Your logic here
    }
}

```

PR #15749, opened 2024-10-08
Remove `Handle<T>` trait implementations that are dependent on `Component`

tim-blackbird:
# Objective

- Another step towards #15716
- Remove trait implementations that are dependent on `Handle<T>` being a `Component`

## Solution

- Remove unused `ExtractComponent` trait implementation for `Handle<T>`
- Remove unused `ExtractInstance` trait implementation for `AssetId`
  - Although the `ExtractInstance` trait wasn't used, the `AssetId`s were being stored inside of `ExtractedInstances` which has an `ExtractInstance` trait bound on its contents.
  I've upgraded the `RenderMaterialInstances` type alias to be its own resource, identical to `ExtractedInstances<AssetId<M>>` to get around that with minimal breakage.
## Testing

Tested `many_cubes`, rendering did not explode

### DIFF
--- a/crates/bevy_pbr/src/material.rs
+++ b/crates/bevy_pbr/src/material.rs
@@ -18,6 +18,7 @@ use bevy_core_pipeline::{
 };
 use bevy_derive::{Deref, DerefMut};
 use bevy_ecs::{
+    entity::EntityHashMap,
     prelude::*,
     system::{lifetimeless::SRes, SystemParamItem},
 };
@@ -25,7 +26,6 @@ use bevy_reflect::std_traits::ReflectDefault;
 use bevy_reflect::Reflect;
 use bevy_render::{
     camera::TemporalJitter,
-    extract_instances::ExtractedInstances,
     extract_resource::ExtractResource,
     mesh::{Mesh3d, MeshVertexBufferLayoutRef, RenderMesh},
     render_asset::{PrepareAssetError, RenderAsset, RenderAssetPlugin, RenderAssets},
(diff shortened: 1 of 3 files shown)


crates/bevy_pbr/src/material.rs
```diff
@@ -490,7 +490,15 @@ impl<P: PhaseItem, M: Material, const I: usize> RenderCommand<P> for SetMaterial
     }
 }
 
-pub type RenderMaterialInstances<M> = ExtractedInstances<AssetId<M>>;
+/// Stores all extracted instances of a [`Material`] in the render world.
+#[derive(Resource, Deref, DerefMut)]
+pub struct RenderMaterialInstances<M: Material>(pub EntityHashMap<AssetId<M>>);
```
IceSentry, line 495: Am I missing something? This doesn't seem to be used anywhere.

tim-blackbird: You must be, it's being use in quite a few places :)

None of them show up in the diff because it's API is exactly what it was, in case that's where you were looking

alice-i-cecile (approved):

crates/bevy_render/src/extract_component.rs
```diff
@@ -197,18 +195,7 @@ impl<C: ExtractComponent> Plugin for ExtractComponentPlugin<C> {
...
-    type Out = Handle<T>;
-
-    #[inline]
-    fn extract_component(handle: QueryItem<'_, Self::QueryData>) -> Option<Self::Out> {
-        Some(handle.clone_weak())
-    }
-}
-
-/// This system extracts all components of the corresponding [`ExtractComponent`], for entities that are synced via [`crate::sync_world::SyncToRenderWorld`].
+/// This system extracts all components of the corresponding [`ExtractComponent`], for entities that are synced via [`SyncToRenderWorld`].
```
Jondolf, line 198: This is making CI unhappy
```suggestion
/// This system extracts all components of the corresponding [`ExtractComponent`], for entities that are synced via [`SyncToRenderWorld`](crate::sync_world::SyncToRenderWorld).
```

NiseVoid (approved):

Jondolf (approved):
Looks correct to me. An alternative would probably be to somehow generically derive `ExtractInstance` for materials such that `QueryData` is `Read<C<A>>`, where `C` is some generic wrapper component storing a handle to a material, but that'd probably be unnecessarily complex.

crates/bevy_render/src/extract_component.rs
tim-blackbird: I botched the merge a bit :(

PR #15750, opened 2024-10-08
Ensure Bevy's rendering byte usage is little-endian

### LINKED ISSUE #15701: Ensure Bevy's rendering byte usage is little-endian
I got the following response from a wgpu maintainer on Discord:

> fwiw wasm is defined to be little endian, so if you're in the browser it's the browsers job to ensure everything looks like the world is little endian. Furthermore wgsl requires everything to be little-endian. But for all your native host code all bets are off, if you're on  BE system it is indeed your job to handle that

The way I interpret that is that I was originally right and it is indeed our responsibility to provide data to wgpu as little endian, not native endian. So `le_bytes` it is, then.

In this case, it might be nice to make a follow-up PR to fix any remaining uses of `ne_bytes` in Bevy.

_Originally posted by @inodentry in https://github.com/bevyengine/bevy/issues/10392#issuecomment-2396819418_

chompaa:
# Objective

- Fixes (partially) #15701.

## Solution

- Use little-endian bytes over native-endian bytes where applicable.

## Testing

- Ran CI.

## Open Questions

- Should we config-gate these for big-endian targets? It looks like there are [very few targets](https://doc.rust-lang.org/nightly/rustc/platform-support.html) that use big-endian.

### DIFF
--- a/crates/bevy_image/src/hdr_texture_loader.rs
+++ b/crates/bevy_image/src/hdr_texture_loader.rs
@@ -54,10 +54,10 @@ impl AssetLoader for HdrTextureLoader {
         for rgb in image_buffer.pixels() {
             let alpha = 1.0f32;
 
-            rgba_data.extend_from_slice(&rgb.0[0].to_ne_bytes());
-            rgba_data.extend_from_slice(&rgb.0[1].to_ne_bytes());
-            rgba_data.extend_from_slice(&rgb.0[2].to_ne_bytes());
-            rgba_data.extend_from_slice(&alpha.to_ne_bytes());
+            rgba_data.extend_from_slice(&rgb.0[0].to_le_bytes());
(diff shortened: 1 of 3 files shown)


alice-i-cecile:
IMO no cfg gating for now. If users on big-endian platforms ever appear, we can get them to tell us what's broken.

alice-i-cecile (approved):

alice-i-cecile:
@clarfonthey can I get your review here?

clarfonthey (approved):
So, this is a strict improvement, but it doesn't fix the issue completely.

There are plenty of places where `bytemuck` is used to cast slices directly without going through these methods and those need to be fixed too.

So, I'm fine approving, since this is an improvement, but under the condition the original issue stays open. (Right now it would get auto closed by this PR.)

alice-i-cecile:
Edited the PR description to ensure the linked issue stays open :)

PR #15754, opened 2024-10-08
Rename `App/World::observe` to `add_observer`, `EntityWorldMut::observe_entity` to `observe`.

### LINKED ISSUE #15752: `observe_entity` on `EntityWorldMut` is confusing/inconsistent
https://github.com/bevyengine/bevy/pull/15616 changed `observe` to `observe_entity` to avoid scenarios where users would chain `world.observe(..).observe(..)` and unintentionally end up observing their observer. This is because `world.observe` is really more like `world.spawn_observer` than the similarly named `app.observe` which just returns `App`.

However, the `observe_entity` name is a bit confusing or unintuitive in the case of `world.spawn(..).observe_entity(..)`, since it seems expected that after spawning an entity you'd observe that entity. The `_entity` suffix adds additional information here that could imply it's doing something other than the obvious, i.e. adds verbosity that actually reduces clarity in the name of preventing other confusion.

Possible solutions:
- Rename `observe` on `App` and `World` to something like `add_observer`.
- Have `observe` return `&mut World` instead of `EntityWorldMut`.

ItsDoot:
# Objective

- Closes #15752

Calling the functions `App::observe` and `World::observe` doesn't make sense because you're not "observing" the `App` or `World`, you're adding an observer that listens for an event that occurs *within* the `World`. We should rename them to better fit this.

## Solution

Renames:
- `App::observe` -> `App::add_observer`
- `World::observe` -> `World::add_observer`
- `Commands::observe` -> `Commands::add_observer`
- `EntityWorldMut::observe_entity` -> `EntityWorldMut::observe`

(Note this isn't a breaking change as the original rename was introduced earlier this cycle.)

## Testing

Reusing current tests.

### DIFF
--- a/benches/benches/bevy_ecs/observers/propagation.rs
+++ b/benches/benches/bevy_ecs/observers/propagation.rs
@@ -106,15 +106,15 @@ fn add_listeners_to_hierarchy<const DENSITY: usize, const N: usize>(
     world: &mut World,
 ) {
     for e in roots.iter() {
-        world.entity_mut(*e).observe_entity(empty_listener::<N>);
+        world.entity_mut(*e).observe(empty_listener::<N>);
     }
     for e in leaves.iter() {
-        world.entity_mut(*e).observe_entity(empty_listener::<N>);
+        world.entity_mut(*e).observe(empty_listener::<N>);
     }
     let mut rng = deterministic_rand();
     for e in nodes.iter() {
         if rng.gen_bool(DENSITY as f64 / 100.0) {
-            world.entity_mut(*e).observe_entity(empty_listener::<N>);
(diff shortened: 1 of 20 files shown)


alice-i-cecile (approved):
I like this rename better :) It would be nice to have a `commands.add_observer` API too, but that can wait for follow-up.

tychedelia (approved):
Thanks! This definitely reads nicer to me. 👍

LiamGallagher737 (approved):

Shatur:
The PR is missing the migration section, which prevents it from appearing in [the migration guide](https://bevyengine.org/learn/migration-guides/0-14-to-0-15).

PR #15524, opened 2024-09-29
Migrate meshes and materials to required components

Jondolf:
# Objective

A big step in the migration to required components: meshes and materials!

## Solution

As per the [selected proposal](https://hackmd.io/@bevy/required_components/%2Fj9-PnF-2QKK0on1KQ29UWQ):

- Deprecate `MaterialMesh2dBundle`, `MaterialMeshBundle`, and `PbrBundle`.
- Add `Mesh2d` and `Mesh3d` components, which wrap a `Handle<Mesh>`.
- Add `MeshMaterial2d<M: Material2d>` and `MeshMaterial3d<M: Material>`, which wrap a `Handle<M>`.
- Meshes *without* a mesh material should be rendered with a default material. The existence of a material is determined by `HasMaterial2d`/`HasMaterial3d`, which is required by `MeshMaterial2d`/`MeshMaterial3d`. This gets around problems with the generics.

Previously:

```rust
commands.spawn(MaterialMesh2dBundle {
    mesh: meshes.add(Circle::new(100.0)).into(),
    material: materials.add(Color::srgb(7.5, 0.0, 7.5)),
    transform: Transform::from_translation(Vec3::new(-200., 0., 0.)),
    ..default()
});
```

Now:

```rust
commands.spawn((
    Mesh2d(meshes.add(Circle::new(100.0))),
    MeshMaterial2d(materials.add(Color::srgb(7.5, 0.0, 7.5))),
    Transform::from_translation(Vec3::new(-200., 0., 0.)),
));
```

If the mesh material is missing, previously nothing was rendered. Now, it renders a white default `ColorMaterial` in 2D and a `StandardMaterial` in 3D (this can be overridden). Below, only every other entity has a material:

Näyttökuva 2024-09-29 181746

Näyttökuva 2024-09-29 181918

Why white? This is still open for discussion, but I think white makes sense for a *default* material, while *invalid* asset handles pointing to nothing should have something like a pink material to indicate that something is broken (I don't handle that in this PR yet). This is kind of a mix of Godot and Unity: Godot just renders a white material for non-existent materials, while Unity renders nothing when no materials exist, but renders pink for invalid materials. I can also change the default material to pink if that is preferable though.

## Testing

I ran some 2D and 3D examples to test if anything changed visually. I have not tested all examples or features yet however. If anyone wants to test more extensively, it would be appreciated!

## Implementation Notes

[...]

PR #15755, opened 2024-10-09
Fix meshlet materials

Jondolf:
# Objective

After #15524, there are these bunny-shaped holes in rendering in the meshlet example!

broken

This is because (1) they're using a raw asset handle instead of `MeshMaterial3d`, and (2) the system that extracts mesh materials into the render world has an unnecessary `With<Mesh3d>` filter, which makes it not account for meshlets.

## Solution

Remove the redundant filter and use `MeshMaterial3d`. The bunnies got some paint!

fixed

### DIFF
--- a/examples/3d/meshlet.rs
+++ b/examples/3d/meshlet.rs
@@ -86,18 +86,20 @@ fn setup(
     for x in -2..=2 {
         commands.spawn(MaterialMeshletMeshBundle {
             meshlet_mesh: meshlet_mesh_handle.clone(),
-            material: standard_materials.add(StandardMaterial {
-                base_color: match x {
-                    -2 => Srgba::hex("#dc2626").unwrap().into(),
-                    -1 => Srgba::hex("#ea580c").unwrap().into(),
-                    0 => Srgba::hex("#facc15").unwrap().into(),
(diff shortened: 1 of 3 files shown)


ricky26 (approved):

examples/3d/meshlet.rs
```diff
@@ -86,18 +86,20 @@ fn setup(
...
-                    -1 => Srgba::hex("#ea580c").unwrap().into(),
-                    0 => Srgba::hex("#facc15").unwrap().into(),
-                    1 => Srgba::hex("#16a34a").unwrap().into(),
-                    2 => Srgba::hex("#0284c7").unwrap().into(),
-                    _ => unreachable!(),
-                },
-                perceptual_roughness: (x + 2) as f32 / 4.0,
-                ..default()
-            }),
+            material: standard_materials
```
JMS55, line 89: Nit: This looks better wrapping it in MeshMaterial3d() rather than using into()

JMS55 (approved):

JMS55:
We also need to make sure `ResourceManager::previous_depth_pyramids` is working properly. It should not be recreating a texture every frame for each views.

Issue #15460, opened 2024-09-26
Tracking Issue: MVP `no_std` Bevy


bushrat011899:
This is a tracking issue for progress on a `no_std` compatible subset of `Bevy`. The tasks are ordered roughly in sequence (`bevy_app` can't be `no_std` until `bevy_ecs` is for example).

# Overview

Below is a visualisation of the crates planned to be ported to `no_std`. Edges indicate the primary blocker for a crate being ported. The 🚧 symbol indicates that this crate is in-progress, while ✅ indicates it is now available in `no_std`.

```mermaid
graph TD;
    bevy_a11y[bevy_a11y ✅]
    bevy_app[bevy_app ✅]
    bevy_asset
    bevy_color[bevy_color ✅]
    bevy_core[bevy_core* ✅]
    bevy_diagnostic[bevy_diagnostic ✅]
    bevy_ecs[bevy_ecs ✅]
    bevy_hierarchy[bevy_hierarchy* ✅]
    bevy_input[bevy_input ✅]
    bevy_input_focus[bevy_input_focus ✅]
    bevy_math[bevy_math ✅]
    bevy_mikktspace[bevy_mikktspace ✅]
    bevy_ptr[bevy_ptr ✅]
    bevy_reflect[bevy_reflect✅]
    bevy_state[bevy_state ✅]
    bevy_tasks[bevy_tasks ✅]
    bevy_time[bevy_time ✅]
    bevy_transform[bevy_transform ✅]
    bevy_utils[bevy_utils ✅]
    bevy_window[bevy_window ✅]

    bevy_ptr-->bevy_ecs;

    bevy_utils-->bevy_ecs;

    bevy_tasks-->bevy_ecs;

    bevy_ecs-->bevy_app;

    bevy_app-->bevy_core;
    bevy_app-->bevy_a11y;
    bevy_app-->bevy_asset;
    bevy_app-->bevy_time;
    bevy_app-->bevy_log;

    bevy_core-->bevy_input;
    bevy_core-->bevy_diagnostic;
    bevy_core-->bevy_hierarchy;
    
    bevy_asset-->bevy_image;
    bevy_asset-->bevy_scene;

    bevy_color-->bevy_image;

    bevy_mikktspace-->bevy_mesh;

    bevy_hierarchy-->bevy_remote;
    bevy_hierarchy-->bevy_state;
    bevy_hierarchy-->bevy_transform;
    bevy_hierarchy-->bevy_input_focus;

[...]

PR #15772, opened 2024-10-09
Remove `thiserror` from `bevy_gltf`

bushrat011899:
# Objective

- Contributes to #15460

## Solution

- Removed `thiserror` from `bevy_gltf`

### DIFF
--- a/crates/bevy_gltf/src/vertex_attributes.rs
+++ b/crates/bevy_gltf/src/vertex_attributes.rs
@@ -4,11 +4,11 @@ use bevy_render::{
     render_resource::VertexFormat,
 };
 use bevy_utils::HashMap;
+use derive_more::derive::{Display, Error};
 use gltf::{
     accessor::{DataType, Dimensions},
     mesh::util::{ReadColors, ReadJoints, ReadTexCoords, ReadWeights},
 };
-use thiserror::Error;
 
 /// Represents whether integer data requires normalization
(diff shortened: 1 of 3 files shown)


crates/bevy_gltf/src/vertex_attributes.rs
```diff
@@ -241,13 +241,16 @@ enum ConversionMode {
...
+#[error(ignore)]
 pub(crate) enum ConvertAttributeError {
-    #[error("Vertex attribute {0} has format {1:?} but expected {3:?} for target attribute {2}")]
+    #[display(
+        "Vertex attribute {_0} has format {_1:?} but expected {_3:?} for target attribute {_2}"
+    )]
     WrongFormat(String, VertexFormat, String, VertexFormat),
-    #[error("{0} in accessor {1}")]
+    #[display("{0} in accessor {_1}")]
     AccessFailed(AccessFailed, usize),
```
dtolnay, line 252: This is not going to do what is intended in this code. This is going to print as something like "99 in accessor 99" because derive_more's expansion of this looks like `write!(formatter, "{0} in accessor {_1}", _1=self.1)` in which `{0}` refers to the `_1` named argument.

bushrat011899: Interesting! I'll look into that and make the appropriate fixes, I suspect this might've occurred in a few places.

PR #15777, opened 2024-10-09
Remove `thiserror` from `bevy_color`

bushrat011899:
# Objective

- Contributes to #15460

## Solution

- Removed `thiserror` from `bevy_color`

### DIFF
--- a/crates/bevy_color/src/color.rs
+++ b/crates/bevy_color/src/color.rs
@@ -4,6 +4,7 @@ use crate::{
 };
 #[cfg(feature = "bevy_reflect")]
 use bevy_reflect::prelude::*;
+use derive_more::derive::From;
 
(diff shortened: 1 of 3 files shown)


crates/bevy_color/src/color.rs
```diff
@@ -40,7 +41,7 @@ use bevy_reflect::prelude::*;
 /// due to its perceptual uniformity and broad support for Bevy's color operations.
 /// To avoid the cost of repeated conversion, and ensure consistent results where that is desired,
 /// first convert this [`Color`] into your desired color space.
-#[derive(Debug, Clone, Copy, PartialEq)]
+#[derive(Debug, Clone, Copy, PartialEq, From)]
```
alice-i-cecile, line 44: 😍

PR #15793, opened 2024-10-09
Add NoDefaultMaterial3d

IceSentry:
# Objective

- Spawning a Mesh3d without a material will automatically add a default material. This isn't always a wanted behaviour.

## Solution

- Add a marker component to not automatically add a default material. I'd prefer if the default material feature just wasn't a thing
- We can't just manually add the `HasMaterial3d` component because the engine uses that to fined entities with a material and the point is that these meshes don't have materials.

## Testing

- Did you test these changes? 
	- I modified the 3d_scene example to use the new component and the cube disappeared as expected.

### DIFF
--- a/crates/bevy_pbr/src/material.rs
+++ b/crates/bevy_pbr/src/material.rs
@@ -559,10 +559,22 @@ fn extract_mesh_materials<M: Material>(
     }
 }
 
+#[derive(Component)]
+pub struct NoDefaultMaterial3d;
+
 /// Extracts default materials for 3D meshes with no [`MeshMaterial3d`].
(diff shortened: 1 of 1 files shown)


alice-i-cecile:
We [discussed this on Discord](https://discord.com/channels/691052431525675048/743663924229963868/1293623090491949251), but I'm in favor of reverting the default material completely. I don't think it's wise to introduce these semantic changes as part of a complex migration, and this is a surprising automatic behavior.

cart:
Yup lets close this in favor of a full remove of the default behavior (including the HasMaterial3d/2d components)

PR #15795, opened 2024-10-09
promote atlas sources texture_ids to pub visibility

brandon-reinhart:
In order to create texture atlases from other systems (custom game solutions) that are compatible with the ones generated by the bevy builders, it would be nice to have the interface be fully public. This field is pub(crate). Unless there's a good reason, can we promote this to pub?

Alternatives:
- Don't do it.


alice-i-cecile (approved):

alice-i-cecile:
I think this is a reasonable escape hatch.

PR #15796, opened 2024-10-09
Remove the `Component` trait implementation from `Handle`

### LINKED ISSUE #15716: Remove `Component` trait impl from `Handle<T>`

### LINKED ISSUE #15718: Stop using `Handle<MeshletMesh>` as a component
See #15716

tim-blackbird:
# Objective

- Closes #15716
- Closes #15718

## Solution

- Replace `Handle<MeshletMesh>` with a new `MeshletMesh3d` component
- As expected there were some random things that needed fixing:
  - A couple tests were storing handles just to prevent them from being dropped I believe, which seems to have been unnecessary in some.
  - The `SpriteBundle` still had a `Handle<Image>` field. I've removed this.
  - Tests in `bevy_sprite` incorrectly added a `Handle<Image>` field outside of the `Sprite` component.
  - A few examples were still inserting `Handle`s, switched those to their corresponding wrappers.
  - 2 examples that were still querying for `Handle<Image>` were changed to query `Sprite`

## Testing

- I've verified that the changed examples work now

## Migration Guide

`Handle` can no longer be used as a `Component`. All existing Bevy types using this pattern have been wrapped in their own semantically meaningful type. You should do the same for any custom `Handle` components your project needs.

The `Handle<MeshletMesh>` component is now `MeshletMesh3d`.

The `WithMeshletMesh` type alias has been removed. Use `With<MeshletMesh3d>` instead.

### DIFF
--- a/crates/bevy_animation/src/animation_event.rs
+++ b/crates/bevy_animation/src/animation_event.rs
@@ -58,7 +58,7 @@ pub(crate) fn trigger_animation_event(
 ///     let mut player = AnimationPlayer::default();
 ///     player.play(animation_index).repeat();
 ///     
-///     commands.spawn((graphs.add(graph), player));
+///     commands.spawn((AnimationGraphHandle(graphs.add(graph)), player));
 /// }
 /// #
 /// # bevy_ecs::system::assert_is_system(setup_animation);
--- a/crates/bevy_asset/src/handle.rs
+++ b/crates/bevy_asset/src/handle.rs
@@ -3,7 +3,6 @@ use crate::{
     UntypedAssetId,
 };
 use alloc::sync::Arc;
-use bevy_ecs::prelude::*;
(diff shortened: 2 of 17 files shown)


alice-i-cecile:
I've extended the migration guide for you :)

NiseVoid (approved):

JMS55:
Thanks for doing this! Small nit, can we remove the WithMeshletMesh type alias? I never liked it to begin with, and it's not even that complex now that handle is gone. Feels uneeded.

alice-i-cecile (approved):

PR #15797, opened 2024-10-09
Rename TextBlock to TextLayout

UkoeHB:
# Objective

- Improve clarity when spawning a text block. See [this discussion](https://github.com/bevyengine/bevy/pull/15591/#discussion_r1787083571).

## Solution

- Rename `TextBlock` to `TextLayout`.


alice-i-cecile (approved):
I prefer this naming. I think this is much clearer: it's more adjective-y, rather than being a distinct noun.

cart:
[From my perspective](https://github.com/bevyengine/bevy/pull/15591/#discussion_r1787003818) this should still be a "driving concept" with a noun. `TextLayout` feels like it is "just" describing the text layout configuration. But imo we can (and should) make this the driving concept for "thing that receives text spans and renders them". `TextBlock` feels like the better name in that context.

UkoeHB:
Right now the requires are
- `Text -> {TextStyle, TextBlock -> {ComputedTextBlock, TextLayoutInfo}}`.

What if we rearrange like this
- `Text -> {TextStyle, ComputedTextBlock -> {TextLayout, TextLayoutInfo}}` or
- `Text -> {TextStyle, TextLayout, ComputedTextBlock -> {TextLayout, TextLayoutInfo}}`?

cart:
Hmm that would have conceptual / ergonomic implications when trying to use the "TextBlock" concept  (in your proposal `ComputedTextBlock`) directly:

```rust
commands.spawn((Node, ComputedTextBlock, TextLayout::new_with_justify(JustifyText::Center))
```

Vs

```rust
commands.spawn((Node, TextBlock::new_with_justify(JustifyText::Center))
```

That being said, with the current impl this standalone usage is not currently possible, and I'm not _really_ sure we even need it.

In the short term I'm cool with TextLayout, provided it is used "just" for config. I agree that the introduction of a "non layout" concept does muddle things when that new concept isn't currently actually _used_ anywhere outside of layout. If we later decide that TextBlock is a useful user facing concept, we can re-open the discussion.

cart (approved):

PR #15800, opened 2024-10-09
Add mesh picking backend and `MeshRayCast` system parameter

### LINKED ISSUE #15545: Add a mesh picking backend for `bevy_picking`
This is very useful for getting basic 3D picking functionality working in games, examples and for debuggging without needing to setup colliders.

`bevy_mod_raycast` is a solid existing implementation. @Jondolf has already started on this.

Jondolf:
# Objective

Closes #15545.

`bevy_picking` supports UI and sprite picking, but not mesh picking. Being able to pick meshes would be extremely useful for various games, tools, and our own examples, as well as scene editors and inspectors. So, we need a mesh picking backend!

Luckily, [`bevy_mod_picking`](https://github.com/aevyrie/bevy_mod_picking) (which `bevy_picking` is based on) by @aevyrie already has a [backend for it](https://github.com/aevyrie/bevy_mod_picking/blob/74f0c3c0fbc8048632ba46fd8f14e26aaea9c76c/backends/bevy_picking_raycast/src/lib.rs) using [`bevy_mod_raycast`](https://github.com/aevyrie/bevy_mod_raycast). As a side product of adding mesh picking, we also get support for performing ray casts on meshes!

## Solution

Upstream a large chunk of the immediate-mode ray casting functionality from `bevy_mod_raycast`, and add a mesh picking backend based on `bevy_mod_picking`. Huge thanks to @aevyrie who did all the hard work on these incredible crates!

All meshes are pickable by default. Picking can be disabled for individual entities by adding `PickingBehavior::IGNORE`, like normal. Or, if you want mesh picking to be entirely opt-in, you can set `MeshPickingBackendSettings::require_markers` to `true` and add a `RayCastPickable` component to the desired camera and target entities.

You can also use the new `MeshRayCast` system parameter to cast rays into the world manually:

```rust
fn ray_cast_system(mut ray_cast: MeshRayCast, foo_query: Query<(), With<Foo>>) {
    let ray = Ray3d::new(Vec3::ZERO, Dir3::X);

    // Only ray cast against entities with the `Foo` component.
    let filter = |entity| foo_query.contains(entity);

    // Never early-exit. Note that you can change behavior per-entity.
    let early_exit_test = |_entity| false;

    // Ignore the visibility of entities. This allows ray casting hidden entities.
    let visibility = RayCastVisibility::Any;

    let settings = RayCastSettings::default()
        .with_filter(&filter)
        .with_early_exit_test(&early_exit_test)
        .with_visibility(visibility);

    // Cast the ray with the settings, returning a list of intersections.
    let hits = ray_cast.cast_ray(ray, &settings);
}
```

This is largely a direct port, but I did make several changes to match our APIs better, remove things we don't need or that I think are unnecessary, and do some general improvements to code quality and documentation.

### Changes Relative to `bevy_mod_raycast` and `bevy_mod_picking`

- Every `Raycast` and "raycast" has been renamed to `RayCast` and "ray cast" (similar reasoning as the "Naming" section in #15724)
- `Raycast` system param has been renamed to `MeshRayCast` to avoid naming conflicts and to be explicit that it is not for colliders
- `RaycastBackend` has been renamed to `MeshPickingBackend`
- `RayCastVisibility` variants are now `Any`, `Visible`, and `VisibleInView` instead of `Ignore`, `MustBeVisible`, and `MustBeVisibleAndInView`
- `NoBackfaceCulling` has been renamed to `RayCastBackfaces`, to avoid implying that it affects the rendering of backfaces for meshes (it doesn't)
- `SimplifiedMesh` and `RayCastBackfaces` live near other ray casting API types, not in their own 10 LoC module
- All intersection logic and types are in the same `intersections` module, not split across several modules
- Some intersection types have been renamed to be clearer and more consistent
	- `IntersectionData` -> `RayMeshHit` 
	- `RayHit` -> `RayTriangleHit`
- General documentation and code quality improvements

### Removed / Not Ported

- Removed unused ray helpers and types, like `PrimitiveIntersection`
- Removed getters on intersection types, and made their properties public
- There is no `2d` feature, and `Raycast::mesh_query` and `Raycast::mesh2d_query` have been merged into `MeshRayCast::mesh_query`, which handles both 2D and 3D
	- I assume this existed previously because `Mesh2dHandle` used to be in `bevy_sprite`. Now both the 2D and 3D mesh are in `bevy_render`.
- There is no `debug` feature or ray debug rendering
- There is no deferred API (`RaycastSource`)
- There is no `CursorRayPlugin` (the picking backend handles this)

### Note for Reviewers

In case it's helpful, the [first commit](https://github.com/bevyengine/bevy/commit/281638ef10ebd25f783bee2533509fdf756f7ab4) here is essentially a one-to-one port. The rest of the commits are primarily refactoring and cleaning things up in the ways listed earlier, as well as changes to the module structure.

It may also be useful to compare the original [picking backend](https://github.com/aevyrie/bevy_mod_picking/blob/74f0c3c0fbc8048632ba46fd8f14e26aaea9c76c/backends/bevy_picking_raycast/src/lib.rs) and [`bevy_mod_raycast`](https://github.com/aevyrie/bevy_mod_raycast) to this PR. Feel free to mention if there are any changes that I should revert or something I should not include in this PR.

## Testing

I tested mesh picking and relevant components in some examples, for both 2D and 3D meshes, and added a new `mesh_picking` example. I also ~~stole~~ ported over the [ray-mesh intersection benchmark](https://github.com/aevyrie/bevy_mod_raycast/blob/dbc5ef32fe48997a1a7eeec7434d9dd8b829e52e/benches/ray_mesh_intersection.rs) from `bevy_mod_raycast`.

---

## Showcase

Below is a version of the `2d_shapes` example modified to demonstrate 2D mesh picking. This is not included in this PR.

https://github.com/user-attachments/assets/7742528c-8630-4c00-bacd-81576ac432bf

And below is the new `mesh_picking` example:

https://github.com/user-attachments/assets/b65c7a5a-fa3a-4c2d-8bbd-e7a2c772986e

There is also a really cool new `mesh_ray_cast` example ported over from `bevy_mod_raycast`:

https://github.com/user-attachments/assets/3c5eb6c0-bd94-4fb0-bec6-8a85668a06c9


aevyrie (approved):
This is a bit of a drive-by review, but the summary of changes look great. Lots of things I haven't had the energy to clean up. :)

Thanks for your effort!

aevyrie:
It might be nice to include one of the raycasting examples. I'm partial to the laser one, because it looks cool, and shows how you can use the result of one raycast to create another one.

alice-i-cecile:
Release note candidate right there :D Seems useful too!

Jondolf:
Agreed, I'll add that! I've always though that example looks super cool :)

(And I'll fix CI...)

Jondolf:
Added the example, you can see it in the PR description, I originally added a way to hide the red automatic ray by pressing Space, but I felt like it looks nicer without the instruction text, so I kept it simple for now :P

Also got to use Glam's [`Vec3::reflect`](https://docs.rs/glam/latest/glam/f32/struct.Vec3.html#method.reflect) for the first time! Definitely not confusingly named for us...

crates/bevy_picking/src/mesh_picking/ray_cast/intersections.rs
```diff
@@ -0,0 +1,405 @@
...
+                Vec3A::from(vertex_positions[i + 1]),
+                Vec3A::from(vertex_positions[i + 2]),
+            ];
+            let tri_normals = vertex_normals.map(|normals| {
+                [
+                    Vec3A::from(normals[i]),
+                    Vec3A::from(normals[i + 1]),
+                    Vec3A::from(normals[i + 2]),
+                ]
+            });
```
tbillington, line 204: Would prefer this use `.chunks_exact(3).enumerate()` for the same reason as above.

Also, there's no guarantee there will be normals at the given index, and it will panic if they aren't there, so i'd recommend this code not use raw indexing.

crates/bevy_picking/src/mesh_picking/ray_cast/intersections.rs
```diff
@@ -0,0 +1,405 @@
...
+        }
+        Backfaces::Include => {
+            // ray and triangle are parallel if det is close to 0
+            if determinant.abs() < f32::EPSILON {
+                return None;
+            }
+        }
+    }
+
+    let determinant_inverse = 1.0 / determinant;
```
tbillington, line 307: is it possible `determinant` is ever 0?

crates/bevy_picking/src/mesh_picking/ray_cast/mod.rs
```diff
@@ -0,0 +1,306 @@
...
+                        nearest_blocking_hit = distance.min(nearest_blocking_hit);
+                    }
+                    self.hits.push((distance, (*entity, intersection)));
+                };
+            });
+
+        self.hits.retain(|(dist, _)| *dist <= nearest_blocking_hit);
+        self.hits.sort_by_key(|(k, _)| *k);
+        let hits = self.hits.iter().map(|(_, (e, i))| (*e, i.to_owned()));
+        *self.output = hits.collect();
```
tbillington, line 303: This should be `self.output.extend(hits)` https://doc.rust-lang.org/std/iter/trait.Extend.html, assuming you're clearing and keeping around the output Vec to avoid allocations.

Currently it's clearing `self.output` for no reason, since it's a new allocation every time from `hits.collect()`

I like the caching the Vec though to avoid allocations :)

tbillington (commented):
Awesome work :) just a few opportunities to harden the engine against badly formed user data, and perf stuff.

crates/bevy_picking/src/mesh_picking/ray_cast/intersections.rs
```diff
@@ -0,0 +1,405 @@
...
+    }
+}
+
+/// Checks if a ray intersects a mesh, and returns the nearest intersection if one exists.
+pub fn ray_mesh_intersection(
+    ray: Ray3d,
+    mesh_transform: &Mat4,
+    vertex_positions: &[[f32; 3]],
+    vertex_normals: Option<&[[f32; 3]]>,
+    indices: Option<&Vec<impl IntoUsize>>,
```
bushrat011899, line 122:
```suggestion
/// Checks if a ray intersects a mesh, and returns the nearest intersection if one exists.
pub fn ray_mesh_intersection(
    ray: Ray3d,
    mesh_transform: &Mat4,
    vertex_positions: &[[f32; 3]],
    vertex_normals: Option<&[[f32; 3]]>,
    indices: Option<&Vec<impl Into<usize> + Copy>>,
```

The trait `IntoUsize` seems redundant, since `usize` implements `From<u32>` and `From<u16>` anyway.

crates/bevy_picking/src/mesh_picking/ray_cast/intersections.rs
bushrat011899: I believe it could happen if the ray is coplanar with the triangle (passing through the surface of the triangle and multiple points on a line), as that would produce a `p_vec` value which is orthogonal to `vector_v0_to_v2`, making their dot product zero.

crates/bevy_picking/src/mesh_picking/ray_cast/intersections.rs
```diff
@@ -0,0 +1,405 @@
...
+    // The distance between ray origin and intersection is t.
+    let t: f32 = vector_v0_to_v2.dot(q_vec) * determinant_inverse;
+
+    Some(RayTriangleHit {
+        distance: t,
+        barycentric_coords: (u, v),
+    })
+}
+
+// TODO: It'd be nice to use `RayCast3d` from `bevy_math` instead. It caches the direction reciprocal.
```
bushrat011899, line 330: Is there a reason this `TODO` hasn't been acted on?

bushrat011899 (approved):
I'm not familiar with the original `bevy_mod_picking`, but this looks good to me. I do have some comments, but I don't think they should block this from being merged.

crates/bevy_picking/src/mesh_picking/ray_cast/intersections.rs
```diff
@@ -0,0 +1,405 @@
...
+    mesh::{Indices, Mesh, PrimitiveTopology, VertexAttributeValues},
+    primitives::Aabb,
+};
+use bevy_utils::tracing::{error, warn};
+
+use super::Backfaces;
+
+/// A ray intersection with a mesh.
+#[derive(Debug, Clone, Reflect)]
+pub struct RayMeshHit {
```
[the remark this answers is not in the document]
Jondolf: The origin of the ray? You can't compute it from just the information stored here, it'd need the ray direction to compute it based on the point of intersection and distance.

crates/bevy_picking/src/mesh_picking/ray_cast/intersections.rs
```diff
@@ -0,0 +1,405 @@
...
+                triangle_index,
+            });
+            closest_hit_distance = hit.distance;
+        }
+    }
+
+    closest_hit
+}
+
+#[inline(always)]
```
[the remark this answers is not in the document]
Jondolf: Haven't tested here yet (these were in bevy_mod_raycast) but I did notice most of my ray casts in #15724 being *significantly* faster when inlined. I can bench later today though.

crates/bevy_picking/src/mesh_picking/ray_cast/intersections.rs
Jondolf: I think it is possible, but the `!(0.0..=1.0).contains(&u)` check should return early in that case, so it shouldn't return any infinite or NaN values

crates/bevy_picking/src/mesh_picking/ray_cast/intersections.rs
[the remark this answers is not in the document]
Jondolf: Yeah, it's roughly what I expected. Removing the inlining causes a 6-12% regression.
regression
In my experience, more often than not, inlining is beneficial for these types of functions and methods. Obviously not everywhere though.

crates/bevy_picking/src/mesh_picking/ray_cast/intersections.rs
```diff
@@ -0,0 +1,405 @@
...
+            let tri_vertex_positions = [
+                Vec3A::from(vertex_positions[index[0].into_usize()]),
+                Vec3A::from(vertex_positions[index[1].into_usize()]),
+                Vec3A::from(vertex_positions[index[2].into_usize()]),
+            ];
+            let tri_normals = vertex_normals.map(|normals| {
+                [
+                    Vec3A::from(normals[index[0].into_usize()]),
+                    Vec3A::from(normals[index[1].into_usize()]),
+                    Vec3A::from(normals[index[2].into_usize()]),
```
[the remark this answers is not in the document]
Jondolf: Seems to be roughly a 2-12% improvement in the benchmarks depending on if there are intersections. Nice!

crates/bevy_picking/src/mesh_picking/ray_cast/intersections.rs
Jondolf: Using `get` for the normals is a 20-45% perf regression :/ I don't think I'll add that for now, if it ends up being a problem then we can reconsider. bevy_mod_picking and bevy_mod_raycast have been widely used for a long time, and I haven't seen anyone complain about panics.

crates/bevy_picking/src/mesh_picking/ray_cast/intersections.rs
```diff
@@ -0,0 +1,405 @@
...
+        );
+        return None;
+    }
+
+    // Get the vertex positions and normals from the mesh.
+    let vertex_positions: &Vec<[f32; 3]> = match mesh.attribute(Mesh::ATTRIBUTE_POSITION) {
+        None => panic!("Mesh does not contain vertex positions"),
+        Some(vertex_values) => match &vertex_values {
+            VertexAttributeValues::Float32x3(positions) => positions,
+            _ => panic!("Unexpected types in {:?}", Mesh::ATTRIBUTE_POSITION),
```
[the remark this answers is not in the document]
Jondolf: Changed to errors. IIRC Bevy itself has panics for invalid attributes for quite a few mesh methods though.

crates/bevy_picking/src/mesh_picking/ray_cast/intersections.rs
Jondolf: It does implement `From<u16>`, but not `From<u32>`, because that isn't guaranteed to be a lossless conversion since `u32` doesn't fit inside a `usize` on all platforms (like some microcontrollers with a native size of `u16`). I believe `as` simply truncates the number so that it fits in those cases.

It *does* implement `TryFrom<u32>` though, which afaik returns an error when the number doesn't fit. I changed it to use this for now, didn't see any meaningful performance change in benchmarks.

crates/bevy_picking/src/mesh_picking/mod.rs
```diff
@@ -0,0 +1,132 @@
...
+#[reflect(Resource, Default)]
+pub struct MeshPickingBackendSettings {
+    /// When set to `true` ray casting will only happen between cameras and entities marked with
+    /// [`RayCastPickable`]. Off by default.
+    ///
+    /// This setting is provided to give you fine-grained control over which cameras and entities
+    /// should be used by the ray cast backend at runtime.
+    pub require_markers: bool,
+
+    /// When set to [`RayCastVisibility::Any`], hidden meshes can be ray casted against.
```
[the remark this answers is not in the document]
Jondolf: Changed it to
```rust
    /// Determines how mesh picking should consider [`Visibility`]. When set to [`RayCastVisibility::Any`],
    /// ray casts can be performed against both visible and hidden entities.
    ///
    /// Defaults to [`RayCastVisibility::VisibleInView`], only performing picking against visible entities
    /// that are in the view of a camera.
```
Technically it uses `ViewVisibility` and `InheritedVisibility`, so referring to entities with specifically `Visibility::Hidden` wouldn't have been entirely accurate

crates/bevy_picking/src/mesh_picking/mod.rs
```diff
@@ -0,0 +1,132 @@
+//! A [mesh ray casting](ray_cast) backend for [`bevy_picking`](crate).
+//!
+//! If a pointer passes through this camera's render target, it will automatically shoot rays into
+//! the scene and be able to pick meshes.
```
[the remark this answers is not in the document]
Jondolf: Updated this

crates/bevy_picking/src/mesh_picking/mod.rs
```diff
@@ -0,0 +1,132 @@
...
+    picking_cameras: Query<(&Camera, Option<&RayCastPickable>, Option<&RenderLayers>)>,
+    pickables: Query<&PickingBehavior>,
+    marked_targets: Query<&RayCastPickable>,
+    layers: Query<&RenderLayers>,
+    mut ray_cast: MeshRayCast,
+    mut output: EventWriter<PointerHits>,
+) {
+    for (&ray_id, &ray) in ray_map.map().iter() {
+        let Ok((camera, cam_pickable, cam_layers)) = picking_cameras.get(ray_id.camera) else {
+            continue;
```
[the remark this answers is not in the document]
Jondolf: I probably wouldn't want it to spam errors every frame, so it'd need to be tracked in some way so that it only logs it once. Either way, I don't think that error/warning belongs here; it should be handled higher up in the picking logic where the `RayMap` is managed. Otherwise you could get a situation where every backend emits its own duplicate warnings.

I also don't think it's even really possible for rays to exist in the `RayMap` without a camera, since the `RayMap` is cleared and repopulated every frame based on pointer positions, which requires cameras.

crates/bevy_picking/src/mesh_picking/ray_cast/intersections.rs
Jondolf: I realized it probably wouldn't work: you can't use the cached reciprocal because the ray needs to be transformed into model space first. We also can't reuse `RayCast3d::aabb_intersection_at` because it assumes normalized rays, while this method (I think) can have scaled rays because of the transformation.

I changed this implementation now to be otherwise closer to `RayCast3d::aabb_intersection_at` though, since it takes advantage of SIMD better and is a bit faster.

bushrat011899: Ok that makes sense, thanks for clarifying!

crates/bevy_picking/src/mesh_picking/ray_cast/intersections.rs
tbillington: Ah that's annoying, do you remember which benchmark? I might have a play later.

crates/bevy_picking/src/mesh_picking/ray_cast/intersections.rs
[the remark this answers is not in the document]
tbillington: Nice!

crates/bevy_picking/src/mesh_picking/ray_cast/intersections.rs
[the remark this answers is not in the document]
tbillington: Ah, I misread the normal as the direction that the ray hit at.

crates/bevy_picking/src/mesh_picking/ray_cast/intersections.rs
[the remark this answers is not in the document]
tbillington: Hmm, thanks for the tip, if I get some free time I might go looking and see if they can be fixed.

NthTensor:
Reviewing this today. Have we benchmarked the changes from the "Optimize ray-AABB intersection" commit? They look reasonable but I'd like to know how much of an improvement they are before merging.

Jondolf:
@NthTensor Hard to get accurate traces on the overall effect, but with a simple benchmark for that function alone with 100,000 slightly randomized AABBs, I'm getting roughly a 14% improvement.

bench

(note that this bench currently includes creating the AABBs, but that's the same before and after)

I did also briefly profile picking in the Caldera test scene with both the initial (roughly) one-to-one port, and the current version. Couldn't get entirely reliable results, but perf seemed to be basically identical. I think I might've slightly regressed some things (maybe removing `2d` feature and handling both 2D and 3D by default? hard to say) while improving perf elsewhere, which roughly cancels out.

Either way, I don't think performance is really a big concern here. In the Caldera scene, `update_hits` for the mesh picking backend was just 0.34% of total frame time in my test.

trace

NthTensor:
Looks great, that pretty much satisfies me. As a policy, I don't like merging things with the word "optimize" without some discussion of performance.

alice-i-cecile:
Thank you to everyone involved with the authoring or reviewing of this PR! This work is relatively important and needs release notes! Head over to https://github.com/bevyengine/bevy-website/issues/1723 if you'd like to help out.

PR #15801, opened 2024-10-09
Update hashbrown to 0.15

clarfonthey:
Updating dependencies; adopted version of #15696. (Supercedes #15696.)

Long answer: hashbrown is no longer using ahash by default, meaning that we can't use the default-hasher methods with ahasher. So, we have to use the longer-winded versions instead. This takes the opportunity to also switch our default hasher as well, but without actually enabling the default-hasher feature for hashbrown, meaning that we'll be able to change our hasher more easily at the cost of all of these method calls being obnoxious forever.

One large change from 0.15 is that `insert_unique_unchecked` is now `unsafe`, and for cases where unsafe code was denied at the crate level, I replaced it with `insert`.

## Migration Guide

`bevy_utils` has updated its version of `hashbrown` to 0.15 and now defaults to `foldhash` instead of `ahash`. This means that if you've hard-coded your hasher to `bevy_utils::AHasher` or separately used the `ahash` crate in your code, you may need to switch to `foldhash` to ensure that everything works like it does in Bevy.

### DIFF
--- a/crates/bevy_ecs/src/bundle.rs
+++ b/crates/bevy_ecs/src/bundle.rs
@@ -381,7 +381,7 @@ impl BundleInfo {
 
         if deduped.len() != component_ids.len() {
             // TODO: Replace with `Vec::partition_dedup` once https://github.com/rust-lang/rust/issues/54279 is stabilized
-            let mut seen = HashSet::new();
+            let mut seen = <HashSet<_>>::default();
             let mut dups = Vec::new();
             for id in component_ids {
                 if !seen.insert(id) {
@@ -1422,8 +1422,11 @@ impl Bundles {
             .or_insert_with(|| {
                 let (id, storages) =
                     initialize_dynamic_bundle(bundle_infos, components, Vec::from(component_ids));
-                self.dynamic_bundle_storages
-                    .insert_unique_unchecked(id, storages);
+                // SAFETY: The ID always increases when new bundles are added, and so, the ID is unique.
+                unsafe {
+                    self.dynamic_bundle_storages
+                        .insert_unique_unchecked(id, storages);
+                }
                 (component_ids.into(), id)
             });
         *bundle_id
--- a/crates/bevy_ecs/src/schedule/graph/graph_map.rs
+++ b/crates/bevy_ecs/src/schedule/graph/graph_map.rs
@@ -4,10 +4,10 @@
 //!
 //! [`petgraph`]: https://docs.rs/petgraph/0.6.5/petgraph/
 
-use bevy_utils::{hashbrown::HashSet, AHasher};
+use bevy_utils::{hashbrown::HashSet, FixedHasher};
 use core::{
     fmt,
-    hash::{BuildHasher, BuildHasherDefault, Hash},
+    hash::{BuildHasher, Hash},
 };
 use indexmap::IndexMap;
 use smallvec::SmallVec;
@@ -20,13 +20,13 @@ use Direction::{Incoming, Outgoing};
 ///
 /// For example, an edge between *1* and *2* is equivalent to an edge between
 /// *2* and *1*.
-pub type UnGraph<S = BuildHasherDefault<AHasher>> = Graph<false, S>;
+pub type UnGraph<S = FixedHasher> = Graph<false, S>;
 
 /// A `Graph` with directed edges.
 ///
 /// For example, an edge from *1* to *2* is distinct from an edge from *2* to
 /// *1*.
-pub type DiGraph<S = BuildHasherDefault<AHasher>> = Graph<true, S>;
+pub type DiGraph<S = FixedHasher> = Graph<true, S>;
 
 /// `Graph<DIRECTED>` is a graph datastructure using an associative array
 /// of its node weights `NodeId`.
@@ -45,7 +45,7 @@ pub type DiGraph<S = BuildHasherDefault<AHasher>> = Graph<true, S>;
 ///
 /// `Graph` does not allow parallel edges, but self loops are allowed.
 #[derive(Clone)]
-pub struct Graph<const DIRECTED: bool, S = BuildHasherDefault<AHasher>>
+pub struct Graph<const DIRECTED: bool, S = FixedHasher>
 where
     S: BuildHasher,
 {
@@ -63,14 +63,6 @@ impl<const DIRECTED: bool, S> Graph<DIRECTED, S>
 where
     S: BuildHasher,
 {
-    /// Create a new `Graph`
-    pub(crate) fn new() -> Self
-    where
-        S: Default,
-    {
-        Self::default()
-    }
-
     /// Create a new `Graph` with estimated capacity.
     pub(crate) fn with_capacity(nodes: usize, edges: usize) -> Self
     where
@@ -274,7 +266,7 @@ where
     }
 }
 
-impl<S: BuildHasher> Graph<true, S> {
+impl<S: BuildHasher> DiGraph<S> {
     /// Iterate over all *Strongly Connected Components* in this graph.
     pub(crate) fn iter_sccs(&self) -> impl Iterator<Item = SmallVec<[NodeId; 4]>> + '_ {
         super::tarjan_scc::new_tarjan_scc(self)
@@ -408,7 +400,7 @@ mod tests {
     fn node_order_preservation() {
         use NodeId::System;
 
-        let mut graph = Graph::<true>::new();
+        let mut graph = <DiGraph>::default();
 
         graph.add_node(System(1));
         graph.add_node(System(2));
@@ -450,7 +442,7 @@ mod tests {
     fn strongly_connected_components() {
         use NodeId::System;
 
-        let mut graph = Graph::<true>::new();
+        let mut graph = <DiGraph>::default();
 
         graph.add_edge(System(1), System(2));
         graph.add_edge(System(2), System(1));
--- a/crates/bevy_reflect/Cargo.toml
+++ b/crates/bevy_reflect/Cargo.toml
@@ -45,6 +45,9 @@ bevy_utils = { path = "../bevy_utils", version = "0.15.0-dev", default-features 
 ] }
 bevy_ptr = { path = "../bevy_ptr", version = "0.15.0-dev" }
 
+# used by bevy-utils, but it also needs reflect impls
+foldhash = { version = "0.1.3", default-features = false }
+
 # other
 erased-serde = { version = "0.4", default-features = false, features = [
   "alloc",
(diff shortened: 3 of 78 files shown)


mnmaita:
@clarfonthey looks like there's a formatting issue somewhere, could you try running the formatter and push again?

clarfonthey:
Yeah, I was running into some difficulties fixing the lints but will try and fix this today or tomorrow.

clarfonthey:
Rereading the changelog, I misunderstood what hashbrown did and they actually switched to a better hash that we should probably use instead. Gonna probably find a way to factor that in.

crates/bevy_ecs/src/bundle.rs
```diff
@@ -1422,8 +1422,11 @@ impl Bundles {
             .or_insert_with(|| {
                 let (id, storages) =
                     initialize_dynamic_bundle(bundle_infos, components, Vec::from(component_ids));
-                self.dynamic_bundle_storages
-                    .insert_unique_unchecked(id, storages);
+                // SAFETY: We know the ID is unique
```
clarfonthey, line 1425: I'm adding this comment to appease the lint, but I actually don't know if we do know if the ID is unique.

crates/bevy_reflect/Cargo.toml
```diff
@@ -34,6 +34,9 @@ bevy_reflect_derive = { path = "derive", version = "0.15.0-dev" }
 bevy_utils = { path = "../bevy_utils", version = "0.15.0-dev" }
 bevy_ptr = { path = "../bevy_ptr", version = "0.15.0-dev" }
 
+# used by bevy-utils, but it also needs reflect impls
+foldhash = { version = "0.1.3", default-features = false }
+
```
clarfonthey, line 39: This felt like the best approach: rather than re-exporting foldhash in `bevy_utils` and further bloating up that crate, just re-import it here and add impls for it.

crates/bevy_utils/Cargo.toml
```diff
@@ -10,23 +10,18 @@ keywords = ["bevy"]
...
-ahash = { version = "0.8.7", default-features = false, features = [
-  "compile-time-rng",
-] }
+foldhash = { version = "0.1.3", default-features = false }
 tracing = { version = "0.1", default-features = false }
-hashbrown = { version = "0.14.2", default-features = false }
+hashbrown = { version = "0.15.0", features = [
+  "equivalent",
+  "raw-entry",
+], optional = true, default-features = false }
```
clarfonthey, line 24: This is to explicitly exclude `default-hasher` here. It gives us the flexibility to change our hasher whenever since nobody will be relying on the default hasher, even though it *is* a pain to do this.

clarfonthey:
This should be ready for review, modulo whatever tiny nitpicks CI throws at me.

Not sure if the migration guide is adequate: not sure how much we care about bevy_utils' public API, as far as like, whether people should be using it or not.

crates/bevy_reflect/Cargo.toml
bushrat011899: My only critique here would be to make the `foldhash` dependency optional, keeping inline with how other reflection implementations for 3rd party types are done. Not a blocker, just a comment.

bushrat011899 (approved):

hymm:
What is better than foldhash vs hash? Is it really worth taking the extra dependency?

clarfonthey:
> What is better about foldhash vs ahash? Is it really worth taking the extra dependency?

Note: it's not an extra dependency, just a different dependency. Since hashbrown is using it now, we'd actually be taking an extra dependency by continuing to use ahash.

But supposedly foldhash performs better than ahash, which seems pretty worth it.

crates/bevy_reflect/Cargo.toml
clarfonthey: I didn't make it optional only because it isn't actually optional: `bevy_utils` always includes it, so, putting it optional here would just give people the option to accidentally remove impls for it when they need them.

crates/bevy_utils/Cargo.toml
mnmaita: FYI `0.15.0` was yanked so you either want to use `0.15.1` (or `0.15` maybe?).

clarfonthey:
I figure this will be a post-0.15 change, and so will likely have more merge conflicts going forward, but I'm going to try and at least get the current conflicts resolved. Seems to be taking a few tries.

crates/bevy_utils/Cargo.toml
clarfonthey: I actually marked it as 0.15.1 here because 0.15.0 was un-yanked, but may have some advisories filed for it, even though none of them affect us. Don't want to accidentally trigger any lints.

clarfonthey:
(I fixed everything on my side, but an advisory issue is showing up that seems unrelated to this PR.)

BenjaminBrienen:
Fixes https://github.com/bevyengine/bevy/issues/16478

BenjaminBrienen (approved):

clarfonthey:
Rebasing to account for the new graph types in ECS: I think that once someone looks at that version, we should mark this as ready for final review, since I'd like to get this out before any more big PRs if possible.

crates/bevy_ecs/src/schedule/graph/graph_map.rs
```diff
@@ -408,7 +400,7 @@ mod tests {
     fn node_order_preservation() {
         use NodeId::System;
 
-        let mut graph = Graph::<true>::new();
+        let mut graph = <DiGraph>::default();
```
clarfonthey, line 403: Here's some cursed Rust knowledge:

Yes, `<DiGraph>` and `DiGraph` mean different things here. `<DiGraph>` is equivalent to `<DiGraph<FixedHasher>>` (using the defaults) whereas `DiGraph` is equivalent to `<DiGraph<_>>` (inferring the parameter.

crates/bevy_ecs/src/schedule/graph/graph_map.rs
```diff
@@ -20,13 +20,13 @@ use Direction::{Incoming, Outgoing};
 ///
 /// For example, an edge between *1* and *2* is equivalent to an edge between
 /// *2* and *1*.
-pub type UnGraph<S = BuildHasherDefault<AHasher>> = Graph<false, S>;
```
clarfonthey, line 23: I'm not sure if the choice to use `BuildHasherDefault<AHasher>` here was heavily motivated, but it felt prudent to use the same default as everywhere else, in particular, using the chosen fixed hasher.

clarfonthey:
(Apologies for re-requesting reviews from people: I figure that it's better to "dismiss" the old approvals whenever I make changes, but there's no expectation to review if you're busy.)

crates/bevy_ecs/src/schedule/graph/graph_map.rs
bushrat011899: Can confirm I just chose a hasher without particular concern. Consistency is a good reason to change it!

crates/bevy_ecs/src/schedule/graph/graph_map.rs
bushrat011899: Interesting! Is that a deliberate feature of the language or just a quirk in the compiler?

bushrat011899 (approved):
Nice work! As a heads-up, I believe the `no_std` of `bevy_reflect` #16256 will merge soon too, which might add some new conflicts (I don't know for certain, might merge well!). If there are conflicts please feel free to @ me or re-request my review!

clarfonthey:
Since the PR @bushrat011899 mentioned has been merged now, I think it's probably okay to push this through soon. Will wait for CI before I bother people, though.

(Yes, I know that this comment counts as bothering people. Shh)

crates/bevy_ecs/src/bundle.rs
mnmaita: I´ve checked the places where `Bundles::init_dynamic_info` was being used and most of them are internal facing, but I found `EntityWorldMut::insert_by_ids`, `EntityWorldMut::retain` and `EntityWorldMut::clear` which might be public facing I think?

`HashMap::insert_unique_unchecked` says:

```rust
    /// This operation is safe if a key does not exist in the map.
    ///
    /// However, if a key exists in the map already, the behavior is unspecified:
    /// this operation may panic, loop forever, or any following operation with the map
    /// may panic, loop forever or return arbitrary result.
```

So I'm unsure if we should say something like the caller ensures the uniqueness, and even add some notes in the `EntityWorldMut` methods if applicable? I guess I'm thinking out loud. 😅

mnmaita (approved):
This looks good codewise, I'm just leaving a small comment.

crates/bevy_ecs/src/bundle.rs
clarfonthey: So, I'm honestly fine just changing this to `insert` and leaving it at that :p

bushrat011899 (approved):
Nice work managing this change!

crates/bevy_ecs/src/bundle.rs
clarfonthey: Okay, so, I double-checked and the bundle ID *is* always increasing here, and without concurrency, we don't need to worry about data races. So, I updated the comment to clarify it is actually unique and leaving it.

clarfonthey:
Final changes made:

* Update "ID is unique" comment to actually clarify why ID is unique
* Normalise `let x: HashMap<_, _> = HashMap::` to `let x = <HashMap<_, _>>::` and similarly for `HashSet`, to make things less verbose

clarfonthey:
(If someone could mark this as ready for final review, I'd appreciate it. I think this should qualify as ready.)

clarfonthey:
(I'll try and take a look at the failures soon)

BenjaminBrienen (approved):

clarfonthey:
Ah, fun: the singular PR that got in the queue before this one didn't result in any merge conflicts, but it did add more uses of `insert_unique_unchecked` that now need unsafe blocks :(

clarfonthey:
(kind of shocked that `ci compile` didn't catch that, tbh)

clarfonthey:
(Okay, this time I waited until CI passed. Should be ready to merge now. <3)

alice-i-cecile:
Thanks for shepherding this across the finish line <3

clarfonthey:
So, I had no idea about these patches that were being run in CI, but they seem pretty brittle-- it would be much better (IMHO) to use `#[cfg]` or `cfg!` to enable/disable these features than use patches, IMHO, so that at least they would be visible to people searching through the code. That said, I'll try and see what I can do.

clarfonthey:
I also have *genuinely* no idea why that patch failed on this PR. It appears to be referencing an earlier change where the patch should have broken, but it somehow did not trigger until now. And there were no recent changes to the CI job, either. It just makes absolutely no sense why this PR would be the one to trigger it, but at least, it did, and I fixed it.

PR #15804, opened 2024-10-09
Fix mesh flags

ecoskey:
Fixed incorrect mesh flags offsets

## Testing

Ran OIT example since that was the affected flag


crates/bevy_pbr/src/render/mesh.rs
```diff
@@ -1494,8 +1494,8 @@ bitflags::bitflags! {
         const SCREEN_SPACE_REFLECTIONS          = 1 << 16;
         const HAS_PREVIOUS_SKIN                 = 1 << 17;
         const HAS_PREVIOUS_MORPH                = 1 << 18;
-        const OIT_ENABLED                       = 1 << 18;
-        const LAST_FLAG                         = Self::HAS_PREVIOUS_MORPH.bits();
```
alice-i-cecile, line 1498: Naive question: why does this constant exist?

ecoskey: Looking below, apparently it's used to calculate the offset for MSAA_SHIFT_BITS, etc.

JMS55 (approved):

Issue #15508, opened 2024-09-28
regression in GLTF Scene loading


eidloi:
## Bevy version

0.15.0-dev

## What you did

Migrated code base with a SceneView example to 0.15.0-dev.

## What went wrong

`let scene: Handle<Scene> = asset_server.load(asset);` causes Bevy to hang and never load the asset. This is in a regular system with `asset_server: Res<AssetServer>`.
On the previous version (0.14.2), the asset loaded correctly. There is no error message.

## Additional information

Asset in question: https://github.com/UmbraLuminosa/sickle_ui/blob/main/assets/examples/Low_poly_scene.gltf#Scene0

BenjaminBrienen:
Can you `git bisect` the issue?

eidloi:
> Can you `git bisect` the issue?

It is this commit: https://github.com/bevyengine/bevy/commit/f06cd448db9171bda656ded1e63c68581d6e19aa
NOTE: even before this commit there was a 50-50 chance that the asset would not load either. It would not hang Bevy, you would just get a black screen.

eidloi:
The 50/50 render bug seems to be a side effect of https://github.com/bevyengine/bevy/commit/d9527c101c2f49c6884763cce36ea1d27dd6a597

eidloi:
CC @robtfm, @tychedelia ⬆️

eidloi:
To reproduce I created a new project with toml:

```toml
[package]
name = "find_gltf_bug"
version = "0.1.0"
edition = "2021"

[dependencies]
bevy = { git = "https://github.com/bevyengine/bevy.git", rev = "d9527c101c2f49c6884763cce36ea1d27dd6a597", default-features = false, features = [
    "animation",
    "bevy_asset",
    "bevy_color",
    "bevy_gltf",
    "bevy_state",
    "bevy_text",
    "bevy_ui",
    "bevy_winit",
    "bevy_pbr",
    "bevy_scene",
    "bevy_picking",
    "png",
    "tonemapping_luts",
] }

# Enable a small amount of optimization in debug mode
[profile.dev]
opt-level = 1
```
And main.rs:
```rust
use bevy::prelude::*;

fn main() {
    let mut app = App::new();

    app.add_plugins(DefaultPlugins.set(WindowPlugin {
        primary_window: Some(Window {
            title: "Find GLTF Bug!".into(),
            resolution: (1280., 720.).into(),
            ..default()
        }),
        ..default()
    }))
[...]

PR #15509, opened 2024-09-28
bevy_asset: Improve `NestedLoader` API

aecsocket:
# Objective

The `NestedLoader` API as it stands right now is somewhat lacking:

- It consists of several types `NestedLoader`, `UntypedNestedLoader`, `DirectNestedLoader`, and `UntypedDirectNestedLoader`, where a typestate pattern on `NestedLoader` would be make it more obvious what it does, and allow centralising the documentation
- The term "untyped" in the asset loader code is overloaded. It can mean either:
  - we have literally no idea what the type of this asset will be when we load it (I dub this "unknown type")
  - we know what type of asset it will be, but we don't know it statically - we only have a TypeId (I dub this "dynamic type" / "erased")
- There is no way to get an `UntypedHandle` (erased) given a `TypeId`

## Solution

Changes `NestedLoader` into a type-state pattern, adding two type params:
- `T` determines the typing
  - `StaticTyped`, the default, where you pass in `A` statically into `fn load<A>() -> ..`
  - `DynamicTyped`, where you give a `TypeId`, giving you a `UntypedHandle`
  - `UnknownTyped`, where you have literally no idea what type of asset you're loading, giving you a `Handle<LoadedUntypedAsset>`
- `M` determines the "mode" (bikeshedding TBD, I couldn't come up with a better name)
  - `Deferred`, the default, won't load the asset when you call `load`, but it does give you a `Handle` to it (this is nice since it can be a sync fn)
  - `Immediate` will load the asset as soon as you call it, and give you access to it, but you must be in an async context to call it

Changes some naming of internals in `AssetServer` to fit the new definitions of "dynamic type" and "unknown type". Note that I didn't do a full pass over this code to keep the diff small. That can probably be done in a new PR - I think the definiton I laid out of  unknown type vs. erased makes it pretty clear where each one applies.

<details>
<summary>Old issue</summary>

The only real problem I have with this PR is the requirement to pass in `type_name` (from `core::any::type_name`) into Erased. Users might not have that type name, only the ID, and it just seems sort of weird to *have* to give an asset type name. However, the reason we need it is because of this:
```rs
    pub(crate) fn get_or_create_path_handle_erased(
        &mut self,
        path: AssetPath<'static>,
        type_id: TypeId,
        type_name: &str,
        loading_mode: HandleLoadingMode,
        meta_transform: Option<MetaTransform>,
    ) -> (UntypedHandle, bool) {
        let result = self.get_or_create_path_handle_internal(
            path,
            Some(type_id),
[...]

PR #15808, opened 2024-10-10
Fix potential deadlock in `AssetServer` on single-threaded modes.

### LINKED ISSUE #15807: Possible deadlock in AssetServer in single-threaded modes.
## Bevy version

HEAD at `bd0c74644fda1c6cbdd2524c2920e81fce063648`

## What you did

According to the docs at https://github.com/bevyengine/bevy/blob/bd0c74644fda1c6cbdd2524c2920e81fce063648/crates/bevy_asset/src/server/mod.rs#L423C1-L425C21, we're supposed to be dropping the infos guard. However this is no longer true after #15509.

We should either pass the mutex guard to this function, or drop the guard before the function call.

andriyDev:
# Objective

Fixes #15807 

## Solution

We move the guard into this function.

## Testing

N/A, This is just reverting to the old behavior before #15509.

### DIFF
--- a/crates/bevy_asset/src/server/mod.rs
+++ b/crates/bevy_asset/src/server/mod.rs
@@ -36,7 +36,7 @@ use either::Either;
 use futures_lite::{FutureExt, StreamExt};
 use info::*;
 use loaders::*;
-use parking_lot::RwLock;
+use parking_lot::{RwLock, RwLockWriteGuard};
 use std::path::{Path, PathBuf};
 
 /// Loads and tracks the state of [`Asset`] values from a configured [`AssetReader`](crate::io::AssetReader). This can be used to kick off new asset loads and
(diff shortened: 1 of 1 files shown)


andriyDev:
Note an alternative is to do the dropping of the lock outside the function, but that would mean duplicating it twice.

aecsocket (approved):
I think this change makes sense, I don't entirely like passing the `RwLockWriteGuard` across functions but that's fine if this logic is more correct.

JMS55 (approved):
Didn't test, but seems fine to me.

mockersf:
does this fix also https://github.com/bevyengine/bevy/issues/15508?

andriyDev:
> does this fix also #15508?

Yes, this also seems to fix #15508, at least the reproduction case provided (I tested it on HEAD first to make sure it was still broken). I don't know why @alice-i-cecile reopened the issue though, so maybe there's another cause?

PR #15810, opened 2024-10-10
Add `no_std` Support to `bevy_math`

bushrat011899:
# Objective

- Contributes to #15460

## Solution

- Added two new features, `std` (default) and `alloc`, gating `std` and `alloc` behind them respectively.
- Added missing `f32` functions to `std_ops` as required. These `f32` methods have been added to the `clippy.toml` deny list to aid in `no_std` development.

## Testing

- CI
- `cargo clippy -p bevy_math --no-default-features --features libm --target "x86_64-unknown-none"`
- `cargo test -p bevy_math --no-default-features --features libm`
- `cargo test -p bevy_math --no-default-features --features "libm, alloc"`
- `cargo test -p bevy_math --no-default-features --features "libm, alloc, std"`
- `cargo test -p bevy_math --no-default-features --features "std"`

## Notes

The following items require the `alloc` feature to be enabled:

- `CubicBSpline`
- `CubicBezier`
- `CubicCardinalSpline`
- `CubicCurve`
- `CubicGenerator`
- `CubicHermite`
- `CubicNurbs`
- `CyclicCubicGenerator`
- `RationalCurve`
- `RationalGenerator`
- `BoxedPolygon`
- `BoxedPolyline2d`
- `BoxedPolyline3d`
- `SampleCurve`
- `SampleAutoCurve`
- `UnevenSampleCurve`
- `UnevenSampleAutoCurve`
- `EvenCore`
- `UnevenCore`
- `ChunkedUnevenCore`

This requirement could be relaxed in certain cases, but I had erred on the side of gating rather than modifying. Since `no_std` is a new set of platforms we are adding support to, and the `alloc` feature is enabled by default, this is not a breaking change.


crates/bevy_math/Cargo.toml
```diff
@@ -40,7 +40,24 @@ bevy_math = { path = ".", version = "0.15.0-dev", features = ["approx"] }
...
+  "rand_distr?/std",
+]
+alloc = [
+  "itertools/use_alloc",
+  "serde?/alloc",
+  "rand?/alloc",
+  "rand_distr?/alloc",
+]
+
 serialize = ["dep:serde", "glam/serde"]
```
BenjaminBrienen, line 61: Is this a copy paste error?

crates/bevy_math/src/ops.rs
```diff
@@ -227,6 +227,38 @@ mod std_ops {
...
+    ///
+    /// Precision is specified when the `libm` feature is enabled.
+    #[inline(always)]
+    pub fn abs(x: f32) -> f32 {
+        f32::abs(x)
+    }
+
+    /// Returns the square root of a number.
+    ///
+    /// Precision is specified when the `libm` feature is enabled.
```
BenjaminBrienen, line 249: f32::sqrt is deterministic and precise apparently

> Precision
> The result of this operation is guaranteed to be the rounded infinite-precision result. It is specified by IEEE 754 as squareRoot and guaranteed not to change.

Source:
https://doc.rust-lang.org/std/primitive.f32.html#method.sqrt

Double check the other operators if you don't mind

BenjaminBrienen (changes requested):

crates/bevy_math/src/ops.rs
bushrat011899: Good catch! I've updated the descriptions for all the `std` versions of the operations I've added. The `libm` versions I've left as-is, since their precision isn't immediately apparent to me in the `libm` documentation.

crates/bevy_math/Cargo.toml
bushrat011899: Whoops! Missed that one!

BenjaminBrienen (approved):

crates/bevy_math/src/ops.rs
```diff
@@ -442,6 +501,68 @@ mod libm_ops {
...
+    #[inline(always)]
+    pub fn rem_euclid(x: f32, y: f32) -> f32 {
+        let result = libm::remainderf(x, y);
+
+        if result < 0. {
+            result + y
+        } else {
+            result
+        }
+    }
```
bushrat011899, line 517: I noticed when experimenting with the `curve` feature that `libm::remainderf` was returning a negative value for `remainderf(0.99, 1.0)`. I can't find any documentation stating that it _must_ return a non-negative value, so I am going to assume it's intended behaviour. As such, we must check if it is negative and return the next value on the modulus.

mweatherley: It's explained [here](https://sourceware.org/newlib/libm.html#remainder) that `remainderf` produces something in the range `-y/2 .. y/2` :)

mweatherley (changes requested):
A few notes:

1. The purpose of the `ops` module is (or perhaps *was*) to guarantee that alternatives to `std` floating-point functions with unspecified precision were available, in the interest of cross-platform determinism. Since `libm` is `no_std`-compatible, I think this extension of that is pretty reasonable, but this is a break from its original purpose.

2. I have concerns about the ongoing maintenance burden of these changes. For example, we lint against `std` math operations with unspecified precision for the aforementioned determinism reasons; right now, as far as I'm aware, this doesn't extend to the new methods added, so I'm not sure if breaks to `no_std` compatibility are actually discoverable.

3. `cubic_splines` is feature-gated behind `alloc`, but plenty of other areas use `alloc` as well. What is the logic behind this?

crates/bevy_math/src/ops.rs
bushrat011899: Thanks for pointing that out! I might change my shim here to instead always add `x/2.0` to avoid the branch.

bushrat011899:
> 1. The purpose of the `ops` module is (or perhaps *was*) to guarantee that alternatives to `std` floating-point functions with unspecified precision were available, in the interest of cross-platform determinism. Since `libm` is `no_std`-compatible, I think this extension of that is pretty reasonable, but this is a break from its original purpose.

That makes sense, thanks for clarifying the original intent of the module for me! I think to highlight this change in intent I'll update the module documentation to say there are two reasons you'd want to use `std_ops`: determinism, and `no_std` compatibility.

> 2. I have concerns about the ongoing maintenance burden of these changes. For example, we lint against `std` math operations with unspecified precision for the aforementioned determinism reasons; right now, as far as I'm aware, this doesn't extend to the new methods added, so I'm not sure if breaks to `no_std` compatibility are actually discoverable.

I'll look into adding these new methods to the lint as well, with some documentation explaining that it's for `no_std` compatibility. It's not included in this PR, but my intention is to add `no_std` testing to the CI as well. Right now it's sufficient to attempt compilation for a `no_std` target (e.g., `x86_64-unknown-none`), but this is still an open question (which `no_std` platforms are people actually _using_ Bevy on in this hypothetical future? Bare metal x86 is an example, but perhaps not realistic)

> 3. `cubic_splines` is feature-gated behind `alloc`, but plenty of other areas use `alloc` as well. What is the logic behind this?

My reasoning here was I saw a large amount of `alloc` usage contained within this module, so I feature gated the whole lot. I will look into it more closely to see if there are some elements which can still be exported without `alloc`, and if so, move the feature gate inside the module.

crates/bevy_math/src/ops.rs
bushrat011899: I tried adding `y/2.` in all cases but it did not provide results consistent with `f32::rem_euclid`, so I've opted to keep this version instead. It will only be used on platforms without the `std` feature enabled, so there is no change in behaviour for the existing Bevy targets.

bushrat011899:
Added to the 0.16 milestone to indicate this is _not_ intended to be merged during the 0.15 release cycle.

BenjaminBrienen (approved):

Jondolf (approved):
This does increase maintenance burden somewhat / make some things slightly uglier, but I would say it's worth it for `no_std` support, and to move the `no_std` efforts as a whole forward. Changes look good to me, didn't notice anything wrong.

crates/bevy_math/src/bounding/raycast2d.rs
```diff
@@ -224,21 +229,29 @@ mod tests {
...
             ),
         ] {
-            let case = format!(
+            assert!(
+                test.intersects(volume),
                 "Case:\n  Test: {:?}\n  Volume: {:?}\n  Expected distance: {:?}",
-                test, volume, expected_distance
+                test,
+                volume,
+                expected_distance
```
tim-blackbird, line 237: The variables for all the assert messages could be inside the string, like this:
```rs
"Case:\n Test: {test:?}\n Volume: {volume:?}\n Expected distance: {expected_distance:?}"
```

mweatherley (approved):
Just one thing that I found kind of confusing at first glance. 

My main concern with this is just making sure that we have enough tools (in CI or otherwise) to maintain this matrix of features without breaking compilation in some combination.

crates/bevy_math/src/bounding/raycast2d.rs
bushrat011899: Yeah that makes sense, shaves a few lines off the diff too!

crates/bevy_math/src/ops.rs
```diff
@@ -444,12 +447,149 @@ mod libm_ops {
...
 pub use libm_ops::*;
 
 #[cfg(not(feature = "libm"))]
 pub use std_ops::*;
 
+#[cfg(feature = "std")]
+pub use std_ops_for_no_std::*;
+
+#[cfg(all(feature = "libm", not(feature = "std")))]
+pub use libm_ops_for_no_std::*;
```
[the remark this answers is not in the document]
bushrat011899: Yeah I'm not a fan of the name but I'm not sure what else to call them haha! I've added some documentation to help explain. The idea is `std` provides _some_ of the maths operations rather than `core`, so we need to provide a compatibility shim for those operations.

In place of a better name, I've added some documentation to these new modules explaining the issue explicitly.

bushrat011899:
> My main concern with this is just making sure that we have enough tools (in CI or otherwise) to maintain this matrix of features without breaking compilation in some combination.

Agreed. I've added `bevy_math` to `compile-check-no-std` in CI, so the specific feature combination of `--no-default-features` and `--feature libm` will be checked, along with the standard set of features in the rest of CI. I do agree that there is a flag combinatorics problem here that `no_std` exacerbates.

PR #15811, opened 2024-10-10
Add `Trigger::components`, which lists the component targets that were triggered

### LINKED ISSUE #14774: Add a method to Trigger to get a list of the triggered component IDs

`Observer`s are triggered with `Event`s, but can also narrow those events with specific `Component`s. However, there's currently no way to get to the list of components the observer was triggered with, from inside the observer function.

```rust
impl Trigger {
    pub fn components(&self) -> impl Iterator<Item = ComponentId>;
}
```

For now I'm passing the `ComponentId` in the event itself.

## Additional context

Originally brought up [on discord](https://discord.com/channels/691052431525675048/749335865876021248/1273836049696817163).

ItsDoot:
# Objective

- Closes #14774 

## Solution

Added:

```rust
impl<'w, E, B: Bundle> Trigger<'w, E, B> {
    pub fn components(&self) -> &[ComponentId];
}
```

I went with storing it in the trigger as a `SmallVec<[Component; 1]>` because a singular target component will be the most common case, and it remains the same size as `Vec<ComponentId>`.

## Testing

Added a test.

### DIFF
--- a/crates/bevy_ecs/src/observer/mod.rs
+++ b/crates/bevy_ecs/src/observer/mod.rs
@@ -200,13 +200,20 @@ impl ObserverDescriptor {
 #[derive(Debug)]
 pub struct ObserverTrigger {
     /// The [`Entity`] of the observer handling the trigger.
-    observer: Entity,
+    pub observer: Entity,
     /// The [`Event`] the trigger targeted.
-    event_type: ComponentId,
+    pub event_type: ComponentId,
     /// The [`ComponentId`]s the trigger targeted.
     components: SmallVec<[ComponentId; 2]>,
     /// The entity the trigger targeted.
-    entity: Entity,
+    pub entity: Entity,
+}
+
+impl ObserverTrigger {
+    /// Returns the components that the trigger targeted.
+    pub fn components(&self) -> &[ComponentId] {
+        &self.components
+    }
 }
 
 // Map between an observer entity and its runner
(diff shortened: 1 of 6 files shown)


crates/bevy_ecs/src/observer/mod.rs
```diff
@@ -1268,9 +1281,6 @@ mod tests {
 
     #[test]
     fn observer_invalid_params() {
-        #[derive(Event)]
-        struct EventA;
```
ItsDoot, line 1272: I removed these as they are already defined at the top of the tests module.

BenjaminBrienen (approved):

crates/bevy_ecs/src/observer/mod.rs
```diff
@@ -192,13 +200,13 @@ impl ObserverDescriptor {
...
-    pub observer: Entity,
-
-    /// The [`ComponentId`] the trigger targeted.
-    pub event_type: ComponentId,
-
+    observer: Entity,
+    /// The [`Event`] the trigger targeted.
+    event_type: ComponentId,
+    /// The [`ComponentId`]s the trigger targeted.
+    components: SmallVec<[ComponentId; 1]>,
```
bushrat011899, line 207: I believe if we enable the [`union` feature](https://docs.rs/smallvec/latest/smallvec/#union) for `smallvec` (if it isn't already!), you can replace `[ComponentId; 1]` with `[ComponentId; 2]` within the same footprint due to alignment.

bushrat011899 (approved):
Looks good! I have an optimisation comment around `smallvec`, but it's non-blocking, and I'm not certain of it anyway.

PR #15812, opened 2024-10-10
Move `ImageLoader` and `CompressedImageSaver` to `bevy_image`.

andriyDev:
# Objective

This is a follow-up to #15650. While the core `Image` stuff moved from `bevy_render` to `bevy_image`, the `ImageLoader` and the `CompressedImageSaver` remained in `bevy_render`.

## Solution

I moved `ImageLoader` and `CompressedImageSaver` to `bevy_image` and re-exported everything out from `bevy_render`. The second step isn't strictly necessary, but `bevy_render` is already doing this for all the other `bevy_image` types, so I kept it the same for consistency.

Unfortunately I had to give `ImageLoader` a constructor so I can keep the `RenderDevice` stuff in `bevy_render`.

## Testing

It compiles!

## Migration Guide

- `ImageLoader` can no longer be initialized directly through `init_asset_loader`. Now you must use `app.register_asset_loader(ImageLoader::new(supported_compressed_formats))` (check out the implementation of `bevy_render::ImagePlugin`). This only affects you if you are initializing the loader manually and does not affect users of `bevy_render::ImagePlugin`.

## Followup work

- We should be able to move most of the `ImagePlugin` to `bevy_image`. This would likely require an `ImagePlugin` and a `RenderImagePlugin` or something though.

### DIFF
--- a/crates/bevy_render/src/texture/mod.rs
+++ b/crates/bevy_render/src/texture/mod.rs
@@ -10,11 +10,11 @@ pub use crate::render_resource::DefaultImageSampler;
 pub use bevy_image::ExrTextureLoader;
 #[cfg(feature = "hdr")]
 pub use bevy_image::HdrTextureLoader;
-use bevy_image::ImageLoader;
 pub use bevy_image::{
-    BevyDefault, CompressedImageFormats, Image, ImageAddressMode, ImageFilterMode, ImageFormat,
-    ImageSampler, ImageSamplerDescriptor, ImageType, IntoDynamicImageError, TextureError,
-    TextureFormatPixelInfo,
+    BevyDefault, CompressedImageFormats, FileTextureError, Image, ImageAddressMode,
+    ImageFilterMode, ImageFormat, ImageFormatSetting, ImageLoader, ImageLoaderError,
+    ImageLoaderSettings, ImageSampler, ImageSamplerDescriptor, ImageType, IntoDynamicImageError,
+    TextureError, TextureFormatPixelInfo,
 };
 #[cfg(feature = "basis-universal")]
 pub use compressed_image_saver::*;
--- a/crates/bevy_render/src/texture/mod.rs
+++ b/crates/bevy_render/src/texture/mod.rs
@@ -1,5 +1,3 @@
-#[cfg(feature = "basis-universal")]
-mod compressed_image_saver;
 mod fallback_image;
 mod gpu_image;
 mod texture_attachment;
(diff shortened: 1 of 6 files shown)


bushrat011899 (approved):

crates/bevy_render/src/texture/mod.rs
```diff
@@ -12,15 +9,15 @@ pub use bevy_image::ExrTextureLoader;
...
 pub use bevy_image::{
-    BevyDefault, CompressedImageFormats, Image, ImageAddressMode, ImageFilterMode, ImageFormat,
-    ImageSampler, ImageSamplerDescriptor, ImageType, IntoDynamicImageError, TextureError,
-    TextureFormatPixelInfo,
+    BevyDefault, CompressedImageFormats, FileTextureError, Image, ImageAddressMode,
+    ImageFilterMode, ImageFormat, ImageFormatSetting, ImageLoader, ImageLoaderError,
+    ImageLoaderSettings, ImageSampler, ImageSamplerDescriptor, ImageType, IntoDynamicImageError,
+    TextureError, TextureFormatPixelInfo,
 };
 #[cfg(feature = "basis-universal")]
```
atlv24, line 17: it might make sense to to change the bevy_image import to a ::* to remove the basis-universal feature from bevy_render

atlv24 (approved):
nice work!

crates/bevy_render/src/texture/mod.rs
andriyDev: We still need the `basis-universal` feature to know whether we add the `CompressedImageSaver` to `bevy_render::ImagePlugin`, so I think we should hold off (to when/if we move the plugin to `bevy_image::ImagePlugin`)

I feel like we should just remove all these imports from `bevy_render` and have people depend on `bevy_image` directly, but that's another PR (and also might be more controversial since it bloats the migration guide).

BenjaminBrienen (approved):

MiniaczQ (approved):

crates/bevy_render/src/texture/mod.rs
clarfonthey: I also explicitly kept all of the "texture" formats as feature flags on `bevy_render` when I added them all to `bevy_image`, under the assumption that it's okay to have feature gates for specifically texture formats since they might require custom render options, whereas stuff like PNG will always be rendered just as pixels and thus doesn't need its own render flag.

So, it feels like `basis-universal` will always have a render flag because of that.

PR #15814, opened 2024-10-10
Add window drag move and drag resize without decoration example.

### LINKED ISSUE #15734: Add drag move example
> Yeah, I'd like an example here for ease of testing :)
_Originally posted by @alice-i-cecile in https://github.com/bevyengine/bevy/issues/15674#issuecomment-2399780483_

The linked PR has a nice code snippet that can be adapted into an example.

shanecelis:
# Objective

Add an example for the new drag move and drag resize introduced by PR #15674 and fix #15734.

## Solution

I created an example that allows the user to exercise drag move and drag resize separately. The user can also choose what direction the resize works in.

Screenshot 2024-10-10 at 4 06 43 AM

### Name

The example is called `window_drag_move`. Happy to have that bikeshedded.

### Contentious Refactor?

This PR removed the `ResizeDirection` enumeration in favor of using `CompassOctant` which had the same variants. Perhaps this is contentious.

### Unsafe?

In PR #15674 I mentioned that `start_drag_move()` and `start_drag_resize()`'s requirement to only be called in the presence of a left-click looks like a compiler-unenforceable contract that can cause intermittent panics when not observed, so perhaps the functions should be marked them unsafe. **I have not made that change** here since I didn't see a clear consensus on that.

## Testing

I exercised this on x86 macOS. However, winit for macOS does not support drag resize. It reports a good error when `start_drag_resize()` is called. I'd like to see it tested on Windows and Linux.

---

## Showcase

Example window_drag_move shows how to drag or resize a window without decoration.


IsseW:
> @IsseW can I get your review here?

I can review tomorrow!

shanecelis:
Re: unsafe

I've submitted a [PR to winit](https://github.com/rust-windowing/winit/pull/3944) so it may not panic in the future when a left-click event isn't present, which I'd then be satisfied without any unsafe markers on the functions.

Cargo.toml
```diff
@@ -2934,6 +2934,7 @@ name = "window_fallthrough"
 path = "examples/ui/window_fallthrough.rs"
 doc-scrape-examples = true
 
+
```
IsseW, line 2937: Was this added by mistake?

crates/bevy_window/src/window.rs
```diff
@@ -379,7 +379,7 @@ impl Window {
     ///
     /// There is no guarantee that this will work unless the left mouse button was
     /// pressed immediately before this function was called.
-    pub fn start_drag_resize(&mut self, direction: ResizeDirection) {
+    pub fn start_drag_resize(&mut self, direction: CompassOctant) {
```
IsseW, line 382: I like this idea of using the existing `CompassOctant` :smile:

IsseW (approved):
I think this looks good, just one comment about a newline.

Cargo.toml
shanecelis: Yes, that was a mistake. Will fix.

examples/window/window_drag_move.rs
```diff
@@ -0,0 +1,140 @@
...
+        *writer.text(example_text.single(), 7) = format!("{:?}", DIRECTIONS[dir.0]);
+    }
+
+    if input.just_pressed(KeyCode::KeyD) {
+        dir.0 = (dir.0 + 1) % DIRECTIONS.len();
+        *writer.text(example_text.single(), 7) = format!("{:?}", DIRECTIONS[dir.0]);
+    }
+}
+
+fn move_windows(
```
MiniaczQ, line 118:
```suggestion
fn drag_or_resize_window(
```
I feel like move only refers to drag

examples/window/window_drag_move.rs
```diff
@@ -0,0 +1,140 @@
...
+) {
+    // Both `start_drag_move()` and `start_drag_resize()` must be called after a
+    // left mouse button press as done here.
+    //
+    // winit 0.30.5 may panic when initiated without a left mouse button press.
+    if input.just_pressed(MouseButton::Left) {
+        for mut window in windows.iter_mut() {
+            match *action {
+                LeftClickAction::Nothing => (),
+                LeftClickAction::Drag => window.start_drag_move(),
```
MiniaczQ, line 132:
```suggestion
                LeftClickAction::Move => window.start_drag_move(),
```
this would be more consistent with `drag_move` vs `drag_resize`

MiniaczQ (approved):
Few naming nits, I like the example 👍 

One side question, would it be possible to call redraw while we're resizing, so window gets updated immediately, rather than just stretching until we're done?

examples/window/window_drag_move.rs
```diff
@@ -0,0 +1,140 @@
+//! This example illustrates drag move and drag resize without window
+//! decorations.
+//!
+//! When window decorations are not present, the user cannot drag the window.
+//! The `start_drag_move()` function will permit the application to make the
+//! window draggable. It does require that the left mouse button was pressed
```
BenjaminBrienen, line 6: *is being* pressed sounds better to me. Currently the tenses are conflicting.

examples/window/window_drag_move.rs
```diff
@@ -0,0 +1,140 @@
...
+//! when it is called.
```
BenjaminBrienen, line 7: When (inconsistent capitalization with other comments)

BenjaminBrienen (approved):
Only nits

alice-i-cecile:
@shanecelis this needs minor changes following #15887.

shanecelis:
Will do.

examples/window/window_drag_move.rs
shanecelis: I apologize. I don't quite follow this suggestion. I have rewritten that paragraph. Please see if your comment still applies.

shanecelis:
Strange. CI failed, but the same command works on my M2 macOS. It complains about the UiTextWriter not being found, which is strange because many examples use that symbol.

MiniaczQ:
> Strange. CI failed, but the same command works on my M2 macOS. It complains about the UiTextWriter not being found, which is strange because many examples use that symbol.

It's `TextUiWriter` after #15887

PR #15817, opened 2024-10-10
Add macros to reduce boilerplates

VitalyAnkh:
# Objective

- Reduce some boilerplate code.

## Solution

- Added two macros, `impl_handle_wrapper` and `impl_generic_handle_wrapper`, to reduce boilerplate code.
## Testing

- CI.

### DIFF
--- a/crates/bevy_utils/src/lib.rs
+++ b/crates/bevy_utils/src/lib.rs
@@ -407,6 +407,54 @@ pub fn error<E: Debug>(result: Result<(), E>) {
     }
 }
 
+/// Boilerplate for handle wrapper
+#[macro_export]
+macro_rules! impl_handle_wrapper {
+    ($handle_wrapper_type:ident, $asset_type:ty) => {
+        impl From<$handle_wrapper_type> for AssetId<$asset_type> {
+            fn from(handle: $handle_wrapper_type) -> Self {
+                handle.id()
+            }
+        }
+
+        impl From<&$handle_wrapper_type> for AssetId<$asset_type> {
+            fn from(handle: &$handle_wrapper_type) -> Self {
+                handle.id()
+            }
+        }
+    };
+}
+
+/// Boilerplate for generic handle wrapper
+#[macro_export]
+macro_rules! impl_generic_handle_wrapper {
+    ($handle_type:ident, $asset_type:ty, $asset_type_parameter:ident, $parameter_trait:ident) => {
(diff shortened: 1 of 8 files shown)


crates/bevy_utils/src/lib.rs
```diff
@@ -407,6 +407,54 @@ pub fn error<E: Debug>(result: Result<(), E>) {
     }
 }
 
+/// Boilerplate for handle wrapper
+#[macro_export]
+macro_rules! impl_handle_wrapper {
```
MrGVSV, line 412: If anything these should probably exist inside `bevy_asset`

crates/bevy_utils/src/once.rs
```diff
@@ -2,7 +2,7 @@
 #[macro_export]
 macro_rules! once {
     ($expression:expr) => {{
-        use ::core::sync::atomic::{AtomicBool, Ordering};
+        use core::sync::atomic::{AtomicBool, Ordering};
```
MrGVSV, line 5: Accidental change? 
```suggestion
        use ::core::sync::atomic::{AtomicBool, Ordering};
```

crates/bevy_utils/src/lib.rs
```diff
@@ -407,6 +407,54 @@ pub fn error<E: Debug>(result: Result<(), E>) {
...
+
+/// Boilerplate for generic handle wrapper
+#[macro_export]
+macro_rules! impl_generic_handle_wrapper {
+    ($handle_type:ident, $asset_type:ty, $asset_type_parameter:ident, $parameter_trait:ident) => {
+        impl<$asset_type_parameter: $parameter_trait> Default
+            for $handle_type<$asset_type_parameter>
+        {
+            fn default() -> Self {
+                Self(Handle::default())
```
MrGVSV, line 436: As a rule of thumb, macros should use fully qualified paths to avoid issues where certain items are not in scope.

MrGVSV (changes requested):
Couple comments. Personally I don't think we need a macro for this. The boilerplate cost isn't too bad and is actually more readable since you can see exactly what traits are being implemented.

I also wonder if we could use the newly added `derive_more` crate to help here instead with the `From` impls and whatnot

alice-i-cecile:
I don't think this is a meaningful enough improvement to justify the added maintenance / abstraction. Closing, sorry :)

PR #15818, opened 2024-10-10
fix: removed WinitPlugin from headless_renderer example

cxreiff:
# Objective

The `headless_renderer` example is meant to showcase running bevy as a headless renderer, but if run without a display server (for example, over an SSH connection), a panic occurs in `bevy_winit` despite never creating a window:

```rust
bevy_winit-0.14.1/src/lib.rs:132:14:
winit-0.30.5/src/platform_impl/linux/mod.rs:
neither WAYLAND_DISPLAY nor WAYLAND_SOCKET nor DISPLAY is set.
```

This example should run successfully in situations without an available display server, as although the GPU is used for rendering, no window is ever created.

## Solution

Disabling WinitPlugin, where the above panic occurs, allows the example to run in a fully headless environment.

## Testing

- I tested this change in normal circumstances with a display server (on macOS Sequoia and Asahi Linux) and behavior was normal.
- I tested with no display server by connecting via SSH, and running the example (on Asahi Linux). Previously this panics, but with this change it runs normally.

## Considerations

- One could argue that ultimately the user should not need to remove `WinitPlugin`, and instead bevy should only throw the above panic when the application first attempts to create a window.

### DIFF
--- a/examples/app/headless_renderer.rs
+++ b/examples/app/headless_renderer.rs
@@ -24,6 +24,7 @@ use bevy::{
         texture::{BevyDefault, TextureFormatPixelInfo},
         Extract, Render, RenderApp, RenderSet,
     },
+    winit::WinitPlugin,
 };
 use crossbeam_channel::{Receiver, Sender};
 use std::{
@@ -86,7 +87,9 @@ fn main() {
                     primary_window: None,
                     exit_condition: bevy::window::ExitCondition::DontExit,
                     close_when_requested: false,
-                }),
+                })
+                // WinitPlugin will panic in environments without a display server.
+                .disable::<WinitPlugin>(),
         )
         .add_plugins(ImageCopyPlugin)
         // headless frame capture
--- a/examples/app/headless_renderer.rs
+++ b/examples/app/headless_renderer.rs
@@ -82,11 +82,11 @@ fn main() {
         .add_plugins(
             DefaultPlugins
                 .set(ImagePlugin::default_nearest())
-                // Do not create a window on startup.
+                // Not strictly necessary, as the inclusion of ScheduleRunnerPlugin below
+                // replaces the bevy_winit app runner and so a window is never created.
                 .set(WindowPlugin {
                     primary_window: None,
-                    exit_condition: bevy::window::ExitCondition::DontExit,
-                    close_when_requested: false,
+                    ..default()
                 })
                 // WinitPlugin will panic in environments without a display server.
                 .disable::<WinitPlugin>(),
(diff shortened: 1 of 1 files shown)


cxreiff:
Related to this change: from my own testing, the `WindowPlugin` settings above my change (e.g. setting primary_window to None) are not required, as they are redundant with adding `ScheduleRunnerPlugin`, which replaces the bevy_winit app runner and prevents a window from being created. Are there non-obvious performance reasons that the primary_window should be manually set to None? Or can these lines be removed?

```
DefaultPlugins.set(WindowPlugin {
   primary_window: None,
   exit_condition: bevy::window::ExitCondition::DontExit,
   close_when_requested: false,
}),
```

HugoPeters1024 (approved):
Nice fix, works on my machine!

> Are there non-obvious performance reasons that the primary_window should be manually set to None? Or can these lines be removed?

There are no performance considerations, but I think that it couldn't hurt to make it explicit how we don't want a window to be created. That currently it already isn't created because `WinitPlugin` is removed is not a property that we should rely for the future nor as didactic material.

Though if that really doesn't sit right with you, I'd suggest we still leave the `primary_window: None` but add a comment about how it is technically not necessary. WYT?

cxreiff:
Its not that it doesn't do anything with `WinitPlugin` removed, it is that it _has never_ done anything in this example, because of the `ScheduleRunnerPlugin` that has always been included. I agree that there should be a more explicit way of stating no window should be created, but I don't think I agree with the idea of configuring plugins a certain way for strictly documentation purposes. Happy to leave it in to get this merged, however, as I don't feel particularly strongly about it.

Note about my own experience though, I came to this example to learn how headless rendering works in bevy, and seeing the `WindowPlugin` lines instantly gave me several false assumptions about how bevy worked that I needed to dig into source code to dispel: that `primary_window: None` was the reason no window was created (it's actually the alternate app runner), that configs like `exit_condition`/`close_when_requested` were external to bevy_winit and so would still apply without a window (they don't), etcetera. So I personally would push for removing them completely, maybe with a new comment above the line that is actually preventing the window from being created (ScheduleRunnerPlugin). I think I would be a little confused coming to the example and seeing the `primary_window: None` line with a note saying it doesn't actually need to be there.

cxreiff:
Added a commit with comments on both.

BenjaminBrienen (approved):

alice-i-cecile:
Definitely think it's weird that we don't just disable `WindowPlugin` too (or build this up from the new `HeadlessPlugins`). That said, this is a strict improvement from the existing code, so I'm merging it.

PR #15825, opened 2024-10-10
add AsVal trait that helps with constructing Val

TotalKrill:
# Objective

To reduce friction when writing Ui code by letting the code follow the left-to-right mindset. 

Mental model: "The width should be 20 pixels" , as in "affect this, with value,of unit"

instead of:
`ui_rect.width = Val::Px(20.0);`
let us go: 
`ui_rect.width = 20.px();`

## Solution

Implement a trait in geometry and implement for the common types

## Testing

Tested this by modifying the text_debug example to use this instead and it worked.

---

### DIFF
--- a/crates/bevy_ui/src/geometry.rs
+++ b/crates/bevy_ui/src/geometry.rs
@@ -195,17 +195,17 @@ impl Neg for Val {
 /// assert_eq!(ui_rect.bottom, Val::Percent(40.0));
 /// ```
 pub trait AsVal {
-    /// turn the value into `Val::Px(self)`
+    /// Turns the value into `Val::Px(self)`.
     fn px(self) -> Val;
-    /// turn the value into `Val::Percent(self)`
+    /// Turns the value into `Val::Percent(self)`.
     fn percent(self) -> Val;
-    /// turn the value into `Val::Vw(self)`
+    /// Turns the value into `Val::Vw(self)`.
     fn vw(self) -> Val;
-    /// turn the value into `Val::Vh(self)`
+    /// Turns the value into `Val::Vh(self)`.
     fn vh(self) -> Val;
-    /// turn the value into `Val::VMin(self)`
+    /// Turns the value into `Val::VMin(self)`.
     fn vmin(self) -> Val;
-    /// turn the value into `Val::VMax(self)`
+    /// Turns the value into `Val::VMax(self)`.
     fn vmax(self) -> Val;
 }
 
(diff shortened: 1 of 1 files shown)


alice-i-cecile (approved):
I like the syntax, and the implementation is clean. The use a macro is justified here IMO: there's a ton of boilerplate and the code is super simple.

alice-i-cecile:
Bumping to the 0.16 milestone since this is implementable externally and it's a new feature :)

crates/bevy_ui/src/geometry.rs
```diff
@@ -175,6 +175,73 @@ impl Neg for Val {
...
+///     40.0.percent(),
+/// );
+///
+/// assert_eq!(ui_rect.left, Val::Px(10.0));
+/// assert_eq!(ui_rect.right, Val::Px(20.0));
+/// assert_eq!(ui_rect.top, Val::Percent(30.0));
+/// assert_eq!(ui_rect.bottom, Val::Percent(40.0));
+/// ```
+pub trait AsVal {
+    /// turn the value into `Val::Px(self)`
```
alice-i-cecile, line 198: Nit: `turn` should be capitalized as `Turn` in these doc comments :)

Jondolf (approved):

alice-i-cecile:
If we're going to be doing this, we should export the trait in the prelude TBH.

TotalKrill:
It's there in the prelude already. From bevy_ui::prelude

atornity (approved):

cart:
As a heads up, within the context of BSN I was planning on doing: `px(10.0)`. That will notably play nicer with function reflection in the context of scene assets, as this type of trait-function-reflection requires manual special casing to resolve. Idk if we should have two very similar ways to do this / better to have one way that works in every situation.

alice-i-cecile:
I'm fine with either solution. This plays a bit nicer with autocomplete. We should merge one or the other for 0.16 though :)

cart:
> This plays a bit nicer with autocomplete

The downside here (if this is in the prelude) is that it adds `.px()` to autocomplete for every float in every context.

ickshonpe:
It wasn't popular the last time I suggested it, but I think the most conveniant thing for users would just be to make all the functions that accept `Val` take `Into<Val>` instead and impl `From` for `f32` and `i32` to `Val::Px`.

cart:
I'm closing this in favor of `px(10.)`-style helper functions (or alternatively, exporting `Val::Px` as `Px()` in the prelude)

PR #15830, opened 2024-10-10
Deprecate SpatialBundle

NiseVoid:
# Objective

- Required components replace bundles, but `SpatialBundle` is yet to be deprecated

## Solution

- Deprecate `SpatialBundle`
- Insert `Transform` and `Visibility` instead in examples using it
- In `spawn` or `insert` inserting a default `Transform` or `Visibility` with component already requiring either, remove those components from the tuple

## Testing

Yes, I ran the examples I changed and tests
The `gamepad_viewer` and and `custom_shader_instancing` examples don't work as intended due to entirely unrelated code, didn't check main.
Run examples, or just check that all spawned values are identical
Linux, wayland trough x11 (cause that's the default feature)

---

## Migration Guide

`SpatialBundle` is now deprecated, insert `Transform` and `Visibility` instead which will automatically insert all other components that were in the bundle. If you do not specify these values and any other components in your `spawn`/`insert` call already requires either of these components you can leave that one out.

before:
```rust
commands.spawn(SpatialBundle::default());
```

after:
```rust
commands.spawn((Transform::default(), Visibility::default());
```

### DIFF
--- a/examples/2d/bounding_2d.rs
+++ b/examples/2d/bounding_2d.rs
@@ -203,20 +203,14 @@ const OFFSET_Y: f32 = 75.;
 fn setup(mut commands: Commands) {
     commands.spawn(Camera2d);
     commands.spawn((
-        SpatialBundle {
-            transform: Transform::from_xyz(-OFFSET_X, OFFSET_Y, 0.),
-            ..default()
-        },
+        Transform::from_xyz(-OFFSET_X, OFFSET_Y, 0.),
         Shape::Circle(Circle::new(45.)),
         DesiredVolume::Aabb,
         Intersects::default(),
     ));
 
     commands.spawn((
-        SpatialBundle {
-            transform: Transform::from_xyz(0., OFFSET_Y, 0.),
-            ..default()
-        },
+        Transform::from_xyz(0., OFFSET_Y, 0.),
         Shape::Rectangle(Rectangle::new(80., 80.)),
         Spin,
         DesiredVolume::Circle,
@@ -224,10 +218,7 @@ fn setup(mut commands: Commands) {
     ));
 
     commands.spawn((
-        SpatialBundle {
-            transform: Transform::from_xyz(OFFSET_X, OFFSET_Y, 0.),
-            ..default()
-        },
+        Transform::from_xyz(OFFSET_X, OFFSET_Y, 0.),
         Shape::Triangle(Triangle2d::new(
             Vec2::new(-40., -40.),
             Vec2::new(-20., 40.),
@@ -239,10 +230,7 @@ fn setup(mut commands: Commands) {
     ));
 
     commands.spawn((
-        SpatialBundle {
-            transform: Transform::from_xyz(-OFFSET_X, -OFFSET_Y, 0.),
-            ..default()
-        },
+        Transform::from_xyz(-OFFSET_X, -OFFSET_Y, 0.),
         Shape::Line(Segment2d::new(Dir2::from_xy(1., 0.3).unwrap(), 90.)),
         Spin,
         DesiredVolume::Circle,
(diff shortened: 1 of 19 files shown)


alice-i-cecile:
> The gamepad_viewer and and custom_shader_instancing examples don't work as intended due to entirely unrelated code, didn't check main.

Can you check the issues for these? I think the former is not yet reported.

examples/shader/custom_phase_item.rs
```diff
@@ -198,18 +198,16 @@ fn main() {
...
     // Spawn a single entity that has custom rendering. It'll be extracted into
     // the render world via [`ExtractComponent`].
-    commands
-        .spawn(SpatialBundle {
-            visibility: Visibility::Visible,
-            transform: Transform::IDENTITY,
-            ..default()
-        })
+    commands.spawn((
+        Visibility::default(),
```
mnmaita, line 202: Shouldn't this be `Visibility::Visible`?

examples/2d/bounding_2d.rs
mnmaita, line 1: Shouldn't `Visibility::default()` be added on these removals too?

examples/3d/fog_volumes.rs
```diff
@@ -30,20 +30,17 @@ fn main() {
...
-    commands
-        .spawn(SpatialBundle {
-            visibility: Visibility::Visible,
-            transform: Transform::from_xyz(0.0, 0.5, 0.0),
-            ..default()
-        })
-        .insert(FogVolume {
+    commands.spawn((
+        Transform::from_xyz(0.0, 0.5, 0.0),
+        FogVolume {
```
mnmaita, line 35: This should also have `Visibility::Visible` right? I assume `FogVolume` would be requiring `Visibility::default()` if not. It might not matter though...

examples/3d/scrolling_fog.rs
```diff
@@ -113,11 +113,7 @@ fn setup(
 
     // Spawn a FogVolume and use the repeating noise texture as its density texture.
     commands.spawn((
-        SpatialBundle {
-            visibility: Visibility::Visible,
-            transform: Transform::from_xyz(0.0, 32.0, 0.0).with_scale(Vec3::splat(64.0)),
-            ..default()
-        },
+        Transform::from_xyz(0.0, 32.0, 0.0).with_scale(Vec3::splat(64.0)),
```
mnmaita, line 116: Same comment as in `fog_volumes.rs`

examples/shader/custom_phase_item.rs
NiseVoid: `Visibility::default()` becomes `Visibility::Inherited` which behaves as `Visibility::Visible` if there is no parent hidden afaik, which in this case there isn't.

examples/2d/bounding_2d.rs
NiseVoid: The bounding example actually doesn't use visibility at all (nor `GlobalTransform`), I created this example but I'm not sure why I used `SpatialBundle` instead of just `Transform`.

examples/3d/fog_volumes.rs
NiseVoid: Same as above, in this case `FogVolume` requires `Visibility` which means it defaults to `Inherited`, and functions as visible.

mnmaita: Thanks for addressing my comments!

tom-frantz (approved):
Looks good to me! @mnmaita 's comments look like they were all covered, and the other examples had the necessary required components. I didn't see any other uses of `SpatialBundle` throughout the code base either, so I'm happy to approve given alice-i-cecile 's comments are covered too!

alice-i-cecile (approved):

PR #15839, opened 2024-10-10
bevy_ecs: Special-case `Entity::PLACEHOLDER` formatting

MrGVSV:
# Objective

Oftentimes, users will store an entity on a component or resource. To make this component/resource `Default`-able, they might initialize it with `Entity::PLACEHOLDER`. This is sometimes done to avoid the need for an `Option<Entity>`, especially if it complicates other logic. 

For example, it's used in this `Selection` resource to denote "no selection":

```rust
#[derive(Resource, Debug)]
struct Selection(Entity);

impl Default for Selection {
    fn default() -> Self {
        Self(Entity::PLACEHOLDER)
    }
}
```

The problem is that if we try to `Debug` the current `Selection`, we get back: `4294967295v1#8589934591`. It's not immediately obvious whether or not the entity is an actual entity or the placeholder.

Now while it doesn't take long to realize that this is in fact just the value of `Entity::PLACEHOLDER`, it would be a lot clearer if this was made explicit, especially for these particular use cases.

## Solution

This PR makes the `Debug` and `Display` impls for `Entity` return `PLACEHOLDER` for the `Entity::PLACEHOLDER` constant.

~~Feel free to bikeshed the actual value returned here. I think `PLACEHOLDER` on its own could work too.~~ Swapped to `PLACEHOLDER` from `Entity::PLACEHOLDER`.

## Testing

You can test locally by running:

```
cargo test --package bevy_ecs
```

---

## Migration Guide

The `Debug` and `Display` impls for `Entity` now return `PLACEHOLDER` for the `Entity::PLACEHOLDER` constant. If you had any code relying on these values, you may need to account for this change.

### DIFF
--- a/crates/bevy_ecs/src/entity/mod.rs
+++ b/crates/bevy_ecs/src/entity/mod.rs
@@ -397,7 +397,7 @@ impl<'de> Deserialize<'de> for Entity {
 ///
 /// This takes the format: `{index}v{generation}#{bits}`.
 ///
-/// For [`Entity::PLACEHOLDER`], this outputs `Entity::PLACEHOLDER`.
+/// For [`Entity::PLACEHOLDER`], this outputs `PLACEHOLDER`.
 ///
 /// # Usage
 ///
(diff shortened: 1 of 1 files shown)


alice-i-cecile (commented):
I prefer just PLACEHOLDER.

MrGVSV:
Two votes for `PLACEHOLDER` so I swapped to that for now. If anyone has any other opinions, feel free to share!

ItsDoot (approved):

LiamGallagher737 (approved):

lewiszlw (approved):

PR #15843, opened 2024-10-10
Add `compile-check-no-std` Command to CI Tool

### LINKED ISSUE #15840: Add `no_std` Checks to CI

Bevy contains a number of `no_std` crates (with more hopefully on the way!), but there are currently no tests in the CI to ensure they actually _are_ `no_std` compatible. If any `std` dependency is added (even transiently!), it will break this guarantee.

- Add one or more `no_std` targets to a specific `no_std` `cargo build` CI action. The most obvious platform would be `x86_64_unknown-none`, since Bevy is known to work on `x86_64` across many operating systems. However, more could be added based on community usage (e.g., Raspberry Pi, STM32, etc.). This would ensure it is possible to build Bevy and/or its crates without the `std` library being linked.
- Add a test CI action which disables the `std` feature (and possibly `alloc` as well) to ensure unit tests pass in a `no_std` environment. Note that these tests can be run on a platform _with_ the `std` library, since the above test ensures that the `std` library isn't _used_ directly or transiently.

We could do nothing for now. Since the Bevy `no_std` community is just starting, there are currently no users to complain if `no_std` support is broken.

bushrat011899:
# Objective

- Fixes #15840

## Solution

Added a new subcommand to the CI tool, `compile-check-no-std`, which will attempt to compile each `no_std` crate in Bevy with the appropriate features (no-defaults, `libm`, etc.) for `x86_64-unknown-none`. The exact target chosen could be changed to any reasonable platform which does not include the `std` library.

The currently tested crates are:

- `bevy_ptr`
- `bevy_utils`
- `bevy_mikktspace`

As more crates have `no_std` support added, they _should_ be added to this CI command. Once Bevy itself can be `no_std`, the individual checks can be replaced with just checking Bevy, since it will transiently check all other crates as appropriate.

## Testing

- Ran CI. From a clean target directory (`cargo clean`), these new checks take approximately 10 seconds total.


tools/ci/src/commands/compile_check_no_std.rs
```diff
@@ -0,0 +1,35 @@
...
+use xshell::cmd;
+
+/// Checks that the project compiles for a `no_std` target.
+#[derive(FromArgs, Default)]
+#[argh(subcommand, name = "compile-check-no-std")]
+pub struct CompileCheckNoStdCommand {}
+
+impl Prepare for CompileCheckNoStdCommand {
+    fn prepare<'a>(&self, sh: &'a xshell::Shell, _flags: Flag) -> Vec<PreparedCommand<'a>> {
+        vec![PreparedCommand::new::<Self>(
```
BenjaminBrienen, line 12: Can we reduce the amount of copy-pasting that will happen as more crates are added here by building the vec of commands from a vec of crate name strings?

BenjaminBrienen (changes requested):

tools/ci/src/commands/compile_check_no_std.rs
bushrat011899: We _could_, but some crates will require features as well (e.g., `bevy_mikktspace` requires the `libm` feature). Once we have `no_std` exposed through `bevy` itself rather than its crates we'll be able to remove all these individual checks anyway, since it would be sufficient to just check that `bevy` itself works on `no_std`

BenjaminBrienen (approved):

.github/workflows/ci.yml
```diff
@@ -127,6 +127,32 @@ jobs:
...
+            ~/.cargo/registry/index/
+            ~/.cargo/registry/cache/
+            ~/.cargo/git/db/
+            target/
+            crates/bevy_ecs_compile_fail_tests/target/
+            crates/bevy_reflect_compile_fail_tests/target/
+          key: ${{ runner.os }}-cargo-check-compiles-no-std-${{ hashFiles('**/Cargo.toml') }}
+      - uses: dtolnay/rust-toolchain@stable
+        with:
+          toolchain: stable
```
mockersf, line 150:
```suggestion
```

tools/ci/src/commands/compile_check_no_std.rs
```diff
@@ -0,0 +1,35 @@
+use crate::{Flag, Prepare, PreparedCommand};
+use argh::FromArgs;
+use xshell::cmd;
+
+/// Checks that the project compiles for a `no_std` target.
+#[derive(FromArgs, Default)]
+#[argh(subcommand, name = "compile-check-no-std")]
+pub struct CompileCheckNoStdCommand {}
```
mockersf, line 8: could you make the target as a parameter to the subcommand, defaulting to `x86_64-unknown-none`?

bushrat011899: Can do! First time working with `argh` but I think I've done this correctly.

mockersf (approved):

PR #15561, opened 2024-10-01
Overflow clip margin

ickshonpe:
# Objective

Limited implementation of the CSS property `overflow-clip-margin`
https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-clip-margin

Allows you to control the visible area for clipped content when using overfllow-clip, -hidden, or -scroll and expand it with a margin.

Based on #15442

Fixes #15468

## Solution

Adds a new field to Style: `overflow_clip_margin: OverflowClipMargin`.
The field is ignored unless overflow-clip, -hidden or -scroll is set on at least one axis.

`OverflowClipMargin` has these associated constructor functions:
```
pub const fn content_box() -> Self;
pub const fn padding_box() -> Self;
pub const fn border_box() -> Self;
```
You can also use the method `with_margin` to increases the size of the visible area:
```
commands
  .spawn(NodeBundle {
      style: Style {
          width: Val::Px(100.),
          height: Val::Px(100.),
          padding: UiRect::all(Val::Px(20.)),
          border: UiRect::all(Val::Px(5.)),
          overflow: Overflow::clip(),
          overflow_clip_margin: OverflowClipMargin::border_box().with_margin(25.),
          ..Default::default()
      },
      border_color: Color::BLACK.into(),
      background_color: GRAY.into(),
      ..Default::default()
  })
```
`with_margin` expects a length in logical pixels, negative values are clamped to zero.

## Notes
* To keep this PR as simple as possible I omitted responsive margin values support. This could be added in a follow up if we want it.
* CSS also supports a `margin-box` option but we don't have access to the margin values in `Node` so it's probably not feasible to implement atm.

## Testing

```cargo run --example overflow_clip_margin```

overflow-clip-margin

## Migration Guide

Style has a new field `OverflowClipMargin`.  It allows users to set the visible area for clipped content when using overflow-clip, -hidden, or -scroll and expand it with a margin.

There are three associated constructor functions `content_box`, `padding_box` and `border_box`:
* `content_box`: elements painted outside of the content box area (the innermost part of the node excluding the padding and border) of the node are clipped. This is the new default behaviour.
* `padding_box`: elements painted outside outside of the padding area of the node are clipped. 
* `border_box`:  elements painted outside of the bounds of the node are clipped. This matches the behaviour from Bevy 0.14.

There is also a `with_margin` method that increases the size of the visible area by the given number in logical pixels, negative margin values are clamped to zero.

[...]

PR #15844, opened 2024-10-10
Upgrade to Taffy 0.6

nicoburns:
# Objective

- Keep Taffy version up to date

Taffy 0.6 doesn't include a huge amount relevant to Bevy. But it does:

- Add the `box_sizing` style
- Expose the computed `margin` in layout
- Traitifies the `Style` struct, which opens up the possibility of using Bevy's `Style` struct directly (although Bevy currently does some style resolution at conversion time which would no longer be cached if it was used directly).
- Have a few bug fixes in the layout algorithms

## Solution

- Upgrade Taffy to `0.6.0`

## Testing

- I've run the `grid` example. All looks good.
- More testing is probably warranted. We have had regressions from Taffy upgrades before
- Having said that, most of the algorithm changes this cycle were driven by fixing WPT tests run through the new Servo integration. So they're possibly less likely than usual to cause regressions.

## Breaking changes

The only "breaking" change is adding a field to `Style`. Probably doesn't bear mentioning?

### DIFF
--- a/crates/bevy_ui/src/ui_node.rs
+++ b/crates/bevy_ui/src/ui_node.rs
@@ -323,6 +323,15 @@ pub struct Node {
     /// <https://developer.mozilla.org/en-US/docs/Web/CSS/display>
     pub display: Display,
 
+    /// Which part of a Node's box length styles like width and height control
+    ///   - [`BoxSizing::BorderBox`]: They refer to the "border box" size (size including padding and border)
+    ///   - [`BoxSizing::ContentBox`]: They refer to the "content box" size (size excluding padding and border)
+    ///
+    /// `BoxSizing::BorderBox` is generally considered more intuitive and is the default in Bevy even though it is not on the web.
+    ///
+    /// See: <https://developer.mozilla.org/en-US/docs/Web/CSS/box-sizing>
+    pub box_sizing: BoxSizing,
+
     /// Whether a node should be laid out in-flow with, or independently of its siblings:
     ///  - [`PositionType::Relative`]: Layout this node in-flow with other nodes using the usual (flexbox/grid) layout algorithm.
     ///  - [`PositionType::Absolute`]: Layout this node on top and independently of other nodes.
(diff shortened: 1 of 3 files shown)


crates/bevy_ui/src/ui_node.rs
```diff
@@ -866,6 +881,33 @@ impl Default for Display {
     }
 }
 
+/// Whether to show or hide overflowing items
```
rparrett, line 884: Stray comments from another `Overflow`

crates/bevy_ui/src/ui_node.rs
```diff
@@ -271,6 +271,20 @@ pub struct Style {
     /// <https://developer.mozilla.org/en-US/docs/Web/CSS/display>
     pub display: Display,
 
+    /// Which layout algorithm to use when laying out this node's contents:
+    ///   - [`Display::Flex`]: Use the Flexbox layout algorithm
+    ///   - [`Display::Grid`]: Use the CSS Grid layout algorithm
+    ///   - [`Display::None`]: Hide this node and perform layout as if it does not exist.
+    ///
```
rparrett, line 278: Stray comments from `Display`

crates/bevy_ui/src/ui_node.rs
```diff
@@ -866,6 +881,33 @@ impl Default for Display {
...
+/// Whether to show or hide overflowing items
+#[derive(Copy, Clone, PartialEq, Eq, Debug, Reflect)]
+#[reflect(Default, PartialEq)]
+#[cfg_attr(
+    feature = "serialize",
+    derive(serde::Serialize, serde::Deserialize),
+    reflect(Serialize, Deserialize)
+)]
+
+/// Which part of a Node's box length styles like width and height control
```
rparrett, line 893: nit: I think we're generally placing doc comments above attributes, and not including new lines between

rparrett (changes requested):

ickshonpe (approved):
Looks good besides the comments nits.

Adding the computed margin values is something I've wanted for a while and means I can add support for `margin-box` to #15561.

nicoburns:
Nits addressed

rparrett (approved):

mnmaita (approved):

BenjaminBrienen (approved):

robtfm:
fixes #16304

alice-i-cecile:
Yep! I'll grab this in my next merge train.

PR #15846, opened 2024-10-11
Meshlet new error projection

JMS55:
* New error projection code taken from @zeux's meshoptimizer nanite.cpp demo for determining LOD (thanks zeux!)
* Builder: `compute_lod_group_data()`
* Runtime: `lod_error_is_imperceptible()`

### DIFF
--- a/crates/bevy_pbr/src/meshlet/cull_clusters.wgsl
+++ b/crates/bevy_pbr/src/meshlet/cull_clusters.wgsl
@@ -140,17 +140,22 @@ fn cull_clusters(
     meshlet_raster_clusters[buffer_slot] = cluster_id;
 }
 
-fn lod_error_is_imperceptible(sphere: MeshletBoundingSphere, error: f32, world_from_local: mat4x4<f32>, world_scale: f32) -> bool {
-    let cp_world = world_from_local * vec4(sphere.center, 1.0);
-    let r_view = world_scale * sphere.radius;
-    let cp_view = (view.view_from_world * vec4(cp_world.xyz, 1.0)).xyz;
+// TODO: This doesn't account for perspective and other edge cases as well as Nanite does
+// https://github.com/zeux/meshoptimizer/blob/1e48e96c7e8059321de492865165e9ef071bffba/demo/nanite.cpp#L115
+fn lod_error_is_imperceptible(lod_sphere: MeshletBoundingSphere, simplification_error: f32, world_from_local: mat4x4<f32>, world_scale: f32) -> bool {
+    let sphere_world_space = (world_from_local * vec4(lod_sphere.center, 1.0)).xyz;
+    let radius_world_space = world_scale * lod_sphere.radius;
+    let error_world_space = world_scale * simplification_error;
 
-    // TODO: Handle view clipping / being inside sphere bounds
-    let aabb = project_view_space_sphere_to_screen_space_aabb(cp_view, r_view);
-    let screen_size = max(aabb.z - aabb.x, aabb.w - aabb.y);
-    let meters_per_pixel = sphere.radius / screen_size;
+    let distance_to_closest_point_on_sphere = distance(sphere_world_space, view.world_position) - radius_world_space;
+    let distance_to_closest_point_on_sphere_clamped_to_znear = max(distance_to_closest_point_on_sphere, view.clip_from_view[3][2]);
+    var projected_error = error_world_space / distance_to_closest_point_on_sphere_clamped_to_znear;
+    if view.clip_from_view[3][3] != 1.0 {
+        projected_error *= view.clip_from_view[1][1];
+    }
+    projected_error *= view.viewport.w;
 
-    return error < meters_per_pixel;
+    return projected_error < 1.0;
 }
 
 // https://zeux.io/2023/01/12/approximate-projected-bounds
--- a/crates/bevy_pbr/src/meshlet/cull_clusters.wgsl
+++ b/crates/bevy_pbr/src/meshlet/cull_clusters.wgsl
@@ -147,15 +147,19 @@ fn lod_error_is_imperceptible(lod_sphere: MeshletBoundingSphere, simplification_
     let radius_world_space = world_scale * lod_sphere.radius;
     let error_world_space = world_scale * simplification_error;
 
-    let distance_to_closest_point_on_sphere = distance(sphere_world_space, view.world_position) - radius_world_space;
-    let distance_to_closest_point_on_sphere_clamped_to_znear = max(distance_to_closest_point_on_sphere, view.clip_from_view[3][2]);
-    var projected_error = error_world_space / distance_to_closest_point_on_sphere_clamped_to_znear;
+    var projected_error = error_world_space;
     if view.clip_from_view[3][3] != 1.0 {
-        projected_error *= view.clip_from_view[1][1];
+        // Perspective
+        let distance_to_closest_point_on_sphere = distance(sphere_world_space, view.world_position) - radius_world_space;
+        let distance_to_closest_point_on_sphere_clamped_to_znear = max(distance_to_closest_point_on_sphere, view.clip_from_view[3][2]);
+        projected_error *= view.clip_from_view[1][1] / distance_to_closest_point_on_sphere_clamped_to_znear;
+    } else {
+        // Orthographic
+        projected_error *= view.clip_from_view[1][1] * 0.5;
     }
     projected_error *= view.viewport.w;
 
-    return projected_error < 1.0;
+    return projected_error < 0.5;
 }
 
 // https://zeux.io/2023/01/12/approximate-projected-bounds
--- a/crates/bevy_pbr/src/meshlet/cull_clusters.wgsl
+++ b/crates/bevy_pbr/src/meshlet/cull_clusters.wgsl
@@ -1,6 +1,7 @@
 #import bevy_pbr::meshlet_bindings::{
     meshlet_cluster_meshlet_ids,
     meshlet_bounding_spheres,
+    meshlet_simplification_errors,
     meshlet_cluster_instance_ids,
     meshlet_instance_uniforms,
     meshlet_second_pass_candidates,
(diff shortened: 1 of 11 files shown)


crates/bevy_pbr/src/meshlet/cull_clusters.wgsl
```diff
@@ -148,14 +140,22 @@ fn cull_clusters(
...
+// https://github.com/zeux/meshoptimizer/blob/1e48e96c7e8059321de492865165e9ef071bffba/demo/nanite.cpp#L115
+fn lod_error_is_imperceptible(lod_sphere: MeshletBoundingSphere, simplification_error: f32, world_from_local: mat4x4<f32>, world_scale: f32) -> bool {
+    let sphere_world_space = (world_from_local * vec4(lod_sphere.center, 1.0)).xyz;
+    let radius_world_space = world_scale * lod_sphere.radius;
+    let error_world_space = world_scale * simplification_error;
+
+    let distance_to_closest_point_on_sphere = distance(sphere_world_space, view.world_position) - radius_world_space;
+    let distance_to_closest_point_on_sphere_clamped_to_znear = max(distance_to_closest_point_on_sphere, view.clip_from_view[3][2]);
+    var projected_error = error_world_space / distance_to_closest_point_on_sphere_clamped_to_znear;
+    if view.clip_from_view[3][3] != 1.0 {
```
zeux, line 153: Is this an ortho check? If so I think that should be done separately; in an ortho projection you don't need to divide by distance at all (but you do still need to multiply by ortho scaling similar to perspective)

JMS55: Yep, this is an orthographic projection check. Thanks for letting me know, I'll take another stab at trying to understand the theory and figure out what to do for Ortho.

JMS55:
After thinking on it I think I should raise the 1px threshold, to some value determined empirically by doing some comparisons vs forcing lod 0. I think I'm probably overestimating the error and using too-fine grained LODs based on how this PR is responding to camera movement. Will have to test before merging.

crates/bevy_pbr/src/meshlet/cull_clusters.wgsl
JMS55: Is this correct? Not sure I applied the ortho scaling factor correctly

```rust
var projected_error = error_world_space;
if view.clip_from_view[3][3] != 1.0 {
    // Perspective
    let distance_to_closest_point_on_sphere = distance(sphere_world_space, view.world_position) - radius_world_space;
    let distance_to_closest_point_on_sphere_clamped_to_znear = max(distance_to_closest_point_on_sphere, view.clip_from_view[3][2]);
    projected_error *= view.clip_from_view[1][1] / distance_to_closest_point_on_sphere_clamped_to_znear;
} else {
    // Orthographic
    projected_error *= view.clip_from_view[1][1] * 0.5;
}
projected_error *= view.viewport.w;
```

Where the ortho matrix is
```rust
pub fn orthographic_rh(
    left: f32,
    right: f32,
    bottom: f32,
    top: f32,
    near: f32,
    far: f32,
) -> Self {
    let rcp_width = 1.0 / (right - left);
    let rcp_height = 1.0 / (top - bottom);
    let r = 1.0 / (near - far);
    Self::from_cols(
        Vec4::new(rcp_width + rcp_width, 0.0, 0.0, 0.0),
        Vec4::new(0.0, rcp_height + rcp_height, 0.0, 0.0),
        Vec4::new(0.0, 0.0, r, 0.0),
        Vec4::new(
            -(left + right) * rcp_width,
            -(top + bottom) * rcp_height,
            r * near,
            1.0,
        ),
    )
}```

**Comment by JMS55** [comment]
Strangely enough I find it impossible to ever render LOD 0 🤔. Higher LODs always seem to have low enough error.

**Comment by JMS55** [comment]
On the other hand, 0.5 threshold means the cliffs seem to take too long to start using coarser LODs as I move away from it. Doing error projection sucks 😭.

File: crates/bevy_pbr/src/meshlet/cull_clusters.wgsl (hunk 1 above)
Comment (zeux) [reply crates/bevy_pbr/src/meshlet/cull_clusters.wgsl:151]: I think you’d want 0.5 factor for perspective as well? Ortho math seems correct (just eyeballing)

**Comment by JMS55** [comment]
I also found https://hhoppe.com/vdrpm.pdf, section 4 "Screen-space geometric error" seems to have a formula. Reading time.

File: crates/bevy_pbr/src/meshlet/cull_clusters.wgsl (hunk 1 above)
Comment (JMS55) [reply crates/bevy_pbr/src/meshlet/cull_clusters.wgsl:151]: Oh yeah, I missed that when copying your code. Thanks for the extra pair of eyes :)

Comment (JMS55) [reply crates/bevy_pbr/src/meshlet/cull_clusters.wgsl:151]: Reduces down to this:
```rust
var projected_error = error_world_space;
if view.clip_from_view[3][3] != 1.0 {
    // Perspective
    let distance_to_closest_point_on_sphere = distance(sphere_world_space, view.world_position) - radius_world_space;
    let distance_to_closest_point_on_sphere_clamped_to_znear = max(distance_to_closest_point_on_sphere, view.clip_from_view[3][2]);
    projected_error *= view.clip_from_view[1][1] / distance_to_closest_point_on_sphere_clamped_to_znear;
}
projected_error *= view.clip_from_view[1][1] * 0.5;
projected_error *= view.viewport.w;
```

JMS55:
Think I'm happy with the state of this PR now, actually. Going to upload a new bunny and let people test, but I'm calling this done (and an improvement over main).

zeux:
Just a note, and I'm *not* suggesting a change here :) but if you ever hit the perspective distortion issues and want a quick tweak, this switches from radial distance to the actual projected distance:

```patch
diff --git a/crates/bevy_pbr/src/meshlet/cull_clusters.wgsl b/crates/bevy_pbr/src/meshlet/cull_clusters.wgsl
index 8ff42bced..48b10399c 100644
--- a/crates/bevy_pbr/src/meshlet/cull_clusters.wgsl
+++ b/crates/bevy_pbr/src/meshlet/cull_clusters.wgsl
@@ -147,11 +147,12 @@ fn lod_error_is_imperceptible(lod_sphere: MeshletBoundingSphere, simplification_
     let sphere_world_space = (world_from_local * vec4(lod_sphere.center, 1.0)).xyz;
     let radius_world_space = world_scale * lod_sphere.radius;
     let error_world_space = world_scale * simplification_error;
+    let sphere_view_space = (view.view_from_world * vec4(sphere_world_space, 1.0)).xyz;
 
     var projected_error = error_world_space;
     if view.clip_from_view[3][3] != 1.0 {
         // Perspective
-        let distance_to_closest_point_on_sphere = distance(sphere_world_space, view.world_position) - radius_world_space;
+        let distance_to_closest_point_on_sphere = -sphere_view_space.z - radius_world_space;
         let distance_to_closest_point_on_sphere_clamped_to_znear = max(distance_to_closest_point_on_sphere, view.clip_from_view[3][2]);
         projected_error /= distance_to_closest_point_on_sphere_clamped_to_znear;
     }
```

This works about the same for objects near the center of the screen. For objects close to screen edges, current formula (radial distance) returns a larger distance to sphere, which may effectively underestimate the projected error as objects near edges get stretched more. The tweak above should keep the error estimate conservative.

(.z is negated because in Bevy's view space coordinate system, points in front of the camera have negative Z, similarly to `project_view_space_sphere_to_screen_space_aabb`)

I personally don't like this because this introduces a dependence on the camera rotation, so LODs switch when camera rotates around its viewpoint; it's *correct* to do this, however, so in case this ever becomes necessary the change is simple.

JMS55:
@zeux I tried it out, and I actually agree with you. It feels weird that LODs get more detailed on the edges of the screen. If anything, I would want the reverse, so that the things you're looking directly at are more detailed.

Thanks for the code though, good to try both ways and confirm.

atlv24 (approved):

crates/bevy_pbr/src/meshlet/from_mesh.rs
```diff
@@ -330,10 +327,55 @@ fn simplify_meshlet_group(
...
+        let meshlet_lod_bounding_sphere = bounding_spheres[*meshlet_id].lod_group_sphere;
+        group_bounding_sphere.center +=
+            meshlet_lod_bounding_sphere.center * meshlet_lod_bounding_sphere.radius;
+        weight += meshlet_lod_bounding_sphere.radius;
+    }
+    group_bounding_sphere.center /= weight;
+
+    // Force parent group sphere to contain all child group spheres (we're currently building the parent from its children)
+    // TODO: This does not produce the absolute minimal bounding sphere. Doing so is non-trivial.
+    //       "Smallest enclosing balls of balls" http://www.inf.ethz.ch/personal/emo/DoctThesisFiles/fischer05.pdf
```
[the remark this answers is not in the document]
zeux: The issue is that if the parent sphere does not contain all spheres of immediate children, the error function stops being monotonic - there will exist a camera position from which the distance to parent is larger vs distance to one of the children, so there's a risk that the parent error will be computed as imperceptible but the child error won't be. Monotonicity assumption is important to be able to select the clusters to render independently from each other (via the "parent's error is too large but mine is okay"); in this scenario you could have multiple overlapping clusters get selected I believe. So the containment of immediate children bounds is a hard requirement for correctness, and then it's just a question of how tight the parent sphere is over the bounds.

JMS55:
I think this PR is bugged actually. Getting some really weird errors. 

image

alice-i-cecile:
@JMS55 can you explain why this is important to get into the 0.15 release for me please?

JMS55:
It's a significant improvement/bugfix for meshlets. We don't _have_ to merge any of the ones in the milestone, but I'd like to so that the blog post I'm going to write covering meshlet changes from bevy 0.14->0.15 can match with what's in mainline.

If you want don't want to block the release or RC on it that's fine, but I would like to get them merged sooner than later so I can move on to the next set of improvements.

alice-i-cecile:
Great, thanks for the context. I'll let Cart or one of the rendering SMEs make the final call: I don't feel strongly either way.

cart (approved):
Reasonably straightforward change (and fully scoped to the experimental meshlets module). I'm cool with including this in 0.15. In the future I'd appreciate a slightly more fleshed out PR description in cases like this. Had to actually read the code to understand the scope of the change and motivation.

PR #15850, opened 2024-10-11
Fix `minimising` example minimizing every frame

rparrett:
# Objective

The `minimising` example is a bit annoying to run locally, because it attempts to minimize the window every frame, so un-minimizing it is difficult.

## Solution

Only minimize once.

The contents of the example can now be inspected, and the window easily closed after the minimization happens.

## Testing

`cargo run --example minimising`

I tested on macos only.

### DIFF
--- a/tests/window/minimising.rs
+++ b/tests/window/minimising.rs
@@ -30,7 +30,7 @@ fn minimise_automatically(
     let mut window = windows.single_mut();
     window.set_minimized(true);
 
-    *done = true
+    *done = true;
 }
 
 /// A simple 3d scene, taken from the `3d_scene` example
--- a/tests/window/minimising.rs
+++ b/tests/window/minimising.rs
@@ -21,16 +21,14 @@ fn main() {
 fn minimise_automatically(
     mut windows: Query<&mut Window>,
     frames: Res<FrameCount>,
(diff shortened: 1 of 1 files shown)


tests/window/minimising.rs
```diff
@@ -18,13 +18,19 @@ fn main() {
...
-        let mut window = windows.single_mut();
-        window.set_minimized(true);
-    } else {
-        *frames += 1;
+fn minimise_automatically(
+    mut windows: Query<&mut Window>,
+    frames: Res<FrameCount>,
+    mut done: Local<bool>,
+) {
+    if *done || frames.0 < 60 {
```
mockersf, line 26:
```suggestion
) {
    if frames.0 != 60 {
```

tests/window/minimising.rs
```diff
@@ -18,13 +18,19 @@ fn main() {
...
+        return;
     }
+
+    let mut window = windows.single_mut();
+    window.set_minimized(true);
+
+    *done = true;
```
mockersf, line 33:
```suggestion
```

tim-blackbird (approved):

mockersf (approved):

PR #15857, opened 2024-10-11
split up `TextStyle`

ickshonpe:
# Objective

Currently text is recomputed unnecessarily on any changes to its color, which is extremely expensive.

## Solution
Split up `TextStyle` into two separate components `TextFont` and `TextColor`.

## Testing

I added this system to `many_buttons`:
```rust
fn set_text_colors_changed(mut colors: Query<&mut TextColor>) {
    for mut text_color in colors.iter_mut() {
        text_color.set_changed();
    }
}
```

reports ~4fps on main, ~50fps with this PR.

## Migration Guide
`TextStyle` has been renamed to `TextFont` and its `color` field has been moved to a separate component named `TextColor` which newtypes `Color`.


crates/bevy_text/src/text_access.rs
```diff
@@ -281,15 +303,15 @@ impl<'w, 's, R: TextRoot> TextWriter<'w, 's, R> {
     /// Gets the [`TextStyle`] of a text span within a text block at a specific index in the flattened span list.
     ///
     /// Panics if there is no span at the requested index.
-    pub fn style(&mut self, root_entity: Entity, index: usize) -> Mut<TextStyle> {
+    pub fn style(&mut self, root_entity: Entity, index: usize) -> Mut<TextFont> {
```
alice-i-cecile, line 306: Needs to be renamed :)

crates/bevy_text/src/text_access.rs
```diff
@@ -281,15 +303,15 @@ impl<'w, 's, R: TextRoot> TextWriter<'w, 's, R> {
     /// Gets the [`TextStyle`] of a text span within a text block at a specific index in the flattened span list.
```
alice-i-cecile, line 303: Doc links are broken here.

crates/bevy_text/src/text_access.rs
```diff
@@ -305,11 +327,7 @@ impl<'w, 's, R: TextRoot> TextWriter<'w, 's, R> {
     }
 
     /// Invokes a callback on each span's [`TextStyle`] in a text block, starting with the root entity.
-    pub fn for_each_style(
-        &mut self,
-        root_entity: Entity,
-        mut callback: impl FnMut(Mut<TextStyle>),
-    ) {
+    pub fn for_each_style(&mut self, root_entity: Entity, mut callback: impl FnMut(Mut<TextFont>)) {
```
alice-i-cecile, line 330: Also needs a rename.

examples/animation/animation_events.rs
```diff
@@ -108,7 +108,7 @@ fn setup(
 }
 
 // Slowly fade out the text opacity.
-fn animate_text_opacity(mut styles: Query<&mut TextStyle>, time: Res<Time>) {
+fn animate_text_opacity(mut styles: Query<&mut TextFont>, time: Res<Time>) {
```
alice-i-cecile, line 111: styles variable needs a rename.

alice-i-cecile:
I really like this change. It's more granular, allowing us to be lazier about recomputations, but IMO it's also much clearer to users.

crates/bevy_text/src/text_access.rs
rparrett: and similar methods added for `TextColor` I assume?

crates/bevy_text/src/text_access.rs
ickshonpe: Yeah going to take me a while to work through all of these 😅

rparrett:
It looks like the current status of the examples is that they've been find/replaced, but many are likely broken in an obvious way. Feel free to ping me when this is ready for review.

viridia:
This all sounds good to me. Adapting my framework to work with this will not be difficult.

ickshonpe:
> It looks like the current status of the examples is that they've been find/replaced, but many are likely broken in an obvious way. Feel free to ping me when this is ready for review.

Should all be updated now. There should be a few errors with so many changes but I fixed a couple from the text rework, so it's probably a wash.

examples/animation/easing_functions.rs
```diff
@@ -64,10 +64,7 @@ fn setup(mut commands: Commands) {
             commands
                 .spawn((
                     Text2d(format!("{:?}", function)),
-                    TextStyle {
-                        color,
-                        ..text_style.clone()
-                    },
+                    text_style.clone(),
```
rparrett, line 67: `color` got dropped here

examples/stress_tests/bevymark.rs
```diff
@@ -283,14 +276,30 @@ fn setup(
...
+                    font.clone(),
+                    TextColor(LIME.into()),
+                ));
+                p.spawn((TextSpan::new(""), font.clone(), TextColor(AQUA.into())));
+                p.spawn((
+                    TextSpan::new("\nFPS (raw): "),
+                    font.clone(),
+                    TextColor(LIME.into()),
+                ));
+                p.spawn((TextSpan::new(""), font.clone(), TextColor(LIME.into())));
```
rparrett, line 290: This and other FPS values were aqua before

examples/stress_tests/many_buttons.rs
```diff
@@ -96,6 +97,12 @@ fn main() {
     app.insert_resource(args).run();
 }
 
+fn set_text_colors_changed(mut colors: Query<&mut TextColor>) {
```
rparrett, line 100: This staying?

ickshonpe: Yeah I think it should, just in case of regressions.

crates/bevy_text/src/text.rs
```diff
@@ -165,22 +165,22 @@ impl TextLayout {
...
 # let font_handle: Handle<Font> = Default::default();
 # let mut world = World::default();
 #
 world.spawn((
     TextLayout::default(),
-    TextStyle {
+    TextFont {
         font: font_handle.clone().into(),
         font_size: 60.0,
         color: BLUE.into(),
```
rparrett, line 178: Feel like CI should have caught this one

crates/bevy_text/src/text.rs
```diff
@@ -165,22 +165,22 @@ impl TextLayout {
...
     }
 ))
 .with_child((
     TextSpan::new("Hello!"),
-    TextStyle {
+    TextFont {
         font: font_handle.into(),
         font_size: 60.0,
         color: RED.into(),
```
rparrett, line 186: Ditto

rparrett:
Our other `Color` wrappers provide conversions from `Color`. Do we want to do the same here? Not sure if this is actually useful now that we're deprecating bundles.

e.g.
```
impl<T: Into<Color>> From<T> for BackgroundColor {
    fn from(color: T) -> Self {
        Self(color.into())
    }
}
```

crates/bevy_text/src/text.rs
ickshonpe: doc tests need to use with `///` not `/* */` blocks

ickshonpe:
> Our other `Color` wrappers provide conversions from `Into<Color>`. Do we want to do the same here? Not sure if this is actually useful now that we're deprecating bundles.
> 
> e.g.
> 
> ```rust
> impl<T: Into<Color>> From<T> for BackgroundColor {
>     fn from(color: T) -> Self {
>         Self(color.into())
>     }
> }
> ```

Yeah I thought about this, but not sure. It's not really useful unless adding `TextColor` to a field?

UkoeHB:
> Yeah I thought about this, but not sure either. I can't think of any use since there aren't any TextColor fields or parameters?

Easier to use `TextColor::from(Srgba::new(...))` than `TextColor(Srgba::new(...).into())`.

ickshonpe:
> > Yeah I thought about this, but not sure either. I can't think of any use since there aren't any TextColor fields or parameters?
> 
> Easier to use `TextColor::from(Srgba::new(...))` than `TextColor(Srgba::new(...).into())`.

I guess and might be useful if someone is constructing a bunch of spans using an iterator or something.

crates/bevy_text/src/pipeline.rs
```diff
@@ -222,7 +225,7 @@ impl TextPipeline {
         // Extract font ids from the iterator while traversing it.
         let mut glyph_info = core::mem::take(&mut self.glyph_info);
         glyph_info.clear();
-        let text_spans = text_spans.inspect(|(_, _, _, style)| {
+        let text_spans = text_spans.inspect(|(_, _, _, style, _)| {
```
alice-i-cecile, line 228:
```suggestion
        let text_spans = text_spans.inspect(|(_, _, _, font, _)| {
```

crates/bevy_text/src/pipeline.rs
```diff
@@ -427,7 +430,7 @@ impl TextMeasureInfo {
 }
 
 fn load_font_to_fontdb(
-    style: &TextStyle,
+    style: &TextFont,
```
alice-i-cecile, line 433: Missed rename of arg.

crates/bevy_text/src/pipeline.rs
```diff
@@ -461,10 +464,11 @@ fn load_font_to_fontdb(
     }
 }
 
-/// Translates [`TextStyle`] to [`Attrs`].
+/// Translates [`TextFont`] to [`Attrs`].
 fn get_attrs<'a>(
     span_index: usize,
-    style: &TextStyle,
+    style: &TextFont,
```
alice-i-cecile, line 470: Missed rename of arg.

crates/bevy_text/src/pipeline.rs
alice-i-cecile, line 228: style -> font?

crates/bevy_text/src/pipeline.rs
```diff
@@ -151,12 +152,14 @@ impl TextPipeline {
...
         // in cosmic-text.
-        let spans_iter = spans.iter().map(|(span_index, span, style, font_info)| {
-            (
-                *span,
-                get_attrs(*span_index, style, font_info, scale_factor),
-            )
-        });
+        let spans_iter = spans
+            .iter()
+            .map(|(span_index, span, style, font_info, color)| {
```
alice-i-cecile, line 157: style -> font?

alice-i-cecile (approved):
I spotted a couple of missed internal renames, but those are just nits :) Nice stuff!

examples/animation/animated_ui.rs
```diff
@@ -40,7 +40,7 @@ fn main() {
 }
 
 impl AnimatableProperty for FontSizeProperty {
-    type Component = TextStyle;
+    type Component = TextFont;
```
pablo-lua, line 43: If I understand the bevy error 0005 correctly, we shouldn't generally be changing the size of a font this way, and instead use transform or ui_scale. Should this example continue to teach this possible footgun to the users?

examples/animation/easing_functions.rs
```diff
@@ -16,7 +16,7 @@ fn main() {
 fn setup(mut commands: Commands) {
     commands.spawn(Camera2d);
 
-    let text_style = TextStyle {
+    let text_style = TextFont {
```
pablo-lua, line 19: Just for sake of understanding, can we also change this var name to text_font? Probably will be more understandable this way.

examples/games/game_menu.rs
```diff
@@ -433,11 +432,11 @@ mod menu {
                         // Display the game name
                         parent.spawn((
                             Text::new("Bevy Game Menu UI"),
-                            TextStyle {
+                            TextFont {
                                 font_size: 67.0,
```
pablo-lua, line 436: Question: generally we only set the size of the font when creating it. Is it possible that we maybe add a shorthand method for this? like `TextFont::from_size(size: f32)`

pablo-lua (commented):
Big changes. I have some questions

examples/animation/animated_ui.rs
ickshonpe: Yeah I agree it's not a good example, it should probably be removed. Scaling the foot using transform isn't normally a good solution either as the text will end up blurred (or blocky if you turn off anti-aliasing). It's outside the scope of this PR to deal with it though I think.

examples/animation/easing_functions.rs
ickshonpe: Yep missed the one,

examples/games/game_menu.rs
ickshonpe: Yeah I agree it's annoying. `FontSmoothing` is so niche, I'd rather it was a separate component as well. I want to keep this PR to just dealing with `color` though as it's the biggest problem.

rparrett (approved):
I have some slight reservations about ux, particularly the amount of
```rust
let text_style = (
    TextFont {
        // ...
    },
    TextColor(/* ... */)
);
```
now happening, which I think will be very common in user code.

But if we're happy with the design then the code looks good to me.

examples/async_tasks/external_source_external_thread.rs
```diff
@@ -54,7 +54,7 @@ fn read_stream(receiver: Res<StreamReceiver>, mut events: EventWriter<StreamEven
 }
 
 fn spawn_text(mut commands: Commands, mut reader: EventReader<StreamEvent>) {
-    let text_style = TextStyle::default();
+    let text_style = TextFont::default();
```
pablo-lua, line 57: nit: rename this var

pablo-lua (approved):
I'm fine with the changes and has some good improvements. I have one small nit, but otherwise LGTM

examples/async_tasks/external_source_external_thread.rs
ickshonpe: Oh it can be be removed even, it's not needed with required components.

ickshonpe:
> I have some slight reservations about ux, particularly the amount of
> 
> ```rust
> let text_style = (
>     TextFont {
>         // ...
>     },
>     TextColor(/* ... */)
> );
> ```
> 
> now happening, which I think will be very common in user code.
> 
> But if we're happy with the design then the code looks good to me. The benefits are certainly obvious.

Yeah I think what this needs is just text font + color inheritance flowing down the tree, It's trivial to implement a basic version, but maybe depends what's being done with bsn etc.

crates/bevy_dev_tools/src/fps_overlay.rs
```diff
@@ -128,9 +131,10 @@ fn customize_text(
     mut writer: UiTextWriter,
 ) {
     for entity in &query {
-        writer.for_each_style(entity, |mut style| {
-            *style = overlay_config.text_config.clone();
+        writer.for_each_font(entity, |mut font| {
+            *font = overlay_config.text_config.clone();
         });
+        writer.for_each_color(entity, |mut color| color.0 = overlay_config.text_color);
```
UkoeHB, line 137: Can we keep `for_each_style` and have it pass both the font and color as params?

crates/bevy_text/src/text.rs
```diff
@@ -359,12 +381,12 @@ pub fn detect_text_needs_rerender<Root: Component>(
         (
             Or<(
                 Changed<Root>,
-                Changed<TextStyle>,
+                Changed<TextFont>,
                 Changed<TextLayout>,
                 Changed<Children>,
             )>,
```
UkoeHB, line 387: Removing `TextColor` here means animated text colors will not cause glyphs to be updated properly.

UkoeHB (changes requested):
Needs a proper solution for change detection on `TextColor`.

crates/bevy_text/src/text.rs
ickshonpe: > Removing `TextColor` here means animated text colors will not cause glyphs to be updated properly.

I don't follow? Animated colors shouldn't cause an update, isn't that the point?

UkoeHB: > I don't follow? Animated colors shouldn't cause an update, isn't that the point?

The glyphs need to update otherwise the color changes won't show up.

UkoeHB: Oh hold on, this is my mistake: color is always copied from entities when doing text extraction.

crates/bevy_dev_tools/src/fps_overlay.rs
ickshonpe: We could but I'm wondering if all these helper functions are even needed though. Users can just use `for_each` and ignore the fields that they don't need? We could return a struct with named fields instread of an enum tuple to make it more ergonomic.

pablo-lua: Yeah, I believe that we should leave this helpers to a follow-up PR so we can discuss better what to do with them

UkoeHB (approved):

PR #15848, opened 2024-10-11
Store `QueryState` in `Query` as a `Cow` and remove distinct `QueryLens` type

ItsDoot:
# Objective

The `QueryLens` type is *basically* a `Query`, except it stores the `QueryState` itself rather than a reference to it. We can unify `Query` and `QueryLens` by allowing `Query` to hold either a reference or a value, which `Cow` provides. As a byproduct, we can also now also return `Query`s from `World::query`, rather than `QueryState`s.

## Solution

The following changes have been made:
- `QueryState` can now be `Clone`d (required by `Cow`)
    - All `WorldQuery::State` types are now required to be `Clone`.
- Functions on `Query` returning `'s` have been replaced with `'_`.
- The current `World::query`/`World::query_filtered` functions have been renamed to `World::query_state`/`World::query_state_filtered`.
- New `World::query`/`World::query_filtered` functions have been added that return `Query` instead of `QueryState`.
- New type conversions have been added:
    - `Query<D, F>::into_state(self) -> QueryState<D, F>` (clones the state if it was borrowed)
    - `QueryState<D, F>::as_query(&mut self, &mut World) -> Query<D, F>`
    - `QueryState<D, F>::as_readonly_query(&self, &World) -> Query<D::ReadOnly, F>`
    - `QueryState<D, F>::into_query(self, &mut World) -> Query<D, F>`
    - `QueryState<D, F>::into_readonly_query(self, &World) -> Query<D::ReadOnly, F>`
    - `QueryState<D, F>::into_readonly(self) -> QueryState<D::ReadOnly, F>`
        - Similar to `QueryState::as_readonly`, but by value instead of by reference.
- The `QueryLens` type has been removed. `Query` can function exactly like it.
    - `Query::transmute_lens_filtered(&mut self) -> QueryLens` changed to `Query::transmute_filtered(&mut self) -> Query`
    - `Query::transmute_lens(&mut self) -> QueryLens` changed to `Query::transmute(&mut self) -> Query`
    - `Query::join(&mut self, other: &mut Query) -> QueryLens` changed to `Query::join(&mut self, other: &mut Query) -> Query`
    - `Query::join_filtered(&mut self, other: &mut Query) -> QueryLens` changed to `Query::join_filtered(&mut self, other: &mut Query) -> Query`
- The state structs for `QueryData`/`QueryFilter` derives now `impl Clone`.

## Testing

- Current tests pass.
- TODO: change compile fail test for query lens.

---

## Showcase

TODO?

## Migration Guide

[...]

PR #15396, opened 2024-09-23
Let query items borrow from query state to avoid needing to clone

chescock:
# Objective

Improve the performance of `FilteredEntity(Ref|Mut)` and `Entity(Ref|Mut)Except`.  

`FilteredEntityRef` needs an `Access<ComponentId>` to determine what components it can access.  There is one stored in the query state, but query items cannot borrow from the state, so it has to `clone()` the access for each row.  Cloning the access involves memory allocations and can be expensive.  

## Solution

Let query items borrow from their query state.  

Add an `'s` lifetime to `WorldQuery::Item` and `WorldQuery::Fetch`, similar to the one in `SystemParam`, and provide `&'s Self::State` to the fetch so that it can borrow from the state.  

Unfortunately, there are a few cases where we currently return query items from temporary query states: the sorted iteration methods create a temporary state to query the sort keys, and the `EntityRef::components<Q>()` methods create a temporary state for their query.  

To allow these to continue to work with most `QueryData` implementations, introduce a new subtrait `ReleaseStateQueryData` that converts a `QueryItem<'w, 's>` to `QueryItem<'w, 'static>`, and is implemented for everything except `FilteredEntity(Ref|Mut)` and `Entity(Ref|Mut)Except`.  

`#[derive(QueryData)]` will generate `ReleaseStateQueryData` implementations that apply when all of the subqueries implement `ReleaseStateQueryData`.

This PR does not actually change the implementation of `FilteredEntity(Ref|Mut)` or `Entity(Ref|Mut)Except`!  That will be done as a follow-up PR so that the changes are easier to review.  I have pushed the changes as chescock/bevy#5.

## Testing

I ran performance traces of many_foxes, both against main and against chescock/bevy#5, both including #15282.  These changes do appear to make generalized animation a bit faster: 

(Red is main, yellow is chescock/bevy#5)
image

## Migration Guide

The `WorldQuery::Item` and `WorldQuery::Fetch` associated types and the `QueryItem` and `ROQueryItem` type aliases now have an additional lifetime parameter corresponding to the `'s` lifetime in `Query`.  Manual implementations of `WorldQuery` will need to update the method signatures to include the new lifetimes.  Other uses of the types will need to be updated to include a lifetime parameter, although it can usually be passed as `'_`.  In particular, `ROQueryItem` is used when implementing `RenderCommand`.  

Before: 

```rust
fn render<'w>(
    item: &P,
    view: ROQueryItem<'w, Self::ViewQuery>,
[...]

PR #13607, opened 2024-05-31
SystemParam `QueryRecursive` for safe recursive iteration.

mintlu8:
# Objective

Recursive iterating through the hierarchy while keeping a reference to parent requires `unsafe`, as seen in `propagate_transform` in `bevy_transform`. This PR alleviates this issue by generalizing recursive iteration into a safe interface.

## Solution 

Added `QueryRecursive` that contains 2 queries from 5 generic parameters.

```rust
root: Query<(Shared, Root), RootFilter>
child: Query<(Shared, Child), ChildFilter>
```

`QueryData` `Shared` gets passed from parent to children as a reference during iteration. 

Additionally we support passing evaluation results from parent to children in `iter_mut_with`.

### Example: transform pipeline (naive implementation)

```rust
fn propagate_transforms(mut query: QueryRecursive<
    (&Transform, &mut GlobalTransform),
    (),
    (),
    Without<Parent>,
    ()
>) {
    query.for_each_mut(
        |(transform, global_transform), ()|
            **global_transform = (**transform).into(),
        |(_, parent), (transform, global_transform), ()|
            **global_transform = parent.mul_transform(**transform),
    );
}
```

## Testing

Added some tests.

---

## Changelog

Added `QueryRecursive`.

PR #15858, opened 2024-10-11
Introduce methods on QueryState to obtain a Query

chescock:
# Objective

Simplify and expand the API for `QueryState`.  

`QueryState` has a lot of methods that mirror those on `Query`.  These are then multiplied by variants that take `&World`, `&mut World`, and `UnsafeWorldCell`.  In addition, many of them have `_manual` variants that take `&QueryState` and avoid calling `update_archetypes()`.  Not all of the combinations exist, however, so some operations are not possible.  

## Solution

Introduce methods to get a `Query` from a `QueryState`.  That will reduce duplication between the types, and ensure that the full `Query` API is always available for `QueryState`.  

Introduce methods on `Query` that consume the query to return types with the full `'w` lifetime.  This avoids issues with borrowing where things like `query_state.query(&world).get(entity)` don't work because they borrow from the temporary `Query`.  

Finally, implement `Copy` for read-only `Query`s.  `get_inner` and `iter_inner` currently take `&self`, so changing them to consume `self` would be a breaking change.  By making `Query: Copy`, they can consume a copy of `self` and continue to work.  

The consuming methods also let us simplify the implementation of methods on `Query`, by doing `fn foo(&self) { self.as_readonly().foo_inner() }` and `fn foo_mut(&mut self) { self.reborrow().foo_inner() }`.  That structure makes it more difficult to accidentally extend lifetimes, since the safe `as_readonly()` and `reborrow()` methods shrink them appropriately.  The optimizer is able to see that they are both identity functions and inline them, so there should be no performance cost.  

Note that this change would conflict with #15848.  If `QueryState` is stored as a `Cow`, then the consuming methods cannot be implemented, and `Copy` cannot be implemented.  

## Future Work

The next step is to mark the methods on `QueryState` as `#[deprecated]`, and move the implementations into `Query`.  

## Migration Guide

`Query::to_readonly` has been renamed to `Query::as_readonly`.


Trashtalk217:
I agree that there's a lot of duplication between QueryState's and Query's API and it's good that this is lessened, but just out of curiosity (just to offer a potentially dumb idea), would it perhaps be better if QueryState and Query were just one single struct instead of two?

chescock:
> I agree that there's a lot of duplication between QueryState's and Query's API and it's good that this is lessened, but just out of curiosity (just to offer a potentially dumb idea), would it perhaps be better if QueryState and Query were just one single struct instead of two?

I think the idea is that you need access to the world to actually do querying, but that we want to cache some state that lives longer than that borrow.  So a `Query` is a `&QueryState` for the state plus `UnsafeWorldCell<'w>` and `Ticks` from the world.  

I don't see a clean way to get rid of either.  If you don't have `Query` then you'd need to pass a `world` to every method on `QueryState`, which is terrible ergonomics for even the simplest system.  If you don't have `QueryState` then you need to look up `ComponentId`s and match archetypes every time you run a system, which is terrible performance.

chescock:
I realized I could split this into two independent PRs: One for the methods on `QueryState` to obtain a `Query`, and one for the consuming methods on `Query`.  The consuming methods aren't all that useful by themselves, but the methods on `QueryState` should be, and those shouldn't be as contentious.  Would anyone like me to split it up like that?

BenjaminBrienen:
I say go big! Go ahead and remove the methods on QueryState! 😆

You can split it up if you think it would help, but I don't mind bigger PRs.

crates/bevy_ecs/src/system/mod.rs
```diff
@@ -1542,24 +1542,6 @@ mod tests {
         });
     }
 
-    #[test]
-    #[should_panic = "Encountered a mismatched World."]
-    fn query_validates_world_id() {
```
BenjaminBrienen, line 1547: Why remove this test? Is it no longer applicable?

crates/bevy_ecs/src/system/query.rs
```diff
@@ -382,23 +390,18 @@ impl<D: QueryData, F: QueryFilter> core::fmt::Debug for Query<'_, '_, D, F> {
...
+    ///   called in ways that ensure the queries have unique mutable access.
+    /// * `world` must be the world used to create `state`.
     #[inline]
     pub(crate) unsafe fn new(
         world: UnsafeWorldCell<'w>,
         state: &'s QueryState<D, F>,
         last_run: Tick,
         this_run: Tick,
     ) -> Self {
-        state.validate_world(world.id());
```
BenjaminBrienen, line 400: Why remove this check?

BenjaminBrienen (approved):

crates/bevy_ecs/src/system/query.rs
chescock: Oh, right, that was subtle and I should have pointed it out!  

The reason is that I changed all of the `&self` and `&mut self` methods on `Query` to call `self.as_readonly().actual_method()` and `self.reborrow().actual_method()`, and `as_readonly()` and `reborrow()` are implemented in terms of `Query::new()`.  I wanted to make sure that change was zero-cost, but checking the assembly with `cargo-show-asm` showed that it was adding a call to `validate_world` to every existing method on `Query`!  Since we always had a world that was already validated, I changed it to be a safety requirement, and now the assembly appears to be unchanged.  

(Then I removed the test you asked about above because it's no longer true that it panics.)

crates/bevy_ecs/src/query/state.rs
```diff
@@ -319,6 +318,123 @@ impl<D: QueryData, F: QueryFilter> QueryState<D, F> {
...
+    /// # Safety
+    ///
+    /// This does not check for mutable query correctness. To be safe, make sure mutable queries
+    /// have unique access to the components they query.
+    pub unsafe fn query_unchecked<'w, 's>(
+        &'s mut self,
+        world: UnsafeWorldCell<'w>,
+    ) -> Query<'w, 's, D, F> {
+        self.update_archetypes_unsafe_world_cell(world);
+        self.query_unchecked_manual(world)
```
13ros27, line 367: This should probably be in an `unsafe` block for consistency with the other functions, I'm not quite sure what bevy's policy on `unsafe_op_in_unsafe_fn` actually is (although iirc it will be warn in 2024 edition)

crates/bevy_ecs/src/system/query.rs
```diff
@@ -693,11 +736,32 @@ impl<'w, 's, D: QueryData, F: QueryFilter> Query<'w, 's, D, F> {
...
+    /// - [`iter_many_inner`](Self::iter_many_inner) to get mutable query items with the full `'world` lifetime.
     #[inline]
     pub fn iter_many_mut<EntityList: IntoIterator<Item: EntityBorrow>>(
         &mut self,
         entities: EntityList,
     ) -> QueryManyIter<'_, 's, D, F, EntityList::IntoIter> {
+        self.reborrow().iter_many_inner(entities)
+    }
+
+    /// Returns an iterator over the query items generated from an [`Entity`] list.
```
13ros27, line 751: Should these `iter_many_inner`, `iter_combinations_inner` etc. call out in their docs why they are separate methods like `get_inner` does ("with the actual "inner" world lifetime, by consuming the [`Query`]")? To be clear, I'm not sure if they should but thought I should raise it as a possibility.

crates/bevy_ecs/src/system/query.rs
```diff
@@ -1406,6 +1478,36 @@ impl<'w, 's, D: QueryData, F: QueryFilter> Query<'w, 's, D, F> {
...
+    ///     let mut health = query.get_single_inner().expect("Error: Could not find a single player.");
+    ///     health.0 += 1;
+    /// }
+    /// # bevy_ecs::system::assert_is_system(regenerate_player_health_system);
+    /// ```
+    ///
+    /// # See also
+    ///
+    /// - [`get_single`](Self::get_single) to get the read-only query item.
+    /// - [`get_single_mut`](Self::get_single_mut) to get items using a mutable reference.
```
13ros27, line 1508: Maybe "to get the item using a mutable reference."?

crates/bevy_ecs/src/system/query.rs
```diff
@@ -1761,6 +2011,21 @@ impl<'w, 's, D: QueryData, F: QueryFilter> Query<'w, 's, D, F> {
...
 }
 
+impl<'w, 's, D: QueryData, F: QueryFilter> IntoIterator for Query<'w, 's, D, F> {
+    type Item = D::Item<'w>;
+    type IntoIter = QueryIter<'w, 's, D, F>;
+
+    fn into_iter(self) -> Self::IntoIter {
+        // SAFETY:
+        // - `self.world` has permission to access the required components.
+        // - The query is read-only, so it can be aliased even if it was originally mutable.
```
13ros27, line 2021: I'm probably missing something but why is this read-only? The `QueryData` isn't limited and `Item` is the one from `WorldQuery` which would be mutable for a mutable query.

13ros27 (approved):
This looks good to me, mostly just nits although I do have one question about that `IntoIterator` for `Query`?

crates/bevy_ecs/src/system/query.rs
13ros27: Ah, this safety comment comes from `Query::iter`, it should probably take the one from `Query::iter_mut` instead which doesn't mention the query being read-only. This is sound because we take `self` so you can't be holding another `&Query` to iterate over and potentially alias.

chescock: Yup, that safety comment is wrong.  I must have copied it from a read-only version.  I'll change it to "We consume the query, so mutable queries cannot alias.  Read-only queries are `Copy`, but may alias themselves."

crates/bevy_ecs/src/system/query.rs
```diff
@@ -381,6 +381,14 @@ pub struct Query<'world, 'state, D: QueryData, F: QueryFilter = ()> {
     this_run: Tick,
 }
 
+impl<D: ReadOnlyQueryData, F: QueryFilter> Clone for Query<'_, '_, D, F> {
+    fn clone(&self) -> Self {
+        *self
```
cBournhonesque, line 386: sorry, noob question: how does this work?
The `&QueryState` inside the Query is simply copied? i.e. we create a new Query that points to the same QueryState?

crates/bevy_ecs/src/query/state.rs
```diff
@@ -319,6 +318,124 @@ impl<D: QueryData, F: QueryFilter> QueryState<D, F> {
         state
     }
 
+    /// Creates a [`Query`] from the given [`QueryState`] and [`World`].
+    ///
+    /// This will create read-only queries, see [`Self::query_mut`] for mutable queries.
+    pub fn query<'w, 's>(&'s mut self, world: &'w World) -> Query<'w, 's, D::ReadOnly, F> {
```
alice-i-cecile, line 324: Oh my gosh finally.

crates/bevy_ecs/src/system/query.rs
```diff
@@ -528,11 +563,7 @@ impl<'w, 's, D: QueryData, F: QueryFilter> Query<'w, 's, D, F> {
     /// [`iter`](Self::iter) for read-only query items.
     #[inline]
     pub fn iter_mut(&mut self) -> QueryIter<'_, 's, D, F> {
-        // SAFETY: `self.world` has permission to access the required components.
-        unsafe {
-            self.state
-                .iter_unchecked_manual(self.world, self.last_run, self.this_run)
-        }
+        self.reborrow().into_iter()
```
cBournhonesque, line 566: Sorry could you please explain again why we need to re-borrow?

`self.reborrow().into_iter()` creates a new Query with `reborrow()` and then borrows from that with `into_iter()`
`self.into_iter()` would simply create a QueryIter that borrows from the initital `&'a mut Query` so the lifetime is constrained by `a; is that the issue?

Victoronz (changes requested):
~~Sorry to block this~~ (the merge queue got there first), but I don't think `Query` being `Copy` is sound?
IIUC you can go from a mutable `Query` to a read-only `QueryLens`, make a `Query` out of that lens again, then copy that `Query`, `transmute_lens` it again, and get a `Query` from it for a last time. Then you'd have a read-only query that is no longer bound to the `&mut self` of the original mutable `Query`.
Did I overlook something?

crates/bevy_ecs/src/system/query.rs
chescock: `reborrow()` goes from `&'a mut Query<'w>` to `Query<'a>`, turning a borrowed `Query` into an owned one with a shorter lifetime.   

`into_iter()` takes `self` instead of `&mut self`, so it actually *consumes* the `Query`.  

We have an `&mut Query`, so we first `reborrow()` to get an owned `Query` with a shorter lifetime, and then use `into_iter()` to consume the owned query and create a `QueryIter` - still with the shorter lifetime.  

It's more complex when looking at one method, but across the whole type it means we can re-use the safety proofs in `reborrow()` and `as_readonly()`.  That lets us implement this method without `unsafe` code, and means the compiler would catch it if we tried to return `QueryIter<'w, ...>` instead of `QueryIter<'_, ...>`.

alice-i-cecile:
Discuss in #17693.

Victoronz:
Something else I'd like to note is that this PR implements `IntoIterator` for `Query`, however that impl is not restricted to `ReadOnlyQueryData`. This means that a simple `for` loop over a mutable Query will consume it, even if that loop doesn't actually mutate anything!
I'd think this to be quite confusing for users. Especially since there is no "owned" way of iterating `Query`, we're still just borrowing the data in the end.
Implementing it only for `ReadOnlyQueryData` would be confusing as well, since that'd mean that writing mutable for loops now diverges from writing immutable ones!
Is that impl necessary?

chescock:
> Something else I'd like to note is that this PR implements `IntoIterator` for `Query`, however that impl is not restricted to `ReadOnlyQueryData`. This means that a simple `for` loop over a mutable `Query` will consume it, even if that loop doesn't actually mutate anything!

Yup!  That's the generalization of `iter_inner()` to mutable data, and lets you write `query_state.query_mut(world).into_iter()` without needing to promote the temporary `query` to a variable, which may be impossible if you're returning data.  That's necessary to completely replace `query_state.iter_mut(world)`.  

I'm hoping it will be a nicer experience for users: You can write the obvious `for data in query`, and in the common case where you only iterate once it will finally just work!  If you need to iterate multiple times, you'll need to `for data in &mut query`, but that's true for other common collection types like `Vec`, so it should be familiar.  

And because read-only queries also `impl Copy`, you *can* do multiple `for data in query` loops over them, and everything will work!

chescock:
And while I'd been focusing on `QueryState` when writing this, the `IntoIter` impl should also make it nicer to work with `QueryLens`, since you won't always have to put the `Query` in a variable.  

For example: 

```rust
let mut lens: QueryLens<_> = ...;

// Fails with "error[E0716]: temporary value dropped while borrowed"
let first = lens.query().iter_mut().next();
assert!(first.is_some());

// Works!
let first = lens.query().into_iter().next();
assert!(first.is_some());
```

Victoronz:
The reasoning suggests that this is mainly to work around `Query` always borrowing `QueryState`.
The reason why `QueryState` has the fetching methods at all I think was because we couldn't directly create `Query` from `World`, also because it borrows `QueryState`.
The reason we have `QueryLens` is once again the same.

I have some thoughts on `Query` as a whole, but I hadn't quite put them together yet, so I'll just pick this place to state them.
I'd like to make an issue out of this where we can discuss, but it also serves as a response here:

It feels like we are stacking features/implementation logic on this detail, when really the proper solution should be to make a `Query` that owns its `QueryState` (what we basically want to allow is at least `Box<QueryState>` but preferably `Deref<QueryState>`).
The reason this was not done previously was to avoid an additional generic or branching in `Query`.
That doesn't mean it can't be done!
Instead of placing the generic in `Query` itself, we could abstract over types that can produce the necessary parts for fetching, and then have the fetch methods (the various `get`s/`iter`s) be generic. And/Or we could even avoid duplication by deriving the fetch methods from a base trait impl, like `Iterator` does with `next`/`size_hint`.
Looking forward, when querying becomes more complex, it is very likely that we won't get away with a single `Query` type in the first place. #13607 or the uncached query efforts are examples.
Within this, `QueryState` itself should no longer be the place where call fetch methods, like you state in "Future Work".

Ideally, consuming methods will not be needed, but more likely will still have their place.
However, I don't think we should be making them the default:
Almost everywhere, an owned `into_iter` call will yield owned items, but here it instead means "longer-lived" borrowed items, which is quite inconsistent with this notion.
IIUC, then #15396 will mean that the consuming methods will not work for what is not `ReleaseStateQueryData`, meaning that the default way of iterating would not work for some specific types, because they don't only rely on the `'w` lifetime. 

All in all, I'm struggling to keep a mental model of all this `Query`/`QueryState` logic myself, and would like to expose the least amount of that as necessary to the users. Let me know what you think!

alice-i-cecile:
Thank you to everyone involved with the authoring or reviewing of this PR! This work is relatively important and needs release notes! Head over to https://github.com/bevyengine/bevy-website/issues/1990 if you'd like to help out.

PR #15862, opened 2024-10-12
fix mesh2d_manual example

### LINKED ISSUE #15847: `mesh2d_manual` example doesn't render correctly
## Bevy version

main, since #15524

Also tested win11+4080

## What you did

`cargo run --example mesh2d_manual`

## What went wrong

Example should display a yellow star. But instead it displays some wispy blue lines.

awtterpip:
# Objective

- Fixes #15847 

## Solution

- Change mesh2d_manual to use the proper vertex format for color.

## Testing

- The mesh2d_manual example now works for me, tested on linux

### DIFF
--- a/examples/2d/mesh2d_manual.rs
+++ b/examples/2d/mesh2d_manual.rs
@@ -263,7 +263,6 @@ fn vertex(vertex: Vertex) -> VertexOutput {
     // Project the world position of the mesh into screen position
     let model = mesh2d_functions::get_world_from_local(vertex.instance_index);
     out.clip_position = mesh2d_functions::mesh2d_position_local_to_clip(model, vec4<f32>(vertex.position, 1.0));
-    // Unpack the `u32` from the vertex buffer into the `vec4<f32>` used by the fragment shader
     out.color = vertex.color;
     return out;
 }
--- a/examples/2d/mesh2d_manual.rs
+++ b/examples/2d/mesh2d_manual.rs
@@ -150,7 +150,7 @@ impl SpecializedRenderPipeline for ColoredMesh2dPipeline {
             // Position
             VertexFormat::Float32x3,
             // Color
-            VertexFormat::Uint32,
+            VertexFormat::Float32x4,
         ];
 
         let vertex_layout =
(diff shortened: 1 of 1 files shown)


examples/2d/mesh2d_manual.rs
```diff
@@ -264,7 +264,7 @@ fn vertex(vertex: Vertex) -> VertexOutput {
     let model = mesh2d_functions::get_world_from_local(vertex.instance_index);
     out.clip_position = mesh2d_functions::mesh2d_position_local_to_clip(model, vec4<f32>(vertex.position, 1.0));
     // Unpack the `u32` from the vertex buffer into the `vec4<f32>` used by the fragment shader
```
rparrett, line 266: This comment is no longer needed

rparrett:
This looks good, but I'm curious about why we changed the vertex attribute format in that migration. I had the impression that we were intentionally using a custom vertex attribute in that example.

examples/2d/mesh2d_manual.rs
alice-i-cecile, line 266:
```suggestion
```

rparrett:
Would like to know if the change away from the custom vertex attribute in #15847 was intentional or desired before approving.

This fixes the issue, but it could also be fixed by switching back to the custom `u32` based color, probably. And I think that example may have been doing that intentionally.

Jondolf:
@rparrett It was half-intentional, but can be reverted. I originally changed it to fix a crash caused by conflicting vertex color formats with default mesh materials (see [this](https://github.com/bevyengine/bevy/pull/15524#discussion_r1783183321)) but then fixed it in a different way as per Cart's suggestion. I forgot to change it back and thought it was cleaner like that anyway, since it's the default format that [`Mesh::ATTRIBUTE_COLOR`](https://github.com/bevyengine/bevy/blob/813c75958df8a518353c6b50cb62a27353bbfd51/crates/bevy_mesh/src/mesh.rs#L176-L177) uses.

My bad though! Didn't anticipate that it would break things or be particularly undesirable (I recall testing it and it worked like normal...)

alice-i-cecile (approved):

alice-i-cecile:
#15883 was selected over this PR because it better preserves the intent of the example.

PR #15863, opened 2024-10-12
fix spatial audio examples

### LINKED ISSUE #15837: `spatial_audio_2d` and `spatial_audio_3d` examples: emitter jumps when movement is unpaused
## Bevy version

0.14, main

## What you did

- `git checkout 754194fd24b7baaa24fcadca0b1378413e39421b` 
   (example is broken for other reasons in more recent commits)
- `cargo run --example spatial_audio_2d`
- Press `space` to pause emitter movement
- Wait one second
- Press `space` again to unpause

## What went wrong

Emitter jumps in space rather than continuing from pre-pause position

## Additional info

The arrow keys should still be able to move the listeners while the emitter movement is paused, so I'm not sure if we can just pause `Time`.

I would suggest a change starting from

```diff
 #[derive(Component, Default)]
 struct Emitter {
-    stopped: bool,
+    stopwatch: Stopwatch,
 }
 ```

awtterpip:
# Objective

- Fixes #15837 

## Solution

- Change `Emitter` components to use `Stopwatch` to allow the time to be tracked independently.

## Testing

- Changes were tested. 
- Run either the `spatial_audio_2d` or `spatial_audio_3d` example to test

### DIFF
--- a/examples/audio/spatial_audio_2d.rs
+++ b/examples/audio/spatial_audio_2d.rs
@@ -3,6 +3,7 @@ use bevy::{
     audio::{AudioPlugin, SpatialScale},
     color::palettes::css::*,
     prelude::*,
+    time::Stopwatch,
 };
 
 /// Spatial audio uses the distance to attenuate the sound volume. In 2D with the default camera,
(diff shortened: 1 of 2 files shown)


rparrett (approved):
Thanks for picking this one up.

alice-i-cecile (approved):

PR #15865, opened 2024-10-12
examples(shaders/glsl): Update GLSL Shader Example Camera View uniform

jakeswenson:
# Objective
The Custom Material GLSL shader example has an old version of the camera view uniform structure. 
This PR updates the example GLSL custom material shader to have the latest structure. 

## Solution

I was running into issues using the camera world position (it wasn't changing) and someone in discord pointed me to the source of truth.
  `crates/bevy_render/src/view/view.wgsl`

After using this latest uniform structure in my project I'm now able to work with the camera position in my shader.

## Testing
I tested this change by running the example with:
```bash
cargo run --features shader_format_glsl --example shader_material_glsl
```
image

### DIFF
--- a/assets/shaders/custom_material.vert
+++ b/assets/shaders/custom_material.vert
@@ -7,14 +7,19 @@ layout(location = 2) in vec2 Vertex_Uv;
 layout(location = 0) out vec2 v_Uv;
 
 layout(set = 0, binding = 0) uniform CameraViewProj {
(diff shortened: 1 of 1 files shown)


tychedelia (approved):
Thanks!

cart (approved):
I made the change specified by @BorisBoutillier and expanded the comment to describe the constraints.

PR #15872, opened 2024-10-12
Migrate from `Query::single` and friends to `Single`

### LINKED ISSUE #15866: Replace all uses of `Query::single` with `Single`

`Query::single` and `Query::single_mut` are convenient lazy helpers with a nasty side effect: they panic unless exactly one matching entity is found!

While this is fine, in theory, for polished, tiny example code, it teaches bad habits to our users.

We should swap over all panicking uses over to the shiny new [`Single`](https://dev-docs.bevyengine.org/bevy/ecs/prelude/struct.Single.html) system param.

In rare cases, swapping to `Query::get_single` or even `Option<Single<&Component>>` might be preferred, to allow for more detailed error handling or to ensure the system runs to do secondary work even if the single entity it's looking for isn't found.

We could remove these methods entirely, per #14275. This is a much larger and more controversial change. Even if this is chosen, this migration needs to be done anyways, so we might as well do it now.

## Additional context

The nascent `bevy_cli` project introduced a lint to check for this class of error: https://github.com/TheBevyFlock/bevy_cli/pull/95. At some point, we'll want to run that Bevy-specific linter in CI for this project, but it's not ready yet.

pablo-lua:
# Objective

- closes #15866

## Solution

- Simply migrate where possible.

## Testing

- Expect that CI will do most of the work. Examples is another way of testing this, as most of the work is in that area.
---

## Notes
For now, this PR doesn't migrate `QueryState::single` and friends as for now, this look like another issue. So for example, QueryBuilders that used single or `World::query` that used single wasn't migrated. If there is a easy way to migrate those, please let me know.

Most of the uses of `Query::single` were removed, the only other uses that I found was related to tests of said methods, so will probably be removed when we remove `Query::single`.

### DIFF
--- a/benches/benches/bevy_ecs/scheduling/run_condition.rs
+++ b/benches/benches/bevy_ecs/scheduling/run_condition.rs
@@ -67,8 +67,8 @@ pub fn run_condition_yes_with_query(criterion: &mut Criterion) {
     group.warm_up_time(core::time::Duration::from_millis(500));
     group.measurement_time(core::time::Duration::from_secs(3));
     fn empty() {}
-    fn yes_with_query(query: Query<&TestBool>) -> bool {
-        query.single().0
+    fn yes_with_query(query: Single<&TestBool>) -> bool {
+        query.0
     }
     for amount in 0..21 {
         let mut schedule = Schedule::default();
--- a/crates/bevy_app/src/app.rs
+++ b/crates/bevy_app/src/app.rs
@@ -801,7 +801,7 @@ impl App {
     /// }
     ///
     /// fn validate(query: Option<Single<(&A, &B, &C)>>) {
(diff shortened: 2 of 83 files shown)


alice-i-cecile (approved):

alice-i-cecile:
I've had to open this up in Codespaces to review it, as Github's webpage hates PRs with many files changed 🙃 That said, looks great! I particularly appreciate the use of `Option<Single>` in tests to ensure that we're still panicking if things aren't found.

rparrett:
Guessing this is going to have conflicts after #15857. Can look after those get fixed up.

pablo-lua:
@BenjaminBrienen All comments addressed.

BenjaminBrienen (approved):

PR #15877, opened 2024-10-13
Add fully type-erased event triggering

ItsDoot:
# Objective

Use case: A socket receives packets that are disambiguated via a packet ID and the current client state. I need to trigger these packets as events, but I don't statically know the type when triggering the events. `EmitDynamicTrigger` currently exists to try to fulfill this need, but I don't have a type that can stand-in to be passed to `EmitDynamicTrigger`, so it feels a bit implicitly unsafe.
We should be able to trigger events with a `ComponentId` and type-erased `PtrMut` to the event data, instead.

## Solution

Added:
```rust
impl World {
    pub unsafe fn trigger_ref_dynamic<T: Traversal>(
        &mut self,
        event_type: ComponentId,
        event_data: PtrMut<'_>,
        auto_propagate: bool,
    );
    
    pub unsafe fn trigger_targets_ref_dynamic<T: Traversal>(
        &mut self,
        event_type: ComponentId,
        event_data: PtrMut<'_>,
        targets: impl TriggerTargets,
        auto_propagate: bool,
    );
}
```

Note: I have not added a `Command` variant of these because `PtrMut<'a>` is not `'static`, and it's not important for my use case.
`Box<dyn Any>` would probably work, though.

## Testing

- Added two tests.

### DIFF
--- a/crates/bevy_ecs/src/observer/mod.rs
+++ b/crates/bevy_ecs/src/observer/mod.rs
@@ -14,9 +14,10 @@ use crate::{
     observer::entity_observer::ObservedBy,
     prelude::*,
     system::IntoObserverSystem,
+    traversal::Traversal,
     world::{DeferredWorld, *},
 };
-use bevy_ptr::Ptr;
+use bevy_ptr::{Ptr, PtrMut};
 use bevy_utils::HashMap;
 use core::{
     fmt::Debug,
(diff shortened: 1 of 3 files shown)


ItsDoot:
Well, actually I just figured out a better solution involving storing `fn(Commands, RawPacket)` instead, so this isn't necessary for me anymore. 😅 

If anyone else finds a use for it, I'll reopen.

PR #15879, opened 2024-10-13
Make `tracing` Optional for `bevy_utils`

bushrat011899:
# Objective

- Allows `bevy_utils` to compile on platforms without atomic CAS operations (e.g., GameBoy Advance `thumbv4t-none-eabi`)
- Contributes to #15460 

## Solution

- Made `tracing` optional (behind the `tracing` feature, enabled by default).
- Switched to using `log` within `bevy_utils`, and exported it as `tracing`.

## Testing

- `cargo build --target thumbv4t-none-eabi`

## Migration Guide

If you were using `bevy_utils` with default features disabled, and you relied on the `tracing` re-export, and you relied on a `tracing` item that is _not_ available in `log`, enable the `tracing` feature.

## Notes

`tracing` has built-in support for capturing logs from `log` and this is enabled by default. The biggest loss going from `log` to `tracing` is the loss of `info_span`/etc. Otherwise, the crates are largely identical in behaviour, with `log` having no dependencies and more popularity on crates.io.

### DIFF
--- a/crates/bevy_utils/src/lib.rs
+++ b/crates/bevy_utils/src/lib.rs
@@ -39,13 +39,15 @@ pub use ahash::{AHasher, RandomState};
 pub use bevy_utils_proc_macros::*;
 pub use default::default;
 pub use hashbrown;
-pub use log;
 #[cfg(feature = "std")]
 pub use parallel_queue::*;
 pub use time::*;
 
 #[cfg(feature = "tracing")]
 pub use tracing;
+
+#[cfg(not(feature = "tracing"))]
+pub use log as tracing;
 
 #[cfg(feature = "alloc")]
 use alloc::boxed::Box;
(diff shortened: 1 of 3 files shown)


alice-i-cecile:
I don't quite follow the strategy here. It looks like we're unconditionally calling the `log` variants, but we want to call the `tracing` ones if and only if the feature is enabled (just like before).

crates/bevy_utils/src/lib.rs
```diff
@@ -42,8 +42,13 @@ pub use hashbrown;
 #[cfg(feature = "std")]
 pub use parallel_queue::*;
 pub use time::*;
+
+#[cfg(feature = "tracing")]
 pub use tracing;
 
+#[cfg(not(feature = "tracing"))]
+pub use log as tracing;
```
alice-i-cecile, line 50: Comment here please :) This is a mysterious setup for someone to stumble across.

bushrat011899: Good point! Included a little comment explaining why this is being done.

alice-i-cecile (approved):

BenjaminBrienen (approved):

Elabajaba:
Just a small sidenote, log has a much higher performance overhead than tracing for spammy logs and log filtering (see https://github.com/gfx-rs/wgpu/issues/6046#issuecomment-2309022590 for why).

mockersf:
I think the re-exporting of log as tracing should be avoided.

I would prefer to take the breaking change now, and re-exports the common functions macro between tracing and log as bevy_utils::log, and use that through the engine

bushrat011899:
These are both excellent points. I will update the documentation to indicate that users should prefer to use the `tracing` feature unless they _must_ disable it. As for the breaking change, I agree. I will update the PR to expose the common items in `log` and `tracing` from `bevy_utils::log`. For items which are exclusively available from `tracing`, should we export those from `bevy_utils::log`, only export them from `bevy_utils::tracing`, or re-export the whole `tracing` module (causing items like `trace!` to be exported twice)?

I'm leaning towards only exporting `tracing` exclusive items from `bevy_utils::tracing`, and everything else from `bevy_utils::log`.

PR #15880, opened 2024-10-13
Disjoint `QueryData` access

### LINKED ISSUE #14518: Disjointed QueryDataAccess

In generic contexts, or when you have really complex `QueryData` parameters, there is currently no way to have disjointed `QueryData` access in a single query iteration 
For example, you have a generic system that changes Transform based on an output of the generic parameter:
```rust
trait EntitySystem {
    Data: QueryData;
    Out;

    fn run(data: Self::Data) -> Self::Out;
}

fn change_transform<T: EntitySystem<Out = Vec3>>(mut query: Query<DataSet<(T::Data, &mut Transform)>>) {
    for mut set in query.iter_mut() {
        let desirable_translation = T::run(set.d0());
        let mut transform = set.d1();
        transform.translation = desirable_translation;
    }
}
```

For this feature request I will name such `QueryData` implementer `DataSet` (similar to how `ParamSet` for `SystemParam` was named)

Implementation would be similar to the `ParamSet` implementation.
Requires cloning `WorldQuery::Fetch` in call to `WorldQuery::fetch`
Requires adding method to the `WorldQuery` trait similar to `WorldQuery::shrink`, but with `WorldQuery::Fetch` as input and output, so that it is possible to implement `WorldQuery::shrink` for `DataSet`
```rust
pub struct DataSet<'w, T: QueryData> {
    fetch: T::Fetch<'w>,
    entity: Entity,
    table_row: TableRow,
}
```

Using `ParamSet` over two queries
Iterating each query in different cycle 
Storing results of iteration in a `HashMap`
Code equivalent of this alternative:
```rust
fn alternative_function(mut set: ParamSet<(Query<(Entity, D0)>, Query<(Entity, D1)>, ...)>) {
    let mut p0_map: EntityHashMap = HashMap::new();
    for (entity, data) in set.p0().iter_mut() {
        let result = calculation_for_p0(data);
        p0_map.insert(entity, result);
    }

    let mut p1_map: EntityHashMap = HashMap::new();
    for (entity, data) in set.p1().iter_mut() {
        if let Some(p0_result) = p0_hash_map.get_mut(&entity) {
            let result = calculation_for_p1(p0_result, data);
            p1_map.insert(entity, result);
        }
    }

    ...
}
```
This is error prone, requires unnecessary allocations, boilerplate and is inefficient, because in generic contexts it's not always possible to make all queries match the same archetype, which means more entities than necessary will be iterated (there is no archetype filter).

## Additional context

I've built proof-of-concept version without proc marco implementation, with just 2 generics https://github.com/vil-mo/bevy_data_set
Currently working on proc macro version

vil-mo:
# Objective

Fixes https://github.com/bevyengine/bevy/issues/14518

## Solution

Added `QueryData` implementer similarly to how `ParamSet` is implemented

## Testing

2 doc tests and 3 tests that are adapted forms of the same tests for `ParamSet`

## Showcase
Adds `DataSet` that imlements `QueryData` and allows disjoint access to it's members, similarly to how `ParamSet` works.
Mainly would be useful with in generic contexts where you need `QueryData`, but can't guarantee it won't conflict with other `QueryData`. Or with complex custom `QueryData` implementers that conflict with each other.  
This will compile even if A and B conflict:
```rust
fn iterate_generic_data<A: QueryData, B: QueryData>(mut query: Query<DataSet<(A, B)>>) {
    for mut set in query.iter_mut() {
        let a = set.d0();
        let b = set.d1();
    }
}
```


chescock (approved):

BenjaminBrienen (approved):

chescock:
It's definitely more niche than `ParamSet`, but I don't see a simpler way to solve the use case in the linked issue.  There could theoretically be cases with overlapping `#[derive(QueryData)]` queries where it could help, as well.  

I believe it was sound when it was written, since it ensures only one subquery is alive at a time and declares all the access from each.  

It will need an update in light of #16843, though.  It's now valid for a `QueryData` to request mutable access to a resource in `init_fetch`, and it would be unsound if multiple subqueries did that.  I think that can be resolved by adding the resource access from each subquery to the main `access` before checking the next one.  

Also, this uses a proc macro, which was consistent with `ParamSet` when written, but #16847 changed that to be a `macro_rules` macro.  It would be nice to change this to `macro_rules` as well, but that doesn't need to block it.

crates/bevy_ecs/src/system/mod.rs
```diff
@@ -810,6 +810,31 @@ mod tests {
...
+    #[test]
+    fn data_set_system() {
+        fn sys(mut _query: Query<DataSet<(&mut A, &A)>>) {}
+        let mut world = World::default();
+        run_system(&mut world, sys);
+    }
+
+    #[test]
+    #[should_panic]
+    fn conflicting_query_with_data_set_system() {
```
hymm, line 822: Could you add a should_panic test with the `Query<DataSet>` as the first system parameter too. The conflict checking can be order dependent.

crates/bevy_ecs/src/system/mod.rs
```diff
@@ -810,6 +810,31 @@ mod tests {
...
+
+    #[test]
+    #[should_panic]
+    fn conflicting_data_sets_system() {
+        fn sys(_query_1: Query<DataSet<(&mut A,)>>, _query_2: Query<DataSet<(&mut A, &B)>>) {}
+
+        let mut world = World::default();
+        run_system(&mut world, sys);
+    }
+
```
hymm, line 837: Could you add a should_panic test with the conflict in the second item of the DataSet?

crates/bevy_ecs/src/query/fetch.rs
```diff
@@ -1716,6 +1717,154 @@ unsafe impl<'__w, T: Component> QueryData for Mut<'__w, T> {
     type ReadOnly = Ref<'__w, T>;
 }
 
+/// A collection of potentially conflicting [`QueryData`]s allowed by disjoint access.
```
hymm, line 1720: The docs should make it clear that the query will fetch entitieis that match the union of all the QueryData's in the DataSet.

hymm (commented):
If we had real variadics, I would say just merge this. As it is though this adds a fair amount of codegen for a niche feature, so I'm a little hesitant. This is probably only really useful in generic contexts. Overall I do lean towards merging this, since the alternatives are a lot more awkward to write.

We could limit this to 4 items like ParamSet was originally until a user asks for more.

chescock (approved):

alice-i-cecile:
I'm a little reluctant to merge this, given the level of code gen for a niche feature and the absolute headache that ParamSet has been. I'm open to being convinced however.

PR #15881, opened 2024-10-13
Improve `PhantomData` held by curve adaptors

mweatherley:
# Objective

The previous `PhantomData` instances were written somewhat lazily, so they were just things like `PhantomData<T>` for curves with an output type of `T`. This looks innocuous, but it unnecessarily constrains `Send/Sync` inference based on `T`. See [here](https://doc.rust-lang.org/nomicon/phantom-data.html#table-of-phantomdata-patterns).

## Solution

Switch to `PhantomData` of the form `PhantomData<fn() -> T>` for most of these adaptors. Since they only have a functional relationship to `T` (i.e. it shows up in the return type of trait methods), this is more accurate.

## Testing

Tested by compiling Bevy.

### DIFF
--- a/crates/bevy_math/src/curve/adaptors.rs
+++ b/crates/bevy_math/src/curve/adaptors.rs
@@ -85,7 +85,7 @@ pub struct FunctionCurve<T, F> {
     #[cfg_attr(feature = "bevy_reflect", reflect(ignore))]
     pub(crate) f: F,
     #[cfg_attr(feature = "bevy_reflect", reflect(ignore))]
-    pub(crate) _phantom: PhantomData<T>,
+    pub(crate) _phantom: PhantomData<fn() -> T>,
 }
 
 impl<T, F> Debug for FunctionCurve<T, F> {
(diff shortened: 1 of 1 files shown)


mweatherley:
I'm actually still a little bit unsure about `MapCurve` specifically. My reasoning was that it should behave like it owns a `fn() -> S` since it actually owns a curve with that output type; the `fn(S) -> T` phantom type looks reasonable but is also just going to be redundant in all real cases, because `F: Fn(S) -> T` (so, for instance, it could literally be a function pointer `fn(S) -> T`).

MrGVSV (approved):
My brain struggles so hard with trying to understand variance, but this looks right to me

IQuick143 (approved):

MrGVSV:
> I'm actually still a little bit unsure about `MapCurve` specifically. My reasoning was that it should behave like it owns a `fn() -> S` since it actually owns a curve with that output type; the `fn(S) -> T` phantom type looks reasonable but is also just going to be redundant in all real cases, because `F: Fn(S) -> T` whenever the struct is a curve (so, for instance, it could literally be a function pointer `fn(S) -> T`).

When would `MapCurve` not be a curve? Is there a reason `F` isn't defined with the `Fn(S) -> T` bound?

alice-i-cecile (approved):

PR #15882, opened 2024-10-13
use previous ubuntu version for example validation

mockersf:
# Objective

- Example validation job fails in CI
- This happened after GitHub updated the latest version of ubuntu from the 22.04 to the 24.04
- The package libegl1-mesa is not available on ubuntu 24.04

## Solution

- Keep using ubuntu 22.04
- This is a temp fix and we should fix the update

## Testing

- if it can get merged then it works 🤷


alice-i-cecile (approved):
I love CI.

BD103 (approved):
CI spelled backwards is pronounced "ick".

PR #15883, opened 2024-10-13
Alternative fix for `mesh2d_manual` example

### LINKED ISSUE #15847: `mesh2d_manual` example doesn't render correctly

rparrett:
# Objective

Fixes #15847

Alternative to #15862. Would appreciate a rendering person signaling preference for one or the other.

## Solution

Partially revert the changes made to this example in #15524.

Add comment explaining that the non-usage of the built-in color vertex attribute is intentional.

## Testing

`cargo run --example mesh2d_manual`

### DIFF
--- a/examples/2d/mesh2d_manual.rs
+++ b/examples/2d/mesh2d_manual.rs
@@ -11,7 +11,7 @@ use bevy::{
     math::{ops, FloatOrd},
     prelude::*,
     render::{
-        mesh::{Indices, RenderMesh},
+        mesh::{Indices, MeshVertexAttribute, RenderMesh},
         render_asset::{RenderAssetUsages, RenderAssets},
         render_phase::{
             AddRenderCommand, DrawFunctions, PhaseItemExtraIndex, SetItemPipeline,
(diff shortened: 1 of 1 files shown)


MiniaczQ (approved):
I like this more than the alternative,
can confirm the example works

tychedelia (approved):
I think it makes sense to demonstrate something actually custom. Thanks!

IceSentry:
Why does the current code not work? AFAIK attributes haven't changed at all for a long time?

rparrett:
> Why does the current code not work? AFAIK attributes haven't changed at all for a long time?

@IceSentry It was mistake during the migration in #15862. During the process of figuring out why `mesh2d_manual` wasn't working, it was tidied up to not use the built-in color vertex attribute instead of the custom one. But the shader code wasn't updated. The underlying problem with the example was fixed in another way.

So the question right now is just "do we want to use a custom vertex attribute in `mesh2d_manual` or the built-in one?".

IceSentry (approved):
Okay, thank you for the context, in that case, yeah I'm happy to use a custom attribute here.

Issue #5977, opened 2022-09-13
Allow read-write raw WindowEvents within Bevy apps


alice-i-cecile:
## What problem does this solve or what need does it fill?

While working on an input capture and playback library, @plof27 pointed out that I may be able to capture raw events from `winit` directly, rather than dealing with the separated event streams.

Unfortunately, the necessary `WindowEvent` type is not exposed in [`winit_runner_with`](https://github.com/bevyengine/bevy/blob/1914696a24f33ee3abc18effd1a7b14f4f303b82/crates/bevy_winit/src/lib.rs#L339), our `winit`-based app runner.

## What solution would you like?

1. Allow users to read the full list of raw window events sent.
2. Allow users to send raw window events from within systems, which are then processed by the `winit` loop.

These should be exposed as a simple `Events<WindowEvent>`; the complexity comes from ensuring that `winit_runner_with` responds appropriately.

## What alternative(s) have you considered?

I've considered exposing this with a custom runner, however #4537 means this is a tedious and duplication-heavy task.

alice-i-cecile:
This is blocked on https://github.com/rust-windowing/winit/issues/1387, as `WindowEvent` has a mostly pointless lifetime that means we can't effectively store this data without some serious cludgery. Fix is proposed in ttps://github.com/rust-windowing/winit/pull/1456 / , but the PR is very dead.

alice-i-cecile:
This isn't quite blocked, the `WindowEvent::to_static` method exists, and thus `WindowEvent<'static>` can be used safely. However the problematic variant (scale factor changing) cannot be represented.

JamesHovet:
Just wanted to chime in and give a +1 to this issue. I recently went down the same rabbit hole of wanting to access raw events before the winit runner had the chance to convert them into bevy events and was unable to do so. 

(in my case because I wanted to pass the raw events to egui-winit so that I could use egui without using the bevy renderer and a crate like bevy_egui, but that's beside the point)

I think the interface that would have been most ergonomic for me would be to have been able to pass in an "intercept" function that took in winit window events and done whatever processing I wanted with them and then returned a boolean indicating if that event should be propagated down to the winit-runner or not.

HugoPeters1024:
I would also like to +1 this, especially the read access. My use case is for a custom rendering backend that implements egui support transitively via `egui_winit` (`egui_bevy` is unfortunately tightly coupled with Bevy's default rendering backend). 

[...]

PR #15884, opened 2024-10-13
bevy_winit(emit raw winit events)

HugoPeters1024:
# Objective

- Exposes raw winit events making Bevy even more modular and powerful for custom plugin developers (e.g. a custom render plugin).

XRef: https://github.com/bevyengine/bevy/issues/5977
It doesn't quite close the issue as sending events is not supported (or not very useful to be precise). I would think that supporting that requires some extra considerations by someone a bit more familiar with the `bevy_winit` crate. That said, this PR could be a nice step forward.

## Solution

Emit `RawWinitWindowEvent` objects for each received event.

## Testing

I verified that the events are emitted using a basic test app. I don't think it makes sense to solidify this behavior in one of the examples.

---

## Showcase

My example usage for a custom `egui_winit` integration:

```rust
for ev in winit_events.read() {
    if ev.window_id == window.id {
        let _ = egui_winit.on_window_event(&window, &ev.event);
    }
}
```

### DIFF
--- a/crates/bevy_winit/src/lib.rs
+++ b/crates/bevy_winit/src/lib.rs
@@ -19,7 +19,7 @@ use bevy_reflect::prelude::ReflectDefault;
 use bevy_reflect::Reflect;
 use bevy_window::{RawHandleWrapperHolder, WindowEvent};
 use core::marker::PhantomData;
-use winit::event_loop::EventLoop;
+use winit::{event_loop::EventLoop, window::WindowId};
 
 use bevy_a11y::AccessibilityRequested;
 use bevy_app::{App, Last, Plugin};
@@ -122,6 +122,7 @@ impl<T: Event> Plugin for WinitPlugin<T> {
         app.init_non_send_resource::<WinitWindows>()
             .init_resource::<WinitMonitors>()
             .init_resource::<WinitSettings>()
+            .add_event::<RawWinitWindowEvent>()
             .set_runner(winit_runner::<T>)
             .add_systems(
                 Last,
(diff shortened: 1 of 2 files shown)


crates/bevy_winit/src/lib.rs
```diff
@@ -149,6 +150,21 @@ impl<T: Event> Plugin for WinitPlugin<T> {
 #[derive(Debug, Default, Clone, Copy, Event)]
 pub struct WakeUp;
 
+/// The original window event as produced by winit. This is meant as an escape
+/// hatch for power users that wish add custom winit integrations.
```
BenjaminBrienen, line 154: winit uncapitalized

crates/bevy_winit/src/state.rs
```diff
@@ -80,6 +80,8 @@ struct WinitAppRunnerState<T: Event> {
     previous_lifecycle: AppLifecycle,
     /// Bevy window events to send
     bevy_window_events: Vec<bevy_window::WindowEvent>,
+    /// Raw Winit window events to send
```
BenjaminBrienen, line 83: winit capitalized

BenjaminBrienen (approved):
Only a tiny nit! Just change one of them (whichever doesn't follow the typical usage).

BenjaminBrienen (approved):

BenjaminBrienen (approved):

HugoPeters1024:
I rebased to hopefully get the CI fix for mac os screenshots

BenjaminBrienen (approved):

tychedelia (approved):
Not so sure that just doing a plain copy is the best way, esp since 99% of users won't use this, but I did just run into a situation with egui as well that requires the underlying winit events.

HugoPeters1024:
> Not so sure that just doing a plain copy is the best way, esp since 99% of users won't use this, but I did just run into a situation with egui as well that requires the underlying winit events.

Thanks for the review. I acknowledge that indeed most users would not and probably should not use this. But it can be a powerful escape hatch as many 3rd party packages take raw egui events as input.

tychedelia:
Another alternative would be to provide a way to convert bevy events back into winit events. I'm doing this in my project and it seems to work okay, although is a lossy process w.r.t. some of the original event information.

IceSentry (approved):

crates/bevy_winit/src/lib.rs
```diff
@@ -155,6 +156,21 @@ impl<T: Event> Plugin for WinitPlugin<T> {
 #[reflect(Debug, Default)]
 pub struct WakeUp;
 
+/// The original window event as produced by Winit. This is meant as an escape
+/// hatch for power users that wish add custom Winit integrations.
```
IceSentry, line 160:
```suggestion
/// hatch for power users that wish to add custom Winit integrations.
```

PR #15886, opened 2024-10-13
Meshlet builder improvements redux

JMS55:
Take a bunch more improvements from @zeux's nanite.cpp code

* Use position-only vertices (discard other attributes) to determine meshlet connectivity for grouping
* Rather than using the lock borders flag when simplifying meshlet groups, provide the locked vertices ourselves. The lock borders flag locks the entire border of the meshlet group, but really we only want to lock the edges between meshlet groups - outwards facing edges are fine to unlock. This gives a really significant increase to the DAG quality.
* Add back stuck meshlets (group has only a single meshlet, simplification failed) to the simplification queue to allow them to get used later on and have another attempt at simplifying
* Target 8 meshlets per group instead of 4 (second biggest improvement after manual locks)
* Provide a seed to metis for deterministic meshlet building
* Misc other improvements

Need to merge https://github.com/bevyengine/bevy/pull/15846 first

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1209,7 +1209,7 @@ setup = [
     "curl",
     "-o",
     "assets/models/bunny.meshlet_mesh",
-    "https://raw.githubusercontent.com/JMS55/bevy_meshlet_asset/8443bbdee0bf517e6c297dede7f6a46ab712ee4c/bunny.meshlet_mesh",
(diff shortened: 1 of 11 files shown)


atlv24 (approved):

alice-i-cecile:
Same question as in #15846: why is this vital to get into the 0.15 release, rather than simply "would be nice" :)

cart (approved):
Seems well motivated and the code sure is affecting meshlet building :smile: 

This is scoped to the experimental meshlet module so I'm cool with its inclusion in 0.15.

Just sort out merge conflicts and I'm down to merge.

PR #15887, opened 2024-10-13
Text Rework cleanup

MiniaczQ:
# Objective

Cleanup naming and docs, add missing migration guide after #15591 

All text root nodes now use `Text` (UI) / `Text2d`.
All text readers/writers use `Text<Type>Reader`/`Text<Type>Writer` convention.

---

## Migration Guide

Doubles as #15591 migration guide.

Text bundles (`TextBundle` and `Text2dBundle`) were removed in favor of `Text` and `Text2d`.
Shared configuration fields were replaced with `TextLayout`, `TextFont` and `TextColor` components.
Just `TextBundle`'s additional field turned into `TextNodeFlags` component,
while `Text2dBundle`'s additional fields turned into `TextBounds` and `Anchor` components.

Text sections were removed in favor of hierarchy-based approach.
For root text entities with `Text` or `Text2d` components, child entities with `TextSpan` will act as additional text sections.
To still access text spans by index, use the new `TextUiReader`, `Text2dReader` and `TextUiWriter`, `Text2dWriter` system parameters.

### DIFF
--- a/crates/bevy_ui/src/lib.rs
+++ b/crates/bevy_ui/src/lib.rs
@@ -8,7 +8,7 @@
 
 //! This crate contains Bevy's UI system, which can be used to create UI for both 2D and 3D games
 //! # Basic usage
-//! Spawn UI elements with [`node_bundles::ButtonBundle`], [`node_bundles::ImageBundle`], [`TextUi`](prelude::TextUi) and [`node_bundles::NodeBundle`]
+//! Spawn UI elements with [`node_bundles::ButtonBundle`], [`node_bundles::ImageBundle`], [`Text`](prelude::Text) and [`node_bundles::NodeBundle`]
 //! This UI is laid out with the Flexbox and CSS Grid layout models (see <https://cssreference.io/flexbox/>)
 
 pub mod measurement;
@@ -53,7 +53,7 @@ pub mod prelude {
             node_bundles::*,
             ui_material::*,
             ui_node::*,
-            widget::{Button, Label, TextUi, TextUiReader, TextUiWriter},
+            widget::{Button, Label, Text, TextUiReader, TextUiWriter},
             Interaction, UiMaterialHandle, UiMaterialPlugin, UiScale,
         },
         // `bevy_sprite` re-exports for texture slicing
@@ -224,17 +224,17 @@ impl Plugin for UiPlugin {
 fn build_text_interop(app: &mut App) {
     use crate::widget::TextNodeFlags;
     use bevy_text::TextLayoutInfo;
-    use widget::TextUi;
+    use widget::Text;
 
     app.register_type::<TextLayoutInfo>()
         .register_type::<TextNodeFlags>()
-        .register_type::<TextUi>();
+        .register_type::<Text>();
 
     app.add_systems(
         PostUpdate,
         (
             (
-                bevy_text::detect_text_needs_rerender::<TextUi>,
+                bevy_text::detect_text_needs_rerender::<Text>,
                 widget::measure_text_system,
             )
                 .chain()
--- a/crates/bevy_ui/src/lib.rs
+++ b/crates/bevy_ui/src/lib.rs
@@ -8,7 +8,7 @@
 
 //! This crate contains Bevy's UI system, which can be used to create UI for both 2D and 3D games
 //! # Basic usage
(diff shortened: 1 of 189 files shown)


MiniaczQ:
I'm also thinking whether this should reintroduce deprecated `Text2dBundle` and `TextBundle` so people get a deprecation message with migration instructions
@alice-i-cecile

crates/bevy_ui/src/widget/text.rs
```diff
@@ -104,18 +104,18 @@ impl Default for TextNodeFlags {
     Visibility, // TODO: Remove when Node uses required components.
     Transform // TODO: Remove when Node uses required components.
 )]
-pub struct Text(pub String);
+pub struct TextUi(pub String);
```
ickshonpe, line 107: It's good to have consistant naming scheme, but I find this a bit ugly. I'd prefer if we had `Text` refer to UI text and just  add the -`2d` suffix to the text2d types.

MiniaczQ: It's like the core controversy in this PR
I think it's clearer if we have 2d/Ui
also helps with the TextReader situation, where we need to add "Ui" to prevent collision with the generic type

alice-i-cecile: I have a strong preference for just Text, and I think Cart will share that.

alice-i-cecile:
I really like most of these changes :) If we revert the TextUi change IMO it's uncontroversial.

alice-i-cecile:
> I'm also thinking whether this should reintroduce deprecated `Text2dBundle` and `TextBundle` so people get a deprecation message with migration instructions @alice-i-cecile

Yes please!

crates/bevy_ui/src/lib.rs
```diff
@@ -46,14 +46,17 @@ use widget::UiImageSize;
...
+    pub use crate::widget::TextBundle;
     #[doc(hidden)]
     pub use {
         crate::{
             geometry::*,
             node_bundles::*,
             ui_material::*,
             ui_node::*,
-            widget::{Button, Label, Text, UiTextReader, UiTextWriter},
+            widget::{Button, Label, Text, TextUiReader, TextUiWriter},
```
ickshonpe, line 59: The last three should be `"bevy_text"` feature-gated?

ickshonpe (changes requested):

crates/bevy_ui/src/lib.rs
MiniaczQ: Hm, `bevy_ui` doesn't have a `bevy_text` feature flag, guess I'll need to add that

MiniaczQ: `bevy_ui` feature always imports `bevy_text`
It's a bit confusing that `bevy_text` is marked as an optional dependency, so I'll remove that.

MiniaczQ: Hm, that didn't work 🤔 
Do dependencies act as features if marked as optional?
I thought we need to add a feature flag that then does `dep:dependency_crate`

ickshonpe: I'm not sure I don't understand rust feature gating that well. `bevy_ui` is meant to be able to run exclusively `bevy_text` but that doesn't seem possible atm.

alice-i-cecile: > Do dependencies act as features if marked as optional?

Yes.

BenjaminBrienen (approved):

alice-i-cecile (approved):

ickshonpe (approved):
Seem's fine. I'd prefer if `TextUiReader/Writer` was just called `TextReader/Writer` but then we'd need to rename the underlying type. It's not worth fussing over for now.

UkoeHB (approved):

alice-i-cecile:
@Aravind-blip, commenting on PRs that were merged two years ago is unhelpful. Moreover, undisclosed AI use is frustrating and harms trust. Please do not do this again, or you will be banned from this repo.

Aravind-blip:
@alice-i-cecile  Thanks for pointing that out. I was completing an external review exercise and didn’t realize commenting on an old merged PR would cause disruption. That wasn’t my intention. I won’t do that again.

PR #14428, opened 2024-07-21
Remove incorrect equality comparisons for asset load error types

JoJoJet:
# Objective

The type `AssetLoadError` has `PartialEq` and `Eq` impls, which is problematic due to the fact that the `AssetLoaderError` and `AddAsyncError` variants lie in their impls: they will return `true` for any `Box<dyn Error>` with the same `TypeId`, even if the actual value is different. This can lead to subtle bugs if a user relies on the equality comparison to ensure that two values are equal.

More generally, it is an anti-pattern for large error types involving dynamic dispatch, such as `AssetLoadError`, to have equality comparisons. Directly comparing two errors for equality is usually not desired -- if some logic needs to branch based on the value of an error, it is usually more correct to check for specific variants and inspect their fields.

As far as I can tell, the only reason these errors have equality comparisons is because the `LoadState` enum wraps `AssetLoadError` for its `Failed` variant. This equality comparison is only used to check for `== LoadState::Loaded`, which we can easily replace with an `is_loaded` method.

## Solution

Remove the `{Partial}Eq` impls from `LoadState`, which also allows us to remove it from the error types.

## Migration Guide

The types `bevy_asset::AssetLoadError` and `bevy_asset::LoadState` no longer support equality comparisons. If you need to check for an asset's load state, consider checking for a specific variant using `LoadState::is_loaded` or the `matches!` macro. Similarly, consider using the `matches!` macro to check for specific variants of the `AssetLoadError` type if you need to inspect the value of an asset load error in your code.

Issue #13328, opened 2024-05-11
Can't compile bevy_dylib


Fidius-jko:
## Bevy version

0.13
## Relevant system information/What you did
![image](https://github.com/bevyengine/bevy/assets/95738553/63eb3bb5-6d84-4ef8-9521-37492a4698dc)

If i compile bevy i see this error

## What went wrong

bevy_dylib doesn't work

SkiFire13:
Did you follow the instructions from https://bevyengine.org/learn/quick-start/getting-started/setup/#enable-fast-compiles-optional? In particular this part:

> *Important!* On Windows you _must_ also enable the [performance optimizations](https://bevyengine.org/learn/quick-start/getting-started/setup/#compile-with-performance-optimizations) or you will get a [`too many exported symbols`](https://github.com/bevyengine/bevy/issues/1110#issuecomment-1312926923) error.

Fidius-jko:
Thanks, this note was not there before.

PR #15890, opened 2024-10-13
Remove incorrect equality comparisons for asset load error types

BenjaminBrienen:
# Objective

The type `AssetLoadError` has `PartialEq` and `Eq` impls, which is problematic due to the fact that the `AssetLoaderError` and `AddAsyncError` variants lie in their impls: they will return `true` for any `Box<dyn Error>` with the same `TypeId`, even if the actual value is different. This can lead to subtle bugs if a user relies on the equality comparison to ensure that two values are equal.

The same is true for `DependencyLoadState`, `RecursiveDependencyLoadState`.

More generally, it is an anti-pattern for large error types involving dynamic dispatch, such as `AssetLoadError`, to have equality comparisons. Directly comparing two errors for equality is usually not desired -- if some logic needs to branch based on the value of an error, it is usually more correct to check for specific variants and inspect their fields.

As far as I can tell, the only reason these errors have equality comparisons is because the `LoadState` enum wraps `AssetLoadError` for its `Failed` variant. This equality comparison is only used to check for `== LoadState::Loaded`, which we can easily replace with an `is_loaded` method.

## Solution

Remove the `{Partial}Eq` impls from `LoadState`, which also allows us to remove it from the error types.

Adopted from #14428 by @JoJoJet.

## Migration Guide

The types `bevy_asset::AssetLoadError` and `bevy_asset::LoadState` no longer support equality comparisons. If you need to check for an asset's load state, consider checking for a specific variant using `LoadState::is_loaded` or the `matches!` macro. Similarly, consider using the `matches!` macro to check for specific variants of the `AssetLoadError` type if you need to inspect the value of an asset load error in your code.

`DependencyLoadState` and `RecursiveDependencyLoadState` are not released yet, so no migration needed,


mockersf:
Could you either open your PR against @JoJoJet branch so that they can merge it, or copy the original title/description in this PR? 
If merged, this would add a commit "continuation of #13328" in the git log of Bevy which wouldn't be helpful.

Same for sections of the description, it's better for a PR to be self contained rather than have to refer to another. Some are also used by automated tools which don't know how to follow redirections

BenjaminBrienen:
> Could you either open your PR against @JoJoJet branch so that they can merge it, or copy the original title/description in this PR? 
> If merged, this would add a commit "continuation of #14428" in the git log of Bevy which wouldn't be helpful.
> 
> Same for sections of the description, it's better for a PR to be self contained rather than have to refer to another. Some are also used by automated tools which don't know how to follow redirections

Done!

crates/bevy_asset/src/server/mod.rs
```diff
@@ -1500,8 +1496,35 @@ pub enum LoadState {
     Failed(Arc<AssetLoadError>),
 }
 
+impl LoadState {
+    /// Returns `true` if this instance is [`LoadState::Loading`]
+    pub fn is_loading(&self) -> bool {
+        matches!(self, Self::Loading)
```
BenjaminBrienen, line 1502: I could prefer this over the matches in the tests

crates/bevy_asset/src/server/mod.rs
```diff
@@ -1500,8 +1496,35 @@ pub enum LoadState {
...
+    }
+    // NOTE: an `is_not_loaded` method is intentionally not included, as it may mislead some users
+    // into thinking it is complementary to `is_loaded`.
+    // `NotLoaded` is a very specific failure mode and in most cases it is not necessary to directly check for it.
+    // If this is necessary the `matches!` macro can be used instead of a helper method.
+
+    /// Returns `true` if the asset is loaded or in the process of being loaded. If true true,
+    /// then the asset can be considered to be in a "normal" state: the asset either exists
+    /// or will exist, and no errors have been encountered yet.
+    pub fn is_loaded_or_loading(&self) -> bool {
```
BenjaminBrienen, line 1516: Seems a bit extra. Is this a common use case?

BenjaminBrienen:
Please feel free to leave a reply on my review comments and then I'll feel comfortable refining the PR with that feedback.

crates/bevy_asset/src/lib.rs
```diff
@@ -926,25 +924,25 @@ mod tests {
             let (a_load, a_deps, a_rec_deps) = asset_server.get_load_states(a_id).unwrap();
             assert_eq!(a_text.text, "a");
             assert_eq!(a_text.dependencies.len(), 2);
-            assert_eq!(a_load, LoadState::Loaded, "a is loaded");
-            assert_eq!(a_deps, DependencyLoadState::Loading);
-            assert_eq!(a_rec_deps, RecursiveDependencyLoadState::Loading);
+            assert!(a_load.is_loaded(), "a is loaded");
```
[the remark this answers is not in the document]
alice-i-cecile: Yes.

crates/bevy_asset/src/server/mod.rs
alice-i-cecile: Yes.

crates/bevy_asset/src/lib.rs
```diff
@@ -926,25 +924,25 @@ mod tests {
...
+            assert!(matches!(a_rec_deps, RecursiveDependencyLoadState::Loading));
 
             let b_id = a_text.dependencies[0].id();
             let b_text = get::<CoolText>(world, b_id);
             let (b_load, b_deps, b_rec_deps) = asset_server.get_load_states(b_id).unwrap();
             assert!(b_text.is_none(), "b component should not exist yet");
-            assert_eq!(b_load, LoadState::Loading);
-            assert_eq!(b_deps, DependencyLoadState::Loading);
-            assert_eq!(b_rec_deps, RecursiveDependencyLoadState::Loading);
+            assert!(b_load.is_loading());
```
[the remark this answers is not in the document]
alice-i-cecile: Not particularly. I mean, might as well fix this but...

crates/bevy_asset/src/server/mod.rs
alice-i-cecile: IMO cut this: it's extraneous.

BenjaminBrienen:
@alice-i-cecile feedback is addressed

I also realized that the merge conflict involved 2 new error types, which should have the same process done to them. It is much better now, I think. I'll update the migration guide.

crates/bevy_asset/src/lib.rs
```diff
@@ -905,17 +904,9 @@ mod tests {
...
-                DependencyLoadState::Loading,
-                "a deps should still be loading"
-            );
-            assert_eq!(
-                a_rec_deps,
-                RecursiveDependencyLoadState::Loading,
-                "a recursive deps should still be loading"
-            );
+            assert!(a_load.is_loading());
+            assert!(a_deps.is_loaded());
```
andriyDev, line 908: There are a lot of these in the diff where a_deps should be `is_loading`, but is instead `is_loaded`. Is this correct?

crates/bevy_asset/src/lib.rs
```diff
@@ -926,25 +917,25 @@ mod tests {
             let (a_load, a_deps, a_rec_deps) = asset_server.get_load_states(a_id).unwrap();
             assert_eq!(a_text.text, "a");
             assert_eq!(a_text.dependencies.len(), 2);
-            assert_eq!(a_load, LoadState::Loaded, "a is loaded");
-            assert_eq!(a_deps, DependencyLoadState::Loading);
-            assert_eq!(a_rec_deps, RecursiveDependencyLoadState::Loading);
+            assert!(a_load.is_loaded());
+            assert!(a_deps.is_loaded());
```
andriyDev, line 921: is_loading

crates/bevy_asset/src/lib.rs
```diff
@@ -926,25 +917,25 @@ mod tests {
...
 
             let b_id = a_text.dependencies[0].id();
             let b_text = get::<CoolText>(world, b_id);
             let (b_load, b_deps, b_rec_deps) = asset_server.get_load_states(b_id).unwrap();
             assert!(b_text.is_none(), "b component should not exist yet");
-            assert_eq!(b_load, LoadState::Loading);
-            assert_eq!(b_deps, DependencyLoadState::Loading);
-            assert_eq!(b_rec_deps, RecursiveDependencyLoadState::Loading);
+            assert!(b_load.is_loading());
+            assert!(b_deps.is_loaded());
```
andriyDev, line 929: is_loading

crates/bevy_asset/src/lib.rs
```diff
@@ -926,25 +917,25 @@ mod tests {
...
 
             let c_id = a_text.dependencies[1].id();
             let c_text = get::<CoolText>(world, c_id);
             let (c_load, c_deps, c_rec_deps) = asset_server.get_load_states(c_id).unwrap();
             assert!(c_text.is_none(), "c component should not exist yet");
-            assert_eq!(c_load, LoadState::Loading);
-            assert_eq!(c_deps, DependencyLoadState::Loading);
-            assert_eq!(c_rec_deps, RecursiveDependencyLoadState::Loading);
+            assert!(c_load.is_loading());
+            assert!(c_deps.is_loaded());
```
andriyDev, line 937: is_loading

crates/bevy_asset/src/lib.rs
```diff
@@ -956,25 +947,25 @@ mod tests {
             let (a_load, a_deps, a_rec_deps) = asset_server.get_load_states(a_id).unwrap();
             assert_eq!(a_text.text, "a");
             assert_eq!(a_text.dependencies.len(), 2);
-            assert_eq!(a_load, LoadState::Loaded);
-            assert_eq!(a_deps, DependencyLoadState::Loading);
-            assert_eq!(a_rec_deps, RecursiveDependencyLoadState::Loading);
+            assert!(a_load.is_loaded());
+            assert!(a_deps.is_loaded());
```
andriyDev, line 951: is_loading

crates/bevy_asset/src/lib.rs
```diff
@@ -956,25 +947,25 @@ mod tests {
...
 
             let c_id = a_text.dependencies[1].id();
             let c_text = get::<CoolText>(world, c_id);
             let (c_load, c_deps, c_rec_deps) = asset_server.get_load_states(c_id).unwrap();
             assert!(c_text.is_none(), "c component should not exist yet");
-            assert_eq!(c_load, LoadState::Loading);
-            assert_eq!(c_deps, DependencyLoadState::Loading);
-            assert_eq!(c_rec_deps, RecursiveDependencyLoadState::Loading);
+            assert!(c_load.is_loading());
+            assert!(c_deps.is_loaded());
```
andriyDev, line 967: is_loading

crates/bevy_asset/src/lib.rs
```diff
@@ -991,31 +982,29 @@ mod tests {
...
             let (c_load, c_deps, c_rec_deps) = asset_server.get_load_states(c_id).unwrap();
             assert_eq!(c_text.text, "c");
             assert_eq!(c_text.embedded, "ab");
-            assert_eq!(c_load, LoadState::Loaded);
-            assert_eq!(
-                c_deps,
-                DependencyLoadState::Loading,
+            assert!(c_load.is_loaded());
+            assert!(
+                c_deps.is_loaded(),
```
andriyDev, line 1003: is_loading

crates/bevy_asset/src/lib.rs
```diff
@@ -1025,26 +1014,24 @@ mod tests {
...
 
             let d_id = c_text.dependencies[0].id();
             let d_text = get::<CoolText>(world, d_id);
             let (d_load, d_deps, d_rec_deps) = asset_server.get_load_states(d_id).unwrap();
             assert!(d_text.is_none(), "d component should not exist yet");
-            assert_eq!(d_load, LoadState::Loading);
-            assert_eq!(d_deps, DependencyLoadState::Loading);
-            assert_eq!(d_rec_deps, RecursiveDependencyLoadState::Loading);
+            assert!(d_load.is_loading());
+            assert!(d_deps.is_loaded());
```
andriyDev, line 1026: is_loading

crates/bevy_asset/src/lib.rs
```diff
@@ -1349,9 +1323,9 @@ mod tests {
         run_app_until(&mut app, |world| {
             let _a_text = get::<CoolText>(world, a_id)?;
             let (a_load, a_deps, a_rec_deps) = asset_server.get_load_states(a_id).unwrap();
-            assert_eq!(a_load, LoadState::Loaded);
-            assert_eq!(a_deps, DependencyLoadState::Loading);
-            assert_eq!(a_rec_deps, RecursiveDependencyLoadState::Loading);
+            assert!(a_load.is_loaded());
+            assert!(a_deps.is_loaded());
```
andriyDev, line 1327: is_loading

andriyDev (approved):

BenjaminBrienen:
@andriyDev I've gone through and double checked all the assertions. Sorry for the copy paste errors!

crates/bevy_asset/src/lib.rs
```diff
@@ -1025,26 +1014,24 @@ mod tests {
...
             let d_id = c_text.dependencies[0].id();
             let d_text = get::<CoolText>(world, d_id);
             let (d_load, d_deps, d_rec_deps) = asset_server.get_load_states(d_id).unwrap();
             assert!(d_text.is_none(), "d component should not exist yet");
-            assert_eq!(d_load, LoadState::Loading);
-            assert_eq!(d_deps, DependencyLoadState::Loading);
-            assert_eq!(d_rec_deps, RecursiveDependencyLoadState::Loading);
+            assert!(d_load.is_loading());
+            assert!(d_deps.is_loading());
+            assert!(d_rec_deps.is_loaded());
```
andriyDev, line 1027: This line seems to have been changed accidentally as well.

andriyDev (approved):

crates/bevy_asset/src/lib.rs
BenjaminBrienen: Thanks! By the way, you can review it with the status "Request Changes" instead of approving.

andriyDev (approved):

alice-i-cecile (approved):

PR #15898, opened 2024-10-14
Migrate UI bundles to required components

### LINKED ISSUE #15889: Migrate UI bundles to required components
Almost all of our bundle types have been migrated to required components after #14791. However, UI bundles have not been done yet.

#15550 will conflict with that, but since that PR needs more time to bake, we should press on with this work for the 0.15 release.

The strategy to do so is laid out here: https://hackmd.io/@bevy/required_components/%2FpuAA8c18TzeMhjclo36vEQ#Combined-Proposal-1-Conservative-Port-no-major-reworks-Selected

VitalyAnkh:
# Objective

- Migrate UI bundles to required components, fixes #15889

## Solution

- deprecate `NodeBundle` in favor of `Node`
- deprecate `ImageBundle` in favor of `UiImage`
- deprecate `ButtonBundle` in favor of `Button`

## Testing

CI.

## Migration Guide

- Replace all uses of `NodeBundle` with `Node`. e.g.
```diff
     commands
-        .spawn(NodeBundle {
-            style: Style {
+        .spawn((
+            Node::default(),
+            Style {
                 width: Val::Percent(100.),
                 align_items: AlignItems::Center,
                 justify_content: JustifyContent::Center,
                 ..default()
             },
-            ..default()
-        })
+        ))
``` 
- Replace all uses of `ButtonBundle` with `Button`. e.g.
```diff
                     .spawn((
-                        ButtonBundle {
-                            style: Style {
-                                width: Val::Px(w),
-                                height: Val::Px(h),
-                                // horizontally center child text
-                                justify_content: JustifyContent::Center,
-                                // vertically center child text
-                                align_items: AlignItems::Center,
-                                margin: UiRect::all(Val::Px(20.0)),
-                                ..default()
-                            },
-                            image: image.clone().into(),
+                        Button,
+                        Style {
+                            width: Val::Px(w),
+                            height: Val::Px(h),
+                            // horizontally center child text
+                            justify_content: JustifyContent::Center,
+                            // vertically center child text
+                            align_items: AlignItems::Center,
+                            margin: UiRect::all(Val::Px(20.0)),
                             ..default()
                         },
+                        UiImage::from(image.clone()),
                         ImageScaleMode::Sliced(slicer.clone()),
                     ))
```
- Replace all uses of `ImageBundle` with `UiImage`. e.g.
```diff
-    commands.spawn(ImageBundle {
-        image: UiImage {
+    commands.spawn((
+        UiImage {
             texture: metering_mask,
             ..default()
         },
-        style: Style {
+        Style {
             width: Val::Percent(100.0),
             height: Val::Percent(100.0),
             ..default()
         },
-        ..default()
-    });
+    ));
 ```

### CODE REVIEW

File: crates/bevy_ui/src/widget/button.rs [hunk 1]
```diff
@@ -1,7 +1,13 @@
...
 use bevy_ecs::{prelude::Component, reflect::ReflectComponent};
 use bevy_reflect::{std_traits::ReflectDefault, Reflect};
 
 /// Marker struct for buttons
 #[derive(Component, Debug, Default, Clone, Copy, PartialEq, Eq, Reflect)]
 #[reflect(Component, Default, Debug, PartialEq)]
+#[require(Node, FocusPolicy(block_focus_policy), Interaction)]
 pub struct Button;
+
+fn block_focus_policy() -> FocusPolicy {
```
cart, line 11: This should be defined inline as

```rust
#[require(FocusPolicy(|| FocusPolicy::Block))]
```

examples/ui/ui_texture_slice_flip_and_tile.rs
```diff
@@ -60,13 +60,11 @@ fn setup(mut commands: Commands, asset_server: Res<AssetServer>) {
...
             ] {
                 parent.spawn((
-                    NodeBundle {
-                        style: Style {
-                            width: Val::Px(width),
-                            height: Val::Px(height),
-                            ..default()
-                        },
-                        ..Default::default()
+                    Node::default(),
```
cart, line 63: This can be removed as UiImage requires Node. Lets move UiImage to the top of this declaration.

VitalyAnkh: Thanks! It's done.

crates/bevy_ui/src/widget/button.rs
VitalyAnkh: Thank! I didn't realize I could use closure. Fixed.

Should use [In-line Required Component Values](https://hackmd.io/@bevy/required_components/%2FHwaPiH3GRWqxrEkCKRzXYQ) when it's completed.

crates/bevy_ui/src/ui_node.rs
```diff
@@ -25,6 +27,18 @@ use smallvec::SmallVec;
 /// - [`Interaction`](crate::Interaction) to obtain the interaction state of this node
 #[derive(Component, Debug, Copy, Clone, PartialEq, Reflect)]
 #[reflect(Component, Default, Debug)]
+#[require(
+    Style,
+    BackgroundColor,
+    BorderColor,
```
ickshonpe, line 33: I'm not sure that we should require `BackgroundColor` and `BorderColor` for `Node`, it means that we can't distinguish between visable and non-visable nodes intended only to constrain the layout. 

Also extraction for `UiMaterial` nodes has a `Without<BackgroundColor>` query filter, so UI materials is broken atm with this PR. That can be fixed by just removing the query filter from `bevy_ui::render::ui_material_pipeline::extract_ui_material_nodes`  but maybe it goes against the intent of the API, I'm not sure.

ickshonpe (changes requested):

ickshonpe (changes requested):

crates/bevy_ui/src/ui_node.rs
alice-i-cecile: Yeah IMO we should remove this require.

alice-i-cecile:
> Doesn't gets drawn because of the Without<BackgroundColor> query filter in extract_ui_materials.

This means we can't have `BackgroundColor` in the required components for either `Node` or `Style`. We could swap this to store an `Option` and then branch on it in this system. It's a bit wasteful to store it but not a big deal.

The other open question here is the `Node` -> `Style` vs `Style` -> `Node` question. I think I have a mild preference for the latter, since it makes the migration nicer for both now and 0.16, when we're planning to move many of the fields from `Style` into `Node`. The lack of `Node::default` is also a big win for me.

On the other hand though, `commands.spawn(Style {..})` is *super* unclear to beginners and has no obvious semantic meaning. So maybe this way is better. @cart, final call please?

ickshonpe:
Neither `Node` nor `Style` has very helpful semantics, maybe we could do a simple renaming for whatever the core component is for now like to `UiNode` or `UiStyle` so it's clear to new users that this is a UI component whilst retaining enough familiarity that exisiting users won't be too confused. I find the `Ui` prefix a bit ugly though.

ickshonpe:
Yet another option. `Node` is mostly only used internally. We could rename the existing `Node` node type to `ResolvedLayout` or `ComputedNode` or whatever without causing too much friction. Then replace it with a zst named `Node` carrying the required components that doesn't need to be constructed using `default()`.

crates/bevy_ui/src/ui_node.rs
ickshonpe: `BackgroundColor` and `BorderColor` are only accessed internally during extraction, so leaving them out shouldn't cause any problems. I ran some naive benchmarks and the cost of extracting all these extra uinodes can become significant if you have a very complicated layout.

alice-i-cecile:
Moving to `ResolvedLayout` or something seems reasonable: that seems like the broad shape of the design we want in the future.

crates/bevy_ui/src/ui_node.rs
cart: I think we should be conservative here given how close we are to release. Lets give this conversation its own space / issue where we can look at the benchmarks and discuss the implications of not having these everywhere.

cart:
> The other open question here is the Node -> Style vs Style -> Node question. I think I have a mild preference for the latter, since it makes the migration nicer for both now and 0.16, when we're planning to move many of the fields from Style into Node. The lack of Node::default is also a big win for me.

This is conceptually wrong. "Node" is the driving context. "Style" is a property of a "Node". The cost of training people to think otherwise is too high. The current port is already an ergonomic win. We can claim the remaining wins when we break up Style.

> I find the Ui prefix a bit ugly though.

Me too. I prefer the short names. Instead of prefixing everything we should just pick concept names that are "unique enough", with a bias toward short and snappy to please the "ui ergonomics" crowd. I personally think `Node` is fine. We can/should discuss this later, not here.

> Moving to ResolvedLayout or something seems reasonable: that seems like the broad shape of the design we want in the future.

Yeah this seems reasonable. But we should discuss this elsewhere.

> Then replace it with a ZST named Node carrying the required components that doesn't need to be constructed using default().

I still strongly believe Node should contain the "style properties that all nodes have" such as width, height, etc. But thats a post 0.15 conversation.

examples/ui/ui_material.rs
```diff
@@ -23,16 +23,16 @@ fn setup(
...
                 align_items: AlignItems::Center,
                 justify_content: JustifyContent::Center,
                 ..default()
             },
-            ..default()
-        })
+        ))
         .with_children(|parent| {
             let banner_scale_factor = 0.5;
             parent.spawn(MaterialNodeBundle {
```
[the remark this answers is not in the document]
cart: Is there a reason we need BackgroundColor to break UI material extraction? Can we just remove that filter? Seems to work for me.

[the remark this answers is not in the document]
cart: Ah it does seem like it would still be captured by `extract_uinode_background_colors`. Munching on this now.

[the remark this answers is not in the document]
cart: This still works by default because BackgroundColor defaults to transparent, which results in material node backgrounds being skipped by default.

This does slightly break if set the background color to a visible color (because the material and the background z-fight with each other). I think ideally we make background color play nicely with this case.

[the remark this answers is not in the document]
cart: I've fixed this in a satisfying way that also paves the way for other such "multi-layered extraction" cases. I've also ported MaterialNodeBundle to required components and renamed `UiMaterialHandle<M>` to `MaterialNode<M>`. It is now the "driving concept" and requires `Node`.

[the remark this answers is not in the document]
cart: Polishing it up now. I'll push momentarily.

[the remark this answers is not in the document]
cart: Just pushed this.

cart (approved):
I just pushed a change that:

1. Added support for BackgroundColor to UI material nodes,  allowing `require(BackgroundColor)` to work as expected.
2. Renames `UiMaterialHandle` to `MaterialNode`
3. Ports UI materials to Required Components. `MaterialNode` is now requires `Node` (because it "is a" Node).
4. Deprecate `MaterialNodeBundle`

I think this PR is now ready to merge

alice-i-cecile (approved):

crates/bevy_ui/src/render/mod.rs
```diff
@@ -823,7 +842,7 @@ pub fn queue_uinodes(
             pipeline,
             entity: (*entity, extracted_uinode.main_entity),
             sort_key: (
-                FloatOrd(extracted_uinode.stack_index as f32),
+                FloatOrd(extracted_uinode.stack_index as f32 + stack_z_offsets::BACKGROUND_COLOR),
```
ickshonpe, line 845: Could just set `extract_ui_material_nodes` after `extract_ui_background_colors` here.
A stable sort is used so that if two phase items have the same stack index then the z-order is determined by the ordering of the extraction system.

ickshonpe (changes requested):

crates/bevy_ui/src/render/mod.rs
cart: Gotcha. I don't have strong opinions one way or the other. I doubt a single per-entity float add will meaningfully tip the performance scales. Doing this per-entity gives us more granularity than "per thing encapsulated by a system", but we also currently have no use case where this matters. I also like that it puts the sort behavior front-and-center in a centralized place. The current behavior is pretty implicit.

ickshonpe: > Could just set `extract_ui_material_nodes` after `extract_ui_background_colors` instead. TransparentUi uses a stable sort, so if two phase items have the same stack index then the z-order is determined by the ordering of the extraction systems.

Oh wait this was changed.

ickshonpe (approved):

crates/bevy_ui/src/render/mod.rs
ickshonpe: I don't care about the performance impact but I don't like the inconsistancy very much and it breaks any existing code that rendered borders or text above ui materials on the same node.

cart: Perhaps, but it also buys flexibility for future scenarios and decouples sort order from system execution order (which feels like a good thing as execution order could easily need to be different than sort order) .

I still don't have super strong opinions here. I think I prefer the explicitness of defining offsets. But I wouldn't strongly object to a change that moves this to using system execution order.

PR #15902, opened 2024-10-14
Despawn unused light-view entity

### LINKED ISSUE #15897: `scene_viewer` example prints warnings after toggling shadow on and off
## Bevy version

main

This does not reproduce with v0.14.2.

## What you did

- Run `scene_viewer` example and toggle shadow on and off (press <kbd>U</kbd>)

## What went wrong

No visual problem, but this warning appears repeatedly in terminal:
`WARN bevy_pbr::render::gpu_preprocess: The preprocessing index buffer wasn't present`

## Additional information

https://github.com/bevyengine/bevy/blob/82d6f56cfc3432f584632f761c44ec8efd2e7c67/crates/bevy_pbr/src/render/light.rs#L1198-L1206

Hitting these conditions cause living (due to now retained render world) entitiy to be removed from `live_shadow_mapping_lights` in this function, hence removed from `ViewBinnedRenderPhases`, and later disagrees with query result in `GpuPreprocessNode::run`.

akimakinai:
# Objective

- Fixes #15897

## Solution

- Despawn light view entities when they go unused or when the corresponding view is not alive.

## Testing

- `scene_viewer` example no longer prints "The preprocessing index buffer wasn't present" warning
- modified an example to try toggling shadows for all kinds of light: https://gist.github.com/akimakinai/ddb0357191f5052b654370699d2314cf

### DIFF
--- a/crates/bevy_pbr/src/render/light.rs
+++ b/crates/bevy_pbr/src/render/light.rs
@@ -1260,6 +1260,11 @@ pub fn prepare_lights(
                     .map(|_| commands.spawn_empty().id())
                     .collect()
             });
+            if light_view_entities.len() != iter.len() {
+                let entities = core::mem::take(light_view_entities);
+                despawn_entities(&mut commands, entities);
+                light_view_entities.extend((0..iter.len()).map(|_| commands.spawn_empty().id()));
+            }
 
             for (cascade_index, (((cascade, frustum), bound), view_light_entity)) in
                 iter.zip(light_view_entities.iter().copied()).enumerate()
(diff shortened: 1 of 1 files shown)


crates/bevy_pbr/src/render/light.rs
```diff
@@ -459,7 +459,7 @@ fn create_render_visible_mesh_entities(
 }
 
 #[derive(Component, Default, Deref, DerefMut)]
-pub struct LightViewEntities(Vec<Entity>);
+pub struct LightViewEntities(EntityHashMap<Vec<Entity>>);
```
alice-i-cecile, line 462: Non-blocking: can we add basic doc strings to this type while we're here?

BenjaminBrienen (approved):

alice-i-cecile (approved):

alice-i-cecile:
@akimakinai can you test this on WASM and investigate the failure please?

PR #15905, opened 2024-10-14
Add module-level docs for Curve

mweatherley:
# Objective

Improve the average user's ability to understand what the heck is going on with the Curve API.

## Solution

I wrote some docs. I doubt these are perfect; I'm probably far too close to this for that to be the case. :)


torsteingrindvik (approved):
This was a really great read. I've been wondering what's been going on with the curve APIs the last months and I feel I have some insight now and it's looking superb.

No real nits or anything (also TIL _parameterized_ can be spelled _parametrized_).

alice-i-cecile (approved):

MiniaczQ (approved):
The intro felt about the right level for me and if anything, it serves as a reference to what this module offers.
Then the examples after that look approachable for any level and cover most interesting use cases. :)
Great job!

PR #15907, opened 2024-10-14
Fix `clippy::needless_lifetimes` for most crates

clarfonthey:
Part "2" for #15906. This fixes `clippy::needless_lifetimes` for all crates except `bevy_reflect` and `bevy_ecs`, to reduce the amount of code to review.

I say "2" in quotes because all the lint fixes can be merged separately, and don't depend on each other. Although all of them are required to truly make `ci lints` pass without errors.

### DIFF
--- a/crates/bevy_audio/src/audio_output.rs
+++ b/crates/bevy_audio/src/audio_output.rs
@@ -58,7 +58,7 @@ pub struct PlaybackRemoveMarker;
 pub(crate) struct EarPositions<'w, 's> {
     pub(crate) query: Query<'w, 's, (Entity, &'static GlobalTransform, &'static SpatialListener)>,
 }
-impl<'w, 's> EarPositions<'w, 's> {
+impl EarPositions<'_, '_> {
     /// Gets a set of transformed ear positions.
     ///
     /// If there are no listeners, use the default values. If a user has added multiple
--- a/crates/bevy_hierarchy/src/hierarchy.rs
+++ b/crates/bevy_hierarchy/src/hierarchy.rs
@@ -176,7 +176,7 @@ fn despawn_descendants_inner<'v, 'w>(
     world
 }
 
-impl<'w> DespawnRecursiveExt for EntityWorldMut<'w> {
+impl DespawnRecursiveExt for EntityWorldMut<'_> {
     /// Despawns the provided entity and its children.
     /// This will emit warnings for any entity that does not exist.
     fn despawn_recursive(self) {
--- a/crates/bevy_hierarchy/src/query_extension.rs
+++ b/crates/bevy_hierarchy/src/query_extension.rs
@@ -214,7 +214,7 @@ where
     }
 }
 
-impl<'w, 's, D: QueryData, F: QueryFilter> Iterator for DescendantIter<'w, 's, D, F>
+impl<'w, D: QueryData, F: QueryFilter> Iterator for DescendantIter<'w, '_, D, F>
 where
     D::ReadOnly: WorldQuery<Item<'w> = &'w Children>,
 {
@@ -259,7 +259,7 @@ where
     }
 }
 
-impl<'w, 's, D: QueryData, F: QueryFilter> Iterator for DescendantDepthFirstIter<'w, 's, D, F>
+impl<'w, D: QueryData, F: QueryFilter> Iterator for DescendantDepthFirstIter<'w, '_, D, F>
 where
     D::ReadOnly: WorldQuery<Item<'w> = &'w Children>,
 {
@@ -298,7 +298,7 @@ where
     }
 }
 
-impl<'w, 's, D: QueryData, F: QueryFilter> Iterator for AncestorIter<'w, 's, D, F>
+impl<'w, D: QueryData, F: QueryFilter> Iterator for AncestorIter<'w, '_, D, F>
 where
     D::ReadOnly: WorldQuery<Item<'w> = &'w Parent>,
 {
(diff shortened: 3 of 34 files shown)


crates/bevy_audio/src/audio_output.rs
```diff
@@ -58,7 +58,7 @@ pub struct PlaybackRemoveMarker;
 pub(crate) struct EarPositions<'w, 's> {
     pub(crate) query: Query<'w, 's, (Entity, &'static GlobalTransform, &'static SpatialListener)>,
 }
-impl<'w, 's> EarPositions<'w, 's> {
+impl EarPositions<'_, '_> {
```
alice-i-cecile, line 61: This is substantially less clear.

crates/bevy_hierarchy/src/hierarchy.rs
```diff
@@ -176,7 +176,7 @@ fn despawn_descendants_inner<'v, 'w>(
     world
 }
 
-impl<'w> DespawnRecursiveExt for EntityWorldMut<'w> {
+impl DespawnRecursiveExt for EntityWorldMut<'_> {
```
alice-i-cecile, line 179: This is much less clear. The 'w lifetime has semantic meaning here.

crates/bevy_hierarchy/src/query_extension.rs
```diff
@@ -214,7 +214,7 @@ where
     }
 }
 
-impl<'w, 's, D: QueryData, F: QueryFilter> Iterator for DescendantIter<'w, 's, D, F>
+impl<'w, D: QueryData, F: QueryFilter> Iterator for DescendantIter<'w, '_, D, F>
```
alice-i-cecile, line 217: I really dislike this change too.

crates/bevy_macro_utils/src/symbol.rs
```diff
@@ -11,7 +11,7 @@ impl PartialEq<Symbol> for Ident {
     }
 }
 
-impl<'a> PartialEq<Symbol> for &'a Ident {
+impl PartialEq<Symbol> for &Ident {
```
alice-i-cecile, line 14: Good change.

alice-i-cecile:
After, looking at the 3 PRs in this group, I don't think we should do this. These are a serious downgrade in some cases (actual named lifetimes + 'w/'s), neutral in most cases and positive in a very small fraction of cases. There's important information being lost here in favor of a meaningless anonymous lifetime.

I'd prefer to close these PRs and blanket ignore the lint if it gets stabilized. I'll file an issue on the clippy repo too.

crates/bevy_audio/src/audio_output.rs
BenjaminBrienen: Can you provide a reasoning, @alice-i-cecile? I agree with your other ones, but this one seems to me like an improvement to conciseness with no loss in clarity.

crates/bevy_hierarchy/src/query_extension.rs
BenjaminBrienen: I don't like this change because there is still a lifetime parameter, so it's weird to switch between eliding and not eliding.

crates/bevy_hierarchy/src/query_extension.rs
```diff
@@ -259,7 +259,7 @@ where
     }
 }
 
-impl<'w, 's, D: QueryData, F: QueryFilter> Iterator for DescendantDepthFirstIter<'w, 's, D, F>
+impl<'w, D: QueryData, F: QueryFilter> Iterator for DescendantDepthFirstIter<'w, '_, D, F>
```
BenjaminBrienen, line 262: .

crates/bevy_hierarchy/src/query_extension.rs
```diff
@@ -298,7 +298,7 @@ where
     }
 }
 
-impl<'w, 's, D: QueryData, F: QueryFilter> Iterator for AncestorIter<'w, 's, D, F>
+impl<'w, D: QueryData, F: QueryFilter> Iterator for AncestorIter<'w, '_, D, F>
```
BenjaminBrienen, line 301: .

crates/bevy_tasks/src/task_pool.rs
```diff
@@ -602,7 +602,7 @@ pub struct Scope<'scope, 'env: 'scope, T> {
     env: PhantomData<&'env mut &'env ()>,
 }
 
-impl<'scope, 'env, T: Send + 'scope> Scope<'scope, 'env, T> {
+impl<'scope, T: Send + 'scope> Scope<'scope, '_, T> {
```
BenjaminBrienen, line 605: .

crates/bevy_tasks/src/task_pool.rs
```diff
@@ -655,7 +655,7 @@ impl<'scope, 'env, T: Send + 'scope> Scope<'scope, 'env, T> {
     }
 }
 
-impl<'scope, 'env, T> Drop for Scope<'scope, 'env, T>
+impl<'scope, T> Drop for Scope<'scope, '_, T>
```
BenjaminBrienen, line 658: .

crates/bevy_ui/src/ghost_hierarchy.rs
```diff
@@ -30,7 +30,7 @@ pub struct UiRootNodes<'w, 's> {
     ui_children: UiChildren<'w, 's>,
 }
 
-impl<'w, 's> UiRootNodes<'w, 's> {
+impl<'s> UiRootNodes<'_, 's> {
```
BenjaminBrienen, line 33: .

BenjaminBrienen:
I'm going to checkout this commit and play with the code a bit to see if I can do better than clippy at eliding the remaining lifetime parameters. Probably not.

PR #15908, opened 2024-10-14
Fix `clippy::needless_lifetimes` for `bevy_ecs` and `bevy_reflect` crates

clarfonthey:
Part "3" for #15906. This fixes `clippy::needless_lifetimes` for `bevy_reflect` and `bevy_ecs`, since those triggered the lint often.

I say "3" in quotes because all the lint fixes can be merged separately, and don't depend on each other. Although all of them are required to truly make `ci lints` pass without errors.


clarfonthey:
Actually, gonna close and split this up further. It's bigger than I thought.

PR #15913, opened 2024-10-14
[Adopted] Add a method for asynchronously waiting for an asset to load

bushrat011899:
# Objective

Currently, is is very painful to wait for an asset to load from the context of an `async` task. While bevy's `AssetServer` is asynchronous at its core, the public API is mainly focused on being used from synchronous contexts such as bevy systems. Currently, the best way of waiting for an asset handle to finish loading is to have a system that runs every frame, and either listens for `AssetEvents` or manually polls the asset server. While this is an acceptable interface for bevy systems, it is extremely awkward to do this in a way that integrates well with the `async` task system. At my work we had to create our own (inefficient) abstraction that encapsulated the boilerplate of checking an asset's load status and waking up a task when it's done.

## Solution

Add the method `AssetServer::wait_for_asset`, which returns a future that suspends until the asset associated with a given `Handle` either finishes loading or fails to load.

## Testing

- CI

## Notes

This is an adoption of #14431, the above description is directly from that original PR.


crates/bevy_asset/src/server/mod.rs
```diff
@@ -433,7 +433,10 @@ impl AssetServer {
         });
 
         #[cfg(not(any(target_arch = "wasm32", not(feature = "multi_threaded"))))]
-        infos.pending_tasks.insert(handle.id(), task);
+        {
+            let mut infos = infos;
+            infos.pending_tasks.insert(handle.id(), task);
+        }
```
bushrat011899, line 439: This is unrelated to the PR. I added this to suppress a Clippy warning around unused `mut`.

crates/bevy_asset/src/server/mod.rs
```diff
@@ -1336,6 +1339,110 @@ impl AssetServer {
...
+                                | (LoadState::Loaded, RecursiveDependencyLoadState::NotLoaded)
+                        );
+                        if !is_loading {
+                            cx.waker().wake_by_ref();
+                        } else {
+                            // Leave our waker behind
+                            info.waiting_tasks.push(cx.waker().clone());
+                        }
+                    }
+                    Poll::Pending
```
andriyDev, line 1434: Can we make `if !has_waker` a guard clause instead? As in `if has_waker { return Poll::Pending; } ... the unnested body`.

andriyDev (approved):

crates/bevy_asset/src/server/mod.rs
bushrat011899: I've done that _and_ moved the `poll_fn` into its own function to further reduce the indentation level.

MiniaczQ (approved):

PR #15914, opened 2024-10-14
screenshot comparison: fix upload for macOS... again

mockersf:
# Objective

- Used the wrong variable to set metadata
- new fixes after https://github.com/bevyengine/bevy/pull/15911

## Solution

- Use the right one
- Also keep a reference to the original PR when there's one


alice-i-cecile (approved):
Debugging CI is fun :)

PR #15916, opened 2024-10-15
Deny Rust 2018 idioms

BenjaminBrienen:
# Objective

Prepare for when this is deny by default

## Solution

Add `rust_2018_idioms = "deny"` to `[workspace.lints.rust]`
Add elided lifetime parameters
Remove unused extern crate declarations

## Testing

CI


alice-i-cecile:
The elided lifetime change is *really* bad for Bevy. We need to allow this when it's ready. I'll talk to the Rust team...

PR #10437, opened 2023-11-07
UI Materials: don't reserve items on phase from inside the loop

mockersf:
# Objective

- UI Materials reserve an additional nodes length for each node, leading to (number of nodes) ^2 reserved size of array

## Solution

- Reorder the loops and reserve length outside of the loop on nodes

PR #15919, opened 2024-10-15
UI materials: don't reserve in loop when enough capacity

mockersf:
# Objective

- UI materials reserve too much capacity in a vec: for every node in the transparent phase, it reserves enough memory to store all the nodes
- Update #10437 

## Solution

- Only reserve extra memory if there's not enough
- Only reserve the needed memory, not more

### DIFF
--- a/crates/bevy_ui/src/render/ui_material_pipeline.rs
+++ b/crates/bevy_ui/src/render/ui_material_pipeline.rs
@@ -643,9 +643,9 @@ pub fn queue_ui_material_nodes<M: UiMaterial>(
             },
         );
         if transparent_phase.items.capacity() < extracted_uinodes.uinodes.len() {
-            transparent_phase
-                .items
-                .reserve(extracted_uinodes.uinodes.len());
+            transparent_phase.items.reserve_exact(
+                transparent_phase.items.capacity() - extracted_uinodes.uinodes.len(),
+            );
         }
         transparent_phase.add(TransparentUi {
             draw_function,
--- a/crates/bevy_ui/src/render/ui_material_pipeline.rs
+++ b/crates/bevy_ui/src/render/ui_material_pipeline.rs
@@ -642,9 +642,11 @@ pub fn queue_ui_material_nodes<M: UiMaterial>(
                 bind_group_data: material.key.clone(),
             },
         );
-        transparent_phase
-            .items
-            .reserve(extracted_uinodes.uinodes.len());
+        if transparent_phase.items.capacity() < extracted_uinodes.uinodes.len() {
+            transparent_phase
+                .items
+                .reserve(extracted_uinodes.uinodes.len());
+        }
         transparent_phase.add(TransparentUi {
             draw_function,
             pipeline,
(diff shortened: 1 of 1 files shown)


crates/bevy_ui/src/render/ui_material_pipeline.rs
```diff
@@ -642,9 +642,11 @@ pub fn queue_ui_material_nodes<M: UiMaterial>(
                 bind_group_data: material.key.clone(),
             },
         );
-        transparent_phase
-            .items
-            .reserve(extracted_uinodes.uinodes.len());
+        if transparent_phase.items.capacity() < extracted_uinodes.uinodes.len() {
+            transparent_phase.items.reserve_exact(
+                transparent_phase.items.capacity() - extracted_uinodes.uinodes.len(),
```
ickshonpe, line 647: This doesn't look right, the capacity the first time the transparent items buffer is filled its capacity will be 0 so it should panic on subtraction with overflow?

ickshonpe, line 647: This doesn't look right, the capacity the first time the transparent items buffer is filled its capacity will be 0 so it should panic on subtraction with overflow?

ickshonpe (changes requested):

crates/bevy_ui/src/render/ui_material_pipeline.rs
mockersf: right it's the other way around 🤦

crates/bevy_ui/src/render/ui_material_pipeline.rs
```diff
@@ -642,9 +642,11 @@ pub fn queue_ui_material_nodes<M: UiMaterial>(
                 bind_group_data: material.key.clone(),
             },
         );
-        transparent_phase
-            .items
-            .reserve(extracted_uinodes.uinodes.len());
+        if transparent_phase.items.capacity() < extracted_uinodes.uinodes.len() {
```
ickshonpe, line 645: The transparent phase items vec is shared between all  the extracted UI items, not just uimaterial nodes, so this isn't quite right either. Maybe the current length of items needs to be stored before the start of the loop, then this condition could become `transparent_phase.items.capacity() - initial_len < extracted.uinodes.len()`?

ickshonpe (changes requested):

crates/bevy_ui/src/render/ui_material_pipeline.rs
mockersf: so it seems we have no way to know the exact size required on each transparent phase, as not all nodes may go to the same if there are several cameras

Current behaviour on main is that for each entity, it's reserving "number of entities" capacity, so this PR is still an improvement

ickshonpe: Yeah that makes sense. We should merge this and look for further improvements later.

ickshonpe (approved):

alice-i-cecile (approved):

Issue #15853, opened 2024-10-11
`pbr` example is missing one of it labels


rparrett:
## Bevy version

main
bisected to #14848

## Relevant system information

SystemInfo { os: "MacOS 14.5 ", kernel: "23.5.0", cpu: "Apple M1 Max", core_count: "10", memory: "64.0 GiB" }
AdapterInfo { name: "Apple M1 Max", vendor: 0, device: 0, device_type: IntegratedGpu, driver: "", driver_info: "", backend: Metal }

also tested win11+4080

## What you did

`cargo run --example pbr`

## What went wrong

The vertical "Metallic" label on the right side is missing.

ickshonpe:
Oh yeah, I checked through all the UI examples with this but forgot that none of them feature text rotation. I'll add some slanted text to one of the UI text examples so it doesn't get missed next time.

PR #15925, opened 2024-10-15
don't clip text that is rotated

### LINKED ISSUE #15922: Text rotation no more working properly
## Bevy version

acbed6040eeae99199657e44a9668772738ac344

## What you did

I was looking at issue #15853, and found out that this is the rotation of the text that make it disappear.
I wrote the following example, that rotate a text, that showcase the bug:
```rust
use bevy::prelude::*;

fn main() {
    App::new()
        .add_plugins(DefaultPlugins)
        .add_systems(Startup, setup)
        .add_systems(Update, text_rotation_system)
        .run();
}

    commands.spawn(Camera2d);
    commands.spawn((
        Text::new("abcdefghijklmnopqrstuvwxyz"),
        TextFont {
            font_size: 40.0,
            ..default()
        },
        TextLayout::new_with_justify(JustifyText::Center),
        Style {
            align_self: AlignSelf::Center,
            justify_self: JustifySelf::Center,
            ..default()
        },
    ));
}

fn text_rotation_system(time: Res<Time>, mut query: Query<&mut Transform, With<Text>>) {
    for mut transform in &mut query {
        let seconds = time.elapsed_seconds();
        transform.rotation = Quat::from_rotation_z(seconds / 2.0);
    }
}
```

## What went wrong

At some small rotation some letters disappear then the whole text disappear. See video:

https://github.com/user-attachments/assets/c941e152-8cca-4dfc-ba84-a995f00f1ff2

## Additional information

Fixing this issue should also fix #15863 directly.

mockersf:
# Objective

- Fixes #15922 , #15853 
- Don't clip text that is rotated by some angles

## Solution

- Compare to the absolute size before clipping

### DIFF
--- a/crates/bevy_ui/src/render/mod.rs
+++ b/crates/bevy_ui/src/render/mod.rs
@@ -1158,9 +1158,9 @@ pub fn prepare_uinodes(
                                 let transformed_rect_size =
                                     glyph.transform.transform_vector3(rect_size);
                                 if positions_diff[0].x - positions_diff[1].x
-                                    >= transformed_rect_size.x
+                                    >= transformed_rect_size.x.abs()
                                     || positions_diff[1].y - positions_diff[2].y
-                                        >= transformed_rect_size.y
+                                        >= transformed_rect_size.y.abs()
                                 {
                                     continue;
                                 }


ickshonpe (approved):
Makes sense. The "Material" label in the PBR example is visible now with the correct orientation.

MiniaczQ (approved):
Example works correctly after the change

PR #15926, opened 2024-10-15
Remove `ExtractComponent::Out`

akimakinai:
# Objective

- `C: ExtractComponent` inserts `C::Out` instead of `C`, so we need to remove `C::Out`. cc #15904.

## Solution

- `C` -> `C::Out`

## Testing

- CAS has `<ContrastAdaptiveSharpening as ExtractComponent>::Out = (DenoiseCas, CasUniform)`. Setting its strength to zero correctly removes the effect after this change.


MiniaczQ (approved):
Wasn't familiar with the trait in last PR my bad,
this looks appropriate

tychedelia (approved):

PR #15927, opened 2024-10-15
Fix overflow panic on Stopwatch at Duration::MAX

### LINKED ISSUE #15924: Stopwatch panic on reaching Duration::MAX
## Bevy version

0.14.2

#### Describe how you arrived at the problem. If you can, consider providing a code snippet or link.

hello !

I stumbled upon this as I was using a Timer with Duration::MAX and wanted to set it as finished as its used as a cooldown, but calling .tick() on a maxed stopwatch panic
```
thread 'main' panicked at /home/mirsella/.local/share/rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/time.rs:1125:31:
overflow when adding durations
```
repro/example:
```rust
use bevy::time::Stopwatch;
use std::time::Duration;

fn main() {
    let second = Duration::from_secs(1);

    let mut stopwatch = Stopwatch::new();
    // lot of time has passed... or a timer with Duration::MAX that was artificially set has "finished":
    // timer.set_elapsed(timer.remaining());
    stopwatch.set_elapsed(Duration::MAX);
    // panic
    stopwatch.tick(second);

    let mut stopwatch = Stopwatch::new();
    stopwatch.set_elapsed(Duration::MAX - second);
    // this doesnt panic as its still one off the max
    stopwatch.tick(second);
    // this panic
    stopwatch.tick(second);
}
```

from stopwatch impl:
```rust
    pub fn tick(&mut self, delta: Duration) -> &Self {
        if !self.paused() {
            self.elapsed += delta;
        }
        self
    }
```
Maybe we can just replace it with `self.elapsed.saturating_add(delta)`.
If that a valid solution, i would be happy to do a PR.

thanks !

mirsella:
See #15924 for more details
close #15924

from the issue, this code panic:
```rust
use bevy::time::Stopwatch;
use std::time::Duration;

fn main() {
    let second = Duration::from_secs(1);

    let mut stopwatch = Stopwatch::new();
    // lot of time has passed... or a timer with Duration::MAX that was artificially set has "finished":
    // timer.set_elapsed(timer.remaining());
    stopwatch.set_elapsed(Duration::MAX);
    // panic
    stopwatch.tick(second);

    let mut stopwatch = Stopwatch::new();
    stopwatch.set_elapsed(Duration::MAX - second);
    // this doesnt panic as its still one off the max
    stopwatch.tick(second);
    // this panic
    stopwatch.tick(second);
}
```

with this PR changes, the code now doesn't panic.

have a good day !

### DIFF
--- a/crates/bevy_time/src/stopwatch.rs
+++ b/crates/bevy_time/src/stopwatch.rs
@@ -129,7 +129,7 @@ impl Stopwatch {
     /// ```
     pub fn tick(&mut self, delta: Duration) -> &Self {
         if !self.paused() {
-            self.elapsed += delta;
+            self.elapsed = self.elapsed.saturating_add(delta);
         }
         self
     }


alice-i-cecile (approved):
Panics are bad! This is the most reasonable failsafe behavior.

MiniaczQ (approved):

PR #15929, opened 2024-10-15
Allow users to choose how `Commands` and `EntityCommands` should fail

### LINKED ISSUE #2004: Configurable error handling in commands

Commands (like removing components or despawning entities) may fail for various reasons. Right now, for the sake of idempotency (the same function applied twice should do the same thing as the function applied once), robustness and not spamming the user endlessly, these silently fail.

This is not always the desired behavior, and should be configurable at the call site.

### Approach to config

Global config is convenient for reducing boilerplate, but the desired behavior may fail by call site.

Local config is useful for both debugging and persistent configuration.

### Default behavior

This could panic, warn or fail silently.

### Patterns

More methods: this is bad for API explosion.

Extra argument: this is very verbose for common operations because Rust doesn't have default arguments

Builder pattern: nice, as long as we don't *have* to use it.

## Additional context

@Frizi, @BoxyUwU  and myself ran into this while discussing new commands for relations in https://github.com/bevyengine/rfcs/pull/18

### LINKED ISSUE #3845: EntityCommands should have methods to fail gracefully

Commands failing due to an entity being despawned is one of the most common and frustrating sources of panics in Bevy at the moment.

1. Enable better backtrace behavior to enable debugging. In particular, we need to know the list of other commands that operated on this entity, and which systems they were issued in.
2. A tool to enable entity commands to fail gracefully (silently ignore / log / warn) rather than panicking.

### Try

For each of the existing methods on `EntityCommands`, unwrap this option. Create a `try_X` variant for each method.

This creates significant API bloat and doesn't capture the way that entity commands are chained: after the first operation has succeeded, the rest of the operations in the chain are infallible.

### Verified entity flag

Store a `verified_entity_exists` Option<bool> flag on `EntityCommands`.  Each of the existing methods on `EntityCommands` should attempt to set this flag by unwrapping it to `Some(true) if it is `None`, and be skipped if it is false.

Then, create a few methods on `EntityCommands` that set the value of the verification flag with varying behavior:

1. `commands.entity(entity).silently_fail_if_missing().insert(Foo)`
2. `commands.entity(entity).log_if_missing().insert(Foo)`
3. `commands.entity(entity).warn_if_missing().insert(Foo)`

This doesn't work though, because the check is occuring at command-issuing time.

## Additional context

#2241 proposes a much more general approach to this, but has significant complexity and overhead.

#3757 aims for a similar goal, but doesn't actually create a reliable API: entities can be despawned after we check for their existence, resulting in the same panic.

@emiongit expressed interest in this problem in #3840.

### LINKED ISSUE #7118: Potential surprising "entity does not exist" panic due to asynchronous command execution
## Bevy version

0.9.1

## What went wrong

These two systems look innocent on their own, but may cause panic if used together.

```
fn insert(mut commands: Commands, query: Query<Entity, With<Marker>>, ...) {
    for entity in query.iter() {
        if ... {
            // We obtained the entity from the query, so it must be valid, right?
            commands.entity(entity).insert(Foo);
        }
    }
}

fn despawn(mut commands: Commands, query: Query<Entity, With<Marker>>, ...) {
    for entity in query.iter() {
        if ... {
            // This command may execute before the insertion and cause a panic there!
        }
    }
}
```

This panic is hard to avoid in user code, as there is no way to know if there is a pending despawn command for a certain entity.

## Additional information

The `Insert` command is, as far as I know, the only place where a missing entity in command execution is turned into a panic. We should at least replace the panic with a soft error or warning like the other commands.

We are also currently inconsistent in the way missing entities are reported. `Insert` uses panics, `Remove` uses no error reporting at all (we may want to change this), `Despawn` uses warnings, and `DespawnRecursive` use debug messages.

### LINKED ISSUE #10166: Commands should be infallible whenever possible
> Commands in general shouldn't even have errors, and all our built-in commands should be infallible.
>
> The panic that happens when an `insert` or `remove` comes after a `despawn` is something we just arbitrarily do. They should just be ignored (with a `debug!` warning). It isn't a bug if two different systems from two different plugins want to despawn an entity at the same time for different reasons, or if a plugin system wants to despawn an entity at the same time user code wants to give it a component, assuming it'll still exist. Batching does not add anything new to this equation.
>
> All we need to do is make a PR that (1) changes the panic to a `debug!` warning and (2) privates all fields of [`Insert`][1], [`Remove`][2], and [`Despawn`][3] so that users can only make these commands through `EntityCommands`. You can't get an `EntityCommands` for an entity [unless that entity is alive][4] (or will be), so all commands created through `EntityCommands` _must_ be considered valid.
>
> Since commands modify the shared state of the world, I don't think they could be reordered safely, unless we know ahead of time the set of all commands and all of their interactions, which is impossible with custom commands and will make adding new commands very difficult and fragile.
>
> The reality is that custom commands should never make implicit assumptions about which entities exist and what components they may or may not have. i.e. If you pushed `<custom command 1>` assuming that `Mass` won't have been added to `e2` yet, that isn't safe to assume. I don't think we explicitly encouraged assuming things before, but it'd be actively discouraged after this.

Originally posted by @maniwani in https://github.com/bevyengine/bevy/issues/10154#issuecomment-1767078749.

Having seen the problems that can arise, I think I generally agree. It's very hard to get a usable error handling solution (#2004, #2241) without severely compromising performance. And in any case, panicking is a very bad default, both for beginners (frustration), prototyping (wasted time) and production games (crashes are very bad).

# Steps to fix

- remove `try_insert` method
- swap all panics to a `debug` warning in commands
- make fields private
- get SME-ECS to agree that we should do this

JaySpruce:
## Objective
- Fixes #2004
- Fixes #3845
- Fixes #7118
- Fixes #10166

## Solution
Added the following `Commands`:
- `ignore_on_error`
- `log_on_error` (sends messages using `info!`)
- `warn_on_error` (sends messages using `warn!`)
- `panic_on_error`

and the following `EntityCommands`:
- `ignore_if_missing`
- `log_if_missing`
- `warn_if_missing`
- `panic_if_missing`

These change how subsequent commands will respond if errors occur.

Internally:
- Added a `failure_handling_mode` field to `Commands`.
- When a fallible `Command` or an `EntityCommand` is queued, this field is sent with it.
- Fallible `Command`s should check for errors themselves.
- When any `EntityCommand` is applied, it automatically checks if the entity exists.
- If an error occurs, the command matches against `failure_handling_mode` to determine how to respond.

## Showcase

```
#[derive(Component)]
struct X;

let mut queue_a = CommandQueue::default();
let mut queue_b = CommandQueue::default();

let mut commands_a = Commands::new(&mut queue_a, &world);
let mut commands_b = Commands::new(&mut queue_b, &world);

let entity = commands_a.spawn_empty().id();

commands_a.entity(entity).despawn();
commands_b.entity(entity).warn_if_missing().insert(X);

queue_a.apply(&mut world);
queue_b.apply(&mut world); // Doesn't panic
```

## Review Notes
The real changes are all in `system/commands/mod.rs` and `world/mod.rs`.
All the other files are just returning some commands to their old behavior:

- `despawn()` -> `warn_if_missing().despawn()`
- `remove()` -> `ignore_if_missing().remove()`

These two are the only commands that both 1) didn't already panic by default and 2) were used internally (according to VS Code's "Find All References" anyway).
They might not all need to be kept non-panicking, I just didn't want to regress anything.

## Bikeshedding
- `ignore` vs `silent` vs something else?
- `failure_handling_mode` vs `failure_response` vs something else? (Used to be `failure_mode`, but that's wrong)
- `info!` vs `debug!`?

## Possible Additions / Future PRs
- Deprecate `try_` variants of commands
  - I added some comments explaining the change and directing the user to the normal variants, but held off on deprecating since AFAIK that's usually a separate PR.
- Location tracking
  - Just call `Location::caller` in the user-facing command (if not set to `ignore`) and pass it in, so the user can see exactly which command caused an error. Didn't seem to hurt performance meaningfully in my testing, but maybe I'm not the best at benchmarking.
- Generalized errors for `Commands` using `Result`
  - We can't return a `Result` from a command to the caller, but we could wrap commands in something that handles any `Err` returned from the internal function call.

## Migration Guide

The following commands that used to fail silently will now panic by default:

- `EntityCommands::remove`
- `EntityCommands::remove_by_id`
- `EntityCommands::remove_with_requires`
- `EntityCommands::clear`
- `EntityCommands::retain`
- `EntityCommands::observe`
- `EntityEntryCommands::and_modify`

Use `ignore_if_missing` to achieve the previous behavior:
```
commands.entity(entity).ignore_if_missing().clear();
```

The command `EntityCommands::despawn` used to warn upon failure and will now panic by default.
Use `warn_if_missing` to achieve the previous behavior:
```
commands.entity(entity).warn_if_missing().despawn();
```

`try` variants of commands are unchanged and will always fail silently; these will be removed in the future.


JaySpruce:
I don't know how to do Tracy stuff, but I ran the "spawn_commands" benchmark which does this:
```
for i in 0..entity_count {
    let mut entity = commands.spawn_empty();
    entity
        .insert_if(A, || black_box(i % 2 == 0))
        .insert_if(B, || black_box(i % 3 == 0))
        .insert_if(C, || black_box(i % 4 == 0));

    if black_box(i % 5 == 0) {
        entity.despawn();
    }
}
```

and got these results:

# main:
Code_rKDHOq7DKF

# this PR:
Code_jD1W6zViIl

JaySpruce:
Maybe controversial, but since invalid `EntityCommands` just do nothing now, I think it's okay to let `Commands::entity` return one if the entity doesn't exist. It'll panic if the user wants it to panic, but otherwise it'll just complain or do nothing when they try to use it. `Commands::get_entity` is still useful if they want to handle it properly.

JaySpruce:
I've got a version working that tracks the calling location of `EntityCommands`, so that the error messages can be more useful:
Code_aMsZ0VVLHl


It hurts performance of `warn_if_missing` by another ~2%. Other than that, the biggest downside is that `EntityCommands` all look like this:

```
pub fn insert(&mut self, bundle: impl Bundle) -> &mut Self {
    if self.failure_mode().should_respond() {
        self.queue_with_caller(insert(bundle, InsertMode::Replace), Location::caller())
    } else {
        self.queue(insert(bundle, InsertMode::Replace))
    }
}
```

Could leave it for another PR, or could merge it here if it's wanted

BenjaminBrienen:
Bikeshedding: I thought a "failure mode" or a "mode of failure" is the way something went wrong. This type/field describes the strategy for handling when something goes wrong.

crates/bevy_ecs/src/system/commands/mod.rs
```diff
@@ -399,9 +458,17 @@ impl<'w, 's> Commands<'w, 's> {
 
     /// Returns the [`EntityCommands`] for the requested [`Entity`].
     ///
-    /// # Panics
+    /// This method does not guarantee that `EntityCommands` will be successfully applied,
+    /// since another command in the queue may delete the entity before them.
```
BenjaminBrienen, line 462: What is "them"?

crates/bevy_ecs/src/system/commands/mod.rs
```diff
@@ -1027,15 +1162,48 @@ pub trait EntityCommand<Marker = ()>: Send + 'static {
...
     {
-        move |world: &mut World| self.apply(entity, world)
+        move |world: &mut World| {
+            if world.entities.contains(entity) {
+                self.apply(entity, world);
+            } else {
+                match failure_mode {
+                    FailureMode::Ignore => (),
+                    FailureMode::Log => {
+                        info!("Could not execute EntityCommand because its Entity {entity:?} was missing");
```
BenjaminBrienen, line 1176: You may want to use Display formatting. There is an open issue about consistently preferring Display for Entity.

JaySpruce:
Hadn't heard of that, but that does seem to be the case. Maybe just `failure_handling_mode`? Doesn't really need to be short since it's not very public-facing.

Totally open to more bikeshedding, I'm not committed to any of the names here

crates/bevy_ecs/src/system/commands/mod.rs
```diff
@@ -1665,6 +1843,52 @@ pub struct EntityEntryCommands<'a, T> {
 }
 
 impl<'a, T: Component> EntityEntryCommands<'a, T> {
+    /// Sets the [`EntityEntryCommands`] instance to ignore commands if the entity doesn't exist.
```
BenjaminBrienen, line 1846: Inconsistent "when a command is executed" missing from doc comment

crates/bevy_ecs/src/system/commands/mod.rs
```diff
@@ -1706,19 +1926,15 @@ impl<'a, T: Component> EntityEntryCommands<'a, T> {
...
     #[track_caller]
     pub fn or_insert_with(&mut self, default: impl Fn() -> T) -> &mut Self {
         self.or_insert(default())
     }
 
     /// [Insert](EntityCommands::insert) the value returned from `default` into this entity, if `T` is not already present.
     ///
-    /// Unlike [`or_insert_with`](Self::or_insert_with), this will not panic if the entity does not exist.
+    /// [`Self::or_insert_with`] used to panic if the entity was missing, and this was the non-panicking version.
+    /// `EntityCommands` no longer need to handle missing entities individually, so just use [`Self::or_insert_with`]
```
BenjaminBrienen, line 1937: Missing full stop `.`

crates/bevy_ecs/src/system/commands/mod.rs
```diff
@@ -1693,7 +1912,8 @@ impl<'a, T: Component> EntityEntryCommands<'a, T> {
 
     /// [Insert](EntityCommands::insert) `default` into this entity, if `T` is not already present.
     ///
-    /// Unlike [`or_insert`](Self::or_insert), this will not panic if the entity does not exist.
+    /// [`Self::or_insert`] used to panic if the entity was missing, and this was the non-panicking version.
+    /// `EntityCommands` no longer need to handle missing entities individually, so just use [`Self::or_insert`]
```
BenjaminBrienen, line 1916: Missing full stop `.`

BenjaminBrienen (approved):
I love it!

crates/bevy_ecs/src/system/commands/mod.rs
JaySpruce: Does this read better?

> This method does not guarantee that commands queued by the `EntityCommands` will be successful, since the entity could be despawned before they are executed.

BenjaminBrienen: Yes

alice-i-cecile:
Closing in favor of your newer #17043.

PR #15930, opened 2024-10-15
Revert default mesh materials

### LINKED ISSUE #15799: Revert default mesh material
Yup lets close this in favor of a full remove of the default behavior (including the HasMaterial3d/2d components)

_Originally posted by @cart in https://github.com/bevyengine/bevy/issues/15793#issuecomment-2403421027_

Jondolf:
# Objective

Closes #15799.

Many rendering people and maintainers are in favor of reverting default mesh materials added in #15524, especially as the migration to required component is already large and heavily breaking.

## Solution

Revert default mesh materials, and adjust docs accordingly.

- Remove `extract_default_materials`
- Remove `clear_material_instances`, and move the logic back into `extract_mesh_materials`
- Remove `HasMaterial2d` and `HasMaterial3d`
- Change default material handles back to pink instead of white
    - 2D uses `Color::srgb(1.0, 0.0, 1.0)`, while 3D uses `Color::srgb(1.0, 0.0, 0.5)`. Not sure if this is intended.

There is now no indication at all about missing materials for `Mesh2d` and `Mesh3d`. Having a mesh without a material renders nothing.

## Testing

I ran `2d_shapes`, `mesh2d_manual`, and `3d_shapes`, with and without mesh material components.

### DIFF
--- a/crates/bevy_pbr/src/lib.rs
+++ b/crates/bevy_pbr/src/lib.rs
@@ -422,13 +422,13 @@ impl Plugin for PbrPlugin {
             app.add_plugins(DeferredPbrLightingPlugin);
         }
 
-        // Initialize the default material.
+        // Initialize the default material handle.
         app.world_mut()
             .resource_mut::<Assets<StandardMaterial>>()
             .insert(
                 &Handle::<StandardMaterial>::default(),
                 StandardMaterial {
-                    base_color: Color::WHITE,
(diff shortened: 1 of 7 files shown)


tychedelia (approved):

ecoskey (approved):
Have we decided on not needing/wanting a debug warning at least? Otherwise LGTM

PR #15387, opened 2024-09-23
Documentation for variadics

BenjaminBrienen:
# Objective

Relevant: #15208

## Solution

I went ahead and added the variadics documentation in all applicable locations.

## Testing

- I built the documentation and inspected it to see whether the feature is there.

PR #15931, opened 2024-10-15
Infer `StableInterpolate` on tuples

mweatherley:
# Objective

Make `StableInterpolate` "just work" on tuples whose parts are each `StableInterpolate` types. These types arise notably through `Curve::zip` (or just through explicit mapping of a similar form). It would otherwise be kind of frustrating to stumble upon such a thing and then realize that, e.g., automatic resampling just doesn't work, even though there is a very "obvious" way to do it.

## Solution

Infer `StableInterpolate` on tuples of up to size 11. I can make that number bigger, if desired. Unfortunately, I don't think that our standard "fake variadics" tools actually work for this; the anonymous field accessors of tuples are `:tt` for purposes of macro expansion, which means that you can't simplify away the identifiers by doing something clever like using recursion (which would work if they were `:expr`). Maybe someone who knows some incredibly dark magic could chime in with a better solution. 

The expanded impls look like this:
```rust
impl<
        T0: StableInterpolate,
        T1: StableInterpolate,
        T2: StableInterpolate,
        T3: StableInterpolate,
        T4: StableInterpolate,
    > StableInterpolate for (T0, T1, T2, T3, T4)
{
    fn interpolate_stable(&self, other: &Self, t: f32) -> Self {
        (
            <T0 as StableInterpolate>::interpolate_stable(&self.0, &other.0, t),
            <T1 as StableInterpolate>::interpolate_stable(&self.1, &other.1, t),
            <T2 as StableInterpolate>::interpolate_stable(&self.2, &other.2, t),
            <T3 as StableInterpolate>::interpolate_stable(&self.3, &other.3, t),
            <T4 as StableInterpolate>::interpolate_stable(&self.4, &other.4, t),
        )
    }
}
```

## Testing

Expanded macros; it compiles.

## Future

Make a version of the fake variadics workflow that supports this kind of thing.

### DIFF
--- a/crates/bevy_math/src/common_traits.rs
+++ b/crates/bevy_math/src/common_traits.rs
@@ -309,3 +309,81 @@ impl StableInterpolate for Dir3A {
         self.slerp(*other, t)
     }
 }
+
+macro_rules! impl_stable_interpolate_tuple {
+    ($(($T:ident, $n:tt)),*) => {
+        impl<$($T: StableInterpolate),*> StableInterpolate for ($($T,)*) {
+            fn interpolate_stable(&self, other: &Self, t: f32) -> Self {
+                (
+                    $(
+                        <$T as StableInterpolate>::interpolate_stable(&self.$n, &other.$n, t),
+                    )*
+                )
+            }
+        }
+    };
+}
+
+// (See `macro_metavar_expr`, which might make this better.)
+// This currently implements `StableInterpolate` for tuples of up to 11 elements.
+impl_stable_interpolate_tuple!((T, 0));
+impl_stable_interpolate_tuple!((T0, 0), (T1, 1));
+impl_stable_interpolate_tuple!((T0, 0), (T1, 1), (T2, 2));
+impl_stable_interpolate_tuple!((T0, 0), (T1, 1), (T2, 2), (T3, 3));
+impl_stable_interpolate_tuple!((T0, 0), (T1, 1), (T2, 2), (T3, 3), (T4, 4));
+impl_stable_interpolate_tuple!((T0, 0), (T1, 1), (T2, 2), (T3, 3), (T4, 4), (T5, 5));
+impl_stable_interpolate_tuple!(
+    (T0, 0),
+    (T1, 1),
+    (T2, 2),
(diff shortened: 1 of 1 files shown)


alice-i-cecile (changes requested):
Can you add the fake_variadics doc attribute? We've added this elsewhere and it really improves the rustdoc output.

I would suggest using the all_tuples macro here, but it's good to not pull in that dependency.

NiseVoid (approved):
Kind of cursed, but it does look like this is the only way without introducing new deps.

mweatherley:
> Can you add the fake_variadics doc attribute? We've added this elsewhere and it really improves the rustdoc output.

Do you know how to do that outside of the context of the `all_tuples` macro? Everything I try basically winds up with the compiler complaining to me about how I'm not allowed to use `#[doc(fake_variadics)]`, even when I add a module-level `allow`.

> I would suggest using the all_tuples macro here, but it's good to not pull in that dependency.

It doesn't work anyway :( I spent quite a while trying to do it that way and reached the conclusion that it's basically impossible.

alice-i-cecile (approved):
I'm fine with this for now :)

alice-i-cecile:
@BenjaminBrienen @bash y'all have any advice? You were very helpful for this in https://github.com/bevyengine/bevy/pull/15387

mweatherley:
> I'm fine with this for now :)

For what it's worth, I got it working with a new version of `all_tuples` in my branch now. At least long-term, I'd also prefer to do it that way, since I have other tuples where I want to derive the same thing. I'd be happy to do the shove the more complicated thing into a follow-up, too, though.

BenjaminBrienen:
I had the same errors he's describing. It only worked for me within `all_tuples!`.

alice-i-cecile:
Follow-up it is then ;)

bash:
> @BenjaminBrienen @bash y'all have any advice? You were very helpful for this in https://github.com/bevyengine/bevy/pull/15387

Yeah `#[doc(fake_variadic)]` can be a bit of a hassle :))
I created a follow-up PR that adds it: #15933

PR #15933, opened 2024-10-15
Use `#[doc(fake_variadic)]` on `StableInterpolate`

bash:
This is a follow-up to #15931 that adds `#[doc(fake_variadic)]` for improved docs output :)


alice-i-cecile (approved):
Lovely explanation, and thanks for the fix <3

PR #15934, opened 2024-10-15
Observers to log debug message when no observers found for event type…

### LINKED ISSUE #15921: Log when an event was triggered but no observer consumed it

We can easily see where an event was triggered from, but not that it didn't cause any observers to run. We should add a (feature flagged) debug message for when an event is triggered, but no observer consumes it.

Log when an event is not observed by anything. This should be behind a non-default feature flag. We should also consider adding an umbrella "debug" feature flags for all our various ones.

andristarr:
Fixes #15921 

I was using the tracing feature if there are objections: do we want to add a feature for this as there is no current debug or verbose feature?

### DIFF
--- a/crates/bevy_ecs/src/observer/mod.rs
+++ b/crates/bevy_ecs/src/observer/mod.rs
@@ -288,6 +288,11 @@ impl Observers {
             world.increment_trigger_id();
             let observers = world.observers();
             let Some(observers) = observers.try_get_observers(event_type) else {
+                bevy_utils::tracing::debug!(
+                    "No observers found for event type {:?} in invoke",
+                    event_type
+                );
+
                 return;
             };
             // SAFETY: The only outstanding reference to world is `observers`


crates/bevy_ecs/src/observer/mod.rs
```diff
@@ -288,6 +288,12 @@ impl Observers {
             world.increment_trigger_id();
             let observers = world.observers();
             let Some(observers) = observers.try_get_observers(event_type) else {
+                #[cfg(feature = "trace")]
```
alice-i-cecile, line 291:
```suggestion
```

andristarr: Do you not want to add a new feature then for debugging? I can add a new feature called `debug` for example and use it here, so it can be used later.

andristarr: I have addressed these nevertheless and if you feel content with its current state this can be merged, but feel free to let me know if you want me to add the feature too in the scope of this work :)

alice-i-cecile: I like the idea 🤔 Let's split it out though.

alice-i-cecile (approved):

mockersf:
The log given is:
```
 No observers found for event type ComponentId(390) in invoke
```

This is not very helpful.

I think this feature is controversial, because it's expected that Bevy or third party plugins trigger events, and not all of them will have observers plugged. Events are one of the extension points, and won't be always used and it's perfectly normal

PR #15935, opened 2024-10-15
Improve and Debug CI `compile-check-no-std` Command

bushrat011899:
# Objective

- Fix bug where `cargo run -p ci` fails due to differing implementations for default values between `Default` trait and `argh`
- Automatically install target via `rustup` to make first-run simpler.

## Solution

The command will now attempt to install the target via `rustup` by default, which will provide a cleaner error message if a malformed target is passed. It will also avoid confusion when people attempt to use this tool for the first time without the target already installed.

I've added a flag, --skip-install, to disable the attempted installation just-in-case. (e.g., maybe rustup isn't in their path but cargo is?).

Also fixed a bug where the default value for `target` was different between the `Default` trait and `argh`, causing `cargo run -p ci` to fail.

## Testing

- CI
- Subcommand ran directly

## Notes

This issue was originally discovered by @targrub (on Discord):

> Unfortunately, running `cargo run -p ci` still gives that same error as I initially reported (though `cargo run -p ci -- compile-check-no-std` succeeds).  This is after having run `rustup target add x86_64-unknown-none`.

### DIFF
--- a/tools/ci/src/commands/compile_check_no_std.rs
+++ b/tools/ci/src/commands/compile_check_no_std.rs
@@ -3,39 +3,65 @@ use argh::FromArgs;
 use xshell::cmd;
 
 /// Checks that the project compiles for a `no_std` target.
-#[derive(FromArgs, Default)]
+/// Note that this tool will attempt to install the target via rustup.
+/// This can be skipped by passing the "--skip-install" flag.
+#[derive(FromArgs)]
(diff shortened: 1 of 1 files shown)


targrub:
The above changes solved the issue for me.

BenjaminBrienen (approved):

alice-i-cecile (approved):

PR #15939, opened 2024-10-16
Delete `ImageLoader::COUNT` in favour of `ImageLoader::SUPPORTED_FORMATS.len()`.

andriyDev:
# Objective

- This is a followup to #15812.

## Solution

I just deleted the `COUNT` const and replaced it. I didn't realize for loops are not const yet, so improving the other const variables is not obvious.

Note: `slice::len` has been const since Rust 1.39, so we're not relying on a brand new feature or anything.

## Testing

- It builds!

### DIFF
--- a/crates/bevy_image/src/image_loader.rs
+++ b/crates/bevy_image/src/image_loader.rs
@@ -12,72 +12,6 @@ pub struct ImageLoader {
 }
 
 impl ImageLoader {
-    /// Number of image formats, used for computing other constants.
-    const COUNT: usize = {
-        let mut count = 0;
-        #[cfg(feature = "avif")]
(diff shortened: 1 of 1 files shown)


andriyDev:
@bushrat011899 We could also replace `COUNT_FILE_EXTENSIONS` with `Self::SUPPORTED_FORMATS.len()` since today the formats map one-to-one with extensions, but this one feels a lot less bad to me. This just future proofs the code for a tiny cost, and it feels more "algorithm-y" so keeping it is more reliable.

alice-i-cecile (approved):

bushrat011899 (approved):

PR #15943, opened 2024-10-16
Export `Val` Variants

### LINKED ISSUE #15937: Add Val helper functions
I'm closing this in favor of `px(10.)`-style helper functions (or alternatively, exporting `Val::Px` as `Px()` in the prelude)

_Originally posted by @cart in https://github.com/bevyengine/bevy/issues/15825#issuecomment-2414922363_

bushrat011899:
# Objective

- Fixes #15937

## Solution

- Export the variants of `Val` in the `bevy_ui` prelude.
- Updated examples to use the shortened names.

## Testing

- Ran CI.

## Notes

I'm not totally convinced this is _that_ big of a deal ergonomically. It's pretty easy for the end-user to write `use Val::{...};` at the top of their own files if it really helps them. This was a simple enough PR that I just wanted to create this as a way to decide whether #15937 should just be closed.

### DIFF
--- a/crates/bevy_ui/src/layout/mod.rs
+++ b/crates/bevy_ui/src/layout/mod.rs
@@ -591,8 +591,8 @@ mod tests {
         let ui_root = world
             .spawn(NodeBundle {
                 style: Style {
-                    width: Val::Percent(100.),
-                    height: Val::Percent(100.),
+                    width: Percent(100.),
+                    height: Percent(100.),
                     ..default()
                 },
                 ..default()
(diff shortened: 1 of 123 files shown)


ickshonpe (commented):
I don't hate this but I have a strong preference for the free functions alternative. Except for ubiquitous stuff like `Option` I don't like enum exports, (like you say) users can easily add `bevy::Val::*` to their own project's prelude if they want to,  `px` is easier to type than `Px` and a free function API would be more flexible. A `px` function could support:
```
pub trait AsValNum {
    fn as_val_num(self) -> f32;
}

impl AsValNum for i32 {
    fn as_val_num(self) -> f32 {
        self as f32
    }
}

impl AsValNum for f32 {
    fn as_val_num(self) -> f32 {
        self
    }
}

pub fn px(n: AsValNum) -> Val {
    Val::Px(n.as_val_num())
}
```
for instance. Maybe `AsValNum` (or whatever) would be controversial and I'm not suggesting it has to be included but it's available to us as an option with free functions. I just find it unpleasant that probably 99% of Bevy UI code uses integers but the API has to require floats everywhere.

PR #15947, opened 2024-10-16
Fix typos from greyscale -> grayscale

mgi388:
# Objective

I was grepping for "grayscale" and thought I'd fix these while I'm here so I don't need to look for both forms.

## Solution

From [wikipedia]:

> In [digital photography](https://en.wikipedia.org/wiki/Digital_photography), [computer-generated imagery](https://en.wikipedia.org/wiki/Computer-generated_imagery), and [colorimetry](https://en.wikipedia.org/wiki/Colorimetry), a greyscale (more common in [Commonwealth English](https://en.wikipedia.org/wiki/Commonwealth_English)) or grayscale (more common in [American English](https://en.wikipedia.org/wiki/American_English))

[wikipedia]: https://en.wikipedia.org/wiki/Grayscale

### DIFF
--- a/crates/bevy_image/src/image.rs
+++ b/crates/bevy_image/src/image.rs
@@ -1155,7 +1155,7 @@ impl Image {
     /// If you are working with a 32-bit integer [`TextureFormat`], the value will be
     /// inaccurate (as `f32` does not have enough bits to represent it exactly).
     ///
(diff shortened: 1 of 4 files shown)


Jondolf (approved):
Bevy uses American English, so this makes sense

tim-blackbird (approved):

Issue #15871, opened 2024-10-12
Conditional extraction to render world can cause desync


akimakinai:
In retained render world, despawning the main world entity or removing its components (#15582) are the only way to remove already inserted components from render world.
Some extraction systems seem to rely on the previous behavior by extracting components conditionally, expecting render entities to be wiped out every frame.

- [x] Bloom
#15856
https://github.com/bevyengine/bevy/blob/813c75958df8a518353c6b50cb62a27353bbfd51/crates/bevy_core_pipeline/src/bloom/settings.rs#L222-L229
Turning HDR off causes crash because `Bloom` component is not removed.

- [x] ChromaticAberration
https://github.com/bevyengine/bevy/blob/813c75958df8a518353c6b50cb62a27353bbfd51/crates/bevy_core_pipeline/src/post_process/mod.rs#L499
Setting `intensity = 0` will not be synchronized to render world if you have non-zero value currently, for example.

- [x] ContrastAdaptiveSharpening
https://github.com/bevyengine/bevy/blob/813c75958df8a518353c6b50cb62a27353bbfd51/crates/bevy_core_pipeline/src/contrast_adaptive_sharpening/mod.rs#L88

- [ ] extract_depth_of_field_settings
https://github.com/bevyengine/bevy/blob/813c75958df8a518353c6b50cb62a27353bbfd51/crates/bevy_core_pipeline/src/dof/mod.rs#L825

- [ ] extract_taa_settings
https://github.com/bevyengine/bevy/blob/813c75958df8a518353c6b50cb62a27353bbfd51/crates/bevy_core_pipeline/src/taa/mod.rs#L376

- ~~extract_camera_prepass_phase~~
https://github.com/bevyengine/bevy/blob/813c75958df8a518353c6b50cb62a27353bbfd51/crates/bevy_core_pipeline/src/core_3d/mod.rs#L589

- [ ] extract_clusters
https://github.com/bevyengine/bevy/blob/813c75958df8a518353c6b50cb62a27353bbfd51/crates/bevy_pbr/src/cluster/mod.rs#L533

- [ ] extract_camera_previous_view_data
https://github.com/bevyengine/bevy/blob/813c75958df8a518353c6b50cb62a27353bbfd51/crates/bevy_pbr/src/prepass/mod.rs#L587

- [ ] extract_lights
[...]

PR #15948, opened 2024-10-16
Remove components if not extracted

akimakinai:
# Objective

- Fixes https://github.com/bevyengine/bevy/issues/15871
(Camera is done in #15946)

## Solution

- Do the same as #15904 for other extraction systems
- Added missing `SyncComponentPlugin` for DOF, TAA, and SSAO
(According to the [documentation](https://dev-docs.bevyengine.org/bevy/render/sync_component/struct.SyncComponentPlugin.html), this plugin "needs to be added for manual extraction implementations." We may need to check this is done.)

## Testing

Modified example locally to add toggles if not exist.
- [x] DOF - toggling DOF component and perspective in `depth_of_field` example
- [x] TAA - toggling `Camera.is_active` and TAA component
- [x] clusters - not entirely sure, toggling `Camera.is_active` in `many_lights` example (no crash/glitch even without this PR)
- [x] previous_view - toggling `Camera.is_active` in `skybox` (no crash/glitch even without this PR)
- [x] lights - toggling `Visibility` of `DirectionalLight` in `lighting` example
- [x] SSAO - toggling `Camera.is_active` and SSAO component in `ssao` example
- [x] default UI camera view - toggling `Camera.is_active` (nop without #15946 because UI defaults to some camera even if `DefaultCameraView` is not there)
- [x] volumetric fog - toggling existence of volumetric light. Looks like optimization, no change in behavior/visuals

### DIFF
--- a/crates/bevy_core_pipeline/src/dof/mod.rs
+++ b/crates/bevy_core_pipeline/src/dof/mod.rs
@@ -331,7 +331,6 @@ pub struct DepthOfFieldPipeline {
 
 impl ViewNode for DepthOfFieldNode {
     type ViewQuery = (
-        Read<DepthOfField>,
         Read<ViewUniformOffset>,
         Read<ViewTarget>,
         Read<ViewDepthTexture>,
@@ -346,7 +345,6 @@ impl ViewNode for DepthOfFieldNode {
         _: &mut RenderGraphContext,
         render_context: &mut RenderContext<'w>,
         (
-            _depth_of_field,
             view_uniform_offset,
             view_target,
             view_depth_texture,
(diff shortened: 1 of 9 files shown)


alice-i-cecile (approved):

alice-i-cecile:
@rafalh are you able to review this PR for me in turn? :) This is closely related to #15946.

crates/bevy_pbr/src/render/light.rs
```diff
@@ -379,6 +379,14 @@ pub fn extract_lights(
     ) in &directional_lights
     {
         if !view_visibility.get() {
+            commands
+                .get_entity(entity)
+                .expect("Light entity wasn't synced.")
+                .remove::<(
+                    ExtractedDirectionalLight,
+                    RenderCascadesVisibleEntities,
+                    LightViewEntities,
```
rafalh, line 388: `LightViewEntities` is added in an observer and not in this system, so it's not clear at the first glance why we remove it here. Maybe we should remove it in an observer too to be consistent? :thinking:

crates/bevy_core_pipeline/src/dof/mod.rs
```diff
@@ -342,6 +346,7 @@ impl ViewNode for DepthOfFieldNode {
         _: &mut RenderGraphContext,
         render_context: &mut RenderContext<'w>,
         (
+            _depth_of_field,
```
rafalh, line 349: not sure why do we need this and it seems like a hack to me

rafalh (commented):
> @rafalh are you able to review this PR for me in turn? :) This is closely related to https://github.com/bevyengine/bevy/pull/15946.

@alice-i-cecile  I tried but I don't think I understand enough about the render graph to fully review this. I left some comments through. Most of changes looks okay to me.

tychedelia (approved):
Thanks for leaving the breadcrumbs to improve this later.

PR #15950, opened 2024-10-16
register `TextFont` and `TextColor` in app type registry

atornity:
# Objective

`TextFont`  and `TextColor` is not registered in the app type registry and serializing a scene with a a `Text2d` doesn't save the color and font of the text entity.

## Solution

register `TextFont`  and `TextColor`  in the type registry


crates/bevy_text/src/lib.rs
```diff
@@ -114,6 +114,7 @@ impl Plugin for TextPlugin {
     fn build(&self, app: &mut App) {
         app.init_asset::<Font>()
             .register_type::<Text2d>()
+            .register_type::<TextStyle>()
```
alice-i-cecile, line 117: I think we probably want `TextColor` too?

ickshonpe (approved):
Looks good.

PR #15953, opened 2024-10-16
Default UI shadow samples fix

ickshonpe:
# Objective

  In `queue_shadows`, the `UiBoxShadows` option is unwrapped incorrectly which results in the number of shadow samples being set to `u32::default()` instead of `UiBoxShadows::default()` if the camera entity doesn't have the component. 

## Solution

Just use `unwrap_or_default` directly without `map`.


alice-i-cecile (approved):

alice-i-cecile:
Great PR description: really clearly communicates the bug and the fix.

Issue #15918, opened 2024-10-15
Reporting rendering regression on PRs


mockersf:
Good news! Since #13248 (and #15894, #15911 and #15914) CI is now checking screenshots taken during example runs on macOS and Linux for changes.

The list of examples that are executed can be found there: https://github.com/bevyengine/bevy/tree/main/.github/example-run. At the moment, only 2 examples take a screenshot: `breakout` and `load_gltf`

The macOS run is done on every push to a PR, in the merge queue, and on main. You can find an example run there: https://github.com/bevyengine/bevy/actions/runs/11339649537/job/31534799604

The Linux run is done in the merge queue and on main. You can find an example run there: https://github.com/bevyengine/bevy/actions/runs/11339392720/job/31534437159

As you may have noticed, they both fail: this is because the example `load_gltf` is a tiny bit time dependent for the light movement, so not fully deterministic. Failures are currently not reported or blocking for PRs.

It is possible to make examples deterministic even when relying on time by applying the patches [there](https://github.com/bevyengine/bevy/tree/main/tools/example-showcase), or using the `example-showcase` tool.

You can see a list of run here: https://pixel-eagle.com/project/B04F67C0-C054-4A6F-92EC-F599FEC2FD1D

I have some improvements planned for the reporting, but now my question to you, dear contributors: which examples should be ran and checked for changes in CI?

This issue will be fixed by changing the list of screenshots compared to a fully deterministic one (either by selecting deterministic examples, or by enabling stable time with the patches), and changes being reported by a comment on PRs (not blocking merging yet)

ickshonpe:
Really like having this.

For UI the examples I think that would be most valuable to watch would be:
- `pbr`
- `borders`
- `ui`
- `text_wrap_debug`
- `grid`
- `overflow`
- `ui_texture_slice_flp_and_tile`
- `z_index`
- `box_shadow`

I could always add some extra elements to the `ui` example if that list is a bit long.

rparrett:
Do we have a target for the amount of examples we want to run?

It would be nice to have more "roll-up examples" like `ui` that show off many features of an engine area. Beyond `ui`, the following might be nice IMO:

- `lighting`
- `split_screen`
- `breakout`
- `animated_fox` (unfortunately this became non-deterministic at some point)

Maybe our game examples could be fleshed out to use more features, or some additional purpose-built examples could be made.

mockersf:
[...]

PR #15954, opened 2024-10-16
Introduce testbed examples starting with 2d

mockersf:
# Objective

- Make progress for #15918 
- Start with 2d

## Solution

- Remove screenshots for existing examples as they're not deterministic
- Create new "testbed" example category, with a 2d one to start

## Testing

- Run `CI_TESTING_CONFIG=.github/example-run/testbed_2d.ron cargo run --example testbed_2d --features "bevy_ci_testing"`
- ???
- Check the screenshots

### DIFF
--- a/examples/testbed/2d.rs
+++ b/examples/testbed/2d.rs
@@ -30,7 +30,7 @@ enum Scene {
 
 fn switch_scene(
     keyboard: Res<ButtonInput<KeyCode>>,
-    #[cfg(feature = "bevy_ci_testing")] mut ci_events: ResMut<Events<CiTestingCustomEvent>>,
+    #[cfg(feature = "bevy_ci_testing")] mut ci_events: EventReader<CiTestingCustomEvent>,
     scene: Res<State<Scene>>,
     mut next_scene: ResMut<NextState<Scene>>,
 ) {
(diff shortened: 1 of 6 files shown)


alice-i-cecile (approved):
I'm so pleased to have this started. An easy to use single testbed example (or three) will make testing changes so much easier. We should probably migrate our existing `ui` example over to this at some point.

examples/testbed/2d.rs
```diff
@@ -0,0 +1,172 @@
...
+    keyboard: Res<ButtonInput<KeyCode>>,
+    #[cfg(feature = "bevy_ci_testing")] mut ci_events: ResMut<Events<CiTestingCustomEvent>>,
+    scene: Res<State<Scene>>,
+    mut next_scene: ResMut<NextState<Scene>>,
+) {
+    let mut should_switch = false;
+    should_switch |= keyboard.just_pressed(KeyCode::Space);
+    #[cfg(feature = "bevy_ci_testing")]
+    {
+        should_switch |= !ci_events.is_empty();
```
BD103, line 41: I would prefer this actually filter for the `switch_scene` command, so it doesn't break when other custom commands are added in the future.

examples/testbed/2d.rs
```diff
@@ -0,0 +1,172 @@
...
+    }
+    if should_switch {
+        info!("Switching scene");
+        #[cfg(feature = "bevy_ci_testing")]
+        ci_events.clear();
```
BD103, line 46: If you switch to filtering, this should be removed.

BD103 (approved):
Looks good! I'd prefer to actually filter for the `switch_scene` command, but won't block on it.

Issue #15920, opened 2024-10-15
Buffer types + arrayLength() issues


JMS55:
## Bevy version
acbed6040eeae99199657e44a9668772738ac344

## Problem
StorageBuffer/UniformBuffer/etc will reuse its GPU buffer if it's big enough.

Say one frame you want a buffer of size N*2, and the next, N.

Because StorageBuffer/etc use as_entire_buffer_binding(), on the second frame, you'll end up with a buffer of size N*2, and _not_ N.

If the buffer stores an unsized array, any shaders using arrayLength() in the shader will return N*2, but again we'd be expecting N here as that's how many items we wrote this frame.

## Affected shaders
* mesh_preprocess.wgsl
* get_visibility_range_dither_level()
* Some meshlet stuff (I fixed this on a branch)

## Solution
The buffer types should probably not use as_entire_buffer_binding(), and instead should bind only the section of the buffer actually written to in the last write_buffer() call.

tychedelia:
I think the fix here is easy (use offset + len instead of as_entire_binding), but do we need to do this for `0.15` or is the ad hoc fix for meshlets sufficient.

JMS55:
I'd rather we do it for 0.15

alice-i-cecile:
I'm having trouble fully understanding the context here. What are the consequences here, how much code does this affect and why was this marked critical?

JMS55:
GPU-driven rendering and visibility ranges will break if the amount of meshes decreases between frames. Probably also 3rd party crates using our buffer abstractions + arrayLength().

alice-i-cecile:
Okay, that's definitely very bad. Thanks!

djeedai:
I'm having a hard time understanding why `as_entire_buffer_binding()` would return N/2 if the buffer was reallocated. By definition `as_entire_buffer_binding()` will return the entire buffer. So if it returns N/2 that means your buffer is of size N/2, and no amount of manually specifying a buffer size will change that. In fact manually specifying N when the buffer is only of size N/2 would be a bug. The only way I see that would be an issue would be if the binding returned by `as_entire_buffer_binding()` was cached, but that seems very unlikely given it has a lifetime bound.

JMS55:
The buffer is only reallocated if it's too small to fit the queued data (stored in a Vec). If the existing buffer is big enough, it's reused between writes. 

[...]

Issue #14681, opened 2024-08-09
Improve update_mesh_previous_global_transforms() performance


JMS55:
* Using try_insert for every entity is expensive. We should have a system to first add PreviousGlobalTransform to entities missing it, and then a second system to update the value. Or something like that.
* When using the meshlet feature, but no motion vector prepass, Mesh entities unnecessarily get a PreviousGlobalTransform. Not really worth fixing imo, as you probably want to use TAA with meshlets anyways, but I'm making a note of it here.

re0312:
I would like to complete this task, but I'm not sure if it will conflict with your current work.

JMS55:
Feel free to, it won't conflict with any active work that I know of. The only reason I haven't done it myself is that I don't have a computer to benchmark on atm.

JMS55:
Another option would be to include PreviousGlobalTransform in MeshMaterialBundle behind a cargo feature, instead of having a system to add it to entities that don't have it.

JMS55:
With required components, I'm tempted to make PreviousGlobalTransform required for MeshletMesh3d. How to solve this for the general case, idk yet.

JMS55:
> When using the meshlet feature, but no motion vector prepass, Mesh entities unnecessarily get a PreviousGlobalTransform. Not really worth fixing imo, as you probably want to use TAA with meshlets anyways, but I'm making a note of it here.

Don't know why I said this btw, PreviousGlobalTransform is _required_ for the meshlet culling to work correctly.

PR #15955, opened 2024-10-16
Meshlet fill cluster buffers rewritten

JMS55:
# Objective
- Make the meshlet fill cluster buffers pass slightly faster
- Address https://github.com/bevyengine/bevy/issues/15920 for meshlets
- Added PreviousGlobalTransform as a required meshlet component to avoid extra archetype moves, slightly alleviating https://github.com/bevyengine/bevy/issues/14681 for meshlets
- Enforce that MeshletPlugin::cluster_buffer_slots is not greater than 2^25 (glitches will occur otherwise). Technically this field controls post-lod/culling cluster count, and the issue is on pre-lod/culling cluster count, but it's still valid now, and in the future this will be more true.

Needs to be merged after https://github.com/bevyengine/bevy/pull/15846 and https://github.com/bevyengine/bevy/pull/15886

## Solution

- Old pass dispatched a thread per cluster, and did a binary search over the instances to find which instance the cluster belongs to, and what meshlet index within the instance it is.
- New pass dispatches a workgroup per instance, and has the workgroup loop over all meshlets in the instance in order to write out the cluster data.
- Use a push constant instead of arrayLength to fix the linked bug
- Remap 1d->2d dispatch for software raster only if actually needed to save on spawning excess workgroups

## Testing

  - Ran the meshlet example, and an example with 1041 instances of 32217 meshlets per instance. Profiled the second scene with nsight, went from 0.55ms -> 0.40ms. Small savings. We're pretty much VRAM bandwidth bound at this point.
  - Run the meshlet example

## Changelog (non-meshlets)
- PreviousGlobalTransform now implements the Default trait


alice-i-cecile:
This isn't a full fix for #15955, right? It only tackles the meshlet part of the problem.

JMS55:
Yeah. Only meshlets.

JMS55:
For further clarification, while working on this PR I discovered #15955. This PR fixes the bug for meshlets (by removing the code that caused it, as it's no longer needed). But we'll need a separate PR for an actual fix for the issue.

alice-i-cecile:
Perfect, thanks for the clarification. I've edited your PR description to make sure it doesn't accidentally get clobbered :)

atlv24 (approved):

cart (approved):
I've done a pass. I haven't fully grok-ed these changes, nor do I intend to at this point in the game. But they seem roughly shaped the right way and are scoped to the experimental meshlet module, so I'm cool with their inclusion in 0.15.

PR #15959, opened 2024-10-16
Fix AVIF image decoding

brianreavis:
# Objective / Solution

This expands upon:

- https://github.com/bevyengine/bevy/pull/15586

The image crate requires `avif-native` feature for decoding AVIF images ([documented here](https://github.com/image-rs/image?tab=readme-ov-file#supported-image-formats)). The `avif` flag only allows for the encoding of AVIFs.

## Testing

- Tested on a local project

### DIFF
--- a/crates/bevy_internal/Cargo.toml
+++ b/crates/bevy_internal/Cargo.toml
@@ -45,7 +45,7 @@ zlib = ["bevy_image/zlib"]
 zstd = ["bevy_image/zstd"]
 
 # Image format support (PNG enabled by default)
-avif = ["bevy_image/avif-native"]
+avif = ["bevy_image/avif"]
 bmp = ["bevy_image/bmp"]
 ff = ["bevy_image/ff"]
 gif = ["bevy_image/gif"]
--- a/crates/bevy_internal/Cargo.toml
+++ b/crates/bevy_internal/Cargo.toml
@@ -45,7 +45,7 @@ zlib = ["bevy_image/zlib"]
 zstd = ["bevy_image/zstd"]
 
 # Image format support (PNG enabled by default)
-avif = ["bevy_image/avif"]
+avif = ["bevy_image/avif-native"]
 bmp = ["bevy_image/bmp"]
 ff = ["bevy_image/ff"]
 gif = ["bevy_image/gif"]
--- a/crates/bevy_image/Cargo.toml
+++ b/crates/bevy_image/Cargo.toml
@@ -10,7 +10,7 @@ keywords = ["bevy"]
 
 [features]
 # Image formats
-avif = ["image/avif"]
+avif = ["image/avif-native"]
 basis-universal = ["dep:basis-universal"]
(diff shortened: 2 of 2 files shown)


crates/bevy_internal/Cargo.toml
```diff
@@ -45,7 +45,7 @@ zlib = ["bevy_image/zlib"]
 zstd = ["bevy_image/zstd"]
 
 # Image format support (PNG enabled by default)
-avif = ["bevy_image/avif"]
+avif = ["bevy_image/avif-native"]
```
alice-i-cecile, line 48:
```suggestion
avif = ["bevy_image/avif"]
```
We need to enable the `bevy_image` feature flag here, which is still just called `avif`.

alice-i-cecile (changes requested):
The feature flag for `bevy_image` is wrong :)

alice-i-cecile:
@clarfonthey can I get your review here?

brianreavis:
@alice-i-cecile You're quick! Just noticed that too and fixed it.

brianreavis:
One thing that needs to be figured out: the `ci` clippy job fails because it runs with `--all-features`:
https://github.com/bevyengine/bevy/actions/runs/11371344253/job/31633201423?pr=15959#step:6:876

We could install `dav1d`, but that seems overkill to make that particular job work.

mockersf:
AVIF seems to be mostly a format for animated images, which Bevy doesn't support built in.
I think the fix is to not expose it in Bevy, it will still be possible for a third party plugin to enable it.

brianreavis:
>  AVIF seems to be mostly a format for animated images, which Bevy doesn't support built in.

Though AVIF supports animation, it's great for compressing single images: https://jakearchibald.com/2020/avif-has-landed/

clarfonthey (approved):
I'm fine with this; it does look like the decoder is only gated on this feature, not the `avif` feature.

mockersf:
> the `ci` clippy job fails because it runs with `--all-features`

You can either remove built-in support for `AVIF`, which is what I would recommend, or add the `dav1d` library to the linux dependencies ([for ci](https://github.com/bevyengine/bevy/blob/main/.github/actions/install-linux-deps/action.yml) and [in documentation](https://github.com/bevyengine/bevy-website/blob/main/content/learn/book/getting-started/installation.md)).

Both options are possible, it depends how much you want built-in support for `AVIF`

clarfonthey:
I feel like it'd be best as a third option, opt for an embedded AVIF implementation, but in the absence of that, removing it feels better.

alice-i-cecile:
> I feel like it'd be best as a third option, opt for an embedded AVIF implementation, but in the absence of that, removing it feels better.

I don't think this is an important enough feature for that. I'd prefer to remove it and let third-party solutions cover this.

brianreavis:
I’m going to go ahead and close this PR and leave removal to another PR. If I get time I’ll take a stab at it.

PR #15075, opened 2024-09-07
Remove OrthographicProjection.scale (adopted)

richchurcher:
# Objective

Hello! I am adopting #11022 to resolve conflicts with `main`. tldr: this removes `scale` in favour of `scaling_mode`. Please see the original PR for explanation/discussion.

Also relates to #2580.

## Migration Guide

Replace all uses of `scale` with `scaling_mode`, keeping in mind that `scale` is (was) a multiplier. For example, replace
```rust
    scale: 2.0,
    scaling_mode: ScalingMode::FixedHorizontal(4.0),

```
with
```rust
    scaling_mode: ScalingMode::FixedHorizontal(8.0),
```

PR #15969, opened 2024-10-17
Improve API for scaling orthographic cameras

### LINKED ISSUE #15971: Vertically-centered text is too low
## Bevy version

0.15.0-dev, although this bug has been around a long time.

## What you did

In a flexbox or grid, place a text element with vertical centering.

## What went wrong

The text appears 1 or 2 pixels lower than it should, as seen in the following example screenshot:

Image

## Additional information

I had hoped that the migration to cosmic-text would fix this, but apparently not.

@UkoeHB

alice-i-cecile:
# Objective

Fixes #15971.

As raised in #11022, scaling orthographic cameras is confusing! In Bevy 0.14, there were multiple completely redundant ways to do this, and no clear guidance on which to use.

As a result, #15075 removed the `scale` field from `OrthographicProjection` completely, solving the redundancy issue.

However, this resulted in an unintuitive API and a painful migration, as discussed in #15971. Users simply want to change a single parameter to zoom, rather than deal with the irrelevant details of how the camera is being scaled.

## Solution

This PR reverts #15075, and takes an alternate, more nuanced approach to the redundancy problem. `ScalingMode::WindowSize` was by far the biggest offender. This was the default variant, and stored a float that was *fully* redundant to setting `scale`.

All of the other variants contained meaningful semantic information and had an intuitive scale. I could have made these unitless, storing an aspect ratio, but this would have been a worse API and resulted in a pointlessly painful migration.

In the course of this work I've also:

- improved the documentation to explain that you should just set `scale` to zoom cameras
- swapped to named fields for all of the variants in `ScalingMode` for more clarity about the parameter meanings
- substantially improved the `projection_zoom` example
- removed the footgunny Mul and Div impls for ScalingMode, especially since these no longer have the intended effect on `ScalingMode::WindowSize`.
- removed a rounding step because this is now redundant 🎉 

## Testing

I've tested these changes as part of my work in the `projection_zoom` example, and things seem to work fine.

## Migration Guide

`ScalingMode` has been refactored for clarity, especially on how to zoom orthographic cameras and their projections:

- `ScalingMode::WindowSize` no longer stores a float, and acts as if its value was 1. Multiply your camera's scale by any previous value to achieve identical results.
- `ScalingMode::FixedVertical` and `FixedHorizontal` now use named fields.

### DIFF
--- a/crates/bevy_render/src/camera/projection.rs
+++ b/crates/bevy_render/src/camera/projection.rs
@@ -1,7 +1,4 @@
-use core::{
-    marker::PhantomData,
-    ops::{Div, DivAssign, Mul, MulAssign},
-};
+use core::marker::PhantomData;
 
 use crate::{primitives::Frustum, view::VisibilitySystems};
 use bevy_app::{App, Plugin, PostStartup, PostUpdate};
@@ -269,6 +266,11 @@ impl Default for PerspectiveProjection {
 
 /// Scaling mode for [`OrthographicProjection`].
 ///
+/// The effect of these scaling modes are combined with the [`OrthographicProjection::scale`] property.
+///
+/// For example, if the scaling mode is `ScalingMode::Fixed{ width: 100.0, height: 300 }` and the scale is `2.0`,
+/// the projection will be 200 world units wide and 600 world units tall.
+///
 /// # Examples
 ///
 /// Configure the orthographic projection to two world units per window height:
@@ -280,15 +282,18 @@ impl Default for PerspectiveProjection {
 ///    ..OrthographicProjection::default_2d()
 /// });
 /// ```
-#[derive(Debug, Clone, Copy, Reflect, Serialize, Deserialize)]
+#[derive(Default, Debug, Clone, Copy, Reflect, Serialize, Deserialize)]
 #[reflect(Serialize, Deserialize)]
 pub enum ScalingMode {
-    /// Manually specify the projection's size, ignoring window resizing. The image will stretch.
-    /// Arguments are in world units.
-    Fixed { width: f32, height: f32 },
     /// Match the viewport size.
-    /// The argument is the number of pixels that equals one world unit.
-    WindowSize(f32),
+    ///
+    /// With a scale of 1, one world unit will be equal to the number of pixels in the window.
+    #[default]
+    WindowSize,
+    /// Manually specify the projection's size, ignoring window resizing. The image will stretch.
+    ///
+    /// Arguments are in world units when the scale is 1.
+    Fixed { width: f32, height: f32 },
     /// Keeping the aspect ratio while the axes can't be smaller than given minimum.
     /// Arguments are in world units.
     AutoMin { min_width: f32, min_height: f32 },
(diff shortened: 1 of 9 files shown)


BenjaminBrienen (approved):

crates/bevy_render/src/camera/projection.rs
```diff
@@ -280,81 +282,30 @@ impl Default for PerspectiveProjection {
 ///    ..OrthographicProjection::default_2d()
 /// });
 /// ```
-#[derive(Debug, Clone, Copy, Reflect, Serialize, Deserialize)]
+#[derive(Default, Debug, Clone, Copy, Reflect, Serialize, Deserialize)]
 #[reflect(Serialize, Deserialize)]
 pub enum ScalingMode {
+    /// Match the viewport size.
+    ///
+    /// With a scale of 1, one world unit will be equal to the number of pixels in the window.
```
MiniaczQ, line 290:
```suggestion
    /// With a scale of 1, one pixel will equate to one world unit.
```
The description doesn't make sense to me, how does length unit correspond to pixels in the window which are area?
Not sure if suggestion is correct either, do lmk

crates/bevy_render/src/camera/projection.rs
```diff
@@ -280,81 +282,30 @@ impl Default for PerspectiveProjection {
...
     Fixed { width: f32, height: f32 },
-    /// Match the viewport size.
-    /// The argument is the number of pixels that equals one world unit.
-    WindowSize(f32),
     /// Keeping the aspect ratio while the axes can't be smaller than given minimum.
     /// Arguments are in world units.
     AutoMin { min_width: f32, min_height: f32 },
     /// Keeping the aspect ratio while the axes can't be bigger than given maximum.
     /// Arguments are in world units.
     AutoMax { max_width: f32, max_height: f32 },
```
MiniaczQ, line 302:
```suggestion
    /// Keeping the aspect ratio while the axes can't be smaller than given minimum.
    /// Arguments are in world units and are affected by scale.
    AutoMin { min_width: f32, min_height: f32 },
    /// Keeping the aspect ratio while the axes can't be bigger than given maximum.
    /// Arguments are in world units and are affected by scale.
    AutoMax { max_width: f32, max_height: f32 },
```

MiniaczQ: Also, what's the padding behavior, black bars?

MiniaczQ (approved):
Just docs nits,
on the second though, maybe drop all the scale mentions on enum variant docs?
it's already explained in enum docs

alice-i-cecile:
@MiniaczQ I've improved the docs for ScalingMode to further elaborate on "what does a scale of 1" *mean*. This is an important point, so I've tried to be relatively exhaustive about all of the different factors that can affect things.

MiniaczQ:
> @MiniaczQ I've improved the docs for ScalingMode to further elaborate on "what does a scale of 1" _mean_. This is an important point, so I've tried to be relatively exhaustive about all of the different factors that can affect things.

Looks good :)

tim-blackbird (approved):

PR #15972, opened 2024-10-17
Double MAX_JOINTS and MAX_MORPH_WEIGHTS

victorb:
# Objective

I tried to import a 3D model that were slightly above the existing MAX_JOINTS and MAX_MORPH_WEIGHTS limits, got this error:

```
Failed to load asset 'models/humanoid.glb' with asset loader 'bevy_gltf::loader::GltfLoader':
failed to generate morph targets: Bevy only supports up to 64 morph targets (individual poses),
tried to create a model with 77 morph targets
```

## Solution

So tried the simplest solution possible, doubling the limits I hit when trying to load the model. It seems to work without any issues, everything still works exactly the same as far as I can tell. @alice-i-cecile asked me to PR the changes, so here we are :)

Ideally, I'd would have liked the comments for these two values to describe why those exact limits have been chosen, but I don't know the answer to that, maybe someone else could help me explain why the limits are there in the first place, and how the specific numbers were chosen.

### DIFF
--- a/crates/bevy_mesh/src/morph.rs
+++ b/crates/bevy_mesh/src/morph.rs
@@ -15,7 +15,7 @@ const MAX_TEXTURE_WIDTH: u32 = 2048;
 const MAX_COMPONENTS: u32 = MAX_TEXTURE_WIDTH * MAX_TEXTURE_WIDTH;
 
 /// Max target count available for [morph targets](MorphWeights).
-pub const MAX_MORPH_WEIGHTS: usize = 64;
+pub const MAX_MORPH_WEIGHTS: usize = 128;
 
 #[derive(Error, Display, Clone, Debug)]
 pub enum MorphBuildError {
--- a/crates/bevy_pbr/src/render/skin.rs
+++ b/crates/bevy_pbr/src/render/skin.rs
@@ -15,7 +15,7 @@ use bevy_render::{
 use bevy_transform::prelude::GlobalTransform;
 
 /// Maximum number of joints supported for skinned meshes.
-pub const MAX_JOINTS: usize = 256;
+pub const MAX_JOINTS: usize = 512;
 
 #[derive(Component)]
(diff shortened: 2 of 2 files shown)


alice-i-cecile:
Do you have a convenient link to a model that previously failed but now works that we can use to test this?

alice-i-cecile:
This has been moved several times, and I'm trying to piece together the history. It was moved to `bevy_mesh`, and before that split off of `mesh.rs` in https://github.com/bevyengine/bevy/commit/db1e3d36bc8ebdcdf85b17eecca2ecf022dabd78

victorb:
@alice-i-cecile 
> Do you have a convenient link to a model that previously failed but now works that we can use to test this?

Unfortunately not, it's a purchased & proprietary model and license doesn't allow me to redistribute it like that.

I'm wondering if maybe just the weight/joint data however could be extracted and used for a test case, would just be data and should be fine in theory. If this could work for you/Bevy and be beneficial, I can reach out to the original model author and ask if it's OK with them.

mockersf (changes requested):
This values are used to allocate buffers. Their maximum sizes depend on the GPU/platform, the current values are the ones guaranteed to work everywhere.

To allow bigger values, you would have to check against the limits of the GPU at runtime, which would mean not using consts anymore.

victorb:
> To allow bigger values, you would have to check against the limits of the GPU at runtime, which would mean not using consts anymore.

How would that realistically work like? I'm guessing we can't dynamically figure that out, as that would have to take into consideration other things that go into the GPU, or not? Or would it just be a question of reading a value then setting the limits based on that?

Alternatively, would it make sense to have the default values as they were before this patch, and then allow users to set the limit themselves via passed in values? Would be a workaround at best I suppose.

victorb:
Closing this in favor of https://github.com/bevyengine/bevy/issues/17128

PR #15973, opened 2024-10-17
Remove AVIF feature

### LINKED ISSUE #15968: Remove broken AVIF support from `bevy_image`
> I feel like it'd be best as a third option, opt for an embedded AVIF implementation, but in the absence of that, removing it feels better.

I don't think this is an important enough feature for that. I'd prefer to remove it and let third-party solutions cover this.

_Originally posted by @alice-i-cecile in https://github.com/bevyengine/bevy/issues/15959#issuecomment-2418310800_

clarfonthey:
Resolves #15968. Since this feature never worked, and enabling it in the `image` crate requires system dependencies, we've decided that it's best to just remove it and let other plugin crates offer support for it as needed.

## Migration Guide

AVIF images are no longer supported. They never really worked, and require system dependencies (libdav1d) to work correctly, so, it's better to simply offer this support via an unofficial plugin instead as needed. The corresponding types have been removed from Bevy to account for this.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -247,9 +247,6 @@ trace_tracy_memory = [
 # Tracing support
 trace = ["bevy_internal/trace"]
 
-# AVIF image format support
-avif = ["bevy_internal/avif"]
-
 # Basis Universal compressed texture support
 basis-universal = ["bevy_internal/basis-universal"]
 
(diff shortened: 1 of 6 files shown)


Elabajaba (approved):
LGTM

BenjaminBrienen (approved):

mockersf (approved):

PR #15975, opened 2024-10-18
Merge Style properties into Node. Use ComputedNode for computed properties.

cart:
# Objective

Continue improving the user experience of our UI Node API in the direction specified by [Bevy's Next Generation Scene / UI System](https://github.com/bevyengine/bevy/discussions/14437)

## Solution

As specified in the document above, merge `Style` fields into `Node`, and move "computed Node fields" into `ComputedNode` (I chose this name over something like `ComputedNodeLayout` because it currently contains more than just layout info. If we want to break this up / rename these concepts, lets do that in a separate PR). `Style` has been removed.

This accomplishes a number of goals:

## Ergonomics wins

Specifying both `Node` and `Style` is now no longer required for non-default styles

Before:
```rust
commands.spawn((
    Node::default(),
    Style {
        width:  Val::Px(100.),
        ..default()
    },
));
```

After:

```rust
commands.spawn(Node {
    width:  Val::Px(100.),
    ..default()
});
```

## Conceptual clarity

`Style` was never a comprehensive "style sheet". It only defined "core" style properties that all `Nodes` shared. Any "styled property" that couldn't fit that mold had to be in a separate component. A "real" style system would style properties _across_ components (`Node`, `Button`, etc). We have plans to build a true style system (see the doc linked above).

By moving the `Style` fields to `Node`, we fully embrace `Node` as the driving concept and remove the "style system" confusion.

## Next Steps

* Consider identifying and splitting out "style properties that aren't core to Node". This should not happen for Bevy 0.15.

---

## Migration Guide

Move any fields set on `Style` into `Node` and replace all `Style` component usage with `Node`.

Before:
```rust
commands.spawn((
    Node::default(),
    Style {
        width:  Val::Px(100.),
        ..default()
    },
));
```

After:

```rust
commands.spawn(Node {
    width:  Val::Px(100.),
    ..default()
});
```

For any usage of the "computed node properties" that used to live on `Node`, use `ComputedNode` instead:

Before:
```rust
fn system(nodes: Query<&Node>) {
    for node in &nodes {
        let computed_size = node.size();
    }
}
```

After:
```rust
fn system(computed_nodes: Query<&ComputedNode>) {
    for computed_node in &computed_nodes {
        let computed_size = computed_node.size();
    }
}
```

### DIFF
--- a/crates/bevy_ui/src/layout/mod.rs
+++ b/crates/bevy_ui/src/layout/mod.rs
@@ -1,7 +1,7 @@
 use crate::{
     experimental::{UiChildren, UiRootNodes},
-    BorderRadius, ContentSize, DefaultUiCamera, Display, Node, Outline, OverflowAxis,
-    ScrollPosition, Style, TargetCamera, UiScale,
+    BorderRadius, ComputedNode, ContentSize, DefaultUiCamera, Display, Node, Outline, OverflowAxis,
+    ScrollPosition, TargetCamera, UiScale,
 };
 use bevy_ecs::{
     change_detection::{DetectChanges, DetectChangesMut},
@@ -115,22 +115,19 @@ pub fn ui_layout_system(
     mut resize_events: EventReader<bevy_window::WindowResized>,
     mut ui_surface: ResMut<UiSurface>,
     root_nodes: UiRootNodes,
-    mut style_query: Query<
-        (
-            Entity,
-            Ref<Style>,
-            Option<&mut ContentSize>,
-            Option<&TargetCamera>,
-        ),
-        With<Node>,
-    >,
-    node_query: Query<(Entity, Option<Ref<Parent>>), With<Node>>,
+    mut node_query: Query<(
+        Entity,
+        Ref<Node>,
+        Option<&mut ContentSize>,
+        Option<&TargetCamera>,
+    )>,
+    computed_node_query: Query<(Entity, Option<Ref<Parent>>), With<ComputedNode>>,
     ui_children: UiChildren,
     mut removed_components: UiLayoutSystemRemovedComponentParam,
     mut node_transform_query: Query<(
-        &mut Node,
+        &mut ComputedNode,
         &mut Transform,
-        &Style,
+        &Node,
         Option<&BorderRadius>,
         Option<&Outline>,
         Option<&ScrollPosition>,
@@ -173,7 +170,7 @@ pub fn ui_layout_system(
     // Precalculate the layout info for each camera, so we have fast access to it for each node
     camera_layout_info.clear();
 
-    style_query
+    node_query
         .iter_many(root_nodes.iter())
         .for_each(|(entity, _, _, target_camera)| {
             match camera_with_default(target_camera) {
@@ -211,8 +208,8 @@ pub fn ui_layout_system(
         ui_surface.try_remove_node_context(entity);
     }
 
-    // Sync Style and ContentSize to Taffy for all nodes
-    style_query
+    // Sync Node and ContentSize to Taffy for all nodes
+    node_query
         .iter_mut()
         .for_each(|(entity, style, content_size, target_camera)| {
             if let Some(camera) =
@@ -235,7 +232,7 @@ pub fn ui_layout_system(
                     ui_surface.upsert_node(&layout_context, entity, &style, measure);
                 }
             } else {
-                ui_surface.upsert_node(&LayoutContext::DEFAULT, entity, &Style::default(), None);
+                ui_surface.upsert_node(&LayoutContext::DEFAULT, entity, &Node::default(), None);
             }
         });
     scale_factor_events.clear();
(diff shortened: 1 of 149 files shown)


cart:
In #15898 I said this should wait until after Bevy 0.15. I've re-evaluated the cost:benefit of doing this now and I think we should merge this for Bevy 0.15. I was wrong. The wins here are big and the cost of porting is lower than I expected.

I had it in my head that we needed to sort out the more controversial "Style breakup" problems as part of this, but I don't really see the need. Doing this now actually makes a future breakup easier as it can be framed as "just" moving fields out of `Node`.

alice-i-cecile:
I'm on board with doing this now :) I don't think this is the end game, but it's much nicer as an interim solution.

examples/games/game_menu.rs
```diff
@@ -377,15 +374,15 @@ mod menu {
 
     fn main_menu_setup(mut commands: Commands, asset_server: Res<AssetServer>) {
         // Common style for all buttons on the screen
-        let button_style = Style {
```
VitalyAnkh, line 380: Also variable rename here.

examples/games/game_menu.rs
```diff
@@ -377,15 +374,15 @@ mod menu {
...
+        let button_style = Node {
             width: Val::Px(300.0),
             height: Val::Px(65.0),
             margin: UiRect::all(Val::Px(20.0)),
             justify_content: JustifyContent::Center,
             align_items: AlignItems::Center,
             ..default()
         };
-        let button_icon_style = Style {
+        let button_icon_style = Node {
```
VitalyAnkh, line 385: Also variable rename here.

VitalyAnkh (changes requested):

alice-i-cecile (approved):
The semantics are much clearer, I'm extremely glad to have cleared up the Style name, and the reduced boilerplate is nice.

I don't think the migration will be hard at all: rename uses of `Node` to `ComputedNode`, then rename uses of `Style` to `Node`.

UkoeHB (approved):

crates/bevy_ui/src/node_bundles.rs
```diff
@@ -93,7 +93,7 @@ pub struct ImageBundle {
     /// The transform of the node
     ///
     /// This component is automatically managed by the UI layout system.
-    /// To alter the position of the `ImageBundle`, use the properties of the [`Style`] component.
+    /// To alter the position of the `ImageBundle`, use the properties of the [`Node`] component.
```
ickshonpe, line 96: There need to be stronger warnings even to users against modifying the `Transform` of UI nodes but there doesn't seem to be a natural place anymore with required components for context specific documentation like we could have in bundles. Maybe on the require?
```
#[require(
    Node,
    // A Node's Transform is automatically managed by the UI layout system, manually changing its values should be avoided
    Transform,
     // .. rest of required components
)]
```
It doesn't seem great though and wouldn't show up in doc comments, unless something could be done with the require macro?

crates/bevy_ui/src/update.rs
```diff
@@ -37,7 +42,12 @@ pub fn update_clipping_system(
...
-    node_query: &mut Query<(&Node, &GlobalTransform, &Style, Option<&mut CalculatedClip>)>,
+    node_query: &mut Query<(
+        &ComputedNode,
+        &GlobalTransform,
+        &Node,
+        Option<&mut CalculatedClip>,
+    )>,
     entity: Entity,
     mut maybe_inherited_clip: Option<Rect>,
 ) {
```
tim-blackbird, line 53: There is a variable still called `style` just below this comment, I can't comment on the line itself unfortunately

crates/bevy_ui/src/layout/mod.rs
```diff
@@ -211,8 +208,8 @@ pub fn ui_layout_system(
         ui_surface.try_remove_node_context(entity);
     }
 
-    // Sync Style and ContentSize to Taffy for all nodes
-    style_query
+    // Sync Node and ContentSize to Taffy for all nodes
+    node_query
         .iter_mut()
         .for_each(|(entity, style, content_size, target_camera)| {
```
tim-blackbird, line 214: Another variable still called `style`

crates/bevy_ui/src/layout/mod.rs
```diff
@@ -319,9 +318,9 @@ with UI components as a child of an entity without UI components, your UI layout
         ui_surface: &UiSurface,
         root_size: Option<Vec2>,
         node_transform_query: &mut Query<(
```
tim-blackbird, line 320: Another variable still called `style` where this query is used

examples/ui/display_and_visibility.rs
```diff
@@ -48,7 +48,7 @@ trait TargetUpdate {
 }
 
 impl TargetUpdate for Target<Display> {
-    type TargetComponent = Style;
+    type TargetComponent = Node;
     const NAME: &'static str = "Display";
     fn update_target(&self, style: &mut Self::TargetComponent) -> String {
```
tim-blackbird, line 53: Another variable still called `style`

crates/bevy_ui/src/ui_node.rs
```diff
@@ -446,8 +446,8 @@ pub struct Style {
     ///
     /// # Example
     /// ```
-    /// # use bevy_ui::{Style, UiRect, Val};
-    /// let style = Style {
+    /// # use bevy_ui::{Node, UiRect, Val};
+    /// let style = Node {
```
tim-blackbird, line 450: Another variable still called `style`

ickshonpe (approved):
I've been unreasonably miserable all week about the previous required components changes so really pleased with this change, it fixes everything I was unhappy about before.

tim-blackbird (approved):
LGTM, just some missed variable names left

You may want to adjust [these docs](https://github.com/cart/bevy/blob/af0149d2d18751bdfa4ec8fe442d4eba74a4460f/crates/bevy_ecs/src/query/filter.rs#L357) as well

crates/bevy_ui/src/node_bundles.rs
ickshonpe: It's outside the scope of this PR though, can be discussed later.

ickshonpe (approved):

cart:
Ah wait missed a couple of non-code comment things.

Issue #15255, opened 2024-09-17
Extrudable not implemented for Polygons


AnjanaYEAH:
Hi, 

Beginner Rust and Bevy developer here!

I don't see why RegularPolygons are extrudable and Polygons are not extrudable https://docs.rs/bevy/latest/bevy/render/mesh/trait.Extrudable.html

The code below does not work with the following error message:
```
the trait From<bevy::prelude::Extrusion<bevy::prelude::Polygon<_>>> is not implemented for bevy::prelude::Mesh, which is required by bevy::prelude::Extrusion<bevy::prelude::Polygon<_>>: Into<bevy::prelude::Mesh>
```

```rs
fn spawn_polygon(
    mut commands: Commands,
    mut materials: ResMut<Assets<StandardMaterial>>,
    mut meshes: ResMut<Assets<Mesh>>,
) {
    let points2d = vec!
        [
            vec2(-6., 2.),
            vec2(-2., 4.),
            vec2(-3., 5.)
        ];
    let p = Polygon::new(points2d);
    let e = Extrusion::new(p, 2.0);
    let mesh_handle: Handle<Mesh> = meshes.add(e);

    commands.spawn((
        PbrBundle {
            mesh: mesh_handle,
            material: materials.add(StandardMaterial {
               base_color: Srgba::hex("#ffd891").unwrap().into(), 
                ..default()
            }),
            ..default()
        },
        Cube,
    ));
}
```

I want to be able to create custom polygons and extrude them, not sure why this is not possible at the moment?

IQuick143:
I believe the issue is related to us not even implementing Meshing for Polygons, the reason being you can have Polygons represent various ugly concave or even self-intersecting shapes.

Extrusion is one step after that.

Meshing for concave polygons is not an impossible problem, I could draft up some algorithm, although I give no guarantees about how the topology of the mesh will look.

As for self-intersecting polygons... idk.

lynn-lumen:
If we are fine with adding another dependency, we could give libraries like [earcutr](https://github.com/donbright/earcutr) a try. In my experience, they work very well :)

IQuick143:
I'm not sure if we should introduce a dependency for such a relatively niche feature, but the ear-clipping algorithm looks pretty much exactly like what we want if we implement this.

alice-i-cecile:
Yeah, I'm loathe to add a dependency for a very optional feature.

lynn-lumen:
[...]

PR #15981, opened 2024-10-18
Polygon simplicity

lynn-lumen:
# Objective

- This PR adds the ability to determine whether a `Polygon<N>` or `BoxedPolygon` is simple (aka. not self-intersecting) by calling `my_polygon.is_simple()`.
- This may be useful information for users to determine whether their polygons are 'valid' and will be useful when adding meshing for polygons.
  - As such this is a step towards fixing #15255

## Solution

- Implemented the Shamos-Hoey algorithm in its own module `polygon`.

## Testing

- Tests are included, and can be verified visually.

---

## Performance

- The Shamos-Hoey algorithm runs in O(n * log n)
- In reality, the results look more linear to me.
- Determining simplicity for a simple polygon (the worst case) with less than 100 vertices takes less than 0.2ms.

image

### DIFF
--- a/crates/bevy_math/src/primitives/polygon.rs
+++ b/crates/bevy_math/src/primitives/polygon.rs
@@ -154,7 +154,7 @@ impl<'a> SweepLine<'a> {
         }
     }
 
-    /// Determine whther the given edges of the polygon intersect.
+    /// Determine whether the given edges of the polygon intersect.
     fn intersects(&self, edge1: Option<usize>, edge2: Option<usize>) -> bool {
         let Some(edge1) = edge1 else {
             return false;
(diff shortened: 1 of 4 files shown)


alice-i-cecile (approved):

crates/bevy_math/src/primitives/polygon.rs
```diff
@@ -0,0 +1,340 @@
...
+    Left,
+    Right,
+}
+
+/// An event in the [`EventQueue`] is either the left or right vertex of an edge of the polygon.
+///
+/// Events are ordered so that any event `e1` which is to the left of another event `e2` is less than that event.
+/// If `e1.position().x == e2.position().x` the events are ordered from bottom to top.
+#[derive(Debug, Clone, Copy)]
+struct Event {
```
IQuick143, line 20: I know this is not exposed outside the module, but I feel that even inside the module this would deserve a smaller scope, or a more concrete name than just "Event".

lynn-lumen: Something like `SweepLineEvent`?

BenjaminBrienen:
O(n*log_9999999(n)) I guess

BenjaminBrienen (approved):

alice-i-cecile (approved):

pcwalton:
I think this may have broken the test that checks to ensure that `bevy_math` compiles with `libcore`.

Issue #15893, opened 2024-10-14
Spawning or modifying many different 2d or 3d materials hangs for minutes or crashes


DGriffin91:
## Bevy version 89e19aaff0082a9d353d3130f3d79e85ebb122c0

The `many_cubes` example with `cargo run --example many_cubes --release -- --vary-material-data-per-instance` hangs indefinitely (Update: Tried just letting this run and after a little over 2 minutes the example started working).

This regression also affects modifying materials at run time. See example: https://github.com/bevyengine/bevy/issues/15893#issuecomment-2421224174

Windows 10 / RTX3060 / Vulkan

The issue was introduced at https://github.com/bevyengine/bevy/commit/7b81ae7e406e61b108accb30352d67304ff6f044 with [Update WGPU to version 22](https://github.com/bevyengine/bevy/pull/14401)

Apple M1 / Metal: Hangs for 4 minutes
Win10 / GTX1060 / Vulkan / i7 6700k: Hangs for 12 minutes
Win10 / RTX3060 / Vulkan / 7950x: Hangs for 2 minutes
Win10 / RTX3060 / Dx12 / 7950x: Crashes (Note Dx12 also crashes in 0.14)
```
2024-10-14T19:44:43.028155Z ERROR wgpu_hal::dx12::descriptor: Unable to allocate descriptors: RangeAllocationError { fragmented_free_length: 1 }
2024-10-14T19:44:43.028302Z ERROR wgpu::backend::wgpu_core: Handling wgpu errors as fatal by default
thread 'main' panicked at \.cargo\registry\src\index.crates.io-6f17d22bba15001f\wgpu-0.20.1\src\backend\wgpu_core.rs:2996:5:
wgpu error: Validation Error
Caused by:
    In Device::create_bind_group
      note: label = `StandardMaterial`
    Not enough memory left.
```

Minimal-ish 3d example:
```rs
use bevy::{diagnostic::*, prelude::*};
fn main() {
    App::new()
        .add_plugins((
            DefaultPlugins,
            FrameTimeDiagnosticsPlugin,
            LogDiagnosticsPlugin::default(),
        ))
        .add_systems(Startup, setup)
        .run();
}
fn setup(
    mut commands: Commands,
    mut meshes: ResMut<Assets<Mesh>>,
    mut materials: ResMut<Assets<StandardMaterial>>,
) {
    let mesh = Mesh3d(meshes.add(Cuboid::new(1.0, 1.0, 1.0)));
[...]

PR #15988, opened 2024-10-18
Upgrade to wgpu 23

tychedelia:
Fixes https://github.com/bevyengine/bevy/issues/15893


crates/bevy_color/Cargo.toml
```diff
@@ -23,7 +23,7 @@ derive_more = { version = "1", default-features = false, features = [
   "from",
   "display",
 ] }
-wgpu-types = { version = "22", default-features = false, optional = true }
+wgpu-types = { git = "https://github.com/gfx-rs/wgpu", branch = "trunk", default-features = false, optional = true }
```
djeedai, line 26: It seems dangerous to move to a trunk branch no? Shouldn't we specify a locked commit hash instead? (even if it's the same as v23 at the time of writing)

djeedai (commented):
Looks good to merge as long as we're OK to take a dependency on a moving target (I don't think we should). Also maybe the danger of upgrading to a new version close to release (but I understand it fixes some bug).

crates/bevy_color/Cargo.toml
rparrett: wgpu 0.23 isn't released yet, presumably the author intends to update this and take the PR out of draft status once it is released.

JMS55:
No need to wrap bind groups in Some() anymore https://github.com/gfx-rs/wgpu/pull/6452

JMS55:
Actually that makes it worse... Can't automatically convert between bevy's and wgpu's bind group type...
```
note: required by a bound in `wgpu::RenderPass::<'encoder>::set_bind_group`
   --> C:\Users\Jasmine\.cargo\git\checkouts\wgpu-ebe3fb3e55696655\a9033f5\wgpu\src\api\render_pass.rs:80:26
    |
77  |     pub fn set_bind_group<'a>(
    |            -------------- required by a bound in this associated function
...
80  |         bind_group: impl Into<Option<&'a BindGroup>>,
    |                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^ required by this bound in `RenderPass::<'encoder>::set_bind_group`

error[E0277]: the trait bound `std::option::Option<&wgpu::BindGroup>: From<&bind_group::BindGroup>` is not satisfied
   --> crates\bevy_render\src\view\window\screenshot.rs:626:36
    |
626 |             pass.set_bind_group(0, &prepared_state.bind_group, &[]);
    |                  --------------    ^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `From<&bind_group::BindGroup>` is not implemented for `std::option::Option<&wgpu::BindGroup>`, which is required by `&bind_group::BindGroup: Into<std::option::Option<&wgpu::BindGroup>>`
    ```

AlexAegis:
https://github.com/gfx-rs/wgpu/pull/6465

Looks like wgpu 23 is about to release! 🎉

Elabajaba:
This doesn't build on iOS currently according to Francois' testing.

mockersf:
Just re-triggered a build, here is the failure: https://github.com/TheBevyFlock/bevy-example-runner/actions/runs/11533949819/job/32107445980

tychedelia:
> Just re-triggered a build, here is the failure: https://github.com/TheBevyFlock/bevy-example-runner/actions/runs/11533949819/job/32107445980

From last month:

https://github.com/gfx-rs/wgpu/commit/fb0cb1eb11663f8de023f6dd64128ed1f5342ec7#diff-46fec780a1b7071adf9590bb9d55dfea0f97bd15419ccbe55ee89607fc244335R27

BenjaminBrienen:
With `cargo run --example relative_cursor` I have to set the `web-sys` dependency to `0.3.72` and then it compiles.

mockersf:
> Edit: pushed a commit to attempt to fix this.

Building now works 🎉

JMS55:
Might be worth letting wgpu people know that they should include it in the changelog for wgpu 23.

mockersf:
released! https://github.com/gfx-rs/wgpu/releases/tag/v23.0.0

JMS55:
Something we can finally now do is use const arrays with dynamic indexes, for e.g. shadow sampling patterns. 

We should take advantage of this to cleanup a bunch of code.

djeedai:
FYI this also brings CAS atomic operations to Metal. I didn't check if that was the last platform missing them (I expect wasm needs CORS, as with everything atomic due to security reasons, if it even supports CAS at the API level already).

mockersf:
@tychedelia naga_oil 0.16 has been updated and released

kristoff3r (approved):
LGTM, I also tested a bunch of 2d/3d examples on Linux and WebGL and stuff seems to work.

mockersf (approved):

mockersf:
I would like to merge this one quickly and publish a new rc soon after, so get your reviews now! well in the next day or so

JMS55 (approved):

PR #15991, opened 2024-10-18
Fix OIT depth test

IceSentry:
# Objective

- The depth test was only using the final depth but it should be testing each fragments 
- Fully transparent fragments should not be added to the list

## Solution

- Test each fragment after sorting

## Testing

before:
image

after:
image

### DIFF
--- a/crates/bevy_core_pipeline/src/oit/oit_draw.wgsl
+++ b/crates/bevy_core_pipeline/src/oit/oit_draw.wgsl
@@ -5,6 +5,12 @@
 #ifdef OIT_ENABLED
 // Add the fragment to the oit buffer
 fn oit_draw(position: vec4f, color: vec4f) {
+    // Don't add fully transparent fragments to the list
+    // because we don't want to have to sort them in the resolve pass
+    // TODO should this be comparing with < espilon ?
+    if color.a == 0.0 {
+        return;
+    }
     // get the index of the current fragment relative to the screen size
     let screen_index = i32(floor(position.x) + floor(position.y) * view.viewport.z);
     // get the size of the buffer.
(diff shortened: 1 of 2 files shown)


crates/bevy_core_pipeline/src/oit/oit_draw.wgsl
```diff
@@ -5,6 +5,12 @@
 #ifdef OIT_ENABLED
 // Add the fragment to the oit buffer
 fn oit_draw(position: vec4f, color: vec4f) {
+    // Don't add fully transparent fragments to the list
+    // because we don't want to have to sort them in the resolve pass
+    // TODO should this be comparing with < espilon ?
```
hexroll, line 10: I suppose this could be something like:
```rust
const EPSILON: f32 = 4.88e-04;
if color.a < EPSILON { ... }
```
But tbh I'm not seeing that this is done when comparing color values anywhere in the codebase.

hexroll (approved):
LGTM.
I also tested this with two scenes and it's working as expected \o/

tychedelia (approved):
LGTM, might be worth filing a ticket so that we don't lose track unless this is something you think you'll get to for work anyway.

crates/bevy_core_pipeline/src/oit/oit_draw.wgsl
IceSentry: The idea behind that comment is more about having a cutoff for almost transparent pixels. Instead of fully transparent, the reason is that every fragment we add to the list costs us more memory. Right now it doesn't really matter because we have a fixed buffer size, but eventually I want to work on making it smaller.

IceSentry:
All the various TODOs in the OIT code are very much for future me when I'll work on it. Some of the improvements I need for work but some are a bevy specific.

cart (approved):

PR #15994, opened 2024-10-18
Refactor `bevy_gltf`

hukasu:
# Objective

Refactor `bevy_gltf`, the criteria for the split is kind of arbitrary but at least it is not a 2.6k line file.

## Solution

Move methods and structs found in `bevy_gltf/loader.rs` into multiple new modules.

## Testing

`cargo run -p ci`


suprohub:
any updates?

hukasu:
The team is focused on 0.15.1, i'm in wait until 0.16 becames the target, then i will merge and do the comments

hukasu:
@alice-i-cecile can we get this moving now? who would be SME on this topic to do the reviews?

thepackett (commented):
Overall this seems like a good PR. Having read through the GltfLoader code before, this is a significant improvement. I ran CI locally successfully and was able to confirm that examples using Gltf assets were working as expected. The only question I have is whether or not the extension traits should be public or not.

crates/bevy_gltf/src/ext/document.rs
```diff
@@ -0,0 +1,51 @@
+use bevy_platform_support::collections::HashSet;
+
+use gltf::Document;
+
+use super::material::MaterialExt;
+
+pub trait DocumentExt {
```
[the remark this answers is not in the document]
hukasu: they are visible only inside the crate because the modules are not public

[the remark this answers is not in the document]
hukasu: if you go to `lib.rs` you will see that all modules are private, and a bit below it there is the re-exports of what should be visible by users of the crate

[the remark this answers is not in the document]
hukasu: what is visible by users remains the same as before the refactor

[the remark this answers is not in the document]
thepackett: Yep, you are correct. In that case I can't see any issues with this PR as written.

thepackett (approved):

andriyDev (changes requested):
I am not a fan of how many extension traits there are. This doesn't seem necessary, and results in a bunch of tiny files which I hurts readability IMO. For example `crates/bevy_gltf/src/ext/scene.rs` is a whopping 14 lines. Jumping to this file just to see "oh this is just an extension trait" is quite underwhelming.

I would prefer just free functions for most of these (like the original implementation), just organized into modules more. For example, rather than `node.node_name()`, we just have `node_name(&node)` and `node_transform(&node)` and so on. But with the important part being that `node_name` and `node_transform` should be in a `mod node`.

One case where "extension traits" is likely desirable is labels - these could all be a single `trait ToLabel` which Scene, Skin, Texture, etc can implement. Now all these disparate label stuffs can be brought under one roof and we can see how all of these are really the same thing over and over again. Of course whether this really is the way we want to go I'm not sure (for example Skin already breaks this pattern since it wants a `inverse_bind_matrics_label` version - though this could just sit next to the trait implementations as `inverse_bind_matrics_label_for_skin` to again keep the similar implementations together.

hukasu:
@andriyDev see if this solution is satisfactory

rparrett (commented):
I added some specific suggestions about the `ext` module, though I think I generally agree that this is feeling excessively "enterprise-grade."

I don't really agree with the way the changeset is presented. The current arrangement of "the loader stuff is in loader.rs" is logical and not inherently worse than an "arbitrary" split.

crates/bevy_gltf/src/label.rs
```diff
@@ -0,0 +1,165 @@
...
+            GltfAssetLabel::Skin(skin.index())
+        }
+    }
+}
+
+impl From<&gltf::Texture<'_>> for GltfAssetLabel {
+    fn from(texture: &gltf::Texture) -> GltfAssetLabel {
+        GltfAssetLabel::Texture(texture.index())
+    }
+}
```
andriyDev, line 165: Ah I didn't realize how some of these need an "inverse" flag. Sorry about that. I don't think we should go with this. Seeing something like `(material, true).into()` would be very confusing. It's not clear what's going on, what `true` corresponds to, etc. I think just making these regular free functions like `material_to_label(material: &gltf::Material, is_scale_inverted: bool)` is straight forward and clear.

crates/bevy_gltf/src/loader/gltf_ext/mesh.rs
```diff
@@ -0,0 +1,42 @@
+use bevy_render::mesh::PrimitiveTopology;
+
+use gltf::mesh::{Mesh, Mode, Primitive};
+
+use crate::GltfError;
+
+pub trait MeshExt {
```
andriyDev, line 7: As I said before, I think we should just avoid these extension traits. I'd much prefer just free functions, like `primitive_name_for_mesh(mesh: &gltf::Mesh, primitive: &Primitive)`. (Similar for the other extension traits)

andriyDev (changes requested):

crates/bevy_gltf/src/loader/gltf_ext/extras.rs
```diff
@@ -1,13 +1,9 @@
...
-}
+use crate::GltfExtras;
 
-impl ExtrasExt for gltf::json::Extras {
-    fn as_gltf_extras(&self) -> Option<GltfExtras> {
-        self.as_ref().map(|extras| GltfExtras {
-            value: extras.get().to_string(),
-        })
-    }
+pub fn as_gltf_extras(extras: &Extras) -> Option<GltfExtras> {
```
andriyDev, line 5: Can we make basically everything in `crates/bevy_gltf/src/loader/gltf_ext` and `crates/bevy_gltf/src/loader/extensions` `pub(crate)` instead (where it makes sense of course)? I know these aren't leaking out right now, but I'd rather make it clear that we want these to only be implementation details.

andriyDev (changes requested):
This is looking great!

crates/bevy_gltf/src/loader/gltf_ext/extras.rs
hukasu: sure

mockersf:
This feels to me as too many small files, not sure that's better than the current state.

crates/bevy_gltf/src/loader/gltf_ext/extras.rs
hukasu: is it enough like this?
```rust
pub(super) mod extras;
pub(super) mod material;
pub(super) mod mesh;
pub(super) mod scene;
pub(super) mod skin;
pub(super) mod texture;
```
or is it better on each individual struct?

andriyDev: It's better on each individual struct. Making it clear that these are not our public API to readers of the code is useful IMO. Plus it prevents us from accidentally exposing these symbols even if we make a mod public for some reason.

hukasu: understood

hukasu: or would it be ok to have a module doc (i.e. `//!`) that says that all items on that file are internal to `loader.rs`?

andriyDev: Doc comments can go stale. Visibility keywords can't. So I'd prefer we make it explicit and just say `pub(crate)`.

hukasu: ok, coming up in a few seconds

crates/bevy_gltf/src/loader/extensions/khr_materials_anisotropy.rs
```diff
@@ -0,0 +1,71 @@
...
+impl AnisotropyExtension {
+    #[expect(
+        clippy::allow_attributes,
+        reason = "`unused_variables` is not always linted"
+    )]
+    #[allow(
+        unused_variables,
+        reason = "Depending on what features are used to compile this crate, certain parameters may end up unused."
+    )]
+    pub fn parse(
```
andriyDev, line 40: Parsing here should also be `pub(crate)` unless I'm missing something?

crates/bevy_gltf/src/loader/extensions/khr_materials_clearcoat.rs
```diff
@@ -0,0 +1,104 @@
...
+impl ClearcoatExtension {
+    #[expect(
+        clippy::allow_attributes,
+        reason = "`unused_variables` is not always linted"
+    )]
+    #[allow(
+        unused_variables,
+        reason = "Depending on what features are used to compile this crate, certain parameters may end up unused."
+    )]
+    pub fn parse(
```
andriyDev, line 44: Same here.

crates/bevy_gltf/src/loader/extensions/khr_materials_specular.rs
```diff
@@ -0,0 +1,100 @@
...
+    pub specular_texture: Option<Handle<Image>>,
+    pub specular_color_factor: Option<[f64; 3]>,
+    #[cfg(feature = "pbr_specular_textures")]
+    pub specular_color_channel: UvChannel,
+    #[cfg(feature = "pbr_specular_textures")]
+    pub specular_color_texture: Option<Handle<Image>>,
+}
+
+impl SpecularExtension {
+    pub fn parse(
```
andriyDev, line 44: Same here.

andriyDev (approved):

crates/bevy_gltf/src/loader/extensions/khr_materials_anisotropy.rs
```diff
@@ -0,0 +1,71 @@
...
+        Some(AnisotropyExtension {
+            anisotropy_strength: extension.get("anisotropyStrength").and_then(Value::as_f64),
+            anisotropy_rotation: extension.get("anisotropyRotation").and_then(Value::as_f64),
+            #[cfg(feature = "pbr_anisotropy_texture")]
+            anisotropy_channel: anisotropy_channel.unwrap_or_default(),
+            #[cfg(feature = "pbr_anisotropy_texture")]
+            anisotropy_texture,
+        })
+    }
+}
```
andriyDev, line 71: All these pub things should also be `pub(crate)`. Please repeat that for all these extensions. AFAICT, these types weren't public before, so we should mark them as such.

andriyDev (changes requested):

hukasu:
after merge, tested with `many_foxes`, `anisotropy`, and `deferred_rendered` examples, which are some examples that i know use distinct gltfs

andriyDev (approved):

pcwalton:
@mockersf This PR makes more small files than I personally would, but it's definitely better than the current monster `load_gltf` function.

hukasu:
I see following code by jumping between different files easier that jumping between different locations on the same file

mockersf:
> I see following code by jumping between different files easier that jumping between different locations on the same file

There's a balance to reach, and this PR takes it too far in the other direction in my opinion

crates/bevy_gltf/src/loader/mod.rs
```diff
@@ -840,11 +787,11 @@ async fn load_gltf<'a, 'b, 'c>(
                         &skin,
                         joints,
                         skinned_mesh_inverse_bindposes[skin.index()].clone(),
-                        get_gltf_extras(skin.extras()),
+                        as_gltf_extras(skin.extras()),
                     );
 
                     let handle = load_context
-                        .add_labeled_asset(skin_label(&skin), gltf_skin)
+                        .add_labeled_asset(skin_label(&skin).to_string(), gltf_skin)
```
greeble-dev, line 794: I think this can use GltfSkin's method? See related comment on `gltf_ext/skin.rs`.

```suggestion
                        .add_labeled_asset(skin.asset_label(&skin).to_string(), gltf_skin)
```

crates/bevy_gltf/src/loader/gltf_ext/skin.rs
```diff
@@ -0,0 +1,13 @@
+use gltf::Skin;
+
+use crate::GltfAssetLabel;
+
+/// Return the label for the [`Skin`] as [`GltfAssetLabel::Skin`].
+pub(crate) fn skin_label(skin: &Skin<'_>) -> GltfAssetLabel {
```
greeble-dev, line 6: If `loader/mod.rs` is changed to use `GltfSkin::asset_label` then I think this becomes redundant?

hukasu:
> E.g. gltf_ext/mesh.rs is small now, but there's still 100+ lines in the loader that could move there in a follow up PR.
my first go into this at the tail end of 0.14 was a lot more aggressive with the changes, including breaking the contents of `load_gltf`, `load_material`, etc, methods into smaller methods, but there is already pushback with the current changes, so my idea was to leave to a different PR

alice-i-cecile:
I've finally taken a proper look at this, and I'm broadly on board with this. It's *much* better than the status quo, so I'm going to merge this now.

alice-i-cecile (changes requested):
Sorry, I changed my mind. I *do* think this is better than the status quo, but I'm uncomfortable with the single item files, especially since these files feel mostly complete.

Is there another splitting strategy that we can use?

hukasu:
1. I'm ok with merging the types on `assets/` into a single file
2. I could also merge the types on `extensions/` into a single file, but for this one I would vote against, as to have a clearer view of which extensions we already have
3. I could also merge the methods on `gltf_ext/` into a single file, but I would prefer not to, my reasoning for the split was, if the main type of the method is present in `gltf::N` or `gltf::json::N`, I put the method in `gltf_ext/N`, so methods that use `gltf::Gltf` or `gltf::Document` would go on the root, and `gltf::texture::Texture` and `gltf::json::texture::Info` would go into `gltf_ext/texture.rs`, etc
4.  `gltf_ext/extras.rs` I can merge with the module root or implement as `From` for the components

hukasu:
5. `ImageOrPath` and `DataUri` are in separate modules because I couldn't think of a name for the module that would work

alice-i-cecile:
Let's go with 1 and 4,  and then resolve 5 by moving them back into lib.rs for now?

hukasu:
4 by moving to `mod.rs` or `impl From`?

alice-i-cecile:
Let's go with the `From` route. Although if I understand correctly, the conversion is fallible, in which case it should be `TryFrom`.

hukasu:
is my solution for 4 satisfactory?

alice-i-cecile (approved):

alice-i-cecile:
I quite like your final solution for 4. I think this is a nicer balance now; thanks for your patience. These sorts of reorganization PRs are quite painful due to merge conflicts and bikeshedding, but it's important to do them every now and then.

PR #16004, opened 2024-10-19
Added `PressedButtons` to `PickingInteraction::Pressed`.

### LINKED ISSUE #15983: Picking: Filter keys in a component state based workflow

For managing appearance of UI it is easier to work with state, i.e. `PickingInteraction` than events, i.e. `On::<T>` from `bevy_picking`. The limitation of `PickingInteraction` currently is there is no way to distinguish the button of `Pressed`, thus limiting the usage of this pattern.

I don't think there is a clearcut solution, one or multiple of these can be considered:

1). Add a resource to globally disable keys like RMB from triggering `PickingInteraction`.
2). Add a component that filters which key triggers `PickingInteraction`.
3). Change `PickingInteraction::Pressed` into a `PickingInteraction::Pressed(Set<Key>)`.

`3` seems like the best solution but notably complicates the state machine workflow.

Keep things as is.

mintlu8:
# Objective

Fixes #15983. `PickingInteraction` is preferable to events in some use cases for code organization. This PR adds the keys pressed to `PickingInteraction::Pressed`, allowing the user to ignore certain keys like the RMB in such use cases.

## Solution

- Added field `PressedButtons` to `PickingInteraction::Pressed`.

## Testing

* Added a test for merging `PickingInteraction`s.
* Added an example for `PickingInteraction`.

## Migration Guide

`PickingInteraction::Pressed` now has a field, code matching against it should use `matches!(interaction, PickingInteraction::Pressed(..))` instead.

### DIFF
--- a/crates/bevy_picking/src/focus.rs
+++ b/crates/bevy_picking/src/focus.rs
@@ -5,6 +5,7 @@
 
 use alloc::collections::BTreeMap;
 use core::fmt::Debug;
+use core::ops::BitOrAssign;
 use std::collections::HashSet;
 
 use crate::{
@@ -190,14 +191,64 @@ fn build_hover_map(
 /// it will be considered hovered.
 #[derive(Component, Copy, Clone, Default, Eq, PartialEq, Debug, Reflect)]
 #[reflect(Component, Default, PartialEq, Debug)]
+#[repr(u8)]
 pub enum PickingInteraction {
     /// The entity is being pressed down by a pointer.
-    Pressed = 2,
+    Pressed(PressedButtons) = 2,
     /// The entity is being hovered by a pointer.
     Hovered = 1,
     /// No pointers are interacting with this entity.
     #[default]
     None = 0,
+}
+
+impl BitOrAssign for PickingInteraction {
+    fn bitor_assign(&mut self, rhs: Self) {
+        use PickingInteraction::*;
+        match (self, rhs) {
+            (Pressed(a), Pressed(b)) => *a |= b,
+            (Pressed(_), _) => (),
+            (a, Pressed(b)) => *a = Pressed(b),
+            (a @ None, Hovered) => *a = Hovered,
+            _ => (),
+        }
+    }
+}
+
+bitflags::bitflags! {
+    #[repr(transparent)]
+    #[derive(Hash, Clone, Copy, PartialEq, Eq, Debug, Reflect)]
+    #[reflect(opaque)]
+    #[reflect(Hash, PartialEq, Debug)]
+    /// Button pressed in a [`PickingInteraction`]
+    pub struct PressedButtons: u8 {
+        /// Primary mouse button is pressed.
+        const PRIMARY = 1;
+        /// Secondary mouse button is pressed.
+        const SECONDARY = 2;
+        /// Middle mouse button is pressed.
+        const MIDDLE = 4;
+        /// Touch input is pressed.
+        const TOUCH = 8;
+        /// Custom input is pressed.
+        const CUSTOM = 16;
+    }
+}
+
+impl From<PointerPress> for PressedButtons {
+    fn from(value: PointerPress) -> Self {
+        let mut result = PressedButtons::empty();
+        if value.is_primary_pressed() {
+            result |= PressedButtons::PRIMARY;
+        }
+        if value.is_secondary_pressed() {
+            result |= PressedButtons::SECONDARY;
+        }
+        if value.is_middle_pressed() {
+            result |= PressedButtons::MIDDLE;
+        }
+        result
+    }
 }
 
 /// Uses pointer events to update [`PointerInteraction`] and [`PickingInteraction`] components.
@@ -235,7 +286,12 @@ pub fn update_interactions(
             pointer_interaction.sorted_entities = sorted_entities;
 
             for hovered_entity in pointers_hovered_entities.iter().map(|(entity, _)| entity) {
-                merge_interaction_states(pointer_press, hovered_entity, &mut new_interaction_state);
+                merge_interaction_states(
+                    pointer,
+                    pointer_press,
+                    hovered_entity,
+                    &mut new_interaction_state,
+                );
             }
         }
     }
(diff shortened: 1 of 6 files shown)


alice-i-cecile (approved):
I like this. It makes `PickingInteraction` much more useful. The use of bitflag packing to reduce memory bloat is clever too.

alice-i-cecile:
This will be useful as part of #15597, and I'd like to get this merged first.

BenjaminBrienen:
Many computer mice have a "Back" and "Forward" button, but I don't know if those affect focus.

crates/bevy_picking/src/focus.rs
```diff
@@ -190,16 +191,75 @@ fn build_hover_map(
...
+    /// Button pressed in a [`PickingInteraction`]
+    pub struct PressedButtons: u8 {
+        /// Primary mouse button is pressed.
+        const PRIMARY = 1;
+        /// Secondary mouse button is pressed.
+        const SECONDARY = 2;
+        /// Middle mouse button is pressed.
+        const MIDDLE = 4;
+        /// Touch input is pressed.
+        const TOUCH = 8;
```
aevyrie, line 241: I'm not sure this is correct. Touch inputs are probably akin to a primary press, and some touch devices, like pens, can have a primary secondary, tertiary press, just like a mouse.

BenjaminBrienen: I've had something like that before: my Note 9 has a pen and it supports hovering over something and clicking.

NthTensor: A stylus is not a touch pointer. Winit does not currently support stylus input, so neither do we.

aevyrie: It definitely works, I've tested it. It's treated like a touch input.

Regardless, if you are matching on whether a button was pressed, why would you not want it to be considered pressed if it was touched by a different pointing device?

aevyrie (changes requested):
Tying an interaction struct to hardware is asking for people to just default to the mouse, and suddenly a bunch of software doesn't work with other pointing devices for no good reason.

I would take a look at the JS pointer event APIs - they were designed to be hardware agnostic for this reason: https://developer.mozilla.org/en-US/docs/Web/API/Pointer_events#determining_button_states

I would strongly prefer we follow this model, and make the easy path one that is device agnostic. E.g. under the hood, this struct might be a huge bitflag that encodes which device buttons are down, but the easiest-to-access methods are things like `is_primary_pressed`, which return true if the primary button of *any* pointing device is down.

I'd also like to ask whether this structure should even exist anymore. It was originally added to match `Interaction` in bevy_ui. However, with observers, we can do the same thing with less memory and cpu, because we don't need to update the state of every interactive entity even if no one is interested in reading it. That might mean refactoring the pointer events a bit to coalesce `button` information.

NthTensor:
> That might mean refactoring the pointer events a bit to coalesce button information.

Which I am more than happy to do if we get specs for the events we need. Enter/Leave is already on the radar.

aevyrie:
> Yeah, this is increasingly my opinion. It's nice for the "change widget style based on interaction", but that's the only use case where it's nicer than observers.

We definitely have a need to coalesce multiple events to determine state. You can't, for example, use `Over` and `Out` to determine if an entity is hovered, that entity might receive multiple of these events from multiple pointers and bubbled from children. I'm not sure what the right approach is though. I'm kinda leaning towards keeping this struct for that reason though.

I think this PR would be okay to merge with some modifications to make it match the pointer API I linked. It currently mixes the concept of abstract pointer buttons and device button.
1. Change the name to `PointerDeviceButtons` or something similar. This struct tracks which device buttons are down, but presents a public API that aggregates them, to help push users toward correct code.
2. Change `PRIMARY`/`SECONDARY` to something like `LEFT_MOUSE`/`RIGHT_MOUSE`/`MIDDLE_MOUSE`. E.g. every bit field should be device specific, this would also include mouse fwd/back and pen barrel when we have them. This structure would become an aggregate of all types of devices, if there are multiple devices (e.g. touch) all states should be OR'ed (like present).
3. Make the fields private if possible, and expose methods `primary()`, `secondary()`, `middle()`. To expose the internal details, add `pressed_by_any(button: PointerDeviceButtons)` and `pressed_by_all(button: PointerDeviceButtons)` so they can check if a specific button(s) was pressed. 
    - This would also support chording, because you could pass in `pressed_by_any(PointerDeviceButtons::LMB | PointerDeviceButtons::LMB )` or `pressed_by_all(PointerDeviceButtons::LMB | PointerDeviceButtons::LMB )`.
    - LMB and touches should both be considered a primary press. RMB and pen barrel (once it is supported) would be secondary presses. Again, follow the spec used on the web.
4. Virtual (custom) pointers should be forced to decide what kind of input they are mocking, e.g. touch/mouse/pen.

crates/bevy_picking/src/focus.rs
mintlu8: The PR just maps the input from picking in a form that makes the most sense to me. Definitely would like to see more detailed info, but that should come from picking first.

crates/bevy_picking/src/focus.rs
```diff
@@ -252,28 +317,55 @@ pub fn update_interactions(
...
 ) {
-    let new_interaction = match pointer_press.is_any_pressed() {
-        true => PickingInteraction::Pressed,
-        false => PickingInteraction::Hovered,
+    let new_interaction = if !pointer_press.is_any_pressed() {
+        PickingInteraction::Hovered
+    } else {
+        match id {
+            PointerId::Mouse => PickingInteraction::Pressed((*pointer_press).into()),
+            PointerId::Touch(_) => PickingInteraction::Pressed(PressedButtons::TOUCH),
```
alice-i-cecile, line 330: See here for the mapping from `Touch`. I think that this is the right approach for the current PR.

mintlu8:
> 1. Change the name to `PointerDevicePressed` or something similar. This struct tracks which device buttons are down, but presents a public API that aggregates them, to help push users toward correct code. `PressedButtons` is probably fine too, but we should disambiguate from, say keyboard buttons. `PointerPressed`?

Do you think it is fine to just reserve `X1` and `X2` on the bitflags? I don't see a reason to encapsulate it further.

> https://developer.mozilla.org/en-US/docs/Web/API/Pointer_events#determining_button_states

According to this spec the current implementation is a upgrade on touch input since usually we treat touch as left click, but with this implementation you can potentially treat them differently.

crates/bevy_picking/src/focus.rs
mintlu8: Renamed to `left` and `right`.

PR #16005, opened 2024-10-19
Support prefers_home_indicator_hidden

### LINKED ISSUE #15757: iOS prefersHomeIndicatorAutoHidden()

Hides the home indicator (the line at the bottom of the window) after a while, like in a lot of games or media apps like YouTube or Netflix.

Calling the Winits [with_prefers_home_indicator_hidden()](https://smithay.github.io/smithay/winit/platform/ios/trait.WindowAttributesExtIOS.html#tymethod.with_prefers_home_indicator_hidden) or even better, have it as window config option.

See also [official docs](https://developer.apple.com/documentation/uikit/uiviewcontroller/2887510-prefershomeindicatorautohidden).

I've tried to query the window and set the Winit settings myself, like in [this fashion](https://www.nikl.me/blog/2023/notes_on_android_development_using_bevy/), but `with_prefers_home_indicator_hidden()` is not just an option but a method.

Soulghost:
# Objective

- Fixes #15757 

## Solution

- Add the platform specific property `prefers_home_indicator_hidden` to bevy's Window configuration, and applying it by invoking `with_prefers_home_indicator_hidden` in `winit`.

## Testing

- I have tested the `bevy_mobile_example` on the iOS platform.

## Showcase
- Currently, the `prefers_home_indicator_hidden` is enabled in the bevy_mobile_example demo. You can test it with an iOS device. The home indicator will disappear after several seconds of inactivity in the bottom areas.

### DIFF
--- a/examples/mobile/src/lib.rs
+++ b/examples/mobile/src/lib.rs
@@ -27,6 +27,7 @@ fn main() {
                     // on iOS, gestures must be enabled.
                     // This doesn't work on Android
                     recognize_rotation_gesture: true,
+                    prefers_home_indicator_hidden: true,
                     ..default()
                 }),
                 ..default()
--- a/examples/mobile/src/lib.rs
+++ b/examples/mobile/src/lib.rs
@@ -27,6 +27,7 @@ fn main() {
                     // on iOS, gestures must be enabled.
                     // This doesn't work on Android
                     recognize_rotation_gesture: true,
+                    // Only has an effect on iOS
                     prefers_home_indicator_hidden: true,
                     ..default()
                 }),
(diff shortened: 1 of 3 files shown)


examples/mobile/src/lib.rs
```diff
@@ -27,6 +27,7 @@ fn main() {
                     // on iOS, gestures must be enabled.
                     // This doesn't work on Android
                     recognize_rotation_gesture: true,
+                    prefers_home_indicator_hidden: true,
```
alice-i-cecile, line 30:
```suggestion
                    // Only has an effect on iOS
                    prefers_home_indicator_hidden: true,
```

alice-i-cecile (approved):

mnmaita (approved):

BenjaminBrienen (approved):

PR #16018, opened 2024-10-20
fix: add reflect to `SceneInstanceReady` and other observers/events

mrchantey:
# Objective

Built-in observers & events should be `Reflect` so that components that interact with them can be serialized in scenes. This is a similar pr to #14259.


alice-i-cecile:
For context, the `bevy_cli` working group has added a lint to catch these issues :) https://github.com/TheBevyFlock/bevy_cli/pull/139

alice-i-cecile (approved):

PR #16020, opened 2024-10-20
upgrade image crate to v0.25.4

piaoger:
Objective

    Keep image crate update to date.

    image v0.25.4 brings some improvements on webp decoding and fast blur.It also contains some bug fixes.

    It maybe a good idea to keep Image crate up to date for new release.

Solution

    Upgrade image to v0.25.4

Testing

    I've run some 2d/3d examples and all looks good. 

Breaking changes

    No breaking changes :)


mockersf:
This is not needed, this version will be picked up by cargo automatically.

Unless there's a specific need in Bevy for a version, it's better for compatibility to keep the version required as low as possible

Issue #15663, opened 2024-10-05
Add `.contains_aabb` for `Frustum`


torsteingrindvik:
## What problem does this solve or what need does it fill?

Solves making logic decisions based on whether an object is contained in a frustum.

_My specific usecase: N cameras exist, M<N should be active. Figure out which M based on visiblity of some AABB._

## What solution would you like?

Adding `.contains_aabb` to `Frustum`.

## What alternative(s) have you considered?

There is `.intersects_obb`, but that is only _intersects_ and not _contains_.

## Additional context

Discord context feat @Jondolf , @mweatherley : https://discord.com/channels/691052431525675048/1203087353850364004/1292080527125254185

torsteingrindvik:
Related issue for intersection (but not contains) API docs: #6096

PR #16022, opened 2024-10-20
Add .contains_aabb for Frustum

Soulghost:
# Objective

- Fixes: #15663

## Solution

- Add an `is_forward_plane` method to `Aabb`, and a `contains_aabb` method to `Frustum`.

## Test
- I have created a frustum with an offset along with three unit tests to evaluate the `contains_aabb` algorithm.

## Explanation for the Test Cases
- To facilitate the code review, I will explain how the frustum is created. Initially, we create a frustum without any offset and then create a cuboid that is just contained within it.

image

- Secondly, we move the cuboid by 2 units along both the x-axis and the y-axis to make it more general.

## Reference
- [Frustum Culling](https://learnopengl.com/Guest-Articles/2021/Scene/Frustum-Culling#)
- [AABB Plane intersection](https://gdbooks.gitbooks.io/3dcollisions/content/Chapter2/static_aabb_plane.html)

### DIFF
--- a/crates/bevy_render/src/primitives/mod.rs
+++ b/crates/bevy_render/src/primitives/mod.rs
@@ -95,6 +95,33 @@ impl Aabb {
     pub fn max(&self) -> Vec3A {
         self.center + self.half_extents
     }
+
+    #[inline]
+    pub fn is_forward_plane(
+        &self,
+        p_normal: &Vec3A,
+        p_distance: f32,
+        world_from_local: &Affine3A,
+    ) -> bool {
+        let aabb_center_world = world_from_local.transform_point3a(self.center);
+        let bb_axis_x_world = world_from_local.x_axis * self.half_extents.x;
+        let bb_axis_y_world = world_from_local.y_axis * self.half_extents.y;
+        let bb_axis_z_world = world_from_local.z_axis * self.half_extents.z;
+        let half_extents_x_world = bb_axis_x_world.dot(Vec3A::X).abs()
+            + bb_axis_y_world.dot(Vec3A::X).abs()
+            + bb_axis_z_world.dot(Vec3A::X).abs();
+        let half_extents_y_world = bb_axis_x_world.dot(Vec3A::Y).abs()
+            + bb_axis_y_world.dot(Vec3A::Y).abs()
+            + bb_axis_z_world.dot(Vec3A::Y).abs();
+        let half_extents_z_world = bb_axis_x_world.dot(Vec3A::Z).abs()
+            + bb_axis_y_world.dot(Vec3A::Z).abs()
+            + bb_axis_z_world.dot(Vec3A::Z).abs();
+        let r = half_extents_x_world * p_normal.x.abs()
+            + half_extents_y_world * p_normal.y.abs()
+            + half_extents_z_world * p_normal.z.abs();
+        let signed_distance = p_normal.dot(aabb_center_world) - p_distance;
+        signed_distance > r
+    }
 }
 
 impl From<Sphere> for Aabb {
@@ -293,6 +320,17 @@ impl Frustum {
             let p_normal = half_space.normal();
             let relative_radius = aabb.relative_radius(&p_normal, &world_from_local.matrix3);
             if half_space.normal_d().dot(aabb_center_world) + relative_radius <= 0.0 {
+                return false;
+            }
+        }
+        true
+    }
+
+    #[inline]
+    pub fn contains_aabb(&self, aabb: &Aabb, world_from_local: &Affine3A) -> bool {
+        for half_space in &self.half_spaces {
+            let p_normal = half_space.normal();
+            if !aabb.is_forward_plane(&p_normal, half_space.d(), world_from_local) {
                 return false;
             }
         }
--- a/crates/bevy_render/src/primitives/mod.rs
+++ b/crates/bevy_render/src/primitives/mod.rs
@@ -556,7 +556,7 @@ mod tests {
         let half_extent_world = (((49.5 * 49.5) * 0.5) as f32).sqrt() + 0.5f32.sqrt();
         let near = 50.5 - half_extent_world;
         let far = near + 2.0 * half_extent_world;
-        let fov = 2.0f32 * ops::atan(half_extent_world / near);
+        let fov = 2.0 * ops::atan(half_extent_world / near);
         let proj = PerspectiveProjection {
             aspect_ratio: 1.0,
             near,
--- a/crates/bevy_render/src/primitives/mod.rs
+++ b/crates/bevy_render/src/primitives/mod.rs
@@ -108,7 +108,7 @@ impl Aabb {
         // transform the half-extents into world space.
         let half_extents_world = world_from_local.matrix3.abs() * self.half_extents.abs();
         // collapse the half-extents onto the plane normal.
-        let r = (world_from_local.matrix3 * self.half_extents).dot(p_normal.abs());
+        let r = half_extents_world.dot(p_normal.abs());
         let aabb_center_world = world_from_local.transform_point3a(self.center);
         let signed_distance = p_normal.dot(aabb_center_world) + p_distance;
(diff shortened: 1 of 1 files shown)


BenjaminBrienen (approved):

IQuick143:
I'll review later today, I have some comments to the math.

crates/bevy_render/src/primitives/mod.rs
```diff
@@ -95,6 +95,35 @@ impl Aabb {
...
+        let bb_axis_z_world = world_from_local.z_axis * self.half_extents.z;
+        let half_extents_x_world = bb_axis_x_world.dot(Vec3A::X).abs()
+            + bb_axis_y_world.dot(Vec3A::X).abs()
+            + bb_axis_z_world.dot(Vec3A::X).abs();
+        let half_extents_y_world = bb_axis_x_world.dot(Vec3A::Y).abs()
+            + bb_axis_y_world.dot(Vec3A::Y).abs()
+            + bb_axis_z_world.dot(Vec3A::Y).abs();
+        let half_extents_z_world = bb_axis_x_world.dot(Vec3A::Z).abs()
+            + bb_axis_y_world.dot(Vec3A::Z).abs()
+            + bb_axis_z_world.dot(Vec3A::Z).abs();
```
IQuick143, line 120: All these mutliplications by scalar vectors could be reduced to choosing a single component of a vector.
```
        let bb_axis_x_world = world_from_local.x_axis * self.half_extents.x;
        let bb_axis_y_world = world_from_local.y_axis * self.half_extents.y;
        let bb_axis_z_world = world_from_local.z_axis * self.half_extents.z;
        let half_extents_x_world = bb_axis_x_world.x.abs()
            + bb_axis_y_world.x.abs()
            + bb_axis_z_world.x.abs();
        let half_extents_y_world = bb_axis_x_world.y.abs()
            + bb_axis_y_world.y.abs()
            + bb_axis_z_world.y.abs();
        let half_extents_z_world = bb_axis_x_world.z.abs()
            + bb_axis_y_world.z.abs()
            + bb_axis_z_world.z.abs();
```
Then we can push the `.abs()` all the way back to the vectors (in glam (and by extent bevy_math) you can call .abs() on a vector to apply absolute value to each component)
```
        let bb_axis_x_world_abs = (world_from_local.x_axis * self.half_extents.x).abs();
        let bb_axis_y_world_abs = (world_from_local.y_axis * self.half_extents.y).abs();
        let bb_axis_z_world_abs = (world_from_local.z_axis * self.half_extents.z).abs();
        let half_extents_x_world = bb_axis_x_world_abs.x
            + bb_axis_y_world_abs.x
            + bb_axis_z_world_abs.x;
        let half_extents_y_world = bb_axis_x_world_abs.y
            + bb_axis_y_world_abs.y
            + bb_axis_z_world_abs.y;
        let half_extents_z_world = bb_axis_x_world_abs.z
            + bb_axis_y_world_abs.z
            + bb_axis_z_world_abs.z;
```
We can also notice, we always sum up the same components, this reduces to vector addition
```
        let bb_axis_x_world_abs = (world_from_local.x_axis * self.half_extents.x).abs();
        let bb_axis_y_world_abs = (world_from_local.y_axis * self.half_extents.y).abs();
        let bb_axis_z_world_abs = (world_from_local.z_axis * self.half_extents.z).abs();
        let half_extents_world = bb_axis_x_world_abs + bb_axis_y_world_abs + bb_axis_z_world_abs;
```
Then we note that the absolute value can distribute over the multiplication (there are no additions taking place on that line).
```
        let bb_axis_x_world_abs = world_from_local.x_axis.abs() * self.half_extents.abs().x;
        let bb_axis_y_world_abs = world_from_local.y_axis.abs() * self.half_extents.abs().y;
        let bb_axis_z_world_abs = world_from_local.z_axis.abs() * self.half_extents.abs().z;
        let half_extents_world = bb_axis_x_world_abs + bb_axis_y_world_abs + bb_axis_z_world_abs;
```
Finally, one has to notice that we are effectively just doing matrix - vector multiplication, written out.
Affine3A::matrix3 gives you a matrix containing exactly the columns you access with x_axis, y_axis, z_axis, we can apply absolute value to the entire matrix:
```suggestion
        let half_extents_world = world_from_local.matrix3.abs() * self.half_extents.abs();
```

crates/bevy_render/src/primitives/mod.rs
```diff
@@ -95,6 +95,35 @@ impl Aabb {
...
+        let r = half_extents_x_world * p_normal.x.abs()
+            + half_extents_y_world * p_normal.y.abs()
+            + half_extents_z_world * p_normal.z.abs();
```
IQuick143, line 123: By similar operations here, you can simplify this down to:
```suggestion
        let r = (world_from_local.matrix3 * self.half_extents).dot(p_normal.abs())
```
Although I'm not sure if the absolute value is in a correct spot here, I'll think about it more, I've been just doing algebra so far.

crates/bevy_render/src/primitives/mod.rs
```diff
@@ -502,4 +548,45 @@ mod tests {
...
+            fov: 90.0_f32.to_radians(),
+            aspect_ratio: 1.0,
+            near: 1.0,
+            far: 100.0,
+        };
+        proj.compute_frustum(&GlobalTransform::from_translation(Vec3::new(2.0, 2.0, 0.0)))
+    }
+
+    #[test]
+    fn aabb_inside_frustum() {
```
IQuick143, line 564: All of these tests use Affine3A::IDENTITY, which is the most simplistic test case, and can hide issues regarding the Affine computations.

I'd recommend adding more tests with with different Affine transformations.

IQuick143 (changes requested):
I'll have to go over the geometry to make sure the algorithm is correct, but I've left some comments regarding algebraic simplifications of the code. (Most notably they enable better SIMD optimisations by the compiler.)

Soulghost:
Hi @IQuick143 , sorry for the late response. I have resolved all the issues and added some cases involving rotation, please take another look and review them. Thank you!

crates/bevy_render/src/primitives/mod.rs
```diff
@@ -95,6 +95,24 @@ impl Aabb {
     pub fn max(&self) -> Vec3A {
         self.center + self.half_extents
     }
+
+    /// Check if the point is at the front side of the plane.
+    /// Referenced from: [AABB Plane intersection](https://gdbooks.gitbooks.io/3dcollisions/content/Chapter2/static_aabb_plane.html)
+    #[inline]
+    pub fn is_forward_plane(
```
IQuick143, line 102: I think that a better name could be `is_in_front_of_plane` or `is_in_halfspace`.

crates/bevy_render/src/primitives/mod.rs
```diff
@@ -95,6 +95,24 @@ impl Aabb {
     pub fn max(&self) -> Vec3A {
         self.center + self.half_extents
     }
+
+    /// Check if the point is at the front side of the plane.
```
IQuick143, line 99:
```suggestion
    /// Check if the AABB is at the front side of the plane.
```

crates/bevy_render/src/primitives/mod.rs
```diff
@@ -95,6 +95,24 @@ impl Aabb {
...
+    /// Referenced from: [AABB Plane intersection](https://gdbooks.gitbooks.io/3dcollisions/content/Chapter2/static_aabb_plane.html)
+    #[inline]
+    pub fn is_forward_plane(
+        &self,
+        p_normal: &Vec3A,
+        p_distance: f32,
```
IQuick143, line 105: We could perhaps merge these two inputs into a single: `plane: HalfSpace`, this would better align with the use later for Frustum intersections, also the Vec3A already has a size of 4 f32's.

IQuick143 (approved):
Seems good, hope I didn't miss anything.

Thanks for the PR.

mockersf (approved):

PR #16025, opened 2024-10-20
Fix typos in `config_fast_builds.toml`

benfrankel:
From https://github.com/TheBevyFlock/bevy_new_2d/pull/317 plus a few other changes.

### DIFF
--- a/.cargo/config_fast_builds.toml
+++ b/.cargo/config_fast_builds.toml
@@ -21,7 +21,7 @@
 # Mold is a newer linker written by one of the authors of LLD. It boasts even greater performance, specifically
 # through its high parallelism, though it only supports Linux.
 #
-# Mold is  disabled by default in this file. If you wish to enable it, follow the installation instructions for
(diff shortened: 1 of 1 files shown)


MiniaczQ (approved):

mockersf:
do you know why this file isn't checked by the typo CI?

mockersf (approved):

benfrankel:
CI uses the `crate-ci/typos` action, which has info on this in its FAQ: https://github.com/crate-ci/typos?tab=readme-ov-file#why-was--not-corrected

mockersf:
> EDIT: Following this, it seems we have `ignore-hidden` set to `true` (which is the default value) so `typos` skips the `.cargo/` directory.

Thanks!

I think it could be worth a try to check if disabling that doesn't raise too many issues

benfrankel:
> I think it could be worth a try to check if disabling that doesn't raise too many issues

The output seems good to me if we also explicitly exclude `.git`. I can make a separate PR for this.

PR #14194, opened 2024-07-06
Replace `AsyncSeek` trait by `AsyncSeekForward` for `Reader` to address #12880

ldubos:
# Objective

The primary motivation behind this PR is to (partially?) address the limitations imposed by the recently added `AsyncSeek` trait bound discussed in issue #12880. While the `AsyncSeek` trait add some flexibility to the reader, it inadvertently restricts the ability to write asset readers that can truly stream bytes, particularly in scenarios like HTTP requests where backward seeking is not supported. It is also challenging in contexts where assets are stored in compressed formats or require other kinds of transformations.

The logic behind this change is that currently, with `AsyncSeek`, an asset Reader based on streamed data will either 1) fail silently, 2) return an error, or 3) use a buffer to satisfy the trait constraint. I believe that being able to advance in the file without having to "read" it is a good thing. The only issue here is the ability to seek backward. It is highly likely that in this context, we only need to seek forward in the file because we would have already read an entry table upstream and just want to access one or more resources further in the file. I understand that in some cases, this may not be applicable, but I think it is more beneficial not to constrain `Reader`s that want to stream than to allow "Assets" to read files in a completely arbitrary order.

## Solution

Replace the current `AsyncSeek` trait with `AsyncSeekForward` on asset `Reader`

## Changelog

- Introduced a new custom trait, `AsyncSeekForward`, for the asset Reader.
- Replaced the current `AsyncSeek` trait with `AsyncSeekForward` for all asset `Reader` implementations.

## Migration Guide

Replace all instances of `AsyncSeek` with `AsyncSeekForward` in your asset reader implementations.

PR #16027, opened 2024-10-20
Add `AsyncSeekForwardExt` trait to allows a similar API to the previous Bevy version

ldubos:
# Objective

This PR introduces an `AsyncSeekForwardExt` trait, which I forgot in my previous PR #14194.

This new trait is analogous to `AsyncSeekExt` and allows all implementors of `AsyncSeekForward` to directly use the `seek_forward` function in async contexts.

## Solution

- Implement a new `AsyncSeekForwardExt` trait
- Automatically implement this trait for all types that implement `AsyncSeekForward`

## Showcase

This new trait allows a similar API to the previous Bevy version:

```rust
#[derive(Default)]
struct UniverseLoader;

#[derive(Asset, TypePath, Debug)]
struct JustALilAsteroid([u8; 128]);

impl AssetLoader for UniverseLoader {
    type Asset = JustALilAsteroid;
    type Settings = ();
    type Error = std::io::Error;

    async fn load<'a>(
        &'a self,
        reader: &'a mut Reader<'a>,
        _settings: &'a Self::Settings,
        _context: &'a mut LoadContext<'_>,
    ) -> Result<Self::Asset, Self::Error> {
        // read the asteroids entry table
        let entry_offset: u64 = /* ... */;
        let current_offset: u64 = reader.seek_forward(0).await?;

        // jump to the entry
        reader.seek_forward(entry_offset - current_offset).await?;

        let mut asteroid_buf = [0; 128];
        reader.read_exact(&mut asteroid_buf).await?;

        Ok(JustALilAsteroid(asteroid_buf))
    }
    
    fn extensions(&self) -> &[&str] {
        &["celestial"]
    }
}
```

### DIFF
--- a/crates/bevy_asset/src/io/mod.rs
+++ b/crates/bevy_asset/src/io/mod.rs
@@ -31,7 +31,10 @@ use core::{
 use derive_more::derive::{Display, Error, From};
 use futures_io::{AsyncRead, AsyncWrite};
 use futures_lite::{ready, Stream};
-use std::path::{Path, PathBuf};
+use std::{
+    future::Future,
+    path::{Path, PathBuf},
+};
 
 /// Errors that occur while loading assets.
 #[derive(Error, Display, Debug, Clone)]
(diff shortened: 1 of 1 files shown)


BenjaminBrienen (approved):

crates/bevy_asset/src/io/mod.rs
```diff
@@ -120,6 +121,38 @@ impl<T: ?Sized + AsyncSeekForward + Unpin> AsyncSeekForward for Box<T> {
     }
 }
 
+pub trait AsyncSeekForwardExt: AsyncSeekForward {
+    fn seek_forward(&mut self, offset: u64) -> SeekForwardFuture<'_, Self>
```
bushrat011899, line 125: Should include some documentation here. E.g.,

```rust
/// Seek by the provided `offset` in the forwards direction, using the [`AsyncSeekForward`] trait.
```

bushrat011899 (approved):
Looks good with a clean implementation. I do recommend putting some documentation on the method though, even if the method name is self explanatory.

crates/bevy_asset/src/io/mod.rs
ldubos: Good point

cart (approved):
Definitely good to make this easy to call!

PR #16047, opened 2024-10-21
track_change_detection: Also track spawns/despawns

SpecificProtagonist:
# Objective

Expand `track_change_detection` feature to also track entity spawns and despawns. Use this to create better error messages.

# Solution

Adds `Entities::entity_get_spawned_or_despawned_by` as well as `{all entity reference types}::spawned_by`.

This also removes the deprecated `get_many_entities_mut` & co (and therefore can't land in 0.15) because we don't yet have no Polonius.

## Testing

Added a test that checks that the locations get updated and these updates are ordered correctly vs hooks & observers.

---

## Showcase

Access location:
```rust
let mut world = World::new();
let entity = world.spawn_empty().id();
println!("spawned by: {}", world.entity(entity).spawned_by());
```
```
spawned by: src/main.rs:5:24
```
Error message (with `track_change_detection`):
```rust
world.despawn(entity);
world.entity(entity);
```
```
thread 'main' panicked at src/main.rs:11:11:
Entity 0v1#4294967296 was despawned by src/main.rs:10:11
```
and without:
```
thread 'main' panicked at src/main.rs:11:11:
Entity 0v1#4294967296 does not exist (enable `track_change_detection` feature for more details)
```
Similar error messages now also exists for `Query::get`, `World::entity_mut`, `EntityCommands` creation and everything that causes `B0003`, e.g.
```
error[B0003]: Could not insert a bundle (of type `MaterialMeshBundle<StandardMaterial>`) for entity Entity { index: 7, generation: 1 }, which was despawned by src/main.rs:10:11. See: https://bevyengine.org/learn/errors/#b0003
```

### DIFF
--- a/crates/bevy_ecs/src/entity/mod.rs
+++ b/crates/bevy_ecs/src/entity/mod.rs
@@ -61,6 +61,8 @@ use crate::{
     },
     storage::{SparseSetIndex, TableId, TableRow},
 };
+#[cfg(feature = "track_change_detection")]
+use core::panic::Location;
 use core::{fmt, hash::Hash, mem, num::NonZero, sync::atomic::Ordering};
 #[cfg(feature = "serialize")]
 use serde::{Deserialize, Serialize};
@@ -938,6 +940,46 @@ impl Entities {
     pub fn is_empty(&self) -> bool {
         self.len == 0
     }
+
+    /// Sets the source code location from which this entity has last been spawned
+    /// or despawned.
+    #[cfg(feature = "track_change_detection")]
+    #[inline]
+    pub(crate) fn set_spawned_or_despawned_by(&mut self, index: u32, caller: &'static Location) {
+        let meta = self
+            .meta
+            .get_mut(index as usize)
+            .expect("Entity index invalid");
+        meta.spawned_or_despawned_by = Some(caller);
+    }
+
+    /// Returns the source code location from which this entity has last been spawned
+    /// or despawned. Returns `None` if this entity has never existed.
+    #[cfg(feature = "track_change_detection")]
+    pub fn entity_get_spawned_or_despawned_by(
+        &self,
+        entity: Entity,
+    ) -> Option<&'static Location<'static>> {
+        self.meta
+            .get(entity.index() as usize)
+            .and_then(|meta| meta.spawned_or_despawned_by)
+    }
+
+    /// Constructs a message explaining why an entity does not exists, if known.
+    pub(crate) fn entity_does_not_exist_error_details_message(&self, _entity: Entity) -> String {
+        #[cfg(feature = "track_change_detection")]
+        {
+            if let Some(location) = self.entity_get_spawned_or_despawned_by(_entity) {
+                format!("was despawned by {location}",)
+            } else {
+                "was never spawned".to_owned()
+            }
+        }
+        #[cfg(not(feature = "track_change_detection"))]
+        {
+            "does not exist (enable `track_change_detection` feature for more details)".to_owned()
+        }
+    }
 }
 
 #[derive(Copy, Clone, Debug)]
@@ -946,6 +988,9 @@ struct EntityMeta {
     pub generation: NonZero<u32>,
     /// The current location of the [`Entity`]
     pub location: EntityLocation,
+    /// Location of the last spawn or despawn of this entity
+    #[cfg(feature = "track_change_detection")]
+    spawned_or_despawned_by: Option<&'static Location<'static>>,
 }
 
 impl EntityMeta {
@@ -953,10 +998,12 @@ impl EntityMeta {
     const EMPTY: EntityMeta = EntityMeta {
         generation: NonZero::<u32>::MIN,
         location: EntityLocation::INVALID,
+        #[cfg(feature = "track_change_detection")]
+        spawned_or_despawned_by: None,
     };
 }
 
-/// Records where an entity's data is stored.
+/// A location of an entity in an archetype.
 #[derive(Copy, Clone, Debug, PartialEq)]
 pub struct EntityLocation {
     /// The ID of the [`Archetype`] the [`Entity`] belongs to.
--- a/crates/bevy_ecs/src/query/error.rs
+++ b/crates/bevy_ecs/src/query/error.rs
@@ -11,7 +11,7 @@ pub enum QueryEntityError<'w> {
     /// Either it does not have a requested component, or it has a component which the query filters out.
     QueryDoesNotMatch(Entity, UnsafeWorldCell<'w>),
     /// The given [`Entity`] does not exist.
-    NoSuchEntity(Entity),
+    NoSuchEntity(Entity, UnsafeWorldCell<'w>),
     /// The [`Entity`] was requested mutably more than once.
     ///
     /// See [`QueryState::get_many_mut`](crate::query::QueryState::get_many_mut) for an example.
@@ -26,15 +26,22 @@ impl<'w> core::fmt::Display for QueryEntityError<'w> {
             Self::QueryDoesNotMatch(entity, world) => {
                 write!(
                     f,
-                    "The query does not match the entity {entity}, which has components "
+                    "The query does not match entity {entity}, which has components "
                 )?;
                 format_archetype(f, world, entity)
             }
-            Self::NoSuchEntity(entity) => write!(f, "The entity {entity} does not exist"),
-            Self::AliasedMutability(entity) => write!(
-                f,
-                "The entity {entity} was requested mutably more than once"
-            ),
+            Self::NoSuchEntity(entity, world) => {
+                write!(
+                    f,
+                    "Entity {entity} {}",
+                    world
+                        .entities()
+                        .entity_does_not_exist_error_details_message(entity)
+                )
+            }
+            Self::AliasedMutability(entity) => {
+                write!(f, "Entity {entity} was requested mutably more than once")
+            }
         }
     }
 }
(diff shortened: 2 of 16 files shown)


crates/bevy_ecs/Cargo.toml
```diff
@@ -11,7 +11,7 @@ categories = ["game-engines", "data-structures"]
 rust-version = "1.77.0"
 
 [features]
-default = ["bevy_reflect"]
+default = ["bevy_reflect", "track_change_detection"] # TMP DO NO INCLUDE
```
aevyrie, line 14: Adding review note to remember to remove. :)

crates/bevy_ecs/src/query/error.rs
```diff
@@ -30,7 +30,21 @@ impl<'w> core::fmt::Display for QueryEntityError<'w> {
                 )?;
                 format_archetype(f, world, entity)
             }
-            Self::NoSuchEntity(entity) => write!(f, "The entity {entity} does not exist"),
+            Self::NoSuchEntity(entity, world) => {
+                #[cfg(feature = "track_change_detection")]
+                {
+                    if let Some(location) = world.entities().get_entity_spawned_despawned_by(entity)
```
aevyrie, line 36: This is really cool.

aevyrie (approved):

crates/bevy_ecs/src/entity/mod.rs
```diff
@@ -969,13 +1001,18 @@ struct EntityMeta {
     pub generation: NonZero<u32>,
     /// The current location of the [`Entity`]
     pub location: EntityLocation,
+    /// Location of the last spawn or despawn of this entity
+    #[cfg(feature = "track_change_detection")]
+    spawned_despawned_by: &'static Location<'static>,
```
bushrat011899, line 1006: I experimented with this change as well, but noticed this safety comment on `EntityMeta`:

> Safety:
> This type must not contain any pointers at any level, and be safe to fully fill with `u8::MAX`.

I believe this new field breaks _both_ of these safety invariants, since it's a pointer type (static reference), _and_ it's not safe to fill with all ones.

I _do_ think it _should_ be possible to relax the pointer requirement, since this is a read-only static reference (I can't imagine a scenario where some combination of transmuting, dropping, etc. could be unsound). But the all-ones-fill could create an invalid reference.

I think the most transparent workaround here would be to use a `union` type:

```rust
#[repr(C)]
union LocationOrByteFill {
    fill: NonZero<usize>,
    location: &'static Location<'static>,
}
```

Then when reading the value, you'd have to check if the `fill` value is `usize::MAX`, and if so, panic.

Alternatively, maybe you could update `flush_and_reserve_invalid_assuming_no_entities` (the function that performs the fill) to also track-caller and use the real `Location` value. I'm not sure which is easier/more performant.

crates/bevy_ecs/src/query/error.rs
```diff
@@ -30,7 +30,22 @@ impl<'w> core::fmt::Display for QueryEntityError<'w> {
...
+                        _world.entities().get_entity_spawned_despawned_by(entity)
+                    {
+                        write!(f, "The entity {entity} was despawned by {location}",)
+                    } else {
+                        write!(f, "The entity {entity} was never spawned",)
+                    }
+                }
+                #[cfg(not(feature = "track_change_detection"))]
+                {
+                    write!(f, "The entity {entity} does not exist")
```
bushrat011899, line 46: Might be nice to include a hint here to enable `track_change_detection` for a more detailed error message?

crates/bevy_ecs/src/world/entity_ref.rs
```diff
@@ -272,6 +274,14 @@ impl<'w> EntityRef<'w> {
...
         unsafe { self.0.get_components::<Q>() }
     }
+
+    /// Returns the source code location from which this entity has last been spawned
+    /// or despawned. Returns `None` if this entity has never existed.
+    /// In a removal hook or observer, this points to the despawn.
+    #[cfg(feature = "track_change_detection")]
+    pub fn get_spawned_despawned_by(&self) -> &'static Location<'static> {
+        self.0.get_spawned_despawned_by()
+    }
```
bushrat011899, line 284: I believe this documentation is out-of-date, since this function does not return an `Option`.

crates/bevy_ecs/src/world/entity_ref.rs
```diff
@@ -762,6 +772,14 @@ impl<'w> EntityMut<'w> {
...
         unsafe { component_ids.fetch_mut(self.0) }
     }
+
+    /// Returns the source code location from which this entity has last been spawned
+    /// or despawned. Returns `None` if this entity has never existed.
+    /// In a removal hook or observer, this points to the despawn.
+    #[cfg(feature = "track_change_detection")]
+    pub fn get_spawned_despawned_by(&self) -> &'static Location<'static> {
+        self.0.get_spawned_despawned_by()
+    }
```
bushrat011899, line 782: Same as above re:`Option`

crates/bevy_ecs/src/world/entity_ref.rs
```diff
@@ -2416,6 +2467,14 @@ impl<'w> FilteredEntityRef<'w> {
...
             .flatten()
     }
+
+    /// Returns the source code location from which this entity has last been spawned
+    /// or despawned. Returns `None` if this entity has never existed.
+    /// In a removal hook or observer, this points to the despawn.
+    #[cfg(feature = "track_change_detection")]
+    pub fn get_spawned_despawned_by(&self) -> &'static Location<'static> {
+        self.entity.get_spawned_despawned_by()
+    }
```
bushrat011899, line 2477: Same as above re:`Option`

crates/bevy_ecs/src/world/entity_ref.rs
```diff
@@ -2720,6 +2779,14 @@ impl<'w> FilteredEntityMut<'w> {
...
```
bushrat011899, line 2789: Same as above re:`Option`

crates/bevy_ecs/src/world/entity_ref.rs
```diff
@@ -2856,6 +2923,14 @@ where
...
         }
     }
+
+    /// Returns the source code location from which this entity has last been spawned
+    /// or despawned. Returns `None` if this entity has never existed.
+    /// In a removal hook or observer, this points to the despawn.
+    #[cfg(feature = "track_change_detection")]
+    pub fn get_spawned_despawned_by(&self) -> &'static Location<'static> {
+        self.entity.get_spawned_despawned_by()
+    }
```
bushrat011899, line 2933: Same as above re:`Option`

crates/bevy_ecs/src/world/entity_ref.rs
```diff
@@ -2964,6 +3039,14 @@ where
...
```
bushrat011899, line 3049: Same as above re:`Option`

crates/bevy_ecs/src/world/unsafe_world_cell.rs
```diff
@@ -996,6 +996,16 @@ impl<'w> UnsafeEntityCell<'w> {
...
+
+    /// Returns the source code location from which this entity has last been spawned
+    /// or despawned.
+    #[cfg(feature = "track_change_detection")]
+    pub fn get_spawned_despawned_by(self) -> &'static Location<'static> {
+        self.world()
+            .entities()
+            .get_entity_spawned_despawned_by(self.entity)
+            .unwrap()
+    }
```
bushrat011899, line 1008: I suspect this documentation is what should be used for the above methods instead.

bushrat011899 (changes requested):
Really nice change, I think this philosophy should be extended to more areas of Bevy too (e.g., `Component`s). However, there is a safety comment which I believe this change violates. Either the PR needs to be amended to satisfy the invariant, or a `SAFETY` comment should be added to the field to indicate why it is safe.

crates/bevy_ecs/src/entity/mod.rs
hymm: Could potentially remove the method. I don't think it's being used anymore after retained render world

bushrat011899: With the safety comment updated that would also be an acceptable resolution to the issue.

crates/bevy_ecs/src/world/unsafe_world_cell.rs
aevyrie: The `get_` can also be dropped

crates/bevy_ecs/src/query/error.rs
bushrat011899: Yeah that's a good idea. And like the `tracing` filter level features, we should include a note saying library authors should _not_ enable this feature on their published crates, as it removes the ability for end-users to choose themselves.

crates/bevy_ecs/src/world/entity_ref.rs
```diff
@@ -1638,7 +1652,18 @@ impl<'w> EntityWorldMut<'w> {
     /// Despawns the current entity.
     ///
     /// See [`World::despawn`] for more details.
+    #[track_caller]
     pub fn despawn(self) {
+        self.despawn_with_caller(
+            #[cfg(feature = "track_change_detection")]
```
BenjaminBrienen, line 1658: This is pretty clever language design, I think.

BenjaminBrienen (approved):

bushrat011899 (approved):
Looks great!

NthTensor (approved):
Fantastic. I've marked this as ready for review.

alice-i-cecile:
Really lovely. We definitely need to rename the feature, but that's for follow-up. @SpecificProtagonist once merge conflicts are resolved I'll press the button.

alice-i-cecile:
CI is happy; I'm happy. Let's go: this looks so sweet.

alice-i-cecile:
Thank you to everyone involved with the authoring or reviewing of this PR! This work is relatively important and needs release notes! Head over to https://github.com/bevyengine/bevy-website/issues/1974 if you'd like to help out.

PR #16050, opened 2024-10-22
Compute better smooth normals for cheaper, maybe

shanecelis:
# Objective

Avoid a premature normalize operation and get better smooth normals for it.

## Inspiration

@IceSentry suggested `face_normal()` could have its normalize removed based on [this article](https://iquilezles.org/articles/normals/) in PR #16039. 

## Solution

I did not want to change `face_normal()` to return a vector that's not normalized. The name "normal" implies it'll be normalized. Instead I added the `face_area_normal()` function, whose result is not normalized. Its magnitude is equal two times the triangle's area. I've noted why this is the case in its doc comment.

I changed `compute_smooth_normals()` from computing normals from adjacent faces with equal weight to use the area of the faces as a weight. This has the benefit of being cheaper computationally and hopefully produces better normals.

The `compute_flat_normals()` is unchanged and still uses `face_normal()`.

## Testing

One test was added which shows the bigger triangle having an effect on the normal, but the previous test that uses the same size triangles is unchanged.

**WARNING:** No visual test has been done yet. No example exists that demonstrates the compute_smooth_normals(). Perhaps there's a good model to demonstrate what the differences are. I would love to have some input on this.

I'd suggest @IceSentry and @stepancheg to review this PR.

## Further Considerations

It's possible weighting normals by their area is not definitely better than unweighted. It's possible there may be aesthetic reasons to prefer one over the other. In such a case, we could offer two variants: weighted or unweighted. Or we could offer another function perhaps like this: `compute_smooth_normals_with_weights(|normal, area| 1.0)` which would restore the original unweighted sum of normals.

---

## Showcase

Smooth normal calculation now weights adjacent face normals by their area.

## Migration Guide

### DIFF
--- a/crates/bevy_mesh/src/vertex.rs
+++ b/crates/bevy_mesh/src/vertex.rs
@@ -138,6 +138,25 @@ pub(crate) struct MeshAttributeData {
     pub(crate) values: VertexAttributeValues,
 }
 
+/// Compute a vector whose direction is the normal of the triangle formed by
+/// points a, b, c, and whose magnitude is double the area of the triangle. This
+/// is useful for computing smooth normals where the contributing normals are
+/// proportionate to the areas of the triangles as [discussed
+/// here](https://iquilezles.org/articles/normals/).
+///
+/// Question: Why double the area? Because the area of a triangle A is
+/// determined by a cross product like so:
+///
+/// _A = |(b - a) x (c - a)| / 2
+///
+/// By computing _2 A_ we avoid a division operation, and when calculating the
+/// the sum of these vectors which are then normalized, a constant multiple has
+/// no effect.
+pub(crate) fn face_area_normal(a: [f32; 3], b: [f32; 3], c: [f32; 3]) -> [f32; 3] {
+    let (a, b, c) = (Vec3::from(a), Vec3::from(b), Vec3::from(c));
+    (b - a).cross(c - a).into()
+}
+
 pub(crate) fn face_normal(a: [f32; 3], b: [f32; 3], c: [f32; 3]) -> [f32; 3] {
     let (a, b, c) = (Vec3::from(a), Vec3::from(b), Vec3::from(c));
     (b - a).cross(c - a).normalize().into()
--- a/crates/bevy_mesh/src/vertex.rs
+++ b/crates/bevy_mesh/src/vertex.rs
@@ -144,10 +144,10 @@ pub(crate) struct MeshAttributeData {
 /// proportionate to the areas of the triangles as [discussed
 /// here](https://iquilezles.org/articles/normals/).
 ///
-/// Question: Why double the area? Because the area of a triangle A is
-/// determined by a cross product like so:
+/// Question: Why double the area? Because the area of a triangle _A_ is
+/// determined by this equation:
 ///
-/// _A = |(b - a) x (c - a)| / 2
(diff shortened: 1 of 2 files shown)


shanecelis:
I implemented the `compute_smooth_normals_with_weights(|normal, area| 1.0)` simply to demonstrate what the visual difference is. I have not committed that function in this PR. I altered the "custom_primitives" example and make the mesh low resolution for testing and demonstration purposes for this PR. Here is an image of the variants:

normals-still

In the top left, "low poly" shows the flat normals, which are unchanged by this patch. In the top right, "explicit normals" are from the heart builder; they should be considered best. Bottom left is "unweighted"; this is what `compute_smooth_normals()` did prior to the patch. Bottom right is "weighted"; this is what `compute_smooth_normals()` does after the patch.

It is a little difficult to see any changes side to side, so I made a gif where the changes present do tend to pop out more.

normal-demo-4

For the case above, I'd judge the "weighted" image to match the "explicit normals" image slightly better than the "unweighted." The visual differences have been relatively slight.

Because most of the differences are slight though, I would not be inclined to include `compute_smooth_normals_with_weights()` as part of bevy's API.

BenjaminBrienen (approved):

crates/bevy_mesh/src/vertex.rs
```diff
@@ -138,6 +138,25 @@ pub(crate) struct MeshAttributeData {
...
+///
+/// By computing _2 A_ we avoid a division operation, and when calculating the
+/// the sum of these vectors which are then normalized, a constant multiple has
+/// no effect.
+pub(crate) fn face_area_normal(a: [f32; 3], b: [f32; 3], c: [f32; 3]) -> [f32; 3] {
+    let (a, b, c) = (Vec3::from(a), Vec3::from(b), Vec3::from(c));
+    (b - a).cross(c - a).into()
+}
+
 pub(crate) fn face_normal(a: [f32; 3], b: [f32; 3], c: [f32; 3]) -> [f32; 3] {
```
IceSentry, line 160: You could probably reuse face_area_normal here and just call .normalize() on the output.

IceSentry (approved):
I don't know how I feel about the results being slightly different. I guess it makes sense, but I'm wondering if it could break in some edge cases. At least if it breaks it's not too hard for a user to write a custom function that fixes it.

face_normal and face_area_normal could probably be pub. It doesn't hurt anything to expose that to users.

crates/bevy_mesh/src/vertex.rs
shanecelis: In the body of `face_normal()`? I thought about that, but I figured it'd be an extra function call. Maybe that's C thinking though. However, I'd have to call `Vec3::from()` again to call normalize from `face_area_normal()` since we'd outputting `[f32; 3]`.

Happy to do it if you think it's worth it.

IceSentry: Oh, right, I missed that it outputs a `[f32:3]`. Although, I wouldn't be surprised if the compiler just optimizes all of it out. Same thing for the function call. Could add an `inline` annotation if it's an issue. Doesn't really matter though, it's just a few lines so duplication is fine.

shanecelis: I'll add the inlines and make them pub.

PR #16058, opened 2024-10-22
Ensure ghost nodes are skipped when getting parent clipping rect

villor:
# Objective

- Follow up on #16044
- `extract_uinode_borders` uses `bevy_hierarchy` directly instead of going through the traversal utilities, meaning it won't handle `GhostNode`s properly. 

## Solution

- Replaced the use of `bevy_hierarchy::Parent` with `UIChildren::get_parent`

## Testing

- Ran the `overflow` example, clipping looks ok.

---

### DIFF
--- a/crates/bevy_ui/src/render/mod.rs
+++ b/crates/bevy_ui/src/render/mod.rs
@@ -4,6 +4,7 @@ mod render_pass;
 mod ui_material_pipeline;
 pub mod ui_texture_slice_pipeline;
 
+use crate::experimental::UiChildren;
 use crate::{
     BackgroundColor, BorderColor, CalculatedClip, ComputedNode, DefaultUiCamera, Outline,
     ResolvedBorderRadius, TargetCamera, UiAntiAlias, UiBoxShadowSamples, UiImage, UiScale,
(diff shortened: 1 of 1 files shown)


ickshonpe (approved):
Looks good.

Wondering if it would make sense to give all ghostnodes a `CalculatedClip` component instead and then just walk the tree normally during clipping updates, it should just work and I'd expect it to be more efficient. This should be merged though, that's just speculative etc.

viridia (approved):

BenjaminBrienen (approved):

villor:
> Wondering if it would make sense to give all ghostnodes a CalculatedClip component instead and then just walk the tree normally during clipping updates, it should just work and I'd expect it to be more efficient. This should be merged though, that's just speculative etc.

Yeah something like that could be great if the ghosts decide to stick around.  Or a more general optimization strategy like a retained (relations-based) UI node tree that can be walked like a normal hierarchy.

But seeing as ghost nodes are still experimental, and likely to be removed/re-designed if a better solution comes along, IMO it is probably better to hold off on any substantial optimizations right now.

crates/bevy_ui/src/render/mod.rs
```diff
@@ -4,6 +4,7 @@ mod render_pass;
 mod ui_material_pipeline;
 pub mod ui_texture_slice_pipeline;
 
+use crate::experimental::UiChildren;
```
mockersf, line 7: could you group this import with the one below?

villor: Done!

cart (approved):

PR #16063, opened 2024-10-22
Use CosmicFontSystem in public bevy_text APIs and remove cosmic_text re-export

### LINKED ISSUE #16006: cosmic_text Query re-export degrades autocomplete UX
>  When typing Query, RA autocomplete suggests text::cosmic_text::fontdb::Query

Reported by @JMS55 [here](https://discord.com/channels/691052431525675048/692572690833473578/1297054382084456489).

I think we should change how we're re-exporting (probably alias the offending `Query`) to avoid this problem.

cart:
# Objective

Fixes #16006

## Solution

We currently re-export `cosmic_text`, which is seemingly motivated by the desire to use `cosmic_text::FontSystem` in `bevy_text` public APIs instead of our `CosmicFontSystem` resource wrapper type.

This change makes `bevy_text` a "true" abstraction over `cosmic_text` (it in fact, was already built to be that way generally and has this one "leak").

This allows us to remove the `cosmic_text` re-export, which helps clean up the Rust Analyzer imports and generally makes this a "cleaner" API.

### DIFF
--- a/crates/bevy_text/src/lib.rs
+++ b/crates/bevy_text/src/lib.rs
@@ -45,8 +45,6 @@ mod text;
 mod text2d;
 mod text_access;
 
-pub use cosmic_text;
-
 pub use bounds::*;
 pub use error::*;
 pub use font::*;
--- a/crates/bevy_text/src/pipeline.rs
+++ b/crates/bevy_text/src/pipeline.rs
@@ -2,6 +2,7 @@ use alloc::sync::Arc;
 
 use bevy_asset::{AssetId, Assets};
 use bevy_color::Color;
+use bevy_derive::{Deref, DerefMut};
 use bevy_ecs::{
     component::Component,
     entity::Entity,
@@ -26,7 +27,7 @@ use crate::{
 /// The font system is used to retrieve fonts and their information, including glyph outlines.
 ///
 /// This resource is updated by the [`TextPipeline`] resource.
-#[derive(Resource)]
+#[derive(Resource, Deref, DerefMut)]
 pub struct CosmicFontSystem(pub cosmic_text::FontSystem);
 
 impl Default for CosmicFontSystem {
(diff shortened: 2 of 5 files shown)


nicoburns:
Huh, looking into this, it seems that there is no API for manually using a cosmic-text `Buffer` in bevy-text? Is the idea that if you don't want to use `bevy-text`s "text-as-entities" system that you'd manually pull in your own separate text layouting system?

UkoeHB:
> Is the idea that if you don't want to use bevy-texts "text-as-entities" system that you'd manually pull in your own separate text layouting system?

Yes. The default buffer cannot be publicly exposed because the contents are re-built from the ECS every time text is re-rendered. Any modifications by users will be erased.

UkoeHB (approved):

VitalyAnkh (approved):

nicoburns:
> > Is the idea that if you don't want to use bevy-texts "text-as-entities" system that you'd manually pull in your own separate text layouting system?
> 
> Yes. The default buffer cannot be publicly exposed because the contents are re-built from the ECS every time text is re-rendered. Any modifications by users would be erased.

Presumably it could be changed so that this only happens if a marker component is present or similar. Probably too late for the current release cycle, but it could be looked at in future.

nicoburns (approved):

UkoeHB:
> Presumably it could be changed so that this only happens if a marker component is present or similar.

If you need to disable all of the functionality tied to `ComputedTextBlock`, then you might as well not use `ComputedTextBlock` and just implement everything separately.

TotalKrill:
Do we really need to do this, I think its quite nice to allow a user to reach the underlying cosmic-text structures, especially while we are experimenting and wanting to create different base widgets for bevy? 

for example text-input, where there is stuff we could utilize from cosmic-text already? When bevy is mere feature complete regarding text.

TotalKrill:
but wouldnt that block external or third party crates experimenting from using the exact same version of cosmic-text that bevy is using by default?

UkoeHB:
> but wouldnt that block external or third party crates experimenting from using the exact same version of cosmic-text that bevy is using by default?

Why would it? You can look in bevy's `Cargo.toml` if you need the same version.

ickshonpe (approved):

cart:
> Huh, looking into this, it seems that there is no API for arbitrary usage of a cosmic-text Buffer in bevy-text? 

Yup. Worth exploring ways to enable this use case (later) though!

> Why would it? You can look in bevy's Cargo.toml if you need the same version.

Yeah this case is niche enough that I'm cool with having them take on this responsibility, especially given that right now there isn't a supported way for bevy_text to interact with arbitrary user provided cosmic_text ops.

PR #16071, opened 2024-10-23
Fixes for a few minor borders and outlines bugs

ickshonpe:
# Objective

1. Nodes with `Display::None` set are removed from the layout and have no position or size. Outlines should not be drawn for a node with `Display::None` set.
2. The outline and border colors are checked for transparency together. If only one of the two is transparent, both will get queued.
3. The `node.is_empty()` check is insufficient to check if a border is present since a non-zero sized node can have a zero width border.

## Solution

1. Add a check to `extract_uinode_borders` and ignore the node if `Display::None` is set.
2.  Filter the border and outline optional components by `is_fully_transparent`.
3.  Check if all the border widths are zero instead.

## Testing

I added dark cyan outlines around the left and right sections in the `display_and_visibility` example. If you run the example and set the outermost node to `Display::None` on the right, then you'll see the that the outline on the left disappears.


BenjaminBrienen (approved):

crates/bevy_ui/src/render/mod.rs
```diff
@@ -435,24 +437,22 @@ pub fn extract_uinode_borders(
             continue;
         };
 
-        // Skip invisible borders
-        if !view_visibility.get()
-            || maybe_border_color.is_some_and(|border_color| border_color.0.is_fully_transparent())
-                && maybe_outline.is_some_and(|outline| outline.color.is_fully_transparent())
-        {
+        // Skip invisible borders and removed nodes
+        if !view_visibility.get() || node.display == Display::None {
```
BenjaminBrienen, line 441: Clean 🤌

BenjaminBrienen (approved):

viridia (approved):

doup (approved):
As far as I understand, LGTM.

The example initially confused me a little bit because the outline is set with an offset so it seems it's from the white rectangle. Then I realised that the white one and the dark blue are two different nodes.

cart (approved):

Issue #15806, opened 2024-10-10
Nightly CI detects new dead code lints in benchmarks


alice-i-cecile:
https://github.com/bevyengine/bevy/actions/runs/11265648245/job/31327897506?pr=15800

alice-i-cecile:
Split it out so it doesn't get bogged down by the benchmarking please :)

Issue #15660, opened 2024-10-05
Prefer iterator over Vec::push where it makes sense to


bas-ie:
## What problem does this solve or what need does it fill?

Now-closed PR #5775 from `NathanSWard` suggests replacing many instances of `push()` in a loop with an iterator, citing efficiency.

This will not necessarily be a blanket change, as [cart points out](https://github.com/bevyengine/bevy/pull/5775#pullrequestreview-1089488861) and should be benchmarked.

The closed PR is now 2 years old and may have significant bit rot, but could be used as a starting point.

BenjaminBrienen:
This doesn't really have a defined scope. We should just improve specific examples as they are identified instead of having a generic code style enhancement as an issue.

PR #16073, opened 2024-10-23
Avoid vec push

BenjaminBrienen:
# Objective

Partially addresses #15660

I also allowed the dead code warnings that aren't meant to be fixed because it overwhelms the developer experience. One file was also missing from the module tree! I added it and updated the code to actually work. I also took the chance to format the number literals since that is probably too small of a chance to be its own PR.

I want to get some feedback before I do this in more places

## Solution

Used iter methods to avoid vec pushes.

## Testing

No testing yet

## Misc

Edit: I think that this also solves #15806

### DIFF
--- a/crates/bevy_image/src/ktx2.rs
+++ b/crates/bevy_image/src/ktx2.rs
@@ -43,20 +43,17 @@ pub fn ktx2_buffer_to_image(
     let depth = depth.max(1);
 
     // Handle supercompression
-    let mut levels = Vec::new();
-    if let Some(supercompression_scheme) = supercompression_scheme {
-        for (_level, _level_data) in ktx2.levels().enumerate() {
+    let mut levels = if let Some(supercompression_scheme) = supercompression_scheme {
+        ktx2.levels().enumerate().map(|(_level, _level_data)| {
             match supercompression_scheme {
                 #[cfg(feature = "flate2")]
                 SupercompressionScheme::ZLIB => {
                     let mut decoder = flate2::bufread::ZlibDecoder::new(_level_data);
                     let mut decompressed = Vec::new();
-                    decoder.read_to_end(&mut decompressed).map_err(|err| {
-                        TextureError::SuperDecompressionError(format!(
-                            "Failed to decompress {supercompression_scheme:?} for mip {_level}: {err:?}",
-                        ))
-                    })?;
-                    levels.push(decompressed);
+                    decoder.read_to_end(&mut decompressed).map_err(|err| TextureError::SuperDecompressionError(format!(
+                        "Failed to decompress {supercompression_scheme:?} for mip {_level}: {err:?}",
+                    )))?;
+                    Ok(decompressed)
                 }
                 #[cfg(feature = "ruzstd")]
                 SupercompressionScheme::Zstandard => {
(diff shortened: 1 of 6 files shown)


benches/benches/bevy_ecs/observers/simple.rs
```diff
@@ -28,7 +28,7 @@ pub fn observe_simple(criterion: &mut Criterion) {
     group.bench_function("trigger_targets_simple/10000_entity", |bencher| {
         let mut world = World::new();
         let mut entities = vec![];
-        for _ in 0..10000 {
+        for _ in 0..10_000 {
             entities.push(world.spawn_empty().observe(empty_listener_base).id());
```
BenjaminBrienen, line 32: Fix this one

crates/bevy_image/src/ktx2.rs
```diff
@@ -43,44 +43,47 @@ pub fn ktx2_buffer_to_image(
...
                 SupercompressionScheme::ZLIB => {
                     let mut decoder = flate2::bufread::ZlibDecoder::new(_level_data);
                     let mut decompressed = Vec::new();
-                    decoder.read_to_end(&mut decompressed).map_err(|err| {
-                        TextureError::SuperDecompressionError(format!(
-                            "Failed to decompress {supercompression_scheme:?} for mip {_level}: {err:?}",
-                        ))
-                    })?;
-                    levels.push(decompressed);
+                    match decoder.read_to_end(&mut decompressed) {
```
BenjaminBrienen, line 53: Put back as map_err

benches/benches/bevy_picking/ray_mesh_intersection.rs
```diff
@@ -14,23 +14,32 @@ struct SimpleMesh {
...
-    }
+    let normals = std::iter::repeat([0., 1., 0.])
+        .take(vertices_squared as usize)
+        .collect();
+
+    let indices = (0..vertices_squared)
+        .filter(|&p| {
+            p % vertices_per_side != vertices_per_side - 1
+                && p / vertices_per_side != vertices_per_side - 1
+        })
```
bushrat011899, line 33: I believe this filter expression removes `vertices_per_side`-number of items for `p % vertices_per_side != vertices_per_side - 1`, and another `vertices_per_side` worth in `p / vertices_per_side != vertices_per_side - 1`. Meaning the total number of items this produces (after the `flatten`) is between `6 * (vertices_squared - 2 * vertices_per_side)` and `6 * (vertices_squared - vertices_per_side)`. It _might_ be faster to allocate the `Vec` using `Vec::with_capacity(6 * (vertices_squared - vertices_per_side))` (over-estimating to avoid reallocation), since I don't believe a `collect` statement is faster when the size of the iterator isn't knowable ahead of time (which `filter` and `flat_map` can do).

Could also spend a half-hour in Desmos working out the _exact_ length of the `Vec` as a function of `vertices_per_side` to minimise memory usage, but even the overestimate _should_ perform better.

crates/bevy_reflect/src/impls/std.rs
```diff
@@ -1110,17 +1110,16 @@ where
...
+        // BTreeMap doesn't have a `drain` function, so one must be faked by popping one at a time.
+        // See https://github.com/rust-lang/rust/issues/81074
+        core::iter::from_fn(|| self.pop_first())
+            .map(|(key, value)| {
+                (
+                    Box::new(key) as Box<dyn PartialReflect>,
+                    Box::new(value) as Box<dyn PartialReflect>,
+                )
+            })
+            .collect()
```
bushrat011899, line 1122:
```suggestion
        // BTreeMap doesn't have a `drain` function. See
        // https://github.com/rust-lang/rust/issues/81074. So we have to fake one
        let mut tree = Self::new();

        core::mem::swap(self, &mut tree);

        tree.into_iter()
            .map(|(k, v)| {
                (
                    Box::new(k) as Box<dyn PartialReflect>,
                    Box::new(v) as Box<dyn PartialReflect>,
                )
            })
            .collect()
```

By `swap`-ing a new `BTreeMap` in we can just use the `IntoIterator` implementation instead, which _should_ perform better, since it _should_ be an `ExactSizeIterator`, which generally has the best `collect()` implementation. Also I personally think this is cleaner.

bushrat011899 (commented):
Some nice changes here. Just a few suggestions, but overall well done.

benches/benches/bevy_ecs/components/archetype_updates.rs
```diff
@@ -69,7 +69,7 @@ pub fn no_archetypes(criterion: &mut Criterion) {
 pub fn added_archetypes(criterion: &mut Criterion) {
     const SYSTEM_COUNT: usize = 100;
     let mut group = criterion.benchmark_group("added_archetypes");
-    for archetype_count in [100, 200, 500, 1000, 2000, 5000, 10000] {
+    for archetype_count in [100, 200, 500, 1000, 2000, 5000, 10_000] {
```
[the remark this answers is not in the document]
BenjaminBrienen: No, my goal with that is to separate groups of 4 zeroes. I'm not opposed to it, but I don't think it would improve the readability.

bushrat011899 (approved):

bushrat011899:
What's written looks good, but I agree this needs some benchmarking to determine if the change is worth it. The handful of times this PR replaced `vec![]` with `Vec::with_capacity` _should_ be a simple performance improvement (or at least have no regression), but the change from `for ... in ...` to `... take ... collect` may not be as simple to state.

mockersf:
> but the change from `for ... in ...` to `... take ... collect` may not be as simple to state.

When Rust is able to keep the iterator as an [`ExactSizeIterator`](https://doc.rust-lang.org/std/iter/trait.ExactSizeIterator.html) (starts from one, and no filters or operations like that), `collect` will be faster than a push in a for loop.

It's not so sure for cases where Rust doesn't know the size

PR #16082, opened 2024-10-24
Fix `bevy_picking` plugin suffixes

### LINKED ISSUE #16081: Some `bevy_picking` plugins do not have the `Plugin` suffix
Originally found with the `bevy::plugin_not_ending_in_plugin` lint.

There are 3 plugins introduced by `bevy_picking` that do not have a `Plugin` suffix:

1. `MeshPickingBackend`
2. `SpritePickingBackend`
3. `DefaultPickingPlugins`

1 and 2 should be renamed to follow the style of `MeshPickingBackendPlugin` or `MeshPickingPlugin`, while 3 should not even be a plugin, but instead a `PluginGroup` (like `DefaultPlugins`).

Given how all 3 of these types were added after 0.14 was released, their names are not yet public. (Excluding the release candidate.) This means that we still have a chance to rename them before they become introduced to other projects, avoiding breaking changes in the future.

BD103:
# Objective

- `MeshPickingBackend` and `SpritePickingBackend` do not have the `Plugin` suffix
- `DefaultPickingPlugins` is masquerading as a `Plugin` when in reality it should be a `PluginGroup`
- Fixes #16081.

## Solution

- Rename some structures:

|Original Name|New Name|
|-|-|
|`MeshPickingBackend`|`MeshPickingPlugin`|
|`MeshPickingBackendSettings`|`MeshPickingSettings`|
|`SpritePickingBackend`|`SpritePickingPlugin`|
|`UiPickingBackendPlugin`|`UiPickingPlugin`|

- Make `DefaultPickingPlugins` a `PluginGroup`.
- Because `DefaultPickingPlugins` is within the `DefaultPlugins` plugin group, I also added support for nested plugin groups to the `plugin_group!` macro.

## Testing

- I used ripgrep to ensure all references were properly renamed.
- For the `plugin_group!` macro, I used `cargo expand` to manually inspect the expansion of `DefaultPlugins`.

---

## Migration Guide

> [!NOTE]
>
> All 3 of the changed structures were added after 0.14, so this does not need to be included in the 0.14 to 0.15 migration guide.

- `MeshPickingBackend` is now named `MeshPickingPlugin`.
- `MeshPickingBackendSettings` is now named `MeshPickingSettings`.
- `SpritePickingBackend` is now named `SpritePickingPlugin`.
- `UiPickingBackendPlugin` is now named `UiPickingPlugin`.
- `DefaultPickingPlugins` is now a a `PluginGroup` instead of a `Plugin`.

### DIFF
--- a/crates/bevy_app/src/plugin_group.rs
+++ b/crates/bevy_app/src/plugin_group.rs
@@ -48,6 +48,17 @@ use core::any::TypeId;
 /// #   impl Plugin for WebCompatibilityPlugin { fn build(&self, _: &mut App) {} }
 /// # }
 /// #
+/// # mod audio {
+/// #   use bevy_app::*;
+/// #   #[derive(Default)]
+/// #   pub struct AudioPlugins;
+/// #   impl PluginGroup for AudioPlugins {
+/// #     fn build(self) -> PluginGroupBuilder {
+/// #       PluginGroupBuilder::start::<Self>()
+/// #     }
+/// #   }
+/// # }
+/// #
 /// # mod internal {
 /// #   use bevy_app::*;
 /// #   #[derive(Default)]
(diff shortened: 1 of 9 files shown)


cart:
I agree with adding the suffix. I'm thinking we can drop the `Backend` terminology here. A `MeshPickingPlugin` pretty unambiguously provides "mesh picking functionality". Same for `SpritePickingPlugin`. "XPickingBackend" in the context of `bevy_picking` can be interpreted as "connects the common picking infrastructure into X". But I think `XPickingPlugin` also does that job effectively.

BD103:
> I agree with adding the suffix. I'm thinking we can drop the `Backend` terminology here. A `MeshPickingPlugin` pretty unambiguously provides "mesh picking functionality". Same for `SpritePickingPlugin`. "XPickingBackend" in the context of `bevy_picking` can be interpreted as "connects the common picking infrastructure into X". But I think `XPickingPlugin` also does that job effectively.

Done!

cart (approved):

PR #16084, opened 2024-10-24
Move `UiImage` from `ui_node` to the `widget::image` module

ickshonpe:
# Objective

`UiImage` isn't just a general image component now, it's the defining component for the image widget so it belongs in the image widget's module.


alice-i-cecile (approved):
Agreed. Merging as trivial. Adding to the 0.15 milestone to make it easier to backport fixes.

PR #16086, opened 2024-10-24
Texture slicing measure func fix

ickshonpe:
# Objective

UI texture slicing chops and scales an image to fit the size of a node but because the required components changes required `ImageSize` and `ContentSize` for nodes with `UiImage`, texture sliced nodes are laid out using an `ImageMeasure`.

## Solution

Add `Without<ImageScaleMode>` to `UpdateImageFilter` so the measure func isn't set.

## Testing

Remove all the constraints from the `ui_texture_slice` example:

```rust
//! This example illustrates how to create buttons with their textures sliced
//! and kept in proportion instead of being stretched by the button dimensions

use bevy::{
    color::palettes::css::{GOLD, ORANGE},
    prelude::*,
    winit::WinitSettings,
};

fn main() {
    App::new()
        .add_plugins(DefaultPlugins)
        // Only run the app when there is user input. This will significantly reduce CPU/GPU use.
        .insert_resource(WinitSettings::desktop_app())
        .add_systems(Startup, setup)
        .add_systems(Update, button_system)
        .run();
}

fn button_system(
    mut interaction_query: Query<
        (&Interaction, &Children, &mut UiImage),
        (Changed<Interaction>, With<Button>),
    >,
    mut text_query: Query<&mut Text>,
) {
    for (interaction, children, mut image) in &mut interaction_query {
        let mut text = text_query.get_mut(children[0]).unwrap();
        match *interaction {
            Interaction::Pressed => {
                **text = "Press".to_string();
                image.color = GOLD.into();
            }
            Interaction::Hovered => {
                **text = "Hover".to_string();
                image.color = ORANGE.into();
            }
            Interaction::None => {
                **text = "Button".to_string();
                image.color = Color::WHITE;
            }
        }
    }
}

fn setup(mut commands: Commands, asset_server: Res<AssetServer>) {
    let image = asset_server.load("textures/fantasy_ui_borders/panel-border-010.png");

    let slicer = TextureSlicer {
        border: BorderRect::square(22.0),
        center_scale_mode: SliceScaleMode::Stretch,
        sides_scale_mode: SliceScaleMode::Stretch,
        max_corner_scale: 1.0,
    };
    // ui camera
    commands.spawn(Camera2d);
    commands
        .spawn(Node {
            width: Val::Percent(100.0),
            height: Val::Percent(100.0),
            align_items: AlignItems::Center,
            justify_content: JustifyContent::Center,
            ..default()
        })
        .with_children(|parent| {
[...]

PR #16088, opened 2024-10-24
Add image scale mode to uiimage

ickshonpe:
# Objective

UI texture slicing chops and scales an image to fit the size of a node but because the required components changes required ImageSize and ContentSize for nodes with UiImage, texture sliced nodes are laid out using an ImageMeasure.

Alternative implementation of #16086.

This doesn't feel great to me, we end up with a lot of systems iterating through entities they ignore, rather than never pulling them in in the query. It does seem to fit in better with the direction of the other `UiImage` changes though and it makes the feature much more discoverable. It wasn't documented anywhere except in the examples that `UiImage`s could be composed with `ImageScaleMode` which I guess was probably why this error got introduced.

## Solution

Add a `scale_mode: ImageScaleMode` field to `UiImage`. 
Remove queries for `ImageScaleMode` from UI extraction, and refer to the field instead.
Don't set a measure func in `update_image_content_size_system` if `scale_mode` is some.

## Testing
Remove all the constraints from the ui_texture_slice example:

//! This example illustrates how to create buttons with their textures sliced
//! and kept in proportion instead of being stretched by the button dimensions

use bevy::{
    color::palettes::css::{GOLD, ORANGE},
    prelude::*,
    winit::WinitSettings,
};

fn main() {
    App::new()
        .add_plugins(DefaultPlugins)
        // Only run the app when there is user input. This will significantly reduce CPU/GPU use.
        .add_systems(Startup, setup)
        .add_systems(Update, button_system)
        .run();
}

fn button_system(
    mut interaction_query: Query<
        (&Interaction, &Children, &mut UiImage),
        (Changed<Interaction>, With<Button>),
    >,
    mut text_query: Query<&mut Text>,
) {
    for (interaction, children, mut image) in &mut interaction_query {
        let mut text = text_query.get_mut(children[0]).unwrap();
        match *interaction {
            Interaction::Pressed => {
                **text = "Press".to_string();
                image.color = GOLD.into();
            }
            Interaction::Hovered => {
                **text = "Hover".to_string();
                image.color = ORANGE.into();
            }
            Interaction::None => {
                **text = "Button".to_string();
                image.color = Color::WHITE;
            }
        }
    }
}

    let image = asset_server.load("textures/fantasy_ui_borders/panel-border-010.png");

    let slicer = TextureSlicer {
        border: BorderRect::square(22.0),
        center_scale_mode: SliceScaleMode::Stretch,
        sides_scale_mode: SliceScaleMode::Stretch,
        max_corner_scale: 1.0,
    };
    // ui camera
    commands.spawn(Camera2d);
    commands
        .spawn(Node {
            width: Val::Percent(100.0),
            height: Val::Percent(100.0),
            ..default()
        })
        .with_children(|parent| {
            for [w, h] in [[150.0, 150.0], [300.0, 150.0], [150.0, 300.0]] {
                parent
                    .spawn((
                        Button,
                        Node {
                            // width: Val::Px(w),
                            // height: Val::Px(h),
                            // horizontally center child text
                            // vertically center child text
                            margin: UiRect::all(Val::Px(20.0)),
                            ..default()
                        },
                        UiImage::new(image.clone()),
                        ImageScaleMode::Sliced(slicer.clone()),
                    ))
                    .with_children(|parent| {
                        // parent.spawn((
                        //     Text::new("Button"),
                        //     TextFont {
                        //         font: asset_server.load("fonts/FiraSans-Bold.ttf"),
                        //         font_size: 33.0,
                        //         ..default()
                        //     },
                        //     TextColor(Color::srgb(0.9, 0.9, 0.9)),
                        // ));
                    });
            }
        });
}
This should result in a blank window, since without any constraints the texture slice image nodes should be zero sized. But instead we see three nodes:

slicing
With this PR the window is blank.

### DIFF
--- a/crates/bevy_ui/src/render/mod.rs
+++ b/crates/bevy_ui/src/render/mod.rs
@@ -308,18 +308,15 @@ pub fn extract_uinode_images(
     texture_atlases: Extract<Res<Assets<TextureAtlasLayout>>>,
     default_ui_camera: Extract<DefaultUiCamera>,
     uinode_query: Extract<
-        Query<
-            (
-                Entity,
-                &ComputedNode,
-                &GlobalTransform,
-                &ViewVisibility,
-                Option<&CalculatedClip>,
-                Option<&TargetCamera>,
-                &UiImage,
-            ),
-            Without<ImageScaleMode>,
-        >,
+        Query<(
+            Entity,
+            &ComputedNode,
+            &GlobalTransform,
+            &ViewVisibility,
+            Option<&CalculatedClip>,
+            Option<&TargetCamera>,
+            &UiImage,
+        )>,
     >,
     mapping: Extract<Query<RenderEntity>>,
 ) {
@@ -337,6 +334,7 @@ pub fn extract_uinode_images(
         if !view_visibility.get()
             || image.color.is_fully_transparent()
             || image.image.id() == TRANSPARENT_IMAGE_HANDLE.id()
+            || image.scale_mode.is_some()
         {
             continue;
         }
--- a/crates/bevy_ui/src/render/mod.rs
+++ b/crates/bevy_ui/src/render/mod.rs
@@ -40,7 +40,7 @@ use bevy_render::{
     ExtractSchedule, Render,
 };
 use bevy_sprite::TextureAtlasLayout;
-use bevy_sprite::{BorderRect, ImageScaleMode, SpriteAssetEvents};
+use bevy_sprite::{BorderRect, SpriteAssetEvents};
 
 use crate::{Display, Node};
 use bevy_text::{ComputedTextBlock, PositionedGlyph, TextColor, TextLayoutInfo};
--- a/crates/bevy_ui/src/render/ui_texture_slice_pipeline.rs
+++ b/crates/bevy_ui/src/render/ui_texture_slice_pipeline.rs
@@ -256,20 +256,21 @@ pub fn extract_ui_texture_slices(
             Option<&CalculatedClip>,
             Option<&TargetCamera>,
             &UiImage,
-            &ImageScaleMode,
         )>,
     >,
     mapping: Extract<Query<RenderEntity>>,
 ) {
-    for (entity, uinode, transform, view_visibility, clip, camera, image, image_scale_mode) in
-        &slicers_query
-    {
+    for (entity, uinode, transform, view_visibility, clip, camera, image) in &slicers_query {
         let Some(camera_entity) = camera.map(TargetCamera::entity).or(default_ui_camera.get())
         else {
             continue;
         };
 
         let Ok(camera_entity) = mapping.get(camera_entity) else {
+            continue;
+        };
+
+        let Some(ref image_scale_mode) = image.scale_mode else {
             continue;
         };
 
--- a/crates/bevy_ui/src/ui_node.rs
+++ b/crates/bevy_ui/src/ui_node.rs
@@ -9,7 +9,7 @@ use bevy_render::{
     texture::{Image, TRANSPARENT_IMAGE_HANDLE},
     view::Visibility,
 };
-use bevy_sprite::{BorderRect, TextureAtlas};
+use bevy_sprite::{BorderRect, ImageScaleMode, TextureAtlas};
 use bevy_transform::components::Transform;
 use bevy_utils::warn_once;
 use bevy_window::{PrimaryWindow, WindowRef};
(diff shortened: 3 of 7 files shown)


ickshonpe:
This  doesn't fix the problems with measure func removal, going to look at that next.

cart:
I'm on board for the general angle, but I think we should take this a couple of steps farther:

1. Port this pattern to Sprite for consistency.
2. Remove the `Component` impl for `ImageScaleMode`.
3. Make the "default" / `None` scale mode an `ImageScaleMode` variant (ex: `ImageScaleMode::Standard`). This makes the ergonomics better and also allows us to more easily express ideas like `scale_mode.defines_size()` (Standard) and `scale_mode.uses_slices()` (Sliced and Tiled). This also opens the doors to new scale modes in the future that aren't "sliced".

cart:
I've put those changes together. Pushing now for consideration (not dictating this solution, just pushing for convenience of discussion).

ickshonpe:
> I'm on board for the general angle, but I think we should take this a couple of steps further:
> 
> 1. Port this pattern to Sprite for consistency.
> 2. Remove the `Component` impl for `ImageScaleMode`.
> 3. Make the "default" / `None` scale mode an `ImageScaleMode` variant (ex: `ImageScaleMode::Standard`). This makes the ergonomics better and also allows us to more easily express ideas like `scale_mode.defines_size()` (Standard) and `scale_mode.uses_slices()` (Sliced and Tiled). This also opens the doors to new scale modes in the future that aren't "sliced".

Yeah I've just been working on this PR the last hour as well and didn't notice this until I came to merge my changes. But I've been working in the same the direction, agree that `ImageScaleMode` should be non-optional and that it needs another variant for the default behaviour.

The main difference though is that I don't think `Sprite` and `UiImage` can share the `ImageScaleMode` enum since a `UiImage` also has to deal with the constraints on its size imposed by the UI layout. Initially I think we need at least two extra variants 
one that uses a measure func to preserve the images aspect ratio and one that just fills the node without applying any constraints.

ickshonpe:
The name `ImageScaleMode` seems misleading as well for the same reasons, for `UiImage`s since it wouldn't just control how the image is scaled but also how the layout algorithm rearranges the layout to make room for the image.

cart:
> The main difference though is that I don't think Sprite and UiImage can share the ImageScaleMode enum

Yeah a split here seems reasonable. Also open to name alternatives.

ickshonpe:
Merged my changes, called the new enum `UiImageMode` which I kind of hate but I can't think of anything right now that's more specific and isn't super long or potentially confusing. The `Auto` and `Resize` variant names seem bad too.

ickshonpe:
Renamed both `Resize` and `Standard` to `Stretch`, `Standard` and its comment seemed misleading because what it really means is that the image is resized if a `custom_size` is set.

cart:
I like these changes. Perhaps we should call these `SpriteImageMode` and `NodeImageMode` for unified naming conventions?

ImageScaleMode feels "too generic" given that it is really in the "sprite scope" and UiImageMode also feels too generic, as this is really the scale mode for nodes specifically.

It feels like in the current form sprite is really in `Auto` by default, given that it takes on the size of the image by default (just like `UiImageMode::Auto`) and then _implicitly_ switches to Stretch if a custom size is set (which feels like an "auto" behavior to me).

ickshonpe:
> It feels like in the current form sprite is really in `Auto` by default, given that it takes on the size of the image by default (just like 
`UiImageMode::Auto`) and then _implicitly_ switches to Stretch if a custom size is set (which feels like an "auto" behavior to me).

Oh I know, we could get rid of the `custom_size` field completely, and add instead a `CustomSize(Vec2)` variant to the `SpriteImageMode` instead. The `Tiled` and `Sliced` variants would need their own `size` field too but it never really made sense to use them with the base image size anyway.

ickshonpe:
> Perhaps we should call these `SpriteImageMode` and `NodeImageMode` for unified naming conventions?

Yep seem good to me.

cart:
> We could get rid of the custom_size field and replace it with a CustomSize(Vec2) variant on the SpriteImageMode enum instead? The Tiled and Sliced variants would need their own size fields too but it doesn't really make sense to use them with the base image size anyway.

Hmm the common `custom_size` field does make it easier to operate on "sprite custom size" generically without needing to know the exact variant it is currently in. And adding it to the variants would make then make the sprite vs node interface different.

I kind of still like just doing `Stretch -> Auto`, especially given that it makes this a less invasive change. But I also don't have _super_ strong opinions here.

cart:
I also think the field should be `Sprite::image_mode` not `Sprite::mode` , otherwise the variant should be `SpriteMode` and `NodeMode`, but imo thats a bit too generic (in both cases).

cart:
(just pushed that change)

cart (approved):

PR #16091, opened 2024-10-24
Make some associated functions of `Color` const

mamekoro:
# Objective
Make the following functions `const` that will be useful to define colors as constants.

- `Color::srgb_from_array`
- `Color::srgba_u8`
- `Color::srgb_u8`

The last two require Rust 1.82.0.

## Solution
- Make them `const`
- Change MSRV to 1.82.0

## Testing
I tested bevy_color only. My machine does not have enough RAM capacity to test the whole bevy.

`cargo test -p bevy_color`


ickshonpe (approved):
Don't see any problems.

crates/bevy_color/src/color.rs
```diff
@@ -143,7 +143,7 @@ impl Color {
     /// Creates a new [`Color`] object storing a [`Srgba`] color from [`u8`] values.
     ///
     /// A value of 0 is interpreted as 0.0, and a value of 255 is interpreted as 1.0.
-    pub fn srgba_u8(red: u8, green: u8, blue: u8, alpha: u8) -> Self {
+    pub const fn srgba_u8(red: u8, green: u8, blue: u8, alpha: u8) -> Self {
```
ickshonpe, line 146: Won't work because of floating arithmetic.

crates/bevy_color/src/color.rs
```diff
@@ -163,7 +163,7 @@ impl Color {
     /// Creates a new [`Color`] object storing a [`Srgba`] color from [`u8`] values with an alpha of 1.0.
     ///
     /// A value of 0 is interpreted as 0.0, and a value of 255 is interpreted as 1.0.
-    pub fn srgb_u8(red: u8, green: u8, blue: u8) -> Self {
+    pub const fn srgb_u8(red: u8, green: u8, blue: u8) -> Self {
         Self::Srgba(Srgba {
             red: red as f32 / 255.0,
```
ickshonpe, line 168: Same here.

ickshonpe (changes requested):
I think `srgb_from_arrray` can be made const but not the other two.

crates/bevy_color/src/color.rs
alice-i-cecile: const floating point arithmetic just stabilized :)

ickshonpe: ohh I see, I was confused by how the author's bevy_color tests could be passing locally.

mamekoro: I had forgotten about Bevy's MSRV.
Should I also change the value of `package.rust-version` in Cargo.toml to `1.82.0` with this pull request?

alice-i-cecile: Yes, please do.

alice-i-cecile (approved):

PR #16093, opened 2024-10-24
Correctly feature gate `custom_cursor`

Friz64:
# Objective

Currently there's no way to change the window's cursor icon with the `custom_cursor` feature **disabled**. You should still be able to set system cursor icons.

Connections:

- https://github.com/bevyengine/bevy/pull/15649

## Solution

Move some `custom_cursor` feature gates around, as to expose the `CursorIcon` type again.

Note this refactoring was mainly piloted by hunting after the compiler warnings -- I shouldn't have missed anything, but FYI.

## Testing

Disabled the `custom_cursor` feature, ran the `window_settings` example.

### DIFF
--- a/crates/bevy_winit/src/converters.rs
+++ b/crates/bevy_winit/src/converters.rs
@@ -6,7 +6,6 @@ use bevy_input::{
     ButtonState,
 };
 use bevy_math::{CompassOctant, Vec2};
-#[cfg(feature = "custom_cursor")]
 use bevy_window::SystemCursorIcon;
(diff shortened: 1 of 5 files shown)


Friz64:
Context: I encountered this while porting bevy_egui to the release candidate. This should be fixed before 0.15.

BenjaminBrienen (approved):

tychedelia (approved):
Thanks for catching this, would have been a bummer to ship on 0.15.

PR #16095, opened 2024-10-25
Native unclipped depth on supported platforms

### LINKED ISSUE #16078: Use native depth clip control instead of manual DEPTH_CLAMP_ORTHO where possible

The DEPTH_CLAMP_ORTHO stuff we have for directional light shadow mapping uses a fragment shader. Having the fragment shader disables some really important hardware optimizations that exist for depth-only vertex-only passes.

For non-alpha masked draws, on supported platforms, use https://docs.rs/wgpu/latest/wgpu/struct.PrimitiveState.html#structfield.unclipped_depth instead of setting the DEPTH_CLAMP_ORTHO shader def.

Additionally when we do use DEPTH_CLAMP_ORTHO, we should not pass the entire `clip_position_unclamped: vec4<f32>` from the vertex shader to the fragment shader, we only need to pass the `z` value.

We could remove DEPTH_CLAMP_ORTHO entirely, and instead expand the shadow view frustum either by some fixed amount, or based on the AABB's of entities in the scene.

## Additional context

https://github.com/bevyengine/bevy/pull/8877
https://therealmjp.github.io/posts/shadow-maps/#disabling-z-clipping

JMS55:
# Objective
- Fixes #16078

## Solution

- Rename things to clarify that we _want_ unclipped depth for directional light shadow views, and need some way of disabling the GPU's builtin depth clipping
- Use DEPTH_CLIP_CONTROL instead of the fragment shader emulation on supported platforms
  - Pass only the clip position depth instead of the whole clip position between vertex->fragment shader (no idea if this helps performance or not, compiler might optimize it anyways)
- Meshlets
  - HW raster always uses DEPTH_CLIP_CONTROL since it targets a more limited set of platforms
  - SW raster was not handling DEPTH_CLAMP_ORTHO correctly, it ended up pretty much doing nothing.
  - This PR made me realize that SW raster technically should have depth clipping for all views that are not directional light shadows, but I decided not to bother writing it. I'm not sure that it ever matters in practice. If proven otherwise, I can add it.

## Testing

  - Lighting example. Both opaque (no fragment shader) and alpha masked geometry (fragment shader emulation) are working with depth_clip_control, and both work when it's turned off. Also tested meshlet example.
  - Performance. I can't figure out a good test scene.
  - Toggle depth_clip_control_supported in prepass/mod.rs line 323 to turn this PR on or off.
  - Native

---

## Migration Guide
- `MeshPipelineKey::DEPTH_CLAMP_ORTHO` is now `MeshPipelineKey::UNCLIPPED_DEPTH_ORTHO`
- The `DEPTH_CLAMP_ORTHO` shaderdef has been renamed to `UNCLIPPED_DEPTH_ORTHO_EMULATION`
- `clip_position_unclamped: vec4<f32>` is now `unclipped_depth: f32`


atlv24 (approved):

pcwalton (approved):
Awesome!

crates/bevy_pbr/src/prepass/mod.rs
```diff
@@ -328,7 +334,7 @@ where
 
     fn specialize(
         &self,
-        key: Self::Key,
+        mut key: Self::Key,
```
superdump, line 337: Making this mutable looks suspicious to me because the specialisation key is used with the pipeline caching _before_ specialisation is done? But I’d have to re-read the code around it.

JMS55: I think the code is what I wanted, it's just a bit confusing now that you mention it.

We _do_ want DEPTH_CLAMP_ORTHO to be in the key, which tells us that we want unclipped depth.

But after the key is set, we _may or may not_ want the DEPTH_CLAMP_ORTHO shader define in the prepass shader, depending on whether or not the GPU natively supports unclipped depth.

Modifying the key in this function is just so that the rest of the logic in it skips the DEPTH_CLAMP_ORTHO shader workaround we have.

I think I'll rename DEPTH_CLAMP_ORTHO to UNCLIPPED_DEPTH_ORTHO or something.

superdump: Right, good point about the key field expressing the user intent, but the possible problem is that other specialisation functions will build on the pipeline descriptor returned from this and they will not see the change to the key. Maybe it doesn’t matter in practice currently but it feels like it could. So what I mean is - maybe another layer of specialisation looks at the key, sees depth clamp ortho is set, and does something based on that assuming the current behaviour (say, presence of the interpolated unclamped position as fragment shader input)?

superdump: I just saw the changed version. That indeed looks like a good approach to me - differentiate between the intent in the key and the intent in the shader def. Good stuff. This would need a migration doc comment in the description for the change in shader key and shader def. Which is a good thing in this case. :)

JMS55: Yeah I'll write a migration guide once I've tested perf and made sure I didn't break anything.

Glad that the new version is clearer.

JMS55:
Technically I think I should have a version of the meshlet software raster that does depth clipping for all other views besides directional light shadows, but I'm not 100% sure if this even matters in practice so I've elected to just ignore it for now.

JMS55:
Ready for review. Please post perf tests, I was unable to find a good comparison. Also please double check that the fragment shader is actually omitted when the emulation is off, renderdoc was confusing when I tried to check.

JMS55:
@superdump when we set the frag depth to the unclipped depth, should we be writing 1 / unclipped depth or something? Is what we have correct?

BenjaminBrienen (approved):

superdump:
Running `many_cubes --benchmark --shadows` on M1 Max `main` (yellow) vs this PR (red):
Screenshot 2024-11-18 at 11 52 42
This PR looks slightly slower. I did check that depth clip control was supported. Seems odd...

JMS55:
:/ yeah that's what I saw... Confusing.

Elabajaba:
Noticeably faster on the GPU for shadow passes, though overall framerate doesn't improve due to still being CPU bound.
image

I'd expect a bigger win on iGPUs and mobile GPUs where you're way more bandwidth limited than a 6800xt at 720p lol. Number of shaded pixels dropped from 68,743,094 to 17,371,373 according to radeon gpu profiler.

Tested on bistro in scene_viewer with vsync disabled, winit update modes set to continuous, depth prepass added to the camera, camera transform `Transform::from_translation(Vec3::new(-17.654163, 2.798615, -6.699705)).with_rotation(Quat::from_slice(&[0.0038612718, -0.7603651, 0.0045205816, 0.6494687])),`

JMS55:
Ok, so it doesn't improve FPS because we're still CPU bound, but on scenes with a lot of shadow fragments, we _are_ faster now. I think we're good to merge :)

PR #16096, opened 2024-10-25
Layout rounding debug example

ickshonpe:
# Objective

Simple example for debugging layout rounding errors.

layout_rounding_debug

Any white lines are gaps in the layout caused by coordinate rounding errors.

PR #16097, opened 2024-10-25
Remove custom rounding

### LINKED ISSUE #15197: Delay in text-update vs layout-update, creating wiggling text
## Bevy version

Using main to dev (`1fd478277`)

##  Relevant system information

- cargo 1.80.1 (376290515 2024-07-16)
- Linux terska-2k24 6.8.0-44-generic 44-Ubuntu SMP PREEMPT_DYNAMIC Tue Aug 13 13:35:26 UTC 2024 x86_64 x86_64 x86_64 GNU/Linux

## What you did

Started the _text_debug_ example.

## What went wrong

Texts that are on right wiggle one pixel back and forth horizontally.

## Additional information

Not sure if it is related to the texts at the bottom that move around a lot because of the FPS and frame time numbers. You would think the text in the top right would stay still but it still wiggles.

**Note**: You can not see the fps/frame time text until #15190 has been merged because they are pushed out of frame unless you make the window larger.

ickshonpe:
# Objective

Taffy added layout rounding a while ago but it had a couple of bugs and caused some problems with the fussy `ab_glyph` text implementation. So I disabled Taffy's builtin rounding and added some hacks ad hoc that fixed (some) of those issues. Since then though Taffy's rounding algorithm has improved while we've changed layout a lot and migrated to `cosmic-text` so those hacks don't help any more and in some cases cause significant problems.

Also our rounding implementation only rounds to the nearest logical pixel, whereas Taffy rounds to the nearest physical pixel meaning it's much more accurate with high dpi displays.

fixes #15197

## Some examples of layout rounding errors visible in the UI examples

These errors are much more obvious at high scale factor, you might not see any problems at a scale factor of 1.

`cargo run --example text_wrap_debug`

text_debug_gaps

The narrow horizontal and vertical lines are gaps in the layout caused by errors in the coordinate rounding.

`cargo run --example text_debug`

text_debug

The two text blocks here are aligned right to the same boundary but in this screen shot you can see that the lower block is one pixel off to the left. Because the size of this text node changes between frames with the reported framerate the rounding errors cause it to jump left and right.

## Solution

Remove all our custom rounding hacks and reenable Taffy's layout rounding.

The gaps in the `text_wrap_debug` example are gone:
text_wrap_debug_fix

This doesn't fix some of the gaps that occur between borders and content but they seem appear to be a rendering problem as they disappear with `UiAntiAlias::Off` set.

## Testing

Run the examples as described above in the `Objective` section. With this PR the problems mentioned shouldn't appear.

Also added an example in a separate PR #16096 `layout_rounding_debug` for identifying these issues.

## Migration Guide

`UiSurface::get_layout` now also returns the final sizes before rounding. Call `.0` on the `Ok` result to get the previously returned `taffy::Layout` value.


nicoburns (approved):
Makes sense to me

ickshonpe:
`ui_rounding_test` was failing after the layout rounding changes but only fails at extremely low scale factors (below ~0.14). At that scale factor with 7 logical pixels to 1 physical pixel not much is going to visable anyway, so I don't think it's worth worrying about. I set the test to start at `1 - 1/n` instead of `1/n` scale factor and now it passes.

crates/bevy_ui/src/layout/mod.rs
```diff
@@ -1123,7 +1088,7 @@ mod tests {
             .collect::<Vec<Entity>>();
 
         for r in [2, 3, 5, 7, 11, 13, 17, 19, 21, 23, 29, 31].map(|n| (n as f32).recip()) {
-            let mut s = r;
+            let mut s = 1. - r;
```
rparrett, line 1091:
```suggestion
            // This fails with very small / unrealistic scale values
            let mut s = 0.2 + r;
```

Tiny suggestion for the test.

alice-i-cecile:
@ickshonpe CI is failing on a related test: can you take a look? I *suspect* the test itself is wrong.

ickshonpe:
Yep I'll fix it now

alice-i-cecile (approved):

crates/bevy_ui/src/layout/mod.rs
```diff
@@ -351,12 +351,10 @@ with UI components as a child of an entity without UI components, your UI layout
 
             absolute_location += layout_location;
 
-            let rounded_size = approx_round_layout_coords(absolute_location + layout_size)
-                - approx_round_layout_coords(absolute_location);
+            let rounded_size = (absolute_location + layout_size) - (absolute_location);
```
BenjaminBrienen, line 354: Parentheses are redundant

BenjaminBrienen (approved):

alice-i-cecile (approved):

alice-i-cecile:
Removing the unrounded size is a breaking change, but I'm okay if the migration guide is just "please tell us why you wanted this".

BenjaminBrienen (approved):

alice-i-cecile:
@ickshonpe CI failure is pointing out a real, if minor issue :) Ping me once it's resolved!

ickshonpe:
> Removing the unrounded size is a breaking change, but I'm okay if the migration guide is just "please tell us why you wanted this".

Couldn't decide what to say in the migration guide, so went back and worked out a really wretched way to retrieve the unrounded size from Taffy. I guess it's better this way though, there don't seem to be any problems anymore but it feels safer to stick with using the unrounded size for text.

ickshonpe:
> @ickshonpe CI failure is pointing out a real, if minor issue :) Ping me once it's resolved!

Fixed

crates/bevy_ui/src/layout/ui_surface.rs
```diff
@@ -276,14 +275,23 @@ impl UiSurface {
...
     /// Does not compute the layout geometry, `compute_window_layouts` should be run before using this function.
-    pub fn get_layout(&self, entity: Entity) -> Result<&taffy::Layout, LayoutError> {
-        if let Some(taffy_node) = self.entity_to_taffy.get(&entity) {
-            self.taffy
-                .layout(*taffy_node)
-                .map_err(LayoutError::TaffyError)
-        } else {
-            Err(LayoutError::InvalidHierarchy)
-        }
+    pub fn get_layout(&mut self, entity: Entity) -> Result<(taffy::Layout, Vec2), LayoutError> {
```
alice-i-cecile, line 278: Can you document the meaning of the new `Vec2` here? Or replace the returned tuple with a real type.

alice-i-cecile (changes requested):
Small docs and/or quality request, then this LGTM :)

alice-i-cecile (approved):

crates/bevy_ui/src/layout/mod.rs
```diff
@@ -955,11 +914,12 @@ mod tests {
                 .get_single(world)
                 .expect("missing MovingUiNode");
             assert_eq!(expected_camera_entity, target_camera_entity);
-            let ui_surface = world.resource::<UiSurface>();
+            let mut ui_surface = world.resource_mut::<UiSurface>();
 
             let layout = ui_surface
                 .get_layout(ui_node_entity)
-                .expect("failed to get layout");
+                .expect("failed to get layout")
```
BenjaminBrienen, line 921: `expect()` here

crates/bevy_ui/src/layout/mod.rs
```diff
@@ -1043,8 +1003,8 @@ mod tests {
 
         ui_schedule.run(&mut world);
 
-        let ui_surface = world.resource::<UiSurface>();
-        let layout = ui_surface.get_layout(ui_entity).unwrap();
+        let mut ui_surface = world.resource_mut::<UiSurface>();
+        let layout = ui_surface.get_layout(ui_entity).unwrap().0;
```
BenjaminBrienen, line 1007: `unwrap()` here

BenjaminBrienen (approved):
I noticed a small inconsistency with unwrap vs expect in the tests, but other than that, still good.

PR #16102, opened 2024-10-25
Cosmetic tweaks to `query_gltf_primitives`

rparrett:
# Objective

This example is really confusing to look at and tell at a glance whether it's broken or not.

It's displaying a strange shape -- a cube with two vertices stretched in a couple dimensions at an odd angle, and doing its vertex position modification in a way where the intent isn't obvious.

## Solution

- Change the gltf geometry so that the object is a recognizable regular shape
- Change the vertex modification so that the entire cube top is being "lifted" from the cube
- Adjust colors, lighting, and camera location so we can see what's going on
- Also remove some irrelevant shadow and environment map setup

## Before

Image

## After

image


BenjaminBrienen (approved):

alice-i-cecile (approved):
Lovely, thanks for all the care you put into our examples!

PR #16103, opened 2024-10-25
Window picking

NthTensor:
# Objective

On the web, it's common to attach observers to windows. As @viridia has discovered, this can be quite a nice paradigm in bevy as well when applied to observers. The changes here are intended to make this possible.
+ Adds a new default picking back-end as part to the core picking plugin (which can be disabled) that causes pointers on windows to treat the window entity as the final hit, behind everything else. This means clicking empty space now dispatches normal picking events to the window, and is especially nice for drag-and-drop functionality.
+ Adds a new traversal type, specific to picking events, that causes them to bubble up to the window entity after they reach the root of the hierarchy.

## Solution

The window picking back-end is extremely simple, but the bubbling changes are much more complex, since they require doing a different traversal depending on the picking event. 

To achieve this, `Traversal` has been made generic over an associated sized data type `D`. Observer bounds have been changed such that `Event::Traversal<D>` is required for `Trigger<D>`. A blanket implementation has been added for `()` and `Parent` that preserves the existing functionality. A new `PointerTraversal` traversal has been implemented, with a blanket implementation for `Traversal<Pointer<E>>`.

It is still possible to use `Parent` as the traversal for any event, because of the blanket implementation. It is now possible for users to add other custom traversals, which read event data during traversal.

## Testing

I tested these changes locally on some picking UI prototypes I have been playing with. I also tested them on the picking examples.


viridia (approved):

BenjaminBrienen (approved):

crates/bevy_ecs/src/traversal.rs
```diff
@@ -13,16 +13,20 @@ use crate::{entity::Entity, query::ReadOnlyQueryData};
 /// for documenting possible looping behavior, and consumers of those implementations are responsible for
 /// avoiding infinite loops in their code.
 ///
+/// Traversials may be parameterized with additional data. For example, in observer event propagation, the
```
mnmaita, line 16:
```suggestion
/// Traversals may be parameterized with additional data. For example, in observer event propagation, the
```
typo?

PR #16105, opened 2024-10-25
Emit picking event streams

NthTensor:
# Objective

In `bevy_mod_picking` events are accessible through event listeners or `EventReader`s. When I replaced event listeners with observers, I removed the `EventReader` for simplicity. This adds it back.

## Solution

All picking events are now properly registered, and can be accessed through `EventReader<Pointer<E>>`. `Pointer` now tracks the entity the event targeted initially, and this can also be helpful in observers (which don't currently do this).

## Testing

The picking examples run fine. This shouldn't really change anything.

### DIFF
--- a/crates/bevy_picking/src/events.rs
+++ b/crates/bevy_picking/src/events.rs
@@ -39,7 +39,7 @@
 
 use core::fmt::Debug;
 
-use bevy_ecs::prelude::*;
+use bevy_ecs::{prelude::*, system::SystemParam};
 use bevy_hierarchy::Parent;
 use bevy_math::Vec2;
 use bevy_reflect::prelude::*;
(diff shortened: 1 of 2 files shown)


crates/bevy_picking/src/events.rs
```diff
@@ -60,6 +60,8 @@ use crate::{
 #[derive(Clone, PartialEq, Debug, Reflect, Component)]
 #[reflect(Component, Debug)]
 pub struct Pointer<E: Debug + Clone + Reflect> {
+    /// The original target of this picking event, before bubling
```
aevyrie, line 63:
```suggestion
    /// The original target of this picking event, before bubbling
```

aevyrie (approved):

BenjaminBrienen (approved):

alice-i-cecile (approved):
I'm fine with providing this, especially as a migration affordance. I'd like a clearer picture of what events we send as observers-events and what we send as buffered-events cohesively, but that's a cross-cutting concern.

PR #16108, opened 2024-10-26
More `#[doc(fake_variadic)]` goodness

bash:
This PR adds `#[doc(fake_variadic)]` to that were previously not supported by rustdoc.

Thanks to an [upstream contribution](https://github.com/rust-lang/rust/pull/132115) by yours truly, `#[doc(fake_variadic)]` is now supported on impls such as `impl QueryData for AnyOf<(T, ...)>` 🎉 
Requires the latest nightly compiler (2024-10-25) which is already available on [docs.rs](https://docs.rs/about/builds).

image
image

This means that the impl sections for `QueryData` and `QueryFilter` are now nice and tidy ✨ 

---

I also added `fake_variadic` to some impls that use `all_tuples_with_size`, however I'm not entirely happy because the docs are slightly misleading now:

image

Note that the docs say `IntoBindGroupLayoutEntryBuilderArray<1>` instead of
`IntoBindGroupLayoutEntryBuilderArray<N>`.

### DIFF
--- a/crates/bevy_ecs/src/query/fetch.rs
+++ b/crates/bevy_ecs/src/query/fetch.rs
@@ -2030,8 +2030,8 @@ macro_rules! impl_tuple_query_data {
 }
 
 macro_rules! impl_anytuple_fetch {
-    ($(($name: ident, $state: ident)),*) => {
-
+    ($(#[$meta:meta])* $(($name: ident, $state: ident)),*) => {
+        $(#[$meta])*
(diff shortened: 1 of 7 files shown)


janhohenheim:
Wow, you added that into `rustdoc` itself? Awesome :D

BenjaminBrienen (approved):

alice-i-cecile (approved):

PR #16110, opened 2024-10-26
Mesh picking fixes

### LINKED ISSUE #16065: Excessive logging from new `mesh_picking` backend
**Note: This is in regard to the new `0.15.0-rc.1` release candidate.**
I understand that this is a pre-release, and considered waiting to see if the official release notes might shed light on this behaviour,
but thought you might want to know about the rough edges sooner than that.

The Issue
---
After updating to `0.15.0-rc.1`, I get about a thousand of these _every second_:
```
ERROR bevy_picking::mesh_picking::ray_cast::intersections: Invalid intersection check: `TriangleList` is the only supported `PrimitiveTopology`
```

I am indeed using an alternate `PrimitiveTopology`, though I am not yet utilising any `picking` features within my code.

Each of the following is unclear to me:
1. Why it needs to be logging this for every function call (or even every frame), vs using `error_once!`
2. Why it is considered an error and not a warning (given that it is an optional feature)
3. Why it is attempting to compute so many intersections (given that I'm not requesting such information in my systems)
4. Why it even matters which `PrimitiveTopology` is used (the other variants work just fine elsewhere in Bevy)

Reproduction
---
You can reproduce this behaviour with the regular `3d/lines` example, with one tweak:
1. Change one or both usages of `RenderAssetUsages::RENDER_WORLD` to `RenderAssetUsages::default()`
2. Run the example and move your mouse near the lines

aevyrie:
# Objective

- Mesh picking is noisy when a non triangle list is used
- Mesh picking runs even when users don't need it
- Resolve #16065 

## Solution

- Don't add the mesh picking plugin by default
- Remove error spam

### DIFF
--- a/crates/bevy_picking/src/mesh_picking/ray_cast/intersections.rs
+++ b/crates/bevy_picking/src/mesh_picking/ray_cast/intersections.rs
@@ -37,9 +37,7 @@ pub(super) fn ray_intersection_over_mesh(
     backface_culling: Backfaces,
 ) -> Option<RayMeshHit> {
     if mesh.primitive_topology() != PrimitiveTopology::TriangleList {
-        error!(
-            "Invalid intersection check: `TriangleList` is the only supported `PrimitiveTopology`"
-        );
+        // ray_mesh_intersection assumes vertices are laid out in a triangle list
         return None;
     }
 
--- a/crates/bevy_picking/src/mesh_picking/ray_cast/intersections.rs
+++ b/crates/bevy_picking/src/mesh_picking/ray_cast/intersections.rs
@@ -16,7 +16,7 @@ pub struct RayMeshHit {
     /// The distance from the ray origin to the intersection point.
     pub distance: f32,
     /// The vertices of the triangle that was hit.
-    pub triangle: Option<[Vec3A; 3]>,
+    pub triangle: Option<[Vec3; 3]>,
     /// The index of the triangle that was hit.
     pub triangle_index: Option<usize>,
 }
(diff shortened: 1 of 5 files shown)


crates/bevy_picking/src/mesh_picking/ray_cast/intersections.rs
```diff
@@ -37,9 +37,7 @@ pub(super) fn ray_intersection_over_mesh(
     backface_culling: Backfaces,
 ) -> Option<RayMeshHit> {
     if mesh.primitive_topology() != PrimitiveTopology::TriangleList {
-        error!(
-            "Invalid intersection check: `TriangleList` is the only supported `PrimitiveTopology`"
-        );
+        // ray_mesh_intersection assumes vertices are laid out in a triangle list
         return None;
     }
```
rs017991, line 43: Comment on lines 47/53:

While I didn't encounter these, at least [one other person](https://github.com/bevyengine/bevy/issues/16065#issuecomment-2439420880) did.

* How would you feel about removing all of the error logging?
* For any that are deemed useful enough to remain, consider lowering to a lower log level (and at the very least switching to `*_once!`)

aevyrie: Thanks for pointing this out!

aevyrie:
While cleaning this up, I ran some benchmarks to make sure I wasn't regressing performance in this hot path. My initial changes caused a pretty serious regression, but after tweaking things for a bit, this PR is now as significant performance *improvement* compared with `main`.

Performance improvements range from -9% to -19% drops in run time.

[perf.txt](https://github.com/user-attachments/files/17532215/perf.txt)

BenjaminBrienen (approved):
Examples run great

alice-i-cecile (approved):
Great, I'm very happy with this resolution.

PR #16113, opened 2024-10-26
Improve formatting

BenjaminBrienen:
# Objective

Format the code optimally. This helps avoid conflicts and unnecessary changes.

For example, when trying to use Find & Replace to make code changes, formatting everything on 1 line makes it simpler. After the Find & Replace operations, formatting can be returned back to normal. This is what I did in #16112.

## Solution

1. In `rustfmt.toml` set `max_width = 9999`.
2. Run `cargo fmt --all`.
3. Remove `max_width`. (sets it back to `100`)
4. Run `cargo fmt --all`.

I consider this to be a workaround to a bug in rustfmt where the resulting formatted code depends on how it was formatted beforehand.

## Testing

CI


mockersf:
I would prefer this to be blocked until after the 0.15 is released, as it has high chances of conflicts during cherry picks.

IceSentry:
This turns lines that were already above 100 to be even longer. I don't see this as a win at all. Especially the one with long strings that put all the parameters on the same already way too long line.

I'm pretty sure a bunch of those are bugs in rustfmt where it just gives up when lines are super long because of a string.

crates/bevy_asset/src/server/mod.rs
```diff
@@ -913,29 +910,19 @@ impl AssetServer {
...
                 match load_folder(source.id(), path.path(), asset_reader, &server, &mut handles).await {
-                    Ok(_) => server.send_asset_event(InternalAssetEvent::Loaded {
-                        id,
-                        loaded_asset: LoadedAsset::new_with_dependencies(
-                            LoadedFolder { handles },
-                            None,
-                        )
-                        .into(),
-                    }),
+                    Ok(_) => server.send_asset_event(InternalAssetEvent::Loaded { id, loaded_asset: LoadedAsset::new_with_dependencies(LoadedFolder { handles }, None).into() }),
```
IceSentry, line 921: This is one of the changes I'm talking about that aren't an improvement at all and just seem like a rustfmt failure.

crates/bevy_core_pipeline/src/core_3d/main_opaque_pass_3d_node.rs
```diff
@@ -37,15 +37,7 @@ impl ViewNode for MainOpaquePass3dNode {
...
-        (
-            view,
-            camera,
-            target,
-            depth,
-            skybox_pipeline,
-            skybox_bind_group,
-            view_uniform_offset,
-        ): QueryItem<'w, Self::ViewQuery>,
+        (view, camera, target, depth, skybox_pipeline, skybox_bind_group, view_uniform_offset): QueryItem<'w, Self::ViewQuery>,
```
IceSentry, line 40: This isn't an improvement either.

crates/bevy_asset/src/server/mod.rs
BenjaminBrienen: I guess we'll have to manually format these lines. If there aren't too many, maybe we can just catch them in review.

IceSentry: My issue is that it's most of the changes that suffer from this, I only pointed out the first one that were too bad to ignore for me, but the more I scroll the more I see changes that I see as complete downsides.

PR #16114, opened 2024-10-26
Fix `dead_code` warnings in examples

BenjaminBrienen:
# Objective

Fix a few warnings that show up depending on `bevy_ci_testing`

## Solution

Expect dead code depending on the feature flag

## Testing

Tested locally with cargo check.

### DIFF
--- a/examples/testbed/3d.rs
+++ b/examples/testbed/3d.rs
@@ -138,6 +138,10 @@ mod light {
     }
 }
 
+#[cfg_attr(
+    all(feature = "bevy_ci_testing", target_os = "windows"),
+    expect(dead_code)
+)]
 mod bloom {
     use bevy::{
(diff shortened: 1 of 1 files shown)


mockersf:
I think this adds complexity unrelated to the example and without a lot of value

BenjaminBrienen:
Without this change there are constantly 4 warnings for no reason. The value is that there aren't 4 warnings. Maybe a better solution would be to make those scenes work on the CI windows runner?

mockersf:
Aren't those warning only when you build the examples with the feature `bevy_ci_testing` enabled?

BenjaminBrienen:
Yes, which happens in the common case of this config: `"rust-analyzer.check.features": "all",`.

alice-i-cecile:
Agreed, unfortunately I think the value isn't here :(

Issue #16120, opened 2024-10-27
Randomly sprite does not render and systems do not function


Frost2779:
## Bevy version

Reproducible on 0.14.2, 0.15.0-rc1, and main.

## \[Optional\] Relevant system information

Rust: 1.82.0
OS: Nobara Linux 40

```ignore
`AdapterInfo { name: "NVIDIA GeForce RTX 2070 SUPER", vendor: 4318, device: 7812, device_type: DiscreteGpu, driver: "NVIDIA", driver_info: "560.35.03", backend: Vulkan }`
```
I have tested on upstream `wgpu` and experiences no rendering issues.

## What you did

I created a simple project with only the most recent release of bevy as the dependency. I setup the project as needed for what I am trying to accomplish. That being getting a sprite to render to the screen, be able to move around the screen using the middle mouse button, and then use the scroll wheel to be able to zoom in/out. 

Upon running the project, I received the following output:

![Image](https://github.com/user-attachments/assets/34bb7d60-b0be-41a9-90cb-d463b8f47310)

## What went wrong

Sometimes, it seems like the sprite doesn't render, and other times, the sprite renders, but the systems seem not to function as expected, seeming to do nothing. Various println have confirmed that the systems are in fact running, and handle proper mouse input, but seem to have no effect. This usually required me to run the app a few times before both the sprite renders and the systems function. Furthermore, even if the mouse systems section is removed, the issue with the sprite randomly not rendering still persists. 

- what were you expecting?
The sprite would render and the systems would function consistently.

MRE of the project code:
```rust

use bevy::{input::{common_conditions::input_pressed, mouse::{MouseMotion, MouseWheel}}, prelude::*};

#[derive(Component)]
struct MainCamera;

fn main() {
    App::new()
        .add_plugins(
            DefaultPlugins
                .set(ImagePlugin::default_nearest())
        )
        .add_systems(Startup, setup)
        .add_systems(Update, (
            mouse_zoom, 
            mouse_pan.run_if(input_pressed(MouseButton::Middle))
        ))
        .run();
}

fn setup(mut commands: Commands, server: Res<AssetServer>) {
    commands.spawn((Camera2d::default(), MainCamera));
    commands.spawn(Sprite{
        image: server.load("placeholder.png"),
        ..default()
    });
}

fn mouse_zoom(
    mut scroll_events: EventReader<MouseWheel>,
    mut query_camera: Query<&mut OrthographicProjection>
) {
[...]

PR #12542, opened 2024-03-17
Fix crash on Linux Nvidia 550 driver

eero-lehtinen:
# Objective

Fix crashing on Linux with latest stable Nvidia 550 driver when resizing. The crash happens at startup with some setups.

Fixes #12199

I think this would be nice to get into 0.13.1

## Solution

Ignore `wgpu::SurfaceError::Outdated` always on this platform+driver.

It looks like Nvidia considered the previous behaviour of not returning this error a bug:
"Fixed a bug where vkAcquireNextImageKHR() was not returning VK_ERROR_OUT_OF_DATE_KHR when it should with WSI X11 swapchains" (https://www.nvidia.com/Download/driverResults.aspx/218826/en-us/)

What I gather from this is that the surface was outdated on previous drivers too, but they just didn't report it as an error. So behaviour shouldn't change.

In the issue conversation we experimented with calling `continue` when this error happens, but I found that it results in some small issues like bevy_egui scale not updating with the window sometimes. Just doing nothing seems to work better.

## Changelog

- Fixed crashing on Linux with Nvidia 550 driver when resizing the window

PR #12055, opened 2024-02-22
configure_surface needs to be on the main thread on iOS

mockersf:
# Objective

- Bevy fails to change screen orientation on iOS
```
Main Thread Checker: UI API called on a background thread: -[UIView layer]
PID: 37669, TID: 13872050, Thread name: Compute Task Pool (1), Queue name: com.apple.root.default-qos.overcommit, QoS: 0
Backtrace:
4   bevy_mobile_example                 0x0000000102cf92b8 _ZN60_$LT$$LP$$RP$$u20$as$u20$objc..message..MessageArguments$GT$6invoke17h8944e3d8ee34f15fE + 64
5   bevy_mobile_example                 0x0000000102c46358 _ZN4objc7message8platform15send_unverified17h667844cebe2d7931E + 132
6   bevy_mobile_example                 0x0000000102bcbd6c _ZN8wgpu_hal5metal7surface100_$LT$impl$u20$wgpu_hal..Surface$LT$wgpu_hal..metal..Api$GT$$u20$for$u20$wgpu_hal..metal..Surface$GT$9configure17h8a6af0f24cec1328E + 1548
7   bevy_mobile_example                 0x000000010279be50 _ZN9wgpu_core6device6global52_$LT$impl$u20$wgpu_core..global..Global$LT$G$GT$$GT$17surface_configure17h52709bbb3b3f0ff1E + 2792
8   bevy_mobile_example                 0x000000010287aacc _ZN84_$LT$wgpu..backend..wgpu_core..ContextWgpuCore$u20$as$u20$wgpu..context..Context$GT$17surface_configure17h54077b9f040286a4E + 508
9   bevy_mobile_example                 0x00000001028904b4 _ZN47_$LT$T$u20$as$u20$wgpu..context..DynContext$GT$17surface_configure17hfd6a0ac5a67a8f02E + 256
10  bevy_mobile_example                 0x00000001028a1870 _ZN4wgpu7Surface9configure17h97bf7dbd54220473E + 148
11  bevy_mobile_example                 0x0000000101fdc7cc _ZN11bevy_render8renderer13render_device12RenderDevice17configure_surface17h6853eab840b53e07E + 56
12  bevy_mobile_example                 0x000000010228eb64 _ZN11bevy_render4view6window15prepare_windows17hf6f8b3c93ba189b8E + 3248
13  bevy_mobile_example                 0x0000000102169eb8 _ZN4core3ops8function5FnMut8call_mut17h53ae762930afec98E + 192
[...]

PR #16123, opened 2024-10-28
Fix linux nvidia + xwayland freeze at startup

### LINKED ISSUE #16122: NVIDIA drivers on Linux can cause an intermittent failure to render under X11
> Maybe we could rethink the `Couldn't get swap chain texture. This often happens with the NVIDIA drivers on Linux. It can be safely ignored.` error to somehow force update the window on startup. It seems that it infact can't be safely ignored. 

 _Originally posted by @eero-lehtinen in [#16120](https://github.com/bevyengine/bevy/issues/16120#issuecomment-2440215274)_

eero-lehtinen:
# Objective

- Fixes #16122

When the wayland feature is not enabled, xwayland is used on wayland. Nvidia drivers are somewhat bugged on linux and return outdated surfaces on xwayland for seemingly no reason. Oftentimes at startup we get into an infine loop where the surface is permanently outdated and nothing (or sometimes only the first frame) is drawn on the screen.

## Solution

After experimenting I found that we can safely call configure again and the issue seems to resolve itsef. After this change I couldn't reproduce the original issue after many tries. More testing is probably needed though.

The main issue is that `get_current_texture` fails sometimes because the surface remains outdated even after configuring. It would be better to just properly handle and never panic when `get_current_texture` fails. This way we always call configure when outdated and bail when getting the swapchain fails instead of crashing. The number of special cases is also reduced.

## Testing

I tested the example "rotation" manually by trying to move around.

It works with X11 and Xwayland and the non panicing code paths didn't change so other platforms aren't affected.

### DIFF
--- a/crates/bevy_render/src/view/window/mod.rs
+++ b/crates/bevy_render/src/view/window/mod.rs
@@ -5,9 +5,11 @@ use crate::{
 };
 use bevy_app::{App, Plugin};
 use bevy_ecs::{entity::EntityHashMap, prelude::*};
-#[cfg(target_os = "linux")]
-use bevy_utils::warn_once;
-use bevy_utils::{default, tracing::debug, HashSet};
+use bevy_utils::{
+    default,
+    tracing::{debug, warn},
+    HashSet,
+};
 use bevy_window::{
     CompositeAlphaMode, PresentMode, PrimaryWindow, RawHandleWrapper, Window, WindowClosing,
 };
@@ -193,9 +195,6 @@ impl WindowSurfaces {
     }
 }
 
-#[cfg(target_os = "linux")]
-const NVIDIA_VENDOR_ID: u32 = 0x10DE;
-
 /// (re)configures window surfaces, and obtains a swapchain texture for rendering.
 ///
 /// NOTE: `get_current_texture` in `prepare_windows` can take a long time if the GPU workload is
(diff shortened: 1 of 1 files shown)


alice-i-cecile:
@Frost2779 can you test this for me? This should help with what you were seeing in #16120.

crates/bevy_render/src/view/window/mod.rs
```diff
@@ -280,6 +280,7 @@ pub fn prepare_windows(
                 }
                 #[cfg(target_os = "linux")]
                 Err(wgpu::SurfaceError::Outdated) if is_nvidia() => {
+                    render_device.configure_surface(surface, &surface_data.configuration);
```
alice-i-cecile, line 283: Comment please!

crates/bevy_render/src/view/window/mod.rs
```diff
@@ -280,6 +280,7 @@ pub fn prepare_windows(
...
                     warn_once!(
                         "Couldn't get swap chain texture. This often happens with \
```
alice-i-cecile, line 285: Can we remove this warning now?

Frost2779:
@alice-i-cecile I tested with the MRE from the original issue thread with 10 separate launches and did not encounter the unexpected behavior. Sprite rendered and systems seemed to function as expected.

richchurcher:
> @alice-i-cecile I tested with the MRE from the original issue thread with 10 separate launches and did not encounter the unexpected behavior. Sprite rendered and systems seemed to function as expected.

Just out of curiosity, did you test with `wayland` feature on or off?

eero-lehtinen:
I don't really get why we even have this if-else here when the two cases are identical except the else case handles extra stuff that already never happens in the first.

crates/bevy_render/src/view/window/mod.rs
```diff
@@ -431,5 +405,7 @@ pub fn create_surfaces(
             };
             render_device.configure_surface(&data.surface, &data.configuration);
         }
+
+        window_surfaces.configured_windows.insert(window.entity);
```
eero-lehtinen, line 409: I thought marking windows configured would fit after configuring them instead of in the next system where they are configured only when outdated.

BenjaminBrienen (approved):

crates/bevy_render/src/view/window/mod.rs
```diff
@@ -250,60 +249,35 @@ pub fn prepare_windows(
...
+                window.set_swapchain_texture(frame);
+            }
+            Err(wgpu::SurfaceError::Outdated) => {
+                render_device.configure_surface(surface, &surface_data.configuration);
+                let frame = match surface.get_current_texture() {
+                    Ok(frame) => frame,
+                    Err(err) => {
+                        // This is a common occurrence on X11 and Xwayland with NVIDIA drivers
+                        // when opening and resizing the window.
+                        warn!("Couldn't get swap chain texture after configuring. Cause: '{err}'");
```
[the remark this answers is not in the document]
eero-lehtinen: We don't have any platforms that instantly panic when spawning a window or resizing, so this doesn't functionally change anything. Also it is better to handle errors instead of just panicing in general, clearly just having an outdated surface isn't fatal. It's also completely possible that some other platform also returns an error here in the future.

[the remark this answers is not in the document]
tychedelia: Okay, that's good enough for me. Thanks!

eero-lehtinen:
> I don't really get why we even have this if-else here when the two cases are identical except the else case handles extra stuff that already never happens in the first.
> 
> #12542 here I also wanted to remove the if-else but didn't because people didn't want to review too much code, but I really do think it's better that way.

It looks like here https://github.com/bevyengine/bevy/pull/12055/files#diff-b0a34efd7fd2e03ca78f181d1ecce355b927f07669a75501af11498f8b9ec3adL343 we did something extra in the if statement, but now it's not useful anymore.

PR #16129, opened 2024-10-28
Picking example touchups

aevyrie:
# Objective

- Cleanup pass to make the examples a bit more succinct, focusing on the topic at hand.
- Added drag rotation to make the picking examples more interesting, and to demonstrate a simple use case.

### DIFF
--- a/examples/picking/mesh_picking.rs
+++ b/examples/picking/mesh_picking.rs
@@ -21,36 +21,15 @@
 
 use std::f32::consts::PI;
 
-use bevy::{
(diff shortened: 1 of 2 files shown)


aevyrie:
Some changes might be controversial, happy to revert anything.

BenjaminBrienen (approved):

alice-i-cecile (approved):
Eh, the performance note probably belongs in the mesh picking docs. This is a solid improvement.

mnmaita (approved):

PR #16133, opened 2024-10-28
Update notify-debouncer-full requirement from 0.3.1 to 0.4.0

mnmaita:
# Objective

- Supersedes #16126 

## Solution

- Updated code in `file_watcher.rs` to fix breaking changes introduced in the new version.
- Check changelog here: https://github.com/notify-rs/notify/blob/main/CHANGELOG.md#debouncer-full-040-2024-10-25.
- Relevant PR with the breaking change: https://github.com/notify-rs/notify/pull/557.

## Testing

- CI checks passing locally

### DIFF
--- a/crates/bevy_asset/Cargo.toml
+++ b/crates/bevy_asset/Cargo.toml
@@ -65,7 +65,7 @@ wasm-bindgen-futures = "0.4"
 js-sys = "0.3"
 
 [target.'cfg(not(target_arch = "wasm32"))'.dependencies]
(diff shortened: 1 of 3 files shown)


mnmaita:
I checked and `RecommendedCache` translates to `NoCache` on Linux (and Android), and to `FileIdMap` on other platforms. Unsure if we need to add a comment regarding this somewhere.

BenjaminBrienen (approved):

alice-i-cecile (approved):

PR #16135, opened 2024-10-28
Update taffy requirement from 0.5 to 0.6

mnmaita:
# Objective

- Supersedes #16127.

## Solution

- Added missing fields in `taffy::style::Style` struct.

## Testing

- TBD


mnmaita:
Wondering if I should just do `..default()` to fill in the new fields or if it's better to have all the fields explicitly in case there are future breaking changes that might affect the display of UI elements. Feedback will be much appreciated.

mnmaita:
Oh! Thanks for the heads up @rparrett, I'll close this and the dependabot PR and reference that one.

PR #16139, opened 2024-10-28
`ReflectBundle::remove` improvement

### LINKED ISSUE #15676: `ReflectBundle::remove` should return the removed `ReflectBundle`

In the course of my 🧪 over at [i-cant-believe-its-not-bsn](https://github.com/Leafwing-Studios/i-cant-believe-its-not-bsn), I've implemented a `WithChild` component, which takes the data from the component and inserts it as a child.

Doing this for multiple children is much harder though, and requires either an absolute abuse of generics or some form of type erasure.

[`ReflectBundle::remove`](https://dev-docs.bevyengine.org/bevy/ecs/reflect/struct.ReflectBundle.html#method.remove) looked promising, but it doesn't spit the data back out for me to reinsert.

Return the removed component data as a `ReflectBundle`

We could leave the existing behavior alone (or make it return `&mut EntityCommands` for chaining) and add `take`, to match the non-reflected API.

We could (and probably should) also add `EntityWorldMut::take_by_id`, which would allow a similar style of design. That would probably need to return a pile of bytes or something though: there's no guarantee that the components are reflectable.

BenjaminBrienen:
# Objective

Fixes #15676

## Solution

`remove` returns the removed item

Add `take`

## Testing

None yet

## Migration Guide

If you don't need the returned value from `remove`, discard it.


crates/bevy_ecs/src/reflect/bundle.rs
```diff
@@ -85,8 +86,19 @@ impl ReflectBundle {
...
+    pub fn remove(&self, entity: &mut EntityWorldMut) -> ReflectBundle {
+        (self.0.remove)(entity)
+    }
+
+
+    /// Removes all components in the [`Bundle`] from the entity and returns their previous values.
+    ///
+    /// **Note:** If the entity does not have every component in the bundle, this method will not
+    /// remove any of them.
+    // TODO: BundleRemover?
```
alice-i-cecile, line 98: This comment is very unclear to me :) I know it's draft but I want to make sure we don't forget.

alice-i-cecile:
Super easy! Much easier than I was expecting TBH. Thanks for tackling this.

crates/bevy_ecs/src/reflect/bundle.rs
BenjaminBrienen: It's copy pasted from ecs's take for now

crates/bevy_ecs/src/reflect/bundle.rs
```diff
@@ -35,7 +35,8 @@ pub struct ReflectBundleFns {
     /// Function pointer implementing [`ReflectBundle::apply_or_insert()`].
     pub apply_or_insert: fn(&mut EntityWorldMut, &dyn PartialReflect, &TypeRegistry),
     /// Function pointer implementing [`ReflectBundle::remove()`].
-    pub remove: fn(&mut EntityWorldMut),
+    pub remove: fn(&mut EntityWorldMut) -> ReflectBundle,
+    pub take: fn(&mut EntityWorldMut) -> Option<ReflectBundle>,
```
MrGVSV, line 39: I think one downside to this approach is that it couples these pointers to `ReflectBundle`, making them harder to manually construct outside of a `ReflectBundle` (which iirc is part of the reason they were separated into their own struct)

MrGVSV, line 39: Imo `take` should return the bundle as a `Box<dyn Reflect>`

crates/bevy_ecs/src/reflect/bundle.rs
```diff
@@ -85,8 +86,19 @@ impl ReflectBundle {
     }
 
     /// Removes this [`Bundle`] type from the entity. Does nothing if it doesn't exist.
-    pub fn remove(&self, entity: &mut EntityWorldMut) {
-        (self.0.remove)(entity);
+    pub fn remove(&self, entity: &mut EntityWorldMut) -> ReflectBundle {
+        (self.0.remove)(entity)
+    }
```
MrGVSV, line 91: As mentioned on Discord, I think we could move the chaining convenience stuff to these methods rather than on the pointers themselves

crates/bevy_ecs/src/reflect/bundle.rs
MrGVSV: I think this might also be a confusion of concepts (apologies if untrue)? A `ReflectBundle` by itself holds no runtime information. It's all just type information. So to insert the bundle you need to pair it with a matching `dyn Reflect`. Likewise, to take the bundle you need to return it as a `dyn Reflect`.

Hopefully that helps clear things up!

MrGVSV (approved):
Looks good to me! Thanks!

alice-i-cecile (approved):
I think this is an unambiguous improvement and we should merge it as is.

BenjaminBrienen:
> I think this is an unambiguous improvement and we should merge it as is.

Finally, one of my PRs isn't in the grey area! 😆

PR #16141, opened 2024-10-28
Code style consistency with named lifetimes

BenjaminBrienen:
# Objective

A certain level of consistency in how lifetimes are (or aren't) written.

## Solution

Fix `'a` -> `'w` where appropriate.
Fix "`'w` should always be written" where appropriate.
Fix "`'s` should always be written" where appropriate.
Fix "named lifetimes should always be written" where appropriate.

## Testing

Changes should be effectively "code style", not semantic.


alice-i-cecile (approved):

BenjaminBrienen:
Hold on, I made a mistake

BenjaminBrienen:
image
I just want you to know that this is all coming from clippy

BenjaminBrienen:
@alice-i-cecile No longer 2 lines :(

BenjaminBrienen:
Ok, clippy isn't requiring anything more.

crates/bevy_audio/src/audio_output.rs
```diff
@@ -58,7 +58,7 @@ pub struct PlaybackRemoveMarker;
 pub(crate) struct EarPositions<'w, 's> {
     pub(crate) query: Query<'w, 's, (Entity, &'static GlobalTransform, &'static SpatialListener)>,
 }
-impl<'w, 's> EarPositions<'w, 's> {
+impl EarPositions<'_, '_> {
```
alice-i-cecile, line 61: This is a regression.

crates/bevy_ecs/src/change_detection.rs
```diff
@@ -601,7 +601,7 @@ impl<'w, T: Resource> From<Res<'w, T>> for Ref<'w, T> {
     }
 }
 
-impl<'w, 'a, T: Resource> IntoIterator for &'a Res<'w, T>
+impl<'a, T: Resource> IntoIterator for &'a Res<'_, T>
```
alice-i-cecile, line 604: These are regressions.

Victoronz:
All the `'a` lifetimes that are hidden away are a nice cleanup, but this change also hides away a lot of named lifetimes, mostly `'w` and `'s`. 
There are cases when only one of the two is present, and now when hidden, is no longer discernible from just any elided `'a` lifetime. In other places now only one of these two is hidden, which is no longer be discernible from `<'w, 'a>`, which also exists.

crates/bevy_ecs/src/event/mutator.rs
```diff
@@ -48,7 +48,7 @@ pub struct EventMutator<'w, 's, E: Event> {
     events: ResMut<'w, Events<E>>,
 }
 
-impl<'w, 's, E: Event> EventMutator<'w, 's, E> {
+impl<E: Event> EventMutator<'_, '_, E> {
```
alice-i-cecile, line 51: This is a regression.

alice-i-cecile (changes requested):
Rescinding my approval: we shouldn't do this systematically. The 'a lifetimes are fine, but as @Victoronz points out, the loss of named lifetimes is a net downgrade for readability.

Victoronz:
> Yeah, I don't like these changes as a whole, and would prefer to allow this lint.

I agree. My opinion generally is:
Unnamed lifetime -> elide when possible
Named lifetime -> never elide

crates/bevy_core/src/serde.rs
```diff
@@ -24,7 +24,7 @@ impl<'de> Deserialize<'de> for Name {
 
 struct EntityVisitor;
 
-impl<'de> Visitor<'de> for EntityVisitor {
+impl Visitor<'_> for EntityVisitor {
```
alice-i-cecile, line 27: These ones should be reverted; the 'de convention is helpful.

crates/bevy_ecs/src/system/query.rs
```diff
@@ -792,7 +792,11 @@ impl<'w, 's, D: QueryData, F: QueryFilter> Query<'w, 's, D, F> {
     /// [`par_iter_mut`]: Self::par_iter_mut
     /// [`World`]: crate::world::World
     #[inline]
-    pub fn par_iter(&self) -> QueryParIter<'_, '_, D::ReadOnly, F> {
+    pub fn par_iter<'world, 'state>(&self) -> QueryParIter<'world, 'state, D::ReadOnly, F>
```
alice-i-cecile, line 795: Rest of this file uses 'w and 's.

crates/bevy_render/src/texture/fallback_image.rs
```diff
@@ -240,7 +240,7 @@ pub struct FallbackImageMsaa<'w> {
     default_sampler: Res<'w, DefaultImageSampler>,
 }
 
-impl<'w> FallbackImageMsaa<'w> {
+impl FallbackImageMsaa<'_> {
```
alice-i-cecile, line 243: This one should be reverted.

alice-i-cecile (changes requested):
Only a couple more fixes left! I'm in favor of these changes now, and I'm glad to see the lint allowed.

Victoronz:
I would be careful about keeping the scope of this PR down. I see it now also specifies some named lifetimes, which is fine for impls, but there are also some function (and trait function?) definitions in there now. 

Caution should be taken with other locations, because specifying lifetimes in function definitions (i.e. adding generics) can be a breaking change. It also means no longer systematically following/allowing a lint, so that should be a separate decision/PR imo.

BenjaminBrienen:
I totally understand if we aren't interested in tackling "lifetime style consistency" right now. The problems I noticed are:

- Many `'a` should be `'w`, so it's not obvious which changes should be reverted.
- Existing inconsistencies in the way lifetimes are written that go against feedback I've received in the PR, which makes me want to fix those, too.
- Other inconsistencies that are easy fixes, like `Formatter<'_>`.

BenjaminBrienen:
I also prefer to just follow the lints, for similar reasons. I'm just doing my best given the feedback since I think pretty much every member has been against that strategy. I'm trying to find a balance between "We don't want to have to write all the lifetimes" in #15916 (for example, writing `Mut<'w, T>` everywhere and "We don't want to ellide `'w`.

chescock:
> I'm just doing my best given the feedback since I think pretty much every member has been against that strategy.

Just in case it wasn't clear: I think you're doing a great job with this PR!  My comment was for the folks who were against that strategy.  There are definitely tradeoffs to be made here, but I wanted to highlight some of the benefits of your original proposal in case they hadn't been considered.

crates/bevy_asset/src/reflect.rs
```diff
@@ -21,12 +21,13 @@ pub struct ReflectAsset {
...
     // - may only be used to access **at most one** access at once
-    get_unchecked_mut: unsafe fn(UnsafeWorldCell<'_>, UntypedHandle) -> Option<&mut dyn Reflect>,
-    add: fn(&mut World, &dyn PartialReflect) -> UntypedHandle,
-    insert: fn(&mut World, UntypedHandle, &dyn PartialReflect),
-    len: fn(&World) -> usize,
+    get_unchecked_mut:
+        for<'w> unsafe fn(UnsafeWorldCell<'w>, UntypedHandle) -> Option<&'w mut dyn Reflect>,
+    add: for<'w, 'a> fn(&'w mut World, &'a dyn PartialReflect) -> UntypedHandle,
+    insert: for<'w, 'a> fn(&'w mut World, UntypedHandle, &'a dyn PartialReflect),
+    len: for<'w> fn(&World) -> usize,
```
bushrat011899, line 28: I think either the `for<...>` needs to be removed here, or the lifetime for `World` should be `'w`.

crates/bevy_ecs/src/query/fetch.rs
```diff
@@ -442,7 +442,7 @@ unsafe impl ReadOnlyQueryData for EntityLocation {}
 /// `fetch` accesses all components in a readonly way.
 /// This is sound because `update_component_access` and `update_archetype_component_access` set read access for all components and panic when appropriate.
 /// Filters are unchanged.
-unsafe impl<'a> WorldQuery for EntityRef<'a> {
+unsafe impl<'world> WorldQuery for EntityRef<'world> {
```
bushrat011899, line 445: I've seen some of these changes use `'w` and others use `'world`. Is there a particular reason for one over the other? I think it's fine if it's just an inconsistency as both are clear IMO

crates/bevy_ecs/src/query/fetch.rs
```diff
@@ -872,7 +872,7 @@ where
 }
 
 /// SAFETY: `Self` is the same as `Self::ReadOnly`.
-unsafe impl<'a, B> QueryData for EntityRefExcept<'a, B>
+unsafe impl<B> QueryData for EntityRefExcept<'_, B>
```
bushrat011899, line 875: Shouldn't this be `'w`?

bushrat011899 (commented):
In general I like the changes here. My only broad concern is there are a few instances of switching to anonymous `'_` lifetimes where I think `'w` is more appropriate.

crates/bevy_ecs/src/query/fetch.rs
BenjaminBrienen: Sometimes there is a conflicting, existing lifetime declared in an overlapping scope

crates/bevy_asset/src/reflect.rs
BenjaminBrienen: The fact that this isn't caught by a lint (among other things) makes me wish that we just went by the normal clippy lint and elided all lifetimes when possible.

crates/bevy_ecs/src/query/fetch.rs
BenjaminBrienen: Yeah, I'll fix all of those now

bushrat011899 (approved):

chescock:
> The fact that this isn't caught by a lint (among other things) makes me wish that we just went by the normal clippy lint and elided all lifetimes when possible.

I also wish that!  I don't entirely understand the objection to following the lint, and had been hoping [my comment above](https://github.com/bevyengine/bevy/pull/16141#issuecomment-2452043891) would spur a little more debate.  

> Sometimes there is a conflicting, existing lifetime declared in an overlapping scope

I think this is guaranteed to happen on every `SystemParam` and `WorldQuery` impl that uses those lifetimes.  The lifetime parameters on `Self` are never used in the impls.  They only exist to support the magic where ordinary functions can be converted to systems.  Everything gets the actual lifetimes re-applied in `type Item<'w>` and `type Item<'w, 's>`.  

Those traits are some of the cases where I think using `'_` to make it clear that the lifetime is never used is really valuable!

Victoronz:
Revisiting this, I think this is definitely a more intricate issue than it at first seems. This is my current understanding of the lifetimes here:

The `'w` lifetime doesn't semantically mean what it sounds like:
As far as I understand, it is supposed to represent a "borrow from world", which conceptually means that there is a `&'w World` that `'w` refers to. However, that isn't really the case. Semantically, a `'w` lifetime is simply a named lifetime like any other, i.e. `'a`. Its scope must match any other mention of `'w` in the item it is defined in, and satisfy any specified bounds on top of that. Because there is not just one "world borrow lifetime", we are lead to confusing situations where we have both `'world` and `'w`, which are both supposed to be the same, yet are semantically not.

Furthermore, `&'w mut Query<'w, ...>` doesn't mean we have a borrow of `Query` that lives as long as world, but that `'w` is now constrained to match the scope of that query borrow.
If we obtain an `EntityMut<'_>` from that query, what is originally written as `'w` in the struct definition instead becomes scoped on our query borrow (we could call this `'q`). If we now fetch an item from `EntityMut`, we rely on a borrow of that struct, not directly of `Query`, or `World`.
In these constrained cases, switching `'_` to `'w` is actually unsound! See: https://github.com/bevyengine/bevy/pull/17035
(Similar to this are `transmute`s, where the `'s` in `Query` now borrows from a `QueryLens`).

I do like `'_` when it can remove noise, and say "it is inferred, don't worry about this", and I like `'w` when it can clarify noise, and say "it is a world borrow, don't worry about this". 

When either applies is heavily case dependent, and I don't think I can broadly say.

BenjaminBrienen:
I wouldn't be surprised. Manually annotating the lifetimes correctly is nontrivial in sneaky ways. I think what this shows is that it is easier and more maintainable to elide lifetimes when possible. Currently, we do it one way half the time and another way half the time. Choosing to be consistent this way seems really hard.

PR #16142, opened 2024-10-28
Make `BinnedRenderPhase` fields for accessing batchable and unbatchable entities public

### LINKED ISSUE #16080: Make BinnedRenderPhase fields for accessing batchable and unbatchable entities public
BinnedRenderPhase already has two public fields (`batchable_mesh_keys` and `unbatchable_mesh_keys`) which contain the keys to the hashmaps `batchable_mesh_values` and `unbatchable_mesh_values`, so I think it would make sense if the hashamps were public as well.

It would be useful to be able to read the values of `batchable_mesh_values` and `unbatchable_mesh_values` in the render world, in order to be able to see
1. which entities are being batched together and
2. which entities have which `instance_index`.

### This could be used for
- gpu picking of instances (determining which entity has which instance index and sending the instance index from the render world to the app world and comparing it with the instance index from the shader)
- debugging (checking the expected entities are being batched together)

Making `batchable_mesh_values`, `unbatchable_mesh_values`, `UnbatchableBinnedEntities` and `UnbatchableBinnedEntities.entities` public instead of pub(crate).

- The fields could have public getters instead of being public.
- A resource or components could be created that contain a batch group number and the instance_index for each entity. This resource or components could be updated in the `batch_and_prepare_binned_render_phase` systems in `bevy/crates/bevy_render/src/batching/gpu_preprocessing.rs `and `bevy/crates/bevy_render/src/batching/no_gpu_preprocessing.rs`.

## Additional context
I don't know if there is already a way of getting the batch and `instance_index` for each entity that I am missing. By reading the values from `batchable_mesh_values` and `unbatchable_mesh_values`, I think it is possible to determine which entites have which `instance_index` by iterating over the vector of entities for each key in `batchable_mesh_values` and `unbatchable_mesh_values` in the same way as is done in the `batch_and_prepare_binned_render_phase` systems.

BenjaminBrienen:
# Objective

Fixes #16080

## Solution

Make the fields and struct pub as per the suggested solution.

## Testing

None

### DIFF
--- a/crates/bevy_render/src/render_phase/mod.rs
+++ b/crates/bevy_render/src/render_phase/mod.rs
@@ -101,7 +101,7 @@ where
     ///
     /// Each bin corresponds to a single batch set. For unbatchable entities,
     /// prefer `unbatchable_values` instead.
-    pub(crate) batchable_mesh_values: HashMap<BPI::BinKey, Vec<(Entity, MainEntity)>>,
+    pub batchable_mesh_values: HashMap<BPI::BinKey, Vec<(Entity, MainEntity)>>,
 
     /// A list of `BinKey`s for unbatchable items.
     ///
@@ -112,7 +112,7 @@ where
     /// The unbatchable bins.
     ///
     /// Each entity here is rendered in a separate drawcall.
-    pub(crate) unbatchable_mesh_values: HashMap<BPI::BinKey, UnbatchableBinnedEntities>,
+    pub unbatchable_mesh_values: HashMap<BPI::BinKey, UnbatchableBinnedEntities>,
 
     /// Items in the bin that aren't meshes at all.
     ///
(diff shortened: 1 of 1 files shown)


BenjaminBrienen:
@askogseide Let me know if you have any feedback

crates/bevy_render/src/render_phase/mod.rs
```diff
@@ -112,7 +112,7 @@ where
     /// The unbatchable bins.
```
BenjaminBrienen, line 112: Is "bins" correct here?

alice-i-cecile:
@askogseide could you please review this PR to verify that it meets your needs?

askogseide:
This looks good to me! Sorry for not responding earlier and thank you for working on this!

askogseide (approved):

Issue #12207, opened 2024-02-29
bevy_gizmo doc tests hang compute when run locally


alice-i-cecile:
## Bevy version

https://github.com/bevyengine/bevy/commit/b24ab2e9fba5b81aba15da97b6551a4e99aad40c

## \[Optional\] Relevant system information

Windows

## What you did

Run `cargo run -p ci --doc`.

## What went wrong

Doc tests for initial crates pass, and then a dozen or so doc tests within `bevy_gizmos` get stuck, producing a warning that they've been running for more than 60 seconds.

My computer then runs out of RAM (over 12 GB!), freezes, and I have to hard reboot.

## Additional information

Running with fewer threads via e.g. `-j 4` improves responsiveness and slows down time-to-hang, but does not resolve the problem.

This works fine in CI, so it may be OS specific.

solis-lumine-vorago:
This issue is related to 'Microsoft® Incremental Linker' -  the default linker used by the rust compiler on some (most?) windows machines. 
The linkers in the following image use 18.8 GB of RAM!
![gizmos-doc-test-task-manager](https://github.com/bevyengine/bevy/assets/62256001/b527fc3e-ca62-486e-a87b-6e05c7c677f4)

The one thing that sets `bevy_gizmos` apart from other crates like `bevy_ecs` is the increased "usage" of 'Microsoft® Incremental Linker'. Most other doc-tests also use 'Microsoft® Incremental Linker' but it does not use alot of memory there. Instead 'link.exe' uses more memory but remains within acceptable bounds (<1GB).

No single doc-test in particular seems to be responsible for the increased memory usage, as removing any given set of doc-tests results in a decrease in memory usage more or less proportional to the amount of tests removed. Instead, running doc-tests for (very) large projects on windows machines seems to be the issue.

I was unable to find an "in code" fix for this, but there is a [discussion](https://users.rust-lang.org/t/tips-for-compiling-large-rust-projects-with-limited-resources-on-windows/102056) on the rust forums that appears to suggest building for a different but compatible target or using `rust-lld` as the linker. 

However, this problem does not seem to be limited to Windows only, as #11034 details a very similar (the same?) issue on a Linux machine.

alice-i-cecile:
Further discussion on [Discord](https://discord.com/channels/691052431525675048/692572690833473578/1244956150894690386).

[...]

PR #16145, opened 2024-10-28
Add argument to ci for test threads and build jobs

### LINKED ISSUE #16051: Allow passing `--jobs` to `ci` to prevent OOM on home machines

Home machines does not have the same amount as the machines that run CI on GitHub, or maybe it is because the machines that run CI on GitHub have a small amount of CPUs, so running `cargo run -p ci -- doc` may cause OOM.

Allow passing `--jobs` into `ci` (`cargo run -p ci -- doc --jobs n`) to decrease parallelism and prevent OOM.

BenjaminBrienen:
# Objective

Fixes #16051

## Solution

Rework flags into arguments and add an arg for the number of test threads and an arg for the number of build jobs.

`cargo run -p ci -- --test-threads 1 --build-jobs 1`

## Testing

Tested CI locally. Build pipeline should also be a good way to catch mistakes in the command line usage.


tools/ci/src/args.rs
```diff
@@ -0,0 +1,5 @@
+#[derive(Copy, Clone, PartialEq, Eq)]
+pub struct Args {
+    pub keep_going: bool,
+    pub test_threads: Option<u8>,
+}
```
BD103, line 5: This is missing the `build_jobs` flag.

tools/ci/src/ci.rs
```diff
@@ -22,9 +40,14 @@ impl CI {
...
     pub fn run(self) {
         let sh = xshell::Shell::new().unwrap();
+        let mut prepared_commands = self.prepare(&sh);
 
-        let prepared_commands = self.prepare(&sh);
-
+        if let Some(build_jobs) = self.build_jobs {
+            prepared_commands = prepared_commands
+                .into_iter()
+                .map(|pc| pc.with_env_var("RUST_BUILD_JOBS", build_jobs.to_string().leak()))
```
BD103, line 48: This line will allocate and leak the same string multiple times, once for each command prepared. To avoid this, this should be changed to:

1. Only leak the string once.
2. Remove the `'static` requirement and use the `'a` lifetime that `PreparedCommand` already has.

I'd prefer 2, but it does require a bit more Rust magic.

BD103: Additionally, this should probably be converted into a normal `for` loop:

```rust
for pc in prepared_commands.iter_mut() {
    pc.env_vars.push(("RUST_BUILD_JOBS", the_string_variant_of_build_jobs));
}
```

tools/ci/src/ci.rs
```diff
@@ -59,28 +82,23 @@ impl CI {
     }
 
     fn prepare<'a>(&self, sh: &'a xshell::Shell) -> Vec<PreparedCommand<'a>> {
-        let mut flags = Flag::empty();
-
-        if self.keep_going {
-            flags |= Flag::KEEP_GOING;
-        }
-
+        let args = self.into();
```
BD103, line 85: (Nit): A type annotation here would be very useful when reading this code, especially since `into()` is quite general.

```rust
let args: Args = self.into();
```

While not necessary to compile the code (or if you have Rust Analyzer), it's nice when reviewing the code in Github.

tools/ci/src/args.rs
```diff
@@ -0,0 +1,5 @@
+#[derive(Copy, Clone, PartialEq, Eq)]
+pub struct Args {
```
BD103, line 2: (Nit) Could you add some docs to this struct, since the rest of the programs deals with it a lot? A few things I would cover:

1. What is its purpose? (To act as a `Prepare`-friendly version of `CI`)
2. Highlight how this needs to be kept up-to-date with `CI`.

BD103 (changes requested):

tools/ci/src/args.rs
BenjaminBrienen: It wouldn't be used if it were added here at the moment, but I should propagate that downward, too.

tools/ci/src/ci.rs
hukasu: setting the `RUST_BUILD_JOBS` was a workaround, the best would be passing `--jobs` to cargo

hukasu:
Also should fix #12207

tools/ci/src/ci.rs
hukasu: `cargo test --jobs 4 -- --test-threads 4` or something like this

BenjaminBrienen:
> Also should fix #12207 

More of a workaround than a fix, I think.

BD103 (approved):
Looks good!

hukasu:
I'm not understanding how to use the `--jobs` arg.

`--help` gives this `Usage: ci [--keep-going] [--test-threads <test-threads>] [--jobs <jobs>] [<command>] [<args>]`, but using `cargo run -p ci -- --jobs 4 doc` gives `error: Unrecognized option: 'jobs4'`, and using `cargo run -p ci -- --jobs=4 doc` gives `Unrecognized argument: --jobs=4`.

hukasu:
`--test-threads` worked great, no OOM during `bevy_gizmos` tests

bushrat011899 (approved):

hukasu:
@BenjaminBrienen does `--jobs` work for you?

BenjaminBrienen:
I'll test it when I have time later

hukasu:
pulled d7d808dd7418e9680a431339a1784bd5b4361acf and now i can pass `--jobs` but the build is still using more parallelism

`cargo run -p ci -- --jobs 4` fails on `cargo fmt` because it does not take `--jobs`

`cargo run -p ci -- --jobs 4 doc` fails with
```
error: Unrecognized option: 'jobs'
error: doctest failed, to rerun pass `-p benches --doc`
thread 'main' panicked at tools/ci/src/ci.rs:72:13:
One or more CI commands failed:
doc-test: Please fix failing doc tests in output above.
```
and i saw on the console that it is passing as a string(?)
```
cargo test --workspace --doc " --jobs 4" -- 
```

hukasu:
`cargo run -p ci -- --jobs 4 bench-check` also fails with `--jobs` being included between quotes
```
cargo check --benches --target-dir ../target --manifest-path "./benches/Cargo.toml --jobs 4"
```

BenjaminBrienen:
I'll look at this more later thanks

hukasu:
`cargo run -p ci -- --jobs 4 clippy` is quoting
```
cargo clippy --workspace --all-targets "--all-features --jobs 4" -- -Dwarnings
```

BenjaminBrienen:
Reading the documentation for `cmd!`, looks like I'll have to approach this differently.

hukasu:
can you set `env` with that crate? 'cuz you could have `ci` take in the arguments, and pass to the subprocesses through `env`

hukasu:
Adopted

BD103:
Closing in favor of #19359 :)

PR #15428, opened 2024-09-25
improve doc check

BenjaminBrienen:
# Objective

Fixes #15427

## Solution

Add the environment variables to the doc check

## Showcase

**After making a change that only affects nightly:**
image

**After undoing the offensive change:**
image

## Testing

See the showcase and commit failure/success status

PR #16147, opened 2024-10-29
Make doc CI use nightly

### LINKED ISSUE #15427: `cargo run -p ci` doesn't handle doc tests properly
> Maybe the command that you used didn't correctly export the env vars?

Ah, this command did the trick:
`$env:RUSTFLAGS='--cfg docsrs_dep'; $env:RUSTDOCFLAGS='--cfg=docsrs'; cargo +nightly doc`

Fixing the errors now!

_Originally posted by @BenjaminBrienen in https://github.com/bevyengine/bevy/issues/15387#issuecomment-2374009554_

BenjaminBrienen:
# Objective

Fixes #15427
Follow-up to #15428

## Solution

Use nightly and add the environment variables as suggested here: https://github.com/bevyengine/bevy/pull/15428#pullrequestreview-2331294421

## Testing

CI should run

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -316,7 +316,9 @@ jobs:
             ~/.cargo/git/db/
             target/
           key: ${{ runner.os }}-check-doc-${{ hashFiles('**/Cargo.toml') }}
-      - uses: dtolnay/rust-toolchain@stable
+      - uses: dtolnay/rust-toolchain@master
+        with:
+          toolchain: ${{ env.NIGHTLY_TOOLCHAIN }}
       - name: Install Linux dependencies
         uses: ./.github/actions/install-linux-deps
         with:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -127,7 +127,7 @@ jobs:
       - name: Check Compile
         # See tools/ci/src/main.rs for the commands this runs
         run: cargo run -p ci -- compile
-  
+
   check-compiles-no-std:
     runs-on: ubuntu-latest
     timeout-minutes: 30
@@ -301,7 +301,7 @@ jobs:
         with:
           name: example-run-macos
           path: example-run/
-        
+
   check-doc:
     runs-on: ubuntu-latest
     timeout-minutes: 30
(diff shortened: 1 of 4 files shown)


BenjaminBrienen:
@alice-i-cecile I think fixing that error would require making the whole CI use nightly. Do you have a different idea?

alice-i-cecile (approved):

BD103:
Because docs are already built with nightly, I'm fine with enabling it here:

https://github.com/BenjaminBrienen/bevy/blob/2a9c0fad9a90c74a19e3820581b6e46de1fe8769/.github/workflows/ci.yml#L319

```diff
- - uses: dtolnay/rust-toolchain@stable
+ - uses: dtolnay/rust-toolchain@master
+   with:
+     toolchain: ${{ env.NIGHTLY_TOOLCHAIN }}
```

mockersf:
the tool shouldn't force CI but use the version decided by the user

BD103:
> The tool shouldn't force nightly but use the version decided by the user
> 
> They can print a warning that nightly is recommended

I agree. I think the path forward here is:

1. Make the CI checks for docs specifically use nightly.
2. Detect the version of `rustc` (run `rustc --version`) and enable nightly features based on that.

.github/workflows/ci.yml
```diff
@@ -127,7 +127,7 @@ jobs:
       - name: Check Compile
         # See tools/ci/src/main.rs for the commands this runs
         run: cargo run -p ci -- compile
-
```
BenjaminBrienen, line 130: We should look into CI enforced yml formatting

BD103 (approved):
The changes look good, but is it possible to test them before this is merged? (Perhaps test it on your fork?)

BenjaminBrienen:
> The changes look good, but is it possible to test them before this is merged? (Perhaps test it on your fork?)

Can you explain what you mean by that? I'm not sure what I could do on my fork that would be different from the CI here.

BD103:
> Can you explain what you mean by that? I'm not sure what I could do on my fork that would be different from the CI here.

Ah my bad, I thought this PR changed [`docs.yml`](https://github.com/bevyengine/bevy/blob/main/.github/workflows/docs.yml), which wouldn't be caught by CI. Sorry about that, @alice-i-cecile you're free to merge!

Issue #15979, opened 2024-10-18
Specify policy on panics in the Bevy repo.


tbillington:
## What problem does this solve or what need does it fill?

Bevy currently does not have a policy or documentation on the usage of panics in Rust code.

Because unhandled errors (panics) in Bevy result in the termination of the application, they are very harmful to the user experience. A panic in Rust signals the application has entered into an unrecoverable error state, and application termination is the only viable course of action. This is reinforced in the [Rust book](https://doc.rust-lang.org/book/ch09-03-to-panic-or-not-to-panic.html#summary).

> The `panic!` macro signals that your program is in a state it can’t handle and lets you tell the process to stop instead of trying to proceed with invalid or incorrect values.

Panicing is especially sensitive in a library like Bevy since it takes away choice from the user about how to handle a particular situation. To Bevys credit, it often does give the user a choice, such as `Query::get_single` which returns a `Result`, as opposed to `Query::single`, which is good to see.

### Unhandled error fallout

Since Bevy is based upon writing game logic in Rust at the same "level" as core engine code, and there is no separate runtime like engines such as Unity or Godot, unhandled errors due to either user error or engine library error are much more disruptive. With runtime based engine, an unhandled error in user code or much of the engine code will result in an exception being raised, and a error being logged, but otherwise execution of the game will continue relatively uninterrupted.

That resilience and gracefully handling of errors is incredibly useful to game developers. Not only are games often produced under time, budget, and team constraints, but they may also be worked on by large teams with people of varied skill sets that may not know all of the invariants that are supposed to be kept in game logic. Rapid prototyping, iteration, and changes late in development also prevent developers from producing perfect code that handles all edge cases. At their core, games are about delivering an experience for players playing on their own devices. As such, if not handled well, errors can be especially disruptive to the player experience, and ultimately to the games reception and reputation.

A crash in a Bevy game isn't just a stacktrace on the developers machine. It results in a negative experience for players, and possibly worse reviews and refunds, materially impacting the developer.

This being the case, the use of panics in Bevy should be avoided wherever possible, heavily scrutinised, and only used as a last result.

## What solution would you like?

Codify a policy on the use of panics with the Bevy engine, preferable upheld by CI lints/rules.

These rules could be tighter/looser per crate, but that's just minor details at this point.

## Additional context

<details><summary>Motivating experience</summary>

[...]

Issue #15891, opened 2024-10-13
`specialized_mesh_pipeline` example panics when hovering over mesh


rparrett:
## Bevy version

main
bisected to #15800.

## Relevant system information

SystemInfo { os: "MacOS 14.5 ", kernel: "23.5.0", cpu: "Apple M1 Max", core_count: "10", memory: "64.0 GiB" }
AdapterInfo { name: "Apple M1 Max", vendor: 0, device: 0, device_type: IntegratedGpu, driver: "", driver_info: "", backend: Metal }

## What you did

- `cargo run --example specialized_mesh_pipeline`
- hover over over the triangles

## What went wrong

```
thread 'Compute Task Pool (3)' panicked at crates/bevy_picking/src/mesh_picking/ray_cast/intersections.rs:145:29:
index out of bounds: the len is 3 but the index is 3
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
Encountered a panic in system `bevy_picking::mesh_picking::update_hits`!
Encountered a panic in system `bevy_app::main_schedule::Main::run_main`!
```

alice-i-cecile:
I'm pretty sure something is malformed in the mesh. This looks like the exact issue spotted by @tbillington and called out as not a problem in https://github.com/bevyengine/bevy/pull/15800#discussion_r1797081408 by @Jondolf.

rparrett:
It looks like it is because the mesh doesn't have normals.

Jondolf:
The [line](https://github.com/bevyengine/bevy/blob/82d6f56cfc3432f584632f761c44ec8efd2e7c67/crates/bevy_picking/src/mesh_picking/ray_cast/intersections.rs#L145) the panic is referring to is indexing the third vertex position for a triangle, not indexing a normal. The code does check if `Mesh::ATTRIBUTE_NORMAL` exists at all, and should work if it doesn't. I believe normals should only be an issue if there are less normals than there are vertices, which is considered malformed by Bevy anyway.

The panic happens because the example defines the mesh's indices as `[0, 1, 2, 0, 2, 3]`. Index `3` points to nothing since the mesh only has three vertices. Not sure why it's set up like that or how it's even meant to work.

rparrett:
Cool, you're right, thanks. Opened a tiny PR to fix.

PR #16148, opened 2024-10-29
Add tests and remove sources of panics in mesh picking

tbillington:
# Objective

Adds tests to cover various mesh picking cases and removes sources of panics.

It should prevent users being able to trigger panics in `bevy_picking` code via bad mesh data such as https://github.com/bevyengine/bevy/issues/15891, and is a follow up to my comments in https://github.com/bevyengine/bevy/pull/15800#pullrequestreview-2361694213.

This is me "putting my money where my mouth is" re https://github.com/bevyengine/bevy/issues/15979 :)

## Testing

Adds 8 new tests to cover `ray_mesh_intersection` code.

With new tests before `ray_mesh_intersection` changes in this PR there were 3 failures.

```
test mesh_picking::ray_cast::intersections::tests::ray_mesh_intersection_indices ... ok
test mesh_picking::ray_cast::intersections::tests::ray_mesh_intersection_indices_vertex_normals ... ok
test mesh_picking::ray_cast::intersections::tests::ray_mesh_intersection_not_enough_indices ... ok
test mesh_picking::ray_cast::intersections::tests::ray_mesh_intersection_simple ... ok
test mesh_picking::ray_cast::intersections::tests::ray_mesh_intersection_vertex_normals ... ok
test mesh_picking::ray_cast::intersections::tests::ray_mesh_intersection_missing_vertex_normals ... FAILED
test mesh_picking::ray_cast::intersections::tests::ray_mesh_intersection_bad_indices ... FAILED
test mesh_picking::ray_cast::intersections::tests::ray_mesh_intersection_incides_missing_vertex_normals ... FAILED
```

---

## Benchmarks

Small improvement of 0 to 10% in micro-benchmarks which was not what I expected. This was on m1 mac pro. Would be interested in benches on other platforms as benching on a laptop is not ideal.

```
git stash pop
cargo bench ray_mesh_intersection

     Running benches/bevy_picking/ray_mesh_intersection.rs (target/release/deps/ray_mesh_intersection-c384089e071d683e)

ray_mesh_intersection/100_vertices
                        time:   [709.55 ns 711.99 ns 714.46 ns]
                        change: [-3.2272% -2.8377% -2.4864%] (p = 0.00 < 0.05)
                        Performance has improved.
ray_mesh_intersection/10000_vertices
                        time:   [64.404 µs 64.513 µs 64.630 µs]
                        change: [-2.6677% -2.4757% -2.2773%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 5 outliers among 100 measurements (5.00%)
  5 (5.00%) high mild
ray_mesh_intersection/1000000_vertices
                        time:   [6.6422 ms 6.6597 ms 6.6781 ms]
                        change: [-3.2386% -2.8388% -2.4585%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 3 outliers among 100 measurements (3.00%)
  3 (3.00%) high mild

ray_mesh_intersection_no_cull/100_vertices
                        time:   [728.73 ns 733.48 ns 738.97 ns]
                        change: [+0.0399% +0.7068% +1.4202%] (p = 0.05 > 0.05)
                        No change in performance detected.
ray_mesh_intersection_no_cull/10000_vertices
                        time:   [65.945 µs 66.080 µs 66.228 µs]
                        change: [-2.9063% -2.6568% -2.4047%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 3 outliers among 100 measurements (3.00%)
  3 (3.00%) high mild
ray_mesh_intersection_no_cull/1000000_vertices
                        time:   [6.7569 ms 6.7706 ms 6.7856 ms]
                        change: [-3.9866% -3.6929% -3.3857%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 10 outliers among 100 measurements (10.00%)
  8 (8.00%) high mild
  2 (2.00%) high severe

ray_mesh_intersection_no_intersection/100_vertices
                        time:   [391.14 ns 395.55 ns 400.08 ns]
                        change: [-3.4350% -2.6984% -1.8984%] (p = 0.00 < 0.05)
                        Performance has improved.
ray_mesh_intersection_no_intersection/10000_vertices
                        time:   [43.852 µs 43.922 µs 43.993 µs]
                        change: [-6.7318% -6.5286% -6.3175%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 2 outliers among 100 measurements (2.00%)
  1 (1.00%) low severe
  1 (1.00%) low mild
ray_mesh_intersection_no_intersection/1000000_vertices
                        time:   [4.5213 ms 4.5358 ms 4.5510 ms]
                        change: [-10.603% -10.263% -9.9009%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 1 outliers among 100 measurements (1.00%)
  1 (1.00%) high severe
```

### DIFF
--- a/crates/bevy_picking/src/mesh_picking/ray_cast/intersections.rs
+++ b/crates/bevy_picking/src/mesh_picking/ray_cast/intersections.rs
@@ -94,17 +94,20 @@ pub fn ray_mesh_intersection<I: TryInto<usize> + Clone + Copy>(
 
             let triangle_index = Some(a);
             let tri_vertex_positions = &[
-                Vec3::from(positions[a]),
-                Vec3::from(positions[b]),
-                Vec3::from(positions[c]),
+                Vec3::from(*positions.get(a)?),
+                Vec3::from(*positions.get(b)?),
+                Vec3::from(*positions.get(c)?),
             ];
-            let tri_normals = vertex_normals.map(|normals| {
-                [
-                    Vec3::from(normals[a]),
-                    Vec3::from(normals[b]),
-                    Vec3::from(normals[c]),
-                ]
-            });
+            let tri_normals = vertex_normals
+                .map(|normals| -> Result<[Vec3; 3], ()> {
+                    Ok([
+                        Vec3::from(*normals.get(a).ok_or(())?),
+                        Vec3::from(*normals.get(b).ok_or(())?),
+                        Vec3::from(*normals.get(c).ok_or(())?),
+                    ])
+                })
+                .transpose()
+                .ok()?;
 
             let Some(hit) = triangle_intersection(
                 tri_vertex_positions,
@@ -141,13 +144,15 @@ pub fn ray_mesh_intersection<I: TryInto<usize> + Clone + Copy>(
             };
             let triangle_index = Some(i);
             let tri_vertex_positions = &[Vec3::from(a), Vec3::from(b), Vec3::from(c)];
-            let tri_normals = vertex_normals.map(|normals| {
-                [
-                    Vec3::from(normals[i]),
-                    Vec3::from(normals[i + 1]),
-                    Vec3::from(normals[i + 2]),
-                ]
-            });
+            let tri_normals = vertex_normals
+                .map(|normals| -> Result<[Vec3; 3], ()> {
+                    let c = Vec3::from(*normals.get(i + 2).ok_or(())?);
+                    let b = Vec3::from(*normals.get(i + 1).ok_or(())?);
+                    let a = Vec3::from(*normals.get(i).ok_or(())?);
+                    Ok([a, b, c])
+                })
+                .transpose()
+                .ok()?;
 
             let Some(hit) = triangle_intersection(
                 tri_vertex_positions,
@@ -313,6 +318,7 @@ pub fn ray_aabb_intersection_3d(ray: Ray3d, aabb: &Aabb3d, model_to_world: &Mat4
 #[cfg(test)]
 mod tests {
     use bevy_math::Vec3;
+    use bevy_transform::components::GlobalTransform;
 
     use super::*;
 
@@ -334,6 +340,176 @@ mod tests {
         let triangle = [V2.into(), V1.into(), V0.into()];
         let ray = Ray3d::new(Vec3::ZERO, Dir3::X);
         let result = ray_triangle_intersection(&ray, &triangle, Backfaces::Cull);
+        assert!(result.is_none());
+    }
+
+    #[test]
+    fn ray_mesh_intersection_simple() {
+        let ray = Ray3d::new(Vec3::ZERO, Dir3::X);
+        let mesh_transform = GlobalTransform::IDENTITY.compute_matrix();
+        let positions = &[V0, V1, V2];
+        let vertex_normals = None;
+        let indices: Option<&[u16]> = None;
+        let backface_culling = Backfaces::Cull;
+
+        let result = ray_mesh_intersection(
+            ray,
+            &mesh_transform,
+            positions,
+            vertex_normals,
+            indices,
+            backface_culling,
+        );
+
+        assert!(result.is_some());
+    }
+
+    #[test]
+    fn ray_mesh_intersection_indices() {
+        let ray = Ray3d::new(Vec3::ZERO, Dir3::X);
+        let mesh_transform = GlobalTransform::IDENTITY.compute_matrix();
+        let positions = &[V0, V1, V2];
+        let vertex_normals = None;
+        let indices: Option<&[u16]> = Some(&[0, 1, 2]);
+        let backface_culling = Backfaces::Cull;
+
+        let result = ray_mesh_intersection(
+            ray,
+            &mesh_transform,
+            positions,
+            vertex_normals,
+            indices,
+            backface_culling,
+        );
+
+        assert!(result.is_some());
+    }
+
+    #[test]
+    fn ray_mesh_intersection_indices_vertex_normals() {
+        let ray = Ray3d::new(Vec3::ZERO, Dir3::X);
+        let mesh_transform = GlobalTransform::IDENTITY.compute_matrix();
+        let positions = &[V0, V1, V2];
+        let vertex_normals: Option<&[[f32; 3]]> =
+            Some(&[[-1., 0., 0.], [-1., 0., 0.], [-1., 0., 0.]]);
+        let indices: Option<&[u16]> = Some(&[0, 1, 2]);
+        let backface_culling = Backfaces::Cull;
+
+        let result = ray_mesh_intersection(
+            ray,
+            &mesh_transform,
+            positions,
+            vertex_normals,
+            indices,
+            backface_culling,
+        );
+
+        assert!(result.is_some());
+    }
+
+    #[test]
+    fn ray_mesh_intersection_vertex_normals() {
+        let ray = Ray3d::new(Vec3::ZERO, Dir3::X);
+        let mesh_transform = GlobalTransform::IDENTITY.compute_matrix();
+        let positions = &[V0, V1, V2];
+        let vertex_normals: Option<&[[f32; 3]]> =
+            Some(&[[-1., 0., 0.], [-1., 0., 0.], [-1., 0., 0.]]);
+        let indices: Option<&[u16]> = None;
+        let backface_culling = Backfaces::Cull;
+
+        let result = ray_mesh_intersection(
+            ray,
+            &mesh_transform,
+            positions,
+            vertex_normals,
+            indices,
+            backface_culling,
+        );
+
+        assert!(result.is_some());
+    }
+
+    #[test]
+    fn ray_mesh_intersection_missing_vertex_normals() {
+        let ray = Ray3d::new(Vec3::ZERO, Dir3::X);
+        let mesh_transform = GlobalTransform::IDENTITY.compute_matrix();
+        let positions = &[V0, V1, V2];
+        let vertex_normals: Option<&[[f32; 3]]> = Some(&[]);
+        let indices: Option<&[u16]> = None;
+        let backface_culling = Backfaces::Cull;
+
+        let result = ray_mesh_intersection(
+            ray,
+            &mesh_transform,
+            positions,
+            vertex_normals,
+            indices,
+            backface_culling,
+        );
+
+        assert!(result.is_none());
+    }
+
+    #[test]
+    fn ray_mesh_intersection_incides_missing_vertex_normals() {
+        let ray = Ray3d::new(Vec3::ZERO, Dir3::X);
+        let mesh_transform = GlobalTransform::IDENTITY.compute_matrix();
+        let positions = &[V0, V1, V2];
+        let vertex_normals: Option<&[[f32; 3]]> = Some(&[]);
+        let indices: Option<&[u16]> = Some(&[0, 1, 2]);
+        let backface_culling = Backfaces::Cull;
+
+        let result = ray_mesh_intersection(
+            ray,
+            &mesh_transform,
+            positions,
+            vertex_normals,
+            indices,
+            backface_culling,
+        );
+
+        assert!(result.is_none());
+    }
+
+    #[test]
+    fn ray_mesh_intersection_not_enough_indices() {
+        let ray = Ray3d::new(Vec3::ZERO, Dir3::X);
+        let mesh_transform = GlobalTransform::IDENTITY.compute_matrix();
+        let positions = &[V0, V1, V2];
+        let vertex_normals = None;
+        let indices: Option<&[u16]> = Some(&[0]);
+        let backface_culling = Backfaces::Cull;
+
+        let result = ray_mesh_intersection(
+            ray,
+            &mesh_transform,
+            positions,
+            vertex_normals,
+            indices,
+            backface_culling,
+        );
+
+        assert!(result.is_none());
+    }
+
+    #[test]
+    fn ray_mesh_intersection_bad_indices() {
+        let ray = Ray3d::new(Vec3::ZERO, Dir3::X);
+        let mesh_transform = GlobalTransform::IDENTITY.compute_matrix();
+        let positions = &[V0, V1, V2];
+        let vertex_normals = None;
+        let indices: Option<&[u16]> = Some(&[0, 1, 3]);
+        let backface_culling = Backfaces::Cull;
+
+        let result = ray_mesh_intersection(
+            ray,
+            &mesh_transform,
+            positions,
+            vertex_normals,
+            indices,
+            backface_culling,
+        );
+
         assert!(result.is_none());
     }
 }


BenjaminBrienen (commented):
It looks good in itself, but I'm wondering whether the function should return a Result instead. It has multiple failure modes.

aevyrie:
On an M3 Max, I'm seeing 3% regressions to 1% improvements. Probably noise.

tbillington:
Yea I wondered that too. I would lean towards this `ray_mesh_intersection` returning a `Result` with a simple error enum. That way calling code in bevy_picking can decide to use it or ignore it where it makes sense, but users would have the option to do the intersection themselves and handle errors.

Admittedly, malformed meshes are probably not _too_ common. I'm kinda on the fence about whether the default public api for mesh picking should bother with them or not.

I imagine it would be frustrating if some of your meshes "aren't working" when you try to click on them. To track it down to a malformed mesh might be a little bit of a journey. Especially if you have some kind of user generated content and malformed meshes are making it into the game.

crates/bevy_picking/src/mesh_picking/ray_cast/intersections.rs
```diff
@@ -94,17 +94,20 @@ pub fn ray_mesh_intersection<I: TryInto<usize> + Clone + Copy>(
...
+            let tri_normals = vertex_normals
+                .map(|normals| -> Result<[Vec3; 3], ()> {
+                    Ok([
+                        Vec3::from(*normals.get(a).ok_or(())?),
+                        Vec3::from(*normals.get(b).ok_or(())?),
+                        Vec3::from(*normals.get(c).ok_or(())?),
+                    ])
+                })
+                .transpose()
+                .ok()?;
```
aevyrie, line 110: This looks overcomplicated, I think this is equivalent, but more succinct:
```suggestion
            let tri_normals = vertex_normals.map(|normals| {
                Some([
                    Vec3::from(*normals.get(a)?),
                    Vec3::from(*normals.get(b)?),
                    Vec3::from(*normals.get(c)?),
                ])
            })?;
```

crates/bevy_picking/src/mesh_picking/ray_cast/intersections.rs
```diff
@@ -141,13 +144,15 @@ pub fn ray_mesh_intersection<I: TryInto<usize> + Clone + Copy>(
...
-            });
+            let tri_normals = vertex_normals
+                .map(|normals| -> Result<[Vec3; 3], ()> {
+                    let c = Vec3::from(*normals.get(i + 2).ok_or(())?);
+                    let b = Vec3::from(*normals.get(i + 1).ok_or(())?);
+                    let a = Vec3::from(*normals.get(i).ok_or(())?);
+                    Ok([a, b, c])
+                })
+                .transpose()
+                .ok()?;
```
aevyrie, line 155: Same here. I'd double check the order is correct still:
```suggestion
            let tri_normals = vertex_normals.map(|normals| {
                Some([
                    Vec3::from(*normals.get(i)?),
                    Vec3::from(*normals.get(i + 1)?),
                    Vec3::from(*normals.get(i + 2)?),
                ])
            })?;
```

aevyrie:
> I imagine it would be frustrating if some of your meshes "aren't working" when you try to click on them. To track it down to a malformed mesh might be a little bit of a journey. Especially if you have some kind of user generated content and malformed meshes are making it into the game.

It might be better to `continue` the loop instead of early exiting. That way if a single triangle is degenerate, it doesn't completely break raycasting.

tbillington:
> It might be better to `continue` the loop instead of early exiting. That way if a single triangle is degenerate, it doesn't completely break raycasting.

I'm open to the idea, but I think it's equally as likely to cause false positives.

aevyrie:
I think it makes sense for normals - if the positions are coming back fine, but the normals are broken, you can still cast against the mesh as if it had no normals.

BenjaminBrienen:
If it's possible to make the parameter(s) more strongly typed to avoid dealing with certain failure modes (invalid mesh/triangle), that would be nice. Otherwise, this direction is fine.

crates/bevy_picking/src/mesh_picking/ray_cast/intersections.rs
tbillington: That was my first instinct, but the options are swapped to how you'd expect them to behave.

I've changed it to a match which I think is more straightforward.

crates/bevy_picking/src/mesh_picking/ray_cast/intersections.rs
tbillington: see above

crates/bevy_picking/src/mesh_picking/ray_cast/intersections.rs
```diff
@@ -141,13 +142,15 @@ pub fn ray_mesh_intersection<I: TryInto<usize> + Clone + Copy>(
...
-                ]
-            });
+            let tri_normals = match vertex_normals {
+                Some(normals) => {
+                    let c = Vec3::from(*normals.get(i + 2)?);
+                    let b = Vec3::from(*normals.get(i + 1)?);
+                    let a = Vec3::from(*normals.get(i)?);
+                    Some([a, b, c])
+                }
+                None => None,
```
aevyrie, line 152: This is redundant? You can just `map` this.

crates/bevy_picking/src/mesh_picking/ray_cast/intersections.rs
aevyrie: > the options are swapped to how you'd expect them to behave

What do you mean? This should just be a simple `map` operation.

BenjaminBrienen (approved):

BenjaminBrienen:
Approving for now since it is an improvement

tbillington:
@aevyrie the nesting order of the Option<Option<>> is wrong for the `?`, that's why I had `.transpose` before to swap them.

With the suggestions the tests start failing
image

tbillington:
> If it's possible to make the parameter(s) more strongly typed to avoid dealing with certain failure modes (invalid mesh/triangle), that would be nice. Otherwise, this direction is fine.

I like the idea of a pre-validated Mesh*, but that's a much bigger change :P

crates/bevy_picking/src/mesh_picking/ray_cast/intersections.rs
tbillington: Just to clarify, the return value of `vertex_normals.map()` will be `Option<Option<_>>`. Using `?` on the outer option will exit `ray_mesh_intersection` with `None` if `vertex_normals` is `None`, which is not the desired behaviour.

aevyrie:
That's because it is no longer failing when the normals are missing. Instead, is is continuing to raycast, ignoring the normals - I think this is better behavior. There is one mistake though. The map()?; should be and_then();

tbillington:
Ah, the combo of the two was confusing me. I didn't realise wasn't used for hit detection, I care a lot less about wanting to short circuit if they're wrong now 👍 .

Added a benchmark for the no indices case since it wasn't represented. I tried a few ways of indexing and this had the best perf.

tbillington:
Compared to main now, this branch has between 41-47% worse performance. I'm not 100% sure why the difference is so big from my initial commit of 0-10% improvement to now.

I haven't spend much time attempting to optimise anything, so there's likely some opportunities here. For example only fetching normals if there is a triangle intersection instead of before, etc.

I still think this is a preferable direction. For one, this is a new feature to bevy, so there isn't an established/expected performance baseline. Of course this code should be as fast as possible, but ideally users generate optimised colliders and cast against them in a proper physics sense.

aevyrie:
I don't think we can make any sacrifices in perf here, it is already slower than it should be, and 40-50% is a huge hit. I can poke this a bit more and see if I can find a solution as well.

> I didn't realise normals weren't used for hit detection

Yup! Normals are only used to compute the smooth normal used for lighting, but it doesn't change the actual hit position in space. If triangle normals are missing, it uses the face of the triangle for the normal.

tbillington:
Do you have other benchmarks you're going off? All I can find is a few in `benches`.

As far as I've seen there hasn't been any optimisation done since merging, so I'm curious about the pushback.
How can it be a new feature but already we can't compromise any perf for correctness? Especially if this is going to be used in the bevy editor too. Any bad asset or buggy procedural generation will crash the editor and/or the game. I'm really struggling to understand the rationale.

I think there are quite a few options we have, I mentioned deferring normal resolution until a hit is found. It seems like it would be amenable to some level of parallelization too. Not to mention switching to optimised colliders/LOD over full meshes, or any kind of culling.

Thoughts on an "unchecked_picking" feature flag that would avoid the checked indexing for people who are less risk averse?

aevyrie:
> As far as I've seen there hasn't been any optimisation done since merging, so I'm curious about the pushback. 

I wrote the original impl in mod_raycast. This is in the hottest of hot loops. Any slowdown here is multiplied by any meshes that have their AABB hit. There have been 2ish perf improvements since this was merged. The wins are usually on the order of 5%-20% improvements, it's pretty hard to claw performance here.

tbillington:
> There have been 2ish perf improvements since this was merged.

I didn't look very hard then, woops!

Ah, ok good context :) might be cool if we added a picking stress test example so we have something more concrete for contributors to measure against. From my perspective all I see is microbench perf in an unreleased feature so it's hard to evaluate any perf changes in any real sense.

Good to hear. I guess I was looking too closely at just these functions, didn't see the outer parallelization!

Agree on the physics engine bit!

tbillington:
I might have some time to poke around over the week. If you have any suggestions I could try them out, otherwise will just bash my head against it xD

aevyrie:
> From my perspective all I see is microbench perf in an unreleased feature so it's hard to evaluate perf in any real sense.

That is completely understandable. :) It would be nice to add benchmarks that test the full raycast as well, but when this benchmark was made, there was not parallelization or culling, so it really was everything.

For the record, I think the tests you've added here are invaluable, and I don't mean to put a damper on things. Feel free to ignore nits about code brevity, as long as performance hasn't regressed, this should be merged.

tbillington:
@aevyrie ok with some fiddling I got the perf to be either equivalent or faster.

I changed the benchmark names to something more indicative of what is actually being ran. I didn't change the mesh generation though so we can compare before/after as the mesh size is identical.

It would be interesting to add variation here, for example a bench that all triangles are being intersected vs a more sparse example. I'm not really sure the original intention of `vertices_per_side`.

tbillington:
Hang on I think I might be missing a commit, tests failing 🙃

tbillington:
Ok got even perf for intersection & no cull, but no_intersection is between 20-40% slower still.

It's interesting that `no_intersection` bench on `main` is the only set of tests where the times are an outlier. Looking at all other 5 bench categories they're very close... 🤔

tbillington:
I added a new benchmark for mesh without indices. It seems like the much superior case if you want to pick large meshes.

Picking with indices gets around `315m` triangles/s, without gets `1645m` triangles/s, so it's just over 5x faster.

tbillington:
I renamed the benchmark names to more clearly communicate what they're testing, `vertices` didn't really make much sense to me on a triangle focused operation, especially when you factor in indices.

they look like this now `ray_mesh_intersection/19_602 triangles (10_000 positions, 19_602 indices)`

tbillington:
Thanks for the motivating push to try a bit harder with the perf :) it was a fun exercise

tbillington:
I added 4 new benchmarks, 2 using a `Plane3d` subdivided to different levels (one test using indices, one doesn't), and 2 with lots of quads overlapping (front to back order and back to front order).

It seems the `main` code doesn't handle them as well, and is slower by a significant margin in most cases.\

aevyrie:
If you want to benchmark raycasting multiple objects, I'd recommend using the `Raycast` system param, as it does some important optimizations like AABB bounds testing. I wouldn't put too much effort into microbenchmarking without looking at the impact on the system as a whole.

tbillington:
@aevyrie all the new benches are still a single mesh, just with triangles arranged in certain ways. I didn't want to expand into multiple meshes since I haven't looked at anything above `ray_mesh_intersection`.

aevyrie:
I was referring to the overlapping planes test - it wasn't clear to me what case this is supposed to be testing for - I wouldn't expect to optimize for the case where a mesh is has a bunch of self-overlapping planes. The danger with benchmarks is that is what you end up optimizing for. I'd just like to be cautious if we are adding benches, that they reflect the types of scenarios we want to optimize for.

For example, most meshes look somewhat manifold, and a ray is likely only going to hit a few triangles within a single mesh. If we add a test where a mesh has thousands of overlapping planes, you are now optimizing the code for a case that doesn't exist in the real world, and may end up pushing you to make changes to the code that are detrimental to the common case, e.g. a few triangles hit per mesh.

tbillington:
Yea fair point. It's a bit of a stretch stretch but I guess if you did raycasting against a cpu billboarded particle system it _might_ be close to the overlapping quads 😂 

I didn't really intend for that benchmark to "guide" implementation, more as a stress test for the worst case because I was curious how the implementation would handle it.

Tbh I think the most realistic use case might be the plane one I added, but honestly the best might be to have a range of real meshes, eg the helmet model or some kind of generated heightmap terrain and bench against those.

tbillington:
If you think it's distracting or too far off the intended use case I can disable it by default or remove it?

tbillington:
Any more thoughts @aevyrie ? If you prefer I could try and source another reviewer if you don't have the bandwidth available !

aevyrie (commented):
Sorry, this fell off my radar.

- The benchmarks and tests are a straight upgrade, very appreciated, no notes.
- The changes to the intersection code still gives me pause. The only thing I'd like to see before merging is testing this in a real scene, and the closest we have is the caldera hotel. I'd say that as long as we aren't regressing in this case, I'm happy to approve and merge.
    - In all likelihood, performance will be unchanged, but I'd really like to see a representative case tested before merging. This is a hot path and small tweaks can have surprising impacts to performance, so I'm being pretty cautious.
    - I'm happy to help with this, I can also try doing this myself. I think the easiest way to test this will be using the mesh picking backend, loading in the caldera hotel gltf, and running tracy, while mousing over the same spot in the scene.

aevyrie (approved):
I hooked up bevy's scene viewer, loaded caldera hotel, and compared `main` to your branch, and your branch seems to be faster!

`main`
image

PR: 
image

tbillington:
> The changes to the intersection code still gives me pause. The only thing I'd like to see before merging is testing this in a real scene, and the closest we have is the caldera hotel. I'd say that as long as we aren't regressing in this case, I'm happy to approve and merge.
> 
>   -  In all likelihood, performance will be unchanged, but I'd really like to see a representative case tested before merging. This is a hot path and small tweaks can have surprising impacts to performance, so I'm being pretty cautious.
>   -  I'm happy to help with this, I can also try doing this myself. I think the easiest way to test this will be using the mesh picking backend, loading in the caldera hotel gltf, and running tracy, while mousing over the same spot in the scene.

Yea, we could possibly create some kind of picking stress test, but the trouble is having realistic test data, so it's cool you could test with an actual scene.

alice-i-cecile:
@tbillington @BD103 I'd like to merge this; ping me when merge conflicts are resolved please :)

tbillington:
Not finished yet, but while converting the the overlapping_planes benchmarks the test was failing.

I tried for about an hour to figure out why, it seems moving the y position of the vertices of the plane would cause the ray to never intersect. I can't work out if something is being rotated somewhere, or something else is going on so I'm a little hesitant until I can figure out exactly what is wrong, whether it's just the silly mesh generation code I have or if it's actually something related to my picking refactor.

tbillington:
@BD103 I've covered your requests I'm pretty sure, though I stopped short of fully utilising your `Benchmarks` struct for abstracting even more. I feel like I got most of the value from the helper macro_rules I wrote and passing things into the test functions, but open to suggestions!

> The docs and refactor for p_to_x_norm()
> The logarithmic scaling in the graphs
> Correctly black_box()-ing all inputs
> Testing that it does or does not intersect when cargo test --benches is run
> The bench! macro naming

BD103:
Awesome! I'll do a thorough review of your changes later today, but I did notice some commented-out code. If `Benchmarks` doesn't fit your needs, feel free to discard it! I used it as a well to avoid duplicating code, but that was only because every benchmark tested the same thing with different values.

tbillington:
Yea it was more of a work in progress commit to get y of ur eyes on it. But yes fortifying commented out stuff is next

BD103:
@tbillington In order to not hold up this PR, could you split the benchmark changes into a separate pull request? That way the actual optimizations can be merged while we work on the benchmarks separately.

alice-i-cecile:
@tbillington gentle ping on this. I'd like to get this into 0.16, and I think it's largely ready.

alice-i-cecile:
Cutting from the milestone and putting this up for adoption for now: it's been a couple of weeks without a response but I really do think that this work is valuable.

PR #16151, opened 2024-10-29
Expand `ui` example 

ickshonpe:
# Objective

Expand the `ui` example to show more features. 

# Solution

Add some elements to the `ui` example that demonstrate outlines, border-radius, image sizing and image flipping.

## Showcase

uiexample


alice-i-cecile (approved):
Good to test these. @mockersf would you be okay to move this example into the `testbed` folder? It's really not good learning material.

PR #16157, opened 2024-10-29
Fix blending for CameraOutputMode::Skip

chronicl:
# Objective

Currently, if we have two cameras with the same output texture, one with `CameraOutputMode::Write` and one with `CameraOutputMode::Skip`, it is possible for the `CameraOutputMode::Write` camera to be assigned alpha blending (which is the fallback blending when multiple cameras write to the same output texture), although it is the only camera writing to the output texture. This may or may not happen every restart of the app, because the camera iteration order in prepare_view_upscaling_pipelines isn't consistent. Since this is random behaviour I consider this a bug and didn't add a migration guide.

## Solution

In `prepare_view_upscaling_pipelines` make sure we don't consider cameras with CameraOutputMode::Skip to be outputting something to the output texture.

## Testing

I ran a few examples to make sure nothing obvious is broken. There is no example using CameraOutputMode::Skip, so I only tested the change in my own App where this was relevant, which however isn't public.

### DIFF
--- a/crates/bevy_core_pipeline/src/upscaling/mod.rs
+++ b/crates/bevy_core_pipeline/src/upscaling/mod.rs
@@ -46,28 +46,31 @@ fn prepare_view_upscaling_pipelines(
     let mut output_textures = HashSet::new();
     for (entity, view_target, camera) in view_targets.iter() {
         let out_texture_id = view_target.out_texture().id();
-        let blend_state = if let Some(ExtractedCamera {
-            output_mode: CameraOutputMode::Write { blend_state, .. },
-            ..
-        }) = camera
-        {
-            match *blend_state {
-                None => {
-                    // If we've already seen this output for a camera and it doesn't have a output blend
-                    // mode configured, default to alpha blend so that we don't accidentally overwrite
-                    // the output texture
-                    if output_textures.contains(&out_texture_id) {
-                        Some(BlendState::ALPHA_BLENDING)
-                    } else {
-                        None
+        let blend_state = if let Some(extracted_camera) = camera {
+            match extracted_camera.output_mode {
+                CameraOutputMode::Skip => None,
+                CameraOutputMode::Write { blend_state, .. } => {
+                    let already_seen = output_textures.contains(&out_texture_id);
+                    output_textures.insert(out_texture_id);
+
+                    match blend_state {
+                        None => {
+                            // If we've already seen this output for a camera and it doesn't have a output blend
+                            // mode configured, default to alpha blend so that we don't accidentally overwrite
+                            // the output texture
+                            if already_seen {
+                                Some(BlendState::ALPHA_BLENDING)
+                            } else {
+                                None
+                            }
+                        }
+                        _ => blend_state,
                     }
                 }
-                _ => *blend_state,
             }
         } else {
             None
         };
-        output_textures.insert(out_texture_id);
 
         let key = BlitPipelineKey {
             texture_format: view_target.out_texture_format(),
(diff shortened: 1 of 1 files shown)


crates/bevy_core_pipeline/src/upscaling/mod.rs
```diff
@@ -46,28 +46,32 @@ fn prepare_view_upscaling_pipelines(
...
+                                None
+                            }
+                        }
+                        _ => blend_state,
                     }
                 }
-                _ => *blend_state,
             }
         } else {
+            output_textures.insert(out_texture_id);
```
tychedelia, line 72: I think the idea here was to configure a sensible default blend mode in the event we were writing on top of the same output texture in the csae of `CameraOutputMode::Write { blend_state: None, .. }`. If I'm working through the logic correctly, moving this check would mean we'd only configure alpha blending in the following scenario:
- Camera 1: `CameraOutputMode::Skip`
- Camera 2: `CameraOutputMode::Write { blend_state: None }`

tychedelia (commented):
Thanks for noticing this -- it definitely makes sense that's a defect. I think part of the fix here is that we need to enforce a stable iteration order as we do for `SortedCameras`, although the bug you identify where we skip and then write seems like it would be an issue either way.

crates/bevy_core_pipeline/src/upscaling/mod.rs
chronicl: No, in that scenario alpha blending wouldn't be configured, because we do not insert the `out_texture_id` into `output_textures` for Camera 1, because it has `CameraOutputMode::Skip`, so `already_seen` would be false for Camera 2.

Alpha blending is configured for a camera A if it's `blend_state` is None and we've iterated over a `ViewTarget` that is NOT a camera with `CameraOutputMode::Skip` and has the same output texture as camera A. So for example:
- Camera 1: CameraOutputMode::Write { blend_state: ... }
- Camera 2: CameraOutputMode::Write { blend_state: None }

where both cameras have the same output texture would lead to Camera 2 getting a `blend_state` of alpha blending.

chronicl:
Using `SortedCameras` for the iteration makes sense, I would like to add that to this PR, but I have some questions:

This iteration isn't only over cameras, it's over any `ViewTarget`. 
1. Should a `ViewTarget` that isn't a camera but has the same output texture as a camera, make it so that the camera is configured with alpha blending? This is what I added back in with the second commit, because it was the case before.
2. If we use `SortedCameras` for iteration order, should we handle the `ViewTarget`s that aren't cameras before or after the cameras? If the answer to 1. is no, this doesn't matter.

BenjaminBrienen (approved):

tychedelia:
> Using `SortedCameras` for the iteration makes sense, I would like to add that to this PR, but I have some questions:
> 
> 
> 
> This iteration isn't only over cameras, it's over any `ViewTarget`. 
> 
> 1. Should a `ViewTarget` that isn't a camera but has the same output texture as a camera, make it so that the camera is configured with alpha blending? This is what I added back in with the second commit, because it was the case before.
> 
> 2. If we use `SortedCameras` for iteration order, should we handle the `ViewTarget`s that aren't cameras before or after the cameras? If the answer to 1. is no, this doesn't matter.

Yes, we wouldn't be able to use `SortedCameras` but would need something similar for views I think...

chronicl:
> Yes, we wouldn't be able to use `SortedCameras` but would need something similar for views I think...

True.
I think for now I'd like to just get this PR in as it's relatively uncontroversial, because skipping cameras with `CameraOutputMode::Skip` is correct, and it fixes my immediate issue.

A solution for the sorting problem could be to require a new `OutputOrder` component for `ViewTarget` and remove the `order` field from `ExtractedCamera`, but this will be more controversial and there is other solutions too, so I'd like to make it a separate PR/issue.

tychedelia:
> I'd like to make it a separate PR/issue.

Okay, sounds good to me. This is pretty minor all things considered so I'm happy with landing the more ad-hoc fix.

tychedelia (approved):

PR #16158, opened 2024-10-29
Convert mod structure

BenjaminBrienen:
# Objective

Use the new rust module file structure. One advantage is that there aren't hundreds of files all named `mod.rs`, making file searching more difficult.

## Solution

Use https://github.com/BenjaminBrienen/mod_rs_converter on bevy and then clean up a few hardcoded string file paths.

## Testing

It builds the same.


JMS55:
I'm in favor of this. Will need to be done shortly after 0.15 releases otherwise merge conflicts are going to be painful.

alice-i-cecile:
I prefer this style too. `mod.rs` is really annoying to work on in your IDE and tricky to figure out the context for during review.

tychedelia:
Just wanna lodge a vote against. I find renames mildly annoying in git and prefer having the module be contained within the folder it represents. I appreciate the arguments for IDE tooling but have never run into it personally.

JMS55:
Ah I didn't realize that it moves the mod.rs file outside of the folder it's a part of... Yeah I like this less now :/

BenjaminBrienen:
It's idiomatic https://doc.rust-lang.org/rust-by-example/mod/split.html

alice-i-cecile:
Oh right, you can't have it in the same folder. I just want to have `foo/foo.rs` :( Mildly against on the whole.

MrGVSV:
FWIW I prefer using `mod.rs`. A few reasons:

- Changing/adding a module can all be done within a single directory. I find this easier to work with when using git (e.g. it makes it _slightly_ faster to stage) and a bit easier to reason about regarding the scope of your changes.
- It makes it unambiguously clear if a file is the root module or a child module. I don't have to dart my eyes up and down the file tree to see if `foo.rs` exists as `foo/`, which is especially annoying for larger directories (although I suppose this isn't as much of a problem if you like to have your files and folders inter-mingle in the file tree).
- Similar to the above two, if I want to navigate to the root module, I can quickly tap on the `mod.rs` of the directory I'm in without having to pop out of that directory and look for the associated root module file.
- It makes some operations a bit easier. For example, if I wanted to find all serde-related root modules, I could search for `serde` with a `**/mod.rs` file filter. Or if you wanted to quickly see how all the modules in `bevy_foo` are defining their tests, you could search for `#[test]` with the `bevy_foo/**/mod.rs` file filter.
- I'm a TypeScript dev and all we got is `index.ts` lol

I'm not really that picky with this decision: I'll use whatever the community decides is best. But just wanted to share some additional thoughts on the matter :)

IceSentry:
I'm gonna add a vote against too. I heavily dislike how it moves the mod.rs file outside of the folder. I really don't know why the rust team went in this direction. I get not wanting a bunch of mod.rs files, but having the file away from the folder is IMO a worse compromise. It really should have been something like having the same name for the folder and the .rs file inside the folder.

mockersf:
Against this too, as most IDEs sort separately folders and files it means the file is not near its folder.

And it doesn't really help with file search, as you can easily specify the folder while searching for a file so it already works fine

BenjaminBrienen:
Rest in peace

MrGVSV:
Not that both sides can't be solved/improved with a little change in IDE settings and such, but I'm not fully convinced of the problem of having multiple `mod.rs` files open or searching being too chaotic since most editors will qualify tabs/search results with the containing folder.

PR #16160, opened 2024-10-29
Improve `SubApp` documentation example

### LINKED ISSUE #15841: sub app cannot run as the official example suggested
## Bevy version
0.14.2

## What you did

I tried to run some code in sub app, then I use the example from official documentation:

```rust
use bevy_app::{App, AppLabel, Main, SubApp};
use bevy_ecs::system::{Res, Resource};

fn main() {
#[derive(Resource, Default)]
struct Val(pub i32);

#[derive(Debug, Clone, Copy, Hash, PartialEq, Eq, AppLabel)]
struct ExampleApp;

// Create an app with a certain resource.
let mut app = App::new();
app.insert_resource(Val(10));

// Create a sub-app with the same resource and a single schedule.
let mut sub_app = SubApp::new();
sub_app.insert_resource(Val(100));

// Setup an extract function to copy the resource's value in the main world.
sub_app.set_extract(|main_world, sub_world| {
    sub_world.resource_mut::<Val>().0 = main_world.resource::<Val>().0;
});

// Schedule a system that will verify extraction is working.
sub_app.add_systems(Main, |counter: Res<Val>| {
    // The value will be copied during extraction, so we should see 10 instead of 100.
    println!("system of subapp is executing and the Counter: {}", counter.0);
    assert_eq!(counter.0, 10);
});

// Add the sub-app to the main app.
app.insert_sub_app(ExampleApp, sub_app);

// Update the application once (using the default runner).
app.update();

}
```

## What went wrong

The function in sub app can never be executed

BenjaminBrienen:
# Objective

Fix #15841

## Solution

Added an update schedule as recommended in the issue.

## Testing

Doc test is run and passes.

Ran the example in a test app before and after adding the line.

## Showcase

Before:

```
PS C:\Users\BenjaminBrienen\source\bevy_experiment_test> cargo run
    Blocking waiting for file lock on build directory
   Compiling bevy_experiment_test v0.1.0 (C:\Users\BenjaminBrienen\source\bevy_experiment_test)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 1m 41s
     Running `target\debug\bevy_experiment_test.exe`
```

(nothing happens)

After:

```
PS C:\Users\BenjaminBrienen\source\bevy_experiment_test> cargo run
    Blocking waiting for file lock on build directory
   Compiling bevy_experiment_test v0.1.0 (C:\Users\BenjaminBrienen\source\bevy_experiment_test)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 14.64s
     Running `target\debug\bevy_experiment_test.exe`
system of subapp is executing and the Counter: 10
```

### DIFF
--- a/crates/bevy_app/src/sub_app.rs
+++ b/crates/bevy_app/src/sub_app.rs
@@ -37,6 +37,7 @@ type ExtractFn = Box<dyn Fn(&mut World, &mut World) + Send>;
 ///
 /// // Create a sub-app with the same resource and a single schedule.
 /// let mut sub_app = SubApp::new();
+/// sub_app.update_schedule = Some(Main.intern());
 /// sub_app.insert_resource(Val(100));
 ///
 /// // Setup an extract function to copy the resource's value in the main world.


alice-i-cecile (approved):

alice-i-cecile:
@RamonBrisk can I get your review here?

mockersf (approved):

RamonBrisk:
> @RamonBrisk can I get your review here?

This PR effectively improves the SubApp documentation example (#15841) by adding an update schedule, which resolves the previous unresponsive issue in the example.

PR #16161, opened 2024-10-29
Move `all_tuples` to a new crate

### LINKED ISSUE #15941: Move `all_tuples` and friends into its own crate
FYI I'm merging the simple fix in #15933 now, but I do think we should move fully to this. It does make me think we should pull all_tuples out into its own crate though: I'm not a fan of bevy_math relying on bevy_utils.

_Originally posted by @alice-i-cecile in https://github.com/bevyengine/bevy/issues/15936#issuecomment-2415352291_

BenjaminBrienen:
# Objective

Fixes #15941

## Solution

Created https://crates.io/crates/variadics_please and moved the code there; updating references

`bevy_utils/macros` is deleted.

## Testing

cargo check

## Migration Guide

Use `variadics_please::{all_tuples, all_tuples_with_size}` instead of `bevy::utils::{all_tuples, all_tuples_with_size}`.


BenjaminBrienen:
Blocked on the creation of the bevyengine repository.

alice-i-cecile (approved):

alice-i-cecile:
https://github.com/bevyengine/variadics_please now exists :)

bushrat011899 (approved):
Nice to make `bevy_utils` even slimmer!

PR #16163, opened 2024-10-30
Don't reëxport `bevy_image` from `bevy_render`

### LINKED ISSUE #15940: Remove `bevy_image` re-exports from `bevy_render`

Currently, `bevy_render` re-exports everything from `bevy_image`. This means users may be accidentally using `bevy_render` in cases where they only need `bevy_image`.

Remove the `bevy_image` re-exports from `bevy_render`!

Do nothing. Currently users still have to add the `ImagePlugin` through `bevy_render`, so using `Image` through it is also fine.

## Additional context

This should be deferred to 0.16, to ease the migration due to #15650.

BenjaminBrienen:
# Objective

Fixes #15940

## Solution

Remove the `pub use` and fix the compile errors.
Make `bevy_image` available as `bevy::image`.

## Testing

Feature Frenzy would be good here! Maybe I'll learn how to use it if I have some time this weekend, or maybe a reviewer can use it.

## Migration Guide

Use `bevy_image` instead of `bevy_render::texture` items.


crates/bevy_render/src/lib.rs
```diff
@@ -61,12 +61,11 @@ pub mod prelude {
         },
         render_resource::Shader,
         spatial_bundle::SpatialBundle,
-        texture::{Image, ImagePlugin, IntoDynamicImageError},
```
bushrat011899, line 64: `Image` and `IntoDynamicImageError` should be added to a new `bevy_image::prelude` module, and that prelude should be included by the main `bevy::prelude` (just like all the other member crates) to avoid some of the changes in the examples.

bushrat011899 (changes requested):
Looks good! A flatter crate structure will help with compile times and it reduces the coupling between these crates. I do have a suggestion around adding a `prelude` to `bevy_image`, so that `Image` and co. can still be included in the main `bevy::prelude`, but that's a minor change and would help with the migration guide here too.

crates/bevy_render/src/lib.rs
BenjaminBrienen: Ok. Just those two items? I was thinking that `BevyDefault` is one of the more widely used items and could also be in a prelude.

alice-i-cecile:
Definitely agree on a prelude + re-exporting. For 0.15, @atlv24 made the decision not to do this to ease migration, but I 100% agree that this is a better design in the long-term.

crates/bevy_render/src/lib.rs
bushrat011899: That seems reasonable. I only mentioned those two since they are currently a part of the `bevy_render` prelude. Since `BevyDefault` is used a lot it seems reasonable to include too, especially with the `as _` trick.

BD103:
> Feature Frenzy would be good here! Maybe I'll learn how to use it if I have some time this weekend, or maybe a reviewer can use it.

`flag-frenzy` is [a bit broken right now](https://github.com/TheBevyFlock/flag-frenzy/actions) because some coupling changes to rendering crates forced `x11` or `wayland` to be specified for more crates than the CI is currently configured for. (And I believe the features are for sub-dependencies, like `bevy_render/x11`, which `flag-frenzy` doesn't currently support.)

So feel free to try it! The [docs](https://github.com/TheBevyFlock/flag-frenzy/tree/main/docs) are great, but you may hit a wall or two. (I'll work on improving it once `bevy_lint` gets off the ground.)

atlv24 (approved):
i will admit that i didnt make these changes mostly out of laziness but my excuse is that it kept my diff smaller so it was easier to review and merge. this is the way it *should* be done. incremental prs are good. thanks!

BenjaminBrienen:
I'll address the feedback about the prelude this weekend.

bushrat011899 (approved):
Excellent work!

alice-i-cecile:
@mockersf want this in 0.15 or 0.16?

BenjaminBrienen:
Should export Volume according to https://github.com/bevyengine/bevy/pull/16252

cart:
> @mockersf want this in 0.15 or 0.16?

I personally think we should do this for 0.15. The introduction of the new crate is already "breaking" in a way. We might as well finish the job.

cart:
Just pushed some minor changes:

1. Removed a lot of the prelude usage in places where we prefer direct imports
2. Removed Volume from prelude. The original issue was more concerned with Volume not being made available _at all_ not that it needs prelude membership. It being exported from bevy_image in the normal spot is good enough.
3. Removed a couple of unnecessary imports

cart:
4. I also removed a couple of instances of "feature gated imports". They had "redundant imports" in them, and I generally think its better to just specify full paths directly where they are used.

cart (approved):

crates/bevy_internal/src/lib.rs
```diff
@@ -37,6 +37,7 @@ pub use bevy_gizmos as gizmos;
 #[cfg(feature = "bevy_gltf")]
 pub use bevy_gltf as gltf;
 pub use bevy_hierarchy as hierarchy;
+pub use bevy_image as image;
```
mockersf, line 40:
```suggestion
#[cfg(feature = "bevy_image")]
pub use bevy_image as image;
```

crates/bevy_internal/src/prelude.rs
```diff
@@ -1,7 +1,7 @@
 #[doc(hidden)]
 pub use crate::{
-    app::prelude::*, core::prelude::*, ecs::prelude::*, hierarchy::prelude::*, input::prelude::*,
-    log::prelude::*, math::prelude::*, reflect::prelude::*, time::prelude::*,
+    app::prelude::*, core::prelude::*, ecs::prelude::*, hierarchy::prelude::*, image::prelude::*,
+    input::prelude::*, log::prelude::*, math::prelude::*, reflect::prelude::*, time::prelude::*,
```
mockersf, line 4:
```suggestion
    app::prelude::*, core::prelude::*, ecs::prelude::*, hierarchy::prelude::*, input::prelude::*,
    log::prelude::*, math::prelude::*, reflect::prelude::*, time::prelude::*,
```

and you'll need to add

```
#[doc(hidden)]
#[cfg(feature = "bevy_image")]
pub use crate::image::prelude::*;
```

mockersf (changes requested):
`bevy_image` is optional so it needs to be feature gated

mockersf (approved):

PR #16169, opened 2024-10-30
Fix pointer constructor order

NthTensor:
Fixes a small divergence between `bevy_mod_picking` and the up-streamed `bevy_picking`: Both have a `Pointer<E>` constructor with the same types, but in a different order.

This is part of work being done on `bevy_mod_picking` to simplify the migration to `bevy_picking`.


alice-i-cecile (approved):
Merging as trivial.

PR #16173, opened 2024-10-30
Remove the invalid system ordering in the animation example.

chendaohan:
# Objective

In the animation example, there is the code `.add_systems(Update, init_animations.before(animate_targets))`, where `animate_targets` is added to the `PostUpdate` in the `AnimationPlugin`. Therefore, the `.before(animate_targets)` here is ineffective and should be removed.


alice-i-cecile (approved):
Merging as trivial. Thanks for finding and opening this!

mockersf (approved):

PR #16176, opened 2024-10-30
Add `button_just_down` and `button_just_up` methods to `PointerInput`

NthTensor:
# Objective

Upstream two small methods present in `bevy_mod_picking` but not in `bevy_picking`.

There might be a few more of these.

### DIFF
--- a/crates/bevy_picking/src/pointer.rs
+++ b/crates/bevy_picking/src/pointer.rs
@@ -286,6 +286,26 @@ impl PointerInput {
         }
     }
 
+    /// Returns true if the `button` of this pointer was just pressed.
+    #[inline]
+    pub fn button_just_down(&self, target_button: PointerButton) -> bool {
+        if let PointerAction::Pressed { direction, button } = self.action {
+            direction == PressDirection::Down && button == target_button
+        } else {
+            false
+        }
+    }
+
+    /// Returns true if the `button` of this pointer was just released.
+    #[inline]
+    pub fn button_just_up(&self, target_button: PointerButton) -> bool {
+        if let PointerAction::Pressed { direction, button } = self.action {
+            direction == PressDirection::Up && button == target_button
+        } else {
+            false
+        }
+    }
+
     /// Updates pointer entities according to the input events.
     pub fn receive(
         mut events: EventReader<PointerInput>,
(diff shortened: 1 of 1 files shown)


BenjaminBrienen (approved):

alice-i-cecile (approved):

hymm:
should these be called `pressed` and `released` instead of `down` and `up` to match the methods on `ButtonInput`? https://dev-docs.bevyengine.org/bevy/input/struct.ButtonInput.html#method.just_pressed

alice-i-cecile:
I would prefer that, but for ease of migration, let's keep the names for now.

NthTensor:
This is not exactly the same api as in `mod_picking` so exact migration replication isn't needed. These get around `PointerAction` not being `Eq` for some common cases.

I'll change them to `pressed/released`.

cart:
> should these be called pressed and released instead of down and up to match the methods on ButtonInput? https://dev-docs.bevyengine.org/bevy/input/struct.ButtonInput.html#method.just_pressed

If we're doing this, it should be a global concept rename (ex: `Pointer<Down>` and `Pointer<Up>` types, all of the methods)

I'm guessing the Down / Up vs Pressed / Released conversation has already happened somewhere, so we should make sure we aren't retreading ground.

NthTensor:
> I'm guessing the Down / Up vs Pressed / Released conversation has already happened somewhere, so we should make sure we aren't retreading ground.

It has not, actually! I'm all in favor, but I'd rather not do this in `0.15`. The migration from the old crate will already be somewhat complex without renaming some of the events.

cart:
@aevyrie any strong opinions here?

cart:
We do currently have `PointerAction::Pressed`, which can either be "pressed" down or up, which I do find a little confusing given that "press" is generally thought to be a "down" action (at least in my experience).

alice-i-cecile:
Can we rename `PointerAction::Pressed` to `PointerAction::Button`?

cart:
This also seems to conflict with the `PointerPress` component, which uses bool flag indicators (presumably true == Down and false == Up).

NthTensor:
> Can we rename PointerAction::Pressed to PointerAction::Button?

Yep. That along with `Pressed/Released` is consistent with winit's new pointer api as well.

NthTensor:
I made the suggestion changes to the function names. Can we merge this and leave the other suggestions for follow up (ideally slated for 0.16)?

BenjaminBrienen (approved):

PR #16177, opened 2024-10-30
Remove the `Globals` binding from the box shadow shader

ickshonpe:
# Objective

Remove the `Globals` binding from the box shadow shader. It isn't used and was added by mistake.


alice-i-cecile (approved):
Dead code analysis in shaders please :(

IceSentry (approved):

PR #16180, opened 2024-10-30
enable_state_scoped_entities() as a derive attribute

mockersf:
# Objective

- I got tired of calling `enable_state_scoped_entities`, and though it would make more sense to define that at the place where the state is defined

## Solution

- add a derive attribute `#[states(scoped_entities)]` when derive `States` or `SubStates` that enables it automatically when adding the state

## Testing

- Ran the examples using it, they still work

### DIFF
--- a/crates/bevy_state/src/app.rs
+++ b/crates/bevy_state/src/app.rs
@@ -104,7 +104,7 @@ impl AppExtStates for SubApp {
                 exited: None,
                 entered: Some(state),
             });
-            if S::scoped_entities_enabled() {
+            if S::SCOPED_ENTITIES_ENABLED {
                 self.enable_state_scoped_entities::<S>();
             }
         } else {
@@ -129,7 +129,7 @@ impl AppExtStates for SubApp {
                 exited: None,
                 entered: Some(state),
             });
-            if S::scoped_entities_enabled() {
+            if S::SCOPED_ENTITIES_ENABLED {
                 self.enable_state_scoped_entities::<S>();
             }
         } else {
@@ -166,7 +166,7 @@ impl AppExtStates for SubApp {
                 exited: None,
                 entered: state,
             });
-            if S::scoped_entities_enabled() {
+            if S::SCOPED_ENTITIES_ENABLED {
                 self.enable_state_scoped_entities::<S>();
             }
         } else {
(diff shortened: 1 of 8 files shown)


crates/bevy_state/src/state/states.rs
```diff
@@ -64,4 +64,7 @@ pub trait States: 'static + Send + Sync + Clone + PartialEq + Eq + Hash + Debug
     /// Used to help order transitions and de-duplicate [`ComputedStates`](crate::state::ComputedStates), as well as prevent cyclical
     /// `ComputedState` dependencies.
     const DEPENDENCY_DEPTH: usize = 1;
+
+    /// Should scoped entities be enabled for this state?
```
MrGVSV, line 68: This should probably backlink to `StateScoped` or `App::enable_state_scoped_entities`.

crates/bevy_state/src/app.rs
```diff
@@ -104,6 +104,9 @@ impl AppExtStates for SubApp {
                 exited: None,
                 entered: Some(state),
             });
+            if S::scoped_entities_enabled() {
+                self.enable_state_scoped_entities::<S>();
```
MrGVSV, line 108: The docs for `enable_state_scoped_entities` should probably be updated to mention that it doesn't need to be called manually if using the attribute.

aecsocket (approved):

crates/bevy_state/src/state_scoped.rs
```diff
@@ -16,8 +16,7 @@ use crate::state::{StateTransitionEvent, States};
 /// Entities marked with this component will be removed
 /// when the world's state of the matching type no longer matches the supplied value.
 ///
-/// To enable this feature remember to configure your application
-/// with [`enable_state_scoped_entities`](crate::app::AppExtStates::enable_state_scoped_entities) on your state(s) of choice.
+/// To enable this feature remember to add the attribute `#[states(scoped_entities)]` when deriving [`States`].
```
aecsocket, line 19: This could still refer to `enable_state_scoped_entities` as an alternative to the macro attribute

BenjaminBrienen (approved):

andriyDev (approved):

PR #16181, opened 2024-10-30
UI anti-aliasing fix

ickshonpe:
# Objective

UI Anti-aliasing is incorrectly implemented. It always uses an edge radius of 0.25 logical pixels, and ignores the physical resolution. For low dpi screens 0.25 is is too low and on higher dpi screens the physical edge radius is much too large, resulting in visual artifacts.

## Solution

Multiply the distance by the scale factor in the `antialias` function so that the edge radius stays constant in physical pixels.

## Testing

To see the problem really clearly run the button example with `UiScale` set really high. With `UiScale(25.)` on main if you examine the button's border you can see a thick gradient fading away from the edges:

edgg

With this PR the edges are sharp and smooth at all scale factors: 

edge


mockersf:
@ickshonpe is this ready or is it still a draft?

BenjaminBrienen (approved):

ickshonpe:
> @ickshonpe is this ready or is it still a draft?

Ready for review now, though the implementation sucks so much as it passes in the inverse scale factor with every vertex :sweat: That can be removed later by switching to physical coordinates internally.

I've got some before and after test examples but I need to go to sleep, I'll put them up when I wake up.

mockersf (approved):

PR #14520, opened 2024-07-29
Add `OnMutate` observer and demonstrate how to use it for UI reactivity

alice-i-cecile:
# Objective

`OnMutate` hooks and observers are a natural and much desired complement to the hooks and observers feature added in #10839. These were cut from that initial PR due to a) complexity and b) performance concerns, but are incredibly valuable as a general purpose tool for tracking and responding to changes.

P.S. As discussed in my [Vision for Bevy UI](https://hackmd.io/@bevy/HkjcMkJFC) and @cart's #14437, reactivity is a major open question for building complex UI in Bevy.

This PR will not cover their uses within UI to avoid getting bogged down in controversy, but my hope is to add a complex example or two demonstrating this in follow-up work.

## Solution

Users can now watch specific components for mutations by calling `app.generate_on_mutate::<MyComponent>>`.

When this is enabled, `OnMutate` triggers will be sent whenever a component of that type is added or mutated, targeting the entity whose component has been changed.

This is a useful building block for reactive UI when combined with observers:

```rust
    commands
        .spawn(TextBundle { ..default() })
        .insert(CounterValue(0))
        .observe(
            |trigger: Trigger<OnMutate, CounterValue>,
             mut query: Query<(&CounterValue, &mut Text)>| {
                let (counter_value, mut text) = query.get_mut(trigger.entity()).unwrap();
                *text = Text::from_section(counter_value.0.to_string(), TextStyle::default());
            },
        )
```

## To do

- [ ] add an ECS example
- [ ] cut the UI example
- [ ] make the simple example less ugly
- [ ] move the method on `App` into `bevy_app`
- [ ] see if we can make an `OnMutate` hook too
- [ ] move the `OnMutate` event generation into the ECS internals, and remove its generic
- [ ] create an equivalent `generate_change_events` API on `World`

## Follow-up work

To make this pattern broadly useful, we need:

1. Resource equivalents to all of the lifecycle hooks (#12231).
2. Asset change detection (#14444).
3. Broader unification of the two forms of events, so observers can be triggered by batched events (sent with an `EventWriter`).

To make this *great*, I think we want:

1. Archetype-level (#5097) or query-level (#14510) change tracking to make these updates fast.
2. A way to automatically detect whether or not change detection events are needed based on the registered observers (related to #4882).
3. Better tools to gracefully exit observers when standard operations fail.
4. Relations to make defining entity-links more robust and simpler.
5. Nicer picking events to avoid having to use the naive `OnMutate<Interaction>` pattern (#12365).
[...]

PR #16183, opened 2024-10-31
Mutate Observer

pyhinox:
# Objective

* Adopts https://github.com/bevyengine/bevy/pull/14520 and adds OnMutate observers.
* Provide a mechanism for immediate changes in response to mutate event.

## Solution

  A `EntityChanges` which stores the change list is added to `World`. When `Mut::set_changed` is called, a change record will be added to the `EntityChanges`. 
  When `World::flush` is called, `EntityChanges` will be clear and trigger `OnMutate` event to each relative entity.

## Example
```rust
world.observe(|trigger: Trigger<OnMutate, Transform>, query: Query<&Transform>| {
    let transform = query.get(trigger.entity()).unwrap();
});
```

## Testing

    Just test for the base function, I don't know which test i should do.
    `Query::iter` and `Query::get_mut`'s benchmark
     Just treat the `OnMutate` as a common observer. Nothing is specific.
    Mac

### Possible Follow-ups
1. Add an `OnMutate` hook


pyhinox:
@alice-i-cecile Hi alice, i thought this pr is ready to review now. What should I do next?

BenjaminBrienen:
@pyhinox there are some merge conflicts, but I'll go ahead and mark it for review to start getting eyes on it. 🙂

mockersf:
Could you use `core::cell::RefCell` instead of `std::cell::RefCell`?

CI is failing for a few invalid links in documentation, mostly things that are not in scope for automatic doc links

BenjaminBrienen:
I thought we had a lint enabled to catch that as well. Good eye.

alice-i-cecile:
I think that we can use #16208 to remove the need for this. If components can't be mutated, we can do all of our reactive logic in the insertion / removal stage 🎉

bushrat011899:
> I thought we had a lint enabled to catch that as well. Good eye.

I don't believe clippy lints apply to documentation sadly. I got bit by that when adding the lints!

JMS55:
@alice-i-cecile how does #16208 remove the need for this? What happens if you _want_ to observe a component changing?

bushrat011899:
> @alice-i-cecile how does #16208 remove the need for this? What happens if you _want_ to observe a component changing?

I believe the idea was if you want to watch a component, make that component immutable so that the current hooks can capture all state changes. It doesn't cover the case where you want to observe a component which _isn't_ immutable though.

alice-i-cecile:
> I believe the idea was if you want to watch a component, make that component immutable so that the current hooks can capture all state changes.

Precisely.

> It doesn't cover the case where you want to observe a component which isn't immutable though.

Yeah. Ultimately, virtually all of the same drawbacks apply to the immutable components approach to this problem:

- can only be safely mutated during exclusive world access to ensure side effects are processed immediately
- changes are relatively expensive
- must be opted in (the current design doesn't seem to do this, but will incur an unacceptable global perf penalty)

I'd like to see if we can get away with only using a single new concept to achieve these benefits, and then re-evaluate once we have a clearer picture of the final performance and ergonomics.

cart:
> on reflection, I think that this is probably still useful, because it allows users to do reactivity things with any arbitrary component, which is really valuable for UI. I suspect that we should have on-mutate observers fill a "not very strict + deferred reactions" niche, while reserving immutable components + hooks for really critical invariants that must be constantly upheld.

Yeah I'm starting to want something like this as part of our reactivity solution, over less event driven / poll based approaches (tbd though).

But yeah the current unconditional impl should not be merged. If we land this, we can't affect the default Mut path. That is hot / costly enough as it is.

The `Option` also appears to exist to accommodate resources. But I'd rather stopping using Mut with resources (in favor of ResMut or alternatives) than pay the unnecessary size cost of the Option for resources and the unnecessary branching cost of the option for components.

ChristopherBiscardi:
this is a bit contentious, perf sensitive, hasn't gotten the relevant approvals in a number of years, and doesn't look like it will merge in its current implementation here. Reviving it would require care, but I'll close it for now.

PR #16198, opened 2024-11-01
BRP System Ordering

### LINKED ISSUE #16042: System ordering around BRP (Bevy Remote Protocol)

Currently, Bevy's Remote Plugin adds its systems to the `Update` schedule (which is honestly quite confusing ... wasn't `Update` meant strictly for "user code" and not "engine code"?). Further, there is no system set (like most other bevy internals have) to allow users to order their systems before/after BRP.

Given that the whole point of BRP is to allow monitoring and messing with ECS data, the impact of this is quite severe. As is, BRP always runs with ambiguous/undefined/racy ordering with respect to all user systems in `Update`. Every frame, it may or may not see changes to component values made by various user systems.

Honestly, BRP should probably not be in `Update` in the first place.

Regardless of which schedule it runs in, there should be a system set, to allow people to order systems around it.

There aren't any afaik?

mnmaita:
# Objective

- Attempts to fix #16042

## Solution

- Added a new `RemoteSystem` `SystemSet` for the BRP systems.
- Changed the schedule on which these systems run from `Update` to `Last`.

## Testing

- I did not test these changes and would appreciate a hand in doing so. I assume it would be good to test that you can order against these systems easily now.

---

## Migration Guide

- `process_remote_requests`, `process_ongoing_watching_requests` and `remove_closed_watching_requests` now run in the `Last` schedule. Make sure you use `RemoteSystem` `SystemSet` in case you need to order your systems against them.

### DIFF
--- a/crates/bevy_remote/src/lib.rs
+++ b/crates/bevy_remote/src/lib.rs
@@ -438,7 +438,7 @@ impl Plugin for RemotePlugin {
             .init_resource::<RemoteWatchingRequests>()
             .add_systems(PreStartup, setup_mailbox_channel)
             .add_systems(
-                Update,
+                Last,
                 (
                     process_remote_requests,
                     process_ongoing_watching_requests,
--- a/crates/bevy_remote/src/lib.rs
+++ b/crates/bevy_remote/src/lib.rs
@@ -301,11 +301,11 @@
 //! [fully-qualified type name]: bevy_reflect::TypePath::type_path
 
 use async_channel::{Receiver, Sender};
-use bevy_app::prelude::*;
+use bevy_app::{prelude::*, MainScheduleOrder};
 use bevy_derive::{Deref, DerefMut};
 use bevy_ecs::{
     entity::Entity,
-    schedule::{IntoSystemConfigs, SystemSet},
+    schedule::{IntoSystemConfigs, ScheduleLabel},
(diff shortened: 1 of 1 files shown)


BenjaminBrienen (approved):

MiniaczQ:
I mentioned it in the issue, but I'd consider adding a custom schedule for this, to ensure BRP-unaware users don't mix their systems with BRP systems.
In that case a system set seems pretty unnecessary, at least for the current system complexity

mockersf (approved):
No preference from me between `Last` or its own schedule, both seems fine to me

MiniaczQ (approved):

mnmaita:
@MiniaczQ should this custom schedule run before `Last`? Wondering this since if it runs after, `Last` won't technically be the schedule that runs last anymore.

mnmaita:
I've rebased the branch and pushed changes to use a custom schedule with some assumptions. Please correct me if I did something wrong.

MiniaczQ:
lgtm

mnmaita:
@mockersf I've updated the migration guide (although there's really nothing to migrate per se, just a note) post merge since it had outdated information, is that enough or is this required to be changed somewhere else? I'm thinking about the website generated pages and such.

mockersf:
migration guide isn't really required for this as it's new, there's a release note https://github.com/bevyengine/bevy-website/pull/1773 but it doesn't mention anything about ordering

Issue #13464, opened 2024-05-21
High performance sorted queries


iiYese:
## What problem does this solve or what need does it fill?

#13417 adds sorting at an iterator level. These sorts:
- Do not persist across system runs. It will be common for sorted data to rarely be  
- Make query iteration completely random access

## What solution would you like?
1. Command to reorder tables.
2. Rearrange tables for sorted queries by default.
3. Make queries cache sorts & use change detection to early out of resorting.
4. Condense sorts to `low..=high` ranges for cache friendly iteration.
5. Allow caching iteration position to allow suspending and resuming iteration between system runs. This is useful for ui scroll areas and any rhythm game. 

Final API should look something like this:
```rs
fn cmp(left: FilteredEntityRef, right: FilteredEntityRef) -> std::cmp::Ordering {
    left.get::<A>().unwrap() < right.get::<A>().unwrap()
}

fn sys(mut query: Query<(&A, &B)>, mut pos: Local<usize>) {
    for (a, b) in query.sort(cmp)
        // Optional call that will prevent tables from being reorderd
        // Useful if you have multiple systems requesting slightly different sorts or
        // just completely opting out of reordering & doing multiple ordering manually.
        // User could do multiple stable sorts to get dense iteration across multiple systems.
        .no_reorder()
        .iter()
    {
        // ..
    }
    
    for (a, b) in query.sort(cmp)
        // resume iteration from a previous run & update it for a future run
        .seek(&mut *pos)
        .iter()
        .take(5)
    {
        // ..
    }
}
```

## What alternative(s) have you considered?

- Putting things in resources and doing these sorts myself. This is only viable when you've feature frozen your project and need no changes to data architecture. You completely lose the flexibility of ECS doing this.
- Despawning entities & reinserting them into tables. This first requires knowing all the components to take out of an entity and secondly relies on side effects. Table order is not part of the API contract.

## Additional context

Missing prerequisite features:
- `World::parallel_commands`/make the world command queue thread local by default (needed to allow queries to do `2.`).
- Archetype level change detection: #5097 (needed for `3.`).

Not required but helpful missing feature:
- Queries as entities (can avoid introducing the cache structure for queries that aren't sorted).

alice-i-cecile:
[...]

PR #16203, opened 2024-11-01
Share implementation of sort methods.

chescock:
# Objective

The various `Query::sort()` methods have a lot of duplicated code between them, including some unsafe code.  Reduce the duplication to make the code easier to read and maintain.  

## Solution

Extract the duplicated code to a private method, and pass in the sorting strategy as a closure.  

## Testing

I used `cargo-show-asm` to verify that the closures were inlined, but I didn't run anything through a profiler.  The `sort()` method itself even had identical assembly before and after this change, although the others did not.


crates/bevy_ecs/src/query/iter.rs
```diff
@@ -751,43 +641,9 @@ impl<'w, 's, D: QueryData, F: QueryFilter> QueryIter<'w, 's, D, F> {
...
-        // The original query iter has not been iterated on, so no items are aliased from it.
-        let query_lens = unsafe {
-            query_lens_state.iter_unchecked_manual(
-                world,
-                world.last_change_tick(),
-                world.change_tick(),
-            )
-        };
-        let mut keyed_query: Vec<_> = query_lens.collect();
-        keyed_query.sort_by(|(key_1, _), (key_2, _)| compare(key_1, key_2));
```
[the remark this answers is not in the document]
Victoronz: Good catch! This is a bug.

crates/bevy_ecs/src/query/iter.rs
```diff
@@ -1018,6 +804,33 @@ impl<'w, 's, D: QueryData, F: QueryFilter> QueryIter<'w, 's, D, F> {
...
+    fn sort_impl<L: ReadOnlyQueryData + 'w>(
+        self,
+        f: impl FnOnce(&mut Vec<(L::Item<'w>, NeutralOrd<Entity>)>),
+    ) -> QuerySortedIter<
+        'w,
+        's,
+        D,
+        F,
+        impl ExactSizeIterator<Item = Entity> + DoubleEndedIterator + FusedIterator + 'w,
+    > {
```
Victoronz, line 833: Do you see a way to preserve whether `Entity` needs to be wrapped in `NeutralOrd`?

Since this is no longer public, no need to restrict yourself to `impl Trait`, you can just use plain generics here.

Victoronz (commented):
It'd be nice if this could be done without introducing an additonal level of closure, maybe with generics?

Another reason for punting on this was that while these do the same now, that may change when sort caching is introduced.
This might have to look different then.

crates/bevy_ecs/src/query/iter.rs
chescock: It seems simpler to use it everywhere.  It's a simple newtype wrapper, so it won't affect performance.  And the methods that weren't wrapping the `Entity` don't actually use it during sorting, so it won't affect behavior, either.

chescock:
> It'd be nice if this could be done without introducing an additonal level of closure, maybe with generics?

What's wrong with using a closure here?  Closures are just syntactic sugar for generics, and I don't see how to make things more clear here with named types.  (There's no dynamic dispatch, if that's your worry; this is `impl FnOnce`, not `dyn FnOnce`.  It all gets inlined.)

chescock:
> Another reason for punting on this was that while these do the same now, that may change when sort caching is introduced.
> This might have to look different then.

Oh, I didn't know there were plans for sort caching!  This PR certainly isn't urgent, so it can wait if that's happening soon.  

Depending on your plans for caching, though, this change could make it easier.  I wouldn't expect the caching to vary wildly based on the type of sort, so doing this first could mean that you have to add it to one place instead of six.

Victoronz:
> > It'd be nice if this could be done without introducing an additonal level of closure, maybe with generics?
> 
> What's wrong with using a closure here? Closures are just syntactic sugar for generics, and I don't see how to make things more clear here with named types. (There's no dynamic dispatch, if that's your worry; this is `impl FnOnce`, not `dyn FnOnce`. It all gets inlined.)

While this might generate mildly different code, the main concern here instead would be capability to preserve the information that might be useful for caching: the sort kind, and the sort function used, if there is one (maybe as a fn pointer).

However, caching is not yet in the process of being implemented and is still being discussed in #13464, so how it'll look like is up in the air.

iiYese:
> Another reason for punting on this was that while these do the same now, that may change when sort caching is introduced.

It shouldn't. The public user facing APIs for the cached variants will only allow `fn` pointers because closures can be stateful & alter how they sort. The cache would have no way of detecting if this state is changed hence they're not allowed. All `impl Fn`s can accept function pointers.

Victoronz:
> > > It'd be nice if this could be done without introducing an additonal level of closure, maybe with generics?
> > 
> > 
> > What's wrong with using a closure here? Closures are just syntactic sugar for generics, and I don't see how to make things more clear here with named types. (There's no dynamic dispatch, if that's your worry; this is `impl FnOnce`, not `dyn FnOnce`. It all gets inlined.)
> 
> While this might generate mildly different code, the main concern here instead would be capability to preserve the information that might be useful for caching: the sort kind, and the sort function used, if there is one (maybe as a fn pointer).

Actually, the information I am talking about can and would better be passed as an additional parameter to `sort_impl` if need be, so no longer a concern.

crates/bevy_ecs/src/query/iter.rs
Victoronz: Fair, it should be okay then!

Victoronz:
Since #13443 merged first, this should do the same for the `QueryManyIter` sorts, after that it'd be 👍 
 You say that the assembly changed between some of them, could you show that difference?

chescock:
> Since #13443 merged first, this should do the same for the `QueryManyIter` sorts, after that it'd be 👍 

Done!  

> You say that the assembly changed between some of them, could you show that difference?

Not easily, sorry.  I used https://github.com/pacak/cargo-show-asm, but I had a temporary function that called the sort methods, and I didn't keep it anywhere.  I have the files with the output that I used for diffing, but they're 100 KB each and I don't remember which sort method they were calling.  I think a lot of the changes were just reordering blocks of instructions, but there was enough changed that I didn't try to track down exactly what was happening.  I just made sure there weren't any `call` instructions to functions that weren't called before.

Victoronz (approved):

JaySpruce (approved):

Bleachfuel (approved):

PR #16205, opened 2024-11-01
Assorted small small style refactors

sharma-shray:
# Objective

- Describe the objective or issue this PR addresses.
- If you're fixing a specific issue, say "Fixes #X".

## Solution

we can directly assign the result of the loop to name
text.into() can be changing the type conversion method to be more explicit.

## Testing

---

## Showcase

<details>

```rust
```

</details>

## Migration Guide

### DIFF
--- a/crates/bevy_ui/src/accessibility.rs
+++ b/crates/bevy_ui/src/accessibility.rs
@@ -27,13 +27,13 @@ fn calc_name(
     for child in children {
         let values = text_reader
             .iter(child)
-            .map(|(_, _, text, _, _)| text.into())
+            .map(|(_, _, text, _, _)| text.to_string()) 
             .collect::<Vec<String>>();
         if !values.is_empty() {
-            name = Some(values.join(" "));
+            name = Some(values.join(" ").into_boxed_str()); 
         }
     }
-    name.map(String::into_boxed_str)
+    name
 }
 
 fn calc_bounds(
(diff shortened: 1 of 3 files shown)


alice-i-cecile:
Can you justify why these changes are good? One sentence is fine.

crates/bevy_ui/src/focus.rs
```diff
@@ -95,8 +95,7 @@ impl RelativeCursorPosition {
     /// A helper function to check if the mouse is over the node
     pub fn mouse_over(&self) -> bool {
         self.normalized
-            .map(|position| self.normalized_visible_node_rect.contains(position))
-            .unwrap_or(false)
+            .map_or(false, |position| self.normalized_visible_node_rect.contains(position))
```
mockersf, line 98: there's some distaste amongst some contributors (me included) with using `map_or` as it's harder to read because it put the "or" case first

crates/bevy_ui/src/accessibility.rs
```diff
@@ -27,13 +27,13 @@ fn calc_name(
...
-            .map(|(_, _, text, _, _)| text.into())
+            .map(|(_, _, text, _, _)| text.to_string()) 
             .collect::<Vec<String>>();
         if !values.is_empty() {
-            name = Some(values.join(" "));
+            name = Some(values.join(" ").into_boxed_str()); 
         }
     }
-    name.map(String::into_boxed_str)
+    name
```
ickshonpe, line 36: Your changes seem sensible, but this function doesn't look right. Even if it's correct, it could iterate `children` backwards and short circuit return the first success rather than collecting and throwing away all these intermediate vecs?

doup:
Looks like this should be closed in favor of #16250 as it's a subset of the same changes.

PR #16206, opened 2024-11-01
Adding `ScheduleGraph::contains_set`

urben1680:
# Objective

The schedule graph can easily confirm whether a set is contained or not.

This helps me in my personal project where I write an extension trait for `Schedule` and I want to configure a specific set in its methods. The set in question has a run condition though and I don't want to add that condition to the same schedule as many times as the trait methods are called. Since the non-pub set is unknown to the schedule until then, a `contains_set` is sufficient.

It is probably trivial to add a method that returns an `Option<NodeId>` as well but as I personally don't need it I did not add that. If it is desired I can do so here though. It might be unneeded to have a `contains_set` then because one could check `is_some` on the returned id in that case.

## Solution

Added `ScheduleGraph::contains_set`.

## Testing

I put the below showcase code into a temporary unit test and it worked. If wanted I add it as a test too but I did not see that other more somewhat complicated methods have tests

---

## Showcase

```rs
#[derive(ScheduleLabel, Debug, Default, Clone, Copy, PartialEq, Eq, Hash)]
struct MySchedule;

#[derive(SystemSet, Debug, Default, Clone, Copy, PartialEq, Eq, Hash)]
struct MySet;

let mut schedule = Schedule::new(MySchedule);
assert_eq!(schedule.graph().contains_set(MySet), false);
schedule.configure_sets(MySet);
assert_eq!(schedule.graph().contains_set(MySet), true);
```

### DIFF
--- a/crates/bevy_ecs/src/schedule/schedule.rs
+++ b/crates/bevy_ecs/src/schedule/schedule.rs
@@ -650,6 +650,11 @@ impl ScheduleGraph {
             .and_then(|system| system.inner.as_deref())
     }
 
+    /// Returns `true` if the given system set is part of the graph. Otherwise returns `false`.
+    pub fn contains_set(&self, set: impl SystemSet) -> bool {
+        self.system_set_ids.contains_key(&set.intern())
+    }
+
     /// Returns the system at the given [`NodeId`].
     ///
     /// Panics if it doesn't exist.


urben1680:
For funsies, this is my current workaround:

```rs
fn forward_backward_sets_unknown(schedule: &mut Schedule) -> bool {
    let (lower_bound_before, upper_bound_before) = schedule.graph().system_sets().size_hint();
    schedule.configure_sets(ForwardSet);
    let (lower_bound_after, upper_bound_after) = schedule.graph().system_sets().size_hint();

    const EXPECT: &'static str = "ScheduleGraph::system_sets() expected to impl ExactSizeIterator";
    debug_assert_eq!(Some(lower_bound_before), upper_bound_before, "{EXPECT}");
    debug_assert_eq!(Some(lower_bound_after), upper_bound_after, "{EXPECT}");

    lower_bound_before < lower_bound_after
}
```

hymm (approved):

crates/bevy_ecs/src/schedule/schedule.rs
```diff
@@ -650,6 +650,11 @@ impl ScheduleGraph {
             .and_then(|system| system.inner.as_deref())
     }
 
+    /// Returns `true` if the given system set is part of the graph. Otherwise returns `false`.
```
BenjaminBrienen, line 653: *`Otherwise, `

BenjaminBrienen (approved):

BenjaminBrienen (approved):

PR #16207, opened 2024-11-01
bevy_image: PCX image support (off by default)

lupine:
# Objective

PCX is an obsolete raster image format from 1985. However, it was popular in a range of popular Windows games throughout the 1990s, and is in every respect a "proper" format.

I have a hobby game engine recreation project, which is currently written in Go ( https://code.ur.gs/lupine/ordoor ). I'm looking to rewrite it in rust, using Bevy. The first step is support for the various file formats this game needs.

Because I'm dealing with proprietary assets for a game that's still on sale, it's cleanest all round if I avoid processing or rewriting them in any way. So, I need bevy to understand what a pcx file is and load it (from a custom asset source, incidentally, although that doesn't matter too much for this PR).

## Solution

Although I could achieve this without changing bevy through a custom asset loader (and intend to follow that route for a whole set of other file formats), the image -> texture support seems to be handled automatically if bevy_image just supports the pcx format natively, so this PR adds support for that - off by default, with the `pcx` feature flag to turn it on.

A `pcx` crate already exists in the Rust ecosystem, and I recently got support wiring that up to the `image` crate merged: https://github.com/image-rs/image/pull/2364 . So all we're doing here is wiring the image-rs `pcx` code up to Bevy.

PCX images come in a range of colour formats; as an implementation detail, the upstream image-rs code takes paletted images and fakes them into 24-bit RGB, so they work fine as bevy textures.

Fair warning, there are some other "proper" image formats I've got my eye on - particularly ANI and CUR - if there's no objection to old formats in general!

## Testing

In progress. I just wanted to get this PR up early for comment. I assume we'd want to wait for image-rs to tag a new release before merging as well, but it would be good to get thoughts on whether bevy wants to support pcx early on.

I've had this compile and run end-to-end locally, leading to bevy loading a set of .pcx files into memory, but I've not made it display on screen yet. I'll definitely do that before taking this off draft.


lupine:
the `pcx` crate is WTFPL, which is a very permissive license that splits floss people slightly. FSF don't mind it, google do mind it :shrug:

alice-i-cecile:
@superdump @IceSentry @robtfm do any of you have opinions / objections here? I'm personally fine with this on an off-by-default basis, but I would prefer if we could somehow re-architect to make this easy to do externally.

PR #16209, opened 2024-11-02
Allow components to be "explicitly required".

### LINKED ISSUE #16194: Require components that can't be defaulted

Today, a required component either needs to impl `Default` or needs to have an explicit constructor function passed in. This is problematic for components which require some initial state. Consider the following setup:

```rust
#[derive(Component)]
#[require(EnemyTarget)]
struct Enemy;

#[derive(Component)]
struct EnemyTarget(Entity);
```

I want to say that an enemy needs to be inserted with a target, but a defaulted target doesn't really make sense and `Entity` doesn't even impl `Default`.

`require` should be able to not use a constructor at all and just require that the bundle inserts the component. Perhaps this could be done with a keyword like `#[require(EnemyTarget(explicit))]` to say that a component must be explicitly inserted and will fail (at compile time) otherwise.

This would severely decrease the remaining need for bundles.

One downside is that you can still break this invariant by explicitly removing one of the components after the fact, but that is already true.

Wait for archetype invariants. This would unfortunately mean problems would be detected at runtime.

andriyDev:
# Objective

- Fixes #16194.

## Solution

- Add an "explicit" tag to required components to allow them to check (at runtime) that a component is present (or inserted at the same time).

## Testing

- Unit tests are sufficient.

---

## Showcase

```rust
#[derive(Component)]
#[require(EnemyTarget(explicit))]
struct Enemy {
  health: f32,
}

#[derive(Component)]
struct EnemyTarget(Entity);

fn main() {
  App::new()
    .add_plugins(DefaultPlugins)
    .run()
}

  let target = commands.spawn(Transform::default()).id();
  // Panics if we don't add the EnemyTarget.
  commands.spawn((Enemy{ health: 10.0 }, EnemyTarget(target)));
}
```

## Migration Guide

- Required component constructors named `explicit` must be renamed to anything else.

## Risks

Required components so far are safe - they should not cause a panic. Explicitly required components can cause panics now whenever you spawn or insert a component that names an explicitly required component. For example, in the example above, if you spawn the Enemy + EnemyTarget components, then remove the EnemyTarget component, then insert the Enemy component again, you get a panic! This can make the panics more non-deterministic.

### DIFF
--- a/crates/bevy_ecs/macros/src/component.rs
+++ b/crates/bevy_ecs/macros/src/component.rs
@@ -92,11 +92,16 @@ pub fn derive_component(input: TokenStream) -> TokenStream {
             });
             match &require.func {
                 Some(RequireFunc::Path(func)) => {
+                    let constructor = if func.is_ident("explicit") {
+                        quote! { None }
+                    } else {
+                        quote! {Some(|| { let x: #ident = #func().into(); x })}
+                    };
                     register_required.push(quote! {
                         components.register_required_components_manual::<Self, #ident>(
                             storages,
                             required_components,
-                            || { let x: #ident = #func().into(); x },
+                            #constructor,
                             inheritance_depth
                         );
                     });
@@ -106,7 +111,7 @@ pub fn derive_component(input: TokenStream) -> TokenStream {
                         components.register_required_components_manual::<Self, #ident>(
                             storages,
                             required_components,
-                            || { let x: #ident = (#func)().into(); x },
+                            Some(|| { let x: #ident = (#func)().into(); x }),
                             inheritance_depth
                         );
                     });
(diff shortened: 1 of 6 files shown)


crates/bevy_ecs/macros/src/component.rs
```diff
@@ -92,11 +92,16 @@ pub fn derive_component(input: TokenStream) -> TokenStream {
             });
             match &require.func {
                 Some(RequireFunc::Path(func)) => {
+                    let constructor = if func.is_ident("explicit") {
+                        quote! { None }
+                    } else {
+                        quote! {Some(|| { let x: #ident = #func().into(); x })}
+                    };
```
MrGVSV, line 99: This could be surprising behavior to users who name their constructor `explicit` (for whatever reason). Could we make sure this is at least documented?

Alternatively we could do `#[requires(explicit = true)]` and peeking for `=` could be used to disambiguate the two. Not sure that's totally worth the ergonomics hit though.

mockersf:
would it be possible to find different words for require as in "automatically added if not present when adding this component" and require as in "must be present when adding this component"

andriyDev:
Closed as per the discussion in #16194.

PR #16211, opened 2024-11-02
Improved the global transform api to access rotation and scale

Sigma-dev:
# Objective

GlobalTransform's current methods make it unintuitive, long and clunky to access just the rotation or just the scale.

## Solution

Dedicated just_rotation() and scale() methods to access just these properties.

I'm not sure about the naming, I chose just_rotation() to show that try to indicate there is a waste since it also computes the other fields. 

## Testing

I tried logging the methods with a rotating and scaling cube and the values were correct.
My methods are based on existing bevy/glam methods so should be correct from the getgo.
Probably the easiest is using the 3d_rotations example, adding scaling to it and then logging the methods I added

---

## Showcase
```rust
fn log(gt_query: Query<&GlobalTransform>) {
        for global_transform in gt_query().iter() {
                   println!("{} {}", global_transform.just_rotation(), global_transform.scale());
        }
}
```

### DIFF
--- a/crates/bevy_transform/src/components/global_transform.rs
+++ b/crates/bevy_transform/src/components/global_transform.rs
@@ -218,7 +218,7 @@ impl GlobalTransform {
     /// It is calculated using `to_scale_rotation_translation`, meaning that you
     /// should probably use it directly if you also need translation or scale.
     #[inline]
-    pub fn just_rotation(&self) -> Quat {
+    pub fn rotation(&self) -> Quat {
         self.to_scale_rotation_translation().1
     }
 
--- a/crates/bevy_transform/src/components/global_transform.rs
+++ b/crates/bevy_transform/src/components/global_transform.rs
@@ -216,7 +216,7 @@ impl GlobalTransform {
     /// The transform is expected to be non-degenerate and without shearing, or the output will be invalid.
     /// 
     /// It is calculated using `to_scale_rotation_translation`, meaning that you
-    /// shoud probably use it directly if you also need translation or scale.
+    /// should probably use it directly if you also need translation or scale.
     #[inline]
     pub fn just_rotation(&self) -> Quat {
         self.to_scale_rotation_translation().1
--- a/crates/bevy_transform/src/components/global_transform.rs
+++ b/crates/bevy_transform/src/components/global_transform.rs
@@ -214,7 +214,7 @@ impl GlobalTransform {
     /// Get the rotation as a [`Quat`].
     ///
     /// The transform is expected to be non-degenerate and without shearing, or the output will be invalid.
-    /// 
+    ///
     /// It is calculated using `to_scale_rotation_translation`, meaning that you
     /// should probably use it directly if you also need translation or scale.
     #[inline]
--- a/crates/bevy_transform/src/components/global_transform.rs
+++ b/crates/bevy_transform/src/components/global_transform.rs
@@ -213,8 +213,10 @@ impl GlobalTransform {
 
     /// Get the rotation as a [`Quat`].
     ///
+    /// The transform is expected to be non-degenerate and without shearing, or the output will be invalid.
+    /// 
     /// It is calculated using `to_scale_rotation_translation`, meaning that you
-    /// shoud probably use it instead if you also need translation or scale.
+    /// shoud probably use it directly if you also need translation or scale.
     #[inline]
     pub fn just_rotation(&self) -> Quat {
         self.to_scale_rotation_translation().1
--- a/crates/bevy_transform/src/components/global_transform.rs
+++ b/crates/bevy_transform/src/components/global_transform.rs
@@ -229,6 +229,7 @@ impl GlobalTransform {
     /// The transform is expected to be non-degenerate and without shearing, or the output will be invalid.
     #[inline]
     pub fn scale(&self) -> Vec3 {
+        //Formula based on glam's implementation https://docs.rs/glam/latest/src/glam/f32/affine3a.rs.html#285-309
         let det = self.0.matrix3.determinant();
         Vec3::new(
             self.0.matrix3.x_axis.length() * det.signum(),
--- a/crates/bevy_transform/src/components/global_transform.rs
+++ b/crates/bevy_transform/src/components/global_transform.rs
@@ -227,6 +227,9 @@ impl GlobalTransform {
     /// Get the scale as a [`Vec3`].
     ///
     /// The transform is expected to be non-degenerate and without shearing, or the output will be invalid.
+    ///
+    /// Some of the computations overlap with `to_scale_rotation_translation`, which means you should use
+    /// it instead if you also need rotation.
     #[inline]
     pub fn scale(&self) -> Vec3 {
         //Formula based on glam's implementation https://github.com/bitshifter/glam-rs/blob/2e4443e70c709710dfb25958d866d29b11ed3e2b/src/f32/affine3a.rs#L290
--- a/crates/bevy_transform/src/components/global_transform.rs
+++ b/crates/bevy_transform/src/components/global_transform.rs
@@ -393,4 +393,18 @@ mod test {
             t1_prime.compute_transform(),
         );
     }
+
+    #[test]
+    fn scale() {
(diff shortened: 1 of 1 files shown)


crates/bevy_transform/src/components/global_transform.rs
```diff
@@ -211,6 +211,28 @@ impl GlobalTransform {
...
+    /// The transform is expected to be non-degenerate and without shearing, or the output will be invalid.
+    /// 
+    /// It is calculated using `to_scale_rotation_translation`, meaning that you
+    /// shoud probably use it directly if you also need translation or scale.
+    #[inline]
+    pub fn just_rotation(&self) -> Quat {
+        self.to_scale_rotation_translation().1
+    }
+
+    /// Get the scale as a [`Vec3`].
```
mweatherley, line 225: This also needs the caveats about being shear-free, otherwise this is not well-defined. The behavior in the presence of a negative determinant should maybe be documented as well.

crates/bevy_transform/src/components/global_transform.rs
```diff
@@ -211,6 +211,28 @@ impl GlobalTransform {
...
     }
 
+    /// Get the rotation as a [`Quat`].
+    ///
+    /// The transform is expected to be non-degenerate and without shearing, or the output will be invalid.
+    /// 
+    /// It is calculated using `to_scale_rotation_translation`, meaning that you
+    /// shoud probably use it directly if you also need translation or scale.
+    #[inline]
+    pub fn just_rotation(&self) -> Quat {
```
mweatherley, line 221: This name seems kind of clunky to me. Is there a reason this can't just be `rotation`?

mweatherley (changes requested):

crates/bevy_transform/src/components/global_transform.rs
Sigma-dev: I mentionned the reason in the PR message, but I'm not a 100% sure about it, so rotation() is fine with me too.

alice-i-cecile: Definitely prefer `rotation`.

crates/bevy_transform/src/components/global_transform.rs
```diff
@@ -211,6 +211,30 @@ impl GlobalTransform {
         self.0.translation
     }
 
+    /// Get the rotation as a [`Quat`].
+    ///
+    /// The transform is expected to be non-degenerate and without shearing, or the output will be invalid.
+    ///
+    /// It is calculated using `to_scale_rotation_translation`, meaning that you
```
alice-i-cecile, line 218:
```suggestion
    /// # Warning
    ///
    /// This is calculated using `to_scale_rotation_translation`, meaning that you
```

crates/bevy_transform/src/components/global_transform.rs
ickshonpe: Maybe `to_scale` and `to_rotation` so it's clear these aren't simple accessor functions like `translation`?

BenjaminBrienen (approved):

crates/bevy_transform/src/components/global_transform.rs
```diff
@@ -211,6 +211,32 @@ impl GlobalTransform {
...
+    }
+
+    /// Get the scale as a [`Vec3`].
+    ///
+    /// The transform is expected to be non-degenerate and without shearing, or the output will be invalid.
+    #[inline]
+    pub fn scale(&self) -> Vec3 {
+        let det = self.0.matrix3.determinant();
+        Vec3::new(
+            self.0.matrix3.x_axis.length() * det.signum(),
```
alice-i-cecile, line 234: Can you leave a comment here explaining why only the x value is multiplied by the sign of the determinant?

alice-i-cecile (commented):
I'd like to see a couple of tests for the `scale` method, ideally hitting all 8 octants of xyz space.

crates/bevy_transform/src/components/global_transform.rs
Sigma-dev: I mirrored it from the glam code, without knowing much of the math behind. 

A simpler implementation would be to just `to_scale_rotation_translation().0` but there would be a slight performance penalty. Whatever you think is best.

https://docs.rs/glam/latest/src/glam/f32/affine3a.rs.html#285-309

Sigma-dev:
> I'd like to see a couple of tests for the `scale` method, ideally hitting all 8 octants of xyz space.

I'm not very familiar with bevy's testing, could you tell me where these tests would exist in the repo ?

crates/bevy_transform/src/components/global_transform.rs
BenjaminBrienen: You could put that link as a normal comment there

Sigma-dev: Okay, did that

BenjaminBrienen:
> > I'd like to see a couple of tests for the `scale` method, ideally hitting all 8 octants of xyz space.
> 
> I'm not very familiar with bevy's testing, could you tell me where these tests would exist in the repo ?
> Or did you just mean that I should just test them once on my machine, so that we see it works.

In the file where `scale` is defined, in a module called `tests`, add a `#[test]` or two that provides code coverage.

crates/bevy_transform/src/components/global_transform.rs
```diff
@@ -211,6 +211,33 @@ impl GlobalTransform {
...
+    pub fn rotation(&self) -> Quat {
+        self.to_scale_rotation_translation().1
+    }
+
+    /// Get the scale as a [`Vec3`].
+    ///
+    /// The transform is expected to be non-degenerate and without shearing, or the output will be invalid.
+    #[inline]
+    pub fn scale(&self) -> Vec3 {
+        //Formula based on glam's implementation https://docs.rs/glam/latest/src/glam/f32/affine3a.rs.html#285-309
```
alice-i-cecile, line 232: This needs to be a permalink, so it doesn't point to the wrong thing as glam refactors!

mockersf (changes requested):
Why did you copy glam code for `scale`, but used the method for `rotation`?

Both should do the same thing, and I would prefer if it was just calling `self.to_scale_rotation_translation()` and keeping the correct part of the tuple

Jondolf:
> Why did you copy glam code for scale, but used the method for rotation?

It's unnecessary to extract everything for computing the scale. Extracting rotation requires first getting the scale, but extracting just the scale doesn't need any of the rest of the code in [`to_scale_rotation_translation`](https://docs.rs/glam/latest/src/glam/f32/affine3a.rs.html#285-309).

crates/bevy_transform/src/components/global_transform.rs
```diff
@@ -211,6 +211,33 @@ impl GlobalTransform {
...
+    /// This is calculated using `to_scale_rotation_translation`, meaning that you
+    /// should probably use it directly if you also need translation or scale.
+    #[inline]
+    pub fn rotation(&self) -> Quat {
+        self.to_scale_rotation_translation().1
+    }
+
+    /// Get the scale as a [`Vec3`].
+    ///
+    /// The transform is expected to be non-degenerate and without shearing, or the output will be invalid.
```
Jondolf, line 229: I'd like a mention here that if you also want rotation, you should use `to_scale_rotation_translation` instead.

crates/bevy_transform/src/components/global_transform.rs
Sigma-dev: I fixed it by replacing it with:
https://github.com/bitshifter/glam-rs/blob/2e4443e70c709710dfb25958d866d29b11ed3e2b/src/f32/affine3a.rs#L290

I'm not 100% sure that was what is expected of me.

crates/bevy_transform/src/components/global_transform.rs
Sigma-dev: I added something akin to that, idk if it's what you wanted

Sigma-dev:
I added tests and fixed the comments

alice-i-cecile (approved):

PR #16217, opened 2024-11-03
Add `scale_factor` fields to `RenderTarget` variants

aecsocket:
# Objective

Currently, if you set `RenderTarget` to anything other than a window, the render target is treated as having a hardcoded `1.0` scale factor. This means that `bevy_ui` will always render into that target with a 1x scale factor. My use-case is rendering into an `Image` which is later drawn onto a window via `bevy_egui` - but if the scale factor is hardcoded to 1x, UI will look too small if the window actually has a higher scale factor.

Current code:
```rs
pub fn get_render_target_info<'a>(
    &self,
    resolutions: impl IntoIterator<Item = (Entity, &'a Window)>,
    images: &Assets<Image>,
    manual_texture_views: &ManualTextureViews,
) -> Option<RenderTargetInfo> {
    match self {
        NormalizedRenderTarget::Window(window_ref) => resolutions
            .into_iter()
            .find(|(entity, _)| *entity == window_ref.entity())
            .map(|(_, window)| RenderTargetInfo {
                physical_size: window.physical_size(),
                scale_factor: window.resolution.scale_factor(),
            }),
        NormalizedRenderTarget::Image(image_handle) => {
            let image = images.get(image_handle)?;
            Some(RenderTargetInfo {
                physical_size: image.size(),
                scale_factor: 1.0, // <-- hardcoded here!
            })
        }
        NormalizedRenderTarget::TextureView(id) => {
            manual_texture_views.get(id).map(|tex| RenderTargetInfo {
                physical_size: tex.size,
                scale_factor: 1.0, // <-- hardcoded here!
            })
        }
    }
}
```

## Solution

This PR adds `scale_factor` fields to `RenderTarget::Image` and `RenderTarget::TextureView`, which is fed directly into `RenderTargetInfo` (which is used by `bevy_ui`). Old uses of `Image` and `TextureView` constructors are replaced with `RenderTarget::from(image)` or `from(texture_view_handle)`, which keep the old behaviour of having scale factor `1.0`.

## Testing

In `render_ui_to_texture` example, if you press the up or down arrow keys, it will change the scale factor of the `RenderTarget` by 0.25. Tested this example and it works.

---

## Showcase

`RenderTarget`s which are not `Window`s can now have a custom scale factor applied. Previously, if you had UI which rendered into a camera which rendered into a `RenderTarget`, the UI scale factor would be fixed to `1.0`. Now you can explicitly specify the scale factor of the render target. This is useful for if you want to, for example, render the game itself (including UI) into an image, then draw that image onto the final window manually.

You can try out these changes in the updated `render_ui_to_texture` example.

[Screencast From 2024-11-03 15-15-34.webm](https://github.com/user-attachments/assets/ef9e9e62-a768-47ee-b8e1-78c5b30a5a28)

## Migration Guide

`RenderTarget::Image` and `RenderTarget::TextureView` have been turned into field enum variants, and a new field `scale_factor` has been added, letting you change the scale factor of the render target (previously, it was hardcoded to `1.0` unless it was a window).

If you want to keep the old behavior, either set `scale_factor` to 1.0 explicitly, or use `RenderTarget::from(Handle<Image>)` or `from(ManualTextureViewHandle)`, which also creates a target with a scale factor of 1.0

### DIFF
--- a/crates/bevy_render/src/camera/camera.rs
+++ b/crates/bevy_render/src/camera/camera.rs
@@ -41,6 +41,8 @@ use bevy_window::{
 };
 use core::ops::Range;
 use derive_more::derive::From;
+use std::cmp::Ordering;
+use std::hash::{Hash, Hasher};
 use wgpu::{BlendState, TextureFormat, TextureUsages};
 
 /// Render viewport configuration for the [`Camera`] component.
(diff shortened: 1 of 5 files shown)


alice-i-cecile:
This fix is going to be needed by the Bevy Editor: we're doing all sorts of weird viewport things.

aecsocket:
I was just about to update this to latest main and saw that https://github.com/bevyengine/bevy/pull/16796 is already merged - which implements variable scale factor for `Image`s but not `TextureView`s. Does this warrant a separate PR (and closing this PR)?

alice-i-cecile:
> I was just about to update this to latest main and saw that https://github.com/bevyengine/bevy/pull/16796 is already merged - which implements variable scale factor for Images but not TextureViews. Does this warrant a separate PR (and closing this PR)?

Yeah, redoing this work in that context feels like the right choice :)

ChristopherBiscardi:
closing since redoing in the current context seems to be the preferred approach moving forward, as per the above comments.

PR #16218, opened 2024-11-03
Fix fallible param notes

MiniaczQ:
I noticed one of the reflinks doesn't work correctly


alice-i-cecile (approved):

Danielmelody:
This is actually a breaking change from 0.15.0, as some system with `Single<..>` query crashes after a patch update to 0.15.1. Better to intro this in 0.16.0,

PR #16219, opened 2024-11-03
Flush commands after every mutation in `WorldEntityMut`

### LINKED ISSUE #14621: `Commands` apply at unexpected times with exclusive `World` access
## Bevy version

`0.14.0` and `main` (c1c003d3c7c0d12dcd1af09fb57a246e1b401a10)

## What you did

`world.commands().add(...)` from within another command. Take the following example:

```rust
let mut world = World::new();

let mut cmd = world.commands();

// Command 1
cmd.add(|world: &mut World| {
    let entity_1 = world.spawn_empty().id();

    println!("A");

    // Command 3
    world.commands().add(|world: &mut World| {
        println!("BUZZ!");
    });

    println!("B");

    world.entity_mut(entity_1).insert(A);

    let a = world.get::<A>(entity_1);

    println!("C");

    world.spawn(A);

    println!("D");

    world.flush();

    println!("E");
});

// Command 2
cmd.add(|_: &mut World| {
    println!("F");
});

world.flush();
```

When does `BUZZ!` get printed?

## What went wrong

It ends up being in the place I would *least* expect, which is why I'm calling this a bug rather than simply lacking documentation.

Loosely ordered by where I would expect it to occur:

* After F? If the `&mut World` passed to each command is the same world as the outermost, then I would expect `world.commands()` to always refer to the same queue, so Command 1 is added, Command 2 is added, Command 1 runs, it pushes Command 3 into the queue, Command 2 runs, Command 3 runs.
* After D or E? If each command gets a new "nested" queue that's flushed after it finishes, then it would happen between Command 1 and Command 2. But we're also calling `world.flush()`, and the docs for `world.commands()` says they'll run after a call to `World::flush`. So it should be after D.
* After A? We have exclusive `World` access, so `Commands` are a bit redundant. If we ignore the bit about `World::flush` in the `World::commands` docs, then maybe we'll guess that they simply execute immediately, and `World::commands` is just for convenience when you have something that accepts a `Commands` argument. Or maybe they flush when the `Commands` is immediately dropped after our `add`?
* After B? Maybe there's an implicit `World::flush` any time we actually mutate the world or when we read the result of a mutation.
* After C? If it's none of the above, this is the only option left, apart from "never".

And the winner is... after C! It comes down to `World::flush` being implicit in some methods, but not others. There isn't any mention of this in any of their docs. Even if it was called out explicitly in the docs for each method if it does a flush, I wouldn't want to have to keep that context in my head when reasoning about when a command will run. 

To me, the only three possibilities for when a command should run given exclusive world access are:
1. At the end of the outermost command queue
2. Immediately after the parent command
3. Immediately after being added, or at least before any other world access. No explicit flushing or "will they/won't they" implicit flushing.

## Additional information

This comes out of my attempts to get lifecycle `Observer`s to behave the way I want, and they make things even harder to reason about. Specifically, the `OnAdd` trigger will occur immediately upon component addition, but if you add commands to the queue in the trigger, they'll be executed whenever the command that did the component addition does a flush, which can be wildly unpredictable. For example, what if the command pre-spawns a bunch of entities, then adds components? Your observer's commands will run at the end of that command after *all* components have been added. What if instead, it interleaves entity spawning and component adding? Then you'll get a flush before each new entity, and your observer's commands will occur midway through the command that triggered it.

### LINKED ISSUE #16034: Commands queued by component hooks are not applied during exclusive world access
# Problem

When an entity is spawned (or other structural changes occur) as part of exclusive world access, commands to apply changes via hooks are generated, but are not immediately applied.

In order to apply these commands, we must call `world.flush`, as demonstrated [here](https://github.com/Leafwing-Studios/i-cant-believe-its-not-bsn/blob/8e5783dd164f99960f6c33cc3f83b99174c84ce2/src/maybe.rs#L150).

This allows users to accidentally observe the state of the world *before* component lifecycle hooks are applied, but *only* when working with exclusive world access. As a result, we must be extremely careful if we are relying on hooks for soundness.

## Bevy version

0.14.1

## Additional context

This is related to #14621, but is intended as a clearer demonstration of a specific problem.

#12235 may end up being blocked on this, as otherwise we could get UB by calling unsafe code on an invalid entity hierarchy.

### LINKED ISSUE #16212: `observe` on `WorldEntityMut` may cause the reference to be invalidated
## Bevy version

0.15.0-rc.2 (present in 0.14.0) as well

## What you did

```rust
    #[derive(Event)]
    struct Boom;

    #[derive(Event)]
    struct BoomTomorrow;

    fn boom(trigger: Trigger<Boom>, mut commands: Commands) {
        println!("boom, despawning entity: {:?}", trigger.entity());
        commands.entity(trigger.entity()).despawn();
    }

    fn boom_tomorrow(trigger: Trigger<Boom>, mut commands: Commands) {
        // boom tomorrow
    }

    #[test]
    fn test_boom() {
        let mut app = App::new();

        let mut id = app.world_mut().spawn(Name::new("a")).observe(boom).id();
        let mut id2 = app.world_mut().spawn(Name::new("b")).id();
        
        app.world_mut().flush();
        println!("flushed");
        
        let mut a = app.world_mut().entity_mut(id);
        a.trigger(Boom);
        a.observe(boom_tomorrow);
        let name = a.get::<Name>().unwrap();
        assert_eq!(name.as_str(), "b");
    }
```

Result

```
test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 2 filtered out; finished in 0.00s
```

## What went wrong
 
- `WorldEntityMut` depends on the fact that the `EntityLocation` it caches does not change while it is alive, or changes when it itself expects it to.
- This means that there must be no way to flush the command queue while `WorldEntityMut` lives, as this might cause queued despawns to be run. 
- `observe` internally calls `world.spawn()`.
- `world.spawn()` implicitly has a `flush()` in it, as evidenced in bug report #14621.
- Hence, entity gets despawned while `WorldEntityMut` exists, which means any accesses will use outdated `EntityLocation`

## Additional information

Also related as context #16034.

nakedible:
# Objective

- Currently adding observers spawns an entity which implicitly flushes the command queue, which can cause undefined behaviour if the `WorldEntityMut` is used after this
- The reason `WorldEntityMut` attempted to (unsuccessfully) avoid flushing commands until finished was that such commands may move or despawn the entity being referenced, invalidating the cached location.
- With the introduction of hooks and observers, this isn't sensible anymore as running the commands generated by hooks immediately is required to maintain correct ordering of operations and to not expose the world in an inconsistent state
- Objective is to make command flushing deterministic and fix the related issues
- Fixes #16212
- Fixes #14621 
- Fixes #16034

## Solution

- Allow `WorldEntityMut` to exist even when it refers to a despawned entity by allowing `EntityLocation` to be marked invalid
- Add checks to all methods to panic if trying to access a despawned entity
- Flush command queue after every operation that might trigger hooks or observers
- Update entity location always after flushing command queue

## Testing

- Added test cases for currently broken behaviour
- Added test cases that flushes happen in all operations
- Added test cases to ensure hooks and commands are run exactly in correct order when nested

---

Todo:

- [x] Write migration guide
- [x] Add tests that using `EntityWorldMut` on a despawned entity panics
- [x] Add tests that commands are flushed after every operation that is supposed to flush them
- [x] Add tests that hooks, observers and their spawned commands are run in the correct order when nested

---

## Migration Guide

Previously `EntityWorldMut` triggered command queue flushes in unpredictable places, which could interfere with hooks and observers. Now the command queue is flushed always immediately after any call in `EntityWorldMut` that spawns or despawns an entity, or adds, removes or replaces a component. This means hooks and observers will run their commands in the correct order.

As a side effect, there is a possibility that a hook or observer could despawn the entity that is being referred to by `EntityWorldMut`. This could already currently happen if an observer was added while keeping an `EntityWorldMut` referece and would cause unsound behaviour. If the entity has been despawned, calling any methods which require the entity location will panic. This matches the behaviour that `Commands` will panic if called on an already despawned entity. In the extremely rare case where taking a new `EntityWorldMut` reference or otherwise restructuring the code so that this case does not happen is not possible, there's a new `is_despawned` method that can be used to check if the referred entity has been despawned.

### DIFF
--- a/crates/bevy_ecs/src/world/entity_ref.rs
+++ b/crates/bevy_ecs/src/world/entity_ref.rs
@@ -860,7 +860,23 @@ pub struct EntityWorldMut<'w> {
 }
 
 impl<'w> EntityWorldMut<'w> {
+    #[track_caller]
+    #[inline(never)]
+    #[cold]
+    fn panic_despawned(&self) -> ! {
+        panic!("Entity {} has been despawned, possibly by hooks or observers, so must not be accessed through EntityWorldMut after despawn.", self.entity);
+    }
+
+    #[inline(always)]
+    #[track_caller]
+    fn assert_not_despawned(&self) {
+        if self.location.archetype_id == ArchetypeId::INVALID {
+            self.panic_despawned();
+        }
+    }
+
     fn as_unsafe_entity_cell_readonly(&self) -> UnsafeEntityCell<'_> {
+        self.assert_not_despawned();
         UnsafeEntityCell::new(
             self.world.as_unsafe_world_cell_readonly(),
             self.entity,
@@ -868,6 +884,7 @@ impl<'w> EntityWorldMut<'w> {
         )
     }
     fn as_unsafe_entity_cell(&mut self) -> UnsafeEntityCell<'_> {
+        self.assert_not_despawned();
         UnsafeEntityCell::new(
             self.world.as_unsafe_world_cell(),
             self.entity,
@@ -875,6 +892,7 @@ impl<'w> EntityWorldMut<'w> {
         )
     }
     fn into_unsafe_entity_cell(self) -> UnsafeEntityCell<'w> {
+        self.assert_not_despawned();
         UnsafeEntityCell::new(
             self.world.as_unsafe_world_cell(),
             self.entity,
@@ -914,12 +932,14 @@ impl<'w> EntityWorldMut<'w> {
     /// Gets metadata indicating the location where the current entity is stored.
     #[inline]
     pub fn location(&self) -> EntityLocation {
+        self.assert_not_despawned();
         self.location
     }
 
     /// Returns the archetype that the current entity belongs to.
     #[inline]
     pub fn archetype(&self) -> &Archetype {
+        self.assert_not_despawned();
         &self.world.archetypes[self.location.archetype_id]
     }
 
@@ -1208,6 +1228,7 @@ impl<'w> EntityWorldMut<'w> {
         mode: InsertMode,
         #[cfg(feature = "track_change_detection")] caller: &'static core::panic::Location,
     ) -> &mut Self {
+        self.assert_not_despawned();
         let change_tick = self.world.change_tick();
         let mut bundle_inserter =
             BundleInserter::new::<T>(self.world, self.location.archetype_id, change_tick);
@@ -1237,6 +1258,7 @@ impl<'w> EntityWorldMut<'w> {
         component_id: ComponentId,
         component: OwningPtr<'_>,
     ) -> &mut Self {
+        self.assert_not_despawned();
         let change_tick = self.world.change_tick();
         let bundle_id = self
             .world
@@ -1281,6 +1303,7 @@ impl<'w> EntityWorldMut<'w> {
         component_ids: &[ComponentId],
         iter_components: I,
     ) -> &mut Self {
+        self.assert_not_despawned();
         let change_tick = self.world.change_tick();
         let bundle_id = self
             .world
@@ -1315,6 +1338,7 @@ impl<'w> EntityWorldMut<'w> {
     // TODO: BundleRemover?
     #[must_use]
     pub fn take<T: Bundle>(&mut self) -> Option<T> {
+        self.assert_not_despawned();
         let world = &mut self.world;
         let storages = &mut world.storages;
         let components = &mut world.components;
@@ -1584,6 +1608,7 @@ impl<'w> EntityWorldMut<'w> {
     /// See [`EntityCommands::remove`](crate::system::EntityCommands::remove) for more details.
     // TODO: BundleRemover?
     pub fn remove<T: Bundle>(&mut self) -> &mut Self {
+        self.assert_not_despawned();
         let storages = &mut self.world.storages;
         let components = &mut self.world.components;
         let bundle_info = self.world.bundles.register_info::<T>(components, storages);
@@ -1597,6 +1622,7 @@ impl<'w> EntityWorldMut<'w> {
 
     /// Removes all components in the [`Bundle`] and remove all required components for each component in the bundle
     pub fn remove_with_requires<T: Bundle>(&mut self) -> &mut Self {
+        self.assert_not_despawned();
         let storages = &mut self.world.storages;
         let components = &mut self.world.components;
         let bundles = &mut self.world.bundles;
@@ -1614,6 +1640,7 @@ impl<'w> EntityWorldMut<'w> {
     ///
     /// See [`EntityCommands::retain`](crate::system::EntityCommands::retain) for more details.
     pub fn retain<T: Bundle>(&mut self) -> &mut Self {
+        self.assert_not_despawned();
         let archetypes = &mut self.world.archetypes;
         let storages = &mut self.world.storages;
         let components = &mut self.world.components;
@@ -1646,6 +1673,7 @@ impl<'w> EntityWorldMut<'w> {
     ///
     /// Panics if the provided [`ComponentId`] does not exist in the [`World`].
     pub fn remove_by_id(&mut self, component_id: ComponentId) -> &mut Self {
+        self.assert_not_despawned();
         let components = &mut self.world.components;
 
         let bundle_id = self
@@ -1662,6 +1690,7 @@ impl<'w> EntityWorldMut<'w> {
 
     /// Removes all components associated with the entity.
     pub fn clear(&mut self) -> &mut Self {
+        self.assert_not_despawned();
         let component_ids: Vec<ComponentId> = self.archetype().components().collect();
         let components = &mut self.world.components;
 
@@ -1681,6 +1710,7 @@ impl<'w> EntityWorldMut<'w> {
     ///
     /// See [`World::despawn`] for more details.
     pub fn despawn(self) {
+        self.assert_not_despawned();
         let world = self.world;
         let archetype = &world.archetypes[self.location.archetype_id];
 
--- a/crates/bevy_ecs/src/world/entity_ref.rs
+++ b/crates/bevy_ecs/src/world/entity_ref.rs
@@ -4482,4 +4482,27 @@ mod tests {
 
         assert_eq!((&mut X(8), &mut Y(9)), (x_component, y_component));
     }
+
+    #[derive(Event)]
+    struct TestEvent;
+
+    #[test]
+    fn adding_observer_updates_location() {
+        let mut world = World::new();
+        let entity = world
+            .spawn_empty()
+            .observe(|trigger: Trigger<TestEvent>, mut commands: Commands| {
+                commands.entity(trigger.entity()).insert(TestComponent(0));
+            })
+            .id();
+
+        // this should not be needed, but is currently required to tease out the bug
+        world.flush();
+
+        let mut a = world.entity_mut(entity);
+        a.trigger(TestEvent); // this adds command to change entity archetype
+        a.observe(|_: Trigger<TestEvent>| {}); // this flushes commands implicitly by spawning
+        let location = a.location();
+        assert_eq!(world.entities().get(entity), Some(location));
+    }
 }
--- a/crates/bevy_ecs/src/world/entity_ref.rs
+++ b/crates/bevy_ecs/src/world/entity_ref.rs
@@ -4699,6 +4699,52 @@ mod tests {
     }
 
     #[derive(Resource)]
+    struct TestFlush(usize);
+
+    fn count_flush(world: &mut World) {
+        world.resource_mut::<TestFlush>().0 += 1;
+    }
+
+    #[test]
+    fn archetype_modifications_trigger_flush() {
+        let mut world = World::new();
+        world.insert_resource(TestFlush(0));
+        world.add_observer(|_: Trigger<OnAdd, TestComponent>, mut commands: Commands| {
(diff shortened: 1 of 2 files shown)


alice-i-cecile:
Yes please, I'd like a small note here.

maniwani (commented):
Yes, this is pretty much what I was thinking. I assume branch prediction should hide any overhead from this since the condition should basically never be true.

I think hitting this case should issue a warning. And we might want to have a `#[track_caller]` on the affected `EntityWorldMut` methods so that anyone who sees the warning knows where to start looking.

Something like...
```rust
#[inline(never)]
#[cold]
fn on_entity_invalid(entity: Entity, caller: &'static Location) {
    warn!("error[B<insert-error-number-here>]: {caller} tried to modify the entity {:?}, but it no longer exists. See: https://bevyengine.org/learn/errors/b<insert-error-number-here>", entity);
}
```

maniwani:
Ah, wait. Since `EntityWorldMut` can be converted into the other kinds of entity references, we may need to add the invalid location check to them as well (or make the conversion fallible).

Some of the `UnsafeEntityCell` methods those references use probably assume its location is valid.

Maximetinu:
Hello! 👋 

I lack context about the changes in this PR, but by [petition](https://discord.com/channels/691052431525675048/742569353878437978/1303418282065526875) of @alice-i-cecile , I checked out this branch to see if I can still reproduce the crash of `flush() needs to be called before this operation is legal` that I was hitting in my code

The answer is that no, I cannot.

The thing is, I can't repro it in bevy main neither, so this bug must have been fixed at some point between Bevy 0.14.2 (the version where I encountered) and current Bevy main atow

Here, for reference:

----

this is my minimal reproduceable example, on top of Bevy 0.14.2: 

https://github.com/Maximetinu/bevy/commits/minimal-reproduceable-bug-on-remove-hook/

see changes in latest commit

running that modified example does reproduce the bug

notice that the cause is just spawning an empty entity during a on_remove hook

----

This is my minimal reproduceable example cherry-picked to this branch: 

https://github.com/Maximetinu/bevy/commits/world-entity-mut-flushes-checking-if-panic-keeps-happening/

see changes in latest commit, they are the same as in the branch above

running that modified example does **not** reproduce the bug

so, this branch works 🥳 

----

However, bevy main also works:

https://github.com/Maximetinu/bevy/commits/reproducing-on-remove-hook-panic-on-bevy-main/

which means that, well, this branch does not fix per se this bug, because it is already fixing on Bevy main

but at least it means that it's not re-introducing it neither 🙂 

----

Maybe [my minimal reproduceable example ](https://github.com/Maximetinu/bevy/blob/world-entity-mut-flushes-checking-if-panic-keeps-happening/examples/ecs/component_hooks.rs)can help to write a test covering from this specific scenario

I hope this helps!

alice-i-cecile:
I'd love to see that turned into a test, either in this PR or separately :)

Maximetinu:
> I'd love to see that turned into a test, either in this PR or separately :)

I have made a test for this, that fails at 0.14.2 but passes in current bevy main, so it should cover from this regression

```rs
#[test]
fn test_spawn_entities_during_on_remove_hook() {
    #[derive(Component)]
    struct MyComponent;

    App::new()
        .add_plugins(MinimalPlugins)
        .add_systems(Startup, |world: &mut World| {
            world.register_component_hooks::<MyComponent>().on_remove(
                |mut world, _entity, _component_id| {
                    world.commands().spawn_empty();
                },
            );
        })
        .add_systems(
            Update,
            |mut commands: Commands,
                my_component_q: Query<Entity, With<MyComponent>>,
                mut exit_event: EventWriter<AppExit>,
                mut after_first_frame: Local<bool>| {
                for entity in &my_component_q {
                }

                commands.spawn(MyComponent);

                if *after_first_frame {
                    exit_event.send(AppExit::Success);
                }

                *after_first_frame = true;
            },
        )
        .run();
}
```

It's a bit of an unusual test (I'm not even sure if `MinimalPlugins` can be used in a test like this). It's more an integration test rather than a unit test. I don't know if there is some place inside Bevy repo where more of these tests exist. If so, I can open a PR to add it.

alice-i-cecile:
Let's spin this out into a new issue :)

nakedible:
@maniwani Updated design (just quick notes for now):
- Handle also `UnsafeEntityCell`
- Instead of setting location entirely to invalid, instead set both `archetype_id` and `table_id` to "empty", but keep `archetype_row` and `table_row` as invalid. That allows us to behave more like an empty entity for most APIs.

nakedible:
I changed this pull to instead panic rather than warn and try to behave if nothing is wrong. The reason for this is that if we only try to warn, there's a quite wide ranging knock-off effect into `UnsafeEntityCell` and a lot of other places which store `EntityLocation`. If we instead panic, the effects (and the extra runtime check) are limited to just `WorldEntityMut`. Also, Bevy already opts to panicing if using commands to modify an entity that has been despawned (https://bevyengine.org/learn/errors/b0003/) so this is similar behaviour. There should be nobody depending on this behaviour currently as calling any of the methods after the entity has been despawned by hooks would have already been unsound in the past – so if it happens, then the panics will reveal places where there has been already been a bug.

nakedible:
@alice-i-cecile @maniwani I'd love to get feedback if the new panicing approach is good or bad. Also, is it reasonable to add a panics section to all the methods which might panic for this. It should truly be a corner case when this happens, so I'm not sure if it's sensible to spam it everywhere (and it cannot be put into some of the non-fallible `From` conversions).

nakedible:
Implementation of panic is still placeholderish, just to note.

alice-i-cecile:
> Also, Bevy already opts to panicing if using commands to modify an entity that has been despawned

For what it's worth, this behavior is widely hated :)

I think I'm broadly on board with panicking here though, especially as a first pass. This really is an edge case, and I think that trying to figure out better ways to prevent this bad state from ever arising is more useful than bubbling the error up.

nakedible:
If #16499 gets merged before this one, one of the test cases will break, as it should when the ordering of hooks and observers is changed.

nakedible:
@alice-i-cecile -S-Waiting-on-Author +S-Needs-Review 🙇

alice-i-cecile (approved):
Excellent tests and good analysis. I'd like to clean up the (very niche) panics here by swapping to results, but that's best left for follow-up.

crates/bevy_ecs/src/world/entity_ref.rs
```diff
@@ -860,21 +860,39 @@ pub struct EntityWorldMut<'w> {
...
+    #[inline(never)]
+    #[cold]
+    fn panic_despawned(&self) -> ! {
+        panic!("Entity {} has been despawned, possibly by hooks or observers, so must not be accessed through EntityWorldMut after despawn.", self.entity);
+    }
+
+    #[inline(always)]
+    #[track_caller]
+    fn assert_not_despawned(&self) {
+        if self.location.archetype_id == ArchetypeId::INVALID {
```
chescock, line 873: This could use the `is_despawned()` method if you wanted.  

```suggestion
        if self.is_despawned() {
```

chescock (approved):
I don't entirely understand what `world.flush()` does, but I do believe this PR calls `assert_not_despawned();` before every read or write of the entity or its components and calls `world.flush()` and `update_location()` after every write, so `self.location` should always be correct.  

The one exception is that the `world_scope()` method calls `update_location()` but not `world.flush()`.  Should that call `world.flush()` as well?

hymm:
An alternative to checking in every method if the entity has been despawned would be to panic after calling world.flush instead of checking in each method. This would disallow despawning the entity in hooks, but I think that makes sense that you wouldn't be able to despawn the EntityWorldMut while holding it.

nakedible:
> I don't entirely understand what `world.flush()` does, but I do believe this PR calls `assert_not_despawned();` before every read or write of the entity or its components and calls `world.flush()` and `update_location()` after every write, so `self.location` should always be correct.  
> 
> The one exception is that the `world_scope()` method calls `update_location()` but not `world.flush()`.  Should that call `world.flush()` as well?  

World flush runs any `Command`s that have been queued. Commands are queued by hooks and observers if they want to do archetype changes, as they only get a `DeferredWorld`.

The `world_scope` function just allows a function to run with the world, but it doesn't necessarily queue any commands. If it does, it's the responsibility of the function to flush if necessary.

There has been talk of having commands apply immediately when holding a `&mut World`, like in this `world_scope` function, which would solve the whole issue of asking the user to flush.

nakedible:
> An alternative to checking in every method if the entity has been despawned would be to panic after calling world.flush instead of checking in each method. This would disallow despawning the entity in hooks, but I think that makes sense that you wouldn't be able to despawn the EntityWorldMut while holding it.

It would be an alternative, but instead the aim is to go to another direction, to make the methods fallible instead.

Also, there are two other reasons.

First, there aren't any other ways to make archetype changes to entities but to take `WorldEntityMut` via `world.entity_mut(ent)`. So if despawning the current entity would be totally disallowed, then there would not be a way to safely trigger such a change as you can't not hold a reference while doing it.

Second, you can't really prevent hooks and observers from making these types of changes in some special cases. So there isn't a workaround for that. Where as just dropping a `WorldEntityMut` after an archetype mutation operation is a trivial thing to do by the user.

NthTensor:
FWIW i think panicking is fine here; there's similar stuff in the engine that already works this way. We can reconsider it in a broader review of bevy's error handling. But especially for guarding invariant going into async code, panicking is probably the way to go.

PR #16220, opened 2024-11-03
Support creating asset directories

richchurcher:
# Objective

Exposes a means to create an asset directory (and its parent directories). Wasn't sure whether we also wanted the variant to create directories without the parent (i.e. `mkdir` instead of `mkdir -p`)?

Fixes https://github.com/bevyengine/bevy_editor_prototypes/issues/144


richchurcher:
Hrm, I'll check out the CI in a sec. Maybe there's a good reason we didn't already have this :laughing:

mockersf (approved):

alice-i-cecile (approved):

mnmaita (approved):

PR #16222, opened 2024-11-04
Gamepad improvements

### LINKED ISSUE #16221: Allow mocking inputs for `Gamepad`

After #15685 it became inconvenient to mock inputs. On 0.14 we could call `set_pressed` or `set`. But these methods are no longer exposed.

It's still possible to do this by sending events, but it's inconsistent with the rest of the input and not very ergonomic.

I would like to bring back `set_pressed` and `set`.

## Additional context

There is a similar issue regardless gamepad creation that predates the gamepads rework: #4911. Could be solved by adding a constructor for `Gamepad` component.

Shatur:
# Objective

Closes #16221.

## Solution

- Make `Gamepad` fields public and remove delegates / getters.
- Move `impl Into` to `Axis` methods (delegates for `Axis` used `impl Into` to allow passing both `GamepadAxis` and `GamepadButton`).
- Improve docs.

## Testing

- I run tests.

Not sure if the migration guide is needed, since it's a feature from RC, but I wrote it just in case.

---

## Migration Guide

- `Gamepad` fields are now public.
- Instead of using `Gamepad` delegates like `Gamepad::just_pressed`, call these methods directly on the fields.

### DIFF
--- a/crates/bevy_input/src/gamepad.rs
+++ b/crates/bevy_input/src/gamepad.rs
@@ -306,7 +306,7 @@ pub enum ButtonSettingsError {
     },
 }
 
-/// Stores a connected gamepad's metadata such as name and its input.
+/// Stores a connected gamepad's state and any metadata such as the device name.
 ///
 /// An entity with this component is spawned automatically after [`GamepadConnectionEvent`]
 /// and updated by [`gamepad_event_processing_system`].
--- a/crates/bevy_input/src/gamepad.rs
+++ b/crates/bevy_input/src/gamepad.rs
@@ -363,6 +363,11 @@ impl Gamepad {
         }
     }
 
+    /// Creates a gamepad with the given name.
+    pub fn with_name(name: impl Into<String>) -> Self {
+        Self::new(GamepadInfo::new(name))
+    }
+
     /// Returns the left stick as a [`Vec2`]
     pub fn left_stick(&self) -> Vec2 {
         Vec2 {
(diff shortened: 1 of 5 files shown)


crates/bevy_input/src/gamepad.rs
```diff
@@ -306,13 +306,12 @@ pub enum ButtonSettingsError {
     },
 }
 
-/// The [`Gamepad`] [`component`](Component) stores a connected gamepad's metadata such as the `name` and its [`GamepadButton`] and [`GamepadAxis`].
+/// Stores a connected gamepad's metadata such as name and its input.
```
alice-i-cecile, line 309:
```suggestion
/// Stores a connected gamepad's state and any metadata such as the device name.
```

crates/bevy_input/src/gamepad.rs
```diff
@@ -547,6 +423,17 @@ pub struct GamepadInfo {
     pub product_id: Option<u16>,
 }
 
+impl GamepadInfo {
+    /// Creates a new instance with the given name and empty vendor and product IDs.
+    pub fn new(name: impl Into<String>) -> Self {
```
alice-i-cecile, line 428: IMO it would be better to just derive Default here.

alice-i-cecile (commented):
Dramatically less complicated. I much prefer this. A couple of cleanup notes.

@pcwalton @brandon-reinhart: you two worked to port LWIM to these changes. Can I get your reviews / opinions here?

alice-i-cecile (approved):

NiseVoid (approved):

PR #16229, opened 2024-11-04
Expose picking pointer state as a resource

NthTensor:
In `bevy_mod_picking` events are driven by several interlocking state machines, which read and write events, and share state in a few common resources. When I merged theses state machines into one to make event ordering work properly, I combined this state and hid it in a `Local`.

This PR exposes the state in a resource again. Also adds a simple little API for it. Useful for adding debug UI.

### DIFF
--- a/crates/bevy_picking/src/events.rs
+++ b/crates/bevy_picking/src/events.rs
@@ -254,13 +254,49 @@ pub struct DragEntry {
 /// An entry in the cache that drives the `pointer_events` system, storing additional data
 /// about pointer button presses.
 #[derive(Debug, Clone, Default)]
-pub struct PointerState {
+pub struct PointerButtonState {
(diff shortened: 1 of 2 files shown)


NthTensor:
This should hopefully be the final one of this series of little picking tweaks.

alice-i-cecile (changes requested):
This resource doesn't appear to be added to the app anywhere.

Otherwise I like this change!

NthTensor:
oop, yeah I rushed this a bit. Let me fix that and run some proper tests to make sure nothing else slipped through.

NthTensor:
Registered resource, testing mesh picking example and it works so I don't think I did anything else dumb.

alice-i-cecile (approved):
Much better :D

mnmaita (approved):

PR #16234, opened 2024-11-04
Bump accesskit and accesskit_winit.

ndarilek:
# Objective

- Bumps accesskit and accesskit_winit dependencies

## Solution

- Fixes several breaking API changes introduced in accesskit 0.23.

## Testing

- Tested with the ui example and seems to work comparably

### DIFF
--- a/examples/ui/scroll.rs
+++ b/examples/ui/scroll.rs
@@ -1,10 +1,7 @@
 //! This example illustrates scrolling in Bevy UI.
 
 use bevy::{
-    a11y::{
-        accesskit::{Node as Accessible, Role},
-        AccessibilityNode,
-    },
+    a11y::{accesskit::Role, AccessibilityNode},
     input::mouse::{MouseScrollUnit, MouseWheel},
     picking::focus::HoverMap,
     prelude::*,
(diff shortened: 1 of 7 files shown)


alice-i-cecile (approved):

alice-i-cecile (approved):

examples/ui/scroll.rs
```diff
@@ -2,7 +2,7 @@
 
 use bevy::{
     a11y::{
-        accesskit::{NodeBuilder, Role},
+        accesskit::{Node as Accessible, Role},
```
bushrat011899, line 5: This is an unfortunate name conflict (curse `Node` being such a good descriptor for so many things!) Strictly speaking this was an existing issue, since `NodeBuilder` from AccessKit has nothing to do with Bevy's UI node, just made apparent now.

In the examples, you only need to name the AccessKit `Node` in order to construct a Bevy `AccessibilityNode`. Perhaps we could copy the `new()` method into `AccessibilityNode` so end-users can avoid the name conflict?

bushrat011899 (changes requested):
I think this is important to try and land, but the conflict in the name `Node` is pretty unfortunate in the examples. I've proposed adding a `new` method to Bevy's `AccessibilityNode` as a way to avoid needing `accesskit::Node` in the end-user's namespace.

examples/ui/scroll.rs
ndarilek: Oh good call. Fixed, thanks for the suggestion.

PR #16243, opened 2024-11-05
Create `ReactiveSystemParam` trait

matthunz:
# Objective

- https://github.com/bevyengine/bevy/discussions/9538 mentions a possible `IsChanged` trait or similar to create systems that react to changes in their parameters

## Solution

- [x] `ReactiveQueryData`
  - Implement `ReactiveQueryData` for exported `QueryData` implementations:
    - [x] Entity
    - [x] `&T`
    - [ ] `Ref`
    - [x] `(T1, T2, ... T8)`
    - [ ] `AnyOf`
  - Some questions remain:
    - What is the best way to mark some query data items as changed? (e.g. `Entity` is currently always marked as unchanged, in the hope that `Query<(Entity, &T)>` will react when `T` is changed, but I'm not positive this is the right solution)
    - Can all `QueryData` impls be `ReactiveQueryData`? It may be best to avoid some types, or mark them as always or never changing
- [x] `ReactiveSystemParam`
  - Implemented for certain `ReadOnlySystemParam`s, the hope here is to avoid mutation during reactive systems for 2 reasons:
    1. Ordering becomes super tricky with mutating during reactive systems, most UI frameworks use immutability for this reason (but we can mutate by queuing `Command`s after reactions are finished)
    2. Running reactive systems (or potentially `Reaction` components) can be done without blocking the world, and in parallel with other reader systems.
- [ ] `ReactiveSystemParamFunction`

```rs
/// A reactive system parameter that implements change detection.
pub trait ReactiveSystemParam: ReadOnlySystemParam {
    /// The reactive state of this parameter.
    type State: Send + Sync + 'static;

    /// Initializes the reactive state of this parameter.
    fn init_state(
        world: &mut World,
        system_meta: &mut SystemMeta,
    ) -> <Self as ReactiveSystemParam>::State;

    /// Returns `true` if the parameter has changed since the last system run.
    fn is_changed(
        &self,
        world: DeferredWorld,
        state: &mut <Self as ReactiveSystemParam>::State,
    ) -> bool;
}

/// Reactive [`QueryData`].
pub trait ReactiveQueryData<F: QueryFilter>: ReadOnlyQueryData + Sized {
    /// The reactive state of this query.
    type State: Send + Sync + 'static;

    /// Initializes the reactive state of this query.
    fn init(world: &mut World) -> <Self as ReactiveQueryData<F>>::State;

    /// Returns `true` if the query has changed since the last system run.
    fn is_changed(world: DeferredWorld, state: &mut <Self as ReactiveQueryData<F>>::State) -> bool;
}
```

## Testing

- Test `ReactiveSystemParam` impl for `Res<R>`

---

## Showcase

This could eventually lead to `Reaction`s or similar (but would provide primitives for other designs as well):
https://github.com/matthunz/bevy_mod_reaction
```rs
// Reactions can also subscribe to multiple targets.
let a = commands.spawn(Health(100)).id();
let b = commands.spawn(Health(100)).id();

let mut reaction = Reaction::derive(|scope: In<Scope>, query: Query<&Health>| {
    let health = query.get(scope.entity).unwrap();
    Damage(health.0 * 2)
});

reaction.add_target(a);
reaction.add_target(b);

commands.spawn(reaction);
```

### DIFF
--- a/crates/bevy_ecs/src/change_detection.rs
+++ b/crates/bevy_ecs/src/change_detection.rs
@@ -2,6 +2,7 @@
 
 use crate::{
     component::{Tick, TickCells},
+    entity::Entity,
     prelude::Component,
     ptr::PtrMut,
     query::{Changed, QueryData, QueryFilter, ReadOnlyQueryData},
@@ -1153,6 +1154,21 @@ pub trait ReactiveQueryData<F: QueryFilter>: ReadOnlyQueryData + Sized {
 
     /// Returns `true` if the query has changed since the last system run.
     fn is_changed(world: DeferredWorld, state: &mut <Self as ReactiveQueryData<F>>::State) -> bool;
+}
+
+impl<F: QueryFilter> ReactiveQueryData<F> for Entity {
+    type State = ();
+
+    fn init(world: &mut World) -> <Self as ReactiveQueryData<F>>::State {
+        let _ = world;
+    }
+
+    fn is_changed(world: DeferredWorld, state: &mut <Self as ReactiveQueryData<F>>::State) -> bool {
+        let _ = world;
+        let _ = state;
+
+        false
+    }
 }
 
 impl<F, T> ReactiveQueryData<F> for &T
--- a/crates/bevy_ecs/src/change_detection.rs
+++ b/crates/bevy_ecs/src/change_detection.rs
@@ -3,7 +3,8 @@
 use crate::{
     component::{Tick, TickCells},
     ptr::PtrMut,
-    system::{Resource, SystemParam},
+    system::{ReadOnlySystemParam, Resource, SystemMeta},
+    world::World,
 };
 use bevy_ptr::{Ptr, UnsafeCellDeref};
 use core::{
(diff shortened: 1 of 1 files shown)


alice-i-cecile (commented):
I like the direction of this, but it badly needs docs and an example to demonstrate how it can be used :) Even tests are fine to start.

matthunz:
> I like the direction of this, but it badly needs docs and an example to demonstrate how it can be used :) Even tests are fine to start.

Awesome! I was hoping this code was heading in the right direction

I added a small test but I'll try adding some docs as well. I'm imagining these traits as lower-level primitives for reactivity so I'm not sure how ergonomic they are to use as is, but with a higher-level abstraction over top I think these will feel right at home

cart (approved):
Probably not surprising given that I suggested it in the first place, but I do think this is roughly right sized. I implemented a _slightly_ different version of this in my experiments:

1. ReactiveSystemParam does not have a duplicate `init_state`. This is unnecessary
2. Internally, implementations do not retrieve the actual system param to determine change state, as this does more work than necessary. Ex: for Res change detection, just use the stored state to look up the change ticks directly.
3. `ReactiveSystemParam::is_changed` accepts both change ticks as input, allowing the caller to decide the context. This both (1) allows the system to pass in its stored "last run" change tick when relevant and (2) allows the caller to determine what the "current" run is, rather than reading it directly off World. (2) might not be necessary, but as my implementation had looping reaction resolution, I wanted to have that flexibility.

Given how close (I think) we are to landing on the final reactivity impl, it does feel like we might want to temporarily table this until we know it is right-sized.

matthunz:
Thanks for taking a look!

> 1. ReactiveSystemParam does not have a duplicate `init_state`. This is unnecessary

I wasn’t sure about this, Queries for instance seem to demand different state in a reactive context than as a traditional system param 🤔

> 2. Internally, implementations do not retrieve the actual system param to determine change state, as this does more work than necessary. Ex: for Res change detection, just use the stored state to look up the change ticks directly.

This makes a lot of sense, I wasn’t sure about the internals so I just wanted to go the simple route. If query change detection can work without that `Changed` query hack I have here I’d definitely say I agree we should remove `State`.

> 3. `ReactiveSystemParam::is_changed` accepts both change ticks as input, allowing the caller to decide the context. This both (1) allows the system to pass in its stored "last run" change tick when relevant and (2) allows the caller to determine what the "current" run is, rather than reading it directly off World. (2) might not be necessary, but as my implementation had looping reaction resolution, I wanted to have that flexibility.
That does sound more flexible 👀

> Given how close (I think) we are to landing on the final reactivity impl, it does feel like we might want to temporarily table this until we know it is right-sized.
Sounds good 👍 I’m super excited to see the final reactivity design. For now Im definitely eager to see more experimentation

ChristopherBiscardi:
for this to be merged it would need to come out of draft, have conflicts resolved, etc. going to close for now since it is not on track to merge in its current state.

Issue #16247, opened 2024-11-05
Can not easily opt out of rendering UI on a camera


mahulst:
## What you did
I want to provide users a plugin to debug their UI. Think chrome dev tools for bevy UI. 
I also want to use gizmos to add some guides on where the currently inspected element is positioned on screen.
Gizmos are rendering behind the UI by default.
If I want the gizmos to render on top, I need to add an extra camera for my plugin with a higher Order.
The DefaultUiCamera System param defaults to the camera with the highest order. 

I can't provide a plugin that renders gizmos on top of the users UI, as there is a high chance that the plugin's camera will be the default ui camera.

## What went wrong

If a plugin spawns a camera with a higher than default order, there is a big chance it will be the DefaultUiCamera.

- what were you expecting?
I want to opt out of any UI not explicitly targeting this camera.

- what actually happened?
I need to request the user to set the IsDefaultUiCamera on there preferred camera.

## Additional information

Related discussion on #ui-dev in discord:https://discord.com/channels/691052431525675048/743663673393938453/1303324220608217169

PR #16248, opened 2024-11-05
Rename `Camera` to `RenderSurface`

aevyrie:
I'm sure this will be controversial, but I wanted to start a discussion with a concrete first step.

There are other things to rename to be consistent, but rather than do everything at once, I've started very small.

Rough proposal on discord: https://discord.com/channels/691052431525675048/743663673393938453/1300556420881846323

Discord thread converted to issue: #16249

Related to https://github.com/bevyengine/bevy/issues/16247

# Objective

- Attempt to give `Camera` a more accurate name. Reading the docs and fields, this is what the component is for. The colloquial usage of `Camera` is more appropriate for our existing `Camera3d` or `Camera2d`, which split out of the old `Camera` struct, if I recall correctly.
- Picking uses `Camera` as the common component to reason about where pointers are on screen. This is misleading, because `Camera` has nothing to do with 3d (although it still retains transforms, maybe erroneously). The reason picking uses this component is because a `Camera`, or `RenderSurface` just defines:
    - What target to write to (window or texture), and the dimensions of this target
    - A viewport within the target to constrain rendering to
    - The order to render each of these `RenderSurface`s to the render target
    - The end result, is that the job of `RenderSurface` (`Camera`) is to define how to composite layers of 2d images onto a target.
- Move UI and picking towards reasoning about a "render surface" as opposed to a "camera".
- Next steps would be to update bevy_ui to make the root of every UI contain a `RenderSurface`.

## Solution

- Rename the component.

## Migration Guide

- Rename `Camera` to `RenderSurface`.

### DIFF
--- a/crates/bevy_core_pipeline/src/bloom/settings.rs
+++ b/crates/bevy_core_pipeline/src/bloom/settings.rs
@@ -2,7 +2,7 @@ use super::downsampling_pipeline::BloomUniforms;
 use bevy_ecs::{prelude::Component, query::QueryItem, reflect::ReflectComponent};
 use bevy_math::{AspectRatio, URect, UVec4, Vec4};
 use bevy_reflect::{std_traits::ReflectDefault, Reflect};
-use bevy_render::{extract_component::ExtractComponent, prelude::Camera};
+use bevy_render::{extract_component::ExtractComponent, prelude::RenderSurface};
 
 /// Applies a bloom effect to an HDR-enabled 2d or 3d camera.
 ///
@@ -213,7 +213,7 @@ pub enum BloomCompositeMode {
 }
 
 impl ExtractComponent for Bloom {
-    type QueryData = (&'static Self, &'static Camera);
+    type QueryData = (&'static Self, &'static RenderSurface);
 
     type QueryFilter = ();
     type Out = (Self, BloomUniforms);
--- a/crates/bevy_core_pipeline/src/contrast_adaptive_sharpening/mod.rs
+++ b/crates/bevy_core_pipeline/src/contrast_adaptive_sharpening/mod.rs
@@ -71,7 +71,7 @@ impl Default for ContrastAdaptiveSharpening {
 #[reflect(Component, Default)]
 pub struct DenoiseCas(bool);
 
-/// The uniform struct extracted from [`ContrastAdaptiveSharpening`] attached to a [`Camera`].
+/// The uniform struct extracted from [`ContrastAdaptiveSharpening`] attached to a [`RenderSurface`].
 /// Will be available for use in the CAS shader.
 #[doc(hidden)]
 #[derive(Component, ShaderType, Clone)]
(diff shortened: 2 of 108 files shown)


JMS55:
I would like if we could also take this opportunity to rename the Camera::hdr field to something that doesn't imply HDR output to an HDR monitor.

Jondolf:
@cart may or may not have thoughts on this based on the discussions around splitting and/or renaming `Camera` a month ago, starting [here](https://discord.com/channels/691052431525675048/1264881140007702558/1291497144011980984). My understanding was that he overall likes `Camera` as the defining and driving type for cameras, e.g. `Camera2d` "is a" `Camera`, and that it'd be good for people to think of it that way. I feel like renaming `Camera` to `RenderSurface` moves away from the concept of what people think of as a "camera" a bit, for better or for worse. Although the two are heavily related of course.

Personally, I would be in favor of the rename. I think `RenderSurface` is more fitting and accurate in this case, and the reasoning listed in this PR makes sense. I also just find the semantics of `Camera` and the associated compositional inheritance pattern re: `Camera2d` and `Camera3d` to be somewhat confusing currently.

aevyrie:
>  I also just find the semantics of Camera and the associated compositional inheritance pattern re: Camera2d and Camera3d to be somewhat confusing currently.

Agreed, a camera entity currently has multiple `Camera_` components, which is a bit odd.

> My understanding was that he overall likes Camera as the defining and driving type for cameras, e.g. Camera2d "is a" Camera, and that it'd be good for people to think of it that way.

This one is a bit odd to me, because we already have camera components, which I find more confusing than anything. To me `Camera` is more like the "film" than the "camera". `Camera` (or `RenderSurface`) doesn't "capture" anything, it's just the surface the camera films onto, which we can then stack and composite.

cart:
I don't outright _hate_ it (as it does resolve the Camera vs Camera2d concerns), but I do have a number of sticking points that make it feel like a major downgrade to me:

1. `Camera` to me does not feel like the "film". `RenderTarget` feels like the film (although the line _is_ blurry, especially for the `hdr` field). The majority of `Camera` feels like the "camera" domain to me: selecting _what_ film to write to (render target), selecting what lenses / filters / views of the world to use (CameraRenderGraph), choosing what part of the film is exposed (viewport), choosing whether or not the aperture is open (is_active), etc. Much like a camera, you can take out the film and put it in another camera (using render targets across multiple cameras, defining the order the cameras render in).
2. `RenderSurface` feels _way_ too ambiguous with `RenderTarget`. These both read as "the thing you render to" to me. People will find this confusing.

To me, it seems like the right "conceptual framing" is for something that has `Camera` to "be a camera". `CameraRenderGraph` is a bit like configuring the lens and choosing where to point. If you have those two things _you are ready to film with your camera!_. Camera2d and Camera3d are just the stock configurations with presets you buy off the shelf.

If `Camera` was a base class, I don't think we would be having this conversation.  This whole conversation seems like resistance to the "required components / composition by inheritance" idea. And that is concerning because we have been using this pattern elsewhere (ex: MaterialNode requires Node, ImageNode requires Node, etc).

The concern that people will reach for an "empty"  base `Camera` (ex: missing a lens) is reasonable. But I don't think the best solve to that problem is contorting the design to make _less_ sense. I think our "camera" just needs a warning light when you try to film without a lens.

BenjaminBrienen:
Thanks for the great explanation! How would you prefer to see this resolved? Professionals buy a camera expecting to also need to give it a lens, so how do we create that expectation for `Camera`? Documentation?

cart:
I think the "warning light" should be a "warning log entry".

BenjaminBrienen:
triage: Has merge conflicts, received SME feedback. Probably needs to be closed and converted into a discussion where the design can be further discussed.

PR #16250, opened 2024-11-05
some refactoring

sharma-shray:
# Objective

- Describe the objective or issue this PR addresses.
- If you're fixing a specific issue, say "Fixes #X".

## Solution

some refactoring, if let could be a simpler, since the None case has no additional logic

## Testing

---

## Showcase

<details>

```rust
```

</details>

## Migration Guide

### DIFF
--- a/crates/bevy_reflect/src/serde/mod.rs
+++ b/crates/bevy_reflect/src/serde/mod.rs
@@ -15,6 +15,10 @@ mod tests {
     };
     use serde::de::DeserializeSeed;
 
+    fn custom_default() -> i32 {
+        -1
+    }
+
     #[test]
     fn test_serialization_struct() {
         #[derive(Debug, Reflect, PartialEq)]
@@ -29,10 +33,6 @@ mod tests {
             #[reflect(default = "custom_default")]
             d: i32,
             e: i32,
-        }
-
-        fn custom_default() -> i32 {
-            -1
         }
 
         let mut registry = TypeRegistry::default();
@@ -97,10 +97,6 @@ mod tests {
             i32,
             i32,
         );
-
-        fn custom_default() -> i32 {
-            -1
-        }
 
         let mut registry = TypeRegistry::default();
         registry.register::<TestStruct>();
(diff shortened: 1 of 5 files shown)


alice-i-cecile (approved):

alice-i-cecile:
Please try to include a basic PR description and title: it makes this sort of thing much easier to review.

MrGVSV (approved):
Code looks good, but this seems to touch more than just `bevy_reflect`. That should probably be mentioned in the title/description as well.

alice-i-cecile:
Merging. Please split these out and have a proper description next time though.

crates/bevy_ui/src/focus.rs
```diff
@@ -95,8 +95,7 @@ impl RelativeCursorPosition {
     /// A helper function to check if the mouse is over the node
     pub fn mouse_over(&self) -> bool {
         self.normalized
-            .map(|position| self.normalized_visible_node_rect.contains(position))
-            .unwrap_or(false)
+            .map_or(false, |position| self.normalized_visible_node_rect.contains(position))
```
mockersf, line 98: as in #16205, there's some distaste amongst some contributors (me included) with using map_or as it's harder to read because it put the "or" case first

crates/bevy_reflect/src/serde/mod.rs
```diff
@@ -15,6 +15,10 @@ mod tests {
     };
     use serde::de::DeserializeSeed;
 
+    fn custom_default() -> i32 {
+        -1
+    }
```
mockersf, line 20: I would prefer to keep this where it was, duplicated in both test. this function doesn't really make sense as a "global" function, and it reduces readability by making it far away from where it's used

crates/bevy_ui/src/focus.rs
ickshonpe: yes, hate it. 

If `map_or` is used in the middle of a chain of composed functions then I can see it might be okay (maybe?) but if placed at the end of a computation very strongly prefer `unwrap_or`.

doup (approved):

doup:
Looks like  #16205 should be closed in favor of this PR.

theotherphil:
All but one of the changes in this PR have either been fixed elsewhere or rejected by reviewers.

The remaining change - to replace a single `match` statement in `enum_utility` with an `if let Some` sequence - _is_ an improvement to code readability, but a small enough change that I don't think it's worth keeping this long-dormant PR open for.

Closing.

PR #16257, opened 2024-11-06
Remove `accesskit` re-export from `bevy_a11y`

### LINKED ISSUE #16235: AccessKit `Node` Re-export is Confusing with Bevy's `Node`
## Bevy version

`main`/0.15 RC

## Description

The `bevy_a11y` re-exports AccessKit's `Node`, but that can be confusing since `bevy_ui` _also_ defines a `Node`. Since #16234, there is now a way to use AccessKit without importing `accesskit::Node`. Now that it is optional, we should reconsider the way Bevy re-exports AccessKit. For example, only re-exporting `Role`.

## Context

This was raised by @alice-i-cecile in [this Discord message](https://discord.com/channels/691052431525675048/692572690833473578/1303095959529721896).

TurtIeSocks:
# Objective

- Fixes #16235 

## Solution

- Both Bevy and AccessKit export a `Node` struct, to reduce confusion Bevy will no longer re-export `AccessKit` from `bevy_a11y`

## Testing

- Tested locally

## Migration Guide

```diff
# main.rs
--    use bevy_a11y::{
--        accesskit::{Node, Rect, Role},
--        AccessibilityNode,
--    };
++    use bevy_a11y::AccessibilityNode;
++    use accesskit::{Node, Rect, Role};

# Cargo.toml
++    accesskit = "0.17"
```

- Users will need to add `accesskit = "0.17"` to the dependencies section of their `Cargo.toml` file and update their `accesskit` use statements to come directly from the external crate instead of `bevy_a11y`.
- Make sure to keep the versions of `accesskit` aligned with the versions Bevy uses.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -474,6 +474,7 @@ hyper = { version = "1", features = ["server", "http1"] }
 http-body-util = "0.1"
 anyhow = "1"
 macro_rules_attribute = "0.2"
+accesskit = "0.17"
 
 [target.'cfg(not(target_family = "wasm"))'.dev-dependencies]
 smol = "2"
--- a/crates/bevy_a11y/src/lib.rs
+++ b/crates/bevy_a11y/src/lib.rs
@@ -6,13 +6,18 @@
 )]
 
 //! Accessibility for Bevy
+//! 
+//! As of Bevy version 0.15 `accesskit` is no longer re-exported from this crate.
+//! 
+//! If you need to use `accesskit`, you will need to add it as a separate dependency in your `Cargo.toml`.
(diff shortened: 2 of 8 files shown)


TurtIeSocks:
The issue mentioned updating the docs but I wasn't 100% sure where you wanted the note to be added, so I just added it to the top of the `bevy_a11y` module for now and in the body of the PR. Happy to write something better if necessary.

cart (approved):

cart:
I like this general direction. This does create some usability issues, but imo the way these concepts all relate to each other needs work.

PR #14561, opened 2024-07-31
Separate component and resource access

cBournhonesque:
# Objective

- Fixes https://github.com/bevyengine/bevy/issues/13139
- Fixes https://github.com/bevyengine/bevy/issues/7255
- Separates component from resource access so that we can correctly handles edge cases like the issue above
- Inspired from https://github.com/bevyengine/bevy/pull/14472

## Solution

- Update access to have `component` fields and `resource` fields

## Testing

- Added some unit tests

PR #16263, opened 2024-11-06
Enable EntityRef tests that now pass.

chescock:
# Objective

Re-enable some tests in `entity_ref.rs` that are marked as `#[ignore]`, but that pass after #14561.  

## Solution

Remove `#[ignore]` from those tests.


alice-i-cecile (approved):
Yay!

PR #16264, opened 2024-11-06
Text2d scalefactor change detection fix 

### LINKED ISSUE #16223: Text doesn't respond correctly to scale changes
## Bevy version

Bevy 0.15.0-rc.2

## What you did

I spawned a `Text2d` during `Setup`

## What went wrong

The text gets a different size and position at a scale factor of 2 compared to a scale factor of 1.

## Additional information

Text that is updated after the scale factor is applied seems to behave as expected.

Scale factor 1:
Image

Scale factor 2:
Image

The FPS text is updated each frame and thus avoids this issue.

ickshonpe:
# Objective 

Text2d doesn't respond to changes to the window scalefactor.

Fixes #16223

## Solution 

In `update_text2d_layout` store the previous scale factor in a `Local` instead and check against the current scale factor to detect changes.

It seems like previously the text wasn't updated because of a bug with the `WindowScaleFactorChanged` event and it isn't emitted after changes to the scale factor.  That needs to be looked into, but this will work for now.

## Testing

Really simple app that draws a big message in the middle of the window:

```
use bevy::prelude::*;

fn main() {
    App::new()
        .add_plugins(DefaultPlugins)
        .add_systems(Startup, setup)
        .run();
}

    commands.spawn(Camera2d);
    commands.spawn((
        Text2d::new("Hello"),
        TextFont {
            font_size: 400.,
            ..Default::default()
        },
    ));
}
```

Looks fine:
hello1

On main, after changing the monitor's scale factor:
hello2

With this PR the text maintains the same size and position after the scale factor is changed.

### DIFF
--- a/crates/bevy_text/src/text2d.rs
+++ b/crates/bevy_text/src/text2d.rs
@@ -29,7 +29,7 @@ use bevy_sprite::{Anchor, ExtractedSprite, ExtractedSprites, SpriteSource, Textu
 use bevy_transform::components::Transform;
 use bevy_transform::prelude::GlobalTransform;
 use bevy_utils::HashSet;
-use bevy_window::{PrimaryWindow, Window;
+use bevy_window::{PrimaryWindow, Window};
 
 /// [`Text2dBundle`] was removed in favor of required components.
 /// The core component is now [`Text2d`] which can contain a single text segment.
--- a/crates/bevy_text/src/text2d.rs
+++ b/crates/bevy_text/src/text2d.rs
@@ -358,7 +358,7 @@ mod tests {
 
     use bevy_app::{App, Update};
     use bevy_asset::{load_internal_binary_asset, Handle};
-    use bevy_ecs::{event::Events, schedule::IntoSystemConfigs};
+    use bevy_ecs::schedule::IntoSystemConfigs;
 
     use crate::{detect_text_needs_rerender, TextIterScratch};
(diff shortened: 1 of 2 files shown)


alice-i-cecile (approved):

ickshonpe:
> the `WindowScaleFactorChanged` event is also used to update ui layout, I would prefer a complete fix rather than "just" text.

It's actually more complicated than just fixing the event, there are a couple issues with  `update_text2d_layout`:
* It regenerates the text2d layout on recieving a `WindowScaleFactorChanged` event but the render target could be a different window or a texture.
* It uses the scale factor from the primary window regardless of the render target.
* It doesn't delete the old unused font atlases for the previous scale factors.

So storing the scale factor in a local variable is marginallypreferable to watching the event. With the event (if it worked) it triggers updates on changes to the scale factor of any window, not just the primary window. But since text2d always uses the scale factor of the primary window, the text reupdate is needless on changes to non-primary window's scalefactors. 

It's stupid how it works this way though and we need proper multi-window support.

> 
> The issue is probably about the first scale change on window creation. I can't reproduce on macOS, but it may be os dependent

It can be tricky sometimes to construct examples that capture scale factor bugs. Easiest way in this case is if you have two monitors with different scale factors and drag the window from one to the other.

mockersf (approved):

Issue #1481, opened 2021-02-19
Archetype Invariants


alice-i-cecile:
This is a tracking issue for archetype invariants, introduced in [#1312](https://github.com/bevyengine/bevy/discussions/1312#discussioncomment-347678). As this feature's use cases grow beyond resolving ambiguities, it makes sense to have a central place to discuss its design.

# The Basics

An *archetype* is the set of components that are found together on a single entity. The set of *archetypes* present in our `World` is the union of the archetypes of every entity in the world. 

**Archetype invariants** are rules that limit which components can coexist, limiting the possible archetypes that can co-occur. Because of the flexible power of `commands.spawn` and `commands.insert`, the components that an entity could have is unknowable at compile time, preventing us from . 

We can use this information to allow for more granular component access in ways that would otherwise result in inconsistent runtime behavior, or verify that certain logical rules are followed during execution.

Due to the overhead of verifying these rules against the `World`'s archetypes, this is likely best done only during debug mode, or perhaps by post-hoc inspection of logs.

# Use Cases

1. Reducing false positives in system scheduling ambiguity detection (#1312)
2. Assumption checking for safety
3. More granular overlapping queries (to expand on #1320)

# API for specifying archetype invariants

**Primitives:**
- `forbidden(my_bundle)`: an entities archetype cannot have have the specific combination of components listed in the bundle as a subset
- 'A.always_with(B)':  entities that have component A always also have component B
- `A.only_with(my_bundle)': component A never occurs with components other than those in the bundle

**Derived:**
- `inseparable(my_bundle)`: entities that have either component A or B never occur without each other. Equivalent to combining `A.always_with(B)` with the reverse for every pairwise combination
- `disjoint(my_bundle)`: entities can only have at most one component in the bundle. Equivalent to creating a `forbidden` archetype invariant for each pairwise combination
- bundle versions of all of the primitives, where A and B can be replaced by tuples of components (bundles) instead

alice-i-cecile:
A new thing that I want: `or` modifiers to allow one component in a bundle to replace each other locally. And `replacable`, which allows you to do so everywhere. This enables enum-like component groups in a really nice way.

For context, I want this for [overriding behavior](https://github.com/bevy-cheatbook/bevy-cheatbook/issues/18) in a checked and elegant way.

alice-i-cecile:
[...]

PR #16267, opened 2024-11-06
More forms of related components: suggested and included components

### LINKED ISSUE #15580: Support removing a removed Component's required components

Sometimes it is desirable to remove not just a specific component, but also its required components: https://github.com/bevyengine/bevy/pull/15572#discussion_r1783376860.

Add an `entity.remove_with_requires::<SomeComponent>()` API that removes all required components brought in by SomeComponent (recursively).

* This should not be the default behavior of `entity.remove::<SomeComponent>()`, as in many cases this could remove components that are required by _other_ components. If we want to go this route, we need to consider building some sort of reference counting system to track requires across components. But there are still problems there (ex: what if someone manually added a component ... how do we distinguish between overriding a required component value and _inserting_ a new separate persistent component).

### LINKED ISSUE #16246: Suggested / related components

Required components are great for explaining the relation between components when the relationship is mandatory, but aren't great when it's optional.

We've seen this come up in Bevy itself, with various texture atlas components being fully optional, but being intended to work with both UI images and sprites. Similarly, we could split a lot of fields off of `Node` (previously `Style`) by simply making them have no effect when they're missing. It would be really nice to have a standardized and tooling friendly way to explain these links.

To take an example from one of my crates: `ActionState` and `InputMap` are intended to work together, but you *can* use `ActionState` without an `InputMap` (great for AI). It would be nice to be able to document that edge in a structured and straightforward way.

A parallel `Component` annotation to the `requires` , called `suggests`. To start, this just generates documentation for the `Component` trait.

Afterwards, this should store runtime information for tooling. This will allow us to draw a graph of components and their relationships.

Downstream crates should be able to augment the list of suggested components for components they don't own, to show where they hook into existing primitives.

Primarily, this generates documentation using

We could manually document these, but a) it's more tedious to remember to do b) this can't be automatically discovered by tooling.

## Additional context

Raised on [Discord](https://discord.com/channels/691052431525675048/743559241461399582/1301693168378712065) and was broadly popular there!

alice-i-cecile:
# Objective

As discussed in #16246, required components are a great first step, but it makes sense to use this machinery for less stringent relatonships.

Similarly, https://github.com/bevyengine/bevy/issues/15580 complains that required components are too restrictive for power users with unusual needs, and should be able to be disabled at runtime.

## Solution

As discussed [on Discord](https://discord.com/channels/691052431525675048/1303383992174645389), I've added two new variants of required components (collectively, related components):

- Suggested Components: just docs!
- Included Components: like Required Components, but less strict.

See the `Relatedness` docs included with this PR for the exact behavior.

Fixes #16246; fixes #15580.

## Testing

So far, it compiles...

## TODO

- [ ] write tests
- [ ] double check main docs
- [ ] add removal machinery
- [ ] try to make the error handling nicer
- [ ] rename related components to something that can't be confused with relations (probably linked or affiliated components)
- [ ] split suggested components out from this PR

### DIFF
--- a/crates/bevy_ecs/macros/src/component.rs
+++ b/crates/bevy_ecs/macros/src/component.rs
@@ -82,7 +82,7 @@ pub fn derive_component(input: TokenStream) -> TokenStream {
         for require in requires {
             let ident = &require.path;
             register_recursive_requires.push(quote! {
-                <#ident as Component>::register_required_components(
+                <#ident as Component>::register_related_components(
                     requiree,
                     components,
                     storages,
@@ -93,7 +93,7 @@ pub fn derive_component(input: TokenStream) -> TokenStream {
             match &require.func {
                 Some(RequireFunc::Path(func)) => {
                     register_required.push(quote! {
-                        components.register_required_components_manual::<Self, #ident>(
+                        components.register_related_components_manual::<Self, #ident>(
                             storages,
                             required_components,
                             || { let x: #ident = #func().into(); x },
@@ -103,7 +103,7 @@ pub fn derive_component(input: TokenStream) -> TokenStream {
                 }
                 Some(RequireFunc::Closure(func)) => {
                     register_required.push(quote! {
-                        components.register_required_components_manual::<Self, #ident>(
+                        components.register_related_components_manual::<Self, #ident>(
                             storages,
                             required_components,
                             || { let x: #ident = (#func)().into(); x },
@@ -113,7 +113,7 @@ pub fn derive_component(input: TokenStream) -> TokenStream {
                 }
                 None => {
                     register_required.push(quote! {
-                        components.register_required_components_manual::<Self, #ident>(
+                        components.register_related_components_manual::<Self, #ident>(
                             storages,
                             required_components,
                             <#ident as Default>::default,
@@ -145,11 +145,11 @@ pub fn derive_component(input: TokenStream) -> TokenStream {
         #required_component_docs
         impl #impl_generics #bevy_ecs_path::component::Component for #struct_name #type_generics #where_clause {
             const STORAGE_TYPE: #bevy_ecs_path::component::StorageType = #storage;
-            fn register_required_components(
+            fn register_related_components(
                 requiree: #bevy_ecs_path::component::ComponentId,
                 components: &mut #bevy_ecs_path::component::Components,
                 storages: &mut #bevy_ecs_path::storage::Storages,
-                required_components: &mut #bevy_ecs_path::component::RequiredComponents,
+                required_components: &mut #bevy_ecs_path::component::RelatedComponents,
                 inheritance_depth: u16,
             ) {
                 #(#register_required)*
--- a/crates/bevy_ecs/macros/src/component.rs
+++ b/crates/bevy_ecs/macros/src/component.rs
@@ -75,79 +75,31 @@ pub fn derive_component(input: TokenStream) -> TokenStream {
         .predicates
         .push(parse_quote! { Self: Send + Sync + 'static });
 
-    let requires = &attrs.requires;
-    let mut register_required = Vec::with_capacity(attrs.requires.iter().len());
-    let mut register_recursive_requires = Vec::with_capacity(attrs.requires.iter().len());
+    let RelatedComponentsReturn {
+        register_related: register_suggested,
+        register_recursive_related: register_recursive_suggested,
+        docs: suggested_component_docs,
+    } = related_components(&attrs.suggests, Relatedness::Suggested);
+    let RelatedComponentsReturn {
+        register_related: register_included,
+        register_recursive_related: register_recursive_included,
+        docs: included_component_docs,
+    } = related_components(&attrs.includes, Relatedness::Included);
+    let RelatedComponentsReturn {
+        register_related: register_required,
+        register_recursive_related: register_recursive_required,
+        docs: required_component_docs,
+    } = related_components(&attrs.requires, Relatedness::Required);
 
-    if let Some(requires) = requires {
-        let relatedness = quote! { #bevy_ecs_path::component::Relatedness::Required };
-        for require in requires {
-            let ident = &require.path;
-            register_recursive_requires.push(quote! {
-                <#ident as Component>::register_related_components(
-                    requiree,
-                    components,
-                    storages,
-                    required_components,
-                    inheritance_depth + 1,
-                );
-            });
-            match &require.func {
-                Some(RequireFunc::Path(func)) => {
-                    register_required.push(quote! {
-                        components.register_related_components_manual::<Self, #ident>(
-                            storages,
-                            required_components,
-                            || { let x: #ident = #func().into(); x },
-                            inheritance_depth,
-                            #relatedness,
-                        );
-                    });
-                }
-                Some(RequireFunc::Closure(func)) => {
-                    register_required.push(quote! {
-                        components.register_related_components_manual::<Self, #ident>(
-                            storages,
-                            required_components,
-                            || { let x: #ident = (#func)().into(); x },
-                            inheritance_depth,
-                            #relatedness,
-                        );
-                    });
-                }
-                None => {
-                    register_required.push(quote! {
-                        components.register_related_components_manual::<Self, #ident>(
-                            storages,
-                            required_components,
-                            <#ident as Default>::default,
-                            inheritance_depth,
-                            #relatedness,
-                        );
-                    });
-                }
-            }
-        }
-    }
     let struct_name = &ast.ident;
     let (impl_generics, type_generics, where_clause) = &ast.generics.split_for_impl();
-
-    let required_component_docs = attrs.requires.map(|r| {
-        let paths = r
-            .iter()
-            .map(|r| format!("[`{}`]", r.path.to_token_stream()))
-            .collect::<Vec<_>>()
-            .join(", ");
-        let doc = format!("Required Components: {paths}. \n\n A component's Required Components are inserted whenever it is inserted. Note that this will also insert the required components _of_ the required components, recursively, in depth-first order.");
-        quote! {
-            #[doc = #doc]
-        }
-    });
 
     // This puts `register_required` before `register_recursive_requires` to ensure that the constructors of _all_ top
     // level components are initialized first, giving them precedence over recursively defined constructors for the same component type
     TokenStream::from(quote! {
         #required_component_docs
+        #included_component_docs
+        #suggested_component_docs
         impl #impl_generics #bevy_ecs_path::component::Component for #struct_name #type_generics #where_clause {
             const STORAGE_TYPE: #bevy_ecs_path::component::StorageType = #storage;
             fn register_related_components(
@@ -157,8 +109,14 @@ pub fn derive_component(input: TokenStream) -> TokenStream {
                 required_components: &mut #bevy_ecs_path::component::RelatedComponents,
                 inheritance_depth: u16,
             ) {
+                #(#register_suggested)*
+                #(#register_recursive_suggested)*
+
+                #(#register_included)*
+                #(#register_recursive_included)*
+
                 #(#register_required)*
-                #(#register_recursive_requires)*
+                #(#register_recursive_required)*
             }
 
             #[allow(unused_variables)]
@@ -172,8 +130,136 @@ pub fn derive_component(input: TokenStream) -> TokenStream {
     })
 }
 
+// Because the macro crate cannot access the `bevy_ecs` crate, we need to create our own equivalent.
+// This should be kept in sync with the actual `bevy_ecs` crate!
+enum Relatedness {
+    Suggested,
+    Included,
+    Required,
+}
+
+impl Relatedness {
+    /// Returns the token that represents the corresponding `Relatedness` enum variant in `bevy_ecs`.
+    fn token(&self) -> TokenStream2 {
+        let bevy_ecs_path: Path = crate::bevy_ecs_path();
+        match self {
+            Relatedness::Suggested => quote! { #bevy_ecs_path::component::Relatedness::Included },
+            Relatedness::Included => quote! { #bevy_ecs_path::component::Relatedness::Included },
+            Relatedness::Required => quote! { #bevy_ecs_path::component::Relatedness::Required },
+        }
+    }
+
+    /// Returns the stringified name of this `Relatedness` enum variant for use in docs.
+    fn doc_name(&self) -> &'static str {
+        match self {
+            Relatedness::Suggested => "Suggested",
+            Relatedness::Included => "Included",
+            Relatedness::Required => "Required",
+        }
+    }
+
+    /// Returns the doc string fragment that explains the corresponding `Relatedness` enum variant.
+    fn doc_explanation(&self) -> &'static str {
+        match self {
+            Relatedness::Suggested => "This component might work well with the components listed above, unlocking new functionality.",
+            Relatedness::Included => "A component's Included Components are inserted whenever it is inserted. Note that this will also insert the included components _of_ the included components, recursively, in depth-first order.",
+            Relatedness::Required => "A component's Required Components are inserted whenever it is inserted. Note that this will also insert the required components _of_ the required components, recursively, in depth-first order. Unlike included components, this relationship cannot be removed.",
+        }
+    }
+}
+
+struct RelatedComponentsReturn {
+    register_related: Vec<TokenStream2>,
+    register_recursive_related: Vec<TokenStream2>,
+    docs: Option<TokenStream2>,
+}
+
+/// Generates the code needed to add related components to the component's related components list.
+fn related_components(
+    attribute: &Option<Punctuated<Related, Comma>>,
+    relatedness: Relatedness,
+) -> RelatedComponentsReturn {
+    let mut register_related = Vec::with_capacity(attribute.iter().len());
+    let mut register_recursive_related = Vec::with_capacity(attribute.iter().len());
+
+    let relatedness_token = relatedness.token();
+
+    if let Some(related) = attribute {
+        for require in related {
+            let ident = &require.path;
+            register_recursive_related.push(quote! {
+                <#ident as Component>::register_related_components(
+                    requiree,
+                    components,
+                    storages,
+                    required_components,
+                    inheritance_depth + 1,
+                );
+            });
+            match &require.func {
+                Some(RequireFunc::Path(func)) => {
+                    register_related.push(quote! {
+                        components.register_related_components_manual::<Self, #ident>(
+                            storages,
+                            required_components,
+                            || { let x: #ident = #func().into(); x },
+                            inheritance_depth,
+                            #relatedness_token,
+                        );
+                    });
+                }
+                Some(RequireFunc::Closure(func)) => {
+                    register_related.push(quote! {
+                        components.register_related_components_manual::<Self, #ident>(
+                            storages,
+                            required_components,
+                            || { let x: #ident = (#func)().into(); x },
+                            inheritance_depth,
+                            #relatedness_token,
+                        );
+                    });
+                }
+                None => {
+                    register_related.push(quote! {
+                        components.register_related_components_manual::<Self, #ident>(
+                            storages,
+                            required_components,
+                            <#ident as Default>::default,
+                            inheritance_depth,
+                            #relatedness_token,
+                        );
+                    });
+                }
+            }
+        }
+    }
+
+    let docs = attribute.as_ref().map(|r| {
+        let paths = r
+            .iter()
+            .map(|r| format!("[`{}`]", r.path.to_token_stream()))
+            .collect::<Vec<_>>()
+            .join(", ");
+        let doc_name = relatedness.doc_name();
+        let doc_explanation = relatedness.doc_explanation();
+        let doc = format!("{doc_name}: {paths}. \n\n {doc_explanation}");
+        quote! {
+            #[doc = #doc]
+        }
+    });
+
+    RelatedComponentsReturn {
+        register_related,
+        register_recursive_related,
+        docs,
+    }
+}
+
 pub const COMPONENT: &str = "component";
 pub const STORAGE: &str = "storage";
+
+pub const SUGGEST: &str = "suggest";
+pub const INCLUDE: &str = "include";
 pub const REQUIRE: &str = "require";
 
 pub const ON_ADD: &str = "on_add";
@@ -183,7 +269,9 @@ pub const ON_REMOVE: &str = "on_remove";
 
 struct Attrs {
     storage: StorageTy,
-    requires: Option<Punctuated<Require, Comma>>,
+    suggests: Option<Punctuated<Related, Comma>>,
+    includes: Option<Punctuated<Related, Comma>>,
+    requires: Option<Punctuated<Related, Comma>>,
     on_add: Option<ExprPath>,
     on_insert: Option<ExprPath>,
     on_replace: Option<ExprPath>,
@@ -196,7 +284,7 @@ enum StorageTy {
     SparseSet,
 }
 
-struct Require {
+struct Related {
     path: Path,
     func: Option<RequireFunc>,
 }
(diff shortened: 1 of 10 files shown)


crates/bevy_ecs/src/component.rs
```diff
@@ -1622,11 +1641,13 @@ impl<T: Component> FromWorld for InitComponentId<T> {
...
-/// An error returned when the registration of a required component fails.
+/// An error returned when the registration of a related component fails.
+///
+/// See [`RelatedComponents`] for the place where this information is stored and [`Relatedness`] for the strictness of the relationship.
 #[derive(Error, Display, Debug)]
 #[non_exhaustive]
-pub enum RequiredComponentsError {
-    /// The component is already a directly required component for the requiree.
+pub enum RelatedComponentsError {
+    /// The component is already a directly related component for the requiree.
```
omaskery, line 1650: Hrm, what to do about `requiree`?

I wonder if we should just come up with some convention for referring to each end of a related component, in the vein of "source" (requiree) and "destination" (required), or "subject" and "related", or (insert other bikeshedding ideas).

crates/bevy_ecs/src/component.rs
```diff
@@ -1622,11 +1641,13 @@ impl<T: Component> FromWorld for InitComponentId<T> {
...
     #[display("Component {0:?} already directly requires component {_1:?}")]
```
omaskery, line 1651:
```suggestion
    #[display("Component {0:?} already directly related to component {_1:?}")]
```

crates/bevy_ecs/src/component.rs
```diff
@@ -1638,27 +1659,27 @@ pub enum RequiredComponentsError {
...
 #[derive(Clone)]
-pub struct RequiredComponentConstructor(
+pub struct RelatedComponentConstructor(
     pub Arc<dyn Fn(&mut Table, &mut SparseSets, Tick, TableRow, Entity)>,
 );
 
-impl RequiredComponentConstructor {
+impl RelatedComponentConstructor {
     /// # Safety
     /// This is intended to only be called in the context of [`BundleInfo::write_components`] to initialized required components.
```
omaskery, line 1678:
```suggestion
    /// This is intended to only be called in the context of [`BundleInfo::write_components`] to initialize related components.
```

crates/bevy_ecs/src/component.rs
```diff
@@ -1638,27 +1659,27 @@ pub enum RequiredComponentsError {
...
     /// Calling it _anywhere else_ should be considered unsafe.
     ///
     /// `table_row` and `entity` must correspond to a valid entity that currently needs a component initialized via the constructor stored
-    /// on this [`RequiredComponentConstructor`]. The stored constructor must correspond to a component on `entity` that needs initialization.
+    /// on this [`RelatedComponentConstructor`]. The stored constructor must correspond to a component on `entity` that needs initialization.
     /// `table` and `sparse_sets` must correspond to storages on a world where `entity` needs this required component initialized.
```
omaskery, line 1683:
```suggestion
    /// `table` and `sparse_sets` must correspond to storages on a world where `entity` needs this related component initialized.
```

crates/bevy_ecs/src/component.rs
```diff
@@ -1683,11 +1704,13 @@ impl RequiredComponentConstructor {
     }
 }
 
-/// Metadata associated with a required component. See [`Component`] for details.
+/// Metadata associated with a required component.
```
omaskery, line 1707:
```suggestion
/// Metadata associated with a related component.
```

crates/bevy_ecs/src/component.rs
```diff
@@ -1683,11 +1704,13 @@ impl RequiredComponentConstructor {
...
+/// See [`Component`] for details on usage, and [`RelatedComponents`] for the storage.
 #[derive(Clone)]
-pub struct RequiredComponent {
-    /// The constructor used for the required component.
-    pub constructor: RequiredComponentConstructor,
+pub struct RelatedComponent {
+    /// The constructor used for the related component.
+    pub constructor: RelatedComponentConstructor,
 
     /// The depth of the component requirement in the requirement hierarchy for this component.
```
omaskery, line 1715: I can't do suggestions across lines that aren't shown by the diff, but:

```rs
    /// The depth of the component relation in the related components hierarchy for this component.
    /// This is used for determining which constructor is used in cases where there are duplicate relations.
    ///
    /// For example, consider the inheritance tree `X -> Y -> Z`, where `->` indicates a relation.
    /// `X -> Y` and `Y -> Z` are direct relations with a depth of 0, while `Z` is only indirectly
    /// related to `X` with a depth of `1`.
    ///
    /// In cases where there are multiple conflicting relations with the same depth, a higher priority
    /// will be given to components listed earlier in the `require`/`include` attribute, or to the latest added
    /// relation if registered at runtime.
```

crates/bevy_ecs/src/component.rs
```diff
@@ -1702,22 +1725,81 @@ pub struct RequiredComponent {
     pub inheritance_depth: u16,
 }
 
-/// The collection of metadata for components that are required for a given component.
+/// The collection of metadata for components that are related to a given component.
 ///
-/// For more information, see the "Required Components" section of [`Component`].
+/// For usage information, see the "Related Components" section of [`Component`].
+/// For the metadata associated with a relayed component, see [`RelatedComponent`].
```
omaskery, line 1731:
```suggestion
/// For the metadata associated with a related component, see [`RelatedComponent`].
```

crates/bevy_ecs/src/component.rs
```diff
@@ -1702,22 +1725,81 @@ pub struct RequiredComponent {
...
+    ///
+    /// This level is appropriate to link components that should almost always be used together,
+    /// but where power users may wish to opt-out of this behavior.
+    Included,
+    /// This component must be used with the related component.
+    ///
+    /// In addition to the features of [`Relatedness::Suggested`],
+    /// whenever this component is added to an entity, the relationship between these components cannot be removed.
+    /// In the future, we intend to this level to enforce even stricter guarantees, such as removing this component
+    /// if any of its required component is removed.
```
omaskery, line 1791: I think this section maybe got interrupted part way through editing after duplicating from the `Included` comment, perhaps:

```suggestion
    /// In addition to the features of [`Relatedness::Suggested`],
    /// whenever this component is added to an entity, the related component will also be added if it was not already provided.
    /// The relationship between these components cannot be removed.
    /// In the future, we intend to this level to enforce even stricter guarantees, such as removing this component
    /// if any of its required component is removed.
```

crates/bevy_ecs/src/lib.rs
```diff
@@ -82,7 +82,7 @@ pub mod prelude {
 #[cfg(test)]
 mod tests {
```
omaskery, line 83: Is this intended as the minimal change to support the new concepts, and tests can be added in a future PR?

crates/bevy_ecs/src/system/query.rs
```diff
@@ -477,7 +477,7 @@ impl<'w, 's, D: QueryData, F: QueryFilter> Query<'w, 's, D, F> {
     #[inline]
     pub fn iter(&self) -> QueryIter<'_, 's, D::ReadOnly, F> {
         // SAFETY:
-        // - `self.world` has permission to access the required components.
+        // - `self.world` has permission to access the related components.
```
omaskery, line 480: This doesn't appear to be related to required/related components, I think this is just a find&replace oopsie!

crates/bevy_ecs/src/system/query.rs
```diff
@@ -548,7 +548,7 @@ impl<'w, 's, D: QueryData, F: QueryFilter> Query<'w, 's, D, F> {
         &self,
     ) -> QueryCombinationIter<'_, 's, D::ReadOnly, F, K> {
         // SAFETY:
-        // - `self.world` has permission to access the required components.
+        // - `self.world` has permission to access the related components.
```
omaskery, line 551: Same as above: this doesn't appear to be related to required/related components, I think this is just a find&replace oopsie!

crates/bevy_ecs/src/system/query.rs
```diff
@@ -634,7 +634,7 @@ impl<'w, 's, D: QueryData, F: QueryFilter> Query<'w, 's, D, F> {
         entities: EntityList,
     ) -> QueryManyIter<'_, 's, D::ReadOnly, F, EntityList::IntoIter> {
         // SAFETY:
-        // - `self.world` has permission to access the required components.
+        // - `self.world` has permission to access the related components.
```
omaskery, line 637: Same as above: this doesn't appear to be related to required/related components, I think this is just a find&replace oopsie!

crates/bevy_ecs/src/system/query.rs
```diff
@@ -711,7 +711,7 @@ impl<'w, 's, D: QueryData, F: QueryFilter> Query<'w, 's, D, F> {
     #[inline]
     pub unsafe fn iter_unsafe(&self) -> QueryIter<'_, 's, D, F> {
         // SAFETY:
-        // - `self.world` has permission to access the required components.
+        // - `self.world` has permission to access the related components.
```
omaskery, line 714: Same as above: this doesn't appear to be related to required/related components, I think this is just a find&replace oopsie!

crates/bevy_ecs/src/system/query.rs
```diff
@@ -737,7 +737,7 @@ impl<'w, 's, D: QueryData, F: QueryFilter> Query<'w, 's, D, F> {
         &self,
     ) -> QueryCombinationIter<'_, 's, D, F, K> {
         // SAFETY:
-        // - `self.world` has permission to access the required components.
+        // - `self.world` has permission to access the related components.
```
omaskery, line 740: Same as above: this doesn't appear to be related to required/related components, I think this is just a find&replace oopsie!

crates/bevy_ecs/src/system/query.rs
```diff
@@ -764,7 +764,7 @@ impl<'w, 's, D: QueryData, F: QueryFilter> Query<'w, 's, D, F> {
         entities: EntityList,
     ) -> QueryManyIter<'_, 's, D, F, EntityList::IntoIter> {
         // SAFETY:
-        // - `self.world` has permission to access the required components.
+        // - `self.world` has permission to access the related components.
```
omaskery, line 767: Same as above: this doesn't appear to be related to required/related components, I think this is just a find&replace oopsie!

crates/bevy_ecs/src/world/mod.rs
```diff
@@ -293,22 +293,22 @@ impl World {
...
-    /// When `T` is added to an entity, `R` and its own required components will also be added
+    /// When `T` is added to an entity, `R` and its own related components will also be added
     /// if `R` was not already provided. The [`Default`] `constructor` will be used for the creation of `R`.
-    /// If a custom constructor is desired, use [`World::register_required_components_with`] instead.
+    /// If a custom constructor is desired, use [`World::register_related_components_with`] instead.
     ///
-    /// For the non-panicking version, see [`World::try_register_required_components`].
+    /// For the non-panicking version, see [`World::try_register_related_components`].
     ///
     /// Note that requirements must currently be registered before `T` is inserted into the world
```
omaskery, line 304:
```suggestion
    /// Note that relations must currently be registered before `T` is inserted into the world
```

crates/bevy_ecs/src/world/mod.rs
```diff
@@ -293,22 +293,22 @@ impl World {
...
-    /// [required component]: Component#required-components
+    /// [related component]: Component#related-components
     ///
     /// # Panics
     ///
-    /// Panics if `R` is already a directly required component for `T`, or if `T` has ever been added
+    /// Panics if `R` is already a directly related component for `T`, or if `T` has ever been added
     /// on an entity before the registration.
     ///
     /// Indirect requirements through other components are allowed. In those cases, any existing requirements
```
omaskery, line 314:
```suggestion
    /// Indirect relations through other components are allowed. In those cases, any existing relations
```

crates/bevy_ecs/src/world/mod.rs
```diff
@@ -329,34 +329,38 @@ impl World {
     ///
     /// # let mut world = World::default();
     /// // Register B as required by A and C as required by B.
```
omaskery, line 331:
```suggestion
    /// // Register B as related to A and C as related to B.
```

crates/bevy_ecs/src/world/mod.rs
```diff
@@ -329,34 +329,38 @@ impl World {
...
+    /// When `T` is added to an entity, `R` and its own related components will also be added
     /// if `R` was not already provided. The given `constructor` will be used for the creation of `R`.
-    /// If a [`Default`] constructor is desired, use [`World::register_required_components`] instead.
+    /// If a [`Default`] constructor is desired, use [`World::register_related_components`] instead.
     ///
-    /// For the non-panicking version, see [`World::try_register_required_components_with`].
+    /// For the non-panicking version, see [`World::try_register_related_components_with`].
     ///
     /// Note that requirements must currently be registered before `T` is inserted into the world
     /// for the first time. This limitation may be fixed in the future.
```
omaskery, line 357:
```suggestion
    /// Note that relations must currently be registered before `T` is inserted into the world
    /// for the first time. This limitation may be fixed in the future.
```

crates/bevy_ecs/src/world/mod.rs
```diff
@@ -329,34 +329,38 @@ impl World {
...
-    /// [required component]: Component#required-components
+    /// [related component]: Component#related-components
     ///
     /// # Panics
     ///
-    /// Panics if `R` is already a directly required component for `T`, or if `T` has ever been added
+    /// Panics if `R` is already a directly related component for `T`, or if `T` has ever been added
     /// on an entity before the registration.
     ///
     /// Indirect requirements through other components are allowed. In those cases, any existing requirements
```
omaskery, line 366:
```suggestion
    /// Indirect relations through other components are allowed. In those cases, any existing relations
```

crates/bevy_ecs/src/world/mod.rs
```diff
@@ -378,40 +382,41 @@ impl World {
     /// # let mut world = World::default();
     /// // Register B and C as required by A and C as required by B.
     /// // A requiring C directly will overwrite the indirect requirement through B.
```
omaskery, line 384:
```suggestion
    /// // Register B and C as related to A and C as related to B.
    /// // A relating to C directly will overwrite the indirect relation through B.
```

crates/bevy_ecs/src/world/mod.rs
```diff
@@ -378,40 +382,41 @@ impl World {
...
+    /// When `T` is added to an entity, `R` and its own related components will also be added
     /// if `R` was not already provided. The [`Default`] `constructor` will be used for the creation of `R`.
-    /// If a custom constructor is desired, use [`World::register_required_components_with`] instead.
+    /// If a custom constructor is desired, use [`World::register_related_components_with`] instead.
     ///
-    /// For the panicking version, see [`World::register_required_components`].
+    /// For the panicking version, see [`World::register_related_components`].
     ///
     /// Note that requirements must currently be registered before `T` is inserted into the world
     /// for the first time. This limitation may be fixed in the future.
```
omaskery, line 413:
```suggestion
    /// Note that relations must currently be registered before `T` is inserted into the world
    /// for the first time. This limitation may be fixed in the future.
```

crates/bevy_ecs/src/world/mod.rs
```diff
@@ -378,40 +382,41 @@ impl World {
...
     ///
-    /// [required component]: Component#required-components
+    /// [related component]: Component#related-components
     ///
     /// # Errors
     ///
-    /// Returns a [`RequiredComponentsError`] if `R` is already a directly required component for `T`, or if `T` has ever been added
+    /// Returns a [`RelatedComponentsError`] if `R` is already a directly required component for `T`, or if `T` has ever been added
```
omaskery, line 419:
```suggestion
    /// Returns a [`RelatedComponentsError`] if `R` is already a directly related component for `T`, or if `T` has ever been added
```

crates/bevy_ecs/src/world/mod.rs
```diff
@@ -378,40 +382,41 @@ impl World {
...
     /// on an entity before the registration.
     ///
     /// Indirect requirements through other components are allowed. In those cases, any existing requirements
```
omaskery, line 422:
```suggestion
    /// Indirect relations through other components are allowed. In those cases, any existing relations
```

crates/bevy_ecs/src/world/mod.rs
```diff
@@ -432,39 +437,40 @@ impl World {
     ///
     /// # let mut world = World::default();
     /// // Register B as required by A and C as required by B.
```
omaskery, line 439:
```suggestion
    /// // Register B as related to A and C as related to B.
```

crates/bevy_ecs/src/world/mod.rs
```diff
@@ -432,39 +437,40 @@ impl World {
...
+    /// When `T` is added to an entity, `R` and its own related components will also be added
     /// if `R` was not already provided. The given `constructor` will be used for the creation of `R`.
-    /// If a [`Default`] constructor is desired, use [`World::register_required_components`] instead.
+    /// If a [`Default`] constructor is desired, use [`World::register_related_components`] instead.
     ///
-    /// For the panicking version, see [`World::register_required_components_with`].
+    /// For the panicking version, see [`World::register_related_components_with`].
     ///
     /// Note that requirements must currently be registered before `T` is inserted into the world
     /// for the first time. This limitation may be fixed in the future.
```
omaskery, line 467:
```suggestion
    /// Note that relations must currently be registered before `T` is inserted into the world
    /// for the first time. This limitation may be fixed in the future.
```

crates/bevy_ecs/src/world/mod.rs
```diff
@@ -432,39 +437,40 @@ impl World {
...
     ///
-    /// [required component]: Component#required-components
+    /// [related component]: Component#related-components
     ///
     /// # Errors
     ///
-    /// Returns a [`RequiredComponentsError`] if `R` is already a directly required component for `T`, or if `T` has ever been added
+    /// Returns a [`RelatedComponentsError`] if `R` is already a directly required component for `T`, or if `T` has ever been added
```
omaskery, line 473:
```suggestion
    /// Returns a [`RelatedComponentsError`] if `R` is already a directly related component for `T`, or if `T` has ever been added
```

crates/bevy_ecs/src/world/mod.rs
```diff
@@ -432,39 +437,40 @@ impl World {
...
     /// on an entity before the registration.
     ///
     /// Indirect requirements through other components are allowed. In those cases, any existing requirements
```
omaskery, line 476:
```suggestion
    /// Indirect relations through other components are allowed. In those cases, any existing relations
```

crates/bevy_ecs/src/world/mod.rs
```diff
@@ -486,50 +492,55 @@ impl World {
...
 
         // TODO: Remove this panic and update archetype edges accordingly when required components are added
         if self.archetypes().component_index().contains_key(&requiree) {
-            return Err(RequiredComponentsError::ArchetypeExists(requiree));
+            return Err(RelatedComponentsError::ArchetypeExists(requiree));
         }
 
         let required = self.register_component::<R>();
 
         // SAFETY: We just created the `required` and `requiree` components.
```
omaskery, line 522: As previous: would be nice to come up with nouns for each end of the relatedness relationship.

crates/bevy_ecs/src/world/mod.rs
```diff
@@ -486,50 +492,55 @@ impl World {
...
+                constructor,
+                relatedness,
+            )
         }
     }
 
-    /// Retrieves the [required components](RequiredComponents) for the given component type, if it exists.
-    pub fn get_required_components<C: Component>(&self) -> Option<&RequiredComponents> {
+    /// Retrieves the [related components](RelatedComponents) for the given component type, if it exists.
+    pub fn get_required_components<C: Component>(&self) -> Option<&RelatedComponents> {
```
omaskery, line 534: Do there need to be new getters here to expose included and suggested components?

BD103:
Non-blocking, but it may be nice to add / update examples on how to do this. I'm curious how to use this myself, but don't know enough about ECS internals to read the actual source!

crates/bevy_ecs/macros/src/component.rs
```diff
@@ -75,100 +75,191 @@ pub fn derive_component(input: TokenStream) -> TokenStream {
...
+                #on_replace
+                #on_remove
+            }
+        }
+    })
+}
+
+// Because the macro crate cannot access the `bevy_ecs` crate, we need to create our own equivalent.
+// This should be kept in sync with the actual `bevy_ecs` crate!
+enum Relatedness {
```
cart, line 135: This PR needs enough work, and is "involved" enough that I'm worried about rushing it in this late in the game. This feels like the kind of thing that deserves time to be settled and tested. Additionally, this framing has been on our mind for 3 days. I think it _might_ be the right path forward, but the whole thing is still settling in my head. It feels like we're rushing into this.

Given how the required components conversation is going (in the release post PR), I think we can swing require-only for 0.15, _especially_ if we add the "break-glass "`#[remove_require_risky(Transform)]`". Given that this whole "other related components" effort is motivated by implementing the removal behavior for `include` to nail down the difference between the relationships, and given that this PR does not _yet_ implement that behavior, I think our limited resources / time / risk tolerance should be spent on implementing the removal functionality. That will already be reasonably invasive.

That then gives us a whole cycle to decide what the global picture should look like, ensure the implementation is sound, and actually make use of new tools like `include` ( I don't think we have the time in 0.15 to go through the engine and decide what deserves require vs include).

In the 0.15 release post, we could communicate that we plan on adding additional relationship tools for more flexibility.

alice-i-cecile: Yep, I'm on board with that decision. This is involved enough and I have enough to do that I'm not keen on pushing it in last minute.

crates/bevy_ecs/src/bundle.rs
```diff
@@ -401,15 +401,15 @@ impl BundleInfo {
         }
 
         let explicit_components_len = component_ids.len();
-        let mut required_components = RequiredComponents::default();
+        let mut related_components = RelatedComponents::default();
         for component_id in component_ids.iter().copied() {
             // SAFETY: caller has verified that all ids are valid
             let info = unsafe { components.get_info_unchecked(component_id) };
-            required_components.merge(info.required_components());
+            related_components.merge(info.related_components());
```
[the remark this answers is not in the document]
alice-i-cecile: Yeah, @Jondolf is of the same opinion WRT splitting out suggested components, and I think I'm largely convinced at this point.

alice-i-cecile:
This is more involved than I anticipated; bumping to 0.16 to avoid shipping bugs (and also to prioritize migration guide + release notes).

crates/bevy_ecs/src/component.rs
```diff
@@ -1001,58 +1011,61 @@ impl Components {
...
-    /// Returns a [`RequiredComponentsError`] if the `required` component is already a directly required component for the `requiree`.
+    /// Returns a [`RelatedComponentsError`] if the `required` component is already a directly required component for the `requiree`.
     ///
     /// Indirect requirements through other components are allowed. In those cases, the more specific
     /// registration will be used.
-    pub(crate) unsafe fn register_required_components<R: Component>(
+    pub(crate) unsafe fn register_related_components<R: Component>(
         &mut self,
         required: ComponentId,
         requiree: ComponentId,
```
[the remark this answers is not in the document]
krunchington: Suggestion: `source` and `target`, or some variation thereof (e.g. `relation_source` and `relation_target`)

hukasu:
did not read all of it, but does it include an idea of `these components are incompatible, and one of them is not going to work while the other is present`? (refer to #18188)

alice-i-cecile:
It does not. That's a form of #1481, which this PR covers some of, but not that element. I agree that that would be nice though!

PR #16270, opened 2024-11-06
Add a component to enable/disable interactions on a UI node

Exanc:
# Objective

Project goal: [Component type for enabling/disabling UI widget interactions (allow inactive buttons, etc…)](https://github.com/orgs/bevyengine/projects/10?pane=issue&itemId=7845269)

## Solution

The `Interactivity` component.
It stores a state describing whether an Interactive Node can be interacted with or not.

An interactive node can be described as an entity containing both the `Node` and `Interaction` components.

The `Interactivity` component can be defined with the following properties:
```rs
#[derive(Component, ...)]
#[require(Node, Interaction)]
pub enum Interactivity {
	Enabled,
	Disabled,
}
```

It is used by the `ui_focus_system` to determine whether an interactive node should have it's `Interaction` component updated when an interaction does occur. Furthermore, the assertion should be made that a "disabled" entity will always have it's `Interaction` component set to `Interaction::None`.

It can also be used by secondary systems to, for example, change the visual appearance of a UI node, or to decide whether or not to use `FocusPolicy::Block`.

## Testing

Not yet

## Migration Guide

Migrations are not necessary as the absence of the `Interactivity` component is interpreted as the component being enabled.

### DIFF
--- a/crates/bevy_ui/src/focus.rs
+++ b/crates/bevy_ui/src/focus.rs
@@ -34,6 +34,10 @@ use bevy_reflect::{ReflectDeserialize, ReflectSerialize};
 /// when [`ViewVisibility::get()`] is false.
 /// This ensures that hidden UI nodes are not interactable,
 /// and do not end up stuck in an active state if hidden at the wrong time.
+/// 
+/// If a UI node has both [`Interaction`] and [`Interactivity`] components,
+/// [`Interaction`] will always be [`Interaction::None`]
+/// when [`Interactivity::Disabled`] is used.
 ///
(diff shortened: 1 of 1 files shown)


Exanc:
There is still a missing, points: when an Interactivity component is changed, where should it be detected `ui_focus_system` or another on ?
Should we instantly change the interaction to None or let the current on finish ?

alice-i-cecile:
https://github.com/bevyengine/bevy/pull/15597 will be removing `ui_focus_system`. I would integrate with `bevy_picking` and try and use observers where possible.

Exanc:
@alice-i-cecile did not notice this change, there's a lot of info to keep up with 😅

Issue #16168, opened 2024-10-30
Expose `SystemMeta` for system exploration and configuration


MiniaczQ:
## What problem does this solve or what need does it fill?

Currently, there is no unified way to access `SystemMeta`.
We should define a set of methods in `System`, which would work for all: function, exclusive and compound systems.
This is especially needed for system param warning, which currently can be configured only for function systems.

## What solution would you like?

Introduce `System::system_metas(&self)` and `System::system_metas_mut(&mut self)` whicas would return iterator/vector/slice that we can work with.
@alice-i-cecile suggested a vector as a good starting point.

## What alternative(s) have you considered?

Reflection or separate structure for storing system configuration.

alice-i-cecile:
[This comment](https://github.com/bevyengine/bevy/pull/16275#issuecomment-2466813947) gave an excellent overview of the steps needed to solve this.

PR #16275, opened 2024-11-07
Implement `System::system_metas(&self)` and `System::system_metas_mut(&mut self)` for system exploration and configuration

Rajveer100:
Part of #16168

# Objective

Implement `System::system_metas(&self)` and `System::system_metas_mut(&mut self)` for system exploration and configuration.

## Solution

Introduce System::system_metas(&self) and System::system_metas_mut(&mut self) which would return iterator/vector/slice that we can work with.

### DIFF
--- a/crates/bevy_ecs/src/system/adapter_system.rs
+++ b/crates/bevy_ecs/src/system/adapter_system.rs
@@ -1,7 +1,10 @@
 use alloc::vec::Vec;
 use bevy_utils::prelude::DebugName;
 
-use super::{IntoSystem, ReadOnlySystem, RunSystemError, System, SystemParamValidationError};
+use super::{
+    IntoSystem, ReadOnlySystem, RunSystemError, System, SystemMeta, SystemMetaProvider,
+    SystemParamValidationError,
+};
 use crate::{
     schedule::InternedSystemSet,
     system::{input::SystemInput, SystemIn},
@@ -214,5 +217,30 @@ where
         run_system: impl FnOnce(SystemIn<'_, S>) -> Result<S::Out, RunSystemError>,
     ) -> Result<Self::Out, RunSystemError> {
         run_system(input).map(self)
+    }
+}
+
+impl<Func, S> SystemMetaProvider for AdapterSystem<Func, S>
+where
+    Func: Adapt<S>,
+    S: System + SystemMetaProvider,
+{
+    fn system_metas(&self) -> Vec<&SystemMeta> {
+        self.system.system_metas()
+    }
+
+    fn system_metas_mut(&mut self) -> Vec<&mut SystemMeta> {
+        self.system.system_metas_mut()
+    }
+
+    fn extend_with_system_metas<'a>(&'a self, extendable: &mut impl Extend<&'a SystemMeta>) {
+        self.system.extend_with_system_metas(extendable);
+    }
+
+    fn extend_with_system_metas_mut<'a>(
+        &'a mut self,
+        extendable_mut: &mut impl Extend<&'a mut SystemMeta>,
+    ) {
+        self.system.extend_with_system_metas_mut(extendable_mut);
     }
 }
--- a/crates/bevy_ecs/src/system/function_system.rs
+++ b/crates/bevy_ecs/src/system/function_system.rs
@@ -14,7 +14,7 @@ use crate::{
 
 use alloc::{borrow::Cow, vec, vec::Vec};
 use bevy_utils::prelude::DebugName;
-use core::marker::PhantomData;
+use core::{iter, marker::PhantomData};
 use variadics_please::all_tuples;
 
 #[cfg(feature = "trace")]
@@ -37,6 +37,24 @@ pub struct SystemMeta {
     pub(crate) system_span: Span,
     #[cfg(feature = "trace")]
     pub(crate) commands_span: Span,
+}
+
+/// Trait for retrieving system metas.
+pub trait SystemMetaProvider {
+    /// Returns immutable references of system metas as a vector.
+    fn system_metas(&self) -> Vec<&SystemMeta>;
+
+    /// Returns mutable references of system metas as a vector.
+    fn system_metas_mut(&mut self) -> Vec<&mut SystemMeta>;
+
+    /// Helper method for `system_metas`.
+    fn extend_with_system_metas<'a>(&'a self, extendable: &mut impl Extend<&'a SystemMeta>);
+
+    /// Helper method for `system_metas_mut`.
+    fn extend_with_system_metas_mut<'a>(
+        &'a mut self,
+        extendable_mut: &mut impl Extend<&'a mut SystemMeta>,
+    );
 }
 
 impl SystemMeta {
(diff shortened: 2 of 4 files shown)


Rajveer100:
@alice-i-cecile 
I am not quite sure if this is in the right direction.

alice-i-cecile:
@MiniaczQ, opinions?

crates/bevy_ecs/src/system/adapter_system.rs
```diff
@@ -208,6 +208,14 @@ where
     fn set_last_run(&mut self, last_run: crate::component::Tick) {
         self.system.set_last_run(last_run);
     }
+
+    fn system_metas(&self) -> &super::SystemMeta {
```
alice-i-cecile, line 212: This should be a `Vec<&SystemMeta>` :) Same thing below.

alice-i-cecile:
In all of the cases other than for piped systems, this method should return a vector with one element :)

MiniaczQ:
Yeah, I didn't want to tackle this, because it's actually pretty technical.
I'd say the actual scope of this issue can be split into 3 steps.
If you can get the first one down, the rest is relatively easy as follow-up PRs.

### Step 1
We definitely need a trait that has `system_metas(&self) -> Vec<&SystemMeta>` and `system_metas_mut(&mut self) -> Vec<&mut SystemMeta>`.

Now, to make this *actually* work, we need another set of methods: `extend_with_system_metas(&self, &mut Vec<&SystemMeta>)` and `extend_with_system_metas_mut(&self, &mut Vec<&mut SystemMeta>)`.
The original two methods will create a new `Vec` and fill it using those methods recursively.
Non-primitive systems like adapters and combinators will call those methods on their sub-systems.
I'm not sure if the borrow checker will freak out here or not with the `&mut`.

### Step 2
Then there is the matter of in-lining it.
This is not very appealing:
```rs
let mut my_system = IntoSystem::into_system(my_system);
my_system.system_metas_mut()[0].name = "abba".into();
app.add_systems(Update, my_system);
```
so we want a `map_system_metas` method that can do the same thing inline.
```rs
app.add_systems(Update, my_system.map_system_metas(|metas| { metas[0].name = "abba".into(); }));
```
no need for immutable variant.

### Step 3
Sadly this isn't the end for this issue, we also should expose `SystemMeta`s for schedule-configuration wrappers as well.
I believe those are `SystemConfig` and `SystemSetConfig`, but I'd need to look it up to be sure.
This would allow us to do `my_system.after(other_system).map_system_metas(...)`.

crates/bevy_ecs/src/system/adapter_system.rs
Rajveer100: Sure.

Rajveer100:
Thanks for the detailed overview :)

Rajveer100:
@MiniaczQ 
Let me know if the latest changes is the intended way. I think only the non-primitive systems remain:

- AdapterSystems
- CombinatorSystems
- PipedSystems

MiniaczQ:
Looks good!
Looking through CI there seem to be no issues with borrow checker so far, hopefully it stays that way 🤞

crates/bevy_ecs/src/system/function_system.rs
```diff
@@ -50,6 +50,21 @@ pub struct SystemMeta {
...
+/// Trait for retrieving system metas.
+pub trait SystemMetaProvider {
+    /// Returns immutable references of system metas as a vector.
+    fn system_metas(&self) -> Vec<&SystemMeta>;
+
+    /// Returns mutable references of system metas as a vector.
+    fn system_metas_mut(&mut self) -> Vec<&mut SystemMeta>;
+
+    /// Helper method for `system_metas`.
+    fn extend_with_system_metas<'a>(&'a self, vec: &mut Vec<&'a SystemMeta>);
```
hymm, line 62: it might be better for these to take an `impl Extend` instead of a Vec. https://doc.rust-lang.org/core/iter/trait.Extend.html

MiniaczQ: It's neat, but I don't think it's appropriate to implement it here

hymm: I was saying to change the function signature to `fn extend_with_system_metas<'a>(&'a self, extendable: impl Extend<&'a SystemMeta>)`, which would allow the methods to take a bunch of the other std collections and not be limited to just a `Vec`.

chescock: Another option is to take `impl FnMut(&'a SystemMeta)`.  That way, users that are just going to iterate over the result won't even have to allocate a `Vec`.

Rajveer100: Considering we want to be able to map items and do other sorts of behaviour for the metas in the follow ups, this could be more flexible.

MiniaczQ: Yup

chescock: Just to clarify, my point was that `FnOnce` is more general than `Extend` here.  You'd still offer the version that returns a `Vec`!  

```rust
fn for_each_system_meta<'a>(&'a self, f: impl FnOnce(&'a SystemMeta));

fn system_metas(&self) -> Vec<&SystemMeta> {
    let mut result = Vec::new();
    self.for_each_system_meta(|meta| result.push(meta));
    result
}
```
If someone wants to populate another type of collection, they can pass a closure that calls the relevant `push` method.  But if they were just going to do `for meta in system.system_metas() { ... }` then they can avoid an allocation by calling `system.for_each_system_meta(|meta| ...)`.

MiniaczQ: Can't be `FnOnce` since we use it many times, besides that makes sense, I voice no preference then

Rajveer100: Idea seems reasonable.

Rajveer100:
@alice-i-cecile @MiniaczQ 
Got back to this after a long time, review needed :)

CI looks good.

chescock:
Welcome back!  

I'm worried the use case for this change may have disappeared in the meantime, though.  The linked issue says this is "needed for system param warning, which currently can be configured only for function systems", but system param validation has been reworked since then, with `ParamWarnPolicy` removed in #18454.  

Following that and #19477, all `SystemMeta` has is `name`, `last_run`, and `flags`.  They can all be read using the `System` trait, `last_run` can be set, and the flags aren't safe to modify, so the only API that's really missing is `set_name()`.  

Do we really still need this API?

Rajveer100:
At a first glance, looks like this may not be required, but do we have the same way of direct access this API provides?

It's more like a user's choice which allows one to get a deeper access. I can't think of a use case that this especially provides at the moment, but I feel if we introduce this and let users sort of play with it in the next release we can get a better idea.

If not, we could always remove it?

Rajveer100:
@chescock
Any final thoughts?

chescock:
> @chescock Any final thoughts?

Whoops, sorry, missed this message for a while!  

> At a first glance, looks like this may not be required, but do we have the same way of direct access this API provides?

I think my argument here is that we do!  Looking at the fields on `SystemMeta`, nearly all of them are exposed on the `System` trait already.  The only exception is being able to set the name ... but combinator systems store their name separately from the inner `SystemMeta`, so setting the name on those won't change the system name anyway.  

But I'm also just a random reviewer and not a SME or maintainer :).  If you can get some more consensus that this is a good direction then I'm happy to give it a proper review!  I just don't want to waste your time nitpicking the implementation if we aren't going to do this anyway.

Rajveer100:
Cool, thanks for the feedback, maybe in the future if there's a critical discussion in this area we can come back to it.

Looking into other issues :)

PR #16280, opened 2024-11-07
Rename `AnimationGraph`

### LINKED ISSUE #15604: Rename `AnimationGraph` to something like `BlendGraph` or `BlendTree`
From time to time, someone gets confused about what the `AnimationGraph` actually does. Its name is probably too easily construed as something that might resemble a state machine, when all it does is blend animation clips. A name that reflects that is probably better-suited.

BenjaminBrienen:
# Objective

Fixes #15604

## Solution

Replace "AnimationGraph" with "BlendGraph". Rename "animatable" to "blendable" accordingly.

TODO: Update the comments. I see a lot of "`// animation graph`". I'll do this once this is blessed and the scope has been determined.

## Testing

Simple refactor can be evaluated by CI

## Migration Guide

I'll create this once this has been blessed.

### DIFF
--- a/crates/bevy_animation/src/animatable.rs
+++ b/crates/bevy_animation/src/animatable.rs
@@ -6,7 +6,7 @@ use bevy_math::*;
 use bevy_reflect::Reflect;
 use bevy_transform::prelude::Transform;
 
-/// An individual input for [`Animatable::blend`].
+/// An individual input for [`Blendable::blend`].
 pub struct BlendInput<T> {
     /// The individual item's weight. This may not be bound to the range `[0.0, 1.0]`.
     pub weight: f32,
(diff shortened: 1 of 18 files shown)


BenjaminBrienen:
The scope of this task needs to be precisely defined. What other types should be renamed for clarity? Should the term "blend" be used in documentation as well?

hukasu:
Shouldn't the name of the file also be renamed from `animatable.rs` to `blendable.rs`?

And what about the extension of the `AnimationGraphAssetLoader` from `.animgraph` to `.blendgraph`?

BenjaminBrienen:
Good ideas, thanks.

pcwalton (approved):
I don't have any problem with this.

mweatherley (changes requested):
I'm fine with this, but there are still a bunch of places in `animatable.rs` and `animation_curves.rs` where "animatable" should be changed to "blendable" (including the name of the former module, probably).

alice-i-cecile:
Marking as Adopt-Me: I still think this is a good change (and there's support from animation experts), but there's more work to do here and a boatload of merge conflicts.

ChristopherBiscardi:
adoption attempt in https://github.com/bevyengine/bevy/pull/22782 so closing this one

PR #16283, opened 2024-11-07
Return Error instead of panicking when reading invalid dds file

### LINKED ISSUE #15928: Reading malformed DDS file is a panic instead of an error result
## Bevy version

0.13.1

## What you did

Loading a DDS file for rendering that happens to be malformed.

## What went wrong

Image

https://github.com/bevyengine/bevy/blob/de08fb2afa5e8da1d910ed6be424649b6fbe9f0e/crates/bevy_image/src/dds.rs#L19

As use cases might involve loading DDS files specified by the user, this behavior makes it hard to properly handle an incorrect input from the caller's side. 
Or if I'm not using this the correct way?

BenjaminBrienen:
# Objective

Fixes #15928

## Solution

return Error instead of panic

## Testing

I don't know if we need to add a test for this. It is pretty straightforward.

### DIFF
--- a/crates/bevy_image/src/dds.rs
+++ b/crates/bevy_image/src/dds.rs
@@ -1,3 +1,6 @@
+//! DirectDraw Surface functionality.
+//! https://en.wikipedia.org/wiki/DirectDraw_Surface
+
 #[cfg(debug_assertions)]
 use bevy_utils::warn_once;
 use ddsfile::{Caps2, D3DFormat, Dds, DxgiFormat};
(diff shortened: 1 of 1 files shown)


BenjaminBrienen:
@zbx1425 does this meet your needs?

JMS55 (approved):

richchurcher (approved):

zbx1425:
tysm!
I should've PRed it myself, but didn't get around to do this

mockersf (approved):

PR #16284, opened 2024-11-07
Rename `StateScoped` to `DespawnOnStateExit` and add `DespawnOnStateEnter`

### LINKED ISSUE #15849: Rename `StateScoped` to `StateExitDespawn`, add the inverse variant `StateEntryDespawn`

The current naming of the `StateScoped` component is slightly misleading: it could be interpreted as "this entity will not exist outside of this state". Whereas in fact it merely despawns on exit of the given state, it doesn't care if you spawn an entity _outside_ of the given state:

```
for (entity, binding) in &query {
    if binding.0 == *exited {
        #[cfg(feature = "bevy_hierarchy")]
        commands.entity(entity).despawn_recursive();
        #[cfg(not(feature = "bevy_hierarchy"))]
    }
}
```

To me, this functionality is fine, I can't think of a situation where you would want to spawn something in a state it doesn't belong, only for it to be deleted instantly.

There is also no available _inverse_ of this component: despawn an entity on _entry_ to a state, which I recently found a need for, and have written a very similar mechanism to achieve the same thing.

- Rename `StateScoped` to `StateExitDespawn`, or maybe a cleaner looking name than that.
- Add the inverse version: `StateEntryDespawn`, having it operate during state transition, before `OnEnter`.

Here's a snippet of what I produced for my own game:

```
impl AppExtLifetime for SubApp {
    fn enable_state_entry_despawn<S: States>(&mut self) -> &mut Self {
        self.add_systems(
            StateTransition,
            state_entry_despawn::<S>
                .in_set(TransitionSchedules::<S>::default())
                .run_if(on_event::<StateTransitionEvent<S>>()),
        );
        self
    }
}
```

No real alternatives.

## Additional context

State transitions are such a powerful and flexible way to orchestrate game entity lifecyles and the flow of the game itself, it would be nice to see more features in this area.

BenjaminBrienen:
# Objective

Fixes #15849

## Solution

Rename `StateScoped` to `DespawnOnStateExit` (and related function).
Add a copy of the type named `DespawnOnStateEnter` (and modify related function to use `.entered`)
Add the system.

## Showcase

https://github.com/user-attachments/assets/6c1873dd-9497-49c3-8aed-8bda68ba3b2c

## Testing

Tested manually with an example

## Migration Guide
The `StateScoped` component has been renamed to `DespawnOnStateExit`. A corresponding `DespawnOnStateEnter` component has also been added.


BenjaminBrienen:
@Jaso333 does this meet your needs?

Jaso333:
> @Jaso333 does this meet your needs?

absolutely!

BenjaminBrienen:
> > @Jaso333 does this meet your needs?
> 
> absolutely! thanks!

Could review the PR, please? :)

AxiomaticSemantics:
LGTM, haven't tested but with these changes the functionally mirrors my own automated cleanup solution as far as state scoping goes.

AxiomaticSemantics (approved):
The code looks fine and is a means to an end. I think a larger discussion about world object management needs to happen tho.

benfrankel:
I haven't looked at the code; but I approve of this direction and the names chosen. In `pyri_state` I renamed `StateScope` to `DespawnOnExit` already. I like the `StateExit` disambiguation, though, especially if more variants are added. If we get entity disabling _and_ state disabling eventually, `DisableOnStateDisable` makes more sense than `DisableOnDisable` :P

mockersf:
you should also rename `App::enable_state_scoped_entities` to something else if we want to get ride of "state scoped"

BenjaminBrienen:
> you should also rename `App::enable_state_scoped_entities` to something else if we want to get ride of "state scoped" 

I think "state scoped" is a great name for the feature itself.

alice-i-cecile (commented):
I think this is clearer, and the functionality is useful. If you add tests and a migration guide you have my approval. You probably want to do something similar for events too.

I've removed X-Controversial, as this isn't a particularly high-impact change, and the various states devs are broadly in favor of this style of mechanism.

BenjaminBrienen:
That should work for the migration guide, I think. I'll add a test tonight.

MiniaczQ (approved):
Looks good, one suggestion for followup if we want to have even more variants

BenjaminBrienen:
I didn't have time tonight. Definitely tomorrow or the next day.

BenjaminBrienen:
@alice-i-cecile I added an example. I don't see an existing unit test for state_scoped, so I'm not sure how that would work.

Cargo.toml
```diff
@@ -2118,6 +2118,17 @@ description = "Pipe the output of one system into a second, allowing you to hand
...
 wasm = false
 
+[[example]]
+name = "state_scoped"
+path = "examples/ecs/state_scoped.rs"
+doc-scrape-examples = true
+
+[package.metadata.example.state_scoped]
+name = "State Scoped"
+description = "Show how to use state scoped systems"
```
alice-i-cecile, line 2128: This needs a better description. It should explain what the feature is to users who don't know the jargon :)

examples/ecs/state_scoped.rs
```diff
@@ -0,0 +1,123 @@
+//! Shows how to use state scoped systems.
```
alice-i-cecile, line 1: If we're keeping this example, the module docs here need more color on what this feature is and why it's useful.

I'd probably prefer to cut it, in favor of the already solid doc tests. Just make sure that the state scoped stuff is linked in the module docs of `bevy_state`.

alice-i-cecile (commented):
I like the rename and new feature, but I think we can clean up the documentation story a bit :)

mgi388:
I've adopted this in #18818, so have nominated to close this.

PR #16286, opened 2024-11-08
Make visibility range (HLOD) dithering work when prepasses are enabled.

pcwalton:
Currently, the prepass has no support for visibility ranges, so artifacts appear when using dithering visibility ranges in conjunction with a prepass. This patch fixes that problem.

Note that this patch changes the prepass to use sparse bind group indices instead of sequential ones. I figured this is cleaner, because it allows for greater sharing of WGSL code between the forward pipeline and the prepass pipeline.

The `visibility_range` example has been updated to allow the prepass to be toggled on and off.


tbillington (approved):
Not a regular graphics code reviewer, but it looks okay :)

BenjaminBrienen:
https://github.com/user-attachments/assets/d6dcbcf6-2e03-4318-8524-e59fd2b4f532

There is a very noticeable flicker when the model changes with the prepass enabled.

IceSentry (approved):
I did not run the code locally, but the changes look good to me. I agree that using indices for bind group is a good change here.

pcwalton:
Turns out that the bug that @BenjaminBrienen noticed is actually a regression from a much earlier PR. I submitted a fix at https://github.com/bevyengine/bevy/pull/16468.

BenjaminBrienen (approved):

PR #16287, opened 2024-11-08
AudioPlayer::new()

cart:
# Objective

`AudioPlayer::<AudioSource>(assets.load("audio.mp3"))` is awkward and complicated to type because the `AudioSource` generic type cannot be elided. This is especially annoying because `AudioSource` is used in the majority of cases. Most users don't need to think about it.

## Solution

Add an `AudioPlayer::new()` function that is hard-coded to `AudioSource`, allowing `AudioPlayer::new(assets.load("audio.mp3"))`. Prefer using that in the relevant places.

### DIFF
--- a/crates/bevy_audio/src/audio.rs
+++ b/crates/bevy_audio/src/audio.rs
@@ -264,6 +264,17 @@ where
     }
 }
 
+impl AudioPlayer<AudioSource> {
+    /// Creates a new [`AudioPlayer`] with the given [`Handle<AudioSource>`].
+    ///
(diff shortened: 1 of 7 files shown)


rparrett (approved):
Looks good to me

bushrat011899 (approved):
It's not ideal that we have to work around Rust's implied typing here, but I do think this improves the default case without much effort to maintain. If a better design presents itself migration to it from here shouldn't be hard at all.

TurtIeSocks (approved):

PR #16288, opened 2024-11-08
Fix panic in UiSurface if Node is removed and re-added

UkoeHB:
# Objective

- Fix bug where `UiSurface::set_camera_children` (and `UiSurface::update_children` sometimes) will panic if you remove and add a `Node` component in a single tick. This is more likely to happen now because of `remove_with_requires`.

## Solution

- Filter out entities with `Node` when cleaning up entities from `RemovedComponents<Node>`.

## Testing

- Not tested (rust compiler refused to cooperate when I tried to patch this into my project), correct by inspection.

### DIFF
--- a/crates/bevy_ui/src/layout/mod.rs
+++ b/crates/bevy_ui/src/layout/mod.rs
@@ -278,7 +278,12 @@ with UI components as a child of an entity without UI components, your UI layout
 
     let text_buffers = &mut buffer_query;
     // clean up removed nodes after syncing children to avoid potential panic (invalid SlotMap key used)
-    ui_surface.remove_entities(removed_components.removed_nodes.read());
+    ui_surface.remove_entities(
+        removed_components
+            .removed_nodes
+            .read()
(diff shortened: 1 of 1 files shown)


mockersf (approved):

richchurcher (approved):

BenjaminBrienen:
This should have a test to prove that it works and prevent regressions. Can you add one?

crates/bevy_ui/src/layout/mod.rs
```diff
@@ -771,6 +776,38 @@ mod tests {
         assert!(ui_surface.entity_to_taffy.is_empty());
     }
 
+    /// bugfix test, see https://github.com/bevyengine/bevy/pull/16288
```
BenjaminBrienen, line 779:
```
    /// bugfix test, see [#16288](https://github.com/bevyengine/bevy/pull/16288)
```

BenjaminBrienen (changes requested):

BenjaminBrienen (approved):

Issue #10395, opened 2023-11-05
Tonemapping example image loader not working


DGriffin91:
## Bevy version 32a5c7db3535250c43738e3da5877497b1274960

Previously, in the tonemapping example an exr or hdr image file could be dragged onto the window to view the image with the selected tonemapping settings. This no longer works.

The issue was introduced at 35073cf7aa9eaf748e167c103921a6764fcedfb1 with #9885

```
thread 'Compute Task Pool (21)' panicked at 'called `Result::unwrap()` on an `Err` value: InvalidSourceSyntax', crates\bevy_asset\src\path.rs:115:70
```

cart:
General answer is "its complicated".
The AssetSource / AssetReader / AssetWriter abstraction is intentionally scoped to specific folders to allow us to make assumptions about them (ex: this folder is processed, this source might not come from a filesystem, etc). AssetPaths are also intended to be platform (and source-type) agnostic. But absolute paths like `C:\some\path` are a windows-specific convention.

Of course, we should support loading assets in these cases, but I think how we do that needs to change. I think it should probably be a combination of two things:

1. Enable mounting new asset sources at runtime (instead of needing to anticipate at startup)
2. Create an abstraction that can convert from absolute (sometimes OS-specific) file paths to equivalent platform AssetPaths. In the case of a windows-specific path like the one above, this would involve mounting a new `C` asset source and converting windows-specific paths to platform-agnostic AssetPaths. (ex: `C:\some\asset.png` would become `C://some/asset.png`.

Some discussion here on discord:
https://discord.com/channels/691052431525675048/1164297864441249823

cart:
Note that we are planning on being more principled (and platform-agnostic) about AssetPaths in the future. Ex: you should use `/` instead of `\`, windows-style "drives" would not be allowed (ex:  `C:\`), etc.

DGriffin91:
I think this makes sense as long as there's a reasonable way for an end user to specify a file path in a typical OS native fashion. There should be a reasonably straightforward way to make a bevy app that allows the end user to use `C:\`, or `\` instead of `/`

Some example applications would be: digital content creation, scientific visualization, anything with a OS file dialog.

cart:
[...]

PR #16294, opened 2024-11-08
Support loading assets from arbitrary filesystem paths by specifying an absolute path as the asset path

inodentry:
# Objective / Problem

There is no way to access asset files outside of the designated asset directory. Bevy's `FileReader`/`FileWriter` always interpret paths relative to the configured root directory (based on either the executable's path, or env var).

Being able to load (and save) assets anywhere in the filesystem is important for many applications, most notably editors.

The goal of this PR is to implement the simplest possible solution to this problem.

## Solution

If an absolute path is provided when loading/saving an asset, it is used as-is.

If a relative path is provided, everything works exactly the same as before. Technically this isn't a breaking change, as
previously only relative paths were accepted by Bevy.

## Testing

I ran the various bevy examples to make sure they can still load their assets (from the `assets` folder, as before).

I then modified them (locally, not committed, ofc) by changing the paths in the source code to various absolute paths
on my computer.

I can confirm Bevy can now successfully load random images and GLTFs on my PC. :)

I have not tested asset meta files.

I have not tested the "sync" loader (when the `multi_threaded` cargo feature is disabled).

I have not tested asset writing / saving.

I have not tested the file watcher / hot reloading.

Probably also worth testing loading a complex GLTF file that references external files (for textures, etc) using an absolute path, to make sure asset dependencies work when using the new absolute paths.

Nothing special. Just try to load assets from an absolute path.

```rust
asset_server.load("/home/me/Downloads/random_pic.png");
```

Windows:

```rust
asset_server.load("C:\\Users\\Me\\Downloads\\random_pic.png");
```

I have not tested on Windows yet. I will be able to do that soon. Will edit the PR description to remove this note when I do.

The changes in this PR are only relevant to desktop platforms (the file asset io is not used on mobile and web).

---

Migration guide not necessary.

This PR, while straightforward, does enable cool new use cases and is probably deserving of a release note.

### DIFF
--- a/crates/bevy_asset/src/io/file/file_asset.rs
+++ b/crates/bevy_asset/src/io/file/file_asset.rs
@@ -34,10 +34,16 @@ impl Reader for File {}
 
 impl AssetReader for FileAssetReader {
     async fn read<'a>(&'a self, path: &'a Path) -> Result<impl Reader + 'a, AssetReaderError> {
-        let full_path = self.root_path.join(path);
+        let path_buf;
+        let full_path = if path.is_absolute() {
+            path
+        } else {
+            path_buf = self.root_path.join(path);
+            &path_buf
+        };
         File::open(&full_path).await.map_err(|e| {
             if e.kind() == std::io::ErrorKind::NotFound {
-                AssetReaderError::NotFound(full_path)
+                AssetReaderError::NotFound(full_path.to_path_buf())
             } else {
                 e.into()
             }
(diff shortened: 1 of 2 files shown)


mockersf:
I don't know if @cart changed opinion since https://github.com/bevyengine/bevy/issues/10395#issuecomment-1793837440 or https://discord.com/channels/691052431525675048/692572690833473578/1235753754222133279

andriyDev:
Is it possible to instead just make this an `AssetSource`? For example, it could look like `asset_server.load("absolute://C:\\Users\\Me\\Downloads\\random_pic.png");` This would make enabling "absolute asset paths" a matter of adding a `AbsoluteAssetSource` - we don't need to modify any part of the asset system, and users can choose whether they want to support absolute assets at all.

The cost as we can see is that the syntax is kinda ugly, especially on Windows.

copygirl:
@andriyDev This reminds me of the `file:` URI scheme which from the sounds of it only supports absolute paths.

mardzie:
Wouldn't it make sense to extract all these ifs into a separate function?

```rust
let full_path = if path.is_absolute() {
    path
} else {
    path_buf = self.root_path.join(path);
    &path_buf
};
```

ChristopherBiscardi:
as per comments above, this can be done with custom asset sources. So I'm going to close this one

PR #16299, opened 2024-11-08
Improve profiling instructions

JMS55:
# Objective

- Make it easier to understand how to profile things.
- Talk about CPU vs GPU work for graphics.

## Solution

- Add a section on GPU profiling and CPU vs GPU work.
- Rearrange some sections so Tracy is the first backend mentioned.

## Issues
I did this as a very quick fix to clear some things up, but the overall guide still flows poorly and has too much extraneous info distracting from the use case of "I just want to figure out why my app is slow", where the advice should be "use tracy, and if GPU bottlenecked, do this". Someone should do a full rewrite at some point.

I chose to omit talking about RenderDiagnosticsPlugin, but it's definitely worth a mention as a way to easily check GPU + GPU time for graphics work, although it's not hooked up in a lot of the engine, iirc only shadows and the main passes. Again someone else should write about it in the future.

Similarly it might've been useful to have a section describing how to use NSight/RGP/IGA/Xcode for GPU profiling, as they're far from intuitive tools.

### DIFF
--- a/docs/profiling.md
+++ b/docs/profiling.md
@@ -126,6 +126,7 @@ After closing your app, an interactive `svg` file will be produced:
 If CPU profiling has shown that GPU work is the bottleneck, it's time to profile the GPU.
 
 For profiling GPU work, you should use the tool corresponding to your GPU's vendor:
+
 - NVIDIA - [Nsight Graphics](https://developer.nvidia.com/nsight-graphics)
 - AMD - [Radeon GPU Profiler](https://gpuopen.com/rgp)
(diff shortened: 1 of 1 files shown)


BenjaminBrienen:
It should only check siblings for that markdown lint.

BenjaminBrienen:
In https://github.com/bevyengine/bevy/blob/main/.github%2Flinters%2F.markdown-lint.yml add the config in my branch.

BenjaminBrienen (approved):

JMS55:
Oh, I did forget to mention VSync and how a ~16ms frame might not mean slow 😅 . Add that to the TODO list for someone to pick up.

PR #16301, opened 2024-11-08
Expose Pipeline Compilation Zero Initialize Workgroup Memory Option

atlv24:
# Objective

- wgpu 0.20 made workgroup vars stop being zero-init by default. this broke some applications (cough foresight cough) and now we workaround it. wgpu exposes a compilation option that zero initializes workgroup memory by default, but bevy does not expose it.

## Solution

- expose the compilation option wgpu gives us

## Testing

- ran examples: 3d_scene, compute_shader_game_of_life, gpu_readback, lines, specialized_mesh_pipeline. they all work
- confirmed fix for our own problems

---

</details>

## Migration Guide

- add `zero_initialize_workgroup_memory: false,` to `ComputePipelineDescriptor` or `RenderPipelineDescriptor` structs to preserve 0.14 functionality, add `zero_initialize_workgroup_memory: true,` to restore bevy 0.13 functionality.

### DIFF
--- a/crates/bevy_core_pipeline/src/auto_exposure/pipeline.rs
+++ b/crates/bevy_core_pipeline/src/auto_exposure/pipeline.rs
@@ -89,6 +89,7 @@ impl SpecializedComputePipeline for AutoExposurePipeline {
                 AutoExposurePass::Average => "compute_average".into(),
             },
             push_constant_ranges: vec![],
+            zero_initialize_workgroup_memory: false,
         }
     }
 }
--- a/crates/bevy_core_pipeline/src/blit/mod.rs
+++ b/crates/bevy_core_pipeline/src/blit/mod.rs
@@ -98,6 +98,7 @@ impl SpecializedRenderPipeline for BlitPipeline {
                 ..Default::default()
             },
             push_constant_ranges: Vec::new(),
+            zero_initialize_workgroup_memory: false,
         }
     }
 }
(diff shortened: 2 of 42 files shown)


aevyrie (approved):
Can this be added to the default impl of `RenderPipelineDescriptor` so this doesn't break the next time someone adds it and decides to set it to false? Maybe a constructor even?

IceSentry (approved):

IceSentry:
RenderPipelineDescritptor can't impl Default because of the VertexState. But the compute one could I guess.

atlv24:
not sure if it makes sense to add a blanket Default impl, but per-field defaults would be nice. this is something that rust kinda lacks tbh

atlv24:
i kinda like it
```rs
#[derive(Default)]
pub enum WorkgroupMemoryInitialization {
    #[default]
    Zeroed,
    Uninitialized,
}

impl From<WorkgroupMemoryInitialization> for bool {
    fn from(init: WorkgroupMemoryInitialization) -> bool {
        match init {
            WorkgroupMemoryInitialization::Zeroed => true,
            WorkgroupMemoryInitialization::Uninitialized => false,
        }
    }
}
```

aevyrie:
> i kinda like it

Yeah, this feels like a much more idiomatic solution. You can also add some docs to the type!

atlv24:
i'll do it monday

PR #16303, opened 2024-11-09
For Aevyrie :)

atlv24:
# Objective

- Followup to #16301

### DIFF
--- a/crates/bevy_core_pipeline/src/auto_exposure/pipeline.rs
+++ b/crates/bevy_core_pipeline/src/auto_exposure/pipeline.rs
@@ -89,7 +89,7 @@ impl SpecializedComputePipeline for AutoExposurePipeline {
                 AutoExposurePass::Average => "compute_average".into(),
             },
             push_constant_ranges: vec![],
-            zero_initialize_workgroup_memory: false,
+            workgroup_memory_initialization: MemoryInitialization::Uninitialized,
         }
     }
 }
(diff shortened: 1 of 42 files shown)


BenjaminBrienen:
You're going to have to change the PR title before it gets merged, btw.

aevyrie (commented):
Just to check, you *don't* want to zero the memory?

IceSentry:
> Just to check, you don't want to zero the memory?

That's the current default in bevy 0.14 and until us nobody had an issue and there's probably some overhead to zeroing things all the time so it's fine to default to not zeroing I think.

JMS55:
If I remember the wgpu discussion at the time, it has _significant_ overhead. We shouldn't zero by default imo.

atlv24:
@BenjaminBrienen i made the title more descriptive, hope it helps

atlv24:
> If I remember the wgpu discussion at the time, it has _significant_ overhead. We shouldn't zero by default imo.

I added some notes on performance, and all usages in bevy are set to Uninitialized. I think that should be enough, but maybe im missing something

aevyrie:
> > If I remember the wgpu discussion at the time, it has _significant_ overhead. We shouldn't zero by default imo.
> 
> I added some notes on performance, and all usages in bevy are set to Uninitialized. I think that should be enough, but maybe im missing something

The implementation of default has it set to `Zeroed` when it should be `Uninitialized`

atlv24:
I dont see what the point of this PR is if we change the default to Zeroed. we might as well keep the bool as it defaults to false already

aevyrie:
> I dont see what the point of this PR is if we change the default to Zeroed. we might as well keep the bool as it defaults to false already

Isn't that what this PR is doing? https://github.com/bevyengine/bevy/pull/16303/files#diff-9456795cbf69709846adb7ef9633788a1d50b3d463ee632392b3ff4747c4d395R162-R163

atlv24:
Ope i misspoke. I meant changing the default to Uninitialized. The current bool default in main is uninitialized, and this pr changes it to Zeroed

PR #16306, opened 2024-11-09
Fix markdownlint warnings and improve consistency with config

BenjaminBrienen:
# Objective

Markdown files should have a consistent style which promotes readability for development purposes.

## Solution

I removed some lint disables because they don't trigger anyway.
I undisabled the line-length lint and fixed the resulting warnings.
I used is reflowing the source code of the paragraphs without changing the rendered output.
The main other lint that triggered is inconsistency in the `ul` style.


BenjaminBrienen:
I'm expecting feedback on the `.editorconfig` and `.markdownlint.jsonc` files being checked in. I'm also open to criticism about stylistic choices.

alice-i-cecile:
Can you split out the line length lint into its own PR? Everything else looks pretty uncontroversial and this will be much easier to get merged that way.

BenjaminBrienen:
> Can you split out the line length lint into its own PR? Everything else looks pretty uncontroversial and this will be much easier to get merged that way.

Done: #16308

PR #16311, opened 2024-11-09
Add transform lerp

Sigma-dev:
# Objective

I often find myself needing to lerp between Transforms (ex: smooth camera transitions) and I believe that since each of a Transfom's attributes (translation, scale, rotation) are also lerp-able, it makes sense to implement it for the whole. (I was personally surprised when I didn't find it)

## Solution

Adding a lerp method that lerps all the attributes individually, resulting in the whole transform being interpolated.

## Testing

I didn't feel the need to add tests as the method is directly built upon reliable and well tested parts of the engine, though I would have no problem adding tests if needed.


Jondolf:
For interpolating rotation, you typically don't want linear interpolation (lerp) but spherical linear interpolation (slerp). Lerp is more efficient, but doesn't have a constant velocity, and looks worse especially when the two rotations are far apart. Slerp produces a smoother interpolation with uniform angular velocity.

atornity:
[`Animatable`](https://docs.rs/bevy/latest/bevy/animation/animatable/trait.Animatable.html) is already implemented for `Transform` btw. It let's you use `interpolate` the same way you would `lerp` except it uses `slerp` for the rotation.

BenjaminBrienen:
I think we should close this and keep the functionality defined in Animatable

Sigma-dev:
I think animatable is not very easy to find for users if they don't already know about it.
I am not sure how one would use it.

A built-in function has little downside for  big upsides imo.

BenjaminBrienen:
Maybe we can just add thr function, but it just uses Animatable internally.

mockersf:
I don't think it should be called `lerp`, as it's not doing `lerp` on all fields of the transform.

BenjaminBrienen:
Maybe `interpolate` or `blend`, but again, why not just use the fact that it implements `Animatable`?

mockersf:
at the moment, you have to call `Transform::interpolate(transform_a, transform_b, t)` instead of `transform_a.interpolate(transform_b, t)` which is not very good for discoverability

BenjaminBrienen:
I think the first one is better anyway. That's how it is done in Unity, which is where I come from. If the second one has clear advantages, I'm open-minded.

mockersf:
the issue with the first one is that you can't discover it in your IDE when you're trying to autocomplete on a `Transform` variable, so you have to know it exists first

BenjaminBrienen:
Can we make that possible without widespread code duplication by adding a second version of the each trait method which then takes `&self`?

Sigma-dev:
I think the confusing part  is that all other types that include lerping all have them as a &self method, so there are expectations built-up from the developer pov (talking from personal experience)

atornity:
is there any reason for `Animatable::interpolate` to not take `&self`? Changing that seems like the best solution imo.

atornity:
> "interpolate from x to y (current time is z)" reads better than "x interpolate to y (current time is z)" (opinion of course)

If we were to use `&self` as the first argument then `Transform::interpolate(a, b, t)` and `a.interpolate(b, t)` would both work fine.

> Is the advice to take &self whenever possible on trait methods?

not sure if it's advised, but traits like [`Add`](https://doc.rust-lang.org/std/ops/trait.Add.html) and glams [`FloatExt`](https://docs.rs/glam/latest/glam/trait.FloatExt.html) (which includes `lerp`) do take `self`. 

I personally don't see a reason to not using `&self` except: "we don't really want users to use this trait".

BenjaminBrienen:
Ok, I'm convinced. Thanks for the responses!

Sigma-dev:
I'm not sure what needs to be changed, coud anyone help ?

BenjaminBrienen:
> I'm not sure what needs to be changed, coud anyone help ?

I think the consensus is to change `Animatable`'s methods to take `&self` as the first parameter instead. It should work the same way.

mockersf:
I don't remember if there was a reason to make the `Animatable` trait that way, it would be nice to have a confirmation that it doesn't block anything from @james7132

AntoninD-Galadrim:
I might be wrong, but since only the trait Animatable is pub and the methods in it are not, I don't think the end user will be able to use them, so maybe I need to put these methods as pub ?

Sigma-dev:
I might be wrong, but since only the Animatable trait is pub and not it's methods, then this make the interpolate method inacessible to the end user.

So I might need to change it if it is true.

BenjaminBrienen:
Yeah, trait methods are always `pub`. You can test that by creating an `animate_transform` example in the `./examples` folder.

PR #16314, opened 2024-11-09
Procedural atmospheric scattering

ecoskey:
Implement procedural atmospheric scattering from [Sebastien Hillaire's 2020 paper](https://sebh.github.io/publications/egsr2020.pdf). This approach should scale well even down to mobile hardware, and is physically accurate.

## Co-author: @mate-h 

He helped massively with getting this over the finish line, ensuring everything was physically correct, correcting several places where I had misunderstood or misapplied the paper, and improving the performance in several places as well. Thanks!

## Credits

@aevyrie: helped find numerous bugs and improve the example to best show off this feature :)

Built off of @mtsr's original branch, which handled the transmittance lut (arguably the most important part)

## Showcase: 

sunset
twilight

## For followup

- Integrate with pcwalton's volumetrics code
- refactor/reorganize for better integration with other effects
- have atmosphere transmittance affect directional lights
- add support for generating skybox/environment map


pcwalton:
Wow, this is very impressive! I'll review.

aevyrie:
Considering the `Atmosphere` takes parameters about the inner and outer radius of the atmosphere, maybe it would make sense if that component was an another entity? That would allow you to compute the camera's view within (or outside) the atmosphere based on the position of the atmosphere and the camera using their `GlobalTransform`s. Excited to see this implemented!

pcwalton:
Would you like reviews/feedback on this at this stage or is it too early?

crates/bevy_pbr/src/atmosphere/functions.wgsl
```diff
@@ -0,0 +1,208 @@
...
+
+fn sample_multiscattering_lut(r: f32, mu: f32) -> vec3<f32> {
+    let uv = multiscattering_lut_r_mu_to_uv(r, mu);
+    return textureSampleLevel(multiscattering_lut, multiscattering_lut_sampler, uv, 0.0).rgb;
+}
+
+fn sample_sky_view_lut(ray_dir: vec3<f32>) -> vec3<f32> {
+    let lat_long = ray_dir_to_lat_long(ray_dir);
+    let uv = sky_view_lut_lat_long_to_uv(lat_long.x, lat_long.y);
+    return textureSampleLevel(sky_view_lut, sky_view_lut_sampler, uv, 0.0).rgb;
```
pcwalton, line 63: Inverse trigonometric functions are *really* slow on GPU. Can you use a cubemap LUT instead? Directional data is basically why cubemap textures exist.

crates/bevy_pbr/src/atmosphere/functions.wgsl
```diff
@@ -0,0 +1,208 @@
...
+fn get_local_up(view_r: f32, displacement: vec3<f32>) -> vec3<f32> {
+    return normalize(vec3(0.0, view_r, 0.0) + displacement);
+}
+
+fn get_local_r(view_r: f32, view_mu: f32, dist: f32) -> f32 {
+    return sqrt(dist * dist + 2.0 * view_r * view_mu * dist + view_r * view_r);
+}
+
+// Convert uv [0.0 .. 1.0] coordinate to ndc space xy [-1.0 .. 1.0]
+fn uv_to_ndc(uv: vec2<f32>) -> vec2<f32> {
```
pcwalton, line 156: This function already exists under that name in `view_transformations.wgsl`

crates/bevy_pbr/src/atmosphere/functions.wgsl
```diff
@@ -0,0 +1,208 @@
...
+    return uv * vec2(2.0, -2.0) + vec2(-1.0, 1.0);
+}
+
+/// Convert a ndc space position to world space
+fn position_ndc_to_world(ndc_pos: vec3<f32>) -> vec3<f32> {
```
pcwalton, line 161: This function already exists under that name in `view_transformations.wgsl`

crates/bevy_pbr/src/atmosphere/functions.wgsl
```diff
@@ -0,0 +1,208 @@
...
+    return textureSampleLevel(aerial_view_lut, aerial_view_lut_sampler, ndc, 0.0);
+}
+
+// PHASE FUNCTIONS
+
+fn rayleigh(neg_LdotV: f32) -> f32 {
+    return FRAC_3_16_PI * (1 + (neg_LdotV * neg_LdotV));
+}
+
+fn henyey_greenstein(neg_LdotV: f32) -> f32 {
```
pcwalton, line 78: Probably worth factoring this out somewhere so that both volumetric fog and atmosphere can use it

crates/bevy_pbr/src/atmosphere/sky_view_lut.wgsl
```diff
@@ -0,0 +1,55 @@
...
+}
+
+//lat-long projection [-pi, pi] x [-pi/2, pi/2] -> S^2
+fn get_ray_direction(lat_long: vec2<f32>) -> vec3<f32> {
+    let cos_long = cos(lat_long.y);
+    let sin_long = sin(lat_long.y);
+    let horizontal_rotation = mat2x2(cos_long, -sin_long, sin_long, cos_long);
+    let horizontal = horizontal_rotation * vec2(-view.world_from_view[2].xz);
+
+    return normalize(vec3(horizontal.x, sin(lat_long.x), horizontal.y));
```
pcwalton, line 54: Yeah, this would definitely be simpler as a cubemap.

ecoskey:
> Would you like reviews/feedback on this at this stage or is it too early?

The structure of everything is pretty much final so feedback on that would be nice :). At this stage I'm mainly trying to make sure everything is physically correct, so things will change there but if you see something I'm clearly doing wrong lmk

crates/bevy_pbr/src/atmosphere/functions.wgsl
ecoskey: I copied it so that the import didn't bring in all the mesh view bindings. Or would that not happen?

crates/bevy_pbr/src/atmosphere/functions.wgsl
ecoskey: I copied it so that the import didn't bring in all the mesh view bindings. Or would that not happen?

crates/bevy_pbr/src/atmosphere/functions.wgsl
ecoskey: I believe the author did it this way because the latitude is parameterized to put most of the detail near the horizon, which *does* have a noticeable effect on the final result. There's probably room to speed up my math here, but I'd want to make sure a cubemap could preserve that detail.

ecoskey: I should check how Hillaire does this, since apparently it's good enough for unreal

ecoskey:
> Considering the `Atmosphere` takes parameters about the inner and outer radius of the atmosphere, maybe it would make sense if that component was an another entity? That would allow you to compute the camera's view within (or outside) the atmosphere based on the position of the atmosphere and the camera using their `GlobalTransform`s. Excited to see this implemented!

Some of the main assumptions made are that the center of the planet is at position `(view.x, -bottom_radius, view.z)` and that the camera never leaves the atmosphere. So this implementation doesn't support worlds that are *actually* spherical, or space scenes where the camera can view the atmosphere from above. In the future once this gets integrated with pcwalton's raytracing code, it might be work removing these restrictions. In that case I'd agree that the atmosphere/planet should be its own entity

aevyrie:
> Some of the main assumptions made

I thought the Hillaire paper explicitly allows for space views of the atmosphere? Anyway, I totally understand if it is out of scope. Thanks for working on this!

ecoskey:
> > Some of the main assumptions made
> 
> I thought the Hillaire paper explicitly allows for space views of the atmosphere? Anyway, I totally understand if it is out of scope. Thanks for working on this!

It does, but the higher altitude the camera is the worse precision the sky-view and aerial-view LUTs are, even within the atmosphere. The paper recommends switching to per-pixel raytracing for high-altitude or space shots, which is where I'd want to integrate pcwalton's volumetrics code.

crates/bevy_pbr/src/atmosphere/multiscattering_lut.wgsl
```diff
@@ -0,0 +1,107 @@
...
+
+    return vec3(cos_lambda * cos(phi), sin_lambda, cos_lambda * sin(phi));
+}
+
+@compute 
+@workgroup_size(16, 16, 1)
+fn main(@builtin(global_invocation_id) global_id: vec3<u32>) {
+    let uv: vec2<f32> = (vec2<f32>(global_id.xy) + 0.5) / vec2<f32>(settings.multiscattering_lut_size);
+
+    //See Multiscattering LUT paramatrization
```
atlv24, line 39:
```suggestion
    //See Multiscattering LUT parametrization
```

crates/bevy_pbr/src/render/fast_math.wgsl
```diff
@@ -0,0 +1,43 @@
...
+#import bevy_render::maths::PI;
+
+// Reference: XEngine
+// https://github.com/ShawnTSH1229/XEngine/blob/main/Source/Shaders/FastMath.hlsl
+
+// 4th order polynomial approximation
+// 4 VGRP, 16 ALU Full Rate
+// 7 * 10^-5 radians precision
+// Reference : Handbook of Mathematical Functions (chapter : Elementary Transcendental Functions), M. Abramowitz and I.A. Stegun, Ed.
+fn fast_acos(x: f32) -> f32 {
```
JMS55, line 12: Fyi something similar already exists in ssao_utils.wgsl.

ecoskey: oh I didn't notice lol. The polynomial version seems a little more accurate but I don't need that too much. I'll just merge the ssao_utils stuff into fast_math then, except the multibounce stuff.

ecoskey:
Gonna put this up for review, I know the sunrise/set colors aren't perfectly accurate, but I've been unable to fix it. I'd be perfectly fine merging this as an experimental module, considering I still need to merge it and pcwalton's volumetrics stuff into a single module as a followup. I'm just really ready to be done with this pr lmao.

Cargo.toml
```diff
@@ -1236,7 +1247,7 @@ setup = [
     "curl",
     "-o",
     "assets/models/bunny.meshlet_mesh",
-    "https://raw.githubusercontent.com/JMS55/bevy_meshlet_asset/defbd9b32072624d40d57de7d345c66a9edf5d0b/bunny.meshlet_mesh",
```
JMS55, line 1239: I think this file got clobbered in a merge.

crates/bevy_pbr/src/render/fast_math.wgsl
```diff
@@ -0,0 +1,52 @@
+#define_import_path bevy_pbr::fast_math
+
+#import bevy_render::maths::{PI, HALF_PI};
+
+fn fast_sqrt(x: f32) -> f32 {
+    return bitcast<f32>(0x1fbd1df5 + (bitcast<i32>(x) >> 1u));
+}
+
+//slightly less accurate than fast_acos_4, but much simpler.
```
JMS55, line 9:
```suggestion
// Slightly less accurate than fast_acos_4, but much simpler.
```

Cargo.toml
ecoskey: reverted to the ref on `main`

Cargo.toml
aevyrie, line 1: I think this is supposed to be alphabetical

crates/bevy_pbr/src/atmosphere/mod.rs
```diff
@@ -0,0 +1,455 @@
...
+/// since integrating very small atmospheric densities over long distances
+/// might otherwise cause problems. [`AtmosphereSettings`] has a field to set
+/// the conversion factor from scene units to km, which by default assumes that
+/// the scene units are meters.
+#[derive(Clone, Component, Reflect, ShaderType)]
+#[require(AtmosphereSettings)]
+pub struct Atmosphere {
+    /// Radius of the planet
+    ///
+    /// units: km
```
aevyrie, line 246: While it might be beneficial for the computations to be done in km, the public API should use meters to stay consistent with the rest of bevys units. This conversion precision loss will be a one time thing when computing the uniform, and will be completely negligible.

crates/bevy_pbr/src/atmosphere/mod.rs
```diff
@@ -0,0 +1,455 @@
...
+    pub bottom_radius: f32,
+
+    /// Radius at which we consider the atmosphere to 'end' for our
+    /// calculations (from center of planet)
+    ///
+    /// units: km
```
aevyrie, line 252: Ditto

crates/bevy_pbr/src/atmosphere/mod.rs
```diff
@@ -0,0 +1,455 @@
...
+    ///
+    /// THIS VALUE MUST BE POSITIVE
+    ///
+    /// units: N/A
+    pub rayleigh_density_exp_scale: f32,
+
+    /// The scattering optical density of rayleigh particulare, or how
+    /// much light it scatters per kilometer
+    ///
+    /// units: km^-1
```
aevyrie, line 272: I'd use SI unless this is what measurements are usually given in.

crates/bevy_pbr/src/atmosphere/mod.rs
```diff
@@ -0,0 +1,455 @@
...
+    /// THIS VALUE MUST BE POSITIVE
+    ///
+    /// units: N/A
+    pub mie_density_exp_scale: f32,
+
+    /// The scattering optical density of mie particular, or how much light
+    /// it scatters per kilometer.
+    ///
+    /// units: km^-1
+    pub mie_scattering: f32,
```
aevyrie, line 287: I won't bore you by repeating myself any more. 😅

crates/bevy_pbr/src/atmosphere/mod.rs
```diff
@@ -0,0 +1,455 @@
...
+    /// computing the sky-view LUT.
+    pub sky_view_lut_samples: u32,
+
+    /// The number of points to sample for each slice along the z-axis
+    /// of the aerial-view LUT.
+    pub aerial_view_lut_samples: u32,
+
+    /// A conversion factor between scene units and kilometers, used to
+    /// combat floating-point precision issues.
+    pub scene_units_to_km: f32,
```
aevyrie, line 417: Same here obviously, but I think it becomes even clearer when this defaults to 1.0.

crates/bevy_pbr/src/render/fast_math.wgsl
aevyrie, line 1: A comment at the top of the file would be nice, describing when to use this?

aevyrie (approved):
No showstoppers. Looks great! Some public API nits.

I can't comment on the quality of the compute shaders, or most of the rendering for that matter. The comments sure help though!

examples/3d/atmosphere.rs
```diff
@@ -0,0 +1,110 @@
...
+        first_cascade_far_bound: 0.3,
+        maximum_distance: 3.0,
+        ..default()
+    }
+    .build();
+
+    // Sun
+    commands.spawn((
+        DirectionalLight {
+            color: Color::srgb(0.98, 0.95, 0.82),
```
mate-h, line 57: Setting the directional light's color to srgb white (1,1,1) here results in more physically correct result since the transmittance though the atmosphere of the directional light is already accounted for with the throughput variable and the analytical integration method used.

examples/3d/atmosphere.rs
```diff
@@ -0,0 +1,110 @@
...
+            ..default()
+        },
+        Transform::from_xyz(1.0, -1.0, 0.0).looking_at(Vec3::ZERO, Vec3::Y),
+        cascade_shadow_config,
+    ));
+
+    let sphere_mesh = meshes.add(Mesh::from(Sphere { radius: 1.0 }));
+
+    // light probe spheres
+    commands.spawn((
```
mate-h, line 69: The environment does not show up on the light probes, but this will be used as test subject in a future iteration where we generate the environment cube maps from sampling the sky-view LUT.

examples/3d/atmosphere.rs
```diff
@@ -0,0 +1,110 @@
...
+        .add_plugins(DefaultPlugins)
+        .add_systems(Startup, (setup_camera_fog, setup_terrain_scene))
+        .add_systems(Update, dynamic_scene)
+        .run();
+}
+
+fn setup_camera_fog(mut commands: Commands) {
+    commands.spawn((
+        Camera3d::default(),
+        Camera {
```
mate-h, line 23: Setting the exposure lower at noon and higher at sunset would make the showcase look more balanced, as well as adding a secondary directional light (moon) at night time to showcase a day-night cycle.

mate-h (approved):
The compute shaders look physically correct and I've reviewed and tested them and compared it to reference with the Hillarie paper and accompanying example application.

aevyrie:
FYI turning on `MSAA` causes the app to crash. Should probably defend against this. Some render features see MSAA turned on, warn, and don't run anything in the plugin.

ecoskey:
> FYI turning on `Msaa` causes the app to crash. Should probably defend against this. Some render features see MSAA turned on, warn, and don't run anything in the plugin.

this is a bug then, msaa should be compatible. will fix 👍

crates/bevy_pbr/src/atmosphere/aerial_view_lut.wgsl
```diff
@@ -0,0 +1,75 @@
...
+
+            throughput *= sample_transmittance;
+            if all(throughput < vec3(0.001)) {
+                break;
+            }
+
+            sum_transmittance += throughput.r + throughput.g + throughput.b;
+        }
+        //We only have one channel to store transmittance, so we store the mean 
+        let mean_transmittance = sum_transmittance / (f32(settings.aerial_view_lut_samples) * 3.0);
```
mate-h, line 67: I don't think it's correct to take the transmittance though the atmosphere as the throughput, since that is a variable used for analytical integration. As evidenced by the calculation in the `transmittance_lut.wgsl` entry point, we first sum up the optical depth and then take beer's law `exp(-sum_optical_depth)` to determine the actual transmittance.

ecoskey: multiplying the transmittance/throughput of several segments should be equivalent to summing up the optical depth and then taking the exponential, should it not?

mate-h: While your point about summing exponentials being equivalent is valid, I think that still the current transmittance calculation is still incorrect. We're summing up all intermediate RGB throughput values and dividing by (samples * 3), when we should just be taking the RGB average of the final throughput at each slice.

```wgsl
for (each sample) {
  sample_transmittance = exp(-optical_depth)
  throughput *= sample_transmittance
}
let mean_transmittance = (throughput.r + throughput.g + throughput.b) / 3.0;
```

So what I meant to say earlier was more about the point how the average is calculated. We should be taking the final throughput variable as opposed to creating a sum with intermediate throughput values while we are in the ray march loop.

ecoskey: ah yep you're correct

mate-h:
> FYI turning on `Msaa` causes the app to crash. Should probably defend against this. Some render features see MSAA turned on, warn, and don't run anything in the plugin.

I noted that, in my test environment `Msaa::Sample4` works fine with the atmosphere shader as well, while `Msaa::Sample8` crashes which is expected due to limitations in the device. So this is no cause for concern.

JMS55:
For reviewers: Can we get an updated list of known problems posted in the comments? Is this PR still in draft or ready for reviews and merging?

ecoskey:
> For reviewers: Can we get an updated list of known problems posted in the comments? Is this PR still in draft or ready for reviews and merging?

Only known issues:
- banding in sky view lut at high altitudes
- view space "rings" in aerial view when looking down from high altitude.

However, issues at high altitude/long distances are to be expected with the lut based approach. Mitigations are possible but I don't think need to be a blocker.

ecoskey:
Actually there's another: MSAAx8 for Rgba16Float on MacOS. Is a hard error fine here? It's supported for other sample counts so we shouldn't disable the plugin entirely, I'm open to input for error handling standards here.

JMS55:
Issues at high altitudes are fine for this PR imo, we don't need to fix that now as long as it works from the ground and reasonable heights.

Have all the exposure issues been fixed? At least manual exposures, if auto exposure needs tuning that's a separate, unrelated issue.

Does the MSAA issue only occur when rendering the atmosphere/LUTs, or does it occur when you use 8x MSAA on macOS with camera hdr=true? Imo we're fine to just let it crash. Maybe add a warning to either the atmosphere docs or MSAA docs.

ecoskey:
Exposure/sun color should all be fully accurate now. 

Where before we were using a slightly yellow sun with `lux::AMBIENT_SUNLIGHT` in the example, which incorrectly approximated the sun illuminance *after* passing through the atmosphere, now we're using a white sun with `lux::RAW_SUNLIGHT = 130_000`, which gets us much closer to the the look of the unreal implementation. 

We also fixed the issue with the sun disks not having proper bloom applied, which was happening because of mismatched units. Directional lights specify illuminance, or light falling on an area, which isn't accurate for considering light rays directly hitting the camera sensor. To correct this, we divide by the sun's solid angle in the sun disk loop.

examples/3d/atmosphere.rs
```diff
@@ -0,0 +1,108 @@
...
+            hdr: true,
+            ..default()
+        },
+        Exposure::SUNLIGHT,
+        Tonemapping::AcesFitted,
```
IceSentry, line 28: Is there a reason why the example uses a different tonemapper? Was it just for testing the values or is it also recommended to use it?

examples/3d/atmosphere.rs
```diff
@@ -0,0 +1,108 @@
...
+fn main() {
+    App::new()
+        .add_plugins(DefaultPlugins)
+        .add_systems(Startup, (setup_camera_fog, setup_terrain_scene))
+        .add_systems(Update, dynamic_scene)
+        .run();
+}
+
+fn setup_camera_fog(mut commands: Commands) {
+    commands.spawn((
```
IceSentry, line 21: I'd like a few comments explaining the components and which ones are required and which ones aren't.

crates/bevy_pbr/src/lib.rs
```diff
@@ -192,6 +195,12 @@ impl Plugin for PbrPlugin {
             Shader::from_wgsl
         );
         load_internal_asset!(app, UTILS_HANDLE, "render/utils.wgsl", Shader::from_wgsl);
+        load_internal_asset!(
```
IceSentry, line 198: I feel like this should be in `bevy_render`. Some people may want to use this in a custom impl that isn't using our PBR.

crates/bevy_pbr/src/atmosphere/mod.rs
```diff
@@ -0,0 +1,457 @@
...
+    type QueryFilter = (With<Camera3d>, With<Atmosphere>);
+
+    type Out = AtmosphereSettings;
+
+    fn extract_component(item: QueryItem<'_, Self::QueryData>) -> Option<Self::Out> {
+        Some(item.clone())
+    }
+}
+
+fn configure_camera_depth_usages(
```
IceSentry, line 451: I really dislike how we have that system in a bunch of different places. Don't need to do anything in this PR but I think there's something a bit weird with the fact that we need it in so many places.

IceSentry (approved):
I can't really comment on the math, but the results look good and the implementation itself is fine.

I added a few minor comments but none of them are blockers.

examples/3d/atmosphere.rs
ecoskey: Just looked nice for the example. Left a comment noting that any would work :+1:

ecoskey:
bevy_example_runner results: https://pixel-eagle.com/project/B25A040A-A980-4602-B90C-D480AB84076D?filter=PR-16314

pcwalton:
Unassigning myself as this has enough reviews already.

alice-i-cecile:
Agreed, merging :) Thanks!

alice-i-cecile:
Thank you to everyone involved with the authoring or reviewing of this PR! This work is relatively important and needs release notes! Head over to https://github.com/bevyengine/bevy-website/issues/1984 if you'd like to help out.

PR #16319, opened 2024-11-10
Add git hooks (opt-in)

BenjaminBrienen:
# Objective

Make development as convenient as possible for contributors. The issue identified is that it is very common to forget to run `cargo fmt --all` or generate the templated files. Seeing that the CI failed and then doing so and checking that in increases the iteration time.

## Solution

Provide git hooks to developers so that they can have this check done locally automatically. This feature is opt-in, of course. It increases the processing time for each commit and each push.

## Testing

image

### DIFF
--- a/.githooks/pre-commit
+++ b/.githooks/pre-commit
@@ -1,0 +1,15 @@
+#!/bin/sh
+
+# Exit on error, undefined variables, and prevent command masking
+set -euo pipefail
+
+# Set file permissions to read-write for all files
+echo "Setting file permissions to read-write for all files..."
+git ls-files -z | xargs -0 chmod 644
+
+# Run cargo fmt to ensure code is formatted
+echo "Running cargo fmt check..."
+if ! cargo fmt --all --check; then
+    echo "Error: Code is not formatted. Run 'cargo fmt --all' to format it."
+    exit 1
+fi
--- a/.githooks/pre-push
+++ b/.githooks/pre-push
@@ -1,0 +1,30 @@
+#!/bin/sh
+
+# Exit on error, undefined variables, and prevent command masking
+set -euo pipefail
+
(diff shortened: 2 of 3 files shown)


mockersf:
I very much dislike hooks and won't use them.

What's the point of view of other contributors?

bushrat011899 (approved):
I think this is worth including in the repo. Not everyone uses them, but it's pretty low-effort to keep these up-to-date and could help some people, especially new contributors.

I do think it would be better to move these checks into a new `ci` subcommand instead though. Especially since I believe there is some shared functionality here (the format check for example). It would also make debugging a little easier being able to invoke the hook _without_ a commit.

bushrat011899:
> I very much dislike hooks and won't use them.
> 
> What's the point of view of other contributors?

I've used them before, and I'm not really for or against them. They can help, but if you're someone who likes to commit as they go and squash before pushing it can be a pain.

As an optional thing, I don't really see any harm in offering it for those that do like them.

mockersf:
yup on local I do a lot of commits with fixups and squashes and resplitting commits before pushing.

I'm not opposed to adding hooks, but I want to know if some (more than a few) contributors would actually use them

mockersf:
The documentation would probably be better in the contributor book than here. I think we're (slowly) trying to remove those few markdown files for better discoverability, @alice-i-cecile could confirm

BenjaminBrienen:
I was initially going to add it into the CONTRIBUTING, but then I saw that we don't use that file and I don't know how to add to the website. That's all to say, I agree!

mockersf:
you can open a PR on https://github.com/bevyengine/bevy-website, probably on https://github.com/bevyengine/bevy-website/blob/main/content/learn/contribute/helping-out/opening-pull-requests.md

Issue #16316, opened 2024-11-09
bevy_text feature is required to run bevy_ui tests


UkoeHB:
## Bevy version

0.15-dev

## What you did

Ran `cargo test -p bevy_ui`

## What went wrong

Compile errors saying `bevy_text` is missing.

## Additional information

Works if you use `cargo test -p bevy_ui --features bevy_text`. The need to pass the feature here is unexpected and @mockersf says it's a bug.

BenjaminBrienen:
@alice-i-cecile is this 0.15 milestone?

mockersf:
not a blocker for the 0.15, but nice to have if there's a PR for it

StrikeForceZero:
Ran into this updating one of my PR's. I think its just a [few cargo config related tweaks](https://github.com/bevyengine/bevy/pull/16358) missed from 396aff906e401957507726e3af7d997004117c60

Issue #3815, opened 2022-01-30
UiPlugin cannot be used without rendering


alice-i-cecile:
## Bevy version

0.6

## What you did

```
use bevy::prelude::*;
use bevy::ui::UiPlugin;

fn main() {
    App::new()
        .add_plugins(MinimalPlugins)
        .add_plugin(UiPlugin)
        .run();
}
```

## What actually happened

```
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', C:\Users\Alice\.cargo\registry\src\github.com-1ecc6299db9ec823\bevy_ui-0.6.0\src\render\mod.rs:62:70
```
This panic is in https://github.com/bevyengine/bevy/blob/458cb7a9e93dab3127bb99ce7bf8cfc3af18851d/crates/bevy_ui/src/render/mod.rs#L61.

## Additional information

Related to #3155.

This is essential for headless integration testing of UI logic.

Wcubed:
For headless testing of UI, see the example in this pull request for adding headless windows: #3835

YourAlly:
Just started learning Bevy and Rust, where should I start for this issue?

alice-i-cecile:
> Just started learning Bevy and Rust, where should I start for this issue?

Thanks @YourAlly :) So, the steps I would take are:

1. Reproduce the problem locally.
2. Identify why this is happening (it appears that the resource is not being added).
3. Modify the plugin such that this system is not added or compiled if the rendering feature is disabled.
4. Verify that the problem is fixed.
5. Open a PR with your changes.

https://doc.rust-lang.org/cargo/reference/features.html is a useful learning resource here; you can see how to use feature flags to conditionally enable compilation of certain code segments.

Feel free to check in with me or report back on this thread if you get stuck.

kirusfg:
With my PR, the example provided simply _compiles_. I am not sure what UI functionalities must work in a headless mode, hence provided no tests.

kirusfg:
The simple UI test provided in #3835 (check if button is centered in a window without a window) passes with rendering disabled by preventing

- `widget::text_system`
- `widget::image_node_system`

from loading when `UIPlugin` is added. These systems heavily depend on `textures: Res<Assets<Image>>`; that is the reason `TextPlugin` will not work without rendering, too.

[...]

PR #16322, opened 2024-11-10
Feature gate render in UI (fix bevy_test feature)

### LINKED ISSUE #16313: Feature-gate render in `bevy_ui`

Due to `bevy_ui`'s dependency on `bevy_render` as a rendering backend which has a strong integration with `wgpu`, `bevy_ui` cannot currently support `no_std`. However, UI is foundational to many games, and other features of `bevy_ui` that would not depend on `std` (especially layouting) would be great to have in `no_std`. Rendering should be feature-gated out of `bevy_ui` which would allow others to develop their own platform-specific render for UI.

Add a `bevy_render` (or `render`) feature that would gate rendering integration with `bevy_render` behind it. This should be enabled by default.

Break out the base UI Node and layouting systems into a separate crate (`bevy_layout`?) that could then be made `no_std` on its own. This could be confusing for people but also might be worth it if `bevy_ui` gets too big in the future.

## Additional context

`no_std` support for `bevy_ui` is also currently blocked on `no_std` support for `bevy_ecs` and `bevy_app` (#15460), and a dependency with `accesskit` (#16312). These can be done in parallel however.

BenjaminBrienen:
# Objective

Fixes #16313 and #16316 and #3815

## Solution

Add a lot of feature gates. It looks awful. All configurations build with no warnings, though!

## Testing

```powershell
PS C:\Users\BenjaminBrienen\source\bevy> cargo build -p bevy_ui
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.39s
PS C:\Users\BenjaminBrienen\source\bevy> cargo build -p bevy_ui --no-default-features               
   Compiling bevy_ui v0.15.0-dev (C:\Users\BenjaminBrienen\source\bevy\crates\bevy_ui)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 2.98s
PS C:\Users\BenjaminBrienen\source\bevy> cargo build -p bevy_ui --no-default-features --features bevy_text
   Compiling bevy_ui v0.15.0-dev (C:\Users\BenjaminBrienen\source\bevy\crates\bevy_ui)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 3.28s
PS C:\Users\BenjaminBrienen\source\bevy> cargo build -p bevy_ui --no-default-features --features bevy_render
   Compiling bevy_ui v0.15.0-dev (C:\Users\BenjaminBrienen\source\bevy\crates\bevy_ui)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 4.27s
PS C:\Users\BenjaminBrienen\source\bevy> cargo build -p bevy_ui --no-default-features --features bevy_render --features bevy_text
   Compiling bevy_ui v0.15.0-dev (C:\Users\BenjaminBrienen\source\bevy\crates\bevy_ui)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 14.28s
```


BenjaminBrienen:
I think it makes sense to also fix bevy_text feature in this PR because there will be a lot of merge conflicts if split.

BenjaminBrienen:
After merge:
```powershell
PS C:\Users\BenjaminBrienen\source\bevy> cargo build -p bevy_ui --no-default-features
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.41s
PS C:\Users\BenjaminBrienen\source\bevy> cargo build -p bevy_ui --no-default-features --features bevy_text
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.39s
PS C:\Users\BenjaminBrienen\source\bevy> cargo build -p bevy_ui --no-default-features --features bevy_render
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.37s
PS C:\Users\BenjaminBrienen\source\bevy> cargo build -p bevy_ui --no-default-features --features bevy_render --features bevy_text
   Compiling bevy_ui v0.15.0-dev (C:\Users\BenjaminBrienen\source\bevy\crates\bevy_ui)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 6.15s
PS C:\Users\BenjaminBrienen\source\bevy> cargo build -p bevy_ui
   Compiling bevy_ui v0.15.0-dev (C:\Users\BenjaminBrienen\source\bevy\crates\bevy_ui)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 12.60s
```

crates/bevy_ui/src/widget/text.rs
```diff
@@ -48,18 +57,6 @@ impl Default for TextNodeFlags {
     }
 }
 
-/// [`TextBundle`] was removed in favor of required components.
```
alice-i-cecile, line 51: This definitely shouldn't be removed in this PR.

crates/bevy_ui/src/widget/text.rs
```diff
@@ -432,8 +433,8 @@ pub fn text_system(
             continue;
         };
         let scale_factor = match scale_factors_buffer.entry(camera_entity) {
-            Entry::Occupied(entry) => *entry.get(),
-            Entry::Vacant(entry) => *entry.insert(
+            bevy_utils::Entry::Occupied(entry) => *entry.get(),
```
alice-i-cecile, line 436: I don't understand why these diffs exist.

crates/bevy_ui/src/render/mod.rs
```diff
@@ -128,25 +124,36 @@ pub fn build_ui_render(app: &mut App) {
...
-        )
-        .add_systems(
-            Render,
-            (
-                queue_uinodes.in_set(RenderSet::Queue),
-                sort_phase_system::<TransparentUi>.in_set(RenderSet::PhaseSort),
-                prepare_uinodes.in_set(RenderSet::PrepareBindGroups),
-            ),
         );
+    #[cfg(feature = "bevy_text")]
```
alice-i-cecile, line 128: There's too much duplication here. Add the systems unconditionally, and then just conditionally add extract_text_sections.

crates/bevy_ui/src/node_bundles.rs
```diff
@@ -1,227 +0,0 @@
...
-///
-/// Contains the [`Node`] component and other components required to make a container.
-///
-/// See [`node_bundles`](crate::node_bundles) for more specialized bundles like [`ImageBundle`].
-#[derive(Bundle, Clone, Debug, Default)]
-#[deprecated(
-    since = "0.15.0",
-    note = "Use the `Node` component instead. Inserting `Node` will also insert the other components required automatically."
-)]
-pub struct NodeBundle {
```
alice-i-cecile, line 23: These also should not be removed here.

crates/bevy_ui/src/lib.rs
```diff
@@ -178,72 +196,169 @@ impl Plugin for UiPlugin {
...
             return;
         };
 
+        #[cfg(feature = "bevy_render")]
         render_app.init_resource::<UiPipeline>();
     }
 }
 
+#[cfg(all(feature = "bevy_text", feature = "bevy_render"))]
+fn make_systems() -> (
```
alice-i-cecile, line 241: I really dislike this approach; it's very complex.

crates/bevy_render/src/texture/mod.rs
```diff
@@ -21,14 +23,6 @@ use bevy_app::{App, Plugin};
 use bevy_asset::{AssetApp, Assets, Handle};
 use bevy_ecs::prelude::*;
 
-/// A handle to a 1 x 1 transparent white image.
-///
-/// Like [`Handle<Image>::default`], this is a handle to a fallback image asset.
-/// While that handle points to an opaque white 1 x 1 image, this handle points to a transparent 1 x 1 white image.
-// Number randomly selected by fair WolframAlpha query. Totally arbitrary.
-pub const TRANSPARENT_IMAGE_HANDLE: Handle<Image> =
```
alice-i-cecile, line 29: This shouldn't have been removed.

alice-i-cecile (changes requested):
I agree in principle but the diff here is extremely messy and needs to be reduced to a minimal set of changes.

crates/bevy_render/src/texture/mod.rs
BenjaminBrienen: It was just moved because it doesn't need to be behind the `bevy_render` feature gate.

crates/bevy_ui/src/lib.rs
BenjaminBrienen: What is a better approach?

crates/bevy_ui/src/widget/text.rs
BenjaminBrienen: This import was only used when these features are enabled, so rather than having the messy cfg for the tiny `utils` import, I just inlined it.

crates/bevy_ui/src/lib.rs
alice-i-cecile: Put each call to add_systems that needs to be conditional on its own line, with the appropriate cfg flag attached.

crates/bevy_ui/src/widget/text.rs
BenjaminBrienen: I figured this would be in 0.16 anyway, but I can make that a separate PR.

alice-i-cecile: Yeah, separate PR please!

BenjaminBrienen:
Latest:
image

crates/bevy_ui/src/accessibility.rs
```diff
@@ -1,22 +1,18 @@
...
-    schedule::IntoSystemConfigs,
     system::{Commands, Query},
-    world::Ref,
 };
 use bevy_render::{camera::CameraUpdateSystem, prelude::Camera};
 use bevy_transform::prelude::GlobalTransform;
 
-use accesskit::{Node, Rect, Role};
+use accesskit::Rect;
+use accesskit::{Node, Role};
```
bushrat011899, line 15: These changes don't make a lot of sense to me. Items are removed here but then referenced with their FQN further down (e.g., `Ref`).

crates/bevy_ui/src/experimental/ghost_hierarchy.rs
```diff
@@ -3,6 +3,7 @@
 use bevy_ecs::{prelude::*, system::SystemParam};
 use bevy_hierarchy::{Children, HierarchyQueryExt, Parent};
 use bevy_reflect::prelude::*;
+#[cfg(feature = "bevy_render")]
 use bevy_render::view::Visibility;
```
bushrat011899, line 7: Question for future PRs: Bevy `no_std` will likely introduce a lot more of these feature-gated dependencies in most crates. Stylistically, I think it makes sense to move them out of the main block of `use` statements. I personally find it harder to read, since my first thought it "_Ah the use statements below are only for `bevy_render`. Oh wait it's only this line, the others below are still always included._" Maybe not everyone has the same gut reaction but I still think there's room for a clarified style.

crates/bevy_ui/src/experimental/ghost_hierarchy.rs
```diff
@@ -20,7 +21,8 @@ use crate::Node;
 #[derive(Component, Debug, Copy, Clone, Reflect)]
 #[cfg_attr(feature = "ghost_nodes", derive(Default))]
 #[reflect(Component, Debug)]
-#[require(Visibility, Transform)]
+#[cfg_attr(feature = "bevy_render", require(Visibility, Transform))]
+#[cfg_attr(not(feature = "bevy_render"), require(Transform))]
```
bushrat011899, line 25: Future PR: Might be nice to modify the `Component` derive macro to allow multiple `#[require(...)]` attributes (if that's possible?). Would make cases like this cleaner.

crates/bevy_ui/src/focus.rs
```diff
@@ -1,25 +1,23 @@
...
 use bevy_ecs::{
-    change_detection::DetectChangesMut,
-    entity::Entity,
-    prelude::{Component, With},
-    query::QueryData,
-    reflect::ReflectComponent,
+    change_detection::DetectChangesMut as _,
+    prelude::With,
     system::{Local, Query, Res},
 };
```
bushrat011899, line 9: Nit: I personally prefer keeping feature gates to an absolute minimum, so I would write a larger `use` block for the whole gate instead:

```rust
#[cfg(feature = "bevy_render")]
use {
    crate::{DefaultUiCamera, ResolvedBorderRadius, TargetCamera, UiScale, UiStack},
    bevy_ecs::{...},
    ...
};
```

Fewer lines, and I think it's easier to read. "_For the `bevy_render feature, we need **these** items_"

crates/bevy_ui/src/focus.rs
```diff
@@ -127,8 +125,9 @@ impl Default for FocusPolicy {
 
 /// Contains entities whose Interaction should be set to None
 #[derive(Default)]
+#[cfg(feature = "bevy_render")]
 pub struct State {
-    entities_to_reset: SmallVec<[Entity; 1]>,
+    entities_to_reset: smallvec::SmallVec<[Entity; 1]>,
```
bushrat011899, line 130: Again, is there a reason for removing the `use smallvec::SmallVec;` just to use the FQN here instead? I personally prefer having the `use`-statement, since it makes it clear from the top of a file which crates are actually being used by this module.

crates/bevy_ui/src/widget/image.rs
```diff
@@ -3,16 +3,15 @@ use bevy_asset::{Assets, Handle};
...
-use bevy_render::texture::TRANSPARENT_IMAGE_HANDLE;
 use bevy_sprite::{TextureAtlas, TextureAtlasLayout, TextureSlicer};
-use bevy_window::{PrimaryWindow, Window};
+use bevy_window::PrimaryWindow;
+use bevy_window::Window;
 use taffy::{MaybeMath, MaybeResolve};
-
 /// A UI Node that renders an image.
 #[derive(Component, Clone, Debug, Reflect)]
-#[reflect(Component, Default, Debug)]
```
bushrat011899, line 15: Is there a reason reflection was removed here?

bushrat011899 (changes requested):
I like this change, but I think it needs some polishing. There's some formatting changes that I don't understand the reason for, and I believe reflection was removed from `ImageNode` inadvertently.

crates/bevy_ui/src/accessibility.rs
BenjaminBrienen: I can change the style. While working on it, I felt like it was often easier to avoid feature-gating an import by using the FQN.

crates/bevy_ui/src/focus.rs
BenjaminBrienen: I wasn't aware that the formatter would allow that. I assumed it would split it by crate.

crates/bevy_ui/src/focus.rs
BenjaminBrienen: No, just conciseness. I'd be glad to change it. Thanks for the feedback.

crates/bevy_ui/src/widget/image.rs
BenjaminBrienen: That looks like an error on my part. There should be a test to catch such a big change, tsk tsk tsk.

crates/bevy_ui/src/accessibility.rs
bushrat011899: That is fair, perhaps I missed a feature gate that was further up.

crates/bevy_ui/src/experimental/ghost_hierarchy.rs
bushrat011899: I was mistaken, you can actually do this right now, I just assumed it didn't work because I've never seen it written. You should be able to replace this with:

```rust
#[require(Transform)]
#[cfg_attr(feature = "bevy_render", require(Visibility))]
```

crates/bevy_ui/src/ui_node.rs
```diff
@@ -284,16 +287,32 @@ impl From<&Vec2> for ScrollPosition {
 /// - [`Interaction`](crate::Interaction) to obtain the interaction state of this node
 
 #[derive(Component, Clone, PartialEq, Debug, Reflect)]
-#[require(
```
BenjaminBrienen, line 287: Fix this

ickshonpe:
I haven't looked at the changes yet, but I've wanted to have the UI layout and rendering separated for years. 
I think just feature gating the renderer plugin isn't enough though, we need the ui rendering plugin moved into its own crate `bevy_ui_render` or something.

crates/bevy_ui/src/lib.rs
```diff
@@ -178,72 +200,95 @@ impl Plugin for UiPlugin {
...
+        #[cfg(feature = "bevy_text")]
         build_text_interop(app);
 
         #[cfg(feature = "bevy_ui_picking_backend")]
         app.add_plugins(picking_backend::UiPickingPlugin);
 
+        #[cfg(feature = "bevy_render")]
         if !self.enable_rendering {
             return;
         }
```
ickshonpe, line 266:
```suggestion
```

I don't think there is a need for both this and the feature gate.

crates/bevy_ui/src/lib.rs
```diff
@@ -68,28 +80,33 @@ pub mod prelude {
...
     pub enable_rendering: bool,
 }
 
 impl Default for UiPlugin {
     fn default() -> Self {
         Self {
+            #[cfg(feature = "bevy_render")]
             enable_rendering: true,
         }
     }
```
ickshonpe, line 112:
```suggestion
/// The basic plugin for Bevy UI
#[derive(Default)]
pub struct UiPlugin;
```

Shouldn't be needed anymore.

crates/bevy_ui/src/lib.rs
BenjaminBrienen: It would be dead code

BenjaminBrienen:
https://github.com/bevyengine/bevy/pull/16362 might be good to merge first.

BenjaminBrienen:
image

alice-i-cecile:
> I haven't looked at the changes yet, but I've wanted to have the UI layout and rendering separated for years. I think just feature gating the rendering isn't enough though, we need the UI rendering plugin moved into its own crate `bevy_ui_render` or something.

I'm of the same mind, but I'm fine merging this as an incremental step towards that goal. This separation is really important for a number of reasons (alternative backends like Vello, no_std platforms, headless testing).

ickshonpe:
> > I haven't looked at the changes yet, but I've wanted to have the UI layout and rendering separated for years. I think just feature gating the rendering isn't enough though, we need the UI rendering plugin moved into its own crate `bevy_ui_render` or something.
> 
> I'm of the same mind, but I'm fine merging this as an incremental step towards that goal. This separation is really important for a number of reasons (alternative backends like Vello, no_std platforms, headless testing).

My naive thinking is that it might not be much more complicated to do both at the same time and it might make for work later when we do split up the crates. I'm not too worried about that though and fine with merging this as an iterative improvement (as long as it works, just going to do a proper review now).

crates/bevy_ui/src/lib.rs
```diff
@@ -22,44 +23,55 @@ pub mod picking_backend;
 
 use bevy_derive::{Deref, DerefMut};
 use bevy_reflect::{std_traits::ReflectDefault, Reflect};
+#[cfg(all(feature = "bevy_text", feature = "bevy_render"))]
 mod accessibility;
```
ickshonpe, line 27: `accessibility` shouldn't need text. It's using the text component to store strings, but they are meant to be read using a screenreader I think. I'm not sure the module works or makes sense though, so it doesn't matter I guess?

crates/bevy_ui/src/layout/mod.rs
```diff
@@ -292,10 +293,11 @@ with UI components as a child of an entity without UI components, your UI layout
         }
     });
 
-    for (camera_id, mut camera) in camera_layout_info.drain() {
+    for (_camera_id, mut camera) in camera_layout_info.drain() {
         let inverse_target_scale_factor = camera.scale_factor.recip();
 
-        ui_surface.compute_camera_layout(camera_id, camera.size, text_buffers, &mut font_system);
+        #[cfg(feature = "bevy_text")]
+        ui_surface.compute_camera_layout(_camera_id, camera.size, text_buffers, &mut font_system);
```
ickshonpe, line 300: If we don't call `compute_camera_layout` none of the nodes will be updated.

```suggestion
        ui_surface.compute_camera_layout(
            _camera_id,
            camera.size,
            #[cfg(feature = "bevy_text")]
            text_buffers,
            #[cfg(feature = "bevy_text")]
            &mut font_system,
        );
```

crates/bevy_ui/src/layout/ui_surface.rs
```diff
@@ -1,16 +1,16 @@
-use core::fmt;
-
-use taffy::TaffyTree;
-
+use crate::{layout::convert, LayoutContext, LayoutError, Node, NodeMeasure};
+#[cfg(feature = "bevy_text")]
+use crate::{Measure, MeasureArgs};
```
ickshonpe, line 3: Measure funcs are needed for other intrinsically sized content like images, not only text.

```suggestion
use crate::{Measure, MeasureArgs};
```

crates/bevy_ui/src/layout/ui_surface.rs
```diff
@@ -194,10 +194,11 @@ impl UiSurface {
     }
 
     /// Compute the layout for each window entity's corresponding root node in the layout.
+    #[cfg(all(feature = "bevy_render", feature = "bevy_text"))]
```
ickshonpe, line 197:
```suggestion
    #[cfg(feature = "bevy_render")]
```

crates/bevy_ui/src/layout/ui_surface.rs
```diff
@@ -194,10 +194,11 @@ impl UiSurface {
...
     pub fn compute_camera_layout<'a>(
         &mut self,
         camera: Entity,
-        render_target_resolution: UVec2,
+        render_target_resolution: bevy_math::UVec2,
         buffer_query: &'a mut bevy_ecs::prelude::Query<&mut bevy_text::ComputedTextBlock>,
         font_system: &'a mut CosmicFontSystem,
```
ickshonpe, line 203:
```suggestion
        #[cfg(feature = "bevy_text")] buffer_query: &'a mut bevy_ecs::prelude::Query<
            &mut bevy_text::ComputedTextBlock,
        >,
        #[cfg(feature = "bevy_text")] font_system: &'a mut CosmicFontSystem,
```
\
Also need to add three feature gates to the body of the function too:

```rust
        for root_nodes in camera_root_nodes {
            self.taffy
                .compute_layout_with_measure(
                    root_nodes.implicit_viewport_node,
                    available_space,
                    |known_dimensions: taffy::Size<Option<f32>>,
                     available_space: taffy::Size<taffy::AvailableSpace>,
                     _node_id: taffy::NodeId,
                     context: Option<&mut NodeMeasure>,
                     style: &taffy::Style|
                     -> taffy::Size<f32> {
                        context
                            .map(|ctx| {
                                #[cfg(feature = "bevy_text")]
                                let buffer = get_text_buffer(
                                    crate::widget::TextMeasure::needs_buffer(
                                        known_dimensions.height,
                                        available_space.width,
                                    ),
                                    ctx,
                                    buffer_query,
                                );
                                let size = ctx.measure(
                                    MeasureArgs {
                                        width: known_dimensions.width,
                                        height: known_dimensions.height,
                                        available_width: available_space.width,
                                        available_height: available_space.height,
                                        #[cfg(feature = "bevy_text")]
                                        font_system,
                                        #[cfg(feature = "bevy_text")]
                                        buffer,
                                        _phantom: default(),
                                    },
                                    style,
                                );
                                taffy::Size {
                                    width: size.x,
                                    height: size.y,
                                }
                            })
                            .unwrap_or(taffy::Size::ZERO)
                    },
                )
                .unwrap();
```

Then layout should work again without `bevy_text`.

ickshonpe (changes requested):

crates/bevy_ui/src/lib.rs
BenjaminBrienen: I agree. The module uses Label extensively, which I think uses Text. I'll double check that.

crates/bevy_ui/src/layout/mod.rs
```diff
@@ -292,10 +293,11 @@ with UI components as a child of an entity without UI components, your UI layout
         }
     });
 
-    for (camera_id, mut camera) in camera_layout_info.drain() {
+    for (_camera_id, mut camera) in camera_layout_info.drain() {
```
ickshonpe, line 296:
```suggestion
    for (camera_id, mut camera) in camera_layout_info.drain() {
```

ickshonpe (changes requested):

ickshonpe (commented):
There still seems to be something weird with the "bevy_text" feature as well.

Feature gating all the text spawning commands in the `testbed_ui` example by "bevy_text" and compiling without "bevy_text" enabled:

```
cargo run --example testbed_ui --no-default-features --features "android-game-activity android-game-activity android_shared_stdcxx animation bevy_asset bevy_audio bevy_color bevy_core_pipeline bevy_gilrs bevy_gizmos bevy_gltf bevy_mesh_picking_backend bevy_pbr bevy_picking bevy_remote bevy_render bevy_scene bevy_sprite bevy_sprite_picking_backend bevy_state bevy_ui bevy_ui_picking_backend bevy_window bevy_winit custom_cursor default_font hdr multi_threaded png smaa_luts sysinfo_plugin tonemapping_luts vorbis webgl2 x11"
```

results in errors like:

```
error[E0433]: failed to resolve: use of undeclared type `Text`
  --> examples/testbed/ui.rs:74:33
   |
74 | ...                   Text::new("Text Example"),
   |                       ^^^^ use of undeclared type `Text`
```

BenjaminBrienen:
I'll look into it, thanks.

BenjaminBrienen:
> There still seems to be something weird with the "bevy_text" feature as well.
> 
> Feature gating all the text spawning commands in the `testbed_ui` example by "bevy_text" and compiling without "bevy_text" enabled:
> 
> ```
> cargo run --example testbed_ui --no-default-features --features "android-game-activity android-game-activity android_shared_stdcxx animation bevy_asset bevy_audio bevy_color bevy_core_pipeline bevy_gilrs bevy_gizmos bevy_gltf bevy_mesh_picking_backend bevy_pbr bevy_picking bevy_remote bevy_render bevy_scene bevy_sprite bevy_sprite_picking_backend bevy_state bevy_ui bevy_ui_picking_backend bevy_window bevy_winit custom_cursor default_font hdr multi_threaded png smaa_luts sysinfo_plugin tonemapping_luts vorbis webgl2 x11"
> ```
> 
> results in errors like:
> 
> ```
> error[E0433]: failed to resolve: use of undeclared type `Text`
>   --> examples/testbed/ui.rs:74:33
>    |
> 74 | ...                   Text::new("Text Example"),
>    |                       ^^^^ use of undeclared type `Text`
> ```

I get a runtime error if I run this command

BenjaminBrienen:
```
PS C:\Users\BenjaminBrienen\source\bevy> cargo build -p bevy_ui --no-default-features
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.43s
PS C:\Users\BenjaminBrienen\source\bevy> cargo build -p bevy_ui --no-default-features --features bevy_render
   Compiling bevy_ui v0.15.0-dev (C:\Users\BenjaminBrienen\source\bevy\crates\bevy_ui)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 11.25s
PS C:\Users\BenjaminBrienen\source\bevy> cargo build -p bevy_ui --no-default-features --features bevy_text  
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.44s
PS C:\Users\BenjaminBrienen\source\bevy> cargo build -p bevy_ui --no-default-features --features bevy_text,bevy_render
   Compiling bevy_ui v0.15.0-dev (C:\Users\BenjaminBrienen\source\bevy\crates\bevy_ui)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 14.08s
```

alice-i-cecile (approved):

alice-i-cecile:
Cutting from the milestone: valuable, but not essential. At this point I expect this to be easier to redo from scratch: merge conflicts on this sort of work really really suck.

Issue #11478, opened 2024-01-22
Remove or shrink bevy_utils


doonv:
## What problem does this solve or what need does it fill?

`bevy_utils` is a collection of unrelated utilities for Bevy. Utility modules/classes/libraries are generally frowned upon in the programming community since they are most likely a sign of poor organization.

## What solution would you like?

Move `bevy_utils`'s contents into existing bevy crates or into separate crates.

### Long list of what to do with each item in `bevy_utils`

- [ ] `BoxedFuture` could probably be removed since async in traits were added in Rust 1.75
- [ ] all `HashMap` related stuff: no idea where to put that.
- [x] https://github.com/bevyengine/bevy/pull/11498
- [ ] `OnDrop` could be its own crate
- [ ] `tracing` related stuff: put in `bevy_log`
- [ ] `once` related stuff: I would put it in `bevy_log`, but this could cause some issues (see https://github.com/bevyengine/bevy/pull/11419)
- [ ] `get_short_name`: put in `bevy_reflect`
- [ ] `SyncCell`: use `std::sync::Exclusive` when stabilized (https://github.com/rust-lang/rust/issues/98407)
- [ ] `SyncUnsafeCell`: use `std::cell::SyncUnsafeCell` when stablized (https://github.com/rust-lang/rust/issues/95439)
- [x] `generate_composite_uuid`: Maybe contribute this to the `uuid` crate?
- [x] `label` module: Sounds like a `bevy_reflect` thing but isn't actually used anywhere there so idk
- [x] `Interned`: put in `bevy_ecs`
- [ ] `futures` module: only used in `bevy_render`, so put it there.
- [x] `FloatOrd`: could be its own crate
- [ ] `default`: Could be its own crate (The [`default` crate](https://crates.io/crates/default) already exists, so maybe we should use that)
- [ ] `CowArc`: Could be its own crate

alice-i-cecile:
Yes please: I want this crate eliminated!

alice-i-cecile:
I think we may want to spin up new crates in the Bevy org for some of the stand-alone utilities.

@cart, would that make sense to you? Things like the `default` function crate would be good to split out, and will change slowly enough that I don't have concerns about moving it out of the monorepo.

That said, I know you're reluctant about increasing crate number, just because it's a popular misguided punching bag for "dependency bloat".

doonv:
[...]

PR #16325, opened 2024-11-10
Remove `bevy_utils::default()` and prefer `<_>::default()`

BenjaminBrienen:
# Objective

Contributes to #11478

## Solution

This has been attempted in a different manner before, but I want to see how this version is liked since it uses a more brief syntax than the previous attempt. I'm also interested to see if minds have changed in the last couple months. Close if unwanted.

## Testing

CI

## Migration Guide

`bevy_utils::prelude` and `bevy_utils::default` are removed. Use `<_>::default()` instead.


MrGVSV:
I'm pretty sure Cart added `default` to reduce the noise in Bevy code, especially in bundles. This PR seems to add more noise (and subjectively more than just using `Default::default`). It also means more typing since you can't just import `default` once, but now need to add `<_>::` to every call site

PR #14563, opened 2024-07-31
Handle TriggerTargets that are combinations for components/entities

cBournhonesque:
# Objective

- Fixes https://github.com/bevyengine/bevy/issues/14074
- It is currently not possible to issues a trigger that targets a specific list of components AND a specific list of entities

## Solution

- We can now use `((A, B), (entity_1, entity_2))` as a trigger target, as well as the reverse

## Testing

- Added a unit test.
The triggering rules for observers are quite confusing:
- triggers once per entity target
- for each entity target, an observer system triggers if any of its components matches the trigger target components (but it triggers at most once, since we use an internal counter to make sure that an observer can run at most once per entity target)

PR #16326, opened 2024-11-10
Handle `TriggerTargets` that are combinations for components/entities

### LINKED ISSUE #14074: Observers missing API for triggers targetting components on a given Entity

- Only unsafe APIs currently exist that allow you to specify both target components and entities, used internally for `OnAdd`, `OnInsert`, etc. This behaviour should be accessible to users via the safe APIs

- Implement `TriggerTargets` for a combination of `Entity` and `ComponentId`.

## Additional context

This API seems to have been dropped in the refactor from builder to `TriggerTargets`, we should probably try and check in a fix before release although it is possible to just manually implement `TriggerTargets`.

BenjaminBrienen:
# Objective

Fixes #14074

It is currently not possible to issues a trigger that targets a specific list of components AND a specific list of entities

## Solution

We can now use `((A, B), (entity_1, entity_2))` as a trigger target, as well as the reverse

## Testing

Added a unit test.

The triggering rules for observers are quite confusing:

Triggers once per entity target

For each entity target, an observer system triggers if any of its components matches the trigger target components (but it triggers at most once, since we use an internal counter to make sure that an observer can run at most once per entity target)

(copied from #14563)


BenjaminBrienen:
I feel like I leveled up my macro skills on this one.

crates/bevy_ecs/src/observer/mod.rs
```diff
@@ -172,92 +173,131 @@ impl<'w, E, B: Bundle> DerefMut for Trigger<'w, E, B> {
 /// will run.
 pub trait TriggerTargets {
     /// The components the trigger should target.
-    fn components(&self) -> &[ComponentId];
+    fn components(&self) -> impl ExactSizeIterator<Item = ComponentId> + Clone;
```
ItsDoot, line 176:
```suggestion
    fn components(&self) -> impl ExactSizeIterator<Item = ComponentId> + Clone + '_;
```
Could we connect the return type to the `TriggerTargets`' lifetime to avoid allocations in the `Vec<T>` impl?

ItsDoot: The question isn't resolved because the `collect`s are still present, and it's not done for `TriggerTargets::entities` either.

BenjaminBrienen: @ItsDoot Now it is the wrong kind of iterator. Should I change the return type?

ItsDoot: I just tested it locally, you should be able to swap `ExactSizeIterator` to just `Iterator`

BenjaminBrienen: Done

ItsDoot (approved):

NthTensor:
On the face of this, this seems like an improvement. However I'm concerned about the increasing complexity of the observer api (much of which is my doing).

chescock (commented):
This does what it says, and given that a `TriggerTargets` can be a `Vec<ComponentId>` or a `Vec<Entity>` today, this does seem like the simplest extension!  

The result is that this PR allows interleaving components and entities within a list, though.  The behavior of something like `(component_a, entity_1, component_b, entity_2)` seems odd!  Is it worth trying to make that inexpressible?  

Why *are* components and entities packed into a single parameter like that?  I haven't been following the observers design closely, so I apologize if I'm asking about something that's already been discussed a lot!  What goes wrong if we instead split the `targets` parameter into separate `components` and `entities` parameters?  

Alternately, why not create a new `struct` that takes a slice of each and impls `TriggerTargets`, and have callers create one of those if they want to provide both?

BenjaminBrienen:
I don't know the why, tbh. This is just something I adopted because it seemed easy enough.

NthTensor:
> Why are components and entities packed into a single parameter like that? 

@james-j-obrien would know. I'm pretty sure @alice-i-cecile is open to the idea of refactoring the public observer api, but I'm not sure if this PR is the place to do it.

chescock:
> I'm not sure if this PR is the place to do it.

Oh, sorry, I didn't mean to ask to increase the scope of this PR!  I was just trying to understand the reasoning behind the API design.  

The code looks like a good implementation of that design!  (modulo the nits above)  So I'll approve the PR, with the caveat that I'm not familiar enough with observers to understand why we want an API that works like this.

chescock (approved):

james-j-obrien:
> > Why are components and entities packed into a single parameter like that? 
> 
> @james-j-obrien would know. I'm pretty sure @alice-i-cecile is open to the idea of refactoring the public observer api, but I'm not sure if this PR is the place to do it.

I don't think this was part of my original design, I'm pretty sure this was part of the API that cart refactored just before metge. At this point I don't remember what it originally looked like but I do agree the current iteration is confusing.

cBournhonesque (approved):

crates/bevy_ecs/src/storage/sparse_set.rs
```diff
@@ -750,7 +750,7 @@ mod tests {
             vec![(ComponentId::new(1), 0), (ComponentId::new(2), 0),]
         );
 
-        fn init_component<T: Component>(sets: &mut SparseSets, id: usize) {
+        fn register_component<T: Component>(sets: &mut SparseSets, id: usize) {
```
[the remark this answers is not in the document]
alice-i-cecile: I'd prefer a reversion here, but won't block on it.

alice-i-cecile:
@BenjaminBrienen CI is failing; take a look?

bushrat011899:
I've opened #18024 to apply fixes required to get this PR merged. It's just a fork of this PR so reviews here should equally apply there.

Issue #16172, opened 2024-10-30
Document expectations of Bevy crates


mockersf:
Bevy has many crates, with various level of expectations. I would like those levels to be documented, with their expectations and where they currently stand.

I think the current status is:

* Tier 1: bevy!
  * Code must compile on all platforms (macos, windows, linux, android, ios, wasm)
  * Features must be documented
  * Examples must be documented
  * Selected examples must run on all platforms
* Tier 2: standalone crates: bevy_ecs, bevy_reflect, bevy_color, ...
  * If the crate has unsafe code, run through miri
  * If the crate can build in no std, check it
  * Build without default features
* Tier 3: all the others
  * Not much

Tier 3 is mostly useful to plugin authors and power users, and is lacking in documentation. For example, documenting features of all crates would have a lot of value.

Things to do:
* [ ] document current state
* [ ] agree on expectations

alice-i-cecile:
I don't like the numbered tiers, but I agree that a categorization system would be useful!

Proposal:

- core: part of Bevy's default feature set, used to create a cohesive game engine. These must work on all platforms, be fully documented and so on.
- standalone: crates that can be used on their own but are designed specifically for Bevy. Cannot pull in `bevy_ecs`.
- utility: small, stable crates that Bevy happens to maintain, but are not actually tied to Bevy. These belong in the Bevy org but not the Bevy repo.
- optional: game engine functionality that isn't used by many games. Meshlets for example belong here, as would something like tilemap support. Most of the editor-supporting crates and dev tools seem like they belong here too.

bushrat011899:
I think this would be excellent to codify. With `no_std` support (hopefully) growing in the next release, it might be prudent to split `core` into two labels: `core` and `standard`. `core` would apply to anything that compiles in `no_std` (`bevy_ecs`, `bevy_app`, etc.), while `standard` is a superset of `core` and would include things like `bevy_audio` and `bevy_render`. I think this also overlaps quite nicely with the distinction between `core` and `std` within Rust itself.

alice-i-cecile:
This [came up again today](https://discord.com/channels/691052431525675048/789987800014651412/1343724015805661205) in the context of upstreaming various editor components as crates.

bushrat011899:
> Part of me wants to organize the various crates into subfolders based on this schema, but it would be incredibly breaking for various links and PRs, so it's probably not worth it.

[...]

PR #14505, opened 2024-07-28
Fix `bevy_render`'s `image` dependency version

SkiFire13:
# Objective

- `bevy_render` depends on `image 0.25` but uses `image::ImageReader` which was added only in `image 0.25.2`
- users that have `image 0.25` in their `Cargo.lock` and update to the latest `bevy_render` may thus get a compilation due to this (at least I did)

## Solution

- Properly set the correct minimum version of `image` that `bevy_render` depends on.

PR #16333, opened 2024-11-10
Fix MSRVs for standalone crates

BenjaminBrienen:
# Objective

MSRV in the standalone crates should be accurate

## Solution

Determine the msrv of each crate and set it

## Testing

Adding better msrv checks to the CI is a next-step.

### DIFF
--- a/crates/bevy_color/Cargo.toml
+++ b/crates/bevy_color/Cargo.toml
@@ -7,7 +7,7 @@ homepage = "https://bevyengine.org"
 repository = "https://github.com/bevyengine/bevy"
 license = "MIT OR Apache-2.0"
 keywords = ["bevy", "color"]
-rust-version = "1.76.0"
+rust-version = "1.82.0"
 
 [dependencies]
 bevy_math = { path = "../bevy_math", version = "0.15.0-dev", default-features = false, features = [
--- a/crates/bevy_ecs/Cargo.toml
+++ b/crates/bevy_ecs/Cargo.toml
@@ -8,7 +8,7 @@ repository = "https://github.com/bevyengine/bevy"
 license = "MIT OR Apache-2.0"
 keywords = ["ecs", "game", "bevy"]
 categories = ["game-engines", "data-structures"]
-rust-version = "1.77.0"
+rust-version = "1.81.0"
 
 [features]
 default = ["bevy_reflect"]
--- a/crates/bevy_math/Cargo.toml
+++ b/crates/bevy_math/Cargo.toml
@@ -7,7 +7,7 @@ homepage = "https://bevyengine.org"
 repository = "https://github.com/bevyengine/bevy"
 license = "MIT OR Apache-2.0"
 keywords = ["bevy"]
-rust-version = "1.68.2"
+rust-version = "1.81.0"
 
 [dependencies]
 glam = { version = "0.29", features = ["bytemuck"] }
--- a/crates/bevy_ptr/Cargo.toml
+++ b/crates/bevy_ptr/Cargo.toml
@@ -7,6 +7,7 @@ homepage = "https://bevyengine.org"
 repository = "https://github.com/bevyengine/bevy"
 license = "MIT OR Apache-2.0"
 keywords = ["bevy", "no_std"]
+rust-version = "1.81.0"
 
 [dependencies]
 
--- a/crates/bevy_reflect/Cargo.toml
+++ b/crates/bevy_reflect/Cargo.toml
@@ -7,7 +7,7 @@ homepage = "https://bevyengine.org"
 repository = "https://github.com/bevyengine/bevy"
 license = "MIT OR Apache-2.0"
 keywords = ["bevy"]
-rust-version = "1.76.0"
+rust-version = "1.81.0"
 
 [features]
 default = ["smallvec", "debug", "alloc"]


mockersf:
I think most of those changes are not justified outside of "you think it looks better"

I would prefer to not change the version of dependencies, even for something compatible, without a need to change it for that dependency. Even worse to do it for everything at once.

For MSRV, adding it to most crates will just make it more painful for us to maintain: for the crates that can't be used outside of Bevy, adding it doesn't bring us any value and will just need more work.

errors/Cargo.toml
```diff
@@ -1,6 +1,7 @@
 [package]
 name = "errors"
 edition = "2021"
+version = "0.1.0"
 description = "Documentation and tests for Bevy's error codes"
 publish = false
 license = "MIT OR Apache-2.0"
```
BenjaminBrienen, line 7: Missing

benches/Cargo.toml
```diff
@@ -4,12 +4,13 @@ edition = "2021"
...
 publish = false
 license = "MIT OR Apache-2.0"
+rust-version = "1.78.0"
 
 [dev-dependencies]
 glam = "0.29"
 rand = "0.8"
 rand_chacha = "0.3"
-criterion = { version = "0.3", features = ["html_reports"] }
+criterion = { version = "0.5", features = ["html_reports"] }
```
BenjaminBrienen, line 13: Undo

crates/bevy_core_pipeline/src/smaa/mod.rs
```diff
@@ -35,6 +35,8 @@ use crate::{
     core_3d::graph::{Core3d, Node3d},
 };
 use bevy_app::{App, Plugin};
+#[cfg(not(feature = "smaa_luts"))]
```
BenjaminBrienen, line 38: Move to different PR

.gitignore
```diff
@@ -1,18 +1,18 @@
 # Rust build artifacts
```
BenjaminBrienen, line 1: Move to different PR

crates/bevy_ecs/compile_fail/Cargo.toml
```diff
@@ -6,6 +6,7 @@ homepage = "https://bevyengine.org"
 repository = "https://github.com/bevyengine/bevy"
 license = "MIT OR Apache-2.0"
 publish = false
+rust-version = "1.81.0"
```
BenjaminBrienen, line 9: Undo

output.jsonc
BenjaminBrienen, line 1: Undo

tools/build-templated-pages/Cargo.toml
```diff
@@ -1,18 +1,18 @@
 [package]
 name = "build-templated-pages"
 edition = "2021"
+version = "0.1.0"
 description = "Tool that checks and fixes undocumented features and examples"
 publish = false
 license = "MIT OR Apache-2.0"
+rust-version = "1.73.0"
```
BenjaminBrienen, line 8: Undo

tools/build-wasm-example/Cargo.toml
```diff
@@ -1,14 +1,15 @@
 [package]
 name = "build-wasm-example"
 edition = "2021"
+version = "0.1.0"
 description = "Tool for building example for Wasm"
 publish = false
 license = "MIT OR Apache-2.0"
-
+rust-version = "1.74.1"
```
BenjaminBrienen, line 8: Undo

tools/ci/Cargo.toml
```diff
@@ -1,14 +1,16 @@
 [package]
 name = "ci"
 edition = "2021"
+version = "0.1.0"
 description = "Tool that enables running CI checks locally."
 publish = false
 license = "MIT OR Apache-2.0"
+rust-version = "1.72.1"
```
BenjaminBrienen, line 8: Undo

tools/compile_fail_utils/Cargo.toml
```diff
@@ -6,9 +6,10 @@ homepage = "https://bevyengine.org"
 repository = "https://github.com/bevyengine/bevy"
 license = "MIT OR Apache-2.0"
 publish = false
+rust-version = "1.78.0"
```
BenjaminBrienen, line 9: Undo

tools/example-showcase/Cargo.toml
```diff
@@ -1,19 +1,19 @@
 [package]
 name = "example-showcase"
 edition = "2021"
+version = "0.1.0"
 description = "Tool for running examples or generating a showcase page for the website."
 publish = false
 license = "MIT OR Apache-2.0"
+rust-version = "1.74.1"
```
BenjaminBrienen, line 8: Undo

BenjaminBrienen:
image

BenjaminBrienen:
To address the other feedback:

- I do think it looks better. The alternative option would be keeping the version numbers in-line with what versions are actually being used, which is incredibly noisy and unnecessary.

- I have ensured that no dependencies have been downgraded or updated by `diff`ing `Cargo.lock`.

- I have made sure that the `rust-version` is only specified for public crates. The maintenance cost will be hopefully the exact same as it is now or even less after I have finished improving the CI. We can make it so that the developer doesn't have to edit the value themselves, even. The bot can do that.

mockersf:
> * I do think it looks better. 

I don't. And I especially dislike doing it in a change across all crates, without a reason to do it related to the dependency.

> * I have made sure that the `rust-version` is only specified for public crates. The maintenance cost will be hopefully the exact same as it is now or even less after I have finished improving the CI. We can make it so that the developer doesn't have to edit the value themselves, even. The bot can do that.

I don't think having separate msrv for crates that are not standalone has enough value to bother, I don't think they should be added and I won't support adding those checks in CI. This can be revisited later for a 1.0 release, but for the moment the Bevy msrv policy is "latest stable", let's not go into details crate per crate.

BenjaminBrienen:
Ok. I'd like to discuss the second point a bit more if you don't mind. I thought the design philosophy of bevy is to have each crate be usable on their own. A project could just use `bevy_ecs`, for example. Why wouldn't we want the usability there to be the same as the main `bevy` crate? This is assuming we set the msrv of `bevy` accurately for usability and to provide value to dependents. To me, it works the same way for each public crate. Can you please tell me your perspective on that aspect?

crates/bevy_image/Cargo.toml
```diff
@@ -44,26 +45,26 @@ bevy_reflect = { path = "../bevy_reflect", version = "0.15.0-dev", features = [
 bevy_utils = { path = "../bevy_utils", version = "0.15.0-dev" }
 
 # rendering
-image = { version = "0.25.2", default-features = false }
+image = { version = "0.25", default-features = false }
```
rparrett, line 48: These aren't equivalent, and we want at least `0.25.2`. See https://github.com/bevyengine/bevy/pull/14505.

I think there are other instances where this is known to be true, and probably more yet that could be lurking issues if this were merged.

BenjaminBrienen: Wouldn't a `cargo update` fix that? If the developer doesn't want to do it manually, a post-pull hook would solve that problem.

rparrett: What would be the benefit to specifying a version requirement allowing versions that are known to cause issues?

alice-i-cecile:
> Ok. I'd like to discuss the second point a bit more if you don't mind. I thought the design philosophy of bevy is to have each crate be usable on their own. A project could just use `bevy_ecs`, for example. Why wouldn't we want the usability there to be the same as the main `bevy` crate? This is assuming we set the msrv of `bevy` accurately for usability and to provide value to dependents. To me, it works the same way for each public crate. Can you please tell me your perspective on that aspect?

So, this ties into https://github.com/bevyengine/bevy/issues/16172. Only the standalone crates (bevy_ecs, bevy_math, bevy_color, bevy_ptr, bevy_reflect) should have their own MSRV. For any crate that relies on bevy_ecs, their MSRV should be fully tied to bevy_ecs, which for the foreseeable future is going to be on "latest stable".

BenjaminBrienen:
I'll update this PR with the feedback when I have some free time. I'll fix some of the minor versions and remove more of the msrvs as requested. The main goal is consistency, correctness, and usability, so I'm glad to work with the feedback on that.

BenjaminBrienen:
Marking as blessed because I reduced the scope to exactly the changes endorsed by @alice-i-cecile.

alice-i-cecile (approved):
I agree with this scope and these fixes :)

mockersf (approved):

Issue #16244, opened 2024-11-05
Assets<T> should have `get_many` / `get_many_mut` or equivalent functions


Nathan-Fenner:
## What problem does this solve or what need does it fill?

Similar to `Query<...>`, the `Assets<T>` collection type maps from asset IDs to underlying data (e.g. sounds, animations, meshes, textures, ...). Because of this, it is unsound to allow multiple calls to `.get_mut()` on an `Assets<T>`, since the returned references could alias each other:

```rs
fn example(assets: &mut Assets<Image>, handle1: Handle<Image>, handle2: Handle<Image>) {
  let image_data1: &mut Image = assets.get_mut(&handle1);
  let image_data2: &mut Image = assets.get_mut(&handle2);

  println!("{:?} {:?}", image_data1.clone(), image_data2.clone()); // borrow error: multiple mutable borrows of `assets`
}
```

The `Query` type has the `get_many_mut` methods to get around this limitation, allowing multiple `Entity`s to be looked up at once, with a dynamic check to prevent the same entity from being accessed multiple times.

`Assets` should have a similar `get_many_mut` method with a dynamic check, offering safe mutable access to multiple (distinct) assets at once.

(Other designs are possible, but currently it makes sense to align with `Query<...>`)

PR #16334, opened 2024-11-10
Support for `get_many` and `get_many_mut` functions for `Assets<T>`

Rajveer100:
Part of #16244

# Objective

Add functions for `Assets<T>` to retrieve multiple references similar to `Query<...>`

## Solution

The Query type has the `get_many` and `get_many_mut` methods to get around this limitation, allowing multiple Entities to be looked up at once, with a dynamic check to prevent the same entity from being accessed multiple times. A similar pattern can be followed here.

### DIFF
--- a/crates/bevy_asset/src/assets.rs
+++ b/crates/bevy_asset/src/assets.rs
@@ -215,6 +215,19 @@ impl<A: Asset> DenseAssetStorage<A> {
         }
     }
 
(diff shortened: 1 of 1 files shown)


Rajveer100:
For retrieving mutable references, what do you think is an appropriate way?

Rajveer100:
Closing this in favor of https://github.com/bevyengine/bevy/pull/19484

PR #16340, opened 2024-11-11
Add `unregister_system` command

### LINKED ISSUE #16266: Add unregister_system for Commands

`Commands` currently has a `register_system` method, but no way to unregister systems. Unregistration is very useful in on_remove hooks.

Add an `unregister_system` method to `Commands`.

There are several alternatives, such as writing a custom command for this, but it would be cleaner (and more symmetrical) if this was provided on Commands directly.

grind086:
# Objective

Fixes #16266 

## Solution

Added an `UnregisterSystem` command struct and `Commands::unregister_system`. Also renamed `World::remove_system` and `World::remove_system_cached` to `World::unregister_*`

## Testing

It's a fairly simple change, but I tested locally to ensure it actually works.


BenjaminBrienen:
Can you add a unit test?

BenjaminBrienen (approved):

grind086:
Are there any existing tests to use as an example? I couldn't find any for the related commands other than one doctest example for `Commands::run_system`, but I could just be missing them if they're somewhere other than `system::commands::tests` or `system::system_registry::tests`. If not I can come up with something.

BenjaminBrienen:
If there aren't any tests for `register_system`, don't worry about it. We can open a separate issue for adding tests there. (Unless you feel like doing it here)

hymm:
I feel like this shouldn't be called remove_system. That name feels like the opposite of `add_systems`, so I would expect it to work with Schedules.

grind086:
I could change it to `unregister_system` like the original issue suggested if that's preferable. I went with the current name because it's consistent with `World::remove_system`, which is the non-command equivalent.

alice-i-cecile:
Both this and the method it calls should be named `unregister_system`. Once that's done, I'm happy to approve and merge.

crates/bevy_ecs/src/system/commands/mod.rs
```diff
@@ -890,6 +891,17 @@ impl<'w, 's> Commands<'w, 's> {
         SystemId::from_entity(entity)
     }
 
+    /// Removes a system previously registered with [`Commands::register_system`] or [`World::register_system`].
+    ///
+    /// See [`World::remove_system`] for more information.
+    pub fn remove_system<I, O>(&mut self, system_id: SystemId<I, O>)
```
BenjaminBrienen, line 897:
```suggestion
    pub fn unregister_system<I, O>(&mut self, system_id: SystemId<I, O>)
```

crates/bevy_ecs/src/system/commands/mod.rs
```diff
@@ -890,6 +891,17 @@ impl<'w, 's> Commands<'w, 's> {
         SystemId::from_entity(entity)
     }
 
+    /// Removes a system previously registered with [`Commands::register_system`] or [`World::register_system`].
+    ///
+    /// See [`World::remove_system`] for more information.
```
BenjaminBrienen, line 896:
```suggestion
    /// See [`World::unregister_system`] for more information.
```

crates/bevy_ecs/src/system/commands/mod.rs
```diff
@@ -890,6 +891,17 @@ impl<'w, 's> Commands<'w, 's> {
...
+    pub fn remove_system<I, O>(&mut self, system_id: SystemId<I, O>)
+    where
+        I: SystemInput + Send + 'static,
+        O: Send + 'static,
+    {
+        self.queue(RemoveSystem::new(system_id));
```
BenjaminBrienen, line 902:
```suggestion
        self.queue(UnregisterSystem::new(system_id));
```

crates/bevy_ecs/src/system/system_registry.rs
```diff
@@ -544,6 +544,32 @@ where
     }
 }
 
+/// The [`Command`] type for removing one shot systems from [`Commands`](crate::system::Commands).
+pub struct RemoveSystem<I: SystemInput + 'static, O: 'static> {
```
BenjaminBrienen, line 548:
```suggestion
pub struct UnregisterSystem<I: SystemInput + 'static, O: 'static> {
```

crates/bevy_ecs/src/system/system_registry.rs
```diff
@@ -544,6 +544,32 @@ where
     }
 }
 
+/// The [`Command`] type for removing one shot systems from [`Commands`](crate::system::Commands).
```
BenjaminBrienen, line 547:
```suggestion
/// The [`Command`] type for unregistering one-shot systems from [`Commands`](crate::system::Commands).
```

crates/bevy_ecs/src/system/system_registry.rs
```diff
@@ -544,6 +544,32 @@ where
...
+pub struct RemoveSystem<I: SystemInput + 'static, O: 'static> {
+    system_id: SystemId<I, O>,
+}
+
+impl<I, O> RemoveSystem<I, O>
```
BenjaminBrienen, line 552:
```suggestion
impl<I, O> UnregisterSystem<I, O>
```

crates/bevy_ecs/src/system/system_registry.rs
```diff
@@ -544,6 +544,32 @@ where
...
+    I: SystemInput + 'static,
+    O: 'static,
+{
+    /// Creates a new [`Command`] struct, which can be added to [`Commands`](crate::system::Commands).
+    pub fn new(system_id: SystemId<I, O>) -> Self {
+        Self { system_id }
+    }
+}
+
+impl<I, O> Command for RemoveSystem<I, O>
```
BenjaminBrienen, line 563:
```suggestion
impl<I, O> Command for UnregisterSystem<I, O>
```

crates/bevy_ecs/src/system/system_registry.rs
```diff
@@ -544,6 +544,32 @@ where
...
+where
+    I: SystemInput + 'static,
+    O: 'static,
+{
+    fn apply(self, world: &mut World) {
+        let _ = world.remove_system(self.system_id);
```
BenjaminBrienen, line 569:
```suggestion
        let _ = world.unregister_system(self.system_id);
```

crates/bevy_ecs/src/system/commands/mod.rs
```diff
@@ -3,7 +3,8 @@ mod parallel_scope;
 use core::{marker::PhantomData, panic::Location};
 
 use super::{
-    Deferred, IntoObserverSystem, IntoSystem, RegisterSystem, Resource, RunSystemCachedWith,
+    Deferred, IntoObserverSystem, IntoSystem, RegisterSystem, RemoveSystem, Resource,
```
BenjaminBrienen, line 6:
```suggestion
    Deferred, IntoObserverSystem, IntoSystem, RegisterSystem, UnregisterSystem, Resource,
```

BenjaminBrienen (commented):
I just wanted to try out the suggestion feature lol

crates/bevy_ecs/src/system/commands/mod.rs
alice-i-cecile: This will fail formatting because it's not alphabetized :)

grind086:
I think that's everything, assuming checks pass.

alice-i-cecile (approved):

alice-i-cecile:
Squeezing into 0.15 because it avoids us shipping the bad `remove_system` method name.

PR #16347, opened 2024-11-11
Minor docs fixes

richchurcher:
Happened upon a few stray characters while reading about picking.

### DIFF
--- a/crates/bevy_picking/src/backend.rs
+++ b/crates/bevy_picking/src/backend.rs
@@ -6,9 +6,9 @@
 //! detail.
 //!
 //! Because `bevy_picking` is very loosely coupled with its backends, you can mix and match as
-//! many backends as you want. For example, You could use the `rapier` backend to raycast against
+//! many backends as you want. For example, you could use the `rapier` backend to raycast against
 //! physics objects, a picking shader backend to pick non-physics meshes, and the `bevy_ui` backend
-//! for your UI. The [`PointerHits`]s produced by these various backends will be combined, sorted,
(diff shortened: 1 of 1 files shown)


BenjaminBrienen (approved):

mnmaita (approved):

crates/bevy_picking/src/backend.rs
```diff
@@ -6,9 +6,9 @@
 //! detail.
 //!
 //! Because `bevy_picking` is very loosely coupled with its backends, you can mix and match as
-//! many backends as you want. For example, You could use the `rapier` backend to raycast against
+//! many backends as you want. For example, you could use the `rapier` backend to raycast against
 //! physics objects, a picking shader backend to pick non-physics meshes, and the `bevy_ui` backend
-//! for your UI. The [`PointerHits`]s produced by these various backends will be combined, sorted,
+//! for your UI. The [`PointerHits`] produced by these various backends will be combined, sorted,
```
mockersf, line 11:
```suggestion
//! for your UI. The [`PointerHits`] events produced by these various backends will be combined, sorted,
```

`PointerHits` is the individual thing that is produced, so that "double plural" is somewhat correct here

richchurcher: Yeah, it did occur to me... :thinking: what about 
> The [`PointerHits`] instances produced...
to avoid the double 's'? Otherwise I don't feel strongly.

mockersf (approved):

PR #16348, opened 2024-11-11
Remove deprecated `component_reads_and_writes`

### LINKED ISSUE #16339: Migrate `component_reads_and_writes`
Migrate `component_reads_and_writes` for the 0.16 release. bevy uses it internally and it is deprecated.

~~It is being made `pub(crate)` in #16338.~~ didn't work because it is used in an example. That fact makes this a 0.15 issue :)

bushrat011899:
# Objective

- Fixes #16339

## Solution

- Replaced `component_reads_and_writes` and `component_writes` with `try_iter_component_access`.

## Testing

- Ran `dynamic` example to confirm behaviour is unchanged.
- CI

---

## Migration Guide

The following methods (some removed in previous PRs) are now replaced by `Access::try_iter_component_access`:

* `Access::component_reads_and_writes`
* `Access::component_reads`
* `Access::component_writes`

As `try_iter_component_access` returns a `Result`, you'll now need to handle the failing case (e.g., `unwrap()`). There is currently a single failure mode, `UnboundedAccess`, which occurs when the `Access` is for all `Components` _except_ certain exclusions. Since this list is infinite, there is no meaningful way for `Access` to provide an iterator. Instead, get a list of components (e.g., from the `Components` structure) and iterate over that instead, filtering using `Access::has_component_read`, `Access::has_component_write`, etc.

Additionally, you'll need to `filter_map` the accesses based on which method you're attempting to replace:

* `Access::component_reads_and_writes` -> `Exclusive(_) | Shared(_)`
* `Access::component_reads` -> `Shared(_)`
* `Access::component_writes` -> `Exclusive(_)`

To ease migration, please consider the below extension trait which you can include in your project:

```rust
pub trait AccessCompatibilityExt {
    /// Returns the indices of the components this has access to.
    fn component_reads_and_writes(&self) -> impl Iterator<Item = T> + '_;

    /// Returns the indices of the components this has non-exclusive access to.
    fn component_reads(&self) -> impl Iterator<Item = T> + '_;

    /// Returns the indices of the components this has exclusive access to.
    fn component_writes(&self) -> impl Iterator<Item = T> + '_;
}

impl<T: SparseSetIndex> AccessCompatibilityExt for Access<T> {
    fn component_reads_and_writes(&self) -> impl Iterator<Item = T> + '_ {
        self
            .try_iter_component_access()
            .expect("Access is unbounded. Please refactor the usage of this method to directly use try_iter_component_access")
            .filter_map(|component_access| {
                let index = component_access.index().sparse_set_index();

                match component_access {
                    ComponentAccessKind::Archetypal(_) => None,
                    ComponentAccessKind::Shared(_) => Some(index),
                    ComponentAccessKind::Exclusive(_) => Some(index),
                }
            })
    }

    fn component_reads(&self) -> impl Iterator<Item = T> + '_ {
        self
            .try_iter_component_access()
            .expect("Access is unbounded. Please refactor the usage of this method to directly use try_iter_component_access")
            .filter_map(|component_access| {
                let index = component_access.index().sparse_set_index();

                match component_access {
                    ComponentAccessKind::Archetypal(_) => None,
                    ComponentAccessKind::Shared(_) => Some(index),
                    ComponentAccessKind::Exclusive(_) => None,
                }
            })
    }

    fn component_writes(&self) -> impl Iterator<Item = T> + '_ {
        self
            .try_iter_component_access()
            .expect("Access is unbounded. Please refactor the usage of this method to directly use try_iter_component_access")
            .filter_map(|component_access| {
                let index = component_access.index().sparse_set_index();

                match component_access {
                    ComponentAccessKind::Archetypal(_) => None,
                    ComponentAccessKind::Shared(_) => None,
                    ComponentAccessKind::Exclusive(_) => Some(index),
                }
            })
    }
}
```

Please take note of the use of `expect(...)` in these methods. You should consider using these as a starting point for a more appropriate migration based on your specific needs.

## Notes

- This new method is fallible based on whether the `Access` is bounded or unbounded (unbounded occurring with inverted component sets). If bounded, will return an iterator of every item and its access level. I believe this makes sense without exposing implementation details around `Access`.
- The access level is defined by an `enum` `ComponentAccessKind<T>`, either `Archetypical`, `Shared`, or `Exclusive`. As a convenience, this `enum` has a method `index` to get the inner `T` value without a match statement. It does add more code, but the API is clearer.
- Within `QueryBuilder` this new method simplifies several pieces of logic without changing behaviour.
- Within `QueryState` the logic is simplified and the amount of iteration is reduced, potentially improving performance.
- Within the `dynamic` example it has identical behaviour, with the inversion footgun explicitly highlighted by an `unwrap`.


crates/bevy_ecs/src/query/access.rs
```diff
@@ -761,38 +761,61 @@ impl<T: SparseSetIndex> Access<T> {
...
     }
+}
 
-    /// Returns an iterator over the component IDs that this `Access` either
-    /// writes or can't write.
-    ///
-    /// The returned flag specifies whether the list consists of the components
-    /// that the access *can* write (false) or whether the list consists of the
-    /// components that the access *can't* write (true).
-    pub(crate) fn component_writes(&self) -> (impl Iterator<Item = T> + '_, bool) {
```
bushrat011899, line 791: This didn't _have_ to be removed, but its signature is pretty awful to work with, and the new method shadows this one entirely.

crates/bevy_ecs/src/query/access.rs
```diff
@@ -761,38 +761,61 @@ impl<T: SparseSetIndex> Access<T> {
...
-        (
-            self.component_writes.ones().map(T::get_sparse_set_index),
-            self.component_writes_inverted,
-        )
+/// Error returned when attempting to iterate over items included in an [`Access`]
+/// if the access excludes items rather than including them.
+#[derive(Clone, Copy, PartialEq, Eq, Debug, Display, Error)]
+pub struct UnboundedAccess;
```
bushrat011899, line 802: Could be a `non_exhaustive` `enum` instead, but I just wanted to keep it simple. Open to changing it.

examples/ecs/dynamic.rs
```diff
@@ -150,12 +150,12 @@ fn main() {
...
 
                 query.iter_mut(&mut world).for_each(|filtered_entity| {
-                    #[allow(deprecated)]
                     let terms = filtered_entity
                         .access()
-                        .component_reads_and_writes()
-                        .0
-                        .map(|id| {
+                        .try_iter_component_access()
+                        .unwrap()
```
bushrat011899, line 156: Previously this example assumed that the iteration was always going to be over non-inverted access, which would silently misbehave at runtime if the assumption broke. Instead, the `unwrap()` explicitly crashes the program if that invalid state is reached.

BenjaminBrienen:
I think the function is supposed to stay deprecated, not moved. It's just bevy itself needs to not use the function. It can be removed in 0.16.

chescock (approved):

chescock:
The migration guide is written relative to main.  If this is for 0.15, should it be written relative to 0.14 instead?  In 0.14, there were separate `component_read()`, `component_writes()`, and `component_reads_and_writes()` methods that returned plain `impl Iterator`s, and they returned empty iterators when the result would have been unbounded.

crates/bevy_ecs/src/query/access.rs
```diff
@@ -761,38 +761,61 @@ impl<T: SparseSetIndex> Access<T> {
...
+            let sparse_index = T::get_sparse_set_index(index);
+
+            if self.component_writes.contains(index) {
+                ComponentAccessKind::Exclusive(sparse_index)
+            } else {
+                ComponentAccessKind::Shared(sparse_index)
+            }
+        });
+
+        let archetypical = self
```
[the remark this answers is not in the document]
bushrat011899: Chasing it around, the only items `self.archetypal` contains are components explicitly included in a `Query` with `Has<C>`. As such, I'd expect this list to be very short. Additionally, this method was already "wasting" time, since it iterated all `&mut C` items twice (since they're included in `component_writes` _and_ `component_reads_and_writes`). I imagine the net effect here is zero.

My thinking was I could also remove the `archetypal()` method as well, since it's not being used in engine code (or examples). But I decided to just leave them separate just in case to reduce _some_ of the controversy here.

crates/bevy_ecs/src/query/access.rs
```diff
@@ -761,38 +761,61 @@ impl<T: SparseSetIndex> Access<T> {
...
-                .map(T::get_sparse_set_index),
-            self.component_read_and_writes_inverted,
-        )
+    /// Returns `Err(UnboundedAccess)` if the access is unbounded.
+    /// This typically occurs when an [`Access`] is marked as accessing all
+    /// components, and then adding exceptions.
+    pub fn try_iter_component_access(
+        &self,
+    ) -> Result<impl Iterator<Item = ComponentAccessKind<T>> + '_, UnboundedAccess> {
+        if self.component_writes_inverted || self.component_read_and_writes_inverted {
```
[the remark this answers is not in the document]
bushrat011899: Yeah I actually considered changing the function signature to take an iterator of "source" components, so you could always get an iterator since that source would bound the unbounded cases. But that wouldn't have worked for some of the engine code, so I decided to just pessimistically fail instead. In the future we could update the method to return a meaningful result if there's desire to.

bushrat011899:
> I think the function is supposed to stay deprecated, not removed. It's just bevy itself needs to not use the function. It can be removed in 0.16.

I decided to remove them entirely, since they don't actually work as users expect them to in their current form. Basically, the breaking change already happened, so we might as well see it through to a meaningful replacement.

bushrat011899:
> The migration guide is written relative to main. If this is for 0.15, should it be written relative to 0.14 instead? In 0.14, there were separate `component_read()`, `component_writes()`, and `component_reads_and_writes()` methods that returned plain `impl Iterator`s, and they returned empty iterators when the result would have been unbounded.

I've updated the migration guide to explicitly reference the 0.14 state of these methods, thanks for calling that out! To help, I've provided an implementation of an extension trait which should allow the migration to be a drop-in replacement.

mockersf:
I would prefer to just undeprecate it for the 0.15, to have more time for a proper fix https://github.com/bevyengine/bevy/pull/16357

bushrat011899:
Closed in favour of #16357

mockersf:
#16537 is not a fix for the underlying issue, just a way to make it less needed for the 0.15, and this PR seemed a possibility to fix it for longer term

hymm (commented):
Could we add a few tests since this is not used internally anymore?

crates/bevy_ecs/src/query/access.rs
```diff
@@ -761,38 +761,61 @@ impl<T: SparseSetIndex> Access<T> {
...
+            return Err(UnboundedAccess);
```
hymm, line 773: should we return component_writes_inverted and component_read_and_writes_inverted as part of the error? That way if the user has the archetype info they can figure out what components are available.

alice-i-cecile:
@bushrat011899 I think this is a good idea, but it needs a bit of testing and merge conflict resolution still.

crates/bevy_ecs/src/query/access.rs
bushrat011899: I've updated `UnboundedAccess` to include booleans for whether the reads and/or writes are inverted. I _could_ extend it to also include the iterator, which would allow users to further investigate the unbounded access themselves, but I'm not sure if that's what you meant so I'm erring in the smaller side.

bushrat011899:
Added some unit tests and an example in the documentation. I've also improved the information provided by the `UnboundedAccess` error, and resolved merge conflicts.

crates/bevy_ecs/src/query/access.rs
```diff
@@ -773,38 +774,97 @@ impl<T: SparseSetIndex> Access<T> {
...
-    pub(crate) fn component_writes(&self) -> (impl Iterator<Item = T> + '_, bool) {
-        (
-            self.component_writes.ones().map(T::get_sparse_set_index),
-            self.component_writes_inverted,
-        )
+/// Error returned when attempting to iterate over items included in an [`Access`]
+/// if the access excludes items rather than including them.
+#[derive(Clone, Copy, PartialEq, Eq, Debug, Error)]
+#[error("Access is unbounded")]
+pub struct UnboundedAccess {
```
hymm, line 843: Do we postfix our errors with `Error`?

```suggestion
pub struct UnboundedAccessError {
```

crates/bevy_ecs/src/query/builder.rs
```diff
@@ -81,15 +81,14 @@ impl<'w, D: QueryData, F: QueryFilter> QueryBuilder<'w, D, F> {
                 .is_some_and(|info| info.storage_type() == StorageType::Table)
         };
 
-        let (mut component_reads_and_writes, component_reads_and_writes_inverted) =
-            self.access.access().component_reads_and_writes();
-        if component_reads_and_writes_inverted {
+        let Ok(component_accesses) = self.access.access().try_iter_component_access() else {
```
hymm, line 84: Is this right? This logic is slightly different now as it'll return false if component_writes_inverted is true now too. Compared to before where it was just when `component_read_and_writes_inverted` was true.

crates/bevy_ecs/src/query/builder.rs
```diff
@@ -81,15 +81,14 @@ impl<'w, D: QueryData, F: QueryFilter> QueryBuilder<'w, D, F> {
...
+            // Access is unbounded, pessimistically assume it's sparse.
             return false;
-        }
+        };
 
-        component_reads_and_writes.all(is_dense)
-            && self.access.access().archetypal().all(is_dense)
-            && !self.access.access().has_read_all_components()
```
hymm, line 92: why is the check for `has_read_all_components` removed?

crates/bevy_ecs/src/query/builder.rs
chescock: `reads_and_writes` is a superset of `writes`, so I think if `component_writes_inverted` is true then `component_read_and_writes_inverted` is guaranteed to be true.  We could probably simplify `try_iter_component_access()` by only checking `component_read_and_writes_inverted`.  

For this particular case, I'd like to just remove the call!  #16396 would do that, letting us use dense iteration in more cases.

bushrat011899: Yes from my testing `component_read_and_writes_inverted` is true if `component_writes_inverted` is true, making it a superset. I've implemented the simplification, added a unit test to explicitly demonstrate this relationship, and added some documentation too.

crates/bevy_ecs/src/query/builder.rs
bushrat011899: I think that was a mistake when I was resolving conflicts, good catch!

chescock: For what it's worth, `fn has_read_all_components()` is just `component_read_and_writes_inverted && component_read_and_writes.is_clear()`, so it was already covered by the `try_iter_component_access()` check.

PR #16357, opened 2024-11-11
undeprecate `component_reads_and_writes`

### LINKED ISSUE #16339: Migrate `component_reads_and_writes`

mockersf:
# Objective

- Does not correct #16339 but takes it out of the 0.15 milestone

## Solution

- Make it future us problem instead of solving it now

### DIFF
--- a/crates/bevy_ecs/src/query/access.rs
+++ b/crates/bevy_ecs/src/query/access.rs
@@ -772,7 +772,7 @@ impl<T: SparseSetIndex> Access<T> {
     /// `Access`, it's not recommended. Prefer to manage your own lists of
     /// accessible components if your application needs to do that.
     #[doc(hidden)]
-    #[deprecated]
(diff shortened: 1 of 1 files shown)


alice-i-cecile (approved):
I'm not thrilled with shipping a known bug, but fixing this is not something I'd like to rush. And simply deprecating with no replacement isn't acceptable to me either.

cart (approved):

BenjaminBrienen (approved):

bushrat011899 (approved):
Not ideal but I agree, 0.15 is already in the RC stage and this needs design work.

PR #16359, opened 2024-11-11
Update rodio requirement from 0.19 to 0.20

mnmaita:
# Objective

- Supersedes #16344

## Solution

- Updated `rodio` version requirement to `0.20`.
- [Changelog](https://github.com/RustAudio/rodio/blob/master/CHANGELOG.md)

## Testing

- CI checks.


mnmaita:
The `check-advisories` failure is unrelated and also currently present in `main`. It will be fixed when `notify-types` gets bumped to v2.0.

BenjaminBrienen (approved):

alice-i-cecile:
Bumping to 0.16: this shouldn't be shipped without testing.

Issue #16354, opened 2024-11-11
`cosmic_text` font fallback causes inconsistent text rendering results


TurtIeSocks:
## Bevy version

Bevy: Latest `main`
Cargo: `cargo 1.82.0 (8f40fc59f 2024-08-21)`
OS: Windows 11

```ignore
`AdapterInfo { name: "NVIDIA GeForce RTX 4090", vendor: 4318, device: 9860, device_type: DiscreteGpu, driver: "NVIDIA", driver_info: "566.03", backend: Vulkan }`
```

## What you did

I discovered this issue while working on an i18n crate for Bevy that loads dynamic fonts based on the selected locale.

I put together a minimal reproduction repo here:
https://github.com/TurtIeSocks/bevy_font_bug

## What went wrong

While working on this crate I noticed that if a font wasn't using the dynamic font component, it was still able to successfully render characters from the fonts that I had used in other text components, even though the component in question was only using the default font included with Bevy. After putting together the minimal reproduction repo and taking a closer look at the debug logs, it seems that `cosmic_text` is caching fonts to use as a fallback when the given font does not have support for the necessary characters.

### Expectation

Texts should (?) render with the font that is (or isn't) provided, 

### What Happened

Texts are automatically selecting previously used fonts as fallbacks. 

## Additional information

These logs are taken from the minimal reproduction repo linked above:

### Logs from first render (when the font was not cached)

```
2024-11-11T17:18:24.216542Z DEBUG cosmic_text::font::fallback: failed to find family 'Segoe UI'
2024-11-11T17:18:24.216621Z DEBUG cosmic_text::font::fallback: failed to find family 'Segoe UI Emoji'
2024-11-11T17:18:24.218546Z DEBUG cosmic_text::font::fallback: failed to find family 'Segoe UI'
2024-11-11T17:18:24.218683Z DEBUG cosmic_text::font::fallback: failed to find family 'Segoe UI Emoji'
2024-11-11T17:18:24.218804Z DEBUG cosmic_text::font::fallback: failed to find family 'Segoe UI Symbol'
2024-11-11T17:18:24.218902Z DEBUG cosmic_text::font::fallback: failed to find family 'Segoe UI Historic'
2024-11-11T17:18:24.219010Z DEBUG cosmic_text::font::fallback: Failed to find any fallback for [Han] locale 'en-US': '界'
```

### Logs from the second render 
```
[...]

PR #14150, opened 2024-07-05
Expose loading system fonts

Dimchikkk:
# Objective

Expose `cosmic-text` functionality to load system fonts.

## Solution

Add `load_system_fonts` to `TextPlugin` struct.

## Testing

`cargo r --example text --release` by setting `load_system_fonts` to true emoji is rendered correctly:

https://github.com/bevyengine/bevy/assets/111751109/63b25a6d-944b-43f9-8dac-da6de3e08809

---

## Migration Guide

TextPlugin started to have a field, so in places where TextPlugin was used have to use `bevy_text::TextPlugin::default()` or create a struct with `load_system_fonts` field.

PR #16365, opened 2024-11-13
Adds basic support for working with system fonts

### LINKED ISSUE #1325: Support using system fonts
Currently, to display text in bevy UI, a font file needs to be loaded as an asset.

Considering that bevy UI should be usable for general GUI applications (such as the future bevy editor) as well as games, it should support using fonts from the operating system. There should be a way to get the standard fonts that other GUI apps on the user's system use, or other installed fonts.

bushrat011899:
# Objective

- Fixes #1325
- Contributes to the aside at the end of #16354

## Solution

- Created a new feature, `system_font`, which will instruct `cosmic_text` to begin loading system fonts on startup.
- Added a new system parameter, `FontLibrary`, which allows for convenient searching of all loaded fonts, returning a `Handle<Font>` if you have found a suitable candidate.
- Added a new example, `system_fonts`, which demonstrates searching for and using system fonts.

## Testing

- CI
- `system_fonts` example

---

## Showcase

Users now have access to the `FontLibrary` system parameter, which allows searching for available fonts. This helps with using the new `system_fonts` feature, which will load system fonts at runtime.

_In the below snippets, `fonts` is of type `FontLibrary`_

```rust
// Find a monospaced font
let font: Option<Handle<Font>> = fonts.find(|font| font.monospaced);

// Find a font with a particular name
let font: Option<Handle<Font>> = fonts.find(|font| {
    let family_name = &font.families[0].0;
    family_name.contains("Courier")
});
```

system_font_demo

## Notes

- ~~Loading system fonts increases startup time by a few seconds. I suspect this is platform dependent, and is largely out of our control.~~ The startup cost is now negligible (I cannot measure the difference, much less than a second), since moving system font loading onto a background task. I do believe it's worth keeping this behind a feature gate as it's already organised this way, and there _may_ be a performance cost on some platforms. More options for the user is better IMO.
- If the `system_fonts` feature is enabled on a platform `cosmic_text` does not support, it is simply a no-op, making the feature safe to enable in all cases.
- The `FontLibrary` system param could use some more methods, such as an iterator. I have deliberately only exposed a `find` method since we need to do some work with the `Assets<Font>` to ensure it's available for the `TextPipeline`.
- The `Font` type assumes every font will have binary data associated with it (the `data` field). For system fonts this is not the case. I'm simply inserting no data (empty slice) for these fonts. It might be prudent to update `Font` to be an `enum` instead to make this more clear.
- Font families are still handled the same way Bevy currently does, only selecting index 0. Changing this is out of scope for this PR.


crates/bevy_text/src/pipeline.rs
```diff
@@ -32,10 +32,22 @@ pub struct CosmicFontSystem(pub cosmic_text::FontSystem);
...
-        let locale = sys_locale::get_locale().unwrap_or_else(|| String::from("en-US"));
-        let db = cosmic_text::fontdb::Database::new();
-        // TODO: consider using `cosmic_text::FontSystem::new()` (load system fonts by default)
-        Self(cosmic_text::FontSystem::new_with_locale_and_db(locale, db))
+        #[cfg(feature = "system_font")]
+        return {
+            let mut font_system = cosmic_text::FontSystem::new();
+
+            // This is a no-op on unsupported platforms.
+            font_system.db_mut().load_system_fonts();
```
mockersf, line 40: I suppose this is the one taking time? Is it possible to run this in an asynchronous way, or in another thread?

TurtIeSocks:
Awesome! I swapped to this branch locally with my [playground](https://github.com/TurtIeSocks/bevy_simple_i18n/blob/main/examples/changing_locale.rs) example and every locale now displays without the need to load any external font files 🎊 

I suppose this does increase the need for users to have the ability to control `cosmic_text` fallbacks. I should have time later in the week to dig into that and see what's viable.

bushrat011899:
> If it's enabled by default...

I've left it is a feature that's disabled by default. To enable system fonts, you enable the `system_font` feature in Bevy. This feature does have a startup time cost and I completely agree that the functionality should be opt-in, at least while it has a noticeable impact.

TurtIeSocks:
I've just noticed #14150, was that effort abandoned?

bushrat011899:
I never saw that! Looks like it wasn't linked to the original issue. I'm happy to incorporate changes from that PR, or close this out in favour of that one, I'll leave that up to the reviewers.

crates/bevy_text/src/pipeline.rs
JeanMertz: Looking at cosmic-text, could we use two [`Database`s](https://docs.rs/fontdb/0.16.2/fontdb/struct.Database.html), one for the system fonts, populated in a different thread, and the main one that loads the application-provided fonts. Then, once the system fonts are loaded, send the database back to the main thread, and merge that one into the main one.

crates/bevy_text/src/font_library.rs
```diff
@@ -0,0 +1,50 @@
...
+pub struct FontLibrary<'w> {
+    text_pipeline: ResMut<'w, TextPipeline>,
+    font_system: Res<'w, CosmicFontSystem>,
+    font_assets: ResMut<'w, Assets<Font>>,
+}
+
+impl FontLibrary<'_> {
+    /// Find a [`Font`] based on the provided criteria.
+    /// You are given access to the font's [`FaceInfo`] to aid with selection.
+    pub fn find(&mut self, mut f: impl FnMut(&FaceInfo) -> bool) -> Option<Handle<Font>> {
```
rparrett, line 24: There may be a subtle performance footgun here for users.

Am I right that calling `find` multiple times for the same font results in multiple asset handles pointing to the same font?

I think this would result in separate font atlases being created and possibly hamper text batching.

Not sure exactly what to do about this, but some ideas:
- Maybe there's a better name for `find` that reflects this behavior
- The behavior should probably be documented
- Maybe the example could be reworked to show something slightly more practical, including caching the font handles?
- Add an ID -> handle mapping? Not sure if that's actually possible

examples/2d/system_fonts.rs
```diff
@@ -0,0 +1,74 @@
...
+    commands.spawn((
+        Text2d::new("Press Space to change the font"),
+        TextFont {
+            // We load a fall-back font, since we don't know that every platform
+            // even has system fonts to use.
+            font: asset_server.load("fonts/FiraSans-Bold.ttf"),
+            font_size: 50.0,
+            ..default()
+        },
+        TextLayout::new_with_justify(JustifyText::Center),
```
rparrett, line 32: Don't think this is doing anything for an unbounded `Text2d`.

crates/bevy_text/src/font_library.rs
```diff
@@ -0,0 +1,50 @@
...
+use bevy_ecs::system::{Res, ResMut, SystemParam};
+use cosmic_text::fontdb::FaceInfo;
+
+use crate::{CosmicFontSystem, Font, TextPipeline};
+
+/// Provides a method for finding [fonts](`Font`) based on their [`FaceInfo`].
+///
+/// Note that this is most useful with the `system_font` feature, which exposes
+/// fonts installed on the end-users device. Without this feature, the only fonts
+/// available are ones explicitly added within Bevy anyway.
```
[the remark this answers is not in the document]
bushrat011899: Fair, I could improve the wording here. What I'm trying to say is this allows finding a font and getting a `Handle` to it. But if you have system fonts disabled, it'll only show fonts which you added yourself, and therefore you already have a `Handle` to those fonts.

crates/bevy_text/src/font_library.rs
bushrat011899: Agreed. I might resolve this by adding a `HashMap<FontId, Weak<Font>>` so I can reuse `Handle` if it's already created and still active.

crates/bevy_text/src/font_library.rs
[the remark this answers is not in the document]
rparrett: Ahh, okay.

crates/bevy_text/src/pipeline.rs
bushrat011899: Ok what I've done is moved loading the system font onto an `AsyncCompteTaskPool` task. Since I expect this to take multiple frames, and it's entirely in blocking IO, I believe this is the correct pool to use. Once the task resolves, I take these two `Database`s and merge them by simply cloning `FaceInfo`s from one to the other (can use `mem::swap` to copy from the smaller to the larger). I then send a `SystemFontsAvailable` event once this process is completed. It still happens immediately at startup, but it wont prevent other parts of the application from progressing at least.

rparrett:
IMO, the current example could be made more practical for users seeking to use this feature (and also for automated testing). It would be nice to show a more common scenario than cycling through every font one at a time, and to demonstrate how to react to the system fonts becoming available.

rparrett (approved):

examples/2d/system_fonts.rs
```diff
@@ -0,0 +1,73 @@
...
+            name = String::from(family_name);
+            true
+        }) else {
+            *query.0 = Text2d::new("No more fonts to show!");
+            return;
+        };
+
+        *query.0 = Text2d::new(name);
+        query.1.font = font;
+    }
```
krunchington, line 72: Nitpick - I think there's an idiomatic way to avoid some returns here.
```suggestion
    if keyboard_input.just_pressed(KeyCode::Space) {
        let mut name = default();
        
        // The primary way to find a font is through FontLibrary::find,
        // which iterates over all loaded fonts, and returns the first that
        // satisfies the provided predicate.
        // In this example, we're just looking for any font we haven't already
        // shown.
        
        let next_font = fonts.find(|font|{
            let family_name = font.families[0].0.as_str();
            if used.contains(family_name) {
                false
            } else {
                used.insert(Box::from(family_name));
                name = String::from(family_name);
                true
            }
        });
        
        match next_font {
            Some(font) => {
                *query.0 = Text2d::new(name);
                query.1.font = font;
            },
            None => {
                *query.0 = Text2d::new("No more fonts to show!");
            },
        }
    }
```

krunchington: Actually I think for the end, `clippy` usually prefers `if let ... else` also but I did this in github comments not my editor XD

crates/bevy_text/src/font_library.rs
```diff
@@ -0,0 +1,67 @@
...
+
+#[derive(Resource, Default)]
+pub(crate) struct FontIdToAssetMap {
+    map: HashMap<ID, AssetId<Font>>,
+}
+
+impl FontLibrary<'_> {
+    /// Find a [`Font`] based on the provided criteria.
+    /// You are given access to the font's [`FaceInfo`] to aid with selection.
+    pub fn find(&mut self, mut f: impl FnMut(&FaceInfo) -> bool) -> Option<Handle<Font>> {
```
krunchington, line 33: It might be useful to return the `FaceInfo` as part of the `Option` also- noticed in the example you had to use a mutable local variable to store data from the FaceInfo.  Maybe something like `Option<(Handle<Font>, &FaceInfo)>`?

krunchington: Also worth noting that this function signature, through the predicate, "leaks" the abstraction of a `cosmic_text::fontdb::FaceInfo`.  That might be intended and fine, but worth pointing out if it is preferable to keep the bevy interface more opaque (which could make it easier to replace `cosmic_text` in the future if necessary)

examples/2d/system_fonts.rs
krunchington: An example alternate approach that I think ends up a lot more readable, if as in my other comment more info was returned from `.find`
```rs
    if keyboard_input.just_pressed(KeyCode::Space) {
        // The primary way to find a font is through FontLibrary::find,
        // which iterates over all loaded fonts, and returns the first that
        // satisfies the provided predicate.
        // In this example, we're just looking for any font we haven't already
        // shown.
        
        let next_font = fonts.find(|font|{
            let family_name = font.families[0].0.as_str();
            used.contains(family_name)
        });
        
        if let Some((handle, face_info) = next_font {
            let family_name = face_info.families[0].0.as_str();
            *query.0 = Text2d::new(family_name);
            query.1.font = handle;
        } else {
            *query.0 = Text2d::new("No more fonts to show!");
        }
    }
```

asgeo1:
Any chance this can be added to bevy 0.17 or 0.18 milestones? Really looking forward to this functionality and hoping it's still on the roadmap.

dsgallups (approved):
Looks great!

PR #16366, opened 2024-11-13
Add `http` & `https` asset sources

### LINKED ISSUE #16307: Feature: Upstream `bevy_web_asset`, allowing assets loaded via `http`

Being able to load assets from `http` and `https` urls. 
This is particulary relevant for examples, files that are too big for source control are often stored in external zip files requiring awkward download & extract workflows for people just getting started with a crate.
It also makes both wasm and native binaries more portable, as the additional 'assets' folder isnt required to always be placed alongside them, a big win for game jams etc.

For url assets to 'just work', meaning if a path starts with `http/https` the asset will be loaded via the http client:

```rust
# intercepted by http client
asset_server.load("http://example.com/fox.glb");
# https also accepted
asset_server.load("https://example.com/fox.glb");
# load normally
asset_server.load("/fox.glb");
```

### Keep this as a third party crate

Pros: 
- No changes/maintenance needed for bevy core

Cons: 
- Bevy examples face same pain point mentioned above
- No blessed solution for this issue in ecosystem crates

### Considerations

- Disk Caching: Browsers usually quietly (and unpredictably) cache files on disk so that they arent fetched on every reload, we will need something similar for native builds, and possibly a more predictable implementation on wasm builds, for example using IndexedDB. 
- Choice of HTTP client: It makes sense to go with the same client used elsewhere in Bevy, ie `bevy_remote`, unless it is missing some critical feature.
- Feature Flags: Because this adds a http client to `bevy_asset` it should be behind a feature flag, we will need to decide whether this is to be included by default, in my opinion it should be.
- CORS: files fetched by native builds dont need cors enabled but it will be required for wasm builds to work, this creates an inconsistency that we may want to flag on native builds if the headers are not present.
- Work: @johanhelsing has [mentioned](https://discord.com/channels/691052431525675048/749332104487108618/1303853442304839691) that they dont have the capacity to work on this at the moment. I'm happy to get the ball rolling when I get some time, though this will be my most contraversial PR so I may need some mentorship.

`bevy_web_asset` is a tiny crate, with <250 lines of code. I guess the most work required here will be a caching solution.

mrchantey:
# Objective

- Fixes #16307

## Solution

- Add `http` & `https` asset sources

## Testing

- Verify visually by running `cargo run --example http_source`

---

## Showcase

Bevy now supports assets loaded via url!

Add the `http_source` feature to load assets served over the web, with support for both native and wasm targets. On native targets the `http_source_cache` feature will use a basic caching mechanism to avoid repeated network requests.

```rust
  // Simply use a url where you would normally use an `assets` path
  let handle = asset_server.load("https://raw.githubusercontent.com/bevyengine/bevy/refs/heads/main/assets/branding/bevy_bird_dark.png");
  commands.spawn(Sprite::from_image(handle));
```


mrchantey:
It looks like `surf` and `http-cache-surf` have failed licence, vunerability & unmaintaned checks. Does that mean they're not allowed as first party dependencies?

BenjaminBrienen:
It's possible to add an exception for the licenses, but unmaintained and vulnerable is pretty bad.

mnmaita:
This could be done with something like `reqwest` maybe to avoid using the unmaintained crates. I successfully ran the example with it but my code is probably bad 😅. I can still push a PR as a showcase in case it serves as inspiration.

MalekiRe (approved):
Yep this looks awesome to me! Also I wasn't aware of the `surf` crate; I'll have to keep it in mind for any bevy dashboards I make in the future

mrchantey:
> I can still push a PR as a showcase in case it serves as inspiration.

@mnmaita actually that would be great, I've replaced `surf` and `http-cache-surf` with `reqwest` line for line but am a bit stuck at the moment with reqwest demanding to be run inside a tokio runtime.

```rust
// crates/bevy_asset/src/http_source.rs#134

    let client = reqwest_middleware::ClientBuilder::new(Client::new())
        .with(Cache(HttpCache {
            mode: CacheMode::Default,
            manager: CACacheManager::default(),
            options: HttpCacheOptions::default(),
        }))
        .build();

    let response = ContinuousPoll(client.get(str_path).send())
        .await

```

jf908:
FWIW surf has a bug where any erroring http response >8kb stops every subsequent request from succeeding and is unlikely to ever be fixed which made me unable to use bevy_web_asset so I would personally appreciate a switch 😄.

If you can't get reqwest to work perhaps hyper would suffice since it's already in the dep tree for BRP? Perhaps too low level for this.

mrchantey:
The council of bevy async wizards [has spoken](https://discord.com/channels/691052431525675048/692572690833473578/1307168563190173758) and we're going with `ureq`. It doesn't currently have a `http-cache` wrapper but I've [opened an issue](https://github.com/06chaynes/http-cache/issues/91) and the author might look into creating one when they have time. I guess its worth waiting a bit to see if `ureq` will get first party support before rolling our own `http-cache` wrapper, in which case this pr is ready for review :)

A few notes:
- We're still failing `check-advisories` & `check-licences`
- I used gh codespaces so wasnt able to `cargo run --example http_source`, can somebody please try running that?

jf908:
> * I used gh codespaces so wasnt able to `cargo run --example http_source`, can somebody please try running that?

Tested on Windows 11. The example prints a 404 error when the .meta file fails to load and doesn't show the sprite.

If I configure AssetPlugin to AssetMetaCheck::Never, it runs correctly.

I believe the asset is still supposed to load successfully even if the .meta file 404s so I think this is a bug?

BenjaminBrienen:
> We're still failing check-advisories & check-licences
I used gh codespaces

You can ignore that. It isn't a required check. Once those 2 small issues above are fixed, this should be in a pretty good state, I think.

mrchantey:
Ok I can confirm the example is working in wasm and native builds.

I'm a bit uneasy merging without native http caching. One of the primary use-cases for the feature is large assets and the current implementation wil re-download them every time the app is run. Maybe a stupidly naive cache, ie one that never automatically invalidates, is better than nothing?

In the long run it looks like a [`http-cache-ureq`](https://github.com/06chaynes/http-cache/issues/91) implementation would be ideal.

mrchantey:
A simple native cache has been added and is ready for review

crates/bevy_asset/src/http_source.rs
```diff
@@ -0,0 +1,250 @@
...
+            let mut buffer = Vec::new();
+            reader.read_to_end(&mut buffer)?;
+
+            http_asset_cache::save_to_cache(str_path, &buffer)?;
+
+            Ok(Box::new(VecReader::new(buffer)))
+        }
+        // ureq considers all >=400 status codes as errors
+        Err(ureq::Error::Status(code, _response)) => {
+            if code == 404 {
```
BenjaminBrienen, line 90: Adding a special case for Unauthorized might be useful. Just a thought, probably fine without.

crates/bevy_asset/src/http_source.rs
```diff
@@ -0,0 +1,250 @@
...
+    use std::fs::{self, File};
+    use std::io::{self, Read, Write};
+    use std::path::PathBuf;
+
+    const CACHE_DIR: &str = ".http-asset-cache";
+
+    fn url_to_filename(url: &str) -> String {
+        // Basic URL to filename conversion
+        // This is a naive implementation and might need more robust handling
+        url.replace([':', '/', '?', '=', '&'], "_")
```
BenjaminBrienen, line 162: If this is just for caching, why not just hash it and use that as the filename?

crates/bevy_asset/src/http_source.rs
mrchantey: Ok Ive changed non-404 status codes to `AssetReaderError::HttpError` variant, consistent with `HttpWasmAssetReader`, making it easier for error handlers to directly check the status code.

BenjaminBrienen (approved):

crates/bevy_asset/src/http_source.rs
```diff
@@ -0,0 +1,242 @@
...
+        path: &'a Path,
+    ) -> Result<Box<PathStream>, AssetReaderError> {
+        Err(AssetReaderError::NotFound(self.make_uri(path)))
+    }
+}
+
+/// A naive implementation of an HTTP asset cache that never invalidates.
+/// `ureq` currently does not support caching, so this is a simple workaround.
+/// It should eventually be replaced by `http-cache` or similar, see [tracking issue](https://github.com/06chaynes/http-cache/issues/91)
+mod http_asset_cache {
```
jf908, line 142: I'm a bit sceptical about the caching implementation, mostly because I am biased and my game's assets are small enough not to need caching and I don't want my game to create any unnecessary folders. But also because the default behaviour creating a hidden folder might mean some users may not even realise the assets are being cached, especially since it isn't documented on any high level docs of this feature.

It's clearly useful but IMO caching should be disabled by default as it's something that I think people will reach for if they need it. I don't know if its worth blocking on this but I think it deservers a bit more discussion at least😄

jf908 (commented):
I have a couple of concerns but overall looks good. I really want this so I appreciate you spearheading it!

crates/bevy_asset/src/http_source.rs
mrchantey: Yep good point ive exposed a `http_source_cache` top level feature.

Cargo.toml
```diff
@@ -1637,12 +1643,23 @@ path = "examples/asset/extra_source.rs"
...
+[[example]]
+name = "http_source"
+path = "examples/asset/http_source.rs"
+doc-scrape-examples = true
+
+[package.metadata.example.http_source]
+name = "HTTP Asset Source"
+description = "Load an asset from a http source"
+category = "Assets"
+wasm = true
```
jf908, line 1661:
```suggestion
wasm = true
required-features = ["http_source"]
```

jf908 (approved):
http_source should be added a required feature for the example but otherwise looks good to me

crates/bevy_asset/src/http_source.rs
```diff
@@ -0,0 +1,245 @@
...
+        fs::create_dir_all(CACHE_DIR).ok();
+
+        let mut cache_file = File::create(&cache_path)?;
+        cache_file.write_all(data)?;
+
+        Ok(())
+    }
+}
+
+#[cfg(test)]
```
mockersf, line 187: in all the tests, don't use domains owned by someone. either use domains owned by Bevy or examples domains (example.org or example.com, see https://www.iana.org/domains/reserved)

crates/bevy_asset/src/http_source.rs
```diff
@@ -0,0 +1,244 @@
...
+            http_asset_cache::save_to_cache(str_path, &buffer)?;
+
+            Ok(Box::new(VecReader::new(buffer)))
+        }
+        // ureq considers all >=400 status codes as errors
+        Err(ureq::Error::Status(code, _response)) => {
+            if code == 404 {
+                Err(AssetReaderError::NotFound(path))
+            } else {
+                Err(AssetReaderError::HttpError(code))
```
killercup, line 95: Maybe helpful to add some more error context here by also capturing the path in these cases?

crates/bevy_asset/Cargo.toml
```diff
@@ -63,6 +65,7 @@ js-sys = "0.3"
 
 [target.'cfg(not(target_arch = "wasm32"))'.dependencies]
 notify-debouncer-full = { version = "0.4.0", optional = true }
+ureq = { version = "2.10", optional = true }
```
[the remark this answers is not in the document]
jf908: The JSON feature is only default on ureq 3.0 rc on the main branch so it’s not included currently. 2.0’s default is just tls and gzip. No harm in being explicit or future proof though.

crates/bevy_asset/src/http_source.rs
mrchantey: Its a good idea that should probably be applied to all variants of `AssetReaderError`. The change is straightforward but touches about a dozen files and causes things like `PartialEq` to work differently, could be more cleanly done in a seperate pr?

crates/bevy_asset/Cargo.toml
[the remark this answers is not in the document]
mrchantey: k ive set `default-features=false` with `"brotli","gzip","tls"` added

crates/bevy_asset/src/http_source.rs
killercup: Agreed. You could cheat this by wrapping the top-most call in an IO error with the path in the description and the original error nested.

killercup (approved):

mockersf:
a decision on adding unicode 3.0 and MPL 2.0 need to be taken before merging this PR

alice-i-cecile:
> a decision on authorising licenses unicode 3.0 and MPL 2.0 need to be taken before merging this PR

Previous Unicode licenses are already allowed and it is extremely permissive.

MPL 2.0 is already in tree under off-by-default feature flags for various audio formats. This doesn't strike me as any different, and MPL is not dangerously viral to use as a dependency in the same way as e.g. AGPL.

mockersf:
then they should be added to the license check config, otherwise every other pr will have failed ci

mrchantey:
Last license issue is ring (ureq dep) which has a [custom license](https://github.com/briansmith/ring/blob/main/LICENSE):
```
error[unlicensed]: ring = 0.17.8 is unlicensed
  ┌─ registry+https://github.com/rust-lang/crates.io-index#ring@0.17.8:2:9
  │
2 │ name = "ring"
  │         ━━━━ a valid license expression could not be retrieved for the crate
3 │ version = "0.17.8"
4 │ license = ""
  │            ─ license expression was not specified
5 │ license-files = [
6 │     { path = "/home/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/ring-0.17.8/LICENSE", hash = 0xbd0eed23, score = 0.66, license = "OpenSSL" },
  │                                                                                                                                  ──── low confidence in the license text

```

mockersf:
could you try to remove the dependency on ring by disabling default feature also on the dev-dependency to ureq?

mockersf:
ring comes from rustls support, which is enabled by default with the tls feature

This means that we force our users to use rustls, but also that someone has to properly review that license

mockersf:
from the license:
> 3. All advertising materials mentioning features or use of this
> *    software must display the following acknowledgment:
> *    "This product includes software developed by the OpenSSL Project
> *    for use in the OpenSSL Toolkit. (http://www.openssl.org/)"


Seems kind of a pain. I think it means we won't be able to talk about the engine being able to download things from https without copying that

alice-i-cecile:
Looking at the [ring license](https://github.com/briansmith/ring/blob/main/LICENSE), it appears to be *mostly* MIT.

However: 

>  * 3. All advertising materials mentioning features or use of this software
>  *    must display the following acknowledgement:
>  *    "This product includes cryptographic software written by
>  *     Eric Young (eay@cryptsoft.com)"

This clause is *super* annoying. It looks like it's a very old amateur license from 1995 or so. The practical risk is low, but we should not allow this license. Can we change features around to avoid `ring`?

alice-i-cecile:
See https://github.com/briansmith/ring/issues/1827 for a tracking issue for the upstream licensing stuff. We very much shouldn't try to wait on that.

mrchantey:
From what i can gather on these ureq issues, the new rustls default is aws-lc-rs instead of ring. It seems to have better licensing but worse devx, requiring non-rust dependencies for windows builds.
https://github.com/algesten/ureq/issues/751
https://github.com/algesten/ureq/issues/765

alice-i-cecile:
Bah, okay. Can we ship without https support by default and leave the choice up to the end user? I don't particularly care about encrypting e.g. example assets TBH.

mrchantey:
Sounds good, i guess most users of this feature will need it because just about everything is https these days.

The workaround is in the Cargo.toml `dev-dependencies` because the http_source example uses `https://raw.githubusercontent/.../bevy_bird_dark.png`.

We can add tls when ring sorts its license out and in the meantime adding https suppport is a one-liner:
```toml
# enable bevy https asset loader
ureq = { version = "*", features = ["tls"] }
```

## `check-licenses`

`check-licenses` still isnt happy, i think it cant distinguish a dev-dependency when its the same crate that we're toggling a feature on:

```
 ├ ring v0.17.8
    ├── rustls v0.23.20
    │   └── ureq v2.12.1
    │       ├── (dev) bevy v0.15.0-dev
    │       └── bevy_asset v0.15.0-dev
```

crates/bevy_asset/src/http_source.rs
```diff
@@ -0,0 +1,248 @@
...
+use crate::io::{AssetSource, PathStream};
+use crate::AssetApp;
+use bevy_app::App;
+use bevy_utils::ConditionalSendFuture;
+use std::path::{Path, PathBuf};
+
+/// Adds the `http` and `https` asset sources to the app.
+/// Any asset path that begins with `http` or `https` will be loaded from the web
+/// via `fetch`(wasm) or `ureq`(native).
+/// Note that the use of secure `https` sources in non-wasm builds requires the following dependency:
```
alice-i-cecile, line 11: This note and its sibling needs a warning about the licensing situation :)

alice-i-cecile (changes requested):
Once there's a note or two about the license in the docs I'm happy to merge this.

mockersf:
I don't think it's worth it to merge this without https support

Nowadays everything is https. hiding the license behind a not enabled feature is not a fix, because for pretty much all use cases it will need to be enabled and the license issue solved

cpu:
>  It seems to have better licensing but worse devx, requiring non-rust dependencies for windows builds.

FWIW the situation here has improved since the linked discussions. Since [v0.23.13](https://github.com/rustls/rustls/releases/tag/v%2F0.23.13) Rustls [uses `aws-lc-rs`](https://github.com/rustls/rustls/blob/3751e24bbc1fef4f8d0bb3711a3be2a3d2a7b793/Cargo.toml#L44) with the [prebuilt-nasm](https://docs.rs/aws-lc-rs/latest/aws_lc_rs/#prebuilt-nasm) feature to deliver equivalent Windows support as ring.

mrchantey:
> FWIW the situation here has improved since the linked discussions

Thats very helpful. it increasingly looks like `aws-lc-rs` is the path forward, im tagging this pr as blocked until the ureq authors make a call on an option to exclude `ring`.

algesten:
I made a PR in ureq that would allow it to be compiled without ring, and to pick any `CryptoProvider` with `rustls`. https://github.com/algesten/ureq/pull/931

The only drawback is that rustls isn't semver 1.x, which means I can't make semver major version guarantees for ureq in this API.

algesten:
@mrchantey before we merge that PR, I just want to double check. We can't guarantee this API under semver for ureq, is that a blocker for you? I.e. hypothetically, if `rustls` breaks compatibility of `CryptoProvider` between current version 0.23 and a future 0.24, we will only bump ureq minor version. Would that be a problem?

mrchantey:
@algesten Thanks for clarifying, not a problem but good to keep in mind, I'll make a note of it in the Cargo.toml.

algesten:
Cool. Thanks! I merge!

mrchantey:
@algesten fyi im getting compile errors, not sure if i set the feature flags up correctly:
- [the error](https://github.com/bevyengine/bevy/actions/runs/12637302601/job/35211216282?pr=16366#step:6:1020)

```toml
ureq = { git = "https://github.com/algesten/ureq", rev = "fd82f2bef3c2b2a8eaeaac9aae7ea06ffe7449d3", optional = true, default-features = false, features = [
  # "rustls",             # compiles ok
  "rustls-no-provider",   # breaks compilation
  "gzip",
  "json",
] }
rustls = { version = "0.23", optional = true, default-features = false, features = [
  "aws_lc_rs",
  "logging",
  "std",
  "tls12",
] }

```

crates/bevy_asset/Cargo.toml
```diff
@@ -65,10 +65,18 @@ js-sys = "0.3"
 
 [target.'cfg(not(target_arch = "wasm32"))'.dependencies]
 notify-debouncer-full = { version = "0.4.0", optional = true }
-ureq = { version = "2.10", optional = true, default-features = false, features = [
-  "brotli",
+ureq = { git = "https://github.com/algesten/ureq", rev = "fd82f2bef3c2b2a8eaeaac9aae7ea06ffe7449d3", optional = true, default-features = false, features = [
+  # "rustls", # compiles ok
+  "rustls-no-provider", # breaks compilation
```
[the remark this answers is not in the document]
algesten: https://github.com/algesten/ureq/pull/935

algesten:
@mrchantey thanks. Fixing it now. See https://github.com/algesten/ureq/pull/935

Also adding the case to CI so it doesn't regress.

algesten:
@mrchantey try 423aa8f05b30799129080e1bbe821ef067b50ed2

mrchantey:
@algesten awesome all working as expected :pray: 

I'll leave the `S-Blocked` tag until ureq 3 is out, there are now just a couple of changes that will need review:
- allow `OpenSSL` license
- The use of `once_cell` to avoid rebuilding the agent for every request:
  ```rust
  static AGENT: once_cell::sync::Lazy<ureq::Agent> = Lazy::new(|| {
    // build agent
  });
  let response = AGENT.get("example.com")...
  ```

algesten:
@mrchantey We published `3.0.0-rc5` if you want to have a proper crate dep while waiting for the real `3.0.0`. The timeline was discussed here: https://github.com/algesten/ureq/issues/920

This release reset the clock on having some "time of quiet" before going full out on 3.0.0, but there's nothing on the radar that needs fixing.

jf908:
I was in the middle of re-reviewing this PR since ureq 3 is out and I realized that aws-lc also has the problematic license terms that @alice-i-cecile pointed out from the OpenSSL license. aws-lc also requires that you have CMake (but not Clang) installed on Windows which is quite annoying, especially because even if you install it in Visual Studio Installer, it doesn't include it in the path unless you use the special developer console shortcut.

[rustls's README](https://github.com/rustls/rustls?tab=readme-ov-file#third-party-providers) has a list of alternative tls implementations but after a quick look I don't see any which are suitable given that they are mostly C bindings and could lead to more compilation difficulties. I'm wondering if it's possible to dynamically link to the OS TLS implementation to overcome license issues but maybe that's also asking for compilation issues.

It would be very disappointing if the outcome of this is that bevy will be incapable of web requests, is the OpenSSL license out of the question?

mrchantey:
> aws-lc also requires that you have CMake (but not Clang) installed on Windows

Just confirming was this not solved with the introduction of [prebuilt nasm](https://github.com/aws/aws-lc-rs/blob/main/aws-lc-rs/README.md#use-of-prebuilt-nasm-objects)?

algesten:
Maybe you already saw, but there are some more 3rd party providers if neither aws-lc-rs nor ring seems right: https://github.com/rustls/rustls?tab=readme-ov-file#third-party-providers

jf908:
> Just confirming was this not solved with the introduction of [prebuilt nasm](https://github.com/aws/aws-lc-rs/blob/main/aws-lc-rs/README.md#use-of-prebuilt-nasm-objects)?

Yep, CMake is still required, everything else seems to be provided.

mrchantey:
@algesten I'm revisiting this to finalize and am now getting a panic which is weird because it seems nothing has changed in the branch, and it happens both for the git rev that was working earlier and for `ureq="3"`, am I doing this correctly? 
Here's an isolated example ive tried with a vanilla `cargo init`


```toml
[package]
name = "hello-ureq"
version = "0.1.0"
edition = "2024"

[dependencies]
ureq = { version = "3", default-features = false, features = [
	"rustls-no-provider",
] }
rustls = { version = "0.23", features = ["aws_lc_rs"] }
```

```rust
fn main() {
    use std::sync::Arc;
    use ureq::Agent;
    use ureq::tls::{TlsConfig, TlsProvider};

    let crypto = Arc::new(rustls::crypto::aws_lc_rs::default_provider());
    let agent = Agent::config_builder()
        .tls_config(
            TlsConfig::builder()
                .provider(TlsProvider::Rustls)
                // requires rustls or rustls-no-provider feature
                .unversioned_rustls_crypto_provider(crypto)
                .build(),
        )
        .build()
        .new_agent();

    println!("gonna call");
    let res = agent.get("https://example.com").call();
    println!("called");
}
// gonna call
//
// thread 'main' panicked at /home/pete/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/ureq-3.0.3/src/tls/rustls.rs:110:9:
// No CryptoProvider for Rustls. Enable the feature `ring` or configure the Agent.

```

jf908:
Not sure how it was working before but I reported this [already](https://github.com/algesten/ureq/issues/970) and @algesten graciously fixed it so I assume it will be available in 3.0.4 whenever it releases!

crates/bevy_asset/Cargo.toml
```diff
@@ -63,6 +65,19 @@ js-sys = "0.3"
 
 [target.'cfg(not(target_arch = "wasm32"))'.dependencies]
 notify-debouncer-full = { version = "0.4.0", optional = true }
+# note that while ureq is semver stable rustls is not, meaning possible breaking changes on minor releases. https://github.com/bevyengine/bevy/pull/16366#issuecomment-2572890794
+ureq = { git = "https://github.com/algesten/ureq", rev = "423aa8f05b30799129080e1bbe821ef067b50ed2", optional = true, default-features = false, features = [
+  "rustls-no-provider",
+  "gzip",
+  "json",
+] }
+once_cell = { version = "1.20", optional = true }
```
jf908, line 74: Instead of using once_cell I believe you can use LazyLock in the new bevy_platform_support crate (which defaults to the std implementation) to avoid a new dep.

algesten:
> Not sure how it was working before but I reported this [already](https://github.com/algesten/ureq/issues/970) and @algesten graciously fixed it so I assume it will be available in 3.0.4 whenever it releases!

I released 3.0.4 just now

mrchantey:
Awesome everything working as expected on ubuntu and wasm, and `OnceCell` has been replaced with `LazyLock`. The two extra bits of work mentioned in this pr are better AssetReadErrors and user-defined crypto providers, perhaps we can create seperate issue for each of those.

alice-i-cecile:
Can I have a summary of the licensing situation as it currently stands? I want to make sure I have all the details straight :)

mrchantey:
@alice-i-cecile this pr adds two licences:
- ureq > webpki-roots: MPL-2.0
- aws-lc-rs > aws-lc-sys: OpenSSL

ci still picks up ring because `bevy_remote/client` is still using default-features. Perhaps in another pr we can expose a general `fetch` function with the `aws-lc-rs` machinery used in this pr.

On another note this pr commit history has gotten super messy, it has a tonne of commits from when i was working out how to do ci checks locally and i think i did a bad merge, maybe I could open up a fresh one and link to it?

alice-i-cecile:
A fresh PR with a link to this one would be great :)

mrchantey:
Closinng in favor of a fresh pr with clean commit history: #17889

PR #16367, opened 2024-11-13
 Use normal constructors for EasingCurve, FunctionCurve, ConstantCurve

cart:
# Objective

We currently use special "floating" constructors for `EasingCurve`, `FunctionCurve`, and `ConstantCurve` (ex: `easing_curve`). This erases the type being created (and in general "what is happening" structurally), for very minimal ergonomics improvements. With rare exceptions, we prefer normal `X::new()` constructors over floating `x()` constructors in Bevy. I don't think this use case merits special casing here.

## Solution

Add `EasingCurve::new()`, use normal constructors everywhere, and remove the floating constructors.

I think this should land in 0.15 in the interest of not breaking people later.

### DIFF
--- a/crates/bevy_animation/src/animation_curves.rs
+++ b/crates/bevy_animation/src/animation_curves.rs
@@ -7,9 +7,9 @@
 //! `Curve<Vec3>` that we want to use to animate something. That could be defined in
 //! a number of different ways, but let's imagine that we've defined it [using a function]:
 //!
-//!     # use bevy_math::curve::{Curve, Interval, function_curve};
+//!     # use bevy_math::curve::{Curve, Interval, FunctionCurve};
 //!     # use bevy_math::vec3;
-//!     let wobble_curve = function_curve(
+//!     let wobble_curve = FunctionCurve::new(
 //!         Interval::UNIT,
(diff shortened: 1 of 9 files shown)


alice-i-cecile (approved):
I think this is clearer, more consistent and easier to discover (since it's automatically linked in the docs in both directions).

BenjaminBrienen (approved):

mweatherley:
I don't feel particularly strongly about this either way, but some notes here about why these are this way:
1. When the `Curve` API was first conceived, most of the curve return types of adaptor methods were erased. I moved away from that later for various reasons, but this is sort of a vestige of that.
2. The return types of these library functions are not really "worth knowing" in a meaningful way; all of their useful methods just derive from them being curves. That is, the `EasingCurve`/`FunctionCurve`/`ConstantCurve` namespace is effectively otherwise unused.
3. It's significantly fewer keystrokes :)

IQuick143 (approved):
I don't see an issue with this, and I like the consistency.
Though I second all that mw said.

MiniaczQ (approved):
Not controversial

NthTensor (approved):

PR #16368, opened 2024-11-13
Add a bindless mode to `AsBindGroup`.

pcwalton:
This adds the infrastructure necessary for Bevy to support *bindless resources*. Bindless resources significantly improve rendering performance by reducing `wgpu` and driver overhead considerably.

TODO: Write more here.

This should be ready except for documentation. I'm submitting it now as a draft since this shouldn't land before 0.15 is released anyway.

## Migration Guide

* `RenderAssets::prepare_asset` now takes an `AssetId` parameter.
* Bin keys now have Bevy-specific material bind group indices instead of `wgpu` material bind group IDs, as part of the bindless change. Use the new `MaterialBindGroupAllocator` to map from bind group index to bind group ID.

### DIFF
--- a/crates/bevy_core_pipeline/src/core_3d/mod.rs
+++ b/crates/bevy_core_pipeline/src/core_3d/mod.rs
@@ -87,7 +87,7 @@ use bevy_render::{
         ViewSortedRenderPhases,
     },
     render_resource::{
-        BindGroupId, CachedRenderPipelineId, Extent3d, FilterMode, Sampler, SamplerDescriptor,
+        CachedRenderPipelineId, Extent3d, FilterMode, Sampler, SamplerDescriptor,
         Texture, TextureDescriptor, TextureDimension, TextureFormat, TextureUsages, TextureView,
     },
     renderer::RenderDevice,
@@ -233,16 +233,16 @@ pub struct Opaque3dBinKey {
     /// The function used to draw.
     pub draw_function: DrawFunctionId,
 
+    /// The ID of a bind group specific to the material.
+    ///
+    /// In the case of PBR, this is the `MaterialBindGroupIndex`.
+    pub material_bind_group_index: Option<u32>,
+
     /// The asset that this phase item is associated with.
     ///
     /// Normally, this is the ID of the mesh, but for non-mesh items it might be
     /// the ID of another type of asset.
     pub asset_id: UntypedAssetId,
-
-    /// The ID of a bind group specific to the material.
-    ///
-    /// In the case of PBR, this is the `MaterialBindGroupId`.
-    pub material_bind_group_id: Option<BindGroupId>,
 
     /// The lightmap, if present.
     pub lightmap_image: Option<AssetId<Image>>,
--- a/crates/bevy_pbr/src/material_bind_groups.rs
+++ b/crates/bevy_pbr/src/material_bind_groups.rs
@@ -1,0 +1,437 @@
+//! Material bind group management.
+
+use crate::Material;
+use bevy_derive::{Deref, DerefMut};
+use bevy_ecs::{
+    system::Resource,
+    world::{FromWorld, World},
+};
+use bevy_reflect::{std_traits::ReflectDefault, Reflect};
+use bevy_render::{
+    render_resource::{
+        BindGroup, BindGroupEntry, BindGroupLayout, BindGroupLayoutEntry, BindingResource,
+        BindingType, Buffer, BufferBinding, BufferInitDescriptor, BufferUsages,
+        OwnedBindingResource, Sampler, SamplerDescriptor, TextureViewDimension,
+        UnpreparedBindGroup, WgpuSampler, WgpuTextureView,
+    },
+    renderer::RenderDevice,
+    settings::WgpuFeatures,
+    texture::FallbackImage,
+};
+use bevy_utils::{default, tracing::error, HashMap};
+use core::iter;
+use core::marker::PhantomData;
+use core::num::NonZero;
+
+/// An object that creates and stores bind groups for materials.
+///
+/// This object groups bindless materials together as appropriate.
+#[derive(Resource)]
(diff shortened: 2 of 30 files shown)


pcwalton:
I've updated this PR, and because 0.15 is out it should be ready for review.

Here's a new Tracy profile on Bistro, using Blender with alpha mask enabled. This uses an as-yet-unsubmitted follow-up patch that adds bindless mode to `StandardMaterial`. You can see a 2x speedup on `submit_graph_commands`.

Screenshot 2024-11-30 163508

crates/bevy_render/src/render_asset.rs
```diff
@@ -61,8 +65,21 @@ pub trait RenderAsset: Send + Sync + 'static + Sized {
...
+    ///
+    /// You can implement this method if you need to access ECS data (via
+    /// `_param`) in order to perform cleanup tasks when the asset is removed.
+    ///
+    /// The default implementation does nothing.
+    fn finalize_asset(
+        _source_asset: AssetId<Self::SourceAsset>,
+        _param: &mut SystemParamItem<Self::Param>,
+    ) {
+    }
```
kristoff3r, line 82: To me `finalize_asset` could also sound like the last step that happens before the asset is ready to use. The comment was very helpful though.

crates/bevy_pbr/src/render/mesh.rs
```diff
@@ -630,6 +636,13 @@ pub struct RenderMeshInstancesCpu(MainEntityHashMap<RenderMeshInstanceCpu>);
 #[derive(Default, Deref, DerefMut)]
 pub struct RenderMeshInstancesGpu(MainEntityHashMap<RenderMeshInstanceGpu>);
 
+/// FIXME: The fact that we need this is ugly.
```
kristoff3r, line 639: This comment would be more useful if you could also say why we need it, so we don't have to reverse engineer the reasoning later. Or maybe we can create an issue?

I guess it's fallout from the retained render world?

crates/bevy_pbr/src/material_bind_groups.rs
```diff
@@ -0,0 +1,562 @@
...
+                            ) => vec.push(texture_view),
+                            (
+                                &mut BindingResourceArray::Samplers(ref mut vec),
+                                OwnedBindingResource::Sampler(sampler),
+                            ) => vec.push(sampler),
+                            _ => {
+                                error!(
+                                    "Mismatched bind group layouts; can't combine bind groups \
+                                    into a single bindless bind group!"
+                                );
```
kristoff3r, line 456: Can we add the material type and mismatching resource here? That should speed up debugging quite a bit if this happens.

crates/bevy_pbr/src/material_bind_groups.rs
kristoff3r, line 1: This file was a joy to read, I hope we can some day get the rest of the renderer to this standard of code quality and documentation.

kristoff3r (approved):
I tested the example on Linux and WebGL2 in Firefox. Apart for the need to make yet another allocator this was much less invasive than I expected, and all the code makes sense to me.

crates/bevy_core_pipeline/src/core_3d/mod.rs
```diff
@@ -233,17 +233,17 @@ pub struct Opaque3dBinKey {
     /// The function used to draw.
     pub draw_function: DrawFunctionId,
 
+    /// The ID of a bind group specific to the material.
```
JMS55, line 236: Mateiral [type | instance] ?

crates/bevy_render/src/render_asset.rs
JMS55: Yes I'd rather rename it on_unload or something.

crates/bevy_pbr/src/render/mesh.rs
pcwalton: I just removed the comment. I don't actually remember why I wrote it, embarrassingly enough.

crates/bevy_core_pipeline/src/core_3d/mod.rs
pcwalton: Instance. I added that to the comment.

crates/bevy_pbr/src/material_bind_groups.rs
pcwalton: Done. I think what I added is what you wanted.

crates/bevy_render/src/render_asset.rs
pcwalton: It's now called `unload_asset` to go along with `prepare_asset`.

kristoff3r (approved):

examples/shader/shader_material_bindless.rs
```diff
@@ -0,0 +1,73 @@
...
+    commands.spawn((
+        Mesh3d(meshes.add(Cuboid::default())),
+        MeshMaterial3d(materials.add(BindlessMaterial {
+            color: LinearRgba::BLUE,
+            color_texture: Some(asset_server.load("branding/bevy_logo_dark.png")),
+        })),
+        Transform::from_xyz(-2.0, 0.5, 0.0),
+    ));
+
+    // Add a cube with a red tinted texture.
```
IceSentry, line 52:
```suggestion
    // Add a cylinder with a red tinted texture.
```

IceSentry (approved):
LGTM. Very high quality PR as expected. It's a pretty big change, but it's nice that it doesn't affect that many user facing apis.

Once this is merged I'll make a PR to take advantage of it for the default Mesh2d material.

JMS55 (approved):

alice-i-cecile:
Thank you to everyone involved with the authoring or reviewing of this PR! This work is relatively important and needs release notes! Head over to https://github.com/bevyengine/bevy-website/issues/1960 if you'd like to help out.

PR #12287, opened 2024-03-04
Support closures as ComponentHooks

james7132:
# Objective
Component hooks right now only take raw function pointers instead of allowing those registering it to store some state with them.

## Solution
Store an `Arc<dyn Fn(...)>` instead of a function pointer and make the registration functions generic over the function type being passed in.

Right now this only takes read only closures (`Fn` not `FnMut`), though strictly speaking we could support them as we have mutable access when triggering the hooks, accessing the closure mutably through an Arc would be troublesome if the ComponentInfo was cloned. I tried to keep the traits on `ComponentHooks` and `ComponentInfo` intact as much as I could.

This shouldn't change the performance characteristics of the crate when not using a closure since `&dyn Fn(...)` directly includes the function pointer instead of being a double indirection.

This PR also makes the type alias for ComponentHook private since it's now not part of the public interface of the crate.

PR #16371, opened 2024-11-13
Support component hooks closures

mnmaita:
# Objective

- Adopted #12287

## Solution

- Updated the `Component` macro to accept closures for `ComponentHooks`.

## Testing

- Compiled locally and tested by altering the `component_hooks` example.

### DIFF
--- a/crates/bevy_ecs/macros/src/component.rs
+++ b/crates/bevy_ecs/macros/src/component.rs
@@ -179,10 +179,10 @@ pub const ON_REMOVE: &str = "on_remove";
 struct Attrs {
     storage: StorageTy,
     requires: Option<Punctuated<Require, Comma>>,
-    on_add: Option<ExprPath>,
-    on_insert: Option<ExprPath>,
-    on_replace: Option<ExprPath>,
-    on_remove: Option<ExprPath>,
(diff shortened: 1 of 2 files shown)


mnmaita:
I still have some outstanding questions regarding this change:

- [ ] Is the example update any good? Or should I consider doing a "full rework" where we actually showcase the three ways of adding hooks to components in the actual code?
- [ ] Should we consolidate `RequiredFunc` and `HookFunc` enums since they are essentially the same? I went for a simpler approach but was thinking on creating some `DeriveFunc` (or similar) to cover both cases and avoid code repetition.

hymm:
shouldn't this have the changes from `bevy_ecs/src/components.rs` in #12287 ?

Looking at the discussion in that pr, I'd vote not to support closures for hooks unless a user requests it. `Fn` can only borrow immutably, which doesn't feel like it adds much functionality here so I'm not sure how useful it would be.

ChristopherBiscardi:
This adoption unfortunately has so far failed to build the consensus that https://github.com/bevyengine/bevy/pull/12287 also failed to build. Given that it has been a couple years it is somewhat doubtful that this PR is going to achieve that without a new push, which will have to resolve conflicts and such anyway.

PR #14730, opened 2024-08-13
bevy_reflect: Add `#[reflect(readonly)]` attribute

MrGVSV:
# Objective

There are times when you want a type to be `Reflect`, but don't want to allow the modification of certain fields. There a many reasons you might want to do this. One of those reasons is that your type might need to maintain certain invariants in its fields.

For example, we have defined a `History` type for an editor below.

```rust
struct History {
    /// New/redone actions are pushed here.
    undo_stack: Vec<Op>,
    /// Undone actions are pushed here.
    redo_stack: Vec<Op>,
}
```

It's important that the undo and redo actions are performed through the `History::undo` and `History::redo` methods so that the `Op` gets moved to the proper stack.

However, if we derive `Reflect`, then someone could very easily break this flow through reflection.

We _could_ derive it as a `#[reflect_value]`, but then our reflection-based editor pane can no longer display the undo and redo stacks.

Ideally, we'd be able to allow those fields to be visible without sacrificing our desired level of mutability privileges.

## Solution

This PR adds a `#[reflect(readonly)]` field attribute that prevents that field from being mutably accessed via reflection.

Deriving `Reflect` on a type result in one of three subtraits being implemented: `Struct`, `TupleStruct`, or `Enum`. Because these are the only user-configurable traits (at least via the derive macro), these are the main trait affected by this PR.

All `field`, `field_mut`, `field_at`, and `field_at_mut` methods across these three traits have been modified to return a `Result` instead of an `Option`. The error type, `ReflectFieldError`, then has two variants:
- `NotFound` - This takes the place of `None`, where the given field identifier does not exist on the type or the field is marked with `#[reflect(ignore)]`.
- `Readonly` - This is returned when attempting to mutably access a field marked with `#[reflect(readonly)]`.

Additionally, the `PartialReflect::try_apply` method implementations—which rely on these field methods under the hood—will now skip readonly fields.

We also expose the readonly status of the field via type info:
- `NamedField::readonly`
- `UnnamedField::readonly`

This PR also adds the `FieldId` enum to help with error messaging. This will be super useful for error messaging in the future. And as an added benefit, it can even be used to consolidate named-vs-unnamed APIs:

```rust
// BEFORE
fn print_field(value: &dyn PartialReflect, name: &str) {/* ... */}
fn print_field_at(value: &dyn PartialReflect, index: usize) {/* ... */}

// AFTER
fn print_field(value: &dyn PartialReflect, id: impl Into<FieldId>) {/* ... */}
```

[...]

PR #16372, opened 2024-11-13
Add Immutable `Component` Support

### LINKED ISSUE #16208: Allow certain components to be marked immutable

There are needs to enforce invariants, to build indexes or otherwise keep things consistent in the ECS between multiple entities or between multiple components in a single entity. The new "component hooks" system goes a long way towards making it easy to keep things in sync, as they can be triggered on component addition, removal and replace.

However, the hooks system does not catch mutation currently, and catching mutation has a lot of difficult problems to solve if it needs to be synchronously and with every way of mutating a component. Even if all the fields inside a component are marked private, that doesn't prevent the user from assigning a new instance of that component to it, overwriting the data for the old component - and no hooks will be fired, only asynchronous change events. This means that it is currently impossible to use hooks to maintain an invariant if the user mutates components directly.

There could be a flag on components that marks the component immutable. When set, it would mean that no API will return mutable references to the component, making it impossible to mutate the values in the component. This doesn't mean that there would be any change in how `insert`, `remove`, `take` etc. would work, so it would still be fully permitted to remove the component, or add a new one, or replace an existing with a new one. The current hooks system will fire `on_replace` and `on_insert` hooks when a component is replaced this way instead of mutated in-place.

How the mutation APIs should behave when called on a component that is marked immutable is still unspecified. Perhaps they should panic when trying to mutate a component that can't be mutated. Or maybe they act as if the component doesn't exist on the object if attempting to get a mutable reference. Or perhaps this can somehow be extended all the way to the type system so the calls which attempt to take a mutable reference to a component simply won't compile because those are not implemented for components which are immutable.

The big alternative is that absolutely no change is needed. Preventing mutation is just a safeguard for certain kinds of components, and if the fields of the component are private, doing a mutable assignment on a component is not easy to accidentally do – so for most usages it's probably enough to just say in documentation "please don't mutate this component".

The second alternative is also that absolutely no change is needed. Such components may be made private, and exposed through `QueryData` and `Bundle` in interfaces such that the real component is never accessible, hence never mutated. It might still be accessible via reflection, save/load, etc. but all bets are off there anyway. This is not as nice as it requires a bunch of boilerplate, and the whole point of component hooks was to allow such invariants to be implemented without creating a shadow API to do all the ECS operations while keeping invariants.

The third alternative is to somehow make synchronous change detection hooks work. There is already change detection, so it's not so impossible to think that it could run a hook real time, possibly both before change and after a change. Then we could just have `on_mutate` and that would probably cover the vast majority of cases that would benefit from immutable components.

## Additional context

I'm writing this issue mainly as a placeholder, and to see comments. It's one problem related to hooks that may or may not be common enough to warrant doing something about it. If there comes a strong use case for this that isn't easily solved by other means, and this problem is the primary motivation to make components private, then I will likely want to revisit this.

bushrat011899:
The `Component` trait now implies an immutable, readonly component. To add mutability, you must implement `ComponentMut`, which is a simple marker. In the derive macro, `ComponentMut` will be implemented with `Component` unless you add an `#[immutable]` attribute.

# Objective

- Fixes #16208

## Solution

- Made `Component` imply a read-only, immutable component.
- Added new trait, `ComponentMut`, which allows mutation.
- Updated `derive_component` to derive `ComponentMut` unless an `#[immutable]` attribute is added.
- Updated `ReflectComponent` to only expose read-only methods, and added `ReflectComponentMut` for read and write.
- Added `immutable_components` example demonstrating the feature.

## Testing

- CI
- `immutable_components` example.

---

## Showcase

Users can now mark a component as `#[immutable]` to prevent safe mutation of a component while it is attached to an entity:

```rust
#[derive(Component)]
#[immutable]
struct Foo {
    // ...
}
```

This prevents creating an exclusive reference to the component while it is attached to an entity. This is particularly powerful when combined with component hooks, as you can now fully track a component's value, ensuring whatever invariants you desire are upheld.

Before this would be done my making a component private, and manually creating a `QueryData` implementation which only permitted read access.

## Migration Guide

- When implementing `Component` manually, you must now implement `ComponentMut` for mutable components:
```rust
impl ComponentMut for Foo {}
```
- When reflecting a component, you must reflect `ComponentMut` as well in order to have access to mutable reflection methods.
```rust
#[derive(Reflect, Component)]
#[reflect(Component, ComponentMut)]
struct Foo {
    // ...
}
```
- When working with generic components, you may need to specify that your generic parameter implements `ComponentMut` rather than `Component` if you require mutable access to said component.

## Notes

- I've done my best to implement this feature, but I'm not happy with how reflection has turned out. If any reflection SMEs know a way to improve this situation I'd greatly appreciate it.
- I've attempted to prevent all safe mutable access to a component that does not implement `ComponentMut`, but there may still be some methods I have missed. Please indicate so and I will address them, as they are bugs.
- Unsafe is an escape hatch I am _not_ attempting to prevent. Whatever you do with unsafe is between you and your compiler.
- I am marking this PR as ready, but I suspect it will undergo fairly major revisions based on SME feedback.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1944,6 +1944,17 @@ category = "ECS (Entity Component System)"
 wasm = false
 
 [[example]]
+name = "immutable_components"
+path = "examples/ecs/immutable_components.rs"
+doc-scrape-examples = true
+
+[package.metadata.example.immutable_components]
+name = "Immutable Components"
+description = "Demonstrates the creation and utility of immutable components"
+category = "ECS (Entity Component System)"
+wasm = false
+
+[[example]]
 name = "iter_combinations"
 path = "examples/ecs/iter_combinations.rs"
 doc-scrape-examples = true
--- a/crates/bevy_audio/src/audio.rs
+++ b/crates/bevy_audio/src/audio.rs
@@ -51,7 +51,7 @@ pub enum PlaybackMode {
 /// [`AudioSink`][crate::AudioSink] or [`SpatialAudioSink`][crate::SpatialAudioSink]
 /// components. Changes to this component will *not* be applied to already-playing audio.
 #[derive(Component, Clone, Copy, Debug, Reflect)]
-#[reflect(Default, Component, Debug)]
+#[reflect(Default, ComponentMut, Component, Debug)]
 pub struct PlaybackSettings {
     /// The desired playback behavior.
     pub mode: PlaybackMode,
@@ -147,7 +147,7 @@ impl PlaybackSettings {
 /// This must be accompanied by `Transform` and `GlobalTransform`.
 /// Only one entity with a `SpatialListener` should be present at any given time.
 #[derive(Component, Clone, Debug, Reflect)]
-#[reflect(Default, Component, Debug)]
+#[reflect(Default, ComponentMut, Component, Debug)]
 pub struct SpatialListener {
     /// Left ear position relative to the `GlobalTransform`.
     pub left_ear_offset: Vec3,
@@ -249,7 +249,7 @@ pub type AudioBundle = AudioSourceBundle<AudioSource>;
 /// Playback can be configured using the [`PlaybackSettings`] component. Note that changes to the
 /// `PlaybackSettings` component will *not* affect already-playing audio.
 #[derive(Component, Reflect)]
-#[reflect(Component)]
+#[reflect(ComponentMut, Component)]
 #[require(PlaybackSettings)]
 pub struct AudioPlayer<Source = AudioSource>(pub Handle<Source>)
 where
--- a/crates/bevy_ecs/macros/src/component.rs
+++ b/crates/bevy_ecs/macros/src/component.rs
@@ -33,6 +33,8 @@ pub fn derive_event(input: TokenStream) -> TokenStream {
         impl #impl_generics #bevy_ecs_path::component::Component for #struct_name #type_generics #where_clause {
             const STORAGE_TYPE: #bevy_ecs_path::component::StorageType = #bevy_ecs_path::component::StorageType::SparseSet;
         }
+
+        impl #impl_generics #bevy_ecs_path::component::ComponentMut for #struct_name #type_generics #where_clause { }
     })
 }
 
@@ -141,7 +143,7 @@ pub fn derive_component(input: TokenStream) -> TokenStream {
 
     // This puts `register_required` before `register_recursive_requires` to ensure that the constructors of _all_ top
     // level components are initialized first, giving them precedence over recursively defined constructors for the same component type
-    TokenStream::from(quote! {
+    let mut implementation = TokenStream::from(quote! {
         #required_component_docs
         impl #impl_generics #bevy_ecs_path::component::Component for #struct_name #type_generics #where_clause {
             const STORAGE_TYPE: #bevy_ecs_path::component::StorageType = #storage;
@@ -164,7 +166,15 @@ pub fn derive_component(input: TokenStream) -> TokenStream {
                 #on_remove
             }
         }
-    })
+    });
+
+    if !attrs.immutable {
+        implementation.extend(TokenStream::from(quote! {
+            impl #impl_generics #bevy_ecs_path::component::ComponentMut for #struct_name #type_generics #where_clause { }
+        }));
+    }
+
+    implementation
 }
 
 pub const COMPONENT: &str = "component";
@@ -176,6 +186,8 @@ pub const ON_INSERT: &str = "on_insert";
 pub const ON_REPLACE: &str = "on_replace";
 pub const ON_REMOVE: &str = "on_remove";
 
+pub const IMMUTABLE: &str = "immutable";
+
 struct Attrs {
     storage: StorageTy,
     requires: Option<Punctuated<Require, Comma>>,
@@ -183,6 +195,7 @@ struct Attrs {
     on_insert: Option<ExprPath>,
     on_replace: Option<ExprPath>,
     on_remove: Option<ExprPath>,
+    immutable: bool,
 }
 
 #[derive(Clone, Copy)]
@@ -213,6 +226,7 @@ fn parse_component_attr(ast: &DeriveInput) -> Result<Attrs> {
         on_replace: None,
         on_remove: None,
         requires: None,
+        immutable: false,
     };
 
     let mut require_paths = HashSet::new();
@@ -262,6 +276,8 @@ fn parse_component_attr(ast: &DeriveInput) -> Result<Attrs> {
             } else {
                 attrs.requires = Some(punctuated);
             }
+        } else if attr.path().is_ident(IMMUTABLE) {
+            attrs.immutable = true;
         }
     }
 
--- a/crates/bevy_ecs/src/component.rs
+++ b/crates/bevy_ecs/src/component.rs
@@ -73,6 +73,16 @@ use derive_more::derive::{Display, Error};
 ///
 /// # Component and data access
 ///
+/// Components can be marked as immutable by adding the `#[immutable]` attribute when using the
+/// derive macro. Alternatively, a component will be immutable by default when implementing
+/// [`Component`] manually. To make a manually implemented component mutable, also implement
+/// the marker trait [`ComponentMut`].
+///
+/// Immutable components can be removed, replaced, and inserted just like mutable components.
+/// The only guarantee that's enforced is that while an immutable component is attached
+/// to an entity, an exclusive reference `&mut C` will never be produced. This allows
+/// hooks to observe all changes made to an immutable component.
+///
 /// See the [`entity`] module level documentation to learn how to add or remove components from an entity.
 ///
 /// See the documentation for [`Query`] to learn how to access component data from a system.
@@ -391,6 +401,16 @@ pub trait Component: Send + Sync + 'static {
     ) {
     }
 }
+
+/// Marks that a [`Component`] is mutable.
+/// Without this marker, a component is immutable.
+#[diagnostic::on_unimplemented(
+    message = "`{Self}` is not a `ComponentMut`",
+    label = "invalid `ComponentMut`",
+    note = "consider annotating `{Self}` with `#[derive(Component)]`",
+    note = "if `{Self}` is a `Component`, it is immutable"
+)]
+pub trait ComponentMut: Component {}
 
 /// The storage used for a specific component type.
 ///
--- a/crates/bevy_ecs/src/entity/mod.rs
+++ b/crates/bevy_ecs/src/entity/mod.rs
@@ -253,7 +253,7 @@ impl Entity {
     /// # use bevy_ecs::{prelude::*, component::*};
     /// # use bevy_reflect::Reflect;
     /// #[derive(Reflect, Component)]
-    /// #[reflect(Component)]
+    /// #[reflect(ComponentMut, Component)]
     /// pub struct MyStruct {
     ///     pub entity: Entity,
     /// }
--- a/crates/bevy_ecs/src/lib.rs
+++ b/crates/bevy_ecs/src/lib.rs
@@ -46,7 +46,7 @@ pub mod prelude {
     pub use crate::{
         bundle::Bundle,
         change_detection::{DetectChanges, DetectChangesMut, Mut, Ref},
-        component::Component,
+        component::{Component, ComponentMut},
         entity::{Entity, EntityMapper},
         event::{Event, EventMutator, EventReader, EventWriter, Events},
         observer::{Observer, Trigger},
@@ -71,7 +71,7 @@ pub mod prelude {
     #[doc(hidden)]
     #[cfg(feature = "bevy_reflect")]
     pub use crate::reflect::{
-        AppTypeRegistry, ReflectComponent, ReflectFromWorld, ReflectResource,
+        AppTypeRegistry, ReflectComponent, ReflectComponentMut, ReflectFromWorld, ReflectResource,
     };
 
     #[doc(hidden)]
--- a/crates/bevy_ecs/src/reflect/component.rs
+++ b/crates/bevy_ecs/src/reflect/component.rs
@@ -2,43 +2,43 @@
 //! This allows inserting, updating, removing and generally interacting with components
 //! whose types are only known at runtime.
 //!
-//! This module exports two types: [`ReflectComponentFns`] and [`ReflectComponent`].
+//! This module exports two types: [`ReflectMutabelComponentFns`] and [`ReflectMutabelComponent`].
 //!
 //! # Architecture
 //!
-//! [`ReflectComponent`] wraps a [`ReflectComponentFns`]. In fact, each method on
-//! [`ReflectComponent`] wraps a call to a function pointer field in `ReflectComponentFns`.
+//! [`ReflectMutabelComponent`] wraps a [`ReflectMutabelComponentFns`]. In fact, each method on
+//! [`ReflectMutabelComponent`] wraps a call to a function pointer field in `ReflectMutabelComponentFns`.
 //!
-//! ## Who creates `ReflectComponent`s?
+//! ## Who creates `ReflectMutabelComponent`s?
 //!
-//! When a user adds the `#[reflect(Component)]` attribute to their `#[derive(Reflect)]`
+//! When a user adds the `#[reflect(ComponentMut, Component)]` attribute to their `#[derive(Reflect)]`
 //! type, it tells the derive macro for `Reflect` to add the following single line to its
 //! [`get_type_registration`] method (see the relevant code[^1]).
 //!
 //! ```
 //! # use bevy_reflect::{FromType, Reflect};
-//! # use bevy_ecs::prelude::{ReflectComponent, Component};
+//! # use bevy_ecs::prelude::{ReflectMutabelComponent, Component};
 //! # #[derive(Default, Reflect, Component)]
 //! # struct A;
 //! # impl A {
 //! #   fn foo() {
 //! # let mut registration = bevy_reflect::TypeRegistration::of::<A>();
-//! registration.insert::<ReflectComponent>(FromType::<Self>::from_type());
+//! registration.insert::<ReflectMutabelComponent>(FromType::<Self>::from_type());
 //! #   }
 //! # }
 //! ```
 //!
-//! This line adds a `ReflectComponent` to the registration data for the type in question.
-//! The user can access the `ReflectComponent` for type `T` through the type registry,
+//! This line adds a `ReflectMutabelComponent` to the registration data for the type in question.
+//! The user can access the `ReflectMutabelComponent` for type `T` through the type registry,
 //! as per the `trait_reflection.rs` example.
 //!
 //! The `FromType::<Self>::from_type()` in the previous line calls the `FromType<C>`
-//! implementation of `ReflectComponent`.
+//! implementation of `ReflectMutabelComponent`.
 //!
-//! The `FromType<C>` impl creates a function per field of [`ReflectComponentFns`].
+//! The `FromType<C>` impl creates a function per field of [`ReflectMutabelComponentFns`].
 //! In those functions, we call generic methods on [`World`] and [`EntityWorldMut`].
 //!
-//! The result is a `ReflectComponent` completely independent of `C`, yet capable
+//! The result is a `ReflectMutabelComponent` completely independent of `C`, yet capable
 //! of using generic ECS methods such as `entity.get::<C>()` to get `&dyn Reflect`
 //! with underlying type `C`, without the `C` appearing in the type signature.
 //!
@@ -48,8 +48,8 @@
 //! for generics) is generated **unconditionally**, regardless of whether it ends
 //! up used or not.
 //!
-//! Adding `N` fields on `ReflectComponentFns` will generate `N × M` additional
-//! functions, where `M` is how many types derive `#[reflect(Component)]`.
+//! Adding `N` fields on `ReflectMutabelComponentFns` will generate `N × M` additional
+//! functions, where `M` is how many types derive `#[reflect(ComponentMut, Component)]`.
 //!
 //! Those functions will increase the size of the final app binary.
 //!
@@ -60,8 +60,9 @@
 use super::from_reflect_with_fallback;
 use crate::{
     change_detection::Mut,
-    component::{Component, ComponentId},
+    component::ComponentId,
     entity::Entity,
+    prelude::{Component, ComponentMut},
     world::{
         unsafe_world_cell::UnsafeEntityCell, EntityMut, EntityWorldMut, FilteredEntityMut,
         FilteredEntityRef, World,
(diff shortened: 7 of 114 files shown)


BenjaminBrienen:
The behavior of implementing `Component` before this change and the behavior of implementing ` Component + ComponentMut` after this change should be identical. Do I understand that correctly?

bushrat011899:
Yes a pre-this-PR `Component` is identical to a this-PR `Component + ComponentMut`. `Component` contains all the implementation details it had previously, but now only implies an immutable type. Mutability is now explicitly stated by implementing `ComponentMut`. But for the derive macro, `Component + ComponentMut` are implemented by default (since that is the most typical use-case). To opt-out of mutability in the derive macro, you add `#[immutable]`.

ItsDoot:
Small nit: I would prefer `#[component(immutable)]` to keep all component attributes together. It also follows `#[world_query(mutable)]`.

crates/bevy_ecs/src/entity/mod.rs
```diff
@@ -253,7 +253,7 @@ impl Entity {
     /// # use bevy_ecs::{prelude::*, component::*};
     /// # use bevy_reflect::Reflect;
     /// #[derive(Reflect, Component)]
-    /// #[reflect(Component)]
+    /// #[reflect(ComponentMut, Component)]
```
BenjaminBrienen, line 256: I noticed that this one has both, but the others (that I've read so far) were just changed to ComponentMut. Is that intentional?

bushrat011899: It's not intentional, all cases of `#[reflect(Component)]` should be updated to `#[reflect(Component, ComponentMut)]` to preserve previous behaviour. I've made a few changes to fix some of the inconsistencies.

bushrat011899:
Of note, `FilteredEntityMut::get_mut_by_id` is (so far) the only safe method I have found that can bypass immutable components. I did want to add the immutable flag to `ComponentDescriptor`, but propagating that information proved very challenging. If anyone has a suggestion for how to integrate `ComponentMut` and `ComponentDescriptor` in the least impactful way I would be greatly appreciative.

alice-i-cecile:
Why do you prefer the `ComponentMut: Component` design over a `Mutable + Component` design? I have a mild preference for the latter because I think it'll be easier to extend to resources.

NthTensor:
Don't have time to look over this fully, but I like this. I also prefer the version without the trait bound on component.

Just so I am sure this can be used as I want, if we make parent/children immutable, how do we preserve the existing hierarchy commands api? Will we use unsafe within the commands to get mutable access, or go properly immutable with only clones and inserts?

MrGVSV:
> Just so I am sure this can be used as I want, if we make parent/children immutable, how do we preserve the existing hierarchy commands api? Will we use unsafe within the commands to get mutable access, or go properly immutable with only clones and inserts?

I was wondering if it would make sense to have mutable component access require a key type. Then crates could keep that type private to simulate immutability while still being able to mutate the component themselves.

Not sure if that's possible and I don't know how well it fits with this approach, but possibly an option (though I’m going to guess far more complex and involved).

iiYese:
> or go properly immutable with only clones and inserts

It would be this. Either through `Parent`'s on insert hook or a command.

NthTensor:
We'll have to make `Children` immutable too, won't we? That will mean some extra vec cloning, so might want to look at an immutable vector impl. But I'm content with that answer. We can deal with the costs if it turns out to be a problem.

bushrat011899:
> Why do you prefer the `ComponentMut: Component` design over a `Mutable + Component` design? I have a mild preference for the latter because I think it'll be easier to extend to resources.

The three reasons for me were:

1. With reflection, we'd need a name for the trait that represents the mutable trait methods, and I'm not sure if we can do `ReflectComponentMutable`
2. For updating where conditions it was _slightly_ cleaner to write `ComponentMut` than `Component + Mutable`
3. When we do extend this to resources, we'd have to use a separate trait anyway, because there are types that implement both `Resource` _and_ `Component`, which would create a conflict on the common `Mutable`

NthTensor:
It's probably fine to keep this component-only for now, considering that relations don't currently support hooks (so the applications are pretty minimal).

bushrat011899:
> We'll have to make `Children` immutable too, won't we? That will mean some extra vec cloning, so might want to look at an immutable vector impl. But I'm content with that answer. We can deal with the costs if it turns out to be a problem.

This shouldn't be required. The data is only immutable while it is stored on the entity, so remove/mutate/insert can be used to avoid any cloning. If that pattern is cumbersome we could add a `component_scope` method like `resource_scope`, except with ownership of the component.

NthTensor:
> This shouldn't be required. The data is only immutable while it is stored on the entity, so remove/mutate/insert can be used to avoid any cloning. If that pattern is cumbersome we could add a component_scope method like resource_scope, except with ownership of the component.

Wouldn't this create unnecessary archetype moves? I feel like cloning is probably faster. But that's getting off topic.

alice-i-cecile:
You could even do a `mutate_via_reinsertion` API, which could work on immutable components. Very clean, and no archetype moves. Easy followup though.

hymm:
What happens if you ask for a `&mut C` on an immutable C in a system? Is it the compile error where your function doesn't implement system?

bushrat011899:
Compiler error. The `QueryData` implementation is not provided for immutable components.

bushrat011899:
There's some discussion on Discord around the name _immutable_. These components _are_ mutable, but only when outside of the ECS. Once stored on an entity they become immutable. To clarify this, an alternative name is being bikeshedded. My personal preference is for _frozen_, as that has the same implications as immutable, with the added possibility of "thawing" (remove to mutate)

alice-i-cecile:
For posterity, frozen gets my vote as well, but "restricted components" is a safe fallback.

MalekiRe (approved):
Awesome

chrisjuchem:
I'd love if there was a trait like `ComponentNonMut`. `bevy_mod_index` would want to use that as a bound for components that can be reliability tracked with hooks now that mutations can be prevented.

bushrat011899:
We could maybe add an `unsafe trait ComponentNonMut: Component { }` to cover that usecase. We would then have to make `ComponentMut` unsafe too, since they're mutually exclusive. The derive macro could hide these details tho.

alice-i-cecile:
Unfortunately Rust's trait system doesn't play very nice with that sort of negative bound. If we had both `MutableComponent` and `ImmutableComponent` as traits, there's nothing stopping users from implementing *both*. I agree that both forms would be useful though.

Would an associated type work here?

`impl <C: Component<Mutable=true>> Index {}` seems pretty workable, and I don't think it's much worse / less clear than the current design. It would also bypass the nasty reflection problems of the current approach.

crates/bevy_audio/src/audio.rs
```diff
@@ -51,7 +51,7 @@ pub enum PlaybackMode {
 /// [`AudioSink`][crate::AudioSink] or [`SpatialAudioSink`][crate::SpatialAudioSink]
 /// components. Changes to this component will *not* be applied to already-playing audio.
 #[derive(Component, Clone, Copy, Debug, Reflect)]
-#[reflect(Default, Component, Debug)]
+#[reflect(Default, ComponentMut, Component, Debug)]
```
cart, line 54: Haven't done a review yet or a global "is this worth it" analysis, but at first glance needing to reflect `Component` _and_ `ComponentMut` on every component feels like a painful UX price to pay.

alice-i-cecile: Yeah, I think that everyone (including the author) agrees that this is the biggest problem with the current PR.

bushrat011899: Agreed. The issues I encountered that led me here are:

1. The implementation for `ReflectComponent` included mutable methods (e.g., `reflect_mut`) which fail to compile in this PR, so they have to removed.
2. Ideally, we'd have `ReflectComponent` and `ReflectImmutableComponent`, where `ReflectComponent` requires `ComponentMut` and `ReflectImmutableComponent` only requires `Component`.
3. Reflection support for immutable components is a must, since we'd potentially like to use it for types like `Name`, `Parent`, `Children`, etc., which are all _very_ important.

I'll experiment with the PR some more to see if I can remove the need to `#[reflect(ComponentMut)]`, because that's definitely the worst part of this PR right now.

iiYese: Would making this information dynamic as metadata of the component help? All of these problems seem to be coming about from information solely existing in the type system.

bushrat011899: I believe I have addressed the reflection issue entirely now, as well as dynamic components, _and_ can offer a reliable and safe `ComponentImmutable` trait that acts as `!ComponentMut`.

bushrat011899:
Using an associated type on `Component` allows resolving the reflection DX issues previously introduced by this PR. Additionally, we can now offer a marker trait `ComponentImmutable` which is effectively `!ComponentMut`.

crates/bevy_audio/src/audio.rs
bushrat011899: > Would making this information dynamic as metadata of the component help? All of these problems seem to be coming about from information solely existing in the type system.

The information not being available in the type system is an absolute non-starter for me. Letting users write `Query<&mut MyImmutableFoo>` only for it to crash at runtime massively degrades the developer experience.

alice-i-cecile: Agreed. This information needs to exist in both the type-system and dynamically.

crates/bevy_ecs/src/component.rs
```diff
@@ -73,6 +73,14 @@ use derive_more::derive::{Display, Error};
 ///
 /// # Component and data access
 ///
+/// Components can be marked as immutable by adding the `#[component(immutable)]` attribute when using the
+/// derive macro.
+///
+/// Immutable components can be removed, replaced, and inserted just like mutable components.
+/// The only guarantee that's enforced is that while an immutable component is attached
```
alice-i-cecile, line 80: The framing here is backwards. Talk about the way it's different first, then how it's the same.

crates/bevy_audio/src/audio.rs
iiYese: Read that comment again.

> **solely** existing in the type system

crates/bevy_ecs/src/component.rs
```diff
@@ -378,6 +386,18 @@ pub trait Component: Send + Sync + 'static {
...
+    /// mutable, or immutable. Mutable components will have [`ComponentMut`]
+    /// automatically implemented for them, while immutable components will
+    /// receive [`ComponentImmutable`].
+    ///
+    /// * For a component to be mutable, this type must be `Self`.
+    /// * For a component to be immutable, this type must be `()`.
+    ///
+    /// If you use any other type, this component will be immutable but will not
+    /// implement [`ComponentImmutable`].
+    type Mutable: 'static;
```
alice-i-cecile, line 399: IMO we should borrow from the approach taken by `StorageType`, and use a trait bound for marker types, rather than using `Self` / `()`. I think that's clearer, more consistent and it's much easier to extend.

crates/bevy_ecs/src/world/entity_ref.rs
```diff
@@ -516,16 +516,28 @@ impl<'w> EntityMut<'w> {
...
     }
 
+    /// Gets mutable access to the component of type `T` for the current entity.
+    /// Returns `None` if the entity does not have a component of type `T`.
+    ///
+    /// # Safety
+    ///
+    /// - `T` must be a mutable component
+    #[inline]
+    pub unsafe fn get_mut_assume_mutable<T: Component>(&mut self) -> Option<Mut<'_, T>> {
```
alice-i-cecile, line 531: Hmm. I suppose this method should probably exist, over just telling users to use `UnsafeWorldCell` or whatever, since it's easier to uphold its safety requirements.

alice-i-cecile, line 531: This needs a dynamic equivalent.

crates/bevy_ecs/src/component.rs
bushrat011899: Gone with `Immutable` and `Mutable` as ZSTs implementing a new `ComponentMutability` trait. This also simplifies the boolean check for mutability by allowing the value to be stored on the ZST.

crates/bevy_ecs/src/world/entity_ref.rs
bushrat011899: Yeah it's a necessary evil like most unsafe functions.

bushrat011899: The `get_mut_by_id` method is able to perform the check at runtime, since it already needs the `ComponentInfo` to find it. I _could_ add an alternate method to skip that single branch, but I'd rather not duplicate more APIs if it can be avoided. If users need it they'll ask for it.

bushrat011899:
After further iteration, the `ComponentMut` and `ComponentImmutable` traits have been removed, opting instead for matching against the associated type `Mutability`.

```rust
/// Work with a component, regardless of its mutability
fn get_component<C: Component>(/* ... */) { /* ... */ }

/// _Only_ allow mutable components
fn get_component_mut<C: Component<Mutability = Mutable>>(/* ... */) { /* ... */ }

/// _Only_ allow immutable components
fn get_component_immutable<C: Component<Mutability = Immutable>>(/* ... */) { /* ... */ }
```

crates/bevy_ecs/src/world/entity_ref.rs
alice-i-cecile: Aha, got it :)

alice-i-cecile:
I think that my only remaining major request is that this has a test suite for dynamic components. Once that's done I'll do a final polish pass on this at the start of 0.16.

BenjaminBrienen:
I really like how this PR is turning out. I'm a fan of the associated type idea.

bushrat011899:
As an aside, I added `#[component(immutable)]` to `Name` to see if anything broke, and to my pleasant surprise it just worked without any knock-on effects. Bevy already treats `Name` as some immutable data, so making it "official" is a 1-line PR once this is merged. I haven't included that in this PR just to keep the scope as small as it can be.

bushrat011899:
> The associated trait magic with sealed logic types is very cute, I like it. Is the issue with `FilteredEntityMut::get_mut_by_id` resolved or still outstanding? 

Thanks! And yes that issue is resolved. The move to an associated type made enough information available on `Component` for me to have immutability be a `bool` on `ComponentDescriptor`.

NthTensor (approved):
I did some thinking about what this API will let us do over my lunch-break, and I'm very pleased. An immutable component with an insert hook can cache data on a private required mutable component with extremely little overhead (a single archetype move, and some non-structural mutations). This seems to align perfectly with the direction the engine is going with required components and hooks (and, eventually, archetype invariants).

Looked over the code and saw no issues, as far as I am aware this is good to go.

crates/bevy_ecs/src/world/entity_ref.rs
```diff
@@ -2692,12 +2684,29 @@ impl<'w> FilteredEntityMut<'w> {
...
+
+    /// Consumes self and gets mutable access to the component of type `T`
+    /// with the world `'w` lifetime for the current entity.
+    /// Returns `None` if the entity does not have a component of type `T`.
+    ///
+    /// # Safety
+    ///
+    /// - `T` must be a mutable component
+    #[inline]
+    pub unsafe fn into_mut_assume_mutable<T: Component>(self) -> Option<Mut<'w, T>> {
```
bushrat011899, line 2702: These `x_assume_mutable` methods are the escape-hatch to let you work with a _possibly_ mutable component. I don't like expanding the unsafe API (especially the public unsafe API), but there isn't really a clean alternative. As mentioned in earlier reviews, this is at least a very small safety invariant to uphold.

examples/ecs/immutable_components_dynamic.rs
```diff
@@ -0,0 +1,75 @@
...
+    // This pipeline takes our component descriptions, registers them, and gets
+    // their ComponentId's.
+    let my_registered_components = my_dynamic_components
+        .into_iter()
+        .map(|(name, size)| {
+            // SAFETY:
+            // - No drop command is required
+            // - The component will store [u8; size], which is Send + Sync
+            let descriptor = unsafe {
+                ComponentDescriptor::new_immutable_with_layout(
```
bushrat011899, line 28: This is all that really changes when working with _immutable_ dynamic components; the `ComponentDescriptor` just needs to include the immutable flag and then everything else works as it did previously.

examples/ecs/immutable_components_dynamic.rs
```diff
@@ -0,0 +1,75 @@
...
+            }
+        });
+    }
+
+    for (_name, _size, component_id) in &my_registered_components {
+        // With immutable components, we can read the values...
+        assert!(entity.get_by_id(*component_id).is_ok());
+
+        // ...but we cannot gain a mutable reference.
+        assert!(entity.get_mut_by_id(*component_id).is_err());
```
bushrat011899, line 67: A runtime-error is as-good-as it gets for dynamic immutable components. Not ideal but there's no compile-time options for hopefully pretty obvious reasons.

bushrat011899 (commented):
Below is some notes to assist reviewers understand the decisions I've made for this PR.

Cargo.toml
```diff
@@ -1943,6 +1943,28 @@ description = "Creates a hierarchy of parents and children entities"
...
+doc-scrape-examples = true
+
+[package.metadata.example.immutable_components_dynamic]
+name = "Immutable Dynamic Components"
+description = "Demonstrates the creation and utility of dynamic immutable components"
+category = "ECS (Entity Component System)"
+wasm = false
+
+[[example]]
+name = "immutable_components"
```
BenjaminBrienen, line 1958: Capitalization inconsistency with the names

crates/bevy_ecs/macros/src/component.rs
```diff
@@ -213,6 +223,7 @@ fn parse_component_attr(ast: &DeriveInput) -> Result<Attrs> {
         on_replace: None,
         on_remove: None,
         requires: None,
+        immutable: false,
```
BenjaminBrienen, line 226: Reading this reminds me of the double negative problem with naming things. Typically we want to use "positive" variable names. Maybe we could invert this and name the field `mutable`?

bushrat011899: My rationale here was `false` is typically what people think of when they think of a default value for a `bool`. I do agree that it could just as easily be `mutable` though.

Cargo.toml
bushrat011899: I'm not sure what you're referring to here sorry. Comparing to the `iter_combinations` example:

```toml
[[example]]
name = "iter_combinations"
path = "examples/ecs/iter_combinations.rs"
doc-scrape-examples = true

[package.metadata.example.iter_combinations]
name = "Iter Combinations"
description = "Shows how to iterate over combinations of query results"
category = "ECS (Entity Component System)"
wasm = true
```

I believe I'm matching the same lower_snake_case for `example.name`, and Title Case for `package.metadata.example.x.name`.

crates/bevy_ecs/macros/src/component.rs
BenjaminBrienen: Personally, `mutable` sounds better because I think that with this PR, Mutable is *adding* functionality, which is a positive action to me.

In the general case, I also prefer `enabled` over `disabled` for similar reasons.

It's a minor detail, and consistency is more important. This is mostly a nit.

Cargo.toml
BenjaminBrienen: Sorry, I wasn't paying enough attention.

bushrat011899: All good! Just confused haha

crates/bevy_ecs/macros/src/component.rs
bushrat011899: No I agree these are good points. I guess my perspective is a little backwards because we're _adding_ immutable components, even though it's really refactoring `Component` to make mutability not implicit (well, except outside of the derive macro, since it'd be annoying to annotate every component with `#[component(mutable)]`)

NthTensor: I am fine to keep it as `#[component(immutable)]` or to change it to `#[component(mutable = false)]`. Personally I prefer the former, but I don't care strongly. We should not have to label mutable components with anything, it should be the default.

BenjaminBrienen (approved):

crates/bevy_ecs/src/reflect/component.rs
```diff
@@ -265,6 +266,8 @@ impl ReflectComponent {
 
 impl<C: Component + Reflect + TypePath> FromType<C> for ReflectComponent {
     fn from_type() -> Self {
+        // TODO: Currently we panic of a component is immutable and you use
+        // reflection to mutate it. Perhaps the mutation methods should be fallible?
```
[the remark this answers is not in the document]
MrGVSV: I agree with 1 personally. I've also been looking into making certain reflection operations themselves readonly (see #14730) so that should make it easier to extend if such changes would affect `ReflectComponent`.

But yeah feel free to punt this off for now. In the meantime can we add some temporary `# Panics` sections to the docs of the affected methods?

crates/bevy_ecs/src/reflect/component.rs
```diff
@@ -265,6 +266,8 @@ impl ReflectComponent {
 
 impl<C: Component + Reflect + TypePath> FromType<C> for ReflectComponent {
     fn from_type() -> Self {
+        // TODO: Currently we panic of a component is immutable and you use
```
MrGVSV, line 269:
```suggestion
        // TODO: Currently we panic if a component is immutable and you use
```

crates/bevy_ecs/src/reflect/component.rs
```diff
@@ -273,12 +276,22 @@ impl<C: Component + Reflect + TypePath> FromType<C> for ReflectComponent {
...
             },
             apply_or_insert: |entity, reflected_component, registry| {
-                if let Some(mut component) = entity.get_mut::<C>() {
+                if !C::Mutability::MUTABLE {
+                    panic!("This component is immutable, you cannot modify it through reflection");
+                }
+
+                if let Some(mut component) = entity.take::<C>() {
                     component.apply(reflected_component.as_partial_reflect());
+                    entity.insert(component);
```
MrGVSV, line 294: Was this intentional?
```suggestion
```

crates/bevy_ecs/src/reflect/component.rs
```diff
@@ -273,12 +276,22 @@ impl<C: Component + Reflect + TypePath> FromType<C> for ReflectComponent {
                 entity.insert(component);
             },
             apply: |mut entity, reflected_component| {
-                let mut component = entity.get_mut::<C>().unwrap();
+                if !C::Mutability::MUTABLE {
+                    panic!("This component is immutable, you cannot modify it through reflection");
```
MrGVSV, line 280: Nit: For debuggability, could we maybe have these panic messages indicate the method that caused the error? Even better would be to provide the name of the component.

Maybe something like: ``{component_name} is immutable, it cannot be modified via `ReflectComponent::apply` ``?

crates/bevy_ecs/src/reflect/component.rs
nakedible: Simply removing the insert would be an error. The earlier version used `entity.get_mut()`, this one uses `entity.take()` so it needs to reinsert it back.

However, something is off here. Either the panic above is not needed, as mutating component through `take() + modify + insert()` is fully supported for immutable components, or this could still keep on using `entity.get_mut()` (unless there's just some reason why it doesn't compile).

What I believe instead is that `apply()` and `apply_or_insert()` should instead work on mutable components, and always do `take() + apply + insert()`. Or if we want to optimize for hypothetical performance, it can be an if that if the component is mutable, then it will be `get_mut() + apply`, otherwise through the take path.

`reflect_mut` should fail for immutable components obviously.

examples/ecs/immutable_components.rs
```diff
@@ -0,0 +1,129 @@
...
+    }
+}
+
+/// When a [`Name`] is inserted, we will add it to our [`NameIndex`].
+///
+/// Since all mutations to [`Name`] are captured by hooks, we know it is not currently
+/// inserted in the index, and its value will not change without triggering a hook.
+fn on_insert_name(mut world: DeferredWorld<'_>, entity: Entity, _component: ComponentId) {
+    let Some(&name) = world.entity(entity).get::<Name>() else {
+        unreachable!()
```
nakedible, line 73: This is more of a Rust question, but why would you prefer this version instead of `unwrap()`? Both need a runtime check (as you aren't using `unreachable_unchecked!()`), both end up being cold code, both will panic with a static message. I'm interested in learning.

nakedible (commented):
I'm very happy with this pull in general.

crates/bevy_ecs/src/reflect/component.rs
MrGVSV: Oh I totally glanced over the `take`! Then yeah the insert works.

Yeah I don't think we should be taking and reinserting. Wouldn't that unnecessarily trigger `OnAdd`/`Added`?

Either way, I do think we should keep the mutability check. Even if not technically necessary, it'll be way easier to remove than to add if we ever need to.

examples/ecs/immutable_components.rs
alice-i-cecile: I think this can be a bit clearer at times, to explain *why* this should never panic. Especially if you add a message in unreachable.

MrGVSV: Yeah it might be worth adding a message here. The comment explains why but it would probably be clearer to also include a short error message in the `unreachable!`

teohhanhui: But in this case, wouldn't `.expect(...)` be clearer? I think `unreachable!` should usually be reserved for cases where it isn't immediately obvious from the control flow that a certain code path is unreachable. Unwrapping an `Option` is not one of those cases...

bushrat011899: I've updated the usage of `unreachable!` in this example to explain why the code is unreachable.

As to `unreachable` vs `expect` vs `panic`, I personally have this hierarchy I like to use:

1. If it should _never_ happen, `unreachable`
2. If I don't _think_ it should happen, `expect`
3. If it _can_ happen but I have no alternative, `panic`

Even though they all produce the same code and behaviour, I view `unreachable` as a stronger assertion than `expect`, and likewise for `panic`. It might be more verbose, but I think it more clearly conveys to anyone reading the code "_Hey I thought about this, it's a bug if this ever panics here_".

But again, just a personal preference, I wouldn't go around changing code to one or the other.

crates/bevy_ecs/src/reflect/component.rs
bushrat011899: I agree. I've updated the messages to call out the component (using `ShortName` for legibility) and the specific method that failed. I still view these as stop-gap measures, as we discussed earlier a `Result`-based fallible rewrite is the long-term solution here.

crates/bevy_ecs/src/reflect/component.rs
bushrat011899: I _think_ I wrote this before I added the `get_mut_assume_mutable` methods to the various `Entity` cells. You're right that it shouldn't `take`/`apply`/`insert` here, since we know via the `panic!` that at this point in the code-path we have a mutable component, making that option unnecessary.

crates/bevy_ecs/src/reflect/component.rs
```diff
@@ -277,7 +277,8 @@ impl<C: Component + Reflect + TypePath> FromType<C> for ReflectComponent {
             },
             apply: |mut entity, reflected_component| {
                 if !C::Mutability::MUTABLE {
-                    panic!("This component is immutable, you cannot modify it through reflection");
+                    let name = disqualified::ShortName::of::<C>();
```
BenjaminBrienen, line 280: I'm going to give you the same nit you mentioned to me :p

We prefer to add an import instead of using the FQN so that we can see, at a glance, what crates are being pulled in for this module.

crates/bevy_ecs/src/reflect/component.rs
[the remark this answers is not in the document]
bushrat011899: Thanks! I've added some `# Panics` sections (or updated existing) to call-out that this will now panic if the component is immutable.

crates/bevy_ecs/src/reflect/component.rs
bushrat011899: My hubris! It hurts!

examples/ecs/immutable_components.rs
NthTensor: This was my reading as well. If my code hits a `panic` I assume my code has a bug. If it hits an `expect` I consider that the library may have a bug. If it hits `unreachable` I begin to wonder if the rust compiler has a bug.

nakedible: Okay, so I haven't missed anything big. Personally, I will not use three lines if one line can do the job.

bushrat011899: Perfectly reasonable @nakedible! This is just a personal preference of mine.

nakedible (approved):

BenjaminBrienen (approved):

crates/bevy_ecs/src/lib.rs
```diff
@@ -46,7 +46,7 @@ pub mod prelude {
     pub use crate::{
         bundle::Bundle,
         change_detection::{DetectChanges, DetectChangesMut, Mut, Ref},
-        component::Component,
+        component::{Component, Immutable, Mutable},
```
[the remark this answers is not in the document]
alice-i-cecile: Let's take them out of the prelude: these are niche and rare.

crates/bevy_ecs/src/component.rs
```diff
@@ -825,6 +891,29 @@ impl ComponentDescriptor {
...
+            immutable: false,
+        }
+    }
+
+    /// Create a new `ComponentDescriptor` for an immutable [`Component`].
+    ///
+    /// # Safety
+    /// - the `drop` fn must be usable on a pointer with a value of the layout `layout`
+    /// - the component type must be safe to access from any thread (Send + Sync in rust terms)
+    pub unsafe fn new_immutable_with_layout(
```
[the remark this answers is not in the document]
alice-i-cecile: I much prefer that design. Break the users!

crates/bevy_ecs/src/component.rs
```diff
@@ -392,6 +420,35 @@ pub trait Component: Send + Sync + 'static {
     }
 }
 
+mod private {
+    pub trait Seal {}
+}
+
+/// The mutability option for a [`Component`]. This can either be:
```
alice-i-cecile, line 427: This should probably be where the main explanation lives. Move the docs here, and link from Component, then link this trait from the other marker types in this module.

crates/bevy_ecs/src/reflect/component.rs
```diff
@@ -180,6 +188,10 @@ impl ReflectComponent {
     }
 
     /// Gets the value of this [`Component`] type from the entity as a mutable reflected reference.
+    ///
+    /// # Panics
```
alice-i-cecile, line 192: Follow-up: I would prefer to make these all return Results instead of panicking.

crates/bevy_ecs/src/reflect/component.rs
```diff
@@ -265,6 +281,8 @@ impl ReflectComponent {
 
 impl<C: Component + Reflect + TypePath> FromType<C> for ReflectComponent {
     fn from_type() -> Self {
+        // TODO: Currently we panic if a component is immutable and you use
```
alice-i-cecile, line 284: Strongly agree!

alice-i-cecile (changes requested):
Strongly in favor of this, and it's really well-made. I have a few small complaints I'd like to see done here, but once those are resolved I'm very happy to merge it.

I'd also like to merge the immutable_components_dynamic example into the immutable_components example, possibly in a module. We have too many examples already lol.

crates/bevy_ecs/src/component.rs
bushrat011899: Agreed! I've moved the documentation around.

cart (approved):
I think this is the best path forward.

Worth noting that the need to manually specify mutability could go away if [associated type defaults](https://github.com/rust-lang/rust/issues/29661) ever lands.

alice-i-cecile (approved):

PR #16373, opened 2024-11-13
add docs to clip_from_view

atlv24:
more docs


JMS55:
Users can also use a custom projection

atlv24:
im aware, i think if a user is advanced enough to do custom projections then they know what they're doing. i cant possibly document what they are doing anyways

JMS55:
Mhm. Just thinking we should maybe leave a comment reminding people about it. Up to you, I don't feel strongly either way.

JMS55:
The other nice thing to add would be how you check for ortho vs perspective in a shader.

atlv24:
addressed both

aevyrie (approved):

JMS55 (approved):

BenjaminBrienen (approved):

JMS55 (approved):

aevyrie (approved):
Docs always good.
Is there anything we can do to avoid having 3 sets of the same docs from getting out of sync? Newtyping? Including a text document or linking to it?

pcwalton (approved):

alice-i-cecile:
> Docs always good. Is there anything we can do to avoid having 3 sets of the same docs from getting out of sync? Newtyping? Including a text document or linking to it?

You can `include_str` various bits of text in docs. In this case though module docs feel more correct.

PR #16375, opened 2024-11-13
Only use physical coords internally in `bevy_ui`

ickshonpe:
# Objective

We switch back and forwards between logical and physical coordinates all over the place. Systems have to query for cameras and the UiScale when they shouldn't need to. It's confusing and fragile and new scale factor bugs get found constantly. 

## Solution

* Use physical coordinates whereever possible in `bevy_ui`. 
* Store physical coords in `ComputedNode` and tear out all the unneeded scale factor calculations and queries.
* Add an `inverse_scale_factor` field to `ComputedNode` and set nodes changed when their scale factor changes.

## Migration Guide

`ComputedNode`'s fields and methods now use physical coordinates.
`ComputedNode` has a new field `inverse_scale_factor`. Multiplying the physical coordinates by the `inverse_scale_factor` will give the logical values.


doup (approved):

BenjaminBrienen (approved):

crates/bevy_ui/src/render/mod.rs
```diff
@@ -553,41 +546,26 @@ pub fn extract_default_ui_camera_view(
...
-            }),
-            Some(physical_size),
-            Some(scale_factor),
-        ) = (
-            camera.logical_viewport_size(),
-            camera.physical_viewport_rect(),
-            camera.physical_viewport_size(),
-            camera.target_scaling_factor(),
-        ) {
+        if let (Some(physical_viewport_rect),) = (camera.physical_viewport_rect(),) {
```
atlv24, line 559: whats the point of the tuple on either side of this let?
```suggestion
        if let Some(physical_viewport_rect) = camera.physical_viewport_rect() {
```

crates/bevy_ui/src/render/mod.rs
```diff
@@ -762,7 +719,7 @@ pub fn extract_text_sections(
                         camera_entity: render_camera_entity.id(),
                         rect,
                         item: ExtractedUiItem::Glyphs {
-                            atlas_scaling: Vec2::splat(inverse_scale_factor),
+                            atlas_scaling: Vec2::splat(1.),
```
atlv24, line 722: I think there's `Vec2::ONE`

atlv24 (approved):
This needs a migration guide. Its a particularly sneaky type of breaking change, because it causes almost no compile errors, but underlying semantics change. @BenjaminBrienen @doup have you tested these changes/run any examples? I haven't yet.

doup (changes requested):
@atlv24 I ran the UI testbed and a couple more examples. But, now I checked some other random example and there are indeed issues. 😬

So… I've run all the examples and these are the issues I've found:

- border width are different (example: `borders`, `ui_texture_atlas`)
- texture atlas slice border width (example: `ui_texture_atlas_slice`, `ui_texture_slice_flip_and_tile`, `ui_texture_slice`)
- border radius are different (example: `box_shadow`)
- text wrapping (examples: `grid` sidebar text, `scroll`, `text_debug`, `text_wrap_debug`)
- wrong text/box (?) alignment, could be because of the wrapping issue (examples: `display_and_visibility`, `text`, `size_constraints`)
- overflow clip margin (example: `overflow_clip_margin`)
- a panic! (example: `overflow_debug`)

ickshonpe:
> @atlv24 I ran the UI testbed and a couple more examples. But, now I checked some other random example and there are indeed issues. 😬
> 
> So… I've run all the UI examples and this is what I've found:
> 
> * border width are different (example: `borders`, `ui_texture_atlas`)
> * texture atlas slice border width (example: `ui_texture_atlas_slice`, `ui_texture_slice_flip_and_tile`, `ui_texture_slice`)
> * border radius are different (example: `box_shadow`)
> * text wrapping (examples: `grid` sidebar text, `scroll`, `text_debug`, `text_wrap_debug`)
> * wrong text/box (?) alignment, could be because of the wrapping issue (examples: `display_and_visibility`, `text`, `size_constraints`)
> * overflow clip margin (example: `overflow_clip_margin`)
> * a panic! (example: `overflow_debug`)

I think all of these are fixed now.

doup (approved):
Everything is fine now! 🎉

atlv24 (approved):
ran the ui examples, all looks good, and migration guide is good!

cart (approved):

Issue #16155, opened 2024-10-29
PCSS issues


JMS55:
## Bevy version

0.15

## Info 
* Calculate the search radius based on the light radius in world units and the distance to the light in world units (for directional lights this should be configurable and perhaps default to the distance to the sun)
* Average depth in world units when finding the blocker
* Use the correct calculation for the filter radius that leverages the penumbra size calculation
* Use blue noise random rotations of the MSAA sampling pattern

From this conversation on discord https://discord.com/channels/691052431525675048/743663924229963868/1300819753480228895

cart:
Given the issues with the implementation, I think we should make PCSS experimental for 0.15 and hold off on announcing it until we resolve these problems.

@pcwalton does that sound reasonable?

superdump:
Thinking about it a bit more now:

> Calculate the search radius based on the light radius in world units and the distance to the light in world units (for directional lights this should be configurable and perhaps default to the distance to the sun)

This can rather be done based on the tangent of the angle of the light source in the sky. So I think we could make it configurable and default to the value for the Sun.

> Average depth in world units when finding the blocker

I think averaging in NDC is fine to do because with orthographic projections the depth is essentially a linear interpolation between 0 and 1 anyway, based on the near and far.

PR #16382, opened 2024-11-14
Make PCSS experimental

cart:
# Objective

PCSS still has some fundamental issues (#16155). We should resolve them before "releasing" the feature.

## Solution

1. Rename the already-optional `pbr_pcss` cargo feature to `experimental_pbr_pcss` to better communicate its state to developers.
2. Adjust the description of the `experimental_pbr_pcss` cargo feature to better communicate its state to developers.
3. Gate PCSS-related light component fields behind that cargo feature, to prevent surfacing them to developers by default.


pcwalton:
This is ok, as long as we don't break it.

BenjaminBrienen (approved):

mockersf (approved):

PR #16385, opened 2024-11-14
Retain `RenderMeshInstance` and `MeshInputUniform` data from frame to frame.

pcwalton:
This commit moves the front end of the rendering pipeline to a retained model when GPU preprocessing is in use (i.e. by default, except in constrained environments). `RenderMeshInstance` and `MeshUniformData` are stored from frame to frame and are updated only for the entities that changed state. This was rather tricky and requires some careful surgery to keep the data valid in the case of removals.

This patch is built on top of Bevy's change detection. Generally, this worked, except that `ViewVisibility` isn't currently properly tracked. Therefore, this commit adds proper change tracking for `ViewVisibility`. Doing this required adding a new system that runs after all `check_visibility` invocations, as no single `check_visibility` invocation has enough global information to detect changes.

On the Bistro exterior scene, with all textures forced to opaque, this patch improves steady-state `extract_meshes_for_gpu_building` from 93.8us to 34.5us and steady-state `collect_meshes_for_gpu_building` from 195.7us to 4.28us. Altogether this constitutes an improvement from 290us to 38us, which is a 7.46x speedup.

Screenshot 2024-11-13 143841

Screenshot 2024-11-13 143850

This patch is only lightly tested and shouldn't land before 0.15 is released anyway, so I'm releasing it as a draft.


crates/bevy_pbr/src/render/mesh.rs
```diff
@@ -1117,6 +1317,16 @@ pub fn extract_meshes_for_gpu_building(
...
+
+    // Also record info about each mesh that became invisible.
+    let mut queue = render_mesh_instance_queues.borrow_local_mut();
+    for entity in removed_visibilities_query
+        .read()
+        .chain(removed_global_transforms_query.read())
+        .chain(removed_meshes_query.read())
+    {
+        queue.remove(entity.into(), any_gpu_culling);
+    }
```
tim-blackbird, line 1329: This needs to check whether the entity has had the components re-added after they were removed.
Currently, adding the following system to an example will cause the meshes to disappear and re-appear every other frame:
```rust
.add_systems(
    Update,
    |query: Query<Entity, With<Mesh3d>>, mut commands: Commands| {
        for entity in &query {
            commands
                .entity(entity)
                .remove::<GlobalTransform>()
                .insert(GlobalTransform::default());
        }
    },
)
```

pcwalton: Good catch, I pushed a fix.

tim-blackbird: I've verified the fix :+1:

tim-blackbird (approved):

crates/bevy_pbr/src/render/mesh.rs
```diff
@@ -563,11 +563,25 @@ pub enum RenderMeshInstanceGpuQueue {
...
+        removed: Vec<MainEntity>,
+    },
     /// The version of [`RenderMeshInstanceGpuQueue`] that contains the
     /// [`MeshCullingData`], used when any view has GPU culling
     /// enabled.
-    GpuCulling(Vec<(MainEntity, RenderMeshInstanceGpuBuilder, MeshCullingData)>),
+    GpuCulling {
+        /// Stores GPU data for each entity that became visible or changed in
+        /// such a way that necessitates updating the [`MeshInputUniform`] (e.g.
+        /// changed transform).
```
JMS55, line 580: The fields of each variant are identical, right? Can we pull this out to a new type?

crates/bevy_pbr/src/render/mesh.rs
```diff
@@ -636,6 +650,21 @@ pub struct RenderMeshInstancesCpu(MainEntityHashMap<RenderMeshInstanceCpu>);
 #[derive(Default, Deref, DerefMut)]
 pub struct RenderMeshInstancesGpu(MainEntityHashMap<RenderMeshInstanceGpu>);
 
+/// The result of an entity removal.
+///
+/// Whenever entity data is removed from the [`MeshInputUniform`] buffer, the
```
JMS55, line 655: Is this so that the buffer is always compact so that shaders can dispatch over it without gaps?

JMS55 (approved):
Generally approve.

Some followups I think we should do in the future (nothing to do this with this PR):
* Hookup MeshletMesh to this API too (it already uses MeshUniform, but it doesn't use the existing queuing system)
* Switch to writing indices into the mesh uniforms per view, instead of copying the uniforms
* Experiment with using entities again instead of resources for storing data
* Rename MeshUniform to MeshInstanceData or something better

crates/bevy_pbr/src/render/mesh.rs
JMS55: If so please document this.

pcwalton: Yep, you're right. Added some documentation.

crates/bevy_pbr/src/render/mesh.rs
```diff
@@ -349,16 +351,14 @@ pub struct MeshInputUniform {
     pub first_vertex_index: u32,
     /// Index of the material inside the bind group data.
     pub material_bind_group_slot: u32,
-    /// Padding.
```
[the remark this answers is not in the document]
pcwalton: Oops, that was a mistake. I added them back.

PR #16386, opened 2024-11-14
Fix sprite picking backend not considering the viewport of the camera.

andriyDev:
# Objective

- When picking sprites, the pointer is offset from the mouse, causing you to pick sprites you're not mousing over!

## Solution

- Shift over the cursor by the minimum of the viewport.

## Testing

- I was already using the bevy_mod_picking PR for my project, so it seems to work!
- I tested this on the sprite_example (making the camera only render to part of the viewport), and it also works there.

## Notes

- This is just https://github.com/aevyrie/bevy_mod_picking/pull/365 but in Bevy form.
- We don't need to renormalize the viewport in any way since the viewport is specified in pixels, so all that matters is that the origin is correct.

### DIFF
--- a/crates/bevy_sprite/src/picking_backend.rs
+++ b/crates/bevy_sprite/src/picking_backend.rs
@@ -71,8 +71,13 @@ pub fn sprite_picking(
             continue;
         };
 
-        let Ok(cursor_ray_world) = camera.viewport_to_world(cam_transform, location.position)
-        else {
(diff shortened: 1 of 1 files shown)


kristoff3r (approved):
I also tested the sprite_picking example with a smaller viewport and can confirm it works

alice-i-cecile (approved):

PR #16388, opened 2024-11-14
Add optional transparency passthrough for sprite backend with bevy_picking

### LINKED ISSUE #14929: Add support for alpha-testing in sprite picking backend
```
              > Not sure what you mean by "test against alpha". Could you elaborate? EDIT - By that, do you mean whether or not the sprite is visible due to alpha/invisible texels? I'll add docs for that.

Exactly. I think most users would expect fully transparent pixels to not be hittable. Might be something where the backend/component specifies a threshold, so any alpha below that value will be ignored. Doesn't need to be added right now, but would be nice to mention in the plugin docs.
```

_Originally posted by @aevyrie in https://github.com/bevyengine/bevy/issues/14757#issuecomment-2297442070_
            
We can either add a whole new backend for this, or add a configuration option to the existing backend. Checking the alpha value of each pixel is more expensive but more intuitive and should be enabled by default with a way to toggle it off.

vandie:
# Objective

- Allow bevy_sprite_picking backend to pass through transparent sections of the sprite.
- Fixes #14929

## Solution

- After sprite picking detects the cursor is within a sprites rect, check the pixel at that location on the texture and check that it meets an optional transparency cutoff. Change originally created for mod_picking on bevy 0.14 (https://github.com/aevyrie/bevy_mod_picking/pull/373)

## Testing

- Ran Sprite Picking example to check it was working both with transparency enabled and disabled
- ModPicking version is currently in use in my own isometric game where this has been an extremely noticeable issue

## Showcase

Sprite Picking Text

## Migration Guide

Sprite picking now ignores transparent regions (with an alpha value less than or equal to 0.1). To configure this, modify the `SpriteBackendSettings` resource.


crates/bevy_sprite/src/picking_backend.rs
```diff
@@ -11,25 +11,50 @@ use bevy_ecs::prelude::*;
...
 
+/// Runtime settings for the [`SpriteBackend`].
+#[derive(Resource, Reflect)]
+#[reflect(Resource, Default)]
+pub struct SpriteBackendSettings {
+    /// When set to `true` picking will ignore any part of a sprite which is transparent
+    /// Off by default for backwards compatibility. This setting is provided to give you fine-grained
+    /// control over if transparency on sprites is ignored.
+    pub passthrough_transparency: bool,
+    /// How Opaque does part of a sprite need to be in order count as none-transparent (defaults to 10)
```
mgi388, line 27: Can you elaborate on the docs what the units are? 10 what? What is a valid range I could use, as a user?

crates/bevy_sprite/src/picking_backend.rs
```diff
@@ -11,25 +11,50 @@ use bevy_ecs::prelude::*;
...
 use bevy_render::prelude::*;
 use bevy_transform::prelude::*;
 use bevy_window::PrimaryWindow;
 
+/// Runtime settings for the [`SpriteBackend`].
+#[derive(Resource, Reflect)]
+#[reflect(Resource, Default)]
+pub struct SpriteBackendSettings {
+    /// When set to `true` picking will ignore any part of a sprite which is transparent
+    /// Off by default for backwards compatibility. This setting is provided to give you fine-grained
```
mgi388, line 24: It is mainly off by default for performance?

crates/bevy_sprite/src/picking_backend.rs
```diff
@@ -11,25 +11,50 @@ use bevy_ecs::prelude::*;
...
+    /// control over if transparency on sprites is ignored.
+    pub passthrough_transparency: bool,
+    /// How Opaque does part of a sprite need to be in order count as none-transparent (defaults to 10)
+    pub transparency_cutoff: u8,
+}
+
+impl Default for SpriteBackendSettings {
+    fn default() -> Self {
+        Self {
+            passthrough_transparency: false,
```
mgi388, line 34: The mix of transparency as a prefix and suffix feels like this should have a different name, but I don’t have any good suggestions right now. 

Also, worth checking if transparency is the right term compared to alpha. No strong opinions but it stood out to me as a question.

mgi388 (commented):
I’m excited for this change. I did a quick visual code review but haven’t tested it yet.

mgi388:
@vandie also see https://github.com/bevyengine/bevy/issues/14929 I think your PR fixes it.

vandie:
> @vandie also see #14929 I think your PR fixes it.

I think it does yeah.

crates/bevy_sprite/src/picking_backend.rs
vandie: I think you're right. Alpha is absolutely a better term. Updating.

crates/bevy_sprite/src/picking_backend.rs
mgi388: I just saw/remembered per the linked issue @alice-i-cecile seems to suggest it is on by default with a way to turn off. As a user, I want this OOTB myself so on by default makes sense to me, for what it’s worth. To avoid churn feel free to wait for other reviews before changing your code.

mgi388:
> > @vandie also see #14929 I think your PR fixes it.
> 
> I think it does yeah.

You can put “Fixes #{id}” in your PR description so merging this PR closes the issue as fixed. 😌

BenjaminBrienen (approved):

crates/bevy_sprite/src/picking_backend.rs
aevyrie: Agree, on by default makes sense as long as it isn't some big perf footgun to read the image data.

aevyrie: I don't think the back compat is needed here, bevy is happy to make breaking changes between releases pre 1.0.

crates/bevy_sprite/src/picking_backend.rs
aevyrie: Can these two options be combined into a single enum? It doesn't make sense to have an alpha cutoff setting when alpha passthrough is disabled. e.g.

```rs
SpritePickingAlphaTest {
    /// Don't test alpha, only consider the rect of sprites.
    Ignore,
    /// Test the pixel alpha when running hit tests. Alpha values below this will not be considered for picking.
    Threshold(u8)
}
```

vandie:
Seems everyone agrees that this should be enabled by default so I've updated that and introduced a new `SpriteBackendAlphaPassthrough` enum so this can be controlled by a single setting rather than multiple.

alice-i-cecile:
Can you add a small migration guide explaining that this is now on by default? Definitely the better default, but could break existing users.

vandie:
> Can you add a small migration guide explaining that this is now on by default? Definitely the better default, but could break existing users.

Added. Hope that's ok?

alice-i-cecile:
The migration guide was clear and effective, but the tone / target audience was a bit weird. The goal is to communicate "why is my stuff broken and how do I fix it", but your initial draft was closer to a change log :) I've quickly edited the PR description incorporating the information you've added. These get automatically scraped / compiled, so it's good to clean them up here.

crates/bevy_sprite/src/picking_backend.rs
```diff
@@ -11,25 +11,56 @@ use bevy_ecs::prelude::*;
...
 use bevy_render::prelude::*;
 use bevy_transform::prelude::*;
 use bevy_window::PrimaryWindow;
 
+/// Should the backend count transparent pixels as part of the sprite for picking purposes
+#[derive(Debug, Clone, Copy, Reflect)]
+pub enum SpriteBackendAlphaPassthrough {
+    /// Even if a sprite is picked on a transparent pixel, it should still count within the backend.
+    /// Only consider the rect of a given sprite.
+    NoPassthrough,
```
aevyrie, line 24: `NoPassthrough` is a bit repetitive. I'd suggest just calling this varient `Disabled` or something. The type name already includes "Passthrough" so this is currently repeating itself. e.g.
`SpriteBackendAlphaPassthrough::Disabled`

aevyrie: If we really want to bikeshed this :), the type name on its own isn't super clear that it has anything to do with picking.

I think something like this is more succinct and descriptive of the two modes available to users:
```
enum SpritePickingMode {
    BoundingBox,
    AlphaThreshold(u8)
}
```

vandie: Hey, bikeshed away. I'll sort it out tomorrow 😅

vandie: Renamed 😄

aevyrie (approved):

crates/bevy_sprite/src/picking_backend.rs
```diff
@@ -135,12 +168,44 @@ pub fn sprite_picking(
...
+                                    .get(&atlas.layout)
+                                    .map(|f| f.textures[atlas.index])
+                            })
+                            .or(Some(URect::new(0, 0, texture.width(), texture.height())))?;
+                        // get mouse position on texture
+                        let texture_position = (texture_rect.center().as_vec2()
+                            + cursor_pos_sprite.trunc())
+                        .as_uvec2();
+                        // grab pixel
+                        let pixel_index =
```
andriyDev, line 189: We can replace this with `texture.get_color_at(image_position.x, image_position.y)`. Then we can sample the alpha value through the color. Images can come in all sorts of formats, so we should handle it through `get_color_at`.

crates/bevy_sprite/src/picking_backend.rs
```diff
@@ -135,12 +168,44 @@ pub fn sprite_picking(
...
+                            (texture_position.y * texture.width() + texture_position.x) as usize;
+                        // check transparency
+                        match texture.data.get(pixel_index * 4..(pixel_index * 4 + 4)) {
+                            // If possible check the alpha bit is above cutoff
+                            Some(pixel_data) if pixel_data[3] > cutoff => true,
```
andriyDev, line 194: Can we move the condition to replace `true`? If there is pixel data the result is just the alpha cutoff expression.

crates/bevy_sprite/src/picking_backend.rs
```diff
@@ -135,12 +168,44 @@ pub fn sprite_picking(
...
+                            .as_ref()
+                            .and_then(|atlas| {
+                                texture_atlas_layout
+                                    .get(&atlas.layout)
+                                    .map(|f| f.textures[atlas.index])
+                            })
+                            .or(Some(URect::new(0, 0, texture.width(), texture.height())))?;
+                        // get mouse position on texture
+                        let texture_position = (texture_rect.center().as_vec2()
+                            + cursor_pos_sprite.trunc())
```
andriyDev, line 186: We need to multiply the cursor_pos_sprite.y by -1. World coordinates go right and up. Image coordinates go right and down. This is why your gif looks a little wonky at the edges of the character sprite (if you zoom in really far).

andriyDev, line 186: Replace `.trunc()` with `.floor()`. The way this is right now, a pixel coordinate of -0.5 gets rounded **up** to 0. This means the pixels in the top left quadrant are shifted one pixel to the top left, resulting in slightly wrong picking.

crates/bevy_sprite/src/picking_backend.rs
```diff
@@ -11,25 +11,58 @@ use bevy_ecs::prelude::*;
...
 use bevy_math::{prelude::*, FloatExt, FloatOrd};
 use bevy_picking::backend::prelude::*;
+use bevy_reflect::prelude::*;
 use bevy_render::prelude::*;
 use bevy_transform::prelude::*;
 use bevy_window::PrimaryWindow;
 
+/// How should the [`SpritePickingPlugin`] handle picking and how should it handle transparent pixels
+#[derive(Debug, Clone, Copy, Reflect)]
+pub enum SpritePickingMode {
```
andriyDev, line 21: Maybe we should add a None type here. Today we have a feature flag for enabling or disabling sprite picking. This makes it difficult to undo that (for example, I wanted to copy this picking behavior and had to disable all the default features and re-enable everything but sprite picking in order to do so).

andriyDev (commented):
Sorry for the previous two comments not being part of this review, I've been noticing while trying to adapt this picking backend into my project.

crates/bevy_sprite/src/picking_backend.rs
vandie: From what I can see, we already convert from world coords to sprite coords above on line 145?

crates/bevy_sprite/src/picking_backend.rs
vandie: I'm happy to go with the general opinion here but I feel like that probably falls outside the scope of this change into a wider conversation about uses of feature flags right?

If you want to disable this change, going back to the old style then `BoundingBox` already does that.

alice-i-cecile: Yeah, let's leave that for another PR :)

crates/bevy_sprite/src/picking_backend.rs
vandie: Valid suggestion. Making the change. (This does unfortunately mean that with .alpha() the config has to be presented as 0-1. More consistent with other areas of Bevy but less so with art programs)

crates/bevy_sprite/src/picking_backend.rs
vandie: Doing some testing, and if you do invert the y value here it leads to frequent out of range errors from bevy image. which leads me further into thinking this isn't the case. happy to be shown wrong though.

andriyDev: I was printing the color and it was sampling the wrong pixel for me until I multiplied by -1. One way to test this is to just use a sprite that is opaque on the top half and transparent on the bottom half. Without the -1 you'll be able to pick the transparent half rather than the opaque half!

vandie: I'll take another look tomorrow, because if so then I also need to figure out the out or bounds issue

MonforteAdrian:
Testing the changes got this error after a crash:
```
2024-11-22T07:39:32.137594Z  INFO bevy_winit::system: Creating new window "App" (0v1#4294967296)
2024-11-22T07:39:32.137772Z  INFO winit::platform_impl::platform::x11::window: Guessed window scale factor: 1
2024-11-22T07:39:32.326951Z  WARN bevy_render::view::window: Couldn't get swap chain texture after configuring. Cause: 'The underlying surface has changed, and therefore the swap chain must be updated'
thread 'Compute Task Pool (4)' panicked at /home/adrian/bevy_testing/bevy/crates/bevy_image/src/image.rs:974:36:
range end index 8 out of range for slice of length 4
Encountered a panic in system `bevy_sprite::picking_backend::sprite_picking`!
```
I'm using 0.15 release branch with the changes applied on top.
The sprite on the right works fine, but the squares and bevy logos works quite bad and make the app crash.
The test run on linux with x11 as you can see in the errors

andriyDev:
> Testing the changes got this error after a crash:

I verified this on windows as well. Looks like there's something wrong with `Image::get_color_at`. Dang! I'll file another issue.

crates/bevy_sprite/src/picking_backend.rs
vandie: Ok so, it seems that this -1 issue is actually a wider issue with sprite picking. When the check is done to see if the cursor is within the rect, the position needs to have been multiplied by -1 on the y axis before that point.

`let is_cursor_in_sprite = rect.contains(cursor_pos_sprite);`

The out of range is caused by the uncorrected y position indicating it's within the rect leading to us here trying to get the pixel data for a pixel that doesn't exist within my change. If we multiply by -1 on the y axis before the rect check, not only does it look like it's working more accurately but also the out of range errors seem to stop.

vandie:
> > Testing the changes got this error after a crash:
> 
> I verified this on Windows and on main as well. Looks like there's something wrong with `Image::get_color_at`. Dang! I'll file another issue.

This is definitely still an issue as we shouldn't panic when trying to access non valid pixel data but I've got a fix coming for this specific change.

crates/bevy_sprite/src/picking_backend.rs
vandie: Those bevy logos seem off though. Going to keep digging.

andriyDev (commented):
@vandie #16475 has been merged so that should fix one set of panics.

crates/bevy_sprite/src/picking_backend.rs
```diff
@@ -183,7 +185,7 @@ pub fn sprite_picking(
                             })
                             .or(Some(URect::new(0, 0, texture.width(), texture.height())))?;
                         // get mouse position on texture
-                        let texture_position = (texture_rect.center().as_vec2()
+                        let texture_position = (texture_rect.center().as_vec2().floor()
```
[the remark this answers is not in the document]
vandie: Yeah you're probably right on this one.

crates/bevy_sprite/src/picking_backend.rs
```diff
@@ -129,18 +163,49 @@ pub fn sprite_picking(
...
+                            .and_then(|atlas| {
+                                texture_atlas_layout
+                                    .get(&atlas.layout)
+                                    .map(|f| f.textures[atlas.index])
+                            })
+                            .or(Some(URect::new(0, 0, texture.width(), texture.height())))?;
+                        // get mouse position on texture
+                        let texture_position = (texture_rect.center().as_vec2().floor()
+                            + cursor_pos_sprite.floor())
+                        .as_uvec2();
```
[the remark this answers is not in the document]
vandie: so as long as we have the rect and cursor correct, this shouldn't ever be negative but that being said. casting does filter negatives from what I can tell.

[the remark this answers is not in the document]
andriyDev: Ah you're quite right! TIL any negative floats casts to 0u32, whether release or debug (I'm assuming this isn't platform specific though, which I doubt)

andriyDev:
I tested the sprite_picking example, looks like the black squares are offset by half its width and height. Color sprites are actually a 1x1 image, so it looks like we're offset by half a pixel somewhere.

crates/bevy_sprite/src/picking_backend.rs
```diff
@@ -129,18 +163,50 @@ pub fn sprite_picking(
...
+                            .texture_atlas
+                            .as_ref()
+                            .and_then(|atlas| {
+                                texture_atlas_layout
+                                    .get(&atlas.layout)
+                                    .map(|f| f.textures[atlas.index])
+                            })
+                            .or(Some(URect::new(0, 0, texture.width(), texture.height())))?;
+                        // get mouse position on texture
+                        let texture_position = (texture_rect.center().as_vec2()
```
andriyDev, line 188: We're getting the center and then casting to a float. Odd image sizes will mean the center won't be at 0.5, 0.5 like we'd expect. I'm pretty sure we need to cast the rect to f32 and then get the center, so `texture_rect.as_rect().center()`.

andriyDev (commented):
We're not considering the custom_size of sprites, or the `rect` of sprites inside the pixel checking code.

I also think we should consider moving all the pixel picking logic into the `Sprite` struct as a method. It seems reasonable to want to reuse "given this relative position, where am I in the texture for this sprite?" (but like with the float part remaining).

crates/bevy_sprite/src/picking_backend.rs
vandie: Hmm, let me check my commits because I think I had some correction for this. Wonder if I accidentally removed it with all the reworking we've done.

andriyDev:
@vandie When you have time, could you take a look at https://github.com/vandie/bevy/pull/2 ? I tested it and it looks like it solves a lot of issues!

crates/bevy_sprite/src/picking_backend.rs
```diff
@@ -133,14 +151,44 @@ pub fn sprite_picking(
...
+                    return None;
+                };
+
+                // Since the pixel space coordinate is `Ok`, we know the cursor is in the bounds of
+                // the sprite.
+
+                let cursor_in_valid_pixels_of_sprite = 'valid_pixel: {
+                    match settings.picking_mode {
+                        SpritePickingMode::AlphaThreshold(cutoff) => {
+                            let Some(image) = images.get(&sprite.image) else {
```
Diddykonga, line 168: What is this the reason for returning true for an valid pixel here?

andriyDev: `Sprite::from_color` returns a defaulted handle. This handle doesn't return a valid image, so returning false here would make picking "color sprites" impossible.

alice-i-cecile: Can we add a comment here?

alice-i-cecile:
Thank you to everyone involved with the authoring or reviewing of this PR! This work is relatively important and needs release notes! Head over to https://github.com/bevyengine/bevy-website/issues/1966 if you'd like to help out.

Issue #15664, opened 2024-10-05
Improve access to animation curves and clips


mweatherley:
## What problem does this solve or what need does it fill?

Right now (especially after #15434) it's hard to do anything with curve data after it's been added to an `AnimationClip`. For the sake of post-processing and general reuse, it would be good if there was a way of obtaining a `Curve` from an `AnimationCurve`. This would allow, for example, the curve to be inspected, sampled, etc. without needing to actually animate something.

## What solution would you like?

Use the type registry to register pointers that allow conversion `dyn AnimationCurve -> dyn Curve<T>` when the type of the property being animated is known extrinsically. Ideally, these could be packaged together with some utilities in order to allow something like a `Box< dyn Curve<Transform>>` to be extracted from an `AnimationClip` together with a target ID. (Note that `Curve<Transform>` is actually not quite the right idea, since not every property of the `Transform` may actually be animated.)

## What alternative(s) have you considered?

Do something hacky and hijack the `AnimationGraph` machinery in order to get it to read out values instead of animating things. Alternatively, attempt to downcast the `AnimationCurve` types directly into the types used by the glTF loader.

mweatherley:
It later occurred to me that you can do this without using the `TypeRegistry` but with some unsafe code instead; that kind of solution is more compact, at least.

PR #16395, opened 2024-11-15
Add sample_clamped method to animation curves, returning Boxed values

mbrea-c:
# Objective

Currently the evaluation of `AnimationCurve`s is coupled to the Bevy animation system. This PR provides an integration point for plugins that would like to reuse the animation loading (e.g. from GLTF) and sampling code but implement a custom animation composition system (the use case I have in mind is [bevy_animation_graph](https://github.com/mbrea-c/bevy_animation_graph)).

## Solution

The `AnimationCurve` trait provides a `sample_clamped` method, which returns a boxed version of the result of sampling the underlying curve:

```rust
fn sample_clamped(&self, t: f32) -> Box<dyn Any>;
```

Each of the provided implementations of `AnimationCurve` return a custom sampled type, e.g.:

```rust
#[derive(Reflect)]
pub struct TranslationCurveSample(Vec3);

impl<C> AnimationCurve for TranslationCurve<C>
where
    C: AnimationCompatibleCurve<Vec3>,
{
    // ...

    fn sample_clamped(&self, t: f32) -> Box<dyn Any> {
        let value = self.0.sample_clamped(t);

        Box::new(TranslationCurveSample(value))
    }
}
```

This fits with the "open polymorphism" approach that the rest of the rest of the animation curves code is using. Users can then check the `TypeId` of the returned type to determine what to do with the sampled value.

#### Closed polymorphism
Provide an enum value as the result of `AnimationCurve::sample_clamped`:

```rust
pub enum CurveSample {
    Translation(Vec3),
    Rotation(Quat),
    Scale(Vec3),
    AnimatableCurve(AnimatableCurveSample),
}
```

However this would be at odds with the current implementation of `AnimationCurve`, where users can implement their own `AnimationCurve` types.

## Testing

- Added a new doctest, run all unit tests.
- Ran animation examples to ensure no regressions.
- No benchmarking done, since the existing animation flow is untouched.
- Pending testing tasks:
  - [ ] `AnimatableCurve` sampling.
  - [ ] `WeightsCurve` sampling.
  - [ ] Test using this API in `bevy_animation_graph`.

---

## Showcase

<details>
  <summary>bevy_animation_graph example</summary>

[bevy_animation_graph](https://github.com/mbrea-c/bevy_animation_graph) evaluates animation curves in "Clip" nodes, and stores the
sampled values for all targets in a `Pose` struct which is made available to
other nodes as an output.

Currently evaluating animation curves in `bevy_animation_graph` involves abusing
the `Reflect` API to access private fields in the curve evaluators:
```rust
fn sample_animation_curve(curve: &VariableCurve, time: f32) -> CurveValue {
    let mut evaluator = curve.0.create_evaluator();
    let evaluator_type_id = curve.0.evaluator_type();
    let evaluator_type_name = evaluator.reflect_type_info().type_path();
    let node_index = AnimationNodeIndex::default();

    curve
        .0
        .apply(evaluator.as_mut(), time, 1., node_index)
        .unwrap();

    if evaluator_type_id == TypeId::of::<TranslationCurveEvaluator>() {
        let t = evaluator
            .as_reflect()
            .reflect_ref()
            .as_struct()
            .unwrap()
            .field("evaluator")
            .unwrap()
            .reflect_ref()
            .as_struct()
            .unwrap()
            .field("stack")
            .unwrap()
            .reflect_ref()
            .as_list()
            .unwrap()
            .get(0)
            .unwrap()
            .reflect_ref()
            .as_struct()
            .unwrap()
            .field("value")
            .unwrap()
            .try_downcast_ref::<Vec3>()
            .unwrap();
        CurveValue::Translation(*t)
    } else if evaluator_type_id == TypeId::of::<RotationCurveEvaluator>() {
        let r = evaluator
            .as_reflect()
            .reflect_ref()
            .as_struct()
            .unwrap()
            .field("evaluator")
            .unwrap()
            .reflect_ref()
            .as_struct()
            .unwrap()
            .field("stack")
            .unwrap()
            .reflect_ref()
            .as_list()
            .unwrap()
            .get(0)
            .unwrap()
            .reflect_ref()
            .as_struct()
            .unwrap()
            .field("value")
            .unwrap()
            .try_downcast_ref::<Quat>()
            .unwrap();
        CurveValue::Rotation(*r)
    } else if evaluator_type_id == TypeId::of::<ScaleCurveEvaluator>() {
        let s = evaluator
            .as_reflect()
            .reflect_ref()
            .as_struct()
            .unwrap()
            .field("evaluator")
            .unwrap()
            .reflect_ref()
            .as_struct()
            .unwrap()
            .field("stack")
            .unwrap()
            .reflect_ref()
            .as_list()
            .unwrap()
            .get(0)
            .unwrap()
            .reflect_ref()
            .as_struct()
            .unwrap()
            .field("value")
            .unwrap()
            .try_downcast_ref::<Vec3>()
            .unwrap();
        CurveValue::Scale(*s)
    } else if evaluator_type_name
        // Need to resort to this because WeightsCurveEvaluator is private
        .starts_with("bevy_animation::animation_curves::WeightsCurveEvaluator")
    {
        let w = evaluator
            .as_reflect()
            .reflect_ref()
            .as_struct()
            .unwrap()
            .field("stack_morph_target_weights")
            .unwrap()
            .try_downcast_ref::<Vec<f32>>()
            .unwrap()
            .clone();
        CurveValue::BoneWeights(w)
    } else {
        panic!("Animatable curve type not yet supported!");
    }
}
```

With this PR, we can instead do the following:

```rust
fn sample_animation_curve(curve: &VariableCurve, time: f32) -> CurveValue {
    let curve_sample = curve.0.sample_clampled(time);
    let sample_type_id = curve_sample.as_ref().type_id();

    if TypeId::of::<TranslationCurveSample>() == sample_type_id {
        let cast_sample = curve_sample.downcast::<TranslationCurveSample>().unwrap();
        CurveValue::Translation(cast_sample.0)
    } else if TypeId::of::<RotationCurveSample>() == sample_type_id {
        let cast_sample = curve_sample.downcast::<RotationCurveSample>().unwrap();
        CurveValue::Rotation(cast_sample.0)
    } else if TypeId::of::<ScaleCurveSample>() == sample_type_id {
        let cast_sample = curve_sample.downcast::<ScaleCurveSample>().unwrap();
        CurveValue::Scale(cast_sample.0)
    } else if TypeId::of::<WeightsCurveSample>() == sample_type_id {
        let cast_sample = curve_sample.downcast::<WeightsCurveSample>().unwrap();
        CurveValue::Weights(cast_sample.0)
    } else {
        panic!("Animatable curve type not yet supported!");
    }
}
```

Furthermore, the API in this PR opens the door to adding support for
`AnimatableCurve`s, whereas before this was not possible (or at least I could
not find a way).

</details>

### DIFF
--- a/crates/bevy_animation/src/animation_curves.rs
+++ b/crates/bevy_animation/src/animation_curves.rs
@@ -386,7 +386,7 @@ pub struct TranslationCurveEvaluator {
 /// You shouldn't need to interact with this type unless you're manually evaluating animation
 /// curves.
 #[derive(Reflect)]
-pub struct TranslationCurveSample(Vec3);
+pub struct TranslationCurveSample(pub Vec3);
 
 impl<C> AnimationCurve for TranslationCurve<C>
 where
@@ -497,7 +497,7 @@ pub struct RotationCurveEvaluator {
 /// You shouldn't need to interact with this type unless you're manually evaluating animation
 /// curves.
 #[derive(Reflect)]
-pub struct RotationCurveSample(Quat);
+pub struct RotationCurveSample(pub Quat);
 
 impl<C> AnimationCurve for RotationCurve<C>
 where
@@ -608,7 +608,7 @@ pub struct ScaleCurveEvaluator {
 /// You shouldn't need to interact with this type unless you're manually evaluating animation
 /// curves.
 #[derive(Reflect)]
-pub struct ScaleCurveSample(Vec3);
+pub struct ScaleCurveSample(pub Vec3);
 
 impl<C> AnimationCurve for ScaleCurve<C>
 where
@@ -746,7 +746,7 @@ struct WeightsCurveEvaluator {
 /// You shouldn't need to interact with this type unless you're manually evaluating animation
 /// curves.
 #[derive(Reflect)]
-pub struct WeightsCurveSample(Vec<f32>);
+pub struct WeightsCurveSample(pub Vec<f32>);
 
 impl<C> AnimationCurve for WeightsCurve<C>
 where
@@ -1094,6 +1094,44 @@ pub trait AnimationCurve: Reflect + Debug + Send + Sync {
         graph_node: AnimationNodeIndex,
     ) -> Result<(), AnimationEvaluationError>;
 
+    /// Provides a type-erased API for sampling an [`AnimationCurve`]. This is not usually
+    /// needed if you're using the animation system normally, but can be useful as an integration
+    /// point for animation plugins.
+    ///
+    /// Example usage:
+    /// ```rust
+    /// # use bevy_math::Vec3;
+    /// # use bevy_animation::prelude::{
+    /// #     TranslationCurve,
+    /// #     TranslationCurveSample,
+    /// #     VariableCurve,
+    /// #     AnimatableKeyframeCurve
+    /// # };
+    /// # use core::any::TypeId;
+    /// let translation_curve = TranslationCurve(
+    ///     AnimatableKeyframeCurve::new([
+    ///         (0., Vec3::splat(0.)),
+    ///         (1., Vec3::splat(1.)),
+    ///         (2., Vec3::splat(4.)),
+    ///         (3., Vec3::splat(3.)),
+    ///     ])
+    ///     .unwrap(),
+    /// );
+    ///
+    /// // Suppose we don't know the type of the curve
+    /// let curve = VariableCurve(Box::new(translation_curve));
+    ///
+    /// let curve_sample = curve.0.sample_clamped(2.);
+    ///
+    /// assert_eq!(
+    ///     TypeId::of::<TranslationCurveSample>(),
+    ///     curve_sample.as_ref().type_id()
+    /// );
+    ///
+    /// let translation_curve_sample = curve_sample.downcast::<TranslationCurveSample>().unwrap();
+    ///
+    /// assert_eq!(Vec3::splat(4.), translation_curve_sample.0);
+    /// ```
     fn sample_clamped(&self, t: f32) -> Box<dyn Any>;
 }
 
--- a/crates/bevy_animation/src/animation_curves.rs
+++ b/crates/bevy_animation/src/animation_curves.rs
@@ -81,6 +81,7 @@ use core::{
     fmt::{self, Debug, Formatter},
     marker::PhantomData,
 };
+use std::any::Any;
 
 use bevy_ecs::{component::Component, world::Mut};
 use bevy_math::{
@@ -209,6 +210,21 @@ where
     phantom: PhantomData<P>,
 }
 
+pub struct AnimatableCurveSample {
+    pub component_type: TypeId,
+    pub property_type: TypeId,
+    pub get_mut_boxed: fn(&mut dyn Any) -> Option<&mut dyn Reflect>,
+    pub value: Box<dyn Reflect>,
+}
+
+fn get_mut_boxed<P: AnimatableProperty>(component: &mut dyn Any) -> Option<&mut dyn Reflect> {
+    if let Some(component) = component.downcast_mut::<P::Component>() {
+        P::get_mut(component).map(|p| p.as_reflect_mut())
+    } else {
+        panic!("Type mismatch")
+    }
+}
+
 impl<P, C> AnimatableCurve<P, C>
 where
     P: AnimatableProperty,
@@ -294,6 +310,17 @@ where
             });
         Ok(())
     }
+
+    fn sample_clamped(&self, t: f32) -> Box<dyn Any> {
+        let value = self.curve.sample_clamped(t);
+
+        Box::new(AnimatableCurveSample {
+            component_type: TypeId::of::<P::Component>(),
+            property_type: TypeId::of::<P::Property>(),
+            get_mut_boxed: get_mut_boxed::<P>,
+            value: Box::new(value),
+        })
+    }
 }
 
 impl<P> AnimationCurveEvaluator for AnimatableCurveEvaluator<P>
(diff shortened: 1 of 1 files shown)


mweatherley:
This is fairly close to what I had imagined in this direction (see also #15664 ); another idea would be to yield a `dyn Curve<T>` in some way from a `dyn AnimationCurve` when you have extrinsic knowledge of the return type; e.g.
```rust
fn downcast_curve<T>(self) -> Option<Box<dyn Curve<T>>> {
    // yield a boxed version of the underlying curve if `T` matches the curve type, otherwise None
}
```
I think this is a bit more reusable, but the use-cases might be slightly different, since the dependency on the curve type is a bit different. The main upside with something like this would be that you can do other stuff with the curve API instead of only having the output value. (The main downside in my view is that implementing this requires unsafe code, since this is effectively a "custom `Any`".)

mbrea-c:
> This is fairly close to what I had imagined in this direction (see also #15664 ); another idea would be to yield a `dyn Curve<T>` in some way from a `dyn AnimationCurve` when you have extrinsic knowledge of the return type; e.g.
> 
> ```rust
> fn downcast_curve<T>(self) -> Option<Box<dyn Curve<T>>> {
>     // yield a boxed version of the underlying curve if `T` matches the curve type, otherwise None
> }
> ```
> 
> I think this is a bit more reusable, but the use-cases might be slightly different, since the dependency on the curve type is a bit different. The main upside with something like this would be that you can do other stuff with the curve API instead of only having the output value. (The main downside in my view is that implementing this requires unsafe code, since this is effectively a "custom `Any`".)

I like that, but I think there's value in having both. I think only exposing a `Box<dyn Curve<T>>` wouldn't cover the use case I'm thinking of:
1. My main issue is that we may not have have extrinsic knowledge of the return type; e.g. you have an `AnimationClip` and want to evaluate all curves in it; all you know is that there's a bunch of `VariableCurve`s, which could have as output type a `Vec3`, `Quat`, `Vec<f32>` or any other arbitrary property (`AnimatableProperty::Property`). The API I proposed returns a boxed value whose type you can check, so it works in this case. Also, you have a single boxed type for all `AnimatableCurve` types (regardless of `AnimatableProperty`), which currently are hard to deal with without knowing the type in advance as the `AnimatableCurveEvaluator`'s type depends on the `AnimatableCurve` type parameter.
2. Even if return types are equal, we would want to differentiate between different kinds of `AnimationCurve` in order to apply the result correctly (e.g. translation and scale are both `Vec3`). This one we could workaround by checking the type of the evaluator, but that feels more hacky (maybe that's just me).
3.  `WeightsCurve` uses an `IterableCurve<f32>` instead of a `Curve<Vec<f32>>`. I haven't looked through _all_ of the new `Curve` code yet so maybe there's a way to adapt the former to the latter? Otherwise this seems like a problem for this API.

What do you think?

mweatherley:
>     1. My main issue is that we may not have have extrinsic knowledge of the return type; e.g. you have an `AnimationClip` and want to evaluate all curves in it; all you know is that there's a bunch of `VariableCurve`s, which could have as output type a `Vec3`, `Quat`, `Vec<f32>` or any other arbitrary property (`AnimatableProperty::Property`). The API I proposed returns a boxed value whose type you can check, so it works in this case. Also, you have a single boxed type for all `AnimatableCurve` types (regardless of `AnimatableProperty`), which currently are hard to deal with without knowing the type in advance as the `AnimatableCurveEvaluator`'s type depends on the `AnimatableCurve` type parameter.

That makes sense. I have been kind of wondering if the right direction for the animation system long-term would be to tie every animation curve to a specific `AnimatableProperty` type (instead of making `Transform` parts behave specially, for example); that way, the relationship between an `AnimationCurve` and its curve output would at least be reified in a sort of natural way (as the `Property` type of the `AnimatableProperty`). 

I don't think having both is unreasonable :)

>     2. Even if return types are equal, we would want to differentiate between different kinds of `AnimationCurve` in order to apply the result correctly (e.g. translation and scale are both `Vec3`). This one we could workaround by checking the type of the evaluator, but that feels more hacky (maybe that's just me).

Yeah, this is another reason to want to tie things to `AnimatableProperty`.

>     3. `WeightsCurve` uses an `IterableCurve<f32>` instead of a `Curve<Vec<f32>>`. I haven't looked through _all_ of the new `Curve` code yet so maybe there's a way to adapt the former to the latter? Otherwise this seems like a problem for this API.

Every `IterableCurve<f32>` can also be a `Curve<Vec<f32>>` just by calling `collect`, which is probably what I would do for this API, even if it involves special-casing. There isn't really a way around this: glTF morph weights curves don't even have a `'static` return type (since the returned iterator has a lifetime dependence on the curve itself), so they are like `Curve<T>` where `T: !Any`.

(For some background, the only reason that `IterableCurve` exists is that it would be preferable for these curves to just be `Curve<T>` where `T` is an iterator, but the type of that iterator is made impossible to name by the Rust compiler. Hopefully that will change somewhat in the future.)

mbrea-c:
> That makes sense. I have been kind of wondering if the right direction for the animation system long-term would be to tie every animation curve to a specific AnimatableProperty type (instead of making Transform parts behave specially, for example); that way, the relationship between an AnimationCurve and its curve output would at least be reified in a sort of natural way (as the Property type of the AnimatableProperty).

I don't know what the reasons were for keeping the Transform and weight curves separate tbh (performance maybe?), but I agree that having a more "natural" relationship between the `AnimationCurve` and its output type would be a good thing. Still worried about the dynamic use case though&mdash;unless I'm missing something, having a `P: AnimatableProperty` type parameter on all curves would make type id comparisons complicated when there's no extrinsic type information about the curve&mdash;, what would be the API you're thinking of?

(sorry for the long time between replies btw, hard to find time after day job with end of year deadlines approaching)

mbrea-c:
I'm going to close this one as it needs a rewrite based the discussion in the linked cart PR

PR #15207, opened 2024-09-15
Add `EntityRefExcept` and `EntityMutExcept` world queries, in preparation for generalized animation.

pcwalton:
This commit adds two new `WorldQuery` types: `EntityRefExcept` and `EntityMutExcept`. These types work just like `EntityRef` and `EntityMut`, but they prevent access to a statically-specified list of components. For example, `EntityMutExcept<(AnimationPlayer, Handle<AnimationGraph>)>` provides mutable access to all components except for `AnimationPlayer` and `Handle<AnimationGraph>`. These types are useful when you need to be able to process arbitrary queries while iterating over the results of another `EntityMut` query.

The motivating use case is *generalized animation*, which is an upcoming feature that allows animation of any component property, not just rotation, translation, scaling, or morph weights. To implement this, we must change the current `AnyOf<(&mut Transform, &mut MorphWeights)>` to instead be `EntityMutExcept<(AnimationPlayer, Handle<AnimationGraph>)>`. It's possible to use `FilteredEntityMut` in conjunction with a dynamically-generated system instead, but `FilteredEntityMut` isn't optimized for the use case of a large number of allowed components coupled with a small set of disallowed components. No amount of optimization of `FilteredEntityMut` produced acceptable performance on the `many_foxes` benchmark. `Query<EntityMut, Without<AnimationPlayer>>` will not suffice either, as it's legal and idiomatic for an `AnimationTarget` and an `AnimationPlayer` to coexist on the same entity.

An alternate proposal was to implement a somewhat-more-general `Except<Q, CL>` feature, where Q is a `WorldQuery` and CL is a `ComponentList`. I wasn't able to implement that proposal in a reasonable way, because of the fact that methods like `EntityMut::get_mut` and `EntityRef::get` are inherent methods instead of methods on `WorldQuery`, and therefore there was no way to delegate methods like `get` and `get_mut` to the inner query in a generic way. Refactoring those methods into a trait would probably be possible. However, I didn't see a use case for a hypothetical `Except` with arbitrary queries: `Query<Except<(&Transform, &Visibility), Visibility>>` would just be a complicated equivalent to `Query<&Transform>`, for instance. So, out of a desire for simplicity, I omitted a generic `Except` mechanism.

[...]

PR #9774, opened 2023-09-12
Dynamic queries and builder API

james-j-obrien:
# Objective
Expand the existing `Query` API to support more dynamic use cases i.e. scripting.

## Prior Art
 - #6390 
 - #8308 
- #10037

## Solution
- Create a `QueryBuilder` with runtime methods to define the set of component accesses for a built query.
- Create new `WorldQueryData` implementations `FilteredEntityMut` and `FilteredEntityRef` as variants of `EntityMut` and `EntityRef` that provide run time checked access to the components included in a given query.
- Add new methods to `Query` to create "query lens" with a subset of the access of the initial query.

### Query Builder
The `QueryBuilder` API allows you to define a query at runtime. At it's most basic use it will simply create a query with the corresponding type signature:
```rust
let query = QueryBuilder::<Entity, With<A>>::new(&mut world).build();
// is equivalent to
let query = QueryState::<Entity, With<A>>::new(&mut world);
```
Before calling `.build()` you also have the opportunity to add additional accesses and filters. Here is a simple example where we add additional filter terms:
```rust
let entity_a = world.spawn((A(0), B(0))).id();
let entity_b = world.spawn((A(0), C(0))).id();

let mut query_a = QueryBuilder::<Entity>::new(&mut world)
    .with::<A>()
    .without::<C>()
    .build();
            
assert_eq!(entity_a, query_a.single(&world));
```
This alone is useful in that allows you to decide which archetypes your query will match at runtime. However it is also very limited, consider a case like the following:
```rust
let query_a = QueryBuilder::<&A>::new(&mut world)
// Add an additional access
    .data::<&B>()
    .build();
```
This will grant the query an additional read access to component B however we have no way of accessing the data while iterating as the type signature still only includes &A. For an even more concrete example of this consider dynamic components:
```rust
let query_a = QueryBuilder::<Entity>::new(&mut world)
// Adding a filter is easy since it doesn't need be read later
    .with_id(component_id_a)
// How do I access the data of this component?
    .ref_id(component_id_b)
    .build();
```
With this in mind the `QueryBuilder` API seems somewhat incomplete by itself, we need some way method of accessing the components dynamically. So here's one:
### Query Transmutation
If the problem is not having the component in the type signature why not just add it? This PR also adds transmute methods to `QueryBuilder` and `QueryState`. Here's a simple example:
```rust
world.spawn(A(0));
world.spawn((A(1), B(0)));
let mut query = QueryBuilder::<()>::new(&mut world)
    .with::<B>()
[...]

PR #16396, opened 2024-11-15
Allow dense iteration for FilteredEntity(Ref|Mut) in more cases.

chescock:
# Objective

Improve the performance of some dynamic queries with `FilteredEntity(Ref|Mut)` by allowing dense iteration in more cases, and remove a call to the sort-of deprecated `Access::component_reads_and_writes()` method.  

`QueryBuilder` currently requires sparse iteration if any sparse set components may be read.  We do need sparse iteration if sparse set components are used in the filters, but `FilteredEntityRef` can still perform dense iteration when reading optional components or when reading all components.  

Note that the optional case is different from `Option`, which performs sparse iteration when the inner query is sparse so that it can cache whether the inner query matches for an entire archetype.  

## Solution

Change `FilteredEntity(Ref|Mut)` to have `IS_DENSE = true`.  It used to require sparse iteration in order to filter the `Access` for each archetype, but #15207 changed it to copy the entire access.  

Change `QueryBuilder::is_dense()` to check `D::IS_DENSE && F::IS_DENSE` instead of looking at the component reads and writes.  
`QueryBuilder::is_dense()` still checks the *filters*, so `builder.data::<&Sparse>()` will still cause sparse iteration, but `builder.data::<Option<&Sparse>>()` no longer will.  

I believe this is sound, even in the presence of query transmutes.  The only `WorldQuery` implementations that rely on a sparse query being sparse for soundness are `&`, `&mut`, `Ref`, and `Mut`, but they can only be transmuted to if the component is in the `required` set.  If a dynamic query has the component in the `required` set, then it appears in the filters and the query will use sparse iteration.  

Note that `Option` and `Has` will misbehave and report `None` and `false` for all entities if they do a dense query while wrapping a sparse component, but they won't cause UB.  And it's already possible to hit that case by transmuting from `Query<EntityMut>` to `Query<Option<&Sparse>>`.

### DIFF
--- a/crates/bevy_ecs/src/query/builder.rs
+++ b/crates/bevy_ecs/src/query/builder.rs
@@ -81,16 +81,8 @@ impl<'w, D: QueryData, F: QueryFilter> QueryBuilder<'w, D, F> {
                 .map_or(false, |info| info.storage_type() == StorageType::Table)
         };
 
-        #[allow(deprecated)]
-        let (mut component_reads_and_writes, component_reads_and_writes_inverted) =
-            self.access.access().component_reads_and_writes();
-        if component_reads_and_writes_inverted {
-            return false;
-        }
-
-        component_reads_and_writes.all(is_dense)
-            && self.access.access().archetypal().all(is_dense)
-            && !self.access.access().has_read_all_components()
+        D::IS_DENSE
+            && F::IS_DENSE
             && self.access.with_filters().all(is_dense)
             && self.access.without_filters().all(is_dense)
     }
--- a/crates/bevy_ecs/src/query/fetch.rs
+++ b/crates/bevy_ecs/src/query/fetch.rs
@@ -611,7 +611,7 @@ unsafe impl<'a> WorldQuery for FilteredEntityRef<'a> {
         fetch
     }
 
-    const IS_DENSE: bool = false;
+    const IS_DENSE: bool = true;
 
     unsafe fn init_fetch<'w>(
         world: UnsafeWorldCell<'w>,
@@ -706,7 +706,7 @@ unsafe impl<'a> WorldQuery for FilteredEntityMut<'a> {
         fetch
     }
 
-    const IS_DENSE: bool = false;
+    const IS_DENSE: bool = true;
 
     unsafe fn init_fetch<'w>(
         world: UnsafeWorldCell<'w>,


alice-i-cecile:
> Note that Option and Has will misbehave and report None and false for all entities if they do a dense query while wrapping a sparse component, but they won't cause UB. And it's already possible to hit that case by transmuting from Query<EntityMut> to Query<Option<&Sparse>>.

Can you open an issue for this? :)

Victoronz:
>  And it's already possible to hit that case by transmuting from `Query<EntityMut>` to `Query<Option<&Sparse>>`.

This is possible? This is not documented under `Query::transmute_lens`, and I've never heard of this kind of transmute before, do you know what PR introduced `Query<EntityMut>` -> `Query<Option<T>>` transmutes?

chescock:
> This is possible? This is not documented under `Query::transmute_lens`, and I've never heard of this kind of transmute before, do you know what PR introduced `Query<EntityMut>` -> `Query<Option<T>>` transmutes?

It does work!  `EntityMut` has write access to all components, but doesn't require any of them.  So it can't transmute to `&T`, which requires `T`, but it can transmute to `Option<&T>`, which reads it but does not require it.  

I confess I only looked at the implementation and not the documentation, so I didn't realize it wasn't documented :).  It looks like the subset check hasn't changed much since [transmutes were introduced](https://github.com/bevyengine/bevy/pull/9774/files#diff-5fcd1a28865bd71678344bb254b0b7588eb3600f556c60273a289391cc18fa92R358), so I think it's always worked.  

Looking at the documentation now, that list doesn't seem especially exhaustive.  You can also add `Has<T>` for any `T`, or go from `AnyOf<(T, U)>` to `Option<T>`.  And there can always be third-party `WorldQuery` implementations.

crates/bevy_ecs/src/query/builder.rs
```diff
@@ -81,15 +81,8 @@ impl<'w, D: QueryData, F: QueryFilter> QueryBuilder<'w, D, F> {
...
-        let (mut component_reads_and_writes, component_reads_and_writes_inverted) =
-            self.access.access().component_reads_and_writes();
-        if component_reads_and_writes_inverted {
-            return false;
-        }
-
-        component_reads_and_writes.all(is_dense)
-            && self.access.access().archetypal().all(is_dense)
-            && !self.access.access().has_read_all_components()
+        D::IS_DENSE
```
cBournhonesque, line 84: I think we might want more comments here to explain why this is ok

chescock: Good idea!  I added some comments here, and some in `QueryState::transmute_filtered()`.  (I thought to write up an explanation in the PR description, but I didn't think to put it in the code...)

crates/bevy_ecs/src/query/state.rs
```diff
@@ -645,11 +645,19 @@ impl<D: QueryData, F: QueryFilter> QueryState<D, F> {
             core::any::type_name::<(NewD, NewF)>(), core::any::type_name::<(D, F)>()
         );
 
+        // For transmuted queries, the dense-ness of the query is equal to the dense-ness of the original query.
+        // The check for `required` in `component_access.is_subset()` should ensure that this is always sound:
```
cBournhonesque, line 649: I don't find that part super straightforward; how does knowing that the `Lens`'s required is a subset of the `Query`'s required enough to determine that this is sound?

Is it because `dense-ness` only depends on required components (as for those components we need to iterate only through archetypes that have that component)? 
And for optional components we can just always just iterate through the tables.

I think this might be key point that i'm missing

crates/bevy_ecs/src/query/builder.rs
cBournhonesque: I might even add that `FilteredEntityRef|Mut` (which is one of the primary use-cases of QueryBuilder) has D::IS_DENSE=True, so adding `Option<&Sparse>` will maintain dense-ness.

(IIUC, the key point being, as i wrote in the other comment, that dense-ness only depends on required components (because for those components we need to iterate only through the archetypes that have that component, and not through all tables)?
And for optional components we can just always just iterate through the tables.)

Should the check consider all `required` components on top of filters?
Oh I see, it's what you were mentioning, for "normal" query data like `&, Ref, &mut ,Mut` `D::IS_DENSE` is enough (since those query data add the access to required)

cBournhonesque (commented):
I wonder if there is a way to write a test for it?

1) we could have a small test showing that `FilteredEntityRef` with `Option<&Sparse>` access still is dense
2) but how can we show that it doesn't miss any entity during iteration?

crates/bevy_ecs/src/query/state.rs
chescock: > Is it because `dense-ness` only depends on required components

Ooh, that's a good way to distill the idea!  I think that's right, but I hadn't realized I could put it that way.  

Any `QueryData` or `QueryFilter` that relies on sparse iteration for soundness will register a sparse component in `required`.  And any query with a sparse component in `required` will necessarily perform sparse iteration.  So if `D` or `F` require `is_dense = false`, then they `required` a spare component, and the `is_subset()` check means the *original* query had that component `required`, which means the original query had `is_dense = false`.  

`Option<&Sparse>` and `Has<Sparse>` are still weird because they have `IS_DENSE = false` and will normally trigger sparse iteration, but they don't register any `required` components.  But they aren't *unsound* under dense iteration, so that's okay.  

And `With` and `Without` filters will affect "dense-ness" but are ignored during transmutes.  But, again, they don't affect soundness.  

Let me try to update the comments again with that idea in mind!  Thanks!

cBournhonesque (approved):

PR #16400, opened 2024-11-16
Expose BRP system scheduling and add system set

villor:
# Objective
When adding custom BRP methods one might need to:
- Run custom systems in the `RemoteLast` schedule.
- Order those systems before/after request processing and cleanup.

For example in `bevy_remote_inspector` we need a way to perform some preparation _before_ request processing. And to perform cleanup _between_ request processing and watcher cleanup.

## Solution

- Make `RemoteLast` public
- Add `RemoteSet` with `ProcessRequests` and `Cleanup` variants.


mweatherley (approved):
Nice! Thanks for your continued work on this :)

mockersf (approved):

PR #16407, opened 2024-11-16
Add BRP `watch_id` and `bevy/unwatch`

villor:
# Objective
BRP supports watching requests by using the `+watch`-suffix. But it has some limitations:
- There is no way to differentiate one watcher from another, other than by using the client provided params. This makes it hard to deal with any needed state for ongoing watching requests, as well as detecting when a watcher is closed.
- Watching requests can not be closed/canceled manually, which can be problematic when using transports where multiple requests are communicated over a single connection (see #16403)

## Solution

- Add a _server owned_ `watch_id` (`RemoteWatchingRequestId = u32`) used to uniquely identify a specific watcher.
  - Not to be confused with the `id` field in JSON RPC 2.0, which is a client owned id that may or may not exist and might not be unique.
- Change the `+watch`-request handling to always send an initial response with the `watch_id` of that request.
```json
{
  "jsonrpc": "2.0",
  "id": 5,
  "result": {
    "watch_id": 1
  }
}
```
- Update the `In`-param for the watching request handlers to also include the `watch_id`.
- Add a `bevy/unwatch`-method which can be used to close a current watcher (or all watchers).

## Testing

- Tested together with #16403.
  - ✅ +watch-requests now return their watch_id
  - ✅ bevy/unwatch works for both a specific watch_id, and for all watchers
- I have also used this approach while experimenting with `bevy_remote_inspector`, where I use the watch_id to detect closed watchers and perform cleanup in a separate system.

---


crates/bevy_remote/src/lib.rs
```diff
@@ -524,9 +548,38 @@ impl RemoteMethods {
...
+/// Sent in the initial response of all watching requests.
+///
+/// Not to be confused with the `id` field in JSON-RPC 2.0, which is a client owned ID that might be unset or contain duplicates.
+pub type RemoteWatchingRequestId = u32;
+
+/// Holds the [`BrpMessage`]'s of all ongoing watching requests along with their IDs and handlers.
 #[derive(Debug, Resource, Default)]
-pub struct RemoteWatchingRequests(Vec<(BrpMessage, RemoteWatchingMethodSystemId)>);
+pub struct RemoteWatchingRequests(
+    Vec<(
```
alice-i-cecile, line 561: We should change this to a named struct.

alice-i-cecile (approved):
LGTM, but if you have time swapping to a named struct to avoid the complex tuples would be nice.

crates/bevy_remote/src/lib.rs
villor: Done!

crates/bevy_remote/src/builtin_methods.rs
```diff
@@ -348,7 +364,7 @@ pub fn process_remote_get_request(In(params): In<Option<Value>>, world: &World)
 
 /// Handles a `bevy/get+watch` request coming from a client.
 pub fn process_remote_get_watching_request(
-    In(params): In<Option<Value>>,
+    In((_, params)): In<RemoteWatchingSystemParams>,
```
cBournhonesque, line 367: Why are we providing the `watch_id` if they are not used anywhere in the watch systems?
In case they might be used in the future?

crates/bevy_remote/src/builtin_methods.rs
```diff
@@ -793,6 +809,28 @@ pub fn process_remote_list_watching_request(
...
+        _ => None,
+    };
+
+    for request in watching_requests.0.iter() {
+        if watch_id.is_none() || watch_id.unwrap() == request.watch_id {
+            request.message.sender.close();
+        }
+    }
+
+    Ok(Value::Null)
```
cBournhonesque, line 831: We might want to return an error if a `watch_id` is provided that doesn't exist

crates/bevy_remote/src/lib.rs
```diff
@@ -256,6 +256,17 @@
...
+//! ### bevy/unwatch
+//!
+//! Cancels the watcher (`+watch`-suffixed request) with the supplied `watch_id`.
+//!
+//! When `params` is not provided, all running watchers will be canceled.
+//!
+//! `params`:
+//! - `watch_id`: The ID of request to unwatch.
+//!
+//! `result`: null.
```
cBournhonesque, line 268: We might want to return an error if a watch_id is provided that doesn't exist

cBournhonesque (approved):

crates/bevy_remote/src/builtin_methods.rs
villor: The idea is to give watch handlers a way to identify _which_ watcher it is handling, giving them a unique key for use with any needed state/cache that needs to be handled on watcher level, rather than method level.

As you’ve noticed it is not used by any of the built in methods, as they are not stateful. But I have encountered the need for watcher state while building some custom methods.

I’m not completely certain this is the best solution though. There are other ways. For example each watcher could have their own SystemState, allowing use of Local<T> instead of managing state using ids and hash maps. Another benefit is cleanup would be automatic.

villor:
Since in-engine websocket transport _might_ be off the table, this companion PR may seem unnecessary as well. However I believe this functionality is still useful, for any third-party stream-based transports, as well as a potential future binary TCP transport.

While [trying to think](https://github.com/bevyengine/bevy/pull/16403#issuecomment-2515000562) about ways to solve the connection limit issue with SSE, I realized one way could be to re-design watchers a bit by using a single SSE endpoint as a stream for **all** watcher responses. This would ensure any number of watchers can be used even from a web browser.

With the above in mind, as well as the [state stuff](https://github.com/bevyengine/bevy/pull/16407#discussion_r1895813210), I think we should hold off a bit on this PR and try to figure out a better watcher design in general.

ChristopherBiscardi:
as per above comments, it seems like this requires design work and isn't ready for a PR at this stage.

PR #16410, opened 2024-11-17
Fix bubbling of runtime requirements for `#[require(...)]` attribute

### LINKED ISSUE #16406: Runtime required components do not work correctly for components added through `#[require(...)]`
## Bevy version

Bevy 0.15.0-rc.3

## What you did

## What went wrong

```rust
#[test]
    // `A` requires `B` directly.
    #[derive(Component)]
    #[require(B)]
    struct A;

    #[derive(Component, Default)]
    struct B;

    #[derive(Component, Default)]
    struct C;

    let mut world = World::new();

}
```

```rust
#[test]
    #[derive(Component)]
    #[require(B)]
    struct A;

    #[derive(Component, Default)]
    struct B;

    #[derive(Component, Default)]
    struct C;

    let mut world = World::new();

    let id = world.spawn(B).id();
}
```

Jondolf:
# Objective

Fixes #16406.

Currently, the `#[require(...)]` attribute internally registers component requirements using `register_required_components_manual`. This is done recursively in a way where every requirement in the "inheritance tree" is added into a flat `RequiredComponents` hash map with component constructors and inheritance depths stored.

However, this does not consider runtime requirements: if a plugin has already registered `C` as required by `B`, and a component `A` requires `B` through the macro attribute, spawning an entity with `A` won't add `C`. The `required_by` hash set for `C` doesn't have `A`, and the `RequiredComponents` of `A` don't have `C`.

Intuitively, I would've thought that the macro attribute's requirements were always added *before* runtime requirements, and in that case I believe this shouldn't have been an issue. But the macro requirements are based on `Component::register_required_components`, which in a lot of cases (I think) is only called *after* the first time a bundle with the component is inserted. So if a runtime requirement is defined *before* this (as is often the case, during `Plugin::build`), the macro may not take it into account.

## Solution

Register requirements inherited from the `required` component in `register_required_components_manual_unchecked`.

## Testing

I added a test, essentially the same as in #16406, and it now passes. I also ran some of the tests in #16409, and they seem to work as expected. All the existing tests for required components pass.

### DIFF
--- a/crates/bevy_ecs/src/component.rs
+++ b/crates/bevy_ecs/src/component.rs
@@ -1159,15 +1159,14 @@ impl Components {
     // NOTE: This should maybe be private, but it is currently public so that `bevy_ecs_macros` can use it.
     //       We can't directly move this there either, because this uses `Components::get_required_by_mut`,
     //       which is private, and could be equally risky to expose to users.
-    /// Registers the given component `R` as a [required component] for `T`,
-    /// and adds `T` to the list of requirees for `R`.
+    /// Registers the given component `R` and [required components] inherited from it as required by `T`,
+    /// and adds `T` to their lists of requirees.
     ///
     /// The given `inheritance_depth` determines how many levels of inheritance deep the requirement is.
     /// A direct requirement has a depth of `0`, and each level of inheritance increases the depth by `1`.
     /// Lower depths are more specific requirements, and can override existing less specific registrations.
     ///
-    /// This method does *not* recursively register required components for components required by `R`,
-    /// nor does it register them for components that require `T`.
+    /// This method does *not* register any components as required by components that require `T`.
     ///
     /// Only use this method if you know what you are doing. In most cases, you should instead use [`World::register_required_components`],
     /// or the equivalent method in `bevy_app::App`.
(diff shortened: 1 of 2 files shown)


crates/bevy_ecs/src/component.rs
```diff
@@ -1232,6 +1230,27 @@ impl Components {
...
+        for (id, component) in required {
+            // Register the inherited required components for the requiree.
+            // The inheritance depth is increased by `1` since this is a component required by the original required component.
+            required_components.register_dynamic(
+                id,
+                component.constructor.clone(),
+                component.inheritance_depth + 1,
+            );
+            self.get_required_by_mut(id).unwrap().insert(requiree);
+        }
```
Jondolf, line 1253: I tried to make this use `register_inherited_required_components` since the logic is pretty much duplicated, but this needs to use the `&mut RequiredComponents`, and the existing method can't take it as an argument nicely because of borrowing issues

BenjaminBrienen (approved):

mockersf (approved):

crates/bevy_ecs/src/component.rs
alice-i-cecile: I've opened an issue for this :)

PR #16411, opened 2024-11-17
`Cargo.toml` dependency consistency

BenjaminBrienen:
# Objective

Consistency in the way dependencies are declared

## Solution

Update all dependencies to the latest version using Dependi, remove the patch versions for the non `0.x` crates.
Lower the versions for hashbrown, taffy, and smol_str because they require migrations.

This particular style was chosen because only specifying the major version, while it works in practice, would sometimes require `cargo update` to be run, lest it try to compile with an out-of-date version.

## Testing

`cargo check`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,7 +10,6 @@ keywords = ["game", "engine", "gamedev", "graphics", "bevy"]
 license = "MIT OR Apache-2.0"
 repository = "https://github.com/bevyengine/bevy"
 documentation = "https://docs.rs/bevy"
-rust-version = "1.82.0"
 
 [workspace]
 exclude = [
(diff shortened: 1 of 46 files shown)


alice-i-cecile:
There's a lot of churn here, and no automated way to enforce this. Can you say more about the benefits?

BenjaminBrienen:
I guess it's just OCD?

- `serde = "1"` and `serde = "1.0"` in the same workspace
- `rand = "0.8.0"`... but it doesn't use `0.8.0`; it uses `0.8.5`

It just seems inherently bothersome. idk

BenjaminBrienen:
I think one problem with this is that `cargo add` adds the patch version as well, so that's conflicting with this.

BenjaminBrienen:
Feel free to bash on this. This isn't productive and just serves to satisfy my desire for things to be "just right".

mockersf:
this is not the standard in the Rust community, it goes against letting cargo handles version as it was made to do, it overly restricts versions, and it's unenforceable.

globally very against this

PR #16412, opened 2024-11-17
bevy_picking UI example

bas-ie:
# Objective

This example demonstrates responding to picking events within the UI context.

https://github.com/user-attachments/assets/5cca1a82-9646-4e4b-8186-8c3572caebf7

## Solution

It's been fun playing with this, but I'm also open to changing it dramatically if it doesn't quite meet our needs. I was trying to think of a way to demonstrate without making the example too complex. With one-shot systems, I may not have succeeded there... but they do reduce the overall masses of nested UI code, which helps with readability.

I'm also aware other examples partially cover this territory, like `sprite_picking`.

I did have some hover highlighting in there, but took it out to reduce length/complexity.


richchurcher:
Heh. It kinda broke with some of the latest changes... :eyes: :construction:

examples/picking/ui_picking.rs
```diff
@@ -0,0 +1,248 @@
...
+        commands.entity(slot).with_children(|parent| {
+            parent
+                .spawn((
+                    Draggable,
+                    // This node serves solely to be dragged! It provides the transform, using required
+                    // components.
+                    Node::default(),
+                    ImageNode::new(asset_server.load(paths[i])),
+                    // TODO: seems clumsy, better solution?
+                    ZIndex(99),
```
richchurcher, line 122: Note to self: `GlobalZIndex` is a thing.

examples/picking/ui_picking.rs
```diff
@@ -0,0 +1,248 @@
...
+        // Record which slot we started from, in case we need to drop the item back again.
+        drag_details.drag_origin = Some(parent.get());
+
+        // Absolutely-positioned items can be made to follow the cursor.
+        item_node.position_type = PositionType::Absolute;
+
+        // TODO: is there a better solution for the scaling problem? Without this, and without a
+        // parent element, the asset will be un-scaled (and appear tiny). Copying its current
+        // scaling doesn't seem to be helpful.
+        transform.scale *= 4.;
```
richchurcher, line 152: Maybe `ComputedNode::size()`? Something to investigate.

Cargo.toml
```diff
@@ -3788,7 +3788,17 @@ name = "Sprite Picking"
...
-required-features = ["bevy_sprite_picking_backend"]
+
+[[example]]
+name = "ui_picking"
+path = "examples/picking/ui_picking.rs"
+doc-scrape-examples = true
+
+[package.metadata.example.ui_picking]
+name = "UI Picking"
+description = "Demonstrates picking in a UI context"
```
alice-i-cecile, line 3799:
```suggestion
description = "Demonstrates using picking to build a drag-and-drop UI"
```

alice-i-cecile (changes requested):
Generally like this, but I don't understand why we're returning `impl Fn` closures here. They're much more complex to reason about, and I don't think they're needed.

ChristopherBiscardi:
no answer to previous change requests, plus with bsn it definitely doesn't make sense to run_system_cached here.

PR #16415, opened 2024-11-17
Add missing exports in bevy_ecs

ItsDoot:
# Objective

Seemed to have missed the export of `DynamicComponentFetch` from #15593. `TryFromFilteredError` which is returned by `impl TryFrom<FiliteredEntityMut/Ref> for EntityRef/Mut` also seemed to have been missing.

## Solution

Export both of them.


mockersf (approved):

mockersf:
you made them public so they should be documented, after that it's good to go 🙂

mockersf (approved):

PR #16419, opened 2024-11-17
remove gcd impl from bevy_render

atlv24:
# Objective

- bevy_render (poorly) implements gcd (which should be in bevy_math but theres not enough justification to have it there either anyways cus its just one usage)

## Solution

- hardcoded LUT replacement for the one usage

## Testing

- verified the alternative implementation of 4/gcd(4,x) agreed with original for 0..200

### DIFF
--- a/crates/bevy_render/src/mesh/allocator.rs
+++ b/crates/bevy_render/src/mesh/allocator.rs
@@ -955,12 +955,18 @@ impl ElementLayout {
     /// Creates an [`ElementLayout`] for mesh data of the given class (vertex or
     /// index) with the given byte size.
     fn new(class: ElementClass, size: u64) -> ElementLayout {
+        const {
+            assert!(4 == COPY_BUFFER_ALIGNMENT);
+        }
+        // this is equivalent to `4 / gcd(4,size)` but lets us not implement gcd.
+        // ping @atlv if above assert ever fails (likely never)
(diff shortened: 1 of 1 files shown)


alice-i-cecile (approved):

IQuick143 (approved):

BenjaminBrienen:
Shouldn't this have a unit test for at least a few cases?

BenjaminBrienen (approved):

BenjaminBrienen:
`assert_eq!` could've been used, but it doesn't matter at all for this.

atlv24:
`assert_eq!` cant be used, it doesnt work in `const {}` context

atlv24:
i can add tests, most of the renderer doesnt have any

atlv24:
actually adding tests would require implementing gcd again, and if i hardcode values then the tests would literally be testing to check if an array equals itself. not much point here i think

atlv24:
just leaving this here in case anyone wants to slap it into rust playground just to satisfy the itch
```rs
fn main() {
    for i in 0..200 {
        let a = (4 / gcd(i, 4)) as u32;
        let b = [1, 4, 2, 4][i as usize & 3];
        assert_eq!(a, b);
    }
    println!("done");
}

fn gcd(mut a: u64, mut b: u64) -> u64 {
    while b != 0 {
        let t = b;
        b = a % b;
        a = t;
    }
    a
}
```

PR #16421, opened 2024-11-18
Add `gcd` module to `bevy_math`

bushrat011899:
# Objective

- Add a [`gcd`](https://en.wikipedia.org/wiki/Greatest_common_divisor) module to `bevy_math`
- Generalize and document the `4 / gcd(4,size)` algorithm developed by @atlv24 
- Improves upon #16419

## Solution

- Added `gcd`, `gcd_by_table`, and `n_over_gcd_by_table` to `bevy_math`.
- Updated `ElementLayout::new` in `bevy_render` to use `n_over_gcd_by_table`

## Testing

- Added unit tests to confirm the new methods provide accurate results for small cases.
- CI

## Notes

The algorithm provided by @atlv24, while performant, has a bounds check that cannot be removed by the compiler due to the lookup table being a slice. To remove this redundant bounds check, and better document how the algorithm works, I've moved it into `bevy_math` with some documentation, and generalised the technique for any number `N`, not just 4.

For powers of 2, the algorithm is heavily optimised by the rust compiler, entirely omitting divisions and branches. For all other `N`, the algorithm performs a single `mod` (`%`) and a lookup, while still being branchless. For a comparison of this method, the original from @atlv24, and the naive, see [GodBolt](https://godbolt.org/z/zx78hEP3v).

As a summary here, the case where `N` is 4 produces the following assembly (on x86 for demonstration purposes, other platforms will have similar output):

```asm
n_over_gcd_by_table_four:
        and     rdi, 3
        lea     rax, [rip + .L__unnamed_1]
        mov     rax, qword ptr [rax + 8*rdi]
        mov     qword ptr [rsp - 8], rax
        mov     rax, qword ptr [rsp - 8]
        ret
```

Compared to @atlv24's original:

```asm
n_over_gcd_by_table_four_atlv24:
        sub     rsp, 40
        mov     qword ptr [rsp + 8], 1
        mov     qword ptr [rsp + 16], 4
        mov     qword ptr [rsp + 24], 2
        mov     qword ptr [rsp + 32], 4
        and     rdi, 3
        mov     qword ptr [rsp], rdi
        cmp     rdi, 4
        jae     .LBB0_2
        mov     rax, qword ptr [rsp]
        mov     rax, qword ptr [rsp + 8*rax + 8]
        add     rsp, 40
        ret
.LBB0_2:
        mov     rdi, qword ptr [rsp]
        lea     rdx, [rip + .L__unnamed_1]
        mov     rax, qword ptr [rip + core::panicking::panic_bounds_check::hab02a8df06d3a143@GOTPCREL]
        mov     esi, 4
        call    rax
```

Note that while a more efficient algorithm is nice, I'm more concerned with having a generalised and documented implementation. That it produces cleaner assembly is purely a beneficial side effect.


BenjaminBrienen (approved):

crates/bevy_math/src/gcd.rs
```diff
@@ -0,0 +1,158 @@
...
+
+    table[value % N]
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn gcd_tests() {
```
IQuick143, line 94: Due to how the implementation works, I wouldn't take for granted that `gcd(a,b) = gcd(b,a)`, so an extra test with the reversed order of arguments would be nice imo. (at least add one for the a = 0, b != 0 case)

IQuick143 (approved):

crates/bevy_math/src/gcd.rs
bushrat011899: That is a fair concern. I've added a test for symmetry, refactored the implementation of `gcd` to be a little cleaner, and cited where the specific implementation comes from, to hopefully alleviate further concerns around validity.

crates/bevy_render/src/mesh/allocator.rs
```diff
@@ -955,12 +956,8 @@ impl ElementLayout {
...
     /// index) with the given byte size.
     fn new(class: ElementClass, size: u64) -> ElementLayout {
-        const {
-            assert!(4 == COPY_BUFFER_ALIGNMENT);
-        }
-        // this is equivalent to `4 / gcd(4,size)` but lets us not implement gcd.
-        // ping @atlv if above assert ever fails (likely never)
-        let elements_per_slot = [1, 4, 2, 4][size as usize & 3];
+        let elements_per_slot =
+            n_over_gcd_by_table::<{ COPY_BUFFER_ALIGNMENT as usize }>(size as usize) as u32;
```
IQuick143, line 960: Now that I'm looking at it, is there a possibility of a bug caused by usize being smaller than u64?
Can the input values reasonably be more than u32? On a 32-bit platform?

bushrat011899: I've refactored the `gcd` implementations to work on `u64` explicitly, which moves the casts there. By casting from `usize` to `u64`, we avoid size issues on all platforms less than 128-bit.

BD103 (changes requested):
While the code quality itself looks awesome, I disagree with the use of `#[inline(always)]`. In my mind that is reserved for specific optimizations, and should not be applied to functions as broad as these. (Forcing inline could create several copies of the GCD table, for instance.)

Could you please switch to plain `#[inline]`, which gives the compiler room to disagree in the case it isn't worth inlining?

atlv24:
the instruction count comparison on godbolt isnt fair or accurate because you didnt pass optimization level 3 with `-C opt-level=3`.

after doing so, the counts are comparable, the only difference being yours ends up as an actual lookup table:
```asm
four_over_gcd_4:
        and     edi, 3
        lea     rax, [rip + .L__unnamed_1]
        mov     rax, qword ptr [rax + 8*rdi]
        ret

.L__unnamed_1:
        .asciz  "\001\000\000\000\000\000\000\000\004\000\000\000\000\000\000\000\002\000\000\000\000\000\000\000\004\000\000\000\000\000\000"
```
while mine becomes
```asm
four_over_gcd_4:
        mov     qword ptr [rsp - 32], 1
        mov     qword ptr [rsp - 24], 4
        mov     qword ptr [rsp - 16], 2
        mov     qword ptr [rsp - 8], 4
        and     edi, 3
        mov     rax, qword ptr [rsp + 8*rdi - 32]
        ret
```

perf was not the motivation of my change anyhow, this isnt in the hot path afaik. i mostly wanted to take shit out of bevy_render to improve compile time

atlv24:
if we wanna get more arcane and golf instructions theres this buffoonery too:
```rs
fn four_over_gcd_4(i: usize) -> usize {
    (0x4241 >> (i & 3) * 4) & 7
}
```

atlv24:
slight improvement
```rs
fn four_over_gcd_4(i: usize) -> usize {
    (0x04020401 >> (i & 3) * 16) & 7
}
```
```asm
four_over_gcd_4:
        mov     rcx, rdi
        shl     ecx, 4
        mov     eax, 67240961
        shr     rax, cl
        movzx   eax, al
        ret
```

atlv24:
u8 shaves one more instruction
```rs
fn four_over_gcd_4(i: u8) -> u8 {
    (0x04020401 >> i * 16) as u8
}
```

```asm
four_over_gcd_4:
        mov     ecx, edi
        shl     cl, 4
        mov     eax, 67240961
        shr     eax, cl
        ret
```

atlv24:
its not getting more arcane than that tonight. it exploits the fact that shr wraps at 32 to hide the &3 by multiplying by 8 instead because 8*4 = 32, and then also hides the final mask by returning u8. i think it cant be less instructions

crates/bevy_math/src/gcd.rs
```diff
@@ -0,0 +1,168 @@
+//! Provides implementations for methods relating to the [Greatest Common Divisor](https://en.wikipedia.org/wiki/Greatest_common_divisor).
+
+/// Computes the [Greatest Common Divisor](https://en.wikipedia.org/wiki/Greatest_common_divisor)
+/// between the supplied `value` and the constant parameter `N`.
```
mdickopp, line 4: Since this function has parameters `a` and `b` instead of `value` and a constant `N`, the comment should be changed accordingly.

mdickopp (changes requested):

bushrat011899:
Having thought about this more, I'm not really convinced this PR is worth it. I'm going to close this because I'm personally not motivated enough to be an effective advocate. Perfectly happy for someone else to take it though!

PR #16422, opened 2024-11-18
Reactive components prototype

JMS55:
# Objective
- Initial prototype for reactivity in Bevy

## Solution
- A new `ReactiveComponent` type that automatically inserts and updates a component of type `Output` on the same entity whenever the source entity's component of type `Input` changes (as detected by change detection ticks)

## Implementation issues
* There's a single place where reactive components get updated (the Last schedule). Ideally it would be an immediate reaction like how observers work.
* The `update_reactive_components` driver system loops over every single expression any time an expression changes. It would be ideal if we could track affected components, and only re-run expressions that would be affected. But for the initial prototype, performance isn't too important anyways.

## Not implemented/tested
* ReactiveComponents that can add/remove the component, instead of just modifying the value.
* Chaining reactions
* Expressions that depend on multiple components from an entity
* Expressions that depend on multiple entities
* Expressions that depend on resources
* Expressions that depend on queries or other complicated sources of data
* Reactive entity tree structures (e.g. a reactive version of WithChildren/WithChild, using the Name component as a key)

### DIFF
--- a/crates/bevy_app/src/app.rs
+++ b/crates/bevy_app/src/app.rs
@@ -1,6 +1,6 @@
 use crate::{
-    First, Main, MainSchedulePlugin, PlaceholderPlugin, Plugin, Plugins, PluginsState, SubApp,
-    SubApps,
+    First, Last, Main, MainSchedulePlugin, PlaceholderPlugin, Plugin, Plugins, PluginsState,
+    SubApp, SubApps,
 };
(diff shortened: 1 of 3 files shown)


JMS55:
I think template + diffing approaches are the better move here. Change detection is too difficult to get working well I feel.

PR #16425, opened 2024-11-18
Revert most of #16222 and add gamepad accessors

cart:
# Objective

#16222 regressed the user experience of actually using gamepads:

```rust
// Before 16222
gamepad.just_pressed(GamepadButton::South)

// After 16222
gamepad.digital.just_pressed(GamepadButton::South)

// Before 16222
gamepad.get(GamepadButton::RightTrigger2)

// After 16222
gamepad.analog.get(GamepadButton::RighTrigger2)
```

Users shouldn't need to think about "digital vs analog" when checking if a button is pressed. This abstraction was intentional and I strongly believe it is in our users' best interest. Buttons and Axes are _both_ digital and analog, and this is largely an implementation detail.  I don't think reverting this will be controversial.

## Solution

- Revert most of #16222
- Add the `Into<T>` from #16222 to the internals
- Expose read/write `digital` and `analog` accessors on gamepad, in the interest of enabling the mocking scenarios covered in #16222 (and allowing the minority of users that care about the "digital" vs "analog" distinction in this context to make that distinction)

### DIFF
--- a/crates/bevy_input/src/gamepad.rs
+++ b/crates/bevy_input/src/gamepad.rs
@@ -326,7 +326,7 @@ pub enum ButtonSettingsError {
 ///         println!("{name}");
 ///
 ///         if gamepad.just_pressed(GamepadButton::North) {
-///             println!("{} just pressed North", gamepad.name())
+///             println!("{} just pressed North", name)
 ///         }
 ///
 ///         if let Some(left_stick_x) = gamepad.get(GamepadAxis::LeftStickX)  {
--- a/crates/bevy_input/src/gamepad.rs
+++ b/crates/bevy_input/src/gamepad.rs
@@ -307,12 +307,13 @@ pub enum ButtonSettingsError {
     },
 }
 
-/// Stores a connected gamepad's state and any metadata such as the device name.
+/// The [`Gamepad`] [`component`](Component) stores a connected gamepad's metadata such as the `name` and its [`GamepadButton`] and [`GamepadAxis`].
 ///
-/// An entity with this component is spawned automatically after [`GamepadConnectionEvent`]
-/// and updated by [`gamepad_event_processing_system`].
+/// The [`entity`](Entity) representing a gamepad and its [`minimal components`](GamepadSettings) are automatically managed.
 ///
-/// See also [`GamepadSettings`] for configuration.
+/// # Usage
+///
+/// The only way to obtain a [`Gamepad`] is by [`query`](Query).
 ///
 /// # Examples
 ///
@@ -325,11 +326,11 @@ pub enum ButtonSettingsError {
 ///     for (name, gamepad) in &gamepads {
 ///         println!("{name}");
 ///
-///         if gamepad.digital.just_pressed(GamepadButton::North) {
-///             println!("{name} just pressed North")
+///         if gamepad.just_pressed(GamepadButton::North) {
+///             println!("{} just pressed North", gamepad.name())
 ///         }
 ///
-///         if let Some(left_stick_x) = gamepad.analog.get(GamepadAxis::LeftStickX)  {
+///         if let Some(left_stick_x) = gamepad.get(GamepadAxis::LeftStickX)  {
 ///             println!("left stick X: {}", left_stick_x)
 ///         }
 ///     }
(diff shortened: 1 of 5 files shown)


examples/tools/gamepad_viewer.rs
```diff
@@ -393,12 +393,12 @@ fn update_buttons(
     materials: Res<ButtonMaterials>,
     mut query: Query<(&mut MeshMaterial2d<ColorMaterial>, &ReactTo)>,
 ) {
-    for gamepad in &gamepads {
+    for buttons in &gamepads {
```
alice-i-cecile, line 396: I think that `gamepad` is much clearer here.

alice-i-cecile (approved):
I like the end user UX much better here. I'm glad to see accessors exposed for analog / digital, and you can still construct a `Gamepad` externally and it's fully reflect still so I think this should satisfy the needs of power users and input mocking.

crates/bevy_input/src/gamepad.rs
```diff
@@ -340,46 +341,187 @@ pub enum ButtonSettingsError {
...
+
+    /// Returns `true` if the [`GamepadButton`] has been pressed during the current frame.
+    ///
+    /// Note: This function does not imply information regarding the current state of [`ButtonInput::pressed`] or [`ButtonInput::just_released`].
+    pub fn just_pressed(&self, button_type: GamepadButton) -> bool {
+        self.digital.just_pressed(button_type)
+    }
+
+    /// Returns `true` if any item in [`GamepadButton`] has been pressed during the current frame.
+    pub fn any_just_pressed(&self, button_inputs: impl IntoIterator<Item = GamepadButton>) -> bool {
```
Shatur, line 438: `ButtonInput` have `any_just_pressed`, maybe just call it?

crates/bevy_input/src/gamepad.rs
```diff
@@ -340,46 +341,187 @@ pub enum ButtonSettingsError {
...
+        button_inputs
+            .into_iter()
+            .any(|button_type| self.just_pressed(button_type))
+    }
+
+    /// Returns `true` if all items in [`GamepadButton`] have been just pressed.
+    pub fn all_just_pressed(&self, button_inputs: impl IntoIterator<Item = GamepadButton>) -> bool {
```
Shatur, line 445: Maybe add this method to `ButtonInput` as well and pass it here?

crates/bevy_input/src/gamepad.rs
```diff
@@ -340,46 +341,187 @@ pub enum ButtonSettingsError {
...
+
+    /// Returns `true` if the [`GamepadButton`] has been released during the current frame.
+    ///
+    /// Note: This function does not imply information regarding the current state of [`ButtonInput::pressed`] or [`ButtonInput::just_pressed`].
+    pub fn just_released(&self, button_type: GamepadButton) -> bool {
+        self.digital.just_released(button_type)
+    }
+
+    /// Returns `true` if any item in [`GamepadButton`] has just been released.
+    pub fn any_just_released(
```
Shatur, line 459: Same as with `any_just_pressed`.

crates/bevy_input/src/gamepad.rs
```diff
@@ -340,46 +341,187 @@ pub enum ButtonSettingsError {
...
+        &self,
+        button_inputs: impl IntoIterator<Item = GamepadButton>,
+    ) -> bool {
+        button_inputs
+            .into_iter()
+            .any(|button_type| self.just_released(button_type))
+    }
+
+    /// Returns `true` if all items in [`GamepadButton`] have just been released.
+    pub fn all_just_released(
```
Shatur, line 469: Same as with `all_just_pressed`.

crates/bevy_input/src/gamepad.rs
```diff
@@ -307,12 +307,13 @@ pub enum ButtonSettingsError {
     },
 }
 
-/// Stores a connected gamepad's state and any metadata such as the device name.
+/// The [`Gamepad`] [`component`](Component) stores a connected gamepad's metadata such as the `name` and its [`GamepadButton`] and [`GamepadAxis`].
```
Shatur, line 310: It's not very conventional to put a doc link into self. Maybe write it like this?
```suggestion
/// Stores a connected gamepad's metadata such as the name and its [`GamepadButton`] and [`GamepadAxis`].
```

crates/bevy_input/src/gamepad.rs
```diff
@@ -307,12 +307,13 @@ pub enum ButtonSettingsError {
...
 ///
-/// An entity with this component is spawned automatically after [`GamepadConnectionEvent`]
-/// and updated by [`gamepad_event_processing_system`].
+/// The [`entity`](Entity) representing a gamepad and its [`minimal components`](GamepadSettings) are automatically managed.
```
Shatur, line 312:
```suggestion
/// An entity with this component is spawned automatically after [`GamepadConnectionEvent`]
/// and updated by [`gamepad_event_processing_system`].
```

crates/bevy_input/src/gamepad.rs
```diff
@@ -307,12 +307,13 @@ pub enum ButtonSettingsError {
...
 ///
-/// See also [`GamepadSettings`] for configuration.
+/// # Usage
+///
+/// The only way to obtain a [`Gamepad`] is by [`query`](Query).
```
Shatur, line 316: It's obvious that components obtained via query, so maybe keep the old doc?

```suggestion
/// See also [`GamepadSettings`] for configuration.
```

Shatur (changes requested):
Agree, I like this approach more!

Maybe keep some of the doc changes from my PR? See the suggested changes.
Also a few methods could be delegated. I noticed it back then, but didn't fix since they were removed.

examples/tools/gamepad_viewer.rs
cart: Agreed!

Shatur (approved):

PR #16427, opened 2024-11-18
Use multidraw for opaque meshes when GPU culling is in use.

pcwalton:
This commit adds support for *multidraw*, which is a feature that allows multiple meshes to be drawn in a single drawcall. `wgpu` currently implements multidraw on Vulkan, so this feature is only enabled there. Multiple meshes can be drawn at once if they're in the same vertex and index buffers and are otherwise placed in the same bin. (Thus, for example, at present the materials and textures must be identical, but see #16368.) Multidraw is a significant performance improvement during the draw phase because it reduces the number of rebindings, as well as the amount of driver overhead.

This feature is currently only enabled when GPU culling is used: i.e. when `GpuCulling` is present on a camera. Therefore, if you run for example `scene_viewer`, you will not see any performance improvements, because `scene_viewer` doesn't add the `GpuCulling` component to its camera.

Additionally, the multidraw feature is only implemented for opaque 3D meshes and not for shadows or 2D meshes. I plan to make GPU culling the default and to extend the feature to shadows in the future. Also, in the future I suspect that polyfilling multidraw on APIs that don't support it will be fruitful, as even without driver-level support use of multidraw allows us to avoid expensive `wgpu` rebindings.

### DIFF
--- a/crates/bevy_render/src/render_phase/mod.rs
+++ b/crates/bevy_render/src/render_phase/mod.rs
@@ -37,6 +37,7 @@ use encase::{internal::WriteInto, ShaderSize};
 use nonmax::NonMaxU32;
 pub use rangefinder::*;
 
+use crate::batching::gpu_preprocessing::GpuPreprocessingMode;
 use crate::sync_world::MainEntity;
 use crate::{
     batching::{
@@ -53,14 +54,7 @@ use bevy_ecs::{
     prelude::*,
     system::{lifetimeless::SRes, SystemParamItem},
 };
-use core::{
-    fmt::{self, Debug, Formatter},
-    hash::Hash,
-    iter,
-    marker::PhantomData,
-    ops::Range,
-    slice::SliceIndex,
-};
+use core::{fmt::Debug, hash::Hash, iter, marker::PhantomData, ops::Range, slice::SliceIndex};
 use smallvec::SmallVec;
 
 /// Stores the rendering instructions for a single phase that uses bins in all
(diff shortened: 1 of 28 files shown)


pcwalton:
I'm marking this PR as a draft until 0.15 is out, to indicate that we shouldn't merge it before then. It should generally be ready to go, though; feel free to review.

JMS55 (approved):
Can't really test performance because I lack a desktop atm, but code lgtm.

crates/bevy_render/src/render_phase/mod.rs
```diff
@@ -891,114 +1007,40 @@ pub trait PhaseItem: Sized + Send + Sync + 'static {
...
 
-    /// If this extra index describes an indirect parameters index, returns it;
-    /// otherwise, returns `None`.
-    #[inline]
-    pub fn as_indirect_parameters_index(&self) -> Option<u32> {
-        if self.is_indirect_parameters_index() {
-            Some(self.0 & Self::OFFSET_MASK)
-        } else {
-            None
+    /// Returns either a dynamic offset index or [`PhaseItemExtraIndex::NONE`],
```
IceSentry, line 1038:
```suggestion
    /// Returns either a dynamic offset index or [`PhaseItemExtraIndex::None`],
```

IceSentry (approved):
I haven't tested the code locally yet, but everything LGTM

pcwalton:
Merge conflicts addressed.

pcwalton:
Conflicts addressed.

alice-i-cecile:
Thank you to everyone involved with the authoring or reviewing of this PR! This work is relatively important and needs release notes! Head over to https://github.com/bevyengine/bevy-website/issues/1968 if you'd like to help out.

PR #16436, opened 2024-11-19
Fix runtime required components not registering correctly

### LINKED ISSUE #16406: Runtime required components do not work correctly for components added through `#[require(...)]`

andriyDev:
# Objective

- Fixes #16406 
- Fixes an issue where registering a "deeper" required component, then a "shallower" required component, would result in the wrong required constructor being used for the root component.

## Solution

- Make `register_required_components` add any "parent" of a component as `required_by` to the new "child".
- Assign the depth of the `requiree` plus 1 as the depth of a new runtime required component.

## Testing

- Added two new tests.

### DIFF
--- a/crates/bevy_ecs/src/component.rs
+++ b/crates/bevy_ecs/src/component.rs
@@ -1076,10 +1076,9 @@ impl Components {
                         .debug_checked_unwrap()
                 };
 
+                // Register the original required component in the "parent" of the requiree.
+                // The inheritance depth is 1 deeper than the `requiree` wrt `required_by_id`.
                 let depth = required_components.0.get(&requiree).expect("requiree is required by required_by_id, so its required_components must include requiree").inheritance_depth;
-
-                // Register the original required component for the requiree.
-                // The inheritance depth is `1` since this is a component required by the original requiree.
                 required_components.register_by_id(required, constructor, depth + 1);
 
                 for (component_id, component) in inherited_requirements.iter() {


crates/bevy_ecs/src/component.rs
```diff
@@ -1065,16 +1065,22 @@ impl Components {
...
                 // SAFETY: The component is in the list of required components, so it must exist already.
                 let required_components = unsafe {
                     self.get_required_components_mut(required_by_id)
                         .debug_checked_unwrap()
                 };
 
+                let depth = required_components.0.get(&requiree).expect("requiree is required by required_by_id, so its required_components must include requiree").inheritance_depth;
+
                 // Register the original required component for the requiree.
                 // The inheritance depth is `1` since this is a component required by the original requiree.
```
Jondolf, line 1082: This comment is now outdated

andriyDev: Fixed.

Jondolf (approved):

alice-i-cecile (approved):

PR #16439, opened 2024-11-19
feat: mark some functions in bevy_math as const

DaAlbrecht:
# Objective

Mark simpel functions as const in `bevy_math`

## Solution

- Make them const

## Testing

`cargo test -p bevy_math --all-features`


DaAlbrecht:
Sadly most functions I saw can't be marked as const until floating point arithmetic is marked as const in the std. 

I quickly searched in `bevy_color` and `bevy_ui` but this seemed already done.

alice-i-cecile (approved):

PR #16440, opened 2024-11-19
AnimationEvent -> Event and other improvements

cart:
# Objective

Needing to derive `AnimationEvent` for `Event` is unnecessary, and the trigger logic coupled to it feels like we're coupling "event producer" logic with the event itself, which feels wrong. It also comes with a bunch of complexity, which is again unnecessary. We can have the flexibility of "custom animation event trigger logic" without this coupling and complexity.

The current `animation_events` example is also needlessly complicated, due to it needing to work around system ordering issues. The docs describing it are also slightly wrong. We can make this all a non-issue by solving the underlying ordering problem.

Related to this, we use the `bevy_animation::Animation` system set to solve PostUpdate animation order-of-operations issues. If we move this to bevy_app as part of our "core schedule", we can cut out needless `bevy_animation` crate dependencies in these instances.

## Solution

- Remove `AnimationEvent`, the derive,  and all other infrastructure associated with it (such as the `bevy_animation/derive` crate)
- Replace all instances of `AnimationEvent` traits with `Event + Clone`
- Store and use functions for custom animation trigger logic (ex: `clip.add_event_fn()`). For "normal" cases users dont need to think about this and should use the simpler `clip.add_event()`
- Run the `Animation` system set _before_ updating text
- Move `bevy_animation::Animation` to `bevy_app::Animation`. Remove unnecessary `bevy_animation` dependency from `bevy_ui`
- Adjust `animation_events` example to use the simpler `clip.add_event` API, as the workarounds are no longer necessary

This is polishing work that will land in 0.15, and I think it is simple enough and valuable enough to land in 0.15 with it, in the interest of making the feature as compelling as possible.


crates/bevy_app/src/main_schedule.rs
```diff
@@ -181,6 +181,10 @@ pub struct PostUpdate;
 #[derive(ScheduleLabel, Clone, Debug, PartialEq, Eq, Hash)]
 pub struct Last;
 
+/// Animation system set. This exists in [`PostUpdate`].
```
alice-i-cecile, line 184: Fascinating choice. I've been reluctant to clutter this file with this sort of interop / glue, but I think that this is a good way to handle this flavor of indirect ordering depdendency.

crates/bevy_animation/src/lib.rs
```diff
@@ -472,6 +494,21 @@ impl AnimationClip {
             .push(variable_curve);
     }
 
+    /// Add a untargeted [`AnimationEvent`] to this [`AnimationClip`].
```
alice-i-cecile, line 497: The [`AnimationEvent`] type that you're linking to is private, and has no docs. We should probably fix that.

crates/bevy_animation/src/lib.rs
```diff
@@ -472,6 +494,21 @@ impl AnimationClip {
...
+    ///
+    /// The `event` will trigger on the [`AnimationPlayer`] entity once the `time` (in seconds)
```
alice-i-cecile, line 499:
```suggestion
    /// The `event` will be cloned and trigger on the [`AnimationPlayer`] entity once the `time` (in seconds)
```

crates/bevy_animation/src/lib.rs
```diff
@@ -482,26 +519,53 @@ impl AnimationClip {
...
     /// is reached in the animation.
     ///
+    /// For a simpler [`Event`]-based alternative, see [`AnimationClip::add_event`].
     /// See also [`add_event_to_target`](Self::add_event_to_target).
-    pub fn add_event(&mut self, time: f32, event: impl AnimationEvent) {
-        self.add_event_to_target_inner(AnimationEventTarget::Root, time, event);
+    pub fn add_event_fn(
+        &mut self,
+        time: f32,
+        func: impl Fn(&mut Commands, Entity, f32, f32) + Send + Sync + 'static,
```
alice-i-cecile, line 543: This signature is nasty for Rust beginners to read. A doc test would go a long way here.

crates/bevy_app/src/main_schedule.rs
cart: Yeah. "animation" feels like a general enough concept to merit this kind of "concept promotion". I think we should be stingy about what we allow here, but promoting "arbitrary animation happens in this part of PostUpdate" feels reasonable to me, especially given that it makes our compile graph nicer. I think the more opinionated something is, the more we should discourage "promotion".

crates/bevy_animation/src/lib.rs
cart: Ah yeah. It is intentionally private / an implementation detail. Ill switch this to `Event`

alice-i-cecile (changes requested):
I really like how much simpler this. That said, the `AnimationEvent` doc link is both wrong and broken and needs to be fixed. The other comments are just suggestions and can be ignored if you just want to get this shipped.

alice-i-cecile (approved):

BenjaminBrienen:
check-doc is still failing. Other than that, I think this is good.

PR #16450, opened 2024-11-20
remove the detailed_trace macro

mockersf:
# Objective

- Fix CI
- Contribute to #11478 

## Solution

- Remove the `detailed_trace` macro, and replace its use by the `trace` macro
- The macro doesn't work: it doesn't check for the `detailed_trace` feature
- If it checked for the feature, as the macro is expanded where it's used, the feature should be added to all crates
- Log levels are already controllable by features

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -370,9 +370,6 @@ animation = ["bevy_internal/animation", "bevy_animation"]
 # Enable using a shared stdlib for cxx on Android
 android_shared_stdcxx = ["bevy_internal/android_shared_stdcxx"]
 
-# Enable detailed trace event logging. These trace events are expensive even when off, thus they require compile time opt-in
-detailed_trace = ["bevy_internal/detailed_trace"]
-
 # Include tonemapping Look Up Tables KTX2 files. If everything is pink, you need to enable this feature or change the `Tonemapping` method for your `Camera2d` or `Camera3d`.
 tonemapping_luts = ["bevy_internal/tonemapping_luts", "ktx2", "zstd"]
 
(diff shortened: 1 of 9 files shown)


alice-i-cecile (approved):
I agree with removing this in the current broken form, especially since it's now breaking CI.

alice-i-cecile:
@mockersf you're missing some imports, as CI helpfully points out: https://github.com/bevyengine/bevy/actions/runs/11940939339/job/33284703598?pr=16450#step:6:749

cart (approved):

rparrett:
> Log levels are already controllable by features

Just linking those here for visibility: https://docs.rs/tracing/latest/tracing/level_filters/index.html#compile-time-filters

cart (approved):
This does mean people need to remember to disable tracing (via compile time flags) or they will have significant performance degradation (even in release mode). I'm pretty worried that we'll be tanking performance by default now / a high percentage of games will ship with high-traffic / high cost trace logs. That being said, I do like the simplicity of just using normal log levels / removing this plumbing .

cart:
Actually hold on. Can we just fix the macro for 0.15? Maybe that would be better than rushing this in. It would be good to have a full cycle to evaluate the performance implications of this.

mockersf:
In that case, I would prefer to remove those logs completely.

They were not possible to enable through the feature, so no one should miss them

cart:
I think we should discuss that as the permanent solution (ex: evaluate each log for value / do performance cost-benefit / decide the correct long term path for this category of thing). Imo we should minimally fix this for 0.15 without rocking the boat. It feels like removing logs indiscriminately is rocking the boat (even if they don't currently have users).

BenjaminBrienen:
triage: Macro was fixed for 0.15 release. Has merge conflicts. This does still seem relevant.

PR #16452, opened 2024-11-20
Fix detailed_trace!

cart:
Alternative to #16450 

# Objective

detailed_trace! in its current form does not work  (and breaks CI)

## Solution

Fix detailed_trace by checking for the feature properly, adding it to the correct crates, and removing it from the incorrect crates


mockersf:
as mentioned in #16450, I think a better solution is to remove the macro *and* the logs that are using it

cart:
I responded there

cart:
Merging now in the interest of unblocking main. If we decide removing is the move for 0.15, we can do that later

PR #16458, opened 2024-11-21
Inserting Resource Tuples

### LINKED ISSUE #16398: Allow resources to be inserted via tuple
## Whats the problem?

I think it's inconsistent that `System`s and `Plugin`s can be inserted via tuple like this, but not `Resource`s:
```rust
app.add_plugins((
        DefaultPlugins,
        RapierPhysicsPlugin::<NoUserData>::default(),
        bevy_framepace::FramepacePlugin,
));
```
I mean sure, it's just a quality of life change, but it really annoys me haha

## Implementation

If this proposal is green-lit, I want to try implementing it myself. I will probably do something similar to how [`Plugin` tuples are handled](https://github.com/bevyengine/bevy/blob/56ac38120b27f4c297a8355fb458a9e15333329e/crates/bevy_app/src/plugin.rs#L164).
I will also update the [documentation](https://docs.rs/bevy_app/latest/bevy_app/struct.App.html#method.insert_resource) on `App::insert_resource` to reflect the change, but I dont know what kind of unit tests i should write, or what other documentation should change, so I'm happy to listen to suggestions.

Wuketuke:
# Objective

Some structs like `System`s or `Plugin`s can be inserted via tuple, but not `Resource`s, which greatly annoyed me.
Fixed #16398

## Solution

I very closely [copied](https://github.com/Wuketuke/bevy/blob/26aad2cf06f1eeccb0788671142cca6b01184ae8/crates/bevy_ecs/src/system/system_param.rs#L970) what [plugins do](https://github.com/Wuketuke/bevy/blob/26aad2cf06f1eeccb0788671142cca6b01184ae8/crates/bevy_app/src/plugin.rs#L131) in `bevy_app`. I also modified every function signature relating to inserting `Resource`s.
However, I still have some open questions about my implementation:
- I dont exaclty know what `#[track_caller]` does. I tried faithfully keeping that functionality, but its possible that I broke something.
- Somehow, [`World::insert_resource_with_caller`](https://github.com/Wuketuke/bevy/blob/26aad2cf06f1eeccb0788671142cca6b01184ae8/crates/bevy_ecs/src/world/mod.rs#L1764C19-L1764C46) is now unused, and I dont what went wrong
- Some other functions like [`App::init_resource`](https://github.com/Wuketuke/bevy/blob/26aad2cf06f1eeccb0788671142cca6b01184ae8/crates/bevy_ecs/src/world/mod.rs#L1720), which dont currently support tuples. Should I also rework these functions?
- should i rename `insert_resource` to `insert_resource` to reflect this new feature?
- I removed the generic function signature in [a file](https://github.com/Wuketuke/bevy/blob/26aad2cf06f1eeccb0788671142cca6b01184ae8/crates/bevy_state/src/app.rs#L118) on line 118 and 131, which hopefully didnt break anything

## Testing

I added an example in the [documentation](https://github.com/Wuketuke/bevy/blob/26aad2cf06f1eeccb0788671142cca6b01184ae8/crates/bevy_app/src/app.rs#L373), and [two new unit tests](https://github.com/Wuketuke/bevy/blob/26aad2cf06f1eeccb0788671142cca6b01184ae8/crates/bevy_ecs/src/system/system_param.rs#L2784)

### DIFF
--- a/crates/bevy_app/src/app.rs
+++ b/crates/bevy_app/src/app.rs
@@ -362,13 +362,20 @@ impl App {
     /// struct MyCounter {
     ///     counter: usize,
     /// }
+    /// 
+    /// #[derive(Resource)]
+    /// struct GamePaused {
+    ///     paused: bool,
+    /// }
     ///
     /// App::new()
     ///    .insert_resource(MyCounter { counter: 0 });
+    /// App::new()
+    ///    .insert_resource((MyCounter { counter: 0}, GamePaused { paused: false}));
     /// ```
-    pub fn insert_resource<R: Resource>(&mut self, resource: R) -> &mut Self {
+    pub fn insert_resource<R: Resource>(&mut self, resource: impl bevy_ecs::system::Resources<R>) -> &mut Self {
         self.main_mut().insert_resource(resource);
-        self
+        self 
     }
 
     /// Inserts the [`Resource`], initialized with its default value, into the app,
--- a/crates/bevy_app/src/app.rs
+++ b/crates/bevy_app/src/app.rs
@@ -362,7 +362,7 @@ impl App {
     /// struct MyCounter {
     ///     counter: usize,
     /// }
-    /// 
+    ///
     /// #[derive(Resource)]
     /// struct GamePaused {
     ///     paused: bool,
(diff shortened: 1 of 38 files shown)


Wuketuke:
yeah i agree that this isnt needed. I thought it was an easy implementation, but it is more complex than i initially thought (and not really worth since its a small QOL thing)

PR #16460, opened 2024-11-21
Remove flush_and_reserve_invalid_assuming_no_entities

SpecificProtagonist:
# Objective

`flush_and_reserve_invalid_assuming_no_entities` was made for the old rendering world (which was reset every frame) and is usused since the 0.15 retained rendering world, but wasn't removed yet. It is pub, but is undocumented apart from the safety comment.

## Solution

Remove `flush_and_reserve_invalid_assuming_no_entities` and the safety invariants this method required for `EntityMeta`, `EntityLocation`, `TableId` and `TableRow`. This reduces the amount of unsafe code & safety invariants and makes #16047 easier.

## Alternatives
- Document `flush_and_reserve_invalid_assuming_no_entities` and keep it unchanged
- Document `flush_and_reserve_invalid_assuming_no_entities` and change it to be based on `EntityMeta::INVALID`

## Migration Guide
- exchange `Entities::flush_and_reserve_invalid_assuming_no_entities` for `reserve` and `flush_as_invalid` and notify us if that's insufficient

### DIFF
--- a/crates/bevy_ecs/src/entity/mod.rs
+++ b/crates/bevy_ecs/src/entity/mod.rs
@@ -954,7 +954,7 @@ impl EntityMeta {
     };
 }
 
-/// Records where an entities data is stored.
+/// Records where an entity's data is stored.
 #[derive(Copy, Clone, Debug, PartialEq)]
 pub struct EntityLocation {
     /// The ID of the [`Archetype`] the [`Entity`] belongs to.
--- a/crates/bevy_ecs/src/entity/mod.rs
+++ b/crates/bevy_ecs/src/entity/mod.rs
@@ -913,25 +913,6 @@ impl Entities {
         }
     }
 
-    /// # Safety
-    ///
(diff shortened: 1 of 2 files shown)


crates/bevy_ecs/src/entity/mod.rs
```diff
@@ -979,13 +954,8 @@ impl EntityMeta {
     };
 }
 
-// This type is repr(C) to ensure that the layout and values within it can be safe to fully fill
-// with u8::MAX, as required by [`Entities::flush_and_reserve_invalid_assuming_no_entities`].
-// SAFETY:
-// This type must not contain any pointers at any level, and be safe to fully fill with u8::MAX.
-/// A location of an entity in an archetype.
+/// Records where an entities data is stored.
```
BenjaminBrienen, line 957:
```suggestion
/// Records where an entity's data is stored.
```

BenjaminBrienen (changes requested):

BenjaminBrienen (approved):

PR #14449, opened 2024-07-23
Retain Rendering World

re0312:
# Objective

- fix #12144 ,compare to #14252 
- We want Relations! However, the render world currently hinders the development of  `* as entities` because it needs to despawn all entities each frame to ensure consistent entity IDs between the main world and render world.

## Solution
- Transition to a strategy of maintaining entity-to-entity mapping between the main world and render world, abandoning the method of removing all entities every frame.
- Use `Observer `to listen the components that tis entity needs to be synchronized to the render world.
- Before executing the extract schedule, the `entity_sync_system `will be run first, which is responsible for maintaining the entity mapping between the main world and the rendering world

Bevy Schedule 
image

##  Alternative(s)

we could using Resource to manage all rendering entity component data  so that we don't have any entities at all.

## Advantages

- We can continue to use ECS in rendering, which can bring more performance advantages in many aspects(Retain redering world + update in place)
- More consistent primitives, as Bevy is entirely based on ECS.

## Drawbacks 
- Once a component has a dependency relationship on other main world entities, this relationship must also be synchronized to the rendering world, which undoubtedly increases the user's mental burden
- Not all the entities are in sync with the rendering world,  And that makes it pretty confusing to tell apart the entities in the main world from those in the rendering world when we're in the middle of rendering stuff. And considering our extensive use of  `EntityHashmap `to handle rendering data. that's could made everything even more of a headache to deal with.

## Performance
cargo run --release --example many_cubes --features bevy/trace_tracy
red is main
image

cargo run --release --example many_foxes --features bevy/trace_tracy
image

cargo run --release --example many_buttons --features bevy/trace_tracy 

image

~30% loss
The majority of the performance loss from the spawn(~3ms) and despawn(~3ms) operations of ephemeral entity.

## Next Steps

1. Stop spawn temporary entities and depsawn it every frame , it could persist in render world
2. Update in place for instance data instead re-insert it per frame 

## Migration Guide

Render world no longer clear all entities per frame , Instead, You must manually control the spawn and despawn of render entities properly.

Before this pr ,`extract` would copy the related main world entity and its data into render world. Then, the render world clear all render entities at the end of the frame to reserve enough entity space, ensuring that no main world entity ID would be occupied when next `extract`.

For the sake of `* as entities`, **we turn to an entity-to-entity mapping strategy to sync between main world entity and render world entity, where each `synchronized` main entity has a `RenderEntity` component  that holds an Entity ID pointer to its unique counterpart entity in the render world.**

image

[...]

PR #16462, opened 2024-11-21
Use default storage for `TemporaryRenderEntity`

rparrett:
# Objective

`TemporaryRenderEntity` currently uses `SparseSet` storage, but doesn't seem to fit the criteria for a component that would benefit from this.

Typical usage of `TemporaryRenderEntity` (and all current usages of it in engine as far as I can tell) would be to spawn an entity with it once and then iterate over it once to despawn that entity.

`SparseSet` is said to be useful for insert/removal perf at the cost of iteration perf.

## Solution

Use the default table storage

## Testing

Possibly this could show up in stress tests like `many_buttons`. I didn't do any benchmarking.

### DIFF
--- a/crates/bevy_render/src/sync_world.rs
+++ b/crates/bevy_render/src/sync_world.rs
@@ -166,7 +166,6 @@ pub type MainEntityHashSet = hashbrown::HashSet<MainEntity, EntityHash>;
 
 /// Marker component that indicates that its entity needs to be despawned at the end of the frame.
 #[derive(Component, Copy, Clone, Debug, Default, Reflect)]
-#[component(storage = "SparseSet")]
 pub struct TemporaryRenderEntity;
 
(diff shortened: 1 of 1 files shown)


alice-i-cecile (approved):
In the absence of benchmarking we should stick to the default choice. I also think it's the better one here.

BenjaminBrienen (approved):

BenjaminBrienen:
Is there a justification from the original author of this line? I don't have the tools to see which PR added it at this moment.

rparrett:
I couldn't find any discussion about it in either the original or adopted PR that introduced the code.

https://github.com/bevyengine/bevy/pull/14449
https://github.com/bevyengine/bevy/pull/15320

Digging a bit further, it was added in this particular commit https://github.com/bevyengine/bevy/pull/15320/commits/8d2c81a1a887f6786fec944f7b9099e2259351fa but I still don't see any relevant review or conversation.

PR #16464, opened 2024-11-21
Dont enable bevy_remote by default

notmd:
# Objective

- I dont think we want bevy_remote enable by default until our editor is out.

## Solution

- Disable it


mockersf (approved):

notmd:
I have updated the example config, hope everything is good now

Issue #15725, opened 2024-10-08
Transition animations for UI elements


viridia:
## What problem does this solve or what need does it fill?

Most animations in games are based on a timeline with keyframes: that is, you have some parameter which is being animated over a curve. Your options are to start, stop, or reverse the animation, but the control points are usually fixed.

However, a different kind of animation - one often seen in CSS - is an interpolation to a target value, often modulated by some mathematical easing function. (There are also physics-based animations, like mechanical spring simulations). These animations don't have a fixed timeline of keyframes, and in fact the target value can be changed in mid-animation.

These kinds of animations are very useful in UI work for things like popup menus, dialogs, sliding drawers, basically anything that has an "enter" and "exit" transition. "Interpolate to target" animations have an advantage over keyframe-based approaches because the animation is frequently interrupted: the popup menu may be closed before it has finished opening. It also works well for elements that have more than two states, such as a sidebar which might have "hidden", "collapsed" and "expanded" states. The problem with the keyframe approach is that restarting the animation track in mid-animation will cause an unsightly "pop".

A related requirement is that we need some way to poll when the animation is complete. Take for example a dialog or inventory screen: we don't want to keep around the entity hierarchy for the popup when it's not visible. But when the popup closes, we can't despawn the entities right away; we need to wait until the closing animation is complete. Typically this would be done by some conditional expression such as "if open || animation.running" where "running" means that the animation has not reached some quiescent state.

For UI work, it would be ideal to be able to do this by inserting a component into the Node entity, such that this component would continually modify the style and transform properties of the node. 

(In my case, I would want to be able to rely on Bevy component change detection for the polling: in other words, I would like to be able to monitor the progress of the animation by looking for changed components. This would let me easily integrate it with my reactive framework.)

The kinds of properties we most frequently want to animate are:

* scale
* translation
* rotation
* color (background, border, outline or text)
* alpha

## What solution would you like?

I have prototyped a framework for this in bevy_reactor, however this was done over a year ago, and there have been a lot of developments in Bevy since then, and my code doesn't take advantage of any of those improvements. What I would like to see is a solution which integrates all of the various animation ideas that people have had in the last year.

Part of the motivation for this ticket is to start a discussion on what the API might look like.

## What alternative(s) have you considered?

I already have a working solution but it's less than ideal.

viridia:
@mockersf - I think this could dovetail with some of the work you have been doing on interpolation curves.
[...]

PR #16484, opened 2024-11-23
AnimatedProperty and AnimatedPropertyOptional

cart:
# Objective

Animating component properties requires too much boilerplate at the moment:

```rust
#[derive(Reflect)]
struct FontSizeProperty;

impl AnimatableProperty for FontSizeProperty {
    type Component = TextFont;

    type Property = f32;

    fn get_mut(component: &mut Self::Component) -> Option<&mut Self::Property> {
        Some(&mut component.font_size)
    }
}

animation_clip.add_curve_to_target(
    animation_target_id,
    AnimatableKeyframeCurve::new(
        [0.0, 0.5, 1.0, 1.5, 2.0, 2.5, 3.0]
            .into_iter()
            .zip([24.0, 80.0, 24.0, 80.0, 24.0, 80.0, 24.0]),
    )
    .map(AnimatableCurve::<FontSizeProperty, _>::from_curve)
    .expect("should be able to build translation curve because we pass in valid samples"),
);
```

## Solution

This adds `AnimatedProperty` and `AnimatedPropertyOptional`, enabling the following:

```rust
animation_clip.add_curve_to_target(
    animation_target_id,
    AnimatableCurve::new(
        AnimatedProperty::new(|font: &mut TextFont| &mut font.font_size),
        AnimatableKeyframeCurve::new(
            [0.0, 0.5, 1.0, 1.5, 2.0, 2.5, 3.0]
                .into_iter()
                .zip([24.0, 80.0, 24.0, 80.0, 24.0, 80.0, 24.0]),
        )
        .expect(
            "should be able to build translation curve because we pass in valid samples",
        ),
    ),
);
```

This required reworking the internals a bit, namely stripping out a lot of the `Reflect` usage, as that implementation was fundamentally incompatible with the `AnimatedProperty` pattern. `Reflect` was being used in this context just to downcast traits. But we can get downcasting behavior without the `Reflect` requirement by implementing `Downcast` for `AnimationCurveEvaluator`.

### DIFF
--- a/crates/bevy_animation/src/animation_curves.rs
+++ b/crates/bevy_animation/src/animation_curves.rs
@@ -91,7 +91,7 @@ use bevy_math::{
     },
     Quat, Vec3,
 };
-use bevy_reflect::{FromReflect, Reflect, Reflectable, TypePath};
+use bevy_reflect::{FromReflect, Reflect, Reflectable};
 use bevy_render::mesh::morph::MorphWeights;
 use bevy_transform::prelude::Transform;
 
@@ -100,6 +100,7 @@ use crate::{
     prelude::{Animatable, BlendInput},
     AnimationEntityMut, AnimationEvaluationError,
 };
+use downcast_rs::{impl_downcast, Downcast};
 
 /// A value on a component that Bevy can animate.
 ///
@@ -160,17 +161,83 @@ use crate::{
 /// configured above).
 ///
 /// [`AnimationClip`]: crate::AnimationClip
-pub trait AnimatableProperty: Reflect + TypePath {
+pub trait AnimatableProperty {
     /// The type of the component that the property lives on.
     type Component: Component;
 
     /// The type of the property to be animated.
-    type Property: Animatable + FromReflect + Reflectable + Clone + Sync + Debug;
+    type Property: Animatable + Clone + Sync + Debug;
 
     /// Given a reference to the component, returns a reference to the property.
     ///
     /// If the property couldn't be found, returns `None`.
-    fn get_mut(component: &mut Self::Component) -> Option<&mut Self::Property>;
+    fn get_mut<'a>(&self, component: &'a mut Self::Component) -> Option<&'a mut Self::Property>;
+}
+
+/// A [`Component`] property that can be animated, defined by a function that reads the component and returns
+/// the accessed field / property.
+#[derive(Clone)]
+pub struct AnimatedProperty<C, P, F: Fn(&mut C) -> &mut P> {
+    func: F,
+    marker: PhantomData<(C, P)>,
+}
+
+impl<C, P, F> AnimatableProperty for AnimatedProperty<C, P, F>
+where
+    C: Component,
+    P: Animatable + Clone + Sync + Debug,
+    F: Fn(&mut C) -> &mut P,
+{
+    type Component = C;
+
+    type Property = P;
+
+    fn get_mut<'a>(&self, component: &'a mut Self::Component) -> Option<&'a mut Self::Property> {
+        Some((self.func)(component))
+    }
+}
+
+impl<C, P, F: Fn(&mut C) -> &mut P + 'static> AnimatedProperty<C, P, F> {
+    /// Creates a new instance of [`AnimatedProperty`]
+    pub fn new(func: F) -> Self {
+        Self {
+            func,
+            marker: PhantomData,
+        }
+    }
+}
+
+/// A [`Component`] property that can be animated, defined by a function that reads the component and returns
+/// either the accessed field / property, or [`None`] if it does not exist
+#[derive(Clone)]
+pub struct AnimatedPropertyOptional<C, P, F: Fn(&mut C) -> Option<&mut P>> {
+    func: F,
+    marker: PhantomData<(C, P)>,
+}
+
+impl<C, P, F> AnimatableProperty for AnimatedPropertyOptional<C, P, F>
+where
+    C: Component,
+    P: Animatable + Clone + Sync + Debug,
+    F: Fn(&mut C) -> Option<&mut P>,
+{
+    type Component = C;
+
+    type Property = P;
+
+    fn get_mut<'a>(&self, component: &'a mut Self::Component) -> Option<&'a mut Self::Property> {
+        (self.func)(component)
+    }
+}
+
+impl<C, P, F: Fn(&mut C) -> Option<&mut P> + 'static> AnimatedPropertyOptional<C, P, F> {
+    /// Creates a new instance of [`AnimatedPropertyOptional`]
+    pub fn new(func: F) -> Self {
+        Self {
+            func,
+            marker: PhantomData,
+        }
+    }
 }
 
 /// This trait collects the additional requirements on top of [`Curve<T>`] needed for a
@@ -187,12 +254,14 @@ impl<T, C> AnimationCompatibleCurve<T> for C where C: Curve<T> + Debug + Clone +
 #[derive(Reflect, FromReflect)]
 #[reflect(from_reflect = false)]
 pub struct AnimatableCurve<P, C> {
+    /// The property selector, which defines what component to access and how to access
+    /// a property on that component.
+    pub property: P,
+
     /// The inner [curve] whose values are used to animate the property.
     ///
     /// [curve]: Curve
     pub curve: C,
-    #[reflect(ignore)]
-    _phantom: PhantomData<P>,
 }
 
 /// An [`AnimatableCurveEvaluator`] for [`AnimatableProperty`] instances.
@@ -205,8 +274,7 @@ where
     P: AnimatableProperty,
 {
     evaluator: BasicAnimationCurveEvaluator<P::Property>,
-    #[reflect(ignore)]
-    phantom: PhantomData<P>,
+    property: P,
 }
 
 impl<P, C> AnimatableCurve<P, C>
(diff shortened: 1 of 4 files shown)


cart:
I'd like to land this in 0.15 as it:
1. Polishes up a feature being released in 0.15
3. Is seasonably straightforward / risk-free

crates/bevy_animation/src/animation_curves.rs
```diff
@@ -205,8 +274,7 @@ where
     P: AnimatableProperty,
 {
     evaluator: BasicAnimationCurveEvaluator<P::Property>,
-    #[reflect(ignore)]
-    phantom: PhantomData<P>,
+    property: P,
```
mweatherley, line 276: This is basically my only concern with this PR. 

Since evaluators are cached and dispatched based on type IDs, the previous setup with marker structs guaranteed that all `AnimatableCurve`s accessing the same property would map to evaluators with the same type, since they used the same marker type as input. 

Now, since `AnimatedProperty` has a function-type generic `F: Fn(&mut C) -> &mut P`, I think there is a subtle footgun involving the instantiation of evaluators: for example, if you create two `AnimatableCurve`s using `AnimatableProperty`s that are the same, but use closures to specify the property accessor, the two will map to distinct evaluator types, since the types (`P` here) will actually be distinct. As a result, the outputs of those curves will not blend correctly (one will just overwrite the other, I think).

In other words, the previous implementation effectively guaranteed that the property accessors would be type-identical, and that is no longer the case now.

mweatherley (changes requested):

crates/bevy_animation/src/animation_curves.rs
mweatherley: (It might be enough to just warn about this and recommend defining the property in a large lexical scope and then reusing it in calls to `AnimatableCurve::new`; if anything, I'm mostly pointing out that the pattern described in the 'Solution' section of the PR description is subtly an anti-pattern.)

viridia:
There are a few things that I would do differently here.

First, I would define a getter and setter rather than returning a mutable field. This would allow the property being animated to be a simplified abstraction of the underlying property. So for example, I could animate rotation_x as an angle instead of having to deal with a quat - the conversion from angle to quat could be done in the AnimatableProperty impl. Similarly, I could animate scale as a scalar affecting all dimensions instead of separate scale_x, scale_y, scale_z.

See also the discussion in #15725 - there's a difference between "keyframe" curves and "transition" curves (my terms, perhaps there's a better way of describing this), in that the animation target can change in mid-animation (like a button which inflates when hovered, but the mouse quickly enters and leaves, giving no time for the animation to complete). In thee cases, you want to avoid an ugly jump from resetting the animation back to start, and instead start the animation from whatever the current state is.

crates/bevy_animation/src/animation_curves.rs
cart: Hmm yeah nice catch. Current behavior doesn't feel shippable, but I think it points to a more fundamental problem.  Theres an odd coupling right now between evaluators and "property accessor code".

Seems like we should decouple evaluator identity from a statically defined AnimationProperty type, and instead tie it to the id of the component combined with the id of the property/field being accessed. I think we'll want to do that anyway to support things like `ReflectedAnimatableProperty::new(TypeId::of::<Transform>(), "translation")`.

cart: I'll start playing with that.

cart: This would probably also allow making arbitrary transform field animation (via things like `AnimatableProperty`) compatible with a custom `TransformCurve` impl. Rather than using the #15665 approach, which relies on everything using the same custom impls.

cart:
Just pushed some changes:

1. Reworked evaluator identity. Evaluators can now either be a (Component / Field) pair, or a TypeId. This allows properties to reuse evaluators, even if they have different accessor methods. The "contract" here is that for a given (Component / Field) pair, the accessor will return the same value. Fields are identified by their Reflect-ed field index. The (TypeId, usize) is prehashed and cached to optimize for lookup speed.
2. Reworked AnimatableProperty to make it box-able / type-eraseable.
3. Renamed AnimatedProperty to AnimatedField to better encapsulate the meaning of the type
4. Added `animated_field!(Transform::translation`)` macro. This removes the need to manually define the Fn to access the field, allows us to make future get/set internals non-breaking in these cases, and allows us to abstract over field-name-accesses (which require string lookups / unwraps). We still get compile-time protection here because of the generated field-access Fn, which will produce compiler errors if the field does not exist.
5. Removes built-in hard-code TranslationCurve / RotationCurve / ScaleCurve in favor of AnimatableField.

crates/bevy_animation/src/animation_curves.rs
```diff
@@ -131,46 +129,99 @@ use crate::{
...
-    ///
-    /// If the property couldn't be found, returns `None`.
-    fn get_mut(component: &mut Self::Component) -> Option<&mut Self::Property>;
+    /// The [`EvaluatorId`] used to look up the [`AnimationCurveEvaluator`] for this [`AnimatableProperty`].
+    /// For a given animated property, this ID should always be the same to allow things like animation blending to occur.
+    fn evaluator_id(&self) -> EvaluatorId;
+}
+
+/// A [`Component`] field that can be animated, defined by a function that reads the component and returns
+/// the accessed field / property.
```
alice-i-cecile, line 176: Some docs here about the meaning of the C, A and F generics would be helpful.

alice-i-cecile, line 176: These docs should link to the macro.

crates/bevy_animation/src/animation_curves.rs
```diff
@@ -131,46 +129,99 @@ use crate::{
...
+    }
+
+    fn evaluator_id(&self) -> EvaluatorId {
+        EvaluatorId::ComponentField(&self.evaluator_id)
+    }
+}
+
+impl<C: Typed, P, F: Fn(&mut C) -> &mut P + 'static> AnimatedField<C, P, F> {
+    /// Creates a new instance of [`AnimatedField`]. This operates under the assumption that
+    /// `C` is a reflect-able struct with named fields, and that `field_name` is a valid field on that struct.
```
alice-i-cecile, line 209: Missing `Panics` section of the docs.

crates/bevy_animation/src/lib.rs
```diff
@@ -100,182 +98,14 @@ impl VariableCurve {
...
-}
-
 /// A list of [`VariableCurve`]s and the [`AnimationTargetId`]s to which they
 /// apply.
 ///
 /// Because animation clips refer to targets by UUID, they can target any
 /// [`AnimationTarget`] with that ID.
 #[derive(Asset, Reflect, Clone, Debug, Default)]
 pub struct AnimationClip {
+    #[reflect(ignore)]
```
alice-i-cecile, line 108: A comment here about why this is needed would be helpful.

crates/bevy_animation/src/lib.rs
```diff
@@ -878,23 +708,106 @@ pub struct AnimationEvaluationState {
     /// Stores all [`AnimationCurveEvaluator`]s corresponding to properties that
     /// we've seen so far.
     ///
-    /// This is a mapping from the type ID of an animation curve evaluator to
+    /// This is a mapping from the id an animation curve evaluator to
```
alice-i-cecile, line 711:
```suggestion
    /// This is a mapping from the id of an animation curve evaluator to
```

alice-i-cecile (approved):
I like the simplification here, and I think this is a reasonable path forward. DreamerTalin's points above are interesting and reasonable, but definitely out of scope for now!

mbrea-c:
I'm also in favor of the new API. My only concern is that by losing `Reflect` on evaluators we make sampling animation curves outside of the built-in bevy_animation flow will be even harder, from what I can see having to resort to `std::mem::transmute`-ing into copied types or maybe trying to write custom evaluators to override the existing (I'll dig into this more tomorrow to see if I missed something). I documented my own use-case for that in #16395, which I can update to bring in line with this new API; I'd feel much more comfortable about this if we also bring in some changes to open up the sampling of animation curves, be it my linked PR or something else entirely.

That said, there's not many applications for "sampling outside of bevy_animation" flow other than third-party animation crates, and it's not really something that was supported before 0.15 either, so it's completely understandable if we don't block on that (just mentioning here to bring some visibility to the issue).

mweatherley:
> I'm also in favor of the new API. My only concern is that by losing `Reflect` on evaluators we make sampling animation curves outside of the built-in bevy_animation flow will be even harder [...]

I am extremely sympathetic to this, but I also think that improving the story for access to animation curves requires enough design that it's unlikely to make 0.15; I'm also skeptical, I think, that `Reflect` is necessarily the right thing for working dynamically with these long-term in this case.

(From my perspective, the intent with how things work right now was that the level things would be "shareable" was with `Curve<T>` rather than `AnimationCurve` per se.)

mbrea-c:
> I am extremely sympathetic to this, but I also think that improving the story for access to animation curves requires enough design that it's unlikely to make 0.15; 

That's fair, it's probably better to leave it for 0.16

> I'm also skeptical, I think, that `Reflect` is necessarily the right thing for working dynamically with these long-term in this case.

Yeah, I completely agree. I never considered it a long term solution, just an unintended "escape hatch" that could be used as a workaround for the missing animation curve sampling/access APIs

> (From my perspective, the intent with how things work right now was that the level things would be "shareable" was with `Curve<T>` rather than `AnimationCurve` per se.)

Makes sense, the more I think about it the more I agree that it'd be the most "natural" API. Maybe we can continue discussion in #15664 after 0.15?

cart:
> I'm also in favor of the new API. My only concern is that by losing Reflect on evaluators we make sampling animation curves outside of the built-in bevy_animation flow will be even harder, from what I can see having to resort to std::mem::transmute-ing into copied types or maybe trying to write custom evaluators to override the existing (I'll dig into this more tomorrow to see if I missed something). I documented my own use-case for that in https://github.com/bevyengine/bevy/pull/16395, which I can update to bring in line with this new API; I'd feel much more comfortable about this if we also bring in some changes to open up the sampling of animation curves, be it my linked PR or something else entirely.

Yeah I also think this is a worthwhile scenario to cover and agree that Reflect was the wrong way to cover it. We should build actual APIs that enable the scenarios we want.

cart:
For 0.15, I think we should move forward with this and then sort out how to cover these scenarios in 0.16

cart:
(docs need a bit of attention ... working on that now)

PR #7314, opened 2023-01-21
Show basic application integration testing (#2896)

Wcubed:
# Objective

Show an example of how the integration testing of a bevy application can be structured.
I went for basic window interaction and user input for the examples, but there are certainly other areas that would be good to demonstrate, I just haven't worked with the engine enough to point them out. Suggestions are welcome :)

Issue #2896, opened 2021-10-01
Provide a way to test Bevy App


arialpew:
## What problem does this solve or what need does it fill?

Bevy App should be testable in headless mode (no renderer), for Continuous Integration and fast feedback when user logic / system interaction is not what user intendeed.

## What solution would you like?

Idealy, a complete example for testing a Bevy App should be provided out of the box.

Currently, there's a minimal example provided here : https://github.com/bevyengine/bevy/blob/main/crates/bevy_ecs/examples/resources.rs

This doesn't work well because you have no way to test things that happen "later in time". You advance Bevy synchronously with iteration in a for loop. If you sleep the main thread between each iteration, you'll advance in time but the side effect is test will be slower. So this example work well for startup system or system that run each tick, but this doesn't work for system that happen "later" in time.

We need a way to test a Bevy Application and travel time to a future point where Bevy State is updated, so assertion can work, for all case (FixedTimeStep System or not).

1) Provide a way to advance or scale Bevy time, skip frame, for test (custom Scheduler ? Or maybe the current default Scheduler is enough ?).
2) Test should be runnable concurrently with cargo test, some plugins are not compatible (for example, LogPlugin is not usable in context of a test if 2 Bevy App use 2 instance of LogPlugin concurrently).
3) Documentation.

## What alternative(s) have you considered?

Don't bother with testing, stay like this.

Honestly, for small Bevy App you don't need test, but for complex system interaction it's a nice feature.

I'm confident about Bevy behavior and logic, but I don't trust my own code especially when it's inverted control (https://en.wikipedia.org/wiki/Inversion_of_control, the framework is responsible to run my application).

A nice side effect of this feature is if you make Bevy easily testable (out of the box with cargo test), contributors will have a better understanding of Bevy code base/change.

## Additional context

Related : https://github.com/bevyengine/bevy/issues/1057 and https://github.com/bevyengine/bevy/issues/1720 and https://github.com/bevyengine/bevy/issues/2687

bjorn3:
This is already partially supported and used on CI: https://github.com/bevyengine/bevy/blob/5ba2b9adcf5e686273cf024acf1ad8ddfb4f8e18/.github/workflows/ci.yml#L248

[...]

PR #16489, opened 2024-11-23
Simple integration testing (adopted)

richchurcher:
# Objective

This older PR from `Wcubed` seemed well worth saving, adopted from #7314. See also tracking issue #2896 for ongoing discussion of Bevy testability. Thanks `Wcubed`!

## Solution

- Updated for 0.15
- Added the `expected`/`actual` pattern
- Switched to function plugin
- Tweaked a bit of description

## Testing

Green.


Wcubed:
Thanks for adopting the pr 😃. I'd almost forgotten I made this one after two years 😄.

tests/how_to_test_apps.rs
```diff
@@ -0,0 +1,133 @@
+#![allow(missing_docs)]
+/// Demonstrates simple integration testing of Bevy applications.
```
alice-i-cecile, line 2: These should be `//!` comments, since they're at the start of the file.

tests/how_to_test_apps.rs
```diff
@@ -0,0 +1,133 @@
+#![allow(missing_docs)]
```
alice-i-cecile, line 1: I think this is only needed because your doc comments below are broken.

```suggestion
```

alice-i-cecile (commented):
Good call: I do really like this! Straightforward and useful. There's a nit to be fixed about doc comments, but I'm happy to approve this after that.

BenjaminBrienen (approved):
This will be super useful!

tests/how_to_test_apps.rs
richchurcher: Argh, I had it that way and changed it. Have now read the rustdoc book in penance :laughing:

alice-i-cecile (approved):

PR #16491, opened 2024-11-23
Upstream `naga_oil`

tychedelia:
`naga_oil` is a foundational crate for Bevy, but doesn't get enough love from community maintainers. We should include it in the monorepo so that it gets more care and attention. The crate was originally left out of the repo under the "useful to other external users" metric. However, since then, we have included other crates like `bevy_color` in the monorepo under and haven't seen uptake from external users that benefits `naga_oil` maintenence.

On the horizon, [WESL](https://github.com/wgsl-tooling-wg/wesl-spec) promises to solve a lot of problems that we currently rely on `naga_oil` for. However, in the meantime, there are a number of important initiatives that would benefit from improvements made to `naga_oil`, including future initiatives like upstreaming `bevy_hanabi` and improvements to `AsBindGroup` that may require code generation strategies. Additionally, there are number of outstanding bugs ex. https://github.com/bevyengine/naga_oil/issues/54 that are more likely to get fixed if they live in our issue tracker.

## Changes

- History imported so we can have a record all the hard work @robtfm has done here.
- The crate has been renamed to `bevy_naga_oil` so that its version can be reset to match the rest of the bevy crates.
- Internal crates point to `bevy_naga_oil`.


tychedelia:
Realizing that the squash merge will obliterate the history, unfortunately.

crates/bevy_naga_oil/src/compose/mod.rs
```diff
@@ -0,0 +1,1912 @@
...
+            override_functions,
+            module_ir,
+            header_ir,
+            start_offset,
+        };
+
+        Ok(composable_module)
+    }
+
+    // shunt all data owned by a composable into a derived module
```
BenjaminBrienen, line 1206:
```suggestion
    /// Shunts all data owned by a composable into a derived module.
```

crates/bevy_naga_oil/src/compose/mod.rs
```diff
@@ -0,0 +1,1912 @@
...
+                    let span = composable.module_ir.functions.get_span(h_f);
+                    derived.import_function_if_new(f, span);
+                }
+            }
+        }
+
+        derived.clear_shader_source();
+    }
+
+    // add an import (and recursive imports) into a derived module
```
BenjaminBrienen, line 1287:
```suggestion
    /// Adds an import (and recursive imports) into a derived module.
```

crates/bevy_naga_oil/src/compose/mod.rs
```diff
@@ -0,0 +1,1912 @@
...
+            shader_defs,
+            true,
+            true,
+            &preprocessed_source,
+            imports,
+        )
+        .map_err(|err| err.into())
+    }
+
+    // build required ComposableModules for a given set of shader_defs
```
BenjaminBrienen, line 1354:
```suggestion
    // Builds the required [`ComposableModules`] for a given set of shader_defs.
```

BenjaminBrienen (changes requested):

tychedelia:
Thanks for all the copyediting @BenjaminBrienen! It's very much appreciated.

doup:
> Realizing that the squash merge will obliterate the history, unfortunately.

If it's interesting to keep the history then someone with privileges could merge it instead?

crates/bevy_naga_oil/README.md
```diff
@@ -0,0 +1,203 @@
+# Bevy Naga Oil
+
+Bevy Naga Organized Integration Library (`bevy_naga-oil`) is a crate for combining and manipulating shaders.
```
djeedai, line 3: Any reason to mix `_` and `-`? It's already bad that Rust crates use both, but within the same name...

tychedelia: Oh no that's just a mistake!

tychedelia:
Not planned, we're hoping to merge support for WESL and continue on that path.

PR #16499, opened 2024-11-25
Run observers before hooks for `on_replace` and `on_remove`

### LINKED ISSUE #16498: Hook and observer ordering for `on_remove`/`on_replace` should be inverted
## Bevy version

0.15.0-rc.3 (exists before as well)

## What you did

When adding a component, the order of hooks and observers are:
- on_add hook
- on_add observers
- on_insert hook
- on_insert observers

When removing a component, the order of hooks and observers are:
- on_replace hook
- on_replace observers
- on_remove hook
- on_remove observers

## What went wrong

The remove ordering matches the add ordering, while it would be more natural for it to be inverted – running observers before hooks, that is.

## Additional information

Note that this bug is to provoke **discussion**. The current behaviour may be considered correct as well. So this bug should be used to discuss the matter and make a decision so that the decision is then documented for posterity.

nakedible:
# Objective

- Fixes #16498 

## Solution

- Trivially swaps ordering of hooks and observers for all call sites where they are triggered for `on_replace` or `on_remove`

## Testing

- Just CI

---

## Migration Guide

The order of hooks and observers for `on_replace` and `on_remove` has been swapped. Observers are now run before hooks. This is a more natural ordering where the removal ordering is inverted compared to the insertion ordering.

### DIFF
--- a/crates/bevy_ecs/src/bundle.rs
+++ b/crates/bevy_ecs/src/bundle.rs
@@ -902,7 +902,6 @@ impl<'w> BundleInserter<'w> {
             let mut deferred_world = self.world.into_deferred();
 
             if insert_mode == InsertMode::Replace {
-                deferred_world.trigger_on_replace(archetype, entity, add_bundle.iter_existing());
                 if archetype.has_replace_observer() {
                     deferred_world.trigger_observers(
                         ON_REPLACE,
@@ -910,6 +909,7 @@ impl<'w> BundleInserter<'w> {
                         add_bundle.iter_existing(),
                     );
                 }
+                deferred_world.trigger_on_replace(archetype, entity, add_bundle.iter_existing());
             }
         }
 
--- a/crates/bevy_ecs/src/world/entity_ref.rs
+++ b/crates/bevy_ecs/src/world/entity_ref.rs
@@ -4910,14 +4910,14 @@ mod tests {
             "OrdB hook on_insert",
             "OrdB observer on_insert",
             "OrdB command on_add", // command added by OrdB hook on_add, needs to run before despawn command
-            "OrdA hook on_replace", // start of despawn
-            "OrdB hook on_replace",
-            "OrdA observer on_replace",
+            "OrdA observer on_replace", // start of despawn
             "OrdB observer on_replace",
-            "OrdA hook on_remove",
-            "OrdB hook on_remove",
+            "OrdA hook on_replace",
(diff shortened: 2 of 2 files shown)


bushrat011899 (approved):
This makes sense to me conceptually. Nice work!

nakedible:
The fix pull runs now hooks after observers, but observers are run in the same order as on insertion, not in reverse order. Flecs says observer ordering is undefined, and I also think that having undefined ordering is better than saying observers are called in insertion order (or reverse insertion order). So I believe it's fine that the observers are not run in reverse order by this pull.

nakedible:
If guaranteed observer ordering is added (in a separate pull), then that pull should also add functionality to call the observers in reverse order, and this reverse order should be used for `OnReplace` and `OnRemove` observers.

BenjaminBrienen (approved):

nakedible:
@alice-i-cecile Two approving community reviews. Not certain of the process, but -S-Needs-Review +S-Ready-For-Final-Review?

nakedible:
The reason this failed the merge queue is exactly because my other pull added checks for the order of hooks vs. observers :-) the only fix needed is to fix the testcase, which is exactly correct.

nakedible:
```
assertion `left == right` failed
  left: ["OrdA hook on_add", "OrdA observer on_add", "OrdA hook on_insert", "OrdA observer on_insert", "OrdB hook on_add", "OrdB observer on_add", "OrdB hook on_insert", "OrdB observer on_insert", "OrdB command on_add", "OrdA observer on_replace", "OrdB observer on_replace", "OrdA hook on_replace", "OrdB hook on_replace", "OrdA observer on_remove", "OrdB observer on_remove", "OrdA hook on_remove", "OrdB hook on_remove"]
 right: ["OrdA hook on_add", "OrdA observer on_add", "OrdA hook on_insert", "OrdA observer on_insert", "OrdB hook on_add", "OrdB observer on_add", "OrdB hook on_insert", "OrdB observer on_insert", "OrdB command on_add", "OrdA hook on_replace", "OrdB hook on_replace", "OrdA observer on_replace", "OrdB observer on_replace", "OrdA hook on_remove", "OrdB hook on_remove", "OrdA observer on_remove", "OrdB observer on_remove"]
```

BenjaminBrienen (approved):

nakedible:
@alice-i-cecile Mind bumping this back into the merge queue?

PR #16502, opened 2024-11-25
Multiple box shadow support

ickshonpe:
# Objective

Add support for multiple box shadows on a single `Node`.

## Solution

* Rename `BoxShadow` to `ShadowStyle` and remove its `Component` derive.
* Create a new `BoxShadow` component that newtypes a `Vec<ShadowStyle>`.
* Add a `new` constructor method to `BoxShadow` for single shadows.
* Change `extract_shadows` to iterate through a list of shadows per node.

Render order is determined implicitly from the order of the shadows stored in the `BoxShadow` component, back-to-front. 
Might be more efficient to use a `SmallVec<[ShadowStyle; 1]>` for the list of shadows but not sure if the extra friction is worth it.

## Testing

Added a node with four differently coloured shadows to the `box_shadow` example.

---

## Showcase

```
cargo run --example box_shadow
```

four-shadow

## Migration Guide

Bevy UI now supports multiple shadows per node. A new struct `ShadowStyle` is used to set the style for each shadow. And the `BoxShadow` component is changed to a tuple struct wrapping a vector containing a list of `ShadowStyle`s. To spawn a node with a single shadow you can use the `new` constructor function:
```rust
commands.spawn((
    Node::default(),
    BoxShadow::new(
        Color::BLACK.with_alpha(0.8),
        Val::Percent(offset.x),
        Val::Percent(offset.y),
        Val::Percent(spread),
        Val::Px(blur),
    )
));
```

### DIFF
--- a/crates/bevy_ui/src/ui_node.rs
+++ b/crates/bevy_ui/src/ui_node.rs
@@ -2448,6 +2448,11 @@ impl BoxShadow {
 
 #[derive(Copy, Clone, Debug, PartialEq, Reflect)]
 #[reflect(PartialEq, Default)]
+#[cfg_attr(
+    feature = "serialize",
+    derive(serde::Serialize, serde::Deserialize),
+    reflect(Serialize, Deserialize)
+)]
 pub struct ShadowStyle {
     /// The shadow's color
     pub color: Color,
--- a/crates/bevy_ui/src/ui_node.rs
+++ b/crates/bevy_ui/src/ui_node.rs
@@ -2436,13 +2436,13 @@ impl BoxShadow {
         spread_radius: Val,
         blur_radius: Val,
     ) -> Self {
-        Self(ShadowStyle {
+        Self(vec![ShadowStyle {
             color,
             x_offset,
             y_offset,
             spread_radius,
             blur_radius,
-        })
+        }])
     }
 }
 
--- a/crates/bevy_ui/src/ui_node.rs
+++ b/crates/bevy_ui/src/ui_node.rs
@@ -2424,7 +2424,8 @@ impl ResolvedBorderRadius {
     derive(serde::Serialize, serde::Deserialize),
     reflect(Serialize, Deserialize)
 )]
-/// List of shadows to draw for a `Node`.
+/// List of shadows to draw for a [`Node`].
+/// Draw order is determined implicitly from the vector of [`ShadowStyle`]s, back-to-front.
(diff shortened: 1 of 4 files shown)


crates/bevy_ui/src/ui_node.rs
```diff
@@ -2416,14 +2417,49 @@ impl ResolvedBorderRadius {
...
-#[derive(Component, Copy, Clone, Debug, PartialEq, Reflect)]
+#[derive(Component, Clone, Debug, Default, PartialEq, Reflect, Deref, DerefMut)]
 #[reflect(Component, PartialEq, Default)]
 #[cfg_attr(
     feature = "serialize",
     derive(serde::Serialize, serde::Deserialize),
     reflect(Serialize, Deserialize)
 )]
-pub struct BoxShadow {
+/// List of shadows to draw for a `Node`.
```
alice-i-cecile, line 2427: Missing doc link.

alice-i-cecile:
I like the feature and the API. I'd probably do a smallvec myself, but we should bench that and do it in follow-up.

crates/bevy_ui/src/ui_node.rs
```diff
@@ -2416,14 +2417,50 @@ impl ResolvedBorderRadius {
...
-#[derive(Component, Copy, Clone, Debug, PartialEq, Reflect)]
+#[derive(Component, Clone, Debug, Default, PartialEq, Reflect, Deref, DerefMut)]
 #[reflect(Component, PartialEq, Default)]
 #[cfg_attr(
     feature = "serialize",
     derive(serde::Serialize, serde::Deserialize),
     reflect(Serialize, Deserialize)
 )]
-pub struct BoxShadow {
+/// List of shadows to draw for a [`Node`].
```
alice-i-cecile, line 2427:
```suggestion
/// List of shadows to draw for a [`Node`].
///
```

ickshonpe:
Wondering if it's more or less confusing to call optional components non-required/unrequired instead 😕

ickshonpe:
> I like the feature and the API. I'd probably do a smallvec myself, but we should bench that and do it in follow-up.

Yeah it seems to me in practical terms it's unusual to use a very large number of shadows and if you do it's most likely because you're are using multiple shadows in which case the smallvec will allocate on the heap anyway.

doup (approved):
I was thinking it was going to be more complex, same but in a loop. 👌

Issue #15712, opened 2024-10-07
Reflection is not symmetrical: round tripping can fail


UkoeHB:
## Bevy version

0.14.2

## What you did

1. Define unit struct with reflection and ser/de.
2. Use `TypedReflectDeserializer` to deserialize raw JSON into a reflection box.
3. Use `T::from_reflect` to get a `T` out of the box.
4. Re-serialize `T` directly to JSON.
5. Expect the reserialized JSON to equal the starting JSON.

## What went wrong

The `TypedReflectDeserializer` expects unit structs to be JSON arrays. However, JSON *serializes* unit structs to `null`. If you try to use `null` in the original raw JSON, then you get `Error("invalid type: null, expected reflected struct value", ...)`.

## Additional information

Minimum reproducible:

```toml
bevy = { version = "0.14", default-features = true, features = ["serialize"] }
serde = { version = "1.0" }
serde_json = { version = "1.0", features = ["preserve_order"] }
```

```rust
use std::str::FromStr;
use std::fmt::Debug;

use bevy::{prelude::*, reflect::serde::TypedReflectDeserializer};
use serde::de::DeserializeSeed;

use serde::{Deserialize, Serialize};

#[derive(Reflect, Default, Debug, PartialEq, Serialize, Deserialize)]
pub struct UnitStruct;

fn test_equivalence<T>(w: &World, json_raw: &str, val: T)
where
    T: Serialize + for<'de> Deserialize<'de> + Debug + Reflect + FromReflect + PartialEq
{
    // Json raw to json value
    let json_val = serde_json::Value::from_str(json_raw).unwrap();

    // Json value to reflect
    let type_registry = w.resource::<AppTypeRegistry>().read();
    let registration = type_registry.get(std::any::TypeId::of::<T>()).unwrap();
    let deserializer = TypedReflectDeserializer::new(registration, &type_registry);
    let reflected = deserializer.deserialize(json_val.clone()).unwrap();

    // Reflect to val
    let extracted = T::from_reflect(reflected.as_reflect()).unwrap();
    assert_eq!(val, extracted);

    // Val to json value
    let json_val_deser = serde_json::to_value(&val).unwrap();
    assert_eq!(json_val, json_val_deser);
}

#[test]
fn bug()
{
    let mut app = App::new();
    app.register_type::<UnitStruct>();
    test_equivalence(app.world(), "[]", UnitStruct);
}
```

MrGVSV:
[...]

PR #16507, opened 2024-11-25
bugfix: Reflection is not symmetrical 

pikerpoler:
# Objective
fixing #15712

before:
- unit structs were serialized into null
- deserializing null caused an error
- empty arrays deserialized into unit structs

after:
- unit structs are serialized into null
- null deserialize into unit structs
- deserializing empty arrays cause an error

## Solution

checking for the number of fields in the reflect struct and deserializing the json using `deserialize_unit` which handles the types correctly
additionally I had to (trivially) implement visit_unit because the default implementation fails with a type error

## Testing

for testing I added a main file (which is not included as a file change for obvious reasons) with the following code:

```rust
fn main() {}

mod tests {
    use std::fmt::Debug;
    use std::str::FromStr;

    use bevy::{prelude::*, reflect::serde::TypedReflectDeserializer};
    use serde::de::DeserializeSeed;

    use serde::{Deserialize, Serialize};

    #[derive(Reflect, Default, Debug, PartialEq, Serialize, Deserialize)]
    struct UnitStruct;

    fn test_equivalence<T>(w: &World, json_raw: &str, val: T)
    where
        T: Serialize + for<'de> Deserialize<'de> + Debug + Reflect + FromReflect + PartialEq,
    {
        // Json raw to json value
        let json_val = serde_json::Value::from_str(json_raw).unwrap();

        // Json value to reflect
        let type_registry = w.resource::<AppTypeRegistry>().read();
        let registration = type_registry.get(std::any::TypeId::of::<T>()).unwrap();
        let deserializer = TypedReflectDeserializer::new(registration, &type_registry);
        let reflected = deserializer.deserialize(json_val.clone()).unwrap();

        // Reflect to val
        let extracted = T::from_reflect(reflected.as_ref()).unwrap();
        assert_eq!(val, extracted);

        // Val to json value
        let json_val_deser = serde_json::to_value(&val).unwrap();

        assert_eq!(json_val, json_val_deser);
    }

    #[test]
    fn bug() {
        let mut app = App::new();
        app.register_type::<UnitStruct>();
        test_equivalence(app.world(), "null", UnitStruct);
    }
}
```
for this code to run, you need to add a `serde_json` dependancy to Aargo.toml
```rust
serde_json = { version = "1.0", features = ["preserve_order"] }
```

I also tested manually to check if deserializing an empty array as a unit object causes an error which it does.

```rust
```

</details>

## Migration Guide

If you serialized empty arrays into unit objects, now you need to replace them with nulls.

```"[]"``` -> ```"null"```
```json!([])``` ->  ```json!(null)```


BenjaminBrienen (changes requested):
Really well done! Can you please add a couple unit tests to demonstrate that it works and to prevent regressions?

BenjaminBrienen:
> after:
> deserializing empty arrays cause an error

Why is this the desired result? Is it ever possible to serialize into an empty array?

pikerpoler:
> Really well done! Can you please add a couple unit tests to demonstrate that it works and to prevent regressions?

I am working on it. right after I published the PR I realized the dev-dependencies allow me to test this inside the crate.

UkoeHB:
@pikerpoler unfortunately that fix is not strictly valid, because `#[reflect(ignore)]` can cause a struct to 'appear' to have zero fields to reflection when in fact it has fields.

pikerpoler:
> [@pikerpoler](https://github.com/pikerpoler) unfortunately that fix is not strictly valid, because `#[reflect(ignore)]` can cause a struct to 'appear' to have zero fields to reflection when in fact it has fields.

we may solve it in a different way, using `ReflectDerive::UnitStruct` instead. I considered this initially, but it may require a larger refactor, and this suggestion:
> ...We can probably do that pretty easy with a simple field length check (being sure to exclude `skip_serializing` fields). 
worked for my tests

the section I think of changing is this:
https://github.com/bevyengine/bevy/blob/efd2982a881e8756dac14e740c407b249afeefaa/crates/bevy_reflect/derive/src/remote.rs#L52

but It will take a bit more time to figre out and test

PR #16508, opened 2024-11-25
Elliptical border radius

ickshonpe:
# Objective

Elliptical border radius support

## Testing
---

## Migration Guide

### DIFF
--- a/crates/bevy_ui/src/geometry.rs
+++ b/crates/bevy_ui/src/geometry.rs
@@ -615,6 +615,67 @@ impl Default for UiRect {
     }
 }
 
+#[derive(Copy, Clone, Debug, PartialEq, Reflect)]
+#[reflect(PartialEq, Default, Debug)]
+#[cfg_attr(
+    feature = "serialize",
+    derive(serde::Serialize, serde::Deserialize),
+    reflect(Serialize, Deserialize)
+)]
+pub struct UiAxes {
+    pub horizontal: Val,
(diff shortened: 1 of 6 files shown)


viridia:
This needs both a description and a justification.

Honestly, if I really needed elliptical corners, I'd probably do them in one of two ways:
* Use a nine-patch texture.
* Fake it by using circular corners and then apply a non-uniform scaling to the entity.

Given that the `BorderRadius` component is already somewhat unwieldy to work with, I don't see that much benefit in adding additional features.

ickshonpe:
> This needs both a description and a justification.
> 
> Honestly, if I really needed elliptical corners, I'd probably do them in one of two ways:
> 
> * Use a nine-patch texture.
> * Fake it by using circular corners and then apply a non-uniform scaling to the entity.
> 
> Given that the `BorderRadius` component is already somewhat unwieldy to work with, I don't see that much benefit in adding additional features.

I think it is something we want eventually but yeah it's not a priority which is this has remained a draft and I haven't done any more with it.

It's frustrating how much extra complexity and friction support for niche features like this adds. 99% of the time you only want homogenous corners with circular arcs.  The multi-colored `BorderColor` support is annoying in the same way, there's probably not one bevy app that uses it but everyone has to pay for it with the more awkward API now. Maybe we could think about having a feature gate with a much simplified API but then that might become a pain to maintain.

PR #16511, opened 2024-11-25
ensure all Material2d rendering happen in order

mockersf:
# Objective

- `Material2d` rendering order is random
- This can be seen in example `wireframe_2d`, wireframe is randomly visible 50% of the time

## Solution

- Add a temp resource counting the materials added
- Add sets to materials with a unique value
- Order each material to be after those added before
- *This means rendering order is dependent on plugin addition order*

## Testing

```sh
steps=10; git restore .;
echo "wireframe_2d" > test
cargo run -p example-showcase -- run --stop-frame 250 --screenshot-frame 100 --fixed-frame-time 0.05 --example-list test --in-ci;
mv screenshots base;
for prefix in `seq 0 $steps`;
do
  echo step $prefix;
  cargo run -p example-showcase -- run --stop-frame 250 --screenshot-frame 100 --fixed-frame-time 0.05 --example-list test;
  mv screenshots $prefix-screenshots;
done;
mv base screenshots
for prefix in `seq 0 $steps`;
do
  echo check $prefix
  for file in screenshots/*/*;
  do
    echo $file;
    diff $file $prefix-$file;
  done;
done;
```

### DIFF
--- a/crates/bevy_sprite/src/mesh2d/material.rs
+++ b/crates/bevy_sprite/src/mesh2d/material.rs
@@ -236,6 +236,12 @@ impl<M: Material2d> Default for Material2dPlugin<M> {
     }
 }
 
+#[derive(Resource, Debug)]
+struct MaterialCounts(usize);
+
+#[derive(Debug, Hash, PartialEq, Eq, Clone, SystemSet)]
+struct MaterialSet(usize);
+
 impl<M: Material2d> Plugin for Material2dPlugin<M>
 where
     M::Data: PartialEq + Eq + Hash + Clone,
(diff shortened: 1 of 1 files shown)


IceSentry:
I'm not sure how I feel about this. I think the issue is wireframes being wrong. It doesn't feel like that's the rigth fix for the issue. I need to think about this more.

mockersf:
I'm not happy about making it dependent on plugin order, but the fact that `Material2d` order is random is wrong

BenjaminBrienen (approved):

pcwalton:
The problem with wireframes is that they break the 1:1 mapping between entities and meshes/materials, which breaks things in lots of ways that can't be fully covered up by forcing the rendering order. One way to fix wireframes properly would be to add a `WireframePass` and treat wireframes as a sort of prepass-like thing.

IceSentry:
Yeah, at the time I ported the wireframe from their own pass to the current wireframe material thing because it simplified a lot of the code, but in hindsight that isn't as ideal of a solution. It's nice that the code is a bit simpler but it's probably less efficient than it could be.

IceSentry:
The 2d wireframes now use their own render phase so material 2d order should be deterministic

PR #16514, opened 2024-11-26
Refactor `FunctionSystem` to use a single `Option`

ItsDoot:
# Objective

Combine the `Option<_>` state in `FunctionSystem` into a single `Option` to provide clarity and save space.

## Solution

Simplifies `FunctionSystem`'s layout by using a single `Option<FunctionSystemState>` for state that must be initialized before running, and saves a byte by removing the need to store an enum tag. Additionally, calling `System::run` on an uninitialized `System` will now give a more descriptive message prior to verifying the `WorldId`.

## Testing

Ran CI checks locally.

### DIFF
--- a/crates/bevy_ecs/src/system/function_system.rs
+++ b/crates/bevy_ecs/src/system/function_system.rs
@@ -611,7 +611,7 @@ where
     marker: PhantomData<fn() -> Marker>,
 }
 
-/// Stores the state of a [`FunctionSystem`], which must be initialized with
+/// The state of a [`FunctionSystem`], which must be initialized with
 /// [`System::initialize`] before the system can be run. A panic will occur if
 /// the system is run without being initialized.
 struct FunctionSystemState<Marker, F>
@@ -683,7 +683,8 @@ where
     /// Message shown when a system isn't initialized
     // When lines get too long, rustfmt can sometimes refuse to format them.
     // Work around this by storing the message separately.
-    const ERROR_UNINITIALIZED: &'static str = "System's param_state was not found. Did you forget to initialize this system before running it?";
+    const ERROR_UNINITIALIZED: &'static str =
+        "System's state was not found. Did you forget to initialize this system before running it?";
 }
 
 impl<Marker, F> System for FunctionSystem<Marker, F>
@@ -735,11 +736,7 @@ where
 
         let change_tick = world.increment_change_tick();
 
-        let param_state = self
-            .state
-            .as_mut()
-            .map(|state| &mut state.param)
-            .expect(Self::ERROR_UNINITIALIZED);
+        let param_state = &mut self.state.as_mut().expect(Self::ERROR_UNINITIALIZED).param;
         // SAFETY:
         // - The caller has invoked `update_archetype_component_access`, which will panic
         //   if the world does not match.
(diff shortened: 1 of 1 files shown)


crates/bevy_ecs/src/system/function_system.rs
```diff
@@ -602,14 +604,28 @@ where
...
 {
     func: F,
-    pub(crate) param_state: Option<<F::Param as SystemParam>::State>,
-    pub(crate) system_meta: SystemMeta,
-    world_id: Option<WorldId>,
+    state: Option<FunctionSystemState<Marker, F>>,
+    system_meta: SystemMeta,
     archetype_generation: ArchetypeGeneration,
     // NOTE: PhantomData<fn()-> T> gives this safe Send/Sync impls
     marker: PhantomData<fn() -> Marker>,
```
benfrankel, line 611: Is this still needed with `Marker` now appearing in `FunctionSystemState<Marker, F>`?

crates/bevy_ecs/src/system/function_system.rs
```diff
@@ -669,7 +683,7 @@ where
     /// Message shown when a system isn't initialized
     // When lines get too long, rustfmt can sometimes refuse to format them.
     // Work around this by storing the message separately.
-    const PARAM_MESSAGE: &'static str = "System's param_state was not found. Did you forget to initialize this system before running it?";
+    const ERROR_UNINITIALIZED: &'static str = "System's param_state was not found. Did you forget to initialize this system before running it?";
```
benfrankel, line 686: `param_state` -> `state` or reworded here?

crates/bevy_ecs/src/system/function_system.rs
```diff
@@ -721,39 +735,50 @@ where
...
     }
 
     #[inline]
     unsafe fn validate_param_unsafe(&mut self, world: UnsafeWorldCell) -> bool {
-        let param_state = self.param_state.as_ref().expect(Self::PARAM_MESSAGE);
+        let param_state = self
+            .state
+            .as_ref()
+            .map(|state| &state.param)
+            .expect(Self::ERROR_UNINITIALIZED);
```
benfrankel, line 781:
```suggestion
        let param_state = self
            .state
            .as_ref()
            .expect(Self::ERROR_UNINITIALIZED)
            .param;
```
Does this work? If so, it's a bit simpler.
(same comment for the other places this line appears)

crates/bevy_ecs/src/system/function_system.rs
```diff
@@ -602,14 +604,28 @@ where
...
 }
 
+/// Stores the state of a [`FunctionSystem`], which must be initialized with
+/// [`System::initialize`] before the system can be run. A panic will occur if
+/// the system is run without being initialized.
+struct FunctionSystemState<Marker, F>
+where
+    F: SystemParamFunction<Marker>,
+{
+    /// Stores the cached state of the system's [`SystemParam`]s.
```
benfrankel, line 621:
```suggestion
    /// The cached state of the system's [`SystemParam`]s.
```
Optional nit

crates/bevy_ecs/src/system/function_system.rs
ItsDoot: It no longer appears in `FunctionSystemState`, so I'm assuming it still needs to exist.

benfrankel (approved):
I don't know the potential implications of this change, but I've reviewed the diff and it makes sense to me. `Option<FunctionSystemState>` makes the uninitialized vs initialized semantics for function systems obvious.

chescock (approved):

BenjaminBrienen (approved):

mockersf (approved):

PR #16515, opened 2024-11-26
Add PhysicsSystems and AiSystems system sets.

andriyDev:
# Objective

- Allow ordering systems against unknown physics engine / AI crate.

## Solution

- Add an unused generic `PhysicsSystems` and `AiSystems` system set that crates can opt into.

---

## Showcase

The `PhysicsSystems` and `AiSystems` are useful general system sets for ordering systems relative to unknown crates. For example, your crate may care that it manipulates transforms before a physics engine. Physics engines can put their systems in this system set, and your crate can order its systems before this system set. This allows crates to avoid pulling in extra dependencies just for their system sets!

### DIFF
--- a/crates/bevy_app/src/main_schedule.rs
+++ b/crates/bevy_app/src/main_schedule.rs
@@ -185,6 +185,18 @@ pub struct Last;
 #[derive(SystemSet, Debug, Hash, PartialEq, Eq, Clone)]
 pub struct Animation;
 
+/// A generic system set for physics. This is unused by Bevy but provides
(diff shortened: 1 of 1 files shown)


cart:
`AiSystems` feels a _bit_ too generic to me. "AI" can be arbitrary app logic, whereas "Physics" is more often handled in a specific "physics step".

BenjaminBrienen:
AI might not be the best word for that anymore and we can make the use case more clear. Maybe call it "AgentSystems" (for ordering against when entities with agency make decisions).

PR #16516, opened 2024-11-26
one shot system cleanup

### LINKED ISSUE #16497: Clean up one-shot systems API
There are a few inconsistencies and gaps in the current one-shot systems API:

- There should be a `Commands::unregister_system_cached` method
- `Commands::` and `World::run_system_with_input` should be renamed to `run_system_with` for consistency
- `Commands::` and `World::run_system_once_with` should accept `(system, input)` in that order rather than `(input, system)` for consistency

pemattern:
# Objective

- Fixes #16497
- This is my first PR, so I'm still learning to contribute to the project

## Solution

- Added struct `UnregisterSystemCached` and function `unregister_system_cached`
- renamed `World::run_system_with_input` to `run_system_with`
- reordered input parameters for `World::run_system_once_with`

## Testing

- Added a crude test which registers a system via `World::register_system_cached`, and removes it via `Command::unregister_system_cached`.

## Migration Guide

- Change all occurrences of `World::run_system_with_input` to `World::run_system_with`.
- swap the order of input parameters for `World::run_system_once_with` such that the system comes before the input.


pemattern:
added migration guide

crates/bevy_ecs/src/system/system_registry.rs
```diff
@@ -570,6 +570,42 @@ where
...
 }
 
+/// The [`Command`] type for unregistering one-shot systems from [`Commands`](crate::system::Commands).
+pub struct UnregisterSystemCached<I, O, M, S>
+where
+    I: SystemInput + 'static,
+    S: IntoSystem<I, O, M> + Send + 'static,
+{
+    system: S,
+    _phantom: PhantomData<(fn() -> I, fn() -> O, fn() -> M)>,
```
benfrankel, line 580:
```suggestion
    _phantom: PhantomData<fn() -> (I, O, M)>,
```
Optional

crates/bevy_ecs/src/system/system_registry.rs
```diff
@@ -570,6 +570,42 @@ where
     }
 }
 
+/// The [`Command`] type for unregistering one-shot systems from [`Commands`](crate::system::Commands).
+pub struct UnregisterSystemCached<I, O, M, S>
+where
+    I: SystemInput + 'static,
+    S: IntoSystem<I, O, M> + Send + 'static,
+{
+    system: S,
```
benfrankel, line 579: Technically `world.unregister_system_cached` doesn't actually use the system argument passed into it, so this field could be removed. We could add a private `World::unregister_system_cached_inner::<S>` method that `World::unregister_system_cached` and this command both call into, using `S` directly instead of passing in a value of type `S`.

This isn't a big deal though so I'm not blocking on it.

crates/bevy_ecs/src/system/commands/mod.rs
```diff
@@ -2475,6 +2490,23 @@ mod tests {
...
     }
 
+    #[test]
+    fn unregister_system_cached_commands() {
+        let mut world = World::default();
+        let mut queue = CommandQueue::default();
+
+        fn nothing() {}
+        assert!(world.iter_resources().count() == 0);
+        let _id = world.register_system_cached(nothing);
```
benfrankel, line 2500: Could actually use this system ID to check that the entity it points to (`SystemId::entity()`) exists, and later that the entity no longer exists after unregister.

crates/bevy_ecs/src/system/system_registry.rs
chescock: Another option is to remove the `IntoSystem` constraint from the type entirely and just have 

```rust
pub struct UnregisterSystemCached<S> {
    system: S,
}
```

You need the constraints on `fn unregister_system_cached` to have the `Command` impl, but I don't think you actually need them on the `struct`.

chescock (approved):

crates/bevy_ecs/src/system/commands/mod.rs
pemattern: Much better, thanks!

crates/bevy_ecs/src/system/commands/mod.rs
```diff
@@ -2475,6 +2490,22 @@ mod tests {
...
+
+        fn nothing() {}
+
+        let id = world.register_system_cached(nothing);
+        assert!(world.get_entity(id.entity).is_ok());
+
+        let mut commands = Commands::new(&mut queue, &world);
+        commands.unregister_system_cached(nothing);
+        queue.apply(&mut world);
+        assert!(world.get_entity(id.entity).is_err());
```
benfrankel, line 2506: The resource asserts would still be useful here alongside the entity asserts.

benfrankel (changes requested):
Will approve after this change.

benfrankel (approved):

BenjaminBrienen (approved):

PR #16517, opened 2024-11-26
Update `sysinfo` version to 0.32.1

GuillaumeGomez:
Fixes bug for new libc release.

PR #16523, opened 2024-11-26
Revert "Update `sysinfo` version to 0.32.1 (#16517)"

cart:
This reverts commit 3476a9f3a658fe7c8326ca0e618ad543914ca7f4.

# Objective

#16517 makes it impossible to select a libc version, now that the "breaking" libc release was yanked

## Solution

Revert the version bump so we can select a version of sysinfo that builds (the previous version)


mockersf:
not needed on the 0.15 release

PR #16524, opened 2024-11-27
Skip empty spans when updating text buffers

### LINKED ISSUE #16521: Nested text spans are not rendered if the `default_font` feature is disabled
## Bevy version

`0.15.0-rc.3`, also current main

```
```

## What you did

- Disable `default-features`
- Decline to enable `default_font` feature
- Use nested text spans where the (empty) root node uses `TextFont::default()` either by omission of the component, or by explicitly setting it to that value.

https://github.com/rparrett/bevy-default-font-bug

```
git clone https://github.com/rparrett/bevy-default-font-bug
cd bevy-default-font-bug
cargo run
```

## What went wrong

No text is displayed

## Additional Information

Relevant snippet pasted here for convenience

```rust
let font = asset_server.load("FiraSans-Bold.ttf");
commands
    .spawn((
        Text::default(),
        // Uncommenting this makes things work
        // TextFont {
        //     font: font.clone(),
        //     ..default()
        // },
    ))
    .with_children(|parent| {
        parent.spawn((
            TextSpan::new("span, should be visible"),
            TextFont { font, ..default() },
        ));
    });
```

cart:
# Objective

Fixes #16521

## Solution

If an empty span is encountered (such as the default `Text` value), we skip it entirely when updating buffers. This prevents unnecessarily bailing when the font doesn't exist (ex: when the default font is disabled)

### DIFF
--- a/crates/bevy_text/src/pipeline.rs
+++ b/crates/bevy_text/src/pipeline.rs
@@ -107,6 +107,9 @@ impl TextPipeline {
         computed.entities.clear();
 
         for (span_index, (entity, depth, span, text_font, color)) in text_spans.enumerate() {
+            // Save this span entity in the computed text block.
+            computed.entities.push(TextEntity { entity, depth });
+
             if span.is_empty() {
                 continue;
             }
@@ -124,9 +127,6 @@ impl TextPipeline {
 
                 return Err(TextError::NoSuchFont);
             }
-
-            // Save this span entity in the computed text block.
-            computed.entities.push(TextEntity { entity, depth });
 
             // Get max font size for use in cosmic Metrics.
             font_size = font_size.max(text_font.font_size);
@@ -226,7 +226,6 @@ impl TextPipeline {
     ) -> Result<(), TextError> {
         layout_info.glyphs.clear();
         layout_info.size = Default::default();
-        println!("QUEUED");
 
         // Clear this here at the focal point of text rendering to ensure the field's lifecycle has strong boundaries.
         computed.needs_rerender = false;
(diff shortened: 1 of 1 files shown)


crates/bevy_text/src/pipeline.rs
```diff
@@ -223,6 +226,7 @@ impl TextPipeline {
     ) -> Result<(), TextError> {
         layout_info.glyphs.clear();
         layout_info.size = Default::default();
+        println!("QUEUED");
```
UkoeHB, line 229: Leftover?

crates/bevy_text/src/pipeline.rs
```diff
@@ -107,6 +107,9 @@ impl TextPipeline {
         computed.entities.clear();
 
         for (span_index, (entity, depth, span, text_font, color)) in text_spans.enumerate() {
+            if span.is_empty() {
+                continue;
+            }
```
UkoeHB, line 112: For correct indexing, this should go after `computed.entities.push(TextEntity { entity, depth });`.

UkoeHB (changes requested):

crates/bevy_text/src/pipeline.rs
cart: YUUUP

crates/bevy_text/src/pipeline.rs
cart: Ah yeah. I thought I tested this properly, but it was setting the wrong span and I didn't know because the colors were the same. Sure enough, giving the spans different colors surfaced an indexing problem, and the reorder fixes it.

cart: Nice catch!

UkoeHB (approved):

viridia (approved):

PR #16527, opened 2024-11-27
Clarify that `bevy_app::App.world()` (and mut variant) returns the main `SubApp`'s `World`

LikeLakers2:
# Objective
The documentation for `bevy_app::App.world()` (and its mut variant) could confuse some into thinking that this is the only World that the App will contain.

## Solution
Clarify the documentation for `bevy_app::App.world()` (and its mut variant), to say that it returns the main subapp's world. This helps imply that Apps can contain more than one world (albeit, only one per SubApp).

## Testing
This is a documentation change, with no changes to doctests. Thus, testing is not necessary beyond ensuring the link syntax is correct.

### DIFF
--- a/crates/bevy_app/src/app.rs
+++ b/crates/bevy_app/src/app.rs
@@ -1008,12 +1008,18 @@ impl App {
             .try_register_required_components_with::<T, R>(constructor)
     }
 
-    /// Returns a reference to the [`World`].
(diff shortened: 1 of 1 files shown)


alice-i-cecile (approved):
This is definitely clearer.

Vrixyz (approved):
Great!

This opens the idea to have a `MainWorld` and `RenderWorld` akin to entities.. :thinking: but yeah another whole can of worms :)

PR #11279, opened 2024-01-09
bevy_gltf: Add support for using name-based labels

dmlary:
# Objective

Allow the use of names while loading assets from a gltf file.
- fixes #2948 #10715

## Solution

This is a first pass at supporting name-based labels for GLTF, and intended to spark discussion.

This is implemented by adding a the settings field `GltfLoaderSettings.label_by_name`. When set to true, the majority of asset labels will be created using the name of the asset in the gltf file if present.  If there is no name, it will fall back to the index-based naming scheme.

When `label_by_name == true`, Animation, Mesh, Node, and Primitive assets will use any available name when adding labeled assets.  For example, the naming for `models/animated/Fox.glb` becomes:
```txt
'Animation/Run', 'Animation/Survey', 'Animation/Walk', 'Material0', 'Mesh/fox1', 'Mesh/fox1.0', 'Node/_rootJoint', 'Node/b_Head_05', 'Node/b_Hip
_01', 'Node/b_LeftFoot01_017', 'Node/b_LeftFoot02_018', 'Node/b_LeftForeArm_010', 'Node/b_LeftHand_011', 'Node/b_LeftLeg01_015', 'Node/b_LeftLeg02_016', 'Node/b_
LeftUpperArm_09', 'Node/b_Neck_04', 'Node/b_RightFoot01_021', 'Node/b_RightFoot02_022', 'Node/b_RightForeArm_07', 'Node/b_RightHand_08', 'Node/b_RightLeg01_019',
 'Node/b_RightLeg02_020', 'Node/b_RightUpperArm_06', 'Node/b_Root_00', 'Node/b_Spine01_02', 'Node/b_Spine02_03', 'Node/b_Tail01_012', 'Node/b_Tail02_013', 'Node/
b_Tail03_014', 'Node/fox', 'Node/root', 'Scene0', 'Skin0', 'Texture0'
```

There is a usability problem here as the load settings must be set every time an asset is loaded from the file.
```rust
        asset_server.load_with_settings(
            "models/animated/Fox.glb#Animation/Survey",
            |s: &mut GltfLoaderSettings| {
                s.label_by_name = true;
            },
        ),
```

On top of that there's a race-condition in the label naming as the first asset AssetServer loads for the file determines how the labels were created.  This is demonstrated in the modified animaled_fox example in this PR.  When it runs, it's unable to find the name-based assets because the index-based assets have been loaded first.  As a result we get the asset not found error:
```txt
[...]

Issue #10715, opened 2023-11-24
No example for how to load animations by name


alice-i-cecile:
## How can Bevy's documentation be improved?

The [animated_fox](https://github.com/bevyengine/bevy/blob/6f56380826ce4641fccb672c92f5c97601041f4c/examples/animation/animated_fox.rs#L34) example shows how to load animations by path, but not by animation name.

By loading in the `Gltf` object directly and then debugging it, I was able to determine that I needed the path, not the animation name. However, no example of how I might use the animation name itself was found, despite it being substantially more human-readable.

mockersf:
Assuming "by path" is `#Animation0`?

The issue is that animations (and every sub asset in a gltf file) is guaranteed to have a path, but labels are optional.

To keep it consistent, labeled assets are always reachable by path, and it's not possible to have the same asset with two labels (`#Animation0` and `#Animation-Idle`).

alice-i-cecile:
> Assuming "by path" is #Animation0?

Correct, that was what I meant.

dmlary:
I'm working on a name-based GLTF workflow as breaking each individual asset I'm building into individual scenes is a headache.  As a datapoint, with just two sets of wall tiles for a hex world, I'm already over 150 nodes.  There's no easy way for me to figure out which index I want to load at this point.

I'm trying to figure out what are the key barriers to using a name-based asset label (`tiles.glb#Node-cell_base_rock`) vs index-based (`tiles.glb#Node178`).  

> To keep it consistent, labeled assets are always reachable by path, and it's not possible to have the same asset with two labels (#Animation0 and #Animation-Idle).

@mockersf is this a technical limitation of the `AssetServer`? Or specific to the bevy_gltf crate?

If it's a technical limitation of `AssetServer`, would there be opposition to a PR that adds a `GltfLoaderSettings.by_name` field that causes assets to be registered with their name instead of index?  With support for falling back to indexed name if there is no name for the asset.

mockersf:
It's a limitation of how assets work, and adding a setting to the gltf loader could be a good idea 👍

dmlary:
I asked in the discord about how difficult it would be to add `LoadContext::add_asset_alias<A<(&mut self, label: String, handle: Handle<A>)` and @NthTensor pointed me at this PR #10153 (Extension-less Assets).

I've confirmed that with that PR it's possible to add a second label for an asset.  This is the change I made to verify.  It crashes on the main branch, succeeds on the PR branch. 
```diff
[...]

Issue #2948, opened 2021-10-10
Support loading GLTF assets by name


pop:
## What problem does this solve or what need does it fill?

To improve ergonomics of loading GLTF assets, Bevy ought to support loading GLTF assets (Mesh/Primitive/Material/Texture/Node/Scene) by Name as well as by Index

## What solution would you like?

Add the ability to fetch assets by name. Something like this:

```rust
let my_gltf_scene_handle = asset_server.load("my_scenes.gltf#Scene1")
```

Could become:

```rust
let my_gltf_scene_handle = asset_server.load("my_scenes.gltf#Scene:'Winter Scene'")
```

This should _probably_ be _additive_. Users can fetch GLTF assets with either Indexing or Naming.

If you think this feature is worth adding, I would be more than happy to implement it! I am enjoying Bevy and would love to contribute. 😄 

## What alternative(s) have you considered?

One could use the [`named_scenes` field on the GLTF struct](https://github.com/bevyengine/bevy/blob/6d6bc2a8b4c243a670b34032de8ed9dec7d70d44/crates/bevy_gltf/src/lib.rs#L31). One would load the GLTF asset file and then spawn specific scenes from the already loaded GLTF asset. In my opinion is not as ergonomic as it could be.

## Additional context

The unofficial Bevy Cheatsheet [makes it seem like this feature currently exists](https://bevy-cheatbook.github.io/features/assets.html?highlight=includes%20names/label#assetpath-and-labels) , but [when I looked through the code](https://github.com/bevyengine/bevy/blob/997eae61854b362c10810884966ac605121e04d7/crates/bevy_gltf/src/loader.rs#L554-L580) this does not seem to be supported. Instead of correcting the cookbook, I think it's worth adding the feature.

djee-ms:
> The unofficial Bevy Cheatsheet makes it seem like this feature currently exists , but when I looked through the code this does not seem to be supported.

I got caught by this, wasted a lot of time :'(

djeedai:
> The unofficial Bevy Cheatsheet makes it seem like this feature currently exists , but when I looked through the code this does not seem to be supported.

I got caught by this, wasted a lot of time trying to figure out what I was doing wrong...

PR #16529, opened 2024-11-27
Changed GLTF Asset Label to use Strings instead of indexes

Hexorg:
* Changed GLTF Asset Label to use Strings instead of indexes
* Removed Primitive name field as GLTF Spec does not allow Primitive names
* Changed all examples to use new asset label format.

# Objective

It's hard to find out GLTF scene/meshes/materials/animation indexes from gltf, which can also change on every export from Blender. Let's introduce `String` labels instead.

## Solution

Replace compatible `GltfAssetLabel` variants from `usize` to `Cow<'static, str>` instead.

## Testing

I ran gltf unit tests, and each example that this pull request touches. 
---

## Showcase

You can now load, e.g. animations with `assets.load("models.glb#Animation:Walk");`

## Migration Guide

GLTF Assets are now keyed by their names, not indexes. Assets that don't have a name, still use their index. For example

```rust
assets.load("fox.glb#Animation:Walk"); // load by name
assets.load("fox.glb#Scene:0"); // load by index
```

It is theoretically possible to have a gltf that has names only for some objects, but not all. Properly developed tools that output GLTF should be consistent in naming, however if they are not - the indexes will NOT be continuous. A good fix for this is to load your GLTF in blender first and assign proper names to items you're interested in.

If an object has  a name it is now also impossible to load it by index.

### DIFF
--- a/assets/animation_graphs/Fox.animgraph.ron
+++ b/assets/animation_graphs/Fox.animgraph.ron
@@ -12,17 +12,17 @@
                 weight: 1.0,
             ),
             (
-                node_type: Clip(AssetPath("models/animated/Fox.glb#Animation0")),
+                node_type: Clip(AssetPath("models/animated/Fox.glb#Animation:Survey")),
                 mask: 0,
                 weight: 1.0,
             ),
             (
-                node_type: Clip(AssetPath("models/animated/Fox.glb#Animation1")),
+                node_type: Clip(AssetPath("models/animated/Fox.glb#Animation:Walk")),
                 mask: 0,
                 weight: 1.0,
             ),
             (
-                node_type: Clip(AssetPath("models/animated/Fox.glb#Animation2")),
+                node_type: Clip(AssetPath("models/animated/Fox.glb#Animation:Run")),
                 mask: 0,
                 weight: 1.0,
             ),
--- a/crates/bevy_gltf/src/lib.rs
+++ b/crates/bevy_gltf/src/lib.rs
@@ -26,7 +26,7 @@
 //!         // This is equivalent to "models/FlightHelmet/FlightHelmet.gltf#Scene0"
 //!         // The `#Scene0` label here is very important because it tells bevy to load the first scene in the glTF file.
 //!         // If this isn't specified bevy doesn't know which part of the glTF file to load.
-//!         SceneRoot(asset_server.load(GltfAssetLabel::Scene(0).from_asset("models/FlightHelmet/FlightHelmet.gltf"))),
+//!         SceneRoot(asset_server.load(GltfAssetLabel::scene("0").from_asset("models/FlightHelmet/FlightHelmet.gltf"))),
 //!         // You can use the transform to give it a position
 //!         Transform::from_xyz(2.0, 0.0, -5.0),
 //!     ));
@@ -91,6 +91,8 @@
 //! You can use [`GltfAssetLabel`] to ensure you are using the correct label.
 
 extern crate alloc;
+
+use alloc::borrow::Cow;
 
 #[cfg(feature = "bevy_animation")]
 use bevy_animation::AnimationClip;
@@ -167,6 +169,11 @@ impl Plugin for GltfPlugin {
             custom_vertex_attributes: self.custom_vertex_attributes.clone(),
         });
     }
+}
+
+fn gltf_name_or_index_string(name: Option<&str>, index: usize) -> String {
+    name.map(ToOwned::to_owned)
+        .unwrap_or_else(|| format!("{}", index))
 }
 
 /// Representation of a loaded glTF file.
@@ -268,7 +275,7 @@ impl GltfNode {
 
     /// Subasset label for this node within the gLTF parent asset.
     pub fn asset_label(&self) -> GltfAssetLabel {
-        GltfAssetLabel::Node(self.index)
+        GltfAssetLabel::Node(self.name.clone().into())
     }
 }
 
@@ -313,7 +320,7 @@ impl GltfSkin {
 
     /// Subasset label for this skin within the gLTF parent asset.
     pub fn asset_label(&self) -> GltfAssetLabel {
-        GltfAssetLabel::Skin(self.index)
+        GltfAssetLabel::Skin(self.name.clone().into())
     }
 }
 
@@ -342,11 +349,7 @@ impl GltfMesh {
     ) -> Self {
         Self {
             index: mesh.index(),
-            name: if let Some(name) = mesh.name() {
-                name.to_string()
-            } else {
-                format!("GltfMesh{}", mesh.index())
-            },
+            name: gltf_name_or_index_string(mesh.name(), mesh.index()),
             primitives,
             extras,
         }
@@ -354,7 +357,7 @@ impl GltfMesh {
 
     /// Subasset label for this mesh within the gLTF parent asset.
     pub fn asset_label(&self) -> GltfAssetLabel {
-        GltfAssetLabel::Mesh(self.index)
+        GltfAssetLabel::Mesh(self.name.clone().into())
     }
 }
 
@@ -365,10 +368,8 @@ impl GltfMesh {
 pub struct GltfPrimitive {
     /// Index of the primitive inside the mesh
     pub index: usize,
-    /// Index of the parent [`GltfMesh`] of this primitive
-    pub parent_mesh_index: usize,
-    /// Computed name for a primitive - either a user defined primitive name from gLTF or a generated name from index
-    pub name: String,
+    /// Name of the parent [`GltfMesh`] of this primitive
+    pub parent_mesh_name: String,
     /// Topology to be rendered.
     pub mesh: Handle<Mesh>,
     /// Material to apply to the `mesh`.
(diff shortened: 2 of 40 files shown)


Hexorg:
This technically addresses [Issue#10715](https://github.com/bevyengine/bevy/issues/10715) though not directly.

andriyDev (approved):
I'm a little suspicious of the "helper methods for creating gltf asset label" functions, but I guess it's fine since most of these labels are pretty small so the function arguments are fairly obvious.

Hexorg:
@andriyDev I was trying to be a bit memory conscious since most asset labels are likely to be &'static str, but there's a chance someone will need it to be built dynamically. 

I can just use String instead of Cow and let everyone call .to_owned() on their labels without helper functions.

Hexorg:
Actually wait, those are unrelated. I have the helper methods just to avoid calling .into()

viridia (approved):

alice-i-cecile:
I much prefer names, but indexes need to continue to be supported, especially for migration ease.

hukasu:
> @andriyDev I was trying to be a bit memory conscious since most asset labels are likely to be &'static str, but there's a chance someone will need it to be built dynamically.
> 
> I can just use String instead of Cow and let everyone call .to_owned() on their labels without helper functions.

the label wouldn't be `'static` if it were defined on a file like `.animgraph`

Alainx277:
> I much prefer names, but indexes need to continue to be supported, especially for migration ease.

Unfortunately that is (afaik) still not possible due to asset aliases never being implemented and has been the blocker for named GLTF asset labels for multiple years (see https://github.com/bevyengine/bevy/issues/2948#issuecomment-939562038).

I strongly feel that Bevy should start using GLTF labels, as a manual index based workflow is extremely error prone (from my own experience). If backwards compatibility is a concern, https://github.com/bevyengine/bevy/pull/11279 is an old PR that gates labels behind a setting.

aloucks:
Does the GLTF spec guarantee that names are unique?

mockersf:
indexes must still be the main way to do labels for gltfs, names can be added but can't replace those.

maybe the fix will come with the editor and a user friendly way to display the label but use the index in code

viridia:
I only ever use names. Typically a GLTF character asset in my game is a file which contains multiple characters, all of which share the same rig and animations. Both individual characters and animation tracks are referenced by name. The reason for combining these together is twofold: ease of authoring in Blender, and reducing the number of i/o operations for loading a large population of actors.

For example, the NPC "Sergeant Stone" has an actor definition file that looks like this:
```json
  "SergeantStone": {
    "type": "Actr",
    "display_name": "Sergeant Stone",
    "aspects": {
      "Skin": "characters/humanoid-male.glb#WithCoatHeavy",
      "Colors": {
        "Hair": "#918981",
        "FacialHair": "#91806b",
        "Coat": "#24378b",
        "CoatTrim": "#435575",
        "Shirt": "#775e31",
        "Belt": "#452e2e",
        "Gauntlets": "#5f4646",
        "Legs": "#4e4437",
        "Boots": "#2b3a48",
        "HatPrimary": "#1c274f",
        "HatDetail": "#686f83"
      },
      "Features": {
        "HairThinning": true,
        "FacialHairMuttonChops": true,
        "HatTricorn": true
      }
    },
    "extends": "#Proto.HumanMaleCoatHeavy"
  },
```
The GLTF model is referenced by the path "characters/humanoid-male.glb#WithCoatHeavy", in the file "humanoid-male" which contains about a dozen different character models (including skeletons) and about two dozen animation tracks. Because I may add more characters or animation tracks at a later time, I cannot depend on the numeric indices remaining stable; as a result I never use the indices if I can help it.

aloucks:
> Because I may add more characters or animation tracks at a later time, I cannot depend on the numeric indices remaining stable; as a result I never use the indices if I can help it.

Why do you need the index to remain stable? Why can't you query all animations from the asset and then build a mapping between the name and the index?

andriyDev:
> Why do you need the index to remain stable? Why can't you query all animations from the asset and then build a mapping between the name and the index?

The trouble with that is the deferred nature of assets: you can't use that mapping until after the GLTF has been loaded. This means you can't spawn anything until your asset has loaded, and you need a system to listen for asset events to fetch that handle, and only then do whatever you need to do. It's possible, but not exactly ideal from code.

viridia:
> The trouble with that is the deferred nature of assets: you can't use that mapping until after the GLTF has been loaded. This means you can't spawn anything until your asset has loaded, and you need a system to listen for asset events to fetch that handle, and only then do whatever you need to do. It's possible, but not exactly ideal from code.

Yes, exactly. What I do now is split the string on the `#` character, load the whole GLTF file, and then trawl through the root looking for the piece I need. This subverts the whole notion of asset labels.

aloucks:
I'm not disputing the idea that loading it by an alias would be nice. I'm just saying changing the internal representation to use a non-unique identifier is just not the right solution, imo.

mockersf:
looking at the graph in https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#concepts-general
gltf

I think we should be ok using names for nodes that don't have a path coming to them.

To start moving forward, I think we should:
- limit this change to scenes and animations
- add a check that verify names are unique on those, and warn and fallback to indices if they are not

aloucks:
> I think we should be ok using names for types that don't have a path coming to them.
> 
> To start moving forward, I think we should:
> 
> * limit this change to scenes and animations
> * add a check that verify names are unique on those, and warn and fallback to indices if they are not

That's going to lead to a lot of inconsistency. Can't this be addressed in the asset loader with a way to provide aliases?

andriyDev:
> Can't this be addressed in the asset loader with a way to provide aliases?

The problem is not providing a way for an `AssetLoader` to register aliases - that is mostly trivial. The problem is what to do with it. When you load a subasset through the `AssetServer` we don't know whether its an alias or not, so we always create a new strong handle to an asset. But once the subasset gets loaded, where does it go? `Mesh0` or `MyMeshName`?

P.S. This may be worth documenting in a separate issue. I feel like "asset aliases" is a concrete enough feature and at least documenting why we can't do it is useful.

aloucks:
> P.S. This may be worth documenting in a separate issue. I feel like "asset aliases" is a concrete enough feature and at least documenting why we can't do it today is useful.

This sounds like more of an issue with the current asset system in general, not a GTLF specific one.

Alainx277:
I did spend an hour or two to check if adding multi-label support would be an easy win, unfortunately (as @andriyDev points out) the way handles work currently makes this quite difficult (at least for someone who does not have an in-depth understanding of the asset subsystem).

In addition to associating multiple labels to a subasset, 1-N handles need to be tracked per subasset. Perhaps this is not as hard as I imagine it, but I stopped at that point.

It would be great if someone who knows the asset system well could more precisely assess how difficult such a change would be. I can imagine it being useful for other kinds of assets as well (ex. sprite sheets, BSN).

mockersf:
> It would be great if someone who knows the asset system well could more precisely assess how difficult such a change would be. I can imagine it being useful for other kinds of assets as well (ex. sprite sheets, BSN).

it's been tried several times, and either you hit a perf issue when you alias, or a memory issue when you duplicate

Alainx277:
If https://github.com/bevyengine/bevy/pull/15813 (@andriyDev again ❤️ ) is merged this problem would get solved by proxy, as two handles could share an `Arc`'d asset, paying only the setup cost of cloning the `Arc` once.

andriyDev:
@Alainx277 Thanks for noticing #15813, haha! The reason I didn't being it up in this discussion is because using it for asset aliases results in two issues: 

1. Cloning the arc means that you can't modify the asset in place anymore - modifying a subasset requires cloning the subasset first (and now your alias refers to different data) 
2. The aliases are practically two separate assets. They have different handles, and these handles can independently be dropped. So if you drop all the handles to the alias, you need to **reload** the asset just to get it back (even if another alias is still being used).

(Also feel free to leave a review for #15813, I'd love to move it forward haha)

Alainx277:
> 1. Cloning the arc means that you can't modify the asset in place anymore - modifying a subasset requires cloning the subasset first (and now your alias refers to different data)

I think that's an acceptable compromise, considering the lack of other feasible proposals.

> 2. The aliases are practically two separate assets. They have different handles, and these handles can independently be dropped. So if you drop all the handles to the alias, you need to **reload** the asset just to get it back (even if another alias is still being used).

Couldn't some additional bookkeeping be added so the asset system knows the asset is still available?

andriyDev:
@Alainx277 if you could make an issue, that would be great. Linking to this PR and then summarizing the ideas here would also be fantastic.

I also might have an idea on how to do asset aliases but I feel like I'm missing something so I'll tinker with it soon.

Igor-dvr:
Proposal mentioned in discord:

Getting an asset by its name is problematic because names are optional and can be duplicate. We can solve this by introducing duplicate number after an asset name when loading an asset.

Example:
Let's say we have 2 animations with the name "Walk" and 3 without a name.

First animation with the name "Walk" can be loaded like this:
assets.load("models.glb#Animation:Walk 0");
or shortcut:
assets.load("models.glb#Animation:Walk");
Number after a name is not a real asset id. Its number of a duplicate.

Second animation with the name "Walk" can be loaded like this:
assets.load("models.glb#Animation:Walk 1");

To load an asset that doesn't have a name we only specify duplicate number. This loads nameless asset 3:
assets.load("models.glb#Animation: 2"); 

To simplify migration we can add an option in the loader settings that indicates if we want to use old id based approach or new approach with a name + duplicate number.

ChristopherBiscardi:
There is a newer version of this PR in #23505 so I'm going to close this one to de-duplicate

PR #16533, opened 2024-11-27
Fix CAS toggle broken by retained render world

### LINKED ISSUE #16531: Toggling sharpening on then off results in a massively oversharpened image despite sharpening being disabled.
## Bevy version

2a66bf0909992b280420442ecda42a38d37de326 with https://github.com/bevyengine/bevy/pull/16530 so sharpening actually works

## What you did

Run the `anti_aliasing` example, then toggle sharpening on and off. Wait a bit and it starts glitching out.

## What went wrong

Image ends up being oversharpened to the point of looking broken, with the oversharpening artifacts changing over time as well despite sharpening being disabled.

Image

kristoff3r:
# Objective

Fixes #16531

I also added change detection when creating the pipeline, which technically isn't needed but it felt weird leaving it as is.

## Solution

Remove the pipeline if CAS is disabled. The uniform was already being removed, which caused flickering / weirdness.

## Testing

Tested the anti_alias example by toggling CAS a bunch on/off.

### DIFF
--- a/crates/bevy_core_pipeline/src/contrast_adaptive_sharpening/mod.rs
+++ b/crates/bevy_core_pipeline/src/contrast_adaptive_sharpening/mod.rs
@@ -249,6 +249,10 @@ fn prepare_cas_pipelines(
     >,
     mut removals: RemovedComponents<CasUniform>,
 ) {
+    for entity in removals.read() {
+        commands.entity(entity).remove::<ViewCasPipeline>();
+    }
+
     for (entity, view, denoise_cas) in &views {
         let pipeline_id = pipelines.specialize(
             &pipeline_cache,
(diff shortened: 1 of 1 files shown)


Elabajaba (approved):
Tested it and it works.

mockersf:
Would that work correctly if the `CasUniform` component is removed and re-added in the same frame?

kristoff3r:
> Would that work correctly if the `CasUniform` component is removed and re-added in the same frame?

No, that would break. But that can't happen with `ExtractComponentPlugin` afaik, so you'd have to add a custom extraction or render system just to break it.

I think it just works if we run removals first, because in the "added then removed" case the query would be empty.

alice-i-cecile (approved):
Oh hey, this is a situation where @cart's "required components work on removal" would have solved that bug. Anyways, this is fine, merging.

PR #16537, opened 2024-11-28
Add `Commands::run_schedule`

### LINKED ISSUE #16495: Add `Commands::run_schedule`
Systems (especially observers) should be able to queue the execution of schedules (which are essentially a special kind of exclusive system). It's currently possible to queue the execution of a system through `Commands::run_system`, but there's no equivalent `Commands::run_schedule` counterpart for `World::run_schedule`.

Off the top of my head, I can't think of any technical blockers.

bushrat011899:
# Objective

- Fixes #16495

## Solution

- Added `Commands::run_schedule`, which internally calls `World::try_run_schedule`, panicking on any issues.

## Testing

- Added doctest
- Ran CI

## Showcase

Instead of writing:

```rust
commands.queue(|world: &mut World| world.run_schedule(FooSchedule));
```

You can now write:

```rust
commands.run_schedule(FooSchedule);
```

### DIFF
--- a/crates/bevy_ecs/src/system/commands/mod.rs
+++ b/crates/bevy_ecs/src/system/commands/mod.rs
@@ -1017,7 +1017,7 @@ impl<'w, 's> Commands<'w, 's> {
     pub fn run_schedule(&mut self, label: impl ScheduleLabel) {
         self.queue(|world: &mut World| {
             if let Err(error) = world.try_run_schedule(label) {
-                error!("Failed to run schedule: {error}");
+                panic!("Failed to run schedule: {error}");
             }
         });
     }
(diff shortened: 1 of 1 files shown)


alice-i-cecile (approved):
Not just unwrapping I see :P I think this is good default behavior, but we'll need to come back and change this when we tackle command error handling.

bushrat011899:
> Not just unwrapping I see :P I think this is good default behavior, but we'll need to come back and change this when we tackle command error handling.

I was half tempted to do `// TODO: commands need a coherent error handling policy` but I thought I'd just sneak in my preferred solution!

BenjaminBrienen (approved):

mockersf (changes requested):
I would prefer this to panic until there's a way to react in code to errors

Running a schedule that doesn't exist doesn't seem like something that should be ignored and you would still expect everything working

BenjaminBrienen:
Does `error!` not set the right expectation?

mockersf:
no, you can't react to it in code

mockersf (approved):

PR #16545, opened 2024-11-28
feat: support for clip children on windows

### LINKED ISSUE #16544: Add WS_CLIPCHILDREN support for Windows in bevy_winit

The winit version used in Bevy now supports WS_CLIPCHILDREN for Windows. This is particularly useful when trying to render a Wry WebView for which the HTML content has a transparent background because it will prevent the Wry container from clipping the rendering surface used by Bevy. This is the last "obstacle" I see in order to fully/properly use HTML as a user interface layer for Bevy.

It could be to just add `winit_window_attributes = winit_window_attributes.with_clip_children(false);` [here](https://github.com/bevyengine/bevy/blob/main/crates/bevy_winit/src/winit_windows.rs#L128).
It could also be a more "configurable" approach using Window settings. Not sure considering this is Windows only.

I don't really see any other alternatives.

## Additional context

N/A

nicholasc:
# Objective

Support the parametrization of the WS_CLIPCHILDREN style on Windows.

Fixes #16544

## Solution

Added a window configuration in bevy_winit to control the usage of the WS_CLIPCHILDREN style.

## Testing

I did. I was able to create a Wry Webview with a transparent HTML document and was also able to see my Bevy scene behind the webview elements.

 I don't believe so. I assume the option is extensively tested within winit itself.

Test repositiory [here](https://github.com/nicholasc/bevy_wry_test). Bevy's path will need to be updated in the Cargo.toml

This is a Windows specific issue. Should be tested accordingly.


BenjaminBrienen (approved):

crates/bevy_window/src/window.rs
```diff
@@ -277,6 +277,15 @@ pub struct Window {
     ///
     /// - Only supported on Windows.
     pub skip_taskbar: bool,
+    /// Sets whether the window should clip its child windows to its client area.
+    ///
+    /// If `true`, child windows will be clipped to the client area of this window.
+    /// If `false`, child windows can be drawn outside the client area.
```
jf908, line 283: This wording here isn't entirely clear to me but maybe that's just because I'm familiar with Windows terminology. I have a suggestion but it's not blocking.

```suggestion
    /// Sets whether the window should draw over its child windows.
    ///
    /// If `true`, the window excludes drawing over areas obscured by child windows.
    /// If `false`, the window can draw over child windows.
```

jf908 (approved):
Thought I would review this because of an interest in webview stuff. Tested and it works, example here: https://github.com/jf908/bevy-webview-test.

Small doc suggestion but nothing blocking, lgtm.

P.S. It took me ages to find the Microsoft documentation which explains this option, it's here: https://learn.microsoft.com/en-us/windows/win32/winmsg/window-features#clipping

crates/bevy_window/src/window.rs
nicholasc: Your explanation seems more concise than mine but same as you, I was having trouble with with the wording.
I'll go ahead and accept the changes.

BenjaminBrienen (approved):

BenjaminBrienen (approved):

PR #16546, opened 2024-11-28
Clarify inheritance behavior of required components

### LINKED ISSUE #16494: Misleading docs on duplicate required component constructors
The [required components documentation](https://dev-docs.bevyengine.org/bevy/ecs/component/trait.Component.html#required-components) describes the algorithm for resolving multiple constructors for the same component in a required components tree as follows:

> In general, this shouldn’t happen often, but when it does the algorithm is simple and predictable:
>
> 1. Use all of the constructors (including default constructors) directly defined in the spawned component’s require list
> 2. In the order the requires are defined in #[require()], recursively visit the require list of each of the components in the list (this is a Depth First Search). When a constructor is found, it will only be used if one has not already been found.

However, this is not what actually happens. For example, consider this case:

```rust
#[derive(Component)]
#[require(B)]
struct A;

#[derive(Component, Default)]
#[require(C, D(|| D(1)))]
struct B;

#[derive(Component, Default)]
#[require(D(|| D(2)))]
struct C;

#[derive(Component)]
struct D(u8);
```

Now suppose we spawn `A`. Following the steps described in the docs to determine which constructor will be used for `D`:

1. The "spawned component" is `A`, which does not have a constructor for `D`, so go to step 2.
2. A DFS will visit the components in the order `A B C D(2) D(1)`. So when `D(2)` is visited, it will be used because `D` hasn't been found yet; and when `D(1)` is visited, it will not be used because `D` has been found already.

This suggests that `D(2)` will be used as the constructor. But that's incorrect: `D(1)` is actually used.

alice-i-cecile:
Co-authored by: @BenjaminBrienen

# Objective

Fixes #16494. Closes #16539, which this replaces. Suggestions alone weren't enough, so now we have a new PR!

### DIFF
--- a/crates/bevy_ecs/src/component.rs
+++ b/crates/bevy_ecs/src/component.rs
@@ -226,9 +226,9 @@ use derive_more::derive::{Display, Error};
 /// assert_eq!(2, world.entity(id).get::<X>().unwrap().0);
 /// ```
 ///
-/// In general, this shouldn't happen often, but when it does the algorithm is simple and predictable:
-/// 1. Use all of the constructors (including default constructors) directly defined in the spawned component's require list
-/// 2. In the order the requires are defined in `#[require()]`, recursively visit the require list of each of the components in the list (this is a Depth First Search). When a constructor is found, it will only be used if one has not already been found.
+/// In general, this shouldn't happen often, but when it does the algorithm for choosing the constructor from the tree is simple and predictable:
+/// 1. A constructor from a direct #[requires()], if one exists, is selected with priority.
+/// 2. Otherwise, perform a Depth First Search on the tree of requirements and select the first one found.
 ///
 /// From a user perspective, just think about this as the following:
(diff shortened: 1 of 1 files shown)


crates/bevy_ecs/src/component.rs
```diff
@@ -226,9 +226,9 @@ use derive_more::derive::{Display, Error};
 /// assert_eq!(2, world.entity(id).get::<X>().unwrap().0);
 /// ```
 ///
-/// In general, this shouldn't happen often, but when it does the algorithm is simple and predictable:
-/// 1. Use all of the constructors (including default constructors) directly defined in the spawned component's require list
-/// 2. In the order the requires are defined in `#[require()]`, recursively visit the require list of each of the components in the list (this is a Depth First Search). When a constructor is found, it will only be used if one has not already been found.
+/// In general, this shouldn't happen often, but when it does the algorithm for choosing the constructor from the tree is simple and predictable:
+/// 1. A constructor from a direct #[requires()], if one exists, is selected with priority.
```
Jondolf, line 230:
```suggestion
/// 1. A constructor from a direct #[require()], if one exists, is selected with priority.
```

mweatherley (approved):
Definitely clearer :)

BenjaminBrienen:
Perfect

Issue #16541, opened 2024-11-28
iOS mobile example regressions?


extrawurst:
## Bevy version

3d72e08

## \[Optional\] Relevant system information

* cargo 1.82.0 (8f40fc59f 2024-08-21)
* xcode 16.1 (16B40)

## What you did

Just running the `bevy_mobile_example.xcodeproj`

## What went wrong

Console spamming of:

```
Monitor removed 14v3908#16784732192782
processing non `RedrawRequested` event after the main event loop: AboutToWait
Monitor removed 4v3083#13241384173572
processing non `RedrawRequested` event after the main event loop: AboutToWait
Monitor removed 14v3909#16789027160078
processing non `RedrawRequested` event after the main event loop: AboutToWait
Monitor removed 4v3084#13245679140868
processing non `RedrawRequested` event after the main event loop: AboutToWait
Monitor removed 14v3910#16793322127374
processing non `RedrawRequested` event after the main event loop: AboutToWait
Monitor removed 4v3085#13249974108164
processing non `RedrawRequested` event after the main event loop: AboutToWait
Monitor removed 14v3911#16797617094670
processing non `RedrawRequested` event after the main event loop: AboutToWait
Monitor removed 4v3086#13254269075460
processing non `RedrawRequested` event after the main event loop: AboutToWait
Monitor removed 14v3912#16801912061966
processing non `RedrawRequested` event after the main event loop: AboutToWait
Monitor removed 4v3087#13258564042756
processing non `RedrawRequested` event after the main event loop: AboutToWait
Monitor removed 14v3913#16806207029262
processing non `RedrawRequested` event after the main event loop: AboutToWait
Monitor removed 4v3088#13262859010052
```

Otherwise visually the example seems to run fine in the simulator. 

here is the log of the same example on bevy 14.2: https://gist.github.com/extrawurst/7cbbb5b1918b5876a93705b92d65e196

here is the entire log of the above example in 3d72e08:
https://gist.github.com/extrawurst/b91117297242d412091862db171e29ab

If you compare this to the 14.2 behaviour there are a few things very different:
1. we don't have timestamps anymore
2. instead of just the old friend `processing non `RedrawRequested` event after the main event loop: AboutToWait` we now seem to get this second log every frame: `Monitor removed 4v3088#13262859010052`
3. the startup produces an absolutely massive amount of additional logs if you compare both full logs

I wonder if this is all intended? The repeating `Monotitor removed` at the very least looks very wrong.

[...]

PR #16548, opened 2024-11-29
Reduce iOS cpu usage

mockersf:
# Objective

- Avoid recreating the monitor every loop (temp fix until it's done properly on winit side)
- Add a new `WinitSettings` preset for mobile that makes the winit loop wait more and recommend its usage

### DIFF
--- a/crates/bevy_winit/src/system.rs
+++ b/crates/bevy_winit/src/system.rs
@@ -158,6 +158,16 @@ pub fn create_monitors(
                 seen_monitors[idx] = true;
                 continue 'outer;
             }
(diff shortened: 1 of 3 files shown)


alice-i-cecile:
Prompted by #16541, but not a full fix. See that thread for more information.

alice-i-cecile (approved):
I think this is a good fix given the part of the release cycle we're in. There's definitely something fundamentally wrong with how winit / Bevy is interacting with iOS here though.

alice-i-cecile:
> rusticorn: That looks good to me it’s what I did locally as well 👍

That's a second approval. Merging :)

PR #16549, opened 2024-11-29
Use GitHub markdown alerts in README

doup:
# Objective

Use [GitHub's markdown alert](https://github.com/orgs/community/discussions/16925) in README, which looks like this:

> image

## Solution

Use `[!CAUTION]` alert type. I've considered also the `[!WARNING]` one but it's not very _bold_:

> image

## Testing

Checked in the GitHub markdown preview.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,9 +11,12 @@
 
 Bevy is a refreshingly simple data-driven game engine built in Rust. It is free and open-source forever!
 
-## WARNING
-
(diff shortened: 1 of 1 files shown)


doup:
Uhmm… I'm not sure if this is going to work on the Cargo page, probably not. So, maybe it's better to leave it as it was. 🙃

Loweredgames:
> Uhmm… I'm not sure if this is going to work on the Cargo page, probably not. So, maybe it's better to leave it as it was. 🙃

no

BenjaminBrienen:
I think that we should either go ahead and use this feature, or we should block this on adding this feature to crates.io and docs.rs.

BenjaminBrienen (approved):

BenjaminBrienen:
Maybe `markdownlint` too
image

mockersf:
blocked on support on docs.rs, crates.io and probably lib.rs

doup:
I'm closing this, I didn't think through every other place where the README is shown… and I don't want to open issues in these repos for a little nice to have.

PR #16555, opened 2024-11-29
bump to 0.16.0-dev

atlv24:
# Objective

- bump versions to 0.16.0-dev

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "bevy"
-version = "0.15.0-dev"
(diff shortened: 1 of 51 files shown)


Loweredgames:
to be reopened?

> this should be done by CI: https://github.com/bevyengine/bevy/actions/runs/12091144173
> 
> I triggered the job an hour ago but it failed due to a permission problem since GitHub switched Bevy to a company org.
> 
> I would prefer if the job was fixed then relaunched to check that it's working

PR #13764, opened 2024-06-09
add helper method Query::with_entity

Nathan-Fenner:
# Objective

While I'm writing systems, I frequently find myself suddenly wanting to have access to the `Entity` that I'm iterating over with a `Query<&MyComp>` - so I have to refactor to `Query<(Entity, &MyComp)>`. This means changing the parameter type and other iteration sites over the same query, even if those sites don't use the `Entity`.

## Solution

This PR adds a convenience method, `.with_entity()` analogous to e.g. `.enumerate()` on `Iterator` - it returns a `QueryLens` over the same data, but with `Entity` added to the data selector (in other words, it turns a `Query<D, F>` into a `QueryLens<(Entity, D), F>`).

## Testing

- There's a trivial doctest that should make sure it compiles; this is essentially just a convenience wrapper around `transmute_lens_filtered` but without the need to specify generic types as a caller

---

## Changelog

- Added `Query::with_entity()` convenience function that returns a `QueryLens` over the same results but also including the `Entity` id

PR #16560, opened 2024-11-30
add `.include_entity()` helper on query iterators

Nathan-Fenner:
# Objective

Adds a helper method to the  `QueryIter` struct called `.include_entity()`, which automatically adds `(Entity, _)` around whatever is currently being iterated. Since the underlying `Entity` is always tracked anyway, this can be done "for free" at zero additional runtime cost.

This helper makes it more ergonomic to evolve systems over time. For example, as we write a complex system, we encounter a case where we need the `Entity` that we're iterating over:

```rs
fn example(q_data: Query<(&mut Player, &Transform, &Health)>) {
   for (player, transform, _health) in q_data.iter() {
       do_thing1(player, transform);
    }

   for (mut player, transform, health) in q_data.iter_mut() {
       do_thing2(player, transform, health);
    }

   for (mut player, transform, _health) in q_data.iter() {
       do_thing3(todo!() /* need entity! */, player, transform);
    }
}
```

Despite the `Entity` already being tracked at runtime, we then need to go back and update multiple lines of code to request it:

```rs
fn example(q_data: Query<(Entity, &mut Player, &Transform, &Health)>) {   // changed
   for (_entity, player, transform, _health) in q_data.iter() {           // changed
       do_thing1(player, transform);
    }

   for (_entity, mut player, transform, _health) in q_data.iter_mut() {   // changed
       do_thing2(player, transform, health);
    }

   for (entity, mut player, transform, _health) in q_data.iter_mut() {    // changed
       do_thing3(entity, player, transform);                              // changed
    }
}
```

Of the 5 changed lines, only the last 2 are essential to the change we actually want to make. If we use `include_entity()` instead, then we get the much-more-explicit delta:

```rs
fn example(q_data: Query<(&mut Player, &Transform, &Health)>) {
   for (player, transform, _health) in q_data.iter() {
       do_thing1(player, transform);
    }

   for (mut player, transform, _health) in q_data.iter_mut() {
       do_thing2(player, transform, health);
    }

   for (entity, (mut player, transform, _health)) in q_data.iter_mut().include_entity() { // changed
       do_thing3(entity, player, transform);                                              // changed
    }
}
```

## Why not `(Entity, D)`? Why a new `IncludeEntity<D>` type?

The underlying query state for the query `(A, B)` is `(A::State, B::State)`. This means that for `(Entity, D)` the underlying state type is `((), D::State))`, not `D::State`. Despite looking very similar, these types aren't guaranteed to have the same layout, and so we cannot transmute between them.

The `IncludeEntity<_>` query type is not intended for typical use by game developers, but is made public so that other utilities/libraries can also rely on it for transmute safety.

## Testing

- [ ] Needs tests before merging

### DIFF
--- a/crates/bevy_ecs/src/query/fetch.rs
+++ b/crates/bevy_ecs/src/query/fetch.rs
@@ -305,12 +305,18 @@ pub unsafe trait QueryData: WorldQuery {
     ) -> Self::Item<'w>;
 }
 
+unsafe impl<D: QueryData> QueryData for crate::query::IncludeEntity<D> {
+    type ReadOnly = crate::query::IncludeEntity<D::ReadOnly>;
+}
+
 /// A [`QueryData`] that is read only.
 ///
 /// # Safety
 ///
 /// This must only be implemented for read-only [`QueryData`]'s.
 pub unsafe trait ReadOnlyQueryData: QueryData<ReadOnly = Self> {}
+
+unsafe impl<D: ReadOnlyQueryData> ReadOnlyQueryData for crate::query::IncludeEntity<D> {}
 
 /// The item type returned when a [`WorldQuery`] is iterated over
 pub type QueryItem<'w, Q> = <Q as QueryData>::Item<'w>;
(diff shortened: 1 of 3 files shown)


Nathan-Fenner:
I consider this a better alternative to a previous PR I opened at #13764, which might be independently useful but is not as general

hymm:
Definitely prefer this pr to the previous one as the call to transmute_lens in that pr allocates.

I lean towards not wanting this functionality as I prefer the simplicity of having only one way of adding entity to a query. But I don't feel too strongly about it if someone really wants it.

Bleachfuel:
I think it's nicer if u add the last example u have here in the description in the docs. Because the docs don't really show the purpose now of `include_entity`

ItsDoot:
I like the idea! Happy to review + approve once CI is green.

Issue #15059, opened 2024-09-05
Swap most uses of allow in lints to expect


alice-i-cecile:
## What problem does this solve or what need does it fill?

With the advent of [Rust 1.81](https://blog.rust-lang.org/2024/09/05/Rust-1.81.0.html), we can now distinguish between lints that we haven't gotten around to fixing and those where the lint is wrong.

## What solution would you like?

Most localized lints in Bevy should be `expect`: the only exception that comes to mind is `missing_docs`.

Leave the `Cargo.toml` lints set as allow, as funny as it would be to set a crate-level `expect(type_complexity)`.

alice-i-cecile:
I'd also like to turn on the new `allow_attributes_without_reason` lint as part of this.

BD103:
I'll take a stab at this!

BD103:
If someone wants to work on this, I got a lot of progress done in #15118. Feel free to adopt it and follow [my advice here](https://github.com/bevyengine/bevy/pull/15118?scrlybrkr=9612f9dd#issuecomment-2348935105) for finishing it.

alice-i-cecile:
My suggestion for those tackling this is to handle this one crate at a time: that will ensure it's easier to complete and review smaller bits of work.

richchurcher:
Copying BD103's instructions/workflow for ease of reference. Also note that crate-level `expect`/`missing_docs` should be fixed in an upcoming Rust version.

> I'm going to put this PR up for adoption. I no longer have the energy to work on it, and there are still a _lot_ of crates left. If you want to adopt this, my workflow looked like this:
> 
> 1. Run `cargo clippy --workspace --message-format=short`
> 2. `cd` into the first crate that raises warnings.
> 3. Run `cargo watch -c -x clippy` until warnings stop appearing.
> 4. Run `cargo clippy --no-default-features` and `cargo clippy --all-features` for good measure.
> 5. Repeat step 1 until there are no more warning.

richchurcher:
Keeping a list here for my own reference: these were not covered by the previous PR.

- [ ] bevy_core_pipeline
- [ ] bevy_dev_tools
- [ ] bevy_dylib
- [ ] bevy_gilrs
- [ ] bevy_gizmos
- [ ] bevy_gltf
- [ ] ~bevy_input~ _looks pretty well-covered_
- [ ] bevy_internal
- [ ] bevy_pbr
- [ ] bevy_picking
- [ ] bevy_render
- [ ] bevy_sprite
[...]

PR #16561, opened 2024-11-30
Rust 1.83, allow -> expect (missing_docs)

bas-ie:
# Objective

We were waiting for 1.83 to address most of these, due to a bug with `missing_docs` and `expect`. Relates to, but does not entirely complete, #15059.

## Solution

- Upgrade to 1.83
- Switch `allow(missing_docs)` to `expect(missing_docs)`
- Remove a few now-unused `allow`s along the way, or convert to `expect`


richchurcher:
Let's see what the CI has to say now :eyes:

crates/bevy_reflect/src/lib.rs
```diff
@@ -1,6 +1,4 @@
-// FIXME(15321): solve CI failures, then replace with `#![expect()]`.
-#![allow(missing_docs, reason = "Not all docs are written yet, see #3492.")]
-// `rustdoc_internals` is needed for `#[doc(fake_variadics)]`
+#![expect(missing_docs, reason = "Not all docs are written yet, see #3492.")]
 #![allow(internal_features)]
```
BenjaminBrienen, line 4: Why did you remove this? It might have warned about being unused because it is only used with that feature enabled. Maybe put this under a cfg_attr instead?

BenjaminBrienen (changes requested):

BenjaminBrienen:
I think you can use expect there if you use cfg_attr.

crates/bevy_reflect/src/lib.rs
richchurcher: I see! Is https://github.com/bevyengine/bevy/pull/16561/commits/229ef7f5317b8a44c8019798b8a5620ff128c27a what you were thinking?

crates/bevy_render/src/lib.rs
```diff
@@ -1,8 +1,5 @@
-// FIXME(15321): solve CI failures, then replace with `#![expect()]`.
-#![allow(missing_docs, reason = "Not all docs are written yet, see #3492.")]
-#![allow(unsafe_code)]
-// `rustdoc_internals` is needed for `#[doc(fake_variadics)]`
-#![allow(internal_features)]
```
BenjaminBrienen, line 5: Do the same thing here :)

BenjaminBrienen (changes requested):

crates/bevy_render/src/lib.rs
BenjaminBrienen, line 5: Do the same thing here :)

```suggestion
#![cfg_attr(any(docsrs, docsrs_dep), expect(internal_features))]
```

BenjaminBrienen (changes requested):

BenjaminBrienen (approved):

BenjaminBrienen (approved):

examples/2d/bloom_2d.rs
```diff
@@ -65,6 +65,7 @@ fn setup(
             left: Val::Px(12.0),
             ..default()
         },
+        TextColor(Color::srgb(5., 5., 5.)),
```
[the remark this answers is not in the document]
bas-ie: :thinking: why _did_ I do that?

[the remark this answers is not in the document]
bas-ie: I have absolutely no idea, must've snuck in while I was playing around with something! Ty for the spot.

BenjaminBrienen (approved):

PR #16567, opened 2024-11-30
Rename `ArgList::push` methods to `with` and add new `push` methods which take `&mut self`

atornity:
# Objective

The `ArgList::push` family of methods consume `self` and return a new `ArgList` which means they can't be used with `&mut ArgList` references. 

```rust
fn foo(args: &mut ArgList) {
    args.push_owned(47_i32); // doesn't work :(
}
```

It's typical for `push` methods on other existing types to take `&mut self`.

## Solution

Renamed the existing push methods to `with_arg`, `with_ref` etc and added new `push` methods which take `&mut self`.

## Migration Guide

Uses of the `ArgList::push` methods should be replaced with the `with` counterpart.

<details>

| old | new |
| --- | --- |
| push_arg | with_arg |
| push_ref | with_ref |
| push_mut | with_mut |
| push_owned | with_owned | 
| push_boxed | with_boxed |

</details>


BenjaminBrienen (approved):

greeble-dev (changes requested):
There's a lot of comments that now seem incorrect because they're still using push  - one example below. Can these be updated?

https://github.com/bevyengine/bevy/blob/498d31d4a2af49782625b2a7a57c7d336ed6e5af/crates/bevy_reflect/src/func/args/list.rs#L19-L29

MrGVSV (approved):

PR #16575, opened 2024-11-30
Move required components doc to type doc

SpecificProtagonist:
# Objective

Make documentation of a component's required components more visible by moving it to the type's docs

## Solution

Change `#[require]` from a derive macro helper to an attribute macro.

Disadvantages:
- this silences any unused code warnings on the component, as it is used by the macro!
- need to import `require` if not using the ecs prelude (I have not included this in the migration guilde as Rust tooling already suggests the fix)

---

## Showcase
Documentation of Camera

### DIFF
--- a/crates/bevy_core_pipeline/src/motion_blur/mod.rs
+++ b/crates/bevy_core_pipeline/src/motion_blur/mod.rs
@@ -11,7 +11,10 @@ use crate::{
 use bevy_app::{App, Plugin};
 use bevy_asset::{load_internal_asset, Handle};
 use bevy_ecs::{
-    bundle::Bundle, component::Component, query::With, reflect::ReflectComponent,
+    bundle::Bundle,
+    component::{require, Component},
+    query::With,
+    reflect::ReflectComponent,
     schedule::IntoSystemConfigs,
 };
 use bevy_reflect::{std_traits::ReflectDefault, Reflect};
--- a/crates/bevy_core_pipeline/src/taa/mod.rs
+++ b/crates/bevy_core_pipeline/src/taa/mod.rs
@@ -10,7 +10,7 @@ use bevy_app::{App, Plugin};
 use bevy_asset::{load_internal_asset, Handle};
 use bevy_core::FrameCount;
 use bevy_ecs::{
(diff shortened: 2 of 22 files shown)


alice-i-cecile:
Neat, I had no idea this would work. I prefer this: I think it's more discoverable.

Jondolf:
This should probably have a heading or a horizontal rule separating the section from the rest of the docs. Currenly it blends in too much, and sometimes looks like it's a part of some other text even though it should be separate.

docs

With a horizontal rule (just using `\n\n---\n\nRequired Components: {paths}`), it looks like this:

docs

I think this might be a clear enough separation between "type docs" and "docs automatically embedded by Bevy".

We could also use a heading, which would look like this:

docs

The benefit here is that it shows up in the Sections outline on the left, so it's quicker to navigate to the Required Components section for components with long docs. You can also link to the heading, which could be nice so that you can send a direct link if people are asking about a specific component's required components for whatever reason.

The (minor) downside of the heading is that it *must* be a top-level h1 heading, because otherwise it could be shown "beneath" other sections in the outline, which would be wrong. Being an h1 heading makes it quite prominent, for better or for worse.

I think I'm mildly in favor of the heading because of it being shown in the outline and being linkable, but the horizontal rule could work fine too.

Jondolf:
<details>
<summary>By the way, if anyone is wondering if these auto-generated docs show up in VSCode when hover over a type, they do!</summary>

VSCode hover docs
</details>

BenjaminBrienen (approved):

Jondolf:
Oh wow, another nice side effect here is that this makes autocomplete for `require` work!

autocomplete!!!

Previously you got no suggestions and had to remember if it's `require` or `requires`. I saw a few people complaining about that on Discord earlier

Jondolf (approved):

BD103:
I'd like to nominate this for release notes. I think this feature is going to be very valuable to users migrating to required components, especially when it comes to visibility in libraries and plugins.

alice-i-cecile:
Thank you to everyone involved with the authoring or reviewing of this PR! This work is relatively important and needs release notes! Head over to https://github.com/bevyengine/bevy-website/issues/1962 if you'd like to help out.

PR #16577, opened 2024-11-30
Update breakout to use Required Components

Mclilzee:
# Objective

This PR update breakout to use the new 0.15 Required Component feature instead of the Bundle.
Add more information in the comment about where to find more info about Required Components.

## Solution

Replace `#[derive(Bundle)]` with a new Wall component and `#[require()]` Macro to include the other components.

## Testing

Tested with `cargo test` as well tested the game manually with `cargo run --example breakout` It looks to me that it works like it used to before the changes. Tested on Arch Linux, Wayland

### DIFF
--- a/examples/games/breakout.rs
+++ b/examples/games/breakout.rs
@@ -102,9 +102,7 @@ struct CollisionSound(Handle<AudioSource>);
 #[derive(Component, Default)]
 struct Collider;
 
-// This is a collection of the components that define a "Wall" in our game.
-// from bevy version 0.15 onward the main way has shifted from using a Bundle, to Required Components
-// Allowing you to compose their functionality, See https://bevyengine.org/news/bevy-0-15/#required-components
+// This is a collection of the components that define a "Wall" in our game
 #[derive(Component)]
 #[require(Sprite, Transform, Collider)]
 struct Wall;
--- a/examples/games/breakout.rs
+++ b/examples/games/breakout.rs
@@ -98,7 +98,7 @@ struct Brick;
 #[derive(Resource, Deref)]
 struct CollisionSound(Handle<AudioSource>);
 
-// Default is required to be part of the Required Componetns used with the Wall Component bellow
+// Default is required to be part of the Required Components used with the Wall Component bellow
 #[derive(Component, Default)]
 struct Collider;
 
--- a/examples/games/breakout.rs
+++ b/examples/games/breakout.rs
@@ -98,7 +98,7 @@ struct Brick;
 #[derive(Resource, Deref)]
 struct CollisionSound(Handle<AudioSource>);
 
-// Default is required to be part of the Required Components used with the Wall Component bellow
+// Default must be implemented to define this as a required component for the Wall component below
 #[derive(Component, Default)]
 struct Collider;
 
@@ -150,8 +150,7 @@ impl WallLocation {
 impl Wall {
     // This "builder method" allows us to reuse logic across our wall entities,
     // making our code easier to read and less prone to bugs when we change the logic
-    // notice the use of Sprite and Transform alongside Wall even tho they are part of our Required Component
-    // that is because Required Component by default uses Components' default values, but can be overwritten with manual insertion like bellow
+    // Notice the use of Sprite and Transform alongside Wall, overwriting the default values defined for the required components
     fn new(location: WallLocation) -> (Wall, Sprite, Transform) {
         (
             Wall,
--- a/examples/games/breakout.rs
+++ b/examples/games/breakout.rs
@@ -89,10 +89,6 @@ struct Ball;
 #[derive(Component, Deref, DerefMut)]
 struct Velocity(Vec2);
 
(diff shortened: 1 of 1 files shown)


examples/games/breakout.rs
```diff
@@ -101,15 +98,16 @@ struct Brick;
...
-// This bundle is a collection of the components that define a "wall" in our game
-#[derive(Bundle)]
-struct WallBundle {
-    // You can nest bundles inside of other bundles like this
-    // Allowing you to compose their functionality
-    sprite: Sprite,
-    transform: Transform,
-    collider: Collider,
-}
+// Default is required to be part of the Required Componetns used with the Wall Component bellow
```
RCoder01, line 101:
```suggestion
// Default is required to be part of the Required Components used with the Wall Component bellow
```

BenjaminBrienen (approved):

Mclilzee:
@BenjaminBrienen Hey thanks for the review. I noticed the `Waiting Author` Tag, is there anything else that needs to be done?

BenjaminBrienen:
I think there was an error in CI, maybe a merge conflicts. Looks good now :)

alice-i-cecile (approved):
Nice!

examples/games/breakout.rs
```diff
@@ -101,15 +98,16 @@ struct Brick;
...
-    transform: Transform,
-    collider: Collider,
-}
+// Default is required to be part of the Required Components used with the Wall Component bellow
+#[derive(Component, Default)]
+struct Collider;
+
+// This is a collection of the components that define a "Wall" in our game.
+// from bevy version 0.15 onward the main way has shifted from using a Bundle, to Required Components
+// Allowing you to compose their functionality, See https://bevyengine.org/news/bevy-0-15/#required-components
```
Jondolf, line 107: I find it strange to comment on Bevy versions or link to release notes. Examples should demonstrate *current* functionality and APIs, not describe the history of how the API or conventions have changed.

examples/games/breakout.rs
```diff
@@ -101,15 +98,16 @@ struct Brick;
...
-// This bundle is a collection of the components that define a "wall" in our game
-#[derive(Bundle)]
-struct WallBundle {
-    // You can nest bundles inside of other bundles like this
-    // Allowing you to compose their functionality
-    sprite: Sprite,
-    transform: Transform,
-    collider: Collider,
-}
+// Default is required to be part of the Required Components used with the Wall Component bellow
```
Jondolf, line 101: - bellow -> below
- "Default is required" could be worder better to not overlap with required components
- I don't generally see a reason to capitalize common features like Component or Required Components
```suggestion
// Default must be implemented to define this as a required component for the Wall component below
```

examples/games/breakout.rs
```diff
@@ -149,13 +147,16 @@ impl WallLocation {
...
-impl WallBundle {
+impl Wall {
     // This "builder method" allows us to reuse logic across our wall entities,
     // making our code easier to read and less prone to bugs when we change the logic
-    fn new(location: WallLocation) -> WallBundle {
-        WallBundle {
-            sprite: Sprite::from_color(WALL_COLOR, Vec2::ONE),
-            transform: Transform {
+    // notice the use of Sprite and Transform alongside Wall even tho they are part of our Required Component
+    // that is because Required Component by default uses Components' default values, but can be overwritten with manual insertion like bellow
```
Jondolf, line 154:
```suggestion
    // Notice the use of Sprite and Transform alongside Wall, overwriting the default values defined for the required components
```

examples/games/breakout.rs
Mclilzee: I agree with that. In my mind I wanted to mention the shift from bundles since they are not being deprecated and still in the system, so wanted to highlight hem as "another way" of combining components, but moving forward they are not the main way.

I will push an update and, as you said, keeping it to the point of how the current API works.

Mclilzee:
Any thought about removing `Transform` from the required components for the `Wall` since `Sprite` itself requires `Transform`? 

If we remove it, it is potentially less clear that the Wall will have `Transform` for those unfamiliar with `Sprite` and wouldn't know that `Sprite` itself requires `Transform`. Just something I thought about while cleaning up.

alice-i-cecile:
It's more idiomatic to remove Transform, but I'd call out this behavior in a comment since this is learning material.

Mclilzee:
@alice-i-cecile hey, this should be ready, I believe. I re-requested a review from the reviewers to take another look after implementing their changes, but they seem to be busy. I didn't want this to take a long time, as it's one of the core examples of bevy, I think we can move to final-review phase, I'm still new to all of this so learning by observing how things go.

As for the Transform we talked about above, we can open a different PR / Issue to discuss it as for now this PR is straight forward translation from Bundle to Required Component, doesn't need changes in implementation.

alice-i-cecile:
Agreed. This is definitely a straight improvement, so I'm merging as is.

Issue #14275, opened 2024-07-11
Our API suggests to the user that panicking should be the default


janhohenheim:
## What problem does this solve or what need does it fill?

_In spirit of https://github.com/bevyengine/bevy/issues/12660_
_Adapted from https://github.com/bevyengine/bevy/pull/14268_

I believe it is not controversial to say that when an API offers two similar functions with similar names, the shorter will seem like the default. As such, I believe many people will instinctively gravitate to `single` and `single_mut` over `get_single` and `get_single_mut`. This means we are subtly pushing users to prefer the implicitly panicking version over the fallible one. This is bad, as it leads to games that may run well on the developers machine but then panic in an edge case. 
It is currently understood in the wider Rust community that panics should be an exceptional case, a nuclear option for when recovery is impossible. 
We should do our best to make it as easy to do the right thing by default (See [Falling Into The Pit of Success](https://blog.codinghorror.com/falling-into-the-pit-of-success/)). This means that it should be easier to handle an error by propagating it with `?` to a logger or similar than to panic. Our current API does the opposite.

This is an issue for the following APIs:
- `Query::single`
- `Query::get`
- `World::resource`
- `World::non_send_resource` (although eventually this will likely be on App)
- All of the other non-deferred entity-component-manipulating APIs listed in https://github.com/bevyengine/bevy/issues/14231

## What solution would you like?

Deprecate and remove panicking variants. Improve ergonomics by adding macros for early returns. Using a macro for `get_single` as an example:

```rust
fn do_stuff(primary_window: Query<&Window, With<PrimaryWindow>>) {
    let window = get_single!(primary_window);
    // Do stuff with the window. If there's not exactly one, we have already returned.
}
```

Which expands to:

```rust
fn do_stuff(primary_window: Query<&Window, With<PrimaryWindow>>) {
    match primary_window.get_single() {
        Ok(item) => item,
        Err => return default(),
    };
    // Do stuff with the window. If there's not exactly one, we have already returned.
}
```

Note that returning `default()` will allow the macro to work with systems that return an `Option` and get `pipe`d into error handling methods.

Similar macros are already being used by some users, as indicated by https://github.com/tbillington/bevy_best_practices?tab=readme-ov-file#getter-macros

Paraphrasing @alice-i-cecile:

> I'm also coming around on the macro approach: I'd rather this be a Rust feature where we can just ? in function that return (), but failing that this is better than the endless let else return pattern.
>
[...]

Issue #12660, opened 2024-03-22
`unwrap` considered harmful


james7132:
## What problem does this solve or what need does it fill?
Unwrap should generally be avoided. We already encourage contributors to avoid the use of `Option::unwrap`, yet crashes from panic are still one of the most common categories of issues we're seeing now.

## What solution would you like?
 
 * Enable the clippy `unwrap_used` lint, and add `allow-unwrap-in-tests = true` to Clippy.toml.
 * Replace actual expected crashes with `expect` or unwrap_or_else with a panic where possible, otherwise properly gracefully handle the error.
 * Preferably avoid panicking in the renderer entirely, instead failing gracefully where possible by logging a warning and not rendering anything or using blatantly obvious errors (i.e. Valve's missing texture shader).

## What alternative(s) have you considered?
Leaving it as is. Keep crashing from unwraps.

alice-i-cecile:
See https://github.com/bevyengine/bevy/issues/10166 for a major culprit.

SludgePhD:
Another API design issue related to this is `Query::single` / `Query::get_single` – here, the shorter method name that users are most likely to go for will `.unwrap()` implicitly. IMO `Query::single` is almost never the right choice and just making it have the semantics of `get_single` would be better.

alice-i-cecile:
`single/get_single` used to work like that: users at the time found it tedious to constantly check if their player or camera had despawned!

It was also moved to be consistent with `.resource / .get_resource`, which while it still panics is much less prone to unexpected failure, as removing resources is much more rare than despawning entities.

That said, I am personally allergic to using `.single` in my own serious projects due to frustrations with panics in edge cases and during refactors, and return early on errors.

bushrat011899:
Perhaps instead of `get_x` and `x`, we could use `x` and `x_unchecked` instead? This makes the panicking API clearly distinguishable from the non-panicing, while also encouraging the idiomatic path.

Regardless, I agree with this change. `unwrap` is a code smell within a library.

james7132:
`x_unchecked` makes my mind jump immediately to `unsafe`, which we shouldn't be encouraging unless it's paramount that users squeeze every last ounce of performance out.

alice-i-cecile:
I think that `x_unchecked` (by whatever name) isn't worth adding. It's more API to maintain, and not actually much nicer than just another unwrap call.

benfrankel:
[...]

Issue #11562, opened 2024-01-27
Better error handling for systems


musjj:
## What problem does this solve or what need does it fill?

Handling `Result<T, E>` inside systems currently feels a bit awkward. The few options we have right now:

- Call `.unwrap()` everywhere. This feels very unidiomatic in Rust.
- Write custom log messages everywhere. This feels very boilerplaty for what could've been a single `?`.
- Have your system return `Result<T, E>`, then `.pipe()` it into [`bevy::utils::{warn, error, ...}`](https://github.com/bevyengine/bevy/blob/main/examples/ecs/system_piping.rs).

I'm liking the third option the most. But the problem is that the log messages does not include any traceback information whatsoever, such as the name of the system where the error originated from. This makes it very hard to debug.

## What solution would you like?

The usability of `bevy::utils::{warn, error, ...}` should be improved to make it easier to work with.

Or maybe there should be a built-in support for systems that returns `Result<T, E>`.

nicopap:
Another option is to use [`bevy_mod_sysfail`](https://crates.io/crates/bevy_mod_sysfail) (disclaimer, I am the author)

nicopap:
This has been discussed extensively in the past. I'm surprised you are actually the first to open an issue regarding error handling in bevy systems.

The best you are likely to get for now are the pipe error handling systems and `bevy_mod_sysfail`.

Regarding the traceback in the pipe error handling systems, there is an open issue: #8638

musjj:
Thanks for the crate! I tried it out and the log messages now displays the module the error originated from, which is way better than the default `.pipe()`. But it still does not log the actual name of the system. Is there a way to configure this?

nicopap:
A future release will automatically add the system name to the log message. It's going to be published early next week, see https://github.com/nicopap/bevy_mod_sysfail/pull/5

nicopap:
I'm closing this as a Duplicate of #10874 since it's really just asking for the same thing.

Issue #10874, opened 2023-12-05
Add ControlFlow::Continue(())  handling for systems


dimvoly:
## What problem does this solve or what need does it fill?

Many a time, I'd run a query and need to exit early. E.g. if some `Component` is missing or uninitialized and there's no use running the system further. The query may issue out a `Result<R, E>` or I may use a sub-routine that returns a `ControlFlow<()>`.

I would like to be able to use the `?` operator, as [described here](https://doc.rust-lang.org/stable/std/ops/enum.ControlFlow.html) to return early from a system.

## What solution would you like?

Allow for systems to have this in their signature: 

```rust
fn system() -> std::ops::ControlFlow<()> {

    // Then you call a sub-routine like so:
    some_sub_routine_that_can_return_early()?;

    ...
}
```

## What alternative(s) have you considered?

Using a macro to reduce the verbosity of returning early:

```rust
/// Unwrap the Continue or return.
///
#[macro_export]
macro_rules! continue_or_return {
    ( $e:expr ) => {
        match $e {
            std::ops::ControlFlow::Continue(x) => x,
            std::ops::ControlFlow::Break(_) => return,
        }
    };
}
pub use continue_or_return;
```

And use like so:
```rust
fn system() {

    // Then you call a sub-routine like so:
    continue_or_return!(some_sub_routine_that_can_return_early());

    ...
}
```

## Additional context

I'm on bevy 0.10.1, haven't looked in the latest version to see if this is a problem.
Also I don't know the inner workings of bevy, and if this may be impossible for some internal reason.

alice-i-cecile:
So, your suggestion is a variant of system piping.

I would really like to be able to use the ? operator here, but I don't think that as written it's possible. We need a blanket implementation over all systems: only those which both accept and return `()` can be stored in a final schedule: system piping effectively "welds" systems together.

I would want to peek at the implementation again, but I suspect that using a `ControlFlow` return type here would require specialization. Prove me wrong though!

Perhaps we could work upstream in Rust itself to support the use of the `?` operator in functions that return `()`.

nicopap:
This has been discussed heavily. Bevy developers seems uninterested in adding this feature. However, I made a crate to make it possible: https://lib.rs/crates/bevy_mod_sysfail

dimvoly:
Thanks for the suggestions, cool to see a crate's been put together too.

I hadn't thought about system piping that is possible too as a workaround.

[...]

PR #16589, opened 2024-12-01
Fallible systems

NthTensor:
# Objective

Error handling in bevy is hard. See for reference https://github.com/bevyengine/bevy/issues/11562, https://github.com/bevyengine/bevy/issues/10874 and https://github.com/bevyengine/bevy/issues/12660. The goal of this PR is to make it better, by allowing users to optionally return `Result` from systems as outlined by Cart in <https://github.com/bevyengine/bevy/issues/14275#issuecomment-2223708314>.

## Solution

This PR introduces a new `ScheuleSystem` type to represent systems that can be added to schedules. Instances of this type contain either an infallible `BoxedSystem<(), ()>` or a fallible `BoxedSystem<(), Result>`. `ScheuleSystem` implements `System<In = (), Out = Result>` and replaces all uses of `BoxedSystem` in schedules. The async executor now receives a result after executing a system, which for infallible systems is always `Ok(())`. Currently it ignores this result, but more useful error handling could also be implemented.

Aliases for `Error` and `Result` have been added to the `bevy_ecs` prelude, as well as const `OK` which new users may find more friendly than `Ok(())`.

## Testing

- Currently there are not actual semantics changes that really require new tests, but I added a basic one just to make sure we don't break stuff in the future.
- The behavior of existing systems is totally unchanged, including logging.
- All of the existing systems tests pass, and I have not noticed anything strange while playing with the examples

## Showcase

The following minimal example prints "hello world" once, then completes.

```rust
use bevy::prelude::*;

fn main() {
    App::new().add_systems(Update, hello_world_system).run();
}

fn hello_world_system() -> Result {
    println!("hello world");
    Err("string")?;
    println!("goodbye world");
    OK
}
```

## Migration Guide

This change should be pretty much non-breaking, except for users who have implemented their own custom executors. Those users should use `ScheduleSystem` in place of `BoxedSystem<(), ()>` and import the `System` trait where needed. They can choose to do whatever they wish with the result.

## Current Work

+ [x] Fix tests & doc comments
+ [x] Write more tests
+ [x] Add examples
+ [X] Draft release notes

## Draft Release Notes

As of this release, systems can now return results.

First a bit of background: Bevy has hisotrically expected systems to return the empty type `()`. While this makes sense in the context of the ecs, it's at odds with how error handling is typically done in rust: returning `Result::Error` to indicate failure, and using the short-circuiting `?` operator to propagate that error up the call stack to where it can be properly handled. Users of functional languages will tell you this is called "monadic error handling".

Not being able to return `Results` from systems left bevy users with a quandry. They could add custom error handling logic to every system, or manually pipe every system into an error handler, or perhaps sidestep the issue with some combination of fallible assignents, logging, macros, and early returns. Often, users would just litter their systems with unwraps and possible panics.

While any one of these approaches might be fine for a particular user, each of them has their own drawbacks, and none makes good use of the language. Serious issues could also arrise when two different crates used by the same project made different choices about error handling.

Now, by returning results, systems can defer error handling to the application itself. It looks like this:

```rust
// Previous, handling internally
app.add_systems(my_system)
fn my_system(window: Query<&Window>) {
   let Ok(window) = query.get_single() else {
       return;
   };
   // ... do something to the window here
}

// Previous, handling externally
app.add_systems(my_system.pipe(my_error_handler))
fn my_system(window: Query<&Window>) -> Result<(), impl Error> {
   let window = query.get_single()?;
   // ... do something to the window here
   Ok(())
}

// Previous, panicking
app.add_systems(my_system)
fn my_system(window: Query<&Window>) {
   let window = query.single();
   // ... do something to the window here
}

// Now 
app.add_systems(my_system)
fn my_system(window: Query<&Window>) -> Result {
    let window = query.get_single()?;
    // ... do something to the window here
    Ok(())
}
```

There are currently some limitations. Systems must either return `()` or `Result<(), Box<dyn Error + Send + Sync + 'static>>`, with no in-between. Results are also ignored by default, and though implementing a custom handler is possible, it involves writing your own custom ecs executor (which is *not* recomended).

Systems should return errors when they cannot perform their normal behavior. In turn, errors returned to the executor while running the schedule will (eventually) be treated as unexpected. Users and library authors should prefer to return errors for anything that disrupts the normal expected behavior of a system, and should only handle expected cases internally.

We have big plans for improving error handling further:
+ Allowing users to change the error handling logic of the default executors.
+ Adding source tracking and optional backtraces to errors.
+ Possibly adding tracing-levels (Error/Warn/Info/Debug/Trace) to errors.
+ Generally making the default error logging more helpful and inteligent.
+ Adding monadic system combininators for fallible systems.
+ Possibly removing all panicking variants from our api.


bushrat011899:
Even without the actual error handling benefits this provides, just having a more blessed way to use `?` in systems will be _really_ nice. I know we can just pipe the `Result` with current systems, but this will hide that bit of extra boilerplate. This also pairs nicely with making more APIs return `Result` instead of `Option`, and also makes panicking variants less important (possibly even removable TBH).

NthTensor:
I have been informed tests are failing due to some arcane inference issue with `Never` aka `!`.

Putting this here for posterity, as I'm having to wrangle with my rust compiler to get it to update.

```rust
pub struct Never;

impl<F, Marker> IntoSystemConfigs<(Never, Marker)> for F
where
    F: IntoSystem<(), !, Marker>,
{
    fn into_configs(self) -> SystemConfigs {
        let boxed_system = Box::new(IntoSystem::into_system(self));
        SystemConfigs::new_system(ScheduleSystem::Infallible(boxed_system))
    }
}
```

bushrat011899 (approved):
I agree this probably needs an example, but I like the approach. Opens up the possibility of having error handlers in the future, which would resolve the _to-panic or not to-panic_ debate entirely. This also lays the groundwork for how fallibility in `Command`s could work. Really nice work!

bushrat011899:
It was mentioned in Discord, but I'll include it here for posterity: with fallible systems getting first-class treatment, there may be room to consider removing the panicking variants of certain functions (e.g., `Query::get_entity` and `Query::entity`), since the choice of behaviour could be controlled by a system error handler. This would be a large DX win, since the "proper" methods would get the shorter names, and it'd reduce the API surface area.

NthTensor:
> there may be room to consider removing the panicking variants of certain function

That's in line with the third point Cart proposed in https://github.com/bevyengine/bevy/issues/14275#issuecomment-2223708314. He indicated then that it was important to land all the related changes in a single release cycle, and I agree. This PR provides his (1), what (2) and (3) look like is up to @alice-i-cecile and the other designated ecs experts.

crates/bevy_ecs/src/result.rs
```diff
@@ -0,0 +1,10 @@
+//! Contains error and result helpers for use in fallible systems.
+
+/// A dynamic error type for use in fallible systems.
+pub type Error = Box<dyn core::error::Error + Send + Sync + 'static>;
+
+/// A result type for use in fallible systems.
+pub type Result<T = (), E = Error> = core::result::Result<T, E>;
```
tychedelia, line 7: Since `IntoSystemConfigs` doesn't provide generic params, I'm not sure these type defaults make sense, unless there's a fast follow to support generic errors.

tychedelia (approved):
Amazing to see how straightforward this is, all things considered. Very excited!

crates/bevy_ecs/src/result.rs
NthTensor: Agreed, however I want to have these in the bevy prelude. If we put a `Result` without the generics in the prelude, it will prevent the normal result type from being used. By only adding defaults with an alias, I'm hoping we will be able to write systems that use the default `Result` and functions that use `Result<T, E>` in the same module without issue.

teohhanhui:
> as well as const `OK` which new users may find more friendly than `Ok(())`.

Why? This just makes the code more jarring compared to the rest of the Rust ecosystem, and more cognitive load to switch between returning no results vs. returning some value.

It'd make sense if it's something useful like https://docs.rs/anyhow/latest/anyhow/fn.Ok.html

NthTensor:
> This just makes the code more jarring compared to the rest of the Rust ecosystem

In this I am trying to defer to my understanding of Cart's preferences. He uses a const in the linked issue, and I believe has expressed that `Ok(())` is sort of confusing and cumbersome. No strong preference here from me really.

crates/bevy_ecs/src/result.rs
teohhanhui: There's a good precedent here: https://docs.rs/anyhow/latest/anyhow/type.Result.html

> `anyhow::Result` may be used with one or two type parameters.

bushrat011899:
> It'd make sense if it's something useful like https://docs.rs/anyhow/latest/anyhow/fn.Ok.html

Functionally this is identical to `anyhow`'s `Ok` function. The difference is since in Bevy a `ScheduleSystem` can _only_ output an `()`, the function can instead just be a constant.

It's a problem with Rust's syntax at the end of the day. With the addition of the `Try` trait, perhaps we'll get an automatic wrapping as a bit of syntactic sugar, but that's up to the language team at the end of the day.

alice-i-cecile:
@NthTensor let me know when you're done the outstanding tasks and take this out of draft. I'll do a final pass and then merge + create follow-up issues for this.

alice-i-cecile:
> as well as const OK which new users may find more friendly than Ok(()).

Can you please cut this from this PR? I kinda like it, but it's pointlessly controversial here :)

crates/bevy_ecs/src/lib.rs
```diff
@@ -1,5 +1,7 @@
 // FIXME(11590): remove this once the lint is fixed
 #![allow(unsafe_op_in_unsafe_fn)]
+// TODO: remove once Edition 2024 is released
+#![allow(dependency_on_unit_never_type_fallback)]
```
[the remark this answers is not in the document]
bushrat011899: I actually prefer having the explicit allow with a `TODO`. Once 2024 releases, we will want to go back over this and potentially implement the relevant traits for systems that output `!`. If we don't, users who try to add similar `todo!` closures to their apps will fail to compile.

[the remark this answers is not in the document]
NthTensor: I agree, and even if I was overruled, I'd rather not make this change as part of this PR. It's really unrelated and due only to an annoying (and imo mistaken) choice from the rust lang team.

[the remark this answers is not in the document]
chescock: > Once 2024 releases, we will want to go back over this and potentially implement the relevant traits for systems that output `!`.

You could implement those traits now, right?  The 2024 edition will make those impls more useful, but you can create them before then so that you don't have to scramble after the release.  (Not suggesting that we should do that in this PR!  Just noting that we don't have to wait for the edition.)  

> I'd rather not make this change as part of this PR

Sorry, I wasn't trying to make more work for you!  Your message on Discord made it seem like you wanted to resolve the warning but couldn't, and I was just trying to offer a way to do so.  :)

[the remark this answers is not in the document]
bushrat011899: > You could implement those traits now, right?

Sadly not! The `!` type isn't stabilized until Edition 2024, so we can't make the "proper" fix until then.

NthTensor:
@alice-i-cecile requesting removal of the `M-Needs-Release-Note` label.

crates/bevy_ecs/src/lib.rs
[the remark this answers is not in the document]
chescock: Huh, I didn't see that in the edition guide. ... Oh, I see, it's even worse than that: the plan is to stabilize it sometime *after* the edition.  

... wait, then what are we going to do when the new edition is out?  The code with the warnings won't compile on Rust 2024 without impls for `!`, but we can't add those impls until `!` is stabilized at some later date.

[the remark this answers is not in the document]
NthTensor: Exactly. This is why I am a tad frustrated with the rust team. Theoretically this should not be an issue until bevy itself decides to upgrade, since each crate gets to decide its edition individually.

crates/bevy_ecs/src/system/schedule_system.rs
```diff
@@ -0,0 +1,183 @@
...
+    component::{ComponentId, Tick},
+    query::Access,
+    result::Result,
+    schedule::InternedSystemSet,
+    system::{input::SystemIn, BoxedSystem, System},
+    world::{unsafe_world_cell::UnsafeWorldCell, DeferredWorld, World},
+};
+
+/// A type which wraps and unifies the different sorts of systems that can be added to a schedule.
+pub enum ScheduleSystem {
```
hymm, line 15: I’m not a fan of this approach using an enum as this is introducing a lot of branching. I’d rather see all systems returning a result and figuring out a way of wrapping systems that return ‘()’ to return Ok.

NthTensor: I mostly agree. Some considerations:
+ With the enum, I was thinking perhaps the compiler would know which systems are infallible and would eliminate the handler code for them.
+ When writing this, I thought maybe we might eventually want to introduce more variants.

I have since found https://github.com/bevyengine/bevy/pull/8705 and it seems to have a much more streamlined version using adapter systems. I could probably just adopt that. One issue with that version is that it mangles the origin of tracing logs, which is why it was closed. I do not believe the enum does.

NthTensor: If you're concerned about branching performance, we should probably just benchmark to see how bad of a regression it really is.

crates/bevy_ecs/src/schedule/executor/mod.rs
```diff
@@ -140,18 +140,16 @@ mod __rust_begin_short_backtrace {
...
 
     /// # Safety
     /// See `System::run_unsafe`.
     #[inline(never)]
-    pub(super) unsafe fn run_unsafe(
-        system: &mut dyn System<In = (), Out = ()>,
-        world: UnsafeWorldCell,
-    ) {
-        system.run_unsafe((), world);
+    pub(super) unsafe fn run_unsafe(system: &mut ScheduleSystem, world: UnsafeWorldCell) {
```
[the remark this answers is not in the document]
NthTensor: Yep they should be, good catch.

crates/bevy_ecs/src/lib.rs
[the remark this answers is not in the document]
chescock: I don't think I follow.  What do you think the rust team should have done differently?  This looks like a standard case of code that will fail to compile on the next edition causing a warning on the current one.  It failed CI because that builds with `-Dwarnings`, but if you ignore the warning then it still builds with the old semantics.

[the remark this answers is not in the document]
bushrat011899: The big issue as I see it is there'll be a period of time where the only way to accept a function that returns `!` as a parameter is to accept _all_ return types. So in this PR, we have no way of actually accepting a closure like `|| todo!()` as a system, since it returns an unnameable type in Edition 2024. Combined with the statement that they intend on back-porting this change to earlier editions makes things very complicated.

_If_ Edition 2024 launches with the never-type _and_ this change to inference, then it's fine IMO. But since it'll only launch with the inference change, we'll have this transitional period where closures are just worse.

alice-i-cecile:
> @alice-i-cecile requesting removal of the M-Needs-Release-Note label (added in pr description because we don't have a good place for it yet).

Just like migration guides, we should keep this label around even after they're written for searchability and tooling. We might wan to rename that to be more clear though 🤔

crates/bevy_ecs/src/system/schedule_system.rs
hymm: We should check against the single threaded executor perf. I wouldn't be surprised if it isn't measurable, but I think we should still go with the other method if we can, since we know we don't need this to be dynamically settable.

NthTensor: If it's not a regression, and no one else feels strongly about it, I'd prefer to merge as is and refine the implementation in tree as follow-up. It will be much easier to improve various aspects of this once we have something basic working.

alice-i-cecile: Yeah, I don't have strong feelings about the implementation here. Getting the source-logging right is the key element to me.

examples/ecs/fallible_systems.rs
```diff
@@ -0,0 +1,77 @@
...
+    ));
+
+    // Spawn a camera:
+    commands.spawn((
+        Camera3d::default(),
+        Transform::from_xyz(-2.0, 3.0, 5.0).looking_at(Vec3::ZERO, Vec3::Y),
+    ));
+
+    // Create a new sphere mesh:
+    let mut sphere_mesh = Sphere::new(1.0).mesh().ico(7)?;
```
bushrat011899, line 44: What a sight to behold. Once proper (user configurable) handlers are added in a follow-up this will be perfect. Bevy APIs can be simplified _and_ made more reliable without any loss in ergonomics (IMO). Adding `Ok(())` at the end of a system is a small price to pay that (hopefully) Rust will solve on its own (since the issue isn't specific to Bevy)

NthTensor: Yeah :) this example is really just a minimal placeholder. Once we have handlers hooked up, I intend to go through and update all the examples to use this style (where it makes sense).

BenjaminBrienen:
> > @alice-i-cecile requesting removal of the M-Needs-Release-Note label (added in pr description because we don't have a good place for it yet).
> 
> Just like migration guides, we should keep this label around even after they're written for searchability and tooling. We might wan to rename that to be more clear though 🤔 
> 

`M-#[require(ReleaseNote)]`

NthTensor:
Alright, I added the most basic tests and examples in the world. There will be more to do there when handlers are hooked up. Ready for review.

examples/ecs/fallible_systems.rs
```diff
@@ -0,0 +1,77 @@
...
+    /// Add sample points as children of the sphere:
+    for point in distribution.sample_iter(&mut rng).take(10000) {
+        sphere.with_child((
+            Mesh3d(point_mesh.clone()),
+            MeshMaterial3d(point_material.clone()),
+            Transform::from_translation(point),
+        ));
+    }
+
+    /// Indicate the system completed sucsessfully:
```
alice-i-cecile:
```suggestion
    /// Indicate the system completed successfully:
```

crates/bevy_ecs/src/schedule/config.rs
```diff
@@ -517,18 +518,41 @@ impl IntoSystemConfigs<()> for SystemConfigs {
...
 
 impl IntoSystemConfigs<()> for BoxedSystem<(), ()> {
     fn into_configs(self) -> SystemConfigs {
-        SystemConfigs::new_system(self)
+        SystemConfigs::new_system(ScheduleSystem::Infallible(self))
+    }
+}
+
+/// A marker type for a fallible system. Used internally.
+pub struct Fallible;
```
alice-i-cecile, line 541: Nit: This should probably be doc hidden :)

alice-i-cecile (approved):
@nth this is good to go once it's merge-conflict free. I do prefer the "everything is fallible" approach by WrongShoe, but that's easily left to a follow-up refactor. Let's get the ball rolling here.

Neo-Zhixing:
Might be a bit late on this discussion, but this particular implementation feels a bit intrusive to me. Can we instead implement IntoSystemConfigs on systems that returns `Result` so that those systems can be added to the schedule like a normal system?

alice-i-cecile:
Thank you to everyone involved with the authoring or reviewing of this PR! This work is relatively important and needs release notes! Head over to https://github.com/bevyengine/bevy-website/issues/1967 if you'd like to help out.

PR #14385, opened 2024-07-18
Update Access and FilteredAccess to use sorted vecs instead of FixedBitSet

cBournhonesque:
# Objective

As described in the relations RFC: https://github.com/james-j-obrien/rfcs/blob/minimal-fragmenting-relationships/rfcs/79-minimal-fragmenting-relationships.md#access-bitsets-and-component-sparsesets

The reasons while the access bitsets are efficient currently is because the ComponentIds are dense: they are incremented from 0 and should remain small.
With relations, a ComponentId will have some upper bits 1, so the amount of memory allocated in the FixedBitSet to represent the value would be non-trivial.

## Solution

One way to fix the issue would be to replace the FixedBitSets with sorted vectors.
The vectors should remain relatively small since queries usually don't involved hundreds of components.

These allow to do union, difference, intersection operations in O(1).
(however inserting a value is O(n))

## Testing

Ran the `busy_schedule` benchmarks 
```
group                                   main                                   pr
-----                                   ----                                   --
busy_systems/01x_entities_03_systems    1.00     27.7±3.44µs        ? ?/sec    1.02     28.2±5.25µs        ? ?/sec
busy_systems/01x_entities_06_systems    1.00     42.3±1.46µs        ? ?/sec    1.11     47.0±2.95µs        ? ?/sec
busy_systems/01x_entities_09_systems    1.08    67.3±13.90µs        ? ?/sec    1.00     62.1±1.42µs        ? ?/sec
busy_systems/01x_entities_12_systems    1.00     77.9±1.76µs        ? ?/sec    1.08    83.8±14.63µs        ? ?/sec
busy_systems/01x_entities_15_systems    1.00     94.2±1.92µs        ? ?/sec    1.08    101.8±5.94µs        ? ?/sec
busy_systems/02x_entities_03_systems    1.00     42.4±0.86µs        ? ?/sec    1.28    54.1±15.42µs        ? ?/sec
busy_systems/02x_entities_06_systems    1.00     74.0±3.57µs        ? ?/sec    1.15     85.0±5.44µs        ? ?/sec
busy_systems/02x_entities_09_systems    1.00    109.8±1.81µs        ? ?/sec    1.05    114.9±5.82µs        ? ?/sec
busy_systems/02x_entities_12_systems    1.00    142.3±1.42µs        ? ?/sec    1.11   157.5±27.83µs        ? ?/sec
[...]

PR #16591, opened 2024-12-01
Added stress test for large ecs worlds

Trashtalk217:
# Objective

We currently have no benchmarks for large worlds with many entities, components and systems.
Having a benchmark for a world with many components is especially useful for the performance improvements needed for relations. This is also a response to this [comment from cart](https://github.com/bevyengine/bevy/pull/14385#issuecomment-2311292546).

> I'd like both a small bevy_ecs-scoped executor benchmark that generates thousands of components used by hundreds of systems.

## Solution

I use dynamic components and components to construct a benchmark with 2000 components, 4000 systems, and 10000 entities.

## Some notes

- ~I use a lot of random entities, which creates unpredictable performance, I should use a seeded PRNG.~
- Not entirely sure if everything is ran concurrently currently. And there are many conflicts, meaning there's probably a lot of first-come-first-serve going on. Not entirely sure if these benchmarks are very reproducible.
- Maybe add some more safety comments
- Also component_reads_and_writes() is about to be deprecated #16339, but there's no other way to currently do what I'm trying to do.


chescock:
> * Also component_reads_and_writes() is about to be deprecated [Migrate `component_reads_and_writes` #16339](https://github.com/bevyengine/bevy/issues/16339), but there's no other way to currently do what I'm trying to do.

One option is to remember `access_components` yourself.  You could give `base_system` a `Local<Vec<ComponentId>>` param and build it with `LocalBuilder(access_components.collect())`, or you could give it an ordinary `&[ComponentId]` param and do `let access_components: Vec<_> = access_components.collect();` then `.build_system(move |query: Query<FilteredEntityMut>| base_system(&access_components, query))`.  

That might also help reduce noise when testing changes to the implementation of `Access`, since the performance of `component_reads_and_writes()` will change, but a real system wouldn't be calling it.

benches/benches/bevy_ecs/scheduling/schedule.rs
```diff
@@ -143,3 +156,109 @@ pub fn empty_schedule_run(criterion: &mut Criterion) {
...
+
+        // we assign this value to all the components we can write to
+        for component_id in filtered_entity.access().component_reads_and_writes().0 {
+            let ptr = filtered_entity.get_by_id(component_id).unwrap();
+            if filtered_entity.access().has_component_write(component_id) {
+                // SAFETY:
+                // We have exclusive access so the pointer is unique
+                // All components have a u8 layout
+                unsafe {
+                    let value = ptr.assert_unique().deref_mut::<u8>();
```
chescock, line 191: I think you can combine the `get_by_id()` call with the access check if you use `get_mut_by_id()`.  

```suggestion
            if let Some(ptr) = filtered_entity.get_mut_by_id(component_id) {
                // SAFETY:
                // All components have a u8 layout
                unsafe {
                    let value = ptr.with_type::<u8>();
```

Trashtalk217:
> > * Also component_reads_and_writes() is about to be deprecated [Migrate `component_reads_and_writes` #16339](https://github.com/bevyengine/bevy/issues/16339), but there's no other way to currently do what I'm trying to do.
> 
> One option is to remember `access_components` yourself. You could give `base_system` a `Local<Vec<ComponentId>>` param and build it with `LocalBuilder(access_components.collect())`, or you could give it an ordinary `&[ComponentId]` param and do `let access_components: Vec<_> = access_components.collect();` then `.build_system(move |query: Query<FilteredEntityMut>| base_system(&access_components, query))`.
> 
> That might also help reduce noise when testing changes to the implementation of `Access`, since the performance of `component_reads_and_writes()` will change, but a real system wouldn't be calling it.

I'm a fan of passing it as an argument, I'll do that. I'm hesitant to use `Local`s in this case because that's effectively adding another little bit of complexity (I believe Locals are Resources underneath, not entirely sure) which could potentially add to the variance of the benchmark.

chescock:
> I'm a fan of passing it as an argument, I'll do that. I'm hesitant to use `Local`s in this case because that's effectively adding another little bit of complexity (I believe Locals are Resources underneath, not entirely sure) which could potentially add to the variance of the benchmark.

I agree that using a closure is simpler!  But in case you're curious: Locals store their value [right in the param state](https://github.com/bevyengine/bevy/blob/c02696b6092308cc198ce3aee032079245d94df4/crates/bevy_ecs/src/system/system_param.rs#L1123), not in a Resource or anything.  So accessing them should be just a pointer offset from the system struct, same as a closure capture.

hymm:
It might be worth trying this as a stress test example since it's so noisy. That way we could bench using tracy and run it for longer to reduce the noise.

examples/stress_tests/many_components.rs
```diff
@@ -0,0 +1,194 @@
+//! Stress test for large ECS worlds.
+//!
+//! Running this example:
+//!
+//! ```
+//! cargo run --profile stress-test --example many_components <num_entities> <num_components> <num_systems>
```
nakedible, line 6:
```suggestion
//! cargo run --profile stress-test --example many_components [<num_entities>] [<num_components>] [<num_systems>]
```

Maybe also otherwise highlight that there are quite sane defaults, so values only need to be specified if testing something custom.

examples/stress_tests/many_components.rs
```diff
@@ -0,0 +1,194 @@
...
+    let world = app.world_mut();
+
+    // register a bunch of components
+    let component_ids: Vec<ComponentId> = (1..=num_components)
+        .map(|i| {
+            world.register_component_with_descriptor(
+                #[allow(unsafe_code)]
+                // SAFETY:
+                // we don't implement a drop function
+                // u32 is Sync and Send
```
nakedible, line 86:
```suggestion
                // u8 is Sync and Send
```

examples/stress_tests/many_components.rs
```diff
@@ -0,0 +1,194 @@
...
+}
+
+#[allow(missing_docs)]
+pub fn main() {
+    let default_num_entities = 50000;
+    let default_num_components = 1000;
+    let default_num_systems = 800;
+
+    // take input
+    let num_entities = std::env::args()
```
nakedible, line 161: Could we add some limit for number of updates, after which the stress test exits? Even though this isn't a benchmark, it would allow rudimentary benchmarking with `hyperfine` etc. Just looking at FPS values makes it quite hard to compare the impact of changes.

nakedible (approved):
Minor comments, fine to merge as is by me.

examples/stress_tests/many_components.rs
Trashtalk217: I'll think I'll pass on that, as other stress tests don't have them. Most benchmarking is done via tools like tracy.

nakedible (approved):

examples/stress_tests/many_components.rs
```diff
@@ -0,0 +1,196 @@
...
+        .add_plugins(DiagnosticsPlugin)
+        .add_plugins(LogPlugin::default())
+        .add_plugins(FrameTimeDiagnosticsPlugin)
+        .add_plugins(LogDiagnosticsPlugin::filtered(vec![DiagnosticPath::new(
+            "fps",
+        )]));
+    app.run();
+}
+
+#[allow(missing_docs)]
```
BD103, line 156:
```suggestion
#[expect(missing_docs)]
```

(Nit) I largely prefer `#[expect(...)]`, since it emits a warning when the lint goes away.

examples/stress_tests/many_components.rs
```diff
@@ -0,0 +1,196 @@
...
+        // See https://en.wikipedia.org/wiki/Faulhaber%27s_formula
+        // The time is takes to compute this depends on the number of entities and the values in
+        // each entity. This is to ensure that each system takes a different amount of time.
+        let mut total: Wrapping<u8> = Wrapping(0);
+        let mut exponent: u32 = 1;
+        for component_id in &access_components.0 {
+            // find the value of the component
+            let ptr = filtered_entity.get_by_id(*component_id).unwrap();
+
+            #[allow(unsafe_code)]
```
BD103, line 48:
```suggestion
            #[expect(unsafe_code)]
```

examples/stress_tests/many_components.rs
```diff
@@ -0,0 +1,196 @@
...
+                }
+                total += product;
+            }
+            exponent += 1;
+        }
+
+        // we assign this value to all the components we can write to
+        for component_id in &access_components.0 {
+            if let Some(ptr) = filtered_entity.get_mut_by_id(*component_id) {
+                #[allow(unsafe_code)]
```
BD103, line 65:
```suggestion
                #[expect(unsafe_code)]
```

BD103 (approved):
Looks good! Some small nits, but nothing blocking :)

examples/stress_tests/many_components.rs
```diff
@@ -0,0 +1,196 @@
...
+            "fps",
+        )]));
+    app.run();
+}
+
+#[allow(missing_docs)]
+pub fn main() {
+    let default_num_entities = 50000;
+    let default_num_components = 1000;
+    let default_num_systems = 800;
```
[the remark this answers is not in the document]
Trashtalk217: this is gonna sound dumb, but I prefer just normal variables because the caps are too shouty... ^_^

PR #16592, opened 2024-12-02
Update Contributor's Guide link in README.md

james-j-obrien:
# Objective

- Fix outdated link in README.md

Perhaps the link name should be updated as well since the article is no longer labelled as "Contributor's Guide"?
Either way better to avoid making users click twice.

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,7 +50,7 @@ Before contributing or participating in discussions with the community, you shou
 
 ### Contributing
 
-If you'd like to help build Bevy, check out the **[Contributor's Guide](https://github.com/bevyengine/bevy/blob/main/CONTRIBUTING.md)**.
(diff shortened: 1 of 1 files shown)


alice-i-cecile (commented):
We should also change the naming to match.

james-j-obrien:
The phrasing seems like it would also need to change as the title is just "Introduction".

Maybe something like:
If you'd like to help build Bevy, (check out our/you can find an) [introduction here](https://bevyengine.org/learn/contribute/introduction).

Most ways I try to phrase it sound clunky or vague, in CONTRIBUTING.md it's just referred to as "information on contributing".

BenjaminBrienen (approved):

BenjaminBrienen:
Approving as this is strictly an improvement, but those are good suggestions.

PR #16598, opened 2024-12-02
Moving `PlaybackMode` to `AudioPlayer` and defaulting to `PlaybackMode::Remove`

### LINKED ISSUE #12359: Consider making `PlaybackSettings::ONCE` not the default or removing it
## Motivation

Following up on a question posed during the "ecs-based audio" refactor in https://github.com/bevyengine/bevy/pull/8424#issuecomment-1604451931 which I don't think was ever really addressed.

> This begs the question: should `PlaybackMode::Once` even exist? Does it even make sense to retain audio-related components, if sound is no longer playing?

As well as this comment on `PlaybackMode`:

https://github.com/bevyengine/bevy/blob/f67b17d80d410045f0fb0af668521e2f91569159/crates/bevy_audio/src/audio.rs#L76-L79

Motivated by a recent Discord discussion about "how to replay a sound."

`PlaybackSettings` currently defaults to [`PlaybackMode::Once`](https://docs.rs/bevy/latest/bevy/audio/enum.PlaybackMode.html#variant.Once), meaning the audio will play once, and then the audio entity and its components are left alone and not despawned or removed.

That sounds convenient at first. After the `AudioSink` is created, it can be re-used, saving you the CPU cycles of spawning, despawning, and creating the sink. Or archetype-moves adding/removing components.

Unfortunately, it really can't be reused. Once the audio in the sink's queue is played, it is gone, and there is no way for a user to refill the sink and start playback again. The only way I know of to restart audio is to remove the `AudioSink` component from the entity -- `play_queued_audio_system` will create a new one. That trick doesn't work for playing a different sound with the same sink. See #11862.

So why does this mode exist, and why is the default behavior to leak audio entities into the world and never clean them up?

Make the default behavior `PlaybackMode::Remove`, and add an internal system that cleans up "orphaned" entities when audio components are removed and the entity is otherwise "empty."

Leave `PlaybackMode::Once` around in case it is useful to someone. Add docs explaining why it's sort of useless?

1. Make the default behavior `PlaybackMode::Despawn`. This is might be what users want most often. Leave `Once` in case it's useful to someone.
2. Same, but remove `PlaybackMode::Once`.
4. Keep current default behavior, remove the `TODO` comment.
4. Add some sort of machinery for reusing a sink by refilling it or filling it with different audio.
    
    I naively attempted to add `append` to the `AudioSinkPlayback` trait, but it required making the trait generic, and it seemed like a mess. It's probably possible, but that route would lead to a pretty inconvenient API that seems counter to the "ecs-based audio." Using change detection on `Handle<Source>` seems like it would be pretty funky too.

seivan:
# Objective
Fix: https://github.com/bevyengine/bevy/issues/12359

Wanted to make it clear what happens when audio is being played, and the source is related to the `PlaybackMode`.
You're either playing sound effects that are temporary in lifetime, or constant repeated music.
Some sound effects related to existing entity while others are children of said entities. 
I wanted the API to actually reflect that, instead of being split over two components. 
It requires explicit input about `PlaybackMode `

This break existing code in two places: 
* When `AudioPlayer::new()` is used it changes default mode to `Remove`, current default is `Once`.
* Tuple struct syntax requires as second argument. 

## Solution

* Changing the default `PlaybackMode` to `PlaybackMode::Remove`
* Adding `PlaybackMode` to `AudioPlayer`
* Adding convenience methods to `AudioPlayer` to create with different `PlaybackMode`s

## Testing

> - Did you test these changes? If so, how?

* Ran each audio example. 
* Ran tests

> - Are there any parts that need more testing?

Performance? Since `AudioPlayer` now holds `PlaybackMode` there can have been a difference in internal optimizations that I might have missed. 

> - How can other people (reviewers) test your changes? Is there anything specific they need to know?

Run audio examples. 

> - If relevant, what platforms did you test these changes on, and are there any important ones you can't test?

Tested only on `macOS`, I don't think it's relevant for other platforms unless there are performance implications with the new default `remove` for `PlaybackMode`. 

---

### Changed the default behaviour to be `remove`

```rust
    pub fn new(source: Handle<AudioSource>) -> Self {
        Self(source, PlaybackMode::default())
    }
```

### Introducing new functions to create `AudioPlayer`

```rust
    /// Creates a new [`AudioPlayer`] that plays the sound once.
    pub fn with_once(source: Handle<AudioSource>) -> Self {
        Self(source, PlaybackMode::Once)
    }

    /// Creates a new [`AudioPlayer`] that loops the sound forever.
    pub fn with_loop(source: Handle<AudioSource>) -> Self {
        Self(source, PlaybackMode::Loop)
    }

    /// Creates a new [`AudioPlayer`] that despawns the entity when the sound finishes playing.
    pub fn with_despawn(source: Handle<AudioSource>) -> Self {
        Self(source, PlaybackMode::Despawn)
    }

    /// Creates a new [`AudioPlayer`] that removes the audio component from the entity when the sound finishes playing.
    pub fn with_remove(source: Handle<AudioSource>) -> Self {
        Self(source, PlaybackMode::Remove)
    }
```

#### Examples: 

Old
```rust
    commands.spawn(AudioPlayer(
        asset_server.load("sounds/Windless Slopes.ogg"),
    ));
```
New
```rust
    commands.spawn(AudioPlayer::with_once(
        asset_server.load("sounds/Windless Slopes.ogg"),
    ));
```

Old
```rust
        commands.spawn((
            AudioPlayer(pitch_assets.add(Pitch::new(frequency.0, Duration::new(1, 0)))),
            PlaybackSettings::DESPAWN,
        ));
```
New
```rust
        commands.spawn(AudioPlayer(
            pitch_assets.add(Pitch::new(frequency.0, Duration::new(1, 0))),
            PlaybackMode::Despawn,
        ));
```

### DIFF
--- a/crates/bevy_audio/src/audio.rs
+++ b/crates/bevy_audio/src/audio.rs
@@ -33,7 +33,7 @@ impl Volume {
 }
 
 /// The way Bevy manages the sound playback.
-#[derive(Debug, Clone, Copy, Reflect)]
+#[derive(Debug, Clone, Copy, Reflect, Default)]
 pub enum PlaybackMode {
     /// Play the sound once. Do nothing when it ends.
     Once,
@@ -42,6 +42,7 @@ pub enum PlaybackMode {
     /// Despawn the entity and its children when the sound finishes playing.
     Despawn,
     /// Remove the audio components from the entity, when the sound finishes playing.
+    #[default]
     Remove,
 }
 
@@ -53,8 +54,6 @@ pub enum PlaybackMode {
 #[derive(Component, Clone, Copy, Debug, Reflect)]
 #[reflect(Default, Component, Debug)]
 pub struct PlaybackSettings {
-    /// The desired playback behavior.
-    pub mode: PlaybackMode,
     /// Volume to play at.
     pub volume: Volume,
     /// Speed to play at.
@@ -77,40 +76,17 @@ pub struct PlaybackSettings {
 
 impl Default for PlaybackSettings {
     fn default() -> Self {
-        // TODO: what should the default be: ONCE/DESPAWN/REMOVE?
-        Self::ONCE
+        Self {
+            volume: Volume(1.0),
+            speed: 1.0,
+            paused: false,
+            spatial: false,
+            spatial_scale: None,
+        }
     }
 }
 
 impl PlaybackSettings {
-    /// Will play the associated audio source once.
-    pub const ONCE: PlaybackSettings = PlaybackSettings {
-        mode: PlaybackMode::Once,
-        volume: Volume(1.0),
-        speed: 1.0,
-        paused: false,
-        spatial: false,
-        spatial_scale: None,
-    };
-
-    /// Will play the associated audio source in a loop.
-    pub const LOOP: PlaybackSettings = PlaybackSettings {
-        mode: PlaybackMode::Loop,
-        ..PlaybackSettings::ONCE
-    };
-
-    /// Will play the associated audio source once and despawn the entity afterwards.
-    pub const DESPAWN: PlaybackSettings = PlaybackSettings {
-        mode: PlaybackMode::Despawn,
-        ..PlaybackSettings::ONCE
-    };
-
-    /// Will play the associated audio source once and remove the audio components afterwards.
-    pub const REMOVE: PlaybackSettings = PlaybackSettings {
-        mode: PlaybackMode::Remove,
-        ..PlaybackSettings::ONCE
-    };
-
     /// Helper to start in a paused state.
     pub const fn paused(mut self) -> Self {
         self.paused = true;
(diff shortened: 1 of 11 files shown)


rparrett:
Just noting that this would close #12359. I haven't looked closely at the PR yet.

seivan:
> Just noting that this would close #12359. I haven't looked closely at the PR yet.

I actually wanted to remove `::Once` but wasn't sure if it had unintended consequences.

BenjaminBrienen (approved):

rparrett:
For what it's worth, I have changed my opinion since writing up #12359 and I don't personally that automatic despawning ore removal by default is the right choice. The current behavior seems like what I would expect when I spawn any other entity.

I agree that `PlaybackSettings` is really clunky right now, but we should probably come to a consensus on a design if we're reworking it to minimize churn.

PR #16603, opened 2024-12-02
Add alpha mode implementation to shader_material_2d 

Leinnan:
## Objective 

Bevy 0.15 introduced new method in `Material2d` trait- `alpha_mode`. Before that when new material was created it had alpha blending, now it does not. 

## Solution 

While I am okay with it, it could be useful to add the new trait method implementation to one of the samples so users are more aware of it.

### DIFF
--- a/examples/shader/shader_material_2d.rs
+++ b/examples/shader/shader_material_2d.rs
@@ -4,7 +4,7 @@ use bevy::{
     prelude::*,
     reflect::TypePath,
     render::render_resource::{AsBindGroup, ShaderRef},
-    sprite::{Material2d, Material2dPlugin},
+    sprite::{AlphaMode2d, Material2d, Material2dPlugin},
 };
 
 /// This example uses a shader source file from the assets subdirectory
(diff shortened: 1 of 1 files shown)


IceSentry:
This is already covered in more details in the mesh2d_alpha_mode example.

If we had an alpha mode to this example it should be mask, not blend because it doesn't need blending. It should also point to the above example.

Leinnan:
@IceSentry I would argue that is not covered for the Custom materials part. In 0.14 if you implemented a custom material that was by default a blend one, since 0.15 it changed. Other than that, I think it could be `AlphaMode2d::Mask(0.5)`, I will change it later.

examples/shader/shader_material_2d.rs
```diff
@@ -57,4 +57,8 @@ impl Material2d for CustomMaterial {
     fn fragment_shader() -> ShaderRef {
         SHADER_ASSET_PATH.into()
     }
+
+    fn alpha_mode(&self) -> AlphaMode2d {
+        AlphaMode2d::Blend
```
IceSentry, line 62:
```suggestion
        AlphaMode2d::Mask(0.5)
```

IceSentry (approved):

BenjaminBrienen (approved):

PR #16611, opened 2024-12-02
add index_range for TextureAtlas

useyourfeelings:
# Objective

To simplify index updating for TextureAtlas when playing animation.  

## Solution

Add index_range and advance_index() for TextureAtlas.  

So  

```rust
// init
let animation_indices = AnimationIndices { first: 1, last: 6 };

TextureAtlas {
    layout: texture_atlas_layout,
    index: animation_indices.first,
},

// update
atlas.index = if atlas.index == indices.last {
    indices.first
} else {
    atlas.index + 1
};
```

will become

```rust
// init
TextureAtlas {
    layout: texture_atlas_layout,
    index: 1,
    index_range: (1, 6)
},

// update
atlas.advance_index();
```

## Migration Guide

1. Provide `index_range: (my_first_index, my_last_index)` for all your `TextureAtlas` initializations.   
```rust
TextureAtlas {
    layout: texture_atlas_layout,
    index: 1,
    index_range: (1, 6), // add this as defualt index range
},
```
2. Instead of manually updating index like 
```rust
atlas.index = if atlas.index == indices.last {
    indices.first
} else {
    atlas.index + 1
};
```
just call  
```rust
atlas.advance_index(); // index += 1
```

3. Remove unnecessary code like

```rust
let animation_indices = AnimationIndices { first: 1, last: 6 }; // delete this
```

## Testing

Tested all examples related to TextureAtlas.

### DIFF
--- a/crates/bevy_sprite/src/texture_atlas.rs
+++ b/crates/bevy_sprite/src/texture_atlas.rs
@@ -171,7 +171,7 @@ pub struct TextureAtlas {
     pub layout: Handle<TextureAtlasLayout>,
     /// Texture atlas section index
     pub index: usize,
-    /// Texture atlas default index range (first, last)
+    /// The index of the first and last images in the atlas
     pub index_range: (usize, usize),
 }
 
--- a/crates/bevy_sprite/src/texture_atlas.rs
+++ b/crates/bevy_sprite/src/texture_atlas.rs
@@ -32,6 +32,7 @@ impl TextureAtlasSources {
         Some(TextureAtlas {
             layout,
             index: self.texture_index(texture)?,
+            index_range: (0, 0),
         })
     }
 
@@ -170,6 +171,8 @@ pub struct TextureAtlas {
     pub layout: Handle<TextureAtlasLayout>,
     /// Texture atlas section index
     pub index: usize,
+    /// Texture atlas default index range (first, last)
+    pub index_range: (usize, usize),
 }
 
 impl TextureAtlas {
@@ -178,6 +181,14 @@ impl TextureAtlas {
         let atlas = texture_atlases.get(&self.layout)?;
         atlas.textures.get(self.index).copied()
     }
+
+    pub fn advance_index(&mut self) -> &mut Self {
+        self.index += 1;
+        if self.index > self.index_range.1 {
+            self.index = self.index_range.0;
+        }
+        self
+    }
 }
 
 impl From<Handle<TextureAtlasLayout>> for TextureAtlas {
@@ -185,6 +196,7 @@ impl From<Handle<TextureAtlasLayout>> for TextureAtlas {
         Self {
             layout: texture_atlas,
             index: 0,
+            index_range: (0, 0),
         }
     }
 }
--- a/examples/2d/sprite_animation.rs
+++ b/examples/2d/sprite_animation.rs
@@ -64,10 +64,10 @@ fn execute_animations(time: Res<Time>, mut query: Query<(&mut AnimationConfig, &
             if let Some(atlas) = &mut sprite.texture_atlas {
                 if atlas.index == config.last_sprite_index {
                     // ...and it IS the last frame, then we move back to the first frame and stop.
-                    atlas.index = config.first_sprite_index;
+                    atlas.advance_index();
                 } else {
                     // ...and it is NOT the last frame, then we move to the next frame...
-                    atlas.index += 1;
+                    atlas.advance_index();
                     // ...and reset the frame timer to start counting all over again
                     config.frame_timer = AnimationConfig::timer_from_fps(config.fps);
                 }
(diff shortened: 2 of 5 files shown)


crates/bevy_sprite/src/texture_atlas.rs
```diff
@@ -170,6 +171,8 @@ pub struct TextureAtlas {
     pub layout: Handle<TextureAtlasLayout>,
     /// Texture atlas section index
     pub index: usize,
+    /// Texture atlas default index range (first, last)
```
alice-i-cecile, line 174:
```suggestion
    /// The index of the first and last images in the atlas
```

alice-i-cecile (changes requested):
I really like these changes, but I had a hard time understanding the docs at first. I've left a simple suggestion to improve that.

alice-i-cecile (approved):

crates/bevy_sprite/src/texture_atlas.rs
```diff
@@ -178,13 +181,22 @@ impl TextureAtlas {
         let atlas = texture_atlases.get(&self.layout)?;
         atlas.textures.get(self.index).copied()
     }
+
+    pub fn advance_index(&mut self) -> &mut Self {
+        self.index += 1;
+        if self.index > self.index_range.1 {
+            self.index = self.index_range.0;
+        }
```
BenjaminBrienen, line 189: This looks like it should just use a saturating add.

examples/ui/ui_texture_atlas_slice.rs
```diff
@@ -31,7 +31,7 @@ fn button_system(
             Interaction::Pressed => {
                 **text = "Press".to_string();
                 if let Some(atlas) = &mut image.texture_atlas {
-                    atlas.index = (atlas.index + 1) % 30;
```
BenjaminBrienen, line 34: This is supposed to cycle

BenjaminBrienen (changes requested):

rparrett:
Sort of opposed to this change. I feel like it's of way too limited utility to be given prime real estate in the `TextureAtlas` struct, and I'm not sure that animation metadata belongs in there anyway.

Definitely want to add stuff to help with 2d animations... I just think that whatever we end up with will look nothing like what is being done in that PR.

I'd maybe implement it as an optional `AtlasAnimation { frames: Vec<usize> }` or `AtlasAnimation { start: usize, end: usize}` component with helper methods or something.  But that last one feels pretty restrictive.

examples/2d/sprite_animation.rs
```diff
@@ -64,10 +64,10 @@ fn execute_animations(time: Res<Time>, mut query: Query<(&mut AnimationConfig, &
             if let Some(atlas) = &mut sprite.texture_atlas {
                 if atlas.index == config.last_sprite_index {
                     // ...and it IS the last frame, then we move back to the first frame and stop.
-                    atlas.index = config.first_sprite_index;
+                    atlas.advance_index();
                 } else {
                     // ...and it is NOT the last frame, then we move to the next frame...
-                    atlas.index += 1;
+                    atlas.advance_index();
```
mgi388, line 70: FWIW, this change looks subtly wrong to me. Previously the logic was:

- If the index is not the last, then increment it (inline).

Now the logic is (due to `advance_index`):

- If the index is not the last, then increment it (inside `advance_index`).
- (Inside `advance_index`) If the index is more than the last set it to the index of the first.

Even if this particular change is the same as before (I didn't completely check because it looks subtly wrong and it's no longer intuitive, at least to me), my main concerns are left in a comment at the end of the review.

mgi388 (commented):
For me this change doesn't stand out as a net improvement.

The name `advance_index` implies to me that it advances the index, not that it advances and possibly wraps back around.

It feels weird to me that you'd now have to pass `index_range` to your `TextureAtlas` (even if using `Default` to have it receive a default value).

Some problems I have with `index_range`:

1. `index_range` does not imply that it's a range for the _first and last index_. Some valid ranges might just be a section of the atlas (e.g. index 0 to 7 out of an atlas with 256 frames), so `index_range`'s name and functionality seems limited. Further, then `advance_index` only works on the range 0 to 7.
2. An entity with a `TextureAtlas` probably has _multiple_ valid ranges, not just one. A texture atlas can have animations in it for walking north, south, etc., running north, running south, etc. See https://docs.rs/bevy_spritesheet_animation/1.0.0/bevy_spritesheet_animation/spritesheet/struct.Spritesheet.html for an example crate which can create different "ranges" and how there is a wide variety, certainly more than just "start to end incrementing by 1".
3. All my existing code that constructs `TextureAtlas` would now have to pass `index_range` even though it may not even be relevant for my entity. A concrete example follows. I have a UI image with a `TextureAtlas`. I have an atlas of textures and my UI image comes from index 48.

```rust
commands
    .spawn((
        ImageBundle {
            style: Style {
                // ...
                ..Default::default()
            },
            image: UiImage::new(hud_sprite_sheet.texture.clone()),
            ..Default::default()
        },
        TextureAtlas {
            layout: hud_sprite_sheet.texture_atlas_layout.clone(),
            index: 48,
        },
    ))
    .id();
```

I never need to animate this UI image. So if I was forced to pass `index_range` here, even if indirectly through `Default`, it doesn't really make sense to me.

IceSentry:
I agree with the suggestions that a similar feature should be implemented in a separate helper struct/function and not directly on the texture atlas

PR #16612, opened 2024-12-02
Make `StateTransitionSteps` public

### LINKED ISSUE #16594: Make `StateTransitionSteps` public for better ergonomics of State Scoped Resources

As of `0.15` `Entity`s and `Event`s can be bound to a state.
Entities will be despawned on state exit and events will be initialized and cleaned-up on state exit.
There's no official way of doing this for resources but we can make our own.

Consider this naive impl: 
```rust
fn init_state_scoped_resource<R: Resource + FromWorld>(
        &mut self,
        state: impl States,
    ) -> &mut Self {
        self.add_systems(OnEnter(state.clone()), init_state_scoped_resource_impl::<_, R>(state.clone()));
        self.add_systems(OnExit(state.clone()), clear_state_scoped_resource_impl::<_, R>(state));
        self
}
```

```rust
app.init_state_scoped_resource::<Foo>(BazState::Bar);
```
While it does work it can race with this system:
```rust
app.add_systems(OnEnter(BazState::Bar), |foo: Res<Foo>| {});
```

There are 3 solutions I can think of:
1. Keep using naive OnEnter/OnExit and offer some `SystemSet` to let users (me) order racy systems but this is imo quite un-ergonomic and error prone - you can forget to do it.
2. Switch OnEnter/OnExit schedules to `StateTransition` schedule and modify inner init/clear fns to behave like [clear_state_scoped_entities](https://github.com/bevyengine/bevy/blob/56d559102858d4ce8a5b5e3bccd7053dd687a197/crates/bevy_state/src/state_scoped.rs#L67). This fixes possible race and there's no need to offer system set but it will behave differently than other systems running in that schedule and I'm not sure what effects it could have.
3. Like 2. + add `.in_set(StateTransitionSteps::EnterSchedules)` and `.in_set(StateTransitionSteps::ExitSchedules)` to user-defined `init_state_scoped_resource` inner fns. This will offer the "best" compatibility out of the mentioned 3 but it's not possible because `StateTransitionSteps` is [private](https://github.com/bevyengine/bevy/blob/56d559102858d4ce8a5b5e3bccd7053dd687a197/crates/bevy_state/src/state/transitions.rs#L71).

I can't think of other alternatives.

## Additional context

None.

Phoqinu:
# Objective

- Fixes #16594 

## Solution

- `StateTransitionSteps` was made public

## Testing

- _Did you test these changes? If so, how?_
 I am now able to import it and use it.

---

## Showcase

- Users can now write their own state scoped resources

```rust
    fn init_state_scoped_resource<R: Resource + FromWorld>(
        &mut self,
        state: impl States,
    ) -> &mut Self {
        use bevy::state::state::StateTransitionSteps; // this PR in action

        self.add_systems(
            StateTransition,
            (
                clear_state_scoped_resource_impl::<_, R>(state.clone())
                    .in_set(StateTransitionSteps::ExitSchedules), // and here
                init_state_scoped_resource_impl::<_, R>(state)
                    .in_set(StateTransitionSteps::EnterSchedules), // here too
            ),
        );

        self
    }
```

### DIFF
--- a/crates/bevy_state/src/state/transitions.rs
+++ b/crates/bevy_state/src/state/transitions.rs
@@ -68,7 +68,7 @@ pub struct StateTransitionEvent<S: States> {
 ///
 /// These system sets are run sequentially, in the order of the enum variants.
 #[derive(SystemSet, Clone, Debug, PartialEq, Eq, Hash)]
-pub(crate) enum StateTransitionSteps {
+pub enum StateTransitionSteps {
     /// States apply their transitions from [`NextState`](super::NextState)
     /// and compute functions based on their parent states.
     DependentTransitions,


alice-i-cecile (approved):
Agreed: system sets should be public by default.

MiniaczQ (approved):

PR #16613, opened 2024-12-02
BrpQueryRow has field deserialization fix

Leinnan:
# Objective

BrpQueryRow doesn't serialize `has` field if it is empty. That is okay until you try to deserialize it after. Then it will fail to deserialize due to missing field.

## Solution

Serde support using default value when field is missing, this PR adds that.


alice-i-cecile (approved):
Sensible fix, but I'd be happier if there was a test too!

kristoff3r (approved):

Leinnan:
@alice-i-cecile I have already WIP of serialization tests for remote structs on my machine and I will try to create PR before end of the week, but I wanted to have that merged first, since it is blocking issue for me for using bevy_remote module in my remote inspector.

Issue #16085, opened 2024-10-24
Let users control the line height of their text


rparrett:
## What problem does this solve or what need does it fill?

Line height is currently hard-coded to a value of `1.2`. We should let users control this.

https://github.com/bevyengine/bevy/blob/9274bfed277725e8987f8755931dc4d83e5fa08f/crates/bevy_text/src/pipeline.rs#L486

## What solution would you like?

Add a "relative line height" field to `TextFont` and use it in place of the hardcoded `1.2`.

## What alternative(s) have you considered?

We may want to argue about where exactly to put this field.

ickshonpe:
I think this should be a separate component, not a field on `TextFont` probably. It's a property of the text layout, not the font,

BenjaminBrienen:
Should this have its own example, or is there an existing example that demonstrates a bunch of related features to which we could add this?

TheOpenDevProject:
Please merge this in as I really need the ability to dynamically set the line height

BenjaminBrienen:
> Please merge this in as I really need the ability to dynamically set the line height

Would you like to try implementing this yourself? Otherwise, I might try it sometime this week.

rparrett:
Fixed by #16614

PR #16614, opened 2024-12-02
add line height to `TextFont`

Cyborus04:
# Objective

- Allow users to customize the line height of text.
- Implements #16085

## Solution

- Add a `line_height` field to `TextFont` to feed into `cosmic_text`'s `Metrics`.

## Testing

- Tested on my own game, and worked exactly as I wanted.
- My game is only 2D, so I only tested `Text2d`. `Text` still needs tested, but I imagine it'll work fine.
- An example is available [here](https://code.cartoon-aa.xyz/Cyborus/custom-line-height-example)

---

## Showcase

<details>

With font:
```rust
TextFont {
    font: /* unimportant */,
    font_size: 16.0,
    line_height: None,
    ..default()
}
```

image

With font:
```rust
TextFont {
    font: /* unimportant */,
    font_size: 16.0,
    line_height: Some(16.0),
    ..default()
}
```

image

</details>

## Migration Guide

`TextFont` now has a `line_height` field. Any instantiation of `TextFont` that doesn't have `..default()` will need to add this field.

### DIFF
--- a/crates/bevy_text/src/pipeline.rs
+++ b/crates/bevy_text/src/pipeline.rs
@@ -98,6 +98,7 @@ impl TextPipeline {
         // Collect span information into a vec. This is necessary because font loading requires mut access
         // to FontSystem, which the cosmic-text Buffer also needs.
         let mut font_size: f32 = 0.;
+        let mut line_height: f32 = 0.0;
         let mut spans: Vec<(usize, &str, &TextFont, FontFaceInfo, Color)> =
             core::mem::take(&mut self.spans_buffer)
                 .into_iter()
@@ -130,6 +131,7 @@ impl TextPipeline {
 
             // Get max font size for use in cosmic Metrics.
             font_size = font_size.max(text_font.font_size);
+            line_height = line_height.max(text_font.line_height.eval(text_font.font_size));
 
             // Load Bevy fonts into cosmic-text's font system.
             let face_info = load_font_to_fontdb(
@@ -146,7 +148,6 @@ impl TextPipeline {
             spans.push((span_index, span, text_font, face_info, color));
         }
 
-        let line_height = font_size * 1.2;
         let mut metrics = Metrics::new(font_size, line_height).scale(scale_factor as f32);
         // Metrics of 0.0 cause `Buffer::set_metrics` to panic. We hack around this by 'falling
         // through' to call `Buffer::set_rich_text` with zero spans so any cached text will be cleared without
@@ -486,7 +487,13 @@ fn get_attrs<'a>(
         .stretch(face_info.stretch)
         .style(face_info.style)
         .weight(face_info.weight)
-        .metrics(Metrics::relative(text_font.font_size, 1.2).scale(scale_factor as f32))
+        .metrics(
+            Metrics {
+                font_size: text_font.font_size,
+                line_height: text_font.line_height.eval(text_font.font_size),
+            }
+            .scale(scale_factor as f32),
+        )
         .color(cosmic_text::Color(color.to_linear().as_u32()));
     attrs
 }
--- a/crates/bevy_text/src/text.rs
+++ b/crates/bevy_text/src/text.rs
@@ -285,6 +285,11 @@ pub struct TextFont {
     /// A new font atlas is generated for every combination of font handle and scaled font size
     /// which can have a strong performance impact.
     pub font_size: f32,
+    /// The vertical height of a line of text, from the top of one line to the top of the
+    /// next.
+    ///
+    /// Defaults to `LineHeight::RelativeToFont(1.2)`
+    pub line_height: LineHeight,
     /// The antialiasing method to use when rendering text.
     pub font_smoothing: FontSmoothing,
 }
@@ -324,8 +329,36 @@ impl Default for TextFont {
         Self {
             font: Default::default(),
             font_size: 20.0,
+            line_height: LineHeight::default(),
             font_smoothing: Default::default(),
         }
+    }
+}
+
+/// Specifies the height of each line of text for `Text` and `Text2d`
+///
+/// Default is 1.2x the font size
+#[derive(Debug, Clone, Copy, Reflect)]
+#[reflect(Debug)]
+pub enum LineHeight {
+    /// Set line height to a specific number of pixels
+    Px(f32),
+    /// Set line height to a multiple of the font size
+    RelativeToFont(f32),
+}
+
+impl LineHeight {
+    pub(crate) fn eval(self, font_size: f32) -> f32 {
+        match self {
+            LineHeight::Px(px) => px,
+            LineHeight::RelativeToFont(scale) => scale * font_size,
+        }
+    }
+}
+
+impl Default for LineHeight {
+    fn default() -> Self {
+        LineHeight::RelativeToFont(1.2)
     }
 }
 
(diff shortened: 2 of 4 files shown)


crates/bevy_text/src/text.rs
```diff
@@ -284,6 +284,11 @@ pub struct TextFont {
     /// A new font atlas is generated for every combination of font handle and scaled font size
     /// which can have a strong performance impact.
     pub font_size: f32,
+    /// The vertical height of a line of text in pixels, from the top of one line to the top of the
+    /// next.
+    ///
+    /// If not set, the line height will be 1.2 * `font_size`.
+    pub line_height: Option<f32>,
```
ickshonpe, line 291: Is there a point to making this optional, it could just be an `f32` with a default value of 1.2 instead?

ickshonpe (changes requested):

crates/bevy_text/src/text.rs
Cyborus04: It's measured in pixels, not multiples of the font size (though that might be a better interface). Working with `Options` makes it simpler to implement too IMO, but changing it wouldn't be a huge deal

ickshonpe: Oh I see it's not a scalar. I think then it would better if there was a `LineHeight` enum maybe with `Px` and `Scalar` variants or something.

Cyborus04: I like that! Best of both worlds

Cyborus04: in that case, should it be a field of `TextFont`, or its own component?

ickshonpe: Otherwise it looks good, the UI text debug examples are still working with different line heights, don't see any problems.

ickshonpe: > in that case, should it be a field of `TextFont`, or its own component? Being a separate component would make it so it's not a breaking change at all, but I'm not sure yet how to do that

The current direction atm with bevy seems to be less granular APIs so a field on `TextFont` is preferable probably. A field is more discoverable than a distinct component as well.

crates/bevy_text/src/text.rs
```diff
@@ -323,11 +328,38 @@ impl Default for TextFont {
...
+            line_height: LineHeight::default(),
             font_smoothing: Default::default(),
         }
     }
 }
 
+/// Specifies the height of each line of text for `Text` and `Text2d`
+///
+/// Default is 1.2x the font size
+#[derive(Debug, Clone, Copy, Reflect)]
```
coreh, line 340: This is missing a `#[reflect(Debug)]` and also a `.register_type::<LineHeight>()` on the plugin.

crates/bevy_text/src/text.rs
```diff
@@ -323,11 +328,38 @@ impl Default for TextFont {
...
+pub enum LineHeight {
+    /// Set line height to a specific number of pixels
+    Pixels(f32),
```
coreh, line 343: We can't reuse `bevy_ui`'s `Val` here, but it would be nice to align the naming of the pixel size variant to `LineHeight::Px(f32)`, so that it's consistent with `Val::Px(f32)`

crates/bevy_text/src/text.rs
```diff
@@ -323,11 +328,38 @@ impl Default for TextFont {
...
+        match self {
+            LineHeight::Pixels(px) => px,
+            LineHeight::Scalar(scale) => scale * font_size,
+        }
+    }
+}
+
+impl Default for LineHeight {
+    fn default() -> Self {
+        LineHeight::Scalar(1.2)
```
coreh, line 359: To make it easier to expand the feature set in the future (by propagating inherited line height across a hierarchy etc) without having to do a breaking change, would it make sense to have a separate `LineHeight::Auto` variant as the default?

crates/bevy_text/src/pipeline.rs
```diff
@@ -146,7 +153,7 @@ impl TextPipeline {
             spans.push((span_index, span, text_font, face_info, color));
         }
 
-        let line_height = font_size * 1.2;
+        let line_height = line_height_pixel.max(line_height_scalar * font_size);
```
coreh, line 156: Hmm... If I understand it correctly, this should produce a combination of the largest font size with the largest line height scalar, even if they weren't used on the same span. This can produce unexpected line height results if you have a small font with a large scalar and a large font with a small scalar on the same line.

coreh (commented):
Nice! Left a couple comments, mostly about the `LineHeight` enum, but also about a potential issue with the line height math.

crates/bevy_text/src/text.rs
Cyborus04: What would that do?

BenjaminBrienen:
@coreh everything check out?

coreh:
At a quick glance, LGTM! 👍

coreh (approved):

BenjaminBrienen:
You need to fix the no-std build. It can be in final review after 2 community approvals.

Cyborus04:
I'm not sure what I need to do to fix the nostd build, the error was in a different crate than my changes

BenjaminBrienen:
Weird. FYI @bushrat011899

bushrat011899:
@Cyborus04 just tested your PR locally, it'll pass the `no_std` checks once you update this PR to the latest changes on `main`. I don't have a good answer as to _why_ you need to upstream from `main`, but that should fix the issue.

Cyborus04:
I had a suspicion that's all I needed to do, I just wasn't at a computer to do so then

crates/bevy_text/src/text.rs
```diff
@@ -324,11 +329,39 @@ impl Default for TextFont {
...
+/// Specifies the height of each line of text for `Text` and `Text2d`
+///
+/// Default is 1.2x the font size
+#[derive(Debug, Clone, Copy, Reflect)]
+#[reflect(Debug)]
+pub enum LineHeight {
+    /// Set line height to a specific number of pixels
+    Px(f32),
+    /// Set line height to a multiple of the font size
+    Scalar(f32),
```
rparrett, line 347: I am not sure that the name `Scalar` makes sense here.

Maybe `Relative`? `Em`? I will ask other UI people to weigh in.

alice-i-cecile: FontRelative gets my vote.

ickshonpe: `Relative` / `Em` might be a bit confusing and `FontRelative` seems a bit odd.

`RelativeToFontSIze` ? It's a bit long but completely unambiguous.

Jondolf: In CSS, the equivalent here seems to be just a unitless number. So this at least shouldn't be called `Em`, since that behaves differently.
<https://developer.mozilla.org/en-US/docs/Web/CSS/line-height#values>

I think `Relative`, `RelativeToFont`, or `Factor` would be my suggestion

ickshonpe: Yeah `RelativeToFont` seems fine to me.

BenjaminBrienen (approved):

alice-i-cecile (approved):

rparrett:
Making a note that the `Showcase` section of the PR description still has older syntax from a previous iteration of this PR.

PR #16615, opened 2024-12-02
Move UI `Transform` to a field on `Node`

ickshonpe:
# Objective

Animating UI elements by modifying the `Transform` component of UI nodes doesn't work very well because `ui_layout_system` overwrites the translations each frame. The `overflow_debug` example uses a horrible hack where it copies the transform into the position that'll likely cause a panic if any users naively copy it.

## Solution
There have been a couple of other attempts to fix this with `UiTransform` components etc but they were quite complicated and controversial and got bogged down in review. So I thought about it again and came up with this simpler alternative: 
* Opt-out of the builtin transform propagation by requiring `GlobalTransform` instead of `Transform` on `Node`.
* Add a `transform: Transform` field to `Node`.
* Update UI node global transforms during layout updates.

### DIFF
--- a/crates/bevy_ui/src/ui_node.rs
+++ b/crates/bevy_ui/src/ui_node.rs
@@ -9,7 +9,7 @@ use bevy_render::{
     view::Visibility,
 };
 use bevy_sprite::BorderRect;
-use bevy_transform::components::Transform;
+use bevy_transform::components::{GlobalTransform, Transform};
 use bevy_utils::warn_once;
 use bevy_window::{PrimaryWindow, WindowRef};
 use core::num::NonZero;
@@ -305,9 +305,9 @@ impl From<&Vec2> for ScrollPosition {
     BorderRadius,
     FocusPolicy,
     ScrollPosition,
-    Transform,
     Visibility,
-    ZIndex
+    ZIndex,
+    GlobalTransform
 )]
 #[reflect(Component, Default, PartialEq, Debug)]
 #[cfg_attr(
@@ -586,6 +586,9 @@ pub struct Node {
     ///
     /// <https://developer.mozilla.org/en-US/docs/Web/CSS/grid-column>
     pub grid_column: GridPlacement,
+
+    /// Rotate, scale, skew, or translate an element.
+    pub transform: Transform,
 }
 
 impl Node {
@@ -628,6 +631,7 @@ impl Node {
         grid_auto_columns: Vec::new(),
         grid_column: GridPlacement::DEFAULT,
         grid_row: GridPlacement::DEFAULT,
+        transform: Transform::IDENTITY,
     };
 }
 
--- a/crates/bevy_ui/src/layout/convert.rs
+++ b/crates/bevy_ui/src/layout/convert.rs
@@ -436,6 +436,8 @@ impl RepeatedGridTrack {
 
 #[cfg(test)]
 mod tests {
+    use bevy_transform::components::Transform;
+
     use super::*;
 
     #[test]
@@ -509,6 +511,7 @@ mod tests {
             ],
             grid_column: GridPlacement::start(4),
             grid_row: GridPlacement::span(3),
+            transform: Transform::IDENTITY,
         };
         let viewport_values = LayoutContext::new(1.0, bevy_math::Vec2::new(800., 600.));
         let taffy_style = from_node(&node, &viewport_values, false);
--- a/crates/bevy_ui/src/layout/mod.rs
+++ b/crates/bevy_ui/src/layout/mod.rs
@@ -13,10 +13,10 @@ use bevy_ecs::{
     world::Ref,
 };
 use bevy_hierarchy::{Children, Parent};
-use bevy_math::{UVec2, Vec2};
+use bevy_math::{Mat4, UVec2, Vec2};
 use bevy_render::camera::{Camera, NormalizedRenderTarget};
 use bevy_sprite::BorderRect;
-use bevy_transform::components::Transform;
+use bevy_transform::components::GlobalTransform;
 use bevy_utils::tracing::warn;
 use bevy_window::{PrimaryWindow, Window, WindowScaleFactorChanged};
 use derive_more::derive::{Display, Error, From};
@@ -124,10 +124,10 @@ pub fn ui_layout_system(
     computed_node_query: Query<(Entity, Option<Ref<Parent>>), With<ComputedNode>>,
     ui_children: UiChildren,
     mut removed_components: UiLayoutSystemRemovedComponentParam,
-    mut node_transform_query: Query<(
+    mut node_update_query: Query<(
         &mut ComputedNode,
-        &mut Transform,
         &Node,
+        &mut GlobalTransform,
         Option<&BorderRadius>,
         Option<&Outline>,
         Option<&ScrollPosition>,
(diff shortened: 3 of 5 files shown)


alice-i-cecile:
I like this as an incremental move forward. We shouldn't be storing a full `Transform` here, but that's easy to migrate from this PR.

NthTensor:
This does seem like the least-bad option. But I thought there was code in `bevy_transform` to specifically filter out ui elements from propagation. Could you check if that's still there and remove it if it is?

I believe this should fix the issues `big_space` had with the use of `Transform` by ui elements, which is nice.

ickshonpe:
> I like this as an incremental move forward. We shouldn't be storing a full `Transform` here, but that's easy to migrate from this PR.
> 
> Should this be on `ComputedNode` instead though? I don't think this is intended to be user-modifiable.

The new transform field on `Node` is intended to be user-modifiable. It's meant to be simar to the CSS transform property and allows users to add simple UI animations or whatever without weird hacks to get around the conflicts between the transformation propagation and UI layout update systems.

I made another branch first just before this one that removed both `Transform` and `GlobalTransform` from UI nodes and instead added the transform to `ComputedNode` but it required lots of changes to extraction and focus and writing a UI specific visibility system. Performance is much better than main with that approach though and would allows us to add helper methods to `Node` like `contains_point` etc which would simplify a lot of the systems.

alice-i-cecile:
@ickshonpe I'm really keen on this as a step towards the great transform splittification :) Is this ready for review?

ickshonpe:
> @ickshonpe I'm really keen on this as a step towards the great transform splittification :) Is this ready for review?

Probably, I'll take another look at this first thing tomorrow. I've been thinking though do we even need a transform component for UI? We could just stick the transform values in `ComputedNode` along with everything else. `ComputedNode` is huge already but a lot of UI/2D devs are allergic to linear algebra and it would mean we could add all the `contains_point` and `relative_position` helper functions that people have been requesting forever.

The other alternative is a `TransformedNode(&ComputedNode, &UiTransform)` wrapper that we implement all the helper functions on.

ickshonpe:
> Looks ok. I'm not sold on the transform as field in `ComputedNode` but I'm not sold either on any alternative.
> 

Yeah I don't know either. I did it this way because it was easy to implement but it's relatively trivial to switch it to use a separate `UiTransform`/`UiGlobalTransform`. 

> Quick Q: How one would create a 3D UI with this change? Is there a 3D transform at the root of the node hierarchy? Or would the user have to render to an off screen render target the 2D UI then lay it out in 3D in the scene as a texture?

You couldn't really create a 3d UI before this change it was all 2d coords, it just used the 3d transform. Either way you have to render to texture for world UI.

djeedai (approved):

ickshonpe:
@alice-i-cecile I think this is more or less good enough now. Abandoned the weird ideas I had, just got a `UiTransform` component that replaces `Transform`. I wanted to keep this a simple PR but ended up making a lot of other changes, figured it's better to fix everything here than migrate all the bugs to the new setup.

djeedai:
Should this PR move to Draft? Was said to be nearly done, then got 12 commits merged. Also designed completely changed from initial version. I'm not sure as a reviewer when it's safe to spend time reviewing again. Thanks

ickshonpe:
> Should this PR move to Draft? Was said to be nearly done, then got 12 commits merged. Also designed completely changed from initial version. I'm not sure as a reviewer when it's safe to spend time reviewing again. Thanks

Yeah my fault keep changing my mind as to how this should be implemented. I think it's okay now though.

crates/bevy_ui/src/ui_transform.rs
```diff
@@ -0,0 +1,188 @@
...
+    feature = "serialize",
+    derive(serde::Serialize, serde::Deserialize),
+    reflect(Serialize, Deserialize)
+)]
+#[require(UiGlobalTransform)]
+pub struct UiTransform {
+    /// Translate the node.
+    pub translation: Val2,
+    /// Scale the node. A negative value reflects the node in that axis.
+    pub scale: Vec2,
```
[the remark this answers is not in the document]
ickshonpe: Yep only went with scaling in both axis because it was supported with `Transform`/`GlobalTransform` and I wanted to avoid too much controversy with this PR.

NthTensor (approved):

crates/bevy_ui/src/ui_transform.rs
```diff
@@ -0,0 +1,191 @@
+use crate::Val;
+use bevy_derive::Deref;
+use bevy_ecs::component::Component;
+use bevy_ecs::prelude::ReflectComponent;
+use bevy_math::Affine2;
+use bevy_math::Rot2;
+use bevy_math::Vec2;
+use bevy_reflect::prelude::*;
+
+/// A pair of [`Val`]s used to representin a 2-dimensional size or offset.
```
eckz: Typo: *representin*

crates/bevy_ui/src/ui_transform.rs
```diff
@@ -0,0 +1,191 @@
...
+    derive(serde::Serialize, serde::Deserialize),
+    reflect(Serialize, Deserialize)
+)]
+pub struct Val2 {
+    /// Translate the node along the x-axis.
+    /// `Val::Percent` values are resolved based on the computed width of the Ui Node.
+    /// `Val::Auto` is resolved to `0.`.
+    pub x: Val,
+    /// Translate the node along the y-axis.
+    /// `Val::Percent` values are resolved based on the computed width of the Ui Node.
```
eckz: I guess this is a typo, it should say `based on the computed **height** of the Ui Node.`

crates/bevy_ui/src/ui_node.rs
```diff
@@ -324,12 +393,12 @@ impl From<Vec2> for ScrollPosition {
 #[require(
     ComputedNode,
     ComputedNodeTarget,
+    UiTransform,
```
eckz, line 396: Just an idea. You could make `Node` depend only on `UiGlobalTransform`, making `UiTransform` an optional component.

The only difference would be using `Option<&UiTransform>` instead of `&UiTransform` in `ui_layout_system`, saving a call to `compute_affine` when the component is not present.

NthTensor (commented):
I think we may want to consider folding `GlobalZIndex` and `ZIndex` into the UI transform eventually.

crates/bevy_ui/src/ui_transform.rs
alice-i-cecile:
```suggestion
/// A pair of [`Val`]s used to represent a 2-dimensional size or offset.
```

crates/bevy_ui/src/ui_transform.rs
alice-i-cecile:
```suggestion
    /// `Val::Percent` values are resolved based on the computed height of the UI Node.
```

viridia:
> The cursor position is transformed to local node space during picking so that it respects rotations and scalings.

I'm not sure this solves the problem I have been having with pointer interactions for multi-part widgets. Consider a slider with a draggable thumb: you click on the thumb entity, so the picking event you get has the coordinates relative to the thumb. But what you need for the drag calculation is the coordinates relative to the slider as a whole.

What I'd like is for the observer function to have the option to transform the picking coordinates into local coordinates for any arbitrary ancestor of the picked element. The easiest way to do this, I think, is to have the event contain global coordinates, and then have the observer do a conversion to local coordinates using the global transform component on the ancestor.

PR #16619, opened 2024-12-03
switch bevy_mesh to use wgpu-types instead of wgpu

atlv24:
# Objective

- dont depend on wgpu if we dont have to

## Solution

- works towards this, but doesnt fully accomplish it. bevy_mesh depends on bevy_image

## Testing

- 3d_scene runs

### DIFF
--- a/crates/bevy_mesh/src/primitives/extrusion.rs
+++ b/crates/bevy_mesh/src/primitives/extrusion.rs
@@ -72,7 +72,7 @@ impl PerimeterSegment {
 /// ## Warning
 ///
 /// By implementing this trait you guarantee that the `primitive_topology` of the mesh returned by
-/// this builder is [`PrimitiveTopology::TriangleList`](wgpu::PrimitiveTopology::TriangleList)
+/// this builder is [`PrimitiveTopology::TriangleList`](crate::PrimitiveTopology::TriangleList)
 /// and that your mesh has a [`Mesh::ATTRIBUTE_POSITION`] attribute.
 pub trait Extrudable: MeshBuilder {
     /// A list of the indices each representing a part of the perimeter of the mesh.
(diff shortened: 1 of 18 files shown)


mockersf:
the doc for the `Extrudable` trait still mentions `wgpu`

crates/bevy_mesh/src/primitives/extrusion.rs
```diff
@@ -72,7 +72,7 @@ impl PerimeterSegment {
 /// ## Warning
 ///
 /// By implementing this trait you guarantee that the `primitive_topology` of the mesh returned by
-/// this builder is [`PrimitiveTopology::TriangleList`](wgpu::PrimitiveTopology::TriangleList)
+/// this builder is [`PrimitiveTopology::TriangleList`](crate::PrimitiveTopology::TriangleList)
```
BD103, line 75:
```suggestion
/// this builder is [`PrimitiveTopology::TriangleList`](PrimitiveTopology::TriangleList)
```

(Nit) The `crate` shouldn't be necessary, since you import `PrimitiveTopology` already.

BD103 (approved):

crates/bevy_mesh/src/primitives/extrusion.rs
```diff
@@ -72,7 +72,7 @@ impl PerimeterSegment {
 /// ## Warning
 ///
 /// By implementing this trait you guarantee that the `primitive_topology` of the mesh returned by
-/// this builder is [`PrimitiveTopology::TriangleList`](wgpu::PrimitiveTopology::TriangleList)
+/// this builder is [`PrimitiveTopology::TriangleList`](PrimitiveTopology::TriangleList)
```
BD103, line 75:
```suggestion
/// this builder is [`PrimitiveTopology::TriangleList`]
```

This should fix CI! `PrimitiveTopology` is already imported, so the "link" is redundant.

BD103: + The failure on the dependencies check is unrelated, so there's not much to be done about that.

atlv24: thanks!

PR #4166, opened 2022-03-09
Add `&mut World` as a system param and make `.exclusive_system()` optional

maniwani:
## Objective

- finish what started in #2398
- make `.exclusive_system()` optional (`FunctionSystem` handles all systems now)
- lay groundwork for bevyengine/rfcs#45

## Solution

- add ~~`SemiSafeCell<'w, World>`~~ `MaybeUnsafeCell<'w, World>` and have it replace `&'w World` in internal access methods
- add new required method on `SystemParamState` so we can (mostly) statically ensure soundness and detect exclusive systems
- add `&mut World` as a system param
- remove `ExclusiveSystem*` types and traits
- ~~mark `.exclusive_system()` method as deprecated~~ (we might need it later to coerce systems into being exclusive)

## Related

- #2923
- #3001 (main inspiration for cell enum/union type)
- #3946 (superceded by this PR)

PR #16622, opened 2024-12-03
Unify ExclusiveSystemParams with normal SystemParams

ItsDoot:
# Objective

Remove `ExclusiveSystemParam` and `ExclusiveFunctionSystem` in favor of just `SystemParam`s, and define a system's exclusivity based on its parameters' `SystemParam::world_access_level` return value.

## Solution

~~Note: This PR relies on #16642 to be merged first.~~ EDIT: Merged.

- Removed `ExclusiveSystemParam`, `ExclusiveFunctionSystem`, and `ExclusiveSystemParamFunction`.
- `&mut World`, `&mut QueryState<D, F>`, and `&mut SystemState<P>` now implement `SystemParam`.
    - `&mut World` registers exclusive access, so it cannot be used alongside any parameter that accesses the world
    - `&mut QueryState` and `&mut SystemState` don't hold access to the world, so can be used alongside it
- As `SystemParam`s:
    - `&mut QueryState` now will automatically update its archetypes via `SystemParam::new_archetype`
    - `&mut SystemState` will now have its deferred mutations automatically applied via `SystemParam::apply`
- `DeferredWorld`'s `init_state` was updated to align with `&World` and `&mut World`

## TODO

- Observers cannot be allowed to use `&mut World`, but this PR in its current state allows it.
    - We should allow them to use `DeferredWorld`, however. We should consider splitting "exclusivity" into "structural exclusivity" and "deferred exclusivity".
- Decide if `&mut QueryState` and `&mut SystemState` shouldn't do their things above automatically.
- Consider adding new lints to `bevy_lint` to tell users not to use two `&mut World`s in a system or `&mut World` + `Query`/`Res`/etc

## Testing

Most pre-existing tests are passing currently, will need to add some new ones before merging.

---

## Showcase

Systems with `&mut World` no longer require it to be the first parameter (or second if using system input). It can now be placed anywhere in the function arguments:

```rust
// Before:
fn do_stuff(world: &mut World, local: Local<usize>, query: &mut QueryState<&Foo>) {}
// or
fn do_stuff_with(In(v): In<i32>, world: &mut World, local: Local<usize>, query: &mut QueryState<&Foo>) {}

// Now also possible:
fn do_stuff(local: Local<usize>, world: &mut World, query: &mut QueryState<&Foo>) {}
// or
fn do_stuff(local: Local<usize>, query: &mut QueryState<&Foo>, world: &mut World) {}
// or
fn do_stuff_with(In(v): In<i32>, local: Local<usize>, query: &mut QueryState<&Foo>, world: &mut World) {}
// or
fn do_stuff_with(In(v): In<i32>, local: Local<usize>, world: &mut World, query: &mut QueryState<&Foo>) {}
```

## Migration Guide

TODO


chescock:
How are you handling parameters like `&Archetypes` and `Query<Entity>` that have no `Access` but still rely on the structure of the world not changing?  That is, a system like

```rust
fn do_something(world: &mut World, archetypes: &Archetypes, query: Query<Entity>) {}
```

would be UB, since the `&mut World` could change archetypes or add entities.  But those params currently register no access and don't look different from `Local`, so I think your change might allow them.  

I think that's why #4166 had `enum WorldAccessLevel`.

hymm:
Added the controversial tag since there was a pr with this approach, but we merged the current approach instead.

You should add tests for the current SystemParams that conflict with &mut World, since that's now a safety concern.

crates/bevy_ecs/src/schedule/executor/mod.rs
```diff
@@ -118,14 +126,119 @@ impl SystemSchedule {
...
+    unsafe fn run_unsafe(
+        &mut self,
+        _input: SystemIn<'_, Self>,
+        _world: UnsafeWorldCell,
+    ) -> Self::Out {
+        panic!("`bevy_ecs::apply_deferred` should not be run directly")
+    }
+
+    fn run(&mut self, _input: SystemIn<'_, Self>, world: &mut World) -> Self::Out {
+        self.last_run = world.change_tick().relative_to(Tick::MAX);
```
hymm, line 202: shouldn't this method panic too like `run_unsafe`?

crates/bevy_ecs/src/schedule/executor/mod.rs
```diff
@@ -118,14 +126,119 @@ impl SystemSchedule {
 /// - This function (currently) does nothing if it's called manually or wrapped inside a [`PipeSystem`](crate::system::PipeSystem).
 /// - Modifying a [`Schedule`](super::Schedule) may change the order buffers are applied.
 #[doc(alias = "apply_system_buffers")]
-#[allow(unused_variables)]
-pub fn apply_deferred(world: &mut World) {}
+#[allow(unused_variables, non_upper_case_globals)]
```
hymm, line 129: I think this change to apply_deferred should be it's own pr as it's really unrelated to the exclusive system trait change. It's also less controversial than the rest of this pr, so it could be merged quickly.

crates/bevy_ecs/src/schedule/executor/mod.rs
```diff
@@ -118,14 +126,119 @@ impl SystemSchedule {
...
+/// - between systems with dependencies if the dependency
+///   [has deferred buffers](crate::system::System::has_deferred)
+///   (if system `bar` directly or indirectly depends on `foo`, and `foo` uses `Commands` param,
+///   changes to the world in `foo` will be visible in `bar`)
+///
+/// ## Notes
+/// - This system (currently) does nothing if it's called manually or wrapped inside a [`PipeSystem`](crate::system::PipeSystem).
+/// - Modifying a [`Schedule`](super::Schedule) may change the order buffers are applied.
+pub struct ApplyDeferred {
+    last_run: Tick,
```
hymm, line 158: why do we need the last_run Tick?

crates/bevy_ecs/src/system/system_param.rs
```diff
@@ -1004,14 +1013,37 @@ unsafe impl SystemParam for &'_ World {
...
 unsafe impl<'w> SystemParam for DeferredWorld<'w> {
     type State = ();
     type Item<'world, 'state> = DeferredWorld<'world>;
 
     fn init_state(_world: &mut World, system_meta: &mut SystemMeta) -> Self::State {
-        system_meta.component_access_set.read_all();
-        system_meta.component_access_set.write_all();
+        let mut access = Access::default();
+        access.read_all();
+        if !system_meta
```
hymm, line 1028: Does the existing code for DeferredWorld not check for conflicts? This change should probably pulled out if so as that's a safety concern.

crates/bevy_ecs/src/system/system_param.rs
```diff
@@ -1023,6 +1055,111 @@ unsafe impl<'w> SystemParam for DeferredWorld<'w> {
...
+    a part of it. Allowing this would break Rust's mutability rules";
+
+// SAFETY: `&mut World` registers exclusive access to the entire world.
+unsafe impl SystemParam for &mut World {
+    type State = ();
+    type Item<'w, 's> = &'w mut World;
+
+    fn init_state(_world: &mut World, system_meta: &mut SystemMeta) -> Self::State {
+        let mut access = Access::default();
+        access.read_all();
```
hymm, line 1075: this should be writes_all

crates/bevy_ecs/src/system/system_param.rs
```diff
@@ -1023,6 +1055,111 @@ unsafe impl<'w> SystemParam for DeferredWorld<'w> {
...
+            .archetype_component_access
+            .is_compatible(&access)
+        {
+            panic!("{}", MUT_WORLD_ERROR);
+        }
+        system_meta.archetype_component_access.extend(&access);
+
+        let mut filtered_access = FilteredAccess::default();
+
+        filtered_access.read_all();
```
hymm, line 1086: I feel like we shouldn't set any filtered_access as &mut World doesn't filter.

chescock: > I feel like we shouldn't set any filtered_access as &mut World doesn't filter.

`component_access_set` is the one that actually gets checked for conflicts by things like `Res` and `Query`, so you do want to populate it.  `FilteredAccess::default()` is unfiltered, so this is declaring read access with no filter.  It could also be written `FilteredAccess::matches_everything()`, which may be more clear.

chescock:
A fun thing this would enable is putting `&mut World` inside a `ParamSet` so you can switch between full access and typed parameters.  That's not any more powerful than putting the other params in a `SystemState`, but it might be more ergonomic in some cases.  

Another thing this would enable is adding `SystemParamBuilder<SystemState<P>>` and `SystemParamBuilder<QueryState<D, F>>` implementations in the future, which would make it easier to use dynamic queries in exclusive systems.

crates/bevy_ecs/src/schedule/executor/mod.rs
ItsDoot: Done: https://github.com/bevyengine/bevy/pull/16642

crates/bevy_ecs/src/system/system_param.rs
```diff
@@ -1025,6 +1059,113 @@ unsafe impl<'w> SystemParam for DeferredWorld<'w> {
...
+const MUT_WORLD_ERROR: &str = "&mut World requires exclusive access to the \
+    entire world, but a previous system parameter already registered access to \
+    a part of it. Allowing this would break Rust's mutability rules";
+
+// SAFETY: `&mut World` registers exclusive access to the entire world.
+unsafe impl SystemParam for &mut World {
+    type State = ();
+    type Item<'w, 's> = &'w mut World;
+
+    fn init_state(_world: &mut World, system_meta: &mut SystemMeta) -> Self::State {
```
urben1680, line 1077: Coming from the `apply_deferred` PR, I am quite sure you also have to call `system_meta.set_non_send()` in this method because exclusive systems were non-send as well.
I don't know however why `&World` and `DeferredWorld` do not do that themselves too prior this PR since these have methods that read/write non-send resources.

crates/bevy_ecs/src/system/system_param.rs
```diff
@@ -273,6 +273,15 @@ pub unsafe trait SystemParam: Sized {
         world: UnsafeWorldCell<'world>,
         change_tick: Tick,
     ) -> Self::Item<'world, 'state>;
+
+    /// Returns true if this [`SystemParam`] requires exclusive access to the
+    /// entire [`World`]. [`System`]s with exclusive parameters cannot run in
+    /// parallel with other systems.
+    ///
+    /// [`System`]: crate::system::System
+    fn is_exclusive() -> bool {
```
cBournhonesque, line 282: Out of curiosity, what is the need for this compared with just setting the Access to `mut_all_components`? (which would mean that the system is incompatible with any other system)?

Is it for optimization reasons that we want to 'pre-compute' this?

ItsDoot: The only way to know if a system is exclusive is based on its parameters now, since we don't have two separate implementations on `System` now.

crates/bevy_ecs/src/system/system_param.rs
ItsDoot: No, at the time of writing this test passes on `main`:
```rust
    #[test]
    fn deferred_world_last() {
        #[derive(Resource)]
        struct FooRes;

        fn bad_system(foo: ResMut<FooRes>, mut world: DeferredWorld) {
            let foo = foo.into_inner();
            let foo = world.get_resource_mut::<FooRes>().unwrap().into_inner();
        }

        let mut world = World::new();
        world.insert_resource(FooRes);
        let mut bad_system = IntoSystem::into_system(bad_system);
        bad_system.initialize(&mut world);
        bad_system.run((), &mut world);
    }
```

ItsDoot:
Closed in favor of #25507

PR #16625, opened 2024-12-03
Expose `SystemMeta`'s access field as part of public API

vil-mo:
# Objective

Outside of the `bevy_ecs` crate it's hard to implement `SystemParam` trait on params that require access to the `World`, because `init_state` expects user to extend access in `SystemMeta` and access-related fields of `SystemMeta` are private.

## Solution

Expose those fields as a functions


BenjaminBrienen (approved):
^_^

alice-i-cecile (changes requested):
Great docs, and I think we should expose this. That said, the mutable accessors need to be unsafe. If you lie about the access you can get UB by allowing aliased mutability at both the schedule and system levels.

alice-i-cecile (approved):

chescock (approved):

PR #16626, opened 2024-12-03
Added `bevy/components` method to BRP, which lists all world components

Lommix:
# New BRP method to list all available fully qualified components in the active world.

After using the newly added BRP feature and trying to build tools around it, I encountered a common issue. There is no way to get all available fully qualified component names from bevy and user space without accumulating them over each entity. Which has an insane cost attached to it.

This is a Problem If you want to provided auto complete for a query API or just want to know what is available without looking into the source code.

This should not be part of the user space, rather be a common method in bevy_remote itself.

## Solution

-  Added a new method that returns a list of all fully qualified component names, active on the current world.
- `method`: `bevy/components`
- `params`: None
- `result`: An array of fully-qualified type names of all world components

## Testing

- Tested on the new `server` example with curl and JSON-RPC.

---

## Showcase

start the server example

```bash
cargo run --example server --features="bevy_remote"
```

get the list result from the new  `bevy/components` method

```bash
curl --silent --no-buffer -X POST -H 'Content-Type: application/json' -d '{"id": 0 ,"jsonrpc":"2.0", "method": "bevy/components", "params": {}}' http://127.0.0.1:15702
```

### Result
[remote_component_list_response.json](https://github.com/user-attachments/files/17992832/remote_component_list_response.json)

### DIFF
--- a/crates/bevy_remote/src/builtin_methods.rs
+++ b/crates/bevy_remote/src/builtin_methods.rs
@@ -797,11 +797,7 @@ pub fn process_remote_list_watching_request(
 }
 
 /// Handles a `bevy/components` request (list all world components) coming from a client.
-pub fn process_remote_components_request(
-    In(_): In<Option<Value>>,
-    world: &World,
-) -> BrpResult {
-
+pub fn process_remote_components_request(In(_): In<Option<Value>>, world: &World) -> BrpResult {
     let components = world.components();
     let mut response = BrpListResponse::default();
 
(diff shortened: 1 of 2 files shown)


Leinnan:
@Lommix Isn't that essentialy the same as `bevy/list`? Example to check:  
```
curl -X POST -d '{ "jsonrpc": "2.0", "id": 1, "method": "bevy/list" }' 127.0.0.1:15702 | jq .    
```

Lommix:
@Leinnan 

Thank you for the enlightenment, you are correct. I did not read correctly.
This is an unnecessary change. Once again, the biggest bug is not in the code, but sitting in front of it xD

I'll close this silently in shame =)

PR #16630, opened 2024-12-03
Update winit_runner to use spawn_app for wasm32 target

### LINKED ISSUE #12562: Allow winit event loop to be recreated on web / wasm
When using Bevy in web currently it is not possible to run bevy multiple times. This is especially important in hybrid application that use Bevy only in game or multiple bevy instances. 

A similar issue was mentioned here [12195](https://github.com/bevyengine/bevy/discussions/12195) in the bevy repo. The proposed solution there is not suited for my usecase since it makes communication between the iframe and the host application difficult to impossible.

There exists an already implemented feature in winit solving the same problem I showed. The related issue is [2897](https://github.com/rust-windowing/winit/pull/2897) and [2720](https://github.com/rust-windowing/winit/pull/2720) in the winit repo. It should be possible to use this feature similarly to how conditionally already some features of android and co are used in the bevy winit plugin.

Currently there seems at least in the bevy side of things no other solution. It might be possible to completely shutdown wasm and load the bevy application completely from scratch. But these solution seem hacky on not like good solutions.

xuwaters:
# Objective

- Fixes #12562
- Fixes #12195

## Solution

- Use `spawn_app` instead of `run_app` for web platform in `winit_runner` as suggested in the [document](https://docs.rs/winit/latest/winit/platform/web/trait.EventLoopExtWebSys.html#tymethod.spawn_app)

## Testing

Tested on web. Created a react app which renders the bevy WASM app and returns the disposer to JS. Js will call the disposer on component unmount to stop the app, the disposer sends a signal to a `signal` receiver in rust which exits the app like this:

```rust
fn handle_stop_signal(
    signal: NonSendMut<StopSignalReceiver>,
    window_entities: Query<(Entity, &Window)>,
    mut event_writer: EventWriter<WindowCloseRequested>,
) {
    if let Ok(_) = signal.try_recv() {
        for (entity, _window) in window_entities.iter() {
            info!("closing window entity: {:x}", entity.to_bits());
            event_writer.send(WindowCloseRequested { window: entity });
        }
    }
}
```

  - No

  - Are all resources released after stopping the app like this? The WASM is still loaded, the LogPlugin complains on the logger re-initialization, but it's a warning anyway.

  - Tested the WASM version on web platform and the native version on MacOS.

### DIFF
--- a/crates/bevy_winit/src/state.rs
+++ b/crates/bevy_winit/src/state.rs
@@ -872,7 +872,7 @@ pub fn winit_runner<T: Event>(mut app: App) -> AppExit {
             }
             // If everything is working correctly then the event loop only exits after it's sent an exit code.
             runner_state.app_exit.unwrap_or_else(|| {
-                error!("Failed to receive a app exit code! This is a bug");
+                error!("Failed to receive an app exit code! This is a bug");
                 AppExit::error()
             })
         }
--- a/crates/bevy_winit/src/state.rs
+++ b/crates/bevy_winit/src/state.rs
@@ -24,6 +24,8 @@ use bevy_tasks::tick_global_task_pools_on_main_thread;
 use bevy_utils::HashMap;
 use bevy_utils::Instant;
 use core::marker::PhantomData;
+#[cfg(target_arch = "wasm32")]
+use winit::platform::web::EventLoopExtWebSys;
 use winit::{
     application::ApplicationHandler,
     dpi::PhysicalSize,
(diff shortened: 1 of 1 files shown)


crates/bevy_winit/src/state.rs
```diff
@@ -858,16 +860,23 @@ pub fn winit_runner<T: Event>(mut app: App) -> AppExit {
...
+        if #[cfg(target_arch = "wasm32")] {
+            event_loop.spawn_app(runner_state);
+            AppExit::Success
+        } else {
+            if let Err(err) = event_loop.run_app(&mut runner_state) {
+                error!("winit event loop returned an error: {err}");
+            }
+            // If everything is working correctly then the event loop only exits after it's sent an exit code.
+            runner_state.app_exit.unwrap_or_else(|| {
+                error!("Failed to receive a app exit code! This is a bug");
```
mnmaita, line 875:
```suggestion
                error!("Failed to receive an app exit code! This is a bug");
```

Typo here.

BenjaminBrienen (approved):
Maybe "this is a bug" could be improved to recommend whether to open an issue in bevy or winit (whichever is relevant, idk)

PR #16633, opened 2024-12-03
Add `no_std` support to `bevy_color`

bushrat011899:
# Objective

- Contributes to #15460

## Solution

- Added the following new features: 
  - `std` (default)
  - `alloc`
  - `encase` (default)
  - `libm`

## Testing

- Added to `compile-check-no-std` CI command

## Notes

- `ColorCurve` requires `alloc` due to how the underlying `EvenCore` type works.
- `Srgba::to_hex` requires `alloc` to return a `String`.
- This was otherwise a _very_ simple change

### DIFF
--- a/crates/bevy_color/src/lib.rs
+++ b/crates/bevy_color/src/lib.rs
@@ -4,6 +4,7 @@
     html_logo_url = "https://bevyengine.org/assets/icon.png",
     html_favicon_url = "https://bevyengine.org/assets/icon.png"
 )]
+#![cfg_attr(not(feature = "std"), no_std)]
 
 //! Representations of colors in various color spaces.
 //!
@@ -89,8 +90,12 @@
 //! println!("Hsla: {:?}", hsla);
 //! ```
 
+#[cfg(feature = "alloc")]
+extern crate alloc;
+
 mod color;
 pub mod color_difference;
+#[cfg(feature = "alloc")]
 mod color_gradient;
 mod color_ops;
 mod color_range;
@@ -121,6 +126,7 @@ pub mod prelude {
 }
 
 pub use color::*;
+#[cfg(feature = "alloc")]
 pub use color_gradient::*;
 pub use color_ops::*;
 pub use color_range::*;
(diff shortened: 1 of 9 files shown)


viridia (approved):

crates/bevy_color/src/lib.rs
```diff
@@ -4,6 +4,7 @@
     html_logo_url = "https://bevyengine.org/assets/icon.png",
     html_favicon_url = "https://bevyengine.org/assets/icon.png"
 )]
+#![cfg_attr(not(feature = "std"), no_std)]
```
IQuick143, line 7: What does this line do?

IQuick143 (approved):

crates/bevy_color/src/lib.rs
bushrat011899: It removes the standard library if you _don't_ have the `std` feature enabled. The reason for the double negative is Rust features are supposed to be additive. Adding the `std` feature adds the standard library.

IQuick143: Ah, so it resolves to #[no_std] when the std feature is not enabled.
I was confused what the attribute was relating to.

BenjaminBrienen (approved):

TimJentzsch (approved):
I have to say, I'm a bit surprised that stuff like rounding numbers is not available in `core` and has to be imported from `bevy_math`.

But the changes look fine to me :)

bushrat011899:
> I have to say, I'm a bit surprised that stuff like rounding numbers is not available in `core` and has to be imported from `bevy_math`.

Yeah they were in `core` in early versions of Rust, but have since been removed. The reason was they depended on `libm` for their implementation, and they wanted `core` to be as slim as possible. Since `libm` is also a `no_std` library with pretty much perfect platform support it isn't really that big of a change. I'd previously even discussed the idea of adding an extension trait in `bevy_math` that would make this change entirely transparent. The idea was eventually scrapped, as it's probably a good thing to highlight which maths operations are "problematic".

Issue #16578, opened 2024-11-30
Option to toggle/revert to 0.14 behaviour of Fallible SystemParams


TheBiochemic:
## What problem does this solve or what need does it fill?

In bigger Projects there is already a lot going on in the Terminal during runtime. If now forgetting to add some Resource etc. generates only logging (or nothing at all), that something won't happen, but the app otherwise launching and working often entails debugging that wouldn't be necessary in the first place, if it would just crash the program with the log message as the panic message, as it was historically.

If a Resource etc. is missing, it's supposed to interrupt my workflow, which is why i consider the old behavior a safeguard.

## What solution would you like?

Some Sort of functionality to toggle the behavior globally. It could be one of these things:
- A Flag to set it to panic, log or none globally by default, with the option to override it, as seen in 0.15 (could be nice, since it could be disabled for release builds. This could be problematic with other parts of bevy potentially).
- An Option in **DefaultPlugins** (That defines the default behavior for all Systems. Same Problem as first Option)
- An Option to set the Default behavior on a Plugin level (All systems added via this plugin have the following behavior of none, logging once, loggin always, panic)
- Catching nonworking SystemParams (like a missing Resource) in a System without defined behaviour (not using .never_param_warn(), ...) during compile time and throwing a hard error (i'm not sure how possible that would be, if at all)

The Last option would be personally my favourite.

## What alternative(s) have you considered?

Not upgrading to 0.15 just yet, until some interface is available to toggle to this or a similarly obvious behavior.

## Additional context

i'm talking specifically about the feature mentioned in https://bevyengine.org/news/bevy-0-15/#fallible-system-parameters

mockersf:
I think panics by default should be re-introduced

catching it at compile time is impossible as some of the system parameters can be made valid/invalid at runtime

TheBiochemic:
catching at compile time not being possible makes sense, i think it was a suggestion from the bevy discord.

Restoring the previous behaviour would be great yeah. I would say the panicking along side the additional behaviour of logging/none is probably a good compromise.

I guess one of the chores might be to go through the codebase and make sure, that the Systems, that shouldn't panic on missing/unloaded SystemParams need to have the `.never_param_warn()`, etc.

I don't really know, what implications a Plugin wide behavior override mechanic would have, but it might help with that above chore.

seivan:
If it's not possible to catch at compile time, then it needs to be possible to catch at boot. 
[...]

PR #16638, opened 2024-12-03
Set panic as default fallible system param behavior

MiniaczQ:
# Objective

Fixes: #16578

## Solution

This is a patch fix, proper fix requires a breaking change.

Added `Panic` enum variant and using is as the system meta default.
Warn once behavior can be enabled same way disabling panic (originally disabling wans) is.

To fix an issue with the current architecture, where **all** combinator system params get checked together,
combinator systems only check params of the first system.
This will result in old, panicking behavior on subsequent systems and will be fixed in 0.16.

## Testing

Ran unit tests and `fallible_params` example.


MiniaczQ:
Naive approach comes with some issues, I'm not sure when I'll have time to truly sit down and work on this, so feel free to take over as needed

alice-i-cecile:
Can you say a bit more about the issues you encountered?

MiniaczQ:
The main issue is stuff that was fixed in the `System::run() -> Option<S::Out>` rework (didn't land in 0.15).
In a combination system, like `a.and(b)` or `a.pipe(b)` we're running param checks for *all* systems.
This results in `resource_exists::<MyRes>.and(do_stuff)` failing params, because `do_stuff` doesn't have `MyRes` even though it wouldn't run because of failed condition.
When we change default fail behavior to panic, this causes crashes.

I've had time to think about it since this PR.
We didn't ship a complete feature, so I think we should be fine to downscope it.
Since combination systems are incomplete, let's just ensure the first system (or *the* system, if no combinations are used) works correctly with fallible params.
In 0.16 we can ship the correct param checking behavior, where we do it directly before running a system.

alice-i-cecile:
I'm okay with that as the behavior for now :)

crates/bevy_ecs/src/schedule/executor/mod.rs
```diff
@@ -321,15 +321,11 @@ mod tests {
         schedule.set_executor_kind(executor);
         schedule.add_systems(
             (
-                // Combined systems get skipped together.
-                (|mut commands: Commands| {
-                    commands.insert_resource(R1);
-                })
-                .pipe(|_: In<()>, _: Res<R1>| {}),
```
MiniaczQ, line 328: As agreed, this is no longer supported in this scope

alice-i-cecile (approved):

crates/bevy_ecs/src/system/combinator.rs
```diff
@@ -214,7 +214,7 @@ where
     #[inline]
     unsafe fn validate_param_unsafe(&mut self, world: UnsafeWorldCell) -> bool {
         // SAFETY: Delegate to other `System` implementations.
-        unsafe { self.a.validate_param_unsafe(world) && self.b.validate_param_unsafe(world) }
+        unsafe { self.a.validate_param_unsafe(world) }
```
andriyDev, line 217: Why is this ok? Surely we should check that both params are valid?

crates/bevy_ecs/src/system/combinator.rs
```diff
@@ -433,7 +433,7 @@ where
 
     unsafe fn validate_param_unsafe(&mut self, world: UnsafeWorldCell) -> bool {
         // SAFETY: Delegate to other `System` implementations.
-        unsafe { self.a.validate_param_unsafe(world) && self.b.validate_param_unsafe(world) }
+        unsafe { self.a.validate_param_unsafe(world) }
```
andriyDev, line 436: Same here

crates/bevy_ecs/src/system/combinator.rs
andriyDev: Ah I should have read above in the PR discussion, can you just add a to-do or at least a comment explanation why we've decided to do this?

bushrat011899 (approved):

BenjaminBrienen (approved):

mockersf:
https://github.com/MiniaczQ/bevy/pull/1 should get the example working again

Danielmelody:
This is actually a breaking change from 0.15.0, as some system with Single<..> query crashes after a patch update to 0.15.1. Better to intro this in 0.16.0,

PR #16640, opened 2024-12-04
`OverflowClipBox::MarginBox`

ickshonpe:
# Objective

Add support for clipping content overflowing a UI node's margin box.

## Solution

* Add a new field `margin` to `ComputedNode` which is updated in `ui_layout_system` from the margins computed by Taffy.
* Add a `MarginBox` variant to the `OverflowClipBox` enum.
* In `update_clipping` clip to the margin box when a node with `OverflowClipBox::MarginBox` is found.

## Testing
I added an item demonstrating `OverflowClipBox::MarginBox` to the `overflow_clip_margin` example:
```
cargo run --example overflow_clip_margin
```

margin-box

### DIFF
--- a/crates/bevy_sprite/src/texture_slice/border_rect.rs
+++ b/crates/bevy_sprite/src/texture_slice/border_rect.rs
@@ -1,3 +1,5 @@
+use std::ops::Neg;
+
 use bevy_reflect::Reflect;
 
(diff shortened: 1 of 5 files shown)


JMS55 (approved):

ickshonpe:
Probably not useful.

PR #16641, opened 2024-12-04
Fix the `texture_binding_array`, `specialized_mesh_pipeline`, and `custom_shader_instancing` examples after the bindless change.

pcwalton:
The bindless PR (#16368) broke some examples:

* `specialized_mesh_pipeline` and `custom_shader_instancing` failed because they expect to be able to render a mesh with no material, by overriding enough of the render pipeline to be able to do so. This PR fixes the issue by restoring the old behavior in which we extract meshes even if they have no material.

* `texture_binding_array` broke because it doesn't implement `AsBindGroup::unprepared_bind_group`. This was tricky to fix because there's a very good reason why `texture_binding_array` doesn't implement that method: there's no sensible way to do so with `wgpu`'s current bindless API, due to its multiple levels of borrowed references. To fix the example, I split `MaterialBindGroup` into `MaterialBindlessBindGroup` and `MaterialNonBindlessBindGroup`, and allow direct custom implementations of `AsBindGroup::as_bind_group` for the latter type of bind groups. To opt in to the new behavior, return the `AsBindGroupError::CreateBindGroupDirectly` error from your `AsBindGroup::unprepared_bind_group` implementation, and Bevy will call your custom `AsBindGroup::as_bind_group` method as before.

## Migration Guide

* Bevy will now unconditionally call `AsBindGroup::unprepared_bind_group` for your materials, so you must no longer panic in that function. Instead, return the new `AsBindGroupError::CreateBindGroupDirectly` error, and Bevy will fall back to calling `AsBindGroup::as_bind_group` as before.

### DIFF
--- a/crates/bevy_pbr/src/material_bind_groups.rs
+++ b/crates/bevy_pbr/src/material_bind_groups.rs
@@ -38,7 +38,7 @@ where
     M: Material,
 {
     /// The data that the allocator keeps about each bind group.
-    pub bind_groups: Vec<MaterialBindGroup<M>>,
+    bind_groups: Vec<MaterialBindGroup<M>>,
 
     /// Stores IDs of material bind groups that have at least one slot
     /// available.
@@ -60,7 +60,22 @@ where
 }
 
 /// Information that the allocator keeps about each bind group.
-pub struct MaterialBindGroup<M>
+pub enum MaterialBindGroup<M>
+where
+    M: Material,
+{
+    /// Information that the allocator keeps about each bind group with bindless
+    /// textures in use.
+    Bindless(MaterialBindlessBindGroup<M>),
+
+    /// Information that the allocator keeps about each bind group for which
+    /// bindless textures are not in use.
+    NonBindless(MaterialNonBindlessBindGroup<M>),
+}
+
+/// Information that the allocator keeps about each bind group with bindless
+/// textures in use.
+pub struct MaterialBindlessBindGroup<M>
 where
     M: Material,
 {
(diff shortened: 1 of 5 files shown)


crates/bevy_pbr/src/material_bind_groups.rs
```diff
@@ -235,15 +296,156 @@ impl<M> MaterialBindGroup<M>
...
+
+    /// Fills a slot directly with a custom bind group.
+    ///
+    /// This is only a meaningful operation for non-bindless bind groups. It's
+    /// rarely used, but see the `texture_binding_array` example for an example
+    /// demonstrating how this feature might see use in practice.
+    fn init_custom(&mut self, bind_group: BindGroup, extra_data: M::Data) {
+        match *self {
+            MaterialBindGroup::Bindless(_) => {
+                error!("Custom bind groups aren't supported in bindless mode");
```
Elabajaba, line 357: Dumb question (not asking for any changes), does limiting custom bind groups to non-bindless mean eg. custom materials (thinking about either an extended material, or maybe something like a full `StandardMaterial` toon shader replacement?) won't benefit from bindless currently?

Elabajaba (approved):
Examples are working on my system (windows amd 6800xt, tested dx12 and vulkan), code looks good, comments are great as always.

crates/bevy_pbr/src/material_bind_groups.rs
pcwalton: Bindless works for any material as long as you write `#[bindless]` on it; there's nothing specific to `StandardMaterial`. (Actually, `StandardMaterial` doesn't benefit from bindless yet; that's #16644.)

The restriction you quoted only applies in very unusual circumstances, basically only in the `custom_shader_instancing` example.

JMS55 (approved):

PR #16642, opened 2024-12-04
Turn `apply_deferred` into a ZST System

ItsDoot:
# Objective

- Required by #16622 due to differing implementations of `System` by `FunctionSystem` and `ExclusiveFunctionSystem`.
- Optimize the memory usage of instances of `apply_deferred` in system schedules.

## Solution

By changing `apply_deferred` from being an ordinary system that ends up as an `ExclusiveFunctionSystem`, and instead into a ZST struct that implements `System` manually, we save ~320 bytes per instance of `apply_deferred` in any schedule.

## Testing

- All current tests pass.

---

## Migration Guide

- If you were previously calling the special `apply_deferred` system via `apply_deferred(world)`, don't.

### DIFF
--- a/crates/bevy_ecs/src/schedule/executor/mod.rs
+++ b/crates/bevy_ecs/src/schedule/executor/mod.rs
@@ -2,6 +2,9 @@ mod multi_threaded;
 mod simple;
 mod single_threaded;
 
+use alloc::borrow::Cow;
+use core::any::TypeId;
+
 pub use self::{
     multi_threaded::{MainThreadExecutor, MultiThreadedExecutor},
     simple::SimpleExecutor,
@@ -11,9 +14,13 @@ pub use self::{
 use fixedbitset::FixedBitSet;
 
 use crate::{
-    schedule::{BoxedCondition, NodeId},
-    system::BoxedSystem,
-    world::World,
+    archetype::ArchetypeComponentId,
+    component::{ComponentId, Tick},
+    prelude::{IntoSystemSet, SystemSet},
+    query::Access,
+    schedule::{BoxedCondition, InternedSystemSet, NodeId, SystemTypeSet},
+    system::{BoxedSystem, System, SystemIn},
+    world::{unsafe_world_cell::UnsafeWorldCell, DeferredWorld, World},
 };
 
 /// Types that can run a [`SystemSchedule`] on a [`World`].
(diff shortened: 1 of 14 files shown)


crates/bevy_ecs/src/schedule/executor/mod.rs
```diff
@@ -100,32 +107,123 @@ impl SystemSchedule {
...
+    }
+
+    fn archetype_component_access(&self) -> &Access<ArchetypeComponentId> {
+        // This system accesses no archetype components.
+        const { &Access::new() }
+    }
+
+    fn is_send(&self) -> bool {
+        false
+    }
```
ItsDoot, line 164: I would like to comment why this needs to return `false`, however I'm not sure why. If set to `true`, it fails this test if ran with `multithreaded` enabled:
https://github.com/bevyengine/bevy/blob/5ecff1e2dc772a4780b6c1e6c2783a8f53095e48/crates/bevy_ecs/src/system/mod.rs#L462-L463

crates/bevy_ecs/src/schedule/executor/mod.rs
```diff
@@ -100,32 +107,123 @@ impl SystemSchedule {
...
+/// [`Deferred`]: crate::system::Deferred
+/// [`Commands`]: crate::prelude::Commands
+/// [has deferred buffers]: crate::system::System::has_deferred
+/// [`PipeSystem`]: crate::system::PipeSystem
+/// [`Schedule`]: super::Schedule
 #[doc(alias = "apply_system_buffers")]
-#[allow(unused_variables)]
-pub fn apply_deferred(world: &mut World) {}
+#[allow(unused_variables, non_camel_case_types)]
+pub struct apply_deferred;
```
ItsDoot, line 136: I decided to maintain compatibility and not rename it to `ApplyDeferred`. Should I instead rename it and add a deprecated alias for `apply_deferred`?

bushrat011899: I think giving it a proper name of `ApplyDeferred` is better. Include the type-alias for now, but I wouldn't be surprised to see it axed in favour of a breaking change.

crates/bevy_ecs/src/schedule/executor/mod.rs
```diff
@@ -100,32 +107,124 @@ impl SystemSchedule {
...
+/// [`System::apply_deferred`]: crate::system::System::apply_deferred
+/// [`Deferred`]: crate::system::Deferred
+/// [`Commands`]: crate::prelude::Commands
+/// [has deferred buffers]: crate::system::System::has_deferred
+/// [`PipeSystem`]: crate::system::PipeSystem
+/// [`Schedule`]: super::Schedule
 #[doc(alias = "apply_system_buffers")]
-#[allow(unused_variables)]
-pub fn apply_deferred(world: &mut World) {}
+#[allow(unused_variables, non_camel_case_types)]
```
bushrat011899, line 135: I believe `unused_variables` isn't required anymore, since that was in reference to the `world: &mut World` parameter.

bushrat011899 (approved):
Nice! I'm in favour of a breaking change to rename this `ApplyDeferred`. Users aren't supposed to actually call this as a function anyway, and giving it a PascalCase name should help highlight to users that it isn't a normal system.

Memory saving is also nice, especially since this system will be reused a lot in a typical application.

crates/bevy_ecs/src/schedule/executor/mod.rs
ItsDoot: The dream of having `#[expect()]` much earlier in rust 😅

crates/bevy_ecs/src/schedule/executor/mod.rs
urben1680: Exclusive systems are not send either, which the original `apply_deferred` was, see:
https://docs.rs/bevy_ecs/0.15.0/src/bevy_ecs/system/exclusive_function_system.rs.html#93

> exclusive systems should have access to non-send resources the executor runs exclusive systems on the main thread, so this field reflects that constraint

chescock (approved):

crates/bevy_ecs/src/schedule/executor/mod.rs
ItsDoot: Yea, but `apply_deferred` itself doesn't need access to anything, so why does it need to be non-send?

ItsDoot:
I've gone ahead and renamed `apply_deferred` to `ApplyDeferred` and added a deprecated alias for the old identifier. It's not a big lift on our end to include it (4 lines), so for the odd dev still using `apply_deferred` directly, this will give them an immediate actionable solution.

PR #16644, opened 2024-12-04
Make `StandardMaterial` bindless.

pcwalton:
This commit makes `StandardMaterial` use bindless textures, as implemented in PR #16368. Non-bindless mode, as used for example in Metal and WebGL 2, remains fully supported via a plethora of `#ifdef BINDLESS` preprocessor definitions.

Unfortunately, this PR introduces quite a bit of unsightliness into the PBR shaders. This is a result of the fact that WGSL supports neither passing binding arrays to functions nor passing individual *elements* of binding arrays to functions, except directly to texture sample functions. Thus we're unable to use the `sample_texture` abstraction that helped abstract over the meshlet and non-meshlet paths. I don't think there's anything we can do to help this other than to suggest improvements to upstream Naga.

### DIFF
--- a/crates/bevy_pbr/src/extended_material.rs
+++ b/crates/bevy_pbr/src/extended_material.rs
@@ -155,14 +155,32 @@ impl<B: Material, E: MaterialExtension> AsBindGroup for ExtendedMaterial<B, E> {
         layout: &BindGroupLayout,
         render_device: &RenderDevice,
         (base_param, extended_param): &mut SystemParamItem<'_, '_, Self::Param>,
+        mut force_no_bindless: bool,
     ) -> Result<UnpreparedBindGroup<Self::Data>, AsBindGroupError> {
+        // Only allow bindless mode if both the base material and the extension
+        // support it.
+        force_no_bindless = force_no_bindless
+            || B::BINDLESS_SLOT_COUNT.is_none()
+            || E::BINDLESS_SLOT_COUNT.is_none();
+
         // add together the bindings of the base material and the user material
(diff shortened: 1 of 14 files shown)


JMS55 (approved):
Oh boy, absolutely terrible code 😅. We are unlikely to get any improvements to this in naga anytime soon afaik. I think we might want to look into writing a simple macro in naga_oil or bevy for texture sampling.

pcwalton:
Yeah, I'd love a macro in `naga_oil`. We could do a lot with macros, not just this. For example, we could abstract over bindless/non-bindless declarations.

kristoff3r (approved):
My eyes might have glazed over at some point, but looks correct and tested pbr example on linux+webgl2.

I agree we need macros or similar very soon, the pbr shaders weren't pretty before but they're getting borderline unreadable except for mechanical changes like this.

alice-i-cecile:
Testing this via an [example run](https://github.com/TheBevyFlock/bevy-example-runner/actions/runs/12248580234) then merging. I'm not pleased with the reduced clarity of the shaders, but I agree that this is not the place to solve it (nor should this be blocked on that).

pcwalton:
Example run is green.

rparrett:
The example run completed successfully, but shows differences in `fog_volumes` and `volumetric_fog` on macOS and `meshlet` on linux which seem to be breakage.

https://pixel-eagle.com/project/B25A040A-A980-4602-B90C-D480AB84076D?filter=PR-16644

PR #16646, opened 2024-12-04
Use new get_unchecked methods in QuerySortedIter

Victoronz:
# Objective

In `QuerySortedIter::fetch_next`, `Entites::get`. `Archetypes::get`, `Tables::get` are all guaranteed to return correctly, but still do bounds checks/further checks.

## Solution

Introduce `get_unchecked` methods on `Entities`, `Archetypes`, and `Tables`, and use those.

### DIFF
--- a/crates/bevy_ecs/src/archetype.rs
+++ b/crates/bevy_ecs/src/archetype.rs
@@ -868,6 +868,17 @@ impl Archetypes {
         self.archetypes.get(id.index())
     }
 
+    // Fetches an immutable reference to an Archetype using its
+    // ID, without checking whether it exists.
+    //
+    // # Safety
+    //
+    // A corresponding archetype must exist.
+    #[inline]
+    pub(crate) unsafe fn get_unchecked(&self, id: ArchetypeId) -> &Archetype {
+        self.archetypes.get_unchecked(id.index())
+    }
+
     /// # Panics
     ///
     /// Panics if `a` and `b` are equal.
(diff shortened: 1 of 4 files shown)


Victoronz:
Should these methods be public?

Victoronz:
It is the `slice::get_unchecked` that improves the perf here, plus avoiding the additional checks in `Entities::get`.
Not using `debug_unchecked_unwrap` is just a consequence of no longer handling options!

chescock:
> It is the `slice::get_unchecked` that improves the perf here

Right, but I would expect `.get().debug_checked_unwrap()` to compile to the same thing as `.get_unchecked()`.  The optimizer will see the `core::hint::unreachable_unchecked()` on the `None` branch and eliminate any check that causes a `None` to be returned, reasoning that it knows the result of those checks without making them.  

... Hmm, I bet I can check this with `cargo-show-asm`...  

```rust
#[inline(never)]
pub fn asm_test(entities: &Entities, entity: Entity) -> EntityLocation {
    unsafe { entities.get(entity).debug_checked_unwrap() }
}
```

```
> cargo asm -p bevy_ecs --features bevy_reflect asm_test --lib --simplify

    Finished `release` profile [optimized] target(s) in 0.40s
bevy_ecs::entity::asm_test:
        mov rax, rcx
        mov ecx, r8d
        mov rdx, qword ptr [rdx + 8]
        lea rcx, [rcx + 4*rcx]
        movups xmm0, xmmword ptr [rdx + 4*rcx + 4]
        movups xmmword ptr [rax], xmm0
        ret
```

No branches!

Victoronz:
ohhh, I see I misunderstood `debug_checked_unwrap`, and `Entites::get` only has one path to `Some`!

PR #16655, opened 2024-12-05
Do not attempt to send screenshots to Pixel Eagle without a token

LikeLakers2:
# Objective
Running Github Actions on forks helps users to reduce the amount of CI errors they get before submitting a PR. However, due to how workflows are set up on the Bevy repository, this can result in errors occurring for jobs that may not be related to their PR - in this case, uploading screenshots to Pixel Eagle.

## Solution
The Pixel Eagle workflow is skipped if we aren't running on the Bevy repository.

If we are on the Bevy repository, or the user has set it to run elsewhere, we check if the `PIXELEAGLE_TOKEN` secret is set. If it isn't, we skip uploading screenshots to Pixel Eagle.

* Artifacts still continue to generate, in case the user needs them.
* In the event that the Pixel Eagle workflow runs, but the `PIXELEAGLE_TOKEN` secret isn't set, we generate a step summary that notifies the user of why it was skipped. https://github.com/LikeLakers2/bevy/actions/runs/12173329006/attempts/1#summary-33953502068 for an example.

## Testing
Lots. And lots. Of trying to get Github Actions to work with me.

### DIFF
--- a/.github/workflows/send-screenshots-to-pixeleagle.yml
+++ b/.github/workflows/send-screenshots-to-pixeleagle.yml
@@ -24,7 +24,7 @@ jobs:
     name: Send screenshots to Pixel Eagle
     runs-on: ubuntu-24.04
     steps:
-      - name: Notify user on non-existant token
+      - name: Notify user on non-existent token
         if: ${{ env.PIXELEAGLE_TOKEN_EXISTS == 'false' }}
         run: |
           echo "The PIXELEAGLE_TOKEN secret does not exist, so uploading screenshots to Pixel Eagle was skipped." >> $GITHUB_STEP_SUMMARY
--- a/.github/workflows/send-screenshots-to-pixeleagle.yml
+++ b/.github/workflows/send-screenshots-to-pixeleagle.yml
@@ -25,6 +25,10 @@ jobs:
   send-to-pixel-eagle:
     name: Send screenshots to Pixel Eagle
     runs-on: ubuntu-24.04
+    # Pixel Eagle is irrelevant for most forks, even of those that allow workflows to run. Thus, we
+    # disable this job for any forks. Any forks where Pixel Eagle is relevant can comment out the
+    # `if:` conditional below.
+    if: ${{ github.repository == 'bevyengine/bevy' }}
     steps:
       - name: Notify user on non-existent token
         if: ${{ ! fromJSON(env.PIXELEAGLE_TOKEN_EXISTS) }}
--- a/.github/workflows/send-screenshots-to-pixeleagle.yml
+++ b/.github/workflows/send-screenshots-to-pixeleagle.yml
@@ -16,18 +16,27 @@ on:
         required: true
         type: string
 
+env:
+  PIXELEAGLE_TOKEN_EXISTS: ${{ secrets.PIXELEAGLE_TOKEN != '' }}
+
 jobs:
   send-to-pixel-eagle:
     name: Send screenshots to Pixel Eagle
     runs-on: ubuntu-24.04
     steps:
+      - name: Notify user on non-existant token
+        if: ${{ env.PIXELEAGLE_TOKEN_EXISTS == 'false' }}
+        run: |
+          echo "The PIXELEAGLE_TOKEN secret does not exist, so uploading screenshots to Pixel Eagle was skipped." >> $GITHUB_STEP_SUMMARY
       
       - name: Download artifact
+        if: ${{ env.PIXELEAGLE_TOKEN_EXISTS != 'false' }}
         uses: actions/download-artifact@v4
         with:
           pattern: ${{ inputs.artifact }}
       
       - name: Send to Pixel Eagle
+        if: ${{ env.PIXELEAGLE_TOKEN_EXISTS != 'false' }}
         env:
           project: B04F67C0-C054-4A6F-92EC-F599FEC2FD1D
         run: |
--- a/.github/workflows/send-screenshots-to-pixeleagle.yml
+++ b/.github/workflows/send-screenshots-to-pixeleagle.yml
@@ -27,18 +27,18 @@ jobs:
     runs-on: ubuntu-24.04
     steps:
       - name: Notify user on non-existent token
-        if: ${{ env.PIXELEAGLE_TOKEN_EXISTS == 'false' }}
+        if: ${{ ! fromJSON(env.PIXELEAGLE_TOKEN_EXISTS) }}
         run: |
           echo "The PIXELEAGLE_TOKEN secret does not exist, so uploading screenshots to Pixel Eagle was skipped." >> $GITHUB_STEP_SUMMARY
       
       - name: Download artifact
-        if: ${{ env.PIXELEAGLE_TOKEN_EXISTS == 'true' }}
+        if: ${{ fromJSON(env.PIXELEAGLE_TOKEN_EXISTS) }}
         uses: actions/download-artifact@v4
         with:
           pattern: ${{ inputs.artifact }}
       
       - name: Send to Pixel Eagle
-        if: ${{ env.PIXELEAGLE_TOKEN_EXISTS == 'true' }}
+        if: ${{ fromJSON(env.PIXELEAGLE_TOKEN_EXISTS) }}
         env:
           project: B04F67C0-C054-4A6F-92EC-F599FEC2FD1D
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/send-screenshots-to-pixeleagle.yml
```diff
@@ -16,18 +16,29 @@ on:
...
+  # suggests a workaround: Putting the secret in an environment variable, and checking that instead.
+  PIXELEAGLE_TOKEN_EXISTS: ${{ secrets.PIXELEAGLE_TOKEN != '' }}
+
 jobs:
   send-to-pixel-eagle:
     name: Send screenshots to Pixel Eagle
     runs-on: ubuntu-24.04
     steps:
+      - name: Notify user on non-existent token
+        if: ${{ env.PIXELEAGLE_TOKEN_EXISTS == 'false' }}
```
LikeLakers2, line 30: In case you're thinking "why is this a comparison to a stringified boolean?": environment variables are strings, even those that are booleans.

It took me way too long to figure this out.

.github/workflows/send-screenshots-to-pixeleagle.yml
```diff
@@ -16,18 +16,29 @@ on:
...
+  # suggests a workaround: Putting the secret in an environment variable, and checking that instead.
+  PIXELEAGLE_TOKEN_EXISTS: ${{ secrets.PIXELEAGLE_TOKEN != '' }}
+
 jobs:
   send-to-pixel-eagle:
     name: Send screenshots to Pixel Eagle
     runs-on: ubuntu-24.04
     steps:
+      - name: Notify user on non-existent token
+        if: ${{ ! fromJSON(env.PIXELEAGLE_TOKEN_EXISTS) }}
```
LikeLakers2, line 30: In case you're curious:

* Environment variables are stored as strings. However, the strings `'true'` and `'false'` are valid JSON, and return booleans when converted to JSON.
* I wish I could put this `if:` conditional on the job itself - but those don't allow access to environment variables for some reason.

BenjaminBrienen (approved):

.github/workflows/send-screenshots-to-pixeleagle.yml
```diff
@@ -16,18 +16,29 @@ on:
...
+env:
+  # Unfortunately, we can't check secrets in `if:` conditionals. However, Github's own documentation
+  # suggests a workaround: Putting the secret in an environment variable, and checking that instead.
+  PIXELEAGLE_TOKEN_EXISTS: ${{ secrets.PIXELEAGLE_TOKEN != '' }}
+
 jobs:
   send-to-pixel-eagle:
     name: Send screenshots to Pixel Eagle
     runs-on: ubuntu-24.04
     steps:
```
BD103, line 28:
```suggestion
  send-to-pixel-eagle:
    name: Send screenshots to Pixel Eagle
    runs-on: ubuntu-24.04
    if: ${{ github.repository == "bevyengine/bevy" }}
    steps:
```

BD103 (approved):
Your current approach works, which is why I'm approving it, but we can improve it by checking what repository this is being run from.

By adding `if: ${{ github.repository == "bevyengine/bevy" }}`, we can skip saving the secret to an environmental variable and the various other checks.

.github/workflows/send-screenshots-to-pixeleagle.yml
LikeLakers2: So, I'm not actually sure whether to commit this suggestion.

My intention is to allow forks to run the Pixel Eagle job if they desire. That's the reason I chose to lock the Pixel Eagle job behind needing a token, rather than behind what repository it's on.

However, after having slept between sending this PR and now, I'm not actually sure there's any reason a fork would want to run this.

That said, I'm also not sure why we would lock it solely to the Bevy repository.

@BD103 What do you think?

LikeLakers2: So, I'm going to apply this suggestion - but I'm going to reword the comment first, as I don't think it fits with my intention to allow the Pixel Eagle job to run on forks.

.github/workflows/send-screenshots-to-pixeleagle.yml
```diff
@@ -16,18 +16,33 @@ on:
         required: true
         type: string
 
+env:
+  # Unfortunately, we can't check secrets in `if:` conditionals. However, Github's own documentation
+  # suggests a workaround: Putting the secret in an environment variable, and checking that instead.
+  PIXELEAGLE_TOKEN_EXISTS: ${{ secrets.PIXELEAGLE_TOKEN != '' }}
```
BD103, line 22:
```suggestion
```

.github/workflows/send-screenshots-to-pixeleagle.yml
```diff
@@ -16,18 +16,33 @@ on:
...
   send-to-pixel-eagle:
     name: Send screenshots to Pixel Eagle
     runs-on: ubuntu-24.04
+    # Pixel Eagle is irrelevant for most forks, even of those that allow workflows to run. Thus, we
+    # disable this job for any forks. Any forks where Pixel Eagle is relevant can comment out the
+    # `if:` conditional below.
+    if: ${{ github.repository == 'bevyengine/bevy' }}
     steps:
+      - name: Notify user on non-existent token
+        if: ${{ ! fromJSON(env.PIXELEAGLE_TOKEN_EXISTS) }}
```
BD103, line 34:
```suggestion
```

.github/workflows/send-screenshots-to-pixeleagle.yml
```diff
@@ -16,18 +16,33 @@ on:
...
+        run: |
+          echo "The PIXELEAGLE_TOKEN secret does not exist, so uploading screenshots to Pixel Eagle was skipped." >> $GITHUB_STEP_SUMMARY
       
       - name: Download artifact
+        if: ${{ fromJSON(env.PIXELEAGLE_TOKEN_EXISTS) }}
```
BD103, line 39:
```suggestion
```

.github/workflows/send-screenshots-to-pixeleagle.yml
```diff
@@ -16,18 +16,33 @@ on:
...
         uses: actions/download-artifact@v4
         with:
           pattern: ${{ inputs.artifact }}
       
       - name: Send to Pixel Eagle
+        if: ${{ fromJSON(env.PIXELEAGLE_TOKEN_EXISTS) }}
```
BD103, line 45:
```suggestion
```

BD103 (changes requested):
Could you please remove the old logic, since it no longer has any effect? (Also I'd rather avoid putting secrets in environmental variables if possible.)

LikeLakers2:
@BD103
> Could you please remove the old logic, since it no longer has any effect?
It does have an effect, though.

In the old version, if `secrets.PIXELEAGLE_TOKEN == ''`, then certain parameters to `curl` would be given an empty value. This results in an error at the command-line level.

With the logic I added, it sets an environment variable `PIXELEAGLE_TOKEN_EXISTS` to the result of `secrets.PIXELEAGLE_TOKEN != ''`. If this is `'false'`, then `secrets.PIXELEAGLE_TOKEN == ''`. With this, we can prevent some steps from running that would otherwise error out.

> (Also I'd rather avoid putting secrets in environmental variables if possible.)
I agree - and this is why the environment variable isn't the secret itself, but rather a boolean, the result of checking if `secrets.PIXELEAGLE_TOKEN != ''`.

.github/workflows/send-screenshots-to-pixeleagle.yml
LikeLakers2: If I were to make this change, I would have to remove this entire step. There is no point to running this step unless `secrets.PIXELEAGLE_TOKEN == ''`.

BD103 (approved):
Oh, you're right! I misread:

```yaml
PIXELEAGLE_TOKEN_EXISTS: ${{ secrets.PIXELEAGLE_TOKEN != '' }}
```

as

```yaml
PIXELEAGLE_TOKEN_EXISTS: ${{ secrets.PIXELEAGLE_TOKEN }}
```

so I thought the environmental variable stored the secret directly. Sorry for confusion, approved!

.github/workflows/send-screenshots-to-pixeleagle.yml
BenjaminBrienen: I think the environment variable was a superior solution because it allows forks to set that up for themselves, but checking the name of the repository is the existing pattern we use in other context.

BenjaminBrienen: I think the environment variable was a superior solution because it allows forks to set that up for themselves, but checking the name of the repository is the existing pattern we use in other contexts.

PR #16657, opened 2024-12-05
picking: disable raycast backface culling for Mesh2d

globin:
# Objective

- This fixes raycast picking with lyon
- reverse winding of 2D meshes currently results in them being rendered but not pickable as the raycast passes through the backface and would only hit "from below" 

## Solution

- Disables backface culling for Mesh2d

## Testing

- Tested picking with bevy_prototype_lyon
- Could probably use testing with Mesh3d (should not be affected) and SimplifiedMesh (no experience with that, could have the same issue if used for 2D?)

### DIFF
--- a/crates/bevy_picking/src/mesh_picking/ray_cast/mod.rs
+++ b/crates/bevy_picking/src/mesh_picking/ray_cast/mod.rs
@@ -274,9 +274,9 @@ impl<'w, 's> MeshRayCast<'w, 's> {
                     return;
                 };
 
-                let backfaces = match has_backfaces {
-                    true => Backfaces::Include,
-                    false => Backfaces::Cull,
+                let backfaces = match (has_backfaces, mesh2d.is_some()) {
+                    (false, false) => Backfaces::Cull,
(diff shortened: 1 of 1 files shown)


crates/bevy_picking/src/mesh_picking/ray_cast/mod.rs
```diff
@@ -274,9 +274,9 @@ impl<'w, 's> MeshRayCast<'w, 's> {
                     return;
                 };
 
-                let backfaces = match has_backfaces {
-                    true => Backfaces::Include,
-                    false => Backfaces::Cull,
+                let backfaces = match (has_backfaces, mesh2d.is_some()) {
```
aevyrie, line 277:
```suggestion
                // Backfaces of 2d meshes are never culled, unlike 3d mehses.
                let backfaces = match (has_backfaces, mesh2d.is_some()) {
```

aevyrie (approved):

IQuick143 (approved):

PR #16662, opened 2024-12-05
Make `Parent` and `Children` components immutable

tim-blackbird:
# Objective
Minimal changes to make the hierarchy components immutable.

### DIFF
--- a/crates/bevy_ecs/src/reflect/component.rs
+++ b/crates/bevy_ecs/src/reflect/component.rs
@@ -301,19 +301,17 @@ impl<C: Component + Reflect + TypePath> FromType<C> for ReflectComponent {
                 component.apply(reflected_component);
             },
             apply_or_insert: |entity, reflected_component, registry| {
-                if !C::Mutability::MUTABLE {
-                    let name = ShortName::of::<C>();
-                    panic!("Cannot call `ReflectComponent::apply_or_insert` on component {name}. It is immutable, and cannot modified through reflection");
-                }
-
                 // SAFETY: guard ensures `C` is a mutable component
-                if let Some(mut component) = unsafe { entity.get_mut_assume_mutable::<C>() } {
-                    component.apply(reflected_component.as_partial_reflect());
-                } else {
-                    let component = entity.world_scope(|world| {
-                        from_reflect_with_fallback::<C>(reflected_component, world, registry)
-                    });
-                    entity.insert(component);
+                match unsafe { entity.get_mut_assume_mutable::<C>() } {
+                    Some(mut component) if C::Mutability::MUTABLE => {
+                        component.apply(reflected_component.as_partial_reflect());
+                    }
+                    _ => {
+                        let component = entity.world_scope(|world| {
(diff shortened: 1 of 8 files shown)


bushrat011899 (approved):
Just some documentation that needs to be deleted now that `apply_or_insert` doesn't panic, otherwise I think this is good to go! A follow-up that adds hooks for upholding the `Parent`/`Children` relationship invariants would be the next logical step.

crates/bevy_hierarchy/src/components/mod.rs
```diff
@@ -1,5 +1,26 @@
...
+/// Private trait to consolidate unsafe calls to [`get_mut_assume_mutable`](EntityWorldMut::get_mut_assume_mutable).
+pub(crate) trait GetHierarchyComponentsMut {
+    fn get_parent_mut(&mut self) -> Option<Mut<Parent>>;
+    fn get_children_mut(&mut self) -> Option<Mut<Children>>;
+}
+
+#[expect(unsafe_code)]
+impl GetHierarchyComponentsMut for EntityWorldMut<'_> {
+    fn get_parent_mut(&mut self) -> Option<Mut<Parent>> {
+        // SAFETY: The Parent component has no invariants.
```
chescock, line 18: The doc comments for `get_mut_assume_mutable()` give the safety invariant as `` `T` must be a mutable component ``, which this safety comment doesn't satisfy.  If this is intended to be legal, then I think we need to update the documentation for that method.  

I thought the plan was that immutable components could only be updated by re-inserting them, though, and that this was not intended to be legal.

bushrat011899: I think updating the safety requirement for the `*_assume_mutable` methods is appropriate here in the interim. The ideal solution would be to use swapping to actually preserve immutability and avoid the unsafe methods, but that's a little cumbersome at the moment.

bushrat011899: "swapping" as in temporarily replacing `Children`/`Parent` with a dummy value so you can just have ownership over the component while you're editing it. Basically `core::mem::swap`. As an alternative, #16668 will add a `with_component` method which does this for you without the temporary value by directly triggering hooks/observers.

crates/bevy_ecs/src/reflect/component.rs
```diff
@@ -160,9 +160,7 @@ impl ReflectComponent {
 
     /// Uses reflection to set the value of this [`Component`] type in the entity to the given value or insert a new one if it does not exist.
     ///
-    /// # Panics
-    ///
-    /// Panics if [`Component`] is immutable.
+    /// Always inserts if [`Component`] is immutable.
```
alice-i-cecile, line 163: Good change!

crates/bevy_hierarchy/src/components/mod.rs
alice-i-cecile: I'd like to merge #16668 and use it here. This safety comment really confused me.

alice-i-cecile:
Tackled by the official relations implementation! :D

PR #16665, opened 2024-12-05
bevy_input_focus improvements (follow-up PR)

viridia:
This adds a few minor items which were left out of the previous PR:

- Added synchronization from bevy_input_focus to bevy_a11y.
- Initialize InputFocusVisible resource.
- Make `input_focus` available from `bevy` module.

I've tested this using VoiceOver on Mac OS. It works, but it needs considerable polish.

### DIFF
--- a/crates/bevy_input_focus/Cargo.toml
+++ b/crates/bevy_input_focus/Cargo.toml
@@ -10,7 +10,6 @@ keywords = ["bevy", "color"]
 rust-version = "1.76.0"
 
 [dependencies]
-bevy_a11y = { path = "../bevy_a11y", version = "0.15.0-dev", default-features = false }
 bevy_app = { path = "../bevy_app", version = "0.15.0-dev", default-features = false }
 bevy_ecs = { path = "../bevy_ecs", version = "0.15.0-dev", default-features = false }
 bevy_input = { path = "../bevy_input", version = "0.15.0-dev", default-features = false }
--- a/crates/bevy_input_focus/Cargo.toml
+++ b/crates/bevy_input_focus/Cargo.toml
@@ -10,6 +10,7 @@ keywords = ["bevy", "color"]
 rust-version = "1.76.0"
 
 [dependencies]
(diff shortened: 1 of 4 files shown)


crates/bevy_input_focus/src/lib.rs
```diff
@@ -135,6 +138,13 @@ fn dispatch_keyboard_input(
     }
 }
 
+/// System which synchronizes the a11y focus state with the input focus state.
+fn sync_a11y_focus(focus: Res<InputFocus>, mut a11y_focus: ResMut<Focus>) {
```
alice-i-cecile, line 142: This seems silly: shouldn't we just change the a11y focus to use the new resource?

alice-i-cecile (commented):
I think we should split off the bevy_a11y changes and merge the rest as trivial :)

crates/bevy_input_focus/src/lib.rs
viridia: Uh, I did it this way because my understanding was (based on our previous discussion) that we couldn't have `bevy_a11y` depend on `bevy_input_focus`.

alice-i-cecile: Hmm. Maybe I'm misremembering or maybe there was a micommunication, but bevy_a11y depending on bevy_input_focus definitely seems more correct than the reverse (like it is now).

crates/bevy_input_focus/Cargo.toml
```diff
@@ -10,6 +10,7 @@ keywords = ["bevy", "color"]
 rust-version = "1.76.0"
 
 [dependencies]
+bevy_a11y = { path = "../bevy_a11y", version = "0.15.0-dev", default-features = false }
```
alice-i-cecile, line 13: This should be cut from this PR :)

alice-i-cecile (changes requested):
Missed a change, but once that's done this LGTM and I'll merge as trivial.

alice-i-cecile (approved):

PR #16669, opened 2024-12-05
Fix scaling for UI debug overlay

### LINKED ISSUE #16666: `DebugUiPlugin` doesn't respect scale factor properly
## Bevy version

0.15

## What you did

Enable `DebugUiPlugin`'s outline drawing.

## What went wrong

Outline is twice the size of the actual nodes.

## Additional information

This seems to be introduced by https://github.com/bevyengine/bevy/pull/16375, which did not update the debug implementation here: https://github.com/bevyengine/bevy/blob/83c729162f9079b592ee7dce66df599a7e32f6e0/crates/bevy_dev_tools/src/ui_debug_overlay/mod.rs#L40-L48

ickshonpe:
# Objective

Fixes #16666

# Solution

Multiply the coordinates by the inverse of the window scale factor.

### DIFF
--- a/crates/bevy_dev_tools/src/ui_debug_overlay/mod.rs
+++ b/crates/bevy_dev_tools/src/ui_debug_overlay/mod.rs
@@ -156,7 +156,6 @@ struct OutlineParam<'w, 's> {
     children: Query<'w, 's, &'static Children>,
     nodes: Query<'w, 's, NodesQuery>,
     view_visibility: Query<'w, 's, &'static ViewVisibility>,
-    ui_scale: Res<'w, UiScale>,
 }
 
 type CameraQuery<'w, 's> = Query<'w, 's, &'static Camera, With<DebugOverlayCamera>>;
--- a/crates/bevy_dev_tools/src/ui_debug_overlay/mod.rs
+++ b/crates/bevy_dev_tools/src/ui_debug_overlay/mod.rs
@@ -15,7 +15,7 @@ use bevy_render::{
     view::{RenderLayers, VisibilitySystems},
 };
 use bevy_transform::{prelude::GlobalTransform, TransformSystem};
-use bevy_ui::{ComputedNode, DefaultUiCamera, Display, Node, TargetCamera, UiScale};
+use bevy_ui::{ComputedNode, DefaultUiCamera, Display, Node, TargetCamera};
 use bevy_utils::{default, warn_once};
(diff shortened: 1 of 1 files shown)


alice-i-cecile (approved):

JeanMertz:
I backported these changes to 0.15 in https://github.com/bevyengine/bevy/compare/v0.15.0...JeanMertz:bevy:jean/debug-ui-gizmo-size, but I cannot get the debug lines to show up. The bevy-egui-inspector also doesn't show the entities being created to draw the lines.

Need to investigate what happened, but it might just be a mistake on my side, or the fact that this doesn't backport cleanly to 0.15.

ickshonpe:
> I backported these changes to 0.15 in [v0.15.0...JeanMertz:bevy:jean/debug-ui-gizmo-size](https://github.com/bevyengine/bevy/compare/v0.15.0...JeanMertz:bevy:jean/debug-ui-gizmo-size), but I cannot get the debug lines to show up. The bevy-egui-inspector also doesn't show the entities being created to draw the lines.
> 
> Need to investigate what happened, but it might just be a mistake on my side, or the fact that this doesn't backport cleanly to 0.15.

So they worked on main without problem for you?

I could rewrite the debug visualization to use the UI renderer. It would be simpler and it would work with multiple windows and render to texture etc.

ickshonpe:
UI renderer PR: #16693

BenjaminBrienen (approved):

alice-i-cecile:
@ickshonpe @mockersf  I'd like to ship this for 0.15.1: what's the best approach now that #16693 is merged?

JeanMertz:
@alice-i-cecile i don’t believe that implementation has the same issue. I backported the pr you linked to 0.15 and haven’t had any issues with it.

alice-i-cecile:
>  or are you referring to the fact that that PR won’t be part of 0.15.1, so we still need this fix to land?

Precisely: we don't want to ship the PR that this builds on into 0.15.1 because it's breaking. But it would still be good to have a simple fix for the issue.

mockersf:
Tt will probably have to be rewritten on the patch branch. I'll check how hard it is when cherrypicking this PR, and call for help if needed

ickshonpe:
The changes are really simple so I don't think there should any problems.

alice-i-cecile:
I don't think this is impactful enough to spend the time backporting.

PR #16670, opened 2024-12-06
Turn on GPU culling automatically if the target platform supports it.

pcwalton:
This commit removes the undocumented `GpuCulling` component in favor of automatically turning GPU culling on when the platform supports it. The main reason to make GPU culling automatic is that GPU culling enables indirect mode. Indirect mode is needed for multidraw (#16427), because non-indirect multidraw doesn't exist in `wgpu`. Since multidraw is such a win for performance, when that feature is supported the small performance tax that indirect mode incurs is virtually always worth paying.

CPU culling is always used in addition to GPU culling unless the `NoCpuCulling` component is placed on the camera. This results in some amount of redundant computation on the GPU, but the overhead is negligible. I figured that the GPU time savings gained from skipping this computation when not needed didn't outweigh the costs of the added complexity that would be necessarily to implement that optimization, especially with GPU two-phase occlusion culling on the horizon.

## Migration Guide

* The `GpuCulling` component has been removed. GPU culling is now automatically enabled for all cameras if the hardware and platform support it.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -242,7 +242,7 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - name: Check for typos
-        uses: crate-ci/typos@v1.27.3
+        uses: crate-ci/typos@v1.28.2
       - name: Typos info
         if: failure()
         run: |
--- a/.github/workflows/weekly.yml
+++ b/.github/workflows/weekly.yml
@@ -10,12 +10,29 @@ on:
 env:
   CARGO_TERM_COLOR: always
 
+# The jobs listed here are intentionally skipped when running on forks, for a number of reasons:
+#
+# * Scheduled workflows run on the base/default branch, with no way (currently) to change this. On
+#   forks, the base/default branch is usually kept in sync with the main Bevy repository, meaning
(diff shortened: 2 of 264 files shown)


alice-i-cecile (approved):
Good, I think this is a better design.

JMS55 (approved):
Approved once docs are fixed (see CI).

bushrat011899 (approved):
Tested `many_cubes` in Windows 10 on a i5 1240P Framework 13 and everything appears to be working as expected. Code definitely looks cleaner too!

pcwalton:
I believe the failure is https://github.com/bevyengine/bevy/pull/15981.

rparrett:
These seem to be the finished runs:

mac:
https://pixel-eagle.com/project/b25a040a-a980-4602-b90c-d480ab84076d/run/5862/compare/5815
linux:
https://pixel-eagle.com/project/b25a040a-a980-4602-b90c-d480ab84076d/run/5863/compare/5816

Looks like `ssr`, `custom_shader_instancing`, `deferred_rendering` are probably real issues, and the rest are known flakiness.

pcwalton:
Closed in favor of #16755 and #16757.

PR #16673, opened 2024-12-06
Fix atan2 docs

dbeckwith:
# Objective

The parameter names for `bevy::math::ops::atan2` are labelled such that `x` is the first argument and `y` is the second argument, but it passes those arguments directly to [`f32::atan2`](https://doc.rust-lang.org/stable/std/primitive.f32.html#method.atan2), whose parameters are expected to be `(y, x)`. This PR changes the parameter names in the bevy documentation to use the correct order for the operation being performed. You can verify this by doing:

```rust
fn main() {
    let x = 3.0;
    let y = 4.0;
    let angle = bevy::math::ops::atan2(x, y);
    // standard polar coordinates formula
    dbg!(5.0 * angle.cos(), 5.0 * angle.sin());
}
```

This will print `(4.0, 3.0)`, which has flipped `x` and `y`. The problem is that the `atan2` function to calculate the angle was really expecting `(y, x)`, not `(x, y)`.

## Solution

I flipped the parameter names for `bevy::math::ops::atan2` and updated the documentation. I also removed references to `self` and `other` from the documentation which seemed to be copied from the `f32::atan2` documentation.

## Testing

Not really needed, you can compare the `f32::atan2` docs to the `bevy::math::ops::atan2` docs to see the problem is obvious. If a test is required I could add a short one.
## Migration Guide

I'm not sure if this counts as a breaking change, since the implementation clearly meant to use `f32::atan2` directly, so it was really just the parameter names that were wrong.

### DIFF
--- a/crates/bevy_math/src/ops.rs
+++ b/crates/bevy_math/src/ops.rs
@@ -143,7 +143,7 @@ mod std_ops {
         f32::atan(x)
     }
 
-    /// Computes the four quadrant arctangent of `self` (`y`) and `other` (`x`) in radians.
+    /// Computes the four quadrant arctangent of `y` and `x` in radians.
     ///
     /// * `x = 0`, `y = 0`: `0`
     /// * `x >= 0`: `arctan(y/x)` -> `[-pi/2, pi/2]`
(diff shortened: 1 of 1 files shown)


mweatherley (changes requested):
Good catch. We should also change it for the `libm::atan2f` version :)

IQuick143:
I think it'd be best to change in this PR as well, the libm version of this function is used a bit later in this file.

IQuick143 (approved):

pcwalton (approved):
Straightforward.

mweatherley (approved):
Thanks!

PR #16675, opened 2024-12-06
Add some basic tests for easing functions

scottmcm:
# Objective

Context for #16676

The release notes nerd-sniped me again.  I was looking at the `EaseFunction::Exponential*` definition and wondering why that specific definition was chosen.  And looking into it I noticed that it seemed wrong and there weren't any tests, so here are those tests -- which are failing as proof.

## Solution

I don't know what the best answer is here, but I think having tests is at least a good start for a discussion.

## Testing

These *are* tests 🙂

### DIFF
--- a/crates/bevy_math/src/curve/easing.rs
+++ b/crates/bevy_math/src/curve/easing.rs
@@ -459,3 +459,67 @@ impl EaseFunction {
         }
     }
 }
+
+#[cfg(test)]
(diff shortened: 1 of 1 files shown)


BenjaminBrienen:
Marking this as blocked on fixing #16676 with another PR, at which point this can be rebased and the tests should then pass.

PR #16688, opened 2024-12-06
Fix mehses -> meshes typo

Jondolf:
# Objective

The "mehses" typo introduced in a review comment [here](https://github.com/bevyengine/bevy/pull/16657#discussion_r1870834999) hurts my soul, it was merged right as I was about to comment about it :(

## Solution

Fix it :D

(also, why didn't the CI typo checker catch this?)


alice-i-cecile (approved):
Typo checker is manually curated to reduce false positives :)

PR #16691, opened 2024-12-06
More complete documentation of valid query transmutes

chescock:
# Objective

The documentation for `Query::transmute_lens` lists some allowed transmutes, but the list is incomplete.  

## Solution

Document the underlying rules for what transmutes are allowed.  

Add a longer list of examples.  Write them as doc tests to ensure that those examples are actually allowed.  

I'm assuming that anything that can be done today is intended to be supported!  If any of these examples are things we plan to prohibit in the future then we can add some warnings to that effect.

### DIFF
--- a/crates/bevy_ecs/src/system/query.rs
+++ b/crates/bevy_ecs/src/system/query.rs
@@ -1335,7 +1335,7 @@ impl<'w, 's, D: QueryData, F: QueryFilter> Query<'w, 's, D, F> {
     ///
     /// ## Panics
     ///
-    /// This will panic if `NewD` is not a subset of the original fetch `Q`
+    /// This will panic if `NewD` is not a subset of the original fetch `D`
     ///
     /// ## Example
     ///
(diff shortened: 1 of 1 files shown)


crates/bevy_ecs/src/system/query.rs
```diff
@@ -1376,19 +1376,104 @@ impl<'w, 's, D: QueryData, F: QueryFilter> Query<'w, 's, D, F> {
...
+    /// # #[derive(Component)]
+    /// # struct T;
+    /// #
+    /// # #[derive(Component)]
+    /// # struct U;
+    /// #
+    /// # #[derive(Component)]
+    /// # struct V;
+    /// #
+    /// // `&mut T` and `Mut<T>` are equivalent, `&T` and `Ref<T>` are equivalent,
```
hymm, line 1434: Ref<T> and &T aren't equivalent. They return different types. i.e. &T does just return the reference without the change ticks.

chescock: What I want to communicate is that they have equivalent access for the purposes of query transmutes.  Like, `&mut T` to `Ref<T>` is just as valid as `&mut T` to `&T`, but it doesn't feel useful to spell out all four versions of that in examples.  

I'm happy to change the wording if you have a better suggestion!  Everything I can come up with that's more precise winds up way too verbose to be clear, though.

alice-i-cecile: `&T` and `Ref<T>` access the same data, and can be substituted for the purposes of query transmutes.

PR #16692, opened 2024-12-06
Use multidraw for shadows when GPU culling is in use.

pcwalton:
This patch makes shadows use multidraw when the camera they'll be drawn to has the `GpuCulling` component. This results in a significant reduction in drawcalls; Bistro Exterior drops to 3 drawcalls for each shadow cascade.

Note that PR #16670 will remove the `GpuCulling` component, making shadows automatically use multidraw. Beware of that when testing this patch; before #16670 lands, you'll need to manually add `GpuCulling` to your camera in order to see any performance benefits.

### DIFF
--- a/crates/bevy_pbr/src/render/light.rs
+++ b/crates/bevy_pbr/src/render/light.rs
@@ -15,6 +15,7 @@ use bevy_render::{
     batching::gpu_preprocessing::{GpuPreprocessingMode, GpuPreprocessingSupport},
     camera::SortedCameras,
     mesh::allocator::MeshAllocator,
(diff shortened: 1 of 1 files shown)


kristoff3r (approved):
Code LGTM, tested a few lighting/shadow examples with/without GpuCulling.

JMS55 (approved):

alice-i-cecile:
Testing the [example runs](https://github.com/TheBevyFlock/bevy-example-runner/actions/runs/12248612460) before merging <3

pcwalton:
@alice-i-cecile Success

PR #16693, opened 2024-12-06
Draw the UI debug overlay using the UI renderer

### LINKED ISSUE #16666: `DebugUiPlugin` doesn't respect scale factor properly

ickshonpe:
# Objective

Draw the UI debug overlay using the UI renderer.

Significantly simpler and easier to use than `bevy_dev_tools::ui_debug_overlay` which uses `bevy_gizmos`.
* Supports multiple windows and UI rendered to texture.
* Draws rounded debug rects for rounded UI nodes. 

Fixes #16666

## Solution

Removed the `ui_debug_overlay` module from `bevy_dev_tools`.

Added a `bevy_ui_debug` feature gate.

Draw the UI debug overlay using the UI renderer.
Adds a new module `bevy_ui::render::debug_overlay`. 

The debug overlay extraction function queries for the existing UI layout and then adds a border around each UI node with `u32::MAX / 2` added to each stack index so it's drawn on top. 

There is a `UiDebugOptions` resource that can be used to enable or disable the debug overlay and set the line width.

## Testing

The `testbed_ui` example has been changed to use the new debug overlay:

```
cargo run --example testbed_ui --features bevy_ui_debug
```

Press Space to toggle the debug overlay on and off.

---

## Showcase

testbed-ui-new-debug

## Migration Guide

The `ui_debug_overlay` module has been removed from `bevy_dev_tools`.
There is a new debug overlay implemented using the `bevy_ui` renderer. To use it, enable the `bevy_ui_debug` feature and set the `enable` field of the `UiDebugOptions` resource to `true`.


crates/bevy_ui/src/render/debug_overlay.rs
```diff
@@ -0,0 +1,98 @@
...
+        };
+
+        let Ok(render_camera_entity) = mapping.get(camera_entity) else {
+            continue;
+        };
+
+        extracted_uinodes.uinodes.insert(
+            commands.spawn(TemporaryRenderEntity).id(),
+            ExtractedUiNode {
+                stack_index: uinode.stack_index + 2_147_483_647,
```
BenjaminBrienen, line 74: Does this magic value have a better way of writing it? Maybe a constant?

coreh: Yeah, this is `i32::MAX`

JMS55:
Can we get a rounded node in the example to ensure we don't break this later?

ickshonpe:
> Can we get a rounded node in the example to ensure we don't break this later?

The rotated bevy logo rotated in the bottom left has a rounded debug outline (the blue outline).

crates/bevy_ui/src/render/debug_overlay.rs
ickshonpe: It's just a big number added to the uinode's stack index so the overlay is drawn on top.

JMS55:
Oh I missed that 😅

alice-i-cecile (changes requested):
This is a much simpler and more reliable way to do this: I really prefer it. That said, we should clean up and remove the old approach in this PR, and write a migration guide.

I don't feel strongly about "where should this live": bevy_dev_tools is in a weird spot currently, but if this lives in bevy_ui we should probably feature flag it.

crates/bevy_ui/src/render/debug_overlay.rs
BenjaminBrienen: Using a constant and adding a comment are both good enhancements.

ickshonpe:
> This is a much simpler and more reliable way to do this: I really prefer it. That said, we should clean up and remove the old approach in this PR, and write a migration guide.
> 
> I don't feel strongly about "where should this live": bevy_dev_tools is in a weird spot currently, but if this lives in bevy_ui we should probably feature flag it.

I feel like a feature flag is a bad idea, it adds a lot of extra friction and this should be almost zero cost when disabled.

I only added border-box visualizations for this PR as a starting point but there are a lot more features I want to add in follow ups (assuming its accepted) that would be really useful like displaying margin, padding, and content boxes, debug labels, and a marker component to filter for specific nodes.

ickshonpe:
I have a dashed border renderer I wrote for Hedra which I think they'd let me upstream which might be nice for this too,

alice-i-cecile (approved):

JeanMertz:
I backported and tested this on the 0.15 release. Works beautifully.

alice-i-cecile:
@ickshonpe I'm going to count that as a second approval. Ping me when this is up to date with main please :)

ickshonpe:
@alice-i-cecile should be ready now. The check-advisories CI is failing but are we just ignoring that atm?

alice-i-cecile:
Yeah, it's blocked on an upstream fix :(

Issue #16680, opened 2024-12-06
`SystemRunner` param - run systems inside other systems


vil-mo:
## What problem does this solve or what need does it fill?

Currently ways of running systems from other systems are very limited. This proposal aims to provide a way to easily run system as a `SystemParam`

## What solution would you like?

Add a `SystemParam` - `SystemRunner<T: System + FromWorld>`

```rust
#[derive(Default)]
struct MySystem;
impl System for MySystem {
    In = In<i32>;
    Out = String;
    //...
}

fn running_my_system(mut runner: SystemRunner<MySystem>) {
    let output = runner.run(50);
    info!("Output of MySystem: {output}");
}
```

### Unergonomic `System` implementation

To use this param, user would have to explicitly specify the type of the system they will use. Implementing `System` trait on it's own is hard - unsafe code is not something users want to deal with. [`FunctionSystem`](https://dev-docs.bevyengine.org/bevy/ecs/system/struct.FunctionSystem.html)s are the main way people are used to working with the bevy systems, but they don't implement [`FromWorld`](https://dev-docs.bevyengine.org/bevy/ecs/prelude/trait.FromWorld.html), and even if it did, you would have to either specify the marker or box every system.

So, it would be nice to have a macro that labels a function system with a specified type name.

```rust
#[system(MySystem)]
pub fn my_system(query: Query<&mut MyComponent>, resourse: Res<MyRes>) -> i32 {
    //...
}
```
expands to
```rust
pub struct MySystem(SystemState<(Query<'static, 'static, &'static mut MyComponent>, Res<'static, MyRes>)>);

impl FromWorld for MySystem {
    fn from_world(world: &mut World) -> Self {
        MySystem(SystemState::new(world))
    }
}

impl System for MySystem {
    //... Basically the same as FunctionSystem implementation
}
```

### Needed controversial change

There is a conflict between [`System::update_archetype_component_access`](https://dev-docs.bevyengine.org/bevy/ecs/prelude/trait.System.html#tymethod.update_archetype_component_access) and [`SystemParam::new_archetype`](https://dev-docs.bevyengine.org/bevy/ecs/system/trait.SystemParam.html#method.new_archetype). You can't call `update_archetype_component_access` from `new_archetype`. Without that you wouldn't be able to call `update_archetype_component_access` on an inner system and simultaneously have access rights to run the system.

[...]

PR #16694, opened 2024-12-06
Cache systems by `S` instead of `S::System`

benfrankel:
# Objective

- Fixes the issue described in this comment: https://github.com/bevyengine/bevy/issues/16680#issuecomment-2522764239.

## Solution

- Cache one-shot systems by `S: IntoSystem` (which is const-asserted to be a ZST) rather than `S::System`.

## Testing

Added a new unit test named `cached_system_into_same_system_type` to `system_registry.rs`.

---

## Migration Guide

The `CachedSystemId` resource has been changed:

```rust
// Before:
let cached_id = CachedSystemId::<S::System>(id);
assert!(id == cached_id.0);

// After:
let cached_id = CachedSystemId::<S>::new(id);
assert!(id == SystemId::from_entity(cached_id.entity));
```


chescock (approved):

alice-i-cecile (changes requested):
I like this change, but it needs a test :)

benfrankel:
Added a unit test for this.

alice-i-cecile (approved):

alice-i-cecile:
Ping me when CI is green please?

benfrankel:
@alice-i-cecile CI passes now.

PR #16696, opened 2024-12-06
Rename `EntityCommands::clone` to `clone_and_spawn`

JaySpruce:
## Objective

Follow-up to #16672.

`EntityCommands::clone` looks the same as the `Clone` trait, which could be confusing. A discord discussion has made me realize that's probably a bigger problem than I thought. Oops :P

## Solution

Renamed `EntityCommands::clone` to `EntityCommands::clone_and_spawn`, renamed `EntityCommands::clone_with` to `EntityCommands::clone_and_spawn_with`. Also added some docs explaining the commands' relation to `Clone` (components need to implement it (or `Reflect`)).

## Showcase

```
// Create a new entity and keep its EntityCommands
let mut entity = commands.spawn((ComponentA(10), ComponentB(20)));

// Create a clone of the first entity
let mut entity_clone = entity.clone_and_spawn();
```

## The Bikeshed

- `clone_and_spawn` (Alice's suggestion)
- `spawn_clone` (benfrankel's suggestion)
- `spawn_cloned` (rparrett's suggestion)

### DIFF
--- a/crates/bevy_ecs/src/system/commands/mod.rs
+++ b/crates/bevy_ecs/src/system/commands/mod.rs
@@ -1720,9 +1720,11 @@ impl<'a> EntityCommands<'a> {
 
     /// Clones an entity and returns the [`EntityCommands`] of the clone.
     ///
-    /// The clone will have all the components of the original entity that
-    /// can be cloned. A component must implement either `Clone` or `Reflect`
-    /// to be cloned.
+    /// The clone will receive all the components of the original that implement
+    /// [`Clone`] or [`Reflect`](bevy_reflect::Reflect).
+    ///
+    /// To configure cloning behavior (such as only cloning certain components),
+    /// use [`EntityCommands::clone_and_spawn_with`].
     ///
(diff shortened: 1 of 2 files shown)


rparrett:
Here's a link to the referenced conversation.

https://discord.com/channels/691052431525675048/749335865876021248/1314635145650638960

JaySpruce:
Credit to @eugineerd for the neat feature, sorry for repeatedly nitpicking it :P

alice-i-cecile (changes requested):
I definitely think we should rename this, but I find `spawn_clone` quite awkward and unclear. Are you okay with `clone_and_spawn` instead?

rparrett:
What about `spawn_cloned`?

Also cc @benfrankel who suggested `spawn_clone`.

alice-i-cecile:
That's also an improvement, but I think that the order of operations (spawn, then clone) is clearest with my suggestion :)

alice-i-cecile (approved):

PR #16698, opened 2024-12-06
Physical Shadow Softness Parameter

ecoskey:
# Objective

Currently, directional lights use a non-physical `soft_shadow_size` value to determine soft shadows, in units of texels. This won't work with atmospherics, which needs a physical value for the angular size of the light to draw sun disks.

## Solution

- replace `soft_shadow_size` with `angular_size`, which is no longer optional. This field describes the angular size of the light/sun, or the angle between one side of the sun and the other in the sky, through its center.

- Add new `ShadowsStyle` component:
```rs
pub enum ShadowsStyle {
  #[default]
  None, 
  Hard, 
  #[cfg(feature = "experimental_pbr_pcss")]
  Soft
}
```
- make all lights require `ShadowsStyle`

## Testing

Tested pcss example, directional soft shadows don't work quite right. Hard shadows in other examples seem unaffected. 

## Migration Guide

The `shadows_enabled` and `enable_soft_shadows` fields have been removed from all lights. Add the `ShadowsStyle::Hard` component to a light to enable hard shadows, or `ShadowsStyle::Soft` to enable soft shadows (only with the `experimental_pbr_pcss` feature).

before:
```rs
commands.add(DirectionalLight {
  shadows_enabled: true,
  ..default()
});
```

after:
```rs
commands.add((
  DirectionalLight::default(),
  ShadowsStyle::Hard,
));
```

Instead of `soft_shadow_size` on directional lights, use `angular_size`. This field describes the angular size of the light/sun, or the angle between one side of the sun and the other in the sky, through its center. Its default value corresponds to the angular size of Earth's sun in the sky.


JMS55:
I don't really like `ShadowsStyle::Hard` either tbh. Like for us it makes sense, but for new users I think "oh if you want to turn on shadows, you need Shadow::Hard" is weird. Especially because PCSS is more expensive and probably won't be used a ton. Maybe we could have { Off, On, Soft }? Or just keep the existing API, or split the component up or something, idk.

PR #16702, opened 2024-12-07
:pencil2: Fix typos across bevy

homersimpsons:
# Objective

Fixes typos in bevy project, following suggestion in https://github.com/bevyengine/bevy-website/pull/1912#pullrequestreview-2483499337

## Solution

I used https://github.com/crate-ci/typos to find them.

I included only the ones that feel undebatable too me, but I am not in game engine so maybe some terms are expected.

I left out the following typos:
- `reparametrize` => `reparameterize`: There are a lot of occurences, I believe this was expected
- `semicircles` => `hemicircles`: 2 occurences, may mean something specific in geometry
- `invertation` => `inversion`: may mean something specific
- `unparented` => `parentless`: may mean something specific
- `metalness` => `metallicity`: may mean something specific

## Testing

- Did you test these changes? If so, how? I did not test the changes, most changes are related to raw text. I expect the others to be tested by the CI.
- Are there any parts that need more testing? I do not think
- How can other people (reviewers) test your changes? Is there anything specific they need to know? To me there is nothing to test

---

## Migration Guide

(kept in case I include the `reparameterize` change here)

## Questions

- [x] Should I include the above typos? No (https://github.com/bevyengine/bevy/pull/16702#issuecomment-2525271152)
- [ ] Should I add `typos` to the CI? (I will check how to configure it properly)

This project looks awesome, I really enjoy reading the progress made, thanks to everyone involved.

### DIFF
--- a/crates/bevy_asset/src/server/mod.rs
+++ b/crates/bevy_asset/src/server/mod.rs
@@ -1384,7 +1384,7 @@ impl AssetServer {
     /// asset from being dropped.
     /// If you have access to an asset's strong [`Handle`], you should prefer to call
     /// [`AssetServer::wait_for_asset`]
-    /// or [`wait_for_assest_untyped`](Self::wait_for_asset_untyped) to ensure the asset finishes
+    /// or [`wait_for_asset_untyped`](Self::wait_for_asset_untyped) to ensure the asset finishes
     /// loading.
     ///
     /// # Errors
--- a/crates/bevy_image/src/image_texture_conversion.rs
+++ b/crates/bevy_image/src/image_texture_conversion.rs
@@ -229,7 +229,7 @@ mod test {
 
         let image = Image::from_dynamic(initial.clone(), true, RenderAssetUsages::RENDER_WORLD);
 
-        // NOTE: Fails if `is_srbg = false` or the dynamic image is of the type rgb8.
+        // NOTE: Fails if `is_srgb = false` or the dynamic image is of the type rgb8.
         assert_eq!(initial, image.try_into_dynamic().unwrap());
     }
 }
--- a/crates/bevy_scene/src/dynamic_scene.rs
+++ b/crates/bevy_scene/src/dynamic_scene.rs
@@ -331,7 +331,7 @@ mod tests {
                 .get::<Parent>()
                 .expect("something is wrong with this test, and the scene components don't have a parent/child relationship")
                 .get(),
-            "something is wrong with the this test or the code reloading scenes since the relationship between scene entities is broken"
+            "something is wrong with this test or the code reloading scenes since the relationship between scene entities is broken"
         );
     }
 
--- a/examples/animation/animated_fox.rs
+++ b/examples/animation/animated_fox.rs
@@ -275,12 +275,12 @@ fn simulate_particles(
     time: Res<Time>,
 ) {
     for (entity, mut transform, mut particle) in &mut query {
-        if particle.lifeteime_timer.tick(time.delta()).just_finished() {
+        if particle.lifetime_timer.tick(time.delta()).just_finished() {
             commands.entity(entity).despawn();
         } else {
             transform.translation += particle.velocity * time.delta_secs();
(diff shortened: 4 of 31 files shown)


crates/bevy_asset/src/server/mod.rs
```diff
@@ -1384,7 +1384,7 @@ impl AssetServer {
     /// asset from being dropped.
     /// If you have access to an asset's strong [`Handle`], you should prefer to call
     /// [`AssetServer::wait_for_asset`]
-    /// or [`wait_for_assest_untyped`](Self::wait_for_asset_untyped) to ensure the asset finishes
+    /// or [`wait_for_asset_untyped`](Self::wait_for_asset_untyped) to ensure the asset finishes
```
homersimpsons, line 1387: I checked and the typo was only in the documentation.

crates/bevy_image/src/image_texture_conversion.rs
```diff
@@ -229,7 +229,7 @@ mod test {
 
         let image = Image::from_dynamic(initial.clone(), true, RenderAssetUsages::RENDER_WORLD);
 
-        // NOTE: Fails if `is_srbg = false` or the dynamic image is of the type rgb8.
+        // NOTE: Fails if `is_srgb = false` or the dynamic image is of the type rgb8.
```
homersimpsons, line 232: The typo is only in the doc.

crates/bevy_scene/src/dynamic_scene.rs
```diff
@@ -331,7 +331,7 @@ mod tests {
                 .get::<Parent>()
                 .expect("something is wrong with this test, and the scene components don't have a parent/child relationship")
                 .get(),
-            "something is wrong with the this test or the code reloading scenes since the relationship between scene entities is broken"
+            "something is wrong with this test or the code reloading scenes since the relationship between scene entities is broken"
```
homersimpsons, line 334: This is not detected by `typos` but is the original change I wanted to make.

examples/animation/animated_fox.rs
```diff
@@ -316,7 +316,7 @@ fn spawn_particle<M: Material>(
 
 #[derive(Component)]
 struct Particle {
-    lifeteime_timer: Timer,
+    lifetime_timer: Timer,
```
homersimpsons, line 319: I believe this change is safe to do, and looks contained to this file. If there should be any debate around it I can revert it.

atlv24 (approved):

examples/animation/animated_fox.rs
atlv24: this is fine

atlv24:
i read through these, none of these are controversial thanks

homersimpsons:
Thanks @atlv24 for the fast review. Do you have an answer for my 2 questions:

> Should I include the above typos?
> Should I add typos to the CI? (I will check how to configure it properly)

BenjaminBrienen:
Don't fix those typos that you correctly left out. Those should be removed from typos-ci.

BenjaminBrienen (approved):

mweatherley:
I still don't think 'reparametrize' is a typo or wrong (and I'm sort of the only person using this word either way), insofar as it comes ultimately from the notion of 'parametric' curves. (Perhaps the difference is largely Latin vs. Greek.) As a mathematician, this is how I've always seen this word spelled in this context, and both are regarded as correct.

rparrett:
>  Should I add typos to the CI? (I will check how to configure it properly)

We already use `typos` in CI, but currently an older version that doesn't catch some of these.

https://github.com/bevyengine/bevy/blob/main/typos.toml
https://github.com/bevyengine/bevy/blob/main/.github/workflows/ci.yml#L239

Related: https://github.com/bevyengine/bevy/pull/16602#issuecomment-2511825030

I agree with the typos you chose not to fix, except `invertation -> inversion` seems fine to me.

The other ones should get ignored in the config if we update CI.

mockersf (approved):

PR #16705, opened 2024-12-07
Remove COPY_DST from AsBindGroup uniform buffers

JMS55:
# Objective
- Wgpu barrier tracking is expensive. Making buffers read-only makes ideally lets wgpu skip worrying about barriers, although in wgpu 23 it apparently won't yet.

## Solution
- Remove COPY_DST usage from AsBindGroup uniform buffers to allow future wgpu versions to make this cheaper.
- AsBindGroup never updates buffers, so there's no need for COPY_DST. We always recreate all buffers and the bind group every time data changes, which yeah is also expensive.

## Testing
- Ran the animated materials example with/without bindless enabled. No crashes.

### DIFF
--- a/crates/bevy_render/macros/src/as_bind_group.rs
+++ b/crates/bevy_render/macros/src/as_bind_group.rs
@@ -482,7 +482,7 @@ pub fn derive_as_bind_group(ast: syn::DeriveInput) -> Result<TokenStream> {
                         #render_path::render_resource::OwnedBindingResource::Buffer(render_device.create_buffer_with_data(
                             &#render_path::render_resource::BufferInitDescriptor {
                                 label: None,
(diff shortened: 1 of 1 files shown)


pcwalton (approved):
Neat, good catch!

kristoff3r (approved):

IceSentry:
Is this configurable by users? I feel like some users may have been relying on that and this could break them. Bevy itself isn't, but our examples don't cover every use case. I agree it shouldn't be there by default though.

IceSentry:
> I don't think users have access to the buffers generated by AsBindGroup.

Right, I didn't think about that.

IceSentry (approved):

PR #16707, opened 2024-12-07
Add `World::try_resource_scope`

### LINKED ISSUE #16706: `try_resource_scope` that returns an `Option` instead of panicking in case the resource doesn't exist

Guarding against a panic with `resource_scope` currently requires the resource to be checked for existence prior to calling `resource_scope`:

```rust
if !world.contains_resource::<UiStack>() {
    return;
}

world.resource_scope(|world, mut stack: Mut<UiStack>| {
    // ...
});
```

Without the `contains_resource` call I always encounter a panic when closing the window, which I would like to avoid.

A new function with the following signature:

```rust
impl World {
    pub fn try_resource_scope<R: Resource, U>(&mut self, f: impl FnOnce(&mut World, Mut<R>) -> U) -> Option<U>;
}
```

And would be used as follows:

```rust
let ret = world.try_resource_scope(|world, mut stack: Mut<UiStack>| {
    // ...
});

// optionally use the return value
if let Some(ret) = ret {
    // ...
}
```

**I recommend reimplementing `resource_scope` in terms of this function when finished, to avoid duplication.**

atornity:
# Objective

Fixes #16706

## Solution 

- Added new method: `try_resource_scope` which returns `None` if the requested resource doesn't exist.
- Changed the `resource_scope` test to use `try_resource_scope` as well to test for the `None` case.

---

## Showcase

```rust
world.try_resource_scope::<MyResource, _>(|world, mut my_resource| {
    // do something with the resource if it exists
});
```

### DIFF
--- a/crates/bevy_ecs/src/lib.rs
+++ b/crates/bevy_ecs/src/lib.rs
@@ -1486,7 +1486,7 @@ mod tests {
         let mut world = World::default();
         assert!(world.try_resource_scope::<A, _>(|_, _| {}).is_none());
         world.insert_resource(A(0));
-        world.try_resource_scope(|world: &mut World, mut value: Mut<A>| {
+        world.resource_scope(|world: &mut World, mut value: Mut<A>| {
             value.0 += 1;
             assert!(!world.contains_resource::<A>());
         });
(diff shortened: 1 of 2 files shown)


ItsDoot (approved):

alice-i-cecile (approved):
This is nicer. Once we have system error handling sorted, we should probably just remove the panicking variant and tell people to ? / unwrap.

PR #16711, opened 2024-12-08
Use frostbite's specular sampling direction for environment map light

JMS55:
Adopt a slightly more accurate lighting model.

Before:
image

After:
image

## Changelog
- EnvironmentMapLight lighting is now slightly more realistic for metallic materials with high roughness

### DIFF
--- a/crates/bevy_pbr/src/light_probe/environment_map.wgsl
+++ b/crates/bevy_pbr/src/light_probe/environment_map.wgsl
@@ -259,6 +259,6 @@ fn environment_map_light(
 // https://seblagarde.wordpress.com/wp-content/uploads/2015/07/course_notes_moving_frostbite_to_pbr_v32.pdf#page=70
 fn radiance_sample_direction(N: vec3<f32>, R: vec3<f32>, roughness: f32) -> vec3<f32> {
     let smoothness = saturate(1.0 - roughness);
-    let lerp_factor = smoothness * sqrt((smoothness) + roughness);
+    let lerp_factor = smoothness * (sqrt(smoothness) + roughness);
     return mix(N, R, lerp_factor);
 }
--- a/crates/bevy_pbr/src/light_probe/environment_map.wgsl
+++ b/crates/bevy_pbr/src/light_probe/environment_map.wgsl
@@ -256,7 +256,7 @@ fn environment_map_light(
 }
 
 // "Moving Frostbite to Physically Based Rendering 3.0", listing 22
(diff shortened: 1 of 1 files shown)


crates/bevy_pbr/src/light_probe/environment_map.wgsl
```diff
@@ -206,9 +205,8 @@ fn environment_map_light(
     var out: EnvironmentMapLight;
 
     let radiances = compute_radiances(
-        input,
+        (*input).layers[LAYER_BASE],
```
pcwalton, line 208: Did you test the performance here? We used to use by-value struct passing and it had severe impacts on our performance that @Elabajaba noticed (that's why I changed it to `ptr` in the first place).

pcwalton (changes requested):

crates/bevy_pbr/src/light_probe/environment_map.wgsl
```diff
@@ -256,3 +254,11 @@ fn environment_map_light(
 
     return out;
 }
+
+// "Moving Frostbite to Physically Based Rendering 3.0", listing 22
+// https://seblagarde.wordpress.com/wp-content/uploads/2015/07/course_notes_moving_frostbite_to_pbr_v32.pdf#page=70
+fn radiance_sample_direction(N: vec3<f32>, R: vec3<f32>, roughness: f32) -> vec3<f32> {
+    let smoothness = saturate(1.0 - roughness);
+    let lerp_factor = smoothness * sqrt((smoothness) + roughness);
```
pcwalton, line 262: This doesn't match the paper: it should be `smoothness * (sqrt(smoothness) + roughness)` (notice parentheses).

crates/bevy_pbr/src/light_probe/environment_map.wgsl
JMS55: I didn't, but we use all 5 fields of the struct anyways, so I don't see what the difference could be. I'll try and profile.

pcwalton (changes requested):

crates/bevy_pbr/src/light_probe/environment_map.wgsl
JMS55: Ah damn, good catch. I double checked this, and still messed up 😓.

crates/bevy_pbr/src/light_probe/environment_map.wgsl
JMS55: Tested perf, no real difference.

pcwalton (approved):
Works for me then :)

kristoff3r (approved):

PR #16716, opened 2024-12-08
Rename `trigger.entity()` to `trigger.target()`

aevyrie:
# Objective

- A `Trigger` has multiple associated `Entity`s - the entity observing the event, and the entity that was targeted by the event.
- The field `entity: Entity` encodes no semantic information about what the entity is used for, you can already tell that it's an `Entity` by the type signature!

## Solution

- Rename `trigger.entity()` to `trigger.target()`

---

## Changelog

- `Trigger`s are associated with multiple entities. `Trigger::entity()` has been renamed to `Trigger::target()` to reflect the semantics of the entity being returned.

## Migration Guide

- Rename `Trigger::entity()` to `Trigger::target()`.
- Rename `ObserverTrigger::entity` to `ObserverTrigger::target`

### DIFF
--- a/crates/bevy_ecs/README.md
+++ b/crates/bevy_ecs/README.md
@@ -340,8 +340,8 @@ let mut world = World::new();
 let entity = world.spawn_empty().id();
 
 world.add_observer(|trigger: Trigger<Explode>, mut commands: Commands| {
-    println!("Entity {:?} goes BOOM!", trigger.entity());
-    commands.entity(trigger.entity()).despawn();
+    println!("Entity {:?} goes BOOM!", trigger.target());
+    commands.entity(trigger.target()).despawn();
 });
 
 world.flush();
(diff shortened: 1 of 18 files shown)


BenjaminBrienen (approved):
I like this and it lines up with how I'm used to this being called.

BenjaminBrienen:
Obviously, the description needs to be updated lol

alice-i-cecile (approved):
This is much clearer to me, thanks for putting this together.

alice-i-cecile:
CI caught a few more places where this needs to be renamed. Once that's fixed, ping me and I'll merge :)

aevyrie:
@alice-i-cecile looks like I was missing some recent commits to main.

aevyrie:
I'm not sure why CI is failing, the code it is pointing to doesn't exist on main or my branch.

PR #13432, opened 2024-05-20
bevy_reflect: Reflection-based cloning

MrGVSV:
# Objective

Using `Reflect::clone_value` can be somewhat confusing to those unfamiliar with how Bevy's reflection crate works. For example take the following code:

```rust
let value: usize = 123;
let clone: Box<dyn Reflect> = value.clone_value();
```

What can we expect to be the underlying type of `clone`? If you guessed `usize`, then you're correct! Let's try another:

```rust
#[derive(Reflect, Clone)]
struct Foo(usize);

let value: Foo = Foo(123);
let clone: Box<dyn Reflect> = value.clone_value();
```

What about this code? What is the underlying type of `clone`? If you guessed `Foo`, unfortunately you'd be wrong. It's actually `DynamicStruct`.

It's not obvious that the generated `Reflect` impl actually calls `Struct::clone_dynamic` under the hood, which always returns `DynamicStruct`.

There are already some efforts to make this a bit more apparent to the end-user: #7207 changes the signature of `Reflect::clone_value` to instead return `Box<dyn PartialReflect>`, signaling that we're potentially returning a dynamic type.

But why _can't_ we return `Foo`?

`Foo` can obviously be cloned— in fact, we already derived `Clone` on it. But even without the derive, this seems like something `Reflect` should be able to handle. Almost all types that implement `Reflect` either contain no data (trivially clonable), they contain a `#[reflect_value]` type (which, by definition, must implement `Clone`), or they contain another `Reflect` type (which recursively fall into one of these three categories).

This PR aims to enable true reflection-based cloning where you get back exactly the type that you think you do.

## Solution

Add a `Reflect::reflect_clone` method which returns `Result<Box<dyn Reflect>, ReflectCloneError>`, where the `Box<dyn Reflect>` is guaranteed to be the same type as `Self`.

```rust
#[derive(Reflect)]
struct Foo(usize);

let value: Foo = Foo(123);
let clone: Box<dyn Reflect> = value.reflect_clone().unwrap();
assert!(clone.is::<Foo>());
```

Notice that we didn't even need to derive `Clone` for this to work: it's entirely powered via reflection!

Under the hood, the macro generates something like this:

```rust
fn reflect_clone(&self) -> Result<Box<dyn Reflect>, ReflectCloneError> {
    Ok(Box::new(Self {
        // The `reflect_clone` impl for `usize` just makes use of its `Clone` impl
        0: Reflect::reflect_clone(&self.0)?.take().map_err(/* ... */)?,
    }))
[...]

Issue #10154, opened 2023-10-17
Automatically batch similar commands together to improve performance and robustness


alice-i-cecile:
## What problem does this solve or what need does it fill?

Commands are currently applied one at a time, in system graph order, and within that in the order they were sent within systems (FIFO). This causes some problems:

- #5074
- Confusing interactions between despawning entities and other operations like in #10122 or #2004.
- Suboptimal performance, as we are violating the basic principles of data-oriented design

## What solution would you like?

Automatically batch adjacent similar commands together.

## What alternative(s) have you considered?

We could also reorder commands to reduce surprising errors and confusing interactions. A sample ordering might be:

- Spawn
- Insert
- Remove
- Despawn

This would avoid a large number of spurious failures without needing explicit command error handling or system ordering. I'm unsure if that should be done as part of this work, or even at all however.

## Additional context

@maniwani has been investigating this, and thinks we should

maniwani:
> I'm personally less convinced: wouldn't those arguments similarly apply to any custom commands?

Custom commands are fine since they're still commands.

The batching implementation I'm investigating (used by flecs) benefits from having a "bird's eye view" of all queued commands. If there are deferred operations that aren't commands, those would cut off how many commands we can batch or we'd have to completely change the order we do things (e.g. run all commands first, then run all misc. deferred operations). IMO just remove `Deferred`. Commands *are* general-purpose. Why do we need `Deferred` when custom commands exist?

The only thing to know about custom commands is that batching would be able to "move" entity commands to run before them.

RedMindZ:
> We could also reorder commands to reduce surprising errors and confusing interactions

Since commands modify the shared state of the world, I don't think they could be reordered safely, unless we know ahead of time the set of all commands and all of their interactions, which is impossible with custom commands and will make adding new commands very difficult and fragile.

By not reordering the commands, the user has more control over how the world is modified, which is critical when the commands are trying to reflect in the world modifications to the user's state. It also becomes the user's responsibility to ensure commands are added in a valid order, which is much easier to control and debug when the commands are executed in the order they are created in in the user's code.

maniwani:
> We could also reorder commands to reduce surprising errors and confusing interactions.

> This would avoid a large number of spurious failures without needing explicit command error handling or system ordering.

I'm pretty sure batching is orthogonal to this.

Commands in general shouldn't even have errors, and all our built-in commands should be infallible.

[...]

PR #16717, opened 2024-12-08
Faster entity cloning

eugineerd:
# Objective

#16132 introduced entity cloning functionality, and while it works and is useful, it can be made faster. This is the promised follow-up to improve performance.

## Solution

**PREFACE**: This is my first time writing `unsafe` in rust and I have only vague idea about what I'm doing. I would encourage reviewers to scrutinize `unsafe` parts in particular.

The solution is to clone component data to an intermediate buffer and use `EntityWorldMut::insert_by_ids` to insert components without additional archetype moves.

To facilitate this, `EntityCloner::clone_entity` now reads all components of the source entity and provides clone handlers with the ability to read component data straight from component storage using `read_source_component` and write to an intermediate buffer using `write_target_component`. `ComponentId` is used to check that requested type corresponds to the type available on source entity.

Reflect-based handler is a little trickier to pull of: we only have `&dyn Reflect` and no direct access to the underlying data. `ReflectFromPtr` can be used to get `&dyn Reflect` from concrete component data, but to write it we need to create a clone of the underlying data using `Reflect`. For this reason only components that have `ReflectDefault` or `ReflectFromReflect` or `ReflectFromWorld` can be cloned, all other components will be skipped. The good news is that this is actually only a temporary limitation: once #13432 lands we will be able to clone component without requiring one of these `type data`s.

This PR also introduces `entity_cloning` benchmark to better compare changes between the PR and main, you can see the results in the **showcase** section.

## Testing

- All previous tests passing
- Added test for fast reflect clone path (temporary, will be removed after reflection-based cloning lands) 
- Ran miri

## Showcase
Here's a table demonstrating the improvement:

| **benchmark**           | **main, avg** | **PR, avg** | **change, avg** |
| ----------------------- | ------------- | ----------- | --------------- |
| many components reflect | 18.505 µs     | 2.1351 µs   | -89.095%        |
| hierarchy wide reflect*  | 22.778 ms     | 4.1875 ms   | -81.616%        |
| hierarchy tall reflect*  | 107.24 µs     | 26.322 µs   | -77.141%        |
| hierarchy many reflect  | 78.533 ms     | 9.7415 ms   | -87.596%        |
| many components clone   | 1.3633 µs     | 758.17 ns   | -45.937%        |
| hierarchy wide clone*    | 2.7716 ms     | 3.3411 ms   | +20.546%        |
| hierarchy tall clone*    | 17.646 µs     | 20.190 µs   | +17.379%        |
| hierarchy many clone    | 5.8779 ms     | 4.2650 ms   | -27.439%        |

*: these benchmarks have entities with only 1 component

## Considerations
Once #10154 is resolved a large part of the functionality in this PR will probably become obsolete. It might still be a little bit faster than using command batching, but the complexity might not be worth it.

## Migration Guide
- `&EntityCloner` in component clone handlers is changed to `&mut ComponentCloneCtx` to better separate data.
- Changed `EntityCloneHandler` from enum to struct and added convenience functions to add default clone and reflect handler more easily.


eugineerd:
Fixed tests for reflect-based cloning paths not actually checking reflect paths (oops) and `clone_fast` path in reflect handler not actually writing pointers to component data (which the tests should've caught). This fix didn't affect benchmark results in any noticeable way.

alice-i-cecile:
@JaySpruce can I get your review here please?

JaySpruce (approved):
Looks good to me! The unsafe seems reasonable, and inserting the components all at once is much preferable. 

One little thing that I can't highlight because it's not near any changes, there's 2 places in `component.rs` that still use `ComponentCloneHandler::default()` instead of `default_handler`, and it didn't break since they happen to be equivalent. `ComponentCloneHandler` still deriving `Default` probably isn't intentional.

crates/bevy_ecs/src/component.rs
```diff
@@ -2070,42 +2086,65 @@ pub fn component_clone_via_clone<C: Clone + Component>(
...
 /// See [`ComponentCloneHandlers`] for more details.
 #[cfg(feature = "bevy_reflect")]
-pub fn component_clone_via_reflect(world: &mut DeferredWorld, entity_cloner: &EntityCloner) {
-    let component_id = entity_cloner.component_id();
-    let source = entity_cloner.source();
-    let target = entity_cloner.target();
-    world.commands().queue(move |world: &mut World| {
-        world.resource_scope::<crate::reflect::AppTypeRegistry, ()>(|world, registry| {
+pub fn component_clone_via_reflect(world: &mut DeferredWorld, ctx: &mut ComponentCloneCtx) {
+    fn clone_fast(ctx: &mut ComponentCloneCtx) -> Option<()> {
```
alice-i-cecile, line 2090: I'd like some comments here about why this might fail to work, causing us to fallback to the slow mechanism.

crates/bevy_ecs/src/entity/clone_entities.rs
```diff
@@ -348,6 +609,44 @@ mod tests {
         assert!(world.get::<A>(e_clone).is_some_and(|c| *c == component));
     }
 
+    // TODO: remove this when 13432 lands
```
alice-i-cecile, line 612:
```suggestion
    // TODO: remove this when https://github.com/bevyengine/bevy/pull/13432 lands
```

alice-i-cecile:
I don't have the time to do a full audit of the unsafe here, but I like the API and the design makes sense to me.

crates/bevy_ecs/src/component.rs
```diff
@@ -2126,42 +2142,80 @@ pub fn component_clone_via_clone<C: Clone + Component>(
...
+        component_id: ComponentId,
+        source: Entity,
+        target: Entity,
+        registry: &crate::reflect::AppTypeRegistry,
+        world: &mut World,
+    ) -> Option<()> {
+        let registry = registry.read();
+
+        let component_info = world.components().get_info(component_id)?;
+        let type_id = component_info.type_id()?;
```
MrGVSV, line 2189: Seems like some of this is duplicated from `clone_fast`. Maybe we could merge the two to avoid double lookups and unnecessary retries (i.e. if `component_info.type_id()?` fails in `clone_fast`, there's no need to try again in `clone_slow`).

crates/bevy_ecs/src/component.rs
```diff
@@ -2126,42 +2142,80 @@ pub fn component_clone_via_clone<C: Clone + Component>(
...
-            let mut target = world
-                .get_entity_mut(target)
-                .expect("Target entity must exist");
-            reflect_component.apply_or_insert(&mut target, &*source_component, &registry);
+        Some(())
+    }
+
+    // This will try to clone component using ReflectComponent.
+    // If the component does not have this type data registered, it will be ignored.
+    fn clone_slow(
```
MrGVSV, line 2179: How much slower is this? I'm sure it is slower due to the extra `clone_value`, but it basically works the same as `clone_fast` after that: it attempts `ReflectFromReflect`, then `ReflectDefault`, then `ReflectFromWorld`, and then panics otherwise.

crates/bevy_ecs/src/component.rs
```diff
@@ -2126,42 +2142,80 @@ pub fn component_clone_via_clone<C: Clone + Component>(
...
+
+        let component_info = world.components().get_info(component_id)?;
+        let type_id = component_info.type_id()?;
+        let reflect_component =
+            registry.get_type_data::<crate::reflect::ReflectComponent>(type_id)?;
+        let source_component = reflect_component
+            .reflect(world.get_entity(source).ok()?)?
+            .clone_value();
+        let mut target = world.get_entity_mut(target).ok()?;
+        reflect_component.apply_or_insert(&mut target, &*source_component, &registry);
```
MrGVSV, line 2196: Note: This can panic. If that's undesirable, it may be worth creating a non-panicking version.

crates/bevy_ecs/src/component.rs
```diff
@@ -2126,42 +2142,80 @@ pub fn component_clone_via_clone<C: Clone + Component>(
...
             let registry = registry.read();
+            let component_info = ctx.components().get_info(ctx.component_id())?;
+            let type_id = component_info.type_id()?;
+
+            if let Some(reflect_from_reflect) =
+                registry.get_type_data::<bevy_reflect::ReflectFromReflect>(type_id)
+            {
+                reflect_from_reflect.from_reflect(source_component_reflect.as_partial_reflect())
+            } else if let Some(reflect_default) =
+                registry.get_type_data::<bevy_reflect::std_traits::ReflectDefault>(type_id)
```
MrGVSV, line 2163: We can also try `ReflectFromWorld` like `ReflectComponent` does (though it's not as commonly registered).

eugineerd: Wouldn't that require us to give access to `&mut World` to clone handlers (which will be unsound) then?

crates/bevy_ecs/src/component.rs
eugineerd: > How much slower is this

As slow as the results in the `main, avg` column of the benchmark table. 

Yeah, it seems `clone_slow` is pretty much reserved for `ReflectFromWorld` now and can panic. Maybe I should try to do it manually instead of using `ReflectComponent` then.

crates/bevy_ecs/src/component.rs
MrGVSV: Could we not queue it up as a command like you have for `clone_slow`?

eugineerd:
Removed slow path in `Reflect`-based handler in favor of manually implementing what `ReflectComponent` did. This also avoids panics when trying to clone components that don't implement one of `ReflectFromReflect`, `ReflectFromWorld`, `ReflectDefault`. The main difference is that `Reflect`-based clone handler now requires `ReflectFromPtr` to work, but I think this should be fine.

crates/bevy_ecs/src/component.rs
```diff
@@ -2106,62 +2128,100 @@ pub fn enforce_no_required_components_recursion(
...
-                .expect("Source entity must have reflected component")
-                .clone_value();
-            let mut target = world
-                .get_entity_mut(target)
-                .expect("Target entity must exist");
-            reflect_component.apply_or_insert(&mut target, &*source_component, &registry);
+pub fn component_clone_via_reflect(world: &mut DeferredWorld, ctx: &mut ComponentCloneCtx) {
+    // This path will be selected if:
+    // 1. Component has ReflectFromReflect or ReflectDefault registered.
+    // 2. Component has ReflectFromPtr registered.
```
MrGVSV, line 2158: I don't see `ReflectFromPtr` being used. I'm on mobile rn so I could also just be completely skipping past it 😅

crates/bevy_ecs/src/component.rs
```diff
@@ -2106,62 +2128,100 @@ pub fn enforce_no_required_components_recursion(
...
+    // Otherwise, it will fallback to clone_slow.
```
MrGVSV, line 2159: Nit: this should probably specify `ReflectFromWorld` since `clone_slow` no longer exists

crates/bevy_ecs/src/component.rs
```diff
@@ -2106,62 +2128,100 @@ pub fn enforce_no_required_components_recursion(
...
 /// Component [clone handler function](ComponentCloneFn) implemented using reflect.
 /// Can be [set](ComponentCloneHandlers::set_component_handler) as clone handler for any registered component,
 /// but only reflected components will be cloned.
 ///
-/// See [`ComponentCloneHandlers`] for more details.
+/// To clone a component using this handler, the following must be true:
+/// - World has [`AppTypeRegistry`](crate::reflect::AppTypeRegistry)
+/// - Component has [`TypeId`]
+/// - Component is registered
+/// - Component has [`ReflectFromPtr`](bevy_reflect::ReflectFromPtr) registered
```
MrGVSV, line 2147: Same here. Is `ReflectFromPtr` actually used?

crates/bevy_ecs/src/component.rs
eugineerd: It's used in `read_source_component_reflect`

crates/bevy_ecs/src/component.rs
eugineerd: This is actually an outdated comment that I forgot to remove, oops.

crates/bevy_ecs/src/component.rs
```diff
@@ -2106,62 +2128,96 @@ pub fn enforce_no_required_components_recursion(
...
+        let component_id = ctx.component_id();
+        world.commands().queue(move |world: &mut World| {
+            let mut component = reflect_from_world.from_world(world);
+            component.apply(source_component_cloned.as_partial_reflect());
+            // SAFETY:
+            // - component_id is from the same world as target entity
+            // - component is a valid value represented by component_id
+            unsafe {
+                let raw_component =
+                    core::ptr::NonNull::new_unchecked(Box::into_raw(component).cast::<u8>());
```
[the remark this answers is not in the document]
eugineerd: It is leaked according to miri, good catch.

crates/bevy_ecs/src/entity/clone_entities.rs
```diff
@@ -1,20 +1,225 @@
...
+    }
+
+    /// Returns a reference to the component on the source entity.
+    ///
+    /// Will return `None` if `ComponentId` of requested component does not match `ComponentId` of source component
+    pub fn read_source_component<T: Component>(&self) -> Option<&T> {
+        if self
+            .components
+            .component_id::<T>()
+            .is_some_and(|id| id == self.component_id)
```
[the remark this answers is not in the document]
eugineerd: > This also means it can't work with dynamic components.

I think that we can probably allow to clone dynamic components by exposing some unsafe methods on `ComponentCloneCtx` to get source component `Ptr` and to copy from provided `Ptr` to target. This would let someone who created dynamic component to register their own component clone handler.

eugineerd:
I think this should address everything outlined by reviews at the moment. Also added a method `ComponentCloneCtx::write_target_component_ptr` to allow writing custom clone handlers for components without `TypeId`.

chescock (approved):
Looks good!  

I found a safety hole that I missed on the first pass and that we should fix before merging, but the fix is simple so I'm hitting "approve" now.

alice-i-cecile:
@eugineerd I agree with the last two suggestions; ping me when those are addressed and I'll get this merged for you!

eugineerd:
@alice-i-cecile applied suggestions and CI is green now.

alice-i-cecile:
Needs a bit of cleanup to make sure this builds okay with `no_std`. See https://github.com/bevyengine/bevy/actions/runs/12399422024/job/34614292140. @bushrat011899 should be able to give you a hand if you get stuck :)

eugineerd:
@alice-i-cecile fixed `cargo build -p bevy_ecs --no-default-features` build, probably should be fine now?

PR #16723, opened 2024-12-09
Don't unconditionally create temporary render entities for visible objects.

pcwalton:
PR #15756 made us create temporary render entities for all visible objects, even if they had no render world counterpart. This regressed our `many_cubes` time from about 3.59 ms/frame to 4.66 ms/frame.

This commit changes that behavior to use `Entity::PLACEHOLDER` instead of creating a temporary render entity. This improves our `many_cubes` time from 5.66 ms/frame to 3.96 ms/frame, a 43% speedup.

I tested 3D, 2D gizmos, and UI and they seem to work.

See the following graph of `many_cubes` frame time (lower is better). PR #15756 is the one in October.

Time (ms_frame) vs  Date(3)

### DIFF
--- a/crates/bevy_pbr/src/render/light.rs
+++ b/crates/bevy_pbr/src/render/light.rs
@@ -31,7 +31,7 @@ use bevy_render::{
 };
 use bevy_render::{
     mesh::allocator::SlabId,
-    sync_world::{MainEntity, RenderEntity, TemporaryRenderEntity},
+    sync_world::{MainEntity, RenderEntity},
 };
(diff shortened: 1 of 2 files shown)


tychedelia (approved):
Thanks for looking into this.

kristoff3r (approved):

wilk10:
Does this qualify for 0.15.1 ?

pcwalton:
We should let it bake for a bit, because this has a decent probability of breaking something.

PR #16725, opened 2024-12-09
BRP strict field in query

Leinnan:
# Objective

- Allow skiping components that don't have ComponentId yet instead of failing `bevy/query` request.

## Solution

## Testing

My naive approach boils down to:
- bevy/list to get list of all components.
- bevy/query with empty components and has fields and a option that contains result of the bevy/list.

Before that change I end up with bunch of `Component xxx isn't used in the world` because some of the components wasn't spawned at any moment yet in the game. Now it should work.

## Migration Guide

- `BrpQueryParams` now has `strict` boolean field. It serfs as a flag to fail when encountering an invalid component rather than skipping it. Defaults to false.


BenjaminBrienen (approved):

Leinnan:
@alice-i-cecile is there anything missing now?

alice-i-cecile (approved):

PR #16727, opened 2024-12-09
`BorderRect` maintenance

ickshonpe:
# Objective

The doc comments and function namings for `BorderRect` feel imprecise to me. Particularly the `square` function which is used to define a uniform `BorderRect` with equal widths on each edge. But this is potentially confusing since this "square" border could be around an oblong shape.

Using "padding" to refer to the border extents seems undesirable too since "padding" is typically used to refer to the area between border and content, not the border itself.

## Solution
* Rename `square`  to `all` (this matches the name of the similar method on `UiRect`).
* Rename `rectangle` to `axes` (this matches the name of the similar method on `UiRect`).
* Update doc comments. 

## Migration Guide
The `square` and `rectangle` functions belonging to `BorderRect` have been renamed to `all` and `axes`.


crates/bevy_sprite/src/texture_slice/border_rect.rs
```diff
@@ -1,40 +1,41 @@
 use bevy_reflect::Reflect;
 
-/// Struct defining a [`Sprite`](crate::Sprite) border with padding values
+/// Defines the extents of the border of a rectangle.
+///
+/// This struct is used to represent thickness or offsets from the edges
+/// of a rectangle (left, right, top, and bottom), with values increasing inwards.
 #[derive(Default, Copy, Clone, PartialEq, Debug, Reflect)]
 pub struct BorderRect {
-    /// Pixel padding to the left
```
rparrett, line 6: As far as I know, this was the only hint in docs in anything related to texture slicing about the values being in pixels. So it would be great to get that information back somewhere.

ickshonpe: Yep that needs to be in the `TextureSlicer` docs I think. `BorderRect` is used in different coordinate spaces so it doesn't make sense to refer to pixels here.

alice-i-cecile:
I like these changes, but seconding the request not to lose the information about pixels :)

alice-i-cecile (approved):

BenjaminBrienen (approved):

alice-i-cecile:
CI failure seems real and related: I think there's a conditional compilation issue here.

rparrett:
I think it's just new stuff from #16693 that needs to be updated.

BenjaminBrienen (approved):

PR #16728, opened 2024-12-09
Making sure each plugin indicates its default group and feature when …

seivan:
# Objective

- Updating `plugin_group!` macro to include the plugin's meta information.
- Making sure each plugin indicates its default group and feature when applicable.

## Solution

- Output additional `cfg` flags when using `plugin_group!`
- For each plugin, add links to the plugin groups associated and necessary conditions. 

## Testing

- Checked that the docs updated accordingly. 
---

### DIFF
--- a/crates/bevy_a11y/src/lib.rs
+++ b/crates/bevy_a11y/src/lib.rs
@@ -108,7 +108,8 @@ pub enum AccessibilitySystem {
     /// Update the accessibility tree
     Update,
 }
-
+/// - Is part of the [`DefaultPlugins`](https://docs.rs/bevy/latest/bevy/struct.DefaultPlugins.html) plugin group when `bevy_window` feature is enabled.
+///
 /// Plugin managing non-GUI aspects of integrating with accessibility APIs.
 #[derive(Default)]
 pub struct AccessibilityPlugin;
--- a/crates/bevy_animation/src/lib.rs
+++ b/crates/bevy_animation/src/lib.rs
@@ -1233,6 +1233,8 @@ pub fn animate_targets(
         });
 }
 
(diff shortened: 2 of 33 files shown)


alice-i-cecile (changes requested):
Neat idea, and I like the notes.

Two things:

- this macro really needs comments to explain how / why it works. I'm very surprised that we can modify docs for the plugin itself!
- the newly added lines should be below the manually added docs

seivan:
> * this macro really needs comments to explain how / why it works. I'm very surprised that we can modify docs for the plugin itself!

That makes two of us, I wasn't expecting it. Maybe you can ask the original developers to add those, I just modified it to include the `custom` directives from the `cfg` - in fact, I would have preferred it if we could extract the `all`(s), `not`(s) and `target`(s) to work the same way so they could be highlighted better with markdown instead of just blasted as a straight line. But imho, it's good enough for now and you don't have to check source. 


> * the newly added lines should be below the manually added docs

I'll change that. 



You want to me remove the macro changes until someone can come along and explain/comment why it works?

alice-i-cecile:
Leave the macro changes for now: I think they're an improvement :)

BenjaminBrienen (approved):

alice-i-cecile:
@seivan Gentle reminder about this PR :) It's not far off merging now.

mockersf (changes requested):
I don't think the added doc on plugins should be the first line

the first line of docs is used by the html doc as a summary and presented alone in some context

seivan:
> @seivan Gentle reminder about this PR :) It's not far off merging now.

Sorry, was away. I’ll get on the changes requested tomorrow. 👌🏼

seivan:
> I don't think the added doc on plugins should be the first line
> 
> the first line of docs is used by the html doc as a summary and presented alone in some context

Yep @alice-i-cecile said as much as will change.

PR #16732, opened 2024-12-09
Update clipping rects in `ui_layout_system`

ickshonpe:
# Objective

Perform UI clipping updates during the walk to update the rest of the layout properties.

## Solution

* Removed `update_clipping_system` and renamed the `update` module to `update_target_cameras`.
* Migrated the queries and clipping update logic from `update_clipping_systems` to the  `update_uinode_geometry_recursive` function in the `layout` module.
* Nodes with `Display::None` set and their descendants don't need a full update, so added a second function to `layout` called `hide_uinodes_recursive` that just walks down the tree and sets the node sizes to zero.

##  Results

```
cargo run --example many_buttons --release --features "bevy/trace_tracy"
```
tracy-clip-in-update

Yellow is this PR, red is main.

The layout for `many_buttons` is very shallow and I'd expect most Bevy UI apps will see larger improvements in layout performance (but at the same time they'll have probably use less nodes so the absolute gains will be smaller).

```
cargo run --example many_buttons --release --features "bevy/trace_tracy" -- --display-none
```
mb-none


ickshonpe:
I'll do some benchmarks a bit later this evening. Typically it seems to only be a ~10% improvement overall but something like a modal UI with a very deep tree that uses `Display::None` to hide the inactive states you would see very large improvements.

Issue #16647, opened 2024-12-04
Prefix benchmark names with module path


BD103:
# Problem

I've slowly been returning to [`bevy-bencher`](https://github.com/TheBevyFlock/bevy-bencher), and am trying to migrate it to use Bevy's in-tree benchmarks instead of custom ones. One of the original issues I had with this approach is that it's difficult to see what a benchmark is testing from its name alone. For example:

- `layers_intersect`
- `entity_hash`
- `easing_1000`
- `param/combinator_system/8_piped_systems`
- `concrete_list_clone_dynamic`
- `ray_mesh_intersection/1000_vertices`
- `despawn_world_recursive/100_entities`
- `overhead_par_iter/threads_4`
- `run_condition/yes_using_resource`

All of these names were pulled from our current benchmarks, and are the names that would be displayed in Bencher's UI. Can you guess what each benchmark tracks specifically? Probably not, unless you're deeply familiar with that specific subsystem.

# Solution

Now look at the same list again, but with a few changes:

- `render::render_layers::intersect`
- `ecs::world::entity_hash`
- `math::bezier::easing_1000`
- `ecs::param::combinator_system::8_piped_systems`
- `reflect::list::concrete_clone_dynamic`
- `picking::ray_mesh_intersection/1000_vertices`
- `ecs::world::despawn_recursive/100_entities`
- `tasks::overhead_par_iter/threads_4`
- `ecs::scheduling::run_condition/yes_using_resource`

This naming scheme includes the module path in the benchmark name, and removes any redundant words from the benchmark name. There are a few benefits to this approach:

1. The name is far clearer on what is being tested.
2. It's easy to locate the benchmark from the name alone.
   - With the name `render::render_layers::intersect`, you know the benchmark is within `bevy_render/render_layers.rs`.
3. You can easily filter benchmarks to run by category.
   - For instance, you can run `cargo bench -- ecs::world` to run all `World`-related benchmarks.

## Automation

We can automate this naming a little bit using macros, specifically with [`module_path!()`](https://doc.rust-lang.org/stable/std/macro.module_path.html). For a quick sketch, you may be able to do this:

```rust
// I may have messed up this syntax, but you get the idea :)
macro_rules! bench {
  ($name:lit) => {
    concat!(module_path!(), $name)
  }
}

// Crate: `bevy_math`
mod bezier {
  fn easing(c: &mut Criterion) {
    // Name is `bevy_math::bezier::easing`.
[...]

PR #16733, opened 2024-12-09
Make benchmark setup consistent

BD103:
# Objective

- Benchmarks are inconsistently setup, there are several that do not compile, and several others that need to be reformatted.
- Related / precursor to #16647, this is part of my attempt migrating [`bevy-bencher`](https://github.com/TheBevyFlock/bevy-bencher) to the official benchmarks.

## Solution

> [!TIP]
>
> I recommend reviewing this PR commit-by-commit, instead of all at once!

In 5d26f56eb9404d14b8272ce8fa47e7eb318c2b18 I reorganized how benches were registered. Now this is one `[[bench]]` per Bevy crate. In each crate benchmark folder, there is a `main.rs` that calls `criterion_main!`. I also disabled automatic benchmark discovery, which isn't necessarily required, but may clear up confusion with our custom setup. I also fixed a few errors that were causing the benchmarks to fail to compile.

In afc8d33a87d52df65d52392d207acfdfe189f5b2 I ran `rustfmt` on all of the benchmarks.

In d6cdf960abe27bf68a19fdcd5aa97d4aedd2d314 I fixed all of the Clippy warnings.

In ee94d48f5085002566a5a6eada9225a10ff067cb I fixed some of the benchmarks' usage of `black_box()`. I ended up opening https://github.com/rust-lang/rust/pull/133942 due to this, which should help prevent this in the future.

In cbe1688dcd7998771ec99f8765896fa859c7da00 I renamed all of the ECS benchmark groups to be called `benches`, to be consistent with the other crate benchmarks.

In e701c212cdc6bb57f1af37316e7834f2ac982630 and 8815bb78b0d8a87326f365da51940a01cf37e1f9 I re-ordered some imports and module definitions, and uplifted `fragmentation/mod.rs` to `fragementation.rs`.

Finally, in b0065e0b0bf53c9159cb5c210214c2576a803ec5 I organized `Cargo.toml` and bumped Criterion to v0.5.

## Testing

- `cd benches && cargo clippy --benches`
- `cd benches && cargo fmt --all`


benches/benches/bevy_ecs/empty_archetypes.rs
```diff
@@ -70,25 +67,26 @@ fn par_for_each(
...
+    query.par_iter().for_each(|comp| {
         black_box(comp);
     });
 }
 
+#[expect(
+    unused_variables,
+    reason = "`parallel` has no effect, it needs to be removed or parallel support needs to be re-added."
+)]
 fn setup(parallel: bool, setup: impl FnOnce(&mut Schedule)) -> (World, Schedule) {
```
BD103, line 79: @james7132 or @hymm, from the commit history it looks like you both worked on this file before.

I'm not familiar with `ComputeTaskPool`, but it's no longer a resource, so I temporarily removed it. How can I re-add it to work with the current version of Bevy, or should I remove the `parallel` argument completely?

hymm: I think it would be roughly equivalent to switch the executor kind between `SingleThreaded` and `MultiThreaded`. https://dev-docs.bevyengine.org/bevy/ecs/prelude/struct.Schedule.html#method.set_executor_kind

BD103: Thank you for the quick response! Implemented in df26ba2a21c8b33409da8c33866066c15d933aae :)

bushrat011899 (approved):
Looks good to me! Definitely a cleaner structure.

benches/Cargo.toml
```diff
@@ -22,70 +21,47 @@ bevy_render = { path = "../crates/bevy_render" }
...
-path = "benches/bevy_ecs/benches.rs"
-harness = false
-
-[[bench]]
-name = "ray_mesh_intersection"
-path = "benches/bevy_picking/ray_mesh_intersection.rs"
-harness = false
-
-[[bench]]
-name = "reflect_function"
```
MrGVSV, line 49: I was previously running function-reflection-specific benches like:
```
cargo bench --bench reflect_function --all-features
```
Does this change mean I won't be able to run specific sets of benches anymore?

BD103: Yes, but only temporarily! Just like unit tests, you can filter what subset of benchmarks you want to run:

```bash
# Will run `typed/function`, `typed/closure`, `typed/closure_mut`, and more.
cargo bench --bench reflect -- typed
```

My plan with #16647 is to rename all benchmarks to include their module paths, so you'll be able to run:

```bash
# Will run `reflect::function::typed/closure`, etc.
cargo bench --bench reflect -- function
```

This PR was a precursor, since I didn't want to have too many changes at once.

Trashtalk217: I want to second this, because when I still was putting the [stress test ](https://github.com/bevyengine/bevy/pull/16591), it was very annoying to test *just* the bench while not testing everything else. Can you still run specific benchmarks (or groups of them) with this reorganization?

alice-i-cecile:
@BD103 once this is merge conflict free I'll press the button. Thanks for this: the benchmarks folder has been in a state of disarray for way too long.

BD103:
Whoops, looks like I messed up fixing merge conflicts. Please hold!

BD103:
@alice-i-cecile should be good to go now!

PR #16738, opened 2024-12-10
Improve ComputedNode accessibility

UkoeHB:
# Objective

- Enable modifying node size after layout.
- Gain access to a node's content_size. `UiSurface` is a private type so content size can't be looked up.

## Solution

- Make `ComputedNode` fields public.
- Add `content_size` to `ComputedNode`.


alice-i-cecile (commented):
Good, I think making this public with strongly worded docs is the way forward.

alice-i-cecile (approved):

BenjaminBrienen:
It seems weird that something "computed" should be  made public with the intention of being mutable. I have no experience writing bevy UI, so I don't have insight about this idea, but it is executed correctly.

BenjaminBrienen (approved):

alice-i-cecile:
> It seems weird that something "computed" should be made public with the intention of being mutable

Yep, absolutely. That's why it was designed this way originally, but advanced users occasionally need an escape hatch, and this doesn't seem like a tripping hazard to me.

PR #16739, opened 2024-12-10
Feature gate `is_polygon_simple` behind the `alloc` feature.

pcwalton:
CI was failing because `bevy_math` no longer compiled with `libcore`. This was due to PR #15981. This commit fixes the issue by moving the applicable functionality behind `#[cfg(feature = "alloc")]`.


bushrat011899 (approved):
Thanks for fixing this one! I'm surprised it managed to sneak in, I guess the full CI isn't run during the merging perhaps? Regardless, this shouldn't be a problem going forward anyway, since every other crate in Bevy will require `alloc`, so the only new feature will be `std`, which should hopefully be less frequent.

BenjaminBrienen (approved):

Issue #14322, opened 2024-07-15
Use bicubic filtering in lightmaps


mgi388:
## What problem does this solve or what need does it fill?

Lightmap support was added in #10231.

A [TODO note] was left to support bicubic filtering.

[TODO note]: https://github.com/bevyengine/bevy/blob/4a312a12e79ccf32b54ba2d3e99053685028c055/crates/bevy_pbr/src/lightmap/lightmap.wgsl#L23

## What solution would you like?

- Support bicubic filtering
- [GPU Gems 2 - Chapter 20. Fast Third-Order Texture Filtering] seems to be a good reference (a rendering engineer could confirm though)
- Also see **Additional context** below for a solution Griffin used: I haven't checked if that's based on the GPU Gems 2 article, or not

[GPU Gems 2 - Chapter 20. Fast Third-Order Texture Filtering]: https://developer.nvidia.com/gpugems/gpugems2/part-iii-high-quality-rendering/chapter-20-fast-third-order-texture-filtering

## What alternative(s) have you considered?

Leave it as is.

## Additional context

- Griffin [ported some existing bicubic filtering code to WGSL]. You can read the [thread in Discord].
- I pulled that code down and put it into lightmap.wgsl.
- I created a simple plane with a lightmap to compare the before and after:

Before (staircase shadows):

![image](https://github.com/user-attachments/assets/d1097a5e-56f0-426f-9829-fbd009db84ac)

After (smooth shadows):

![image](https://github.com/user-attachments/assets/325ba126-b676-4412-8c5a-88c2631864b3)

- Note: I had to modify Griffin's code to get it to compile. I used this:

<details>
  <summary>lightmap.wgsl</summary>

  ```wgsl
  #define_import_path bevy_pbr::lightmap
  
  #import bevy_pbr::mesh_bindings::mesh
  
  @group(1) @binding(4) var lightmaps_texture: texture_2d<f32>;
  @group(1) @binding(5) var lightmaps_sampler: sampler;
  
  // Samples the lightmap, if any, and returns indirect illumination from it.
  fn lightmap(uv: vec2<f32>, exposure: f32, instance_index: u32) -> vec3<f32> {
      let packed_uv_rect = mesh[instance_index].lightmap_uv_rect;
      let uv_rect = vec4<f32>(vec4<u32>(
          packed_uv_rect.x & 0xffffu,
          packed_uv_rect.x >> 16u,
          packed_uv_rect.y & 0xffffu,
[...]

PR #16740, opened 2024-12-10
Higher quality bicubic lightmap sampling

JMS55:
# Objective
- Closes https://github.com/bevyengine/bevy/issues/14322.

## Solution
- Implement fast 4-sample bicubic filtering based on this shader toy https://www.shadertoy.com/view/4df3Dn, with a small speedup from a ghost of tushima presentation.

## Testing

  - Ran on lightmapped example. Practically no difference in that scene.
  - Lightmapping a better scene.

## Changelog
- Lightmaps now have a higher quality bicubic sampling method (off by default).


DGriffin91:
imo this should be optional (with at least linear being the other option). Try with a low res light map to see the effect.

crates/bevy_pbr/src/lightmap/lightmap.wgsl
```diff
@@ -8,22 +8,72 @@
...
+    let g0x = g0(fuv.x);
+    let g1x = g1(fuv.x);
+    let h0x = h0_approx(fuv.x);
+    let h1x = h1_approx(fuv.x);
+    let h0y = h0_approx(fuv.y);
+    let h1y = h1_approx(fuv.y);
+    let p0 = (vec2(iuv.x + h0x, iuv.y + h0y) - 0.5) * texel_size;
+	let p1 = (vec2(iuv.x + h1x, iuv.y + h0y) - 0.5) * texel_size;
+	let p2 = (vec2(iuv.x + h0x, iuv.y + h1y) - 0.5) * texel_size;
+	let p3 = (vec2(iuv.x + h1x, iuv.y + h1y) - 0.5) * texel_size;
```
pcwalton, line 35: nit: Odd indentation here

crates/bevy_pbr/src/lightmap/lightmap.wgsl
```diff
@@ -8,22 +8,72 @@
...
+fn g1(a: f32) -> f32 {
+    return w2(a) + w3(a);
+}
+
+fn h0_approx(a: f32) -> f32 {
+    return 0.2 + a * (0.24 * a - 0.44);
+}
+
+fn h1_approx(a: f32) -> f32 {
+    return 1.0 + a * (0.24 * a - 0.04);
```
pcwalton, line 78: I saw that comment, but the rounding on the coefficients for `h1_approx` leads to some serious error:
BadApprox

I calculated out the real coefficients here:
GoodApprox

crates/bevy_pbr/src/lightmap/lightmap.wgsl
```diff
@@ -8,22 +8,72 @@
...
+fn g0(a: f32) -> f32 {
+    return w0(a) + w1(a);
+}
+
+fn g1(a: f32) -> f32 {
+    return w2(a) + w3(a);
+}
+
+fn h0_approx(a: f32) -> f32 {
+    return 0.2 + a * (0.24 * a - 0.44);
```
pcwalton, line 74: For comparison with the code below, this is a fine approximation.

OkApprox

crates/bevy_pbr/src/lightmap/lightmap.wgsl
pcwalton:
```suggestion
    return 1.00158 + a * (0.230963 * a - 0.0353001);
```

pcwalton (approved):
Looks good modulo a little inaccuracy.

I could go either way on exposing an option for bilinear, and I won't block the PR on that. A quote from the Bakery author's blog post: https://ndotl.wordpress.com/2018/08/29/baking-artifact-free-lightmaps/

> Bicubic interpolation. If you are not shipping on mobile, there are exactly 0 reasons to not use bicubic interpolation for lightmaps. Many UE3 games in the past did that, and it is a great trick. But some engines (Unity, I’m looking at you) still think they can get away with a single bilinear tap in 2018. Bicubic hides low resolution and makes jagged lines appear smooth. Sometimes I see people fixing jagged sharp shadows by super-sampling during bake, but it feels like a waste of lightmapping time to me.

See the post for a screenshot comparison.

The "if you are not shipping on mobile" comment is 6 years old, so it may well just be time to say bicubic everywhere.

crates/bevy_pbr/src/lightmap/lightmap.wgsl
JMS55: Copy+pasted tabs :/. Fixed.

crates/bevy_pbr/src/lightmap/lightmap.wgsl
pcwalton: Oops, I replaced the - with a +! Sign error. Ignore my comment, your code is fine the way it is

DGriffin91:
Based these images from @pcwalton I don't think bicubic is correct: https://discord.com/channels/691052431525675048/743663924229963868/1316259592308527134

image

JMS55:
Consensus is to make bicubic vs bilinear a toggle, and default to bilinear. Bicubic is oftentimes only a marginal quality improvement, can lead to light leaks, and is more expensive.

JMS55:
I think prepass + deferred might be broken on main. I don't see any usage of `lightmap_image` in `queue_prepass_material_meshes`, unlike how it's used in `queue_material_meshes`.

alice-i-cecile (approved):

alice-i-cecile:
@DGriffin91 @JMS55 what's your final opinion on the status of this? Did https://github.com/bevyengine/bevy/pull/16740/commits/d2243653852619882699c77357783239c7eed8d7 resolve the concerns?

JMS55:
It's fine now imo, but lightmaps being broken with deferred needs fixing first.

JMS55:
Blocked on https://github.com/bevyengine/bevy/pull/16836, and then once that's merged I need to update the deferred queuing code.

crates/bevy_pbr/src/lightmap/lightmap.wgsl
```diff
@@ -13,33 +13,87 @@
...
+    let g1x = g1(fuv.x);
+    let h0x = h0_approx(fuv.x);
+    let h1x = h1_approx(fuv.x);
+    let h0y = h0_approx(fuv.y);
+    let h1y = h1_approx(fuv.y);
+    let p0 = (vec2(iuv.x + h0x, iuv.y + h0y) - 0.5) * texel_size;
+    let p1 = (vec2(iuv.x + h1x, iuv.y + h0y) - 0.5) * texel_size;
+    let p2 = (vec2(iuv.x + h0x, iuv.y + h1y) - 0.5) * texel_size;
+    let p3 = (vec2(iuv.x + h1x, iuv.y + h1y) - 0.5) * texel_size;
+    let color = g0(fuv.y) * (g0x * sample(p0, lightmap_slot) + g1x * sample(p1, lightmap_slot)) + g1(fuv.y) * (g0x * sample(p2, lightmap_slot) + g1x * sample(p3, lightmap_slot));
```
pcwalton, line 42: Pretty sure I already reviewed this, so I'm just assuming this is correct :)

crates/bevy_pbr/src/lightmap/mod.rs
```diff
@@ -100,6 +100,13 @@ pub struct Lightmap {
     /// This field allows lightmaps for a variety of meshes to be packed into a
     /// single atlas.
     pub uv_rect: Rect,
+
+    /// Whether bicubic sampling should be used for sampling this lightmap.
+    ///
+    /// Bicubic sampling is higher quality, but slower, and may lead to light leaks.
+    ///
+    /// If true, the lightmap texture's sampler must be set to [`bevy_image::ImageSampler::linear`].
```
pcwalton, line 108: Might want to mention that the default is false.

pcwalton (approved):
Looks fine, tested and working. I submitted a very small PR to your repo: https://github.com/JMS55/bevy/pull/29 The PR makes the feature easier to test by adding a `--bicubic` option to the `lightmaps` example.

Issue #16743, opened 2024-12-10
derive Copy for Stopwatch/Timer


mirsella:
## What problem does this solve or what need does it fill?

Easier usage of Stopwatch/Timer, including custom struct wrapping a Timer.
Personally im using a macro to generate components with derive and more, and one of them has a timer, which mean all the other components can't be Copy even if they dont use timer.

## Additional context

from https://doc.rust-lang.org/std/marker/trait.Copy.html:
`(Copy) Types whose values can be duplicated simply by copying bits.`
`Generally speaking, if your type can implement Copy, it should.`
`A type can implement Copy if all of its components implement Copy.`

all theses a true, and all the field of Stopwatch/Timer impl copy.

i can open a PR, thanks !

BenjaminBrienen:
The general advice is to implement common traits when possible for this reason, so I'm in agreement.

rparrett:
I think that implicit duplication of these specific types could be pretty confusing.

mirsella:
I thought of that, but i guess its exactly the same with most rust types like Duration ? i agree it can lead to non-wanted duplication, but it seems to be widely used in rust.
in what case (code) would an accidental “hidden” copy happen ? sorry im new and can't seems to find a examples, apart from some that are not really "hidden" so i would count them.

my use case is special and i don't think is revelant:
```rust
define_components! {
    struct TroopHealth { pub current: u32, pub max: u32 },
    struct Cost(pub u32),
    struct GoldGeneration(pub u32, pub Timer),
    struct Damage(pub u32),
    struct AttackCooldown(pub Timer),
```
i have a simple macro like this to create the components, derive things on them, add them to a struct with a builder that allows simple constructions of stats, this struct has a method to insert all of its stats on a entity and bevy_replicon::replicate them.
i have a single derive statement for all the components in the macro, and so they all need to be the same. at Timer doesn't implement Copy, all other Components can't. in most systems that use these components, i need to clone() simple struct even like Damage, or Cost to send events, and create more entity, as they don't implement Copy.

rparrett:
> i have a simple macro like this to create the components, derive things on them

It seems sort misguided to derive `Copy` for all of the components in an entire app unconditionally.

> its exactly the same with most rust types like Duration

The difference between e.g. `Stopwatch` and `Duration` is the `is_paused` field which we expect users to mutate. An unintentional copy isn't just a performance concern, but also a potential source of bugs.

mirsella:
[...]

PR #16748, opened 2024-12-10
derive Copy on Stopwatch and Timer

mirsella:
# Objective

Fixes https://github.com/bevyengine/bevy/issues/16743


alice-i-cecile:
I'm not convinced we should do this. There's important state stored in here, and accidentally copying it seems like a serious footgun.

mirsella:
closing then

PR #16749, opened 2024-12-10
Improve `bevy_input_focus`

eckz:
# Objective

I was curious to use the newly created `bevy_input_focus`, but I found some issues with it
  - It was only implementing traits for `World`.
  - Lack of tests
  - `is_focus_within` logic was incorrect.

## Solution
 This PR includes some improvements to the `bevy_input_focus` crate: 
  - Add new `IsFocusedHelper` that doesn't require access to `&World`. It implements `IsFocused`
  - Remove `IsFocused` impl for `DeferredWorld`. Since it already implements `Deref<Target=World>` it was just duplication of code.
  - impl `SetInputFocus` for `Commands`. There was no way to use `SetFocusCommand` directly. This allows it.
  - The `is_focus_within` logic has been fixed to check descendants. Previously it was checking if any of the ancestors had focus which is not correct according to the documentation.
  - Added a bunch of unit tests to verify the logic of the crate.

## Testing

- Did you test these changes? If so, how? Yes, running newly added unit tests.

---

### DIFF
--- a/crates/bevy_input_focus/Cargo.toml
+++ b/crates/bevy_input_focus/Cargo.toml
@@ -6,7 +6,7 @@ description = "Keyboard focus management"
 homepage = "https://bevyengine.org"
 repository = "https://github.com/bevyengine/bevy"
 license = "MIT OR Apache-2.0"
-keywords = ["bevy", "color"]
+keywords = ["bevy"]
 rust-version = "1.76.0"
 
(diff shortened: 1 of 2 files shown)


viridia (approved):
You learn something new every day :)

alice-i-cecile (approved):
Very nice, thanks :)

PR #16750, opened 2024-12-10
Register type `BoxShadow`

jf908:
# Objective

- Register `BoxShadow` type for reflection

## Testing

- Tested that box shadow example compiles and runs

## Additional

- It would be nice to have this in 0.15.1


ickshonpe (approved):
:sweat: not the first time I forgot this

PR #16755, opened 2024-12-10
Update the prepass shaders and fix the batching logic for bindless and multidraw.

pcwalton:
This commit resolves most of the failures seen in #16670. It contains two major fixes:

1. The prepass shaders weren't updated for bindless mode, so they were accessing `material` as a single element instead of as an array. I added the needed `BINDLESS` check.

2. If the mesh didn't support batch set keys (i.e. `get_batch_set_key()` returns `None`), and multidraw was enabled, the batching logic would try to multidraw all the meshes in a bin together instead of disabling multidraw. This is because we checked whether the `Option<BatchSetKey>` for the previous batch was equal to the `Option<BatchSetKey>` for the next batch to determine whether objects could be multidrawn together, which would return true if batch set keys were absent, causing an entire bin to be multidrawn together. This patch fixes the logic so that multidraw is only enabled if the batch set keys match *and are `Some`*.

Additionally, this commit adds batch key support for bins that use `Opaque3dNoLightmapBinKey`, which in practice means prepasses. Consequently, this patch enables multidraw for the prepass when GPU culling is enabled.

When testing this patch, try adding `GpuCulling` to the camera in the `deferred_rendering` and `ssr` examples. You can see that these examples break without this patch and work properly with it.


JMS55 (approved):

kristoff3r (approved):

kristoff3r:
I played around with the deferred rendering example, and this happens if you first go to "Forward + Prepass", back to just "Forward" and then resize the window. I don't know if it's related to this PR or not, it didn't happen on main but main was broken in other ways.

image

pcwalton:
Probably unrelated, but I can verify for sure.

pcwalton:
@kristoff3r That bug is still there in 0.15, so it's unrelated.

PR #16757, opened 2024-12-10
Make indirect drawing opt-out instead of opt-in, enabling multidraw by default.

pcwalton:
This patch replaces the undocumented `NoGpuCulling` component with a new component, `NoIndirectDrawing`, effectively turning indirect drawing on by default. Indirect mode is needed for the recently-landed multidraw feature (#16427). Since multidraw is such a win for performance, when that feature is supported the small performance tax that indirect mode incurs is virtually always worth paying.

To ensure that custom drawing code such as that in the `custom_shader_instancing` example continues to function, this commit additionally makes GPU culling take the `NoFrustumCulling` component into account.

This PR is an alternative to #16670 that doesn't break the `custom_shader_instancing` example. **PR #16755 should land first in order to avoid breaking deferred rendering, as multidraw currently breaks it**.

## Migration Guide

* Indirect drawing (GPU culling) is now enabled by default, so the `GpuCulling` component is no longer available. To disable indirect mode, which may be useful with custom render nodes, add the new `NoIndirectDrawing` component to your camera.


JMS55:
Did you mean to add the new component to custom_shader_instancing?

pcwalton:
@JMS55 Good catch, fixed.

bushrat011899 (approved):
Code looks good, and it works on my i5-1240P iGPU with Windows 10 and DX12.

pcwalton:
Looks like the tests all passed except for some timeouts on Android that look like test flakiness rather than actual failures.

alice-i-cecile (approved):

PR #16758, opened 2024-12-10
Add `no_std` support to `bevy_ecs`

bushrat011899:
# Objective

- Contributes to #15460

## Solution

  - `std` (default)
  - `async_executor` (default)
  - `edge_executor`
  - `critical-section`
  - `portable-atomic`
- Gated `tracing` in `bevy_utils` to allow compilation on certain platforms
- Switched from `tracing` to `log` for simple message logging within `bevy_ecs`. Note that `tracing` supports capturing from `log` so this should be an uncontroversial change.
- Fixed imports and added feature gates as required 
- Made `bevy_tasks` optional within `bevy_ecs`. Turns out it's only needed for parallel operations which are already gated behind `multi_threaded` anyway.

## Testing

- Added to `compile-check-no-std` CI command
- `cargo check -p bevy_ecs --no-default-features --features edge_executor,critical-section,portable-atomic --target thumbv6m-none-eabi`
- `cargo check -p bevy_ecs --no-default-features --features edge_executor,critical-section`
- `cargo check -p bevy_ecs --no-default-features`

## Draft Release Notes

Bevy's core ECS now supports `no_std` platforms.

In prior versions of Bevy, it was not possible to work with embedded or niche platforms due to our reliance on the standard library, `std`. This has blocked a number of novel use-cases for Bevy, such as an embedded database for IoT devices, or for creating games on retro consoles.

With this release, `bevy_ecs` no longer requires `std`. To use Bevy on a `no_std` platform, you must disable default features and enable the new `edge_executor` and `critical-section` features. You may also need to enable `portable-atomic` and `critical-section` if your platform does not natively support all atomic types and operations used by Bevy.

```toml
[dependencies]
bevy_ecs = { version = "0.16", default-features = false, features = [
  # Required for platforms with incomplete atomics (e.g., Raspberry Pi Pico)
  "portable-atomic",
  "critical-section",

  # Optional
  "bevy_reflect",
  "serialize",
  "bevy_debug_stepping",
  "edge_executor"
] }
```

Currently, this has been tested on bare-metal x86 and the Raspberry Pi Pico. If you have trouble using `bevy_ecs` on a particular platform, please reach out either through a GitHub issue or in the `no_std` working group on the Bevy Discord server.

Keep an eye out for future `no_std` updates as we continue to improve the parity between `std` and `no_std`. We look forward to seeing what kinds of applications are now possible with Bevy!

## Notes

- Creating PR in draft to ensure CI is passing before requesting reviews.
- This implementation has no support for multithreading in `no_std`, especially due to `NonSend` being unsound if allowed in multithreading. The reason is we cannot check the `ThreadId` in `no_std`, so we have no mechanism to at-runtime determine if access is sound.


crates/bevy_ecs/src/component.rs
```diff
@@ -29,6 +29,12 @@ use core::{
 };
 use thiserror::Error;
 
+#[cfg(feature = "portable-atomic")]
+use portable_atomic_util::Arc;
+
+#[cfg(not(feature = "portable-atomic"))]
+use alloc::sync::Arc;
```
bushrat011899, line 36: For a handful of atomic types, this kind of feature gating is required to support platforms without full native atomic support. In general, `portable-atomic` is a drop-in replacement for `alloc::sync`.

crates/bevy_ecs/src/component.rs
```diff
@@ -1983,35 +1989,69 @@ impl RequiredComponents {
...
-                    }
-                });
-            },
-        ));
+        let erased: RequiredComponentConstructor = RequiredComponentConstructor({
+            // `portable-atomic-util` `Arc` is not able to coerce an unsized
+            // type like `std::sync::Arc` can. Creating a `Box` first does the
+            // coercion.
+            //
+            // This would be resolved by https://github.com/rust-lang/rust/issues/123430
```
bushrat011899, line 1997: This area is a little frustrating. In short, `std::sync::Arc` has certain privileges that `portable_atomic_util::Arc` doesn't, the key one being unsized coercion. To work around this, we first create a `Box` (which can do the coercion), and then create an `Arc` from that box. This code is a little verbose to handle the 2×2 feature matrix of tracking changes and portable atomics while also ensuring the `Box` workaround is only used when `portable-atomic` is enabled.

crates/bevy_ecs/src/entity/mod.rs
```diff
@@ -61,20 +62,30 @@ use crate::{
...
 type IdCursor = i64;
 
 /// Most modern platforms support 64-bit atomics, but some less-common platforms
 /// do not. This fallback allows compilation using a 32-bit cursor instead, with
 /// the caveat that some conversions may fail (and panic) at runtime.
-#[cfg(not(target_has_atomic = "64"))]
+#[cfg(all(not(target_has_atomic = "64"), not(feature = "portable-atomic")))]
 use core::sync::atomic::AtomicIsize as AtomicIdCursor;
+#[cfg(all(not(target_has_atomic = "64"), feature = "portable-atomic"))]
+use portable_atomic::AtomicIsize as AtomicIdCursor;
```
bushrat011899, line 88: I haven't tried it, but I believe using `portable-atomic` may allow access to an `AtomicI64` on all platforms, removing the possible panic. This should be investigated in a follow-up PR.

crates/bevy_ecs/src/intern.rs
```diff
@@ -4,9 +4,15 @@
...
 //! and make comparisons for any type as fast as integers.
 
+use alloc::{borrow::ToOwned, boxed::Box};
 use core::{fmt::Debug, hash::Hash, ops::Deref};
+
+#[cfg(feature = "std")]
 use std::sync::{OnceLock, PoisonError, RwLock};
 
+#[cfg(not(feature = "std"))]
+use spin::{once::Once as OnceLock, rwlock::RwLock};
```
bushrat011899, line 14: A handful of sync primitives are conditionally replaced with alternatives from `spin`. I would like to revisit our reliance on `spin` at a later date. Spinning is considered a last-resort sync option, and designing it out may improve `std` performance too.

bushrat011899 (commented):
Here's some comments to assist reviewers with evaluating this PR. It is fairly lengthy so I'd recommend skimming through these first.

crates/bevy_ecs/Cargo.toml
```diff
@@ -11,39 +11,89 @@ categories = ["game-engines", "data-structures"]
...
+portable-atomic-util = { version = "0.2.4", features = [
+  "alloc",
+], optional = true }
+spin = { version = "0.9.8", default-features = false, features = [
+  "spin_mutex",
+  "rwlock",
+  "once",
+] }
+tracing = { version = "0.1", default-features = false, optional = true }
+log = { version = "0.4", default-features = false }
```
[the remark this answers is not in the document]
alice-i-cecile: We should swap to a workspace dependency IMO: synchronizing these manually is silly. But that probably warrants its own independent discussion.

bushrat011899:
CI is passing, so I consider this PR ready for review!

crates/bevy_ecs/src/lib.rs
```diff
@@ -84,6 +89,17 @@ pub mod prelude {
...
+/// Exports used by macros.
+///
+/// These are not meant to be used directly and are subject to breaking changes.
+#[doc(hidden)]
+pub mod __macro_exports {
+    // Cannot directly use `alloc::vec::Vec` in macros, as a crate may not have
+    // included `extern crate alloc;`. This re-export ensures we have access
+    // to `Vec` in `no_std` and `std` contexts.
+    pub use alloc::vec::Vec;
+}
```
bushrat011899, line 101: The new required components recursion check uses a `Vec`, which isn't in the global namespace with `no_std`. This technique of adding a `pub mod __macro_exports { ... }` is how `bevy_reflect` gets around this issue, so this is at least consistent, even if it isn't ideal.

alice-i-cecile: We could define a feature-flagged type alias here instead right? But yeah, consistent is fine for now.

crates/bevy_ecs/Cargo.toml
```diff
@@ -11,39 +11,85 @@ categories = ["game-engines", "data-structures"]
...
+]
+trace = ["std", "dep:tracing", "bevy_utils/tracing"]
 multi_threaded = ["bevy_tasks/multi_threaded", "arrayvec"]
 bevy_debug_stepping = []
-serialize = ["dep:serde"]
+serialize = ["dep:serde", "bevy_utils/serde"]
 track_change_detection = []
 reflect_functions = ["bevy_reflect", "bevy_reflect/functions"]
-detailed_trace = []
+detailed_trace = ["trace"]
```
alice-i-cecile, line 35: Comments and maybe some grouping would make these features a lot easier to follow. There's a ton being added here, and it's hard to understand why they're needed and what they each do.

crates/bevy_ecs/Cargo.toml
```diff
@@ -11,39 +11,89 @@ categories = ["game-engines", "data-structures"]
...
+  "dep:critical-section",
+  "bevy_tasks/critical-section",
+  "portable-atomic?/critical-section",
+]
+portable-atomic = [
+  "dep:portable-atomic",
+  "dep:portable-atomic-util",
+  "bevy_tasks/portable-atomic",
+  "concurrent-queue/portable-atomic",
+]
```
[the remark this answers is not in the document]
alice-i-cecile: Useful, but would be more useful as a comment in the code ;)

alice-i-cecile (commented):
Once the features are more clear this LGTM. Spin-locking on no-std platforms is a fine solution for now.

crates/bevy_ecs/Cargo.toml
bushrat011899: Agreed. I've reorganised the features into distinct sections with short descriptions. It's not quite the same as how other crates (e.g., `bevy_internal`) document their features, but I believe this is more appropriate and easier to understand. But I am happy to change it based on feedback!

Victoronz (commented):
Some small code style changes, then it'd LGTM!

crates/bevy_ecs/src/component.rs
Victoronz: Might be good to note that `#[derive(CoercePointee)]` is currently in FCP for stabilization, which means it is likely that we'll have it by 1.85 (20th February)

Victoronz (commented):
(didn't catch the review comments)

crates/bevy_ecs/src/system/exclusive_function_system.rs
```diff
@@ -10,7 +10,7 @@ use crate::{
     world::{unsafe_world_cell::UnsafeWorldCell, World},
 };
 
-use alloc::borrow::Cow;
+use alloc::{borrow::Cow, vec, vec::Vec};
```
Victoronz, line 13: I think this would usually be written as `vec::{self, Vec}`.

crates/bevy_ecs/src/component.rs
chescock: Would it be easier to combine the code if the closure inside `RequiredComponentConstructor` took a `MaybeLocation`?  The comments on the alias say "Please use this type sparingly", but I'm not sure what actual cost there is to using it.

chescock (approved):

crates/bevy_ecs/src/system/exclusive_function_system.rs
bushrat011899: Unfortunately this doesn't work. `vec::{self}` imports the _module_ `vec`, while `vec` imports the _macro_ `vec!`.

crates/bevy_ecs/src/intern.rs
```diff
@@ -136,15 +142,43 @@ impl<T: Internable + ?Sized> Interner<T> {
...
+        let lock = {
+            #[cfg(feature = "std")]
+            let lock = self.0.get_or_init(Default::default);
+
+            #[cfg(not(feature = "std"))]
+            let lock = self.0.call_once(Default::default);
+
+            lock
+        };
+
```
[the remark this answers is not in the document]
bushrat011899: My original thinking was it'd be "easier" to read if the feature gating was grouped together, but looking at it now, it doesn't help much. I've flattened the structure, removing a few lines from the diff.

crates/bevy_ecs/src/component.rs
bushrat011899: I _do_ think this should be used in more places. I don't like the verbosity that the `cfg(...)` introduces, and I don't believe that it's more performant (a ZST _should_ be removable from a function argument at compile time, since it has zero size)

crates/bevy_ecs/src/system/exclusive_function_system.rs
Victoronz: ah if it's the macro then it is okay!

crates/bevy_ecs/src/component.rs
Victoronz: It'd be good to note that the feature that resolves this, `[derive(CoercePointee)]` is currently in FCP for stabilization, so it will likely hit stable in about 2 months.

Victoronz (approved):

bushrat011899:
That CI issue around building `bevy_ecs` _without_ default features will need to be fixed! I'll try and get that resolved shortly...

Victoronz:
Another question, I just realized that `HashSet`/`HashMap` are neither in `core` nor `alloc`. Yet `EntityHashSet`/`EntityHashMap` doesn't come up in the diff with a feature flag here, has that just not come up yet?

bushrat011899:
> Another question, I just realized that `HashSet`/`HashMap` are neither in `core` nor `alloc`. Yet `EntityHashSet`/`EntityHashMap` doesn't come up in the diff with a feature flag here, has that just not come up yet?

Thankfully, `hashbrown` is `no_std` compatible, and `bevy_utils` has provided `hashbrown` as an alternative to the `std` `HashMap` and `HashSet` for a while now!

bushrat011899:
Ok I've made `bevy_ecs` work without default features. As a nice bonus, the fix involved making `bevy_tasks` optional, which was shockingly simple (only needed for things that were already gated behind `multi_threaded`). This reduces the number of dependencies required to build `bevy_ecs` to 40 on my machine.

alice-i-cecile:
Excellent! That definitely feels like the right design, and something we would have wanted to do anyways.

crates/bevy_ecs/src/world/filtered_resource.rs
```diff
@@ -217,11 +221,21 @@ impl<'w, 's> From<&'w FilteredResourcesMut<'_, 's>> for FilteredResources<'w, 's
 impl<'w> From<&'w World> for FilteredResources<'w, 'static> {
     fn from(value: &'w World) -> Self {
         static READ_ALL_RESOURCES: OnceLock<Access<ComponentId>> = OnceLock::new();
-        let access = READ_ALL_RESOURCES.get_or_init(|| {
-            let mut access = Access::new();
-            access.read_all_resources();
+        let access = {
```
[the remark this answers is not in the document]
bushrat011899: Just tested:
```rust
const READ_ALL_RESOURCES: &'static Access<ComponentId> = {
    const ACCESS: Access<ComponentId> = {
        let mut access = Access::new();
        access.read_all_resources();
        access
    };
    &ACCESS
};
```
You need the nested `const`s to create a static reference while guaranteeing that `Access` wont ever be dropped (compiler refuses to compile if it's possible to drop an `Access` as it's a "non-trivial drop"). But this does remove the use of `OnceLock` from `READ_ALL_RESOURCES` and `WRITE_ALL_RESOURCES`, which is very nice!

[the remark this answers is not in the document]
chescock: Too late now, but you can do it with one `const` if you put the `&` around the whole block like

```rust
const READ_ALL_RESOURCES: &'static Access<ComponentId> = &{
    let mut access = Access::new();
    access.read_all_resources();
    access
};
```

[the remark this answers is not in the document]
bushrat011899: Ahhh that's how you'd do it! I was placing it on the inside of the block.

alice-i-cecile:
Thank you to everyone involved with the authoring or reviewing of this PR! This work is relatively important and needs release notes! Head over to https://github.com/bevyengine/bevy-website/issues/1976 if you'd like to help out.

Issue #16477, opened 2024-11-22
check-advisories Github Action fails due to unmaintained `instant` crate.


andriyDev:
## Bevy version

6741e01df

## What you did

- Ran `cargo deny check advisories` (same thing as the Github Action).

## What went wrong

```
error[unmaintained]: `instant` is unmaintained
    ┌─ E:\Content\bevy/Cargo.lock:293:1
    │
293 │ instant 0.1.13 registry+https://github.com/rust-lang/crates.io-index
    │ ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ unmaintained advisory detected
    │
    ├ ID: RUSTSEC-2024-0384
    ├ Advisory: https://rustsec.org/advisories/RUSTSEC-2024-0384
    ├ This crate is no longer maintained, and the author recommends using the maintained [`web-time`] crate instead.

      [`web-time`]: https://crates.io/crates/web-time
    ├ Solution: No safe upgrade is available!
    ├ instant v0.1.13
      └── notify-types v1.0.0
          ├── notify v7.0.0
          │   └── notify-debouncer-full v0.4.0
          │       └── bevy_asset v0.15.0-dev
          │           ├── bevy_animation v0.15.0-dev
          │           │   ├── bevy_gltf v0.15.0-dev
          │           │   │   └── bevy_internal v0.15.0-dev
          │           │   │       ├── bevy v0.15.0-dev
          │           │   │       └── bevy_dylib v0.15.0-dev
          │           │   │           └── bevy v0.15.0-dev (*)
          │           │   └── bevy_internal v0.15.0-dev (*)
          │           ├── bevy_audio v0.15.0-dev
          │           │   └── bevy_internal v0.15.0-dev (*)
          │           ├── bevy_core_pipeline v0.15.0-dev
          │           │   ├── bevy_dev_tools v0.15.0-dev
          │           │   │   └── bevy_internal v0.15.0-dev (*)
          │           │   ├── bevy_gizmos v0.15.0-dev
          │           │   │   ├── bevy_dev_tools v0.15.0-dev (*)
          │           │   │   └── bevy_internal v0.15.0-dev (*)
          │           │   ├── bevy_gltf v0.15.0-dev (*)
          │           │   ├── bevy_internal v0.15.0-dev (*)
          │           │   ├── bevy_pbr v0.15.0-dev
[...]

PR #16763, opened 2024-12-11
Ignore the 'instant is unmaintained' advisory.

andriyDev:
# Objective

- Hides #16477.

## Solution

Add the advisory ID to the list of ignored advisories.

The notify-types crate has already been switched to web-time upstream, but it's up to the maintainer to publish the crate. There is nothing for us to do, so better to just ignore it so we don't ignore this CI check anymore (and mistakenly miss new advisories).

## Testing

- Tested locally.

### DIFF
--- a/deny.toml
+++ b/deny.toml
@@ -3,7 +3,10 @@ all-features = true
 
 [advisories]
 version = 2
-ignore = []
+ignore = [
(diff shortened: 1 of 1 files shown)


BD103 (approved):
Thank you!

PR #16766, opened 2024-12-11
Use one `BevyManifest` instance in proc macros

raldone01:
# Objective

- Minor consistency improvement in proc macro code.
- Remove `get_path_direct` since it was only used once anyways and doesn't add much.

## Solution
- Possibly a minor performance improvement since the `Cargo.toml` wont be parsed as often.

## Testing

- I don't think it breaks anything.
- This is my first time working on bevy itself. Is there a script to do a quick verify of my pr.


alice-i-cecile:
>  Is there a script to do a quick verify of my pr?
Use `cargo run -p ci`, which runs the `ci` project in the `tools` folder :)

alice-i-cecile (changes requested):
Looks nice, but as the bot says, you'll need to bump the MSRV. I'm pretty sure that's due to the use of LazyLock here?

raldone01:
No the msrv issue was because I also tried to use `CARGO_MANIFEST_PATH` instead of `CARGO_MANIFEST_DIR` which apparently needs a higher msrv.

I did a quick search but did not find any annotations when it was added exactly, so I just reverted back to `CARGO_MANIFEST_DIR`.

As you can see the all the ci runs passed.

raldone01:
> Use `cargo run -p ci`, which runs the `ci` project in the `tools` folder :)

I tried to run the ci scripts by `cd`ing into the folder and running them with `cargo run`.

Consider adding an advisory for `cargo run -p ci` to the PR-template.

alice-i-cecile (approved):

alice-i-cecile:
Nope, I just forgot to update the label after your response :) This needs a second review then it's good to merge.

crates/bevy_macro_utils/src/bevy_manifest.rs
```diff
@@ -1,17 +1,21 @@
...
 
-impl Default for BevyManifest {
-    fn default() -> Self {
-        Self {
+const BEVY: &str = "bevy";
+const BEVY_INTERNAL: &str = "bevy_internal";
+
+impl BevyManifest {
+    /// Returns a global shared instance of the [`BevyManifest`] struct.
+    pub fn default_shared() -> &'static LazyLock<Self> {
```
mockersf, line 17:
```suggestion
    pub fn shared() -> &'static LazyLock<Self> {
```
and all the places where it's called too.

It was called `default` because of the `Default` impl, but it doesn't make sense to keep that part of the name

raldone01: Makes sense.

mockersf (approved):

PR #16770, opened 2024-12-11
Add `compile_fail_utils`, compile fail tests, and benchmarks to Cargo workspace

BD103:
# Objective

- `compile_fail_utils`, `benches`, and all of our `compile_fail` sub-crates are outside of the Cargo workspace, so they are not checked by default nor caught by CI. As a result, these crates have several compile errors and warnings that were never discovered.

## Solution

- Add `compile_fail_utils`, `benches`, and all `compile_fail` crates to the Cargo workspace.
- Fix several compile errors and warnings, and run `rustfmt`.
- Remove custom `release` profile from benchmarks.
  - This may cause a regression in benchmark statistics, since we no longer set `lto = true`.
  - I can re-add this back as a custom profile available to all crates within the workspace, but I'm unsure its worth the hassle.
- Do not run `cargo test --benches` in CI.
  - This would have caused all of the benchmarks to be run for every pull request, which doesn't seem desired.

## Testing

- `cargo check --workspace --all-targets`
- `cd tools/compile_fail_utils && cargo test --test example`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,20 +13,19 @@ documentation = "https://docs.rs/bevy"
 rust-version = "1.82.0"
 
 [workspace]
-exclude = [
-  "benches",
-  "crates/bevy_derive/compile_fail",
-  "crates/bevy_ecs/compile_fail",
-  "crates/bevy_reflect/compile_fail",
-  "tools/compile_fail_utils",
-]
 members = [
+  # All of Bevy's official crates are within the `crates` folder!
   "crates/*",
+  # Several crates with macros have "compile fail" tests nested inside them, also known as UI
+  # tests, that verify diagnostic output does not accidentally change.
+  "crates/*/compile_fail",
+  # Examples of compiling Bevy for mobile platforms.
   "examples/mobile",
-  "tools/ci",
-  "tools/build-templated-pages",
-  "tools/build-wasm-example",
-  "tools/example-showcase",
+  # Benchmarks
+  "benches",
+  # Internal tools that are not published.
+  "tools/*",
+  # Bevy's error codes. This is a crate so we can automatically check all of the code blocks.
   "errors",
 ]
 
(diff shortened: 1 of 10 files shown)


alice-i-cecile (approved):
Good fix, but can we add something to CI to ensure that this is automatically checked?

BD103:
> Good fix, but can we add something to CI to ensure that this is automatically checked?

Yup! The simplest fix I could find was to add all of the `compile_fail` tests to the Cargo workspace. (Which works, surprisingly, so I'm unsure why we didn't do this in the past!)

BenjaminBrienen:
I thought that we don't include the compile_fail tests because it doesn't work with crater. @alice-i-cecile am I hallucinating that? 😆

alice-i-cecile:
I don't remember that! And not seeing any issues on https://github.com/rust-lang/crater. Are you sure you haven't been replaced by an LLM?

Anyways @BD103 CI is now mad at you, so kicking this back to you until it's placated.

BenjaminBrienen:
https://github.com/bevyengine/bevy/blob/main/tools%2Fcompile_fail_utils%2FREADME.md

BD103:
> https://github.com/bevyengine/bevy/blob/main/tools%2Fcompile_fail_utils%2FREADME.md

I wonder why it says that! As far as I can tell, Crater just runs `cargo build && cargo test`. The only time Crater may fail is if it manually runs `rustc` on any `.rs` file it finds, which wouldn't make sense... @cart, do you have anything to add to this?

Either way, let me fix the final few Clippy lints!

BD103:
Turns out CI was running `cargo test --benches`, which would cause all of our benchmarks to be run for every PR. I turned that off in 2d763ab8965c733baa264c1bed4c808f338b6808, since that seems undesired.

BenjaminBrienen:
For the lifetime elisions, disable that clippy lint to be in line with the workspace.

BD103:
While there's a bit more I could do to improve the code quality of the benchmarks, I'm going to save it for a separate PR.

BenjaminBrienen (approved):

BD103:
I'm going to take another shot at this tomorrow. For some reason I'm unable to reproduce CI locally...

Issue #16639, opened 2024-12-03
`Taffy` crashes


Phoqinu:
## Bevy version

`bevy = "0.15.0"`

## What you did

I am making dagerfall inspired game and it's quite basic but providing minimal code might be pretty hard. 
This only happens after I pick and place item in inventory. This code involves changing parent/child with bevy built-in commands.
Then I hit escape which bring-ups menu and then I go to main menu which causes automatic entity/resource/etc clean-up but here it crashes.

## What went wrong

```

thread 'Compute Task Pool (14)' panicked at C:\Users\Phoqinu\.cargo\registry\src\index.crates.io-6f17d22bba15001f\taffy-0.5.2\src\tree\taffy_tree.rs:447:29:
invalid SlotMap key used

```

<details><summary> stack backtrace:</summary>

```

   0: std::panicking::begin_panic<ref$<str$> >
             at C:\Users\Phoqinu\.rustup\toolchains\stable-x86_64-pc-windows-msvc\lib\rustlib\src\rust\library\std\src\panicking.rs:737
   1: slotmap::basic::impl$9::index_mut<slotmap::DefaultKey,enum2$<core::option::Option<taffy::tree::node::NodeId> > >
             at C:\Users\Phoqinu\.cargo\registry\src\index.crates.io-6f17d22bba15001f\slotmap-1.0.7\src\basic.rs:936
   2: taffy::tree::taffy_tree::TaffyTree<enum2$<bevy_ui::measurement::NodeMeasure> >::remove<enum2$<bevy_ui::measurement::NodeMeasure> >
             at C:\Users\Phoqinu\.cargo\registry\src\index.crates.io-6f17d22bba15001f\taffy-0.5.2\src\tree\taffy_tree.rs:447
   3: bevy_ui::layout::ui_surface::UiSurface::remove_camera_entities<core::iter::adapters::map::Map<core::iter::adapters::flatten::Flatten<core::option::IntoIter<core::iter::adapters::cloned::Cloned<bevy_ecs::event::iterators::EventIterator<bevy_ecs::removal_det
             at C:\Users\Phoqinu\.cargo\registry\src\index.crates.io-6f17d22bba15001f\bevy_ui-0.15.0\src\layout\ui_surface.rs:261
   4: bevy_ui::layout::ui_layout_system
             at C:\Users\Phoqinu\.cargo\registry\src\index.crates.io-6f17d22bba15001f\bevy_ui-0.15.0\src\layout\mod.rs:241
[...]

PR #16780, opened 2024-12-12
Upgrade Taffy to 0.7

nicoburns:
# Objective

- Includes https://github.com/DioxusLabs/taffy/pull/749
- Which should fix https://github.com/bevyengine/bevy/issues/16639

## Solution

- Bump taffy version from `0.6` to `0.7`

## Testing

- I have run a couple of examples, but no extensive testing.


alice-i-cecile (approved):
Good, this should be an easy small upgrade. Very pleased that the crash fix is in.

BenjaminBrienen (approved):

PR #16784, opened 2024-12-12
Replace FixedBitSet in Access and FilteredAccess with sorted vectors

Trashtalk217:
Adopted from #14385.

# Objective

As described in the [relations RFC](https://github.com/james-j-obrien/rfcs/blob/minimal-fragmenting-relationships/rfcs/79-minimal-fragmenting-relationships.md#access-bitsets-and-component-sparsesets). 

The access bitsets are currently efficient because the ComponentIds are dense: they are incremented from 0 and should remain small.
With relations, a ComponentId will have some upper bits 1, so the amount of memory allocated in the FixedBitSet to represent the value would be non-trivial.

## Solution

We replace `FixedBitSets` with sorted vectors.
The vectors should remain relatively small since queries usually don't involve hundreds of components.
These allow us to do union, difference, and intersection operations in O(1).
Inserting a value, however, is O(1).

## Testing

TODO (I can't figure out the bencher)

## Migration Guide

TODO (is it really necessary?)


crates/bevy_ecs/src/storage/sorted_small_vec.rs
```diff
@@ -0,0 +1,259 @@
+use smallvec::SmallVec;
+use core::cmp::Ordering;
+
+/// Stores a sorted list of indices with quick implementation for union, difference, intersection.
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub struct SortedSmallVec<const N: usize>(SmallVec<[usize; N]>);
```
andriyDev, line 6: Why is this `pub`? Why not `pub(crate)`?

This extends to `SortedSmallVec::new` and `SortedSmallVec::from_vec`

crates/bevy_ecs/src/storage/sorted_small_vec.rs
```diff
@@ -0,0 +1,259 @@
...
+    /// Duplicates are removed.
+    pub fn from_vec(vec: Vec<usize>) -> Self {
+        let mut sorted_vec = Self(SmallVec::with_capacity(vec.len()));
+        for value in vec {
+            sorted_vec.insert(value);
+        }
+        sorted_vec
+    }
+
+    pub(crate) fn ones(&self) -> impl Iterator<Item = usize> + '_ {
```
andriyDev, line 48: `ones` was kinda confusing here. It makes sense in the context of "replacing FixedBitSet", but it doesn't really make sense for this struct.

Can we just say `iter`? That's pretty much all this is.

crates/bevy_ecs/src/storage/sorted_small_vec.rs
andriyDev, line 6: I feel like the name here isn't clear. Maybe just `SmallVecSet`? `SortedSmallVec` doesn't really make it clear that it's a set, and the fact that it's sorted seems like an implementation detail (although this could be made as a guarantee on `ones`).

crates/bevy_ecs/src/storage/sorted_small_vec.rs
```diff
@@ -0,0 +1,259 @@
...
+        self.intersection(other).next().is_none()
+    }
+
+    /// Returns true if all the elements in `self` are also in `other`.
+    pub(crate) fn is_subset(&self, other: &Self) -> bool {
+        self.difference(other).next().is_none()
+    }
+}
+
+impl<const N: usize> Extend<usize> for SortedSmallVec<N> {
```
andriyDev, line 145: This implementation is wrong if `other` is unsorted. So we should either make this just a loop of inserts or add a method instead of a trait impl that makes it clear the behavior is undefined if you don't call it with a sorted input.

crates/bevy_ecs/src/storage/sorted_small_vec.rs
```diff
@@ -0,0 +1,259 @@
...
+    other: &'a SortedSmallVec<N>,
+    i: usize,
+    j: usize,
+}
+
+impl<'a, const N: usize> Iterator for Intersection<'a, N> {
+    type Item = usize;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        let mut res = None;
```
andriyDev, line 185: We don't need this variable. Just put the var inside the if and return None at the bottom.

crates/bevy_ecs/src/storage/sorted_small_vec.rs
```diff
@@ -0,0 +1,259 @@
...
+        while self.i < self.this.len() && self.j < self.other.len() {
+            if (self.i == 0 || self.this.0[self.i] != self.this.0[self.i - 1])
```
andriyDev, line 187: I don't understand this expression. How can `[self.i] == [self.i - 1]`? We should never have the same value twice.

(If we change anything here, we should mirror it with `Difference`)

crates/bevy_ecs/src/storage/sorted_small_vec.rs
```diff
@@ -0,0 +1,259 @@
...
+            }
+        }
+        res
+    }
+}
+
+impl<'a, const N: usize> From<Intersection<'a, N>> for SortedSmallVec<N> {
+    fn from(intersection: Intersection<'a, N>) -> Self {
+        let mut sorted_vec = SortedSmallVec::new_const();
+        for value in intersection.into_iter() {
```
andriyDev, line 207: I think we can use a cheaper impl here. Since we know Intersection returns elements in sorted order, we can just `SmallVec::from_iter(intersection)` I think.

crates/bevy_ecs/src/storage/sorted_small_vec.rs
```diff
@@ -0,0 +1,259 @@
...
+            self.i += 1;
+        }
+        res
+    }
+}
+
+impl<'a, const N: usize> From<Difference<'a, N>> for SortedSmallVec<N> {
+    fn from(difference: Difference<'a, N>) -> Self {
+        let mut sorted_vec = SortedSmallVec::new_const();
+        for value in difference.into_iter() {
```
andriyDev, line 254: Same thing here, we don't need to insert, we can just move elements in order.

andriyDev (commented):
The O(1)'s in the description should be O(n). Adding more elements will always add more to iterate through.

crates/bevy_ecs/src/storage/sorted_small_vec.rs
chescock: +1 for the `Set` suffix.  That better matches names like `HashSet` and `BTreeSet`.  I'd vote for keeping `Sorted` in the name and making it part of the contract, since `Hash` and `BTree` are also implementation details.  But `SortedSmallVecSet` would be awfully verbose.

crates/bevy_ecs/src/storage/sorted_small_vec.rs
chescock: I think you can handle unsorted input and still be reasonably efficient with just:
```rust
self.0.extend(other);
self.0.sort();
self.0.dedup();
```

hymm:
This ci failure seems real.
https://github.com/bevyengine/bevy/actions/runs/12298594497/job/34322375150?pr=16784

crates/bevy_ecs/src/storage/sorted_small_vec.rs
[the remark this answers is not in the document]
Trashtalk217: I thought about it and decided against it. It isn't needed now and I don't think it'll be needed in the near future. If anyone were to need a generalized version later, it would be easy to implement.

Right now it seems like overengineering.

Trashtalk217: I'll go for `SortedVecSet`

crates/bevy_ecs/src/storage/sorted_small_vec.rs
Trashtalk217: I've elected for a simple rename to iter(), implementing `IntoIterator` for `&SortedVecSet` would have me change `set.ones()` into `(&set).into_iter())` everywhere.

crates/bevy_ecs/src/storage/sorted_small_vec.rs
Trashtalk217: I think that this accounted for possible duplicates, but only for self for some reason. Since we're already assuming that SortedVecSets are in fact 'sorted', I don't think that it's too much to ask for zero duplicates either. I've documented it.

crates/bevy_ecs/src/storage/sorted_small_vec.rs
chescock: > implementing `IntoIterator` for `&SortedVecSet` would have me change `set.ones()` into `(&set).into_iter())` everywhere.

Oh, yes, definitely have an `iter()` method!  I just meant it might make sense to have *both*, like `Vec` does.

crates/bevy_ecs/src/query/access.rs
```diff
@@ -428,23 +378,22 @@ impl<T: SparseSetIndex> Access<T> {
             other.component_read_and_writes_inverted,
         ) {
             (true, true) => {
-                self.component_read_and_writes
-                    .intersect_with(&other.component_read_and_writes);
+                self.component_read_and_writes = self
+                    .component_read_and_writes
+                    .intersection(&other.component_read_and_writes)
```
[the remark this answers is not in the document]
chescock: If it helps, I think you can use `retain` to take care of shuffling the elements around, like:

```rust
fn intersect_with(&mut self, other: &Self) {
    let mut j = 0;
    self.0.retain(|current| {
        // Advance past any smaller elements in other
        while j < other.len() && other.0[j] < *current { j += 1; }
        // It's only in the intersection if it's in other,
        // and this is the only place in other it could be
        j < other.len() && other.0[j] == *current
    });
}
```

`difference_with` would be the same but with a `!` in front of the condition.

crates/bevy_ecs/src/storage/sorted_small_vec.rs
```diff
@@ -0,0 +1,259 @@
...
+
+    /// Adds all the elements from `other` into this vector. (skipping duplicates)
+    pub(crate) fn union_with(&mut self, other: &Self) {
+        let mut i = 0;
+        let mut j = 0;
+        while i < self.len() && j < other.len() {
+            match self.0[i].cmp(&other.0[j]) {
+                Ordering::Less => i += 1,
+                Ordering::Greater => {
+                    self.0.insert(i, other.0[j]);
```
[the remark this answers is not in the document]
chescock: While I'm looking at this, I think I got the bit at the end wrong.  `split_at_mut` won't help because the slices would overlap.  Maybe it should be

```rust
} // end of while loop
// Shift any remaining elements from self into place
// This also removes any extra space
self.0.drain(i..w);
// Copy any remaining elements from other into place
// `drain` doesn't shrink the allocation, so this won't need to allocate
self.0.extend_from_slice(&other.0[j..]);
```

crates/bevy_ecs/src/query/access.rs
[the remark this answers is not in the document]
Trashtalk217: Very clever, thank you

crates/bevy_ecs/src/storage/sorted_small_vec.rs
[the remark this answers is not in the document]
Trashtalk217: I actually noticed that all of the 'extend' calls could practically be replaced by 'union_with' as most of them came in the form `set.extend(other_set.iter())`

crates/bevy_ecs/src/storage/sorted_vec_set.rs
```diff
@@ -0,0 +1,250 @@
...
+    pub(crate) fn difference_with(&mut self, other: &Self) {
+        let mut j = 0;
+        self.0.retain(|current| {
+            // Advance past any smaller elements in other
+            while j < other.len() && other.0[j] < *current {
+                j += 1;
+            }
+            // It's only in the difference if it's not in other,
+            // and this is the only place in other it could be
+            j < other.len() && other.0[j] != *current
```
chescock, line 94: I think you need to invert the entire condition for `difference_with`.  For example, if `other` is empty then we want to keep everything and should always return `true`, but `j < other.len()` would be false.  

```suggestion
            j == other.len() || other.0[j] != *current
```

crates/bevy_ecs/src/query/access.rs
```diff
@@ -1,67 +1,30 @@
...
     /// All exclusively-accessed components, or components that may not be
     /// exclusively accessed if `Self::component_writes_inverted` is set.
-    component_writes: FixedBitSet,
+    component_writes: SortedVecSet<ACCESS_SMALL_VEC_SIZE>,
     /// All accessed resources.
-    resource_read_and_writes: FixedBitSet,
+    resource_read_and_writes: SortedVecSet<ACCESS_SMALL_VEC_SIZE>,
     /// The exclusively-accessed resources.
-    resource_writes: FixedBitSet,
+    resource_writes: SortedVecSet<ACCESS_SMALL_VEC_SIZE>,
```
omaskery, line 27: Is it at all feasible, within this PR or a follow up one, to factor the `x` and `x_inverted` pairings into a utility struct that wraps the `SortedVecSet` and `inverted` boolean, encapsulating all of the logic for inverting these sets? 😮

chescock: I vote for doing that as a follow-up!  It would be a good simplification, but it's not related to the goal of this PR and I don't think we should try to do too many things at once.

PR #16785, opened 2024-12-12
Add std derives to SystemParam types

spectria-limina:
# Objective

- Use `Clone` on `SystemParam`, when applicable, in a generic context.

## Solution

-  Add some derives

## Testing

- I ran `cargo test` once.
- I didn't even look at the output.


crates/bevy_ecs/src/system/builder.rs
```diff
@@ -400,6 +401,7 @@ unsafe impl<P: SystemParam, B: SystemParamBuilder<P>> SystemParamBuilder<Vec<P>>
 ///     set.for_each(|mut query| for mut health in query.iter_mut() {});
 /// }
 /// ```
+#[derive(Debug, Default, Copy, Clone)]
```
alice-i-cecile, line 404: Why `Copy` here?

alice-i-cecile:
It would be helpful to reviewers to briefly explain why `Clone` was needed on these types.

BD103 (approved):
I'm fine with the code changes themselves, but some justification would be appreciated! :)

BenjaminBrienen:
If it were just Clone, that would be easier to approve. Why Copy?

spectria-limina:
Sorry for the delay replying. 

Types should [eagerly implement applicable traits](https://rust-lang.github.io/api-guidelines/interoperability.html?highlight=Copy#types-eagerly-implement-common-traits-c-common-traits).

My use case was playing around with some dynamic manipulation of builders in order to inject data into a system during state initialization, and my test usage with `Local` required cloning the builder so that the same builder could be mapped to multiple arguments of a system.

So I think the question is better asked as: why not?

Builder types are only used in the guts of 

As for `Copy`, I think I actually should have made more of them `Copy`. I don't really see any reason why these shouldn't be `Copy` when wrapping a `Copy` type. And [therefore it should] (https://doc.rust-lang.org/std/marker/trait.Copy.html#when-should-my-type-be-copy).

Note that for actual `SystemParam`s, I think `Clone` or especially `Copy` wouldn't work because the semantics are tied to the system and it's state. But this is just the builders, and they are basically fancy named closures.

mockersf:
> So I think the question is better asked as: why not?

More derives mean more build time

> And [therefore it should] (https://doc.rust-lang.org/std/marker/trait.Copy.html#when-should-my-type-be-copy).

A lot of people think this advice is wrong. Deriving `Copy` change the usage of a type from move semantics to copy semantics, and can lead to unexpected behaviour

alice-i-cecile:
I agree with everything but Copy here, and would approve it if that's removed. The explanation about dynamic manipulation of builder types makes sense to me.

BenjaminBrienen:
Copy on everything doesn't work when performance is a huge concern like it is in a game engine. It does make sense if you are coming from a language like Haskell, I think.

alice-i-cecile:
> Copy on everything doesn't work when performance is a huge concern like it is in a game engine. It does make sense if you are coming from a language like Haskell, I think.

There's no performance concerns here: my objections center around explicitness.

spectria-limina:
Removed `Copy`

mockersf (approved):

mockersf:
failure due to GitHub incident

spectria-limina:
Oops. Guess it was actually worse than I thought.

PR #16786, opened 2024-12-12
Misc. docs and renames for niche ECS internals

JaySpruce:
## Objective

Some structs and methods in the ECS internals have names that don't describe their purpose very well, and sometimes don't have docs either.

Also, the function `remove_bundle_from_archetype` is a counterpart to `BundleInfo::add_bundle_to_archetype`, but isn't a method and is in a different file.

## Solution

- Renamed the following structs and added docs:

| Before               | After                        |
|----------------------|------------------------------|
| `AddBundle`          | `ArchetypeAfterBundleInsert` |
| `InsertBundleResult` | `ArchetypeMoveType`          |

- Renamed the following methods:

| Before                                | After                                        |
|---------------------------------------|----------------------------------------------|
| `Edges::get_add_bundle`               | `Edges::get_archetype_after_bundle_insert`   |
| `Edges::insert_add_bundle`            | `Edges::cache_archetype_after_bundle_insert` |
| `Edges::get_remove_bundle`            | `Edges::get_archetype_after_bundle_remove`   |
| `Edges::insert_remove_bundle`         | `Edges::cache_archetype_after_bundle_remove` |
| `Edges::get_take_bundle`              | `Edges::get_archetype_after_bundle_take`     |
| `Edges::insert_take_bundle`           | `Edges::cache_archetype_after_bundle_take`   |

- Moved `remove_bundle_from_archetype` from `world/entity_ref.rs` to `BundleInfo`. I left the function in entity_ref in the first commit for comparison, look there for the diff of comments and whatnot.
- Tidied up docs:
  - General grammar and spacing.
  - Made the usage of "insert" and "add" more consistent.
  - Removed references to information that isn't there.
- Renamed `BundleInfo::add_bundle_to_archetype` to `BundleInfo::insert_bundle_into_archetype` for consistency.

### DIFF
--- a/crates/bevy_ecs/src/archetype.rs
+++ b/crates/bevy_ecs/src/archetype.rs
@@ -582,11 +582,11 @@ impl Archetype {
         self.entities.reserve(additional);
     }
 
-    /// Removes the entity at `index` by swapping it out. Returns the table row the entity is stored
+    /// Removes the entity at `row` by swapping it out. Returns the table row the entity is stored
     /// in.
     ///
     /// # Panics
-    /// This function will panic if `index >= self.len()`
+    /// This function will panic if `row >= self.entities.len()`
     #[inline]
     pub(crate) fn swap_remove(&mut self, row: ArchetypeRow) -> ArchetypeSwapRemoveResult {
         let is_last = row.index() == self.entities.len() - 1;
--- a/crates/bevy_ecs/src/archetype.rs
+++ b/crates/bevy_ecs/src/archetype.rs
@@ -109,7 +109,7 @@ impl ArchetypeId {
     }
 }
 
-/// Used in [`AddBundle`] to track whether components in the bundle are newly
+/// Used in [`ArchetypeAfterBundleInsert`] to track whether components in the bundle are newly
 /// added or already existed in the entity's archetype.
 #[derive(Copy, Clone, Eq, PartialEq)]
 pub(crate) enum ComponentStatus {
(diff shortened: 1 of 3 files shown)


alice-i-cecile (approved):
Really nice set of changes. `AddBundle` in particular has really confused and aggravated me in the past. I agree with all of these changes, and while the type names are verbose, they're both arcane and internal so I think it's a fine tradeoff.

crates/bevy_ecs/src/archetype.rs
```diff
@@ -109,19 +109,20 @@ impl ArchetypeId {
...
 #[derive(Copy, Clone, Eq, PartialEq)]
 pub(crate) enum ComponentStatus {
     Added,
     Existing,
 }
 
-pub(crate) struct AddBundle {
-    /// The target archetype after the bundle is added to the source archetype
+/// Used in [`Edges`] to cache the result of inserting a bundle into the source archetype.
+pub(crate) struct ArchetypeAfterBundleInsert {
```
hymm, line 121: nit: it might be better to name this and other methods `ArchetypeAfterInsertBundle` and `*_insert_bundle()` to match the `insert_bundle` ordering.

hymm (approved):
changes make sense to me.

crates/bevy_ecs/src/archetype.rs
JaySpruce: I went back and forth on this, but I stuck with `bundle_insert` because I thought `insert_bundle` might sound too much like an "action", like `get_archetype_after_insert_bundle` would actually be doing the insert itself. Not married to it though

hymm: I don't really have too strong feelings here either and I don't think it's worth thinking about too hard. The new names are better, so lets just ship it.

Issue #4596, opened 2022-04-25
The `far` member of `PerspectiveProjection` is irrelevant


HackerFoo:
The `far` member of `PerspectiveProjection` is irrelevant, because the far plane is at negative infinite Z

https://github.com/bevyengine/bevy/blob/989fb8a78d2ca6e3fe6fb1ccd7d79ea88868b8c7/crates/bevy_render/src/camera/projection.rs#L25

mockersf:
maybe related to https://github.com/bevyengine/bevy/pull/4457

superdump:
We could still cull meshes to a fake far plane. I just chose not to to preserve behaviour at the time I think. I’m not sure whether we should or shouldn’t cull to a fake far plane. We do for point lights but not for meshes. We could make it optional but then I don’t know what the default should be. I kind of feel like we could make it an Option<f32> and if present then we cull to it, if not, we don’t.

cart:
Yeah I think providing optional configuration for people to cull at a given distance makes a lot of sense.

superdump:
- I've had a go at this in #5282
(The above is a bullet point because that's the only way I know to force github markdown to show the linked PR title.)

Neopallium:
I have been thinking of culling smaller objects faster than large objects.  This way large buildings would still be visible, but smaller objects inside or around it wouldn't be rendered.

Would it be difficult to do that in the far-plane culling logic?

superdump:
What do you mean by faster? All meshes are chilled against the visible volume every frame.

Neopallium:
Sorry, by "faster" I mean at a closer distance.  A small object (scale dimensions < 1.0) will not be as visible (size on screen) at a distance of 50.0 from the camera vs a larger object (scale dim. > 1.0).

I have a 3d scene that has +40,000 objects (over 100,000 ECS entities).  The object meshes are generated from primitive profile/path parameters and can have different textures on different faces.   So right now one object could have multiple meshes.

I had tried enabling the far plane culling like in your PR to see if it would help with FPS.  It does, but there is a hard cut-off, where even large objects go invisible.

I am now using an adaptive culling based on distance and the Transform.scale of the mesh.  For the scene that I have tested with (large buildings with many smaller objects inside), it allows the buildings that are far away to still be visible while hiding the smaller objects that are not as noticable (even when the build has lots of windows).

Issue #16746, opened 2024-12-10
PerspectiveProjection - Incorrect behaviour for `far`


ptsd:
## Bevy version

0.15.0

## Relevant system information

- `cargo 1.82.0 (8f40fc59f 2024-08-21)`
- `MacOS 13.6.9 (22G830)`

```ignore
`AdapterInfo { name: "AMD Radeon R9 M290X", vendor: 0, device: 0, device_type: DiscreteGpu, driver: "", driver_info: "", backend: Metal }`
```

## What you did

Repro:
```rust
use bevy::prelude::*;
use bevy::window::WindowResolution;

fn main() {
    App::new()
        .add_plugins(DefaultPlugins.set(WindowPlugin {
            primary_window: Some(Window {
                resolution: WindowResolution::new(640., 480.),
                ..default()
            }),
            ..default()
        }))
        .add_systems(Startup, setup)
        .run();
}

fn setup(
    mut commands: Commands,
    mut meshes: ResMut<Assets<Mesh>>,
    mut materials: ResMut<Assets<StandardMaterial>>,
) {
    commands.spawn((
        Mesh3d(meshes.add(Cuboid::new(0.5, 0.5, 0.5))),
        MeshMaterial3d(materials.add(StandardMaterial::from(Color::srgb(0.8, 0., 0.)))),
        Transform::from_xyz(0., 0., 0.),
    ));

    commands.spawn((
        Camera3d::default(),
        Transform::from_xyz(5., 5., 5.).looking_at(Vec3::ZERO, Dir3::Y),
        Projection::from(PerspectiveProjection {
            far: 0.5,
            ..default()
        }),
    ));
}

```

## What went wrong

- Cube is visible, but `far: 0.5` should cull the cube as it is more than `0.5` units from the camera.
 
## Additional information

`PerspectiveProjection { far: 0.5 }`
![Image](https://github.com/user-attachments/assets/61cdbd2f-6c75-4d0d-aeba-4dc36eccbfb1)

Works as expected with `Orthographic` projection:
`OrthographicProjection { far: 1000.0 }`
![Image](https://github.com/user-attachments/assets/2fef05ce-4adc-42af-8b7f-d49a1476e5b8)

`OrthographicProjection { far: 0.5 }`
![Image](https://github.com/user-attachments/assets/e939866e-2b68-43b1-8b2a-247fccf97195)

[...]

PR #8205, opened 2023-03-25
add option for far plane culling (#8173)

NotoriousENG:
# Objective

- Adds option to `Projection` for far plane culling, and updates `check_visibility` to perform culling based on this property
- Before this change Frustum Culling would only account for near, left, right, top, and bottom planes with no trivial user facing overrides
- "Fixes #8173"

## Solution

- In addition to providing the option and updating check visibility, the new `use_far_culling: bool` defaults to false to prevent breaking changes
- with the use_far_culling option enabled, the perspective projection will behave more like the defaults in Godot, Unity, and Unreal where all 6 planes of the frustum are used for visibility tests
- this can lead to performance improvements on weaker targets like mobile with minimal effort

---

## Changelog

- Added use_far_culling option for camera_3d projections. Consider enabling this option for a quick performance boost for weaker targets until occlusion culling is ready.

## Example
Checkout https://github.com/NotoriousENG/bevy_cull_bug/tree/feature/fix_example and build against this PR for bevy to see in action
 
Image Showing Far Plane culling is working
image

Issue #8173, opened 2023-03-23
Bevy 0.10.0 Frustum Culling is not working


NotoriousENG:
## Bevy version

0.10.0

## \[Optional\] Relevant system information

Windows 11 (Desktop)
```ignore
`AdapterInfo { name: "NVIDIA GeForce GTX 1080", vendor: 4318, device: 7040, device_type: DiscreteGpu, driver: "NVIDIA", driver_info: "516.40", backend: Vulkan }`
```
Fedora (Laptop)
```ignore
`AdapterInfo { name: "AMD Radeon Vega 8 Graphics (RADV RAVEN)", vendor: 4098, device: 5597, device_type: IntegratedGpu, driver: "radv", driver_info: "Mesa 22.3.7", backend: Vulkan }`
```

## What you did

Minimal Code Example:  https://github.com/NotoriousENG/bevy_cull_bug

## What went wrong

I was expecting objects behind the far plane to be culled base off of the bevy 0.6.0 changelog https://bevyengine.org/news/bevy-0-6/ 

I'm wondering if it may be due to a frustum struct being empty? Not quite sure if it should be created as part of building a Camera3dBundle?

No objects are being culled, and I can still see beyond where I would expect the view to end.

Originally found it while working here: 
* https://github.com/NotoriousENG/goon_game

And posted to the rendering discussion on Discord:
* https://discord.com/channels/691052431525675048/866787577687310356/1088250126448996432

## Additional information
Please build and run the minimal code example to see my issue https://github.com/NotoriousENG/bevy_cull_bug 

Using
```rs
const FAR: f32 = 10.00;
```

for a camera
```rs
let cam_transform = Transform::from_xyz(0.0, 5.0, -10.0).looking_at(Vec3::ZERO, Vec3::Y);
// camera
commands.spawn(Camera3dBundle {
    projection: PerspectiveProjection {
        far: FAR,
        ..default()
    }
    .into(),
    transform: cam_transform.clone(),
    ..default()
});
```

and cubes spawned such that 
```rs
// cubes
for i in 0..20 {
    let z = i as f32 + 0.5;
    if z > cutoff_z {
        // these should be culled by the frustum culling?
        println!("cube {} is behind the far cutoff {} at {}", i, cutoff_z, z);
    }
    commands.spawn(PbrBundle {
        mesh: meshes.add(Mesh::from(shape::Cube { size: 0.5 })),
[...]

PR #16789, opened 2024-12-12
Introduce support for far culling when using `PerspectiveProjection`

ptsd:
## Note
An update of https://github.com/bevyengine/bevy/pull/8205 for 0.15 and with review suggestions.
All credit to @NotoriousENG

# Objective
* Before this change Frustum Culling on cameras with `PerspectiveProjection` would only account for near, left, right, top, and bottom planes with no trivial user facing overrides
* Fixes:
  * https://github.com/bevyengine/bevy/issues/8173 
  * https://github.com/bevyengine/bevy/issues/16746

## Solution
* Changes return of `CameraProjection::far()` and it's impls from `f32` to `Option<f32>`
* Adds sensible (hopefully non bc breaking) `far` defaults to `OrthographicProjection` and `PerspectiveProjection`
  * `OrthographicProjection` = `Some(OrthographicProjection::DEFAULT_FAR_PLANE)`
     Equivalent to existing behaviour of far clipping at `1000.0`
  * `PerspectiveProjection` = `None`
     Equivalent to existing behaviour of not far clipping at all
* Updates `Frustum::from_clip_from_world_custom_far` to account for `far` becoming optional
* Updates `bevy_render::view::visibility::check_visibility` to perform far culling when appropriate
* Updates users of the above changes as appropriate

## Changelog
* Existing instantiations of  `OrthographicProjection` and `PerspectiveProjection` with an explicit `far` value will need to wrap the value in `Some`
* Custom impls of `CameraProjection` will need updating to account for the changed return type of `far()`
* Usages of `Frustum::from_clip_from_world_custom_far` in downstream code will need updating to account for the changed type of the `far` parameter

### DIFF
--- a/crates/bevy_render/src/camera/projection.rs
+++ b/crates/bevy_render/src/camera/projection.rs
@@ -92,7 +92,7 @@ pub trait CameraProjection {
             &clip_from_world,
             &camera_transform.translation(),
             &camera_transform.back(),
-            self.far().unwrap_or_default(),
+            self.far(),
         )
     }
 }
@@ -176,7 +176,7 @@ pub struct PerspectiveProjection {
     ///
     /// Objects farther from the camera than this value will not be visible.
     ///
-    /// Defaults to None (infinite far plane)
+    /// Defaults to `None` (far plane is ignored for visibility checks)
     pub far: Option<f32>,
 }
 
@@ -260,7 +260,7 @@ impl Default for PerspectiveProjection {
         PerspectiveProjection {
             fov: core::f32::consts::PI / 4.0,
             near: 0.1,
-            far: Option::None,
+            far: None,
             aspect_ratio: 1.0,
         }
     }
@@ -354,7 +354,9 @@ pub struct OrthographicProjection {
     ///
     /// Objects further than this will not be rendered.
     ///
-    /// Defaults to None (infinite far plane)
+    /// Defaults to [`OrthographicProjection::DEFAULT_FAR_PLANE`].
+    ///
+    /// ie: `Some(1000.0)`.
     pub far: Option<f32>,
     /// Specifies the origin of the viewport as a normalized position from 0 to 1, where (0, 0) is the bottom left
     /// and (1, 1) is the top right. This determines where the camera's position sits inside the viewport.
@@ -408,7 +410,7 @@ impl CameraProjection for OrthographicProjection {
             self.area.max.y,
             // NOTE: near and far are swapped to invert the depth range from [0,1] to [1,0]
             // This is for interoperability with pipelines using infinite reverse perspective projections.
-            self.far.unwrap_or_default(),
+            self.far.unwrap_or(Self::DEFAULT_FAR_PLANE),
             self.near,
         )
     }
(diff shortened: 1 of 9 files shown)


crates/bevy_render/src/primitives/mod.rs
```diff
@@ -241,18 +241,24 @@ impl Frustum {
 
     /// Returns a frustum derived from `clip_from_world`,
     /// but with a custom far plane.
+    /// If `far` is `None` behavior is identical to [`Frustum::from_clip_from_world`]
```
ptsd, line 244: This seems like the least expectation defying change but not sure it's the best approach?

crates/bevy_render/src/view/visibility/mod.rs
```diff
@@ -575,17 +583,25 @@ pub fn check_visibility<QF>(
...
+                        let use_far_culling = maybe_projection.is_some_and(|p| p.far().is_some());
+
                         let world_from_local = transform.affine();
                         let model_sphere = Sphere {
                             center: world_from_local.transform_point3a(model_aabb.center),
                             radius: transform.radius_vec3a(model_aabb.half_extents),
                         };
                         // Do quick sphere-based frustum culling
-                        if !frustum.intersects_sphere(&model_sphere, false) {
+                        if !frustum.intersects_sphere(&model_sphere, use_far_culling) {
```
ptsd, line 595: I thought this would preserve existing behaviour when using the default projections, but maybe I'm misunderstanding as `OrthographicProjection` previously far culled even though we always passed `false` here?

crates/bevy_sprite/src/picking_backend.rs
```diff
@@ -114,7 +114,10 @@ fn sprite_picking(
         let Ok(cursor_ray_world) = camera.viewport_to_world(cam_transform, pos_in_viewport) else {
             continue;
         };
-        let cursor_ray_len = cam_ortho.far - cam_ortho.near;
+        let cursor_ray_len = cam_ortho
+            .far
+            .unwrap_or(OrthographicProjection::DEFAULT_FAR_PLANE)
```
ptsd, line 119: This is definitely wrong and I'm open to suggestions. Would also allow `DEFAULT_FAR_PLANE` not to be `pub`.

crates/bevy_render/src/camera/projection.rs
```diff
@@ -549,7 +553,7 @@ impl OrthographicProjection {
         OrthographicProjection {
             scale: 1.0,
             near: 0.0,
-            far: 1000.0,
+            far: Some(Self::DEFAULT_FAR_PLANE),
```
IceSentry, line 556: I feel like this should default to None since bevy currently doesn't do far plane culling.

IceSentry:
I added a few comments. Overall the direction looks good. I'd like @NthTensor or @aevyrie input on the picking part.

crates/bevy_render/src/camera/projection.rs
ptsd: This is the orthographic case, where bevy does currently do far culling. See https://github.com/bevyengine/bevy/issues/16746 (Additional information)

crates/bevy_render/src/primitives/mod.rs
ptsd: Changed.

crates/bevy_render/src/camera/projection.rs
aevyrie: Does `None` even make sense for ortho? It doesn't have the ability to use infinite z, because depth is linear.

BenjaminBrienen:
Closes https://github.com/bevyengine/bevy/issues/4596

PR #16790, opened 2024-12-12
box-shadow clipping fix

ickshonpe:
# Objective

Instead of clipping the non-visable sections of box-shadows, the shadow is scaled to fit into the remaining area after clipping because the normalized coordinates that are meant to border the unclipped subsection of the shadow are always set to `[Vec2::ZERO, Vec2::X, Vec2::ONE, Vec2::Y]`, 

## Solution

Calculate the coordinates for the corners of the visible area.

## Testing

Test app:

```rust
use bevy::color::palettes::css::RED;
use bevy::color::palettes::css::WHITE;
use bevy::prelude::*;

fn main() {
    App::new()
        .add_plugins(DefaultPlugins)
        .add_systems(Startup, setup)
        .run();
}

    commands.spawn(Camera2d);
    commands
        .spawn(Node {
            ..Default::default()
        })
        .with_children(|commands| {
            commands
                .spawn((
                    Node {
                        width: Val::Px(100.),
                        height: Val::Px(100.),
                        margin: UiRect {
                            left: Val::Px(100.),
                            top: Val::Px(300.),
                            ..Default::default()
                        },
                        overflow: Overflow::clip(),
                        ..Default::default()
                    },
                    BackgroundColor(WHITE.into()),
                ))
                .with_children(|commands| {
                    commands.spawn((
                        Node {
                            left: Val::Px(50.),
                            top: Val::Px(50.),
                            width: Val::Px(100.),
                            height: Val::Px(100.),
                            ..Default::default()
                        },
                        BackgroundColor(RED.into()),
                        BoxShadow::from(ShadowStyle {
                            x_offset: Val::ZERO,
                            y_offset: Val::ZERO,
                            spread_radius: Val::Px(50.),
                            blur_radius: Val::Px(6.),
                            ..Default::default()
                        }),
                    ));
                });
        });
}
```

Main:
bad_shadow

This PR:
clipped_shadow


UkoeHB (commented):
Btw possibly related to this PR, this line looks odd to me:
```rust
let positions = QUAD_VERTEX_POSITIONS
    .map(|pos| (box_shadow.transform * (pos * rect_size).extend(1.)).xyz());
```
Shouldn't it be `transform + (pos * rect_size)`?

ickshonpe:
> Btw possibly related to this PR, this line looks odd to me:
> 
> ```rust
> let positions = QUAD_VERTEX_POSITIONS
>     .map(|pos| (box_shadow.transform * (pos * rect_size).extend(1.)).xyz());
> ```
> 
> Shouldn't it be `transform + (pos * rect_size)`?

transform is a 4x4 matrix so it's not possible add it to to a vec3 

> Btw possibly related to this PR, this line looks odd to me:
> 
> ```rust
> let positions = QUAD_VERTEX_POSITIONS
>     .map(|pos| (box_shadow.transform * (pos * rect_size).extend(1.)).xyz());
> ```
> 
> Shouldn't it be `transform + (pos * rect_size)`?

`transform` is a 4x4 matrix so addition doesn't work here, to apply the matrix transformation we have to extend the scaled position to a vec4 and multipy.

UkoeHB (approved):

crates/bevy_ui/src/render/box_shadow.rs
```diff
@@ -479,7 +479,26 @@ pub fn prepare_shadows(
...
+                        Vec2::new(
+                            uinode_rect.max.x + positions_diff[2].x,
+                            uinode_rect.max.y + positions_diff[2].y,
+                        ),
+                        Vec2::new(
+                            uinode_rect.min.x + positions_diff[3].x,
+                            uinode_rect.max.y + positions_diff[3].y,
+                        ),
+                    ]
+                    .map(|pos| pos / uinode_rect.max);
```
andriyDev, line 500: Should this be `max`? or should it be `size()`? I think a non-zero x_offset or y_offset will show an error.

JMS55 (approved):

andriyDev (approved):

PR #16796, opened 2024-12-13
Support scale factor for image render targets

msvbg:
# Objective

I have something of a niche use case. I have a camera rendering pixel art with a scale factor set, and another camera that renders to an off-screen texture which is supposed to match the main camera exactly. However, when computing camera target info, Bevy [hardcodes a scale factor of 1.0](https://github.com/bevyengine/bevy/blob/116c2b02fe8a7589d1777af7dabd84dc756b5b0d/crates/bevy_render/src/camera/camera.rs#L828) for image targets which means that my main camera and my image target camera get different `OrthographicProjections` calculated.

## Solution

This PR adds an `ImageRenderTarget` struct which allows scale factors to be specified. 

## Testing

I tested the affected examples on macOS and they still work. This is an additive change and should not break any existing code, apart from what is trivially fixable by following compiler error messages.

---

## Migration Guide

`RenderTarget::Image` now takes an `ImageRenderTarget` instead of a `Handle<Image>`. You can call `handle.into()` to construct an `ImageRenderTarget` using the same settings as before.

### DIFF
--- a/crates/bevy_render/src/camera/camera.rs
+++ b/crates/bevy_render/src/camera/camera.rs
@@ -28,7 +28,7 @@ use bevy_ecs::{
     world::DeferredWorld,
 };
 use bevy_image::Image;
-use bevy_math::{ops, vec2, Dir3, Mat4, Ray3d, Rect, URect, UVec2, UVec4, Vec2, Vec3};
+use bevy_math::{ops, vec2, Dir3, FloatOrd, Mat4, Ray3d, Rect, URect, UVec2, UVec4, Vec2, Vec3};
 use bevy_reflect::prelude::*;
 use bevy_render_macros::ExtractComponent;
 use bevy_transform::components::{GlobalTransform, Transform};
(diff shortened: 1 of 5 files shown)


JMS55:
This gets tricky, because we need something similar (but not exactly the same) for temporal upscaling like DLSS. Different passes of the same camera will need different scale factors (e.g. main pass, vs certain post processing and UI). It might get confusing if we have both this scale factor on the camera target, and another, similar property on the camera itself.

msvbg:
I think this solution makes the most sense given what Bevy looks like today, since it mirrors the capabilities of the window render target.

DLSS may require a more complicated refactor of camera logic, but that is not in scope for this PR in my opinion. Dynamic scale factors across passes sounds like a different feature altogether to me. This PR is a relatively simple case of unblocking functionality that isn’t possible otherwise.

JMS55:
Right, it's just a matter of figuring out what to name this that won't be confusing when we add camera scale factor in the future.

msvbg:
Maybe something like `downscale_factor`/`upscale_factor` captures the two needs? I'm trying to downscale and you're trying to upscale.

Though I'm not too familiar with how DLSS works and I shouldn't try to design for it here. I'm fine with breaking naming/API changes here if/when DLSS lands. I'd just like this to be in 0.16 in some shape or form so I don't have to maintain a fork.

IceSentry:
Yeah, once we have a way to manage scale between passes I think it should all be relative to the scale of the target anyway. So I think both can coexist. Most users will most likely leave it to the default anyway.

IceSentry (approved):

alice-i-cecile (approved):

PR #16800, opened 2024-12-13
Fix registering all reflection types that are components as reflection components

### LINKED ISSUE #16659: The reflection of bevy_render::sync_world::TemporaryRenderEntity doesn't show up as a Component
## Bevy version

0.15.0

## What you did

I'm extracting type information using the reflection API. I'm trying to get a list of all components with code like this:

```
for ty in type_registry.iter() {
    if ty.contains::<ReflectComponent>() {
        println!("component: {ty:?}");
    } else {
        println!("not a component: {ty:?}");
    }
}
```

## What went wrong

There are a few components that don't have a ReflectComponent, for example `bevy_render::sync_world::TemporaryRenderEntity`:

```
TypeRegistration { type_info: Struct(StructInfo { ty: bevy_render::sync_world::TemporaryRenderEntity, generics: Generics([]), fields: [], field_names: [], field_indices: {}, custom_attributes: {}, docs: Some(" Marker component that indicates that its entity needs to be despawned at the end of the frame.") }) }
```

There are probably more of those. I can't find anything specific that differentiates them from other components where this code works.

anlumo:
# Objective

Fixes #16659

## Solution

- I just added all the `#[reflect(Component)]` attributes where necessary.

## Testing

I wrote a small program that scans the bevy code for all structs and enums that derive `Component` and `Reflect`, but don't have the attribute `#[reflect(Component)]`.

I don't know if this testing program should be part of the testing suite of bevy. It takes a bit of time to scan the whole codebase. In any case, I've published it [here](https://github.com/anlumo/bevy-reflect-check).


BD103 (approved):
Looks good!

MrGVSV (approved):
That's a neat program! We'd love to do something similar for the Bevy linter if you’re ever interested in helping there :)

alice-i-cecile:
Import merge conflict was a bit tricky; let's see if CI likes that.

anlumo:
My code scanner had a bug where it missed

```rust
#[cfg_attr(feature = "bevy_reflect", derive(Reflect), reflect(Debug))]
```

since it only scanned for top-level derives. I found only two instances of the problem this PR fixes, both in the gamepad code.

alice-i-cecile:
Thanks for cleaning this up <3 Merging.

Issue #5069, opened 2022-06-22
A query filter for Asset updates


nicopap:
## What problem does this solve or what need does it fill?

A major pain point of using `Handle<Asset>` as components in the ECS currently is that the assets pointed-to by the handle can change, and there is no simple way of reacting to changes to the asset in the context of the ECS.

For example: AABB calculation. We have a `Handle<Mesh>` and `Aabb` component in the ECS. We want to update `Aabb` whenever the `Mesh` for the `Handle` is updated. The `Handle` is updated in two circumstances:
1. The user changes the `Handle` the entity has by replacing it in a system with a  `Query<&mut Handle<Mesh>>` parameter
2. The user changes the `Mesh` in the `Assets<Mesh>` in a system with a `ResMut<Assets<Mesh>>`  parameter

To react to (1), it's trivial: just add a `Changed<Handle<Mesh>>` to the system that updates the `Aabb`.

To react to (2), it becomes harder. The only way to react to asset changes is to use the `AssetEvent::Modified { handle: Handle<Mesh> }` event. This event only holds a `Handle<Mesh>`, so you either need to iterate over ALL entities with a `Handle<Mesh>` and updates the ones that are equal to the `Modified` field, or keep a `HashMap` mapping handles to entities, and other gory details required to maintain this `HashMap` properly.

Most of asset changes are done through (2), which makes it very difficult to react to asset changes in the general case.

This was observed in https://github.com/bevyengine/bevy/pull/4944 and also in personal project related to scene hot-reloading.

## What solution would you like?

Introducing a `AssetChanged<A>` query filter that is `true` for entities with a `Handle<A>` for which the underlying asset got modified since the last time the system ran.

## What alternative(s) have you considered?

In the case of `Mesh`, it seems that moving Aabb calculation to the generation of the Mesh and keeping it as an optional `Mesh` field is a correct solution. This would fall into the category of "asset pre-processing" which is both possible today (through a constructor) and subject to future possible improvements.

But this is only a specific solution to a specific problem. In general, it seems reasonable to be able to bridge the gap between assets and ECS with such a query filter.

## Additional context

See discussions:
- https://github.com/bevyengine/bevy/pull/4944
- https://github.com/bevyengine/bevy/discussions/3972
- https://github.com/bevyengine/bevy/pull/4980

richchurcher:
[...]

PR #13343, opened 2024-05-12
constrain WorldQuery::get_state to only use &Components

Victoronz:
# Objective

Passing `&World` in the `WorldQuery::get_state` method is unnecessary, as all implementations of this method in the engine either only access `Components` in `&World`, or do nothing with it.
It can introduce UB by necessitating the creation of a `&World` from a `UnsafeWorldCell`.
This currently happens in `Query::transmute_lens`, which obtains a `&World` from the internal `UnsafeWorldCell` solely to pass to `get_state`. `Query::join` suffers from the same issue.
Other cases of UB come from allowing implementors of `WorldQuery` to freely access `&World`, like in the `bevy-trait-query` crate, where a [reference to a resource is obtained](https://github.com/JoJoJet/bevy-trait-query/blob/0c0e7dd646b4a77fa3496ef5e9686107d17fdd1e/src/lib.rs#L445) inside of [`get_state`](https://github.com/JoJoJet/bevy-trait-query/blob/0c0e7dd646b4a77fa3496ef5e9686107d17fdd1e/src/one.rs#L245), potentially aliasing with a `ResMut` parameter in the same system.

`WorldQuery::init_state` currently requires `&mut World`, which doesn't suffer from these issues.
But that too can be changed to receive a wrapper around `&mut Components` and `&mut Storages` for consistency in a follow-up PR.

## Solution

Replace the `&World` parameter in `get_state` with `&Components`.

## Changelog

 `WorldQuery::get_state` now takes `&Components` instead of `&World`.
 The `transmute`, `transmute_filtered`, `join` and `join_filtered` methods on `QueryState` now similarly take `&Components` instead of `&World`.

## Migration Guide

Users of `WorldQuery::get_state` or `transmute`, `transmute_filtered`, `join` and `join_filtered` methods on `QueryState` now need to pass `&Components` instead of `&World`. 
`&Components` can be trivially obtained from either `components` method on `&World` or `UnsafeWorldCell`.
For implementors of `WorldQuery::get_state` that were accessing more than the `Components` inside `&World` and its methods, this is no longer allowed.

PR #5080, opened 2022-06-23
Add the AssetChanged query filter

nicopap:
# Objective

* Closes #5069
* ~~Blocked on: #5105~~

## Solution

* Add the `AssetChanges<T>` resource.
* Add the `AssetChanged<T>` world query
* Add the `AssetOrHandleChanged<T>` world query type alias

`AssetChanges<T>` resource does the following:
* It is a `HashMap<HandleId, Tick>`
* If `AssetChanges<T>` is present, each `AssetEvent<T>` emitted in `Assets::<T>::asset_event_system` will also update the relevant entry in the `AssetChanges` hash map.

`AssetChanged<T>` does the following:
* It acts like `&Handle<T>` with one difference:
* Its `init_state` adds the `AssetChanges<T>` resource in the world if not present
* Its `init_fetch` fetches the `AssetChanges<T>` resource in addition to the `&Handle<T>`. (panics if the resource was removed)
* It checks if the handle was updated since last time the system ran, and filters on it

#### Performance

* `AssetChanges<T>` is a fairly costly operation, since it does a hashmap lookup per entity iterated.
* We do not maintain and update an `AssetChanges<T>` for assets that do not have a `AssetChanged<T>` world query.
* The fact that `AssetChanged` both accesses (and locks) a component `Handle<T>` and a resource `AssetChanges` is unique and might come off as surprising. But I don't see how this could introduce a soundness issue.
* If using "Single arc tree" in #8624 turns out to be impossible for change tick, a few optimizations are possibles:
  * bloom filters seem promising
  * a coarse-grained asset tick, that we could compare directly instead of fetching individual asset change ticks.

#### Relation with Asset v2 (#8624)

The "Single Arc tree" bullet point in "Bevy Assets v2" section mentions we could attach metadata to handles. Using this for change ticks would make the implementation much more performant, and simplify the implementation.

#### `AssetChanges` Update time

I chose to update the `change_ticks` hash map during asset event
propagation rather than when accessing the asset, because we need access
somehow to the current `change_tick`. Updating the tick hash map during
modification would require passing the `change_tick` to all methods of
`Assets` that accepts a `&mut self`. Which would be a huge ergonomic
loss.

As a result, the asset change expressed by `AssetChanged` is only
effective after `asset_event_system` ran. This is in line with how
`AssetEvents` currently works, but might cause a 1 frame lag for asset
change detection.

## Alternatives

1. #4944  (reverted due to perf issues)
[...]

PR #16810, opened 2024-12-13
Add `AssetChanged` query filter

tychedelia:
# Objective

Implement a new `AssetChanged` query filter that allows users to query for entities whose related assets may have changed.

- Closes https://github.com/bevyengine/bevy/issues/5069
- Unblocks #16420. Currently, `cold-specialization`, a key rendering optimization for unlocking ancillary benefits of the retained render world, is blocked on being unable detect all scenarios in which an entity's mesh/material changes using events and observers. An `AssetChanged` filter will drastically simplify our implementation and be more robust to future changes.

Originally implemented by @nicopap in #5080.

## Solution

- Adds a new `AssetChanged` query filter that initializes a `AssetChanges<A>` resource that tracks changed assets and ticks in `asset_events`.
- ~Reverts #13343 and changes the api of `get_state` to accept `impl Into<UnsafeWorldCell<'w>>` to allow accessing the `AssetChanges<A>` resource.~
- Adds a `AsAssetId` trait used for newtype handle wrappers (e.g. `Mesh3d`) that allows associating a component with the underlying `Asset` it represents.

## Testing

- Tests are added for `AssetChanged`.
- TBD on performance. We are going to add this `Mesh3d` and `MeshMaterial3d` (etc) in the renderer. Long term wins in render performance this unblocks should swamp tracking overhead for any realistic workload.

## Migration Guide

- The `asset_events` system is no longer public. Users should order their systems relative to the `AssetEvents` system set.


Victoronz:
Why does this revert the `get_state` PR? In the `get_state` impl of this PR, I only see a call to `resource_id` on `world`, which is already available on `Components`.

tychedelia:
> Why does this revert the `get_state` PR? In the `get_state` impl of this PR, I only see a call to `resource_id` on `world`, which is already available on `&Components`.

Because I didn't know that. Will change it back!

Victoronz:
Separately, how should the resource access in `init_fetch` be reasoned about?
A type being private means it cannot be mentioned in user code, however you do not have to mention a type to obtain it via public API! In this case, this is possible via id-based resource fetching.
`UnsafeWorldCell::get_resource` gives its caller the responsibility to ensure it has access to that resource, however this safety contract cannot be passed on to users of the filter.
For both user, library and other in-engine areas, how should we communicate about this?

crates/bevy_asset/src/asset_changed.rs
```diff
@@ -0,0 +1,451 @@
...
+        let is_newer = |tick: &Tick| tick.is_newer_than(self.last_run, self.this_run);
+        let id = handle.as_asset_id();
+
+        self.change_ticks.get(&id).is_some_and(is_newer)
+    }
+}
+
+/// Filter that selects entities with a `A` for an asset that changed
+/// after the system last ran, where `A` is a component that implements
+/// [`AsAssetId`].
```
pcwalton, line 78: What about components that store handles to multiple assets of the same type? We don't need to do that right now, but we might have to revisit it later. In my app I have components with handles to dozens of `AnimationClip`s for example. I think you could maybe just have `AsAssetId`'s method return `-> impl Iterator<Item = AssetId>`.

This isn't blocking, just a suggestion.

crates/bevy_asset/src/asset_changed.rs
```diff
@@ -0,0 +1,451 @@
...
+/// will only pick up asset changes in schedules following `AssetEvents` or the
+/// next frame. Consider adding the system in the `Last` schedule after [`AssetEvents`] if you need
+/// to react without frame delay to asset changes.
+///
+/// # Performance
+///
+/// When at least one `A` is updated, this will
+/// read a hashmap once per entity with a `A` component. The
+/// runtime of the query is proportional to how many entities with a `A`
+/// it matches.
```
pcwalton, line 106: Nice, I like the way you document the performance.

tychedelia:
> In this case, this is possible via id-based resource fetching.

How would the user forge an id if they do not have access to the underlying type? `UnsafeWorldCell::get_resource` already documents the aliasing requirements as an unsafe API. If users are iterating or creating random opaque `ComponentId`s and getting mutable access they trivially violate that contract by not knowing what resource they are touching.

pcwalton:
Overall this looks fine to me, but as I mentioned we should have an ECS expert look over the `get_state` signature change.

crates/bevy_asset/src/lib.rs
```diff
@@ -141,13 +141,15 @@
 // FIXME(3492): remove once docs are ready
 // FIXME(15321): solve CI failures, then replace with `#![expect()]`.
 #![allow(missing_docs, reason = "Not all docs are written yet, see #3492.")]
+#![allow(unsafe_code)]
```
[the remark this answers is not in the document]
tychedelia: That's what the other crates seem to do, I'm not unsure about our policy here.

crates/bevy_asset/src/asset_changed.rs
tychedelia: The bigger issue is how to handle the associated type. We might have to use some kind of indirection via UntypedAssetId instead to be able to handle multiple kinds of assets.

crates/bevy_asset/src/asset_changed.rs
```diff
@@ -0,0 +1,446 @@
...
+            }
+        }
+    }
+
+    unsafe fn fetch<'w>(_: &mut Self::Fetch<'w>, _: Entity, _: TableRow) -> Self::Item<'w> {}
+
+    #[inline]
+    fn update_component_access(state: &Self::State, access: &mut FilteredAccess<ComponentId>) {
+        <&A>::update_component_access(&state.asset_id, access);
+        access.add_resource_read(state.resource_id);
```
chescock, line 220: I don't think this line is quite enough.  We need `Query` to report the access in `Access<ArchetypeComponentId>` for it to be used by the scheduler, and it currently does that in `QueryState::update_archetype_component_access()`, which only looks at `component_reads_and_writes()` and `component_writes()`.  

I think you need modify `QueryState::new_with_access()` to look at `resource_reads_and_writes()` and `resource_writes()` and convert them to `ArchetypeComponentId`s.  

I think you also need to modify `FilteredAccess::is_compatible()` to consider resources to conflict even when the filters don't overlap.  (That may need to get reworked in the future for resources-as-components, but the semantics seem reasonable.)  

It would be good to add a unit test that verifies that the access is set up correctly.  Maybe check that `fn system(r: ResMut<AssetChanges<A>>, q: Query<Entity, AssetChanged<A>>)` panics when passed to `assert_system_does_not_conflict`, and again with the parameters reversed.

tychedelia: Alright thanks for the close review. Will add a test to validate.

crates/bevy_asset/src/lib.rs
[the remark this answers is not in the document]
alice-i-cecile: Much prefer local allows here.

crates/bevy_asset/src/asset_changed.rs
```diff
@@ -0,0 +1,458 @@
...
+    storage::{Table, TableRow},
+    world::unsafe_world_cell::UnsafeWorldCell,
+};
+use bevy_utils::tracing::error;
+use bevy_utils::HashMap;
+use core::marker::PhantomData;
+use disqualified::ShortName;
+
+#[derive(Resource)]
+pub(crate) struct AssetChanges<A: Asset> {
```
alice-i-cecile, line 21: Basic doc strings for internal types please, especially linking the various parts of the machinery together. It's much easier to review and maintain!

tychedelia: Added docs also noting the safety implications of this resource.

crates/bevy_asset/src/asset_changed.rs
tychedelia: So just added a test and without making any other changes am getting the following error:
```rust
error[B0001]: Query<MyComponent, AssetChanged<MyComponent>> in system bevy_asset::asset_changed::tests::resource_conflict::system accesses component(s)AssetChanges<MyAsset> in a way that conflicts with a previous system parameter. Consider using `Without<T>` to create disjoint Queries or merging conflicting Queries into a `ParamSet`. See: https://bevyengine.org/learn/errors/b0001
```

Thoughts?

chescock: Oh, right, `ResMut` does unfiltered access, so `FilteredAccess::is_compatible()` will see that the filters overlap and catch it.  Great!  So that won't be a problem unless someone tries to do *mutable* resource access from a query.  

I think it's just the `Access<ArchetypeComponentId>` that would be an issue, then.  I'm not coming up with a simple way to write a test for that, though.

tychedelia: So, I think it's just adding the following to `QueryState::update_archetype_component_access`? https://github.com/bevyengine/bevy/pull/16810/commits/d0130a34108ece47c77232d3b4a6e85874090547

crates/bevy_asset/src/asset_changed.rs
```diff
@@ -0,0 +1,465 @@
+//! Define the [`AssetChanged`] query filter.
```
alice-i-cecile, line 1:
```suggestion
//! Defines the [`AssetChanged`] query filter.
```

crates/bevy_asset/src/asset_changed.rs
```diff
@@ -0,0 +1,465 @@
...
+//!
+//! Like [`Changed`](bevy_ecs::prelude::Changed), but for [`Asset`]s.
```
alice-i-cecile, line 3:
```suggestion
//! Like [`Changed`](bevy_ecs::prelude::Changed), but for [`Asset`]s,
//! and triggers whenever the handle or the underlying asset changes.
```

crates/bevy_asset/src/asset_changed.rs
```diff
@@ -0,0 +1,465 @@
...
+impl<'w, A: AsAssetId> Clone for AssetChangedFetch<'w, A> {
+    fn clone(&self) -> Self {
+        Self {
+            inner: self.inner,
+            check: self.check,
+        }
+    }
+}
+
+/// State for [`AssetChanged`].
```
alice-i-cecile, line 142:
```suggestion
/// [`WorldQuery`] state for [`AssetChanged`].
```

crates/bevy_asset/src/asset_changed.rs
```diff
@@ -0,0 +1,465 @@
...
+/// runtime of the query is proportional to how many entities with a `A`
+/// it matches.
+///
+/// If no `A` asset updated since the last time the system ran, then no lookups occur.
+///
+/// [`AssetEvents`]: crate::AssetEvents
+/// [`Assets<Mesh>::get_mut`]: crate::Assets::get_mut
+pub struct AssetChanged<A: AsAssetId>(PhantomData<A>);
+
+/// Fetch for [`AssetChanged`].
```
alice-i-cecile, line 126:
```suggestion
/// [`WorldQuery`] fetch for [`AssetChanged`].
```

crates/bevy_ecs/src/query/state.rs
```diff
@@ -583,6 +583,16 @@ impl<D: QueryData, F: QueryFilter> QueryState<D, F> {
             {
                 access.add_component_write(archetype_component_id);
             }
+            if self.component_access.access.has_resource_read(component_id) {
```
alice-i-cecile, line 586: Oh right, this is an existing soundness bug. Re-adding the tag.

crates/bevy_asset/src/asset_changed.rs
tychedelia: Sorry, I also misunderstood you. Multiple assets of the same type should be much easier to handle. I'd been hung up on if, for example we decided to merge mesh and material as has been discussed before.

crates/bevy_asset/src/asset_changed.rs
chescock: Not quite.  That adds access to components on entities in archetypes matched by the query.  But you want to add access to resources from the world.  Resources on the world have their own `ArchetypeComponentId`s that don't appear in any archetypes.  

If we just want to support this use case of a finite set of read access, I think it's enough to add code to `Query::new_with_access` that does
```rust
for component_id in state.component_access.resource_reads() {
    access.add_resource_read(world.initialize_resource_internal(component_id).id());
}
```

More complete support would need to check for writes and for `reads_all` and `writes_all`, and look a bit like the code in `FilteredResourcesMutParamBuilder` https://github.com/bevyengine/bevy/blob/6178ce93e8537f823685e8c1b617e22ba93696e7/crates/bevy_ecs/src/system/builder.rs#L644

---

The background here is that adding access in a `QueryData` has only ever added access to entities matched by the query.  There has never been a way to declare that you want access to resources from the world.  

But the code you wrote *looks* like it should give access to resources!  And that access is useful!  So I think we should add the plumbing to make it work.  

It might be cleaner to do that in a separate PR, though, and put this one back to what you had originally.  This one would just be unsound until the other one was merged.  If you'd like, I can try to put one together on Monday.

tychedelia: Thanks @chescock, that would be super helpful. I'm mostly trying to land this to unblock some rendering features that depend on asset change detection and some of the ECS internals are a bit above my paygrade. Appreciate your education here, as it's better building my understanding.

alice-i-cecile: > But the code you wrote looks like it should give access to resources! And that access is useful! So I think we should add the plumbing to make it work.

Yes please! This is really useful for unlocking indexes as a first class ECS feature.

chescock: Okay!  #16843

While writing tests there, I also noticed we might need to add
```rust
assert!(!access.access().has_resource_write(component_id), "some useful message");
```
just before the `add_resource_read` call, unless we decide to just ban queries from making mutable resource access.

crates/bevy_asset/src/asset_changed.rs
```diff
@@ -0,0 +1,468 @@
...
+    }
+
+    unsafe fn init_fetch<'w>(
+        world: UnsafeWorldCell<'w>,
+        state: &Self::State,
+        last_run: Tick,
+        this_run: Tick,
+    ) -> Self::Fetch<'w> {
+        // SAFETY: `AssetChanges` is private and only accessed mutably in the `AssetEvents` schedule
+        let Some(changes) = (unsafe { world.get_resource::<AssetChanges<A::Asset>>() }) else {
```
chescock, line 172: Since you already have the `resource_id` in `state`, it would be slightly more efficient to call `get_resource_by_id`.  I think that would look something like: 

```suggestion
        // SAFETY:
        // - `AssetChanges` is private and only accessed mutably in the `AssetEvents` schedule
        // - `resource_id` was obtained from the type ID of `AssetChanges<A::Asset>`.
        let Some(changes) = (unsafe {
            world
                .get_resource_by_id(state.resource_id)
                .map(|ptr| ptr.deref::<AssetChanges<A::Asset>>())
        }) else {
```

crates/bevy_asset/src/asset_changed.rs
tychedelia: I've removed the changes I added and merged your changes from main. I've also added the assertion for write access in the filter.

crates/bevy_ecs/src/query/state.rs
tychedelia: These changes have been removed in favor of @chescock's.

alice-i-cecile (approved):

crates/bevy_asset/src/asset_changed.rs
```diff
@@ -0,0 +1,479 @@
...
+use disqualified::ShortName;
+
+/// A resource that stores the last tick an asset was changed. This is used by
+/// the [`AssetChanged`] filter to determine if an asset has changed since the last time
+/// a query ran.
+///
+/// This resource is automatically managed by the [`AssetEvents`](crate::AssetEvents) schedule and
+/// should not be exposed to the user in order to maintain safety guarantees. Any additional uses of
+/// this resource should be carefully audited to ensure that they do not introduce any safety
+/// issues.
```
pcwalton, line 28: When you mean "should not be exposed", what do you mean? Changing the fields?

I ask because usually safety invariants are scoped to be module-private, not crate-private.

pcwalton (approved):
Seems fine, with the caveat that as usual I'm not an expert on the ECS guts.

crates/bevy_asset/src/asset_changed.rs
tychedelia: This mostly is meant to mean don't make this pub because some asks. It's crate scoped mostly due to code organization, having the filter live in its own file is I think preferable to including all of this code in the root lib file where the asset event system lives simply for the sake of keeping it module private.

PR #16812, opened 2024-12-14
Remove the type parameter from `check_visibility`, and only invoke it once.

pcwalton:
Currently, `check_visibility` is parameterized over a query filter that specifies the type of potentially-visible object. This has the unfortunate side effect that we need a separate system, `mark_view_visibility_as_changed_if_necessary`, to trigger view visibility change detection. That system is quite slow because it must iterate sequentially over all entities in the scene.

This PR moves the query filter from `check_visibility` to a new component, `VisibilityClass`. `VisibilityClass` stores a list of type IDs, each corresponding to one of the query filters we used to use. Because `check_visibility` is no longer specialized to the query filter at the type level, Bevy now only needs to invoke it once, leading to better performance as `check_visibility` can do change detection on the fly rather than delegating it to a separate system.

This commit also has ergonomic improvements, as there's no need for applications that want to add their own custom renderable components to add specializations of the `check_visibility` system to the schedule. Instead, they only need to ensure that the `ViewVisibility` component is properly kept up to date. The recommended way to do this, and the way that's demonstrated in the `custom_phase_item` and `specialized_mesh_pipeline` examples, is to make `ViewVisibility` a required component and to add the type ID to it in a component add hook. This patch does this for `Mesh3d`, `Mesh2d`, `Sprite`, `Light`, and `Node`, which means that most app code doesn't need to change at all.

Note that, although this patch has a large impact on the performance of visibility determination, it doesn't actually improve the end-to-end frame time of `many_cubes`. That's because the render world was already effectively hiding the latency from
`mark_view_visibility_as_changed_if_necessary`. This patch is, however, necessary for *further* improvements to `many_cubes` performance.

`many_cubes` trace before:
Screenshot 2024-12-13 015318

`many_cubes` trace after:
Screenshot 2024-12-13 145735

## Migration Guide

* `check_visibility` no longer takes a `QueryFilter`, and there's no need to add it manually to your app schedule anymore for custom rendering items. Instead, entities with custom renderable components should add the appropriate type IDs to `VisibilityClass`. See `custom_phase_item` for an example.


crates/bevy_render/src/view/visibility/mod.rs
```diff
@@ -126,6 +127,30 @@ impl InheritedVisibility {
...
+/// that renders it: for example, `Mesh3d`, `Mesh2d`, and `Sprite`. The
+/// [`VisibilityClass`] component stores the visibility class or classes that
+/// the entity belongs to. (Generally, an object will belong to only one
+/// visibility class, but in rare cases it may belong to multiple.)
+///
+/// When adding a new renderable component, you'll typically want to write an
+/// add-component hook that adds the type ID of that component to the
+/// [`VisibilityClass`] array. See `custom_phase_item` for an example.
+#[derive(Clone, Component, Default, Reflect, Deref, DerefMut)]
+pub struct VisibilityClass(pub SmallVec<[TypeId; 1]>);
```
ItsDoot, line 152: How much more work would it be to store `ComponentId` instead of `TypeId`? The former would be more introspect-able.

JMS55:
@pcwalton need to do the same for meshlets.

crates/bevy_render/src/view/visibility/mod.rs
pcwalton: @ItsDoot I just tried it, and unfortunately it's not really possible because those IDs get fed into the render world, and component IDs are per-world.

pcwalton:
@JMS55 Good point, added meshlet support.

crates/bevy_render/src/view/visibility/mod.rs
JMS55: Do you mind leaving that as a comment? Would be nice to have for future reference.

pcwalton: Ok, I added a comment. I didn't make it a doc comment because I don't think users need to care about the reasons behind the choice.

alice-i-cecile: The per-world component ids are an interesting argument for an entity-partitioning strategy, rather than true multiple worlds.

crates/bevy_render/src/mesh/components.rs
```diff
@@ -96,3 +116,10 @@ impl From<&Mesh3d> for AssetId<Mesh> {
         mesh.id()
     }
 }
+
+// The `check_visibility` system needs to mark 3D meshes as visible.
+fn add_mesh3d_visibility_class(mut world: DeferredWorld<'_>, entity: Entity, _: ComponentId) {
+    if let Some(mut visibility_class) = world.get_mut::<VisibilityClass>(entity) {
+        visibility_class.push(TypeId::of::<Mesh3d>());
+    }
+}
```
kristoff3r, line 125: These hooks could be generic one generic function and live in the visibility module for slightly less boilerplate? Unless that doesn't work with the macro.

crates/bevy_render/src/view/visibility/mod.rs
```diff
@@ -344,24 +359,25 @@ impl Plugin for VisibilityPlugin {
...
-            ),
-        );
+        app.register_type::<VisibilityClass>()
+            .configure_sets(
+                PostUpdate,
+                (CalculateBounds, UpdateFrusta, VisibilityPropagate)
+                    .before(CheckVisibility)
+                    .after(TransformSystem::TransformPropagate),
+            )
+            .configure_sets(PostUpdate, CheckVisibility.ambiguous_with(CheckVisibility))
```
kristoff3r, line 369: Just noticed this while reading, what does it mean for a set to be ambiguous with itself? It was like this before as well.

kristoff3r (approved):
I like this approach and code LGTM, I haven't tested it

crates/bevy_render/src/view/visibility/mod.rs
alice-i-cecile: Ambiguities between systems within the set will be ignored. This should probably be removed now!

crates/bevy_render/src/mesh/components.rs
JMS55: I like this suggestion. Should work fine.

pcwalton: Done.

alice-i-cecile (approved):

PR #16813, opened 2024-12-14
Add ability to mute audio sinks

mgi388:
# Objective

- Allow users to mute audio.

```rust
fn mute(
    keyboard_input: Res<ButtonInput<KeyCode>>,
    mut sink: Single<&mut AudioSink, With<MyMusic>>,
) {
    if keyboard_input.just_pressed(KeyCode::KeyM) {
        sink.toggle_mute();
    }
}
```

- I want to be able to press, say, `M` and mute all my audio. I want this for dev, but I'm sure it's a useful player setting as well.
- Muting is different to pausing—I don't want to pause my sounds, I want them to keep playing but with no volume. For example if I have background music playing which is made up of 5 tracks, I want to be able to temporarily mute my background music, and if I unmute at, say, track 4, I want to play track 4 rather than have had everything paused and still be on the first track.
- I want to be able to continue to control the volume of my audio even when muted. Like in the example, if I have muted my audio but I use the volume up/down controls, I want Bevy to remember those volume changes so that when I unmute, the volume corresponds to that.

## Solution

- Add methods to audio to allow muting, unmuting and toggling muting.
- To preserve the user's intended volume, each sink needs to keep track of a "managed volume".
- I checked `rodio` and I don't see any built in support for doing this, so I added it to `bevy_audio`.
- I'm interested to hear if this is a good idea or a bad idea. To me, this API looks nice and looks usable, but I'm aware it involves some changes to the existing API and now also requires mutable access in some places compared to before.
- I'm also aware of work on *Better Audio*, but I'm hoping that if this change isn't too wild it might be a useful addition considering we don't really know when we'll eventually get better audio.

## Testing

- Update and run the example:  `cargo run --example audio_control`
- Run the example:  `cargo run --example soundtrack`
- Update and run the example:  `cargo run --example spatial_audio_3d`
- Add unit tests.

---

## Showcase

See 2 changed examples that show how you can mute an audio sink and a spatial audio sink.

## Migration Guide

- The `AudioSinkPlayback` trait now has 4 new methods to allow you to mute audio sinks: `is_muted`, `mute`, `unmute` and `toggle_mute`. You can use these methods on `bevy_audio`'s `AudioSink` and `SpatialAudioSink` components to manage the sink's mute state.
- `AudioSinkPlayback`'s `set_volume` method now takes a mutable reference instead of an immutable one. Update your code which calls `set_volume` on `AudioSink` and `SpatialAudioSink` components to take a mutable reference. E.g.:

Before:

```rust
fn increase_volume(sink: Single<&AudioSink>) {
    sink.set_volume(sink.volume() + 0.1);
}
```

After:

```rust
fn increase_volume(mut sink: Single<&mut AudioSink>) {
    let current_volume = sink.volume();
    sink.set_volume(current_volume + 0.1);
}
```

- The `PlaybackSettings` component now has a `muted` field which you can use to spawn your audio in a muted state. `PlaybackSettings` also now has a helper method `muted` which you can use when building the component. E.g.: 

```rust
commands.spawn((
    // ...
    AudioPlayer::new(asset_server.load("sounds/Windless Slopes.ogg")),
    PlaybackSettings::LOOP.with_spatial(true).muted(),
));
```

### DIFF
--- a/crates/bevy_audio/src/audio_output.rs
+++ b/crates/bevy_audio/src/audio_output.rs
@@ -160,7 +160,7 @@ pub(crate) fn play_queued_audio_system<Source: Asset + Decodable>(
             match settings.mode {
                 PlaybackMode::Loop => sink.append(audio_source.decoder().repeat_infinite()),
                 PlaybackMode::Once | PlaybackMode::Despawn | PlaybackMode::Remove => {
-                    sink.append(audio_source.decoder())
+                    sink.append(audio_source.decoder());
                 }
             };
 
@@ -200,7 +200,7 @@ pub(crate) fn play_queued_audio_system<Source: Asset + Decodable>(
             match settings.mode {
                 PlaybackMode::Loop => sink.append(audio_source.decoder().repeat_infinite()),
                 PlaybackMode::Once | PlaybackMode::Despawn | PlaybackMode::Remove => {
-                    sink.append(audio_source.decoder())
+                    sink.append(audio_source.decoder());
                 }
             };
 
--- a/crates/bevy_audio/src/audio_output.rs
+++ b/crates/bevy_audio/src/audio_output.rs
@@ -159,9 +159,9 @@ pub(crate) fn play_queued_audio_system<Source: Asset + Decodable>(
 
             match settings.mode {
                 PlaybackMode::Loop => sink.append(audio_source.decoder().repeat_infinite()),
-                PlaybackMode::Once => sink.append(audio_source.decoder()),
-                PlaybackMode::Despawn => sink.append(audio_source.decoder()),
-                PlaybackMode::Remove => sink.append(audio_source.decoder()),
+                PlaybackMode::Once | PlaybackMode::Despawn | PlaybackMode::Remove => {
+                    sink.append(audio_source.decoder())
+                }
             };
 
             let mut sink = SpatialAudioSink::new(sink);
@@ -178,8 +178,7 @@ pub(crate) fn play_queued_audio_system<Source: Asset + Decodable>(
             }
 
             match settings.mode {
-                PlaybackMode::Loop => commands.entity(entity).insert(sink),
-                PlaybackMode::Once => commands.entity(entity).insert(sink),
+                PlaybackMode::Loop | PlaybackMode::Once => commands.entity(entity).insert(sink),
                 PlaybackMode::Despawn => commands
                     .entity(entity)
                     // PERF: insert as bundle to reduce archetype moves
@@ -200,9 +199,9 @@ pub(crate) fn play_queued_audio_system<Source: Asset + Decodable>(
 
             match settings.mode {
                 PlaybackMode::Loop => sink.append(audio_source.decoder().repeat_infinite()),
-                PlaybackMode::Once => sink.append(audio_source.decoder()),
-                PlaybackMode::Despawn => sink.append(audio_source.decoder()),
-                PlaybackMode::Remove => sink.append(audio_source.decoder()),
+                PlaybackMode::Once | PlaybackMode::Despawn | PlaybackMode::Remove => {
+                    sink.append(audio_source.decoder())
+                }
             };
 
             let mut sink = AudioSink::new(sink);
@@ -219,8 +218,7 @@ pub(crate) fn play_queued_audio_system<Source: Asset + Decodable>(
             }
 
             match settings.mode {
-                PlaybackMode::Loop => commands.entity(entity).insert(sink),
-                PlaybackMode::Once => commands.entity(entity).insert(sink),
+                PlaybackMode::Loop | PlaybackMode::Once => commands.entity(entity).insert(sink),
                 PlaybackMode::Despawn => commands
                     .entity(entity)
                     // PERF: insert as bundle to reduce archetype moves
--- a/crates/bevy_audio/src/audio_output.rs
+++ b/crates/bevy_audio/src/audio_output.rs
@@ -167,25 +167,25 @@ pub(crate) fn play_queued_audio_system<Source: Asset + Decodable>(
             match settings.mode {
                 PlaybackMode::Loop => {
                     sink.append(audio_source.decoder().repeat_infinite());
-                    commands.entity(entity).insert(SpatialAudioSink { sink });
+                    commands.entity(entity).insert(SpatialAudioSink::new(sink));
                 }
                 PlaybackMode::Once => {
                     sink.append(audio_source.decoder());
-                    commands.entity(entity).insert(SpatialAudioSink { sink });
+                    commands.entity(entity).insert(SpatialAudioSink::new(sink));
                 }
                 PlaybackMode::Despawn => {
                     sink.append(audio_source.decoder());
                     commands
                         .entity(entity)
                         // PERF: insert as bundle to reduce archetype moves
-                        .insert((SpatialAudioSink { sink }, PlaybackDespawnMarker));
+                        .insert((SpatialAudioSink::new(sink), PlaybackDespawnMarker));
                 }
                 PlaybackMode::Remove => {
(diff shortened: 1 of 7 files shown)


alice-i-cecile:
Changes like this definitely shouldn't be blocked on the Better Audio work :)

crates/bevy_audio/src/sinks.rs
```diff
@@ -86,15 +113,46 @@ pub trait AudioSinkPlayback {
...
 }
 
 impl AudioSinkPlayback for AudioSink {
     fn volume(&self) -> f32 {
-        self.sink.volume()
+        if let Some(volume) = self.managed_volume {
+            volume
+        } else {
+            self.sink.volume()
+        }
```
SolarLiner, line 147:
```suggestion
        self.managed_volume.unwrap_or_else(|| self.sink.volume())
```

crates/bevy_audio/src/sinks.rs
```diff
@@ -124,6 +182,22 @@ impl AudioSinkPlayback for AudioSink {
...
+    fn mute(&mut self) {
+        self.managed_volume = Some(self.volume());
+        self.sink.set_volume(0.0);
+    }
+
+    fn unmute(&mut self) {
+        if let Some(volume) = self.managed_volume {
+            self.sink.set_volume(volume);
+            self.managed_volume = None;
+        }
```
SolarLiner, line 199:
```suggestion
        if let Some(volume) = self.managed_volume.take() {
            self.sink.set_volume(volume);
        }
```

crates/bevy_audio/src/sinks.rs
```diff
@@ -139,15 +213,42 @@ impl AudioSinkPlayback for AudioSink {
...
+        Self {
+            sink,
+            managed_volume: None,
+        }
+    }
 }
 
 impl AudioSinkPlayback for SpatialAudioSink {
     fn volume(&self) -> f32 {
         self.sink.volume()
```
SolarLiner, line 243: Did you forget to change it here?

crates/bevy_audio/src/sinks.rs
```diff
@@ -177,6 +278,22 @@ impl AudioSinkPlayback for SpatialAudioSink {
...
+    fn mute(&mut self) {
+        self.managed_volume = Some(self.volume());
+        self.sink.set_volume(0.0);
+    }
+
+    fn unmute(&mut self) {
+        if let Some(volume) = self.managed_volume {
+            self.sink.set_volume(volume);
+            self.managed_volume = None;
+        }
```
SolarLiner, line 295:
```suggestion
        if let Some(volume) = self.managed_volume.take() {
            self.sink.set_volume(volume);
        }
```

SolarLiner (changes requested):
I really like this! This packs quite a lot of punch for such a small change, and I like the design!

I've left a few nitpicks, but otherwise LGTM!

crates/bevy_audio/src/sinks.rs
mgi388: Good pickup thanks. This is now covered by the added unit test assuming CI passes it (creating a rodio `SpatialSink` in a test is different to creating a `Sink`, so I'm keen to see if CI can run it).

mgi388: Yeah CI doesn't like the spatial test, so removed it in https://github.com/bevyengine/bevy/pull/16813/commits/973cf24aee87386104b2662bbe9a9fe4651dac82

examples/audio/spatial_audio_3d.rs
```diff
@@ -128,3 +129,11 @@ fn update_listener(
         listeners.translation.z -= speed * time.delta_secs();
     }
 }
+
+fn mute(keyboard_input: Res<ButtonInput<KeyCode>>, mut sinks: Query<&mut SpatialAudioSink>) {
+    if keyboard_input.just_pressed(KeyCode::KeyM) {
```
rparrett, line 134: On-screen instructions for this example should be updated to reflect this

rparrett:
What's the reasoning for `managed_volume` being an `Option`? At a glance, shuffling the volume setting back and forth between our sink wrapper and the underlying sink seems like an unnecessary bit of complexity.

SolarLiner:
Since there's no native support for sink muting in rodio, this is like the least complex alternative to me. Since if we set the sink output to zero, we lose the original volume, we then have to store it somewhere. The `Option<f32>` for the volumes makes intuitive sense, at least to me.

rparrett:
I get the need to store it separately, just unclear on `Option<f32>` vs `f32`.

rparrett:
It seems like `PlaybackSettings` should get a corresponding `muted` field.

SolarLiner:
> I get the need to store it separately, just unclear on `Option<f32>` vs `f32`.

Ah I didn't get your original question then. `Option` makes sense because it also makes explicit that the managed volume is sometimes not available, or temporary. It makes less of a sense to me to make it non-optional and have to mutate it everywhere we also mutate the sink's volume, which also increases complexity, and more bug-prone.

rparrett:
> Ah I didn't get your original question then. Option makes sense because it also makes explicit that the managed volume is sometimes not available, or temporary.

I feel like "managed volume being `Some`" is sort of a clunky way to represent that the sink is in a muted state vs something like

```rust
volume: f32
muted: bool
```

IMO, "the volume the user wants when the sink is unmuted" is conceptually not a temporary thing.

But that detail is hidden from users, so not something I feel the need to fuss over. 👍

SolarLiner (approved):

rparrett (commented):
It seems like `PlaybackSettings` should get a corresponding `muted` field so that `AudioPlayer` can be spawned in a muted state.

Right now, if an app author wanted to, for example, persist and restore a user's "muted" preference, implementing that would be tricky for them.

mgi388:
@rparrett thanks for the review and suggestion to update `PlaybackSettings`.

Re:

> IMO, "the volume the user wants when the sink is unmuted" is conceptually not a temporary thing.

The volume the user wants when the sink is unmuted is actually whatever is returned from `self.sink.volume()`, so the concept is not actually temporary, it's just returned from a different place (the underlying sink).

I feel the `Option` makes more sense. If we're muted, we _need_ to manage the volume for the user, so we have `Some(volume)`. If we're unmuted, we don't _want_ to manage any extra internal state and want to just defer to `self.sink.volume()`, so we have `None`.

Supposing we used two fields for this per your suggestion, then if we're unmuted, we're unnecessarily tracking the volume in our own `volume` field _and_ in the sink's volume, and that feels a little brittle. Plus it just doesn't seem necessary since we don't need that state around (as above, we just defer to whatever `self.sink.volume()` returns anyway).

rparrett (approved):

SolarLiner (approved):

crates/bevy_audio/src/audio_output.rs
```diff
@@ -10,7 +10,7 @@ use bevy_transform::prelude::GlobalTransform;
 use bevy_utils::tracing::warn;
 use rodio::{OutputStream, OutputStreamHandle, Sink, Source, SpatialSink};
```
jaypotts072-cyber, line 12: @ jay mark Potts /474399695

crates/bevy_audio/src/audio_output.rs
```diff
@@ -165,28 +178,15 @@ pub(crate) fn play_queued_audio_system<Source: Asset + Decodable>(
             }
 
             match settings.mode {
-                PlaybackMode::Loop => {
-                    sink.append(audio_source.decoder().repeat_infinite());
-                    commands.entity(entity).insert(SpatialAudioSink { sink });
-                }
```
jaypotts072-cyber, line 171: Greg humphries

jaypotts072-cyber (changes requested):

PR #16814, opened 2024-12-14
Support `SystemInput` tuples up to 8 elements

ItsDoot:
# Objective

- Writing an API, and I want to allow users to pass in extra data alongside the API provided input, and tuples are the most natural extension in this case.
- Bring `SystemInput` up to par with `SystemParam` for tuple support.

## Solution

- Added impls for tuples up to 8 elements. If you need a 9-arity tuple or more, write your own `SystemInput` type (it's incredibly simple to do).

## Testing

- Added a test demonstrating this.

---

## Showcase

Tuples of arbitrary`SystemInput`s are now supported:
```rust
fn by_value((In(a), In(b)): (In<usize>, In<usize>)) -> usize {
    a + b
}
fn by_mut((InMut(a), In(b)): (InMut<usize>, In<usize>)) {
    *a += b;
}

let mut world = World::new();
let mut by_value = IntoSystem::into_system(by_value);
let mut by_mut = IntoSystem::into_system(by_mut);

by_value.initialize(&mut world);
by_mut.initialize(&mut world);

assert_eq!(by_value.run((12, 24), &mut world), 36);

let mut a = 10;
let b = 5;
by_mut.run((&mut a, b), &mut world);
assert_eq!(*a, 15);
```

### DIFF
--- a/crates/bevy_ecs/src/system/input.rs
+++ b/crates/bevy_ecs/src/system/input.rs
@@ -265,8 +265,8 @@ mod tests {
         fn by_ref((InRef(a), InRef(b)): (InRef<usize>, InRef<usize>)) -> usize {
             *a + *b
         }
-        fn by_mut((InMut(a), InMut(b)): (InMut<usize>, InMut<usize>)) {
-            *a += *b;
+        fn by_mut((InMut(a), In(b)): (InMut<usize>, In<usize>)) {
+            *a += b;
         }
 
         let mut world = World::new();
@@ -279,11 +279,11 @@ mod tests {
         by_mut.initialize(&mut world);
 
         let mut a = 12;
-        let mut b = 24;
+        let b = 24;
 
         assert_eq!(by_value.run((a, b), &mut world), 36);
         assert_eq!(by_ref.run((&a, &b), &mut world), 36);
-        by_mut.run((&mut a, &mut b), &mut world);
+        by_mut.run((&mut a, b), &mut world);
         assert_eq!(a, 36);
     }
 }
--- a/crates/bevy_ecs/src/system/input.rs
+++ b/crates/bevy_ecs/src/system/input.rs
@@ -17,6 +17,25 @@ use crate::{bundle::Bundle, prelude::Trigger, system::System};
 ///
 /// For advanced usecases, you can implement this trait for your own types.
 ///
+/// # Examples
+///
+/// ## Tuples of [`SystemInput`]s
+///
+/// ```
+/// use bevy_ecs::prelude::*;
+///
+/// fn add((InMut(a), In(b)): (InMut<usize>, In<usize>)) {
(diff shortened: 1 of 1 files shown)


alice-i-cecile:
@spectria-limina you were poking at this; can I get your review here?

spectria-limina:
LGTM

crates/bevy_ecs/src/system/input.rs
```diff
@@ -11,6 +13,9 @@ use crate::{bundle::Bundle, prelude::Trigger, system::System};
 /// - [`InMut<T>`]: For mutable references to values
 /// - [`Trigger<E, B>`]: For [`ObserverSystem`]s
 /// - [`StaticSystemInput<I>`]: For arbitrary [`SystemInput`]s in generic contexts
+/// - Tuples of [`SystemInput`]s up to 8 elements
```
alice-i-cecile, line 16: This should also be documented on `In`.

alice-i-cecile (commented):
Surprisingly simple and well-tested. I think we can probably do a bit better with docs.

crates/bevy_ecs/src/system/input.rs
```diff
@@ -227,3 +224,66 @@ impl<'a, I: SystemInput> SystemInput for StaticSystemInput<'a, I> {
...
+            }
+        }
+    };
+}
+
+all_tuples!(
+    #[doc(fake_variadic)]
+    impl_system_input_tuple,
+    0,
+    8,
```
bushrat011899, line 249: Not a problem with this PR specifically, but I feel like Bevy should probably just have a constant somewhere for `LONGEST_REASONABLE_TUPLE`, instead of each site picking a somewhat random value (I think I've seen 12, 15, and 16, now also 8).

bushrat011899 (approved):
Minor gripe around max tuple size documentation that isn't even specific to this PR. Otherwise, this just looks like a nice change!

crates/bevy_ecs/src/system/input.rs
ItsDoot: `In` used to be the *only* way to pass input into systems, but given it's now *just* one of the multiple `SystemInput` types, I don't think `In` needs to be treated any differently from the others. I have added an extra note on all of the input types to check the trait for more information.

crates/bevy_ecs/src/system/input.rs
ItsDoot: That's fair, I just think its way less reasonable to have large-arity input tuples when even using system input at all is somewhat uncommon. So I'd prefer if we could (potentially) save some compilation time by not implementing for arities that are highly unlikely to be used in practice.

BenjaminBrienen (approved):

alice-i-cecile (approved):

PR #16823, opened 2024-12-14
Made UIRect initialisation functions `const`

Kees-van-Beilen:
# Objective
Destructuring in const code blocks isn't allowed, thus using UIRect in const code can be a hassle as it initialisation function aren't const. This Pr makes them const.

## Solution

Removed all destructuring in the UIRect implementation

## Testing

- I've ran a few ui examples to check if i didn't make a mistake,

---

### DIFF
--- a/crates/bevy_ui/src/geometry.rs
+++ b/crates/bevy_ui/src/geometry.rs
@@ -390,11 +390,11 @@ impl UiRect {
     /// assert_eq!(ui_rect.top, Val::ZERO);
     /// assert_eq!(ui_rect.bottom, Val::ZERO);
     /// ```
-    pub fn horizontal(value: Val) -> Self {
-        UiRect {
+    pub const fn horizontal(value: Val) -> Self {
+        Self {
             left: value,
             right: value,
-            ..Default::default()
+            ..Self::DEFAULT
         }
     }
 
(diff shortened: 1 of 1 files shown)


crates/bevy_ui/src/geometry.rs
```diff
@@ -459,11 +457,10 @@ impl UiRect {
...
-    pub fn left(value: Val) -> Self {
-        UiRect {
-            left: value,
-            ..Default::default()
-        }
+    pub const fn left(value: Val) -> Self {
+        let mut rect = UiRect::DEFAULT;
+        rect.left = value;
+        rect
     }
```
ickshonpe, line 465: I'd prefer that we used struct update syntax for the constructor functions, like:

```suggestion
    pub const fn left(left: Val) -> Self {
        Self {
            left,
            ..Self::DEFAULT
        }
    }

```

is simpler and clearer.

ickshonpe (commented):
Looks good, all the fields match up correctly and the examples still work. 
Strongly prefer struct update syntax for the constructor functions though.

crates/bevy_ui/src/geometry.rs
ickshonpe: I thought you can, can't you? `Self::DEFAULT` is a const value, it's not `Self::default()`

ickshonpe (approved):
Everything looks good now.

alice-i-cecile (approved):

PR #16825, opened 2024-12-15
Fix meshlet shaders for bindless mode.

pcwalton:
We have to extract the material ID from the mesh and stuff it in the vertex during visibility buffer resolution.


crates/bevy_pbr/src/meshlet/visibility_buffer_resolve.wgsl
```diff
@@ -152,6 +153,8 @@ fn resolve_vertex_output(frag_coord: vec4<f32>) -> VertexOutput {
 
     let world_tangent = calculate_world_tangent(world_normal, ddx_world_position, ddy_world_position, ddx_uv, ddy_uv);
 
+    let material_bind_group_slot = instance_uniform.material_bind_group_slot;
```
JMS55, line 156: Nit: Move this inline to the output, under instance_id ^ meshlet_id.

crates/bevy_pbr/src/render/parallax_mapping.wgsl
```diff
@@ -5,8 +5,7 @@
     mesh_bindings::mesh
 }
 
-fn sample_depth_map(uv: vec2<f32>, instance_index: u32) -> f32 {
-    let slot = mesh[instance_index].material_bind_group_slot;
+fn sample_depth_map(uv: vec2<f32>, material_bind_group_slot: u32) -> f32 {
```
JMS55, line 8: Why did this change? Was it missed in an earlier PR?

crates/bevy_pbr/src/render/mesh_functions.wgsl
```diff
@@ -12,6 +12,7 @@
 }
 #import bevy_render::maths::{affine3_to_square, mat2x4_f32_to_mat3x3_unpack}
 
+#ifndef MESHLET_MESH_MATERIAL_PASS
```
JMS55, line 15: Why were the `#ifndef MESHLET_MESH_MATERIAL_PASS` added here? Did they cause compilation issues?

JMS55 (changes requested):

crates/bevy_pbr/src/render/parallax_mapping.wgsl
pcwalton: Not quite sure what you mean. It changed because we don't use that `mesh` array in visbuffer resolution. If you're asking whether I didn't fix meshlets in an earlier PR, yes, I broke them (that's the whole point of this PR).

crates/bevy_pbr/src/render/mesh_functions.wgsl
pcwalton: The `mesh` array that's used in those functions isn't present during visbuffer resolution.

crates/bevy_pbr/src/render/parallax_mapping.wgsl
JMS55: Ah nvm I see that it used to get the slot in the function itself, and all you did was move it outside the function so that the slot could be obtained differently for meshlet meshes.

crates/bevy_pbr/src/render/mesh_functions.wgsl
JMS55: Right, but I don't think that causes any actual issues, unless you try and use them? It was never an issue in 0.15. Unless a recent PR changed that. Anyways I'm fine with the ifdef, I was just wondering why it was added.

crates/bevy_pbr/src/render/parallax_mapping.wgsl
pcwalton: Right.

JMS55 (approved):

JMS55:
I can't test meshlets atm (don't have access to my desktop), but approving on the assumption that it was tested and this works. Code looks fine.

crates/bevy_pbr/src/render/mesh_functions.wgsl
pcwalton: It caused the shaders to fail to compile.

JMS55:
I'm also curious to know, is all 5 StandardMaterial bunnies shaded in one draw call with bindless? Because if there's one bind group and pipeline shared across all 5 bunnies, they should be able to be shaded in one pass. So there should be two draws total during the meshlet mesh material shading pass: one for the debug material, and one for standard material.

pcwalton:
@JMS55 Bindless isn't enabled for meshlets. I don't know enough about the meshlet code to implement it; you'll need to implement it if you want it. It doesn't break, though.

IceSentry (approved):

JMS55:
Can confirm it works.
image

Also looking at the renderdoc trace, now we do bind PBR StandardMaterial BG, 5 draws for each of the PBR instances, bind MeshletDebugMaterial, 1 draw for the 5 debug instances. Before bindless this used to be a separate bind + draw for each PBR instance, now it's 1 bind + separate draws.

Ideally ofc this would be 1 draw for all 5 PBR instances since they share the same pipeline + BG, but I can fix that sometime in the future. Probably needs to wait for retained bins, and when we consolidate the binning code across mesh and meshlet mesh.

image

PR #16826, opened 2024-12-15
Extend cloning functionality and add convenience methods to `EntityWorldMut` and `EntityCommands`

### LINKED ISSUE #15350: Moving components from one entity to another

For my reversible systems crate I want to support reversible entity commands that do structural changes.

For example, to create a reversible variant of `EntityCommands::insert`, that takes a `Bundle` as an argument, I first have to move the components that would be overwritten out of the entity so I can make this process undoable.

`BundleInfo` has the methods I need to get the relevant ids including required components. However, there is no further API from here on.

What would work is to move components from one entity to another by id(s). A typed variant should be added there too because it does not exist yet. This is an elegant solution because the data structure of storing untyped components is already there: The ECS storages. This might be useful for other use cases as well.

```rs
// in EntityWorldMut and EntityCommands impl

pub fn move<T: Bundle>(&mut self, to: Entity) -> &mut Self;

pub fn move_by_id(&mut self, to: Entity, component_id: ComponentId) -> &mut Self;

pub fn move_by_ids(
    &mut self, 
    to: Entity,
    component_ids: impl IntoIterator<Item = ComponentId>
) -> &mut Self;
```
Each variant should also come with a `_with_requires` variant like https://github.com/bevyengine/bevy/pull/15026 that has the difference to not construct required components at the target entity but to move them from the source entity as well.

Hooks should be triggered by these operations.

Panics should happen if either the source or target entity is missing or at least one `ComponentId` is invalid.

### Open questions

- Should this come with variants like `try_insert` that fail silently?

An alternative is to move the components entirely out of the ECS storage as some kind of blob. However, no such type exists yet as far I can tell. It would require extra work to design this blob, also because it needs to support generating `OwningPtr` for reinsertion.

I see no alternative in my untyped context. I could mirror the Bundle trait to create an associated function but that would not work with third-party Bundle implementors that do not implement my trait.

Until a feature like this comes out I consider to require only tuple bundles for my crate's API with further restrictions on the upcoming required components feature. This probably makes the API very panic-happy. 🫤 

I see reflection not helping me here as not every component can be reflected.

JaySpruce:
## Objective

Thanks to @eugineerd's work on entity cloning (#16132), we now have a robust way to copy components between entities. We can extend this to implement some useful functionality that would have been more complicated before.

Closes #15350.

## Solution

`EntityCloneBuilder` now automatically includes required components alongside any component added/removed from the component filter.

Added the following methods to `EntityCloneBuilder`:
- `move_components`
- `without_required_components`

Added the following methods to `EntityWorldMut` and `EntityCommands`:
- `clone_with`
- `clone_components`
- `move_components`

Also added `clone_and_spawn` and `clone_and_spawn_with` to `EntityWorldMut` (`EntityCommands` already had them).

## Showcase

```
assert_eq!(world.entity(entity_a).get::<B>(), Some(&B));
assert_eq!(world.entity(entity_b).get::<B>(), None);
world.entity_mut(entity_a).clone_components::<B>(entity_b);
assert_eq!(world.entity(entity_a).get::<B>(), Some(&B));
assert_eq!(world.entity(entity_b).get::<B>(), Some(&B));

assert_eq!(world.entity(entity_a).get::<C>(), Some(&C(5)));
assert_eq!(world.entity(entity_b).get::<C>(), None);
world.entity_mut(entity_a).move_components::<C>(entity_b);
assert_eq!(world.entity(entity_a).get::<C>(), None);
assert_eq!(world.entity(entity_b).get::<C>(), Some(&C(5)));
```

### DIFF
--- a/crates/bevy_ecs/src/entity/clone_entities.rs
+++ b/crates/bevy_ecs/src/entity/clone_entities.rs
@@ -18,6 +18,7 @@ pub struct EntityCloner {
     filter_allows_components: bool,
     filter: Arc<HashSet<ComponentId>>,
     clone_handlers_overrides: Arc<HashMap<ComponentId, ComponentCloneHandler>>,
+    move_components: bool,
 }
 
 impl EntityCloner {
@@ -35,16 +36,20 @@ impl EntityCloner {
                 .filter(|id| self.is_cloning_allowed(id)),
         );
 
-        for component in components {
+        for component in &components {
             let global_handlers = world.components().get_component_clone_handlers();
-            let handler = match self.clone_handlers_overrides.get(&component) {
-                None => global_handlers.get_handler(component),
+            let handler = match self.clone_handlers_overrides.get(component) {
+                None => global_handlers.get_handler(*component),
                 Some(ComponentCloneHandler::Default) => global_handlers.get_default_handler(),
                 Some(ComponentCloneHandler::Ignore) => component_clone_ignore,
                 Some(ComponentCloneHandler::Custom(handler)) => *handler,
             };
-            self.component_id = Some(component);
+            self.component_id = Some(*component);
             (handler)(&mut world.into(), self);
+        }
+
+        if self.move_components {
+            world.entity_mut(self.source).remove_by_ids(&components);
         }
     }
 
@@ -145,6 +150,8 @@ pub struct EntityCloneBuilder<'w> {
     filter_allows_components: bool,
     filter: HashSet<ComponentId>,
     clone_handlers_overrides: HashMap<ComponentId, ComponentCloneHandler>,
+    attach_required_components: bool,
+    move_components: bool,
 }
 
 impl<'w> EntityCloneBuilder<'w> {
@@ -155,6 +162,8 @@ impl<'w> EntityCloneBuilder<'w> {
             filter_allows_components: false,
             filter: Default::default(),
             clone_handlers_overrides: Default::default(),
+            attach_required_components: false,
+            move_components: false,
         }
     }
 
@@ -165,6 +174,7 @@ impl<'w> EntityCloneBuilder<'w> {
             filter_allows_components,
             filter,
             clone_handlers_overrides,
+            move_components,
             ..
         } = self;
 
@@ -175,10 +185,40 @@ impl<'w> EntityCloneBuilder<'w> {
             filter_allows_components,
             filter: Arc::new(filter),
             clone_handlers_overrides: Arc::new(clone_handlers_overrides),
+            move_components,
         }
         .clone_entity(world);
 
         world.flush_commands();
+    }
+
+    /// Allows for a scoped mode where any changes to the filter that allow/deny components
+    /// will also allow/deny those components' required components, recursively.
+    pub fn with_required_components(
+        &mut self,
+        builder: impl FnOnce(&mut EntityCloneBuilder) + Send + Sync + 'static,
+    ) -> &mut Self {
+        self.attach_required_components = true;
+        builder(self);
+        self.attach_required_components = false;
+        self
+    }
+
+    /// Sets the cloner to remove any components that were cloned,
+    /// effectively moving them from the source entity to the target.
+    ///
+    /// This only applies to components that are allowed through the filter
+    /// at the time [`EntityCloneBuilder::clone_entity`] is called.
+    pub fn enable_move_on_clone(&mut self) -> &mut Self {
+        self.move_components = true;
+        self
+    }
+
+    /// The default setting for the cloner. The source entity will keep
+    /// all components that are cloned.
+    pub fn disable_move_on_clone(&mut self) -> &mut Self {
+        self.move_components = false;
+        self
     }
 
     /// Adds all components of the bundle to the list of components to clone.
@@ -186,18 +226,10 @@ impl<'w> EntityCloneBuilder<'w> {
     /// Note that all components are allowed by default, to clone only explicitly allowed components make sure to call
     /// [`deny_all`](`Self::deny_all`) before calling any of the `allow` methods.
     pub fn allow<T: Bundle>(&mut self) -> &mut Self {
-        if self.filter_allows_components {
-            T::get_component_ids(self.world.components(), &mut |id| {
-                if let Some(id) = id {
-                    self.filter.insert(id);
-                }
-            });
-        } else {
-            T::get_component_ids(self.world.components(), &mut |id| {
-                if let Some(id) = id {
-                    self.filter.remove(&id);
-                }
-            });
+        let bundle = self.world.register_bundle::<T>();
+        let ids = bundle.explicit_components().to_owned();
+        for id in ids {
+            self.filter_allow(id);
         }
         self
     }
@@ -207,12 +239,8 @@ impl<'w> EntityCloneBuilder<'w> {
     /// Note that all components are allowed by default, to clone only explicitly allowed components make sure to call
     /// [`deny_all`](`Self::deny_all`) before calling any of the `allow` methods.
     pub fn allow_by_ids(&mut self, ids: impl IntoIterator<Item = ComponentId>) -> &mut Self {
-        if self.filter_allows_components {
-            self.filter.extend(ids);
-        } else {
-            ids.into_iter().for_each(|id| {
-                self.filter.remove(&id);
-            });
+        for id in ids {
+            self.filter_allow(id);
         }
         self
     }
@@ -222,15 +250,10 @@ impl<'w> EntityCloneBuilder<'w> {
     /// Note that all components are allowed by default, to clone only explicitly allowed components make sure to call
     /// [`deny_all`](`Self::deny_all`) before calling any of the `allow` methods.
     pub fn allow_by_type_ids(&mut self, ids: impl IntoIterator<Item = TypeId>) -> &mut Self {
-        let ids = ids
-            .into_iter()
-            .filter_map(|id| self.world.components().get_id(id));
-        if self.filter_allows_components {
-            self.filter.extend(ids);
-        } else {
-            ids.into_iter().for_each(|id| {
-                self.filter.remove(&id);
-            });
+        for type_id in ids {
+            if let Some(id) = self.world.components().get_id(type_id) {
+                self.filter_allow(id);
+            }
         }
         self
     }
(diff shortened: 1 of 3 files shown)


crates/bevy_ecs/src/world/entity_ref.rs
```diff
@@ -2107,6 +2107,248 @@ impl<'w> EntityWorldMut<'w> {
...
+        let component_ids = bundle_info.contributed_components().to_owned();
+
+        let mut builder = EntityCloneBuilder::new(self.world);
+        builder.deny_all().allow_by_ids(component_ids);
+        builder.clone_entity(self.entity, target);
+
+        self.world.flush();
+        self.update_location();
+        self
+    }
```
eugineerd, line 2220: I'm pretty sure allowing and denying components with their required components can be done as a mode for `EntityCloneBuilder`, which should be more flexible:
```rs
builder.with_required_components(|builder| {
  builder
    .allow::<B>()       // Calling with_required_components can set an internal flag which would
    .allow_by_ids(ids); // make all allow* and deny* methods also include contributed_components
});
```

crates/bevy_ecs/src/world/entity_ref.rs
```diff
@@ -2107,6 +2107,248 @@ impl<'w> EntityWorldMut<'w> {
...
+        let bundles = &mut self.world.bundles;
+        let bundle_id = bundles.register_info::<B>(components, storages);
+
+        // SAFETY: the `BundleInfo` for this `BundleId` is initialized above
+        unsafe { self.remove_bundle(bundle_id) };
+
+        self.world.flush();
+        self.update_location();
+        self
+    }
```
eugineerd, line 2277: Moving components can also be a configuration option for `EntityCloneBuilder` that would call `remove_bundle` as a final step in `clone_entity`. Theoretically, it would also allow us to add an optimization in the future that would allow `EntityCloner` to move components that don't implement `Clone` or `Reflect`.

crates/bevy_ecs/src/world/entity_ref.rs
```diff
@@ -2107,6 +2107,248 @@ impl<'w> EntityWorldMut<'w> {
...
+    /// inserts them into another entity, then removes the components from this entity.
+    ///
+    /// Components can only be cloned if they implement
+    /// [`Clone`] or [`Reflect`](bevy_reflect::Reflect).
+    ///
+    /// # Panics
+    ///
+    /// - If this entity has been despawned while this `EntityWorldMut` is still alive.
+    /// - If the target entity does not exist.
+    pub fn move_components_with_requires<B: Bundle>(&mut self, target: Entity) -> &mut Self {
```
eugineerd, line 2289: Maybe we can avoid combinatorial explosion of methods by just having `move_components` and `move_components_with` similar to `spawn_and_clone`? Although I'm not sure if configuring through `EntityCloneBuilder` is a good idea - it would allow to set cloning hierarchy which would not make much sense in the context of cloning components from one entity to the next. Maybe there should be a `ComponentCloneBuilder` or something similar that would re-expose only the methods that should be allowed?

eugineerd (commented):
I would suggest moving most of the functionality to `EntityCloneBuilder` which should allow to use it in `clone_and_spawn` methods as well. Also I'm a bit worried about the combinatorial explosion of methods - that is why I made `EntityCloneBuilder` in the first place. ( Now It's my time to nitpick :P )

crates/bevy_ecs/src/world/entity_ref.rs
JaySpruce: Something like this?
<details open>
<summary>Code</summary>

```
pub fn with_required_components(
    &mut self,
    f: impl FnOnce(&mut EntityCloneBuilder) + Send + Sync + 'static,
) -> &mut Self {
    self.attach_required_components = true;
    f(self);
    self.attach_required_components = false;
    self
}

pub fn allow<T: Bundle>(&mut self) -> &mut Self {
    if self.filter_allows_components {
        T::get_component_ids(self.world.components(), &mut |id| {
            if let Some(id) = id {
                self.filter.insert(id);
                if self.attach_required_components {
                    if let Some(info) = self.world.components().get_info(id) {
                        for required in info.required_components().iter_ids() {
                            self.filter.insert(required);
                        }
                    }
                }
            }
        });
    } else {
        T::get_component_ids(self.world.components(), &mut |id| {
            if let Some(id) = id {
                self.filter.remove(&id);
                if self.attach_required_components {
                    if let Some(info) = self.world.components().get_info(id) {
                        for required in info.required_components().iter_ids() {
                            self.filter.remove(&required);
                        }
                    }
                }
            }
        });
    }
    self
}
```
</details>

eugineerd: I think this code can actually be deduplicated a bit if we replace `if self.filter_allows_components` with something like this:
```rs
fn apply_filter(
    &mut self,
    id: ComponentId,
) {
    if self.filter_allows_components {
        self.filter.insert(id);
    } else {
        self.filter.remove(&id);
    }
}
```

JaySpruce:
All true, I think I'll cut it down to these:
- `clone_with` (gives plain access to the builder)
- `clone_and_spawn`
- `clone_and_spawn_with`
- `clone_components`
- `move_components`

I'll add the required components stuff to the builder, and tinker with the moving functionality too

crates/bevy_ecs/src/world/entity_ref.rs
JaySpruce: I can't get around the borrow checker to get that to work, but moving the `if`s around is still an improvement

alice-i-cecile:
I really like this direction, and @eugineerd's feedback is excellent. Ping me when you need a review please :)

JaySpruce:
I added the helper functions for the filter, so the allow/deny methods are slimmer now.

Required components work as discussed, the `with_required_components` method opening a sub-builder thing that includes required components in any `allow`/`deny`.

Moving components is just a bool in `EntityCloneBuilder`, which enables or disables removing all components that were cloned at the end of `clone_entity`.

This is the full syntax:
```
world.entity_mut(entity_a).clone_with(entity_b, |builder| {
    builder.deny_all();
    builder.enable_move_on_clone();
    builder.with_required_components(|builder| {
        builder.allow::<B>();
    });
});
```

Also added `EntityWorldMut::remove_by_ids` because I needed it (there was `remove_by_id`, so remove matches insert now, which has `insert_by_id` and `insert_by_ids`).

I think we're good to go for round 2, @alice-i-cecile & @eugineerd.

crates/bevy_ecs/src/entity/clone_entities.rs
```diff
@@ -175,29 +185,51 @@ impl<'w> EntityCloneBuilder<'w> {
...
+            move_components,
         }
         .clone_entity(world);
 
         world.flush_commands();
     }
 
+    /// Allows for a scoped mode where any changes to the filter that allow/deny components
+    /// will also allow/deny those components' required components, recursively.
+    pub fn with_required_components(
```
alice-i-cecile, line 197: Very cool. @cart is almost certainly going to want to want this on by default, possibly without the ability to disable it. The vision is to treat these as "one indivisible object" wherever possible, at least for "required" components :)

I agree with that philosophy, and I think the toggle should basically be "replace required components with their default values". The default value should be "copy all required components that you can".

crates/bevy_ecs/src/entity/clone_entities.rs
```diff
@@ -175,29 +185,51 @@ impl<'w> EntityCloneBuilder<'w> {
...
+    /// effectively moving them from the source entity to the target.
+    ///
+    /// This only applies to components that are allowed through the filter
+    /// at the time [`EntityCloneBuilder::clone_entity`] is called.
+    pub fn enable_move_on_clone(&mut self) -> &mut Self {
+        self.move_components = true;
+        self
+    }
+
+    /// The default setting for the cloner. The source entity will keep
```
alice-i-cecile, line 217: The bit about this being the default should be in the second line, not the first.

crates/bevy_ecs/src/system/commands/mod.rs
```diff
@@ -1808,15 +1835,40 @@ impl<'a> EntityCommands<'a> {
...
     }
+
+    /// Clones the specified components of this entity and inserts them into another entity.
+    ///
+    /// Components can only be cloned if they implement
+    /// [`Clone`] or [`Reflect`](bevy_reflect::Reflect).
+    ///
+    /// # Panics
+    ///
+    /// The command will panic when applied if either of the entities do not exist.
```
alice-i-cecile, line 1855: Gah more panics :( Not your problem: this needs a holistic solution.

crates/bevy_ecs/src/world/entity_ref.rs
```diff
@@ -2107,6 +2132,150 @@ impl<'w> EntityWorldMut<'w> {
...
     }
+
+    /// Clones parts of an entity (components, observers, etc.) onto another entity,
+    /// configured through [`EntityCloneBuilder`].
+    ///
+    /// By default, the other entity will receive all the components of the original that implement
+    /// [`Clone`] or [`Reflect`](bevy_reflect::Reflect).
+    ///
+    /// Configure through [`EntityCloneBuilder`] as follows:
+    /// ```ignore
```
alice-i-cecile, line 2143: Don't ignore this: make it work (possibly hiding lines for importing dependencies / boilerplate). Ignored doc tests are extremely brittle.

crates/bevy_ecs/src/entity/clone_entities.rs
```diff
@@ -175,29 +185,51 @@ impl<'w> EntityCloneBuilder<'w> {
...
+        self.attach_required_components = false;
+        self
+    }
+
+    /// Sets the cloner to remove any components that were cloned,
+    /// effectively moving them from the source entity to the target.
+    ///
+    /// This only applies to components that are allowed through the filter
+    /// at the time [`EntityCloneBuilder::clone_entity`] is called.
+    pub fn enable_move_on_clone(&mut self) -> &mut Self {
```
alice-i-cecile, line 212: Bikeshed: I think these methods might be clearer if they were `move_components` and `keep_components`. The enable/disable terminology was confusing to me at first.

crates/bevy_ecs/src/entity/clone_entities.rs
eugineerd: A bit off-topic for this PR, but I wonder if `DynamicSceneBuilder` should also extract required components in that case?

crates/bevy_ecs/src/entity/clone_entities.rs
eugineerd: I think this should just be one method: `move_components` that takes a bool. That way it would be in line with the other `EntityCloneBuilder` options (like `add_observers`). It was designed to look somewhat similar to `std::fs::OpenOptions`.

crates/bevy_ecs/src/entity/clone_entities.rs
JaySpruce: That's essentially what the toggle is, since components are ultimately inserted with normal `insert` methods. I can switch the default though.

Here's a couple colors for the bikeshed:
- `default_required_components`
- `with_default_required_components`
- `replace_required_components`
- `replace_required_components_with_default`

Also, if `move_on_clone` is on, should required components still be removed if they weren't cloned?

eugineerd (approved):
Looks good to me! I like the addition of various convenience methods and move component functionality is pretty useful.

I have also prepared an optimization based on code from #16717 for `move_components` that would allow us to move components that don't implement `Reflect` or `Clone` after both PRs are merged.

crates/bevy_ecs/src/system/commands/mod.rs
JaySpruce: I do have another PR for that :P (not holistic, just commands). Not sure how it'll interact with work in the fallible systems area though

crates/bevy_ecs/src/entity/clone_entities.rs
alice-i-cecile: Yeah, we should probably be unifying the mechanisms between this and the scenes stuff. Another PR / issue though! As for the bikeshed, go with the long and clear one please: this is quite niche :)

alice-i-cecile (approved):
I'm happy with the state of this now; let's merge and refine in follow-up :)

alice-i-cecile:
Thank you to everyone involved with the authoring or reviewing of this PR! This work is relatively important and needs release notes! Head over to https://github.com/bevyengine/bevy-website/issues/1972 if you'd like to help out.

PR #16828, opened 2024-12-15
Fix rounding bug in camera projection

### LINKED ISSUE #16773: Visual tearing with `ImageFilterMode::Nearest`
## Bevy version

`0.15.0`

## What you did

When `ImageFilterMode::Nearest` is set as the default sampler, tearing occurs to text, sprites, etc. when the window resolution is an odd number (`101x303`, etc.). You can test by spawning a sprite and then resizing the window:

```rust
use bevy::prelude::*;

fn main() -> AppExit {
    App::new()
        .add_plugins(DefaultPlugins.set(ImagePlugin::default_nearest()))
        .add_systems(Startup, setup)
        .run()
}

    commands.spawn(Camera2d);

    commands.spawn(Sprite::from_image(asset_server.load("grid_480x270.png")));
}
```

The notable exception is that it doesn't happen to sprites with dimensions that is the power of two (`512x256`, etc.).

You can test a more elaborate, interactive reproduction here: https://github.com/musjj/nearest_filter_bug. Slowly and gradually resize the window and you should be able to see the artifacts popping in and out.

## What went wrong

Visual tearing occurs.

## Additional information

An example of **no** tearing:

Image

An example of tearing:

Image

A closeup of the tearing (the lines are supposed to be perfectly straight):

Image

musjj:
# Objective

Fixes #16773

## Solution

https://github.com/bevyengine/bevy/pull/15969 introduces a regression that causes image tearing when `ImageFilterMode::Nearest` is used as the sampler. Rounding the origin fixes the issue.

Credits to @sophrosyne97 for bisecting and finding the commit that broke it.

## Testing

You can test it with this repository: https://github.com/musjj/nearest_filter_bug

```bash
git clone https://github.com/musjj/nearest_filter_bug
cd nearest_filter_bug
echo '[patch.crates-io]\nbevy = { git = "https://github.com/musjj/bevy.git", branch = "rounding-bug" }' >>Cargo.toml
sed -i -e 's/0.15.0/0.15.0-dev/' Cargo.toml
cargo update && cargo run
```
Gradually resize the window to various sizes, while paying attention to the sprite and font. No tearing should occur on any window resolution, even on odd numbers.

Tested on Linux with Gnome. Tests on other platforms would be appreciated.

---

## Showcase

Before:

image

After:

image

### DIFF
--- a/crates/bevy_render/src/camera/projection.rs
+++ b/crates/bevy_render/src/camera/projection.rs
@@ -492,8 +492,8 @@ impl CameraProjection for OrthographicProjection {
             ScalingMode::Fixed { width, height } => (width, height),
         };
 
-        let origin_x = projection_width * self.viewport_origin.x;
-        let origin_y = projection_height * self.viewport_origin.y;
+        let origin_x = (projection_width * self.viewport_origin.x).round();
+        let origin_y = (projection_height * self.viewport_origin.y).round();
 
         self.area = Rect::new(
             self.scale * -origin_x,


mockersf (approved):

alice-i-cecile (approved):
Thank you both for spotting this. I was the one to introduce this regression!

PR #9582, opened 2023-08-26
Updated `VolumeLevel` to use Decibel and Amplitude Ratio Units

bushrat011899:
# Objective

- Fixes #9507

## Solution

- Breaking changes in both the `AudioSinkPlayback` `trait` and `VolumeLevel` type.
- Replaced `VolumeLevel` with an `enum` representing either the amplitude ratio or decibel level. Names and types were chosen to be consistent with 3rd party crates like `Kira`, and the internally used types in the rest of `bevy_audio`

---

## Changelog

- Moved `VolumeLevel` into its own file `volume_level.rs` to reduce clutter in `audio.rs`.
- Implemented total equality and ordering for `VolumeLevel`, as both variants `Amplitude` and `Decibels` are isomorphic and internally possess both properties.
- Implemented `amplitude()` and `decibels()` methods to get the desired unit type from `VolumeLevel` directly without a `match` statement.
- Added `debug_assert!` statements to ensure calculated values are within their appropriate bounds.
- Added a unit test for conversion between `Amplitude` and `Decibels`, with a truth table obtained from [Wikipedia](https://web.archive.org/web/20230810185300/https://en.wikipedia.org/wiki/Decibel)
- Updated internal functions to either accept the `VolumeLevel` type, or explicitly request an amplitude which is then wrapped in a `VolumeLevel::Amplitude` where appropriate.
- Removed `VolumeLevel::get` and `VolumeLevel::new` due to ambiguity in the unit to return.

## Migration Guide

- `VolumeLevel::new(volume)` -> `VolumeLevel::Amplitude(volume)`
- `VolumeLevel::get()` -> `VolumeLevel::amplitude()`
- `AudioSinkPlayback::volume()` -> `AudioSinkPlayback::volume().amplitude()`
- `AudioSinkPlayback::set_volume(volume)` -> `AudioSinkPlayback::set_volume(VolumeLevel::Amplitude(volume))`

PR #16838, opened 2024-12-16
Move Volume and GlobalVolume to own file

mgi388:
# Objective

- Prework for reviving #9582.

## Solution

- Move the two types to volume.rs and made it compile.
- Also `#[reflect(Debug)]` on `Volume` while I'm here. 

## Testing

- Ran example locally.
- Rely on CI.

### DIFF
--- a/crates/bevy_audio/src/audio.rs
+++ b/crates/bevy_audio/src/audio.rs
@@ -1,36 +1,10 @@
 #![expect(deprecated)]
 
-use crate::{AudioSource, Decodable};
+use crate::{AudioSource, Decodable, Volume};
 use bevy_asset::{Asset, Handle};
-use bevy_derive::Deref;
 use bevy_ecs::prelude::*;
 use bevy_math::Vec3;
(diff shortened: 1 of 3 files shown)


SolarLiner (approved):
We should put a pin on improving the documentation of the `Volume` newtype, because it's unclear from the doc comments which scale we are using (linear can be inferred from the `debug_assert!`, but it would be nice to make it explicit).

Additionally, we should add a second constructor for setting the volume in decibels.

mgi388:
> We should put a pin on improving the documentation of the `Volume` newtype, because it's unclear from the doc comments which scale we are using (linear can be inferred from the `debug_assert!`, but it would be nice to make it explicit).
> 
> Additionally, we should add a second constructor for setting the volume in decibels.

Absolutely. We can look at these in a follow up PR since this PR is just a simple move around. FWIW, I also noticed this from a user perspective. The constructor just says `Create a new volume level.` which leaves the user wondering :)

mnmaita (approved):

PR #16841, opened 2024-12-16
Toggleable UI layout rounding

### LINKED ISSUE #16731: Bevy UI nodes are always pixel-aligned and there is no way to disable it
## Bevy version

- 0.15.0

```
- cargo 1.83.0 (5ffbef321 2024-10-29)
- MacOS Sonoma 14.1.1 (23B81)
- SystemInfo { os: "MacOS 14.1.1 ", kernel: "23.1.0", cpu: "Apple M1 Max", core_count: "10", memory: "64.0 GiB" }
- AdapterInfo { name: "Apple M1 Max", vendor: 0, device: 0, device_type: IntegratedGpu, driver: "", driver_info: "", backend: Metal }
```

## What you did

Made UI follow 3d entity. Bascially I want to make health bar (and some more info) to stick above/belove each unit.

## What went wrong

It seams that Bevy UI nodes are always aligned to pixels. This is a good thing for static UI to make it sharper but introduces jaggines when UI is smoothly moved. This is especially noticeable when UI is placed near the object.

I think default behaviour is correct but there should be a way to disable it and allow sub-pixel positions/sizes for UI. It probably should be some component added to the root node or a flag in the `TargetCamera`.

## Additional information

Here is example from the game in question:
Image

As you can see model moves smoothly but green borders and red rect (all are nodes within a single root node) are always jumping here and there. If you just look/focus at the model, it becomes much more noticeable.

UI position is calculated by projecting each corner of the hexagon and taking min/max x/y values:
```rust
let corners = Map::get_layout().hex_corners(agent.hex)
    .map(|e| e.extend(0.0).xzy())
    .map(|e| camera.world_to_viewport(camera_transform, e))
    .map(|e| e.unwrap_or(Vec2::INFINITY));

if corners.iter().any(|e| !e.is_finite()) {
    continue;
}
let Some((min_x, max_x)) = corners.iter().minmax_by_key(|e| e.x).into_option() else {
    continue;
};
let Some((min_y, max_y)) = corners.iter().minmax_by_key(|e| e.y).into_option() else {
    continue;
};

node.top = Val::Px(min_y.y);
node.left = Val::Px(min_x.x);
node.width = Val::Px(max_x.x - min_x.x);
node.height = Val::Px(max_y.y - min_y.y);
```

Here is simpler example:
Image

UI scaling in this case is 2, so it takes half of the logical pixel to see UI moving.

ickshonpe:
# Objective

Allow users to enable or disable layout rounding for specific UI nodes and their descendants.

Fixes #16731

## Solution

New component `LayoutConfig` that can be added to any UiNode entity. Setting the `use_rounding` field of `LayoutConfig` determines if the Node and its descendants should be given rounded or unrounded coordinates.

## Testing

Not tested this extensively but it seems to work and it's not very complicated. 
This really basic test app returns fractional coords:

```rust
use bevy::prelude::*;

fn main() {
    App::new()
        .add_plugins(DefaultPlugins)
        .add_systems(Startup, setup)
        .add_systems(Update, report)
        .run();
}

    commands.spawn(Camera2d);
    commands.spawn((
        Node {
            left: Val::Px(0.1),
            width: Val::Px(100.1),
            height: Val::Px(100.1),
            ..Default::default()
        },
        LayoutConfig { use_rounding: false },
    ));
}

fn report(node: Query<(Ref<ComputedNode>, &GlobalTransform)>) {
    for (c, g) in node.iter() {
        if c.is_changed() {
            println!("{:#?}", c);
            println!("position = {:?}", g.to_scale_rotation_translation().2);
        }
    }
}
```


alice-i-cecile:
@matrixdev can I get your review here please?

alice-i-cecile (approved):
Sad to add a branch here, but the use case in the linked issue is very reasonable.

MatrixDev:
@alice-i-cecile, I can confirm that it becomes significantly smoother with this change and no longer pixel-aligned.

With `use_rounding: false`:
sc

With `use_rounding: true`:
cc

There is still one problem with bars but I don't know if it is related or not. All tree bars (reb, blue and yellow) are inside of the column with zero row_gap/colum_gap and ideally those should always stick to each other but sometimes there are gaps shown between them (in both with or without rounding):
Screenshot 2024-12-16 at 22 20 49
Screenshot 2024-12-16 at 22 20 55
Screenshot 2024-12-16 at 22 21 01
Screenshot 2024-12-16 at 22 21 07

In any case, from my point of view, original issue is resolved now.

crates/bevy_ui/src/layout/ui_surface.rs
```diff
@@ -277,11 +277,21 @@ impl UiSurface {
...
+    ) -> Result<(taffy::Layout, Vec2), LayoutError> {
         let Some(taffy_node) = self.entity_to_taffy.get(&entity) else {
             return Err(LayoutError::InvalidHierarchy);
         };
 
+        if use_rounding {
+            self.taffy.enable_rounding();
+        } else {
+            self.taffy.disable_rounding();
+        }
```
UkoeHB, line 293: This is potentially leaky if the `self.taffy.layout(..)` errors-out. It seems this function should always end with rounding enabled.

crates/bevy_ui/src/ui_node.rs
```diff
@@ -2528,6 +2528,23 @@ impl Default for ShadowStyle {
...
+#[derive(Component, Copy, Clone, Debug, PartialEq, Reflect)]
+#[reflect(Component, Debug, PartialEq, Default)]
+#[cfg_attr(
+    feature = "serialize",
+    derive(serde::Serialize, serde::Deserialize),
+    reflect(Serialize, Deserialize)
+)]
+pub struct LayoutConfig {
+    pub use_rounding: bool,
+}
```
UkoeHB, line 2540: Needs docs

UkoeHB (changes requested):

JeanMertz:
Any reason to add this as a field to the new `LayoutConfig` struct, as opposed to a newtype struct (or enum) specifically for this purpose? Do we expect more layout options to be configurable soon?

ickshonpe:
> Sad to add a branch here, but the use case in the linked issue is very reasonable.

Yeah taffy's rounding api is really unpleasant with how this setting is used both to enable or disable rounding during updates, and then afterwards to access the unrounded values. I'll make a taffy PR as well with some alternative then we'd be able to clean this up too later.

crates/bevy_ui/src/layout/ui_surface.rs
ickshonpe: Yeah this function is fine I think as this is the only place we read the layout values and we reset the rounding mode each time we enter it, but we need to call `enable_rounding()` each frame before we recompute the layout since like you say the update walk could end with `disable_rounding` set.

ickshonpe:
> Any reason to add this as a field to the new `LayoutConfig` struct, as opposed to a newtype struct (or enum) specifically for this purpose? Do we expect more layout options to be configurable soon?

I'm not really sure. I'm not against using an enum or an `UnroundedLayout` marker component or something instead. I guess I think if we did have any other similar niche configuration options it'd be preferable to consolidate them together in a single struct like this for the sake of discoverability. If there were other settings we'd have to add an inherit from parent option to the `use_rounding` setting.

crates/bevy_ui/src/layout/ui_surface.rs
UkoeHB: > Yeah this function is fine I think as this is the only place we read the layout values and we reset the rounding mode each time we enter it

Imo this reasoning is too fragile. It assumes the current design will stay the same, and it implies if the design does change then whoever changes it will somehow know the right thing to do.

crates/bevy_ui/src/layout/ui_surface.rs
```diff
@@ -209,6 +209,7 @@ impl UiSurface {
             width: taffy::style::AvailableSpace::Definite(render_target_resolution.x as f32),
             height: taffy::style::AvailableSpace::Definite(render_target_resolution.y as f32),
         };
+        self.taffy.enable_rounding();
         for root_nodes in camera_root_nodes {
```
UkoeHB, line 213:
```suggestion
        };
        for root_nodes in camera_root_nodes {
```
Looks like a leftover

crates/bevy_ui/src/ui_node.rs
```diff
@@ -2550,6 +2550,26 @@ impl Default for ShadowStyle {
...
+#[cfg_attr(
+    feature = "serialize",
+    derive(serde::Serialize, serde::Deserialize),
+    reflect(Serialize, Deserialize)
+)]
+/// This component can be added to any UI node to modify its layout behavior.
+pub struct LayoutConfig {
+    /// If set to true the coordinates for this node and its descendents will be rounded to the nearest physical pixel.
+    /// This can help prevent visual artifacts like blurry images or semi-transparent edges that can occur with sub-pixel positioning.
+    pub use_rounding: bool,
```
UkoeHB, line 2564:
```suggestion
    /// This can help prevent visual artifacts like blurry images or semi-transparent edges that can occur with sub-pixel positioning.
    ///
    /// Defaults to true.
    pub use_rounding: bool,
```

UkoeHB (changes requested):

BenjaminBrienen (approved):

PR #16843, opened 2024-12-16
Support declaring resource access in Queries.

chescock:
# Objective

Allow resources to be accessed soundly by `QueryData` and `QueryFilter` implementations.  

This mostly works today, and is used in `bevy-trait-query` and will be used by #16810.  The problem is that the access is not made visible to the executor, so it would be possible for a system with resource access in a query to run concurrently with a system that accesses the resource with `ResMut`, resulting in Undefined Behavior.  

## Solution

Define calling `add_resource_read` or `add_resource_write` in `WorldQuery::update_component_access` to be a supported way to declare resource access in a query.  
Modify `QueryState::new_with_access` to check for resource access and report it in `archetype_component_acccess`.  
Modify `FilteredAccess::is_compatible` to consider resource access conflicting even on queries with disjoint filters.

### DIFF
--- a/crates/bevy_ecs/src/query/state.rs
+++ b/crates/bevy_ecs/src/query/state.rs
@@ -187,6 +187,24 @@ impl<D: QueryData, F: QueryFilter> QueryState<D, F> {
             }
         }
         state.archetype_generation = world.archetypes.generation();
+
+        // Resource access is not part of any archetype and must be handled separately
+        if state.component_access.access().has_read_all_resources() {
+            access.read_all_resources();
+        } else {
+            for component_id in state.component_access.access().resource_reads() {
+                access.add_resource_read(world.initialize_resource_internal(component_id).id());
+            }
+        }
+
+        if state.component_access.access().has_write_all_resources() {
+            access.write_all_resources();
+        } else {
+            for component_id in state.component_access.access().resource_writes() {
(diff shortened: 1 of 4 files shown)


crates/bevy_ecs/src/query/state.rs
```diff
@@ -187,6 +187,24 @@ impl<D: QueryData, F: QueryFilter> QueryState<D, F> {
...
+        // Resource access is not part of any archetype and must be handled separately
+        if state.component_access.access().has_read_all_resources() {
+            access.read_all_resources();
+        } else {
+            for component_id in state.component_access.access().resource_reads() {
+                access.add_resource_read(world.initialize_resource_internal(component_id).id());
+            }
+        }
+
+        if state.component_access.access().has_write_all_resources() {
```
chescock, line 200: This will also support *mutable* resource access in queries, mainly because it was straightforward to do so, and because allowing `access.add_resource_write` to compile but do nothing seemed dangerous.  But it's hard to do anything useful with mutable resource access.  Multiple query items may be alive at once, so you can't use a mutable borrow in the fetch item.  

Would it make more sense to simply prohibit mutable access to resources, and panic here if there is any?

alice-i-cecile (approved):
Straightforward. I'm excited to use this for indexes.

bushrat011899 (approved):
Looks good to me!

Issue #16829, opened 2024-12-15
`ui_material` example should show how to use `Globals`


rparrett:
## How can Bevy's documentation be improved?

It's easy for a user to get tripped up by using the wrong globals import, e.g. from `bevy_pbr`. We don't seem to show the correct import anywhere in our docs or examples.

Something like
```c
#import bevy_render::globals::Globals

@group(0) @binding(1) var<uniform> globals: Globals;

// globals.time
```
Seems to work, but I'm not 100% sure if this is the canonical way to do it.

PR #16846, opened 2024-12-16
Fix inaccurate comment in custom_ui_material.wgsl shader

### LINKED ISSUE #16830: Comments in `custom_ui_material.wgsl` are outdated

In #14031, this shader was changed so that it renders a progress bar. But the comments still say that it renders a circle.

https://github.com/bevyengine/bevy/blob/09d229201699730766ca452c462ff7ef2ba34ec1/assets/shaders/custom_ui_material.wgsl#L1

kurk070ff:
# Objective

- Modify a comment in the shader file to describe what the shader actually does
- Fixes #16830

## Solution

- Changed the comment.

## Testing

- Testing is not relevant to fixing comments (as long as the comment is accurate)

### DIFF
--- a/assets/shaders/custom_ui_material.wgsl
+++ b/assets/shaders/custom_ui_material.wgsl
@@ -1,4 +1,5 @@
-// This shader draws a circle with a given input color
+// This shader draws a progress bar that changes its color based on time elapsed
+// and cycles between 0% full and 100% full
 #import bevy_ui::ui_vertex_output::UiVertexOutput
 
 @group(1) @binding(0) var<uniform> color: vec4<f32>;
--- a/assets/shaders/custom_ui_material.wgsl
+++ b/assets/shaders/custom_ui_material.wgsl
@@ -1,5 +1,4 @@
-// This shader draws a progress bar that changes its color based on time elapsed
-// and cycles between 0% full and 100% full
+// Draws a progress bar with properties defined in CustomUiMaterial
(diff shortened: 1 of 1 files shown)


alice-i-cecile (approved):

BenjaminBrienen (approved):

rparrett (changes requested):
This is an improvement, but to me it sounds like it is claiming some responsibility in the shader for stuff that is happening in rust code in the example.

The shader is just responsible for drawing the bar according to those material properties that are modified there.

This distinction is sort of relevant to #16829 actually, because this sort of animation *could* be done in the shader, but that's not what's going on here at the moment.

kurk070ff:
> This is an improvement, but to me it sounds like it is claiming some responsibility in the shader for stuff that is happening in rust code in the example.
> 
> The shader is just responsible for drawing the bar according to those material properties that are modified there.
> 
> This distinction is sort of relevant to #16829 actually, because this sort of animation _could_ be done in the shader, and people probably want to know how to do that, but that's not what's going on here at the moment.

So it's better to describe as "The shader draws a progress bar in accordance with the data from the game/rust side of the game"?

rparrett:
Maybe something like

> Draws a progress bar with properties defined in `CustomUiMaterial`.

alice-i-cecile (approved):

rparrett (approved):

PR #16857, opened 2024-12-17
Resource Params as references. `fn system(_res: &Res){}`

eckz:
# Objective

- In the same way that queries can be written like `Query<&Component>`, one would expect to refer to resources as `&Resource` while writing systems.
 - I couldn't find any issue or pull request talking about this in the past, but if this is not align with the API that Bevy wants to provide, I can  close this pull request. I understand this would be a controversial change, but it would be a big quality of life improvement for many systems

## Solution

- This PR allows resources to be use just by reference by implementing `SystemParam` for `&T where T: Resource `.
- The behaviour should be exactly the same one as `Res<T>` without the access to track changes.
- `Res<T>` will continue to exist and it can be used when it makes sense, like `Ref` exists for components.
- Honestly I'm not sure if this was not possible in the past, and I was not expecting this change to even compile, but it seem to work just fine without any disadvantages

## Testing

-. 
- `cargo run -p ci`: OK
- `MIRIFLAGS="-Zmiri-ignore-leaks -Zmiri-disable-isolation" cargo +nightly miri test -p bevy_ecs`: OK

---

## Showcase

With this PR, some systems that don't need the methods provided by `Res` can be simplified, common examples are `ButtonInput<T>`, `AssetServer` or `Time`.

Now you can write systems like this:

```rust
fn my_system(keyboard: &ButtonInput<KeyCode>, asset_server: &AssetServer, time: &Time) {
    // Do stuff
}

```

### DIFF
--- a/crates/bevy_ecs/src/schedule/mod.rs
+++ b/crates/bevy_ecs/src/schedule/mod.rs
@@ -60,7 +60,7 @@ mod tests {
         world.resource_mut::<SystemOrder>().0.push(u32::MAX);
     }
 
-    fn counting_system(counter: Res<Counter>) {
+    fn counting_system(counter: &Counter) {
         counter.0.fetch_add(1, Ordering::Relaxed);
     }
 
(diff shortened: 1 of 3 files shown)


alice-i-cecile:
Hmm. I don't think that this is an improvement. I can see why your intuition might guide you here, but it's better to have one clear way to do things :)

`&MyResource` would be nice, but we can't use it as a system param, because then you lose the change detection bits stored in the `Res`/`ResMut` smart pointers.

PR #16858, opened 2024-12-17
Add benchmarks and `compile_fail` tests back to workspace

### LINKED ISSUE #16801: Add benchmarks and compile fail tests back to workspace
The benchmarks and `compile_fail` UI tests are currently excluded from the main Cargo workspace:

https://github.com/bevyengine/bevy/blob/00722b8d0fb1224db99d161b92fd892595fc2823/Cargo.toml#L16-L22

This is a problem because these things get checked less often, and often lag behind. For example, in #16733 I had to fix a compile error for a function that was removed in 0.10 (we just released 0.15). The best solution to prevent this from happening in the future is to re-add the benchmarks and UI tests to the main Cargo workspace.

Past attempt: #16770

BD103:
# Objective

- Our benchmarks and `compile_fail` tests lag behind the rest of the engine because they are not in the Cargo workspace, so not checked by CI.
- Fixes #16801, please see it for further context!

## Solution

- Add benchmarks and `compile_fail` tests to the Cargo workspace.
- Fix any leftover formatting issues and documentation.

## Testing

- I think CI should catch most things!

## Questions

<details>
  <summary>Outdated issue I was having with function reflection being optional</summary>

The `reflection_types` example is failing in Rust-Analyzer for me, but not a normal check.

```rust
error[E0004]: non-exhaustive patterns: `ReflectRef::Function(_)` not covered
   --> examples/reflection/reflection_types.rs:81:11
    |
81  |     match value.reflect_ref() {
    |           ^^^^^^^^^^^^^^^^^^^ pattern `ReflectRef::Function(_)` not covered
    |
note: `ReflectRef<'_>` defined here
   --> /Users/bdeep/dev/bevy/bevy/crates/bevy_reflect/src/kind.rs:178:1
    |
178 | pub enum ReflectRef<'a> {
    | ^^^^^^^^^^^^^^^^^^^^^^^
...
188 |     Function(&'a dyn Function),
    |     -------- not covered
    = note: the matched value is of type `ReflectRef<'_>`
help: ensure that all possible cases are being handled by adding a match arm with a wildcard pattern or an explicit pattern as shown
    |
126 ~         ReflectRef::Opaque(_) => {},
127 +         ReflectRef::Function(_) => todo!()
    |
```

I think it is because the following line is feature-gated:

https://github.com/bevyengine/bevy/blob/cc0f6a8db43581755bd84302072c7b97ea51bc0f/examples/reflection/reflection_types.rs#L117-L122

My theory for why this is happening is because the benchmarks enabled `bevy_reflect`'s `function` feature, which gets merged with the rest of the features when RA checks the workspace, but the `#[cfg(...)]` gate in the example isn't detecting it:

https://github.com/bevyengine/bevy/blob/cc0f6a8db43581755bd84302072c7b97ea51bc0f/benches/Cargo.toml#L19

Any thoughts on how to fix this? It's not blocking, since the example still compiles as normal, but it's just RA and the command `cargo check --workspace --all-targets` appears to fail.

</summary>

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,6 +13,7 @@ documentation = "https://docs.rs/bevy"
 rust-version = "1.83.0"
 
 [workspace]
+resolver = "2"
 members = [
   # All of Bevy's official crates are within the `crates` folder!
   "crates/*",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,20 +13,19 @@ documentation = "https://docs.rs/bevy"
 rust-version = "1.83.0"
 
 [workspace]
-exclude = [
-  "benches",
-  "crates/bevy_derive/compile_fail",
-  "crates/bevy_ecs/compile_fail",
-  "crates/bevy_reflect/compile_fail",
-  "tools/compile_fail_utils",
-]
 members = [
+  # All of Bevy's official crates are within the `crates` folder!
   "crates/*",
+  # Several crates with macros have "compile fail" tests nested inside them, also known as UI
+  # tests, that verify diagnostic output does not accidentally change.
+  "crates/*/compile_fail",
+  # Examples of compiling Bevy for mobile platforms.
   "examples/mobile",
-  "tools/ci",
-  "tools/build-templated-pages",
-  "tools/build-wasm-example",
-  "tools/example-showcase",
+  # Benchmarks
+  "benches",
+  # Internal tools that are not published.
+  "tools/*",
+  # Bevy's error codes. This is a crate so we can automatically check all of the code blocks.
   "errors",
 ]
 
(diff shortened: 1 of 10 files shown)


alice-i-cecile:
I take it the RA bug you mentioned is now fixed?

BD103:
> I take it the RA bug you mentioned is now fixed?

No, not yet unfortunately, and it's also surfaced in the CI failure. There are a few solutions, but I don't like any of them:

1. Conditionally-compile function reflection benchmarks, no longer enable the feature by default in `benches/Cargo.toml`.
2. Enable function reflection by default and assume it's available in the example.
3. Don't add `benches` to the Cargo workspace.
4. Change how we detect function reflection in the example. I have no idea how we'd do this, but it would be preferable.

BD103:
I chose to feature-flag the function reflection benchmarks. While not amazing, it's more desirable then the other solutions I thought of. @MrGVSV next time you want to benchmark function reflection, you will need to run this:

```sh
cargo bench --bench reflect --feature reflect_functions
```

alice-i-cecile (approved):

BenjaminBrienen (approved):

PR #16862, opened 2024-12-17
Add `SubApp::take_extract()`

### LINKED ISSUE #16850: Manually insert `RenderEntity` / reverse extract

The new retained render world allows (and, forces) synchonizing entities between both worlds. This can be done via the `entity_sync_system()` with the `SyncToRenderWorld` component, but there's no way to manually insert the `RenderEntity` component on an entity of the `MainWorld`. The reverse is possible; you simply use the extract phase to read the `RenderEntity` from the main world, and write the `MainEntity` in the render world. But since there's no (as I know) "reverse extract" where the render world is read-only and the main world can be mutated, it's impossible to manually insert the `RenderEntity` into the main world with custom logic.

One of:
- Have a `ReverseExtract` phase at the end of rendering, and helper system param to do render->main sync.
- Allow the extract phase to mutate the main world (incidentally, this allows stealing resources from main to render without copying them, which can't be done simply if the main world is immutable, and requires workarounds like interior mutability).

I'm kind of stuck because I looked at `entity_sync_system()` and it's using a hard-coded hook into the `SubApp`, where it's explicitly called before the `extract(main, render)` call. Since I can neither 1) read the existing `SubApp::extract` function (to be able to call it from a custom one which would replace it); nor 2) add a second custom extract function to the render app, as a consequence I cannot install my own sync system with custom logic because I would overwrite the default Bevy one and break all extract systems.

## Additional context

I'm trying to convert Hanabi to use the retained render world, and need to sync the particle effect on the main world and the cache one on the render world, both pointing to each other. I believe this should greatly simplify/clarify the code, by leveraging ECS to store the data instead of custom containers, and allow more granular and clear systems to operate on partial effect data (as opposed to know, few systems doing many different things on one giant blob of data per effect). However I want to run custom logic when creating the render app entity, because I only want to do so if the main world component passed a few basic validations, to prune potentially invalid effects. I also hate having that `SyncToRenderWorld` marker component which is unrelated to my library, and is redundant with the fact all entities with my particle component should already synchronize by design. Finally, the `SyncWorldPlugin` says the sync is done at the very beginning of the frame, but I want this sync done just before extract, because I don't want to incur a one frame delay on newly spawned effects.

djeedai:
# Objective

Fixes #16850

## Solution

Add a new function `SubApp::take_extract()`, similar to `Option::take()`, which allows stealing the currently installed extract function of a sub-app, with the intent to replace it with a custom one calling the original one via `set_extract()`.

This pattern enables registering a custom "world sync" function similar to the existing one `entity_sync_system()`, to run custom world sync logic with mutable access to both the main and render worlds.

## Testing

`cargo r -p ci` currently doesn't build locally, event after upgrading rustc to latest and doing a `cargo update`.

### DIFF
--- a/crates/bevy_app/src/sub_app.rs
+++ b/crates/bevy_app/src/sub_app.rs
@@ -182,8 +182,8 @@ impl SubApp {
     ///     // [...]
     ///
     ///     // Call Bevy's default, which executes the Extract phase
-    ///     if let Some(fn) = default_fn.as_mut() {
-    ///         fn(main, render);
+    ///     if let Some(f) = default_fn.as_mut() {
+    ///         f(main, render);
     ///     }
     ///
     ///     // Do post-extract custom logic
--- a/crates/bevy_app/src/sub_app.rs
+++ b/crates/bevy_app/src/sub_app.rs
@@ -165,6 +165,32 @@ impl SubApp {
         self
     }
 
+    /// Take the function that will be called by [`extract`](Self::extract) out of the app, if any was set,
+    /// and replace it with `None`.
+    ///
+    /// If you use Bevy, `bevy_render` will set a default extract function used to extract data from
+    /// the main world into the render world as part of the Extract phase. In that case, you cannot replace
+    /// it with your own function. Instead, take the Bevy default function with this, and install your own
+    /// instead which calls the Bevy default.
+    ///
+    /// ```
+    /// # let mut app = SubApp::new();
+    /// let default_fn = app.take_extract();
+    /// app.set_extract(move |(main, render)| {
+    ///     // Do pre-extract custom logic
+    ///     // [...]
+    ///
+    ///     // Call Bevy's default, which executes the Extract phase
+    ///     default_fn(main, render);
+    ///
+    ///     // Do post-extract custom logic
+    ///     // [...]
+    /// });
+    /// ```
+    pub fn take_extract(&mut self) -> Option<ExtractFn> {
+        self.extract.take()
+    }
+
     /// See [`App::insert_resource`].
     pub fn insert_resource<R: Resource>(&mut self, resource: R) -> &mut Self {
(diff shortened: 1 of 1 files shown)


djeedai:
`cargo r -p ci`

```
error: constant `CURRENT_SCENE` is never used
   --> examples/testbed/3d.rs:147:11
    |
147 |     const CURRENT_SCENE: super::Scene = super::Scene::Bloom;
    |           ^^^^^^^^^^^^^
    |
    = note: `-D dead-code` implied by `-D warnings`
    = help: to override `-D warnings` add `#[allow(dead_code)]`

error: function `setup` is never used
   --> examples/testbed/3d.rs:149:12
    |
149 |     pub fn setup(
    |            ^^^^^

error: constant `CURRENT_SCENE` is never used
   --> examples/testbed/3d.rs:197:11
    |
197 |     const CURRENT_SCENE: super::Scene = super::Scene::Gltf;
    |           ^^^^^^^^^^^^^

error: function `setup` is never used
   --> examples/testbed/3d.rs:199:12
    |
199 |     pub fn setup(mut commands: Commands, asset_server: Res<AssetServer>) {
    |            ^^^^^

error: could not compile `bevy` (example "testbed_3d") due to 4 previous errors
```

alice-i-cecile:
That will be fixed by #16866.

djeedai:
Note that local CI (`cargo r -p ci`) still fails because it seems to run docs in stable but uses attributes of nightly:

```
error[E0554]: `#![feature]` may not be used on the stable release channel
 --> src\lib.rs:2:21
  |
2 | #![cfg_attr(docsrs, feature(doc_auto_cfg))]
  |   
```

hymm (approved):
Seems like a reasonable way of doing this without having to know what the underlying sub app is doing.

alice-i-cecile (approved):

PR #16863, opened 2024-12-17
Replace bevy_a11y::Focus with InputFocus

alice-i-cecile:
# Objective

Bevy now has first-class input focus handling! We should use this for accessibility purpose via accesskit too.

## Solution

- Removed bevy_a11y::Focus.
- Replaced all usages of Focus with InputFocus
- Changed the dependency tree so bevy_a11y relies on bevy_input_focus
- Moved initialization of the focus (starts with the primary window) from bevy_window to bevy_input_focus to avoid circular dependencies (and it's cleaner)

## Testing

TODO

## Migration Guide

`bevy_a11y::Focus` has been replaced with `bevy_input_focus::Focus`.

### DIFF
--- a/crates/bevy_a11y/src/lib.rs
+++ b/crates/bevy_a11y/src/lib.rs
@@ -22,11 +22,10 @@ use accesskit::Node;
 use bevy_app::Plugin;
 use bevy_derive::{Deref, DerefMut};
 use bevy_ecs::{
-    prelude::{Component, Entity, Event, ReflectResource},
+    prelude::{Component, Event},
     schedule::SystemSet,
     system::Resource,
 };
-use bevy_reflect::{std_traits::ReflectDefault, Reflect};
 
 /// Wrapper struct for [`accesskit::ActionRequest`]. Required to allow it to be used as an `Event`.
 #[derive(Event, Deref, DerefMut)]
@@ -97,11 +96,6 @@ impl From<Node> for AccessibilityNode {
     }
 }
 
-/// Resource representing which entity has keyboard focus, if any.
-#[derive(Resource, Default, Deref, DerefMut, Reflect)]
-#[reflect(Resource, Default)]
-pub struct Focus(pub Option<Entity>);
-
 /// Set enum for the systems relating to accessibility
 #[derive(Debug, Hash, PartialEq, Eq, Clone, SystemSet)]
 pub enum AccessibilitySystem {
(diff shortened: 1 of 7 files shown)


crates/bevy_a11y/src/lib.rs
```diff
@@ -97,11 +96,6 @@ impl From<Node> for AccessibilityNode {
     }
 }
 
-/// Resource representing which entity has keyboard focus, if any.
-#[derive(Resource, Default, Deref, DerefMut, Reflect)]
```
alice-i-cecile, line 101: I didn't keep the Deref / DerefMut derives around: do you want me to readd them?

crates/bevy_winit/src/accessibility.rs
```diff
@@ -241,7 +243,7 @@ fn update_adapter(
     TreeUpdate {
         nodes: to_update,
         tree: None,
-        focus: NodeId(focus.unwrap_or(primary_window_id).to_bits()),
+        focus: NodeId(focus.0.unwrap_or(primary_window_id).to_bits()),
```
alice-i-cecile, line 246: Diff is due to the removed Deref / DerefMut impl.

crates/bevy_input_focus/src/lib.rs
```diff
@@ -376,6 +385,9 @@ mod tests {
         };
         app.world_mut().spawn((window, PrimaryWindow));
 
+        // Run the world for a single frame to set up the initial focus
+        app.update();
```
alice-i-cecile, line 389: This was needed as we're now setting the input focus in a startup system, rather than inside the windowing event loop.

crates/bevy_input_focus/src/lib.rs
```diff
@@ -149,12 +149,21 @@ pub struct InputDispatchPlugin;
...
+            .insert_resource(InputFocus(None))
             .insert_resource(InputFocusVisible(false))
             .add_systems(PreUpdate, dispatch_keyboard_input);
     }
 }
 
+/// Sets the initial focus to the primary window, if any.
+pub fn set_initial_focus(
+    mut input_focus: ResMut<InputFocus>,
+    window: Single<Entity, With<PrimaryWindow>>,
```
bushrat011899, line 162: I forget, does `Single` panic when there _isn't_ a single matching entity, or does this system just not run? I know there was some heated debate around the defaults here.

bushrat011899 (approved):
This makes sense to me. `Focus` is something more broadly applicable than just for accessibility, Just a minor note around the use of `Single`. LGTM

viridia (approved):

crates/bevy_input_focus/src/lib.rs
alice-i-cecile: System just won't run :)

eidloi:
@alice-i-cecile this locks out developers from using the `bevy_a11y` create entirely without going through the input focus crate, which is neither suitable nor correctly implements focus tracking (or is actually functional).

alice-i-cecile:
If you have concerns about the focus tracking, I would encourage you to open issues to that effect.

PR #16866, opened 2024-12-17
Feature flag testbed_3d code correctly

alice-i-cecile:
# Objective

Rust-Analyzer was reporting problems with dead code in the 3d testbed scene.

## Solution

These scenes don't work in CI on the Windows runner (because they're too weak).

Mirror the feature flags from above onto the offending modules.

## Testing

RA no longer complains.

### DIFF
--- a/examples/testbed/3d.rs
+++ b/examples/testbed/3d.rs
@@ -138,6 +138,7 @@ mod light {
     }
 }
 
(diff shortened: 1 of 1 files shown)


bushrat011899 (approved):
My awful Windows environment will work again!

djeedai (approved):

PR #16870, opened 2024-12-17
Remove `OnceLock` usage from `bevy_ecs`

### LINKED ISSUE #16868: Remove OnceLock's in bevy_ecs by using const
We might be able to get rid of these `OnceLock`s completely now that `const` is more powerful.  I think we could make `read_all_resources()` a `const fn` and then do `let access = &const { <what init does now> }`.

That isn't actually related to this PR, though, so it should probably be a follow-up.

_Originally posted by @chescock in https://github.com/bevyengine/bevy/pull/16758#discussion_r1887633983_

bushrat011899:
# Objective

- Fixes #16868

## Solution

- Replaced several usages of `OnceLock` within `bevy_ecs` with `const`s

## Testing

- CI

### DIFF
--- a/crates/bevy_ecs/Cargo.toml
+++ b/crates/bevy_ecs/Cargo.toml
@@ -130,7 +130,6 @@ portable-atomic-util = { version = "0.2.4", features = [
 spin = { version = "0.9.8", default-features = false, features = [
   "spin_mutex",
   "rwlock",
-  "once",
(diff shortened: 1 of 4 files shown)


BenjaminBrienen (approved):

alice-i-cecile (approved):
Very nice!

PR #16874, opened 2024-12-17
Add `no_std` support to `bevy_app`

bushrat011899:
# Objective

- Contributes to #15460

## Solution

  - `std` (default)
  - `bevy_tasks` (default)
  - `downcast ` (default)
  - `portable-atomic`
  - `critical-section`
- `downcast` and `bevy_tasks` are now optional dependencies for `bevy_app`.

## Testing

- CI
- Personal UEFI and Raspberry Pi Pico demo applications compile and run against this branch

## Draft Release Notes

Bevy's application framework now supports `no_std` platforms.

Following up on `bevy_ecs` gaining `no_std` support, `bevy_app` extends the functionality available on these targets to include the powerful `App` and `Plugin` abstractions. With this, library authors now have the option of making their plugins `no_std` compatible, or even offering plugins specifically to improve Bevy on certain embedded platforms!

To start making a `no_std` compatible plugin, simply disable default features when including `bevy_app`:

```toml
[dependencies]
bevy_app = { version = "0.16", default-features = false }
```

We encourage library authors to do this anyway, as it can also help with compile times and binary size on all platforms.

Keep an eye out for future `no_std` updates as we continue to improve the parity between `std` and `no_std`. We look forward to seeing what kinds of applications are now possible with Bevy!

## Notes

- `downcast-rs` is optional as it isn't compatible with `portable-atomic`. I will investigate making a PR upstream to add support for this functionality, as it should be very straightforward.
- In line with the `bevy_ecs` no-std-ification, I've added documentation to all features, and grouped them as well.
- ~~Creating this PR in draft while CI runs and so I can polish before review.~~

### DIFF
--- a/crates/bevy_app/src/app.rs
+++ b/crates/bevy_app/src/app.rs
@@ -2,6 +2,10 @@ use crate::{
     First, Main, MainSchedulePlugin, PlaceholderPlugin, Plugin, Plugins, PluginsState, SubApp,
     SubApps,
 };
+use alloc::{
+    boxed::Box,
+    string::{String, ToString},
+};
 pub use bevy_derive::AppLabel;
 use bevy_ecs::{
     component::RequiredComponentsError,
@@ -11,15 +15,22 @@ use bevy_ecs::{
     schedule::{ScheduleBuildSettings, ScheduleLabel},
     system::{IntoObserverSystem, SystemId, SystemInput},
 };
-#[cfg(feature = "trace")]
-use bevy_utils::tracing::info_span;
-use bevy_utils::{tracing::debug, HashMap};
+use bevy_utils::HashMap;
 use core::{fmt::Debug, num::NonZero, panic::AssertUnwindSafe};
+use log::debug;
+use thiserror::Error;
+
+#[cfg(feature = "trace")]
+use tracing::info_span;
+
+#[cfg(feature = "std")]
 use std::{
     panic::{catch_unwind, resume_unwind},
     process::{ExitCode, Termination},
 };
-use thiserror::Error;
+
+#[cfg(feature = "downcast")]
+use alloc::vec::Vec;
 
 bevy_ecs::define_label!(
     /// A strongly-typed class of labels used to identify an [`App`].
@@ -458,12 +469,21 @@ impl App {
             .push(Box::new(PlaceholderPlugin));
 
         self.main_mut().plugin_build_depth += 1;
-        let result = catch_unwind(AssertUnwindSafe(|| plugin.build(self)));
+
+        let f = AssertUnwindSafe(|| plugin.build(self));
+
+        #[cfg(feature = "std")]
+        let result = catch_unwind(f);
+
+        #[cfg(not(feature = "std"))]
+        f();
+
         self.main_mut()
             .plugin_names
             .insert(plugin.name().to_string());
         self.main_mut().plugin_build_depth -= 1;
 
+        #[cfg(feature = "std")]
         if let Err(payload) = result {
             resume_unwind(payload);
         }
@@ -499,6 +519,7 @@ impl App {
     /// # app.add_plugins(ImagePlugin::default());
     /// let default_sampler = app.get_added_plugins::<ImagePlugin>()[0].default_sampler;
     /// ```
+    #[cfg(feature = "downcast")]
     pub fn get_added_plugins<T>(&self) -> Vec<&T>
     where
         T: Plugin,
@@ -1294,7 +1315,7 @@ type RunnerFn = Box<dyn FnOnce(App) -> AppExit>;
 
 fn run_once(mut app: App) -> AppExit {
     while app.plugins_state() == PluginsState::Adding {
-        #[cfg(not(target_arch = "wasm32"))]
+        #[cfg(all(not(target_arch = "wasm32"), feature = "bevy_tasks"))]
         bevy_tasks::tick_global_task_pools_on_main_thread();
     }
     app.finish();
@@ -1364,6 +1385,7 @@ impl From<u8> for AppExit {
     }
 }
 
+#[cfg(feature = "std")]
 impl Termination for AppExit {
     fn report(self) -> ExitCode {
         match self {
--- a/crates/bevy_app/src/lib.rs
+++ b/crates/bevy_app/src/lib.rs
@@ -11,6 +11,7 @@
     html_logo_url = "https://bevyengine.org/assets/icon.png",
     html_favicon_url = "https://bevyengine.org/assets/icon.png"
 )]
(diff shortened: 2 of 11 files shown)


crates/bevy_ecs/Cargo.toml
```diff
@@ -88,8 +88,9 @@ critical-section = [
 portable-atomic = [
   "dep:portable-atomic",
   "dep:portable-atomic-util",
-  "bevy_tasks/portable-atomic",
+  "bevy_tasks?/portable-atomic",
   "concurrent-queue/portable-atomic",
+  "spin/portable_atomic",
```
bushrat011899, line 93: The changes to `bevy_ecs/Cargo.toml` _could_ be split into their own PR, but for now I'm just including them here for brevity. I'm happy to split it out if anyone has any concerns.

tools/ci/src/commands/compile_check_no_std.rs
```diff
@@ -102,6 +102,14 @@ impl Prepare for CompileCheckNoStdCommand {
             "Please fix compiler errors in output above for bevy_ecs no_std compatibility.",
         ));
 
+        commands.push(PreparedCommand::new::<Self>(
+            cmd!(
+                sh,
+                "cargo check -p bevy_app --no-default-features --features bevy_reflect --target {target}"
```
bushrat011899, line 108: Now that `bevy_app` is compatible, we can start looking into replacing/augmenting this `compile_check_no_std` CI task with something more robust and tangible, like a `no_std` example crate.

crates/bevy_app/src/schedule_runner.rs
```diff
@@ -121,7 +127,10 @@ impl Plugin for ScheduleRunnerPlugin {
                     {
                         loop {
                             match tick(&mut app, wait) {
-                                Ok(Some(delay)) => std::thread::sleep(delay),
+                                Ok(Some(_delay)) => {
+                                    #[cfg(feature = "std")]
+                                    std::thread::sleep(_delay);
```
bushrat011899, line 132: Without access to `std::thread::sleep`, `no_std` apps will run as-fast-as-possible by default. This isn't ideal, but I believe a reasonable compromise.

crates/bevy_app/src/schedule_runner.rs
```diff
@@ -3,7 +3,10 @@ use crate::{
     plugin::Plugin,
     PluginsState,
 };
-use bevy_utils::{Duration, Instant};
+use bevy_utils::Duration;
+
+#[cfg(any(target_arch = "wasm32", feature = "std"))]
+use bevy_utils::Instant;
```
bushrat011899, line 9: I'm not a fan of how we bring this `Instant` in. I think it'd be worth considering some kind of `bevy_platform_support` crate which moves things like `spin`, `portable-atomic`, `critical-section`, `wasm-time`, etc. into a single spot. It'd be more focussed than `bevy_utils`, and possibly clean up some of the `cfg(...)` noise we have across Bevy.

crates/bevy_app/src/plugin.rs
```diff
@@ -1,8 +1,16 @@
...
 
 use crate::App;
 use core::any::Any;
 
+#[cfg(not(feature = "downcast"))]
+#[doc(hidden)]
+pub trait Downcast {}
+
+#[cfg(not(feature = "downcast"))]
+impl<T: ?Sized> Downcast for T {}
```
bushrat011899, line 12: `downcast-rs` isn't compatible with `portablie-atomic` (though it could be with a PR I believe!), so I've included a little dummy trait so the `Plugin: Downcast` bound can remain.

crates/bevy_app/src/app.rs
```diff
@@ -11,15 +15,22 @@ use bevy_ecs::{
...
-#[cfg(feature = "trace")]
-use bevy_utils::tracing::info_span;
-use bevy_utils::{tracing::debug, HashMap};
+use bevy_utils::HashMap;
 use core::{fmt::Debug, num::NonZero, panic::AssertUnwindSafe};
+use log::debug;
+use thiserror::Error;
+
+#[cfg(feature = "trace")]
+use tracing::info_span;
```
bushrat011899, line 24: As with `bevy_ecs`, I'm flattening our relationship with `bevy_utils::tracing` and also relying on `log` for non-span logging. It's already been mentioned that we should probably move things like `log` and `tracing` into workspace dependencies, but I'm just maintaining status quo for now.

crates/bevy_app/src/lib.rs
```diff
@@ -23,7 +24,7 @@ mod plugin;
 mod plugin_group;
 mod schedule_runner;
 mod sub_app;
-#[cfg(not(target_arch = "wasm32"))]
+#[cfg(all(not(target_arch = "wasm32"), feature = "std"))]
 mod terminal_ctrl_c_handler;
```
bushrat011899, line 28: Can't use Control-C in the terminal if your platform doesn't have a terminal...or a keyboard...

crates/bevy_app/src/sub_app.rs
```diff
@@ -332,6 +333,7 @@ impl SubApp {
     }
 
     /// See [`App::get_added_plugins`].
+    #[cfg(feature = "downcast")]
     pub fn get_added_plugins<T>(&self) -> Vec<&T>
```
bushrat011899, line 337: This is the only real consequence of losing `downcast-rs` support on certain `no_std` targets (not all!). I would like to bring feature parity regardless, but it is fairly minor IMO.

crates/bevy_app/src/schedule_runner.rs
alice-i-cecile: Yes please!

crates/bevy_app/src/sub_app.rs
```diff
@@ -332,6 +333,7 @@ impl SubApp {
     }
 
     /// See [`App::get_added_plugins`].
+    #[cfg(feature = "downcast")]
```
alice-i-cecile, line 336: Leave a TODO to clean this up later please, linking to the PR upstream.

alice-i-cecile:
No serious problems here; let me know when this is ready.

crates/bevy_app/src/sub_app.rs
bushrat011899: Made a `TODO` comment, I'll update it with a link to the PR when it's published too.

BenjaminBrienen (approved):

heydocode (approved):

alice-i-cecile:
Thank you to everyone involved with the authoring or reviewing of this PR! This work is relatively important and needs release notes! Head over to https://github.com/bevyengine/bevy-website/issues/1977 if you'd like to help out.

PR #16876, opened 2024-12-18
Generalize bubbling focus input events to other kinds of input

alice-i-cecile:
# Objective

The new `bevy_input_focus` crates has a tool to bubble input events up the entity hierarchy, ending with the window, based on the currently focused entity. Right now though, this only works for keyboard events!

Both `bevy_ui` buttons and `bevy_egui` should hook into this system (primarily for contextual hotkeys), and we would like to drive `leafwing_input_manager` via these events, to help resolve longstanding pain around "absorbing" / "consuming" inputs based on focus. In order to make that work properly though, we need gamepad support!

## Solution

The logic backing this has been changed to be generic for any cloneable event types, and the machinery to make use of this externally has been made `pub`. 

Within the engine itself, I've added support for gamepad button and scroll events, but nothing else. Mouse button / touch bubbling is handled via bevy_picking, and mouse / gamepad motion doesn't really make sense to bubble.

## Testing

The `tab_navigation` example continues to work, and CI is green.

## Future Work

I would like to add more complex UI examples to stress test this, but not here please.

We should take advantage of the bubbled mouse scrolling when defining scrolled widgets.

### DIFF
--- a/crates/bevy_input_focus/src/lib.rs
+++ b/crates/bevy_input_focus/src/lib.rs
@@ -20,16 +20,10 @@ pub mod tab_navigation;
 
 use bevy_app::{App, Plugin, PreUpdate};
 use bevy_ecs::{
-    component::Component,
-    entity::Entity,
-    event::{Event, EventReader},
-    query::{QueryData, With},
-    system::{Commands, Query, Res, Resource, SystemParam},
-    traversal::Traversal,
-    world::{Command, DeferredWorld, World},
+    prelude::*, query::QueryData, system::SystemParam, traversal::Traversal, world::DeferredWorld,
 };
 use bevy_hierarchy::{HierarchyQueryExt, Parent};
-use bevy_input::keyboard::KeyboardInput;
+use bevy_input::{gamepad::GamepadButtonChangedEvent, keyboard::KeyboardInput};
 use bevy_window::{PrimaryWindow, Window};
 use core::fmt::Debug;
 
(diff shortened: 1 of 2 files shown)


NthTensor (approved):
Interesting! This makes `FocusedInput` equivalent to `Pointer` in some respects. 

I'm fine to merge this, but hopefully I will have time to take another pass at traversal soon and we can generalize this even further. I want to be able to pull the focused window directly out of the ecs during traversal instead of having to embed it in the event type.

alice-i-cecile:
Yep, I'm increasingly convinced that we can probably unify this with `Pointer` completely TBH. One step at a time though, and this is moving in the right direction.

viridia (approved):

eidloi:
@alice-i-cecile you probably didn't notice but the bubbling does not consider ghost nodes at all. Since those are UI only, this won't work for UI.

eidloi:
> Yep, I'm increasingly convinced that we can probably unify this with `Pointer` completely TBH. One step at a time though, and this is moving in the right direction.

you should be sus at this point that the input focus create is entirely redundant and offers no value besides adding a layer of complexity and plumbing to get to the pointer. Focus management only ever needed moving a couple markers around, so I hope I don't have to re-implement the entire picking too to avoid this. Please.

alice-i-cecile:
@eidloi, you have been [previously sanctioned](https://github.com/bevyengine/bevy/pull/15341#issuecomment-2394453522) for a code of conduct violation on a closely related topic. While that has expired, your tone here is unproductive and disrespectful. There will not be another warning.

That said, I'm always interested in gathering constructive feedback, and we may have missed something important.

> @alice-i-cecile you probably didn't notice but the bubbling does not consider ghost nodes at all. Since those are UI only, this won't work for UI.

As far as I can see, the bubbling should work fine. Can you explain what you mean more clearly?

> you should be sus at this point that the input focus create is entirely redundant and offers no value besides adding a layer of complexity and plumbing to get to the pointer

The input_focus crate works without a pointer at all: that's why the keyboard-based tab navigation exists. Making sure that our UI functions even without a pointer-based device is critical for accessibility.

>  Focus management only ever needed moving a couple markers around

We've discussed the resource-based vs marker-component design quite a bit. I think that the resource-based design integrates more cleanly with `accesskit`, and does a better job of preserving the key invariant: at most one entity can be focused at once.

Like I said in https://github.com/bevyengine/bevy/pull/16863#issuecomment-2551982774, tangible, constructive feedback on `bevy_input_focus` is welcome. It's still early in development, and I want to make sure we get it right!

eidloi:
wffon8de3uxd1

@alice-i-cecile I sincerely apologize, it's the stress. My comments came out rude which is unintended, but I feel the criminal treatment is unwarranted. I don't have time to type out my suggestions, my wife gave birth yesterday. I'll open an issue once I have time to type out my explanation & original meaning of my comments. Sorry again.

PR #16878, opened 2024-12-18
Retain `RenderMaterialInstances` and `RenderMeshMaterialIds` from frame to frame.

pcwalton:
This commit makes Bevy use change detection to only update `RenderMaterialInstances` and `RenderMeshMaterialIds` when meshes have been added, changed, or removed. `extract_mesh_materials`, the system that extracts these, now follows the pattern that
`extract_meshes_for_gpu_building` established.

This improves frame time of `many_cubes` from 3.9ms to approximately 3.1ms, which slightly surpasses the performance of Bevy 0.14.

Screenshot 2024-12-17 182109

### DIFF
--- a/crates/bevy_pbr/src/material.rs
+++ b/crates/bevy_pbr/src/material.rs
@@ -37,6 +37,7 @@ use bevy_render::{
     render_phase::*,
     render_resource::*,
     renderer::RenderDevice,
+    sync_world::MainEntity,
     view::{ExtractedView, Msaa, RenderVisibilityRanges, ViewVisibility},
     Extract,
 };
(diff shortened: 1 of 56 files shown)


JMS55:
CI is failing

crates/bevy_pbr/src/material.rs
```diff
@@ -582,28 +583,121 @@ pub const fn screen_space_specular_transmission_pipeline_key(
...
+/// resources from the meshes in the scene.
 fn extract_mesh_materials<M: Material>(
     mut material_instances: ResMut<RenderMaterialInstances<M>>,
     mut material_ids: ResMut<RenderMeshMaterialIds>,
     mut material_bind_group_allocator: ResMut<MaterialBindGroupAllocator<M>>,
-    query: Extract<Query<(Entity, &ViewVisibility, &MeshMaterial3d<M>)>>,
+    changed_meshes_query: Extract<
+        Query<
+            (Entity, &ViewVisibility, &MeshMaterial3d<M>),
+            Or<(Changed<ViewVisibility>, Changed<MeshMaterial3d<M>>)>,
```
JMS55, line 595: Question on this: MeshMaterial3d might not change, but the asset pointed to by its handle might change. Do we need to handle that here? I don't have the context of what extract is doing here anymore.

pcwalton: We don't need to handle that here because the only thing `extract_mesh_materials` needs to do is to allocate within the slabs and store the resulting texture *indices*. If the material changes, then whatever textures (and other bindings) the material has will be updated in-place within the slabs, so the indices won't change, so this system doesn't need to do anything.

tychedelia (changes requested):
Would you mind updating `extract_mesh_materials_2d` as well to make sure these don't diverge too much?

pcwalton:
Maybe #16825 should land first, so we can make sure meshlets don't regress.

pcwalton:
I don't know what happened to this branch, but I needed to force-push to fix it, sorry.

pcwalton:
@tychedelia Done

pcwalton:
I'm going to close and resubmit this so that the commit history is cleaned up.

PR #16881, opened 2024-12-18
Update linux dependency documentation for NixOS (and nix)

to-bak:
# Objective
- Updates the linux-dependencies docs to 1) fix problems with nix `alsa-lib`, and 2) include additional documentation to run bevy with nix on non NixOS systems. (nix is a package manager that can be run outside of NixOS).

## Solution
1. the nix `alsa-lib` package doesn't include `alsa-plugins`, which bevy depends on. Instead, use the `alsa-lib-with-plugins` package, which wraps `alsa-plugins` into `alsa-lib`. For more information see: https://github.com/NixOS/nixpkgs/pull/277180
2. using nix on non NixOS systems, software like `nixGL` is required to correctly link graphics drivers. 

## Testing
- Tested on ubuntu 22.04 with nix.


mockersf:
`nixGL` seems to be a wrapper for OpenGL, but also contains `nixVulkan` which wraps Vulkan. Could you mention it first instead of running with `nixGL`?

NthTensor (commented):
Can't speak to the change, as I'm on `NixOS` and these instructions are for using the `nix` package manager on other operating systems. But seems to be correct.

NthTensor (approved):

to-bak:
> `nixGL` seems to be a wrapper for OpenGL, but also contains `nixVulkan` which wraps Vulkan. Could you mention it first instead of running with `nixGL`?

I updated the documentation to make this more clear. Please have another look

BenjaminBrienen (approved):

PR #16884, opened 2024-12-18
Add dashed lines

### LINKED ISSUE #16873: Make `GizmoLineStyle::Dotted` more configurable

Right now there is no way to configure the distance between segments. For example, in my game it looks like this:

Image

But I trying to make it look like this:

Image

It would be nice to add a length parameter to the enum.

lynn-lumen:
# Objective

- Fixes #16873

## Solution

- Added  `GizmoLineStyle::Dashed {gap_scale, line_scale}`
  - The `gap_scale` and `line_scale` describe the lengths of the gaps and visible line-segments in terms of line-widths. For example, if `gap_scale == 1.0` and `line_scale == 3.0` the gaps are square and the the visible segments are three line-widths long.
- The new `GizmoLineStyle` can be used both in 3D and 2D and with both perspective and orthographic cameras.
- Updated the `2d_gizmos` and `3d_gizmos` examples to include the new line-style.
- Display a warning, when using negative `gap_scale` or `line_scale`.
- Notably, `Hash` and `Eq` are manually implemented for `GizmoLineStyle` since both are not implemented for `f32` which prevents deriving these traits for `GizmoLineStyle`.

## Testing

- The results can be verified visually

---

## Showcase
The following images depict dashed lines with `gap_scale == 3.0` and `line_scale == 5.0` in perspective 3D and orthographic 2D.

linestyle-dashed-2d
linestyle-dashed-3d

### DIFF
--- a/crates/bevy_gizmos/src/config.rs
+++ b/crates/bevy_gizmos/src/config.rs
@@ -14,10 +14,10 @@ use bevy_reflect::{std_traits::ReflectDefault, Reflect, TypePath};
 use bevy_utils::TypeIdMap;
 use core::{
     any::TypeId,
+    hash::Hash,
     ops::{Deref, DerefMut},
     panic,
 };
-use std::hash::Hash;
 
 /// An enum configuring how line joints will be drawn.
 #[derive(Debug, Default, Copy, Clone, Reflect, PartialEq, Eq, Hash)]
--- a/crates/bevy_gizmos/src/config.rs
+++ b/crates/bevy_gizmos/src/config.rs
@@ -17,6 +17,7 @@ use core::{
     ops::{Deref, DerefMut},
     panic,
 };
+use std::hash::Hash;
 
 /// An enum configuring how line joints will be drawn.
 #[derive(Debug, Default, Copy, Clone, Reflect, PartialEq, Eq, Hash)]
@@ -36,7 +37,7 @@ pub enum GizmoLineJoint {
 }
 
 /// An enum used to configure the style of gizmo lines, similar to CSS line-style
-#[derive(Copy, Clone, Debug, Default, Hash, PartialEq, Eq, Reflect)]
+#[derive(Copy, Clone, Debug, Default, PartialEq, Reflect)]
 #[non_exhaustive]
 pub enum GizmoLineStyle {
     /// A solid line without any decorators
@@ -44,6 +45,32 @@ pub enum GizmoLineStyle {
     Solid,
     /// A dotted line
     Dotted,
+    /// A dashed line with configurable gap and line sizes
+    Dashed {
+        /// The length of the gap in `line_width`s
+        gap_scale: f32,
+        /// The length of the visible line in `line_width`s
+        line_scale: f32,
+    },
+}
+impl Eq for GizmoLineStyle {}
+impl Hash for GizmoLineStyle {
+    fn hash<H: core::hash::Hasher>(&self, state: &mut H) {
+        match self {
+            Self::Solid => {
+                0u64.hash(state);
+            }
+            Self::Dotted => 1u64.hash(state),
+            Self::Dashed {
+                gap_scale,
+                line_scale,
+            } => {
+                2u64.hash(state);
+                gap_scale.to_bits().hash(state);
+                line_scale.to_bits().hash(state);
+            }
+        }
+    }
 }
 
 /// A trait used to create gizmo configs groups.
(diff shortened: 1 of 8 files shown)


alice-i-cecile (approved):
Very pretty, and very easy. Thanks!

crates/bevy_gizmos/src/config.rs
```diff
@@ -36,14 +37,40 @@ pub enum GizmoLineJoint {
...
+    /// A dashed line with configurable gap and line sizes
+    Dashed {
+        /// The length of the gap in `line_width`s
+        gap_scale: f32,
+        /// The length of the visible line in `line_width`s
+        line_scale: f32,
+    },
+}
+impl Eq for GizmoLineStyle {}
+impl Hash for GizmoLineStyle {
```
Shatur, line 57: Let's add newlines for consistency:
```suggestion

impl Eq for GizmoLineStyle {}

impl Hash for GizmoLineStyle {
```

crates/bevy_gizmos/src/retained.rs
```diff
@@ -115,13 +118,30 @@ pub(crate) fn extract_linegizmos(
         } else {
             0
         };
+        let (gap_scale, line_scale) = if let GizmoLineStyle::Dashed {
+            gap_scale,
+            line_scale,
+        } = gizmo.line_config.style
+        {
+            if gap_scale <= 0.0 {
+                warn_once!("When using gizmos with the line style `GizmoLineStyle::Dashed{{..}}` the gap scale should be greater than zero.");
```
Shatur, line 127:
```suggestion
                warn_once!("when using gizmos with the line style `GizmoLineStyle::Dashed{{..}}` the gap scale should be greater than zero");
```

crates/bevy_gizmos/src/retained.rs
```diff
@@ -115,13 +118,30 @@ pub(crate) fn extract_linegizmos(
...
+            }
+            if line_scale <= 0.0 {
+                warn_once!("When using gizmos with the line style `GizmoLineStyle::Dashed{{..}}` the line scale should be greater than zero.");
```
Shatur, line 130: Warnings usually starts without capital letter and dot at the end:
```suggestion
                warn_once!("when using gizmos with the line style `GizmoLineStyle::Dashed{{..}}` the line scale should be greater than zero");
```

crates/bevy_gizmos/src/lib.rs
```diff
@@ -419,6 +419,9 @@ fn extract_gizmo_data(
     handles: Extract<Res<GizmoHandles>>,
     config: Extract<Res<GizmoConfigStore>>,
 ) {
+    use bevy_utils::warn_once;
+    use config::GizmoLineStyle;
```
Shatur, line 423: Maybe import globally?

Shatur (approved):
Awesome! Left a few minor nitpicks:

crates/bevy_gizmos/src/lines.wgsl
```diff
@@ -168,3 +175,14 @@ fn fragment_dotted(in: FragmentInput) -> FragmentOutput {
     
     return FragmentOutput(vec4(in.color.xyz, in.color.w * alpha));
 }
+@fragment
```
Shatur, line 178: A newline between functions:
```suggestion

@fragment
```

Shatur (changes requested):

Shatur (approved):

globin:
Would it be possible to extend this to not only allow `(line, gap)`, but rater `Vec<(line, gap)>` in order to produce a dash-dot/dash-dot-dot pattern etc?

lynn-lumen:
> Would it be possible to extend this to not only allow `(line, gap)`, but rater `Vec<(line, gap)>` in order to produce a dash-dot/dash-dot-dot pattern etc?
> 

I'm not sure if it's sensible, as it would require a pretty big adjustment to the current system, which is primarily meant to aid development. 

Currently, the `(line, gap)` pair is part of the gizmos-uniform whisch _should_ not hold large amounts of data and cannot hold dynamic amounts of data. As such, we would need smth. like an extra buffer for that, which seems fairly excessive for a development tool.

globin:
I am aware that gizmos were originally intended for debugging, but with retained mode and other improvements they have moved more toward a polyline implementation, which bevy currently is lacking. I am currently using bevy_prototype_lyon for it, but could imagine moving to retained Gizmos for that.

PR #16886, opened 2024-12-18
Add `Text2d` picking backend

coreh:
# Objective

- This PR adds support for picking `Text2d`.

## Solution

- Introduce a new feature flag, `bevy_text2d_picking_backend`, and add all the required plumbing;
- Extend the `TextPlugin` struct to include a `add_picking` setting;
- Port over the existing sprite picking backend to `bevy_text`, using `TextLayoutInfo` for the bounds;
- Add a `text2d_picking` example.

Note: I'm deliberately not calling the feature `bevy_text_picking_backend`, to avoid misleading people into thinking this might be used for UI Text.

## Testing

- Did you test these changes? If so, how? Yes, by running the `text2d_picking` example.
- Are there any parts that need more testing? Not that I'm aware of
- How can other people (reviewers) test your changes? Is there anything specific they need to know? Run the `text2d_picking` example.
- If relevant, what platforms did you test these changes on, and are there any important ones you can't test? macOS

### DIFF
--- a/crates/bevy_text/src/text2d_picking_backend.rs
+++ b/crates/bevy_text/src/text2d_picking_backend.rs
@@ -3,7 +3,7 @@
 //! 2d text is pickable even in its transparent areas.
 //!
 //! **Note:** This backend is pretty much a 1:1 port of the [`bevy_sprite`] picking backend,
-//! and *is not reposible for handling the picking of UI Text*. For that, please refer to the
+//! and *is not responsible for handling the picking of UI Text*. For that, please refer to the
 //! picking backend implementation under the `bevy_ui` crate.
 
 use core::cmp::Reverse;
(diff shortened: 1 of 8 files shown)


crates/bevy_text/src/text2d_picking_backend.rs
```diff
@@ -0,0 +1,176 @@
...
+                        picking_behavior,
+                    ))
+                } else {
+                    None
+                }
+            },
+        )
+        .collect();
+
+    sorted_text2d.sort_by_key(|x| Reverse(FloatOrd(x.3.translation().z)));
```
andriyDev, line 66: Can you update this to match #17174 ? I would be shocked if the perf characteristics weren't the same.

andriyDev (approved):

baozaolaoba-top:
@coreh  Dear coreh, thank you. This is a very useful feature. Could you update it? The changes are minimal and only include sorting/type names and the viewport.

PR #16889, opened 2024-12-18
Beef up the InputFocusVisible docs

alice-i-cecile:
# Objective

The docs for InputFocusVisible could do a better job explaining how the resource is intended to be used.

## Solution

Add more detail and do an editing pass. Link to the `IsFocused` trait for breadcrumbs too.


viridia (approved):

BenjaminBrienen (approved):
I think an example would be very valuable for this, too.

alice-i-cecile:
> I think an example would be very valuable for this, too.

Agreed: I'd like to extend the game menu example to cover tab navigation properly.

PR #16890, opened 2024-12-18
Fix several regressions from recent rendering changes.

pcwalton:
This commit fixes the following regressions:

1. Transmission-specific calls to shader lighting functions didn't pass the `enable_diffuse` parameter, breaking the `transmission` example.

2. The combination of bindless `StandardMaterial` and bindless lightmaps caused us to blow past the 128 texture limit on M1/M2 chips in some cases, in particular the `depth_of_field` example. https://github.com/gfx-rs/wgpu/issues/3334 should fix this, but in the meantime this patch reduces the number of bindless lightmaps from 16 to 4 in order to stay under the limit.

3. The renderer was crashing on startup on Adreno 610 chips. This PR simply disables bindless on Adreno 610 and lower.

### DIFF
--- a/crates/bevy_pbr/src/lightmap/mod.rs
+++ b/crates/bevy_pbr/src/lightmap/mod.rs
@@ -50,6 +50,7 @@ use bevy_reflect::{std_traits::ReflectDefault, Reflect};
 use bevy_render::{
     render_asset::RenderAssets,
     render_resource::{Sampler, Shader, TextureView, WgpuSampler, WgpuTextureView},
+    renderer::RenderAdapter,
     sync_world::MainEntity,
     texture::{FallbackImage, GpuImage},
     view::ViewVisibility,
@@ -71,7 +72,7 @@ pub const LIGHTMAP_SHADER_HANDLE: Handle<Shader> =
 ///
 /// If bindless textures aren't in use, then only a single lightmap can be bound
 /// at a time.
-pub const LIGHTMAPS_PER_SLAB: usize = 16;
+pub const LIGHTMAPS_PER_SLAB: usize = 4;
 
 /// A plugin that provides an implementation of lightmaps.
 pub struct LightmapPlugin;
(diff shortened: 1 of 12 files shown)


crates/bevy_render/src/batching/gpu_preprocessing.rs
```diff
@@ -356,26 +356,7 @@ impl FromWorld for GpuPreprocessingSupport {
 
         // filter some Qualcomm devices on Android as they crash when using GPU preprocessing.
         fn is_non_supported_android_device(adapter: &RenderAdapter) -> bool {
-            if cfg!(target_os = "android") {
-                let adapter_name = adapter.get_info().name;
-
-                // Filter out Adreno 730 and earlier GPUs (except 720, as it's newer than 730)
-                // while also taking suffixes into account like Adreno 642L.
```
rparrett, line 363: I think both lines of this comment would be useful to preserve. (but second should be moved to `get_adreno_model`)

alice-i-cecile:
I've started an example-runner test run for this :) We should expect diffs though.

rparrett (approved):
I am not qualified to review most of the rendering changes, but:

- Confirmed that this fixes `depth_of_field` on my m1 mac 
- Confirmed that this fixes `transmission`
- Changes related to Android make sense to me, but I can't test
- Didn't seem to cause any additional problems in the example run

crates/bevy_pbr/src/lightmap/mod.rs
```diff
@@ -71,7 +72,7 @@ pub const LIGHTMAP_SHADER_HANDLE: Handle<Shader> =
 ///
 /// If bindless textures aren't in use, then only a single lightmap can be bound
 /// at a time.
-pub const LIGHTMAPS_PER_SLAB: usize = 16;
+pub const LIGHTMAPS_PER_SLAB: usize = 4;
```
IceSentry, line 75: Is there a reason for the 4 specifically or is it just a nice number that happens to still work because it's small enough??

IceSentry (approved):

IceSentry:
I haven't tested anything locally, but the changes LGTM

PR #16891, opened 2024-12-18
Revert "Make doc CI use nightly (#16147)"

### LINKED ISSUE #16871: Local CI (`cargo r -p ci`) fails because it seems to run docs in stable but uses attributes of nightly:
Note that local CI (`cargo r -p ci`) still fails because it seems to run docs in stable but uses attributes of nightly:

```
error[E0554]: `#![feature]` may not be used on the stable release channel
 --> src\lib.rs:2:21
  |
2 | #![cfg_attr(docsrs, feature(doc_auto_cfg))]
  |   
```

_Originally posted by @djeedai in https://github.com/bevyengine/bevy/issues/16862#issuecomment-2549726848_

alice-i-cecile:
This reverts commit 58a73924ebf4b8afadaa48a9f0bbf2f057292a43.

# Objective

This is causing local `cargo run -p ci` runs to fail on `stable` Rust when used locally. Fixes #16871.

## Solution

Revert #16147. We shouldn't have merged this until we had a full solution (I missed a Controversial tag, sorry!).

## Testing

`cargo run -p ci` fails for me before this change in the reported fashion. After this reversion, it passes.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -126,7 +126,6 @@ jobs:
       - name: Check Compile
         # See tools/ci/src/main.rs for the commands this runs
         run: cargo run -p ci -- compile
-  
   check-compiles-no-std:
     runs-on: ubuntu-latest
     timeout-minutes: 30
@@ -295,7 +294,6 @@ jobs:
         with:
           name: example-run-macos
           path: example-run/
-        
   check-doc:
     runs-on: ubuntu-latest
     timeout-minutes: 30
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -126,7 +126,7 @@ jobs:
       - name: Check Compile
         # See tools/ci/src/main.rs for the commands this runs
         run: cargo run -p ci -- compile
-
+  
   check-compiles-no-std:
     runs-on: ubuntu-latest
     timeout-minutes: 30
@@ -295,7 +295,7 @@ jobs:
         with:
           name: example-run-macos
           path: example-run/
-
+        
   check-doc:
     runs-on: ubuntu-latest
     timeout-minutes: 30
@@ -310,9 +310,7 @@ jobs:
             ~/.cargo/git/db/
             target/
           key: ${{ runner.os }}-check-doc-${{ hashFiles('**/Cargo.toml') }}
-      - uses: dtolnay/rust-toolchain@master
-        with:
-          toolchain: ${{ env.NIGHTLY_TOOLCHAIN }}
+      - uses: dtolnay/rust-toolchain@stable
       - name: Install Linux dependencies
         uses: ./.github/actions/install-linux-deps
         with:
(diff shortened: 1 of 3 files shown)


.github/workflows/weekly.yml
```diff
@@ -80,6 +80,24 @@ jobs:
         # See tools/ci/src/main.rs for the commands this runs
         run: cargo run -p ci -- compile
 
+  check-doc:
```
alice-i-cecile, line 83: The reversion suggested I re-add this, but I'm not sure if that's correct.

.github/workflows/weekly.yml
```diff
@@ -98,7 +116,7 @@ jobs:
             --jq '.[0].number')
           if [[ -n $previous_issue_number ]]; then
             gh issue comment $previous_issue_number \
-              --body "Weekly pipeline still fails: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}"
+              --body "Weekly pipeline still fails: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}"
```
alice-i-cecile, line 119: We should probably keep the white space cleanup from that PR.

.github/workflows/ci.yml
```diff
@@ -126,7 +126,7 @@ jobs:
       - name: Check Compile
         # See tools/ci/src/main.rs for the commands this runs
         run: cargo run -p ci -- compile
-
+
```
alice-i-cecile, line 129:
```suggestion

```

.github/workflows/ci.yml
```diff
@@ -295,7 +295,7 @@ jobs:
         with:
           name: example-run-macos
           path: example-run/
-
+
```
alice-i-cecile, line 298:
```suggestion

```

.github/workflows/weekly.yml
alice-i-cecile, line 119:
```suggestion
              --body "Weekly pipeline still fails: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}"
```

.github/workflows/weekly.yml
```diff
@@ -110,6 +128,6 @@ jobs:
           GH_REPO: ${{ github.repository }}
           TITLE: Main branch fails to compile on Rust beta.
           LABELS: C-Bug,S-Needs-Triage
-          BODY: |
+          BODY: |
```
alice-i-cecile, line 131:
```suggestion
          BODY: |
```

.github/workflows/weekly.yml
BD103: Yup, that shouldn't have been reverted. If you look at the [original PR](https://github.com/bevyengine/bevy/pull/16147/files#diff-b077dee47bccc37c2b6fef32cdd102f36acadad73ede1f410c3941f39c50cd94), it doesn't remove this snippet from the code.

BD103 (approved):
Once the strange addition in `weekly.yml` is resolved, this is good to go!

BenjaminBrienen (approved):

Issue #16892, opened 2024-12-19
Split up `bevy_core`


ItsDoot:
## What problem does this solve or what need does it fill?

Title. This crate is basically a utilities crate, which we would like to avoid (`bevy_utils` is slowly being split up, for example). See here for its current API surface: https://docs.rs/bevy_core/latest/bevy_core/

## What solution would you like?

We should:
- Move `FrameCount` and its plugin to `bevy_diagnostics` or `bevy_window`
- Move other plugins to `bevy_app`
- Move everything else to `bevy_ecs`

## What alternative(s) have you considered?

Leave it as-is.

## Additional context

Related to #11478 and #16877.

[Originally brought up on the Bevy Discord.](https://discord.com/channels/691052431525675048/692572690833473578/1319072745010761749)

alice-i-cecile:
Related to #11478.

bushrat011899:
Recording here for visibility:

My list of little restructures I'd like:
* Move bevy_core::Name  into bevy_ecs
* Move bevy_core::TypeRegistrationPlugin and bevy_core::TaskPoolPlugin into bevy_app
* Move bevy_core::FrameCountPlugin into bevy_diagnostic
* Merge bevy_log into bevy_diagnostic (significant overlap between the two crates conceptually IMO)
* Move platform compatibility stuff out of bevy_utils and out of being inlined in crates and into its own new crate, bevy_platform_support (portable-atomic-util, hashbrown, web_time, etc.)
* Take all the Future and parallel stuff out of bevy_utils and put it in bevy_tasks
* Remove all the tracing/log stuff out of bevy_utils entirely, maybe putting some of it into bevy_diagnostic.
* Put things like bevy_utils::default and bevy_utils::assert_object_safe in a pair of micro-crates that'll never be updated ever again.

That'd remove bevy_utils, bevy_log, and bevy_core and replace them with 2 micro crates that can be moved out of the repo, and 1 more targeted crate bevy_platform_support

_Originally posted on [Discord](https://discord.com/channels/691052431525675048/692572690833473578/1319097890739716210)_

bushrat011899:
Also related to #16877

PR #16894, opened 2024-12-19
Move `Name` out of `bevy_core`

bushrat011899:
# Objective

- Contributes to #16892

## Solution

- Moved `Name` and `NameOrEntity` into `bevy_ecs::name`, and added them to the prelude.

## Testing

- CI

## Migration Guide

If you were importing `Name` or `NameOrEntity` from `bevy_core`, instead import from `bevy_ecs::name`.

### DIFF
--- a/crates/bevy_ecs/src/name.rs
+++ b/crates/bevy_ecs/src/name.rs
@@ -2,7 +2,10 @@
 
 use crate::{self as bevy_ecs, component::Component, entity::Entity, query::QueryData};
 
-use alloc::borrow::Cow;
+use alloc::{
+    borrow::{Cow, ToOwned},
+    string::String,
+};
 use bevy_utils::FixedHasher;
 use core::{
     hash::{BuildHasher, Hash, Hasher},
--- a/crates/bevy_ecs/src/name.rs
+++ b/crates/bevy_ecs/src/name.rs
@@ -109,7 +109,6 @@ impl core::fmt::Debug for Name {
 /// Convenient query for giving a human friendly name to an entity.
 ///
 /// ```
-/// # use bevy_core::prelude::*;
 /// # use bevy_ecs::prelude::*;
 /// # #[derive(Component)] pub struct Score(f32);
 /// fn increment_score(mut scores: Query<(NameOrEntity, &mut Score)>) {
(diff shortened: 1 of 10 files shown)


crates/bevy_ecs/src/name.rs
```diff
@@ -213,10 +221,44 @@ impl Deref for Name {
...
+
+#[cfg(feature = "serialize")]
+impl<'de> Deserialize<'de> for Name {
+    fn deserialize<D: Deserializer<'de>>(deserializer: D) -> Result<Self, D::Error> {
+        deserializer.deserialize_str(EntityVisitor)
+    }
+}
+
+#[cfg(feature = "serialize")]
+struct EntityVisitor;
```
ItsDoot, line 239: Could we give a better name for this while we're at it?

alice-i-cecile:
This will be one of the first "blessed" components, so it's not a fully uncontroversial decision. I think this is the right choice though, since my understanding is that this will be needed for the BSN macro.

crates/bevy_hierarchy/src/valid_parent_check_plugin.rs
```diff
@@ -53,7 +53,7 @@ impl<T> Default for ReportHierarchyIssue<T> {
 /// (See B0004 explanation linked in warning message)
 pub fn check_hierarchy_component_has_valid_parent<T: Component>(
     parent_query: Query<
-        (Entity, &Parent, Option<&bevy_core::Name>),
```
ItsDoot, line 56: Not necessary for this PR, but we could replace the `Option<&Name>` + `Entity` with `NameOrEntity`

ItsDoot (approved):
Looks good! Left a few comments

crates/bevy_ecs/src/name.rs
bushrat011899: Yeah I agree, it's a bad name. I've chosen `NameVisitor`, since it's a visitor for the `Name` component.

crates/bevy_hierarchy/src/valid_parent_check_plugin.rs
bushrat011899: Yeah I agree, I'll leave it off just to keep things less controversial, but it'd be a nice little follow-up.

PR #16897, opened 2024-12-19
Remove `bevy_core`

### LINKED ISSUE #16892: Split up `bevy_core`

bushrat011899:
# Objective

- Fixes #16892

## Solution

- Removed `TypeRegistryPlugin` (`Name` is now automatically registered with a default `App`)
- Moved `TaskPoolPlugin` to `bevy_app`
- Moved `FrameCountPlugin` to `bevy_diagnostic`
- Deleted now-empty `bevy_core`

## Testing

- CI

## Migration Guide

- `TypeRegistryPlugin` no longer exists. If you can't use a default `App` but still need `Name` registered, do so manually with `app.register_type::<Name>()`.
- References to `TaskPoolPlugin` and associated types will need to import it from `bevy_app` instead of `bevy_core`
- References to `FrameCountPlugin` and associated types will need to import it from `bevy_diagnostic` instead of `bevy_core`

## Notes

This strategy was agreed upon by Cart and several other members in [Discord](https://discord.com/channels/691052431525675048/692572690833473578/1319137218312278077).


bushrat011899:
I've marked this as `X-Blessed` based on Cart's approval of my proposal, but please remove it if that's inappropriate!

aevyrie (approved):
Probably worth noting the path change to `Name` for people migrating their scene files.

ItsDoot:
> Probably worth noting the path change to `Name` for people migrating their scene files.

That should be covered by the migration guide in #16894

bushrat011899:
> Probably worth noting the path change to `Name` for people migrating their scene files.

Yeah as @ItsDoot mentioned it's covered by #16894's migration guide. I just missed this instance and it slipped through CI, so I fixed it here since it's somewhat related.

PR #16906, opened 2024-12-19
Unrequire `ComputedNode`

### LINKED ISSUE #16904: `Display::None` propagation bug
## Bevy version

Main

## What you did

Added red outlines to the innermost node of Bevy's `display_and_visibility` example.

```cargo run --example display_and_visibility```

1. Without `Display::None` set:
Image

2. With `Display::None` set on the innermost node:
Image

3. With `Display::None` set on the outermost node:
Image

## What went wrong

During layout recomputation if a node is found with `Display::None` set, that node and all its descendants are given a zero size. 
This means most systems ignore the those nodes since you can't mouse-over or draw a zero-sized element. 

However zero-sized nodes can have outlines and outlines extraction only checks the local `Display::None` value for the outlined node and not the `Display` value of its ancestors. Which is why in `3` there is a visible outline around a UI node that should be removed from the layout and shouldn't even have a position to draw an outline around.

### PR DESCRIPTION
**Pull request by ickshonpe**
# Objective

UI nodes with `Display::None` set are meant to be removed from the layout and ignored but because all we do to signal removal is set the size to zero the inactive nodes get caught up in all the UI queries which is catastrophically terrible for performance and causes a few bugs.

Fixes #16904

## Solution

Remove `ComputedNode` from `Node`'s required components and instead add or remove it automatically in `ui_stack_system` during the walk to update the stack indices depending on the state of the `Node::display` field.

I bashed this out really quickly and it's a bit rough. I expect there to be some bugs and some things need a clean up. I'm convinced the idea is sound though and it's ready enough to test and review.

## Testing

```
cargo run --example many_buttons --release --features trace-tracy -- --display-none
```
unreq-computed-node

---

## Migration Guide

`ComputedNode` is no longer a required component for `Node`, instead it's added or removed automatically by `ui_stack_system`.

### CODE REVIEW

**Comment by alice-i-cecile** [comment]
Oh good catch. @pcwalton this might help explain your weird perf issues with UI layout.

After some chewing, I like this approach better than adding some disabling marker: it doesn't make sense for us to keep the data around due to how long-lived Display::None states are. It also prevents weird footguns where you're checking stale derived data.

**Comment by ickshonpe** [comment]
> Oh good catch. @pcwalton this might help explain your weird perf issues with UI layout.
>

I think his issue is because the stack indices, clipping rects and border values are alway recomputed every frame regardless of changes to the layout. It used to be not such a big deal but recent changes, especially ghost nodes, have made these updates a lot more expensive.
 
> After some chewing, I like this approach better than adding some disabling marker: it doesn't make sense for us to keep the data around due to how long-lived Display::None states are. It also prevents weird footguns where you're checking stale derived data.

Yeah lots of query filters can be confusing, I think this is definitely better.

File: crates/bevy_ui/src/lib.rs [hunk 1]
```diff
@@ -176,7 +176,7 @@ impl Plugin for UiPlugin {
                 PostUpdate,
                 (
                     CameraUpdateSystem,
-                    UiSystem::Prepare.before(UiSystem::Stack).after(Animation),
+                    UiSystem::Prepare.after(UiSystem::Stack).after(Animation),
                     UiSystem::Layout,
                     UiSystem::PostLayout,
```
UkoeHB, line 181: `UiSystem::Prepare` was explicitly added to be the first ui system set. That way you can reliably do `.before(UiSystem::Prepare)` and your systems will never mess up ui calculations.

It looks like you're using the results of `UiSystem::Stack` in systems in `Prepare` now. I think you need to do a larger rearranging of systems to accommodate this so `Prepare` remains the first set.

UkoeHB (changes requested):

cart:
Imo the alternative to consider is something like `ComputedNode::Enabled { /* computed state */  }` and `ComputedNode::Disabled`, or a field like `ComputedNode::enabled`. This has the benefit of not moving things around in the ECS as "normal" part of the flow, and allows us to spawn node trees without immediately doing an archetype move. The downside is you then need to check the state in the internals, but this feels pretty reasonable to me (and making it an enum variant would force that). I dont think ComputedNode is so big that we need to optimize it out whenever something isn't being displayed, especially when we (1) factor in the cost of archetype moves when switching and (2) consider that disabled nodes will essentially always be enabled at some point.

Ultimately I think adding enabled/disabled "runtime" states to components in Bevy ECS would be a nice generic solve to this problem. Lets us keep things in memory while also "hiding" components from systems. But in lieu of that feature, I'm kind of biased toward leaving ComputedNode as required and eating the complexity of checking the enabled flags in the few internal places where that matters.

crates/bevy_ui/src/stack.rs
```diff
@@ -40,32 +40,43 @@ impl ChildBufferCache {
...
     }
 
-    for (id, global_zindex, maybe_zindex) in zindex_global_node_query.iter() {
+    for (id, node, global_zindex, maybe_zindex) in zindex_global_node_query.iter() {
+        if node.display == Display::None {
+            commands.entity(id).remove::<ComputedNode>();
+        }
         if visited_root_nodes.contains(&id) {
             continue;
         }
```
UkoeHB, line 82:
```suggestion
        if visited_root_nodes.contains(&id) {
            continue;
        }
        if node.display == Display::None {
            commands.entity(id).remove::<ComputedNode>();
        }
```

crates/bevy_ui/src/stack.rs
```diff
@@ -40,32 +40,43 @@ impl ChildBufferCache {
...
-            id,
-            (
-                maybe_global_zindex.map(|zindex| zindex.0).unwrap_or(0),
-                maybe_zindex.map(|zindex| zindex.0).unwrap_or(0),
-            ),
-        ));
+    for (id, node, maybe_global_zindex, maybe_zindex) in
+        root_node_query.iter_many(ui_root_nodes.iter())
+    {
+        if node.display != Display::None {
```
UkoeHB, line 62: This behavior is not the same as the `zindex_global_node_query` case below (which looks correct)

crates/bevy_ui/src/stack.rs
```diff
@@ -94,20 +107,32 @@ pub fn ui_stack_system(
...
     ui_children: &UiChildren,
-    zindex_query: &Query<Option<&ZIndex>, (With<ComputedNode>, Without<GlobalZIndex>)>,
+    node_query: &Query<&Node>,
+    zindex_query: &Query<Option<&ZIndex>, (With<Node>, Without<GlobalZIndex>)>,
     ui_stack: &mut Vec<Entity>,
 ) {
+    if node_query.get(node_entity).unwrap().display == Display::None {
+        remove_computed_nodes_recursive(node_entity, commands, ui_children);
+    }
+
```
UkoeHB, line 131:
```suggestion
    if node_query.get(node_entity).unwrap().display == Display::None {
        remove_computed_nodes_recursive(node_entity, commands, ui_children);
        return;
    }

```

UkoeHB (changes requested):

PR #16913, opened 2024-12-20
many_components stress test improvements

hymm:
# Objective

- I was getting familiar with the many_components example to test some recent pr's for executor changes and saw some things to improve.

## Solution

- Use `insert_by_ids` instead of `insert_by_id`. This reduces the number of archetype moves and improves startup times substantially.
- Add a tracing span to `base_system`. I'm not sure why, but tracing spans weren't showing for this system. I think it's something to do with how pipe system works, but need to investigate more. The approach in this pr is a little better than the default span too, since it allows adding the number of entities queried to the span which is not possible with the default system span.
- println the number of archetype component id's that are created. This is useful since part of the purpose of this stress test is to test how well the use of FixedBitSet scales in the executor.

## Testing

- Ran the example with `cargo run --example many_components -F trace_tracy 1000000` and connected with tracy
- Timed the time it took to spawn 1 million entities on main (240 s) vs this pr (15 s)

---

## Showcase

image

## Future Work

- Currently systems are created with a random set of components and entities are created with a random set of components without any correlation between the randomness. This means that some systems won't match any entities and some entities could not match any systems. It might be better to spawn the entities from the pool of components that match the queries that the systems are using.

### DIFF
--- a/examples/stress_tests/many_components.rs
+++ b/examples/stress_tests/many_components.rs
@@ -162,7 +162,10 @@ fn stress_test(num_entities: u32, num_components: u32, num_systems: u32) {
         }
     }
 
-    println!("Number of Archetype-Components: {}", world.archetypes().archetype_components_len());
+    println!(
+        "Number of Archetype-Components: {}",
+        world.archetypes().archetype_components_len()
+    );
 
     // overwrite Update schedule in the app
     app.add_schedule(schedule);
--- a/examples/stress_tests/many_components.rs
+++ b/examples/stress_tests/many_components.rs
@@ -140,7 +140,7 @@ fn stress_test(num_entities: u32, num_components: u32, num_systems: u32) {
 
         let mut entity = world.spawn_empty();
         // We use `ManuallyDrop` here as we need to avoid dropping the u8's when `values` is dropped
-        // since ownership of the values is passed to the world in `insert_by_ids`. 
+        // since ownership of the values is passed to the world in `insert_by_ids`.
         // But we do want to deallocate the memory when values is dropped.
         let mut values: Vec<ManuallyDrop<u8>> = components
             .iter()
--- a/examples/stress_tests/many_components.rs
+++ b/examples/stress_tests/many_components.rs
@@ -143,10 +143,13 @@ fn stress_test(num_entities: u32, num_components: u32, num_systems: u32) {
             .iter()
             .map(|_id| {
                 let mut value: u8 = rng.gen_range(0..255);
-                let value = &mut value;
+                let value = &mut value as *mut u8;
                 // SAFETY: value is non null since it was initialized from an &mut above
                 #[allow(unsafe_code)]
-                let ptr = unsafe { NonNull::new_unchecked(value as *mut u8) };
+                let ptr = unsafe { NonNull::new_unchecked(value) };
+                // SAFETY:
+                // * ptr was created from a u8, so is valid, aligned, and has correct provenance.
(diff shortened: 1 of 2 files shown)


examples/stress_tests/many_components.rs
```diff
@@ -124,23 +127,40 @@ fn stress_test(num_entities: u32, num_components: u32, num_systems: u32) {
...
+                // SAFETY: value is non null since it was initialized from an &mut above
+                #[allow(unsafe_code)]
+                let ptr = unsafe { NonNull::new_unchecked(value as *mut u8) };
                 #[allow(unsafe_code)]
-                // SAFETY:
-                // component_id is from the same world
-                // value is u8, so ptr is a valid reference for component_id
                 unsafe {
-                    entity.insert_by_id(component_id, ptr);
+                    OwningPtr::new(ptr)
```
chescock, line 146: Isn't `ptr` pointing to a variable local to the closure?  I think this will be dangling by the time it gets passed to `insert_by_ids`.  You might need to create a `Vec<u8>` with values so that they can all be alive at once.

hymm: Whoops, you're right. Surprised this doesn't blow up.

hymm: Fixed it by storing it as a `Vec<ManuallyDrop<u8>>`.  The extra allocation slowed things down from 10s -> 15s.

examples/stress_tests/many_components.rs
```diff
@@ -124,23 +127,51 @@ fn stress_test(num_entities: u32, num_components: u32, num_systems: u32) {
...
+        // SAFETY:
+        // * component_id's are from the same world
+        // * `values` was initialized above, so references are valid
+        #[allow(unsafe_code)]
+        unsafe {
+            entity.insert_by_ids(&components, ptrs.into_iter());
         }
     }
 
+    println!(
```
chescock, line 170: Should this be `info!` so it can be configured with the logging framework?

examples/stress_tests/many_components.rs
chescock: If the allocation is the slow part, you might be able to get some perf back by re-using the `Vec`.  Declare it outside the loop, `extend()` to populate it, and `drain()` instead of `into_iter()` to consume it.  

I think `ManuallyDrop<u8>` is overkill since `u8` is already `Copy`, but it's harmless at worst.  

It might be a little safer to do `PtrMut::from(value).promote()` so that the compiler can check the lifetimes, but I'm pretty sure what you have is sound.

chescock (approved):

examples/stress_tests/many_components.rs
hymm: ah nice, I didn't know about that way of constructing a OwningPtr

examples/stress_tests/many_components.rs
hymm: I'm not sure. I see some examples use `info` and some use `println`.  This example was using println already, so I just kept it consistent.

BenjaminBrienen:
Needing unsafe in the benchmarks makes me feel like we could improve the API.

hymm:
> Needing unsafe in the benchmarks makes me feel like we could improve the API.

it's because we're spawning things dynamically, for which I don't think we can avoid unsafe.

JaySpruce (approved):

hymm:
I added a optional dependency on the tracing lib to the base Cargo.toml which is enabled with the `trace` feature. This shouldn't be a major change as tracing should already be in the dependency tree when that feature is enabled.

PR #16916, opened 2024-12-20
Add a custom render phase example

IceSentry:
# Objective

- It's currently very hard for beginners and advanced users to get a full understanding of a complete render phase

## Solution

- Implement a full custom render phase
- 
---

## Showcase

<details>

```rust
```

</details>

### DIFF
--- a/examples/shader/custom_render_phase.rs
+++ b/examples/shader/custom_render_phase.rs
@@ -8,22 +8,36 @@
 use std::ops::Range;
 
 use bevy::{
+    core_pipeline::core_3d::graph::{Core3d, Node3d},
     ecs::{
         entity::EntityHashSet,
-        query::ROQueryItem,
+        query::{QueryItem, ROQueryItem},
         system::{
             lifetimeless::{Read, SRes},
             SystemParamItem,
         },
     },
     math::FloatOrd,
-    pbr::{DrawMesh, RenderMeshInstances, SetMeshBindGroup, SetMeshViewBindGroup},
+    pbr::{
+        material_bind_groups::MaterialBindGroupSlot, DrawMesh, MeshInputUniform, MeshPipeline,
+        MeshPipelineKey, MeshPipelineViewLayoutKey, MeshUniform, RenderMeshInstances,
+        SetMeshBindGroup, SetMeshViewBindGroup,
+    },
     prelude::*,
 };
 use bevy_render::{
+    batching::{
+        gpu_preprocessing::{batch_and_prepare_sorted_render_phase, IndirectParametersBuffer},
+        GetBatchData, GetFullBatchData, NoAutomaticBatching,
+    },
+    camera::ExtractedCamera,
+    diagnostic::RecordDiagnostics,
     extract_component::{ExtractComponent, ExtractComponentPlugin, UniformComponentPlugin},
-    mesh::{MeshVertexBufferLayoutRef, RenderMesh},
+    mesh::{allocator::MeshAllocator, MeshVertexBufferLayoutRef, RenderMesh},
     render_asset::{PrepareAssetError, RenderAsset, RenderAssetPlugin, RenderAssets},
+    render_graph::{
+        NodeRunError, RenderGraphApp, RenderGraphContext, RenderLabel, ViewNode, ViewNodeRunner,
+    },
     render_phase::{
         sort_phase_system, AddRenderCommand, CachedRenderPipelinePhaseItem, DrawFunctionId,
         DrawFunctions, PhaseItem, PhaseItemExtraIndex, RenderCommand, RenderCommandResult,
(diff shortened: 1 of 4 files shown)


tychedelia (approved):
This is a great start. I think it would be cool if this did something more notable like an outline. I don't have many specific comments, other than I think it would be nice if we had a way to demonstrate a custom draw command here, but it's difficult for me to think what that would be. The only other thing is that we should demonstrate the cold specialization and retained bin pattern once those are settled, but this is also an issue with our other existing examples and out of scope here. This is going to be sooo helpful to share I think and we can def iterate more, tysm for doing this.

IceSentry:
Yeah, an outline would be cool, but it would require a lot more code and the example is already really long.

It's for the same reason I used a sorted phase and didn't use cold specialization. It made the code even more complicated.

IIRC we already have an example showing a custom draw command. If not, we could set the shader color from a uniform that is bound in a custom command. Shouldn't be too hard to add to this example.

ChristopherBiscardi (approved):
I've run the example on an m1 mac and previously used this example (plus a custom post_process) to implement [outlines](https://github.com/rust-adventure/bevy-examples/tree/39edda7da297564f694891ade8104fb4bd07708b/examples/edge-detection-custom-phase). 

image

I think the outlines are probably more than this PR should sustain, I like that it stops at rendering to the view as that's a concrete jumping-off point for anyone who wants to get started.

The example shows off a lot of useful concepts already. The only real downside I see is that adding hdr to the camera will break it due to the color attachment format (maybe worth changing in a future PR).

ChristopherBiscardi:
The miri CI failure was a github network request issue, so I've moved it to ready-for-final since all other ci is passing and it now has two reviews.

tychedelia:
> It's for the same reason I used a sorted phase and didn't use cold specialization. It made the code even more complicated.

Okay I'm convinced, it's definitely better to start small. 

> IIRC we already have an example showing a custom draw command.

It's the custom phase item example which this one kinda overlaps with. I don't think we need to make this an everything example, just interacting with some of the more complex retained rendering stuff we've added recently is on my radar as conspicuously missing from our examples atm. This is covering a huge gap though so let's def just start here!

examples/shader/custom_render_phase.rs
```diff
@@ -0,0 +1,630 @@
+//! This example demonstrates how to write a custom phase
+//!
+//! Render phases in bevy are used whenever you need to draw a group of meshes in a specific way.
+//! For example, bevy's main pass has an opaque phase, a transparent phase for both 2d and 3d.
+//! Sometimes, you may want to only draw a subset of meshes before or after the builtin phase. In
+//! those situations you need to write your own phase.
+//!
+//! This example showcases how writing a custom phase to draw a stencil of a bevy mesh could look
```
cBournhonesque, line 8: I tried to google it but could only find some vague definitions; what is a 'stencil' in this context?

examples/shader/custom_render_phase.rs
```diff
@@ -0,0 +1,630 @@
...
+    SetItemPipeline,
+    // This will set the view bindings in group 0
+    SetMeshViewBindGroup<0>,
+    // This will set the mesh bindings in group 1
+    SetMeshBindGroup<1>,
+    // This will draw the mesh
+    DrawMesh,
+);
+
+// This is the data required when you define a custom phase in bevy. More specifically this is the
```
cBournhonesque, line 248: I think it might be more clear to also mention that this is the data needed to render a single phase item (i.e. an entity)

examples/shader/custom_render_phase.rs
```diff
@@ -0,0 +1,630 @@
...
+            .add_render_command::<StencilPhase, DrawMesh3dStencil>()
+            .init_resource::<ViewSortedRenderPhases<StencilPhase>>()
+            .add_systems(ExtractSchedule, extract_camera_phases)
+            .add_systems(
+                Render,
+                (
+                    sort_phase_system::<StencilPhase>.in_set(RenderSet::PhaseSort),
+                    batch_and_prepare_sorted_render_phase::<StencilPhase, StencilPipeline>
+                        .in_set(RenderSet::PrepareResources),
+                    queue_custom_meshes.in_set(RenderSet::QueueMeshes),
```
cBournhonesque, line 132: I would put the systems in a different order, as I was a bit confused when reading this; it seems like the order of execution is actually
```
// queue entities with DrawStencil as PhaseItems for the new render phase
queue_custom_meshes.in_set(RenderSet::QueueMeshes),
// sort
sort_phase_system::<StencilPhase>.in_set(RenderSet::PhaseSort),
// batch
batch_and_prepare_sorted_render_phase::<StencilPhase, StencilPipeline>
                        .in_set(RenderSet::PrepareResources),
```

examples/shader/custom_render_phase.rs
```diff
@@ -0,0 +1,630 @@
...
+}
+
+impl CachedRenderPipelinePhaseItem for StencilPhase {
+    #[inline]
+    fn cached_pipeline(&self) -> CachedRenderPipelineId {
+        self.pipeline
+    }
+}
+
+impl GetBatchData for StencilPipeline {
```
cBournhonesque, line 334: What does the batching do in this case?

cBournhonesque (approved):
Super helpful example!
As a rendering noob, most of it made sense to me (the only thing I didn't understand is the batching logic and how it's used)

examples/shader/custom_render_phase.rs
IceSentry: It refers to essentially drawing the shape of the mesh. It's a bit of graphics programming terminology, but it's useful when you want to mask out something or want to create outlines. It's not really the purpose of the example though. It was just to have something a bit more concrete since it's most likely the most common use case for a custom phase.

examples/shader/custom_render_phase.rs
IceSentry: That's part of the future docs improvements I want to do, but the docs on the trait explain what is going on a bit more and I didn't want to just copy paste the doc, but maybe I should.

examples/shader/custom_render_phase.rs
IceSentry: I just wrote them in the order I originally wrote the systems. I could maybe try ordering everything based on the execution order.

PR #16917, opened 2024-12-20
impl EntityBorrow for more types

Victoronz:
# Objective

Some types like `RenderEntity` and `MainEntity` are just wrappers around `Entity`, so they should be able to implement `EntityBorrow`/`TrustedEntityBorrow`. This allows using them with `EntitySet` functionality.
The `EntityRef` family are more than direct wrappers around `Entity`, but can still benefit from being unique in a collection.

## Solution

Implement `EntityBorrow` and `TrustedEntityBorrow` for simple `Entity` newtypes and `EntityRef` types.
These impls are an explicit decision to have the `EntityRef` types compare like just `Entity`.
`EntityWorldMut` is omitted from this impl, because it explicitly contains a `&mut World` as well, and we do not ever use more than one at a time.

Add `EntityBorrow` to the `bevy_ecs` prelude.

## Migration Guide

`NormalizedWindowRef::entity` has been replaced with an `EntityBorrow::entity` impl.

### DIFF
--- a/crates/bevy_ecs/src/world/entity_ref.rs
+++ b/crates/bevy_ecs/src/world/entity_ref.rs
@@ -387,7 +387,7 @@ impl Eq for EntityRef<'_> {}
 
 #[expect(clippy::non_canonical_partial_ord_impl)]
 impl PartialOrd for EntityRef<'_> {
-    /// EntityRef's comparison trait implementations match the underlying [`Entity`],
+    /// [`EntityRef`]'s comparison trait implementations match the underlying [`Entity`],
     /// and cannot discern between different worlds.
     fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
         self.entity().partial_cmp(&other.entity())
@@ -925,7 +925,7 @@ impl Eq for EntityMut<'_> {}
 
 #[expect(clippy::non_canonical_partial_ord_impl)]
 impl PartialOrd for EntityMut<'_> {
-    /// EntityMut's comparison trait implementations match the underlying [`Entity`],
+    /// [`EntityMut`]'s comparison trait implementations match the underlying [`Entity`],
     /// and cannot discern between different worlds.
     fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
         self.entity().partial_cmp(&other.entity())
@@ -3063,7 +3063,7 @@ impl Eq for FilteredEntityRef<'_> {}
 
 #[expect(clippy::non_canonical_partial_ord_impl)]
 impl PartialOrd for FilteredEntityRef<'_> {
-    /// FilteredEntityRef's comparison trait implementations match the underlying [`Entity`],
+    /// [`FilteredEntityRef`]'s comparison trait implementations match the underlying [`Entity`],
     /// and cannot discern between different worlds.
     fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
         self.entity().partial_cmp(&other.entity())
@@ -3390,7 +3390,7 @@ impl Eq for FilteredEntityMut<'_> {}
 
 #[expect(clippy::non_canonical_partial_ord_impl)]
 impl PartialOrd for FilteredEntityMut<'_> {
-    /// FilteredEntityMut's comparison trait implementations match the underlying [`Entity`],
+    /// [`FilteredEntityMut`]'s comparison trait implementations match the underlying [`Entity`],
     /// and cannot discern between different worlds.
     fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
         self.entity().partial_cmp(&other.entity())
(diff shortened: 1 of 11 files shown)


crates/bevy_ecs/src/world/unsafe_world_cell.rs
```diff
@@ -650,9 +650,9 @@ impl Debug for UnsafeWorldCell<'_> {
 /// A interior-mutable reference to a particular [`Entity`] and all of its components
 #[derive(Copy, Clone)]
 pub struct UnsafeEntityCell<'w> {
-    world: UnsafeWorldCell<'w>,
-    entity: Entity,
-    location: EntityLocation,
+    pub(crate) world: UnsafeWorldCell<'w>,
+    pub(crate) entity: Entity,
```
chescock, line 654: Is it really necessary to loosen the privacy here?  

It looks like this is only needed so you can get an `&Entity` instead of an `Entity` to `impl Borrow<Entity>` for a few types.  Since we're no longer using `Borrow<Entity>` internally, could we just leave out those impls?  

Alternately, could we add a `pub (crate)` method to get `&Entity` so that it's clear nothing outside of this module modifies these fields?

crates/bevy_ecs/src/world/entity_ref.rs
```diff
@@ -869,6 +919,47 @@ impl<'a> TryFrom<&'a mut FilteredEntityMut<'_>> for EntityMut<'a> {
     }
 }
 
+impl PartialEq for EntityMut<'_> {
+    fn eq(&self, other: &Self) -> bool {
+        self.entity() == other.entity()
```
chescock, line 926: I understand why we need to have this, but it's a little amusing since there can't possibly be two `EntityMut`s alive for the same `entity`.  It's also not very useful with `Query::iter_many`, since you can never have an `EntityMut` that would match the query.  Well, it can get combined with this other impls as part of the `EntityPtr` work and then it won't seem odd anymore.

crates/bevy_ecs/src/world/entity_ref.rs
```diff
@@ -386,6 +386,8 @@ impl PartialEq for EntityRef<'_> {
 
 impl Eq for EntityRef<'_> {}
 
+// We intentionally choose to have EntityRef compare like Entity.
```
chescock, line 389: I suppose the danger here is that entities from different `world`s that happen to have the same ID will compare equal.  That seems unlikely to be a problem in practice, although it might be worth documenting somewhere.

crates/bevy_ecs/src/world/unsafe_world_cell.rs
Victoronz: right, there is a much simpler solution: I hadn't thought of just adding an `entity` method on `UnsafeEntityCell` itself for some reason. Should `UnsafeEntityCell` itself implement `EntityBorrow`?

crates/bevy_ecs/src/world/entity_ref.rs
Victoronz: Yeah, it is a bit unfortunate that we can't have both, but it seems to me that being able to compare them like `Entity` is a lot more valuable. I think I'll add some documentation to the types themselves.

crates/bevy_ecs/src/world/entity_ref.rs
Victoronz: You still can have its entity match another query! That is one thing the `nesting_queries()` test in `EntitySet` shows. But it is definitely more obscure.

crates/bevy_ecs/src/world/unsafe_world_cell.rs
Victoronz: Ah, I remembered why I didn't do this initially. Unless you return `&Entity` from a field, you will get a temporary lifetime error. So unless we increase the visibility here, or move the type to `entity_ref.rs`, we cannot impl `Borrow<Entity>` for the `EntityRef` types.
The reason why I added this impl is for the original purpose of `Borrow`: to allow its use in `Map` and `Set` types.
But I can omit it instead then.

crates/bevy_ecs/src/world/entity_ref.rs
chescock: Oh, I just meant `EntityMut` specifically.  It has write access to all components, so it's going to conflict with any other query that matches the same archetypes.  :)

crates/bevy_ecs/src/world/unsafe_world_cell.rs
chescock: I think it works if the method returns a reference to the original value: 

```rust
impl UnsafeEntityCell<'_> {
    pub(crate) fn entity_ref(&self) -> &Entity {
        &self.entity
    }
}

impl Borrow<Entity> for EntityMut<'_> {
    fn borrow(&self) -> &Entity {
        self.0.entity_ref()
    }
}
```

But it's a moot point now that you removed them.

chescock (approved):

crates/bevy_ecs/src/world/unsafe_world_cell.rs
Victoronz: ahhh, you're right! Had I taken my train of thought just a bit further I could've realized, we can just impl `Borrow<Entity>` and `EntityBorrow` for `UnsafeEntityCell`...
I'll add them back.

crates/bevy_ecs/src/world/entity_ref.rs
Victoronz: I also meant `EntityMut` specifically! You can still query for an archetype without accessing any components :)

bushrat011899 (approved):
Looks good! I like seeing gaps in the API filled like this.

PR #16922, opened 2024-12-21
Refactor `batch_and_prepare_binned_render_phase` in preparation for bin retention.

pcwalton:
This commit makes the following changes:

* `IndirectParametersBuffer` has been changed from a `BufferVec` to a `RawBufferVec`. This won about 20us or so on Bistro by avoiding `encase` overhead.

* The methods on the `GetFullBatchData` trait no longer have the `entity` parameter, as it was unused.

* `PreprocessWorkItem`, which specifies a transform-and-cull operation, now supplies the mesh instance uniform output index directly instead of having the shader look it up from the indirect draw parameters. Accordingly, the responsibility of writing the output index to the indirect draw parameters has been moved from the CPU to the GPU. This is in preparation for retained indirect instance draw commands, where the mesh instance uniform output index may change from frame to frame, while the indirect instance draw commands will be cached. We won't want the CPU to have to upload the same indirect draw parameters again and again if a batch didn't change from frame to frame.

* `batch_and_prepare_binned_render_phase` and `batch_and_prepare_sorted_render_phase` now allocate indirect draw commands for an entire batch set at a time when possible, instead of one batch at a time. This change will allow us to retain the indirect draw commands for whole batch sets.

* `GetFullBatchData::get_batch_indirect_parameters_index` has been replaced with `GetFullBatchData::write_batch_indirect_parameters`, which takes an offset and writes into it instead of allocating. This is necessary in order to use the optimization mentioned in the previous point.

* At the WGSL level, `IndirectParameters` has been factored out into `mesh_preprocess_types.wgsl`. This is because we'll need a new compute shader that zeroes out the instance counts in preparation for a new frame. That shader will need to access `IndirectParameters`, so it was moved to a separate file.

* Bins are no longer raw vectors but are instances of a separate type, `RenderBin`. This is so that the bin can eventually contain its retained batches.


IceSentry (approved):

crates/bevy_pbr/src/render/mesh_preprocess_types.wgsl
```diff
@@ -0,0 +1,19 @@
...
+// The `wgpu` indirect parameters structure. This is a union of two structures.
+// For more information, see the corresponding comment in
+// `gpu_preprocessing.rs`.
+struct IndirectParameters {
+    // `vertex_count` or `index_count`.
+    vertex_count_or_index_count: u32,
+    // `instance_count` in both structures.
+    instance_count: atomic<u32>,
+    // `first_vertex` in both structures.
+    first_vertex: u32,
```
kristoff3r, line 14: This comment does not agree with the name or comments on the Rust struct, there it's `first_vertex_or_first_index`.

crates/bevy_pbr/src/render/mesh_preprocess_types.wgsl
```diff
@@ -0,0 +1,19 @@
...
+    // `first_instance` or `base_vertex`.
+    first_instance_or_base_vertex: u32,
```
kristoff3r, line 16: In the Rust struct it's called `base_vertex_or_first_instance`, both are fine but they should agree I think.

kristoff3r (approved):

crates/bevy_pbr/src/render/mesh_preprocess_types.wgsl
pcwalton: I changed it to `base_vertex_or_first_instance`.

crates/bevy_pbr/src/render/mesh_preprocess_types.wgsl
pcwalton: I changed it to `first_vertex_or_first_index`.

alice-i-cecile:
Running an example check to spot unexpected regressions: https://github.com/TheBevyFlock/bevy-example-runner/issues/74

pcwalton:
Looks like this broke shadows. I'll look.

mockersf:
this fixed shadows, https://github.com/bevyengine/bevy/pull/16836 broke them

pcwalton:
Shadows still seem broken even with this PR. I wonder if the issue is intermittent.

PR #16926, opened 2024-12-22
Implement FromStr for Val

VictorKoenders:
# Objective

This PR implements `FromStr` for `Val`, so developers can parse values like `10px` and `50%`

## Testing

Added tests for this. I think they cover pretty much everything, and it's a fairly simple unit test.

## Limitations

Currently the following float values are not parsed:
- `inf`, `-inf`, `+infinity`, `NaN`
- `2.5E10`, `2.5e10`, `2.5E-10`

For my use case this is perfectly fine but other developers might want to support these values

### DIFF
--- a/crates/bevy_ui/src/geometry.rs
+++ b/crates/bevy_ui/src/geometry.rs
@@ -86,7 +86,7 @@ impl core::str::FromStr for Val {
 
         let Some(end_of_number) = s
             .bytes()
-            .position(|c| !(b'0'..=b'9').contains(&c) && c != b'.' && c != b'-')
+            .position(|c| !(c.is_ascii_digit() || c == b'.' || c == b'-' || c == b'+'))
         else {
             return Err(ValParseError::UnitMissing);
         };
--- a/crates/bevy_ui/src/geometry.rs
+++ b/crates/bevy_ui/src/geometry.rs
@@ -55,9 +55,10 @@ pub enum Val {
     VMax(f32),
 }
 
-#[derive(Debug, Error)]
+#[derive(Debug, Error, PartialEq, Eq)]
 pub enum ValParseError {
     UnitMissing,
+    ValueMissing,
     InvalidValue,
     InvalidUnit,
 }
@@ -66,6 +67,7 @@ impl std::fmt::Display for ValParseError {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         match self {
             ValParseError::UnitMissing => write!(f, "unit missing"),
+            ValParseError::ValueMissing => write!(f, "value missing"),
             ValParseError::InvalidValue => write!(f, "invalid value"),
             ValParseError::InvalidUnit => write!(f, "invalid unit"),
         }
@@ -88,6 +90,10 @@ impl std::str::FromStr for Val {
         else {
             return Err(ValParseError::UnitMissing);
         };
+
+        if end_of_number == 0 {
+            return Err(ValParseError::ValueMissing);
+        }
 
         let (value, unit) = s.split_at(end_of_number);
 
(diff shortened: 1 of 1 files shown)


crates/bevy_ui/src/geometry.rs
```diff
@@ -45,6 +55,70 @@ pub enum Val {
...
+            ValParseError::InvalidValue => write!(f, "invalid value"),
+            ValParseError::InvalidUnit => write!(f, "invalid unit"),
+        }
+    }
+}
+
+impl core::str::FromStr for Val {
+    type Err = ValParseError;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
```
bushrat011899, line 80: Would love to see this pulled out as a `const` function if it's compatible. Not a blocker if it's not a simple change though.

bushrat011899 (approved):
A sensible addition IMO. I would expect many `bsn`-like alternatives would have a similar implementation anyway, so it might as well be first-party. I would like this to be const though, just for that extra utility.

crates/bevy_ui/src/geometry.rs
VictorKoenders: I checked and every single function that's being called is non-const :laughing:

VictorKoenders: Specifically: `str::trim`, `str::bytes()`, `iterator::position`, `str::split_at`, `?`, `f32::parse`

bushrat011899: Damn, thanks for clarifying!

crates/bevy_ui/src/geometry.rs
```diff
@@ -45,6 +55,70 @@ pub enum Val {
...
+        let s = s.trim();
+
+        if s.eq_ignore_ascii_case("auto") {
+            return Ok(Val::Auto);
+        }
+
+        let Some(end_of_number) = s
```
mweatherley, line 87: Only small nit with this PR is that this doesn't accommodate all of rust's valid `f32` parsing inputs; I don't think people will often want to make values of `1e5 px` or whatever, but it's possible that values like that may show up in automatically-formatted float output, for example.

mweatherley (approved):

crates/bevy_ui/src/geometry.rs
VictorKoenders: This doesn't support: `NaN`, `inf`, `infinity`, `+inf`, `1E4` and `1e5`. I think most developers won't need this, and it can be added later

mweatherley: Indeed. That's why I approved it :)

BenjaminBrienen (approved):

crates/bevy_ui/src/geometry.rs
```diff
@@ -45,6 +55,70 @@ pub enum Val {
...
+            Ok(Val::Vw(value))
+        } else if unit.eq_ignore_ascii_case("vh") {
+            Ok(Val::Vh(value))
+        } else if unit.eq_ignore_ascii_case("vmin") {
+            Ok(Val::VMin(value))
+        } else if unit.eq_ignore_ascii_case("vmax") {
+            Ok(Val::VMax(value))
+        } else {
+            Err(ValParseError::InvalidUnit)
+        }
```
ickshonpe, line 118: The long `else if` chain is a bit ugly, is there a neater construction we could use?

Maybe something like:
```suggestion
        const VALS: &[(&str, fn(f32) -> Val)] = &[
            ("px", Val::Px),
            ("%", Val::Percent),
            ("vw", Val::Vw),
            ("vh", Val::Vh),
            ("vmin", Val::VMin),
            ("vmax", Val::VMax),
        ];

        for (name, val) in VALS {
            if unit.eq_ignore_ascii_case(name) {
                return Ok(val(value));
            }
        }

        Err(ValParseError::InvalidUnit)
```

Or less nice but non-looping:
```
        let val = if unit.eq_ignore_ascii_case("px") {
            Val::Px
        } else if unit.eq_ignore_ascii_case("%") {
            Val::Percent
        } else if unit.eq_ignore_ascii_case("vw") {
            Val::Vw
        } else if unit.eq_ignore_ascii_case("vh") {
            Val::Vh
        } else if unit.eq_ignore_ascii_case("vmin") {
            Val::VMin
        } else if unit.eq_ignore_ascii_case("vmax") {
            Val::VMax
        } else {
            return Err(ValParseError::InvalidUnit);
        };

        Ok(val(value))
```

This isn't that important though, and otherwise everything looks good.

ickshonpe (approved):

crates/bevy_ui/src/geometry.rs
BenjaminBrienen: I think it's fine how it is. Your suggestion might be nicer if it didn't require a loop, but the way it is seems the most straightforward and readable.

PR #16930, opened 2024-12-22
Refactor non-core Curve methods into extension traits

mweatherley:
# Objective

The way `Curve` presently achieves dyn-compatibility involves shoving `Self: Sized` bounds on a bunch of methods to forbid them from appearing in vtables. (This is called *explicit non-dispatchability*.) The `Curve` trait probably also just has way too many methods on its own. 

In the past, using extension traits instead to achieve similar functionality has been discussed. The upshot is that this would allow the "core" of the curve trait, on which all the automatic methods rely, to live in a very simple dyn-compatible trait, while other functionality is implemented by extensions. For instance, `dyn Curve<T>` cannot use the `Sized` methods, but `Box<dyn Curve<T>>` is `Sized`, hence would automatically implement the extension trait, containing the methods which are currently non-dispatchable.

Other motivations for this include modularity and code organization: the `Curve` trait itself has grown quite large with the addition of numerous adaptors, and refactoring it to demonstrate the separation of functionality that is already present makes a lot of sense. Furthermore, resampling behavior in particular is dependent on special traits that may be mimicked or analogized in user-space, and creating extension traits to achieve similar behavior in user-space is something we ought to encourage by example.

## Solution

`Curve` now contains only `domain` and the `sample` methods. 

`CurveExt` has been created, and it contains all adaptors, along with the other sampling convenience methods (`samples`, `sample_iter`, etc.). It is implemented for all `C` where `C: Curve<T> + Sized`.

`CurveResampleExt` has been created, and it contains all resampling methods. It is implemented for all `C` where `C: Curve<T> + ?Sized`. 

## Testing

It compiles and `cargo doc` succeeds.

---

## Future work

- Consider writing extension traits for resampling curves in related domains (e.g. resampling for `Curve<T>` where `T: Animatable` into an `AnimatableKeyframeCurve`). 
- `CurveExt` might be further broken down to separate the adaptor and sampling methods.

---

## Migration Guide

`Curve` has been refactored so that much of its functionality is now in extension traits. Adaptors such as `map`, `reparametrize`, `reverse`, and so on now require importing `CurveExt`, while the resampling methods `resample_*` require importing `CurveResampleExt`. Both of these new traits are exported through `bevy::math::curve` and through `bevy::math::prelude`.

### DIFF
--- a/crates/bevy_math/src/curve/adaptors.rs
+++ b/crates/bevy_math/src/curve/adaptors.rs
@@ -18,6 +18,10 @@ mod paths {
     pub(super) const THIS_CRATE: &str = "bevy_math";
 }
 
+// Import `CurveExt` just for doc links.
+#[expect(unused)]
+use super::CurveExt;
+
 // NOTE ON REFLECTION:
 //
 // Function members of structs pose an obstacle for reflection, because they don't implement
@@ -173,7 +177,7 @@ where
 }
 
 /// A curve whose samples are defined by mapping samples from another curve through a
-/// given function. Curves of this type are produced by [`Curve::map`].
+/// given function. Curves of this type are produced by [`CurveExt::map`].
 #[derive(Clone)]
 #[cfg_attr(feature = "serialize", derive(serde::Serialize, serde::Deserialize))]
 #[cfg_attr(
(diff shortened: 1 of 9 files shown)


crates/bevy_math/src/curve/adaptors.rs
```diff
@@ -18,6 +18,10 @@ mod paths {
     pub(super) const THIS_CRATE: &str = "bevy_math";
 }
 
+// Import `CurveExt` just for doc links.
```
BenjaminBrienen, line 21: Nit: I recommend using `, reason = "" ` instead of a comment.

BenjaminBrienen (approved):

crates/bevy_math/src/curve/adaptors.rs
mweatherley: Good point!

IQuick143 (approved):

PR #16931, opened 2024-12-22
Use `variadics_please` to implement `StableInterpolate` on tuples.

mweatherley:
# Objective

Now that `variadics_please` has a 1.1 release, we can re-implement the original solution. 

## Solution

Copy-paste the code from the [original PR](https://github.com/bevyengine/bevy/pull/15931) branch :)


BenjaminBrienen (approved):
Double check that 1..11 is correct.

PR #12827, opened 2024-04-01
Consolidate Render(Ui)Materials(2d) into RenderAssets

superdump:
# Objective

- Replace `RenderMaterials` / `RenderMaterials2d` / `RenderUiMaterials` with `RenderAssets` to enable implementing changes to one thing, `RenderAssets`, that applies to all use cases rather than duplicating changes everywhere for multiple things that should be one thing.
- Adopts #8149 

## Solution

- Make RenderAsset generic over the destination type rather than the source type as in #8149 
- Use `RenderAssets<PreparedMaterial<M>>` etc for render materials

---

## Changelog

- Changed:
  - The `RenderAsset` trait is now implemented on the destination type. Its `SourceAsset` associated type refers to the type of the source asset.
  - `RenderMaterials`, `RenderMaterials2d`, and `RenderUiMaterials` have been replaced by `RenderAssets<PreparedMaterial<M>>` and similar.

## Migration Guide

- `RenderAsset` is now implemented for the destination type rather that the source asset type. The source asset type is now the `RenderAsset` trait's `SourceAsset` associated type.

PR #16937, opened 2024-12-22
Remove unnecessary cast in DynamicTextureAtlasBuilder

mgi388:
# Summary 

- I started experimenting if `TextureAtlas` and friends can be moved to `bevy_image`. See [Discord](https://discord.com/channels/691052431525675048/692572690833473578/1320176054911897642) thread.
- While doing that, and moving `DynamicTextureAtlasBuilder` to `bevy_image`, it revealed that `DynamicTextureAtlasBuilder` depends on `bevy_render::GpuImage`, but we can't have `bevy_image` depend on `bevy_render`.
- The reason for the dependency is an assertion introduced in [this PR](https://github.com/bevyengine/bevy/pull/12827/files?show-viewed-files=true&file-filters%5B%5D=#diff-d9afd2170466f4aae340b244bdaa2a80aef58e979268c003878ca6c95860eb37R59).
- [It doesn't seem like there was a specific reason for that change](https://discord.com/channels/691052431525675048/743663924229963868/1320506862067650580), so should be safe to change it.
- So instead of the cast, just look up `asset_usage` directly on the concrete `Image` type.
- Also update the message which referred to a non-existent variable `atlas_texture_handle` (it was renamed during a subsequent refactor PR).

# Testing

- Checked on Discord if there was any known reason this had to stay like this.
- CI builds it.

### DIFF
--- a/crates/bevy_sprite/src/dynamic_texture_atlas_builder.rs
+++ b/crates/bevy_sprite/src/dynamic_texture_atlas_builder.rs
@@ -1,10 +1,7 @@
 use crate::TextureAtlasLayout;
 use bevy_image::{Image, TextureFormatPixelInfo};
 use bevy_math::{URect, UVec2};
-use bevy_render::{
-    render_asset::{RenderAsset, RenderAssetUsages},
-    texture::GpuImage,
-};
+use bevy_render::render_asset::RenderAssetUsages;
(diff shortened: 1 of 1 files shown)


kristoff3r (approved):
Seems reasonable

BenjaminBrienen (approved):

mockersf (approved):

PR #16940, opened 2024-12-23
Add BRP method to mutate a component

purplg:
# Objective

Add a method to mutate components with BRP.

Currently the only way to modify a component on an entity with BRP is to insert a new one with the new values. This isn't ideal for several reasons, one reason being that the client has to know what all the fields are of the component and stay in sync with the server.

## Solution

Add a new BRP method called `bevy/mutate_component` to mutate a single field in a component on an entity.

## Testing

Tested on a simple scene on all `Transform`, `Name`, and a custom component.

---

## Showcase

Example JSON-RPC request to change the `Name` of an entity to "New name!"

```json
{
    "jsonrpc": "2.0",
    "id": 0,
    "method": "bevy/mutate_component",
    "params": {
        "entity": 4294967308,
        "component": "bevy_ecs::name::Name",
        "path": ".name",
        "value": "New name!"
    }
}
```

Or setting the X translation to 10.0 on a Transform:

```json
{
    "jsonrpc": "2.0",
    "id": 0,
    "method": "bevy/mutate_component",
    "params": {
        "entity": 4294967308,
        "component": "bevy_transform::components::transform::Transform",
        "path": ".translation.x",
        "value": 10.0
    }
}
```

Clip of my Emacs BRP package using this method:

https://github.com/user-attachments/assets/a786b245-5c20-4189-859f-2261c5086a68

### DIFF
--- a/crates/bevy_remote/src/builtin_methods.rs
+++ b/crates/bevy_remote/src/builtin_methods.rs
@@ -736,8 +736,7 @@ pub fn process_remote_mutate_component_request(
     let app_type_registry = world.resource::<AppTypeRegistry>().clone();
     let type_registry = app_type_registry.read();
 
-    // Get the type of the component to be mutated.
-    // E.g. `bevy_transform::components::transform::Transform`
+    // Get the fully-qualified type names of the component to be mutated.
     let component_type: &TypeRegistration = type_registry
         .get_with_type_path(&component)
         .ok_or_else(|| {
--- a/crates/bevy_remote/src/builtin_methods.rs
+++ b/crates/bevy_remote/src/builtin_methods.rs
@@ -205,13 +205,20 @@ pub struct BrpListParams {
 /// The server responds with a null.
 #[derive(Debug, Serialize, Deserialize, Clone, PartialEq)]
 pub struct BrpMutateParams {
-    /// TODO
+    /// The entity of the component to mutate.
     pub entity: Entity,
-    /// TODO
+
+    /// The [full path] of component to mutate.
+    ///
+    /// [full path]: bevy_reflect::TypePath::type_path
     pub component: String,
-    /// TODO
+
+    /// The [path] of the field within the component.
+    ///
+    /// [path]: bevy_reflect::GetPath
     pub path: String,
-    /// TODO
+
+    /// The value to insert at `path`.
     pub value: Value,
 }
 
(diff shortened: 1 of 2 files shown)


crates/bevy_remote/src/builtin_methods.rs
```diff
@@ -48,6 +48,9 @@ pub const BRP_REPARENT_METHOD: &str = "bevy/reparent";
 /// The method path for a `bevy/list` request.
 pub const BRP_LIST_METHOD: &str = "bevy/list";
 
+/// The method path for a `bevy/reparent` request.
+pub const BRP_MUTATE_COMPONENT_METHOD: &str = "bevy/mutate_component";
```
purplg, line 52: I'm not sure what the convention is for multiple words in a method name is. Should the method be named differently? I'm trying to consider future methods for modifying other things like Resource's. 

I had a few other ideas for methods:

- `bevy/modify/component`
- `bevy/mutate/component`
- `bevy/component_mut` *this feels too Rusty in this context*

crates/bevy_remote/src/builtin_methods.rs
```diff
@@ -657,6 +682,71 @@ pub fn process_remote_insert_request(
...
 }
 
+/// Handles a `bevy/mutate_component` request coming from a client.
+///
+/// This method allows you to mutate a single field inside an Entity's
+/// component.
+pub fn process_remote_mutate_component_request(
+    In(params): In<Option<Value>>,
+    world: &mut World,
+) -> BrpResult {
```
purplg, line 692: I feel like the method handler could probably be done better. Any pointers here would be appreciated.

cBournhonesque (approved):
This generally LGTM.
It's a shame that we don't have unit tests for BRP

purplg:
> It's a shame that we don't have unit tests for BRP

I agree. I did look at writing tests for this, but I didn't see an existing way to do this.

LiamGallagher737 (changes requested):
It would be useful to link to the following docs in the crate level docs for this method to show how the path field should be structured.

https://docs.rs/bevy/latest/bevy/prelude/trait.GetPath.html#syntax

purplg:
Good point on that doc link. Added.

LiamGallagher737 (approved):

BenjaminBrienen:
CI still failing

purplg:
I think the msrv one just needed a rebase.

Not sure why the CI check required me to backtick `bevy/mutate_component`. None of the other ones required it, heh.

Anyway, it looks like it all passes now. Thanks

mockersf (approved):

PR #16942, opened 2024-12-23
Key render phases off the main world view entity, not the render world view entity.

pcwalton:
We won't be able to retain render phases from frame to frame if the keys are unstable. It's not as simple as simply keying off the main world entity, however, because some main world entities extract to multiple render world entities. For example, directional lights extract to multiple shadow cascades, and point lights extract to one view per cubemap face. Therefore, we key off a new type, `RetainedViewEntity`, which contains the main entity plus a *subview ID*.

This is part of the preparation for retained bins.


ickshonpe (commented):
Nothing gets rendered for the UI, but I'm not very familiar with the camera extraction internals and I couldn't quite unravel the problem. 

Very loosely, from what I can tell: 
`extract_default_ui_camera_view` spawns a default camera view entity for each camera which has an `ExtractedView` with `subview_index: 1` and then creates a corresponding `TransparentUi` render phase. 

But the extracted uinodes are given the entity id of the non-ui camera `ExtractedView` entity so that in `queue_uinodes` the non-ui view with `subview_index: 0` is retrieved and then no `TransparentUi` phases are found because they were all keyed to `subview_index: 1`.

pcwalton:
This should fix the issue with the UI. The proper fix was a bit tricky and involved creating a new subgraph runner to make sure to run the UI rendering subgraph on the right render world entity.

crates/bevy_ui/src/render/mod.rs
```diff
@@ -250,6 +255,31 @@ impl ExtractedUiNodes {
...
+        &self,
+        graph: &mut RenderGraphContext,
+        _: &mut RenderContext<'w>,
+        world: &'w World,
+    ) -> Result<(), NodeRunError> {
+        // Fetch the UI view.
+        let Some(mut render_views) = world.try_query::<&DefaultCameraView>() else {
+            return Ok(());
+        };
+        let Some(default_camera_view) = render_views.iter(world).next() else {
```
ickshonpe, line 273: The `multiple_windows` example doesn't render the label for the second window. It looks like this just needs to look up the correct view using the `graph.view_entity()`.
```suggestion
        let Ok(default_camera_view) = render_views.get(world, graph.view_entity()) else {
```

ickshonpe (changes requested):
Multi-window UI isn't working, I think I found the problem though.

Otherwise everything looks good now, the different shaders etc all seem to be working correctly.

IceSentry:
It looks like there's still some issues with UI based on the example runner https://pixel-eagle.com/project/B25A040A-A980-4602-B90C-D480AB84076D?filter=PR-16942

ickshonpe:
> It looks like there's still some issues with UI based on the example runner https://pixel-eagle.com/project/B25A040A-A980-4602-B90C-D480AB84076D?filter=PR-16942

Ah yeah you're right. Shadows and texture slices still don't work, I must have ran those examples on the wrong branch by mistake.

ickshonpe:
The extraction and queue functions just needed to be updated. I put in a pr on this pr with the needed changes.

pcwalton:
Cherry picked over, thanks!

ickshonpe (approved):
The UI rendering is all fixed now. There is a slight performance regression from the extra indirection but I don't think it's a problem with this PR really, more of a sign that we need to rethink how the UI handle its camera views.

crates/bevy_ui/src/render/mod.rs
```diff
@@ -532,16 +562,39 @@ const UI_CAMERA_FAR: f32 = 1000.0;
 // TODO: Evaluate if we still need this.
 const UI_CAMERA_TRANSFORM_OFFSET: f32 = -0.1;
 
+/// The ID of the subview associated with a camera on which UI is to be drawn.
+///
+/// When UI is present, 3D cameras extract to two views: a 3D one and a UI one.
+/// The main 3D camera gets subview 0, and the corresponding UI camera gets this
+/// subview, 1.
```
IceSentry, line 569: Why does this comment only mentions 3d cameras? Isn't this also true for 2d?

IceSentry (approved):

crates/bevy_ui/src/render/mod.rs
ickshonpe: Yes every active 2d or 3d camera gets an extracted UI view. Doesn't matter if there are any UI entities present, only that the UI rendering plugin is enabled.

pcwalton: Updated the comment.

alice-i-cecile:
Let me know when this has passed an example check run and merge conflicts are resolved please :)

pcwalton:
@alice-i-cecile I kicked off a run. https://pixel-eagle.com/project/B25A040A-A980-4602-B90C-D480AB84076D?filter=PR-16942 Looks like the failures are known intermittents, so we should be good to go.

PR #16944, opened 2024-12-23
show these 'fully qualified paths' for bevy_remote's rpc

alphastrata:
# Objective

It is not obvious that one would know these:
```json
{
  "id": 1,
  "jsonrpc": "2.0",
  "result": [
    "bevy_animation::AnimationPlayer",
    "bevy_animation::AnimationTarget",
    "bevy_animation::graph::AnimationGraphHandle",
    "bevy_animation::transition::AnimationTransitions",
    "bevy_audio::audio::PlaybackSettings",
    "bevy_audio::audio::SpatialListener",
    "bevy_core_pipeline::bloom::settings::Bloom",
    "bevy_core_pipeline::contrast_adaptive_sharpening::ContrastAdaptiveSharpening",
   **... snipping for brevity ...**
    "bevy_ui::ui_node::Node",
    "bevy_ui::ui_node::Outline",
    "bevy_ui::ui_node::ScrollPosition",
    "bevy_ui::ui_node::TargetCamera",
    "bevy_ui::ui_node::UiAntiAlias",
    "bevy_ui::ui_node::ZIndex",
    "bevy_ui::widget::button::Button",
    "bevy_window::monitor::Monitor",
    "bevy_window::window::PrimaryWindow",
    "bevy_window::window::Window",
    "bevy_winit::cursor::CursorIcon",
    "server::Cube"
  ]
}
```
Especially if you for example, are reading the GH examples because:
image
If you for example expand these things, due to the number of places bevy re-exports things you'll find it difficult to find the true path of something.
i.e you'd probably be forgiven for writing a query (using the `client.rs` example):
```sh
$ cargo run --example client --  bevy_pbr::mesh_material::MeshMaterial3d | jq
{
  "error": {
    "code": -23402,
    "message": "Unknown component type: `bevy_pbr::mesh_material::MeshMaterial3d`"
  },
  "id": 1,
  "jsonrpc": "2.0"
}
```
which is where https://github.com/bevyengine/bevy/blob/8d9a00f5483b2ba28f40081759bd3d6f8b8a686d/crates/bevy_pbr/src/mesh_material.rs#L41 would lead you to believe it is...

I've worked with bevy a lot, so it's no issue for me, but for others... ?

## Solution

- Add some more docs, including a sample request (`json` and `rust`)

## Testing

N/A

---

## Showcase
N/A


examples/remote/list_all_paths.rs
```diff
@@ -0,0 +1,29 @@
...
+    builtin_methods::BRP_LIST_METHOD, http::DEFAULT_ADDR, http::DEFAULT_PORT, BrpRequest,
+};
+
+/// The application entry point.
+fn main() -> AnyhowResult<()> {
+    // Create the URL. We're going to need it to issue the HTTP request.
+    let host_part = format!("{}:{}", DEFAULT_ADDR, DEFAULT_PORT);
+    let url = format!("http://{}/", host_part);
+
+    // If you pass no params, you'll be able to see all the optinos for arguments you can pass to the `./client.rs` remote example.
```
BenjaminBrienen, line 14:
```suggestion
    // If you pass no params, you'll be able to see all the options for arguments you can pass to the `./client.rs` remote example.
```

BenjaminBrienen (changes requested):

BenjaminBrienen (approved):

alphastrata:
👍🏽

alice-i-cecile (commented):
Hmm. I can see why this would be useful, but I think it's probably better suited as a doc test in the module docs.

alphastrata:
> Hmm. I can see why this would be useful, but I think it's probably better suited as a doc test in the module docs.

Am I correct in thinking that if they are here [in examples] they're scraped for the docs, but not the other way around?

Happy to move them though. @alice-i-cecile whatever you want to do is fine with me.

alice-i-cecile:
Examples getting scraped is a good argument, but Bevy's examples are already rather numerous. I'd like to reserve them for teaching things that *must* be interactive, so let's move this information into the docs.

BenjaminBrienen:
Examples being visual examples only will help delineate that and it makes sense, since examples are shown on the website as a "look at this".

alice-i-cecile:
> Although, many crates don't have anything to do with visuals and might benefit from bigger examples that aren't well suited to a doc test.

Yeah, things that demonstrate end-to-end usage patterns can be valuable to teach in examples too! But small snippets / notes like this should just go in the docs.

alphastrata:
> > Although, many crates don't have anything to do with visuals and might benefit from bigger examples that aren't well suited to a doc test.
> 
> Yeah, things that demonstrate end-to-end usage patterns can be valuable to teach in examples too! But small snippets / notes like this should just go in the docs.

I will move them over the weekend. --standby

alphastrata:
We thinking something like this:
image
i.e have them on the BrpRequest docs? or something more... prominent?

alice-i-cecile:
On the BrpRequest docs makes sense to me :)

alice-i-cecile (approved):

PR #16947, opened 2024-12-23
METIS-based meshlet generation

JMS55:
# Objective
Improve DAG building for virtual geometry

## Solution

- Use METIS to group triangles into meshlets which lets us minimize locked vertices which improves simplification, instead of using meshopt which prioritizes culling efficiency. Also some other minor tweaks.
- Currently most meshlets have 126 triangles, and not 128. Fixing this might involve calling METIS recursively ourselves to manually bisect the graph, not sure. Not going to attempt to fix this in this PR.

## Testing

  - Tested on bunny.glb and cliff.glb
  - No
  - Download the new bunny asset, run the meshlet example.

---

## Showcase

New 
image

Old
image


crates/bevy_pbr/src/meshlet/from_mesh.rs
```diff
@@ -67,12 +67,29 @@ impl MeshletMesh {
         // Validate mesh format
         let indices = validate_input_mesh(mesh)?;
 
-        // Split the mesh into an initial list of meshlets (LOD 0)
+        // Get meshet vertices
```
IceSentry, line 70:
```suggestion
        // Get meshlet vertices
```

IceSentry (approved):

crates/bevy_pbr/src/meshlet/from_mesh.rs
```diff
@@ -243,8 +252,103 @@ fn validate_input_mesh(mesh: &Mesh) -> Result<Cow<'_, [u32]>, MeshToMeshletMeshC
...
+        }
+    }
+    xadj.push(adjncy.len() as i32);
+
+    let mut options = [-1; metis::NOPTIONS];
+    options[metis::option::Seed::INDEX] = 17;
+    options[metis::option::UFactor::INDEX] = 1; // Important that there's very little imbalance between partitions
+
+    let mut meshlet_per_triangle = vec![0; triangle_count];
+    let partition_count = triangle_count.div_ceil(126); // Need to undershoot to prevent METIS from going over 128 triangles per meshlet
```
atlv24, line 316: by how much does it overshoot the partitions it makes if you set this to 128?

atlv24 (approved):

crates/bevy_pbr/src/meshlet/from_mesh.rs
JMS55: 1-2 triangles, which then leads to tiny meshlets with 1 triangle. Need to undershoot unfortunately :(

PR #16950, opened 2024-12-24
don't trigger drag events if there's no movement

### LINKED ISSUE #16571: Drag picking event fires on pointer up
## Bevy version

0.15.0

## What went wrong

`Trigger<Pointer<Drag>>` observer fires on `Pointer<Up>` even without drag movement. `Drag` is supposedly only supposed to fire if the cursor moved this tick.

## What you did

Reproduced:

```rust
use bevy::prelude::*;

fn main() {
    App::new()
        .add_plugins(DefaultPlugins)
        .add_systems(Startup, setup_scene)
        .run();
}

fn setup_scene(
    mut commands: Commands,
) {
    commands
        .spawn(Text::new("Click Me or Drag Me"))
        .observe(
            |_: Trigger<Pointer<Down>>| {
                println!("down");
            },
        )
        .observe(
            |event: Trigger<Pointer<Drag>>| {
                println!("drag {:?}", event.event());
            },
        );

    commands.spawn(Camera2d::default());
}
```

mockersf:
# Objective

- Fixes #16571

## Solution

- When position delta is zero, don't trigger `Drag` or `DragOver` events

## Testing

- tested with the code from the issue

### DIFF
--- a/crates/bevy_picking/src/events.rs
+++ b/crates/bevy_picking/src/events.rs
@@ -692,6 +692,10 @@ pub fn pointer_events(
 
                     // Emit Drag events to the entities we are dragging
                     for (drag_target, drag) in state.dragging.iter_mut() {
+                        let delta = location.position - drag.latest_pos;
+                        if delta == Vec2::ZERO {
+                            continue; // No need to emit a Drag event if there is no movement
+                        }
                         let drag_event = Pointer::new(
                             pointer_id,
                             location.clone(),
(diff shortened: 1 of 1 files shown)


BenjaminBrienen (approved):

alice-i-cecile (approved):
Good fix. I think that an epsilon might be warranted, but let's ship this very simple change and see if it resolves things.

PR #16951, opened 2024-12-24
Move `futures.rs`, `ConditionalSend` and `BoxedFuture` types to `bevy_tasks`

mnmaita:
# Objective

- Related to https://github.com/bevyengine/bevy/issues/11478

## Solution

- Moved `futures.rs`, `ConditionalSend` `ConditionalSendFuture` and `BoxedFuture` from `bevy_utils` to `bevy_tasks`.

## Testing

- CI checks

## Migration Guide

- Several modules were moved from `bevy_utils` into `bevy_tasks`:
  - Replace `bevy_utils::futures` imports with `bevy_tasks::futures`.
  - Replace `bevy_utils::ConditionalSend` with `bevy_tasks::ConditionalSend`.
  - Replace `bevy_utils::ConditionalSendFuture` with `bevy_tasks::ConditionalSendFuture`.
  - Replace `bevy_utils::BoxedFuture` with `bevy_tasks::BoxedFuture`.

### DIFF
--- a/crates/bevy_tasks/Cargo.toml
+++ b/crates/bevy_tasks/Cargo.toml
@@ -10,9 +10,7 @@ keywords = ["bevy"]
 
 [features]
 default = ["std", "async_executor"]
-alloc = []
 std = [
-  "alloc",
   "futures-lite/std",
   "async-task/std",
   "spin/std",
--- a/crates/bevy_tasks/Cargo.toml
+++ b/crates/bevy_tasks/Cargo.toml
@@ -10,7 +10,9 @@ keywords = ["bevy"]
 
 [features]
 default = ["std", "async_executor"]
+alloc = []
 std = [
+  "alloc",
   "futures-lite/std",
   "async-task/std",
   "spin/std",
--- a/crates/bevy_tasks/src/lib.rs
+++ b/crates/bevy_tasks/src/lib.rs
@@ -8,6 +8,37 @@
 
 extern crate alloc;
 
+#[cfg(not(target_arch = "wasm32"))]
+mod conditional_send {
+    /// Use [`ConditionalSend`] to mark an optional Send trait bound. Useful as on certain platforms (eg. Wasm),
+    /// futures aren't Send.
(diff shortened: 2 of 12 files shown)


mnmaita:
Wondering if the added `alloc` feature is the proper way to do it or if there's a different way.

crates/bevy_tasks/Cargo.toml
```diff
@@ -10,7 +10,9 @@ keywords = ["bevy"]
 
 [features]
 default = ["std", "async_executor"]
+alloc = []
```
bushrat011899, line 13: Within `bevy_tasks`, `alloc` is mandatory, so this feature isn't required.

crates/bevy_tasks/src/lib.rs
```diff
@@ -8,6 +8,37 @@
...
+}
+
+pub use conditional_send::*;
+
+/// Use [`ConditionalSendFuture`] for a future with an optional Send trait bound, as on certain platforms (eg. Wasm),
+/// futures aren't Send.
+pub trait ConditionalSendFuture: core::future::Future + ConditionalSend {}
+impl<T: core::future::Future + ConditionalSend> ConditionalSendFuture for T {}
+
+#[cfg(feature = "alloc")]
```
bushrat011899, line 33: It should be sufficient to just remove these `cfg(...)` attributes in order to remove the `alloc` feature. This will also need to be done in `futures.rs`.

bushrat011899 (changes requested):
Totally agree with this change, just need to remove the `alloc` feature from `bevy_tasks` since we can safely assume access to an allocator.

crates/bevy_tasks/src/lib.rs
mnmaita: Thanks for chiming in on this one, I was unsure what to do with it.

Removed the feature and the cfgs. There were no instances of this in `futures.rs` FWIW.

bushrat011899 (approved):

BenjaminBrienen (approved):

PR #16953, opened 2024-12-24
Hide `SubApps` from the documentation

LikeLakers2:
# Objective
`SubApps` is visible within the documentation for `bevy_app`. However, no way of accessing the `SubApps` field in `App` is currently available.

## Solution
Hides `SubApps` from the documentation.

The other solution is to expose access to `SubApps` in `Apps`, which I  submitted as a PR at <https://github.com/bevyengine/bevy/pull/16952>.

## Testing
Because of the simplicity of the changes, I only tested by compiling `bevy_app` - which compiled successfully.


alice-i-cecile:
Closing in favor of #16952.

PR #16958, opened 2024-12-24
Fix path checking for FileWatcher for virtual workspace projects

### LINKED ISSUE #16879: `file_watcher` feature cannot find assets folder in workspace
## Bevy version

v0.15.0  #3f2baf8ec

Observed on both Arch Linux and macOS.

## What you did

A basic workspace setup where the actual bevy App launches from a subcrate, like the following Cargo.toml setup:

```
[workspace]
resolver = "2"
members = [
    "library_a",
    "main_bevy_game",
]
```

The assets folder for the game resides in the `main_bevy_game` root. There are no other assets folders in the workspace.
`main_bevy_game` enables `file_watcher` feature, and then in `main.rs`:

```
    App::new()
        .add_plugins(DefaultPlugins.set(AssetPlugin {
            watch_for_changes_override: Some(true),
            ..Default::default()
        }));
```

## What went wrong

On launch, bevy issues the following warning: 

`WARN bevy_asset::io::source: Skip creating file watcher because path "assets" does not exist.`

Creating a folder at the root of the workspace will suppress this warning, but file changes are not recognized.

Moving the assets folder to the root of the workspace causes a panic:

`Failed to create file watcher from path "assets", Error { kind: PathNotFound, paths: ["/Users/notarealuser/workspace/main_bevy_game/assets"] }`

(Path altered for privacy)

## Additional information

I believe this is from using a raw `std::path::PathBuf` to test if the assets folder exists whereas in most other cases it gets built out.
I have a quick one-line fix:

`bevy_asset` crate @ `bevy_assets/src/io/source.rs`

 line 534:
`                let path = super::file::get_base_path().join(path.clone());`

This builds the path as it does in `file::FileWatcher::new()` and seems to make everything work as expected.

This might not be the appropriate fix, or perhaps my project structure is not as intended. I'm not sure, so I haven't opened a PR for this one line change.

super-saturn:
# Objective

Fixes #16879

## Solution

Moved the construction of the root path of the assets folder out of `FileWatcher::new()` and into `source.rs`, as the path is checked there with `path.exists()` and fails in certain configurations eg., virtual workspaces.

## Testing

Applied fix to a private fork and tested against both standard project setups and virtual workspaces. Works without issue on both. Have tested under macOS and Arch Linux.

### DIFF
--- a/crates/bevy_asset/src/io/file/file_watcher.rs
+++ b/crates/bevy_asset/src/io/file/file_watcher.rs
@@ -30,7 +30,7 @@ impl FileWatcher {
         sender: Sender<AssetSourceEvent>,
         debounce_wait_time: Duration,
     ) -> Result<Self, notify::Error> {
-        let root = normalize_path(super::get_base_path().join(root).as_path());
+        let root = normalize_path(&path);
         let watcher = new_asset_event_debouncer(
             root.clone(),
             debounce_wait_time,
--- a/crates/bevy_asset/src/io/file/file_watcher.rs
+++ b/crates/bevy_asset/src/io/file/file_watcher.rs
@@ -26,13 +26,13 @@ pub struct FileWatcher {
 
 impl FileWatcher {
     pub fn new(
-        root: PathBuf,
+        path: PathBuf,
         sender: Sender<AssetSourceEvent>,
         debounce_wait_time: Duration,
     ) -> Result<Self, notify::Error> {
         let root = normalize_path(&path);
         let watcher = new_asset_event_debouncer(
-            root.clone(),
+            path.clone(),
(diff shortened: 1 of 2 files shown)


BenjaminBrienen (approved):

alice-i-cecile:
@super-saturn can you explain why you closed this so others trying to fix this bug better understand what's going on?

super-saturn:
@alice-i-cecile Apologies, this just has a small issue, I was merging these fixes in from my own cloned copy of bevy, and I was trying to be very clean about what I picked and this won't compile as-is. Triple-checking my work before I make a small change and re-open.

kristoff3r (approved):

PR #16964, opened 2024-12-25
Move the stage that acquires a new texture from the swap chain as late as possible in the rendering pipeline.

pcwalton:
Currently, we retrieve a new texture from the pipeline in `prepare_windows`, which executes during the `ManageViews` rendering phase. This is inefficient if we're GPU bound, because `ManageViews` happens before `queue_material_meshes` and
`batch_and_prepare_binned_render_phase` among other frequently-slow systems. Retrieving a new texture from the swap chain can block for considerable time, during which we don't want the CPU to be sitting idle. Additionally, we don't want the GPU to be sitting idle while the CPU works on `queue_material_meshes` and so forth.

This commit fixes the issue by moving the acquisition of a new swap chain texture late in the pipeline, to a new phase `PrepareWindows` that occurs right before render graph evaluation. Because render graph evaluation needs to bind the swap chain textures in order to issue drawing commands, this is the last possible moment where it's safe to do so.

This shows Bistro on `main` with shadows enabled before these changes. Note that `queue_material_meshes` and `batch_and_prepare_binned_render_phase` occur after `prepare_windows` and therefore the GPU is idle during this time.
Screenshot 2024-12-24 185330

This shows Bistro with these changes applied and shadows enabled. Observe that `queue_material_meshes` and `batch_and_prepare_binned_render_phase` are moved before `prepare_windows`, allowing for simultaneous GPU/CPU execution.
Screenshot 2024-12-24 190909

### DIFF
--- a/crates/bevy_core_pipeline/src/auto_exposure/mod.rs
+++ b/crates/bevy_core_pipeline/src/auto_exposure/mod.rs
@@ -73,7 +73,9 @@ impl Plugin for AutoExposurePlugin {
             .add_systems(
                 Render,
                 (
-                    prepare_buffers.in_set(RenderSet::Prepare),
+                    prepare_buffers
+                        .in_set(RenderSet::Prepare)
+                        .before(RenderSet::PrepareWindows),
                     queue_view_auto_exposure_pipelines.in_set(RenderSet::Queue),
                 ),
             )
--- a/crates/bevy_core_pipeline/src/bloom/mod.rs
+++ b/crates/bevy_core_pipeline/src/bloom/mod.rs
@@ -64,8 +64,12 @@ impl Plugin for BloomPlugin {
             .add_systems(
                 Render,
                 (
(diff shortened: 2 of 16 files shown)


JMS55:
Do we want this? Not sure how the synchronization is implemented in wgpu, but for DirectX12 I did this: https://github.com/JMS55/bevy_solari_directx/blob/main/src/swapchain.rs#L70-L76

Maybe @cwfitzgerald can weigh in on when the best time to block on the swapchain in wgpu is for good frame pacing / minimizing latency / maximizing FPS.

pcwalton:
I don't really understand why you'd want to not wait until the last possible moment. Right now the GPU is sitting idle for a long time and that makes no sense to me.

pcwalton:
To be clear, not having this definitely loses us FPS in Bistro.

pcwalton:
@Elabajaba did measurements and found that this has no effect, so closing.

PR #16968, opened 2024-12-25
Make `RawHandleWrapper` fields private to save users from themselves

### LINKED ISSUE #16683: create_surfaces is unsound
## Bevy version

0.15

## What went wrong

In [`create_surfaces`](https://github.com/bevyengine/bevy/blob/24c3bd5f005867aa483237071e3add4f13a43904/crates/bevy_render/src/view/window/mod.rs#L328), the underlying `create_surface_unsafe` is called under the assumption that any surface target inside of `ExtractedWindows` is guaranteed to be valid.

This isn't true, however. The fields on [`RawHandleWrapper`](https://github.com/bevyengine/bevy/blob/24c3bd5f005867aa483237071e3add4f13a43904/crates/bevy_window/src/raw_handle.rs#L52) are public, so any user could insert garbage data and cause the call to `create_surface_unsafe` to use invalid data.

## Propose solution

Make `RawHandleWrapper`'s fields private, replacing them with accessors and `unsafe` constructors.

Brezak:
# Objective

Fixes #16683

## Solution

Make all fields ine `RawHandleWrapper` private.

## Testing

- CI
- `cargo clippy`
- The lightmaps example
---

## Migration Guide

The `window_handle` and `dispay_handle` fields on `RawHandleWrapper` are no longer public. Use the newly added getters and setters to manipulate them instead.

### DIFF
--- a/crates/bevy_window/src/raw_handle.rs
+++ b/crates/bevy_window/src/raw_handle.rs
@@ -49,9 +49,9 @@ impl<W: 'static> Deref for WindowWrapper<W> {
 pub struct RawHandleWrapper {
     _window: Arc<dyn Any + Send + Sync>,
     /// Raw handle to a window.
-    pub window_handle: RawWindowHandle,
+    window_handle: RawWindowHandle,
     /// Raw handle to the display server.
-    pub display_handle: RawDisplayHandle,
+    display_handle: RawDisplayHandle,
 }
 
 impl RawHandleWrapper {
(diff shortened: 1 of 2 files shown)


crates/bevy_window/src/raw_handle.rs
```diff
@@ -75,6 +75,41 @@ impl RawHandleWrapper {
...
+
+    /// Gets the stored window handle.
+    pub fn get_window_handle(&self) -> RawWindowHandle {
+        self.window_handle
+    }
+
+    // The use of an explicit setter instead of a getter for a mutable reference is to limit the amount of time unsoundness can happen.
+    // If we handed out a mutable reference the user would have to maintain safety invariants throughout its lifetime. For consistency
+    // we also prefer to handout copies of the handles instead of immutable references.
+    /// Sets the window handle.
```
alice-i-cecile, line 87: This should be the first line of the docs here.

alice-i-cecile:
Simple docs issue then this LGTM. @spectria-limina, does this line up with what you were hoping for?

spectria-limina:
Yes, LGTM after a cursory look. Thanks!

alice-i-cecile (approved):

PR #16977, opened 2024-12-26
Added App::with_component_hooks 

### LINKED ISSUE #16921: make register_component_hooks method for App
currently, you can register component hooks only when you impl Componet or have mut world access.
if you want to register hooks when building your App you need to call app.world_mut().register_component_hooks() and then register them

add a `fn register_component_hooks<T>(&mut self, hook: ComponentHook)` to App so it is easy to register hooks when making an app

add a helper `fn register_component_hooks(&mut self)` to App
this makes it easier to find instead of having to know it exists on World that you can access from App

pin3-free:
# Objective

- Fixes #16921
- Right now, to add component hooks, you either have to implement them in the Component trait implementation, or use `app.world_mut().register_component_hooks()`, which is somewhat clunky. This PR aims to introduce a more concise way to add component hooks directly from the app interface

## Solution

- I could not find a good way to implement the interface proposed in the solution directly, since the user would need to not only pass the hook itself, but also specify which hook method (`on_add`, `on_insert`, etc.) they want to add the hook to. Plus there's the issue of fallible hook add methods.
- The proposed solution is similar to methods such as `with_children` on the `EntityCommands`, passing a closure that will mutate the target (in this case -- `&mut ComponentHooks`) and return the parent, allowing it to chain nicely into the multitude of calls usually associated with `App`.

## Testing

- I tested this by creating a new example called `with_component_hooks.rs` in the `ecs` section of examples. It works identically to the `component_hooks` example, but uses the new API
---

## Showcase

> Here's a snipped from the `component_hooks.rs` example and the comparison to the `with_component_hooks.rs` example

```rust
// === component_hooks.rs ===

fn main() {
    App::new()
        .add_plugins(DefaultPlugins)
        .add_systems(Startup, setup)
        .add_systems(Update, trigger_hooks)
        .init_resource::<MyComponentIndex>()
        .add_event::<MyEvent>()
        .run();
}

fn setup(world: &mut World) {
    world
        .register_component_hooks::<MyComponent>()
        .on_add(|mut world, entity, component_id| {
            let value = world.get::<MyComponent>(entity).unwrap().0;
            println!("Component: {component_id:?} added to: {entity:?} with value {value:?}");
            world
                .resource_mut::<MyComponentIndex>()
                .insert(value, entity);
            world.send_event(MyEvent);
        })
        .on_insert(|world, _, _| {
            println!("Current Index: {:?}", world.resource::<MyComponentIndex>());
        })
        .on_replace(|mut world, entity, _| {
            let value = world.get::<MyComponent>(entity).unwrap().0;
            world.resource_mut::<MyComponentIndex>().remove(&value);
        })
        .on_remove(|mut world, entity, component_id| {
            let value = world.get::<MyComponent>(entity).unwrap().0;
            println!("Component: {component_id:?} removed from: {entity:?} with value {value:?}");
            world.commands().entity(entity).despawn();
        });
}
```

```rust
// === with_component_hooks.rs ===

// The setup function is no longer needed in this instance
fn main() {
    App::new()
        .add_plugins(DefaultPlugins)
        .add_systems(Update, trigger_hooks)
        .with_component_hooks::<MyComponent>(|hooks| {
            hooks
                .on_add(|mut world, entity, component_id| {
                    let value = world.get::<MyComponent>(entity).unwrap().0;
                    println!(
                        "Component: {component_id:?} added to: {entity:?} with value {value:?}"
                    );
                    world
                        .resource_mut::<MyComponentIndex>()
                        .insert(value, entity);
                    world.send_event(MyEvent);
                })
                .on_insert(|world, _, _| {
                    println!("Current Index: {:?}", world.resource::<MyComponentIndex>());
                })
                .on_replace(|mut world, entity, _| {
                    let value = world.get::<MyComponent>(entity).unwrap().0;
                    world.resource_mut::<MyComponentIndex>().remove(&value);
                })
                .on_remove(|mut world, entity, component_id| {
                    let value = world.get::<MyComponent>(entity).unwrap().0;
                    println!(
                        "Component: {component_id:?} removed from: {entity:?} with value {value:?}"
                    );
                    world.commands().entity(entity).despawn();
                });
        })
        .init_resource::<MyComponentIndex>()
        .add_event::<MyEvent>()
        .run();
}

```

### DIFF
--- a/benches/Cargo.toml
+++ b/benches/Cargo.toml
@@ -7,6 +7,11 @@ license = "MIT OR Apache-2.0"
 # Do not automatically discover benchmarks, we specify them manually instead.
 autobenches = false
 
+[dependencies]
+# The primary crate that runs and analyzes our benchmarks. This is a regular dependency because the
+# `bench!` macro refers to it in its documentation.
+criterion = { version = "0.5.1", features = ["html_reports"] }
+
 [dev-dependencies]
 # Bevy crates
 bevy_app = { path = "../crates/bevy_app" }
@@ -22,7 +27,6 @@ bevy_tasks = { path = "../crates/bevy_tasks" }
 bevy_utils = { path = "../crates/bevy_utils" }
 
 # Other crates
-criterion = { version = "0.5.1", features = ["html_reports"] }
 glam = "0.29"
 rand = "0.8"
 rand_chacha = "0.3"
--- a/benches/benches/bevy_reflect/function.rs
+++ b/benches/benches/bevy_reflect/function.rs
@@ -1,14 +1,25 @@
-use bevy_reflect::func::{ArgList, IntoFunction, IntoFunctionMut, TypedFunction};
-use criterion::{criterion_group, BatchSize, Criterion};
+use core::hint::black_box;
 
-criterion_group!(benches, typed, into, call, overload, clone);
+use benches::bench;
+use bevy_reflect::func::{ArgList, IntoFunction, IntoFunctionMut, TypedFunction};
+use criterion::{criterion_group, BatchSize, BenchmarkId, Criterion};
+
+criterion_group!(
+    benches,
+    typed,
+    into,
+    call,
+    clone,
+    with_overload,
+    call_overload,
+);
 
 fn add(a: i32, b: i32) -> i32 {
     a + b
 }
 
 fn typed(c: &mut Criterion) {
-    c.benchmark_group("typed")
+    c.benchmark_group(bench!("typed"))
         .bench_function("function", |b| {
             b.iter(|| add.get_function_info());
         })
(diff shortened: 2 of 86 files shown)


BenjaminBrienen (approved):

alice-i-cecile:
No complaints about the implementation, but I don't think we should do this. There's already too many methods on `App` and too much API duplication. Medium-term, I'd like to push folks towards the app.world_mut pattern in general, and this PR is going in the opposite direction.

BenjaminBrienen:
Is there a compromise where we can allow for mutably "inspecting" the app's world from an App, which would allow for keeping the builder-like construction of an App without API duplication.

alice-i-cecile:
It feels like that method would make sense as `modify_world`? Overall I think that the App <-> World duplication / API mess around multiple worlds needs a real design to solve, and we should be nervous about just adding a dozen little fixes.

Issue #16718, opened 2024-12-08
Fallible systems need to report failures


alice-i-cecile:
As of https://github.com/bevyengine/bevy/pull/16589, systems which fail will just silently do nothing.

This is a fine choice to offer, but not a good default behavior since it masks real bugs in users games. Until we get a more sophisticated error handling mechanism in place, we should make this panic for now.

alice-i-cecile:
I've opened this issue to ensure that we don't *ship* the current behavior. The suggestion to panic is a reasonable stop-gap, and should have been done in the initial PR.

PR #16979, opened 2024-12-26
panic on system error

mockersf:
# Objective

- First step for #16718 
- #16589 introduced an api that can only ignore errors, which is risky

## Solution

- Panic instead of just ignoring the errors

## Testing

- Changed the `fallible_systems` example to return an error
```
Encountered an error in system `fallible_systems::setup`: TooManyVertices { subdivisions: 300, number_of_resulting_points: 906012 }
Encountered a panic in system `fallible_systems::setup`!
```


alice-i-cecile (approved):

alice-i-cecile:
This is a safer default for now, although I'd *really* like to make this configurable in time for 0.16.

BenjaminBrienen (approved):

PR #16985, opened 2024-12-26
Retain `RenderMaterialInstances` and `RenderMeshMaterialIds` from frame to frame.

pcwalton:
This commit makes Bevy use change detection to only update `RenderMaterialInstances` and `RenderMeshMaterialIds` when meshes have been added, changed, or removed. `extract_mesh_materials`, the system that extracts these, now follows the pattern that
`extract_meshes_for_gpu_building` established.

This improves frame time of `many_cubes` from 3.9ms to approximately 3.1ms, which slightly surpasses the performance of Bevy 0.14.

(Resubmitted from #16878 to clean up history.)

Screenshot 2024-12-17 182109


IceSentry (approved):
Very nice and simple. It's nice that we can finally take advantage of the retained world.

It's unfortunate that we need to duplicate this for 2d, but figuring out an abstraction for that is definitely future work.

tychedelia (changes requested):
I noticed that in practice the query for `Changed<ViewVisibility>` was firing every frame. It seems to be an issue in our lighting code see https://github.com/pcwalton/bevy/pull/7.

pcwalton:
@tychedelia Cherry-picked over, thanks!

tychedelia (approved):

pcwalton:
There are some fairly serious bugs that have required a rewrite of a good deal of this patch.

pcwalton:
This should be ready for re-review. I had to redo a couple of things that the `view_visibility` problem highlighted:

1. The whole reference counting idea was not only incorrect in edge cases, but it also made little sense as we *already* have a mechanism for managing asset lifetimes: the `RenderAsset` infrastructure. I moved binding array allocation over to the normal `prepare_asset`/`unload_asset` method pair, allowing us to eliminate all the error-prone referencing counting.

2. If you change the `MeshMaterial3d` component on a mesh, then we need to re-extract the mesh that that material is extracted to, because `collect_meshes_for_gpu_building` writes the material binding into the `MeshInputUniform`. The easiest way to re-extract a mesh is just to mark its `Mesh3d` as changed, so I added a system to do that when the material changes. (Note that changing the material *itself* doesn't require re-extraction of meshes; the binding is still good. That's why we don't need asset change detection.)

3. Populating the `RenderVisibilityRanges` resource needs to happen before mesh extraction, because `extract_meshes_for_gpu_building` consults that resource. I added the ordering necessary to make this happen.

IceSentry (approved):

Elabajaba (approved):

alice-i-cecile:
Running another example test run; I'll merge this in if that comes up clean :)

alice-i-cecile:
The colors in the examples in this PR seem messed up; are materials being batched correctly? See https://github.com/TheBevyFlock/bevy-example-runner/issues/87#issuecomment-2591520084

pcwalton:
Hmm, I'm having trouble reproducing the failures. It may be some kind of race...

pcwalton:
OK, looks like that fixed the race. I'll look at the regressions.

pcwalton:
All of the failures look either like things that were broken already (the pre-existing wireframe problem having to do with the fact that it breaks the 1:1 entity:material relationship), or are false positives. So I think this is ready to go.

PR #16833, opened 2024-12-15
Remove benchmarks from CI test

sophrosyne97:
# Objective

Fixes [Do not test benchmarks in CI #16803](https://github.com/bevyengine/bevy/issues/16803), part of [Add benchmarks and compile fail tests back to workspace #16801](https://github.com/bevyengine/bevy/issues/16801)

## Solution

Removes the `--benches` flag from the CI tool test.

## Testing

`cargo run -p ci --quiet -- test`

Issue #16803, opened 2024-12-13
Do not test benchmarks in CI


BD103:
Our CI tool currently runs `cargo test --benches`:

https://github.com/bevyengine/bevy/blob/00722b8d0fb1224db99d161b92fd892595fc2823/tools/ci/src/commands/test.rs#L20

Because our benchmarks use a custom harness, though, they do not detect this and just run all benchmarks. This is undesired because benchmarks are lengthy and resource-intensive, and should not contain unit tests. The `--benches` flag should be removed.

PR #16987, opened 2024-12-26
Test benchmarks in CI

BD103:
# Objective

- This reverts #16833, and completely goes against #16803.
- Turns out running `cargo test --benches` runs each benchmark once, without timing it, just to ensure nothing panics. This is actually desired because we can use it to verify benchmarks are working correctly without all the time constraints of actual benchmarks.

## Solution

- Add the `--benches` flag to the CI test command.

## Testing

- `cargo run -p ci -- test`

### DIFF
--- a/tools/ci/src/commands/test.rs
+++ b/tools/ci/src/commands/test.rs
@@ -17,7 +17,8 @@ impl Prepare for TestCommand {
         vec![PreparedCommand::new::<Self>(
             cmd!(
                 sh,
-                // `--benches` runs each benchmark once, to verify that they behave correctly.
+                // `--benches` runs each benchmark once in order to verify that they behave
+                // correctly and do not panic.
                 "cargo test --workspace --lib --bins --tests --benches {no_fail_fast}"
             ),
             "Please fix failing tests in output above.",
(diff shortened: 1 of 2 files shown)


BenjaminBrienen:
Something happened with the bench

```
thread 'main' panicked at benches/benches/bevy_tasks/iter.rs:80:25:
```

tools/ci/src/commands/test.rs
```diff
@@ -17,7 +17,8 @@ impl Prepare for TestCommand {
         vec![PreparedCommand::new::<Self>(
             cmd!(
                 sh,
-                "cargo test --workspace --lib --bins --tests {no_fail_fast}"
+                // `--benches` runs each benchmark once, to verify that they behave correctly.
```
BenjaminBrienen, line 20:
```suggestion
                // `--benches` runs each benchmark once in order to verify that they behave correctly. (Don't panic)
```

BenjaminBrienen (approved):

BenjaminBrienen:
Approving since the failure should be fixed in a separate PR

BD103:
> Approving since the failure should be fixed in a separate PR

Surprisingly not! The panic was an integer overflow that was never caught in our benchmarks because they are usually compiled in release mode. I fixed it in aa2e600f3aab8427821a6a21b73fc82895a3cc77, though, to explicitly wrap.

BenjaminBrienen (approved):

Issue #16988, opened 2024-12-26
Crashes on startup due to too many bindings


BenjaminBrienen:
## Bevy version

c03e494: Migrate reflection benchmarks to new naming system (#16986) …

## \[Optional\] Relevant system information

```
PS C:\Users\BenjaminBrienen\source\bevy> cargo version                                                                                                                                                                              
cargo 1.83.0 (5ffbef321 2024-10-29)
```
```
SystemInfo { os: "Windows 11 Pro", kernel: "26100", cpu: "12th Gen Intel(R) Core(TM) i7-12700H", core_count: "14", memory: "31.7 GiB" }
```
```
AdapterInfo { name: "Intel(R) Iris(R) Xe Graphics", vendor: 32902, device: 18086, device_type: IntegratedGpu, driver: "Intel Corporation", driver_info: "101.6325", backend: Vulkan }
```

My drivers are very up-to-date.

## What you did

```
PS C:\Users\BenjaminBrienen\source\bevy> cargo run --example testbed_ui --no-default-features --features "android-game-activity android-game-activity android_shared_stdcxx animation bevy_asset bevy_audio bevy_color bevy_core_pipeline bevy_gilrs bevy_gizmos bevy_gltf bevy_mesh_picking_backend bevy_pbr bevy_picking bevy_remote bevy_render bevy_scene bevy_sprite bevy_sprite_picking_backend bevy_state bevy_ui bevy_ui_picking_backend bevy_window bevy_winit custom_cursor default_font hdr multi_threaded png smaa_luts sysinfo_plugin tonemapping_luts vorbis webgl2 x11"
```

## What went wrong

```
warning: error finalizing incremental compilation session directory `\\?\C:\Users\BenjaminBrienen\source\bevy\target\debug\incremental\bevy_gltf-3d0vzuikmb2aw\s-h33e87o7fs-12utzwx-working`: Access is denied. (os error 5)

warning: `bevy_gltf` (lib) generated 1 warning
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.84s
     Running `target\debug\examples\testbed_ui.exe`
2024-12-26T21:32:35.430861Z  INFO bevy_diagnostic::system_information_diagnostics_plugin::internal: SystemInfo { os: "Windows 11 Pro", kernel: "26100", cpu: "12th Gen Intel(R) Core(TM) i7-12700H", core_count: "14", memory: "31.7 GiB" }
[...]

PR #16989, opened 2024-12-27
Fix logic error in loading_screen.rs

### LINKED ISSUE #16792: Error in loading_screen.rs example
## Bevy version

v0.15.0

## What went wrong

I think there is an error in the [example](https://github.com/bevyengine/bevy/blob/latest/examples/games/loading_screen.rs#L220) on line 220.
If there are two elements in the array and the zero element is deleted on the first iteration, then on the second iteration an attempt to delete the element with index 1 will end with a panic because the deletion of the zero element caused a shift.

BenjaminBrienen:
# Objective

Fix #16792

## Solution

Fix the logic to retain loaded ones

## Testing

Unable to test due to #16988


MiniaczQ:
Example appears to work correctly

alice-i-cecile (approved):

examples/games/loading_screen.rs
```diff
@@ -204,21 +204,13 @@ fn update_loading_data(
         // we reset the confirmation frame count.
         loading_data.confirmation_frames_count = 0;
 
-        // Go through each asset and verify their load states.
-        // Any assets that are loaded are then added to the pop list for later removal.
-        let mut pop_list: Vec<usize> = Vec::new();
-        for (index, asset) in loading_data.loading_assets.iter().enumerate() {
+        loading_data.loading_assets.retain(|asset| {
             if let Some(state) = asset_server.get_load_states(asset) {
```
rparrett, line 208: nit: could be
```rust
asset_server
    .get_recursive_dependency_load_state(asset)
    .map_or(true, |state| !state.is_loaded())
```

BenjaminBrienen: Actually, I think that would trigger a lint for `is_none_or`. :p

rparrett: It doesn't but that is even nicer :)

PR #16991, opened 2024-12-27
Fix `Text2d` performance regression

### LINKED ISSUE #16972: Poor performance for Text2d with TextSpan
- ~5 fps for Text2d with TextSpan 
- ~20 fps when i use only Text2d
- ~60 fps when i use old bevy 0.13 -_-

_Originally posted by @MikMikus in https://github.com/bevyengine/bevy/issues/16783#issuecomment-2545826342_

rparrett:
# Objective

Probably fixes #16972

## Solution

With 100k text2d, tracy was showing most time being spent in `extract_components<bevy_sprite::SpriteSource>`.

image

Browsing Bevy's code, this `SpriteSource` component is seemingly not even used in the render world. So I just ~~deleted the code that was extracting it~~ it.

## Testing

`cargo run --example text2d` still seems to work.

The example from [my comment](https://github.com/bevyengine/bevy/issues/16972#issuecomment-2562680876) in the linked issue shows a ~50x speedup.

### DIFF
--- a/crates/bevy_sprite/src/lib.rs
+++ b/crates/bevy_sprite/src/lib.rs
@@ -36,7 +36,6 @@ pub mod prelude {
     };
 }
 
-use bevy_reflect::{std_traits::ReflectDefault, Reflect};
 pub use bundle::*;
 pub use dynamic_texture_atlas_builder::*;
 pub use mesh2d::*;
@@ -58,7 +57,7 @@ use bevy_render::{
     primitives::Aabb,
     render_phase::AddRenderCommand,
     render_resource::{Shader, SpecializedRenderPipelines},
-    view::{self, NoFrustumCulling, VisibilityClass, VisibilitySystems},
+    view::{NoFrustumCulling, VisibilitySystems},
     ExtractSchedule, Render, RenderApp, RenderSet,
 };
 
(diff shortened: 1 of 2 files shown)


crates/bevy_sprite/src/lib.rs
```diff
@@ -229,18 +223,6 @@ pub fn calculate_bounds_2d(
     }
 }
 
-impl ExtractComponent for SpriteSource {
```
alice-i-cecile, line 232: We should probably remove `SpriteSource` completely if we're doing this.

rparrett:
I found another minor optimization. `update_text2d_layout_system` is checking each text entity to see if it's in a local hashset of texts that need to be retried. This wasn't using `EntityHashSet`.

image

rparrett:
I am realizing now that removing `SpriteSource` makes this a breaking change. This type isn't new to `0.15`, so I am not sure about removing it entirely now if this PR is going to be included in a patch release.

pcwalton (approved):
I don’t think this can be backported to 0.15 in the first place because visibility classes were introduced in 0.16.

rparrett:
> I don’t think this can be backported to 0.15 in the first place because visibility classes were introduced in 0.16.

Oh, oops. I can do a separate non-breaking PR for 0.15.

alice-i-cecile (approved):

PR #16993, opened 2024-12-27
Remove `Send + Sync` bounds from `PartialReflect` and friends (and move the constraint to usages thereof)

### LINKED ISSUE #16713: Reflect trait should not require Send + Sync bounds
## Bevy version

0.15

## Problem

`Reflect` has no intrinsic need to be Send + Sync. This limits its uses, especially in a broader non-Bevy context.

## Solution

Remove `Send` + `Sync` trait bounds, adding them locally to any methods that fail to compile.

## Context

According to @MrGVSV, this was originally for compatibility with components, before the explicit `Component` trait was added.

BenjaminBrienen:
# Objective

Fixes #16713

## Solution

Move the bounds to the usage sites

## Disclaimer

**If you read this while it is a draft, be aware that this is not ready for a round of full feedback.**


BenjaminBrienen:
I might pick this up again in 4 or 5 weeks when rust 1.86 lands, which includes https://github.com/rust-lang/rust/issues/65991

alice-i-cecile:
Cutting from the milestone: there's nothing urgent about this, as much as I think it's great.

BenjaminBrienen:
A follow up is needed to remove the ` + Send + Sync and truly make the engine more generic. This is just the first step.

crates/bevy_asset/src/handle.rs
```diff
@@ -691,16 +691,17 @@ mod tests {
                     "Inserting the asset should result in a strong count of 1"
                 );
 
-                let reflected: &dyn Reflect = &handle;
-                let _cloned_handle: Box<dyn Reflect> = reflected.reflect_clone().unwrap();
+                let reflected: &(dyn Reflect + Send + Sync) = &handle;
```
MrGVSV, line 694: Are the bounds necessary here?

crates/bevy_reflect/src/reflect.rs
```diff
@@ -106,32 +106,34 @@ where
     /// Casts this type to a boxed, reflected value.
     ///
     /// This is useful for coercing trait objects.
-    fn into_partial_reflect(self: Box<Self>) -> Box<dyn PartialReflect>;
+    fn into_partial_reflect(self: Box<Self>) -> Box<dyn PartialReflect + Send + Sync>;
```
MrGVSV, line 109: Hm, I feel like this sort of defeats the purpose, no?

The only way we can return `Box::new(self)` here is if `Self: Send + Sync`.

PR #16995, opened 2024-12-27
Add `no_std` support to `bevy_input`

bushrat011899:
# Objective

- Contributes to #15460

## Solution

  - `std` (default)
  - `smol_str` (default)
  - `portable-atomic`
  - `critical-section`
  - `libm`
- Fixed an existing issue where `bevy_reflect` wasn't properly feature gated.

## Testing

- CI

## Notes

- There were some minor issues with `bevy_math` and `bevy_ecs` noticed in this PR which I have also resolved here. I can split these out if desired, but I've left them here for now as they're very small changes and I don't consider this PR itself to be very controversial.
- `libm`, `portable-atomic`, and `critical-section` are shortcuts to enable the relevant features in dependencies, making the usage of this crate on atomically challenged platforms possible and simpler.
- `smol_str` is gated as it doesn't support atomically challenged platforms (e.g., Raspberry Pi Pico). I have an issue and a [PR](https://github.com/rust-analyzer/smol_str/pull/91) to discuss this upstream.

### DIFF
--- a/crates/bevy_input/Cargo.toml
+++ b/crates/bevy_input/Cargo.toml
@@ -9,7 +9,7 @@ license = "MIT OR Apache-2.0"
 keywords = ["bevy"]
 
 [features]
-default = ["std", "bevy_reflect"]
+default = ["std", "bevy_reflect", "bevy_ecs/async_executor", "smol_str"]
 
 # Functionality
 
@@ -29,6 +29,9 @@ serialize = [
   "bevy_math/serialize",
 ]
 
+## Uses the small-string optimization provided by `smol_str`.
+smol_str = ["dep:smol_str", "bevy_reflect/smol_str"]
+
 # Platform Compatibility
 
 ## Allows access to the `std` crate. Enabling this feature will prevent compilation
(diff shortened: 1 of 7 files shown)


BenjaminBrienen (approved):

crates/bevy_input/src/keyboard.rs
```diff
@@ -72,8 +72,14 @@ use bevy_ecs::{
     event::{Event, EventReader},
     system::ResMut,
 };
+
 #[cfg(feature = "bevy_reflect")]
 use bevy_reflect::Reflect;
+
+#[cfg(not(feature = "smol_str"))]
+use alloc::string::String as SmolStr;
```
BenjaminBrienen, line 80: Cursed

BenjaminBrienen (approved):

crates/bevy_input/src/keyboard.rs
bushrat011899: Just avoids further changing code without reason, but I do agree this isn't pretty haha. I'm hoping the PR I've made upstream will be worked-on/accepted so we can just remove this feature gate entirely.

crates/bevy_input/Cargo.toml
```diff
@@ -9,36 +9,72 @@ license = "MIT OR Apache-2.0"
...
+libm = ["bevy_math/libm"]
 
 [dependencies]
 # bevy
 bevy_app = { path = "../bevy_app", version = "0.15.0-dev", default-features = false }
-bevy_ecs = { path = "../bevy_ecs", version = "0.15.0-dev", default-features = false, features = [
-  "serialize",
-] }
-bevy_math = { path = "../bevy_math", version = "0.15.0-dev", default-features = false, features = [
-  "rand",
```
IQuick143, line 28: Seems that the `bevy_math/rand` feature just disappeared during these changes.
Is that intentional? (I mean I don't see a reason for bevy_input to need random numbers.)

bushrat011899: Yeah I deliberately left that feature off since I couldn't see anything in `bevy_input` that actually needed it.

alice-i-cecile (approved):

PR #16998, opened 2024-12-27
Add `no_std` support to `bevy_hierarchy`

bushrat011899:
# Objective

- Contributes to #15460

## Solution

  - `std` (default)

## Testing

- CI

## Notes

- There was a minor issue with `bevy_reflect`'s `smallvec` feature noticed in this PR which I have also resolved here. I can split this out if desired, but I've left it here for now as it's a very small change and I don't consider this PR itself to be very controversial.

### DIFF
--- a/crates/bevy_hierarchy/src/hierarchy.rs
+++ b/crates/bevy_hierarchy/src/hierarchy.rs
@@ -8,7 +8,7 @@ use bevy_ecs::{
     system::EntityCommands,
     world::{Command, DeferredWorld, EntityWorldMut, World},
 };
-use bevy_utils::tracing::debug;
+use log::debug;
 
 /// Despawns the given entity and all its children recursively
 #[derive(Debug)]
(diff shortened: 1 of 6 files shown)


crates/bevy_hierarchy/src/hierarchy.rs
```diff
@@ -8,7 +8,7 @@ use bevy_ecs::{
     system::EntityCommands,
     world::{Command, DeferredWorld, EntityWorldMut, World},
 };
-use bevy_utils::tracing::debug;
+use log::debug;
```
BenjaminBrienen, line 11: Why is log necessary if we have tracing?

BenjaminBrienen (approved):

crates/bevy_hierarchy/src/hierarchy.rs
bushrat011899: `tracing` doesn't work on certain platforms if they don't support atomic CAS, whereas `log` works on every platform. Most notable example is Raspberry Pi Pico and other embedded ARM CPUs.

alice-i-cecile (approved):

PR #8895, opened 2023-06-20
implement insert and remove reflected entity commands

NoahShomette:
Objective
Implement two new EntityCommands functions for inserting and removing reflected components.
Solution
Implements a new EntityCommands extension trait for reflection related functions in the reflect module of bevy_ecs.
Implements two new functions, insert_reflected and remove_reflected. Both functions take a Box<dyn Reflect> component as well as a TypeRegistry. The type registry is used to get the reflection data for the component and enable the inserting and removing.
Derive Clone for TypeRegistry to enable cloning it into the commands.
Made EntityCommands fields pub(crate) to allow access in the reflect module. (Might be worth making these just public to enable user end custom entity commands?)
Added basic tests to ensure the commands are actually working.
Documentation of functions.
Note: I'm not entirely sure on cloning the TypeRegistry a bunch. An alternative solution could be to take a reference of the TypeRegistry in the command function, get the registration data out of it right when its called, and then save that data into the command struct. Basically instead of saving the whole TypeRegistry just grab the ReflectComponent out of it and save that into the command struct. However that would involve options and woudl move panics to when the command initially gets called rather than when the command is actually applied. Not sure on exactly how much of a command is allowed to happen before its applied or if this would be preferred. It would prevent cloning the TypeRegistry though.

Changelog
Added:

Implements two new functions, insert_reflected and remove_reflected. Both functions take a Box<dyn Reflect> component as well as a TypeRegistry. The type registry is used to get the reflection data for the component and enable the inserting and removing.
Changed:

Derive Clone for TypeRegistry to enable cloning it into the commands.
Made EntityCommands fields pub(crate) to allow access in the reflect module.

> Hopefully this time it works. Please don't make me rebase again ☹

PR #16999, opened 2024-12-27
Move some structs that `impl Command` to methods on `World` and `EntityWorldMut`

JaySpruce:
## Objective

Commands were previously limited to structs that implemented `Command`. Now there are blanket implementations for closures, which (in my opinion) are generally preferable.

Internal commands within `commands/mod.rs` have been switched from structs to closures, but there are a number of internal commands in other areas of the engine that still use structs. I'd like to tidy these up by moving their implementations to methods on `World`/`EntityWorldMut` and changing `Commands` to use those methods through closures.

This PR handles the following:
- `TriggerEvent` and `EmitDynamicTrigger` double as commands and helper structs, and can just be moved to `World` methods.
- Four structs that enabled insertion/removal of components via reflection. This functionality shouldn't be exclusive to commands, and can be added to `EntityWorldMut`.
- Five structs that mostly just wrapped `World` methods, and can be replaced with closures that do the same thing.

## Solution

- __Observer Triggers__ (`observer/trigger_event.rs` and `observer/mod.rs`)
  - Moved the internals of `TriggerEvent` to the `World` methods that used it.
  - Replaced `EmitDynamicTrigger` with two `World` methods:
    - `trigger_targets_dynamic`
    - `trigger_targets_dynamic_ref`
  - `TriggerTargets` was now the only thing in `observer/trigger_event.rs`, so it's been moved to `observer/mod.rs` and `trigger_event.rs` was deleted.
- __Reflection Insert/Remove__ (`reflect/entity_commands.rs`)
  - Replaced the following `Command` impls with equivalent methods on `EntityWorldMut`:
    - `InsertReflect` -> `insert_reflect`
    - `InsertReflectWithRegistry` -> `insert_reflect_with_registry`
    - `RemoveReflect` -> `remove_reflect`
    - `RemoveReflectWithRegistry` -> `remove_reflect_with_registry`
- __System Registration__ (`system/system_registry.rs`)
  - The following `Command` impls just wrapped a `World` method and have been replaced with closures:
    - `RunSystemWith`
    - `UnregisterSystem`
    - `RunSystemCachedWith`
    - `UnregisterSystemCached`
  - `RegisterSystem` called a helper function that basically worked as a constructor for `RegisteredSystem` and made sure it came with a marker component. That helper function has been replaced with `RegisteredSystem::new` and a `#[require]`.

## Possible Addition

The extension trait that adds the reflection commands, `ReflectCommandExt`, isn't strictly necessary; we could just `impl EntityCommands`. We could even move them to the same files as the main impls and put it behind a `#[cfg]`.

The PR that added it [had a similar conversation](https://github.com/bevyengine/bevy/pull/8895#discussion_r1234713671) and decided to stick with the trait, but we could revisit it here if so desired.


LikeLakers2 (approved):
When I read the title and PR description, I was initially concerned that you were removing functions from `Commands`.

However, I read through the code, and I'm glad to see that isn't the case - you were simply changing their code from queuing up a `impl Command` struct, to queuing up a closure.

I'm not a bevy team member, but I approve this nonetheless.

alice-i-cecile (approved):

alice-i-cecile:
@LikeLakers2 thanks for the review here: that's exactly the pattern of community review that Bevy is built around :)

PR #17005, opened 2024-12-28
Add fallible add methods to `PluginGroupBuilder`

### LINKED ISSUE #17001: Add `contains` or `has` to `PluginGroupBuilder`

`PluginGroupBuilder` has several panicking methods such as `set` and `disable`, which panic if the given type of plugin is missing from the plugin group. These methods have no error-returning variants, and there is currently no way to check from the outside if a `PluginGroupBuilder` has a given plugin, even though internally it would just be a simple `contains_key` call to the plugin hash map.

This can be problematic for APIs that take or produce a `PluginGroupBuilder` and want to perform some operations on it, or otherwise just need to check if a plugin exists.

Add a `contains` or `has` method to `PluginGroupBuilder`.

- Add non-panicking variants for all of the panicking methods (not exclusive with a `contains` method, both could exist)
- Make the panicking methods non-panicking (doesn't necessarily solve just wanting to know if a plugin exists)

Brezak:
# Objective

Make working with `PluginGroupBuilder` less panicky.
Fixes #17001

## Solution

Expand the `PluginGroupBuilder` api with fallible add methods + a contains method.
Also reorder the `PluginGroupBuilder` tests because before should come before after.

## Testing

Ran the `PluginGroupBuilder` tests which call into all the newly added methods.

### DIFF
--- a/crates/bevy_app/src/plugin_group.rs
+++ b/crates/bevy_app/src/plugin_group.rs
@@ -4,7 +4,7 @@ use alloc::{
     string::{String, ToString},
     vec::Vec,
 };
-use bevy_utils::TypeIdMap;
+use bevy_utils::{hashbrown::hash_map::Entry, TypeIdMap};
 use core::any::TypeId;
 use log::{debug, warn};
 
@@ -224,11 +224,6 @@ impl PluginGroup for PluginGroupBuilder {
     }
 }
 
-/// Helper method to get the [`TypeId`] of a value without having to name its type.
-fn type_id_of_val<T: 'static>(_: &T) -> TypeId {
-    TypeId::of::<T>()
-}
-
 /// Facilitates the creation and configuration of a [`PluginGroup`].
 ///
 /// Provides a build ordering to ensure that [`Plugin`]s which produce/require a [`Resource`](bevy_ecs::system::Resource)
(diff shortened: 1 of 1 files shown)


crates/bevy_app/src/plugin_group.rs
```diff
@@ -311,15 +310,28 @@ impl PluginGroupBuilder {
...
-        });
-        entry.plugin = Box::new(plugin);
-        self
+        })
+    }
+
+    /// Tries to set the value of the given [`Plugin`], if it exists.
+    ///
+    /// If the given plugin doesn't exist returns self and the passed in [`Plugin`].
+    // TODO: Should we return the original plugin value?
```
alice-i-cecile, line 325: Eh, probably not useful enough to complicate the return value.

alice-i-cecile (approved):

crates/bevy_app/src/plugin_group.rs
```diff
@@ -357,23 +379,105 @@ impl PluginGroupBuilder {
...
-        self.order.insert(target_index, type_id_of_val(&plugin));
+    ///
+    /// If the plugin was already the group, it is removed from its previous place.
+    ///
+    /// # Panics
+    ///
+    /// Panics if `Target` is not already in this [`PluginGroupBuilder`].
+    pub fn add_after<Target: Plugin>(self, plugin: impl Plugin) -> Self {
+        self.try_add_after_overwrite::<Target, _>(plugin)
+            .unwrap_or_else(|_| {
```
BenjaminBrienen, line 440: Isn't this just `expect()`?

BenjaminBrienen (approved):

crates/bevy_app/src/plugin_group.rs
alice-i-cecile: Not quite! The unwrap_or_else pattern is useful as a performance optimization when string formatting needs to be done in the error case. IIRC otherwise things are pessimized even in the happy case.

PR #17011, opened 2024-12-28
Use -D warnings in all relevant CI

### LINKED ISSUE #17009: Build warnings with certain feature combinations
In CI, we don't use warnings as errors in all relevant places, so warnings are found, but don't block merging. These warnings should be fixed and the CI should fail on warnings.

https://github.com/bevyengine/bevy/actions/runs/12519981025/job/34925080257#step:5:25

https://doc.rust-lang.org/stable/clippy/continuous_integration/index.html

BenjaminBrienen:
# Objective

Fixes #17009

See: https://doc.rust-lang.org/stable/clippy/continuous_integration/index.html

## Solution

Add the env

## Testing

CI should start to fail, then I'll fix it.

## Showcase

image

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -170,6 +170,8 @@ jobs:
         with:
           target: wasm32-unknown-unknown
       - name: Check wasm
+        env:
+          RUSTFLAGS: "-D warnings"
         run: cargo check --target wasm32-unknown-unknown
 
   build-wasm-atomics:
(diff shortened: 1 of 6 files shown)


alice-i-cecile:
I broadly agree with this: deny warnings is the standard practice for Rust + Bevy, and blocking CI is the best way to ensure that these are fixed and stay fixed.

BenjaminBrienen:
An alternative direction that we could take is to deny warnings in CI *except* for unused imports, variables, and muts. This might help keep verbosity down while still catching important lints. If we do that though, I would expect all the cfg stuff to be removed in all existing places we already do do that.

BD103 (approved):

alice-i-cecile:
> I don't like this though because I don't want irrelevant warnings when developing locally and there isn't enough granularity with allowing warnings in unchanged code.

Agreed. Half of the reason why these lints exist are to provide a better local development experience.

mockersf:
OK to add `-D warnings` to CI, but I would prefer the warnings in task pool creating to be solved in a way that keep their builder for wasm and not wasm common

BenjaminBrienen:
I can do that

alice-i-cecile (approved):

mockersf (approved):

PR #17012, opened 2024-12-28
Fix msrvs

### LINKED ISSUE #17008: `rust-version`s are incorrect
The `rust-verssion` in bevy_color, bevy_math, bevy_ecs, and bevy_input_focus are all wrong.

BenjaminBrienen:
# Objective

The rust-versions are out of date.
Fixes #17008

## Solution

Update the values

Cherry-picked from #17006 in case it is controversial

## Testing

Validated locally and in #17006

### DIFF
--- a/crates/bevy_animation/src/lib.rs
+++ b/crates/bevy_animation/src/lib.rs
@@ -1057,8 +1057,8 @@ pub fn animate_targets(
                     (player, graph_handle.id())
                 } else {
                     trace!(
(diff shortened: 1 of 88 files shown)


alice-i-cecile (approved):
There's a broader question of whether or not these crates should all *have* a listed MSRV, but given that it currently exists it should not be wrong.

BenjaminBrienen:
I can remove the one for input_focus if you want @alice-i-cecile

PR #17013, opened 2024-12-28
Rename `NodeImageMode` to `ImageNodeMode`

### LINKED ISSUE #16909: Consider renaming `NodeImageMode` to `ImageNodeMode`
- In 0.15, we now have `ImageNode`, and I just noticed it has a field `image_mode` of type `NodeImageMode` (notice the inconsistency):

```rust
pub struct ImageNode {
    /// Controls how the image is altered to fit within the layout and how 
    /// the layout algorithm determines the space to allocate for the image.
    pub image_mode: NodeImageMode,
}
```

- Looks like `image_mode` was added in https://github.com/bevyengine/bevy/pull/16088 and merged on 5 Nov.
- `UiImage` was renamed by cart in https://github.com/bevyengine/bevy/pull/16271 and merged on 8 Nov.
- It's possible the close timing here meant it was missed.
- I can't see `NodeImageMode` used by anything else that would suggest keeping this name was deliberate.

BenjaminBrienen:
# Objective

Fixes #16909 

## Solution

Rename the type

## Migration Guide

`widget::image::NodeImageMode` is now `widget::image::ImageNodeMode`

### DIFF
--- a/crates/bevy_ui/src/render/ui_texture_slice_pipeline.rs
+++ b/crates/bevy_ui/src/render/ui_texture_slice_pipeline.rs
@@ -275,10 +275,10 @@ pub fn extract_ui_texture_slices(
         };
 
         let image_scale_mode = match image.image_mode.clone() {
-            widget::NodeImageMode::Sliced(texture_slicer) => {
+            widget::ImageNodeMode::Sliced(texture_slicer) => {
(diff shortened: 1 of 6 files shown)


ickshonpe (commented):
I prefer `NodeImageMode`.

PR #17014, opened 2024-12-28
Add doc to `derive_as_bind_group`

### LINKED ISSUE #16955: AsBindGroup derive macro should link to the AsBindGroup trait documentation
The derive macro documentation is [here](https://docs.rs/bevy/latest/bevy/render/render_resource/derive.AsBindGroup.html).
Currently, the derive attributes (`#[uniform]`, `#[texture_storage]`, etc) are well documented in the [trait](https://docs.rs/bevy/latest/bevy/render/render_resource/trait.AsBindGroup.html) documentation. From the trait perspective, this makes sense, because the vast majority of users will never manually implement this trait, however it would make more sense to put this information in the derive macro documentation and link to it in the trait documentation.
I think that's probably more effort than it's worth though, so instead I'm suggesting to just add a sentence to the macro documentation saying that there is much more detailed information available in the trait documentation (an a link to it).

BenjaminBrienen:
# Objective

Fixes #16955

## Solution

## Testing

See showcase

## Showcase

image

### DIFF
--- a/crates/bevy_render/macros/src/lib.rs
+++ b/crates/bevy_render/macros/src/lib.rs
@@ -66,6 +66,7 @@ pub fn derive_extract_component(input: TokenStream) -> TokenStream {
         bindless
     )
 )]
+/// See [`AsBindGroup`](crate::render_resource::AsBindGroup) for detailed information.
 pub fn derive_as_bind_group(input: TokenStream) -> TokenStream {
     let input = parse_macro_input!(input as DeriveInput);
 


BenjaminBrienen:
https://github.com/rust-lang/api-guidelines/discussions/186#discussioncomment-232142

BenjaminBrienen:
This doesn't seem to be possible. Adding the documentation within bevy_render_macros causes the link to not work because it is an external link. Trying to add the documentation on the re-export in bevy_render just breaks the documentation page outright.

PR #17015, opened 2024-12-29
Scale input to account for deadzones

### LINKED ISSUE #3450: Deadzone clipping in AxisSettings and ButtonSettings results in a compression of values
## Problem

The [current behavior](https://github.com/bevyengine/bevy/blob/e48f9d8cd4562d5ea6b483764ae9288feb13cc72/crates/bevy_input/src/gamepad.rs#L188) of `AxisSettings` is:

- If an axis's value is below the low-threshold, set them to 0.0.
- If an axis's is above the high-threshold, set them to 1.0.

This is sensible enough, and it's useful to be able to configure this in order to control dead zones and partially account for controller drift.

However, this results in a compression of values: rather than the controller always being able to send values from -1.0 to 1.0, they can only send values in a restricted range: the high, low and near-zero values are just clipped out.

This is undesirable because it causes player-configurable settings which are intended to account for strange quirks in physical controller behavior to have unintuitive and far-reaching gameplay effects.

The same effect occurs for analogue buttons like triggers in `ButtonSetttings`.

## Proposed Solution

Rescale values to cover the full range regardless of the configuration. The formula (for positive values) is:

```rust
let new_value = (value - positive_low) / (positive_high - positive_low);
```

If the game cares about the original raw input values instead, they can intercept the event stream themselves.

## Alternatives

Make this a configurable setting, and keep the current behavior if the `rescale_values` field in the `AxisSettings` is set to false.

BenjaminBrienen:
# Objective

Fixes #3450 

## Solution

Scale the input to account for the range

## Testing

Updated unit tests

## Migration Guide

`GamepadButtonChangedEvent.value` is now linearly rescaled to be from `0.0..=1.0` (instead of `low..=high`) and `GamepadAxisChangedEvent.value` is now linearly rescaled to be from `-1.0..=0.0`/`0.0..=1.0` (accounting for the deadzone).


BenjaminBrienen:
I can't work on fixing the unit tests until this is fixed:
```
error[E0599]: no method named `to_string` found for reference `&str` in the current scope
   --> crates\bevy_ecs\src\name.rs:252:24
    |
252 |         Ok(Name::new(v.to_string()))
    |                        ^^^^^^^^^ method not found in `&str`
    |
    = help: items from traits can only be used if the trait is in scope
help: trait `ToString` which provides `to_string` is implemented but not in scope; perhaps you want to import it
    |
3   + use crate::alloc::string::ToString;
```

crates/bevy_input/src/gamepad.rs
```diff
@@ -1666,10 +1715,10 @@ mod tests {
             (0.99, None, Some(1.0)),
             (0.96, None, Some(1.0)),
             (0.95, None, Some(1.0)),
-            (0.9499, None, Some(0.9499)),
-            (0.84, None, Some(0.84)),
-            (0.43, None, Some(0.43)),
-            (0.05001, None, Some(0.05001)),
+            (0.9499, None, Some(0.9998889)),
```
alice-i-cecile, line 1718: These lines need a comment explaining why they don't match the obvious value.

crates/bevy_input/src/gamepad.rs
```diff
@@ -1239,13 +1240,61 @@ impl AxisSettings {
     ///
     /// Returns the clamped `new_value` if the change exceeds the settings threshold,
     /// and `None` otherwise.
-    pub fn filter(&self, new_value: f32, old_value: Option<f32>) -> Option<f32> {
-        let new_value = self.clamp(new_value);
+    fn filter(&self, new_value: f32, old_value: Option<f32>) -> Option<ScaledAxisPosition> {
+        let axis_position = self.get_axis_position_from_value(new_value);
+        if !self.should_register_change(new_value, old_value) {
```
alice-i-cecile, line 1245: This would be clearer as a match statement.

crates/bevy_input/src/gamepad.rs
alice-i-cecile: Hmm, this is a recurring problem 🤔

crates/bevy_input/src/gamepad.rs
```diff
@@ -1239,13 +1240,61 @@ impl AxisSettings {
...
-        if self.should_register_change(new_value, old_value) {
-            return Some(new_value);
+#[inline(always)]
+fn scale(value: f32, low: f32, high: f32) -> f32 {
+    (value - low) / (high - low)
+}
+
+#[derive(Debug, Clone, Copy)]
+/// Deadzone-aware axis position.
+enum ScaledAxisPosition {
```
alice-i-cecile, line 1282: Doc strings for each variant please.

alice-i-cecile (commented):
I agree that this is the right default behavior. Otherwise setting a deadzone makes it very hard / impossible to get small directional inputs.

This PR needs a bit of cleanup still. It would be particularly helpful to cleanup the tests: right now, the values just look like magic numbers. Maybe we should add some algebra + consts inside of the Some arms?

alice-i-cecile:
For the alloc error, I think you need to fix the feature flags in a new PR. @bushrat011899, FYI.

crates/bevy_input/src/gamepad.rs
BenjaminBrienen: I copied it from the PR that I adopted. I'll add a comment once I can compile and test bevy again.

BenjaminBrienen:
> For the alloc error, I think you need to fix the feature flags in a new PR. @bushrat011899, FYI. 

He's already got that in the no_std bevy_input PR.

bushrat011899:
Can confirm I have this issue resolved in the `no_std` `bevy_input` PR, along with some other issues with how `bevy_input` feature gates `bevy_reflect`

crates/bevy_input/src/gamepad.rs
```diff
@@ -1750,27 +1885,42 @@ mod tests {
 
     #[test]
     fn test_axis_settings_default_filter_with_old_values() {
+        let threshold = 0.01;
```
BenjaminBrienen, line 1888: I chose not to hardcode the values here because the logic is more important than seeing the literal value

crates/bevy_input/src/gamepad.rs
```diff
@@ -1661,15 +1770,21 @@ mod tests {
...
+            // high enough to be rounded to 1
             (1.0, None, Some(1.0)),
             (0.99, None, Some(1.0)),
             (0.96, None, Some(1.0)),
             (0.95, None, Some(1.0)),
-            (0.9499, None, Some(0.9499)),
-            (0.84, None, Some(0.84)),
-            (0.43, None, Some(0.43)),
-            (0.05001, None, Some(0.05001)),
+            // for the following, remember that 0.05 is the "low" value and 0.95 is the "high" value
```
BenjaminBrienen, line 1780: most tests are just testing the scaling, so it makes more sense to hardcode the expected result

alice-i-cecile (approved):

alice-i-cecile:
Let's do a quick migration guide noting this change (since it's otherwise surprising), but this LGTM now. Good work on the tests; I like the balance you struck.

BenjaminBrienen:
Sorry for the noise. I realized that I could make the implementation and test data cleaner by storing the raw values in the gamepad, which means the old values will be raw values. Everything is very optimal now, imo.

BenjaminBrienen:
Migration guide made me realize the doc comment should be improved. Both are done now.

alice-i-cecile (approved):
I like the latest changes, and particularly appreciate the comments on the tests.

crates/bevy_input/src/gamepad.rs
```diff
@@ -1213,39 +1216,138 @@ impl AxisSettings {
...
         }
     }
 
-    /// Determines whether the change from `old_value` to `new_value` should
+    /// Determines whether the change from `old_raw_value` to `new_raw_value` should
     /// be registered as a change, according to the [`AxisSettings`].
-    fn should_register_change(&self, new_value: f32, old_value: Option<f32>) -> bool {
-        if old_value.is_none() {
-            return true;
+    #[inline(always)]
```
Shatur, line 1233: Never inline private methods. Same for others. [Here](https://matklad.github.io/2021/07/09/inline-in-rust.html) is a good article about it.
```suggestion
```

crates/bevy_input/src/gamepad.rs
```diff
@@ -1213,39 +1216,138 @@ impl AxisSettings {
...
+                scaled: self.get_axis_position_from_value(clamped_unscaled),
+                raw: new_raw_value,
+            }),
+            false => None,
+        }
+    }
+
+    #[inline(always)]
+    fn get_axis_position_from_value(&self, value: f32) -> ScaledAxisWithDeadZonePosition {
+        if value < self.deadzone_upperbound && value > self.deadzone_lowerbound {
```
Shatur, line 1263: Maybe use match? I think this way you won't need the `unreachable`.

crates/bevy_input/src/gamepad.rs
```diff
@@ -1213,39 +1216,138 @@ impl AxisSettings {
...
+                self.livezone_lowerbound..=self.deadzone_lowerbound,
+                -1.0..=0.0,
+            ))
+        } else {
+            unreachable!();
+        }
+    }
+}
+
+/// A linear remapping of `value` from `old` to `new`
```
Shatur, line 1287: Use `.` for all doc comments in Rust:
```suggestion
/// A linear remapping of `value` from `old` to `new`.
```

crates/bevy_input/src/gamepad.rs
```diff
@@ -1648,130 +1778,161 @@ mod tests {
...
-            (0.95, Some(0.945), Some(1.0)),
-            (-0.43, Some(-0.44001), Some(-0.43)),
-            (-0.43, Some(-0.44), None),
-            (-0.43, Some(-0.43), None),
-            (-0.43, Some(-0.41999), Some(-0.43)),
-            (-0.43, Some(-0.17), Some(-0.43)),
-            (-0.43, Some(-0.84), Some(-0.43)),
-            (-0.05, Some(-0.055), Some(0.0)),
-            (-0.95, Some(-0.945), Some(-1.0)),
+            // enough increase to change
```
Shatur, line 1903: I would suggest to split this single test into multiple small tests named like `enough_increase_to_change` instead of comments. But it's up to you since it was like this before.

Shatur (changes requested):
Looks good to me! Left a few small suggestions.

Personally, I think deadzone is a more high-level functionality and should be handled by input managers🤔 But since we already have this feature, it's definitely an improvement!
I handle deadzones in the same way in the `bevy_enhanced_input` library.

alice-i-cecile:
> Personally, I think deadzone is a more high-level functionality and should be handled by input managers

Agreed. In the medium term, I think this functionality should be removed from bevy_input and moved up a level. But until we have a first party input manager / action crate we're blocked there.

crates/bevy_input/src/gamepad.rs
BenjaminBrienen: I tried but it didn't look better because it was all within guard clauses and it still needed an unreachable `_ => ...` branch. The compiler doesn't reason about guard clauses.

Shatur: Ah, I see :(

BenjaminBrienen:
I think deadzones and scaling are good here because they have to do with accounting for the hardware limitations. Before then, the events are "raw". The part I would move into another crate is actually the clamping/rounding. Those are more about what " feels right".

Shatur:
I would consider `bevy_input` as a source for raw input and its *processing* being a higher-level abstraction. That's what I currently do with my crate (I use methods that return raw values and let user apply any modifiers on top of them).

We on the same page with @alice-i-cecile, this PR is good.
I just wanted to share my thoughts for possible future considerations.

BenjaminBrienen:
I'm choosing to not split up the unit test. It can be done as an improvement in a separate PR if someone would like. :)

Shatur (approved):

PR #17016, opened 2024-12-29
Overhaul bezier curve benchmarks

BD103:
# Objective

- Part of #16647.
- The benchmarks for bezier curves have several issues and do not yet use the new `bench!` naming scheme.

## Solution

- Make all `bevy_math` benchmarks use the `bench!` macro for their name.
- Delete the `build_accel_cubic()` benchmark, since it was an exact duplicate of `build_pos_cubic()`.
- Remove `collect::<Vec<_>>()` call in `build_pos_cubic()` and replace it with a `for` loop.
- Combine all of the benchmarks that measure `curve.position()` under a single group, `curve_position`, and extract the common bench routine into a helper function.
- Move the time calculation for the `curve.ease()` benchmark into the setup closure so it is not tracked.
- Rename the benchmarks to be more descriptive on what they do.
  - `easing_1000` -> `segment_ease`
  - `cubic_position_Vec2` -> `curve_position/vec2`
  - `cubic_position_Vec3A` -> `curve_position/vec3a`
  - `cubic_position_Vec3` -> `curve_position/vec3`
  - `build_pos_cubic_100_points` -> `curve_iter_positions`

## Testing

- `cargo test -p benches --bench math`
- `cargo bench -p benches --bench math`
  - Then open `./target/criterion/report/index.html` to see the report!

### DIFF
--- a/benches/benches/bevy_math/bezier.rs
+++ b/benches/benches/bevy_math/bezier.rs
@@ -1,10 +1,9 @@
 use benches::bench;
 use bevy_math::{prelude::*, VectorSpace};
-use criterion::{
-    criterion_group, measurement::Measurement, BatchSize, BenchmarkGroup, BenchmarkId,
-    Criterion,
-};
 use core::hint::black_box;
+use criterion::{
+    criterion_group, measurement::Measurement, BatchSize, BenchmarkGroup, BenchmarkId, Criterion,
+};
 
 criterion_group!(benches, segment_ease, curve_position, curve_iter_positions);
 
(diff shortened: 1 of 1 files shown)


BenjaminBrienen (approved):

IQuick143 (approved):
Not entirely familiar with the benchmarking APIs, but the changes all seem reasonable.

alice-i-cecile:
Thanks for formatting this for me: manual merge conflict resolution often causes that failure.

BD103:
> Thanks for formatting this for me: manual merge conflict resolution often causes that failure.

No problem!

PR #17025, opened 2024-12-29
fix visibility propagation during reparenting

### LINKED ISSUE #17024: Using .set_parent on an entity whose parent was Visibility:Hidden fails to refresh visibility
## Bevy version
0.15 

 I found a super annoying bug !!

```
   commands
                        .entity(*equipment_link_node)
                        .set_parent(*parent_entity);

                        //this doesnt refresh inherited visibility !
```

If you use set_parent and the entity had a 'false' inherited visibility, then the inherited visibility doesnt refresh!
what gives ? is this a bug in bevy or am i supposed to handle this
i think it is because of this 

```
fn visibility_propagate_system(
    changed: Query<
        (Entity, &Visibility, Option<&Parent>, Option<&Children>),
        (With<InheritedVisibility>, Changed<Visibility>),
    >,
    mut visibility_query: Query<(&Visibility, &mut InheritedVisibility)>,
    children_query: Query<&Children, (With<Visibility>, With<InheritedVisibility>)>,
) {
```

doesnt account for changed<parent>   ... ?   seems like it doesnt

ethereumdegen:
# Objective
Fixes #17024 

## Testing
 By adding 

```
if let Some(mut cmd) = commands.get_entity( *equipment_link_node ){
                         cmd.insert(Visibility::Inherited); // a hack for now 
                     }

```
in my build after .set_parent() , this fixes the issue.  This is why i think that this change will fix the issue.  Unfortunately i was not able to test the Changed (parent )  , this actual code change, because no matter how i 'patch', it breaks my project.  I got super close but still had 23 errors due to Reflect being angry.  

---

### DIFF
--- a/crates/bevy_render/src/view/visibility/mod.rs
+++ b/crates/bevy_render/src/view/visibility/mod.rs
@@ -757,6 +757,45 @@ mod test {
         );
     }
 
+
+
+    #[test]
+    fn test_visibility_propagation_on_parent_change() {
+        // Setup the world and schedule
+        let mut app = App::new();
+         
+        app.add_systems(Update,visibility_propagate_system);
+
+        // Create entities with visibility and hierarchy
+        let parent1 = app.world_mut().spawn((Visibility::Hidden,)).id();
+        let parent2 = app.world_mut().spawn((Visibility::Visible,)).id();
+        let child1 = app.world_mut().spawn((Visibility::Inherited,)).id();
+        let child2 = app.world_mut().spawn((Visibility::Inherited,)).id();
+
+        // Build hierarchy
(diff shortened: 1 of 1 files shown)


alice-i-cecile (approved):

alice-i-cecile (changes requested):
Can you please add a regression test for this?

ethereumdegen:
Alright @alice-i-cecile  i have added your test and it indeed fails when i use the original code and passes when i put my fix in .  !

alice-i-cecile (approved):

BenjaminBrienen (approved):
I'm fairly confident that this will work recursively, although the test doesn't check for that.

PR #17027, opened 2024-12-29
Add `portable-atomic` support to `bevy_utils` for `once!`

bushrat011899:
# Objective

- Improves platform compatibility for `bevy_utils`

## Solution

- Added `portable-atomic` to allow using the `once!` macro on more platforms (e.g., Raspberry Pi Pico)

## Testing

- CI

## Notes

- This change should be entirely hidden thanks to the use of `doc(hidden)`. Enabling the new `portable-atomic` feature just allows using the `once!` macro on platforms which previously could not.
- I took the liberty of updating the feature documentation to be more in line with how I've documented features in `bevy_ecs`/`bevy_app`/etc. for their `no_std` updates.

### DIFF
--- a/crates/bevy_utils/Cargo.toml
+++ b/crates/bevy_utils/Cargo.toml
@@ -10,11 +10,42 @@ keywords = ["bevy"]
 
 [features]
 default = ["std", "serde", "tracing"]
-std = ["alloc", "tracing?/std", "foldhash/std", "dep:thread_local"]
-alloc = ["hashbrown"]
-detailed_trace = []
+
+# Functionality
+
+## Adds serialization support through `serde`.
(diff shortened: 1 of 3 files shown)


alice-i-cecile (approved):
Very nice :) I particularly appreciate the docs on the features.

BenjaminBrienen:
I think eventually we should have this kind of code live in a `bevy_platform_adaptors` or whatever.

BenjaminBrienen (approved):

bushrat011899:
> I think eventually we should have this kind of code live in a `bevy_platform_adaptors` or whatever.

Totally agree, just wanted to sneak this change in as-is to unblock #17028.

PR #17031, opened 2024-12-30
Add `no_std` support to `bevy_window`

bushrat011899:
# Objective

- Contributes to #15460

## Solution

  - `std` (default)
  - `bevy_reflect` (default)
  - `libm`

## Testing

- CI

## Notes

- `bevy_reflect` was previously always enabled, which isn't how most other crates handle reflection. I've brought this in line with how most crates gate `bevy_reflect`. This is where the majority of the changes come from in this PR.


tools/ci/src/commands/compile_check_no_std.rs
```diff
@@ -134,6 +134,14 @@ impl Prepare for CompileCheckNoStdCommand {
             "Please fix compiler errors in output above for bevy_state no_std compatibility.",
         ));
 
+        commands.push(PreparedCommand::new::<Self>(
```
alice-i-cecile, line 137: We should probably start deduplicating this logic soon.

alice-i-cecile (approved):

tools/ci/src/commands/compile_check_no_std.rs
bushrat011899: Yeah @BD103 and I had a chat about this earlier today. I like the idea of moving this out into a `toml`, `csv` or some other "plain data" file format.

bushrat011899: I also think we're rapidly reaching a critical mass of functionality where we could replace this CI task with just an example crate instead.

BenjaminBrienen: It would be nice if this could be rolled up into a loop, but the macro requires a string literal, which means no format string.

TimJentzsch (approved):
I like the improved features documentation!

PR #17033, opened 2024-12-30
Overhaul picking benchmarks

BD103:
# Objective

- Part of #16647.
- This PR goes through our `ray_cast::ray_mesh_intersection()` benchmarks and overhauls them with more comments and better extensibility. The code is also a lot less duplicated!

## Solution

- Create a `Benchmarks` enum that describes all of the different kind of scenarios we want to benchmark.
- Merge all of our existing benchmark functions into a single one, `bench()`, which sets up the scenarios all at once.
- Add comments to `mesh_creation()` and `ptoxznorm()`, and move some lines around to be a bit clearer.
- Make the benchmarks use the new `bench!` macro, as part of #16647.
- Rename many functions and benchmarks to be clearer.

## For reviewers

I split this PR up into several, easier to digest commits. You might find it easier to review by looking through each commit, instead of the complete file changes.

None of my changes actually modifies the behavior of the benchmarks; they still track the exact same test cases. There shouldn't be significant changes in benchmark performance before and after this PR.

## Testing

List all picking benchmarks: `cargo bench -p benches --bench picking -- --list`

Run the benchmarks once in debug mode: `cargo test -p benches --bench picking`

Run the benchmarks and analyze their performance: `cargo bench -p benches --bench picking`

- Check out the generated HTML report in `./target/criterion/report/index.html` once you're done!

---

## Showcase

List of all picking benchmarks, after having been renamed:

image

Example report for `picking::ray_mesh_intersection::cull_intersect/100_vertices`:

image

### DIFF
--- a/benches/benches/bevy_picking/ray_mesh_intersection.rs
+++ b/benches/benches/bevy_picking/ray_mesh_intersection.rs
@@ -64,8 +64,13 @@ fn create_mesh(vertices_per_side: u32) -> SimpleMesh {
 /// An enum that represents the configuration for all variations of the ray mesh intersection
 /// benchmarks.
 enum Benchmarks {
+    /// The ray intersects the mesh, and culling is enabled.
     CullIntersect,
+
+    /// The ray intersects the mesh, and culling is disabled.
     NoCullIntersect,
+
+    /// The ray does not intersect the mesh, and culling is enabled.
     CullNoIntersect,
 }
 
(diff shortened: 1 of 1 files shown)


BD103:
cc @aevyrie, since you're the person who I know is most familiar with picking, and most likely to catch any mistakes in the documentation!

cc @mockersf, since you wrote the original version of these benchmarks in https://github.com/aevyrie/bevy_mod_raycast/pull/17.

No pressure if neither of you want to review, just thought you might be interested!

alice-i-cecile (approved):

BenjaminBrienen (approved):

alice-i-cecile:
@BD103 what do you think still needs to be done before we merge this? Looks like the CI failure is just `cargo fmt`.

BD103:
> @BD103 what do you think still needs to be done before we merge this? Looks like the CI failure is just `cargo fmt`.

I also wanted to add one more test case to verify that the ray does / doesn't intersect the mesh when `cargo test --bench picking` was run, but it's ready now!

BenjaminBrienen:
Sometimes I use the mobile app to put the Waiting label on my own PRs so that I can remember to fix a merge conflict or a CI check when I get to my laptop. It looks like they're doing the same thing to me.

PR #15184, opened 2024-09-13
Support systems that take references as input

ItsDoot:
# Objective

- Fixes #14924
- Closes #9584

## Solution

- We introduce a new trait, `SystemInput`, that serves as a type function from the `'static` form of the input, to its lifetime'd version, similarly to `SystemParam` or `WorldQuery`.
- System functions now take the lifetime'd wrapped version, `SystemInput::Param<'_>`, which prevents the issue presented in #14924 (i.e. `InRef<T>`).
- Functions for running systems now take the lifetime'd unwrapped version, `SystemInput::Inner<'_>` (i.e. `&T`).
- Due to the above change, system piping had to be re-implemented as a standalone type, rather than `CombinatorSystem` as it was previously.
- Removes the `Trigger<'static, E, B>` transmute in observer runner code.

## Testing

- All current tests pass.
- Added additional tests and doc-tests.

---

## Showcase

```rust
let mut world = World::new();

let mut value = 2;

// Currently possible:
fn square(In(input): In<usize>) -> usize {
    input * input
}
value = world.run_system_once_with(value, square);

// Now possible:
fn square_mut(InMut(input): InMut<usize>) {
    *input *= *input;
}
world.run_system_once_with(&mut value, square_mut);

// Or:
fn square_ref(InRef(input): InRef<usize>) -> usize {
    *input * *input
}
value = world.run_system_once_with(&value, square_ref);
```

## Migration Guide

- All current explicit usages of the following types must be changed in the way specified:
    - `SystemId<I, O>` to `SystemId<In<I>, O>`
    - `System<In = T>` to `System<In = In<T>>`
    - `IntoSystem<I, O, M>` to `IntoSystem<In<I>, O, M>`
    - `Condition<M, T>` to `Condition<M, In<T>>`
- `In<Trigger<E, B>>` is no longer a valid input parameter type. Use `Trigger<E, B>` directly, instead.

PR #17034, opened 2024-12-30
Make 8 methods public and updated input parameter generics for SystemState::build_system_with_input

Neo-Zhixing:
# Objective

- Made certain methods public for advanced use cases. Methods that returns mutable references are marked as unsafe due to the possibility of violating internal lifetime constraint assumptions.
- Fixes an issue introduced by #15184

### DIFF
--- a/crates/bevy_ecs/src/system/function_system.rs
+++ b/crates/bevy_ecs/src/system/function_system.rs
@@ -476,7 +476,7 @@ impl<Param: SystemParam> SystemState<Param> {
 
     /// Gets the metadata for this instance.
     #[inline]
-    pub unsafe fn meta_mut(&mut self) -> &mut SystemMeta {
+    pub fn meta_mut(&mut self) -> &mut SystemMeta {
         &mut self.meta
     }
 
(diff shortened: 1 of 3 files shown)


crates/bevy_ecs/src/system/function_system.rs
```diff
@@ -396,7 +396,7 @@ macro_rules! impl_build_system {
                 Input: SystemInput,
                 Out: 'static,
                 Marker,
-                F: FnMut(In<Input>, $(SystemParamItem<$param>),*) -> Out
+                F: FnMut(Input, $(SystemParamItem<$param>),*) -> Out
```
Neo-Zhixing, line 399: After #15184, the input parameter no longer have to be wrapped in a `In`

crates/bevy_ecs/src/system/function_system.rs
```diff
@@ -644,6 +650,22 @@ impl<Param: SystemParam> SystemState<Param> {
...
+        &self.param_state
+    }
+
+    /// Returns a mutable reference to the current system param states.
+    /// Marked as unsafe because modifying the system states may result in violation to certain
+    /// assumptions made by the [`SystemParam`]. Use with care.
+    /// 
+    /// Safety: Modifying the system param states may have unintented consequences.
+    /// the param state is generally considered to be owned by the [`SystemParam`]. Modifications
+    /// should respect any invariants as required by the [`SystemParam`].
```
ItsDoot: It might be helpful to provide an example of when this might be the case, as I'm not sure why it needs to be unsafe.

Neo-Zhixing: For example, modifying the system state of `ResMut` without also updating [`SystemMeta::component_access_set`] will obviously create issues.

alice-i-cecile (approved):

crates/bevy_ecs/src/system/function_system.rs
alice-i-cecile: Interesting. Can we remove `In` completely in a follow-up then?

PR #14748, opened 2024-08-15
Unshorten lifetimes on Query getter functions

ItsDoot:
# Objective

Lifetimes on the `Query` getter functions are unnecessarily shortened, when they could be returning `'w`. They could be lengthened since the `QueryState` methods return `'w`:

https://github.com/bevyengine/bevy/blob/fe94728d0c5de493d3cd8f404b5bdaf84360bdd1/crates/bevy_ecs/src/query/state.rs#L915-L921

This is currently preventing me from implementing a custom `SystemParam` that uses a `Query`: https://gist.github.com/ItsDoot/3727d6b219ba22a2f9b6865762905943

```
error[E0499]: cannot borrow `self.query` as mutable more than once at a time
  --> src\ecs.rs:71:19
   |
55 |     pub fn get(&mut self, start: Entity) -> Option<Mut<T>> {
   |                - let's call the lifetime of this reference `'1`
...
71 |             match self.query.get_mut(current) {
   |                   ^^^^^^^^^^ `self.query` was mutably borrowed here in the previous iteration of the loop
...
75 |                     return Some(component);
   |                            --------------- returning this value requires that `self.query` is borrowed for `'1`
```

## Solution

Lengthen them to `'w`.

PR #17035, opened 2024-12-30
Update return items lifetime for Query to capture 'world only

Neo-Zhixing:
# Objective

- When a system has a return value with lifetimes, allow the system return value to capture `'world` instead of `'world + 'state` when working with query parameters

## Solution

- Items returned from a `Query` do not actually reference the internals of a `QueryState`. As such, it appears safe to annotate these methods to return `'world` lifetime without capturing `'state`.


ItsDoot:
This violates memory safety rules unfortunately as I found out a few months ago: #14748

ItsDoot:
Additional context on the discord here: https://discord.com/channels/691052431525675048/749335865876021248/1273440592554102835

Neo-Zhixing:
Ok so sounds like the solution is to use Query::into_inner to consume the query instead.

PR #17036, opened 2024-12-30
System Yielding

PixelDust22:
# Objective

- In some cases, systems may not be immediately ready for execution due to unmet constraints. For example, the execution of a certain system may be dependent upon the completion of an asynchronous GPU submission. 
- In those cases, we may not want to skip the execution of this system for the current frame altogether. Instead, we would like to **defer** the execution of this system without violating system dependencies. We schedule all other systems first, and once we're done, we come back and pick up where we left off.
- This is also particularly useful for the scheduling of **async systems**. We can create a system wrapper that wraps an async system. Each time the system gets scheduled, we call `poll` on the future. And we don't mark the system as completed until the future returns `Poll::Ready`.

## Solution

- Added a new method, `yielded`, on the `System` trait. If `yielded` returns true, the scheduler defer the execution of the system until all other currently ready systems have completed execution.
- Maybe `pending` is a better name for this method?

## Testing

Added new unit tests. Also tested as a part of a custom implementation of a Vulkan-based rendering backend that returns GPUFutures.

## Draft Release Notes

System Deferral: Introduced a new mechanism in the ECS scheduler to defer the execution of systems. This allows systems to temporarily defer their execution by allowing the scheduler to execute other systems first. This is useful in cases where the execution of a system is dependent upon the completion of an asynchronous operation. It is also useful for implementing asynchronous systems.

### DIFF
--- a/crates/bevy_ecs/src/schedule/executor/mod.rs
+++ b/crates/bevy_ecs/src/schedule/executor/mod.rs
@@ -386,4 +386,92 @@ mod tests {
         assert!(world.get_resource::<R1>().is_none());
         assert!(world.get_resource::<R2>().is_none());
     }
+
+    #[test]
+    fn system_yielding() {
+        for executor in [
+            ExecutorKind::Simple,
+            ExecutorKind::SingleThreaded,
+            ExecutorKind::MultiThreaded,
+        ] {
+            system_yielding_core(executor);
+        }
+    }
+
+    fn system_yielding_core(executor: ExecutorKind) {
+        use crate::system::{System, SystemMeta};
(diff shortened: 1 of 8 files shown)


mrchantey:
Awesome! Great to see the building blocks for async systems being put into place.
I guess the provided test demonstrates a low-level implementation of yielding, and implementing `IntoSystemConfigs` for systems that return futures (or generators?) would be a seperate pr?

Neo-Zhixing:
Implementing IntoSystemConfigs for systems that return futures turns out to be a bit difficult due to #17037

PR #11184, opened 2024-01-02
Nward/commands error handling

EngoDev:
# Objective
Fixes: #2004

Original PR: https://github.com/bevyengine/bevy/pull/2241
Thank you very much @NathanSWard for all your hard work! ⭐ I mostly just updated their code to work with the current version of bevy.

Also this is my first bevy PR, feedback is very much appreciated :)  

### Added

- `FallibleCommand`
    - this is for command that can potentially fail, and allow people to provide error handlers if an error occurs
  
- `CommandErrorHandling`
    - this is for function like types that can handle command errors.
- `CommandErrorHandler`
    -  Contains handlers for typical use cases of handling errors like:
        - `ignore`
        - `panic`
        - `log` - This is the default behavior instead of `panic`. The error will be logged via `error!`.

### Changed
If a command is fallible, then new return type is a `Config` type.
The `Config` type allows a user to optionally provide an error handler via `on_err`.

### Example
```rust
#[derive(Component)]
struct Marker(u32);

#[derive(Resource)]
struct Food(u32);

fn system(mut commands: Commands) {
    commands.entity(e)
        .insert(Marker(32))
        .on_err(|error, ctx| { /*..*/ });

    command
        .remove_resource::<Food>()
        .on_err(CommandErrorHandler::ignore);

    commands
        .entity(e)
        .despawn()
        .on_err(CommandErrorHandler::panic);

    commands
        .entity(e)
        .insert(Marker(32))
        .insert(Marker(64)); // you can also ignore the error handling and just chain regular commands
}
```

PR #17043, opened 2024-12-30
Add `Result` handling to `Commands` and `EntityCommands`

### LINKED ISSUE #2004: Configurable error handling in commands

### LINKED ISSUE #3845: EntityCommands should have methods to fail gracefully

### LINKED ISSUE #7118: Potential surprising "entity does not exist" panic due to asynchronous command execution

### LINKED ISSUE #10166: Commands should be infallible whenever possible

JaySpruce:
## Objective

Fixes #2004
Fixes #3845
Fixes #7118
Fixes #10166

## Solution

- The crux of this PR is the new `Command::with_error_handling` method. This wraps the relevant command in another command that, when applied, will apply the original command and handle any resulting errors.
- To enable this, `Command::apply` and `EntityCommand::apply` now return `Result`.
- `Command::with_error_handling` takes as a parameter an error handler of the form `fn(&mut World, CommandError)`, which it passes the error to.
  - `CommandError` is an enum that can be either `NoSuchEntity(Entity)` or `CommandFailed(Box<dyn Error>)`.

### Closures
- Closure commands can now optionally return `Result`, which will be passed to `with_error_handling`.

### Commands
- Fallible commands can be queued with `Commands::queue_fallible` and `Commands::queue_fallible_with`, which call `with_error_handling` before queuing them (using `Commands::queue` will queue them without error handling).
- `Commands::queue_fallible_with` takes an `error_handler` parameter, which will be used by `with_error_handling` instead of a command's default.
- The `command` submodule provides unqueued forms of built-in fallible commands so that you can use them with `queue_fallible_with`.
- There is also an `error_handler` submodule that provides simple error handlers for convenience.

### Entity Commands
- `EntityCommand` now automatically checks if the entity exists before executing the command, and returns `NoSuchEntity` if it doesn't.
- Since all entity commands might need to return an error, they are always queued with error handling.
- `EntityCommands::queue_with` takes an `error_handler` parameter, which will be used by `with_error_handling` instead of a command's default.
- The `entity_command` submodule provides unqueued forms of built-in entity commands so that you can use them with `queue_with`.

### Defaults
- In the future, commands should all fail according to the global error handling setting. That doesn't exist yet though.
- For this PR, commands all fail the way they do on `main`.
- Both now and in the future, the defaults can be overridden by `Commands::override_error_handler` (or equivalent methods on `EntityCommands` and `EntityEntryCommands`).
- `override_error_handler` takes an error handler (`fn(&mut World, CommandError)`) and passes it to every subsequent command queued with `Commands::queue_fallible` or `EntityCommands::queue`.
- The `_with` variants of the queue methods will still provide an error handler directly to the command.
- An override can be reset with `reset_error_handler`.

## Future Work

- After a universal error handling mode is added, we can change all commands to fail that way by default.
  - Once we have all commands failing the same way (which would require either the full removal of `try` variants or just making them useless while they're deprecated), `queue_fallible_with_default` could be removed, since its only purpose is to enable commands having different defaults.


JaySpruce:
Pinging @alice-i-cecile as requested. Not totally polished, but I felt like this could use some feedback/cooperation in regards to other fallible stuff going on.

crates/bevy_ecs/src/system/commands/mod.rs
```diff
@@ -1767,7 +1833,52 @@ impl<'a> EntityCommands<'a> {
...
+        &mut self,
+        command: impl EntityCommand<M>,
+        default_error_mode: CommandErrorMode,
+    ) -> &mut Self {
+        self.commands
+            .queue_fallible_with_error_mode(command.with_entity(self.entity), default_error_mode);
+        self
+    }
+
+    ///
```
alice-i-cecile, line 1855: Reminder to finish these doc strings before we merge :)

crates/bevy_ecs/src/system/commands/mod.rs
```diff
@@ -2219,54 +2407,36 @@ where
...
+/// An [`EntityCommand`] that despawns a specific entity.
 ///
 /// # Note
 ///
 /// This won't clean up external references to the entity (such as parent-child relationships
 /// if you're using `bevy_hierarchy`), which may leave the world in an invalid state.
 #[track_caller]
-fn despawn(log_warning: bool) -> impl EntityCommand {
+fn despawn() -> impl EntityCommand<World> {
+    #[cfg(feature = "track_change_detection")]
```
alice-i-cecile, line 2418: This change seems correct regardless, maybe we should split this out? Ditto the docs above.

crates/bevy_ecs/src/world/error.rs
```diff
@@ -25,60 +23,12 @@ pub enum EntityComponentError {
 }
 
 /// An error that occurs when fetching entities mutably from a world.
-#[derive(Clone, Copy)]
-pub enum EntityFetchError<'w> {
+#[derive(Error, Debug, Clone, Copy, PartialEq, Eq)]
+pub enum EntityFetchError {
```
alice-i-cecile, line 27: Wow I much prefer removing the lifetime + UnsafeWorldCell from this. Can we split this out?

crates/bevy_ecs/src/world/mod.rs
```diff
@@ -96,13 +97,44 @@ use unsafe_world_cell::{UnsafeEntityCell, UnsafeWorldCell};
 ///     commands.queue(AddToCounter(42));
 /// }
 /// ```
-pub trait Command: Send + 'static {
+pub trait Command<Marker = ()>: Send + 'static {
```
alice-i-cecile, line 100: Docs on why the Marker generic is required please.

crates/bevy_ecs/src/world/mod.rs
```diff
@@ -96,13 +97,44 @@ use unsafe_world_cell::{UnsafeEntityCell, UnsafeWorldCell};
...
+                    CommandErrorMode::Panic => panic!("Command returned an error: {error}"),
+                }
+            }
+        }
+    }
+}
+
+/// How a [`Command`] should respond upon encountering an [`Error`](crate::result::Error).
+#[derive(Clone, Copy, PartialEq, Eq)]
+pub enum CommandErrorMode {
```
alice-i-cecile, line 131: I think that we need to be able to provide a closure here as well, for actually responding to the error in bespoke ways. This is relatively niche, but an important requirement. Unfortunately, I think we need to add that in the initial PR, since it's the most complex technical requirement and I can't confidently say whether or not this approach will handle it just by looking.

crates/bevy_ecs/src/world/mod.rs
```diff
@@ -96,13 +97,44 @@ use unsafe_world_cell::{UnsafeEntityCell, UnsafeWorldCell};
...
+    /// Fail silently.
+    Silent,
+    /// Send a warning.
+    Warn,
```
alice-i-cecile, line 135: Follow-up: we should add an error level too.

alice-i-cecile:
I'm willing to eat a 5% (worst case) performance regression on Commands here. This is a huge usability problem with Bevy, and I'm convinced that we can't solve this without somehow regressing perf there. The big performance gains to be found here are in command batching anyways.

NthTensor:
I'd like to do a proper review of this, if I can find the time. Writing this as a reminder to myself.

crates/bevy_ecs/src/world/mod.rs
JaySpruce: I got it working with `fn()`, doing it with `FnOnce()` would be more complicated. Is `fn()` enough or should I keep working on it?

alice-i-cecile: That's enough :)

crates/bevy_ecs/src/world/mod.rs
```diff
@@ -96,13 +97,52 @@ use unsafe_world_cell::{UnsafeEntityCell, UnsafeWorldCell};
...
+    /// The command will send a warning.
+    Warn,
+    /// The command will send an error.
+    Error,
+    /// The command will start a panic.
+    Panic,
+    /// The command will call a custom function.
+    ///
+    /// `fn()` can be a closure if it doesn't capture its environment.
+    Override(fn()),
```
alice-i-cecile, line 145: In order for this to be useful, we need to pass in both the `Result` (or at least the error arm) and `&mut World` to actually respond. Without any data access this is only really viable for custom logging.

EngoDev:
> @EngoDev can I get your opinions on this work? I think I prefer the basic idea here over that of #11184. Making commands fallible by default (with commands that can't fail always returning `Ok(())`) feels like the right direction.

I agree with you, this approach makes a lot of sense. I do think we can find a more optimized solution though. Having proper error handling for commands is worth the 5% decrease in performance but I'm positive we can minimize it.

I'll try to find time over the weekend to play around with this PR and give more based feedback.

crates/bevy_ecs/src/world/mod.rs
```diff
@@ -86,23 +87,79 @@ use unsafe_world_cell::{UnsafeEntityCell, UnsafeWorldCell};
...
+    /// Upon encountering an error, the command will send an error message with [`error!`].
+    Error,
+    /// Upon encountering an error, the command will start a panic.
+    Panic,
+    /// Upon encountering an error, the command will call a custom function,
+    /// passing in its `&mut World` access and the error.
+    ///
+    /// `fn()` can be a closure if it doesn't capture its environment.
+    ///
+    /// If `CommandErrorMode` is used in a situation without `&mut World` access
```
alice-i-cecile, line 160: This seems like an odd situation. What methods would try to resolve and handle commands without exclusive world access?

crates/bevy_ecs/src/system/commands/mod.rs
```diff
@@ -444,23 +457,25 @@ impl<'w, 's> Commands<'w, 's> {
...
+        if !self.entities.contains(entity) {
+            match self.error_mode.unwrap_or(CommandErrorMode::Panic) {
+                CommandErrorMode::Silent => (),
+                CommandErrorMode::Warning => warn!(
+                    "Attempted to create an EntityCommands for entity {entity}, which does not exist; returned invalid EntityCommands"
+                ),
+                CommandErrorMode::Error => error!(
+                    "Attempted to create an EntityCommands for entity {entity}, which does not exist; returned invalid EntityCommands"
+                ),
+                CommandErrorMode::Panic | CommandErrorMode::Custom(_) => panic!(
```
alice-i-cecile, line 469: Aha, here's the panic. This should be called out more clearly at least, but I think that passing down a more complex Result or adding an error arm is the better way. The custom error handler should be able to know whether or not the entity failed to exist at all, and respond accordingly.

crates/bevy_ecs/src/system/commands/mod.rs
```diff
@@ -606,6 +626,116 @@ impl<'w, 's> Commands<'w, 's> {
...
+        self.error_mode = Some(CommandErrorMode::Panic);
+    }
+
+    /// Sets the [`EntityCommands`] instance call a custom function upon encountering an error.
+    ///
+    /// You can use [`Self::reset_error_mode`] to undo this.
+    ///
+    /// `fn()` can be a closure if it doesn't capture its environment.
+    ///
+    /// If an error is encountered in a situation without `&mut World` access
```
alice-i-cecile, line 713: Same note here about docs / alternative fix for this problem.

crates/bevy_ecs/src/world/mod.rs
```diff
@@ -86,23 +87,79 @@ use unsafe_world_cell::{UnsafeEntityCell, UnsafeWorldCell};
...
+    /// (e.g. a non-deferred [`Commands`] method), this mode will instead panic.
+    Custom(fn(&mut World, Result)),
```
alice-i-cecile, line 162: I thought about this a bit more: this should actually have a `Box<dyn Error>`, not a `Result`. There's no need for error handling if everything went okay, so we can save our users some repetitive matching.

Then, combined with the ideas about how `commands.entity(my_entity)` shouldn't panic, we get:

```rust
enum CommandError {
    NoSuchEntity(Entity),
    CommandFailed(Box<dyn Error>)
}
```

which we pass to the custom error handlers to give them the context needed to solve the problem.

crates/bevy_ecs/src/system/commands/mod.rs
```diff
@@ -81,6 +86,10 @@ pub use parallel_scope::*;
 pub struct Commands<'w, 's> {
     queue: InternalQueue<'s>,
     entities: &'w Entities,
+    /// How the [`Commands`] instance will response upon encountering an error.
+    ///
+    /// If this is unset ([`None`]), commands can choose their error mode individually.
```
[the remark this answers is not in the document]
NthTensor: I agree with this assessment. I'll add that currently the global handling mode is effectively based on `Box<dyn Error>` and always panics.

crates/bevy_ecs/src/system/commands/mod.rs
NthTensor: I was kind of hoping we wouldn't have to put handling logic on this level. I feel like this sort of match statement should ideally only appear once, and everything should be funneled through it on a lower level. That's just a feeling on my part.

NthTensor:
I like this direction a lot, but I do have a gripe with `CommandErrorMode`. It seems to me like this should not be an enum, but rather just make everything an `Option<fn(&mut World, Box<dyn Error>)>`. Then we can have a hot path where it is `None` and the errors bubble up to the handler embedded in the schedule (through a fallible `ApplyDeferred`), and we can have a `#[cold]` path that calls the function as part of command application.

You have `queue_fallible` insert the `None` and something like a `queue_fallible_with(command, panic_on_error)` where `panic_on_error` is a normal function.

These are all suggestions mind you. You're doing great work, and I wouldn't block the current line of work over any of this. It just strikes me that using function pointers might be more extensible.

crates/bevy_ecs/src/system/commands/mod.rs
```diff
@@ -444,23 +457,25 @@ impl<'w, 's> Commands<'w, 's> {
     #[inline]
     #[track_caller]
     pub fn entity(&mut self, entity: Entity) -> EntityCommands {
-        #[inline(never)]
-        #[cold]
-        #[track_caller]
```
hymm, line 449: Some of the perf regressions are probably from losing these annotations. We should try to add them back.

crates/bevy_ecs/src/system/commands/mod.rs
hymm: Do we need to remove the panic here? I feel like if the user doesn't want it to panic they should just be using `get_entity`.

NthTensor: Eventually I think the plan is to collapse our panicking and non-panicking variants into the same call.

crates/bevy_ecs/src/system/commands/mod.rs
NthTensor: Good catch, that would do it.

crates/bevy_ecs/src/system/commands/mod.rs
hymm: I think that would need the `impl EntityCommand for impl FnOnce(EntityWorldMut)` changes. I do think that should be pulled out of this pr as that is something that could be considered separately.

crates/bevy_ecs/src/system/commands/mod.rs
alice-i-cecile: Yeah, I'd like to remove `get_entity` completely in favor of this error handling technique. It's frustrating that there's so many different edge cases to guard for, and no unified way to make sure you've handled them all.

crates/bevy_ecs/src/system/commands/mod.rs
JaySpruce: I did smaller benchmarks before and after making that change (only the `insert_commands` benchmark, because that had been the most volatile) and didn't see any difference. We're aiming to remove the panic from `entity` anyway, I'm not sure there's a place for those

crates/bevy_ecs/src/system/commands/mod.rs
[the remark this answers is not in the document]
JaySpruce: It's the only way I could think for it to work without vastly changing how commands work. Calling `queue` *immediately* turns the command into a bunch of unknown bytes, so if you want to customize the command, it has to know what you want *before* `queue` is called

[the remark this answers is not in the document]
alice-i-cecile: Alright, I'm fine with this for now. Leave a comment to this effect on this field please :)

crates/bevy_ecs/src/system/commands/mod.rs
hymm: That makes sense, but I still feel like this pr should be focused on the deferred errors. Or are you saying that creating a EntityCommands will only panic when the command queue is applied?

crates/bevy_ecs/src/system/commands/mod.rs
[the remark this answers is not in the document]
JaySpruce: Actually, here's an alternative: we could re-add those helper functions we just removed that return the commands as closures, and we could make them `pub` to let people queue built-in commands with their own error handlers:
`commands.entity(entity).queue_with(insert(bundle), |world, error| { /* whatever */ });`

crates/bevy_ecs/src/system/commands/mod.rs
alice-i-cecile: > That makes sense, but I still feel like this pr should be focused on the deferred errors

I agree with that, but I think we can do some future-proofing here. It's fine to leave it for follow-up though: I just think it's a bit messy as is.

hymm:
Hmm, I'm trying to figure out what would even cause regressions here if it's not the #[cold] annotation. I feel like there hasn't been much change to the code paths for fake_commands and sized_commands_0_bytes. The only thing I really see is that we now need to drop the result from `command.apply`. Not sure what else it could be.

https://github.com/bevyengine/bevy/pull/17043/files#diff-badf600e116f81aa58524486dea99ee4bb337143cca99a57339dda74f4661bc4R190

JaySpruce:
I think I'll revert `entity` anyway, handling the error there doesn't really fit and changing it to return `Result` would make a massive diff (344 references across 144 files...) which ought to be its own PR

JaySpruce:
How would y'all feel about something like this:

```
pub mod handler {
	pub fn log_warning() -> fn(&mut World, Box<dyn Error>) {
		|_, error| warn!("{error}")
	}
	[more of them]
}

pub mod command {
	pub fn insert(impl Bundle) -> impl EntityCommand {
		move |EntityWorldMut| {
			/*command code*/
		}
	}
	[more of them]
}
```

So that users could do this if they wanted:
`commands.entity(entity).queue_with(command::insert(bundle), handler::log_warning());`

I'm mostly unsure what the stance is on using `pub mod` like namespaces.

JaySpruce:
Performance somehow got better? I guess reverting `Commands::entity` might've helped, but that doesn't explain the improvements over main:

<details>
<summary>Benchmarks</summary>

with_commanderror_1
with_commanderror_2
with_commanderror_3

</details>

NthTensor:
Well that's very encouraging. I'll see if I can replicate.

crates/bevy_ecs/src/system/commands/mod.rs
```diff
@@ -606,6 +623,65 @@ impl<'w, 's> Commands<'w, 's> {
         }
     }
 
+    /// Pushes a generic [`Command`] to the command queue with error handling.
+    ///
+    /// If the command returns an error, it will be passed to the global error handler
+    /// (or the [`Commands`] instance's error handler override, if set).
```
alice-i-cecile, line 629: This should link to the docs about how to set the error handler.

crates/bevy_ecs/src/system/commands/mod.rs
```diff
@@ -81,6 +84,12 @@ pub use parallel_scope::*;
 pub struct Commands<'w, 's> {
```
alice-i-cecile, line 84: We  need docs on `Commands` about how to set the error handler for the specific instance of this.

crates/bevy_ecs/src/system/commands/mod.rs
```diff
@@ -1744,19 +1820,29 @@ impl<'a> EntityCommands<'a> {
...
+        self.queue_with_default(entity_command::despawn(), handler::silent());
     }
 
     /// Pushes an [`EntityCommand`] to the queue, which will get executed for the current [`Entity`].
     ///
+    /// All entity commands are fallible, because they must return an error if the entity
+    /// doesn't exist when a command is executed. Therefore, all entity commands are
+    /// queued with error handling.
+    ///
+    /// If the command returns an error, it will be passed to the global error handler
```
alice-i-cecile, line 1841: We also need docs on how to set the `EntityCommands` error handler.

crates/bevy_ecs/src/system/commands/mod.rs
```diff
@@ -2157,26 +2322,364 @@ impl<F> Command for F
...
+    }
+}
+
+/// Convenience functions that return simple error handlers for use with the following methods:
+/// - [`Commands::queue_fallible_with`](super::Commands::queue_fallible_with)
+/// - [`Commands::override_error_handler`](super::Commands::override_error_handler)
+/// - [`EntityCommands::queue_with`](super::EntityCommands::queue_with)
+/// - [`EntityCommands::override_error_handler`](super::EntityCommands::override_error_handler)
+/// - [`EntityEntryCommands::override_error_handler`](super::EntityEntryCommands::override_error_handler)
+pub mod handler {
```
alice-i-cecile, line 2403: Move this module to its own file please: this file is *way* too long already.

crates/bevy_ecs/src/system/commands/mod.rs
```diff
@@ -2157,26 +2322,364 @@ impl<F> Command for F
...
+    /// the panic message.
+    pub fn panic() -> fn(&mut World, CommandError) {
+        |_, error| panic!("{error}")
+    }
+}
+
+/// Convenience functions that return fallible commands for use with
+/// [`Commands::queue_fallible_with`](super::Commands::queue_fallible_with)
+/// to customize how the commands should handle errors.
+pub mod fallible_command {
```
alice-i-cecile, line 2432: The organization feels weird here. Why would users be searching for "fallible commands"?

crates/bevy_ecs/src/system/commands/mod.rs
```diff
@@ -2157,26 +2322,364 @@ impl<F> Command for F
...
+    }
+}
+
+/// Convenience functions that return entity commands for use with
+/// [`EntityCommands::queue_with`](super::EntityCommands::queue_with)
+/// to customize how the commands should handle errors.
+///
+/// All built-in entity commands are fallible. They will check if the entity exists
+/// at the time of execution and will return an error if it does not.
+pub mod entity_command {
```
alice-i-cecile, line 2534: Ditto on moving to its own file.

alice-i-cecile (changes requested):
Alright, this is really close. I'm happy with the core mechanism here, and the PR's diff is clean.

Two things to improve, ideally before merging. If you're busy, let me know and I'll PR to your branch.

1. This needs better documentation. There's several layers of command error handling here, which should be clearly laid out and tested via doc tests. I think module docs are the right place for this.
2. commands/mod.rs is way too big: we should try and split things out from it when possible.

crates/bevy_ecs/src/system/commands/mod.rs
JaySpruce: All in the same file, or should each module have its own? If they're all in the same, any recommendations for the name?

crates/bevy_ecs/src/system/commands/mod.rs
JaySpruce: I felt weird about calling it `command` if it only has fallible commands in it, and I also felt weird about moving non-fallible commands into it since it'd bloat the file even more. Since we're moving it to its own file though, might as well do both of those.

crates/bevy_ecs/src/system/commands/mod.rs
alice-i-cecile: Each module should have its own please :)

crates/bevy_ecs/src/system/commands/mod.rs
alice-i-cecile: Perfect :)

crates/bevy_ecs/src/system/commands/mod.rs
JaySpruce: Should I move the `Command` and `EntityCommand` traits into those modules too?

`Command` was previously moved to `world/mod.rs` with the reasoning that it's a `World` mutation and could be used separately from `Commands`. I think it's kinda weird for it to be so far from other command stuff, and these modules could be a nice place for it.

alice-i-cecile: Yes please :)

alice-i-cecile (approved):
I love the new module organization and the docs are straightforward and clear. Thank you so much; you've done a great job with this complex and important work!

crates/bevy_ecs/src/system/commands/command.rs
```diff
@@ -0,0 +1,345 @@
...
+            Err(error) => Err(CommandError::CommandFailed(error)),
+        }
+    }
+
+    /// Returns a new [`Command`] that, when applied, will apply the original command
+    /// and pass any resulting error to the provided `error_handler`.
+    fn with_error_handling(
+        self,
+        error_handler: Option<fn(&mut World, CommandError)>,
+    ) -> impl FnOnce(&mut World) + Send + 'static
```
LikeLakers2, line 87: This method strikes me as the sort of provided method which should return a wrapper struct, rather than an `impl Trait`, like how `Iterator`'s provided methods do.

Even if we don't want to do a wrapper struct, the return type should be `impl Command`, to line up with the documentation saying "Returns a new `Command`".

NthTensor (approved):
I don't have time to really look this over properly, so take this approval as you will. But this seems to be pretty much exactly what I was hoping for.

crates/bevy_ecs/src/system/commands/entity_command.rs
```diff
@@ -0,0 +1,329 @@
...
+    fn apply(self, entity: Entity, world: &mut World) -> Result;
+
+    /// Returns a [`Command`] which executes this [`EntityCommand`] for the given [`Entity`].
+    ///
+    /// This method is called when adding an [`EntityCommand`] to a command queue via [`Commands`](crate::system::Commands).
+    /// You can override the provided implementation if you can return a `Command` with a smaller memory
+    /// footprint than `(Entity, Self)`.
+    /// In most cases the provided implementation is sufficient.
+    #[must_use = "commands do nothing unless applied to a `World`"]
+    fn with_entity(self, entity: Entity) -> impl Command<(Result, CommandError)>
```
LikeLakers2, line 103: Same here as in my other comment - this strikes me as the sort of method which should probably return a wrapper struct (i.e. `WithEntity<Self>`).

LikeLakers2 (changes requested):
I skimmed through all the changes, and I only have a couple changes to suggest.

That said, you can otherwise consider this an approval.

crates/bevy_ecs/src/system/commands/command.rs
```diff
@@ -0,0 +1,345 @@
...
+        SpawnBatchIter::new(
+            world,
+            bundles_iter.into_iter(),
+            #[cfg(feature = "track_location")]
+            caller,
+        );
+    }
+}
+
+/// A [`Command`] that consumes an iterator to add a series of [`Bundles`](Bundle) to a set of entities.
```
LikeLakers2, line 157:
```suggestion
/// A [`Command`] that consumes an iterator to add a series of [`Bundle`]s to a set of entities.
```

crates/bevy_ecs/src/system/commands/command.rs
```diff
@@ -0,0 +1,345 @@
...
+        world.insert_batch_with_caller(
+            batch,
+            mode,
+            #[cfg(feature = "track_location")]
+            caller,
+        );
+    }
+}
+
+/// A [`Command`] that consumes an iterator to add a series of [`Bundles`](Bundle) to a set of entities.
```
LikeLakers2, line 179:
```suggestion
/// A [`Command`] that consumes an iterator to add a series of [`Bundle`]s to a set of entities.
```

LikeLakers2 (changes requested):
I looked back over this, looking specifically at docs (since I mostly looked at the API structure on my other review), and found a couple small docs improvements that could be made.

cBournhonesque (approved):
I feel like we could add tests for the CommandOverride, or to have more examples on how the ErrorHandling would be used in practice. 
(can be done in separate PRs as this is already huge)

I like the overall direction!

crates/bevy_ecs/src/system/commands/command.rs
```diff
@@ -0,0 +1,345 @@
...
+where
+    F: FnOnce(&mut World) + Send + 'static,
+{
+    fn apply(self, world: &mut World) -> Result {
+        self(world);
+        Ok(())
+    }
+}
+
+impl<F> Command<Result> for F
```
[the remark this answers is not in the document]
JaySpruce: Nope, you would need to box it. Pretty sure `?` boxes things automatically if necessary. Unfortunate that the type gets erased, but I don't think there's a way around that

crates/bevy_ecs/src/system/commands/command.rs
JaySpruce: I tried `impl Command` early on and it didn't work, but now it does. I guess I changed something that broke it. I'll do that real quick

crates/bevy_ecs/src/system/commands/command.rs
JaySpruce: That kinda adds a weird space between the word and the s when rendered. I do that when they're both real things (like `Command` and `Commands`), but otherwise I think it's fine to make it prettier

LikeLakers2: Well, the main reason I wrote this is to request consistency, as the docs for the `spawn_batch` function just above this use the <code>[\`MyStruct\`]s</code> style.

JaySpruce: I just copy-pasted that from the old helper function, didn't notice :P I'll take care of that

LikeLakers2: That works. Feel free to mark these two doc-comment reviews as resolved. :)

LikeLakers2 (approved):
While I do think some improvements could still be made (see my suggestions regarding `Command::with_error_handling` and `EntityCommand::with_entity`), I'd be happy with having this merged as-is. :)

NthTensor:
@JoJoJet points out in discord (https://discord.com/channels/691052431525675048/1325887475729698837/1325917241396564069) that we may want to use concrete error types if possible.

alice-i-cecile:
I would also prefer concrete error types that get cast at the last minute if at all possible, but I don't feel that *has* to be done in this PR.

JaySpruce:
I only have a vague idea of how to implement that so this might not be true, but I think it would have to be less convenient for struct-based commands for that to work. I've tried to keep it so that, if you `impl Command for SomeStruct`, you still only have to implement `apply` and it'll just work. I think we could have more concrete errors, but structs would have to do more work to implement `Command`.

I don't know how much we still care about struct commands, I've just tried to avoid making them worse. If a decision-maker decides that they're less important, that's fine by me

alice-i-cecile:
I think that we can overwhelmingly move away from struct-based commands :) All of our internal experiments to move away from them have gone smoothly. Even when you need to capture data / settings, closures work fine.

I don't mind regressing their UX there.

cart (approved):
I've wrapped up my review. In general I think this is the right general direction but I think some things should change. I've put together a branch with fixes to my points below. I think we should merge this PR and then review and merge my PR (in reasonably quick succession to avoid churn for people following main).

1. Command and EntityCommand always returning a `Result` is unnecessary and allows for a lot of weirdness. For example, this pr (likely accidentally) added a bunch of unnecessary error handlers to Commands that need no error handling. This could have been detected / disallowed if it was encoded in the type system.
2. As others have said, making commands generic on error type instead of boxing in-place is a good idea.
3. I think we can make the EntityCommand and Command traits _much_ simpler and remove the marker type. For EntityCommand, this hinges on moving to `EntityWorldMut`, which I strongly believe we should do anyway, as it allows batched entity commands to be applied with a single entity lookup, and it makes the declaration of most entity commands much simpler.
4. I'm not a fan of defining temporary error handlers on Commands. It adds more branching, makes Commands bigger / more expensive to initialize (note that we construct it at high frequencies / treat it like a pointer type), makes the code harder to follow, and introduces a bunch of additional functions. I think we should rely on a default error handler in `queue_fallible` in combination with calling `commands.queue(command.with_error_handler(handler))` directly on commands that need to opt out for some reason. This comes with the tradeoff of "command shorthand methods" on Commands only using the default error handler, but imo this is worth it.

My PR should be ready tomorrow. The branch works, but it needs a bit of polish + documentation.

alice-i-cecile:
I'm going to wait until your PR is up to merge this, but I'm considering this "ready-to-merge" :)

alice-i-cecile:
Thank you to everyone involved with the authoring or reviewing of this PR! This work is relatively important and needs release notes! Head over to https://github.com/bevyengine/bevy-website/issues/1979 if you'd like to help out.

PR #17048, opened 2024-12-30
Simplify sort/max_by calls

Victoronz:
# Objective

Some sort calls and `Ord` impls are unnecessarily complex.

## Solution

Rewrite the "match on cmp, if equal do another cmp" as either a comparison on tuples, or `Ordering::then_with`, depending on whether the compare keys need construction.

`sort_by` -> `sort_by_key` when symmetrical. Do the same for `min_by`/`max_by`.

Note that `total_cmp` can only work with `sort_by`, and not on tuples.

When sorting collected query results that contain `Entity`/`MainEntity`/`RenderEntity` in their `QueryData`, with that `Entity` in the sort key: 
stable -> unstable sort (all queried entities are unique)

If key construction is not simple, switch to `sort_by_cached_key` when possible.

Sorts that are only performed to discover the maximal element are replaced by `max_by_key`.

Dedicated comparison functions and structs are removed where simple.

Derive `PartialOrd`/`Ord` when useful.

Misc. closure style inconsistencies.

## Testing
- Existing tests.

### DIFF
--- a/crates/bevy_animation/src/animatable.rs
+++ b/crates/bevy_animation/src/animatable.rs
@@ -125,7 +125,7 @@ impl Animatable for bool {
     #[inline]
     fn blend(inputs: impl Iterator<Item = BlendInput<Self>>) -> Self {
         inputs
-            .max_by(|a, b| FloatOrd(a.weight).cmp(&FloatOrd(b.weight)))
+            .max_by_key(|x| FloatOrd(x.weight))
             .map(|input| input.value)
             .unwrap_or(false)
     }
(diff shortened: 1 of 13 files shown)


crates/bevy_animation/src/animatable.rs
```diff
@@ -125,7 +125,7 @@ impl Animatable for bool {
     #[inline]
     fn blend(inputs: impl Iterator<Item = BlendInput<Self>>) -> Self {
         inputs
-            .max_by(|a, b| FloatOrd(a.weight).cmp(&FloatOrd(b.weight)))
```
BenjaminBrienen, line 128: I wonder if this could be a clippy lint

BenjaminBrienen (approved):

alice-i-cecile (approved):

PR #17049, opened 2024-12-30
Fix `no_std` CI Warnings and WASM Compatibility

bushrat011899:
# Objective

- Resolve several warnings encountered when compiling for `no_std` around `dead_code`
- Fix compatibility with `wasm32-unknown-unknown` when using `no_std` (identified by Sachymetsu on [Discord](https://discord.com/channels/691052431525675048/692572690833473578/1323365426901549097))

## Solution

- Removed some unused imports
- Added `allow(dead_code)` for certain private items when compiling on `no_std`
- Fixed `bevy_app` and `bevy_tasks` compatibility with WASM when compiling without `std` by appropriately importing `Box` and feature gating panic unwinding

## Testing

- CI

### DIFF
--- a/crates/bevy_math/src/primitives/polygon.rs
+++ b/crates/bevy_math/src/primitives/polygon.rs
@@ -1,12 +1,17 @@
 #[cfg(feature = "alloc")]
-use alloc::{collections::BTreeMap, vec::Vec};
+use {
+    super::{Measured2d, Triangle2d},
+    alloc::{collections::BTreeMap, vec::Vec},
+};
 
 use core::cmp::Ordering;
 
 use crate::Vec2;
 
-use super::{Measured2d, Triangle2d};
-
+#[cfg_attr(
+    not(feature = "alloc"),
+    expect(dead_code, reason = "this type is only used with the alloc feature")
+)]
 #[derive(Debug, Clone, Copy)]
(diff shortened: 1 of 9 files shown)


crates/bevy_math/src/primitives/polygon.rs
```diff
@@ -136,6 +153,10 @@ impl Ord for Segment {
 }
 
 /// Holds information about which segment is above and which is below a given [`Segment`]
+#[cfg_attr(
```
BenjaminBrienen, line 156: I thought that we normally use conditional compilation instead of `expect`s.

BenjaminBrienen (approved):

crates/bevy_math/src/primitives/polygon.rs
bushrat011899: I decided to use `expect` here to communicate that the item itself _is_ `no_std` compatible, it's just not _needed_. But I'm happy to switch to just conditional compilation instead, it should give the same end-results

alice-i-cecile (approved):

crates/bevy_math/src/primitives/polygon.rs
BenjaminBrienen: I like your explanation actually

PR #8705, opened 2023-05-28
Allow systems to return results

JoJoJet:
# Objective

When writing an app in rust, the `main` function usually returns nothing, but it can also return a `Result`; when doing this, returning an `Err` value will result in a panic. This makes writing fallible code much more ergonomic, since `?` can be used to propagate errors.

This flexibility should be extended to systems: users should be able to return a `Result` from their systems. This can be done by calling `system.pipe(unwrap)` when adding the system, however this approach requires more boilerplate, incurs a runtime overhead due to the `PipeSystem`, removes context from the error message (since the panic message has no way of accessing the system's name), and makes the name of the combined system messier (which is stored by allocating a new `String`).

## Solution

A system can only be added to a schedule if it implements `IntoSystemConfigs<>`. Currently, this is only implemented for systems that take no input and return nothing. Now, this trait is also implemented for any systems that return `Result<(), impl Debug>`. A custom system adapter is used that panics when the system returns an error and includes the system name in the message.

Initially, I was concerned that this may make the compile error for systems with invalid return types more confusing. However in my testing, the error message does not seem to be any worse.

---

## Changelog

- Systems may now return values of type `Result<(), E>`, where `E` is any type that implements `std::fmt::Debug`.

PR #17051, opened 2024-12-30
Fallible systems trait based

hymm:
# Objective

- #16589 added an enum to switch between fallible and infallible system. This branching should be unnecessary if we wrap infallible systems in a function to return `Ok(())`.

## Solution

- Create a wrapper system for `System<(), ()>`s that returns `Ok` on the call to `run` and `run_unsafe`. The wrapper should compile out, but I haven't checked.
- I removed the `impl IntoSystemConfigs for BoxedSystem<(), ()>` as I couldn't figure out a way to keep the impl without double boxing.

## Testing

- ran `many_foxes` example to check if it still runs.

## Migration Guide

- `IntoSystemConfigs` has been removed for `BoxedSystem<(), ()>`. Either use `InfallibleSystemWrapper` before boxing or make your system return `bevy::ecs::prelude::Result`.


alice-i-cecile:
@Neo-Zhixing how do you feel about this implementation instead?

alice-i-cecile (approved):
I prefer this design: I think it's clearer and reducing the branching is good.

NthTensor:
Seems great! Have you tested logging and panics to see if this mangles the source tracking, as was the issue in https://github.com/bevyengine/bevy/pull/8705?

hymm:
> Have you tested logging and panics to see if this mangles the source tracking, as was the issue in https://github.com/bevyengine/bevy/pull/8705?

I'll check, but it should be fine. This should just add one more line to the backtrace.

NthTensor (approved):
This is clearly a much better approach. Thanks for tidying up after me :).

hymm:
Seems ok.
```text
thread '<unnamed>' panicked at examples/stress_tests/many_foxes.rs:116:5:
testing panics backtraces
Encountered a panic in system `many_foxes::setup`!
error: process didn't exit successfully: `target\debug\examples\many_foxes.exe` (exit code: 101)
```

Neo-Zhixing:
Oops I'm a bit late to the discussions, but this is clearly a much better approach.

PR #17054, opened 2024-12-31
small documentation update and issue template fix

HyperCodec:
# Objective
Fix some outdated `bevy_state` documentation examples.

## Solution
- updated some doc examples in `bevy_state` that hadn't been updated with the API.
- fixed an outdated link in the documentation issue template that referred to a 404 page instead of the contribution guide.

## Testing
No necessary testing aside from the usual doctests.

---

## Showcase
N/A

## Migration Guide
N/A

### DIFF
--- a/.github/ISSUE_TEMPLATE/docs_improvement.md
+++ b/.github/ISSUE_TEMPLATE/docs_improvement.md
@@ -10,4 +10,4 @@ assignees: ''
 
 Provide a link to the documentation and describe how it could be improved. In what ways is it incomplete, incorrect, or misleading?
 
-If you have suggestions on exactly what the new docs should say, feel free to include them here. Alternatively, make the changes yourself and [create a pull request](https://bevyengine.org/learn/book/contributing/code/) instead.
+If you have suggestions on exactly what the new docs should say, feel free to include them here. Alternatively, make the changes yourself and [create a pull request](https://bevyengine.org/learn/contribute/helping-out/writing-docs/) instead.
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -170,6 +170,8 @@ jobs:
         with:
           target: wasm32-unknown-unknown
       - name: Check wasm
+        env:
+          RUSTFLAGS: "-D warnings"
         run: cargo check --target wasm32-unknown-unknown
 
   build-wasm-atomics:
@@ -195,7 +197,7 @@ jobs:
       - name: Check wasm
         run: cargo check --target wasm32-unknown-unknown -Z build-std=std,panic_abort
         env:
-          RUSTFLAGS: "-C target-feature=+atomics,+bulk-memory"
+          RUSTFLAGS: "-C target-feature=+atomics,+bulk-memory -D warnings"
 
   markdownlint:
     runs-on: ubuntu-latest
@@ -319,7 +321,7 @@ jobs:
         run: cargo run -p ci -- doc
         env:
           CARGO_INCREMENTAL: 0
-          RUSTFLAGS: "-C debuginfo=0"
+          RUSTFLAGS: "-C debuginfo=0 -D warnings"
       # This currently report a lot of false positives
       # Enable it again once it's fixed - https://github.com/bevyengine/bevy/issues/1983
       # - name: Installs cargo-deadlinks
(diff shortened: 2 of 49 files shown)


alice-i-cecile (approved):
Thanks! Normally I would suggest that you put semi-related changes in different PRs but these are both trivial enough that I don't mind.

alice-i-cecile:
Interesting: CI says you need to reformat your code.

HyperCodec:
> Interesting: CI says you need to reformat your code.

I think that was just a space or something.

 Now I’m just tweaking some stuff to get the CI to compile the actual doctests (Schedule::add_systems and App::add_systems are different, and it was using Schedule before)

HyperCodec:
can't run the doctests on my local machine for whatever reason (there are some random compile errors that don't appear in the CI, probably #[cfg] shenanigans) so this is mostly going to be a process of trial and error where i commit and then wait for the CI to tell me i'm wrong.

HyperCodec:
Ok now it's complaining that the `Counter` resource hasn't been registered, but in all of those doctests there is a `.insert_resource::<Counter>()`, so I'm not entirely sure what's wrong with it now (although it could just be something really obvious that I overlooked).

alice-i-cecile:
@BD103 @janhohenheim do either of you know what's going wrong here?

BenjaminBrienen:
The wasm-atomics will be fixed if you update your branch

HyperCodec:
That merge fixed the wasm-atomics CI, but the issue of Counter not existing persists. I ran it as an actual binary with the following code for easier stacktrace following:

```rust
use bevy_ecs::prelude::*;
use bevy_state::prelude::*;
use bevy_app::{App, Update};
use bevy_state::app::StatesPlugin;

#[derive(Resource, Default)]
struct Counter(u8);

#[derive(States, Clone, Copy, Default, Eq, PartialEq, Hash, Debug)]
enum GameState {
    #[default]
    Playing,
    Paused,
}

fn main() {
    let mut app = App::new();
    app
        .init_resource::<Counter>()
        .add_plugins(StatesPlugin)
        .add_systems(Update,
            // `state_exists` will only return true if the
            // given state exists
            my_system.run_if(state_exists::<GameState>),
        );
    // `GameState` does not yet exist `my_system` won't run
    app.run();
    assert_eq!(app.world().resource::<Counter>().0, 0);
    app.init_state::<GameState>();
    // `GameState` now exists so `my_system` will run
    app.run();
    assert_eq!(app.world().resource::<Counter>().0, 1);
}

fn my_system(mut counter: ResMut<Counter>) {
    counter.0 += 1;
}
```

and this fails on `app.world().resource::<Counter>()` in the first condition. If I comment the first `app.run()` and `assert_eq!` out, the system runs properly, but it still panics on the second assert. I think this narrows down the cause to the method of access.

HyperCodec:
This code works as expected:
```rust
use bevy_ecs::prelude::*;
use bevy_state::prelude::*;
use bevy_app::{App, Update};
use bevy_state::app::StatesPlugin;

#[derive(Resource, Default, Debug)]
struct Counter(u8);

#[derive(States, Clone, Copy, Default, Eq, PartialEq, Hash, Debug)]
enum GameState {
    #[default]
    Playing,
    Paused,
}

fn main() {
    let mut app = App::new();
    app
        .init_resource::<Counter>()
        .add_plugins(StatesPlugin)
        .add_systems(Update,
            // `state_exists` will only return true if the
            // given state exists
            (
                my_system.run_if(state_exists::<GameState>),
                print_counter.after(my_system),
            )
        )
        .init_state::<GameState>();
    // `GameState` now exists so `my_system` will run
    app.run();
}

fn my_system(mut counter: ResMut<Counter>) {
    counter.0 += 1;
}

fn print_counter(world: &mut World) {
    dbg!(world.resource::<Counter>());
}
```

Why would using a system retrieve a different world than `app.world()`?

HyperCodec:
Is this a bug or undocumented, intended behavior?

mockersf:
yup it's intended: 
https://github.com/bevyengine/bevy/blob/294e0db71930313e3412e633426ef3713d0ef56b/crates/bevy_app/src/app.rs#L183

to run, the runner has to consume the app, so it's replace by an empty one:
https://github.com/bevyengine/bevy/blob/294e0db71930313e3412e633426ef3713d0ef56b/crates/bevy_app/src/app.rs#L183

HyperCodec:
CI says it works now.

PR #17057, opened 2024-12-31
Update picking docs

aevyrie:
More updates to picking docs. Addresses some nits with wording, and fixes some out of date information.


laundmo (approved):
the seemingly autoformatter based changes are a bit odd and annoying. i wonder how that happens - i'd have assumed all PRs use rustfmt and thats that.

BenjaminBrienen (approved):

PR #17065, opened 2024-12-31
Do not lowercase asset file extensions

### LINKED ISSUE #17064: Asset file extensions are sometimes handled as case-insensitive
## Bevy version

0.15

## What you did

* Register an asset loader for the extension `Custom`
* Try to load the asset `asset.Custom` **untyped**

## What went wrong

The asset cannot be loaded: `ERROR bevy_asset::server: Could not find an asset loader matching: Loader Name: None; Asset Type: None; Extension: None; Path: Some("data/asset.Custom");`

## Additional information

I [pushed a Bevy branch to my fork with a slightly changed `custom_asset` example that reproduces the issue](https://github.com/NiklasEi/bevy/tree/reproduce-issue-with-case-sensitive-asset-file-extensions).

The issue originally came up in NiklasEi/bevy_common_assets#44

Bevy seems case-sensitive for some parts of the asset file names (e.g. you cannot load the asset `Test.custom` as `test.custom`), but the file endings seem case-insensitive. I can register the loader for `custom` and it will be found and used when loading `data.Custom`.

NiklasEi:
# Objective

Resolves #17064

## Solution

- Bevy no longer converts asset file extensions to lowercase before trying to resolve an asset loader

## Testing

- I adapted the `custom_asset` example (see comment in #17064)
- The changes were tested on Linux

As far as I know, Windows has a case-insensitive file system by default, so case-sensitive asset file extensions are probably bad practice in a game. But we should be case-sensitive everywhere or handle asset paths completely case-insensitive. 

Before this PR:
* asset loader extensions are case-sensitive
* asset file names are case-sensitive
* asset file extensions are converted to lowercase :zap: 

Now everything should be case-sensitive

### DIFF
--- a/crates/bevy_asset/src/path.rs
+++ b/crates/bevy_asset/src/path.rs
@@ -454,7 +454,7 @@ impl<'a> AssetPath<'a> {
     pub fn get_full_extension(&self) -> Option<String> {
         let file_name = self.path().file_name()?.to_str()?;
         let index = file_name.find('.')?;
-        let mut extension = file_name[index + 1..].to_lowercase();
+        let mut extension = file_name[index + 1..].to_owned();
 
         // Strip off any query parameters
         let query = extension.find('?');
@@ -972,5 +972,8 @@ mod tests {
 
         let result = AssetPath::from("http://a.tar.bz2?foo=bar#Baz");
         assert_eq!(result.get_full_extension(), Some("tar.bz2".to_string()));
+
+        let result = AssetPath::from("asset.Custom");
+        assert_eq!(result.get_full_extension(), Some("Custom".to_string()));
     }
 }


BenjaminBrienen (approved):

ickshonpe (approved):
Looks good, agree wrt consistancy.

NiklasEi:
@alice-i-cecile I don't think the WASM atomics failure is due to my changes

BenjaminBrienen:
> @alice-i-cecile I don't think the WASM atomics failure is due to my changes

Just update your branch. It was already fixed. 🙂

nilclass:
I want to argue that the previous behavior was actually better (even though it seems inconsistent at first) :see_no_evil:.

My argument rests on three points
1. that it's more convenient from a user's perspective, for the extension to be case-insensitive
2. that it's more convenient from a plugin author's perspective
3. that there is no good use-case for needing case-sensitive extensions

From a user's perspective: if my assets are coming from a third party, I may not have control over the case of the file extension.
Example: some STL files in [this repository](https://github.com/google-deepmind/mujoco_menagerie/tree/cfd91c5605e90f0b77860ae2278ff107366acc87/unitree_g1/assets) use an uppercase extension, others use lowercase (example shared by @dlhxzb in https://github.com/nilclass/bevy_stl/pull/17).

Similarly when collaborating with people who work on systems with case-insensitive filesystems, asking them to use a consistent case for the extension can create unnecessary friction.

From a plugin author's perspective: suppose I want my users to be able to load files with any case (e.g. `*.stl` files and `*.STL` files as in the example above), then I would need to specify `["stl", "STL"]` as the list of supported extensions.
Once I do that though, one could argue that `*.Stl` should be supported as well. So do I add that to the list too?
Ultimately that decision will be up to the plugin author, so case-sensitivity of file extensions would not be consistent across the bevy ecosystem, as some authors may support only uppercase extensions, while others support only lowercase extensions, while others would support some combination.

This leaves the question: is there a good use-case for case-sensitive extensions?
I can find some examples where it is, but I'd argue they are esoteric.
Looking at the [spec for shared-mime-info's "glob" attribute](https://specifications.freedesktop.org/shared-mime-info-spec/0.21/ar01s02.html#id-1.3.7), it matches extensions case-insensitive by default, unless explicitly requested:
> Applications MUST match globs case-insensitively, except when the case-sensitive attribute is set to true. This is so that e.g. main.C will be seen as a C++ file, but IMAGE.GIF will still use the *.gif pattern. 

Looking at the actual usage of the `case-sensitive` attribute in the freedesktop.org database, there are just a few entries:
```
$ grep case-sensitive /usr/share/mime/packages/freedesktop.org.xml
<!ATTLIST glob case-sensitive CDATA #IMPLIED>
    <glob pattern="core" case-sensitive="true"/>
    <glob pattern="perf.data" case-sensitive="true"/>
    <glob pattern="*.C" case-sensitive="true"/>
    <glob pattern="*.c" case-sensitive="true"/>
    <glob pattern="*.gs" case-sensitive="true"/>
```
None of these look relevant for the bevy asset loader.

Finally, what about @NiklasEi's original issue, of registering an asset loader for the `Custom` extension?
I think this could be solved, by:
1. Documenting that `AssetLoader.extensions` should return all-lowercase strings
2. (optionally) Logging a warning if an `AssetLoader` specifies an extension that contains uppercase characters, pointing to the documentation

PR #17066, opened 2024-12-31
reduce nesting in the `sparse_set` module

ickshonpe:
# Objective

Just being fussy but I hate this `.map(|v| v.is_some()).unwrap_or(false)` stuff.

## Solution
Reduce it using `and_then`.

### DIFF
--- a/crates/bevy_ecs/src/storage/sparse_set.rs
+++ b/crates/bevy_ecs/src/storage/sparse_set.rs
@@ -50,7 +50,7 @@ macro_rules! impl_sparse_array {
             #[inline]
             pub fn contains(&self, index: I) -> bool {
                 let index = index.sparse_set_index();
-                self.values.get(index).and_then(Option::as_ref).is_some()
+                self.values.get(index).is_some_and(|v| v.is_some())
             }
 
             /// Returns a reference to the value at `index`.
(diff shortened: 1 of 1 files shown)


crates/bevy_ecs/src/storage/sparse_set.rs
```diff
@@ -50,7 +50,7 @@ macro_rules! impl_sparse_array {
             #[inline]
             pub fn contains(&self, index: I) -> bool {
                 let index = index.sparse_set_index();
-                self.values.get(index).map(|v| v.is_some()).unwrap_or(false)
+                self.values.get(index).and_then(Option::as_ref).is_some()
```
Jondolf, line 53: This is equivalent as well, right? To me this is easier to read
```suggestion
                self.values.get(index).is_some_and(|v| v.is_some())
```

ickshonpe: I quite like how in the `and_then` version it flattens it down and ends on `.is_some()` rather than going "inside" the option but yeah that seems clearer.

Jondolf (approved):

Victoronz (approved):
It'd be nice to use the dedicated `Option`/`Result` methods properly across the engine!

BenjaminBrienen (approved):

PR #17067, opened 2024-12-31
Replace `map` + `unwrap_or(false)` with `is_some_and`

ickshonpe:
# Objective

The `my_option.map(|inner| inner.is_whatever).unwrap_or(false)` pattern is fragile and ugly. 

Replace it with `is_some_and` everywhere.

### DIFF
--- a/crates/bevy_animation/src/animatable.rs
+++ b/crates/bevy_animation/src/animatable.rs
@@ -126,8 +126,7 @@ impl Animatable for bool {
     fn blend(inputs: impl Iterator<Item = BlendInput<Self>>) -> Self {
         inputs
             .max_by_key(|x| FloatOrd(x.weight))
-            .map(|input| input.value)
-            .unwrap_or(false)
+            .is_some_and(|input| input.value)
     }
 }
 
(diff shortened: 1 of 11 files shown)


alice-i-cecile (approved):

BenjaminBrienen (approved):

Victoronz:
I think `map_or(true, ...)` also be done in this PR, as it is the same pattern.

BenjaminBrienen:
> I think `map_or(true, ...)` also be replaced in this PR, as it is the same pattern.

That is already linted by clippy and I fixed that already in another PR.

Victoronz:
> > I think `map_or(true, ...)` also be replaced in this PR, as it is the same pattern.
> 
> That is already linted by clippy and I fixed that already in another PR.

Does it capture everything? I had done a quick search in the engine and still found some `map_or(false/true, ...)`.
Not many, but still some.

alice-i-cecile:
> I think map_or(true/false, ...) also be replaced in these PRs where applicable, as it is the same pattern.

I'm fine either way. I'll merge this on Monday if ickshonpe doesn't get to it, or doesn't want to bother.

PR #17069, opened 2024-12-31
Make the input focus docs less keyboard-centric

alice-i-cecile:
# Objective

Following #16876, `bevy_input_focus` works with more than just keyboard inputs! The docs should reflect that.

## Solution

Fix a few missed mentions in the documentation.

Also add a brief reference to navigation frameworks within the module docs to help give more breadcrumbs.

### DIFF
--- a/crates/bevy_input_focus/src/lib.rs
+++ b/crates/bevy_input_focus/src/lib.rs
@@ -5,12 +5,13 @@
     html_favicon_url = "https://bevyengine.org/assets/icon.png"
 )]
 
-//! Keyboard focus system for Bevy.
+//! A UI-centric focus system for Bevy.
 //!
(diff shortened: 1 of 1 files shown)


viridia (approved):

viridia:
I think that at some point we are also going to need a guide that talks about how to structure input in your app. This wouldn't just be about bevy_input_focus specifically, but more about using observers for input generally, including things like integration with game states.

alice-i-cecile:
> I think that at some point we are also going to need a guide that talks about how to structure input in your app. This wouldn't just be about bevy_input_focus specifically, but more about using observers for input generally, including things like integration with game states.

Yep, absolutely. I want to hammer out the patterns first, then move to examples, then move to written docs and unified guidance.

PR #17071, opened 2024-12-31
Return `Result` from tab navigation API

alice-i-cecile:
# Objective

Tab navigation can fail in all manner of ways. The current API recognizes this, but merely logs a warning and returns `None`.

We should supply the actual reason for failure to the caller, so they can handle it in whatever fashion they please (including logging a warning!).

Swapping to a Result-oriented pattern is also a bit more idiomatic and makes the code's control flow easier to follow.

## Solution

- Refactor the `tab_navigation` module to return a `Result` rather than an `Option` from its key APIs.
- Move the logging to the provided prebuilt observer. This leaves the default behavior largely unchanged, but allows for better user control.
- Make the case where no tab group was found for the currently focused entity an error branch, but provide enough information that we can still recover from it.

## Testing

The `tab_navigation` example continues to function as intended.

### DIFF
--- a/crates/bevy_input_focus/Cargo.toml
+++ b/crates/bevy_input_focus/Cargo.toml
@@ -17,6 +17,8 @@ bevy_hierarchy = { path = "../bevy_hierarchy", version = "0.15.0-dev", default-f
 bevy_utils = { path = "../bevy_utils", version = "0.15.0-dev", default-features = false }
 bevy_window = { path = "../bevy_window", version = "0.15.0-dev", default-features = false }
 
+thiserror = { version = "2", default-features = false }
+
 [dev-dependencies]
 smol_str = "0.2"
 
--- a/crates/bevy_input_focus/src/tab_navigation.rs
+++ b/crates/bevy_input_focus/src/tab_navigation.rs
@@ -315,7 +315,7 @@ pub fn handle_tab_navigation(
         && key_event.state == ButtonState::Pressed
         && !key_event.repeat
     {
-        let next = nav.navigate(
+        let maybe_next = nav.navigate(
             &focus,
             if keys.pressed(KeyCode::ShiftLeft) || keys.pressed(KeyCode::ShiftRight) {
                 NavAction::Previous
@@ -324,7 +324,7 @@ pub fn handle_tab_navigation(
             },
         );
 
-        match next {
+        match maybe_next {
             Ok(next) => {
                 trigger.propagate(false);
                 focus.set(next);
--- a/crates/bevy_input_focus/src/tab_navigation.rs
+++ b/crates/bevy_input_focus/src/tab_navigation.rs
@@ -80,17 +80,25 @@ impl TabGroup {
     }
 }
 
-/// A navigation action for tabbing.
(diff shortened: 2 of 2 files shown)


crates/bevy_input_focus/Cargo.toml
```diff
@@ -17,6 +17,8 @@ bevy_hierarchy = { path = "../bevy_hierarchy", version = "0.15.0-dev", default-f
 bevy_utils = { path = "../bevy_utils", version = "0.15.0-dev", default-features = false }
 bevy_window = { path = "../bevy_window", version = "0.15.0-dev", default-features = false }
 
+thiserror = { version = "2", default-features = false }
```
alice-i-cecile, line 20: We aren't using workspace dependencies yet, but I checked to ensure that this is synced with the other crates.

crates/bevy_input_focus/src/tab_navigation.rs
```diff
@@ -144,19 +180,29 @@ impl TabNavigation<'_, '_> {
...
+                    Err(TabNavigationError::NoTabGroupForCurrentFocus {
+                        previous_focus: focus.0.unwrap(),
+                        new_focus: entity,
+                    })
+                } else {
+                    Ok(entity)
+                }
+            }
+            Err(e) => Err(e),
         }
```
ickshonpe, line 197:
```suggestion
        self.navigate_in_group(tabgroup, focus, action)
            .and_then(|entity| {
                if focus.0.is_some() && tabgroup.is_none() {
                    Err(TabNavigationError::NoTabGroupForCurrentFocus {
                        previous_focus: focus.0.unwrap(),
                        new_focus: entity,
                    })
                } else {
                    Ok(entity)
                }
            })
```

ickshonpe: I've become obsessed with rewriting everything using chained combinators everywhere for some reason today, just ignore me if you hate this style.

BenjaminBrienen (approved):

crates/bevy_input_focus/src/tab_navigation.rs
alice-i-cecile: Ha, I prefer the nested matching / if's: it's easier for me to piece together what's going on without remembering what all of the little methods do.

viridia (approved):
Beautiful.

crates/bevy_input_focus/src/tab_navigation.rs
seritools: Maybe I'm missing something, but couldn't _both_ nesting _and_ combinators be prevented here with a regular ol `?`? Also swapped `is_some` + `unwrap` with `if let`, which reintroduces one level of nesting until let-chains are available I guess.

Also same disclaimer, feel free to ignore ^^

```suggestion
        let entity = self.navigate_in_group(tabgroup, focus, action)?;

        if let Some(previous_focus) = focus.0 {
            if tabgroup.is_none() {
                return Err(TabNavigationError::NoTabGroupForCurrentFocus {
                    previous_focus,
                    new_focus: entity,
                });
            }
        }

        Ok(entity)
```

PR #17075, opened 2025-01-01
Rename `track_change_detection` flag to `track_location`

### LINKED ISSUE #16852: Rename `track_change_detection` feature of `bevy_ecs`
As seen in #16047 and so on, this feature now encompasses much more than just change detection. We should rename it to `track_location`, since tracking source code locations for various events is the core idea.

skimmedsquare:
# Objective

- As stated in the related issue, this PR is to better align the feature flag name with what it actually does and the plans for the future.
- Fixes #16852 

## Solution

- Simple find / replace

## Testing

- Local run of `cargo run -p ci`

## Migration Guide

The `track_change_detection` feature flag has been renamed to `track_location` to better reflect its extended capabilities.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -455,7 +455,7 @@ ios_simulator = ["bevy_internal/ios_simulator"]
 bevy_state = ["bevy_internal/bevy_state"]
 
 # Enables source location tracking for change detection and spawning/despawning, which can assist with debugging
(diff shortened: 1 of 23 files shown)


Olle-Lukowski (approved):

alice-i-cecile (approved):
Very nice, thanks!

PR #17077, opened 2025-01-01
Add `NameOrEntity::get_from_query` method

### LINKED ISSUE #17061: `NameOrEntity` is difficult to `Query::get`

Iterating a `Query<NameOrEntity>` is pretty ergonomic: you just iterate through the query and it lets you display print the entity name. Quite nice! However if you need to use `Query::get` the situation is more complicated. If you `Query::<NameOrEntity>::get`, you get back a `Result` since the entity may not exist - this is standard and expected, however handling this is a pain today.

For a concrete example, I have an observer that is triggered on `Pointer<Out>`. However another system can despawn this entity before the `Pointer<Out>` event is triggered. Trying to debug something, I added a `Query<NameOrEntity>` and did `names.get(trigger.entity())`. `Result` doesn't implement `Display` though, so I unwrap the result - but since the entity is already despawned, the query doesn't find anything. Therefore, my app panics.

The "proper" way to handle this is with:

```rust
names.get(trigger.entity()).unwrap_or(NameOrEntityItem { name: None, entity: trigger.entity() })
```

Ideally all this logic would be bundled into an associated function on `NameOrEntity`, for example `NameOrEntity::get_from_query(names, trigger.entity())`.

We could instead make an extension trait for `Query<NameOrEntity>`, but this seems less discoverable and not that much more useful.

Olle-Lukowski:
# Objective

Fixes #17061.

## Solution

Implemented a method to easily get the `NameOrEntity` from an `Entity`, and add some docs for it.

## Testing

All tests pass, doctests work fine too.

### DIFF
--- a/crates/bevy_ecs/src/name.rs
+++ b/crates/bevy_ecs/src/name.rs
@@ -1,6 +1,12 @@
 //! Provides the [`Name`] [`Component`], used for identifying an [`Entity`].
 
-use crate::{self as bevy_ecs, component::Component, entity::Entity, query::QueryData};
+use crate::{
+    self as bevy_ecs,
+    component::Component,
+    entity::Entity,
+    query::{QueryData, QueryFilter},
+    system::Query,
+};
 
 use alloc::{
     borrow::{Cow, ToOwned},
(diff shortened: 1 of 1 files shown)


crates/bevy_ecs/src/name.rs
```diff
@@ -141,6 +147,36 @@ pub struct NameOrEntity {
     pub entity: Entity,
 }
 
+impl NameOrEntity {
+    /// Get a [`NameOrEntity`] from a [`Query`].
+    /// This is a convenience wrapper around [`Query::get`],
+    /// which uses the fact that a [`NameOrEntity`] can always be obtained,
+    /// even if the entity does not have a [`Name`].
```
andriyDev, line 154: This line isn't quite right. The problem is not whether the entity has a name, but whether the entity is still "alive" (not despawned). Getting an entity without a name works already!

andriyDev (changes requested):
Thanks for writing this up so quickly!!

crates/bevy_ecs/src/name.rs
Olle-Lukowski: Ah right, probably messed that up in my "quick writing" xD, will fix this!

PR #17078, opened 2025-01-01
Fix sprite performance regression since retained render world

superdump:
# Objective

- Fix sprite rendering performance regression since retained render world changes
  - The retained render world changes moved `ExtractedSprites` from using the highly-optimised `EntityHasher` with an `Entity` to using `FixedHasher` with `(Entity, MainEntity)`. This was enough to regress framerate in bevymark by 25%.

## Solution

- Move the render world entity into a member of `ExtractedSprite` and change `ExtractedSprites` to use `MainEntityHashMap` for its storage
- Disable sprite picking in bevymark

## Testing

M4 Max. `bevymark --waves 100 --per-wave 1000 --benchmark`. main in yellow vs PR in red:

Screenshot 2025-01-01 at 16 36 22

20.2% median frame time reduction.

Screenshot 2025-01-01 at 16 38 37

49.7% median extract_sprites execution time reduction.

Comparing 0.14.2 yellow vs PR red:
Screenshot 2025-01-01 at 16 40 06

~6.1% median frame time reduction.

---

## Migration Guide

- `ExtractedSprites` is now using `MainEntityHashMap` for storage, which is keyed on `MainEntity`.
- The render world entity corresponding to an `ExtractedSprite` is now stored in the `render_entity` member of it.

### DIFF
--- a/crates/bevy_sprite/src/render/mod.rs
+++ b/crates/bevy_sprite/src/render/mod.rs
@@ -19,7 +19,6 @@ use bevy_ecs::{
 };
 use bevy_image::{BevyDefault, Image, ImageSampler, TextureFormatPixelInfo};
 use bevy_math::{Affine3A, FloatOrd, Quat, Rect, Vec2, Vec4};
-use bevy_render::sync_world::MainEntity;
 use bevy_render::view::RenderVisibleEntities;
 use bevy_render::{
     render_asset::RenderAssets,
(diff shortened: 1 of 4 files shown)


superdump:
I'm not totally sure this is correct. I don't see why it wouldn't be, but I am unfamiliar with all the ins and outs of the retained render world changes.

tychedelia (approved):
Thanks @superdump, this fix looks exactly right. We shouldn't ever need to key off the entity pair together. Our policy has pretty much been to make sure we key all collections to `MainEntity`.

FYI: The reason for the `(RenderEntity, MainEntity)` shenanigans in the render phase is because the `RenderCommand` api supports looking up components on the `PhaseItem` via `RenderCommand::ItemQuery`, which needs the render world id to function. So while our internal render code almost exclusively looks up things via a `MainEntity` keyed collection on `RenderCommand::Param`, user code may not. I believe the exception here in our own code is some UI stuff. It's a bit of a mess!

tychedelia:
I added this to `0.15.2` although I realize now it technically has a breaking change in `ExtractedSprites`. The performance regression is probably not extreme in non-benchmark use cases.

alice-i-cecile (approved):

alice-i-cecile:
Very neat to see that we have a net performance gain here!

PR #17079, opened 2025-01-01
Fix depth_bias and build errors on less capable platforms

### LINKED ISSUE #14169: `depth_bias` is ignored on `StandardMaterial`
## Bevy version

0.14.0

```
cargo 1.79.0 (ffa9cf99a 2024-06-03)
```
```ignore
SystemInfo { os: "Windows 11 Pro", kernel: "22631", cpu: "AMD Ryzen 9 7950X 16-Core Processor", core_count: "16", memory: "31.6 GiB" }
AdapterInfo { name: "AMD Radeon RX 7900 XTX", vendor: 4098, device: 29772, device_type: DiscreteGpu, driver: "AMD proprietary driver", driver_info: "24.5.1 (LLPC)", backend: Vulkan }
```

## What you did

Spawned a small plane mesh with a transparent ring texture and a `depth_bias` of `f32::INFINITY`.
```
(
    PbrBundle {
        mesh: meshes.add(Plane3d::new(Vec3::Y, Vec2::splat(0.3))),
        material: materials.add(StandardMaterial {
            base_color_texture: Some(
                asset_cache.cache(asset_server.load("textures/ring.ktx2")), // NOTE: `asset_cache.cache` simply clones the strong handle into a vector and returns the original handle
            ),
            base_color: Srgba::WHITE.with_alpha(0.12549019).into(), // 32 alpha
            alpha_mode: AlphaMode::Blend,
            unlit: true,
            depth_bias: f32::INFINITY, // Render the ring above all other 3d objects
            ..Default::default()
        }),
        transform: Transform::from_translation(Vec3::new(
            0.0,
            -HALF_PLAYER_HEIGHT - GROUND_OFFSET, // Places the ring exactly on the surface level
            0.0,
        )),
        ..Default::default()
    },
    NotShadowCaster,
)
```

## What went wrong

Z-fighting:
image

## Additional information

This was previously functional on Bevy 0.13.2. As a "workaround" (probably how I should've done it from the beginning) I made a different camera on a higher render layer to render the ring on top of all other 3d meshes.

nandee95:
# Objective

- I'm compiling (parts of) bevy for an embedded platform with no 64bit atomic and ctrlc handler support. Some compilation errors came up. This PR contains the fixes for those.
- Fix depth_bias casting in PBR material (Fixes #14169)
  - Negative depth_bias values were casted to 0 before this PR
  - f32::INFINITY depth_bias value was casted to -1 before this PR

## Solutions

- Restrict 64bit atomic reflection to supported platforms
- Restrict ctrlc handler to supported platforms (linux, windows or macos instead of "not wasm")
- The depth bias value (f32) is first casted to i32 then u64 in order to preserve negative values

## Testing
- This version compiles on a platform with no 64bit atomic support, and no ctrlc support
- CtrlC handler still works on Linux and Windows (I can't test on Macos)
- depth_bias:
```rust
println!("{}",f32::INFINITY as u64 as i32); // Prints: -1 (old implementation)
println!("{}",f32::INFINITY as i32 as u64 as i32); // Prints: 2147483647 (expected, new implementation)
```
Also ran a modified version of 3d_scene example with the following results:

RED cube depth_bias: -1000.0
BLUE cube depth_bias: 0.0
image

RED cube depth_bias: -INF
BLUE cube depth_bias: 0.0
image

RED cube depth_bias: INF (case reported in #14169)
BLUE cube depth_bias: 0.0
(Im not completely sure whats going on with the shadows here, it seems like depth_bias has some affect to those aswell, if this is unintentional this issue was not introduced by this PR)
image

### DIFF
--- a/crates/bevy_app/Cargo.toml
+++ b/crates/bevy_app/Cargo.toml
@@ -93,7 +93,7 @@ portable-atomic-util = { version = "0.2.4", features = [
   "alloc",
 ], optional = true }
 
-[target.'cfg(not(target_arch = "wasm32"))'.dependencies]
+[target.'cfg(any(unix, windows))'.dependencies]
 ctrlc = { version = "3.4.4", optional = true }
 
 [target.'cfg(target_arch = "wasm32")'.dependencies]
--- a/crates/bevy_pbr/src/pbr_material.rs
+++ b/crates/bevy_pbr/src/pbr_material.rs
@@ -1238,7 +1238,9 @@ impl From<&StandardMaterial> for StandardMaterialKey {
         }
 
         key.insert(StandardMaterialKey::from_bits_retain(
-            (material.depth_bias as u64) << STANDARD_MATERIAL_KEY_DEPTH_BIAS_SHIFT,
+            // Casting to i32 first to ensure the full i32 range is preserved.
+            // (wgpu expects the depth_bias as an i32 when this is extracted in a later step)
+            (material.depth_bias as i32 as u64) << STANDARD_MATERIAL_KEY_DEPTH_BIAS_SHIFT,
         ));
         key
     }
--- a/crates/bevy_app/src/lib.rs
+++ b/crates/bevy_app/src/lib.rs
@@ -30,7 +30,7 @@ mod schedule_runner;
 mod sub_app;
 #[cfg(feature = "bevy_tasks")]
 mod task_pool_plugin;
-#[cfg(all(not(target_arch = "wasm32"), feature = "std"))]
+#[cfg(all(any(unix, windows), feature = "std"))]
 mod terminal_ctrl_c_handler;
 
 pub use app::*;
@@ -42,7 +42,7 @@ pub use schedule_runner::*;
 pub use sub_app::*;
 #[cfg(feature = "bevy_tasks")]
 pub use task_pool_plugin::*;
-#[cfg(all(not(target_arch = "wasm32"), feature = "std"))]
+#[cfg(all(any(unix, windows), feature = "std"))]
 pub use terminal_ctrl_c_handler::*;
 
 /// The app prelude.
--- a/crates/bevy_dev_tools/src/ci_testing/mod.rs
+++ b/crates/bevy_dev_tools/src/ci_testing/mod.rs
@@ -62,7 +62,7 @@ impl Plugin for CiTestingPlugin {
 
         // The offending system does not exist in the wasm32 target.
         // As a result, we must conditionally order the two systems using a system set.
-        #[cfg(not(target_arch = "wasm32"))]
+        #[cfg(any(unix, windows))]
         app.configure_sets(
             Update,
             SendEvents.before(bevy_app::TerminalCtrlCHandlerPlugin::exit_on_flag),
--- a/crates/bevy_internal/src/default_plugins.rs
+++ b/crates/bevy_internal/src/default_plugins.rs
@@ -18,7 +18,7 @@ plugin_group! {
         bevy_window:::WindowPlugin,
         #[cfg(feature = "bevy_window")]
         bevy_a11y:::AccessibilityPlugin,
-        #[custom(cfg(not(target_arch = "wasm32")))]
+        #[custom(cfg(any(unix, windows)))]
         bevy_app:::TerminalCtrlCHandlerPlugin,
         #[cfg(feature = "bevy_asset")]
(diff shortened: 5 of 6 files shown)


BD103:
Out of curiosity, which platform + OS were you compiling for?

nandee95:
ESP32 microcontroller. It runs FreeRTOS and have a compatibility layer for std.
https://docs.esp-rs.org/book/overview/using-the-standard-library.html

The compiler target_os config is the following:
```rust
#[cfg(target_os = "espidf")]
```
I don't think bevy should have any traces of this platform thats why I thought its a better idea to just specify the OSes that actually have support for ctrlc.

crates/bevy_pbr/src/pbr_material.rs
```diff
@@ -1238,7 +1238,7 @@ impl From<&StandardMaterial> for StandardMaterialKey {
         }
 
         key.insert(StandardMaterialKey::from_bits_retain(
-            (material.depth_bias as u64) << STANDARD_MATERIAL_KEY_DEPTH_BIAS_SHIFT,
+            (material.depth_bias as i32 as u64) << STANDARD_MATERIAL_KEY_DEPTH_BIAS_SHIFT,
```
BenjaminBrienen, line 1241: Why is this correct?

nandee95: The `depth_bias` is actually stored as an `i32` past this step, the `i32` value will go to wgpu. Casting from f32 to an unsigned will cause it to lose its value when its negative and will not work well with i32 out of range values(e.g. f32::INFINITY). If we cast to `i32` first, it will then cast to `u64` it will retain the full range of i32 and will be passed to wgpu correctly.

I also made you a small demo here to better understand the casting (left os old behaviour, right is new):
https://godbolt.org/z/d94h7nfYT

crates/bevy_app/Cargo.toml
```diff
@@ -93,7 +93,7 @@ portable-atomic-util = { version = "0.2.4", features = [
   "alloc",
 ], optional = true }
 
-[target.'cfg(not(target_arch = "wasm32"))'.dependencies]
+[target.'cfg(any(target_os = "macos", target_os = "linux", target_os = "windows"))'.dependencies]
```
BD103, line 96: Will this work for all platforms supported by `ctrlc`? Looks like they use [`#[cfg(unix)]` and `#[cfg(windows)]`](https://github.com/Detegr/rust-ctrlc/blob/3502f44eca2fd006fd578d59e41471ca57d26191/src/platform/mod.rs#L10-L14):

```rust
#[cfg(unix)]
mod unix;

#[cfg(windows)]
mod windows;
```

BD103 (approved):
The `ctrlc` and reflection feature gate look fine, given all the platforms supported by `ctrlc` are supported by us. I'm not sure I completely understand the `as i32 as u64` casting, so @pcwalton I would appreciate your eyes on it!

crates/bevy_app/Cargo.toml
BD103: Specifically, are there platforms that are `#[cfg(unix)]` that are not Linux or MacOS that we should worry about?

BenjaminBrienen: We support FreeBSD and there are others like TvOS we could support.

nandee95: Changed all occurrences from
```rust
cfg(any(target_os = "macos", target_os = "linux", target_os = "windows"))
```
to
```rust
cfg(any(unix, windows))
```

Re-tested on Windows and Linux and CtrlC handler still works.

BenjaminBrienen (approved):

crates/bevy_pbr/src/pbr_material.rs
IceSentry: Can you add a comment describing this? I'm fine if it's even just copy pasted from what you just wrote. It's not immediately obvious otherwise.

nandee95: I gave it some thought. I wouldn't copy my previous comment as is since it was made in the context of the current issue which may cause more confusion to a future contributor. 
 Instead I explained why the double casting is necessary to make sure its not removed accidentally by someone else:
 ```
            // Casting to i32 first to ensure the full i32 range is preserved.
            // (wgpu expects the depth_bias as an i32 when this is extracted in a later step)
 ```

BenjaminBrienen (approved):

PR #17083, opened 2025-01-01
Lint against `clippy::allow_attributes` and `clippy::allow_attributes_without_reason`

LikeLakers2:
This PR is a WIP - but please feel free to comment anything that comes to mind.

# Objective

## Solution
Set the `clippy::allow_attributes` and `clippy::allow_attributes_without_reason` lints to `deny`, and bring Bevy's code base in line with the new restrictions.

This PR **does not** modify any code other than lint attributes, as I consider that out-of-scope - except in cases such as the `unused_variables` lint, where the lint attribute can be removed entirely via a very simple change.

## Testing
Uhh... I ran `cargo clippy`? lol

### DIFF
--- a/crates/bevy_color/src/oklaba.rs
+++ b/crates/bevy_color/src/oklaba.rs
@@ -216,7 +216,10 @@ impl ColorToComponents for Oklaba {
     }
 }
 
-#[allow(clippy::excessive_precision)]
+#[expect(
+    clippy::excessive_precision,
+    reason = "The code with excessive precision was copied from another codebase."
+)]
 impl From<LinearRgba> for Oklaba {
     fn from(value: LinearRgba) -> Self {
         let LinearRgba {
@@ -239,7 +242,10 @@ impl From<LinearRgba> for Oklaba {
     }
 }
 
-#[allow(clippy::excessive_precision)]
+#[expect(
+    clippy::excessive_precision,
+    reason = "The code with excessive precision was copied from another codebase."
+)]
 impl From<Oklaba> for LinearRgba {
     fn from(value: Oklaba) -> Self {
         let Oklaba {
--- a/crates/bevy_ecs/src/bundle.rs
+++ b/crates/bevy_ecs/src/bundle.rs
@@ -322,7 +322,7 @@ macro_rules! tuple_impl {
             fn get_components(self, func: &mut impl FnMut(StorageType, OwningPtr<'_>)) {
                 #[allow(
                     non_snake_case,
-                    reason = "The name of these variables is determined at the invocation location, not by this macro."
+                    reason = "Variable names use parameters provided by the macro invocation."
                 )]
                 let ($(mut $name,)*) = self;
                 $(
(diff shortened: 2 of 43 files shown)


crates/bevy_color/src/oklaba.rs
```diff
@@ -216,7 +216,10 @@ impl ColorToComponents for Oklaba {
     }
 }
 
-#[allow(clippy::excessive_precision)]
+#[expect(
+    clippy::excessive_precision,
+    reason = "The code with excessive precision was copied from another codebase."
+)]
```
LikeLakers2, line 222: I'm not really happy with this lint attribute (as well as the other one changed in this file), as there really isn't a reason for the floats to have excessive precision, beyond the code having been copied verbatim from another code-base.

That said, I feel truncating the floats is out-of-scope for this PR, hence me changing them to be `expect(...)`.

crates/bevy_ecs/src/bundle.rs
```diff
@@ -594,7 +630,10 @@ impl BundleInfo {
     /// This method _should not_ be called outside of [`BundleInfo::write_components`].
     /// For more information, read the [`BundleInfo::write_components`] safety docs.
     /// This function inherits the safety requirements defined there.
-    #[allow(clippy::too_many_arguments)]
+    #[expect(
+        clippy::too_many_arguments,
+        reason = "Can't be rewritten with less arguments."
+    )]
     pub(crate) unsafe fn initialize_required_component(
```
LikeLakers2, line 637: Ditto.

crates/bevy_ecs/src/entity/entity_set.rs
```diff
@@ -379,26 +379,25 @@ impl<I: Iterator<Item: TrustedEntityBorrow> + Debug> Debug for UniqueEntityIter<
...
     use super::UniqueEntityIter;
 
     #[derive(Component, Clone)]
     pub struct Thing;
 
-    #[allow(clippy::iter_skip_zero)]
     #[test]
+    #[expect(
+        clippy::iter_skip_zero,
+        reason = "Uniqueness of entities should be preserved, even when skipping zero entities."
```
LikeLakers2, line 399: I don't know how I feel about this reason... but I don't know if the `.skip(0)` should be removed.

crates/bevy_ecs/src/world/entity_ref.rs
```diff
@@ -1688,7 +1688,6 @@ impl<'w> EntityWorldMut<'w> {
             })
         };
 
-        #[allow(clippy::undocumented_unsafe_blocks)] // TODO: document why this is safe
```
LikeLakers2, line 1693: Yes, this means clippy will complain about an undocumented unsafe block.

Unfortunately, I have *no* idea what the safety doc should be, or if it'd even be correct.

LikeLakers2, line 1693: Yes, this means clippy will complain about an undocumented unsafe block.

Unfortunately, I have *no* idea what the safety doc should be, or if it'd even be correct.

LikeLakers2:
Closing in favor of splitting this PR up.

PR #17084, opened 2025-01-02
Update entity cloning benchmarks

BD103:
# Objective

- `entity_cloning` was separated from the rest of the ECS benchmarks.
- There was some room for improvement in the benchmarks themselves.
- Part of #16647.

## Solution

- Merge `entity_cloning` into the rest of the ECS benchmarks.
- Apply the `bench!` macro to all benchmark names.\
- Reorganize benchmarks and their helper functions, with more comments than before.
- Remove all the extra component definitions (`C2`, `C3`, etc.), and just leave one. Now all entities have exactly one component.

## Testing

```sh
# List all entity cloning benchmarks, to verify their names have updated.
cargo bench -p benches --bench ecs entity_cloning -- --list

# Test benchmarks by running them once.
cargo test -p benches --bench ecs entity_cloning

# Run all benchmarks (takes about a minute).
cargo bench -p benches --bench ecs entity_cloning
```

---

## Showcase

image

Interestingly, using `Clone` instead of `Reflect` appears to be 2-2.5 times faster. Furthermore, there were noticeable jumps in time when running the benchmarks:

image

I theorize this is because the `World` is allocating more space for all the entities, but I don't know for certain. Neat!

### DIFF
--- a/benches/benches/bevy_ecs/entity_cloning.rs
+++ b/benches/benches/bevy_ecs/entity_cloning.rs
@@ -4,7 +4,7 @@ use benches::bench;
 use bevy_ecs::bundle::Bundle;
 use bevy_ecs::component::ComponentCloneHandler;
 use bevy_ecs::reflect::AppTypeRegistry;
-use bevy_ecs::{component::Component, reflect::ReflectComponent, world::World};
+use bevy_ecs::{component::Component, world::World};
 use bevy_hierarchy::{BuildChildren, CloneEntityHierarchyExt};
 use bevy_math::Mat4;
 use bevy_reflect::{GetTypeRegistration, Reflect};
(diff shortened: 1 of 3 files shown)


BD103:
@eugineerd if you have the time, I would appreciate your review! Since you originally created entity cloning and its benchmarks, you can catch any mistakes that I made or things that I left out. Thanks!

alice-i-cecile (approved):

benches/benches/bevy_ecs/entity_cloning.rs
```diff
@@ -1,173 +1,206 @@
...
+criterion_group!(
+    benches,
+    single,
+    hierarchy_tall,
+    hierarchy_wide,
+    hierarchy_many,
+);
 
 #[derive(Component, Reflect, Default, Clone)]
 #[reflect(Component)]
```
eugineerd, line 22: I don't think `#[reflect(Component)]` is actually required for cloning functionality anymore. Maybe this can be removed?

eugineerd (commented):
This is a nice cleanup. I created this benchmark just to test the changes in #16717 and didn't put too much effort into making it right (not to mention this is my first time using criterion in general), so this is certainly an improvement. 

My only concern is that 1 component per entity is not really representative of the average workload for this command. I would expect entities to have at least a couple different components in real-world scenario. If you look at the table in #16717 , you can see that the 1 component benchmarks actually regressed a bit, even though multiple components gained a big perf boost. I'd say that just having benchmark work for multiple components should be enough as 1 component isn't as common.

benches/benches/bevy_ecs/entity_cloning.rs
BD103: Yes, it doesn't seem like these are needed anymore. I tested removing `#[reflect(Component)]` and there wasn't any change in benchmark performance for the `Reflect` variants. If it was required, I would assume that the clone command would skip those components and significantly decrease the benchmark time, or just panic.

BD103:
> My only concern is that 1 component per entity is not really representative of the average workload for this command. I would expect entities to have at least a couple different components in real-world scenario. If you look at the table in #16717 , you can see that the 1 component benchmarks actually regressed a bit, even though multiple components gained a big perf boost. I'd say that just having benchmark work for multiple components should be enough as 1 component isn't as common.

Thank you for the review, this was exactly the kind of thing I was hoping you'd catch! It should all be fixed now in e50ab98db0ac22ad3a85d0357d9dbea9a42b3862, but let me know if anything else needs changed. :)

BenjaminBrienen (approved):

eugineerd (approved):
Should be good now.

PR #17086, opened 2025-01-02
Remove Implicit `std` Prelude from `no_std` Crates

bushrat011899:
# Background

In `no_std` compatible crates, there is often an `std` feature which will allow access to the standard library. Currently, with the `std` feature _enabled_, the [`std::prelude`](https://doc.rust-lang.org/std/prelude/index.html) is implicitly imported in all modules. With the feature _disabled_, instead the [`core::prelude`](https://doc.rust-lang.org/core/prelude/index.html) is implicitly imported. This creates a subtle and pervasive issue where `alloc` items _may_ be implicitly included (if `std` is enabled), or must be explicitly included (if `std` is not enabled).

# Objective

- Make the implicit imports for `no_std` crates consistent regardless of what features are/not enabled.

## Solution

- Replace the `cfg_attr` "double negative" `no_std` attribute with conditional compilation to _include_ `std` as an external crate.
```rust
// Before
#![cfg_attr(not(feature = "std"), no_std)]

// After
#![no_std]

#[cfg(feature = "std")]
extern crate std;
```
- Fix imports that are currently broken but are only now visible with the above fix.

## Testing

- CI

## Notes

I had previously used the "double negative" version of `no_std` based on general consensus that it was "cleaner" within the Rust embedded community. However, this implicit prelude issue likely was considered when forming this consensus. I believe the reason why is the items most affected by this issue are provided by the `alloc` crate, which is rarely used within embedded but extensively used within Bevy.

### DIFF
--- a/crates/bevy_app/src/lib.rs
+++ b/crates/bevy_app/src/lib.rs
@@ -16,7 +16,7 @@
 
 //! This crate is about everything concerning the highest-level, application layer of a Bevy app.
 
-#![cfg(feature = "std")]
+#[cfg(feature = "std")]
 extern crate std;
 
 extern crate alloc;
--- a/crates/bevy_app/src/lib.rs
+++ b/crates/bevy_app/src/lib.rs
@@ -11,9 +11,12 @@
     html_logo_url = "https://bevyengine.org/assets/icon.png",
     html_favicon_url = "https://bevyengine.org/assets/icon.png"
 )]
-#![cfg_attr(not(feature = "std"), no_std)]
+#![no_std]
 
 //! This crate is about everything concerning the highest-level, application layer of a Bevy app.
+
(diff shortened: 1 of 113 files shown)


BenjaminBrienen (approved):

alice-i-cecile (approved):
Very reasonable diff: I like that this is more explicit and less likely to cause weird CI problems for contributors. Ping me when CI is behaving :)

alice-i-cecile:
Ooh, you've unlocked the next, harder level of CI.

bushrat011899:
Ok @alice-i-cecile I'm ready for the next attempt at merging this thing!

bushrat011899:
Ok one last CI action is failing in the merge queue. Back to the IDE...

PR #17090, opened 2025-01-02
bevy_color: Apply `#![deny(clippy::allow_attributes, clippy::allow_attributes_without_reason)]`

LikeLakers2:
# Objective

## Solution
Set the `clippy::allow_attributes` and `clippy::allow_attributes_without_reason` lints to `deny`, and bring `bevy_audio` in line with the new restrictions.

## Testing

### DIFF
--- a/crates/bevy_color/src/oklaba.rs
+++ b/crates/bevy_color/src/oklaba.rs
@@ -216,7 +216,10 @@ impl ColorToComponents for Oklaba {
     }
 }
 
-#[allow(clippy::excessive_precision)]
+#[expect(
+    clippy::excessive_precision,
+    reason = "The code with excessive precision was copied from another codebase."
+)]
 impl From<LinearRgba> for Oklaba {
     fn from(value: LinearRgba) -> Self {
         let LinearRgba {
@@ -239,7 +242,10 @@ impl From<LinearRgba> for Oklaba {
     }
 }
 
-#[allow(clippy::excessive_precision)]
+#[expect(
+    clippy::excessive_precision,
+    reason = "The code with excessive precision was copied from another codebase."
+)]
 impl From<Oklaba> for LinearRgba {
     fn from(value: Oklaba) -> Self {
         let Oklaba {
--- a/crates/bevy_color/src/lib.rs
+++ b/crates/bevy_color/src/lib.rs
@@ -1,6 +1,10 @@
 #![cfg_attr(docsrs, feature(doc_auto_cfg))]
 #![forbid(unsafe_code)]
-#![deny(clippy::allow_attributes, clippy::allow_attributes_without_reason)]
+#![deny(
+    clippy::allow_attributes,
+    clippy::allow_attributes_without_reason,
+    reason = "See #17111; To be removed once all crates are in-line with these attributes"
+)]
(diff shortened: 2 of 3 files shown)


IQuick143 (approved):

crates/bevy_color/src/oklaba.rs
```diff
@@ -216,7 +216,10 @@ impl ColorToComponents for Oklaba {
     }
 }
 
-#[allow(clippy::excessive_precision)]
+#[expect(
+    clippy::excessive_precision,
+    reason = "The code with excessive precision was copied from another codebase."
+)]
```
LikeLakers2, line 222: See review comment.

crates/bevy_color/src/oklaba.rs
```diff
@@ -239,7 +242,10 @@ impl From<LinearRgba> for Oklaba {
...
```
LikeLakers2, line 248: See review comment.

crates/bevy_color/src/oklaba.rs
IQuick143: I'd say the rationale could be to be consistent with other implementations?
But I agree that truncating the floats is out of scope here, also maybe for documentation purposes (If someone were to use our code as a template and worked with a greater precision. Alternatively you're expressing the intent of multiplying with these precise values, even if it's not going to be achievable with f32's.)

BenjaminBrienen (approved):

crates/bevy_color/src/oklaba.rs
alice-i-cecile: I like the reason in your last message better!

eero-lehtinen: The original source of these numbers is actually from https://bottosson.github.io/posts/oklab/#converting-from-linear-srgb-to-oklab.

eero-lehtinen: We should probably reference that instead of other Rust crates.

LikeLakers2: So should I change the reason to this?
```rust
reason = "Excessive precision is used here to be consistent with other implementations."
```

Or are we still discussing what it should be?

eero-lehtinen: I would just truncate.

LikeLakers2: @eero-lehtinen I *would* truncate - but honestly, doing so feels like it'd be out-of-scope for this PR.

eero-lehtinen: That message or what I said is completely fine. Just ship it.

LikeLakers2: Truncated floats are visible at this PR: https://github.com/bevyengine/bevy/pull/17109

I'm not a fan of the truncated floats, as they reduce readability... I would personally just do as I have in this PR, and move the allow to an expect.

alice-i-cecile (approved):

IQuick143 (approved):

LikeLakers2:
@alice-i-cecile Requesting this be re-added to the merge queue.

PR #17091, opened 2025-01-02
bevy_math: Apply `#[deny(clippy::allow_attributes, clippy::allow_attributes_without_reason)]`

LikeLakers2:
# Objective

## Solution
Set the `clippy::allow_attributes` and `clippy::allow_attributes_without_reason` lints to `deny`, and bring `bevy_math` in line with the new restrictions.

## Testing

### DIFF
--- a/crates/bevy_math/src/ops.rs
+++ b/crates/bevy_math/src/ops.rs
@@ -22,7 +22,7 @@
 #[cfg(not(feature = "libm"))]
 #[expect(
     clippy::disallowed_methods,
-    reason = "Many of the disallowed methods point to methods which are ultimately defined here."
+    reason = "Many of the disallowed methods are disallowed to force code to use the feature-conditional re-exports from this module, but this module itself is exempt from that rule."
 )]
 mod std_ops {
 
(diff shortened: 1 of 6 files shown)


crates/bevy_math/src/ops.rs
```diff
@@ -23,6 +20,10 @@
 // - `f32::ln_gamma`
 
 #[cfg(not(feature = "libm"))]
+#[expect(
+    clippy::disallowed_methods,
+    reason = "Many of the disallowed methods point to methods which are ultimately defined here."
```
IQuick143, line 25: I'd suggest an (imo clearer) version.
```suggestion
    reason = "Many of the disallowed methods are disallowed to force code to use the feature-conditional re-exports from this module, but this module itself is exempt from that rule."
```

crates/bevy_math/src/ops.rs
```diff
@@ -519,6 +520,10 @@ mod libm_ops_for_no_std {
 }
 
 #[cfg(feature = "std")]
+#[expect(
+    clippy::disallowed_methods,
+    reason = "Many of the disallowed methods point to methods which are ultimately defined here."
```
IQuick143, line 525:
```suggestion
    reason = "Many of the disallowed methods are disallowed to force code to use the feature-conditional re-exports from this module, but this module itself is exempt from that rule."
```

IQuick143 (approved):

BenjaminBrienen (approved):

PR #14828, opened 2024-08-19
Apply unused_qualifications lint

EdJoPaTo:
# Objective

Fixes #14782

## Solution

Enable the lint and fix all upcoming hints (`--fix`). Also try to figure out the false-positive (see review comment). Maybe split this PR up into multiple parts where only the last one enables the lint, so some can already be merged resulting in less many files touched / less potential for merge conflicts?

Currently, there are some cases where it might be easier to read the code with the qualifier, so perhaps remove the import of it and adapt all cases? In the current stage it's just a plain adoption of the suggestions in order to have a base to discuss.

## Testing

`cargo clippy` and `cargo run -p ci` are happy.

PR #17092, opened 2025-01-02
bevy_reflect: Apply `#[deny(clippy::allow_attributes, clippy::allow_attributes_without_reason)]`

LikeLakers2:
# Objective

## Solution
Set the `clippy::allow_attributes` and `clippy::allow_attributes_without_reason` lints to `deny`, and bring `bevy_reflect` in line with the new restrictions.

## Testing

### DIFF
--- a/crates/bevy_reflect/src/impls/std.rs
+++ b/crates/bevy_reflect/src/impls/std.rs
@@ -1,5 +1,7 @@
-// Temporary workaround for impl_reflect!(Option/Result false-positive
-#![allow(unused_qualifications)]
+#![expect(
+    unused_qualifications,
+    reason = "Temporary workaround for impl_reflect!(Option/Result false-positive"
+)]
 
 use crate::{
     self as bevy_reflect, impl_type_path, map_apply, map_partial_eq, map_try_apply,
@@ -236,7 +238,6 @@ macro_rules! impl_reflect_for_atomic {
             #[cfg(feature = "functions")]
             crate::func::macros::impl_function_traits!($ty);
 
-            #[allow(unused_mut)]
             impl GetTypeRegistration for $ty
             where
                 $ty: Any + Send + Sync,
--- a/crates/bevy_reflect/src/lib.rs
+++ b/crates/bevy_reflect/src/lib.rs
@@ -1,4 +1,5 @@
 #![expect(missing_docs, reason = "Not all docs are written yet, see #3492.")]
+#![deny(clippy::allow_attributes, clippy::allow_attributes_without_reason)]
 #![cfg_attr(
     any(docsrs, docsrs_dep),
     expect(
@@ -10,10 +11,6 @@
 #![doc(
     html_logo_url = "https://bevyengine.org/assets/icon.png",
     html_favicon_url = "https://bevyengine.org/assets/icon.png"
-)]
-#![deny(
-    clippy::allow_attributes,
-    clippy::allow_attributes_without_reason
 )]
 
 //! Reflection in Rust.
--- a/crates/bevy_reflect/src/lib.rs
+++ b/crates/bevy_reflect/src/lib.rs
@@ -1,5 +1,9 @@
 #![expect(missing_docs, reason = "Not all docs are written yet, see #3492.")]
-#![deny(clippy::allow_attributes, clippy::allow_attributes_without_reason)]
+#![deny(
+    clippy::allow_attributes,
+    clippy::allow_attributes_without_reason,
+    reason = "See #17111; To be removed once all crates are in-line with these attributes"
+)]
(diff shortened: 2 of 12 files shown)


crates/bevy_reflect/src/impls/std.rs
```diff
@@ -1,5 +1,7 @@
-// Temporary workaround for impl_reflect!(Option/Result false-positive
-#![allow(unused_qualifications)]
+#![expect(
+    unused_qualifications,
+    reason = "Temporary workaround for impl_reflect!(Option/Result false-positive"
```
BenjaminBrienen, line 3: I don't understand the parenthesis here

crates/bevy_reflect/src/lib.rs
```diff
@@ -683,8 +684,7 @@ pub mod __macro_exports {
         note = "consider annotating `{Self}` with `#[derive(Reflect)]`"
     )]
     pub trait RegisterForReflection {
-        #[allow(unused_variables)]
-        fn __register(registry: &mut TypeRegistry) {}
+        fn __register(_registry: &mut TypeRegistry) {}
```
BenjaminBrienen, line 687: What happens if you change this to:

```rs
        fn __register(registry: &mut TypeRegistry);
```

crates/bevy_reflect/src/serde/mod.rs
```diff
@@ -199,7 +199,7 @@ mod tests {
 
         #[reflect_trait]
         trait Enemy: Reflect + Debug {
-            #[allow(dead_code, reason = "this method is purely for testing purposes")]
+            #[expect(dead_code, reason = "this method is purely for testing purposes")]
```
BenjaminBrienen, line 202: This method may become used during said testing, and it would be annoying if this generated a warning during testing.

crates/bevy_reflect/src/type_registry.rs
```diff
@@ -79,8 +79,7 @@ pub trait GetTypeRegistration: 'static {
     ///
     /// This method is called by [`TypeRegistry::register`] to register any other required types.
     /// Often, this is done for fields of structs and enum variants to ensure all types are properly registered.
-    #[allow(unused_variables)]
-    fn register_type_dependencies(registry: &mut TypeRegistry) {}
+    fn register_type_dependencies(_registry: &mut TypeRegistry) {}
```
BenjaminBrienen, line 82: What happens if this is changed to:

```rs
    fn register_type_dependencies(registry: &mut TypeRegistry);
```

BenjaminBrienen (changes requested):

crates/bevy_reflect/src/impls/std.rs
alice-i-cecile: Yeah, I'd like to clean this up before merging. @MrGVSV, what did you mean here?

crates/bevy_reflect/src/serde/mod.rs
alice-i-cecile: I think I prefer leaving this as an expect for now.

crates/bevy_reflect/src/type_registry.rs
LikeLakers2: It starts linting `unused_variables`... which causes it to fail CI.

crates/bevy_reflect/src/impls/std.rs
MrGVSV: This actually wasn't me. Looks like it was added in #14828. My guess is that since `Option` and `Result` are in the prelude, their full qualification is unnecessary, but we rely on it for generating the `TypePath` impl within `impl_reflect!`.

crates/bevy_reflect/src/lib.rs
LikeLakers2: I consider this out-of-scope for this PR, but out of curiousity, I tried it anyways.

Changing it as you suggested seems to work - after you place `fn __register(_registry: &mut TypeRegistry) {}` in a bunch of impls (such as `impl RegisterForReflection for DynamicEnum`).

crates/bevy_reflect/src/type_registry.rs
LikeLakers2: I consider this out-of-scope for this PR, but out of curiousity, I tried it anyways.

It appears to cause a bunch of "not all trait items implemented" errors, whether from attributes like `#[remote_reflect(...)]`, or inside invocations to a proc macro called `impl_from_reflect_opaque`.

BenjaminBrienen (approved):

alice-i-cecile (approved):
Doesn't make things worse; no reason to block on fixing that weird comment.

PR #17096, opened 2025-01-02
Anamorphic Bloom

aevyrie:
https://github.com/user-attachments/assets/e2de3d20-4246-4eba-a0a7-8469a468dddb

The _JJ Abrahams_

https://github.com/user-attachments/assets/2dce3df9-665b-46ff-b687-e7cb54364f30

The _Cyberfunk 2025_

image

# Objective

- Add the ability to scale bloom for artistic control, and to mimic anamorphic blurs.

## Solution

- Add a scale factor in bloom settings, and plumb this to the shader.

## Testing

- Added runtime-tweak-able setting to the `bloom_3d`/`bloom_2d ` example

---

## Showcase

image

- Added `scale` parameter to `Bloom` to improve artistic control and enable anamorphic bloom.

### DIFF
--- a/crates/bevy_core_pipeline/src/bloom/bloom.wgsl
+++ b/crates/bevy_core_pipeline/src/bloom/bloom.wgsl
@@ -52,33 +52,61 @@ fn karis_average(color: vec3<f32>) -> f32 {
 
 // [COD] slide 153
 fn sample_input_13_tap(uv: vec2<f32>) -> vec3<f32> {
-#ifdef FIRST_DOWNSAMPLE
-    // Prevents noticeable sampling/ghosting artifacts when using large scales.
-    let scale = vec2<f32>(1.0, 1.0);
-#else
-    let scale = uniforms.scale;
-#endif
-    let x_s = scale.x / f32(textureDimensions(input_texture).x);
-    let y_s = scale.y / f32(textureDimensions(input_texture).y);
-    let x_l = 2.0 * x_s;
-    let y_l = 2.0 * y_s;
-    let x_ns = -1.0 * x_s;
-    let y_ns = -1.0 * y_s;
-    let x_nl = -2.0 * x_s;
-    let y_nl = -2.0 * y_s;
-    let a = textureSample(input_texture, s, uv + vec2<f32>(x_nl, y_l)).rgb;
-    let b = textureSample(input_texture, s, uv + vec2<f32>(0.0, y_l)).rgb;
-    let c = textureSample(input_texture, s, uv + vec2<f32>(x_l, y_l)).rgb;
-    let d = textureSample(input_texture, s, uv + vec2<f32>(x_nl, 0.0)).rgb;
+#ifdef UNIFORM_SCALE
+    // This is the fast path. When the bloom scale is uniform, the 13 tap sampling kernel can be
+    // expressed with constant offsets.
+    //
+    // It's possible that this isn't meaningfully faster than the "slow" path. However, because it
+    // is hard to test performance on all platforms, and uniform bloom is the most common case, this
+    // path was retained when adding non-uniform (anamorphic) bloom. This adds a small, but nonzero,
+    // cost to maintainability, but it does help me sleep at night.
+    let a = textureSample(input_texture, s, uv, vec2<i32>(-2, 2)).rgb;
+    let b = textureSample(input_texture, s, uv, vec2<i32>(0, 2)).rgb;
+    let c = textureSample(input_texture, s, uv, vec2<i32>(2, 2)).rgb;
+    let d = textureSample(input_texture, s, uv, vec2<i32>(-2, 0)).rgb;
     let e = textureSample(input_texture, s, uv).rgb;
-    let f = textureSample(input_texture, s, uv + vec2<f32>(x_l, 0.0)).rgb;
-    let g = textureSample(input_texture, s, uv + vec2<f32>(x_nl, y_nl)).rgb;
-    let h = textureSample(input_texture, s, uv + vec2<f32>(0.0, y_nl)).rgb;
-    let i = textureSample(input_texture, s, uv + vec2<f32>(x_l, y_nl)).rgb;
-    let j = textureSample(input_texture, s, uv + vec2<f32>(x_ns, y_s)).rgb;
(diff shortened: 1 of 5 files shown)


crates/bevy_core_pipeline/src/bloom/bloom.wgsl
```diff
@@ -51,19 +52,33 @@ fn karis_average(color: vec3<f32>) -> f32 {
...
+#endif
+    let x_s = scale.x / f32(textureDimensions(input_texture).x);
+    let y_s = scale.y / f32(textureDimensions(input_texture).y);
+    let x_l = 2.0 * x_s;
+    let y_l = 2.0 * y_s;
+    let x_ns = -1.0 * x_s;
+    let y_ns = -1.0 * y_s;
+    let x_nl = -2.0 * x_s;
+    let y_nl = -2.0 * y_s;
+    let a = textureSample(input_texture, s, uv + vec2<f32>(x_nl, y_l)).rgb;
```
aevyrie, line 69: Note for reviewers - I was not able to use the offset syntax used before, because I was hitting shader compiler errors when trying to pass computed offsets into what it said was a const context.

examples/3d/bloom_3d.rs
aevyrie, line 1: The tweaks to appearance in this file are purely cosmetic. The existing settings didn't seem to showcase bloom as favorably as it could.

crates/bevy_core_pipeline/src/bloom/bloom.wgsl
```diff
@@ -51,19 +52,33 @@ fn karis_average(color: vec3<f32>) -> f32 {
...
-    let b = textureSample(input_texture, s, uv, vec2<i32>(0, 2)).rgb;
-    let c = textureSample(input_texture, s, uv, vec2<i32>(2, 2)).rgb;
-    let d = textureSample(input_texture, s, uv, vec2<i32>(-2, 0)).rgb;
+#ifdef FIRST_DOWNSAMPLE
+    // Prevents noticeable sampling/ghosting artifacts when using large scales.
+    let scale = vec2<f32>(1.0, 1.0);
+#else
+    let scale = uniforms.scale;
+#endif
+    let x_s = scale.x / f32(textureDimensions(input_texture).x);
```
mate-h, line 61: Would make sense to combine these variables to a vec2

mate-h (commented):
Looks good to me, just style changes

crates/bevy_core_pipeline/src/bloom/bloom.wgsl
tychedelia: I'm not sure. I'd be curious to learn from others whether there are dedicated hardware paths for offsets when texture fetching on modern gpus. I wonder if it could also affect mips.

The other way we could handle this is by implementing support for [`@override`] (https://www.w3.org/TR/WGSL/#override-decls) although it's worth noting that this would create a new pipeline for each override so wouldn't be suitable for scenarios where you want to animate the bloom scale effect (and would be limiting to the +/-8 of that offset param). 

Assuming there aren't major performance concerns that others are familiar with this makes sense to me.

tychedelia (approved):
Approved pending SME feedback.

crates/bevy_core_pipeline/src/bloom/bloom.wgsl
aevyrie: Another option would be to branch on the special case when the offsets == 1, in which case we can take the happy path and use the const offset syntax for sampling. From my understanding, this branching should be cheap, because it is branching on a uniform, meaning all groups will take the same branch.

aevyrie: The spec doesn't really mention anything about perf. I just says the offset needs to be a const expression.
https://www.w3.org/TR/WGSL/#texturesample

tychedelia: Using a uniform branch is a good idea too, but I'd say let's keep it simple unless there's evidence of performance issue. The main question is always mobile. I'll bench mark on desktop this weekend if someone more knowledgeable doesn't get back with a more definitive answer about hardware.

IceSentry:
I don't have time to benchmark right now, but if someone benchmarks this and it shows no perf impact then the code LGTM. I'd just like to see some numbers first. My guess is that if there's an impact it will be pretty small, but would still be nice to know.

crates/bevy_core_pipeline/src/bloom/bloom.wgsl
aevyrie: Oops, well, I added the uniform branch. If anything, it will make it easier for you to test if there is any perf impact of the slow path. As soon as the scale is non-uniform, it takes the slow path automatically.

aevyrie:
I added a uniform branch, so the fast path should still exist, and the new code is only compiled in when using nonuniform bloom scale.

crates/bevy_core_pipeline/src/bloom/bloom.wgsl
tychedelia: Using the shader def is even better. I'd still like to benchmark this (just because I'm curious), but this certainly eliminates any concerns.

aevyrie: I am not englishing very well today. Yeah, I added a shader def, figured that would be pretty bulletproof, though it comes at the cost of some extra boilerplate on the setup side.

IceSentry (approved):
With the ifdef I'm happy to merge it without benchmarking, but it would be nice to have numbers and yeet the ifdef if it's not necessary.

aevyrie:
I'd like to try messing with the weights instead. After fixing the upscaling issue, non uniform blur looks worse because it can't rely on incorrect upsamples misaligned with the frags.

aevyrie:
I couldn't find a decent path with weights without making more significant changes. I think this is good as-is now. I've tested with the car scene again with the latest settings, and I'm happy with how it looks.

alice-i-cecile:
Thank you to everyone involved with the authoring or reviewing of this PR! This work is relatively important and needs release notes! Head over to https://github.com/bevyengine/bevy-website/issues/1978 if you'd like to help out.

PR #17041, opened 2024-12-30
`ExtractedSprites` slice buffer

ickshonpe:
# Objective

Instead of extracting an individual sprite per glyph of a text spawn or slice of a nine-patched sprite, add a buffer to store the extracted slice geometry.

Fixes #16972

## Solution

* New struct `ExtractedSlice` to hold sprite slice size, position and atlas info (for text each glyph is a slice).
* New resource `ExtractedSlices` that wraps the `ExtractedSlice` buffer. This is a separate resource so it can be used without sprites (with a text material, for example).
* New enum `ExtractedSpriteKind` with variants `Single` and `Slices`. `Single` represents a single sprite, `Slices` contains a range into the `ExtractedSlice` buffer.
* Only queue a single `ExtractedSprite` for sets of glyphs or slices and push the geometry for each individual slice or glyph into the `ExtractedSlice` buffer. 
* Modify `ComputedTextureSlices` to return an `ExtractedSlice` iterator instead of `ExtractedSprites`.
* Modify `extract_text2d_sprite` to only queue new `ExtractedSprite`s on font changes and otherwise push slices.

I don't like the name `ExtractedSpriteKind` much, it's a bit redundant and too haskellish. But although it's exported, it's not something users will interact with most of the time so don't want to overthink it.

## Testing
yellow = this pr, red = main

```cargo run --example many_glyphs --release --features "trace_tracy" -- --no-ui```

many-glyphs

```cargo run --example many_text2d --release --features "trace_tracy"```
many-text2d

## Migration Guide
* `ExtractedSprite` has a new `kind: ExtractedSpriteKind` field with variants `Single` and `Slices`. 
   - `Single` represents a single sprite. `ExtractedSprite`'s `anchor`, `rect`, `scaling_mode` and `custom_size` fields have been moved into `Single`. 
   - `Slices` contains a range that indexes into a new resource `ExtractedSlices`. Slices are used to draw elements composed from multiple sprites such as text or nine-patched borders.
* `ComputedTextureSlices::extract_sprites` has been renamed to `extract_slices`. Its `transform` and `original_entity` parameters have been removed.

PR #17101, opened 2025-01-02
`Text2d` and sprite slicing fix

### LINKED ISSUE #17098: `Text2d` and nine-slicing is broken
## Bevy version

main, bisected to #17078

SystemInfo { os: "macOS 14.5 Sonoma", kernel: "23.5.0", cpu: "Apple M1 Max", core_count: "10", memory: "64.0 GiB" }

## What you did

`cargo run --example text2d`

## What went wrong

It seems that only one glyph per entity gets drawn.

Image

ickshonpe:
# Objective

Fix #17098

`text2d` example:
image

## Solution

When extracting multiple sprites for text2d and texture slicing spawn new temporary render entities to act as keys in the `ExtractedSprites` `MainEntityHashMap`.

It wraps a temporary render world entity inside a `MainEntity` which doesn't seem great, but it seems to work?

### DIFF
--- a/crates/bevy_sprite/src/render/mod.rs
+++ b/crates/bevy_sprite/src/render/mod.rs
@@ -396,7 +396,7 @@ pub fn extract_sprites(
                         commands.spawn(TemporaryRenderEntity).id(),
                         sprite,
                     )
-                    .map(|e| (original_entity.into(), e)),
+                    .map(|e| (commands.spawn(TemporaryRenderEntity).id().into(), e)),
             );
         } else {
             let atlas_rect = sprite
@@ -417,7 +417,7 @@ pub fn extract_sprites(
 
             // PERF: we don't check in this function that the `Image` asset is ready, since it should be in most cases and hashing the handle is expensive
             extracted_sprites.sprites.insert(
-                original_entity.into(),
+                commands.spawn(TemporaryRenderEntity).id().into(),
                 ExtractedSprite {
                     color: sprite.color.into(),
                     transform: *transform,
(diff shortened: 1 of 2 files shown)


superdump:
While this may work, it for sure feels like a workaround, and probably hurts performance too (spawning new entities every frame?).

I think I have understood that the problem is the one to many mapping of main world entities to render world entities for text and nine-patch. We have this pattern for shadow mapping views but they are relative few so there’s no performance issue. However, for UI I expect it could easily become an issue?

Maybe a more flexible solution for the performance regression than what I did would be to implement some optimised hashing for a tuple of entities and go back to a hashmap of render world entity, main world entity?

ickshonpe:
> While this may work, it for sure feels like a workaround, and probably hurts performance too (spawning new entities every frame?).
> 
> I think I have understood that the problem is the one to many mapping of main world entities to render world entities for text and nine-patch. We have this pattern for shadow mapping views but they are relative few so there’s no performance issue. However, for UI I expect it could easily become an issue?
> 
> Maybe a more flexible solution for the performance regression than what I did would be to implement some optimised hashing for a tuple of entities and go back to a hashmap of render world entity, main world entity?

Yeah this is just intended as temporary fix for main as it's completely broken now.
It would be easy to add a pool of retained entities stored in a local to reuse as keys for one one to many sprite entities I guess.

#17041 is another option.

tychedelia:
Oof, yeah. I didn't consider the possibility for one to many when reviewing https://github.com/bevyengine/bevy/pull/17078, sorry!

I don't love adding more tech debt in this way and the semantics of the fix for sprite perf were clearly just incorrect.

> Maybe a more flexible solution for the performance regression than what I did would be to implement some optimised hashing for a tuple of entities and go back to a hashmap of render world entity, main world entity?

I think this makes more sense.

rparrett:
Superseded by #17123

PR #17102, opened 2025-01-02
Add basic directional (gamepad) navigation for UI (and non-UI)

alice-i-cecile:
# Objective

While directional navigation is helpful for UI in general for accessibility reasons, it is *especially* valuable for a game engine, where menus may be navigated primarily or exclusively through the use of a game controller.

Thumb-stick powered cursor-based navigation can work as a fallback, but is generally a pretty poor user experience. We can do better!

## Prior art

Within Bevy, https://github.com/nicopap/ui-navigation and https://github.com/rparrett/bevy-alt-ui-navigation-lite exist to solve this same problem. This isn't yet a complete replacement for that ecosystem, but hopefully we'll be there for 0.16.

## Solution

UI navigation is complicated, and the right tradeoffs will vary based on the project and even the individual scene.

We're starting with something simple and flexible, hooking into the existing `InputFocus` resource, and storing a manually constructed graph of entities to explore in a `DirectionalNavigationMap` resource. The developer experience won't be great (so much wiring to do!), but the tools are all there for a great user experience.

We could have chosen to represent these linkages via component-flavored not-quite-relations. This would be useful for inspectors, and would give us automatic cleanup when the entities were despawned, but seriously complicates the developer experience when building and checking this API. For now, we're doing a dumb "entity graph in a resource" thing and `remove` helpers. Once relations are added, we can re-evaluate.

I've decided to use a `CompassOctant` as our key for the possible paths. This should give users a reasonable amount of precise control without being fiddly, and playing reasonably nicely with arrow-key navigation. This design lets us store the set of entities that we're connected to as a 8-byte array (yay Entity-niching). In theory, this is maybe nicer than the double indirection of two hashmaps. but if this ends up being slow we should create benchmarks.

To make this work more pleasant, I've added a few utilities to the `CompassOctant` type: converting to and from usize, and adding a helper to find the 180 degrees opposite direction. These have been mirrored onto `CompassQuadrant` for consistency: they should be generally useful for game logic.

## Future work

This is a relatively complex initiative! In the hopes of easing review and avoiding merge conflicts, I've opted to split this work into bite-sized chunks.

Before 0.16, I'd like to have:

- An example demonstrating gamepad and tab-based navigation in a realistic game menu
- Helpers to convert axis-based inputs into compass quadrants / octants
- Tools to check the listed graph desiderata
- A helper to build a graph from a grid of entities
- A tool to automatically build a graph given a supplied UI layout

One day, it would be sweet if:

- We had an example demonstrating how to use focus navigation in a non-UI scene to cycle between game objects
- Standard actions for tab-style and directional navigation with a first-party bevy_actions integration
- We had a visual debugging tool to display these navigation graphs for QC purposes
- There was a built-in way to go "up a level" by cancelling the current action
- The navigation graph is built completely out of relations

## Testing

- tests for the new `CompassQuadrant` / `CompassOctant` methods
- tests for the new directional navigation module


crates/bevy_input_focus/src/directional_navigation.rs
```diff
@@ -0,0 +1,379 @@
...
+    /// Get the neighbor for a given [`CompassOctant`].
+    pub const fn get(&self, octant: CompassOctant) -> Option<Entity> {
+        self.neighbors[octant.to_index()]
+    }
+
+    /// Set the neighbor for a given [`CompassOctant`].
+    pub const fn set(&mut self, octant: CompassOctant, entity: Entity) {
+        self.neighbors[octant.to_index()] = Some(entity);
+    }
+}
```
NthTensor, line 66: nice that these can all be const.

crates/bevy_input_focus/src/directional_navigation.rs
```diff
@@ -0,0 +1,379 @@
...
+/// - **Not self-connected**: An entity should not be a neighbor of itself; use [`None`] instead.
+///
+/// For now, this graph must be built manually, and the developer is responsible for ensuring that it meets the above criteria.
+#[derive(Resource, Debug, Default, Clone, PartialEq)]
+pub struct DirectionalNavigationMap {
+    /// A directed graph of focusable entities.
+    ///
+    /// Pass in the current focus as a key, and get back a collection of up to 8 neighbors,
+    /// each keyed by a [`CompassOctant`].
+    pub neighbors: EntityHashMap<NavNeighbors>,
```
NthTensor, line 87: Just a thought, could it make sense for `NavNeighbors` to be a component? Possibly an immutable one?

crates/bevy_input_focus/src/directional_navigation.rs
```diff
@@ -0,0 +1,379 @@
...
+    pub fn remove(&mut self, entity: Entity) {
+        self.neighbors.remove(&entity);
+
+        for node in self.neighbors.values_mut() {
+            for neighbor in node.neighbors.iter_mut() {
+                if *neighbor == Some(entity) {
+                    *neighbor = None;
+                }
+            }
+        }
```
NthTensor, line 108: If we forced the graph to be symmetric we could make this significantly more efficient (and constant-time bounded).

crates/bevy_input_focus/src/directional_navigation.rs
```diff
@@ -0,0 +1,379 @@
...
+    }
+
+    /// Adds a symmetrical edge between two entities in the navigation map.
+    /// The A -> B path will use the provided direction, while B -> A will use the [`CompassOctant::opposite`] variant.
+    ///
+    /// Any existing connections between the two entities will be overwritten.
+    pub fn add_symmetrical_edge(&mut self, a: Entity, b: Entity, direction: CompassOctant) {
+        self.add_edge(a, b, direction);
+        self.add_edge(b, a, direction.opposite());
+    }
```
NthTensor, line 158: As mentioned above, I think there might be good perf reasons to only have the symmetric case. It's hard for me to imagine a good UI that is not symmetric and the non-symmetric edges do have a real cost.

crates/bevy_input_focus/src/directional_navigation.rs
```diff
@@ -0,0 +1,379 @@
...
+    /// Add symettrical edges between all entities in the provided slice, looping back to the first entity at the end.
+    ///
+    /// This is useful for creating a circular navigation path between a set of entities, such as a menu.
+    pub fn add_looping_edges(&mut self, entities: &[Entity], direction: CompassOctant) {
+        for i in 0..entities.len() {
+            let a = entities[i];
+            let b = entities[(i + 1) % entities.len()];
+            self.add_symmetrical_edge(a, b, direction);
+        }
+    }
```
NthTensor, line 169: This is a nice little convenience.

NthTensor (approved):
I left a few comments about the implementation, but it looks great!

It's worth pointing out that I don't think this will work well for highly dynamic or (diegetic 3d) UIs. You'd need either (a) a very fast auto-generation framework or (b) a lot of manual management. So for stuff that moves around on screen in complex nonlinear ways, we'll need something else.

That said, I like this a lot. Most of the UIs we have seen people make are really static (especially the types we tend to see in game jams). Having something purpose-built for the static case that seems great.

crates/bevy_input_focus/src/directional_navigation.rs
NthTensor: @IQuick143 has pointed out to me that requiring symmetry may not be a good idea, and I will also admit that perf of graph modification isn't really a major concern.

I'll weaken this to an extremely surface-level nit: Perhaps we should make the symmetric edge fn (which I think is probably what most people will want most of the time) the "default" by making it shorter: `add_edge` and `add_asymetric_edge`? Just a thought.

crates/bevy_input_focus/src/directional_navigation.rs
```diff
@@ -0,0 +1,379 @@
...
+
+        assert_eq!(map.get_neighbor(a, CompassOctant::SouthEast), Some(b));
+        assert_eq!(
+            map.get_neighbor(b, CompassOctant::SouthEast.opposite()),
+            None
+        );
+    }
+
+    #[test]
+    fn symettrical_edges() {
```
skimmedsquare, line 273: Nit: `symettrical` -> `symmetrical`

crates/bevy_math/src/compass.rs
```diff
@@ -72,6 +113,57 @@ pub enum CompassOctant {
...
+            _ => None,
+        }
+    }
+
+    /// Converts a [`CompassOctant`] to a standard index.
+    ///
+    /// Starts at 0 for [`CompassOctant::North`] and increments clockwise.
+    pub const fn to_index(self) -> usize {
+        match self {
+            Self::North => 0,
```
skimmedsquare, line 139: Random thought I had, (this may be irrelevant or out of scope for this change) - is it worth foregoing the nice index integer incrementing and instead making the cardinal directions match up between `CompassOctant` and `CompassQuadrant`? I'm thinking of a scenario where the integer data is serialized & conversion between the two becomes a bit frustrating for the user of the API.

IQuick143: I think it's better to just have a `From`+`TryFrom` impls for the conversions, the indices are better to have a reasonable order, because they'll be used as indices to arrays.

viridia:
An alternate approach: rather than a fixed length array of neighbors, have a `Vec<(CompassOctant, Entity)>`. This avoids the need to convert compass directions into indices. However, the real reason for doing it this way is that it's more "relation-like", meaning that migrating to relations will be a smaller step.

crates/bevy_input_focus/src/directional_navigation.rs
```diff
@@ -0,0 +1,379 @@
...
+    ///
+    /// Any existing connections between the two entities will be overwritten.
+    pub fn add_symmetrical_edge(&mut self, a: Entity, b: Entity, direction: CompassOctant) {
+        self.add_edge(a, b, direction);
+        self.add_edge(b, a, direction.opposite());
+    }
+
+    /// Add symettrical edges between all entities in the provided slice, looping back to the first entity at the end.
+    ///
+    /// This is useful for creating a circular navigation path between a set of entities, such as a menu.
```
IQuick143, line 162: Docs should note that these edges are symmetrical.

alice-i-cecile:
> It's worth pointing out that I don't think this will work well for highly dynamic or (diegetic 3d) UIs. You'd need either (a) a very fast auto-generation framework or (b) a lot of manual management. So for stuff that moves around on screen in complex nonlinear ways, we'll need something else.

Yep, I think that this is best suited for highly tuned mostly static UIs. For extremely dynamic stuff, you'll be better off with a heuristic combining distance and position like Aevyrie suggested IMO :)

crates/bevy_input_focus/src/directional_navigation.rs
alice-i-cecile: They do? It's the second (misspelled) word.

crates/bevy_input_focus/src/directional_navigation.rs
alice-i-cecile: Yeah, I chewed on this a fair bit. IMO until we have relations we shouldn't try to encode these graphs via components.

crates/bevy_input_focus/src/directional_navigation.rs
alice-i-cecile: Yeah, sadly I don't think that's viable. As discussed on Discord, the "three small elements beside one big one" layout is not symmetric.

crates/bevy_input_focus/src/directional_navigation.rs
alice-i-cecile: I think that symmetric edges present too much of a footgun to make the default :)

NthTensor: I'm ok with that. What do you think about allowing multiple vertices of the focus graph to correspond to the same ui element entity, instead of making them 1-to-1? Later in the same discussion we ran into some situations where this seemed desireable.

crates/bevy_math/src/compass.rs
alice-i-cecile: Would you like me to add From + TryFrom traits too? I'd like to keep the existing methods since they're const.

alice-i-cecile:
> An alternate approach: rather than a fixed length array of neighbors, have a Vec<(CompassOctant, Entity)>. This avoids the need to convert compass directions into indices. However, the real reason for doing it this way is that it's more "relation-like", meaning that migrating to relations will be a smaller step.

I think that the current approach is a bit nicer. I don't expect the conversion to relations to be particularly challenging, and if it is we can do this refactor just before that change.

NthTensor:
Yeah I'm personally a fan of the octant approach. I can see how it would make for a clean editor UI as well.

alice-i-cecile:
We could also do eight fields on the struct. I'm not sure if that's faster / cleaner though. The little index -> compass octant helpers seem generally quite nice.

crates/bevy_input_focus/src/directional_navigation.rs
alice-i-cecile: I think that's reasonable, but best left to follow-up. This is the sort of situation that was pushing me towards having a history,  but I think there's a bunch of design to chew on.

NthTensor: Totally fair. This has good bones, I think we should run with it.

crates/bevy_math/src/compass.rs
skimmedsquare: I'm fine either way - I'll defer to your judgement here. I think I was partially confused too and thought that the `NavNeighbors` `get/set` took a `CompassQuadrant` as well - though I do think it's a potential pitfall that `NavNeighbors.neighbors[CompassOctant::East.to_index()] != NavNeighbors.neighbors[CompassQuadrant::East.to_index()]`

crates/bevy_input_focus/src/directional_navigation.rs
IQuick143: Never mind me, I just cannot read, my bad.

crates/bevy_math/src/compass.rs
alice-i-cecile: Yeah, I think that the strongly typed helper methods are going to be preferred so we should be safe.

IQuick143:
Personally not a huge fan of eight fields on a struct, I think the best type-safe option would be to make a newtype around `[T; 8]` (where T is some relevant type) and implement `Index<CompassOctant>` and such.
Fields on a struct just lead to the same memory layout yet worse accessing semantics, because even if you know the octant/offset, you have to name the variable and use something like a match arm.

crates/bevy_input_focus/src/directional_navigation.rs
```diff
@@ -0,0 +1,164 @@
...
+    ///
+    /// An [`EntityHashSet`] must be provided as it is noticeably faster than the standard hasher or a [`Vec`].
+    pub fn remove_multiple(&mut self, entities: EntityHashSet) {
+        for entity in &entities {
+            self.neighbors.remove(entity);
+        }
+
+        for node in self.neighbors.values_mut() {
+            for neighbor in node.neighbors.iter_mut() {
+                if let Some(entity) = *neighbor {
```
andriyDev, line 125: Consider using let-else syntax for less nesting!

crates/bevy_input_focus/src/directional_navigation.rs
```diff
@@ -0,0 +1,164 @@
...
+    /// The currently focused entity.
+    pub focus: ResMut<'w, InputFocus>,
+    /// The navigation map containing the connections between entities.
+    pub map: Res<'w, DirectionalNavigationMap>,
+}
+
+impl DirectionalNavigation<'_> {
+    /// Navigates to the neighbor in a given direction from the current focus, if any.
+    pub fn navigate(&mut self, octant: CompassOctant) {
+        if let Some(current_focus) = self.focus.0 {
```
andriyDev, line 207: Let-else here and the next line perhaps

crates/bevy_input_focus/src/directional_navigation.rs
```diff
@@ -0,0 +1,379 @@
...
+}
+
+impl DirectionalNavigation<'_> {
+    /// Navigates to the neighbor in a given direction from the current focus, if any.
+    ///
+    /// Returns the new focus if successful.
+    /// Returns an error if there is no focus set or if there is no neighbor in the requested direction.
+    ///
+    /// If the result was `Ok`, the [`InputFocus`] resource is updated to the new focus as part of this method call.
+    pub fn navigate(
```
andriyDev, line 203: This is for another PR, but I wonder if we should also consider "adjacent octants". It may be confusing to users if they accidentally pressed up-and-left on a UI with just a left and their action had no effect. I recognize this is ambiguous (what if you have both a left and an up but no top-left), but I think picking any direction is better than doing nothing.

crates/bevy_input_focus/src/directional_navigation.rs
```diff
@@ -0,0 +1,379 @@
...
+    pub fn get_neighbors(&self, entity: Entity) -> Option<&NavNeighbors> {
+        self.neighbors.get(&entity)
+    }
+}
+
+/// A system parameter for navigating between focusable entities in a directional way.
+#[derive(SystemParam, Debug)]
+pub struct DirectionalNavigation<'w> {
+    /// The currently focused entity.
+    pub focus: ResMut<'w, InputFocus>,
```
andriyDev, line 191: For another PR: we might want to consider making `InputFocus` a component. It would be nice to support multiple "cursors". For example, in a couch-coop game you may want to vote for the next minigame or something - having independent cursors for each player would be nice!

andriyDev (approved):

crates/bevy_input_focus/src/directional_navigation.rs
alice-i-cecile: We've talked a bit about multi-focus before :) It's relatively niche, and we're worried that it will complicate the happy path for handling UI. Open to having my mind changed by a design in the future though.

viridia: Agreed; there are always going to be minority use cases that don't fit within the model. Fortunately, it's not that hard to roll your own implementation, which doesn't require any changed to Bevy. This is an area where third-party crates can fill the gaps.

PR #11870, opened 2024-02-15
Fix asset loader registration warning

thepackett:
# Objective

When registering and preregistering asset loaders, there would be a `warn!` if multiple asset loaders use a given extension, and an `info!` if multiple asset loaders load the same asset type. Since both of these situations are individually fine, it was decided that these messages should be removed.

## Solution

Replace both of these messages with a new `warn!` that notes that if multiple asset loaders share the same asset type _and_ share extensions, that the loader must be specified in the `.meta` file for those assets in order to solve the ambiguity. This is a more useful message, since it notes when a user must take special action / consideration.

PR #17105, opened 2025-01-02
Fix duplicate asset loader registration warning

arunke:
# Objective

The original fix (bevyengine/bevy#11870) did not actually implement the described logic. It checked if there were independently multiple loaders for a given asset type and multiple loaders for a given extension. However, this did not handle the case where those loaders were not the same. For example, there could be a loader for type `Foo` and extension `.foo`. Anther loader could exist for type `Bar` but extension `.bar`. If a third loader was added for type `Foo` but extension `.bar`, the warning would have been incorrectly logged.

## Solution

Instead of independently checking to see if there are preexisting loaders for both the extension and type, look up the indices of the loaders for the type in question. Then check to see if the loaders registered for the extensions has any overlap. Only log if there are loaders that fit this criteria.

## Testing

Ran CI tests. Locally tested the situation describe in the objective section for the normal `App::init_asset_loader` flow. I think testing could be done on the pre-registration flow for loaders still. I tested on Windows, but the changes should not be affected by platform.

### DIFF
--- a/crates/bevy_asset/src/server/loaders.rs
+++ b/crates/bevy_asset/src/server/loaders.rs
@@ -46,6 +46,7 @@ impl AssetLoaders {
             };
 
         if is_new {
+            let existing_loaders_for_type_id = self.type_id_to_loaders.get(&loader_asset_type);
             let mut duplicate_extensions = Vec::new();
             for extension in AssetLoader::extensions(&*loader) {
                 let list = self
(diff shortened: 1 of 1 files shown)


alice-i-cecile:
@thepackett can I get your review here please?

BenjaminBrienen (approved):

alice-i-cecile (approved):

PR #17106, opened 2025-01-02
Fix documentation for system set method

arunke:
# Objective

Fix incorrect comment on `IntoSystemSetConfigs::after` likely caused by copy-paste error. It said "before" instead of "after".

## Solution

Update the comment to the correct text.

## Testing

CI tests pass. This is just updating a comment.


alice-i-cecile (approved):

alice-i-cecile:
Thanks :)

Issue #17111, opened 2025-01-03
Warn the `clippy::allow_attributes` and `clippy::allow_attributes_without_reason` lints


LikeLakers2:
This is a tracking issue for warning on the `clippy::allow_attributes` and `clippy::allow_attributes_without_reason` lints across all bevy crates.

### `clippy::allow_attributes`
This lints against any instances of `#[allow(...)]`, suggesting they be changed to `#[expect(...)]`.

The use case for this is to avoid letting `#[allow(...)]`s pile up within Bevy's database, especially since `#[expect(...)]` is now available. By using `#![warn(clippy::allow_attributes)]`, we can remove any unnecessary instances of `#[allow(...)]`.

Do note that it's not possible to forbid this lint - there are instances where we are forced to `#[allow(...)]`, as the lints may not always be triggered (such as in macros for tuples).

Notably, this **does not** lint against `#![allow(...)]` (notice the `!`), per the lint's description:
> This lint only warns outer attributes (`#[allow]`), as inner attributes (`#![allow]`) are usually used to enable or disable lints on a global scale.

However, I will make an attempt to look for instances of `#![allow(...)]` anyways... if I remember.

See also: 
https://rust-lang.github.io/rust-clippy/master/index.html#allow_attributes

### `clippy::allow_attributes_without_reason`
This lints against any instances of `#[allow(...)]` or `#[expect(...)]` that do not include a `reason = "..."` field.

This ensures that if an expect or allow is used, it is given a reason - which helps to document why it's there in the first place. 

Notably, this **does not** lint against `#[warn(...)]` or `#[deny(...)]`.

See also:
https://rust-lang.github.io/rust-clippy/master/index.html#allow_attributes_without_reason

## Checklist
PRs marked as drafts in the list below mean that something (i.e. another PR) blocking them. Once the blocking PRs are merged, each applicable PR will be updated to fix any conflicts, and marked as ready to merge.

- [x] For each Bevy crate, create a PR where we add `#![warn(clippy::allow_attributes, clippy::allow_attributes_without_reason)]` to the root, and modify the rest of the crate as necessary to fulfill these new restrictions.
    - bevy_a11y
        https://github.com/bevyengine/bevy/pull/17093
    - bevy_animation
        https://github.com/bevyengine/bevy/pull/17094
    - bevy_app
        https://github.com/bevyengine/bevy/pull/17095
    - bevy_asset
        https://github.com/bevyengine/bevy/pull/17113
    - bevy_audio
[...]

PR #17113, opened 2025-01-03
bevy_asset: Apply `#![deny(clippy::allow_attributes, clippy::allow_attributes_without_reason)]`

LikeLakers2:
# Objective

## Solution
Set the `clippy::allow_attributes` and `clippy::allow_attributes_without_reason` lints to `deny`, and bring `bevy_asset` in line with the new restrictions.

## Testing
`cargo clippy` and `cargo test --package bevy_asset --features multi_threaded` were run, and no errors were encountered.

### DIFF
--- a/crates/bevy_asset/src/lib.rs
+++ b/crates/bevy_asset/src/lib.rs
@@ -139,6 +139,7 @@
 //! This trait mirrors [`AssetLoader`] in structure, and works in tandem with [`AssetWriter`](io::AssetWriter), which mirrors [`AssetReader`](io::AssetReader).
 
 #![expect(missing_docs, reason = "Not all docs are written yet, see #3492.")]
+#![deny(clippy::allow_attributes, clippy::allow_attributes_without_reason)]
 #![cfg_attr(docsrs, feature(doc_auto_cfg))]
 #![doc(
     html_logo_url = "https://bevyengine.org/assets/icon.png",
@@ -1767,8 +1768,11 @@ mod tests {
     #[derive(Asset, TypePath)]
     pub struct TestAsset;
 
-    #[allow(dead_code)]
     #[derive(Asset, TypePath)]
+    #[expect(
+        dead_code,
+        reason = "This exists to ensure that `#[derive(Asset)]` works on enums. The inner variants are known not to be used."
+    )]
     pub enum EnumTestAsset {
         Unnamed(#[dependency] Handle<TestAsset>),
         Named {
@@ -1783,7 +1787,6 @@ mod tests {
         Empty,
     }
 
-    #[allow(dead_code)]
     #[derive(Asset, TypePath)]
     pub struct StructTestAsset {
         #[dependency]
@@ -1792,7 +1795,6 @@ mod tests {
         embedded: TestAsset,
     }
 
-    #[allow(dead_code)]
     #[derive(Asset, TypePath)]
     pub struct TupleTestAsset(#[dependency] Handle<TestAsset>);
 }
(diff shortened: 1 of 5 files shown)


crates/bevy_asset/src/lib.rs
```diff
@@ -1765,8 +1766,11 @@ mod tests {
     #[derive(Asset, TypePath)]
     pub struct TestAsset;
 
-    #[allow(dead_code)]
     #[derive(Asset, TypePath)]
+    #[expect(
+        dead_code,
+        reason = "This exists to ensure that `#[derive(Asset)]` works on enums. The inner variants are known not to be used."
+    )]
```
LikeLakers2, line 1773: I have no idea why this only lints on the enum test case, despite the struct test cases also being unused.

crates/bevy_asset/src/server/loaders.rs
```diff
@@ -342,17 +342,14 @@ mod tests {
     use super::*;
 
     // The compiler notices these fields are never read and raises a dead_code lint which kill CI.
-    #[allow(dead_code)]
     #[derive(Asset, TypePath, Debug)]
-    struct A(usize);
+    struct A;
```
LikeLakers2, line 346: In case you're wondering: I checked and double-checked to make sure that removing these `usize` fields wouldn't affect tests.

If they affect tests, I didn't see it during my testing.

crates/bevy_asset/src/server/loaders.rs
```diff
@@ -342,17 +342,14 @@ mod tests {
     use super::*;
 
     // The compiler notices these fields are never read and raises a dead_code lint which kill CI.
```
BenjaminBrienen, line 344: out of date comment

crates/bevy_asset/src/server/loaders.rs
BenjaminBrienen: I don't think the tests are meant to test specifically unit struct assets. Unit struct assets don't sound very useful anyway.

BenjaminBrienen (changes requested):

BenjaminBrienen:
In benches, we silence dead code warnings due to similar dead_code unit tests in this PR.

crates/bevy_asset/src/server/loaders.rs
LikeLakers2: my bad
```suggestion
```

crates/bevy_asset/src/server/loaders.rs
LikeLakers2: The tests compile and finish successfully either way - so I'm not sure the `(usize)` here matters.

If we start needing a value, they can be added back.

LikeLakers2:
@BenjaminBrienen These aren't benches - thus, I'm not going to treat them like benches.

BenjaminBrienen:
It's fine to me if we fix the dead code warning by removing the fields, but I wonder if it might make sense to instead use those fields in the unit tests. If not, nevermind me.

LikeLakers2:
@BenjaminBrienen Adding a new test that uses those fields would probably make more sense - but the addition of tests is out-of-scope for this PR.

My job here is to apply `#![deny(clippy::allow_attributes, clippy::allow_attributes_without_reason)]` to bevy_asset.

crates/bevy_asset/src/server/loaders.rs
IQuick143: It's true that as far as I can see there seems to be no behavior in the tests hinging on the structs having any fields. However, idk if they're relevant for `TypeId` calculations.
It would be best to ask someone who wrote this or knows about reflection.

LikeLakers2: @IQuick143 I don't believe they're relevant for `TypeId` calculations: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=99bbd15c51d685da51caa2a4beecf51b

I mean, it appears that changing them to include `(usize)` does change their `TypeId` - but two unit structs still have two different `TypeId`s.

BenjaminBrienen: In that case, I think this fix is good :)

BenjaminBrienen (approved):

BenjaminBrienen (approved):

PR #17115, opened 2025-01-03
Optimize `Entities::entity_does_not_exist_error_details_message`, remove `UnsafeWorldCell` from error

JaySpruce:
## Objective

The error `EntityFetchError::NoSuchEntity` has an `UnsafeWorldCell` inside it, which it uses to call `Entities::entity_does_not_exist_error_details_message` when being printed. That method returns a `String` that, if the `track_location` feature is enabled, contains the location of whoever despawned the relevant entity.

I initially had to modify this error while working on #17043. The `UnsafeWorldCell` was causing borrow problems when being returned from a command, so I tried replacing it with the `String` that the method returns, since that was the world cell's only purpose.

Unfortunately, `String`s are slow, and it significantly impacted performance (on top of that PR's performance hit):
<details>
<summary>17043 benchmarks</summary>

### With `String`
error_handling_insert_slow

### No `String`
error_handling_insert_fixed

</details>

For that PR, I just removed the error details entirely, but I figured I'd try to find a way to keep them around.

## Solution

- Replace the `String` with a helper struct that holds the location, and only turn it into a string when someone actually wants to print it.
- Replace the `UnsafeWorldCell` with the aforementioned struct.
- Do the same for `QueryEntityError::NoSuchEntity`.

## Benchmarking

This had some interesting performance impact:

<details>
<summary>This PR vs main</summary>

dne_rework_1
dne_rework_2
dne_rework_3

</details>

## Other work

`QueryEntityError::QueryDoesNotMatch` also has an `UnsafeWorldCell` inside it. This one would be more complicated to rework while keeping the same functionality.

## Migration Guide

The errors `EntityFetchError::NoSuchEntity` and `QueryEntityError::NoSuchEntity` now contain an `EntityDoesNotExistDetails` struct instead of an `UnsafeWorldCell`. If you were just printing these, they should work identically.

### DIFF
--- a/crates/bevy_ecs/src/entity/mod.rs
+++ b/crates/bevy_ecs/src/entity/mod.rs
@@ -995,7 +995,10 @@ impl Entities {
         &self,
         _entity: Entity,
     ) -> EntityDoesNotExistDetails {
-        #[cfg(feature = "track_location")]
+        return EntityDoesNotExistDetails {
+            #[cfg(feature = "track_location")]
+            location: self.entity_get_spawned_or_despawned_by(_entity),
+        };
         return EntityDoesNotExistDetails {
             location: self.entity_get_spawned_or_despawned_by(_entity),
         };
--- a/crates/bevy_ecs/src/entity/mod.rs
+++ b/crates/bevy_ecs/src/entity/mod.rs
@@ -999,11 +999,6 @@ impl Entities {
             #[cfg(feature = "track_location")]
             location: self.entity_get_spawned_or_despawned_by(_entity),
         };
-        return EntityDoesNotExistDetails {
-            location: self.entity_get_spawned_or_despawned_by(_entity),
-        };
-        #[cfg(not(feature = "track_location"))]
-        return EntityDoesNotExistDetails {};
     }
 }
 
--- a/crates/bevy_ecs/src/entity/mod.rs
+++ b/crates/bevy_ecs/src/entity/mod.rs
@@ -995,10 +995,10 @@ impl Entities {
         &self,
         _entity: Entity,
     ) -> EntityDoesNotExistDetails {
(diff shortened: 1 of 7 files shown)


crates/bevy_ecs/src/world/error.rs
```diff
@@ -25,53 +27,17 @@ pub enum EntityComponentError {
 }
 
 /// An error that occurs when fetching entities mutably from a world.
-#[derive(Clone, Copy)]
-pub enum EntityFetchError<'w> {
+#[derive(Error, Debug, Clone, Copy)]
+pub enum EntityFetchError {
```
alice-i-cecile, line 31: Yay no lifetime!

alice-i-cecile (approved):
Much cleaner. I'd like to merge this first.

crates/bevy_ecs/src/entity/mod.rs
```diff
@@ -991,19 +991,40 @@ impl Entities {
     }
 
     /// Constructs a message explaining why an entity does not exists, if known.
-    pub(crate) fn entity_does_not_exist_error_details_message(&self, _entity: Entity) -> String {
+    pub(crate) fn entity_does_not_exist_error_details_message(
+        &self,
+        _entity: Entity,
+    ) -> EntityDoesNotExistDetails {
         #[cfg(feature = "track_location")]
```
BenjaminBrienen, line 998:
```suggestion
        return EntityDoesNotExistDetails {
            #[cfg(feature = "track_location")]
            location: self.entity_get_spawned_or_despawned_by(_entity),
        };
```

BenjaminBrienen (approved):
I suggested a small code style improvement. Otherwise, looks good!

BenjaminBrienen (approved):

PR #17116, opened 2025-01-03
Disallow requesting write resource access in Queries

vil-mo:
Related to https://github.com/bevyengine/bevy/pull/16843

Since `WorldQuery::Fetch` is `Clone`, it can't store mutable references to resources, so it doesn't make sense to mutably access resources. In that sense, it is hard to find usecases of mutably accessing resources and to clearly define, what mutably accessing resources would mean, so it's been decided to disallow write resource access.
Also changed documentation of safety requirements of `WorldQuery::init_fetch` and `WorldQuery::fetch` to clearly state to the caller, what safety invariants they need to uphold.

### DIFF
--- a/crates/bevy_ecs/src/query/state.rs
+++ b/crates/bevy_ecs/src/query/state.rs
@@ -199,6 +199,8 @@ impl<D: QueryData, F: QueryFilter> QueryState<D, F> {
             }
         }
 
+        debug_assert!(!state.component_access.access().has_any_resource_write(), "Mutable resource access in queries is not allowed");
+
         state
     }
 
--- a/crates/bevy_ecs/src/query/state.rs
+++ b/crates/bevy_ecs/src/query/state.rs
@@ -199,14 +199,6 @@ impl<D: QueryData, F: QueryFilter> QueryState<D, F> {
             }
         }
 
-        if state.component_access.access().has_write_all_resources() {
-            access.write_all_resources();
-        } else {
-            for component_id in state.component_access.access().resource_writes() {
-                access.add_resource_write(world.initialize_resource_internal(component_id).id());
-            }
-        }
-
         state
     }
 
--- a/crates/bevy_ecs/src/query/state.rs
+++ b/crates/bevy_ecs/src/query/state.rs
@@ -199,7 +199,10 @@ impl<D: QueryData, F: QueryFilter> QueryState<D, F> {
             }
         }
 
-        debug_assert!(!state.component_access.access().has_any_resource_write(), "Mutable resource access in queries is not allowed");
+        debug_assert!(
+            !state.component_access.access().has_any_resource_write(),
+            "Mutable resource access in queries is not allowed"
+        );
 
         state
     }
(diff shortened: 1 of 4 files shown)


crates/bevy_ecs/src/query/world_query.rs
```diff
@@ -60,11 +60,14 @@ pub unsafe trait WorldQuery {
     fn shrink_fetch<'wlong: 'wshort, 'wshort>(fetch: Self::Fetch<'wlong>) -> Self::Fetch<'wshort>;
 
     /// Creates a new instance of this fetch.
+    /// Readonly accesses resourses registered in [`update_component_access`].
```
LikeLakers2, line 63:
```suggestion
    /// Readonly accesses resources registered in [`update_component_access`].
```

BenjaminBrienen (approved):

crates/bevy_ecs/src/query/state.rs
```diff
@@ -199,14 +199,6 @@ impl<D: QueryData, F: QueryFilter> QueryState<D, F> {
             }
         }
 
-        if state.component_access.access().has_write_all_resources() {
```
chescock, line 202: It might be worth adding a check like

```rust
debug_assert!(!state.component_access.access().has_any_resource_write());
```

so that anyone who tries it gets an error.

chescock (approved):
Yeah, I think this is the right move.  I only added the code for mutable resources because it looked simple and not because we had a real use case, and it's clear that I was wrong about it being simple.  

We should also remove [the `has_resource_write()` check in `AssetChanged`](https://github.com/bevyengine/bevy/pull/16810/files#diff-7e8eda4ef0e59b96f20c695f7394df2de7cfcb6185a26eb135ab197029a19b86R236-R240).  It's the only first-party use of resource access in queries, so users might use it as an example, and leaving that check there might trick them into thinking that mutable resource access is supported.

vil-mo:
> We should also remove [the `has_resource_write()` check in `AssetChanged`](https://github.com/bevyengine/bevy/pull/16810/files#diff-7e8eda4ef0e59b96f20c695f7394df2de7cfcb6185a26eb135ab197029a19b86R236-R240).  It's the only first-party use of resource access in queries, so users might use it as an example, and leaving that check there might trick them into thinking that mutable resource access is supported.  

It should not be removed, because user could request write access to the resource in the `SystemParam` prior to the query

chescock:
> It should not be removed, because user could request write access to the resource in the `SystemParam` prior to the query

I believe that's checked in `<Query as SystemParam>::init_state()` after the whole query is built.  The check in `<AssetChanged as WorldQuery>::update_component_access()` only has the `FilteredAccess` from the current query, not the `FilteredAccessSet` that includes other system params' access.

vil-mo:
> I believe that's checked in `<Query as SystemParam>::init_state()` after the whole query is built

It's not checked there
```rust
fn init_state(world: &mut World, system_meta: &mut SystemMeta) -> Self::State {
    let state = QueryState::new_with_access(world, &mut system_meta.archetype_component_access);
    init_query_param(world, system_meta, &state);
    state
}
```

chescock:
The check is in `init_query_param`.  It's in a method so it can be shared by `QueryParamBuilder`.

alice-i-cecile:
Note that write access to resources during queries may be useful for certain index updating strategies. That said, it doesn't seem to actually work, so we should clean this up for now and re-add it later when we have clearer requirements.

PR #16338, opened 2024-11-10
Remove all `deprecated` code

BenjaminBrienen:
# Objective

Release cycle things

## Solution

Delete items deprecated in 0.15 and migrate bevy itself.

## Testing

CI

PR #17119, opened 2025-01-03
bevy_audio: Apply `#![deny(clippy::allow_attributes, clippy::allow_attributes_without_reason)]`

LikeLakers2:
# Objective

## Solution
Set the `clippy::allow_attributes` and `clippy::allow_attributes_without_reason` lints to `deny`, and bring `bevy_audio` in line with the new restrictions.

## Testing
`cargo clippy` and `cargo test --package bevy_audio` were run, and no errors were encountered.


crates/bevy_audio/src/audio.rs
```diff
@@ -267,6 +274,15 @@ impl AudioPlayer<AudioSource> {
...
 )]
+// NOTE: When removing this, please remove the `#[expect(deprecated)]` at the top of this file.
+//
+// For whatever reason, this struct counts as a use of itself, causing rustc to lint about a use of
+// a deprecated struct. However, adding an `#[expect(deprecated)]` to this strict causes rustc to
+// say that the lint expectation is unfulfilled.
+//
+// The only solution I could find to this was to mark the whole module as expecting a deprecated
+// lint. So, if this item is ever removed, please remove the `#[expect(deprecated)]` at the top of
+// this file.
```
LikeLakers2, line 285: I'm gonna be honest: I have NO idea what's happening with rustc's linter here. It's marking a deprecated item's definition as a use of a deprecated item, *and* it's marking the impls for that deprecated item as a use of that deprecated item.

But I tried my best to silence the lints, while simultaneously changing all `allow`s to `expect`s, and giving them all `reason = "..."`.

So I hope this comment and the module-wide `expect` are okay, because they were the best I could do.

crates/bevy_audio/src/audio.rs
LikeLakers2, line 1: I'm gonna be honest: I have NO idea what's happening with rustc's linter here. It's marking a deprecated item's definition as a use of a deprecated item, *and* it's marking the impls for that deprecated item as a use of that deprecated item.

But I tried my best to silence the lints, while simultaneously changing all `allow`s to `expect`s, and giving them all `reason = "..."`.

LikeLakers2: I've figured it out - it's the `#[derive(Bundle)]` placed atop `AudioSourceBundle`.

That said, I have *no* idea what to do to fix that derive to make it not generate `deprecated` warnings. Even adding `#[automatically_derived]` to the two impls it generates doesn't do much.

BenjaminBrienen:
I disagree with your interpretation of rustc's deprecation behavior. An impl is a usage. I think this should be blocked on #16338, which would clean-up this changeset.

LikeLakers2:
@BenjaminBrienen Ok. I don't really care much either way - just notify me when that PR is merged and I'll update this PR.

LikeLakers2:
This PR is blocked on https://github.com/bevyengine/bevy/pull/16338 per Benjamin's request. Once that is merged, I will fix any conflicts and mark it as ready to merge.

IQuick143 (approved):

BenjaminBrienen (approved):

PR #17120, opened 2025-01-03
More DefaultUiCamera fixes

ickshonpe:
# Objective

Found more excessive `DefaultUiCamera` queries outside of extraction.
The default UI camera lookup only needs to be done once. Do it first, not per node.

### DIFF
--- a/crates/bevy_ui/src/widget/text.rs
+++ b/crates/bevy_ui/src/widget/text.rs
@@ -1,3 +1,5 @@
+use std::default;
+
 use crate::{
     ComputedNode, ContentSize, DefaultUiCamera, FixedMeasure, Measure, MeasureArgs, Node,
     NodeMeasure, TargetCamera, UiScale,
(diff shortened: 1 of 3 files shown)


crates/bevy_ui/src/widget/text.rs
```diff
@@ -1,3 +1,5 @@
+use std::default;
+
```
LikeLakers2, line 2: You forgot to remove this.
```suggestion
```

BenjaminBrienen (approved):
Great find!

rparrett (approved):
> The default UI camera lookup only needs to be done once.

Minor point -- I think it doesn't need to be done at all if the user is using `TargetCamera` for all their UI hierarchies.

But this should still be an optimization in that case too.

ickshonpe:
> > The default UI camera lookup only needs to be done once.
> 
> Minor point -- I think it doesn't need to be done at all if the user is using `TargetCamera` for all their UI hierarchies.
> 
> But this should still be an optimization in that case too.

Yep I thought about adding some sort of memoization, but it doesn't seem necessary as one query per system isn't going to be expensive.

PR #14284, opened 2024-07-11
Add custom cursors

eero-lehtinen:
# Objective

- Add custom images as cursors
- Fixes #9557 

## Solution

- Change cursor type to accommodate both native and image cursors
- I don't really like this solution because I couldn't use `Handle<Image>` directly. I would need to import `bevy_assets` and that causes a circular dependency. Alternatively we could use winit's `CustomCursor` smart pointers, but that seems hard because the event loop is needed to create those and is not easily accessable for users. So now I need to copy around rgba buffers which is sad.
- I use a cache because especially on the web creating cursor images is really slow
- Sorry to #14196 for yoinking,  I just wanted to make a quick solution for myself and thought that I should probably share it too.

Update:
- Now uses `Handle<Image>`, reads rgba data in `bevy_render` and uses resources to send the data to `bevy_winit`, where the final cursors are created.

## Testing

- Added example which works fine at least on Linux Wayland (winit side has been tested with all platforms).
- I haven't tested if the url cursor works.

## Migration Guide

- `CursorIcon` is no longer a field in `Window`, but a separate component can be inserted to a window entity. It has been changed to an enum that can hold custom images in addition to system icons.
- `Cursor` is renamed to `CursorOptions` and `cursor` field of `Window` is renamed to `cursor_options`
- `CursorIcon` is renamed to `SystemCursorIcon`

PR #17121, opened 2025-01-03
Support texture atlases in CustomCursor::Image

mgi388:
# Objective

- Bevy 0.15 added support for custom cursor images in https://github.com/bevyengine/bevy/pull/14284.
- However, to do animated cursors using the initial support shipped in 0.15 means you'd have to animate the `Handle<Image>`: You can't use a `TextureAtlas` like you can with sprites and UI images.
- For my use case, my cursors are spritesheets. To animate them, I'd have to break them down into multiple `Image` assets, but that seems less than ideal.

## Solution

- Allow users to specify a `TextureAtlas` field when creating a custom cursor image.
- To create parity with Bevy's `TextureAtlas` support on `Sprite`s and `ImageNode`s, this also allows users to specify `rect`, `flip_x` and `flip_y`. In fact, for my own use case, I need to `flip_y`.

## Testing

- I added unit tests for `calculate_effective_rect` and `extract_and_transform_rgba_pixels`.
- I added a brand new example for custom cursor images. It has controls to toggle fields on and off. I opted to add a new example because the existing cursor example (`window_settings`) would be far too messy for showcasing these custom cursor features (I did start down that path but decided to stop and make a brand new example).
- The new example uses a [Kenny cursor icon] sprite sheet. I included the licence even though it's not required (and it's CC0).
- I decided to make the example just loop through all cursor icons for its animation even though it's not a _realistic_ in-game animation sequence.
- I ran the PNG through https://tinypng.com. Looks like it's about 35KB.
- I'm open to adjusting the example spritesheet if required, but if it's fine as is, great.

[Kenny cursor icon]: https://kenney-assets.itch.io/crosshair-pack

---

## Showcase

https://github.com/user-attachments/assets/8f6be8d7-d1d4-42f9-b769-ef8532367749

## Migration Guide

The `CustomCursor::Image` enum variant has some new fields. Update your code to set them.

Before:

```rust
CustomCursor::Image {
    handle: asset_server.load("branding/icon.png"),
    hotspot: (128, 128),
}
```

After:

```rust
CustomCursor::Image {
    handle: asset_server.load("branding/icon.png"),
    texture_atlas: None,
    flip_x: false,
    flip_y: false,
    rect: None,
    hotspot: (128, 128),
}
```

## References

- Feature request [originally raised in Discord].

[originally raised in Discord]: https://discord.com/channels/691052431525675048/692572690833473578/1319836362219847681


mgi388:
Another showcase video. This is a more realistic in-game animated cursor. The pointer animates every 5s (blink and you'll miss it):

https://github.com/user-attachments/assets/4a15225c-2490-4333-abb7-3866f428d5f8

LikeLakers2:
> The new example uses a [Kenny cursor icon](https://kenney-assets.itch.io/crosshair-pack) sprite sheet. I included the licence even though it's not required (and it's CC0).

First off: Thank you for attributing to Kenney. He's a cool person. :D

Second off: I don't see any assets added by this PR? All the files touched are code, except for `examples/README.md`.

mgi388:
@eero-lehtinen 👋 would you be interested in reviewing this PR for texture atlas support on custom cursor images?

crates/bevy_winit/src/cursor.rs
```diff
@@ -192,26 +244,311 @@ fn on_remove_cursor_icon(trigger: Trigger<OnRemove, CursorIcon>, mut commands: C
...
-                .collect(),
-        ),
+                .collect::<Vec<u8>>();
+
+            Some(&image_data_as_u8s)
+        }
         _ => None,
+    };
+
+    let image_data = image_data?;
```
eero-lehtinen, line 278: I would like to keep a fast path when there is no trasformations and just return the data.

eero-lehtinen (changes requested):
Seems fine to me except for a couple of things.

crates/bevy_winit/Cargo.toml
```diff
@@ -18,7 +18,13 @@ serialize = ["serde", "bevy_input/serialize", "bevy_window/serialize"]
 android-native-activity = ["winit/android-native-activity"]
 android-game-activity = ["winit/android-game-activity"]
 
-custom_cursor = ["bevy_image", "bevy_asset", "bytemuck", "wgpu-types"]
+custom_cursor = [
+  "bevy_asset",
+  "bevy_image",
+  "bevy_sprite",
```
eero-lehtinen, line 24: It seems very undesireable to enable `bevy_sprite` because it enables the `SpritePlugin` in `DefaultPlugins` and includes all sprite stuff when 3d apps have no need for that.

eero-lehtinen (changes requested):

crates/bevy_winit/src/cursor.rs
mgi388: Makes sense, I think I can adjust this.

crates/bevy_winit/Cargo.toml
mgi388: Do you have any personal suggestions/prefernce on a better place for `TextureAtlas` and friends?

It seems like maybe this is going to need to be solved before this PR can go through, after all.

eero-lehtinen: I would move it to `bevy_image`. `TextureAtlas` looks like a pretty lean type so it wouldn't bloat too much and `bevy_image` includes a lot of texture stuff already.

mgi388: OK cool thanks. Per my PR desc comment:

> But bevy_image currently has no such Bevy plugin in there, and so at the very least it meant I needed to add a new plugin so that I could register TextureAtlasLayout asset (and register reflection, etc.). See [WIP branch](https://github.com/mgi388/bevy/commit/a186b5f348888e11b0015040888dd1c1d008d2bc#diff-b3dd4b626d8f0abfdc08f3de21a90b9fb4523eed3ee310e7dfd06e55bcfe3646R56).

The only thing that makes it less lean is the requirement for a Bevy plugin so we can register the types.

eero-lehtinen: Yeah I guess that makes it pretty bad. `bevy_image` probably shouldn't include `bevy_app`. Technically the type registration could just remain in `bevy_sprite` and also be added in the custom cursor plugin. Then it is registered if either of them are included.

eero-lehtinen: A new crate might be the most sensible option, even if it's sad to see ever increasing number of crates.

mgi388: @alice-i-cecile I'm inclined to add a new micro crate `bevy_texture_atlas`. Could we get on board with this? Alternatively, I'm a bit unsure of how to proceed. Per this thread, and my PR desc, `bevy_image` _seems_ like a good choice but when you start doing it, it quickly feels like a bad choice.

alice-i-cecile: I think that's likely the right way, but @atlv24 has thought about this organization problem harder than I have.

mgi388: @BenjaminBrienen presumably for that suggestion it would that mean going through every block of code in `bevy_sprite` and putting it behind a feature (including the texture atlas stuff behind a `texture_atlas` feature, on by default), and then in `bevy_winit`, when adding the `bevy_sprite` dep, only enabling the `texture_atlas` feature?

If yes, I can't help but feel like that's going to be a hard change to design (I could be wrong). E.g. take the `Sprite` and `Anchor` struct. Do they get feature gated, or are they OK to end up in `bevy_winit`?

I'm conscious of bringing in more cooks, but @bushrat011899 I know you've also had some thoughts about `bevy_sprite` organization. Can you see a path forward here keeping `TextureAtlas` and friends in `bevy_sprite` and heavily feature-gating the crate? Feel free to nack if you don't feel you can comment on this!

mgi388: From @atlv24 via [Discord]:

> bevy_winit should not depend on bevy_sprite because bevy_sprite depends on bevy_render, and that would bring back the bevy_winit dependency on bevy_render which is very undesirable. bevy_image already depends on bevy_asset, which depends on bevy_app. just move sprite sheet and anything it needs into bevy_image, the bevy_app dependency is already there just make it explicit

Me:

> Great, thanks! The only thing to note with it being in bevy_image is that TextureAtlas needs to be registered (app.register_type::<TextureAtlas>() is currently being done in SpritePlugin) (and TextureAtlasLayout for that matter), so it needs a new plugin added, hence bevy_image ends up having a plugin added to it.

atlv24:

> yeah just add a plugin
> thats basically free

[Discord]: https://discord.com/channels/691052431525675048/692572690833473578/1326064294445449246

eero-lehtinen: That is the way to go then.

mgi388: Thank for the help everyone, hopefully this PR hits the mark: https://github.com/bevyengine/bevy/pull/17219

crates/bevy_winit/src/cursor.rs
mgi388: Done. There are technically some cases where texture atlas might be the full image rect, but I decided to not let those follow the fast path since it's not obvious to me how to check that properly.

BenjaminBrienen:
I don't have time to do a full code review, but I'll run the examples when I get to my laptop

BenjaminBrienen (approved):
custom_cursor_image works. window_settings actually seems broken here and on main.

mgi388:
> custom_cursor_image works. window_settings actually seems broken here and on main.

@BenjaminBrienen how so? Seems to work for me. You need to left/right click to cycle the cursor icon if you didn't already do that.

BenjaminBrienen:
> > custom_cursor_image works. window_settings actually seems broken here and on main.
> 
> @BenjaminBrienen how so? Seems to work for me. You need to left/right click to cycle the cursor icon if you didn't already do that.

See #17227

crates/bevy_winit/src/custom_cursor.rs
```diff
@@ -0,0 +1,490 @@
...
+        | TextureFormat::Rgba8Uint
+        | TextureFormat::Rgba8Sint => Some(image.data.clone()),
+        TextureFormat::Rgba32Float => Some(
+            image
+                .data
+                .chunks(4)
+                .map(|chunk| {
+                    let chunk = chunk.try_into().unwrap();
+                    let num = bytemuck::cast_ref::<[u8; 4], f32>(chunk);
+                    (num * 255.0) as u8
```
eero-lehtinen, line 91: This is actually an oversight from my previous code, but we should `round()` here to follow the same conventions as `bevy_color`, maybe even `clamp()` too. https://github.com/bevyengine/bevy/blob/b20e23dd415b379ca2d91f1a9832c11087d42a15/crates/bevy_color/src/linear_rgba.rs#L309

eero-lehtinen (approved):
Nice tests, looks good.

crates/bevy_winit/src/custom_cursor.rs
mgi388: Thanks, updated. Tested by hacking the new example to load a .exr image, and I observed that it worked. And I added temporary logging to this branch to prove that it was followed.

```
cargo run --example custom_cursor_image --features=custom_cursor,exr
```

image

Image from https://openexr.com/en/latest/test_images/TestImages/GammaChart.html

https://github.com/user-attachments/assets/0be65d64-5992-48e8-9384-4bad06d3eebd

eero-lehtinen (approved):

PR #17126, opened 2025-01-03
Update `typos`

rparrett:
# Objective

Use the latest version of `typos` and fix the typos that it now detects

# Additional Info

By the way, `typos` has a "low priority typo suggestions issue" where we can throw typos we find that `typos` doesn't catch.

(This link may go stale) https://github.com/crate-ci/typos/issues/1200

### DIFF
--- a/crates/bevy_core_pipeline/src/smaa/smaa.wgsl
+++ b/crates/bevy_core_pipeline/src/smaa/smaa.wgsl
@@ -44,7 +44,7 @@
  * Here you'll find instructions to get the shader up and running as fast as
  * possible.
  *
- * IMPORTANTE NOTICE: when updating, remember to update both this file and the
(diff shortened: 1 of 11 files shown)


crates/bevy_core_pipeline/src/smaa/smaa.wgsl
```diff
@@ -44,7 +44,7 @@
  * Here you'll find instructions to get the shader up and running as fast as
  * possible.
  *
- * IMPORTANTE NOTICE: when updating, remember to update both this file and the
+ * IMPORTANT NOTICE: when updating, remember to update both this file and the
```
rparrett, line 47: Slightly sad to see this one go, but it's either this or `NOTICE IMPORTANTE`.

mockersf (approved):

PR #17131, opened 2025-01-03
Make the `get` function on `InstanceInputUniformBuffer` less error prone

Bleachfuel:
# Objective

the `get` function on [`InstanceInputUniformBuffer`] seems very error-prone. This PR hopes to fix this. 

## Solution

Do a few checks to ensure the index is in bounds and that the `BDI` is not removed.
Return `Option<BDI>` instead of `BDI`. 

## Testing

added a test to verify that the instance buffer works correctly

## Future Work
Performance decreases when using .binary_search(). However this is likely due to the fact that [`InstanceInputUniformBuffer::get`] for now is never used, and only get_unchecked.

## Migration Guide
`InstanceInputUniformBuffer::get` now returns `Option<BDI>` instead of `BDI` to reduce panics. If you require the old functionality of `InstanceInputUniformBuffer::get` consider using  `InstanceInputUniformBuffer::get_unchecked`.

### DIFF
--- a/crates/bevy_render/src/batching/gpu_preprocessing.rs
+++ b/crates/bevy_render/src/batching/gpu_preprocessing.rs
@@ -219,12 +219,12 @@ where
     ///
     /// Returns [`None`] if the index is out of bounds or the data is removed.
     pub fn get(&self, uniform_index: u32) -> Option<BDI> {
-        if (uniform_index as usize) < self.buffer.len()
-            || !self.free_uniform_indices.contains(&uniform_index)
+        if (uniform_index as usize) >= self.buffer.len()
+            || self.free_uniform_indices.contains(&uniform_index)
         {
-            Some(self.get_unchecked(uniform_index))
-        } else {
             None
+        } else {
+            Some(self.get_unchecked(uniform_index))
         }
     }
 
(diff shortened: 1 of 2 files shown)


BenjaminBrienen (approved):

IceSentry (approved):

crates/bevy_render/src/batching/gpu_preprocessing.rs
```diff
@@ -921,3 +941,21 @@ pub fn write_indirect_parameters_buffer(
         .write_buffer(&render_device, &render_queue);
     indirect_parameters_buffer.buffer.clear();
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn instance_buffer_correct_behaviour() {
```
LikeLakers2, line 950:
```suggestion
    fn instance_buffer_correct_behavior() {
```
Typos-cli apparently hates British English, lol.

Bleachfuel: Yeah hahahaha, Can i check typos-cli localy? And do you know if i have to open a differnt pr if i want to change the implementation from `contains()` to `binary_search()`. Because i think it's better for performance, but i'm unsure and this PR is small anyways.

BenjaminBrienen: There is a `typos` extension on VSCode from the same project. I'm sure if you look at their repository, they'll have more info.

PR #17137, opened 2025-01-04
bevy_core_pipeline: Apply `#![warn(clippy::allow_attributes, clippy::allow_attributes_without_reason)]`

LikeLakers2:
# Objective

## Solution
Set the `clippy::allow_attributes` and `clippy::allow_attributes_without_reason` lints to `warn`, and bring `bevy_core_pipeline` in line with the new restrictions.

## Testing
`cargo clippy` and `cargo test --package bevy_core_pipeline` were run, and no warnings were encountered.


BenjaminBrienen:
I'll review after https://github.com/bevyengine/bevy/pull/16338 merged and this is updated to fix the conflict :)

crates/bevy_core_pipeline/src/auto_exposure/compensation_curve.rs
```diff
@@ -136,7 +136,10 @@ impl AutoExposureCompensationCurve {
                 let lut_inv_range = 1.0 / (lut_end - lut_begin);
 
                 // Iterate over all LUT entries whose pixel centers fall within the current segment.
-                #[allow(clippy::needless_range_loop)]
+                #[expect(
+                    clippy::needless_range_loop,
+                    reason = "This for-loop also uses `i` to calculate a value `t`."
+                )]
```
LikeLakers2, line 142: I want to mention: I actually spent some time seeing if I could rewrite this loop to use an iterator (as the lint suggests to do).

However, I ultimately opted not to do so, as I found that I couldn't easily get `i` to be the same values.

I don't believe it's impossible to change it to use an iterator though - and ideally, a follow-up PR might do so.

Bleachfuel: maybe 

```rust
let iter  = lut_begin.ceil() as usize..=lut_end.floor() as usize

for i in iter.iter()
```
seems a bit unnessecary tbh.

crates/bevy_core_pipeline/src/core_2d/camera_2d.rs
```diff
@@ -1,4 +1,7 @@
-#![expect(deprecated)]
+#![expect(
```
Bleachfuel, line 1: Does this have to be here? As you already have #[expect] on the individual impl?

Bleachfuel (commented):
Those Impls seem double to me.

crates/bevy_core_pipeline/src/core_2d/camera_2d.rs
LikeLakers2: Unfortunately, yes. I didn't copy the comment because I knew this PR would be blocked on https://github.com/bevyengine/bevy/pull/16338, but this is actually required.

What's happening is that any `impl` blocks for a deprecated struct/enum are being marked as a use of that deprecated item. This includes `impl` blocks generated by derive macros - in this case, a `derive(Bundle)` is causing the "use of deprecated item" warning.

Unfortunately, short of modifying the derive code, I can't silence those warnings without doing a module-wide silence.

Bleachfuel: Ah okay!

crates/bevy_core_pipeline/src/auto_exposure/compensation_curve.rs
LikeLakers2: For what it's worth, that *would* silence the lint - but I suspect that clippy may eventually start linting on that case too (where the iterator is itself a range).

LikeLakers2:
This PR is blocked on https://github.com/bevyengine/bevy/pull/16338 per Benjamin's request. Once that is merged, I will fix any conflicts and mark it as ready to merge.

LikeLakers2:
No longer blocked by https://github.com/bevyengine/bevy/pull/16338 but I am personally blocking this on https://github.com/bevyengine/bevy/pull/17185

BenjaminBrienen (approved):

IQuick143 (approved):

BenjaminBrienen (approved):

BenjaminBrienen (approved):

PR #17140, opened 2025-01-04
Relax documented safety requirements for `*_assume_mut`

vil-mo:
# Objective

Often you want to lock end users from mutating component, while still having the ability to mutate it in the library internals. For example there is a `Child` and `Parent` relationship in the `bevy_hierarchy` https://github.com/bevyengine/bevy/pull/16662

For that to be possible, the safety requirements of `*_assume_mutable` methods should be relaxed, since they are currently too restrictive for no reason, and do not allow mutation of an immutable component under any circumstances.

## Solution

Current safety requirements:
```
- T must be a mutable component
```
Replaced with:
```
- One of the following should be true:
   - Either `T` is mutable, or
   - `OnReplace` hooks and observers for that component on that entity should trigger immediately before the mutation
     and `OnInsert` should trigger immediately after the mutation, or
   - The user should uphold documented invariants of `T`. If no such documentation provided, it is impossible for the end user to uphold this.
```

# Note

[The pr linked above already breaks current safety requirements](https://github.com/bevyengine/bevy/pull/16662#discussion_r1872019871), and the [suggested alternative](https://github.com/bevyengine/bevy/pull/16662#discussion_r1872303035) also breaks it.

### DIFF
--- a/crates/bevy_ecs/src/world/entity_ref.rs
+++ b/crates/bevy_ecs/src/world/entity_ref.rs
@@ -581,7 +581,11 @@ impl<'w> EntityMut<'w> {
     ///
     /// # Safety
     ///
-    /// - `T` must be a mutable component
+    /// - One of the following should be true:
+    ///   - Either `T` is mutable, or
+    ///   - `OnReplace` hooks and observers for that component on that entity should trigger immediately before the mutation
+    ///     and `OnInsert` should trigger immediately after the mutation, or
+    ///   - The user should uphold documented invariants of `T`. If no such documentation provided, it is impossible for the end user to uphold this.
     #[inline]
     pub unsafe fn get_mut_assume_mutable<T: Component>(&mut self) -> Option<Mut<'_, T>> {
         // SAFETY: &mut self implies exclusive access for duration of returned value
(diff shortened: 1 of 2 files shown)


BenjaminBrienen (approved):

crates/bevy_ecs/src/world/entity_ref.rs
```diff
@@ -581,7 +581,11 @@ impl<'w> EntityMut<'w> {
     ///
     /// # Safety
     ///
-    /// - `T` must be a mutable component
+    /// - One of the following should be true:
+    ///   - Either `T` is mutable, or
+    ///   - `OnReplace` hooks and observers for that component on that entity should trigger immediately before the mutation
+    ///     and `OnInsert` should trigger immediately after the mutation, or
+    ///   - The user should uphold documented invariants of `T`. If no such documentation provided, it is impossible for the end user to uphold this.
```
chescock, line 588: Is this requirement necessary?  If it is, then I think `fn modify_component()` would need to be `unsafe` and include a similar requirement.  

But it seems like this could be handled by privacy: Anything that would violate the invariants of `T` just shouldn't be `pub`.  That just leaves things like `mem::swap`, but as long as you're firing triggers that should be indistinguishable from overwriting the component with `insert`.

chescock (commented):
Loosening the requirement like this seems like a good idea, because people will *expect* this to work, as shown in the `Parent`/`Child` PR.  

(If we don't, then we might want to `debug_assert!` that the component is *actually* mutable to help catch misuses.)  

On the other hand, it seems hard to have a case where firing the triggers manually is simpler than calling `modify_component()`.  When would this method actually be called on an immutable component?  

Allowing this will prohibit at least one possible optimization:  With the current rules, the `changed_tick` is always equal to the `added_tick` for an immutable component, and it would be valid to store only a single copy, with the column storing two pointers to the same allocation.  It would then be legal to create a `Ref`, but UB to create a `Mut`, since the ticks pointers would alias.  (That optimization might be a bad idea anyway!  I just want to illustrate that changing the safety requirement here isn't completely free.)

ItsDoot (commented):
> On the other hand, it seems hard to have a case where firing the triggers manually is simpler than calling `modify_component()`.

Agreed, an example of this would be helpful.

PR #17143, opened 2025-01-04
Simplified UI tree navigation without `ghost_nodes`

ickshonpe:
# Objective

There is a large performance regression in the UI systems in 0.15 because the `UiChildren` and `UiRootRootNodes` system params (even with `ghost_nodes` disabled) are really inefficient compared to regular queries and can trigger a heap allocation with large numbers of children.

## Solution

Replace the `UiChildren` and `UiRootRootNodes` system params with simplified versions when the `ghost_nodes` feature is disabled.

## Testing

yellow this PR, red main

cargo run --example many_buttons --features "trace_tracy" --release

`ui_stack_system`
stack

`ui_layout_system`
unghosted

`update_clipping_system`
clipping


BenjaminBrienen (approved):

alice-i-cecile (approved):

alice-i-cecile:
FYI @cart @viridia :)

viridia:
Yeah I noticed this but didn't comment earlier.

Here's my take: right now ghost nodes are in this kind of quantum superposition state where we haven't committed to keeping them or not. There are three possible outcomes as I see it:

1) Ghost nodes are rejected, in which case there may not be a need for UiChildren at all, we can just go back to doing things the old way. However, this also means that the reactive strategies developed in bevy_reactor and thorium_ui are dead ends, and we'll need to come up with some other strategy for handling dynamic-conditional and dynamic-iterative children. I've written about this too much already, so I'm not going to elaborate further here.

2) Ghost nodes get used by BSN for reactivity, which pretty much requires that they be enabled always, in which case the optimizations here will be irrelevant.

3) Child nodes get reimplemented as relations, in which case all bets are off; any optimizations we do now may or may not be useful in that new world.

The only world in which this PR is really helpful is one in which ghost nodes continue to be supported as an optional feature, and honestly I hope that won't be the case, that we make up our minds one way or another. I'm taking a "wait and see" attitude for now.

That being said, the instant that cart's new relations spawning API drops, I'll be providing an update to thorium_ui which uses the new API and which offers a new and better developer experience, and this will be using ghost nodes. I've been deliberately holding off on promoting any of my ideas until this happens.

alice-i-cecile:
Yep, I agree with that analysis :) I definitely like this as a "merge for now" solution. I suspect we can optimize ghost node performance quite a bit in the future, one way or another, assuming we decide to keep them around. This PR makes sure we have a good target for that work :)

ickshonpe:
I'm not adding any complex optimisations here, I'm just replacing the ghost nodes system param stuff with regular queries for the `Children` and `Parent` components when the feature is disabled. 

I mostly agree otherwise ghost nodes are needed and shouldn't be optional. The current implementation is really bad though, it doesn't need to be this expensive or allocate and the feature-gate is broken. It there is a 0.15.2 release maybe this could included, users shouldn't be paying the cost for a disabled ghost nodes feature.

UkoeHB:
Btw the ghost path can be optimized by using the design of `TextSpanIter`

PR #17149, opened 2025-01-05
Cleanup entity reference types

ItsDoot:
# Objective

Cleanup `EntityRef`, `EntityMut`, and `EntityWorldMut` in preparation for my "Scoped Entity References" PR.

## Solution

- Switched `EntityRef`/`EntityMut` from tuple structs to normal ones.
- Ensured all conversion trait impls use the same `entity` argument name.
- Replaced some `unsafe` with delegated calls from `EntityMut` to `EntityRef`
    - Added `EntityMut::into_readonly` to make the replacements clearer
- Replaced some `unsafe` with delegated calls from `EntityWorldMut` to `EntityMut` and `EntityRef`
    - Added `EntityWorldMut::into_readonly`, `::as_readonly`, `::into_mutable`, `::as_mutable` to make the replacements clearer

## Testing

Reusing current tests.

### DIFF
--- a/crates/bevy_ecs/src/world/entity_ref.rs
+++ b/crates/bevy_ecs/src/world/entity_ref.rs
@@ -306,7 +306,7 @@ impl<'a> From<&'a EntityWorldMut<'_>> for EntityRef<'a> {
     fn from(entity: &'a EntityWorldMut<'_>) -> Self {
         // SAFETY:
         // - `EntityWorldMut` guarantees exclusive access to the entire world.
-        // - `&value` ensures no mutable accesses are active.
+        // - `&entity` ensures no mutable accesses are active.
         unsafe { EntityRef::new(entity.as_unsafe_entity_cell_readonly()) }
     }
 }
@@ -323,7 +323,7 @@ impl<'a> From<&'a EntityMut<'_>> for EntityRef<'a> {
     fn from(entity: &'a EntityMut<'_>) -> Self {
         // SAFETY:
         // - `EntityMut` guarantees exclusive access to all of the entity's components.
-        // - `&value` ensures there are no mutable accesses.
+        // - `&entity` ensures there are no mutable accesses.
         unsafe { EntityRef::new(entity.cell) }
     }
 }
--- a/crates/bevy_ecs/src/world/entity_ref.rs
+++ b/crates/bevy_ecs/src/world/entity_ref.rs
@@ -49,7 +49,9 @@ use super::{unsafe_world_cell::UnsafeEntityCell, Ref, ON_REMOVE, ON_REPLACE};
 /// # bevy_ecs::system::assert_is_system(disjoint_system);
 /// ```
 #[derive(Copy, Clone)]
-pub struct EntityRef<'w>(UnsafeEntityCell<'w>);
+pub struct EntityRef<'w> {
+    cell: UnsafeEntityCell<'w>,
+}
 
 impl<'w> EntityRef<'w> {
     /// # Safety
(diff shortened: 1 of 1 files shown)


crates/bevy_ecs/src/world/entity_ref.rs
```diff
@@ -274,105 +276,106 @@ impl<'w> EntityRef<'w> {
...
+        unsafe { EntityRef::new(entity.cell) }
     }
 }
 
 impl<'a> From<&'a EntityMut<'_>> for EntityRef<'a> {
-    fn from(value: &'a EntityMut<'_>) -> Self {
+    fn from(entity: &'a EntityMut<'_>) -> Self {
         // SAFETY:
         // - `EntityMut` guarantees exclusive access to all of the entity's components.
         // - `&value` ensures there are no mutable accesses.
```
TimJentzsch, line 326: The variable name should be updated here as well

crates/bevy_ecs/src/world/entity_ref.rs
```diff
@@ -1039,6 +1056,28 @@ impl<'w> EntityWorldMut<'w> {
...
+    pub fn into_readonly(self) -> EntityRef<'w> {
+        EntityRef::from(self)
+    }
+
+    /// Gets read-only access to all of the entity's components.
+    pub fn as_readonly(&self) -> EntityRef<'_> {
+        EntityRef::from(self)
+    }
+
+    /// Consumes `self` and returns non-structural mutable access to all of the
```
TimJentzsch, line 1070: What does "non-structural" mean in this context?

ItsDoot: Structural access (`&mut World` access) means you can insert/remove components, or despawn the entity. Non-structural access allows none of that; instead you have sliced access to the world which means you aren't allowed to perform changes that might affect other entities.

crates/bevy_ecs/src/world/entity_ref.rs
```diff
@@ -594,7 +602,19 @@ impl<'w> EntityMut<'w> {
...
+
+    /// Gets mutable access to the component of type `T` for the current entity.
+    /// Returns `None` if the entity does not have a component of type `T`.
+    ///
+    /// # Safety
+    ///
+    /// - `T` must be a mutable component
+    #[inline]
+    pub unsafe fn into_mut_assume_mutable<T: Component>(self) -> Option<Mut<'w, T>> {
+        // SAFETY: consuming `self` implies exclusive access
```
chescock, line 616: This should probably mention that the "the component `T` is mutable" requirement is satisfied by the caller, although I see this was copied from `fn get_mut_assume_mutable()` which is also missing that part of the comment.

chescock (approved):
This looks nice!  This would all be good cleanup even if it weren't necessary for the `EntityPtr` work.  I especially like the reduction of `unsafe` by doing `self.as_mutable().thing()` and `self.as_readonly().thing()`.

crates/bevy_ecs/src/world/entity_ref.rs
ItsDoot: Added!

JaySpruce (approved):

BenjaminBrienen (approved):

crates/bevy_ecs/src/world/entity_ref.rs
TimJentzsch: I see, thanks for the clarification!

TimJentzsch (approved):

hymm (approved):

PR #17081, opened 2025-01-01
Add `-D warnings` in a few steps

BenjaminBrienen:
# Objective

I missed a couple checks in #17011 

## Solution

Add env

## Testing

CI

PR #17156, opened 2025-01-05
msrv needs build

BenjaminBrienen:
# Objective

My previous PR is causing the github bot to comment whenver a PR build fails

## Solution

msrv should depend on build succeeding

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -397,6 +397,7 @@ jobs:
   msrv:
     runs-on: ubuntu-latest
     timeout-minutes: 30
+    needs: build
     steps:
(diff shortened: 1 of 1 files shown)


alice-i-cecile (approved):

mockersf:
could you re-add the other dependencies you removed in #17081?

BenjaminBrienen:
What do the wasm builds need it for?

mockersf:
it's to reduce the amount of work that is not useful, when we already know something will fail

pcwalton:
I don't really know anything about the CI unfortunately.

mockersf (approved):

Issue #16877, opened 2024-12-18
Reorganise Platform Support Dependencies


bushrat011899:
> I'm not a fan of how we bring this `Instant` in. I think it'd be worth considering some kind of `bevy_platform_support` crate which moves things like `spin`, `portable-atomic`, `critical-section`, `wasm-time`, etc. into a single spot. It'd be more focussed than `bevy_utils`, and possibly clean up some of the `cfg(...)` noise we have across Bevy.

_Originally posted by @bushrat011899 in https://github.com/bevyengine/bevy/pull/16874#discussion_r1889405000_

alice-i-cecile:
Part of #11478.

bushrat011899:
Closing as completed by the various contributing PRs above!

PR #17158, opened 2025-01-05
Remove everything except `Instant` from `bevy_utils::time`

bushrat011899:
# Objective

- Contributes to #11478
- Contributes to #16877

## Solution

- Removed everything except `Instant` from `bevy_utils::time`

## Testing

- CI

---

## Migration Guide

If you relied on any of the following from `bevy_utils::time`:

- `Duration`
- `TryFromFloatSecsError`

Import these directly from `core::time` regardless of platform target (WASM, mobile, etc.)

If you relied on any of the following from `bevy_utils::time`:

- `SystemTime`
- `SystemTimeError`

Instead import these directly from either `std::time` or `web_time` as appropriate for your target platform.

## Notes

`Duration` and `TryFromFloatSecsError` are both re-exports from `core::time` regardless of whether they are used from `web_time` or `std::time`, so there is no value gained from re-exporting them from `bevy_utils::time` as well. As for `SystemTime` and `SystemTimeError`, no Bevy internal crates or examples rely on these types. Since Bevy doesn't have a `Time<Wall>` resource for interacting with wall-time (and likely shouldn't need one), I think removing these from `bevy_utils` entirely and waiting for a use-case to justify inclusion is a reasonable path forward.

### DIFF
--- a/crates/bevy_animation/src/transition.rs
+++ b/crates/bevy_animation/src/transition.rs
@@ -10,7 +10,7 @@ use bevy_ecs::{
 };
 use bevy_reflect::{std_traits::ReflectDefault, Reflect};
 use bevy_time::Time;
-use bevy_utils::Duration;
+use core::time::Duration;
 
 use crate::{graph::AnimationNodeIndex, ActiveAnimation, AnimationPlayer};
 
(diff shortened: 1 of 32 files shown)


BenjaminBrienen (approved):

alice-i-cecile (approved):
I agree with this. I was skeptical at first, because the split web_time imports suck, but you're right about us not using them and not having an immediate use for it. As I'm sure you know, moving this into `bevy_platform_compat` is the next step here.

PR #17161, opened 2025-01-05
Remove `tracing` re-export from `bevy_utils`

bushrat011899:
# Objective

- Contributes to #11478

## Solution

- Made `bevy_utils::tracing` `doc(hidden)`
- Re-exported `tracing` from `bevy_log` for end-users
- Added `tracing` directly to crates that need it.

## Testing

- CI

---

## Migration Guide

If you were importing `tracing` via `bevy::utils::tracing`, instead use `bevy::log::tracing`. Note that many items within `tracing` are also directly re-exported from `bevy::log` as well, so you may only need `bevy::log` for the most common items  (e.g., `warn!`, `trace!`, etc.). This also applies to the `log_once!` family of macros.

## Notes

- While this doesn't reduce the line-count in `bevy_utils`, it further decouples the internal crates from `bevy_utils`, making its eventual removal more feasible in the future.
- I have just imported `tracing` as we do for all dependencies. However, a workspace dependency may be more appropriate for version management.


crates/bevy_utils/src/lib.rs
```diff
@@ -63,6 +63,7 @@ pub use hashbrown;
 pub use parallel_queue::*;
 pub use time::*;
 #[cfg(feature = "tracing")]
+#[doc(hidden)]
 pub use tracing;
```
BenjaminBrienen, line 67: Why not delete this?

BenjaminBrienen (approved):

crates/bevy_utils/src/lib.rs
bushrat011899: It's still being used by some of the other exported macros (`warm_once!`, etc.)

alice-i-cecile:
> I have just imported tracing as we do for all dependencies. However, a workspace dependency may be more appropriate for version management.

As @mockersf pointed out to me, the challenge with workspace dependencies is that the features must be shared across all users of the crate. That might be appropriate for tracing, but it's something we should consider (and write down).

crates/bevy_utils/src/lib.rs
alice-i-cecile: I'd prefer to fix that now, and fully remove this.

bushrat011899: I've entirely removed `tracing` from `bevy_utils`. The `X_once!` macros that remained have been moved into `bevy_log` for end users, and simply expanded for internal use to avoid adding a `bevy_log` dependency to too many crates. All it does is replace `warn_once!(...)` with `once!(warn!(...))`, which isn't too bad IMO.

BenjaminBrienen (approved):

cBournhonesque (approved):

PR #17166, opened 2025-01-05
[`InstanceInputUniformBuffer::remove`] could cause incorrect behaviour.

Bleachfuel:
# Objective

[`InstanceInputUniformBuffer::remove`] could, when the wrong index was inserted overwrite data when adding it to the buffer.

## Solution

make [`InstanceInputUniformBuffer::remove`] do some checks before removing. And having[`InstanceInputUniformBuffer::remove_unchecked`] for when you don't need that behaviour.

## Testing

Added unit test

### DIFF
--- a/crates/bevy_render/src/batching/gpu_preprocessing.rs
+++ b/crates/bevy_render/src/batching/gpu_preprocessing.rs
@@ -231,7 +231,7 @@ where
 
     /// Removes a piece of buffered data from the uniform buffer.
     ///
-    /// Ensure uniform_index is within bounds of [`Self::buffer`] and not removed twice,
+    /// Ensure `uniform_index` is within bounds of [`Self::buffer`] and not removed twice,
     /// as this may cause incorrect behavior or panicking when inserting new data.
     pub fn remove_unchecked(&mut self, uniform_index: u32) {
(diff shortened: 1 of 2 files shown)


crates/bevy_render/src/batching/gpu_preprocessing.rs
```diff
@@ -211,16 +210,51 @@ where
...
         self.free_uniform_indices.push(uniform_index);
     }
 
     /// Returns the piece of buffered data at the given index.
-    pub fn get(&self, uniform_index: u32) -> BDI {
+    ///
+    /// Returns [`None`] if the index is out of bounds or the data is removed.
+    pub fn get(&self, uniform_index: u32) -> Option<BDI> {
+        if (uniform_index as usize) >= self.buffer.len()
+            || self.free_uniform_indices.contain(uniform_index)
```
pcwalton, line 237: Pretty sure this should be `contains`

pcwalton:
Does this actually fix a bug? We should never be removing twice to begin with.

Bleachfuel:
Not really, more so a unsoundess possibility. We could also just add notes to the function. This is more if u accidentally remove twice.

BenjaminBrienen (approved):

PR #17169, opened 2025-01-05
Run example validation jobs on `ubuntu-latest`

### LINKED ISSUE #17164: example running on runner ubuntu-latest fails because of system dependency change
When installing dependencies on ubuntu-latest, it fails with:
```
E: Unable to locate package libegl1-mesa
```

this package has changed for ubuntu 24.04

workaround for now is to keep it running on ubuntu-22.04, but it's not viable long term

rparrett:
# Objective

Fixes #17164

## Solution

- Switch to `ubuntu-latest`
- Remove `libegl1-mesa`. It seems that whatever this was providing is now covered by other deps or their deps?

## Testing

Tested [over here](https://github.com/rparrett/bevy/pull/21) by commenting out workflow `if` condition.

### DIFF
--- a/.github/start-wasm-example/package.json
+++ b/.github/start-wasm-example/package.json
@@ -8,7 +8,7 @@
   "author": "",
   "license": "ISC",
   "devDependencies": {
-    "@playwright/test": "^1.28.1"
+    "@playwright/test": "^1.49.1"
   },
   "dependencies": {
     "dotenv": "^16.0.1"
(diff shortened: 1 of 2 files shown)


TimJentzsch (approved):

BenjaminBrienen (approved):

mockersf:
oh nice, didn't try to just remove it to check if it works 😄 

it should already mostly work, but it's blocked until January 17th as GitHub is rolling out the migration of ubuntu-latest to ubuntu-24.04: https://github.com/actions/runner-images/issues/10636

mockersf (approved):

rparrett:
It looks like playwright also needs to be updated. See https://github.com/microsoft/playwright/issues/30368

I previously only tested the vulkan job with an explicit `runs-on: ubuntu-24.04`, but tested again [over here](https://github.com/rparrett/bevy/pull/22).

mockersf (approved):

mockersf:
should be good now 🎉

PR #17173, opened 2025-01-05
Fix confusing variable name

rparrett:
# Objective

Make a confusing variable name less confusing.

`x` is presumably short for `xform` which is short for `transform`. But it just reads like "x", which is confusing when we're sorting by `z`. And in this case, `x` isn't the transform, it's a tuple with the transform in the middle somewhere.

`x.2.translation().z` -> `transform.translation().z`

## Solution

Destructure the tuple and name the thing what it is


rparrett:
Opening an alternative PR that uses radsort here.

Issue #16469, opened 2024-11-22
Disabling bevy_*_picking_backend does not disable backends.


andriyDev:
In my project I want to disable the `bevy_sprite_picking_backend` feature and write my own (I want to fork #16388). However adding the following to my Cargo.toml does not work:

```toml
bevy = { version = "0.15.0-rc.3", default-features = false, features = [
  # Default features
  "android-game-activity",
  "android-game-activity",
  "android_shared_stdcxx",
  "animation",
  "bevy_asset",
  "bevy_audio",
  "bevy_color",
  "bevy_core_pipeline",
  "bevy_gilrs",
  "bevy_gizmos",
  "bevy_gltf",
  # "bevy_mesh_picking_backend",
  "bevy_pbr",
  "bevy_picking",
  "bevy_remote",
  "bevy_render",
  "bevy_scene",
  "bevy_sprite",
  # "bevy_sprite_picking_backend",
  "bevy_state",
  "bevy_text",
  "bevy_ui",
  # "bevy_ui_picking_backend",
  "bevy_winit",
  "custom_cursor",
  "default_font",
  "hdr",
  "multi_threaded",
  "png",
  "smaa_luts",
  "sysinfo_plugin",
  "tonemapping_luts",
  "vorbis",
  "webgl2",
  "x11",
] }
```

My PreUpdate graph still includes the sprite and ui backends. The mesh backend isn't present, but that's just because the `MeshPickingPlugin` is intentionally not added by default.

![Image](https://github.com/user-attachments/assets/13fbe464-b99d-4388-ac03-875bb9e504b0)

It looks like the default features for `bevy_sprite` sets `bevy_sprite_picking_backend`. `bevy` then has a feature `bevy_sprite` that enables that crate which by default uses the picking backend. The `bevy_sprite_picking_backend` feature in `bevy` then does nothing since `bevy_sprite` itself sets this feature.

A similar story exists for `bevy_ui_picking_backend` and `bevy_mesh_picking_backend`.

PR #17175, opened 2025-01-05
Allow users to easily use `bevy_sprite` and `bevy_ui` without picking

rparrett:
# Objective

RIght now it's not possible to opt out of sprite or ui picking at the feature level without jumping through some hoops.

If you add the `bevy_sprite` feature, you get `bevy_sprite_picking_backend`.

If you add the `bevy_ui` feature, you get `bevy_ui_picking_backend`

To get `bevy_sprite` without picking, I think you would have to do something like this, which seems **very** annoying.

```toml
[dependencies]
bevy = { version = "0.15", default-features = false, features = [
    # ... omitted
    # "bevy_sprite",
    # "bevy_ui", # this also brings in bevy_sprite
    # "bevy_text", # this also brings in bevy_sprite
    "bevy_render",
    "bevy_core_pipeline",
    "bevy_color",
] }
bevy_internal = { version = "0.15", default-features = false, features = [
    "bevy_sprite",
    "bevy_ui",
    "bevy_text"
] }
```

## Solution

- Remove `bevy_sprite_picking_backend` from the `bevy_sprite` feature.
- Remove `bevy_ui_picking_backend` from the `bevy_ui` feature.

These are still in Bevy's `default-plugins`.

## Testing

I did some basic testing in a minimal project based on the `sprite_picking` example to verify that "picking stuff" didn't get included with `bevy_sprite`.

I would appreciate help testing. I am just unraveling these features and think that this is correct, but I am not 100% sure.

## Migration Guide

`bevy_sprite_picking_backend` is no longer included by default when using the `bevy_sprite` feature. If you are using Bevy without default features and relied on sprite picking, add this feature to your `Cargo.toml`.

`bevy_ui_picking_backend` is no longer included by default when using the `bevy_ui` feature. If you are using Bevy without default features and relied on sprite picking, add this feature to your `Cargo.toml`.

## Additional info

It looks like we attempted to fix this earlier in #16469, but the fix was incomplete?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -213,7 +213,6 @@ bevy_sprite = [
   "bevy_render",
   "bevy_core_pipeline",
   "bevy_color",
-  "bevy_sprite_picking_backend",
 ]
 
 # Provides text functionality
@@ -226,7 +225,6 @@ bevy_ui = [
   "bevy_text",
   "bevy_sprite",
   "bevy_color",
-  "bevy_ui_picking_backend",
 ]
 
 # Windowing layer


rparrett:
@andriyDev could you take a look?

andriyDev (approved):
Ah thank you for finishing this up. I eventually got #16473 merged which fixed the problem in a better way IMO.

alice-i-cecile (approved):
Definitely better :)

PR #17177, opened 2025-01-05
Upstream `DebugPickingPlugin` from `bevy_mod_picking`

skimmedsquare:
# Objective

The debug features (`DebugPickingPlugin`) from `bevy_mod_picking` were not upstreamed with the rest of the core changes, this PR reintroduces it for usage inside `bevy_dev_tools`

## Solution

Vast majority of this code is taken as-is from `bevy_mod_picking` aside from changes to ensure compilation and code style, as such @aevyrie was added as the co-author for this change.

### Main changes
* `multiselection` support - the relevant code was explicitly not included in the process of upstreaming the rest of the package, so it also has been omitted here.
* `bevy_egui` support - the old package had a preference for using `bevy_egui` instead of `bevy_ui` if possible, I couldn't see a way to support this in a core crate, so this has been removed.

Relevant code has been added to the `bevy_dev_tools` crate instead of `bevy_picking` as it is a better fit and requires a dependency on `bevy_ui` for drawing debug elements. 

### Minor changes
* Changed the debug text size from `60` to `12` as the former was so large as to be unreadable in the new example.

## Testing
* `cargo run -p ci`
* Added a new example in `dev_tools/picking_debug` and visually verified the in-window results and the console messages


aevyrie:
Might make more sense to put the example in the picking group. That way people looking for the feature will see it. That, or just enable for all picking examples instead of making a standalone example.

skimmedsquare:
> Might make more sense to put the example in the picking group. That way people looking for the feature will see it. That, or just enable for all picking examples instead of making a standalone example.

Sounds reasonable - I don't have a strong opinion here, so I'll wait on some other reviewers to chime in before doing this.

aevyrie (approved):
The code is a bit janky tbh. I'm not sure if it makes sense to store the data as a component instead of just holding a map in a resource. At times like this I really wish bevy UI was immediate mode.

This is probably fine to upstream as-is, but I'm not super pleased with the code quality, which is my own fault.

@alice-i-cecile this is missing the pointer click and drag overlay that the egui version has, because it's such a pita to just draw shapes on the screen in bevy. We really need screen space gizmos (screen space immediate mode line drawing) to do this kind of thing. This also highlights the usefulness of immediate mode UI when you just need to throw text, boxes, and circles on screen with a system, especially in dev tools. This entire feature could be a single system with imgui.

crates/bevy_dev_tools/src/picking_debug.rs
```diff
@@ -0,0 +1,323 @@
+//! Text and on-screen debugging tools
```
alice-i-cecile, line 1: This needs to be updated to reflect its new context in bevy_dev_tools :)

crates/bevy_dev_tools/src/picking_debug.rs
```diff
@@ -0,0 +1,323 @@
...
+                        },
+                        h.to_owned(),
+                    )
+                })
+                .collect(),
+        };
+    }
+}
+
+/// Enum representing either a named or unnamed entity.
```
alice-i-cecile, line 247: There's a `NameOrEntity` system param that we should be using instead here.

alice-i-cecile (changes requested):
Just a bit of cleanup to do, but this is almost ready.

skimmedsquare:
> Might make more sense to put the example in the picking group. That way people looking for the feature will see it. That, or just enable for all picking examples instead of making a standalone example.

@alice-i-cecile Any thoughts on the suggestion from @aevyrie above? Was waiting on this before fixing the `check-missing-examples-in-docs` failure

alice-i-cecile:
Mild preference to move it into the picking group :)

crates/bevy_dev_tools/src/picking_debug.rs
```diff
@@ -0,0 +1,323 @@
...
+use bevy_picking::backend::HitData;
+use bevy_picking::hover::HoverMap;
+use bevy_picking::pointer::{Location, PointerId, PointerPress};
+use bevy_picking::prelude::*;
+use bevy_picking::{pointer, PickSet};
+use bevy_reflect::prelude::*;
+use bevy_render::prelude::*;
+use bevy_text::prelude::*;
+use bevy_ui::prelude::*;
+use bevy_utils::tracing::{debug, trace};
```
LikeLakers2, line 19: As of https://github.com/bevyengine/bevy/pull/17161, Tracing is no longer exported from `bevy_utils`.
```suggestion
use tracing::{debug, trace};
```

LikeLakers2, line 19: Simply updating the branch doesn't fix the issue here - you have to change this use statement as well.
```suggestion
use tracing::{debug, trace};
```

skimmedsquare: I'm aware. I'm working on the rest of the changes before pushing again

LikeLakers2: My bad, then.

skimmedsquare: All good, no worries 🙂

alice-i-cecile (approved):
That cleaned up nicely. Thanks!

PR #17179, opened 2025-01-05
can hide status bar on iOS

mockersf:
# Objective

- I want to hide the clock and the battery indicator on iOS

## Solution

- Add the platform specific property `prefers_status_bar_hidden` on Window creation, and map it to `with_prefers_status_bar_hidden` in winit.

## Testing

- Tested on iOS


alice-i-cecile (approved):

BenjaminBrienen (approved):
Looks right, haven't tested

eero-lehtinen (approved):
Haven't tested this PR but have used the same winit function in a Bevy app before and it has always worked.

Winit has `set_prefers_status_bar_hidden` so that value could technically also be updated at runtime.

PR #17195, opened 2025-01-06
Remove references to old sample_texture function

alex5nader:
# Objective

Remove some outdated docs from 0.15 that mention a removed function.

## Solution

In `pbr_functions.wgsl`, I think it's fine to just remove the mention. In `meshlet/asset.rs`, I think it would be nice to still have a note on how texture samples should be done. Unfortunately, there isn't a nice abstraction for it any more. Current workaround, for reference:

https://github.com/bevyengine/bevy/blob/b386d08d0f905d07843ddfed936e404845a6a10b/crates/bevy_pbr/src/render/pbr_fragment.wgsl#L184-L208

For now, I've just removed the mention.

### DIFF
--- a/crates/bevy_pbr/src/render/pbr_functions.wgsl
+++ b/crates/bevy_pbr/src/render/pbr_functions.wgsl
@@ -33,9 +33,8 @@
 #endif
 
 
-// Biasing info needed to sample from a texture when calling `sample_texture`.
-// How this is done depends on whether we're rendering meshlets or regular
-// meshes.
+// Biasing info needed to sample from a texture. How this is done depends on
+// whether we're rendering meshlets or regular meshes.
 struct SampleBias {
 #ifdef MESHLET_MESH_MATERIAL_PASS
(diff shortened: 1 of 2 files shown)


crates/bevy_pbr/src/render/pbr_functions.wgsl
```diff
@@ -33,9 +33,8 @@
 #endif
 
 
-// Biasing info needed to sample from a texture when calling `sample_texture`.
-// How this is done depends on whether we're rendering meshlets or regular
-// meshes.
+// Biasing info needed to sample from a texture. How this is done depends on
+// whether we're rendering meshlets or regular meshes.
 struct SampleBias {
```
JMS55, line 38: Not for this PR, but as a side note the derivative stuff for meshlet meshes are not biases. We should probably rename this type to TextureSampleInfo in the future or something.

JMS55 (approved):

PR #17197, opened 2025-01-06
Make EntityMapper fallible

### LINKED ISSUE #17193: Provide a fallible map_entity api

The current MapEntity trait only provides a `map_entity(entity) -> Entity` method.
There are cases where mapping the entity would fail, for example in networking.

Maybe we could provide a second method:
`try_map_entity(entity) -> Result<Entity>`

or just modify the main api to `map_entity(Entity) -> Result<Entity>`.

I think historically the api was not fallible because the only EntityMapper was SceneEntityMapper which is not fallible.
(it either maps the entity or spawns a new one)

Update `map_entity` to return `Result<Entity>`

Unknown

cBournhonesque:
# Objective

- Fixes https://github.com/bevyengine/bevy/issues/17193

## Solution

Update the `EntityMapper` trait so that 
`map_entity(&mut self, entity: Entity)` returns an `Option<Entity>` instead of `Entity` to account for cases where the mapping fails or cannot be done.

## Testing

Updated the unit tests.
Maybe we should be using a Result, but I didn't know what to put in the error type. Maybe `Box<dyn Error>`?

## Migration Guide

- The `map_entity` method of the `EntityMapper` trait now returns an `Option<Entity>` instead of an `Entity` to support cases where the mapping fails. You will probably need to update any custom implementations of the `MapEntities` trait.

### DIFF
--- a/crates/bevy_ecs/src/entity/map_entities.rs
+++ b/crates/bevy_ecs/src/entity/map_entities.rs
@@ -39,8 +39,13 @@ use super::{EntityHashMap, VisitEntitiesMut};
 ///
 /// impl MapEntities for Spring {
 ///     fn map_entities<M: EntityMapper>(&mut self, entity_mapper: &mut M) {
-///         self.a = entity_mapper.map_entity(self.a);
-///         self.b = entity_mapper.map_entity(self.b);
+///
+///         if let Some(mapped_a) = entity_mapper.map_entity(self.a) {
+///             self.a = mapped_a;
+///         }
+///         if let Some(mapped_b) = entity_mapper.map_entity(self.b) {
+///             self.b = mapped_b;
+///         }
 ///     }
 /// }
 /// ```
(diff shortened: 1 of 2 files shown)


Shatur (commented):
After thinking about it a bit more, I think we always should replace the entity with `Entity::PLACEHOLDER` instead of keeping it untouched.
If yes, do we need to return `Option`?

cBournhonesque:
> After thinking about it a bit more, I think we always should replace the entity with `Entity::PLACEHOLDER` instead of keeping it untouched. If yes, do we need to return `Option`?

So you're saying that you update `MapEntities` to return `Entity::PLACEHOLDER` if it fails,
and then you update `EntityMapper` to do something like:
```
    fn map_entities<M: EntityMapper>(&mut self, entity_mapper: &mut M) {
        let mapped = entity_mapper.map_entity(self.entity);
        if mapped != Entity::PLACEHOLDER {
            self.entity = mapped;
        }
    }
```

I think that works but it requires a bit of user-knowledge. It's kind of 'hidden' pattern that the user can remember to use Entity::PLACEHOLDER as an error value in both `MapEntities` and `EntityMapper`. I would probably prefer things to be more explicit.

Shatur:
> So you're saying that you update MapEntities to return Entity::PLACEHOLDER if it fails, and then you update EntityMapper to do something like:

No, I suggesting to always assign the entity even if it's `Entity::PLACEHOLDER`. It just better then to pointing to some random entity.

PR #17206, opened 2025-01-07
Replace unsafe blocks in `World` and `DeferredWorld` with safe equivalents

ItsDoot:
# Objective

Reduce the number of unsafe blocks.

## Solution

Replaced 5 unsafe blocks with safe equivalents.

## Testing

Reusing current tests


alice-i-cecile (approved):

alice-i-cecile:
Love to see it frankly.

cart (approved):

PR #17207, opened 2025-01-07
Use `as _` to avoid namespace pollution

### LINKED ISSUE #13959: Extension traits are re-exported by name
Extension traits can be re-exported without bringing their name into scope using [underscore imports](https://doc.rust-lang.org/reference/items/use-declarations.html#underscore-imports):

```rust
pub mod prelude {
    // This brings the trait AND its name into scope. This is how bevy currently does it.
    //pub use crate::commands::BuildChildrenTransformExt;

    // This brings the trait into scope.
    pub use crate::commands::BuildChildrenTransformExt as _;
}
```

Users can still pull in the name if needed:

```rust
use bevy_transform::prelude::*;
use bevy_transform::commands::BuildChildrenTransformExt;
```

Benefits:

- Clearly mark certain re-exports as extension traits
- Reduce name pollution (not that extension traits have high-value names, but still)

### LINKED ISSUE #16268: Using `trait as _` to avoid namespace pollution
> This is really interesting, I hadn't considered using `trait as _` to avoid name collisions. At a glance this also works in [preludes](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=2d58697888b02fc3ff00a4ecb0234e2f). For a future PR, could be nice to do this for extension style traits (e.g., `AppAssetExt` etc.) so the preludes aren't so heavy with names.

_Originally posted by @bushrat011899 in https://github.com/bevyengine/bevy/pull/16163#discussion_r1821705357_

BenjaminBrienen:
# Objective

Make it clear when a trait is being used for its extension methods, not for itself.
Fixes #16268 
Fixes #13959

## Solution

Use `as _` for obvious cases.

## Testing

Code compiles


BenjaminBrienen:
FYI @benfrankel

alice-i-cecile (approved):
I think I like this style a bit better.

BenjaminBrienen:
Oh, is that not enabled? I'll check that.

crates/bevy_app/src/panic_handler.rs
```diff
@@ -15,7 +15,7 @@ use crate::{App, Plugin};
 /// * Other platforms are currently not setup.
 ///
 /// ```no_run
-/// # use bevy_app::{App, NoopPluginGroup as MinimalPlugins, PluginGroup, PanicHandlerPlugin};
```
hymm, line 18: Does the lint catch these? Otheriwise changing these just feels like it's just extra noise here.

BenjaminBrienen: Apparently there is a lint, yeah. Let me redo this using the lint.

cart (changes requested):
I strongly believe that this is worse. Its harder to type, harder to read, it requires manual action (RA won't do this for us, so every auto-imported RA action needs fixing up), its another thing that Bevy devs need to remember, and its something we'll get punished for if we lint for it (and if we're adopting the pattern globally, we should lint for it).

I think this is especially heinous for non exported `use` statements, as they only have impact on the local file. The cost of preemptively anticipating and protecting against namespace collisions in local files is _so_ much worse than actually dealing with them when they occur (which basically never happens, and is trivially easy to resolve when it does).

There is an argument to be made for doing this in public prelude exports. Its not something I would personally push for, but I at least see the utility, as it has a more global / unconstrained scope.

BenjaminBrienen:
Conflicts with https://rust-lang.github.io/rust-clippy/master/index.html#unsafe_removed_from_name

benfrankel (approved):
I looked over the diff and it looks correct.

PR #17208, opened 2025-01-07
Fix `scale_factor_override` not used in window creation

### LINKED ISSUE #17188: Window size isn't correct until second frame when scale factor is overridden
## Bevy version

main

SystemInfo { os: "macOS 14.5 Sonoma", kernel: "23.5.0", cpu: "Apple M1 Max", core_count: "10", memory: "64.0 GiB" }

High DPI display (scale factor 2)

## What you did

```rust
use bevy::{diagnostic::FrameCount, prelude::*, window::WindowResolution};

fn main() {
    App::new()
        .add_plugins(DefaultPlugins.set(WindowPlugin {
            primary_window: Some(Window {
                resolution: WindowResolution::new(1920.0, 1080.0).with_scale_factor_override(1.0),
                ..default()
            }),
            ..default()
        }))
        .add_systems(Startup, startup)
        .add_systems(Update, print_windows)
        .run();
}

fn startup(windows: Query<&Window>, frame_count: Res<FrameCount>) {
    for window in &windows {
        info!("{} {:?} (startup)", frame_count.0, window.resolution);
    }
}

fn print_windows(windows: Query<&Window>, frame_count: Res<FrameCount>) {
    for window in &windows {
        info!("{} {:?} (update)", frame_count.0, window.resolution);
    }
}
```

## What went wrong

The `WindowResolution` is incorrect during `Startup` and for one run of the `Update` schedule, and then becomes correct on the next frame.

```
2025-01-06T12:55:54.906932Z  INFO bevy_winit::system: Creating new window App (0v1)
2025-01-06T12:55:54.932602Z  INFO sfo: Frame 0: WindowResolution { physical_width: 3840, physical_height: 2160, scale_factor_override: Some(1.0), scale_factor: 2.0 } (startup)
2025-01-06T12:55:54.941606Z  INFO sfo: Frame 0: WindowResolution { physical_width: 3840, physical_height: 2160, scale_factor_override: Some(1.0), scale_factor: 2.0 } (update)
2025-01-06T12:55:55.002012Z  INFO sfo: Frame 1: WindowResolution { physical_width: 1920, physical_height: 1080, scale_factor_override: Some(1.0), scale_factor: 2.0 } (update)
```

## Additional information

Related to #17183

chompaa:
# Objective

Fixes #17188. When `scale_factor_override` is set on `WindowResolution`, the resulting `WindowResolution` is incorrect for the first frame. This is because when windows are first created in `bevy::winit::create_windows`, the resolution is set without respect to `WindowResolution::scale_factor_override`:
```rust
window
    .resolution
    .set_scale_factor_and_apply_to_physical_size(winit_window.scale_factor() as f32);
```
Afterwards, `bevy::winit::changed_windows` updates the resolution at the end of the first frame.

## Solution

Check if `scale_factor_override` is set before using OS resolution settings on window creation.

## Testing

Tested using the code from the linked issue.

### DIFF
--- a/crates/bevy_winit/src/system.rs
+++ b/crates/bevy_winit/src/system.rs
@@ -77,7 +77,17 @@ pub fn create_windows<F: QueryFilter + 'static>(
             window.window_theme = Some(convert_winit_theme(theme));
         }
 
-        if window.resolution.scale_factor_override().is_none() {
+        if let Some(forced_factor) = window.resolution.scale_factor_override() {
+            let size = window.resolution.physical_size();
+            let size = PhysicalSize::new(
+                ((size.x as f32) * forced_factor) as u32,
+                ((size.y as f32) * forced_factor) as u32,
+            );
+            window
+                .resolution
+                .set_physical_resolution(size.width, size.height);
+            let _ = winit_window.request_inner_size(size);
+        } else {
             window
                 .resolution
                 .set_scale_factor_and_apply_to_physical_size(winit_window.scale_factor() as f32);
--- a/crates/bevy_winit/src/system.rs
+++ b/crates/bevy_winit/src/system.rs
@@ -77,9 +77,11 @@ pub fn create_windows<F: QueryFilter + 'static>(
             window.window_theme = Some(convert_winit_theme(theme));
         }
 
-        window
-            .resolution
-            .set_scale_factor_and_apply_to_physical_size(winit_window.scale_factor() as f32);
+        if window.resolution.scale_factor_override().is_none() {
+            window
+                .resolution
+                .set_scale_factor_and_apply_to_physical_size(winit_window.scale_factor() as f32);
+        }
 
         commands.entity(entity).insert(CachedWindow {
             window: window.clone(),


BenjaminBrienen:
Should there be an `else` that sets it using the `scale_factor_override`? I know you mentioned it gets set later in the first frame, but maybe it makes sense to set it here?

chompaa:
> Should there be an `else` that sets it using the `scale_factor_override`? I know you mentioned it gets set later in the first frame, but maybe it makes sense to set it here?

I don't think so. By getting set later, I meant that the initially (incorrect) set `physical_width` and `physical_height` of the `WindowResolution` is corrected. Now it is initially set correctly.

As an aside, I am a bit confused as to why `set_scale_factor_and_apply_to_physical_size` exists. From my understanding: 

$$\text{logical\\_size} = \text{physical\\_size} \times \text{scale\\_factor}$$

Maybe someone else can shed light on this, but from my point of view it should just set `scale_factor` and not the physical size.

BenjaminBrienen (approved):

BenjaminBrienen (approved):
example works

tauanbinato (approved):

SimplyLiz:
Confirming this PR fixes a real bug on iOS. Cherry-picked onto v0.15.3 and tested on iPhone 16 Pro (logical 874×402, scale 3.0, physical 2622×1206) running iOS 18.

**Symptom without the PR:** Bevy's default `WindowResolution` (1280×720) is passed through to winit's `with_inner_size` on iOS, which sizes the `UIView` to 1280×720 logical points — larger than the actual device screen. The Metal surface then renders at the oversized view size and the bottom-right of the framebuffer falls off-screen. Setting `scale_factor_override` on the `Window` descriptor before `app.run()` had no effect because winit's iOS backend was ignoring the override at surface creation.

**With this PR applied:** `scale_factor_override` is honored at window creation and the physical size is scaled correctly, so the Metal surface is sized to the device's actual pixel dimensions.

Note for iOS users arriving here: this PR is necessary but not sufficient on its own for all devices — the default 1280×720 logical size still lands on-screen as an oversized `UIView` unless something also resizes the UIView/UIWindow to the real screen bounds. In our case we also added a runtime system that walks `view.window.windowScene.screen.bounds` (the `UIScreen.mainScreen` API is deprecated in iOS 16+ and returns stale bounds) and resizes both the `UIWindow` and the `UIView` to the screen rect every frame. Those two fixes combined — this PR + runtime UIKit sync — produced a correctly-sized Metal framebuffer.

Would be great to see this merged; it resolves a hard-to-diagnose class of iOS sizing bugs and anyone hitting #17563 is going to need it.

Aceeri (approved):
I think this is worthwhile to do and I don't think this really makes anything worse than the current situation (aside from users already doing this themselves).

PR #17209, opened 2025-01-07
Add Image constructor specialised for rendering to a texture

### LINKED ISSUE #7358: Provide built-in API for generating "render target texture" and/or simplify "Render to Texture" example

A snippet from the [Render to Texture example](https://github.com/bevyengine/bevy/blob/main/examples/3d/render_to_texture.rs) on main:
```rust
fn setup(
    // ...
) {
    let size = Extent3d {
        width: 512,
        height: 512,
        ..default()
    };

    // This is the texture that will be rendered to.
    let mut image = Image {
        texture_descriptor: TextureDescriptor {
            label: None,
            size,
            dimension: TextureDimension::D2,
            format: TextureFormat::Bgra8UnormSrgb,
            mip_level_count: 1,
            sample_count: 1,
            usage: TextureUsages::TEXTURE_BINDING
                | TextureUsages::COPY_DST
                | TextureUsages::RENDER_ATTACHMENT,
        },
        ..default()
    };

    // fill image.data with zeroes
    image.resize(size);
    
    // ...
}
```

There's quite a lot going on in this snippet that, not coming from a rendering background, and just wanting to make pretty pictures, I simply don't understand.

My less-preferred solution is to simplify the example by creating a helper function, that uses the `Default` impl for `Image`, and calling it in the `setup`.
```rust
fn create_render_target_texture(width: u32, height: u32) -> Image {
    let mut image = Image::default();

    // `Image::resize` replaces `Image.texture_descriptor.size` as well as fills `Image.data` with zeroes.
    image.resize(Extent3d {
        width,
        height,
        ..default()
    });

    image.texture_descriptor.usage =
        TextureUsages::TEXTURE_BINDING | TextureUsages::COPY_DST | TextureUsages::RENDER_ATTACHMENT;

    image
}

fn setup(
    // ...
) {
    let image = create_render_target_texture(512, 512);
    // ...
}
```

By using the `Default` impl it also removes 3 `use`s, and perhaps the remaining `use`s within `fn create_render_target_texture` could be put into the body or fully qualified.

My more-preferred solution would be to expose this same helper function (with a better name) as a constructor on `Image`. The example can simply call that constructor.

Leave it as it is, and force future users to dive into (some of) the implementation details to understand what's going on.

Perhaps there is a missing higher level API to achieve the same outcomes?

Being new to rendering in general, there are a large number of properties that I'm not really sure what they do (hence my reliance on the `Default` impl). Maybe some of these it makes sense to expose as parameters for the helper function / constructor.

## Additional context

Initially all I had read was the Bevy 0.8 release notes, and just tried to construct an `Image` and see if I could get it working 
(I thought it couldn't be that hard), then I got a few crashes / things I had to import / all these weird rendering concepts / etc., until I folded and decided to see if there was an example (and of course there was).

Strictly speaking, there are two small differences between my proposed version and the example:

1. The example sets the `size` when initially constructing the `Image`, then `resize`s, but `resize` sets the `size` anyway so we don't need to do this extra step.

2. The example sets `Image.texture_descriptor.format` to `TextureFormat::Bgra8UnormSrgb`, but the default impl sets this to `TextureFormat::Rgba8UnormSrgb` via `wgpu::TextureFormat::bevy_default()`. I don't know what sort of impact this has, but it works on my machine.

tigregalis:
# Objective

Fixes #7358
Redo of #7360

Ergonomics. There's a bunch of enigmatic boilerplate for constructing a texture for rendering to, which could be greatly simplified for the external user-facing API.

## Solution

- Take part of the render_to_target example and turn it into a new constructor for `Image`, with minimal changes beyond the `Default` implementation.
- Update the render_to_target example to use the new API.

Strictly speaking, there are two small differences between the constructor and the example:

1. The example sets the `size` when initially constructing the `Image`, then `resize`s, but `resize` sets the `size` anyway so we don't need to do this extra step.

2. The example sets `Image.texture_descriptor.format` to `TextureFormat::Bgra8UnormSrgb`, but the default impl sets this to `TextureFormat::Rgba8UnormSrgb` via `wgpu::TextureFormat::bevy_default()`. I don't know what sort of impact this has, but it works on my machine.

I've deliberately chosen to only include `width` and `height` as parameters, but maybe it makes sense for some of the other properties to be exposed as parameters.

---

### Added

Added `Image::new_target_texture` constructor for simpler creation of render target textures.

---

Notes:

- This is a re-do of https://github.com/bevyengine/bevy/pull/7360 - there's some relevant discussion on code style there.
- The docs for the method want to refer to `bevy_render::camera::Camera` and bevy_render::camera::RenderTarget::Image`. `bevy_image` used to be part of `bevy_render` and was split out in the past, and `bevy_image` doesn't depend on `bevy_render`. What's the recommendation here?

### DIFF
--- a/crates/bevy_image/src/image.rs
+++ b/crates/bevy_image/src/image.rs
@@ -761,6 +761,26 @@ impl Image {
         value
     }
 
+    /// Create a new zero-filled image with a given size, which can be rendered to.
+    /// This is primarily for use as a render target for a [`Camera`](`bevy_render::camera::Camera`).
+    /// See [`RenderTarget::Image`](`bevy_render::camera::RenderTarget::Image`).
+    pub fn new_target_texture(width: u32, height: u32) -> Self {
+        let mut image = Self::default();
+
+        image.resize(Extent3d {
+            width,
+            height,
+            ..Default::default()
+        });
+
+        // You need to set these texture usage flags in order to use the image as a render target
+        image.texture_descriptor.usage = TextureUsages::TEXTURE_BINDING
+            | TextureUsages::COPY_DST
+            | TextureUsages::RENDER_ATTACHMENT;
+
+        image
+    }
+
     /// Returns the width of a 2D image.
     #[inline]
     pub fn width(&self) -> u32 {
--- a/examples/3d/render_to_texture.rs
+++ b/examples/3d/render_to_texture.rs
@@ -2,14 +2,7 @@
 
 use std::f32::consts::PI;
 
-use bevy::{
-    prelude::*,
-    render::{
-        render_asset::RenderAssetUsages,
-        render_resource::{Extent3d, TextureDimension, TextureFormat, TextureUsages},
-        view::RenderLayers,
-    },
-};
+use bevy::{prelude::*, render::view::RenderLayers};
 
 fn main() {
     App::new()
@@ -33,23 +26,8 @@ fn setup(
     mut materials: ResMut<Assets<StandardMaterial>>,
     mut images: ResMut<Assets<Image>>,
 ) {
-    let size = Extent3d {
-        width: 512,
-        height: 512,
-        ..default()
-    };
-
     // This is the texture that will be rendered to.
-    let mut image = Image::new_fill(
-        size,
-        TextureDimension::D2,
-        &[0, 0, 0, 0],
-        TextureFormat::Bgra8UnormSrgb,
-        RenderAssetUsages::default(),
-    );
-    // You need to set these texture usage flags in order to use the image as a render target
-    image.texture_descriptor.usage =
-        TextureUsages::TEXTURE_BINDING | TextureUsages::COPY_DST | TextureUsages::RENDER_ATTACHMENT;
+    let image = Image::new_target_texture(512, 512);
 
     let image_handle = images.add(image);
 


tigregalis:
@BenjaminBrienen I've "rebased" this

crates/bevy_image/src/image.rs
```diff
@@ -761,6 +761,26 @@ impl Image {
         value
     }
 
+    /// Create a new zero-filled image with a given size, which can be rendered to.
+    /// This is primarily for use as a render target for a [`Camera`](`bevy_render::camera::Camera`).
+    /// See [`RenderTarget::Image`](`bevy_render::camera::RenderTarget::Image`).
+    pub fn new_target_texture(width: u32, height: u32) -> Self {
+        let mut image = Self::default();
+
+        image.resize(Extent3d {
```
chompaa, line 770: I noticed you discussed being more explicit and constructing `Image` directly in your previous PR. Personally, I prefer that approach over using `resize`. Was there a reason this wasn't changed?

crates/bevy_image/src/image.rs
```diff
@@ -761,6 +761,26 @@ impl Image {
         value
     }
 
+    /// Create a new zero-filled image with a given size, which can be rendered to.
+    /// This is primarily for use as a render target for a [`Camera`](`bevy_render::camera::Camera`).
+    /// See [`RenderTarget::Image`](`bevy_render::camera::RenderTarget::Image`).
```
chompaa, line 766: This fixes those pesky doc links :)

```suggestion
  /// Create a new zero-filled image with a given size, which can be rendered to. This is primarily
  /// for use as a render target for a [`Camera`]. See [`RenderTarget::Image`].
  ///
  /// [`Camera`]: https://docs.rs/bevy/latest/bevy/render/camera/struct.Camera.html
  /// [`RenderTarget::Image`]: https://docs.rs/bevy/latest/bevy/render/camera/enum.RenderTarget.html#variant.Image
```

crates/bevy_image/src/image.rs
tigregalis: I've pushed a new commit that inlines the logic for default and resize and tidied it it up - it is definitely clearer overall.

To answer your question, it's mostly because there wasn't really a clear direction given on the style, and the style for the constructors are already inconsistent within this file:

- `Default::default`: struct literal
- `Image::new`: start with `Default::default`
- `Image::transparent`: struct literal
- `Image::new_fill`: start with `Default:;default`
- (this) `Image::new_target_texture` (original): start with `Default:;default`
- (this) `Image::new_target_texture` (new): struct literal

I think part of the awkwardness comes from the fact that `TextureDescriptor` (a foreign type) doesn't have a default, and it's not clear what a sensible default should be. Somewhat related to this, is this comment:
```rs
pub struct Image {
    // ..
    // TODO: this nesting makes accessing Image metadata verbose. Either flatten out descriptor or add accessors
    pub texture_descriptor: TextureDescriptor<Option<&'static str>, &'static [TextureFormat]>,
    // ..
}
```

BenjaminBrienen (approved):

JMS55:
Shouldn't hard code to default the texture format. HDR cameras need a different format.

IceSentry (approved):
Can you also update headless_renderer, gpu_readback, and compute_shader_game_of_life

I think those are all the examples that use this pattern

IceSentry:
> Shouldn't hard code to default the texture format. HDR cameras need a different format.

Yeah, this should either take a `hdr` parameter or just a texture format.

tigregalis:
In the discord there were a few comments:

1. `BevyDefault` which is defined within `bevy_image` and only has a single implementer (`TextureFormat`) should be removed and any calls to `TextureFormat::bevy_default()` should be replaced with (now) `bevy_image::TEXTURE_FORMAT_SDR`.
2. Perhaps this new constructor really wants to live in `bevy_render`, either as a free function or as an extension trait.
3. Perhaps `TEXTURE_FORMAT_SDR` and `TEXTURE_FORMAT_HDR` should really live in `bevy_render`, but the other constructors on `Image` also want to use `TEXTURE_FORMAT_SDR` (as a "default" `TextureFormat`).

For now I have left the constructor in `bevy_image`, added a `hdr: bool` parameter, created the two constants, and have `ViewTarget::TEXTURE_FORMAT_HDR` point to `bevy_image::TEXTURE_FORMAT_HDR`.

BenjaminBrienen (approved):

tigregalis:
> Can you also update headless_renderer, gpu_readback, and compute_shader_game_of_life
> 
> I think those are all the examples that use this pattern.
> 
> (Sorry, that wasn't an approve, just a comment)

I've done this but the discrepancies make me think maybe these should have their own helper constructors, possibly with guidance in the doc comment.

- `headless_renderer`: additional texture usage `COPY_SRC`
- `gpu_readback`: additional texture usage `COPY_SRC` + `STORAGE_BINDING`, different texture format `R32Uint`, different asset usage `RENDER_WORLD`
- `compute_shader_game_of_life`: different texture usage `COPY_DST` + `STORAGE_BINDING` + `TEXTURE_BINDING`, different texture format `R32Float`, different asset usage `RENDER_WORLD`

IceSentry (changes requested):
Hey, sorry for forgetting about this PR. After looking at the change again, I think I'm almost happy with everything, but I realize now that maybe we should just have the texture format be a direct parameter to the function instead of an HDR flag. That way it can easily be used in other contexts and people can inline the `if hdr` pattern if they need it and if they don't it's pretty easy to just provide the right texture format.

I think it's fine if more advanced use cases just modify the resulting image like you did in the examples. It's still nicer to have the constructor as a starting point for that boilerplate.

tigregalis:
@IceSentry 

I've updated `Image::new_target_texture` to take `TextureFormat` parameter instead of `hdr` flag.

I've improved the documentation.

The examples I've updated are:
- `compute_shader_game_of_life`
- `render_to_texture`
- `headless_renderer`

Since example `gpu_readback` now uses the `Image::new_uninit` constructor, I've opted to revert that change and leave that is is; `Image::new_target_texture` felt like a misuse for it anyway.

IceSentry:
This looks good, but could you make the changes to add the TEXTURE_FORMAT consts to bevy_image in a separate PR? I think it's a worthwhile change, but I'd prefer if it wasn't bundled with a feature PR.

crates/bevy_image/src/image.rs
```diff
@@ -817,6 +820,51 @@ impl Image {
...
+    /// for Standard Dynamic Range (SDR) and High Dynamic Range (HDR) respectively.
+    ///
+    /// The default texture usages are
+    /// [`TEXTURE_BINDING`](TextureUsages::TEXTURE_BINDING),
+    /// [`COPY_DST`](TextureUsages::COPY_DST),
+    /// [`RENDER_ATTACHMENT`](TextureUsages::RENDER_ATTACHMENT).
+    ///
+    /// [`Camera`]: https://docs.rs/bevy/latest/bevy/render/camera/struct.Camera.html
+    /// [`RenderTarget::Image`]: https://docs.rs/bevy/latest/bevy/render/camera/enum.RenderTarget.html#variant.Image
+    pub fn new_target_texture(width: u32, height: u32, format: TextureFormat) -> Self {
```
IceSentry, line 837: The doc should probably add a note about the `RenderAssetUsages` being default. For some edge cases people might prefer using RENDER_WORLD instead of default. I think for simplicity's sake it's fine to not make it an argument, but documenting it so advanced users know they'll need to change it would be a good thing.

tigregalis: Are you sure? In the primary `render_to_texture` example, which doesn't seem to be accessing the image from the CPU after creation:

With `RenderAssetUsages::default()` (`RenderAssetUsages::MAIN_WORLD | RenderAssetUsages::RENDER_WORLD`):

image

But with only `RenderAssetUsages::RENDER_WORLD` the first pass image isn't being lit properly?

image

tigregalis: Could it be a lighting issue when something is `RENDER_WORLD` only? A similar problem occurs with example `headless_renderer`. Below, `cpu_image` is set to both `MAIN_WORLD | RENDER_WORLD`.

```rs
    // This is the texture that will be rendered to.
    let mut render_target_image =
        Image::new_target_texture(size.width, size.height, TextureFormat::bevy_default());
    render_target_image.texture_descriptor.usage |= TextureUsages::COPY_SRC;
    // render_target_image.asset_usage = Default::default(); // necessary so that the scene is lit
    let render_target_image_handle = images.add(render_target_image);

    // This is the texture that will be copied to.
    let mut cpu_image =
        Image::new_target_texture(size.width, size.height, TextureFormat::bevy_default());
    cpu_image.asset_usage = Default::default(); // necessary because it will crash otherwise
    let cpu_image_handle = images.add(cpu_image);
```

`render_target_image.asset_usage = RENDER_WORLD`:


`render_target_image.asset_usage = MAIN_WORLD | RENDER_WORLD`:
old-000

IceSentry: Uh, that seems like a bug to me, I don't see why this would happen because of the usages. I guess let's ignore it for this PR for now.

IceSentry (approved):

IceSentry:
Approved, but I would still like the texture format changes in bevy_image to be in a separate PR

crates/bevy_image/src/image.rs
tigregalis: Alright, for this PR I've used `RenderAssetUsages::default()` as the default and documented it, since it makes the example code read "as intended". When that (lighting?) bug is fixed we can switch it to `RENDER_WORLD` only as the default and update the docs.

tigregalis:
> Approved, but I would still like the texture format changes in bevy_image to be in a separate PR

Dropped the constants out of this PR.

janhohenheim:
Tagging as `Waiting-on-Author` because this has 2 community approvals, but now has merge conflicts

tigregalis:
@janhohenheim Rebased and passing CI.

janhohenheim:
Cool, tagged it to be ready for the next merge train :)

PR #17211, opened 2025-01-07
Use `multi_draw_indirect_count` where available, in preparation for two-phase occlusion culling.

pcwalton:
This commit allows Bevy to use `multi_draw_indirect_count` for drawing meshes. The `multi_draw_indirect_count` feature works just like `multi_draw_indirect`, but it takes the number of indirect parameters from a GPU buffer rather than specifying it on the CPU.

Currently, the CPU constructs the list of indirect draw parameters with the instance count for each batch set to zero, uploads the resulting buffer to the GPU, and dispatches a compute shader that bumps the instance count for each mesh that survives culling. Unfortunately, this is inefficient when we support `multi_draw_indirect_count`. Draw commands corresponding to meshes for which all instances were culled will remain present in the list when calling
`multi_draw_indirect_count`, causing overhead. Proper use of `multi_draw_indirect_count` requires eliminating these empty draw commands.

To address this inefficiency, this PR makes Bevy fully construct the indirect draw commands on the GPU instead of on the CPU. Instead of writing instance counts to the draw command buffer, the mesh preprocessing shader now writes them to a separate *indirect metadata buffer*. A second compute dispatch known as the *build indirect parameters* shader runs after mesh preprocessing and converts the indirect draw metadata into actual indirect draw commands for the GPU. The build indirect parameters shader operates on a batch at a time, rather than an instance at a time, and as such each thread writes only 0 or 1 indirect draw parameters, simplifying the current logic in `mesh_preprocessing`, which currently has to have special cases for the first mesh in each batch. The build indirect parameters shader emits draw commands in a tightly packed manner, enabling maximally efficient use of `multi_draw_indirect_count`.

Along the way, this patch switches mesh preprocessing to dispatch one compute invocation per render phase per view, instead of dispatching one compute invocation per view. This is preparation for two-phase occlusion culling, in which we will have two mesh preprocessing stages. In that scenario, the first mesh preprocessing stage must only process opaque and alpha tested objects, so the work items must be separated into those that are opaque or alpha tested and those that aren't. Thus this PR splits out the work items into a separate buffer for each phase. As this patch rewrites so much of the mesh preprocessing infrastructure, it was simpler to just fold the change into this patch instead of deferring it to the forthcoming occlusion culling PR.

Finally, this patch changes mesh preprocessing so that it runs separately for indexed and non-indexed meshes. This is because draw commands for indexed and non-indexed meshes have different sizes and layouts. *The existing code is actually broken for non-indexed meshes*, as it attempts to overlay the indirect parameters for non-indexed meshes on top of those for indexed meshes. Consequently, right now the parameters will be read incorrectly when multiple non-indexed meshes are multi-drawn together. *This is a bug fix* and, as with the change to dispatch phases separately noted above, was easiest to include in this patch as opposed to separately.

## Migration Guide

* Systems that add custom phase items now need to populate the indirect drawing-related buffers. See the `specialized_mesh_pipeline` example for an example of how this is done.


crates/bevy_pbr/src/render/mesh_preprocess_types.wgsl
```diff
@@ -2,18 +2,96 @@
...
-    // `base_vertex` or `first_instance`.
-    base_vertex_or_first_instance: u32,
-    // A read-only copy of `instance_index`.
+// Per-frame data that the CPU supplies to the GPU.
+struct MeshInput {
+    // The model transform.
+    world_from_local: mat3x4<f32>,
+    // The lightmap UV rect, packed into 64 bits.
+    lightmap_uv_rect: vec2<u32>,
+    // Various flags.
```
atlv24, line 11: I know this comment is just moved from another file, but this comment sucks. What flags are these? You don't have to resolve this, it can be dont in another (potentially someone else's) PR.

atlv24, line 11:
```suggestion
    // Bits from the rust struct `MeshFlags`.
```

crates/bevy_pbr/src/render/mesh.rs
```diff
@@ -907,11 +923,23 @@ impl RenderMeshInstanceGpuBuilder {
...
+                None => (0, 0),
+            };
+        let (mesh_is_indexed, first_index_index, index_count) =
+            match mesh_allocator.mesh_index_slice(&self.shared.mesh_asset_id) {
+                Some(mesh_index_slice) => (
+                    true,
+                    mesh_index_slice.range.start,
+                    mesh_index_slice.range.end - mesh_index_slice.range.start,
+                ),
+                None => (false, !0, !0),
```
atlv24, line 941: why `!0` here but `0` above?

crates/bevy_pbr/src/render/mesh.rs
```diff
@@ -1698,86 +1734,31 @@ impl GetFullBatchData for MeshPipeline {
...
-            }
-        }
-        RenderMeshBufferInfo::NonIndexed => IndirectParameters {
-            vertex_or_index_count: mesh.vertex_count,
+        let indirect_parameters = IndirectParametersMetadata {
+            mesh_index,
+            base_output_index,
+            batch_set_index: match batch_set_index {
+                Some(batch_set_index) => u32::from(batch_set_index),
+                None => !0,
```
atlv24, line 1750: Why `!0`?

crates/bevy_sprite/src/mesh2d/mesh.rs
```diff
@@ -403,56 +403,33 @@ impl GetFullBatchData for Mesh2dPipeline {
...
-                first_vertex_or_first_index: vertex_buffer_slice.range.start,
-                base_vertex_or_first_instance: 0,
-                // Use `0xffffffff` as a placeholder to tell the mesh
-                // preprocessing shader that this is a non-indexed mesh.
-                first_instance: !0,
+        let indirect_parameters = IndirectParametersMetadata {
+            mesh_index: input_index,
+            base_output_index,
+            batch_set_index: match batch_set_index {
+                None => !0,
```
atlv24, line 421: Why `!0`?

crates/bevy_pbr/src/render/build_indirect_params.wgsl
```diff
@@ -0,0 +1,106 @@
...
+    // dynamically reserve an index for the indirect parameters we're to
+    // generate.
+#ifdef MULTI_DRAW_INDIRECT_COUNT_SUPPORTED
+    if (instance_count == 0u) {
+        return;
+    }
+
+    // If this batch belongs to a batch set, then allocate space for the
+    // indirect commands in that batch set.
+    if (batch_set_index != 0xffffffffu) {
```
atlv24, line 79: I guess this is one of the `!0`s

atlv24 (approved):
this is good stuff! nice!

crates/bevy_pbr/src/render/mesh.rs
pcwalton: Line 79 of `build_indirect_params.wgsl` checks this value to see whether the batch belongs to a batch set.

crates/bevy_pbr/src/render/build_indirect_params.wgsl
pcwalton: Yep, it's the `batch_set_index` value.

crates/bevy_sprite/src/mesh2d/mesh.rs
pcwalton: Line 79 of build_indirect_params.wgsl checks this value to see whether the batch belongs to a batch set.

crates/bevy_pbr/src/render/mesh.rs
pcwalton: I switched it to `0, 0`. In an earlier version of this patch, I used an index buffer range starting with `!0` to indicate that a mesh was non-indexed, but now there's no need to do that as indexed and non-indexed meshes are kept fully separated throughout the pipeline.

crates/bevy_pbr/src/render/mesh_preprocess_types.wgsl
pcwalton: I updated the comment to mention that these are `MeshFlags`.

IceSentry (approved):

pcwalton:
Many test failures: https://pixel-eagle.com/project/B25A040A-A980-4602-B90C-D480AB84076D/run/6860/compare/6846

pcwalton:
I believe all the regressions are fixed. The ones worth calling out are:

* `wireframe_2d` -- I believe that my PR actually fixes the rendering from `main`.
* `texture_atlas` -- Locally I can't reproduce the bots' rendering on `main`. My "before" and "after" are identical to the bot's "after". So I think it's some kind of pre-existing race.
* `2d_gizmos` -- seems to be a bug fix from `main`, possibly related to the indexed/non-indexed fix.

PR #17215, opened 2025-01-07
Improved Command Errors

cart:
# Objective

Rework / build on #17043 to simplify the implementation. #17043 should be merged first, and the diff from this PR will get much nicer after it is merged (this PR is net negative LOC).

## Solution

1. Command and EntityCommand have been vastly simplified. No more marker components. Just one function.
2. Command and EntityCommand are now generic on the return type. This enables result-less commands to exist, and allows us to statically distinguish between fallible and infallible commands, which allows us to skip the "error handling overhead" for cases that don't need it.
3. There are now only two command queue variants: `queue` and `queue_fallible`. `queue` accepts commands with no return type. `queue_fallible` accepts commands that return a Result (specifically, one that returns an error that can convert to `bevy_ecs::result::Error`).
4. I've added the concept of the "default error handler", which is used by `queue_fallible`. This is a simple direct call to the `panic()` error handler by default. Users that want to override this can enable the `configurable_error_handler` cargo feature, then initialize the GLOBAL_ERROR_HANDLER value on startup. This is behind a flag because there might be minor overhead with `OnceLock` and I'm guessing this will be a niche feature. We can also do perf testing with OnceLock if someone really wants it to be used unconditionally, but I don't personally feel the need to do that.
5. I removed the "temporary error handler" on Commands (and all code associated with it). It added more branching, made Commands bigger / more expensive to initialize (note that we construct it at high frequencies / treat it like a pointer type), made  the code harder to follow, and introduced a bunch of additional functions. We instead rely on the new default error handler used in `queue_fallible` for most things. In the event that a custom handler is required, `handle_error_with` can be used.
6. EntityCommand now _only_ supports functions that take `EntityWorldMut` (and all existing entity commands have been ported). Removing the marker component from EntityCommand hinged on this change, but I strongly believe this is for the best anyway, as this sets the stage for more efficient batched entity commands.
7. I added `EntityWorldMut::resource` and the other variants for more ergonomic resource access on `EntityWorldMut` (removes the need for entity.world_scope, which also incurs entity-lookup overhead).

## Open Questions

1. I believe we could merge `queue` and `queue_fallible` into a single `queue` which accepts both fallible and infallible commands (via the introduction of a `QueueCommand` trait). Is this desirable?


spectria-limina:
I don't have time for a full review, but I skimmed, and two quick comments. 

First, I think the output should be an associated type, not a parameter. Can't think of any good reason to allow a single type to implement the trait multiple times with different return types.

Second, if I read correctly, there isn't a way to make built in commands like `insert` return errors, right? I think that option needs to exist. Apologies if I missed it in the PR.

crates/bevy_ecs/src/system/commands/command.rs
```diff
@@ -45,91 +45,57 @@ use crate::{
...
+            Err(err) => (error_handler)(world, err),
+        }
     }
+}
 
-    fn apply_internal(self, world: &mut World) -> Result<(), CommandError> {
-        self(world)
+/// Takes a [`Command`] that returns a [`Result`] with an error that can be converted into the [`Error`] type
+/// and returns a [`Command`] that internally converts that error to [`Error`] (if it occurs).
+pub fn map_command_err<T, E: Into<Error>>(
```
cBournhonesque, line 93: Would it be worth combining this with the impl HandleError above with

```
impl<C: Command<Result<T, E>>, T, E: Into<Error>> HandleError for C { 
   ...
}
```
Since it's essentially what we want in the end?

I see this function is only used once in queue_fallible to create a Command compatible with `HandleError`

crates/bevy_ecs/src/world/entity_ref.rs
```diff
@@ -1284,6 +1284,45 @@ impl<'w> EntityWorldMut<'w> {
...
 
+    /// Gets a reference to the resource of the given type
+    ///
+    /// # Panics
+    ///
+    /// Panics if the resource does not exist.
+    /// Use [`get_resource`](EntityWorldMut::get_resource) instead if you want to handle this case.
+    #[inline]
+    #[track_caller]
+    pub fn resource<R: Resource>(&self) -> &R {
```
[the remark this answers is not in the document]
cart: I can see the argument, although porting one of the doc examples is what motivated this change. Switching to `EntityCommand` to accept `EntityWorldMut` resulted in increased UX pressure to add these. I personally consider them to be reasonably uncontroversial. But if people feel differently I'm happy to break this out.

cart:
@spectria-limina
> First, I think the output should be an associated type, not a parameter. Can't think of any good reason to allow a single type to implement the trait multiple times with different return types.

An associated type would prevent default values, which would force `Command<Out = ()>` instead of `Command`. Likewise `Command<Out = Result>` is just a bit nastier to write than `Command<Result>`. It also forces implementations that look like this:

```rust
impl Command for Foo {
  type Out = ();
  fn apply(self, world: &mut World) {
  }
}
```

Instead of this:

```rust
impl Command for Foo {
  fn apply(self, world: &mut World) {
  }
}
```

I think normal generics are the move for this one.

> Second, if I read correctly, there isn't a way to make built in commands like insert return errors, right? I think that option needs to exist. Apologies if I missed it in the PR.

First: I'll not that most of the built in commands as phrased in #17043 do not (yet) return errors, and I think updating everything is out of scope for this PR.

That being said, multiple paths have been laid for built-in commands to return handled errors.

The first path: all built-in commands are now exported as functions in the `command` and `entity_command` modules. If any of them were rephrased to return an error, then this would immediately be possible:

```rust
entity.queue_fallible(insert(Team::Blue).handle_error_with(warn()));
```

The more common case of using the built-in shorthand functions would use the default global error handler, which defaults to panicking but can be configured:

```rust
entity.insert(Team::Blue);
```

@cBournhonesque

> I feel like the EntityCommands change to only support EntityWorldMut should be part of a separate PR. It's kind of independent to error-handling. Having this change in the same PR makes it harder to discuss implications specific to the EntityWorldMut change

As mentioned in the description, the simplification of EntityCommands hinges on there being only one blanket function impl. We'd need to remove one of the two existing implementations anyway.

> I think queue and queue_fallible should be unified if possible. I have tons of commands in my project, some of them are tiny and return (), some of them are more complex and are fallible. I don't really want to have to remember if they return () or Result to see if I need to call queue or queue_fallible, I just want to queue them. If they return Result then I want the Result to be handled.

Good points. I'm on "team unify" as well. I'll investigate.

spectria-limina:
Normal generics would allow for both `queue` and `queue_fallible` to be the same function though, wouldn't they? Because the return type would be inferred from the type of the `Command`. So in that case, you wouldn't need to write `Command<Out = ()>` until you got to a point where you really needed that specifically. Code that doesn't care about the output type would never need to mention it.

There'd probably need to be some sort of trait bound on the output type, I guess, to ensure that it can be reduced to a `()`, say by logging and discarding the error or by panicking. But that again wouldn't need to be mentioned in most places.

JaySpruce:
I feel like a `queue_fallible_with` would help users without really costing anything. If someone wants to modify how an entity command reacts to the entity not existing, they have to do everything internal themselves:
```
commands.queue(entity_command::insert(bundle).with_entity(entity).handle_error_with(error_handler::warn());
```

Users don't have any reason to even know `with_entity` exists prior to this.

With `queue_fallible_with`, it's not really shorter, but users only have to worry about stuff they already know:
```
commands.entity(entity).queue_fallible_with(entity_command::insert(bundle), error_handler::warn());
```

NthTensor:
Ok, starting on a review of this now. Having not yet looked at the code, I'm really pleased with the high-level goals. There is just one note, before I forget:

> I've added the concept of the "default error handler", which is used by queue_fallible. This is a simple direct call to the panic() error handler by default. Users that want to override this can enable the configurable_error_handler cargo feature, then initialize the GLOBAL_ERROR_HANDLER value on startup. 

Ok, I like this, I want default handlers. But it's weird that this is not unified with the system result handler, which I placed in the schedule runner here in https://github.com/bevyengine/bevy/pull/16589. Personally I don't really care where the default handler lives, and if we want to do a global-install-style handler that's fine with me, but imo we should pick one or the other.

> This is behind a flag because there might be minor overhead with OnceLock and I'm guessing this will be a niche feature. We can also do perf testing with OnceLock if someone really wants it to be used unconditionally, but I don't personally feel the need to do that.

OnceLock is probably alright here. I had some idea that we could be returning these errors to the scheduler through the sync-point system, idk if that's a viable. I don't really think this will be a niche feature though. Don't we anticipate the editor using custom error handling to populate an error console? But that's nothing to do with the implementation.

cart:
> I feel like a queue_fallible_with (or queue_with) would help users without really costing anything. If someone wants to modify how an entity command reacts to the entity not existing, they have to do everything internal themselves:

The code you provided works, but it also isn't necessary. This is already supported:

```rust
commands.entity(entity).queue_fallible(entity_command::insert(bundle).handle_error_with(error_handler::warn()));
```

Which is roughly as good as `queue_fallible_with`:

```rust
commands.entity(entity).queue_fallible_with(entity_command::insert(bundle), error_handler::warn());
```

I personally like keeping the variants down / having one way to do things. It gets even closer ergonomically if we unify queue and queue_fallible:

```rust
commands.entity(entity).queue(entity_command::insert(bundle).handle_error_with(error_handler::warn()));
```

cart:
> But it's weird that this is not unified with the system result handler

Hmm yeah we should think about how to unify these. I think applying the global handler to the system result handler by default makes some sense. I'd prefer that we tackle unification in a separate PR though.

cart:
> I had some idea that we could be returning these errors to the scheduler through the sync-point system, idk if that's a viable.

Worth considering all of our options. Imo the direct handler approach is nice because it gives us both granularity and nice performance (no need to queue results up, look up which handler to use, etc). I think we should start from there with essentially optimal performance and then compare against the fancier options later.

>  I don't really think this will be a niche feature though. Don't we anticipate the editor using custom error handling to populate an error console? But that's nothing to do with the implementation

Hmmm yeah thats true. Although thats an implementation we'd own ourselves, and editor integration would likely be behind a cargo feature (so we could abstract that out from a user perspective). I think the majority of users shouldn't need to explicitly _think_ about custom global error handlers.

NthTensor:
All fair points. I'm pretty happy with this, there's obviously going to be more work to do here but it will be in the long tail of gradual improvement. The changes to commands are really nice.

crates/bevy_ecs/src/system/commands/mod.rs
```diff
@@ -631,66 +611,9 @@ impl<'w, 's> Commands<'w, 's> {
...
-
-    /// Resets the [`Commands`] instance's error handler, allowing commands
-    /// to respond to errors in their default manner.
-    ///
-    /// This is only useful if the instance's error handler was previously overridden
-    /// by [`Self::override_error_handler`].
-    pub fn reset_error_handler(&mut self) {
-        self.error_handler_override = None;
+    /// This will use the default error handler. See [`Command`] for details on error handling.
+    pub fn queue_fallible<E: Into<Error>>(&mut self, command: impl Command<Result<(), E>>) {
```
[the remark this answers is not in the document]
cart: This was largely just to leave doors open in case we find a use for "commands that produce outputs". My thought was that its probably preferable to outright not support outputs here than to silently ignore them, but I don't have strong feelings.

JaySpruce:
> This is already supported:
>```
>commands.entity(entity).queue_fallible(entity_command::insert(bundle).handle_error_with(error_handler::warn()));
>```

Sorry, I'm still not seeing it. Wouldn't that result in an `EntityFetchError` being handled by the default error handler?
```
move |world: &mut World {			   <- handle_error      (default error handler)
    move |world: &mut World| {                     <- with_entity
	let entity = world.get_entity_mut(entity)?;
        move |entity: EntityWorldMut| {            <- handle_error_with (user's error handler)
	    move |mut entity: EntityWorldMut| {    <- entity_command::insert
	        entity.insert_with_caller(bundle, InsertMode::Replace);
            }
        }
    }
}
```

cart:
I just pushed a commit that merges `queue` and `queue_fallible`.

> Sorry, I'm still not seeing it. Wouldn't that result in an EntityFetchError being handled by the default error handler?

Hmmm yeah I see your point. You can't customize the error handler for an internal `with_entity` call, whereas the `handle_error_with` call could be added to the end of the chain internally with a specialized function.

crates/bevy_ecs/src/system/commands/entity_command.rs
```diff
@@ -74,96 +76,117 @@ use bevy_ptr::OwningPtr;
...
+                Err(err) => (error_handler)(world, err),
+            }
+        }
     }
 }
 
-impl<F> EntityCommand<(World, Result)> for F
+/// Takes a [`EntityCommand`] that returns a [`Result`] with an error that can be converted into the [`Error`] type
+/// and returns a [`EntityCommand`] that internally converts that error to [`Error`] (if it occurs).
+pub fn map_entity_command_err<T, E: Into<crate::result::Error>>(
```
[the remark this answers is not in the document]
cart: Good call. That also helped with merging queue with queue_fallible

cart:
Fundamentally the problem @JaySpruce has called out is that the custom error handler resolves "too early" before we call `entity_command.with_entity` internally, resulting in the default error handler (panicking!) being used for the with_entity call. Given that `with_entity` is one of the more likely things to fail, we do need to prioritize resolving this.

As @JaySpruce said, adding `queue_fallible_with` does resolve this (I prefer `queue_with_handler`, `queue_with`, or `queue_handled` in the "unified queue" world). However I don't think adding that alone is optimal:

* It makes `handle_error_with` footgun-ey in the context of EntityCommands (its generally safe to use it in the context of Commands, unless you are manually replicating the same footgun-ey entity command order).
* It adds two way to do things, which I generally don't like
* The two ways to do things look functionally identical, but they have subtly different behaviors

I think we have a few options:

1. Remove the public `handle_error_with` function entirely (to resolve the issues above) and _require_ using `queue_handled` if you want to define a custom handler, both for EntityCommands _and_ Commands. This is a reduction in flexibility (as you can't pre-bake commands with custom handlers anymore), but I think in general it is fine.
2. Rework Command init to "tack on" the handler via a "command descriptor" object, deferring the final construction of the Command with the baked in handler until _after_ `with_entity` is called internally.
3. Just add `queue_handled` and accept that the corner case of calling `handle_error_with` and passing the result into EntityCommands will not use the custom handler for `with_entity` errors.

(1) does feel slightly inelegant / it removes functionality. But it does also make the UX simple / consistent / footgun free.
(2) provides the exact behavior we want at the cost of additional complexity and codegen (to generate the descriptor type). I don't love this solution.
(3) Is the easiest and allows people to do the right thing, but it complicates the UX and still allows people to shoot themselves in the foot.

hymm (commented):
> If anyone has other ideas let me know.

Not sure I like either of these ideas, but posting them since they have some merit.

5. Pass a `Result<EntityWorldMut, EntityFetchError>` instead of `EntityWorldMut` to EntityCommands. Not great, but just need to `?` to get the EntityWorldMut.
6. Restructure things so you can attach an error handler to `commands.entity(e).with_error_handler` and then pass the EntityWorldMut to each entity command without having to reconstruct it every time. This feels like a pretty complex change and probably would need a EntityCommandQueue or maybe a intermediate buffer of components to be stored somewhere. It would probably slow down the case where you have only one entity command.

crates/bevy_ecs/src/system/commands/mod.rs
```diff
@@ -1712,19 +1620,26 @@ impl<'a> EntityCommands<'a> {
...
     }
 
     /// Despawns the entity.
+    ///
     /// This will not emit a warning if the entity does not exist, essentially performing
     /// the same function as [`Self::despawn`] without emitting warnings.
     #[track_caller]
     pub fn try_despawn(&mut self) {
-        self.queue(despawn(false));
+        self.queue(entity_command::despawn());
```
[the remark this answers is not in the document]
cart: Probably a comment for the other (now merged) PR, but I think I agree.

crates/bevy_ecs/src/system/commands/command.rs
```diff
@@ -0,0 +1,314 @@
...
+        self(world)
+    }
+}
+
+/// Takes a [`Command`] that returns a Result and uses a given error handler function to convert it into
+/// a [`Command`] that internally handles an error if it occurs and returns `()`.
+pub trait HandleError<T = ()> {
+    /// Takes a [`Command`] that returns a Result and uses a given error handler function to convert it into
+    /// a [`Command`] that internally handles an error if it occurs and returns `()`.
+    fn handle_error_with(self, error_handler: fn(&mut World, Error)) -> impl Command;
```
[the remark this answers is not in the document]
cart: Worth considering!

cart:
I've opted to remove the `HandleEntityError` trait entirely, add `Commands::queue_handled` and `EntityCommands::queue_handled`, and leave `HandleError` as-is. This prevents the weirdness by embracing the fact that directly tacking on `EntityCommand` error handling is disallowed.  `HandleError` is left in because this pattern is still valid and we need a public trait to drive this functionality.

Resolving some CI things and sorting out what `try` command variants should look like, then I think this is ready.

crates/bevy_ecs/src/system/commands/mod.rs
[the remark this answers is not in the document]
cart: I've made despawn and try_despawn use their old behavior for now (warn() and silent() respectively).

cart:
I've made the `try_` variants behave as expected (how they did previously). I think its worth reconsidering the `try_x` APIs as shorthand for manually calling `queue_handled(entity_command::x, error_handler::warn())`, but I think thats a topic for a separate PR.

I think this PR is now ready to merge.

crates/bevy_ecs/src/system/commands/mod.rs
```diff
@@ -92,37 +91,25 @@ use crate::{
...
-/// Commands return a [`Result`](crate::result::Result), which gets passed to
+/// Commands can return a [`Result`](crate::result::Result), which can be passed to
 /// an error handler. Error handlers are functions/closures of the form
 /// `fn(&mut World, CommandError)`.
 ///
-/// By default, if a command returns an error, it will be passed to the
-/// global error handler. Currently, the global error handler just panics;
-/// in the future, this will be configurable.
+/// Fallible commands should generally be passed to [`Commands::queue_fallible`], which will
+/// use the default error handler.
```
JeanMertz, line 99: outdated documentation – `queue_fallible` no longer exists.

crates/bevy_ecs/src/system/commands/mod.rs
```diff
@@ -1189,24 +1149,18 @@ impl<'w, 's> Commands<'w, 's> {
...
-/// Entity commands return a [`Result`](crate::result::Result), which gets passed to
+/// [`EntityCommands`] can return a [`Result`](crate::result::Result), which can be passed to
 /// an error handler. Error handlers are functions/closures of the form
 /// `fn(&mut World, CommandError)`.
 ///
-/// By default, if a command returns an error, it will be passed to the
-/// global error handler. Currently, the global error handler just panics;
-/// in the future, this will be configurable.
+/// Fallible commands should generally be passed to [`EntityCommands::queue_fallible`], which will
+/// use the default error handler.
```
JeanMertz, line 1157: outdated documentation – queue_fallible no longer exists.

crates/bevy_ecs/src/system/commands/command.rs
```diff
@@ -34,102 +34,70 @@ use crate::{
...
-/// of `Command` for closures, like so:
-/// ```ignore (This would conflict with the real implementations)
-/// impl Command for FnOnce(&mut World)
-/// impl Command<Result> for FnOnce(&mut World) -> Result
-/// ```
-/// Without the generic, Rust would consider the two implementations to be conflicting.
-///
-/// The type used for `Marker` has no connection to anything else in the implementation.
-pub trait Command<Marker = ()>: Send + 'static {
+pub trait Command<T = ()>: Send + 'static {
```
alice-i-cecile, line 47: T needs docs or a better name. Ideally both!

crates/bevy_ecs/src/system/commands/entity_command.rs
```diff
@@ -74,96 +76,75 @@ use bevy_ptr::OwningPtr;
...
-pub trait EntityCommand<Marker = ()>: Send + 'static {
+pub trait EntityCommand<T = ()>: Send + 'static {
     /// Executes this command for the given [`Entity`] and
     /// returns a [`Result`] for error handling.
-    fn apply(self, entity: Entity, world: &mut World) -> Result;
+    fn apply(self, entity: EntityWorldMut) -> T;
+}
+/// Passes in a specific entity to an [`EntityCommand`], resulting in a [`Command`] that
+/// internally runs the [`EntityCommand`] on that entity.
+pub trait CommandWithEntity<T> {
```
alice-i-cecile, line 86: Comments here on why this rather convoluted design was needed would be useful. A link back to this PR thread would be sufficient even.

crates/bevy_ecs/src/system/commands/entity_command.rs
```diff
@@ -74,96 +76,75 @@ use bevy_ptr::OwningPtr;
...
-    }
+/// An error that occurs when running an [`EntityCommand`] on a specific entity.
+#[derive(thiserror::Error, Debug)]
+pub enum EntityCommandError<E> {
+    /// The entity this [`EntityCommand`] tried to run on could not be fetched.
+    #[error(transparent)]
+    EntityFetchError(#[from] EntityFetchError),
+    /// An error that occurred while running the [`EntityCommand`].
+    #[error("{0}")]
+    Error(E),
```
alice-i-cecile, line 132: This variant name is pretty bad. `EntityCommandError::Error`? I liked the `CommandFailed` name from the last PR.

crates/bevy_ecs/src/system/commands/error_handler.rs
```diff
@@ -1,32 +1,53 @@
...
+#[cfg(not(feature = "configurable_error_handler"))]
+#[inline]
+pub fn default() -> fn(&mut World, Error) {
+    panic()
+}
+
+/// A global error handler. This can be set at startup, as long as it is set before
+/// any uses. This should generally be configured _before_ initializing the app.
+///
+/// If the `configurable_error_handler` cargo feature is enabled, this will be used
```
alice-i-cecile, line 40: OnceLocks with a function pointer are a bit exotic for something that's this user facing. Can we add a simple doc test teaching users how to set this?

crates/bevy_ecs/src/system/commands/mod.rs
```diff
@@ -1781,83 +1710,53 @@ impl<'a> EntityCommands<'a> {
...
+    ///     );
+    /// # }
+    /// # bevy_ecs::system::assert_is_system(my_system);
+    /// ```
+    pub fn queue_handled<C: EntityCommand<T> + CommandWithEntity<M>, T, M>(
         &mut self,
-        command: impl EntityCommand<M>,
-        default_error_handler: fn(&mut World, CommandError),
+        command: C,
+        error_handler: fn(&mut World, Error),
```
alice-i-cecile, line 1756: We're using boxed errors here, but in my ideal world this "add an error handler to a specific command" would handle receive the non-erased error type here. That gives us much better recovery options, beyond just log / panic / ignore.

 Can we change that?

alice-i-cecile (commented):
Good cleanup! The "EntityCommands ~= EntityWorldMut` changes are great, and I like the new helper methods on EntityWorldMut.

Requests for better documentation (shocking), and a more serious question about getting the concrete error type in queue_handled. I think this is an improvement and I wouldn't be upset to see this merged, but I think that getting that last detail right is valuable.

crates/bevy_ecs/src/system/commands/mod.rs
cart: I do agree that this could probably handle non-erased error types instead in the "specific-command error handler" case. I spent a bit of time investigating and got this largely working. I added a `type Error` associated type to `HandleError` and `CommandWithEntity`, and added an `ErrorHandle<E>`  type alias for a generic error handler function. This works for adding error handlers to specific commands, but is a problem: The "default error handler" fundamentally requires the general-purpose bevy `Error` type, making it (not directly) usable for a signature that expects a specific error type. We want `queue` to be "default error conversion compatible" and `queue_handled` to be flexible. I'm reasonably certain this requires a duplication of traits.

This is a big enough change that I think my preference is to defer it to a later PR.

alice-i-cecile (approved):

PR #17216, opened 2025-01-07
Stop automatically generating meta files for assets while using asset processing.

andriyDev:
# Objective

- Today, enabling asset processing can generate many meta files. This makes it a painful transition for users as they get a "mega commit" containing tons of meta files.

## Solution

- Stop automatically generating meta files! Users can just leave the meta files defaulted.
- Add a function `AssetServer::write_default_meta_file_for_path`

## Testing

- Tested this manually on the asset_processing example (by removing the meta files for the assets that had default meta files).
- I did not add a unit test for the `write_default_meta_file_for_path` since we don't have an in-memory asset writer. Writing one could be useful in the future.

---

## Showcase

Asset processing no longer automatically generates meta files! This makes it much easier to transition to using asset processing since you don't suddenly get many meta files when turning it on.

You can still manually generate meta files using the new `AssetServer::write_default_meta_file_for_path` function.


andriyDev:
I don't think this PR needs a migration guide since users don't need to change any existing stuff for their assets to work. New assets will also continue to work, though new meta files will have to be created using the provided function or manually created.

viridia (approved):
Looks fine to me. Do we want to wait for cart to weigh in on this?

viridia:
In terms of guides, I can write up something once all changes are in and the dust has settled. Not a migration guide so much as a description of the new workflow.

viridia (approved):

BenjaminBrienen:
FYI @dreamertalin @nisevoid @ethereumdegen 

And for reference:
https://discord.com/channels/691052431525675048/749332104487108618/1324891672127078460

(This is somewhat blessed)

mockersf (changes requested):
to keep the same behaviour, I would prefer if the new function didn't overwrite files

andriyDev:
I went with returning an error if the meta file already exists (as opposed to silently succeeding). Unfortunately this required me to try to read the meta file just to check if it's there, but that doesn't seem too bad.

crates/bevy_asset/src/server/mod.rs
```diff
@@ -1466,6 +1467,40 @@ impl AssetServer {
...
+
+        let meta = loader.default_meta();
+        let serialized_meta = meta.serialize();
+
+        let source = self.get_source(path.source())?;
+
+        let reader = source.reader();
+        match reader.read_meta_bytes(path.path()).await {
+            Ok(_) => return Err(WriteDefaultMetaError::MetaAlreadyExists),
+            Err(_) => {
```
BenjaminBrienen, line 1490: Is it possible to match on the true error case and early return here directly?

andriyDev: The thing I'm unclear of is what to do if this fails for a non-`NotFound` reason? Should we propagate the `AssetReaderError`? This seems a little undesirable, especially since this is a writing function.

andriyDev: After chatting on Discord, I decided to just duplicate all the `AssetReaderError` variants. This prevents propagating the NotFound variant, and there's few enough variants that it's ok.

BenjaminBrienen (approved):

crates/bevy_asset/src/server/mod.rs
```diff
@@ -1479,6 +1480,44 @@ impl AssetServer {
             }
         }
     }
+
+    /// Writes the default meta file for the provided `path`.
+    ///
+    /// Note if there is already a meta file for `path`, this function returns
+    /// `Err(WriteDefaultMetaError::MetaAlreadyExists)`.
+    pub async fn write_default_meta_file_for_path(
```
cart, line 1488: I do like this general path, but using this method to generate the meta files for a project would ignore the "default processor configuration", as it just writes the default loader meta.

I'm cool with this method as-is, but I think we need an equivalent on the AssetProcessor as well that looks for a default processor configuration first and uses that _and then_ falls back to the logic here (aka the same logic that the processor uses currently when finding what meta to use at runtime).

Ultimately I think we'll want something that orchestrates the creation of meta files for a project for a given set of filters + configurations, but we can build that if and when we decide its needed.

andriyDev: Done. Renamed this function to `write_default_loader_meta_file_for_path` and wrote a new version `AssetProcessor::write_default_meta_file_for_path`.

alice-i-cecile:
Waiting for @cart's re-review before merging, but if that's not forthcoming I intend to merge this before we cut the 0.16 release candidate :)

cart (approved):

alice-i-cecile:
Thank you to everyone involved with the authoring or reviewing of this PR! This work is relatively important and needs release notes! Head over to https://github.com/bevyengine/bevy-website/issues/2000 if you'd like to help out.

PR #17219, opened 2025-01-07
Move `TextureAtlas` and friends into `bevy_image`

mgi388:
# Objective

- Allow other crates to use `TextureAtlas` and friends without needing to depend on `bevy_sprite`.
- Specifically, this allows adding `TextureAtlas` support to custom cursors in https://github.com/bevyengine/bevy/pull/17121 by allowing `bevy_winit` to depend on `bevy_image` instead of `bevy_sprite` which is a [non-starter].

[non-starter]: https://github.com/bevyengine/bevy/pull/17121#discussion_r1904955083

## Solution

- Move `TextureAtlas`, `TextureAtlasBuilder`, `TextureAtlasSources`, `TextureAtlasLayout` and `DynamicTextureAtlasBuilder` into `bevy_image`.
- Add a new plugin to `bevy_image` named `TextureAtlasPlugin` which allows us to register `TextureAtlas` and `TextureAtlasLayout` which was previously done in `SpritePlugin`. Since `SpritePlugin` did the registration previously, we just need to make it add `TextureAtlasPlugin`.

## Testing

- CI builds it.
- I also ran multiple examples which hopefully covered any issues:

```
$ cargo run --example sprite
$ cargo run --example text
$ cargo run --example ui_texture_atlas
$ cargo run --example sprite_animation
$ cargo run --example sprite_sheet
$ cargo run --example sprite_picking
```

---

## Migration Guide

The following types have been moved from `bevy_sprite` to `bevy_image`: `TextureAtlas`, `TextureAtlasBuilder`, `TextureAtlasSources`, `TextureAtlasLayout` and `DynamicTextureAtlasBuilder`.

If you are using the `bevy` crate, and were importing these types directly (e.g. before `use bevy::sprite::TextureAtlas`), be sure to update your import paths (e.g. after `use bevy::image::TextureAtlas`)

If you are using the `bevy` prelude to import these types (e.g. `use bevy::prelude::*`), you don't need to change anything.

If you are using the `bevy_sprite` subcrate, be sure to add `bevy_image` as a dependency if you do not already have it, and be sure to update your import paths.

### DIFF
--- a/crates/bevy_image/src/lib.rs
+++ b/crates/bevy_image/src/lib.rs
@@ -47,16 +47,3 @@ pub use texture_atlas_builder::*;
 
 pub(crate) mod image_texture_conversion;
 pub use image_texture_conversion::IntoDynamicImageError;
-
-use bevy_app::prelude::*;
-use bevy_asset::AssetApp as _;
-/// Adds support for texture atlases.
-pub struct TextureAtlasPlugin;
-
-impl Plugin for TextureAtlasPlugin {
-    fn build(&self, app: &mut App) {
-        app.init_asset::<TextureAtlasLayout>()
-            .register_asset_reflect::<TextureAtlasLayout>()
-            .register_type::<TextureAtlas>();
-    }
-}
--- a/crates/bevy_image/src/lib.rs
+++ b/crates/bevy_image/src/lib.rs
@@ -1,8 +1,14 @@
 #![expect(missing_docs, reason = "Not all docs are written yet, see #3492.")]
 #![allow(unsafe_code)]
 
+extern crate alloc;
+
 pub mod prelude {
-    pub use crate::{BevyDefault as _, Image, ImageFormat, TextureError};
+    pub use crate::{
+        dynamic_texture_atlas_builder::DynamicTextureAtlasBuilder,
+        texture_atlas::{TextureAtlas, TextureAtlasLayout, TextureAtlasSources},
+        BevyDefault as _, Image, ImageFormat, TextureAtlasBuilder, TextureError,
+    };
 }
 
 mod image;
@@ -13,6 +19,7 @@ mod basis;
 mod compressed_image_saver;
 #[cfg(feature = "dds")]
 mod dds;
+mod dynamic_texture_atlas_builder;
 #[cfg(feature = "exr")]
 mod exr_texture_loader;
 #[cfg(feature = "hdr")]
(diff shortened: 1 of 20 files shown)


crates/bevy_image/src/lib.rs
```diff
@@ -13,25 +19,44 @@ mod basis;
...
+/// Adds support for texture atlases.
+pub struct TextureAtlasPlugin;
+
+impl Plugin for TextureAtlasPlugin {
+    fn build(&self, app: &mut App) {
+        app.init_asset::<TextureAtlasLayout>()
+            .register_asset_reflect::<TextureAtlasLayout>()
+            .register_type::<TextureAtlas>();
+    }
+}
```
kristoff3r, line 62: I would either move this to `texture_atlas.rs` or rename it to something like `ImagePlugin` which could then potentially be used for other stuff.

kristoff3r (approved):

crates/bevy_sprite/src/lib.rs
```diff
@@ -111,13 +104,15 @@ impl Plugin for SpritePlugin {
             "render/sprite_view_bindings.wgsl",
             Shader::from_wgsl
         );
-        app.init_asset::<TextureAtlasLayout>()
-            .register_asset_reflect::<TextureAtlasLayout>()
-            .register_type::<Sprite>()
+
+        if !app.is_plugin_added::<TextureAtlasPlugin>() {
+            app.add_plugins(TextureAtlasPlugin);
```
eero-lehtinen, line 109: Should this go to default plugins or somehow automatically enabled for games not using sprites?

mgi388: I figured it was best to preserve the existing behavior: The texture atlas stuff was only ever enabled if you added `SpritePlugin`.

If you make a 3d game and don't have `SpritePlugin`, I am not sure that you should get `TextureAtlasPlugin` enabled just because it was in `DefaultPlugins`.

That was my thinking at least, not to say I'm right.

eero-lehtinen: Yeah I guess we can preserve the current behaviour for now. Then make sure that custom cursors work in 3d examples in the other pr and fix it if not.

crates/bevy_image/src/lib.rs
mgi388: Moved to texture_atlas.rs. I'm not too fussed either way, but possibly inclined to keep it as a micro plugin for now.

I suppose if there is more in `bevy_image` that seems to suggest callers want an `ImagePlugin` to encapsulate all these things, we could add that later?

eero-lehtinen (approved):

crates/bevy_sprite/src/lib.rs
```diff
@@ -111,13 +104,15 @@ impl Plugin for SpritePlugin {
             "render/sprite_view_bindings.wgsl",
             Shader::from_wgsl
         );
-        app.init_asset::<TextureAtlasLayout>()
-            .register_asset_reflect::<TextureAtlasLayout>()
-            .register_type::<Sprite>()
+
+        if !app.is_plugin_added::<TextureAtlasPlugin>() {
```
BenjaminBrienen, line 108: "If plugin X isn't added, add plugin X". Is that necessary to specify? If the plugin is already added, I would think that adding the plugin again doesn't do anything. If it does do something, then why is this the only place that check is done?

alice-i-cecile (approved):

PR #17223, opened 2025-01-07
Update `downcast-rs` to version 2

bushrat011899:
# Objective & Solution

- Update `downcast-rs` to the latest version, 2.
- Disable (new) `sync` feature to improve compatibility with atomically challenged platforms.
- Remove stub `downcast-rs` alternative code from `bevy_app`

## Testing

- CI

## Notes

The only change from version 1 to version 2 is the addition of a new `sync` feature, which allows disabling the `DowncastSync` parts of `downcast-rs`, which require access to `alloc::sync::Arc`, which is not available on atomically challenged platforms. Since Bevy makes no use of the functionality provided by the `sync` feature, I've disabled it in all crates. Further details can be found [here](https://github.com/marcianx/downcast-rs/pull/22).

### DIFF
--- a/crates/bevy_animation/Cargo.toml
+++ b/crates/bevy_animation/Cargo.toml
@@ -32,7 +32,7 @@ petgraph = { version = "0.6", features = ["serde-1"] }
 ron = "0.8"
 serde = "1"
 blake3 = { version = "1.0" }
-downcast-rs = "1.2.0"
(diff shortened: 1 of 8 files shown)


alice-i-cecile (approved):
Nice work upstream!

BenjaminBrienen (approved):

Issue #16783, opened 2024-12-12
`Text2d` and `TextSpan` is not center-aligned when `JustifyText::Center` is specified.


Ramirisu:
## Bevy version

0.15

## What went wrong

In bevy 0.14, the text sections correctly reflect the `JustifyText::Center` parameter, texts are center-aligned:

```rust
fn setup(mut commands: Commands) {
    commands.spawn(Camera2dBundle::default());

    commands.spawn(Text2dBundle {
        text: Text::from_sections([
                        TextSection {
                value: "Text Justification\n".into(),
                style: TextStyle {
                    color: WHITE.into(),
                    ..default()
                },
                ..default()
            },
            TextSection {
                value: "Center\n".into(),
                style: TextStyle {
                    color: YELLOW.into(),
                    ..default()
                },
                ..default()
            },
            TextSection {
                value: "xxxxxxxxxxCenterxxxxxxxxxx".into(),
                style: TextStyle {
                    color: RED.into(),
                    ..default()
                },
                ..default()
            },
        ])
        .with_justify(JustifyText::Center),
        transform: Transform::from_translation(Vec3::default()),
        ..default()
    });
}

```
![Image](https://github.com/user-attachments/assets/d7e76b8c-64e6-478c-9dc7-efc5ce9bf998)

However, after upgrading to bevy 0.15, the `Text2d` and `TextSpan` are always left-aligned even though the `TextLayout::new_with_justify(JustifyText::Center)` is specified:

```rust

fn setup(mut commands: Commands) {
    commands.spawn(Camera2d::default());

    commands
        .spawn((
            Text2d::new("Text Justification\n"),
            TextLayout::new_with_justify(JustifyText::Center),
            TextColor::from(WHITE),
            Transform::from_translation(Vec3::default()),
        ))
        .with_children(|parent| {
            parent.spawn((TextSpan::new("Center\n"), TextColor::from(YELLOW)));
            parent.spawn((
                TextSpan::new("xxxxxxxxxxCenterxxxxxxxxxx"),
                TextColor::from(RED),
            ));
        });
}

```

![Image](https://github.com/user-attachments/assets/de2dae9c-9384-4487-8550-50d39cb8ca93)

MikMikus:
Does anyone have a working temporary solution?

Btw, do you also notice a performance drop when using Text2d with TextSpan?
In my case - hundreds labels with text - i have:
- ~5 fps for Text2d with TextSpan 
- ~20 fps when i use only Text2d
[...]

PR #17224, opened 2025-01-07
Add a simple directional UI navigation example

alice-i-cecile:
# Objective

Gamepad / directional navigation needs an example, for both teaching and testing purposes.

## Solution

- Add a simple grid-based example.
- Fix an intermittent panic caused by a race condition with bevy_a11y
- Clean up small issues noticed in bevy_input_focus

image

## To do: this PR

- [x] figure out why "enter" isn't doing anything
- [x] change button color on interaction rather than printing
- [x] add on-screen directions
- [x] move to an asymmetric grid to catch bugs
- [x] ~~fix colors not resetting on button press~~ lol this is mostly just a problem with hacking `Interaction` for this
- [x] swap to using observers + bubbling, rather than `Interaction`

## To do: future work

- when I increase the button size, such that there is no line break, the text on the buttons is no longer centered :( EDIT: this is https://github.com/bevyengine/bevy/issues/16783
- add gamepad stick navigation
- add tools to find the nearest populated quadrant to make diagonal inputs work
- add a `add_edges` method to `DirectionalNavigationMap`
- add a `add_grid` method to `DirectionalNavigationMap`
- make the example's layout more complex and realistic
- add tools to automatically generate this list
- add button shake on failed navigation rather than printing an error
- make Pressed events easier to mock: default fields, PointerId::Focus

## Testing

`cargo run --example directional_navigation`


rparrett:
Sent a couple tiny PRs your way

examples/ui/directional_navigation.rs
```diff
@@ -0,0 +1,417 @@
...
+) {
+    for (mut reset_timer, mut color) in query.iter_mut() {
+        reset_timer.tick(time.delta());
+        if reset_timer.just_finished() {
+            color.0 = NORMAL_BUTTON.into();
+        }
+    }
+}
+
+// We're spawning a simple 3x3 grid of buttons
```
rparrett, line 101:
```suggestion
// We're spawning a simple 5x3 grid of buttons
```

examples/ui/directional_navigation.rs
```diff
@@ -0,0 +1,417 @@
...
+            },
+        ))
+        .id();
+
+    // Set up the root entity to hold the grid
+    let grid_root_entity = commands
+        .spawn(Node {
+            display: Display::Grid,
+            // Allow the grid to take up the full height and width of the window
+            width: Val::Percent(60.),
```
rparrett, line 143: IMO

- Make instructions node wider, maybe `Px(300.0)` 
- Leave this at `100.0`.
- Adjust this comment to say "take up the full height and the rest of the width."

Seems okay for the grid to not be centered in the screen to me.

examples/ui/directional_navigation.rs
```diff
@@ -0,0 +1,417 @@
...
+                    },
+                    ResetTimer::default(),
+                    BorderRadius::all(Val::Px(16.0)),
+                    BackgroundColor::from(NORMAL_BUTTON),
+                    Name::new(button_name.clone()),
+                ))
+                // Add a text element to the button
+                .with_child((
+                    Text::new(button_name),
+                    // And center the text within the label
```
rparrett, line 187:
```suggestion
                    // And center the text if it flows onto multiple lines
```

examples/ui/directional_navigation.rs
```diff
@@ -0,0 +1,417 @@
...
+const NORMAL_BUTTON: Srgba = bevy::color::palettes::tailwind::BLUE_400;
+const PRESSED_BUTTON: Srgba = bevy::color::palettes::tailwind::BLUE_500;
+const FOCUSED_BORDER: Srgba = bevy::color::palettes::tailwind::BLUE_50;
+
+// This observer will be triggered whenever a button is pressed
+// In a real project, each button would also have its own unique behavior,
+// to capture the actual intent of the user
+fn universal_button_click_behavior(
+    // We're using a on-mouse-down trigger to improve responsiveness;
+    // Clicked is better when you want roll-off cancellation
```
rparrett, line 69: Might be nicer to have separate effects that get triggered on `Pressed` and `Clicked` rather than explain why it looks like we're doing the "wrong" thing.

But that could be a future improvement.

examples/ui/directional_navigation.rs
```diff
@@ -0,0 +1,417 @@
...
+    // Add instruction to the left of the grid
+    let instructions = commands
+        .spawn((
+            Text::new("Use arrow keys or D-pad to navigate. \
+            Click the buttons, or press Enter / the South gamepad button to interact with the focused button."),
+            Node {
+                width: Val::Px(200.0),
+                justify_content: JustifyContent::Center,
+                align_items: AlignItems::Center,
+                margin: UiRect::all(Val::Px(16.0)),
```
rparrett, line 132:
```suggestion
                margin: UiRect::all(Val::Px(12.0)),
```
For consistency with other examples with top-left instructions

rparrett (changes requested):

rparrett:
Would you mind rewriting that last commit message without the tag?

Approving, but continuing to nitpick to clarify the reasoning for a previous review.

image

I think that either

- The grid node should be `width: 100%` so that it takes up the rest of the space and the buttons are centered within the non-instruction space
- The instructions node should given a percentage width of `(100% - grid%) / 2` so that things are centered
- The grid node should be `width: 60%` or whatever, but the text node absolutely positioned.

My eyes are just drawn to the "implicit green box" in the layout and it looks off-center.

viridia (approved):

PR #17225, opened 2025-01-07
Make sprite picking opt-in

### LINKED ISSUE #16903: Consider changing the default picking behavior of sprites to be non-pickable by default
From [this discord conversation](https://discord.com/channels/691052431525675048/768253008416342076/1319121525831499817):

> While I think having all UI nodes pickable by default is a very good choice, making all sprites pickable and blocking lower entities by default is not ideal, especially as sprites tend to be used for misc non-interactive things (like overlays, or nested cameras for rendering at lower resolutions) more often than not. You end up having to patch lots of spawning code adding `PickingBehavior` to suppress that behavior.

Also, some precedent from mesh picking, which has the `RayCastPickable` component.

(Also, if #16886 is merged, this should probably also apply to that.)

chompaa:
# Objective

Fixes #16903.

## Solution

- Make sprite picking opt-in by requiring a new `SpritePickingCamera` component for cameras and usage of a new `Pickable` component for entities.
- Update the `sprite_picking` example to reflect these changes.
- Some reflection cleanup (I hope that's ok).

## Testing

Ran the `sprite_picking` example

## Open Questions

<del>
   <ul>
    <li>Is the name `SpritePickable` appropriate?</li>
    <li>Should `SpritePickable` be in `bevy_sprite::prelude?</li>
  </ul> 
</del>

## Migration Guide

The sprite picking backend is now strictly opt-in using the `SpritePickingCamera` and `Pickable` components. You should add the `Pickable` component any entities that you want sprite picking to be enabled for, and mark their respective cameras with `SpritePickingCamera`.

### DIFF
--- a/crates/bevy_picking/src/backend.rs
+++ b/crates/bevy_picking/src/backend.rs
@@ -41,10 +41,18 @@ use bevy_reflect::Reflect;
 pub mod prelude {
     pub use super::{ray::RayMap, HitData, PointerHits};
     pub use crate::{
+        backend::Pickable,
         pointer::{PointerId, PointerLocation},
         PickSet, PickingBehavior,
     };
 }
+
+/// Component used to mark entities as pickable by backends.
+///
+/// You should use this component if you are using a backend with opt-in behavior.
+#[derive(Component, Reflect, Debug)]
+#[reflect(Component, Debug)]
+pub struct Pickable;
 
 /// An event produced by a picking backend after it has run its hit tests, describing the entities
 /// under a pointer.
--- a/crates/bevy_picking/src/backend.rs
+++ b/crates/bevy_picking/src/backend.rs
@@ -41,18 +41,10 @@ use bevy_reflect::Reflect;
 pub mod prelude {
     pub use super::{ray::RayMap, HitData, PointerHits};
     pub use crate::{
-        backend::Pickable,
         pointer::{PointerId, PointerLocation},
         PickSet, PickingBehavior,
     };
 }
-
-/// Component used to mark entities as pickable by backends.
-///
-/// You should use this component if you are using a backend with opt-in behavior.
-#[derive(Component, Reflect, Debug)]
-#[reflect(Component, Debug)]
-pub struct Pickable;
 
 /// An event produced by a picking backend after it has run its hit tests, describing the entities
 /// under a pointer.
--- a/crates/bevy_sprite/src/picking_backend.rs
+++ b/crates/bevy_sprite/src/picking_backend.rs
@@ -5,7 +5,7 @@
 //! # Usage
 //!
 //! This backend is strictly opt-in. For entities to be considered for sprite picking, you should
-//! mark them and their respective cameras with the [`SpritePickable`] component.
+//! mark them with [`Pickable`] and their respective cameras with [`SpritePickingCamera`].
 
 use crate::{Sprite, TextureAtlasLayout};
 use bevy_app::prelude::*;
--- a/crates/bevy_sprite/src/picking_backend.rs
+++ b/crates/bevy_sprite/src/picking_backend.rs
@@ -17,6 +17,7 @@ use bevy_window::PrimaryWindow;
 
 /// How should the [`SpritePickingPlugin`] handle picking and how should it handle transparent pixels
 #[derive(Debug, Clone, Copy, Reflect)]
+#[reflect(Debug)]
 pub enum SpritePickingMode {
     /// Even if a sprite is picked on a transparent pixel, it should still count within the backend.
     /// Only consider the rect of a given sprite.
@@ -30,6 +31,12 @@ pub enum SpritePickingMode {
 #[derive(Resource, Reflect)]
 #[reflect(Resource, Default)]
 pub struct SpritePickingSettings {
+    /// When set to `true` sprite picking will only consider cameras and entities explicitly marked
+    /// with [`SpritePickable`]. `true` by default.
+    ///
+    /// This setting is provided to give you fine-grained control over which cameras and entities
+    /// should be used by the sprite picking backend at runtime.
+    pub require_markers: bool,
     /// Should the backend count transparent pixels as part of the sprite for picking purposes or should it use the bounding box of the sprite alone.
     ///
     /// Defaults to an inclusive alpha threshold of 0.1
@@ -39,10 +46,19 @@ pub struct SpritePickingSettings {
 impl Default for SpritePickingSettings {
     fn default() -> Self {
         Self {
+            require_markers: true,
             picking_mode: SpritePickingMode::AlphaThreshold(0.1),
         }
     }
 }
+
+/// An optional component that marks cameras and target entities that should be used in the
+/// [`SpritePickingPlugin`].
+///
+/// Only needed if [`SpritePickingSettings::require_markers`] is set to `true`, and ignored otherwise.
+#[derive(Debug, Clone, Default, Component, Reflect)]
+#[reflect(Debug, Default, Component)]
+pub struct SpritePickable;
 
 #[derive(Clone)]
 pub struct SpritePickingPlugin;
@@ -50,6 +66,7 @@ pub struct SpritePickingPlugin;
 impl Plugin for SpritePickingPlugin {
     fn build(&self, app: &mut App) {
         app.init_resource::<SpritePickingSettings>()
+            .register_type::<(SpritePickingMode, SpritePickingSettings, SpritePickable)>()
             .add_systems(PreUpdate, sprite_picking.in_set(PickSet::Backend));
     }
 }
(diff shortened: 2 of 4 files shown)


crates/bevy_sprite/src/picking_backend.rs
```diff
@@ -39,17 +46,27 @@ pub struct SpritePickingSettings {
...
     }
 }
 
+/// An optional component that marks cameras and target entities that should be used in the
+/// [`SpritePickingPlugin`].
+///
+/// Only needed if [`SpritePickingSettings::require_markers`] is set to `true`, and ignored otherwise.
+#[derive(Debug, Clone, Default, Component, Reflect)]
+#[reflect(Debug, Default, Component)]
+pub struct SpritePickable;
```
alice-i-cecile, line 61: Part of me feels like this should be backend agnostic as `Pickable` 🤔

crates/bevy_sprite/src/picking_backend.rs
```diff
@@ -60,14 +77,21 @@ impl Plugin for SpritePickingPlugin {
...
+        &Projection,
+    )>,
     primary_window: Query<Entity, With<PrimaryWindow>>,
     images: Res<Assets<Image>>,
     texture_atlas_layout: Res<Assets<TextureAtlasLayout>>,
     settings: Res<SpritePickingSettings>,
     sprite_query: Query<(
         Entity,
         &Sprite,
+        Option<&SpritePickable>,
```
alice-i-cecile, line 94: Don't we need to use a `With` filter to actually get the desired performance benefits?

crates/bevy_sprite/src/picking_backend.rs
```diff
@@ -30,6 +31,12 @@ pub enum SpritePickingMode {
 #[derive(Resource, Reflect)]
 #[reflect(Resource, Default)]
 pub struct SpritePickingSettings {
+    /// When set to `true` sprite picking will only consider cameras and entities explicitly marked
+    /// with [`SpritePickable`]. `true` by default.
+    ///
+    /// This setting is provided to give you fine-grained control over which cameras and entities
+    /// should be used by the sprite picking backend at runtime.
+    pub require_markers: bool,
```
alice-i-cecile, line 39: I don't like this field: it adds a surprising behavior change with little benefit. We should pick a standard option, and then stick to it.

crates/bevy_sprite/src/picking_backend.rs
chompaa: I'm inclined to agree, but what if there existed a scenario where multiple backends are in use, and you want to opt-in to just one of them for a specific entity?

crates/bevy_sprite/src/picking_backend.rs
```diff
@@ -60,14 +77,21 @@ impl Plugin for SpritePickingPlugin {
 )]
 fn sprite_picking(
     pointers: Query<(&PointerId, &PointerLocation)>,
-    cameras: Query<(Entity, &Camera, &GlobalTransform, &Projection)>,
+    cameras: Query<(
+        Entity,
+        &Camera,
+        Option<&SpritePickable>,
```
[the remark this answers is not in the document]
chompaa: Agreed, I think there should be another marker component. Not sure on a name, though. `SpritePickingCamera` comes to mind.

BenjaminBrienen (approved):

crates/bevy_sprite/src/picking_backend.rs
alice-i-cecile: Each entity should only be covered by one backend. I don't think we should try to plan for extremely pathological cases like that.

alice-i-cecile (changes requested):
Looking better :) Final changes:

1. Move marker component to bevy_picking, and call it `Pickable`.
2. Split apart the components and call the camera one `SpritePickingCamera`.

Follow-up:

- use this design for mesh-picking as well

chompaa:
>     2. Split apart the components and call the camera one `SpritePickingCamera`.

What are your thoughts on a `PickingCamera` component in bevy_picking? I can see it being used in the mesh picking backend and the UI picking backend if we were to support diegetic UI.

I'd be happy to tackle the follow-up :)

chompaa:
> What are your thoughts on a `PickingCamera` component in bevy_picking? I can see it being used in the mesh picking backend and the UI picking backend if we were to support diegetic UI.

On second thought, I don't think this makes sense. Cameras could be used for distinct backends, so generalizing seems wrong.

BenjaminBrienen (approved):

alice-i-cecile:
> On second thought, I don't think this makes sense. Cameras could be used for distinct backends, so generalizing seems wrong.

Exactly my thinking.

alice-i-cecile (approved):

crates/bevy_picking/src/backend.rs
```diff
@@ -41,11 +41,19 @@ use bevy_reflect::Reflect;
...
         PickSet, PickingBehavior,
     };
 }
 
+/// Component used to mark entities as pickable by backends.
+///
+/// You should use this component if you are using a backend with opt-in behavior.
+#[derive(Component, Reflect, Debug)]
+#[reflect(Component, Debug)]
+pub struct Pickable;
```
aevyrie, line 55: This type existed in mod_picking before upstream. It then turned into `PickingBehavior`, an optional component. We should discuss before adding this back.

aevyrie, line 55: This type existed in mod_picking before upstream. It then turned into PickingBehavior, an optional component. We should discuss before adding this back.

aevyrie (changes requested):

crates/bevy_picking/src/backend.rs
aevyrie: I'd rather backends use the existing component. E.g. if a backed is set to opt-in, it should be looking for the `PickingBehavior` component.

https://docs.rs/bevy_mod_picking/latest/bevy_mod_picking/backend/prelude/struct.Pickable.html

That said, I'm not a huge name of the rename to `PickingBehavior`. I much prefer `Pickable` as a component that makes an entity pickable.

alice-i-cecile: I agree that we should unify this. No strong feelings on the naming.

crates/bevy_sprite/src/picking_backend.rs
```diff
@@ -1,6 +1,11 @@
 //! A [`bevy_picking`] backend for sprites. Works for simple sprites and sprite atlases. Works for
 //! sprites with arbitrary transforms. Picking is done based on sprite bounds, not visible pixels.
 //! This means a partially transparent sprite is pickable even in its transparent areas.
+//!
+//! # Usage
+//!
+//! This backend is strictly opt-in. For entities to be considered for sprite picking, you should
+//! mark them with [`Pickable`] and their respective cameras with [`SpritePickingCamera`].
```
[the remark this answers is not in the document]
alice-i-cecile: Hmm, the debugging use case is a good argument for making this configurable 🤔 I would like to unify this across all of our backends though.

crates/bevy_picking/src/backend.rs
chompaa: My initial thoughts were that these were describing two distinct things, but I see that converging on one makes sense now. I have a preference towards `Pickable` since I think it better conveys that it will be considered in the backend.

crates/bevy_sprite/src/picking_backend.rs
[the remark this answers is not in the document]
chompaa: My only qualm is that we lose some performance benefit, but I'm in agreement.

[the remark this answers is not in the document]
chompaa: There is the question of whether it should be opt in by default or not. Backend consistency says it should be opt out, but perhaps we want to push users to opt in?

chompaa:
I've pushed some changes:

- Removed `Pickable` in favor of using `PickingBehavior` (I'd like to follow up on a rename).
- Added `require_markers` to `SpritePickingSettings` (defaults to `false` to align with the mesh picking backend).

I also think that the `SpritePickingPlugin` should be explicitly added by the user, and not a part of `DefaultPlugins` (and in general, all backends should be), but that seems out of scope for this PR.

aevyrie:
> I've pushed some changes:
> 
> * Removed `Pickable` in favor of using `PickingBehavior` (I'd like to follow up on a rename).
> * Added `require_markers` to `SpritePickingSettings` (defaults to `false` to align with the mesh picking backend).
> 
> I also think that the `SpritePickingPlugin` should be explicitly added by the user, and not a part of `DefaultPlugins` (and in general, all backends should be), but that seems out of scope for this PR.

Agree with everything here. Doing a pass to standardize all backends and make them all opt-in would be great.

crates/bevy_sprite/src/picking_backend.rs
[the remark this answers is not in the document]
aevyrie: From my experience working on mod_picking, it's a better experience for users if it starts as opt-out, because it remove friction and steps to just get things working. Once it is working, it's then another small step to switch to opt-in if they decide to.

PR #17228, opened 2025-01-08
Add documentation to `DynamicStorageBuffer` to clarify purpose

### LINKED ISSUE #15052: The doc for `DynamicStorageBuffer` doesn't really explain the difference with `StorageBuffer`
https://docs.rs/bevy/latest/bevy/render/render_resource/struct.DynamicStorageBuffer.html

> Each item [push](https://docs.rs/bevy/latest/bevy/render/render_resource/struct.DynamicStorageBuffer.html#method.push)ed into this structure will additionally be aligned to meet dynamic offset alignment requirements.

Compared with https://docs.rs/bevy/latest/bevy/render/render_resource/struct.StorageBuffer.html#method.write_buffer

> must conform to [std430 alignment/padding requirements](https://www.w3.org/TR/WGSL/#address-spaces-storage), which is automatically enforced by this structure.

Those sound like the same thing phrased in two different ways. The type `T` also has the same `ShadeType` bound, so there's no telling from that what a difference might be.

A table explaining the differences and pros/cons of all buffers, probably under the module's page, would be quite useful.

BenjaminBrienen:
# Objective

Fixes #15052

## Solution

Add sentence from @JMS55

### DIFF
--- a/crates/bevy_render/src/render_resource/storage_buffer.rs
+++ b/crates/bevy_render/src/render_resource/storage_buffer.rs
@@ -156,6 +156,8 @@ impl<'a, T: ShaderType + WriteInto> IntoBinding<'a> for &'a StorageBuffer<T> {
 
 /// Stores data to be transferred to the GPU and made accessible to shaders as a dynamic storage buffer.
 ///
+/// DynamicStorageBuffer is just a StorageBuffer, but also allows you to set dynamic offsets.
+///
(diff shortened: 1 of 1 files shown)


BenjaminBrienen:
Feel free to suggest a different comment. I'm no expert, just trying to implement it based on what I read in the issue.

djeedai (approved):

PR #17232, opened 2025-01-08
bevy_winit: Apply `#![deny(clippy::allow_attributes, clippy::allow_attributes_without_reason)]`

LikeLakers2:
# Objective

## Solution
Set the `clippy::allow_attributes` and `clippy::allow_attributes_without_reason` lints to `deny`, and bring `bevy_winit` in line with the new restrictions.

## Testing
`cargo clippy --tests` was run, and no errors were encountered.

### DIFF
--- a/crates/bevy_winit/src/winit_windows.rs
+++ b/crates/bevy_winit/src/winit_windows.rs
@@ -43,7 +43,10 @@ pub struct WinitWindows {
 
 impl WinitWindows {
     /// Creates a `winit` window and associates it with our entity.
-    #[allow(clippy::too_many_arguments)]
+    #[expect(
+        clippy::too_many_arguments,
+        reason = "Could be rewritten with less arguments using a QueryData-implementing struct, but doesn't need to be."
+    )]
     pub fn create_window(
         &mut self,
         event_loop: &ActiveEventLoop,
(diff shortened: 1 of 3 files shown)


crates/bevy_winit/src/winit_windows.rs
```diff
@@ -243,7 +246,11 @@ impl WinitWindows {
                 winit_window_attributes.with_min_inner_size(min_inner_size)
             };
 
-        #[allow(unused_mut)]
+        #[expect(clippy::allow_attributes, reason = "`unused_mut` is not always linted")]
+        #[allow(
+            unused_mut,
+            reason = "This variable needs to be mutable if `cfg(target_arch = "wasm32")`"
```
mnmaita, line 252: I have a feeling that the quotation marks around "wasm32" are messing this string up. You might need to escape them?

BenjaminBrienen, line 252:
```suggestion
            reason = "This variable needs to be mutable if `cfg(target_arch = \"wasm32\")`"
```

BenjaminBrienen (changes requested):

crates/bevy_winit/src/winit_windows.rs
LikeLakers2: Woops...

BenjaminBrienen (approved):

Issue #17230, opened 2025-01-08
Clicking the mouse once will also trigger DragStart and DragEnd


rendaoer:
## Bevy version

v0.15.1

## What you did

Yes, Drag is fixed, but what about DragStart and DragEnd? Why are they triggered even when there is no movement?

## What went wrong

In theory, a click of the mouse should only trigger Down and Up, and DragStart and DragEnd should not be triggered.

## Additional information

```rust
use bevy::prelude::*;

fn main() {
    App::new()
        .add_plugins(DefaultPlugins.set(ImagePlugin::default_nearest()))
        .add_systems(Startup, setup)
        .run();
}

fn setup(mut commands: Commands, asset_server: Res<AssetServer>) {
    commands.spawn(Camera2d);

    commands
        .spawn((
            Sprite {
                color: Color::srgb(1.0, 0.0, 0.0),
                custom_size: Some(Vec2::new(100.0, 100.0)),
                ..default()
            },
            Transform::from_xyz(0.0, 0.0, 0.0),
        ))
        .observe(down_handler)
        .observe(up_handler)
        .observe(drag_start_handler)
        .observe(drag_handler)
        .observe(drag_end_handler);
}

fn down_handler(_: Trigger<Pointer<Down>>) {
    info!("Down");
}

fn up_handler(_: Trigger<Pointer<Up>>) {
    info!("Up");
}

fn drag_start_handler(_: Trigger<Pointer<DragStart>>) {
    info!("DragStart");
}

fn drag_handler(_: Trigger<Pointer<Drag>>) {
    info!("Drag");
}

fn drag_end_handler(_: Trigger<Pointer<DragEnd>>) {
    info!("DragEnd");
}

```

PR #17233, opened 2025-01-08
If there is no movement, DragStart is not triggered.

rendaoer:
# Objective

Fixed the issue where DragStart was triggered even when there was no movement
https://github.com/bevyengine/bevy/issues/17230

## Solution

- When position delta is zero, don't trigger DragStart events, DragStart is not triggered, so DragEnd is not triggered either. Everything is fine.

## Testing

- tested with the code from the issue

---

## Migration Guide

> Fix the missing part of Drag https://github.com/bevyengine/bevy/pull/16950

### DIFF
--- a/crates/bevy_picking/src/events.rs
+++ b/crates/bevy_picking/src/events.rs
@@ -668,6 +668,9 @@ pub fn pointer_events(
 
                     // Emit DragEntry and DragStart the first time we move while pressing an entity
                     for (press_target, (location, _, hit)) in state.pressing.iter() {
+                        if delta == Vec2::ZERO {
+                            continue; // No need to emit a DragStart event if there is no movement
+                        }
                         if state.dragging.contains_key(press_target) {
                             continue; // This entity is already logged as being dragged
                         }


NthTensor:
Startled that direct float equality works here, and that the previous pr didn't use an epsilon.

NthTensor (changes requested):
IMO the check added by @mockersf in #16950 needs to be moved to line 664. Ether the pointer moves or it doesn't. This should be changing the semantics of the `PointerAction::Moved` event not which things we trigger.

rendaoer:
If the delta of a move event is zero, subsequent events should be skipped.

NthTensor (approved):

NthTensor:
Looking good, thanks. After further review, looks like the other check is a slightly different cross-frame delta. Looks correct now.

It's still weird that we don't need an epsilon here, but I'll take it.

alice-i-cecile (approved):
I think that the "raw" inputs that we get from the OS / input devices are filtered to remove tiny perturbations, which is why we end up not needing an epsilon. I've watched the mouse input streams before and they're shockingly un-jittery.

rendaoer:
If it moves, Drag should be triggered again, so since it has been checked above, the check in Drag below should be unnecessary, right?

rendaoer:
It doesn't seem to be a logical conclusion. Since it's correct, I won't bother with it anymore (sorry)

BenjaminBrienen (approved):

PR #17234, opened 2025-01-08
Allow tuple structs in animated_field! macro

### LINKED ISSUE #16736: Add support to the `animated_field!` macro for tuple structs

Components like `TextColor` and `BackgroundColor` cannot currently be easily animated without creating a custom `AnimatableProperty`.

Add support to the `animated_field!` macro (and consequently `AnimatedField`) for tuple structs. I'm not sure what the cleanest way to do this is from an implementation perspective, since the `$ident` pattern doesn't match tuple struct field indices and I'm not sure if it makes sense to use `$literal`, as that matches more than just field indices.

- Add an alternative easy way to animate tuple structs (e.g. a separate macro/type)
- Maybe something could be done with `Deref` and a blanket implementation of `AnimatableProperty`? But this puts the burden on components to implement `Deref`, only works for tuple structs with single fields, and also introduces potential conflicts with trait implementations because it would be a blanket implementation

## Additional context

I was specifically trying to animate UI color alpha when I ran into this, and I wonder if it's worth considering if there should be easy ways to get a `&mut` to a `Color`'s alpha and also possibly animate individual fields of a component's fields (e.g. animating `Transform::translation::x`), but that's a separate issue and probably a little more complicated.

wolf-in-space:
# Objective

Allow tuple structs in the animated_field macro.
-  for example `animated_field!(MyTupleStruct::0)`.

Fixes #16736 
- This issue was partially fixed in #16747, where support for tuple structs was added to `AnimatedField::new_unchecked`.

## Solution

Change the designator for `$field` in the macro from `ident` to `tt`.

## Testing

Expanded the doc test on `animated_field!` to include an example with a tuple struct.

### DIFF
--- a/crates/bevy_animation/src/animation_curves.rs
+++ b/crates/bevy_animation/src/animation_curves.rs
@@ -984,6 +984,7 @@ where
 ///
 /// ```
 /// # use bevy_animation::{animation_curves::AnimatedField, animated_field};
+/// # use bevy_color::Srgba;
 /// # use bevy_ecs::component::Component;
 /// # use bevy_math::Vec3;
 /// # use bevy_reflect::Reflect;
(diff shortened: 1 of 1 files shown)


BenjaminBrienen (approved):

alice-i-cecile (approved):

alice-i-cecile:
Thanks :) Macros really do feel arcane to me some days...

PR #17235, opened 2025-01-08
Rename `DefaultCameraView`

ickshonpe:
# Objective

The name `DefaultCameraView` is confusing and misleading:
* It isn't the default UI camera, which is either the camera with the `IsDefaultUiCamera` marker component or, if no such camera is found, the camera with the highest order which has the primary window as its render target.
* It doesn't make sense to call it a "default", every active 2d and 3d camera is given its own `DefaultCameraView`.
* The name doesn't make it clear that it's UI specific component.

## Solution

Rename `DefaultCameraView` to `UiCameraView`, add a doc comment for it and rename a few other fields and variables.

## Migration Guide

`DefaultCameraView` has been renamed to `UiCameraView`

### DIFF
--- a/crates/bevy_ui/src/render/render_pass.rs
+++ b/crates/bevy_ui/src/render/render_pass.rs
@@ -1,7 +1,7 @@
 use core::ops::Range;
 
 use super::{ImageNodeBindGroups, UiBatch, UiMeta};
-use crate::DefaultCameraView;
+use crate::UiCameraView;
 use bevy_ecs::{
     prelude::*,
     system::{lifetimeless::*, SystemParamItem},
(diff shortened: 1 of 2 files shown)


crates/bevy_ui/src/render/render_pass.rs
```diff
@@ -56,19 +57,20 @@ impl Node for UiPassNode {
...
-        let Ok((target, camera)) = self.ui_view_query.get_manual(world, input_view_entity) else {
+        let Ok((target, camera)) = self.camera_view_query.get_manual(world, input_view_entity)
+        else {
             return Ok(());
         };
         if transparent_phase.items.is_empty() {
             return Ok(());
         }
 
-        // use the "default" view entity if it is defined
```
mnmaita, line 66: The quoting here makes a bit evident the need for a better naming 😅

mnmaita (approved):

mnmaita:
Oh, btw @ickshonpe is this change breaking? Since this is a `pub` struct it might be but please correct me if I'm wrong. If so, it would be great to have the Migration section in the PR description.

alice-i-cecile (approved):

PR #17239, opened 2025-01-08
Rename animated fox examples to better communicate their purpose

### LINKED ISSUE #17192: Rename animated fox examples to better communicate their purpose
> I like the split a lot, but I think we should rename these examples to not refer to foxes. That's an implementation detail that doesn't matter to learners :)

_Originally posted by @alice-i-cecile in https://github.com/bevyengine/bevy/pull/17191#pullrequestreview-2532611157_

greeble-dev:
Fixes #17192.

Replaces "animated_fox" with "animated_mesh".

I considered a few different names - should it say "skinned_mesh" to be precise? Should it mention gltf? But "animated_mesh" seems intuitive and keeps it short.

## Testing

- Ran all three examples (Windows 10).


alice-i-cecile (approved):
Looks correct to me, thanks :)

🫡 🦊

greeble-dev:
Oh, forgot to mention that I didn't change `examples/stress_tests/many_foxes`. I figured that's more of an internal engine benchmark and not something beginners will look at, so a more precise name is ok. And honestly I just think it's kinda cute. ❤️ 🦊 🦊 🦊

BenjaminBrienen (approved):
Perfect

PR #17242, opened 2025-01-08
bevy_input: Apply `#![deny(clippy::allow_attributes, clippy::allow_attributes_without_reason)]`

LikeLakers2:
# Objective

## Solution
Set the `clippy::allow_attributes` and `clippy::allow_attributes_without_reason` lints to `deny`, and bring `bevy_input` in line with the new restrictions.

## Testing
`cargo clippy --tests --package bevy_input` was run, and no errors were encountered.

### DIFF
--- a/crates/bevy_input/src/keyboard.rs
+++ b/crates/bevy_input/src/keyboard.rs
@@ -238,7 +238,10 @@ pub enum NativeKeyCode {
     all(feature = "serialize", feature = "bevy_reflect"),
     reflect(Serialize, Deserialize)
 )]
-#[allow(clippy::doc_markdown)] // Clippy doesn't like our use of <kbd>.
+#[expect(
+    clippy::doc_markdown,
+    reason = "We use camel-case words inside `<kbd>` tags to represent keyboard keys, which are not identifiers that we should be putting inside backticks."
+)]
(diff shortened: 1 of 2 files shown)


alice-i-cecile (approved):

crates/bevy_input/src/keyboard.rs
```diff
@@ -764,7 +767,10 @@ pub enum NativeKey {
     all(feature = "serialize", feature = "bevy_reflect"),
     reflect(Serialize, Deserialize)
 )]
-#[allow(clippy::doc_markdown)] // Clippy doesn't like our use of <kbd>.
+#[expect(
+    clippy::doc_markdown,
+    reason = "We use camel-case words inside `<kbd>` tags to represent keyboard keys, which are not identifiers that we should be putting inside backticks."
```
BenjaminBrienen, line 772: This isn't a reason. It reads more like a TODO. I suggest implementing the fix with backticks.

LikeLakers2: It's not though - `<kbd>something</kbd>` and <code>\`something\`</code> look very different. As an example, here's `bevy_input::keyboard::Key::Alt`'s documentation, which uses both:

image

PR #17244, opened 2025-01-08
Improved UI camera mapping

ickshonpe:
# Objective

Two more optimisations for UI extraction:
* We only need to query for the camera's render entity when the target camera changes. If the target camera is the same as for the previous UI node we can use the previous render entity.
* The cheap checks for visibility and zero size should be performed first before the camera queries.

## Solution
Add a new system param `UiCameraMap` that resolves the correct render camera entity and only queries when necessary.

tracee

I don't like the `UiCameraMap` + `UiCameraMapper` implementation very much, maybe someone else can suggest a better construction.

This is partly motivated by #16942 which adds further indirection and these changes would ameliorate that performance regression.


pcwalton (approved):
Seems fine. I'm not intimately familiar with the details of the UI code, so someone else should review this in addition to me.

BenjaminBrienen (approved):
Good idea

alice-i-cecile (approved):

alice-i-cecile:
@ickshonpe looks great; ping me when merge conflicts are resolved.

ickshonpe:
@alice-i-cecile ready now

PR #17247, opened 2025-01-08
`add_edges` and `add_grid` helpers for `DirectionalNavMap`

alice-i-cecile:
# Objective

Prompted by #17224, I realized that we need more helpers to make configuring `DirectionalNavMap`s for directional (gamepad) navigation easier.

## Solution

- Add a simple `add_edges` helper, which does not loop around the end of the list.
- Add a `add_grid` helper for working with grids.
  - Obviously, this is really helpful for wiring up e.g. Terraria-style inventory screens.
  - Less obviously, this is a powerful mid-level API for defining directional nav layouts more broadly. The coordinates don't need to be physical, and the ability to duplicate entities in the map makes modelling large objects a breeze.
- Added more helpers to CompassQuadrant / CompassOctant to make this code prettier: these simple methods should be generally useful to users.

## Testing

I'm currently relying on unit tests for these changes. Once #17224 is merged, we can update that the example to test these in a more hands-on way.

## To do

- [ ] make sure tests pass and logic is correct

### DIFF
--- a/crates/bevy_input_focus/src/directional_navigation.rs
+++ b/crates/bevy_input_focus/src/directional_navigation.rs
@@ -25,6 +25,7 @@ use bevy_math::{CompassOctant, IVec2};
 #[cfg(feature = "bevy_reflect")]
 use bevy_reflect::{prelude::*, Reflect};
 use bevy_utils::{hashbrown, HashMap};
+use core::ops::Deref;
 use thiserror::Error;
 
 use crate::InputFocus;
@@ -402,7 +403,7 @@ impl Default for NavGrid {
     }
 }
 
-impl core::ops::Deref for NavGrid {
+impl Deref for NavGrid {
     type Target = HashMap<IVec2, Entity>;
 
     fn deref(&self) -> &Self::Target {
--- a/crates/bevy_input_focus/src/directional_navigation.rs
+++ b/crates/bevy_input_focus/src/directional_navigation.rs
@@ -513,6 +513,25 @@ mod tests {
     }
 
     #[test]
+    fn edges() {
+        let mut world = World::new();
+        let a = world.spawn_empty().id();
+        let b = world.spawn_empty().id();
+        let c = world.spawn_empty().id();
+
+        let mut map = DirectionalNavigationMap::default();
+        map.add_edges(&[a, b, c], CompassOctant::East);
+
+        assert_eq!(map.get_neighbor(a, CompassOctant::East), Some(b));
+        assert_eq!(map.get_neighbor(b, CompassOctant::East), Some(c));
+        assert_eq!(map.get_neighbor(c, CompassOctant::East), None);
+
+        assert_eq!(map.get_neighbor(a, CompassOctant::West), None);
+        assert_eq!(map.get_neighbor(b, CompassOctant::West), Some(a));
+        assert_eq!(map.get_neighbor(c, CompassOctant::West), Some(b));
+    }
+
+    #[test]
     fn looping_edges() {
         let mut world = World::new();
         let a = world.spawn_empty().id();
--- a/crates/bevy_input_focus/src/directional_navigation.rs
+++ b/crates/bevy_input_focus/src/directional_navigation.rs
@@ -203,8 +203,8 @@ impl DirectionalNavigationMap {
     /// Missing entities will be ignored, so sparse grids are supported.
     /// If an element is missing, their neighbors will be connected to the nearest existing entity in the given direction.
     ///
-    /// Diagonals travel in a purely 1:1 step fashion, so an entity at (0, 0) will connect to an entity at (2, 2) in the SouthEast direction,
-    /// but would never see an entity at (1, 2).
+    /// Diagonals travel in a purely 1:1 step fashion, so an entity at (0, 0) will connect to an entity at (2, 2)
+    /// in the [`SouthEast`](CompassOctant::SouthEast) direction, but would never see an entity at (1, 2).
     /// That said, you can add the same entity at multiple locations in the grid!
     /// This works well to model larger entities,
     /// even if the entities aren't truly grid-aligned.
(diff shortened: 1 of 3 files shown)


crates/bevy_input_focus/src/directional_navigation.rs
```diff
@@ -173,7 +175,18 @@ impl DirectionalNavigationMap {
         self.add_edge(b, a, direction.opposite());
     }
 
-    /// Add symmetrical edges between all entities in the provided slice, looping back to the first entity at the end.
+    /// Add symettrical edges between each consecutive pair of entities in the provided slice.
```
IQuick143, line 178:
```suggestion
    /// Add symmetrical edges between each consecutive pair of entities in the provided slice.
```

crates/bevy_input_focus/src/directional_navigation.rs
```diff
@@ -184,6 +197,168 @@ impl DirectionalNavigationMap {
...
     }
 
+    /// Adds edges to a grid of entities, connecting each entity to its neighbors in each of the eight directions.
+    ///
+    /// The grid is provided as a hashmap of [`IVec2`] to entities,
+    /// where the `x` represents columns from the top and `y` represents rows from the left.
+    /// Missing entities will be ignored, so sparse grids are supported.
+    /// If an element is missing, their neighbors will be connected to the nearest existing entity in the given direction.
+    ///
+    /// Diagonals travel in a purely 1:1 step fashion, so an entity at (0, 0) will connect to an entity at (2, 2)
```
IQuick143, line 207:
```suggestion
    /// Diagonals travel in a purely 1:1 step fashion, so an entity at (0, 0) will connect to an entity at (2, 2) if an entity is not present at (1, 1)
```

crates/bevy_input_focus/src/directional_navigation.rs
```diff
@@ -184,6 +197,168 @@ impl DirectionalNavigationMap {
...
+    /// That said, you can add the same entity at multiple locations in the grid!
+    /// This works well to model larger entities,
+    /// even if the entities aren't truly grid-aligned.
+    ///
+    /// If `should_loop` is set to `true`, the top and bottom rows will be connected, as will the left and right columns.
+    // The keys are specified as i32 for compatibility with taffy's grid system,
+    // which uses u16 values. We need signed integers here, but we should be able to cast them losslessly.
+    // PERF: This is a very flexible / "clever" implementation, but it won't be the fastest for simple cases.
+    // If there's user demand for it, we can add a more optimized / less flexible version that requires non-sparse rectangular grids.
+    pub fn add_grid(&mut self, entity_grid: NavGrid, should_loop: bool) {
```
IQuick143, line 218: This promises support for placing entities at multiple locations, however it's not well defined what the result is.
Consider the following grid:
```
X A
Y A
```
Now... what is the left neighbour of A? It seems to be completely implementation dependent.

crates/bevy_input_focus/src/directional_navigation.rs
```diff
@@ -184,6 +197,168 @@ impl DirectionalNavigationMap {
...
+                        position,
+                    )
+                };
+
+                if let Some(neighbor) = maybe_neighbor {
+                    // Entities should not be neighbors of themselves
+                    // Inserting the same entity at multiple locations is supported,
+                    // so we need to check for this case
+                    if neighbor != entity {
+                        // PERF: we could also add the reverse edge here, to save work later
```
IQuick143, line 267: Multiple positions for the same node would make this very tricky/not possible.

crates/bevy_input_focus/src/directional_navigation.rs
```diff
@@ -184,6 +197,168 @@ impl DirectionalNavigationMap {
...
+                        max_rows,
+                        max_columns,
+                        direction,
+                        position,
+                    )
+                };
+
+                if let Some(neighbor) = maybe_neighbor {
+                    // Entities should not be neighbors of themselves
+                    // Inserting the same entity at multiple locations is supported,
```
IQuick143, line 264: Note: This collision can happen also by just an entity looping to itself, which may or (probably) may not be desired.

crates/bevy_input_focus/src/directional_navigation.rs
alice-i-cecile: Yeah, we should probably choose, document and test a rule for this.

crates/bevy_input_focus/src/directional_navigation.rs
alice-i-cecile: Right 🤔 Okay, I'll cut the PERF note.

crates/bevy_math/src/compass.rs
```diff
@@ -114,6 +137,18 @@ pub enum CompassOctant {
 }
 
 impl CompassOctant {
+    /// The list of all possible [`CompassOctant`] variants.
+    pub const VARIANTS: [CompassOctant; 8] = [
```
IQuick143, line 141: Might be nice to test that this is in the same order as `from_index` (could we implement `from_index` like that?)

crates/bevy_input_focus/src/directional_navigation.rs
```diff
@@ -21,9 +21,11 @@ use bevy_ecs::{
     prelude::*,
```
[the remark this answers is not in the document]
cart: This is especially important because many lists of nodes will have different targets based on their current layout. We essentially _need_ those cases to automatically/implicitly adapt to the current context.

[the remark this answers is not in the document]
alice-i-cecile: > As a general thought, I think my ideal system is an automatic system that is overridden by manually defined nav targets.

Fully agree here! I've been exploring APIs and algorithms to make this possible :)

alice-i-cecile:
I think this design of add_grid is a bit too complex / fraught for now. I may revisit it if we want to come back to this later.

PR #17249, opened 2025-01-09
Allow `clippy::too_many_arguments` to lint without warnings

LikeLakers2:
# Objective
Many instances of `clippy::too_many_arguments` linting happen to be on systems - functions which we don't call manually, and thus there's not much reason to worry about the argument count.

## Solution
Allow `clippy::too_many_arguments` globally, and remove all lint attributes related to it.


BenjaminBrienen (approved):

rparrett (approved):
Looks good to me.

LikeLakers2:
I keep feeling like I went way too fast in submitting this PR. As a result, I'm not actually sure if this is what we want, or if we want to do something slightly different.

So I'm switching this to draft mode for now.

alice-i-cecile:
>  I think its easiest for the bevy community as a whole if we just encourage globally disabling that lint as a matter of course

- @cart, [on Discord](https://discord.com/channels/691052431525675048/692572690833473578/1326716428115509439).

I agree with this position too.

LikeLakers2:
Well, if two maintainers (one of them being the project lead) think that, then I think I'm going to un-draft this PR.

BenjaminBrienen (approved):

Issue #4906, opened 2022-06-03
WASM cannot build for wasm32-wasi


schtauffen:
## Bevy version

0.7

## What you did

Cargo.toml
```toml
<snip>

[target.'cfg(target_arch="wasm32")'.dependencies]
lunatic = "0.9"
bevy = { version = "0.7", default-features = false }
```
.cargo/config.toml
```toml
[target.wasm32-wasi]
runner = "lunatic"
```

ran `cargo run --bin server --target wasm32-wasi`

## What went wrong

I expect bevy to compile for use inside wasm. Instead, it gives errors:
```
error[E0463]: can't find crate for `js_sys`
 --> C:\Users\Zach and Andrea\.cargo\registry\src\github.com-1ecc6299db9ec823\cpal-0.13.5\src\host\webaudio\mod.rs:1:1
  |
1 | extern crate js_sys;
  | ^^^^^^^^^^^^^^^^^^^^ can't find crate

error[E0463]: can't find crate for `wasm_bindgen`
 --> C:\Users\Zach and Andrea\.cargo\registry\src\github.com-1ecc6299db9ec823\cpal-0.13.5\src\host\webaudio\mod.rs:2:1
  |
2 | extern crate wasm_bindgen;
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^ can't find crate

error[E0463]: can't find crate for `web_sys`
 --> C:\Users\Zach and Andrea\.cargo\registry\src\github.com-1ecc6299db9ec823\cpal-0.13.5\src\host\webaudio\mod.rs:3:1
  |
3 | extern crate web_sys;
  | ^^^^^^^^^^^^^^^^^^^^^ can't find crate

   Compiling image v0.23.14
error[E0433]: failed to resolve: use of undeclared type `AudioContextOptions`
   --> C:\Users\Zach and Andrea\.cargo\registry\src\github.com-1ecc6299db9ec823\cpal-0.13.5\src\host\webaudio\mod.rs:218:31    
    |
218 |         let mut stream_opts = AudioContextOptions::new();
    |                               ^^^^^^^^^^^^^^^^^^^ use of undeclared type `AudioContextOptions`

error[E0433]: failed to resolve: use of undeclared type `AudioContext`
   --> C:\Users\Zach and Andrea\.cargo\registry\src\github.com-1ecc6299db9ec823\cpal-0.13.5\src\host\webaudio\mod.rs:221:13    
    |
[...]

PR #17250, opened 2025-01-09
Create `bevy_platform_support` Crate

bushrat011899:
# Objective

- Contributes to #16877

## Solution

- Initial creation of `bevy_platform_support` crate.
- Moved `bevy_utils::Instant` into new `bevy_platform_support` crate.
- Moved `portable-atomic`, `portable-atomic-util`, and `critical-section` into new `bevy_platform_support` crate.

## Testing

- CI

---

## Showcase

Instead of needing code like this to import an `Arc`:

```rust
#[cfg(feature = "portable-atomic")]
use portable_atomic_util::Arc;

#[cfg(not(feature = "portable-atomic"))]
use alloc::sync::Arc;
```

We can now use:

```rust
use bevy_platform_support::sync::Arc;
```

This applies to many other types, but the goal is overall the same: allowing crates to use `std`-like types without the boilerplate of conditional compilation and platform-dependencies.

## Migration Guide

- Replace imports of `bevy_utils::Instant` with `bevy_platform_support::time::Instant`
- Replace imports of `bevy::utils::Instant` with `bevy::platform_support::time::Instant`

## Notes

- `bevy_platform_support` hasn't been reserved on `crates.io`
- ~~`bevy_platform_support` is not re-exported from `bevy` at this time. It may be worthwhile exporting this crate, but I am unsure of a reasonable name to export it under (`platform_support` may be a bit wordy for user-facing).~~
- I've included an implementation of `Instant` which is suitable for `no_std` platforms that are not Wasm for the sake of eliminating feature gates around its use. It may be a controversial inclusion, so I'm happy to remove it if required.
- There are many other items (`spin`, `bevy_utils::Sync(Unsafe)Cell`, etc.) which should be added to this crate. I have kept the initial scope small to demonstrate utility without making this too unwieldy.


alice-i-cecile:
> bevy_platform_support is not re-exported from bevy at this time. It may be worthwhile exporting this crate, but I am unsure of a reasonable name to export it under (platform_support may be a bit wordy for user-facing).

Just re-export it as is, the name is fine and this is less surprising.

bushrat011899:
> Just re-export it as is, the name is fine and this is less surprising.

Done!

alice-i-cecile:
Marking as blessed due to [this comment](https://discord.com/channels/691052431525675048/692572690833473578/1319137218312278077) by @cart. I agree with it too!

TimJentzsch (approved):
I don't really understand the fallback implementation for `Instant`, but the rest looks good to me

crates/bevy_platform_support/README.md
```diff
@@ -0,0 +1,56 @@
...
+These needs are handled by this crate, `bevy_platform_support`.
+The goal of this crate is to provide alternatives and extensions to the Rust standard library which minimize friction when developing with and for Bevy across multiple platforms.
+
+## Getting Started
+
+Like any dependency from [crates.io](https://crates.io/), include the following in you `Cargo.toml` `dependencies` table:
+
+```toml
+bevy_platform_support = "0.16.0-dev"
+```
```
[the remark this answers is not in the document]
bushrat011899: Yeah that's a good point. I've also replaced a second instance of `0.16.0-dev` with just `x.y.z` to make it more robust to version number changes.

bushrat011899:
> I don't really understand the fallback implementation for `Instant`, but the rest looks good to me

I should update the documentation to be more clear on how this works. Basically it's a straight copy of the `Instant` from `std`, but it includes two key changes:

1. `Instant::set_elapsed` is a new method that lets you tell `Instant` _how_ to know how much time has elapsed. Since we don't have an OS, we can't use the typical syscalls. A simpler alternative would be to just provide `Instant` with the elapsed time directly, rather than giving it a function to determine elapsed time. But the benefit here is now `Instant` can update itself only when it needs to, instead of the user needing to eagerly update it (e.g., in the mainloop)
2. `Instant::now` calls the method provided to `Instant::set_elapsed` to determine how much time has elapsed, instead of using a syscall directly like in `std`.

For the sake of convenience, I've included an `unset_getter` method, which is the default function called by `Instant::now` if you haven't set one yourself. If you're on `x86` or `aarch64`, we have access to a built-in instruction for getting a system tick in nanoseconds, so we can actually provide a working default implementation on those platforms. For all others, this method panics, explaining that you need to call `Instant::set_elapsed` before `Instant::now` is called.

crates/bevy_platform_support/Cargo.toml
```diff
@@ -0,0 +1,52 @@
...
+]
+
+alloc = ["portable-atomic-util?/alloc"]
+
+## `critical-section` provides the building blocks for synchronization primitives
+## on all platforms, including `no_std`.
+critical-section = ["dep:critical-section", "portable-atomic?/critical-section"]
+
+## `portable-atomic` provides additional platform support for atomic types and
+## operations, even on targets without native support.
```
BenjaminBrienen, line 33: Is "native support" the same as hardware support?

crates/bevy_reflect/src/func/dynamic_function.rs
```diff
@@ -92,8 +93,21 @@ impl<'env> DynamicFunction<'env> {
...
+        #[cfg(feature = "portable-atomic")]
+        #[expect(
+            unsafe_code,
+            reason = "unsized coercion is an unstable feature for non-std types"
+        )]
+        // SAFETY:
+        // - Coercion from `T` to `dyn for<'a> Fn(ArgList<'a>) -> FunctionResult<'a> + Send + Sync + 'env`
+        //   is valid as `T: for<'a> Fn(ArgList<'a>) -> FunctionResult<'a> + Send + Sync + 'env`
+        // - `Arc::from_raw` receives a valid pointer from a previous call to `Arc::into_raw`
+        let arc = unsafe { ArcFn::<'env>::from_raw(Arc::into_raw(arc) as *const _) };
```
BenjaminBrienen, line 107: I'm just going to assume that this incantation is correct 😅

crates/bevy_tasks/src/single_threaded_task_pool.rs
```diff
@@ -1,17 +1,12 @@
...
 
 #[cfg(feature = "std")]
 use std::thread_local;
 
-#[cfg(feature = "portable-atomic")]
-use portable_atomic_util::Arc;
-
-#[cfg(not(feature = "portable-atomic"))]
-use alloc::sync::Arc;
-
```
BenjaminBrienen, line 14: I like it

BenjaminBrienen (approved):
Pretty straightforward and well done!

crates/bevy_platform_support/Cargo.toml
bushrat011899: I believe so, there might be some nuance I'm unaware of, but I think the two terms could be interchangeable in this context.

crates/bevy_reflect/src/func/dynamic_function.rs
bushrat011899: Yeah there's a safe alternative to this where you first `Box` the item, do the coercion inside the `Box`, and then use `Arc::from(box)`. But that requires a second allocation, and isn't much easier to understand IMO.

crates/bevy_platform_support/src/time.rs
```diff
@@ -0,0 +1,192 @@
+//! Provides `Instant` for all platforms.
+
+pub use time::Instant;
+
+#[cfg(target_arch = "wasm32")]
+use web_time as time;
```
TimJentzsch, line 6: Hmmm, I'm actually not sure if this is "safe" to do without a feature flag or similar.
Unfortunately, we cannot assume that Wasm = Browser (although for Bevy apps its most likely true...) and this will only work in the browser.
In fact, AFAIK there is no safe way to detect if running in the browser

crates/bevy_platform_support/src/time.rs
```diff
@@ -0,0 +1,192 @@
...
+    /// which when called supplies a monotonically increasing count of elapsed nanoseconds relative
+    /// to some arbitrary point in time.
+    #[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+    pub struct Instant(Duration);
+
+    impl Instant {
+        /// Returns an instant corresponding to "now".
+        #[must_use]
+        pub fn now() -> Instant {
+            let getter = ELAPSED_GETTER.load(Ordering::Relaxed);
```
chescock, line 51: Are you sure you don't need `Ordering::Acquire` here and `Ordering::Release` below?  I can imagine someone writing a value to a global variable and then calling `set_elapsed` with a function that reads that variable.  Without synchronization, another thread could see the new function pointer without seeing the updates to the global variable.

crates/bevy_platform_support/src/time.rs
bushrat011899: This is true, but no worse than the status-quo of Bevy. Currently, Bevy assumes all `wasm32` targets are the browser. Bevy _should_ be refactored to have a `web` feature which enables JS interop, but that would require more substantial changes that I believe are better suited for a follow up intended to add support for WASI, the non-web `wasm` target that makes sense for Bevy (e.g., hosting a multiplayer server on a server less platform)

crates/bevy_platform_support/src/time.rs
bushrat011899: I reckon you're correct! Atomic `Ordering` is something I'm still learning.ill update this PR to use the orderings you've suggested here. Thanks!

crates/bevy_platform_support/src/time.rs
TimJentzsch: Ah I see, fair point!

TimJentzsch: Do you know if we already have a ticket regarding support for other wasm platforms?
Especially Wasi support sounds very promising, especially since it includes full support for `std` (or at least the file APIs) if I remember correctly.

bushrat011899: That would be #4906

mockersf:
could you add the new crate (in the correct order by depdendency) in https://github.com/bevyengine/bevy/blob/main/tools/publish.sh?

bushrat011899:
> could you add the new crate (in the correct order by depdendency) in https://github.com/bevyengine/bevy/blob/main/tools/publish.sh?

Done! `bevy_platform_support` is required by `bevy_utils`, so I've placed it at the very top of the list directly above it.

BenjaminBrienen (approved):

BD103:
> * `bevy_platform_support` hasn't been reserved on `crates.io`

Once this PR is merged, we should add the crate to [`bevy-crate-reservations`](https://github.com/bevyengine/bevy-crate-reservations).

cc @cart :)

cart:
Reserved!

PR #17258, opened 2025-01-09
Proportional scaling for the sprite's texture.

silvestrpredko:
# Objective

Bevy sprite image mode lacks proportional scaling for the underlying texture. In many cases, it's required. For example, if it is desired to support a wide variety of screens with a single texture, it's okay to cut off some portion of the original texture.

## Solution

I added scaling of the texture during the preparation step. To fill the sprite with the original texture, I scaled UV coordinates accordingly to the sprite size aspect ratio and texture size aspect ratio. To fit texture in a sprite the original `quad` is scaled and then the additional translation is applied to place the scaled quad properly.

## Testing

For testing purposes could be used `2d/sprite_scale.rs`. Also, I am thinking that it would be nice to have some tests for a `crates/bevy_sprite/src/render/mod.rs:sprite_scale`.

---

## Showcase

image


examples/2d/sprite_scale.rs
```diff
@@ -0,0 +1,139 @@
...
+    let font = asset_server.load("fonts/FiraSans-Bold.ttf");
+    let style = TextFont {
+        font: font.clone(),
+        ..default()
+    };
+
+    let square = asset_server.load("textures/slice_square_2.png");
+    let banner = asset_server.load("branding/banner.png");
+
+    let rects = vec![
```
tbillington, line 23: Ultra nit-picking, feel free to ignore xD 
- I don't think specifying the font is necessary any more, bevy has a default font built in
- Don't need a Vec here, it can just be an array `let rects = [ ... ];

silvestrpredko: I'll go ahead and sort this out as soon as other's review comments appear. I just want to gather more comments for further work.

crates/bevy_sprite/src/texture_slice/computed_slices.rs
```diff
@@ -120,8 +121,8 @@ fn compute_sprite_slices(
             };
             slice.tiled(*stretch_value, (*tile_x, *tile_y))
         }
-        SpriteImageMode::Auto => {
-            unreachable!("Slices should not be computed for SpriteImageMode::Stretch")
+        SpriteImageMode::Auto | SpriteImageMode::ScaleMode(_) => {
```
alice-i-cecile, line 124: This will be clearer for debugging if you keep them as two different match arms. Then you'll get a more precise warning if this ever breaks.

crates/bevy_sprite/src/sprite.rs
```diff
@@ -185,6 +188,49 @@ impl SpriteImageMode {
...
+    pub const fn scale(&self) -> Option<TextureScale> {
+        if let SpriteImageMode::ScaleMode(scale) = self {
+            Some(*scale)
+        } else {
+            None
+        }
+    }
+}
+
+/// Represents various modes for proportional scaling of a texture
```
alice-i-cecile, line 204:
```suggestion
/// Represents various modes for proportional scaling of a texture.
///
/// Can be used in [`SpriteImageMode::ScaleMode`].
```

crates/bevy_sprite/src/sprite.rs
```diff
@@ -162,6 +162,9 @@ pub enum SpriteImageMode {
     /// The sprite will take on the size of the image by default, and will be stretched or shrunk if [`Sprite::custom_size`] is set.
     #[default]
     Auto,
+    /// The texture will be scaled to fit the rect bounds defined in [`Sprite::custom_size`].
+    /// Otherwise no scaling will be applied.
+    ScaleMode(TextureScale),
```
alice-i-cecile, line 167: We shouldn't duplicate `Mode` here. Just `SpriteImageMode::Scale` is clearer here.

crates/bevy_sprite/src/render/mod.rs
```diff
@@ -874,3 +890,96 @@ impl<P: PhaseItem> RenderCommand<P> for DrawSpriteBatch {
         RenderCommandResult::Success
     }
 }
+
+/// Scales a texture to fit within a given quad size with keeping the aspect ratio.
+fn scale_texture(
```
alice-i-cecile, line 895: I think this makes more sense as a public method on `TextureScale`. This is generally useful logic, and I think there's a good chance we move this out of `bevy_sprite` in the future.

alice-i-cecile: `apply_scaling` feels like the right name to me.

crates/bevy_sprite/src/render/mod.rs
```diff
@@ -339,6 +339,7 @@ pub struct ExtractedSprite {
     /// For cases where additional [`ExtractedSprites`] are created during extraction, this stores the
     /// entity that caused that creation for use in determining visibility.
     pub original_entity: Option<Entity>,
+    pub texture_scale: Option<TextureScale>,
```
alice-i-cecile, line 342: `TextureScale` is a bit of a weird name. What about `ScalingMode`?

alice-i-cecile:
I like the feature, and the algorithms seem to be correct. Some mild style / docs / naming feedback, but do push back if you disagree with any of them.

Much of this feels like it belongs in `bevy_image`, but we should move that later. Being able to share this style of logic across UI/decals/sprites would be great.

alice-i-cecile (approved):

crates/bevy_sprite/src/render/mod.rs
```diff
@@ -875,3 +891,96 @@ impl<P: PhaseItem> RenderCommand<P> for DrawSpriteBatch {
...
+                *quad_translation = Vec2::new(0.0, offset.y);
+                *quad_size = new_quad;
+            } else {
+                let scale = Vec2::new(texture_ratio / quad_ratio, 1.0);
+                let new_quad = *quad_size * scale;
+                let offset = *quad_size - new_quad;
+                *quad_translation = Vec2::new(-offset.x, 0.0);
+                *quad_size = new_quad;
+            }
+        }
```
kristoff3r, line 984: These are identical except for 1 minus sign, and there are similar duplication in the other Start/End cases. On one hand it would be nice if the repetition could be factored out so e.g. FitEnd is just FitStart and set the minus. On the other hand this code is pretty simple and easy to read.

It might also become simpler if you match on `(scaling_mode, quad_ratio > texture_ratio)` since every case has that if inside.

kristoff3r (approved):

crates/bevy_sprite/src/sprite.rs
```diff
@@ -185,6 +188,51 @@ impl SpriteImageMode {
...
+#[derive(Debug, Clone, Copy, PartialEq, Default, Reflect)]
+#[reflect(Debug)]
+pub enum ScalingMode {
+    /// Scale the texture uniformly (maintain the texture's aspect ratio)
+    /// so that both dimensions (width and height) of the texture will be equal
+    /// to or larger than the corresponding dimension of the rect.
+    /// Fill rect with a centered texture.
+    #[default]
+    FillCenter,
+    /// Scale the texture to fill the rect with a start of the texture,
```
ickshonpe, line 216:
```suggestion
    /// Scale the texture to fill the rect with the start of the texture,
```

This isn't really very clear about what the "start" of the texture is.

crates/bevy_sprite/src/sprite.rs
```diff
@@ -185,6 +188,51 @@ impl SpriteImageMode {
...
+    /// maintaining the aspect ratio.
+    FillStart,
+    /// Scale the texture to fill the rect with a end of the texture,
```
ickshonpe, line 219:
```suggestion
    /// Scale the texture to fill the rect with the end of the texture,
```

crates/bevy_sprite/src/sprite.rs
```diff
@@ -185,6 +188,51 @@ impl SpriteImageMode {
...
+    /// maintaining the aspect ratio.
+    FillEnd,
+    /// Scaling the texture will maintain the original aspect ratio
+    /// and ensure that the original texture fits entirely inside the rect.
+    /// At least one axis (X or Y) will fit exactly. The result is centered inside the rect.
```
ickshonpe, line 224:
```suggestion
    /// At least one axis (x or y) will fit exactly. The result is centered inside the rect.
```

crates/bevy_sprite/src/sprite.rs
```diff
@@ -185,6 +188,51 @@ impl SpriteImageMode {
...
+    FitCenter,
+    /// Scaling the texture will maintain the original aspect ratio
+    /// and ensure that the original texture fits entirely inside rect.
+    /// At least one axis (X or Y) will fit exactly.
```
ickshonpe, line 228:
```suggestion
    /// At least one axis (x or y) will fit exactly.
```

crates/bevy_sprite/src/sprite.rs
```diff
@@ -185,6 +188,51 @@ impl SpriteImageMode {
...
+    /// Aligns the result to the left and top edges of rect.
+    FitStart,
+    /// Scaling the texture will maintain the original aspect ratio
+    /// and ensure that the original texture fits entirely inside rect.
+    /// At least one axis (X or Y) will fit exactly.
```
ickshonpe, line 233:
```suggestion
    /// At least one axis (x or y) will fit exactly.
```

ickshonpe (changes requested):

examples/2d/sprite_scale.rs
```diff
@@ -0,0 +1,133 @@
...
+        .add_systems(Startup, setup)
+        .run();
+}
+
+fn setup(mut commands: Commands, asset_server: Res<AssetServer>) {
+    commands.spawn(Camera2d);
+
+    let square = asset_server.load("textures/slice_square_2.png");
+    let banner = asset_server.load("branding/banner.png");
+
```
ickshonpe, line 17: There should be some example cases with a sprite from a texture atlas.

crates/bevy_ui/src/render/ui_texture_slice_pipeline.rs
```diff
@@ -813,7 +813,10 @@ fn compute_texture_slices(
             [[0., 0., 1., 1.], [0., 0., 1., 1.], [1., 1., rx, ry]]
         }
         SpriteImageMode::Auto => {
-            unreachable!("Slices should not be computed for ImageScaleMode::Stretch")
+            unreachable!("Slices should not be computed for SpriteImageMode::Stretch")
```
ickshonpe, line 816:
```suggestion
            unreachable!("Slices can not be computed for SpriteImageMode::Auto")
```

crates/bevy_ui/src/render/ui_texture_slice_pipeline.rs
```diff
@@ -813,7 +813,10 @@ fn compute_texture_slices(
...
+        }
+        SpriteImageMode::Scale(_) => {
+            unreachable!("Slices should not be computed for SpriteImageMode::Scale")
```
ickshonpe, line 819:
```suggestion
            unreachable!("Slices can not be computed for SpriteImageMode::Scale")
```

ickshonpe (changes requested):

crates/bevy_sprite/src/render/mod.rs
```diff
@@ -430,6 +431,7 @@ pub fn extract_sprites(
                     image_handle_id: sprite.image.id(),
                     anchor: sprite.anchor.as_vec(),
                     original_entity: Some(original_entity),
+                    scaling_mode: sprite.image_mode.scale(),
```
ickshonpe, line 434: I'd prefer it if the sprite geometry was calculated during extraction rather than in prepare.

crates/bevy_sprite/src/sprite.rs
```diff
@@ -185,6 +188,51 @@ impl SpriteImageMode {
...
+
+/// Represents various modes for proportional scaling of a texture.
+///
+/// Can be used in [`SpriteImageMode::Scale`].
+#[derive(Debug, Clone, Copy, PartialEq, Default, Reflect)]
+#[reflect(Debug)]
+pub enum ScalingMode {
+    /// Scale the texture uniformly (maintain the texture's aspect ratio)
+    /// so that both dimensions (width and height) of the texture will be equal
+    /// to or larger than the corresponding dimension of the rect.
```
ickshonpe, line 212: I guess you mean the `rect` field of the sprite? Needs to be clearer I think.

ickshonpe (changes requested):

crates/bevy_sprite/src/sprite.rs
```diff
@@ -185,6 +188,51 @@ impl SpriteImageMode {
...
+    /// maintaining the aspect ratio.
+    FillEnd,
+    /// Scaling the texture will maintain the original aspect ratio
+    /// and ensure that the original texture fits entirely inside the rect.
+    /// At least one axis (X or Y) will fit exactly. The result is centered inside the rect.
+    FitCenter,
+    /// Scaling the texture will maintain the original aspect ratio
+    /// and ensure that the original texture fits entirely inside rect.
+    /// At least one axis (X or Y) will fit exactly.
+    /// Aligns the result to the left and top edges of rect.
```
ickshonpe, line 229: What about top and right? This seems incomplete like it needs some per-axis options or something.

ickshonpe (changes requested):
This seems like a good start but looks like it still needs some work. It took me a while to understand the API and the docs aren't very clear. It probably doesn't make much that much impact but I'd like to see some sprite benchmarks as well. 

Maybe there needs to be some sort of scale with overflow option as well, but I'm not really sure and that could be added in a followup.

crates/bevy_sprite/src/render/mod.rs
ickshonpe: Maybe this API would be better with two enums `Fill/Fit` and `Start/Center/End`.

crates/bevy_sprite/src/sprite.rs
```diff
@@ -185,6 +188,51 @@ impl SpriteImageMode {
...
+        }
+    }
+}
+
+/// Represents various modes for proportional scaling of a texture.
+///
+/// Can be used in [`SpriteImageMode::Scale`].
+#[derive(Debug, Clone, Copy, PartialEq, Default, Reflect)]
+#[reflect(Debug)]
+pub enum ScalingMode {
```
ickshonpe, line 209: I think maybe this type might be better split up into different two enums, one `Fill/Fit` and the other `Start/Center/End`.

crates/bevy_sprite/src/render/mod.rs
silvestrpredko: I understand your concern about code duplication, and I agree that maintaining clarity is important. However, considering the minimal amount of code duplication we currently have, I believe it's reasonable to keep things as they are. The existing structure seems to be functioning well without introducing unnecessary complexity.

examples/2d/sprite_scale.rs
silvestrpredko: > There should be some example cases with a sprite from a texture atlas.

It's important to note that the sprite sheet won't function correctly with scaling. Scaling completely overrides both the scaling and offset for the sprite, disregarding the `rect`.

crates/bevy_sprite/src/render/mod.rs
silvestrpredko: All the geometry logic happens in [`prepare`](https://github.com/bevyengine/bevy/pull/17258/files#diff-77918ecd6990bcb870a2d09f505f056ac01f54ce72fba80d692acd0484153810R776-R798) for now. I am going with less resistance and fewer changes in the codebase to achieve the necessary functionality.

crates/bevy_sprite/src/sprite.rs
silvestrpredko: > I think maybe this type might be better split up into different two enums, one `Fill/Fit` and the other `Start/Center/End`.

I don't think that will be convenient for the end user. There are only six variants and no more are planned to add I guess. That led to the single field in Sprite that corresponds to scaling, adding one more enum adds more complexity.

Then `SpriteImageMode::Scale` needs to be a struct enum variant with two fields, one for `Fill/Fit` and the second one for direction `Start/Center/End`. In my humble opinion, it will not bring convenience for the current `API`.

examples/2d/sprite_scale.rs
silvestrpredko: I just updated the documentation to properly notify a user that it couldn't be used for a sprite sheet. For now, I don't know how to easily add additional scaling of the sprite sheet to the desired rect

silvestrpredko:
I've been considering how to effectively apply scaling to a specific rectangle within a texture. My approach involves summing the scaling factors and then appropriately adjusting the offsets to accommodate the various scaling options. This way, I can ensure that the scaling is applied uniformly and maintains the correct alignment within the texture.

Please take a look. @alice-i-cecile @ickshonpe @BenjaminBrienen


https://github.com/user-attachments/assets/36765e07-ff25-4899-8728-a5542d8a3aa9

silvestrpredko:
@ickshonpe just a gentle ping to review the latest changes.

Issue #11429, opened 2024-01-20
Set a max-history-duration for Diagnostics


cBournhonesque:
## What problem does this solve or what need does it fill?

I feel like with diagnostics, the value I often want to know is:
'what was the average value over the last X seconds' 
(for example what was the bandwidth over the last 1 second, or what was the fps averaged over the last 3 seconds).

The current diagnostics requires you to specify a `max_history_length` which will be the number of values stored in the buffer.
But personally I don't particulary care for the 'bandwidth over the last 64 measurements`, i'm more interested knowing in knowing the value averaged over a specific amount of time.

## What solution would you like?

The API could accept a parameter like
```
pub enum BufferSize {
   HistoryLength(u32),
   HistoryDuration(Duration)
}
```

PR #17259, opened 2025-01-09
Allow users to customize history length in `FrameTimeDiagnosticsPlugin`

### LINKED ISSUE #10948: with_smoothing_factor is currently kind of useless
trying to apply with_smoothing_factor to the default fps counter from `FrameTimeDiagnosticsPlugin` doesn't actually do anything (and seems impossible to do in a way that it would mutate that counter), you'd have to create your own, different fps counter with a custom smoothing factor to be able to use it. I'm not sure if this was intended. I'm looking at numbers changing so fast between 240 and 290 fps that I'm not even sure that's the correct range.

rparrett:
# Objective

I have an application where I'd like to measure average frame rate over the entire life of the application, and it would be handy if I could just configure this on the existing `FrameTimeDiagnosticsPlugin`.

Probably fixes #10948?

## Solution

Add `max_history_length` to `FrameTimeDiagnosticsPlugin`, and because `smoothing_factor` seems to be based on history length, add that too.

## Discussion

I'm not totally sure that `DEFAULT_MAX_HISTORY_LENGTH` is a great default for `FrameTimeDiagnosticsPlugin` (or any diagnostic?). That's 1/3 of a second at typical game frame rates. Moreover, the default print interval for `LogDiagnosticsPlugin` is 1 second. So when the two are combined, you are printing the average over the last third of the duration between now and the previous print, which seems a bit wonky. (related: #11429)

I'm pretty sure this default value discussed and the current value wasn't totally arbitrary though.

Maybe it would be nice for `Diagnostic` to have a `with_max_history_length_and_also_calculate_a_good_default_smoothing_factor` method? And then make an explicit smoothing factor in `FrameTimeDiagnosticsPlugin` optional?

Or add a `new(max_history_length: usize)` method to `FrameTimeDiagnosticsPlugin` that sets a reasonable default `smoothing_factor`? edit: This one seems like a no-brainer, doing it.

## Alternatives

It's really easy to roll your own `FrameTimeDiagnosticsPlugin`, but that might not be super interoperable with, for example, third party FPS overlays. Still, might be the right call.

## Testing

`cargo run --example many_sprites` (modified to use a custom `max_history_length`)

## Migration Guide

`FrameTimeDiagnosticsPlugin` now contains two fields. Use `FrameTimeDiagnosticsPlugin::default()` to match Bevy's previous behavior or, for example, `FrameTimeDiagnosticsPlugin::new(60)` to configure it.

### DIFF
--- a/crates/bevy_dev_tools/src/fps_overlay.rs
+++ b/crates/bevy_dev_tools/src/fps_overlay.rs
@@ -42,7 +42,7 @@ impl Plugin for FpsOverlayPlugin {
     fn build(&self, app: &mut bevy_app::App) {
         // TODO: Use plugin dependencies, see https://github.com/bevyengine/bevy/issues/69
         if !app.is_plugin_added::<FrameTimeDiagnosticsPlugin>() {
-            app.add_plugins(FrameTimeDiagnosticsPlugin);
+            app.add_plugins(FrameTimeDiagnosticsPlugin::default());
         }
         app.insert_resource(self.config.clone())
             .add_systems(Startup, setup)
--- a/crates/bevy_diagnostic/src/frame_time_diagnostics_plugin.rs
+++ b/crates/bevy_diagnostic/src/frame_time_diagnostics_plugin.rs
@@ -23,7 +23,7 @@ impl Default for FrameTimeDiagnosticsPlugin {
     }
 }
 impl FrameTimeDiagnosticsPlugin {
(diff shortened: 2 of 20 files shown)


alice-i-cecile (approved):

alice-i-cecile:
I think that DEFAULT_MAX_HISTORY_LENGTH is a bad default, but that change should be split out IMO. The existing changes are nicely uncontroversial :)

BenjaminBrienen (approved):

PR #17266, opened 2025-01-09
Rename `PickingBehavior` to `Pickable`

chompaa:
# Objective

PR #17225 allowed for sprite picking to be opt-in. After some discussion, it was agreed that `PickingBehavior` should be used to opt-in to sprite picking behavior for entities. This leads to `PickingBehavior` having two purposes: mark an entity for use in a backend, and describe how it should be picked. Discussion led to the name `Pickable`making more sense (also: this is what the component was named before upstreaming).

A follow-up pass will be made after this PR to unify backends.

## Solution

Replace all instances of `PickingBehavior` and `picking_behavior` with `Pickable` and `pickable`, respectively.

## Testing

CI

## Migration Guide

Change all instances of `PickingBehavior` to `Pickable`.

### DIFF
--- a/crates/bevy_dev_tools/src/picking_debug.rs
+++ b/crates/bevy_dev_tools/src/picking_debug.rs
@@ -293,7 +293,7 @@ pub fn debug_draw(
                         ..Default::default()
                     },
                 ))
-                .insert(PickingBehavior::IGNORE)
+                .insert(Pickable::IGNORE)
                 .insert(TargetCamera(camera));
         }
     }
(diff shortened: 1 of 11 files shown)


alice-i-cecile (approved):
I agree with this. Once the backends are unified in behavior, we should add docs to explain that `Pickable` is always required.

BenjaminBrienen (approved):

aevyrie:
> I agree with this. Once the backends are unified in behavior, we should add docs to explain that Pickable is always required

@alice-i-cecile is that correct? I thought we wanted both opt in and opt out? E.g. in opt-out, `Pickable` is not required.

PR #17269, opened 2025-01-09
Adds an example for an extended material that uses a custom vertex shader 

### LINKED ISSUE #16970: Need an example of a simple material extension as a vertex shader

I would like a nice simple example of an ExtensionMaterial<standardmaterial> that uses a custom wgsl shader of just a vertex shader, perhaps that accepts a prop for Velocity and a gradient texture . 

In this way , the vertices (mapped by uv map) will be translated by the (velocity x gradient texture) 

For simulating grass, hair, etc 

Other solutions to solve and/or work around the problem presented.

## Additional context

Any other information you would like to add such as related previous work,
screenshots, benchmarks, etc.

ethereumdegen:
# Objective

There was no example of an extended material that uses a custom vertex shader so i made one. 
 
 Someone please check over my example and make sure I dont have any unnecessary code in there.  Im not sure if i need the specialize() impl or if that is unnecessary  for example. 
 
 
Fixes #16970.

---

## Showcase
 
vert_shader


ethereumdegen:
this is for #16970 16970

examples/shader/extended_vertex_material.rs
```diff
@@ -0,0 +1,125 @@
...
+    pbr::{ExtendedMaterial, MaterialExtension, OpaqueRendererMethod},
+    prelude::*,
+    render::render_resource::*,
+};
+use bevy_render::mesh::MeshVertexAttribute;
+use bevy_render::mesh::MeshVertexBufferLayoutRef;
+
+/// This example uses a shader source file from the assets subdirectory
+//const SHADER_ASSET_PATH: &str = "shaders/extended_material.wgsl";
+
```
LikeLakers2, line 16: CI's saying something about a empty line after a doc-comment here.
```suggestion
```

hukasu:
@ethereumdegen do you intend on updating this example and fixing the errors reported by CI? or can this be put for adoption?

IceSentry:
I'm not convinced this is a necessary example. Custom vertex shaders could be showcased with a normal material and the concepts will be the same for extracted materials. The code has also diverged since the original PR.

PR #17277, opened 2025-01-10
bevy_pbr: Apply `#![deny(clippy::allow_attributes, clippy::allow_attributes_without_reason)]`

LikeLakers2:
# Objective

## Solution
Set the `clippy::allow_attributes` and `clippy::allow_attributes_without_reason` lints to `deny`, and bring `bevy_pbr` in line with the new restrictions.

## Testing
`cargo clippy --tests --package bevy_pbr` was run, and no errors were encountered.

### DIFF
--- a/crates/bevy_pbr/src/lib.rs
+++ b/crates/bevy_pbr/src/lib.rs
@@ -1,6 +1,11 @@
 #![expect(missing_docs, reason = "Not all docs are written yet, see #3492.")]
 #![cfg_attr(docsrs, feature(doc_auto_cfg))]
-#![deny(unsafe_code)]
+#![forbid(unsafe_code)]
+#![deny(
+    clippy::allow_attributes,
+    clippy::allow_attributes_without_reason,
+    reason = "See #17111; To be removed once all crates are in-line with these attributes"
+)]
 #![doc(
     html_logo_url = "https://bevyengine.org/assets/icon.png",
     html_favicon_url = "https://bevyengine.org/assets/icon.png"
--- a/crates/bevy_pbr/src/ssao/mod.rs
+++ b/crates/bevy_pbr/src/ssao/mod.rs
@@ -752,10 +752,13 @@ fn prepare_ssao_bind_groups(
     }
 }
 
-#[allow(clippy::needless_range_loop)]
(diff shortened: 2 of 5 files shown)


crates/bevy_pbr/src/ssao/mod.rs
```diff
@@ -752,10 +752,13 @@ fn prepare_ssao_bind_groups(
...
 }
 
-#[allow(clippy::needless_range_loop)]
 fn generate_hilbert_index_lut() -> [[u16; 64]; 64] {
     let mut t = [[0; 64]; 64];
 
+    #[expect(
+        clippy::needless_range_loop,
+        reason = "Rewriting this loop to use iterators would make the code much harder to read for very little performance gain."
+    )]
```
LikeLakers2, line 761: It *would* be possible to rewrite this iterator with clearer code, if `hilbert_index()` took `usize`s instead of `u16`s... but I consider that out-of-scope.

crates/bevy_pbr/src/lib.rs
```diff
@@ -1,6 +1,11 @@
 #![expect(missing_docs, reason = "Not all docs are written yet, see #3492.")]
 #![cfg_attr(docsrs, feature(doc_auto_cfg))]
-#![deny(unsafe_code)]
+#![forbid(unsafe_code)]
```
LikeLakers2, line 3: I searched through `bevy_pbr`, and found no unsafe code - so I figured this was meant to be a `forbid`.

ickshonpe (approved):
I wonder how all those `unused_variable` lints survived for so long, shows why this is a good change.

crates/bevy_pbr/src/ssao/mod.rs
BenjaminBrienen: Do you like this?

```rs
let t: [[u16; 64]; 64] = std::array::from_fn(|x| {
    std::array::from_fn(|y| hilbert_index(x as u16, y as u16))
});
```

BenjaminBrienen (approved):

BenjaminBrienen (approved):

alice-i-cecile (approved):

PR #17282, opened 2025-01-10
`custom_ui_material`  border fix

ickshonpe:
# Objective

The order of the border edges in `UiVertexOutput` is left, right, top, bottom but in `custom_ui_material` the selectors switch them so left is right and top is bottom.

## Solution

Reverse the conditions so that the correct border values are selected.

### DIFF
--- a/assets/shaders/custom_ui_material.wgsl
+++ b/assets/shaders/custom_ui_material.wgsl
@@ -12,8 +12,8 @@
 fn fragment(in: UiVertexOutput) -> @location(0) vec4<f32> {
     let r = in.uv - 0.5;
     let b = vec2(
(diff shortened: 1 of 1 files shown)


rparrett (approved):
Confirmed issue and fix by modifying the `ui_material` example.

```rust
border: UiRect {
    top: Val::Px(20.),
    left: Val::Px(20.),
    bottom: Val::Px(10.),
    right: Val::Px(10.),
},
```

Since you seem to understand this shader code, maybe you could sprinkle a few comments and/or expand the mysterious one letter variable names while you're in there?

alice-i-cecile (approved):

PR #17288, opened 2025-01-10
bevy_state: Apply `#![deny(clippy::allow_attributes, clippy::allow_attributes_without_reason)]`

LikeLakers2:
# Objective

## Solution
Set the `clippy::allow_attributes` and `clippy::allow_attributes_without_reason` lints to `deny`, and bring `bevy_state` in line with the new restrictions.

## Testing
Rust-analyzer did not return any errors once the deny was added.


alice-i-cecile (approved):
I am *so* looking forward to cleaning these up.

PR #17291, opened 2025-01-10
Create missing constructors for 2D primitive mesh builders

BD103:
# Objective

- While all [`MeshBuilder`](https://dev-docs.bevyengine.org/bevy/prelude/trait.MeshBuilder.html)s can be created by first creating the primitive `bevy_math` type and using [`Meshable`](https://dev-docs.bevyengine.org/bevy/prelude/trait.Meshable.html), most builders have their own `const` constructors that can be used instead. Some builders are missing constructors, however, making them unavailable in `const` contexts.

## Solution

- Add a `const` constructor for `RegularPolygonMeshBuilder`, `RhombusMeshBuilder`, `Triangle2dMeshBuilder`, and `RectangleMeshBuilder`.
- Add a note on the requirements of `ConvexPolygonMeshBuilder`, and recommend using `ConvexPolygon::new().mesh()` instead.
  - A constructor cannot easily be created for this type, since it requires duplicating all of `ConvexPolygon::new()`'s verification code. I may be able to work around this, but it requires touching a bit more code surface. Opinions?

## Testing

Not much beyond CI! The changes are trivial enough that only a cursory glance for typos and switched variables should be necessary.

## Note for Reviewers

Hi! I haven't directly used the types I modify in this PR beyond some benchmarking work I did this morning. If you're familiar with these types, please let me know if any of the constructors need additional validation (or if the constructors shouldn't be there at all). Thanks!

### DIFF
--- a/crates/bevy_mesh/src/primitives/dim2.rs
+++ b/crates/bevy_mesh/src/primitives/dim2.rs
@@ -754,7 +754,10 @@ pub struct RhombusMeshBuilder {
 impl RhombusMeshBuilder {
     /// Creates a new [`RhombusMeshBuilder`] from a horizontal and vertical diagonal size.
     pub const fn new(horizontal_diagonal: f32, vertical_diagonal: f32) -> Self {
-        debug_assert!(horizontal_diagonal >= 0.0, "rhombus has a negative horizontal");
+        debug_assert!(
+            horizontal_diagonal >= 0.0,
+            "rhombus has a negative horizontal",
+        );
         debug_assert!(vertical_diagonal >= 0.0, "rhombus has a negative vertical");
 
         Self {
--- a/crates/bevy_mesh/src/primitives/dim2.rs
+++ b/crates/bevy_mesh/src/primitives/dim2.rs
@@ -462,7 +462,7 @@ impl RegularPolygonMeshBuilder {
     ///
     /// # Panics
     ///
-    /// Panics if `circumradius` is negative, or if `sides` is less than 3.
+    /// Panics in debug mode if `circumradius` is negative, or if `sides` is less than 3.
     pub const fn new(circumradius: f32, sides: u32) -> Self {
         debug_assert!(
             circumradius.is_sign_positive(),
@@ -753,6 +753,10 @@ pub struct RhombusMeshBuilder {
 
 impl RhombusMeshBuilder {
     /// Creates a new [`RhombusMeshBuilder`] from a horizontal and vertical diagonal size.
+    /// 
+    /// # Panics
+    /// 
+    /// Panics in debug mode if `horizontal_diagonal` or `vertical_diagonal` is negative.
     pub const fn new(horizontal_diagonal: f32, vertical_diagonal: f32) -> Self {
         debug_assert!(
             horizontal_diagonal >= 0.0,
@@ -896,6 +900,10 @@ pub struct RectangleMeshBuilder {
 
 impl RectangleMeshBuilder {
     /// Creates a new [`RectangleMeshBuilder`] from a full width and height.
+    /// 
+    /// # Panics
+    /// 
+    /// Panics in debug mode if `width` or `height` is negative.
     pub const fn new(width: f32, height: f32) -> Self {
         debug_assert!(width >= 0.0, "rectangle has a negative width");
         debug_assert!(height >= 0.0, "rectangle has a negative height");
--- a/crates/bevy_mesh/src/primitives/dim2.rs
+++ b/crates/bevy_mesh/src/primitives/dim2.rs
@@ -464,11 +464,11 @@ impl RegularPolygonMeshBuilder {
     ///
     /// Panics if `circumradius` is negative, or if `sides` is less than 3.
     pub const fn new(circumradius: f32, sides: u32) -> Self {
-        assert!(
+        debug_assert!(
             circumradius.is_sign_positive(),
             "polygon has a negative radius"
         );
-        assert!(sides > 2, "polygon has less than 3 sides");
+        debug_assert!(sides > 2, "polygon has less than 3 sides");
 
         Self {
             circumradius,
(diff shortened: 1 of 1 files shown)


alice-i-cecile (approved):

BD103:
Looks like the CI failure is unrelated (maybe with the recent Rust release?). I ran `cargo clippy -p bevy_mesh` locally and it passed!

crates/bevy_mesh/src/primitives/dim2.rs
```diff
@@ -452,6 +455,28 @@ pub struct RegularPolygonMeshBuilder {
...
     sides: u32,
 }
+
+impl RegularPolygonMeshBuilder {
+    /// Creates a new [`RegularPolygonMeshBuilder`] from the radius of a circumcircle and a number
+    /// of sides.
+    ///
+    /// # Panics
+    ///
+    /// Panics if `circumradius` is negative, or if `sides` is less than 3.
```
IQuick143, line 465: How is +-0 handled?

crates/bevy_mesh/src/primitives/dim2.rs
```diff
@@ -726,6 +751,15 @@ pub struct RhombusMeshBuilder {
     half_diagonals: Vec2,
 }
 
+impl RhombusMeshBuilder {
+    /// Creates a new [`RhombusMeshBuilder`] from a horizontal and vertical diagonal size.
+    pub const fn new(horizontal_diagonal: f32, vertical_diagonal: f32) -> Self {
```
IQuick143, line 756: Not sure, but I think you may want a check for positive values here, using negative values will most likely give you the wrong winding order on the resulting mesh.

crates/bevy_mesh/src/primitives/dim2.rs
```diff
@@ -452,6 +455,28 @@ pub struct RegularPolygonMeshBuilder {
...
+
+impl RegularPolygonMeshBuilder {
+    /// Creates a new [`RegularPolygonMeshBuilder`] from the radius of a circumcircle and a number
+    /// of sides.
+    ///
+    /// # Panics
+    ///
+    /// Panics if `circumradius` is negative, or if `sides` is less than 3.
+    pub const fn new(circumradius: f32, sides: u32) -> Self {
+        assert!(
```
IQuick143, line 467: The asserts should probably be debug asserts, panicking in production is not good and these code paths are unlikely to lead to data corruption.

crates/bevy_mesh/src/primitives/dim2.rs
```diff
@@ -843,6 +888,15 @@ pub struct RectangleMeshBuilder {
     half_size: Vec2,
 }
 
+impl RectangleMeshBuilder {
+    /// Creates a new [`RectangleMeshBuilder`] from a full width and height.
+    pub const fn new(width: f32, height: f32) -> Self {
```
IQuick143, line 893: Same potential consideration as the Rhombus builder.

crates/bevy_mesh/src/primitives/dim2.rs
IQuick143, line 467:
```suggestion
        debug_assert!(
```

crates/bevy_mesh/src/primitives/dim2.rs
```diff
@@ -452,6 +455,28 @@ pub struct RegularPolygonMeshBuilder {
...
+    ///
+    /// # Panics
+    ///
+    /// Panics if `circumradius` is negative, or if `sides` is less than 3.
+    pub const fn new(circumradius: f32, sides: u32) -> Self {
+        assert!(
+            circumradius.is_sign_positive(),
+            "polygon has a negative radius"
+        );
+        assert!(sides > 2, "polygon has less than 3 sides");
```
IQuick143, line 471:
```suggestion
        debug_assert!(sides > 2, "polygon has less than 3 sides");
```

IQuick143 (changes requested):
I think the

crates/bevy_mesh/src/primitives/dim2.rs
BD103: If `circumradius` is 0 (positive or negative), the mesh still builds successfully but is not visible. I don't think it should panic in this case.

crates/bevy_mesh/src/primitives/dim2.rs
BD103: Yup! As a fun note, the mesh builder tries to allocate an incredibly large `Vec` if `sides < 2`.

crates/bevy_mesh/src/primitives/dim2.rs
BD103: After testing, negative values made the mesh disappear. I'll add a `debug_assert!` warning against it.

crates/bevy_mesh/src/primitives/dim2.rs
BD103: Same thing as rhombus: a negative width / height causes the mesh to disappear. Adding a `debug_assert!`!

crates/bevy_mesh/src/primitives/dim2.rs
```diff
@@ -754,6 +754,12 @@ pub struct RhombusMeshBuilder {
 impl RhombusMeshBuilder {
     /// Creates a new [`RhombusMeshBuilder`] from a horizontal and vertical diagonal size.
     pub const fn new(horizontal_diagonal: f32, vertical_diagonal: f32) -> Self {
+        debug_assert!(
+            horizontal_diagonal >= 0.0,
+            "rhombus has a negative horizontal",
```
IQuick143, line 759:
```suggestion
            "rhombus has a negative horizontal size",
```

crates/bevy_mesh/src/primitives/dim2.rs
```diff
@@ -754,6 +754,12 @@ pub struct RhombusMeshBuilder {
...
+        );
+        debug_assert!(vertical_diagonal >= 0.0, "rhombus has a negative vertical");
```
IQuick143, line 761:
```suggestion
        debug_assert!(vertical_diagonal >= 0.0, "rhombus has a negative vertical size");
```

crates/bevy_mesh/src/primitives/dim2.rs
```diff
@@ -464,11 +464,11 @@ impl RegularPolygonMeshBuilder {
     ///
     /// Panics if `circumradius` is negative, or if `sides` is less than 3.
```
IQuick143, line 465: The `Panic` comment is not entirely accurate, as the panics happen only in debug. (At least I think this is how Panic doc comments should be interpreted?)

IQuick143 (approved):

PR #17297, opened 2025-01-10
Make `ObservedBy` public

ad-kr:
# Objective
- Currently, the `ObservedBy`-component is only public within the `bevy_ecs` crate. Sometimes it is desirable to refer to this component in the "game-code". Two examples that come in mind:
  - Clearing all components in an entity, but intending to keep the existing observers: Making `ObservedBy` public allows us to use `commands.entity(entity).retain::<ObservedBy>();`, which clears all other components, but keeps `ObservedBy`, which prevents the Observers from despawning.
  - The opposite of the above, clearing all of entities' Observers: `commands.entity(entity).remove::<ObservedBy>` will despawn all associated Observers. Admittedly, a cleaner solution would be something like `commands.entity(entity).clear_observers()`, but this is sufficient.

## Solution

- Removed `(crate)` "rule" and added `ObservedBy` to the prelude-module

## Testing

- Linked `bevy_ecs` locally with another project to see if `ObservedBy` could be referenced.

### DIFF
--- a/crates/bevy_ecs/src/lib.rs
+++ b/crates/bevy_ecs/src/lib.rs
@@ -64,7 +64,7 @@ pub mod prelude {
         entity::{Entity, EntityBorrow, EntityMapper},
         event::{Event, EventMutator, EventReader, EventWriter, Events},
         name::{Name, NameOrEntity},
-        observer::{CloneEntityWithObserversExt, Observer, Trigger},
+        observer::{CloneEntityWithObserversExt, ObservedBy, Observer, Trigger},
         query::{Added, AnyOf, Changed, Has, Or, QueryBuilder, QueryState, With, Without},
         removal_detection::RemovedComponents,
         result::{Error, Result},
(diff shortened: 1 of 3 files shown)


BenjaminBrienen (approved):

alice-i-cecile:
This definitely shouldn't be in the prelude :) It's too bloated already and this is quite niche.

I'm on the fence about making this public at all: I'm nervous that it's the sort of implementation detail that's likely to change in subtle and breaking ways. Probably fine (migration path is just to relations), but let me sleep on it.

ad-kr:
@alice-i-cecile I agree on it not really belonging in the prelude, so I've removed it.

I can definitely see a reason to keep some implementation details private. In this case, the `ObservedBy` component is being added to an entity that was likely spawned by "you". And though it is an "internal" thing, personally I think the developer should be able to control the composition of the entity they have spawned. Having this component private kind of limits that.

Right now, in `0.15.1`, there doesn't seem to be a way to preserve Observers when "clearing" all other components*. Just to give an example, my current use-case is that I am trying to implement reactive UI; I want to clear all components and add the "new" ones when rerendering. Adding new observers for each "rerender" really "messes" with event-bubbling, so I'd rather keep the Observers unchanged.

_* It might be possible to spawn an Observer manually, add a temporary entity to force it to not despawn. Clear the entity, and than make the observer watch this entity again. I haven't tested it, but this or something similar might work. I'm not convinced that it is the most optimal approach._

PR #17311, opened 2025-01-11
Renamed members of `ParamWarnPolicy` to reflect new behaviour.

### LINKED ISSUE #17262: Rename `never_param_warn` into something like `ignore_param_missing`

I found this name confusing.
In 0.15 we originally used warnings for missing system params by default. And `never_param_warn` disables it. But later we changed the default behavior to panic.

I think we need to rename `never_param_warn` into something like `ignore_param_missing`. Because this function now changes the behavior from panic into system skip.

AlephCubed:
- `Once` renamed to `Warn`.
- `param_warn_once()` renamed to `warn_param_missing()`.
- `never_param_warn()` renamed to `ignore_param_missing()`.

Also includes changes to the documentation of the above methods.

Fixes #17262.

## Migration Guide
- `ParamWarnPolicy::Once` has been renamed to `ParamWarnPolicy::Warn`.
- `ParamWarnPolicy::param_warn_once` has been renamed to `ParamWarnPolicy::warn_param_missing`.
- `ParamWarnPolicy::never_param_warn` has been renamed to `ParamWarnPolicy::ignore_param_missing`.

### DIFF
--- a/crates/bevy_ecs/src/schedule/executor/mod.rs
+++ b/crates/bevy_ecs/src/schedule/executor/mod.rs
@@ -344,7 +344,7 @@ mod tests {
                 (|mut commands: Commands| {
                     commands.insert_resource(R2);
                 })
-                .param_warn_once(),
+                .warn_param_missing(),
             )
                 .chain(),
         );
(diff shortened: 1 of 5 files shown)


BenjaminBrienen (approved):

alice-i-cecile:
I'll merge this once there's a migration guide :)

AlephCubed:
Added :)

PR #7947, opened 2023-03-07
remove bors and small CI improvements

mockersf:
# Objective

- Remove bors
- Some CI improvements

## Solution

- Moving from bors to merge queues:
  - Remove bors config and support for bors branches
  - Add a unique status job for the matrix in validation that can be used in branch protection rules
- Reduce load:
  - Remove cron execution on dependency check jobs. did anyone check those?
  - Move the unused dependency check from CI to validation so that it runs less often. it doesn't raise issues enough to keep it on every push
- Use less runners at the start without changing the total duration. This reduce the "burst" usage so that more PRs can run in parallel
  - Reorder markdown jobs (examples, features, lint) so that they run sequentially
  - Reduce parallelism of "no default features" job matrix

PR #17321, opened 2025-01-12
Set certain workflows to run in the merge queue, but not on main

LikeLakers2:
# Objective
I discovered that certain workflows run more often than they need to. In particular, the CI and Validation Jobs workflows run twice when merged - once in the merge queue, and another when properly merged into `main`.

Additionally, the Dependencies workflow was not running in the merge queue at all.

## Solution
Set CI, Validation Jobs, and Dependencies to run in the merge queue, but not in workflows.

The "Deploy Docs" workflow has not been changed; it will continue to run only on pushes to `main` (and when dispatched manually), as it deploys to Github pages - and we don't want failed merge queue stuff to deploy to github pages!

## Testing
Well I guess we'll find out if these changes work, won't we?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,7 +5,6 @@ on:
   pull_request:
   push:
     branches:
-      - main
       - release-*
 
 env:
--- a/.github/workflows/dependencies.yml
+++ b/.github/workflows/dependencies.yml
@@ -1,6 +1,7 @@
 name: Dependencies
 
 on:
+  merge_group:
   pull_request:
     paths:
       - '**/Cargo.toml'
@@ -9,8 +10,6 @@ on:
     paths:
       - '**/Cargo.toml'
       - 'deny.toml'
-    branches:
-      - main
 
 concurrency:
   group: ${{github.workflow}}-${{github.ref}}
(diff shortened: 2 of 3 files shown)


mockersf:
Yup, *some* of them are needed on main, but not all of them. 

I can fix that as I need to change some of those for #15918, but I won't have time to do it before the end of the month.

Adding the dependencies workflow to the merge queue won't do anything as those jobs are not (and can't be) required

LikeLakers2:
> Adding the dependencies workflow to the merge queue won't do anything as those jobs are not (and can't be) required

Well it doesn't make sense for them to run after the merge to main... so what should I do here?

mockersf:
> so what should I do here?

Dig through history, try to understand why it was done that way, document those needs and improve the separations.

Or wait and I'll do that by the end of the month

LikeLakers2:
So digging through history, I found that `dependencies` running on `main` was added by you... And your reason was *also* to reduce load.
https://github.com/bevyengine/bevy/pull/7947

So frankly, I'm not sure I need to do much here.

mockersf:
Closing then, I'll pick that up with comments in the changes for screenshots

PR #17327, opened 2025-01-12
added `any_match_filter` common condition

### LINKED ISSUE #17326: Add `any_for_filter` to built-in run conditions

Currently, the only built-in run condition like this is `any_with_component`. However, this very quickly loses value on the systems that depend on a combination of components including the provided component type. Compounding `any_with_component` doesn't address this as it doesn't look at archetypes like a filter does.

Take the following system for example:

```
fn update(mut enemies: Query<(&EnemyMovement, &mut Transform)>, time: Res<Time>) {
    for (movement, mut transform) in enemies.iter_mut() {
        transform.translation.y -= time.delta_secs() * movement.speed;
    }
}
```

This is a very basic system, but the built-in run conditions do not provide means to limit the scheduling of this system. If I were to use `any_with_component::<EnemyMovement>`, this system will still run if there are `EnemyMovement` components that exist without `Transform`, and therefore time would be wasted scheduling this system for execution in that frame. Instead, I've used the suggested condition:

```
app.add_systems(
  Update,
  update.run_if(any_for_filter::<(With<EnemyMovement>, With<Transform>)>),
);
```

I'm using this in every project in order to appropriately limit my scheduled systems:

```
pub fn any_for_filter<F: QueryFilter>(entities: Query<(), F>) -> bool {
    !entities.is_empty()
}
```

No alternatives.

## Additional context

N/A

Jaso333:
# Objective

resolves #17326.

## Solution

Simply added the suggested run condition.

## Testing

A self-explanatory run condition. Fully verified by the operation of `QueryFilter` in a system.

### DIFF
--- a/crates/bevy_ecs/src/schedule/condition.rs
+++ b/crates/bevy_ecs/src/schedule/condition.rs
@@ -940,7 +940,7 @@ pub mod common_conditions {
 
     /// A [`Condition`]-satisfying system that returns `true`
     /// if there are any entities that match the given [`QueryFilter`].
-    pub fn any_for_filter<F: QueryFilter>(query: Query<(), F>) -> bool {
+    pub fn any_match_filter<F: QueryFilter>(query: Query<(), F>) -> bool {
         !query.is_empty()
     }
 
@@ -1404,7 +1404,7 @@ mod tests {
                 .distributive_run_if(resource_removed::<TestResource>)
                 .distributive_run_if(on_event::<TestEvent>)
                 .distributive_run_if(any_with_component::<TestComponent>)
-                .distributive_run_if(any_for_filter::<With<TestComponent>>)
+                .distributive_run_if(any_match_filter::<With<TestComponent>>)
                 .distributive_run_if(not(run_once)),
         );
     }
--- a/crates/bevy_ecs/src/schedule/condition.rs
+++ b/crates/bevy_ecs/src/schedule/condition.rs
@@ -398,6 +398,7 @@ pub mod common_conditions {
         change_detection::DetectChanges,
         event::{Event, EventReader},
         prelude::{Component, Query, With},
+        query::QueryFilter,
         removal_detection::RemovedComponents,
         system::{In, IntoSystem, Local, Res, Resource, System, SystemInput},
     };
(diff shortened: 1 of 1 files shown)


BenjaminBrienen (approved):

alice-i-cecile (changes requested):
I like this but the name is quite unclear to me :) How about `any_match_filter`?

Jaso333:
@alice-i-cecile agreed, it was more of a placeholder name initially, forgot to mention the fact. The documentation above the method even describes it according to your suggestion... 😅

```
/// A [`Condition`]-satisfying system that returns `true`
/// if there are any entities that match the given [`QueryFilter`].
```

I have updated the code accordingly.

benfrankel:
Grammatically it should be `any_matches_filter`.

Jaso333:
> Grammatically it should be `any_matches_filter`.
> 
> EDIT: I guess "any" can be either singular or plural. So I'll amend that to "IMO `any_matches_filter` reads better".

I think I prefer `any_match_filter` as its more natural, in my opinion, to think of the "any" as a plural in the context of a set of entities. This was probably due to me thinking the same way with `any_with_component`, but that also works both ways too!

alice-i-cecile (approved):

chescock:
Do we still need this now that we have [`Populated`](https://docs.rs/bevy/latest/bevy/ecs/prelude/struct.Populated.html)?  The motivating example in the linked issue seems like it would be better served by replacing `Query` with `Populated`, since that avoids having to duplicate the query filters.

Jaso333:
> Do we still need this now that we have [`Populated`](https://docs.rs/bevy/latest/bevy/ecs/prelude/struct.Populated.html)? The motivating example in the linked issue seems like it would be better served by replacing `Query` with `Populated`, since that avoids having to duplicate the query filters.

There's a few reasons I can think of:

- Populated fails validation and emits a warning when there are no entities. This can be turned off, but it's an extra step that goes against the engine's default behaviour.
- There is another example in my game where I want to run a system if a component combination _exists_, but I don't want to do anything with those entities. Without this, I would have to include an unused parameter on my system.
- any_with_component already exists, and has the same purpose as this condition, only more restrictive.

PR #17333, opened 2025-01-12
Remove `Event: Component` trait bound

alice-i-cecile:
# Objective

As raised in https://github.com/bevyengine/bevy/pull/17317, the `Event: Component` trait bound is confusing to users.

In general, a type `E` (like `AppExit`) which implements `Event` should not:

- be stored as a component on an entity
- be a valid option for `Query<&AppExit>`
- require the storage type and other component metadata to be specified

Events are not components (even if they one day use some of the same internal mechanisms), and this trait bound is confusing to users.

We're also automatically generating `Component` impls with our derive macro, which should be avoided when possible to improve explicitness and avoid conflicts with user impls.

P.S. This prompted a [larger hackmd design](https://hackmd.io/@bevy/SypE1qZP1l) from me about the core principles of "make everything entities and components". 

## Context

Dynamic events (which observers support IIRC) require a mapping from `TypeId` to "some arbitrary stable ID". We decided to use `ComponentId` there when implementing this, as the code / concept already exists. The easiest way to get the observers API compiling as a result was to just add a `Component` trait bound.

## Solution

1. Add a dead simple way to get a `ComponentId` for any Rust types.
2. Generate the `ComponentId` for events (inside of the observers API) using this new method.
3. Relax the `Component` trait bound on `Event` to `Send + Sync + 'static`.

Unlike I initially thought, we cannot (and should not) relax the `Component` trait bound on `register_component`, as other work is done to make sure the component is properly set up (like registering hooks).

This change does not:

- restrict our ability to implement dynamic buffered events (the main improvement over #17317)
   - there's still a fair bit of work to do, but this is a step in the right direction
- limit our ability to store event metadata on entities in the future
- make it harder for users to work with types that are both events and components (just add the derive / trait bound)

## Future work

1. Use `generate_component_id` inside of our existing methods to improve clarity and reduce duplication.
2. Consider renaming `ComponentId` to something that isn't component-specific.

## Testing

Code compiles, and `alien_cake_addict` works. If something goes wrong here *everything* will break.

## Migration Guide

### DIFF
--- a/crates/bevy_ecs/src/component.rs
+++ b/crates/bevy_ecs/src/component.rs
@@ -1098,6 +1098,20 @@ pub struct Components {
 }
 
 impl Components {
+    /// Generates a new [`ComponentId`] for the provided Rust type from its [`TypeId`],
+    /// returning any existing [`ComponentId`] if the type has already been registered.
+    ///
+    /// For read-only access, see [`Components::component_id`].
+    pub fn generate_component_id<T: ?Sized + 'static>(&mut self) -> ComponentId {
+        if let Some(component_id) = self.indices.get(&TypeId::of::<T>()) {
+            component_id.clone()
+        } else {
+            let id = ComponentId(self.components.len());
+            self.indices.insert(TypeId::of::<T>(), id);
+            id
+        }
+    }
+
     /// Registers a [`Component`] of type `T` with this instance.
     /// If a component of this type has already been registered, this will return
     /// the ID of the pre-existing component.
--- a/crates/bevy_ecs/src/event/base.rs
+++ b/crates/bevy_ecs/src/event/base.rs
@@ -20,7 +20,7 @@ use core::{
 /// This trait can be derived.
 ///
 /// Each event type has a unique [`ComponentId`] associated with it, which is used to store the event in the world.
-/// This can be obtained using [`World::generate_component_id<Self>`].
+/// This can be obtained using [`World::generate_component_id<Self>`](crate::world::World::generate_component_id).
 ///
 /// Each individual event sent has an [`EventId`] associated with it, which can be used to trace the flow of an event.
 ///
--- a/crates/bevy_ecs/src/event/base.rs
+++ b/crates/bevy_ecs/src/event/base.rs
@@ -19,8 +19,10 @@ use core::{
 ///
 /// This trait can be derived.
 ///
-/// Each event has a unique [`ComponentId`] associated with it, which is used to store the event in the world.
+/// Each event type has a unique [`ComponentId`] associated with it, which is used to store the event in the world.
 /// This can be obtained using [`World::generate_component_id<Self>`].
+///
+/// Each individual event sent has an [`EventId`] associated with it, which can be used to trace the flow of an event.
 ///
 /// Events must be thread-safe.
 ///
--- a/crates/bevy_ecs/src/event/base.rs
+++ b/crates/bevy_ecs/src/event/base.rs
@@ -1,4 +1,4 @@
-use crate::{component::Component, traversal::Traversal};
+use crate::traversal::Traversal;
 #[cfg(feature = "bevy_reflect")]
 use bevy_reflect::Reflect;
 #[cfg(feature = "track_location")]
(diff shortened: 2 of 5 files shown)


crates/bevy_ecs/src/component.rs
```diff
@@ -1098,6 +1098,20 @@ pub struct Components {
 }
 
 impl Components {
+    /// Generates a new [`ComponentId`] for the provided Rust type from its [`TypeId`],
+    /// returning any existing [`ComponentId`] if the type has already been registered.
+    ///
+    /// For read-only access, see [`Components::component_id`].
+    pub fn generate_component_id<T: ?Sized + 'static>(&mut self) -> ComponentId {
+        if let Some(component_id) = self.indices.get(&TypeId::of::<T>()) {
```
alice-i-cecile, line 1106: Is `.entry` faster in general? I find this style a bit clearer.

crates/bevy_ecs/src/observer/runner.rs
```diff
@@ -398,13 +398,13 @@ fn hook_on_add<E: Event, B: Bundle, S: ObserverSystem<E, B>>(
...
     world.commands().queue(move |world: &mut World| {
-        let event_type = world.register_component::<E>();
+        let event_id = world.generate_component_id::<E>();
         let mut components = Vec::new();
         B::component_ids(&mut world.components, &mut world.storages, &mut |id| {
             components.push(id);
         });
         let mut descriptor = ObserverDescriptor {
-            events: vec![event_type],
+            events: vec![event_id],
```
alice-i-cecile, line 407: Renamed for consistency.

crates/bevy_ecs/src/event/base.rs
```diff
@@ -19,9 +19,10 @@ use core::{
 ///
 /// This trait can be derived.
 ///
-/// Events implement the [`Component`] type (and they automatically do when they are derived). Events are (generally)
-/// not directly inserted as components. More often, the [`ComponentId`] is used to identify the event type within the
-/// context of the ECS.
+/// Each event type has a unique [`ComponentId`] associated with it, which is used to store the event in the world.
+/// This can be obtained using [`World::generate_component_id<Self>`].
+///
+/// Each individual event sent has an [`EventId`] associated with it, which can be used to trace the flow of an event.
```
alice-i-cecile, line 25: Bonus docs!

crates/bevy_ecs/src/component.rs
```diff
@@ -1098,6 +1098,20 @@ pub struct Components {
...
+            component_id.clone()
+        } else {
+            let id = ComponentId(self.components.len());
```
james-j-obrien, line 1109: If we aren't registering a component info won't this return the same ID multiple times since the length of self.components does not increase?

chescock:
> On top of that, if the user calls `generate_component_id::<Foo>` before the first time `register_component::<Foo>` is called, the initialization of the component descriptor will be skipped.
> 
> I see two solutions:
> 
> * Add a `HasComponentId` trait that can optionally return a component descriptor (no specialization sadge) and remove `generate_component_id`; keep track of maximum free id and allocate ids of non-components from high to low so the component descriptor array is kept dense (I'd favor this version)
> * or have component descriptors be stored in a sparse array, with `register_component` doing the initialization even if the id is already allocated (but there isn't an entry in the array)

It would be a little unfortunate to be unable to record *any* type information for these. 
 `ComponentDescriptor` has fields like `layout` and `drop` that are useful when working with untyped data.  A variation on the second idea could be to split up `ComponentInfo` into things we can get from any type, and things that only apply to components.  Then it's only the component-specific stuff that needs to be sparse.  

A third solution is to do what `Resource` does and have a third `TypeIdMap<ComponentId>` for these that is separate from `indices` and `resource_indices`.  Then a single type could be used in all three ways with potentially different configuration.

alice-i-cecile:
I think I'm going to go with solution 3 here: it's the simplest incremental change. I *really* don't like that you can get duplicate IDs for different types, but that's a pre-existing problem.

alice-i-cecile:
> > but that's a pre-existing problem.
> 
> How so?

The resources and component indexes are incremented independently now.

alice-i-cecile:
Closing in favor of #17380, which is even simpler.

PR #17334, opened 2025-01-13
Reflect and register the wireframe materials

SludgePhD:
# Objective

These were missing, but can trivially be reflected.

## Solution

Do that.


SludgePhD:
On second thought, there might not be much of a point to this since the materials just mirror what the wireframe component contains.

BenjaminBrienen (approved):

BenjaminBrienen:
https://github.com/bevyengine/bevy/issues/17474

PR #17335, opened 2025-01-13
bevy_ecs: Apply `#![warn(clippy::allow_attributes, clippy::allow_attributes_without_reason)]`

LikeLakers2:
# Objective

## Solution
Set the `clippy::allow_attributes` and `clippy::allow_attributes_without_reason` lints to `warn`, and bring `bevy_ecs` in line with the new restrictions.

## Testing
This PR is a WIP; testing will happen after it's finished.

### DIFF
--- a/crates/bevy_ecs/src/archetype.rs
+++ b/crates/bevy_ecs/src/archetype.rs
@@ -391,7 +391,7 @@ impl Archetype {
         let (min_sparse, _) = sparse_set_components.size_hint();
         let mut flags = ArchetypeFlags::empty();
         let mut archetype_components = SparseSet::with_capacity(min_table + min_sparse);
-        for (component_id, archetype_component_id) in table_components {
+        for (idx, (component_id, archetype_component_id)) in table_components.enumerate() {
             // SAFETY: We are creating an archetype that includes this component so it must exist
             let info = unsafe { components.get_info_unchecked(component_id) };
             info.update_archetype_flags(&mut flags);
@@ -409,7 +409,7 @@ impl Archetype {
             component_index
                 .entry(component_id)
                 .or_default()
-                .insert(id, ArchetypeRecord {});
+                .insert(id, ArchetypeRecord { column: Some(idx) });
         }
 
         for (component_id, archetype_component_id) in sparse_set_components {
@@ -427,7 +427,7 @@ impl Archetype {
             component_index
                 .entry(component_id)
                 .or_default()
-                .insert(id, ArchetypeRecord {});
+                .insert(id, ArchetypeRecord { column: None });
         }
         Self {
             id,
@@ -786,7 +786,15 @@ pub struct Archetypes {
 }
 
 /// Metadata about how a component is stored in an [`Archetype`].
-pub struct ArchetypeRecord {}
+pub struct ArchetypeRecord {
+    /// Index of the component in the archetype's [`Table`](crate::storage::Table),
+    /// or None if the component is a sparse set component.
+    #[expect(
+        dead_code,
+        reason = "Currently unused, but planned to be used to implement a component index to improve performance of fragmenting relations."
+    )]
+    pub(crate) column: Option<usize>,
+}
 
 impl Archetypes {
     pub(crate) fn new() -> Self {
--- a/crates/bevy_ecs/src/archetype.rs
+++ b/crates/bevy_ecs/src/archetype.rs
@@ -789,7 +789,7 @@ pub struct Archetypes {
 pub struct ArchetypeRecord {
     /// Index of the component in the archetype's [`Table`](crate::storage::Table),
     /// or None if the component is a sparse set component.
-    #[allow(dead_code)]
+    #[expect(dead_code)]
     pub(crate) column: Option<usize>,
 }
 
@@ -827,7 +827,7 @@ impl Archetypes {
 
     /// Fetches the total number of [`Archetype`]s within the world.
     #[inline]
-    #[allow(clippy::len_without_is_empty)] // the internal vec is never empty.
+    #[expect(clippy::len_without_is_empty)] // the internal vec is never empty.
     pub fn len(&self) -> usize {
         self.archetypes.len()
     }
--- a/crates/bevy_ecs/src/archetype.rs
+++ b/crates/bevy_ecs/src/archetype.rs
@@ -391,7 +391,7 @@ impl Archetype {
         let (min_sparse, _) = sparse_set_components.size_hint();
         let mut flags = ArchetypeFlags::empty();
         let mut archetype_components = SparseSet::with_capacity(min_table + min_sparse);
-        for (idx, (component_id, archetype_component_id)) in table_components.enumerate() {
+        for (component_id, archetype_component_id) in table_components {
             // SAFETY: We are creating an archetype that includes this component so it must exist
             let info = unsafe { components.get_info_unchecked(component_id) };
             info.update_archetype_flags(&mut flags);
@@ -409,7 +409,7 @@ impl Archetype {
             component_index
                 .entry(component_id)
                 .or_default()
-                .insert(id, ArchetypeRecord { column: Some(idx) });
+                .insert(id, ArchetypeRecord {});
         }
 
         for (component_id, archetype_component_id) in sparse_set_components {
@@ -427,7 +427,7 @@ impl Archetype {
             component_index
                 .entry(component_id)
                 .or_default()
-                .insert(id, ArchetypeRecord { column: None });
+                .insert(id, ArchetypeRecord {});
         }
         Self {
             id,
@@ -786,12 +786,7 @@ pub struct Archetypes {
 }
 
 /// Metadata about how a component is stored in an [`Archetype`].
-pub struct ArchetypeRecord {
-    /// Index of the component in the archetype's [`Table`](crate::storage::Table),
-    /// or None if the component is a sparse set component.
-    #[expect(dead_code, reason = "This field is written to, but never read.")]
-    pub(crate) column: Option<usize>,
-}
+pub struct ArchetypeRecord {}
 
 impl Archetypes {
     pub(crate) fn new() -> Self {
--- a/crates/bevy_ecs/src/archetype.rs
+++ b/crates/bevy_ecs/src/archetype.rs
@@ -789,10 +789,7 @@ pub struct Archetypes {
 pub struct ArchetypeRecord {
     /// Index of the component in the archetype's [`Table`](crate::storage::Table),
     /// or None if the component is a sparse set component.
-    #[expect(
-        dead_code,
-        reason = "This field is written to, but never read."
-    )]
+    #[expect(dead_code, reason = "This field is written to, but never read.")]
     pub(crate) column: Option<usize>,
 }
 
--- a/crates/bevy_ecs/src/archetype.rs
+++ b/crates/bevy_ecs/src/archetype.rs
@@ -789,7 +789,10 @@ pub struct Archetypes {
 pub struct ArchetypeRecord {
     /// Index of the component in the archetype's [`Table`](crate::storage::Table),
     /// or None if the component is a sparse set component.
-    #[expect(dead_code)]
+    #[expect(
+        dead_code,
+        reason = "This field is written to, but never read."
+    )]
     pub(crate) column: Option<usize>,
 }
 
@@ -827,7 +830,10 @@ impl Archetypes {
 
     /// Fetches the total number of [`Archetype`]s within the world.
     #[inline]
-    #[expect(clippy::len_without_is_empty)] // the internal vec is never empty.
+    #[expect(
+        clippy::len_without_is_empty,
+        reason = "The internal vec is never empty"
+    )]
     pub fn len(&self) -> usize {
         self.archetypes.len()
     }
--- a/crates/bevy_ecs/src/entity/entity_set.rs
+++ b/crates/bevy_ecs/src/entity/entity_set.rs
@@ -385,11 +385,11 @@ mod tests {
     use crate::prelude::{Schedule, World};
 
     use crate::component::Component;
+    use crate::entity::Entity;
     use crate::query::{QueryState, With};
     use crate::system::Query;
     use crate::world::Mut;
     use crate::{self as bevy_ecs};
-    use crate::entity::Entity;
 
     use super::UniqueEntityIter;
 
--- a/crates/bevy_ecs/src/entity/entity_set.rs
+++ b/crates/bevy_ecs/src/entity/entity_set.rs
@@ -382,17 +382,17 @@ impl<I: Iterator<Item: TrustedEntityBorrow> + Debug> Debug for UniqueEntityIter<
 mod tests {
     use alloc::{vec, vec::Vec};
 
-    #[allow(unused_imports)]
+    #[expect(unused_imports)]
     use crate::prelude::{Schedule, World};
 
-    #[allow(unused_imports)]
+    #[expect(unused_imports)]
     use crate::component::Component;
     use crate::query::{QueryState, With};
     use crate::system::Query;
     use crate::world::Mut;
-    #[allow(unused_imports)]
+    #[expect(unused_imports)]
     use crate::{self as bevy_ecs};
-    #[allow(unused_imports)]
+    #[expect(unused_imports)]
     use crate::{entity::Entity, world::unsafe_world_cell};
 
     use super::UniqueEntityIter;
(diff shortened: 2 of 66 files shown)


crates/bevy_ecs/src/lib.rs
```diff
@@ -60,7 +60,10 @@ pub use bevy_ptr as ptr;
 ///
 /// This includes the most common types in this crate, re-exported for your convenience.
 pub mod prelude {
-    #[expect(deprecated)]
+    #[expect(
+        deprecated,
+        reason = "`crate::schedule::apply_deferred` is considered deprecated; however, it may still be used by crates which consume `bevy_ecs`, so its removal here may cause confusion."
```
LikeLakers2, line 65: I really couldn't think of a better reason, but I also don't know why it's still here.

crates/bevy_ecs/src/lib.rs
```diff
@@ -152,9 +155,8 @@ mod tests {
     #[derive(Component, Debug, PartialEq, Eq, Clone, Copy)]
     struct C;
 
-    #[expect(dead_code)]
     #[derive(Default)]
-    struct NonSendA(usize, PhantomData<*mut ()>);
+    struct NonSendA(PhantomData<*mut ()>);
```
LikeLakers2, line 159: In case you're wondering: I checked to see if this `usize` had any purpose. It did in the past, [when this struct was first added](https://github.com/bevyengine/bevy/commit/0bf7f3153d94543078ef518b2a699b1f24b08dd5#diff-696a3460ea2e00a5e92135018cb36f5bc0e08214a1e24aa0dac608f3c6d3a865R1269-R1278). However, it seems that test has since been removed; for what reason, I'm unsure.

Regardless, the `usize` is no longer needed, so I've removed it.

crates/bevy_ecs/src/lib.rs
```diff
@@ -171,8 +173,10 @@ mod tests {
         }
     }
 
-    // TODO: The compiler says the Debug and Clone are removed during dead code analysis. Investigate.
-    #[expect(dead_code)]
+    #[expect(
+        dead_code,
+        reason = "This struct is used to test how `Drop` behavior works in regards to SparseSet storage, and as such is solely a wrapper around `DropCk` to make it use the SparseSet storage. Because of this, the inner field is intentionally never read."
```
LikeLakers2, line 178: I'm particularly proud of this comment. It took quite a bit to figure out the struct's use, and longer still to nail the messaging, but I'm proud of it.

crates/bevy_ecs/src/archetype.rs
```diff
@@ -789,7 +789,10 @@ pub struct Archetypes {
 pub struct ArchetypeRecord {
     /// Index of the component in the archetype's [`Table`](crate::storage::Table),
     /// or None if the component is a sparse set component.
-    #[expect(dead_code)]
+    #[expect(
+        dead_code,
+        reason = "This field is written to, but never read."
+    )]
     pub(crate) column: Option<usize>,
```
LikeLakers2, line 796: To clarify: This variable is given a value in certain places, but that value never seems to be used. So I... don't actually know what the point of this variable is.

crates/bevy_ecs/src/entity/entity_set.rs
```diff
@@ -382,25 +382,24 @@ impl<I: Iterator<Item: TrustedEntityBorrow> + Debug> Debug for UniqueEntityIter<
...
 
     use super::UniqueEntityIter;
 
     #[derive(Component, Clone)]
     pub struct Thing;
 
-    #[expect(clippy::iter_skip_zero)]
+    #[expect(
+        clippy::iter_skip_zero,
+        reason = "Uniqueness of entities should be preserved, even when skipping zero entities."
```
LikeLakers2, line 401: I'm not actually convinced this is the reason... But I need to hear back from @Victoronz first, before removing the `.skip(0)`.

crates/bevy_ecs/src/archetype.rs
BenjaminBrienen: It isn't `pub`, so it can just be removed as the fix.

crates/bevy_ecs/src/world/entity_ref.rs
```diff
@@ -1758,7 +1758,7 @@ impl<'w> EntityWorldMut<'w> {
             })
         };
 
-        #[allow(clippy::undocumented_unsafe_blocks)] // TODO: document why this is safe
+        #[expect(clippy::undocumented_unsafe_blocks)] // TODO: document why this is safe
```
BenjaminBrienen, line 1761: Can you create an issue for this and link it in the reason?

crates/bevy_ecs/src/archetype.rs
LikeLakers2: I'll remove this once I've finished linting work, as I want to run `cargo test` if I'm going to change this.

crates/bevy_ecs/src/lib.rs
```diff
@@ -2643,40 +2659,49 @@ mod tests {
 
     // These structs are primarily compilation tests to test the derive macros. Because they are
     // never constructed, we have to manually silence the `dead_code` lint.
-    #[allow(dead_code)]
+    #[expect(
```
BenjaminBrienen, line 2662: I think this can go on the whole function with the reason simplified to "this is a compilation test".

crates/bevy_ecs/src/lib.rs
```diff
@@ -1,7 +1,14 @@
...
-#![allow(dependency_on_unit_never_type_fallback)]
+#![expect(
+    unsafe_op_in_unsafe_fn,
+    reason = "See #11590. To be removed once all applicable unsafe code has an unsafe block with a safety comment."
+)]
+#![cfg_attr(
+    test,
+    expect(
+        dependency_on_unit_never_type_fallback,
+        reason = "To be removed once Edition 2024 is released"
```
BenjaminBrienen, line 9: Can you link a new issue here? The issue will just have the blocked status for now.

crates/bevy_ecs/src/lib.rs
LikeLakers2: These aren't in a function - they're in the `tests` module.

crates/bevy_ecs/src/world/entity_ref.rs
LikeLakers2: When I get to this `expect()`, I will definitely do so.

crates/bevy_ecs/src/lib.rs
LikeLakers2: https://github.com/bevyengine/bevy/pull/17335/commits/578e392f09adc9648bd3de1912be3f5934f663ac

crates/bevy_ecs/src/lib.rs
LikeLakers2: Are you suggesting I put these into a function? Currently, these structs and expects are within the `tests` module, not within a test function body.

BenjaminBrienen: I misread it, sorry! Nevermind me.

crates/bevy_ecs/src/world/entity_ref.rs
LikeLakers2: Handled in https://github.com/bevyengine/bevy/pull/17335/commits/9fc969846f5cd4cf653da7d3e4b474c89760589d

crates/bevy_ecs/src/world/entity_ref.rs
```diff
@@ -3653,12 +3652,11 @@ impl<B: Bundle> PartialEq for EntityRefExcept<'_, B> {
 
 impl<B: Bundle> Eq for EntityRefExcept<'_, B> {}
 
-#[expect(clippy::non_canonical_partial_ord_impl)]
 impl<B: Bundle> PartialOrd for EntityRefExcept<'_, B> {
     /// [`EntityRefExcept`]'s comparison trait implementations match the underlying [`Entity`],
     /// and cannot discern between different worlds.
     fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
-        self.entity().partial_cmp(&other.entity())
+        Some(self.cmp(other))
```
BenjaminBrienen, line 3659: Does this change how this does the comparison? In other words, is this identical to the previous code?

LikeLakers2: It doesn't change the comparison - it just makes the call stack slightly different.

`Entity`'s `PartialOrd` impl is written in terms of its `Ord` impl:

https://github.com/bevyengine/bevy/blob/141b7673ab3b0af13d49349b97f053286bb16227/crates/bevy_ecs/src/entity/mod.rs#L214-L220

...which means both `self.entity().cmp(&other.entity())` and `self.entity().partial_cmp(&other.entity())` eventually call the same thing.

And since [`clippy::non_canonical_partial_ord_impl`](https://rust-lang.github.io/rust-clippy/master/index.html#non_canonical_partial_ord_impl) wants us to define `PartialOrd` in terms of `Ord`, and the comparison is the same either way, I figured it was easier to rewrite this.

crates/bevy_ecs/src/system/commands/mod.rs
```diff
@@ -2163,7 +2162,10 @@ mod tests {
         sync::atomic::{AtomicUsize, Ordering},
     };
 
-    #[expect(dead_code)]
+    #[expect(
+        dead_code,
+        reason = "This struct is used to test how `Drop` behavior works in regards to SparseSet storage, and as such is solely a wrapper around `DropCk` to make it use the SparseSet storage. Because of this, the inner field is intentionally never read."
```
LikeLakers2, line 2167: This feels like a duplicated drop check test (the other instance being in `crates/bevy_ecs/src/lib.rs`, with the structs `DropCk` and `DropCkSparse`)...

LikeLakers2:
This PR is nearly finished. I'm going to let CI run, and see if there's anything else that needs fixing up. Then, I'll mark this as ready for review.

crates/bevy_ecs/src/archetype.rs
```diff
@@ -789,7 +789,7 @@ pub struct Archetypes {
 pub struct ArchetypeRecord {
     /// Index of the component in the archetype's [`Table`](crate::storage::Table),
     /// or None if the component is a sparse set component.
-    #[allow(dead_code)]
+    #[expect(dead_code, reason = "This field is written to, but never read.")]
```
BenjaminBrienen, line 792: That's true, but why is that a justification? The reason should focus on the "why".

BenjaminBrienen: Oh, I forgot about that too lol

crates/bevy_ecs/src/archetype.rs
```diff
@@ -786,12 +786,7 @@ pub struct Archetypes {
 }
 
 /// Metadata about how a component is stored in an [`Archetype`].
-pub struct ArchetypeRecord {
-    /// Index of the component in the archetype's [`Table`](crate::storage::Table),
-    /// or None if the component is a sparse set component.
-    #[expect(dead_code, reason = "This field is written to, but never read.")]
-    pub(crate) column: Option<usize>,
-}
+pub struct ArchetypeRecord {}
```
LikeLakers2: Per Periwink on Discord (<https://discord.com/channels/691052431525675048/749335865876021248/1328528070323208273>):

> It was supposed to be used for the move toward FragmentingRelation and the work to add a ComponentIndex. If it's un-read right now it can probably be removed for now and re-added later

crates/bevy_ecs/src/archetype.rs
LikeLakers2: https://github.com/bevyengine/bevy/pull/17335#discussion_r1913984135

crates/bevy_ecs/src/lib.rs
alice-i-cecile: This will be removed in the Bevy 0.17 cycle.

crates/bevy_ecs/src/query/iter.rs
```diff
@@ -2935,10 +2935,9 @@ impl<T> PartialEq for NeutralOrd<T> {
 
 impl<T> Eq for NeutralOrd<T> {}
 
-#[allow(clippy::non_canonical_partial_ord_impl)]
 impl<T> PartialOrd for NeutralOrd<T> {
-    fn partial_cmp(&self, _other: &Self) -> Option<Ordering> {
-        Some(Ordering::Equal)
+    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
+        Some(self.cmp(other))
```
Victoronz, line 2940: This should just return `Some(Ordering::Equal)`, since that is what this call will always evaluate to!

LikeLakers2: That causes it to lint [`clippy::non_canonical_partial_ord_impl`](https://rust-lang.github.io/rust-clippy/master/index.html#non_canonical_partial_ord_impl) though.

Since `Ord` always returns `Ordering::Equal`, and deferring to the `Ord` impl (as the lint wants us to) will ensure the two impls agree, I figured the easier option was to just do this.

crates/bevy_ecs/src/lib.rs
LikeLakers2: I will edit the reason to mention this.

crates/bevy_ecs/src/query/iter.rs
Victoronz: I'd usually agree, but in this case, this exact neutral `PartialOrd`/`Ord` impl is the very point of this struct! 
Because of that, this impl must not/will never be refactored, and the purpose of this lint does not apply.

LikeLakers2: ...What? This new version has the exact same effect as before, and doesn't require an `#[expect()]` atop the trait impl block like putting `Some(Ordering::Equal)` would. I don't get what the issue is here.

BenjaminBrienen (approved):

crates/bevy_ecs/src/query/iter.rs
BenjaminBrienen: They're saying that their suggestion would suit this type more and be more clear and explicit despite triggering clippy. I'm fine with either.

LikeLakers2: I see. In that case: If I put `Some(Ordering::Equal)`, I have to explain why it's not just calling into the `Ord` impl in the `#[expect()]`. And currently, I'm not sure what reason I would put down.

Victoronz: As a reason:
"`PartialOrd` and `Ord` on this struct must only ever return `Ordering::Equal`, so no need to worry about refactoring."

Going off of the "why is this bad" section in the [clippy lint index](https://rust-lang.github.io/rust-clippy/master/index.html#/non_canonical_partial_ord_impl).

alice-i-cecile: Agreed; this should be an expect and I like Victoronz reason here :)

Victoronz: To elaborate on why I'd like to keep the lint here: This is a non-canonical `PartialOrd`/`Ord` impl, in the actual meaning of those words!
These impls intentionally don't do anything/return anything useful, which I'd like to have remain clear.
While currently private, this is potentially a useful type to have public, somewhat akin to `FloatOrd`.
In the presence of looking at type impls via IDEs, I personally find it to be less mental load to only have to look at one trait impl and know it is constant, vs being redirected to another, especially when this type can wrap non-standard ordering types like floats.
This is pedantic! But that is what lint work to be, IMO.

LikeLakers2: https://github.com/bevyengine/bevy/pull/17335/commits/727c4a717da2ea2c41f68cf98732ec9a1b09926a

crates/bevy_ecs/src/archetype.rs
```diff
@@ -786,12 +786,7 @@ pub struct Archetypes {
 }
 
 /// Metadata about how a component is stored in an [`Archetype`].
-pub struct ArchetypeRecord {
-    /// Index of the component in the archetype's [`Table`](crate::storage::Table),
-    /// or None if the component is a sparse set component.
-    #[allow(dead_code)]
-    pub(crate) column: Option<usize>,
-}
+pub struct ArchetypeRecord {}
```
alice-i-cecile, line 789: This is a really messy state. I'd rather keep the dead code around, with an expect that links to an issue for what should be done to actually use it.

alice-i-cecile (changes requested):
Good stuff! I like the improved clarity here, and your comments are great. I do think we should revert the `ArchetypeRecord` change for this PR though: it's too much for a lint PR and I would rather not have to redo that work. If needed, link to this PR for future plans if no issue exists.

alice-i-cecile (approved):

PR #17336, opened 2025-01-13
Fix docs mistake in bevy_ecs::world

berylllium:
# Objective

- Correct a mistake in the rustdoc for bevy_ecs::world::World.

## Solution

- The rustdoc wrongly stated that "Each component can have up to one instance of each component type.". This sentence should presumably be "Each *Entity* can have up to one instance of each component type.". Applying this change makes the prior sentence "Each [`Entity`] has a set of components." redundant.


crates/bevy_ecs/src/world/mod.rs
```diff
@@ -73,9 +73,8 @@ use unsafe_world_cell::{UnsafeEntityCell, UnsafeWorldCell};
 /// Stores and exposes operations on [entities](Entity), [components](Component), resources,
 /// and their associated metadata.
 ///
-/// Each [`Entity`] has a set of components. Each component can have up to one instance of each
-/// component type. Entity components can be created, updated, removed, and queried using a given
-/// [`World`].
+/// Each [`Entity`] can have up to one instance of each type of component. Entity
+/// components can be created, updated, removed, and queried using a given [`World`].
```
LikeLakers2, line 77: Perhaps something like
> Each [`Entity`] has a set of components, but only one instance of each type of component.
so that it's clear that multiple types of components can exist, but only one of each type.

```suggestion
/// Each [`Entity`] has a set of components, but only one instance of each type of
/// component. Entity components can be created, updated, removed, and queried using
/// a given [`World`].
```

mockersf, line 77:
```suggestion
/// Each [`Entity`] has a set of unique components, based on their type.
/// Entity components can be created, updated, removed, and queried using a given
```
This feels clearer to me

berylllium: I agree.

mockersf (approved):

Carter0 (approved):

alice-i-cecile (approved):

PR #17338, opened 2025-01-13
Change Camera.viewport_to_world_2d to rescale according to the rendering target's scaling factor

### LINKED ISSUE #17312: Need to take scale factor into account when transforming Node to world position
## Bevy version

0.15, latest main

## What went wrong

1. Convert UI node position to world position using camera.viewport_to_world_2d()
2. Set an object in the world to that position, everything seems OK:

Image

3. Someone else runs the app with a scale factor other than 1.0 and sees this:

Image

## Additional information

Here is the system in question:
```rust
fn move_square_to_node(
    mut box_query: Single<&mut Transform, With<Mesh2d>>,
    text_query: Single<&GlobalTransform, With<Text>>,
    camera_query: Single<(&Camera, &GlobalTransform), With<Camera>>,
    window: Single<&Window>,
) {
    let (camera, camera_transform) = *camera_query;

    let Ok(point) = camera.viewport_to_world_2d(
        camera_transform,
        text_query.translation().xy(), // need to divide by window.scale_factor() here
    ) else {
        return;
    };

    box_query.translation = point.extend(0.0);
}
```
Full MRE here: https://gist.github.com/GitGhillie/44e02ab79f7e054ba939281b738fa10f

Example usecase: Player picks up a coin in the world, and to show this you create an animation where the coin moves from its position in the world to the coin icon in the UI. You want to keep it in the world as long as possible so you can apply effects etc while it is going to the UI position.

If I missed a better approach please let me know!

Suibeom:
Camera.viewport_to_world_2d doesn't behave as expected with the target window is rescaled, or didn't have a 1:1 scale factor in the first place. If the rendering target has enough information for us to get its scaling factor, we should use it to rescale the viewport coordinates before computing normalized device coordinates and then converting to world coordinates. This change should preserve the existing behavior if rendering target information is not available.

Fixes #17312

## Solution

Get the target_scaling_factor option from our camera and unwrap_or, with default 1.0, so we will use the scaling factor if it exists. It might be that we want to return an error if this option is `None`, but it didn't care before, so perhaps the safer change is to default to the previous behavior if we can't rescale in the first place.

## Testing

Issue reported included an MRE here:

[Issue MRE](https://gist.github.com/GitGhillie/44e02ab79f7e054ba939281b738fa10f)

This change successfully addresses the undesirable behavior in the MRE

Screenshot 2025-01-12 at 9 05 39 PM

---

## Migration Guide

If you were using Camera.viewport_to_world_2d previously and manually correcting for scaling, it will probably break!

### DIFF
--- a/crates/bevy_render/src/camera/camera.rs
+++ b/crates/bevy_render/src/camera/camera.rs
@@ -593,7 +593,7 @@ impl Camera {
 
         // Rescale by the rendering target's scaling factor if we have it.
         let scale_factor = self.target_scaling_factor().unwrap_or(1.);
-        viewport_position = viewport_position / scale_factor;
+        viewport_position /= scale_factor;
 
         // Flip the Y co-ordinate origin from the top to the bottom.
         viewport_position.y = target_size.y - viewport_position.y;
--- a/crates/bevy_render/src/camera/camera.rs
+++ b/crates/bevy_render/src/camera/camera.rs
@@ -590,6 +590,11 @@ impl Camera {
         let target_size = self
             .logical_viewport_size()
             .ok_or(ViewportConversionError::NoViewportSize)?;
+
+        // Rescale by the rendering target's scaling factor if we have it.
+        let scale_factor = self.target_scaling_factor().unwrap_or(1.);
+        viewport_position = viewport_position / scale_factor;
+
         // Flip the Y co-ordinate origin from the top to the bottom.
         viewport_position.y = target_size.y - viewport_position.y;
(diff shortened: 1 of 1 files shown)


rparrett:
This breaks even our own `2d_viewport_to_world` example, which expects logical coordinates. I think this method **should** take logical coordinates.

If there's anything to be done about #17312, maybe it's modifying the UI camera projection or whatever so that world coordinates match up with logical coordinates instead of physical coordinates.

PR #17339, opened 2025-01-13
Child spawning optimizations

### LINKED ISSUE #17301: Spawning child entities is extremely slow
## Bevy version

0.15, has existed for as long as constructing `Parent` and `Children` has been locked down.

## What you did

Attempt to optimize spawning large batches of children in a custom command for `big_space`.

## What went wrong

Profiling shows that applying system commands dominates system time. Spawning 320k entities took 31s.

The problem appears to be that even when spawning a brand new entity, bevy is doing reparenting checks for every child when its `Parent` component is added, and again for every child when the children are added to the parent entity's `Children` component. These checks are completely unnecessary - the entities are new and cannot yet exist in the hierarchy.

By doing some `transmute` shenanigans, I was able to see what a niave solution would look like if I could simply construct the `Parent` and `Children` entities myself. The performance improvement was stark: what previously took 31s now took 83ms.

I was also able to work around this *without* unsafe by using reflection to construct a `Parent` and `Children`, bypassing the private constructors.

aevyrie:
# Objective

- Fixes some of #17301 

## Solution

- Avoid re-parenting and related checks when spawning new entities, which cannot exist in the hierarchy yet.

## Testing

- Added a `spawn_children` benchmark

---

## Showcase

Fixed exponential time complexity when spawning children on an entity that already has existing children. This case can now complete in linear time.

```
spawn_children/10_entities
                        time:   [1.4008 µs 1.4195 µs 1.4392 µs]
                        change: [-7.2431% -5.7892% -4.2650%] (p = 0.00 < 0.05)
spawn_children/100_entities
                        time:   [10.150 µs 10.302 µs 10.461 µs]
                        change: [-22.516% -21.350% -20.218%] (p = 0.00 < 0.05)
spawn_children/1000_entities
                        time:   [88.756 µs 89.768 µs 90.867 µs]
                        change: [-75.726% -75.512% -75.309%] (p = 0.00 < 0.05)
spawn_children/10000_entities
                        time:   [912.39 µs 924.42 µs 936.40 µs]
                        change: [-96.697% -96.659% -96.618%] (p = 0.00 < 0.05)
spawn_children/100000_entities
                        time:   [9.3389 ms 9.3936 ms 9.4529 ms]
                        change: [-99.655% -99.653% -99.651%] (p = 0.00 < 0.05)
spawn_children/1000000_entities
                        time:   [111.12 ms 111.84 ms 112.58 ms]
```

Note the 1 million case takes 111ms on this branch. On `main`, I had to kill the test because it was estimated to take 8 hours.

### DIFF
--- a/crates/bevy_hierarchy/src/child_builder.rs
+++ b/crates/bevy_hierarchy/src/child_builder.rs
@@ -173,6 +173,7 @@ fn remove_children(parent: Entity, children: &[Entity], world: &mut World) {
 /// ```
 pub struct ChildBuilder<'a> {
     commands: Commands<'a, 'a>,
+    /// These children are guaranteed to be spawned entities, and do not need re-parenting checks.
     children: SmallVec<[Entity; 8]>,
     parent: Entity,
 }
@@ -213,13 +214,15 @@ impl ChildBuild for ChildBuilder<'_> {
         Self: 'a;
 
     fn spawn(&mut self, bundle: impl Bundle) -> EntityCommands {
-        let e = self.commands.spawn(bundle);
+        // Spawn the entity with the `Parent` component to avoid archetype changes.
+        let e = self.commands.spawn((bundle, Parent(self.parent)));
         self.children.push(e.id());
         e
     }
 
     fn spawn_empty(&mut self) -> EntityCommands {
-        let e = self.commands.spawn_empty();
+        // Spawn the entity with the `Parent` component to avoid archetype changes.
+        let e = self.commands.spawn(Parent(self.parent));
         self.children.push(e.id());
         e
     }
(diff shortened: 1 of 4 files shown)


crates/bevy_hierarchy/src/child_builder.rs
```diff
@@ -346,20 +349,37 @@ impl BuildChildren for EntityCommands<'_> {
...
         let children = builder.children;
-        if children.contains(&parent) {
-            panic!("Entity cannot be a child of itself.");
-        }
+
+        // Send hierarchy events
+        let events = children
+            .iter()
+            .map(|&child| HierarchyEvent::ChildAdded { child, parent })
+            .collect::<alloc::vec::Vec<_>>();
```
bushrat011899, line 358: Would prefer `use alloc::vec::Vec;` at the top of the file.

bushrat011899 (approved):
The rationale for your changes looks sound to me. Just a minor nit around `alloc::vec::Vec`.

BenjaminBrienen (approved):

Jondolf (approved):
Changes make sense and look solid to me.

aevyrie:
I have a few more changes locally I need to push. I noticed that some of the child methods send a hierarchy event, but some do not. I can't tell if this is an error or not. The docs don't really spell out any intent.

aevyrie:
Can anyone with more familiarity please comment on the `HierarchyEvent`s? It seems like it is outright missing in a few places but I'm not sure if this is a bug or not.

alice-i-cecile:
@aevyrie I think that this can be closed now that relations has landed :) (what a sentence to say)

PR #17343, opened 2025-01-13
Ambient component

### LINKED ISSUE #7193: Move AmbientLight to a component, instead of a resource.

robtfm:
# Objective

allow setting ambient light via component on cameras. 
arguably fixes #7193
note i chose to use a component rather than an entity since it was not clear to me how to manage multiple ambient sources for a single renderlayer, and it makes for a very small changeset.

## Solution

- make ambient light a component as well as a resource
- extract it
- use the component if present on a camera, fallback to the resource

## Testing

i added 
```rs
        if index == 1 {
            commands.entity(camera).insert(AmbientLight{
                color: Color::linear_rgba(1.0, 0.0, 0.0, 1.0),
                brightness: 1000.0,
                ..Default::default()                
            });
        }
```
at line 84 of the split_screen example

### DIFF
--- a/crates/bevy_pbr/src/light/ambient_light.rs
+++ b/crates/bevy_pbr/src/light/ambient_light.rs
@@ -4,6 +4,8 @@ use super::*;
 ///
 /// This resource is inserted by the [`PbrPlugin`] and by default it is set to a low ambient light.
 ///
+/// It can also be added to a camera to override the resource (or default) ambient for that camera only.
+/// 
 /// # Examples
 ///
 /// Make ambient light slightly brighter:
@@ -15,8 +17,8 @@ use super::*;
 ///    ambient_light.brightness = 100.0;
 /// }
 /// ```
-#[derive(Resource, Clone, Debug, ExtractResource, Reflect)]
-#[reflect(Resource, Debug, Default)]
+#[derive(Resource, Component, Clone, Debug, ExtractResource, Reflect)]
+#[reflect(Resource, Component, Debug, Default)]
 pub struct AmbientLight {
     pub color: Color,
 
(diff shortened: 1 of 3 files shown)


pcwalton (approved):
I didn't realize how small of a change this would be!

I'd be in favor of marking the resource as deprecated, but we can do that later.

mgi388 (approved):
I tested this out in my game and it works. It solves the use case I have: For my minimap camera, I want to be able to have different ambient light compared to my main world camera so that I can separately control the brightness of just the minimap view.

Here's my minimap camera being spawned, and with just the simple `AmbientLight` component addition it worked great.

```rust
const MAP_RENDER_LAYER: usize = 1;

  commands.spawn((
      Name::new("Minimap camera"),
      Camera3d::default(),
      Camera {
          is_active: false, // starts inactive
          order: -1,
          hdr: true,
          target: image_handle.clone().into(),
          ..default()
      },
      RenderLayers::from_layers(&[0, MAP_RENDER_LAYER]),
      AmbientLight { // <--- added this as a test for this PR
          color: WHITE.into(),
          brightness: 2000.0,
      },
  ));
```

As already said, I see there's room for future work here, but for me, the value is here already 🙏

IceSentry (approved):

mockersf (approved):

crates/bevy_pbr/src/light/ambient_light.rs
```diff
@@ -15,8 +17,9 @@ use super::*;
 ///    ambient_light.brightness = 100.0;
 /// }
 /// ```
-#[derive(Resource, Clone, Debug, ExtractResource, Reflect)]
-#[reflect(Resource, Debug, Default)]
+#[derive(Resource, Component, Clone, Debug, ExtractResource, ExtractComponent, Reflect)]
+#[reflect(Resource, Component, Debug, Default)]
+#[require(Camera)]
```
mockersf, line 22: Not sure I like requiring `Camera` here, but not a strong enough opinion to block

PR #17348, opened 2025-01-13
Unify picking backends

chompaa:
# Objective

Currently, our picking backends are inconsistent:

- Mesh picking and sprite picking both have configurable opt in/out behavior. UI picking does not.
  - Sprite picking uses `SpritePickingCamera` and `Pickable` for control, but mesh picking uses `RayCastPickable`.
- `MeshPickingPlugin` is not a part of `DefaultPlugins`. `SpritePickingPlugin` and `UiPickingPlugin` are.

## Solution

- Add configurable opt in/out behavior to UI picking (defaults to opt out).
- Replace `RayCastPickable` with `MeshPickingCamera` and `Pickable`.
- Remove `SpritePickingPlugin` and `UiPickingPlugin` from `DefaultPlugins`.

## Testing

Ran some examples.

## Migration Guide

`UiPickingPlugin` and `SpritePickingPlugin` are no longer included in `DefaultPlugins`. They must be explicitly added.

`RayCastPickable` has been replaced in favor of the `MeshPickingCamera` and `Pickable` components. You should add them to cameras and entities, respectively, if you have `MeshPickingSettings::require_markers` set to `true`.


alice-i-cecile:
@notmd @fjkorf can I get your reviews over here? I'd like to unify all this for 0.16.

fjkorf (approved):
These changes make sense to me. The naming is more consistent. I appreciate the extra comments scattered throughout. I'm not sure if it makes sense to leave picking out of the default plugins, but I don't think either option is bad. I was able to run this code locally and try the examples. All said this looks good to me.

crates/bevy_ui/src/picking_backend.rs
```diff
@@ -122,6 +168,10 @@ pub fn ui_picking(
             continue;
         };
 
+        if settings.require_markers && node.pickable.is_none() {
+            continue;
```
notmd, line 172: I think we should not make picking opt-out with `require_markers`. Maybe introduce a second flag or remove configurable opt-out. The `require_markers` should only use to config the camera, not the UI elements picking opt-in/out

chompaa: > Maybe introduce a second flag

To me, this would decrease readability for opt-in behavior. In my opinion, if you opt-in, you are opting to be explicit, so not having to add `UiPickingCamera` when all your UI needs `Pickable` seems off. Happy to hear your thoughts though :)

> [...] or remove configurable opt-out

To the best of my knowledge UI picking is going to drive a lot of the interactions going forward, so I think it makes sense to be opt-out by default (and consequently have the option to opt-in since there's no (?) perf loss).

notmd: Hi, sorry for not making it clear.  I mean a common scenario is you start with picking opt-out (all elements are pickable without inserting `Pickable`) and 1 camera. Then you start adding more cameras and now want to disable picking for them. You starting set make `require_markers` to true, but this will regret (even if you make your main camera to allow picking by inserting `UiPickingCamera`) all UI elements you have before because now it requires to explicit to insert `Pickable`. For UI, I only want to insert `Pickable` when I want to custom the behavior of `Pickable`. The only reason for UI picking is opt-in is performance, but I believe you will soon realize that the performance gain is not worth the DX.

fjkorf: I think renaming the flag inside of `UIPickingSettings` to something more explicit like `only_pickable` is reasonable.

That said the reason for making picking opt-in is performance, and that performance gain is worth a learning moment in the developer experience. The DX of updating your UI to reflect your picking preferences does not seem onerous.

alice-i-cecile: I think we should remove the ability to toggle opt-out behavior at the plugin level, and then use required components as needed to get "automatic picking". We don't need more mechanisms for this! I also think that the UiPickingPlugin should be added as part of UiPlugin.

chompaa: Right, that makes sense, I agree required components do solve the DX issue. How do we want to handle it for cameras though? I feel like having the plugin level toggle is nice in this case. I can see why `UiPickingPlugin` should be added as a part of `UiPlugin`, but I'm not sure I agree with `SpritePickingPlugin` being added by default. What's the consensus there? If we decide that `SpritePickingPlugin` be added by default then I think `MeshPickingPlugin` should be too.

alice-i-cecile: I think no `SpritePickingPlugin` by default; it's definitely feasible to write games that never use it. The same cannot be said for building UIs that don't!

Not sure on the camera config TBH 🤔

cart:
Yeah I think sprite picking and ui picking should be enabled by default.  Sprite picking has almost no cost associated with it and provides high utility, and UI picking is a core piece of the UI puzzle.

crates/bevy_ui/src/picking_backend.rs
cart: I think sprite picking and ui picking should be enabled by default. Sprite picking has almost no cost associated with it (per-entity opt-in, filtered out cheaply via archetype queries) and provides high utility, and UI picking is a core piece of the UI puzzle.

NthTensor (approved):

PR #17349, opened 2025-01-13
Fix `bevy_time` tests occasionally failing on optimised Windows builds

greeble-dev:
# Objective

Fix the `bevy_time` unit tests occasionally failing on optimised Windows builds.

# Background

I noticed that the `bevy_time` unit tests would fail ~50% of the time after enabling `opt-level=1` in config.toml, or adding `--release` to cargo test.

```
> cargo test -p bevy_time --release

thread 'real::test::test_update' panicked at crates\bevy_time\src\real.rs:164:9:
assertion `left != right` failed
  left: Some(Instant { t: 9458.0756664s })
 right: Some(Instant { t: 9458.0756664s })
 ```

Disabling optimisations would fix the issue, as would switching from Windows to Linux.

The failing path is roughly:

```rust
let mut time = Time::<Real>::new(Instant::now());
time.update();
time.update();
assert_ne!(time.last_update(), time.first_update());
```

Which kinda boils down to:

```rust
let left = Instant::now();
let right = Instant::now();
assert_ne!(left, right);
```

So the failure seems legit, since there's no guarantee that `Instant::now()` increases between calls.

I suspect it only triggers with a combination of Windows + fast CPU + optimisations (Windows has a lower resolution clock than Linux/MacOS). That would explain why it doesn't fail on the Bevy Github CI (optimisations disabled, and I'm guessing the runner CPUs are clocked lower).

# Solution

Make sure `Instant::now()` has increased before calling `time.update()`.

I also considered:

1. Change the unit tests to accept `Instant:now()` not increasing.
    - In retrospect this is maybe the better change?
    - There's other unit tests that cover time increasing.
    - Could also add a deterministic test for zero delta updates.
    - I can switch the PR to this if desired.
2. Avoid any paths that hit `Instant::now()` in unit tests.
    - Arguably unit tests should always be deterministic.
    - But that would mean a bunch of paths aren't tested.

## Testing

`cargo test -p bevy_time --release`

## System Info

`os: "Windows 10 Pro", kernel: "19045", cpu: "AMD Ryzen 9 7900 12-Core Processor", core_count: "12", memory: "63.2 GiB"`

Also tested on same computer with Linux pop-os 6.9.3.

### DIFF
--- a/crates/bevy_time/src/real.rs
+++ b/crates/bevy_time/src/real.rs
@@ -139,6 +139,18 @@ impl Time<Real> {
 mod test {
     use super::*;
 
+    // Waits until Instant::now() has increased.
+    //
+    // ```
+    // let previous = Instant::now();
+    // wait();
+    // assert!(Instant::now() > previous);
+    // ```
+    fn wait() {
+        let start = Instant::now();
+        while Instant::now() <= start {}
+    }
+
     #[test]
     fn test_update() {
         let startup = Instant::now();
@@ -150,6 +162,7 @@ mod test {
         assert_eq!(time.delta(), Duration::ZERO);
         assert_eq!(time.elapsed(), Duration::ZERO);
 
+        wait();
         time.update();
 
         assert_ne!(time.first_update(), None);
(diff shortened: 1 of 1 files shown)


alice-i-cecile (approved):

alice-i-cecile:
I like this solution: it's minimally intrusive and clearly explains why it's needed.

BenjaminBrienen (approved):

mockersf (approved):
that is sooo not pretty 😄

mockersf:
failure due to GitHub incident

PR #17352, opened 2025-01-13
`FromWorld` derive macro

ecoskey:
simple derive macro for `FromWorld`. Going to be needed for composable pipeline specializers but probably a nice thing to have regardless

## Testing

simple manual testing, nothing seemed to blow up. I'm no proc macro pro though, so there's a chance I've mishandled spans somewhere or something.


MrGVSV (commented):
Could you add some docs to `FromWorld` explaining how to use the derive macro?

crates/bevy_ecs/macros/src/lib.rs
```diff
@@ -611,3 +612,60 @@ pub fn derive_states(input: TokenStream) -> TokenStream {
...
+    let field_initializers: Punctuated<FieldValue, Token![,]> = match fields {
+        syn::Fields::Named(fields_named) => fields_named
+            .named
+            .iter()
+            .map(|field| FieldValue {
+                attrs: Vec::new(),
+                member: Member::Named(field.ident.clone().unwrap()),
+                colon_token: Some(Token![:](field.span())),
+                expr: field_init_expr.clone(),
+            })
```
MrGVSV, line 644: Nit: any reason to use `FieldValue` instead of quoting the ident/index and initializer directly?

ecoskey: No reason other than it felt more structured than a `Vec<TokenStream>`, but I can change it if that's more idiomatic

crates/bevy_ecs/src/world/mod.rs
```diff
@@ -18,6 +18,7 @@ pub use crate::{
     change_detection::{Mut, Ref, CHECK_TICK_THRESHOLD},
     world::command_queue::CommandQueue,
 };
+pub use bevy_ecs_macros::FromWorld;
```
MrGVSV, line 21: Nit: I think you can move this import into the doc example since that's where it's being used

MrGVSV (approved):

crates/bevy_ecs/src/world/mod.rs
ecoskey: No this is so the macro is publicly available

Carter0 (approved):

tychedelia (approved):

crates/bevy_ecs/macros/src/lib.rs
```diff
@@ -611,3 +611,46 @@ pub fn derive_states(input: TokenStream) -> TokenStream {
...
+pub fn derive_from_world(input: TokenStream) -> TokenStream {
+    let bevy_ecs_path = bevy_ecs_path();
+    let ast = parse_macro_input!(input as DeriveInput);
+    let struct_name = ast.ident;
+    let (impl_generics, ty_generics, where_clauses) = ast.generics.split_for_impl();
+
+    let Data::Struct(DataStruct { fields, .. }) = &ast.data else {
+        return syn::Error::new(
+            Span::call_site(),
+            "#[derive(FromWorld)]` only supports structs",
```
MrGVSV, line 625: Totally not required for this PR but one thing that just occurred to me is that we could add a `#[from_world]` helper that works sorta like `Default`'s `#[default]`, allowing for enums to utilize this derive macro too.

ecoskey: ooh good idea. Might need to rework the internals for that a bit, so might be good for a followup :)

PR #17354, opened 2025-01-13
Diagnostics smoothing factor fix

ickshonpe:
# Objective

Diagnostics is reporting incorrect FPS and frame time.

## Solution

Looks like the smoothing value should be `2 / (history_length + 1)` not `(history_length + 1) / 2`.


rparrett (approved):
Yep, oops.
https://github.com/bevyengine/bevy/blob/141b7673ab3b0af13d49349b97f053286bb16227/crates/bevy_diagnostic/src/diagnostic.rs#L179)

mockersf:
CI failure because of GitHub incident, retriggering CI

mockersf (approved):

PR #17359, opened 2025-01-14
Group rendering crates in common subfolder

ecoskey:
discord convo: https://discord.com/channels/691052431525675048/743663924229963868/1328606441350627329

@IceSentry's idea, don't kill me please 🙏

# Objective

- group the rendering crates together so it's either to move between them when working. 

## Migration Guide

There's no way this gets merged, but if so I'm sorry to anybody working on a rendering pr. I wouldn't submit this other than the fact we want this in some form at some point probably.


BenjaminBrienen:
I like how this makes the feature area more obvious. The problem is that maybe we consider something fringe like bevy_image or bevy_mesh to be its own area in the future.

mockersf:
I don't like it, and I don't think it can that be clear cut. gltf, image and mesh are assets. mikktspace is maths. encase_derive and render are crate glue. pbr is physics (ok that one is far-fetched, but it actually has physic in the name)

also you would have to change the dependency paths on all the crates depending on the ones you move, some of our tools would have to be fixed

ecoskey:
Agree to both. Gonna close this out since any actual reorganization will probably need a working group :). This wasn't the most serious pr, as you could probably tell lol

PR #17370, opened 2025-01-14
docs: enhance documentation in `query.rs` to clarify borrowing rules

younes-io:
docs: enhance documentation in `query.rs` to clarify borrowing rules.

Please, let me know if you don't agree with the wording.. There is always room for improvement.

Tested locally and it looks like this:

image

### DIFF
--- a/crates/bevy_ecs/src/system/query.rs
+++ b/crates/bevy_ecs/src/system/query.rs
@@ -229,7 +229,7 @@ use core::{
 /// // This will panic!
 /// // EntityRef provides read access to ALL components on an entity.
 /// // When combined with &mut ComponentA in the same query, it creates
-/// // a conflict because EntityRef could read ComponentA while the &mut 
+/// // a conflict because EntityRef could read ComponentA while the &mut
 /// // attempts to modify it - violating Rust's borrowing rules of no
 /// // simultaneous read+write access.
 /// query: Query<(EntityRef, &mut ComponentA)>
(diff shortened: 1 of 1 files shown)


crates/bevy_ecs/src/system/query.rs
```diff
@@ -239,11 +244,16 @@ use core::{
...
+/// // This creates a perfect separation where:
+/// // 1. First query reads entities that have ComponentA
+/// // 2. Second query modifies ComponentB only on entities that DON'T have ComponentA
+/// // Result: No entity can ever be accessed by both queries simultaneously
 /// query_a: Query<EntityRef, With<ComponentA>>,
 /// query_b: Query<&mut ComponentB, Without<ComponentA>>,
 /// # ) {}
 /// # bevy_ecs::system::assert_system_does_not_conflict(system);
 /// ```
+/// The core principle is immutable: `EntityRef` can read everything on an entity, so it must be completely separated from any mutable access. The `With`/`Without` filters achieve this by ensuring the queries operate on entirely different sets of entities.
```
alice-i-cecile, line 256: "The core principle is immutable" doesn't make sense to me. Can you try rephrasing? Rest looks good.

younes-io: done

alice-i-cecile (approved):

Carter0 (approved):

alice-i-cecile:
This needs formatting before CI will let me merge it @younes-io :)

BenjaminBrienen (approved):

PR #17371, opened 2025-01-14
Add `any_filtered` condition

JeanMertz:
This adds an `any_filtered` condition that returns `true` if there are any entities matching the given filter.

Example:

```rust
my_system.run_if(any_filtered::<Changed<A>>)
```

I often implement this method in my own projects, so I figured I'd upstream it, in case others have a use for it.


alice-i-cecile:
Duplicate of #17327 :)

PR #17373, opened 2025-01-14
Composable Pipeline Specialization

ecoskey:
Currently, our specialization API works through a series of wrapper structs and traits, which make things confusing to follow and difficult to generalize. 

This pr takes a different approach, where "specializers" (types that implement `Specialize`) are composable, but "flat" rather than composed of a series of wrappers. The key is that specializers don't *produce* pipeline descriptors, but instead *modify* existing ones: 

```rs
pub trait Specialize<T: Specializable> {
    type Key: SpecializeKey;
    
    fn specialize(
        &self, 
        key: Self::Key, 
        descriptor: &mut T::Descriptor
    ) -> Result<Canonical<Self::Key>, BevyError>;
}
```

This lets us use some derive magic to stick multiple specializers together: 

```rs
pub struct A;
pub struct B;

impl Specialize<RenderPipeline> for A { ... }
impl Specialize<RenderPipeline> for A { ... }

#[derive(Specialize)]
#[specialize(RenderPipeline)]
struct C {
    // specialization is applied in struct field order
    applied_first: A,
    applied_second: B,
}

type C::Key = (A::Key, B::Key);

```

This approach is much easier to understand, IMO, and also lets us separate concerns better. Specializers can be placed in fully separate crates/modules, and key computation can be shared as well.

The only real breaking change here is that since specializers only modify descriptors, we need a "base" descriptor to work off of. This can either be manually supplied when constructing a `Specializer` (the new collection replacing `Specialized[Render/Compute]Pipelines`), or supplied by implementing `HasBaseDescriptor` on a specializer. See `examples/shader/custom_phase_item.rs` for an example implementation.

## Testing

- Did some simple manual testing of the derive macro, it seems robust.

---

## Showcase

```rs
#[derive(Specialize, HasBaseDescriptor)]
#[specialize(RenderPipeline)]
pub struct SpecializeMeshMaterial<M: Material> {
    // set mesh bind group layout and shader defs
    mesh: SpecializeMesh,
    // set view bind group layout and shader defs
    view: SpecializeView,
    // since type SpecializeMaterial::Key = (), 
    // we can hide it from the wrapper's external API
    #[key(default)]
    // defer to the GetBaseDescriptor impl of SpecializeMaterial, 
    // since it carries the vertex and fragment handles
    #[base_descriptor]
    // set material bind group layout, etc
    material: SpecializeMaterial<M>,
}

// implementation generated by the derive macro
impl <M: Material> Specialize<RenderPipeline> for SpecializeMeshMaterial<M> {
    type Key = (MeshKey, ViewKey);

    fn specialize(
        &self, 
        key: Self::Key, 
        descriptor: &mut RenderPipelineDescriptor
    ) -> Result<Canonical<Self::Key>, BevyError>  {
        let mesh_key = self.mesh.specialize(key.0, descriptor)?;
        let view_key = self.view.specialize(key.1, descriptor)?;
        let _ = self.material.specialize((), descriptor)?;
        Ok((mesh_key, view_key));
    }
}

impl <M: Material> HasBaseDescriptor<RenderPipeline> for SpecializeMeshMaterial<M> {
    fn base_descriptor(&self) -> RenderPipelineDescriptor {
        self.material.base_descriptor()
    }
}
```


crates/bevy_render/macros/src/specialize.rs
```diff
@@ -0,0 +1,439 @@
...
+    }
+
+    if used_count == 1 {
+        field_info[single_index].key = Key::Whole;
+    }
+
+    Ok(field_info)
+}
+
+fn get_struct_fields<'a>(ast: &'a DeriveInput, derive_name: &str) -> syn::Result<&'a Fields> {
```
Bleachfuel, line 195: u can use bevy_macro_utils::get_struct_fields. Probably also for more things here.

ecoskey: good point. I used a custom version originally for better error messages, but I should probably just upstream those changes to `get_struct_fields`

Bleachfuel (approved):

tychedelia (commented):
How does this account for `SpecializedMeshPipeline` which needs a `MeshVertexBufferLayoutRef` that's derived from the mesh that's currently being specialized? This is the problem that I ran into when trying to make something similarly generic. Maybe we could have some kind of "extra data" that could piggyback along with the key? Or just add an extra parameter slot / associated type that's `()` for most implementations?

ecoskey:
Very good point. I think ideally we'd include it as part of the normal key, but it's true that it's not `Hash`. Could we add a wgpu style global id to `MeshVertexBufferLayoutRef`, and accept that if users don't handle things right they might have duplicate pipelines? 

I think @superdump also mentioned that pipelines layouts and mesh vertex buffer layouts don't have to be strictly equal? If so, I'm not sure how we handle everything generically. I'd rather not add another associated type for the layout (especially if it makes the derive macro more hellish) but if we have to I guess that's fine.

ecoskey:
Note: blocked on mesh rework, or at least a small part of it. This pr needs vertex attributes to be pre-processed, small, and hashable before they reach the specializer

JMS55 (approved):
I can't comment on the macro code, but I'm very much onboard with the idea.

This has the potential to massively cleanup the amount of specialization code scattered around main/prepass/deferred/meshlet/shadows specialize functions.

We could extract the repetitive parts into a common base specializer.

tychedelia (approved):
This makes sense to me and helps clean up our existing specialization traits which are particularly hard to be generic over.

release-content/migration-guides/composable_specialization.md
```diff
@@ -0,0 +1,153 @@
...
+        descriptor.fragment.targets[0].as_mut().unwrap().blend_mode = key.blend_state;
+        Ok(key)
+    }
+}
+
+impl GetBaseDescriptor for MySpecializer {
+    fn get_base_descriptor(&self) -> RenderPipelineDescriptor {
+        RenderPipelineDescriptor {
+            label: Some("my_pipeline".into()),
+            layout: vec![self.layout.clone()],
```
atlv24, line 126: does this need to be self.layout or can it just be some defaulted value? (or maybe even empty vec and have specialize insert-or-update?

release-content/migration-guides/composable_specialization.md
```diff
@@ -0,0 +1,153 @@
...
+                entry_point: "vertex".into(),
+                buffers: vec![],
+            },
+            primitive: Default::default(),
+            depth_stencil: None,
+            multisample: MultiSampleState::default(),
+            fragment: Some(FragmentState {
+                shader: self.fragment.clone(),
+                shader_defs: vec![],
+                entry_point: "fragment".into(),
```
atlv24, line 140: i feel like the base descriptor should probably be passed in a fully defaulted value and this should only have to set a handful of things on it. Especially once we have wgpu automatic entrypoint selection, it would be really nice to unify base and specialize so that you just write one type of function. you'd only have to set the frag and vert shader, and the layout. down the line, potentially even the layout could be introspected from the shader and you wouldnt have to set that either

atlv24 (approved):
i agree with the overall direction of this

superdump:
From the high level description, I’m on board with this. I feel like the naming is a bit off. SpecializeMeshPipeline for example does feel like it’s missing a d which would make it the same as the existing name. I’d say make the names feel right, break APIs, and document migrations.

release-content/migration-guides/composable_specialization.md
tychedelia: This also came to mind when I was using this PR on a branch. There isn't actually that much you can usefully set on a "base" descriptor besides perhaps a label, but that's also easy to set elsewhere as well.

crates/bevy_render/src/render_resource/specialize.rs
```diff
@@ -0,0 +1,401 @@
...
+/// In this case, "canonical" means that each unique value of this type will produce
+/// a unique specialized result, which isn't true in general. `MeshVertexBufferLayout`
+/// is a good example of a type that's `Eq + Hash`, but that isn't canonical: vertex
+/// attributes could be specified in any order, or there could be more attributes
+/// provided than the specialized pipeline requires. Its `Canonical` key type would
+/// be `VertexBufferLayout`, the final layout required by the pipeline.
+///
+/// Processing keys into canonical keys this way allows the `Specializer` to reuse
+/// resources more eagerly where possible.
+pub trait SpecializeKey: Clone + Hash + Eq {
```
superdump: SpecializationKey?

crates/bevy_render/src/render_resource/specialize.rs
```diff
@@ -0,0 +1,401 @@
...
+    /// Denotes whether this key is canonical or not. This should only be `true`
+    /// if and only if `Canonical = Self`.
+    const IS_CANONICAL: bool;
```
superdump: Base key? This is false for compound keys that combine multiple specialisation keys?

superdump: Hmm. No, I see that base is used for the base pipeline descriptor but this could be a key corresponding to one specialisation layer. Simple key vs compound key?

ecoskey: It's not whether it's simple/compound, more like the distinction between MeshVertexLayoutRef and VertexBufferLayout. The docs explain it a bit, but maybe I should expand them?

release-content/migration-guides/composable_specialization.md
tychedelia: In fact, I kind of feel like this is the right call and would help simplify the API. Thoughts, @ecoskey?

release-content/migration-guides/composable_specialization.md
ecoskey: Either would work, I think having a set of both placeholder values and wrapped accessors is needed. I was gonna experiment with that in an immediate followup, but I can do that in this pr if that's better.

release-content/migration-guides/composable_specialization.md
ecoskey: Composing specializers is the main reason I didn't do this. IMO having nice composition means that specializers should overall be really minimal and only adapt pipelines to fit their keys, and only rarely set anything that would otherwise be static. `GetBaseDescriptor` (which encourages putting all the static handles and such on the specializer) is meant mostly for convenience with monolithic specializers.

Also, maybe this doesn't apply for materials (whose keys override a lot of the descriptor) but a lot of the other engine specializers can put most stuff in the `base_descriptor`, and keep the actual specializer really simple.

ecoskey: This is kind of also why I chose `Specializer` over the more passive `SpecializedPipeline`. Rather than being the pipeline itself, they're things that act on pipelines.

PR #17376, opened 2025-01-14
Change `World::try_despawn` and `World::try_insert_batch` to return `Result`

JaySpruce:
## Objective

Most `try` methods on `World` return a `Result`, but `try_despawn` and `try_insert_batch` don't. Since Bevy's error handling is advancing, these should be brought in line.

## Solution

- Added `TryDespawnError` and `TryInsertBatchError`.
- `try_despawn`, `try_insert_batch`, and `try_insert_batch_if_new` now return their respective errors.
- Fixed slightly incorrect behavior in `try_insert_batch_with_caller`.
  - The method was always meant to continue with the rest of the batch if an entity was missing, but that only worked after the first entity; if the first entity was missing, the method would exit early. This has been resolved.

## Migration Guide
- `World::try_despawn` now returns a `Result` rather than a `bool`.
- `World::try_insert_batch` and `World::try_insert_batch_if_new` now return a `Result` where they previously returned nothing.


crates/bevy_ecs/src/world/mod.rs
```diff
@@ -2526,99 +2544,126 @@ impl World {
...
-                        insert_mode,
-                        #[cfg(feature = "track_location")]
-                        caller,
-                    )
-                };
+                    };
+                    break Some(cache);
+                }
+                // This is effectively in an `else` block, since it can only be reached if the
+                // above `if` fails.
```
BenjaminBrienen, line 2617:
```suggestion
```

crates/bevy_ecs/src/world/error.rs
```diff
@@ -15,6 +16,22 @@ use crate::{
 #[error("The schedule with the label {0:?} was not found.")]
 pub struct TryRunScheduleError(pub InternedScheduleLabel);
 
+/// The error type returned by [`World::try_despawn`] if the provided entity does not exist.
+///
+/// [`World::try_despawn`]: crate::world::World::try_despawn
+#[derive(Error, Debug, Clone, Copy)]
+#[error("Could not despawn the entity with ID {0} because it {1}")]
+pub struct TryDespawnError(pub Entity, pub EntityDoesNotExistDetails);
```
alice-i-cecile, line 24: Named structs instead of tuple structs please. Operations on tuple structs are much less clear, so we avoid them as a matter of style unless they're 1-element newtypes.

crates/bevy_ecs/src/world/mod.rs
```diff
@@ -2332,7 +2354,7 @@ impl World {
     ///
     /// This function will panic if any of the associated entities do not exist.
     ///
-    /// For the non-panicking version, see [`World::try_insert_batch`].
+    /// For the fallible version, see [`World::try_insert_batch`].
     #[track_caller]
     pub fn insert_batch<I, B>(&mut self, batch: I)
```
alice-i-cecile, line 2359: I would prefer to merge all of the fallible / infallible variants, but that should be done in follow-up.

alice-i-cecile (approved):

alice-i-cecile:
@JaySpruce let me know when merge conflicts are resolved here please <3

JaySpruce:
@alice-i-cecile, good to go

PR #11349, opened 2024-01-15
Asset Error Events & Load Retry Plugin

brianreavis:
## Objective

This PR introduces public-facing asset failure events (#11288). It also includes an out-of-the-box plugin for retrying failed assets, which is extremely useful if you have assets coming over a network. It's configured for the WASM asset reader by default, but can easily support third-party asset readers (like [bevy_web_asset](https://github.com/johanhelsing/bevy_web_asset))
### Asset Error Events

* **Added  `AssetLoadFailedEvent<A>`** – I started implementing `AssetEvent::Failed` like suggested in #11288, but decided it was better as its own type because:
	* I think it makes sense for `AssetEvent` to only refer to assets that actually exist. 
	* In order to return `AssetLoadError` in the event (which is useful information for error handlers that might attempt a retry) we would have to remove `Copy` from `AssetEvent`.
	* There are numerous places in the render app that match against `AssetEvent`, and I don't think it's worth introducing extra noise about assets that don't exist.
* **Added `UntypedAssetLoadErrorEvent`** – Useful in places that need to support type flexibility, like the asset retry plugin.

### Asset Retry Plugin

This retry plugin in this PR is more opinionated than the error events and can be split out if desired. I found that writing a retry plugin couldn't be done in user space without small adjustments to `bevy_asset`, like making HTTP errors first-class citizens.

#### Usage
The `AssetLoadRetryPlugin` is added to the `DefaultPlugins`, so it will work out-of-the box. To get retries to happen, you must use an `AssetReader` that provides `get_retry_settings` (like `HttpWasmAssetReader`), or provide your own source or asset-specific settings via the `AssetLoadRetrier` resource.
##### Retry Settings By Source
```rust
let mut retrier = app.world.resource_mut::<AssetLoadRetrier>();
retrier.set_source_settings(
    AssetSourceId::Default,
    IoErrorRetrySettingsProvider::from(AssetLoadRetrySettings {
        max_attempts: 5,
        max_delay: Duration::from_millis(5000),
        starting_delay: Duration::from_millis(100),
        time_multiple: 2.0,
    }),
);
```
##### Retry Settings By Asset Type
```rust
let mut retrier = app.world.resource_mut::<AssetLoadRetrier>();
// Retry images with their own interval
retrier.set_asset_settings::<Image>(IoErrorRetrySettingsProvider::from(AssetLoadRetrySettings {
    max_attempts: 1,
    max_delay: Duration::from_millis(1000),
    starting_delay: Duration::from_millis(1000),
[...]

PR #17377, opened 2025-01-15
fix: limit the number of concurrent load_internal calls to avoid exce…

### LINKED ISSUE #14542: On iOS, loading many preprocessed assets causes some assets to fail to load
## Bevy version

0.14.0

iOS 17.5.1

## What you did

Loaded 128 images all at once with `AssetMode::Processed` in iOS. The images aren't preprocessed, they're just configured to be copied into `imported_assets`.

## What went wrong

5 of the assets failed to load with `[2m2024-07-30T19:31:05.814567Z[0m [31mERROR[0m [2mbevy_asset::server[0m[2m:[0m Encountered an I/O error while loading asset: Too many open files (os error 24)`.

## Additional information

I tracked the loading states of the asset with logging:
```
[2m2024-07-30T19:31:33.444959Z[0m [32m INFO[0m [2mbevy_mobile_example[0m[2m:[0m == LOAD STATES ==
[2m2024-07-30T19:31:33.444966Z[0m [32m INFO[0m [2mbevy_mobile_example[0m[2m:[0m no_state   0
[2m2024-07-30T19:31:33.444970Z[0m [32m INFO[0m [2mbevy_mobile_example[0m[2m:[0m not_loaded 0
[2m2024-07-30T19:31:33.444975Z[0m [32m INFO[0m [2mbevy_mobile_example[0m[2m:[0m loading    0
[2m2024-07-30T19:31:33.444980Z[0m [32m INFO[0m [2mbevy_mobile_example[0m[2m:[0m loaded     123
[2m2024-07-30T19:31:33.444984Z[0m [32m INFO[0m [2mbevy_mobile_example[0m[2m:[0m failed     5
```

This issue doesn't occur in `AssetMode::Unprocessed` with 512 images. I haven't tested with more than that. I suspect there's already a mechanism in Bevy that prevents too many assets from loading all at once, but it doesn't take into account `.meta` files.

Behold, a repro: https://github.com/Seldom-SE/testetst/blob/279fa27a26bf0b845a5558a815c1f8d031c26fa2/src/lib.rs It's based on `bevy_mobile_example`. To run it:

1. Make sure you're on the linked commit
2. `./copy_assets.sh`. This just copies the one image 127 times.
3. `cargo run --features preprocess`
4. Configure the Xcode project to use your certificate. It might just do this automatically.
5. Run from Xcode (or with the Makefile, but I haven't tested that)

Braymatter:
fix: limit the number of concurrent load_internal calls to avoid exceeding the OS file descriptor limit
# Objective

Fixes #14542 

## Solution

Added a Semaphore to the AssetServer that load_internal blocks on, preventing the app from requesting too many file descriptors from the OS. 

## Testing

Tested using the repro in #14542 with 1500 images, and 1500 glb's with asset preprocessing enabled. On IOS at a permit-limit of 127 experienced zero failures repeatedly on an iPhone 16 physical device.  

## Migration Guide
I don't think there are any breaking changes (unsure if adding a field to a struct counts)


crates/bevy_asset/src/server/mod.rs
```diff
@@ -112,6 +116,13 @@ impl AssetServer {
         let (asset_event_sender, asset_event_receiver) = crossbeam_channel::unbounded();
         let mut infos = AssetInfos::default();
         infos.watching_for_changes = watching_for_changes;
+
+        #[cfg(target_os = "ios")]
+        let file_limit = 127; // The normal limit is 256, cut in half for .meta files and sub 1 because 128 still throws the occasional error (3 failed files out of 1500)
+
+        #[cfg(not(target_os = "ios"))]
+        let file_limit = 16000;
```
BenjaminBrienen, line 124: Source?

Braymatter: Not much:
https://stackoverflow.com/questions/57007894/permanently-increase-file-descriptor-ulimit-in-android-without-being-root

Seemed like a reasonable limit, considering it wasn't limited at all before - it's going to vary with platform OS and I didn't find exhaustive documentation on specific limits for every OS and distro. 

I'm not very concerned as I've never run into this problem on a desktop machine and my projects are bigger than most.

JMS55:
Do we not retry failed asset loads? I would assume they would fail, back off, and then retry.

crates/bevy_asset/src/server/mod.rs
BenjaminBrienen: In terms of correctness, these values seem very magic and arbitrary. They seem like they could change at any time or even by configurable by the gamer. Is there a way to instead detect whether the limit has been reached when performing the operation instead of assuming what the limit is at compile-time?

BenjaminBrienen: These values should have a primary source linked in a comment. If there is an authoritative resource, maaaayybe we can rely on them.

Braymatter: In general I seriously doubt that anyone is changing the file descriptor limit on ios. The ios limit of 256 is somewhat well documented online, but can always change. I ran ulimit -n on my macos machine and got 2560, which is where the other value came from. 

16000 as a limit was a wild-guess. 

I was unable to find a cross platform, exhaustive way to detect file system limits. As you mentioned they can change arbitrarily. If anything I would set it to 127 for every platform as most apps aren't even going to have that many assets, and the ones that do will likely have a negligible/user undetectable increase in load times.

Braymatter:
@JMS55 - My projects use bevy_asset_loader so I'm not entirely sure. I didn't see any retry logs etc. in my main project, it would just fail and never transition to the next state. I don't recall seeing any retry logging in the repro etc. today either.

crates/bevy_asset/src/server/mod.rs
```diff
@@ -112,6 +116,16 @@ impl AssetServer {
         let (asset_event_sender, asset_event_receiver) = crossbeam_channel::unbounded();
         let mut infos = AssetInfos::default();
         infos.watching_for_changes = watching_for_changes;
+
+        #[cfg(target_os = "ios")]
+        let file_limit = 255; // The normal limit is 256, cut in half for .meta files and sub 1 because 128 still throws the occasional error (3 failed files out of 1500)
+
+        #[cfg(target_os = "macos")]
+        let file_limit = 2559;
```
alice-i-cecile, line 124: Link / explanation of where these numbers came from please!

JMS55:
Looks like we do this in a test https://github.com/bevyengine/bevy/blob/26bb0b40d26ee93a8938986ed68a58a897a0e166/crates/bevy_asset/src/lib.rs#L1644, but don't actually have it hooked up...

crates/bevy_asset/src/server/mod.rs
```diff
@@ -61,6 +62,9 @@ pub(crate) struct AssetServerData {
     sources: AssetSources,
     mode: AssetServerMode,
     meta_check: AssetMetaCheck,
+
+    ///Used to ensure the `asset_server` does not try to acquire more loaders (and thus `file_handles`) than the OS allows
+    asset_counter: Semaphore,
```
cart, line 67: This approach looks good to me, but I'm thinking we should scope this to FileAssetReader instead, as this is a "file descriptor" limitation protection? We wouldn't want this limit on an in-memory asset source

Braymatter: My initial instinct was to keep in in file_asset as well, but I couldn't think of a way to detect when the file was actually closed/dropped, as the AssetReader returns the File back to the caller. 

If I were to add the Semaphore to the AssetReader impl file_asset.rs it would release the permit as soon as it loaded the file and returned, and not after it was finished being read/processed etc.  My understanding is that the host OS will keep the file descriptor allocated until the process closes (or in Rust's case drops) the file. 

I could definitely be misreading/misunderstanding something so open to suggestions.

crates/bevy_asset/src/server/mod.rs
Braymatter: Updated, and added what I was able to find. I searched around quite a bit in an effort to come up with something better but docs are very scarce because the limit is (very) changeable per platform. 

We could potentially invoke `ulimit -n` on startup, but that didn't seem like a very good solution.

Braymatter:
@JMS55 - That would be very nice to have hooked up! I think in any case we probably don't want to thrash the file system more than we have to so some limit is probably good to have in place. This would be a great fallback for platforms / systems where the file descriptor limit is non-standard / unexpectedly low

alice-i-cecile:
> Do we not retry failed asset loads? I would assume they would fail, back off, and then retry.

We currently do not.

crates/bevy_asset/src/server/mod.rs
Braymatter: Some previous discussion here:
https://discord.com/channels/691052431525675048/749332104487108618/1328572829930881066

We could read the limit from a package.metadata parameter and leave it to the client application developers to set, while defaulting to u32::Max or something sensible. 

It would then be opt-in, and wouldn't affect network/memory etc. AssetReaders by default. The downside being client app developers would have to use the lowest limit of any of their target platforms.

crates/bevy_asset/src/server/mod.rs
BenjaminBrienen: ulimit could be changed during the middle of asset loading

crates/bevy_asset/src/server/mod.rs
cart: In theory we could implement the `Reader` trait for a wrapper containing both a semaphore handle and the File, then return that.

cart: Idk if the async_lock semaphore impl will meet the send / sync  requirements there.

crates/bevy_asset/src/server/mod.rs
```diff
@@ -112,6 +116,32 @@ impl AssetServer {
...
+        let file_limit = 127; // The normal limit is 256, cut in half for .meta files and sub 1 because 128 still throws the occasional error (3 failed files out of 1500)
+
+        /*
+        https://krypted.com/mac-os-x/maximum-files-in-mac-os-x/
+        Running `ulimit -n` on a MBP M3-Max yields 2560. In empirical testing when using the exact limit
+        some failures would still squeak through. This also leaves a small amount of headroom for direct
+        std::fs calls by the client application
+        */
+        #[cfg(target_os = "macos")]
+        let file_limit = 1279;
```
Bleachfuel, line 135: why are these not just CONST?

crates/bevy_asset/src/server/mod.rs
Braymatter: I took a stab at it here: https://github.com/Braymatter/bevy_filecount_fix/commit/a415213b062b3da731cff657672473bfd5f7679a

It... compiles but the Semaphores are no longer preventing too many file handles from being opened. Interestingly it seems like the app is provisioning two FileAssetReaders based on this screenshot / log showing up twice in the xcode debugger:

image

I think this would actually be a problem anyway because you may want to load assets from the normal assets/imported_assets AND load downloaded assets from (in my case) the App sandbox fs. 

There's also the issue of the AssetWriter also potentially reserving file descriptors. 

We could do some kind of global/shared semaphore, but I could also be misunderstanding how some of this fits together. 

Open to more suggestions - I'm unsure why this isn't working; I would expect that halving the amount of Semaphore permits would resolve the issue with two asset loaders but it doesn't seem to be the case.

Braymatter:
On further reflection - I think we should expose the limit for concurrent asset loads as a package.metadata field if possible, and keep the initial implementation of this. Since the file descriptor limit is per-process (and not per-bevy-asset) we should let the user set a limit themselves in case they're doing odd things with the filesystem directly via std::fs.

crates/bevy_asset/src/server/mod.rs
mockersf, line 135: this limit is at 256 on my mac (m4 max).

crates/bevy_asset/src/server/mod.rs
```diff
@@ -112,6 +116,32 @@ impl AssetServer {
...
+
+        /*
+        https://docs.pingidentity.com/pingdirectory/latest/installing_the_pingdirectory_suite_of_products/pd_ds_config_file_descriptor_limits.html#:~:text=Many%20Linux%20distributions%20have%20a,large%20number%20of%20concurrent%20connections.
```
mockersf, line 138: not a fan of this source of truth, but couldn't find a better one...

crates/bevy_asset/src/server/mod.rs
Braymatter: Ah! That's weird and disappointing. 

image
Mine is 2560. 

If there are no objections, I'll go ahead and implement [this](https://github.com/bevyengine/bevy/pull/17377#issuecomment-2596135239) idea, and add 256 as the default limit for osx/ios if that package.metadata field isn't present. 

I think long term we likely want to have some level of retry logic, and it would be good to find a reliable way to identify the descriptor limit at runtime (in a perfect world). 

For my app we worked around this by using multiple loading states, which works, even if inelegant.

mockersf:
https://github.com/GuillaumeGomez/sysinfo/issues/1202 would help with getting the actual max value

crates/bevy_asset/src/server/mod.rs
Braymatter: I searched around for about an hour trying to find some reasonably 'verified' information, but came to the conclusion that it's so configurable by distro/os etc. it really can be all over the place. Identifying the limit at runtime is probably the only way to be reliable.

alice-i-cecile:
@Braymatter, I'd like to get this in for 0.16; can you clean this up when you have a bit?

Braymatter:
> @Braymatter, I'd like to get this in for 0.16; can you clean this up when you have a bit?

I'll take a look this weekend

mockersf:
the limits are now available in sys info: https://github.com/GuillaumeGomez/sysinfo/pull/1480

Braymatter:
Closing this PR as the issue is addressable using this plugin when (once merged):

https://github.com/bevyengine/bevy/pull/11349

mockersf:
this PR is a better fix for the issue than #11349. retrying on failures is good, but avoiding failures is even better

crates/bevy_asset/src/server/mod.rs
cart: I think a shared semaphore between anything file-related that needs it (FileAssetReader / FileAssetWriter) makes the most sense to me. I'd like to avoid leaking internal filesystem-specific constraints into the high level asset server.

cart:
Yeah my current stance is that this approach is the move, but the logic / semaphore should be scoped to the FileAssetReader/FileAssetWriter (and shared between instances).  Exposing filesystem-specific constraints in the high level asset server feels wrong to me (and is overly strict when non-filesystem asset sources enter the picture).

alice-i-cecile:
Cutting from 0.16 milestone since I don't think it's going to be ready in time :(

PR #17380, opened 2025-01-15
Remove Event: Component trait bound using a wrapper type which impls Component

alice-i-cecile:
# Objective

As raised in https://github.com/bevyengine/bevy/pull/17317, the `Event: Component` trait bound is confusing to users.

In general, a type `E` (like `AppExit`) which implements `Event` should not:

- be stored as a component on an entity
- be a valid option for `Query<&AppExit>`
- require the storage type and other component metadata to be specified

Events are not components (even if they one day use some of the same internal mechanisms), and this trait bound is confusing to users.

We're also automatically generating `Component` impls with our derive macro, which should be avoided when possible to improve explicitness and avoid conflicts with user impls.

Closes #17317, closes #17333

## Solution

- We only care that each unique event type gets a unique `ComponentId`
- dynamic events need their own tools for getting identifiers anyways
- This avoids complicating the internals of `ComponentId` generation.
- Clearly document why this cludge-y solution exists.

In the medium term, I think that either a) properly generalizing `ComponentId` (and moving it into `bevy_reflect?) or b) using a new-typed `Entity` as the key for events is more correct. This change is stupid simple though, and removes the offending trait bound in a way that doesn't introduce complex tech debt and does not risk changes to the internals.

This change does not:

- restrict our ability to implement dynamic buffered events (the main improvement over #17317)
   - there's still a fair bit of work to do, but this is a step in the right direction
- limit our ability to store event metadata on entities in the future
- make it harder for users to work with types that are both events and components (just add the derive / trait bound)

## Migration Guide

### DIFF
--- a/crates/bevy_ecs/src/event/base.rs
+++ b/crates/bevy_ecs/src/event/base.rs
@@ -33,7 +33,7 @@ use core::{
     label = "invalid `Event`",
     note = "consider annotating `{Self}` with `#[derive(Event)]`"
 )]
-pub trait Event: Component {
+pub trait Event: Send + Sync + 'static {
     /// The component that describes which Entity to propagate this event to next, when [propagation] is enabled.
     ///
     /// [propagation]: crate::observer::Trigger::propagate
@@ -49,12 +49,20 @@ pub trait Event: Component {
 
 /// An internal type that implements [`Component`] for a given [`Event`] type.
 ///
-/// This exists so we can easily get access to a unique [`ComponentId`](crate::component::ComponentId) for each [`Event`] type,
+/// This exists so we can easily get access to a unique [`ComponentId`] for each [`Event`] type,
 /// without requiring that [`Event`] types implement [`Component`] directly.
+/// [`ComponentId`] is used internally as a unique identitifier for events because they are:
 ///
-/// This is an implementation detail and should never be made public.
+/// - Unique to each event type.
+/// - Can be quickly generated and looked up.
+/// - Are compatible with dynamic event types, which aren't backed by a Rust type.
+///
+/// This type is an implementation detail and should never be made public.
+///
+/// [`ComponentId`]: crate::component::ComponentId
+// TODO: refactor events to store their metadata on distinct entities, rather than using `ComponentId`
 #[derive(Component)]
-struct EventWrapperComponent<E: Event>(PhantomData<E>);
+pub(crate) struct EventWrapperComponent<E: Event>(PhantomData<E>);
 
 /// An `EventId` uniquely identifies an event stored in a specific [`World`].
 ///
--- a/crates/bevy_ecs/src/event/base.rs
+++ b/crates/bevy_ecs/src/event/base.rs
@@ -1,4 +1,6 @@
 use crate as bevy_ecs;
+use crate::component::ComponentId;
+use crate::world::World;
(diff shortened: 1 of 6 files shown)


crates/bevy_ecs/src/observer/runner.rs
```diff
@@ -398,13 +400,13 @@ fn hook_on_add<E: Event, B: Bundle, S: ObserverSystem<E, B>>(
     _: ComponentId,
 ) {
     world.commands().queue(move |world: &mut World| {
-        let event_type = world.register_component::<E>();
+        let event_id = world.register_component::<EventWrapperComponent<E>>();
```
alice-i-cecile, line 403: Renamed for consistency with other callsites.

james-j-obrien (changes requested):
Nice simple solution, needs updates to `World::bootstrap` for the built-in event types otherwise I think it looks good.

james-j-obrien (approved):

crates/bevy_ecs/src/event/base.rs
```diff
@@ -48,8 +47,41 @@ pub trait Event: Component {
...
+/// [`ComponentId`] is used internally as a unique identitifier for events because they are:
+///
+/// - Unique to each event type.
+/// - Can be quickly generated and looked up.
+/// - Are compatible with dynamic event types, which aren't backed by a Rust type.
+///
+/// This type is an implementation detail and should never be made public.
+// TODO: refactor events to store their metadata on distinct entities, rather than using `ComponentId`
+#[derive(Component)]
+struct EventWrapperComponent<E: Event + ?Sized>(PhantomData<E>);
```
chescock, line 83: Note that doing it this way won't give usable values for `layout` or `drop` in the `ComponentInfo`.  I don't think anyone is actually using those right now, but the reason I originally suggested a `#[repr(transparent)]` wrapper instead of `PhantomData` is that those values would be valid for the event type.  

```suggestion
#[repr(transparent)]
struct EventWrapperComponent<E: Event>(E);
```

crates/bevy_ecs/src/event/base.rs
```diff
@@ -48,8 +47,41 @@ pub trait Event: Component {
...
     /// [`Trigger::propagate`]: crate::observer::Trigger::propagate
     const AUTO_PROPAGATE: bool = false;
+
+    /// Generates the [`ComponentId`] for this event type.
+    ///
+    /// If this type has already been registered,
+    /// this will return the existing [`ComponentId`].
+    ///
+    /// This is used by various dynamically typed observer APIs.
+    fn register_component_id(world: &mut World) -> ComponentId {
```
chescock, line 57: Making these trait methods technically leaves open the ability for users to implement them manually.  If a manual implementation returned, say, `ON_ADD`, that would be bad.  I don't know whether it would actually be unsound, though, and I don't think anyone *would* do that.

chescock (approved):
Looks good!  I thought of some interesting edge cases, but I don't think they should block making progress.

alice-i-cecile:
@chescock, repr transparent caused issues with `Sized` that I couldn't work out immediately. Leaving for now.

crates/bevy_ecs/src/event/base.rs
alice-i-cecile: Left a note!

PR #17383, opened 2025-01-15
Add Cursor* things to bevy::prelude

mgi388:
# Objective and solution

- Add useful cursor types to `bevy::prelude`.
- Also add `bevy_image::prelude` doc while I was here since it seemed to be the only one missing the doc boilerplate.

## Testing

- CI
- Ran example: `cargo run --example custom_cursor_image --features=custom_cursor`
- Observed that we removed a bunch of manually imported types from 3 examples which is the goal of the PR.

## Changes

- `CursorGrabMode`, `SystemCursorIcon`, `CursorIcon` and `CustomCursor` are now a part of `bevy::prelude`.

### DIFF
--- a/crates/bevy_image/src/lib.rs
+++ b/crates/bevy_image/src/lib.rs
@@ -2,6 +2,9 @@
 
 extern crate alloc;
 
+/// The image prelude.
+///
+/// This includes the most common types in this crate, re-exported for your convenience.
 pub mod prelude {
     pub use crate::{
         dynamic_texture_atlas_builder::DynamicTextureAtlasBuilder,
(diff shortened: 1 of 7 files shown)


alice-i-cecile (commented):
I like the prelude docs, but I don't think these types are worth re-exporting in the prelude. We're currently too aggressive with that, and these aren't super commonly used.

mgi388:
> I like the prelude docs, but I don't think these types are worth re-exporting in the prelude. We're currently too aggressive with that, and these aren't super commonly used.

All good, that's fine. I wasn't really sure the conditions for deciding what should go in a prelude or not.

I'm happy to just close this.

PR #17388, opened 2025-01-15
Add external assets to .gitignore

alex5nader:
Added an external assets section to .gitignore. This prevents contributors from accidentally adding or committing them.

I believe currently the only externel asset is the meshlet bunny.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,6 +22,9 @@ assets/**/*.meta
 crates/bevy_asset/imported_assets
 imported_assets
 
+# External Assets
+assets/models/bunny.meshlet_mesh
+
 # Bevy Examples
 example_showcase_config.ron
 example-showcase-reports/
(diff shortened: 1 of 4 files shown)


BenjaminBrienen (approved):

mockersf (changes requested):
instead of ignoring a specific file, could you create a new folder, something like `assets/external`, ignore everything in it, and update the meshlet example and instructions to use it from there?

mockersf:
path also needs to be changed there: https://github.com/bevyengine/bevy/blob/64ab33c6a731b54894b27ce68e5b621162de06fe/Cargo.toml#L1255

and ping @JMS55 so you know why running the example will fail next time you pull after this is merged: you need to put the file at a new path

JMS55:
A little bit annoying since those folders won't exist, and I'll have to create them manually, but it's not a big deal.

I think you also need to update the path in Cargo.toml `[package.metadata.example.meshlet]` right?

alice-i-cecile (approved):

PR #17389, opened 2025-01-15
`add_edges` helper for directional navigation

alice-i-cecile:
# Objective

While `add_looping_edges` is a helpful method for manually defining directional navigation maps, we don't always want to loop around!

## Solution

Add a non-looping variant.

These commits are cherrypicked from the more complex #17247.

## Testing

I've updated the `directional_navigation` example to use these changes, and verified that it works.

### DIFF
--- a/crates/bevy_input_focus/src/directional_navigation.rs
+++ b/crates/bevy_input_focus/src/directional_navigation.rs
@@ -173,7 +173,7 @@ impl DirectionalNavigationMap {
         self.add_edge(b, a, direction.opposite());
     }
 
-    /// Add symettrical edges between each consecutive pair of entities in the provided slice.
+    /// Add symmetrical edges between each consecutive pair of entities in the provided slice.
     ///
     /// Unlike [`add_looping_edges`](Self::add_looping_edges), this method does not loop back to the first entity.
     pub fn add_edges(&mut self, entities: &[Entity], direction: CompassOctant) {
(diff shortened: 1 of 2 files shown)


BenjaminBrienen (commented):
Ignore this

crates/bevy_input_focus/src/directional_navigation.rs
```diff
@@ -173,7 +173,18 @@ impl DirectionalNavigationMap {
...
+    /// Add symettrical edges between each consecutive pair of entities in the provided slice.
+    ///
+    /// Unlike [`add_looping_edges`](Self::add_looping_edges), this method does not loop back to the first entity.
+    pub fn add_edges(&mut self, entities: &[Entity], direction: CompassOctant) {
+        for i in 0..entities.len() - 1 {
+            let a = entities[i];
+            let b = entities[i + 1];
+            self.add_symmetrical_edge(a, b, direction);
+        }
+    }
```
BenjaminBrienen, line 185: Rewrote it for fun, feel free to ignore
```suggestion
    pub fn add_edges(&mut self, entities: &[Entity], direction: CompassOctant) {
        for &[a, b] in entities.windows(2) {
            self.add_symmetrical_edge(a, b, direction);
        }
    }
```

BenjaminBrienen (approved):

crates/bevy_input_focus/src/directional_navigation.rs
```diff
@@ -173,7 +173,18 @@ impl DirectionalNavigationMap {
         self.add_edge(b, a, direction.opposite());
     }
 
-    /// Add symmetrical edges between all entities in the provided slice, looping back to the first entity at the end.
+    /// Add symettrical edges between each consecutive pair of entities in the provided slice.
```
rparrett, line 176:
```suggestion
    /// Add symmetrical edges between each consecutive pair of entities in the provided slice.
```

rparrett (approved):

viridia (approved):

PR #17398, opened 2025-01-16
Relationships (non-fragmenting, one-to-many)

### LINKED ISSUE #3742: Entity-entity relations 🌈

### LINKED ISSUE #12235: Use component lifecycle hooks to make `Parent` + `Children` hierarchy code simpler and more robust

### LINKED ISSUE #15299: despawn_recursive causing Tracy error
## Bevy version
0.14.2

## What you did

I have a project that involves spawning and despawning many entities. I am having trouble running Tracy with the project because of a Tracy error caused by despawn_recursive.

## What went wrong

Tracy errors out because it runs over the limit of systems that can be tracked (32,000).

## Additional information

The reason there are over 32,000 systems running is that DespawnRecursive uses the entity that was despawned to create the trace name, creating a unique system to be tracked in Tracy. This can be seen in the Statistics window in Tracy.

I was able to workaround the problem by limiting despawn_recursives, but this only delays the problem, as my project still eventually hits the 32k limit. Changing despawn_recursive to despawn, fixes the issue, but I need despawn_recursive in some places.

### Tracy Error
tracy_error

### Tracy Statistics, filled with 32,000 unique DespawnRecursive commands
tracy_statistics

### Code causing the problem in `bevy_hierarchy/src/hierarchy.rs`
https://github.com/bevyengine/bevy/blob/1bb8007dceb03f41b20317bb44fcdfec1d70513c/crates/bevy_hierarchy/src/hierarchy.rs#L57C1-L81C2
```
impl Command for DespawnRecursive {
    fn apply(self, world: &mut World) {
        #[cfg(feature = "trace")]
        let _span = bevy_utils::tracing::info_span!(
            "command",
            name = "DespawnRecursive",
            entity = bevy_utils::tracing::field::debug(self.entity)
        )
        .entered();
        despawn_with_children_recursive(world, self.entity);
    }
}

impl Command for DespawnChildrenRecursive {
    fn apply(self, world: &mut World) {
        #[cfg(feature = "trace")]
        let _span = bevy_utils::tracing::info_span!(
            "command",
            name = "DespawnChildrenRecursive",
            entity = bevy_utils::tracing::field::debug(self.entity)
        )
        .entered();
        despawn_children_recursive(world, self.entity);
    }
}
```

### LINKED ISSUE #15308: Move `bevy_hierarchy` back into `bevy_ecs`

`bevy_hierarchy` is in an uncomfortable place: technically existing outside of `bevy_ecs`, but in practice generally privileged by the ecosystem.

One day (🥺), #3742 will replace `bevy_hierarchy` completely, allowing for generalized relations in a more robust fashion. This needs to be part of `bevy_ecs`, so `bevy_hierarchy`'s days are numbered.

1. Move the core parent-child components and logic to `bevy_ecs`.
2. Feature flag this under an on-by-default `hierarchy` flag.
3. Move any transform propagation logic to `bevy_transform`.
4. Move any visibility propagation logic to `bevy_render`.

As follow-up, consider changing the default despawn behavior to act recursively.

We could leave this alone, but this is causing problems when trying to tackle how #11090 interacts with hierarchies.

## Additional context

Desired by @mockersf (maintainer), @maniwani (SME-ECS) and myself (SME-ECS and maintainer).

### LINKED ISSUE #17301: Spawning child entities is extremely slow

cart:
This adds support for one-to-many non-fragmenting relationships (with planned paths for fragmenting and non-fragmenting many-to-many relationships). "Non-fragmenting" means that entities with the same relationship type, but different relationship targets, are not forced into separate tables (which would cause "table fragmentation").

Functionally, this fills a similar niche as the current Parent/Children system. The biggest differences are:

1. Relationships have simpler internals and significantly improved performance and UX. Commands and specialized APIs are no longer necessary to keep everything in sync. Just spawn entities with the relationship components you want and everything "just works".
2. Relationships are generalized. Bevy can provide additional built in relationships, and users can define their own.

**REQUEST TO REVIEWERS**: _please don't leave top level comments and instead comment on specific lines of code. That way we can take advantage of threaded discussions. Also dont leave comments simply pointing out CI failures as I can read those just fine._

## Built on top of what we have

Relationships are implemented on top of the Bevy ECS features we already have: components, immutability, and hooks. This makes them immediately compatible with all of our existing (and future) APIs for querying, spawning, removing, scenes, reflection, etc. The fewer specialized APIs we need to build, maintain, and teach, the better.

## Why focus on one-to-many non-fragmenting first?

1. This allows us to improve Parent/Children relationships immediately, in a way that is reasonably uncontroversial. Switching our hierarchy to fragmenting relationships would have significant performance implications. ~~Flecs is heavily considering a switch to non-fragmenting relations after careful considerations of the performance tradeoffs.~~ _(Correction from @SanderMertens: Flecs is implementing non-fragmenting storage specialized for asset hierarchies, where asset hierarchies are many instances of small trees that have a well defined structure)_
2. Adding generalized one-to-many relationships is currently a priority for the [Next Generation Scene / UI effort](https://github.com/bevyengine/bevy/discussions/14437). Specifically, we're interested in building reactions and observers on top. 

## The changes

This PR does the following:

1. Adds a generic one-to-many Relationship system
3. Ports the existing Parent/Children system to Relationships, which now lives in `bevy_ecs::hierarchy`. The old `bevy_hierarchy` crate has been removed.
4. Adds on_despawn component hooks
5. Relationships can opt-in to "despawn descendants" behavior, meaning that the entire relationship hierarchy is despawned when `entity.despawn()` is called. The built in Parent/Children hierarchies enable this behavior, and `entity.despawn_recursive()` has been removed.
6. `world.spawn` now applies commands after spawning. This ensures that relationship bookkeeping happens immediately and removes the need to manually flush. This is in line with the equivalent behaviors recently added to the other APIs (ex: insert).
7. Removes the ValidParentCheckPlugin (system-driven / poll based) in favor of a `validate_parent_has_component` hook.

## Using Relationships

The `Relationship` trait looks like this:

```rust
pub trait Relationship: Component + Sized {
    type RelationshipSources: RelationshipSources<Relationship = Self>;
    fn get(&self) -> Entity;
    fn from(entity: Entity) -> Self;
}
```

A relationship is a component that:

1. Is a simple wrapper over a "target" Entity.
2. Has a corresponding `RelationshipSources` component, which is a simple wrapper over a collection of entities. Every "target entity" targeted by a "source entity" with a `Relationship` has a `RelationshipSources` component, which contains every "source entity" that targets it.  

For example, the `Parent` component (as it currently exists in Bevy) is the `Relationship` component and the entity containing the Parent is the "source entity". The entity _inside_ the `Parent(Entity)` component is the "target entity". And that target entity has a `Children` component (which implements `RelationshipSources`).

In practice, the Parent/Children relationship looks like this:

```rust
#[derive(Relationship)]
#[relationship(relationship_sources = Children)]
pub struct Parent(pub Entity);

#[derive(RelationshipSources)]
#[relationship_sources(relationship = Parent)]
pub struct Children(Vec<Entity>);
```

The Relationship and RelationshipSources derives automatically implement Component with the relevant configuration (namely, the hooks necessary to keep everything in sync).

The most direct way to add relationships is to spawn entities with relationship components:

```rust
let a = world.spawn_empty().id();
let b = world.spawn(Parent(a)).id();

assert_eq!(world.entity(a).get::<Children>().unwrap(), &[b]);
```

There are also convenience APIs for spawning more than one entity with the same relationship:

```rust
world.spawn_empty().with_related::<Children>(|s| {
    s.spawn_empty();
    s.spawn_empty();
})
```

The existing `with_children` API is now a simpler wrapper over `with_related`. This makes this change largely non-breaking for existing spawn patterns.

```rust
world.spawn_empty().with_children(|s| {
    s.spawn_empty();
    s.spawn_empty();
})
```

There are also other relationship APIs, such as `add_related` and `despawn_related`.

## Automatic recursive despawn via the new on_despawn hook

`RelationshipSources` can opt-in to "despawn descendants" behavior, which will despawn all related entities in the relationship hierarchy:

```rust
#[derive(RelationshipSources)]
#[relationship_sources(relationship = Parent, despawn_descendants)]
pub struct Children(Vec<Entity>);
```

This means that `entity.despawn_recursive()` is no longer required. Instead, just use `entity.despawn()` and the relevant related entities will also be despawned.

To despawn an entity _without_ despawning its parent/child descendants, you should remove the `Children` component first, which will also remove the related `Parent` components:

```rust
entity
    .remove::<Children>()
    .despawn()
```

This builds on the on_despawn hook introduced in this PR, which is fired when an entity is despawned (before other hooks).

## Relationships are the source of truth

`Relationship` is the _single_ source of truth component. `RelationshipSources` is merely a reflection of what all the `Relationship` components say. By embracing this, we are able to significantly improve the performance of the system as a whole. We can rely on component lifecycles to protect us against duplicates, rather than needing to scan at runtime to ensure entities don't already exist (which results in quadratic runtime). A single source of truth gives us constant-time inserts. This does mean that we cannot directly spawn populated `Children` components (or directly add or remove entities from those components). I personally think this is a worthwhile tradeoff, both because it makes the performance much better _and_ because it means theres exactly one way to do things (which is a philosophy we try to employ for Bevy APIs).

As an aside: treating both sides of the relationship as "equivalent source of truth relations" does enable building simple and flexible many-to-many relationships. But this introduces an _inherent_ need to scan (or hash) to protect against duplicates. [`evergreen_relations`](https://github.com/EvergreenNest/evergreen_relations) has a very nice implementation of the "symmetrical many-to-many" approach. Unfortunately I think the performance issues inherent to that approach make it a poor choice for Bevy's default relationship system. 

## Followup Work

* Discuss renaming `Parent` to `ChildOf`. I refrained from doing that in this PR to keep the diff reasonable, but I'm personally biased toward this change (and using that naming pattern generally for relationships).
* [Improved spawning ergonomics](https://github.com/bevyengine/bevy/discussions/16920)
* Consider adding relationship observers/triggers for "relationship targets" whenever a source is added or removed. This would replace the current "hierarchy events" system, which is unused upstream but may have existing users downstream. I think triggers are the better fit for this than a buffered event queue, and would prefer not to add that back.
* Fragmenting relations: My current idea hinges on the introduction of "value components" (aka: components whose type _and_ value determines their ComponentId, via something like Hashing / PartialEq). By labeling a Relationship component such as `ChildOf(Entity)` as a "value component",  `ChildOf(e1)` and `ChildOf(e2)` would be considered "different components". This makes the transition between fragmenting and non-fragmenting a single flag, and everything else continues to work as expected.
* Many-to-many support
    * Non-fragmenting: We can expand Relationship to be a list of entities instead of a single entity. I have largely already written the code for this.
    * Fragmenting: With the "value component" impl mentioned above, we get many-to-many support "for free", as it would allow inserting multiple copies of a Relationship component with different target entities.

Fixes #3742 (If this PR is merged, I think we should open more targeted followup issues for the work above, with a fresh tracking issue free of the large amount of less-directed historical context)
Fixes #17301
Fixes #12235 
Fixes #15299
Fixes #15308 

## Migration Guide

* Replace `ChildBuilder` with `ChildSpawnerCommands`.
* Replace calls to `.set_parent(parent_id)` with `.insert(Parent(parent_id))`.
* Replace calls to `.replace_children()` with `.remove::<Children>()` followed by `.add_children()`. Note that you'll need to manually despawn any children that are not carried over.
* Replace calls to `.despawn_recursive()` with `.despawn()`.
* Replace calls to `.despawn_descendants()` with `.despawn_related::<Children>()`.
* If you have any calls to `.despawn()` which depend on the children being preserved, you'll need to remove the `Children` component first.


crates/bevy_ecs/src/relationship/mod.rs
```diff
@@ -0,0 +1,262 @@
...
+    }
+}
+
+/// A [`Component`] containing the collection of entities that relate to this [`Entity`] via the associated `Relationship` type.
+/// See the [`Relationship`] documentation for more information.
+///
+// The "deprecated" state is used to prevent users from mutating the internal RelationshipSource collection.
+// These internals are allowed to modify the internal RelationshipSource collection.
+#[allow(deprecated)]
+pub trait RelationshipSources: Component<Mutability = Mutable> + Sized {
```
cart, line 137: I am open to alternative names for this concept, but I would prefer to keep the `Relationship` name as it is, as that is the "driving concept" for the "relationship feature".

Carter0 (commented):
omg its happening

bushrat011899 (approved):
Excellent to see this landing! Mostly comments, but I do have one question around the need for `Component<Mutability = Mutable>` on `RelationshipSource`.

crates/bevy_ecs/src/world/component_constants.rs
```diff
@@ -13,6 +13,8 @@ pub const ON_INSERT: ComponentId = ComponentId::new(1);
 pub const ON_REPLACE: ComponentId = ComponentId::new(2);
 /// [`ComponentId`] for [`OnRemove`]
 pub const ON_REMOVE: ComponentId = ComponentId::new(3);
+/// [`ComponentId`] for [`OnDespawn`]
+pub const ON_DESPAWN: ComponentId = ComponentId::new(4);
```
alice-i-cecile, line 17: This doesn't seem to have been added to `World::bootstrap`, so the component ID won't be set correctly.

crates/bevy_ecs/macros/src/relationship.rs
```diff
@@ -0,0 +1,235 @@
...
+            }
+
+            fn from_collection(collection: Self::Collection) -> Self {
+                Self(collection)
+            }
+        }
+
+        impl #impl_generics #bevy_ecs_path::component::Component for #struct_name #type_generics #where_clause {
+            const STORAGE_TYPE: #bevy_ecs_path::component::StorageType = #bevy_ecs_path::component::StorageType::Table;
+            type Mutability = #bevy_ecs_path::component::Mutable;
```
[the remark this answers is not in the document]
cart: Yes this was deliberate for a couple of reasons:

1. RelationshipSources like Children will want user-facing APIs to reorder children, as order may be semantically relevant in some cases (or may just be organizational ... ex: dragging to reorder children in the editor). This does not violate the relationship correctness, so it can / should be allowed.
2. The immutable model doesn't really makes sense, given that we're appending to a list regularly as new children are added. We could hack around this, but that would break the user-facing immutable data model.

crates/bevy_ecs/src/world/component_constants.rs
cart: Hmm interesting that the relevant recursive despawn tests still pass. Probably just getting lucky on the id allocation :)

I'll fix this.

crates/bevy_ecs/macros/src/relationship.rs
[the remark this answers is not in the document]
bushrat011899: That makes sense, I figured it had something to do with ordering. Is there any mechanism to prevent/detect/dissuade users from actually _changing_ the values in a `RelationshipSources` type like `Children`, or is that the responsibility of the library author defining the type (e.g., `Children` has its internal `Vec<Entity>` private)?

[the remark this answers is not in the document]
alice-i-cecile: A comment here would be useful for future readers of this code :) I agree with and really like this explanation!

cBournhonesque (commented):
Didn't read everything, added a couple comments.
Is there a test for the `commands.remove::<Parent>().despawn()` doesn't despawn descendants? I think that would be good to have

crates/bevy_ecs/src/hierarchy.rs
```diff
@@ -0,0 +1,305 @@
...
+    Component,
+    MapEntities,
+    VisitEntities,
+    VisitEntitiesMut,
+    PartialEq,
+    Debug,
+    FromWorld
+)]
+#[relationship(relationship_sources = Children)]
+pub struct Parent(pub Entity);
```
physgun, line 40: >     * Discuss renaming `Parent` to `ChildOf`.

When I started learning how to use Bevy recently, this was the first stumbling block. It seemed straightforward that `Parent` was a marker component (or "driving concept") for a parent entity!

crates/bevy_ecs/src/system/commands/entity_command.rs
```diff
@@ -254,8 +254,8 @@ pub fn retain<T: Bundle>() -> impl EntityCommand {
 ///
 /// # Note
 ///
-/// This won't clean up external references to the entity (such as parent-child relationships
-/// if you're using `bevy_hierarchy`), which may leave the world in an invalid state.
+/// This will also despawn the entities in any [`RelationshipSources`](crate::relationship::RelationshipSources) that are configured
+/// to despawn descendants. For example, this will recursively despawn [`Children`](crate::hierarchy::Children).
 pub fn despawn() -> impl EntityCommand {
```
MrGVSV, line 259: Would it be worth adding `despawn_recursive` here so it can be marked as deprecated? Or maybe copy over the original trait and mark that?

crates/bevy_ecs/src/relationship/mod.rs
vultix: It definitely feels oxymoronic that `Relationship` is considered the source of truth, and `RelationshipSources` isn’t.  The Parent component isn’t the source of the relationships, but the recipient.

vultix: Some other options:

- `RelationshipTarget`
- `RelationshipParent`
- `RelationshipRecipient`
- `RelationshipHub`
- `RelationshipCollector`
- `RelationshipTracker`
- `RelationshipContainer`

I think `RelationshipContainer` is my favorite. 
```
Relationship: ChildOf, StudentOf, EmployeeOf
RelationshipContainer: Children, Students, Employees
```

theotherphil: I was initially very confused by the PR explanation until I realised that the RelationshipSources component goes on the relationship target.

I think main confusion was from not spotting which of  two plausible metaphors is being used:

* Properties: you _have_ a Parent just like you _have_ a velocity or position
* Markers: entities can _be_ a parent just like they can _be_ a Monster

This PR uses the properties metaphor but the market metaphor feels more natural to me. A child entity _is_ a Relationship(Source) and a parent _is_ a Relationship target. If you're already thinking of switching Parent to Child of then this would make RelationshipTarget a natural name for RelationshipSources.

Bleachfuel: maybe RelationshipWith

> Relationship: ChildOf, StudentOf, EmployeeOf
RelationshipWith: Children, Students, Employees

Althought, U also have a relation when u are childOf, with another parent for example. But since RelationShip doesn't have to think about that. RelationshipWith would be a very clear name with who you have a relationship with!

crates/bevy_ecs/src/relationship/relationship_query.rs
```diff
@@ -0,0 +1,241 @@
...
+        <D as QueryData>::ReadOnly: WorldQuery<Item<'w> = &'w S>,
+    {
+        self.get(entity)
+            .into_iter()
+            .flat_map(RelationshipSources::iter)
+    }
+
+    /// Recursively walks up the tree defined by the given `R` [`Relationship`] until
+    /// there are no more related entities, returning the "root entity" of the relationship hierarchy.
+    pub fn root_ancestor<R: Relationship>(&'w self, entity: Entity) -> Entity
```
eckz, line 36: Generic relations can be circular and there is nothing really preventing it.
So majority of these methods could create infinite loops.

Also 
 - `root_ancestor `
 - `iter_leaves `
 - `iter_siblings `
 - `iter_descendants`
 - `iter_ancestors`

Make sense when talking about Child/Parent relationships, in other type of relationship it might not make sense, specially circular ones.

crates/bevy_ecs/src/relationship/relationship_source_collection.rs
```diff
@@ -0,0 +1,58 @@
...
+    fn add(&mut self, entity: Entity);
+
+    /// Removes the given `entity` from the collection.
+    fn remove(&mut self, entity: Entity);
+
+    /// Iterates all entities in the collection.
+    fn iter(&self) -> impl DoubleEndedIterator<Item = Entity>;
+
+    /// Takes all entities in the given collection.
+    fn take(&mut self) -> Vec<Entity>;
```
eckz, line 20: Is this really necessary?, It looks like it's not being used.

It could be make more generic too:

```
    /// Takes all entities in the given collection.
    fn drain(&mut self) -> impl Iterator<Item=Entity>;
```

crates/bevy_ecs/src/system/commands/entity_command.rs
```diff
@@ -254,8 +254,8 @@ pub fn retain<T: Bundle>() -> impl EntityCommand {
 ///
 /// # Note
 ///
-/// This won't clean up external references to the entity (such as parent-child relationships
-/// if you're using `bevy_hierarchy`), which may leave the world in an invalid state.
+/// This will also despawn the entities in any [`RelationshipSources`](crate::relationship::RelationshipSources) that are configured
+/// to despawn descendants. For example, this will recursively despawn [`Children`](crate::hierarchy::Children).
```
eckz, line 258: Given how prominent Child/Parent relations are, I think it's worth mentioning despawning of Children the first thing, and then any other generic relationship.

```rust
/// An [`EntityCommand`] that despawns an entity with all it's descendants.
///
/// # Note
/// ...
```

Same thing would apply to any `*despawn` method.

crates/bevy_ecs/src/traversal.rs
```diff
@@ -30,3 +30,14 @@ impl<D> Traversal<D> for () {
         None
     }
 }
+
+/// This provides generalized hierarchy traversal for use in [event propagation].
+///
+/// `Parent::traverse` will never form loops in properly-constructed hierarchies.
```
eckz, line 36: I guess this comment is not correct since nothing prevents relationship loops.

crates/bevy_ecs/src/relationship/mod.rs
```diff
@@ -0,0 +1,262 @@
...
+
+    /// Returns a reference to the stored [`RelationshipSources::Collection`].
+    fn collection(&self) -> &Self::Collection;
+    /// Returns a mutable reference to the stored [`RelationshipSources::Collection`].
+    ///
+    /// # Warning
+    /// This should generally not be called by user code, as modifying the internal collection could invalidate the relationship.
+    /// This uses the "deprecated" state to warn users about this.
+    #[deprecated = "Modifying the internal RelationshipSource collection should only be done by internals as it can invalidate relationships."]
+    fn collection_mut(&mut self) -> &mut Self::Collection;
```
eckz, line 151: Marking a function as `deprecated` to deter people from using it's very weird.

I see other options:

#### Make the functions unsafe:
```rust
unsafe fn collection_mut(&mut self) -> &mut Self::Collection;
```
####  Change the name to be more scary, like 
```rust
fn __internal_collection_mut(&mut self) -> &mut Self::Collection;
```

####  Create a wrapper that can be created publicly, but only accessed by the bevy crate.

```rust

/// Wrapper of a type can be constructed publicly, but only dereference by this crate.
#[doc(hidden)]
pub struct InternalCollectionWrapper<T>(T);
impl <T> InternalCollectionWrapper<T> {
    /// Creates 
    pub fn new(value: T) -> Self {
        Self(value)
    }
}

#[allow(deprecated)]
pub trait RelationshipSources: Component<Mutability = Mutable> + Sized {
    // [...]
    #[doc(hidden)]
    fn collection_mut(&mut self) -> InternalCollectionWrapper<&mut Self::Collection>;
}
```

`from_collection` could be removed and let the `proc_macro` directly generate `fn RelationshipSources::with_capacity(capacity: usize) {}` since it's the only place used.

This way the only possible operation would be to create an empty Collection (with some capacity), and that shouldn't be unsafe.

crates/bevy_ecs/src/relationship/mod.rs
cart: > This PR uses the fields metaphor but the marker metaphor feels more natural to me. A child entity is a Relationship(Source) and a parent is a RelationshipTarget. If you're already thinking of switching Parent to ChildOf then this would make RelationshipTarget a natural name for RelationshipSources.

Solid argument. I'm pretty sure I like RelationshipTarget.

cart:
I misrepresented the scope of the Flecs effort to switch to non-fragmenting relations (sorry @SanderMertens!). I've replaced the relevant text in the description with the following text:

> Flecs is implementing non-fragmenting storage specialized for asset hierarchies, where asset hierarchies are many instances of small trees that have a well defined structur

crates/bevy_ecs/macros/src/component.rs
```diff
@@ -207,6 +210,7 @@ pub const ON_ADD: &str = "on_add";
 pub const ON_INSERT: &str = "on_insert";
 pub const ON_REPLACE: &str = "on_replace";
 pub const ON_REMOVE: &str = "on_remove";
+pub const ON_DESPAWN: &str = "on_despawn";
```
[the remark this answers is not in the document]
cart: We can benchmark, but I don't expect significant degradation.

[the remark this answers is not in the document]
bushrat011899: I wouldn't expect any problems either, agreed.

crates/bevy_ecs/macros/src/relationship.rs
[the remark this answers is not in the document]
cart: Just added a comment.

crates/bevy_ecs/src/system/commands/entity_command.rs
cart: Good call. I added the methods directly to EntityCommands / EntityWorldMut. The trait path is already broken, so I don't see much sense in keeping it around. This feels like a reasonable middle ground.

crates/bevy_ecs/src/relationship/relationship_source_collection.rs
cart: Good call. Removed!

crates/bevy_ecs/src/relationship/mod.rs
cart: Yeah I definitely admit it is weird. I opted for the current path for a few reasons:

1. It displays an informative compiler warning when developers try to call these methods. They can choose to suppress it after considering the tradeoffs.
2. Using `unsafe` for code that doesn't actually involve "memory safety" is generally frowned upon and is its own kind of weird. This also doesn't directly inform users about the context.

My preferences is for either the deprecation approach or the "scary name" approach.

crates/bevy_ecs/src/relationship/relationship_query.rs
cart: I added some warnings here. I'm not sure this is the kind of constraint we should add at the "relationship API level", as I don't really think paying the loop detection cost is worth it, so it wouldn't actually be enforced.

mintlu8:
Would it be more general to implement the macro like this? This can also work with one way relation from parent to child.
```rust
#[derive(Component)]
#[relationship_sources(relationship = Parent)]
pub struct Children(#[despawns] Vec<Entity>);
```

viridia (approved):
The API looks good. I've gone ahead and ported one of my UI libraries to the new API and encountered no significant problems.

viridia:
# Migration Guide

* Replace `ChildBuilder` with `ChildSpawnerCommands`.
* Replace calls to `.set_parent(parent_id)` with `.insert(Parent(parent_id))`.
* Replace calls to `.replace_children()` with `.remove::<Children>()` followed by `.add_children()`. Note that you'll need to manually despawn any children that are not carried over.
* Replace calls to `.despawn_recursive()` with `.despawn()`.
* Replace calls to `.despawn_descendants()` with `.despawn_related::<Children>()`.
* If you have any calls to `.despawn()` which depend on the children being preserved, you'll need to remove the `Children` component first.

crates/bevy_ecs/src/relationship/mod.rs
cart: I'm actually a bit torn. On the one hand, I think we should be embracing the IsA(X) naming convention for most components over HasA(X). And RelationshipTarget feels pretty right in that context. 

However the actual instances of this trait will have names like `Children`, `Observers`, `Reactions`, etc. Which follows the HasA pattern. RelationshipSources behaves like a collection type. In that regard, `RelationshipSources::iter()` makes way more sense than `RelationshipTarget::iter()`, as the latter kind of feels like we're iterating the targets, not the sources.

crates/bevy_ecs/src/relationship/mod.rs
alice-i-cecile: I'd much prefer "scary name" over permanently deprecated :)

crates/bevy_ecs/src/hierarchy.rs
```diff
@@ -0,0 +1,345 @@
...
+
+impl Parent {
+    pub fn get(&self) -> Entity {
+        self.0
+    }
+}
+
+// TODO: We need to impl either FromWorld or Default so Parent can be registered as Reflect.
```
alice-i-cecile, line 48: The pattern that users have come up with is to create a panicking constructor in the `Default` impl. I don't think that would work here though, since you need to actually create an instance.

We *could* store an `Option` inside, but that's a fair bit of UX pain.

crates/bevy_ecs/src/hierarchy.rs
```diff
@@ -0,0 +1,345 @@
...
+    #[inline(always)]
+    fn from_world(_world: &mut World) -> Self {
+        Parent(Entity::PLACEHOLDER)
+    }
+}
+
+#[derive(RelationshipSources, Default, Reflect, VisitEntitiesMut)]
+#[relationship_sources(relationship = Parent, despawn_descendants)]
+#[reflect(Component, MapEntities, VisitEntities, VisitEntitiesMut)]
+pub struct Children(Vec<Entity>);
```
alice-i-cecile, line 62: `Parent` and `Children` definitely need docs before we ship this.

crates/bevy_ecs/src/hierarchy.rs
alice-i-cecile: I think we should consider renaming these components this cycle. Easy to leave for follow-up though.

crates/bevy_ecs/src/hierarchy.rs
```diff
@@ -0,0 +1,345 @@
+// TODO: REMOVE THIS
+#![allow(missing_docs)]
+
+//! Parent-Child relationships for entities.
```
alice-i-cecile, line 4: More docs on the "standard properties" of parent-child relationships would be good here.

crates/bevy_ecs/src/relationship/mod.rs
```diff
@@ -0,0 +1,262 @@
+//! This module provides [`Relationship`] functionality. See the [`Relationship`] trait for more info.
```
alice-i-cecile, line 1:
```suggestion
//! This module provides functionality to link entities to each other using specialized components called "relationships". See the [`Relationship`] trait for more info.
```

crates/bevy_ecs/src/hierarchy.rs
cart, line 62: @mockersf just suggested that we auto-generate RelationshipSources on behalf of the developer. So this:

```rust
#[derive(Component)]
#[relationship(relationship_sources = Children, despawn_descendants)]
pub struct Parent(pub Entity);
```

Would fully generate the `Children` struct and the relevant derives.

Some pros:

1. It reduces a lot of boilerplate for the cases that don't need customization
2. Implementing the RelationshipSources side is a bit complicated. Ex: it needs to wrap a single unnamed field, which needs to be private, it needs to point back to the relationship via the `relationship_sources` attribute, to be compatible with the current scene system it needs to support MapEntities/VisitEntities, etc.
3. We could still support direct derives / normal configuration by making it possible to opt-out of manual generation.

Some cons:

1. It means by default the user has less control over the type. They cannot derive new traits (nicely) while they are using auto-generation, add requires, define the component storage type, etc. They also cannot naturally write documentation for it.
2. It has a "weirdness factor". The `Children` type is no longer directly specified in code.

crates/bevy_ecs/src/relationship/mod.rs
```diff
@@ -0,0 +1,262 @@
...
+/// #[derive(Relationship)]
+/// #[relationship(relationship_sources = Children)]
+/// pub struct Parent(pub Entity);
+///
+/// #[derive(RelationshipSources)]
+/// #[relationship_sources(relationship = Parent, despawn_descendants)]
+/// pub struct Children(Vec<Entity>);
+/// ```
+///
+// NOTE: The "deprecated" state is used to prevent users from mutating the internal RelationshipSource collection.
```
alice-i-cecile, line 66: Mentioned elsewhere, but I much prefer the "scary name" approach over the "permanently deprecated" strategy :)

crates/bevy_ecs/src/hierarchy.rs
cart: Given that I'm pretty sure we want the manual mode no matter what, maybe we should punt this to a followup PR.

crates/bevy_ecs/src/relationship/mod.rs
```diff
@@ -0,0 +1,262 @@
...
+    type RelationshipSources: RelationshipSources<Relationship = Self>;
+
+    /// Gets the [`Entity`] ID of the related entity.
+    fn get(&self) -> Entity;
+
+    /// Creates this [`Relationship`] from the given `entity`.
+    fn from(entity: Entity) -> Self;
+
+    /// The `on_insert` component hook that maintains the [`Relationship`] / [`RelationshipSources`] connection.
+    fn on_insert(mut world: DeferredWorld, entity: Entity, _: ComponentId) {
```
alice-i-cecile, line 81: This feels like a weird place for this method. I'd kinda prefer making `on_insert` etc methods for `Component` (and removing the awkward `register_component_hooks` API), and then generating those methods in the `Component` trait with the `Relationship` derive macro.

crates/bevy_ecs/src/relationship/mod.rs
alice-i-cecile: I like `RelationshipTarget` + changing the name of `Children` etc to use `IsA` framing.

crates/bevy_ecs/src/relationship/relationship_query.rs
```diff
@@ -0,0 +1,261 @@
...
+    relationship::{Relationship, RelationshipSources},
+    system::Query,
+};
+use alloc::collections::VecDeque;
+use smallvec::SmallVec;
+
+impl<'w, 's, D: QueryData, F: QueryFilter> Query<'w, 's, D, F> {
+    /// If the given `entity` contains the `R` [`Relationship`] component, returns the
+    /// target entity of that relationship.
+    pub fn related<R: Relationship>(&'w self, entity: Entity) -> Option<Entity>
```
alice-i-cecile, line 13: I am extremely pleased to see these generalized and baked into bevy_ecs :D

crates/bevy_ecs/src/relationship/relationship_source_collection.rs
```diff
@@ -0,0 +1,51 @@
...
+    fn len(&self) -> usize;
+
+    /// Returns true if the collection contains no entities.
+    #[inline]
+    fn is_empty(&self) -> bool {
+        self.len() == 0
+    }
+}
+
+impl RelationshipSourceCollection for Vec<Entity> {
```
alice-i-cecile, line 29: We should add an impl for (and benchmark using) `SmallVec<Entity>` here. Most children lists are going to be *small*.

`EntityHashSet` would also be a really nice alternative.

crates/bevy_text/src/text.rs
```diff
@@ -510,7 +508,7 @@ pub fn detect_text_needs_rerender<Root: Component>(
             ));
             continue;
         };
-        let mut parent: Entity = **span_parent;
+        let mut parent: Entity = span_parent.0;
```
alice-i-cecile, line 511: If we impl `Deref` for `Parent` we should be able to revert these changes and reduce breakage in the engine and for users.

cart: I personally think `**` is _significantly_ nastier and would prefer to remove it. Reducing breakage is nice, but im not sure deferring this pain (if we agree that it is inevitable) will be particularly beneficial.

alice-i-cecile: Given that `Entity` doesn't have any meaningful methods on it, I think that's a reasonable position to take.

cart: Although I guess another reasonable position to take is that a Deref impl doesn't really hurt anyone and we can always just not use it by convention. Nevermind I'll add it in the interest of keeping breakage down, and just stop using it in engine code.

crates/bevy_ecs/src/relationship/mod.rs
alice-i-cecile: Tracked in #17411.

crates/bevy_ecs/src/relationship/mod.rs
cart: What would `Children` be called? I don't have a satisfying alternative. Note that this is the term we'd use for the improved spawn APIs, ex:

```rust
world.spawn((
    Foo,
    Children::spawn_one(Bar),
))

world.spawn((
    Foo,
    children![
        Bar,
    ]
));
```

cart: I think `Children` / `Observers` / `Reactions`, etc are already the ideal names for these cases.

cart: IsA vs HasA does break down for common cases. An entity is not a `Health` or a `Display`. It _has_ a Display property and _has_ Health.

viridia: I suggest that we separate the names for the constructors (`Children`) from the name of the component that represents the relation (`ParentOf`). `Children` doesn't just insert a component, it inserts a two-way relation that consists of both `ParentOf` and `ChildOf`, so there's some justification for not having the same name.

I agree that in terms of construction ergonomics that `Children` is perfectly intuitive. But I don't necessarily think that intuition carries over into the underlying representation.

crates/bevy_ecs/src/hierarchy.rs
cart: Note that this is carryover from the current hierarchy/scene code. I don't feel any particular inclination to sort this out here.

crates/bevy_ecs/src/relationship/mod.rs
viridia: If we do decide to rename these components, I'm very much in favor of doing this as a separate PR (#17412) as it will result in a lot of churn.

crates/bevy_ecs/macros/src/component.rs
```diff
@@ -202,11 +283,14 @@ pub fn document_required_components(attr: TokenStream, item: TokenStream) -> Tok
...
 pub const STORAGE: &str = "storage";
 pub const REQUIRE: &str = "require";
+pub const RELATIONSHIP: &str = "relationship";
+pub const RELATIONSHIP_SOURCES: &str = "relationship_sources";
 
 pub const ON_ADD: &str = "on_add";
 pub const ON_INSERT: &str = "on_insert";
 pub const ON_REPLACE: &str = "on_replace";
 pub const ON_REMOVE: &str = "on_remove";
+pub const ON_DESPAWN: &str = "on_despawn";
```
Jondolf, line 293: Probably not something that should be done in this PR, but do we also want an `on_spawn` hook for completeness? There's already nice symmetry with `on_add`/`on_remove` and `on_insert`/`on_replace` (or `on_discard` if we want to rename it to that)

crates/bevy_ecs/src/relationship/mod.rs
cart: Yeah the rename was intentionally deferred to keep the diff in check.

I would prefer it if we didn't need to build special-cased constructors for these things. We shouldn't have two names for the same concept. I'd like these APIs to be symbol-driven. If ParentOf is the component name, then it should also be the thing that we use during spawn. `ParentOf` _is_ actually reasonable in this context:

```rust
world.spawn((
    Foo,
    ParentOf::spawn((
        Spawn(Bar),
        Spawn(Baz),
    )),
))

world.spawn((
    Foo,
    parent_of![
        Bar,
        Baz,
    ]
));
```

cart: If we do want to refer to the "collection of childof relationships" as "children" in this context, then we should violate our "IsA" constraint and just give it the name that makes us happy. Needing to connect ParentOf to a new "children" concept name is confusing and needlessly complicated. I don't think losing IsA here is a particular loss. It wouldn't be the first or the last case where "IsA" naming doesn't quite fit (as mentioned above).

viridia: All right. I don't feel strongly about any of these name changes, I was merely trying to fine-tune the wordsmithing.

I realized also that there are additional sources of naming confusion: for example, the "observe" and "own" verbs point in opposite directions: sub-entities "observe" super-entities, while super-entities "own" sub-entities. This means that the suffix "...By" has the opposite meaning in "ObservedBy" vs "OwnedBy".

The upshot here is that no matter what we do, people are going to simply have to remember which direction is associated with which name.

viridia (approved):
I've updated my repo to pick up the latest changes (using Component derive) and everything still works.

crates/bevy_ecs/src/relationship/mod.rs
cart: Actually I'll copy this over to the issue. We should discuss there.

crates/bevy_ecs/macros/src/component.rs
cart: Worth discussing, but I can't personally think of a spawn use case that wouldn't be better off in OnAdd or OnInsert. Given that every hook has a cost associated with it, we need real use cases to justify their existence.

Part of me thinks we might even be better off hard-coding the "recursive despawn" logic rather than using hooks, as I also can't think of a use case for on_despawn other than recursive despawn.

crates/bevy_ecs/src/relationship/relationship_source_collection.rs
cart: Agreed. Although I think that might be better off as followup work.

crates/bevy_ecs/src/relationship/mod.rs
cart: Done!

crates/bevy_ecs/src/relationship/mod.rs
cart: I've renamed RelationshipSources to RelationshipTarget.

alice-i-cecile:
@cart I have no concerns about this at this stage and we have plenty of community review; feel free to merge when you feel it's ready.

crates/bevy_ecs/src/relationship/mod.rs
```diff
@@ -0,0 +1,253 @@
...
+/// # use bevy_ecs::entity::Entity;
+/// #[derive(Component)]
+/// #[relationship(relationship_target = Children)]
+/// pub struct Parent(pub Entity);
+///
+/// #[derive(Component)]
+/// #[relationship_target(relationship = Parent, despawn_descendants)]
+/// pub struct Children(Vec<Entity>);
+/// ```
+pub trait Relationship: Component + Sized {
```
nxsaken, line 62: Is there a reason this is called `Relationship` and not `Relation`? I like the latter more since it's shorter

mrchantey: I suppose relationship is more accurate when describing specific instances of relations
https://www.britannica.com/dictionary/eb/qa/relations-and-relationship

rparrett:
Could we offer users some advice on migrating the removed `HierarchyEvent`?

PR #17399, opened 2025-01-16
Make mocking picking events for bevy_ui easier

alice-i-cecile:
# Objective

As discussed in https://github.com/bevyengine/bevy/pull/17267, we'd like to unify input-focus driven widget activations (e.g. press enter to click the focused button) with mouse-driven picking interactions by going through the bevy_picking infrastructure.

While you *can* fake this with observer triggers, it is:

- inaccurate: you're feeding in garbage information
- extremely verbose
- rather arcane

## Solution

As @aevyrie points out, we have access to everything we need to send realistic pointer events! Doing this *correctly* requires rather a lot of information. To make users lives easier, I've introduced a system param with all of the required information, which simulates pointer events at the center of the UI node specified.

The `directional_navigation` example has been updated to demonstrate / test this change.

## Discussion

- Some small improvements to this general area of the code have been split out into #17395. They'd be nice here, but I wanted to reduce complexity here.
- While PointerId::Custom exists, it is challenging to both generate a UUID and to match against it if you want to respond differently if the Pointer event is from a focused element. While I could have used sentinel values here, I think it's more idiomatic and nicer to use if we just add an enum arm for a common use.
- We should probably rename `bevy_ui/picking_backend.rs`, but I've opted to defer that to make the diff easier to read.
- This was originally a custom command, but has been refactored to a `SystemParam` for flexibility, easier error handling and performance.
- As follow-up, we could add some simple helper methods to the `SystemParam` to make certain actions even easier.
- In theory, other backends could do the same thing! The logic required is slightly specialized, so I stuck to just UI for now.

## Testing

I've tested this with the `directional_navigation` example. While I hope that this works properly with more exotic windowing / rendering target setups (viewports, portals, multiple windows), we're not equipped to do that effectively right now.

## Migration Guide

TODO

### DIFF
--- a/crates/bevy_picking/src/input.rs
+++ b/crates/bevy_picking/src/input.rs
@@ -79,7 +79,7 @@ impl Default for PointerInputPlugin {
 impl Plugin for PointerInputPlugin {
     fn build(&self, app: &mut App) {
         app.insert_resource(*self)
-            .add_systems(Startup, spawn_mouse_pointer)
+            .add_systems(Startup, (spawn_mouse_pointer, spawn_focus_pointer))
             .add_systems(
                 First,
                 (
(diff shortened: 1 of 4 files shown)


viridia:
A concern I have is that the motivation, use case, and overall theory of operation won't be apparent to people who are looking at the docs for Pointer. I would like to see a more detailed note, and possibly link to the emulated pointer documentation.

aevyrie:
I have a lot of thoughts but I need some time to write them down. I've explored this space a lot at work, and there are some design problems I think we need to solve first to avoid backing ourselves into a corner.

alice-i-cecile:
Needs to be redone / rethought. Closing.

PR #17403, opened 2025-01-16
Rename `TargetCamera` to `UiTargetCamera`

ickshonpe:
# Objective

It's not immediately obvious that `TargetCamera` only works with UI node entities. It's natural to assume from looking at something like the `multiple_windows` example that it will work with everything.

## Solution

Rename `TargetCamera` to `UiTargetCamera`.

## Migration Guide

`TargetCamera` has been renamed to `UiTargetCamera`.


alice-i-cecile (approved):
Annoying that this isn't more unified, but making this clear is better.

Carter0 (approved):

PR #17404, opened 2025-01-16
Reworked Segment types into their cartesian forms

Sigma-dev:
# Objective

Segment2d and Segment3d are currently hard to work with because unlike many other primary shapes, they are bound to the origin.
The objective of this PR is to allow these segments to exist anywhere in cartesian space, making them much more useful in a  variety of contexts.

## Solution

Reworking the existing segment type's internal fields and methods to allow them to exist anywhere in cartesian space.
I have done both reworks for 2d and 3d segments but I was unsure if I should just have it all here or not so feel free to tell me how I should proceed, for now I have only pushed Segment2d changes.

As I am not a very seasoned contributor, this first implementation is very likely sloppy and will need some additional work from my end, I am open to all criticisms and willing to work to get this to bevy's standards.

## Testing

I am not very familiar with the standards of testing. Of course my changes had to pass the thorough existing tests for primitive shapes.
I also checked the gizmo 2d shapes intersection example and everything looked fine.

I did add a few utility methods to the types that have no tests yet. I am willing to implement some if it is deemed necessary

## Migration Guide

The segment type constructors changed so if someone previously created a Segment2d with a direction and length they would now need to use the `from_direction` constructor

### DIFF
--- a/crates/bevy_gizmos/src/primitives/dim2.rs
+++ b/crates/bevy_gizmos/src/primitives/dim2.rs
@@ -635,7 +635,7 @@ where
         }
 
         let segment = Segment2d::new(self.point1, self.point2)
-            .offset(self.isometry.translation)
+            .translated(self.isometry.translation)
             .rotated(self.isometry.rotation);
 
         if self.draw_arrow {
--- a/crates/bevy_gizmos/src/primitives/dim2.rs
+++ b/crates/bevy_gizmos/src/primitives/dim2.rs
@@ -635,8 +635,8 @@ where
         }
 
         let segment = Segment2d::new(self.point1, self.point2)
-            .translated(self.isometry.translation)
-            .rotated_around_center(self.isometry.rotation);
+            .rotated(self.isometry.rotation)
+            .translated(self.isometry.translation);
 
         if self.draw_arrow {
             self.gizmos
--- a/crates/bevy_gizmos/src/primitives/dim2.rs
+++ b/crates/bevy_gizmos/src/primitives/dim2.rs
@@ -540,7 +540,7 @@ where
         }
         // draw normal of the plane (orthogonal to the plane itself)
         let normal = primitive.normal;
-        let normal_segment = Segment2d::from_direction(normal, HALF_MIN_LINE_LEN * 2.);
+        let normal_segment = Segment2d::from_direction_and_length(normal, HALF_MIN_LINE_LEN * 2.);
         self.primitive_2d(
             &normal_segment,
             // offset the normal so it starts on the plane line
--- a/crates/bevy_gizmos/src/primitives/dim2.rs
+++ b/crates/bevy_gizmos/src/primitives/dim2.rs
@@ -636,7 +636,7 @@ where
 
         let segment = Segment2d::new(self.point1, self.point2)
             .translated(self.isometry.translation)
-            .rotated(self.isometry.rotation);
+            .rotated_around_center(self.isometry.rotation);
 
         if self.draw_arrow {
             self.gizmos
--- a/crates/bevy_gizmos/src/primitives/dim2.rs
+++ b/crates/bevy_gizmos/src/primitives/dim2.rs
@@ -540,10 +540,7 @@ where
         }
         // draw normal of the plane (orthogonal to the plane itself)
         let normal = primitive.normal;
-        let normal_segment = Segment2d {
-            direction: normal,
-            half_length: HALF_MIN_LINE_LEN,
-        };
+        let normal_segment = Segment2d::from_direction(normal, HALF_MIN_LINE_LEN * 2.);
         self.primitive_2d(
             &normal_segment,
             // offset the normal so it starts on the plane line
@@ -577,8 +574,8 @@ where
 {
     gizmos: &'a mut GizmoBuffer<Config, Clear>,
 
-    direction: Dir2,  // Direction of the line segment
-    half_length: f32, // Half-length of the line segment
+    point1: Vec2, // First point of the segment
+    point2: Vec2, // Second point of the segment
 
     isometry: Isometry2d, // isometric transformation of the line segment
     color: Color,         // color of the line segment
@@ -616,8 +613,8 @@ where
     ) -> Self::Output<'_> {
         Segment2dBuilder {
             gizmos: self,
-            direction: primitive.direction,
-            half_length: primitive.half_length,
+            point1: primitive.point1(),
+            point2: primitive.point2(),
 
             isometry: isometry.into(),
             color: color.into(),
@@ -637,14 +634,16 @@ where
             return;
         }
 
-        let direction = self.direction * self.half_length;
-        let start = self.isometry * (-direction);
-        let end = self.isometry * direction;
+        let segment = Segment2d::new(self.point1, self.point2)
+            .offset(self.isometry.translation)
+            .rotated(self.isometry.rotation);
 
         if self.draw_arrow {
-            self.gizmos.arrow_2d(start, end, self.color);
+            self.gizmos
+                .arrow_2d(segment.point1(), segment.point2(), self.color);
         } else {
-            self.gizmos.line_2d(start, end, self.color);
+            self.gizmos
+                .line_2d(segment.point1(), segment.point2(), self.color);
         }
     }
 }
--- a/crates/bevy_gizmos/src/primitives/dim3.rs
+++ b/crates/bevy_gizmos/src/primitives/dim3.rs
@@ -228,9 +228,11 @@ where
             return;
         }
 
-        let isometry = isometry.into();
-        let direction = primitive.direction.as_vec3();
-        self.line(isometry * direction, isometry * (-direction), color);
+        let isometry: Isometry3d = isometry.into();
+        let transformed = primitive
+            .rotated(isometry.rotation)
+            .offset(isometry.translation.into());
+        self.line(transformed.point1(), transformed.point1(), color);
     }
 }
 
--- a/crates/bevy_gizmos/src/primitives/dim3.rs
+++ b/crates/bevy_gizmos/src/primitives/dim3.rs
@@ -230,8 +230,8 @@ where
 
         let isometry: Isometry3d = isometry.into();
         let transformed = primitive
-            .rotated(isometry.rotation)
-            .offset(isometry.translation.into());
+            .translated(isometry.translation.into())
+            .rotated_around_center(isometry.rotation);
         self.line(transformed.point1(), transformed.point2(), color);
     }
 }
--- a/crates/bevy_gizmos/src/primitives/dim3.rs
+++ b/crates/bevy_gizmos/src/primitives/dim3.rs
@@ -230,8 +230,8 @@ where
 
         let isometry: Isometry3d = isometry.into();
         let transformed = primitive
-            .translated(isometry.translation.into())
-            .rotated_around_center(isometry.rotation);
+            .rotated(isometry.rotation)
+            .translated(isometry.translation.into());
         self.line(transformed.point1(), transformed.point2(), color);
     }
 }
--- a/crates/bevy_math/src/bounding/bounded2d/primitive_impls.rs
+++ b/crates/bevy_math/src/bounding/bounded2d/primitive_impls.rs
@@ -272,7 +272,7 @@ impl Bounded2d for Segment2d {
     fn bounding_circle(&self, isometry: impl Into<Isometry2d>) -> BoundingCircle {
         let isometry: Isometry2d = isometry.into();
         let local_center = self.center();
-        let radius = local_center.distance(self.point1);
+        let radius = local_center.distance(self.point1());
         let local_circle = BoundingCircle::new(local_center, radius);
         local_circle.transformed_by(isometry.translation, isometry.rotation)
     }
--- a/crates/bevy_math/src/bounding/bounded2d/primitive_impls.rs
+++ b/crates/bevy_math/src/bounding/bounded2d/primitive_impls.rs
@@ -269,7 +269,11 @@ impl Bounded2d for Segment2d {
     }
 
     fn bounding_circle(&self, isometry: impl Into<Isometry2d>) -> BoundingCircle {
-        BoundingCircle::from_point_cloud(isometry, &[self.point1(), self.point2()])
+        let isometry: Isometry2d = isometry.into();
+        let local_center = self.center();
+        let radius = local_center.distance(self.point1);
+        let local_circle = BoundingCircle::new(local_center, radius);
+        local_circle.transformed_by(isometry.translation, isometry.rotation)
     }
 }
 
--- a/crates/bevy_math/src/bounding/bounded2d/primitive_impls.rs
+++ b/crates/bevy_math/src/bounding/bounded2d/primitive_impls.rs
@@ -334,9 +334,8 @@ impl Bounded2d for Triangle2d {
         if let Some((point1, point2)) = side_opposite_to_non_acute {
             // The triangle is obtuse or right, so the minimum bounding circle's diameter is equal to the longest side.
             // We can compute the minimum bounding circle from the line segment of the longest side.
-            let segment = Segment2d::new(point1, point2).centered();
-            let center = (point1 + point2) / 2.;
-            segment.bounding_circle(isometry * Isometry2d::from_translation(center))
+            let segment = Segment2d::new(point1, point2);
+            segment.bounding_circle(isometry)
         } else {
             // The triangle is acute, so the smallest bounding circle is the circumcircle.
             let (Circle { radius }, circumcenter) = self.circumcircle();
--- a/crates/bevy_math/src/bounding/bounded2d/primitive_impls.rs
+++ b/crates/bevy_math/src/bounding/bounded2d/primitive_impls.rs
@@ -411,7 +411,10 @@ impl Bounded2d for Capsule2d {
         let isometry = isometry.into();
 
         // Get the line segment between the semicircles of the rotated capsule
-        let segment = Segment2d::from_direction(isometry.rotation * Dir2::Y, self.half_length * 2.);
+        let segment = Segment2d::from_direction_and_length(
+            isometry.rotation * Dir2::Y,
+            self.half_length * 2.,
+        );
         let (a, b) = (segment.point1(), segment.point2());
 
         // Expand the line segment by the capsule radius to get the capsule half-extents
--- a/crates/bevy_math/src/bounding/bounded2d/primitive_impls.rs
+++ b/crates/bevy_math/src/bounding/bounded2d/primitive_impls.rs
@@ -1,6 +1,7 @@
 //! Contains [`Bounded2d`] implementations for [geometric primitives](crate::primitives).
 
 use crate::{
+    bounding::BoundingVolume,
     ops,
     primitives::{
         Annulus, Arc2d, Capsule2d, Circle, CircularSector, CircularSegment, Ellipse, Line2d,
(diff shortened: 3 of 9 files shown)


crates/bevy_math/src/primitives/dim2.rs
```diff
@@ -1245,27 +1241,82 @@ impl Segment2d {
...
+    }
 
-        (
-            // We are dividing by the length here, so the vector is normalized.
-            Self::new(Dir2::new_unchecked(diff / length), length),
-            (point1 + point2) / 2.,
-        )
+    /// Create a new `Segment2d` at the origin from a `direction` and `half_length`
+    #[inline(always)]
+    pub fn from_direction(direction: Dir2, length: f32) -> Segment2d {
```
IQuick143, line 1251: Not sure if it won't be too verbose, but I suppose the name should be `from_direction_and_length` or something along those lines.

alice-i-cecile: I prefer the clearer, longer name here. Also note that the docs for this are inconsistent.

Sigma-dev: Both should be fixed

crates/bevy_math/src/primitives/dim2.rs
```diff
@@ -1245,27 +1241,82 @@ impl Segment2d {
...
+        self.point1().distance(self.point2())
+    }
+
+    /// Get the segment offset by a vector
+    #[inline(always)]
+    pub fn offset(&self, offset: Vec2) -> Segment2d {
+        Self::new(self.point1() + offset, self.point2() + offset)
+    }
+
+    /// Get the segment rotated around it's center
```
alice-i-cecile, line 1286:
```suggestion
    /// Compute a new segment, based on the original segment rotated around its center
```

alice-i-cecile (changes requested):
Looks good so far. but I'd prefer to do the 3D segment changes in the same PR. This is straightforward to review and I *really* don't want to ship one without the other.

alice-i-cecile (approved):

crates/bevy_math/src/primitives/dim2.rs
```diff
@@ -1245,27 +1241,82 @@ impl Segment2d {
...
+    #[inline(always)]
+    pub fn length(&self) -> f32 {
+        self.point1().distance(self.point2())
+    }
+
+    /// Get the segment offset by a vector
+    #[inline(always)]
+    pub fn offset(&self, offset: Vec2) -> Segment2d {
+        Self::new(self.point1() + offset, self.point2() + offset)
+    }
```
Jondolf, line 1284: `translated` is more consistent with our other APIs. `offset` is also both a noun and a verb, which could be confusing.

```suggestion
    /// Get the segment translated by the given vector
    #[inline(always)]
    pub fn translated(&self, translation: Vec2) -> Segment2d {
        Self::new(self.point1() + translation, self.point2() + translation)
    }
```

crates/bevy_math/src/primitives/dim2.rs
```diff
@@ -1245,27 +1241,82 @@ impl Segment2d {
...
+
+    /// Compute a new segment, based on the original segment rotated around its center
+    #[inline(always)]
+    pub fn rotated(&self, rotation: Rot2) -> Segment2d {
```
Jondolf, line 1288: So in a lot of cases (*most* cases I'd argue), this is not how you'd want rotations to be handled.

Whenever you have a line segment relative to some origin, like say you use line segments to form a triangle or some other polygon, you typically want to rotate them relative to that origin. If you rotate a triangle, its line segments should rotate around the triangle's center, not around their local centers.

Rotating around the origin is also how the transformations for bounding volumes work.

So I would split this to two or three methods:

- `rotated` or `rotated_around_origin`: rotates around the origin.
- `rotated_local` or `rotated_around_center`: rotates around the center of the segment (the current version)
- `rotated_around`: rotates around the given point.

I think I'd like the names `rotated`, `rotated_local`, and `rotated_around` the most, but I'm fine with other names too. This could probably be left to a follow-up too if you'd like.

crates/bevy_math/src/primitives/dim2.rs
```diff
@@ -1245,27 +1241,82 @@ impl Segment2d {
...
+        pub fn rotate_point(p: Vec2, rotation: Rot2) -> Vec2 {
+            Vec2::new(
+                p.x * rotation.cos - p.y * rotation.sin,
+                p.x * rotation.sin + p.y * rotation.cos,
+            )
+        }
```
Jondolf, line 1294: There's no need for this function. `Rot2` is intended to abstract 2D rotations much like `Quat` in 3D, so you can just do `rotation * p`.

crates/bevy_math/src/primitives/dim2.rs
```diff
@@ -1245,27 +1241,82 @@ impl Segment2d {
...
+        let offset_from_origin = self.center();
+        let centered = self.centered();
+        let centered_rotated: Segment2d = Segment2d::new(
+            rotate_point(centered.point1(), rotation),
+            rotate_point(centered.point2(), rotation),
+        );
+        centered_rotated.offset(offset_from_origin)
+    }
+
+    /// Get the segment with it's center is at the origin
```
Jondolf, line 1305:
```suggestion
    /// Get the segment with its center at the origin
```

crates/bevy_math/src/primitives/dim2.rs
```diff
@@ -1245,27 +1241,82 @@ impl Segment2d {
...
+    /// Compute a new segment, based on the original segment rotated around its center
+    #[inline(always)]
+    pub fn rotated(&self, rotation: Rot2) -> Segment2d {
+        pub fn rotate_point(p: Vec2, rotation: Rot2) -> Vec2 {
+            Vec2::new(
+                p.x * rotation.cos - p.y * rotation.sin,
+                p.x * rotation.sin + p.y * rotation.cos,
+            )
+        }
+        // We center the segment for the purpose of the rotation, then offset back to it's original position
```
Jondolf, line 1295:
```suggestion
        // We center the segment for the purpose of the rotation, then offset back to its original position
```

crates/bevy_math/src/primitives/dim3.rs
```diff
@@ -359,22 +359,25 @@ impl Primitive3d for Line3d {}
     reflect(Serialize, Deserialize)
 )]
 pub struct Segment3d {
```
Jondolf, line 361: I won't copy all of my comments from the 2D version, but most of them also apply here in 3D.

crates/bevy_math/src/primitives/dim3.rs
```diff
@@ -359,22 +359,25 @@ impl Primitive3d for Line3d {}
...
     /// Create a new `Segment3d` from a direction and full length of the segment
     #[inline(always)]
-    pub fn new(direction: Dir3, length: f32) -> Self {
-        Self {
-            direction,
-            half_length: length / 2.0,
-        }
+    pub fn from_direction_and_length(direction: Dir3, length: f32) -> Self {
+        let half_length = length / 2.;
+        Self::new(direction * -half_length, direction * -half_length)
```
Jondolf, line 380: This produces a segment with zero length.
```suggestion
    pub fn from_direction_and_length(direction: Dir3, length: f32) -> Self {
        let half_length = length / 2.;
        Self::new(direction * -half_length, direction * half_length)
```

crates/bevy_math/src/primitives/dim3.rs
```diff
@@ -384,26 +387,59 @@ impl Segment3d {
...
-        let diff = point2 - point1;
-        let length = diff.length();
+        let segment = Self::new(point1, point2);
 
-        (
-            // We are dividing by the length here, so the vector is normalized.
-            Self::new(Dir3::new_unchecked(diff / length), length),
-            (point1 + point2) / 2.,
-        )
+        (segment, segment.center())
```
Jondolf, line 392: This should be deprecated like in 2D. Also, why is the implementation here slightly different?

examples/2d/bounding_2d.rs
```diff
@@ -224,13 +224,16 @@ fn setup(mut commands: Commands) {
             Vec2::new(40., 50.),
         )),
         Spin,
-        DesiredVolume::Aabb,
+        DesiredVolume::Circle,
```
Jondolf, line 227: Why was this changed? I believe the example was intended to have AABBs and bounding circles alternating

example

crates/bevy_math/src/primitives/dim2.rs
```diff
@@ -1245,27 +1241,82 @@ impl Segment2d {
...
     pub fn point2(&self) -> Vec2 {
-        *self.direction * self.half_length
+        self.point2
+    }
+
+    /// Get the segment's center
+    #[inline(always)]
+    pub fn center(&self) -> Vec2 {
+        (self.point1() + self.point2()) / 2.
+    }
```
Jondolf, line 1272: I would add a `midpoint` alias since Glam has it for vectors. E.g. this could be `self.point1().midpoint(self.point2())`.
```suggestion
    /// Get the segment's center
    #[inline(always)]
    #[doc(alias = "midpoint")]
    pub fn center(&self) -> Vec2 {
        (self.point1() + self.point2()) / 2.
    }
```
Same for 3D.

crates/bevy_math/src/primitives/dim2.rs
```diff
@@ -1221,22 +1221,18 @@ impl Primitive2d for Line2d {}
 )]
 #[doc(alias = "LineSegment2d")]
 pub struct Segment2d {
```
Jondolf, line 1223: A pretty fundamental question here is, how consistent do we want the simplex types to be? I would argue this should be:
```rust
pub struct Segment2d {
    /// The endpoints of the line segment.
    pub vertices: [Vec2; 2],
}
```
for consistency with `Triangle2d`/`Triangle3d` and `Tetrahedron`. Or alternatively, also switch those types to have each vertex be its own property, but that's a larger rework.

alice-i-cecile: I prefer that!

mweatherley: I'd like if they were all consistent in this way :)

Sigma-dev: I'm not sure I understand, do you prefer point1, point2... or [Vec2, N] ?

crates/bevy_math/src/bounding/bounded2d/primitive_impls.rs
```diff
@@ -336,7 +329,8 @@ impl Bounded2d for Triangle2d {
         if let Some((point1, point2)) = side_opposite_to_non_acute {
             // The triangle is obtuse or right, so the minimum bounding circle's diameter is equal to the longest side.
             // We can compute the minimum bounding circle from the line segment of the longest side.
-            let (segment, center) = Segment2d::from_points(point1, point2);
+            let segment = Segment2d::new(point1, point2).centered();
+            let center = (point1 + point2) / 2.;
```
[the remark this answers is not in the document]
Sigma-dev: I tried the simple approach but it failed in the example and the tests.
I believe it has to do with the old implementation automatically "centering" the segment even if the given points were not centered, but the center returned from from_points was not centered itself.
This weird behavior is what caused me to do this weird code to reproduce the original behavior as close as I could.

This could be a case of something being messed up somewhere else, or just me not understanding the math properly.

crates/bevy_math/src/primitives/dim2.rs
Sigma-dev: I tried but I got `cannot multiply `glam::Vec2` by `Rot2`, maybe i'm doing something wrong

crates/bevy_math/src/primitives/dim3.rs
Sigma-dev: For the things you havn't specificially mentionned for the Segment3d, I have done the translation and rotation changes, tell me if I missed anything else

crates/bevy_math/src/primitives/dim2.rs
```diff
@@ -1245,27 +1241,82 @@ impl Segment2d {
     ///
     /// Panics if `point1 == point2`
     #[inline(always)]
+    #[deprecated(since = "0.15.1", note = "Use the `new` constructor instead")]
```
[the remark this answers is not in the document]
Sigma-dev: It would seem like changing the deprecated version seems to fail the CI job with:
`error: the since field must contain a semver-compliant version`

It's not my domain of expertise so I don't know what to do.

Sigma-dev:
If I didn't miss everything (unlikely) I should have adressed everything except the stuff I left questions about.
I'm open to any new required changes as well of course

crates/bevy_math/src/primitives/dim2.rs
IQuick143: I think Jondolf is suggesting adding the `#[doc(alias = "midpoint")]` attribute for easier discoverability of the method.

crates/bevy_math/src/primitives/dim2.rs
IQuick143: I'd prefer (and I think the others do too) to use `[Vec2; 2]`, to be consistent with other similar pieces of code.
(Also in my opinion this is better because it lets you convert indices to values without a match arm like:
```rust
match i {
  0 => point1,
  1 => point2,
  _ => ...
}
``` Which, while niche here, is imo a good property of the code.)

File: crates/bevy_math/src/primitives/dim2.rs (hunk 12 above)
Comment (Sigma-dev) [reply crates/bevy_math/src/primitives/dim2.rs:1272]: I see but what would `self.point1().midpoint(self.point2())` be about ?

Comment (Jondolf) [reply crates/bevy_math/src/primitives/dim2.rs:1272]: Sorry, my comment was a bit confusing there. I just meant that this `center` method is functionally equivalent to using Glam's `midpoint` method like `segment.point1().midpoint(segment.point2())`, which is why I think adding a `midpoint` alias could be nice here.

So yeah, just adding the alias is what I intended with my suggestion.

File: crates/bevy_math/src/primitives/dim2.rs (hunk 13 above)
Comment (Jondolf) [reply crates/bevy_math/src/primitives/dim2.rs:1223]: Yeah, so change the type to store `pub vertices: [Vec2; 2]`, but probably keep the `point1()` and `point2()` helpers.

File: crates/bevy_math/src/bounding/bounded2d/primitive_impls.rs (hunk 14 above)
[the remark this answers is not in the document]
Comment (Jondolf) [reply crates/bevy_math/src/bounding/bounded2d/primitive_impls.rs:338]: It does pass tests and work in the example for me atleast. Just this:
```diff
- let segment = Segment2d::new(point1, point2).centered();
- let center = (point1 + point2) / 2.;
- segment.bounding_circle(isometry * Isometry2d::from_translation(center))
+ let segment = Segment2d::new(point1, point2);
+ segment.bounding_circle(isometry)
```

crates/bevy_math/src/primitives/dim2.rs
Bleachfuel: you should probably multiply Rot2 by glam::Vec2

crates/bevy_math/src/primitives/dim3.rs
```diff
@@ -359,51 +359,98 @@ impl Primitive3d for Line3d {}
...
+    pub fn rotated_around_center(&self, rotation: Quat) -> Segment3d {
+        self.rotated_around(rotation, self.center())
+    }
+
+    /// Get the segment offset so that it's center is at the origin
+    #[inline(always)]
+    pub fn centered(&self) -> Segment3d {
+        let center = self.center();
+        self.translated(-center)
     }
```
Jondolf, line 454: `resized` is missing from `Segment3d` but is on `Segment2d`

Jondolf (commented):
Some inconsistencies between 2D and 3D

Sigma-dev:
Everything should be fixed now

Jondolf:
This is still unadressed, though I'm fine fixing it in a follow-up too if you'd prefer https://github.com/bevyengine/bevy/pull/17404#discussion_r1921060088

Jondolf (approved):
Looks great now, thanks for your patience on this! I'll add some more helpers and do a small clean-up pass in a follow-up.

crates/bevy_math/src/bounding/bounded2d/primitive_impls.rs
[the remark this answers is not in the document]
Sigma-dev: should be ok now

Sigma-dev:
I appreciate your patience as well, there was quite a lot to fix

crates/bevy_gizmos/src/primitives/dim2.rs
```diff
@@ -637,14 +634,16 @@ where
             return;
         }
 
-        let direction = self.direction * self.half_length;
-        let start = self.isometry * (-direction);
-        let end = self.isometry * direction;
+        let segment = Segment2d::new(self.point1, self.point2)
+            .translated(self.isometry.translation)
+            .rotated_around_center(self.isometry.rotation);
```
Jondolf, line 639: Picture a collider with a segment shape that is offset relative to the transform of the entity. The segment should stay at the same position in the local space of the collider even when that entity rotates. In other words, the segment should rotate around the center of that collider, not around its own center.

This is currently translating the segment to the origin and then applying the isometry, rather than applying the isometry directly to the vertices. I think this should instead be `rotated` rather than rotating around the center of the segment. Also, rotation should be done before translation for [isometries](https://docs.rs/bevy_math/latest/bevy_math/struct.Isometry2d.html).
```suggestion
        let segment = Segment2d::new(self.point1, self.point2)
            .rotated(self.isometry.rotation)
            .translated(self.isometry.translation);
```

crates/bevy_gizmos/src/primitives/dim3.rs
```diff
@@ -228,9 +228,11 @@ where
             return;
         }
 
-        let isometry = isometry.into();
-        let direction = primitive.direction.as_vec3();
-        self.line(isometry * direction, isometry * (-direction), color);
+        let isometry: Isometry3d = isometry.into();
+        let transformed = primitive
+            .translated(isometry.translation.into())
+            .rotated_around_center(isometry.rotation);
```
Jondolf, line 234: Same as in 2D
```suggestion
        let transformed = primitive
            .rotated(isometry.rotation)
            .translated(isometry.translation.into());
```

crates/bevy_math/src/primitives/dim3.rs
```diff
@@ -359,51 +359,108 @@ impl Primitive3d for Line3d {}
...
+        self.point1().distance(self.point2())
+    }
+
+    /// Get the segment translated by a vector
+    #[inline(always)]
+    pub fn translated(&self, translation: Vec3) -> Segment3d {
+        Self::new(self.point1() + translation, self.point2() + translation)
+    }
+
+    /// Get the segment rotated around it's center
```
Jondolf, line 425:
```suggestion
    /// Compute a new segment, based on the original segment rotated around the origin
```

crates/bevy_math/src/primitives/dim2.rs
```diff
@@ -1245,27 +1241,94 @@ impl Segment2d {
...
+            Vec2::new(
+                p.x * rotation.cos - p.y * rotation.sin,
+                p.x * rotation.sin + p.y * rotation.cos,
+            )
+        }
+        Segment2d::new(
+            rotate_point(self.point1(), rotation),
+            rotate_point(self.point2(), rotation),
+        )
+    }
```
Jondolf, line 1300: This was meant to be just this
```suggestion
    /// Compute a new segment, based on the original segment rotated around the origin
    #[inline(always)]
    pub fn rotated(&self, rotation: Rot2) -> Segment2d {
        Segment2d::new(rotation * self.point1(), rotation * self.point2())
    }
```

Jondolf (commented):
Found a few more things

Carter0 (approved):

PR #17413, opened 2025-01-17
Implement experimental GPU two-phase occlusion culling for the standard 3D mesh pipeline.

pcwalton:
*Occlusion culling* allows the GPU to skip the vertex and fragment shading overhead for objects that can be quickly proved to be invisible because they're behind other geometry. A depth prepass already eliminates most fragment shading overhead for occluded objects, but the vertex shading overhead, as well as the cost of testing and rejecting fragments against the Z-buffer, is presently unavoidable for standard meshes. We currently perform occlusion culling only for meshlets. But other meshes, such as skinned meshes, can benefit from occlusion culling too in order to avoid the transform and skinning overhead for unseen meshes.

This commit adapts the same [*two-phase occlusion culling*] technique that meshlets use to Bevy's standard 3D mesh pipeline when the new `OcclusionCulling` component, as well as the `DepthPrepass` component, are present on the camera. It has these steps:

1. *Early depth prepass*: We use the hierarchical Z-buffer from the previous frame to cull meshes for the initial depth prepass, effectively rendering only the meshes that were visible in the last frame.

2. *Early depth downsample*: We downsample the depth buffer to create another hierarchical Z-buffer, this time with the current view transform.

3. *Late depth prepass*: We use the new hierarchical Z-buffer to test all meshes that weren't rendered in the early depth prepass. Any meshes that pass this check are rendered.

4. *Late depth downsample*: Again, we downsample the depth buffer to create a hierarchical Z-buffer in preparation for the early depth prepass of the next frame. This step is done after all the rendering, in order to account for custom phase items that might write to the depth buffer.

Note that this patch has no effect on the per-mesh CPU overhead for occluded objects, which remains high for a GPU-driven renderer due to the lack of `cold-specialization` and retained bins. If `cold-specialization` and retained bins weren't on the horizon, then a more traditional approach like potentially visible sets (PVS) or low-res CPU rendering would probably be more efficient than the GPU-driven approach that this patch implements for most scenes. However, at this point the amount of effort required to implement a PVS baking tool or a low-res CPU renderer would probably be greater than landing `cold-specialization` and retained bins, and the GPU driven approach is the more modern one anyway. It does mean that the performance improvements from occlusion culling as implemented in this patch *today* are likely to be limited, because of the high CPU overhead for occluded meshes.

Note also that this patch currently doesn't implement occlusion culling for 2D objects or shadow maps. Those can be addressed in a follow-up. Additionally, note that the techniques in this patch require compute shaders, which excludes support for WebGL 2.

This PR is marked experimental because of known precision issues with the downsampling approach when applied to non-power-of-two framebuffer sizes (i.e. most of them). These precision issues can, in rare cases, cause objects to be judged occluded that in fact are not. (I've never seen this in practice, but I know it's possible; it tends to be likelier to happen with small meshes.) As a follow-up to this patch, we desire to switch to the [SPD-based hi-Z buffer shader from the Granite engine], which doesn't suffer from these problems, at which point we should be able to graduate this feature from experimental status. I opted not to include that rewrite in this patch for two reasons: (1) @JMS55 is planning on doing the rewrite to coincide with the new availability of image atomic operations in Naga; (2) to reduce the scope of this patch.

A new example, `occlusion_culling`, has been added. It demonstrates objects becoming quickly occluded and disoccluded by dynamic geometry and shows the number of objects that are actually being rendered. Also, a new `--occlusion-culling` switch has been added to `scene_viewer`, in order to make it easy to test this patch with large scenes like Bistro.

[*two-phase occlusion culling*]:
https://medium.com/@mil_kru/two-pass-occlusion-culling-4100edcad501

[Aaltonen SIGGRAPH 2015]:
https://www.advances.realtimerendering.com/s2015/aaltonenhaar_siggraph2015_combined_final_footer_220dpi.pdf

[Some literature]:
https://gist.github.com/reduz/c5769d0e705d8ab7ac187d63be0099b5?permalink_comment_id=5040452#gistcomment-5040452

[SPD-based hi-Z buffer shader from the Granite engine]: https://github.com/Themaister/Granite/blob/master/assets/shaders/post/hiz.comp

## Migration guide

* When enqueuing a custom mesh pipeline, work item buffers are now created with `bevy::render::batching::gpu_preprocessing::get_or_create_work_item_buffer`, not `PreprocessWorkItemBuffers::new`. See the `specialized_mesh_pipeline` example.

## Showcase

Occlusion culling example:
Screenshot 2025-01-15 175051

Bistro zoomed out, before occlusion culling:
Screenshot 2025-01-16 185425

Bistro zoomed out, after occlusion culling:
Screenshot 2025-01-16 184949

In this scene, occlusion culling reduces the number of meshes Bevy has to render from 1591 to 585.


bushrat011899 (commented):
Code looks good, just a minor comment around the experimental module and marking it as `doc(hidden)` for Sem-Ver reasons. I unfortunately couldn't get the new `occlusion_culling` example to run on my laptop (Intel i5-1240p iGPU Windows 10) with either DX12 or the Vulkan backends.

crates/bevy_core_pipeline/src/lib.rs
```diff
@@ -14,6 +14,7 @@ pub mod core_2d;
 pub mod core_3d;
 pub mod deferred;
 pub mod dof;
+pub mod experimental;
```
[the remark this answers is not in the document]
pcwalton: Do we care about semver compatibility here though if we aren't shipping this in a point release? My concern about `#[doc(hidden)]` is that it makes the feature less discoverable, and we want testing on it as it's the thing that could have bugs.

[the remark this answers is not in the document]
bushrat011899: Fair point! While Bevy is pre-1.0 it's probably not important anyway, since every release is a breaking release.

bushrat011899 (approved):
Can confirm the example now runs on my i5-1240p. In the `DX12` backend it says my platform doesn't support occlusion culling, but runs the example fine otherwise. On `Vulkan` it works as expected, culling approximately 30 meshes. Nice work!

tychedelia (commented):
Seeing a panic on my M2 MBP:

```
2025-01-20T01:09:54.639116Z ERROR wgpu::backend::wgpu_core: Handling wgpu errors as fatal by default
thread 'Compute Task Pool (4)' panicked at /Users/char/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/wgpu-23.0.1/src/backend/wgpu_core.rs:996:18:
wgpu error: Validation Error

Caused by:
  In Device::create_bind_group, label = 'preprocess_late_indexed_gpu_occlusion_culling_bind_group'
    Buffer offset 320 does not respect device's requested `min_storage_buffer_offset_alignment` limit 256


Encountered a panic in system `bevy_pbr::render::gpu_preprocess::prepare_preprocess_bind_groups`!
```

crates/bevy_pbr/src/meshlet/visibility_buffer_raster_node.rs
```diff
@@ -311,10 +314,11 @@ impl Node for MeshletVisibilityBufferRasterPassNode {
                 resolve_depth_shadow_view_pipeline,
                 camera,
             );
-            downsample_depth(
+            meshlet_view_resources.depth_pyramid.downsample_depth(
+                "main downsample depth",
```
JMS55, line 318: The labels for the meshlet shadow pass downsample seem wrong to me.

crates/bevy_pbr/src/prepass/mod.rs
```diff
@@ -202,9 +202,6 @@ where
 #[derive(Resource)]
 struct AnyPrepassPluginLoaded;
 
-#[cfg(not(feature = "meshlet"))]
-type PreviousViewFilter = (With<Camera3d>, With<MotionVectorPrepass>);
-#[cfg(feature = "meshlet")]
 type PreviousViewFilter = Or<(With<Camera3d>, With<ShadowView>)>;
```
JMS55, line 205: We can just inline this and remove the type def now.

crates/bevy_core_pipeline/src/core_3d/mod.rs
```diff
@@ -16,7 +16,9 @@ pub mod graph {
     #[derive(Debug, Hash, PartialEq, Eq, Clone, RenderLabel)]
     pub enum Node3d {
         MsaaWriteback,
-        Prepass,
+        EarlyPrepass,
```
JMS55, line 19: Is the early prepass running the _full_ prepass? E.g. for deferred is it doing the gbuffer rendering too?

Not sure, but it might make more sense to do depth only in the early pass, and then depth + other attachments in the late pass.

crates/bevy_pbr/src/meshlet/cull_clusters.wgsl
```diff
@@ -82,19 +83,15 @@ fn cull_clusters(
     let occlusion_culling_bounding_sphere_center_view_space = (view.view_from_world * vec4(occlusion_culling_bounding_sphere_center.xyz, 1.0)).xyz;
 #endif
 
-    var aabb = project_view_space_sphere_to_screen_space_aabb(occlusion_culling_bounding_sphere_center_view_space, occlusion_culling_bounding_sphere_radius);
```
JMS55, line 85: Please do not change cull_clusters.wgsl. I am modifying the occlusion culling soon to use AABBs instead.

Feel free to copy the code to use in other shaders, but leave this here.

crates/bevy_core_pipeline/src/experimental/mip_generation/downsample_depth.wgsl
```diff
@@ -1,8 +1,16 @@
 #ifdef MESHLET_VISIBILITY_BUFFER_RASTER_PASS_OUTPUT
 @group(0) @binding(0) var<storage, read> mip_0: array<u64>; // Per pixel
```
JMS55, line 2: I believe all the meshlet-specific stuff is going to disappear here once wgpu 24 is merged and I can switch back to image atomics.

JMS55:
Did part of my review, will do the rest another time.

Focused mainly on meshlets, the depth downsample, and culling test parts. Haven't yet looked at the code for applying the occlusion culling to our main pipeline.

crates/bevy_pbr/src/prepass/mod.rs
JMS55: I only added the typedef in the first place to deal with the meshlet feature. In hindsight, not actually all that necessary, this probably costs very little perf (unlike the previous instance system).

crates/bevy_pbr/src/meshlet/cull_clusters.wgsl
pcwalton: I don't agree with having duplicate code, but OK.

crates/bevy_core_pipeline/src/core_3d/mod.rs
pcwalton: I'm pretty sure that Griffin said that you usually split the full prepass into early and late phases rather than having a separate z-prepass, but I'd rather not make that change here as this patch is too big as it is. Instead I just documented that occlusion culling is currently incompatible with deferred. We can add support for deferred in a followup.

crates/bevy_pbr/src/meshlet/visibility_buffer_raster_node.rs
pcwalton: Ah, I didn't realize that was for shadows. Makes sense. Should be fixed now

crates/bevy_core_pipeline/src/experimental/mip_generation/mod.rs
```diff
@@ -0,0 +1,750 @@
...
+    fn run<'w>(
+        &self,
+        _: &mut RenderGraphContext,
+        render_context: &mut RenderContext<'w>,
+        (view_depth_pyramid, view_downsample_depth_bind_group, view_depth_texture): QueryItem<
+            'w,
+            Self::ViewQuery,
+        >,
+        world: &'w World,
+    ) -> Result<(), NodeRunError> {
```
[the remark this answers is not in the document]
pcwalton: Done.

crates/bevy_core_pipeline/src/experimental/mip_generation/mod.rs
```diff
@@ -0,0 +1,750 @@
...
+                } else {
+                    texture_2d(TextureSampleType::Depth)
+                },
+                // All the mip levels follow:
+                texture_storage_2d(TextureFormat::R32Float, StorageTextureAccess::WriteOnly),
+                texture_storage_2d(TextureFormat::R32Float, StorageTextureAccess::WriteOnly),
+                texture_storage_2d(TextureFormat::R32Float, StorageTextureAccess::WriteOnly),
+                texture_storage_2d(TextureFormat::R32Float, StorageTextureAccess::WriteOnly),
+                texture_storage_2d(TextureFormat::R32Float, StorageTextureAccess::WriteOnly),
+                texture_storage_2d(TextureFormat::R32Float, StorageTextureAccess::ReadWrite),
```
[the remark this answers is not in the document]
pcwalton: We call `textureStore` on it. See `mip_6` in `downsample_depth.wgsl`.

crates/bevy_pbr/src/prepass/mod.rs
pcwalton: Done.

pcwalton:
I believe I've fixed all the issues that I know about and have addressed the review comments. See the commit descriptions for more information.

crates/bevy_core_pipeline/src/experimental/mip_generation/mod.rs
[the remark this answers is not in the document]
tychedelia: Ah yup, I see it's the handoff point between first and second.

tychedelia (approved):
Crash on mac fixed. Lgtm! ✨

crates/bevy_pbr/src/render/gpu_preprocess.rs
```diff
@@ -673,77 +1256,171 @@ impl FromWorld for PreprocessPipelines {
...
+            early_phase: preprocess_phase_pipelines.clone(),
+            late_phase: preprocess_phase_pipelines.clone(),
+            main_phase: preprocess_phase_pipelines.clone(),
         }
     }
 }
 
 fn preprocess_direct_bind_group_layout_entries() -> DynamicBindGroupLayoutEntries {
-    DynamicBindGroupLayoutEntries::sequential(
+    DynamicBindGroupLayoutEntries::new_with_indices(
```
atlv24, line 1309: I wasn't aware we had a `new_with_indices`, this is much nicer to use

crates/bevy_pbr/src/render/mesh_preprocess.wgsl
```diff
@@ -129,6 +185,115 @@ fn main(@builtin(global_invocation_id) global_invocation_id: vec3<u32>) {
...
+        // transform.
+        let world_pos = (world_from_local * vec4(local_pos, 1.0)).xyz;
+        let view_pos = position_world_to_view(world_pos);
+        let ndc_pos = position_world_to_ndc(world_pos);
+#endif  // EARLY_PHASE
+
+        let uv_pos = ndc_to_uv(ndc_pos.xy);
+
+        // Update the AABB and maximum view-space depth.
+        if (i == 0u) {
```
atlv24, line 240: this `i == 0u` case can be removed by initializing `max_depth_view` to `-inf`, and `aabb` to `vec4(inf, inf, -inf, -inf)` i think

crates/bevy_pbr/src/render/occlusion_culling.wgsl
```diff
@@ -0,0 +1,30 @@
...
+
+    let depth_pyramid_size_mip_0 = vec2<f32>(textureDimensions(depth_pyramid, 0));
+    let depth_level = max(0, i32(ceil(log2(max(aabb_width_pixels, aabb_height_pixels))))); // TODO: Naga doesn't like this being a u32
+    let depth_pyramid_size = vec2<f32>(textureDimensions(depth_pyramid, depth_level));
+    let aabb_top_left = vec2<u32>(aabb.xy * depth_pyramid_size);
+
+    let depth_quad_a = textureLoad(depth_pyramid, aabb_top_left, depth_level).x;
+    let depth_quad_b = textureLoad(depth_pyramid, aabb_top_left + vec2(1u, 0u), depth_level).x;
+    let depth_quad_c = textureLoad(depth_pyramid, aabb_top_left + vec2(0u, 1u), depth_level).x;
+    let depth_quad_d = textureLoad(depth_pyramid, aabb_top_left + vec2(1u, 1u), depth_level).x;
```
atlv24, line 28: i believe it is recommended to sample 16 pixels from a 1-step-finer mip level. this is definitely something to punt to another pr though

crates/bevy_render/src/experimental/occlusion_culling/mod.rs
```diff
@@ -0,0 +1,87 @@
...
+/// Occlusion culling inherently has some overhead, because Bevy must examine
+/// the objects' bounding boxes, and create an acceleration structure
+/// (hierarchical Z-buffer) to perform the occlusion tests. Therefore, occlusion
+/// culling is disabled by default. Only enable it if you measure it to be a
+/// speedup on your scene. Note that, because Bevy's occlusion culling runs on
+/// the GPU and is quite efficient, it's rare for occlusion culling to result in
+/// a significant slowdown.
+///
+/// Occlusion culling currently requires a `DepthPrepass`. If no depth prepass
+/// is present on the view, the [`OcclusionCulling`] component will be ignored.
```
atlv24, line 61: Maybe have OcclusionCulling required component DepthPrepass?

atlv24 (approved):
Another herculean effort from you on all fronts, the docs are great thanks.

a few questions
- whats the path forward for occlusion culling shadow views? not planned, not worth it, or worth it but annoying to do/not now?
- how does hzb test interact with TAA jitter?
- what frustum are you culling by for early depth pass? culling by only prev view frustum will render more than necessary, and culling only by current view frustum will test against unwritten parts of the prev hzb, likely resulting in overdraw. this overdraw cost will have to be eaten either in the early pass or the depth pass though because we do not have information for the newly disoccluded region of the screen, so i don't think it matters, meaning culling by current view is ideal.

crates/bevy_render/src/experimental/occlusion_culling/mod.rs
atlv24: i realize that the current way is just aligning with how all the other prepasses do it. lets punt on this to a required-components prepass migration pr

pcwalton: I tried to do that but the problem is that `DepthPrepass` lives in `bevy_core_pipeline` which is downstream of `bevy_render`, so `OcclusionCulling` can't refer to it.

pcwalton:
> whats the path forward for occlusion culling shadow views? not planned, not worth it, or worth it but annoying to do/not now?

Planned, but didn't want to do it in this patch since it'd make it bigger and more complex.

> how does hzb test interact with TAA jitter?

I think TAA jitter is basically just a different view matrix, so it's essentially just the same as a regular camera movement (i.e. it should just work).

> what frustum are you culling by for early depth pass? culling by only prev view frustum will render more than necessary, and culling only by current view frustum will test against unwritten parts of the prev hzb, likely resulting in overdraw. this overdraw cost will have to be eaten either in the early pass or the depth pass though because we do not have information for the newly disoccluded region of the screen, so i don't think it matters, meaning culling by current view is ideal.

It tests against the current frame frustum.

crates/bevy_pbr/src/render/mesh_preprocess.wgsl
pcwalton: OK, I did that. Note that I had to use a bitcast because Naga complained if I tried `1.0 / 0.0` or `-1.0 / 0.0`.

pcwalton:
Bevy example runner output looks good; all the changes seem to be false positives.

crates/bevy_pbr/src/render/occlusion_culling.wgsl
JMS55: I haven't heard of that. I have seen using a 3x3 sample based off a condition (I forget what) though. Something we can improve on in a future PR.

alice-i-cecile:
Doing a final example runner test and then merging.

alice-i-cecile:
Thank you to everyone involved with the authoring or reviewing of this PR! This work is relatively important and needs release notes! Head over to https://github.com/bevyengine/bevy-website/issues/1988 if you'd like to help out.

PR #17415, opened 2025-01-17
Improve `TextSpan` docs

### LINKED ISSUE #16794: TextUiReader requires a parent Text to function, despite TextSpan docs
## Bevy version

Bevy 0.15.0

## What you did

I constructed a root `TextLayout` with child `TextSpans`, as per the [`TextSpan` documentation](https://docs.rs/bevy/latest/bevy/prelude/struct.TextSpan.html).

## What went wrong

`TextUiReader` should allow me to properly iterate over the child `TextSpan`s of the entity with `TextLayout`, but it will return an empty iterator if there is no root parent `Text` component.

I'm not sure what the expected behavior is here; if I don't want to use a root `Text` component, should `TextUiReader` still work? Or is the example for `TextSpan` incorrect?

rparrett:
# Objective

Our [`TextSpan`](https://docs.rs/bevy/latest/bevy/prelude/struct.TextSpan.html) docs include a code example that does not actually "work." The code silently does not render anything, and the `Text*Writer` helpers fail.

This seems to be by design, because we can't use `Text` or `Text2d` from `bevy_ui` or `bevy_sprite` within docs in `bevy_text`. (Correct me if I am wrong)

I have seen multiple users confused by these docs.

Also fixes #16794

## Solution

Remove the code example from `TextSpan`, and instead encourage users to seek docs on `Text` or `Text2d`.

Add examples with nested `TextSpan`s in those areas.

### DIFF
--- a/crates/bevy_text/src/text.rs
+++ b/crates/bevy_text/src/text.rs
@@ -158,7 +158,10 @@ impl TextLayout {
     }
 }
 
-/// A span of UI text in a tree of spans under an entity with [`TextLayout`] and `Text` or `Text2d`.
+/// A span of text in a tree of spans.
+///
+/// `TextSpan` is only valid as a child of an entity with [`TextLayout`], which is provided by `Text`
+/// for text in `bevy_ui` or `Text2d` for text in 2d world-space.
 ///
 /// Spans are collected in hierarchy traversal order into a [`ComputedTextBlock`] for layout.
 ///
--- a/crates/bevy_text/src/text.rs
+++ b/crates/bevy_text/src/text.rs
@@ -173,6 +173,8 @@ impl TextLayout {
 /// # let mut world = World::default();
 /// #
 /// world.spawn((
+///     // `Text` or `Text2d` are needed, and will provide default instances
+///     // of the following components.
 ///     TextLayout::default(),
 ///     TextFont {
 ///         font: font_handle.clone().into(),
@@ -182,6 +184,7 @@ impl TextLayout {
 ///     TextColor(BLUE.into()),
 /// ))
 /// .with_child((
+///     // Children must be `TextSpan`, not `Text` or `Text2d`.
 ///     TextSpan::new("Hello!"),
 ///     TextFont {
 ///         font: font_handle.into(),
--- a/crates/bevy_text/src/text2d.rs
+++ b/crates/bevy_text/src/text2d.rs
@@ -47,7 +47,7 @@ use bevy_window::{PrimaryWindow, Window};
 /// ```
 /// # use bevy_asset::Handle;
 /// # use bevy_color::Color;
-/// # use bevy_color::palettes::basic::{BLUE, RED};
+/// # use bevy_color::palettes::basic::BLUE;
 /// # use bevy_ecs::world::World;
 /// # use bevy_ecs::prelude::Component;
 /// # use bevy_hierarchy::{ChildBuild, BuildChildren};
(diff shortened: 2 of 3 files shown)


crates/bevy_text/src/text.rs
```diff
@@ -170,11 +170,13 @@ impl TextLayout {
...
 /// # use bevy_hierarchy::BuildChildren;
-///
+/// #
 /// # let font_handle: Handle<Font> = Default::default();
 /// # let mut world = World::default();
 /// #
 /// world.spawn((
+///     // `Text` or `Text2d` are needed, and will provide default instances
+///     // of the following components.
 ///     TextLayout::default(),
```
ickshonpe, line 180: Shouldn't this have a `Text` or `Text2d` component here as part of the example?

rparrett: Yep, but this is in `bevy_text`, which doesn't have `bevy_ui` and/or `bevy_sprite` as a dependency, so I don't think we can do that?

ickshonpe: Yeah I think that would be best.

crates/bevy_text/src/text2d.rs
```diff
@@ -47,13 +47,18 @@ use bevy_window::{PrimaryWindow, Window};
...
-/// # use bevy_text::{Font, JustifyText, Text2d, TextLayout, TextFont, TextColor};
+/// # use bevy_ecs::prelude::Component;
+/// # use bevy_hierarchy::{ChildBuild, BuildChildren};
+/// # use bevy_text::{Font, JustifyText, Text2d, TextLayout, TextFont, TextColor, TextSpan};
 /// #
 /// # let font_handle: Handle<Font> = Default::default();
 /// # let mut world = World::default();
 /// #
+/// # #[derive(Component)]
+/// # struct WorldSpan;
```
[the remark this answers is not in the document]
rparrett: To demonstrate that arbitrary marker components can be placed on text spans. Maybe that's confusing? I don't mind getting rid of it.

Carter0 (approved):

ickshonpe (approved):

mockersf (approved):

crates/bevy_text/src/text2d.rs
```diff
@@ -49,7 +49,9 @@ use bevy_window::{PrimaryWindow, Window};
 /// # use bevy_color::Color;
 /// # use bevy_color::palettes::basic::BLUE;
 /// # use bevy_ecs::world::World;
-/// # use bevy_text::{Font, JustifyText, Text2d, TextLayout, TextFont, TextColor};
+/// # use bevy_ecs::prelude::Component;
+/// # use bevy_hierarchy::{ChildBuild, BuildChildren};
```
mockersf, line 53: doesn't exist anymore

crates/bevy_ui/src/widget/text.rs
```diff
@@ -63,7 +63,9 @@ impl Default for TextNodeFlags {
 /// # use bevy_color::Color;
 /// # use bevy_color::palettes::basic::BLUE;
 /// # use bevy_ecs::world::World;
-/// # use bevy_text::{Font, JustifyText, TextLayout, TextFont, TextColor};
+/// # use bevy_ecs::prelude::Component;
+/// # use bevy_hierarchy::{ChildBuild, BuildChildren};
```
mockersf, line 67: doesn't exist anymore

crates/bevy_text/src/text2d.rs
rparrett: Ugh, okay, thanks for the poke. I didn't realize this failed to merge.

Issue #14852, opened 2024-08-21
`AnimationPlayer` doesn't explain how it is created


aecsocket:
## How can Bevy's documentation be improved?

[`bevy_animation::AnimationPlayer`](https://docs.rs/bevy_animation/latest/bevy_animation/struct.AnimationPlayer.html) doesn't explain:
- how you should create an animation player
- if you even *should* create an animation player manually
- how this interacts with `AnimationTransitions` and glTF scenes

When I was trying to spawn an entity with animations, I was following the animated fox example, but tried to simplify it. My code looked approximately like:
```rust
let mut anim_player = AnimationPlayer::default();
let mut transitions = AnimationTransitions::new();

transitions
    .play(&mut anim_player, idle_animation, Duration::ZERO)
    .repeat();

commands
    .spawn((
        SceneBundle {
            scene: my_entity_gltf_scene.clone(),
            ..default()
        },
        anim_player,
        transitions,
        my_anim_graph.clone(),
    ));
```

However, this is wrong because here I'm creating my own `AnimationPlayer`, when I should actually be using the player created by the scene when it's instantiated. What I actually have to do is spawn the SceneBundle in one system, wait for the scene to be instantiated, and use a second system to use the automatically-created `AnimationPlayer` (I assume because the scene automatically populates it with some important data?). It's not really documented that this is specifically how you're meant to do this.

aevyrie:
Ran into this as well. Trying to use the animation system to play a GLTF animation is pretty frustrating right now, and requires you to pipe data around without any docs or explanation. There is also no documentation on `AnimationPlayer`.

mockersf:
not saying this is friendly or easy to discover, but here's what I do:

```

#[derive(Resource)]
struct Animation {
    animation: AnimationNodeIndex,
    graph: Handle<AnimationGraph>,
}

pub fn setup(
    mut commands: Commands,
    asset_server: Res<AssetServer>,
    mut graphs: ResMut<Assets<AnimationGraph>>,
) {
    let (graph, node) = AnimationGraph::from_clip(
        asset_server.load(GltfAssetLabel::Animation(2).from_asset("my_file.gltf")),
    );

    let graph_handle = graphs.add(graph);
    commands.insert_resource(Animation {
        animation: node,
        graph: graph_handle,
    });

    commands
        .spawn(SceneRoot(
            asset_server.load(GltfAssetLabel::Scene(0).from_asset("my_file.gltf")),
        ))
        .observe(play_animation);
[...]

PR #17421, opened 2025-01-18
Make the `animated_mesh` example more intuitive 

greeble-dev:
# Objective

Make the `animated_mesh` example more intuitive and easier for the user to extend.

# Solution

The `animated_mesh` example shows how to spawn a single mesh and play a single animation. The original code is roughly:

1. In `setup_mesh_and_animation`, spawn an entity with a SceneRoot that will load and spawn the mesh. Also record the animation to play as a resource.
2. Use  `play_animation_once_loaded` to detect when any animation players are spawned, then play the animation from the resource.

When I used this example as a starting point for my own app, I hit a wall when trying to spawn multiple meshes with different animations. `play_animation_once_loaded` tells me an animation player spawned somewhere, but how do I get from there to the right animation? The entity it runs on is spawned by the scene so I can't attach any data to it?

The new code takes a different approach. Instead of a global resource, the animation is recorded as a component on the entity with the SceneRoot. Instead of detecting animation players spawning wherever, an observer is attached to that specific entity.

This feels more intuitive and localised, and I think most users will work out how to get from there to different animations and meshes. The downside is more lines of code, and the "find the animation players" part still feels a bit magical and inefficient.

# Side Notes

- The solution was mostly stolen from https://github.com/bevyengine/bevy/issues/14852#issuecomment-2481401769.
- The example still feels too complicated.
    - "Why do I have to make this graph to play one animation?"
    - "Why can't I choose and play the animation in one step and avoid this temporary component?"
    - I think this requires engine changes.
- I originally started on a separate example of multiple meshes ([branch](https://github.com/bevyengine/bevy/compare/main...greeble-dev:bevy:animated-mesh-multiple)).
    - I decided that the user could probably work this out themselves from the single animation example.
    - But maybe still worth following through.

# Testing

`cargo run --example animated_mesh`

### DIFF
--- a/examples/animation/animated_mesh.rs
+++ b/examples/animation/animated_mesh.rs
@@ -21,7 +21,7 @@ fn main() {
 }
 
 // A component that records what animation we want to play. This is created when
-// we start loading the mesh (see `setup_mesh_and_animation`) and picked up when
+// we start loading the mesh (see `setup_mesh_and_animation`) and read when
 // the mesh is spawned (see `play_animation_once_loaded`).
 #[derive(Component)]
 struct AnimationToPlay {
@@ -43,7 +43,7 @@ fn setup_mesh_and_animation(
     // Register the animation graph as an asset.
     let graph_handle = graphs.add(graph);
 
-    // Create a component that will spawn our mesh after it has loaded.
+    // Create a SceneRoot component that will spawn our mesh after it has loaded.
     let mesh_scene = SceneRoot(asset_server.load(GltfAssetLabel::Scene(0).from_asset(GLTF_PATH)));
 
     // Create a component that records which animation we want to play on the
@@ -53,7 +53,7 @@ fn setup_mesh_and_animation(
         index,
     };
 
-    // Spawn an entity containing our components, and connect it to our
+    // Spawn an entity with our components and connect it to our
     // play_animation_once_loaded trigger.
     commands
         .spawn((mesh_scene, animation_to_play))
@@ -67,12 +67,12 @@ fn play_animation_once_loaded(
     animations_to_play: Query<&AnimationToPlay>,
     mut players: Query<&mut AnimationPlayer>,
 ) {
-    // The entity we spawned in setup_mesh_and_animation is the trigger's target.
+    // The entity we spawned in `setup_mesh_and_animation` is the trigger's target.
     // Start by finding the AnimationToPlay component we added to that entity.
     if let Ok(animation_to_play) = animations_to_play.get(trigger.target()) {
-        // The mesh and an animation player will have been spawned in a
-        // descendent of this entity. Search the descendents to find the
-        // animation player.
+        // The SceneRoot component will have spawned the mesh and an animation
+        // player as a descendent of our entity. Search the descendents to find
+        // the animation player.
         for child in children.iter_descendants(trigger.target()) {
             if let Ok(mut player) = players.get_mut(child) {
                 // Tell the animation player to start the animation and keep
--- a/examples/animation/animated_mesh.rs
+++ b/examples/animation/animated_mesh.rs
@@ -89,10 +89,6 @@ fn play_animation_when_ready(
                 commands
                     .entity(child)
                     .insert(AnimationGraphHandle(animation_to_play.graph_handle.clone()));
-
-                // We know that our scene will only contain a single mesh and
-                // animation player, so stop searching.
-                break;
             }
         }
     }
--- a/examples/animation/animated_mesh.rs
+++ b/examples/animation/animated_mesh.rs
@@ -86,6 +86,10 @@ fn play_animation_once_loaded(
                 commands
                     .entity(child)
                     .insert(AnimationGraphHandle(animation_to_play.graph_handle.clone()));
+
+                // We know that our scene will only contain a single mesh and
+                // animation player, so stop searching.
+                break;
             }
         }
     }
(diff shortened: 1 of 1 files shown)


examples/animation/animated_mesh.rs
```diff
@@ -17,62 +17,78 @@ fn main() {
...
-    // Tell the engine to start loading our mesh and animation, and then spawn
-    // them as a scene when ready.
-    commands.spawn(SceneRoot(
-        asset_server.load(GltfAssetLabel::Scene(0).from_asset(GLTF_PATH)),
-    ));
+    // Spawn an entity with our components and connect it to our
+    // play_animation_once_loaded trigger.
+    commands
+        .spawn((mesh_scene, animation_to_play))
+        .observe(play_animation_once_loaded);
```
rparrett, line 60: Nits:

Why create these bindings that are used only once? Just for tidyness of the comments? I think that something like below looks fine.

Some suggestions for the comments themselves down there too.

- I don't think we need to explain that `spawn` creates an entity with the given components at this stage
- It would be great to mention that the mesh will be buried in a scene hierarchy

```rust
    commands
        .spawn((
            // The SceneRoot component will spawn the scene hierarchy, including our mesh
            // after the scene has loaded.
            SceneRoot(asset_server.load(GltfAssetLabel::Scene(0).from_asset(GLTF_PATH))),
            // Store the specific animation we want to play once the mesh has spawned.
            AnimationToPlay {
                graph_handle,
                index,
            },
        ))
        // This observer is triggered `SceneInstanceReady` and will run once the scene has spawned
        .observe(play_animation_once_loaded);
```

examples/animation/animated_mesh.rs
```diff
@@ -17,62 +17,78 @@ fn main() {
         .add_plugins(DefaultPlugins)
         .add_systems(Startup, setup_mesh_and_animation)
         .add_systems(Startup, setup_camera_and_environment)
-        .add_systems(Update, play_animation_once_loaded)
         .run();
 }
 
-#[derive(Resource)]
-struct Animations {
+// A component that records what animation we want to play. This is created when
```
rparrett, line 23: "records what animation" feels a little strange.

Maybe "stores a reference to the animation we want to play"?

examples/animation/animated_mesh.rs
```diff
@@ -17,62 +17,78 @@ fn main() {
...
+    // animation from our example asset, which has an index of two.
     let (graph, index) = AnimationGraph::from_clip(
-        // We want the "run" animation from our example asset, which has an
-        // index of two.
         asset_server.load(GltfAssetLabel::Animation(2).from_asset(GLTF_PATH)),
     );
 
-    // Keep our animation graph in a Resource so that it can be added to the
-    // correct entity once the scene loads.
+    // Register the animation graph as an asset.
```
rparrett, line 43: I might avoid the word "register" which is slightly loaded in this engine. "Store"?

examples/animation/animated_mesh.rs
```diff
@@ -17,62 +17,78 @@ fn main() {
...
-        // Insert the animation graph with our selected animation. This
-        // connects the animation player to the mesh.
-        commands
-            .entity(entity)
-            .insert(AnimationGraphHandle(animations.graph_handle.clone()));
+                // Add the animation graph. This only needs to be done once to
+                // connect the animation player to the mesh.
+                commands
+                    .entity(child)
+                    .insert(AnimationGraphHandle(animation_to_play.graph_handle.clone()));
```
rparrett, line 89: Maybe a `break` statement here?

rparrett (commented):
I dig the major changes here. Have some minor thoughts about comment tweaks.

Carter0 (approved):

examples/animation/animated_mesh.rs
greeble-dev: Updated, although I changed "the animation" to "an animation". Which reads a little clunky, but hints that the component can be reused with different animations.

examples/animation/animated_mesh.rs
greeble-dev: Added a break and a comment, although I wonder if this is over complicating things.

```rust
// We know that our scene will only contain a single mesh and
// animation player, so stop searching.
break;
```

examples/animation/animated_mesh.rs
greeble-dev: I've made updates in https://github.com/bevyengine/bevy/pull/17421/commits/36f62fa7459bb6a7eb9c5b96db02f06e2fba1b86 

I've kept the step by step setup. Maybe just personal taste but I think it's clearer for beginners. 

On the "spawn an entity" comment, I think this makes the narrative flow. Also updated it to emphasise that the observer is connected to that particular entity - I was worried the user might assume it's a global thing.

I've expanded a bit on the scene and hierarchy. I chose to explain the hierarchy in the trigger as that's where it really comes into play.

examples/animation/animated_mesh.rs
rparrett: I think you're probably right and this isn't a big deal / is too noisy.

examples/animation/animated_mesh.rs
```diff
@@ -17,62 +17,80 @@ fn main() {
...
-        // If you want to try stopping and switching animations, see the
-        // `animated_mesh_control.rs` example.
-        player.play(animations.index).repeat();
+    // The entity we spawned in `setup_mesh_and_animation` is the trigger's target.
+    // Start by finding the AnimationToPlay component we added to that entity.
+    if let Ok(animation_to_play) = animations_to_play.get(trigger.target()) {
+        // The SceneRoot component will have spawned the scene as a hierarchy
+        // of entities parented to our entity. Since the asset contained a skinned
+        // mesh and animations, it will also have spawned an animation player
+        // component. Search our entity's descendents to find the animation player.
```
rparrett, line 77:
```suggestion
        // component. Search our entity's descendants to find the animation player.
```

rparrett (approved):

PR #17423, opened 2025-01-18
BRP resource methods

mweatherley:
# Objective

So far, built-in BRP methods allow users to interact with entities' components, but global resources have remained beyond its reach. The goal of this PR is to take the first steps in rectifying this shortfall.

## Solution

Added five new default methods to BRP:
- `bevy/get_resource`: Extracts the value of a given resource from the world.
- `bevy/insert_resource`: Serializes an input value to a given resource type and inserts it into the world.
- `bevy/remove_resource`: Removes the given resource from the world.
- `bevy/mutate_resource`: Replaces the value of a field in a given resource with the result of serializing a given input value.
- `bevy/list_resources`: Lists all resources in the type registry with an available `ReflectResource`. 

## Testing

Added a test resource to the `server` example scene that you can use to mess around with the new BRP methods.

## Showcase

Resources can now be retrieved and manipulated remotely using a handful of new BRP methods. For example, a resource that looks like this:
```rust
#[derive(Resource, Reflect, Serialize, Deserialize)]
#[reflect(Resource, Serialize, Deserialize)]
pub struct PlayerSpawnSettings {
  pub location: Vec2,
  pub lives: u8,
}
```

can be manipulated remotely as follows.

Retrieving the value of the resource:
```json
{
  "jsonrpc": "2.0",
  "id": 1,
  "method": "bevy/get_resource",
  "params": {
    "resource": "path::to::my::module::PlayerSpawnSettings"
  }
}
```

Inserting a resource value into the world:
```json
{
  "jsonrpc": "2.0",
  "id": 2,
  "method": "bevy/insert_resource",
  "params": {
    "resource": "path::to::my::module::PlayerSpawnSettings",
    "value": {
      "location": [
        2.5, 
        2.5
      ],
      "lives": 25
    }
  }
}
```

Removing the resource from the world:
```json
{
  "jsonrpc": "2.0",
  "id": 3,
  "method": "bevy/remove_resource",
  "params": {
    "resource": "path::to::my::module::PlayerSpawnSettings"
  }
}
```

Mutating a field of the resource specified by a path:
```json
{
  "jsonrpc": "2.0",
  "id": 4,
  "method": "bevy/mutate_resource",
  "params": {
    "resource": "path::to::my::module::PlayerSpawnSettings",
    "path": ".location.x",
    "value": -3.0
  }
}
```

Listing all manipulable resources in the type registry:
```json
{
  "jsonrpc": "2.0",
  "id": 5,
  "method": "bevy/list_resources"
}
```

### DIFF
--- a/crates/bevy_remote/src/builtin_methods.rs
+++ b/crates/bevy_remote/src/builtin_methods.rs
@@ -17,7 +17,7 @@ use bevy_hierarchy::BuildChildren as _;
 use bevy_reflect::{
     prelude::ReflectDefault,
     serde::{ReflectSerializer, TypedReflectDeserializer},
-    GetPath as _, NamedField, OpaqueInfo, PartialReflect, ReflectDeserialize, ReflectSerialize,
+    GetPath, NamedField, OpaqueInfo, PartialReflect, ReflectDeserialize, ReflectSerialize,
     TypeInfo, TypeRegistration, TypeRegistry, VariantInfo,
 };
 use bevy_utils::HashMap;
@@ -58,6 +58,21 @@ pub const BRP_GET_AND_WATCH_METHOD: &str = "bevy/get+watch";
 
 /// The method path for a `bevy/list+watch` request.
 pub const BRP_LIST_AND_WATCH_METHOD: &str = "bevy/list+watch";
+
+/// The method path for a `bevy/get_resource` request.
+pub const BRP_GET_RESOURCE_METHOD: &str = "bevy/get_resource";
+
+/// The method path for a `bevy/insert_resource` request.
+pub const BRP_INSERT_RESOURCE_METHOD: &str = "bevy/insert_resource";
+
+/// The method path for a `bevy/remove_resource` request.
+pub const BRP_REMOVE_RESOURCE_METHOD: &str = "bevy/remove_resource";
+
+/// The method path for a `bevy/mutate_resource` request.
+pub const BRP_MUTATE_RESOURCE_METHOD: &str = "bevy/mutate_resource";
+
+/// The method path for a `bevy/list_resources` request.
+pub const BRP_LIST_RESOURCES_METHOD: &str = "bevy/list_resources";
 
 /// The method path for a `bevy/registry/schema` request.
 pub const BRP_REGISTRY_SCHEMA_METHOD: &str = "bevy/registry/schema";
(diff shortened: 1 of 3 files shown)


purplg:
This looks good to me. Excited to see these new methods implemented.

Do we want to consider renaming the methods to be a little more hierarchical in the future or may some other scheme? For example:

- `bevy/mutate_component` -> `bevy/component/mutate`
- `bevy/list_resource` -> `bevy/resource/list`
- `bevy/insert_resource` -> `bevy/resource/insert`
- `bevy/remove_resource` -> `bevy/resource/remove`
- `bevy/mutate_resource` -> `bevy/resource/mutate`

I feel like they may get a little unwieldy over time as more and more methods are added.

mweatherley:
> Do we want to consider renaming the methods to be a little more hierarchical in the future or maybe some other scheme? 

This is a very good idea. I'll change them when I have time :) 

(I guess there's also the question of which other methods, if any, should get moved to the `bevy/component/` namespace.)

alice-i-cecile (approved):
Straightforward, tested and uncontroversially useful. Thanks!

alice-i-cecile:
@Leinnan I actually need a second review over here. Can you do a pass on this?

crates/bevy_remote/src/builtin_methods.rs
```diff
@@ -1491,6 +1733,30 @@ fn get_component_type_registration<'r>(
...
+    let resource_registration = get_resource_type_registration(type_registry, resource_path)?;
+
+    resource_registration
+        .data::<ReflectResource>()
+        .ok_or_else(|| anyhow!("Resource `{}` isn't reflectable", resource_path))
+}
+
+/// Given a resource's type path, return the associated [`TypeRegistration`] from the given
+/// `type_registry` if possible.
+fn get_resource_type_registration<'r>(
```
Leinnan, line 1751: Really minor: this could be imho named `get_type_registration` since it would work the same way for the non resource type. In case that is changed, change also the `resource_path` argument name.

Leinnan (approved):
Really good job. :D I've tested it locally, works great! I agree with @purplg that names could use cleanup and I would love that to happened before `0.16` release. But this is a separate change so I've created ticket for it: #18055

alice-i-cecile:
Thank you to everyone involved with the authoring or reviewing of this PR! This work is relatively important and needs release notes! Head over to https://github.com/bevyengine/bevy-website/issues/1997 if you'd like to help out.

PR #17424, opened 2025-01-18
Flattened `PointerAction::Pressed` into `Press` and `Release`.

### LINKED ISSUE #17397: Rename PointerAction::Pressed to PointerAction::Button
This variatn is also used for `PressDirection::Released`, which is a bit silly.

AlephCubed:
Fixes #17397.
Also renamed all variants into present-tense.
## Migration Guide
- `PointerAction::Pressed` has been seperated into two variants, `PointerAction::Press` and `PointerAction::Release`.
- `PointerAction::Moved` has been renamed to `PointerAction::Move`. 
- `PointerAction::Canceled` has been renamed to `PointerAction::Cancel`.

### DIFF
--- a/crates/bevy_picking/src/events.rs
+++ b/crates/bevy_picking/src/events.rs
@@ -51,9 +51,7 @@ use tracing::debug;
 use crate::{
     backend::{prelude::PointerLocation, HitData},
     hover::{HoverMap, PreviousHoverMap},
-    pointer::{
-        Location, PointerAction, PointerButton, PointerId, PointerInput, PointerMap, PressDirection,
-    },
+    pointer::{Location, PointerAction, PointerButton, PointerId, PointerInput, PointerMap},
 };
 
 /// Stores the common data needed for all pointer events.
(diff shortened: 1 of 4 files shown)


alice-i-cecile (approved):
I think this is much clearer, but I *did* make the issue.

Carter0 (commented):
I see, this logic isn't just for pressing things, but for releasing things as well. I like the idea, but I didn't get what the logic for the change was until I saw the issue. 

Why not use PressedOrReleased instead of Button? I think its clearer that way personally

aevyrie:
I don't understand the motivation for this. I would expect a verb like "move" or "press". The action is "a press". The direction can be "up" or "down". I think Alice's desire in the issue is caused more by the nesting. What if we flatten this a bit to simplify things?

```rs
/// Event sent to drive a pointer.
pub enum PointerAction {
    /// Causes the pointer to press a button.
    Press(PointerButton),
    /// Causes the pointer to release a button.
    Release(PointerButton),
    /// Move the pointer.
    Move {
        /// How much the pointer moved from the previous position.
        delta: Vec2,
    },
    /// The pointer has been canceled. The OS can cause this to happen to touch events.
    Canceled,
}
```

Note I also changed the tense and docs. It seems strange to use past-tense for an action the implementor is sending. These are "driver" events for pointer state.

For that reason I'm also thinking the name `PointerAction` is too close to the consumer API. These events are meant for implementors to drive pointers, which is why they are prefixed with `Input` downstream. I could see people seeing these in the docs thinking they should be reading them somewhere. I keep saying driver, so if maintainers didn't like `PointerInput` for some reason, what about `PointerDriver`?

AlephCubed:
I do like that flattened structure. I guess one argument for using the nested version is to match `MouseButtonInput`, but they already use different naming conventions and structures.

I also agree that `PointerDriver` is a more explicit name.

> I'm also noticing that in the upstream it looks like inputs have been mixed up? The position of a pointer is absolute, not relative, so I'm not sure why we even have a move input? I need to look at the source more closely.

You seem to be correct. The only times `PointerAction::Moved` is created, the absolute position it is converted to relative. Some downstream events like `bevy_picking::events::Move` do use that relative position, but that conversion could be done later.
I guess some cursors could benefit from using relative movement, i.e. a cursor controlled by a joystick, but that's an unusual use case.

alice-i-cecile:
I like the flattened enum too! I'd be very happy to see that.

alice-i-cecile:
Remember to update the PR description and title :)

alice-i-cecile (approved):
Lovely. I like the imperative names for all the actions too.

Carter0 (approved):
I mostly get what's going on and I think it looks good

NthTensor (approved):
Looks good to me. Btw for other reviewers, we've got to take extreme care when modifying logic in `pointer_events` because it is very easy to change the ordering patterns. Looks like that care has been taken here, I don't see any issues.

I think I chose these names initially to match the language used in `mod_picking` but I could be wrong about that.

PR #17427, opened 2025-01-19
Parent -> ChildOf

### LINKED ISSUE #17412: Rename `Parent` to `ChildOf` and `Children` to `ParentOf`
When I started learning how to use Bevy recently, this was the first stumbling block. It seemed straightforward that `Parent` was a marker component (or "driving concept") for a parent entity!

_Originally posted by @physgun in https://github.com/bevyengine/bevy/pull/17398#discussion_r1917639410_

cart:
Fixes #17412

## Objective

`Parent`  uses the "has a X" naming convention. There is increasing sentiment that we should use the "is a X" naming convention for  relationships (following #17398). This leaves `Children` as-is because there is prevailing sentiment that `Children` is clearer than `ParentOf` in many cases (especially when treating it like a collection).

This renames `Parent` to `ChildOf`.

This is just the implementation PR. To discuss the path forward, do so in #17412.

## Migration Guide

- The `Parent` component has been renamed to `ChildOf`.

### DIFF
--- a/crates/bevy_ecs/src/hierarchy.rs
+++ b/crates/bevy_ecs/src/hierarchy.rs
@@ -1,5 +1,5 @@
 //! The canonical "parent-child" [`Relationship`] for entities, driven by
-//! the [`ChildOf`] [`Relationship`] and the [`Children`] [`RelationshipTarget`].
+//! the [`ChildOf`] [`Relationship`] and the [`ParentOf`] [`RelationshipTarget`].
 //!
 //! See [`ChildOf`] for a full description of the relationship and how to use it.
 //!
@@ -29,7 +29,7 @@ use log::warn;
 
 /// A [`Relationship`](crate::relationship::Relationship) component that creates the canonical
 /// "parent / child" hierarchy. This is the "source of truth" component, and it pairs with
-/// the [`Children`] [`RelationshipTarget`](crate::relationship::RelationshipTarget).
+/// the [`ParentOf`] [`RelationshipTarget`](crate::relationship::RelationshipTarget).
 ///
 /// This relationship should be used for things like:
 ///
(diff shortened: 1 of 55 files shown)


crates/bevy_ecs/src/hierarchy.rs
```diff
@@ -123,32 +123,32 @@ impl Deref for Parent {
...
 #[derive(Component, Default, VisitEntitiesMut, Debug, PartialEq, Eq)]
-#[relationship_target(relationship = Parent, despawn_descendants)]
+#[relationship_target(relationship = ChildOf, despawn_descendants)]
 #[cfg_attr(feature = "bevy_reflect", derive(bevy_reflect::Reflect))]
 #[cfg_attr(
     feature = "bevy_reflect",
     reflect(Component, MapEntities, VisitEntities, VisitEntitiesMut, FromWorld)
 )]
-pub struct Children(Vec<Entity>);
+pub struct ParentOf(Vec<Entity>);
```
MrGVSV, line 149: Personally I prefer `Children`. Much like I think `TeacherOf` should probably be `Students` and `EmployerOf` should be `Employees`. It doesn't fit every case, but where one "owns" another and there's a good plural form of it, I feel like that kind of naming works.

The whole "is a" vs "has a" is confusing anyways since you'd say "this entity has `Health`" not "this entity is a `Health`" (but not so for things like `Player`).

MrGVSV, line 149: Alternatively I like `XTo` as well. And it may be possible to use that sort of naming convention to convey certain meaning such as ownership or directionality (or sources vs targets).

physgun: It comes down to how much everyone wants to stick with a central, un-lintable naming scheme. It is possible to adhere to IsA, for example a `Health` component would be something like `Mortal` and/or `Destructible` instead.

cart:
I removed the `Children -> ParentOf` rename, following the discussion in #17412. I think this is good to merge now.

alice-i-cecile (approved):

Carter0 (approved):
It seems our bikes have been successfully shed.

viridia (approved):

PR #17442, opened 2025-01-19
Switch `bevy_asset` to `core::prelude`

bushrat011899:
Makes use of `std` explicit, simplifying a possible `no_std` port.

# Objective

- Contributes to #15460
- Simplify future `no_std` work on `bevy_asset`

## Solution

- Add `#![no_std]` to switch to `core::prelude` instead of `std::prelude`

## Testing

- CI

---

## Notes

This is entirely a change around the names of imports and has no impact on functionality. This just reduces the quantity of changes involved in the (likely more controversial) `no_std`-ification of `bevy_asset`.

### DIFF
--- a/crates/bevy_asset/src/asset_changed.rs
+++ b/crates/bevy_asset/src/asset_changed.rs
@@ -288,7 +288,9 @@ unsafe impl<A: AsAssetId> QueryFilter for AssetChanged<A> {
 #[cfg(test)]
 mod tests {
     use crate::{self as bevy_asset, AssetEvents, AssetPlugin, Handle};
+    use alloc::{vec, vec::Vec};
     use core::num::NonZero;
+    use std::println;
 
     use crate::{AssetApp, Assets};
     use bevy_app::{App, AppExit, Last, Startup, TaskPoolPlugin, Update};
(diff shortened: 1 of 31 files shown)


crates/bevy_asset/src/asset_changed.rs
```diff
@@ -288,7 +288,9 @@ unsafe impl<A: AsAssetId> QueryFilter for AssetChanged<A> {
 #[cfg(test)]
 mod tests {
     use crate::{self as bevy_asset, AssetEvents, AssetPlugin, Handle};
+    use alloc::{vec, vec::Vec};
     use core::num::NonZero;
+    use std::println;
```
alice-i-cecile, line 293: Boo!

alice-i-cecile (approved):

crates/bevy_asset/src/asset_changed.rs
bushrat011899: Exactly, I wonder if Bevy should actually prohibit `eprintln` and `println` since we always have access to `log` and/or `tracing`...

bushrat011899:
Ok that _should_ be the last fix.

PR #17443, opened 2025-01-19
Add an example teaching users about custom relationships

alice-i-cecile:
# Objective

After #17398, Bevy now has relations! We don't teach users how to make / work with these in the examples yet though, but we definitely should.

## Solution

- Add a simple abstract example that goes over defining, spawning, traversing and removing a custom relations.
- ~~Add `Relationship` and `RelationshipTarget` to the prelude: the trait methods are really helpful here.~~
  - this causes subtle ambiguities with method names and weird compiler errors. Not doing it here!
- Clean up related documentation that I referenced when writing this example.

## Testing

`cargo run --example relationships`

## Notes to reviewers

1. Yes, I know that the cycle detection code could be more efficient. I decided to reduce the caching to avoid distracting from the broader point of "here's how you traverse relationships".
2. Instead of using an `App`, I've decide to use `World::run_system_once` + system functions defined inside of `main` to do something closer to literate programming.


examples/ecs/hierarchy.rs
```diff
@@ -1,4 +1,8 @@
-//! Creates a hierarchy of parents and children entities.
+//! Demonstrates techniques for creating a hierarchy of parents and children entities.
```
Jondolf, line 1:
```suggestion
//! Demonstrates techniques for creating a hierarchy of parent and child entities.
```

examples/ecs/relationships.rs
```diff
@@ -0,0 +1,192 @@
+//! Entities generally don't exist in isolation. Instead, they are related to other entities in various ways.
+//! While Bevy comes with a built-in [`Parent`]/[`Children`] relationship
+//! (which enables transform and visibility propagation),
+//! you can define your own relationships using components.
+//!
+//! Every relation has two sides: the source and the target.
```
Jondolf, line 6: Teeny-weeny nit: this is the only instance of "relation" in this PR, it's "relationship" everywhere else
```suggestion
//! Every relationship has two sides: the source and the target.
```

examples/ecs/relationships.rs
```diff
@@ -0,0 +1,192 @@
+//! Entities generally don't exist in isolation. Instead, they are related to other entities in various ways.
+//! While Bevy comes with a built-in [`Parent`]/[`Children`] relationship
+//! (which enables transform and visibility propagation),
+//! you can define your own relationships using components.
+//!
+//! Every relationship has two sides: the source and the target.
+//! The source is the entity that has the relationship with the target,
+//! while the target keeps track of all the entities that have a relationship with it.
+//! For the standard hierarchy, the source is stored in the [`Parent`] component,
+//! while the target is stored in the [`Children`] component.
```
Jondolf, line 10: This is maybe a bit confusing to read and I *think* the end is wrong.

If I understand correctly, `ChildOf` (the current `Parent`) is the *relationship*, the entity that stores it is the *source*, and the entity stored in the relationship component is the *target*. The target entity has a `Children` component that keeps track of all the source entities that have the `ChildOf` relationship with it.

This is currently saying that the source is stored in the `Parent` component and the target (should be plural for `Children`) is stored in the `Children` component, but it should be the other way around. The relationship component (`ChildOf`/`Parent`) is on the source entity and stores the entity it is targeting.

The current naming also definitely doesn't help. It's easy to think `Parent` is on the parent entity of the relationship, when in reality it is on the child entity and stores the parent as the target of a "has a parent" relationship. `ChildOf` would make this much clearer as an "is a child of" relationship.

examples/ecs/relationships.rs
```diff
@@ -0,0 +1,192 @@
...
+    }
+
+    world.run_system_once(debug_relationships).unwrap();
+
+    // Systems can return errors,
+    // which can be used to signal that something went wrong during the system's execution.
+    #[derive(Debug)]
+    #[expect(
+        dead_code,
+        reason = "This error struct is only used for debugging in this example."
```
MinerSebas, line 144: While this reason would be fine in the internals, I don't think this good when used inside an example that could be read by beginners.

Maybe something like this to explain why the expect is necessary here.
```suggestion
        reason = "Rust considers types that are only used by their debug trait as dead code."
```

examples/ecs/relationships.rs
alice-i-cecile: I've attempted to rewrite this, and it should be much clearer now.

examples/ecs/relationships.rs
```diff
@@ -0,0 +1,215 @@
...
+                targeting_name = target_name.clone();
+
+                if visited.contains(&targeting) {
+                    return Err(TargetingCycle {
+                        initial_entity,
+                        visited,
+                    });
+                } else {
+                    visited.insert(targeting);
+                }
```
kristoff3r, line 178: `HashSet::insert` returns false if the value is already present, so this can be written as

```suggestion
                if !visited.insert(targeting) {
                    return Err(TargetingCycle {
                        initial_entity,
                        visited,
                    });
                }
```

It's shorter but might not be as clear when you don't know the pattern.

kristoff3r (approved):

examples/ecs/relationships.rs
```diff
@@ -0,0 +1,215 @@
...
+struct TargetedBy(Vec<Entity>);
+
+fn main() {
+    // Operating on a raw `World` and running systems one at a time
+    // is great for writing tests and teaching abstract concepts!
+    let mut world = World::new();
+
+    // We're going to spawn a few entities and relate them to each other in a complex way.
+    // To start, Alice will target Bob, Bob will target Charlie,
+    // and Charlie will target Alice. This creates a loop in the relationship graph.
```
nekosaur, line 43: Am I being dumb and thinking about this the wrong way, or is this description not the opposite of what is happening in the initial setup of the example?

Bob targets Alice, Charlie targets Bob, Devon targets Charlie, and then Alice also targets Charlie?

Carter0: Yeah the comment above is wrong. Alice is never targeting Bob.

examples/ecs/relationships.rs
```diff
@@ -0,0 +1,213 @@
...
+    //
+    // Then, we'll spawn Devon, who will target Charlie,
+    // creating a more complex graph with a branching structure.
+    fn spawning_entities_with_relationships(mut commands: Commands) {
+        // Calling .id() after spawning an entity will return the `Entity` identifier of the spawned entity,
+        // even though the entity itself is not yet instantiated in the world.
+        // This works because Commands will reserve the entity ID before actually spawning the entity,
+        // through the use of atomic counters.
+        let alice = commands.spawn(Name::new("Alice")).id();
```
Carter0, line 52: Crazy question, but where is 'Name' coming from?

Carter0 (approved):
Minor comments, but overall I think It looks good!

examples/ecs/relationships.rs
alice-i-cecile: This is a `bevy_core` component! It's quite nice!

examples/ecs/relationships.rs
alice-i-cecile: Thanks for catching this <3

examples/ecs/relationships.rs
```diff
@@ -0,0 +1,213 @@
+//! Entities generally don't exist in isolation. Instead, they are related to other entities in various ways.
+//! While Bevy comes with a built-in [`Parent`]/[`Children`] relationship
```
alice-i-cecile, line 2:
```suggestion
//! While Bevy comes with a built-in [`ChildOf`]/[`Children`] relationship
```

examples/ecs/relationships.rs
```diff
@@ -0,0 +1,213 @@
...
+//! (which enables transform and visibility propagation),
+//! you can define your own relationships using components.
+//!
+//! We can define a custom relationship by creating two components:
+//! one to store the relationship itself, and another to keep track of the reverse relationship.
+//! Bevy's [`Parent`] component implements the [`Relationship`] trait, serving as the source of truth,
```
alice-i-cecile, line 8:
```suggestion
//! Bevy's [`ChildOf`] component implements the [`Relationship`] trait, serving as the source of truth,
```

alice-i-cecile:
I have *no* idea what's going on with CI here: I can't reproduce the failures locally or on `main`.

PR #17447, opened 2025-01-20
Support non-Vec data structures in relations

alice-i-cecile:
# Objective

The existing `RelationshipSourceCollection` uses `Vec` as the only possible backing for our relationships. While a reasonable choice, benchmarking use cases might reveal that a different data type is better or faster.

For example:

- Not all relationships require a stable ordering between the relationship sources (i.e. children). In cases where we a) have many such relations and b) don't care about the ordering between them, a hash set is likely a better datastructure than a `Vec`.
- The number of children-like entities may be small on average, and a `smallvec` may be faster

## Solution

- Implement `RelationshipSourceCollection` for `EntityHashSet`, our custom entity-optimized `HashSet`.
-~~Implement `DoubleEndedIterator` for `EntityHashSet` to make things compile.~~
   -  This implementation was cursed and very surprising.
   - Instead, by moving the iterator type on `RelationshipSourceCollection` from an erased RPTIT to an explicit associated type we can add a trait bound on the offending methods!
- Implement `RelationshipSourceCollection` for `SmallVec`

## Testing

I've added a pair of new tests to make sure this pattern compiles successfully in practice!

## Migration Guide

`EntityHashSet` and `EntityHashMap` are no longer re-exported in `bevy_ecs::entity` directly. If you were not using `bevy_ecs` / `bevy`'s `prelude`, you can access them through their now-public modules, `hash_set` and `hash_map` instead.

## Notes to reviewers

The `EntityHashSet::Iter` type needs to be public for this impl to be allowed. I initially renamed it to something that wasn't ambiguous and re-exported it, but as @Victoronz pointed out, that was somewhat unidiomatic.

In https://github.com/bevyengine/bevy/pull/17447/commits/1a8564898f1803cae4e99bbf909a40ab824fb8a0, I instead made the `entity_hash_set` public (and its `entity_hash_set`) sister public, and removed the re-export. I prefer this design (give me module docs please), but it leads to a lot of churn in this PR.

Let me know which you'd prefer, and if you'd like me to split that change out into its own micro PR.

### DIFF
--- a/crates/bevy_ecs/src/entity/hash_set.rs
+++ b/crates/bevy_ecs/src/entity/hash_set.rs
@@ -68,8 +68,8 @@ impl EntityHashSet {
     /// The iterator element type is `&'a Entity`.
     ///
     /// Equivalent to [`HashSet::iter`].
-    pub fn iter(&self) -> EntityHashSetIter<'_> {
-        EntityHashSetIter(self.0.iter(), PhantomData)
+    pub fn iter(&self) -> Iter<'_> {
+        Iter(self.0.iter(), PhantomData)
     }
 
     /// Drains elements which are true under the given predicate,
@@ -98,10 +98,10 @@ impl DerefMut for EntityHashSet {
 impl<'a> IntoIterator for &'a EntityHashSet {
     type Item = &'a Entity;
 
-    type IntoIter = EntityHashSetIter<'a>;
+    type IntoIter = Iter<'a>;
 
     fn into_iter(self) -> Self::IntoIter {
-        EntityHashSetIter((&self.0).into_iter(), PhantomData)
+        Iter((&self.0).into_iter(), PhantomData)
     }
 }
 
@@ -200,16 +200,16 @@ impl FromIterator<Entity> for EntityHashSet {
 /// This struct is created by the [`iter`] method on [`EntityHashSet`]. See its documentation for more.
 ///
 /// [`iter`]: EntityHashSet::iter
-pub struct EntityHashSetIter<'a, S = EntityHash>(hash_set::Iter<'a, Entity>, PhantomData<S>);
+pub struct Iter<'a, S = EntityHash>(hash_set::Iter<'a, Entity>, PhantomData<S>);
 
-impl<'a> EntityHashSetIter<'a> {
+impl<'a> Iter<'a> {
     /// Returns the inner [`Iter`](hash_set::Iter).
     pub fn into_inner(self) -> hash_set::Iter<'a, Entity> {
         self.0
     }
 }
 
-impl<'a> Deref for EntityHashSetIter<'a> {
+impl<'a> Deref for Iter<'a> {
     type Target = hash_set::Iter<'a, Entity>;
 
     fn deref(&self) -> &Self::Target {
@@ -217,13 +217,13 @@ impl<'a> Deref for EntityHashSetIter<'a> {
     }
 }
 
-impl DerefMut for EntityHashSetIter<'_> {
+impl DerefMut for Iter<'_> {
     fn deref_mut(&mut self) -> &mut Self::Target {
         &mut self.0
     }
 }
 
-impl<'a> Iterator for EntityHashSetIter<'a> {
+impl<'a> Iterator for Iter<'a> {
     type Item = &'a Entity;
 
     fn next(&mut self) -> Option<Self::Item> {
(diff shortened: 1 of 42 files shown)


crates/bevy_ecs/src/relationship/relationship_source_collection.rs
```diff
@@ -1,9 +1,21 @@
...
 /// This is not intended to be modified directly by users, as it could invalidate the correctness of relationships.
 pub trait RelationshipSourceCollection {
+    /// The type of iterator returned by the `iter` method.
+    ///
+    /// This is an associated type (rather than using a method that returns an opaque return-position impl trait)
+    /// to ensure that all methods and traits (like [`DoubleEndedIterator`]) of the underlying collection's iterator
+    /// are available to the user when implemented without unduly restricting the possible collections.
+    ///
+    /// The [`SourceIter`](super::SourceIter) type alias can be helpful to reduce confusion when working with this associated type.
+    type SourceIter<'a>: Iterator<Item = Entity>
```
bushrat011899, line 15: This is definitely the right path forward. RPIT is great 90% of the time, but this is that 10% case.

bushrat011899 (approved):
Nice!

crates/bevy_ecs/src/entity/hash_set.rs
```diff
@@ -38,6 +42,16 @@ impl EntityHashSet {
...
+    /// Returns the number of elements in the set.
+    pub fn len(&self) -> usize {
+        self.0.len()
+    }
+
+    /// Returns `true` if the set contains no elements.
+    pub fn is_empty(&self) -> bool {
+        self.0.is_empty()
+    }
+
```
Victoronz, line 54: Are these not available via the `Deref` to the inner hashbrown set?

viridia (approved):

crates/bevy_ecs/src/entity/hash_set.rs
```diff
@@ -186,61 +200,61 @@ impl FromIterator<Entity> for EntityHashSet {
 /// This struct is created by the [`iter`] method on [`EntityHashSet`]. See its documentation for more.
 ///
 /// [`iter`]: EntityHashSet::iter
-pub struct Iter<'a, S = EntityHash>(hash_set::Iter<'a, Entity>, PhantomData<S>);
+pub struct EntityHashSetIter<'a, S = EntityHash>(hash_set::Iter<'a, Entity>, PhantomData<S>);
```
Victoronz, line 203: Currently the rename is still in here, given the "Note for Reviewers" section I'm not sure, was that intentional?

crates/bevy_ecs/src/entity/hash_set.rs
alice-i-cecile: Just relying on the `Deref` impl was failing for these, as the compiler was getting confused about the circular trait methods due to the same name / signature.

crates/bevy_ecs/src/entity/hash_set.rs
alice-i-cecile: Whoops!

crates/bevy_ecs/src/relationship/relationship_source_collection.rs
alice-i-cecile: It's a mouthful, but I really think that the added flexibility / correctness is worth it here.

crates/bevy_ecs/src/entity/hash_set.rs
Victoronz: ah, that is because of method resolution!
when `len` is an inherent method, then `self.len()` in a `len()` trait method impl works, but if the inherent `len()` is down a step in deref chain, the trait method resolves first, resulting in a cycle.
The solution should be any of `self.0.len()`/`(*self).len()`/`self.deref().len()` in the trait method impl.

crates/bevy_ecs/src/relationship/relationship_source_collection.rs
```diff
@@ -1,9 +1,21 @@
-use crate::entity::Entity;
+use crate::entity::{hash_set::EntityHashSet, Entity};
 use alloc::vec::Vec;
+use smallvec::SmallVec;
 
 /// The internal [`Entity`] collection used by a [`RelationshipTarget`](crate::relationship::RelationshipTarget) component.
 /// This is not intended to be modified directly by users, as it could invalidate the correctness of relationships.
 pub trait RelationshipSourceCollection {
```
mintlu8, line 7: Why not use something like `for<'t> &T: IntoIterator<&'t Entity> + Extend<Entity> + Default`?

mintlu8 (changes requested):

PR #17449, opened 2025-01-20
implement EntityIndexMap/Set

Victoronz:
# Objective

We do not have `EntityIndexMap`/`EntityIndexSet`.

Usual `HashMap`s/`HashSet`s do not guarantee any order, which can be awkward for some use cases.
The `indexmap` versions remember insertion order, which then also becomes their iteration order.
They can be thought of as a `HashTable` + `Vec`, which means fast iteration and removal, indexing by index (not just key),  and slicing!
Performance should otherwise be comparable.

## Solution

Because `indexmap` is structured to mirror `hashbrown`, it suffers the same issue of not having the `Hasher` generic on their iterators. #16912 solved this issue for `EntityHashMap`/`EntityHashSet` with a wrapper around the hashbrown version, so this PR does the same.

Hopefully these wrappers can be removed again in the future by having `hashbrown`/`indexmap` adopt that generic in their iterators themselves!

### DIFF
--- a/crates/bevy_ecs/src/entity/index_map.rs
+++ b/crates/bevy_ecs/src/entity/index_map.rs
@@ -1,0 +1,541 @@
+use core::{
+    fmt::{self, Debug, Formatter},
+    hash::BuildHasher,
+    iter::FusedIterator,
+    marker::PhantomData,
+    ops::{Deref, DerefMut, Index, IndexMut, RangeBounds},
+};
+
+#[cfg(feature = "bevy_reflect")]
+use bevy_reflect::Reflect;
+use indexmap::map::{self, IndexMap};
+
+use super::{Entity, EntityHash, EntitySetIterator, TrustedEntityBorrow};
+
+/// A [`IndexMap`] pre-configured to use [`EntityHash`] hashing.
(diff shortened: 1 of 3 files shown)


alice-i-cecile (approved):

alice-i-cecile:
As discussed on Discord, we should adopt https://github.com/bevyengine/bevy/pull/17447 to apply to the `EntityIndexMap` type once this is merged.

Victoronz:
Note: this currently only includes slicing via deref. It'd put this PR well over a thousand lines to add the wrappers for those as well, so I'll do that as a follow-up.
This means some types here seem useless to wrap: In actuality, they can be sliced, which in turn can then return `EntitySetIterator`s.

BenjaminBrienen (approved):

alice-i-cecile:
Actually, before I merge this, can you say a bit more on the motivation here? Adding code like this isn't free, and we don't have any internal users currently.

Victoronz:
> Actually, before I merge this, can you say a bit more on the motivation here? Adding code like this isn't free, and we don't have any internal users currently.

Fair, I've added some more extensive motivation.

PR #17450, opened 2025-01-20
remove unsound DerefMut impls from `EntityHashMap`/`EntityHashSet`

Victoronz:
# Objective

Noticed while doing #17449, I had left these `DerefMut` impls in.
Obtaining mutable references to those inner iterator types allows for `mem::swap`, which can be used to swap an incorrectly behaving instance into the wrappers.

## Solution

Remove them!

### DIFF
--- a/crates/bevy_ecs/src/entity/hash_map.rs
+++ b/crates/bevy_ecs/src/entity/hash_map.rs
@@ -165,12 +165,6 @@ impl<'a, V> Deref for Keys<'a, V> {
     }
 }
 
-impl<V> DerefMut for Keys<'_, V> {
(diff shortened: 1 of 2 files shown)


alice-i-cecile (approved):

BenjaminBrienen (approved):
Should we have more code coverage using `mem::swap` so that miri can catch this?

Victoronz:
> Should we have more code coverage using `mem::swap` so that miri can catch this?

I don't think this is tenable, since what we'd need to test is an absence of an API, not how the present one behaves. It is not that `DerefMut` impls for wrappers that maintain safety invariants are wrong, sometimes such an impl is used to return another wrapper.

PR #17453, opened 2025-01-20
Include ReflectFromReflect in all dynamic data types.

### LINKED ISSUE #17416: HashMaps don't include ReflectFromReflect
## Bevy version

0.15.1

## What you did

I'm trying to instantiate concrete bevy types using the reflection features.

## What went wrong

There are some types in the type registry that use hashmaps, for example:

```
bevy_utils::hashbrown::HashMap<petgraph::graph::NodeIndex, bevy_animation::ActiveAnimation, bevy_utils::hashbrown::hash_map::DefaultHashBuilder>
bevy_utils::hashbrown::HashMap<bevy_ecs::entity::Entity, alloc::vec::Vec<bevy_pbr::light::Cascade>, bevy_ecs::entity::hash::EntityHash>
bevy_utils::hashbrown::HashMap<bevy_animation::AnimationEventTarget, alloc::vec::Vec<bevy_animation::TimedAnimationEvent>, bevy_utils::hashbrown::hash_map::DefaultHashBuilder>
bevy_utils::hashbrown::HashMap<bevy_animation::AnimationTargetId, u64, bevy_utils::hashbrown::hash_map::DefaultHashBuilder>
bevy_utils::hashbrown::HashMap<bevy_asset::id::AssetId<bevy_animation::graph::AnimationGraph>, bevy_animation::graph::ThreadedAnimationGraph, bevy_utils::hashbrown::hash_map::DefaultHashBuilder>
```

All of them do not have the `ReflectFromReflect` type data object attached. This means that I cannot convert from a `DynamicMap` to a concrete reflected instance of that type.

## Additional information

I was able to work around this for a specific type by doing this after the general registrations:

```rust
let graphmap = registry
    .get_mut(std::any::TypeId::of::<
        bevy::utils::HashMap<petgraph::graph::NodeIndex, bevy::animation::ActiveAnimation>,
    >())
    .unwrap();
graphmap.insert::<ReflectFromReflect>(bevy::reflect::FromType::<
    bevy::utils::HashMap<petgraph::graph::NodeIndex, bevy::animation::ActiveAnimation>,
>::from_type());
```

Unfortunately, there's no way to loop over these due to the strict types. I think the macro that generates the registration just doesn't handle `ReflectFromReflect` and has to be updated:

https://github.com/bevyengine/bevy/blob/b66c3ceb0ee39374ff1759ffb1b5bee2e4b93e99/crates/bevy_reflect/src/impls/std.rs#L808-L812

I think this is true for other macros in that file as well, for example `impl_reflect_for_hashset`.

anlumo:
# Objective

Fixes #17416

## Solution

I just included ReflectFromReflect in all macros and implementations. I think this should be ok, at least it compiles properly and does fix the errors in my test code.

## Testing

I generated a DynamicMap and tried to convert it into a concrete `HashMap` as a `Box<dyn Reflect>`. Without my fix, it doesn't work, because this line panics:

```rust
let rfr = ty.data::<ReflectFromReflect>().unwrap();
```

where `ty` is the `TypeRegistration` for the (matching) `HashMap`.

I don't know why `ReflectFromReflect` wasn't included everywhere, I assume that it was an oversight and not an architecture decision I'm not aware of.

# Migration Guide

The hasher in reflected `HashMap`s and `HashSet`s now have to implement `Default`. This is the case for the ones provided by Bevy already, and is generally a sensible thing to do.

### DIFF
--- a/crates/bevy_reflect/src/impls/std.rs
+++ b/crates/bevy_reflect/src/impls/std.rs
@@ -559,6 +559,7 @@ macro_rules! impl_reflect_for_veclike {
             fn get_type_registration() -> TypeRegistration {
                 let mut registration = TypeRegistration::of::<$ty>();
                 registration.insert::<ReflectFromPtr>(FromType::<$ty>::from_type());
+                registration.insert::<ReflectFromReflect>(FromType::<$ty>::from_type());
                 registration
             }
 
@@ -803,11 +804,12 @@ macro_rules! impl_reflect_for_hashmap {
         where
             K: FromReflect + MaybeTyped + TypePath + GetTypeRegistration + Eq + Hash,
             V: FromReflect + MaybeTyped + TypePath + GetTypeRegistration,
-            S: TypePath + BuildHasher + Send + Sync,
+            S: TypePath + BuildHasher + Send + Sync + Default,
         {
             fn get_type_registration() -> TypeRegistration {
                 let mut registration = TypeRegistration::of::<Self>();
                 registration.insert::<ReflectFromPtr>(FromType::<Self>::from_type());
+                registration.insert::<ReflectFromReflect>(FromType::<Self>::from_type());
                 registration
             }
 
@@ -1028,11 +1030,12 @@ macro_rules! impl_reflect_for_hashset {
         impl<V, S> GetTypeRegistration for $ty
         where
             V: FromReflect + TypePath + GetTypeRegistration + Eq + Hash,
-            S: TypePath + BuildHasher + Send + Sync,
+            S: TypePath + BuildHasher + Send + Sync + Default,
         {
             fn get_type_registration() -> TypeRegistration {
                 let mut registration = TypeRegistration::of::<Self>();
                 registration.insert::<ReflectFromPtr>(FromType::<Self>::from_type());
+                registration.insert::<ReflectFromReflect>(FromType::<Self>::from_type());
                 registration
             }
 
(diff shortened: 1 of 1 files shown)


MrGVSV (approved):
Looks good to me! I’m not against requiring `Default` on the `BuildHasher` argument.

MrGVSV:
One thing you may want to change is the title of this PR. It might be confusing since reflection's dynamic types are explicitly not `FromReflect` (currently). This change mainly affects _concrete_ map, list, and set types.

crates/bevy_reflect/src/impls/std.rs
```diff
@@ -1028,11 +1030,12 @@ macro_rules! impl_reflect_for_hashset {
         impl<V, S> GetTypeRegistration for $ty
         where
             V: FromReflect + TypePath + GetTypeRegistration + Eq + Hash,
-            S: TypePath + BuildHasher + Send + Sync,
+            S: TypePath + BuildHasher + Send + Sync + Default,
```
BenjaminBrienen, line 1033: Is this a breaking change?

MrGVSV: Yes I just forgot to add the label, good call!

MrGVSV: Wait is there no "Breaking Change" label? Or am I just not seeing it?

alice-i-cecile (approved):

PR #17461, opened 2025-01-21
EaseFunction svg graphs in doc

SpecificProtagonist:
# Objective

The docs of `EaseFunction` don't visualize the different functions, requiring you to check out the Bevy repo and running the `easing_function` example.

## Solution

- Add tool to generate suitable svg graphs. This only needs to be re-run when adding new ease functions.
- works with all themes
- also add missing easing functions to example.

---

## Showcase

Graphs


BenjaminBrienen (approved):

bushrat011899 (approved):
I think this is a good idea. The documentation becomes massively more approachable when it's more than just type definitions; examples, tables, figures, all make it much nicer.

I do have a concern around needing to maintain this tool separately to `bevy_math`. Perhaps it would be worth trying to refactor this into a proc-macro? Especially considering we're using the `doc` attribute to just include the SVG string in the documentation anyway.

bushrat011899:
like this idea, my only concern would be whether this would affect build time.

Could use `cfg_attr(doc, ...)` to make the macro only run when building documentation?

bushrat011899:
> Will do! I'm assuming the `doc` would be a feature enabled on docs.rs? Or is there a way to find out if it's getting build by rustdoc?

https://doc.rust-lang.org/rustdoc/advanced-features.html

`doc` is a built-in config set by `rust-doc` when it builds documentation. So if you do:

```rust
#[cfg_attr(doc, my_special_macro(...))]
```

It'll only include `my_specual_macro` when building documentation.

bushrat011899:
After some experimentation, I don't think it'll be possible to use a proc-macro to generate the graphs. It'd require a cyclic dependency between the macro crate and `bevy_math`, which is forbidden outside of `dev-dependencies`.

tools/build-easefunction-graphs/src/main.rs
```diff
@@ -0,0 +1,125 @@
...
+    Document,
+};
+
+fn main() {
+    let directory = "../../crates/bevy_math/images/easefunction";
+    if !std::fs::exists(directory).unwrap() {
+        eprintln!("{directory} does not exist");
+        eprintln!("Make sure to run this from the tools/build-easefunction-graphs folder");
+        std::process::exit(1);
+    }
```
mockersf, line 14: could you make this relative to `CARGO_MANIFEST_DIR` so that it doesn't matter where you run it?

examples/animation/easing_functions.rs
```diff
@@ -37,6 +37,12 @@ fn setup(mut commands: Commands) {
         EaseFunction::QuinticIn,
         EaseFunction::QuinticOut,
         EaseFunction::QuinticInOut,
+        EaseFunction::SmoothStepIn,
+        EaseFunction::SmoothStepOut,
+        EaseFunction::SmoothStep,
+        EaseFunction::SmootherStepIn,
+        EaseFunction::SmootherStepOut,
+        EaseFunction::SmootherStep,
```
mockersf, line 45: adding those means that some of the functions are displayed outside of the window default size

mockersf (approved):

PR #17463, opened 2025-01-21
Add insert and remove recursive methods on `EntityWorldMut` and `EntityCommands`

alice-i-cecile:
# Objective

While being able to quickly add / remove components down a tree is broadly useful (material changing!), it's particularly necessary when combined with the newly added #13120.

## Solution

Write four methods: covering both adding and removal on both `EntityWorldMut` and `EntityCommands`.

These methods are generic over the `RelationshipTarget`, thanks to the freshly merged relations 🎉 

## Testing

I've added a simple unit test for these methods.

### DIFF
--- a/crates/bevy_ecs/src/relationship/related_methods.rs
+++ b/crates/bevy_ecs/src/relationship/related_methods.rs
@@ -5,6 +5,7 @@ use crate::{
     system::{Commands, EntityCommands},
     world::{EntityWorldMut, World},
 };
+use alloc::vec::Vec;
 use core::marker::PhantomData;
 
 impl<'w> EntityWorldMut<'w> {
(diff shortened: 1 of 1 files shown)


crates/bevy_ecs/src/relationship/related_methods.rs
```diff
@@ -45,6 +46,55 @@ impl<'w> EntityWorldMut<'w> {
...
         self
     }
+
+    /// Inserts a component or bundle of components into the entity and all related entities,
+    /// traversing the relationship tracked in `S` in a breadth-first manner.
+    ///
+    /// # Warning
+    ///
+    /// This method should only be called on relationships that form a tree-like structure.
+    /// Any cycles will cause this method to loop infinitely.
```
bushrat011899, line 56: Follow-up: I reckon a `detect_cyclic_traversal` debug feature could be nice. Internally just keep an `EntityHashSet` and `panic!` when a cycle is detected.

crates/bevy_ecs/src/relationship/related_methods.rs
```diff
@@ -79,6 +129,38 @@ impl<'a> EntityCommands<'a> {
...
+        self.commands.queue(move |world: &mut World| {
+            world.entity_mut(id).insert_recursive::<S>(bundle);
+        });
+        self
+    }
+
+    /// Removes a component or bundle of components of type `B` from the entity and all related entities,
+    /// traversing the relationship tracked in `S` in a breadth-first manner.
+    ///
+    /// # Warning
```
bushrat011899, line 154:
```suggestion
    /// # Warning
    ///
```

bushrat011899 (approved):
Makes sense! just a missing new-line in the documentation.

crates/bevy_ecs/src/relationship/related_methods.rs
alice-i-cecile: Agreed. There's a ton of methods that could use this.

AlephCubed (approved):

PR #17331, opened 2025-01-12
Improve `bevy_ecs` inspectability and add debugging methods.

AlephCubed:
My attempt at adopting #7758.
## Solution
- Add `len`, `is_empty`, and `iter` methods to `Bundle`.
- Add diagnostic methods to `World`, `Schedules`, `Dag`, and `ComponentId`.
    - `Schedules` and `World` also include a `diagnose_with_flattened` method that includes a flattened version of the dependancy graph. This was moved to a separate method as getting the flattened graph requires mutable access.
## Example output:
This is an example output for `World::diagnose_with_flattened`:
```
world:
  summary:
    component: 4
    bundle: 9
    archetype: 5
    entity: 1
    resource: 2
    resource(non send): 0
    table: 4
    sparse set: 1
  detail:
    components:
      ComponentId(0)(bevy_ecs::event::base::EventWrapperComponent<bevy_ecs::world::component_constants::OnAdd>) Table send_sync
      ComponentId(1)(bevy_ecs::event::base::EventWrapperComponent<bevy_ecs::world::component_constants::OnInsert>) Table send_sync
      ComponentId(2)(bevy_ecs::event::base::EventWrapperComponent<bevy_ecs::world::component_constants::OnReplace>) Table send_sync
      ComponentId(3)(bevy_ecs::event::base::EventWrapperComponent<bevy_ecs::world::component_constants::OnRemove>) Table send_sync
      ComponentId(4)(bevy_ecs::schedule::schedule::Schedules) Table send_sync
      ComponentId(5)(world_diagnostic::GameState) Table send_sync
      ComponentId(6)(world_diagnostic::Counter) Table send_sync
      ComponentId(7)(world_diagnostic::Player) Table send_sync
      ComponentId(8)(world_diagnostic::HighlightFlag) SparseSet send_sync
    bundles:
      BundleId(0): ["ComponentId(7)(world_diagnostic::Player)"]
      BundleId(1): ["ComponentId(6)(world_diagnostic::Counter)", "ComponentId(8)(world_diagnostic::HighlightFlag)"]
      BundleId(2): ["ComponentId(6)(world_diagnostic::Counter)"]
      BundleId(3): ["ComponentId(8)(world_diagnostic::HighlightFlag)"]
    archetypes:
      ArchetypeId(0): components:[] table:TableId(0) entity:0
      ArchetypeId(1): components:["ComponentId(7)(world_diagnostic::Player)"] table:TableId(1) entity:0
[...]

PR #17464, opened 2025-01-21
Added helper methods to `Bundles`.

AlephCubed:
Added `len`, `is_empty`, and `iter` methods to `Bundles`.

Separated out from #17331.


alice-i-cecile (approved):

alice-i-cecile:
CI says this needs formatting :)

PR #17469, opened 2025-01-21
Move `Resource` trait to its own file

alice-i-cecile:
# Objective

`bevy_ecs`'s `system` module is something of a grab bag, and *very* large. This is particularly true for the `system_param` module, which is more than 2k lines long!

While it could be defensible to put `Res` and `ResMut` there (lol no they're in change_detection.rs, obviously), it doesn't make any sense to put the `Resource` trait there. This is confusing to navigate (and painful to work on and review).

## Solution

- Create a root level `bevy_ecs/resource.rs` module to mirror `bevy_ecs/component.rs`
- move the `Resource` trait to that module
- move the `Resource` derive macro to that module as well (Rust really likes when you pun on the names of the derive macro and trait and put them in the same path)
- fix all of the imports

## Notes to reviewers

- We could probably move more stuff into here, but I wanted to keep this PR as small as possible given the absurd level of import changes.
- This PR is ground work for my upcoming attempts to store resource data on components (resources-as-entities). Splitting this code out will make the work and review a bit easier, and is the sort of overdue refactor that's good to do as part of more meaningful work.

## Testing

cargo build works!

## Migration Guide

`bevy_ecs::system::Resource` has been moved to `bevy_ecs::resource::Resource`.

### DIFF
--- a/crates/bevy_a11y/src/lib.rs
+++ b/crates/bevy_a11y/src/lib.rs
@@ -24,7 +24,7 @@ use bevy_derive::{Deref, DerefMut};
 use bevy_ecs::{
     prelude::{Component, Event},
     schedule::SystemSet,
-    system::Resource,
+    resource::Resource,
 };
 
 /// Wrapper struct for [`accesskit::ActionRequest`]. Required to allow it to be used as an `Event`.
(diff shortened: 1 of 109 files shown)


viridia (approved):

BenjaminBrienen (approved):

BenjaminBrienen:
Small CI error

Jondolf (approved):
Looks good once CI passes

PR #17470, opened 2025-01-21
Move `spin` to `bevy_platform_support` out of other crates

bushrat011899:
# Objective

- Contributes to #16877

## Solution

- Expanded `bevy_platform_support::sync` module to provide API-compatible replacements for `std` items such as `RwLock`, `Mutex`, and `OnceLock`.
- Removed `spin` from all crates except `bevy_platform_support`.

## Testing

- CI

---

## Notes

- The sync primitives, while verbose, entirely rely on `spin` for their implementation requiring no `unsafe` and not changing the status-quo on _how_ locks actually work within Bevy. This is just a refactoring to consolidate the "hacks" and workarounds required to get a consistent experience when either using `std::sync` or `spin`.
- I have opted to rely on `std::sync` for `std` compatible locks, maintaining the status quo. However, now that we have these locks factored out into the own module, it would be trivial to investigate alternate locking backends, such as `parking_lot`.
- API for these locking types is entirely based on `std`. I have implemented methods and types which aren't currently in use within Bevy (e.g., `LazyLock` and `Once`) for the sake of completeness. As the standard library is highly stable, I don't expect the Bevy and `std` implementations to drift apart much if at all.


bushrat011899:
Marking as `X-Blessed` for the same reasons outlined [here](https://github.com/bevyengine/bevy/pull/17250#issuecomment-2579226902).

crates/bevy_ecs/src/intern.rs
```diff
@@ -143,24 +138,16 @@ impl<T: Internable + ?Sized> Interner<T> {
...
             }
         }
 
         {
-            #[cfg(feature = "std")]
             let mut set = self.0.write().unwrap_or_else(PoisonError::into_inner);
 
-            #[cfg(not(feature = "std"))]
-            let mut set = self.0.write();
-
```
BenjaminBrienen, line 163: I like how this cleans up

crates/bevy_platform_support/src/sync/once.rs
```diff
@@ -0,0 +1,232 @@
...
+    }
+
+    impl<T: Clone> Clone for OnceLock<T> {
+        fn clone(&self) -> OnceLock<T> {
+            let cell = Self::new();
+            if let Some(value) = self.get() {
+                match cell.set(value.clone()) {
+                    Ok(()) => (),
+                    Err(_) => unreachable!(),
+                }
```
BenjaminBrienen, line 131: I think this could be written more simple and succinct.

crates/bevy_tasks/src/usages.rs
BenjaminBrienen, line 1: So clean!

BenjaminBrienen (approved):
Looks pretty good. I'm impressed with how much progress is being made and I like how it makes engine development easier by making the code less noisy with feature gates.

crates/bevy_platform_support/src/sync/once.rs
bushrat011899: Yeah I was tossing around the idea of `unsafe` and using some `unchecked` method, but I'd rather keep it on the safer side. I'll have another look at this though...

crates/bevy_ecs/src/intern.rs
bushrat011899: That's exactly my goal here. It's technically more lines of code total, but it's _far_ easier to reason about.

crates/bevy_platform_support/src/sync/once.rs
BenjaminBrienen: Why not `unwrap()`?

bushrat011899: Yeah honestly I should just use that, no need to be fancy here. I've cleaned up a few of these patterns to use `ok_or`, etc.

BenjaminBrienen (approved):

crates/bevy_platform_support/src/lib.rs
```diff
@@ -17,3 +17,11 @@ extern crate alloc;
 
 pub mod sync;
 pub mod time;
+
+/// Frequently used items which would typically be included in most contexts.
+pub mod prelude {
+    #[cfg(feature = "alloc")]
+    pub use alloc::{
+        borrow::ToOwned, boxed::Box, format, string::String, string::ToString, vec, vec::Vec,
+    };
```
BD103, line 26: Could you add a comment on why we're re-exporting types in the `std` prelude? (Maybe link to discussion in previous PRs, so people aren't confused about this in the future.)

BD103 (approved):
Implementation is cleanly done, love it! 👍

crates/bevy_platform_support/src/lib.rs
bushrat011899: I've added some documentation explaining the purpose of this prelude (ease transition for users when writing `no_std` code with Bevy). I've also added a private comment listing out which items are missing from the prelude, as a sort-of todo list for the future.

BD103: Thank you! The changes look good :)

PR #17475, opened 2025-01-21
Mesh::merge to return a Result

Vrixyz:
# Objective

Make `Mesh::merge` more resilient to use.

Currently, it's difficult to make sure `Mesh::merge` will not panic (we'd have to check if all attributes are compatible).

- I'd appreciate it for utility function to convert different mesh representations such as: https://github.com/dimforge/bevy_rapier/pull/628.

## Solution

- Make `Mesh::merge` return a `Result`.

## Testing

- It builds

## Migration Guide

- `Mesh::merge` now returns a `Result<(), MeshMergeError>`.


crates/bevy_mesh/src/mesh.rs
```diff
@@ -785,20 +795,21 @@ impl Mesh {
     ///
     /// `Aabb` of entities with modified mesh are not updated automatically.
     ///
-    /// # Panics
+    /// # Errors
     ///
-    /// Panics if the vertex attribute values of `other` are incompatible with `self`.
+    /// Errors out if the vertex attribute values of `other` are incompatible with `self`.
```
greeble-dev, line 800:
```suggestion
    /// Returns [`Err(MergeMeshError)`](MergeMeshError) if the vertex attribute values of `other` are incompatible with `self`.
```

crates/bevy_mesh/src/mesh.rs
```diff
@@ -785,20 +795,21 @@ impl Mesh {
...
     /// For example, [`VertexAttributeValues::Float32`] is incompatible with [`VertexAttributeValues::Float32x3`].
+    ///
+    /// In that case, [`self`] is not modified.
```
greeble-dev, line 803: That doesn't seem true if I'm reading the code right? It could append some attributes but then error out on a later one?

IceSentry (approved):

greeble-dev (approved):
Added some documentation nitpicks.

I feel like the names of the `attributes` and `attributes_data` accessors are now a bit surprising. But I guess that's outside the scope of this PR.

alice-i-cecile:
@Vrixyz ping me when this is ready and I'll get this in.

Vrixyz:
I removed the wrong comment, thanks reviewers ; @alice-i-cecile here goes your requested ping.

> I feel like the names of the attributes and attributes_data accessors are now a bit surprising? But I guess that's outside the scope of this PR.

Indeed. The new accessor is `pub(crate)` though so as it's not public facing API, so I'd agree that's outside this PR's scope.

alice-i-cecile:
I've resolved merge conflicts for you but CI is 100% gonna gripe about that allow statement :)

crates/bevy_mesh/src/mesh.rs
```diff
@@ -788,19 +798,19 @@ impl Mesh {
...
     /// `Aabb` of entities with modified mesh are not updated automatically.
     ///
-    /// # Panics
+    /// # Errors
     ///
-    /// Panics if the vertex attribute values of `other` are incompatible with `self`.
+    /// Returns [`Err(MergeMeshError)`](MergeMeshError) if the vertex attribute values of `other` are incompatible with `self`.
     /// For example, [`VertexAttributeValues::Float32`] is incompatible with [`VertexAttributeValues::Float32x3`].
-    pub fn merge(&mut self, other: &Mesh) {
+    #[allow(clippy::match_same_arms)]
```
BenjaminBrienen, line 805:
```suggestion
```

Issue #17468, opened 2025-01-21
Runtime generation of simple environment maps


viridia:
## What problem does this solve or what need does it fill?

This proposes to kill several birds with one stone:
* Simplifying shaders by removing ambient lights and replacing them with simple 1x1 environment maps.
* Implementing hemispherical lights.

Creating environment maps for Bevy is a difficult process, particularly on Mac where you have to compile multiple C++ packages to get the tooling needed to do the conversion. Other game frameworks such as three.js provide a means to generate environment maps from an image at runtime.

Ambient lights can be emulated by a 1x1 environment map, which would mean that we could remove the shader code for ambient lights, thereby reducing the lines of code for our standard shaders.

## What solution would you like?

From [the discord](https://discord.com/channels/691052431525675048/743663924229963868/1331105565195112551):

```rust
foo.insert(EnvironmentMapLight::solid_color(SLATE_BLUE.into()))
foo.insert(EnvironmentMapLight::hemispherical_gradient(BLUE.into(), BROWN.into()))
```

The idea is to provide methods for generating simple environment maps from a set of parameters.

We should also offer a means to generate an environment map from an equirectangular image, but that is a more advanced project, and should probably be a separate issue.

For backwards compatibility, the `AmbientLight` can be transformed into an environment map light, either via the constructor or perhaps by a hook of some kind.

## What alternative(s) have you considered?

What I have done in the past is posted an image on Discord and had someone convert it to an environment map for me.

## Additional context

Supersedes: #12027

mate-h:
We have definitely partially addressed this issue. Since we landed  #19076 and #20529 I want to repurpose this issue for creating a dedicated example for runtime generation of environment maps with a simple gradient compute shader, that draws a hemisphere light as described in the original issue. 
Therefore I am including it in the tracking task for the environment map filtering at #20212 .

Igor-dvr:
> Implementing hemispherical lights.

There is OpenUSD spec for this. Dome light type supports spherical and hemispherical map and also solid color iiuc.

I left comment here https://github.com/bevyengine/bevy/pull/18207#issuecomment-3303176627 :

> Iiuc we should probably have environment/dome light type that encapsulates an enum: Color/AmbientLight, EnvironmetTexture, SkyTexture, something else. You can configure this light with only one option but can have multiple such lights with different options in a scene. Related:
[...]

PR #17482, opened 2025-01-21
Replace Ambient Lights with Environment Map Lights

SparkyPotato:
# Objective

Transparently uses simple `EnvironmentMapLight`s to mimic `AmbientLight`s. Implements the first part of #17468, but I can implement hemispherical lights in this PR too if needed.

## Solution

- A function `EnvironmentMapLight::solid_color(&mut Assets<Image>, Color)` is provided to make an environment light with a solid color.
- A new system is added to `SimulationLightSystems` that maps `AmbientLight`s on views or the world to a corresponding `EnvironmentMapLight`.

I have never worked with (or on) Bevy before, so nitpicky comments on how I did things are appreciated :).

## Testing

Testing was done on a modified version of the `3d/lighting` example, where I removed all lights except the ambient light. I have not included the example, but can if required.

## Migration
`bevy_pbr::AmbientLight` has been deprecated, so all usages of it should be replaced by a `bevy_pbr::EnvironmentMapLight` created with `EnvironmentMapLight::solid_color` placed on the camera. There is no alternative to ambient lights as resources.


crates/bevy_pbr/src/light/mod.rs
```diff
@@ -522,6 +523,26 @@ pub enum SimulationLightSystems {
     CheckLightVisibility,
 }
 
+pub fn map_ambient_lights(
```
IceSentry, line 526: I think we may want to add a deprecation notice to the AmbientLight component and remove it after 0.16. For this PR I think this is fine as is though.

IceSentry (approved):
This is great, thank you!

IceSentry:
>  I can implement hemispherical lights in this PR too if needed

The PR is simple enough that you could add this in too I think. Up to you, but I would approve it either way.

SparkyPotato:
> > I can implement hemispherical lights in this PR too if needed
> 
> The PR is simple enough that you could add this in too I think. Up to you, but I would approve it either way.

Might as well add it in then.

robtfm:
code looks good. 

simplifying the shaders is nice, but is the cost of binding an extra texture per view and looking up for every pixel really worth the saving of 7-8 lines of shader code?

IceSentry:
> simplifying the shaders is nice, but is the cost of binding an extra texture per view and looking up for every pixel really worth the saving of 7-8 lines of shader code?

I guess what we could do is keep the constructor thing in this PR and do the ambient light thing separately? I can see why that part could be a bit more controversial and that people might still want that around.

pcwalton:
It's not just about simplifying the shader code: it's that the interaction between ambient and environment map light never made any sense. They should be considered basically mutually exclusive (strictly speaking, ambient should be considered a type of light probe, which this PR moves us toward).

crates/bevy_pbr/src/light_probe/environment_map.rs
```diff
@@ -114,6 +116,80 @@ pub struct EnvironmentMapLight {
...
+                    mid_color,
+                    top_color,
+                    bottom_color,
+                    mid_color,
+                    mid_color,
+                ]
+                .into_iter()
+                .flat_map(Srgba::to_u8_array)
+                .collect(),
+                TextureFormat::Rgba8UnormSrgb,
```
mate-h, line 179: This does not need to be a HDR texture right, since the user is defining the color in sRGB and an intensity?

robtfm:
> The lighting example does go from about 0.51 ms to 0.55 ms on my machine (and enabling deferred makes that 0.46 ms to 0.49 ms)

is that in release? if so then i'd argue a ~10% fps reduction is not justified, but it's surprising how large the impact is.


> They should be considered basically mutually exclusive (strictly speaking, ambient should be considered a type of light probe, which this PR moves us toward).

makes sense. if the impact is really this big then perhaps we need some investigation into env map perf before going further though?

SparkyPotato:
> > The lighting example does go from about 0.51 ms to 0.55 ms on my machine (and enabling deferred makes that 0.46 ms to 0.49 ms)
> 
> is that in release? if so then i'd argue a ~10% fps reduction is not justified, but it's surprising how large the impact is.

In release, the difference is 0.49 ms 0.53 ms, but the deferred timings don't change.

> > They should be considered basically mutually exclusive (strictly speaking, ambient should be considered a type of light probe, which this PR moves us toward).
> 
> makes sense. if the impact is really this big then perhaps we need some investigation into env map perf before going further though?

I'd think ambient lights are usually used only in simple scenes where we'd be CPU bound anyways. For anything more complex, users would probably switch to an environment map or one baked by the procedural sky.

SparkyPotato:
I decided to profile the case without an ambient light or envmap, which takes 0.47 ms on this branch.

This means we'll actually be slightly _faster_ in scenes with an envmap after this PR. For example: 
- the deferred rendering example goes from 1.16 to 1.12 ms
- SSR goes from 0.92 to 0.89 ms
- post processing goes from 1.02 to 0.96 ms
- meshlet goes from 1.68 to 1.67 ms
- color grading goes from 0.88 to 0.87 ms

Note that these are all overall frametimes and not individual pass timings like that of the lighting example (for lighting, the overall frame went from 0.98 to 0.96 ms without any ambient or env light).

crates/bevy_pbr/src/light_probe/environment_map.rs
```diff
@@ -114,6 +116,80 @@ pub struct EnvironmentMapLight {
     pub affects_lightmapped_mesh_diffuse: bool,
 }
 
+impl EnvironmentMapLight {
```
JMS55, line 119: Needs docs for the new functions. Esp since EnviornmentMapLight is supposed to have diffuse/specular components that get run through a bunch of sampling and multiscattering stuff, having a single "color" is a bit confusing. Please add some docs and I guess note that it's a bit of a hack? Something along these lines.

crates/bevy_pbr/src/light/mod.rs
JMS55: +1 to adding a deprecated attribute to ambient light, and then removing this system in 0.17.

JMS55 (changes requested):
Code looks fine. Needs some docs, as well as a changelog and migration guide in the PR description. Ask on discord how to do that, I think we changed our process recently and I forget what the new system is.

SparkyPotato:
I think I've addressed all requested changes and updated all the examples to use `EnvironmentMapLight` except when disabling the default `AmbientLight` (which will, of course, no longer be required when we remove `AmbientLight`).

IceSentry:
I started an example runner run here https://github.com/TheBevyFlock/bevy-example-runner/issues/117

Other than that LGTM. If someone is concerned with the texture sampling for performance reasons they probably want to use a custom material that is just simpler in general too, but I think this setup makes sense for most users of StandardMaterial.

JMS55 (approved):

alice-i-cecile:
Thank you to everyone involved with the authoring or reviewing of this PR! This work is relatively important and needs release notes! Head over to https://github.com/bevyengine/bevy-website/issues/1998 if you'd like to help out.

PR #17486, opened 2025-01-21
Move all internal dependancies into the workspace.

### LINKED ISSUE #16652: Use Cargo workspace features for internal dependency versions
Currently all of the Bevy sub-packages which depend on other parts of Bevy use explicit version numbers. This means that whenever a version of a package is bumped, you have to change the version number in every Cargo.toml that references it.

It would be much less work to use Cargo's workspace versioning feature:

```
bevy_app = { workspace = true }
```
This will use the version number specified in the top-level Cargo.toml (in the section `workspace.dependencies`).

Doing this means that the package version for each crate only needs to be specified in two places: in the package itself, and in the top-level `workspace.dependencies`. This will making bumping the versions much easier.

AlephCubed:
This moves all internal crate dependancies into the root `Cargo.toml`, which can be referenced using `workspace = true`.

Fixes #16652.

### Default features
Because of the way workspace dependancies work, `default-features` are now opt-in (see [this warning](https://doc.rust-lang.org/nightly/edition-guide/rust-2024/cargo-inherited-default-features.html) for details).


crates/bevy_animation/Cargo.toml
```diff
@@ -10,21 +10,21 @@ keywords = ["bevy"]
...
-bevy_log = { path = "../bevy_log", version = "0.16.0-dev" }
-bevy_math = { path = "../bevy_math", version = "0.16.0-dev" }
-bevy_reflect = { path = "../bevy_reflect", version = "0.16.0-dev", features = [
+bevy_app = { workspace = true, default-features = true }
+bevy_asset = { workspace = true, default-features = true }
+bevy_color = { workspace = true, default-features = true }
+bevy_derive = { workspace = true, default-features = true }
+bevy_log = { workspace = true, default-features = true }
+bevy_math = { workspace = true, default-features = true }
+bevy_reflect = { workspace = true, default-features = true, features = [
```
alice-i-cecile, line 19: Won't this enable these features for all of the crates that rely on bevy_reflect across the workspace? My understanding was that only a single version of the crate is selected.

AlephCubed: According to [this](https://doc.rust-lang.org/cargo/reference/resolver.html#features), it seems like that is already the case for the current method:
> When building multiple packages in a workspace (such as with --workspace or multiple -p flags), the features of the dependencies of all of those packages are unified. If you have a circumstance where you want to avoid that unification for different workspace members, you will need to build them via separate cargo invocations.

[This post](https://nickb.dev/blog/cargo-workspace-and-the-feature-unification-pitfall/) goes over some of the pitfalls of dependancy unification, but that appears to apply to workspaces in general, not just workspace level dependancies.

But I could totally be wrong here. It's kinda confusing.

BenjaminBrienen: I think the main concern is when a project has a direct dependency on a crate like bevy_ecs, they shouldn't have extra features.

AlephCubed: I don't _think_ that will be effected. Only members built at the same time should have their dependencies unified. 
> Just beware that if you build multiple workspace members at the same time, the features will be unified so that if one member sets default-features = true (which is the default if not explicitly set), the default-features will be enabled for all members using that dependency.

- [source](https://doc.rust-lang.org/nightly/edition-guide/rust-2024/cargo-inherited-default-features.html)

But upon rereading this, I think I was wrong about `default-features` being opt out when disabled on a workspace level. 
I will see about fixing that tomorrow.

AlephCubed: I just tested running `cargo tree -p bevy_ui -e features` and only the `bevy` feature (and its sub-features) are enabled for `bevy_reflect`, not the default features.
So if you compile an individual sub-crate, only its required features should be enabled.

AlephCubed: I just tested running `cargo tree -p bevy_ecs -e features` and `bevy_reflect` has no features enabled, same as in main.
So if you compile an individual sub-crate, only its required features should be enabled.

mockersf:
blocked on https://github.com/dependabot/dependabot-core/issues/7896

janhohenheim:
Unblocked by now, labeling as waiting for author due to conflicts. Approving in anticipation :)

alice-i-cecile:
I think we should use workspace dependencies, but this is a big project-level decision. Marked as X-Controversial. Probably going to be easier to debate this and then redo the PR from scratch rather than resolving conflicts.

PR #17489, opened 2025-01-22
Adds update interval config for FpsOverlayPlugin

### LINKED ISSUE #17487: Add update interval config for `FpsOverlayPlugin`

Current the FpsOverlayPlugin has a very short update interval, like the video below:

https://github.com/user-attachments/assets/94730357-8933-4c9a-9481-26f1c9d87944

Of course we can see that it's approximately 59 fps, because it a relatively stable frame rate. But sometimes we may have some unstable frame rate that may fluctuate from 30+ FPS to 40+ FPS where we can hardly say what the specific frame rate is(I just met this problem recently, and I may add a video example of that if I am available tomorrow).

Add `interval` (or other name, I am not good at naming) field, and the FPS displayed will be the average fps of the past interval duration.

```rust
App::new()
        .add_plugins((DefaultPlugins, FpsOverlayPlugin {
            config: FpsOverlayConfig {
                interval: bevy::utils::Duration::from_millis(200),
                ..default()
            },
        }))
        .add_systems(Startup, setup)
        .run();
```

- Write a simple fps overlay plugin myself
- Use [MSI Afterburner](https://www.msi.com/Landing/afterburner) or Xbox Game Bar

spvky:
# Objective
Fixes #17487 

- Adds a new field `refresh_interval` to `FpsOverlayConfig` to allow the user setting a minimum time before each refresh of the FPS display

## Solution

- Add `refresh_interval` to `FpsOverlayConfig`
- When updating the on screen text, check a duration of `refresh_interval` has passed, if not, don't update the FPS counter

## Testing

- Created a new bevy project
- Included the `FpsOverlayPlugin` with the default `refresh_interval` (100 ms)
- Included the `FpsOverlayPlugin` with an obnoxious `refresh_interval` (2 seconds)
---

### DIFF
--- a/crates/bevy_dev_tools/src/fps_overlay.rs
+++ b/crates/bevy_dev_tools/src/fps_overlay.rs
@@ -116,7 +116,6 @@ fn update_text(
     diagnostic: Res<DiagnosticsStore>,
     query: Query<Entity, With<FpsText>>,
     mut writer: TextUiWriter,
-    //
     time: Res<Time>,
     config: Res<FpsOverlayConfig>,
     mut time_since_rerender: Local<Duration>,
--- a/crates/bevy_dev_tools/src/fps_overlay.rs
+++ b/crates/bevy_dev_tools/src/fps_overlay.rs
@@ -21,7 +21,7 @@ use bevy_ui::{
     widget::{Text, TextUiWriter},
     GlobalZIndex, Node, PositionType,
 };
-use std::time::Duration;
+use core::time::Duration;
 
 /// [`GlobalZIndex`] used to render the fps overlay.
 ///
--- a/crates/bevy_dev_tools/src/fps_overlay.rs
+++ b/crates/bevy_dev_tools/src/fps_overlay.rs
@@ -8,6 +8,7 @@ use bevy_ecs::{
     change_detection::DetectChangesMut,
     component::Component,
     entity::Entity,
+    prelude::Local,
     query::With,
     resource::Resource,
     schedule::{common_conditions::resource_changed, IntoSystemConfigs},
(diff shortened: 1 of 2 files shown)


crates/bevy_dev_tools/src/fps_overlay.rs
```diff
@@ -110,11 +116,19 @@ fn update_text(
     diagnostic: Res<DiagnosticsStore>,
     query: Query<Entity, With<FpsText>>,
     mut writer: TextUiWriter,
+    //
```
BenjaminBrienen, line 119:
```suggestion
```

BenjaminBrienen (approved):

greeble-dev (approved):

crates/bevy_dev_tools/src/fps_overlay.rs
```diff
@@ -110,11 +116,18 @@ fn update_text(
...
+    config: Res<FpsOverlayConfig>,
+    mut time_since_rerender: Local<Duration>,
 ) {
-    for entity in &query {
-        if let Some(fps) = diagnostic.get(&FrameTimeDiagnosticsPlugin::FPS) {
-            if let Some(value) = fps.smoothed() {
-                *writer.text(entity, 1) = format!("{value:.2}");
+    *time_since_rerender += time.delta();
+    if *time_since_rerender >= config.refresh_interval {
+        *time_since_rerender = Duration::default();
```
alice-i-cecile, line 125:
```suggestion
        *time_since_rerender = Duration::ZERO;
```

crates/bevy_dev_tools/src/fps_overlay.rs
```diff
@@ -65,6 +68,8 @@ pub struct FpsOverlayConfig {
     pub text_color: Color,
     /// Displays the FPS overlay if true.
     pub enabled: bool,
+    /// How often the FPS overlay re-renders.
```
alice-i-cecile, line 71:
```suggestion
    /// The period after which the FPS overlay re-renders.
    ///
    /// Defaults to once every 100 ms.
```

alice-i-cecile:
I think our smoothing window is also too small by default, but this is a nice improvement to waste fewer resources :)

PR #17490, opened 2025-01-22
`no_std` Support for `bevy_input_focus`

bushrat011899:
# Objective

- Contributes to #15460

## Solution

- Switched `tracing` for `log` for the atomically challenged platforms
- Setup feature flags as required
- Added to `compile-check-no-std` CI task

## Testing

- CI

---

## Notes

- _Very_ easy one this time. Most of the changes here are just feature definitions and documentation within the `Cargo.toml`

### DIFF
--- a/crates/bevy_input_focus/Cargo.toml
+++ b/crates/bevy_input_focus/Cargo.toml
@@ -10,7 +10,9 @@ keywords = ["bevy"]
 rust-version = "1.83.0"
 
 [features]
-default = ["bevy_reflect"]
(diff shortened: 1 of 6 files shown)


BenjaminBrienen (approved):

TimJentzsch (approved):
Are there any negative effects when part of the crates use tracing and others log, e.g. harder to configure the log messages?

If yes, perhaps we can make it feature dependent in some way. Wouldnt have to happen in this PR though.

PR #17495, opened 2025-01-22
Only include distance fog in the PBR shader if the view uses it.

pcwalton:
Right now, we always include distance fog in the shader, which is unfortunate as it's complex code and is rare. This commit changes it to be a `#define` instead. I haven't confirmed that removing distance fog meaningfully reduces VGPR usage, but it can't hurt.


IceSentry (approved):

Elabajaba (approved):
Checked with radeon gpu profiler and it does improve our register pressure.

JMS55:
Queue material meshes system is forward only. Missing deferred lighting pass, and forward meshlet pass (yes our specialization system is very confusing now).

PR #9660, opened 2023-08-31
Add font size warning to TextSection

hate:
# Objective

- `TextSection`s are unable to display their text value if the `font_size` is less than or equal to 0, thus @ickshonpe suggested we emit a warning to the console whenever this occurs.
- Fixes #9655

## Solution

- Emit a warning if we detect that the font_size is less than or equal to 0.
- Add `bevy_log` as a dependency to `bevy_text` to be able to access the `warn!` macro.

PR #17501, opened 2025-01-22
Add warning for font sizes `<= 0.0`

### LINKED ISSUE #9655: Negative `font_size` warning

At the moment if text is spawned with a negative `font_size` value, it fails silently and nothing is displayed. 

I think we should emit a warning at least.

rparrett:
# Objective

Alternative to #9660, which is outdated since "required components" landed.

Fixes #9655

## Solution

This is a different approach than the linked PR, slotting the warning into an existing check for zero or negative font sizes in the text pipeline.

## Testing

Replaced a font size with `0.0` in `examples/ui/text.rs`.

```
2025-01-22T23:26:08.239688Z  INFO bevy_winit::system: Creating new window App (0v1)
2025-01-22T23:26:08.617505Z  WARN bevy_text::pipeline: Text span 10v1 has a font size <= 0.0. Nothing will be displayed.
```

### DIFF
--- a/crates/bevy_text/src/pipeline.rs
+++ b/crates/bevy_text/src/pipeline.rs
@@ -10,6 +10,7 @@ use bevy_ecs::{
     system::{ResMut, Resource},
 };
 use bevy_image::prelude::*;
+use bevy_log::{once, warn};
 use bevy_math::{UVec2, Vec2};
 use bevy_reflect::{std_traits::ReflectDefault, Reflect};
 use bevy_utils::HashMap;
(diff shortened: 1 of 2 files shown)


alice-i-cecile (approved):

ickshonpe (approved):

crates/bevy_text/src/pipeline.rs
```diff
@@ -141,6 +142,10 @@ impl TextPipeline {
 
             // Save spans that aren't zero-sized.
             if scale_factor <= 0.0 || text_font.font_size <= 0.0 {
+                once!(warn!(
+                    "Text span {entity} has a font size <= 0.0. Nothing will be displayed.",
```
ickshonpe, line 146: Maybe it might be more useful to quote the text from the text span than its entity id (or give both)?

ickshonpe (approved):

crates/bevy_text/src/pipeline.rs
mockersf: or make the size random each frame, that should help quickly noticing where's the error 😄

mockersf (approved):

BenjaminBrienen (approved):

PR #17503, opened 2025-01-22
Refactored `ComponentHook` Parameters into `HookContext`

bushrat011899:
# Objective

- Make the function signature for `ComponentHook` less verbose

## Solution

- Refactored `Entity`, `ComponentId`, and `Option<&Location>` into a new `HookContext` struct.

## Testing

- CI

---

## Migration Guide

Update the function signatures for your component hooks to only take 2 arguments, `world` and `context`. Note that because `HookContext` is plain data with all members public, you can use de-structuring to simplify migration.

```rust
// Before
fn my_hook(
    mut world: DeferredWorld,
    entity: Entity,
    component_id: ComponentId,
) { ... }

// After
fn my_hook(
    mut world: DeferredWorld,
    HookContext { entity, component_id, caller }: HookContext,
) { ... }
``` 

Likewise, if you were discarding certain parameters, you can use `..` in the de-structuring:

```rust
// Before
fn my_hook(
    mut world: DeferredWorld,
    entity: Entity,
    _: ComponentId,
) { ... }

// After
fn my_hook(
    mut world: DeferredWorld,
    HookContext { entity, .. }: HookContext,
) { ... }
```

### DIFF
--- a/crates/bevy_core_pipeline/src/oit/mod.rs
+++ b/crates/bevy_core_pipeline/src/oit/mod.rs
@@ -70,11 +70,11 @@ impl Component for OrderIndependentTransparencySettings {
     type Mutability = Mutable;
 
     fn register_component_hooks(hooks: &mut ComponentHooks) {
-        hooks.on_add(|world, entity, _, caller| {
-            if let Some(value) = world.get::<OrderIndependentTransparencySettings>(entity) {
+        hooks.on_add(|world, context| {
+            if let Some(value) = world.get::<OrderIndependentTransparencySettings>(context.entity) {
                 if value.layer_count > 32 {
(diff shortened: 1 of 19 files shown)


bushrat011899:
Marking as `X-Blessed` based on [this message](https://discord.com/channels/691052431525675048/749335865876021248/1331762792797179984) from @cart.

alice-i-cecile (approved):

viridia (approved):

hymm:
I don't think the migration guide for `before` should include the `caller` field, since that's new since 0.15.

bushrat011899:
> I don't think the migration guide for `before` should include the `caller` field, since that's new since 0.15.

Good catch! I always forget to make sure the migration is from a 0.15 perspective and not a `main` one.

Bleachfuel (approved):

PR #17510, opened 2025-01-23
Combine the shader point and spot light lookup into one loop.

pcwalton:
Point and spot lights are treated similarly. In fact, the point light calculations are part of the spot light calculations. So we can simplify the code by unifying the two loops. This might make life easier for the shader compiler too.

### DIFF
--- a/crates/bevy_pbr/src/render/pbr_lighting.wgsl
+++ b/crates/bevy_pbr/src/render/pbr_lighting.wgsl
@@ -547,7 +547,7 @@ fn point_light(
 // light to give it the spot shape.
 //
 // The resulting value should be multiplied by the luminance that results from
-// treating the spot light as thought it were a point light (i.e. the results of
+// treating the spot light as though it were a point light (i.e. the results of
 // calling `point_light`).
 fn spot_light(light_id: u32, input: ptr<function, LightingInput>) -> f32 {
     let light = &view_bindings::clusterable_objects.data[light_id];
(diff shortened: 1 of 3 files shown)


crates/bevy_pbr/src/render/pbr_lighting.wgsl
```diff
@@ -543,14 +543,13 @@ fn point_light(
...
-    enable_diffuse: bool
-) -> vec3<f32> {
-    // reuse the point light calculations
-    let point_light = point_light(light_id, input, enable_diffuse);
-
+// Returns the additional attenuation that should be applied to the given spot
+// light to give it the spot shape.
+//
+// The resulting value should be multiplied by the luminance that results from
+// treating the spot light as thought it were a point light (i.e. the results of
```
wilk10, line 550:
```suggestion
// treating the spot light as though it were a point light (i.e. the results of
```

IceSentry (approved):
I started an example run https://github.com/TheBevyFlock/bevy-example-runner/issues/94

Code looks good

Elabajaba:
This regresses both VGPR and SGPR on my machine.

pcwalton:
@Elabajaba Ugh, that's so annoying. I wonder if it'll end up winning when we do more (like disable PCSS when not in use) though? This stuff is hard to predict.

pcwalton:
I'm going to close this for now because it regresses VGPR.

PR #17513, opened 2025-01-23
implement FromEntitySetIterator

Victoronz:
# Objective

Some collections are more efficient to construct when we know that every element is unique in advance.
We have `EntitySetIterator`s from #16547, but currently no API to safely make use of them this way.

## Solution

Add `FromEntitySetIterator` as a subtrait to `FromIterator`, and implement it for the `EntityHashSet`/`hashbrown::HashSet` types.
To match the normal `FromIterator`, we also add a `EntitySetIterator::collect_set` method.
It'd be better if these methods could shadow `from_iter` and `collect` completely, but https://github.com/rust-lang/rust/issues/89151 is needed for that.

While currently only `HashSet`s implement this trait, future `UniqueEntityVec`/`UniqueEntitySlice` functionality comes with more implementors.

Because `HashMap`s are collected from tuples instead of singular types, implementing this same optimization for them is more complex, and has to be done separately.

## Showcase

This is basically a free speedup for collecting `EntityHashSet`s!

```rust
pub fn collect_milk_dippers(dippers: Query<Entity, (With<Milk>, With<Cookies>)>) {
    dippers.iter().collect_set::<EntityHashSet>();
    // or
    EntityHashSet::from_entity_set_iter(dippers);
}


BenjaminBrienen (approved):

Victoronz:
> slightly faster than the standard `collect` (the spawning part of this benchmark takes about 26ms, with this PR improving the iter/collect part from ~49ms to ~47ms). Not a huge difference :/

What system/compilation settings did you use to bench this?

Victoronz:
I've also changed the for loop into a fold, since internal iteration is known to generate bit better code.
Might not/probably does not have an impact for simple/common iterators though.

PR #17514, opened 2025-01-23
Add simple Disabled marker

### LINKED ISSUE #17458: Add a first-party `Disabled` component
With https://github.com/bevyengine/bevy/pull/13120 merging, it's important that we ship a simple `Disabled` unit struct and register it as a default query filter to make sure this feature is useful to beginners out of the box.

NiseVoid:
# Objective

We have default query filters now, but there is no first-party marker for entity disabling yet
Fixes #17458

## Solution

Add the marker, cool recursive features and/or potential hook changes should be follow up work

## Testing

Added a unit test to check that the new marker is enabled by default

### DIFF
--- a/crates/bevy_ecs/src/entity_disabling.rs
+++ b/crates/bevy_ecs/src/entity_disabling.rs
@@ -28,10 +28,14 @@ use crate::{
 };
 use bevy_ecs_macros::{Component, Resource};
 
+#[cfg(feature = "bevy_reflect")]
+use {crate::reflect::ReflectComponent, bevy_reflect::Reflect};
+
 /// A marker component for disabled entities. See [the module docs] for more info.
 ///
 /// [the module docs]: crate::entity_disabling
 #[derive(Component)]
+#[cfg_attr(feature = "bevy_reflect", derive(Reflect), reflect(Component))]
 pub struct Disabled;
 
 /// The default filters for all queries, these are used to globally exclude entities from queries.
--- a/crates/bevy_ecs/src/entity_disabling.rs
+++ b/crates/bevy_ecs/src/entity_disabling.rs
@@ -26,7 +26,13 @@ use crate::{
     component::{ComponentId, Components, StorageType},
     query::FilteredAccess,
 };
-use bevy_ecs_macros::Resource;
+use bevy_ecs_macros::{Component, Resource};
+
+/// A marker component for disabled entities. See [the module docs] for more info.
+///
+/// [the module docs]: crate::entity_disabling
+#[derive(Component)]
+pub struct Disabled;
 
 /// The default filters for all queries, these are used to globally exclude entities from queries.
 /// See the [module docs](crate::entity_disabling) for more info.
@@ -37,10 +43,6 @@ pub struct DefaultQueryFilters {
 }
 
 impl DefaultQueryFilters {
-    #[cfg_attr(
-        not(test),
-        expect(dead_code, reason = "No Disabled component exist yet")
-    )]
     /// Set the [`ComponentId`] for the entity disabling marker
     pub(crate) fn set_disabled(&mut self, component_id: ComponentId) -> Option<()> {
         if self.disabled.is_some() {
--- a/crates/bevy_ecs/src/lib.rs
+++ b/crates/bevy_ecs/src/lib.rs
@@ -1531,6 +1531,7 @@ mod tests {
     #[test]
     fn filtered_query_access() {
         let mut world = World::new();
+        // We remove entity disabling so it doesn't affect our query filters
         world.remove_resource::<DefaultQueryFilters>();
         let query = world.query_filtered::<&mut A, Changed<B>>();
 
--- a/crates/bevy_ecs/src/lib.rs
+++ b/crates/bevy_ecs/src/lib.rs
@@ -134,6 +134,7 @@ mod tests {
         change_detection::Ref,
         component::{require, Component, ComponentId, RequiredComponents, RequiredComponentsError},
         entity::Entity,
+        entity_disabling::DefaultQueryFilters,
         prelude::Or,
         query::{Added, Changed, FilteredAccess, QueryFilter, With, Without},
         resource::Resource,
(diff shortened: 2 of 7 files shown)


BenjaminBrienen:
I think that this should close that issue. Would you mind fixing the description? :)

crates/bevy_ecs/src/entity_disabling.rs
```diff
@@ -26,7 +26,13 @@ use crate::{
...
     query::FilteredAccess,
 };
-use bevy_ecs_macros::Resource;
+use bevy_ecs_macros::{Component, Resource};
+
+/// A marker component for disabled entities. See [the module docs] for more info.
+///
+/// [the module docs]: crate::entity_disabling
+#[derive(Component)]
+pub struct Disabled;
```
MrGVSV, line 35: Should this be reflected?
```suggestion
#[derive(Component)]
#[cfg_attr(feature = "bevy_reflect", derive(Reflect), reflect(Component))]
pub struct Disabled;
```

crates/bevy_ecs/src/world/mod.rs
```diff
@@ -3217,6 +3223,15 @@ impl World {
...
 
 impl World {
+    /// Returns true if the type id is used internally, components and resources used internally
+    /// most likely should not be affected by things like saving and loading
+    pub fn is_internal_type(type_id: TypeId) -> bool {
+        if type_id == TypeId::of::<DefaultQueryFilters>() {
+            return true;
+        }
+        false
+    }
```
MrGVSV, line 3233: I feel like this could be a little misleading. Users may expect this to return `true` for all internal-ish types. In their minds this might include computer components like `ComputedVisibility`, observer data, or other engine-specific types.

So rather than introducing this method now with only a single type actually included, what if we simply hardcode this logic in `be y_scene` where it seems to be used?

crates/bevy_scene/src/scene.rs
```diff
@@ -61,6 +62,15 @@ impl Scene {
 
         // Resources archetype
         for (component_id, resource_data) in self.world.storages().resources.iter() {
+            if world
+                .components()
+                .get_info(component_id)
+                .and_then(ComponentInfo::type_id)
+                .filter(|&type_id| World::is_internal_type(type_id))
```
MrGVSV, line 69: I think it might be more consistent to filter this resource by its `ComponentId` rather than its `TypeId`.

crates/bevy_scene/src/scene_filter.rs
```diff
@@ -154,6 +155,9 @@ impl SceneFilter {
     ///
     /// [`Unset`]: SceneFilter::Unset
     pub fn is_allowed_by_id(&self, type_id: TypeId) -> bool {
+        if World::is_internal_type(type_id) {
+            return false;
+        }
```
MrGVSV, line 160: `SceneFilter` is meant to be somewhat general purpose and allow for any type to be filtered. As such, I think this code should remain the same and this sort of hardcoding done in the actual scene logic like you already have.

Note that eventually we do want to have constructors for `SceneFilter` that come with sensible defaults, but that's likely outside the scope of this PR as it has a few decisions/problems that need working out.

crates/bevy_ecs/src/lib.rs
```diff
@@ -1530,6 +1531,7 @@ mod tests {
     #[test]
     fn filtered_query_access() {
         let mut world = World::new();
+        world.remove_resource::<DefaultQueryFilters>();
```
cBournhonesque, line 1534: I think it might be worth adding a comment explaining that the DefaultQueryFilters modify the access and would make the test harder to reason about

crates/bevy_ecs/src/world/mod.rs
```diff
@@ -159,6 +160,11 @@ impl World {
 
         let on_despawn = OnDespawn::register_component_id(self);
         assert_eq!(ON_DESPAWN, on_despawn);
+
+        let disabled = self.register_component::<Disabled>();
+        let mut filters = DefaultQueryFilters::default();
+        filters.set_disabled(disabled);
```
cBournhonesque, line 166: I'm a bit worried that users might do a 
```app.init_resource::<DefaultQueryFilters>()``` 

and lose the `Disabled` functionality inadvertently..

Comment (NiseVoid) [reply crates/bevy_ecs/src/world/mod.rs:166]: `init_resource` doesn't overwrite existing resources. You'd have to explicitly call `.insert_resource(DefaultQueryFilters::default())` which is probably verbose enough to make it obvious what will happen. Currently it is intended that you can opt-out of entity disabling, but as the feature matures we could consider making it so that it is always there.

File: crates/bevy_ecs/src/entity_disabling.rs [hunk 7]
```diff
@@ -26,7 +26,17 @@ use crate::{
...
 };
-use bevy_ecs_macros::Resource;
+use bevy_ecs_macros::{Component, Resource};
+
+#[cfg(feature = "bevy_reflect")]
+use {crate::reflect::ReflectComponent, bevy_reflect::Reflect};
+
+/// A marker component for disabled entities. See [the module docs] for more info.
+///
+/// [the module docs]: crate::entity_disabling
```
chescock, line 36: Should we update the module docs to mention that the `Disabled` component is disabled in the default `DefaultQueryFilters`?  Right now they talk about it as a hypothetical.

crates/bevy_scene/src/dynamic_scene_builder.rs
```diff
@@ -348,9 +351,17 @@ impl<'w> DynamicSceneBuilder<'w> {
     /// [`deny_resource`]: Self::deny_resource
     #[must_use]
     pub fn extract_resources(mut self) -> Self {
+        let original_world_dqf_id = self
+            .original_world
+            .components()
+            .get_resource_id(TypeId::of::<DefaultQueryFilters>());
```
chescock, line 357: It will be nice when resources-as-components lets us use default query filters to exclude the default query filters resource from our resource query :).

chescock (approved):

cBournhonesque (approved):

Issue #16883, opened 2024-12-18
Provide a mechanism for selecting a VideoMode for FullScreen Exclusive


dev-inigmas:
## What problem does this solve or what need does it fill?

I want to be able to provide users with a list of `VideoMode`s available for their monitor and allow them to select the `VideoMode` they want to use. Sensible defaults - like the ones Bevy currently employs - are good. But users may want to manually set their Resolution and Refresh Rate.

Maybe as a "streamer" the user would like to record/stream the game. Or as a "developer", the user would like to make promotional recordings. Let's say that the user's monitor supports 4k resolution and up to 120hz refresh rate depending on resolution. But they want to stream/record via OBS (or other tool) natively at 1080p 60hz while viewing the app at full screen - either for visual preference, or performance reasons.

I haven't tried it yet, but based on the API, I'm assuming that I _can_ allow the user to set the fullscreen resolution. But I don't see a way to adjust the monitor's refresh rate to a user-desired value.

## What solution would you like?

From what I've seen, maybe something like this would work.
* Add variant, `FullscreenMode(MonitorSelection, VideoMode)` to the `WindowMode` enum.
* Rather than calculating the "best" `VideoMode` to use, simply use the one provided by the enum.

## What alternative(s) have you considered?

From what I can tell based on what little source code I've seen - I think that Bevy will automatically determine the "best" refresh rate and adjust to it. So, even _if_ the user were to manually adjust their monitor's refresh rate - Bevy will bump it up to the highest available. Similarly, if the developer tried adjusting the refresh rate outside of Bevy - Bevy would reset it to the highest available.

## Additional context

Thanks again for adding `.with_scale_factor_override(f32)` and similar functions.

jf908:
I have a branch that mostly implements this which I’m considering turning into a PR soon.

It removes SizedFullscreen and changes Fullscreen to WindowMode::Fullscreen(MonitorSelection, VideoModeSelection) where
VideoModeSelection is an enum of Current or a specific VideoMode. Not sure if that would be controversial but I think it would be more intuitive and flexible than the current API.

dev-inigmas:
Personally, I think that many devs may find the `SizedFullscreen|Fullscreen` variants (as they are now) to be useful. I suspect that a PR would have less friction for acceptance if there were simply a new variant. But as an actual contributor, I'm sure you're much more aware of "how things work around here" than I am.

Either way, I'm glad I'm not the only one interested in being able to manually set the VideoMode.

[...]

PR #17525, opened 2025-01-24
Rework WindowMode::Fullscreen API

jf908:
# Objective

- #16883 
- Improve the default behaviour of the exclusive fullscreen API.

## Solution

This PR changes the exclusive fullscreen window mode to require the type `WindowMode::Fullscreen(MonitorSelection, VideoModeSelection)` and removes `WindowMode::SizedFullscreen`. This API somewhat intentionally more closely resembles Winit 0.31's upcoming fullscreen and video mode API.

The new VideoModeSelection enum is specified as follows:

```rust
pub enum VideoModeSelection {
    /// Uses the video mode that the monitor is already in.
    Current,
    /// Uses a given [`crate::monitor::VideoMode`]. A list of video modes supported by the monitor
    /// is supplied by [`crate::monitor::Monitor::video_modes`].
    Specific(VideoMode),
}
```

### Changing default behaviour

This might be contentious because it removes the previous behaviour of `WindowMode::Fullscreen` which selected the highest resolution possible. While the previous behaviour would be quite easy to re-implement as additional options, or as an impl method on Monitor, I would argue that this isn't an implementation that should be encouraged.

From the perspective of a Windows user, I prefer what the majority of modern games do when entering fullscreen which is to preserve the OS's current resolution settings, which allows exclusive fullscreen to be entered faster, and to only have it change if I manually select it in either the options of the game or the OS. The highest resolution available is not necessarily what the user prefers.

I am open to changing this if I have just missed a good use case for it. 

Likewise, the only functionality that `WindowMode::SizedFullscreen` provided was that it selected the resolution closest to the current size of the window so it was removed since this behaviour can be replicated via the new `VideoModeSelection::Specific` if necessary.

## Out of scope

WindowResolution and scale factor act strangely in exclusive fullscreen, this PR doesn't address it or regress it.

## Testing

- Tested on Windows 11 and macOS 12.7
- Linux untested

## Migration Guide

`WindowMode::SizedFullscreen(MonitorSelection)` and `WindowMode::Fullscreen(MonitorSelection)` has become `WindowMode::Fullscreen(MonitorSelection, VideoModeSelection)`. Previously, the VideoMode was selected based on the closest resolution to the current window size for SizedFullscreen and the largest resolution for Fullscreen. It is possible to replicate that behaviour by searching `Monitor::video_modes` and selecting it with `VideoModeSelection::Specific(VideoMode)` but it is recommended to use `VideoModeSelection::Current` as the default video mode when entering fullscreen.


IceSentry (approved):

crates/bevy_window/src/window.rs
```diff
@@ -1286,28 +1307,15 @@ pub enum WindowMode {
...
+    /// The window should be in "true"/"legacy"/"exclusive" Fullscreen mode on the given [`MonitorSelection`].
+    ///
+    /// The resolution, refresh rate, and bit depth are selected based on the given [`VideoModeSelection`].
     ///
     /// Note: As this mode respects the scale factor provided by the operating system,
     /// the window's logical size may be different from its physical size.
     /// If you want to avoid that behavior, you can use the [`WindowResolution::set_scale_factor_override`] function
     /// or the [`WindowResolution::with_scale_factor_override`] builder method to set the scale factor to 1.0.
-    Fullscreen(MonitorSelection),
+    Fullscreen(MonitorSelection, VideoModeSelection),
```
kristoff3r, line 1318: Maybe better to use named fields here? I usually switch to that for structs/variants with more than 1 field.

kristoff3r (approved):
This seems much less convoluted than before. I'm not sure if the old behavior is sometimes useful but in that case we can add them to the VideoModeSelection struct as needed.

crates/bevy_window/src/window.rs
IceSentry: Personally I use named fields when there are primitive types, but when it's only named types I think it's fine

jf908: I think I also prefer (MonitorSelection, VideoModeSelection) here just because `.0` being MonitorSelection matches the other window modes and it also also matches the Fullscreen type from (future) winit.

jf908:
I suppose this is just the spurious wasm test and I don't have to change anything? 😄

alice-i-cecile:
Yep, nothing needed on your end. Trying again!

Issue #13682, opened 2024-06-05
Stochastic texture sampling


torsteingrindvik:
## What problem does this solve or what need does it fill?

I want to easily make tiled textures look non-tiled.

## What solution would you like?

I'd like Bevy support where as a user I have some tunable settings that allow me to adapt my textures to look non-tiled.

## What alternative(s) have you considered?

Using tools such as krita, photoshop etc. and following tutorials to make textures look non-tiled, or rolling my own shader.

## Additional context

[This reddit post](https://www.reddit.com/r/Unity3D/comments/dhr5g2/i_made_a_stochastic_texture_sampling_shader/)
has a great video, and they also show their shader code, usage example, link to paper its based on etc. and might be a good starting point (either for development or further research)

torsteingrindvik:
https://discordapp.com/channels/691052431525675048/743663924229963868/1247968409522802850

> ...it should be an extended material. We should probably have a discussion about how we want to handle this kind of thing

> The problem being this: StandardMaterial is going to be too big if we keep adding things to it without limit. My proposal is to set the limit at the intersection of glTF + ratified KHR extensions and Principled BSDF in Blender. That means basically we add sheen and then we're done. (edited)
[7:42 PM]
But there's obviously room for a lot more features and the question is what to do with them. The ideal would be to use MaterialX and allow the user to create their own ubershader with a node graph. But that's a lot of work and I think we would want a shorter term solution.
[7:47 PM]
My current thought is to add documented helper WGSL functions that allow users to create their own ubershaders, and then some sort of convenience abstraction for common usages. Perhaps decals, for example, could have a built-in ExtendedMaterial that allows for albedo + normal map decals on the StandardMaterial. Likewise triplanar mapping could have a built-in ExtendedMaterial that does blending on a StandardMaterial. Etc

@pcwalton adding your comments from discord here since they might be useful context in the future.

pcwalton:
Also see Inigo Quilez' articles on this https://iquilezles.org/articles/texturerepetition/

ThomasCartier:
I would like to avoid writing the shaders manually. Not skilled enough for that.
What is the current path that is the easiest and compatible with wasm?

JMS55:
Not sure if it's related or not, but there's this very cool recent paper on stochastic texture filtering, with accompanying shader toys https://research.nvidia.com/publication/2024-05_filtering-after-shading-stochastic-texture-filtering.

PR #17527, opened 2025-01-24
Add an example for stochastic texture sampling

cBournhonesque:
# Objective

This PR adds an example that showcases how a custom material can be used to repeat a texture stochastically while avoiding repetition artifacts.
It is related to https://github.com/bevyengine/bevy/issues/13682. 

It doesn't provide a seamless solution for users but at least it can provide an example.

I don't think it's the best implementation out there (it's my first time writing a shader and I mostly glued external code) but it seems to be working!

https://github.com/user-attachments/assets/9c2d9a50-a317-4354-9470-629e25c2ad38


IceSentry:
I personally don't like that kind of example because it doesn't really show any bevy api that isn't already documented in another example. I think for an example like that to be included it would need to show something that requires a lot of complexe use of bevy apis, but this just shows a simple use of the mesh and material api. Most of the complexity is in the shader. If people want it despite that I won't push back more, but I just wanted to voice this.

cBournhonesque:
> I personally don't like that kind of example because it doesn't really show any bevy api that isn't already documented in another example. I think for an example like that to be included it would need to show something that requires a lot of complexe use of bevy apis, but this just shows a simple use of the mesh and material api. Most of the complexity is in the shader. If people want it despite that I won't push back more, but I just wanted to voice this.

I think I agree. Would we want to support this kind of behaviour in-tree or let external crates provide this kind of stuff?

Mantissa-23:
> I personally don't like that kind of example because it doesn't really show any bevy api that isn't already documented in another example. I think for an example like that to be included it would need to show something that requires a lot of complexe use of bevy apis, but this just shows a simple use of the mesh and material api. Most of the complexity is in the shader. If people want it despite that I won't push back more, but I just wanted to voice this.

I think the broader question here is "is it the job of Bevy documentation to teach people _game development_ in addition to _how to use Bevy_?"

I would argue the answer is "yes," due in part to Bevy's popularity, but perhaps it's valuable to introduce some kind of organizational structure here, or provide this kind of thing in a "recipe book," as opposed to an example that directly demonstrates use of the engine.. Because I agree, this is the kind of content that is Bevy-agnostic and could apply to pretty much any game engine.

IceSentry:
> I would argue the answer is "yes,"

I personally disagree, but that's only my personal opinion. The bevy community has decided to start merging tutorials that are aimed at more general game development in the context of bevy and not just teaching bevy itself. https://github.com/bevyengine/bevy-website/pull/2131

PR #17528, opened 2025-01-24
expose OverflowAxis::Hidden as Overflow functions

LucDrenth:
# Objective
expose `OverflowAxis::Hidden` as functions of `Overflow`, just as it is done for `OverflowAxis::Hidden` and `OverflowAxis::Scroll`.

### DIFF
--- a/crates/bevy_ui/src/ui_node.rs
+++ b/crates/bevy_ui/src/ui_node.rs
@@ -1064,6 +1064,30 @@ impl Overflow {
         }
     }
 
+    /// Hide overflowing items on both axes
+    pub const fn hidden() -> Self {
+        Self {
+            x: OverflowAxis::Hidden,
+            y: OverflowAxis::Hidden,
+        }
+    }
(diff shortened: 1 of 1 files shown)


mnmaita (approved):

crates/bevy_ui/src/ui_node.rs
```diff
@@ -1064,6 +1064,30 @@ impl Overflow {
         }
     }
 
+    /// Hide overflowing items on both axes
```
ickshonpe, line 1067: The comments needs to explain that hidden sections nodes are non-visible but still considered part of the layout (unlike clipped sections which are ignored completely).

ickshonpe (changes requested):
Doc comments need a little more just explaining why hidden is different to clipped.
Otherwise looks good.

ickshonpe (approved):

BenjaminBrienen (approved):

Issue #16066, opened 2024-10-23
Update function runs once after closing the window


beholdnec:
## Bevy version

9930df83ed42008f7eb2c02cc7350040f0250c2e

## Relevant system information

Windows 11

## What you did

I experimented with a few of the example apps and noticed that the Update function runs once after I close the window.

## What went wrong

My understanding is that closing the window should tear down Bevy without calling Update any additional times. However, it seems that the Update schedule is run once after exiting.

This can be demonstrated with most of the example apps. For instance, in the camera/camera_orbit.rs file, try adding a println to its Update function, "orbit":

```rust
fn orbit(
    mut camera: Single<&mut Transform, With<Camera>>,
    camera_settings: Res<CameraSettings>,
    mouse_buttons: Res<ButtonInput<MouseButton>>,
    mouse_motion: Res<AccumulatedMouseMotion>,
    time: Res<Time>,
) {
    println!("running update...");
    let delta = mouse_motion.delta;
    let mut delta_roll = 0.0;
```

This will result in the following output upon exiting the app:

```
running update...
running update...
running update...
running update...
running update...
running update...
2024-10-23T02:50:50.183869Z  INFO bevy_window::system: No windows are open, exiting
2024-10-23T02:50:50.186436Z  INFO bevy_winit::system: Closing window 0v1#4294967296
running update...
2024-10-23T02:50:50.192336Z  INFO bevy_window::system: No windows are open, exiting
```

As you can see, the update is called after closing the window, even though bevy_window claims it is exiting.

beholdnec:
Understood. I will close this issue.

PR #17533, opened 2025-01-25
Fix window close in example cause panic

### LINKED ISSUE #17532: In example parallel_query close window cause panic
## Bevy version

[0.15.1](https://github.com/bevyengine/bevy/releases/tag/v0.15.1)

```PostScipt
2025-01-25T08:15:34.071475Z  INFO bevy_diagnostic::system_information_diagnostics_plugin::internal: SystemInfo { os: "Windows 11 Home", kernel: "26100", cpu: "AMD Ryzen 7 5800 8-Core Processor", core_count: "8", memory: "15.9 GiB" }
2025-01-25T08:15:34.508783Z  INFO bevy_render::renderer: AdapterInfo { name: "NVIDIA GeForce RTX 3060 Ti", vendor: 4318, device: 9353, device_type: DiscreteGpu, driver: "NVIDIA", driver_info: "566.36", backend: Vulkan }
```

## What you did

In example parallel_query build&run example then close window cause panic. I have tested both on windows and macos. I found this issue will happen on windows.

this panic is caused by window resource has been recycle but bounce_system is still trying to get and then cause panic.
```rust
// Bounce sprites outside the window
fn bounce_system(window: Single<&Window>, mut sprites: Query<(&Transform, &mut Velocity)>) {
    let width = window.width();
    let height = window.height();
    let left = width / -2.0;
    let right = width / 2.0;
    let bottom = height / -2.0;
    let top = height / 2.0;
    // The default batch size can also be overridden.
    // In this case a batch size of 32 is chosen to limit the overhead of
    // ParallelIterator, since negating a vector is very inexpensive.
    sprites
        .par_iter_mut()
        .batching_strategy(BatchingStrategy::fixed(32))
        .for_each(|(transform, mut v)| {
            if !(left < transform.translation.x
                && transform.translation.x < right
                && bottom < transform.translation.y
                && transform.translation.y < top)
            {
                // For simplicity, just reverse the velocity; don't use realistic bounces
                v.0 = -v.0;
            }
        });
}
```

Fixed in #17533 

so I'm not sure this is a proper solution. but unregister all system first then recycle window maybe not easy on all platfoms

jiangheng90:
# Objective

Fixes #17532 

## Solution

- check window valide

### DIFF
--- a/examples/2d/2d_viewport_to_world.rs
+++ b/examples/2d/2d_viewport_to_world.rs
@@ -12,9 +12,13 @@ fn main() {
 
 fn draw_cursor(
     camera_query: Single<(&Camera, &GlobalTransform)>,
-    window: Single<&Window>,
+    window: Query<&Window>,
     mut gizmos: Gizmos,
 ) {
+    let Ok(window) = window.get_single() else {
+        return;
+    };
+
     let (camera, camera_transform) = *camera_query;
 
     let Some(cursor_position) = window.cursor_position() else {
--- a/examples/3d/3d_viewport_to_world.rs
+++ b/examples/3d/3d_viewport_to_world.rs
@@ -13,9 +13,13 @@ fn main() {
 fn draw_cursor(
     camera_query: Single<(&Camera, &GlobalTransform)>,
     ground: Single<&GlobalTransform, With<Ground>>,
-    windows: Single<&Window>,
+    windows: Query<&Window>,
     mut gizmos: Gizmos,
 ) {
+    let Ok(windows) = windows.get_single() else {
+        return;
+    };
+
     let (camera, camera_transform) = *camera_query;
 
     let Some(cursor_position) = windows.cursor_position() else {
--- a/examples/ecs/observers.rs
+++ b/examples/ecs/observers.rs
@@ -172,9 +172,13 @@ fn draw_shapes(mut gizmos: Gizmos, mines: Query<&Mine>) {
 fn handle_click(
     mouse_button_input: Res<ButtonInput<MouseButton>>,
     camera: Single<(&Camera, &GlobalTransform)>,
-    windows: Single<&Window>,
+    windows: Query<&Window>,
     mut commands: Commands,
 ) {
+    let Ok(windows) = windows.get_single() else {
+        return;
+    };
+
     let (camera, camera_transform) = *camera;
     if let Some(pos) = windows
         .cursor_position()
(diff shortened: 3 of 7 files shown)


Selene-Amanita:
While your PR fixes the issue from the point of view of the example, I feel like there's something wrong regarding how Bevy handles this that should probably be fixed.

In order, during a `Main` schedule/frame, Bevy:
1. in `Update`, will check if all windows have been closed (aka no entity with a Window component) with `exit_on_all_closed` and send an `AppExit` event if that's the case https://docs.rs/bevy/latest/bevy/window/fn.exit_on_all_closed.html
2. in `Last`, will despawn Window entities corresponding to a closed Window with `despawn_windows` https://github.com/bevyengine/bevy/blob/deb135c25c3a7eafdc432eaefe0957d6ffdcca62/crates/bevy_winit/src/lib.rs#L134 / https://github.com/bevyengine/bevy/blob/deb135c25c3a7eafdc432eaefe0957d6ffdcca62/crates/bevy_winit/src/system.rs#L205
3. at the end of the schedule in `bevy_winit` will check if an `AppExit` even has been sent to stop the app

I feel like step 1 should be done between 2 and 3, to avoid having a frame where the Window entity is despawned despite Bevy being configured to exit when that happens. So `exit_on_all_closed` (and `exit_on_primary_closed`) should be ordered in `Last` after `despawn_windows` (and a flush of the despawns). But maybe I'm missing something? Maybe it's not done that way because it messes with parallelization in the `Last` schedule, or maybe Bevy does expect to have one frame to deal with the no-window scenario for whatever reason?

Selene-Amanita (approved):
Apparently the order in which Bevy does thing is to allow people to write a system to react to an AppExit event in Update. I'm not sure this is right, but as long as it stays that way and Single panics, systems should probably never expect that there will always be a Window, and so the `Single<&Window>` should be reverted back. So, approving this PR. Might be needed for other examples too.

beholdnec:
I think this is the same issue I reported here: https://github.com/bevyengine/bevy/issues/16066

BenjaminBrienen:
This doesn't seem right to me. The problem at hand is a general one. We shouldn't have to validate parameters like this.

jiangheng90:
I found in previous version 0.14.0 the example:
```rust
fn bounce_system(windows: Query<&Window>, mut sprites: Query<(&Transform, &mut Velocity)>) {
    let window = windows.single();
```
and in this version the window close will not cause panic.
then I use this api to now v0.15.1 the panic is still there
I think I need a bisect....

jiangheng90:
@Selene-Amanita @BenjaminBrienen I have fix this problem in #17543

BenjaminBrienen:
Closing in favor of the other PR

mockersf (approved):

jiangheng90:
I just have changed ```Single<&Window>```,  which will call in update and cause panic. not all of it.

PR #17538, opened 2025-01-25
added Hash to MouseScrollUnit;

AustinHellerRepo:
# Objective

This allows for the usage of the MouseScrollUnit as a key to a HashSet and HashMap. I have a need for this, but this basic functionality is currently missing.

## Solution

Add the derive Hash attribute to the MouseScrollUnit type.

## Testing

No, but I did perform a `cargo build`. My laptop is failing to run `cargo test` without crashing.
If someone could run a `cargo test` for completeness, that would be great but this is a trivial change.
They simply need to ensure that the common Hash derive macro works as expected for the basic MouseScrollUnit type.
Ubuntu 22.04

### DIFF
--- a/crates/bevy_input/src/mouse.rs
+++ b/crates/bevy_input/src/mouse.rs
@@ -105,7 +105,7 @@ pub struct MouseMotion {
 ///
 /// The value of the event can either be interpreted as the amount of lines or the amount of pixels
 /// to scroll.
-#[derive(Debug, Clone, Copy, Eq, PartialEq)]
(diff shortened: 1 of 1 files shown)


AustinHellerRepo:
I wanted to keep track of vertical and horizontal deltas per unit type in a HashMap, for when both types of scroll units are found at the same time. I wrote a work-around, but it would be nice to refactor it after this PR is merged in.

BenjaminBrienen (approved):

alice-i-cecile (approved):
Thanks for the justification :) That's the primary thing I look for when reviewing these PRs.

Issue #17276, opened 2025-01-10
CustomCursor::Image could use a custom size option


marioferpa:
## What problem does this solve or what need does it fill?

The new CustomCursor::Image solution in Bevy 0.15 is a great addition, making it simple to replace the cursor with an image instead of hiding the original cursor and making a Sprite follow its position.

```
commands
  .entity(window_entity)
  .insert(
      CursorIcon::Custom(
          CustomCursor::Image {
              handle:     asset_server.load("data/vanilla/sprites/kenney_hand.png").into(),
              hotspot:    (0, 0)
          }
      )
  )
;
```

An advantage that the previous method had, however, was complete control about the image size, rotation, etc. In my current project, as an example, sprites are not represented with their original resolution, making the cursor look too small in comparison.

## What solution would you like?

An extra field in CustomCursor::Image allowing for the sprite to be stretched or shrinked. 

## What alternative(s) have you considered?

A quick solution would be to modify the source file to make it bigger. This would mean however that using a different tileset with a different resolution would require making a new cursor as well, so their relative sizes keep being the same. 

If one wanted to animate the cursor (making it grow when approaching something, or making it grow and shrink rhythmically) it would require making a cursor spritesheet, when changing the size value would be simpler.

mgi388:
I started looking at this and it got me wondering about the best behavior for `hotspot`s when users provide image manipulation settings like `flip_x|y` and now `custom_size`. Trying to work out if users would expect Bevy to adjust their provided hotspot based on the image transforms, or if users would expect to also provide their transformed hotspot. Example:

```rust
CursorIcon::Custom(CustomCursor::Image {
    handle: asset_server.load("rotate-cursor.png"), // source is 32x32
    texture_atlas: None,
    flip_x: true,
    flip_y: false,
    rect: None,
    hotspot: (15, 14), // should users pass the hotspot that matches the source or the transformed image?
    custom_size: Some((64, 64).into()), // x2
})
```

Also note that the cursor has `flip_y`, and so the `hotspot.y` field would also need to change.

Should users provided the final `hotspot` value, or should Bevy mutate that based on image transforms it did?

Another alternative is to not add `custom_size` but instead add `scale: Option<Vec2>`. The above would change to this:

```rust
CursorIcon::Custom(CustomCursor::Image {
    handle: asset_server.load("rotate-cursor.png"), // source is 32x32
    texture_atlas: None,
    flip_x: true,
    flip_y: false,
    rect: None,
[...]

PR #17540, opened 2025-01-26
Automatically transform cursor hotspot user asks to flip cursor image

mgi388:
# Objective

- As discussed in https://github.com/bevyengine/bevy/issues/17276#issuecomment-2611203714, we should transform the cursor's hotspot if the user is asking for the image to be flipped.
- This becomes more important when a `scale` transform option exists. It's harder for users to transform the hotspot themselves when using `scale` because they'd need to look up the image to get its dimensions. Instead, we let Bevy handle the hotspot transforms and make the `hotspot` field the "original/source" hotspot.
- Refs #17276.

## Solution

- When the image needs to be transformed, also transform the hotspot. If the image does not need to be transformed (i.e. fast path), no hotspot transformation is applied.

## Testing

- Ran the example: `cargo run --example custom_cursor_image --features=custom_cursor`.
- Add unit tests for the hotspot transform function.
- I also ran the example I have in my `bevy_cursor_kit` crate, which I think is a good illustration of the reason for this PR.
- In the following videos, there is an arrow pointing up. The button hover event fires as I move the mouse over it.
- When I press `Y`, the cursor flips. 
- In the first video, on `bevy@main` **before** this PR, notice how the hotspot is wrong after flipping and no longer hovering the button. The arrow head and hotspot are no longer synced.
- In the second video, on the branch of **this** PR, notice how the hotspot gets flipped as soon as I press `Y` and the cursor arrow head is in the correct position on the screen and still hovering the button. Speaking back to the objective listed at the start: The user originally defined the _source_ hotspot for the arrow. Later, they decide they want to flip the cursor vertically: It's nice that Bevy can automatically flip the _source_ hotspot for them at the same time it flips the _source_ image.

First video (main):

https://github.com/user-attachments/assets/1955048c-2f85-4951-bfd6-f0e7cfef0cf8

Second video (this PR):

https://github.com/user-attachments/assets/73cb9095-ecb5-4bfd-af5b-9f772e92bd16

### DIFF
--- a/crates/bevy_winit/src/custom_cursor.rs
+++ b/crates/bevy_winit/src/custom_cursor.rs
@@ -18,15 +18,13 @@ pub struct CustomCursorImage {
     pub texture_atlas: Option<TextureAtlas>,
     /// Whether the image should be flipped along its x-axis.
     ///
-    /// If true, the cursor's hotspot will be automatically flipped along with
-    /// the image. You don't need to adjust the `hotspot` field to account for
-    /// the flip.
+    /// If true, the cursor's `hotspot` automatically flips along with the
+    /// image.
     pub flip_x: bool,
     /// Whether the image should be flipped along its y-axis.
     ///
-    /// If true, the cursor's hotspot will be automatically flipped along with
-    /// the image. You don't need to adjust the `hotspot` field to account for
-    /// the flip.
+    /// If true, the cursor's `hotspot` automatically flips along with the
+    /// image.
     pub flip_y: bool,
     /// An optional rectangle representing the region of the image to render,
     /// instead of rendering the full image. This is an easy one-off alternative
@@ -37,6 +35,10 @@ pub struct CustomCursorImage {
     pub rect: Option<URect>,
     /// X and Y coordinates of the hotspot in pixels. The hotspot must be within
     /// the image bounds.
+    ///
+    /// If you are flipping the image using `flip_x` or `flip_y`, you don't need
+    /// to adjust this field to account for the flip because it is adjusted
+    /// automatically.
     pub hotspot: (u16, u16),
 }
 
--- a/crates/bevy_winit/src/custom_cursor.rs
+++ b/crates/bevy_winit/src/custom_cursor.rs
@@ -17,8 +17,16 @@ pub struct CustomCursorImage {
     /// An optional texture atlas used to render the image.
     pub texture_atlas: Option<TextureAtlas>,
     /// Whether the image should be flipped along its x-axis.
+    ///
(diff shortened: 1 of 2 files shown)


mgi388:
@eero-lehtinen 🙏

mgi388:
@BenjaminBrienen looks like cursors / this one's `A-Windowing` FYI :) https://github.com/bevyengine/bevy/pulls?q=is%3Apr+author%3Amgi388+is%3Aclosed

eero-lehtinen (approved):

eero-lehtinen:
Thanks for the quality PRs 👍

crates/bevy_winit/src/custom_cursor.rs
```diff
@@ -17,8 +17,16 @@ pub struct CustomCursorImage {
...
+    ///
+    /// If true, the cursor's hotspot will be automatically flipped along with
+    /// the image. You don't need to adjust the `hotspot` field to account for
+    /// the flip.
     pub flip_x: bool,
     /// Whether the image should be flipped along its y-axis.
+    ///
+    /// If true, the cursor's hotspot will be automatically flipped along with
+    /// the image. You don't need to adjust the `hotspot` field to account for
+    /// the flip.
```
Bleachfuel, line 29:
```suggestion
    ///
    /// The `hotspot` will flip alongside the flipped image
```

I feel like the description has too many unnessecary words

crates/bevy_winit/src/custom_cursor.rs
```diff
@@ -17,8 +17,16 @@ pub struct CustomCursorImage {
     /// An optional texture atlas used to render the image.
     pub texture_atlas: Option<TextureAtlas>,
     /// Whether the image should be flipped along its x-axis.
+    ///
+    /// If true, the cursor's hotspot will be automatically flipped along with
+    /// the image. You don't need to adjust the `hotspot` field to account for
+    /// the flip.
```
Bleachfuel, line 23:
```suggestion
    ///
    /// The `hotspot` will flip alongside the flipped image
```

Bleachfuel (changes requested):

crates/bevy_winit/src/custom_cursor.rs
BenjaminBrienen: Missing punctuation at the end of the suggestion

crates/bevy_winit/src/custom_cursor.rs
BenjaminBrienen: Missing punctuation

Bleachfuel (approved):

PR #14155, opened 2024-07-05
Dirty fix for App hanging when windows are invisible on WindowsOS

MiniaczQ:
# Objective

- Fixes #14135 

## Solution

- If no windows are visible, app updates will run regardless of redraw call result.

This a relatively dirty fix, a more robust solution is desired in the long run:
https://github.com/bevyengine/bevy/issues/1343#issuecomment-770091684
https://discord.com/channels/691052431525675048/1253771396832821270/1258805997011730472
The solution would disconnect rendering from app updates.

## Testing

- `window_settings` now works

## Other platforms

Not a problem on Linux: https://discord.com/channels/691052431525675048/692572690833473578/1259526650622640160
Not a problem on MacOS: https://discord.com/channels/691052431525675048/692572690833473578/1259563986148659272

Issue #14135, opened 2024-07-04
visible: false on WindowPlugin Stops Rendering


igorlp:
## Bevy version
0.14

## Relevant system information
```ignore
`SystemInfo { os: "Windows 11 Home", kernel: "22635", cpu: "AMD Ryzen 7 5800X3D 8-Core Processor", core_count: "8", memory: "31.9 GiB" }`
`AdapterInfo { name: "NVIDIA GeForce RTX 4060 Ti", vendor: 4318, device: 10243, device_type: DiscreteGpu, driver: "", driver_info: "", backend: Dx12 }`
```

## What you did

Set visible: false on WindowPlugin

## What went wrong

Systems added to "Update" no longer trigger.

Using:
```
fn setup_window(mut window: Query<&mut Window>, frames: Res<FrameCount>) {
    println!("{}", frames.0);
    if frames.0 == 3 {
        window.single_mut().visible = true;
    }
}
```

If visible is set to false by default, we only see `0` in the console, once. If set to true, we see the frame count as expected for every frame.

## Additional information
Also happens on Bevy example found [here](https://bevyengine.org/examples-webgpu/window/window-settings/)

mockersf:
> Also happens on Bevy example found [here](https://bevyengine.org/examples-webgpu/window/window-settings/)

How do you see that issue in Wasm? It's supposed to stay black...

Themayu:
Issue does not appear on the WASM example; the `make_visible` system runs fine, and all other `Update` systems in the example function normally.

Perhaps they were trying to say that it appears if you run the example locally?

alice-i-cecile:
Can you try just before and after https://github.com/bevyengine/bevy/pull/14023 please?

igorlp:
> Perhaps they were trying to say that it appears if you run the example locally?

Correct!

> Can you try just before and after https://github.com/bevyengine/bevy/pull/14023 please?

Works with:
```
bevy = { version = "=0.14.0-rc.4" }
bevy_winit = { version = "=0.14.0-rc.4" }
```

Does not work with:
```
bevy = { version = "0.14.0" }
bevy_winit = { version = "0.14.0" }
```

janhohenheim:
Little update: this issue can *not* be reproduced on the current `main` on Intel macOS and Linux (with or without wayland).
It really does seem to be Windows-specific.

PR #17543, opened 2025-01-26
Fix update execute while app exit

jiangheng90:
# Objective

Fixes #17533  #16066 

update execute while app is exit, if we query object in update, it will cause panic becuse these object has been recycle

this problem is caused from #14155, by fix  App hanging in example window_settings. when windows all destroyed to be zero the all_invisible  will be true, then cause update execute after app exit.

## Solution

before update check app state is exit. if exit no execute.

another choice is to see the number of windows if it goes to zero, just make all_invisible to be false

but I think check app state is more reliable, because maybe in some case we need app to run background with no window. 

for now, I just check if app is exit.

### DIFF
--- a/crates/bevy_winit/src/state.rs
+++ b/crates/bevy_winit/src/state.rs
@@ -571,12 +571,17 @@ impl<T: Event> ApplicationHandler<T> for WinitAppRunnerState<T> {
             // If no windows exist, this will evaluate to `true`.
             let all_invisible = windows.iter().all(|w| !w.1.visible);
 
+            let app_is_exit = matches!(self.app_exit, Some(AppExit::Success));
+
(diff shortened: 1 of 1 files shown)


BenjaminBrienen (approved):

mockersf:
we want to be able to run update when the app is exiting, so this is not a good fix

jiangheng90:
> we want to be able to run update when the app is exiting, so this is not a good fix

@mockersf  I see that the dirty fix of https://github.com/bevyengine/bevy/issues/14135 is just a temporary solution. 

I'm a new guy learning bevy and rust. this issue is not that important for me.  I just found this issue and try to fix it. 

so what is your suggestion for this issue. just leave it or you have some idea maybe I can try ?

jiangheng90:
close in favor of  #17533

PR #17545, opened 2025-01-26
feat: impl `Ease` for `Isometry[2/3]d`

### LINKED ISSUE #17539: Implement the trait `Ease` for `Isometry3d`

Being able to do something like:
```
let isometry_curve = EasingCurve::new(
            Isometry3d::new(Vec3::ZERO, Quat::IDENTITY),
            Isometry3d::new(Vec3::ZERO, Quat::IDENTITY),
            EaseFunction::CubicInOut,
        )
let isometry = isometry_curve.sample(t).unwrap();
```

Making `Isometry3d` struct implement `Ease`. (Not sure what else is needed to make the above code work?)

Creating a curve singularly for both `Vec3` and `Quat`.

RobWalt:
# Objective

- We kind of missed out on implementing the `Ease` trait for some objects like `Isometry2D` and `Isometry3D` even though it makes sense and isn't that hard
- Fixes #17539

## Testing

- wrote some minimal tests
- ~~noticed that quat easing isn't working as expected yet~~ I just confused degrees and radians once again 🙈

### DIFF
--- a/crates/bevy_math/src/curve/easing.rs
+++ b/crates/bevy_math/src/curve/easing.rs
@@ -770,10 +770,23 @@ mod tests {
             quat_curve.sample(0.0).unwrap(),
             Quat::from_axis_angle(Vec3::Z, 0.0)
         );
-        assert_abs_diff_eq!(
-            quat_curve.sample(0.5).unwrap(),
-            Quat::from_axis_angle(Vec3::Z, 45.0)
-        );
+        // {
+        //     let (before_mid_axis, before_mid_angle) =
+        //         quat_curve.sample(0.25).unwrap().to_axis_angle();
+        //     assert_abs_diff_eq!(before_mid_axis, Vec3::Z);
+        //     assert_abs_diff_eq!(before_mid_angle.to_degrees(), 22.5);
+        // }
+        {
+            // doesnt work ?!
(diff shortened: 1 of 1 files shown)


Carter0 (approved):

crates/bevy_math/src/curve/easing.rs
```diff
@@ -719,4 +758,113 @@ mod tests {
             );
         }
     }
+
+    #[test]
+    fn ease_quats() {
+        let quat_start = Quat::from_axis_angle(Vec3::Z, 0.0);
+        let quat_end = Quat::from_axis_angle(Vec3::Z, 90.0_f32.to_radians());
```
greeble-dev, line 765: Suggestion: instead of each case giving the angle in degrees, maybe define one angle and multiply it by the curve parameter per case?

```rust
let angle = 90.0_f32.to_radians();
let quat_end = Quat::from_axis_angle(Vec3::Z, angle);
...
assert_abs_diff_eq!(quat_curve.sample(0.5).unwrap(), Quat::from_axis_angle(Vec3::Z, 0.5 * angle));
```

greeble-dev (approved):
Approved with one minor suggestion.

crates/bevy_math/src/curve/easing.rs
RobWalt: Good point, will do!

mockersf (approved):

PR #17547, opened 2025-01-26
impl Eq + Hash for BindGroup/Layout

djeedai:
# Objective

Implement `Eq` and `Hash` for the `BindGroup` and `BindGroupLayout` wrappers.

## Solution

Implement based on the same assumption that the ID is unique, for consistency with `PartialEq`.

## Testing

None; this should be straightforward. If there's an issue that would be a design one.

### DIFF
--- a/crates/bevy_render/src/render_resource/bind_group_layout.rs
+++ b/crates/bevy_render/src/render_resource/bind_group_layout.rs
@@ -27,8 +27,8 @@ impl PartialEq for BindGroupLayout {
 
 impl Eq for BindGroupLayout {}
 
-impl std::hash::Hash for BindGroupLayout {
-    fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
+impl core::hash::Hash for BindGroupLayout {
+    fn hash<H: core::hash::Hasher>(&self, state: &mut H) {
         self.id.0.hash(state);
     }
 }
--- a/crates/bevy_render/src/render_resource/bind_group_layout.rs
+++ b/crates/bevy_render/src/render_resource/bind_group_layout.rs
@@ -5,6 +5,14 @@ use core::ops::Deref;
 
 define_atomic_id!(BindGroupLayoutId);
 
+/// Bind group layouts define the interface of resources (e.g. buffers, textures, samplers)
+/// for a shader. The actual resource binding is done via a [`BindGroup`](super::BindGroup).
+///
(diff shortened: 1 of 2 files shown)


crates/bevy_render/src/render_resource/bind_group_layout.rs
```diff
@@ -5,6 +5,14 @@ use core::ops::Deref;
 
 define_atomic_id!(BindGroupLayoutId);
 
+/// Bind group layouts define the interface of resources (e.g. buffers, textures, samplers)
```
IceSentry, line 8: Is a layout really an interface?

I think I prefer how the webgpu docs phrases that "defines the structure and purpose of related GPU resources such as buffers that will be used in a pipeline" https://developer.mozilla.org/en-US/docs/Web/API/GPUBindGroupLayout

IceSentry (approved):

crates/bevy_render/src/render_resource/bind_group_layout.rs
djeedai: I'm fine either way, but yes, the bind group layout defines the resource interface of the shader. I've already seen that terminology in several places, _e.g._:

https://www.w3.org/TR/webgpu/#gpubindgrouplayout

> A [GPUBindGroupLayout](https://www.w3.org/TR/webgpu/#gpubindgrouplayout) defines the interface between a set of resources bound in a [GPUBindGroup](https://www.w3.org/TR/webgpu/#gpubindgroup) and their accessibility in shader stages.

IceSentry: I think the part that confused me is the "interface of resources" part. The interface is for the set of resources and how they are laid out in the bind group. It's not for the resources themselves and the current wording made me think of that.

djeedai: "...the shader interface to some GPU resources" ?

IceSentry: Something like "... the interface for the set of GPU resources (e.g. buffers, textures, samplers) used by the shader" maybe?

I kinda want to say to just use the webgpu spec considering that's what it is anyway.

But this is just me nitpicking because I was initially confused by it and is not at all a blocker.

pcwalton (approved):
Seems fine.

PR #17550, opened 2025-01-26
`Image::get_color_at` and `Image::set_color_at`: Support 16-bit float values

EmbersArc:
# Objective

- Also support `f16` values when getting and setting colors.

## Solution

- Use the `half` crate to work with `f16` until it's in stable Rust.


alice-i-cecile:
Neat! Do GPUs natively support f16 then? I've seen them used for machine learning, but never for rendering before.

alice-i-cecile (approved):
I like the implementation, and I'm curious to see what people do with this.

EmbersArc:
> Do GPUs natively support f16 then?

They do! I'm using them to store height and normal values for terrain textures. Those don't require as much precision and only having to store half as much data compared to `f32` is vital because those textures are _big_.

alice-i-cecile:
Oh fancy :D Right, and this sort of thing is also really handy for weird compute shaders that are passing in "images" 😂

alice-i-cecile:
Note: the Contentious label is just because this pulls in a new dependency.

mate-h (approved):
My only concern was introducing `half` would hinder the `no_std` support, but after checking the official crates.io page for `half` they cited:
> This crate provides [no_std](https://rust-embedded.github.io/book/intro/no-std.html) support so can easily be used in embedded code where a smaller float format is most useful.

Everything in this PR looks good to me, seems consistent with how the other texture formats are handled.

crates/bevy_image/src/image.rs
```diff
@@ -1400,6 +1420,14 @@ impl Image {
                 let r = (r as f64 * u32::MAX as f64) as u32;
                 bytes[0..4].copy_from_slice(&u32::to_le_bytes(r));
             }
+            TextureFormat::R16Float => {
+                // Convert to grayscale with minimal loss if color is already gray
+                let linear = LinearRgba::from(color);
+                let luminance = Xyza::from(linear).y;
+                let [r, _, _, _] = LinearRgba::gray(luminance).to_f32_array();
```
DGriffin91, line 1427: Why is this step needed? Couldn't luminance be used directly? Users may also expect R to be used here, but I can see why you chose G.

DGriffin91: Oh I see it was already done this way. ..I wonder why

PR #17555, opened 2025-01-27
Atmosphere LUT parameterization improvements

mate-h:
# Objective

- Fix the atmosphere LUT parameterization in the aerial -view and sky-view LUTs
- Correct the light accumulation according to a ray-marched reference
- Avoid negative values of the sun disk illuminance when the sun disk is below the horizon

## Solution

- Adding a Newton's method iteration to `fast_sqrt` function
- Switched to using `fast_acos_4` for better precision of the sun angle towards the horizon (view mu angle = 0)
- Simplified the function for mapping to and from the Sky View UV coordinates by removing an if statement and correctly apply the method proposed by the [Hillarie paper](https://sebh.github.io/publications/egsr2020.pdf) detailed in section 5.3 and 5.4.
- Replaced the `ray_dir_ws.y` term with a shadow factor in the `sample_sun_illuminance` function that correctly approximates the sun disk occluded by the earth from any view point

## Testing

- Ran the atmosphere and SSAO examples to make sure the shaders still compile and run as expected.

---

## Showcase

showcase-img

### DIFF
--- a/crates/bevy_pbr/src/atmosphere/aerial_view_lut.wgsl
+++ b/crates/bevy_pbr/src/atmosphere/aerial_view_lut.wgsl
@@ -32,7 +32,10 @@ fn main(@builtin(global_invocation_id) idx: vec3<u32>) {
 
     for (var slice_i: u32 = 0; slice_i < settings.aerial_view_lut_size.z; slice_i++) {
         for (var step_i: u32 = 0; step_i < settings.aerial_view_lut_samples; step_i++) {
-            let t_i = t_max * (f32(slice_i) + ((f32(step_i) + MIDPOINT_RATIO) / f32(settings.aerial_view_lut_samples))) / f32(settings.aerial_view_lut_size.z);
+            // Offset by -0.5 to align sampling position with slice boundaries, 
+            // since each texel stores the integral over its entire slice
+            var t_i = t_max * (f32(slice_i) - 0.5 + ((f32(step_i) + MIDPOINT_RATIO) / f32(settings.aerial_view_lut_samples))) / f32(settings.aerial_view_lut_size.z);
+            t_i = max(t_i, 0.0);
             let dt = (t_i - prev_t);
             prev_t = t_i;
 
--- a/crates/bevy_pbr/src/atmosphere/aerial_view_lut.wgsl
+++ b/crates/bevy_pbr/src/atmosphere/aerial_view_lut.wgsl
@@ -55,8 +55,11 @@ fn main(@builtin(global_invocation_id) idx: vec3<u32>) {
                 break;
             }
         }
-        //We only have one channel to store transmittance, so we store the mean
+        // We only have one channel to store transmittance, so we store the mean
         let mean_transmittance = (throughput.r + throughput.g + throughput.b) / 3.0;
-        textureStore(aerial_view_lut_out, vec3(vec2<u32>(idx.xy), slice_i), vec4(total_inscattering, mean_transmittance));
+
+        // Store the optical depth so that it can be linearly sampled between slices
+        let optical_depth = -log(max(mean_transmittance, 1e-6)); // Avoid log(0)
+        textureStore(aerial_view_lut_out, vec3(vec2<u32>(idx.xy), slice_i), vec4(total_inscattering, optical_depth));
     }
 }
--- a/crates/bevy_pbr/src/atmosphere/aerial_view_lut.wgsl
+++ b/crates/bevy_pbr/src/atmosphere/aerial_view_lut.wgsl
@@ -32,10 +32,7 @@ fn main(@builtin(global_invocation_id) idx: vec3<u32>) {
 
     for (var slice_i: u32 = 0; slice_i < settings.aerial_view_lut_size.z; slice_i++) {
         for (var step_i: u32 = 0; step_i < settings.aerial_view_lut_samples; step_i++) {
-            // Offset by -0.5 to align sampling position with slice boundaries, 
-            // since each texel stores the integral over its entire slice
-            var t_i = t_max * (f32(slice_i) - 0.5 + ((f32(step_i) + MIDPOINT_RATIO) / f32(settings.aerial_view_lut_samples))) / f32(settings.aerial_view_lut_size.z);
-            t_i = max(t_i, 0.0);
+            let t_i = t_max * (f32(slice_i) + ((f32(step_i) + MIDPOINT_RATIO) / f32(settings.aerial_view_lut_samples))) / f32(settings.aerial_view_lut_size.z);
             let dt = (t_i - prev_t);
             prev_t = t_i;
 
@@ -61,9 +58,9 @@ fn main(@builtin(global_invocation_id) idx: vec3<u32>) {
         // We only have one channel to store transmittance, so we store the mean
         let mean_transmittance = (throughput.r + throughput.g + throughput.b) / 3.0;
 
-        // Store the optical depth so that it can be linearly sampled between slices
-        let optical_depth = -log(max(mean_transmittance, 1e-6)); // Avoid log(0)
+        // Store in log space to allow linear interpolation of exponential values between slices
+        let log_transmittance = -log(max(mean_transmittance, 1e-6)); // Avoid log(0)
         let log_inscattering = log(max(total_inscattering, vec3(1e-6)));
-        textureStore(aerial_view_lut_out, vec3(vec2<u32>(idx.xy), slice_i), vec4(log_inscattering, optical_depth));
+        textureStore(aerial_view_lut_out, vec3(vec2<u32>(idx.xy), slice_i), vec4(log_inscattering, log_transmittance));
     }
 }
--- a/crates/bevy_pbr/src/atmosphere/aerial_view_lut.wgsl
+++ b/crates/bevy_pbr/src/atmosphere/aerial_view_lut.wgsl
@@ -7,7 +7,7 @@
             sample_transmittance_lut, sample_atmosphere, rayleigh, henyey_greenstein,
             sample_multiscattering_lut, AtmosphereSample, sample_local_inscattering,
             get_local_r, get_local_up, view_radius, uv_to_ndc, max_atmosphere_distance,
-            uv_to_ray_direction
+            uv_to_ray_direction, MIDPOINT_RATIO
         },
     }
 }
@@ -32,7 +32,7 @@ fn main(@builtin(global_invocation_id) idx: vec3<u32>) {
 
     for (var slice_i: u32 = 0; slice_i < settings.aerial_view_lut_size.z; slice_i++) {
         for (var step_i: u32 = 0; step_i < settings.aerial_view_lut_samples; step_i++) {
-            let t_i = t_max * (f32(slice_i) + ((f32(step_i) + 0.5) / f32(settings.aerial_view_lut_samples))) / f32(settings.aerial_view_lut_size.z);
+            let t_i = t_max * (f32(slice_i) + ((f32(step_i) + MIDPOINT_RATIO) / f32(settings.aerial_view_lut_samples))) / f32(settings.aerial_view_lut_size.z);
             let dt = (t_i - prev_t);
             prev_t = t_i;
 
@@ -60,6 +60,7 @@ fn main(@builtin(global_invocation_id) idx: vec3<u32>) {
 
         // Store the optical depth so that it can be linearly sampled between slices
         let optical_depth = -log(max(mean_transmittance, 1e-6)); // Avoid log(0)
-        textureStore(aerial_view_lut_out, vec3(vec2<u32>(idx.xy), slice_i), vec4(total_inscattering, optical_depth));
+        let log_inscattering = log(max(total_inscattering, vec3(1e-6)));
+        textureStore(aerial_view_lut_out, vec3(vec2<u32>(idx.xy), slice_i), vec4(log_inscattering, optical_depth));
     }
 }
--- a/crates/bevy_pbr/src/atmosphere/functions.wgsl
+++ b/crates/bevy_pbr/src/atmosphere/functions.wgsl
@@ -75,7 +75,9 @@ fn sky_view_lut_r_mu_azimuth_to_uv(r: f32, mu: f32, azimuth: f32) -> vec2<f32> {
     let horizon_zenith = PI - beta;
     let view_zenith = fast_acos_4(mu);
 
-    // Latitude parameterization
+    // Apply non-linear transformation to compress more texels 
+    // near the horizon where high-frequency details matter most
+    // l is latitude in [-π/2, π/2] and v is texture coordinate in [0,1]
     let l = view_zenith - horizon_zenith;
     let abs_l = abs(l);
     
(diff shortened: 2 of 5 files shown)


crates/bevy_pbr/src/atmosphere/functions.wgsl
```diff
@@ -71,41 +71,34 @@ fn sky_view_lut_r_mu_azimuth_to_uv(r: f32, mu: f32, azimuth: f32) -> vec2<f32> {
...
-    if !ray_intersects_ground(r, mu) {
-        let coord = sqrt(1.0 - view_zenith / horizon_zenith);
-        v = (1.0 - coord) * 0.5;
-    } else {
-        let coord = (view_zenith - horizon_zenith) / beta;
-        v = sqrt(coord) * 0.5 + 0.5;
-    }
+    let view_zenith = fast_acos_4(mu);
+
+    // Latitude parameterization
```
alice-i-cecile, line 78: This comment could be more clear for those who aren't familiar with this area.

crates/bevy_pbr/src/atmosphere/functions.wgsl
```diff
@@ -236,15 +243,18 @@ fn sample_local_inscattering(local_atmosphere: AtmosphereSample, ray_dir: vec3<f
...
 fn sample_sun_illuminance(ray_dir_ws: vec3<f32>, transmittance: vec3<f32>) -> vec3<f32> {
+    let r = view_radius();
+    let mu_view = ray_dir_ws.y;
+    let shadow_factor = f32(!ray_intersects_ground(r, mu_view));
     var sun_illuminance = vec3(0.0);
     for (var light_i: u32 = 0u; light_i < lights.n_directional_lights; light_i++) {
         let light = &lights.directional_lights[light_i];
         let neg_LdotV = dot((*light).direction_to_light, ray_dir_ws);
-        let angle_to_sun = fast_acos(neg_LdotV);
+        let angle_to_sun = fast_acos_4(neg_LdotV);
```
alice-i-cecile, line 253: A comment here about why acos_4 was chosen would be nice.

crates/bevy_pbr/src/atmosphere/functions.wgsl
```diff
@@ -127,12 +120,26 @@ fn sample_sky_view_lut(r: f32, ray_dir_as: vec3<f32>) -> vec3<f32> {
...
+// RGB channels: total inscattered light along the camera ray to the current sample.
+// A channel: average transmittance across all wavelengths to the current sample.
 fn sample_aerial_view_lut(uv: vec2<f32>, depth: f32) -> vec4<f32> {
     let view_pos = view.view_from_clip * vec4(uv_to_ndc(uv), depth, 1.0);
-    let dist = length(view_pos.xyz / view_pos.w) * settings.scene_units_to_m;
-    let uvw = vec3(uv, dist / settings.aerial_view_lut_max_distance);
+    let view_ray_dist = length(view_pos.xyz / view_pos.w) * settings.scene_units_to_m;
+    let t_max = settings.aerial_view_lut_max_distance;
+
+    // Special handling for first slice to avoid extrapolation
```
jnhyatt, line 130: I would clarify here just a little bit. Your explanation on Discord helped a lot

jnhyatt (approved):
Just a bit of clarification on top of alice's, otherwise lgtm. Did some additional testing, I'll post a video showing the from-space view of the sun disc correctly dropping below the planet edge rather than the 0 degree zenith.

jnhyatt:
https://github.com/user-attachments/assets/3c09a456-7a3c-4091-a14e-54335c33f409

crates/bevy_pbr/src/atmosphere/functions.wgsl
```diff
@@ -127,12 +120,26 @@ fn sample_sky_view_lut(r: f32, ray_dir_as: vec3<f32>) -> vec3<f32> {
...
+    let delta_slice = t_max / f32(settings.aerial_view_lut_size.z);
+    if (view_ray_dist < delta_slice) {
+        let f = view_ray_dist / delta_slice;
+        let first_slice_uvw = vec3(uv, 0.5 / f32(settings.aerial_view_lut_size.z));
+        let sample = textureSampleLevel(aerial_view_lut, aerial_view_lut_sampler, first_slice_uvw, 0.0);
+        return vec4(sample.rgb * f, pow(sample.a, f));
```
ecoskey, line 136: why the `pow` here?

crates/bevy_pbr/src/atmosphere/functions.wgsl
```diff
@@ -127,12 +120,26 @@ fn sample_sky_view_lut(r: f32, ray_dir_as: vec3<f32>) -> vec3<f32> {
...
+    }
+
+    // Offset by 0.5 slice to sample at the center of each slice
+    let normalized_depth = (view_ray_dist / t_max) * f32(settings.aerial_view_lut_size.z - 1u);
+    let w = (normalized_depth - 0.5) / f32(settings.aerial_view_lut_size.z);
+    let uvw = vec3(uv, clamp(w, 0.0, 1.0));
```
ecoskey, line 145: maybe I'm missing something but this feels overcomplicated compared to just offsetting w by `0.5 / size.z`

ecoskey: actually why do this at all? we're not doing any math with texel coords so it's not like we're trying to avoid sampling in between texels--that's the point right?

ecoskey (approved):
few small nits that could use clarification, otherwise lgtm :)

crates/bevy_pbr/src/atmosphere/functions.wgsl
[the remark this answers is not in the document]
mate-h: Because the transmittance needs to be interpolated according to Beer-Lambert law, I just realized that sampling the LUT with the linear sampler is incorrect at the end of this function too. So instead, I think we should store the `log(transmittance)` in the Aerial View LUT and then later recover it with `exp(sample)`

crates/bevy_pbr/src/atmosphere/functions.wgsl
mate-h: Good points here, I was looking into simplifying this function some more and avoiding that offset by 0.5 to get a good match to the ray marched reference, will continue playing around and add a commit for a simplification and fix.

crates/bevy_pbr/src/atmosphere/functions.wgsl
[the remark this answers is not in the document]
ecoskey: makes sense 👍 

storing optical depth seems fine as long as it doesn't exceed the texel range, but that's probably pretty unlikely

crates/bevy_pbr/src/atmosphere/functions.wgsl
mate-h: After some testing and deliberation, I still settled on sampling the center of each LUT to produce matching result with the raymarched reference, this is because of better numerical precision when approximating the scattering integral using analytical approximation using the midpoint rule. Added code comment about this

ecoskey: If it better matches the reference, I'm all for it. I'm a little confused though, because it seems like this would mean sampling potentially far away from where we're meant to be sampling. For example, with 32 1km slices and a sample 1km away from the camera, we'd ideally interpolate evenly between slices 0 and 1, which are sampled at 500m and 1500m respectively. But with this change we'd sample purely from the 500m slice, and only get about half the optical density, unless I'm misunderstanding 🤔

mate-h:
image

mate-h:
After some more discussions, this discovery points to an issue in how the sample points are calculated. Therefore in order to remove the 0.5 texel offset in the sampling function, we need to fix the math that compute the sample positions. I will work on that tomorrow.

mate-h:
I've corrected the quadratic sampling in the sky view lut, reverted that to linear to recover the numerical accuracy. I've added a constant variable for the `MIDPOINT_RATIO` (open to renaming it) but it's now being used consistently in all raymarch loops. 
Additionally, I am using the `log(value)` and `exp(sample)` method to linearly sample the inscattered light in log space, so that gets rid of some of the banding at high altitudes for a lut that extends far.

crates/bevy_pbr/src/atmosphere/aerial_view_lut.wgsl
```diff
@@ -32,7 +32,10 @@ fn main(@builtin(global_invocation_id) idx: vec3<u32>) {
 
     for (var slice_i: u32 = 0; slice_i < settings.aerial_view_lut_size.z; slice_i++) {
         for (var step_i: u32 = 0; step_i < settings.aerial_view_lut_samples; step_i++) {
-            let t_i = t_max * (f32(slice_i) + ((f32(step_i) + 0.5) / f32(settings.aerial_view_lut_samples))) / f32(settings.aerial_view_lut_size.z);
+            // Offset by -0.5 to align sampling position with slice boundaries, 
+            // since each texel stores the integral over its entire slice
```
mate-h, line 36: I added this offset here and comment based on the discussion on discord, I think I am happy with this version overall.

crates/bevy_pbr/src/atmosphere/aerial_view_lut.wgsl
```diff
@@ -55,8 +58,12 @@ fn main(@builtin(global_invocation_id) idx: vec3<u32>) {
                 break;
             }
         }
-        //We only have one channel to store transmittance, so we store the mean
+        // We only have one channel to store transmittance, so we store the mean
         let mean_transmittance = (throughput.r + throughput.g + throughput.b) / 3.0;
-        textureStore(aerial_view_lut_out, vec3(vec2<u32>(idx.xy), slice_i), vec4(total_inscattering, mean_transmittance));
+
+        // Store the optical depth so that it can be linearly sampled between slices
+        let optical_depth = -log(max(mean_transmittance, 1e-6)); // Avoid log(0)
```
mate-h, line 65: Could call this variable neg_optical_depth or log_transmittance and remove the minus from both here and the sample function, might make it more readable.

crates/bevy_pbr/src/atmosphere/functions.wgsl
```diff
@@ -38,11 +38,17 @@
...
+const FRAC_2_PI: f32 = 0.15915494309;  // 1 / (2π)
 const FRAC_3_16_PI: f32 = 0.0596831036594607509; // 3 / (16π)
 const FRAC_4_PI: f32 = 0.07957747154594767; // 1 / (4π)
 const ROOT_2: f32 = 1.41421356; // √2
 
+// During raymarching, each segment is sampled at a single point. This constant determines
+// where in the segment that sample is taken (0.0 = start, 0.5 = middle, 1.0 = end).
+// We use 0.3 to sample closer to the start of each segment, which better approximates
+// the exponential falloff of atmospheric density.
+const MIDPOINT_RATIO: f32 = 0.3;
```
mate-h, line 50: This could be exposed as a uniform variable, since it can differ depending on the scale height of the scatterers in the atmosphere.

crates/bevy_pbr/src/atmosphere/aerial_view_lut.wgsl
mate-h: image
for a bit more of a visual explanation as well, in this photo the red represents the center of each slice while the green is the boundaries in the texture coordinate space.

crates/bevy_pbr/src/atmosphere/functions.wgsl
ecoskey: possibly, if it doesn't add more complexity to the user-facing api. Alternatively we could also just do the integration differently, for example piecewise linearly or quadratically (trapezoid rule or simpson's rule)

crates/bevy_pbr/src/atmosphere/aerial_view_lut.wgsl
```diff
@@ -32,7 +32,10 @@ fn main(@builtin(global_invocation_id) idx: vec3<u32>) {
 
     for (var slice_i: u32 = 0; slice_i < settings.aerial_view_lut_size.z; slice_i++) {
         for (var step_i: u32 = 0; step_i < settings.aerial_view_lut_samples; step_i++) {
-            let t_i = t_max * (f32(slice_i) + ((f32(step_i) + 0.5) / f32(settings.aerial_view_lut_samples))) / f32(settings.aerial_view_lut_size.z);
+            // Offset by -0.5 to align sampling position with slice boundaries, 
+            // since each texel stores the integral over its entire slice
+            var t_i = t_max * (f32(slice_i) - 0.5 + ((f32(step_i) + MIDPOINT_RATIO) / f32(settings.aerial_view_lut_samples))) / f32(settings.aerial_view_lut_size.z);
```
ecoskey, line 37: I feel really bad for this back-and-forth, but I actually feel like doing the offset in `sample_aerial_view_lut` makes more sense, since then the texel values mean the same thing for each slice. It does remove the need for the special first-slice handling, but I feel like handling it explicitly is better than hiding it in how the lut is parametrized.

crates/bevy_pbr/src/atmosphere/aerial_view_lut.wgsl
ecoskey: agreed, that's definitely simpler 👍

crates/bevy_pbr/src/atmosphere/functions.wgsl
```diff
@@ -71,41 +77,38 @@ fn sky_view_lut_r_mu_azimuth_to_uv(r: f32, mu: f32, azimuth: f32) -> vec2<f32> {
...
-    } else {
-        let coord = (view_zenith - horizon_zenith) / beta;
-        v = sqrt(coord) * 0.5 + 0.5;
-    }
+    let view_zenith = fast_acos_4(mu);
+
+    // Apply non-linear transformation to compress more texels 
+    // near the horizon where high-frequency details matter most
+    // l is latitude in [-π/2, π/2] and v is texture coordinate in [0,1]
+    let l = view_zenith - horizon_zenith;
```
ecoskey, line 89: just wanted to mention how happy I am that we get to remove that mess inside the if-else from before :)

crates/bevy_pbr/src/atmosphere/sky_view_lut.wgsl
```diff
@@ -36,19 +37,13 @@ fn main(@builtin(global_invocation_id) idx: vec3<u32>) {
 
     // Raymarch with quadratic distribution
```
ecoskey, line 38:
```suggestion
```
Should remove this if we're back to midpoint rule

mate-h:
Okay great! I think I've addressed everything and this PR is ready for the merge queue @alice-i-cecile

crates/bevy_pbr/src/atmosphere/aerial_view_lut.wgsl
mate-h: Okay let's be more explicit about that, agreed. I also updated a code comment to be more clear about the intent of the texel offset.

crates/bevy_pbr/src/atmosphere/functions.wgsl
```diff
@@ -135,10 +135,18 @@ fn sample_sky_view_lut(r: f32, ray_dir_as: vec3<f32>) -> vec3<f32> {
...
+    // align sampling position with slice boundaries, since each texel 
+    // stores the integral over its entire slice
+    let uvw = vec3(uv, saturate(dist / t_max - 0.5 / num_slices));
     let sample = textureSampleLevel(aerial_view_lut, aerial_view_lut_sampler, uvw, 0.0);
+    // Treat the first slice specially since there is 0 scattering at the camera
+    let delta_slice = t_max / num_slices;
+    let fade = saturate(dist / delta_slice);
     // Recover the values from log space
-    return vec4(exp(sample.rgb), exp(-sample.a));
+    return vec4(exp(sample.rgb) * fade, exp(sample.a) * fade);
```
ecoskey, line 149: one tiny nit: could use `return exp(sample) * fade;` here :)

mate-h: Good suggestion! Added ⚡️

alice-i-cecile:
Great cleanup and bug fixes; thanks a ton <3

PR #17556, opened 2025-01-27
Decouple the main sub app update and render sub app update

### LINKED ISSUE #1343: Frame rate limiting
Add a way to limit the frame rate.

Ideally, allow for more flexible control over updates, to support non-game gui applications that should sleep when idle.

SarthakSingh31:
# Objective

- Decouple the main sub app update and the render sub app update so that they can run at different speeds. The is really useful for creating desktop applications where we don't need to render a frame every world update. 
- Add a reliable way to cap the frame rate.
- Resolves #1343

## Solution

- Allow users to disable sub apps from updating with the main sub app update.
- Allow users to define a which sub apps to update on each window event.
- Disable the render sub app from updating with the main world.
- Add the render sub app to update on request redraw.
- Provide a new set of `UpdateMode` to control the update of the main and render sub apps.

## Testing

- Tested it by creating the `update_mode.rs` example.
- Most of the other features like 2d_sprites and animation need more testing.
- You can test these changes by inserting a `WinitSettings` where the main sub app update rate is faster than the render sub app.
- I have only tested these changes on linux.

---

## Showcase

The new `update_mode.rs` example:

output

## Migration Guide

- If the user is setting the `WinitSettings` they will need to add the `RenderUpdateMode` to both of the its fields.
- Any system that executes in the `ExtractSchedule` now can run multiple times before a the render sub app is updated. This means that the systems running inside the `ExtractSchedule` should no longer assume that the render sub app will cleanup any resources or entities that are spawned.

### DIFF
--- a/crates/bevy_ui/src/render/mod.rs
+++ b/crates/bevy_ui/src/render/mod.rs
@@ -99,6 +99,7 @@ pub const UI_SHADER_HANDLE: Handle<Shader> = Handle::weak_from_u128(130128470471
 
 #[derive(Debug, Hash, PartialEq, Eq, Clone, SystemSet)]
 pub enum RenderUiSystem {
+    Clear,
     ExtractCameraViews,
     ExtractBoxShadows,
     ExtractBackgrounds,
@@ -128,6 +129,7 @@ pub fn build_ui_render(app: &mut App) {
         .configure_sets(
             ExtractSchedule,
             (
+                RenderUiSystem::Clear,
                 RenderUiSystem::ExtractCameraViews,
                 RenderUiSystem::ExtractBoxShadows,
                 RenderUiSystem::ExtractBackgrounds,
@@ -142,6 +144,7 @@ pub fn build_ui_render(app: &mut App) {
         .add_systems(
             ExtractSchedule,
             (
+                clear_extracted_ui_nodes.in_set(RenderUiSystem::Clear),
                 extract_ui_camera_view.in_set(RenderUiSystem::ExtractCameraViews),
                 extract_uinode_background_colors.in_set(RenderUiSystem::ExtractBackgrounds),
                 extract_uinode_images.in_set(RenderUiSystem::ExtractImages),
@@ -274,6 +277,17 @@ impl RenderGraphNode for RunUiSubgraphOnUiViewNode {
         graph.run_sub_graph(SubGraphUi, vec![], Some(default_camera_view.0))?;
         Ok(())
     }
+}
+
+pub fn clear_extracted_ui_nodes(
+    mut commands: Commands,
+    mut extracted_uinodes: ResMut<ExtractedUiNodes>,
+) {
+    extracted_uinodes
+        .uinodes
+        .drain()
+        .for_each(|(entity, _)| commands.entity(entity).despawn());
+    extracted_uinodes.glyphs.clear();
 }
 
 pub fn extract_uinode_background_colors(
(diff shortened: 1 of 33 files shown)


crates/bevy_ui/src/render/mod.rs
```diff
@@ -276,6 +279,17 @@ impl RenderGraphNode for RunUiSubgraphOnUiViewNode {
...
 }
 
+pub fn clear_extracted_ui_nodes(
+    mut commands: Commands,
+    mut extracted_uinodes: ResMut<ExtractedUiNodes>,
+) {
+    extracted_uinodes
+        .uinodes
+        .drain()
+        .for_each(|(entity, _)| commands.entity(entity).despawn());
```
ickshonpe, line 289: Isn't it enough to just clear the buffer here? They are all `TemporaryRenderEntity` entites spawned during extraction so there are shouldn't be any collisions and they'll be despawned automatically?

ickshonpe (commented):
On windows get a panic: 
```
thread 'main' panicked at std\src\time.rs:433:33:
overflow when subtracting duration from instant
stack backtrace:
   0: std::panicking::begin_panic_handler
             at /rustc/9fc6b43126469e3858e2fe86cafb4f0fd5068869\library/std\src\panicking.rs:665
   1: core::panicking::panic_fmt
             at /rustc/9fc6b43126469e3858e2fe86cafb4f0fd5068869\library/core\src\panicking.rs:76
   2: core::panicking::panic_display
             at /rustc/9fc6b43126469e3858e2fe86cafb4f0fd5068869\library/core\src\panicking.rs:269
   3: core::option::expect_failed
             at /rustc/9fc6b43126469e3858e2fe86cafb4f0fd5068869\library/core\src\option.rs:2018
   4: core::option::Option::expect
             at /rustc/9fc6b43126469e3858e2fe86cafb4f0fd5068869\library/core\src\option.rs:933
   5: std::time::impl$3::sub
             at /rustc/9fc6b43126469e3858e2fe86cafb4f0fd5068869\library/std\src\time.rs:433
   6: bevy_winit::state::winit_runner
   7: core::ops::function::FnOnce::call_once{{vtable.shim}}
   8: bevy_app::app::App::run
   9: bevy_ecs::component::Components::register_component
```

crates/bevy_ui/src/render/mod.rs
SarthakSingh31: I did this because I was worried that the entities spawned might still be mistakenly used in the next stages.
Each of those entity has some component attached to it.
In the `Render` schedule a user can query for the `Component` instead of accessing the entities using the list in the resource. In that case they will get both the current as well as the outdated entities.

SarthakSingh31:
@ickshonpe Does it work on windows now? I removed the only place where I was doing a unchecked subtraction.

crates/bevy_ui/src/render/mod.rs
ickshonpe: IIRC I think they just get a `UiBatch` added to them and we only ever look them up individually by their render entity id,  not by querying for the component.

ickshonpe (commented):
There's a minor problem with the `multiple_windows` example I just noticed.
If you switch the focus to a different non-bevy window or try dragging either bevy window, the text from both windows disappears. It immediately comes back once you stop dragging or switch back to the bevy window. The 3d model is unaffected.

janhohenheim:
Triage: has merge conflicts, maintainers think this may not be the best move for the moment
@SarthakSingh31  do you want to update this PR or should I tag it `S-Needs-Adoption`? :)

SarthakSingh31:
@janhohenheim Unfortunately the company that was sponsoring this work has moved away for bevy for the moment. I might pick this work back up in the future but I think its best to close this PR for now.

Issue #17478, opened 2025-01-21
Missing a method to insert related entities at a specific index (`insert_children`)


villor:
## What problem does this solve or what need does it fill?

Before #17398 there was a pretty handy method for inserting a slice of children at a specific index: [insert_children](https://docs.rs/bevy_hierarchy/0.15.1/src/bevy_hierarchy/child_builder.rs.html#645)

Not sure if it was intentionally left out, or just lost in the process. It's pretty useful for UI sometimes when needing to insert one or more children in specific spots.

While migrating `bevy_editor_prototypes` I resorted to an extension for now:
```rust
impl InsertChildrenExt for EntityWorldMut<'_> {
    fn insert_children(&mut self, index: usize, children: &[Entity]) {
        let prev_children = self.take::<Children>().unwrap_or_default();
        let (prev_left, prev_right) = prev_children.split_at(index);

        let parent_id = self.id();
        self.world_scope(|world| {
            for child in prev_left.iter().chain(children).chain(prev_right) {
                world.entity_mut(*child).insert(ChildOf(parent_id));
            }
        });
    }
}
```

## What solution would you like?

- Add `insert_related` method on `EntityWorldMut`/`EntityCommands`, inserting a `&[Entity]` (or iterator?) at a specific index.
- Add `insert_children` for convenience and less breakage when migrating

## What alternative(s) have you considered?

- Could just skip it and let "user-space" handle it like the extension above or similar. But seeing as it was in there before, I guess at some point it was deemed useful enough to keep upstream.

bjoernp116:
Working on this!

PR #17558, opened 2025-01-27
Insert children

bjoernp116:
# Objective

This PR implements the requested, previously deprecated feature from #17478. Which is concerning the `insert_children` of `EntityWorldMut`. This method inserts children into the tree, at a specific index.

## Solution

I mostly copied the code from the deprecated crate [`bevy_hierarchy`](https://docs.rs/bevy_hierarchy/0.15.1/src/bevy_hierarchy/child_builder.rs.html#379) though this is no longer a trait method.
This code also used `SmallVec` which has later been switched by `Vec`. This was easily interchangeable with working code.

## Testing

I wrote a unit test covering both `EntityWorldMut::add_entities` and `EntityWorldMut::insert_entities`, simply called `hierarchy::tests::insert_children`. I was unsure what error handling architecture i should be using, but i think the simplest to debug in this case would just be a panic in the case of an index out of bounds.

## Showcase

I find this feature pretty self exoplanetary, but the unit test can also bee informing about what this feature implements:
<details>

```rust
let mut world = World::new();
let child1 = world.spawn_empty().id();
let child2 = world.spawn_empty().id();
let child3 = world.spawn_empty().id();
let child4 = world.spawn_empty().id();

let mut entity_world_mut = world.spawn_empty();

let first_children = entity_world_mut.add_children(&[child1, child2]);

let root = first_children.insert_children(1, &[child3, child4]).id();

let hierarchy = get_hierarchy(&world, root);
assert_eq!(
     hierarchy,
     Node::new_with(root, vec![Node::new(child1), Node::new(child3), Node::new(child4), Node::new(child2)])
);
```

</details>

### DIFF
--- a/crates/bevy_ecs/src/hierarchy.rs
+++ b/crates/bevy_ecs/src/hierarchy.rs
@@ -204,7 +204,7 @@ impl<'w> EntityWorldMut<'w> {
             }
             children_component.0.reserve(children.len());
             let mut v = children_component.0.split_off(index);
-            children_component.0.extend_from_slice(&children);
+            children_component.0.extend_from_slice(children);
             children_component.0.append(&mut v);
         } else {
             self.insert(Children(children.to_vec()));
--- a/crates/bevy_ecs/src/hierarchy.rs
+++ b/crates/bevy_ecs/src/hierarchy.rs
@@ -192,7 +192,11 @@ impl<'w> EntityWorldMut<'w> {
             panic!("Cannot insert entity as a child of itself.");
         }
         if let Some(mut children_component) = self.get_mut::<Children>() {
+
             children_component.0.retain(|value| !children.contains(value));
+            if index >= children_component.len() {
+                panic!("Index {} out of bounds! There are only {} children!", index, children.len());
+            }
             children_component.0.reserve(children.len());
             let mut v = children_component.0.split_off(index);
             children_component.0.extend_from_slice(&children);
--- a/crates/bevy_ecs/src/hierarchy.rs
+++ b/crates/bevy_ecs/src/hierarchy.rs
@@ -192,16 +192,20 @@ impl<'w> EntityWorldMut<'w> {
             panic!("Cannot insert entity as a child of itself.");
         }
         if let Some(mut children_component) = self.get_mut::<Children>() {
-
-            children_component.0.retain(|value| !children.contains(value));
+            children_component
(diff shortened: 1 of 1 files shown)


BenjaminBrienen:
Fixes https://github.com/bevyengine/bevy/issues/17478

crates/bevy_ecs/src/hierarchy.rs
```diff
@@ -185,6 +185,33 @@ impl<'w> EntityWorldMut<'w> {
         self.add_related::<ChildOf>(children)
     }
 
+    /// Insert children at specific index
```
BenjaminBrienen, line 188: What happens if the entity is already a child? As a user, I'd like to know whether that is okay ok or not. For example, does doing so reorder it?

crates/bevy_ecs/src/hierarchy.rs
```diff
@@ -185,6 +185,33 @@ impl<'w> EntityWorldMut<'w> {
...
+                .retain(|value| !children.contains(value));
+            if index >= children_component.len() {
+                panic!(
+                    "Index {} out of bounds! There are only {} children!",
+                    index,
+                    children.len()
+                );
+            }
+            children_component.0.reserve(children.len());
+            let mut v = children_component.0.split_off(index);
```
cart, line 206: This does not update the other side of the relationship (`ChildOf`), invalidating the assumptions of the system as a whole.

alice-i-cecile:
@bjoernp116 this is a blocker to 0.16; are you interested in finishing this up in the next week or two? Totally fine if not; we can always get someone else to finish this.

crates/bevy_ecs/src/hierarchy.rs
```diff
@@ -185,6 +185,33 @@ impl<'w> EntityWorldMut<'w> {
         self.add_related::<ChildOf>(children)
     }
 
+    /// Insert children at specific index
+    pub fn insert_children(&mut self, index: usize, children: &[Entity]) -> &mut Self {
+        let parent = self.id();
+        if children.contains(&parent) {
+            panic!("Cannot insert entity as a child of itself.");
```
marko-lazic, line 192: Is it possible to change the return to Bevy Result and then return `Err()` here?

alice-i-cecile:
Closing as adopted <3 Thanks for getting started on this!

PR #17560, opened 2025-01-27
Remove unnecessary `PartialReflect` bound on `DeserializeWithRegistry`

aecsocket:
# Objective

The new `DeserializeWithRegistry` trait in 0.15 was designed to be used with reflection, and so has a `trait DeserializeWithRegistry: PartialReflect` bound. However, this bound is not actually necessary for the trait to function properly. And this `PartialReflect` bound already exists on:
```rs
impl<T: PartialReflect + for<'de> DeserializeWithRegistry<'de>> FromType<T>
    for ReflectDeserializeWithRegistry
```
So there is no point in constraining the trait itself with this bound as well.

This lets me use `DeserializeWithRegistry` with non-`Reflect` types, which I want to do to avoid making a bunch of `FooDeserializer` structs and `impl DeserializeSeed` on them.

## Solution

Removes this unnecessary bound.

## Testing

Trivial change, does not break compilation or `bevy_reflect` tests.

## Migration Guide

`DeserializeWithRegistry` types are no longer guaranteed to be `PartialReflect` as well. If you were relying on this type bound, you should add it to your own bounds manually.
```diff
- impl<T: DeserializeWithRegistry> Foo for T { .. }
+ impl<T: DeserializeWithRegistry + PartialReflect> Foo for T { .. }
```


alice-i-cecile (approved):
This makes sense and I trust the compiler here.

Carter0 (approved):

PR #17570, opened 2025-01-28
Automatically enable `portable-atomic` when required

bushrat011899:
# Objective

- Contributes to #15460
- Reduce quantity and complexity of feature gates across Bevy

## Solution

- Used `target_has_atomic` configuration variable to automatically detect impartial atomic support and automatically switch to `portable-atomic` over the standard library on an as-required basis.

## Testing

- CI

## Notes

To explain the technique employed here, consider getting `Arc` either from `alloc::sync` _or_ `portable-atomic-util`. First, we can inspect the `alloc` crate to see that you only have access to `Arc` _if_ `target_has_atomic = "ptr"`. We add a target dependency for this particular configuration _inverted_:

```toml
[target.'cfg(not(target_has_atomic = "ptr"))'.dependencies]
portable-atomic-util = { version = "0.2.4", default-features = false }
```

This ensures we only have the dependency when it is needed, and it is entirely excluded from the dependency graph when it is not. Next, we adjust our configuration flags to instead of checking for `feature = "portable-atomic"` to instead check for `target_has_atomic = "ptr"`:

```rust
// `alloc` feature flag hidden for brevity

#[cfg(not(target_has_atomic = "ptr"))]
use portable_atomic_util as arc;

#[cfg(target_has_atomic = "ptr")]
use alloc::sync as arc;

pub use arc::{Arc, Weak};
```

The benefits of this technique are three-fold:

1. For platforms without full atomic support, the functionality is enabled automatically.
2. For platforms with atomic support, the dependency is never included, even if a feature was enabled using `--all-features` (for example)
3. The `portable-atomic` feature no longer needs to virally spread to all user-facing crates, it's instead something handled within `bevy_platform_support` (with some extras where other dependencies also need their features enabled).

### DIFF
--- a/crates/bevy_a11y/Cargo.toml
+++ b/crates/bevy_a11y/Cargo.toml
@@ -50,15 +50,6 @@ critical-section = [
   "bevy_input_focus/critical-section",
 ]
 
-## `portable-atomic` provides additional platform support for atomic types and
-## operations, even on targets without native support. 
-portable-atomic = [
-  "bevy_app/portable-atomic",
-  "bevy_ecs/portable-atomic",
-  "bevy_reflect?/portable-atomic",
-  "bevy_input_focus/portable-atomic",
-]
-
 ## Uses the `libm` maths library instead of the one provided in `std` and `core`.
 libm = ["bevy_input_focus/libm"]
 
--- a/crates/bevy_ecs/Cargo.toml
+++ b/crates/bevy_ecs/Cargo.toml
@@ -87,15 +87,6 @@ critical-section = [
   "bevy_reflect?/critical-section",
 ]
 
-## `portable-atomic` provides additional platform support for atomic types and
-## operations, even on targets without native support.
-portable-atomic = [
-  "bevy_tasks?/portable-atomic",
-  "bevy_platform_support/portable-atomic",
-  "concurrent-queue/portable-atomic",
-  "bevy_reflect?/portable-atomic",
-]
-
 [dependencies]
 bevy_ptr = { path = "../bevy_ptr", version = "0.16.0-dev" }
 bevy_reflect = { path = "../bevy_reflect", version = "0.16.0-dev", default-features = false, optional = true }
@@ -130,6 +121,11 @@ variadics_please = { version = "1.1", default-features = false }
 tracing = { version = "0.1", default-features = false, optional = true }
 log = { version = "0.4", default-features = false }
 bumpalo = "3"
+
+[target.'cfg(not(all(target_has_atomic = "8", target_has_atomic = "16", target_has_atomic = "32", target_has_atomic = "64", target_has_atomic = "ptr")))'.dependencies]
+concurrent-queue = { version = "2.5.0", default-features = false, features = [
+  "portable-atomic",
+] }
 
 [dev-dependencies]
 rand = "0.8"
--- a/crates/bevy_platform_support/Cargo.toml
+++ b/crates/bevy_platform_support/Cargo.toml
@@ -24,32 +24,20 @@ serialize = ["hashbrown/serde"]
 std = [
   "alloc",
   "critical-section?/std",
-  "portable-atomic?/std",
-  "portable-atomic-util?/std",
+  "portable-atomic/std",
+  "portable-atomic-util/std",
   "spin/std",
   "foldhash/std",
 ]
 
-alloc = ["portable-atomic-util?/alloc", "dep:hashbrown"]
+alloc = ["portable-atomic-util/alloc", "dep:hashbrown"]
 
 ## `critical-section` provides the building blocks for synchronization primitives
 ## on all platforms, including `no_std`.
-critical-section = ["dep:critical-section", "portable-atomic?/critical-section"]
-
-## `portable-atomic` provides additional platform support for atomic types and
-## operations, even on targets without native support.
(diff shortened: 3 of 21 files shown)


crates/bevy_reflect/src/impls/std.rs
```diff
@@ -217,11 +217,6 @@ impl_reflect_opaque!(::core::num::Wrapping<T: Clone + Send + Sync>());
 impl_reflect_opaque!(::core::num::Saturating<T: Clone + Send + Sync>());
 impl_reflect_opaque!(::bevy_platform_support::sync::Arc<T: Send + Sync + ?Sized>);
 
-// We check despite `portable-atomic` being enabled, if the standard library `Arc` is
-// also available, and implement Reflect for it.
-#[cfg(all(feature = "portable-atomic", target_has_atomic = "ptr"))]
-impl_reflect_opaque!(::alloc::sync::Arc<T: Send + Sync + ?Sized>);
-
```
bushrat011899, line 224: No longer required, since we can now guarantee that `bevy_platform_support::sync::Arc` will be `alloc::sync::Arc` if it exists.

crates/bevy_a11y/Cargo.toml
```diff
@@ -50,15 +50,6 @@ critical-section = [
...
 
-## `portable-atomic` provides additional platform support for atomic types and
-## operations, even on targets without native support. 
-portable-atomic = [
-  "bevy_app/portable-atomic",
-  "bevy_ecs/portable-atomic",
-  "bevy_reflect?/portable-atomic",
-  "bevy_input_focus/portable-atomic",
-]
-
```
bushrat011899, line 61: Since `bevy_platform_support` can now automatically detect the need for `portable-atomic`, user-facing crates no longer need to provide that feature.

crates/bevy_ecs/Cargo.toml
```diff
@@ -131,6 +122,11 @@ tracing = { version = "0.1", default-features = false, optional = true }
 log = { version = "0.4", default-features = false }
 bumpalo = "3"
 
+[target.'cfg(not(all(target_has_atomic = "8", target_has_atomic = "16", target_has_atomic = "32", target_has_atomic = "64", target_has_atomic = "ptr")))'.dependencies]
+concurrent-queue = { version = "2.5.0", default-features = false, features = [
+  "portable-atomic",
+] }
+
```
bushrat011899, line 129: Since `bevy_ecs` has a dependency which itself has a `portable-atomic` feature, we replicate the hack here to just automatically enable its feature on required targets.

crates/bevy_platform_support/Cargo.toml
```diff
@@ -68,6 +56,17 @@ hashbrown = { version = "0.15.1", features = [
...
+[target.'cfg(not(all(target_has_atomic = "8", target_has_atomic = "16", target_has_atomic = "32", target_has_atomic = "64", target_has_atomic = "ptr")))'.dependencies]
+portable-atomic = { version = "1", default-features = false, features = [
+  "fallback",
+] }
+spin = { version = "0.9.8", default-features = false, features = [
+  "portable_atomic",
+] }
+
+[target.'cfg(not(target_has_atomic = "ptr"))'.dependencies]
+portable-atomic-util = { version = "0.2.4", default-features = false }
```
bushrat011899, line 68: Since we only need `portable-atomic-util` for `Arc`, and `Arc` is only available when `target_has_atomic = "ptr"`, we can simplify its target dependency a little.

crates/bevy_platform_support/Cargo.toml
```diff
@@ -68,6 +56,17 @@ hashbrown = { version = "0.15.1", features = [
 web-time = { version = "1.1", default-features = false }
 getrandom = { version = "0.2.0", default-features = false, features = ["js"] }
 
+[target.'cfg(not(all(target_has_atomic = "8", target_has_atomic = "16", target_has_atomic = "32", target_has_atomic = "64", target_has_atomic = "ptr")))'.dependencies]
+portable-atomic = { version = "1", default-features = false, features = [
+  "fallback",
+] }
+spin = { version = "0.9.8", default-features = false, features = [
+  "portable_atomic",
+] }
```
bushrat011899, line 65: Essentially, "_If any atomic feature isn't available, pessimistically include `portable-atomic`_"

crates/bevy_tasks/src/task_pool.rs
```diff
@@ -74,16 +74,20 @@ impl TaskPoolBuilder {
...
+            unsafe_code,
+            reason = "unsized coercion is an unstable feature for non-std types"
+        )]
+        // SAFETY:
+        // - Coercion from `impl Fn` to `dyn Fn` is valid
+        // - `Arc::from_raw` receives a valid pointer from a previous call to `Arc::into_raw`
+        let arc = unsafe {
+            Arc::from_raw(Arc::into_raw(arc) as *const (dyn Fn() + Send + Sync + 'static))
+        };
+
```
bushrat011899, line 90: Just a little tidy-up while I'm here. I'm confident the pointer-cast is valid and it avoids a second allocation.

bushrat011899 (commented):
Some notes to help reviewers.

mockersf (approved):

alice-i-cecile (approved):

PR #17571, opened 2025-01-28
Fix cursor hotspot out of bounds when flipping

mgi388:
# Objective

- Fix off by one error introduced in https://github.com/bevyengine/bevy/pull/17540 causing:

```
Cursor image StrongHandle<Image>{ id: Index(AssetIndex { generation: 0, index: 3 }), path: Some(cursors/kenney_crosshairPack/Tilesheet/crosshairs_tilesheet_white.png) } is invalid: The specified hotspot (64, 64) is outside the image bounds (64x64).
```

- First PR commit and run shows the bug: https://github.com/bevyengine/bevy/actions/runs/13009405866/job/36283507530?pr=17571
- Second PR commit fixes it.

## Solution

- Hotspot coordinates are 0-indexed, so we need to subtract 1 from the width and height.

## Testing

- Fix the tests which included the off-by-one error in their expected values.
- Consolidate the tests into a single test for brevity.
- Test round trip transform to ensure we can "undo" to get back to the original value.
- Add a specific bounds test.
- Ran the example again and observed there are no more error logs: `cargo run --example custom_cursor_image --features=custom_cursor`.

### DIFF
--- a/crates/bevy_winit/src/custom_cursor.rs
+++ b/crates/bevy_winit/src/custom_cursor.rs
@@ -204,11 +204,16 @@ pub(crate) fn transform_hotspot(
 ) -> (u16, u16) {
     let hotspot_x = hotspot.0 as f32;
     let hotspot_y = hotspot.1 as f32;
+
     let (width, height) = (rect.width(), rect.height());
 
-    let hotspot_x = if flip_x { width - hotspot_x } else { hotspot_x };
+    let hotspot_x = if flip_x {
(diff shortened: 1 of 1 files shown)


mgi388:
@eero-lehtinen 🙏

eero-lehtinen (approved):
👍

mgi388:
@BenjaminBrienen  fancy reviewing this if you get a chance?

alice-i-cecile (approved):

PR #17574, opened 2025-01-28
don't use bevy_pbr for base bevy_gizmos plugin

rambip:
# Objective

This PR enables `bevy_gizmos` to be used without `bevy_pbr`, for user who want to create their custom mesh rendering logic.

It can also be useful for user who just want to use bevy for drawing lines (why not).

This work is part of a bigger effort to make the bevy rendering pipeline more modular. I would like to contribute an exemple to render custom meshes without `bevy_pbr`. Something like [this](https://github.com/rambip/plant-mesh/blob/main/src/shader/mod.rs)

## Solution

Now, `bevy_pbr` is an optional dependency, and used only to debug lights.

I query the `ViewUniforms` manually, instead of using `bevy_pbr` to get the heavy `MeshViewLayout`

## Testing

I'm not used to testing with bevy at all, but I was able to use successfully in my project.
It might break for some different mesh pipelines, but I don't think so.

---

## Showcase

image
So nice ...

## Migration Guide

I don't think there is any breaking change

# Remaining work

Before merging it, it would be useful to:
- remove the `bevy_sprite` depedency too
- move `view.rs` to `bevy_render`, so that it can be used in a more modular way.
- include the most recent changes from 0.16


mockersf:
you should open PR against the `main` branch

PR #17577, opened 2025-01-28
Improve ergonomics of `platform_support`'s `Instant`

Mathspy:
# Objective

- Make working with `bevy_time` more ergonomic in `no_std` environments.

Currently `bevy_time` expects the getter in environments where time can't be obtained automatically via the instruction set or the standard library to be of type `*mut fn() -> Duration`. [`fn()`](https://doc.rust-lang.org/beta/std/primitive.fn.html) is already a function pointer, so `*mut fn()` is a _pointer to a function pointer_. This is harder to use and error prone since creating a pointer out of something like `&mut fn() -> Duration` when the lifetime of the reference isn't static will lead to an undefined behavior once the reference is freed

## Solution

- Accept a `fn() -> Duration` instead

## Testing

- I made a whole game on the Playdate that relies on `bevy_time` heavily, see: [bevydate_time](https://github.com/Mathspy/bevydate/blob/1b4f02adcde079cf9757fd3c7d20331c9ab04513/src/lib.rs#L510-L546) for usage of the Instant's getter.

---

## Showcase

<details>

https://github.com/user-attachments/assets/f687847f-6b62-4322-95f3-c908ada3db30

</details>

## Migration Guide

This is a breaking change but it's not for people coming from Bevy v0.15

### Small thank you note
Thanks to my friend https://github.com/repnop for helping me understand how to deal with function pointers in `unsafe` environments

### DIFF
--- a/crates/bevy_platform_support/src/time.rs
+++ b/crates/bevy_platform_support/src/time.rs
@@ -30,7 +30,7 @@ mod fallback {
         time::Duration,
     };
 
-    static ELAPSED_GETTER: AtomicPtr<fn() -> Duration> = AtomicPtr::new(unset_getter as *mut _);
+    static ELAPSED_GETTER: AtomicPtr<()> = AtomicPtr::new(unset_getter as *mut _);
 
     /// Fallback implementation of `Instant` suitable for a `no_std` environment.
     ///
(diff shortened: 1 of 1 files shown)


Mathspy:
> Yeah that is better, thanks for submitting this! And thanks for confirming that the Playdate works too!

Thank YOU for all the hard work championing Bevy's `no_std` support in such a short time ❤️ I was really excited when I saw https://github.com/bevyengine/bevy/issues/15460 a few months ago when I ordered my Playdate and couldn't wait to run Bevy on it

BD103 (approved):

BD103:
The changes look good, and the demo video was super cool! (Now I want a Playdate ^^)

mockersf (approved):

PR #17579, opened 2025-01-28
Store UI render target info locally per node

### LINKED ISSUE #15143: non-root `TargetCamera`s are not ignored
## Bevy version

main

## What you did

Added this system to `render_ui_to_texture`:

```rust
fn target_cameras(
    mut commands: Commands,
    mut count: Local<usize>,
    mut query: Query<&mut TargetCamera, With<Parent>>,
) {
    *count += 1;
    if *count == 2 {
        // spawn a UI camera!
        let ui_camera = commands
            .spawn(Camera2dBundle {
                camera: Camera {
                    order: 10,
                    ..Default::default()
                },
                ..Default::default()
            })
            .id();

        for mut c in query.iter_mut() {
            c.0 = ui_camera;
        }
    }
}
```

## What went wrong

rtt

Non-root `TargetCamera`s are meant to have no effect but `update_target_camera_system` only does an update if it detects changes to the root node's `TargetCamera` and not the `TargetCamera` components of any of the rest of the entities in the UI node tree.

### LINKED ISSUE #17578: `measure_text_system` scale factor change detection watches the camera target's scale factor, not the text node's
## Bevy version

main

## What's wrong

`measure_text_system` queries for the camera entity of a text node and then using the camera entity looks up the camera target's scale factor. Then to detect if the scale factor has changed it looks up the camera's scale factor from the previous frame it keeps stored in a `Local` and compares them. But if the `text` node's target camera was changed since the previous frame it will be comparing the wrong scale factors and the `text` node's measure func won't be updated if its scale factor changed.

ickshonpe:
# Objective

It's difficult to understand or make changes to the UI systems because of how each system needs to individually track changes to scale factor, windows and camera targets in local hashmaps, particularly for new contributors. Any major change inevitably introduces new scale factor bugs. 

Instead of per-system resolution we can resolve the camera target info for all UI nodes in a system at the start of `PostUpdate` and then store it per-node in components that can be queried with change detection.

Fixes #17578
Fixes #15143

## Solution

Store the UI render target's data locally per node in a component that is updated in `PostUpdate` before any other UI systems run.

This component can be then be queried with change detection so that UI systems no longer need to have knowledge of cameras and windows and don't require fragile custom change detection solutions using local hashmaps. 

## Showcase
Compare `measure_text_system` from main (which has a bug the causes it to use the wrong scale factor when a node's camera target changes):
```
pub fn measure_text_system(
    mut scale_factors_buffer: Local<EntityHashMap<f32>>,
    mut last_scale_factors: Local<EntityHashMap<f32>>,
    fonts: Res<Assets<Font>>,
    camera_query: Query<(Entity, &Camera)>,
    default_ui_camera: DefaultUiCamera,
    ui_scale: Res<UiScale>,
    mut text_query: Query<
        (
            Entity,
            Ref<TextLayout>,
            &mut ContentSize,
            &mut TextNodeFlags,
            &mut ComputedTextBlock,
            Option<&UiTargetCamera>,
        ),
        With<Node>,
    >,
    mut text_reader: TextUiReader,
    mut text_pipeline: ResMut<TextPipeline>,
    mut font_system: ResMut<CosmicFontSystem>,
) {
    scale_factors_buffer.clear();

    let default_camera_entity = default_ui_camera.get();

    for (entity, block, content_size, text_flags, computed, maybe_camera) in &mut text_query {
        let Some(camera_entity) = maybe_camera
            .map(UiTargetCamera::entity)
            .or(default_camera_entity)
        else {
            continue;
        };
        let scale_factor = match scale_factors_buffer.entry(camera_entity) {
            Entry::Occupied(entry) => *entry.get(),
            Entry::Vacant(entry) => *entry.insert(
                camera_query
                    .get(camera_entity)
                    .ok()
                    .and_then(|(_, c)| c.target_scaling_factor())
                    .unwrap_or(1.0)
                    * ui_scale.0,
            ),
        };

        if last_scale_factors.get(&camera_entity) != Some(&scale_factor)
            || computed.needs_rerender()
            || text_flags.needs_measure_fn
            || content_size.is_added()
        {
            create_text_measure(
                entity,
                &fonts,
                scale_factor.into(),
                text_reader.iter(entity),
                block,
                &mut text_pipeline,
                content_size,
                text_flags,
                computed,
                &mut font_system,
            );
        }
    }
    core::mem::swap(&mut *last_scale_factors, &mut *scale_factors_buffer);
}
```

with `measure_text_system` from this PR (which always uses the correct scale factor):
```
pub fn measure_text_system(
    fonts: Res<Assets<Font>>,
    mut text_query: Query<
        (
            Entity,
            Ref<TextLayout>,
            &mut ContentSize,
            &mut TextNodeFlags,
            &mut ComputedTextBlock,
            Ref<ComputedNodeTarget>,
        ),
        With<Node>,
    >,
    mut text_reader: TextUiReader,
    mut text_pipeline: ResMut<TextPipeline>,
    mut font_system: ResMut<CosmicFontSystem>,
) {
    for (entity, block, content_size, text_flags, computed, computed_target) in &mut text_query {
        // Note: the ComputedTextBlock::needs_rerender bool is cleared in create_text_measure().
        if computed_target.is_changed()
            || computed.needs_rerender()
            || text_flags.needs_measure_fn
            || content_size.is_added()
        {
            create_text_measure(
                entity,
                &fonts,
                computed_target.scale_factor.into(),
                text_reader.iter(entity),
                block,
                &mut text_pipeline,
                content_size,
                text_flags,
                computed,
                &mut font_system,
            );
        }
    }
}
```

## Testing

I removed an alarming number of tests from the `layout` module but they were mostly to do with the deleted camera synchronisation logic. The remaining tests should all pass now.

The most relevant examples are `multiple_windows` and `split_screen`, the behaviour of both should be unchanged from main.


UkoeHB (approved):
Very nice PR

ickshonpe:
> Very nice PR
> 
> Only comment is perhaps you could consolidate `ComputedNodeTargetSize` and `ComputedNodeScaleFactor` into one component.

Thanks for reviewing this. I agree, I was overestimating the advantages of more granular change detection. In practice changing the resolution or scale factor is almost always going to result in a full relayout of everything so there's no benefit to having the separate components.

crates/bevy_ui/src/ui_node.rs
```diff
@@ -2763,6 +2766,43 @@ impl Default for BoxShadowSamples {
     }
 }
 
+/// ID of the UI node's camera target.
```
alice-i-cecile, line 2769:
```suggestion
/// Derived information about the camera target for this UI node.
```

alice-i-cecile (approved):

PR #17580, opened 2025-01-28
feat(ecs): add `EntityEntryCommands::entity()` method chaining

JeanMertz:
This allows you to continue chaining method calls after calling `EntityCommands::entry`:

```rust
commands
    .entity(player.entity)
    .entry::<Level>()
    // Modify the component if it exists
    .and_modify(|mut lvl| lvl.0 += 1)
    // Otherwise insert a default value
    .or_insert(Level(0))
    // Return the EntityCommands for the entity
    .entity()
    // And continue chaining method calls
    .insert(Name("Player".to_string()));
```

### DIFF
--- a/crates/bevy_ecs/src/system/commands/mod.rs
+++ b/crates/bevy_ecs/src/system/commands/mod.rs
@@ -2169,6 +2169,38 @@ impl<'a, T: Component> EntityEntryCommands<'a, T> {
             .queue(entity_command::insert_from_world::<T>(InsertMode::Keep));
         self
     }
+
+    /// Get the [`EntityCommands`] from which the [`EntityEntryCommands`] was initiated.
+    ///
+    /// This allows you to continue chaining method calls after calling [`EntityCommands::entry`].
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// # use bevy_ecs::prelude::*;
+    /// # #[derive(Resource)]
+    /// # struct PlayerEntity { entity: Entity }
+    /// #[derive(Component)]
+    /// struct Level(u32);
+    ///
+    /// fn level_up_system(mut commands: Commands, player: Res<PlayerEntity>) {
+    ///     commands
+    ///         .entity(player.entity)
+    ///         .entry::<Level>()
+    ///         // Modify the component if it exists
+    ///         .and_modify(|mut lvl| lvl.0 += 1)
+    ///         // Otherwise insert a default value
+    ///         .or_insert(Level(0))
+    ///         // Return the EntityCommands for the entity
+    ///         .entity()
+    ///         // And continue chaining method calls
+    ///         .insert(Name("Player".to_string()));
+    /// }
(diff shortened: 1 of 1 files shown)


alice-i-cecile (approved):

chescock (approved):
Cool!  

There's also an `EntityWorldMut::entry()` method.  Do we want to add the same thing to its `Entry` type?

JeanMertz:
Actually, @chescock, that `Entry` implementation is fairly different than this one (and more closely matches Rust's `Entry' variant).



Specifically, the one I modified _always_ returns `&mut Self`, whereas in that one, _only_ `and_modify` returns `Self`, and `insert_entry` returns `OccupiedEntry`, which would still allow going back to the original `EntityWorldMut`. All the others return other types that no longer retain the required reference to `EntityWorldMut`.

So the only path you can take would be `world.entry::<T>.and_modify().world()` or `world.entry::<T>.insert_entry().world()` (assuming we'd name `entity()` from this impl to `world()`). All other paths won't work. There is no way for the "and modify or insert" path to have a reference back to `world`. So implementing this there seems mostly pointless.

Do you still want me to add the implementation, or leave it as-is?

chescock:
> Do you still want me to add the implementation, or leave it as-is?

Oh, this PR is great the way it is!  Sorry, I'm just brainstorming possible future work.  :)  

> Specifically, the one I modified _always_ returns `&mut Self`, whereas in that one, _only_ `and_modify` returns `Self`, and `insert_entry` returns `OccupiedEntry`, which would still allow going back to the original `EntityWorldMut`. All the others return other types that no longer retain the required reference to `EntityWorldMut`.

Which methods don't return an entry?  I'm looking at https://dev-docs.bevyengine.org/bevy/ecs/world/enum.Entry.html, and I see the two methods you listed plus three `or_` methods that all return `OccupiedEntry`.  I guess `OccupiedEntry::insert` doesn't return `self`, but you're unlikely to chain that after another `insert`.  

But I'm just arguing that it's possible, not that it's actually *useful*, so maybe it doesn't matter anyway.  :)

JeanMertz:
> Which methods don't return an entry?

It looks like this code changed since the last stable release. I was going off of https://docs.rs/bevy/latest/bevy/ecs/world/enum.Entry.html, but I should have verified in the code itself.

Let's to it like this: merge this PR if you're happy with it, and then I'll follow up with one for `world::Entry`.

JeanMertz:
@chescock I experimented with the implementation for `EntityMutWorld::entry()` [here](https://github.com/dcdpr/bevy/commit/c0e90a28cd3f3bf5ae3d1c64fcc0f46637dbf0c9), but I couldn't get the tests to pass, and after [discussing it on Discord](https://discord.com/channels/691052431525675048/749335865876021248/1334454104306618410), it turns out the implementation is unsound, so I won't be providing a PR for that one.

chescock:
Makes sense!  Sorry for sending you on a wild goose chase!

PR #17581, opened 2025-01-28
don't use bevy_pbr for base bevy_gizmos plugin

rambip:
# Objective

This PR enables `bevy_gizmos` to be used without `bevy_pbr`, for user who want to create their custom mesh rendering logic.

It can also be useful for user who just want to use bevy for drawing lines (why not).

This work is part of a bigger effort to make the bevy rendering pipeline more modular. I would like to contribute an exemple to render custom meshes without `bevy_pbr`. Something like [this](https://github.com/rambip/plant-mesh/blob/main/src/shader/mod.rs)

## Solution

Now, `bevy_pbr` is an optional dependency, and used only to debug lights.

I query the `ViewUniforms` manually, instead of using `bevy_pbr` to get the heavy `MeshViewLayout`

## Testing

I'm not used to testing with bevy at all, but I was able to use successfully in my project.
It might break for some different mesh pipelines, but I don't think so.

---

## Showcase

image
So nice ...

## Migration Guide

I don't think there is any breaking change

# Remaining work

Before merging it, it would be useful to:
- rewrite the `pipeline_2d.rs` logic to remove the `bevy_sprite` depedency too
- move `view.rs` to `bevy_render`, so that it can be used in a more modular way.
~~- include the most recent changes from 0.16~~


crates/bevy_gizmos/src/view.rs
```diff
@@ -0,0 +1,88 @@
...
+        RenderCommandResult::Success
+    }
+}
+
+pub(crate) fn prepare_view_bind_groups(
+    mut commands: Commands,
+    render_device: Res<RenderDevice>,
+    view_uniforms: Res<ViewUniforms>,
+    layout: Res<OnlyViewLayout>,
+    views: Query<Entity>,
```
IceSentry, line 62: This _needs_ a filter, this currently iterates on all entities in the render world.

crates/bevy_gizmos/src/view.rs
```diff
@@ -0,0 +1,88 @@
...
+) {
+    if let Some(view_binding) = view_uniforms.uniforms.binding() {
+        for entity in &views {
+            let entries = DynamicBindGroupEntries::new_with_indices(((0, view_binding.clone()),));
```
IceSentry, line 66: This can be moved to be before the loop.

crates/bevy_gizmos/src/view.rs
```diff
@@ -0,0 +1,88 @@
...
+}
+
+pub(crate) fn prepare_view_bind_groups(
+    mut commands: Commands,
+    render_device: Res<RenderDevice>,
+    view_uniforms: Res<ViewUniforms>,
+    layout: Res<OnlyViewLayout>,
+    views: Query<Entity>,
+) {
+    if let Some(view_binding) = view_uniforms.uniforms.binding() {
```
IceSentry, line 64: I'd prefer a let-else with early return but that's a tiny nitpick and purely optional.

crates/bevy_gizmos/src/view.rs
IceSentry:
```suggestion
    views: Query<Entity, With<ExtractedView>>,
```

IceSentry (changes requested):
I like where this is going, but needs a few changes.

rambip:
I did the same thing for `pipeline_2d`, pretty straightforward.

Now `bevy_sprite` is not a dependency anymore and `bevy_pbr` is optional.

Right now `pipeline3d` and `pipeline2d` look very similar, maybe they share some functionality that could be merged.

I would love `view.rs` to move to `bevy_render` so that I can fill a PR to rewrite the `custom_mesh_pipeline` example to work without `bevy_pbr`.

IceSentry:
I'd be more than happy to review and approve a PR that moves shared things from 2d/3d to bevy_render. We want to break appart bevy_render because it's giant, but moving shared things to it still makes sense.

I'd also be happy to review and approve a PR that moves the OnlyViewLayout thing to bevy render, but I think this could be done after this current PR is merged.

rambip:
Ok, I prepare a new PR for 
- 1) moving `OnlyViewLayout` to `bevy_render`
- 2) updating the example.

Do you know where in the `bevy_render` crate the view stuff should go ?

IceSentry:
Answered in discord to keep this discussion relevant to this PR

IceSentry (approved):

rambip:
Is there something missing to merge this issue ?

IceSentry:
A second review. Bevy needs 2 reviews per PR.

alice-i-cecile (approved):

alice-i-cecile:
Sorry @rambip; we needed a second reviewer. This looks sensible to me: if you get merge conflicts resolved I'll merge this ASAP.

atlv24 (approved):

PR #17586, opened 2025-01-28
Use more headers in AsBindGroup docs

IceSentry:
# Objective

- Linking to a specific AsBindGroup attribute is hard because it doesn't use any headers and all the docs is in a giant block

## Solution

- Make each attribute it's own sub-header so they can be easily linked

---

## Showcase

Here's what the rustdoc output looks like with this change

image

## Notes

I kept the bullet point so the text is still indented like before. Not sure if we should keep that or not

### DIFF
--- a/crates/bevy_render/src/render_resource/bind_group.rs
+++ b/crates/bevy_render/src/render_resource/bind_group.rs
@@ -179,6 +179,7 @@ impl Deref for BindGroup {
 /// | `visibility(...)`      | `all`, `none`, or a list-combination of `vertex`, `fragment`, `compute`                    | `compute`     |
 ///
 /// ## `sampler(BINDING_INDEX, arguments)`
+///
 /// * This field's [`Handle<Image>`](bevy_asset::Handle) will be used to look up the matching [`Sampler`] GPU
 ///     resource, which will be bound as a sampler in shaders. The field will be assumed to implement [`Into<Option<Handle<Image>>>`]. In practice,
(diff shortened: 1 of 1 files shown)


crates/bevy_render/src/render_resource/bind_group.rs
```diff
@@ -175,22 +178,24 @@ impl Deref for BindGroup {
...
-///     * The field's [`Handle<Storage>`](bevy_asset::Handle) will be used to look up the matching [`Buffer`] GPU resource, which
-///       will be bound as a storage buffer in shaders. If the `storage` attribute is used, the field is expected a raw
-///       buffer, and the buffer will be bound as a storage buffer in shaders.
-///     * It supports an optional `read_only` parameter. Defaults to false if not present.
+///
+/// ## `storage(BINDING_INDEX, arguments)`
+///
+/// * The field's [`Handle<Storage>`](bevy_asset::Handle) will be used to look up the matching [`Buffer`] GPU resource, which
+///   will be bound as a storage buffer in shaders. If the `storage` attribute is used, the field is expected a raw
+///   buffer, and the buffer will be bound as a storage buffer in shaders.
```
IceSentry, line 198: I made a tiny edit here to remove the last bullet point because it's covered by the table right under it so it was redundant.

mockersf (approved):

mockersf:
trivial as it's doc reorg for better readability

PR #17596, opened 2025-01-29
Fix Taffy viewport node leaks

### LINKED ISSUE #17594: Taffy viewport nodes leak
## Bevy version

main

## What you did

```
fn main() {
    App::new()
        .add_plugins(DefaultPlugins)
        .add_systems(Startup, setup)
        .add_systems(Update, update)
        .run();
}

    commands.spawn(Camera2d);
}

fn update(mut commands: Commands, node_query: Query<Entity, With<Node>>) {
    for entity in node_query.iter() {
    }

    commands.spawn(Node::default());
}
```

## What went wrong

The number of UI nodes stays the same but the number of taffy nodes increases by 1 every frame.

## Additional Information

For most UI node entities theres a 1-to-1 mapping from the entity to an associated taffy node ID. Root UI nodes are an exception though, their corresponding taffy node in the taffy tree is also given a parent that represents the viewport. These viewport taffy nodes are not removed when a root UI node is despawned.

Parenting of an existing root UI node with an associated viewport taffy node also results in the leak of the viewport node.

ickshonpe:
# Objective

For most UI node entities there's a 1-to-1 mapping from the entity to its associated Taffy node. Root UI nodes are an exception though, their corresponding Taffy node in the Taffy tree is also given a parent that represents the viewport. These viewport Taffy nodes are not removed when a root UI node is despawned.

Parenting of an existing root UI node with an associated viewport Taffy node also results in the leak of the viewport node.

These tests fail if added to the `layout` module's tests on the main branch:

```rust
    #[test]
    fn no_viewport_node_leak_on_root_despawned() {
        let (mut world, mut ui_schedule) = setup_ui_test_world();

        let ui_root_entity = world.spawn(Node::default()).id();

        // The UI schedule synchronizes Bevy UI's internal `TaffyTree` with the
        // main world's tree of `Node` entities.
        ui_schedule.run(&mut world);

        // Two taffy nodes are added to the internal `TaffyTree` for each root UI entity.
        // An implicit taffy node representing the viewport and a taffy node corresponding to the
        // root UI entity which is parented to the viewport taffy node.
        assert_eq!(
            world.resource_mut::<UiSurface>().taffy.total_node_count(),
            2
        );

        world.despawn(ui_root_entity);

        // The UI schedule removes both the taffy node corresponding to `ui_root_entity` and its
        // parent viewport node.
        ui_schedule.run(&mut world);

        // Both taffy nodes should now be removed from the internal `TaffyTree`
        assert_eq!(
            world.resource_mut::<UiSurface>().taffy.total_node_count(),
            0
        );
    }

    #[test]
    fn no_viewport_node_leak_on_parented_root() {
        let (mut world, mut ui_schedule) = setup_ui_test_world();

        let ui_root_entity_1 = world.spawn(Node::default()).id();
        let ui_root_entity_2 = world.spawn(Node::default()).id();

        ui_schedule.run(&mut world);

        // There are two UI root entities. Each root taffy node is given it's own viewport node parent,
        // so a total of four taffy nodes are added to the `TaffyTree` by the UI schedule.
        assert_eq!(
            world.resource_mut::<UiSurface>().taffy.total_node_count(),
            4
        );

        // Parent `ui_root_entity_2` onto `ui_root_entity_1` so now only `ui_root_entity_1` is a
        // UI root entity.
        world
            .entity_mut(ui_root_entity_1)
            .add_child(ui_root_entity_2);

        // Now there is only one root node so the second viewport node is removed by
        // the UI schedule.
        ui_schedule.run(&mut world);

        // There is only one viewport node now, so the `TaffyTree` contains 3 nodes in total.
        assert_eq!(
            world.resource_mut::<UiSurface>().taffy.total_node_count(),
            3
        );
    }
```

Fixes #17594

## Solution

Change the `UiSurface::entity_to_taffy` to map to `LayoutNode`s. A `LayoutNode` has a `viewport_id: Option<taffy::NodeId>`  field which is the id of the corresponding implicit "viewport" node if the node is a root UI node, otherwise it is `None`. When removing or parenting nodes this field is checked and the implicit viewport node is removed if present.

## Testing

There are two new tests in `bevy_ui::layout::tests` included with this PR:
* `no_viewport_node_leak_on_root_despawned`
* `no_viewport_node_leak_on_parented_root`

### DIFF
--- a/crates/bevy_ui/src/layout/ui_surface.rs
+++ b/crates/bevy_ui/src/layout/ui_surface.rs
@@ -21,9 +21,25 @@ pub struct RootNodePair {
     pub(super) user_root_node: taffy::NodeId,
 }
 
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+pub struct LayoutNode {
+    // Implicit "viewport" node if this `LayoutNode` corresponds to a root UI node entity
+    pub(super) viewport_id: Option<taffy::NodeId>,
+    // The id of the node in the taffy tree
+    pub(super) id: taffy::NodeId,
+}
+
+impl From<taffy::NodeId> for LayoutNode {
+    fn from(value: taffy::NodeId) -> Self {
+        LayoutNode {
+            viewport_id: None,
+            id: value,
+        }
+    }
+}
 #[derive(Resource)]
 pub struct UiSurface {
-    pub(super) entity_to_taffy: EntityHashMap<taffy::NodeId>,
+    pub(super) entity_to_taffy: EntityHashMap<LayoutNode>,
     pub(super) camera_entity_to_taffy: EntityHashMap<EntityHashMap<taffy::NodeId>>,
     pub(super) camera_roots: EntityHashMap<Vec<RootNodePair>>,
     pub(super) taffy: TaffyTree<NodeMeasure>,
@@ -77,25 +93,25 @@ impl UiSurface {
 
         match self.entity_to_taffy.entry(entity) {
             Entry::Occupied(entry) => {
-                let taffy_node_id = *entry.get();
+                let taffy_node = *entry.get();
                 let has_measure = if new_node_context.is_some() {
                     taffy
-                        .set_node_context(taffy_node_id, new_node_context)
+                        .set_node_context(taffy_node.id, new_node_context)
                         .unwrap();
                     true
                 } else {
-                    taffy.get_node_context(taffy_node_id).is_some()
+                    taffy.get_node_context(taffy_node.id).is_some()
                 };
 
                 taffy
                     .set_style(
-                        taffy_node_id,
+                        taffy_node.id,
                         convert::from_node(node, layout_context, has_measure),
                     )
                     .unwrap();
             }
             Entry::Vacant(entry) => {
-                let taffy_node_id = if let Some(measure) = new_node_context.take() {
+                let taffy_node = if let Some(measure) = new_node_context.take() {
                     taffy.new_leaf_with_context(
                         convert::from_node(node, layout_context, true),
                         measure,
@@ -103,7 +119,7 @@ impl UiSurface {
                 } else {
                     taffy.new_leaf(convert::from_node(node, layout_context, false))
                 };
-                entry.insert(taffy_node_id.unwrap());
+                entry.insert(taffy_node.unwrap().into());
             }
         }
     }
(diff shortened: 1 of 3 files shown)


crates/bevy_ui/src/layout/ui_surface.rs
```diff
@@ -265,11 +289,16 @@ impl UiSurface {
...
+    /// Removes each entity from the internal map and then removes their associated nodes from taffy
     pub fn remove_entities(&mut self, entities: impl IntoIterator<Item = Entity>) {
         for entity in entities {
             if let Some(node) = self.entity_to_taffy.remove(&entity) {
-                self.taffy.remove(node).unwrap();
+                self.taffy.remove(node.id).unwrap();
+                if let Some(viewport_node) = node.viewport_id {
+                    if self.taffy.get_node_context(viewport_node).is_some() {
+                        self.taffy.remove(viewport_node).ok();
+                    }
```
ickshonpe, line 300: This feels silly but I couldn't find a simpler way to check for existence before removal.

Carter0 (approved):

crates/bevy_ui/src/layout/ui_surface.rs
UkoeHB: Maybe `let _ = self.taffy.remove(viewport_node);`?

UkoeHB (approved):

crates/bevy_ui/src/layout/ui_surface.rs
```diff
@@ -21,9 +21,26 @@ pub struct RootNodePair {
     pub(super) user_root_node: taffy::NodeId,
 }
 
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+pub struct LayoutNode {
+    // Implicit "viewport" node if this `LayoutNode` corresponds to a root UI node entity
+    pub(super) viewport_id: Option<taffy::NodeId>,
```
mockersf, line 27: adding it here means that it will allocate memory for each entry even though only few of them will be actually filled

mockersf: that's an additional `u64` for every UI entity. not sure if it's a problem or not

crates/bevy_ui/src/layout/ui_surface.rs
ickshonpe: The problem is  thateven though `remove` returns a result it doesn't return `Err` on failure, it just panics

crates/bevy_ui/src/layout/ui_surface.rs
ickshonpe: We have really fat UI nodes already so it shouldn't make much difference. I really didn't want to add any more complexity to these modules in this PR and this was the solution I came up with that required the fewest changes to the exisiting code.

mockersf (approved):

PR #17600, opened 2025-01-29
Implementing Reflect on *MeshBuilder types

eckz:
# Objective

- Most of the `*MeshBuilder` classes are not implementing `Reflect`

## Solution

- Implementing `Reflect` for all `*MeshBuilder` were is possible.
- Make sure all `*MeshBuilder` implements `Default`.
- Adding new `MeshBuildersPlugin` that registers all `*MeshBuilder` types.

## Testing

- `cargo run -p ci`
- Tested some examples like `3d_scene` just in case something was broken.

### DIFF
--- a/crates/bevy_mesh/src/primitives/dim2.rs
+++ b/crates/bevy_mesh/src/primitives/dim2.rs
@@ -12,10 +12,12 @@ use bevy_math::{
     },
     FloatExt, Vec2,
 };
+use bevy_reflect::prelude::*;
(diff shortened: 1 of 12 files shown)


alice-i-cecile:
Interesting; why do you need Reflect on these? I didn't expect these types to be persisted much of anywhere.

eckz:
The usecase was editing `Meshable` items, like cubes planes, etc, directly using `bevy_egui_inspector`.

alice-i-cecile (approved):

PR #17601, opened 2025-01-29
Fixes #17508: bevy_color::Color constructor docs get docs matching underlying constructor

### LINKED ISSUE #17508: Color docs should specify acceptable value ranges
https://docs.rs/bevy/latest/bevy/prelude/enum.Color.html

Most of the methods to construct a `Color` accept values in the range `[0..1]`. A common beginner mistake is assuming the range `[0..255]`. Hue is typically in the range `[0..360]`. LABA/LCHA have a Lightness in the range `[0..1.5]`. I might have missed some exceptions. 

These ranges are documented for the color-space structs being constructed by each method, but IMO should be documented in the method, too.

janis-bhm:
# Objective

Fixes #17508

`bevy_color::Color` constructors don't have docs explaining the valid range for the values passed.

## Solution

I've mostly copied the docs from the respective underlying type's docs, because that seemed most consistent and accurate.

### DIFF
--- a/crates/bevy_color/src/color.rs
+++ b/crates/bevy_color/src/color.rs
@@ -85,6 +85,13 @@ impl Color {
     }
 
     /// Creates a new [`Color`] object storing a [`Srgba`] color.
+    ///
+    /// # Arguments
+    ///
+    /// * `red` - Red channel. [0.0, 1.0]
+    /// * `green` - Green channel. [0.0, 1.0]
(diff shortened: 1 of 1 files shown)


Carter0 (approved):

alice-i-cecile (approved):
I will note that sometimes these values are used outside of their standard range (bloom! HDR!), but this is still very useful to have.

PR #17603, opened 2025-01-29
Remove IDE gitignores 

SpecificProtagonist:
Remove gitignores for IDE files, recommend how to set up local gitignore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
-# Use .git/info/exclude to ignore files related to your IDE
+# If your IDE needs additional project specific files, configure git to ignore them:
+# https://docs.github.com/en/get-started/getting-started-with-git/ignoring-files#configuring-ignored-files-for-all-repositories-on-your-computer
 
 # Rust build artifacts
 target
@@ -6,6 +7,10 @@ crates/**/target
 benches/**/target
 tools/**/target
 **/*.rs.bk
+
+# DX12 wgpu backend
+dxcompiler.dll
+dxil.dll
 
 # Cargo
 Cargo.lock
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,5 @@
+# Use .git/info/exclude to ignore files related to your IDE
+
 # Rust build artifacts
 target
 crates/**/target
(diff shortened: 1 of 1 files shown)


IceSentry:
These kinds of ignores should be in your own local .gitignore. 

With that said, I'm surprised those other editor specific ones are already in there because in the past that was refused.

GitGhillie (approved):

aloucks:
This looks like a recipe for seeing PRs come in with IDE files. What is the rational behind requiring contributors to modify `.git/info/exclude` after cloning? How is this beneficial? Adding IDE patterns in `.gitignore` is extremely common.

.gitignore
```diff
@@ -1,3 +1,5 @@
+# Use .git/info/exclude to ignore files related to your IDE
```
mockersf, line 1: Usual recommendation is to add your IDE to your computer gitignore, not to this file on each repo
Maybe a link to https://docs.github.com/en/get-started/getting-started-with-git/ignoring-files#configuring-ignored-files-for-all-repositories-on-your-computer would give more informations

mockersf:
> This looks like a recipe for seeing PRs come in with IDE files.

It's already what we have. Well not that often


> Adding IDE patterns in `.gitignore` is extremely common.

You should add *your* IDE to *your* global gitignore file, not on each project

.gitignore
```diff
@@ -10,13 +12,6 @@ Cargo.lock
 .cargo/config
 .cargo/config.toml
 
-# IDE files
-.idea
-.vscode
-.zed
-dxcompiler.dll
```
IceSentry, line 17: The dx ignore should still be kept. They aren't IDE files, they are pretty specific to using bevy with a DX12 backend so I think it make sense to keep them[

aloucks:
> > This looks like a recipe for seeing PRs come in with IDE files.
> 
> It's already what we have. Well not that often
> 
> > Adding IDE patterns in `.gitignore` is extremely common.
> 
> You should add _your_ IDE to _your_ global gitignore file, not on each project

It just seems like a change opens up a lot of potential headaches. Adding the IDE details to `.gitignore` is a precaution. When you remove this, you end up with PRs including those files. You ask the author to remove them and end up with them deleting them in a new commit, but the history remains. This results in asking them to squash it out... or no one notices and this garbage history gets introduced. Why open up this potential back-and-forth and friction?

IceSentry:
We always squash all PRs

IceSentry:
We can't have ignores for all editors and people should learn about the existence of the global gitignore and use it. We haven't had any editor specific ignores for a while and it wasn't an issue. The current vscode and idea additions are relatively recent.

alice-i-cecile (approved):

Issue #1592, opened 2021-03-07
Entity groups


alice-i-cecile:
## What problem does this solve or what need does it fill?

In non-trivial games, entities commonly need to be bundled together to share common behavior, despite being mostly heterogenous. As a few concrete examples:
- character models with multiple moving parts
- visual effects associated with sprites
- UI widgets composed with multiple parts, each of which has its own transform, style, sprite etc

## Possible Solutions

1. Allowing multiple copies of each component per entity (like nodes in Godot) is a Pandora's box of special-casing and is likely to do horrible things to our storages and conceptual clarity.
2. Using a parent-child-like hierarchy tree and traversing it to apply behavior. This results in a lot of wasted recomputation.
3. Storing a "WidgetID" component (or analogous) on each entity, then coordinating their behavior by looking it up within each relevant system.
4. Adding a unique marker component for each entity in the entity group, then querying for it. We generally can't define these at run time, since the number of entity groups is unknown, so min const generics might work well here.

Solutions 1 and 2 are clearly not adequate. Solution 3 is probably fine, but can result in broad queries and doesn't leverage the ECS super well (e.g. for data fragmentation).

Regardless of the solution that we reach consensus on as best-practice, we should make sure it's properly supported and has good examples. These patterns are surprisingly tricky, and I've seen various complaints about how hard these designs are to implement in ECS.

## Additional context

This challenge / pattern was brought up in the context of [widgets](https://discord.com/channels/691052431525675048/743663673393938453/818194214751830016), prompting this issue.

The pattern selected here should be used when designing our UI framework: see #254.

alice-i-cecile:
Solution 3 could be implemented with / improved by the use of index (see #1587, #1205).

cart:
Its worth pointing out that (4) is similar to the "entity relationships" that [flecs](https://github.com/SanderMertens/flecs) has, which is similar to Tags, as defined here #1527

Davier:
What I've been doing in my widgets prototype is to store the entity id of the "widget root" inside the other entities of the widget, and also the reverse when needed (saving the entity id of a "widget child" in the root to make it accessible to all). It feels a little bloated but it works well.

alice-i-cecile:
From @davier: Solution 3 can be naturally implemented using scenes, with the scene instance ID serving as the `WidgetId`.

GuillaumeGomez:
[...]

Issue #17564, opened 2025-01-27
Ensure that bevy_ecs can handle large numbers of archetypes


alice-i-cecile:
As discussed in https://github.com/bevyengine/rfcs/pull/79 and laid out by @james-j-obrien, Bevy's ECS architecture is not optimized for use in worlds with a very large number of archetypes. Fundamentally:

> `bevy_ecs` currently operates under the assumption that archetype and component ids are dense and strictly increasing.

While this is a blocker for efficient implementation of fragmenting / archetypal relations, it's also worth exploring and fixing in its own right.

There are several steps to this:

- [] Remove the sparse set in table storage https://github.com/bevyengine/bevy/pull/14928
- [] Replace the FixedBitSet in Access / FilteredAccess https://github.com/bevyengine/bevy/pull/16784
- [] Implement a component index, which tracks the tables and archetypes each `ComponentId` belongs to as well as the column in that table the id inhabits (if any)
- [] Query and system caches need to be easier to access.
- [] Query and system caches must be updated to reflect changes to archetypes
- [] Create and expose tools for deleting component ids
- [] Create and expose tools for deleting archetypes

This is a tracking issue: as more related work comes along, please link it here and bother Alice to update this issue description.

cart:
Might be worth reframing this / renaming it. The "dense and strictly increasing" behavior works great (essentially optimally) with a large number of archetypes, provided they are all being used. This is about adapting Bevy ECS to dataflow where the set of archetypes is constantly changing.

chescock:
#16885 may also be helpful.  It would let us get rid of `ArchetypeComponentId`, which grows and changes even faster than `ArchetypeId` itself!

PR #17608, opened 2025-01-30
Query by Indexed Component Value using Fragmenting Marker Components

### LINKED ISSUE #4513: Indexes:

An "index" is a way to quickly find entries based on their value.
In the context of an ECS, indexes would let us e.g. "find all entities with components of this enum value", identify all entities inside a spatial region using a quadtree on `Transform`, quickly look up entities by their `Name`, or filter entities to find all components whose value lies within a range. 

Provide a native, efficient, fool-proof indexing solution.
Ideally, this should be user-extensible to account for other forms of acceleration data-structure.
This will be naturally coupled with change detection to ensure efficient updating.

The design of this is rather challenging, and requires a full RFC.

Linear time search will always work... slowly. Parallel query tools will help though.

Users can hand-roll this themselves right now, using a resource and `Query::get`. This is not particularly ergonomic, requires writing a lot of boilerplate, forces users to duplicate the API work for each project, and forces them to reckon with the perils of synchronization alone.

## Additional context

May be useful for #3742. Would be useful for #1470.

### LINKED ISSUE #5639: Query filter for component value rather than component existence

I'm working on a game that uses a lot of different 'status effects'. These status effects are presently implemented on each character as an inserted/removed component, but I am interested in exploring these with always-inserted bool fields instead (akin to `Visibility`). The main issue I have is that I won't be able to filter out entities as easily as I can presently (like finding all poisoned entities by using `With<Poisoned>`).

A Query Filter pair akin to `With`/`Without`, that instead of checking for component existence, checks for component _value_. Given we don't have full const generics yet, this should probably be limited to `true`/`false` for simplicity. Alternatively, the current value could be compared to the `Default` value.

The name for these filters? Unsure. Maybe `Truthy`/`Falsey` or `Enabled`/`Disabled`?

### Illustration

Today:
```rust
#[derive(Component)]
struct Poisoned;
fn all_poisoned(entities: Query<&Name, Without<Poisoned>>) {
    // ...
}
```

New query filter:
```rust
#[derive(Component)]
struct Poisoned(bool);
// We will need some machinery like Deref or Into to extract the current value, but I'll leave that for discussion.
fn all_poisoned(entities: Query<&Name, Enabled<Poisoned>>) {
}
```

Filtering the way that the `Visibility` struct does presently. It's not the best ergonomics, though.

## Additional context

Related to the discussion found in #3796.

### LINKED ISSUE #6556: Entity Markers

A Marker trait that can mark an entity. example

```rust
#[derive(Marker)]
pub enum Players {
    Red,
    Blue,
    Green,
}
```

then we can assign an entity with a marker
```rust
commands.spawn(PlayerBundle::default()).mark(Players::Red);
```

and we can easily query players
```rust
fn add_items_to_red_and_blue(red_query: Query<Player, Or<MarkedAs<Players::Red>>, MarkedAs<Players::Blue>>>) {
    // ...
}
```

An Marker that works like a Resource but holds a EntityID and the Marker can be used in Queries.

Manually make Resources holding EntityIDs and then using a normal query and using .get with the entityID from the resource.

bushrat011899:
# Objective

- Fixes #4513
- Possible fix for #1592
- Fixes #6556
- Fixes #5639
- Add a first-class option for querying by component _value_.

## Solution

Users can now call `app.add_index(...)` for an appropriate component `C` to have Bevy automatically track its value and provide a high-performance archetypical indexing solution. Users can then query by-value using the `QueryByIndex` system parameter all entities with a specific indexed value. The value is provided within the body of a system, meaning `QueryByIndex` can access any entity matching the provided compile-time query filters.

This is achieved by adding `OnInsert` and `OnReplace` observers and requiring that the index component `C` be immutable, ensuring those two hooks are able to capture all mutations. Since this is done with observers, users can index 3rd party components, provided they are immutable and implement the traits required by the index storage back-end chosen.

By default, indexing uses a `HashMap` as its backing storage, which requires that the indexed component implements `Eq + Hash + Clone`. If these traits aren't available on the component to be indexed, users can either use `BTreeMap` as a back-end, only requiring `Ord + Clone`, or implement the `IndexStorage` trait on their own type. Any storage back-end can be used with an index, but only _one_ index will ever exist for a given component.

Within the observers, a private index resource is managed which tracks the component value against a collection of runtime marker components. These runtime marker components are used to group entities by-value at the archetype level for the index component. By using multiple marking components, their combination can be used to uniquely address a particular component value, allowing a relatively small number of markers to totally service even complex components.

This grouping allows for maximum performance in iteration, since querying is simply filtering for a hidden component by-archetype. 

## To-Do / Follow-up

- It would be nice to index against a whole _bundle_ instead. This requires an immutable-bundle concept and for the user to provide a function to create their derived index from the supplied bundle.

## Testing

- New `index` example
- CI

## Performance

I've added some `index` benchmarks to the `ecs` bench which test _iteration_ speed (`index_iter`) and _mutation_ speed (`index_update`). Running these benchmarks locally, I see a large performance improvement (20x) when looking for rare values amongst a large selection of entities (1-in-1,000 amongst 1,000,000), and a notable performance loss (6x) when mutating indexed components. This confirms expectation: indexing is only a performance win in scenarios where a value is rarely updated but frequently searched for.

## Indexing Strategy

To find an `Entity` with a particular value of a component `C`, a first choice might just be to use `filter` on a `Query`. No setup required, no cache invalidation, nice and simple.

Untitled

Big downside to this approach is you need to iterate _all_ entities matching your `Query`. If _most_ entities match your filter, this may be perfectly fine, but if you have a large number of entities to search, and only a small number with the value you're searching for, it quickly becomes too costly.

First optimisation available: store a mapping from `C` to `Entity` and use `Query::get`.

Untitled

This avoids iterating all entities, but introduces some new problems to solve. First, multiple entities could have the same value for `C`, so we need a mapping from `C` to many entities such as an `EntityHashSet`. Next, iteration performance suffers as we're no longer iterating over entities and tables in-order; we're randomly selecting entities based on however they're stored in our index. For the best performance, we want to iterate the `Query` and _check_ against the index, just "somehow" skipping entities that don't match our value.

Second optimisation: use marker components to group entities with like `C` values.

Untitled-1

What we do here is we instead map values of `C` to `ComponentId`s rather than `Entity`s, and then ensure every entity with a `C` has the required marker component for that value _and only that marker component_. This fragments the archetypes on the value of `C`, grouping all entities with the `C` component by its value, giving us maximum iteration performance. This isn't free however. First, changing the value of `C` now requires an archetype move. Secondly, we need 1 `ComponentId` per unique value of `C` currently in the `World`. Since there are only `usize` `ComponentId`s, that limits us to 2^64 unique values _across all indexes_ (on 64-bit platforms, even worse on 32-bit). We _could_ just hope users don't have too many unique values in the index, but this is an easy mistake to make. Indexing a component that internally stores just _2_ `usize`s would allow for the index to totally exhaust all `ComponentId`s.

Final optimisation: use the _absence_ of a marker component to contain information. Instead of marking all entities of a certain value with a _single_ marker component, we could instead encode a binary address using `With` and `Without` of several marker components to build a unique filter. This allows us to address 2^64 unique values with only 64 unique marker components, exponentially fewer components required.

Untitled

This still gives us fragmentation on `C` values for dense iteration, at the expense of a slightly more complex `QueryFilter`, since we need to _exclude_ and _include_ some combination of markers. This is the technique adopted by this PR.

---

## Release Notes

Users can now query by-value using indexes. To setup an index, first create an immutable component with `Clone`, `Eq`, and `Hash` implementations:

```rust
#[derive(Component, PartialEq, Eq, Hash, Clone)]
#[component(immutable)]
struct TilePosition {
    x: i32,
    y: i32,
}
```

Next, request that an index be managed for this component:

```rust
app.add_index(IndexOptions::<TilePosition>::default());
```

Finally, you can query by-value using the new `QueryByIndex` system parameter:

```rust
fn query_for_tiles(mut query: QueryByIndex<TilePosition, (Entity, &TilePosition)>) {
    let mut lens = query.at(&TilePosition { x: 1, y: 1 });
    let count = lens.query().iter().count();

    println!("Found {count} at (1,1)!");
}
```

`QueryByIndex` is just like `Query` except it has a third generic argument, `C`, which denotes the component you'd like to use as the index:

```rust
// Typical query
query: Query<&Foo, With<Bar>>,

// Indexed query
query: QueryByIndex<MyIndex, &Foo, With<Bar>>,
```

Internally, the index will move entities with differing values into their own archetypes, allowing for fast iteration at the cost of some added overhead when modifying the indexed value. As such, indexes are best used where you have an infrequently changing component that you frequently need to access a subset of.

Additional control around _how_ a component is indexed is provided by the `IndexOptions` argument passed to `add_index`. This allows controlling:
* the storage type of the marker components (`SparseSet` or `Table`),
* the address space used, limiting the number of marker components allocated
* the storage back-end used to map component values to a unique ID number

```rust
#[derive(Component, PartialEq, Eq, PartialOrd, Ord, Clone)]
#[component(immutable)]
enum Planet {
    Mercury,
    Venus,
    Earth,
    Mars,
    Jupiter,
    Saturn,
    Uranus,
    Neptune,
}

app.add_index(IndexOptions::<Planet, _> {
    // SparseSet may give better performance if an indexed
    // value frequently changes.
    marker_storage: StorageType::Table,

    // The above Planet enum only has 8 possible values,
    // so only 4 bits are required instead of the implied 8.
    // By default, the index will use `size_of` to determine the
    // number of bits required, which can be excessive for certain types.
    address_space: 4,

    // An alternate storage back-end may be more appropriate
    // for your index component, always benchmark to confirm!
    index_storage: BTreeMap::<Planet, usize>::new(),

    ..default()
});
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2044,6 +2044,7 @@ wasm = false
 name = "index"
 path = "examples/ecs/index.rs"
 doc-scrape-examples = true
+required-features = ["bevy_dev_tools"]
 
 [package.metadata.example.index]
 name = "Indexes"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2050,7 +2050,7 @@ required-features = ["bevy_dev_tools"]
 name = "Indexes"
 description = "Demonstrates querying by component value using indexing"
 category = "ECS (Entity Component System)"
-wasm = false
+wasm = true
 
 [[example]]
 name = "iter_combinations"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2041,6 +2041,17 @@ category = "ECS (Entity Component System)"
 wasm = false
 
 [[example]]
+name = "index"
+path = "examples/ecs/index.rs"
+doc-scrape-examples = true
+
+[package.metadata.example.index]
+name = "Indexes"
+description = "Demonstrates the creation and utility of indexes"
+category = "ECS (Entity Component System)"
+wasm = false
+
+[[example]]
 name = "iter_combinations"
 path = "examples/ecs/iter_combinations.rs"
 doc-scrape-examples = true
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2047,7 +2047,7 @@ doc-scrape-examples = true
 
 [package.metadata.example.index]
 name = "Indexes"
-description = "Demonstrates the creation and utility of indexes"
+description = "Demonstrates querying by component value using indexing"
 category = "ECS (Entity Component System)"
 wasm = false
 
--- a/benches/benches/bevy_ecs/index/index_iter_indexed.rs
+++ b/benches/benches/bevy_ecs/index/index_iter_indexed.rs
@@ -1,0 +1,38 @@
+use bevy_ecs::{prelude::*, system::SystemId};
+use criterion::black_box;
+use glam::*;
+
+const PLANETS: u8 = 16;
+const SPAWNS: usize = 10_000;
+
+#[derive(Component, Copy, Clone, PartialEq, Eq, Hash)]
+#[component(immutable)]
+struct Planet(u8);
+
+fn find_planet_zeroes_indexed(mut query: QueryByIndex<Planet, &Planet>) {
+    let query = query.at(&Planet(0));
+    for planet in query.iter() {
+        let _ = black_box(planet);
+    }
+}
+
+pub struct Benchmark(World, SystemId);
+
+impl Benchmark {
+    pub fn new() -> Self {
+        let mut world = World::new();
+
+        world.add_index::<Planet>();
+
+        world.spawn_batch((0..PLANETS).map(Planet).cycle().take(SPAWNS));
+
+        let id = world.register_system(find_planet_zeroes_indexed);
+
+        Self(world, id)
+    }
+
+    #[inline(never)]
+    pub fn run(&mut self) {
+        let _ = self.0.run_system(self.1);
+    }
+}
--- a/benches/benches/bevy_ecs/index/index_iter_indexed.rs
+++ b/benches/benches/bevy_ecs/index/index_iter_indexed.rs
@@ -9,7 +9,7 @@ const SPAWNS: usize = 1_000_000;
 #[component(immutable)]
 struct Planet(u16);
 
-fn find_planet_zeroes_indexed(mut query: QueryByIndex<Planet, &Planet>) {
+fn find_planet_zeroes_indexed(query: QueryByIndex<Planet, &Planet>) {
     let mut query = query.at(&Planet(0));
     for planet in query.query().iter() {
         let _ = black_box(planet);
--- a/benches/benches/bevy_ecs/index/index_iter_indexed.rs
+++ b/benches/benches/bevy_ecs/index/index_iter_indexed.rs
@@ -10,8 +10,8 @@ const SPAWNS: usize = 1_000_000;
 struct Planet(u16);
 
 fn find_planet_zeroes_indexed(mut query: QueryByIndex<Planet, &Planet>) {
-    let query = query.at_mut(&Planet(0));
-    for planet in query.iter() {
+    let mut query = query.at(&Planet(0));
+    for planet in query.query().iter() {
         let _ = black_box(planet);
     }
 }
--- a/benches/benches/bevy_ecs/index/index_iter_indexed.rs
+++ b/benches/benches/bevy_ecs/index/index_iter_indexed.rs
@@ -1,5 +1,5 @@
 use bevy_ecs::{prelude::*, system::SystemId};
-use criterion::black_box;
+use core::hint::black_box;
 use glam::*;
 
 const PLANETS: u8 = 16;
--- a/benches/benches/bevy_ecs/index/index_iter_indexed.rs
+++ b/benches/benches/bevy_ecs/index/index_iter_indexed.rs
@@ -22,7 +22,7 @@ impl Benchmark {
     pub fn new() -> Self {
         let mut world = World::new();
 
-        world.add_index::<Planet>();
+        world.add_index(IndexOptions::<Planet>::default());
 
         world.spawn_batch((0..PLANETS).map(Planet).cycle().take(SPAWNS));
 
--- a/benches/benches/bevy_ecs/index/index_iter_indexed.rs
+++ b/benches/benches/bevy_ecs/index/index_iter_indexed.rs
@@ -2,12 +2,12 @@ use bevy_ecs::{prelude::*, system::SystemId};
 use core::hint::black_box;
 use glam::*;
 
-const PLANETS: u8 = 16;
-const SPAWNS: usize = 10_000;
+const PLANETS: u16 = 1_000;
+const SPAWNS: usize = 1_000_000;
 
 #[derive(Component, Copy, Clone, PartialEq, Eq, Hash)]
 #[component(immutable)]
-struct Planet(u8);
+struct Planet(u16);
 
 fn find_planet_zeroes_indexed(mut query: QueryByIndex<Planet, &Planet>) {
     let query = query.at(&Planet(0));
--- a/benches/benches/bevy_ecs/index/index_iter_indexed.rs
+++ b/benches/benches/bevy_ecs/index/index_iter_indexed.rs
@@ -10,7 +10,7 @@ const SPAWNS: usize = 1_000_000;
 struct Planet(u16);
 
 fn find_planet_zeroes_indexed(mut query: QueryByIndex<Planet, &Planet>) {
-    let query = query.at(&Planet(0));
+    let query = query.at_mut(&Planet(0));
     for planet in query.iter() {
         let _ = black_box(planet);
     }
--- a/benches/benches/bevy_ecs/index/index_iter_naive.rs
+++ b/benches/benches/bevy_ecs/index/index_iter_naive.rs
@@ -2,12 +2,12 @@ use bevy_ecs::{prelude::*, system::SystemId};
 use core::hint::black_box;
 use glam::*;
 
-const PLANETS: u8 = 16;
-const SPAWNS: usize = 10_000;
+const PLANETS: u16 = 1_000;
+const SPAWNS: usize = 1_000_000;
 
 #[derive(Component, Copy, Clone, PartialEq, Eq, Hash)]
 #[component(immutable)]
-struct Planet(u8);
+struct Planet(u16);
 
 fn find_planet_zeroes_naive(query: Query<&Planet>) {
     for planet in query.iter().filter(|&&planet| planet == Planet(0)) {
--- a/benches/benches/bevy_ecs/index/index_iter_naive.rs
+++ b/benches/benches/bevy_ecs/index/index_iter_naive.rs
@@ -1,0 +1,35 @@
+use bevy_ecs::{prelude::*, system::SystemId};
+use criterion::black_box;
+use glam::*;
+
+const PLANETS: u8 = 16;
+const SPAWNS: usize = 10_000;
+
+#[derive(Component, Copy, Clone, PartialEq, Eq, Hash)]
+#[component(immutable)]
+struct Planet(u8);
+
+fn find_planet_zeroes_naive(query: Query<&Planet>) {
+    for planet in query.iter().filter(|&&planet| planet == Planet(0)) {
+        let _ = black_box(planet);
+    }
+}
+
+pub struct Benchmark(World, SystemId);
+
+impl Benchmark {
+    pub fn new() -> Self {
+        let mut world = World::new();
+
+        world.spawn_batch((0..PLANETS).map(Planet).cycle().take(SPAWNS));
+
+        let id = world.register_system(find_planet_zeroes_naive);
+
+        Self(world, id)
+    }
+
+    #[inline(never)]
+    pub fn run(&mut self) {
+        let _ = self.0.run_system(self.1);
+    }
+}
--- a/benches/benches/bevy_ecs/index/index_iter_naive.rs
+++ b/benches/benches/bevy_ecs/index/index_iter_naive.rs
@@ -1,5 +1,5 @@
 use bevy_ecs::{prelude::*, system::SystemId};
-use criterion::black_box;
+use core::hint::black_box;
 use glam::*;
 
 const PLANETS: u8 = 16;
--- a/benches/benches/bevy_ecs/index/mod.rs
+++ b/benches/benches/bevy_ecs/index/mod.rs
@@ -1,0 +1,38 @@
+mod index_iter_indexed;
+mod index_iter_naive;
+mod index_update_indexed;
+mod index_update_naive;
+
+use criterion::{criterion_group, Criterion};
+
+criterion_group!(benches, index_iter, index_update,);
+
+fn index_iter(c: &mut Criterion) {
+    let mut group = c.benchmark_group("index_iter");
+    group.warm_up_time(core::time::Duration::from_millis(500));
+    group.measurement_time(core::time::Duration::from_secs(4));
+    group.bench_function("naive", |b| {
+        let mut bench = index_iter_naive::Benchmark::new();
+        b.iter(move || bench.run());
+    });
+    group.bench_function("indexed", |b| {
+        let mut bench = index_iter_indexed::Benchmark::new();
+        b.iter(move || bench.run());
+    });
+    group.finish();
+}
+
+fn index_update(c: &mut Criterion) {
+    let mut group = c.benchmark_group("index_update");
+    group.warm_up_time(core::time::Duration::from_millis(500));
+    group.measurement_time(core::time::Duration::from_secs(4));
+    group.bench_function("naive", |b| {
+        let mut bench = index_update_naive::Benchmark::new();
+        b.iter(move || bench.run());
+    });
+    group.bench_function("indexed", |b| {
+        let mut bench = index_update_indexed::Benchmark::new();
+        b.iter(move || bench.run());
+    });
+    group.finish();
+}
--- a/crates/bevy_app/src/app.rs
+++ b/crates/bevy_app/src/app.rs
@@ -1351,7 +1351,7 @@ impl App {
     /// app.add_index::<FavoriteColor>();
     ///
     /// fn find_red_fans(mut query: QueryByIndex<FavoriteColor, Entity>) {
-    ///     for entity in query.at(FavoriteColor::Red).iter() {
+    ///     for entity in query.at(&FavoriteColor::Red).iter() {
     ///         println!("{entity:?} likes the color Red!");
     ///     }
     /// }
--- a/crates/bevy_app/src/app.rs
+++ b/crates/bevy_app/src/app.rs
@@ -18,7 +18,7 @@ use bevy_ecs::{
     system::{IntoObserverSystem, SystemId, SystemInput},
 };
 use bevy_platform_support::collections::HashMap;
-use core::{fmt::Debug, hash::Hash, num::NonZero, panic::AssertUnwindSafe};
+use core::{fmt::Debug, num::NonZero, panic::AssertUnwindSafe};
 use log::debug;
 use thiserror::Error;
 
@@ -1460,7 +1460,7 @@ impl Termination for AppExit {
 
 #[cfg(test)]
 mod tests {
-    use core::{iter, marker::PhantomData};
+    use core::{iter, marker::PhantomData, hash::Hash};
     use std::sync::Mutex;
 
     use bevy_ecs::{
--- a/crates/bevy_app/src/app.rs
+++ b/crates/bevy_app/src/app.rs
@@ -1331,7 +1331,7 @@ impl App {
     /// This allows querying by component _value_ to be very performant, at the expense of a small
     /// amount of overhead when modifying the indexed component.
     ///
-    /// See [`IndexComponent`] for further details.
+    /// See [`IndexableComponent`] for further details.
     ///
     /// # Examples
     ///
--- a/crates/bevy_app/src/app.rs
+++ b/crates/bevy_app/src/app.rs
@@ -1331,55 +1331,14 @@ impl App {
     /// This allows querying by component _value_ to be very performant, at the expense of a small
     /// amount of overhead when modifying the indexed component.
     ///
-    /// See also [`add_index_with_options`](App::add_index_with_options).
+    /// The [options](IndexOptions) provided allow control over how `C` is indexed, which may
+    /// allow you to improve the ergonomics or performance of an index over recommended defaults.
     ///
     /// # Examples
     ///
     /// ```rust
     /// # use bevy_app::prelude::*;
     /// # use bevy_ecs::prelude::*;
-    /// #
-    /// # let mut app = App::new();
-    /// #[derive(Component, PartialEq, Eq, Hash, Clone)]
-    /// #[component(immutable)]
-    /// enum FavoriteColor {
-    ///     Red,
-    ///     Green,
-    ///     Blue,
-    /// }
-    ///
-    /// app.add_index::<FavoriteColor>();
-    ///
-    /// fn find_red_fans(mut query: QueryByIndex<FavoriteColor, Entity>) {
-    ///     let mut lens = query.at(&FavoriteColor::Red);
-    ///     for entity in lens.query().iter() {
-    ///         println!("{entity:?} likes the color Red!");
-    ///     }
-    /// }
-    /// # app.add_systems(Update, find_red_fans);
-    /// ```
-    pub fn add_index<C: Component<Mutability = Immutable> + Eq + Hash + Clone>(
-        &mut self,
-    ) -> &mut Self {
-        self.world_mut().add_index::<C>();
-        self
-    }
-
-    /// Creates and maintains an index for the provided immutable [`Component`] `C` using observers.
-    ///
-    /// This allows querying by component _value_ to be very performant, at the expense of a small
-    /// amount of overhead when modifying the indexed component.
-    ///
-    /// The [options](IndexOptions) provided allow finer control over how `C` is indexed, which may
-    /// allow you to improve the ergonomics or performance of an index over recommended defaults.
-    ///
-    /// See also [`add_index`](App::add_index).
-    ///
-    /// # Examples
-    ///
-    /// ```rust
-    /// # use bevy_app::prelude::*;
-    /// # use bevy_ecs::{prelude::*, index::IndexOptions};
     /// # use bevy_utils::default;
     /// #
     /// # let mut app = App::new();
@@ -1391,7 +1350,7 @@ impl App {
     ///     Blue,
     /// }
     ///
-    /// app.add_index_with_options(IndexOptions::<FavoriteColor> {
+    /// app.add_index(IndexOptions::<FavoriteColor> {
     ///     // FavoriteColor only has 3 states, so can be totally addressed in 2 bits.
     ///     address_space: 2,
     ///     ..default()
@@ -1405,11 +1364,11 @@ impl App {
     /// }
     /// # app.add_systems(Update, find_red_fans);
     /// ```
-    pub fn add_index_with_options<C: Component<Mutability = Immutable>, S: IndexStorage<C>>(
+    pub fn add_index<C: Component<Mutability = Immutable>, S: IndexStorage<C>>(
         &mut self,
         options: IndexOptions<C, S>,
     ) -> &mut Self {
-        self.world_mut().add_index_with_options(options);
+        self.world_mut().add_index(options);
         self
     }
 }
--- a/crates/bevy_app/src/app.rs
+++ b/crates/bevy_app/src/app.rs
@@ -9,16 +9,16 @@ use alloc::{
 };
 pub use bevy_derive::AppLabel;
 use bevy_ecs::{
-    component::RequiredComponentsError,
+    component::{Immutable, RequiredComponentsError},
     event::{event_update_system, EventCursor},
-    index::IndexableComponent,
+    index::{IndexOptions, IndexStorage},
     intern::Interned,
     prelude::*,
     schedule::{ScheduleBuildSettings, ScheduleLabel},
     system::{IntoObserverSystem, SystemId, SystemInput},
 };
 use bevy_platform_support::collections::HashMap;
-use core::{fmt::Debug, num::NonZero, panic::AssertUnwindSafe};
+use core::{fmt::Debug, hash::Hash, num::NonZero, panic::AssertUnwindSafe};
 use log::debug;
 use thiserror::Error;
 
@@ -1331,7 +1331,7 @@ impl App {
     /// This allows querying by component _value_ to be very performant, at the expense of a small
     /// amount of overhead when modifying the indexed component.
     ///
-    /// See [`IndexableComponent`] for further details.
+    /// See also [`add_index_with_options`](App::add_index_with_options).
     ///
     /// # Examples
     ///
@@ -1358,8 +1358,58 @@ impl App {
     /// }
     /// # app.add_systems(Update, find_red_fans);
     /// ```
-    pub fn add_index<C: IndexableComponent>(&mut self) -> &mut Self {
+    pub fn add_index<C: Component<Mutability = Immutable> + Eq + Hash + Clone>(
+        &mut self,
+    ) -> &mut Self {
         self.world_mut().add_index::<C>();
+        self
+    }
+
+    /// Creates and maintains an index for the provided immutable [`Component`] `C` using observers.
+    ///
+    /// This allows querying by component _value_ to be very performant, at the expense of a small
+    /// amount of overhead when modifying the indexed component.
+    ///
+    /// The [options](IndexOptions) provided allow finer control over how `C` is indexed, which may
+    /// allow you to improve the ergonomics or performance of an index over recommended defaults.
+    ///
+    /// See also [`add_index`](App::add_index).
+    ///
+    /// # Examples
+    ///
+    /// ```rust
+    /// # use bevy_app::prelude::*;
+    /// # use bevy_ecs::{prelude::*, index::IndexOptions};
+    /// # use bevy_utils::default;
+    /// #
+    /// # let mut app = App::new();
+    /// #[derive(Component, PartialEq, Eq, Hash, Clone)]
+    /// #[component(immutable)]
+    /// enum FavoriteColor {
+    ///     Red,
+    ///     Green,
+    ///     Blue,
+    /// }
+    ///
+    /// app.add_index_with_options(IndexOptions::<FavoriteColor> {
+    ///     // FavoriteColor only has 3 states, so can be totally addressed in 2 bits.
+    ///     address_space: 2,
+    ///     ..default()
+    /// });
+    ///
+    /// fn find_red_fans(mut query: QueryByIndex<FavoriteColor, Entity>) {
+    ///     let mut lens = query.at(&FavoriteColor::Red);
+    ///     for entity in lens.query().iter() {
+    ///         println!("{entity:?} likes the color Red!");
+    ///     }
+    /// }
+    /// # app.add_systems(Update, find_red_fans);
+    /// ```
+    pub fn add_index_with_options<C: Component<Mutability = Immutable>, S: IndexStorage<C>>(
+        &mut self,
+        options: IndexOptions<C, S>,
+    ) -> &mut Self {
+        self.world_mut().add_index_with_options(options);
         self
     }
 }
--- a/crates/bevy_app/src/app.rs
+++ b/crates/bevy_app/src/app.rs
@@ -1460,7 +1460,7 @@ impl Termination for AppExit {
 
 #[cfg(test)]
 mod tests {
-    use core::{iter, marker::PhantomData, hash::Hash};
+    use core::{hash::Hash, iter, marker::PhantomData};
     use std::sync::Mutex;
 
     use bevy_ecs::{
--- a/crates/bevy_app/src/app.rs
+++ b/crates/bevy_app/src/app.rs
@@ -11,6 +11,7 @@ pub use bevy_derive::AppLabel;
 use bevy_ecs::{
     component::RequiredComponentsError,
     event::{event_update_system, EventCursor},
+    index::IndexComponent,
     intern::Interned,
     prelude::*,
     schedule::{ScheduleBuildSettings, ScheduleLabel},
@@ -1322,6 +1323,42 @@ impl App {
         observer: impl IntoObserverSystem<E, B, M>,
     ) -> &mut Self {
         self.world_mut().add_observer(observer);
+        self
+    }
+
+    /// Creates and maintains an index for the provided immutable [`Component`] `C` using observers.
+    ///
+    /// This allows querying by component _value_ to be very performant, at the expense of a small
+    /// amount of overhead when modifying the indexed component.
+    ///
+    /// See [`IndexComponent`] for further details.
+    ///
+    /// # Examples
+    ///
+    /// ```rust
+    /// # use bevy_app::prelude::*;
+    /// # use bevy_ecs::prelude::*;
+    /// #
+    /// # let mut app = App::new();
+    /// #[derive(Component, PartialEq, Eq, Hash, Clone)]
+    /// #[component(immutable)]
+    /// enum FavoriteColor {
+    ///     Red,
+    ///     Green,
+    ///     Blue,
+    /// }
+    ///
+    /// app.add_index::<FavoriteColor>();
+    ///
+    /// fn find_red_fans(mut query: QueryByIndex<FavoriteColor, Entity>) {
+    ///     for entity in query.at(FavoriteColor::Red).iter() {
+    ///         println!("{entity:?} likes the color Red!");
+    ///     }
+    /// }
+    /// # app.add_systems(Update, find_red_fans);
+    /// ```
+    pub fn add_index<C: IndexComponent>(&mut self) -> &mut Self {
+        self.world_mut().add_index::<C>();
         self
     }
 }
--- a/crates/bevy_app/src/app.rs
+++ b/crates/bevy_app/src/app.rs
@@ -11,7 +11,7 @@ pub use bevy_derive::AppLabel;
 use bevy_ecs::{
     component::RequiredComponentsError,
     event::{event_update_system, EventCursor},
-    index::IndexComponent,
+    index::IndexableComponent,
     intern::Interned,
     prelude::*,
     schedule::{ScheduleBuildSettings, ScheduleLabel},
@@ -1357,7 +1357,7 @@ impl App {
     /// }
     /// # app.add_systems(Update, find_red_fans);
     /// ```
-    pub fn add_index<C: IndexComponent>(&mut self) -> &mut Self {
+    pub fn add_index<C: IndexableComponent>(&mut self) -> &mut Self {
         self.world_mut().add_index::<C>();
         self
     }
--- a/crates/bevy_app/src/app.rs
+++ b/crates/bevy_app/src/app.rs
@@ -1351,7 +1351,8 @@ impl App {
     /// app.add_index::<FavoriteColor>();
     ///
     /// fn find_red_fans(mut query: QueryByIndex<FavoriteColor, Entity>) {
-    ///     for entity in query.at(&FavoriteColor::Red).iter() {
+    ///     let mut lens = query.at(&FavoriteColor::Red);
+    ///     for entity in lens.query().iter() {
     ///         println!("{entity:?} likes the color Red!");
     ///     }
     /// }
--- a/crates/bevy_ecs/src/index/mod.rs
+++ b/crates/bevy_ecs/src/index/mod.rs
@@ -137,7 +137,6 @@ use crate::{
     component::{Component, ComponentDescriptor, ComponentId, Immutable, StorageType},
     entity::Entity,
     prelude::Trigger,
-    query::{QueryBuilder, QueryData, QueryFilter},
     system::{Commands, Query, ResMut},
     world::{FromWorld, OnInsert, OnReplace, World},
 };
@@ -310,26 +309,6 @@ impl<C: IndexableComponent> Index<C> {
             .enumerate()
             .filter_map(|(i, &id)| (index & (1 << i) > 0).then_some(id))
             .collect::<Vec<_>>()
-    }
-
-    fn filter_query_for<D: QueryData, F: QueryFilter>(
-        &self,
-        builder: &mut QueryBuilder<D, F>,
-        value: &C,
-    ) {
-        let Some(&index) = self.mapping.get(value) else {
-            // If there is no marker, create a no-op query by including With<C> and Without<C>
-            builder.without::<C>();
-            return;
-        };
-
-        for (i, &id) in self.markers.iter().enumerate() {
-            if index & (1 << i) > 0 {
-                builder.with_id(id);
-            } else {
-                builder.without_id(id);
-            }
-        }
     }
 }
 
--- a/crates/bevy_ecs/src/index/mod.rs
+++ b/crates/bevy_ecs/src/index/mod.rs
@@ -76,7 +76,7 @@
 //! # #[component(immutable)]
 //! # struct Planet(&'static str);
 //! # let mut world = World::new();
-//! world.add_index::<Planet>();
+//! world.add_index(IndexOptions::<Planet>::default());
 //! ```
 //!
 //! This sets up the necessary mechanisms behind the scenes to track `Planet` components and make
@@ -140,7 +140,6 @@ use alloc::{boxed::Box, format, vec::Vec};
 use bevy_ecs_macros::Resource;
 use bevy_platform_support::{collections::HashMap, hash::FixedHasher, sync::Arc};
 use bevy_ptr::OwningPtr;
-use bevy_utils::default;
 use core::{alloc::Layout, hash::Hash, marker::PhantomData, ptr::NonNull};
 use thiserror::Error;
 
@@ -265,7 +264,7 @@ impl<C: Component<Mutability = Immutable>> Index<C> {
                     match error {
                         TrackEntityError::AddressSpaceExhausted => {
                             log::error!(
-                                "Entity {:?} could not be indexed by component {} as the total addressable space ({} bits) has been exhausted. Consider increasing the address space using `add_index_with_options`.",
+                                "Entity {:?} could not be indexed by component {} as the total addressable space ({} bits) has been exhausted. Consider increasing the address space using `IndexOptions::address_space`.",
                                 entity,
                                 disqualified::ShortName::of::<C>(),
                                 index.markers.len(),
@@ -395,45 +394,26 @@ impl<C: Component<Mutability = Immutable> + Eq + Hash + Clone> Default
     }
 }
 
-/// Extension methods for [`World`] to assist with indexing components.
-pub trait WorldIndexExtension {
-    /// Create and track an index for `C` with the provided options.
-    /// This is required to use the [`QueryByIndex`] system parameter.
-    ///
-    /// See also [`add_index`](WorldIndexExtension::add_index).
-    fn add_index_with_options<C: Component<Mutability = Immutable>, S: IndexStorage<C>>(
-        &mut self,
-        options: IndexOptions<C, S>,
-    ) -> &mut Self;
+impl<C: Component<Mutability = Immutable>, S: IndexStorage<C>> IndexOptions<C, S> {
+    /// Performs initial setup for an index.
+    // Note that this is placed here instead of inlined into `World` to allow most
+    // most of the indexing internals to stay private.
+    #[inline]
+    pub(crate) fn setup_index(self, world: &mut World) {
+        if world.get_resource::<Index<C>>().is_none() {
+            let mut index = Index::<C>::new(world, self);
 
-    /// Create and track an index for `C`.
-    /// This is required to use the [`QueryByIndex`] system parameter.
-    ///
-    /// See also [`add_index_with_options`](WorldIndexExtension::add_index_with_options).
-    fn add_index<C: Component<Mutability = Immutable> + Eq + Hash + Clone>(&mut self) -> &mut Self {
-        self.add_index_with_options::<C, HashMap<C, usize>>(default())
-    }
-}
-
-impl WorldIndexExtension for World {
-    fn add_index_with_options<C: Component<Mutability = Immutable>, S: IndexStorage<C>>(
-        &mut self,
-        options: IndexOptions<C, S>,
-    ) -> &mut Self {
-        if self.get_resource::<Index<C>>().is_none() {
-            let mut index = Index::<C>::new(self, options);
-
-            self.query::<(Entity, &C)>()
-                .iter(self)
+            world.query::<(Entity, &C)>()
+                .iter(world)
                 .map(|(entity, _)| entity)
                 .collect::<Vec<_>>()
                 .into_iter()
                 .for_each(|entity| {
-                    if let Err(error) = index.track_entity(self, entity) {
+                    if let Err(error) = index.track_entity(world, entity) {
                         match error {
                             TrackEntityError::AddressSpaceExhausted => {
                                 log::error!(
-                                    "Entity {:?} could not be indexed by component {} as the total addressable space ({} bits) has been exhausted. Consider increasing the address space using `add_index_with_options`.",
+                                    "Entity {:?} could not be indexed by component {} as the total addressable space ({} bits) has been exhausted. Consider increasing the address space using `IndexOptions::address_space`.",
                                     entity,
                                     disqualified::ShortName::of::<C>(),
                                     index.markers.len(),
@@ -446,11 +426,9 @@ impl WorldIndexExtension for World {
                     }
                 });
 
-            self.insert_resource(index);
-            self.add_observer(Index::<C>::on_insert);
-            self.add_observer(Index::<C>::on_replace);
+            world.insert_resource(index);
+            world.add_observer(Index::<C>::on_insert);
+            world.add_observer(Index::<C>::on_replace);
         }
-
-        self
     }
 }
--- a/crates/bevy_ecs/src/index/mod.rs
+++ b/crates/bevy_ecs/src/index/mod.rs
@@ -92,7 +92,7 @@
 //! fn get_earthlings(mut query: QueryByIndex<Planet, Entity>) {
 //!     let earthlings = query.at(&Planet("Earth"));
 //!
-//!     for earthling in &earthlings {
+//!     for earthling in &earthlings.query() {
 //!         // ...
 //!     }
 //! }
--- a/crates/bevy_ecs/src/index/mod.rs
+++ b/crates/bevy_ecs/src/index/mod.rs
@@ -34,7 +34,9 @@ impl WorldIndexExtension for World {
     }
 }
 
-/// Marker describing the requirements for a [`Component`] to be suitable for indexing.
+/// A marker trait describing the requirements for a [`Component`] to be indexed and accessed via [`QueryByIndex`].
+///
+/// See the module docs for more information.
 pub trait IndexableComponent: Component<Mutability = Immutable> + Eq + Hash + Clone {}
 
 impl<C: Component<Mutability = Immutable> + Eq + Hash + Clone> IndexableComponent for C {}
--- a/crates/bevy_ecs/src/index/mod.rs
+++ b/crates/bevy_ecs/src/index/mod.rs
@@ -131,6 +131,7 @@
 use crate::{
     self as bevy_ecs,
     component::{Component, ComponentDescriptor, ComponentId, Immutable, StorageType, Tick},
+    entity::Entity,
     prelude::Trigger,
     query::{QueryBuilder, QueryData, QueryFilter, QueryState, With},
     system::{Commands, Query, Res, ResMut, SystemMeta, SystemParam},
@@ -152,11 +153,21 @@ pub trait WorldIndexExtension {
 impl WorldIndexExtension for World {
     fn add_index<C: IndexableComponent>(&mut self) -> &mut Self {
         if self.get_resource::<Index<C>>().is_none() {
+            let mut index = Index::<C>::default();
+
+            self.query::<(Entity, &C)>()
+                .iter(&self)
+                .map(|(entity, _)| entity)
+                .collect::<Vec<_>>()
+                .into_iter()
+                .for_each(|entity| {
+                    index.track_entity(self, entity);
+                });
+
+            self.insert_resource(index);
             self.add_observer(Index::<C>::on_insert);
             self.add_observer(Index::<C>::on_replace);
         }
-
-        self.init_resource::<Index<C>>();
 
         self
     }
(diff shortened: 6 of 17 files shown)


crates/bevy_ecs/src/index/mod.rs
```diff
@@ -0,0 +1,177 @@
...
+
+            let query = builder.build();
+
+            query
+        })
+    }
+}
+
+/// Marker describing the requirements for a [`Component`] to be suitable for indexing.
+pub trait IndexComponent: Component<Mutability = Immutable> + Eq + Hash + Clone {}
```
alice-i-cecile, line 63: I think Hash is too strong. BTreeMaps or QuadTrees will work just fine.

bushrat011899: Both of those would require `Ord`, which might be difficult to justify for spatial indexes. Perhaps we can offer a more generic version of this API that would allow users to choose which trait bounds and storage option makes the most sense for them?

bushrat011899:
Marking as ready-for-review as I have the finalised MVP API I would like for this feature. As mentioned, it would be nice to allow `Ord` instead of `Hash` as an alternative, but I'll keep it simple for now just to see if the underlying strategy of fragmenting on an immutable component's value using observers is valid.

crates/bevy_ecs/src/index/mod.rs
alice-i-cecile: Yeah, that's my hope. Ideally we move the bounds off of here, and onto the lookup methods.

crates/bevy_ecs/src/index/mod.rs
```diff
@@ -0,0 +1,286 @@
...
+        }
+
+        self.init_resource::<Index<C>>();
+
+        self
+    }
+}
+
+/// Marker describing the requirements for a [`Component`] to be suitable for indexing.
+pub trait IndexComponent: Component<Mutability = Immutable> + Eq + Hash + Clone {}
```
alice-i-cecile, line 38: This should be called `IndexableComponent`.

crates/bevy_ecs/src/index/mod.rs
```diff
@@ -0,0 +1,286 @@
...
+            self.add_observer(Index::<C>::on_replace);
+        }
+
+        self.init_resource::<Index<C>>();
+
+        self
+    }
+}
+
+/// Marker describing the requirements for a [`Component`] to be suitable for indexing.
```
alice-i-cecile, line 37:
```suggestion
/// A marker trait describing the requirements for a [`Component`] to be indexed and accessed via [`QueryByIndex`].
///
/// See the module docs for more information.
```

crates/bevy_ecs/src/index/mod.rs
```diff
@@ -0,0 +1,286 @@
...
+    ///
+    /// world.add_index::<FavoriteColor>();
+    ///
+    /// fn find_red_fans(mut query: QueryByIndex<FavoriteColor, Entity>) {
+    ///     for entity in query.at(&FavoriteColor::Red).iter() {
+    ///         println!("{entity:?} likes the color Red!");
+    ///     }
+    /// }
+    /// ```
+    pub fn at(&mut self, value: &C) -> Query<'_, '_, D, (F, With<C>)> {
```
alice-i-cecile, line 75: Follow-up: this implies the ability to do `between`, which would be super sweet. Another argument for an `Ord` backend.

crates/bevy_ecs/src/index/mod.rs
```diff
@@ -0,0 +1,286 @@
...
+            state: None,
+            last_run: system_meta.last_run,
+            this_run: change_tick,
+            index,
+        }
+    }
+}
+
+#[derive(Resource)]
+struct Index<C: IndexComponent> {
```
alice-i-cecile, line 157: Doc strings please; it is internal but it really eases maintenance.

crates/bevy_ecs/src/index/mod.rs
```diff
@@ -0,0 +1,286 @@
...
+    mapping: HashMap<C, ComponentId>,
+    spare_markers: Vec<ComponentId>,
+}
+
+impl<C: IndexComponent> Default for Index<C> {
```
alice-i-cecile, line 162:
```suggestion
// Rust's derives assume that C must impl Default for this to hold,
// but that's not true.
impl<C: IndexComponent> Default for Index<C> {
```

crates/bevy_ecs/src/index/mod.rs
```diff
@@ -0,0 +1,286 @@
...
+
+        let value = query
+            .get(entity)
+            .expect("observer should see a component in on_replace");
+
+        let component_id = index
+            .mapping
+            .get(value)
+            .copied()
+            .expect("somehow didn't track this value");
```
alice-i-cecile, line 229: The value should be reported here.

crates/bevy_ecs/src/index/mod.rs
```diff
@@ -0,0 +1,286 @@
...
+        trigger: Trigger<OnReplace, C>,
+        query: Query<&C>,
+        index: Res<Index<C>>,
+        mut commands: Commands,
+    ) {
+        let entity = trigger.target();
+
+        let value = query
+            .get(entity)
+            .expect("observer should see a component in on_replace");
```
alice-i-cecile, line 223: This is worse than just unwrapping.

crates/bevy_ecs/src/index/mod.rs
```diff
@@ -0,0 +1,286 @@
...
+
+                // On removal, we check if this was the last usage of this marker.
+                // If so, we can recycle it for a different value
+                let Self {
+                    ref mut mapping,
+                    ref mut spare_markers,
+                    ..
+                } = index.as_mut();
+
+                // TODO: It may be more performant to make a clone of the old value and lookup directly
```
alice-i-cecile, line 244:
```suggestion
                // PERF: It may be more performant to make a clone of the old value and lookup directly
```

crates/bevy_ecs/src/index/mod.rs
```diff
@@ -0,0 +1,286 @@
...
+            .copied()
+            .expect("somehow didn't track this value");
+
+        commands.queue(move |world: &mut World| {
+            world.resource_scope::<Self, _>(|world, mut index| {
+                // The old marker is no longer applicable since the value has changed/been removed.
+                world.entity_mut(entity).remove_by_id(component_id);
+
+                // On removal, we check if this was the last usage of this marker.
+                // If so, we can recycle it for a different value
```
alice-i-cecile, line 237: Oh that's pretty.

crates/bevy_ecs/src/index/mod.rs
```diff
@@ -0,0 +1,286 @@
...
+        self
+    }
+}
+
+/// Marker describing the requirements for a [`Component`] to be suitable for indexing.
+pub trait IndexComponent: Component<Mutability = Immutable> + Eq + Hash + Clone {}
+
+impl<C: Component<Mutability = Immutable> + Eq + Hash + Clone> IndexComponent for C {}
+
+/// This system parameter allows querying by an indexed component value.
```
alice-i-cecile, line 42: Doc test here please!

crates/bevy_ecs/src/index/mod.rs
```diff
@@ -0,0 +1,286 @@
+//! Provides indexing support for the ECS.
```
alice-i-cecile, line 1: Expand and add an end-to-end doc test demonstrating the feature.

We particularly need to touch on the fact that this is only suitable for components with a relatively small number of values. Using this strategy for e.g. floats would quickly exhaust the component id space.

examples/README.md
```diff
@@ -313,6 +313,7 @@ Example | Description
 [Generic System](../examples/ecs/generic_system.rs) | Shows how to create systems that can be reused with different types
 [Hierarchy](../examples/ecs/hierarchy.rs) | Creates a hierarchy of parents and children entities
 [Immutable Components](../examples/ecs/immutable_components.rs) | Demonstrates the creation and utility of immutable components
+[Indexes](../examples/ecs/index.rs) | Demonstrates the creation and utility of indexes
```
alice-i-cecile, line 316: Describe what an index is instead 😂

examples/ecs/index.rs
```diff
@@ -0,0 +1,56 @@
+//! Demonstrates how to use indexes.
```
alice-i-cecile, line 1: Describe what an index is.

examples/ecs/index.rs
```diff
@@ -0,0 +1,56 @@
...
+
+use bevy::prelude::*;
+
+#[derive(Component, PartialEq, Eq, Hash, Clone, Debug)]
+#[component(immutable)]
+struct TilePosition {
```
alice-i-cecile, line 7: Can we swap these to an i16 or something? As is, the space is as big as the space we're using for ComponentId 😅

alice-i-cecile:
Before we merge this, I want to have benchmarks added in this PR. We need to make sure that this is faster than the naive strategy and stays that way.

crates/bevy_ecs/src/index/mod.rs
bushrat011899: Definitely! Generalised storage backends could allow some _really_ nice methods here for sure.

crates/bevy_ecs/src/index/mod.rs
bushrat011899: There's some pretty _wild_ Rust in this file!

crates/bevy_ecs/src/index/mod.rs
bushrat011899: I've added some doc-tests, background, and drawbacks to the module-level documentation.

examples/ecs/index.rs
bushrat011899: Oh yeah that's a good point lol

crates/bevy_ecs/src/index/mod.rs
```diff
@@ -0,0 +1,454 @@
...
+    ///
+    /// world.add_index::<FavoriteColor>();
+    ///
+    /// fn find_red_fans(mut query: QueryByIndex<FavoriteColor, Entity>) {
+    ///     for entity in query.at(&FavoriteColor::Red).iter() {
+    ///         println!("{entity:?} likes the color Red!");
+    ///     }
+    /// }
+    /// ```
+    pub fn at(&mut self, value: &C) -> Query<'_, '_, D, (F, With<C>)> {
```
tychedelia, line 238: I'd call this `where` personally but maybe `at` has a more ecs-y flavor that I'm not familiar with.

bushrat011899: I like `at` personally because it's like we're getting the query _at_ a particular value. But I'd be happy to change it

bushrat011899:
Preliminary benchmarking shows the index speeds up iteration by a factor of two when 1-in-16 entities has the target value. However, it degrades mutation performance of that index component by a factor of ten in that same scenario. As expected, indexing is beneficial when you're searching for values at a much higher rate than updating them.

bushrat011899:
With a little refactoring I was able to close the performance gap in mutation to a factor of 4 in the benchmark. Still costly, but an improvement nonetheless. One last thing I want to change is `app.add_index` doesn't check for _existing_ spawned entities. It should for correctness.

cBournhonesque:
> I've added some index benchmarks to the ecs bench which test iteration speed (index_iter) and mutation speed (index_update). Running these benchmarks locally, I see a performance improvement by a factor of 2 when iterating using the index (where 1 in 16 are the target value) vs a naive iter().filter(...). However there is a performance degradation by a factor of 4 in that same scenario.

We do want to benchmark again `iter().filter(...)` which just filters over all entities, but I think what we also really want is to benchmark against a more naive index implementation that just maintains a `HashMap<value, Entity>`, right?

Carter0 (approved):
I know my review doesn't mean much on this, but I think the code looks really clean and I think the pitch for this implementation is really good.

crates/bevy_ecs/src/index/mod.rs
```diff
@@ -0,0 +1,468 @@
...
+        // Need a marker component for this entity
+        let component_id = match self.mapping.get_mut(value) {
+            // This particular value already has an assigned marker, reuse it.
+            Some(state) => {
+                state.live_count += 1;
+                state.component_id
+            }
+            None => {
+                // Need to clone the index value for later lookups
+                let value = value.clone();
```
cBournhonesque, line 380: It doesn't have to be in this PR, but I see a world where the value that is used here is instead of the result of some `IndexableComponent::hash_value()` function, so that we can group components into 'buckets' of values.

That way we could avoid the issue of needing too many ZSTs components for components with float values, as we could for example map all the entities with values between 0.0-1.0 in the same bucket (and the query would then have to do an additional check via a `.iter().filter()` in the `QueryIndex` since there's no guarantee that the entities returned by the index have the exact value that we want).

I'm not super knowledgeable about indexes but basically I think we can re-use a bunch of tricks used by traditional DB indexes

crates/bevy_ecs/src/index/mod.rs
```diff
@@ -0,0 +1,468 @@
...
+//! This gives us the same iteration performance as if we had created all those planet
+//! marker components at compile time.
+//!
+//! # Drawbacks
+//!
+//! Indexing by a component value isn't free unfortunately. If it was, it would be enabled by default!
+//!
+//! ## Fragmentation
+//!
+//! To provide the maximum iteration speed, the indexable component is fragmented, meaning each unique
```
cBournhonesque, line 113: That does seem

crates/bevy_ecs/src/index/mod.rs
```diff
@@ -0,0 +1,468 @@
...
+//! value is stored in its own archetype.
+//! This makes iterating through a subset of the total archetype faster, but decreases the performance
+//! of iterating the whole archetype by a small amount.
+//!
+//! ## Component ID Exhaustion
+//!
+//! Fragmenting requires unique [component IDs](ComponentId), and they are finite.
+//! For components with a small number of values in use at any one time, this is acceptable.
+//! But for components with a _large_ number of unique values (e.g., components containing floating point numbers),
+//! the pool of available component IDs will be quickly exhausted.
```
cBournhonesque, line 123: That does seem dangerous, I wonder if we could return a warning after a certain number of ZSTs are used?

Also I guess the work to make the ECS work well with higher component IDs will become more important

crates/bevy_ecs/src/index/mod.rs
```diff
@@ -0,0 +1,468 @@
...
+/// ```rust
+/// # use bevy_ecs::prelude::*;
+/// # let mut world = World::new();
+/// #[derive(Component, PartialEq, Eq, Hash, Clone)]
+/// #[component(immutable)]
+/// struct Player(u8);
+///
+/// // Indexing is opt-in through `World::add_index`
+/// world.add_index::<Player>();
+/// # world.spawn(Player(0));
```
cBournhonesque, line 196: nit: the doc is a bit lengthy, maybe having 0 `Player(0)`, 1 `Player(1)`, 2 `Player(2)`s is enough to demonstrate how this works

crates/bevy_ecs/src/index/mod.rs
cBournhonesque: I prefer `at` to `where` personally. I was otherwise thinking of `for`

crates/bevy_ecs/src/index/mod.rs
```diff
@@ -0,0 +1,468 @@
...
+    pub fn at(&mut self, value: &C) -> Query<'_, '_, D, (F, With<C>)> {
+        self.state = {
+            // SAFETY: Mutable references do not alias and will be dropped after this block
+            let mut builder = unsafe { QueryBuilder::new(self.world.world_mut()) };
+
+            match self.index.mapping.get(value) {
+                // If there is a marker, restrict to it
+                Some(state) => builder.with_id(state.component_id),
+                // Otherwise, create a no-op query by including With<C> and Without<C>
+                None => builder.without::<C>(),
```
cBournhonesque, line 258: clever!

examples/ecs/index.rs
```diff
@@ -0,0 +1,76 @@
...
+    x: i8,
+    y: i8,
+}
+
+fn main() {
+    App::new()
+        // Simply call `add_index` to start indexing a component.
+        // Make sure to call this _before_ any components are spawned, otherwise
+        // the index will miss those entities!
+        .add_index::<TilePosition>()
```
cBournhonesque, line 26: Maybe it would be worthwhile to change `Index<C>` from a `Resource` to a `Component` `Index<C>` that can be added on any entity that we want to be indexable by C.

This lets us:
- ignore some entities that we don't want to get indexed (cheaper indexation). In your version they would still get indexed and the user would have to filter them out in their `QueryByIndex`
- have an easy way to remove an entity from an index (by removing the `Index<C>` component)

For example I might only want to index the `Transform` of only a few entities (player entities), but not of sprites, walls, etc.

Maybe both version can co-exist, because your version has ergonomic benefits (users don't have to add Index<C> on every entity that needs to be indexed)

crates/bevy_ecs/src/index/mod.rs
```diff
@@ -0,0 +1,468 @@
...
+    /// fn find_red_fans(mut query: QueryByIndex<FavoriteColor, Entity>) {
+    ///     for entity in query.at(&FavoriteColor::Red).iter() {
+    ///         println!("{entity:?} likes the color Red!");
+    ///     }
+    /// }
+    /// ```
+    pub fn at(&mut self, value: &C) -> Query<'_, '_, D, (F, With<C>)> {
+        self.state = {
+            // SAFETY: Mutable references do not alias and will be dropped after this block
+            let mut builder = unsafe { QueryBuilder::new(self.world.world_mut()) };
```
chescock, line 252: I think this call to `world_mut()` is unsound.  It will conflict with uses in other params in the same system, or in other systems running in parallel.  It might be impossible to use `QueryBuilder` like this.  

One approach that might work is to use the `QueryState` you're already storing in the parameter state, and try to re-filter the `matched_storage_ids` (and somehow put them back when you're done).  That would also mean you only have to iterate archetypes/tables matching the overall filter, instead of having to check *all* of them as `QueryBuilder` does.

bushrat011899: I was concerned by this. I'm currently refactoring how this works to be more conservative with component IDs but once that is done I'd like to revisit how this works. Your suggestion might be the right way forward (for performance too probably!)

chescock: Hmm, your plan is to preallocate 32 component ids to use as bits, and use those to encode a 32-bit index in the archetype, right?  I think you could build something that stores a `Vec<QueryState>` and does all of the archetype filtering in `new_archetype`.  That would give you no overhead at query execution time, and would let you get queries with `&self` instead of `&mut self`!  

The idea would be to implement `SystemParam::new_archetype()` manually by first doing the same filtering as any `Query<D, (F, With<C>)>`, but then by calling `archetype.contains` 32 times to get the 32-bit index encoded in that archetype.  Then you'd look up the `QueryState` at that index in the `Vec`, and tell only that one about the archetype.  So each `QueryState` in the `Vec` would always have the full list of archetypes with the given index, and none with other indices!  

You'd need a way to make new `QueryState`s when you first see higher indices, but I think that's straightforward.  Everything inside it is already `Clone` except for `D::State` and `F::State`, and I think all actual `WorldQuery` impls have `State: Clone`, so you could just add a bound there.

bushrat011899: I'm going to try and do this now that I've got the combinatorial component ID indexing working. Seriously, thank you for explaining this!

bushrat011899: Ok so after much heartache, pain, and fear, I believe I have a sound implementation for this system parameter. Firstly, I pre-allocate all the component IDs as discussed, which then allows the `QueryByIndex` system parameter to know all relevant markers ahead of time.

Then I use `QueryBuilder` inside the parameter's `init_state` function to create 2 `QueryState`s per bit required for the index; one for `with` and one for `without`. I need both because exclusion is how the binary addressing works. This allows me to maintain all those query states within the normal lifecycle of a system parameter, treating them as the `QueryState` for a hypothetical `Query<(), With<C>>` (the `With<C>` is to ensure the `without` filters don't get overly inclusive).

Finally, I can use `QueryState::join_filtered` to combine the appropriate combination of `with` and `without` `QueryState`s to isolate the required subset of archetypes.

Thankfully, this also doesn't require implementing `Clone` for `QueryState`, since `join_filtered` _kinda_ already clones in that it takes two references and creates an owned third.

And thankfully this is still faster than just using `filter_map`; looking for a 1-in-1,000 entity amongst 1,000,000 takes 500us on my machine using `filter_map` and only 20us using the index.

crates/bevy_ecs/src/index/mod.rs
bushrat011899: I've combined the test a little but still left it extensive. It actually helped quite a bit during a major refactor having the more detailed set of results to compare against.

crates/bevy_ecs/src/index/mod.rs
bushrat011899: I have now resolved this issue by instead using a unique combination of marking ZSTs per value. This puts a ceiling of 32 components required to index every single entity individually.

crates/bevy_ecs/src/index/mod.rs
```diff
@@ -0,0 +1,343 @@
...
+    /// Creates a new marker component for this index.
+    /// It represents a ZST and is not tied to a particular value.
+    /// This allows moving entities into new archetypes based on the indexed value.
+    fn alloc_new_marker(world: &mut World) -> ComponentId {
+        // SAFETY:
+        // - ZST is Send + Sync
+        // - No drop function provided or required
+        let descriptor = unsafe {
+            ComponentDescriptor::new_with_layout(
+                format!("Index Marker ({})", core::any::type_name::<Self>()),
```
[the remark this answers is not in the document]
bushrat011899: Yeah good idea!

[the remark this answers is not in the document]
bushrat011899: Oh yeah much nicer

crates/bevy_ecs/src/index/mod.rs
```diff
@@ -0,0 +1,343 @@
...
+                    .flatten();
+
+                match spare_slot {
+                    Some((index, slot)) => {
+                        slot.active += 1;
+                        index
+                    }
+                    None => {
+                        let index = self.slots.len();
+                        self.mapping.insert(value.clone(), index);
```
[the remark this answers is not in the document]
bushrat011899: Yeah you're right! I guess I hadn't hit a code-path yet where this fails. Good catch!

crates/bevy_ecs/src/index/mod.rs
```diff
@@ -0,0 +1,343 @@
...
+                self.slots[index].active += 1;
+                index
+            }
+            None => {
+                let spare_slot = (self.slots.len() > self.mapping.len())
+                    .then(|| {
+                        self.slots
+                            .iter_mut()
+                            .enumerate()
+                            .find(|(_, slot)| slot.active == 0)
```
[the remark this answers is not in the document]
bushrat011899: I use `enumerate().find()` because I want the index _and_ the stored state. I've switched to a stack of spare slots to remove the search regardless. Originally I was concerned about memory usage, but it "only" doubles the memory usage compared to the `slots` list on its own, so might as well.

crates/bevy_ecs/src/index/query_by_index.rs
```diff
@@ -0,0 +1,287 @@
...
+
+impl<C: IndexableComponent, D: QueryData + 'static, F: QueryFilter + 'static>
+    QueryByIndexState<C, D, F>
+{
+    fn init_state(world: &mut World, system_meta: &mut SystemMeta) -> Self {
+        let primary_query_state =
+            <Query<D, (F, With<C>)> as SystemParam>::init_state(world, system_meta);
+        let index_state = <Res<Index<C>> as SystemParam>::init_state(world, system_meta);
+
+        let ids = world.resource::<Index<C>>().markers.clone();
```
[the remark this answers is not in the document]
bushrat011899: I've updated the method to warn the user to setup the index themselves, but `QueryByIndex` will just call `add_index` after the warning so it can "work". Personally, I don't want users accidentally relying on a feature which they should _really_ think about before using. Indexing changes how a component performs quite substantially so it should be a deliberate choice to enable it IMO.

crates/bevy_ecs/src/system/query.rs
```diff
@@ -1881,6 +1881,33 @@ impl<'w, Q: QueryData, F: QueryFilter> QueryLens<'w, Q, F> {
...
+    ) -> Self {
+        state.validate_world(world.id());
+
+        Self {
+            world,
+            state,
+            last_run,
+            this_run,
+        }
+    }
```
bushrat011899, line 1910: I switched to returning `QueryLens` instead of `Query` in `QueryByIndex::at(_mut)`, which requires making the constructor for a lens accessible to other parts of the crate. I believe this is a sound API to expose, but this may need scrutiny.

examples/ecs/index.rs
```diff
@@ -0,0 +1,181 @@
...
+#[component(immutable, storage = "SparseSet")]
+struct Alive;
+
+#[derive(Component, PartialEq, Eq, Hash, Clone, Copy, Debug)]
+#[component(immutable)]
+struct Position(i8, i8);
+
+#[derive(Component, PartialEq, Eq, Hash, Clone, Copy, Debug)]
+#[component(immutable)]
+struct Chunk(i8, i8);
```
bushrat011899, line 46: Switched to a Conway's Game of Life demonstration for the index feature. Instead of indexing against `Position`, I index against a `Chunk` marker, which groups a 4x4 sub-grid of cells. It performs better and avoids having a _huge_ amount of archetypes. But this is further motivation for letting the user provide a function which can map a component/bundle to a different type for indexing, since it would let the user opt-in to this chunking without a second component.

examples/ecs/index.rs
```diff
@@ -0,0 +1,181 @@
...
+
+                        if count.0 > 8 {
+                            panic!("Wait how did that happen???");
+                        }
+                    }
+                }
+            }
+        }
+    }
+}
```
bushrat011899, line 153: Pretty sure this can be improved. Will go back over it later.

chescock (approved):
This looks awesome!

crates/bevy_ecs/src/index/query_by_index.rs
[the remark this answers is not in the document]
chescock: Oh, wait `Index` isn't `pub`, so that's not actually a danger.  Never mind!

notnotmelon (approved):
I just want to give my 2 cents and say this PR is amazing.

My workflow with bevy is changed from:
   - Should I use the ECS -> Do I need to hash the position?
   
 To now:
   - Should I use the ECS -> **Yes**
   
 I can't wait to remove all the spaghetti code in my project. Thank you!

crates/bevy_ecs/src/index/query_by_index.rs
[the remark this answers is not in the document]
bushrat011899: Yeah I want to keep _how_ the index works as private as possible since this is a _very_ experimental API (who knows, maybe `QueryByIndex` could be made redundant by changing `Query` itself).

examples/ecs/index.rs
```diff
@@ -0,0 +1,179 @@
...
+        for (entity, mut material) in query.iter_mut() {
+            if rng.0.gen::<f32>() < 0.25 {
+                commands.entity(entity).insert(Alive);
+                material.0 = handles.alive_material.clone();
+            }
+        }
+    }
+}
+
+fn spread_livlihood(
```
Jondolf, line 116:
```suggestion
fn spread_livelihood(
```

examples/ecs/index.rs
```diff
@@ -0,0 +1,179 @@
...
+                }),
+                ..default()
+            }),
+            FpsOverlayPlugin::default(),
+        ))
+        .add_index::<Chunk>()
+        .insert_resource(Time::<Fixed>::from_seconds(0.1))
+        .add_systems(Startup, setup)
+        .add_systems(Update, randomly_revive)
+        .add_systems(FixedUpdate, (spread_livlihood, update_state).chain())
```
Jondolf, line 32:
```suggestion
        .add_systems(FixedUpdate, (spread_livelihood, update_state).chain())
```

Cargo.toml
```diff
@@ -2040,6 +2040,18 @@ description = "Demonstrates the creation and utility of immutable components"
...
+name = "index"
+path = "examples/ecs/index.rs"
+doc-scrape-examples = true
+required-features = ["bevy_dev_tools"]
+
+[package.metadata.example.index]
+name = "Indexes"
+description = "Demonstrates querying by component value using indexing"
+category = "ECS (Entity Component System)"
+wasm = false
```
mockersf, line 2053: ny reason this isn't wasm compatible?

bushrat011899: No reason at all! I suspect that was a leftover from the copy/paste of the example above it.

crates/bevy_ecs/src/index/mod.rs
bushrat011899: I've updated this PR to include `add_index_with_options`, which allows configuring the storage backend for an index to an arbitrary type. This also moves the bound out of an `IndexableComponent` trait (now deleted) and up to the call-sites directly. I have included `BTreeMap` as an alternative to `HashMap`, but left `HashMap` as the default.

bushrat011899:
I've updated this PR to allow users to define custom storage back-ends for indexing, and also control other parameters such as the `StorageType` or marker components and the number of bits used to address the component. `StorageType` _may_ want to be changed by a user if they find that `SparseSet`'s better add/remove performance is more important than `Table`'s iteration speed. The number of bits used in addressing would likely want to be changed when working with enums, since the smallest enum will have a `size_of` 8 bits, even if it only has 2 possible states and thus _could_ be uniquely addressed with only 1 bit.

Finally, the storage back-end can replaced with any type implementing the new `IndexStorage` trait. This requires only basic `get/insert/remove` operations and so should be simple for users to implement for their own storage systems as required. Internally, the storage back-end is erased so regardless of the storage used, only one `Index<C>` resource will ever exist for a given component `C`, avoiding ambiguity when using `QueryByIndex`.

I haven't documented it as I don't know if we want to encourage it, but this _can_ allow users to change _how_ a component is indexed. For example, a `Position(f32, f32)` component could be internally truncated to a `Position(i8, i8)` to chunk positions. I haven't encouraged this since it would lead to the confusing (but safe) situation where a user queries for entities with a value `Position(1.1, 1.2)` and they get entities at `Position(1, 1)` as well.

I did try to offer indexing against multiple components (a `Bundle`), but due to the slight differences in `QueryData` and `Bundle`, and the lack of `ImmutableBundle`, I don't think it would be particularly simple to implement. For example, I cannot use `With<(A, B)>` to work with `Bundle` in queries, and I cannot use `Trigger<OnInsert, (&A, &B)>` to work with `QueryData` in observers.

tychedelia:
> I've updated this PR to allow users to define custom storage back-ends for indexing

Does boxing the storage add any additional overhead via indirection in your testing?

bushrat011899:
> Does boxing the storage add any additional overhead via indirection in your testing?

I personally didn't notice any performance loss in the benchmarks, probably since `HashMap` already had 1 level of indirection anyway. Since it's just a `Box` without any locks or sync primitives that's not too surprising IMO. But it's _definitely_ preferable to the alternative of needing to add an extra type parameter to `Index`, making it possible to have multiple indexes for the same component at once.

chescock (approved):

chescock:
> I did try to offer indexing against multiple components

I vote for leaving that out of indexing and making it a separate feature.  

If we're going to map multiple components to one value for indexing, ... then we might as well store that value as a component.  And then we don't even need to tie that feature into indexing!  We build a feature that lets you automatically populate a component derived from a `Query` on immutable components.  If you want to index on a query, you use that feature to populate a derived component, and then index the derived component.

Diddykonga (approved):
Conceptually LGTM
Even has shiny nobs for tuning 😄

crates/bevy_app/src/app.rs
```diff
@@ -1324,6 +1325,93 @@ impl App {
...
+    ///
+    /// fn find_red_fans(mut query: QueryByIndex<FavoriteColor, Entity>) {
+    ///     let mut lens = query.at(&FavoriteColor::Red);
+    ///     for entity in lens.query().iter() {
+    ///         println!("{entity:?} likes the color Red!");
+    ///     }
+    /// }
+    /// # app.add_systems(Update, find_red_fans);
+    /// ```
+    pub fn add_index<C: Component<Mutability = Immutable> + Eq + Hash + Clone>(
```
alice-i-cecile, line 1361: IMO we should just condense this to a single method / API that takes index-setting options (with a nice default). Adding indexes is both rare and inherently a power-user tool that should be tweaked for each usage.

I don't think that the complexity / API surface cost of the two method variants is better than just `IndexOptions::default`.

crates/bevy_ecs/src/index/mod.rs
```diff
@@ -0,0 +1,456 @@
...
+
+/// Options when configuring an index for a given indexable component `C`.
+pub struct IndexOptions<
+    C: Component<Mutability = Immutable>,
+    S: IndexStorage<C> = HashMap<C, usize>,
+> {
+    /// Marker components will be added to indexed entities to allow for efficient lookups.
+    /// This controls the [`StorageType`] that will be used with these markers.
+    /// It is recommended to use [`Table`](StorageType::Table) for more efficient querying, but a
+    /// [`SpareSet`](StorageType::SparseSet) may be more appropriate in certain circumstances.
```
alice-i-cecile, line 357: "in certain circumstances" is a bit vague 😅

crates/bevy_ecs/src/index/mod.rs
```diff
@@ -0,0 +1,456 @@
...
+    /// This controls how many markers will be used to create that unique address.
+    /// Note that a value greater than 32 will be reduced down to 32.
+    /// Bevy's [`World`] only supports 2^32 entities alive at any one moment in time, so an address space
+    /// of 32 is sufficient to uniquely refer to every individual [`Entity`] in the entire [`World`].
+    ///
+    /// Selecting a non-default value may lead to a panic at runtime or entities missing from the
+    /// index if the address space is exhausted.
+    ///
+    /// This defaults to [`size_of<C>`]
+    pub address_space: usize,
```
alice-i-cecile, line 372: Using a usize for this when the maximum allowed value is 32 feels pretty weird 😅

crates/bevy_ecs/src/index/mod.rs
```diff
@@ -0,0 +1,456 @@
...
+    /// This defaults to [`Table`](StorageType::Table).
+    pub marker_storage: StorageType,
+    /// Marker components are combined into a unique address for each distinct value of the indexed
+    /// component.
+    /// This controls how many markers will be used to create that unique address.
+    /// Note that a value greater than 32 will be reduced down to 32.
+    /// Bevy's [`World`] only supports 2^32 entities alive at any one moment in time, so an address space
+    /// of 32 is sufficient to uniquely refer to every individual [`Entity`] in the entire [`World`].
+    ///
+    /// Selecting a non-default value may lead to a panic at runtime or entities missing from the
```
alice-i-cecile, line 368:
```suggestion
    /// Selecting a value lower than the default value may lead to a panic at runtime or entities missing from the
```

crates/bevy_ecs/src/index/mod.rs
```diff
@@ -0,0 +1,456 @@
...
+        Self {
+            marker_storage: StorageType::Table,
+            address_space: size_of::<C>(),
+            index_storage: HashMap::with_hasher(FixedHasher),
+            _phantom: PhantomData,
+        }
+    }
+}
+
+/// Extension methods for [`World`] to assist with indexing components.
```
alice-i-cecile, line 398: We're living in bevy_ecs already; just make these methods on World. See above about wanting to only have a single method though.

crates/bevy_ecs/src/index/query_by_index.rs
```diff
@@ -0,0 +1,304 @@
...
+/// #       by_player.at(&Player(5)).query().iter().count(),
+/// #    ), (1, 2, 3, 4, 5, 6));
+/// }
+/// # world.run_system_cached(find_all_player_one_entities);
+/// ```
+pub struct QueryByIndex<
+    'world,
+    'state,
+    C: Component<Mutability = Immutable>,
+    D: QueryData + 'static,
```
alice-i-cecile, line 56: That's a weird bound. Can you leave a comment here on why 'static is required here?

crates/bevy_ecs/src/index/query_by_index.rs
```diff
@@ -0,0 +1,304 @@
...
+                state = state.join_filtered(self.world, filter);
+            }
+        }
+
+        state
+    }
+}
+
+#[doc(hidden)]
+pub struct QueryByIndexState<
```
alice-i-cecile, line 169:
```suggestion
/// The [`SystemParam::State`] type for [`QuerybyIndex`], which caches information needed for queries.
pub struct QueryByIndexState<
```

crates/bevy_ecs/src/index/query_by_index.rs
```diff
@@ -0,0 +1,304 @@
...
+    C: Component<Mutability = Immutable>,
+    D: QueryData + 'static,
+    F: QueryFilter + 'static,
+> {
+    primary_query_state: QueryState<D, (F, With<C>)>,
+    index_state: ComponentId,
+
+    // TODO: THERE MUST BE A BETTER WAY
+    without_states: Vec<QueryState<(), With<C>>>, // No, With<C> is not a typo
```
alice-i-cecile, line 178: Comment needs to be clearer here. Why on earth does this work?

crates/bevy_ecs/src/index/storage.rs
```diff
@@ -0,0 +1,59 @@
...
+use crate::{component::Immutable, prelude::Component};
+
+/// A storage backend for indexing the [`Component`] `C`.
+///
+/// Typically, you would use a [`HashMap`] for this purpose, but you may be able to further optimize
+/// the indexing performance of a particular component with a custom implementation of this backend.
+///
+/// For example, the component `C` may be uniquely identifiable with a subset of its data, or implement
+/// traits like [`Hash`] or [`Ord`] which would allow for specialized storage options.
+pub trait IndexStorage<C: Component<Mutability = Immutable>>: 'static + Send + Sync {
```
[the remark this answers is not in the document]
alice-i-cecile: Agreed, this is really appealing for follow-up.

crates/bevy_ecs/src/system/query.rs
```diff
@@ -1881,6 +1881,33 @@ impl<'w, Q: QueryData, F: QueryFilter> QueryLens<'w, Q, F> {
...
+
+    /// Creates a new [`QueryLens`].
+    ///
+    /// # Panics
+    ///
+    /// This will panic if the world used to create `state` is not `world`.
+    ///
+    /// # Safety
+    ///
+    /// This will create a query that could violate memory safety rules. Make sure that this is only
```
alice-i-cecile, line 1893: This safety comment is a bit handwavey for my liking.

examples/ecs/index.rs
```diff
@@ -0,0 +1,199 @@
+//! Demonstrates how to query for a component _value_ using indexes.
```
alice-i-cecile, line 1:
```suggestion
//! Demonstrates how to efficiently query for a component _value_ using automatically updated indexes.
//!
//! This pattern can be substantially faster than iterating over and filtering for matching components,
//! at the expense of slower mutations and component insertion.
//!
//! See the [`bevy::ecs::index`] module docs for in-depth information.
```

alice-i-cecile (changes requested):
A few cleanup items, the most serious of which is the `add_index_with_settings` complaint :)

Overall I'm very happy with this: quality is generally great, and this is a really elegant, reliable and powerful design for an essential feature.

crates/bevy_app/src/app.rs
bushrat011899: Yeah that's fair honestly. I've kept the internals of the method inside a `pub(crate) fn` inside the `index` module just so I can keep almost all of the internals private (I think it's easier to iterate knowing the rest of `bevy_ecs` won't break randomly IMO)

crates/bevy_ecs/src/index/mod.rs
bushrat011899: Yeah good point! I've reduced this to a `u8`, _very_ roomy still haha

crates/bevy_ecs/src/index/query_by_index.rs
bushrat011899: Oh good catch! I think this was a leftover from an earlier draft that just didn't cause any issues. Removed!

crates/bevy_ecs/src/system/query.rs
bushrat011899: Yeah it's a straight copy from `Query` since it has the same invariants. I've fiddled with the way `QueryLens` is setup to make it clear what safety invariants need to be upheld and why.

crates/bevy_ecs/src/index/query_by_index.rs
chescock: For what it's worth, I have such a follow up PR mostly ready to go, with the design I mentioned on Discord.  So I wouldn't worry too much how well documented this version is.  And it would be nice not to have too many rebase conflicts :).

crates/bevy_ecs/src/index/mod.rs
```diff
@@ -0,0 +1,436 @@
...
+//! # Drawbacks
+//!
+//! Indexing by a component value isn't free unfortunately. If it was, it would be enabled by default!
+//!
+//! ## Fragmentation
+//!
+//! To provide the maximum iteration speed, the indexable component is fragmented, meaning each unique
+//! value is stored in its own archetype.
+//! This makes iterating through a subset of the total archetype faster, but decreases the performance
+//! of iterating the whole archetype by a small amount.
```
alice-i-cecile, line 116:
```suggestion
//! This makes iterating through a subset of the total archetypes faster, but decreases the performance
//! of iterating all archetypes by a small amount.
```

crates/bevy_ecs/src/index/mod.rs
```diff
@@ -0,0 +1,436 @@
...
+//! Instead of loading archetypes, then the entities, and comparing to our value, we first check an
+//! index for our value and only load the archetypes with that value.
+//! This gives us the same iteration performance as if we had created all those planet
+//! marker components at compile time.
+//!
+//! # Drawbacks
+//!
+//! Indexing by a component value isn't free unfortunately. If it was, it would be enabled by default!
+//!
+//! ## Fragmentation
```
alice-i-cecile, line 111: This needs a paragraph on the perceived "memory leaks" raised by Sander [here](https://discord.com/channels/691052431525675048/749335865876021248/1336823666469769257), which are caused by the lack of archetype cleanup.

crates/bevy_ecs/src/index/mod.rs
```diff
@@ -0,0 +1,436 @@
...
+//!
+//! To provide the maximum iteration speed, the indexable component is fragmented, meaning each unique
+//! value is stored in its own archetype.
```
alice-i-cecile, line 114:
```suggestion
//! value is stored in its own archetype.
//! Archetypes are reused when values are no longer in use;
//! and so the cost paid scales with the maximum number of unique values alive at once.
```

alice-i-cecile (commented):
Nice cleanup of the issues I raised; thanks!

I've left a few more suggestions on documenting performance drawbacks, capturing the discussion on Discord from earlier today. Once those are in, I'm willing to give this my blessing.

For posterity, this *will* cause issues with archetype fragmentation that affect various parts of Bevy, including the parallel scheduler (see #16784). These are broadly tracked by #17564.

That said, I think that this is a valuable feature (for some applications) today, and an excellent base to build off of. Indexes are a really critical tool for optimizing the performance characteristics of table-based data stores, and this is a robust design with good places for us to add options for users to test in their specific applications. If this design ends up being slower / worse than relatively simple "hashmap in a resource" designs, or we want to move to using a distinct marker component for each value, we can *do* that, without breaking end user code.

Finally the fact that this sort of thing is possible (and the fact that archetypal relations, value queries or other advanced features might be needed or desired in the future) means that we probably can't just dodge the perf problems that come with huge numbers of archetypes forever. This is a relatively advanced, performance-focused feature, which makes it more likely (but not guaranteed) that users playing with this are a) actually benchmarking the impact of using this feature/ changes we make to it b) unlikely to be introduced willy-nilly by crates that people are relying on. As a result, I think that this offers us a real world testbed for the optimizations proposed in #17564, which were previously only getting stress-tested on very artificial benchmarks at best.

crates/bevy_ecs/src/index/mod.rs
bushrat011899: I've added a paragraph and included a call-to-action to file a bug report if you come across the issue so that we have a real-world example to optimise against.

alice-i-cecile (approved):

crates/bevy_ecs/src/index/mod.rs
```diff
@@ -0,0 +1,444 @@
...
+use bevy_ecs_macros::Resource;
+use bevy_platform_support::{collections::HashMap, hash::FixedHasher, sync::Arc};
+use bevy_ptr::OwningPtr;
+use core::{alloc::Layout, hash::Hash, marker::PhantomData, ptr::NonNull};
+use thiserror::Error;
+
+/// This [`Resource`] is responsible for managing a value-to-[`ComponentId`] mapping, allowing
+/// [`QueryByIndex`] to simply filter by [`ComponentId`] on a standard [`Query`].
+#[derive(Resource)]
+struct Index<C: Component<Mutability = Immutable>> {
```
cart, line 157: First, I do generally like the shape of this PR (relatively standalone observer-driven index resources with query helpers). 

My biggest concern with this implementation is that it is _very_ memory hungry.

3d_scene on its own idles at 360 MB in debug mode.

If I spawn an index on a 500x500 grid (250,000 entities) in 3d_scene, it uses 2.8 GB. Additionally, the spawn takes 17 seconds. This behaves the same if I split it into two component types each having 125,000 indices.

(note that in release mode the memory usage stays the same, but the 17 seconds drops to 3)

For comparison, if I just manually create a hashmap of the spawned entities, it uses 400 MB and starts instantly. I suspect the observer execution does play a role in that expense (it would be interesting to see how much).

But from this perspective, the stated concern about the 2^64 = 18,446,744,070,000,000,000 value limit imposed by using a single ComponentId feels a bit silly. We could not possibly store that many values in memory using the system in place currently (we would need a computer with 2 x 10^14 GB of memory).

Probably related to this: each value constituting a new Archetype, and each value contributing a very large number (ex: 64) components to that archetype means each value creates 64 new ArchetypeComponentIds. In a world where you have 250,000 values, thats 16,000,000 new ArchetypeComponentIds. Note that this ArchetypeComponentId explosion also counteracts the "ComponentId identiy-space saving" we accomplish with the components-as-bits approach.

Under the current constraints, its feels very hard to endorse this approach. It seems like a simple `value->ComponentId` map would perform better and use less memory, which actually increases the number of unique indices we can use (even if theoretically we're limiting ourselves more).

bushrat011899: Thanks for taking the time to read through the PR!

That level of memory usage is definitely unacceptable. Before going down this path of using `ComponentId`s to form an address I was unaware of the `ArchetypeComponentId` system and how poorly it interacts with the kind of abuse the addressing scheme imparts.

I will attempt to refactor this PR to allow choosing between this addressing scheme and a more simplified 1:1 mapping of `value` to `ComponentId`. Part of my concern with `value` to `ComponentId` 1:1 mapping is that complicates the `QueryByIndex` implementation. Right now it can know ahead of time exactly which marker components it needs to track, regardless of how many unique values are spawned.

Regarding spawn times, I definitely think that has to do with the observers and would be largely unavoidable. In the addressing scheme I insert all the required marker components in a single bundle, which should have the same cost as inserting a single marker component (creating the new archetype if required, moving the entity, etc.).

I am glad that you somewhat approve of the overall user-facing elements of the PR, as that should help this PR (or a followup if this goes south) by giving a relatively clear API to strive for.

bushrat011899: Straight away, changing the storage type for the marker components to `SparseSet` shaves the memory usage down to 1.5GB instead of 2.8GB in my local testing. Since I have exposed this as a user configurable option I'll change the default to the more memory conservative option and let users who need iteration performance change it back.

bushrat011899:
Initial testing appears to indicate that component-per-value would _not_ solve the memory usage issues identified by Cart. Comparing the addressing technique used in this PR to just spawning a unique dynamic component with an entity:

Entities | Indexing with Addresses | Component per Value
-- | -- | --
10 | 1.6MB | 1.7MB
100 | 1.8MB | 2.1MB
1000 | 3.8MB| 11.9MB
10000 | 23.2MB | 789MB
50000 | 106MB | 19.2GB
100000 | 213MB | _OOM_
1000000 | 2.57GB | _OOM_

From this, I'm confident in saying that without substantial changes to Bevy to allow component-per-value to not use so much memory, that the only way to do _fragmenting_ indexing is via the addressing technique proposed in this PR. Non-fragmenting indexing, such as with a `HashMap<C, EntityHashSet>`, may still be preferable to this PR, I need to do more testing.

I will also note here that indexing memory usage drops substantially _without_ `bevy_render` enabled (1.5GB down to 600MB), indicating there is an additional interaction here that needs investigation.

bushrat011899:
Ok I have created an as-close-as-possible replication of this PR except using a `HashMap<T, EntityHashSet>` as the way of mapping component values to entities. You can find a single-file Gist [here](https://gist.github.com/bushrat011899/d6a687ab98f58cbd6fe69c5203e8220c), which also includes a benchmark comparing it against this PR. To run the benchmark, you must be on this branch (e.g., paste this benchmark into `examples/empty.rs`).

In this benchmark, I create 10,000,000 entities and place them on one of 10 planets randomly (seeded RNG for consistency). I also randomly add a pair of marker components to simulate having different archetypes within the indexed set (e.g., some entities are `Players`, some are `Enemies`, etc.). I then iterate all entities _by planet_. In the benchmark I just count them, but this is a stand-in for real work you may do on a per-entity basis. For example, in Avian, we may iterate over all `Transform`s by a particular physics world and compute collisions, etc.

With this PR, it takes 0.00003 seconds (30us) to complete this lookup. With the non-fragmenting alternative in the Gist it takes 0.17 seconds, (about 5,000x slower). This is to be expected, as the non-fragmenting version iterates out of order, potentially going back and forth across archetypes even. While the fragmenting version is able to densely iterate as if the user had inserted an ahead-of-time ZST marker to isolate their particular entities.

Comparing memory usage, non-fragmenting uses 770MB of RAM, while fragmenting uses 710MB. I am unsure _why_ the fragmenting version is more memory efficient. It could be that since there are only 10 unique values, it is more memory efficient to only have 10 entries in a hashmap from `C` value to a set of `ComponentId`s as a marker compared to a mapping to `EntityHashSet`, which must store every entity.

Regardless, I believe that this technique is an acceptable compromise between the two next-best alternatives:

- `ComponentId`-per-value has catastrophic memory consumption properties at moderate levels of indexing, making it trivial for users to OOM. In addition, registering and tracking these `ComponentId`s in a system parameter that can cleanly interface with `Query` was challenging for me in earlier iterations of this PR. Not impossible, but may require further changes to the rest of `bevy_ecs`.
- `HashMap<C, EntityHashSet>` and other value-to-entity techniques sacrifice iteration performance by orders of magnitude. This may be suitable for unique or low-count lookups however, and so there is value in trying to offer this as either an option within this PR or a follow-up.

In my opinion, this PR strikes a careful middle-ground between potential footguns that makes it suitable as a first-party offering. The memory consumption observed _is_ high and _should_ be improved (startup times as well!), but the scaling this technique has makes it far better suited as a general purpose option that a user can _reasonably_ reach for.

maniwani:
Do we need to use component IDs at all? 

If I've understood things so far, you've tried
- filtering entities in the inner loop on `HashMap<T, EntityHashSet>`
- filtering archetypes in the outer loop (building the query cache) *and* entities in the inner loop by fragmenting archetypes based on component value
	- by using 64 `ComponentId` for each type to encode each unique value OR
	- by using 1 `ComponentId` for each unique value

You've noted a huge performance gap between fragmenting the archetypes and the just filtering entities using a `HashMap<T, EntityHashSet>`, but I wonder if we could mostly attribute that to the difference in archetype filtering.

For a fairer comparison, I think you could still filter archetypes in the non-fragmenting case.

IIRC a query builds (or could build) its cache by iterating the list of archetypes matching its first term and filtering them using the other terms. If an archetype passes all terms, it's added to the cache and its entities will be iterated.

If you're using observers to maintain a `HashMap<T, HashSet<Entity>>`, you can also maintain a `HashMap<T, HashMap<ArchetypeId, usize>>`, which can be used just like a component index to filter archetypes going into the query cache.

If a query uses the value index, you'd now have to rebuild its cache whenever an archetype is added to or removed from it.

bushrat011899:
Ok based on discussions on discord and feedback here (both of which I'm incredibly appreciative of!) I have decided to change my focus here. Clearly the question of _how_ to index entities is far from settled. _But_ there is consensus that it _should_ cleanly interact with `Query`. Instead of providing a first party indexing solution that is a jack of all trades and master of none, I am going to change this PR to instead provide a consistent API that users can supply their own indexing solution into, and then provide a good-enough implementation as an example.

After sleeping on it, I believe `QueryFilter` and the associated query iteration types can be _slightly_ extended to allow supplying the `QueryFilter` with information from within the body of a system. My goal API is:

```rust
let query: Query<Foo, ByIndex<Planet, (With<Bar>, ...)>>;

for foo in query.iter().provide_filter(&Planet("Earth")) {
    // ...
}
```

Since this would be used to _reduce_ the total results from a query (either by archetype or by row as determined by the filter itself) I don't foresee any issues around access or safety. 

Once (and if) I have a working prototype I will close this PR in favor of the new proposal.

cart:
> Initial testing appears to indicate that component-per-value would not solve the memory usage issues identified by Cart. Comparing the addressing technique used in this PR to just spawning a unique dynamic component with an entity

Yup I whipped up a quick impl myself and also got these results. A big contributor to the memory usage is `SparseSet<ComponentId, X>`. By swapping that out for a HashSet in Table and Archetype I get 1.1 GB with the "componentid -> value map" impl. #14928 attempted to prepare us for this (although it uses a slightly different approach). I think embarking on the quest to make higher componentids less expensive is the right path to be on. Given that folks have already been working on this (@james-j-obrien, @cBournhonesque , @chescock, @sandermertens), I'm guessing we could make progress on that in short order.

Also note that I'm pretty interested in making "value components" a first-class internalized concept (ex: store the maps internally instead of using observers). That was one part of my "incremental fragmenting relations" path. Doing that would also require solving this exact problem space (as they're essentially the same implementation).

> Instead of providing a first party indexing solution that is a jack of all trades and master of none, I am going to change this PR to instead provide a consistent API that users can supply their own indexing solution into, and then provide a good-enough implementation as an example.

I like this plan. I think we should come back to built-in fragmenting indexing once we sort out the "high componentids" problem.

SanderMertens:
From the discussion yesterday, these are the numbers in Flecs when creating a component per value:

| Entities/values | Flecs component per value |
|--|--|
| 10 | 1.4 MB |
| 100 | 1.5 MB |
| 1000 | 2.5 MB | 
| 10000 | 12.8 MB |
| 50000 | 57.2 MB |
| 100000 | 113.9 MB |
| 250000 | 282.9 MB |
| 1000000 | 1.09 GB |

Can be reproduced with:
```c
#include "flecs.h"

int main() {
  ecs_world_t *world = ecs_mini();

  for (int i = 0; i < 1000 * 1000; i ++) {
    ecs_entity_t v = ecs_new(world);
    ecs_entity_t e = ecs_new(world);
    ecs_add_id(world, e, v);
  }

  while (true) {}

  return ecs_fini(world);
}
```

benches/benches/bevy_ecs/index/index_iter_indexed.rs
```diff
@@ -0,0 +1,38 @@
...
+const SPAWNS: usize = 1_000_000;
+
+#[derive(Component, Copy, Clone, PartialEq, Eq, Hash)]
+#[component(immutable)]
+struct Planet(u16);
+
+fn find_planet_zeroes_indexed(query: QueryByIndex<Planet, &Planet>) {
+    let mut query = query.at(&Planet(0));
+    for planet in query.query().iter() {
+        let _ = black_box(planet);
```
BD103, line 15:
```suggestion
        black_box(planet);
```

(Nit) The `let _ =` isn't necessary, since `black_box()` forces the expression it is passed to be evaluated.

> In practice, black_box serves two purposes:
>
> 1. It prevents the compiler from making optimizations related to the value returned by black_box
> 2. It forces the value passed to black_box to be calculated, even if the return value of black_box is unused

[`black_box()` - How to use this](https://doc.rust-lang.org/beta/std/hint/fn.black_box.html#how-to-use-this)

benches/benches/bevy_ecs/index/mod.rs
```diff
@@ -0,0 +1,38 @@
...
+mod index_update_naive;
+
+use criterion::{criterion_group, Criterion};
+
+criterion_group!(benches, index_iter, index_update,);
+
+fn index_iter(c: &mut Criterion) {
+    let mut group = c.benchmark_group("index_iter");
+    group.warm_up_time(core::time::Duration::from_millis(500));
+    group.measurement_time(core::time::Duration::from_secs(4));
```
BD103, line 13: Is there a specific reason you chose these times, which are smaller than the default? `criterion` by default has a warm-up time of 3 seconds and a measurements time of 5 seconds.

benches/benches/bevy_ecs/index/index_iter_naive.rs
```diff
@@ -0,0 +1,35 @@
...
+        let mut world = World::new();
+
+        world.spawn_batch((0..PLANETS).map(Planet).cycle().take(SPAWNS));
+
+        let id = world.register_system(find_planet_zeroes_naive);
+
+        Self(world, id)
+    }
+
+    #[inline(never)]
```
BD103, line 31: Why did you choose to add `#[inline(never)]`? It seems counter-intuitive to me, since we're just trying to benchmark the contained `run_system()`.

BD103 (approved):
I just reviewed the benchmarks, since I have no qualifications to look over the actual ECS implementation. The benchmarks overall look good, though I do have a few questions!

bushrat011899:
@alice-i-cecile [here](https://github.com/bushrat011899/bevy/commit/bfbf90dc56cb58e30fcb91d6edafb7a1548c9bdf) is my first-draft for what I'm calling "Adjustable Query Filters", which I believe supersedes this PR and provides a low-controversy path for indexing (and other features too). To summarise the feature, we introduce a new trait, `AdjustableQueryFilter`:

```rust
pub unsafe trait AdjustableQueryFilter: QueryFilter {
    type Input<'a>;

    unsafe fn adjust_filter<'a>(
        state: &mut <Self as WorldQuery>::Fetch<'a>,
        input: Self::Input<'a>,
    );
}
```

This allows `QueryIter<D, F: AdjustableQueryFilter>` to provide a method `provide_filter`:

```rust
impl<'w, 's, D: QueryData, F: AdjustableQueryFilter> QueryIter<'w, 's, D, F> {
    pub fn provide_filter(&mut self, input: F::Input<'w>) -> &mut Self {
        unsafe {
            F::adjust_filter(&mut self.cursor.filter, input);
        }

        self
    }
}
```

This allows a user to call `query.iter()` and _then_ provide additional information to the `QueryFilter` used by that iterator, with _what_ information and _why_ being entirely up to the `QueryFilter` itself. For the case of indexing, we can provide an index resource and a desired index value as inputs to the filter. To unlock the full potential of such an API, there may need to be other changes to how `QueryFilter` works. For example, my draft branch adds a `pre_filter` method, which asks the filter if a given `Archetype`/`Table` _can_ return results. This would allow an adjustable query filter to dynamically filter out entire archetypes/tables based on user input.

I have yet to publish a PR for this change, as it is far from complete and really needs an SME-ECS to work through it for potential issues.

PR #17609, opened 2025-01-30
Remove unsound SystemParam impl for DeferredWorld

ItsDoot:
# Objective

The following test currently passes (i.e. no panic) on main:

```rust
#[test]
fn deferred_world_last() {
    #[derive(Resource)]
    struct FooRes;

    fn bad_system(foo: ResMut<FooRes>, mut world: DeferredWorld) {
        let foo = foo.into_inner();
        let foo2 = world.get_resource_mut::<FooRes>().unwrap().into_inner();
    }

    let mut world = World::new();
    world.insert_resource(FooRes);
    let mut bad_system = IntoSystem::into_system(bad_system);
    bad_system.initialize(&mut world);
    bad_system.run((), &mut world);
}
```

## Solution

Remove the unsound `SystemParam` impl for `DeferredWorld`.

---

## Migration Guide

The `SystemParam` implementation for `DeferredWorld` was removed due to unsoundness. Consider using `Commands` instead.

### DIFF
--- a/crates/bevy_ecs/src/system/system_param.rs
+++ b/crates/bevy_ecs/src/system/system_param.rs
@@ -1066,27 +1066,6 @@ unsafe impl SystemParam for &'_ World {
     }
 }
 
-/// SAFETY: `DeferredWorld` can read all components and resources but cannot be used to gain any other mutable references.
-unsafe impl<'w> SystemParam for DeferredWorld<'w> {
-    type State = ();
-    type Item<'world, 'state> = DeferredWorld<'world>;
-
-    fn init_state(_world: &mut World, system_meta: &mut SystemMeta) -> Self::State {
-        system_meta.component_access_set.read_all();
-        system_meta.component_access_set.write_all();
-        system_meta.set_has_deferred();
-    }
-
-    unsafe fn get_param<'world, 'state>(
-        _state: &'state mut Self::State,
-        _system_meta: &SystemMeta,
-        world: UnsafeWorldCell<'world>,
-        _change_tick: Tick,
-    ) -> Self::Item<'world, 'state> {
-        world.into_deferred()
-    }
-}
-
 /// A system local [`SystemParam`].
 ///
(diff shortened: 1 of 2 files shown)


alice-i-cecile (approved):

JeanMertz:
Does this mean `DeferredWorld` can no longer be used in regular systems?

chescock:
I think the only thing we need to do to make it sound is to check for existing access like

```rust
if system_meta.component_access_set.has_any_component_read()
    || system_meta.component_access_set.has_any_resource_read()
{
    panic!("Useful message")
}
```

Oh, and 

```rust
system_meta.archetype_component_access.write_all();
```

Could we do that instead?

hymm:
> I think the only thing we need to do to make it sound is to check for existing access like

DeferredWorld pushes into the World's command queue, not one specific to the world and that requires exclusive access to the world.

chescock:
Well, it was a little more complicated than I expected, but I created a competing PR that tries to leave the `impl` in place but make it sound:  #17616

chescock:
> DeferredWorld pushes into the World's command queue, not one specific to the system and that requires exclusive access to the world.

Oh!  I missed that.  It might still be sound, though.  If we declare that accessing the World's command queue requires write-all-components access, then no more than one system can claim it.

alice-i-cecile:
On second thought, I'd prefer fixing this in 0.16 cycle.

hymm:
> Oh! I missed that. It might still be sound, though. If we declare that accessing the World's command queue requires write-all-components access, then no more than one system can claim it.

It might be sound, but when do the command queue get applied? I don't think there's a determined place when they ever will get applied. Just whenever the world's queue happen's to get applied. Fixing this properly would be to somehow pass a system stored command queue to the constructor of DeferredWorld.

hymm (approved):
My pref here is to remove this for now and it can just be readded once an appropriate pr shows up.

alice-i-cecile:
Closing in favor of #17616.

PR #17614, opened 2025-01-30
Fix `update_image_content_size_system` to use the camera target's scale factor

### LINKED ISSUE #17613: UI image size is only updated on changes to the primary window's scale factor
## Bevy version

main + 0.15

## What you did

With this example:

```
use bevy::{
    prelude::*, render::camera::RenderTarget, ui::widget::ImageNodeSize, window::WindowRef,
};

fn main() {
    App::new()
        .add_plugins(DefaultPlugins)
        .add_systems(Startup, setup)
        .add_systems(Update, update)
        .run();
}

    let first_window_camera = commands.spawn((Camera3d::default(),)).id();
    let image = asset_server.load("branding/icon.png");
    // Spawn a second window
    let second_window = commands
        .spawn(Window {
            title: "Second window".to_owned(),
            ..default()
        })
        .id();

    let second_window_camera = commands
        .spawn((
            Camera2d::default(),
            Camera {
                target: RenderTarget::Window(WindowRef::Entity(second_window)),
                ..default()
            },
        ))
        .id();

    commands
        .spawn((Node::default(), UiTargetCamera(first_window_camera)))
        .with_children(|commands| {
            commands.spawn((Node::default(), ImageNode::new(image.clone())));
        });

    commands
        .spawn((Node::default(), UiTargetCamera(second_window_camera)))
        .with_children(|commands| {
            commands.spawn((ImageNode::new(image.clone()),));
        });
}

fn update(image_size_query: Query<Entity, Changed<ImageNodeSize>>) {
    for entity in image_size_query.iter() {
        println!("image size updated for {}", entity.index());
    }
}
```

If you change the scale factor of the primary window it reports that both image sizes have been updated, even though the target scale factor only changed for the image on the primary window.

Or if you change the scale factor of the secondary window it doesn't update the image size for either image, even though the target scale factor changed for the image on the secondary window.

## What went wrong

The UI image size is only updated on changes to the primary window's scale factor.

ickshonpe:
# Objective

Update UI image node size on changes to its camera target's scale factor.

Fixes #17613

## Solution

Query for the camera target and use its scale factor.

Also changes the size stored in `ImageNodeSize` to physical pixels. It's only used in change detection and by using physical coords we can avoid having to track scale factor separately.

## Testing

If you run the example from #17613 you'll see that the image sizes are updated correctly now.

### DIFF
--- a/crates/bevy_ui/src/widget/image.rs
+++ b/crates/bevy_ui/src/widget/image.rs
@@ -3,14 +3,12 @@ use crate::{
 };
 use bevy_asset::{Assets, Handle};
 use bevy_color::Color;
-use bevy_ecs::entity::hash_map::EntityHashMap;
 use bevy_ecs::prelude::*;
 use bevy_image::prelude::*;
-use bevy_math::{Rect, UVec2, Vec2};
+use bevy_math::{Rect, Vec2};
 use bevy_reflect::{std_traits::ReflectDefault, Reflect};
 use bevy_render::{camera::Camera, texture::TRANSPARENT_IMAGE_HANDLE};
 use bevy_sprite::TextureSlicer;
-use bevy_window::{PrimaryWindow, Window};
 use taffy::{MaybeMath, MaybeResolve};
 
 /// A UI Node that renders an image.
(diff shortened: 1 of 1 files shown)


ickshonpe:
Fixed by the resolved target's PR

PR #17618, opened 2025-01-30
Extract UI nodes into a Vec

ickshonpe:
# Objective

Extract UI nodes into a `Vec` instead of an `EntityHashMap`.

## Solution

Extract UI nodes into a `Vec` instead of an `EntityHashMap`.
Store an index into the `Vec` in each transparent UI item.
Compare both the index and render entity in prepare so there aren't any collisions.

## Showcase

Yellow this PR, Red main

```
cargo run --example many_buttons --release --features trace_tracy
```

`extract_uinode_background_colors`
extract_uinode_background_colors

`extract_uinode_images`
extract_uinode_images

`prepare_uinodes`
prepare_uinodes_vec

### DIFF
--- a/crates/bevy_ui/src/render/box_shadow.rs
+++ b/crates/bevy_ui/src/render/box_shadow.rs
@@ -12,7 +12,6 @@ use bevy_color::{Alpha, ColorToComponents, LinearRgba};
 use bevy_ecs::prelude::*;
 use bevy_ecs::{
     prelude::Component,
-    storage::SparseSet,
     system::{
         lifetimeless::{Read, SRes},
         *,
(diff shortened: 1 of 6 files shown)


alice-i-cecile:
How does an `EntityIndexSet` compare? @Victoronz may have ideas.

mockersf (approved):

alice-i-cecile (approved):

Victoronz:
It seems that the speedup gained from just not hashing at all outweighs the slowdown from the n^2 filtering required to retrieve it again.
While `EntityIndexMap` could be faster than the `EntityHashMap` case, I think the `Vec` here would still be faster than both.
What I am however curious about is this:
Given that it was a `HashMap` before, there was no defined order, which means that if we are storing these items in a `Vec` instead, we could use the ordering to signify where an item is stored.

Instead of using indexes, you could sort the initial list of keys for which you want to push stuff by `Entity`. When retrieving, you just resort the keys, and you can retrieve the items in the same manner, no filtering required.
That is what I would guess to be fastest here!

alice-i-cecile:
Leaving Victoronz suggestions for follow-up; they're great but this is a clear huge win.

PR #17626, opened 2025-01-31
Add a test for direct recursion in required components.

ElliottjPierce:
I realized there wasn't a test for this yet and figured it would be trivial to add. Why not? Unless there was a test for this, and I just missed it?

I appreciate the unique error message it gives and wanted to make sure it doesn't get broken at some point. Or worse, endlessly recurse.

### DIFF
--- a/crates/bevy_ecs/src/lib.rs
+++ b/crates/bevy_ecs/src/lib.rs
@@ -2662,6 +2662,18 @@ mod tests {
         World::new().register_component::<A>();
     }
 
+    #[test]
+    #[should_panic = "Recursive required components detected: A → A\nhelp: Remove require(A)."]
+    fn required_components_self_errors() {
+        #[derive(Component, Default)]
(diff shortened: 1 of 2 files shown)


mockersf (approved):

Carter0 (approved):

crates/bevy_ecs/src/lib.rs
```diff
@@ -2662,6 +2662,18 @@ mod tests {
...
+    fn required_components_self_errors() {
+        #[derive(Component, Default)]
+        #[require(A)]
+        struct A;
+
+        World::new().register_component::<A>();
+    }
+
+    // These structs are primarily compilation tests to test the derive macros. Because they are
+    // never constructed, we have to manually silence the `dead_code` lint.
```
alice-i-cecile, line 2676:
```suggestion
```

alice-i-cecile (approved):

PR #6659, opened 2022-11-17
Opt out change detection

james7132:
# Objective
Allow users to opt out of change detection at the type level. Change detection comes with a performance cost on every mutative query. Disabling change detection at the type level can act as a performance optimization for all queries it's involved in.

## Solution
Add the constant `CHANGE_DETECTION_ENABLED` to `Component` and have it default to true. Extend the `Component` derive macro to allow easy configuration of this constant. Use this constant to return constant `false` results to change detection queries.

```rust
#[derive(Component)]
#[component(change_detection = false)]
pub struct ChangeDetectionless;
```

Disabling change detection will make `Mut::is_added`, `Mut::is_changed`, `Added<T>` queries, and `Changed<T>` all to return false. Any query using the filters will return empty results.

ZSTs will always make `Mut::set_changed` a no-op, even if change detection is enabled.

### Follow-up

 * Disable change detection on all render world types.
 * Disable storage based on this constant (if possible without incurring a performance penalty).

Issue #4882, opened 2022-05-31
Change Detection as Components / Opt-in or opt-out of Change Detection


james7132:
## What problem does this solve or what need does it fill?
Currently change detection is hard-coded from userspace all the way to ECS storage. This adds quite a bit of extra effort to maintain it at every level in bevy_ecs. Logically speaking, the change ticks included are always kept parallel to the base storage, and can be seen as a companion component that is always present with every component in the World. If this is the case, it may make sense to actually make the change detection ticks their own component.

This can allow for disabling change detection entirely for types which can remove the memory and CPU overhead for small component types that don't really need it (i.e. usize newtypes or ZSTs). 

This would also remove all storage-level special casing required for change detection. For example, moving entities from table to table doesn't require knowledge of the change ticks. It's copied from column to column like any other component.

## What solution would you like?

- Remove all special casing for `ComponentTicks` in storage and add `ChangeTicks<T: Component>` as an engine provided component.
- Add Component hooks that automatically add or remove the ticks if the component is added/removed. This can be done via hooks or be the only hard-coded part of this.
- Update `Mut` to fetch both `&mut T` and `&mut ChangeTicks<T>`.
- Add an associated type to `Component` that changes `WriteFetch` to retrieve `&mut T` when change detection is disabled, and `Mut<T>` when it isn't. Update the derive macro to enable/disable this behavior.
- Update change detection fetches/filters to rely on `Component<Write=Mut<Self>>`

## What alternative(s) have you considered?

1. Leave change detection as it is.
2. Alter storage to use `Option<Vec<UnsafeCell<ComponentTicks>>>` instead.

alice-i-cecile:
My initial instinct is that the component-based design is going to be the most maintainable form of this.

As for default behavior, ideally I'd like opt-out for most things, but opt-in for ZSTs.

bjorn3:
Isn't a component based approach less efficient for sparse storage?

james7132:
Arguably. We're still randomly accessing the change ticks now. Though this does add one additional SparseArray lookup in that path.

james7132:
> As for default behavior, ideally I'd like opt-out for most things, but opt-in for ZSTs.

I would actually argue for opt-in by default, but have it explicitly enabled on all user-facing first-party bevy components (i.e. Transform). In a lot of my own game code, I use the change detection filters and checks surprisingly rarely in the default case.

bjorn3:
[...]

PR #17629, opened 2025-02-01
As-needed change detection

SpecificProtagonist:
# Objective

Make change detection optional in a runtime-configurable way.
See #4882.
For prior art, also see #6659, which statically determined whether query items would be change-detection-enabled based on the type of the component.

## Solution

Querying for `&mut T` now gives you a `Mut<T>` (newly introduced type) instead of a `RefMut<T>` (previously called `Mut`). This still writes change ticks, but doesn't let you read them. Resources are unchanged.

By default, change detection is disabled for all components. It can be manually enabled via `World::enable_change_detection::<T>()`. Using a query with a `Ref<T>`, `Mut<T>`, `Added<T>` or `Changed<T>` term also enables change detection for `T` if not yet enabled, but prints a warning if entities with this component exist (as their change ticks were missed).

Updating change ticks via `MutMarkChanges` does not branch: When change detection for `T` is disabled, they're instead written to a dummy sink.

This PR doesn't turn change ticks into components of their own, which can happen in a followup if desired.

|typed|mutation|tick read|old type  |new type           |old query term  |new query term|
|-----|--------|---------|----------|-------------------|----------------|--------------|
|❌   |❌      |❌       |Ptr       |Ptr                |—               |—             |
|❌   |❌      |✅       |          |                   |—               |—             |
|❌   |✅      |❌       |          |MutUntyped         |—               |—             |
|❌   |✅      |✅       |MutUntyped|RefMutUntyped      |—               |—             |
|✅   |❌      |❌       |&T        |&T                 |&T              |&T            |
|✅   |❌      |✅       |Ref\<T\>  |Ref\<T\>           |Ref\<T\>        |Ref\<T\>      |
|✅   |✅      |❌       |          |Mut\<T\>           |                |&mut T        |
|✅   |✅      |✅       |Mut\<T\>  |RefMut\<T\>        |&mut T, Mut\<T\>|RefMut\<T\>   |

Many methods that previously worked with `Mut`/`MutUntyped` now have both a version with and without read access to change ticks (this is responsible for much of the line count; if you review this PR, going commit by commit helps). There's likely some opportunity for deduplication there, but I wanted to go with the simple solution for the initial PR.

This design should be easy to expand to coarse-grained (per-table) ticks, but doesn't make multi-level or equality based change detection easier.

## Status

This is in draft. I haven't pushed quite all the changes described above, but because of several ecs changes that have been merged since, it will be easier to recreate the changes than to rebase them.

Most of the diff (and growth in line count) is because many `_mut` methods have been split into a `_mut` and `_mut_with_ticks` variant. Can and should we prevent this duplication by making them generic over this – e.g. instead of `get_mut<T: Component<Mutability = Mutable>>(…` and `get_mut_with_ticks<T: Component<Mutability = Mutable>>(…` we would have `get_mut<T: Component<Mutability = Mutable>, const TICKS: bool>(`?

Besides this, there is more important work to do:
- Performance
  - Benchmark this!
  - Does this allow autovectorization? Should the dummy sink be turned into an array of 4 or 8 or so change ticks to help with that? 
  - Work on parallel iteration so multiple threads aren't fighting over ownership of the same cache line
- Is this the user interface we want – are there footguns (e.g. with `SystemRegistry`), bad edge cases, or unwieldiness?
- Add change tick access for dynamic queries & ReflectComponent
- Consider splitting enabling of added and changed ticks
- Consider pointer type which optionally lets you read change ticks, if present. Use this to reduce API duplication.
- Adjust APIs that expect change tracking to always be present, write new tests, write documentation incl. migration guide

## Testing

(todo)

## Showcase

```rust
let id = world.register_component::<Foo>();
assert!(!world.components().get_info(id).unwrap().change_detection());
```

## Migration Guide
(todo)

### DIFF
--- a/crates/bevy_ecs/src/world/entity_ref.rs
+++ b/crates/bevy_ecs/src/world/entity_ref.rs
@@ -4420,7 +4420,7 @@ pub unsafe trait DynamicComponentFetch {
 // - No mutable references are returned by `fetch_ref`.
 unsafe impl DynamicComponentFetch for ComponentId {
     type Ref<'w> = Ptr<'w>;
-    type Mut<'w> = MutUntyped<'w>;
+    type Mut<'w> = MutMarkChangesUntyped<'w>;
 
     unsafe fn fetch_ref(
         self,
@@ -4445,7 +4445,7 @@ unsafe impl DynamicComponentFetch for ComponentId {
 // - No mutable references are returned by `fetch_ref`.
 unsafe impl<const N: usize> DynamicComponentFetch for [ComponentId; N] {
     type Ref<'w> = [Ptr<'w>; N];
-    type Mut<'w> = [MutUntyped<'w>; N];
+    type Mut<'w> = [MutMarkChangesUntyped<'w>; N];
 
     unsafe fn fetch_ref(
         self,
@@ -4467,7 +4467,7 @@ unsafe impl<const N: usize> DynamicComponentFetch for [ComponentId; N] {
 // - No mutable references are returned by `fetch_ref`.
 unsafe impl<const N: usize> DynamicComponentFetch for &'_ [ComponentId; N] {
     type Ref<'w> = [Ptr<'w>; N];
-    type Mut<'w> = [MutUntyped<'w>; N];
+    type Mut<'w> = [MutMarkChangesUntyped<'w>; N];
 
     unsafe fn fetch_ref(
         self,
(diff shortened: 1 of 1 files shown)


alice-i-cecile:
> I'd also be happy to hear about better name suggestions – e.g. should Mut be renamed to something like MutWithTicks and MutMarkChanges to Mut (and which one would be easier for migration)?

I would prefer this flavor of design: the "nice" name should be the one returned by default.

PR #17631, opened 2025-02-01
anti-alias outside the edges of UI nodes, not across them

### LINKED ISSUE #17561: UI Node with border radius causes background pixels to come through on the inside of rounded corners
## Bevy version

from 0.14 (where rounded corners were introduced) until the current main branch

```
SystemInfo { os: "macOS 15.2 Sequoia", kernel: "24.2.0", cpu: "Apple M2", core_count: "8", memory: "16.0 GiB" }
```

Although I believe this bug exists on every device.

## What you did

Spawn a Node with a border radius that is bigger than the border size, so that the inside of the border become rounded. 

Example `main.rs` to reproduce (tested on main branch but should also work with bevy 15.x):
```
use bevy::{
    color::palettes::css::{BLUE, GREEN, WHITE},
    prelude::*,
};

fn main() {
    App::new()
        .add_plugins(DefaultPlugins)
        .add_systems(Startup, setup)
        .run();
}

    commands.spawn(Camera2d);
    commands
        .spawn((
            Node {
                width: Val::Percent(100.),
                height: Val::Percent(100.),
                ..default()
            },
            BackgroundColor(WHITE.into()),
        ))
        .with_children(|parent| {
            parent.spawn((
                Node {
                    width: Val::Percent(80.0),
                    height: Val::Percent(80.0),
                    border: UiRect::all(Val::Px(20.)),
                    ..default()
                },
                BackgroundColor(BLUE.into()),
                BorderColor(GREEN.into()),
                BorderRadius::all(Val::Px(30.0)),
            ));
        });
}
```

## What went wrong

If we zoom in to the edges, you can see that the background comes through
Image
Image

While i'd expect it to look like this
Image
Image

The issue becomes more apparent with more contrasting colors, and especially once you start zooming in.

## Cause of the issue
As a bit of background info, a Node with a border gets rendered by having 2 rectangles: 1 for the border (green part of my example images) and 1 for the background (blue part of my example images).

The problem is that the rectangle for the background is not quite big enough to fully reach the anti-aliased pixels of the inside of the border corners.

## What I tried to do to solve this issue
1) **Remove anti-aliasing of the background rectangle** 
I did this by using a `t` value of 1.0 instead of any pixel that has a higher `t` value of 0 in the current code. Here `t` refers to `let t` in the relevant fragment shader: `crates/bevy_ui/src/render/ui.wgsl`. Even with this, there was still some bleed-through.

2) **Make the background rectangle bigger** 
I increased the size of the background rectangle by halve of the border size. This fixes the issue if the border has no transparency (which is how I got the images of what I expect a fixed version to look like), but it creates another issue for borders with transparency: the overlapping part of the border and the background make another border (so now there's unintentionally two borders!).

## Proposed solution
Unify the two separate rectangles of border and background in to 1 rectangle. This way we have fine grained control over the pixel values and we can properly calculate the color without the background coming through.

I don't think there is any real impact or migration guide necessary for this approach.

Once I get some eyes on this to confirm that this is the right approach, I'd love to implement this myself.

## Related issues
- probably related to #14456

ickshonpe:
# Objective

Fixes #17561

## Solution

The anti-aliasing function used by the UI fragment shader is this:
```wgsl
fn antialias(distance: f32) -> f32 {
    return saturate(0.5 - distance);      // saturate clamps between 0 and 1
}
```
The returned value is  multiplied with the alpha channel value to get the anti-aliasing effect.

The `distance` is a signed distance value. A positive `distance` means we are outside the shape we're drawing and a negative `distance` means we are on the inside. 

So with `distance` at `0` (on the edge of the shape):
```
antialias(0) = saturate(0.5 - 0) = saturate(0.5) = 0.5
```
but we want it to be `1` at this point, so the entire interior of the shape is given a solid colour, and then decrease as the signed distance increases.

So in this PR we change it to:
```wgsl
fn antialias(distance: f32) -> f32 {
    return saturate(1. - distance);
}
```
Then:
```
antialias(-0.5) = saturate(1 - (-1)) = saturate(2) = 1
antialias(1) = saturate(1 - 0) = 1
antialias(0.5) = saturate(1 - 0.5) = 0.5
antialias(1) = saturate(1 - 1) = 0
```
as desired.

## Testing

```cargo run --example button```

On main:
bleg

With this PR:
bbwhite

Modified the `button` example to draw a white background to make the bleeding more obvious.

### DIFF
--- a/crates/bevy_ui/src/render/ui.wgsl
+++ b/crates/bevy_ui/src/render/ui.wgsl
@@ -117,7 +117,7 @@ fn sd_inset_rounded_box(point: vec2<f32>, size: vec2<f32>, radius: vec4<f32>, in
 // get alpha for antialiasing for sdf
 fn antialias(distance: f32) -> f32 {
     // Using the fwidth(distance) was causing artifacts, so just use the distance.
-    return saturate(0.5 - distance);
+    return saturate(1.0 - distance);
 }
 
 fn draw(in: VertexOutput, texture_color: vec4<f32>) -> vec4<f32> {


hukasu (approved):
e57f73207ef5afa233eeea5311b59b8522659946
image

419f98fa7ce10fd5780c4df64c881e0439efc1cd
image

hukasu:
on the description, in the `Then:` section, there is `antialias(1)` twice, where one of them should be `antialias(0)`

mockersf (approved):

hymm:
Your logic isn't correct. The pixels that are directly on the sdf edge are half in the shape and half out of the shape. See this image
image.

https://mortoray.com/antialiasing-with-a-signed-distance-field/

ickshonpe:
Yeah realised this sucks. I was thinking that pushing out the size half a pixel was a good tradeoff, but it also gives rounded corners an uglier sharper angle where the corner begins. I experimented with a couple of other hacks but maybe it's better just to accept the gaps for now and users can set a background color to fill them if needed.

PR #17633, opened 2025-02-01
Respect viewport position in coordinate conversion functions

### LINKED ISSUE #16200: `Camera::viewport_to_world_2d` doesn't take into account the viewport position
## Bevy version

0.14.2

## What you did

Set the camera viewport rect to have a minumum X/Y other than zero.

## What went wrong

`Camera::viewport_to_world_2d` gives an incorrect world position.

In order to overcome this, I have to apply the effect of the rect min before using the function. Here's an example where I am setting the X position of something based on the mouse cursor position:

```
// because of centralized viewport, viewport_to_world_2d needs the position modified according to the left edge of the viewport
// viewport_to_world_2d doesn't take viewport rect into account
if let Some(position) = windows
    .iter()
    .next()
    .and_then(|w| w.cursor_position())
    .and_then(|p| {
        camera.viewport_to_world_2d(
            camera_transform,
            p - camera
                .logical_viewport_rect()
                .and_then(|r| Some(r.min.x))
                .unwrap_or_default() as f32,
        )
    })
{
    transform.translation.x = position.x;
}
```

## Additional information

N/A, this is quite a simple issue.

Novakasa:
# Objective

- In `Camera::viewport_to_world_2d`, `Camera::viewport_to_world`, `Camera::world_to_viewport` and `Camera::world_to_viewport_with_depth`, the results were incorrect when the `Camera::viewport` field was configured with a viewport position that was non-zero. This PR attempts to correct that.
- Fixes #16200 

## Solution

- This PR now takes the viewport position into account in the functions mentioned above.
- Extended `2d_viewport_to_world` example to test the functions with a dynamic viewport position and size, camera positions and zoom levels. It is probably worth discussing whether to change the example, add a new one or just completely skip touching the examples.

## Testing

Used the modified example to test the functions with dynamic camera transform as well as dynamic viewport size and position.

### DIFF
--- a/examples/2d/2d_viewport_to_world.rs
+++ b/examples/2d/2d_viewport_to_world.rs
@@ -1,12 +1,24 @@
-//! This example demonstrates how to use the `Camera::viewport_to_world_2d` method.
+//! This example demonstrates how to use the `Camera::viewport_to_world_2d` method with a dynamic viewport and camera.
 
-use bevy::{color::palettes::basic::WHITE, prelude::*};
+use bevy::{
+    color::palettes::{
+        basic::WHITE,
+        css::{GREEN, RED},
+    },
+    math::ops::powf,
+    prelude::*,
+    render::camera::Viewport,
+};
 
 fn main() {
     App::new()
         .add_plugins(DefaultPlugins)
         .add_systems(Startup, setup)
-        .add_systems(Update, draw_cursor)
+        .add_systems(FixedUpdate, controls)
+        .add_systems(
+            PostUpdate,
+            draw_cursor.after(TransformSystem::TransformPropagate),
+        )
         .run();
 }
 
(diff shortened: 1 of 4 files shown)


alice-i-cecile:
> ## Migration Guide
> Not sure if this does this need a migration guide entry, since this can probably be considered a bugfix. 
> 
> This PR changes the result of the functions `Camera::viewport_to_world_2d`, `Camera::viewport_to_world`, `Camera::world_to_viewport` and `Camera::world_to_viewport_with_depth` in the case when the viewport position is not `UVec2::ZERO`. With a viewport position of `UVec2::ZERO`, the behaviour is not changed.

Good migration guide, but yes, bug fix :)

examples/2d/2d_viewport_to_world.rs
```diff
@@ -1,12 +1,16 @@
 //! This example demonstrates how to use the `Camera::viewport_to_world_2d` method.
 
-use bevy::{color::palettes::basic::WHITE, prelude::*};
+use bevy::render::camera::Viewport;
+use bevy::{
+    color::palettes::{basic::WHITE, css::GREEN},
+    prelude::*,
+};
```
mockersf, line 7:
```suggestion
use bevy::{
    color::palettes::{basic::WHITE, css::GREEN},
    prelude::*,
    render::camera::Viewport,
};
```

tychedelia (changes requested):
Thanks for taking the time to update the example to make sure this doesn't regress. The example looks a little funny on macOS, likely due to window scale factor, although seems to be functional. I'd suggest testing with scale factor = 2.0 to emulate macOS.

image

Novakasa:
Polished the example quite a bit
- Initialize viewport relative to physical window size (should hopefully fix MacOS weirdness)
- Controls now in `FixedUpdate` and respect fixed delta time
- Simplified viewport bounds visualization with a background mesh
- Draw cursor after Transform propagation to get rid of camera control lag
- Add comments

Novakasa:
@tychedelia would appreciate another test of the example, since I think the scale_override behaviour one Windows might be bit different than MacOS defaults.

tychedelia (approved):
Looks great and even more improved! Thanks.

Novakasa:
Rebased & fixed merge conflict, switched from `get_single()` to `single()`

Technically, it can be argued that these functions should rather be called `window_to_world_` rather than `viewport_to_world_`, but I would be fine with either, not sure if it would be worth introducing a breaking change over.

alice-i-cecile (approved):

Issue #17351, opened 2025-01-13
`SpriteBatch`es can be overwritten in `prepare_sprite_image_bind_groups`


ickshonpe:
## Bevy version 
Main + 0.15 

## What you did
```
fn main() {
    App::new()
        .add_plugins(DefaultPlugins.set(WindowPlugin {
            primary_window: Some(Window {
                resolution: WindowResolution::new(1000., 1000.).with_scale_factor_override(1.),
                title: "Window 1".to_string(),
                ..default()
            }),
            ..default()
        }))
        .add_systems(Startup, setup)
        .run();
}

fn setup(mut commands: Commands) {
    // The primary window is the default camera target and 0 is the default render layer, there's no need to set these normally
    // the explicitness is just to make this example as clear as possible.

    commands.spawn((
        Camera2d::default(),
        Camera {
            target: RenderTarget::Window(WindowRef::Primary),
            ..default()
        },
        RenderLayers::layer(0),
    ));

    // Spawn a second window
    let window_2 = commands
        .spawn(Window {
            title: "Window 2".to_owned(),
            resolution: WindowResolution::new(1000., 1000.).with_scale_factor_override(1.),
            ..default()
        })
        .id();

    commands.spawn((
        Camera2d::default(),
        Camera {
            target: RenderTarget::Window(WindowRef::Entity(window_2)),
            ..default()
        },
        RenderLayers::layer(1),
    ));

    // Text drawn to window 1
    commands.spawn((Text2d::new("1"), RenderLayers::layer(0)));

    // Text drawn to window 2
    commands.spawn((Text2d::new("2"), RenderLayers::layer(1)));

    commands.spawn((
        Text2d::new("3"),
        Transform::from_translation(-25. * Vec3::Y),
        RenderLayers::from_layers(&[0, 1]),
    ));
}
```

## What went wrong

In the second window it should display 2 and 3, but sometimes it displays the wrong glyphs.

## Additional information

Font's use a texture atlas, so there is only one texture.
So each render phase generates a single `SpriteBatch`.
Then each `SpriteBatch` is inserted on the render entity for the first phase item in that batch.
This means that if the phase item corresponding to the "3" glyph is the first item in each batch, both `SpriteBatch`s will be inserted on the same render entity, overwriting one of the batches. 
The "3" glyph sprite is visible in both views so that single `SpriteBatch` is then drawn for both views.

#
It also seems odd how the same sprite batch goes to each view as well.

[...]

PR #17636, opened 2025-02-01
Move sprite batches to resource

### LINKED ISSUE #16867: the render layers: they are broken
## Bevy version

0.15

## What you did

1. I rendered a bird into RenderLayers 0 and another into RenderLayers 0 and 1.
2. Camera A has RenderLayers 0
3. Camera B has RenderLayers 1 and a RenderTarget Image.

## What went wrong

We expect to see both birds on Camera A and only one bird on Camera B, which is drawn as an ImageNode to display the RenderTarget contents.

Instead we see... a random combination of birds on both cameras. ?!?

Example code is here: https://github.com/dead-money/bevy_render_layers_broken

Every few seconds the example despawns everything and recreates the scene. You'll see arbitrary results.

tychedelia:
# Objective

Currently, `prepare_sprite_image_bind_group` spawns sprite batches onto an individual representative entity of the batch. This poses significant problems for multi-camera setups, since an entity may appear in multiple phase instances.

## Solution

Instead, move batches into a resource that is keyed off the view and the representative entity. Long term we should switch to mesh2d and use the existing BinnedRenderPhase functionality rather than naively queueing into transparent and doing our own ad-hoc batching logic.

Fixes #16867, #17351

## Testing

Tested repros in above issues.

### DIFF
--- a/crates/bevy_sprite/src/render/mod.rs
+++ b/crates/bevy_sprite/src/render/mod.rs
@@ -10,6 +10,7 @@ use bevy_core_pipeline::{
         TonemappingLuts,
     },
 };
+use bevy_derive::{Deref, DerefMut};
 use bevy_ecs::{
     prelude::*,
     query::ROQueryItem,
@@ -19,7 +20,7 @@ use bevy_image::{BevyDefault, Image, ImageSampler, TextureAtlasLayout, TextureFo
 use bevy_math::{Affine3A, FloatOrd, Quat, Rect, Vec2, Vec4};
 use bevy_platform_support::collections::HashMap;
 use bevy_render::sync_world::MainEntity;
-use bevy_render::view::RenderVisibleEntities;
+use bevy_render::view::{RenderVisibleEntities, RetainedViewEntity};
 use bevy_render::{
     render_asset::RenderAssets,
     render_phase::{
@@ -483,7 +484,10 @@ pub struct SpriteViewBindGroup {
     pub value: BindGroup,
 }
 
-#[derive(Component, PartialEq, Eq, Clone)]
+#[derive(Resource, Deref, DerefMut, Default)]
+pub struct SpriteBatches(HashMap<(RetainedViewEntity, MainEntity), SpriteBatch>);
+
+#[derive(PartialEq, Eq, Clone)]
 pub struct SpriteBatch {
     image_handle_id: AssetId<Image>,
     range: Range<u32>,
@@ -616,8 +620,6 @@ pub fn prepare_sprite_view_bind_groups(
 }
 
 pub fn prepare_sprite_image_bind_groups(
-    mut commands: Commands,
-    mut previous_len: Local<usize>,
     render_device: Res<RenderDevice>,
     render_queue: Res<RenderQueue>,
     mut sprite_meta: ResMut<SpriteMeta>,
@@ -627,6 +629,7 @@ pub fn prepare_sprite_image_bind_groups(
     extracted_sprites: Res<ExtractedSprites>,
     mut phases: ResMut<ViewSortedRenderPhases<Transparent2d>>,
     events: Res<SpriteAssetEvents>,
+    mut batches: ResMut<SpriteBatches>,
 ) {
     // If an image has changed, the GpuImage has (probably) changed
     for event in &events.images {
@@ -640,7 +643,7 @@ pub fn prepare_sprite_image_bind_groups(
         };
     }
 
-    let mut batches: Vec<(Entity, SpriteBatch)> = Vec::with_capacity(*previous_len);
+    batches.clear();
 
     // Clear the sprite instances
     sprite_meta.sprite_instance_buffer.clear();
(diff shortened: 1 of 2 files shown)


tychedelia:
I guess this is technically a breaking change but can probably be safely picked into `0.15.2`. Your call @alice-i-cecile.

hukasu:
Ran codes provided by each of the linked issues and they seem to work as intended now

hukasu (commented):
While i was looking for info on the issues, i saw line 565
```
let index = extracted_sprite.original_entity.unwrap_or(*entity).index();
```
should the value inside the `unwrap_or` be `main_entity.id()`?

tychedelia:
> While i was looking for info on the issues, i saw line 565
> 
> ```
> let index = extracted_sprite.original_entity.unwrap_or(*entity).index();
> ```
> 
> should the value inside the `unwrap_or` be `main_entity.id()`?

Yes, the `original_entity` field in `ExtractedSprite` should likely be `MainEntity`, but that's unrelated to the issue here and mostly a correctness issue (it's easy to accidentally use a `RenderEntity` for something that's keyed off of `MainEntity`, e.g.).

crates/bevy_sprite/src/render/mod.rs
```diff
@@ -650,7 +653,8 @@ pub fn prepare_sprite_image_bind_groups(
 
     let image_bind_groups = &mut *image_bind_groups;
 
-    for transparent_phase in phases.values_mut() {
+    for (retained_view, transparent_phase) in phases.iter_mut() {
+        let mut batch_key = None;
```
hukasu, line 657: I dislike how the way the flow of the method requires that this is initialized with `None`, i would personally collect all distinct phase items into something like a `HashMap<AssetId<Image>, Vec<Entity>>`

superdump: Without having reminded myself of how this code works, not having thought enough about what you mean, using HashMap to hold collections of entities with not-highly-optimised keys seems likely to introduce significant performance regressions.

superdump:
Do you have any Tracy traces to show the performance of bevymark using the sprite mode with say 100-150k sprites?

crates/bevy_sprite/src/render/mod.rs
```diff
@@ -483,7 +484,10 @@ pub struct SpriteViewBindGroup {
     pub value: BindGroup,
 }
 
-#[derive(Component, PartialEq, Eq, Clone)]
+#[derive(Resource, Deref, DerefMut, Default)]
+pub struct SpriteBatches(HashMap<(RetainedViewEntity, MainEntity), SpriteBatch>);
```
superdump, line 488: Do we have an optimised hash of a pair of entities yet? If not then perhaps we need one? In the degenerate case with one batch per sprite I suppose this would hurt. But, sprites should use sprite sheets so that should not happen for large numbers of sprites in practice.

tychedelia: I think using `EntityHash` should work no matter how many entities you use in the key. I'll include that instead of the default hasher and confirm.

tychedelia:
This appears to be solidly faster, which is somewhat surprising to me but maybe the cost of batch spawning is a lot greater than even all the hashing?

image

crates/bevy_sprite/src/render/mod.rs
```diff
@@ -771,7 +776,7 @@ pub fn prepare_sprite_image_bind_groups(
             transparent_phase.items[batch_item_index]
                 .batch_range_mut()
                 .end += 1;
-            batches.last_mut().unwrap().1.range.end += 1;
+            batches.get_mut(&batch_key.unwrap()).unwrap().range.end += 1;
```
ickshonpe, line 779: Could add a `current_batch` variable containing a reference to the current `SpriteBatch` in the sprite batches map and avoid the extra hashmap look up here each iteration.

crates/bevy_sprite/src/render/mod.rs
tychedelia: This didn't seem to matter in my testing.

crates/bevy_sprite/src/render/mod.rs
tychedelia: This was a substantial improvement, now compared to main:

image

crates/bevy_sprite/src/render/mod.rs
```diff
@@ -690,8 +694,7 @@ pub fn prepare_sprite_image_bind_groups(
                     });
 
                 batch_item_index = item_index;
-                batches.push((
-                    item.entity(),
+                current_batch = Some(batches.entry((*retained_view, item.main_entity())).insert(
```
ickshonpe, line 697: Do we need to use the whole `RetainedViewEntity` as a key, it seems like just using `retained_view.main_entity` would be enough to avoid collisions?

tychedelia: I think it probably would be fine, especially because this is not yet retained and we are rebuilding this every frame, but we're moving towards using `RetainedViewEntity` everywhere for safety reasons. In my brief benching, using just the `MainEntity` didn't seem to have a notable performance impact.

hukasu (approved):

ickshonpe (approved):
Looks good. The batches aren't overwritten any more, the example from #17351 works correctly now.

PR #14343, opened 2024-07-15
Export glTF skins as a Gltf struct

barsoosayque:
# Objective

- Make skin data of glTF meshes available for users, so it would be possible to create skinned meshes without spawning a scene.
- I believe it contributes to https://github.com/bevyengine/bevy/issues/13681 ?

## Solution

- Add a new `GltfSkin`, representing skin data from a glTF file, new member `skin` to `GltfNode` and both `skins` + `named_skins` to `Gltf` (a la meshes/nodes).
- Rewrite glTF nodes resolution as an iterator which sorts nodes by their dependencies (nodes without dependencies first). So when we create `GltfNodes` with their associated `GltfSkin` while iterating, their dependencies already have been loaded.
- Make a distinction between `GltfSkin` and `SkinnedMeshInverseBindposes` in assets: prior to this PR, `GltfAssetLabel::Skin(n)` was responsible not for a skin, but for one of skin's components. Now `GltfAssetLabel::InverseBindMatrices(n)` will map to `SkinnedMeshInverseBindposes`, and `GltfAssetLabel::Skin(n)` will map to `GltfSkin`.

## Testing

- New test `skin_node` does just that; it tests whether or not `GltfSkin` was loaded properly.

## Migration Guide

- Change `GltfAssetLabel::Skin(..)` to `GltfAssetLabel::InverseBindMatrices(..)`.

PR #17641, opened 2025-02-02
Fix Maya-exported rigs by not trying to topologically sort glTF nodes.

pcwalton:
The code added in #14343 seems to be trying to ensure that a `Handle` for each glTF node exists by topologically sorting the directed graph of glTF nodes containing edges from parent to child and from skin to joint. Unfortunately, such a graph can contain cycles, as there's no guarantee that joints are descendants of nodes with the skin. In particular, glTF exported from Maya using the popular babylon.js export plugin create skins attached to nodes that animate their parent nodes. This was causing the topological sort code to enter an infinite loop.

Assuming that the intent of the topological sort is indeed to ensure that `Handle`s exist for each glTF node before populating them, there's a better mechanism for this: `LoadContext::get_label_handle`. This is the documented way to obtain a handle for a node before populating it, obviating the need for a topological sort. This patch replaces the topological sort with a pre-pass that uses
`LoadContext::get_label_handle` to get handles for each `Node` before populating them. This fixes the problem with Maya rigs, in addition to making the code simpler and faster.

### DIFF
--- a/crates/bevy_gltf/Cargo.toml
+++ b/crates/bevy_gltf/Cargo.toml
@@ -59,6 +59,8 @@ gltf = { version = "1.4.0", default-features = false, features = [
 ] }
 thiserror = { version = "2", default-features = false }
 base64 = "0.22.0"
+fixedbitset = "0.5"
+itertools = "0.13"
 percent-encoding = "2.1"
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1"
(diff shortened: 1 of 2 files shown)


bushrat011899 (approved):
What a great fix, nice work!

tychedelia (approved):

pcwalton:
The tests failed because I wasn't checking for cycles. I added code to do that.

mockersf (approved):

PR #17645, opened 2025-02-02
Improve cubic segment bezier functionality

### LINKED ISSUE #17642: `CubicSegment` lacks functionality of `CubicCurve`

`CubicCurve`, which is a `Vec<CubicSegment>`, can be used in many ways, e.g. creating roads and paths.

However, in many cases `CubicSegment` is very useful by itself. For example, I want to check if my entity will not collide with anything, while following a curved path. But the check is still local enough, so I don't need the whole `CubicCurve`

**Performance considerations.** 

`CubicCurve` contains a `Vec`, which is heap allocated. It is expensive, if spawned every frame for many entities.

`CubicSegment`, on the other hand, is fixed size, which is ideal for such scenario.

However, the [implementation](https://github.com/bevyengine/bevy/blob/082d87141fdf7951ed04dc3550d6542f4a9da07f/crates/bevy_math/src/cubic_splines/mod.rs#L1035) of `CubicSegment::new_bezier` includes creating a `CubicCurve` of size 1, only to take the first element later. It is still an allocation, and in this case, an unnecessary one.

1. I want be able to create `CubicSegment` from 4 control points, not just 2. Right now it seems like it is just intended for animation easing purposes.
2. I would like it to be as fast as can be expected from creating a fixed size struct.
3. I would like `CubicSegment::iter_positions` and other cool functions that `CubicCurve` has

I implemented my own hacky trait to have this functionality:

```rust
pub trait CubicSegment4 {
    fn new_bezier_4(p1: impl Into<Vec2>, p2: impl Into<Vec2>, p3: impl Into<Vec2>, p4: impl Into<Vec2>) -> Self;

    fn iter_positions(self, subdivisions: usize) -> impl Iterator<Item=Vec2>;
}

impl CubicSegment4 for CubicSegment<Vec2> {
    fn new_bezier_4(p1: impl Into<Vec2>, p2: impl Into<Vec2>, p3: impl Into<Vec2>, p4: impl Into<Vec2>) -> Self {
        // A derivation for this matrix can be found in "General Matrix Representations for B-splines" by Kaihuai Qin.
        // <https://xiaoxingchen.github.io/2020/03/02/bspline_in_so3/general_matrix_representation_for_bsplines.pdf>
        // See section 4.2 and equation 11.
        let char_matrix = [
            [1., 0., 0., 0.],
            [-3., 3., 0., 0.],
            [3., -6., 3., 0.],
            [-1., 3., -3., 1.],
        ];
        let p: [Vec2; 4] = [p1.into(), p2.into(), p3.into(), p4.into()];
        coefficients(p, char_matrix)
    }

    fn iter_positions(self, subdivisions: usize) -> impl Iterator<Item=Vec2> {
        let step = 1.0 / subdivisions as f32;
        (0..=subdivisions).map(move |i| i as f32 * step)
        .map(move |t| self.position(t))
    }
}

fn coefficients<P: VectorSpace>(p: [P; 4], char_matrix: [[f32; 4]; 4]) -> CubicSegment<P> {
    let [c0, c1, c2, c3] = char_matrix;
    // These are the polynomial coefficients, computed by multiplying the characteristic
    // matrix by the point matrix.
    let coeff = [
        p[0] * c0[0] + p[1] * c0[1] + p[2] * c0[2] + p[3] * c0[3],
        p[0] * c1[0] + p[1] * c1[1] + p[2] * c1[2] + p[3] * c1[3],
        p[0] * c2[0] + p[1] * c2[1] + p[2] * c2[2] + p[3] * c2[3],
        p[0] * c3[0] + p[1] * c3[1] + p[2] * c3[2] + p[3] * c3[3],
    ];
    CubicSegment::<P> { coeff }
}
```

It does the thing without memory allocations. But has some duplicated code.

Example usage:
```rust
pub fn get_curve(
    pos_1: Vec2,
    pos_2: Vec2,
    anchor: Option<Vec2>
) -> CubicSegment<Vec2> {
    if let Some(anchor) = anchor {
        // Curve with anchor point
        CubicSegment::new_bezier_4(
            pos_1,
            pos_1.lerp(anchor, 0.66),
            pos_2.lerp(anchor, 0.66),
            pos_2
        )
    } else {
        // Straight line
        CubicSegment::new_bezier_4(
            pos_1,
            pos_1.lerp(pos_2, 0.33),
            pos_1.lerp(pos_2, 0.66),
            pos_2
        )
    }
}
```

## Additional context

Looking at code, looks like `CubicSegment` is being sort of ignored in favor of the `CubicCurve`. I think many functions like the `new_bezier` and `new_bezier_4` should really be implemented on the base structure.

hocop:
# Objective

- Fixes #17642

## Solution

- Implemented method `new_bezier(points: [P; 4]) -> Self` for `CubicSegment<P>`
- Old implementation of `new_bezier` is now `new_bezier_easing(p1: impl Into<Vec2>, p2: impl Into<Vec2>) -> Self` (**breaking change**)
- ~~added method `new_bezier_with_anchor`, which can make a bezier curve between two points with one control anchor~~
- added methods `iter_positions`, `iter_velocities`, `iter_accelerations`, the same as in `CubicCurve` (**copied code, potentially can be reduced)**
- bezier creation logic is moved from `CubicCurve` to `CubicSegment`, removing the unneeded allocation

## Testing

  - Run tests inside `crates/bevy_math/`
  - Tested the functionality in my project
  - Did not run `cargo test` on the whole bevy directory because of OOM
  - Performance improvements are expected when creating `CubicCurve` with `new_bezier` and `new_bezier_easing`, but not tested
  - Use in any code that works created `CubicCurve::new_bezier`
  - I don't think relevant

---

## Showcase

```rust
// Imagine a car goes towards a local target

// Create a simple `CubicSegment`, without using heap
let planned_path = CubicSegment::new_bezier([
    car_pos,
    car_pos + car_dir * turn_radius,
    target_point - target_dir * turn_radius,
    target_point,
]);

// Check if the planned path itersect other entities
for pos in planned_path.iter_positions(8) {
   // do some collision checks
}
```

## Migration Guide

- Replace `CubicCurve::new_bezier` with `CubicCurve::new_bezier_easing`


crates/bevy_math/src/cubic_splines/mod.rs
```diff
@@ -994,6 +984,34 @@ impl<P: VectorSpace> CubicSegment<P> {
...
+    /// This function can also be used to approximate a quarter circle,
+    /// if start, anchor and end points form a right isosceles triangle.
+    pub fn new_bezier_with_anchor(start: P, anchor: P, end: P) -> Self {
+        Self::new_bezier([
+            start,
+            start.lerp(anchor, 0.667),
+            end.lerp(anchor, 0.667),
+            end,
+        ])
+    }
```
mweatherley, line 1013: Is this a well-known construction? It looks kind of random to me. (Also, it is very easy to implement in user-space)

crates/bevy_math/src/cubic_splines/mod.rs
```diff
@@ -994,6 +984,34 @@ impl<P: VectorSpace> CubicSegment<P> {
...
+        // <https://xiaoxingchen.github.io/2020/03/02/bspline_in_so3/general_matrix_representation_for_bsplines.pdf>
+        // See section 4.2 and equation 11.
+        let char_matrix = [
+            [1., 0., 0., 0.],
+            [-3., 3., 0., 0.],
+            [3., -6., 3., 0.],
+            [-1., 3., -3., 1.],
+        ];
+        Self::coefficients(points, char_matrix)
+    }
```
mweatherley, line 999: Moving these closer to `CubicSegment` is a good idea, especially with the `no_std` changes. My only gripe here is that I think that, in this case, we should probably refactor out *all* of these characteristic matrices so that the spline constructions use methods on `CubicSegment` instead of calling `CubicSegment::coefficients` directly. Hermite and Cardinal spline constructions would still be useful with only a single segment.

Of course, you're under no obligation to do that in this PR. :)

crates/bevy_math/src/cubic_splines/mod.rs
```diff
@@ -1015,6 +1033,46 @@ impl<P: VectorSpace> CubicSegment<P> {
...
+
+    /// Iterate over the curve split into `subdivisions`, sampling the velocity at each step.
+    pub fn iter_velocities(&self, subdivisions: usize) -> impl Iterator<Item = P> + '_ {
+        self.iter_samples(subdivisions, Self::velocity)
+    }
+
+    /// Iterate over the curve split into `subdivisions`, sampling the acceleration at each step.
+    pub fn iter_accelerations(&self, subdivisions: usize) -> impl Iterator<Item = P> + '_ {
+        self.iter_samples(subdivisions, Self::acceleration)
+    }
```
mweatherley, line 1075: Not a fan of the code duplication, but I guess that's kind of unavoidable 🤔 

I think there's a reasonable shot that all of these will be outmoded by the `Curve` API implementations in the future, but for now it's more ergonomic, I guess. 

(E.g. one right now can write something like
```rs
cubic_segment.with_two_derivatives().map(|j| j.second_derivative).samples(n)
```
to similar effect, and the intention is for this to become more ergonomic in the near future.)

mweatherley (changes requested):

mweatherley:
You can delete the entire `cfg_attr` block at like 1016 if you want CI to stop complaining :)

crates/bevy_math/src/cubic_splines/mod.rs
hocop: This is actually just a quadratic curve, expressed as a cubic one. https://stackoverflow.com/questions/3162645/convert-a-quadratic-bezier-to-a-cubic-one

I agree that it is easy to implement. It is more of a proposal from me. I find it useful, but I don't know if many people will.

If, as you say, the api is going to be changed in the near future, then I guess there is no need to add many small functions. I am inclined to remove this one then

crates/bevy_math/src/cubic_splines/mod.rs
hocop: Agree, but I don't think I'm ready to do it now. I haven't used these other curves, so it will take some time to figure out how to not break anything. Maybe I'll open another PR later, if no one does it :)

crates/bevy_math/src/cubic_splines/mod.rs
hocop: Maybe it's better to move these functions to a separate trait, and implement it on both `CubicCurve` and `CubicSegment`. This can help avoid some duplication. What do you think?

Carter0 (approved):
I think the code and doc comments look good. You might want to add some labels to your PR though if you can. Sometimes people skip over PR's if they don't have labels.

villor (approved):
Glad to see these changes! 😍 I’ll finally be able to get rid of some copy paste workarounds in [bevy_lookup_curve](https://github.com/villor/bevy_lookup_curve/blob/96dde7a47c42d14aa0505d3cb900fcd6b729f596/src/lib.rs#L538)

Another thing on the wishlist is a way to pass custom `MAX_ERROR` and/or `MAX_ITERS` to `CubicSegment<Vec2>::ease`

IQuick143 (approved):

PR #17648, opened 2025-02-02
Add user supplied mesh tag

### LINKED ISSUE #13373: User provided mesh binding / per-instance mesh data

A common pattern in creative coding is to use the instance index in a shader in order to sample from something like a texture or read from a SSBO. For example, the shader might use one texture in order to offset the position of the mesh and a second texture in order to determine the vertex color. 

Currently, batch rendered entities (i.e. entities that share the same mesh and material) are instanced in a random order relative to how they were spawned. This is due a variety of causes:
- Instability in ECS queries.
- GPU preprocessing compute shader.
- Order of rendering bins.

Consequently, because the instance index is "random", it's not possible in a custom shader to know where to sample from an input texture. 

See the following pseudo UV maps to see what this looks like with different rendering features enabled:

The user should be able to provide a custom binding representing per-instance mesh data, i.e. a SSBO that they can index into with the instance index to retrieve additional data required for their vertex shader.

The difficult part here is that in order to write into the user data buffer, it needs to be plumbed everywhere the `MeshUniform` is created, including our new preprocessing compute shader. In order not to pollute all our rendering code, this means we need some kind of generic plugin for instance data that gets type erased in our rendering code so that it can optionally injected everywhere it needs to go.

We could, alternatively, track some kind of total "spawn order" and add this as a new index to `MeshUniform`. The user could then bind their buffer to their material and index into that instead. This has a few disadvantages:
- Adding a `u32` to every mesh uniform that is likely unnecessary for most users.
- Requiring all possible data to be input to the material even though many of the instances may be culled.
- Difficult to track when the spawn order should be when meshes are despawned and respawned.

## Additional context

There's prior art here I'm happy to reference in terms of how this typically works in creative coding applications.

tychedelia:
# Objective

Because of mesh preprocessing, users cannot rely on `@builtin(instance_index)` in order to reference external data, as the instance index is not stable, either from frame to frame or relative to the total spawn order of mesh instances.

## Solution

Add a user supplied mesh index that can be used for referencing external data when drawing instanced meshes.

Closes #13373

## Testing

Benchmarked `many_cubes` showing no difference in total frame time.

## Showcase

https://github.com/user-attachments/assets/80620147-aafc-4d9d-a8ee-e2149f7c8f3b


assets/shaders/mesh_instance_index.wgsl
```diff
@@ -0,0 +1,41 @@
...
+fn vertex(vertex: Vertex) -> VertexOutput {
+    var out: VertexOutput;
+    let mesh_instance_index = mesh_functions::get_mesh_instance_index(vertex.instance_index);
+    var world_from_local = mesh_functions::get_world_from_local(vertex.instance_index);
+    out.world_position = mesh_functions::mesh_position_local_to_world(world_from_local, vec4(vertex.position, 1.0));
+    out.clip_position = position_world_to_clip(out.world_position.xyz);
+
+    let tex_dim = textureDimensions(texture);
+    // Find the texel coordinate from the mesh_instance_index
+let texel_coord = vec2<u32>(mesh_instance_index % tex_dim.x, mesh_instance_index / tex_dim.x);
```
IceSentry, line 30:
```suggestion
    let texel_coord = vec2<u32>(mesh_instance_index % tex_dim.x, mesh_instance_index / tex_dim.x);
```

examples/shader/storage_buffer.rs
```diff
@@ -39,22 +38,20 @@ fn setup(
...
-    commands.insert_resource(CustomMaterialHandles(material_handles.clone()));
+    commands.insert_resource(CustomMaterialHandle(material_handle.clone()));
 
     // Spawn cubes with the custom material
-    let mut current_color_id = 0;
+    let mut current_color_id: u32 = 0;
     for i in -6..=6 {
         for j in -3..=3 {
             commands.spawn((
                 Mesh3d(meshes.add(Cuboid::from_size(Vec3::splat(0.3)))),
```
IceSentry, line 52: Does it still work if you clone the cube mesh?

pcwalton (commented):
In the decal code I have a similar feature, but I called it the "tag". I think that might be a less confusing name than "instance index", because with your code as it currently stands "instance index" has two separate meanings.

IceSentry (approved):
I have no idea if the slightly higher memory usage is something we want to care about or now, but other than that this is a really useful PR and the implementation is nice and simple.

IceSentry:
Yes, I agree that the name is a bit confusing though.

tychedelia:
Okay, I think "tag" is a nice alternative for sure especially if it's already used in the engine, I'll switch to that instead 👍 .

examples/shader/mesh_instance_index.rs
```diff
@@ -0,0 +1,90 @@
+//! This examples demonstrates using the `MeshInstanceIndex` component to sample from a
```
IceSentry, line 1: I wonder if this should just be merged with the automatic_instancing example?

Just add a few comments to explain that the new component is optional.

IceSentry: This one is a lot prettier though, so I'm happy if it keeps the look of it. Just also add the comments explaining the automatic instancing bits.

tychedelia: I think that's a good idea actually, as it will be more discoverable and is often what users are curious about when they reach for instancing.

JMS55:
I somewhat wish we could instead always have a tag corresponding to an Entity, and have the instance_index builtin be used to index into a list of tags. So you do `draw instance_index within a phase -> tags[instance_index] -> transforms/other_data[tag]`.

I think we've discussed this on discord in the past, don't remember the conclusion we came to.

tychedelia:
> I somewhat wish we could instead always have a tag corresponding to an Entity, and have the instance_index builtin be used to index into a list of tags. So you do draw instance_index within a phase -> tags[instance_index] -> transforms/other_data[tag].

This definitely makes sense as a design and could help unify some of these patterns. @IceSentry brought up shader mesh picking as an example of something that could be useful for the engine to ship. I'd be happy if someone implemented that but it will require deeper incisions into things and make my primary use case more complicated. This implementation is still very general and can be used to implement other similar things (e.g. write the entity index into the tag) as well. I do think storing the entity on the gpu may be somewhat inevitable but I'd like to unblock this first, especially to see what other user needs arise.

examples/shader/automatic_instancing.rs
```diff
@@ -1,51 +1,97 @@
...
     App::new()
-        .add_plugins(DefaultPlugins)
+        .add_plugins((DefaultPlugins, MaterialPlugin::<CustomMaterial>::default()))
         .add_systems(Startup, setup)
+        .add_systems(Update, update)
         .run();
 }
 
-/// set up a simple 3D scene
+/// Set up our scene
```
IceSentry, line 24: This should probably have quick comment explaining the end result people should expect to see when running the example. It's not super obvious otherwise why it starts by loading a texture.

examples/shader/automatic_instancing.rs
```diff
@@ -1,51 +1,97 @@
 //! Shows that multiple instances of a cube are automatically instanced in one draw call
 //! Try running this example in a graphics profiler and all the cubes should be only a single draw call.
```
IceSentry, line 2: I just did a quick sanity check and it is indeed still a single draw call 🎉

examples/shader/automatic_instancing.rs
```diff
@@ -1,51 +1,102 @@
...
+    mut materials: ResMut<Assets<CustomMaterial>>,
 ) {
-    // camera
+    // We will use this image as our external data for our material to sample from in the vertex shader
+    let image = assets.load("branding/icon.png");
+
+    // Our single mesh handle that will be instanced
+    let mesh_handle = meshes.add(Cuboid::from_size(Vec3::splat(0.01)));
+
+    // Create the custom material with a reference to our texture
```
IceSentry, line 42:
```suggestion
    // Create the custom material with a reference to our texture
    // Automatic instancing works with any Material, including the `StandardMaterial`.
    // This custom material is used to demonstrate the optional `MeshTag` feature.
```

Maybe add something like this too? Just to make sure people understand there's nothing special about the custom material in terms of automatic instancing.

crates/bevy_pbr/src/render/mesh.rs
```diff
@@ -536,8 +538,8 @@ pub struct MeshInputUniform {
     /// Low 16 bits: index of the material inside the bind group data.
     /// High 16 bits: index of the lightmap in the binding array.
     pub material_and_lightmap_bind_group_slot: u32,
-    /// Padding.
-    pub pad_a: u32,
+    /// User supplied tag to identify this mesh instance.
+    pub tag: u32,
     /// Padding.
     pub pad_b: u32,
```
pcwalton, line 544: Probably should change this to just `pad` now that there's only 1 field.

crates/bevy_render/src/experimental/occlusion_culling/mesh_preprocess_types.wgsl
```diff
@@ -19,7 +19,8 @@ struct MeshInput {
     // Low 16 bits: index of the material inside the bind group data.
     // High 16 bits: index of the lightmap in the binding array.
     material_and_lightmap_bind_group_slot: u32,
-    pad_a: u32,
+    // User supplied index to identify the mesh instance
+    tag: u32,
     pad_b: u32,
```
pcwalton, line 24: Ditto, change to `pad`

pcwalton (approved):
Works for me.

Issue #17647, opened 2025-02-02
`WorldQuery` types for working with relationships


alice-i-cecile:
## What problem does this solve or what need does it fill?

Working with relationships in Bevy is cumbersome, requiring multiple queries, manual iteration and repeated calls to Query::get to correlate the information about the relationship graph with the data you actually want to fetch.

This is particularly annoying in the context of exclusive world access, as keeping multiple queries alive at once requires the use of `SystemState`.

Many of these systems follow very common patterns: surely we can write simple abstractions over them.

## What solution would you like?

Add 4 query filter types:

- [ ] `RelatedTo<R: Relationship, F: QueryFilter>`: is this entity related to an entity that matches the query filter? 
- [ ] `Scan<RT: RelationshipTarget, F: QueryFilter>`: do any of the entities which are related to this entity match the query filter?
- [ ] `WithAncestor<R: Relationship, F: QueryFilter>`: is this entity transitively related to any entities (searching up recursively) that match the given query filter?
- [ ] `WithDescendant<RT: RelationshipTarget, F: QueryFilter>`: do any of the entities which are transitively related to this entity (searching down recursively)  match the query filter?

**Implementation note:** both WithAncestor and WithDescendant are naively O(N^2). We can cache the decisions made and check the cache to reduce that to O(N). That doesn't have to be in the first PR.

Add an initial query data type:

- [ ] `Related<R: Relationship, D: QueryData>`: fetch the data of type `D` from the entity that this entity is related to via `R`.

Note that `Option<Related<R, D>>` is extremely useful for looking up data if and only if it exists.

Finally, add a type alias

- [ ] `type Parent<D: QueryData> = Related<R: ChildOf, D: QueryData>`

## What alternative(s) have you considered?

Users who need these helpers can continue to work with multiple queries per system. This is not very pleasant, and often obscures meaning.

Alternatively, these helpers should generally be implementable outside of bevy_ecs. Doing so is a non-trivial amount of work, and requires fairly complex unsafe code.

## Additional context

These query helper types build on the work in #17398; while they are intended to be ergonomic, pleasant to work with and forward-compatible, performance is a secondary concern. Archetypal relations (as seen in flecs, or proposed as part of #17564) would likely improve the speed of many of these operations.

These types were first designed in https://hackmd.io/UCyLaF4YStC0cTSE8Fm7cw, and discussed [on Discord](https://discord.com/channels/691052431525675048/1334239065599512718/1334239208826732666).

[...]

PR #17649, opened 2025-02-02
Add a `RelatedTo` filter for querying relations

alice-i-cecile:
# Objective

Looking for entities that have a parent (or other relationship) where the parent meets some filter is a common task.

Part of #17647.

## Solution

Make it easier by adding a `RelatedTo` type which implements `QueryFilter`.

Tasks:

- [x] create the basic type
- [x] successfully implement WorldQuery and QueryFilter
- [x] add comprehensive tests
- [x] extend documentation 
- [ ] ensure tests pass and implementation is correct

## Testing

I've added a number of unit tests (and a simple doc test) checking this implementation against various forms of query filters.

### DIFF
--- a/crates/bevy_ecs/src/relationship/relationship_query.rs
+++ b/crates/bevy_ecs/src/relationship/relationship_query.rs
@@ -390,13 +390,16 @@ unsafe impl<R: Relationship, F: QueryFilter> WorldQuery for RelatedTo<R, F> {
         table_row: TableRow,
     ) -> Self::Item<'w> {
         // Look up the relationship
-        let relation =
-            <&'static R as WorldQuery>::fetch(&mut fetch.relation_fetch, entity, table_row);
+        // SAFETY: the safety requirements for calling `fetch` on `R` are a subset of the safety requirements for calling this method
+        let relation = unsafe {
+            <&'static R as WorldQuery>::fetch(&mut fetch.relation_fetch, entity, table_row)
+        };
         // Then figure out what the related entity is
         let related_entity = relation.get();
 
         // Finally, check if the related entity matches the filter
-        <F as WorldQuery>::fetch(&mut fetch.filter_fetch, related_entity, table_row)
+        // SAFETY: the safety requirements for calling `fetch` on `F` are a subset of the safety requirements for calling this method
+        unsafe { <F as WorldQuery>::fetch(&mut fetch.filter_fetch, related_entity, table_row) }
     }
 
     fn update_component_access(
--- a/crates/bevy_ecs/src/relationship/relationship_query.rs
+++ b/crates/bevy_ecs/src/relationship/relationship_query.rs
@@ -507,6 +507,27 @@ mod tests {
     struct A;
 
     #[test]
+    fn related_to_empty_filter() {
+        let mut world = World::default();
+        let parent = world.spawn_empty().id();
+        let child = world.spawn(ChildOf(parent)).id();
+        let _unrelated = world.spawn_empty().id();
+        let grandchild = world.spawn(ChildOf(child)).id();
+
+        let mut query_state = world.query_filtered::<Entity, RelatedTo<ChildOf, ()>>();
+        for matching_entity in query_state.iter(&world) {
+            let matches_child_or_grandchild =
+                matching_entity == child || matching_entity == grandchild;
+            assert!(
+                matches_child_or_grandchild,
+                "Entity {matching_entity} should have a parent"
+            );
+        }
+
+        assert_eq!(query_state.iter(&world).count(), 2);
+    }
+
+    #[test]
     fn related_to() {
         let mut world = World::default();
         let entity = world.spawn(A).id();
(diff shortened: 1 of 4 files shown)


alice-i-cecile:
Currently, `WorldQuery::shrink` is complaining about:
> lifetime may not live long enough
> consider adding the following bound: `'wshort: 'wlong`

And `WorldQuery::fetch` is complaining about:

> lifetime may not live long enough
> consider adding the following bound: `'wshort: 'wlong`
> requirement occurs because of the type `RelatedToFetch<'_, R, F>`, which makes the generic argument `'_` invariant
> the struct `RelatedToFetch<'w, R, F>` is invariant over the parameter `'w`
> see <https://doc.rust-lang.org/nomicon/subtyping.html> for more information about variance

Obviously, neither of those suggestions are good ('short must be less than 'long!), but I'm not immediately sure how to fix it.

chescock:
> Currently, `WorldQuery::shrink` is complaining about:
> 
> > lifetime may not live long enough
> > consider adding the following bound: `'wshort: 'wlong`
> 
> And `WorldQuery::fetch` is complaining about:

I think you need to call `F::shrink(item)` and do something like 
```rust
RelatedToFetch { 
  relation_fetch: <&R>::shrink_fetch(fetch.relation_fetch),
  filter_fetch: F::shrink_fetch(fetch.filter_fetch),
}
```

The implementations that have concrete types can just return `item` and `fetch` because the compiler knows the concrete types are covariant.  Here we have associated types on generic parameters, so the compiler assumes they might be invariant and we have to call the `shrink` methods explicitly.

crates/bevy_ecs/src/relationship/relationship_query.rs
```diff
@@ -270,3 +276,356 @@ where
...
+                matches_child_or_grandchild,
+                "Entity {matching_entity} should have a parent"
+            );
+        }
+
+        assert_eq!(query_state.iter(&world).count(), 2);
+    }
+
+    #[test]
+    fn related_to_with() {
```
alice-i-cecile, line 538: Fails.

crates/bevy_ecs/src/relationship/relationship_query.rs
```diff
@@ -270,3 +276,356 @@ where
...
+        let parent = world.spawn(A).id();
+        let child = world.spawn(ChildOf(parent)).id();
+        let mut query_state = world.query_filtered::<Entity, RelatedTo<ChildOf, With<A>>>();
+        let fetched_child = query_state.iter(&world).next().unwrap();
+
+        assert_eq!(child, fetched_child);
+    }
+
+    #[test]
+    fn related_to_same_with() {
```
alice-i-cecile, line 549: Passes?!

crates/bevy_ecs/src/relationship/relationship_query.rs
```diff
@@ -270,3 +276,356 @@ where
...
+            world.query_filtered::<Entity, (With<A>, RelatedTo<ChildOf, With<A>>)>();
+        let fetched_child = query_state.iter(&world).next().unwrap();
+
+        assert_eq!(child, fetched_child);
+    }
+
+    // If this test fails but the one above passes,
+    // the RelatedTo filter is probably not matching enough archetypes.
+    #[test]
+    fn related_to_different_with() {
```
alice-i-cecile, line 563: Also fails.

crates/bevy_ecs/src/relationship/relationship_query.rs
```diff
@@ -270,3 +276,356 @@ where
...
+    }
+
+    fn get_state(components: &crate::component::Components) -> Option<Self::State> {
+        Some(RelatedFilterState {
+            relation_state: <&'static R as WorldQuery>::get_state(components)?,
+            filter_state: <F as WorldQuery>::get_state(components)?,
+        })
+    }
+
+    fn matches_component_set(
```
alice-i-cecile, line 428: Given the test pattern, I suspect that this method is wrong, but I can't figure out why.

alice-i-cecile (commented):
Alright, I think I've figured out the core of what's going wrong here and why some tests pass but others fail. Take this with a grain of salt and correct me: I'm not particularly familiar with manual implementations of `WorldQuery`.

The key pair of tests to look at is `related_to_with` (fails) and `related_to_with_same_archetype` (passes?!).

In order for `WorldQuery::fetch` to work correctly, we must first set the table / archetype correctly, via `WorldQuery::set_table`. Unfortunately, the right table is *different* based on whether we're looking at the base entity (e.g. child) or the target entity (e.g. parent). This means that the case demonstrated in the `related_to_with` component fails to match the right archetype, and `QueryFilter::filter_fetch` / `WorldQuery::fetch` is never called at all. I tested this by sticking a panic in the `fetch` method: it's not hit for the failing tests.

However, when we share components between the base and target entities, we can accidentally look at the right archetypes, and then `WorldQuery::fetch` functions as intended.

As for the fix, I'm really not sure. `WorldQuery` is *really* designed to operate on only a single entity at once, hence the set_archetype / set_table design. We could relax that, but I worry that the performance characteristics would be untenable. We can find out though!

We could instead swap this over to a `SystemParam` type, but that's really not much better than the existing query helpers: the `WorldQuery` impl here / in the linked issue are key to the power and ergonomics.

P.S. I don't think that the API as designed here would work with archetypal relations either 🤔 The core problem is that we need to look at a different table halfway through `WorldQuery::fetch`, and that method doesn't have access to the parameters required to call `set_archetype` and/or `set_table`.

PR #17654, opened 2025-02-03
Improve docs for `WorldQuery`

alice-i-cecile:
# Objective

While working on #17649, I found the docs for `WorldQuery` and the related traits frustratingly vague.

## Solution

Clarify them and add some more tangible advice.

Also fix a copy-pasted typo in related comments.

### DIFF
--- a/crates/bevy_ecs/src/query/filter.rs
+++ b/crates/bevy_ecs/src/query/filter.rs
@@ -431,7 +431,7 @@ macro_rules! impl_or_query_filter {
             unsafe fn init_fetch<'w>(world: UnsafeWorldCell<'w>, state: &Self::State, last_run: Tick, this_run: Tick) -> Self::Fetch<'w> {
                 let ($($filter,)*) = state;
                 ($(OrFetch {
-                    // SAFETY: The invariants are uphold by the caller.
+                    // SAFETY: The invariants are upheld by the caller.
                     fetch: unsafe { $filter::init_fetch(world, $filter, last_run, this_run) },
                     matches: false,
                 },)*)
@@ -444,7 +444,7 @@ macro_rules! impl_or_query_filter {
                 $(
                     $filter.matches = $filter::matches_component_set($state, &|id| table.has_column(id));
                     if $filter.matches {
-                        // SAFETY: The invariants are uphold by the caller.
+                        // SAFETY: The invariants are upheld by the caller.
                         unsafe { $filter::set_table(&mut $filter.fetch, $state, table); }
                     }
                 )*
@@ -462,7 +462,7 @@ macro_rules! impl_or_query_filter {
                 $(
                     $filter.matches = $filter::matches_component_set($state, &|id| archetype.contains(id));
                     if $filter.matches {
-                        // SAFETY: The invariants are uphold by the caller.
+                        // SAFETY: The invariants are upheld by the caller.
                        unsafe { $filter::set_archetype(&mut $filter.fetch, $state, archetype, table); }
                     }
                 )*
@@ -475,7 +475,7 @@ macro_rules! impl_or_query_filter {
                 table_row: TableRow
             ) -> Self::Item<'w> {
                 let ($($filter,)*) = fetch;
-                // SAFETY: The invariants are uphold by the caller.
+                // SAFETY: The invariants are upheld by the caller.
                 false $(|| ($filter.matches && unsafe { $filter::filter_fetch(&mut $filter.fetch, entity, table_row) }))*
             }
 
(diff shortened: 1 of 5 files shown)


bushrat011899 (approved):

crates/bevy_ecs/src/query/world_query.rs
```diff
@@ -59,7 +63,8 @@ pub unsafe trait WorldQuery {
     /// This function manually implements subtyping for the query fetches.
     fn shrink_fetch<'wlong: 'wshort, 'wshort>(fetch: Self::Fetch<'wlong>) -> Self::Fetch<'wshort>;
 
-    /// Creates a new instance of this fetch.
+    /// Creates a new instance of [`Self::Fetch`](WorldQuery::Fetch),
+    /// by combining data from the [`World`] with the cashed [`Self::State`](WorldQuery::State).
```
james-j-obrien, line 67:
```suggestion
    /// by combining data from the [`World`] with the cached [`Self::State`](WorldQuery::State).
```

james-j-obrien (approved):

crates/bevy_ecs/src/query/filter.rs
```diff
@@ -93,6 +93,11 @@ pub unsafe trait QueryFilter: WorldQuery {
     ///
     /// Note that this is called after already restricting the matched [`Table`]s and [`Archetype`]s to the
     /// ones that are compatible with the Filter's access.
+    /// If the implementation of [`WorldQuery::matches_component_set`] is incorrect,
```
cBournhonesque, line 96: Those 2 lines are a bit confusing to me; you mean that your `matches_component_set` was overly restrictive and was causing some tables/archetypes to be missed completely, which is why this function was not called?

crates/bevy_ecs/src/query/world_query.rs
```diff
@@ -40,12 +40,16 @@ use variadics_please::all_tuples;
 /// [`QueryFilter`]: crate::query::QueryFilter
 pub unsafe trait WorldQuery {
     /// The item returned by this [`WorldQuery`]
-    /// For `QueryData` this will be the item returned by the query.
+    /// For `QueryData` this will be the data retrieved by the query,
+    /// and is visible to the end user when calling e.g. `Query<Self>::get`.
+    ///
     /// For `QueryFilter` this will be either `()`, or a `bool` indicating whether the entity should be included
     /// or a tuple of such things.
+    /// Archetypal query filters (like `With`) set this to `()`, as the filtering is done by selecting the archetypes to iterate over,
```
cBournhonesque, line 48: Great! i might even add that the archetypal query filters rely on `matches_component_set` while the non-archetypal rely on `filter_fetch`

chescock: > archetypal query filters rely on `matches_component_set` while the non-archetypal rely on `filter_fetch`

Non-archetypal filters rely on *both*, right?  `Changed<T>` filters to archetypes that contain a `T`.

chescock, line 48: It's kind of weird that `QueryFilter`s even have an `Item`.  I think it's mostly because they used to be a single trait, and the change to separate them tried to minimize changes.  

Would be more clear if we moved `Item` and `fetch` from `WorldQuery` to `QueryData`, and then did all the filtering in `filter_fetch`?  (Not in this PR, of course.)  It doesn't look like `fetch` is ever called on a `QueryFilter` outside its own implementation, so that might work.

crates/bevy_ecs/src/query/filter.rs
```diff
@@ -93,6 +93,11 @@ pub unsafe trait QueryFilter: WorldQuery {
...
+    /// this method may never be called, and your query filter may fail to check entities that should be included.
+    ///
+    /// Implementors of this method will generally either have a trivial `true` body (for archetypal filters),
```
chescock, line 99: One thing that wasn't clear to me at first that may be worth saying explicitly: 

If `IS_ARCHETYPAL` is `true`, then `filter_fetch` must *always* return `true`.

chescock (approved):

Carter0 (approved):
This is good stuff!

crates/bevy_ecs/src/query/filter.rs
alice-i-cecile: I think that this should be removed.

crates/bevy_ecs/src/query/world_query.rs
alice-i-cecile: Spun out into #17662. Fully agree.

alice-i-cecile:
Really excellent suggestions; thank you all. Waiting on one re-review before merging.

chescock (approved):

PR #17660, opened 2025-02-03
Revert #17631

ickshonpe:
# Objective

Revert #17631

After some more experimentation, realised it's not the right approach.


alice-i-cecile:
See the bottom of that thread for comments on why it was wrong.

PR #16457, opened 2024-11-21
fix: setting content size to rect size if image has a rect

MarcoMeijer:
# Objective

When using a rect for a ui image, its content size is still equal to the size of the full image instead of the size of the rect.

## Solution

Use the rect size if it is present.

## Testing

I tested it using all 4 possible combinations of having a rect and texture atlas or not. See the showcase section.

---

## Showcase

<details>
  <summary>Click to view showcase</summary>

```rust
use bevy::prelude::*;

fn main() {
    App::new()
        .add_plugins(DefaultPlugins.set(ImagePlugin::default_nearest()))
        .add_systems(Startup, create_ui)
        .run();
}

fn create_ui(
    mut commands: Commands,
    assets: Res<AssetServer>,
    mut texture_atlas_layouts: ResMut<Assets<TextureAtlasLayout>>,
    mut ui_scale: ResMut<UiScale>,
) {
    let texture = assets.load("textures/fantasy_ui_borders/numbered_slices.png");
    let layout = TextureAtlasLayout::from_grid(UVec2::splat(16), 3, 3, None, None);
    let texture_atlas_layout = texture_atlas_layouts.add(layout);

    ui_scale.0 = 2.;

    commands.spawn(Camera2d::default());

    commands
        .spawn(Node {
            display: Display::Flex,
            align_items: AlignItems::Center,
            ..default()
        })
        .with_children(|parent| {
            // nothing
            parent.spawn(ImageNode::new(texture.clone()));

            // with rect
            parent.spawn(ImageNode::new(texture.clone()).with_rect(Rect::new(0., 0., 16., 16.)));

            // with rect and texture atlas
            parent.spawn(
                ImageNode::from_atlas_image(
                    texture.clone(),
                    TextureAtlas {
                        layout: texture_atlas_layout.clone(),
                        index: 1,
                    },
                )
                .with_rect(Rect::new(0., 0., 8., 8.)),
            );

            // with texture atlas
            parent.spawn(ImageNode::from_atlas_image(
                texture.clone(),
                TextureAtlas {
                    layout: texture_atlas_layout.clone(),
                    index: 2,
                },
            ));
        });
}
```

Before this change:
Screenshot 2024-11-21 at 11 55 45

After the change:
Screenshot 2024-11-21 at 11 54 54

</details>

PR #17664, opened 2025-02-03
added an example showing how to use with_rect

### LINKED ISSUE #16482: Demonstrate how to use ImageNode::rect
Looks good.

I think I noticed this was wrong during the UiImage required components migration but forgot about it again in the rush.

There should be a UI example that demonstrates how to use `ImageNode::rect`. Maybe you could add a bit of explanatory text to your showcase example and include it with this PR?

_Originally posted by @ickshonpe in https://github.com/bevyengine/bevy/pull/16457#pullrequestreview-2451651855_

fjkorf:
# Objective

Add an example demonstrating the use of `ImageNode::rect`
Fixes #16482 

## Solution

Modified the code in the original issue #16457

## Testing

to test run

`cargo run --example image_rect  --release`


alice-i-cecile (changes requested):
I like this, but IMO this sort of documentation on "how to use a specific API" is best suited to API docs. Examples are harder to discover and much more expensive to maintain.

Can you condense this and move it over to a doc test (and maybe some extra explanation there?).

crates/bevy_ui/src/widget/image.rs
```diff
@@ -119,6 +119,55 @@ impl ImageNode {
         self
     }
 
+    /// `with_rect` crops an `ImageNode` to the portion described by
```
alice-i-cecile, line 122:
```suggestion
    /// Crops an `ImageNode` to the portion described by
```

crates/bevy_ui/src/widget/image.rs
```diff
@@ -119,6 +119,55 @@ impl ImageNode {
...
+    /// a `Rect`. This can be applied to `ImageNode`s created from
```
alice-i-cecile, line 123:
```suggestion
    /// the provided `Rect`, measured from the top-left corner. This can be applied to `ImageNode`s created from
```
Not sure about the top-left bit; please fill in the correct answer bit of information there if it's wrong!

alice-i-cecile (commented):
Nice stuff; this is clear and helpful. I have a few style / clarity nits.

fjkorf:
Thanks!

alice-i-cecile (approved):

alice-i-cecile:
@Carter0, I'd appreciate your review here if you have a minute :)

IceSentry (approved):
I feel like the example could be rewritten with bsn but this can happen in a future PR

mockersf (approved):

PR #17668, opened 2025-02-03
Skip camera look ups in queue uinodes 

ickshonpe:
# Objective

`queue_uinodes` looks up the `ExtractedView` for every extracted UI node, but there's no need to look it up again if consecutive nodes have the same `extracted_camera_entity`.

## Solution

 In queue uinodes reuse the previously looked up extracted view if the `extracted_camera_entity` doesn't change

## Showcase

```
cargo run --example many_buttons --release --features "trace_tracy"
```

queue-ui-improvement

yellow is this PR, red is main


greeble-dev (approved):
The optimised code seems equivalent as far as I can tell - I'm not a UI expert. Tested `multiple_windows` and `split_screen` examples.

Nitpick: would have liked the PR description to mention examples that are mostly likely to reveal issues.

mockersf (approved):

PR #17671, opened 2025-02-04
Isolate component registration

ElliottjPierce:
# Objective

Progresses #17569. The end goal here is to synchronize component registration. See the other PR for details for the motivation behind that. 

For this PR specifically, the objective is to decouple `Components` from `Storages`. What components are registered etc should have nothing to do with what Storages looks like. Storages should only care about what entity archetypes have been spawned.

## Solution

Previously, this was used to create sparse sets for relevant components when those components were registered. Now, we do that when the component is inserted/spawned.

This PR proposes doing that in `BundleInfo::new`, but there may be a better place.

## Testing

In theory, this shouldn't have changed any functionality, so no new tests were created. I'm not aware of any examples that make heavy use of sparse set components either.

## Migration Guide

- Remove storages from functions where it is no longer needed.
- Note that SparseSets are no longer present for all registered sparse set components, only those that have been spawned.

### DIFF
--- a/crates/bevy_ecs/src/bundle.rs
+++ b/crates/bevy_ecs/src/bundle.rs
@@ -435,7 +435,7 @@ impl BundleInfo {
             .collect();
 
         // ensure components are valid
-        for id in component_ids.iter().cloned() {
+        for id in component_ids.iter().copied() {
             // Safety: caller ensures the ids are valid
             let info = unsafe { components.get_info_unchecked(id) };
             storages.prepare_component(info);
--- a/crates/bevy_ecs/src/bundle.rs
+++ b/crates/bevy_ecs/src/bundle.rs
@@ -151,11 +151,7 @@ use variadics_please::all_tuples;
 pub unsafe trait Bundle: DynamicBundle + Send + Sync + 'static {
     /// Gets this [`Bundle`]'s component ids, in the order of this bundle's [`Component`]s
     #[doc(hidden)]
-    fn component_ids(
-        components: &mut Components,
-        storages: &mut Storages,
-        ids: &mut impl FnMut(ComponentId),
-    );
+    fn component_ids(components: &mut Components, ids: &mut impl FnMut(ComponentId));
 
     /// Gets this [`Bundle`]'s component ids. This will be [`None`] if the component has not been registered.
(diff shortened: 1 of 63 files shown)


ElliottjPierce:
> Maybe also add a comment to that effect on `Storages.sparse_sets`?

Done! Thanks @SpecificProtagonist for the review and suggestions.

chescock (approved):
Looks good!  Thanks for pulling this out of the larger PR!

crates/bevy_ecs/src/query/fetch.rs
```diff
@@ -1105,17 +1105,8 @@ unsafe impl<T: Component> WorldQuery for &T {
             components: StorageSwitch::new(
                 || None,
                 || {
-                    // SAFETY: The underlying type associated with `component_id` is `T`,
-                    // which we are allowed to access since we registered it in `update_archetype_component_access`.
```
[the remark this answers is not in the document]
ElliottjPierce: That makes sense. After reading it again, I agree. I'll fix this shortly.

crates/bevy_ecs/src/bundle.rs
```diff
@@ -446,6 +434,13 @@ impl BundleInfo {
             })
             .collect();
 
+        // ensure components are valid
+        for id in component_ids.iter().copied() {
```
cart, line 438: We could avoid double-iterating this list by calling prepare_component during the first iteration of the "baseline" version of the list, then also calling prepare_component() when we push the required components to the list.

Slightly harder to keep correct, but this code doesn't change often and this is a hot path.

ElliottjPierce: I've gone ahead and done that. I agree that the speed is worth it.

alice-i-cecile:
Good stuff; I think this is a worthwhile cleanup regardless of the final design choice.

crates/bevy_ecs/src/bundle.rs
cart: > Not terribly hot though – it's only called once per type of bundle.

Its all relative I suppose. Spawning speed is important, and this will get much hotter in the world of "fragmenting relations" / "value components".

PR #17672, opened 2025-02-04
Use dual-source blending for rendering the sky

ecoskey:
# Objective

Since previously we only had the alpha channel available, we stored the mean of the transmittance in the aerial view lut, resulting in a grayer fog than should be expected. 

## Solution

- Calculate transmittance to scene in `render_sky` with two samples from the transmittance lut
- use dual-source blending to effectively have per-component alpha blending


alice-i-cecile:
@mate-h this one is for you :)

mate-h:
I've had an initial look at this, and using dual source blending is a great idea, the diff looks good to me. I went to test things out and noticed some inaccuracies between what I expect to see ground truth, vs the results seem to be shifting colors in the rendered geometry toward orange. 
I dug a bit deeper and it's likely an issue in either how we sample the transmittance lut and calculate the transmittance though the atmosphere for a path segment to the scene depth from the camera. 
I will continue investigating and testing this further. 

I pushed all my testing code to the https://github.com/mate-h/bevy/tree/atmosphere-test branch and will keep working on that.
Screenshot 2025-02-04 at 12 19 38 AM

mate-h:
After this last iteration and more thorough testing we were able to identify the source of the issue. I've tested both cases of the geometry above and below the horizon and the math checks out compared to the raymarched reference. 
Test case for when the ray intersects the ground:
image

Test case for when the ray does not intersect the ground (else block):
image

mate-h (approved):

crates/bevy_pbr/src/atmosphere/functions.wgsl
```diff
@@ -118,6 +118,23 @@ fn sample_transmittance_lut(r: f32, mu: f32) -> vec3<f32> {
     return textureSampleLevel(transmittance_lut, transmittance_lut_sampler, uv, 0.0).rgb;
 }
 
+//should be in bruneton_functions, but wouldn't work in that module bc imports. What to do wrt licensing?
```
mate-h, line 121: I think we should just leave a comment saying that the license in the `bruneton_functions.wgsl` file applies to this function, or come up with some way to reorganize the imports. Also just in general, since the code is not exaclty the same as bruneton's verison, it is a WGSL re-write, not sure how the licensing works because I am no legal export. but does it still apply?

crates/bevy_pbr/src/atmosphere/render_sky.wgsl
```diff
@@ -18,22 +18,30 @@
...
+    if depth == 0.0 {
+        let ray_dir_as = direction_world_to_atmosphere(ray_dir_ws.xyz);
+        transmittance = sample_transmittance_lut(r, mu);
+        inscattering += sample_sky_view_lut(r, ray_dir_as);
+        inscattering += sample_sun_illuminance(ray_dir_ws.xyz, transmittance);
     } else {
-        return sample_aerial_view_lut(in.uv, depth);
+        let t = ndc_to_camera_dist(vec3(uv_to_ndc(in.uv), depth));
+        inscattering = sample_aerial_view_lut(in.uv, t);
+        transmittance = sample_transmittance_lut_segment(r, mu, t);
```
mate-h, line 44: Just a note on performance, we should start profiling this since we are now taking 3 texture samples per fragment here to determine both inscattering and transmittance, as opposed to 1 that we used before.

ecoskey: Compared to full per-pixel tamarin I'm not too concerned. If it ends up being an issue we can always fallback to the old way

mate-h:
@alice-i-cecile this one is ready for final review and merge when you get a chance!

crates/bevy_pbr/src/atmosphere/functions.wgsl
superdump: If it's a direct port with no significant practical changes then I'd say the license still applies.

alice-i-cecile:
Needs a second rendering review :)

crates/bevy_pbr/src/atmosphere/functions.wgsl
alice-i-cecile, line 121: This needs to be resolved before we can merge. I'd leave a little comment here.

crates/bevy_pbr/src/atmosphere/functions.wgsl
```diff
@@ -130,23 +147,31 @@ fn sample_sky_view_lut(r: f32, ray_dir_as: vec3<f32>) -> vec3<f32> {
     return textureSampleLevel(sky_view_lut, sky_view_lut_sampler, uv, 0.0).rgb;
 }
 
+fn ndc_to_camera_dist(ndc: vec3<f32>) -> f32 {
```
mate-h, line 150: Note here that there is also a bevy PBR builtin for this `bevy_pbr::view_transformations::depth_ndc_to_view_z` and then the return value could be just multiplied by `settings.scene_units_to_m` Also seems to consider projection in that function.

ecoskey: I believe that function only returns the "parallel" distance from the camera to the object, i.e. `|P - C| cosθ` where `θ` is the angle between `P - C` and the camera forward vector. So we could use it, but we'd have to divide out that `cosθ` term, which isn't too difficult. There's also the issue of bindings, so we'd still have to copy the function into the local shaders

mate-h: Oh true I didn't think about the bindings, scratch that :D

tychedelia (commented):
image
image

Can still see the sun shining when it's night, which didn't seem to be the case on main.

mate-h:
Good point! There is a couple issues on main that may be unrelated to this PR:
1. The shadow maps are broken on OSX 
2. The PBR is not yet integrated to the atmosphere, I started working on that now at https://github.com/mate-h/bevy/tree/pbr-atmosphere

So we might see light on objects in this example because of shadows maps not working, and the directional light not being occluded by the atmosphere or the planet earth. As far as I can tell this has not changed compared to main.

tychedelia:
> So we might see light on objects in this example because of shadows maps not working, and the directional light not being occluded by the atmosphere or the planet earth. As far as I can tell this has not changed compared to main.

I tested this PR on macOS and main on windows so that would make sense!

alice-i-cecile (changes requested):
The comment about the license needs to be resolved before I can merge this. To resolve this:

1. Mention that this is related to the bruneton_functions
2. Credit the original author in the comment
3. Name the license that it's used under in that comment.

ecoskey:
@alice-i-cecile fixed! :)

alice-i-cecile (approved):

Issue #17670, opened 2025-02-04
Text2d glyphs sometimes do not render


rparrett:
## Bevy version

main, bisected to #17636

##  Relevant system information

SystemInfo { os: "macOS 15.3 Sequoia", kernel: "24.3.0", cpu: "Apple M4 Max", core_count: "16", memory: "64.0 GiB" }    
AdapterInfo { name: "Apple M4 Max", vendor: 0, device: 0, device_type: IntegratedGpu, driver: "", driver_info: "", backend: Metal }

## What you did

`cargo run --example text2d`

## What went wrong

The particular glyphs rendered change every frame. It's very exciting to watch.

<img width="1280" alt="Image" src="https://github.com/user-attachments/assets/4850a0a1-5736-4409-8235-c75ff9f599aa" />

ickshonpe:
Sprite slicing is also broken for the same reason in a similarly exciting fashion:
<img width="961" alt="Image" src="https://github.com/user-attachments/assets/1320de56-4b78-4266-8736-293adaf90df9" />

PR #17674, opened 2025-02-04
Fix text-2d.

tychedelia:
# Objective

Fix text 2d. Fixes https://github.com/bevyengine/bevy/issues/17670

## Solution

Evidently there's a 1:N extraction going on here that requires using the render entity rather than main entity.

## Testing

Text 2d example

### DIFF
--- a/crates/bevy_sprite/src/render/mod.rs
+++ b/crates/bevy_sprite/src/render/mod.rs
@@ -485,9 +485,9 @@ pub struct SpriteViewBindGroup {
 }
 
 #[derive(Resource, Deref, DerefMut, Default)]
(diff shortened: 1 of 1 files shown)


ickshonpe (approved):
This was my fault, I don't how I missed that it broke Text2d when I reviewed the sprite batch resource when I've spent so much time fixing Text2d bugs :sweat:

Texture slicing also does one-to-many extraction and is broken in the same way, `sprite_
slice` example:

sprite-splice



 With this PR it's perfect again:
sprite-slice

Carter0 (approved):

mockersf (approved):

PR #17675, opened 2025-02-04
Add missing return in default `Relationship::on_insert` impl

grind086:
# Objective

There was a bug in the default `Relationship::on_insert` implementation that caused it to not properly handle entities targeting themselves in relationships. The relationship component was properly removed, but it would go on to add itself to its own target component.

## Solution

Added a missing `return` and a couple of tests (`self_relationship_fails` failed on its second assert prior to this PR).

## Testing

See above.


alice-i-cecile (approved):
Good tests, and a solid explanation. Thanks :)

mockersf (approved):

PR #17677, opened 2025-02-04
Add `edit_material_on_gltf` example

hukasu:
# Objective

Create a minimal example of how to modify the material from a `Gltf`. This is frequently asked about on the help channel of the discord.

## Solution

Create the example.

## Showcase

image

### DIFF
--- a/examples/3d/edit_material_on_gltf.rs
+++ b/examples/3d/edit_material_on_gltf.rs
@@ -1,0 +1,89 @@
+//! Showcases how to change the material of a `Scene` spawned from a Gltf
+
+use bevy::{
+    app::{App, PluginGroup, Startup},
+    asset::{AssetServer, Assets},
+    audio::AudioPlugin,
+    color::{palettes, Color},
+    gltf::GltfAssetLabel,
+    math::{Dir3, Vec3},
+    pbr::{DirectionalLight, MeshMaterial3d, StandardMaterial},
+    prelude::{Camera3d, Children, Commands, Component, Query, Res, ResMut, Transform, Trigger},
+    scene::{SceneInstanceReady, SceneRoot},
+    DefaultPlugins,
+};
+
+fn main() {
+    App::new()
+        .add_plugins(DefaultPlugins.build().disable::<AudioPlugin>())
+        .add_systems(Startup, setup_scene)
+        .add_observer(change_material)
+        .run();
+}
+
+/// Overrides the color of the `StandardMaterial` of a mesh
+#[derive(Component)]
+struct ColorOverride(Color);
+
+fn setup_scene(mut commands: Commands, asset_server: Res<AssetServer>) {
+    // Camera
+    commands.spawn((
+        Camera3d::default(),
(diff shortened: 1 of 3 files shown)


examples/3d/edit_material_on_gltf.rs
```diff
@@ -0,0 +1,89 @@
...
+        .add_observer(change_material)
+        .run();
+}
+
+/// Overrides the color of the `StandardMaterial` of a mesh
+#[derive(Component)]
+struct ColorOverride(Color);
+
+fn setup_scene(mut commands: Commands, asset_server: Res<AssetServer>) {
+    // Camera
```
rparrett, line 29: These comments don't seem super useful. I'd suggest removing the camera and directional light comments and adding e.g.

```
// This scene will keep its original materials
// These scenes will have their materials overridden with the specified color
```

rparrett (changes requested):
Cool. This comes up frequently enough that I did a short blog post about it at one point. It will be nice to have an example like this to point to.

examples/3d/edit_material_on_gltf.rs
```diff
@@ -0,0 +1,94 @@
...
+
+fn main() {
+    App::new()
+        .add_plugins(DefaultPlugins.build().disable::<AudioPlugin>())
+        .add_systems(Startup, setup_scene)
+        .add_observer(change_material)
+        .run();
+}
+
+/// This is added to a [`SceneRoot`] and will cause the [`StandardMaterial::base_color`] of all materials
```
rparrett, line 24: This sentence needs work

examples/3d/edit_material_on_gltf.rs
```diff
@@ -0,0 +1,89 @@
...
+            .get(child)
+            .ok()
+            .and_then(|id| standard_materials.get_mut(id.id()))
+        {
+            let mut new_material = material.clone();
+            new_material.base_color = color_override.0;
+
+            commands
+                .entity(child)
+                .insert(MeshMaterial3d(standard_materials.add(new_material)));
```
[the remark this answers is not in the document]
rparrett: I still think that the act of adding a brand new material every time this observer fires is an important enough detail to call out more explicitly in a comment.

This is a silent performance footgun, and users likely want to find a way to reuse the materials.

examples/3d/edit_material_on_gltf.rs
```diff
@@ -0,0 +1,95 @@
...
+        if let Some(material) = mesh_materials
+            .get(descendants)
+            .ok()
+            .and_then(|id| asset_materials.get_mut(id.id()))
+        {
+            // Create a copy of the material and override base color
+            let mut new_material = material.clone();
+            new_material.base_color = color_override.0;
+
+            // Override `MeshMaterial3d` with new material
```
rparrett, line 89: Re-surfacing this comment from a now-outdated review:

I still think that the act of adding multiple brand new materials every time this observer fires is an important enough detail to call out more explicitly in a comment.
 
This is a silent performance footgun, and users likely want to find a way to reuse the materials. I feel like actually "fixing" this may be advanced / context dependent so a comment should suffice.

rparrett (approved):

alice-i-cecile (approved):

PR #17678, opened 2025-02-04
Interactable hidden overflow

ickshonpe:
# Objective

Enable interactions for areas of UI Node's made non-visible with `OverflowAxis::Hidden`.

## Solution

Replace `CalculatedClip`'s `clip` field into two new fields `visible` and `interactable`.

In `update_clipping` update and propagate both the `interactable` and `visible` fields.

Rendering intersects UI node rects with `CalculatedClip::visible` to find the visible area of the nodes.
Interaction intersects UI node rects with `CalculatedClip::interactable` to find the interactable area of the nodes.

## Showcase
Extended the `overflow` example to showcase the new behaviours:
```
cargo run --example overflow
```

overflow

## Testing

The `overflow` example above can be used for testing.

You should see that with `OverflowAxis::Hidden` interactions are enabled for the non-visible sections of the image node buttons on that axis. The sections clipped with `OverflowAxis::Clip` should be non-interactive.

## Migration Guide
`CalculatedClip` `clip: Rect` field has been split into two new fields:
```
pub struct CalculatedClip {
    /// Content outside this `Rect` should not be visible
    pub visible: Rect,
    /// Content outside this `Rect` should not be interactable
    pub interactable: Rect,
}
```
These fields can be intersected with a Node's bounding rect to find it's visible and interactable areas respectively.

### DIFF
--- a/crates/bevy_ui/src/ui_node.rs
+++ b/crates/bevy_ui/src/ui_node.rs
@@ -1093,6 +1093,11 @@ impl Overflow {
         self.x.is_visible() && self.y.is_visible()
     }
 
+    /// Overflow is interactable on both axes
+    pub const fn is_interactable(&self) -> bool {
+        self.x.is_interactable() && self.y.is_interactable()
+    }
+
     pub const fn scroll() -> Self {
         Self {
             x: OverflowAxis::Scroll,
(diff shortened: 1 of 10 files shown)


crates/bevy_ui/src/ui_node.rs
```diff
@@ -1149,6 +1154,11 @@ impl OverflowAxis {
     pub const fn is_visible(&self) -> bool {
         matches!(self, Self::Visible)
     }
+
+    /// Overflow is interactable on this axis
+    pub const fn is_interactable(&self) -> bool {
+        matches!(self, Self::Visible | Self::Hidden)
```
BenjaminBrienen, line 1160: Why is Hidden intractable?

PR #17679, opened 2025-02-04
Move `Item` and `fetch` to `QueryData` from `WorldQuery` 

### LINKED ISSUE #17662: `Item` and `fetch` should be moved to `QueryData` from `WorldQuery`
It's kind of weird that `QueryFilter`s even have an `Item`.  I think it's mostly because they used to be a single trait, and the change to separate them tried to minimize changes.  

Would be more clear if we moved `Item` and `fetch` from `WorldQuery` to `QueryData`, and then did all the filtering in `filter_fetch`?  (Not in this PR, of course.)  It doesn't look like `fetch` is ever called on a `QueryFilter` outside its own implementation, so that might work.

_Originally posted by @chescock in https://github.com/bevyengine/bevy/pull/17654#discussion_r1939602703_

couyit:
# Objective

Fixes #17662

## Solution

Moved `Item` and `fetch` from `WorldQuery` to `QueryData`, and adjusted their implementations accordingly.

Currently, documentation related to `fetch` is written under `WorldQuery`. It would be more appropriate to move it to the `QueryData` documentation for clarity.

I am not very experienced with making contributions. If there are any mistakes or areas for improvement, I would appreciate any suggestions you may have.

## Migration Guide

The `WorldQuery::Item` type and `WorldQuery::fetch` method have been moved to `QueryData`, as they were not useful for `QueryFilter` types.

### DIFF
--- a/crates/bevy_ecs/src/query/fetch.rs
+++ b/crates/bevy_ecs/src/query/fetch.rs
@@ -278,6 +278,38 @@ use variadics_please::all_tuples;
 pub unsafe trait QueryData: WorldQuery {
     /// The read-only variant of this [`QueryData`], which satisfies the [`ReadOnlyQueryData`] trait.
     type ReadOnly: ReadOnlyQueryData<State = <Self as WorldQuery>::State>;
+
+    /// The item returned by this [`WorldQuery`]
+    /// For `QueryData` this will be the data retrieved by the query,
+    /// and is visible to the end user when calling e.g. `Query<Self>::get`.
+    ///
+    /// For `QueryFilter` this will be either `()`, or a `bool` indicating whether the entity should be included
+    /// or a tuple of such things.
+    /// Archetypal query filters (like `With`) set this to `()`,
+    /// as the filtering is done by selecting the archetypes to iterate over via [`WorldQuery::matches_component_set`],
+    /// while non-archetypal query filters (like `Changed`) set this to a `bool` and evaluate the filter for each entity,
+    /// after the set of possible archetypes has been narrowed down.
+    type Item<'a>;
+
+    /// This function manually implements subtyping for the query items.
+    fn shrink<'wlong: 'wshort, 'wshort>(item: Self::Item<'wlong>) -> Self::Item<'wshort>;
+
+    /// Fetch [`Self::Item`](`WorldQuery::Item`) for either the given `entity` in the current [`Table`],
+    /// or for the given `entity` in the current [`Archetype`]. This must always be called after
+    /// [`WorldQuery::set_table`] with a `table_row` in the range of the current [`Table`] or after
+    /// [`WorldQuery::set_archetype`]  with an `entity` in the current archetype.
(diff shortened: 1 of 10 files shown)


Carter0 (approved):
Makes sense. And I think the rust compiler would help guide you through this.

crates/bevy_ecs/src/query/fetch.rs
```diff
@@ -278,6 +278,38 @@ use variadics_please::all_tuples;
...
+    /// The item returned by this [`WorldQuery`]
+    /// For `QueryData` this will be the data retrieved by the query,
+    /// and is visible to the end user when calling e.g. `Query<Self>::get`.
+    ///
+    /// For `QueryFilter` this will be either `()`, or a `bool` indicating whether the entity should be included
+    /// or a tuple of such things.
+    /// Archetypal query filters (like `With`) set this to `()`,
+    /// as the filtering is done by selecting the archetypes to iterate over via [`WorldQuery::matches_component_set`],
+    /// while non-archetypal query filters (like `Changed`) set this to a `bool` and evaluate the filter for each entity,
+    /// after the set of possible archetypes has been narrowed down.
```
chescock, line 291: We don't need to mention `QueryFilter` here anymore :)

```suggestion
    /// This will be the data retrieved by the query,
    /// and is visible to the end user when calling e.g. `Query<Self>::get`.
```

crates/bevy_ecs/src/query/fetch.rs
```diff
@@ -1713,8 +1728,26 @@ unsafe impl<'__w, T: Component> WorldQuery for Mut<'__w, T> {
 }
 
 // SAFETY: access of `Ref<T>` is a subset of `Mut<T>`
-unsafe impl<'__w, T: Component> QueryData for Mut<'__w, T> {
+unsafe impl<'__w, T: Component<Mutability = Mutable>> QueryData for Mut<'__w, T> {
```
chescock, line 1731: !!!  Wait, it's currently possible to query for `Mut<T>` on an immutable component?  Good catch!

crates/bevy_ecs/src/query/world_query.rs
```diff
@@ -14,11 +13,11 @@ use variadics_please::all_tuples;
 /// # Safety
 ///
 /// Implementor must ensure that
-/// [`update_component_access`], [`matches_component_set`], [`fetch`] and [`init_fetch`]
+/// [`update_component_access`], [`matches_component_set`], [`QueryData::fetch`] and [`init_fetch`]
```
chescock, line 16: We probably want to mention `QueryFilter::filter_fetch` alongside `QueryData::fetch` here and below.

chescock (approved):
Looks good!  Thanks for fixing this so quickly!

alice-i-cecile (approved):

alice-i-cecile:
I've added a migration guide for you :) Really excellent stuff; thanks for tackling this. I hope it was fun and/or educational :D

PR #17680, opened 2025-02-04
Make `batch_and_prepare_binned_render_phase` only record information about the first batch in each batch set.

pcwalton:
Data for the other batches is only accessed by the GPU, not the CPU, so it's a waste of time and memory to store information relating to those other batches.

On Bistro, this reduces time spent in
`batch_and_prepare_binned_render_phase` from 85.9 us to 61.2 us, a 40% speedup.

Screenshot 2025-02-04 093315

### DIFF
--- a/crates/bevy_render/src/render_phase/mod.rs
+++ b/crates/bevy_render/src/render_phase/mod.rs
@@ -189,9 +189,16 @@ pub enum BinnedRenderPhaseBatchSets<BK> {
     MultidrawIndirect(Vec<BinnedRenderPhaseBatchSet<BK>>),
 }
 
+/// A group of entities that will be batched together into a single multi-draw
+/// call.
 pub struct BinnedRenderPhaseBatchSet<BK> {
-    pub(crate) batches: Vec<BinnedRenderPhaseBatch>,
+    /// The first batch in this batch set.
+    pub(crate) first_batch: BinnedRenderPhaseBatch,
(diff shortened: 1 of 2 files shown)


crates/bevy_render/src/render_phase/mod.rs
```diff
@@ -189,9 +189,16 @@ pub enum BinnedRenderPhaseBatchSets<BK> {
     MultidrawIndirect(Vec<BinnedRenderPhaseBatchSet<BK>>),
 }
 
+/// A group of entities that will be batched together into a single multi-draw
+/// call.
 pub struct BinnedRenderPhaseBatchSet<BK> {
-    pub(crate) batches: Vec<BinnedRenderPhaseBatch>,
+    /// The first batch in this batch set.
+    pub(crate) first_batch: BinnedRenderPhaseBatch,
```
tychedelia, line 196: nit: `representative_batch`? `base_batch`? "first" seems somewhat incidental to how the batch set is built vs what it's used for.

tychedelia (approved):
Lgtm! Pretty notable win for such a small change.

IceSentry (approved):

crates/bevy_render/src/render_phase/mod.rs
pcwalton: Well, it does need to be the first one because offsets are calculated relative to it.

superdump (approved):

PR #17685, opened 2025-02-04
Split `Component::register_component_hooks` into individual methods

### LINKED ISSUE #17411: Refactor `Component::init_component_hooks` to split apart component hook methods
`Component`'s `init_component_hooks` method is intimidating and a bit frustrating to use. The obvious solution is to simply make each type of component hook its own method.

As @james-o-brien [points out](https://discord.com/channels/691052431525675048/1329275973501190216/1329632664616960091), the most naive solution won't work, because we want to cache whether or not each component has each kind of hook (stored in the flags on archetypes). As far as we know, there's no way to statically check if the function provided is actually a no-op.

This came up in the context of relationships.

> This feels like a weird place for this method. I'd kinda prefer making `on_insert` etc methods for `Component` (and removing the awkward `register_component_hooks` API), and then generating those methods in the `Component` trait with the `Relationship` derive macro.

_Originally posted by @alice-i-cecile in https://github.com/bevyengine/bevy/pull/17398#discussion_r1919445892_

bushrat011899:
# Objective

- Fixes #17411

## Solution

- Deprecated `Component::register_component_hooks`
- Added individual methods for each hook which return `None` if the hook is unused.

## Testing

- CI

---

## Migration Guide

`Component::register_component_hooks` is now deprecated and will be removed in a future release. When implementing `Component` manually, also implement the respective hook methods on `Component`.

```rust
// Before
impl Component for Foo {
    // snip
    fn register_component_hooks(hooks: &mut ComponentHooks) {
            hooks.on_add(foo_on_add);
    }
}

// After
impl Component for Foo {
    // snip
    fn on_add() -> Option<ComponentHook> {
            Some(foo_on_add)
    }
}
```

## Notes

I've chosen to deprecate `Component::register_component_hooks` rather than outright remove it to ease the migration guide. While it is in a state of deprecation, it must be used by `Components::register_component_internal` to ensure users who haven't migrated to the new hook definition scheme aren't left behind. For users of the new scheme, a default implementation of `Component::register_component_hooks` is provided which forwards the new individual hook implementations.

Personally, I think this is a cleaner API to work with, and would allow the documentation for hooks to exist on the respective `Component` methods (e.g., documentation for `OnAdd` can exist on `Component::on_add`). Ideally, `Component::on_add` would be the hook itself rather than a getter for the hook, but it is the only way to early-out for a no-op hook, which is important for performance.

## Migration Guide

`Component::register_component_hooks` has been deprecated. If you are manually implementing the `Component` trait and registering hooks there, use the individual methods such as `on_add` instead for increased clarity.

### DIFF
--- a/crates/bevy_ecs/macros/src/component.rs
+++ b/crates/bevy_ecs/macros/src/component.rs
@@ -71,15 +71,11 @@ pub fn derive_component(input: TokenStream) -> TokenStream {
 
     let storage = storage_path(&bevy_ecs_path, attrs.storage);
 
-    let on_add = hook_register_function_call(quote! {on_add}, attrs.on_add);
-    let mut on_insert = hook_register_function_call(quote! {on_insert}, attrs.on_insert);
-    let mut on_replace = hook_register_function_call(quote! {on_replace}, attrs.on_replace);
-    let on_remove: Option<TokenStream2> =
-        hook_register_function_call(quote! {on_remove}, attrs.on_remove);
-    let mut on_despawn = hook_register_function_call(quote! {on_despawn}, attrs.on_despawn);
+    let on_add_path = attrs.on_add.map(|path| path.to_token_stream());
+    let on_remove_path = attrs.on_remove.map(|path| path.to_token_stream());
 
-    if relationship.is_some() {
-        if on_insert.is_some() {
+    let on_insert_path = if relationship.is_some() {
+        if attrs.on_insert.is_some() {
             return syn::Error::new(
                 ast.span(),
                 "Custom on_insert hooks are not supported as relationships already define an on_insert hook",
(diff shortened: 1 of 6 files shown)


james-j-obrien (approved):

james-j-obrien:
Needs minor doc changes to pass CI, but I like this API and think the changes generally look good.

crates/bevy_ecs/macros/src/component.rs
```diff
@@ -71,15 +71,11 @@ pub fn derive_component(input: TokenStream) -> TokenStream {
...
-    let on_add = hook_register_function_call(quote! {on_add}, attrs.on_add);
-    let mut on_insert = hook_register_function_call(quote! {on_insert}, attrs.on_insert);
-    let mut on_replace = hook_register_function_call(quote! {on_replace}, attrs.on_replace);
-    let on_remove: Option<TokenStream2> =
-        hook_register_function_call(quote! {on_remove}, attrs.on_remove);
-    let mut on_despawn = hook_register_function_call(quote! {on_despawn}, attrs.on_despawn);
-
-    if relationship.is_some() {
-        if on_insert.is_some() {
+    let on_add_path = attrs.on_add.map(|path| path.to_token_stream());
```
ElliottjPierce, line 74: IDK, but could you do Path::to_token_stream here instead of passing a closure? I'm no expert here, but if that's possible, it might make it a bit more readable. There's a few other places here too that could use that if it's possible.

ElliottjPierce (approved):
Looks good to me! Definitely need another opinion on the proc macro code. I'm no expert there. 

Only downside here is that it may add more complexity to add more kinds of hooks, but I don't see that happening very often, and I think it's worth it for the readability. IDK why it wasn't designed like this in the first place.

PR #17688, opened 2025-02-05
Don't mark a previous mesh transform as changed if it didn't actually change.

pcwalton:
This patch fixes a bug whereby we're re-extracting every mesh every frame. It's a regression from PR #17413. The code in question has actually been in the tree with this bug for quite a while; it's that just the code didn't actually run unless the renderer considered the previous view transforms necessary. Occlusion culling expanded the set of circumstances under which Bevy computes the previous view transforms, causing this bug to appear more often.

This patch fixes the issue by checking to see if the previous transform of a mesh actually differs from the current transform before copying the current transform to the previous transform.

### DIFF
--- a/crates/bevy_pbr/src/prepass/mod.rs
+++ b/crates/bevy_pbr/src/prepass/mod.rs
@@ -226,7 +226,7 @@ pub fn update_previous_view_data(
     }
 }
 
-#[derive(Component, Default)]
+#[derive(Component, PartialEq, Default)]
 pub struct PreviousGlobalTransform(pub Affine3A);
 
 #[cfg(not(feature = "meshlet"))]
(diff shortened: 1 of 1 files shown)


crates/bevy_pbr/src/prepass/mod.rs
```diff
@@ -237,15 +237,19 @@ type PreviousMeshFilter = Or<(With<Mesh3d>, With<MeshletMesh3d>)>;
...
-        for (entity, transform) in &meshes {
-            commands
-                .entity(entity)
-                .try_insert(PreviousGlobalTransform(transform.affine()));
+        for (entity, transform, old_previous_transform) in &meshes {
+            let new_previous_transform = PreviousGlobalTransform(transform.affine());
+            // Make sure not to trigger change detection on
+            // `PreviousGlobalTransform` if the previous transform hasn't
+            // changed.
+            if old_previous_transform != Some(&new_previous_transform) {
```
IceSentry, line 250: IIRC there's a way to skip change detection while letting you still mutate the component directly instead of reinserting. I don't know if it would have a significant perf impact on this though.

IceSentry (approved):

crates/bevy_pbr/src/prepass/mod.rs
pcwalton: We would still need to potentially insert because the mesh might not have a `PreviousGlobalTransform` to begin with.

tychedelia (approved):

bushrat011899:
Wasm CI issues should be resolved once you upstream from `main`.

superdump (approved):

PR #17690, opened 2025-02-05
Add checked methods to add vertex attributes to mesh only if values is not empty

hukasu:
# Objective

When creating a custom mesh, there might be an attribute that is optional, that way you would need to store the `Mesh` in a variable, insert the required attributes, then check if the optional attribute exists or not, and insert if it does. With this checked methods you can use with the builder patterns

## Solution

Create methods `insert_attribute_if_not_empty` and `with_inserted_attribute_if_not_empty` (names are tentative).

## Testing

Used in a modification to `SphereMeshBuilder` (PR #17691)

## Showcase

Before
```rust
        let mut mesh = Mesh::new(
            PrimitiveTopology::TriangleList,
            RenderAssetUsages::default(),
        )
        .with_inserted_indices(indices)
        .with_inserted_attribute(Mesh::ATTRIBUTE_POSITION, points)
        .with_inserted_attribute(Mesh::ATTRIBUTE_NORMAL, normals)
        .with_inserted_attribute(Mesh::ATTRIBUTE_UV_0, uvs);

         if optional_attribute {
                 let optional_attribute = generate_optional_attribute();
                 mesh.insert_attribute(OPTIONAL_ATTRIBUTE, optional_attribute);
        }

       mesh
```

After
```rust
let optional_attribute = if optional_attribute {
          generate_optional_attribute()
} else {
    vec![]
};

Mesh::new(
    PrimitiveTopology::TriangleList,
    RenderAssetUsages::default(),
)
.with_inserted_indices(indices)
.with_inserted_attribute(Mesh::ATTRIBUTE_POSITION, points)
.with_inserted_attribute(Mesh::ATTRIBUTE_NORMAL, normals)
.with_inserted_attribute(Mesh::ATTRIBUTE_UV_0, uvs)
.with_inserted_attribute_if_not_empty(OPTIONAL_ATTRIBUTE, optional_attribue)
```

### DIFF
--- a/crates/bevy_mesh/src/mesh.rs
+++ b/crates/bevy_mesh/src/mesh.rs
@@ -237,6 +237,35 @@ impl Mesh {
             .insert(attribute.id, MeshAttributeData { attribute, values });
     }
 
+    /// Sets the data for a vertex attribute (position, normal, etc.) if not empty.
+    /// The name will often be one of the associated constants such
+    /// as [`Mesh::ATTRIBUTE_POSITION`].
+    ///
+    /// `Aabb` of entities with modified mesh are not updated automatically.
+    ///
+    /// # Panics
+    /// Panics when the format of the values does not match the attribute's format.
+    #[inline]
+    pub fn insert_attribute_if_not_empty(
+        &mut self,
+        attribute: MeshVertexAttribute,
+        values: impl Into<VertexAttributeValues>,
+    ) {
+        let values = values.into();
+        if !values.is_empty() {
+            let values_format = VertexFormat::from(&values);
+            if values_format != attribute.format {
+                panic!(
+                "Failed to insert attribute. Invalid attribute format for {}. Given format is {values_format:?} but expected {:?}",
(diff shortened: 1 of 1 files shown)


janhohenheim (changes requested):
I like the functionality, but shouldn't this rather be the default? Seems more like what the user would want in general, IMO.

hukasu:
have `with_inserted_attribute` already taking in `Option`?

janhohenheim:
No, I mean making this exact implementation the implementation of `with_inserted_attribute` and `insert_attribute`.

ChristopherBiscardi:
The PR this was used in was closed, and making this the default behavior also has implications for the shader code that would be handling the meshes. This is generally fine in some cases but could be surprising in others.

The user also has the power to do this and more complex checks.

Since this has been open for over a year without approvals I'm going to close, taking that lack of advocacy and response as this not being particularly desired at the moment, but it could be conceivably revived and advocated for.

PR #17691, opened 2025-02-05
Primitive 3d custom tangents

hukasu:
# Objective

Anisotropy requires tangents, and the current method of building tangents depends on UVs, which causes seams on most of the primitive shapes. But most of the primitive shapes have simple methods of calculating tangents that do not require the UVs.

## Solution

Create custom calculation of the tangent for some of the 3d primitives.

## Testing

Add custom primitives to the `anisotropy` example.

## Showcase

Sphere with custom tangent calculation in the `anisotropy` example
[Screencast_20250205_110935.webm](https://github.com/user-attachments/assets/9cc8daf9-7201-4a40-96e3-a4f7663d32a3)

## Migration Guide

This PR requires #17690

### DIFF
--- a/crates/bevy_mesh/src/mesh.rs
+++ b/crates/bevy_mesh/src/mesh.rs
@@ -237,6 +237,35 @@ impl Mesh {
             .insert(attribute.id, MeshAttributeData { attribute, values });
     }
 
+    /// Sets the data for a vertex attribute (position, normal, etc.) if not empty.
+    /// The name will often be one of the associated constants such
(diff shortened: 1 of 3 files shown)


hukasu:
The calculation that i'm doing is incredibly naive, it is just doing the cross product between the normal and an axis, but doing the cross product between 2 colinear vectors results in zero, what would be the best solution for this?

Sphere
Screenshot_20250205_202947

Capsule
Screenshot_20250205_202721

PR #17695, opened 2025-02-05
Weak handle migration

SludgePhD:
# Objective

- Make use of the new `weak_handle!` macro added in https://github.com/bevyengine/bevy/pull/17384

## Solution

- Migrate bevy from `Handle::weak_from_u128` to the new `weak_handle!` macro that takes a random UUID
- Deprecate `Handle::weak_from_u128`, since there are no remaining use cases that can't also be addressed by constructing the type manually

## Testing

- `cargo run -p ci -- test`

---

## Migration Guide

Replace `Handle::weak_from_u128` with `weak_handle!` and a random UUID.


alice-i-cecile (approved):
Nice! Do we need to note that the UUIDs associated with these handles has changed in the migration guide?

mockersf (approved):

PR #17567, opened 2025-01-27
Cold Specialization

tychedelia:
# Cold Specialization

## Objective

An ongoing part of our quest to retain everything in the render world, cold-specialization aims to cache pipeline specialization so that pipeline IDs can be recomputed only when necessary, rather than every frame. This approach reduces redundant work in stable scenes, while still accommodating scenarios in which materials, views, or visibility might change, as well as unlocking future optimization work like retaining render bins.

## Solution

Queue systems are split into a specialization system and queue system, the former of which only runs when necessary to compute a new pipeline id. Pipelines are invalidated using a combination of change detection and ECS ticks.

### The difficulty with change detection

Detecting “what changed” can be tricky because pipeline specialization depends not only on the entity’s components (e.g., mesh, material, etc.) but also on which view (camera) it is rendering in. In other words, the cache key for a given pipeline id is a view entity/render entity pair. As such, it's not sufficient simply to react to change detection in order to specialize -- an entity could currently be out of view or could be rendered in the future in camera that is currently disabled or hasn't spawned yet.

### Why ticks?

Ticks allow us to ensure correctness by allowing us to compare the last time a view or entity was updated compared to the cached pipeline id. This ensures that even if an entity was out of view or has never been seen in a given camera before we can still correctly determine whether it needs to be re-specialized or not.

## Testing

TODO: Tested a bunch of different examples, need to test more.

## Migration Guide

TODO

- `AssetEvents` has been moved into the `PostUpdate` schedule.

Issue #17535, opened 2025-01-25
GLTF loading and animation performance regression


aloucks:
## Bevy version

main
deb135c25c3a7eafdc432eaefe0957d6ffdcca62

## Relevant system information
```
$ cargo --version -vvv
cargo 1.83.0 (5ffbef321 2024-10-29)
release: 1.83.0
commit-hash: 5ffbef3211a8c378857905775a15c5b32a174d3b
commit-date: 2024-10-29
host: x86_64-pc-windows-msvc
libgit2: 1.8.1 (sys:0.19.0 vendored)
libcurl: 8.9.0-DEV (sys:0.4.74+curl-8.9.0 vendored ssl:Schannel)
os: Windows 10.0.22631 (Windows 11 Professional) [64-bit]
```
## What you did

```
cargo run --example scene_viewer --features animation assets/models/UndeadMale/scourgemale.gltf
```

## What went wrong

The gltf file never finishes loading when running with the dev profile (or at least it takes a very very long time - I've killed the process after a few minutes). It will eventually load when running it with `--release` but it's slow and the animation playback is very choppy.

## Additional information

The file loads nearly instantly with `v0.15.1` and animation playback is smooth, even with the dev profile. 

[UndeadMale.zip](https://github.com/user-attachments/files/18546742/UndeadMale.zip)

rparrett:
Are you able to bisect?

aloucks:
> Are you able to bisect?

The first bad commit is 21f1e3045c80230bc8c0055ba70dbf2e98d934a3

cart:
I was able to repo. This is the result of propagating over a million transforms per frame, when prior to #17398 we were propagating 28,000 per frame. We have the same number of transforms on both branches. This is actually caused by Children containing duplicates. I suspect that this is caused by buggy "scene spawn" code that spawns _both_ a fully populated Children collection _and_ a ChildOf, resulting in duplicates. This isn't possible via "normal" Rust flows, as you can't spawn a populated Children collection.

cart:
Looking at the source, `Children` is stored in the `Scene`'s `World`, and the scene spawn logic duplicates that into the main World using FromReflect (which means the full Children are written on insert, bypassing the normal flow and resulting in duplicates).

pcwalton:
If, like me, you're blocked because of this bug, the following diff will work around it:

```diff
diff --git a/crates/bevy_scene/src/scene.rs b/crates/bevy_scene/src/scene.rs
[...]

PR #17698, opened 2025-02-05
Retain bins from frame to frame.

pcwalton:
This PR makes Bevy keep entities in bins from frame to frame if they haven't changed. This reduces the time spent in `queue_material_meshes` and related functions to near zero for static geometry. This patch uses the same change tick technique that #17567 uses to detect when meshes have changed in such a way as to require re-binning.

In order to quickly find the relevant bin for an entity when that entity has changed, we introduce a new type of cache, the *bin key cache*. This cache stores a mapping from main world entity ID to cached bin key, as well as the tick of the most recent change to the entity. As we iterate through the visible entities in `queue_material_meshes`, we check the cache to see whether the entity needs to be re-binned. If it doesn't, then we mark it as clean in the `valid_cached_entity_bin_keys` bit set. If it does, then we insert it into the correct bin, and then mark the entity as clean. At the end, all entities not marked as clean are removed from the bins.

This patch has a dramatic effect on the rendering performance of most benchmarks, as it effectively eliminates `queue_material_meshes` from the profile. Note, however, that it generally simultaneously regresses `batch_and_prepare_binned_render_phase` by a bit (not by enough to outweigh the win, however). I believe that's because, before this patch, `queue_material_meshes` put the bins in the CPU cache for `batch_and_prepare_binned_render_phase` to use, while with this patch, `batch_and_prepare_binned_render_phase` must load the bins into the CPU cache itself.

On Caldera, this reduces the time spent in `queue_material_meshes` from 5+ ms to 0.2ms-0.3ms. Note that benchmarking on that scene is very noisy right now because of https://github.com/bevyengine/bevy/issues/17535.

Screenshot 2025-02-05 153458

### DIFF
--- a/crates/bevy_render/src/batching/gpu_preprocessing.rs
+++ b/crates/bevy_render/src/batching/gpu_preprocessing.rs
@@ -4,6 +4,7 @@ use core::any::TypeId;
 
 use bevy_app::{App, Plugin};
 use bevy_ecs::{
+    prelude::Entity,
     query::{Has, With},
     resource::Resource,
     schedule::IntoSystemConfigs as _,
@@ -1326,8 +1327,9 @@ pub fn batch_and_prepare_binned_render_phase<BPI, GFBD>(
                 let first_output_index = data_buffer.len() as u32;
                 let mut batch: Option<BinnedRenderPhaseBatch> = None;
 
-                for &(entity, main_entity) in &bin.entities {
-                    let Some(input_index) = GFBD::get_binned_index(&system_param_item, main_entity)
+                for main_entity in bin.entities() {
+                    let Some(input_index) =
+                        GFBD::get_binned_index(&system_param_item, *main_entity)
                     else {
                         continue;
                     };
@@ -1378,7 +1380,7 @@ pub fn batch_and_prepare_binned_render_phase<BPI, GFBD>(
                                 },
                             );
                             batch = Some(BinnedRenderPhaseBatch {
-                                representative_entity: (entity, main_entity),
+                                representative_entity: (Entity::PLACEHOLDER, *main_entity),
                                 instance_range: output_index..output_index + 1,
                                 extra_index: PhaseItemExtraIndex::maybe_indirect_parameters_index(
                                     NonMaxU32::new(indirect_parameters_index),
@@ -1424,8 +1426,8 @@ pub fn batch_and_prepare_binned_render_phase<BPI, GFBD>(
             let first_output_index = data_buffer.len() as u32;
 
             let mut batch: Option<BinnedRenderPhaseBatch> = None;
-            for &(entity, main_entity) in &phase.batchable_mesh_values[key].entities {
-                let Some(input_index) = GFBD::get_binned_index(&system_param_item, main_entity)
+            for main_entity in phase.batchable_mesh_values[key].entities() {
+                let Some(input_index) = GFBD::get_binned_index(&system_param_item, *main_entity)
                 else {
                     continue;
                 };
@@ -1487,7 +1489,7 @@ pub fn batch_and_prepare_binned_render_phase<BPI, GFBD>(
                             },
                         );
                         batch = Some(BinnedRenderPhaseBatch {
-                            representative_entity: (entity, main_entity),
+                            representative_entity: (Entity::PLACEHOLDER, *main_entity),
                             instance_range: output_index..output_index + 1,
                             extra_index: PhaseItemExtraIndex::IndirectParametersIndex {
                                 range: indirect_parameters_index..(indirect_parameters_index + 1),
@@ -1507,7 +1509,7 @@ pub fn batch_and_prepare_binned_render_phase<BPI, GFBD>(
                             },
                         );
                         batch = Some(BinnedRenderPhaseBatch {
-                            representative_entity: (entity, main_entity),
+                            representative_entity: (Entity::PLACEHOLDER, *main_entity),
                             instance_range: output_index..output_index + 1,
                             extra_index: PhaseItemExtraIndex::None,
                         });
@@ -1559,8 +1561,8 @@ pub fn batch_and_prepare_binned_render_phase<BPI, GFBD>(
                 )
             };
 
-            for &(_, main_entity) in &unbatchables.entities {
-                let Some(input_index) = GFBD::get_binned_index(&system_param_item, main_entity)
+            for main_entity in unbatchables.entities.keys() {
+                let Some(input_index) = GFBD::get_binned_index(&system_param_item, *main_entity)
                 else {
                     continue;
                 };
--- a/crates/bevy_render/src/render_phase/mod.rs
+++ b/crates/bevy_render/src/render_phase/mod.rs
@@ -239,18 +239,6 @@ where
     }
 }
 
-impl<BPI> PartialEq for CachedBinKey<BPI>
-where
-    BPI: BinnedPhaseItem,
-{
-    fn eq(&self, other: &Self) -> bool {
-        self.batch_set_key == other.batch_set_key
-            && self.bin_key == other.bin_key
-            && self.phase_type == other.phase_type
-            && self.change_tick == other.change_tick
-    }
-}
-
 /// How we store and render the batch sets.
 ///
 /// Each one of these corresponds to a [`GpuPreprocessingMode`].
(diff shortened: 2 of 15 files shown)


crates/bevy_render/src/render_phase/mod.rs
```diff
@@ -1357,3 +1528,127 @@ impl BinnedRenderPhaseType {
...
+impl<'a> Iterator for ReverseFixedBitSetZeroesIterator<'a> {
+    type Item = usize;
+
+    fn next(&mut self) -> Option<usize> {
+        while self.bit_index >= 0 {
+            // Unpack the bit index into block and bit.
+            let block_index = self.bit_index / (Block::BITS as isize);
+            let bit_pos = self.bit_index % (Block::BITS as isize);
+
+            // Grab the block. Mask off all bits before the one we're scanning
```
tychedelia, line 1594: "Before" here caught me up because we're going in reverse. Above? Previous?

examples/shader/custom_phase_item.rs
tychedelia, line 1: Hmm, out of scope for this PR, but makes me wonder whether we don't want to also demonstrate a "custom" specialization system here.

tychedelia (approved):
Looks great!

examples/shader/custom_phase_item.rs
pcwalton: There's a `specialized_mesh_pipeline` example, is that what you mean?

crates/bevy_render/src/render_phase/mod.rs
pcwalton: Changed to "above".

examples/shader/custom_phase_item.rs
tychedelia: No just that we're eagerly specializing and rebuilding the bins. I still like that this example is relatively minimal (and it's not as if it's a problem necessary), was just unsure whether we should demonstrate the cached version.

pcwalton: Perhaps at some point in a follow-up. I'm not super happy with `specialized_mesh_pipeline` and `custom_shader_instancing` and we might want to think about what a more comprehensive example would look like.

ickshonpe (commented):
With the event registered, everything seems to work.

I removed the `update_scroll_position` from the `testbed_ui` example (in `examples/`testbed/ui.rs`) and added an observer to the scrollable node:

```rust
// Scrolling list
parent
    .spawn((
        Node {
            align_self: AlignSelf::Stretch,
            height: Val::Percent(50.),
            overflow: Overflow::scroll_y(),
            ..default()
        },
        BackgroundColor(Color::srgb(0.10, 0.10, 0.10)),
    ))
    .observe(|
        trigger: Trigger<Pointer<Scroll>>,
        mut scroll_position_query: Query<&mut ScrollPosition>,
    | {
        if let Ok(mut scroll_position) = scroll_position_query.get_mut(trigger.target) {
            let (dx, dy) = match trigger.unit {
                MouseScrollUnit::Line => (trigger.x * 20., trigger.y * 20.),
                MouseScrollUnit::Pixel => (trigger.x, trigger.y),
            };
    
            scroll_position.offset_x -= dx;
            scroll_position.offset_y -= dy;
        }
    })
    .with_children(|parent| {
        // List items
        for i in 0..25 {
            parent
                .spawn((
                    Text(format!("Item {i}")),
                    TextFont {
                        ..default()
                    },
                    Label,
                    AccessibilityNode(Accessible::new(Role::ListItem)),
                ))
                .insert(Pickable {
                    should_block_lower: false,
                    ..default()
                });
        }
    });
```

which worked with problems.

pcwalton:
I pushed a fix for `deferred_rendering`. The problem was that we were only accounting for entities being added or removed from a phase, not entities changing bins within the same phase.

pcwalton:
All the pixel eagle failures look like false positives or pre-existing. I think this is good.

pcwalton:
Two of the failures in the Bevy example runner are fixed by #17717 and #17736.

crates/bevy_render/src/batching/gpu_preprocessing.rs
```diff
@@ -1378,7 +1380,7 @@ pub fn batch_and_prepare_binned_render_phase<BPI, GFBD>(
                                 },
                             );
                             batch = Some(BinnedRenderPhaseBatch {
-                                representative_entity: (entity, main_entity),
+                                representative_entity: (Entity::PLACEHOLDER, *main_entity),
```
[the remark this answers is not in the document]
superdump: Yup. And that is definitely a separate PR. :)

crates/bevy_render/src/render_phase/mod.rs
```diff
@@ -1357,3 +1617,127 @@ impl BinnedRenderPhaseType {
...
+        while self.bit_index >= 0 {
+            // Unpack the bit index into block and bit.
+            let block_index = self.bit_index / (Block::BITS as isize);
+            let bit_pos = self.bit_index % (Block::BITS as isize);
+
+            // Grab the block. Mask off all bits above the one we're scanning
+            // from by setting them all to 1.
+            let mut block = self.bitset.as_slice()[block_index as usize];
+            if bit_pos + 1 < (Block::BITS as isize) {
+                block |= (!0) << (bit_pos + 1);
```
superdump, line 1687: Why !0 instead of just 1? Seems unnecessarily indirect?

crates/bevy_render/src/render_phase/mod.rs
```diff
@@ -1357,3 +1617,127 @@ impl BinnedRenderPhaseType {
...
+            }
+
+            // Search for the next unset bit. Note that the `leading_ones`
+            // function counts from the MSB to the LSB, so we need to flip it to
+            // get the bit number.
+            let pos = (Block::BITS as isize) - (block.leading_ones() as isize) - 1;
```
superdump, line 1693: I don’t understand why leading_ones() is used here. Doesn’t the algorithm scan from the end of the set backwards to find zeros? Shouldn’t it look at trailing ones?

crates/bevy_sprite/src/lib.rs
```diff
@@ -151,6 +152,8 @@ impl Plugin for SpritePlugin {
                             .ambiguous_with(queue_material2d_meshes::<ColorMaterial>),
                         prepare_sprite_image_bind_groups.in_set(RenderSet::PrepareBindGroups),
                         prepare_sprite_view_bind_groups.in_set(RenderSet::PrepareBindGroups),
+                        sort_binned_render_phase::<Opaque2d>.in_set(RenderSet::PhaseSort),
+                        sort_binned_render_phase::<AlphaMask2d>.in_set(RenderSet::PhaseSort),
```
superdump, line 156: Were these just accidentally missing before this PR?

superdump (commented):
Just a couple of questions. Otherwise LGTM. Nice!

crates/bevy_render/src/render_phase/mod.rs
pcwalton: `!` in Rust actually flips all the bits, like `~` in C.

crates/bevy_render/src/render_phase/mod.rs
pcwalton: Yeah, so `fixedbitset` actually fills each block from LSB to MSB, so bit 0 is `(1 << 0)`, bit 1 is `(1 << 1)`, etc. `leading_ones` searches from MSB to LSB. So we need to use `leading_ones` to iterate in reverse. (`trailing_ones` would search from LSB to MSB, which would be wrong in this case.)

crates/bevy_sprite/src/lib.rs
pcwalton: Yep, they were accidentally missing. And we need them now.

crates/bevy_render/src/render_phase/mod.rs
pcwalton: And we need to fill all the bits so that we can set all the bits "past the end" of the bit vector. I don't actually know what value `fixedbitset` fills them with, but I doubt we can rely on them all being 1s (if anything, they're more likely to be 0s). So I mask them off to be safe.

crates/bevy_render/src/render_phase/mod.rs
superdump: Ahhh, ok. I guess they did that for simpler indexing? Regardless, fine.

superdump (approved):

PR #17701, opened 2025-02-06
Generalized and atomic staging utils

ElliottjPierce:
# Objective

The goal here is to provide a general alternative to a `RwLock` for rarely changed collections. When a change is made, it stages it in temporary storage without actually modifying the data. Then, at user defined points, the changes are applied to the data, draining the staged changes.

This has come up a lot regarding components, ex: #17569, but the same treatment has been discussed for reflection and other areas where a type has to be registered once and only once, on demand. This allows that functionality to be put on separate threads, which could unblock a lot of additional ideas, like assets as entities and read-only queries.

Further, it may be useful to put these "StageOnWrite" data structures in an Arc, so that has been implemented too. This could unlock benefits like not having to translate component ids when spawning scenes or cloning entities, but more on that in future PRs and discussions.

There's a lot of things this PR benefits but it should be evaluated as a stand-alone addition for these utilities. Still, those goals and use cases did influence the design, so keep that in mind.

## Solution

- Implement a generic `StagedChanges` trait that tracks changes to a "target" data structure, `Cold`.
- Implement types that track these staged changes, `StageOnWrite` and `AtomicStageOnWrite`.
- Implement both locking and unlocked ways of accessing the staged and un-staged data together, mutably and immutably.

## Testing

I did create a test module with an example, but the potential for bugs here is mostly in systems using this feature, so I kept it kind of minimal for now.


bushrat011899:
The `no_std` CI failure can be resolved by feature-gating your new types behind `feature = "alloc"`

alice-i-cecile (commented):
Can we put this in bevy_platform_support? I think it's likely more fitting there, and I'm nervous about adding anything new to bevy_utils, since we're trying to eliminate it.

ElliottjPierce:
> Can we put this in bevy_platform_support? I think it's likely more fitting there, and I'm nervous about adding anything new to bevy_utils, since we're trying to eliminate it.

Ah. I wasn't aware of that. I can absolutely move it.

bushrat011899:
I'd personally prefer if it stayed in `bevy_utils` until it could get yeeted into it's own micro crate. The goal for `bevy_platform_support` is to just get `std`-like functionality for all platforms and I'd like to keep its scope narrowed to that, otherwise I see it just being more `bevy_utils`. But I'm willing to be overruled on that front if we consider this something that _should_ be in the standard library.

crates/bevy_utils/src/staging.rs
```diff
@@ -0,0 +1,460 @@
+//! Provides an abstracted system for staging modifications to data structures that rarely change. See [`StageOnWrite`] as a starting point.
+
+use core::ops::Deref;
+
+use alloc::sync::Arc;
+use bevy_platform_support::sync::{RwLock, RwLockReadGuard, RwLockWriteGuard};
```
bushrat011899:
```suggestion
use bevy_platform_support::sync::{RwLock, RwLockReadGuard, RwLockWriteGuard};
use core::ops::Deref;

#[cfg(feature = "alloc")]
use bevy_platform_support::sync::Arc;
```

crates/bevy_utils/src/staging.rs
```diff
@@ -0,0 +1,460 @@
...
+    Cold(C),
+    /// There is staging necessary.
+    Staged(S),
+}
+
+/// A struct that allows read-optimized operations while still allowing mutation. When mutattions are made,
+/// they are staged. Then, at user-defined times, they are applied to the read-optimized storage. This allows muttations
+/// through [`RwLock`]s without needing to constantly lock old or cold data.
+///
+/// This is not designed for atomic use (ie. in an [`Arc`]). See [`AtomicStageOnWrite`] for that.
```
bushrat011899:
```suggestion
/// This is not designed for atomic use (ie. in an `Arc`).
#[cfg_attr(feature = "alloc", doc = "See [`AtomicStageOnWrite`] for that.")]
```

crates/bevy_utils/src/staging.rs
```diff
@@ -0,0 +1,460 @@
...
+#[derive(Default)]
+pub struct StageOnWrite<T: StagedChanges> {
+    /// Cold data is read optimized.
+    cold: T::Cold,
+    /// Staged data stores recent modifications to cold. It's [`RwLock`] coordinates mutations.
+    staged: RwLock<T>,
+}
+
+#[derive(Default)]
```
bushrat011899:
```suggestion
#[cfg(feature = "alloc")]
#[derive(Default)]
```

crates/bevy_utils/src/staging.rs
```diff
@@ -0,0 +1,460 @@
...
+    staged: RwLock<T>,
+}
+
+/// A version of [`StageOnWrite`] designed for atomic use. See [`StageOnWrite`] for details.
+///
+/// This type includes a baked in [`Arc`], so it can be shared similarly. Many of it's methods take `&mut self` even though
+/// it doesn't technically need the mutation. This is done to signify that the methods involve a state change of the data and to prevent deadlocks.
+/// Because everything that involves a write lock requires `&mut self`, it is impossible to deadlock because doing so would require another lock guard
+/// with the same lifetime, which rust will complaine about. If you do not want this behavior, see [`AtomicStageOnWriteInner`].
+#[derive(Clone)]
```
bushrat011899:
```suggestion
#[cfg(feature = "alloc")]
#[derive(Clone)]
```

crates/bevy_utils/src/staging.rs
```diff
@@ -0,0 +1,460 @@
...
+    /// # Deadlocks
+    ///
+    /// This deadlocks if there are any write lock guards on this thread for this value.
+    #[inline]
+    pub fn read_scope_locked<R>(&self, f: impl FnOnce(&StagedRef<T>) -> R) -> R {
+        f(&self.read_lock().as_staged_ref())
+    }
+}
+
+impl<T: StagedChanges> AtomicStageOnWrite<T> {
```
bushrat011899:
```suggestion
#[cfg(feature = "alloc")]
impl<T: StagedChanges> AtomicStageOnWrite<T> {
```

crates/bevy_utils/src/staging.rs
```diff
@@ -0,0 +1,460 @@
...
+impl<'a, T: StagedChanges> Clone for StagedRef<'a, T> {
+    fn clone(&self) -> Self {
+        Self {
+            staged: self.staged,
+            cold: self.cold,
+        }
+    }
+}
+
+impl<T: StagedChanges> Default for AtomicStageOnWrite<T>
```
bushrat011899:
```suggestion
#[cfg(feature = "alloc")]
impl<T: StagedChanges> Default for AtomicStageOnWrite<T>
```

crates/bevy_platform_support/src/lib.rs
```diff
@@ -21,6 +21,8 @@ pub mod time;
 
 #[cfg(feature = "alloc")]
 pub mod collections;
+#[cfg(feature = "alloc")]
```
bushrat011899, line 24:
```suggestion
```

crates/bevy_platform_support/src/staging.rs
```diff
@@ -0,0 +1,460 @@
...
+impl<'a, T: StagedChanges> Clone for StagedRef<'a, T> {
+    fn clone(&self) -> Self {
+        Self {
+            staged: self.staged,
+            cold: self.cold,
+        }
+    }
+}
+
+impl<T: StagedChanges> Default for AtomicStageOnWrite<T>
```
bushrat011899, line 405:
```suggestion
#[cfg(feature = "alloc")]
impl<T: StagedChanges> Default for AtomicStageOnWrite<T>
```

crates/bevy_platform_support/src/staging.rs
```diff
@@ -0,0 +1,460 @@
...
+    /// # Deadlocks
+    ///
+    /// This deadlocks if there are any write lock guards on this thread for this value.
+    #[inline]
+    pub fn read_scope_locked<R>(&self, f: impl FnOnce(&StagedRef<T>) -> R) -> R {
+        f(&self.read_lock().as_staged_ref())
+    }
+}
+
+impl<T: StagedChanges> AtomicStageOnWrite<T> {
```
bushrat011899, line 210:
```suggestion
#[cfg(feature = "alloc")]
impl<T: StagedChanges> AtomicStageOnWrite<T> {
```

crates/bevy_platform_support/src/staging.rs
```diff
@@ -0,0 +1,460 @@
...
+    staged: RwLock<T>,
+}
+
+/// A version of [`StageOnWrite`] designed for atomic use. See [`StageOnWrite`] for details.
+///
+/// This type includes a baked in [`Arc`], so it can be shared similarly. Many of it's methods take `&mut self` even though
+/// it doesn't technically need the mutation. This is done to signify that the methods involve a state change of the data and to prevent deadlocks.
+/// Because everything that involves a write lock requires `&mut self`, it is impossible to deadlock because doing so would require another lock guard
+/// with the same lifetime, which rust will complaine about. If you do not want this behavior, see [`AtomicStageOnWriteInner`].
+#[derive(Clone)]
```
bushrat011899, line 89:
```suggestion
#[cfg(feature = "alloc")]
#[derive(Clone)]
```

crates/bevy_platform_support/src/staging.rs
```diff
@@ -0,0 +1,460 @@
...
+/// This is not designed for atomic use (ie. in an [`Arc`]). See [`AtomicStageOnWrite`] for that.
+#[derive(Default)]
+pub struct StageOnWrite<T: StagedChanges> {
+    /// Cold data is read optimized.
+    cold: T::Cold,
+    /// Staged data stores recent modifications to cold. It's [`RwLock`] coordinates mutations.
+    staged: RwLock<T>,
+}
+
+#[derive(Default)]
```
bushrat011899, line 73:
```suggestion
#[cfg(feature = "alloc")]
#[derive(Default)]
```

crates/bevy_platform_support/src/staging.rs
```diff
@@ -0,0 +1,460 @@
...
+    Cold(C),
+    /// There is staging necessary.
+    Staged(S),
+}
+
+/// A struct that allows read-optimized operations while still allowing mutation. When mutattions are made,
+/// they are staged. Then, at user-defined times, they are applied to the read-optimized storage. This allows muttations
+/// through [`RwLock`]s without needing to constantly lock old or cold data.
+///
+/// This is not designed for atomic use (ie. in an [`Arc`]). See [`AtomicStageOnWrite`] for that.
```
bushrat011899, line 64:
```suggestion
/// This is not designed for atomic use (ie. in an `Arc`).
#[cfg_attr(feature = "alloc", doc = "See [`AtomicStageOnWrite`] for that.")]
```

crates/bevy_platform_support/src/staging.rs
```diff
@@ -0,0 +1,460 @@
+//! Provides an abstracted system for staging modifications to data structures that rarely change. See [`StageOnWrite`] as a starting point.
+
+use core::ops::Deref;
+
+use crate::alloc::sync::Arc;
+use crate::sync::{RwLock, RwLockReadGuard, RwLockWriteGuard};
```
bushrat011899, line 6:
```suggestion
use crate::sync::{RwLock, RwLockReadGuard, RwLockWriteGuard};
use core::ops::Deref;

#[cfg(feature = "alloc")]
use crate::sync::Arc;
```

bushrat011899 (changes requested):
This should fix the `alloc` issue you were having before and also allow using most of the functionality here even without `alloc` enabled.

ElliottjPierce:
> I'd personally prefer if it stayed in `bevy_utils` until it could get yeeted into it's own micro crate. The goal for `bevy_platform_support` is to just get `std`-like functionality for all platforms and I'd like to keep its scope narrowed to that, otherwise I see it just being more `bevy_utils`. But I'm willing to be overruled on that front if we consider this something that _should_ be in the standard library.

I think the idea for the utils crate was to hold micro utils that could arguably be part of std. Many of the utilities in there are in nightly rust's std already. I'll keep it in platform_support for now, but I'm happy to move it back if a different consensus is reached. I definitely respect your opinion @bushrat011899

crates/bevy_platform_support/src/staging.rs
```diff
@@ -0,0 +1,460 @@
...
+pub enum MaybeStaged<C, S> {
+    /// There is no staging necessary.
+    Cold(C),
+    /// There is staging necessary.
+    Staged(S),
+}
+
+/// A struct that allows read-optimized operations while still allowing mutation. When mutattions are made,
+/// they are staged. Then, at user-defined times, they are applied to the read-optimized storage. This allows muttations
+/// through [`RwLock`]s without needing to constantly lock old or cold data.
```
bushrat011899, line 63:
```suggestion
/// A struct that allows read-optimized operations while still allowing mutation.
/// When mutations are made, they are staged.
/// Then, at user-defined times, they are applied to the read-optimized storage.
/// This allows mutations through [`RwLock`]s without needing to constantly lock old or cold data.
```
Simple typo and reformatting for new lines. In Rust documentation, you can put new sentences in the same paragraph on the next documentation line. I personally find that easier to read, and it's usually more friendly for Git diffs.

crates/bevy_platform_support/src/staging.rs
```diff
@@ -0,0 +1,460 @@
...
+
+use crate::alloc::sync::Arc;
+use crate::sync::{RwLock, RwLockReadGuard, RwLockWriteGuard};
+
+/// Signifies that this type represents staged changes to [`Cold`](Self::Cold).
+pub trait StagedChanges: Default {
+    /// The more compact data structure that these changes compact into.
+    type Cold;
+
+    /// This applies these changes to the passed [`Cold`](Self::Cold). When this is done, there should be no more staged changes, and [`any_staged`](Self::any_staged) should be false.
```
bushrat011899, line 14:
```suggestion
    /// This applies these changes to the passed [`Cold`](Self::Cold).
    /// When this is done, there should be no more staged changes, and [`any_staged`](Self::any_staged) should be `false`.
```

crates/bevy_platform_support/src/staging.rs
```diff
@@ -0,0 +1,460 @@
...
+    fn apply_staged(&mut self, storage: &mut Self::Cold);
+
+    /// Returns true if and only if there are staged changes that could be applied.
```
bushrat011899, line 17:
```suggestion
    /// Returns `true` if and only if there are staged changes that could be applied.
```

crates/bevy_platform_support/src/staging.rs
```diff
@@ -0,0 +1,460 @@
...
+    fn any_staged(&self) -> bool;
+}
+
+/// A trait that signifies that it holds an immutable reference to a cold type (ie. [`StagedChanges::Cold`]).
+pub trait ColdStorage<T: StagedChanges>: Deref<Target = T::Cold> {}
+
+/// A struct that allows staging changes while reading from cold storage. Generally, staging changes should be implemented on this type.
```
bushrat011899, line 24:
```suggestion
/// A struct that allows staging changes while reading from cold storage.
/// Generally, staging changes should be implemented on this type.
```

crates/bevy_platform_support/src/staging.rs
```diff
@@ -0,0 +1,460 @@
...
+pub struct Stager<'a, T: StagedChanges> {
+    /// The storage that is read optimized.
+    pub cold: &'a T::Cold,
+    /// The staged changes.
+    pub staged: &'a mut T,
+}
+
+/// A struct that allows accessing changes while reading from cold storage. Generally, reading data should be implemented on this type.
```
bushrat011899, line 32:
```suggestion
/// A struct that allows accessing changes while reading from cold storage.
/// Generally, reading data should be implemented on this type.
```

crates/bevy_platform_support/src/staging.rs
```diff
@@ -0,0 +1,460 @@
...
+#[derive(Copy)]
+pub struct StagedRef<'a, T: StagedChanges> {
+    /// The storage that is read optimized.
+    pub cold: &'a T::Cold,
+    /// The staged changes.
+    pub staged: &'a T,
+}
+
+/// A locked version of [`Stager`]. Use this to hold a lock guard while using [`StagerLocked::as_stager`] or similar.
```
bushrat011899, line 41:
```suggestion
/// A locked version of [`Stager`].
/// Use this to hold a lock guard while using [`StagerLocked::as_stager`] or similar.
```

crates/bevy_platform_support/src/staging.rs
```diff
@@ -0,0 +1,460 @@
...
+pub struct StagerLocked<'a, T: StagedChanges, C: ColdStorage<T>> {
+    cold: C,
+    staged: RwLockWriteGuard<'a, T>,
+}
+
+/// A locked version of [`StagedRef`] Use this to hold a lock guard while using [`StagerLocked::as_staged_ref`].
```
bushrat011899, line 47:
```suggestion
/// A locked version of [`StagedRef`].
/// Use this to hold a lock guard while using [`StagerLocked::as_staged_ref`].
```

crates/bevy_platform_support/src/staging.rs
```diff
@@ -0,0 +1,460 @@
...
+    cold: T::Cold,
+    /// Staged data stores recent modifications to cold. It's [`RwLock`] coordinates mutations.
+    staged: RwLock<T>,
+}
+
+#[derive(Default)]
+struct AtomicStageOnWriteInner<T: StagedChanges> {
+    /// Cold data is read optimized. This lives behind a [`RwLock`], but it is only written to for applying changes in
+    /// a non-blocking way. In other worlds this locks, but almost never blocks. (It can technically block if a thread
+    /// tries to read from it while it is having changes applied, but that is extremely rare.)
```
bushrat011899, line 80:
```suggestion
    /// Cold data is read optimized.
    /// This lives behind a [`RwLock`], but it is only written to for applying changes in a non-blocking way.
    /// This will only block if a thread tries to read from it while it is having changes applied, but that is extremely rare.
```

crates/bevy_platform_support/src/staging.rs
```diff
@@ -0,0 +1,460 @@
...
+    cold: RwLock<T::Cold>,
+    /// Staged data stores recent modifications to cold. It's [`RwLock`] coordinates mutations.
+    staged: RwLock<T>,
+}
+
+/// A version of [`StageOnWrite`] designed for atomic use. See [`StageOnWrite`] for details.
```
bushrat011899, line 86:
```suggestion
/// A version of [`StageOnWrite`] designed for atomic use.
/// See [`StageOnWrite`] for details.
```

crates/bevy_platform_support/src/staging.rs
```diff
@@ -0,0 +1,460 @@
...
+///
+/// This type includes a baked in [`Arc`], so it can be shared similarly. Many of it's methods take `&mut self` even though
+/// it doesn't technically need the mutation. This is done to signify that the methods involve a state change of the data and to prevent deadlocks.
+/// Because everything that involves a write lock requires `&mut self`, it is impossible to deadlock because doing so would require another lock guard
+/// with the same lifetime, which rust will complaine about. If you do not want this behavior, see [`AtomicStageOnWriteInner`].
```
bushrat011899, line 91:
```suggestion
/// This type includes a baked in [`Arc`], so it can be shared across threads.
/// Many of it's methods take `&mut self` to ensure access is exclusive, preventing possible deadlocks.
/// If you do not want this behavior, see [`AtomicStageOnWriteInner`].
```

crates/bevy_platform_support/src/staging.rs
```diff
@@ -0,0 +1,460 @@
...
+impl<T: StagedChanges> StageOnWrite<T> {
+    /// Constructs a new [`StageOnWrite`] with the given value and no staged changes.
+    pub fn new(value: T::Cold) -> Self {
+        Self {
+            cold: value,
+            staged: RwLock::default(),
+        }
+    }
+
+    /// Gets the inner cold data if it is safe. If [`any_staged`](Self::any_staged) is known to be false, this can be safely unwrapped.
```
bushrat011899, line 105:
```suggestion
    /// Gets the inner cold data if it is safe.
    /// If [`any_staged`](Self::any_staged) is known to be false, this can be safely unwrapped.
```
What does _"if it is safe"_ refer to here? I generally avoid using the term "safe" in Rust documentation unless it is explicitly around `unsafe`/safety.

crates/bevy_platform_support/src/staging.rs
```diff
@@ -0,0 +1,460 @@
...
+impl<T: StagedChanges> AtomicStageOnWrite<T> {
+    /// Constructs a new [`AtomicStageOnWrite`] with the given value and no staged changes.
+    pub fn new(value: T::Cold) -> Self {
+        Self(Arc::new(AtomicStageOnWriteInner {
+            cold: RwLock::new(value),
+            staged: RwLock::default(),
+        }))
+    }
+
+    /// Gets the inner cold data if it is safe. If [`any_staged`](Self::any_staged) is known to be false, this can be safely unwrapped.
```
bushrat011899, line 224:
```suggestion
    /// Gets the inner cold data if it is safe.
    /// If [`any_staged`](Self::any_staged) is known to be false, this can be safely unwrapped.
```
See above regarding use of the term "safe".

alice-i-cecile:
I agree with bushrat here :) Thanks for making the case.

crates/bevy_platform_support/src/staging.rs
```diff
@@ -0,0 +1,460 @@
...
+    pub fn apply_staged_for_full_locked(&mut self) -> RwLockWriteGuard<'_, T::Cold> {
+        let mut staged = self.0.staged.write().unwrap();
+        let mut cold = self.0.cold.write().unwrap();
+        if staged.any_staged() {
+            staged.apply_staged(&mut cold);
+        }
+        cold
+    }
+
+    /// Gets the inner cold data if it is safe.
```
bushrat011899, line 250: See above regarding "safe".

crates/bevy_platform_support/src/staging.rs
```diff
@@ -0,0 +1,460 @@
...
+                if staged.any_staged() {
+                    None
+                } else {
+                    Some(cold)
+                }
+            }
+        }
+    }
+
+    /// If possible applies any staged changes. Returns true if it can guarantee there are no more staged changes.
```
bushrat011899, line 283:
```suggestion
    /// If possible applies any staged changes.
    /// Returns true if it can guarantee there are no more staged changes.
```

crates/bevy_platform_support/src/staging.rs
```diff
@@ -0,0 +1,460 @@
...
+        }
+    }
+
+    /// Easily run a stager function to stage changes.
+    #[inline]
+    pub fn stage_scope_locked<R>(&mut self, f: impl FnOnce(&mut Stager<T>) -> R) -> R {
+        f(&mut self.stage_lock().as_stager())
+    }
+
+    /// Easily run a stager function to stage changes. Then, tries to apply those changes if doing so wouldn't lock.
```
bushrat011899, line 355:
```suggestion
    /// Easily run a stager function to stage changes.
    /// Then, tries to apply those changes if doing so wouldn't lock.
```

crates/bevy_platform_support/src/staging.rs
```diff
@@ -0,0 +1,460 @@
...
+        }
+    }
+
+    // This is commented out, as it intentionally does not compile. This demonstrates how `AtomicStageOnWrite` prevents deadlock using the borrow checker.
+    // #[test]
+    // fn test_no_compile_for_deadlock() {
+    //     let mut stage_on_write = AtomicStageOnWrite::<StagedNumVec>::default();
+    //     let reading = stage_on_write.read_lock();
+    //     stage_on_write.apply_staged_non_blocking();
+    // }
```
bushrat011899, line 457: Instead, place this in the documentation for `AtomicStageOnWrite` as a `compile_fail`:

\```compile_fail
use ...
let mut stage_on_write = AtomicStageOnWrite::<StagedNumVec>::default();
let reading = stage_on_write.read_lock();
stage_on_write.apply_staged_non_blocking();
\```

crates/bevy_platform_support/src/staging.rs
```diff
@@ -0,0 +1,460 @@
...
+
+    #[test]
+    fn test_simple_stage() {
+        let mut data = StageOnWrite::<StagedNumVec>::default();
+        data.stage_scope_locked(|stager| stager.staged.added.push(5));
+        let full = data.apply_staged_for_full();
+        assert_eq!(full[0], 5);
```
bushrat011899, line 464:
```suggestion
        assert_eq!(full[..], &[5]);
```
Stronger test just to ensure the value isn't added twice, etc.

crates/bevy_platform_support/src/staging.rs
```diff
@@ -0,0 +1,466 @@
...
+        Self {
+            cold: value,
+            staged: RwLock::default(),
+        }
+    }
+
+    /// Gets the inner cold data if it is safe. If [`any_staged`](Self::any_staged) is known to be false, this can be safely unwrapped.
+    #[inline]
+    pub fn full(&mut self) -> Option<&mut T::Cold> {
+        if self.staged.get_mut().unwrap().any_staged() {
```
bushrat011899, line 108: Instead of `unwrap()`, it is more typical to use `.unwrap_or_else(PoisonError::into_inner)` when dealing with `RwLock`'s possibility to be poisoned.

NthTensor:
I'll defer to @alice-i-cecile as to judging whether this is properly motivated.
Some thoughts, having skimmed but not really reviewed:
+ It would be nice to see performance comparison between this and RwLock. I'm generally hesitant to merge ostensible optimizations without seeing numbers.
+ There's existing crates that do this. How does this differ from, for example, `left_right`?

ElliottjPierce:
> I'll defer to @alice-i-cecile as to judging whether this is properly motivated.

Agreed. This is definitely tricky to determine. The end goal is to enable `&Components` to be effectively passed around freely, with full rights to register components etc. This would have lots of exciting implications, some of which have been discussed elsewhere. This staging idea is the best architecture for that I could come up with, but if there is a better one, I'm all for it. 

>  It would be nice to see performance comparison between this and RwLock. I'm generally hesitant to merge ostensible optimizations without seeing numbers.

Totally agree here too. The simplicity of RwLock is appealing. I haven't bothered to test it yet because when this idea was first discussed on discord Cart was pretty sure we should avoid solutions where registering a new component would block old components from being read. `RwLock<Components>` would have that problem. Maybe that simplicity would still be worth it, but his points steered me away from that for now. 

> There's existing crates that do this. How does this differ from, for example, `left_right`?

I wish `left_right` would work for this too. It was one of the first things I thought about, but it's just barely not enough for what we need (I think). It only allows one writer at a time. We could just through the writer in an `Arc<RwLock<Writer>>`, but there are other problems. For example, what if a threads want to register the a component and then read its `ComponentInfo`? Well, the info will only exist on the writer, so the thread will have to use only the writer instead of a reader. This could be fixed by pushing the changes on each registration, buy know you're back to effectively a `RwLock`.

`left_right` *could* be incorporated into a direct solution, but it would need a lot of wrappers to make it safe, and the synchronizing in that might be even slower than `RwLock`. Besides all, it wouldn't work for `no_std`. I'm no expert on rust's third party synchronizing crates, but the ones I researched ahead of time, like `left_right`, don't *feel* promising to me. I could always be wrong.

That's my thoughts at least. Believe me, I wouldn't be making this PR if I saw an easier way, but I don't. But if anyone else wants to save the day here, I'm all ears.

crates/bevy_utils/src/staging.rs
```diff
@@ -0,0 +1,894 @@
...
+impl<T: StagedChanges> StagableWritesTypes for StageOnWrite<T> {
+    type Staging = T;
+
+    type ColdRef<'a>
+        = &'a T::Cold
+    where
+        T::Cold: 'a;
+}
+
+impl<T: StagedChanges> StagableWritesTypes for AtomicStageOnWrite<T> {
```
bushrat011899, line 403:
```suggestion
#[cfg(feature = "alloc")]
impl<T: StagedChanges> StagableWritesTypes for AtomicStageOnWrite<T> {
```

crates/bevy_utils/src/staging.rs
```diff
@@ -0,0 +1,894 @@
...
+    fn raw_read_cold(&self) -> Self::ColdRef<'_> {
+        &self.cold
+    }
+
+    fn raw_read_staged(&self) -> RwLockReadGuard<'_, Self::Staging> {
+        self.staged.read().unwrap_or_else(PoisonError::into_inner)
+    }
+}
+
+impl<T: StagedChanges> StagableWrites for AtomicStageOnWrite<T> {
```
bushrat011899, line 422:
```suggestion
#[cfg(feature = "alloc")]
impl<T: StagedChanges> StagableWrites for AtomicStageOnWrite<T> {
```

bushrat011899 (changes requested):
Minor fixes for `compile-check-no-std`

bushrat011899:
> I wish `left_right` would work for this too. It was one of the first things I thought about, but it's just barely not enough for what we need (I think). It only allows one writer at a time. We could just through the writer in an `Arc<RwLock<Writer>>`, but there are other problems. For example, what if a threads want to register the a component and then read its `ComponentInfo`? Well, the info will only exist on the writer, so the thread will have to use only the writer instead of a reader. This could be fixed by pushing the changes on each registration, buy know you're back to effectively a `RwLock`.
> 
> `left_right` _could_ be incorporated into a direct solution, but it would need a lot of wrappers to make it safe, and the synchronizing in that might be even slower than `RwLock`. Besides all, it wouldn't work for `no_std`. I'm no expert on rust's third party synchronizing crates, but the ones I researched ahead of time, like `left_right`, don't _feel_ promising to me. I could always be wrong.

Definitely worth documenting this thought process either in the PR description or even the documentation on the module itself. Knowing exactly what niche this is trying to fill helps justify its existence, and will inform future maintainers about _why_ certain tradeoffs have been made.

ElliottjPierce:
There's a few things I'm trying to clean up at the moment, and then I'll go hands off for a bit. Making the example showed me some areas to improve.

NthTensor:
Great, thanks for the thorough response. Seems like we should def proceed with this then.

ElliottjPierce:
@bushrat011899 I'm satisfied with where it's at at the moment. You had a lot of great suggestions in the last round. Sorry it took so long to do them. Looking forward to more!

crates/bevy_utils/src/staging.rs
```diff
@@ -0,0 +1,1108 @@
...
+//!     staged: RwLockReadGuard<'a, StagedPlayerChanges>,
+//!     // must be valid
+//!     id: PlayerId,
+//! }
+//!
+//! struct LockedNameColdRef<'a, T: StagableWritesCore<Staging = StagedPlayerChanges> + 'a> {
+//!     cold: T::ColdRef<'a>,
+//!     // must be valid
+//!     id: PlayerId,
+//! }
```
andriyDev, line 258: Can we link both of these as workarounds to stabilization of https://doc.rust-lang.org/std/sync/struct.MappedRwLockReadGuard.html ? Just to make it clear that these are not strictly necessary.

crates/bevy_utils/src/staging.rs
```diff
@@ -0,0 +1,1108 @@
...
+    /// Same as [`raw_read_staged`](StagableWritesCore::raw_read_staged), but never blocks.
+    fn raw_read_staged_non_blocking(&self) -> Option<RwLockReadGuard<'_, Self::Staging>>;
+
+    /// Allows raw access to reading staged changes, which may be missing context of cold storage.
+    fn raw_write_staged(&self) -> RwLockWriteGuard<'_, Self::Staging>;
+
+    /// Same as [`raw_write_staged`](StagableWritesCore::raw_write_staged), but never blocks.
+    fn raw_write_staged_non_blocking(&self) -> Option<RwLockWriteGuard<'_, Self::Staging>>;
+
+    /// Allows raw access to both staged and cold data.
```
andriyDev, line 397: `raw_write_cold_mut` and `raw_write_staged_mut` both mention that they don't lock. Should we add this to the comment here? e.g., `This should not do any locking, and shouldn't need to as we have exclusive references.`

crates/bevy_utils/src/staging.rs
```diff
@@ -0,0 +1,1108 @@
...
+        f: impl FnOnce(&StagedRef<<Self as StagableWritesTypes>::Staging>) -> R,
+    ) -> R {
+        f(&self.read_lock().as_staged_ref())
+    }
+}
+
+/// This trait provides some conviniencies around [`StagableWritesCore`].
+///
+/// For example, mutable references are used to enforce safety for some functions.
+pub trait StagableWrites {
```
andriyDev, line 524: Why does this need to be a separate trait? Can't we just `impl StagableWritesCore { /*methods go here*/ }`?

crates/bevy_utils/src/staging.rs
```diff
@@ -0,0 +1,1108 @@
...
+/// Then, at user-defined times, they are applied to the read-optimized storage.
+/// This allows mutations through [`RwLock`]s without needing to constantly lock old or cold data.
+///
+/// This is not designed for atomic use (ie. in an `Arc`).
+#[cfg_attr(feature = "alloc", doc = "See [`AtomicStageOnWrite`] for that.")]
+#[derive(Default, Debug)]
+pub struct StageOnWrite<T: StagedChanges> {
+    /// Cold data is read optimized.
+    cold: T::Cold,
+    /// Staged data stores recent modifications to cold. It's [`RwLock`] coordinates mutations.
```
andriyDev, line 618: "It's" doesn't make sense here - shorthand for "It is RwLock coordinates mutations". Honestly just removing it seems fine to me.

Repeat for `AtomicStageOnWrite`.

(weewoo grammar police)

crates/bevy_utils/src/staging.rs
```diff
@@ -0,0 +1,1108 @@
...
+    staged: RwLock<T>,
+}
+
+/// A version of [`StageOnWrite`] designed for atomic use.
+/// It functions fully without needing `&mut self`.
+/// See [`StageOnWrite`] for details.
```
andriyDev, line 624: I think this line is redundant. We already link to it on the first line.

Repeat for ArcStageOnWrite.

crates/bevy_utils/src/staging.rs
```diff
@@ -0,0 +1,1108 @@
...
+///
+/// This type includes a baked in [`Arc`], so it can be shared across threads.
+///
+/// Remember to use [`apply_staged_non_blocking`](AtomicStageOnWrite::apply_staged_non_blocking) or similar methods periodically as a best practice.
+#[cfg(feature = "alloc")]
+#[derive(Clone)]
+pub struct ArcStageOnWrite<T: StagedChanges>(pub Arc<AtomicStageOnWrite<T>>);
+
+/// Although it is is often enough to pass around references to a [`StagableWritesCore`], it is sometimes desierable to encapsulate that reference here.
```
andriyDev, line 646: `desierable` -> `desirable`

andriyDev (changes requested):
I haven't yet reviewed the implementation, but I need a break, so here's most of the review. I'll come back for another pass on the impl later.

crates/bevy_utils/src/staging.rs
ElliottjPierce: Thats a really good question.

`StagableWritesCore` gives information about the type coordinating the stage-on-write solution. This is the core of the actual implementation. `StagableWrites` protects against deadlocks by requiring mutable access to a reference to the `StagableWritesCore` type, not to the inner type itself. That reference could be a `ArcStageOnWrite` or a `RefStageOnWrite`. 

As an example, let's say we have a `ArcStageOnWrite` in thread A, and we share that reference to threads B and C by cloning the arc. Thread's B and C can prevent internal deadlocks by using `StagableWrites` on their arcs (provided they don't clone the arcs internally). If the two traits were unified, we would need to clone a mutable reference between threads (not allowed) or use the unsafe, deadlock-prone writing methods on `StagableWritesCore`.

The extra trait `StagableWrites` was the best way I could think of to balance deadlock prevention with thread sharing. If you have any other ideas, I'm all ears.

ElliottjPierce:
@andriyDev Thanks for the great docs review. I don't know how I managed some of those typos lol.

ElliottjPierce:
There are a few improvements I can make here after #18028 is merged. Just cleaning up some undocumented behavior.

crates/bevy_utils/src/staging.rs
```diff
@@ -0,0 +1,1124 @@
...
+    /// Easily run a stager function to stage changes.
+    /// Then, tries to apply those changes if doing so wouldn't lock.
+    ///
+    /// # Deadlocks
+    ///
+    /// This can still deadlock if this [`Arc`] has been cloned around on the same thread and is still being locked on.
+    /// But that is very unlikely.
+    #[inline]
+    pub fn stage_scope_locked_eager<R>(&mut self, f: impl FnOnce(&mut Stager<T>) -> R) -> R {
+        // Safety: Since this has mutible access to self, we can be reasonably sure this is safe.
```
andriyDev, line 937: `mutible` -> `mutable`

andriyDev (approved):

alice-i-cecile:
Closing in favor of #18173.

PR #17704, opened 2025-02-06
Add scroll functionality to bevy_picking

colepoirier:
# Objective

`bevy_picking` currently does not support scroll events.

## Solution

This pr adds a new event type for scroll, and updates the default input system for mouse pointers to read and emit this event.

## Testing

I haven't tested these changes, if the reviewers can advise me how to do so I'd appreciate it!


crates/bevy_picking/src/events.rs
```diff
@@ -750,6 +765,28 @@ pub fn pointer_events(
                     event_writers.move_events.send(move_event);
                 }
             }
+            PointerAction::Scroll { x, y, unit } => {
+                // If it's a press, emit a Pressed event and mark the hovered entities as pressed
```
aevyrie, line 769: this comment looks like a copy paste error

ickshonpe (commented):
You need to add 
```
 .add_event::<Pointer<Scroll>>()
```
to `InteractionPlugin` `Plugin` impl, otherwise you get a panic because it can't find the scroll event resource.

ickshonpe (commented):
With the event registered, there don't seem to be any other problems.

To test the changes, I removed the `update_scroll_position` from the `testbed_ui` example (in `examples/testbed/ui.rs`) and added an observer to the scrollable node:

```rust
// Scrolling list
parent
    .spawn((
        Node {
            align_self: AlignSelf::Stretch,
            height: Val::Percent(50.),
            overflow: Overflow::scroll_y(),
            ..default()
        },
        BackgroundColor(Color::srgb(0.10, 0.10, 0.10)),
    ))
    .observe(|
        trigger: Trigger<Pointer<Scroll>>,
        mut scroll_position_query: Query<&mut ScrollPosition>,
    | {
        if let Ok(mut scroll_position) = scroll_position_query.get_mut(trigger.target) {
            let (dx, dy) = match trigger.unit {
                MouseScrollUnit::Line => (trigger.x * 20., trigger.y * 20.),
                MouseScrollUnit::Pixel => (trigger.x, trigger.y),
            };
    
            scroll_position.offset_x -= dx;
            scroll_position.offset_y -= dy;
        }
    })
    .with_children(|parent| {
        // List items
        for i in 0..25 {
            parent
                .spawn((
                    Text(format!("Item {i}")),
                    TextFont {
                        ..default()
                    },
                    Label,
                    AccessibilityNode(Accessible::new(Role::ListItem)),
                ))
                .insert(Pickable {
                    should_block_lower: false,
                    ..default()
                });
        }
    });
```

scrolling working as before.

crates/bevy_picking/src/events.rs
```diff
@@ -285,6 +286,19 @@ pub struct DragEntry {
     pub latest_pos: Vec2,
 }
 
+/// Fires while a pointer is scrolling over the `target` entity.
```
ickshonpe, line 289:
```suggestion
/// Fires while a pointer is scrolling over the `target` entity.
```

The way this is written seems a bit odd, does a pointer scroll? Maybe I'm being fussy, I'm not sure.

aevyrie: This is just mirroring the language used in the other event docs. I think it's a valid description. The pointing device sends the scroll, but it is definitely tied to the pointer. That is, pointer scrolling events depend on where the pointer is located w.r.t. picking, it's not just an input.

colepoirier: Does that mean we should leave it as is?

colepoirier:
> You need to add
> 
> ```
>  .add_event::<Pointer<Scroll>>()
> ```
> 
> to `InteractionPlugin`'s `Plugin` impl, otherwise you get a panic because it can't find the scroll event resource.

Done!

crates/bevy_picking/src/events.rs
colepoirier: I removed that and added an appropriate comment mirroring the other events' implementations lower down in that code block.

crates/bevy_picking/src/events.rs
ickshonpe: Yeah it's fine, I was just being nitpicky.

ickshonpe (approved):

aevyrie (approved):
Nice work, thank you!

PR #17710, opened 2025-02-06
Add entity disabling example

alice-i-cecile:
# Objective

The entity disabling / default query filter work added in #17514 and #13120 is neat, but we don't teach users how it works!

We should fix that before 0.16.

## Solution

Write a simple example to teach the basics of entity disabling!

## Testing

`cargo run --example entity_disabling`

## Showcase

image


examples/ecs/entity_disabling.rs
```diff
@@ -0,0 +1,155 @@
...
+    ignored_query: Query<(), Or<(With<Window>, With<Text>)>>,
+    mut commands: Commands,
+) {
+    let clicked_entity = trigger.target();
+    // Windows and text are entities and can be clicked!
+    // We definitely don't want to disable the window itself,
+    // because that would cause the app to close!
+    if ignored_query.contains(clicked_entity) {
+        return;
+    }
```
bushrat011899, line 43: I think it would be nicer to create a `DisableOnClick` marker component and explicitly indicate which components we will allow to be disabled on click. Performs better (not that it's a concern here) and I think is a better demonstration of  the functionality.

examples/ecs/entity_disabling.rs
```diff
@@ -0,0 +1,155 @@
...
+
+    // Just add the `Disabled` component to the entity to disable it.
+    // Note that the `Disabled` component is *only* added to the entity,
+    // its children are not affected.
+    commands.entity(clicked_entity).insert(Disabled);
```
bushrat011899, line 48: That's interesting! Since the entity is disabled, does that cause the parent/child relationship to break, orphaning those children? Or does the relationship stick around and the children just have a disabled parent?

bushrat011899 (approved):
Looks good! I love an interactive example, especially small ones like this. I have an alternate method for avoiding disabling the window and text entities suggested, and some minor documentation items.

examples/ecs/entity_disabling.rs
alice-i-cecile: The latter :) I don't think much of our code is robust to it yet though!

examples/ecs/entity_disabling.rs
```diff
@@ -0,0 +1,155 @@
+//! Disabling entities is a  powerful feature that allows you to hide entities from the ECS without deleting them.
+//!
+//! This can be useful for implementing features like "sleeping" objects that are offscreen
+//! or managing networked entities.
+//!
+//! Note that you should *not* use this feature to simply make entities invisible!
```
Carter0, line 6: It's a little odd that this comment says not to use this feature to make entities invisible, yet the example makes entities invisible! Still I guess its sorta the easiest way to showcase the example

examples/ecs/entity_disabling.rs
```diff
@@ -0,0 +1,155 @@
...
+
+// The query here will not find entities with the `Disabled` component,
+// because it does not explicitly include it.
+fn list_all_named_entities(
+    query: Query<&Name>,
+    mut name_text_query: Query<&mut Text, With<EntityNameText>>,
+    mut commands: Commands,
+) {
+    let mut text_string = String::from("Named entities found:\n");
+    // Query iteration order is not guaranteed, so we sort the names
```
Carter0, line 62: I didn't know that either!

Carter0 (approved):
I ran this example and it works well! Only real feedback is on the visibility comment above, but I don't really think it matters all that much.

rparrett (changes requested):

examples/ecs/entity_disabling.rs
```diff
@@ -0,0 +1,155 @@
...
+//! Note that you should *not* use this feature to simply make entities invisible!
+//! [`Visibility`](bevy::prelude::Visibility) should be used for that:
+//! disabled entities are skipped entirely, which can lead to subtle bugs.
+//!
+//! # Default query filters
+//!
+//! Under the hood, Bevy uses a "default query filter" that skips entities with the
+//! the [`Disabled`] component.
+//! These filters act as a by-default exclusion list for all queries,
+//! and can be bypassed by explicitly including these components in your queries.
```
Diddykonga, line 15: Maybe provide an example of what "explicitly including components in the query" looks like.
`e.x 'Query<(A, B, Disabled)>' or 'Query<(A, B), With<Disabled>>'`

examples/ecs/entity_disabling.rs
```diff
@@ -0,0 +1,155 @@
...
+// because it does not explicitly include it.
+fn list_all_named_entities(
+    query: Query<&Name>,
+    mut name_text_query: Query<&mut Text, With<EntityNameText>>,
+    mut commands: Commands,
+) {
+    let mut text_string = String::from("Named entities found:\n");
+    // Query iteration order is not guaranteed, so we sort the names
+    // to ensure the output is consistent.
+    let mut names: Vec<Name> = query.iter().cloned().collect();
```
Diddykonga, line 64: This would be a good opportunity to use the QueryIter::sort methods.
```rust
let names = query.iter().sort_unstable<&Name>();
```

examples/ecs/entity_disabling.rs
alice-i-cecile: Yeah, that's part of why the name-listing system exists :) It will *also* make things invisible of course!

examples/ecs/entity_disabling.rs
```diff
@@ -0,0 +1,155 @@
...
+            Update,
+            (list_all_named_entities, reenable_entities_on_space),
+        )
+        .add_systems(Startup, (setup_scene, display_instructions))
+        .run();
+}
+
+fn disable_entities_on_click(
+    trigger: Trigger<Pointer<Click>>,
+    ignored_query: Query<(), Or<(With<Window>, With<Text>)>>,
```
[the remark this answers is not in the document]
alice-i-cecile: This will avoid actually fetching / locking the data :) We don't actually care about the data contained inside!

rparrett (approved):

PR #17711, opened 2025-02-06
Public easing

### LINKED ISSUE #17665: Expose `bevy_math::curve::easing::easing_functions` to public

I want to use easing functions directly.

I would like `bevy_math::curve::easing::easing_functions` to have public visibility.

I can use the `simple_easing` crate.

## Additional context

There isn't a reason for these not to be publicly visible.

fjkorf:
# Objective

Make easing functions publicly accessible 
Fixes #17665 

## Solution

add `pub` to mod

## Testing

local build and run

### DIFF
--- a/crates/bevy_math/src/curve/easing.rs
+++ b/crates/bevy_math/src/curve/easing.rs
@@ -367,21 +367,25 @@ pub mod easing_functions {
 
     use crate::{ops, FloatPow};
 
+    /// Linear
     #[inline]
-    pub(crate) fn linear(t: f32) -> f32 {
+    pub fn linear(t: f32) -> f32 {
         t
     }
 
+    /// Quadratic In
     #[inline]
-    pub(crate) fn quadratic_in(t: f32) -> f32 {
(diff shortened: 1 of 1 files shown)


alice-i-cecile (commented):
Is this useful if the functions within are pub(crate)?

alice-i-cecile (approved):

greeble-dev:
These functions can already be used from the public EasingCurve API, although it's a little verbose:

```rust
let x = EasingCurve::new(0.0, 1.0, EaseFunction::SmoothStep).sample_clamped(t);
```

Also these low-level functions do not clamp their inputs, which might surprise people. Unity/Unreal/Godot and shader language smoothsteps all clamp.

On the other hand, having access to these low-level functions might be useful for optimisation. Maybe it's fine if `mod easing_functions` is renamed to `mod unchecked_easing_functions`?

mockersf (changes requested):
I don't think those should be public, they're more of an implementation detail

Issue #17537, opened 2025-01-25
Remove `bevy_pbr::generate_view_layouts`


rambip:
I was surprised to see that `generate_view_layouts` is not used at all in the repository, it would be nice to remove it.

The more useful function, `layout_entries`, is private.

The best thing to do IMO is to make layout_entries public and to deprecate generate_view_layouts.

Utope:
looking at this right now. generate_view_layouts is referenced in this comment. his struct is referenced in a bunch of other places so I can't just remove it. I don't know enough about this system how should i reword it?
```

bitflags::bitflags! {
    /// A key that uniquely identifies a [`MeshPipelineViewLayout`].
    ///
    /// Used to generate all possible layouts for the mesh pipeline in [`generate_view_layouts`],
    /// so special care must be taken to not add too many flags, as the number of possible layouts
    /// will grow exponentially.
    #[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
    #[repr(transparent)]
    pub struct MeshPipelineViewLayoutKey: u32 {
        const MULTISAMPLED                = 1 << 0;
        const DEPTH_PREPASS               = 1 << 1;
        const NORMAL_PREPASS              = 1 << 2;
        const MOTION_VECTOR_PREPASS       = 1 << 3;
        const DEFERRED_PREPASS            = 1 << 4;
        const OIT_ENABLED                 = 1 << 5;
    }
}
```

Utope:
realized you said to deprecate not remove it. So I'm not sure if we should change the comment yet. going to create a pr.

BD103:
This will be closed by #17714, although I'm not able to link them together.

beicause:
`generate_view_layouts` is removed in #23982

PR #13266, opened 2024-05-06
Extract mesh view layouts logic

IceSentry:
Copied almost verbatim from the volumetric fog PR

# Objective

- Managing mesh view layouts is complicated

## Solution

- Extract it to it's own struct
- This was done as part of #13057 and is copied almost verbatim. I wanted to keep this part of the PR it's own atomic commit in case we ever have to revert fog or run a bisect. This change is good whether or not we have volumetric fog.

Co-Authored-By: @pcwalton

PR #17714, opened 2025-02-07
deprecated generate_view_layouts in favor of layout_entries (#17537)

Utope:
# Objective

deprecated generate_view_layouts in favor of layout_entries as described in #17537

### DIFF
--- a/crates/bevy_pbr/src/render/mesh_view_bindings.rs
+++ b/crates/bevy_pbr/src/render/mesh_view_bindings.rs
@@ -195,7 +195,7 @@ pub(crate) fn buffer_layout(
 }
 
 /// Returns the appropriate bind group layout vec based on the parameters
-fn layout_entries(
+pub fn layout_entries(
     clustered_forward_buffer_binding_type: BufferBindingType,
     visibility_ranges_buffer_binding_type: BufferBindingType,
     layout_key: MeshPipelineViewLayoutKey,
@@ -470,6 +470,7 @@ impl MeshPipelineViewLayouts {
 
 /// Generates all possible view layouts for the mesh pipeline, based on all combinations of
 /// [`MeshPipelineViewLayoutKey`] flags.
+#[deprecated(since = "0.16.0", note = "Use layout_entries instead")]
 pub fn generate_view_layouts(
     render_device: &RenderDevice,
     render_adapter: &RenderAdapter,


crates/bevy_pbr/src/render/mesh_view_bindings.rs
```diff
@@ -470,6 +470,7 @@ impl MeshPipelineViewLayouts {
 
 /// Generates all possible view layouts for the mesh pipeline, based on all combinations of
 /// [`MeshPipelineViewLayoutKey`] flags.
+#[deprecated(since = "0.16.0", note = "Use layout_entries instead")]
```
fjkorf, line 473: backticks on `layout_entries`

fjkorf (approved):
I've read through the Issue and the code changes. I have a minor string gripe but otherwise this looks good.

crates/bevy_pbr/src/render/mesh_view_bindings.rs
Utope: Would you like me to change this real quick?

fjkorf: thanks! looks good

IceSentry:
Did someone investigate in which commit generate_view_layout stopped being used? I'm a bit surprised it's dead right now.

Utope:
The issue had Read 'Read-For-Implemenation-Tag' so I assumed someone had checked it

levydsa:
> Did someone investigate in which commit generate_view_layout stopped being used? I'm a bit surprised it's dead right now.

I was exploring the codebase for some unrelated stuff and found that the only two references to `MeshPipelineViewLayoutKey::label` was in the implementation of `FromWorld` of `MeshPipelineViewLayoutKey` and `generate_view_layouts`. They looked strangely similar, what lead me to here.

That `FromWorld` implementation was introduced on #13266 and seams to have inlined the `generate_view_layouts` on the process, leading to it being unused.

crates/bevy_pbr/src/render/mesh_view_bindings.rs
```diff
@@ -470,6 +470,7 @@ impl MeshPipelineViewLayouts {
 
 /// Generates all possible view layouts for the mesh pipeline, based on all combinations of
 /// [`MeshPipelineViewLayoutKey`] flags.
+#[deprecated(since = "0.16.0", note = "Use `layout_entries` instead")]
```
IceSentry, line 473:
```suggestion
#[deprecated(since = "0.19.0", note = "Use `layout_entries` instead")]
```

IceSentry (approved):
Thank you for investigating! I completely forgot about this PR but it should be good to go now.

Issue #13735, opened 2024-06-07
EntityRef::get_ref should always use the ticks of the system if possible


cBournhonesque:
## What problem does this solve or what need does it fill?

1) When you query a `Ref<C>`, the `last_run` tick inside `Ref<C>` is populated from the last time the system has run.
2) This is also the case when you call `EntityRef::get_ref::<C>` from within an exclusive system, it is also the case.
(probably because of https://docs.rs/bevy/latest/bevy/ecs/world/struct.World.html#method.last_change_tick

> When called from within an exclusive system (a [System](https://docs.rs/bevy/latest/bevy/ecs/system/trait.System.html) that takes &mut World as its first parameter), this method returns the [Tick](https://docs.rs/bevy/latest/bevy/ecs/component/struct.Tick.html) indicating the last time the exclusive system was run.

3) Otherwise, this returns the Tick indicating the last time that [World::clear_trackers](https://docs.rs/bevy/latest/bevy/ecs/world/struct.World.html#method.clear_trackers) was called.)

But when you call `EntityRef::get_ref::<C>` outside of an exclusive system, the `last_run` tick is from the last time `World::clear_trackers` was called (and it's called in every `Last` schedule).

I think this is fairly misleading, as users are taught to expect that `Ref::is_added()` and `Ref::is_changed()` will return true if the component was added/changed since **the last time the system was run**.

(in my use case, I was calling `EntityRef::get_ref` from a system that uses the `&World` SystemParam, so I was in case 3) )

Some code that shows the current behaviour:
```rust
#[test]
    fn test_ref() {
        let mut world = World::new();
        world.init_component::<A>();

        let entity = world.spawn(A).id();
        let update_a = world.register_system(move |world: &mut World | {
            world.entity_mut(entity).get_mut::<A>().unwrap().set_changed();
        });

        let check_change_detection_via_entity_ref = world.register_system(move |world: &World| {
            let a_ref = world.entity(entity).get_ref::<A>().unwrap();
            // dbg!(&a_ref.ticks);
            dbg!("via entity_ref: ", a_ref.is_changed());
        });
        let check_change_detection_via_entity_ref_in_exclusive_system = world.register_system(move |world: &mut World| {
            let a_ref = world.entity(entity).get_ref::<A>().unwrap();
            // dbg!(&a_ref.ticks);
            dbg!("via entity_ref in exclusive system: ", a_ref.is_changed());
        });
[...]

PR #17716, opened 2025-02-07
Expose method to update the internal ticks of Ref and Mut

cBournhonesque:

There are some situations (https://github.com/bevyengine/bevy/issues/13735) where the ticks that are present inside `Ref` are incorrect, for example if `Ref` is created outside of a `SystemParam`.
I still want to use `Ref` because it has convenient `is_added` and `is_changed` methods.

My current solution is to build my own `Ref` by copy-pasting most the bevy code to do that via something like
```rust
/// This method is necessary because there is no easy way to 
pub(crate) fn get_ref<C: Component>(
    world: &World,
    entity: Entity,
    last_run: Tick,
    this_run: Tick,
) -> Ref<C> {
    unsafe {
        let component_id = world
            .components()
            .get_id(TypeId::of::<C>())
            .unwrap_unchecked();
        let world = world.as_unsafe_world_cell_readonly();
        let entity_cell = world.get_entity(entity).unwrap_unchecked();
        get_component_and_ticks(
            world,
            component_id,
            C::STORAGE_TYPE,
            entity,
            entity_cell.location(),
        )
        .map(|(value, cells, _caller)| {
            Ref::new(
                value.deref::<C>(),
                cells.added.deref(),
                cells.changed.deref(),
                last_run,
                this_run,
                #[cfg(feature = "track_location")]
                _caller.deref(),
            )
        })
        .unwrap_unchecked()
    }
}

// Utility function to return
#[inline]
unsafe fn get_component_and_ticks(
    world: UnsafeWorldCell<'_>,
    component_id: ComponentId,
    storage_type: StorageType,
    entity: Entity,
    location: EntityLocation,
) -> Option<(Ptr<'_>, TickCells<'_>, MaybeUnsafeCellLocation<'_>)> {
    match storage_type {
        StorageType::Table => {
            let table = unsafe { world.storages().tables.get(location.table_id) }?;

            // SAFETY: archetypes only store valid table_rows and caller ensure aliasing rules
            Some((
                table.get_component(component_id, location.table_row)?,
                TickCells {
                    added: table
                        .get_added_tick(component_id, location.table_row)
                        .unwrap_unchecked(),
                    changed: table
                        .get_changed_tick(component_id, location.table_row)
                        .unwrap_unchecked(),
                },
                #[cfg(feature = "track_location")]
                table
                    .get_changed_by(component_id, location.table_row)
                    .unwrap_unchecked(),
                #[cfg(not(feature = "track_location"))]
                (),
            ))
        }
        StorageType::SparseSet => {
            let storage = unsafe { world.storages() }.sparse_sets.get(component_id)?;
            storage.get_with_ticks(entity)
        }
    }
}
```

It would be very convenient if instead bevy exposed a way to create a `Ref` object with custom `last_run` and `this_run` ticks.
This PR does this by exposing a function to overwrite the `last_run` and `this_run` ticks.
(Same with `Mut`)

I am ok with marking the method unsafe or risky if it's deemed to risky for end-users.

### DIFF
--- a/crates/bevy_ecs/src/change_detection.rs
+++ b/crates/bevy_ecs/src/change_detection.rs
@@ -859,6 +859,15 @@ impl<'w, T: ?Sized> Ref<'w, T> {
             changed_by: caller,
         }
     }
+
+    /// Overwrite the `last_run` and `this_run` tick that are used for change detection.
+    ///
+    /// This is an advanced feature. `Ref`s are usually _created_ by engine-internal code and
+    /// _consumed_ by end-user code.
+    pub fn set_ticks(&mut self, last_run: Tick, this_run: Tick) {
+        self.ticks.last_run = last_run;
+        self.ticks.this_run = this_run;
+    }
 }
 
 impl<'w, 'a, T> IntoIterator for &'a Ref<'w, T>
@@ -977,6 +986,15 @@ impl<'w, T: ?Sized> Mut<'w, T> {
             #[cfg(feature = "track_location")]
             changed_by: caller,
         }
+    }
+
+    /// Overwrite the `last_run` and `this_run` tick that are used for change detection.
+    ///
+    /// This is an advanced feature. `Mut`s are usually _created_ by engine-internal code and
+    /// _consumed_ by end-user code.
+    pub fn set_ticks(&mut self, last_run: Tick, this_run: Tick) {
+        self.ticks.last_run = last_run;
+        self.ticks.this_run = this_run;
     }
 }
 


chescock (approved):
Yeah, there are too many places in Bevy where we pull the ticks out of the `World` without any way to set them correctly from within a system.  This seems like a good way to let users work around that issue!  

I do think we could at least make `Query<EntityRef>` work automatically, but making it automatic for methods on `&World` will be tough.  

> It would be very convenient if instead bevy exposed a way to create a `Ref` object with custom `last_run` and `this_run` ticks.

My first thought was to note that `Ref::new()` lets you do that.  But you can't get the `added` and `changed` ticks back *out* of an existing `Ref`, so there's still no way to map an existing `Ref` to one with different system ticks!  (I assume you know this, but I'm writing it in case other reviewers have the same first reaction that I did.)

alice-i-cecile:
> Exactly, exposing methods to get the added/changed ticks from an existing Ref would also work for me!

I would like this! Would this completely replace this PR?

I'm alright with these methods existing as a carefully documented escape hatch, but I'd much prefer other ways to avoid opening this can of worms.

cBournhonesque:
> > Exactly, exposing methods to get the added/changed ticks from an existing Ref would also work for me!
> 
> I would like this! Would this completely replace this PR?
> 
> I'm alright with these methods existing as a carefully documented escape hatch, but I'd much prefer other ways to avoid opening this can of worms.

Actually this wouldn't work for me. I remember now why `Ref::new()` didn't work for my usecase, it's because of the addition of the `track_location` feature. If `track_location` is enabled I need to pass a `&'static Location` which is only accessible in `unsafe fn get_component_and_ticks`.

`unsafe fn get_component_and_ticks` requires me to provide ECS internals such as `EntityLocation` and `StorageType`.

Maybe I can provide methods `get_location` which returns `&'static Location` needed for the `track_location` feature.

You think that being able to modify the ticks of a Ref is too dangerous? it's true that it is unneeded for 99%+ of usecases

chescock:
> If `track_location` is enabled I need to pass a `&'static Location` which is only accessible in `unsafe fn get_component_and_ticks`.

If that feature is enabled, then you can get one out of the old `Ref` by calling `changed_by()`, right?  

Oh, that might be more difficult now that #17602 is merged, though.  Now `Ref::new()` requires a `MaybeLocation<&'w &'static Location<'static>>` for consistency with `Mut`, but `Ref::changed_by()` only returns a `MaybeLocation<&'static Location<'static>>`.  

We could work around that for `Ref` by adding methods to return the underlying references, but for `Mut` there wouldn't be a way to get the `&'w mut T` *and* the `MaybeLocation<&'w mut &'static Location<'static>>` with full `'w` lifetimes without a consuming method that returned both... and at that point you're returning everything except for `Ticks`, so it's just a more roundabout way to do what this PR is doing.  

> I'm alright with these methods existing as a carefully documented escape hatch, but I'd much prefer other ways to avoid opening this can of worms.

Since the public `Ref::new()` function lets you pass in your own `Tick` values, isn't the can already open?

alice-i-cecile (approved):
Okay, I've chewed on this and poked around. This is definitely the simplest solution, and a pretty natural escape hatch. We already allow for manually setting change ticks in DetectChangesMut, so this really isn't doing anything new.

Merging.

PR #17717, opened 2025-02-07
Fix motion vector computation after #17688.

pcwalton:
PR #17688 broke motion vector computation, and therefore motion blur, because it enabled retention of `MeshInputUniform`s, and `MeshInputUniform`s contain the indices of the previous frame's transform and the previous frame's skinned mesh joint matrices. On frame N, if a `MeshInputUniform` is retained on GPU from the previous frame, the `previous_input_index` and `previous_skin_index` would refer to the indices for frame N - 2, not the index for frame N - 1.

This patch fixes the problems. It solves these issues in two different ways, one for transforms and one for skins:

1. To fix transforms, this patch supplies the *frame index* to the shader as part of the view uniforms, and specifies which frame index each mesh's previous transform refers to. So, in the situation described above, the frame index would be N, the previous frame index would be N - 1, and the `previous_input_frame_number` would be N - 2. The shader can now detect this situation and infer that the mesh has been retained, and can therefore conclude that the mesh's transform hasn't changed.

2. To fix skins, this patch replaces the explicit `previous_skin_index` with an invariant that the index of the joints for the current frame and the index of the joints for the previous frame are the same. This means that the `MeshInputUniform` never has to be updated even if the skin is animated. The downside is that we have to copy joint matrices from the previous frame's buffer to the current frame's buffer in `extract_skins`.

The rationale behind (2) is that we currently have no mechanism to detect when joints that affect a skin have been updated, short of comparing all the transforms and setting a flag for `extract_meshes_for_gpu_building` to consume, which would regress performance as we want `extract_skins` and
`extract_meshes_for_gpu_building` to be able to run in parallel.

To test this change, use `cargo run --example motion_blur`.


tychedelia (approved):
Makes sense. Tested on macOS.

crates/bevy_render/src/view/mod.rs
```diff
@@ -876,6 +878,12 @@ impl ViewDepthTexture {
     }
 }
 
+/// A number that starts at 0 and increments by 1 every frame.
+///
+/// This is useful for determining which frame cached data belongs to.
+#[derive(Resource, Default, Deref, DerefMut)]
+pub struct FrameNumber(pub u32);
```
IceSentry, line 885: The FrameCount resouce already provides that.

tychedelia: I was just wondering whether we had this today.

pcwalton: Oh cool, thanks. I removed `FrameNumber` in favor of `FrameCount`.

aevyrie:
Motion blur seems to be broken, not sure if introduced here though
image

aevyrie (approved):
Issues appear to be resolved in latest commit.

crates/bevy_pbr/src/render/skin.rs
```diff
@@ -232,3 +233,18 @@ pub fn no_automatic_skin_batching(
         commands.entity(entity).try_insert(NoAutomaticBatching);
     }
 }
+
+pub fn mark_meshes_as_changed_if_their_skins_changed(
```
superdump, line 237: For consistency:
```suggestion
pub fn mark_3d_meshes_as_changed_if_their_skins_changed(
```

It would be nice to port all this to 2D as well... :)

crates/bevy_pbr/src/render/mesh.rs
```diff
@@ -536,17 +546,15 @@ pub struct MeshInputUniform {
...
     /// The current skin index, or `u32::MAX` if there's no skin.
     pub current_skin_index: u32,
-    /// The previous skin index, or `u32::MAX` if there's no previous skin.
     pub previous_skin_index: u32,
     /// The material and lightmap indices, packed into 32 bits.
     ///
     /// Low 16 bits: index of the material inside the bind group data.
     /// High 16 bits: index of the lightmap in the binding array.
     pub material_and_lightmap_bind_group_slot: u32,
+    pub timestamp: u32,
```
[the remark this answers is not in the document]
superdump: Maybe `frame_count_when_updated` or something like that? That seems to be the semantics of it.

superdump (approved):
Just a minor variable naming thing for clarity. Otherwise LGTM.

pcwalton:
This should be ready to go assuming CI is green.

superdump (approved):

PR #17719, opened 2025-02-07
Move specialize_* to QueueMeshes.

tychedelia:
# Objective

Things were breaking post-cs.

## Solution

`specialize_mesh_materials` must run after `collect_meshes_for_gpu_building`. Therefore, its placement in the `PrepareAssets` set didn't make sense (also more generally). To fix, we put this class of system in ~`PrepareResources`~ `QueueMeshes`, although it potentially could use a more descriptive location. We may want to review the placement of `check_views_need_specialization` which is also currently in `PrepareAssets`.

### DIFF
--- a/crates/bevy_pbr/src/material.rs
+++ b/crates/bevy_pbr/src/material.rs
@@ -314,9 +314,10 @@ where
                     Render,
                     (
                         specialize_material_meshes::<M>
-                            .in_set(RenderSet::PrepareResources)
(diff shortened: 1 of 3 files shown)


HackerFoo (approved):
I tested these changes on my development build of Noumenal, and the glitches caused by https://github.com/bevyengine/bevy/pull/17567

pcwalton (approved):
LGTM

superdump (approved):

PR #17724, opened 2025-02-07
`Query::get_many` should not check for duplicates

chescock:
# Objective

Restore the behavior of `Query::get_many` prior to #15858.  

When passed duplicate `Entity`s, `get_many` is supposed to return results for all of them, since read-only queries don't alias.  However, #15858 merged the implementation with `get_many_mut` and caused it to return `QueryEntityError::AliasedMutability`.  

## Solution

Introduce a new `Query::get_many_readonly` method that consumes the `Query` like `get_many_inner`, but that is constrained to `D: ReadOnlyQueryData` so that it can skip the aliasing check.  Implement `Query::get_many` in terms of that new method.  Add a test, and a comment explaining why it doesn't match the pattern of the other `&self` methods.

### DIFF
--- a/crates/bevy_ecs/src/query/mod.rs
+++ b/crates/bevy_ecs/src/query/mod.rs
@@ -439,6 +439,18 @@ mod tests {
     }
 
     #[test]
+    fn get_many_only_mut_checks_duplicates() {
+        let mut world = World::new();
+        let id = world.spawn(A(10)).id();
+        let mut query_state = world.query::<&mut A>();
+        let mut query = query_state.query_mut(&mut world);
+        let result = query.get_many([id, id]);
+        assert_eq!(result, Ok([&A(10), &A(10)]));
+        let mut_result = query.get_many_mut([id, id]);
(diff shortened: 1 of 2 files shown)


chescock:
> Yeah, but wouldn't you always want to check for duplicates? Like having ([player_id, player_id]) would be weird even if you can't alter the player data cause it's read-only.

I *think* the goal there is just performance.  Doing the check isn't free.  We need it for mutable data for soundness, but for read-only data we can skip it.  

But I haven't actually used this method myself, so I may not be the best person to justify it :).

Carter0:
> > Yeah, but wouldn't you always want to check for duplicates? Like having ([player_id, player_id]) would be weird even if you can't alter the player data cause it's read-only.
> 
> I _think_ the goal there is just performance. Doing the check isn't free. We need it for mutable data for soundness, but for read-only data we can skip it.
> 
> But I haven't actually used this method myself, so I may not be the best person to justify it :).

yoki doki makes sense

PR #17725, opened 2025-02-07
Button example using `bevy_picking`

ickshonpe:
# Objective

Rewrite the `UI` button example to use `bevy_picking`

## Solution

Couldn't use `PickingInteration` by itself because unlike with `Interaction` the `Pressed` state is changed to `None` when the pointer leaves the UI node, so I added a `ButtonState` component that tracks the pressed state updated by some observers. 

Not very happy with this, feels like there should be a much simpler construction I'm missing. I tried a version that used the `PointerState` resource, but it was even more awkward.


hukasu:
Wouldn't be better for the trigger to edit the values of the button on them instead of having to set a flag and then on a system update the values?

hukasu:
Was closing this intended?

PR #17726, opened 2025-02-07
Fix cloning entities with multiple `linked_spawn` relationships

eugineerd:
# Objective
#17687 changed recursive cloning to queue cloning commands for entities in `linked_spawn` `RelationshipTarget`s. However, if there are multiple relationships with `linked_spawn` on the same entity, multiple clone commands will be queued for the same entity.

## Solution
Maintain a list of already cloned entities during one `clone_entity_mapped` call and skip as necessary.

## Testing
- Added a test when cloning entities with multiple `linked_spawn` relations.

### DIFF
--- a/crates/bevy_ecs/src/entity/clone_entities.rs
+++ b/crates/bevy_ecs/src/entity/clone_entities.rs
@@ -9,6 +9,7 @@ use alloc::boxed::Box;
 
 use crate::component::{ComponentCloneBehavior, ComponentCloneFn};
 use crate::entity::hash_map::EntityHashMap;
+use crate::entity::hash_set::EntityHashSet;
 use crate::entity::EntityMapper;
 use crate::system::Commands;
 use crate::{
@@ -273,7 +274,9 @@ impl<'a, 'b> ComponentCloneCtx<'a, 'b> {
     }
 
     /// Queues the `entity` to be cloned by the current [`EntityCloner`]
-    pub fn queue_entity_clone(&self, entity: Entity) {
+    ///
+    /// This will create exactly one clone, even if called multiple times with the same `entity`.
+    pub fn queue_entity_clone_once(&self, entity: Entity) {
         self.entity_cloner
             .clone_queue
             .borrow_mut()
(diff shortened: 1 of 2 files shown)


alice-i-cecile (approved):
Good fix. I'd like to also see a test with multiple linked relations joining the same pair of entities, to make sure we're not panicking or something there. But that's not strictly necessary, so I'm approving.

eugineerd:
> a test with multiple linked relations joining the same pair of entities

What configuration do you mean by that? The test in this PR currently looks like this:
```mermaid
flowchart LR
    child --ChildOf--> root
    child2 --ChildOf--> child
    child --Rel--> root
    child2 --Rel--> root
```

alice-i-cecile:
Ah I see, you already have that as a subset of what's being tested :) That works for me!

crates/bevy_ecs/src/entity/clone_entities.rs
```diff
@@ -564,12 +567,17 @@ impl EntityCloner {
         mapper: &mut dyn EntityMapper,
     ) -> Entity {
         let target = self.clone_entity_internal(world, source, mapper);
+        let mut already_cloned = EntityHashSet::new();
```
cart, line 570: This is the right fix if we really need this behavior. But given that this adds marginal (but still present) overhead, we should discuss whether we actually need this behavior. From my perspective LINKED_SPAWN is a form of ownership, and this adds support for "multiple owners". Do you have a use case in mind for this, or is it a purely theoretical fix? If we can't find solid real world use cases, I think the alternative is to just document that multiple LINKED_SPAWN relationships on the same entity will result in duplicates on clone.

eugineerd: Just something I've noticed, no real use for it yet. I think one use case for this is @viridia's `DynChildren`, but I'm not sure. There isn't really any protection from having multiple `linked_spawn` components on one entity, so if a situation like that arises it'd be a bit strange to see only some things break. One alternative would be to restrict the `recursive` behavior to `Children` instead of all `linked_spawn` relations. Another is to ensure only one `linked_spawn` component can exist on an entity (but this is probably less practical). The performance hit doesn't seem to be that big for the normal case (~5% on my machine), but maybe we can wait until someone actually opens an issue for this?

alice-i-cecile: Good point on the overhead. IMO we should go with docs for now, and then consider this fix once we have a true bug report / use case.

PR #17728, opened 2025-02-07
Cleanup publish process

mockersf:
# Objective

- publish script copy the license files to all subcrates, meaning that all publish are dirty. this breaks git verification of crates
- the order and list of crates to publish is manually maintained, leading to error. cargo 1.84 is more strict and the list is currently wrong

## Solution

- duplicate all the licenses to all crates and remove the `--allow-dirty` flag
- instead of a manual list of crates, get it from `cargo package --workspace`
- remove the `--no-verify` flag to... verify more things?

### DIFF
--- a/crates/bevy_a11y/LICENSE-APACHE
+++ b/crates/bevy_a11y/LICENSE-APACHE
@@ -1,0 +1,176 @@
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+   1. Definitions.
+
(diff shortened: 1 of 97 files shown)


cart (approved):

alice-i-cecile:
Prompted by https://github.com/rust-lang/cargo/issues/15151

mockersf:
> Prompted by [rust-lang/cargo#15151](https://github.com/rust-lang/cargo/issues/15151)

no, that one has just been itching me for a while... it's https://github.com/bevyengine/bevy/pull/17727 that's related to that cargo issue

alice-i-cecile (approved):
I low-key hate the duplication of licenses here, but I agree that it's the least bad option.

PR #17736, opened 2025-02-08
Set late indirect parameter offsets every frame again.

pcwalton:
PR #17684 broke occlusion culling because it neglected to set the indirect parameter offsets for the late mesh preprocessing stage if the work item buffers were already set. This PR moves the update of those values to a new function, `init_work_item_buffers`, which is unconditionally called for every phase every frame.

Note that there's some complexity in order to handle the case in which occlusion culling was enabled on one frame and disabled on the next, or vice versa. This was necessary in order to make the occlusion culling toggle in the `occlusion_culling` example work again.


superdump (approved):
LGTM

superdump:
Though CI disagrees :)

tychedelia (approved):
Makes sense, thanks!

IceSentry (approved):

PR #17739, opened 2025-02-08
feat(ecs): add `Assume` and `Unpack` traits for `Result` conversion

JeanMertz:
Adds the following traits, and implements them for `Option`:

```rust
/// Unpack `Self<T>` to `T`, otherwise return [`Unpack::Error`].
///
/// This can be a drop-in replacement for `unwrap`, combined with the question mark operator and
/// [`Result`](super::Result) return type, to get the same ergonomics as `unwrap` but without the
/// panicking behavior (when using a non-panicking error handler).
pub trait Unpack<T> {
    /// The error type returned by [`Unpack::unpack`].
    ///
    /// Typically implements the [`Error`] trait, allowing it to match Bevy's fallible system
    /// [`Result`](super::Result) return type.
    type Error;

    /// Convert `Self<T>` to a `Result<T, Self::Error>`.
    fn unpack(self) -> Result<T, Self::Error>;
}

/// Assume that `Self<T>` is `T`, otherwise return the provided error.
///
/// This can be a drop-in replacement for `expect`, combined with the question mark operator and
/// [`Result`](super::Result) return type, to get the same ergonomics as `expect` but without the
/// panicking behavior (when using a non-panicking error handler).
pub trait Assume<T> {
    /// The error type returned by [`Assume::assume`].
    ///
    /// Typically implements the [`Error`] trait, allowing it to match Bevy's fallible system
    /// [`Result`](super::Result) return type.
    type Error;

    /// Convert `Self<T>` to a `Result<T, Self::Error>`.
    fn assume<E: Into<Self::Error>>(self, err: E) -> Result<T, Self::Error>;
}
```

This allows for a more concise and readable way of handling `Option` values in Bevy systems that return `Result`s, by applying the question mark operator:

```rust
fn my_system(world: &World) -> bevy::ecs::Result {
    world.get::<MyComponent>(Entity::PLACEHOLDER).unpack()?;

    world.get_resource::<MyResource>().assume("resource exists")?;

    Ok(())
}
```

In future commits, more of Bevy's API will be converted to return `Result`s directly, but there will always be (non-)Bevy methods that return `Option` values, so this extension trait will have its place, regardless.

This was suggested by @cart here: https://github.com/bevyengine/bevy/issues/14275#issuecomment-2223733195


mockersf:
> c) we come up with a better place to combine all of this.

I think the better place for this is Rust std. Could you try opening the discussion there?

JeanMertz:
> I think the better place for this trait at least is Rust std. Could you try opening the discussion there?

I’m happy to, but I don’t suppose you propose waiting for that resolution? Any suggestion where it should live until/when it lands in Rust std?

mockersf:
> I’m happy to, but I don’t suppose you propose waiting for that resolution?

Yes I do? If it happens fast, it could be in rust 1.86 which may happen before (or around) the next version of Bevy

JeanMertz:
> > I’m happy to, but I don’t suppose you propose waiting for that resolution?
> 
> Yes I do? If it happens fast, it could be in rust 1.86 which may happen before (or around) the next version of Bevy

I see. I would propose that both can be done in parallel; add the convenience methods in Bevy to help with fallible system adoption, and concurrently propose upstreaming the methods. If they are upstreamed, we can remove the custom impl here, if it's not upstreamed, then we don't need to change anything on our side.

Regardless, I'll find some time next week to propose the methods upstream.

mockersf:
> I see. I would propose that both can be done in parallel; add the convenience methods in Bevy to help with fallible system adoption, and concurrently propose upstreaming the methods. If they are upstreamed, we can remove the custom impl here, if it's not upstreamed, then we don't need to change anything on our side.

I would prefer to wait, this is not that much better that it's worth changing then reverting

alice-i-cecile:
I like the end result, but I'm not super keen on shipping this in Bevy itself. Extension traits on super common types like Option are likely to badly confuse beginners. Would enthusiastically support using this if it was in the `std` though!

JeanMertz:
I haven’t looked too deep yet, but `NoneError` existed in Rust before, but was removed and there have apparently been many lengthy discussions on it. So I would not hold my breath on that landing any time soon. I’ll dig a bit deeper on Monday and see if opening a PR is worth the effort or not.

JeanMertz:
Alright, I looked into it more, I really don't see _any_ chance of this getting upstreamed.

For `fn assume<E: Into<Box<dyn Error>>>(self, err: E) -> Result<T, Box<dyn Error>>;`, this is _close_ to Rust's [`ok_or`](https://doc.rust-lang.org/1.84.1/std/option/enum.Option.html#method.ok_or), except that there `E` is generic, whereas we explicitly take anything that can be converted into the specific error type returned by our fallible systems. Now, the question mark operator `?` already applies `From::from` to the error value, but there are still cases where you have to append a `map_err(Into::into)` to make it work. So you go from:

```rust
optional.assume(SomeError)?
```

to (worst case):

```rust
optional.ok_or(SomeError).map_err(Into::into)?
```

_(I should note that `optional.assume(SomeError)` doesn't _read_ correctly, while `optional.assume("should exist")` does, as does `optional.ok_or(SomeError)`, but that's a different discussion)_

While I _could_ make a case for `assume` to be valuable (given that our `Error` type is just a type alias for `Box<dyn Error>`), it's also the case that `std::result::Result<T, E>` doesn't make any assumptions about what type `E` is, in fact there is no method in there that expects `E: Error`, as far as I can tell. That is what makes `assume` more convenient in our case, because it expects its input to be convertible to our system error return type.

For `fn unpack(self) -> Result<T, NoneError>;`, there are [many](https://github.com/rust-lang/rust/issues/46871) [discussions](https://internals.rust-lang.org/t/should-noneerror-implement-the-error-trait) to be [found](https://github.com/rust-lang/rust/issues/42327#issuecomment-366840247) on the topic. But here's basically [the most succinct explanation](https://github.com/rust-lang/rust/issues/46871#issuecomment-618186642) on why `NoneError` was removed from Rust. Their reasoning for removal is sound, but it does mean our fallible systems are a _little bit_ less ergonomical without this extension trait. I don't feel it's worth it for me to re-open that discussion, nor do I want to waste people's time re-hashing resolved discussions.

---

Having said all that. I'm happy to close this and use the extension trait in my own projects. I added this because it was proposed in https://github.com/bevyengine/bevy/issues/14275#issuecomment-2223733195, but if we think the gain in convenience is too small for the cost of people having to learn non-std-Rust methods, then I can see that reasoning as well.

alice-i-cecile:
We'll leave this open and let @cart decide when he gets around to it :) This is just a matter of taste IMO.

crates/bevy_utils/src/option_ext.rs
```diff
@@ -0,0 +1,90 @@
+//! Extensions to the [`Option`] type used in Bevy.
```
cart, line 1: I think these should live in `bevy_ecs::error`. Currently our "boxed error type" is "just" `core::error::Error`, but there is a _very_ good chance that we will be swapping that for something custom (or anyhow). The `assume` coercion will be tied directly to the bevy ecs error type, so it should live there.

crates/bevy_utils/src/option_ext.rs
```diff
@@ -0,0 +1,90 @@
...
+impl fmt::Display for NoneError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "Unexpected None value.")
+    }
+}
+
+impl Error for NoneError {}
+
+/// Extension trait for [`Option`].
+pub trait OptionExt<T> {
```
cart, line 19: For clarity I think we can afford to split this into two different traits: `Unpack` and `Assume`.

crates/bevy_utils/src/lib.rs
```diff
@@ -20,8 +20,12 @@ extern crate alloc;
 /// This includes the most common types in this crate, re-exported for your convenience.
 pub mod prelude {
     pub use crate::default;
+    #[cfg(feature = "alloc")]
+    pub use crate::option_ext::OptionExt as _;
```
cart, line 24: I think the most controversial aspect of this PR is including the traits in the prelude. We could hold off on this until we're certain that these are used at a frequency to justify it.

crates/bevy_utils/src/option_ext.rs
cart, line 1: I also think the odds of these being included in `std` are slim. And `assume`, once Bevy's error type isn't just a boxed core Error, couldn't possibly live in `std`. And even if `unpack` were to land in std, it would likely be closer to the "years" timeframe than the "months" timeframe.

cart (commented):
I think this is a good move, with some minor changes.

alice-i-cecile:
Agreed with Cart's feedback, and I'm fine to follow his direction on this. I agree with his / your assumption that upstreaming will be *hard*.

crates/bevy_utils/src/option_ext.rs
JeanMertz: Resolved in [210faa7](https://github.com/bevyengine/bevy/pull/17739/commits/210faa7418348289342a517c183eca8e6f7c30c7).

crates/bevy_utils/src/option_ext.rs
JeanMertz: Resolved in [210faa7](https://github.com/bevyengine/bevy/pull/17739/commits/210faa7418348289342a517c183eca8e6f7c30c7).

JeanMertz:
This PR has been rebased on main and all feedback has been resolved.

crates/bevy_ecs/src/result.rs
```diff
@@ -61,6 +61,33 @@
 //! If you need special handling of individual fallible systems, you can use Bevy's [`system piping
 //! feature`] to capture the `Result` output of the system and handle it accordingly.
 //!
+//! # Conveniently returning `Result`
+//!
+//! The [`Unpack`] and [`Assume`] traits can be used to unpack and assume any value into a
```
alice-i-cecile, line 66:
```suggestion
//! The [`Unpack`] and [`Assume`] traits can be used to transform any value into a
```

alice-i-cecile: Explaining something using the same words is generally unhelpful :)

crates/bevy_ecs/src/result.rs
```diff
@@ -61,6 +61,33 @@
...
+//! [`Result`] that can be handled by Rust's `?` operator. This makes working with fallible systems
+//! more ergonomic.
+//!
```
alice-i-cecile, line 69:
```suggestion
//!
//! [`Unpack::unpack`] should be used in place of `unwrap` (to quickly get a value,
//! while [`Assume::assume`] should be used in place of `expect`
//! (if you also want a string explaining the assumption that was violated).
```

crates/bevy_ecs/src/result/assume.rs
```diff
@@ -0,0 +1,59 @@
+use super::Error;
+
+/// Assume that `Self<T>` is `T`, otherwise return the provided error.
```
alice-i-cecile, line 3: This needs more context on why this might be useful, and the fact that it replaces `unwrap`.

crates/bevy_ecs/src/result/unpack.rs
```diff
@@ -0,0 +1,49 @@
+use core::{error::Error, fmt};
+
+/// Unpack `Self<T>` to `T`, otherwise return [`Unpack::Error`].
```
alice-i-cecile, line 3: Ditto on context here.

crates/bevy_ecs/src/result/assume.rs
```diff
@@ -0,0 +1,59 @@
...
+pub trait Assume<T> {
+    /// The error type returned by [`Assume::assume`].
+    type Error;
+
+    /// Convert `Self<T>` to a `Result<T, Self::Error>`.
```
alice-i-cecile, line 8: These docs will not be very helpful on mouse-over tooltips.

crates/bevy_ecs/src/result/assume.rs
```diff
@@ -0,0 +1,59 @@
+use super::Error;
+
+/// Assume that `Self<T>` is `T`, otherwise return the provided error.
+pub trait Assume<T> {
+    /// The error type returned by [`Assume::assume`].
```
alice-i-cecile, line 5: Context about Bevy's `Result` type would be useful here.

alice-i-cecile, line 5: While I don't think we should make it a trait bound, we should mention that these types should typically implement the `Error` trait.

crates/bevy_ecs/src/result/unpack.rs
```diff
@@ -0,0 +1,49 @@
...
+
+impl<T> Unpack<T> for Option<T> {
+    type Error = NoneError;
+
+    fn unpack(self) -> Result<T, Self::Error> {
+        self.ok_or(NoneError)
+    }
+}
+
+/// A custom type which implements [`Error`], used to indicate that an [`Option`] was [`None`].
```
alice-i-cecile, line 20:
```suggestion
/// An [`Error`] which indicates that an [`Option`] was [`None`].
```

alice-i-cecile:
A round of documentation hardening for you :) New users are going to see these traits / methods pretty prominently, so I want to make sure that they're approachable and have the context needed to defend the design.

alice-i-cecile (approved):

cart:
Ok sorry for the back-and-forth, but I talked this over with @mockersf and I've come around to his perspective. First, I've been informed that this already works on Bevy `main`:

```rust
fn system() -> Result {
  let x = Option(10);
  x.ok_or("x does not exist")?;
  Ok(())
}
```

And once we transition to anyhow-style erroring in our result type, this will still capture the stacktrace in the right spot during the From conversion.

This makes `ok_or` directly equivalent to `assume`, so there is no reason to include it.

From there, it does start to make sense to aim for `ok_or` "pattern parity" and "rust nativeness". `ok_or(NoneError)?` is about as hard to type and less useful than `ok_or("helpful message")?`. Most of our built-in APIs should return Result with a helpful error message. And any app code calling something that doesn't probably _should_ provide a helpful error message. So I'm inclined to just encourage that pattern for now and see how it goes.

I _don't_ think we should immediately port every `unwrap()` to `ok_or("error message")?` in our codebase, as that would both be laborious and produce way too much noise in the codebase. Instead, we should focus on making APIs return Result where it make sense, replace those instances of `unwrap()` with error handling, and then consider what to do with the remaining unwraps _after_ that.

JeanMertz:
Reasonable perspective!

Copy/pasted over to my own project(s) for convenience, until Bevy transitions to anyhow-style erroring.

PR #17742, opened 2025-02-08
Tidy up `easing_functions` example

rparrett:
# Objective

After #17461, the ease function labels in this example are a bit cramped, especially in the bottom row.

This adjusts the spacing slightly and centers the labels.

## Solution

- The label is now a child of the plot and they are drawn around the center of the transform
- Plot size and extents are now constants, and this thing has been banished:
  
  ```rust
  i as f32 * 95.0 - 1280.0 / 2.0 + 25.0,
  -100.0 - ((j as f32 * 250.0) - 300.0),
  0.0,
  ```

- There's room for expansion in another row, so make that easier by doing the chunking by row
- Other misc tidying of variable names, sprinkled in a few comments, etc.

## Before

Screenshot 2025-02-08 at 7 33 14 AM

## After

Screenshot 2025-02-08 at 7 32 45 AM

### DIFF
--- a/examples/animation/easing_functions.rs
+++ b/examples/animation/easing_functions.rs
@@ -1,9 +1,10 @@
 //! Demonstrates the behavior of the built-in easing functions.
 
-use bevy::{prelude::*, sprite::Anchor};
+use bevy::prelude::*;
 
 #[derive(Component)]
-struct SelectedEaseFunction(EaseFunction, Color);
+#[require(Visibility, Transform)]
+struct EaseFunctionPlot(EaseFunction, Color);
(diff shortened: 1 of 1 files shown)


mockersf:
what do you think of starting the fourth row now, with the earrings that don't really fit in the in/out/inout split? so linear, steps, elastic. maybe with more example of the last two with different parameters if you want to fill it

rparrett:
Looks great, IMO
image

mockersf (approved):

mweatherley (approved):

PR #17743, opened 2025-02-08
Fix rounding in steps easing function

rparrett:
# Objective

While working on #17742, I noticed that the `Steps` easing function looked a bit suspicious.

image

Comparing to the options available in [css](https://developer.mozilla.org/en-US/docs/Web/CSS/easing-function/steps#description):

image

It is "off the charts," so probably not what users are expecting.

## Solution

Use `floor` when rounding to match the default behavior (jump-end, top right) in css.

image

## Testing

I had to modify an existing test that was testing against the old behavior. This function and test were introduced in #14788 and I didn't see any discussion about the rounding there.

`cargo run --example easing_functions`

## Migration Guide

`EaseFunction::Steps` now behaves like css's default, "jump-end." If you were relying on the old behavior, we plan on providing it. See https://github.com/bevyengine/bevy/issues/17744.

### DIFF
--- a/crates/bevy_math/src/curve/easing.rs
+++ b/crates/bevy_math/src/curve/easing.rs
@@ -685,7 +685,7 @@ mod easing_functions {
 
     #[inline]
     pub(crate) fn steps(num_steps: usize, t: f32) -> f32 {
-        ops::round(t * num_steps as f32) / num_steps.max(1) as f32
+        ops::floor(t * num_steps as f32) / num_steps.max(1) as f32
     }
 
     #[inline]
--- a/crates/bevy_math/src/curve/mod.rs
+++ b/crates/bevy_math/src/curve/mod.rs
@@ -1064,14 +1064,12 @@ mod tests {
         let curve = EasingCurve::new(start, end, EaseFunction::Steps(4));
         [
             (0.0, start),
-            (0.124, start),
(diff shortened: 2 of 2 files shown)


rparrett:
cc @RobWalt

RobWalt (approved):
LGTM, thanks!

I think it would be good to support all of the CSS modi of steps via configuration options somewhere. Can you open an issue and link to this PR please? I'll tackle it later on when I'm back home (on mobile atm)

Carter0 (commented):
Can you take a screenshot of what it looks like with your changes? 

It does look odd on close inspection. It starts like halfway through a step I think. And then I swear the vertical lines are slightly bent. Maybe I’m seeing things tho

rparrett:
> It starts like halfway through a step I think.

That's the appearance, but I would say "it advances half-way through the step."

> And then I swear the vertical lines are slightly bent. Maybe I’m seeing things tho

That's to be expected based on the sampling that's happening in the example.

> Can you take a screenshot of what it looks like with your changes

Sure

rparrett:
Looking into the failing test. It succeeds locally.

mweatherley (approved):

Carter0 (approved):

mockersf (approved):

Issue #17747, opened 2025-02-08
Change defaults with Local<> to be more ergonomic


newclarityex:
## What problem does this solve or what need does it fill?

I often use Local for systems that keep track of local ui input, timers, counters, etc. For data that only needs to be accessed in a single system, it's very convenient especially for prototyping to not have to manage resources or components. Using Local<> with a default value however is a lot more complex and makes Local<> much less appealing to use. The workaround described in the docs is to use a resource and convert your system into a capturing closure, which is more complex. 

## What solution would you like?
A possible solution would be a function on Local<> to set an intial value that is only in the first run of the system.

### Example:
```
fn counter(mut count: Local<i32>) {
    count.set_default(10);
    *count += 1;
}
```

## What alternative(s) have you considered?

An alternative workaround is to create a wrapping type, and impl default on that type, then Local<Wrapper>. This is still simpler than the closure solution, and makes more sense, but I feel this can be simplified further.

### Example:
```
struct CounterLocal(i32);

impl Default for CounterLocal {
    fn default() -> Self {
        CounterLocal(10)
    }
}

fn counter(mut count: Local<CounterLocal>) {
    count.0 += 1;
}
```

alice-i-cecile:
Can you use Option<T> as your Local type?

newclarityex:
Using Option seems like a decent alternative as well, although trying to write with it seems a little janky to mutate, although there might be a better way than what I'm doing here

```
fn counter(mut count: Local<Option<i32>>) {
    let cur_count = count.unwrap_or(10);
    *count = Some(cur_count);

    // do stuff with cur_count

    *count = Some(cur_count + 1);
}
```

newclarityex:
An additional edgecase are things that don't impl Default. They currently have to be wrapped to be used at all in Local.

ilyvion:
There's `Option::get_or_insert` which gives you a `&mut T` with the given default if `None`; so for your example, you'd do
```rust
fn counter(mut count: Local<Option<i32>>) {
    let cur_count = count.get_or_insert(10);

    // do stuff with cur_count

    *cur_count = cur_count + 1;
}
```

newclarityex:
That's perfect, I'll update my pr

theotherphil:
The linked PR fixed this issue - github just doesn't recognise "solves" as a linking keyword.

PR #17751, opened 2025-02-09
adds example for local defaults

newclarityex:
# Objective
Solves https://github.com/bevyengine/bevy/issues/17747.

## Solution

- Adds an example for creating a default value for Local.

## Testing

- Example code compiles and passes assertions.


alice-i-cecile (approved):
Perfect, that's lovely.

PR #17752, opened 2025-02-09
Add ways to configure `EasingFunction::Steps` via new `StepConfig`

### LINKED ISSUE #17744: Support more "step" easing modes
> I think it would be good to support all of the CSS modi of steps via configuration options somewhere. Can you open an issue and link to this PR please? I'll tackle it later on when I'm back home (on mobile atm)
> 
> _Originally posted by @RobWalt in https://github.com/bevyengine/bevy/pull/17743#pullrequestreview-2603874387_

See https://developer.mozilla.org/en-US/docs/Web/CSS/easing-function/steps#description.

After #17743, our `Step` easing will be like "jump-end" (the default) in CSS, but it would be nice to support other modes.

RobWalt:
# Objective

- In #17743, attention was raised to the fact that we supported an unusual kind of step easing function. The author of the fix kindly provided some links to standards used in CSS. It would be desirable to support generally agreed upon standards so this PR here tries to implement an extra configuration option of the step easing function
- Resolve #17744

## Solution

- Introduce `StepConfig`
- `StepConfig` can configure both the number of steps and the jumping behavior of the function
- `StepConfig` replaces the raw `usize` parameter of the `EasingFunction::Steps(usize)` construct.
- `StepConfig`s default jumping behavior is `end`, so in that way it follows #17743 

## Testing

- I added a new test per `JumpAt` jumping behavior. These tests replicate the visuals that can be found at https://developer.mozilla.org/en-US/docs/Web/CSS/easing-function/steps#description

## Migration Guide

- `EasingFunction::Steps` now uses a `StepConfig` instead of a raw `usize`. You can replicate the previous behavior by replaceing `EasingFunction::Steps(10)` with `EasingFunction::Steps(StepConfig::new(10))`.


alice-i-cecile (changes requested):
I think the `StepConfig` is a bit convoluted; can you swap to `EaseFunctionSteps(usize, JumpAt)` to make this a bit nicer to use? I think that's a good compromise between the initial design and the one proposed in #17777.

rparrett:
One thing I appreciated about the other PR was the consistency with regard to the SVGs, generating them with the same tool that generated the others.

The SVGs in this PR are more informative, but they are also very large, and the style doesn't seem to fit quite as well.

Although in 17777, the svgs did not actually depict the separate step modes properly.

<details>
<summary>Expand StepConfig docs</summary>

image
</details>

<details>
<summary>Expand 17777 Step_ docs</summary>

image
</details>

crates/bevy_math/src/curve/easing.rs
```diff
@@ -903,4 +985,71 @@ mod tests {
...
+    fn step_config_start() {
+        let config = JumpAt::Start;
+        let num_steps = 4;
+
+        [
+            (0.0, 0.25),
+            (0.25, 0.5),
+            (0.5, 0.75),
+            (0.75, 1.0),
+            (1.0, 1.0),
```
greeble-dev, line 999: Taking a leaf from #17777, could this include a few edge case checks?

```suggestion
            (0.0, 0.25),
            (0.24, 0.25),
            (0.25, 0.5),
            (0.49, 0.5),
            (0.5, 0.75),
            (0.74, 0.75),
            (0.75, 1.0),
            (1.0, 1.0),
```

crates/bevy_math/src/curve/easing.rs
```diff
@@ -903,4 +985,71 @@ mod tests {
             );
         });
     }
+
+    #[test]
+    fn step_config_start() {
```
mockersf, line 990: not sure those tests are really needed

examples/animation/easing_functions.rs
```diff
@@ -68,7 +68,7 @@ fn setup(mut commands: Commands) {
         EaseFunction::BounceInOut,
         // "Other" row
         EaseFunction::Linear,
-        EaseFunction::Steps(4),
+        EaseFunction::Steps(4, JumpAt::End),
```
mockersf, line 71:
```suggestion
        EaseFunction::Steps(4, JumpAt::End),
        EaseFunction::Steps(4, JumpAt::Start),
        EaseFunction::Steps(4, JumpAt::Both),
        EaseFunction::Steps(4, JumpAt::None),
```

alice-i-cecile (approved):

rparrett (approved):

crates/bevy_math/src/curve/easing.rs
```diff
@@ -1027,6 +1075,94 @@ mod tests {
...
+            (0.499, 0.4),
+            (0.5, 0.6),
+            (0.749, 0.6),
+            (0.75, 0.8),
+            (0.999, 0.8),
+            (1.0, 1.0),
+        ]
+        .into_iter()
+        .for_each(|(t, expected)| {
+            assert_abs_diff_eq!(jump_at.eval(num_steps, t), expected);
```
mockersf, line 1163:
```suggestion
        .for_each(|(t, expected)| {
            assert_abs_diff_eq!(jump_at.eval(num_steps, t), expected);
        });
```

mockersf (approved):

PR #17731, opened 2025-02-07
feat(ecs): implement fallible observer systems

JeanMertz:
This commit builds on top of the work done in #16589 and #17051, by adding support for fallible observer systems.

As with the previous work, the actual results of the observer system are suppressed for now, but the intention is to provide a way to handle errors in a global way.

Until then, you can use a `PipeSystem` to manually handle results.

PR #17753, opened 2025-02-09
feat(ecs): configurable error handling for fallible systems

JeanMertz:
You can now configure error handlers for fallible systems. These can be configured on several levels:

- Globally via `App::set_systems_error_handler`
- Per-schedule via `Schedule::set_error_handler`
- Per-system via a piped system (this is existing functionality)

The "fallible_systems" example demonstrates the new functionality.

This builds on top of #17731, #16589, #17051.

### DIFF
--- a/crates/bevy_ecs/src/schedule/schedule.rs
+++ b/crates/bevy_ecs/src/schedule/schedule.rs
@@ -29,7 +29,7 @@ use crate::{
     component::{ComponentId, Components, Tick},
     prelude::Component,
     resource::Resource,
-    result::Result,
+    result::{default_error_handler, Error},
     schedule::*,
     system::ScheduleSystem,
     world::World,
@@ -40,20 +40,30 @@ pub use stepping::Stepping;
 use Direction::{Incoming, Outgoing};
 
 /// Resource that stores [`Schedule`]s mapped to [`ScheduleLabel`]s excluding the current running [`Schedule`].
-#[derive(Default, Resource)]
+#[derive(Resource)]
 pub struct Schedules {
     inner: HashMap<InternedScheduleLabel, Schedule>,
     /// List of [`ComponentId`]s to ignore when reporting system order ambiguity conflicts
     pub ignored_scheduling_ambiguities: BTreeSet<ComponentId>,
+
+    /// Error handler to use for systems that return a [`Result`](crate::result::Result).
+    pub error_handler: fn(Error, &ScheduleSystem),
+}
+
+impl Default for Schedules {
+    fn default() -> Self {
+        Self {
+            inner: default(),
+            ignored_scheduling_ambiguities: default(),
+            error_handler: default_error_handler,
+        }
+    }
 }
 
 impl Schedules {
     /// Constructs an empty `Schedules` with zero initial capacity.
     pub fn new() -> Self {
-        Self {
-            inner: HashMap::default(),
-            ignored_scheduling_ambiguities: BTreeSet::new(),
-        }
+        Self::default()
     }
 
     /// Inserts a labeled schedule into the map.
@@ -300,6 +310,7 @@ pub struct Schedule {
     executable: SystemSchedule,
     executor: Box<dyn SystemExecutor>,
     executor_initialized: bool,
+    error_handler: Option<fn(Error, &ScheduleSystem)>,
 }
 
 #[derive(ScheduleLabel, Hash, PartialEq, Eq, Debug, Clone)]
@@ -324,6 +335,7 @@ impl Schedule {
             executable: SystemSchedule::new(),
             executor: make_executor(ExecutorKind::default()),
             executor_initialized: false,
+            error_handler: None,
         };
         // Call `set_build_settings` to add any default build passes
         this.set_build_settings(Default::default());
@@ -401,6 +413,11 @@ impl Schedule {
         self
     }
 
+    /// Set the error handler to use for systems that return a [`Result`](crate::result::Result).
+    pub fn set_error_handler(&mut self, error_handler: fn(Error, &ScheduleSystem)) {
+        self.error_handler = Some(error_handler);
+    }
+
     /// Returns the schedule's current `ScheduleBuildSettings`.
     pub fn get_build_settings(&self) -> ScheduleBuildSettings {
         self.graph.settings.clone()
@@ -438,8 +455,11 @@ impl Schedule {
         self.initialize(world)
             .unwrap_or_else(|e| panic!("Error when initializing schedule {:?}: {e}", self.label));
 
+        let error_handler = self.error_handler.expect("schedule initialized");
+
         #[cfg(not(feature = "bevy_debug_stepping"))]
-        self.executor.run(&mut self.executable, world, None);
+        self.executor
+            .run(&mut self.executable, world, None, error_handler);
 
         #[cfg(feature = "bevy_debug_stepping")]
         {
@@ -448,8 +468,12 @@ impl Schedule {
                 Some(mut stepping) => stepping.skipped_systems(self),
             };
 
-            self.executor
-                .run(&mut self.executable, world, skip_systems.as_ref());
+            self.executor.run(
+                &mut self.executable,
+                world,
+                skip_systems.as_ref(),
+                error_handler,
+            );
         }
     }
 
@@ -472,6 +496,10 @@ impl Schedule {
             )?;
             self.graph.changed = false;
             self.executor_initialized = false;
+        }
+
+        if self.error_handler.is_none() {
+            self.error_handler = Some(world.get_resource_or_init::<Schedules>().error_handler);
         }
 
         if !self.executor_initialized {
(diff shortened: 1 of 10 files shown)


NthTensor:
Wonderful, thanks! I will try very hard to find the time to review this sometime this week.

examples/ecs/fallible_systems.rs
```diff
@@ -77,3 +97,14 @@ fn setup(
...
     Ok(())
 }
+
+#[derive(Resource)]
+struct UninitializedResource;
+
+fn failing_system(world: &mut World) -> Result {
+    world
+        .get_resource::<UninitializedResource>()
+        .ok_or("Resource not initialized")?;
```
alice-i-cecile, line 107: A comment here about why putting a string in here works would be helpful.

examples/ecs/fallible_systems.rs
```diff
@@ -6,13 +6,33 @@ use bevy::prelude::*;
...
+        .set_systems_error_handler(|err, system| warn!("{:?} failed: {err}", system.name()))
+        // You can handle individual systems by piping the output result.
+        .add_systems(
+            PostStartup,
+            failing_system.pipe(|result: In<Result>| {
+                result.0.inspect_err(|err| info!("captured error: {err}"))
+            }),
+        );
+
+    // You can also set a custom error handler per `Schedule` (optionally at runtime through
```
alice-i-cecile, line 26: Phrasing here could be cleaned up. I think the parenthetical can just be removed TBH.

crates/bevy_ecs/src/world/mod.rs
```diff
@@ -3666,6 +3666,18 @@ impl World {
         schedules.allow_ambiguous_resource::<T>(self);
         self.insert_resource(schedules);
     }
+
+    /// Set the error handler to use for systems that return a [`Result`](crate::result::Result) in
+    /// a specific schedule.
+    pub fn set_schedule_error_handler(
```
alice-i-cecile, line 3672: This doesn't feel particularly worthwhile as a method on `World`. I don't think users will be trying to access this data via `World` very often.

crates/bevy_ecs/src/schedule/schedule.rs
```diff
@@ -40,20 +40,30 @@ pub use stepping::Stepping;
...
 
 /// Resource that stores [`Schedule`]s mapped to [`ScheduleLabel`]s excluding the current running [`Schedule`].
-#[derive(Default, Resource)]
+#[derive(Resource)]
 pub struct Schedules {
     inner: HashMap<InternedScheduleLabel, Schedule>,
     /// List of [`ComponentId`]s to ignore when reporting system order ambiguity conflicts
     pub ignored_scheduling_ambiguities: BTreeSet<ComponentId>,
+
+    /// Error handler to use for systems that return a [`Result`](crate::result::Result).
```
alice-i-cecile, line 49: This needs to explain the fallback / prioritization system.

examples/ecs/fallible_systems.rs
JeanMertz: Rust provides `impl<'a> From<&str> for Box<dyn Error + 'a>`.

See: https://doc.rust-lang.org/1.84.1/src/alloc/boxed/convert.rs.html#686

Does it make sense to add this as a comment?

alice-i-cecile (changes requested):
Basic strategy looks good to me. Exposing the different error handlers at the schedule level is definitely the most natural way to do this, even if I think that users will almost never set error handlers for specific schedules independently.

That said, I think we should:

1. Improve the documentation to more clearly explain the core model. The `bevy_ecs::result` module feels like the best place for a detailed explanation, which should be linked wherever these setters are exposed.
2. Cut the error handling setter from `World`: it's neither convenient nor natural, and it's super easy to implement externally.
3. Add a few conveniences to the `result` module to make setting this behavior to info/warn/error/ignore easier.

3 can easily happen in follow-up though :)

crates/bevy_ecs/src/world/mod.rs
JeanMertz: I mostly followed `allow_ambiguous_component`, which also weaves the same method through the call stack.

That one also seems highly unlikely to be used at runtime, more than this one, I guess?

Happy to remove it to keep the API of `World` a bit cleaner.

examples/ecs/fallible_systems.rs
alice-i-cecile: Yep! That's exactly the sort of information that's really helpful to a beginner but really hard to piece together.

crates/bevy_ecs/src/world/mod.rs
JeanMertz: Never mind, I just realized I refactored this at some point and this method is no longer used anywhere. I'll remove it.

examples/ecs/fallible_systems.rs
JeanMertz: Resolved in [95b54ba](https://github.com/bevyengine/bevy/pull/17753/commits/95b54bac8e63257562d3e5d6ef805455233e17bc).

examples/ecs/fallible_systems.rs
JeanMertz: Resolved in [95b54ba](https://github.com/bevyengine/bevy/pull/17753/commits/95b54bac8e63257562d3e5d6ef805455233e17bc).

crates/bevy_ecs/src/world/mod.rs
JeanMertz: Resolved in [95b54ba](https://github.com/bevyengine/bevy/pull/17753/commits/95b54bac8e63257562d3e5d6ef805455233e17bc).

JeanMertz:
I realized #17731 does not give us access to `ScheduleSystem` (because observer systems aren't "scheduled" in the ECS the way regular systems are).

So it probably makes sense to review and merge #17731 first, then update this PR to change the error handler signature from `fn(Error, &ScheduledSystem)` to (e.g.) `fn(Error, SystemErrorContext)`, which allows us to pass some context such as the system name for both scheduled systems and observer systems.

It also warrants a discussion on where our default `error_handler` should live. As I _can_ fetch the `Schedules` resource to get the default error handler for observer systems, but for obvious reasons that's a weird place to get the data from.

Given that we're already doing a `world.get_resource::<Schedules>` to get the global error handler, I'm thinking we could instead introduce a new `DefaultSystemsErrorHandler` resource that we pull instead, both for scheduled systems and observer systems. That resource would be fetched once when initializing a schedule, or when spawning an observer, so it's not in any significantly hot path.

alice-i-cecile:
> Given that we're already doing a world.get_resource::<Schedules> to get the global error handler, I'm thinking we could instead introduce a new DefaultSystemsErrorHandler resource that we pull instead, both for scheduled systems and observer systems. That resource would be fetched once when initializing a schedule, or when spawning an observer, so it's not in any significantly hot path.

I like this!

examples/ecs/fallible_systems.rs
```diff
@@ -6,13 +6,53 @@ use bevy::prelude::*;
...
+    // types of systems the same way, except for the error handling.
+    app.add_systems(Startup, (setup, failing_system));
+
+    // By default, fallible systems that return an error will panic.
+    //
+    // We can change this by setting a custom error handler. This can be done globally for all
+    // systems in a given `App`. Here we set the global error handler using one of the built-in
+    // error handlers. Bevy provides built-in handlers for `panic`, `error`, `warn`, `info`,
+    // `debug`, `trace` and `ignore`.
+    app.set_systems_error_handler(bevy::ecs::result::warn);
```
cart, line 24: I slightly prefer `app.set_system_error_handler` and `DefaultSystemErrorHandler` over the plural form, as error handler selection happens from the perspective of a single system.

cart (approved):
I think this is the right approach. Well done!

examples/ecs/fallible_systems.rs
JeanMertz: Resolved in [7ac274a](https://github.com/bevyengine/bevy/pull/17753/commits/7ac274a3edda041444296e84f87e1bfbe6d783aa).

notmd:
Nice work, just small question. How does Rust Analyzer work with this when suggesting autocomplete? Does it conflict with the info! macro etc from tracing? If it conflicts I think we should wrap them in a `SystemErrorHandler` struct. As this probably use one time so additional namespace is fine.

JeanMertz:
> Nice work, just small question. How does Rust Analyzer work with this when suggesting autocomplete? Does it conflict with the `warn!`, `error` macro etc from tracing? If it conflicts I think we should wrap them in a `SystemErrorHandler` struct. As this probably use one time so additional namespace is fine.

The bare functions aren’t exposed in the prelude, so it shouldn’t be an issue.

alice-i-cecile (approved):
Nice refinement!

alice-i-cecile:
Thank you to everyone involved with the authoring or reviewing of this PR! This work is relatively important and needs release notes! Head over to https://github.com/bevyengine/bevy-website/issues/1993 if you'd like to help out.

PR #17758, opened 2025-02-09
Trait tags on docs.rs

SpecificProtagonist:
# Objective

Bevy's ECS provides several core traits such as `Component`, `SystemParam`, etc that determine where a type can be used. When reading the docs, this currently requires scrolling down to and scanning the "Trait Implementations" section. Make these core traits more visible.

## Solution

Add a color-coded labels below the type heading denoting the following types:
- `Component`
  - immutable components are labeled as such
- `Resource`
- `Asset`
- `Event`
- `Plugin` & `PluginGroup`
- `ScheduleLabel` & `SystemSet`
- `SystemParam`

As docs.rs does not provide an option for post-processing the html, these are added via JS with traits implementations being detected by scanning the DOM. Rustdoc's html output is unstable, which could potentially lead to this detection (or the adding of the labels) to break, however it only needs to work when a new release is deployed and falls back to the status quo of not displaying these labels.

Idea by JMS55, implementation by Jondolf (see https://github.com/Jondolf/bevy_docs_extension_demo/).

## Testing

Run this in Bevy's root folder:
```bash
 RUSTDOCFLAGS="--html-after-content docs-rs/trait-tags.html --cfg docsrs_dep" RUSTFLAGS="--cfg docsrs_dep" cargo doc --no-deps -p <some_bevy_package>
```

---

## Showcase
Check it out on [docs.rs](https://docs.rs/bevy_docs_extension_demo/0.1.1/bevy_docs_extension_demo/struct.TestAllTraits.html)

trait tags

## Release Notes

On docs.rs, Bevy now displays labels indicating which core traits a type implements:
trait tags small

If you want to add these to your own crate, check out [these instructions](https://github.com/bevyengine/bevy/blob/main/docs-rs#3rd-party-crates).


alice-i-cecile:
I love the end result. I'll let others chime in on the impl though.

docs-rs/trait-tags.html
```diff
@@ -0,0 +1,165 @@
...
+        // Create a container for the tags.
+        const heading = document.body.querySelector(".main-heading h1");
+        const tagContainer = document.createElement('div');
+        tagContainer.className = 'bevy-tag-container';
+        heading.appendChild(tagContainer);
+
+        // Check if an implemented trait has a `type Mutability = Immutable` associated type.
+        // This is used to determine if a `Component` is immutable or not.
+        // TODO: Ideally we should just check the associated types of the `Component` trait,
+        //       but the docs.rs layout makes it tricky to do so in a robust wau.
```
Carter0, line 35:
```suggestion
        //       but the docs.rs layout makes it tricky to do so in a robust way.
```

BD103 (approved):
Changes look good! I'm not sure I like creating a new root-level folder, but there are no better alternatives right now. Thank you, this is really cool!

mockersf:
could you add the new flags to https://github.com/bevyengine/bevy/blob/main/.github/workflows/docs.yml#L64?

mockersf (approved):

alice-i-cecile:
Thank you to everyone involved with the authoring or reviewing of this PR! This work is relatively important and needs release notes! Head over to https://github.com/bevyengine/bevy-website/issues/1994 if you'd like to help out.

Issue #17746, opened 2025-02-08
Allow `Image` to be created without data


tychedelia:
## What problem does this solve or what need does it fill?

Currently, the `RenderAsset` implementation for `GpuImage` unconditionally calls `create_texture_with_data.` This is unhelpful for creating textures that are intended to be filled in on the GPU, for example during a compute pass that runs prior to rendering.

## What solution would you like?

`ShaderStorageBufffer` already has a similar solution where different constructors are provided with/without data. Do the same for `Image` and determine whether we need to create a texture with data based on the presence of data in `Image`.

Lege19:
this can probably be closed now

PR #17760, opened 2025-02-09
Allowed creating uninitialized images (for use as storage textures)

Lege19:
# Objective
https://github.com/bevyengine/bevy/issues/17746
## Solution
- Change `Image.data` from being a `Vec<u8>` to a `Option<Vec<u8>>`
- Added functions to help with creating images
## Testing

All current tests pass
Tested a variety of existing examples to make sure they don't crash (they don't)
Linux x86 64-bit NixOS 
---
## Migration Guide
Code that directly access `Image` data will now need to use unwrap or handle the case where no data is provided.
Behaviour of new_fill slightly changed, but not in a way that is likely to affect anything. It no longer panics and will fill the whole texture instead of leaving black pixels if the data provided is not a nice factor of the size of the image.

### DIFF
--- a/crates/bevy_image/src/dynamic_texture_atlas_builder.rs
+++ b/crates/bevy_image/src/dynamic_texture_atlas_builder.rs
@@ -2,6 +2,7 @@ use crate::{Image, TextureAtlasLayout, TextureFormatPixelInfo as _};
 use bevy_asset::RenderAssetUsages;
 use bevy_math::{URect, UVec2};
 use guillotiere::{size2, Allocation, AtlasAllocator};
+use tracing::error;
 
 /// Helper utility to update [`TextureAtlasLayout`] on the fly.
 ///
@@ -80,8 +81,15 @@ impl DynamicTextureAtlasBuilder {
             let end = begin + rect_width * format_size;
             let texture_begin = texture_y * rect_width * format_size;
             let texture_end = texture_begin + rect_width * format_size;
-            atlas_texture.data[begin..end]
-                .copy_from_slice(&texture.data[texture_begin..texture_end]);
+            let Some(ref mut atlas_data) = atlas_texture.data else {
+                error!("Atlas texture has no texture data");
+                return;
+            };
+            let Some(ref data) = texture.data else {
+                error!("Source texture provided has no texture data");
+                return;
+            };
+            atlas_data[begin..end].copy_from_slice(&data[texture_begin..texture_end]);
         }
     }
 }
--- a/crates/bevy_image/src/image.rs
+++ b/crates/bevy_image/src/image.rs
@@ -338,7 +338,11 @@ impl ImageFormat {
     reflect(opaque, Default, Debug)
 )]
 pub struct Image {
-    pub data: Vec<u8>,
+    /// Raw pixel data.
+    /// If the image is being used as a storage texture which doesn't need to be initialzied by the
+    /// CPU, then this should be `None`
+    /// Otherwise, it should always be `Some`
+    pub data: Option<Vec<u8>>,
     // TODO: this nesting makes accessing Image metadata verbose. Either flatten out descriptor or add accessors
     pub texture_descriptor: TextureDescriptor<Option<&'static str>, &'static [TextureFormat]>,
     /// The [`ImageSampler`] to use during rendering.
@@ -691,10 +695,61 @@ impl<'a> From<SamplerDescriptor<'a>> for ImageSamplerDescriptor {
 impl Default for Image {
     /// default is a 1x1x1 all '1.0' texture
     fn default() -> Self {
+        let mut image = Image::default_uninit();
+        image.data = Some(vec![255; image.texture_descriptor.format.pixel_size()]);
+        image
+    }
+}
+
+impl Image {
+    /// Creates a new image from raw binary data and the corresponding metadata.
+    ///
+    /// # Panics
+    /// Panics if the length of the `data`, volume of the `size` and the size of the `format`
+    /// do not match.
+    pub fn new(
+        size: Extent3d,
+        dimension: TextureDimension,
+        data: Vec<u8>,
+        format: TextureFormat,
+        asset_usage: RenderAssetUsages,
+    ) -> Self {
+        let mut image = Image::new_uninit(size, dimension, format, asset_usage);
+        debug_assert_eq!(
+            size.volume() * format.pixel_size(),
+            data.len(),
+            "Pixel data, size and format have to match",
+        );
+        image.data = Some(data);
+        image
+    }
+    /// Exactly the same as [`Image::new`], but doesn't initialize the image
+    pub fn new_uninit(
+        size: Extent3d,
+        dimension: TextureDimension,
+        format: TextureFormat,
+        asset_usage: RenderAssetUsages,
+    ) -> Self {
+        let mut image = Image::default_uninit();
+        image.texture_descriptor.dimension = dimension;
+        image.texture_descriptor.size = size;
+        image.texture_descriptor.format = format;
+        image.asset_usage = asset_usage;
+        image
+    }
+
+    /// A transparent white 1x1x1 image.
+    ///
+    /// Contrast to [`Image::default`], which is opaque.
+    pub fn transparent() -> Image {
+        // We rely on the default texture format being RGBA8UnormSrgb
+        // when constructing a transparent color from bytes.
+        // If this changes, this function will need to be updated.
         let format = TextureFormat::bevy_default();
-        let data = vec![255; format.pixel_size()];
+        debug_assert!(format.pixel_size() == 4);
+        let data = vec![255, 255, 255, 0];
         Image {
-            data,
+            data: Some(data),
             texture_descriptor: TextureDescriptor {
                 size: Extent3d {
                     width: 1,
@@ -714,49 +769,11 @@ impl Default for Image {
             asset_usage: RenderAssetUsages::default(),
         }
     }
-}
-
-impl Image {
-    /// Creates a new image from raw binary data and the corresponding metadata.
-    ///
-    /// # Panics
-    /// Panics if the length of the `data`, volume of the `size` and the size of the `format`
-    /// do not match.
-    pub fn new(
-        size: Extent3d,
-        dimension: TextureDimension,
-        data: Vec<u8>,
-        format: TextureFormat,
-        asset_usage: RenderAssetUsages,
-    ) -> Self {
-        debug_assert_eq!(
-            size.volume() * format.pixel_size(),
-            data.len(),
-            "Pixel data, size and format have to match",
-        );
-        let mut image = Self {
-            data,
-            ..Default::default()
-        };
-        image.texture_descriptor.dimension = dimension;
-        image.texture_descriptor.size = size;
-        image.texture_descriptor.format = format;
-        image.asset_usage = asset_usage;
-        image
-    }
-
-    /// A transparent white 1x1x1 image.
-    ///
-    /// Contrast to [`Image::default`], which is opaque.
-    pub fn transparent() -> Image {
-        // We rely on the default texture format being RGBA8UnormSrgb
-        // when constructing a transparent color from bytes.
-        // If this changes, this function will need to be updated.
+    /// Creates a new uninitialized 1x1x1 image
+    pub fn default_uninit() -> Image {
         let format = TextureFormat::bevy_default();
-        debug_assert!(format.pixel_size() == 4);
-        let data = vec![255, 255, 255, 0];
         Image {
-            data,
+            data: None,
             texture_descriptor: TextureDescriptor {
                 size: Extent3d {
                     width: 1,
@@ -779,9 +796,6 @@ impl Image {
 
     /// Creates a new image from raw binary data and the corresponding metadata, by filling
     /// the image data with the `pixel` data repeated multiple times.
-    ///
-    /// # Panics
-    /// Panics if the size of the `format` is not a multiple of the length of the `pixel` data.
     pub fn new_fill(
         size: Extent3d,
         dimension: TextureDimension,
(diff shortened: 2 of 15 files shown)


Lege19:
When I ran cargo check I didn't enable all features. :(

crates/bevy_image/src/dynamic_texture_atlas_builder.rs
```diff
@@ -80,8 +81,15 @@ impl DynamicTextureAtlasBuilder {
             let end = begin + rect_width * format_size;
             let texture_begin = texture_y * rect_width * format_size;
             let texture_end = texture_begin + rect_width * format_size;
-            atlas_texture.data[begin..end]
-                .copy_from_slice(&texture.data[texture_begin..texture_end]);
+            let Some(ref mut atlas_data) = atlas_texture.data else {
```
IceSentry, line 84: I think this function should be updated to return a `Result` instead of just logging a warning. add_texture should probably also be updated to return a `Result`. Also, this should probably be moved out of the loop.

crates/bevy_image/src/image.rs
```diff
@@ -729,15 +714,23 @@ impl Image {
...
             data.len(),
             "Pixel data, size and format have to match",
         );
-        let mut image = Self {
-            data,
-            ..Default::default()
-        };
+        image.data = Some(data);
+        image
+    }
```
IceSentry, line 725:
```suggestion
    }
    
```

crates/bevy_image/src/image.rs
```diff
@@ -756,7 +749,31 @@ impl Image {
...
+                usage: TextureUsages::TEXTURE_BINDING | TextureUsages::COPY_DST,
+                view_formats: &[],
+            },
+            sampler: ImageSampler::Default,
+            texture_view_descriptor: None,
+            asset_usage: RenderAssetUsages::default(),
+        }
+    }
+    /// Creates a new uninitialized 1x1x1 image
+    pub fn default_uninit() -> Image {
```
IceSentry, line 773: The implementation should use new_uninit(). It's also missing a new line just above the doc comment.

crates/bevy_image/src/image.rs
```diff
@@ -779,38 +796,23 @@ impl Image {
...
         pixel: &[u8],
         format: TextureFormat,
         asset_usage: RenderAssetUsages,
     ) -> Self {
-        let mut value = Image::default();
-        value.texture_descriptor.format = format;
-        value.texture_descriptor.dimension = dimension;
-        value.asset_usage = asset_usage;
-        value.resize(size);
+        let mut image = Image::default_uninit();
```
IceSentry, line 806: Why were the debug_asserts removed?

crates/bevy_image/src/image.rs
```diff
@@ -849,10 +851,12 @@ impl Image {
     /// Does not properly resize the contents of the image, but only its internal `data` buffer.
     pub fn resize(&mut self, size: Extent3d) {
         self.texture_descriptor.size = size;
-        self.data.resize(
-            size.volume() * self.texture_descriptor.format.pixel_size(),
-            0,
-        );
+        if let Some(ref mut data) = self.data {
```
IceSentry, line 854: I _think_ resizing an empty image should maybe log a warning? I'm not sure in what scenario that would be a valid thing to do? Or maybe resize should just create the image data if it doesn't exist already.

crates/bevy_image/src/texture_atlas_builder.rs
```diff
@@ -118,8 +118,15 @@ impl<'a> TextureAtlasBuilder<'a> {
             let end = begin + rect_width * format_size;
             let texture_begin = texture_y * rect_width * format_size;
             let texture_end = texture_begin + rect_width * format_size;
-            atlas_texture.data[begin..end]
-                .copy_from_slice(&texture.data[texture_begin..texture_end]);
+            let Some(ref mut atlas_data) = atlas_texture.data else {
```
IceSentry, line 121: This could be moved out of the loop and it should maybe be turned into an error instead of a log.

crates/bevy_pbr/src/render/mesh.rs
```diff
@@ -1732,7 +1732,7 @@ impl FromWorld for MeshPipeline {
             let format_size = image.texture_descriptor.format.pixel_size();
             render_queue.write_texture(
                 texture.as_image_copy(),
-                &image.data,
+                image.data.as_ref().unwrap(),
```
IceSentry, line 1735:
```suggestion
                image.data.as_ref().expect("Image was created with data"),
```

crates/bevy_image/src/image.rs
Lege19: They didn't make sense with the reimplementation, as there was no reason I could see why what they  were there to prevent shouldn't be allowed. Neither of them would have caused anything unexpected as far as I could see, so why make a function panic that has no reason to?

IceSentry (changes requested):
Overall this looks like a useful change, but I would prefer if you used expect() instead of unwrap in code that isn't inside a test. I didn't add a comment on every unwrap but a few of them should definitely be updated.

crates/bevy_image/src/image.rs
Lege19: well, new_uninit() currently uses default_uninit(), but it would probably make more sense to swap that around

crates/bevy_image/src/image.rs
Lege19: I thought about this, and I decided it was more likely to be done intentionally than by accident. And I feel quite strongly that it shouldn't create image data. There's no reason a user calling that function would expect that to happen. It should make the minimum modification to the `Image` necessary to do its job.
I don't really mind about the warning thing though, so I could add that.

crates/bevy_image/src/image.rs
IceSentry: The pixel.len() and format.pixel_size() should still be validated even with the new impl. It currently doesn't let users know if they used an invalid pixel size. The code will work, but users should still be warned because it was most likely a mistake to use an invalid pixel size.

crates/bevy_image/src/texture_atlas_builder.rs
Lege19: I'll move it outside the loop, although I'm not sure what you mean by an error instead of a log. It's an error already, just not a panic, but we shouldn't panic unless we have to

crates/bevy_image/src/image.rs
IceSentry: Oh, right, the code folding in the PR review UI confused me and made it look like it was the opposite.

IceSentry: Still, I think it should be the other way around. Like you said.

crates/bevy_pbr/src/render/mesh.rs
Lege19: If it's Bevy policy to use expect instead then sure, but the image is created just a few lines before, so the unwrap will always be valid.

crates/bevy_image/src/image.rs
IceSentry: I somewhat disagree that it is unexpected because for a long time people used `.resize()` instead of `new_fill()` to fill an image with data.

I think that was a wrong way to use that api though, so I'm fine with not creating data if it doesn't exist, but that should log a warning because I don't think it makes sense in most case to call .resize() on an uninit image.

crates/bevy_render/src/texture/fallback_image.rs
```diff
@@ -114,7 +113,7 @@ fn fallback_image_new(
             render_queue,
             &image.texture_descriptor,
             TextureDataOrder::default(),
-            &image.data,
+            &image.data.unwrap(),
```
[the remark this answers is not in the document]
Lege19: I don't see it, elaborate?

crates/bevy_image/src/texture_atlas_builder.rs
IceSentry: Error as in, return a `Result::Err` not just a log.

crates/bevy_pbr/src/render/mesh.rs
IceSentry: It's not a strict policy, but this kind of thing is to avoid future errors if someone ever updated the code above to something that doesn't initialize the data. I will approve with or without the expect, but I prefer it if it's there.

crates/bevy_image/src/image.rs
Lege19: what do you mean by invalid pixel size though?

Lege19: isn't it useful to be able to specify `&[0xff]` to mean all white and full opacity?

crates/bevy_render/src/texture/fallback_image.rs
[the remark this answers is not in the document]
IceSentry: on line 100, you changed default() to default_uninit(). Unless I'm missing something, this means the data will be uninitialized once it reaches this unwrap.

crates/bevy_image/src/image.rs
IceSentry: I mean, for that specific use case, sure, but if someone for some reasons gives it 3 values for a format that expects 4 values it won't generate something that makes sense. Especially if it's not just 0xff. It's trivial to check for this and for users that want to fill it with the same byte they can just do `&[0xff; 4]` or whatever the size they need.

crates/bevy_image/src/image.rs
Lege19: I've just pushed a commit to rectify this

crates/bevy_image/src/texture_atlas_builder.rs
Lege19: should I also keep the error! ?

Lege19: this requires me to add another variant to TextureAtlasBuilderError. Which isn't currently non_exhaustive, should I make it non_exhaustive?

Lege19: more generally, is it Bevy policy to always make error enums non_exhaustive

crates/bevy_render/src/texture/fallback_image.rs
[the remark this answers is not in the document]
Lege19: line 100 only runs if !create_texture_with data, line 116 only runs if create_texture_with_data. I'm aware it's not ideal to have code that takes this much analysis to validate, but I'm pretty sure it is valid

crates/bevy_image/src/dynamic_texture_atlas_builder.rs
Lege19: done

crates/bevy_image/src/image.rs
Lege19: done

crates/bevy_pbr/src/render/mesh.rs
Lege19: changed this for engine code, but not examples or tests

crates/bevy_image/src/image.rs
Lege19: just put this back. CI is waiting on miri, but I'm sure it will pass

crates/bevy_render/src/texture/fallback_image.rs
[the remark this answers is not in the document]
IceSentry: oh, yeah, I completely missed that if on line 111.

crates/bevy_image/src/dynamic_texture_atlas_builder.rs
```diff
@@ -67,22 +81,30 @@ impl DynamicTextureAtlasBuilder {
...
         let atlas_width = atlas_texture.width() as usize;
         let rect_width = rect.width() as usize;
         let format_size = atlas_texture.texture_descriptor.format.pixel_size();
 
+        let Some(ref mut atlas_data) = atlas_texture.data else {
+            error!("Atlas texture has no texture data");
+            return Err(DynamicTextureAtlasBuilderError::UninitializedAtlas);
+        };
+        let Some(ref data) = texture.data else {
+            error!("Source texture provided has no texture data");
```
IceSentry, line 97: I think we should just let the error bubble up to the user and let them decide if they want to log it.

IceSentry (approved):
I think we should just return `Err` and let users decide if they want to log. Other than that LGTM

alice-i-cecile (approved):

alice-i-cecile:
Nice error handling, and a well-motivated little fix. Thanks.

crates/bevy_image/src/texture_atlas_builder.rs
alice-i-cecile: Given the prevalence of breaking changes at this stage, we don't particularly care about non_exhaustive.

And in some cases, like the ECS, we can pretty reliably predict that error enums *are* exhaustive.

mintlu8:
Please revert this change. Uninitialized images can be expressed as a zero sized Vec. This change hurts the usability of `Image` a lot. You cannot use 
```rust
let Some(data) = &mut data else {return;}
```
in a lot of situations since `Image` has a lot of methods that cannot be split borrowed for now. Basically the only thing you can do that has any form of ergonomics is `unwrap`, which is bad.

Lege19:
There is certainly a valid argument with the unwrap thing. But using an empty Vec to represent an uninitialised Image is just a really bad idea. It's functionally equivalent to an Option but utterly fails to express the intent and is very arguably not 'correct'. I don't consider it an option. There definitely needs to be a way to have uninitialised images, and there is significant gain from unifying all image related things with this one type. Most users don't need to read image data, and writing image data isn't a problem.

This has already passed code review and been merged so I'm not sure what the protocol is for reverting something. But any case I don't have the power to do that.

Lege19:
Also doesn't
```rust

let Some(ref mut data) = &mut data else {return;}

```
work just fine?

mintlu8:
> There definitely needs to be a way to have uninitialized images.

I agree.

> Most users don't need to access the data directly and it makes sense to use Option for uninitialized images.

After some thought, I think the real solution is we should add some methods to edit the image, then we can sort of side step this issue.

```rust
image.edit_region::<Srgba>(rect, |x, y, rgba: &mut Rgba| {

})
```

PR #17764, opened 2025-02-10
Build batches across phases in parallel.

pcwalton:
Currently, invocations of `batch_and_prepare_binned_render_phase` and `batch_and_prepare_sorted_render_phase` can't run in parallel because they write to scene-global GPU buffers. After PR #17698, `batch_and_prepare_binned_render_phase` started accounting for the lion's share of the CPU time, causing us to be strongly CPU bound on scenes like Caldera when occlusion culling was on (because of the overhead of batching for the Z-prepass). Although I eventually plan to optimize `batch_and_prepare_binned_render_phase`, we can obtain significant wins now by parallelizing that system across phases.

This commit splits all GPU buffers that
`batch_and_prepare_binned_render_phase` and
`batch_and_prepare_sorted_render_phase` touches into separate buffers for each phase so that the scheduler will run those phases in parallel. At the end of batch preparation, we gather the render phases up into a single resource with a new *collection* phase. Because we already run mesh preprocessing separately for each phase in order to make occlusion culling work, this is actually a cleaner separation. For example, mesh output indices (the unique ID that identifies each mesh instance on GPU) are now guaranteed to be sequential starting from 0, which will simplify the forthcoming work to remove them in favor of the compute dispatch ID.

On Caldera, this brings the frame time down to approximately 9.1 ms with occlusion culling on.

Screenshot 2025-02-08 210720


examples/3d/occlusion_culling.rs
```diff
@@ -185,6 +190,10 @@ fn main() {
                 .set(RenderPlugin {
                     allow_copies_from_indirect_parameters: true,
                     ..default()
+                })
+                .set(PbrPlugin {
+                    allow_copies_from_indirect_parameters: true,
```
superdump, line 195: Could this be reused from the RenderPlugin? Rather than having to set it in two places.

pcwalton: Could you elaborate as to how that would work? The problem is that the `PbrPlugin` can't reach into the `RenderPlugin` to check its value.

tychedelia: I think this is fine for now but it's a rather particular option and I might suggest subsuming it some kind of broader "debug renderer" setting if we ever have a second instance of this kind of thing.

pcwalton: Yeah, good point. Or maybe a "debug flags"?

pcwalton: I went ahead and switched this to a `RenderDebugFlags` so that we can have more of them.

crates/bevy_render/src/batching/gpu_preprocessing.rs
```diff
@@ -702,6 +758,68 @@ pub struct IndirectBatchSet {
 /// These buffers will remain empty if indirect drawing isn't in use.
 #[derive(Resource)]
 pub struct IndirectParametersBuffers {
```
tychedelia, line 760: Total aesthetic preference but I would `Deref` these.

tychedelia (approved):
Lgtm. Thanks for creating the debug flags, a lot cleaner. Being able to lean on the ECS scheduler for this is nice and clean.

crates/bevy_render/src/batching/gpu_preprocessing.rs
pcwalton: Done.

superdump (approved):

PR #17765, opened 2025-02-10
Meshlet texture atomics

JMS55:
* Use texture atomics rather than buffer atomics for the visbuffer (haven't tested perf on a raster-heavy scene yet)
  * Unfortunately to clear the visbuffer we now need a compute pass to clear it. Using wgpu's clear_texture function internally uses a buffer -> image copy that's insanely expensive. Ideally it should be using vkCmdClearColorImage, which I've opened an issue for https://github.com/gfx-rs/wgpu/issues/7090. For now we'll have to stick with a custom compute pass and all the extra code that brings.
* Faster resolve depth pass by discarding 0 depth pixels instead of redundantly writing zero (2x faster for big depth textures like shadow views)


crates/bevy_pbr/src/meshlet/visibility_buffer_raster_node.rs
```diff
@@ -362,6 +374,29 @@ fn fill_cluster_buffers_pass(
     );
 }
 
+// TODO: Replace this with vkCmdClearColorImage once wgpu supports it
```
tychedelia, line 377: Bummer this is necessary

tychedelia (approved):
Everything reads quite a bit clearer with the storage texture. Tested on M2 macOS.

SparkyPotato (approved):
Looks good! Also seems to lead to a ~40% perf improvement for the main and shadow raster pass in the example.

alice-i-cecile:
Failed in CI on merge: https://github.com/bevyengine/bevy/actions/runs/13274053444/job/37059925752

JMS55:
@alice-i-cecile wgpu's DirectX12 backend does not work with push constants correctly. They need to implement something to wrap all push constants in structs, because HLSL does not allow push constants to be a primitive type. I'd rather not modify our shaders to use only structs for push constants.

https://github.com/gfx-rs/wgpu/issues/5683

JMS55:
@alice-i-cecile I've reverted back to a struct for downsample_depth's push constants, which should fix non-meshlet examples on DX12 (I still don't care enough to modify all of the meshlet code to use structs). Should be good to merge now.

alice-i-cecile:
Thanks :) Sorry that platform support is such a pain in the ass for rendering!

JMS55:
Summarizing impact of this PR for when I need to write my next blog post:
* Texture vs buffer atomics impact for the actual raster currently unmeasured (need to do that)
* Faster depth downsample for the mip 0 section of the code because sampling a texture is more cache-friendly
* Faster clear visbuffer texture via compute shader instead of wgpu's builtin vkCmdFillBuffer to zero a buffer
* Faster depth/material depth resolve by discarding background (0 depth) fragments instead of writing 0 (2x faster for shadow views)

ariofrio:
For reference, it seems that neither `TEXTURE_INT64_ATOMIC` nor `SHADER_INT64_ATOMIC_MIN_MAX` are supported on M1 Macs (2020).

Verified that `meshlets` panics on startup because of this on my M1 Pro (even after I revert this PR) and also cross-referenced with [Apple's Metal Feature Set Tables](https://developer.apple.com/metal/Metal-Feature-Set-Tables.pdf) (search for "64-bit atomics").

According to those tables, it seems that `TEXTURE_INT64_ATOMIC` and `SHADER_INT64_ATOMIC_MIN_MAX` is only supported on M2 Macs (2022) or newer. On mobile, it is supported on [A17](https://en.wikipedia.org/wiki/Apple_A17) and above, which ship with iPhone 15 Pro (2023) or newer and iPad Mini (7th generation, 2024) or newer.

JMS55:
Yes, you need an M2 GPU or newer.

PR #17767, opened 2025-02-10
Use unchecked shaders for better performance

JMS55:
# Objective
- Wgpu has some expensive code it injects into shaders to avoid the possibility of things like infinite loops. Generally our shaders are written by users who won't do this, so it just makes our shaders perform worse.

## Solution

- Turn off the checks.
  - We could try to conditionally keep them, but that complicates the code and 99.9% of users won't want this.

## Migration Guide

-  Bevy no longer turns on wgpu's runtime safety checks https://docs.rs/wgpu/latest/wgpu/struct.ShaderRuntimeChecks.html. If you were using Bevy with untrusted shaders, please file an issue.

### DIFF
--- a/crates/bevy_render/src/renderer/render_device.rs
+++ b/crates/bevy_render/src/renderer/render_device.rs
@@ -76,6 +76,11 @@ impl RenderDevice {
         }
 
         #[cfg(not(feature = "spirv_shader_passthrough"))]
+        // SAFETY: we are interfacing with shader code, which may contain undefined behavior,
+        // such as indexing out of bounds.
+        // The checks required are prohibitively expensive and a poor default for game engines.
+        // TODO: split this method into safe and unsafe variants, and propagate the safety requirements from
+        // https://docs.rs/wgpu/latest/wgpu/struct.Device.html#method.create_shader_module_trusted to the unsafe form.
         unsafe {
             self.device
                 .create_shader_module_trusted(desc, wgpu::ShaderRuntimeChecks::unchecked())
--- a/crates/bevy_render/src/renderer/render_device.rs
+++ b/crates/bevy_render/src/renderer/render_device.rs
@@ -64,11 +64,17 @@ impl RenderDevice {
                         })
                 }
             }
-            _ => self.device.create_shader_module(desc),
+            _ => unsafe {
+                self.device
+                    .create_shader_module_trusted(desc, wgpu::ShaderRuntimeChecks::unchecked())
+            },
         }
 
         #[cfg(not(feature = "spirv_shader_passthrough"))]
-        self.device.create_shader_module(desc)
+        unsafe {
+            self.device
+                .create_shader_module_trusted(desc, wgpu::ShaderRuntimeChecks::unchecked())
+        }
     }
 
     /// Check for resource cleanups and mapping callbacks.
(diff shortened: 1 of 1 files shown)


crates/bevy_render/src/renderer/render_device.rs
```diff
@@ -64,11 +64,17 @@ impl RenderDevice {
                         })
                 }
             }
-            _ => self.device.create_shader_module(desc),
+            _ => unsafe {
```
alice-i-cecile, line 67: SAFETY comment here please! Or at least a comment explaining why we're disabling these checks.

alice-i-cecile:
I do think we should do this though. This is the right default for 99% of users. Unless you're making VRChat or a shader playground, you're not running random untrusted shaders.

I'd like to expose an option to enable validation for some shaders, but that needs a real user to inform us of the right design.

cart (approved):
I agree that this is unquestionably the better default. I won't push too hard for it in this PR (as this is better and I don't look a gift horse in the mouth), but making this configurable from our Shader type is low hanging fruit, and makes a very valuable and hard-to-implement-yourself feature available to people that want it.

Edit: I don't think doing that needs design work. Just make it an enum on Shader and expose the functionality via a new method on RenderDevice.

alice-i-cecile (approved):
Eh, let's merge this now :)

crates/bevy_render/src/renderer/render_device.rs
alice-i-cecile:
```suggestion
            // SAFETY: we are interfacing with shader code, which may contain undefined behavior,
            // such as indexing out of bounds.
            // The checks required are prohibitively expensive and a poor default for game engines.
            // TODO: split this method into safe and unsafe variants, and propagate the safety requirements from
            // https://docs.rs/wgpu/latest/wgpu/struct.Device.html#method.create_shader_module_trusted to the unsafe form.
            _ => unsafe {
```

alice-i-cecile:
@JMS55 CI is failing due to no safety comment (probably good!). I've taken my best crack at it, but please feel free to revise it as you see fit.

JMS55:
@alice-i-cecile there's two usages of unsafe, you'll probably need to add the same comment to the second one to satisfy CI

crates/bevy_render/src/renderer/render_device.rs
```diff
@@ -64,11 +64,22 @@ impl RenderDevice {
...
+            // https://docs.rs/wgpu/latest/wgpu/struct.Device.html#method.create_shader_module_trusted to the unsafe form.
+            _ => unsafe {
+                self.device
+                    .create_shader_module_trusted(desc, wgpu::ShaderRuntimeChecks::unchecked())
+            },
         }
 
         #[cfg(not(feature = "spirv_shader_passthrough"))]
-        self.device.create_shader_module(desc)
+        unsafe {
```
alice-i-cecile, line 79:
```suggestion
            // SAFETY: we are interfacing with shader code, which may contain undefined behavior,
            // such as indexing out of bounds.
            // The checks required are prohibitively expensive and a poor default for game engines.
            // TODO: split this method into safe and unsafe variants, and propagate the safety requirements from
            // https://docs.rs/wgpu/latest/wgpu/struct.Device.html#method.create_shader_module_trusted to the unsafe form.
        unsafe {
```

Issue #10949, opened 2023-12-12
Scene prefab


djeedai:
## What problem does this solve or what need does it fill?

Ability to use `Scene` as a prefab, in the sense given by Unity: one-way irreversible spawning of a group of entities and components.

## What solution would you like?

No `SceneBundle`, no `SceneInstance`, no `SceenSpawner`. Just a way to spawn the content of a scene anywhere (including at the `World` root, without a common parent entity, so ability to spawn multiple top-level entities as a flat list) and forget forever that this ever was a `Scene`.

## What alternative(s) have you considered?

- `SceneBundle`. This has all the mentioned extra "features" which are just getting in the way most of the time for typical "use a prefab to spawn multiple things" usage. This is also missing some features like spawning without a single comment parent entity.
- Spawning function. This only works if you hardcode the collection of entities and components, and is hard to work with (spawning via code is tedious).

## Additional context

Possibly related to #3877, #10907

Issue #2565, opened 2021-07-29
Add an abstraction like bundles that can define child entities as well


ricardoalcantara:
## What problem does this solve or what need does it fill?

It's just a syntax sugar for better entity code organization

## What solution would you like?

```rust
#[derive(Bundle)]
pub struct PlayerBundle {
    player: Player,
    movement: Movement,
    transform: Transform,
    global_transform: GlobalTransform,
    
    #[child]
    sprite: SpriteSheetBundle,
}
```
## What alternative(s) have you considered?

I can already spawn two entities and parent them together, but I would be easier to spawn an entity and alto spawn it's children.

blaind:
I think some solution to the bundle with children -creation should be present. Ideally, it'd reduce the boilerplate code needed, and make the code cleaner.

However, some questions related to the proposed solution:
* how would nested bundles / bundle trees be handled?
* how would it handle cases where code needs to be executed while the bundle is being created?

I experimented with this solution a few days ago:
```rust
use bevy::{ecs::system::EntityCommands, prelude::*};
        
pub trait EntityBuilder {
    type Bundle: bevy::prelude::Bundle;
    
    fn new() -> Self;
        
    fn spawn<'a, 'b>(
        &self,
        cmd: &'b mut Commands<'a>,
        meshes: &mut Assets<Mesh>,
        materials: &mut Assets<StandardMaterial>,
    ) -> EntityCommands<'a, 'b> {
        let mut entity_cmd = cmd.spawn_bundle(self.build_bundle());
        entity_cmd.with_children(self.build_child_fn(meshes, materials));
        entity_cmd
    }

    fn build_bundle(&self) -> Self::Bundle;

    fn build_child_fn<'a>(
        &self,
        meshes: &mut Assets<Mesh>,
        materials: &mut Assets<StandardMaterial>,
    ) -> Box<dyn FnOnce(&mut ChildBuilder)>;
```
usage:

```rust
    let entity_id = MyBundleBuilder::new()
        .set_transform(Transform::from_xyz(2., 3., 4.)) // the struct where `EntityBuilder` is implemented for can have additional methods
        .spawn(&mut commands, &mut meshes, &mut materials)
        .id();
```

and then the struct (e.g. `PlayerStructBuilder` in above example) would implement the code for adding necessary children in the `build_child_fn` method.

Currently meshes & materials are hardcoded, that'll cause problems if other references are needed.

Davier:
[...]

Issue #1446, opened 2021-02-14
Workflow for working with prefab entities


alice-i-cecile:
**What problem does this solve or what need does it fill?**

Prefabs allow users to rapidly add copies of predefined entities to their game, and create variations on them.

**What solution would you like?**

The basic workflow would be:
1. Create some interesting prefab entities in their own `World` either using its own systems, directly from code or by reading in a serialized format.
2. Turn an entity into its component bundle.
3. Create a new entity from that component bundle, with variations as needed.
4. Use addition-detecting systems (or maybe `State` on_enter systems?) to finish initializing the entity in the main game world.

There are a few things we'd need to have this work:
1. Multiple, interacting `Worlds`.
2. A way to turn a created entity back into a bundle. 
3. A nice pattern for cloning an entity that allows you to change some of its components. If you have a bundle from 2., you could probably get away with `.spawn(cloned_entity_bundle).with(modified_component)`, but it would need to be clear that `.with` is overwriting.

**What alternative(s) have you considered?**

Use component bundles with the struct-update Default pattern used in e.g. `SpriteBundle`.

This works quite well for statically defined prefabs, which can each get their own type. I'm not sure it works correctly (or naturally) for prefabs that are created dynamically or loaded from a serialized format. You would need to create new Bundles at runtime, which I'm not sure you could do.

The functionality needed all seems useful for other purposes though, and the pattern seems quite natural, so it may still be worthing pushing users towards it even if those barriers can be overcome.

**Additional context**

This issue ties into discussion in #255, although it doesn't actually use `Scenes` to accomplish its goals.

This issue was prompted by [this](https://discord.com/channels/691052431525675048/742884593551802431/810349051115077652)  conversation on Discord. 

Multiple interacting `Worlds` was mentioned in #1343 for decoupling graphics and physics.

alice-i-cecile:
In order to turn a created entity into a bundle again, we can borrow from the approach taken in [`DynamicScene`](https://github.com/bevyengine/bevy/blob/a880b5450829411b0248e47cc066915234371d86/crates/bevy_scene/src/dynamic_scene.rs#L30)

TheRawMeatball:
Something to consider is how this would interact with #1442.

Davier:
I believe that scenes are the intended solution for that, what feature are they missing?

DarkMara:
> I believe that scenes are the intended solution for that, what feature are they missing?

[...]

PR #17769, opened 2025-02-10
Add an example demonstrating how to use default query filtering and entity cloning to create a prefab system

alice-i-cecile:
# Objective

Bevy users have long wished for tools for working with a "prefab"-style workflows: #10949, #3877, #2565, #1446...

The core idea is pretty simple: define an "object" (made up of some collection of entities) that we can save, instantiated, into the world, and then clone and modify them as needed.

## Solution

Instead of creating a first-class solution, I've instead chosen to add an example demonstrating how this pattern can be achieved quite simply with a combination of default query filters and recursive entity cloning.

This is useful for teaching users how to create their own prefab-flavored solution, and gives us time to iterate on the ideal solution.

Long-term, I think that Bevy would be well-served by making a first party prefab workflow, complete with a standardized state-powered asset loading solution. It's nice to be able to teach this pattern directly to new users, and more importantly, it allows us to view and building tooling around prefabs for e.g. a GUI-based level editor.

But Bevy itself is a poor place to prove out the architectural details for this: we aren't actually testing it in a large game project, and so the solution isn't subject to the appropriate pressures. Once the community has developed a robust solution, we should upstream it (likely as a standalone crate) and update this example!

## To do

This PR is not yet complete!

Before it's merged, we need to:

- [ ] merge https://github.com/bevyengine/bevy/pull/17768
- [ ] figure out why foxes flicker in and out of existence when spawning

It might also be nice to:

- [ ] add instruction text
- [ ] load more than one model, and actually make use of that hash map
- [ ] define and use an asset loading state (please discuss!)
- [ ] somehow extract the asset path from the SceneInstanceReady event and cache that

Like always, feel free to PR to my branch if you want to fix these.

## Testing

`cargo run --example prefabs`

## Showcase

image

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -320,6 +320,7 @@ Example | Description
 [Observers](../examples/ecs/observers.rs) | Demonstrates observers that react to events (both built-in life-cycle events and custom events)
 [One Shot Systems](../examples/ecs/one_shot_systems.rs) | Shows how to flexibly run systems without scheduling them
 [Parallel Query](../examples/ecs/parallel_query.rs) | Illustrates parallel queries with `ParallelIterator`
+[Prefabs](../examples/ecs/prefabs.rs) | Demonstrates how to load, store and clone template entity collections
 [Relationships](../examples/ecs/relationships.rs) | Define and work with custom relationships between entities
 [Removal Detection](../examples/ecs/removal_detection.rs) | Query for entities that had a specific component removed earlier in the current frame
 [Run Conditions](../examples/ecs/run_conditions.rs) | Run systems only when one or multiple conditions are met
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2085,6 +2085,17 @@ category = "ECS (Entity Component System)"
 wasm = false
 
 [[example]]
+name = "prefabs"
+path = "examples/ecs/prefabs.rs"
+doc-scrape-examples = true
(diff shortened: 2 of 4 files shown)


ChristopherBiscardi:
video that shows the flickering (and longer non-rendering meshes)

https://github.com/user-attachments/assets/42cb69cd-9126-4fd7-a69b-7e3a2932f70f

examples/README.md
```diff
@@ -320,6 +320,7 @@ Example | Description
 [Observers](../examples/ecs/observers.rs) | Demonstrates observers that react to events (both built-in life-cycle events and custom events)
 [One Shot Systems](../examples/ecs/one_shot_systems.rs) | Shows how to flexibly run systems without scheduling them
 [Parallel Query](../examples/ecs/parallel_query.rs) | Illustrates parallel queries with `ParallelIterator`
+[Prefabs](../examples/ecs/prefabs.rs) | Demonstrates how to load, store and clone template entity collections
```
cart, line 323: I think this deserves a qualifier or a different more specific name less loaded with less context. People will see this name and think "this is the Bevy prefab system", which has implications like "an asset format", "inheritance", etc.

alice-i-cecile: Good call; I definitely think that's the right call for the current stage of maturity.

PR #17771, opened 2025-02-10
Assert a minimum thread count for `TaskPoolOptions`

jmskov:
# Objective

 TaskPool policies assign one thread per group, so three threads are assigned at minimum. Adds an assertion for a minimum thread count for `TaskPoolOptions`. Passing 0, 1 or 2 for thread count has no real effect.

## Testing

- Module tests pass

### DIFF
--- a/crates/bevy_app/src/task_pool_plugin.rs
+++ b/crates/bevy_app/src/task_pool_plugin.rs
@@ -159,6 +159,7 @@ impl Default for TaskPoolOptions {
 impl TaskPoolOptions {
     /// Create a configuration that forces using the given number of threads.
     pub fn with_num_threads(thread_count: usize) -> Self {
+        assert!(thread_count > 2);
         TaskPoolOptions {
             min_total_threads: thread_count,
             max_total_threads: thread_count,
--- a/crates/bevy_app/src/task_pool_plugin.rs
+++ b/crates/bevy_app/src/task_pool_plugin.rs
@@ -159,7 +159,7 @@ impl Default for TaskPoolOptions {
 impl TaskPoolOptions {
     /// Create a configuration that forces using the given number of threads.
     pub fn with_num_threads(thread_count: usize) -> Self {
-        assert!(thread_count > 2);
+        assert!(thread_count > 2, "Default task pool policies requires at least 3 threads");
         TaskPoolOptions {
             min_total_threads: thread_count,
             max_total_threads: thread_count,
--- a/crates/bevy_app/src/task_pool_plugin.rs
+++ b/crates/bevy_app/src/task_pool_plugin.rs
@@ -12,7 +12,7 @@ use alloc::string::ToString;
 use bevy_platform_support::sync::Arc;
 use bevy_tasks::{AsyncComputeTaskPool, ComputeTaskPool, IoTaskPool, TaskPoolBuilder};
 use core::{fmt::Debug, marker::PhantomData};
-use log::trace;
+use log::{trace, warn};
 
 #[cfg(not(target_arch = "wasm32"))]
 use {crate::Last, bevy_ecs::prelude::NonSend};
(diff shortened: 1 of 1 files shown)


crates/bevy_app/src/task_pool_plugin.rs
```diff
@@ -159,6 +159,7 @@ impl Default for TaskPoolOptions {
 impl TaskPoolOptions {
     /// Create a configuration that forces using the given number of threads.
     pub fn with_num_threads(thread_count: usize) -> Self {
+        assert!(thread_count > 2);
```
alice-i-cecile, line 162: This needs a comment about why it's being set.

crates/bevy_app/src/task_pool_plugin.rs
```diff
@@ -159,6 +159,10 @@ impl Default for TaskPoolOptions {
 impl TaskPoolOptions {
     /// Create a configuration that forces using the given number of threads.
     pub fn with_num_threads(thread_count: usize) -> Self {
+        assert!(
+            thread_count > 2,
+            "Default task pool policies require at least 3 threads"
```
alice-i-cecile, line 164: Good, but we should add a bit more explanation here. Your PR description was really helpful to me, and users should learn those things too!

jmskov: Sounds good, I'll improve this message :)

alice-i-cecile (approved):

crates/bevy_app/src/task_pool_plugin.rs
```diff
@@ -159,6 +159,10 @@ impl Default for TaskPoolOptions {
 impl TaskPoolOptions {
     /// Create a configuration that forces using the given number of threads.
```
alice-i-cecile, line 160:
```suggestion
    /// Create a configuration that forces using the given number of threads.
    ///
    /// # Panics
    ///
    /// Values below 3 are meaningless, and will panic.
```

alice-i-cecile (approved):

mockersf:
why an `assert` rather than
- an `assert_dbg`?
- returning an `Option`/`Result`?
- ignoring values below 3, and using 3 by default in those cases?

jmskov:
> why an `assert` rather than
> 
> * an `assert_dbg`?
> * returning an `Option`/`Result`?
> * ignoring values below 3, and using 3 by default in those cases?

I opened this after using the `with_num_threads` method at [the plugin setup stage](https://docs.rs/bevy/latest/bevy/core/struct.TaskPoolOptions.html#scraped-examples). I was confused why 1, 2 (and 0) all worked the same (effectively defaulted to 3) until I looked at the source. Since this method is exposed all the time, `assert` was the simplest way to avoid this ambiguity for others.

Not sure if `debug_assert` is more or less appropriate. 😄

mockersf:
> Edit: Or a warning, just so the user knows what is happening.

I would prefer the warning I think.

An assert usually means "if the assert fail, it wouldn't have been possible to continue executing the code in a safe way", it's not the case here

alice-i-cecile:
Seconding the warning; I think that's a more appropriate failure level.

jmskov:
Is it possible to raise a warning (or other message for that matter) from a constructor that is passed to add_plugins, i.e. from `with_num_threads` below?
```rust
.add_plugins(DefaultPlugins.set(TaskPoolPlugin {
            task_pool_options: TaskPoolOptions::with_num_threads(4),
        }))
```
I've tried both `log::warn` and `tracing::warn`, and neither work here. Both work in the `create_default_pools` system called later.

jmskov:
I suppose it has to do with both TaskPoolPlugin and LogPlugin being in the DefaultPlugins set, because adding plugins after this can raise log messages using a constructor method. At this point, maybe it's best just to add a note to the documentation of `with_num_threads` and call it a day 😊

alice-i-cecile:
> I suppose it has to do with both TaskPoolPlugin and LogPlugin being in the DefaultPlugins set, because adding plugins after this can raise log messages using a constructor method. At this point, maybe it's best just to add a note to the documentation of with_num_threads and call it a day 😊

You can probably swap the ordering of the two plugins and make this work properly :)

jmskov:
Updated to a warning.

> You can probably swap the ordering of the two plugins and make this work properly :)

I can raise the warning if I add plugins separately, but the following does not raise a warning (or any trace, debug, etc.)
```rust
.add_plugins(DefaultPlugins.set(TaskPoolPlugin {    // This does not raise a warning!
             task_pool_options: TaskPoolOptions::with_num_threads(1),
}))
```

PR #17772, opened 2025-02-10
Cache `MeshInputUniform` indices in each `RenderBin`.

pcwalton:
Currently, we look up each `MeshInputUniform` index in a hash table that maps the main entity ID to the index every frame. This is inefficient, cache unfriendly, and unnecessary, as the `MeshInputUniform` index for an entity remains the same from frame to frame (even if the input uniform changes). This commit changes the `IndexSet` in the `RenderBin` to an `IndexMap` that maps the `MainEntity` to `MeshInputUniformIndex` (a new type that this patch adds for more type safety).

On Caldera with parallel `batch_and_prepare_binned_render_phase`, this patch improves that function from 3.18 ms to 2.42 ms, a 31% speedup.

### DIFF
--- a/examples/shader/specialized_mesh_pipeline.rs
+++ b/examples/shader/specialized_mesh_pipeline.rs
@@ -427,6 +427,7 @@ fn queue_custom_mesh_pipeline(
                     asset_id: AssetId::<Mesh>::invalid().untyped(),
                 },
                 (render_entity, visible_entity),
+                mesh_instance.current_uniform_index,
                 // This example supports batching, but if your pipeline doesn't
                 // support it you can use `BinnedRenderPhaseType::UnbatchableMesh`
                 BinnedRenderPhaseType::BatchableMesh,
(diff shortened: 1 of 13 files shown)


superdump (approved):

pcwalton:
Example runner looks good.

examples/shader/specialized_mesh_pipeline.rs
```diff
@@ -427,6 +427,7 @@ fn queue_custom_mesh_pipeline(
                     asset_id: AssetId::<Mesh>::invalid().untyped(),
                 },
                 (render_entity, visible_entity),
+                mesh_instance.current_uniform_index,
```
tychedelia, line 430: I think this is the same as `input_index` provided by `MeshPipeline::get_binned_index` above, which raises the bigger question of whether we could remove the `get_binned_index` method entirely, as it's only being used for unbatchables atm, which may make implementing it confusing. Minimally, the comment on `get_binned_index` should be changed to note it's not used for binning just unbatchables.

tychedelia (approved):
Lgtm.

examples/shader/specialized_mesh_pipeline.rs
pcwalton: Good point, I changed the comment.

PR #17777, opened 2025-02-10
added additional easing functions

### LINKED ISSUE #17744: Support more "step" easing modes

fjkorf:
# Objective

Fixes #17744 

## Solution

This updated adds three additional easing functions to support the full set of functions available in CSS
https://developer.mozilla.org/en-US/docs/Web/CSS/easing-function/steps#description

Three new `EaseFunction`s were added, and the original `EaseFunction::Steps` was modified.

I thought it made the most sense to have `Steps` be equivalent to `jump-none`, and added additional functions for the remaining options. `StepsStart` for `jump-start`, and so on. The original `Steps` function is now `StepsEnd`, which may be a problem .

## Testing

I updated the `EaseFunction`s example to include the three new functions, and successfully tested the code locally. 

---

## Showcase

Screenshot from 2025-02-10 11-26-56

## Migration Guide

Calls to `EaseFunction::Steps` would need to be updated to `EaseFunction::StepsEnd` to display the same behavior

### DIFF
--- a/crates/bevy_math/src/curve/easing.rs
+++ b/crates/bevy_math/src/curve/easing.rs
@@ -432,7 +432,7 @@ pub enum EaseFunction {
     ///
     #[doc = include_str!("../../images/easefunction/Steps.svg")]
     Steps(usize),
-    /// `n` steps connecting the start and the end. A jump is included at the start 
+    /// `n` steps connecting the start and the end. A jump is included at the start
     ///
     #[doc = include_str!("../../images/easefunction/Steps.svg")]
     StepsStart(usize),
--- a/crates/bevy_math/src/curve/easing.rs
+++ b/crates/bevy_math/src/curve/easing.rs
@@ -432,6 +432,18 @@ pub enum EaseFunction {
     ///
     #[doc = include_str!("../../images/easefunction/Steps.svg")]
     Steps(usize),
+    /// `n` steps connecting the start and the end. A jump is included at the start 
+    ///
+    #[doc = include_str!("../../images/easefunction/Steps.svg")]
+    StepsStart(usize),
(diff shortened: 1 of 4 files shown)


alice-i-cecile:
@greeble-dev can I get your review here? :)

alice-i-cecile (changes requested):
This feels both clearer and more consistent with CSS to add an enum and specify this behavior inside of the existing `Steps` variant. That will also make the migration easier and more automatic.

greeble-dev:
Awkward.... I think this PR and https://github.com/bevyengine/bevy/pull/17752 both implement the same feature.

#17752 has the advantage of addressing @alice-i-cecile 's comment about making the jump behaviour an enum, and includes updated svgs. But it adds a `StepConfig` intermediate struct that makes things a bit verbose.

```rust
let c = EasingCurve::new(2.0, 4.0, EaseFunction::Steps(StepConfig::new(4).with_jump_at(JumpAt::Start)));
```

I'd be happy to review either PR in more detail, but I'm reluctant to do that with both in play.

greeble-dev:
Although I'll walk that back slightly - if a maintainer says "please review both PRs and then we'll see what happens" then I can do that.

alice-i-cecile:
Let's coalesce the efforts in #17752 :)

PR #17778, opened 2025-02-10
Expand `EasingCurve` documentation

greeble-dev:
# Objective

- Expand the documentation for `EasingCurve`.
    - I suspect this might have avoided the confusion in https://github.com/bevyengine/bevy/pull/17711.
- Also add a shortcut for simple cases.

## Solution

- Added various examples and extra context.
- Implemented `Curve<T>` for `EaseFunction`.
    - This means `EasingCurve::new(0.0, 1.0, EaseFunction::X)` can be shortened to `EaseFunction::X`.
    - In some cases this will be a minor performance improvement.
    - Added test to confirm they're the same.
- ~~Added some benchmarks for bonus points.~~

## Side Notes

- I would have liked to rename `EaseFunction` to `EaseFn` for brevity, but that would be a breaking change and maybe controversial.
    - Also suspect `EasingCurve` should be `EaseCurve`, but say la vee.
- Benchmarks show that calling `EaseFunction::Smoothstep` is still slower than calling `smoothstep` directly.
    - I think this is because the compiler refuses to inline `EaseFunction::eval`.
    - I don't see any good solution - might need a whole different interface.

## Testing

```sh
cargo test --package bevy_math

cargo doc --package bevy_math
./target/doc/bevy_math/curve/easing/struct.EasingCurve.html

cargo bench --package benches --bench math -- easing
```


alice-i-cecile:
@fjkorf  can I get your review here?

alice-i-cecile (commented):
I like the expanded usage docs. So-so on the benchmark; I think I'd prefer to have that in its own PR.

mockersf (approved):

alice-i-cecile (approved):

PR #17783, opened 2025-02-10
run example in CI on windows using static dxc

mockersf:
# Objective

- Run more things on windows

## Solution

- With the update of wgpu and the statically linked dxc, examples now run on windows in CI


IceSentry (approved):

alice-i-cecile (approved):
Let's go~!

PR #17787, opened 2025-02-10
Sweep bins after queuing so as to only sweep them once.

### LINKED ISSUE #17759: Entities that should be visible are being culled
## Bevy version

main @ `556b75078288464e9628d3e939afdedfea34e958`

```
2025-02-09T14:21:00.168791Z  INFO bevy_diagnostic::system_information_diagnostics_plugin::internal: SystemInfo { os: "macOS 15.3 Sequoia", kernel: "24.3.0", cpu: "Apple M4 Max", core_count: "16", memory: "64.0 GiB" }    
2025-02-09T14:21:00.234400Z  INFO bevy_render::renderer: AdapterInfo { name: "Apple M4 Max", vendor: 0, device: 0, device_type: IntegratedGpu, driver: "", driver_info: "", backend: Metal }
```

## What you did

Run `cargo run --release --example many_cubes`

## What went wrong

Expected: all cubes that should be visible are visible in every frame

Observed: some cubes are not visible when they should be.

## Additional information

As they are only missing for a few frames it is difficult to take a screenshot but here is a short screen capture where the missing cubes can be seen as they come into view from the top of the frame.

https://github.com/user-attachments/assets/58e0be83-d8ad-43d6-9aa2-ea7f44fd732b

pcwalton:
Currently, we *sweep*, or remove entities from bins when those entities became invisible or changed phases, during `queue_material_meshes` and similar phases. This, however, is wrong, because `queue_material_meshes` executes once per material type, not once per phase. This could result in sweeping bins multiple times per phase, which can corrupt the bins. This commit fixes the issue by moving sweeping to a separate system that runs after queuing.

This manifested itself as entities appearing and disappearing seemingly at random.

Closes #17759.

### DIFF
--- a/crates/bevy_pbr/src/render/light.rs
+++ b/crates/bevy_pbr/src/render/light.rs
@@ -1955,7 +1955,7 @@ pub fn queue_shadows<M: Material>(
             }
 
             // Remove invalid entities from the bins.
-            shadow_phase.sweep_old_entities();
+            //shadow_phase.sweep_old_entities();
         }
     }
 }
--- a/crates/bevy_pbr/src/render/light.rs
+++ b/crates/bevy_pbr/src/render/light.rs
@@ -1953,9 +1953,6 @@ pub fn queue_shadows<M: Material>(
                     *current_change_tick,
                 );
             }
-
-            // Remove invalid entities from the bins.
-            //shadow_phase.sweep_old_entities();
         }
     }
 }
--- a/crates/bevy_pbr/src/material.rs
+++ b/crates/bevy_pbr/src/material.rs
@@ -1047,10 +1047,6 @@ pub fn queue_material_meshes<M: Material>(
                 }
             }
         }
-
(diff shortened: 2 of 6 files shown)


alice-i-cecile:
I spotted this over in #17769 too, for anyone looking for a reproduction.

crates/bevy_pbr/src/render/light.rs
```diff
@@ -1955,7 +1955,7 @@ pub fn queue_shadows<M: Material>(
             }
 
             // Remove invalid entities from the bins.
-            shadow_phase.sweep_old_entities();
+            //shadow_phase.sweep_old_entities();
```
tychedelia, line 1958: This can be deleted with the comment above

tychedelia (approved):

superdump (approved):
Approved aside from @tychedelia 's comment.

Issue #17763, opened 2025-02-09
Sprite in `bloom_2d` example is not rendered


rparrett:
## Bevy version

main, bisected to #17735

## Relevant system information

`SystemInfo { os: "macOS 15.3 Sequoia", kernel: "24.3.0", cpu: "Apple M4 Max", core_count: "16", memory: "64.0 GiB" }`
`AdapterInfo { name: "Apple M4 Max", vendor: 0, device: 0, device_type: IntegratedGpu, driver: "", driver_info: "", backend: Metal }`

But this also reproduces in CI in [the example report](https://thebevyflock.github.io/bevy-example-runner/).

## What you did

`cargo run --example bloom_2d`

## What went wrong

The bevy logo sprite that should be displayed in the center of the screen is not.

hukasu:
disabling tonemapping causes the sprite to show

![Image](https://github.com/user-attachments/assets/634af12e-ac42-4134-a0a1-031729525e07)

hukasu:
https://github.com/user-attachments/assets/920b2d6a-b237-4b19-be9d-9b651b4dc913

Elabajaba:
`main_opaque_pass_2d` writes to `main_texture_sampled` and `main_texture_a`, then `main_transparent_pass_2d` writes to `main_texture_sampled` and `main_texture_b`, but then the first bloom pass takes in `main_texture_a`.

I don't _think_ we expect to be flipping the main_texture_a/b between the main passes (we don't for 3d, and the flip-flopping is for post-processing so we don't have to allocate new textures for every pass).

hukasu:
confirmed the fix on d41de7f3dfebb351efac1cba998e07f1f88e9954

PR #17789, opened 2025-02-11
Add tonemapping switch to bloom 2d example

hukasu:
# Objective

Allow switching through available Tonemapping algorithms on `bloom_2d` example to compare between them

## Solution

Add a resource to `bloom_2d` that holds current tonemapping algorithm, a method to get the next one, and a check of key press to make the switch

## Testing

Ran `bloom_2d` example with modified code

## Showcase

https://github.com/user-attachments/assets/920b2d6a-b237-4b19-be9d-9b651b4dc913
Note: Sprite flashing is already described in #17763

### DIFF
--- a/examples/2d/bloom_2d.rs
+++ b/examples/2d/bloom_2d.rs
@@ -16,12 +16,14 @@ fn main() {
         .run();
 }
 
+/// Resource to allow cycling through available Tonemapping algorithms
 #[derive(Resource)]
 struct Tonemapper {
     tonemapper: Tonemapping,
 }
 
 impl Tonemapper {
+    /// Modifies resources tonemapping algorithm for the next one and returns it
     fn next(&mut self) -> Tonemapping {
         let next = match self.tonemapper {
             Tonemapping::None => Tonemapping::AcesFitted,
--- a/examples/2d/bloom_2d.rs
+++ b/examples/2d/bloom_2d.rs
@@ -79,10 +79,10 @@ fn update_bloom_settings(
     keycode: Res<ButtonInput<KeyCode>>,
     time: Res<Time>,
 ) {
-    let bloom = camera.into_inner();
+    let (camera_entity, tonemapping, bloom) = camera.into_inner();
 
     match bloom {
-        (entity, _, Some(mut bloom)) => {
+        Some(mut bloom) => {
             text.0 = "Bloom (Toggle: Space)\n".to_string();
             text.push_str(&format!("(Q/A) Intensity: {}\n", bloom.intensity));
             text.push_str(&format!(
@@ -112,7 +112,7 @@ fn update_bloom_settings(
             text.push_str(&format!("(I/K) Horizontal Scale: {}\n", bloom.scale.x));
 
             if keycode.just_pressed(KeyCode::Space) {
-                commands.entity(entity).remove::<Bloom>();
+                commands.entity(camera_entity).remove::<Bloom>();
             }
 
             let dt = time.delta_secs();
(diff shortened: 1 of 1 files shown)


examples/2d/bloom_2d.rs
```diff
@@ -110,6 +139,7 @@ fn update_bloom_settings(
                 bloom.prefilter.threshold_softness
             ));
             text.push_str(&format!("(I/K) Horizontal Scale: {}\n", bloom.scale.x));
+            text.push_str(&format!("(O) Tonemapper: {:?}\n", tonemapper.tonemapper));
```
rparrett, line 142: It would be nice if these instructions were still displayed, and the cycling still worked when bloom is disabled.

examples/2d/bloom_2d.rs
```diff
@@ -16,12 +16,40 @@ fn main() {
         .run();
 }
 
+/// Resource to allow cycling through available Tonemapping algorithms
+#[derive(Resource)]
+struct Tonemapper {
```
rparrett, line 21: It feels a bit redundant to store the current tonemapping method as a resource, when it's already quite accessible as a component on the camera.

hukasu: it is more to be able to implement the `next` method, but i can replace it for a method that is not part of a struct or a ext trait

rparrett: > method that is not part of a struct

Would personally prefer this (tucked away at the bottom of the file) over an ext trait or the resource

examples/2d/bloom_2d.rs
```diff
@@ -210,18 +180,33 @@ fn update_bloom_settings(
...
         }
     }
+
+    text.push_str(&format!("(O) Tonemapper: {:?}\n", bloom.1));
+    if keycode.just_pressed(KeyCode::KeyO) {
+        commands.entity(bloom.0).insert(next_tonemap(bloom.1));
+    }
+}
+
+/// Modifies resources tonemapping algorithm for the next one and returns it
```
rparrett: Comment is now outdated. Feel free to remove or update.

examples/2d/bloom_2d.rs
```diff
@@ -210,18 +180,33 @@ fn update_bloom_settings(
...
+        (entity, _, None) => {
+            text.0 = "Bloom: Off (Toggle: Space)\n".to_string();
 
             if keycode.just_pressed(KeyCode::Space) {
                 commands.entity(entity).insert(Bloom::default());
             }
         }
     }
+
+    text.push_str(&format!("(O) Tonemapper: {:?}\n", bloom.1));
```
rparrett, line 194: Just a nit.

`bloom.1` is a little sad.

This whole thing would clean up well with:

```rust
    let (entity, tonemapping, maybe_bloom) = camera.into_inner();

    match maybe_bloom {
        Some(mut bloom) => {
```

At the top of the fn.

examples/2d/bloom_2d.rs
```diff
@@ -210,18 +180,33 @@ fn update_bloom_settings(
...
+    if keycode.just_pressed(KeyCode::KeyO) {
+        commands.entity(bloom.0).insert(next_tonemap(bloom.1));
+    }
+}
+
+/// Modifies resources tonemapping algorithm for the next one and returns it
+fn next_tonemap(tonemap: &Tonemapping) -> Tonemapping {
```
rparrett, line 201: Just a nit: we've got "Tonemapper" "tonemapping" and "tonemap" all referring to this value. Would love to standardize on "tonemapping."

rparrett (approved):
Nice, this seems useful beyond the debugging you were using it for. I played with this once a while back when I was trying to figure out if bloom would work for one of my games.

I don't want to nitpick this to death. I could follow up with the other suggestions later, but we should fix the outdated comment.

examples/2d/bloom_2d.rs
rparrett: Nice, beat me to it.

JMS55:
I'm not a huge fan of this, personally. The example should be for demonstrating bloom, not tonemapping.

hukasu:
> I'm not a huge fan of this, personally. The example should be for demonstrating bloom, not tonemapping.

but tonemapping affects the appearance of bloom, wouldn't it be important for the user to see the difference?

alice-i-cecile (approved):

PR #17790, opened 2025-02-11
Silence deprecation warning in `Bundle` derive macro (#17369)

### LINKED ISSUE #17369: Deprecated bundles cause compiler warnings
## Bevy version

0.15.1

## What you did

Mark a bundle as deprecated in favor of required components.

```rust
#[deprecated(note = "all uses can be safely removed, uses required components")]
struct SomeBundle {
    // ...
}
```

## What went wrong

The derived implementation of `Bundle` and `DynamicBundle` cause compiler warnings about using a deprecated struct. IMO, `#[allow(deprecated)]` should be included in the derived implementations to silence these warnings.

Person-93:
# Objective

- Fixes #17369

## Solution

- Add `#[allow(deprecated)]` to the generated code.


alice-i-cecile (commented):
Ah interesting, apparently the expect wasn't needed. I'm still a bit ehh about this idea in general: I think it masks real problems to ease one-time migration pain.

alice-i-cecile:
Hmm, no I think I was wrong. This is only for the Bundle code gen, which isn't a particularly useful case to warn about.

alice-i-cecile (approved):

PR #17793, opened 2025-02-11
Don't relocate the meshes when mesh slabs grow.

### LINKED ISSUE #17766: Rendered geometry "breaks" when adding many primitives
## Bevy version

`main` after #16985 (`72ddac140a985bbfd66f8869110f3ac5e1c22b5c` is the commit when bisecting `main` and `v0.15.2`)

Ubuntu Linux 22.04
dynamic_linking feature
rustc 1.86.0-nightly (a730edcd6 2025-01-30)

### Adapter info

```ignore
`AdapterInfo { name: "NVIDIA GeForce RTX 4080", vendor: 4318, device: 9988, device_type: DiscreteGpu, driver: "NVIDIA", driver_info: "535.183.01", backend: Vulkan }`
```
## What you did

I'm just toying around with a tower-defense-style game.  When adding the tower-placement systems, I was placing lots of `Cylinder`s on a plane. 

## What went wrong

Expected:
I expected to see lots and lots of cylinders, maybe seeing the FPS drop as the number gets very large

Actual:
After a few 10s or 100s of cylinders, the geometry "turns inside out"  (see videos)

## Additional information
- I'm very new to Bevy, I could be using some bad memory-access pattern, and only getting lucky with ordering/optimization on older commits.
- No errors or warning in logs
- Maybe something to do with memory getting stomped?
- Occurs in both release and debug build profiles

###  Bad
https://github.com/user-attachments/assets/0cfbab52-6eee-4aa9-b142-197122691ddf

### Good
https://github.com/user-attachments/assets/3fc9d181-a375-4124-80ea-3f71e0fe674f

pcwalton:
Currently, when a mesh slab overflows, we recreate the allocator and reinsert all the meshes that were in it in an arbitrary order. This can result in the meshes moving around. Before `MeshInputUniform`s were retained, this was slow but harmless, because the `MeshInputUniform` that contained the positions of the vertex and index data in the slab would be recreated every frame. However, with mesh retention, there's no guarantee that the `MeshInputUniform`, which could be cached from the previous frame, will reflect the new position of the mesh data within the buffer if that buffer happened to grow. This manifested itself as seeming mesh data corruption when adding many meshes dynamically to the scene.

There are three possible ways that I could have fixed this that I can see:

1. Invalidate and rebuild all the `MeshInputUniform`s belonging to all meshes in a slab when that mesh grows.

2. Introduce a second layer of indirection so that the `MeshInputUniform` points to a *mesh allocation table* that contains the current locations of the data of each mesh.

3. Avoid moving meshes when reallocating the buffer.

To be efficient, option (1) would require scanning meshes to see if their positions changed, a la
`mark_meshes_as_changed_if_their_materials_changed`. Option (2) would add more runtime indirection and would require additional bookkeeping on the part of the allocator.

Therefore, this PR chooses option (3), which was remarkably simple to implement. The key is that the offset allocator happens to allocate addresses from low addresses to high addresses. So all we have to do is to *conceptually* allocate the full 512 MiB mesh slab as far as the offset allocator is concerned, and grow the underlying backing store from 1 MiB to 512 MiB as needed. In other words, the allocator now allocates *virtual* GPU memory, and the actual backing slab resizes to fit the virtual memory. This ensures that the location of mesh data remains constant for the lifetime of the mesh asset, and we can remove the code that reinserts meshes one by one when the slab grows in favor of a single buffer copy.

Closes #17766.

### DIFF
--- a/crates/bevy_render/src/mesh/allocator.rs
+++ b/crates/bevy_render/src/mesh/allocator.rs
@@ -792,10 +792,11 @@ impl MeshAllocator {
 
     /// Reallocates a slab that needs to be resized, or allocates a new slab.
     ///
-    /// This performs the actual growth operation that [`GeneralSlab::try_grow`]
-    /// scheduled. We do the growth in two phases so that, if a slab grows
-    /// multiple times in the same frame, only one new buffer is reallocated,
-    /// rather than reallocating the buffer multiple times.
+    /// This performs the actual growth operation that
+    /// [`GeneralSlab::grow_if_necessary`] scheduled. We do the growth in two
+    /// phases so that, if a slab grows multiple times in the same frame, only
+    /// one new buffer is reallocated, rather than reallocating the buffer
+    /// multiple times.
     fn reallocate_slab(
         &mut self,
         render_device: &RenderDevice,
--- a/crates/bevy_render/src/mesh/allocator.rs
+++ b/crates/bevy_render/src/mesh/allocator.rs
@@ -16,7 +16,7 @@ use bevy_ecs::{
     system::{Res, ResMut},
     world::{FromWorld, World},
 };
-use bevy_platform_support::collections::{HashMap, HashSet};
+use bevy_platform_support::collections::{hash_map::Entry, HashMap, HashSet};
 use bevy_utils::default;
 use offset_allocator::{Allocation, Allocator};
 use tracing::error;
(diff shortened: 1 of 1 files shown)


superdump (approved):

alice-i-cecile (commented):
How hard would it be to add a regression test for this? Doesn't need to be fully realistic.

pcwalton:
@alice-i-cecile We don't really have any rendering regression tests, so we'd have to have a framework for that. We have examples, but this doesn't really lend itself to an example.

alice-i-cecile (approved):

PR #17330, opened 2025-01-12
Harden proc macro path resolution and add integration tests.

raldone01:
This pr uses the `extern crate self as` trick to make proc macros behave the same way inside and outside bevy.

# Objective

- Removes noise introduced by `crate as` in the whole bevy repo.
- Fixes #17004.
- Hardens proc macro path resolution.

## TODO

- [x] `BevyManifest` needs cleanup.
- [x] Cleanup remaining `crate as`.
- [x] Add proper integration tests to the ci.

## Notes

- `cargo-manifest-proc-macros` is written by me and based/inspired by the old `BevyManifest` implementation and [`bkchr/proc-macro-crate`](https://github.com/bkchr/proc-macro-crate).
- What do you think about the new integration test machinery I added to the `ci`?
  More and better integration tests can be added at a later stage.
  The goal of these integration tests is to simulate an actual separate crate that uses bevy. Ideally they would lightly touch all bevy crates.

## Testing

- Needs RA test
- Needs testing from other users
- Others need to run at least `cargo run -p ci integration-test` and verify that they work.

PR #17795, opened 2025-02-11
Fix failing proc macros when depending on bevy through dev and normal dependencies.

### LINKED ISSUE #17780: Hardening proc macro path resolution broke cargo overrides
This broke all my usages of bevy derive with messages like this:

```
error[E0433]: failed to resolve: use of undeclared crate or module `bevy_reflect`
   --> C:\Users\pcwal\Source\Rust\leafwing-input-manager\src\lib.rs:144:41
    |
144 | #[derive(Debug, Clone, Copy, PartialEq, Reflect, Serialize, Deserialize)]
    |                                         ^^^^^^^ use of undeclared crate or module `bevy_reflect`
    |
    = note: struct `crate::user_input::trait_reflection::tripleaxislike::TypeRegistration` exists but is inaccessible
    = note: this error originates in the derive macro `Reflect` (in Nightly builds, run with -Z macro-backtrace for more info)
```

and

```
error[E0277]: `user_input::gamepad::GamepadStick` can not provide dynamic type information through reflection
   --> C:\Users\pcwal\Source\Rust\leafwing-input-manager\src\user_input\gamepad.rs:548:23
    |
548 | impl DualAxislike for GamepadStick {
    |                       ^^^^^^^^^^^^ the trait `bevy::bevy_reflect::Typed` is not implemented for `user_input::gamepad::GamepadStick`
```

For reference, I'm using Cargo overrides to redirect Bevy 0.16.0 to the local repo.

_Originally posted by @pcwalton in https://github.com/bevyengine/bevy/issues/17330#issuecomment-2646851696_

raldone01:
This is a follow up fix for #17330 and fixes #17780.
There was a logic error in the ambiguity detection of `cargo-manifest-proc-macros`.
`cargo-manifest-proc-macros` now has a test for this case to prevent the issue in the future.

I also opted to hard fail if the `cargo-manifest-proc-macros` crate fails. That way the error is more obvious and easier to fix and diagnose.

## Testing

- The reproducer: https://github.com/bevyengine/bevy_editor_prototypes/pull/178 works for me using these fixes.

### DIFF
--- a/crates/bevy_macro_utils/Cargo.toml
+++ b/crates/bevy_macro_utils/Cargo.toml
@@ -12,7 +12,7 @@ keywords = ["bevy"]
 syn = "2.0"
 quote = "1.0"
 proc-macro2 = "1.0"
-cargo-manifest-proc-macros = "0.3.3"
+cargo-manifest-proc-macros = "0.3.4"
 
 [lints]
 workspace = true
--- a/crates/bevy_macro_utils/src/bevy_manifest.rs
+++ b/crates/bevy_macro_utils/src/bevy_manifest.rs
@@ -48,8 +48,7 @@ impl BevyManifest {
     /// Returns the path for the crate with the given name.
     pub fn get_path(&self, name: &str) -> syn::Path {
         self.maybe_get_path(name)
-            //.expect("Failed to get path for crate")
-            .unwrap_or_else(|_err| Self::parse_str(name))
+            .expect("Failed to get path for crate")
     }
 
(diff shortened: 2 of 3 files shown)


villor:
I can confirm that this fixes the issues encountered in https://github.com/bevyengine/bevy_editor_prototypes/pull/178
Using this PR in place of bevy main, the code compiles with no errors.

alice-i-cecile (approved):

alice-i-cecile:
Fantastic; thank you so much for the prompt investigation and fix! The upstream fix was particularly nice :)

PR #17796, opened 2025-02-11
implement iterators that yield UniqueEntitySlice

Victoronz:
# Objective

Continuation of #17589 and #16547.

Slices have several methods that return iterators which themselves yield slices, which we have not yet implemented.
An example use is `par_iter_many` style logic.

## Solution

Their implementation is rather straightforward, we simply delegate all impls to `[T]`.
The resulting iterator types need their own wrappers in the form of `UniqueEntitySliceIter` and `UniqueEntitySliceIterMut`.

We also add three free functions that cast slices of entity slices to slices of `UniqueEntitySlice`. 
These three should be sufficient, though infinite nesting is achievable with a trait (like `TrustedEntityBorrow` works over infinite reference nesting), should the need ever arise.

### DIFF
--- a/crates/bevy_ecs/src/entity/unique_slice.rs
+++ b/crates/bevy_ecs/src/entity/unique_slice.rs
@@ -1,7 +1,8 @@
 use core::{
     borrow::Borrow,
     cmp::Ordering,
-    fmt::Debug,
+    fmt::{self, Debug, Formatter},
+    iter::FusedIterator,
     ops::{
         Bound, Deref, Index, IndexMut, Range, RangeFrom, RangeFull, RangeInclusive, RangeTo,
         RangeToInclusive,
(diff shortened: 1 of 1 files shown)


alice-i-cecile (approved):

alice-i-cecile:
As always, clean work. Thanks!

crates/bevy_ecs/src/entity/unique_slice.rs
```diff
@@ -973,3 +1313,368 @@ impl<'a, T: TrustedEntityBorrow> UniqueEntityIter<slice::IterMut<'a, T>> {
...
+        // SAFETY:
+        unsafe { cast_slice_of_unique_entity_slice(self.iter.as_ref()) }
+    }
+}
+
+impl<'a, T: TrustedEntityBorrow + 'a, I: Iterator<Item = &'a [T]>> Debug
+    for UniqueEntitySliceIter<'a, T, I>
+{
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        f.debug_struct("UniqueEntitySliceIter").finish()
```
chescock, line 1402: Why not just `#[derive(Debug)]`?  (And similarly in `UniqueEntitySliceIterMut`.)  If `I` has a useful `Debug` impl, then it seems like a shame not to use it.  And if it doesn't, then it seems reasonable for a wrapped `I` not to, either.

crates/bevy_ecs/src/entity/unique_slice.rs
```diff
@@ -973,3 +1313,368 @@ impl<'a, T: TrustedEntityBorrow> UniqueEntityIter<slice::IterMut<'a, T>> {
         unsafe { UniqueEntitySlice::from_slice_unchecked(self.as_inner().as_slice()) }
     }
 }
+
+/// An iterator that yields unique entity/entity borrow slices.
```
chescock, line 1317: This comment was a little unclear to me at first.  I wasn't sure whether this meant the slices were unique, or the contents of each slice were unique, or the contents of all slices together were unique.  Since you have `fn windows()`, it must be the middle one.  

Maybe it would help to use the `UniqueEntitySlice` type name?  That would make the meaning of `Unique` unambiguous.  Maybe something like "An iterator that yields `&UniqueEntitySlice`s.  Note that an entity may appear in more than one `UniqueEntitySlice`, but no more than once within each `UniqueEntitySlice`."  

(And similarly for `UniqueEntitySliceIterMut`.)

chescock (approved):

crates/bevy_ecs/src/entity/unique_slice.rs
Victoronz: I had overlooked this! Fixed.

crates/bevy_ecs/src/entity/unique_slice.rs
Victoronz: I've clarified a bit!

PR #17799, opened 2025-02-11
Fix error in scene example

### LINKED ISSUE #17798: some examples don't work
`cargo run --example scene`
- serializes just fine into the terminal but the deserialization doesn't:
- ERROR bevy_asset::server: Failed to load asset 'scenes/load_scene_example.scn.ron' with asset loader 'bevy_scene::scene_loader::SceneLoader': Could not parse RON: 10:33: Expected string

`cargo run --example desk_toy`
- the transparency doesn't work. the window is black instead of see-through

`cargo run --example texture_atlas`
- all text is strobbing (flashing in and out, like a broken neon lamp). this has to be unintended.

all examples in directory "testbed" cannot be run
- error: no example target named `2d`
- error: no example target named `3d`
- error: no example target named `ui_layout_rounding`
- error: no example target named `ui`
- error: no example target named `helpers`

i've not tried every example. maybe i will encounter more

rparrett:
# Objective

After #16894, this example started logging errors:

```
ERROR bevy_asset::server: Failed to load asset 'scenes/load_scene_example.scn.ron' with asset loader 'bevy_scene::scene_loader::SceneLoader': Could not parse RON: 10:33: Expected string
```

Fixes #17798, this is the only actionable/unreported issue in there as far as I can tell.

## Solution

Update the serialized scene with the expected format for `Name`

## Testing

`cargo run --example scene`

## Discussion

This example breaks very often and we don't always catch it. It might be nice to have this scene either

1. produce visual output so that it can be checked
2. panic if the scene fails to load (check for LoadState::Failed)

Either of those would make the failures visible in [the example report](https://thebevyflock.github.io/bevy-example-runner/). Not sure which method would best suit the example.

### DIFF
--- a/assets/scenes/load_scene_example.scn.ron
+++ b/assets/scenes/load_scene_example.scn.ron
@@ -7,10 +7,7 @@
   entities: {
     4294967296: (
       components: {
-        "bevy_ecs::name::Name": (
-          hash: 17588334858059901562,
-          name: "joe",
-        ),
+        "bevy_ecs::name::Name": "name",
         "bevy_transform::components::global_transform::GlobalTransform": ((1.0, 0.0, 0.0, 0.0, 1.0, 0.0, 0.0, 0.0, 1.0, 0.0, 0.0, 0.0)),
         "bevy_transform::components::transform::Transform": (
           translation: (0.0, 0.0, 0.0),
--- a/assets/scenes/load_scene_example.scn.ron
+++ b/assets/scenes/load_scene_example.scn.ron
@@ -7,7 +7,7 @@
   entities: {
     4294967296: (
       components: {
-        "bevy_ecs::name::Name": "name",
+        "bevy_ecs::name::Name": "joe",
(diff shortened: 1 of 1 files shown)


assets/scenes/load_scene_example.scn.ron
```diff
@@ -7,10 +7,7 @@
   entities: {
     4294967296: (
       components: {
-        "bevy_ecs::name::Name": (
-          hash: 17588334858059901562,
-          name: "joe",
-        ),
+        "bevy_ecs::name::Name": "name",
```
mockersf, line 10: could you use a name that is not "name", just to avoid confusion? 🙂

rparrett: Oops, nice catch!

mockersf (approved):

alice-i-cecile (approved):

mockersf:
> This example breaks very often and we don't always catch it. It might be nice to have this scene either
> 
> 1. produce visual output so that it can be checked
> 2. panic if the scene fails to load (check for LoadState::Failed)
> 
> Either of those would make the failures visible in [the example report](https://thebevyflock.github.io/bevy-example-runner/). Not sure which method would best suit the example.

option 2 for me!

Issue #17784, opened 2025-02-10
Entity{Ref,Mut}Except is duplicate functionality to FilteredEntity{Ref,Mut}


anlumo:
## Bevy version

0.15.2

## What you did

Working on animation stuff.

## What went wrong

`Entity{Ref,Mut}Except` are missing some functionality I need, like `get_by_id`. However, it appears that these types aren't used anywhere except the animation system, and it should be possible to replace them with the existing `FilteredEntity{Ref,Mut}`, which has all the functions I need.

anlumo:
Calling [Access::read_all_components](https://docs.rs/bevy_ecs/0.15.2/src/bevy_ecs/query/access.rs.html#52) first inverts the behavior to a blocklist instead of an allowlist.

chescock:
The "Pointing Fingers: Deduplicating Entity APIs with EntityPtr" working group is planning to merge a lot of the code between these types, so they'll eventually be able to share implementations of methods like `get_by_id`.  I'd vote for just adding the methods you need to `EntityMutExcept`, since we already have a plan to reduce the maintenance burden in the future.  

Note that `FilteredEntityMut` currently has some performance issues because it has to `clone()` an `Access` for each instance.  `EntityMutExcept` embeds that all in the type, so it can be created cheaply.  (We may be able to fix that with #15396, which I'm planning to get back in shape now that some of the blockers have been cleared out.)

anlumo:
> Note that `FilteredEntityMut` currently has some performance issues because it has to `clone()` an `Access` for each instance.

I see nothing in the type that would stop a switch to `Arc<Access>`, but I get that this would still be slower than storing this in the type system.

PR #17800, opened 2025-02-11
Add ComponentId-taking functions to Entity{Ref,Mut}Except to mirror FilteredEntity{Ref,Mut}

anlumo:
# Objective

Related to #17784. The ticket is actually about just getting rid of `Entity{Ref,Mut}Except` in favor of `FilteredEntity{Ref,Mut}`, but I got told the unification of Entity types is a bigger endeavor that has been going on for a while now (as the "Pointing Fingers" working group) and I should just add the functions I actually need in the meantime.

## Solution

This PR adds all of the functions necessary to access components by TypeId or ComponentId instead of static types.

## Testing

> Did you test these changes? If so, how?

Haven't tested it yet, but the changes are mostly copy/paste from other implementations in the same file, since there is a lot of duplicated functionality there.

## Not a Migration Guide

There shouldn't be any breaking changes, it's just a few new functions on existing types.

I had to shuffle around the lifetimes in `From<&EntityMutExcept<'a, B>> for EntityRefExcept<'a, B>` (originally it was `From<&'a EntityMutExcept<'_, B>> for EntityRefExcept<'_, B>`) to make the borrow checker happy, but I don't think that this should have an impact on user code (correct me if I'm wrong).

### DIFF
--- a/crates/bevy_ecs/src/world/entity_ref.rs
+++ b/crates/bevy_ecs/src/world/entity_ref.rs
@@ -3737,13 +3737,100 @@ where
     pub fn spawned_by(&self) -> MaybeLocation {
         self.entity.spawned_by()
     }
+
+    /// Gets the component of the given [`ComponentId`] from the entity.
+    ///
+    /// **You should prefer to use the typed API [`Self::get`] where possible and only
+    /// use this in cases where the actual component types are not known at
+    /// compile time.**
+    ///
+    /// Unlike [`EntityRefExcept::get`], this returns a raw pointer to the component,
+    /// which is only valid while the [`EntityRefExcept`] is alive.
+    #[inline]
+    pub fn get_by_id(&self, component_id: ComponentId) -> Option<Ptr<'w>> {
+        let components = self.entity.world().components();
+        (!bundle_contains_component::<B>(components, component_id))
+            .then(|| {
+                // SAFETY: We have read access for this component
+                unsafe { self.entity.get_by_id(component_id) }
+            })
+            .flatten()
+    }
+
+    /// Returns `true` if the current entity has a component of type `T`.
+    /// Otherwise, this returns `false`.
+    ///
+    /// ## Notes
+    ///
+    /// If you do not know the concrete type of a component, consider using
+    /// [`Self::contains_id`] or [`Self::contains_type_id`].
+    #[inline]
+    pub fn contains<T: Component>(&self) -> bool {
+        self.contains_type_id(TypeId::of::<T>())
+    }
+
+    /// Returns `true` if the current entity has a component identified by `component_id`.
+    /// Otherwise, this returns false.
+    ///
+    /// ## Notes
+    ///
+    /// - If you know the concrete type of the component, you should prefer [`Self::contains`].
+    /// - If you know the component's [`TypeId`] but not its [`ComponentId`], consider using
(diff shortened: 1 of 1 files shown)


alice-i-cecile (approved):

crates/bevy_ecs/src/world/entity_ref.rs
```diff
@@ -3737,13 +3773,100 @@ where
...
+    /// Gets the component of the given [`ComponentId`] from the entity.
+    ///
+    /// **You should prefer to use the typed API [`Self::get`] where possible and only
+    /// use this in cases where the actual component types are not known at
+    /// compile time.**
+    ///
+    /// Unlike [`EntityRefExcept::get`], this returns a raw pointer to the component,
+    /// which is only valid while the [`EntityRefExcept`] is alive.
+    #[inline]
+    pub fn get_by_id(&self, component_id: ComponentId) -> Option<Ptr<'w>> {
```
chescock, line 3786: I think this could be generalized using `DynamicComponentFetch` to match the other `get_by_id()` methods.  But that should happen anyway when these get merged by the "Pointing Fingers: Deduplicating Entity APIs with EntityPtr" working group, so this is fine for now!

crates/bevy_ecs/src/world/entity_ref.rs
```diff
@@ -3737,13 +3773,100 @@ where
...
+            .then(|| {
+                // SAFETY: We have read access
+                unsafe { self.entity.get_change_ticks_by_id(component_id) }
+            })
+            .flatten()
+    }
 }
 
-impl<'a, B> From<&'a EntityMutExcept<'_, B>> for EntityRefExcept<'a, B>
+impl<'a, B> From<&EntityMutExcept<'a, B>> for EntityRefExcept<'a, B>
```
chescock, line 3865: I think this change is unsound.  

Given `&'c EntityMutExcept<'d, B>`, it's valid to create an `EntityRefExcept<'c, B>` because we know nothing has mutable access through the original `EntityMutExcept` for the duration of `'c`.  But if we change it to create an `EntityRefExcept<'d, B>`, then once `'c` ends we can have both the original `EntityMutExcept<'d, B>` and the derived `EntityRefExcept<'d, B>` alive at the same time, which would allow unsound aliasing.  

That is, this change lets the following bad code compile:
```rust
fn bad_aliasing<'a, B: Bundle, C: Component<Mutability = Mutable>>(mut m: EntityMutExcept<'a, B>) {
    let r: EntityRefExcept<B> = EntityRefExcept::from(&m);
    let rc: &C = r.get::<C>().unwrap();
    let mc: &mut C = &mut *m.get_mut::<C>().unwrap();
    // Both &C and &mut C are alive at the same time!
    assert!(std::ptr::eq(rc, mc));
}
```

What was the borrow checker issue you were trying to resolve?

anlumo: I moved this back to the original version and the compiler doesn't complain any more, so I guess I solved it as a side effect of some other change.

anlumo:
@chescock I reverted the unsound lifetime change by an amend commit, apparently that removes the entire conversation from GitHub. Everything seems to be fine, so the issues I was having apparently were solved by some other change during development.

chescock (approved):

Issue #17720, opened 2025-02-07
Children in spawned gltf scenes are incorrectly ordered


rparrett:
## Bevy version

main, bisected to #17687

## Relevant system information

SystemInfo { os: "macOS 15.3 Sequoia", kernel: "24.3.0", cpu: "Apple M4 Max", core_count: "16", memory: "64.0 GiB" }    
AdapterInfo { name: "Apple M4 Max", vendor: 0, device: 0, device_type: IntegratedGpu, driver: "", driver_info: "", backend: Metal }

Also failing on linux/mac in the [example runner](https://thebevyflock.github.io/bevy-example-runner/).

## What you did

`cargo run --example gltf_skinned_mesh`

## What went wrong

```
thread 'Compute Task Pool (1)' panicked at examples/animation/gltf_skinned_mesh.rs:61:68:
called `Result::unwrap()` on an `Err` value: QueryDoesNotMatch(12v1 with components Transform, GlobalTransform, Visibility, InheritedVisibility, ViewVisibility, VisibilityClass, ChildOf, PreviousGlobalTransform, Mesh3d, SkinnedMesh, Name, MeshMaterial3d<StandardMaterial>, Aabb)
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
Encountered a panic in system `gltf_skinned_mesh::joint_animation`!
Encountered a panic in system `bevy_app::main_schedule::Main::run_main`!
```

rparrett:
It seems that the order of the children within the spawned hierarchy has changed.

```
// Bevy 0.15.2
8v1
 9v1
  GltfNode0
   Mesh
   GltfNode1
    GltfNode2

// Bevy main
8v1
 9v1
  GltfNode0
   GltfNode1
    GltfNode2
   Mesh
```

So the following from `gltf_skinned_mesh` is no longer true:

```
// First joint is the second child of the mesh node.
```

So the following patch fixes the example:
```diff
diff --git a/examples/animation/gltf_skinned_mesh.rs b/examples/animation/gltf_skinned_mesh.rs
index 90e95d53f..2a8a5228d 100644
--- a/examples/animation/gltf_skinned_mesh.rs
+++ b/examples/animation/gltf_skinned_mesh.rs
@@ -56,7 +56,7 @@ fn joint_animation(
         let mesh_node_parent = parents.get(mesh_node_entity).unwrap();
 
         // First joint is the second child of the mesh node.
[...]

Issue #8099, opened 2023-03-15
Fox.glb has incorrectly scaled normals


NotAFile:
## Bevy version

35e107ca

## What went wrong

The `Fox.glb` model has incorrect scaling of the vertex normals.

e.g. for the first vertex, the normals are:

```
0.176476 -0.507666 -0.242626
```

correctly scaled would be:

```
0.299268 -0.860901 -0.411446
```

This has led to issues such as  #6528. That has been fixed, but it would be good to fix this at the source anyway.

alecgarza96:
New to Bevy, would love to help out. Still learning the codebase

Does this change involve edits to the 3D model itself? I looked under examples/animated_fox.rs and didn't find any code performing scaling. 

I've been searching under crates/bevy_render as well for where this error could be originating

alecgarza96:
> > Does this change involve edits to the 3D model itself? I looked under examples/animated_fox.rs and didn't find any code performing scaling.
> 
> Looks like it. There's no normalization code we do on any meshes by default. IMO we should any or all of the following:
> 
> * Update the 3D model itself to include normalized normals
> * Add a validation step when loading meshes to ensure normals are normalized.
> * Add a asset preprocessing step to do this when that pipeline is ready (contingent on the asset rework @cart's working on)

Looking into adding the validation step for when meshes are normalized

I found the compute_flat_normals method which utilizes the face_normal

would it be possible to reuse this functionality for the validation check, or do we want to calculate the standard normals instead of flat?

Also, at what point should the validation check occur? My initial thoughts are in either ::new() or inert_attribute()

JohnTheCoolingFan:
While you are figuring out validation, I've been trying to fix the model itself. I've loaded it into blender, recalculated normals (removed duplicate vertices, recalculated, disconnected faces back, export). Visually nothing changed (judging by blender's normal visualization), I tried it on commit specified in #6528 and it's still black. I've uploaded my updated model to a fork: https://github.com/JohnTheCoolingFan/bevy/blob/dfdda55edb9e1ac6f6d497259faed32dd300068b/assets/models/animated/Fox.glb

NotAFile:
For reference, the way I personally did the normalization was via MeshLab, which has a "normalize normals" option. I unfortunately could not get to output a drop in replacement .glb file though.

mockersf:
If the wrong normals are confirmed on the upstream fox, fixing them would be very nice. If not, we can update ours with the original 👍

[...]

PR #17801, opened 2025-02-11
Fix validation errors in `Fox.glb`

rparrett:
# Objective

Fix gltf validation errors in `Fox.glb`.

Inspired by #8099, but that issue doesn't appear to describe a real bug to fix, as far as I can tell.

## Solution

Use the latest version of the Fox from [glTF-Sample-Assets](https://github.com/KhronosGroup/glTF-Sample-Assets/blob/main/Models/Fox/glTF-Binary/Fox.glb).

## Testing

Dropped both versions in https://github.khronos.org/glTF-Validator/

`cargo run --example animated_mesh` seems to still look fine.

Before:

```
The asset contains errors.
"numErrors": 126,
"numWarnings": 4184,
```

After:

```
The asset is valid.
"numErrors": 0,
"numWarnings": 0,
```

## Discussion

The 3d testbed was panicking with
```
thread 'main' panicked at examples/testbed/3d.rs:288:60:
called `Result::unwrap()` on an `Err` value: QueryDoesNotMatch(35v1 with components Transform, GlobalTransform, Visibility, InheritedVisibility, ViewVisibility, ChildOf, Children, Name)
```
Which is bizarre. I think this might be related to #17720, or maybe the structure of the gltf changed.

I fixed it by using updating the testbed to use a more robust method of finding the correct entity as is done in `animated_mesh`.


Vrixyz:
Cool!
nitpick: maybe refer to https://github.com/KhronosGroup/glTF-Sample-Assets/blob/main/Models/Fox/glTF-Binary/Fox.glb (in the future?) ; I think it's the same model but I was surprised to see "latest model" linking to an archived repo.

rparrett:
Oops, updated link and confirmed that it's the same.

mockersf (approved):

alice-i-cecile (approved):

PR #17807, opened 2025-02-11
Remove BlitPlugin and introduce TextureBlitter

IceSentry:
# Objective

wgpu 24 added a new TextureBlitter that handles doing blitting between 2 textures. Unfortunately, it's currently a bit limited but we will likely want to start use it once it has been updated to support the required features. It should be able to replace the need for the blit plugin and simplifies doing. a blit.

## Solution

- Remove the blit plugin
- Copy the texture_blitter module from wgpu with some small modifications done for our use cases
- Add the blit() function to the render graph context. This is needed because we need both the render device and the command encoder to do a blit.
- Update the upscaling and msaa writeback node to use this new texture blitter.

## Testing

Confirmed that msaa writeback still works with the split_screen example

---

## Migration Guide

TODO explain how to use the new texture blitter


IceSentry:
Oof, that doesn't work on wasm. Back to drawing board. I'll probably need a cache stored in a resource or maybe just use the asset system.

IceSentry:
If anyone wants this feel free to revive it.

PR #17808, opened 2025-02-11
Split PointerId into distinct id and PointerKind fields

alice-i-cecile:
# Objective

As discussed in https://github.com/bevyengine/bevy/pull/17399, sending fake picking events (as you might want to do for clicking a focused element) is frustrating.

After discussing this with @aevyrie (notes [here](https://hackmd.io/swhNP8IYTmGqCOOkyfUFFA)), part of the problem here is that our `PointerId` type is questionably designed. There are several small problems with it:

1. Constructing Uuids is very annoying.
2. Storing the id inside of each variant, rather than splitting the ID apart from the variants encourages inconsistent design and makes matching harder.
3. The `Custom` variant has unclear semantics. When mocking inputs for tests, you almost always want to pretend to be a real mouse or touch (often at a level above this, sending raw input events). But when sending picking events for "click on the focused element", it's not clear what whether you should pretend to be a mouse, a custom input, or even a touch.

## Solution

1. Store the id of each pointer separately from the kind of pointer it is.
2. Standardize on a u64 identifier, and cut uuid from bevy_picking completely.
3. Add a `PointerKind::Virtual` variant with clear semantics for this sort of software-driven pointer that doesn't correspond to real hardware.

## Future work

It would be nice to support multiple mice at some point in the future (I've had creative coding users asking me for this). Unfortunately, while this PR ensures that bevy_picking could easily accommodate it, winit still assumes a single mouse (and keyboard).

Unsurprisingly, patching winit is outside of the scope of this PR.

## Testing

Existing tests pass. The picking examples seem to work fine, although I haven't tested them with touch inputs.

## Migration Guide

`PointerId` has been refactored for consistency. Check the `kind` field to see the `PointerKind` of the pointer, and read the `id` field to recover the unique identifier.

To construct a `PointerId::Mouse`, use the `PointerId::MOUSE` constant. If you were using `PointerId::Custom`, either use `PointerKind::Mouse` (if you were mocking inputs for tests) or use `PointerKind::Virtual` if you were emulating a gamepad-controlled pointer or mocking pointer events to interact with various objects.

### DIFF
--- a/crates/bevy_picking/src/pointer.rs
+++ b/crates/bevy_picking/src/pointer.rs
@@ -16,8 +16,6 @@ use bevy_reflect::prelude::*;
 use bevy_render::camera::{Camera, NormalizedRenderTarget};
 use bevy_window::PrimaryWindow;
 
-use uuid::Uuid;
-
 use core::{fmt::Debug, ops::Deref};
 
 use crate::backend::HitData;
@@ -29,39 +27,67 @@ use crate::backend::HitData;
 #[derive(Debug, Default, Clone, Copy, Eq, PartialEq, Hash, Component, Reflect)]
 #[require(PointerLocation, PointerPress, PointerInteraction)]
 #[reflect(Component, Default, Debug, Hash, PartialEq)]
-pub enum PointerId {
-    /// The mouse pointer.
-    #[default]
-    Mouse,
-    /// A touch input, usually numbered by window touch events from `winit`.
-    Touch(u64),
-    /// A custom, uniquely identified pointer. Useful for mocking inputs or implementing a software
-    /// controlled cursor.
-    #[reflect(ignore)]
-    Custom(Uuid),
+pub struct PointerId {
+    /// A stable identifier for a specific pointer.
+    ///
+    /// This is typically provided by winit,
+    /// and is unique for each touch input or mouse.
+    pub id: u64,
+    /// The type of hardware device that created this pointer.
+    pub kind: PointerKind,
 }
 
 impl PointerId {
+    /// The [`PointerId`] of the mouse.
+    ///
+    /// Currently, only one mouse pointer is supported.
+    pub const MOUSE: Self = Self {
+        id: 0,
+        kind: PointerKind::Mouse,
+    };
+
     /// Returns true if the pointer is a touch input.
     pub fn is_touch(&self) -> bool {
-        matches!(self, PointerId::Touch(_))
+        matches!(self.kind, PointerKind::Touch)
     }
     /// Returns true if the pointer is the mouse.
     pub fn is_mouse(&self) -> bool {
-        matches!(self, PointerId::Mouse)
+        matches!(self.kind, PointerKind::Mouse)
     }
-    /// Returns true if the pointer is a custom input.
-    pub fn is_custom(&self) -> bool {
-        matches!(self, PointerId::Custom(_))
+    /// Returns true if the pointer is a virtual input.
+    pub fn is_virtual(&self) -> bool {
+        matches!(self.kind, PointerKind::Virtual)
     }
     /// Returns the touch id if the pointer is a touch input.
     pub fn get_touch_id(&self) -> Option<u64> {
-        if let PointerId::Touch(id) = self {
-            Some(*id)
+        if let PointerKind::Touch = self.kind {
+            Some(self.id)
         } else {
             None
         }
     }
+}
+
+/// The input device that created a pointer.
+#[derive(Default, Debug, Clone, Copy, PartialEq, Eq, Hash, Reflect)]
+#[reflect(Default, Debug, Hash, PartialEq)]
+pub enum PointerKind {
+    /// A pointer that is controlled by a mouse.
+    ///
+    /// Not a rodent; the device that you move around on a desk.
+    Mouse,
+    /// A pointer that is controlled by a touch input.
+    ///
+    /// Typically with your fingers on a touch screen,
+    /// but can also be a stylus or other touch input device.
+    Touch,
+    /// A pointer that is driven by code.
+    ///
+    /// No underlying hardware is associated with this pointer,
+    /// and it might represent a gamepad moving a cursor,
+    /// or an input sent to a focused element in a UI.
+    #[default]
+    Virtual,
 }
 
 /// Holds a list of entities this pointer is currently interacting with, sorted from nearest to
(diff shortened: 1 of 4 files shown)


crates/bevy_picking/src/pointer.rs
```diff
@@ -29,41 +27,69 @@ use crate::backend::HitData;
...
+    /// The [`PointerId`] of the mouse.
+    ///
+    /// Currently, only one mouse pointer is supported.
+    pub const MOUSE: Self = Self {
+        id: 0,
+        kind: PointerKind::Mouse,
+    };
+
     /// Returns true if the pointer is a touch input.
     pub fn is_touch(&self) -> bool {
```
alice-i-cecile, line 50: I've left these utility methods alone to ease migration, but I'm not convinced they're particularly valuable now.

crates/bevy_picking/src/pointer.rs
```diff
@@ -29,41 +27,69 @@ use crate::backend::HitData;
...
-    /// The mouse pointer.
-    #[default]
-    Mouse,
-    /// A touch input, usually numbered by window touch events from `winit`.
-    Touch(u64),
-    /// A custom, uniquely identified pointer. Useful for mocking inputs or implementing a software
-    /// controlled cursor.
-    #[reflect(ignore)]
-    Custom(Uuid),
+pub struct PointerId {
```
cart, line 30: Given that PointerId is now a common "thing" maybe it should just be `PointerId(u64)` and `PointerKind` would just be stored adjacent to it in places where that is relevant (ex: as a separate Component)? To distinguish between two pointers, don't we only need to compare / hash the ID, not the kind?

cart: If it is _not_ a common thing. Then I think we should move back to a typed-wrapper to protect against comparing "unlike" ids across types.

alice-i-cecile: Right now (and before this PR), the IDs are only meaningfully used by touches. In theory you could encode information for Virtual / Custom types in the u64 / UUID, but I'm not sure what you would use that for.

In either case, in order for two PointerId to match, both the kind and the number associated with it must match. As a result, I don't think we should split these up, but I'm fine to go back to an enum design here. Which do you prefer?

1. The current design, with split Id + kind.
2. `enum PointerId`, only `Touch` gets a `u64`.
3.  `enum PointerId`, each kind has a `u64`.

I think I prefer solution 2, but went with solution 1 to ease migration.

crates/bevy_picking/src/pointer.rs
```diff
@@ -29,41 +27,69 @@ use crate::backend::HitData;
...
 }
 
+/// The input device that created a pointer.
+#[derive(Default, Debug, Clone, Copy, PartialEq, Eq, Hash, Reflect)]
+#[reflect(Default, Debug, Hash, PartialEq)]
+pub enum PointerKind {
+    /// A pointer that is controlled by a mouse.
+    ///
+    /// Not a rodent; the device that you move around on a desk.
+    Mouse,
```
ickshonpe, line 78: I think that `PointerKind` is meant to describe the behaviour of the pointer, not the actual input device, but it seems a bit unclear. It would be nice to be able to support things like a "mouse" pointer that's controlled by a joystick or a mouse controlled first person pov with diegetic touch screen UI.

alice-i-cecile: Yeah, the intent is that things like UI libraries will match on this variant, and behave accordingly.

NthTensor:
I don't currently have time for review, but I'd like us to consider that winit is shipping a pointer abstraction in it's next major release that does away with the distinction between mouse and touch entirely.

crates/bevy_picking/src/pointer.rs
cart: I also prefer solution (2). I think thats the move.

alice-i-cecile:
We *could* just do a single u64 then 🤔 I'd prefer to reduce controversy and get this merged though; there's relatively important work (like removing `Interaction`) that this is soft-blocking.

crates/bevy_picking/src/pointer.rs
```diff
@@ -29,41 +27,69 @@ use crate::backend::HitData;
...
             None
         }
     }
 }
 
+/// The input device that created a pointer.
+#[derive(Default, Debug, Clone, Copy, PartialEq, Eq, Hash, Reflect)]
+#[reflect(Default, Debug, Hash, PartialEq)]
+pub enum PointerKind {
+    /// A pointer that is controlled by a mouse.
```
ickshonpe, line 75:
```suggestion
    /// A pointer that behaves like a traditional mouse cursor.
```

crates/bevy_picking/src/pointer.rs
```diff
@@ -29,41 +27,69 @@ use crate::backend::HitData;
...
+    ///
+    /// Not a rodent; the device that you move around on a desk.
+    Mouse,
+    /// A pointer that is controlled by a touch input.
```
ickshonpe, line 79:
```suggestion
    /// A pointer that behaves like it is controlled by touch.
```

Like an obvious thing I think we'd want to support here is touch input emulation with the mouse?

crates/bevy_picking/src/pointer.rs
```diff
@@ -29,41 +27,69 @@ use crate::backend::HitData;
...
-    Touch(u64),
-    /// A custom, uniquely identified pointer. Useful for mocking inputs or implementing a software
-    /// controlled cursor.
-    #[reflect(ignore)]
-    Custom(Uuid),
+pub struct PointerId {
+    /// A stable identifier for a specific pointer.
+    ///
+    /// This is typically provided by winit,
+    /// and is unique for each touch input or mouse.
```
ickshonpe, line 34: The id here is for multitouch corresponding to a specific touch on the device and it's only guaranteed to be unique for the duration of that touch I think. It doesn't feel correct to compare a single touch to a mouse device?

ickshonpe:
> We _could_ just do a single u64 then 🤔 I'd prefer to reduce controversy and get this merged though; there's relatively important work (like removing `Interaction`) that this is soft-blocking.

Maybe we shouldn't have the id here at all? The pointer entity which is spawned the touch starts and despawned after it ends can be used to identify the touch within bevy and have a winit touch id -> pointer entity map somewhere that we can look up when we dispatch events.

colepoirier (commented):
The code looks correct to me, but given the comments of Cart in this PR and Aevyrie in ickshonpe's pr, I don't feel I have the requisite domain knowledge to properly weigh in on this.

PR #17813, opened 2025-02-11
Deterministic fallible_systems example

mockersf:
# Objective

- `fallible_systems` example is not deterministic

## Solution

- Make it deterministic
- Also fix required feature declaration

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -4003,7 +4003,6 @@ name = "Sprite Picking"
 description = "Demonstrates picking sprites and sprite atlases"
 category = "Picking"
 wasm = true
-required-features = ["bevy_sprite_picking_backend"]
 
 [[example]]
 name = "debug_picking"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2159,13 +2159,13 @@ wasm = false
 name = "fallible_systems"
 path = "examples/ecs/fallible_systems.rs"
 doc-scrape-examples = true
+required-features = ["bevy_mesh_picking_backend"]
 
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -2159,13 +2159,13 @@ wasm = false
 name = "fallible_systems"
 path = "examples/ecs/fallible_systems.rs"
 doc-scrape-examples = true
+required-features = ["bevy_mesh_picking_backend"]
```
rparrett, line 2162: Is this needed? Example seems to work on main before this fix and I see `bevy_mesh_picking_backend` in `default`.

rparrett:
Is the point of the sampled points just to add some more `?`able methods to `setup`?

I find it sort of confusing that they are pickable. I went down a tiny rabbit hole trying to figure out why the points sometimes separated from the sphere.

JeanMertz:
> I find it sort of confusing that they are pickable. I went down a tiny rabbit hole trying to figure out why the points sometimes separated from the sphere.

I agree, that is unfortunate. I wanted to have some way to trigger an observer that – under certain conditions – results in an error. This was _a_ way I could think of doing that, but I'm open to changing it to something that better illustrates the point we're trying to make.

JeanMertz (approved):

Cargo.toml
mockersf: strictly speaking, no, it's a default feature

rparrett (approved):

alice-i-cecile (approved):

Issue #10752, opened 2023-11-26
Fetch::init_fetch's use in QueryState::par_for_each_unchecked_manual is unsound.


james7132:
## Bevy version
Potentially every one since QueryState::par_for_each_unchecked_manual was introduced.

## What went wrong
The safety invariant on `Fetch::init_fetch` requires that must be legal for the fetch to access all matched components for a query, which is not true for `QueryState::par_for_each_unchecked_manual`, since it's initializing multiple Fetches in parallel and forcing them to access mutually exclusive parts of the query results. Either:

1. ...the safety invariant needs to be more clearly worded and dependent safety invariants updated.
2.  ...we should be using a more scoped way to fetch query results. 

Given that Fetch implementations are the lowest in the access hierarchy, I'd suggest option 1 and being more stringent in how we pair Fetch initialization and retrieval.

PR #17815, opened 2025-02-12
implement par_iter_many and par_iter_many_unique

Victoronz:
# Objective

Continuation of #16547.

We do not yet have parallel versions of `par_iter_many` and `par_iter_many_unique`. It is currently very painful to try and use parallel iteration over entity lists. Even if a list is not long, each operation might still be very expensive, and worth parallelizing.
Plus, it has been requested several times!

## Solution

Once again, we implement what we lack!

These parallel iterators collect their input entity list into a `Vec`/`UniqueEntityVec`, then chunk that over the available threads, inspired by the original `par_iter`.  

Since no order guarantee is given to the caller, we could sort the input list according to `EntityLocation`, but that would likely only be worth it for very large entity lists.

There is some duplication which could likely be improved, but I'd like to leave that for a follow-up.

## Testing

The doc tests on `for_each_init` of `QueryParManyIter` and `QueryParManyUniqueIter`.

### DIFF
--- a/crates/bevy_ecs/src/query/par_iter.rs
+++ b/crates/bevy_ecs/src/query/par_iter.rs
@@ -59,12 +59,12 @@ impl<'w, 's, D: QueryData, F: QueryFilter> QueryParIter<'w, 's, D, F> {
     /// fn system(query: Query<&T>){
     ///     let mut queue: Parallel<usize> = Parallel::default();
     ///     // queue.borrow_local_mut() will get or create a thread_local queue for each task/thread;
-    ///     query.par_iter().for_each_init(|| queue.borrow_local_mut(),|local_queue,item| {
+    ///     query.par_iter().for_each_init(|| queue.borrow_local_mut(),|local_queue, item| {
     ///         **local_queue += 1;
     ///      });
     ///     
     ///     // collect value from every thread
-    ///     let entity_count: usize = queue.iter_mut().map(|v| *v).sum();
+    ///     let entity_count = queue.iter().copied().sum();
     /// }
     /// ```
     ///
@@ -202,6 +202,7 @@ impl<'w, 's, D: ReadOnlyQueryData, F: QueryFilter, E: EntityBorrow + Sync>
     /// ```
     /// use bevy_utils::Parallel;
     /// use crate::{bevy_ecs::prelude::{Component, Res, Resource, Entity}, bevy_ecs::system::Query};
+    /// # use core::slice;
     /// use bevy_platform_support::prelude::Vec;
     /// # fn some_expensive_operation(item: T) -> usize {
     /// #     0
@@ -211,9 +212,9 @@ impl<'w, 's, D: ReadOnlyQueryData, F: QueryFilter, E: EntityBorrow + Sync>
     /// struct T;
     ///
     /// #[derive(Resource)]
-    /// struct V(Vec<Entity>)
+    /// struct V(Vec<Entity>);
     ///
-    /// impl<'a> IntoIterator for &V {
+    /// impl<'a> IntoIterator for &'a V {
     /// // ...
     /// #   type Item = &'a Entity;
     /// #   type IntoIter = slice::Iter<'a, Entity>;
@@ -231,7 +232,7 @@ impl<'w, 's, D: ReadOnlyQueryData, F: QueryFilter, E: EntityBorrow + Sync>
     ///     });
     ///     
     ///     // collect value from every thread
-    ///     let entity_count: usize = queue.iter_mut().map(|v| *v).sum();
+    ///     let final_value = queue.iter().copied().sum();
     /// }
     /// ```
     ///
@@ -368,7 +369,8 @@ impl<'w, 's, D: QueryData, F: QueryFilter, E: TrustedEntityBorrow + Sync>
     ///
     /// ```
     /// use bevy_utils::Parallel;
-    /// use crate::{bevy_ecs::prelude::{Component, Res, Entity}, bevy_ecs::{entity::UniqueEntityVec, system::Query}};
+    /// use crate::{bevy_ecs::prelude::{Component, Res, Resource, Entity}, bevy_ecs::{entity::UniqueEntityVec, system::Query}};
+    /// # use core::slice;
     /// # fn some_expensive_operation(item: T) -> usize {
     /// #     0
     /// # }
@@ -377,9 +379,9 @@ impl<'w, 's, D: QueryData, F: QueryFilter, E: TrustedEntityBorrow + Sync>
     /// struct T;
     ///
     /// #[derive(Resource)]
-    /// struct V(UniqueEntityVec<Entity>)
+    /// struct V(UniqueEntityVec<Entity>);
     ///
-    /// impl<'a> IntoIterator for &V {
+    /// impl<'a> IntoIterator for &'a V {
     /// // ...
     /// #   type Item = &'a Entity;
     /// #   type IntoIter = UniqueEntityVec<slice::Iter<'a, Entity>>;
@@ -397,7 +399,7 @@ impl<'w, 's, D: QueryData, F: QueryFilter, E: TrustedEntityBorrow + Sync>
     ///     });
     ///     
     ///     // collect value from every thread
-    ///     let entity_count: usize = queue.iter_mut().map(|v| *v).sum();
+    ///     let final_value = queue.iter().copied().sum();
     /// }
     /// ```
     ///
--- a/crates/bevy_ecs/src/query/par_iter.rs
+++ b/crates/bevy_ecs/src/query/par_iter.rs
@@ -201,15 +201,29 @@ impl<'w, 's, D: ReadOnlyQueryData, F: QueryFilter, E: EntityBorrow + Sync>
     ///
     /// ```
     /// use bevy_utils::Parallel;
-    /// use crate::{bevy_ecs::prelude::{Component, Res, Entity}, bevy_ecs::system::Query};
-    /// use std::collections::Vec;
+    /// use crate::{bevy_ecs::prelude::{Component, Res, Resource, Entity}, bevy_ecs::system::Query};
+    /// use bevy_platform_support::prelude::Vec;
     /// # fn some_expensive_operation(item: T) -> usize {
     /// #     0
     /// # }
     ///
     /// #[derive(Component)]
     /// struct T;
-    /// fn system(query: Query<&T>, entities: Res<Vec<Entity>>){
+    ///
+    /// #[derive(Resource)]
+    /// struct V(Vec<Entity>)
+    ///
+    /// impl<'a> IntoIterator for &V {
+    /// // ...
+    /// #   type Item = &'a Entity;
+    /// #   type IntoIter = slice::Iter<'a, Entity>;
+    /// #
+    /// #    fn into_iter(self) -> Self::IntoIter {
+    /// #        self.0.iter()
+    /// #    }
+    /// }
+    ///
+    /// fn system(query: Query<&T>, entities: Res<V>){
     ///     let mut queue: Parallel<usize> = Parallel::default();
     ///     // queue.borrow_local_mut() will get or create a thread_local queue for each task/thread;
     ///     query.par_iter_many(entities).for_each_init(|| queue.borrow_local_mut(),|local_queue, item| {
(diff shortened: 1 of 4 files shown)


crates/bevy_ecs/src/query/par_iter.rs
```diff
@@ -146,3 +151,292 @@ impl<'w, 's, D: QueryData, F: QueryFilter> QueryParIter<'w, 's, D, F> {
...
+/// A parallel iterator over the unique query items generated from an [`Entity`] list.
+///
+/// This struct is created by the [`Query::par_iter_many`] method.
+///
+/// [`Entity`]: crate::entity::Entity
+/// [`Query::par_iter_many`]: crate::system::Query::par_iter_many
+pub struct QueryParManyIter<'w, 's, D: QueryData, F: QueryFilter, E: EntityBorrow> {
+    pub(crate) world: UnsafeWorldCell<'w>,
+    pub(crate) state: &'s QueryState<D, F>,
+    pub(crate) entity_list: Vec<E>,
```
chescock, line 164: Could this be generalized to `EntityList: AsRef<[E]>`?  (And similarly with `AsRef<UniqueEntitySlice<E>>` for `QueryParManyUniqueIter`.)  That would let you use it for borrowed slices without needing to pass ownership of the `Vec`, as well as allowing some other collections like `SmallVec`.

crates/bevy_ecs/src/query/state.rs
```diff
@@ -1868,6 +1871,126 @@ impl<D: QueryData, F: QueryFilter> QueryState<D, F> {
...
+                    let accum = init_accum();
+                    self.iter_many_unique_unchecked_manual(batch, world, last_run, this_run)
+                        .fold(accum, &mut func);
+                });
+            }
+
+            #[cfg(feature = "trace")]
+            let _span = self.par_iter_span.enter();
+            let accum = init_accum();
+            self.iter_many_unique_unchecked_manual(&entity_list[rem..], world, last_run, this_run)
```
chescock, line 1927: Rather than calculate `rem` yourself, you could use `ChunksExact::remainder()`.  A method I learned about yesterday when reviewing your implementation of it for `UniqueEntitySliceIter` :).

crates/bevy_ecs/src/query/state.rs
```diff
@@ -1868,6 +1871,126 @@ impl<D: QueryData, F: QueryFilter> QueryState<D, F> {
...
+            let rem = len - len % batch_size;
+
+            for batch in entity_list.chunks_exact(batch_size) {
+                let mut func = func.clone();
+                let init_accum = init_accum.clone();
+                scope.spawn(async move {
+                    #[cfg(feature = "trace")]
+                    let _span = self.par_iter_span.enter();
+                    let accum = init_accum();
+                    self.iter_many_unique_unchecked_manual(batch, world, last_run, this_run)
```
chescock, line 1919: I think we want `SAFETY` comments here (and in the non-`unique` version) for when we turn on `unsafe_op_in_unsafe_fn`.

crates/bevy_ecs/src/query/par_iter.rs
Victoronz: This isn't part of the public API, this is a `Vec` that input iterators are collected into because storing an iterator or some other non-owned type here would make the lifetimes and bounds very complicated.
The reason we specifically use `Vec` here is to take advantage of an optimization that exists in Rust; When `Vec::into_iter` is collected into a `Vec` again, it becomes practically a no-op.

crates/bevy_ecs/src/query/state.rs
Victoronz: Sweet, done :)

crates/bevy_ecs/src/query/par_iter.rs
chescock: Right, I think I'm asking why it does a `collect()` at all and doesn't just keep the original `entities: EntityList` like the non-`Par` versions.  

I guess that's because the implementation requires a slice, but you want to accept an `EntitySet` that isn't a slice?  

Are we worried that the `collect()` is a performance risk?  Even if you have a `Vec`, you might want to only pass a reference to it, so that you can use it later.  And going from a `&Vec` to a `Vec` isn't a no-op.  Parallel iteration is mainly useful when you have a large list of entities, so that might be an expensive `Vec` to copy!

crates/bevy_ecs/src/query/state.rs
Victoronz: Hmm, we can't actually write proper safety comments here right now I just noticed, because the original `par_iter` is itself unsound: https://github.com/bevyengine/bevy/issues/10752
Can we punt on this, and fix all `par_iter` methods at once?

crates/bevy_ecs/src/query/par_iter.rs
Victoronz: It is an expense we'd rather not have, however not doing it would make these methods very complex, which shouldn't be done in the first PR imo.
By collecting the input iterator into a `Vec` we ensure that we know the exact length the list has so we can chunk most effectively.
By owning the type also do not need to worry about splitting up the same iterator across several threads, which would require several `Send`/`Sync` bounds on the iterator and its items.
I did try, but couldn't get it to work, so I left it for later.

crates/bevy_ecs/src/query/par_iter.rs
```diff
@@ -146,3 +151,292 @@ impl<'w, 's, D: QueryData, F: QueryFilter> QueryParIter<'w, 's, D, F> {
...
+    ///
+    /// ```
+    /// use bevy_utils::Parallel;
+    /// use crate::{bevy_ecs::prelude::Component, bevy_ecs::system::Query};
+    /// #[derive(Component)]
+    /// struct T;
+    /// fn system(query: Query<&T>){
+    ///     let mut queue: Parallel<usize> = Parallel::default();
+    ///     // queue.borrow_local_mut() will get or create a thread_local queue for each task/thread;
+    ///     query.par_iter().for_each_init(|| queue.borrow_local_mut(),|local_queue,item| {
```
cBournhonesque, line 210: Should the docstrig use `par_iter_many`?

crates/bevy_ecs/src/query/par_iter.rs
```diff
@@ -146,3 +151,292 @@ impl<'w, 's, D: QueryData, F: QueryFilter> QueryParIter<'w, 's, D, F> {
...
+    {
+        let func = |mut init, item| {
+            func(&mut init, item);
+            init
+        };
+        #[cfg(any(target_arch = "wasm32", not(feature = "multi_threaded")))]
+        {
+            let init = init();
+            // SAFETY:
+            // This method can only be called once per instance of QueryParIter,
```
cBournhonesque, line 238: QueryParManyIter

crates/bevy_ecs/src/query/par_iter.rs
Victoronz: Ah, nice catch, I forgot to update these.

crates/bevy_ecs/src/system/query.rs
```diff
@@ -1082,6 +1082,94 @@ impl<'w, 's, D: QueryData, F: QueryFilter> Query<'w, 's, D, F> {
...
     }
 
+    /// Returns a parallel iterator over the read-only query items generated from an [`Entity`] list.
+    ///
+    /// Entities that don't match the query are skipped. Iteration order and thread assignment is not guaranteed.
+    ///
+    /// If the `multithreaded` feature is disabled, iterating with this operates identically to [`Iterator::for_each`]
+    /// on [`QueryManyIter`].
+    ///
+    /// This can only be called for read-only queries, there is no `par_iter_many_mut` equivalent.
```
cBournhonesque, line 1092: /// This can only be called for read-only queries, there is no `par_iter_many_mut` equivalent.
```suggestion
    /// This can only be called for read-only queries, there is no `par_iter_many_mut` equivalent to avoid potential aliasing
```

crates/bevy_ecs/src/query/par_iter.rs
```diff
@@ -146,3 +151,292 @@ impl<'w, 's, D: QueryData, F: QueryFilter> QueryParIter<'w, 's, D, F> {
...
     }
 }
+
+/// A parallel iterator over the unique query items generated from an [`Entity`] list.
+///
+/// This struct is created by the [`Query::par_iter_many`] method.
+///
+/// [`Entity`]: crate::entity::Entity
+/// [`Query::par_iter_many`]: crate::system::Query::par_iter_many
+pub struct QueryParManyIter<'w, 's, D: QueryData, F: QueryFilter, E: EntityBorrow> {
```
cBournhonesque, line 161: Out of curiosity, why not put a ReadOnlyQueryData bound here? 
Because it's good practice to only put bounds on `impl` instead of on the scope itself?

crates/bevy_ecs/src/query/par_iter.rs
```diff
@@ -146,3 +151,292 @@ impl<'w, 's, D: QueryData, F: QueryFilter> QueryParIter<'w, 's, D, F> {
...
+    ///
+    /// ```
+    /// use bevy_utils::Parallel;
+    /// use crate::{bevy_ecs::prelude::Component, bevy_ecs::system::Query};
+    /// #[derive(Component)]
+    /// struct T;
+    /// fn system(query: Query<&T>){
+    ///     let mut queue: Parallel<usize> = Parallel::default();
+    ///     // queue.borrow_local_mut() will get or create a thread_local queue for each task/thread;
+    ///     query.par_iter().for_each_init(|| queue.borrow_local_mut(),|local_queue,item| {
```
cBournhonesque, line 358: same comments as above

crates/bevy_ecs/src/query/state.rs
```diff
@@ -1868,6 +1871,126 @@ impl<D: QueryData, F: QueryFilter> QueryState<D, F> {
...
+        mut func: FN,
+        last_run: Tick,
+        this_run: Tick,
+    ) where
+        FN: Fn(T, D::Item<'w>) -> T + Send + Sync + Clone,
+        INIT: Fn() -> T + Sync + Send + Clone,
+        E: TrustedEntityBorrow + Sync,
+    {
+        // NOTE: If you are changing query iteration code, remember to update the following places, where relevant:
+        // QueryIter, QueryIterationCursor, QueryManyIter, QueryCombinationIter,QueryState::par_fold_init_unchecked_manual
```
cBournhonesque, line 1906: Some things here need to be updated

crates/bevy_ecs/src/query/state.rs
```diff
@@ -1868,6 +1871,126 @@ impl<D: QueryData, F: QueryFilter> QueryState<D, F> {
...
+        mut func: FN,
+        last_run: Tick,
+        this_run: Tick,
+    ) where
+        FN: Fn(T, D::Item<'w>) -> T + Send + Sync + Clone,
+        INIT: Fn() -> T + Sync + Send + Clone,
+        E: EntityBorrow + Sync,
+    {
+        // NOTE: If you are changing query iteration code, remember to update the following places, where relevant:
+        // QueryIter, QueryIterationCursor, QueryManyIter, QueryCombinationIter,QueryState::par_fold_init_unchecked_manual
```
cBournhonesque, line 1966: Same here

cBournhonesque (approved):
Amazing to see the results of the EntitySet work!

crates/bevy_ecs/src/query/par_iter.rs
Victoronz: Yes!

crates/bevy_ecs/src/query/par_iter.rs
```diff
@@ -146,3 +151,329 @@ impl<'w, 's, D: QueryData, F: QueryFilter> QueryParIter<'w, 's, D, F> {
...
+    /// [`ComputeTaskPool`]: bevy_tasks::ComputeTaskPool
+    #[inline]
+    pub fn for_each<FN: Fn(QueryItem<'w, D>) + Send + Sync + Clone>(self, func: FN) {
+        self.for_each_init(|| {}, |_, item| func(item));
+    }
+
+    /// Runs `func` on each query result in parallel on a value returned by `init`.
+    ///
+    /// `init` may be called multiple times per thread, and the values returned may be discarded between tasks on any given thread.
+    /// Callers should avoid using this function as if it were a parallel version
```
Carter0, line 197: This might be worth elaborating more on

crates/bevy_ecs/src/query/par_iter.rs
```diff
@@ -146,3 +151,329 @@ impl<'w, 's, D: QueryData, F: QueryFilter> QueryParIter<'w, 's, D, F> {
...
+    /// of [`Iterator::fold`].
```
Carter0, line 198: What do you mean by this? Can you elaborate?

Carter0 (commented):
Yeah this is a little above me lol. If you added some tests that might help tho

crates/bevy_ecs/src/query/par_iter.rs
Victoronz: This is has not changed from `QueryParIter`; `for_each_init` has similar inputs to `Iterator::fold`, but must not be mistaken for fold, it doesn't accumulate one overall value.

crates/bevy_ecs/src/query/par_iter.rs
chescock: Oh, I wasn't proposing trying to split up an arbitrary `Iterator`!  I agree that would be too complex.  

I'm proposing `pub fn par_iter_many<E: EntityBorrow, EntityList: AsRef<[E]>>`, and then storing the `entities: EntityList` directly.  So, you could only pass an owned or borrowed slice, and not an arbitrary `Iterator`.  

If you had some other `Iterator` then you'd have to `collect()` it yourself.  That's less convenient, but it makes the cost more visible.  

But in return, you could pass `&Vec<Entity>` without consuming or copying the `Vec`!  And you could still pass an owned `Vec` if you needed a `QueryParManyIter` that didn't borrow from anything.

Victoronz:
oops, that was too early!

chescock (approved):

chescock:
> oops, that was too early!

Only because I had forgotten to actually hit the approve button :)

PR #17817, opened 2025-02-12
Add `TypeRegistry::register_by_val`

NyxAlexandra:
# Objective

It is impossible to register a type with `TypeRegistry::register` if the type is unnameable (in the current scope).

## Solution

Add `TypeRegistry::register_by_val` which mirrors std's `size_of_val` and friends.

## Testing

There's a doc test (unrelated but there seem to be some pre-existing broken doc links in `bevy_reflect`).

### DIFF
--- a/crates/bevy_reflect/src/type_registry.rs
+++ b/crates/bevy_reflect/src/type_registry.rs
@@ -169,6 +169,8 @@ impl TypeRegistry {
     ///
     /// See [`register`] for more details.
     ///
+    /// # Example
+    ///
     /// ```
     /// # use bevy_reflect::{Reflect, TypeRegistry};
     /// # use core::any::TypeId;
--- a/crates/bevy_reflect/src/type_registry.rs
+++ b/crates/bevy_reflect/src/type_registry.rs
@@ -165,6 +165,41 @@ impl TypeRegistry {
         }
     }
 
+    /// Attempts to register the referenced type `T` if it has not yet been registered.
+    ///
+    /// See [`register`] for more details.
+    ///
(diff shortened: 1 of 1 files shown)


crates/bevy_reflect/src/type_registry.rs
```diff
@@ -165,6 +165,41 @@ impl TypeRegistry {
         }
     }
 
+    /// Attempts to register the referenced type `T` if it has not yet been registered.
+    ///
+    /// See [`register`] for more details.
+    ///
```
MrGVSV, line 171: Nit:
```suggestion
    ///
    /// # Example
    ///
```

MrGVSV (approved):

MrGVSV:
> I'm wondering, what do you do with the type, once it's registered? [get_with_type_path](https://docs.rs/bevy/latest/bevy/reflect/struct.TypeRegistry.html#method.get_with_type_path) doesn't work, because it can't be named and thus not converted to a string.

You can actually. We have dynamic methods for accessing type paths such as `DynamicTypePath::reflect_type_path`. And while these types are "unnameable", Rust will still give them names with `std::any::type_name[_of_val]`, which is the default behavior for `TypePath`. And users can always hard code a type path as well.

> This means that the only way to look it up is to do it by TypeId, and that's not really useful for most purposes (such as scripting).

The idea is that this could be used for types like `Foo<T>` where `T` is an unnameable type, such as a function. You can still access its type path and `TypeId`, but you can't name it for direct `register` calls.

alice-i-cecile (approved):

Issue #16929, opened 2024-12-22
Bevy Panics when trying to load a single mesh from gltf file containing armature


PatrickChodowski:
## Bevy version
0.15 (happened on 0.14 too)

## Relevant system information

os: "Linux 22.04 Ubuntu"
kernel: "6.8.0-49-generic"
cpu: "Intel(R) Core(TM) i7-7700HQ CPU @ 2.80GHz"
core_count: "4"
memory: "15.5 GiB" 
AdapterInfo { name: "NVIDIA GeForce GTX 1050", vendor: 4318, device: 7309, device_type: DiscreteGpu, driver: "NVIDIA", driver_info: "535.183.01", backend: Vulkan}

## What you did
Tried to use specific mesh/primitive from Gltf file, that also contains armature inside

## What went wrong

Received wgpu error:

```
2024-12-22T13:38:16.509152Z ERROR wgpu::backend::wgpu_core: Handling wgpu errors as fatal by default    
thread 'Compute Task Pool (2)' panicked at /home/patrick/.cargo/registry/src/index.crates.io-6f17d22bba15001f/wgpu-23.0.1/src/backend/wgpu_core.rs:2976:18:
wgpu error: Validation Error

Caused by:
  In RenderPass::end
    In a draw command, kind: Draw
      The BindGroupLayout with 'mesh_layout' label of current set BindGroup with 'model_only_mesh_bind_group' label at index 1 is not compatible with the corresponding BindGroupLayout with 'skinned_mesh_layout' label of RenderPipeline with 'pbr_opaque_mesh_pipeline' label
        Expected entry with binding 1 not found in assigned bind group layout


note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
Encountered a panic in system `bevy_render::renderer::render_system`!
thread 'main' panicked at /home/patrick/.cargo/registry/src/index.crates.io-6f17d22bba15001f/wgpu-hal-23.0.1/src/vulkan/instance.rs:173:58:
Trying to destroy a SurfaceSemaphores that is still in use by a SurfaceTexture
```

## Additional information

[Minimal reproduction app](https://github.com/PatrickChodowski/gltf_bug)  (you can comment and uncomment parts to test)

It contains 2 assets: They contain same mesh and material but one of the files also contains armature. File **without** armature loads correctly inside bevy, but the one **with** armature loads only as a scene. I dont like using scenes much, so I tried extracting mesh and material manually but it causes game to crash

[...]

PR #17818, opened 2025-02-12
Retain skins from frame to frame.

pcwalton:
Currently, Bevy rebuilds the buffer containing all the transforms for joints every frame, during the extraction phase. This is inefficient in cases in which many skins are present in the scene and their joints don't move, such as the Caldera test scene.

To address this problem, this commit switches skin extraction to use a set of retained GPU buffers with allocations managed by the offset allocator. I use fine-grained change detection in order to determine which skins need updating. Note that the granularity is on the level of an entire skin, not individual joints. Using the change detection at that level would yield poor performance in common cases in which an entire skin is animated at once. Also, this patch yields additional performance from the fact that changing joint transforms no longer requires the skinned mesh to be re-extracted.

Note that this optimization can be a double-edged sword. In `many_foxes`, fine-grained change detection regressed the performance of `extract_skins` by 3.4x. This is because every joint is updated every frame in that example, so change detection is pointless and is pure overhead. Because the `many_foxes` workload is actually representative of animated scenes, this patch includes a heuristic that disables fine-grained change detection if the number of transformed entities in the frame exceeds a certain fraction of the total number of joints. Currently, this threshold is set to 25%. Note that this is a crude heuristic, because it doesn't distinguish between the number of transformed *joints* and the number of transformed *entities*; however, it should be good enough to yield the optimum code path most of the time.

Finally, this patch fixes a bug whereby skinned meshes are actually being incorrectly retained if the buffer offsets of the joints of those skinned meshes changes from frame to frame. To fix this without retaining skins, we would have to re-extract every skinned mesh every frame. Doing this was a significant regression on Caldera. With this PR, by contrast, mesh joints stay at the same buffer offset, so we don't have to update the `MeshInputUniform` containing the buffer offset every frame. This also makes PR #17717 easier to implement, because that PR uses the buffer offset from the previous frame, and the logic for calculating that is simplified if the previous frame's buffer offset is guaranteed to be identical to that of the current frame.

On Caldera, this patch reduces the time spent in `extract_skins` from 1.79 ms to near zero. On `many_foxes`, this patch regresses the performance of `extract_skins` by approximately 10%-25%, depending on the number of foxes. This has only a small impact on frame rate.

### DIFF
--- a/crates/bevy_pbr/src/render/mesh.rs
+++ b/crates/bevy_pbr/src/render/mesh.rs
@@ -1,4 +1,7 @@
-use crate::material_bind_groups::{MaterialBindGroupIndex, MaterialBindGroupSlot};
+use crate::{
+    material_bind_groups::{MaterialBindGroupIndex, MaterialBindGroupSlot},
+    skin::mark_skins_as_changed_if_their_inverse_bindposes_changed,
+};
 use allocator::MeshAllocator;
 use bevy_asset::{load_internal_asset, AssetId, UntypedAssetId};
 use bevy_core_pipeline::{
@@ -26,7 +29,7 @@ use bevy_render::{
         no_gpu_preprocessing, GetBatchData, GetFullBatchData, NoAutomaticBatching,
     },
     camera::Camera,
-    mesh::*,
+    mesh::{skinning::SkinnedMesh, *},
     primitives::Aabb,
     render_asset::RenderAssets,
     render_phase::{
@@ -46,7 +49,7 @@ use bevy_transform::components::GlobalTransform;
 use bevy_utils::{default, Parallel};
 use core::mem::size_of;
 use material_bind_groups::MaterialBindingId;
-use render::skin::{self, SkinIndex};
+use render::skin;
 use tracing::{error, warn};
 
 use self::irradiance_volume::IRRADIANCE_VOLUMES_ARE_USABLE;
@@ -163,7 +166,11 @@ impl Plugin for MeshRenderPlugin {
 
         app.add_systems(
             PostUpdate,
-            (no_automatic_skin_batching, no_automatic_morph_batching),
+            (
+                no_automatic_skin_batching,
+                no_automatic_morph_batching,
+                mark_skins_as_changed_if_their_inverse_bindposes_changed,
+            ),
         )
         .add_plugins((
             BinnedRenderPhasePlugin::<Opaque3d, MeshPipeline>::default(),
@@ -178,7 +185,6 @@ impl Plugin for MeshRenderPlugin {
         if let Some(render_app) = app.get_sub_app_mut(RenderApp) {
             render_app
                 .init_resource::<MeshBindGroups>()
-                .init_resource::<SkinIndices>()
                 .init_resource::<MorphUniforms>()
                 .init_resource::<MorphIndices>()
                 .init_resource::<MeshCullingDataBuffer>()
@@ -478,8 +484,6 @@ pub struct MeshUniform {
     pub first_vertex_index: u32,
     /// The current skin index, or `u32::MAX` if there's no skin.
     pub current_skin_index: u32,
-    /// The previous skin index, or `u32::MAX` if there's no previous skin.
-    pub previous_skin_index: u32,
     /// The material and lightmap indices, packed into 32 bits.
     ///
     /// Low 16 bits: index of the material inside the bind group data.
@@ -487,6 +491,8 @@ pub struct MeshUniform {
     pub material_and_lightmap_bind_group_slot: u32,
     /// User supplied tag to identify this mesh instance.
     pub tag: u32,
+    /// Padding.
+    pub pad: u32,
 }
 
 /// Information that has to be transferred from CPU to GPU in order to produce
@@ -536,8 +542,6 @@ pub struct MeshInputUniform {
     pub index_count: u32,
     /// The current skin index, or `u32::MAX` if there's no skin.
     pub current_skin_index: u32,
-    /// The previous skin index, or `u32::MAX` if there's no previous skin.
-    pub previous_skin_index: u32,
     /// The material and lightmap indices, packed into 32 bits.
     ///
     /// Low 16 bits: index of the material inside the bind group data.
@@ -546,7 +550,9 @@ pub struct MeshInputUniform {
     /// User supplied tag to identify this mesh instance.
     pub tag: u32,
     /// Padding.
-    pub pad: u32,
+    pub pad_a: u32,
+    /// Padding.
+    pub pad_b: u32,
 }
 
 /// Information about each mesh instance needed to cull it on GPU.
@@ -579,7 +585,6 @@ impl MeshUniform {
         material_bind_group_slot: MaterialBindGroupSlot,
         maybe_lightmap: Option<(LightmapSlotIndex, Rect)>,
         current_skin_index: Option<u32>,
-        previous_skin_index: Option<u32>,
         tag: Option<u32>,
     ) -> Self {
         let (local_from_world_transpose_a, local_from_world_transpose_b) =
@@ -598,10 +603,10 @@ impl MeshUniform {
             flags: mesh_transforms.flags,
             first_vertex_index,
             current_skin_index: current_skin_index.unwrap_or(u32::MAX),
-            previous_skin_index: previous_skin_index.unwrap_or(u32::MAX),
             material_and_lightmap_bind_group_slot: u32::from(material_bind_group_slot)
                 | ((lightmap_bind_group_slot as u32) << 16),
             tag: tag.unwrap_or(0),
+            pad: 0,
         }
     }
 }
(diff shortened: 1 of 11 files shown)


greeble-dev:
I hit a panic if I force `skins_use_uniform_buffers` to return true. I'm not sure if it's actually valid to force that on, but flagging it up just in case. 

Repro: `cargo run --example custom_skinned_mesh`. Does not occur in `many_foxes` or `animated_mesh` examples. Does not occur if I roll back to before the PR (153ce468efb1d596c744378be6bdb1a83dea05b8).

```
Caused by:
  In RenderPass::end
    In a set_bind_group command
      Dynamic binding offset index 0 with offset 2048 would overrun the buffer bound to BindGroup with 'skinned_mesh_bind_group' label 1 -> binding 1. Buffer size is 16384 bytes, the binding binds bytes 0..16384, meaning the maximum the binding can be offset is 0 bytes

SystemInfo { os: "Windows 10 Pro", kernel: "19045" }
AdapterInfo { name: "NVIDIA GeForce GT 1030", vendor: 4318, device: 7425, device_type: DiscreteGpu, driver: "NVIDIA", driver_info: "560.94", backend: Vulkan }
```

crates/bevy_pbr/src/render/skin.rs
```diff
@@ -24,17 +30,39 @@ use bevy_transform::prelude::GlobalTransform;
 /// of the GPU at runtime, which would mean not using consts anymore.
 pub const MAX_JOINTS: usize = 256;
 
+/// The total number of joints we support.
+///
+/// This is 256 GiB worth of joint matrices, which we will never hit under any
+/// reasonable circumstances.
+const MAX_TOTAL_JOINTS: u32 = 1024 * 1024 * 1024;
```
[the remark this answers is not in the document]
pcwalton: You're reading more into it than I intended :) It's actually a completely arbitrary number, set to be "higher than you'll ever need but not so high as to be absurd", so as to catch bugs before OOM happens.

crates/bevy_pbr/src/render/skin.rs
```diff
@@ -24,17 +30,39 @@ use bevy_transform::prelude::GlobalTransform;
...
+const JOINTS_PER_ALLOCATION_UNIT: u32 = (256 / size_of::<Mat4>()) as u32;
+
+/// The maximum ratio of the number of entities whose transforms changed to the
+/// total number of joints before we re-extract all joints.
+///
+/// We use this as a heuristic to decide whether it's worth switching over to
+/// fine-grained detection to determine which skins need extraction. If the
+/// number of changed entities is over this threshold, we skip change detection
+/// and simply re-extract the transforms of all joints.
+const JOINT_EXTRACTION_THRESHOLD_FACTOR: f64 = 0.25;
```
tychedelia, line 53: This heuristic intuitively makes sense to me, but I wonder if it should be exposed as configuration. Maybe something for the future with more testing to understand the balance here. I also hope that our change detection (particularly w.r.t. assets) will become cheaper in the future.

crates/bevy_pbr/src/render/skin.rs
```diff
@@ -150,70 +277,334 @@ pub fn prepare_skins(
...
+    skin_uniforms.total_joints -= old_skin_uniform_info.joints.len();
+}
+
+/// Marks [`SkinnedMesh`] components as changed if the
+/// [`SkinnedMeshInverseBindposes`] assets that they refer to changed.
+///
+/// This is needed to ensure that we pick up the new inverse bindposes in
+/// [`extract_skins`].
+pub fn mark_skins_as_changed_if_their_inverse_bindposes_changed(
+    mut events: EventReader<AssetEvent<SkinnedMeshInverseBindposes>>,
```
tychedelia, line 586: If you impl `AsAssetId` for `SinnedMesh` , you can use the new `AssetChanged<SkinnedMesh>` filter. Since you're already doing the table scan, this should be pretty much equivalent in performance, I believe.

tychedelia (approved):
Lgtm other than the note about using the new asset filter.

crates/bevy_pbr/src/render/skin.rs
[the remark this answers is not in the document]
greeble-dev: Hah, that makes sense. I misunderstood what the allocator was doing.

I think it will panic a while before this due to `byte_offset: u32`? But doesn't seem like a pressing issue.

crates/bevy_pbr/src/render/mesh.rs
```diff
@@ -2660,44 +2660,40 @@ pub fn prepare_mesh_bind_group(
     // Create the skinned mesh bind group with the current and previous buffers
     // (the latter being for motion vector computation). If there's no previous
     // buffer, just use the current one as the shader will ignore it.
```
greeble-dev, line 2662: Comment seems out of date.

```suggestion
    // (the latter being for motion vector computation).
```

crates/bevy_pbr/src/render/mesh.rs
```diff
@@ -2864,14 +2859,11 @@ impl<P: PhaseItem, const I: usize> RenderCommand<P> for SetMeshBindGroup<I> {
         if has_motion_vector_prepass {
             // Attach the previous skin index for motion vector computation. If
             // there isn't one, just use zero as the shader will ignore it.
```
greeble-dev, line 2861:
```suggestion
            // Attach the current skin index for motion vector computation.
```

greeble-dev (approved):
Clicking the approve button with some minor suggestions, but take with salt as I don't have much renderer experience.

In case it's relevant, I'm planning to look at #16929 once things have calmed down, and that will involve adding a test for skinning corner cases (missing joints, adding/removing meshes). So that might shake out any subtle issues with this PR.

crates/bevy_pbr/src/render/skin.rs
pcwalton: Do you mean plugin configuration?

tychedelia: No, I agree we're fine now, let's just revisit this heuristic if, e.g., asset change detection improves in the future.

crates/bevy_pbr/src/render/skin.rs
pcwalton: Nice! Made that change.

crates/bevy_pbr/src/render/skin.rs
```diff
@@ -24,17 +30,39 @@ use bevy_transform::prelude::GlobalTransform;
...
+#[derive(Clone, Copy)]
+pub struct SkinByteOffset {
     /// The byte offset of the first joint matrix.
     pub byte_offset: u32,
 }
 
-impl SkinIndex {
+impl SkinByteOffset {
     /// Index to be in address space based on the size of a skin uniform.
     const fn new(start: usize) -> Self {
```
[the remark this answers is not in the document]
pcwalton: Fixed.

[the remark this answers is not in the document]
pcwalton: Fixed.

crates/bevy_pbr/src/render/skin.rs
```diff
@@ -104,20 +196,54 @@ pub fn skins_use_uniform_buffers(render_device: &RenderDevice) -> bool {
...
-        .current_buffer
-        .write_buffer(&render_device, &render_queue);
+    // Swap current and previous buffers.
+    mem::swap(&mut uniform.current_buffer, &mut uniform.prev_buffer);
+
+    // Resize the buffer if necessary. Include extra space equal to `MAX_JOINTS`
+    // because we need to be able to bind a full uniform buffer's worth of data
+    // if skins use uniform buffers on this platform.
+    let needed_size = (uniform.current_staging_buffer.len() as u64 + MAX_JOINTS as u64)
+        * size_of::<Mat4>() as u64;
```
superdump, line 218: Couldn't it just be `.max(MAX_JOINTS)`?

crates/bevy_pbr/src/render/skin.rs
```diff
@@ -104,20 +196,54 @@ pub fn skins_use_uniform_buffers(render_device: &RenderDevice) -> bool {
...
+    if uniform.current_buffer.size() < needed_size {
+        let mut new_size = uniform.current_buffer.size();
+        while new_size < needed_size {
+            // 1.5× growth factor.
+            new_size += new_size / 2;
```
superdump, line 223: For safety?
```suggestion
            new_size += (new_size / 2).max(1);
```

superdump (approved):
LGTM aside from a couple of minor, possibly irrelevant nits.

crates/bevy_pbr/src/render/skin.rs
pcwalton: The issue is that, in CPU-driven mode, the uniform buffer is a sliding window that has to be exactly `MAX_JOINTS` in size. So if the start of the window is placed at the end, then we have to have `MAX_JOINTS - 1` extra entities at the end. I guess this can actually be `uniform.current_starting_buffer.len().max(1) as u64 + MAX_JOINTS as u64 - 1)`.

superdump: This is fine enough.

PR #17819, opened 2025-02-12
Switch bins from parallel key/value arrays to `IndexMap`s.

pcwalton:
Conceptually, bins are ordered hash maps. We currently implement these as a list of keys with an associated hash map. But we already have a data type that implements ordered hash maps directly: `IndexMap`. This patch switches Bevy to use `IndexMap`s for bins. Because we're memory bound, this doesn't affect performance much, but it is cleaner.

### DIFF
--- a/crates/bevy_core_pipeline/src/deferred/node.rs
+++ b/crates/bevy_core_pipeline/src/deferred/node.rs
@@ -147,9 +147,9 @@ impl ViewNode for DeferredGBufferPrepassNode {
             }
 
             // Opaque draws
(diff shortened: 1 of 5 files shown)


alice-i-cecile:
@villor you're also qualified to review this one :)

pcwalton:
I added a commit that switches the inner map to an `IndexMap` too. This had a dramatic improvement in performance of `batch_and_prepare_binned_render_phase`: 2.15 ms to 1.42 ms on Caldera.

alice-i-cecile (approved):
Cleaner and faster, very nice. This one doesn't involve any shaders, so I'm quite comfortable reviewing it for a change :)

PR #17824, opened 2025-02-12
Shader validation enum

### LINKED ISSUE #17786: Make checked vs unchecked shaders configurable
I agree that this is unquestionably the better default. I won't push too hard for it in this PR (as this is better and I don't look a gift horse in the mouth), but making this configurable from our Shader type is low hanging fruit, and makes a very valuable and hard-to-implement-yourself feature available to people that want it.

Edit: I don't think doing that needs design work. Just make it an enum on Shader and expose the functionality via a new method on RenderDevice.

_Originally posted by @cart in https://github.com/bevyengine/bevy/pull/17767#pullrequestreview-2607300869_

fjkorf:
# Objective

Make checked vs unchecked shaders configurable
Fixes #17786 

## Solution

Added `ValidateShaders` enum to `Shader` and added `create_and_validate_shader_module` to `RenderDevice`

## Testing

I tested the shader examples locally and they all worked. I'd like to write a few tests to verify but am unsure how to start.


fjkorf:
Does this effectively expose this functionality to the user?

crates/bevy_render/src/render_resource/pipeline_cache.rs
```diff
@@ -318,7 +318,19 @@ impl ShaderCache {
                 render_device
                     .wgpu_device()
                     .push_error_scope(wgpu::ErrorFilter::Validation);
-                let shader_module = render_device.create_shader_module(module_descriptor);
+
+                // `validate_shader` is evaluated to determine whether or not to perform
```
alice-i-cecile, line 322: I don't think this comment is useful: it's just explaining what the code below says.

crates/bevy_render/src/render_resource/pipeline_cache.rs
```diff
@@ -318,7 +318,19 @@ impl ShaderCache {
...
+                //  validation checks
+                let validate_shader = shader.validate_shader.clone();
```
alice-i-cecile, line 324: This shouldn't be needed: you can just make your little config enums `Copy`. Or in cases where that's more expensive, match on `&validate_shader` or `shader.validate_shader`.

crates/bevy_render/src/render_resource/shader.rs
```diff
@@ -21,6 +21,11 @@ pub enum ShaderReflectError {
     #[error(transparent)]
     Validation(#[from] naga::WithSpan<naga::valid::ValidationError>),
 }
+#[derive(Clone, Debug)]
```
alice-i-cecile, line 24: Needs doc strings!

alice-i-cecile, line 24:
```suggestion

#[derive(Clone, Debug)]
```

crates/bevy_render/src/render_resource/shader.rs
```diff
@@ -21,6 +21,11 @@ pub enum ShaderReflectError {
...
+pub enum ValidateShaders {
+    Disabled,
+    Enabled,
+}
```
alice-i-cecile, line 28:
```suggestion
}

```

crates/bevy_render/src/render_resource/shader.rs
```diff
@@ -36,6 +41,7 @@ pub struct Shader {
     // we must store strong handles to our dependencies to stop them
     // from being immediately dropped if we are the only user.
     pub file_dependencies: Vec<Handle<Shader>>,
+    pub validate_shader: ValidateShaders,
```
alice-i-cecile, line 44: Variable name is singular, type name is plural. They should both be singular: this setting is per-shader, as it should be.

crates/bevy_render/src/render_resource/pipeline_cache.rs
```diff
@@ -318,7 +318,19 @@ impl ShaderCache {
...
+                let shader_module = match validate_shader {
+                    ValidateShaders::Enabled => {
+                        render_device.create_and_validate_shader_module(module_descriptor)
+                    }
+                    ValidateShaders::Disabled => {
+                        render_device.create_shader_module(module_descriptor)
```
alice-i-cecile, line 330: create_shader_module should be made `unsafe`, borrowing the SAFETY comments from the method it calls.

alice-i-cecile (changes requested):
Thanks for tackling this :) Definitely the right direction! I've highlighted a few small cleanup items, some docs to write, and then we can bubble up the unsafe to the end user who's actually calling these things.

Shouldn't be too bad, and feel free to ask questions or for advice on the tricky bits.

fjkorf:
not sure if I got the `unsafe` bit in correctly. I'm getting a warning about it

crates/bevy_render/src/render_resource/shader.rs
```diff
@@ -36,6 +45,7 @@ pub struct Shader {
     // we must store strong handles to our dependencies to stop them
     // from being immediately dropped if we are the only user.
     pub file_dependencies: Vec<Handle<Shader>>,
+    pub validate_shader: ValidateShader,
```
alice-i-cecile, line 48: Missing doc strings :)

crates/bevy_render/src/renderer/render_device.rs
```diff
@@ -45,7 +45,15 @@ impl RenderDevice {
 
     /// Creates a [`ShaderModule`](wgpu::ShaderModule) from either SPIR-V or WGSL source code.
     #[inline]
-    pub fn create_shader_module(&self, desc: wgpu::ShaderModuleDescriptor) -> wgpu::ShaderModule {
+    /// # Safety
+    ///
+    /// we are interfacing with shader code, which may contain undefined behavior,
```
alice-i-cecile, line 50: These safety comments should be copied from the underlying unsafe wgpu method that this calls :)

crates/bevy_render/src/renderer/render_device.rs
```diff
@@ -45,7 +45,15 @@ impl RenderDevice {
 
     /// Creates a [`ShaderModule`](wgpu::ShaderModule) from either SPIR-V or WGSL source code.
     #[inline]
```
alice-i-cecile, line 47: This annotation should go below the docs :)

crates/bevy_render/src/renderer/render_device.rs
```diff
@@ -64,29 +72,41 @@ impl RenderDevice {
...
             },
         }
-
         #[cfg(not(feature = "spirv_shader_passthrough"))]
-        // SAFETY: we are interfacing with shader code, which may contain undefined behavior,
-        // such as indexing out of bounds.
-        // The checks required are prohibitively expensive and a poor default for game engines.
-        // TODO: split this method into safe and unsafe variants, and propagate the safety requirements from
-        // https://docs.rs/wgpu/latest/wgpu/struct.Device.html#method.create_shader_module_trusted to the unsafe form.
+        // SAFETY: use code without runtime checks
```
alice-i-cecile, line 82:
```suggestion
        // SAFETY: the caller is responsible for upholding the safety requirements
```

crates/bevy_render/src/renderer/render_device.rs
```diff
@@ -64,29 +72,41 @@ impl RenderDevice {
...
+        desc: wgpu::ShaderModuleDescriptor,
+    ) -> wgpu::ShaderModule {
+        #[cfg(feature = "spirv_shader_passthrough")]
+        match &desc.source {
+            wgpu::ShaderSource::SpirV(_source)
+                if self
+                    .features()
+                    .contains(wgpu::Features::SPIRV_SHADER_PASSTHROUGH) =>
+            // SAFETY:
+            // we are using the SpirV Shader Passthrough which is unsafe
```
alice-i-cecile, line 102: This one definitely isn't right 🤔

crates/bevy_render/src/render_resource/shader.rs
```diff
@@ -21,6 +21,15 @@ pub enum ShaderReflectError {
     #[error(transparent)]
     Validation(#[from] naga::WithSpan<naga::valid::ValidationError>),
 }
+
+/// Describes whether or not to perform runtime checks on shaders
```
alice-i-cecile, line 25: This needs more details on the costs, risks and when you might choose each of these options.

crates/bevy_render/src/renderer/render_device.rs
```diff
@@ -44,8 +44,17 @@ impl RenderDevice {
     }
 
     /// Creates a [`ShaderModule`](wgpu::ShaderModule) from either SPIR-V or WGSL source code.
+    /// # Safety
+    ///
+    /// creates a shader module with user-customizable runtime checks which allows shaders to
```
alice-i-cecile, line 49:
```suggestion
    /// Creates a shader module with user-customizable runtime checks which allows shaders to
```

crates/bevy_render/src/renderer/render_device.rs
```diff
@@ -44,8 +44,17 @@ impl RenderDevice {
...
+    /// perform operations which can lead to undefined behavior like indexing out of bounds,
+    /// thus it's the caller responsibility to pass a shader which doesn't perform any of this
+    /// operations.
```
alice-i-cecile, line 52:
```suggestion
    /// To avoid UB, ensure your shaders are sound!
```

crates/bevy_render/src/renderer/render_device.rs
```diff
@@ -64,29 +73,48 @@ impl RenderDevice {
...
+        #[cfg(feature = "spirv_shader_passthrough")]
+        match &desc.source {
+            wgpu::ShaderSource::SpirV(_source)
+                if self
+                    .features()
+                    .contains(wgpu::Features::SPIRV_SHADER_PASSTHROUGH) =>
+            // SAFETY:
+            //
+            // This function passes binary data to the backend as-is and can potentially result in a
+            // driver crash or bogus behaviour. No attempt is made to ensure that data is valid SPIR-V.
```
alice-i-cecile, line 110:
```suggestion
            // driver crash or bogus behavior. No attempt is made to ensure that data is valid SPIR-V.
```

crates/bevy_render/src/render_resource/shader.rs
```diff
@@ -36,6 +45,8 @@ pub struct Shader {
     // we must store strong handles to our dependencies to stop them
     // from being immediately dropped if we are the only user.
     pub file_dependencies: Vec<Handle<Shader>>,
+    // enable or disable runtime validation
```
alice-i-cecile, line 48:
```suggestion
    /// Enable or disable runtime shader validation, trading safety against speed.
    /// 
    /// Please read the [`ValidateShader`] docs for a discussion of the tradeoffs involved. 
```

fjkorf:
Thanks!

crates/bevy_render/src/render_resource/shader.rs
```diff
@@ -21,6 +21,17 @@ pub enum ShaderReflectError {
     #[error(transparent)]
     Validation(#[from] naga::WithSpan<naga::valid::ValidationError>),
 }
+
+/// Describes whether or not to perform runtime checks on shaders.
+/// Runtime checks can be enabled for safety at the cost of speed.
+/// By default no runtime checks will be performed.
```
robtfm, line 27: can we please add a note that no checks are performed for spirv, regardless of this setting?

... or if we panic for validated spirv as i suggest below, we can note that here instead

crates/bevy_render/src/renderer/render_device.rs
```diff
@@ -64,29 +73,48 @@ impl RenderDevice {
...
+        match &desc.source {
+            wgpu::ShaderSource::SpirV(_source)
+                if self
+                    .features()
+                    .contains(wgpu::Features::SPIRV_SHADER_PASSTHROUGH) =>
+            // SAFETY:
+            //
+            // This function passes binary data to the backend as-is and can potentially result in a
+            // driver crash or bogus behavior. No attempt is made to ensure that data is valid SPIR-V.
+            unsafe { self.create_shader_module(desc) },
```
robtfm, line 111: having an unsafe function to create unchecked shaders, and having a safe function that also creates unchecked shaders in an unsafe block if you pass in spirv feels wrong (and the comment on the function is misleading).

i think we should probably panic for spirv here.

crates/bevy_render/src/renderer/render_device.rs
```diff
@@ -64,29 +73,48 @@ impl RenderDevice {
...
-            // such as indexing out of bounds.
-            // The checks required are prohibitively expensive and a poor default for game engines.
-            // TODO: split this method into safe and unsafe variants, and propagate the safety requirements from
-            // https://docs.rs/wgpu/latest/wgpu/struct.Device.html#method.create_shader_module_trusted to the unsafe form.
+            // SAFETY:
+            //
+            // creates a shader module with user-customizable runtime checks which allows shaders to
+            // perform operations which can lead to undefined behavior like indexing out of bounds,
+            // thus it's the caller responsibility to pass a shader which doesn't perform any of this
+            // operations.
```
robtfm, line 81: i'd copy the short safety comment below, not much use in copying the long version from the top of the function.

robtfm (changes requested):

crates/bevy_render/src/render_resource/shader.rs
```diff
@@ -21,6 +21,19 @@ pub enum ShaderReflectError {
...
+/// Because no runtime checks are performed for spirv,
```
alice-i-cecile, line 28:
```suggestion
/// 
/// # Panics
/// Because no runtime checks are performed for spirv,
```

crates/bevy_render/src/render_resource/shader.rs
```diff
@@ -21,6 +21,19 @@ pub enum ShaderReflectError {
...
+/// enabling `ValidateShader` for spirv will cause a panic
+#[derive(Clone, Debug, Default)]
+pub enum ValidateShader {
+    #[default]
+    Disabled,
```
alice-i-cecile, line 33:
```suggestion
    /// No runtime checks for soundness (e.g. bound checking) are performed.
    ///
    /// This is suitable for trusted shaders, written by your program or dependencies you trust.
    Disabled,
```

crates/bevy_render/src/render_resource/shader.rs
```diff
@@ -21,6 +21,19 @@ pub enum ShaderReflectError {
...
+    Enabled,
```
alice-i-cecile, line 34:
```suggestion
    /// Enable's runtime checks for soundness (e.g. bound checking).
    ///
    /// While this can have a meaningful impact on performance,
    /// this setting should *always* be enabled when loading untrusted shaders.
    /// This might occur if you are creating a shader playground, running user-generated shaders
    /// (as in VRChat), or writing a web browser in Bevy.
    Enabled,
```

crates/bevy_render/src/renderer/render_device.rs
```diff
@@ -64,29 +73,42 @@ impl RenderDevice {
...
-        // TODO: split this method into safe and unsafe variants, and propagate the safety requirements from
-        // https://docs.rs/wgpu/latest/wgpu/struct.Device.html#method.create_shader_module_trusted to the unsafe form.
+        // SAFETY: the caller is responsible for upholding the safety requirements
         unsafe {
             self.device
                 .create_shader_module_trusted(desc, wgpu::ShaderRuntimeChecks::unchecked())
         }
     }
 
+    /// Creates and validates a [`ShaderModule`](wgpu::ShaderModule) from either SPIR-V or WGSL source code.
```
alice-i-cecile, line 93:
```suggestion
    /// Creates and validates a [`ShaderModule`](wgpu::ShaderModule) from either SPIR-V or WGSL source code.
    ///
    /// See [`ValidateShaders`] for more information on the tradeoffs involved with shader validation.
```

crates/bevy_render/src/renderer/render_device.rs
```diff
@@ -44,8 +44,17 @@ impl RenderDevice {
     }
 
     /// Creates a [`ShaderModule`](wgpu::ShaderModule) from either SPIR-V or WGSL source code.
+    /// # Safety
```
alice-i-cecile, line 47:
```suggestion
    ///
    /// # Safety
```

alice-i-cecile (commented):
A couple more nits and additions to the docs, but this is nearly ready :) Thanks!

fjkorf:
Thank you!

crates/bevy_render/src/renderer/render_device.rs
```diff
@@ -64,29 +74,44 @@ impl RenderDevice {
...
             self.device
                 .create_shader_module_trusted(desc, wgpu::ShaderRuntimeChecks::unchecked())
         }
     }
 
+    /// Creates and validates a [`ShaderModule`](wgpu::ShaderModule) from either SPIR-V or WGSL source code.
+    ///
+    /// See [`ValidateShader`](bevy_render::render_resource::ValidateShader) for more information on the tradeoffs involved with shader validation.
+    #[inline]
+    pub fn create_and_validate_shader_module(
```
IceSentry, line 98: Maybe I'm missing something, but this looks like it just ends up calling `.create_shader_module()` and that function doesn't do any validation.

fjkorf: This calls `create_shader_module` in the wgpu device. In that implementation runtime checks are enabled

https://docs.rs/wgpu/latest/src/wgpu/api/device.rs.html#80-85

```
    pub fn create_shader_module(&self, desc: ShaderModuleDescriptor<'_>) -> ShaderModule {
        let module = self
            .inner
            .create_shader_module(desc, wgt::ShaderRuntimeChecks::checked());
        ShaderModule { inner: module }
    }
   ```

Comment (IceSentry) [reply crates/bevy_render/src/renderer/render_device.rs:98]: Ah, I thought it was calling the bevy one, that makes way more sense.

**Review by alice-i-cecile — APPROVED** [review]

**Review by IceSentry — APPROVED** [review]

File: crates/bevy_render/src/renderer/render_device.rs [hunk 26]
```diff
@@ -64,29 +74,44 @@ impl RenderDevice {
...
+    ) -> wgpu::ShaderModule {
+        #[cfg(feature = "spirv_shader_passthrough")]
+        match &desc.source {
+            wgpu::ShaderSource::SpirV(_source)
+                if self
+                    .features()
+                    .contains(wgpu::Features::SPIRV_SHADER_PASSTHROUGH) =>
+            panic!("no safety checks are performed for spirv shaders. use `create_shader_module` instead"),
+            _ => self.device.create_shader_module(desc),
+        }
```
robtfm, line 110: what happens if we don't have SPIRV_SHADER_PASSTHROUGH here?

fjkorf: If we don't have SPRIV_SHADER_PASSTHROGH the match guard will be false. The `_` branch will be executed I believe

robtfm: ok. i'm not sure it has much value here; it will presumably result in failure either way, but a different failure depending on the enabled features. can we just remove it so we always panic on this path?

fjkorf: I have removed the match guard

robtfm (approved):

Issue #3492, opened 2021-12-30
Tracking issue: deny missing docs, one crate at a time


alice-i-cecile:
Docs are great! However, it's easy to let them atrophy.

As Bevy grows and stabilizes, we want to be able to turn on `#![warn(missing_docs)]` (which will cause CI to fail if violated). This is helpful because it helpful for new users and contributors, and ensures that our code base's documentation state doesn't regress.

However, not everything is stable or well documented enough to do so already.
We should turn this on one crate at a time, once it hits 100% coverage. If there's a crate

Let's check the current doc coverage.

1. Set the following environment variable (on Windows, I used the [`$Env:` syntax](https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_environment_variables?view=powershell-7.2)): `RUSTDOCFLAGS="-Z unstable-options --show-coverage"`
2. Run `cargo +nightly doc --workspace --all-features --no-deps` 

Run on 2023-03-14:
**Crate | Docs Coverage** 

- [x] `bevy_a11y` | 100.0%
- [x] `bevy_animation` | 100.0%
- [ ] `bevy_utils` | 53.3%
- [x] `bevy_tasks` | 100.0% (#3509)
- [ ] `bevy_derive` | 33%
- [ ] `bevy_macro_utils` | 28.6%
- [x] `bevy_math` | 100.0% (#3503 | #4591)
- [x] `bevy_app` | 100.0% (#3539)
- [ ] `bevy_ecs_macros` | 40.0%
- [ ] `bevy_dynamic_plugin` | 42.9%
- [ ] `bevy_window` | 97.8% (#4333)
- [x] `bevy_crevice` | **100%** (Not our problem)
- [x] `bevy_log` | 100.0%
- [x] `bevy_transform` | 100.0%
- [x] `bevy_core` | 100.0%
- [ ] `bevy_reflect`| 61.7%
- [ ] `bevy_input` | 90.6% (Claimed by @KDecay)
- [ ] `bevy_gilrs` | 0%
- [ ] `bevy_diagnostic` | 50%
- [ ] `bevy_winit` | 56%
- [x] `bevy_audio` | 100.0%
- [ ] `bevy_scene` | 22.1%
- [x] `bevy_asset` | 100.0% (https://github.com/bevyengine/bevy/pull/3536)
- [ ] `bevy_core_pipeline` | 3.4%
- [ ] `bevy_gltf` |16.7%
[...]

PR #17829, opened 2025-02-12
Fill out some missing docs for bevy_assets

alice-i-cecile:
# Objective

`bevy_assets` has long been unapproachable for contributors and users. More and better documentation would help that.

We're gradually moving towards globally denying missing docs (#3492)! 
However, writing all of the hundreds of missing doc strings in a single go will be miserable to review.

## Solution

Remove the allow for missing docs temporarily, and then pick some easy missing doc warnings largely at random to tackle.

Stop when the change set is starting to feel intimidating.

### DIFF
--- a/crates/bevy_asset/src/id.rs
+++ b/crates/bevy_asset/src/id.rs
@@ -26,7 +26,9 @@ pub enum AssetId<A: Asset> {
     ///
     /// [`Assets`]: crate::Assets
     Index {
+        /// The unstable, opaque index of the asset.
         index: AssetIndex,
+        /// A marker to store the type information of the asset.
         #[reflect(ignore)]
         marker: PhantomData<fn() -> A>,
     },
@@ -34,7 +36,10 @@ pub enum AssetId<A: Asset> {
     /// with one.
     ///
     /// [`Assets`]: crate::Assets
-    Uuid { uuid: Uuid },
+    Uuid {
+        /// The UUID provided during asset registration.
+        uuid: Uuid,
+    },
 }
 
 impl<A: Asset> AssetId<A> {
@@ -165,12 +170,22 @@ pub enum UntypedAssetId {
     /// explicitly registered that way.
     ///
     /// [`Assets`]: crate::Assets
-    Index { type_id: TypeId, index: AssetIndex },
+    Index {
+        /// An identifier that records the underlying asset type.
+        type_id: TypeId,
+        /// The unstable, opaque index of the asset.
+        index: AssetIndex,
+    },
     /// A stable-across-runs / const asset identifier. This will only be used if an asset is explicitly registered in [`Assets`]
     /// with one.
     ///
     /// [`Assets`]: crate::Assets
-    Uuid { type_id: TypeId, uuid: Uuid },
+    Uuid {
+        /// An identifier that records the underlying asset type.
+        type_id: TypeId,
+        /// The UUID provided during asset registration.
+        uuid: Uuid,
+    },
 }
 
 impl UntypedAssetId {
@@ -404,7 +419,12 @@ impl<A: Asset> TryFrom<UntypedAssetId> for AssetId<A> {
 pub enum UntypedAssetIdConversionError {
     /// Caused when trying to convert an [`UntypedAssetId`] into an [`AssetId`] of the wrong type.
     #[error("This UntypedAssetId is for {found:?} and cannot be converted into an AssetId<{expected:?}>")]
-    TypeIdMismatch { expected: TypeId, found: TypeId },
+    TypeIdMismatch {
+        /// The TypeId of the asset that we are trying to convert to.
+        expected: TypeId,
+        /// The TypeId of the asset that we are trying to convert from.
+        found: TypeId,
+    },
 }
 
 #[cfg(test)]
--- a/crates/bevy_asset/src/id.rs
+++ b/crates/bevy_asset/src/id.rs
@@ -420,9 +420,9 @@ pub enum UntypedAssetIdConversionError {
     /// Caused when trying to convert an [`UntypedAssetId`] into an [`AssetId`] of the wrong type.
     #[error("This UntypedAssetId is for {found:?} and cannot be converted into an AssetId<{expected:?}>")]
     TypeIdMismatch {
-        /// The TypeId of the asset that we are trying to convert to.
+        /// The [`TypeId`] of the asset that we are trying to convert to.
         expected: TypeId,
-        /// The TypeId of the asset that we are trying to convert from.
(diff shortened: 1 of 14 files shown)


andriyDev (approved):
Looks good!

crates/bevy_asset/src/id.rs
```diff
@@ -404,7 +419,12 @@ impl<A: Asset> TryFrom<UntypedAssetId> for AssetId<A> {
 pub enum UntypedAssetIdConversionError {
     /// Caused when trying to convert an [`UntypedAssetId`] into an [`AssetId`] of the wrong type.
     #[error("This UntypedAssetId is for {found:?} and cannot be converted into an AssetId<{expected:?}>")]
-    TypeIdMismatch { expected: TypeId, found: TypeId },
+    TypeIdMismatch {
+        /// The TypeId of the asset that we are trying to convert to.
```
alice-i-cecile, line 423:
```suggestion
        /// The [`TypeId`] of the asset that we are trying to convert to.
```

crates/bevy_asset/src/id.rs
```diff
@@ -404,7 +419,12 @@ impl<A: Asset> TryFrom<UntypedAssetId> for AssetId<A> {
...
+        expected: TypeId,
+        /// The TypeId of the asset that we are trying to convert from.
```
alice-i-cecile, line 425:
```suggestion
        /// The [`TypeId`] of the asset that we are trying to convert from.
```

alice-i-cecile:
@Carter0, want to take a look at this PR too?

crates/bevy_asset/src/id.rs
```diff
@@ -26,15 +26,20 @@ pub enum AssetId<A: Asset> {
     ///
     /// [`Assets`]: crate::Assets
     Index {
+        /// The unstable, opaque index of the asset.
```
Carter0, line 29: This comment sent me down a giant rabbit whole lol. 

I get that opaque means that there is some data known to the compiler but not to the user. I don't really know what unstable means in this context though. 

Maybe it's worth linking to the AssetIndex struct for users to read more?

crates/bevy_asset/src/io/embedded/embedded_watcher.rs
```diff
@@ -24,6 +24,7 @@ pub struct EmbeddedWatcher {
 }
 
 impl EmbeddedWatcher {
+    /// Creates a new `EmbeddedWatcher` that watches for changes to the embedded assets in the given `dir`.
```
Carter0, line 27: I don't really think you need this. The comment above the struct is better.

crates/bevy_asset/src/io/embedded/mod.rs
```diff
@@ -15,6 +15,8 @@ use std::path::{Path, PathBuf};
 #[cfg(feature = "embedded_watcher")]
 use alloc::borrow::ToOwned;
 
+/// The name of the `embedded` [`AssetSource`],
+/// as stored in the [`AssetSourceBuilders`] resource.
 pub const EMBEDDED: &str = "embedded";
```
Carter0, line 20: What does "embedding" and asset mean? I looked at the example and it didn't really explain it either.

crates/bevy_asset/src/io/file/mod.rs
```diff
@@ -65,15 +65,14 @@ impl FileAssetReader {
     }
 }
 
+/// A writer for the local filesystem.
```
Carter0, line 68: This isn't that useful. Maybe you could add a little more? What's it writing and why do we need it?

crates/bevy_asset/src/io/source.rs
```diff
@@ -132,26 +132,34 @@ impl<'a> PartialEq for AssetSourceId<'a> {
...
     pub writer: Option<Box<dyn FnMut(bool) -> Option<Box<dyn ErasedAssetWriter>> + Send + Sync>>,
+    /// The [`AssetWatcher`] to use for unprocessed assets, if any.
     pub watcher: Option<
         Box<
             dyn FnMut(crossbeam_channel::Sender<AssetSourceEvent>) -> Option<Box<dyn AssetWatcher>>
                 + Send
                 + Sync,
         >,
     >,
+    /// The [`ErasedAssetReader`] to use for processed assets.
```
Carter0, line 147: Whats the difference between a "processed" and "unprocessed" asset?

crates/bevy_asset/src/processor/log.rs
```diff
@@ -32,8 +32,10 @@ pub struct ProcessorTransactionLog {
 /// An error that occurs when reading from the [`ProcessorTransactionLog`] fails.
 #[derive(Error, Debug)]
```
Carter0, line 33: Also might be worth linking to this above https://github.com/bevyengine/bevy/blob/main/crates/bevy_asset/src/processor/mod.rs

Carter0 (approved):
As a new user to the bevy_assets crate, there is a lot of docs here for fields in structs. That's good and I think that its a step towards requiring docs in the lints, which is also good! 

Personally though, I find the docs over the small stuff is only useful if you already know the big picture stuff. Thats why I like links to the big docs in mod files or above structs because its a place for people like me to learn the concept first, then dig into the specifics. 

Idk if Bevy has a documentation roadmap or plan already. Documentation is hard because there are several different kinds of documentation and they all serve different purposes. 

Anyway, Thanks for doing this and I do like it!

crates/bevy_asset/src/io/embedded/embedded_watcher.rs
alice-i-cecile: Yeah this is lint-proofing.

crates/bevy_asset/src/id.rs
kristoff3r: Opaque means you can't "see inside" it, or e.g. do arithmetic with it, but only use it by giving it to APIs. The compiler helps us enforce this but that's not a requirement for using the term.

Unstable means you can't expect to get the same index between restarts of the application, so it doesn't make sense to serialize it for example.

In all cases I can think of where doc links are rendered you'd also have the type right there, so a link is probably redundant.

crates/bevy_asset/src/io/embedded/mod.rs
kristoff3r: It means putting the asset inside the game executable file, in contrast to other assets that are normally loaded from disk to memory when needed. It's normally only used for small engine-related assets so you can use them without having them in your own asset directory.  

It sounds like the example could use a line or 2 describing that.

kristoff3r (approved):

fjkorf (approved):
These doc strings make sense to me.

PR #17833, opened 2025-02-12
Update `HitData` position docs

### LINKED ISSUE #17832: `HitData` docs don't accurately describe the behavior of UI picking
> Good point, I overlooked that! Perhaps docs could say something along the lines of `The position reported by the backend, if the data is available.`, as giving mouse coords a la web would still be in screen space rather than world space. Maybe a mention of different spaces in docs would be useful too?

I guess I was thinking solely of my own use-case for having position data in relative local space, and knowing that the pointer position is accessible elsewhere in an observer trigger opted for this space.

_Originally posted by @bytemunch in https://github.com/bevyengine/bevy/issues/17681#issuecomment-2654616820_

bytemunch:
# Objective

Updates the now inaccurate position docs
Fixes #17832 

## Solution

From
`The position of the intersection in the world, if the data is available from the backend.`
To
`The position reported by the backend, if the data is available. Position data may be in any space (e.g. World space, Screen space, Local space), specified by the backend providing it.`

## Testing

uhh reading :)

### DIFF
--- a/crates/bevy_picking/src/backend.rs
+++ b/crates/bevy_picking/src/backend.rs
@@ -105,7 +105,8 @@ pub struct HitData {
     /// distance from the pointer to the hit, measured from the near plane of the camera, to the
     /// point, in world space.
     pub depth: f32,
-    /// The position of the intersection in the world, if the data is available from the backend.
+    /// The position reported by the backend, if the data is available. Position data may be in any
+    /// space (e.g. World space, Screen space, Local space), specified by the backend providing it.
     pub position: Option<Vec3>,
     /// The normal vector of the hit test, if the data is available from the backend.
     pub normal: Option<Vec3>,


alice-i-cecile (approved):

alice-i-cecile:
@notmd, how's this?

notmd (approved):
Maybe we can add which space is used in each backend/plugin? About the normalize value for UI, I think we should use the non normalized value instead. In web it is https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/offsetX which use quite often. The normalized value can easily be computed from non normalized one.

alice-i-cecile:
Normalized values of the clicks in UI is quite helpful for building widgets, while non-normalized value is usually just the mouse position 🤔 That's for a different thread though; maybe spin up an issue?

PR #17836, opened 2025-02-13
docs(example): Example how to insert reflected `Component` onto entities

favilo:
# Objective

We needed an example that showed how to insert a `Component` onto an entity that was deserialized into a type that implements `Reflect`.

`ReflectCommandExt::insert_reflect` was not easy to discover.

## Solution

This new example shows how to use `insert_reflect()`, while also showing how to deserialize data from glTF extras, and add the deserialized `Component`s automatically to the corresponding scene's children entities on instantiation.

## Testing

Tested locally on a Linux machine. Runs fine.

It's an example using standard features, so I assume it'll work where ever

```sh
$ cargo run --example load_gltf_components
```

Linux desktop machine. I could potentially test on other platforms like wasm, but I haven't yet.


ChristopherBiscardi:
IMO this example should be greatly reduced to essentially only the `deserialize_extra` that is written and it should be moved to the reflection folder with the more minimal result. gltf extras handling is already shown in the gltf extras example.

Something closer to this would make sense, without any gltf code.

```rust
let json_component = r#"{"skein::tests::SomeThings":{"OneThing":{"name":"testing"}}}"#
let type_registry = type_registry.read();

let reflect_deserializer = ReflectDeserializer::new(&type_registry);
let mut deserializer = serde_json::Deserializer::from_str(value);
let reflect_value = reflect_deserializer
    .deserialize(json_component)
    .unwrap();

commands
    .entity(entity)
    .insert_reflect(reflect_value);
```

Also ron is used in the other examples, I'm not sure if the Bevy repo has a preference for ron over serde_json for the examples

favilo:
@ChristopherBiscardi, what do you think about this example. I've got both json and ron deserializing, and verify that the component is constructed properly.

ChristopherBiscardi (approved):
overall lgtm.

A lot of the example is the same as [`examples/reflection/reflection.rs`](https://github.com/bevyengine/bevy/blob/0f1c75796b75ce518ae465d8f733c17f852de3a6/examples/reflection/reflection.rs#L103) but I think showing two different serializations (ron/json) is enough to validate the additional example even though `insert_reflect` could be added to the other example. 

I also dislike Foo/Bar naming but the other example this is extended from also uses Foo/Bar so matching that is probably better for ease of recognizing commonalities. CI is failing on the variable `foo` though so that will have to change to merge.

Notably `insert_reflect` is hard to find in the documentation because it is on an extension trait, so having an example for it is a good thing.

PR #17840, opened 2025-02-13
Parallel Transform Propagation

aevyrie:
# Objective

- Make transform propagation faster.

## Solution

- Work sharing worker threads
- Parallel tree traversal excluding leaves
- Second cache friendly wide pass over all leaves
- 3-10x faster than main

## Testing

- Tracy
- Caldera hotel is showing 3-7x faster on my M4 Max. Timing for bevy's existing transform system shifts wildly run to run, so I don't know that I would advertise a particular number. But this implementation is faster in a... statistically significant way.
image


aevyrie:
Getting this up now that it seems to function, marking as draft because it still needs cleanup.

crates/bevy_transform/src/systems.rs
```diff
@@ -36,148 +41,159 @@ pub fn sync_simple_transforms(
...
+            // SAFETY:
+            //
+            // The hierarchy is guaranteed to exist without cycles. Traversing the hierarchy is then
+            // guaranteed to visit disjoint entities, which means it is impossible to mutably alias
+            // two entities.
+            #[expect(unsafe_code, reason = "Mutating disjoint entities in parallel")]
+            unsafe {
+                let mut message = Vec::new();
+                let mut children = nodes.iter_many_unsafe(children);
+                while let Some((child, transform, mut global_transform, _)) = children.fetch_next()
```
Victoronz: This situation is a perfect fit for `iter_many_unique_mut`!

Instead of having the whole iteration be be unsafe, you can wrap the input iterator in `UniqueEntityIter` and the pass that to `iter_many_unique_mut`:
```Rust
let children_iter = UniqueEntityIter::from_iterator_unchecked(children.iter())
nodes.iter_many_unique_mut(children_iter)
```
What this does is it allows to use normal iterator infrastructure instead of a `while let` loop.
This is a more common code structure so should generate better code in some cases, though there is no guarantee there will be a difference.
F.e.: I don't think SIMD is an optimization available to the compiler when no more than 1 element is allowed to be live at the same time.
Even if there isn't a perf difference it should be somewhat cleaner/less restricting!

crates/bevy_transform/src/systems.rs
```diff
@@ -36,148 +41,159 @@ pub fn sync_simple_transforms(
...
+            // SAFETY:
+            //
+            // The hierarchy is guaranteed to exist without cycles. Traversing the hierarchy is then
+            // guaranteed to visit disjoint entities, which means it is impossible to mutably alias
+            // two entities.
+            #[expect(unsafe_code, reason = "Mutating disjoint entities in parallel")]
+            unsafe {
+                for parent in tasks.drain(..) {
+                    let (_, _, parent_transform, children) = nodes.get_unchecked(parent).unwrap();
+                    let mut children = nodes.iter_many_unsafe(children);
```
Victoronz: Same here.

crates/bevy_transform/src/systems.rs
```diff
@@ -36,150 +44,159 @@ pub fn sync_simple_transforms(
...
+        for parent in tasks.drain(..) {
+            // SAFETY: Visiting the hierarchy as a tree without cycles
+            #[expect(unsafe_code, reason = "Mutating disjoint entities in parallel")]
+            unsafe {
+                let (_, _, parent_transform, children) = nodes.get_unchecked(parent).unwrap();
+                propagate_to_child_unchecked(parent_transform, children, nodes, &mut outbox);
+            }
+        }
+        for chunk in outbox.chunks(1024) {
+            queue.sender.send(chunk.to_vec()).ok();
```
Victoronz, line 147: It should be slightly more efficient to send boxed slices instead of `Vec`s, would that work on the receiving end?
```suggestion
            queue.sender.send(Box::from(chunk)).ok();
```

crates/bevy_ui/src/layout/mod.rs
```diff
@@ -409,7 +409,8 @@ mod tests {
                 ApplyDeferred,
                 ui_layout_system,
                 sync_simple_transforms,
-                propagate_transforms,
+                propagate_transform_nodes,
+                compute_transform_leaves,
```
notmd, line 413: I think it's time to add a SystemSet for transform propagate

crates/bevy_transform/src/systems.rs
```diff
@@ -36,150 +45,197 @@ pub fn sync_simple_transforms(
...
+        Entity,
+        Ref<'_, Transform>,
+        &mut GlobalTransform,
+        &Children,
+        &ChildOf,
+    )>,
+) {
+    let mut outbox = queue.local_queue.borrow_local_mut();
+    loop {
+        // Try to acquire a lock on the work queue in a tight loop.
```
cBournhonesque, line 135: I'm not too familiar with mutexes; would something like this https://marabos.nl/atomics/basics.html#condvar be benefitial to avoid a spinloop?

crates/bevy_transform/src/systems.rs
```diff
@@ -36,150 +45,197 @@ pub fn sync_simple_transforms(
...
+    queue: &WorkQueue,
+    nodes: &Query<(
+        Entity,
+        Ref<'_, Transform>,
+        &mut GlobalTransform,
+        &Children,
+        &ChildOf,
+    )>,
+) {
+    let mut outbox = queue.local_queue.borrow_local_mut();
```
cBournhonesque, line 133: So this doesn't allocate because we're re-using the thread-local `Vec<Entity>` (from Parallel) and we are always re-using the same threads from the TaskPool?

crates/bevy_transform/src/systems.rs
notmd: Just wonder if cloning the outbox to do 1 big allocation is better than multiple small allocations?

Victoronz: > Is it because the Box is one usize but a Vec is 2 usize?
> 
> Or because `chunk.to_vec()` makes a copy of the vec, while `Box::from(chunk)` reuses the allocation?

`chunk` is a slice, both `.to_vec()` and `Box::from` will make an allocation, however what is being sent over the channel is just the pointer. `Box<[T]>` is 16 bytes, while `Vec<T>` is 24 bytes.
If these allocations are never extended, then the `capacity` field on `Vec` has no use here.

aevyrie: I tried using a boxed slice, but profiling showed it was the same or slower than the vector.

crates/bevy_transform/src/systems.rs
aevyrie: Maybe. The workers should ever be waiting other than the very beginning, so there is likely no real benefit.

aevyrie: To justify this claim, the gap between tasks (when spinning happens) is only300-500ns long
image

cBournhonesque: I think if we intend to spin loop we can provide a hint to the compiler to make some optimizations: https://doc.rust-lang.org/std/hint/fn.spin_loop.html

alice-i-cecile:
I'd love a before / after tracy histogram BTW. I've seen enough tests on Discord to be convinced that this is markedly (3x or so) faster, but it would be lovely to show in the release notes.

aevyrie:
For wide hierarchies, it's 8-10 times faster. An entity with 10 million children:
image

crates/bevy_transform/src/systems.rs
```diff
@@ -236,6 +236,18 @@ unsafe fn propagate_to_child_unchecked(
     // Safety: `Children` is guaranteed to hold unique entities.
     #[expect(unsafe_code, reason = "Mutating disjoint entities in parallel")]
     let children_iter = unsafe { UniqueEntityIter::from_iterator_unchecked(children.iter()) };
+    // Performance note: iter_many tests every child to see if it meets the query. For leaf nodes,
```
JMS55, line 239: Is there a way you can add a filter to the query like `Without<IsParent>` to filter out the leaves?

aevyrie: That filter is already present. The problem is that we have a list of entities from `Children` that need to be tested against that query.

crates/bevy_transform/src/systems.rs
```diff
@@ -89,100 +131,397 @@ pub fn propagate_transforms(
...
+                );
+            }
+        }
+    }
+}
+
+/// Parallel hierarchy traversal with an optimized batching scheduler. Often 2-5 times faster than
+/// the serial version.
+///
+/// Relies on `std` until a `no_std` `mpsc` channel is available.
```
bushrat011899, line 230: I do want to solve this. It appears a simple API compatible `no_std` alternative just isn't readily available on crates.io right now (certainly not one I could easily find). I might solve this in an external crate potentially since it's very tangential to Bevy itself.

crates/bevy_transform/src/systems.rs
```diff
@@ -89,100 +131,396 @@ pub fn propagate_transforms(
...
+            unsafe { nodes.iter_many_unique_unsafe(children_iter) }
+        {
+            assert!(child_of.get() == parent);
+            if parent_transform.is_changed()
+                || transform.is_changed()
+                || global_transform.is_added()
+            {
+                *global_transform = parent_transform.mul_transform(*transform);
+            }
+            outbox.push(child);
```
Victoronz, line 516: Since you now have an iterator, it would reduce wasteful resizes to use `outbox.extend()` instead of series of pushes in a loop.

aevyrie: Good catch, I always forget about `.extend()`. 🙂

crates/bevy_transform/src/systems.rs
```diff
@@ -89,100 +131,404 @@ pub fn propagate_transforms(
...
+                if transform.is_changed()
+                    || parent_transform.is_added()
+                    || orphans.binary_search(&parent).is_ok()
+                {
+                    *parent_transform = GlobalTransform::from(*transform);
+                }
+
+                // SAFETY: the parent entities passed into this function are taken from iterating
+                // over the root entity query. Queries iterate over disjoint entities, preventing
+                // mutable aliasing, and making this call safe.
```
superdump, line 282: Do we have any hierarchy validity enforcement such that it is not possible to have an entity be a child of more than one parent?

aevyrie: This is done during propagation, to ensure there are no cycles when propagating. It would be nice if we could enforce this somehow, but with reflection, it's always possible to poison the hierarchy.

superdump: i guess the consequences of such a graph, rather than a tree, would be just as nonsensical with the safe version?

aevyrie: Yes, propagation would follow the same path, the only difference is that the unsafe one can be done in parallel. Note that our current impl is parallel at the roots only, is unsafe, and does this same test and panics if a cycle is found. The change here doesn't really change much in terms of unsafety - we are still pinky promising the compiler that our access to entities in parallel with unsafe is disjoint.

crates/bevy_transform/src/systems.rs
```diff
@@ -89,100 +131,404 @@ pub fn propagate_transforms(
...
+                    let (_, (_, p_global_transform), (p_children, _)) =
+                        nodes.get_unchecked(parent).unwrap();
+                    propagate_descendants_unchecked(
+                        parent,
+                        p_global_transform,
+                        p_children,
+                        nodes,
+                        &mut outbox,
+                        queue,
+                        10_000,
```
superdump, line 356: Empirically selected? :)

superdump (approved):
It looks complicated at first glance, but it's actually not so bad, especially because of the comments on all the non-obvious stuff. Awesome improvements! Thanks for your efforts. And I'm glad we have unit tests for this. :D

crates/bevy_transform/src/systems.rs
aevyrie: You could call it that. :)
It's big enough to be larger than any test or scene we have, but small enough that it won't hang the application. I'm actually looking at tweaking it to improve work sharing.

crates/bevy_transform/src/systems.rs
```diff
@@ -36,27 +41,60 @@ pub fn sync_simple_transforms(
...
+                || transform.is_changed()
+                || global_transform.is_added()
+            {
+                *global_transform = parent_transform.mul_transform(*transform);
+            }
+        });
+}
+
+/// Serial hierarchy traversal. Useful in `no_std` or single threaded contexts.
+#[cfg(not(feature = "std"))]
```
mockersf, line 69: this cfg makes it not usable in single threaded context with std enabled

aevyrie: We discussed this a bit on discord. Basically, this PR conflates no_std and single threaded, because we don't have everything needed to make them fully independent. The `serial` implementation isn't actually serial, it parallelizes across the roots.

What we want to do in a future refactor is take the current "single threaded" implementation, and actually make it single threaded. This will remove any overhead associated with working on a task pool when you only have a single thread, and will have the benefit of removing the need for any unsafe.

We would then make the multithreaded implementation work across std and no_std, but this needs a no_std compatible `Channel`.

So, with that context out of the way:

> this cfg makes it not usable in single threaded context with std enabled

I don't think this is true. If the task pool only has a single thread, the parallel implementation will still work just fine, it will only have a single worker on the main thread, but it is arguably less efficient than a specialized single-threaded implementation, which we don't have yet because of the context above.

mockersf: > I don't think this is true.

I meant, the `serial` mod that is being documented and cfg-gated. Its documentation mentions single threaded, but the cfg is on std.

Could you change the doc comment, and maybe add a todo with the direction for the potential future refactor?

superdump (approved):

crates/bevy_transform/src/systems.rs
aevyrie: Good call! I've added a note.

superdump (approved):

hymm (commented):
Nice work. Overall makes sense to me, but I'm going to give this another pass tomorrow and try and understand the logic a little better.

crates/bevy_transform/src/systems.rs
```diff
@@ -89,100 +145,421 @@ pub fn propagate_transforms(
...
+        ),
+    >;
+
+    /// A queue shared between threads for transform propagation.
+    pub struct WorkQueue {
+        /// A semaphore that tracks how many threads are busy doing work. Used to determine when
+        /// there is no more work to do.
+        busy_threads: AtomicI32,
+        sender: Sender<Vec<Entity>>,
+        receiver: Arc<Mutex<Receiver<Vec<Entity>>>>,
```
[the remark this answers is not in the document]
aevyrie: I'm using the dependencies I have on hand. If there is a dep already in tree, we can always switch to that later.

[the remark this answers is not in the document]
aevyrie: From profiling, batching seems to alleviate any issue caused by perceived slowness of `std::mpsc`. The gap between tasks is on the order of a few hundred nanoseconds. There is also the fact that this is already a big performance improvement. There might be more on the table, but this is worth merging as-is.

crates/bevy_transform/src/plugins.rs
```diff
@@ -31,28 +33,27 @@ impl Plugin for TransformPlugin {
...
-                sync_simple_transforms
-                    .in_set(TransformSystem::TransformPropagate)
-                    // FIXME: https://github.com/bevyengine/bevy/issues/4381
-                    // These systems cannot access the same entities,
-                    // due to subtle query filtering that is not yet correctly computed in the ambiguity detector
-                    .ambiguous_with(PropagateTransformsSet),
-                propagate_transforms.in_set(PropagateTransformsSet),
-            ),
+            ((
+                sync_simple_transforms,
```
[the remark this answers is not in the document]
aevyrie: I can retry this though!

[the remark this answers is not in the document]
alice-i-cecile: This would be a helpful comment :)

crates/bevy_transform/src/systems.rs
```diff
@@ -89,100 +145,421 @@ pub fn propagate_transforms(
...
+                    {
+                        *global_transform = p_global_transform.mul_transform(*transform);
+                    }
+                    last_child = Some((child, global_transform, children));
+                    child
+                },
             );
+            outbox.extend(new_children);
+
+            if depth >= max_depth {
```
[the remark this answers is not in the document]
aevyrie: Yeah, good call.

crates/bevy_transform/src/systems.rs
```diff
@@ -89,100 +145,423 @@ pub fn propagate_transforms(
...
+            //     the entire hierarchy.
             //
             // For example, consider the following malformed hierarchy:
             //
             //     A
             //   /   \
-            //  B     C
-            //   \   /
+            //  B     C 
+            //   \   /
```
pcwalton, line 189: uber-nit: trailing whitespace

crates/bevy_transform/src/systems.rs
```diff
@@ -89,100 +145,423 @@ pub fn propagate_transforms(
...
             // Also consider the following case, where A and B are roots:
             //
-            //  A       B
-            //   \     /
-            //    C   D
-            //     \ /
+            //  A       B 
+            //   \     / 
+            //    C   D 
+            //     \ /
```
pcwalton, line 201: uber-nit: trailing whitespace

crates/bevy_transform/src/systems.rs
```diff
@@ -89,100 +145,423 @@ pub fn propagate_transforms(
...
+                };
+                tasks.append(&mut extra_task);
+            }
+
+            // At this point, we know there is work to do, so we increment the busy thread counter,
+            // and drop the mutex guard *after* we have incremented the counter. This ensures that
+            // if another thread is able to acquire a lock, the busy thread counter will already be
+            // incremented.
+            queue.busy_threads.fetch_add(1, Ordering::Relaxed);
+            drop(rx); // Important: drop after atomic and before work starts.
```
pcwalton, line 370: Can we move the work queue logic to `bevy_tasks` or some place? Normally, I'm not big on generalizing prematurely, but from a code organization perspective it feels better to have `bevy_transform` be for higher level logic and `bevy_tasks` (or whatever) be for low-level parallelism nuts and bolts.

crates/bevy_transform/src/systems.rs
```diff
@@ -89,100 +145,423 @@ pub fn propagate_transforms(
...
+                // detection to mark dirty subtrees when the transform is mutated.
+                nodes.iter_many_unique_unsafe(UniqueEntityIter::from_iterator_unchecked(
+                    p_children.iter(),
+                ))
+            };
+
+            let mut last_child = None;
+            let new_children = children_iter.map(
+                |(child, (transform, mut global_transform), (children, child_of))| {
+                    assert!(child_of.get() == parent);
```
pcwalton, line 467: nit: Could be `assert_eq!` I think?

crates/bevy_transform/src/systems.rs
```diff
@@ -89,100 +145,423 @@ pub fn propagate_transforms(
...
+            queue.busy_threads.fetch_add(-1, Ordering::Relaxed);
+        }
+    }
+
+    /// Propagate transforms from `parent` to its non-leaf `children`, pushing updated child
+    /// entities to the `outbox`. Propagation does not visit leaf nodes; instead, they are computed
+    /// in [`compute_transform_leaves`](super::compute_transform_leaves), which can optimize much
+    /// more efficiently.
+    ///
+    /// This function will continue propagating transforms to descendents in a depth-first
```
pcwalton, line 404: typo: should be "descendants" for consistency

crates/bevy_transform/src/systems.rs
```diff
@@ -89,100 +145,423 @@ pub fn propagate_transforms(
...
+                // (`compute_transform_leaves`) and instead simply doing that here. However, that
+                // proved to be much slower than two pass for a few reasons:
+                // - it's less cache friendly and is outright slower than the tight loop in the
+                //   second pass
+                // - it prevents parallelism, as all children must be iterated in series
+                //
+                // The only way I can see to make this faster when there are many leaf nodes is to
+                // speed up archetype checking to make the iterator skip leaf entities more quickly,
+                // or encoding the hierarchy level as a component. That, or use some kind of change
+                // detection to mark dirty subtrees when the transform is mutated.
```
pcwalton, line 458: Yeah, I think we'll want to do the latter eventually. It sounds like what you have here is so fast that it tends to remove transformation propagation from the profile though.

crates/bevy_transform/src/systems.rs
```diff
@@ -89,100 +145,423 @@ pub fn propagate_transforms(
...
+                    break;
+                };
+                tasks.append(&mut extra_task);
+            }
+
+            // At this point, we know there is work to do, so we increment the busy thread counter,
+            // and drop the mutex guard *after* we have incremented the counter. This ensures that
+            // if another thread is able to acquire a lock, the busy thread counter will already be
+            // incremented.
+            queue.busy_threads.fetch_add(1, Ordering::Relaxed);
```
pcwalton, line 369: I *think* `Relaxed` is ok here because of the mutex, but you might want `Acquire` to make sure that the compiler doesn't reorder relative to the mutex unlock you're about to perform?

crates/bevy_transform/src/systems.rs
```diff
@@ -89,100 +145,423 @@ pub fn propagate_transforms(
...
+            // Try to acquire a lock on the work queue in a tight loop. Profiling shows this is much
+            // more efficient than relying on `.lock()`, which causes gaps to form between tasks.
+            let Ok(rx) = queue.receiver.try_lock() else {
+                core::hint::spin_loop(); // No apparent impact on profiles, but best practice.
+                continue;
+            };
+            // If the queue is empty and no other threads are busy processing work, we can conclude
+            // there is no more work to do, and end the task by exiting the loop.
+            let Some(mut tasks) = rx.try_iter().next() else {
+                if queue.busy_threads.load(Ordering::Relaxed) == 0 {
```
pcwalton, line 346: See next comment: does this need to be `Ordering::Release`? I don't think so, but not sure.

pcwalton (commented):
Looks good, just had a few questions. Overall thanks a ton for doing this work, it really helps!

crates/bevy_transform/src/systems.rs
aevyrie: To be honest, I'm not really sure. I found the docs pretty unhelpful.

aevyrie: > Possible values are [SeqCst], [Acquire] and [Relaxed].

crates/bevy_transform/src/plugins.rs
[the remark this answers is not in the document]
hymm: Makes sense. So it might be an improvement at low numbers, but hurts the critical path when there's enough work.

It might be worth tweaking the default batch size and retrying, but that can be done as a followup.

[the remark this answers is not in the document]
aevyrie: I'v updated to run `sync_simple_transforms` alongside `compute_transform_leaves`. We should probably tweak the batch sizes, as you say, because the current default leaves a lot of space in the schedule. I'lll leave that as future work.

crates/bevy_transform/src/systems.rs
aevyrie: I read the docs for cpp, and I've adjusted this to be a bit more correct, I think. No measurable impact to perf.

crates/bevy_transform/src/systems.rs
aevyrie: I tried to do this, but it ended up being a pretty major time sink, and I couldn't quite figure out how to express the lifetimes with generic queries. I've mentioned this exists to Alice to keep it in mind during future work. `nth` is working on some scheduler improvements, and is also aware of this work. I think that is probably a decent place to leave this for now, I might try extracting this logic again if I can make some time, but it's possible the work `nth` is doing will be usable here similar to rayon.

NthTensor:
Alright, I am finally going to be able to give this a look. Right out of the gate I find the increase in run variance concerning, but I'm in favor of merging and leaving that as follow up if we can't easily identify the cause.

mockersf (approved):

aevyrie:
> Alright, I am finally going to be able to give this a look. Right out of the gate I find the increase in run variance concerning, but I'm in favor of merging and leaving that as follow up if we can't easily identify the cause.

If you are referring to my comment, that was about bevy's *existing* system. It swings wildly +- 1ms. Variance (timing distribution) within a run is consistent for both. Variance between runs in this PR seems better than main. My guess is this is caused by E-cores vs P-cores. The old implementation usually ended up spending most time in a single core, whereas the new one is spread across all available cores in the thread pool. I think this explains why inter-run variance has improved.

aevyrie:
The CI failure was not caused by this PR, seems to be some sort of MESA failure according to discord.

alice-i-cecile:
Thank you to everyone involved with the authoring or reviewing of this PR! This work is relatively important and needs release notes! Head over to https://github.com/bevyengine/bevy-website/issues/1995 if you'd like to help out.

Issue #17108, opened 2025-01-03
Per `Entity` sprite picking (and maybe friends) opt-out


Phoqinu:
## What problem does this solve or what need does it fill?

Currently `sprite_picking` is relatively expensive, I did some light testing with tracy and 200k sprites with like only 2k of them being in camera view can result in the `sprite_picking` taking 500µs. Which is not much but frame time was 2.66ms so it took "only" almost 20% of frame time.

I don't need picking for my sprites and in `debug` mode - without any bevy optimizations - the performace penalty is so big that you can reach single digit fps.

My specific use-case aside I can imagine that some people would want picking only for some sprites and they might want to turn off picking for the rest.

## What solution would you like?

For `sprite_picking` a simple component: 
```rust
#[derive(Debug, Component)]
pub struct NoSpritePicking;
```
then in the `sprite_picking` system sprite query could be modified from: 
```rust
sprite_query: Query<(
        Entity,
        &Sprite,
        &GlobalTransform,
        Option<&PickingBehavior>,
        &ViewVisibility,
    )>,
```
to
```rust
sprite_query: Query<(
        Entity,
        &Sprite,
        &GlobalTransform,
        Option<&PickingBehavior>,
        &ViewVisibility,
    ), Without<NoSpritePicking>>,
```

Then users if they don't want picking for their sprites they can just add `NoSpritePicking` component to sprite entity.

## What alternative(s) have you considered?

I tried adding `PickingBehavior::IGNORE` to my sprites but that did not bring any speed-up.

Maybe it could result in some weird behavior with `PickingBehavior` ? idk

I think it would be nice to have similar marking ignore components for other backends.
Maybe even global `NoPicking` component or something ?

Faster and better `sprite_picking` system ?- It's currently doing filtering, sorting and filtering on all sprites...

Picking being opt-in ?

## Additional context

Red - current default system with/out `PickingBehavior`
Yellow - my proposed feature with `NoSpritePicking` on my sprites and modified system
200k sprites (2k on camera, rest off-camera), release mode

**(This is  for my use-case, ofc different apps won't probably see similar speed-up)**
![Image](https://github.com/user-attachments/assets/854e03b6-9da3-4b1e-97ac-4d704053d8ab)

alice-i-cecile:
We should consider making sprite picking opt-in instead.

rparrett:
Related #16903

notmd:
[...]

PR #17842, opened 2025-02-13
Make sprite picking opt-in

notmd:
# Objective

Fix https://github.com/bevyengine/bevy/issues/17108
See https://github.com/bevyengine/bevy/issues/17108#issuecomment-2653020889

## Solution

- Make the query match `&Pickable` instead `Option<&Pickable>`

## Testing

- Run the `sprite_picking` example and everything still work

## Migration Guide

- Sprite picking are now opt-in, make sure you insert `Pickable` component when using sprite picking.
```diff
-commands.spawn(Sprite: {..} );
+commands.spawn((Sprite: {..}, Pickable::default());
```

### DIFF
--- a/crates/bevy_sprite/src/picking_backend.rs
+++ b/crates/bevy_sprite/src/picking_backend.rs
@@ -89,7 +89,7 @@ fn sprite_picking(
         Entity,
         &Sprite,
         &GlobalTransform,
-        Option<&Pickable>,
+        &Pickable,
         &ViewVisibility,
     )>,
     mut output: EventWriter<PointerHits>,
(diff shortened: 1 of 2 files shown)


rparrett:
Related PR: #17348

chompaa:
I think having the ability to opt-out is still valuable (e.g. debugging). One solution that comes to mind is adding `SpritePickingPlugin::opt_in` (default `true`). We could have `sprite_picking` use generic `QueryData` and use `Option<&Pickable>` or `&Pickable` based on `opt_in`.

notmd:
I think there is still a way to make it opt-out but it hurts performance and potential footgun.

alice-i-cecile:
I think the best way for users to restore opt-out behavior (e.g. during debugging) is to just add an observer of their own, and add the required component automatically. There's no need for a dedicated setting for that.

fjkorf:
I looked over this update (and the preceding) and it all makes sense. I ran the `sprite_picking` example locally and it did, in fact, still work. Looks good to me.

PR #17847, opened 2025-02-13
Remove the id from `PointerId::Touch`

ickshonpe:
# Objective

Instead of storing winit's `u64` finger id for touches on the pointer entity in its `PointerId` component, just use the pointer entity itself to represent the touch within Bevy.

Related #17808

## Solution

* Changed `touch_pick_event` so its local `touch_cache` hashmap also tracks the entity id of touch pointers as well as the intial touch states.
* Removed the `deactivate_touch_pointers` system. Instead `touch_pick_event` despawns the corresponding touch pointer entity when a touch ends or is cancelled.
* Added a `pointer_entity: Option<Entity>` field to the `PointerInput` event. This is used to match events to the correct pointer for inputs that support multiple pointers.  For touch it would be used to distinguish between touches for multi-touch. 

Made `pointer_entity` optional so that this PR doesn't have to touch any of the other pointer event systems but it would probably make sense to require it (if this PR is adopted).

## Testing

Works when running the bevy examples in chrome with touch enabled in dev tools.

### DIFF
--- a/crates/bevy_picking/src/pointer.rs
+++ b/crates/bevy_picking/src/pointer.rs
@@ -33,8 +33,8 @@ pub enum PointerId {
     /// The mouse pointer.
     #[default]
     Mouse,
-    /// A touch input, usually numbered by window touch events from `winit`.
-    Touch(u64),
+    /// A touch input.
+    Touch,
     /// A custom, uniquely identified pointer. Useful for mocking inputs or implementing a software
     /// controlled cursor.
     #[reflect(ignore)]
@@ -44,7 +44,7 @@ pub enum PointerId {
 impl PointerId {
     /// Returns true if the pointer is a touch input.
     pub fn is_touch(&self) -> bool {
-        matches!(self, PointerId::Touch(_))
+        matches!(self, PointerId::Touch)
     }
     /// Returns true if the pointer is the mouse.
     pub fn is_mouse(&self) -> bool {
@@ -53,14 +53,6 @@ impl PointerId {
     /// Returns true if the pointer is a custom input.
     pub fn is_custom(&self) -> bool {
         matches!(self, PointerId::Custom(_))
-    }
-    /// Returns the touch id if the pointer is a touch input.
-    pub fn get_touch_id(&self) -> Option<u64> {
-        if let PointerId::Touch(id) = self {
-            Some(*id)
-        } else {
-            None
-        }
     }
 }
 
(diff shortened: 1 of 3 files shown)


alice-i-cecile:
My understanding from @aevyrie is that this doesn't work because of interrupted touches. They will be despawned in Bevy but keep their identifier.

ickshonpe:
I'm not sure what problems that could cause, maybe I'm misunderstanding what you mean by an interrupted touch? The winit docs say that their finger ids are only unique for the duration of a touch and if another touch event happens after a
`TouchPhase::Ended` or `TouchPhase::Cancelled` event with the same id, it should be interpreted as a new touch event.

alice-i-cecile:
To be clear, I much prefer this solution if it works. I'm not confident on my understanding of the behavior here.

fjkorf:
I worked through the code and believe I understand it. Seems sound. I was also able to run the default_picking example locally without error

aevyrie:
We already have some pointer id changes in flight, and I'm a bit worried about doing a partial switch to just one type of pointer input. I think we need to better understand the lifecycle of a pointer, and evaluate if the pointer id is needed at all.

I think the split to Entity + PointerKind is *probably* right, but last time I used Entity to identify pointers, it was causing things to blow up when part of the system tried to access an entity that didn't exist. If I recall correctly it was something like dragging outside a window would remove the pointer when it got to the edge of the screen, and when it returned and respawned, it caused drag/up events to explode.

I don't know if this is still the case, but that's why I'd like to have a bit more of a thorough understanding of the lifecycle of a pointer before changing the ID. The current state of most of this stuff is an accumulation of bug fixes from edge cases discovered in use.

crates/bevy_picking/src/pointer.rs
```diff
@@ -270,6 +262,8 @@ pub enum PointerAction {
 pub struct PointerInput {
     /// The id of the pointer.
     pub pointer_id: PointerId,
+    /// The pointer entity
+    pub pointer_entity: Option<Entity>,
```
aevyrie, line 266: Picking is already inconsistent enough across backends, I would strongly prefer if we didn't add more inconsistency, though I appreciate your reasoning.

aevyrie (commented):
Only one real comment, other than "changing `PointerId` without a better understanding of pointer entity lifecycles strikes a deep fear in my heart".

PR #17848, opened 2025-02-13
Add `NodeImageMode` to the UI prelude

ickshonpe:
# Objective

Add `NodeImageMode` to `bevy_ui::prelude`.


alice-i-cecile (approved):
SpriteImageMode is in the prelude so sure.

PR #17853, opened 2025-02-13
Combine `output_index` and `indirect_parameters_index` into one field in `PreprocessWorkItem`.

pcwalton:
The `output_index` field is only used in direct mode, and the `indirect_parameters_index` field is only used in indirect mode. Consequently, we can combine them into a single field, reducing the size of `PreprocessWorkItem`, which
`batch_and_prepare_{binned,sorted}_render_phase` must construct every frame for every mesh instance, from 96 bits to 64 bits.


alice-i-cecile (approved):
Good clarity, and a clever little trick.

tychedelia (approved):

PR #17855, opened 2025-02-13
Add `EntityDoesNotExistError`, replace cases of `Entity` as an error, do some easy Resultification

JaySpruce:
## Objective
There's no general error for when an entity doesn't exist, and some methods are going to need one when they get Resultified. The closest thing is `EntityFetchError`, but that error has a slightly more specific purpose.

## Solution
- Added `EntityDoesNotExistError`.
  - Contains `Entity` and `EntityDoesNotExistDetails`.
- Changed `EntityFetchError` and `QueryEntityError`:
  - Changed `NoSuchEntity` variant to wrap `EntityDoesNotExistError` and renamed the variant to `EntityDoesNotExist`.
- Renamed `EntityFetchError` to `EntityMutableFetchError` to make its purpose clearer.
- Renamed `TryDespawnError` to `EntityDespawnError` to make it more general.
- Changed `World::inspect_entity` to return `Result<[ok], EntityDoesNotExistError>` instead of panicking.
- Changed `World::get_entity` and `WorldEntityFetch::fetch_ref` to return `Result<[ok], EntityDoesNotExistError>` instead of `Result<[ok], Entity>`.
- Changed `UnsafeWorldCell::get_entity` to return `Result<UnsafeEntityCell, EntityDoesNotExistError>` instead of `Option<UnsafeEntityCell>`.

## Migration Guide
- `World::inspect_entity` now returns `Result<impl Iterator<Item = &ComponentInfo>, EntityDoesNotExistError>` instead of `impl Iterator<Item = &ComponentInfo>`.
- `World::get_entity` now returns `EntityDoesNotExistError` as an error instead of `Entity`. You can still access the entity's ID through the error's `entity` field.
- `UnsafeWorldCell::get_entity` now returns `Result<UnsafeEntityCell, EntityDoesNotExistError>` instead of `Option<UnsafeEntityCell>`.

### DIFF
--- a/crates/bevy_ecs/src/world/entity_fetch.rs
+++ b/crates/bevy_ecs/src/world/entity_fetch.rs
@@ -127,12 +127,10 @@ unsafe impl WorldEntityFetch for Entity {
         self,
         cell: UnsafeWorldCell<'_>,
     ) -> Result<Self::Mut<'_>, EntityMutableFetchError> {
-        let location =
-            cell.entities()
-                .get(self)
-                .ok_or(EntityMutableFetchError::EntityDoesNotExist(
-                    EntityDoesNotExistError::new(self, cell.entities()),
-                ))?;
+        let location = cell
+            .entities()
+            .get(self)
+            .ok_or(EntityDoesNotExistError::new(self, cell.entities()))?;
         // SAFETY: caller ensures that the world cell has mutable access to the entity.
         let world = unsafe { cell.world_mut() };
         // SAFETY: location was fetched from the same world's `Entities`.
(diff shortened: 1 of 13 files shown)


crates/bevy_ecs/src/world/entity_fetch.rs
```diff
@@ -199,22 +211,21 @@ unsafe impl<const N: usize> WorldEntityFetch for &'_ [Entity; N] {
...
 
         let mut refs = [const { MaybeUninit::uninit() }; N];
         for (r, &id) in core::iter::zip(&mut refs, self) {
-            let ecell = cell.get_entity(id).ok_or(EntityFetchError::NoSuchEntity(
-                id,
-                cell.entities().entity_does_not_exist_error_details(id),
-            ))?;
+            let ecell = cell
+                .get_entity(id)
+                .map_err(EntityMutableFetchError::EntityDoesNotExist)?;
```
chescock, line 228: If we `impl From<EntityDoesNotExistError> for EntityMutableFetchError`, then `?` will do the error mapping for us, right?  Would that be worth doing?

alice-i-cecile: I like that impl: I think it's always the correct mapping.

alice-i-cecile (approved):

alice-i-cecile:
@JaySpruce I'll merge this once the new conflicts are resolved.

JaySpruce:
@alice-i-cecile, good to go

PR #17857, opened 2025-02-14
Rustdoc html postprocessing

SpecificProtagonist:
# Objective

#17758 added tags indicating core Bevy traits to doc pages of types. This was implemented by adding JavaScript that modifies the DOM.
#17821 expands this to sort types in sections in module doc pages. Because no information about which item implements which trait is available in the module page, it employs a two step process where the source code is modified to invisibly embed this data. Besides being more complex, this leaves the repository in a dirty state when publishing.

Reimplement #17758 by postprocessing the html output instead. As a side benefit this works with JS disabled and allows using the same source of truth for trait tags and trait sections.

#17821 can then be reimplemented on top of this PR.

## Solution

- Implement a wrapper around `rustdoc` and use it on docs.rs & the docs ci workflow
- Keep getting the information about which types implement which trait from the html (instead of switching to reading rustdoc's json output), as the mapping from item to html file isn't available in an easily digestible format
- Gracefully degrade when the structure of rustdoc's html output changes in an incompatible way, so at worst it will be the same as if we didn't do this customization.

## Testing

Run
```bash
cargo build --package rustdoc-wrapper --release
cargo doc --config "build.rustdoc = \"target/release/rustdoc-wrapper\"" --workspace
```
and check out the doc. The tags should appear the same as on main.

### DIFF
--- a/tools/rustdoc-wrapper/src/main.rs
+++ b/tools/rustdoc-wrapper/src/main.rs
@@ -8,7 +8,7 @@ use std::{
     sync::LazyLock,
 };
 
-use nipper::Document;
+use dom_query::Document;
 use walkdir::WalkDir;
 
 fn main() {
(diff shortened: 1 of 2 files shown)


tools/rustdoc-wrapper/src/main.rs
```diff
@@ -0,0 +1,173 @@
...
+        let path = path.path();
+        if path.extension() == Some(OsStr::new("html")) {
+            let mut doc = Document::from(&read_to_string(path).unwrap());
+            post_process_type(&mut doc);
+            std::fs::write(path, doc.html().as_bytes()).unwrap();
+        }
+    }
+}
+
+// We only use the HTML and not rustdoc's JSON output because
```
mockersf, line 55: because? 😄

JeanMertz:
I love this.

I would also love for this to extend to the search feature (perhaps adding a fourth column). There seem to be many opportunities to improve the docs once this post-processing binary lands.

I also wonder if there's anything worth upstreaming (or at least suggesting) into rustdoc itself?

JeanMertz (approved):

JeanMertz:
I guess I'm using this as a dumping ground for suggestions, but here's another one that has bothered me often:

Whenever I search for a type, the first result (as seen in the above screenshot) that shows up is usually in the `prelude`. I would rather that one not be first, or completely filtered out/replaced, so that I go directly to the "source" module of the type, instead of any re-export location.

GuillaumeGomez:
So if I understand correctly what you're trying to do here, you want for each type to list which components it requires, right? If so, instead of generating it like this, why not generating a markdown file you include like this:

```rust
#[cfg_attr(doc, doc = include!(concat!(DOC_DIR, "ITEM_NAME-traits.md"))]
```

The only thing you need then is to generate this file for all your items. You can pretty easily write a small rustc extension to do it for you and update these files before generating docs.

However:

1. This is not fully automatic, you need to add the `#[cfg_attr()]` attribute wherever needed for a type
2. You need to rerun the command to update the doc files

Luckily it can be mitigated with (proc-)macros that will add the doc file (if needed) for you, preventing doc generation to work if the file is empty or was not created.

mockersf:
I don't think this is a good approach, and one we will want to maintain.

I would prefer to solve #17821 by including markdown files as suggested by @GuillaumeGomez

alice-i-cecile:
I think this is too intrusive / unstable to build on top of docs.rs. I like the core concept, but I think this is likely to be a maintenance headache.

mockersf:
We want to reduce the custom work we do on the doc, not increase it. 

#17758 was borderline OK as it was very limited, but I would rather the work be done to improve things on rustdoc side, by proposing new features or style improvements, than piling more things on Bevy side

BenjaminBrienen:
Closing as Wont-Fixed based on feedback from maintainers and lack of new ideas in this direction.,

PR #17858, opened 2025-02-14
Preserve spawned RelationshipTarget order and other improvements

### LINKED ISSUE #17720: Children in spawned gltf scenes are incorrectly ordered

cart:
Fixes #17720

## Objective

Spawning RelationshipTargets from scenes currently fails to preserve RelationshipTarget ordering (ex: `Children` has an arbitrary order). This is because it uses the normal hook flow to set up the collection, which means we are pushing onto the collection in _spawn order_ (which is currently in archetype order, which will often produce mismatched orderings).

We need to preserve the ordering in the original RelationshipTarget collection. Ideally without expensive checking / fixups.

## Solution

One solution would be to spawn in hierarchy-order. However this gets complicated as there can be multiple hierarchies, and it also means we can't spawn in more cache-friendly orders (ex: the current per-archetype spawning, or future even-smarter per-table spawning). Additionally, same-world cloning has _slightly_ more nuanced needs (ex: recursively clone linked relationships, while maintaining _original_ relationships outside of the tree via normal hooks).

The preferred approach is to directly spawn the remapped RelationshipTarget collection, as this trivially preserves the ordering. Unfortunately we can't _just_ do that, as when we spawn the children with their Relationships (ex: `ChildOf`), that will insert a duplicate.

We could "fixup" the collection retroactively by just removing the back half of duplicates, but this requires another pass / more lookups / allocating twice as much space. Additionally, it becomes complicated because observers could insert additional children, making it harder (aka more expensive) to determine which children are dupes and which are not.

The path I chose is to support "opting out" of the relationship target hook in the contexts that need that, as this allows us to just cheaply clone the mapped collection. The relationship hook can look for this configuration when it runs and skip its logic when that happens. A "simple" / small-amount-of-code way to do this would be to add a "skip relationship spawn" flag to World. Sadly, any hook / observer that runs _as the result of an insert_ would also read this flag. We really need a way to scope this setting to a _specific_ insert.

Therefore I opted to add a new `RelationshipInsertHookMode` enum and an `entity.insert_with_relationship_insert_hook_mode` variant. Obviously this is verbose and ugly. And nobody wants _more_ insert variants. But sadly this was the best I could come up with from a performance and capability perspective. If you have alternatives let me know!

There are three variants:

1. `RelationshipInsertHookMode::Run`: always run relationship insert hooks (this is the default)
2. `RelationshipInsertHookMode::Skip`: do not run any relationship insert hooks for this insert (this is used by spawner code)
3. `RelationshipInsertHookMode::RunIfNotLinked`: only run hooks for _unlinked_ relationships (this is used in same-world recursive entity cloning to preserve relationships outside of the deep-cloned tree)

Note that I have intentionally only added "insert with relationship hook mode" variants to the cases we absolutely need (everything else uses the default `Run` mode), just to keep the code size in check. I do not think we should add more without real _very necessary_ use cases.

I also made some other minor tweaks:

1. I split out `SourceComponent` from `ComponentCloneCtx`. Reading the source component no longer needlessly blocks mutable access to `ComponentCloneCtx`.
2. Thanks to (1), I've removed the `RefCell` wrapper over the cloned component queue.
3. (1) also allowed me to write to the EntityMapper while queuing up clones, meaning we can reserve entities during the component clone and write them to the mapper _before_ inserting the component, meaning cloned collections can be mapped on insert.
4. I've removed the closure from `write_target_component_ptr` to simplify the API / make it compatible with the split `SourceComponent` approach.
5. I've renamed `EntityCloner::recursive` to `EntityCloner::linked_cloning` to connect that feature more directly with `RelationshipTarget::LINKED_SPAWN`
6. I've removed `EntityCloneBehavior::RelationshipTarget`. This was always intended to be temporary, and this new behavior removes the need for it.

### DIFF
--- a/benches/benches/bevy_ecs/entity_cloning.rs
+++ b/benches/benches/bevy_ecs/entity_cloning.rs
@@ -77,7 +77,7 @@ fn reflection_cloner<B: Bundle + GetTypeRegistration>(
     for component in component_ids {
         builder.override_clone_behavior_with_id(component, ComponentCloneBehavior::reflect());
     }
-    builder.recursive(recursive);
+    builder.linked_cloning(recursive);
 
     builder.finish()
 }
@@ -136,7 +136,7 @@ fn bench_clone_hierarchy<B: Bundle + Default + GetTypeRegistration>(
         reflection_cloner::<B>(&mut world, true)
     } else {
         let mut builder = EntityCloner::build(&mut world);
-        builder.recursive(true);
+        builder.linked_cloning(true);
         builder.finish()
     };
 
--- a/crates/bevy_ecs/macros/src/component.rs
+++ b/crates/bevy_ecs/macros/src/component.rs
@@ -207,7 +207,7 @@ pub fn derive_component(input: TokenStream) -> TokenStream {
         .unwrap_or(quote! { #bevy_ecs_path::component::Mutable });
 
     let clone_behavior = if relationship_target.is_some() {
-        quote!(#bevy_ecs_path::component::ComponentCloneBehavior::RelationshipTarget(#bevy_ecs_path::relationship::clone_relationship_target::<Self>))
+        quote!(#bevy_ecs_path::component::ComponentCloneBehavior::Custom(#bevy_ecs_path::relationship::clone_relationship_target::<Self>))
     } else {
         quote!(
             use #bevy_ecs_path::component::{DefaultCloneBehaviorBase, DefaultCloneBehaviorViaClone};
--- a/crates/bevy_ecs/src/relationship/mod.rs
+++ b/crates/bevy_ecs/src/relationship/mod.rs
@@ -12,7 +12,7 @@ pub use relationship_source_collection::*;
 
 use crate::{
     component::{Component, HookContext, Mutable},
-    entity::{ComponentCloneCtx, Entity},
+    entity::{ComponentCloneCtx, Entity, SourceComponent},
     system::{
         command::HandleError,
         entity_command::{self, CommandWithEntity},
@@ -73,7 +73,24 @@ pub trait Relationship: Component + Sized {
     fn from(entity: Entity) -> Self;
 
     /// The `on_insert` component hook that maintains the [`Relationship`] / [`RelationshipTarget`] connection.
-    fn on_insert(mut world: DeferredWorld, HookContext { entity, caller, .. }: HookContext) {
+    fn on_insert(
+        mut world: DeferredWorld,
+        HookContext {
+            entity,
+            caller,
+            relationship_insert_hook_mode,
+            ..
+        }: HookContext,
+    ) {
+        match relationship_insert_hook_mode {
+            RelationshipInsertHookMode::Run => {}
+            RelationshipInsertHookMode::Skip => return,
+            RelationshipInsertHookMode::RunIfNotLinked => {
+                if <Self::RelationshipTarget as RelationshipTarget>::LINKED_SPAWN {
+                    return;
+                }
+            }
+        }
         let target_entity = world.entity(entity).get::<Self>().unwrap().get();
         if target_entity == entity {
             warn!(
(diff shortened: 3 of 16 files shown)


rparrett:
It would be great to see a test that verifies that order is preserved.

I attempted to write one that spawns a scene containing a hierarchy into a world with a different set of archetypes, but it seems that I don't understand the issue well enough to produce a test that fails on main despite the thorough explanation of the issue.

cart:
Done! I added a test that forces the archetype init order to mismatch the child order during spawn. This fails on main and passes on this branch.

benches/benches/bevy_ecs/entity_cloning.rs
```diff
@@ -77,7 +77,7 @@ fn reflection_cloner<B: Bundle + GetTypeRegistration>(
     for component in component_ids {
         builder.override_clone_behavior_with_id(component, ComponentCloneBehavior::reflect());
     }
-    builder.recursive(recursive);
+    builder.linked_cloning(recursive);
```
mockersf, line 80: should the `recursive` variable be renamed?

cart: Yup!

cart: Yup!

crates/bevy_ecs/macros/src/component.rs
```diff
@@ -255,7 +260,12 @@ pub fn derive_component(input: TokenStream) -> TokenStream {
     })
 }
 
-fn visit_entities(data: &Data, bevy_ecs_path: &Path, is_relationship: bool) -> TokenStream2 {
+fn visit_entities(
+    data: &Data,
+    bevy_ecs_path: &Path,
+    is_relationship: bool,
+    is_relatonship_target: bool,
```
andresovela, line 267:
```suggestion
    is_relationship_target: bool,
```

crates/bevy_ecs/src/relationship/mod.rs
```diff
@@ -73,7 +73,24 @@ pub trait Relationship: Component + Sized {
...
+    ) {
+        match relationship_insert_hook_mode {
+            RelationshipInsertHookMode::Run => {}
+            RelationshipInsertHookMode::Skip => return,
+            RelationshipInsertHookMode::RunIfNotLinked => {
+                if <Self::RelationshipTarget as RelationshipTarget>::LINKED_SPAWN {
+                    return;
+                }
+            }
+        }
```
eugineerd, line 93: I wonder if this has to run inside the hook. Is it possible to have a more generic system (e.g. `insert_with_hooks_run_mode(..., <bitset of hooks/observers to ignore/run>)`) that would allow to ignore a set of hooks/observers on insert, remove, etc? This functionality is something I wanted to use for allowing moving non-cloneable components in `EntityCloner`, but to do that I need to first run the remove hooks, then bitwise-copy data and only then remove data from the entity (copying before running hooks is unsound because hooks can move data out of component and lead to double free).

Using the generic ignored hooks system, `Run` and `Skip` should be trivial to implement and `RunIfNotLinked` can probably be emulated by allowing `ComponentCloneCtx` to set hook run flags. That way `ComponentCloneBehavior` for the relationship component will be responsible for skipping `LINKED_SPAWN` components instead.

alice-i-cecile:
Feedback on the minor changes. Grumble grumble about them being rolled into one PR though!

> 1. I split out SourceComponent from ComponentCloneCtx. Reading the source component no longer needlessly blocks mutable access to ComponentCloneCtx.

Great!

> 2. Thanks to (1), I've removed the RefCell wrapper over the cloned component queue.

Always nice.

> 3. (1) also allowed me to write to the EntityMapper while queuing up clones, meaning we can reserve entities during the component clone and write them to the mapper before inserting the component, meaning cloned collections can be mapped on insert.

Good.

> 4. I've removed the closure from write_target_component_ptr to simplify the API / make it compatible with the split SourceComponent approach.

This is much nicer.

> 5. I've renamed EntityCloner::recursive to EntityCloner::linked_cloning to connect that feature more directly with RelationshipTarget::LINKED_SPAWN

Good, I think this is clearer.

6. I've removed EntityCloneBehavior::RelationshipTarget. This was always intended to be temporary, and this new behavior removes the need for it.

Sure :)

crates/bevy_ecs/src/relationship/mod.rs
cart: I see the appeal, but doing bitwise stuff here and externalizing this would mean we pay the price of set operations / checks for every insert, which will have a cost (note that a bitset of componentids could easily require allocations). Additionally, not running hooks in many cases could be a safety / correctness issue. Its worth having this discussion but I don't think this is the place to resolve it, and I'd want benchmarks to show there are no major regressions.

PR #17861, opened 2025-02-14
Add `TypePath::crate_version`

MarcGuiselin:
# Objective

For the purposes of modding or scripting, it becomes valuable to know not just from what crate a type is originating from, but also the version of that crate.

If we imagine a fairly simple example, like a game (v0.2.0) and separate level editor (v0.1.0) compiled with two slightly different versions of bevy. They communicate with each other using brp. Suppose a modder opens the editor and tries to move an object in the game by altering it's Transform. As long as bevy reflect is able to deserialize the message on the game's side, there won't be an issue. But if there is a version mismatch between bevy_transform in the game and in the editor that results in a deserialization error, nothing will move and instead they will see a vague error in the game's logs.

Adding this version info opens up the possibility of adding valuable context to serialization errors (for example that a type is from bevy_transform 0.1.0 but remote version is from bevy_transform 0.2.0).

In my wip bevy modding framework, this version information would be useful to differentiate the types used by different mods and determine which mods are incompatible/outdated compared to others.

## Solution

This pr adds this version info to reflected types.

## Testing

- Compiles fine on nightly and stable

## Migration guide

- Manual implementers of `TypePath` should eventually implement `crate_version()`. However, their code will continue to compile due to the default trait implementation that returns a `None`.

### DIFF
--- a/crates/bevy_asset/src/io/embedded/mod.rs
+++ b/crates/bevy_asset/src/io/embedded/mod.rs
@@ -144,7 +144,7 @@ macro_rules! embedded_path {
     }};
 
     ($source_path: expr, $path_str: expr) => {{
-        let crate_name = env!("CARGO_CRATE_NAME");
+        let crate_name = option_env!("CARGO_CRATE_NAME").unwrap_or("");
         $crate::io::embedded::_embedded_asset_path(
             crate_name,
             $source_path.as_ref(),
--- a/crates/bevy_asset/src/io/embedded/mod.rs
+++ b/crates/bevy_asset/src/io/embedded/mod.rs
@@ -144,7 +144,7 @@ macro_rules! embedded_path {
     }};
 
     ($source_path: expr, $path_str: expr) => {{
-        let crate_name = module_path!().split(':').next().unwrap();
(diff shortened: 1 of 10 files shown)


MarcGuiselin:
I expect the change for determining `crate_name` to be a point of contention. I thought this would be an improvement, but this can be easily reverted if we don't like the changes it brings.

One of the side effects of using `env!("CARGO_CRATE_NAME")` instead of `module_path!().split(':').next().unwrap()` is that for crate names with dashes like `my-crate`, the new implementation will return the crate name with a dash, while the old implementation did not.

MarcGuiselin:
Hello @cart, was the closing and reopening of this PR intentional? If there is an interest in merging I can try to get this up to date with main this weekend.

cart:
The closing and reopening was part of an automated process to fix PRs which were broken by a history rewrite on the main branch. I'd say hold off for now / leave it as-is. I suspect we might want this, especially as editors / plugins / migrations start to become a thing, but I'm a bit hesitant as this is more per-type data compiled into binaries, and we have a ton of types.

PR #17863, opened 2025-02-14
Split out the `IndirectParametersMetadata` into CPU-populated and GPU-populated buffers.

pcwalton:
The GPU can fill out many of the fields in `IndirectParametersMetadata` using information it already has:

* `early_instance_count` and `late_instance_count` are always initialized to zero.

* `mesh_index` is already present in the work item buffer as the `input_index` of the first work item in each batch.

This patch moves these fields to a separate buffer, the *GPU indirect parameters metadata* buffer. That way, it avoids having to write them on CPU during `batch_and_prepare_binned_render_phase`. This effectively reduces the number of bits that that function must write per mesh from 160 to 64 (in addition to the 64 bits per mesh *instance*).

Additionally, this PR refactors `UntypedPhaseIndirectParametersBuffers` to add another layer, `MeshClassIndirectParametersBuffers`, which allows abstracting over the buffers corresponding indexed and non-indexed meshes. This patch doesn't make much use of this abstraction, but forthcoming patches will, and it's overall a cleaner approach.

This didn't seem to have much of an effect by itself on `batch_and_prepare_binned_render_phase` time, but subsequent PRs dependent on this PR yield roughly a 2× speedup.

### DIFF
--- a/crates/bevy_render/src/batching/gpu_preprocessing.rs
+++ b/crates/bevy_render/src/batching/gpu_preprocessing.rs
@@ -624,10 +624,10 @@ pub struct PreprocessWorkItem {
     pub input_index: u32,
 
     /// In direct mode, the index of the mesh uniform; in indirect mode, the
-    /// index of the [`IndirectParametersMetadata`].
+    /// index of the [`IndirectParametersGpuMetadata`].
     ///
     /// In indirect mode, this is the index of the
-    /// [`IndirectParametersMetadata`] in the
+    /// [`IndirectParametersGpuMetadata`] in the
     /// `IndirectParametersBuffers::indexed_metadata` or
     /// `IndirectParametersBuffers::non_indexed_metadata`.
     pub output_or_indirect_parameters_index: u32,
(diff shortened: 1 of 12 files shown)


tychedelia (approved):
Lgtm! Tested on macOS.

crates/bevy_render/src/batching/gpu_preprocessing.rs
```diff
@@ -823,72 +831,162 @@ where
...
+    pub fn clear(&mut self) {
+        self.indexed.clear();
+        self.non_indexed.clear();
+    }
+}
+
+/// The buffers containing all the information that indirect draw commands use
+/// to draw the scene, for a single mesh class (indexed or non-indexed), for a
+/// single phase.
+pub struct MeshClassIndirectParametersBuffers<IP>
```
superdump, line 937: While I'm not super keen on the word 'class', I don't have a better suggestion. I like how much gets ripped out of this.

superdump (approved):

PR #17865, opened 2025-02-14
Adding More Comments to the Embedded Assets Example

Carter0:
# Objective

I noticed when I was looking at the embedded assets example that there wasn't any comments on it to indicate what an embedded asset is and why anyone would want to make one.

## Solution

I added some more comments to the example that gives more detail about embedded assets and how they work. Feel free to be aggressive with rewriting these comments however, I just think the example could use something haha.

### DIFF
--- a/examples/asset/embedded_asset.rs
+++ b/examples/asset/embedded_asset.rs
@@ -1,7 +1,11 @@
 //! Example of loading an embedded asset.
 
 //! An embedded asset is an asset included in the program's memory, in contrast to other assets that are normally loaded from disk to memory when needed.
-//! An embedded asset can reduce loading times since assets are immediately available, rather than being on disk.
+//! The below example embeds the asset at program startup, unlike the common use case of embedding an asset at build time. Embedded an asset at program startup can be useful
+//! for things like loading screens, since it might be nice to display some art while other, non-embedded, assets are loading.
+
+//! Embedded assets are often created when the executable is being made. By embedding an asset at build time rather than runtime the program never needs to go to disk for the asset at all,
+//! since it is already located in the program's binary executable.
 use bevy::{
     asset::{embedded_asset, io::AssetSourceId, AssetPath},
     prelude::*,
@@ -28,10 +32,6 @@ impl Plugin for EmbeddedAssetPlugin {
     }
 }
 
-// This function loads the asset's binary from memory at runtime.
-// Whereas other asset data needs to be loaded from disk, this data is already in memory.
-// This means the asset can be loaded much quicker since it is faster for the processor to grab
-// memory already in use.
 fn setup(mut commands: Commands, asset_server: Res<AssetServer>) {
     commands.spawn(Camera2d);
 
--- a/examples/asset/embedded_asset.rs
+++ b/examples/asset/embedded_asset.rs
@@ -1,8 +1,6 @@
 //! Example of loading an embedded asset.
 
-// An embedded asset is an asset included in the programs memory, in contrast to other assets that are normally loaded from disk to memory when needed. 
-// It's normally only used for small engine-related assets so they can be used without having them in your own asset directory.
-
+// An embedded asset is an asset included in the program's memory, in contrast to other assets that are normally loaded from disk to memory when needed. 
 // An embedded asset can reduce loading times since assets are immediately available, rather than being on disk.
 use bevy::{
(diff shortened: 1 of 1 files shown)


Carter0:
Can I add labels to this? Or is that beyond my current power levels?

alice-i-cecile:
I think it's beyond your current power level, but let me fix that with a Bevy Org invite :)

examples/asset/embedded_asset.rs
```diff
@@ -1,5 +1,7 @@
 //! Example of loading an embedded asset.
 
+// An embedded asset is an asset included in the program's memory, in contrast to other assets that are normally loaded from disk to memory when needed.
+// An embedded asset can reduce loading times since assets are immediately available, rather than being on disk.
```
alice-i-cecile, line 4:
```suggestion
//! An embedded asset is an asset included in the program's memory, in contrast to other assets that are normally loaded from disk to memory when needed.
//! An embedded asset can reduce loading times since assets are immediately available, rather than being on disk.
```

examples/asset/embedded_asset.rs
```diff
@@ -15,6 +17,7 @@ fn main() {
 
 struct EmbeddedAssetPlugin;
 
+// This plugin embeds the asset's binary into the program's memory at runtime.
```
alice-i-cecile, line 20:
```suggestion
```

I don't think this line is  particularly helpful.

alice-i-cecile (approved):
Good stuff :) A few suggestions:

1. You can probably get a cleaner flow by merging the first two sentences of this file.
2. The other benefit of embedded assets is easier distribution, since there's no separate asset files to worry about. That's why Bevy ships default fonts and shaders like this.
3. The speedup isn't a particularly strong argument: it just moves the loading from a loading screen to project startup. That's usually not desired: actions should show immediate effect. I would only use this feature for packaging convenience or for the initial assets used by a loading screen.

Carter0:
> Good stuff :) A few suggestions:

Thanks for the response ❤️ 

> 2. The other benefit of embedded assets is easier distribution, since there's no separate asset files to worry about. That's why Bevy ships default fonts and shaders like this.

Yeah, I was thinking about that. It feels like this feature really shines at embedding assets when the executable is being made, rather than when the program is first launched. That way there is no runtime cost. 

I'm guessing the 'cargo build' command would be how someone would take their Bevy project and turn it into an executable that someone could run by just double clicking an icon? Is there an example that describes how to use this embedded assets at build time rather than at runtime? 

I suppose I can just mention the benefits of this feature in regards to building assets at the top of this example file.

> 3. The speedup isn't a particularly strong argument: it just moves the loading from a loading screen to project startup. That's usually not desired: actions should show immediate effect. I would only use this feature for packaging convenience or for the initial assets used by a loading screen.

That makes sense. I never really thought about it, but it would be a little bit odd to load a loading screen asset haha. I'll add something like that to the comments as well.

examples/asset/embedded_asset.rs
```diff
@@ -1,5 +1,11 @@
 //! Example of loading an embedded asset.
 
+//! An embedded asset is an asset included in the program's memory, in contrast to other assets that are normally loaded from disk to memory when needed.
+//! The below example embeds the asset at program startup, unlike the common use case of embedding an asset at build time. Embedded an asset at program startup can be useful
+//! for things like loading screens, since it might be nice to display some art while other, non-embedded, assets are loading.
+
+//! Embedded assets are often created when the executable is being made. By embedding an asset at build time rather than runtime the program never needs to go to disk for the asset at all,
```
alice-i-cecile, line 7: "Often created when the binary is being made" is a bit unclear to me.

examples/asset/embedded_asset.rs
```diff
@@ -1,5 +1,11 @@
 //! Example of loading an embedded asset.
 
+//! An embedded asset is an asset included in the program's memory, in contrast to other assets that are normally loaded from disk to memory when needed.
+//! The below example embeds the asset at program startup, unlike the common use case of embedding an asset at build time. Embedded an asset at program startup can be useful
```
alice-i-cecile, line 4: TBH, we should probably demonstrate the typical usage instead.

Carter0: Yeah I agree, but I don't know how to do it :(

fjkorf (commented):
This writeup makes sense to me.

PR #17868, opened 2025-02-15
Disable sprite picking in bevymark

aevyrie:
# Objective

- Don't waste CPU cycles in a benchmark

## Solution

- Don't add the picking plugin

## Testing

- Tracy

### DIFF
--- a/examples/stress_tests/bevymark.rs
+++ b/examples/stress_tests/bevymark.rs
@@ -14,7 +14,7 @@ use bevy::{
         render_asset::RenderAssetUsages,
         render_resource::{Extent3d, TextureDimension, TextureFormat},
     },
-    sprite::AlphaMode2d,
+    sprite::{AlphaMode2d, SpritePlugin},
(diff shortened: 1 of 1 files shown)


hukasu:
what is the difference in performance setting `SpritePlugin::add_picking` to false vs disabling https://docs.rs/bevy/latest/bevy/prelude/struct.DefaultPickingPlugins.html

rparrett:
This should probably also be done in `many_sprites` and `many_animated_sprites`.

But there are also a few existing efforts to
- Make picking for an individual sprite opt-in, which should mostly remove the perf penalty:
  #17842
- Not add the sprite picking plugin by default, among other things, which would completely remove the perf penalty:
  #17348

I'd personally rather wait for one of those. This benchmark is telling us that we have a problem that needs to be addressed.

> SpritePlugin::add_picking

TIL. Between that setting, disabling the plugin, and disabling the feature, I feel like we have far too many ways to do this.

PR #17870, opened 2025-02-15
Correctly resolve third party bevy crates via `bevy_macro_utils`

### LINKED ISSUE #17816: `BevyManifest` incorrectly assumes that all crates starting with `bevy_` are exposed through `::bevy::`
When I modified `BevyManifest`, I thought that it was bevy internal. However it is also used by https://github.com/bevyengine/bevy_editor_prototypes/blob/f41e2240174e29f1c9206bb6b080431463466ea5/crates/bevy_bsn/src/macros/src/derive_construct.rs#L15 and possibly in other crates too. As long as only official bevy crates are resolved though it, there are no issues.

However, should a user crate that extends bevy and is named `bevy_...` use the `BevyManifest` in their proc macros, it will incorrectly assume their crate is available through `::bevy::`.

A fix would be to only strip the prefix from the `bevy_` crates that are actually available through `::bevy::`.

I can create a pr for this and see two options:
1) List dir in `crates/` to get all the bevy internal crates. More complex but should automatically keep up with bevy but possibly yield wrong results if crates are added that are not rexported through bevy.
2) Just using a manual string list. Simpler but needs to be kept in sync manually.

raldone01:
Fixes #17816.

## Objective

Currently `BevyManifest` incorrectly assumes all crates starting with `bevy_` are official bevy crates that are re-exported through the main `bevy` crate.

## Solution

- `BevyManifest` should handle third party crates starting with `bevy_` correctly.

## Testing

- I ran the ci.
- It should be verified that the `build.rs` does not break when uploaded to `crates.io`.

## Migration Guide

- Users of `BevyManifest::get_subcrate` should migrate to `BevyManifest::maybe_get_path` or `BevyManifest::get_path`.


crates/bevy_macro_utils/src/official_bevy_crates.rs
```diff
@@ -0,0 +1,49 @@
+// This file is automatically generated by build.rs
+
+/// List of official bevy crates.
+/// This is used to correctly resolve re-exported third-party crates starting with `bevy_`.
+pub const OFFICIAL_BEVY_CRATES: &[&str] = &[
```
alice-i-cecile, line 5: @mockersf can we share this list between this and our publishing tooling somehow?

NyxAlexandra (commented):
I'm pretty sure convention is to use `OUT_DIR`, as `src/` is supposed to br immutable

raldone01:
> I'm pretty sure convention is to use `OUT_DIR`, as `src/` is supposed to be immutable

That won't work. athe file must be committed to source control.

See https://github.com/bevyengine/bevy/issues/17816#issuecomment-2660932114.
Otherwise when `bevy_macro_utils` is pulled as a dependency, it does not have the files necessary to generate the official crates list.

raldone01:
I marked it as a draft again because I was concerned, that publishing could trigger the following error introduced via https://github.com/rust-lang/cargo/issues/5073:
```
error: failed to verify package tarball

Caused by:
  Source directory was modified by build.rs during cargo publish. Build scripts should not modify anything outside of OUT_DIR.
...
```

I ran `cargo publish --dry-run --allow-dirty -p bevy_macro_utils` but it worked fine.

When `cargo publish` builds the package, it is being built in a separate directory: `Current working directory: "/argon_bfs/user_main/argon_bfs_desktop/bevy_dev/bevy/target/package/bevy_macro_utils-0.16.0-dev"`.
Since the parent directory is not `crates` no files in `src` are modified.

raldone01:
I am now confident that the `build.rs` should not cause any issues during publishing or when `bevy_macro_utils` is pulled in as dependency.

alice-i-cecile (approved):
I think this fix is mandatory for 0.16, and I'm content with this solution for generating the crate lists for now. We might be able to piggyback off Cargo.toml metadata in the future to make this less messy.

raldone01:
If you know a better way just ping me and I can try to make it happen.
Also if you want to somehow make the official package list available to the ci, I think it would have to be moved to a separate package.

I checked all `Cargo.toml` files that are shipped with `bevy_macro_utils` but they don't contain a list of all official bevy packages.

Would you be interested in a `bevy_build_info` package that contains the following kind of information?

```rust
/// This is information about the bevy build and the user build.
#[derive(Debug, Clone, Eq, PartialEq, Hash, Serialize, Deserialize)]
pub struct BuildInfo {
  // Version information
  pub cargo_pkg_version: Semver,
  pub git_version: String,
  pub git_submodule_versions: Vec<(String, String)>,

  // Build information
  pub build_date: DateTime<Utc>,
  pub library_linking_type: LibraryLinkingType,
  pub enabled_features: Vec<String>,
  pub overflow_checks: bool,
  pub panic_strategy: String,
  pub relocation_model: String,

  // Target information
  pub target_os: String,
  pub target_family: String,
  pub target_arch: String,
  pub target_features: Vec<String>,
  pub target_pointer_width: usize,
  pub target_endianness: Endianness,
  pub target_env: String,
  // Target information less relevant
  pub target_has_atomic: Vec<String>,
  pub target_has_atomic_equal_alignment: Vec<String>,
  pub target_has_atomic_load_store: Vec<String>,

  // Compiler information
  pub rustc_version: String,
  pub rustup_toolchain: String,
  pub target_triple: String,
  pub host_triple: String,
  pub opt_level: String,
  pub debug: bool,
  pub profile: String,
  pub rustflags: Vec<String>,
  // Linker information
  pub linker: String,
  pub clang_version: String,
  pub lld_version: String,
  pub ld_version: String,
  pub mold_version: String,
}

/// This is information about the machine that is running bevy
pub struct GPUInformation {
  pub model: String,
  pub vendor: String,
  pub memory_total: usize,
  pub memory_used: usize,
  pub memory_available: usize,
}

/// This is information about the machine that is running bevy
pub struct MachineInformation {
  // OS Information
  pub os_name: String,
  pub os_version: String,
  pub kernel_version: String,
  pub current_working_directory: PathBuf,
  pub executable_path: PathBuf,
  /// Time since the machine booted
  pub machine_uptime: Duration,

  // CPU Information
  pub cpu_count: usize,
  pub cpu_model: String,
  pub cpu_vendor: String,
  /// The frequency of the CPU in MHz.
  pub cpu_frequency: Option<f64>,
  /// L1, L2, L3 in KB
  pub cpu_cache_sizes: BTreeMap<String, usize>,

  // GPU Information
  pub gpu_information: Vec<GPUInformation>,

  // IO Information
  /// The amount of installed memory in bytes.
  pub memory_total: usize,
  /// The amount of used memory in bytes.
  pub memory_used: usize,
  /// The amount of free memory in bytes.
  pub memory_free: usize,

  /// The amount of total SWAP in bytes.
  pub swap_total: usize,
  /// The amount of used SWAP in bytes.
  pub swap_used: usize,
  /// The amount of free SWAP in bytes.
  pub swap_free: usize,

  /// The total disk space in bytes of the drive where the executable is located.
  pub disk_usage_total: usize,
  /// The used disk space in bytes of the drive where the executable is located.
  pub disk_usage_used: usize,
  /// The free disk space in bytes of the drive where the executable is located.
  pub disk_usage_free: usize,

  pub active_network_interfaces: Vec<String>,

  // Important Environment Variables
  pub rust_log: Option<String>,
  pub rust_backtrace: Option<String>,
  pub path: String,

  // Environment Variables Library Loading
  pub ld_preload: Option<String>,
  pub ld_library_path: Option<String>,
  pub dyld_library_path: Option<String>,

  // Environment Variables Proxy
  pub http_proxy: Option<String>,
  pub https_proxy: Option<String>,
  pub no_proxy: Option<String>,

  // Environment Variables Locale
  pub lang: Option<String>,
  pub lc_all: Option<String>,
  pub lc_messages: Option<String>,

  // Environment Variables Timezone
  pub tz: Option<String>,
  pub tzdata: Option<String>,
  pub tzdir: Option<String>,

  // Environment Variables remaining
  pub environment_variables: BTreeMap<String, String>,
}
```

The above is unrefined and just here to give you an idea of what I did for my own project.

alice-i-cecile:
> If you know a better way just ping me and I can try to make it happen.

Could we explore the dependency tree of the `bevy` crate? I've played a bit with cargo manifest tools and think that should work; I'm just not sure if that information will be available where we need it.

raldone01:
I can investigate it in the coming weeks. Feel free to merge this whenever though. I will just create a new pr when I get around to doing it.

PR #17871, opened 2025-02-15
Staged Components

ElliottjPierce:
# TLDR

This pr builds on  #17701 for the agenda in  #17569. This has a number of advantages described below. Further, this seems to have no performance cost in the grand scheme of things. 

# Motivation

Right now, component registration requires mutable access to a world's components. However, registering a component is not considered a state change, since it doesn't actually spawn or modify any entity in the world. 

This discrepancy has been an annoyance. It prevents truly readonly queries, multithreaded component access (for things like assets as entities), and it generally encumbers processes that would be much simpler if registering components and retrieving their information could be done with only immutable access. 

As a cherry on top, allowing component registration immutably also further clarifies that this is not considered a state change, which make it more self-documenting.

It is my understanding that the Bevy community as a whole wants some form of this to be implemented *if* it can be done for *free*. Ergonomics would not be worth a performance loss here.

# History

 I had originally prototyped this in #17569 but am now splitting that PR into others. #17671 was the first step in this.

#17701 Generalizes the staging mechanism in #17569. This is how we afford immutable registration for practically free.

I recommend looking at #17701 for details and discussion around why this design is being explored. There has been lots of great discussion about it, and I'm still very open to new ideas.

Per @alice-i-cecile request, before  #17701 is merged, we should review how it will be used. This pr is how it will be used.

# Objective

The goal here is to implement the staging mechanism built in #17701 on components. This does not change how it is used right now. World components and the general user experience are unchanged. If this pr gets a green light, I will open a future one changing the backing world component storage to the staging implementation.

# Costs

The biggest cost is the increased complexity, but I feel it's worth it. In terms of performance though, this appears to be totally free!

On my M2 Max computer, the original components implementation ran in 9.3 microseconds. Almost all of the new implementations are much faster:

Screenshot 2025-02-15 at 10 26 06 PM

The top line "directly" is the new no nonsense implementation at 8.4 microseconds. There was a hash set clone that I changed to a `.iter().collect::<SmallVec>()` which I belvie is what caused the speedup.

In the benchmarks:
- "stage" means we had mutable access incidentally to the synchronized collection.
- "lock" means we had to lock the synchronized collection.
- "sync" means we to lock and unlock the collection in between each registration (very unrealistic for batch registration).
- "eager" means we had to do locking checks before applying the staged changes.

In short, the only new implementation that was as slow as the original was "ArcStageOnWrite sync eager", which is also the least likely to be used.

Further, since these times are all so close to each other, we can confirm that the locking cost of the staging system is practically free. That's not to say it can't be improved, but for every use, this component implementation is already  faster than the previous.

I also see potential ways to make this even faster in terms of syncing, but I don't want to get ahead of myself.

The biggest concern is the performance cost when the locking causes a block. But that blocking would last for a maximum of a few microseconds (if it is for a batch registration), and it would be incredibly unlikely. Component registration is already very rare. What are the odds that two threads try to register brand new components in the same couple microseconds? Pretty much zero.

So unless anyone has other performance concerns, I would think this is free from a performance perspective.

# Future work

If this is merged, I will next work on changing the backing structure of the world. I might make a controversial pr for renaming and generalizing other Component usages since now, Components really just mean ColdComponents. There are lots of interfaces to clean up too by easing up mutable access requirements.

# Breaking changes

I have recorded these in the migration guide. Note that there will be related breaking changes in future PRs that might ought to be grouped together.

## Migration Guide

To generalize `Components`, a number of associated functions have been moved to `ComponentsReader` and `ComponentsWriter`.

`ComponentsReader` now includes:
- `len`
- `is_empty`
- `get_info`
- `get_name`
- `get_info_unchecked`
- `get_id`
- `component_id`
- `get_resource_id`
- `resource_id`

`ComponentsWriter` now includes:
- `register_component`
- `register_component_with_descriptor`
- `register_required_components_manual`
- `register_resource`
- `register_resource_with_descriptor`
- `register_non_send`

These functions are no longer available on `Components` without the respective trait being used in the module.

Additionally, the `ComponentsInfo::required_components` function has been removed. Use `ComponentsReader::get_required_components` instead. This returns `RequiredComponentsRef`, which shares `iter_ids` with the `RequiredComponents`. If you need full `RequiredComponents`, use `RequiredComponentsRef::get_all`.

Similarly, `World::get_required_components` and `World::get_required_components_by_id` now returns `RequiredComponentsRef` instead of `&RequiredComponents`.

`Component::register_required_components`'s parameter `components: &mut Components` now takes  `&mut impl ComponentsWriter`. This will only affect manual `Component` implementations. Other than updating the signature and potentially importing `ComponentsReader`, this will likely not break the implementation body.

# Open questions

Should we implement ComponentsWriter, etc on the StageOnWrite types themselves? This could give more speed when trying to register existing components and a few other use cases, but it would also mean more code. Maybe for a future pr?

Should we add ComponentsWriter and ComponentsReader to the prelude? We could implement these on `World` and add it to the prelude, which might be desirable but IDK.

Should we use StageOnWrite or ArcStageOnWrite for world storage? StageOnWrite could be faster, but ArcStageOnWrite could be fully sent between threads. This might be especially useful for next-gen scene spawning. If the arcs are the same, we could skip component id translation for entity cloning between worlds. This might give enough of a performance edge to be worth it. IMO we do StageOnWrite for now and re-evaluate when we know more about bsn and can do some proper benchmarks.


crates/bevy_ecs/src/component.rs
```diff
@@ -1146,147 +1162,371 @@ pub struct Components {
...
+        storage.indices.extend(self.indices.drain());
+        storage
+            .resource_indices
+            .extend(self.resource_indices.drain());
+
+        for (id, additional_required_components) in self.additional_required_components.drain() {
+            // SAFETY: ids are known to be valid and cold.
+            unsafe { storage.components.get_mut(id.0).debug_checked_unwrap() }
+                .required_components
+                .merge(&additional_required_components);
```
ElliottjPierce, line 1197: Should we be merging the additional requirements into the cold, keeping the cold RequiredComponent if they overlap, or should we be doing the opposite, inserting the updated set into the cold? I'm not aware of a way there will be any overlap in the first place, so I don't know that it matters, but I think its worth clarifying. IDK what the best approach is here.

crates/bevy_ecs/src/component.rs
```diff
@@ -1538,173 +1728,596 @@ impl Components {
...
+                RequiredByStagedRef {
+                    working: &info.required_by,
+                    cold: None,
+                }
+            }
+        })
+    }
+}
+
+impl ComponentsInternalReader for Stager<'_, StagedComponents> {
```
ElliottjPierce, line 1943: This is mostly duplicated code. We could add new traits of each of the ComponentsWriter, ComponentsReader, ComponentsInternalWriter, and ComponentsInternalReader that include lifetimes. That would solve the duplicated code problem, but it would also be messier. I am open to either, but I started with the simpler option. If anyone feels strongly otherwise, I can make that abstraction, or I can even do it in a future PR if we like.

crates/bevy_ecs/src/component.rs
```diff
@@ -1146,147 +1162,371 @@ pub struct Components {
...
-            #[expect(
-                deprecated,
-                reason = "need to use this method until it is removed to ensure user defined components register hooks correctly"
-            )]
-            // TODO: Replace with `info.hooks.update_from_component::<T>();` once `Component::register_component_hooks` is removed
-            T::register_component_hooks(&mut info.hooks);
+/// The [`Deref`] trait ties the lifetime of the returned reference to the lifetime of the value itself.
+/// However, when the returned reference has nothing to do with the value containing it, this behavior is undesirable.
+/// This trait gets around that by putting the lifetime in the trait, allowing a number of niche conviniencies.
+pub trait DerefByLifetime<'a>: Deref {
```
ElliottjPierce, line 1216: This should probably be moved to bevy_utils or something, but I'm not sure where it would be best for this to live. I've left it here for now, but if anyone has a destination in mind, let me know.

This lets us avoid *some* of the "reference to dropped value" errors. I think schedule.rs has a good example of that being super useful. Line 1897 I think.

crates/bevy_ecs/src/component.rs
```diff
@@ -1146,147 +1162,371 @@ pub struct Components {
...
+        -> ComponentId;
+
+    /// Registers a [non-send resource](crate::system::NonSend) of type `T` with this instance.
+    /// If a resource of this type has already been registered, this will return
+    /// the ID of the pre-existing resource.
+    fn register_non_send<T: Any>(&mut self) -> ComponentId;
+}
+
+/// This trait provides easily misused read only access to [`Component`] collections intended for use only within this crate.
+pub(crate) trait ComponentsInternalReader: ComponentsReader {
```
ElliottjPierce, line 1504: These methods were originally pub(crate) so I left it that way for now, but could these be made public safely? If so, we could roll this whole trait into `ComponentsReader` and loose the "unused" lint. 

This might be useful for exposing the shape of the requirement tree to users. For example, this could let the bevy editor display a graph of component requirements visually, which might be a good debugging and docs tool. IDK.

crates/bevy_ecs/src/component.rs
```diff
@@ -1713,6 +2326,90 @@ impl Components {
     }
 }
 
+/// Allows modifying required components with potential staged changes.
+pub(crate) struct RequiredComponentsStagedMut<'a> {
```
ElliottjPierce, line 2330: Because the naming here has been confusing in #17569, I want to go ahead and clarify it. Unless anyone has a better way of naming this, I'll add this explanation to the docs before merging.

For all these types:

`working` refers to the most changeable version of the data. Every component must have a working field. If the working is needed immutably, (`RequiredByStagedRef` and `RequiredComponentsStagedRef`), it can be in cold or staged. If it needs to be mutable, (`RequiredByStagedMut` and `RequiredComponentsStagedMut`), it needs to come from staged, so we create it in staged storage if it doesn't exist already and return the reference. (unless we have mutable access to cold storage in which case we can just use that, ex Components.)

`cold` refers to any tacked on extra data in addition to `working`. I've named it cold because the only place it can ever come from is cold storage. If there's anything in staged, staged will always go to `working`, since that is the "most changeable"/closest to the "top". `cold` is used if and only if there was staged data to put in `working` *and* cold data to put in `cold`. If there is only one of those, it goes in `working` regardless of it it came from cold or staged storage, hence that being named `working` instead of staged or something.

Hopefully that makes sense, but let me know if it doesn't. Better naming ideas are always welcome of course.

crates/bevy_ecs/src/component.rs
chescock: To clarify: 

The duplication is between `impl ComponentsInternalReader for Stager` and `impl ComponentsInternalReader for StagedRef`.  The difference between them is that `Stager` has `&mut T` while `StagedRef` has `&T`.  There is an `as_staged_ref()` method to convert, but that gives you an owned `StagedRef`, so `self.as_staged_ref().get_required_components(id)` would fail because it's borrowing from a temporary.  

Is that right?  

Would it make sense to extract common functions that take `cold: &'a Components, staged: &'a ComponentsStaged` along with the other parameters, and have the trait methods delegate to those?  

---

We have similar issues with reference-like types in `Query`.  The general pattern I'd recommend is having methods take `StagedRef` by *value*.  You can `impl Copy for StagedRef` so that true references aren't actually consumed, and you can use the existing `Stager::as_staged_ref()` to implement methods on `Stager` that reborrow and delegate to `StagedRef`.  The traits make that a little more awkward, but I think it still works.  

But `StagedRef` is just a pair of references, so instead of taking one by value, it seems simpler to just write a function that takes the two references.  Then you can even make it a method on one type or the other.

crates/bevy_ecs/src/component.rs
ElliottjPierce: For now, you're absolutely right. This particular PR I think could get away without this interface. However, one of the open questions is if we want to implement ComponentsReader/Writer and friends on the StageOnWrite coordinator type. (Either `StagedOnWrite` or `ArcStagedOnWrite`). 

Consider the following scenario: Thread A wants to get `ComponentInfo` for a component in cold storage. Thread B wants to register a brand new component. Currently, for thread A has to lock to get its info, where the lifetime of the info is tied to the lock guard, `StageRef` in this case. But that guard also blocks mutable access to staged/hot storage, even though all the `ComponentInfo` only lives on cold. That means thread A is going to be blocking thread B, or worse, thread B is going to be blocking thread A.

However, if we implement ComponentsReader/Writer and friends on the StageOnWrite coordinator type. (Either `StagedOnWrite` or `ArcStagedOnWrite`), this problem goes away. Instead of returning `&ComponentInfo` as the `DerefByLifetime` we can return a `MappedRwReadLock<ComponentInfo>` (Or a stand in type until that is stabilized.) That way, thread A only retains a guard to cold storage, and thread B can register its Component without blocking. This seems pretty desirable to me.

The reason I didn't just go ahead and do this is because we need to have more discussion over whether to use `StagedOnWrite` or `ArcStagedOnWrite`. The implementation details are a little different depending on which one we use. 

We also need to be careful with which locks we keep when returning `DerefByLifetime` types. For example, if thread A was requesting a readonly list of the components required by a component, we would also need to give a lock guard on the staged components, even though there may not be any relevant data there. This is because thread B could then register a component that requires the component thread A was interested in, and now, thread A is dealing with outdated information. This is a case where they really would need to block.

I think I might be able to safely remove this trait, but I suspect we would just need to add it back for the next phase of #17569. I will defer to you all as to which way to proceed, but hopefully that gives some clarity on why I created this trait. I think keeping it will make the future PRs easier to follow and review, since I won't have to change this work to work with it when/if we add it back in.

crates/bevy_ecs/src/component.rs
ElliottjPierce: > self.as_staged_ref().get_required_components(id) would fail because it's borrowing from a temporary.

Bingo! Duplicating the code was the simplest solution, so I wanted to start review with that approach.

> it seems simpler to just write a function that takes the two references

Kinda? IMO the extra code, docs, and such make this unappealing to me. In #17569, I did this (though it was an associated function), and if I remember correctly, you had said (and I agree) that the extra indirection is a little confusing.

> I'd recommend is having methods take StagedRef by value.

I really like that approach. I was initially trying to preserve dyn trait compatibility, so I had originally dismissed this, but since I ended having to use `impl DerefByLifetime`, that ship has already sailed. 

Only downside is that where we used to have `&impl ComponentsWriter` we will now need `impl ComponentsWriter + Copy`, and same with other traits. Kinda annoying, but maybe better than duplicated code? Unless anyone has another idea, I will implement this solution maybe tomorrow.

crates/bevy_ecs/src/component.rs
chescock: > For example, if thread A was requesting a readonly list of the components required by a component, we would also need to give a lock guard on the staged components, even though there may not be any relevant data there. This is because thread B could then register a component that requires the component thread A was interested in, and now, thread A is dealing with outdated information. This is a case where they really would need to block.

This is one of the other bits I'm still working on understanding.  Why do we keep `required_components` in cold storage at all?  (I haven't been following the required components implementation closely, so I may be missing something obvious.)  

For the main `ComponentInfo` struct, the value is that we can read it from cold storage with no locks even while unrelated components are being added.  But for `required_components`, we always need to read from *both* places, to account for any newly-staged additions.  Right?  

It seems like it would be simpler to store those outside of the `ComponentInfo` in a parallel `Vec<(RequiredComponents, HashSet<ComponentId>)>` that lives only in `StagedComponents`.  Then the cold `Vec<ComponentInfo>` would be fully immutable.  We wouldn't lose any parallelism, since readers needed `StagedComponents` anyway.  And we'd never have to deal with types that are part cold and part staged, which might cut down on the complexity.  

Would that be possible?  And if it is, would it be any simpler?  

A variation on that idea would be to have `required_components` be only in `ComponentInfo`, and not have the `additional_` fields in `StagedComponents`.  New components could go to staging with their static required components, but adding runtime required components would require `&mut World` to update cold components in place.

ElliottjPierce: > In general, if you have an impl DerefByLifetime<'a, T> value, you can always get the &'a T out and then drop the value.

Huh. I guess you're right. I hadn't anticipated people explicitly dropping the value while retaining the reference. To fix that, we would need to also tie the lifetime to the object being dereferenced, but now that's back to just `Deref`. I can understand that. I'll probably need to switch to just `Deref` then. That will add some complexity, but it should be safer to your point.

> Why do we keep `required_components` in cold storage at all?

Great question. Long term, we don't. Especially if we go for `ArcStagedOnWrite`. But, right now, the backing data structure for the world is just `Components`. So it still needs to be there so `Components` can implement `ComponentsWriter` and friends. It may also still be desirable to keep these implemented on `Components`. See `StagableWritesCore::maybe_stage` for example. If we want to do batch registration, and there are no staged changes, and there are no active lock guards on anything, doing the bulk registration would be faster on `Components` than on `Stager`. If we want to keep that door open, we need to have a copy of all relevant data on `Components` and only store changes on `StagedComponents`. Hopefully that makes sense.

If this is merged, the next step is to pick a backing `StageOnWrite` type, and transition `World` to use that. Once that happens, I'll do another PR to try to simplify any remaining architecture. Ex: I would also like to move the required components info out of `ComponentInfo` and try to keep `ComponentInfo` immutable everywhere. But that's best done in future PRs IMO. 

> &mut World to update cold components in place.

Although that is possible, it kinda goes against the goal here IMO. We want to be able to toss `RefStageOnWrite` or `ArcStagedOnWrtire` around, share them between threads, etc. I don't particularly like gating some actions around having full access. This would be possible with `ArcStagedOnWrtire` though. With that, you could get `&mut Components` from `&ArcStagedOnWrtire`, if you're ok with blocking.

ElliottjPierce: I'll push a commit up here in a bit moving to just `&`s instead of `impl DerefByLifetime`s. I'll make it work with taking `self` instead of `&self` too. I think for the special ways of minimally locking for `ComponentInfo`, we can just implement them directly on the `StageOnWrite` type.

chescock: > > Why do we keep `required_components` in cold storage at all?
> 
> Great question. Long term, we don't. Especially if we go for `ArcStagedOnWrite`.
> ...
> If this is merged, the next step is to pick a backing `StageOnWrite` type, and transition `World` to use that. Once that happens, I'll do another PR to try to simplify any remaining architecture. Ex: I would also like to move the required components info out of `ComponentInfo` and try to keep `ComponentInfo` immutable everywhere. But that's best done in future PRs IMO.

Okay!  I'm wondering if there is a simpler path where we do things like refactoring `ComponentInfo` to be immutable *first*.  There is a lot of code here to review, and I'd like to skip reviewing anything that won't be in the final version :).  

Suppose we do one PR to extract `required_components` and make `ComponentId` immutable, followed by one PR to go directly to `ArcStagedOnWrite`.  And instead of `ArcStagedOnWrite` being generic over `T: StagedChanges` we start by making it specific to `Components`.  Would that get us to the right place?  How many of these types and traits would we no longer need?

ElliottjPierce: > refactoring ComponentInfo to be immutable first.

I could do that, but it I don't think it would buy us very much. I think there's a lot of code required to carry out the immutable `ComponentInfo` idea, and there's even been some loose discussion of making it static to the `Component` trait. I also think the immutability is related to this PR but not relevant to it if that makes since. That is, this idea works wether or not we make it immutable, so it doesn't feel right to me blocking this PR until/if we make `ComponentInfo` immutable.

> There is a lot of code here to review, and I'd like to skip reviewing anything that won't be in the final version :).

Totally agree here, and thank you for your time reviewing this. I'm trying to do incremental changes as best as I can. That does mean, some things will be changed in future PRs, but I. think that's better than just one colossal PR. And I don't think much will change anyway. Just some minor details I think.

> And instead of `ArcStagedOnWrite` being generic over `T: StagedChanges` we start by making it specific to `Components`.

That was the approach I took in #17569, but since there's already been discussion of reusing the interface for reflection and even some render world data structures, I think keeping it generic is the right move. Much of the generic stuff has also already been reviewed in #17701. I'm not saying don't review it, I'm just saying it's not the focus of the PR, and there has already been lots of discussion of details there IIRC.

ElliottjPierce: Quick update.

> I'll push a commit up here in a bit moving to just &s instead of `impl DerefByLifetimes`. I'll make it work with taking `self` instead of `&self` too. I think for the special ways of minimally locking for `ComponentInfo`, we can just implement them directly on the StageOnWrite type.

The first part went ok. I did one commit to remove a potential foot gun and another to remove `DerefByLifetime`.

> I'll make it work with taking `self` instead of `&self`.

I spent longer than I'd like to admit working o this. The goal was to be able to associate lifetimes with the return type of some functions. This would cut down on duplicated code regarding `ComponentsInternalReader::get_required_by`, `ComponentsInternalReader::get_required_components`, and `ComponentsReader::get_info_unchecked`.

For `ComponentsInternalReader` I could do this relatively easily. But to keep it as a super trait of `ComponentsInternalWriter` (which is needed so that the massive `ComponentsWriter` trait doesn't need duplicated code for its implementations), I would have to do one of two things. I could make Copy a super trait of `ComponentsInternalReader`, but then I have to implement the whole family of these traits on &T and &mut T instead of just T.  That would be duplicated code, which defeats the point. I could alternatively change the super trait to `where &Self: ComponentsInternalReader`, but that adds complexity and may introduce more duplicated code when we get to situations like `& &mut T`.

`ComponentsReader::get_info_unchecked` is more problematic. I can't move it out to a different trait without having to copy the implementations of `ComponentsReader::get_info` and `ComponentsReader::get_name`. It also inherits the other problems from `ComponentsInternalReader`, only worse because the trait has to also still support methods with `&self`.

Finally, I tried making functions for this to fix the duplication more directly, but since they also require information from `ComponentsReader`, we're back to square one. I could just remove that dependency by inlining the functions I need, but again, that's more duplicated code.

Ultimately, there are only three relatively small functions being duplicated. I'm the first to admit that I don't like duplicated code, but in this case, I think the least complex way to get rid of it would be to throw the whole thing in a declarative macro. I would be happy to do that, (It's a pattern I've seen and used other places.) but I'm not so sure it's worth it here.

Hopefully that ramble made sense. Let me know if you have any other ideas.

crates/bevy_ecs/src/component.rs
ElliottjPierce: I have gone ahead and added this to the docs. If there are any other questions or better naming ideas, let me know, but for now, I'm going to mark this as resolved.

crates/bevy_ecs/src/component.rs
```diff
@@ -1146,147 +1156,312 @@ pub struct Components {
...
+    fn is_id_staged(&self, id: ComponentId) -> bool;
+
+    /// Gets the [`ComponentRegistrationStatus`] for a [`ComponentId`].
+    /// See [`ComponentRegistrationStatus`] and its variants for more
     #[inline]
-    pub unsafe fn get_info_unchecked(&self, id: ComponentId) -> &ComponentInfo {
-        debug_assert!(id.index() < self.components.len());
-        // SAFETY: The caller ensures `id` is valid.
-        unsafe { self.components.get_unchecked(id.0) }
+    fn get_registration_status(&self, id: ComponentId) -> ComponentRegistrationStatus {
```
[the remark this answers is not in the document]
ElliottjPierce: Another great question. 

First, I don't want to add anything to `ComponentInfo` because that's another thing to store and update that can just be calculated as needed.

I added this for completeness, but I can imagine it being used for things like profiling or as a utility for future functionality here (ex. `match get_registration_status() ...` ). 

Anyway, its free additional information to those who might end up needing it.

crates/bevy_ecs/src/component.rs
```diff
@@ -1330,22 +1504,76 @@ impl Components {
...
+            .register_by_id(required, constructor, 0);
 
         // Add the requiree to the list of components that require the required component.
         // SAFETY: The component is in the list of required components, so it must exist already.
         let required_by = unsafe { self.get_required_by_mut(required).debug_checked_unwrap() };
-        required_by.insert(requiree);
+        required_by.working.insert(requiree);
+
+        // Get required components inherited from the `required` component.
+        let inherited_requirements: Vec<(ComponentId, RequiredComponent)> = {
```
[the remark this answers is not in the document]
ElliottjPierce: > Is that right?

Exactly. I can't keep the helper method in `ComponentsWriter` because of privacy. I can't move it to `ComponentsInternalWriter` because there aren't enough super traits to have a default implementation there, so it would result in duplicate code. Instead, I inlined it.

> That's going to make it awkward if we ever want to refactor this code to have other helper methods like register_inherited_required_components, but I don't have any ideas on how to make that easier.

Me neither. This is the best I got. I suppose we could add function definitions inside of the functions themselves, but I think the comments are just as clean.

crates/bevy_ecs/src/component.rs
```diff
@@ -1146,147 +1156,312 @@ pub struct Components {
...
-                deprecated,
-                reason = "need to use this method until it is removed to ensure user defined components register hooks correctly"
-            )]
-            // TODO: Replace with `info.hooks.update_from_component::<T>();` once `Component::register_component_hooks` is removed
-            T::register_component_hooks(&mut info.hooks);
+impl StagedChanges for StagedComponents {
+    type Cold = Components;
+
+    fn apply_staged(&mut self, storage: &mut Self::Cold) {
+        storage.components.append(&mut self.components);
```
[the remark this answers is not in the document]
ElliottjPierce: Should clear it automatically. See [here](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.append).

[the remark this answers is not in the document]
chescock: Oh, sorry!  I kept reading it as `extend`.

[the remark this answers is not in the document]
ElliottjPierce: No worries. I did literately the same thing on discord earlier lol.

crates/bevy_ecs/src/component.rs
```diff
@@ -1114,16 +1113,13 @@ impl ComponentCloneBehavior {
 
     /// Set clone handler based on `Reflect` trait.
     #[cfg(feature = "bevy_reflect")]
-    pub fn reflect() -> Self {
+    pub fn reflect_handler() -> Self {
```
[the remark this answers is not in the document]
ElliottjPierce: I'll mark this as resolved, but let me know if you see anything else like this. I had a nasty merge lol.

ElliottjPierce:
Just a quick reminder that #17701 is still under review and has some changes that are not reflected here.

My plan is to merge that first, and then merge main into this branch, applying any changes from  #17701. Hopefully that makes sense.

It looks like GitHub doesn't have a convinient way to update this branch ahead of time, but if I'm missing a simple solution, let me know.

crates/bevy_ecs/src/component.rs
ElliottjPierce: I have gone ahead and unified these to make migration easier. If anyone has concerns, let me know.

crates/bevy_utils/src/staging.rs
```diff
@@ -0,0 +1,1119 @@
+#![expect(
+    unsafe_code,
+    reason = "This module marks some items as unsafe to alert users to deadlock potential."
+)]
+//! Provides an abstracted system for staging modifications to data structures that rarely change.
+//! See [`StageOnWrite`] as a starting point.
+//!
+//! # Rational
```
cBournhonesque, line 8:
```suggestion
//! # Rationale
```

ElliottjPierce:
There are a few improvements I can make here after #18028 is merged. Just cleaning up some undocumented behavior.

ElliottjPierce:
#16519 Fixed a code quality issue with the previous components implementation regarding `register_inherited_required_components`. I see what the idea was there, but I had to inline `register_inherited_required_components` here to preserve privacy rules. See [this comment](https://github.com/bevyengine/bevy/pull/17871/#discussion_r1966566892) for why. So, my most recent merging main into this branch, effectively undid #16519 on this branch.

I Believe I can add back `register_inherited_required_components` and apply the changes in #16519, but doing so would require either adding even more traits or changing the visibility of some methods. Either are possible, but I am trying to keep this PR as slim as I can for the sake of review. If this is merged, I will open an issue to try to resolve this.

Specifically, the options I found are to either change the visibility of `ComponentsInternalWriter`'s `get_required_components_mut`, `get_required_by_mut`, and `register_descriptor` to private (by moving them to `ComponentsPrivateWriter`, as these are not currently used outside the components module), or to add another extension trait for `T: ComponentsInternalWriter` to implement `ComponentsInternalWriter`'s `register_required_components_manual_unchecked` and `register_required_components`.

If a strong consensus is expressed here between those options, I would be happy to go ahead and add that to this PR. I would personally prefer to change the visibility unless anyone knows of plans to use them somewhere else in bevy_ecs. But in the meantime, I'll leave this for later.

crates/bevy_ecs/src/entity/clone_entities.rs
```diff
@@ -779,8 +779,8 @@ impl<'w> EntityClonerBuilder<'w> {
             self.entity_cloner.filter.remove(&id);
         }
         if self.attach_required_components {
-            if let Some(info) = self.world.components().get_info(id) {
-                for required_id in info.required_components().iter_ids() {
+            if let Some(info) = self.world.components().get_required_components(id) {
```
andriyDev, line 782: `info` -> `required_components`

crates/bevy_ecs/src/entity/clone_entities.rs
```diff
@@ -799,8 +799,8 @@ impl<'w> EntityClonerBuilder<'w> {
             self.entity_cloner.filter.insert(id);
         }
         if self.attach_required_components {
-            if let Some(info) = self.world.components().get_info(id) {
-                for required_id in info.required_components().iter_ids() {
+            if let Some(info) = self.world.components().get_required_components(id) {
```
andriyDev, line 802: `info` -> `required_components`

crates/bevy_ecs/src/component.rs
```diff
@@ -1492,208 +1713,504 @@ impl Components {
...
 
-    /// Type-erased equivalent of [`Components::component_id()`].
     #[inline]
-    pub fn get_id(&self, type_id: TypeId) -> Option<ComponentId> {
-        self.indices.get(&type_id).copied()
+    fn is_id_staged(&self, id: ComponentId) -> bool {
+        self.cold.len() <= id.0
+    }
+
+    fn get_required_components(&self, id: ComponentId) -> Option<RequiredComponentsRef> {
```
andriyDev, line 1789: This implementation is identical to the one for `Stager<'_, StagedComponents>`. I'm not sure if this is possible, but can we make a free generic function that takes `impl ComponentsReader` and implements this function? Alternatively, can maybe we should split up this trait to have the parts that are needed for this impl.

I believe the same logic applies for get_required_by.

andriyDev (changes requested):

alice-i-cecile:
@Soulghost, do you have opinions on that last question from @ElliottjPierce?

crates/bevy_ecs/src/component.rs
ElliottjPierce: You are correct. This is duplicated code and it's painful to me as well.

Unfortunately, I could not find a solution to this that didn't involve even more duplicated code. See [this comment](https://github.com/bevyengine/bevy/pull/17871#discussion_r1957233563) followed by [this one](https://github.com/bevyengine/bevy/pull/17871#discussion_r1966291109). 

The short version is, most solutions end up being more complex than the duplicated code. If we really want to dedupe it, I think the best solution is probably just to throw it in a macro. The helper method approach from #17569 was more confusing IIRC. But personally, I think the duplicated code, painful as it is, is the best option. But if a different consensus is reached, I'm happy to implement another solution too.

But I think deduping it might best be brought up as its own PR/Issue since there are so many options:
- change lifetimes of traits
- add delegate functions
- use macros
- create another trait for core of `Stager` and `StagedRef` and auto implement for T: that trait
- use unsafe code to force the lifetimes to be valid (cuz they are, at least in this version of the code.)

Again, if there is a consensus here, I'll implement it, but otherwise, I think we should come up with a fix separately from this PR IMO.

ElliottjPierce: I've done some more digging, and I want to clarify a little bit more.

I'll use `get_required_components` as an example.

For `StagedRef`, the lifetime of the `RequiredComponentsRef` it returns is `'a`, the same lifetime it immutably borrows. That means, we can read from the `StagableWrites` type with `StagedRef`, call `get_required_components`, and then drop the `StagedRef` without invalidating the lifetime. This is really useful.

For `Stager`, the lifetime of the `RequiredComponentsRef` it returns is `'s` as in `&'s self`, which includes `'a`. That means, we can't drop `StagedRef` without invalidating the lifetime of the returned `RequiredComponentsRef`. That coupling of lifetimes is annoying, but that's how rust is preventing some bugs. If the lifetimes were not coupled, `Stager` could then be modified and resized, causing a dangling pointer and a use after free danger. So, while it's annoying, it's worth it.

In other words, the two implementations technically have different return types. So treating these as the same signature ends up taking the more restrictive approach to be safe. So, right now, `ComponentsReader` technically follows this more restrictive path, but we could change this in the future. For example, we may need to change this to keep some references valid after we change the backing components structure in `World`. If so, any unification of these implementations would have to be undone. For example, delegate functions wouldn't work, and the unsafe magic described above would become unsound. The macro solution would still work, which is why it's my favored solution at the moment.

Anyway, that's why I think we should let the dust settle a bit here before deduping it. It's a tricky problem to solve and the way we solve it has implications regarding how restrictive our lifetimes may need to be, which will determine if we need to pass around dummy values that just hold lifetimes or not. So I think we should cross that bridge when/if we get there.

andriyDev: I'm not sure that's correct. The function roughly looks like:

```rust
impl ComponentsReader for StagedRef<'_, StagedComponents> {
    fn get_required_components(&self, id: ComponentId) -> Option<RequiredComponentsRef> {
        ...
    }
}
```

The lifetime ellision rules say if there's one lifetime in the input, then the output will use that lifetime for all outputs. Here's a simplified example https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=5ff4f5011cd25264d8d19f455efd9d87 It turns out it considers `&self` a single lifetime even if `&self` is actually `&'a StagedRef<'s>`

This should happen in both cases, so I don't see the difference here.

Putting that aside, I am totally fine to punt this to a future PR.

andriyDev (approved):

crates/bevy_ecs/src/component.rs
ElliottjPierce: Kinda.

When you change it to a trait, the lifetime is tied to self too. Here's your example in a trait: [example](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=ef2100e1e3a2f2249eb375e763d14ac2) Note that it no longer compiles with the drop.

Hopefully that explains my thought process here. When we change the backing structure of the world's components, we may need to add lifetimes to the trait, which would make the duplicated code diverge.

That said, I'm no lifetime expert, so maybe I'm missing something.

alice-i-cecile:
Closing in favor of #18173.

PR #17874, opened 2025-02-15
Auto sync point before exclusive system despite `*_ignore_deferred`

### LINKED ISSUE #17828: Unflushed commands before exclusive system when configured with `*_ignore_deferred`
## Bevy version

I saw this behavior at around 0.14 I think but only now got around making a minimal reproducible example on `main`, or at least a fairly recent commit behind `main`.

## What you did

I ran this code:

```rs
use bevy::prelude::*;

#[derive(Resource)]
struct InsertTest;

fn insert_system(mut commands: Commands) {
    commands.insert_resource(InsertTest);
}
fn assert_system(world: &mut World) {
    assert!(world.contains_resource::<InsertTest>())
}

fn main() {
    App::new()
        .add_systems(
            Update, 
            (insert_system, assert_system).chain_ignore_deferred()
        )
        .update();
}
```

## What went wrong

`assert_system` is an exclusive system, therefore it is expected that the command in `insert_system` is flushed before it runs. The assertion should not fail but it does.

## Additional information

I did some extensive testing and noticed that this bug occurs no matter how I do this configuration as long I make use of the `_ignore_deferred` variants. The following table lists these configurations and gives each a number because that matters for another observation I elaborate further down.

| inserting config by | asserting config by | config | test |
| - | - | - | - | 
| system | system | `after_ignore_deferred` | ❌ `1` |
| system | system | `before_ignore_deferred` | ❌ `2` |
| system | system | `chain_ignore_deferred` | ❌ `3` |
| system | set | `after_ignore_deferred` | ❌ `4` |
| system | set | `before_ignore_deferred` | ❌ `5` |
| set | system | `after_ignore_deferred` | ❌ `6` |
| set | system | `before_ignore_deferred` | ❌ `7` |
| set | set | `after_ignore_deferred` | ❌ `8` |
| set | set | `before_ignore_deferred` | ❌ `9` |
| set | set | `chain_ignore_deferred` | ❌ `10` |

For example the configuration... 
```rs
(insert_system, assert_system).chain_ignore_deferred()
```
... is the third row in the table, where "inserting config by" with "system" means that the first element in the chain tuple is the system, and "asserting config by" with "system" means that the second element in the tuple is also the system itself.

"set" means if the argument is a set, so a `configure_sets` with this configuration... 
```rs
(SetWithInsertSystem, SetWithAssertSystem).chain_ignore_deferred()
```
... is the last row.

`chain` cannot mix systems and sets, but `before` and `after` can mix, therefore these have more rows.

**General observations regarding the two system:**
- the tests do not fail if `insert_system` is exclusive
- the tests do not fail if `insert_system` is exclusive and `assert_system` is not exclusive

### Third system affecting the outcome of the tests

Now I am adding a third, empty system and order them after the two above. I observed that this results in some test going green suddenly.

```rs
fn third() {}
```

| insert-assert config by | third config by | config | ✅ fixes the failing test of ... |
| - | - | - | - | 
| each system | system | `after` | `1` ` ` `3` `4` ` ` ` ` `7` `8` `9` `10` |
| config tuple | system | `before` | `1` ` ` `3` `4` ` ` ` ` `7` `8` `9` `10`  |
| config tuple | system | `chain` | ` ` ` ` ` ` `4` ` ` ` ` `7` `8` `9` `10` |
| each system | set | `after` |`1` `2` `3` ` ` `5` `6` ` ` `8` `9` `10` |
| config tuple | set | `before` | `1` ` ` `3` `4` ` ` ` ` `7` `8` `9` `10` |
| set | system | `after` | `1` `2` `3` `4` `5` ` ` `7` `8` `9` `10` |
| set | system | `before` | `1` `2` `3` ` ` `5` `6` ` ` ` ` ` ` `  ` |
| set | set | `after`| `1` `2` `3` ` ` `5` `6` ` ` ` ` ` ` `  ` |
| set | set | `before` | `1` `2` `3` ` ` `5` `6` ` ` ` ` ` ` `  ` |
| set | set | `chain`| `1` `2` `3` ` ` `5` `6` ` ` ` ` ` ` `  ` |

For example this is the first row with `3`...
```rs
fn main() {
    App::new()
        .add_systems(
            Update, 
            (
                (insert_system, assert_system).chain_ignore_deferred(),
                third.after(insert_system).after(assert_system)
            )
        )
        .update();
}
```
... and this the second row with `3` ...
```rs
fn main() {
    App::new()
        .add_systems(
            Update, 
            (
                (insert_system, assert_system)
                    .chain_ignore_deferred().before(third),
                third
            )
        )
        .update();
}
```
... which both no longer fail.

**General observations regarding the third system:**
- it does not matter if `third` is exclusive or not
- If `third` is configured with `_ignore_deferred` variants, it does not affect the results of the first tests
- If `third` is put before the two, not after, it does not affect the result of the first tests

### Test file

I generated the tests with [this file](https://gist.github.com/urben1680/f6b323b9e70e1f9c9375123990de31b2) which also tests many other combinations that do not fail. It might come useful for testing fixes for this issue. Sorry if it looks horribly, that is how I generate tests. 🫣 Also worth noting that I double-checked that like 5 times but mistakes may only come up with a sixth time. I am just comfortable enough to list the results.

urben1680:
# Objective

Fixes #17828

Exclusive systems are expected to see the effects of all precedent commands. This was previously not made sure of by the scheduler. This showed especially with `*_ignore_deferred` configurations.

## Solution

Every edge that leads to an exclusive system will contain a sync point that are combined into one if needed.

There needs to be no extra sync point after the exclusive system because it can only queue commands in the global commands queue and that is flushed already in the [`ExclusiveFunctionSystem::run`](https://github.com/bevyengine/bevy/blob/3c9e696faa961ac01cec5933dcb4d16e653a5ecf/crates/bevy_ecs/src/system/exclusive_function_system.rs#L131) implementation.

## Testing

See the new unit tests. 
I did not add mirrored tests of those in the `no_sync_edges` and `no_sync_chain` modules below the new test because I think these are enough. Though I can add more if desired.

## Migration Guide

I am not sure if this needs any. The API did not change and I can think up no scenario where someone would rely on absent sync points before exclusive systems.

### DIFF
--- a/crates/bevy_ecs/src/schedule/schedule.rs
+++ b/crates/bevy_ecs/src/schedule/schedule.rs
@@ -2167,13 +2167,14 @@ mod tests {
         schedule.add_systems(
             (
                 |mut commands: Commands| commands.insert_resource(Resource1),
+                || (),
                 |world: &mut World| assert!(world.contains_resource::<Resource1>()),
             )
                 .chain_ignore_deferred(),
         );
         schedule.run(&mut world);
 
-        assert_eq!(schedule.executable.systems.len(), 3);
+        assert_eq!(schedule.executable.systems.len(), 4);
     }
 
     mod no_sync_edges {
--- a/crates/bevy_ecs/src/schedule/schedule.rs
+++ b/crates/bevy_ecs/src/schedule/schedule.rs
@@ -2177,6 +2177,30 @@ mod tests {
         assert_eq!(schedule.executable.systems.len(), 4);
     }
 
+    fn insert_resource(mut commands: Commands) {
+        commands.insert_resource(Resource1);
+    }
+
+    fn another_with_deferred(_: Commands) {}
+
+    fn contains_resource(world: &mut World) {
+        assert!(world.contains_resource::<Resource1>());
+    }
+
+    #[test]
+    fn merges_sync_points_before_exclusive_system() {
+        let mut schedule = Schedule::default();
+        let mut world = World::default();
+        schedule.add_systems((
+            insert_resource.before_ignore_deferred(contains_resource),
+            another_with_deferred.before_ignore_deferred(contains_resource),
+            contains_resource,
+        ));
+        schedule.run(&mut world);
+
+        assert_eq!(schedule.executable.systems.len(), 4);
+    }
+
     mod no_sync_edges {
         use super::*;
 
--- a/crates/bevy_ecs/src/schedule/schedule.rs
+++ b/crates/bevy_ecs/src/schedule/schedule.rs
@@ -2160,6 +2160,22 @@ mod tests {
         assert_eq!(schedule.executable.systems.len(), 2);
     }
 
+    #[test]
+    fn adds_sync_point_before_exclusive_system_despite_ignore_deferred() {
+        let mut schedule = Schedule::default();
+        let mut world = World::default();
+        schedule.add_systems(
+            (
+                |mut commands: Commands| commands.insert_resource(Resource1),
+                |world: &mut World| assert!(world.contains_resource::<Resource1>()),
+            )
+                .chain_ignore_deferred(),
+        );
+        schedule.run(&mut world);
+
+        assert_eq!(schedule.executable.systems.len(), 3);
+    }
+
     mod no_sync_edges {
         use super::*;
 
@@ -2215,7 +2231,6 @@ mod tests {
                     .configure_sets(Sets::A.after_ignore_deferred(insert_resource));
             });
         }
-
         #[test]
         fn set_to_system_before() {
             check_no_sync_edges(|schedule| {
--- a/crates/bevy_ecs/src/schedule/schedule.rs
+++ b/crates/bevy_ecs/src/schedule/schedule.rs
@@ -2231,6 +2231,7 @@ mod tests {
                     .configure_sets(Sets::A.after_ignore_deferred(insert_resource));
             });
         }
+
         #[test]
         fn set_to_system_before() {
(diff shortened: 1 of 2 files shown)


crates/bevy_ecs/src/schedule/schedule.rs
```diff
@@ -2215,7 +2231,6 @@ mod tests {
                     .configure_sets(Sets::A.after_ignore_deferred(insert_resource));
             });
         }
-
```
urben1680, line 2218:
```suggestion

```

urben1680:
I noticed this is incomplete. This does not work if there is a system in between that has no deferred parameter.

crates/bevy_ecs/src/schedule/schedule.rs
```diff
@@ -2160,6 +2160,47 @@ mod tests {
...
+    }
+
+    #[test]
+    fn merges_sync_points_before_exclusive_system() {
+        let mut schedule = Schedule::default();
+        let mut world = World::default();
+        schedule.add_systems((
+            insert_resource.before_ignore_deferred(contains_resource),
+            another_with_deferred.before_ignore_deferred(contains_resource),
+            contains_resource,
```
urben1680, line 2197: I guess I could make a chain of three anonymous systems but I wanted to express here that two nodes have edges with `contains_resource`, not one between the two nodes and one between the latter and the exclusive system.

urben1680:
I think this approach has too many flaws, I will make a new PR when I fixed some issues I noticed.

PR #17880, opened 2025-02-16
Bubble sync points if `ignore_deferred`, do not ignore if target system is exclusive

### LINKED ISSUE #17828: Unflushed commands before exclusive system when configured with `*_ignore_deferred`

urben1680:
# Objective

Fixes #17828

This fixes two bugs:

1. Exclusive systems should see the effect of all commands queued to that point. That does not happen when the system is configured with `*_ignore_deferred` which may lead to surprising situations. These configurations should not behave like that.
2. If `*_ignore_deferred` is used, no sync point may be added at all **after** the config. Currently this can happen if the last nodes in that config have no deferred parameters themselves. Instead, sync points should always be added after such a config, so long systems have deferred parameters.

## Solution

1. When adding sync points on edges, do not consider `AutoInsertApplyDeferredPass::no_sync_edges` if the target is an exclusive system.
2. when going through the nodes in a directed way, store the information that `AutoInsertApplyDeferredPass::no_sync_edges` suppressed adding a sync point at the target node. Then, when the target node is evaluated later by the iteration and that prior suppression was the case, the target node will behave like it has deferred parameters even if the system itself does not.

## Testing

I added a test for each bug, please let me know if more are wanted and if yes, which cases you would want to see.
These tests also can be read as examples how the current code would fail.


urben1680:
Tagging @hymm as the original author of the modified file. :wave:

alice-i-cecile (changes requested):
I'd like the tests to be more explicit about what the final inferred schedule looks like, and why that schedule is correct.

urben1680:
> I'd like the tests to be more explicit about what the final inferred schedule looks like, and why that schedule is correct.

As in doing more assertions on the schedule's internal values or adding more comments?

Btw I think there is some other problem to address. As it is now, bubbling a sync point further up means more than one sync point could be generated if there is more than one outgoing edge. But that makes no sense, the original system that wants it's commands applied needs only one sync point.

I try to think what to do here. Until then I put this one into draft again.

alice-i-cecile:
More comments mostly :) I had a hard time following what the internal logic of the tests were, and why they were correct.

urben1680:
I was unable to trigger any issues I assumed to be there so I decided to not change my approach.

crates/bevy_ecs/src/schedule/auto_insert_apply_deferred.rs
```diff
@@ -82,33 +81,45 @@ impl ScheduleBuildPass for AutoInsertApplyDeferredPass {
...
             for target in dependency_flattened.neighbors_directed(*node, Direction::Outgoing) {
-                let add_sync_on_edge = add_sync_after
-                    && !is_apply_deferred(graph.systems[target.index()].get().unwrap())
-                    && !self.no_sync_edges.contains(&(*node, target));
+                let mut add_sync_on_edge = add_sync_after;
+                let mut add_syncs_after_target = false;
+
+                if add_sync_on_edge {
+                    let target_system = graph.systems[target.index()].get().unwrap();
+                    add_sync_on_edge &= !is_apply_deferred(target_system);
```
urben1680, line 102:
```suggestion
                    add_sync_on_edge = !is_apply_deferred(target_system);
```

crates/bevy_ecs/src/schedule/auto_insert_apply_deferred.rs
```diff
@@ -23,25 +23,25 @@ use super::{
...
-            .or_else(|| {
+            .unwrap_or_else(|| {
                 let node_id = self.add_auto_sync(graph);
-                self.auto_sync_node_ids.insert(distance, node_id);
-                Some(node_id)
+                self.auto_sync_node_ids.push(node_id);
+                node_id
             })
-            .unwrap()
     }
```
urben1680, line 45: (Sorry I keep making commits to this PR, I can't help and notice things to tweak)

I noticed that `auto_sync_node_ids` can be a vector:

- The first call to this method is always with a  distance of 1. 
  - `distances_and_pending_sync` is empty at the beginning or has distances with zero values stored
  - If this method is called for the first time, then only when such a zero value was increased from zero and then stored in `distances_and_pending_sync`
- This method is always called with a distance value that is at most 1 higher than it was ever called with before.
  - All unique distances in `distances_and_pending_sync` larger than zero caused a sync point to be buffered here
  - As `target_distance` is increased by `1` at most, a missing sync point for that distance can be easily generated with a single additional push

So I decided to make this method take an `usize` index and offset it by 1 at the calling site.

crates/bevy_ecs/src/schedule/schedule.rs
```diff
@@ -2140,6 +2140,52 @@ mod tests {
...
+            |mut commands: Commands| commands.insert_resource(Resource1),
+            || {},
+        )
+            .chain_ignore_deferred();
+
+        // This chain is configured to add a sync point before the last system, but that would not happen if only the
+        // no-op system in `insert_resource_config` was evaluated as that one has no deferred parameters.
+        // So despite having no direct edge from the system that queues the command to the system that attempts to read
+        // the resource, a sync point is still added between the two tuple elements by the command system's has_deferred
+        // bubbling up to the no-op system.
```
andriyDev, line 2182: Just to be clear, previously no sync point would be generated since "system 1" is ignore_deferred against "system 2" and "system 2" has no deferred commands to apply before "system 3"? And now, "system 1" basically signals "system 2" that there are still pending commands?

crates/bevy_ecs/src/schedule/auto_insert_apply_deferred.rs
```diff
@@ -82,39 +82,53 @@ impl ScheduleBuildPass for AutoInsertApplyDeferredPass {
...
+
+                let target_system = graph.systems[target.index()].get().unwrap();
+
+                if is_apply_deferred(target_system) {
+                    // if target system is `ApplyDeferred` there is no point in adding another sync point
+                    continue;
+                }
+
+                if !target_system.is_exclusive() && self.no_sync_edges.contains(&(*node, target)) {
+                    // add sync point on a later edge
```
andriyDev, line 118: Can we make this comment a little more descriptive? e.g., "`node` has deferred commands to apply, but this edge is a no sync edge. Mark the target node as 'delaying' those commands to a future edge".

crates/bevy_ecs/src/schedule/auto_insert_apply_deferred.rs
```diff
@@ -23,25 +23,25 @@ use super::{
...
 pub struct IgnoreDeferred;
 
 impl AutoInsertApplyDeferredPass {
-    /// Returns the `NodeId` of the cached auto sync point. Will create
-    /// a new one if needed.
-    fn get_sync_point(&mut self, graph: &mut ScheduleGraph, distance: u32) -> NodeId {
+    /// Returns the `NodeId` of the cached auto sync point. Will create a new one if needed.
+    ///
+    /// Is expected to be called with an index not larger than the amount of cached sync points so far.
+    fn get_sync_point(&mut self, graph: &mut ScheduleGraph, index: usize) -> NodeId {
```
andriyDev, line 36: Should we maybe add an assert (or maybe just a debug_assert) that the `index <= auto_sync_node_ids.len()`?

crates/bevy_ecs/src/schedule/auto_insert_apply_deferred.rs
hymm: I'm not sure if this change is worth it. This isn't particularly perf sensitive code and it adds a invariant that the sync's are added in topological order, which might not always be the case.

crates/bevy_ecs/src/schedule/schedule.rs
urben1680: Correct. I take your confusion that I try to make this comment more clear.

crates/bevy_ecs/src/schedule/auto_insert_apply_deferred.rs
urben1680: That is a good point, I reverted this change.

crates/bevy_ecs/src/schedule/auto_insert_apply_deferred.rs
urben1680: I reverted this back to a HashMap

crates/bevy_ecs/src/schedule/auto_insert_apply_deferred.rs
andriyDev: I personally don't agree with this revert, it actually made it much more clear in my head that we are always only adding the next necessary sync point. 

Regardless, this could be a follow-up PR so I don't think we should block this PR on it.

urben1680:
Merged @andriyDev's changes into this PR. I must confess both our changes together make this hard to understand. I probably will try to make everything clearer, maybe with more comments, other variable names or such stuff.

If someone has good ideas to make this more clear, please write them in a comment. :smiley:

alice-i-cecile (approved):
Good test comments now! Thanks for tackling some cleanup work while you were here; I agree that this was getting quite tricky to follow 😅

PR #13569, opened 2024-05-29
Allow custom prepass without normal attribute

Ansraer:
# Objective

There are a couple of problems when trying to use CustomMaterials in cobination with a prepass:

1. Right now when you try to run a custom prepass material that doesn't have a normal attribute (e.g. because you manually pack your mesh data, or generate it on the fly) bevy refuses to use it and logs the following:
`Mesh is missing requested attribute: Vertex_Normal (MeshVertexAttributeId(1), pipeline type: Some("bevy_pbr::prepass::PrepassPipeline<bevy_voxel::voxel_engine::rendering::CustomMaterialA>"))`

This PR should also fix #13054.

2. Fixing the first problem revealed a second problem: the shader function `deferred_output` is used across the engine both for generic prepasses and deferred repasses. However, it lacked necessary checks to ensure the deferred data is only written if the prepass actually uses deferred rendering.

## Solution

1. I added an if check to make sure the normal attribute is actually used.
2. The old `deferred_output` method was replaced with `prepass_output`, which works both when deferred rendering is disabled and enabled.

## Testing

- Did you test these changes? If so, how?
Yes, I ran my experimental game project and bevy no longer complained about the missing attribute.
- Are there any parts that need more testing?
Not that I am aware of.
- How can other people (reviewers) test your changes? Is there anything specific they need to know?
Try to use a prepass material that doesn't have a normal attribute.
- If relevant, what platforms did you test these changes on, and are there any important ones you can't test?
I tested this on linux/vulkan, but I don't see why that would have any impact on this particular bug.

## Migration Guide

The shader function `pbr_deferred_functions::deferred_output` was replaced with `prepass_utils::prepass_output`. The new function should behave identical to the old one, with the added benefit of no longer crashing when running in a prepass without deferred lighting.

PR #17881, opened 2025-02-16
Allow prepass to run without ATTRIBUTE_NORMAL

### LINKED ISSUE #13054: Mesh is missing requested attribute: Vertex_Normal error
## Bevy version

0.13.1

If your bug is rendering-related, copy the adapter info that appears when you run Bevy.

```ignore
AdapterInfo { name: "NVIDIA GeForce RTX 4060", vendor: 4318, device: 10370, device_type: DiscreteGpu, driver: "NVIDIA", driver_info: "550.54.14", backend: Vulkan }
```

## What you did
I created a custom shader that uses compressed normals and position data into a `u32` . Meaning I only needed my custom attribute in my shader layout so I removed the rest of the standard layout items like `POSITION` and `NORMAL`

This is my code:
```Rust
pub const ATTRIBUTE_VOXEL_DATA: MeshVertexAttribute =
    MeshVertexAttribute::new("VoxelTypeId", 988540917, VertexFormat::Uint32x2);


#[derive(Asset, TypePath, AsBindGroup, Debug, Clone, Default)]
pub struct ChunkMaterial {
    #[texture(1, dimension = "2d_array")]
    #[sampler(2)]
    pub texture: Handle<Image>,
}

impl Material for ChunkMaterial {
    fn vertex_shader() -> ShaderRef {
        "shaders/chunk_shader_new.wgsl".into()
    }

    fn fragment_shader() -> ShaderRef {
        "shaders/chunk_shader_new.wgsl".into()
    }

    fn alpha_mode(&self) -> AlphaMode {
        AlphaMode::Opaque
    }

    fn specialize(
            _pipeline: &bevy::pbr::MaterialPipeline<Self>,
            descriptor: &mut bevy::render::render_resource::RenderPipelineDescriptor,
            layout: &bevy::render::mesh::MeshVertexBufferLayout,
            _key: bevy::pbr::MaterialPipelineKey<Self>,
        ) -> Result<(), bevy::render::render_resource::SpecializedMeshPipelineError> {

        // The shader location needs to match the one in the prepass_io
        // https://github.com/bevyengine/bevy/blob/main/crates/bevy_pbr/src/prepass/prepass_io.wgsl
        let vertex_layout = layout.get_layout(&[
            ATTRIBUTE_VOXEL_DATA.at_shader_location(30),
        ])?;

        descriptor.vertex.buffers = vec![vertex_layout];
        // descriptor.primitive.cull_mode = None;

        Ok(())
    }

    fn prepass_vertex_shader() -> ShaderRef {
        "shaders/chunk_prepass.wgsl".into()
    }

    fn prepass_fragment_shader() -> ShaderRef {
        "shaders/chunk_prepass.wgsl".into()
    }
}
```
## What went wrong
I get the following error from bevy:
`Mesh is missing requested attribute: Vertex_Normal (MeshVertexAttributeId(1), pipeline type: Some("bevy_pbr::prepass::PrepassPipeline<bevy_voxel::voxel_engine::rendering::ChunkMaterial>"))`

## Additional information

More info can be checked by reading this message and the conversation that lead to it and after:
https://discord.com/channels/691052431525675048/866787577687310356/1231657933490753666

It contains a possible explanation from Jasmine on what is the problem.

I used a workaround to add the `Normal` attribute in my layout and then supply dummy data to make it happy. 

```Rust
let vertex_layout = layout.get_layout(&[
    Mesh::ATTRIBUTE_NORMAL.at_shader_location(3),
    ATTRIBUTE_VOXEL_DATA.at_shader_location(30),
])?;
```

```Rust
let mut normals: Vec<[f32; 3]> = Vec::new();
normals.resize(mesh.vertices.len(), [0.0, 0.0, 0.0]);
mesh.insert_attribute(Mesh::ATTRIBUTE_NORMAL, normals);
```

terahlunah:
# Objective

Allow prepass to run without ATTRIBUTE_NORMAL. 
This is needed for custom materials with non-standard vertex attributes. For example a voxel material with manually packed vertex data.

Fixes #13054.

This PR covers the first part of the **stale** PR #13569 to only focus on fixing #13054.

## Solution

- Only push normals `vertex_attributes` when the layout contains `Mesh::ATTRIBUTE_NORMAL`

## Testing

  **Tested the fix on my own project with a mesh without normal attribute.**
  **I don't think so.**
  **Prepass should not be blocked on a mesh without normal attributes (with or without custom material).**
  **Probably irrelevant, but Windows/Vulkan.**

### DIFF
--- a/crates/bevy_pbr/src/prepass/mod.rs
+++ b/crates/bevy_pbr/src/prepass/mod.rs
@@ -46,7 +46,7 @@ use bevy_render::{
     Extract,
 };
 use bevy_transform::prelude::GlobalTransform;
-use tracing::error;
+use tracing::{error, warn};
 
 #[cfg(feature = "meshlet")]
 use crate::meshlet::{
@@ -508,8 +508,15 @@ where
             .mesh_key
             .intersects(MeshPipelineKey::NORMAL_PREPASS | MeshPipelineKey::DEFERRED_PREPASS)
         {
-            vertex_attributes.push(Mesh::ATTRIBUTE_NORMAL.at_shader_location(3));
             shader_defs.push("NORMAL_PREPASS_OR_DEFERRED_PREPASS".into());
+            if layout.0.contains(Mesh::ATTRIBUTE_NORMAL) {
+                shader_defs.push("VERTEX_NORMALS".into());
+                vertex_attributes.push(Mesh::ATTRIBUTE_NORMAL.at_shader_location(3));
+            } else if key.mesh_key.contains(MeshPipelineKey::NORMAL_PREPASS) {
+                warn!(
+                    "The default normal prepass expects the mesh to have vertex normal attributes."
+                );
+            }
             if layout.0.contains(Mesh::ATTRIBUTE_TANGENT) {
                 shader_defs.push("VERTEX_TANGENTS".into());
                 vertex_attributes.push(Mesh::ATTRIBUTE_TANGENT.at_shader_location(4));
--- a/crates/bevy_pbr/src/prepass/prepass.wgsl
+++ b/crates/bevy_pbr/src/prepass/prepass.wgsl
@@ -96,6 +96,7 @@ fn vertex(vertex_no_morph: Vertex) -> VertexOutput {
 #endif // VERTEX_UVS_B
 
 #ifdef NORMAL_PREPASS_OR_DEFERRED_PREPASS
+#ifdef VERTEX_NORMALS
 #ifdef SKINNED
     out.world_normal = skinning::skin_normals(world_from_local, vertex.normal);
 #else // SKINNED
@@ -106,6 +107,7 @@ fn vertex(vertex_no_morph: Vertex) -> VertexOutput {
         vertex_no_morph.instance_index
     );
 #endif // SKINNED
+#endif // VERTEX_NORMALS
 
 #ifdef VERTEX_TANGENTS
     out.world_tangent = mesh_functions::mesh_tangent_local_to_world(
--- a/crates/bevy_pbr/src/prepass/prepass_io.wgsl
+++ b/crates/bevy_pbr/src/prepass/prepass_io.wgsl
@@ -15,7 +15,9 @@ struct Vertex {
 #endif
 
 #ifdef NORMAL_PREPASS_OR_DEFERRED_PREPASS
+#ifdef VERTEX_NORMALS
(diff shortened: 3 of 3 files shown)


alice-i-cecile (approved):

crates/bevy_pbr/src/prepass/mod.rs
```diff
@@ -508,8 +508,11 @@ where
             .mesh_key
             .intersects(MeshPipelineKey::NORMAL_PREPASS | MeshPipelineKey::DEFERRED_PREPASS)
         {
-            vertex_attributes.push(Mesh::ATTRIBUTE_NORMAL.at_shader_location(3));
             shader_defs.push("NORMAL_PREPASS_OR_DEFERRED_PREPASS".into());
+            if layout.0.contains(Mesh::ATTRIBUTE_NORMAL) {
```
IceSentry, line 512: I feel like I'd want to see a warning if the NORMAL_PREPASS is enabled without normals

crates/bevy_pbr/src/prepass/prepass.wgsl
```diff
@@ -96,11 +96,17 @@ fn vertex(vertex_no_morph: Vertex) -> VertexOutput {
...
+    var vertex_normal = vec3(0.0);
+#endif // VERTEX_NORMALS
+
 #ifdef SKINNED
-    out.world_normal = skinning::skin_normals(world_from_local, vertex.normal);
+    out.world_normal = skinning::skin_normals(world_from_local, vertex_normal);
 #else // SKINNED
     out.world_normal = mesh_functions::mesh_normal_local_to_world(
-        vertex.normal,
+        vertex_normal,
```
IceSentry, line 109: I don't really think it makes sense to pass it a vec3 with zeroes. If the mesh has no normals it shouldn't try to do anything with them

crates/bevy_pbr/src/prepass/mod.rs
terahlunah: I added a warning if the NORMAL_PREPASS is enabled without normals

crates/bevy_pbr/src/prepass/prepass.wgsl
terahlunah: Yeah that's fair, I've surrounded the whole block with `#ifdef VERTEX_NORMALS` instead of calling it with a default value.

IceSentry (approved):

PR #17889, opened 2025-02-17
Add `http` & `https` asset sources (clean commit history)

### LINKED ISSUE #16307: Feature: Upstream `bevy_web_asset`, allowing assets loaded via `http`

mrchantey:
# Objective

This is a duplicate of #16366 because I made a mess of the commit history. See that pr for more context and discussion.

- Fixes #16307 
- Closes #16366

## Solution

- Add `http` & `https` asset sources
- Allows `CDLA-Permissive-2.0` in `deny.toml`

## Testing

- Verify visually by running `cargo run --example http_source --features="http_source"`

---

## Showcase

Bevy now supports assets loaded via url!

Add the `http_source` feature to load assets served over the web, with support for both native and wasm targets. On native targets the `http_source_cache` feature will use a basic caching mechanism to avoid repeated network requests.

```rust
  // Simply use a url where you would normally use an `assets` path
  let handle = asset_server.load("https://raw.githubusercontent.com/bevyengine/bevy/refs/heads/main/assets/branding/bevy_bird_dark.png");
  commands.spawn(Sprite::from_image(handle));
```

## Further Work

- Better AssetReaderError: The asset path is not always collected in the error. This is likely a common error that actual non-developer users will see so its important to get it right.
- Possibly expose the native request mechanic as a general `fetch`, in the future we'll have many reasons to be making requests.


Elabajaba:
Just a quick note, [ring is now unmaintained for the foreseeable future.](https://github.com/briansmith/ring/discussions/2414)

recatek:
Adding MPL seems like it would have complications for distributing game binaries, no?

mockersf:
> Just a quick note, [ring is now unmaintained for the foreseeable future.](https://github.com/briansmith/ring/discussions/2414)

ring is maintained again, now by the rustls team

.gitignore
```diff
@@ -1,6 +1,6 @@
 # If your IDE needs additional project specific files, configure git to ignore them:
 # https://docs.github.com/en/get-started/getting-started-with-git/ignoring-files#configuring-ignored-files-for-all-repositories-on-your-computer
-
+.vscode
```
mockersf, line 3:
```suggestion
```

deny.toml
```diff
@@ -18,7 +18,9 @@ allow = [
   "ISC",
   "MIT",
   "MIT-0",
+  "MPL-2.0",
```
mockersf, line 21: I don't think the MPL license can be allowed

https://www.mozilla.org/en-US/MPL/2.0/
> 3.2. Distribution of Executable Form
requires to distribute the source
> 3.3. Distribution of a Larger Work
even if it's just part of a larger work

mockersf:
blocked on finding where the MPL license comes from and removing it

mockersf:
waiting on a new release of webpki-root with this commit: https://github.com/rustls/webpki-roots/commit/90c48f38672f4300eea5f1b712d4ccca8ad82123#diff-8aec0a2c31031e00524df726f854726660ec8289dcddfa70005000fc21b00511

jf908:
As of right now, ring is not actually included in this PR, it was replaced by aws-lc-rs. Now that ring has [removed the OpenSSL](https://github.com/briansmith/ring/commit/0d80ad56a24bdcc3daeed338551379d8adbed976) parts of the license, I think we can remove aws-lc-rs, remove OpenSSL from deny.toml and use the ureq's default ring dependency again. This would be especially nice because aws-lc-rs requires installing CMake on Windows.

mrchantey:
Okeydokey back to `ring` with OpenSSL and MPL licences removed from the allow list, the next release of `webpki-roots` should get check-licenses passing in ci 🤞

jf908:
I poked the [rustls discord](https://discord.com/channels/976380008299917365/1015156984007381033/1365802386668650597) and they pushed a new webpki-roots release! Just have to wait for ureq to update too now.

Although the rustls discord also suggested that we make use of rustls-platform-verifier rather than webpki-roots. rustls-platform-verifier uses the OS's certificates which is the default behaviour of most applications since it won't be on the onus of bevy and bevy games to keep certificates regularly up to date. See [here](https://github.com/rustls/rustls-platform-verifier/tree/main/rustls-platform-verifier#deployment-considerations). ureq uses webpki-roots as the default to reduce the default number of dependencies but I feel like that might be the wrong default for a bevy application. With that said, maybe we just worry about that in a follow up PR, this one has already been through quite the cycle of development :)

alice-i-cecile:
I'm fine to call this unblocked once `ureq` releases :) I agree with your analysis: that's a better default, but can be done in follow-up.

Very eager to get this into the engine to help unblock the usage of better assets in our examples (in addition to the direct value to users).

examples/asset/http_source.rs
```diff
@@ -0,0 +1,22 @@
...
+        .add_systems(Startup, setup)
+        .run();
+}
+
+fn setup(mut commands: Commands, asset_server: Res<AssetServer>) {
+    commands.spawn(Camera2d);
+
+    commands.spawn(
+        // Simply use a url where you would normally use an asset folder relative path
+        Sprite::from_image(asset_server.load("https://raw.githubusercontent.com/bevyengine/bevy/refs/heads/main/assets/branding/bevy_bird_dark.png"))
```
rlidwka, line 20: I suggest to reformat it like this:

```rs
commands.spawn(Sprite::from_image(asset_server.load(
    // Simply use a url where you would normally use an asset folder relative path
    "https://raw.githubusercontent.com/bevyengine/bevy/refs/heads/main/assets/branding/bevy_bird_dark.png"
)));
```

As is, the line is too long to fit even in github UI.

docs/cargo_features.md
```diff
@@ -86,6 +86,8 @@ The default feature set enables most of the expected features of a game engine,
 |ghost_nodes|Experimental support for nodes that are ignored for UI layouting|
 |gif|GIF image format support|
 |glam_assert|Enable assertions to check the validity of parameters passed to glam|
+|http_source|Enables using assets from HTTP sources|
+|http_source_cache|Assets downloaded from HTTP sources are cached|
```
rlidwka, line 90: why is `http_source_cache` a feature instead of AssetLoader argument?

caching doesn't pull in any dependencies as far as I see, so benefits for conditional compilation are minimal

crates/bevy_asset/src/lib.rs
```diff
@@ -352,6 +355,9 @@ impl AssetPlugin {
 
 impl Plugin for AssetPlugin {
     fn build(&self, app: &mut App) {
+        #[cfg(feature = "http_source")]
+        app.add_plugins(http_source::http_source_plugin);
```
rlidwka, line 359: Question: do you want this to be enabled by default? Or perhaps force users to add it explicitly if people need it?

Pulling stuff from the internet is a potential security risk, e.g. if your GLTF loader has code execution vulnerability, this asset loader just upgrades it to remote code execution.

examples/asset/http_source.rs
```diff
@@ -0,0 +1,22 @@
+//! Example usage of the `http` asset source to load assets from the web.
+//!
+//! Run with the feature `http_source`, and optionally `http_source_cache`
+//! for a simple caching mechanism that never invalidates.
+//!
+use bevy::prelude::*;
+
+fn main() {
+    App::new()
+        .add_plugins(DefaultPlugins)
```
rlidwka, line 10:
```rs
App::new()
        .add_plugins(DefaultPlugins)
        .add_plugins(HttpAssetLoader::with_cache_dir(".http-cache"))
        .add_systems(Startup, setup)
        .run();
```

crates/bevy_asset/src/http_source.rs
```diff
@@ -0,0 +1,254 @@
...
+
+    const CACHE_DIR: &str = ".http-asset-cache";
+
+    fn url_to_hash(url: &str) -> String {
+        let mut hasher = DefaultHasher::new();
+        url.hash(&mut hasher);
+        std::format!("{:x}", hasher.finish())
+    }
+
+    pub fn try_load_from_cache(url: &str) -> Result<Option<Vec<u8>>, io::Error> {
```
rlidwka, line 170: can it use `async-fs` like the rest of bevy assets?

crates/bevy_asset/src/http_source.rs
```diff
@@ -0,0 +1,254 @@
...
+mod http_asset_cache {
+    use alloc::string::String;
+    use alloc::vec::Vec;
+    use core::hash::{Hash, Hasher};
+    use std::collections::hash_map::DefaultHasher;
+    use std::fs::{self, File};
+    use std::io::{self, Read, Write};
+    use std::path::PathBuf;
+
+    const CACHE_DIR: &str = ".http-asset-cache";
```
rlidwka, line 162: Is there any point in moving it inside `./assets`?

Three things are worrying me here:
1. bevy would create it in current directory; if you run app from other `cwd`, it'll create a different directory
2. there's already `assets` folder managed by bevy, me no likey too many directories
3. no ability to override (e.g. some people would want it to be in `/home/user/.cache`, some in `./`, some weirdos prefer `/tmp` for anything important)

crates/bevy_asset/src/http_source.rs
```diff
@@ -0,0 +1,254 @@
...
+            Self::Http => "http://",
+            Self::Https => "https://",
+        })
+        .join(path)
+    }
+
+    /// See [`crate::io::get_meta_path`]
+    fn make_meta_uri(&self, path: &Path) -> Option<PathBuf> {
+        let mut uri = self.make_uri(path);
+        let mut extension = path.extension()?.to_os_string();
```
rlidwka, line 43: How and why can this fail?

If you have a file called `http://example.com/1234567890`, surely `http://example.com/1234567890.meta` is a reasonable path to look metadata in?

crates/bevy_asset/src/http_source.rs
```diff
@@ -0,0 +1,254 @@
...
+        }
+        // ureq considers all >=400 status codes as errors
+        Err(ureq::Error::StatusCode(code)) => {
+            if code == 404 {
+                Err(AssetReaderError::NotFound(path))
+            } else {
+                Err(AssetReaderError::HttpError(code))
+            }
+        }
+        Err(err) => Err(AssetReaderError::Io(
```
rlidwka, line 106: It will be useful to mention to users (in the example perhaps) how to get this error out of bevy and display it to a user.

Once you introduce remote assets, errors will become quite common, so need to teach proper error handling to people.

crates/bevy_asset/src/http_source.rs
```diff
@@ -0,0 +1,254 @@
...
+        )
+    })?;
+
+    #[cfg(feature = "http_source_cache")]
+    if let Some(data) = http_asset_cache::try_load_from_cache(str_path)? {
+        return Ok(Box::new(VecReader::new(data)));
+    }
+    use ureq::Agent;
+
+    static AGENT: LazyLock<Agent> = LazyLock::new(|| Agent::config_builder().build().new_agent());
```
rlidwka, line 83: Need to set `Bevy/0.16` as User-Agent with possibility to configure by the user.

In general, people may want to configure their http client, e.g. amount of parallel requests, redirects, retries, and all the timeouts.

crates/bevy_asset/src/http_source.rs
```diff
@@ -0,0 +1,254 @@
...
+
+    match AGENT.get(str_path).call() {
```
rlidwka, line 85: I'd suggest to add `bevy::log::info` on every http request for debugging purposes (unless `ureq` does it already).

rlidwka (commented):
Would it be too much to ask for current download progress statistics?

*edit*: and a user-defined validation function to check digital signature or checksum

crates/bevy_asset/src/http_source.rs
```diff
@@ -0,0 +1,254 @@
...
+    use crate::io::wasm::HttpWasmAssetReader;
+
+    HttpWasmAssetReader::new("")
+        .fetch_bytes(path)
+        .await
+        .map(|r| Box::new(r) as Box<dyn Reader>)
+}
+
+#[cfg(not(target_arch = "wasm32"))]
+async fn get(path: PathBuf) -> Result<Box<dyn Reader>, AssetReaderError> {
```
rlidwka, line 61: need to pass client in from outside instead of having it as a static variable

perhaps, user can pass it as an argument (see above regarding user-agent override)

crates/bevy_asset/src/lib.rs
jf908: I’m in favour of having this disabled by default because the dependencies needed to do https requests are reasonably large.

crates/bevy_asset/src/http_source.rs
jf908: I’m not a fan of using the assets folder. I think something being in the asset folder implies it’s a source asset because some plugins like asset preprocessing and embedded assets operate blankly on everything in the assets folder so it would be easy to accidentally ship cached assets.

I’m also not a particular fan of the .http_asset_cache but I can’t think of any good alternatives

jf908 (approved):
I just realized this PR is no longer blocked - I was wrong about needing to wait on a ureq release since patch dependencies can be upgraded automatically.

Approved from me other than .gitignore change and adding `CDLA-Permissive-2.0` to allowed licenses in deny.toml. I guess a maintainer will have to give their ok to the license in final review.

jf908:
@rlidwka has raised some very good review points but personally I feel this PR is in a good enough state to merge and then we can start making incremental follow ups - feel free to disagree with me if you think there is anything worth blocking

docs/cargo_features.md
mrchantey: the cache is a naive placeholder for a crate like [http-cache](https://crates.io/crates/http-cache) which [doesnt yet support ureq](https://github.com/06chaynes/http-cache/issues/91)

mrchantey:
I agree, more iteration is requried but this pr is already substantial. How about we create a follow-up issue for discussion on whats needed to complete `http asset sources` and open smaller prs from there.

crates/bevy_asset/src/lib.rs
alice-i-cecile: Definitely disabled by default.

crates/bevy_asset/src/http_source.rs
mrchantey: our time may be better spend implementing `http-cache` for ureq than polishing this hand-rolled solution.

crates/bevy_asset/src/http_source.rs
mrchantey: asset errors in general need a rework. i consider it out of scope for this initial implementation.

crates/bevy_asset/src/http_source.rs
mrchantey: see further work: general fetch

crates/bevy_asset/src/http_source.rs
mrchantey: see further work: general fetch

examples/asset/http_source.rs
mrchantey: our time may be better spend implementing http-cache for ureq than polishing this hand-rolled solution.

crates/bevy_asset/src/http_source.rs
mrchantey: see further work: general fetch

alice-i-cecile:
I've spoken with @cart about this, and we'd like to move forward with this as a solution for example assets (in addition to thinking it's broadly useful). We just need a second approval now to get this merged.

crates/bevy_asset/Cargo.toml
```diff
@@ -90,6 +92,11 @@ bevy_reflect = { path = "../bevy_reflect", version = "0.17.0-dev", default-featu
 
 [target.'cfg(not(target_arch = "wasm32"))'.dependencies]
 notify-debouncer-full = { version = "0.5.0", default-features = false, optional = true }
+# updating ureq: while ureq is semver stable, it depends on rustls which is not, meaning unlikely but possible breaking changes on minor releases. https://github.com/bevyengine/bevy/pull/16366#issuecomment-2572890794
+ureq = { version = "3", optional = true, default-features = false, features = [
+  "gzip",
+  "json",
```
mockersf: those features shouldn't be enabled by default
we should either re-expose features for them, or document how to enable them on `ureq` when needed

crates/bevy_asset/src/http_source.rs
```diff
@@ -0,0 +1,246 @@
...
+use bevy_tasks::ConditionalSendFuture;
+use std::path::{Path, PathBuf};
+
+/// Adds the `http` and `https` asset sources to the app.
+/// Any asset path that begins with `http` or `https` will be loaded from the web
+/// via `fetch`(wasm) or `ureq`(native).
+pub fn http_source_plugin(app: &mut App) {
+    app.register_asset_source(
+        "http",
+        AssetSource::build().with_reader(|| Box::new(HttpSourceAssetReader::Http)),
```
mockersf: I would prefer if `http` support was behind a non default feature flag, to force people who want to use something non secure to enable it manually

mockersf (changes requested):
similar to how the file asset source exposes only a directory and not the whole filesystem, I think an https source should only allow access to a domain and not the whole internet

crates/bevy_asset/src/http_source.rs
jf908: This module is actually already gated here by a non default feature https://github.com/bevyengine/bevy/blob/de3e2b62dd1502d6390e2d249a55ae44e4cd9e72/crates/bevy_asset/src/lib.rs#L189

mockersf: I would prefer an additional feature just for `http`, or different features for `https` and `http`

docs/cargo_features.md
```diff
@@ -91,6 +91,8 @@ The default feature set enables most of the expected features of a game engine,
 |gif|GIF image format support|
 |glam_assert|Enable assertions to check the validity of parameters passed to glam|
 |gltf_convert_coordinates_default|Enable converting glTF coordinates to Bevy's coordinate system by default. This will be Bevy's default behavior starting in 0.18.|
+|http_source|Enables using assets from HTTP sources|
+|http_source_cache|Assets downloaded from HTTP sources are cached|
 |hotpatching|Enable hotpatching of Bevy systems|
```
mockersf:
```suggestion
|hotpatching|Enable hotpatching of Bevy systems|
|http_source|Enables using assets from HTTP sources|
|http_source_cache|Assets downloaded from HTTP sources are cached|
```

jf908:
> similar to how the file asset source exposes only a directory and not the whole filesystem, I think an https source should only allow access to a domain and not the whole internet

I am quite biased towards allowing more than 1 origin because I’m currently making a bevy game where not all my assets are hosted over the same server. Otherwise I see the upside of being able to neatly switch out the asset server host name but there’s also the downside that we would lose the neatness of the asset path being the URL.

Do you think a config option to specify allowed host names would be a good alternative?

jf908:
I have given this some more thought and I'm starting to turn around to the idea.

I don't think a config option for allowed host names would be that useful since its quite easy to parse the url and do any checking you want manually before you load an asset.

I didn't know this until now but file assets can be loaded from absolute paths but they will be rejected unless unapproved_path_mode is set accordingly. We could probably do the same for web assets by having the same unapproved_path_mode option to allow access to the entire internet if required. The main problem I see is that relative paths already default to the file system and afaik there is no common syntax to define a relative http URL so we would have to come up with our own syntax which doesn't clash with absolute http URLs which I don't think would be very intuitive.

Another issue is that unlike the assets folder, there is no sensible default origin. If we want to force people to pick an origin (which would be good practice), we wouldn't be able to make the Plugin Default and that would make it annoying to construct (in the future when it has many config options).

mockersf:
> If we want to force people to pick an origin (which would be good practice), we wouldn't be able to make the Plugin Default and that would make it annoying to construct (in the future when it has many config options).

I think this is more a feature than an issue. Having to opt in to http on a specific domain is a better way to use it in my opinion. If the number of config options becomes too large, we can either rely on a default implementation for most options, or a builder pattern

crates/bevy_asset/src/http_source.rs
```diff
@@ -0,0 +1,264 @@
...
+        path: &'a Path,
+    ) -> Result<Box<PathStream>, AssetReaderError> {
+        Err(AssetReaderError::NotFound(self.make_uri(path)))
+    }
+}
+
+/// A naive implementation of an HTTP asset cache that never invalidates.
+/// `ureq` currently does not support caching, so this is a simple workaround.
+/// It should eventually be replaced by `http-cache` or similar, see [tracking issue](https://github.com/06chaynes/http-cache/issues/91)
+#[cfg(feature = "http_source_cache")]
```
jiangheng90: this may not be effect in wasm32. so I think may be use #[cfg(all(not(target_arch = "wasm32"), feature = "http_source_cache"))] is proper

crates/bevy_asset/src/http_source.rs
```diff
@@ -0,0 +1,264 @@
...
+            HttpSourceAssetReader::Https
+                .make_meta_uri(Path::new("example.com/favicon.png"))
+                .to_str()
+                .unwrap(),
+            "https://example.com/favicon.png.meta"
+        );
+    }
+
+    #[test]
+    fn make_https_without_extension_meta_uri() {
```
jiangheng90, line 268: this test can not pass on macOS, it will  be "https://example.com/favicon..meta"

mrchantey: are we missing something in ci?

jiangheng90: > are we missing something in ci?

I think there is something wrong with your handle with "example.com/favicon" this case

your original extension will be "" and after u push ".meta", it will be "example.com/favicon..meta"

compare with  "example.com/favicon.png", the origin extension will be "png", after pushing ".meta", it becomes "png.meta".

so maybe you can think about this
```rust
fn make_meta_uri(&self, path: &Path) -> PathBuf {
    let mut uri = self.make_uri(path);
    let mut extension = path.extension().unwrap_or_default().to_os_string();
    if !extension.is_empty() {
        extension.push(".");
    }
    extension.push("meta");
    uri.set_extension(extension);
    uri
}
```

mrchantey: i mean the tests passed in macOS ci, how can i replicate the failed test?

mrchantey:
> file assets can be loaded from absolute paths but they will be rejected unless unapproved_path_mode is set accordingly. We could probably do the same for web assets by having the same unapproved_path_mode option to allow access to the entire internet if required.

Sounds like a plan, looking at the mechanism im a bit confused, it looks like this check is done on the `AssetServer::load..` level, not the individual loader level? How do we opt into the same checks?

crates/bevy_asset/src/http_source.rs
jiangheng90: > i mean the tests passed in macOS ci, how can i replicate the failed test?

I think it will not pass on all platform. here is the doc on Path extension

```rust
use std::path::Path;

assert_eq!("rs", Path::new("foo.rs").extension().unwrap());
assert_eq!("gz", Path::new("foo.tar.gz").extension().unwrap());
```

in doc. the extension will get with out the "." when something like "example.com/favicon.png", it will get "png", 
and "example.com/favicon" will get "".

so I think you can run this test only and it will get failure in any platform.

mrchantey: ah yep good catch its not running in ci due to feature flags, this may be a broader issue than `http sources`, the impl is verbatim from [`get_meta_path`](https://github.com/mrchantey/bevy/blob/7e4a750497af18b91b83697639e618eef2d70676/crates/bevy_asset/src/io/mod.rs#L767). They should probably share this part of the code.

mrchantey: Ah yep good catch its not running in ci due to feature flags. Its a verbatim copy of [`get_meta_path`](https://github.com/mrchantey/bevy/blob/7e4a750497af18b91b83697639e618eef2d70676/crates/bevy_asset/src/io/mod.rs#L767) which has the same bug, ill update and reuse that one.

Cargo.toml
```diff
@@ -516,6 +516,15 @@ file_watcher = ["bevy_internal/file_watcher"]
 # Enables watching in memory asset providers for Bevy Asset hot-reloading
 embedded_watcher = ["bevy_internal/embedded_watcher"]
 
+# Enables downloading assets from HTTP sources
+http = ["bevy_internal/http"]
+
+# Enables downloading assets from HTTPS sources
+https = ["bevy_internal/https"]
+
+# Assets downloaded from HTTP sources are cached
```
atlv24: Are https assets not cachable? (may have to rename the feature if not)
```suggestion
# Enable caching downloaded assets on the filesystem. NOTE: this cache currently never invalidates entries!
```

crates/bevy_asset/src/http_source.rs
```diff
@@ -0,0 +1,261 @@
...
+    if let Some(data) = http_asset_cache::try_load_from_cache(str_path).await? {
+        return Ok(Box::new(VecReader::new(data)));
+    }
+    use ureq::Agent;
+
+    static AGENT: LazyLock<Agent> = LazyLock::new(|| Agent::config_builder().build().new_agent());
+
+    match AGENT.get(str_path).call() {
+        Ok(mut response) => {
+            // let mut reader = response.into_reader();
```
atlv24:
```suggestion
```

crates/bevy_internal/Cargo.toml
```diff
@@ -218,6 +218,15 @@ debug_glam_assert = ["bevy_math/debug_glam_assert"]
 
 default_font = ["bevy_text?/default_font"]
 
+# Enables downloading assets from HTTP sources
+http = ["bevy_asset?/http"]
+
+# Enables downloading assets from HTTPS sources
+https = ["bevy_asset?/https"]
+
+# Assets downloaded from HTTP sources are cached
```
atlv24:
```suggestion
# Enable caching downloaded assets on the filesystem. NOTE: this cache currently never invalidates entries!
```

docs/cargo_features.md
```diff
@@ -92,6 +92,9 @@ The default feature set enables most of the expected features of a game engine,
 |glam_assert|Enable assertions to check the validity of parameters passed to glam|
 |gltf_convert_coordinates_default|Enable converting glTF coordinates to Bevy's coordinate system by default. This will be Bevy's default behavior starting in 0.18.|
 |hotpatching|Enable hotpatching of Bevy systems|
+|http|Enables downloading assets from HTTP sources|
+|http_source_cache|Assets downloaded from HTTP sources are cached|
```
atlv24:
```suggestion
|http_source_cache|Enable caching downloaded assets on the filesystem. NOTE: this cache currently never invalidates entries!|
```

crates/bevy_asset/src/http_source.rs
```diff
@@ -0,0 +1,261 @@
+use crate::io::{AssetReader, AssetReaderError, Reader};
+use crate::io::{AssetSource, PathStream};
+use crate::AssetApp;
+use alloc::boxed::Box;
+use bevy_app::{App, Plugin};
+use bevy_tasks::ConditionalSendFuture;
+use std::path::{Path, PathBuf};
+
+/// Adds the `http` and `https` asset sources to the app.
+///
```
atlv24, line 11:
```suggestion
///
/// NOTE: Make sure to add this plugin *before* `AssetPlugin` to properly register http asset sources.
///
```

crates/bevy_internal/src/default_plugins.rs
```diff
@@ -21,6 +21,10 @@ plugin_group! {
         #[cfg(feature = "std")]
         #[custom(cfg(any(all(unix, not(target_os = "horizon")), windows)))]
         bevy_app:::TerminalCtrlCHandlerPlugin,
+        // NOTE: Load this before AssetPlugin to properly register http asset sources.
```
atlv24, line 24: This is an interesting case of plugin dependency ordering: usually, i would expect "A depends on B" to mean A must load *after* B, but here it is inverted, and we definitely dont want AssetPlugin to depend on HttpSourcePlugin. Something to have in mind for eventual plugin dependencies i guess.

crates/bevy_asset/src/io/mod.rs
```diff
@@ -762,11 +763,16 @@ impl Reader for SliceReader<'_> {
...
+/// - `foo` becomes `foo.meta`
+/// - `foo.bar` becomes `foo.bar.meta`
 pub(crate) fn get_meta_path(path: &Path) -> PathBuf {
     let mut meta_path = path.to_path_buf();
     let mut extension = path.extension().unwrap_or_default().to_os_string();
-    extension.push(".meta");
+    if !extension.is_empty() {
+        extension.push(".");
+    }
+    extension.push("meta");
```
atlv24, line 775: Nice bug fix, im glad that you added tests to cover this case too!

crates/bevy_asset/src/io/wasm.rs
```diff
@@ -52,7 +52,8 @@ fn js_value_to_err(context: &str) -> impl FnOnce(JsValue) -> std::io::Error + '_
 }
 
 impl HttpWasmAssetReader {
-    async fn fetch_bytes(&self, path: PathBuf) -> Result<impl Reader, AssetReaderError> {
+    // Also used by HttpSourceAssetReader
```
atlv24:
```suggestion
    // Also used by [`HttpSourceAssetReader`](crate::HttpSourceAssetReader)
```

atlv24 (commented):
Nice work! It would be cool to get this in for 0.17, imo.

I quickly realized that `http` is kinda used as a synonym for `web` in this PR. I think it could use some terminology cleanup, its a bit unclear that https is subject to all the same implementation. At least the the user facing level, the cache feature and plugin should probably use `web` instead of `http`.

jf908:
> I quickly realized that http is kinda used as a synonym for web in this PR. I think it could use some terminology cleanup, its a bit unclear that https is subject to all the same implementation. At least the the user facing level, the cache feature and plugin should probably use web instead of http.

My only worry about this is that web is also synonymous with browser/wasm. Some bevy crates have a feature called "web" that indicates that web apis are available. I think if its clear enough that its specifically about assets from the web it'll be ok.

crates/bevy_asset/src/http_source.rs
```diff
@@ -0,0 +1,262 @@
...
+        app.register_asset_source(
+            "https",
+            AssetSource::build()
+                .with_reader(|| Box::new(HttpSourceAssetReader::Https))
+                .with_processed_reader(|| Box::new(HttpSourceAssetReader::Https)),
+        );
+    }
+}
+
+impl Default for HttpSourcePlugin {
```
andriyDev, line 47: Just derive this?

deny.toml
```diff
@@ -23,6 +23,7 @@ allow = [
   "BSD-3-Clause",
   "BSL-1.0",
   "CC0-1.0",
+  "CDLA-Permissive-2.0",
```
andriyDev, line 26: What's going on with this deny? Why add it here?

crates/bevy_asset/src/http_source.rs
andriyDev: I agree with @jf908. Asset processing already processes assets into `imported_assets` folder separate from `assets` folder.

I would like a followup to make this configurable if possible but I won't block this PR for that.

crates/bevy_asset/src/http_source.rs
```diff
@@ -0,0 +1,262 @@
...
+    fn default() -> Self {
+        Self
+    }
+}
+
+/// Asset reader that treats paths as urls to load assets from.
+pub enum HttpSourceAssetReader {
```
andriyDev, line 54: Can we add a doc comment here saying that this source reader does not support directories? I don't think that's a limitation we can fix, but it's one I want documented.

andriyDev (changes requested):
The main thing that bothers me here is that ureq calls are blocking. That means that loading any asset through this loader in single threaded environments will entirely freeze your app until the asset loads. That's worrisome. Based on the README for `ureq` it sounds like they are *proud* to not be async, so I'm doubtful we can push on them to make an async API or something.

crates/bevy_asset/src/http_source.rs
jf908: I added this with the expectation we will need an option to for unapproved http domains (discussed above). If that does not come to fruition through this default please derive.

deny.toml
jf908: This is the license for webpki-roots which is required by ureq when rustls is enabled for the https feature

mrchantey:
> The main thing that bothers me here is that ureq calls are blocking. 

I think the reasoning in the ecs discord chat was because its lightweight but yes it is limited.

Maybe we can have some kind of `FetchProvider` trait? I can see the appeal in using `ureq` in development (faster build times) and `reqwest` in production (async & concurrent requests).

mockersf:
If I remember correctly, `reqwest` needs a tokio runtime to be able to be asynchronous.

The main environment where single threadness and http requests should happen is wasm, and it should be handled through the browser which will manage non blocking for us

andriyDev:
@mockersf what do you mean by "should be handled through the browser"? As in use like wasm-bindgen to talk to the `fetch` API or something? That's quite a lot of pain and not something I'd expect a user to go through or know about.

I think I won't fully block on this (pun intended), just because the http request will eventually terminate, so we will never actually get deadlocked on this or anything, but it is sad that a user's game can stall out in singlethreaded mode for just downloading assets...

jf908:
> what do you mean by "should be handled through the browser"? 

This PR just reuses bevy's current default asset loader on wasm which is already async and uses wasm-bindgen to talk the fetch API. The user doesn't have to do anything wasm-specific to enable this. Notably the new HttpAssetPlugin plugin is still useful on wasm because it lets you access URLs outside of the current one.

I think it will be quite rare that you need to run bevy single-threaded and also need web assets but It's still a shame its not async. Looking through the rust ecosystem I don't feel like there's any alternatives that slot in as nicely as ureq but hopefully we can address that in the future.

alice-i-cecile:
@mockersf I would like to ship this and start using it internally. It's had a long and troubled history, but fundamentally we have consensus that we want this functionality. What needs to be done before you're comfortable merging this?

NthTensor:
I'm planning to do a full review of this, but before I really get into it I noticed that we've rolled out own wrappers around `ureq` and browser fetch. Would instead using [`ehttp`](<https://docs.rs/ehttp/latest/ehttp/>) (which does more or less the same thing) simplify things?

crates/bevy_asset/src/http_source.rs
```diff
@@ -0,0 +1,271 @@
...
+
+    #[cfg(all(not(target_arch = "wasm32"), feature = "http_source_cache"))]
+    if let Some(data) = http_asset_cache::try_load_from_cache(str_path).await? {
+        return Ok(Box::new(VecReader::new(data)));
+    }
+    use ureq::Agent;
+
+    static AGENT: LazyLock<Agent> = LazyLock::new(|| Agent::config_builder().build().new_agent());
+
+    match AGENT.get(str_path).call() {
```
NthTensor: Using `call` blocks the async context. This will clog the executor, and prevent other tasks from executing. Hard blocker for me, IMO this cannot merge as is.

Every call to `ureq` needs to be done on a new thread so as not to block the async runtime. The `ehttp` crate does this automatically. We could also do this manually using `bevy_tasks` or `unblock`.

NthTensor (changes requested):
Generally looks pretty good, thanks for your work.

I rarely submit blocking feedback, but unfortunately I don't think the way this mixes blocking and async calls is correct.

mockersf:
> @mockersf I would like to ship this and start using it internally. It's had a long and troubled history, but fundamentally we have consensus that we want this functionality. What needs to be done before you're comfortable merging this?

It needs to limit access to user specified domains

mrchantey:
> > @mockersf I would like to ship this and start using it internally. It's had a long and troubled history, but fundamentally we have consensus that we want this functionality. What needs to be done before you're comfortable merging this?
> 
> It needs to limit access to user specified domains

fyi vero is [looking into this](https://discord.com/channels/691052431525675048/749332104487108618/1396280573965762560)

alice-i-cecile:
Bumping to the 0.18 milestone: this isn't ready to go just yet, and isn't a release blocker. This is important to me though, so please let me / us know if you need a hand.

mrchantey:
Great catch by @NthTensor, that does sound problematic. I'm not familiar with async executors, thread woken state etc, unclogging the async runtime will need help from somebody with experience in the area.

NthTensor:
I'd be happy to help.

crates/bevy_asset/src/http_source.rs
jf908: Every asset load is already done inside of an IoTaskPool thread so is it still necessary to wrap ureq calls in another new thread? Would it actually result in less blocking or is it more just good practice because this function is async?

andriyDev: In single threaded mode, the task pool does not use separate threads. So blocking one of these loading tasks blocks the entire Bevy app.

So no, this is not just "good practice" - this is a real concern.

jf908: @NthTensor’s proposed solutions wouldn’t work in single threaded mode either though?

NthTensor: As was recently discussed on discord, "Single-threaded mode" means the bevy schedule runner is synchronous, not that the app will not spawn additional threads. In single threaded mode, we should still spawn a new thread for each request (either directly or using a pool like `unblock`).

andriyDev: This was a good question to ask!

Also for historical reasons, here's a link to the discussion: https://discord.com/channels/691052431525675048/692572690833473578/1402346312774582294

jf908:
I took a stab at solving the async issue: https://github.com/mrchantey/bevy/pull/6

I used nth's idea to use `unblock` which adds an extra dependency but it made it fairly trivial to implement!

crates/bevy_asset/src/http_source.rs
```diff
@@ -0,0 +1,271 @@
...
+
+        let mut cache_file = async_fs::File::create(&cache_path).await?;
+        cache_file.write_all(data).await?;
+
+        Ok(())
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
andriyDev, line 220: Sorry I realized this is also missing a test for the loader itself. Can we add (at least) one?

NthTensor (commented):
Looks good, using `blocking` resolves the async issues.

I believe @mockersf's concerns about domain safety are still outstanding, yes?

atlv24:
> I believe mockersf's concerns about domain safety are still outstanding, yes?

Yes

mrchantey:
closing in favor of #20660 🤞

PR #17890, opened 2025-02-17
Fix 1x1 dds textures being interpreted as 1-dimensional

### LINKED ISSUE #8615: Crash after load a DDS texture.
## Bevy version

0.10

## What you did

Use asset server to load this DDS texture ([Tex_0012_0.dds.zip](https://github.com/bevyengine/bevy/files/11477301/Tex_0012_0.dds.zip))

## What went wrong

Bevy crashed with following error:

```
thread 'Compute Task Pool (0)' panicked at 'wgpu error: Validation Error

Caused by:
    In Device::create_bind_group
    texture binding 1 expects dimension = D2, but given a view with dimension = D1

', /Users/allen/.cargo/registry/src/github.com-1ecc6299db9ec823/wgpu-0.15.1/src/backend/direct.rs:3024:5
```

rparrett:
# Objective

Fixes #8615

## Solution

Bevy currently interprets 1x1 dds textures as 1-dimensional. I think it might be more common for game engines to assume two dimensions in this ambiguous case. [citation needed]

I reworked the dimension choosing logic to only use 1d if there's a dimension > 1, and assume 2d otherwise. I kept the assumption that compressed textures are probably 2d.

## Testing

Modified `sprite.rs` to use `Tex_0012_0.dds` from the linked issue.

### DIFF
--- a/crates/bevy_image/src/dds.rs
+++ b/crates/bevy_image/src/dds.rs
@@ -66,10 +66,13 @@ pub fn dds_buffer_to_image(
     image.texture_descriptor.format = texture_format;
     image.texture_descriptor.dimension = if dds.get_depth() > 1 {
         TextureDimension::D3
-    } else if image.is_compressed() || dds.get_height() > 1 {
-        TextureDimension::D2
-    } else {
+    } else if ((dds.get_width() > 1 || dds.get_height() > 1)
+        && !(dds.get_width() > 1 && dds.get_height() > 1))
+        && !image.is_compressed()
+    {
         TextureDimension::D1
+    } else {
+        TextureDimension::D2
     };
     if is_cubemap {
         let dimension = if image.texture_descriptor.size.depth_or_array_layers > 6 {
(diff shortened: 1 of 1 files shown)


crates/bevy_image/src/dds.rs
```diff
@@ -66,10 +66,13 @@ pub fn dds_buffer_to_image(
     image.texture_descriptor.format = texture_format;
     image.texture_descriptor.dimension = if dds.get_depth() > 1 {
         TextureDimension::D3
-    } else if image.is_compressed() || dds.get_height() > 1 {
-        TextureDimension::D2
-    } else {
+    } else if ((dds.get_width() > 1 || dds.get_height() > 1)
```
alice-i-cecile, line 69:
```suggestion
    // 1x1 textures should generally be interpreted as solid 2D
    } else if ((dds.get_width() > 1 || dds.get_height() > 1)
```

alice-i-cecile (approved):
I agree with this fix: that's a much more common pattern. I've left a suggestion for a clarifying comment; feel free to revise.

IceSentry (approved):

mockersf (approved):

PR #14174, opened 2024-07-06
bevy_reflect: Feature-gate function reflection

MrGVSV:
# Objective

Function reflection requires a lot of macro code generation in the form of several `all_tuples!` invocations, as well as impls generated in the `Reflect` derive macro.

Seeing as function reflection is currently a bit more niche, it makes sense to gate it all behind a feature.

## Solution

Add a `functions` feature to `bevy_reflect`, which can be enabled in Bevy using the `reflect_functions` feature.

## Testing

You can test locally by running:

```
cargo test --package bevy_reflect
```

That should ensure that everything still works with the feature disabled.

To test with the feature on, you can run:

```
cargo test --package bevy_reflect --features functions
```

---

## Changelog

- Moved function reflection behind a Cargo feature (`bevy/reflect_functions` and `bevy_reflect/functions`)
- Add `IntoFunction` export in `bevy_reflect::prelude`

## Internal Migration Guide

> [!important]
> Function reflection was introduced as part of the 0.15 dev cycle. This migration guide was written for developers relying on `main` during this cycle, and is not a breaking change coming from 0.14.

Function reflection is now gated behind a feature. To use function reflection, enable the feature:
- If using `bevy_reflect` directly, enable the `functions` feature
- If using `bevy`, enable the `reflect_functions` feature

PR #17892, opened 2025-02-17
Added top level `reflect_documentation` feature flag.

### LINKED ISSUE #17811: No top level feature enabling `bevy_reflect/documentation`
## Bevy version

`0.15.1`

## Problem

As far as I could tell there is no way to enable this via top level features, and there's nothing mentioning the feature in the rust docs:`https://docs.rs/bevy/latest/bevy/#optional-features

AlephCubed:
Fixes #17811.


alice-i-cecile (approved):

Cargo.toml
```diff
@@ -473,6 +473,9 @@ track_location = ["bevy_internal/track_location"]
 # Enable function reflection
 reflect_functions = ["bevy_internal/reflect_functions"]
 
+# Enable documentation reflection
+reflect_documentation = ["bevy_internal/reflect_documentation"]
```
mockersf, line 477: feels a bit weird to document "documentation reflection" and then have it called `reflect_documentation`
```suggestion
documentation_reflection = ["bevy_internal/documentation_reflection"]
```

crates/bevy_internal/Cargo.toml
```diff
@@ -266,6 +266,9 @@ reflect_functions = [
   "bevy_ecs/reflect_functions",
 ]
 
+# Enable documentation reflection
+reflect_documentation = ["bevy_reflect/documentation"]
```
mockersf, line 270:
```suggestion
documentation_reflection = ["bevy_reflect/documentation"]
```

docs/cargo_features.md
```diff
@@ -90,6 +90,7 @@ The default feature set enables most of the expected features of a game engine,
 |pbr_transmission_textures|Enable support for transmission-related textures in the `StandardMaterial`, at the risk of blowing past the global, per-shader texture limit on older/lower-end GPUs|
 |pnm|PNM image format support, includes pam, pbm, pgm and ppm|
 |qoi|QOI image format support|
+|reflect_documentation|Enable documentation reflection|
```
mockersf, line 93:
```suggestion
|documentation_reflection|Enable documentation reflection|
```

Cargo.toml
AlephCubed: True, although you could say the same about `reflect_functions`. 
Would it make more sense to change the comment to match the flags?

mockersf: most features are a noun, `reflect_documentation` is an action... I would prefer to rename `reflect_function`

mockersf: well https://github.com/bevyengine/bevy/pull/14174#discussion_r1667259154
@alice-i-cecile an opinion there? should reflect be different than everything else for the sake of reflect features being grouped?

alice-i-cecile: Yeah, I think it's worth it for grouping.

mockersf (approved):

PR #17898, opened 2025-02-17
Use global binding arrays for bindless resources.

pcwalton:
Currently, Bevy's implementation of bindless resources is rather unusual: every binding in an object that implements `AsBindGroup` (most commonly, a material) becomes its own separate binding array in the shader. This is inefficient for two reasons:

1. If multiple materials reference the same texture or other resource, the reference to that resource will be duplicated many times. This increases `wgpu` validation overhead.

2. It creates many unused binding array slots. This increases `wgpu` and driver overhead and makes it easier to hit limits on APIs that `wgpu` currently imposes tight resource limits on, like Metal.

This PR fixes these issues by switching Bevy to use the standard approach in GPU-driven renderers, in which resources are de-duplicated and passed as global arrays, one for each type of resource.

Along the way, this patch introduces per-platform resource limits and bumps them from 16 resources per binding array to 64 resources per bind group on Metal and 2048 resources per bind group on other platforms. (Note that the number of resources per *binding array* isn't the same as the number of resources per *bind group*; as it currently stands, if all the PBR features are turned on, Bevy could pack as many as 496 resources into a single slab.) The limits have been increased because `wgpu` now has universal support for partially-bound binding arrays, which mean that we no longer need to fill the binding arrays with fallback resources on Direct3D 12. The `#[bindless(LIMIT)]` declaration when deriving `AsBindGroup` can now simply be written `#[bindless]` in order to have Bevy choose a default limit size for the current platform. Custom limits are still available with the new
`#[bindless(limit(LIMIT))]` syntax: e.g. `#[bindless(limit(8))]`.

The material bind group allocator has been completely rewritten. Now there are two allocators: one for bindless materials and one for non-bindless materials. The new non-bindless material allocator simply maintains a 1:1 mapping from material to bind group. The new bindless material allocator maintains a list of slabs and allocates materials into slabs on a first-fit basis. This unfortunately makes its performance O(number of resources per object * number of slabs), but the number of slabs is likely to be low, and it's planned to become even lower in the future with `wgpu` improvements. Resources are de-duplicated with in a slab and reference counted. So, for instance, if multiple materials refer to the same texture, that texture will exist only once in the appropriate binding array.

To support these new features, this patch adds the concept of a *bindless descriptor* to the `AsBindGroup` trait.  The bindless descriptor allows the material bind group allocator to probe the layout of the material, now that an array of `BindGroupLayoutEntry` records is insufficient to describe the group. The `#[derive(AsBindGroup)]` has been heavily modified to support the new features. The most important user-facing change to that macro is that the struct-level `uniform` attribute, `#[uniform(BINDING_NUMBER, StandardMaterial)]`, now reads `#[uniform(BINDLESS_INDEX, MATERIAL_UNIFORM_TYPE,
binding_array(BINDING_NUMBER)]`, allowing the material to specify the binding number for the binding array that holds the uniform data.

To make this patch simpler, I removed support for bindless `ExtendedMaterial`s, as well as field-level bindless uniform and storage buffers. I intend to add back support for these as a follow-up. Because they aren't in any released Bevy version yet, I figured this was OK.

Finally, this patch updates `StandardMaterial` for the new bindless changes. Generally, code throughout the PBR shaders that looked like `base_color_texture[slot]` now looks like
`bindless_2d_textures[material_indices[slot].base_color_texture]`.

This patch fixes a system hang that I experienced on the [Caldera test] when running with `caldera --random-materials --texture-count 100`. The time per frame is around 19.75 ms, down from 154.2 ms in Bevy 0.14: a 7.8× speedup.

[Caldera test]: https://github.com/DGriffin91/bevy_caldera_scene

### DIFF
--- a/crates/bevy_pbr/src/material_bind_groups.rs
+++ b/crates/bevy_pbr/src/material_bind_groups.rs
@@ -1,137 +1,258 @@
 //! Material bind group management for bindless resources.
 //!
-//! In bindless mode, Bevy's renderer groups materials into small bind groups.
-//! This allocator manages each bind group, assigning slots to materials as
+//! In bindless mode, Bevy's renderer groups materials into bind groups. This
+//! allocator manages each bind group, assigning slots to materials as
 //! appropriate.
 
-use crate::Material;
+use core::{marker::PhantomData, mem};
+
 use bevy_derive::{Deref, DerefMut};
 use bevy_ecs::{
     resource::Resource,
     world::{FromWorld, World},
 };
-use bevy_platform_support::collections::HashMap;
-use bevy_reflect::{std_traits::ReflectDefault, Reflect};
+use bevy_platform_support::collections::{HashMap, HashSet};
+use bevy_reflect::{prelude::ReflectDefault, Reflect};
 use bevy_render::{
     render_resource::{
-        BindGroup, BindGroupEntry, BindGroupLayout, BindGroupLayoutEntry, BindingResource,
-        BindingType, Buffer, BufferBinding, BufferInitDescriptor, BufferUsages,
-        OwnedBindingResource, Sampler, SamplerDescriptor, TextureViewDimension,
-        UnpreparedBindGroup, WgpuSampler, WgpuTextureView,
+        BindGroup, BindGroupEntry, BindGroupLayout, BindingNumber, BindingResource,
+        BindingResources, BindlessDescriptor, BindlessIndex, BindlessResourceType, Buffer,
+        BufferBinding, BufferDescriptor, BufferId, BufferUsages, CompareFunction, FilterMode,
+        OwnedBindingResource, PreparedBindGroup, RawBufferVec, Sampler, SamplerDescriptor,
+        SamplerId, TextureView, TextureViewDimension, TextureViewId, UnpreparedBindGroup,
+        WgpuSampler, WgpuTextureView,
     },
-    renderer::RenderDevice,
+    renderer::{RenderDevice, RenderQueue},
     texture::FallbackImage,
 };
 use bevy_utils::default;
-use core::{any, iter, marker::PhantomData, num::NonZero};
-use tracing::error;
+use tracing::{error, trace};
 
-/// An object that creates and stores bind groups for a single material type.
+use crate::Material;
+
+/// A resource that places materials into bind groups and tracks their
+/// resources.
 ///
-/// This object collects bindless materials into groups as appropriate and
-/// assigns slots as materials are created.
+/// Internally, Bevy has separate allocators for bindless and non-bindless
+/// materials. This resource provides a common interface to the specific
+/// allocator in use.
 #[derive(Resource)]
-pub struct MaterialBindGroupAllocator<M>
+pub enum MaterialBindGroupAllocator<M>
 where
     M: Material,
 {
-    /// The data that the allocator keeps about each bind group.
-    bind_groups: Vec<MaterialBindGroup<M>>,
+    /// The allocator used when the material is bindless.
+    Bindless(Box<MaterialBindGroupBindlessAllocator<M>>),
+    /// The allocator used when the material is non-bindless.
+    NonBindless(Box<MaterialBindGroupNonBindlessAllocator<M>>),
+}
 
-    /// Stores IDs of material bind groups that have at least one slot
-    /// available.
-    free_bind_groups: Vec<u32>,
-
-    /// The layout for this bind group.
+/// The allocator that places bindless materials into bind groups and tracks
+/// their resources.
+pub struct MaterialBindGroupBindlessAllocator<M>
+where
+    M: Material,
+{
+    /// The slabs, each of which contains a bind group.
+    slabs: Vec<MaterialBindlessSlab<M>>,
+    /// The layout of the bind groups that we produce.
     bind_group_layout: BindGroupLayout,
-
-    /// Dummy buffers that are assigned to unused slots.
-    fallback_buffers: MaterialFallbackBuffers,
-
-    /// Whether this material is actually using bindless resources.
+    /// Information about the bindless resources in the material.
     ///
-    /// This takes the availability of bindless resources on this platform into
-    /// account.
-    bindless_enabled: bool,
+    /// We use this information to create and maintain bind groups.
+    bindless_descriptor: BindlessDescriptor,
 
+    /// Dummy buffers that we use to fill empty slots in buffer binding arrays.
+    ///
+    /// There's one fallback buffer for each buffer in the bind group, each
+    /// appropriately sized. Each buffer contains one uninitialized element of
+    /// the applicable type.
+    fallback_buffers: HashMap<BindlessIndex, Buffer>,
+
+    /// The maximum number of resources that can be stored in a slab.
+    ///
+    /// This corresponds to `SLAB_CAPACITY` in the `#[bindless(SLAB_CAPACITY)]`
+    /// attribute, when deriving `AsBindGroup`.
+    slab_capacity: u32,
+}
+
+/// A single bind group and the bookkeeping necessary to allocate into it.
+pub struct MaterialBindlessSlab<M>
+where
+    M: Material,
+{
+    /// The current bind group, if it's up to date.
+    ///
+    /// If this is `None`, then the bind group is dirty and needs to be
+    /// regenerated.
+    bind_group: Option<BindGroup>,
+
+    /// A GPU-accessible buffer that holds the mapping from binding index to
+    /// bindless slot.
+    ///
+    /// This is conventionally assigned to bind group binding 0.
+    bindless_index_table: MaterialBindlessIndexTable<M>,
+
+    /// The binding arrays containing samplers.
+    samplers: HashMap<BindlessResourceType, MaterialBindlessBindingArray<Sampler>>,
+    /// The binding arrays containing textures.
+    textures: HashMap<BindlessResourceType, MaterialBindlessBindingArray<TextureView>>,
+    /// The binding arrays containing data buffers.
+    buffers: HashMap<BindlessIndex, MaterialBindlessBindingArray<Buffer>>,
+
+    /// Holds extra CPU-accessible data that the material provides.
+    ///
+    /// Typically, this data is used for constructing the material key, for
+    /// pipeline specialization purposes.
+    extra_data: Vec<Option<M::Data>>,
+
+    /// A free list of slot IDs.
+    free_slots: Vec<MaterialBindGroupSlot>,
+    /// The total number of materials currently allocated in this slab.
+    live_allocation_count: u32,
+    /// The total number of resources currently allocated in the binding arrays.
+    allocated_resource_count: u32,
+}
+
+/// A GPU-accessible buffer that holds the mapping from binding index to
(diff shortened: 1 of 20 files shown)


crates/bevy_pbr/src/material_bind_groups.rs
```diff
@@ -155,634 +276,1152 @@ impl From<u32> for MaterialBindGroupIndex {
...
+        // Invalidate the cached bind group.
         self.bind_group = None;
-    }
 
-    /// Returns true if all the slots are full or false if at least one slot in
-    /// this bind group is free.
-    fn is_full(&self) -> bool {
-        self.used_slot_bitmap == (1 << (self.unprepared_bind_groups.len() as u32)) - 1
+        // Release the slot ID.
+        self.free_slots.push(slot);
```
aevyrie, line 1059: If a bad descriptor is passed in somehow and no resources are found in the table, this code to push the slot and decrement the allocation count will still run. I *think* this is still correct, but wanted to highlight for review in case this is a real edge case.

crates/bevy_render/macros/src/as_bind_group.rs
```diff
@@ -39,6 +42,11 @@ enum BindingState<'a> {
     },
 }
 
+enum BindlessSlabResourceLimitAttr {
+    Auto,
+    Limit(LitInt),
+}
+
 pub fn derive_as_bind_group(ast: syn::DeriveInput) -> Result<TokenStream> {
```
aevyrie, line 50: This macro is terrifying.

aevyrie (approved):
I'm a bit out of my depth here technically. I can't comment on the correctness of changes to bindless, however I did review the code and documentation to the best of my ability.

Documentation, code, and comments are all easy to read, though they require some background I am missing.

crates/bevy_pbr/src/material_bind_groups.rs
pcwalton: The use of ref counting means that double frees are going to cause havoc in general unfortunately. We just have to be careful.

crates/bevy_pbr/src/render/pbr_bindings.wgsl
```diff
@@ -3,20 +3,53 @@
...
+    clearcoat_normal_texture: u32,      // 25
+    clearcoat_normal_sampler: u32,      // 26
+#endif  // PBR_MULTI_LAYER_MATERIAL_TEXTURES_SUPPORTED
+#ifdef PBR_SPECULAR_TEXTURES_SUPPORTED
+    specular_texture: u32,              // 27
+    specular_sampler: u32,              // 28
+    specular_tint_texture: u32,         // 29
+    specular_tint_sampler: u32,         // 30
+#endif  // PBR_SPECULAR_TEXTURES_SUPPORTED
+}
```
superdump, line 46: :)

crates/bevy_render/src/render_resource/bindless.rs
```diff
@@ -0,0 +1,313 @@
...
+    /// using it.
+    #[default]
+    Auto,
+
+    /// A custom value for the resource limit.
+    ///
+    /// Bevy will allocate no more than this number of resources in a slab,
+    /// unless exceeding this value is necessary in order to allocate at all
+    /// (i.e. unless the number of bindless resources in your bind group exceeds
+    /// this value), in which case Bevy can exceed it.
```
superdump, line 77: Does this exception just mean that in the cases where more bindings are needed than in this limit, we allow it and accept whatever the probably minimal performance hit is for this one material binding?

crates/bevy_render/src/render_resource/bindless.rs
```diff
@@ -0,0 +1,313 @@
...
+            BindlessSlabResourceLimit::Auto => AUTO_BINDLESS_SLAB_RESOURCE_LIMIT,
+            BindlessSlabResourceLimit::Custom(limit) => limit,
+        }
+    }
+}
+
+impl BindlessResourceType {
+    /// Returns the binding number for the common array of this resource type.
+    ///
+    /// For example, if you pass `BindlessResousrceType::Texture2d`, this will
```
superdump, line 260:
```suggestion
    /// For example, if you pass `BindlessResourceType::Texture2d`, this will
```

crates/bevy_pbr/src/material_bind_groups.rs
```diff
@@ -1,145 +1,266 @@
...
+    /// A mapping from [`MaterialBindGroupIndex`] to the bind group allocated in
+    /// each slot.
+    bind_groups: Vec<Option<MaterialNonBindlessAllocatedBindGroup<M>>>,
+    /// The bind groups that are dirty and need to be prepared.
+    ///
+    /// To prepare the bind groups, call
+    /// [`MaterialBindGroupAllocator::prepare_bind_groups`].
+    to_prepare: HashSet<MaterialBindGroupIndex>,
+    /// A free list of bind group indices.
+    free_list: Vec<MaterialBindGroupIndex>,
```
atlv24, line 185: free lists are a specific kind of data structure https://en.wikipedia.org/wiki/Free_list
```suggestion
    /// A list of free bind group indices.
    free_indices: Vec<MaterialBindGroupIndex>,
```

crates/bevy_pbr/src/material_bind_groups.rs
```diff
@@ -155,634 +276,1150 @@ impl From<u32> for MaterialBindGroupIndex {
...
+        while self.buffer.len() < range.end {
+            self.buffer.push(0);
+        }
+
+        for (&bindless_index, &resource_slot) in allocated_resource_slots {
+            self.buffer
+                .set(*bindless_index + range.start as u32, resource_slot);
         }
+
+        // Mark the buffer as needing to be recreated, in case we grew it.
```
atlv24, line 588: is it possible to detect whether the buffer grew (by saving capacity before and comparing after) and only mark it if so?

atlv24 (approved):
Nice job! I'm a bit apprehensive of the amount of panics and expects, however i think this is already very common in the rendering codebase. This is something that should be looked into in the future as a separate more targeted effort, and is not blocking for this PR.

crates/bevy_render/src/render_resource/bindless.rs
pcwalton: More or less, yes. I don't really consider it a performance hit per se because the alternative is to not be able to render at all.

crates/bevy_pbr/src/material_bind_groups.rs
pcwalton: Yes, but I'd like to do this in a follow up because the optimal is actually more complex than that: we can actually avoid recreating the bind group at all.

pcwalton:
@atlv24 Yeah, originally I had fewer panics and expects, but it really made the code messy to have all these extraneous `error!` and `return` paths.

crates/bevy_render/src/render_resource/bindless.rs
```diff
@@ -0,0 +1,313 @@
...
+    (BindlessResourceType::TextureCube, BindingNumber(8)),
+    (BindlessResourceType::TextureCubeArray, BindingNumber(9)),
+];
+
+/// The maximum number of resources that can be stored in a slab.
+///
+/// This limit primarily exists in order to work around `wgpu` performance
+/// problems involving large numbers of bindless resources. Also, some
+/// platforms, such as Metal, currently enforce limits on the number of
+/// resources in use.
```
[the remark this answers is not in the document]
pcwalton: I added the customizable limit so that applications that are seeing performance problems from large binding arrays have a way to address the issues. The idea is that, if you're seeing problems, you lower the limit until the performance is acceptable. This seems inherently application-specific, so I don't think we can query it from wgpu.

crates/bevy_render/src/render_resource/bind_group.rs
```diff
@@ -259,12 +263,32 @@ impl Deref for BindGroup {
...
+///   `BINDLESS_INDEX` specifies the offset of the buffer in the bindless index
+///   table.
+///
+///   For example, suppose that the material slot is stored in a variable named
+///   `slot`, the bindless index table is named `materials`, and that the first
+///   field (index 0) of the bindless index table type is named `material`. Then
+///   specifying `#[uniform(0, StandardMaterialUniform, binding_array(10)]` will
+///   create a binding array buffer declared in the shader as `var<storage>
+///   material: binding_array<StandardMaterialUniform>` and accessible as
+///   `material[materials[slot].material]`.
```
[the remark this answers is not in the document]
pcwalton: OK, I made these changes:
* `pbr_bindings::material` is now `pbr_bindings::material_array`.
* `pbr_bindings::material_bindings` is now `pbr_bindings::material_indices`.
* I added an ASCII art diagram to the `AsBindGroup` documentation to illustrate how this works.

crates/bevy_render/src/render_resource/bindless.rs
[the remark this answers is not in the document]
superdump: Right, and that's a good choice and useful to have. But is Metal's value purely a performance constraint on Metal because these overheads are higher? Or is it because of a limit exposed by wgpu for the current way it is implemented in wgpu?

crates/bevy_render/src/render_resource/bind_group.rs
```diff
@@ -273,30 +298,77 @@ impl Deref for BindGroup {
...
+/// |    Material Bindless Indices               +----------+----------+-----+     |   |
+/// |   +----+-----------------------------+          ^                            |   |
+/// +-->|  0 | material                    +----------+                            |   |
+///     +----+-----------------------------+                                       |   |
+///     |  1 | base_color_texture          +---------------------------------------+   |
+///     +----+-----------------------------+                                           |
+///     |  2 | base_color_sampler          +-------------------------------------------+
+///     +----+-----------------------------+
+///     | .. |            ...              |
+///     +----+-----------------------------+
```
superdump, line 370: Oh wow, very nice!!!

crates/bevy_pbr/src/material_bind_groups.rs
```diff
@@ -155,634 +276,1150 @@ impl From<u32> for MaterialBindGroupIndex {
...
+        self.resource_to_slot.insert(binding_resource_id, slot);
 
-    /// Deletes the stored bind group.
-    fn free(&mut self, _: MaterialBindGroupSlot) {
-        self.allocation = MaterialNonBindlessBindGroupAllocation::Unallocated;
-    }
+        if self.bindings.len() < slot as usize + 1 {
+            self.bindings.resize_with(slot as usize + 1, || None);
+        }
+        self.bindings[slot as usize] = Some(MaterialBindlessBinding::new(resource));
```
superdump, line 1380: Why not push?

superdump (approved):

crates/bevy_pbr/src/material_bind_groups.rs
pcwalton: I think it's slightly less code this way. Otherwise it'd be something like:
```rust
if self.bindings.len() < slot as usize + 1 {
    if self.bindings.len() < slot as usize {
        self.bindings.resize_with(slot as usize, || None);
    }
    self.bindings.push(Some(MaterialBindlessBinding::new(resource));
} else {
    self.bindings[slot as usize] = Some(MaterialBindlessBinding::new(resource));
}
```

PR #17900, opened 2025-02-17
Add byte information to `PositionedGlyph`

bytemunch:
# Objective

Adds information about the byte length and index of a glyph to `PositionedGlyph`. Useful for text picking, allows for picking with multi-byte characters. Also adds a `line` field that helps with converting back and forth from cosmic's `Cursor`.

## Solution

Retrieve the relevant data from cosmic and add it to the glyph in the text pipeline.

## Testing

`cargo r -p ci`

---

## Migration Guide

`PositionedGlyph::new()` has been removed as there is no longer an unused field. Create new `PositionedGlyph`s directly.


alice-i-cecile (approved):

greeble-dev (approved):
Clicking approve as the changes seem sensible and I've tested that example `text_debug` gives the same result. But I'm not familiar with the text code so take with salt.

I think the iteration can be simplified a bit but it's not a big deal: https://github.com/bevyengine/bevy/commit/9038ac455c629159f1cf1ef9dd48242fea907511?w=1

PR #17901, opened 2025-02-17
`GhostNode` trait

ickshonpe:
# Objective

Replace the UI's ghost node implementation with a trait that can be used with any hierarchy, not just UI `Node`s.

## Solution

Make `GhostNode` into a trait with an `Actual` associated type. 
`GhostNode` should be implemented on a marker component that must be inserted on every entity in the collapsible hierarchy.
The flattened view of the hierarchy sees only those nodes with an `Actual` component.

The UI specfic parts of the old ghost nodes implementation and the feature gate have been moved to a new module named `navigation`.

The marker component that implements `GhostNode` is called `UiNode` and the associated `Actual` type is `Node`. `Node` requires `UiNode`.

The navigation `SystemParam`s are mostly unchanged except for some renamings:
* `iter_ui_children` -> `iter_actual_children`
* `is_ui_node` -> `is_actual`

## Migration Guide


janhohenheim:
Triage: has merge conflicts
@ickshonpe  do you want to update this PR or should I tag it `S-Needs-Adoption`? :)

ickshonpe:
> Triage: has merge conflicts @ickshonpe do you want to update this PR or should I tag it `S-Needs-Adoption`? :)

I can't even remember writing these ghost node PRs. I think updating them is pointless, need to get the computed node and UI transform changes merged first.

PR #17905, opened 2025-02-17
Change ChildOf to Childof { parent: Entity} and support deriving Relationship and RelationshipTarget with named structs

### LINKED ISSUE #17896: Make `ChildOf` a named struct and support named struct `Relationship` types

As discussed in #17886, some users find the new semantics and API of `ChildOf` frustrating and confusing (see #17247 for that change).

@cart and I think that we can improve the developer experience by swapping from a tupled struct to a named struct:

```rust
struct ChildOf{
    parent: Entity
}
```

This helps clarify what's going on by giving a name to the data that's being stored.

In order to do that, you'll need to:

1.  Modify the `Relationship` derive code (actually part of the Component macro) to work with named structs with a single `Entity` field.
2. Add a doc test for this.
3. Refactor the `ChildOf` component and clean up all of the breakage.
4. Rename a few variables to `parent` to make the code cleaner.

Take a peek at the Deref / DerefMut derives for tips on implementation, although you don't need to get fancy with attributes (see below).

When this idea was initially proposed, Cart and I thought that marking a specific field as `#[relationship_target]` would be helpful to increase flexibility.

This isn't needed for the initial PR. Moreover, implementing [`Relationship::from`](https://dev-docs.bevyengine.org/bevy/ecs/relationship/trait.Relationship.html) in the derive for arbitrary structs which might contain data is not particularly feasible without adding quite a bit more complexity.

Don't bother with it for now! Advanced users can do a manual impl for now.

## Additional context

I started tackling this today, hence the detailed notes. I'm not great at macros though, so I didn't finish it and I'm shouting into the void in the hopes that one of our lovely contributors finishes it. Bother me if you want to see my draft branch, but it's not particularly helpful beyond these notes.

Bleachfuel:
# Objective

fixes #17896 

## Solution

Change ChildOf ( Entity ) to ChildOf { parent: Entity }

by doing this we also allow users to use named structs for relationship derives, When you have more than 1 field in a struct with named fields the macro will look for a field with the attribute #[relationship] and all of the other fields should implement the Default trait.  Unnamed fields are still supported. 

When u have a unnamed struct with more than one field the macro will fail. 
Do we want to support something like this ? 

```rust
 #[derive(Component)]
 #[relationship_target(relationship = ChildOf)]
 pub struct Children (#[relationship] Entity, u8);
```
I could add this, it but doesn't seem nice.
## Testing

crates/bevy_ecs - cargo test

## Showcase

```rust

use bevy_ecs::component::Component;
use bevy_ecs::entity::Entity;

 #[derive(Component)]
 pub struct ChildOf {
     #[relationship]
     pub parent: Entity,
     internal: u8,
 };

 #[derive(Component)]
 #[relationship_target(relationship = ChildOf)]
 pub struct Children {
     children: Vec<Entity>
 };

```

### DIFF
--- a/crates/bevy_ecs/src/hierarchy.rs
+++ b/crates/bevy_ecs/src/hierarchy.rs
@@ -201,9 +201,9 @@ impl<'w> EntityWorldMut<'w> {
     ///
     /// [`with_children`]: EntityWorldMut::with_children
     pub fn with_child(&mut self, bundle: impl Bundle) -> &mut Self {
-        let id = self.id();
+        let parent = self.id();
         self.world_scope(|world| {
-            world.spawn((bundle, ChildOf { parent: id }));
+            world.spawn((bundle, ChildOf { parent }));
         });
         self
     }
--- a/crates/bevy_ecs/src/hierarchy.rs
+++ b/crates/bevy_ecs/src/hierarchy.rs
@@ -380,9 +380,9 @@ mod tests {
     fn hierarchy() {
         let mut world = World::new();
         let root = world.spawn_empty().id();
-        let child1 = world.spawn(ChildOf {parent: root}).id();
-        let grandchild = world.spawn(ChildOf {parent: root}).id();
-        let child2 = world.spawn(ChildOf {parent: root}).id();
+        let child1 = world.spawn(ChildOf { parent: root }).id();
+        let grandchild = world.spawn(ChildOf { parent: root }).id();
+        let child2 = world.spawn(ChildOf { parent: root }).id();
 
         // Spawn
         let hierarchy = get_hierarchy(&world, root);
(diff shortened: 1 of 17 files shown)


alice-i-cecile:
> When u have a unnamed struct with more than one field the macro will fail.
> Do we want to support something like this ?

No, that's unpleasant, and trivially worked around.

crates/bevy_ecs/src/hierarchy.rs
```diff
@@ -198,7 +203,7 @@ impl<'w> EntityWorldMut<'w> {
     pub fn with_child(&mut self, bundle: impl Bundle) -> &mut Self {
         let id = self.id();
         self.world_scope(|world| {
-            world.spawn((bundle, ChildOf(id)));
+            world.spawn((bundle, ChildOf { parent: id }));
```
alice-i-cecile, line 206: I found the diff was nicer when I renamed these internal variables to `parent`.

alice-i-cecile (approved):
Nice work :) Thanks a ton for finishing this up; you're much more adept with the macros than I am!

crates/bevy_ecs/src/hierarchy.rs
Bleachfuel: yeah but self is not really the parent. And you can't name a variable to self. maybe make it self_id ? Or like u say make it parent.

alice-i-cecile: My suggestion is `let parent = self.id();` :)

cart (approved):
Well done! Thanks for putting this together!

PR #17907, opened 2025-02-17
Implemented `ComputedNode.logical_size()` method, returns node size after accounting for window scale

### LINKED ISSUE #17462: Add ComputedNode.logical_size() method

When dragging, we often need to compute the relative position of the mouse pointer based on the width and height of the widget. However, `node.size()` returns the physical size, not the logical size of the entity. This means that the node size and the mouse position are in different units if the window scale factor is not 1.

Fortunately, the node contains an `inverse_scale_factor` property, so getting the logical size is trivial: multiply the physical size by `inverse_scale_factor`. However, new users may not realize that they need to do this multiplication, especially if they are developing on a window system with a scale factor of 1.

Add a `logical_size()` method to `ComputedNode` which simply returns the size * `inverse_scale_factor`.

## Additional context

This is a super-trivial addition, and would be a great starter task for a novice user who is looking for a way to get started hacking on Bevy.

bsibb22:
Fixes #17462  

## Solution

- Implements a `ComputedNode.logical_size()` method, which returns the size of the node in physical pixels after accounting for the scale factor of the working window.

## Testing

- The change is a rather trivial addition and was not explicitly tested
- Added method can be tested by calling the method on windows of different sizes


crates/bevy_ui/src/ui_node.rs
```diff
@@ -118,6 +118,13 @@ impl ComputedNode {
         self.unrounded_size
     }
 
+    /// The calculated node size as width and height in physical pixels, accounting for a non-uniform window scale factor.
+    ///
```
mnmaita, line 122:
```suggestion
```
This extra line is not necessary I think?

mnmaita (approved):

crates/bevy_ui/src/ui_node.rs
```diff
@@ -118,6 +118,13 @@ impl ComputedNode {
         self.unrounded_size
     }
 
+    /// The calculated node size as width and height in physical pixels, accounting for a non-uniform window scale factor.
```
chompaa, line 121:
```suggestion
    /// The calculated node size as width and height in logical pixels, accounting for a non-uniform window scale factor.
```

chompaa:
It's worth noting this function was present before and removed. See the linked issue,

ickshonpe (changes requested):
The implementation if obviously fine.

I have problems though:
* This doesn't account for scaling by the UI node's `GlobalTransform`
* I think the `scale_factor` field of `ComputedNodeTarget` should be the single source of truth for local scale factor and that the `inverse_scale_factor` field of `ComputedNode` should be removed. `ComputedNodeTarget` is updated in the `Prepare` system set and then `ComputedNode` isn't updated until the `Layout` system set. So it could cause errors if any systems using the `inverse_scale_factor` on a `ComputedNode` after `update_ui_context_system` in `Prepare` but before `ui_layout_system` in `Layout`. `ComputedNodeTarget` needs to be a seperate component to `ComputedNode` for change detection. 
* The logical size of the node isn't that useful by itself normally, you need the position as well for hit detection etc.
* There have been lots of bugs in the UI where the wrong coordinate spaces are used. I think it's better to stick to physical coordinates as much as possible to reduce those errors and transform any coordinates that come into the UI from "outside" like the mouse position to physical coords before we use them. The removal of `logical_size` was intentional, to discourage the use of logical coordinates.

My suggestion (if you are inclined) isn't to abandon this PR but to rework it to add some helper functions. Maybe just a `contains_logical_point` function that takes a logical position, `ComputedNode`, scale factor and `GlobalTransform` and returns true if it's within the UI node's transformed bounds.

Functions that returned the normalized coords of a logical point within a UI node or an axis-aligned minimal bounding rectangle containing the UI node would also be useful.

mnmaita:
@ickshonpe hi! is this change still relevant now that there's an `UiTransform` implementation?

ickshonpe:
> @ickshonpe hi! is this change still relevant now that there's an `UiTransform` implementation?

Don't think so and it's over a year old.

PR #17908, opened 2025-02-17
Anonymous ghost nodes

ickshonpe:
# Objective

Another `GhostNode` alternative:
* Remove the GhostNode component.
* When we walk the tree to update nodes just ignore any entities in the hierarchy that don't have a Node component
* Don't use the Transform with the UI. Instead have fields on `Node` for translation, rotation and scaling. Propagate them after updating the layout during the update for `ComputedNode`s.  Same with `Visibility`.

## Solution
* UI specific transform and visibility propagation based on #16615 
* Simplified ghost navigation based on #17901
* Removed `GhostNode`

This removes the ability to have root ghost nodes, because without a ghost node marker component we can't query using a `With<GhostNode>, Without<ChildOf>` filter.

There are lots of warnings, probably have to replace the `GlobalTranform` and `Visibility` components with UI specific types to fix them. But otherwise it seems to work. Some weird problems with the imports too.


janhohenheim:
Triage: has merge conflicts
@ickshonpe  do you want to update this PR or should I tag it `S-Needs-Adoption`? :)

PR #17915, opened 2025-02-17
Make the specialized pipeline cache two-level.

pcwalton:
Currently, the specialized pipeline cache maps a (view entity, mesh entity) tuple to the retained pipeline for that entity. This causes two problems:

1. Using the view entity is incorrect, because the view entity isn't stable from frame to frame.

2. Switching the view entity to a `RetainedViewEntity`, which is necessary for correctness, significantly regresses performance of `specialize_material_meshes` and `specialize_shadows` because of the loss of the fast `EntityHash`.

This patch fixes both problems by switching to a *two-level* hash table. The outer level of the table maps each `RetainedViewEntity` to an inner table, which maps each `MainEntity` to its pipeline ID and change tick. Because we loop over views first and, within that loop, loop over entities visible from that view, we hoist the slow lookup of the view entity out of the inner entity loop.

Additionally, this patch fixes a bug whereby pipeline IDs were leaked when removing the view. We still have a problem with leaking pipeline IDs for deleted entities, but that won't be fixed until the specialized pipeline cache is retained.

This patch improves performance of the [Caldera benchmark] from 7.8× faster than 0.14 to 9.0× faster than 0.14, when applied on top of the global binding arrays PR, #17898.

[Caldera benchmark]: https://github.com/DGriffin91/bevy_caldera_scene

### DIFF
--- a/crates/bevy_pbr/src/material.rs
+++ b/crates/bevy_pbr/src/material.rs
@@ -28,6 +28,7 @@ use bevy_ecs::{
     },
 };
 use bevy_platform_support::collections::{HashMap, HashSet};
+use bevy_platform_support::hash::FixedHasher;
 use bevy_reflect::std_traits::ReflectDefault;
 use bevy_reflect::Reflect;
 use bevy_render::mesh::mark_3d_meshes_as_changed_if_their_assets_changed;
@@ -825,7 +826,7 @@ pub fn specialize_material_meshes<M: Material>(
 {
     // Record the retained IDs of all shadow views so that we can expire old
     // pipeline IDs.
-    let mut all_views = HashSet::new();
+    let mut all_views: HashSet<RetainedViewEntity, FixedHasher> = HashSet::default();
 
     for (view, visible_entities) in &views {
         all_views.insert(view.retained_view_entity);
--- a/crates/bevy_pbr/src/material.rs
+++ b/crates/bevy_pbr/src/material.rs
@@ -811,7 +811,7 @@ pub fn specialize_material_meshes<M: Material>(
         Res<ViewSortedRenderPhases<Transmissive3d>>,
         Res<ViewSortedRenderPhases<Transparent3d>>,
     ),
-    views: Query<(&MainEntity, &ExtractedView, &RenderVisibleEntities)>,
+    views: Query<(&ExtractedView, &RenderVisibleEntities)>,
     view_key_cache: Res<ViewKeyCache>,
     entity_specialization_ticks: Res<EntitySpecializationTicks<M>>,
     view_specialization_ticks: Res<ViewSpecializationTicks>,
@@ -827,7 +827,7 @@ pub fn specialize_material_meshes<M: Material>(
     // pipeline IDs.
     let mut all_views = HashSet::new();
 
-    for (view_entity, view, visible_entities) in &views {
+    for (view, visible_entities) in &views {
         all_views.insert(view.retained_view_entity);
 
         if !transparent_render_phases.contains_key(&view.retained_view_entity)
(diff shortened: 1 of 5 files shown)


superdump (approved):

tychedelia (changes requested):
Much improved, thanks. Can you just do the same for the 2d and prepass specialization checks as well?

crates/bevy_pbr/src/material.rs
```diff
@@ -817,11 +842,15 @@ pub fn specialize_material_meshes<M: Material>(
             continue;
         };
 
+        let view_tick = view_specialization_ticks.get(view_entity).unwrap();
+        let view_specialized_material_pipeline_cache = specialized_material_pipeline_cache
+            .entry(view.retained_view_entity)
+            .or_default();
+
```
brianreavis, line 849: Add an early-out here to avoid looping visible entities if not needed? Ish:
```rust 
if !entity_specialization_ticks.is_changed() && !view_tick.is_newer_than(view_specialized_material_pipeline_cache.tick, ticks.this_run())  {
    continue;
}
```

crates/bevy_pbr/src/render/light.rs
```diff
@@ -1744,13 +1768,17 @@ pub fn specialize_shadows<M: Material>(
             // NOTE: Lights with shadow mapping disabled will have no visible entities
             // so no meshes will be queued
 
+            let view_tick = light_specialization_ticks
+                .get(&extracted_view_light.retained_view_entity)
+                .unwrap();
+            let view_specialized_material_pipeline_cache = specialized_material_pipeline_cache
+                .entry(extracted_view_light.retained_view_entity)
+                .or_default();
+
```
brianreavis, line 1777: Same here

crates/bevy_pbr/src/material.rs
pcwalton: Are you sure that's correct if entities have been added since the last time `entity_specialization_ticks` was updated?

Really, the solution here is going to be to use change detection with special cases for when entities are added and removed, like our other retained passes.

pcwalton:
@tychedelia Done. I changed the `ViewSpecializationTicks` and `ViewKeyCache` to key off the `RetainedViewEntity` as well.

tychedelia (approved):

crates/bevy_pbr/src/material.rs
brianreavis: I could be missing something, but `entity_specialization_ticks` is always updated with new entities every frame. Driven by:

```rust
pub fn check_entities_needing_specialization<M>(
    needs_specialization: Query<
        Entity,
        Or<(
            Changed<Mesh3d>, // <-- Will pick up new entities
            AssetChanged<Mesh3d>,
            Changed<MeshMaterial3d<M>>, // <-- Will pick up new entities
            AssetChanged<MeshMaterial3d<M>>,
        )>,
    >,
    mut entities_needing_specialization: ResMut<EntitiesNeedingSpecialization<M>>,
) where
    M: Material,
{
    entities_needing_specialization.clear();
    for entity in &needs_specialization {
        // Feeds entity_specialization_ticks
        entities_needing_specialization.push(entity);
    }
}
```
Regardless, this can be deferred to another PR. Just thought it could be a big quick win for static scenes.

superdump (approved):

superdump:
This can be merged as soon as CI passes.

crates/bevy_pbr/src/material.rs
pcwalton: Yeah, let's defer this to another PR; this has already gotten big.

PR #8240, opened 2023-03-28
Replace `Transform` in the UI with `NodeTransform`

ickshonpe:
# Objective

Bevy UI uses `Transform` and `GlobalTransform` to manage node positions. There are lots of problems:
* Users can modify `Transform` breaking the UI layout.
* `Transform` and `GlobalTransform` are big, 112 bytes together in total.
* The UI logic is much more complex and fragile than it needs to be. Look at #8197 which has persisted despite being fixed multiple times.
* Users struggle to understand how to locate UI elements within the layout using `GlobalTransform`.
* Users naturally expect to be able to control the scaling and rotation of a UI element by modifying its `Transform`.
* `propagate_transforms` has to be run after `flex_layout_system`
* The UI node tree is walked multiple times. 

## Solution

Remove the `Transform` and `GlobalTransform` components and queries from the UI bundles and systems. Replace them with a a tuple struct component wrapping an `Affine2` called `NodeTransform`.

[Uploading Bevy App 2023-03-30 14-12-09.mp4…](https://user-images.githubusercontent.com/27962798/228847365-15464586-778c-4b51-86f4-78ac6f248837.mp4)

---

## Changelog
* Added `NodeTransform` which is a tuple struct component wrapping an `Affine2` called `NodeTransform`.
* Added `NodeTransform` to the UI node bundles.
* Removed `Transform and `GlobalTransform` from the UI node bundles.
* Removed `TransformSystem` after and before conditions from `UiPlugin`
* Modified the UI clipping, focus, extraction and flex layout systems to use `NodePosition` instead of  `Transform`/`GlobalTransform`.
* Fixed `ui_focus_system` to respond correctly to interactions with rotated and scaled elements.
* Added an example `node_transform` that demos interactions with translated, rotated and scaled buttons.
* Added `NodeTranslation`, `NodeRotation` and `NodeScale` components.

## Migration Guide

The `Transform` and `GlobalTransform` components have been replaced in Bevy UI's node bundles by the component `NodeTransform`.

The components `NodeTranslation`, `NodeRotation` and `NodeScale` can be added to UI node entities to allow translation, rotation and scaling of a node and its children.

PR #17919, opened 2025-02-17
`Node::transform` field

ickshonpe:
# Objective

Animating UI elements by modifying the Transform component of UI nodes doesn't work very well because ui_layout_system overwrites the translations each frame. The overflow_debug example uses a horrible hack where it copies the transform into the position that'll likely cause a panic if any users naively copy it.

## Solution

* Add a `transform` field to `Node`.
* In layout updates apply the layout position translation to the transform field and then `set_if_neq` it to the node entitie's `Transform` component. 

The difference with this from my other five or so UI transform PRs is that this one still uses regular transform propagation to update the GlobalTransform`s.

## Testing

Reviewers can look at the `overflow_debug` example, which is the only current UI example that modifies the transform.

### DIFF
--- a/crates/bevy_ui/src/layout/convert.rs
+++ b/crates/bevy_ui/src/layout/convert.rs
@@ -448,6 +448,8 @@ impl RepeatedGridTrack {
 
 #[cfg(test)]
 mod tests {
+    use bevy_transform::components::Transform;
+
     use super::*;
 
     #[test]
@@ -522,6 +524,7 @@ mod tests {
             ],
             grid_column: GridPlacement::start(4),
             grid_row: GridPlacement::span(3),
+            transform: Transform::IDENTITY,
         };
         let viewport_values = LayoutContext::new(1.0, bevy_math::Vec2::new(800., 600.));
         let taffy_style = from_node(&node, &viewport_values, false);
--- a/crates/bevy_ui/src/layout/mod.rs
+++ b/crates/bevy_ui/src/layout/mod.rs
@@ -288,9 +288,8 @@ with UI components as a child of an entity without UI components, your UI layout
                 .max(0.);
             }
 
-            if transform.translation.truncate() != node_center {
-                transform.translation = node_center.extend(0.);
-            }
+            transform
+                .set_if_neq(style.transform * Transform::from_translation(node_center.extend(0.)));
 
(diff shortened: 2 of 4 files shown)


aevyrie:
Isn't this a bit anti-ECS? The `Node` struct is absolutely ginormous. What's the motivation for not using `Transform`?

ickshonpe:
> Isn't this a bit anti-composition? The `Node` struct is absolutely ginormous. 

Yep, this isn't good at all, it's just the minimum I thought might possibly get merged as an intermediate step towards some of the problems with the UI transforms. #8240 is my preferred approach but it's going to take forever to update for main and it's been really difficult to reach concensus on changes like this. 

> What's the motivation for not using `Transform`?

`ui_layout_system` sets UI node's  `Transform`'s `translation` with the position offset of the node, which overwrites any value users set to the `translation` field themselves. You can write a system that adjusts the translations post layout but before transform propagation but that's kind of horrible and a lot of attempts I've seen fail with a panic. 

We could make `Transform` work by backing up the values of `Transform` before `ui_layout_system` runs and then restoring the values after the transforms are propagated but that seems really ugly.

aevyrie:
This seems like a good use case for a new transform type that better matches the semantics of how it is being used. I don't want to suggest that should be done here, but it does seem more and more compelling. Even if we make UI rendering more like the rest of our 3d rendering stack, as long as they layout system produces a `GlobalTransform`, UI can use whatever transform type works best.

I hope you don't take my initial comment as a vote against this change. I happened to stumble upon it and wanted to understand what the motivation was, the first post doesn't really explain why we want this.

ickshonpe:
> This seems like a good use case for a new transform type that better matches the semantics of how it is being used. I don't want to suggest that should be done here, but it does seem more and more compelling. Even if we make UI rendering more like the rest of our 3d rendering stack, as long as the layout system produces a `GlobalTransform`, UI can use whatever transform type works best.
> 
> I hope you don't take my initial comment as a vote against this change. I happened to stumble upon it and wanted to understand what the motivation was, the first post doesn't really explain why we want this.

Yeah I had some older PRs that added UI transform types, but it was hard to get any sort of concensus, so I was going for the lowest impact change that would fix the bugs here. I don't think it's a great approach though, closing.

PR #17920, opened 2025-02-18
Make CursorGrabMode a State

Atlas16A:
# Objective

Keeping various systems from running when the window isn't viewed/holding the mouse such as player movement requires needless extra steps.

## Solution

Makes it so that users can use cursor grab mode as a state to control system runs.


mockersf:
It's possible to have several windows with different `CursorGrabMode`. What would the state mean in this case?

I would rather not have this in Bevy, and it's easy enough for an application that needs this to newtype `CursorGrabMode` and handle it as a state on its own.

rparrett:
A couple other potential issues:
- Users would need to synchronize their window's state with this state. This seems awkward.
- This is a three-variant enum where you *most likely* only care about two variants, but need to deal with all three if you're managing the cursor in a cross-platform way.

You can probably get your desired behavior with a run condition.

```rust
app.add_systems(Update, some_system.run_if(cursor_grabbed));

fn cursor_grabbed(windows: Query<&Window, With<PrimaryWindow>>) -> bool {
    for window in &windows {
        return window.cursor_options.grab_mode != CursorGrabMode::None;
    }
    false
}
```

PR #17922, opened 2025-02-18
Don't mark newly-hidden meshes invisible until all visibility-determining systems run.

pcwalton:
The `check_visibility` system currently follows this algorithm:

1. Store all meshes that were visible last frame in the `PreviousVisibleMeshes` set.

2. Determine which meshes are visible. For each such visible mesh, remove it from `PreviousVisibleMeshes`.

3. Mark all meshes that remain in `PreviousVisibleMeshes` as invisible.

This algorithm would be correct if the `check_visibility` were the only system that marked meshes visible. However, it's not: the shadow-related systems `check_dir_light_mesh_visibility` and
`check_point_light_mesh_visibility` can as well. This results in the following sequence of events for meshes that are in a shadow map but *not* visible from a camera:

A. `check_visibility` runs, finds that no camera contains these meshes,
   and marks them hidden, which sets the changed flag.

B. `check_dir_light_mesh_visibility` and/or
   `check_point_light_mesh_visibility` run, discover that these meshes
   are visible in the shadow map, and marks them as visible, again
   setting the `ViewVisibility` changed flag.

C. During the extraction phase, the mesh extraction system sees that
   `ViewVisibility` is changed and re-extracts the mesh.

This is inefficient and results in needless work during rendering.

This patch fixes the issue in two ways:

* The `check_dir_light_mesh_visibility` and `check_point_light_mesh_visibility` systems now remove meshes that they discover from `PreviousVisibleMeshes`.

* Step (3) above has been moved from `check_visibility` to a separate system, `mark_newly_hidden_entities_invisible`. This system runs after all visibility-determining systems, ensuring that `PreviousVisibleMeshes` contains only those meshes that truly became invisible on this frame.

This fix dramatically improves the performance of [the Caldera benchmark], when combined with several other patches I've submitted.

[the Caldera benchmark]: https://github.com/DGriffin91/bevy_caldera_scene

### DIFF
--- a/crates/bevy_pbr/src/render/skin.rs
+++ b/crates/bevy_pbr/src/render/skin.rs
@@ -280,7 +280,7 @@ pub fn extract_skins(
     skinned_meshes: Extract<Query<(Entity, &SkinnedMesh)>>,
     changed_skinned_meshes: Extract<
         Query<
-            (Entity, Ref<ViewVisibility>, Ref<SkinnedMesh>),
+            (Entity, &ViewVisibility, &SkinnedMesh),
             Or<(
                 Changed<ViewVisibility>,
                 Changed<SkinnedMesh>,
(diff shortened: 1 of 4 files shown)


superdump (approved):

crates/bevy_pbr/src/render/skin.rs
tychedelia, line 1: Included by accident?

tychedelia (approved):
Awesome, this makes a lot more sense.

crates/bevy_pbr/src/render/skin.rs
pcwalton: Oops, yes. Removed.

PR #17923, opened 2025-02-18
Rewrite the multidrawable batch set builder for performance.

pcwalton:
This commit restructures the multidrawable batch set builder for better performance in various ways:

* The bin traversal is optimized to make the best use of the CPU cache.

* The inner loop that iterates over the bins, which is the hottest part of `batch_and_prepare_binned_render_phase`, has been shrunk as small as possible.

* Where possible, multiple elements are added to or reserved from GPU buffers as a batch instead of one at a time.

* Methods that LLVM wasn't inlining have been marked `#[inline]` where doing so would unlock optimizations.

This code has also been refactored to avoid duplication between the logic for indexed and non-indexed meshes via the introduction of a `MultidrawableBatchSetPreparer` object.

Together, this improved the `batch_and_prepare_binned_render_phase` time on Caldera by approximately 2×.

Eventually, we should optimize the batchable-but-not-multidrawable and unbatchable logic as well, but these meshes are much rarer, so in the interests of keeping this patch relatively small I opted to leave those to a follow-up.

### DIFF
--- a/crates/bevy_pbr/src/render/mesh.rs
+++ b/crates/bevy_pbr/src/render/mesh.rs
@@ -3129,7 +3129,7 @@ impl<P: PhaseItem> RenderCommand<P> for DrawMesh {
                         indirect_parameters_buffer.get(&TypeId::of::<P>())
                     else {
                         warn!(
-                            "Not rendering mesh because indexed indirect parameters buffer \
+                            "Not rendering mesh because non-indexed indirect parameters buffer \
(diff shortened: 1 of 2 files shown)


pcwalton:
Example runner looks good.

IceSentry (approved):

tychedelia (approved):
Thanks, the comments in the new implementation are really nice.

superdump (approved):

PR #17928, opened 2025-02-18
Add AnnularSector primitive shape

RJ:
Adds an `AnnularSector` primitive 2d shape.

<small>(a pie-slice of an annulus, which is the 2d-donut)</small>

## Solution

- Based on the existing Annulus, to which i added a `half_angle` defined and oriented the same way as `CircularSector`.

## Testing

- Visually inspected the 2d shape and the extruded mesh
- Added it to the `2d_shapes` and `3d_shapes` examples to inspect it.
- Modified the 2d_shapes example to display shapes in two rows, since it was getting crowded.
- TODO: I've not verified if `AnnularSector::closest_point` is correct, is there an example that visualizes this?

---

## Showcase

**2d_shapes example**
Screenshot 2025-02-18 at 21 59 42

**3d_shapes example**
annular sector extrusion 3d_shapes

**gratuitous screenshot**
Screenshot 2025-02-18 at 21 05 52

### DIFF
--- a/crates/bevy_math/src/primitives/dim2.rs
+++ b/crates/bevy_math/src/primitives/dim2.rs
@@ -1017,6 +1017,116 @@ impl Measured2d for Annulus {
     }
 }
 
+/// A sector of an [`Annulus`] defined by a half angle.
+/// The sector middle is at `Vec2::Y`, extending by `half_angle` radians on either side.
+#[derive(Clone, Copy, Debug, PartialEq)]
+#[cfg_attr(feature = "serialize", derive(serde::Serialize, serde::Deserialize))]
+#[cfg_attr(
+    feature = "bevy_reflect",
+    derive(Reflect),
+    reflect(Debug, PartialEq, Default)
+)]
+#[cfg_attr(
+    all(feature = "serialize", feature = "bevy_reflect"),
+    reflect(Serialize, Deserialize)
+)]
+pub struct AnnularSector {
+    /// The annulus that the sector is taken from.
+    pub annulus: Annulus,
+    /// The half angle of the sector.
+    pub half_angle: f32,
+}
+impl Primitive2d for AnnularSector {}
+
+impl Default for AnnularSector {
+    /// Returns a sector of the default [`Annulus`] with a half angle of 1.0.
+    fn default() -> Self {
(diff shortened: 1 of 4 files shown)


Jondolf:
My gut reaction is that this should probably not be its own primitive shape, but rather the `AnnulusMeshBuilder` should have a helper for limiting the sector angle for the generated mesh. A sector of an annulus feels a bit too niche to be considered a first-party geometric primitive, and I don't want to add more shapes for the sake of adding them; are there common practical use cases that would motivate this to be its own primitive?

(Personally, I kinda feel like the circular segment, circular sector, and rhombus types are already a bit too niche. The circular sector is also kind of annoying for collision detection and other geometric queries since they can be concave, like this annular sector.)

alice-i-cecile:
Can you explain what you're hoping to use this for?

rdrpenguin04:
This might just be me, but this reminds me of the ARINC 661 widget "crown". It's essentially used as a wide arc in different parts of the heads-up display that show angles, particularly roll.

I'd personally vouch for this being generally useful, and specifically useful in that it can't be constructed with the other primitive shapes. Additionally, this could effectively *be* the primitive for circles, arcs, and annuluses/annuli, if such consolidation is needed.

alice-i-cecile:
>  Additionally, this could effectively be the primitive for circles, arcs, and annuluses/annuli, if such consolidation is needed.

We should avoid this consolidation in general for our primitives: these are often performance critical, and wasted memory will have measurable impacts.

RJ:
I'm using an annular sector mesh for markers for things in stable circular orbits. I don't need the primitive shape for that, i just assemble the mesh. I added the primitive shape because I was copying the approach used by `CircularSector`.

I could change this to just be a feature of `AnnulusMeshBuilder` like @jondolf suggested, and it would still be fine for my needs.

Seems a bit arbitrary to have a circular sector primitive but not an annular sector, given circle and annulus exist. Concave shapes being a pain for physics is a good reason though, too bad CircularSector already exists.

ChristopherBiscardi:
While this could theoretically be useful, it has not gotten approvals in the last year and at least needs to be revisited for conflicts, etc in addition to being advocated for.

PR #17930, opened 2025-02-19
`extract_text_shadows` camera query fix

ickshonpe:
# Objective

 `extract_text_shadows` was still using `UiTargetCamera` and  `DefaultUiCamera` for UI camera resolution, which no longer always selects the right camera. 

To see this modify the last lines of the `multiple_windows` example from:
```rust
    commands.spawn((
        Text::new("First window"),
        node.clone(),
        // Since we are using multiple cameras, we need to specify which camera UI should be rendered to
        UiTargetCamera(first_window_camera),
    ));

    commands.spawn((
        Text::new("Second window"),
        node,
        UiTargetCamera(second_window_camera),
    ));
```
to:
```rust
    commands
        .spawn((
            node.clone(),
            // Since we are using multiple cameras, we need to specify which camera UI should be rendered to
            UiTargetCamera(first_window_camera),
        ))
        .with_child((Text::new("First window"), TextShadow::default()));

    commands
        .spawn((node, UiTargetCamera(second_window_camera)))
        .with_child((Text::new("Second window"), TextShadow::default()));
```

which results in the shadow that is meant to be displayed for the "Second Window" label instead being written over the first:

first_window_label

## Solution
Remove the `UiTargetCamera` query and the `default_camera` parameter from `extract_text_shadows` and use `UiCameraMap` with `ComputedNodeTarget` instead.

## Testing
The `multiple_windows` example for this PR has been updated to add text shadow to the window labels. You should see that it displays the "Second Window" label's shadow correctly now.


greeble-dev (approved):
Tested the `multiple_windows` example before and after fix. I'm not that familiar with `bevy_ui`, but changes seem sensible and match other UI code.

PR #17934, opened 2025-02-19
Implement occlusion culling for the deferred rendering pipeline.

pcwalton:
Deferred rendering currently doesn't support occlusion culling. This PR implements it in a straightforward way, mirroring what we already do for the non-deferred pipeline.

On the rend3 sci-fi base test scene, this resulted in roughly a 2× speedup when applied on top of my other patches. For that scene, it was useful to add another option, `--add-light`, which forces the addition of a shadow-casting light, to the scene viewer, which I included in this patch.

### DIFF
--- a/crates/bevy_core_pipeline/src/deferred/node.rs
+++ b/crates/bevy_core_pipeline/src/deferred/node.rs
@@ -1,7 +1,8 @@
 use bevy_ecs::{prelude::*, query::QueryItem};
+use bevy_render::experimental::occlusion_culling::OcclusionCulling;
 use bevy_render::render_graph::ViewNode;
 
-use bevy_render::view::ExtractedView;
+use bevy_render::view::{ExtractedView, NoIndirectDrawing};
 use bevy_render::{
     camera::ExtractedCamera,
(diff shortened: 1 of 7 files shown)


crates/bevy_core_pipeline/src/deferred/node.rs
```diff
@@ -18,76 +19,151 @@ use crate::prepass::ViewPrepassTextures;
...
-                .deferred
-                .as_ref()
-                .map(|deferred_texture| {
+    color_attachments.push(
+        view_prepass_textures
+            .deferred
+            .as_ref()
+            .map(|deferred_texture| {
+                if is_late {
+                    deferred_texture.get_attachment()
```
superdump, line 165: Should the same WebGL2, WebGPU, etc cfg guard be here too? Or practically it doesn’t matter because it won’t be set there?

superdump (approved):
Assuming there is no missing cfg check, LGTM. This was much easier to review hiding white space changes. :)

atlv24 (approved):
Looks good to me

PR #17938, opened 2025-02-19
Only despawn scene entities still in the hierarchy

### LINKED ISSUE #17883: Calling despawn_recursive() despawns entities that are no longer descendants
## Bevy version

0.15.2

## What you did

I'm making a "gib mesh" system for my game, so when an object is destroyed, it needs to activate some gib meshes and despawn the entity. For example if a space ship blows up, the space ship despawns, and some broken space ship physics objects are there instead.  I realized a good workflow for how I'm making levels would be to create child objects in Blender and mark them as a "gib". When any scene is spawned, all gibs are hidden, then when that entity blows up, it calls remove_parent_in_place() on the gibs, and then calls despawn_recursive() on the entity itself.

Then the issue arose: the gib meshes are despawned when despawn_recursive() is called, even though the gib meshes are no longer descendants.

I thought maybe I just need to wait a frame or two to let the remove_parent work, but even waiting hundreds of frames, it still happens.

## What went wrong

- what were you expecting?
I was expecting the gib meshes (child entities) to remain in the world after calling despawn_recursive on its prior parent.

- what actually happened?
The gib meshes were despawned.

## Additional information

I made a MRE just to make sure it wasn't anything else in my project causing the issue. I attached the MRE to this issue as a zip file.

[unparent.zip](https://github.com/user-attachments/files/18815776/unparent.zip)

cart:
Fixes #17883

# Objective + Solution

When doing normal scene root entity despawns (which are notably now recursive), do not despawn instanced entities that are no longer in the hierarchy.

(I would not classify this as a bug, but rather a behavior change)

## Migration Guide

If you previously relied on scene entities no longer in the hierarchy being despawned when the scene root is despawned , use `SceneSpawner::despawn_instance()` instead.

### DIFF
--- a/crates/bevy_scene/src/lib.rs
+++ b/crates/bevy_scene/src/lib.rs
@@ -81,7 +81,7 @@ impl Plugin for ScenePlugin {
                     if let Some(instance_ids) = scene_spawner.spawned_dynamic_scenes.get_mut(&id) {
                         instance_ids.remove(&scene_instance);
                     }
-                    scene_spawner.despawn_instance(scene_instance);
+                    scene_spawner.unregister_instance(scene_instance);
                 }
             });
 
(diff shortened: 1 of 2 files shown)


mockersf (approved):

Carter0 (approved):

rparrett (approved):
Cool. The previous default behavior is logical but this definitely seems like what people will want/expect most of the time.

PR #17942, opened 2025-02-20
Fix observer/hook OnReplace and OnRemove triggering when removing a bundle even when the component is not present on the entity

### LINKED ISSUE #17897: All observers triggered when any element of a bundle is removed.
## Bevy version

- Bevy 0.15.2
- Also tested on main at commit 45c26665

## What you did

1. Create an observer for `OnReplace` or `OnRemove` some component A
2. Create another observer for `OnReplace` or `OnRemove` some component B
3. Spawn an entity with some component A
4. Remove the bundle `(A, B)` from that entity

```rust
use bevy::prelude::*;

fn main() {
    App::new()
        .add_plugins(DefaultPlugins)
        .add_systems(Startup, spawn_stuff)
        .add_observer(observe_remove_a)
        .add_observer(observe_remove_b)
        .run();
}

fn spawn_stuff(mut commands: Commands) {
    let e = commands.spawn(A).id();
    commands.entity(e).remove::<(A, B)>();
}

#[derive(Component)]
struct A;

#[derive(Component)]
struct B;

fn observe_remove_a(trigger: Trigger<OnRemove, A>) {
    info!("Removed A");
}

fn observe_remove_b(trigger: Trigger<OnRemove, B>) {
    info!("Removed B");
}
```

## What went wrong

Both observers are triggered, even though the entity has no component `B`.

This only happens if you are observing the same trigger for another component in the bundle that matches. So in the above example, removing `observe_remove_a` does not result in `observe_remove_b` triggering. Or changing `observe_remove_b` to be `OnReplace` also doesn't make `observe_remove_b` trigger.

andriyDev:
# Objective

- Fixes #17897.

## Solution

- When removing components, we filter the list of components in the removed bundle based on whether they are actually in the archetype.

## Testing

- Added a test.

### DIFF
--- a/crates/bevy_ecs/src/world/entity_ref.rs
+++ b/crates/bevy_ecs/src/world/entity_ref.rs
@@ -2653,34 +2653,29 @@ unsafe fn trigger_on_replace_and_on_remove_hooks_and_observers(
     bundle_info: &BundleInfo,
     caller: MaybeLocation,
 ) {
+    let bundle_components_in_archetype = || {
+        bundle_info
+            .iter_explicit_components()
+            .filter(|component_id| archetype.contains(*component_id))
+    };
     if archetype.has_replace_observer() {
         deferred_world.trigger_observers(
             ON_REPLACE,
             entity,
-            bundle_info.iter_explicit_components(),
+            bundle_components_in_archetype(),
             caller,
         );
     }
-    deferred_world.trigger_on_replace(
-        archetype,
-        entity,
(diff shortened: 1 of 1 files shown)


fjkorf (commented):
this makes sense. and the test makes it clear what we're up to.

mnmaita (approved):

andriyDev:
@fjkorf Did you intend to approve? 😅 A second approval to move to this into final review would be great!

fjkorf (approved):

PR #17943, opened 2025-02-20
Remove `T: Default` bound for `Parallel<T>` `Default` impl

ItsDoot:
# Objective

`ThreadLocal::default()` doesn't require `T: Default`, so `Parallel<T>` shouldn't require it either.

## Solution

Replace the `Default` derive with a manually specified impl.

### DIFF
--- a/crates/bevy_utils/src/parallel_queue.rs
+++ b/crates/bevy_utils/src/parallel_queue.rs
@@ -5,7 +5,6 @@ use thread_local::ThreadLocal;
 /// A cohesive set of thread-local values of a given type.
 ///
 /// Mutable references can be fetched if `T: Default` via [`Parallel::scope`].
-#[derive(Default)]
 pub struct Parallel<T: Send> {
     locals: ThreadLocal<RefCell<T>>,
 }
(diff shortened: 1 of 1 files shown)


ItsDoot:
Eh, this isn't really necessary given the two of the main functions requires `T: Default` anyways.

ItsDoot:
😅 Actually it is useful *if* we introduce an alternative to `Parallel::borrow_local_mut` that calls `ThreadLocal::get_or` instead of `ThreadLocal::get_or_default`, which instead accepts a function to create the value.

crates/bevy_utils/src/parallel_queue.rs
```diff
@@ -21,6 +20,27 @@ impl<T: Send> Parallel<T> {
...
     }
+
+    /// Retrieves the thread-local value for the current thread and runs `f` on it.
+    ///
+    /// If there is no thread-local value, it will be initialized to the result
+    /// of `create`.
+    pub fn scope_or<R>(&self, create: impl FnOnce() -> T, f: impl FnOnce(&mut T) -> R) -> R {
+        let mut cell = self.locals.get_or(|| RefCell::new(create())).borrow_mut();
+        let ret = f(cell.deref_mut());
+        ret
```
chescock, line 31: I realize this was copied from existing code, but a few nits: 

- I don't think you need the locals here
- The first line might be simpler if you call `borrow_local_mut_or`
- I think `&mut cell` is more idiomatic than `cell.deref_mut()`

```suggestion
        f(&mut self.borrow_local_mut_or(create))
```

chescock (approved):
Looks good!  

Would it make sense to change the implementations of `scope` and `borrow_local_mut` to be `self.scope_or(Default::default, f)` and `self.borrow_local_mut_or(self, Default::default)`?  That seems like it would share more of the interesting code.

atlv24 (approved):

alice-i-cecile (approved):

alice-i-cecile:
@ItsDoot, merge conflicts need resolving :)

PR #17947, opened 2025-02-20
Fix bug in bevy_reflect serde deserialzie: OptionVisitor does not set index, when deserialzing Some

richterger:
# Objective

- When deserializing an `Option` with the value `Some` the index is not set

## Solution

- Call set_variant_with_index with and pass index 1, instead of just set_variant

## Testing

- Small bug fix, with no extra test
- I did test it in my code. Useing the deserializing Option, failed before the fix and work now


alice-i-cecile:
Can you add a quick regression test please?

PR #17950, opened 2025-02-20
do_not_recommend interned Labels

SpecificProtagonist:
# Objective

Follow up on #17441 now that `do_not_recommend` is stable.


alice-i-cecile (approved):
Yes *please*. I'd like to see more of these, but this is a good start.

PR #17951, opened 2025-02-20
Implement occlusion culling for directional light shadow maps.

pcwalton:
Two-phase occlusion culling can be helpful for shadow maps just as it can for a prepass, in order to reduce vertex and alpha mask fragment shading overhead. This patch implements occlusion culling for shadow maps from directional lights, when the `OcclusionCulling` component is present on the entities containing the lights. Shadow maps from point lights are deferred to a follow-up patch. Much of this patch involves expanding the hierarchical Z-buffer to cover shadow maps in addition to standard view depth buffers.

The `scene_viewer` example has been updated to add `OcclusionCulling` to the directional light that it creates.

This improved the performance of the rend3 sci-fi test scene when enabling shadows.

### DIFF
--- a/crates/bevy_pbr/src/render/light.rs
+++ b/crates/bevy_pbr/src/render/light.rs
@@ -1555,13 +1555,13 @@ pub fn prepare_lights(
 
                 // If this light is using occlusion culling, add the appropriate components.
                 if light.occlusion_culling {
-                    commands
-                        .entity(view_light_entity)
-                        .insert(OcclusionCulling)
-                        .insert(OcclusionCullingSubview {
+                    commands.entity(view_light_entity).insert((
+                        OcclusionCulling,
+                        OcclusionCullingSubview {
                             depth_texture_view,
                             depth_texture_size: directional_light_shadow_map.size as u32,
-                        });
+                        },
+                    ));
                     view_occlusion_culling_lights.push(view_light_entity);
                 }
 
--- a/crates/bevy_pbr/src/render/light.rs
+++ b/crates/bevy_pbr/src/render/light.rs
@@ -15,6 +15,9 @@ use bevy_ecs::{
 use bevy_math::{ops, Mat4, UVec4, Vec2, Vec3, Vec3Swizzles, Vec4, Vec4Swizzles};
 use bevy_platform_support::collections::{HashMap, HashSet};
 use bevy_platform_support::hash::FixedHasher;
+use bevy_render::experimental::occlusion_culling::{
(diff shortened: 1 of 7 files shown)


crates/bevy_pbr/src/render/light.rs
```diff
@@ -1543,6 +1553,18 @@ pub fn prepare_lights(
 
                 view_lights.push(view_light_entity);
 
+                // If this light is using occlusion culling, add the appropriate components.
+                if light.occlusion_culling {
+                    commands
+                        .entity(view_light_entity)
+                        .insert(OcclusionCulling)
```
superdump, line 1560: You know that insert accepts a tuple of components?

crates/bevy_pbr/src/render/light.rs
```diff
@@ -2115,31 +2147,97 @@ impl CachedRenderPipelinePhaseItem for Shadow {
...
+#[derive(Deref, DerefMut)]
+pub struct LateShadowPassNode(ShadowPassNode);
+
+/// Encapsulates rendering logic shared between the early and late shadow pass
+/// nodes.
 pub struct ShadowPassNode {
+    /// The query that finds cameras in which shadows are visible.
     main_view_query: QueryState<Read<ViewLightEntities>>,
-    view_light_query: QueryState<(Read<ShadowView>, Read<ExtractedView>)>,
+    /// The query that finds shadow cascades.
```
superdump, line 2170: Is it necessarily only for directional light cascade shadow maps or does it work for point and spot like shadow maps too?

superdump (approved):
I think this looks good.

crates/bevy_pbr/src/render/light.rs
pcwalton: It'd require more work to make this work for point lights and spot lights because they're cubemaps, so I opted to only go for directional lights for now. We can do point lights/spot lights in a followup.

pcwalton: I'm actually not 100% sure that it would even be worth doing for cubemaps, because Hi-Z construction for cubemaps is going to be fairly expensive.

crates/bevy_pbr/src/render/light.rs
pcwalton: Yeah, just old bad habit :) Fixed.

atlv24 (approved):
Looks good to me!

Makes me think about the possibility of having all the hzb build downsample passes for cascades or etc unified by stuffing all the depth targets into one as a future optimization. Not sure how worthwhile it would be to avoid those state changes though

PR #17953, opened 2025-02-20
Add support for experimental WESL shader source

tychedelia:
# Objective

WESL's pre-MVP `0.1.0` has been [released](https://docs.rs/wesl/latest/wesl/)!

Add support for WESL shader source so that we can begin playing and testing WESL, as well as aiding in their development.

## Solution

Adds a `ShaderSource::WESL` that can be used to load `.wesl` shaders.

Right now, we don't support mixing `naga-oil`. Additionally, WESL shaders currently need to pass through the naga frontend, which the WESL team is aware isn't great for performance (they're working on compiling into naga modules). Also, since our shaders are managed using the asset system, we don't currently support using file based imports like `super` or package scoped imports. Further work will be needed to asses how we want to support this.

---

## Showcase

See the `shader_material_wesl` example. Be sure to press space to activate party mode (trigger conditional compilation)!

https://github.com/user-attachments/assets/ec6ad19f-b6e4-4e9d-a00f-6f09336b08a4


examples/shader/shader_material_wesl.rs
```diff
@@ -0,0 +1,126 @@
...
+        .add_systems(Update, update)
+        .run();
+}
+
+/// A plugin that loads the custom material shader
+pub struct CustomMaterialPlugin;
+
+impl Plugin for CustomMaterialPlugin {
+    fn build(&self, app: &mut App) {
+        load_internal_asset!(
```
IceSentry, line 39: Is this necessary?

tychedelia: Right now, yes, because this is using the asset system for all shader imports, but my next goal is to evaluate what it would look like to use crate/filesystem based lookups.

fjkorf (approved):
This is very cool work. They're presenting on WESL at an [upcoming conference](https://www.khronos.org/events/3d-on-the-web-2025) 
The code here looks good. Well organized and implemented. I ran the demo without issue.  Glad we're adding this feature

IceSentry (approved):
I think this is good to merge. It's a nice step towards better wesl adoption and it let's us start playing with it without fully committing to it yet.

alice-i-cecile:
Thank you to everyone involved with the authoring or reviewing of this PR! This work is relatively important and needs release notes! Head over to https://github.com/bevyengine/bevy-website/issues/2002 if you'd like to help out.

PR #17956, opened 2025-02-21
Text cursor

ickshonpe:
# Objective

Draw a text cursor for a block of text.

## Solution

New `TextCursor` component. Add it to a `Text` entity and a cursor will be drawn above the glyph at the given index of the text block.

`extract_text_sections` handles the rendering.

This only draws a cursor. Movement, blink, focus, etc need to be handled by the user.

## Testing

Added a cursor to the `text` example, run with: 

```
cargo run --example text
```

---

## Showcase

cursor

### DIFF
--- a/crates/bevy_ui/src/render/mod.rs
+++ b/crates/bevy_ui/src/render/mod.rs
@@ -7,9 +7,9 @@ pub mod ui_texture_slice_pipeline;
 #[cfg(feature = "bevy_ui_debug")]
 mod debug_overlay;
 
-use crate::widget::ImageNode;
+use crate::widget::{ImageNode, TextCursor, TextCursorWidth};
 use crate::{
-    BackgroundColor, BorderColor, BoxShadowSamples, CalculatedClip, ComputedNode,
+    BackgroundColor, BorderColor, BorderRadius, BoxShadowSamples, CalculatedClip, ComputedNode,
     ComputedNodeTarget, DefaultUiCamera, Outline, ResolvedBorderRadius, TextShadow, UiAntiAlias,
     UiTargetCamera,
 };
@@ -52,7 +52,7 @@ pub use debug_overlay::UiDebugOptions;
 
 use crate::{Display, Node};
 use bevy_platform_support::collections::{HashMap, HashSet};
-use bevy_text::{ComputedTextBlock, PositionedGlyph, TextColor, TextLayoutInfo};
+use bevy_text::{ComputedTextBlock, GlyphAtlasInfo, PositionedGlyph, TextColor, TextLayoutInfo};
 use bevy_transform::components::GlobalTransform;
 use box_shadow::BoxShadowPlugin;
 use bytemuck::{Pod, Zeroable};
@@ -707,6 +707,7 @@ pub fn extract_text_sections(
             &ComputedNodeTarget,
             &ComputedTextBlock,
             &TextLayoutInfo,
+            Option<&TextCursor>,
         )>,
     >,
     text_styles: Extract<Query<&TextColor>>,
@@ -722,9 +723,10 @@ pub fn extract_text_sections(
         global_transform,
         inherited_visibility,
         clip,
-        camera,
+        target,
         computed_block,
         text_layout_info,
+        maybe_cursor,
     ) in &uinode_query
     {
         // Skip if not visible or if size is set to zero (e.g. when a parent is set to `Display::None`)
@@ -732,12 +734,12 @@ pub fn extract_text_sections(
             continue;
         }
 
-        let Some(extracted_camera_entity) = camera_mapper.map(camera) else {
+        let Some(extracted_camera_entity) = camera_mapper.map(target) else {
             continue;
         };
 
-        let transform = global_transform.affine()
-            * bevy_math::Affine3A::from_translation((-0.5 * uinode.size()).extend(0.));
+        let transform: Mat4 = global_transform.compute_matrix()
+            * Mat4::from_translation((-0.5 * uinode.size()).extend(0.));
 
         for (
             i,
@@ -788,6 +790,56 @@ pub fn extract_text_sections(
 
             end += 1;
         }
+
+        let Some(cursor) = maybe_cursor.filter(|cursor| !cursor.color.is_fully_transparent())
+        else {
+            continue;
+        };
+
+        let Some(glyph) = text_layout_info.glyphs.get(cursor.index) else {
+            continue;
+        };
+
+        let width = match cursor.width {
+            TextCursorWidth::All => glyph.size.x,
+            TextCursorWidth::Px(width) => width * target.scale_factor,
+        };
+
+        let height = computed_block.buffer().metrics().line_height;
+
+        let position = if cursor.width == TextCursorWidth::All {
+            glyph.position
+        } else {
+            glyph.position - 0.5 * (glyph.size.x + width) * Vec2::X
+        };
+
+        extracted_uinodes.uinodes.push(ExtractedUiNode {
+            render_entity: commands.spawn(TemporaryRenderEntity).id(),
+            stack_index: uinode.stack_index,
+            color: LinearRgba::from(cursor.color),
+            image: AssetId::default(),
+            clip: clip.map(|clip| clip.clip),
+            extracted_camera_entity,
+            rect: Rect {
+                min: Vec2::ZERO,
+                max: Vec2::new(width, height),
+            },
+            item: ExtractedUiItem::Node {
+                atlas_scaling: None,
+                flip_x: false,
+                flip_y: false,
+                border_radius: ResolvedBorderRadius {
+                    top_left: cursor.radius * target.scale_factor,
+                    top_right: cursor.radius * target.scale_factor,
+                    bottom_left: cursor.radius * target.scale_factor,
+                    bottom_right: cursor.radius * target.scale_factor,
+                },
+                border: BorderRect::ZERO,
+                node_type: NodeType::Rect,
+                transform: transform * Mat4::from_translation(position.extend(0.)),
+            },
+            main_entity: entity.into(),
+        });
     }
 }
 
--- a/crates/bevy_ui/src/lib.rs
+++ b/crates/bevy_ui/src/lib.rs
@@ -49,7 +49,7 @@ pub mod prelude {
     #[cfg(feature = "bevy_ui_debug")]
     pub use crate::render::UiDebugOptions;
     #[doc(hidden)]
-    pub use crate::widget::{Text, TextUiReader, TextUiWriter};
+    pub use crate::widget::{Text, TextCursor, TextCursorWidth, TextUiReader, TextUiWriter};
     #[doc(hidden)]
     pub use {
         crate::{
--- a/crates/bevy_ui/src/widget/text.rs
+++ b/crates/bevy_ui/src/widget/text.rs
@@ -397,3 +397,32 @@ pub fn text_system(
         }
     }
 }
+
+/// Add to a [`Text`] entity to draw a cursor for a block of text.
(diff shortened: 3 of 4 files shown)


crates/bevy_ui/src/render/mod.rs
```diff
@@ -788,6 +790,56 @@ pub fn extract_text_sections(
...
+
+        let width = match cursor.width {
+            TextCursorWidth::All => glyph.size.x,
+            TextCursorWidth::Px(width) => width * target.scale_factor,
+        };
+
+        let height = computed_block.buffer().metrics().line_height;
+
+        let position = if cursor.width == TextCursorWidth::All {
+            glyph.position
```
rparrett, line 811: This doesn't work for the last character of the text.

ickshonpe: Have you got an example, it worked here:

image

The index starts at 0 and ignores escape characters.

rparrett: I just modified the `text` example with
```
fn move_cursor(buttons: Res<ButtonInput<KeyCode>>, mut cursors: Query<&mut TextCursor>) {
    for mut cursor in &mut cursors {
        if buttons.just_pressed(KeyCode::ArrowLeft) {
            cursor.index = cursor.index.saturating_sub(1);
        }
        if buttons.just_pressed(KeyCode::ArrowRight) {
            cursor.index += 1;
        }
    }
}
```

https://github.com/user-attachments/assets/c4271730-f6de-4353-919d-acb655a0c8ae

rparrett: The displayed cursor height is super interesting too.

rparrett: For reference, here's `cosmic`: https://github.com/pop-os/cosmic-text/blob/main/src/edit/editor.rs#L66

rparrett: To phrase more precisely, it should be possible to place the cursor in front of the last character.

rparrett: Assuming we can get a position for the cursor when it's ahead of the last character, the next problem is that it will be clipped.

crates/bevy_ui/src/render/mod.rs
```diff
@@ -788,6 +790,56 @@ pub fn extract_text_sections(
...
+        let Some(glyph) = text_layout_info.glyphs.get(cursor.index) else {
+            continue;
+        };
+
+        let width = match cursor.width {
+            TextCursorWidth::All => glyph.size.x,
+            TextCursorWidth::Px(width) => width * target.scale_factor,
+        };
+
+        let height = computed_block.buffer().metrics().line_height * cursor.height;
```
rparrett, line 808: Line height may be unrelated to font size. I think this should be based on `max_ascent` + `max_descent`, but I'm not sure how difficult it would be to obtain that here.

crates/bevy_ui/src/render/mod.rs
ickshonpe: Oh yeah of course, I was half a sleep when I bashed this out using the glyph positions is nonsense. I made a single line scrolling input for ab_glyph and IIRC I just added an implicit space at the end of the string to make sure there was always room for the cursor.

ickshonpe: For clipping it would be the editor widget's responsibility to make sure the cursor remains in view I think.

crates/bevy_ui/src/render/mod.rs
ickshonpe: We can experiment I guess, we could even make a height enum that lets you base it off either font or line height. It's a less important part of this PR I think, we can adjust it later.

rparrett: Hm, it's hard for me to imagine line height being what people want, except when it's coincidentally a value based on font size. I've never personally seen an editor behave that way. I don't personally think that an a enum is going to be useful, but I don't mind punting this for later. People doing multi-line inputs with custom line heights are probably rare, and single-line people could probably work around the issue by using margins instead of line height.

crates/bevy_ui/src/render/mod.rs
rparrett: >  IIRC I used an implicit space added at the end

That sounds plausible / workable for me and also addresses the clipping issue. It does feel a little janky though.

rparrett: After playing around with this a bit more and dealing with clipping and cursor positioning in a single line text input, I am going to say that it is very awkward that a `Text` with a `TextCursor`'s calculated size / clipping rect do not include the cursor.

It's potentially a little less awkward now that the cosmic buffer is available, because I can at least look up the real width of the characters, even if they are a space. (though this is tedious) But if we are going to provide "Text with an optional cursor" it feels like we should support that cursor being at the end of the text.

ickshonpe: I added basic `fake` cursor support so it displays a cursor at the end if the cursor index overflows:
Screenshot 2025-02-25 172012

I'd rather just keep this PR simple and leave scrolling support to a follow up.

examples/ui/text_cursor.rs
```diff
@@ -0,0 +1,109 @@
...
+                offset: Val::Px(25.),
+            },
+        ))
+        .with_child((
+            // Accepts a `String` or any type that converts into a `String`, such as `&str`
+            Text::new(
+                "Lorem ipsum dolor sit amet,\nconsectetur adipiscing elit,\nsed do eiusmod tempor incididunt\nut labore et dolore magna aliqua."),
+            TextFont {
+                // This font is loaded and will be used instead of the default font.
+                font: asset_server.load("fonts/FiraMono-Medium.ttf"),
```
rparrett, line 64: Do we need this whole font or will the default font do?

examples/ui/text_cursor.rs
```diff
@@ -0,0 +1,109 @@
...
+    App::new()
+        .add_plugins((DefaultPlugins, FrameTimeDiagnosticsPlugin::default()))
+        .add_systems(Startup, setup)
+        .add_systems(Update, move_cursor)
+        .run();
+}
+fn setup(mut commands: Commands, asset_server: Res<AssetServer>) {
+    // UI camera
+    commands.spawn(Camera2d);
+    // Text with one section
```
rparrett, line 22: Some of these comments, presumably from another example are a little odd in this context.

crates/bevy_ui/src/render/mod.rs
```diff
@@ -788,6 +792,71 @@ pub fn extract_text_sections(
...
+                .find_map(|glyph| {
+                    if cursor.index <= glyph.end {
+                        Some((glyph.x, glyph.w))
+                    } else {
+                        None
+                    }
+                })
+                .unwrap_or((last_glyph.x + last_glyph.w, last_glyph.w))
+        } else {
+            (0., 3. * target.scale_factor)
```
rparrett, line 823: `3.`? Maybe this is meant to be a value based on the cursor width?

examples/ui/text_cursor.rs
```diff
@@ -0,0 +1,109 @@
...
+fn move_cursor(
+    keys: Res<ButtonInput<KeyCode>>,
+    mut cursors: Query<(&mut TextCursor, &mut TextCursorStyle)>,
+) {
+    for (mut cursor, mut style) in &mut cursors {
+        if keys.just_pressed(KeyCode::ArrowLeft) {
+            cursor.index = cursor.index.saturating_sub(1);
+        }
+        if keys.just_pressed(KeyCode::ArrowRight) {
+            cursor.index += 1;
```
rparrett, line 86: Presumably, this doesn't work as a user might expect with multi-byte characters?

The "first byte + line" as a user-facing API is a little confusing to me.

examples/ui/text_cursor.rs
```diff
@@ -0,0 +1,109 @@
...
+        }
+        if keys.just_pressed(KeyCode::KeyQ) {
+            cursor.line = cursor.line.saturating_sub(1);
+        }
+        if keys.just_pressed(KeyCode::ArrowDown) {
+            cursor.line += 1;
```
rparrett, line 92: I might personally prefer keeping this single line if we don't want to go through the effort of making a multi-line cursor behave well in this example.

examples/ui/text_cursor.rs
```diff
@@ -0,0 +1,109 @@
...
+        }
+        if keys.just_pressed(KeyCode::Insert) {
```
rparrett, line 94: Can we use a key that you can find on a laptop keyboard? And add some instructions?

crates/bevy_ui/src/render/mod.rs
ickshonpe: This is for the `TextCursorWidth::All` mode if the text is empty and there is no previous glyph to base the width on. Maybe we can get a default width here or it can be worked around by using an implicit space.

examples/ui/text_cursor.rs
ickshonpe: Yeah I don't like it but it's what cosmic text seems to do with their cursor indices, so I thought its better to go with it for now. Any builtin text input widgets will have a much more abstract API I hope.

rparrett: I know this is just mirroring cosmic. I think the code to move a cursor around in some text "properly" might end up being pretty gross for users though.

I don't mind providing something barebones, but I wouldn't want the example to immediately break when multi-byte text is introduced. That seems painfully ascii-centric of us.

rparrett: At the very least, maybe we just get a bit more explicit about this in comments/docs.

examples/ui/text_cursor.rs
ickshonpe: Yeah just the default is fine, it's just to contrast mono and variable width fonts.

examples/ui/text_cursor.rs
ickshonpe: We could add bounds, repeat on hold and cursor blinking but I'm not sure it's worth the effort atm, it's just here as proof that the cursors works. This example shouldn't be useful anymore once we have a working text input widget,

rparrett: That's fine. My concern was that the example felt broken rather than incomplete, which impedes its use a diagnostic tool as well.

With the text alignment change and up-button fix (didn't notice that, lol) it's feeling good enough.

examples/ui/text_cursor.rs
rparrett: I think with the cosmic buffer public now users have some options to lean on cosmic a bit for cursor movement. It's still gross because you need a mutable ref to a buffer, but it can be done.

So I'm happy to call this resolved with a nit that I wouldn't mind if the docs got tweaked a bit to warn users.

examples/ui/text_cursor.rs
rparrett: > This example shouldn't be useful anymore once we have a working text input widget,

I don't want to underestimate how long this will take. We will have probably 4-8 months of people trying to use this to create their own text inputs.

ickshonpe: Added some doc comments explaining the limitations.

rparrett (approved):
With the cosmic buffer now being public, this no longer seems *vital* to include in-engine, but it does seem useful.

viridia:
It seems like this should be part of the text input widget. What's the motivation for having a cursor for read-only text spans?

ickshonpe:
> It seems like this should be part of the text input widget. What's the motivation for having a cursor for read-only text spans?

This was made before 0.16, I didn't think I we'd get a text input widget for 0.16 and wasn't sure if I'd implement it or someone else would, so I made some PRs that were a small step towards it like this one.

PR #17958, opened 2025-02-21
Fix issue with `define_label!` instantiation in a 3rd party crate

ItsDoot:
# Objective

Calling `define_label!` in a `no_std` 3rd party crate currently requires the user to import `Box` themselves due to a non-fully-specified reference to `Box`.

## Solution

Add a fully specified path for `Box` in the one location necessary, to match all of the other cases.


bushrat011899:
CI failure will be fixed by #17959

alice-i-cecile (approved):

Issue #14195, opened 2024-07-06
Support .run_if() on observers


forbjok:
## What problem does this solve or what need does it fill?

Just like with systems, sometimes observers are only relevant for a specific state. It would be much more convenient to be able to do this on observers using .run_if() or something similar, than having to manually check the state in the function code.

## What solution would you like?

To be able to use .run_if() on observers to limit when they will run, just like on regular systems.

iiYese:
Just forwarding my thoughts from discord
> On run conditions for observers specifically they don't make sense & add a lot of complexity. This came up before in #ecs-dev. Run conditions are for things in schedules. IMO this is a problem with how bevys systems are designed. They always run regardless of if their params match. In flecs what determines if an observer runs is the observers query (queries in flecs aren't limited to iterating entities & matching overlapping archetypes). Like right now if a system asks for `Res<Foo>` where `Foo` doesn't exist it panics. If the behavior was instead that the system doesn't run then observers checking for state would just be `InState<Menu>`. Bevy's type system limitations are also showing here because doing this requires more const generics which would be too limiting anyway cause it would have zero dynamic capabilities.

forbjok:
> Like right now if a system asks for Res<Foo> where Foo doesn't exist it panics.

This is one thing that makes the lack of support for run conditions on observers even more annoying. I just ran into this as well, where when I added the manual state check to the observer function, triggering it would cause a crash because a resource that would only exist in the specific state that observer was relevant in, correctly and as expected did not exist. To work around that I also had to wrap it in an option and use a let-else statement to unwrap it or return.

benfrankel:
My take:

> Run conditions are good for composition. You write a run condition once and then you can apply it to whatever system you want. But not to observers atm.

So you couldn't use a run condition like `in_state(Foo)`, you'd have to add `Res<State<Foo>>` as an argument to your observer and check in there. For more complex run conditions, this would become very annoying.

benfrankel:
I do think that systems not running if one of their required resources does not exist would be desirable btw, I've mentioned this before here: https://github.com/bevyengine/bevy/issues/12660#issuecomment-2016641765.

Peepo-Juice:
Just want to throw my hat into this and say that I also think this should be a thing. I just ran into a use case where this would be super helpful, cus now i have to add in the manual checks inside a bunch of Observers and im not looking forward to it.

olekspickle:
[...]

PR #17962, opened 2025-02-21
Generic system config

newclarityex:
# Objective
Prevents duplicate implementation between IntoSystemConfigs and IntoSystemSetConfigs using a generic, adds a NodeType trait for more config flexibility (opening the door to implement https://github.com/bevyengine/bevy/issues/14195?).

## Solution
Followed writeup by @ItsDoot:
https://hackmd.io/@doot/rJeefFHc1x

Removes IntoSystemConfigs and IntoSystemSetConfigs, instead using IntoNodeConfigs with generics.

## Testing
Pending

---

## Showcase
N/A

## Migration Guide
SystemSetConfigs -> NodeConfigs<InternedSystemSet>
SystemConfigs -> NodeConfigs<ScheduleSystem>
IntoSystemSetConfigs -> IntoNodeConfigs<InternedSystemSet, M>
IntoSystemConfigs -> IntoNodeConfigs<ScheduleSystem, M>


crates/bevy_ecs/src/schedule/config.rs
```diff
@@ -36,24 +36,68 @@ fn ambiguous_with(graph_info: &mut GraphInfo, set: InternedSystemSet) {
     }
 }
 
+/// Stores data to differentiate different Node types
+pub trait NodeType {
+    /// Used in `NodeConfig`
```
ItsDoot, line 41:
```suggestion
    /// Additional data used to configure a node. Stored in [`NodeConfig`].
```

We could elaborate here a bit.

BTW, (for bevy at least) 99% of the time you want to wrap anything quoted in ticks with square brackets so rustdoc picks them up.

crates/bevy_ecs/src/schedule/config.rs
```diff
@@ -36,24 +36,68 @@ fn ambiguous_with(graph_info: &mut GraphInfo, set: InternedSystemSet) {
...
+    type Metadata;
+    /// Used in `NodeConfigs`
```
ItsDoot, line 43:
```suggestion
    /// Additional data used to configure a group of nodes. Stored in [`NodeConfigs`].
```
Same as above.

crates/bevy_ecs/src/schedule/config.rs
```diff
@@ -36,24 +36,68 @@ fn ambiguous_with(graph_info: &mut GraphInfo, set: InternedSystemSet) {
...
+    type GroupMetadata;
+
+    /// Initialize `NodeConfig` from `NodeType`
```
ItsDoot, line 46:
```suggestion
    /// Initializes a configuration from this node.
```
We can use normal english here since `NodeConfig` is already given as the return type, and this is a trait method on `Nodetype`.

crates/bevy_ecs/src/schedule/config.rs
```diff
@@ -36,24 +36,68 @@ fn ambiguous_with(graph_info: &mut GraphInfo, set: InternedSystemSet) {
...
+    fn config(self) -> NodeConfig<Self>
```
ItsDoot, line 47:
```suggestion
    fn into_config(self) -> NodeConfig<Self>
```
This change follows rust naming conventions for functions that take `self`.

crates/bevy_ecs/src/schedule/config.rs
```diff
@@ -36,24 +36,68 @@ fn ambiguous_with(graph_info: &mut GraphInfo, set: InternedSystemSet) {
...
+    where
+        Self: Sized;
+}
+
+impl NodeType for ScheduleSystem {
+    type Metadata = GraphInfo;
+    type GroupMetadata = Chain;
+
+    fn config(self) -> NodeConfig<Self> {
+        let sets = self.default_system_sets().into_iter().collect();
```
ItsDoot, line 57:
```suggestion
        let sets = self.default_system_sets().clone();
```
This is actually something I noticed a while ago but haven't gotten around to changing, lol.

crates/bevy_ecs/src/schedule/config.rs
```diff
@@ -62,35 +106,19 @@ pub enum NodeConfigs<T> {
         configs: Vec<NodeConfigs<T>>,
         /// Run conditions applied to everything in the tuple.
         collective_conditions: Vec<BoxedCondition>,
+        /// Metadata for `NodeConfigs`
+        metadata: T::GroupMetadata,
         /// See [`Chain`] for usage.
         chained: Chain,
```
ItsDoot, line 112:
```suggestion
```
This can be removed because it's provided by the `metadata: T::GroupMetadata` (recall `type GroupMetadata = Chain;`).

crates/bevy_ecs/src/schedule/config.rs
```diff
@@ -62,35 +106,19 @@ pub enum NodeConfigs<T> {
         configs: Vec<NodeConfigs<T>>,
         /// Run conditions applied to everything in the tuple.
         collective_conditions: Vec<BoxedCondition>,
+        /// Metadata for `NodeConfigs`
```
ItsDoot, line 109:
```suggestion
        /// Metadata to be applied to all elements in the tuple.
```
We can be a little more descriptive.

crates/bevy_ecs/src/schedule/config.rs
```diff
@@ -450,12 +505,16 @@ where
...
+    where
+        T: NodeType<Metadata = GraphInfo>,
+    {
         self.into_configs().chain_ignore_deferred()
     }
 }
 
-impl IntoSystemConfigs<()> for SystemConfigs {
+impl<T: NodeType<Metadata = GraphInfo>> IntoNodeConfigs<T, ()> for NodeConfigs<T> {
+    // impl<T: NodeType<Metadata = GraphInfo>, Marker> IntoNodeConfigs<T, Marker> for NodeConfigs<T> {
```
ItsDoot, line 517:
```suggestion
```
Left over comment?

crates/bevy_ecs/src/schedule/config.rs
```diff
@@ -496,12 +555,18 @@ impl IntoSystemConfigs<()> for SystemConfigs {
         self
     }
 
-    fn distributive_run_if<M>(mut self, condition: impl Condition<M> + Clone) -> SystemConfigs {
+    fn distributive_run_if<M>(mut self, condition: impl Condition<M> + Clone) -> NodeConfigs<T>
+    where
+        T: NodeType,
+    {
```
ItsDoot, line 561:
```suggestion
    fn distributive_run_if<M>(mut self, condition: impl Condition<M> + Clone) -> NodeConfigs<T> {
```
This is a redundant bound since the trait already specifies `T: NodeType`.

crates/bevy_ecs/src/schedule/config.rs
```diff
@@ -496,12 +555,18 @@ impl IntoSystemConfigs<()> for SystemConfigs {
...
         self.distributive_run_if_inner(condition);
         self
     }
 
-    fn run_if<M>(mut self, condition: impl Condition<M>) -> SystemConfigs {
+    fn run_if<M>(mut self, condition: impl Condition<M>) -> NodeConfigs<T>
+    where
+        T: NodeType,
+    {
```
ItsDoot, line 569:
```suggestion
    fn run_if<M>(mut self, condition: impl Condition<M>) -> NodeConfigs<T> {
```
Same as above.

crates/bevy_ecs/src/schedule/schedule.rs
```diff
@@ -372,7 +375,10 @@ impl Schedule {
 
     /// Configures a collection of system sets in this schedule, adding them if they does not exist.
     #[track_caller]
-    pub fn configure_sets(&mut self, sets: impl IntoSystemSetConfigs) -> &mut Self {
+    pub fn configure_sets<Marker>(
+        &mut self,
+        sets: impl IntoNodeConfigs<InternedSystemSet, Marker>,
+    ) -> &mut Self {
```
ItsDoot, line 381:
```suggestion
    pub fn configure_sets<M>(
        &mut self,
        sets: impl IntoNodeConfigs<InternedSystemSet, M>,
    ) -> &mut Self {
```
For functions that use the marker generic, we use `M` instead of the full `Marker` name.

crates/bevy_ecs/src/schedule/schedule.rs
```diff
@@ -950,15 +960,18 @@ impl ScheduleGraph {
     }
 
     #[track_caller]
-    fn configure_sets(&mut self, sets: impl IntoSystemSetConfigs) {
+    fn configure_sets<Marker>(&mut self, sets: impl IntoNodeConfigs<InternedSystemSet, Marker>) {
```
ItsDoot, line 963:
```suggestion
    fn configure_sets<M>(&mut self, sets: impl IntoNodeConfigs<InternedSystemSet, M>) {
```
Same deal as above.

crates/bevy_ecs/src/schedule/schedule.rs
```diff
@@ -1537,7 +1550,9 @@ struct ProcessConfigsResult {
 /// Trait used by [`ScheduleGraph::process_configs`] to process a single [`NodeConfig`].
 trait ProcessNodeConfig: Sized {
     /// Process a single [`NodeConfig`].
-    fn process_config(schedule_graph: &mut ScheduleGraph, config: NodeConfig<Self>) -> NodeId;
+    fn process_config(schedule_graph: &mut ScheduleGraph, config: NodeConfig<Self>) -> NodeId
+    where
+        Self: NodeType;
```
ItsDoot, line 1555: Since `ProcessNodeConfig` only has this single function, the `NodeType` bound should be instead moved to the trait definition like so:

```rust
trait ProcessNodeConfig: NodeType + Sized {
```

I can't directly suggest it unfortunately because GitHub won't let me.

ItsDoot (changes requested):
First pass done! Good work so far!

However, one thing I don't mention below is that the `impl_system_collection` and `impl_system_set_collection` macros that provide implementations for tuples should be merged into a single `impl_node_type_collection` macro that uses a `T: NodeType` bound instead of hardcoding to `ScheduleSystem` or `InternedSystemSet`. It should be as simple as taking one of the macros, renaming it, adding the bounds, and using `T` in place of either of the aforementioned types.

hymm:
`NodeConfigs` feels a little too generic. Would it make sense to call it `ScheduleConfigs`?

newclarityex:
> `NodeConfigs` feels a little too generic. Would it make sense to call it `ScheduleConfigs`?

Yup, updated NodeConfigs and NodeConfig to ScheduleConfigs, although kept NodeType as is.

crates/bevy_ecs/src/schedule/config.rs
```diff
@@ -36,61 +36,87 @@ fn ambiguous_with(graph_info: &mut GraphInfo, set: InternedSystemSet) {
     }
 }
 
+/// Stores data to differentiate different Node types
```
alice-i-cecile, line 39: Needs doc links to `Node`.

crates/bevy_ecs/src/schedule/config.rs
```diff
@@ -36,61 +36,87 @@ fn ambiguous_with(graph_info: &mut GraphInfo, set: InternedSystemSet) {
...
 }
 
-/// Stores configuration for a single system.
-pub type SystemConfig = NodeConfig<ScheduleSystem>;
-
-/// A collections of generic [`NodeConfig`]s.
-pub enum NodeConfigs<T> {
+/// A collections of generic [`ScheduleConfig`]s.
+pub enum ScheduleConfigs<T: NodeType> {
     /// Configuration for a single node.
```
alice-i-cecile, line 101: Is this doc comment still true?

alice-i-cecile (changes requested):
I really like the net effect here, but IMO the `NodeType` name is quite confusing, and not a good fit for a trait. How would `Schedulable` work for you?

Some little doc nits too :)

newclarityex:
> I really like the net effect here, but IMO the `NodeType` name is quite confusing, and not a good fit for a trait. How would `Schedulable` work for you?
> 
> Some little doc nits too :)

Renamed and updated some docs :>

alice-i-cecile (approved):

newclarityex:
> Manually resolved merge conflicts. You might need a cargo fmt in there. Ping me to re-attempt the merge if needed.

i think we can reattempt it now

alice-i-cecile:
Awesome, thanks :) I appreciate it: merge conflicts can be super annoying on a project this big.

mnmaita:
FYI the migration guide doesn't seem to be up to date based on some naming changes made later.

newclarityex:
> FYI the migration guide doesn't seem to be up to date based on some naming changes made later.


Updated

PR #17963, opened 2025-02-21
Reextract a mesh on the next frame if its material couldn't be prepared on the frame we first encountered it.

pcwalton:
We might not be able to prepare a material on the first frame we encounter a mesh using it for various reasons, including that the material hasn't been loaded yet or that preparing the material is exceeding the per-frame cap on number of bytes to load. When this happens, we currently try to find the material in the `MaterialBindGroupAllocator`, fail, and then fall back to group 0, slot 0, the default `MaterialBindGroupId`, which is obviously incorrect. Worse, we then fail to dirty the mesh and reextract it when we *do* finish preparing the material, so the mesh will continue to be rendered with an incorrect material.

This patch fixes both problems. In `collect_meshes_for_gpu_building`, if we fail to find a mesh's material in the `MeshBindGroupAllocator`, then we detect that case, bail out, and add it to a list, `MeshesToReextractNextFrame`. On subsequent frames, we process all the meshes in `MeshesToReextractNextFrame` as though they were changed. This ensures both that we don't render a mesh if its material hasn't been loaded and that we start rendering the mesh once its material does load.

This was first noticed in the intermittent Pixel Eagle failures in the `testbed_3d` patch in #17898, although the problem has actually existed for some time. I believe it just so happened that the changes to the allocator in that PR caused the problem to appear more commonly than it did before.

### DIFF
--- a/crates/bevy_pbr/src/render/mesh.rs
+++ b/crates/bevy_pbr/src/render/mesh.rs
@@ -11,7 +11,7 @@ use bevy_derive::{Deref, DerefMut};
 use bevy_diagnostic::FrameCount;
 use bevy_ecs::{
     prelude::*,
-    query::ROQueryItem,
+    query::{QueryData, ROQueryItem},
     system::{lifetimeless::*, SystemParamItem, SystemState},
 };
 use bevy_image::{BevyDefault, ImageSampler, TextureFormatPixelInfo};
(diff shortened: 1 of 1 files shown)


rparrett:
I documented some of the breakage that appeared after #17898 over here: https://github.com/bevyengine/bevy/issues/17970

I am not sure if there's a separate issue or if the fix here is incomplete.

hukasu:
on `deferred_rendering` example, the right cube (below the light) got fixed, but there is still a different issue on `Forward + Prepass`

superdump (approved):

atlv24 (approved):
Appreciate the minor cleanup refactors and type extraction.

PR #17965, opened 2025-02-21
Add a new `#[data]` attribute to `AsBindGroup` that allows packing data for multiple materials into a single array.

pcwalton:
Currently, the structure-level `#[uniform]` attribute of `AsBindGroup` creates a binding array of individual buffers, each of which contains data for a single material. A more efficient approach would be to provide a single buffer with an array containing all of the data for all materials in the bind group. Because `StandardMaterial` uses `#[uniform]`, this can be notably inefficient with large numbers of materials.

This patch introduces a new attribute on `AsBindGroup`, `#[data]`, which works identically to `#[uniform]` except that it concatenates all the data into a single buffer that the material bind group allocator itself manages. It also converts `StandardMaterial` to use this new functionality. This effectively provides the "material data in arrays" feature.


crates/bevy_pbr/src/material_bind_groups.rs
```diff
@@ -857,6 +915,10 @@ where
                     }
                 }
 
+                OwnedBindingResource::Data(_) => {
+                    // The size of a data buffer is unlimited.
```
superdump, line 919: Storage and uniform buffers have maximum sizes, so perhaps that is the limit here. Uniform buffer limits are likely to be hit in some cases, but storage buffer ones practically not.

superdump: Ah but this is only for bindless? Hrm. I was hoping flat buffers would be done for all cases. Do you see a way to do that? And also support uniform buffers?

pcwalton: Non-bindless support wouldn't be useful until the non-bindless allocator has the ability to merge allocations that have identical resources other than buffers, like the bindless allocator can. In order to not have allocation be O(n), this would probably require implementing some kind of hashing scheme. So I'd like to have non-bindless support, but it'd probably be better for its own PR so as not to have too much unrelated stuff in this one.

crates/bevy_render/src/render_resource/bind_group.rs
```diff
@@ -291,6 +291,49 @@ impl Deref for BindGroup {
...
+///
+/// ```ignore
+/// #[uniform(0, StandardMaterialUniform, binding_array(10))]
+/// struct StandardMaterial { ... }
+/// ```
+///
+/// This will produce a binding matching the following WGSL declaration:
+///
+/// ```wgsl
+/// @group(2) @binding(10) var<storage> material_array: binding_array<StandardMaterial>;
```
superdump, line 310: It seems strange that annotating with uniform gives a storage buffer. Hmm.

crates/bevy_render/src/render_resource/bind_group.rs
```diff
@@ -291,6 +291,49 @@ impl Deref for BindGroup {
...
+///
+/// ```ignore
+/// #[data(0, StandardMaterialUniform, binding_array(10))]
+/// struct StandardMaterial { ... }
+/// ```
+///
+/// Then Bevy produces a binding that matches this WGSL declaration instead:
+///
+/// ```wgsl
+/// @group(2) @binding(10) var<storage> material_array: array<StandardMaterial>;
```
superdump, line 323: In the bindless case only, right? We perhaps need to illustrate both here so people can copy and paste?

superdump (approved):

crates/bevy_render/src/render_resource/bind_group.rs
pcwalton: It is weird, yeah, but it's convenient, because there are a lot of cases in which you want a uniform in the non-bindless case (because UBOs are often faster) but want a storage buffer in the bindless case.

crates/bevy_render/src/render_resource/bind_group.rs
pcwalton: Good point. I clarified that this example is only for bindless mode and added a new example illustrating how it looks in non-bindless mode.

tychedelia (approved):
Given the amount of functionality in the macro, it might be nice to have an example covering some of these use cases outside of `StandardMaterial` in the engine code, but otherwise lgtm!

PR #17967, opened 2025-02-21
Upgrade to Rust Edition 2024

### LINKED ISSUE #17960: Upgrade to Rust edition 2024
Rust edition 2024 has recently released with some new features (my personal favourite is the combined doc tests for speedy testing). We should upgrade to the new edition!

bushrat011899:
# Objective

- Fixes #17960

## Solution

- Followed the [edition upgrade guide](https://doc.rust-lang.org/edition-guide/editions/transitioning-an-existing-project-to-a-new-edition.html)

## Testing

- CI

---

### Documentation Indentation

When using lists in documentation, proper indentation is now linted for. This means subsequent lines within the same list item must start at the same indentation level as the item.

```rust
/* Valid */
/// - Item 1
///   Run-on sentence.
/// - Item 2
struct Foo;

/* Invalid */
/// - Item 1
///     Run-on sentence.
/// - Item 2
struct Foo;
```

### Implicit `!` to `()` Conversion

`!` (the never return type, returned by `panic!`, etc.) no longer implicitly converts to `()`. This is particularly painful for systems with `todo!` or `panic!` statements, as they will no longer be functions returning `()` (or `Result<()>`), making them invalid systems for functions like `add_systems`. The ideal fix would be to accept functions returning `!` (or rather, _not_ returning), but this is blocked on the [stabilisation of the `!` type itself](https://doc.rust-lang.org/std/primitive.never.html), which is not done.

The "simple" fix would be to add an explicit `-> ()` to system signatures (e.g., `|| { todo!() }` becomes `|| -> () { todo!() }`). However, this is _also_ banned, as there is an existing lint which (IMO, incorrectly) marks this as an unnecessary annotation.

So, the "fix" (read: workaround) is to put these kinds of `|| -> ! { ... }` closuers into variables and give the variable an explicit type (e.g., `fn()`).

```rust
// Valid
let system: fn() = || todo!("Not implemented yet!");
app.add_systems(..., system);

// Invalid
app.add_systems(..., || todo!("Not implemented yet!"));
```

### Temporary Variable Lifetimes

The order in which temporary variables are dropped has changed. The simple fix here is _usually_ to just assign temporaries to a named variable before use.

### `gen` is a keyword

We can no longer use the name `gen` as it is reserved for a future generator syntax. This involved replacing uses of the name `gen` with `r#gen` (the raw-identifier syntax).

### Formatting has changed

Use statements have had the order of imports changed, causing a substantial +/-3,000 diff when applied. For now, I have opted-out of this change by amending `rustfmt.toml`

```toml
style_edition = "2021"
```

This preserves the original formatting for now, reducing the size of this PR. It would be a simple followup to update this to 2024 and run `cargo fmt`.

### New `use<>` Opt-Out Syntax

Lifetimes are now implicitly included in RPIT types. There was a handful of instances where it needed to be added to satisfy the borrow checker, but there may be more cases where it _should_ be added to avoid breakages in user code.

### `MyUnitStruct { .. }` is an invalid pattern

Previously, you could match against unit structs (and unit enum variants) with a `{ .. }` destructuring. This is no longer valid.

### Pretty much every use of `ref` and `mut` are gone

Pattern binding has changed to the point where these terms are largely unused now. They still serve a purpose, but it is far more niche now.

### `iter::repeat(...).take(...)` is bad

New lint recommends using the more explicit `iter::repeat_n(..., ...)` instead.

## Migration Guide

The lifetimes of functions using return-position impl-trait (RPIT) are likely _more_ conservative than they had been previously. If you encounter lifetime issues with such a function, please create an issue to investigate the addition of `+ use<...>`.

## Notes

- Check the individual commits for a clearer breakdown for what _actually_ changed.


bushrat011899:
This currently has 2 CI failures, `msrv` and `check-advisories`. `check-advisories` is due to [`ring` no longer being actively maintained](https://rustsec.org/advisories/RUSTSEC-2025-0007.html) and is unrelated to the Edition 2024 bump, so I will not attempt to resolve that issue in this PR.

`msrv` does appear to be related to the edition change, but I am unsure _how_ it is broken (the command extracting the Rust version appears to be working when I test it locally, yielding 1.85.0 as expected). As such, I'm not sure how to fix it either! I'm opening up this PR out of draft in the hopes that someone else with a greater understanding around that CI task may be able to assist!

bushrat011899:
I've marked this PR as controversial because there is active debate around whether this should be done crate-by-crate, or some other approach. I don't believe the goal of using Rust 2024 is controversial at all.

benches/benches/bevy_ecs/world/entity_hash.rs
```diff
@@ -11,16 +11,16 @@ fn make_entity(rng: &mut impl Rng, size: usize) -> Entity {
...
+    let x: f64 = rng.r#gen();
+    let r#gen = 1.0 + -(1.0 - x).log2() * 2.0;
 
     // this is not reliable, but we're internal so a hack is ok
-    let bits = ((gen as u64) << 32) | (id as u64);
+    let bits = ((r#gen as u64) << 32) | (id as u64);
     let e = Entity::from_bits(bits);
     assert_eq!(e.index(), id as u32);
-    assert_eq!(e.generation(), gen as u32);
+    assert_eq!(e.generation(), r#gen as u32);
```
mockersf, line 23: we could rename this one to `generation`

bushrat011899: Yeah for this case that makes more sense, much cleaner!

examples/asset/asset_loading.rs
```diff
@@ -36,13 +36,16 @@ fn setup(
...
+    match meshes.get(&sphere_handle) {
+        Some(sphere) => {
+            // You might notice that this doesn't run! This is because assets load in parallel without
+            // blocking. When an asset has loaded, it will appear in relevant Assets<T>
+            // collection.
+            info!("{:?}", sphere.primitive_topology());
+        }
+        _ => {
+            info!("sphere hasn't loaded yet");
+        }
```
mockersf, line 48: The change from `if let` to `match` is not needed.

The edition 2024 fix changes that to keep the same drop order, but we're ok with the change here

examples/games/alien_cake_addict.rs
```diff
@@ -271,24 +271,29 @@ fn focus_camera(
...
-    if let (Some(player_entity), Some(bonus_entity)) = (game.player.entity, game.bonus.entity) {
-        let transform_query = transforms.p1();
-        if let (Ok(player_transform), Ok(bonus_transform)) = (
-            transform_query.get(player_entity),
-            transform_query.get(bonus_entity),
-        ) {
-            game.camera_should_focus = player_transform
-                .translation
-                .lerp(bonus_transform.translation, 0.5);
+    match (game.player.entity, game.bonus.entity) {
```
mockersf, line 274: same for this `if let` -> `match` change

mockersf:
removing the controversial tag, this PR is already done and almost ready, let's not block on that as we don't have an SME matching this PR

mockersf (approved):

.github/workflows/ci.yml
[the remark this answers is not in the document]
bushrat011899: That was indeed the trick, thanks!

chescock (approved):
Looks good!  And today I learned about `repeat_n` and `div_ceil`.

crates/bevy_ecs/src/observer/mod.rs
```diff
@@ -1115,11 +1115,10 @@ mod tests {
     fn observer_despawn() {
         let mut world = World::new();
 
-        let observer = world
-            .add_observer(|_: Trigger<OnAdd, A>| {
-                panic!("Observer triggered after being despawned.")
-            })
-            .id();
+        let system: fn(Trigger<OnAdd, A>) = |_| {
```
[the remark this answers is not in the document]
bushrat011899: Agreed! I'm just going to leave it as-is because I definitely would prefer that we _actually_ support `|...| -> !` closures as systems. It'll make the eventual PR that adds that support more explicit in its benefits too.

mnmaita (approved):

mockersf:
@bushrat011899 this has some conflicts from #17840. Could you resolve them?

bushrat011899:
@mockersf Merged and CI is green!

mockersf (approved):

bushrat011899:
@mockersf Issue was caused by the use of `#[no_mangle]` within `#[bevy_main]` on mobile platforms, as in Rust Edition 2024, `no_mangle` is `unsafe`. I've added a safety comment but left a `TODO` around potentially making `bevy_main` itself `unsafe` (if that's even possible) or trying to strengthen its guards to make it safe to call.

I don't have a setup locally for checking this resolves the issue in CI, but it should be clear to merge now.

alice-i-cecile:
Thank you to everyone involved with the authoring or reviewing of this PR! This work is relatively important and needs release notes! Head over to https://github.com/bevyengine/bevy-website/issues/1996 if you'd like to help out.

PR #17971, opened 2025-02-21
Improve relationship attribute macro syntax

### LINKED ISSUE #17949: Remove "relationship" stutter in relationship attribute macro

Current

```rust 
#[derive(Component, Debug)]
#[relationship(relationship_target = TargetedBy)]
struct Targeting(Entity);

#[derive(Component, Debug)]
#[relationship_target(relationship = Targeting)]
struct TargetedBy(Vec<Entity>);
```

Proposed

```rust 
#[derive(Component, Debug)]
#[relationship(target = TargetedBy)]
struct Targeting(Entity);

#[derive(Component, Debug)]
#[target(relationship = Targeting)]
struct TargetedBy(Vec<Entity>);
```

I think the proposed syntax is more readable than what we have now.

Olle-Lukowski:
# Objective

Fixes #17949.

## Solution

Implement the suggested changes.

## Testing

Ran `cargo test` and tried the related example, still works!

### DIFF
--- a/crates/bevy_ecs/src/hierarchy.rs
+++ b/crates/bevy_ecs/src/hierarchy.rs
@@ -92,7 +92,7 @@ use log::warn;
     feature = "bevy_reflect",
     reflect(Component, PartialEq, Debug, FromWorld)
 )]
-#[relationship(relationship_target = Children)]
+#[relationship(target = Children)]
 #[doc(alias = "IsChild", alias = "Parent")]
 pub struct ChildOf(pub Entity);
 
@@ -142,7 +142,7 @@ impl FromWorld for ChildOf {
 ///
 /// [`RelationshipTarget`]: crate::relationship::RelationshipTarget
 #[derive(Component, Default, Debug, PartialEq, Eq)]
-#[relationship_target(relationship = ChildOf, linked_spawn)]
+#[target(relationship = ChildOf, linked_spawn)]
 #[cfg_attr(feature = "bevy_reflect", derive(bevy_reflect::Reflect))]
 #[cfg_attr(feature = "bevy_reflect", reflect(Component, FromWorld))]
 #[doc(alias = "IsParent")]
--- a/crates/bevy_ecs/src/hierarchy.rs
+++ b/crates/bevy_ecs/src/hierarchy.rs
@@ -142,7 +142,7 @@ impl FromWorld for ChildOf {
 ///
 /// [`RelationshipTarget`]: crate::relationship::RelationshipTarget
 #[derive(Component, Default, Debug, PartialEq, Eq)]
-#[target(relationship = ChildOf, linked_spawn)]
+#[relationship(source = ChildOf, linked_spawn)]
 #[cfg_attr(feature = "bevy_reflect", derive(bevy_reflect::Reflect))]
 #[cfg_attr(feature = "bevy_reflect", reflect(Component, FromWorld))]
 #[doc(alias = "IsParent")]
--- a/examples/ecs/relationships.rs
+++ b/examples/ecs/relationships.rs
@@ -30,7 +30,7 @@ struct Targeting(Entity);
 /// but can safely read its field. In a larger project, we could enforce this through the use of
 /// private fields and public getters.
 #[derive(Component, Debug)]
-#[target(relationship = Targeting)]
+#[relationship(source = Targeting)]
 struct TargetedBy(Vec<Entity>);
 
 fn main() {
--- a/examples/ecs/relationships.rs
+++ b/examples/ecs/relationships.rs
@@ -20,7 +20,7 @@ use bevy::prelude::*;
 /// This is the source of truth for the relationship,
 /// and can be modified directly to change the target.
 #[derive(Component, Debug)]
-#[relationship(relationship_target = TargetedBy)]
+#[relationship(target = TargetedBy)]
 struct Targeting(Entity);
 
 /// All entities that are targeting this entity.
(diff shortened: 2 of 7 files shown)


dmyyy (approved):
LGTM

Edit: There is another suggestion in the issue that I think is worth considering

crates/bevy_ecs/src/hierarchy.rs
```diff
@@ -142,7 +142,7 @@ impl FromWorld for ChildOf {
 ///
 /// [`RelationshipTarget`]: crate::relationship::RelationshipTarget
 #[derive(Component, Default, Debug, PartialEq, Eq)]
-#[relationship_target(relationship = ChildOf, linked_spawn)]
+#[target(relationship = ChildOf, linked_spawn)]
```
nxsaken, line 145: How about `#[relationship(source = ChildOf, ...)]`?

alice-i-cecile (approved):

alice-i-cecile:
I like this a fair bit better, but I'm going to let @cart do a pass on this before merging.

Olle-Lukowski:
Re-requesting review since I just implemented the syntax that seems to have the general consensus in the issue.

alice-i-cecile (approved):

examples/ecs/relationships.rs
```diff
@@ -30,7 +30,7 @@ struct Targeting(Entity);
 /// but can safely read its field. In a larger project, we could enforce this through the use of
 /// private fields and public getters.
 #[derive(Component, Debug)]
-#[relationship_target(relationship = Targeting)]
+#[relationship(source = Targeting)]
```
cart, line 33: I think that the top-level attributes should correspond _directly_ to the name of the traits being defined. As written, there is literally no indicator that this Component is a RelationshipTarget. No amount of saved text is worth that level of implicitness / context-erasure. Not including an explicit RelationshipTarget derive is already nasty.

I'm _more_ open to the `source` / `target` names to avoid stutter, but from my perspective the stutter _makes sense_. We have exactly two trait names here: `Relationship` and `RelationshipTarget`. The `relationship` attribute produces a `Relationship` trait impl. The `relationship_target` attribute produces a `RelationshipTarget` trait impl. Likewise, when these traits connect to each other, they use the `Relationship::RelationshipTarget` and `RelationshipTarget::Relationship` associated types. The 1:1 mapping here was intentional. By using `source` and `target`, we are introducing new concept names. Developers need to learn that arbitrarily `source` == `RelationshipTarget::Relationship` and `target` == `Relationship::RelationshipTarget`. I do not think this is helpful for people.

I personally think we should keep the attributes as-is, as they are the most direct maps to the types and terms that we actually use.

dmyyy: I agree that it's preferable to have attributes map to the traits they implement.

However, I also see macro attributes as a pseudo dsl that can help in understanding what's going on from a quick glance. From that perspective the current naming unfortunately feels like word soup (it certainly doesn't help that relationship is a long word).

```rust
#[relationship(relationship_target = ...)]
#[relationship_target(relationship = ...)]
```

It's difficult for me to parse what's going on at a glance - even knowing what I'm supposed to be looking at beforehand.

Personally it was a little jarring for me that `Relationship` is one end of the relationship, and `RelationshipTarget` is the other. Intuitively I felt like a `Relationship` describes the edge between two entities. It took reading the docs, and mentally putting together that `Relationship` is the "source of truth" for it to click for me. In this case I think introducing the `source` keyword is a good thing.

nxsaken: This one is similar to JPA annotations:

```rust
#[derive(Component, Debug)]
#[many_to_one(inverse = Children)] // Relationship -> ManyToOne, owning side
struct ChildOf(Entity);

#[derive(Component, Debug)]
#[one_to_many(owned_by = ChildOf)] // RelationshipTarget -> OneToMany, inverse side
struct Children(Vec<Entity>);
```

cart: Giving the "nice" Relationship name to the "relationship source" was intentional. The feature is "relationships" and the behavior-driving type is the Relationship type. I think having an actual type we can point to that "is" a Relationship is a useful simplifying concept. A Relationship is a one-directional edge. A RelationshipTarget is the collection of all reverse edges for a given relationship. To me, asking "what is this RelationshipTarget's Relationship?" is no more confusing than asking "What is this RelationshipTarget's  RelationshipSource?". The addition of the "source" context in some ways confuses things more, as it creates the question "where is the relationship?" With the current terminology, someone can ask "what is the relationship?" and the answer is easily ChildOf.

I'm still of the opinion that we should try very hard to treat concept names as actual symbols. It is true that we are building DSLs here, but from my perspective it should be the most rust-ey, statically-typed DSL it can be. One way to do that is to treat concept names and attributes as actual symbols.

I personally find the `A->B` and `B->A` symbol relationship helpful in creating a true binding between the two types.

```rust
#[derive(Component)]
struct ChildOf(Entity);

#[derive(Component)]
#[relationship_target(relationship = ChildOf)]
struct Children(Vec<Entity>);
```

This makes a number of things clear:

1. `ChildOf` _is_ the Relationship. That is a very strong hint that if I want to add a relationship, I should use that component. 
2. These are the reverse of each other, thanks to the full `A->B`, `B->A` symbol parity.
3. I know that ChildOf's `relationship_target` is Children. When I see the `RelationshipTarget` trait somewhere in code, I can trivially make the connection because the terms are identical.
4. I know that Children's `relationship` is `ChildOf`. When I see the `Relationship` trait somewhere in code, I can trivially make the connection because the terms are identical.

From my perspective, the current names and symbols are already optimal.

dmyyy: That makes sense - I can see how that maps better to the underlying implementation. I'll maintain that it can be confusing when you first start working with it - at the end of the day it's a trade off between that initial weirdness, and being faithful to not hiding what's happening under the hood.

No intent to pointlessly bike-shed so I'll defer to your judgement.

alice-i-cecile:
Closing due to Cart's disagreement with this direction. I don't feel strongly enough to pick a fight here :)

PR #17972, opened 2025-02-21
Fix unsound lifetimes in `Query::join` and `Query::join_filtered`

chescock:
# Objective

Fix unsound lifetimes in `Query::join` and `Query::join_filtered`.  

The joined query allowed access from either input query, but it only took the `'world` lifetime from `self`, not from `other`.  This meant that after the borrow of `other` ended, the joined query would unsoundly alias `other`.  

## Solution

Change the lifetimes on `join` and `join_filtered` to require mutable borrows of the *same* lifetime for the input queries.  This ensures both input queries are borrowed for the full lifetime of the joined query.  

Change `join_inner` to take `other` by value instead of reference so that the returned query is still usable without needing to borrow from a local variable.  

## Testing

Added a compile-fail test.


alice-i-cecile (approved):

crates/bevy_ecs/src/system/query.rs
```diff
@@ -2309,10 +2309,10 @@ impl<'w, 's, D: QueryData, F: QueryFilter> Query<'w, 's, D, F> {
...
     /// Like `transmute_lens` the query terms can be changed with some restrictions.
     /// See [`Self::transmute_lens`] for more details.
-    pub fn join<OtherD: QueryData, NewD: QueryData>(
-        &mut self,
-        other: &mut Query<OtherD>,
-    ) -> QueryLens<'_, NewD> {
+    pub fn join<'a, OtherD: QueryData, NewD: QueryData>(
+        &'a mut self,
+        other: &'a mut Query<OtherD>,
+    ) -> QueryLens<'a, NewD> {
```
Victoronz, line 2315: another victim of the `'w` but not `'w` lifetime, makes me wish we began using `'q` for the self-borrowing lifetime

Victoronz (approved):

alice-i-cecile:
@chescock can you resolve conflicts here / regenerate the .stderr file? I don't trust myself to do this by hand.

PR #17977, opened 2025-02-21
Renamed `EventWriter::send` methods to `write`.

### LINKED ISSUE #17856: Rename `EventWriter::send()` to `EventWriter::write()`

The `EventWriter` uses the function `.send()` to write a new event to it's queue. It's a bit confusing considering the type has Writer in it's name. Renaming it to `.write()` would remove this confusion. Also, it might help indicate a bit better that the `EventWriter` will not instantly send the event.

Rename `EventWriter::send()` to `EventWriter::write()` and do the same for `.send_default()`.

Deprecate `.send()` and `.send_default()`

Just deal with the inconsistency

## Additional context

I don't often use events, but every time I do I end up writing `.write()` because that's what the type is called.

I made a poll on discord and at the moment of writing this issue it has 16 votes to rename it to `.write()` and only 1 vote to rename it to `EventSender` instead.

AlephCubed:
Fixes #17856.

## Migration Guide
- `EventWriter::send` has been renamed to `EventWriter::write`.
- `EventWriter::send_batch` has been renamed to `EventWriter::write_batch`.
- `EventWriter::send_default` has been renamed to `EventWriter::write_default`.


mockersf (approved):

examples/ecs/event.rs
IceSentry, line 1: I know this is merged already, but the comments in this file should be updated to not say 'sent' in a  few places.

PR #17978, opened 2025-02-21
Prevent last_trigger_id from overflowing

OwlyCode:
# Objective

This prevents overflowing the `last_trigger_id` property that leads to a panic in debug mode.

```bash
panicked at C:\XXX\.cargo\registry\src\index.crates.io-6f17d22bba15001f\bevy_ecs-0.15.2\src\world\unsafe_world_cell.rs:630:18:
attempt to add with overflow
Encountered a panic when applying buffers for system `bevy_sprite::calculate_bounds_2d`!
Encountered a panic in system `bevy_ecs::schedule::executor::apply_deferred`!
```

## Solution

As this value is only used for detecting a change, we can wrap when it reaches max value.

## Testing

This can be verified by running `cargo run --example observers`


chescock (approved):
Makes sense!  

We could also think about changing the type from `u32` to `Wrapping<u32>` to make the behavior more clear, but that's probably overkill if this is the only place we modify it.

JaySpruce (approved):

Issue #17970, opened 2025-02-21
`animated_material` and other examples are broken


rparrett:
## Bevy version

main, bisected to #17898

## \[Optional\] Relevant system information

SystemInfo { os: "macOS 15.3.1 Sequoia", kernel: "24.3.0", cpu: "Apple M4 Max", core_count: "16", memory: "64.0 GiB" }    
AdapterInfo { name: "Apple M4 Max", vendor: 0, device: 0, device_type: IntegratedGpu, driver: "", driver_info: "", backend: Metal }

but this also reproduces in CI on linux/mac in [the example report](https://thebevyflock.github.io/bevy-example-runner/).

## What you did

`cargo run --example animated_material`
`cargo run --example specular_tint --features="pbr_specular_textures"`
`cargo run --example alien_cake_addict`
`cargo run --example texture`

This list probably isn't comprehensive.

## What went wrong

- `animated_material`: Colors should be changing, but they stay the same.
- `specular_tint`: Pressing space should to stuff, but it doesn't.
- `alien_cake_addict`: Panics on game over. Start the game and just let it run until game over. (there may be a [subtle issue with cake shadows](https://pixel-eagle.com/project/b25a040a-a980-4602-b90c-d480ab84076d/run/8531/compare/8519?screenshot=Games/alien_cake_addict.png), but the panic is a bit easier to notice) Can't be sure the panic reproduces in CI, it doesn't run long enough.
- `cargo run --example texture`: The three textures should have transparency and two should have red/blue tints.

<details>
<summary>Expand <code>alien_cake_addict</code> panic</summary>

```
thread 'Compute Task Pool (7)' panicked at crates/bevy_pbr/src/material_bind_groups.rs:1618:9:
assertion failed: self.bind_groups[*binding_id.group as usize].is_none()
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
Encountered a panic in system `bevy_render::render_asset::prepare_assets<bevy_pbr::material::PreparedMaterial<bevy_pbr::pbr_material::StandardMaterial>>`!
```
</details>

## Additional information

These are *not* fixed by #17963, though that PR seems to fix other examples. (e.g. `texture` is clearly fixed by that PR).

I am not sure if this is a separate issue or if that fix is incomplete.

greeble-dev:
The bug title says `animated_material` but the text says `animated_mesh`?

rparrett:
[...]

PR #17979, opened 2025-02-21
Reallocate materials when they change.

pcwalton:
PR #17898 regressed this, causing much of #17970. This commit fixes the issue by freeing and reallocating materials in the `MaterialBindGroupAllocator` on change. Note that more efficiency is possible, but I opted for the simple approach because (1) we should fix this bug ASAP; (2) I'd like #17965 to land first, because that unlocks the biggest potential optimization, which is not recreating the bind group if it isn't necessary to do so.

### DIFF
--- a/crates/bevy_pbr/src/material.rs
+++ b/crates/bevy_pbr/src/material.rs
@@ -29,6 +29,7 @@ use bevy_ecs::{
         SystemParamItem,
     },
 };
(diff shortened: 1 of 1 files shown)


rparrett:
With this PR I am now seeing the same panic I was seeing in `alien_cake_addict` in `animated_material` and `specular_tint` as well.

This may be metal-specific.

atlv24 (approved):

PR #17980, opened 2025-02-21
Fix bugs in the new non-bindless mesh material allocator.

pcwalton:
This patch fixes two bugs in the new non-bindless material allocator that landed in PR #17898:

1. A debug assertion to prevent double frees had been flipped: we checked to see whether the slot was empty before freeing, while we should have checked to see whether the slot was full.

2. The non-bindless allocator returned `None` when querying a slab that hadn't been prepared yet instead of returning a handle to that slab. This resulted in a 1-frame delay when modifying materials. In the `animated_material` example, this resulted in the meshes never showing up at all, because that example changes every material every frame.

Together with #17979, this patch locally fixes the problems with `animated_material` on macOS that were reported in #17970.

### DIFF
--- a/crates/bevy_pbr/src/material_bind_groups.rs
+++ b/crates/bevy_pbr/src/material_bind_groups.rs
@@ -342,9 +342,15 @@ where
 
 /// A single bind group that the [`MaterialBindGroupNonBindlessAllocator`]
 /// manages.
-struct MaterialNonBindlessSlab<'a, M>(&'a PreparedBindGroup<M::Data>)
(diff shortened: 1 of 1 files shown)


rparrett (approved):
Mostly over my head, but seems to make sense and look good. Applying this and 17979 seems to fix most of #17970.

Of the stuff that didn't get documented there immediately, it seems that the `texture` example is still broken, so they may yet be one more lingering issue.

atlv24 (approved):

superdump (approved):

Issue #17877, opened 2025-02-16
Loading certain GLTF assets causes extreme lag


ethereumdegen:
## Bevy version

0.15  -- this has happened since 0.12 and STILL happens 
 
## What you did
Load in certain GLTF models 

## What went wrong
 There is a warning in the console while it loads 
2025-02-16T01:22:29.929757Z  WARN bevy_gltf::loader: Unknown vertex attribute COLOR_1
2025-02-16T01:22:29.992113Z  WARN bevy_gltf::loader: Unknown vertex attribute COLOR_1
2025-02-16T01:22:29.992163Z  WARN bevy_gltf::loader: Unknown vertex attribute COLOR_2
2025-02-16T01:22:29.998664Z  WARN bevy_gltf::loader: Unknown vertex attribute COLOR_1
2025-02-16T01:22:29.998715Z  WARN bevy_gltf::loader: Unknown vertex attribute COLOR_2
2025-02-16T01:22:30.001401Z  WARN bevy_gltf::loader: Unknown vertex attribute COLOR_1
2025-02-16T01:22:30.010192Z  WARN bevy_gltf::loader: Unknown vertex attribute COLOR_1
2025-02-16T01:22:30.010268Z  WARN bevy_gltf::loader: Unknown vertex attribute COLOR_2

Then , in my game, i get 20 fps instead of 120 fps.  And mind you the model isnt being rendered its just READY to be rendered.  its loaded.

alice-i-cecile:
Can you upload or link a problematic model? We can't reproduce this without one.

torsteingrindvik:
When exporting from Blender it's important to select the option to generate mesh tangents.

Are they missing? I remember in the past that they were generated on GLTF load and it being very time consuming.

Issue #16179, opened 2024-10-30
Anisotropy example has gaps and continuity errors


hukasu:
## Bevy version

4db18d85c12c15e86a39f848829d67a6bb0ee2e0

## \[Optional\] Relevant system information

```2024-10-30T22:18:58.292727Z  INFO bevy_diagnostic::system_information_diagnostics_plugin::internal: SystemInfo { os: "Linux 24.05 NixOS", kernel: "6.6.51", cpu: "Intel(R) Core(TM) i7-14700KF", core_count: "20", memory: "31.2 GiB" }
DRM kernel driver 'nvidia-drm' in use. NVK requires nouveau.
2024-10-30T22:18:58.345559Z ERROR wgpu_hal::vulkan::instance: GENERAL [../src/nouveau/vulkan/nvk_physical_device.c:935 (0x0)]
        VK_ERROR_INCOMPATIBLE_DRIVER    
2024-10-30T22:18:58.345590Z ERROR wgpu_hal::vulkan::instance:   objects: (type: INSTANCE, hndl: 0x557e6959faf0, name: ?)    
2024-10-30T22:18:58.366831Z  INFO bevy_render::renderer: AdapterInfo { name: "NVIDIA GeForce RTX 4060 Ti", vendor: 4318, device: 10243, device_type: DiscreteGpu, driver: "NVIDIA", driver_info: "550.78", backend: Vulkan }
```

## What you did

`cargo run --example anisotropy --features="jpeg pbr_anisotropy_texture wayland"`

## What went wrong

There are gaps and continuity errors on the light reflected from the lamp.

## Additional information

![Image](https://github.com/user-attachments/assets/c1f5ad53-621a-4455-805b-451f95be387a)
![Image](https://github.com/user-attachments/assets/5634b48e-6b0f-4ac1-a192-45b5f3bf719c)
![Image](https://github.com/user-attachments/assets/00915f6d-c6e2-4782-801c-06399f83e4ab)

BenjaminBrienen:
Test after https://github.com/bevyengine/bevy/pull/16111

hukasu:
same thing after ca34e974837009d8a807e91d24366fe72c83f9bb

hukasu:
That PR is related to meshlets though, changes nothing on normal rendering

hukasu:
I added a `Sphere` to the scene to see how it looks and here is them

Anisotropy off

[...]

PR #17989, opened 2025-02-22
Add gram-schmidt fast approximation tangent generation algorithm

### LINKED ISSUE #17834: Mesh tangent generation is very slow

As the title states, generating tangents is quite slow. It's less of an issue when it's a one time operation but becomes more noticeable for frequent mesh updates or when generating many meshes. 
 
It would be nice if the API allowed for the user to select which algorithm is used for tangent computation.

```rust
impl Mesh {
  #[deprecated(note = "use compute_tangents")]
  pub fn generate_tangents(&mut self) -> Result<(), GenerateTangentsError> { .. }

  pub fn compute_tangents(&mut self, tangent_strategy: TangentStrategy) -> Result<(), GenerateTangentsError> { .. }
  
  pub fn compute_normals(&mut self) { .. }
}

pub enum TangentStrategy {
  /// Uses the Gram-Schmidt fast approximation algorithm. Produces potentially lower quality tangents, but very fast.
  FastApproximation,
  /// Uses the `mikktspace` algorithm. Produces potentially higher quality tangents, but much slower.
  HighQuality,
}
```

I'm current using the following to compute the Gram-Schmidt tangents:

EDIT: This example is incorrect and producing a tangent per face instead of per vertex. 

```rust
fn compute_tangents_fast(mesh: &mut Mesh) -> Vec<[f32; 4]> {
    let trinagle_count = mesh.indices().expect("indices").len() / 3;
    let mut tangents = Vec::with_capacity(trinagle_count);
    for triangle in mesh.triangles().expect("triangles") {
        let normal = Vec3::from(triangle.normal().expect("normal")).into();
        let verts = triangle.vertices;
        let verts: [Vec3A; 3] = [verts[0].into(), verts[1].into(), verts[2].into()];
        let edge1 = verts[1] - verts[0];
        let mut tangent = edge1 - normal * edge1.dot(normal);
        if tangent.length_squared() < f32::EPSILON {
            let edge2 = verts[2] - verts[1];
            tangent = edge2 - normal * edge2.dot(normal)
        }
        tangent = tangent.normalize();
        let bitangent_approximation = normal.cross(tangent).normalize();
        let w = if normal.cross(tangent).dot(bitangent_approximation) < 0.0 {
            -1.0
        } else {
            1.0
        };
        tangents.push(tangent.extend(w).to_array());
    }
    tangents
}
```

The method above is more than `20x` faster for the terrain mesh chunks that I'm working on and I can't visually tell any difference between them.
```
generate_tangents:      2.4916 ms  (mikktspace)
compute_tangents_fast:  0.1065 ms  (Gram-Schmidt)
```

I have a working alternative but it would be nice if this was in the bevy API.

## Additional context

I can clean up the example code and start a PR if there's interest in this. 

### mikktspace

note: the thin green line is an axis poking up through the mesh. I think the camera was oriented very slightly differently.

Image

### gram-schmidt

Image

aloucks:
# Provide a fast alternative algorithm for tangent generation

Fixes: #17834
Related (maybe): #17877

## Solution

Add a new enum `TangentStrategy` that gives the user the option to select how tangents are generated. 

The GLTF loader now has a property to set which algorithm is used for models that are missing tangents.

The `FastApproximation` strategy is implemented with the Gram-Schmidt fast approximation technique. The `HighQuality` variant uses MikkTSpace.

## Testing

Unit tests have been added to assert that both tangent computation strategies produce similar tangents for simple shapes. I also tried to check several examples to verify that the visual results are acceptable with both algorithms. I wasn't able to tell any visible difference.

> How can other people (reviewers) test your changes? Is there anything specific they need to know?

I checked the `deferred_rendering`, `parallax_mapping`, and `anisotropy` examples. Suspiciously, when I produced intentionally invalid tangents, the deferred_rendering and parallax_mapping examples seemed to render fine. I didn't try the anisotropy example with intentionally invalid values. I'm curious if there is an unrelated bug in the shader where tangents are ignored (or the examples I tested somehow didn't leverage tangents). 

## Migration Guide

- `GenerateTangentsError::MikktspaceError` has been renamed `GenerateTangentsError::AlgorithmError`
- The `Mesh::generate_tangents` and `Mesh::with_generated_tangents` methods are now deprecated. They are replaced by `compute_tangents`  and `with_computed_tangents`, both of which accept a `TangentStrategy` parameter. The naming conventino of these methods now matches `compute_normals`.

### DIFF
--- a/crates/bevy_gltf/src/lib.rs
+++ b/crates/bevy_gltf/src/lib.rs
@@ -102,7 +102,7 @@ use bevy_platform_support::collections::HashMap;
 use bevy_app::prelude::*;
 use bevy_asset::AssetApp;
 use bevy_image::CompressedImageFormats;
-use bevy_mesh::{MeshVertexAttribute, TangentCalculationStrategy};
+use bevy_mesh::{MeshVertexAttribute, TangentAlgorithm};
 use bevy_render::renderer::RenderDevice;
 
 /// The glTF prelude.
@@ -119,8 +119,8 @@ pub use {assets::*, label::GltfAssetLabel, loader::*};
 #[derive(Default)]
 pub struct GltfPlugin {
     custom_vertex_attributes: HashMap<Box<str>, MeshVertexAttribute>,
-    /// The strategy to use when computing tangents for meshes without them.
-    pub tangent_calculation_strategy: TangentCalculationStrategy,
+    /// The algorithm to use when computing tangents for meshes without them.
+    pub tangent_calculation_algorithm: TangentAlgorithm,
 }
 
 impl GltfPlugin {
@@ -138,12 +138,12 @@ impl GltfPlugin {
         self
     }
 
-    /// The strategy to use when computing mesh tangents.
-    pub fn with_tangent_calculation_strategy(
+    /// The algorithm to use when computing mesh tangents.
+    pub fn with_tangent_calculation_algorithm(
         mut self,
-        tangent_strategy: TangentCalculationStrategy,
+        tangent_algorithm: TangentAlgorithm,
     ) -> Self {
-        self.tangent_calculation_strategy = tangent_strategy;
+        self.tangent_calculation_algorithm = tangent_algorithm;
         self
     }
 }
@@ -171,7 +171,7 @@ impl Plugin for GltfPlugin {
         app.register_asset_loader(GltfLoader {
             supported_compressed_formats,
             custom_vertex_attributes: self.custom_vertex_attributes.clone(),
-            tangent_calculation_strategy: self.tangent_calculation_strategy,
+            tangent_calculation_algorithm: self.tangent_calculation_algorithm,
         });
     }
 }
--- a/crates/bevy_gltf/src/lib.rs
+++ b/crates/bevy_gltf/src/lib.rs
@@ -102,7 +102,7 @@ use bevy_platform_support::collections::HashMap;
 use bevy_app::prelude::*;
 use bevy_asset::AssetApp;
 use bevy_image::CompressedImageFormats;
-use bevy_mesh::{MeshVertexAttribute, TangentStrategy};
+use bevy_mesh::{MeshVertexAttribute, TangentCalculationStrategy};
 use bevy_render::renderer::RenderDevice;
 
 /// The glTF prelude.
@@ -119,8 +119,8 @@ pub use {assets::*, label::GltfAssetLabel, loader::*};
 #[derive(Default)]
 pub struct GltfPlugin {
     custom_vertex_attributes: HashMap<Box<str>, MeshVertexAttribute>,
-    /// The strategy to use when computing mesh tangents.
-    pub computed_tangent_strategy: TangentStrategy,
+    /// The strategy to use when computing tangents for meshes without them.
+    pub tangent_calculation_strategy: TangentCalculationStrategy,
 }
 
 impl GltfPlugin {
@@ -139,8 +139,11 @@ impl GltfPlugin {
     }
 
     /// The strategy to use when computing mesh tangents.
-    pub fn with_computed_tangent_strategy(mut self, tangent_strategy: TangentStrategy) -> Self {
-        self.computed_tangent_strategy = tangent_strategy;
+    pub fn with_tangent_calculation_strategy(
+        mut self,
+        tangent_strategy: TangentCalculationStrategy,
+    ) -> Self {
+        self.tangent_calculation_strategy = tangent_strategy;
         self
     }
 }
@@ -168,7 +171,7 @@ impl Plugin for GltfPlugin {
         app.register_asset_loader(GltfLoader {
             supported_compressed_formats,
             custom_vertex_attributes: self.custom_vertex_attributes.clone(),
-            computed_tangent_strategy: self.computed_tangent_strategy,
+            tangent_calculation_strategy: self.tangent_calculation_strategy,
         });
     }
 }
--- a/crates/bevy_gltf/src/lib.rs
+++ b/crates/bevy_gltf/src/lib.rs
@@ -102,7 +102,7 @@ use bevy_platform_support::collections::HashMap;
 use bevy_app::prelude::*;
 use bevy_asset::AssetApp;
 use bevy_image::CompressedImageFormats;
-use bevy_mesh::MeshVertexAttribute;
+use bevy_mesh::{MeshVertexAttribute, TangentStrategy};
 use bevy_render::renderer::RenderDevice;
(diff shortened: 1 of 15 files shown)


crates/bevy_gltf/src/lib.rs
```diff
@@ -124,6 +124,8 @@ pub mod prelude {
 #[derive(Default)]
 pub struct GltfPlugin {
     custom_vertex_attributes: HashMap<Box<str>, MeshVertexAttribute>,
+    /// The strategy to use when computing mesh tangents.
+    pub computed_tangent_strategy: TangentStrategy,
```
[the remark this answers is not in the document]
aloucks: renamed

hukasu:
the `anisotropy` example does use tangents, this issue shows some issues caused by not optimal tangents #16179

hukasu:
i was making tangent generation for the primitives, but with a very naive algorithm #17691, if your method works without causing seams then my PR becames redundant

aloucks:
> the `anisotropy` example does use tangents, this issue shows some issues caused by not optimal tangents #16179

The sphere looks much better with the gram-schmidt tangents. I actually left `Sphere` out of the tests that I added because there was at least one vertex where the mikktspace generated tangent differed from the gram-schmidt tangent enough that it was outside of the test threshold.

## mikktspace

https://github.com/user-attachments/assets/da7e29eb-8169-4696-b402-51c5cf435967

## gram-schmidt

https://github.com/user-attachments/assets/763b3bea-de11-4ff6-b6cf-dfc1df21e626

aloucks:
> Can the strategy names be changed to just the name of the algorithm? Mikktspace is very dependant on the UVs, so just calling it `HighQuality` might be misleading

Bikeshedding time! 🎨 

@JMS55 had suggested to inject `Calculation` into the name, but I think it's less necessary if we change `Strategy` to `Algorithm`.  The word "algorithm" implies the "calculation" part, imo. Also, do we want to keep HighQuality/Mikktspace the default?

@JMS55  / @hukasu  what are your thoughts?

```rust
pub enum TangentAlgorithm {
  Mikktspace,
  GramSchmidt,
}
```

aloucks:
@JMS55 Anymore feedback on the enum and variant naming? Do we want to leave mikktspace the default or change to gram-schmidt?

crates/bevy_mesh/src/gramschmidt.rs
```diff
@@ -0,0 +1,216 @@
...
+                    bi_tangents[i2] += bi_tangent;
+                    bi_tangents[i3] += bi_tangent;
+                }
+            }
+
+            let mut result_tangents = Vec::with_capacity(vertex_count);
+
+            for i in 0..vertex_count {
+                let normal = Vec3A::from_array(normals[i]);
+                let tangent = tangents[i].normalize();
```
IQuick143, line 108: Is there a numerical reason to normalize here?
The next line already involves a normalization and is linear in the `tangent` variable.

On the other hand, I don't see the normal getting normalised, and that would cause problems in the G-S algo.

aloucks: > Is there a numerical reason to normalize here?
> The next line already involves a normalization and is linear in the tangent variable.

The examples I've seen either normalize the accumulated tangent it or average it by dividing by the number of vertices that were used to accumulate it. I don't think the direction would be correct if we only normalized the final result and omitted the normalizing here.

> On the other hand, I don't see the normal getting normalised, and that would cause problems in the G-S algo.

This assumes that the vertex normals are normalized. Is that assumption false? If so, I'll need to normalize the normal here.

IQuick143: I'm not familiar with the code enough to determine whether the normals are normalised, although I would expect so.

Regarding the pre-normalisation of the tangent, that part is mathematically unnecessary as the projection is linear.
```
tangent = P(t).normalise()
```
where `P` is the projection operator `x -> x - x.dot(normal) * normal`, which is linear.

This tells us that if `t` is not normalised by a positive factor `k` we have:
```
P(k*t).normalise() = (k * P(t)).normalise() = P(t).normalise()
```
Where the first inequality comes from the linearity of P, and the second from normalisation destroying these scalar factors.

So from a purely theoretical standpoint the normalisation has no effect.

The only reason I can think of for the normalisation would be to avoid some sort of numerical error, although I cannot really think of the reasons it would be relevant here given how floating point values work.

aloucks: @IQuick143 

How about this? 

```rust
let normal = Vec3A::from_array(normals[i]);
// Assume that the normal has not been pre-normalized
let normal = normal.normalize();
// Although it is not mathematically necessary to normalize the accumulated
// tangent, doing so may provide better numerical stability in extreme cases.
let tangent = tangents[i].normalize();
// Gram-Schmidt orthogonalization
let tangent = (tangent - normal * normal.dot(tangent)).normalize();
```

IQuick143: You can save a few operations (a square root and some divides) by doing:
```rust
let tangent = (tangent - normal * (normal.dot(tangent) / normal.magnitude_squared())).normalize();
```
Instead of normalising the normal.

aloucks: > You can save a few operations (a square root and some divides) by doing:
> 
> ```rust
> let tangent = (tangent - normal * (normal.dot(tangent) / normal.magnitude_squared())).normalize();
> ```
> 
> Instead of normalising the normal.

Unfortunately that causes other issues where the normal is broken and has length of zero. The `cone` unit test fails. This is why I pushed a later commit to use `normalize_or_zero` instead of `normalize` when normalizing the normal.

aloucks:
@hukasu I renamed the enum to `TangentAlgorithm` and the associated variants. Any chance this can land before `0.16` is released?

hukasu:
i don't have authority to say when it can be merged

aloucks:
@alice-i-cecile Is this too late for 0.16?

alice-i-cecile:
Yeah, we're down to "essential bug fixes and polish" at this stage, sorry!

aloucks:
@alice-i-cecile Now that 0.16 is out, could this be considered for 0.17?

ChristopherBiscardi:
This is pretty reasonable, especially since it is adding a new option, but would need an update to merge as it has unfortunately been left off a few milestones now and has developed conflicts.

IceSentry:
The conflicts need to be resolved but once they are I'm happy with the solution to support multiple tangent algorithm. I don't know about the actual math though. I have no opinions about it.

PR #17990, opened 2025-02-23
impl `Eq`/`PartialEq` for `MeshMaterial{2|3}d`

jnhyatt:
# Objective

`Eq`/`PartialEq` are currently implemented for `MeshMaterial{2|3}d` only through the derive macro. Since we don't have perfect derive yet, the impls are only present for `M: Eq` and `M: PartialEq`. On the other hand, I want to be able to compare material components for my toy reactivity project.

## Solution

Switch to manual `Eq`/`PartialEq` impl.

## Testing

Boy I hope this didn't break anything!

### DIFF
--- a/crates/bevy_pbr/src/mesh_material.rs
+++ b/crates/bevy_pbr/src/mesh_material.rs
@@ -36,7 +36,7 @@ use derive_more::derive::From;
 ///     ));
 /// }
 /// ```
-#[derive(Component, Clone, Debug, Deref, DerefMut, Reflect, PartialEq, Eq, From)]
+#[derive(Component, Clone, Debug, Deref, DerefMut, Reflect, From)]
 #[reflect(Component, Default)]
 pub struct MeshMaterial3d<M: Material>(pub Handle<M>);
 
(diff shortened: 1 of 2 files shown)


alice-i-cecile (approved):

hukasu:
why exactly does the derive not works for `M` that are not `PartialEq` or `Eq`?

and what about adding `PartialEq` and `Eq` to the required traits of `Material2d` and `Material3d`?

```rust
pub trait Material2d:
    [Sized](https://doc.rust-lang.org/nightly/core/marker/trait.Sized.html)
    + [AsBindGroup](https://docs.rs/bevy/latest/bevy/render/render_resource/trait.AsBindGroup.html)
    + [Asset](https://docs.rs/bevy/latest/bevy/prelude/trait.Asset.html)
    + [Clone](https://doc.rust-lang.org/nightly/core/clone/trait.Clone.html)
    + PartialEq
    + Eq;
```

hukasu:
and for testing you could run `cargo run -p ci`

add `CARGO_BUILD_JOBS` and `RUST_TEST_THREADS` to the env to prevent OOM

alice-i-cecile:
> why exactly does the derive not works for M that are not PartialEq or Eq?

This is a standard Rust problem: "perfect derive" is the name of the solution for this. The derive macros assume that the internal types must also be compared, but when you're using things like `PhantomData` that isn't the case.

> and what about adding PartialEq and Eq to the required traits of Material2d and Material3d?

Generally, we should try to avoid adding bounds without a good reason, and prefer to add them at methods etc. It's less verbose to work with and more flexible.

jnhyatt:
@hukasu There shouldn't be any behavior changes. Tests run before merging anyway, that's why I didn't bother running tests locally.

Issue #17488, opened 2025-01-22
WakeUp doesn't unfreeze systems when dragging


chlobes:
## Bevy version  
0.15.1

## Relevant system information  
Windows 11

```ignore
`AdapterInfo { name: "NVIDIA GeForce RTX 2070", vendor: 4318, device: 7938, device_type: DiscreteGpu, backend: Vulkan }`
```

## What you did  
We stuff the `EventLoopProxy` into a static, so that we can wake the app remotely from a seperate thread

## What went wrong
The systems do not wake under some conditions, such as dragging the window, or winit's `drag_resize_window`.

## Additional information
Appears to work ok on Linux.
I am not entirely sure if this is a bevy issue, a winit issue, or a Win11 issue.
Originally reproduced for 14.2, updated to 15.1 and it is still possible to reproduce.

## Minimal Repro
```rust
/*
[dependencies]
bevy = "0.15.1"
bevy_winit = "0.15.1"
once_cell = "1.19.0"
*/

use std::sync::{Mutex, OnceLock};
use std::time::{Duration, Instant};

use bevy::{prelude::*, sprite::Sprite, window::RequestRedraw};
use bevy_winit::{EventLoopProxy, EventLoopProxyWrapper, UpdateMode, WakeUp, WinitSettings};
use once_cell::sync::Lazy;

static START: Lazy<Instant> = Lazy::new(Instant::now);

fn redraw(
    mut redraw: EventWriter<RequestRedraw>,
) {
    println!("redrawing {}", START.elapsed().subsec_millis());
    redraw.send(RequestRedraw);
}

//bobbing square so that we can see if the window is still responsive
#[derive(Component)]
struct Square;

fn spawn_square(mut commands: Commands) {
    // Spawn camera
    commands.spawn(Camera2d);
    commands.spawn((
        Sprite {
            color: Color::WHITE,
            custom_size: Some(Vec2::new(100.0, 100.0)),
            ..Default::default()
        },
        Square,
    ));
}

fn bob_square(mut query: Query<&mut Transform, With<Square>>) {
    if let Ok(mut transform) = query.get_single_mut() {
        let elapsed = START.elapsed().as_secs_f32();
        transform.translation.y = (elapsed * 1.5).sin() * 50.0;
    }
}

fn setup_wake_thread() {
    std::thread::spawn(move || {
        loop {
            EVENT_LOOP_PROXY.wake_event_loop();
            println!("waking {}", START.elapsed().subsec_millis());
[...]

PR #17991, opened 2025-02-23
Fixes missed RequestRedraw events in about_to_wait()

### LINKED ISSUE #16817: `RequestRedraw` doesn't work as expected
## Bevy version

0.15.0

```ignore
AdapterInfo { name: "NVIDIA GeForce RTX 4070 SUPER", vendor: 4318, device: 10115, device_type: DiscreteGpu, driver: "NVIDIA", driver_info: "565.90", backend: Vulkan }
```

## What you did

Here is a white button. When you press it, it becomes black with a transition in 0.2 s. It's a desktop app, so I need to send `RequestRedraw` manually. If the transition has not finished, it will call `redraw_request_events.send(RequestRedraw)`, so this system will still be called to redraw next frame.

```rust
fn update_background_color(
    mut buttons: Query<(&mut CustomButton, &mut BackgroundColor)>,
    mut redraw_request_events: EventWriter<RequestRedraw>,
) {
    for (mut button, mut background_color) in buttons.iter_mut() {
        if let Some(transition) = &button.transition {
            /// Unit: second
            const TRANSITION_DURATION: f32 = 0.2;
            let t = transition.start_time.elapsed().unwrap().as_secs_f32();

            info!("t = {t} s");
            if t > TRANSITION_DURATION {
                button.transition = None;
                info!("transition finished");
                continue;
            }
            let y = EasingCurve::new(0.0, 1.0, EaseFunction::SineIn)
                .sample(t / TRANSITION_DURATION)
                .unwrap();
            background_color.0 = transition.start.mix(&transition.end, y);
            redraw_request_events.send(RequestRedraw);
        }
    }
}
```

However, the output is: 
```
2024-12-14T09:16:26.746642Z  INFO bevy_demo: t = 0.0000055 s
2024-12-14T09:16:27.136166Z  INFO bevy_demo: t = 0.3895302 s
2024-12-14T09:16:27.136300Z  INFO bevy_demo: transition finished
```

It shows that nearly 400ms after `09:16:26.746642`, the next redraw is called.

<details>
<summary>Click me to expand the `main.rs`</summary>

```rust
use bevy::{prelude::*, window::RequestRedraw, winit::WinitSettings};
use std::time::SystemTime;

const DEFAULT_BUTTON_BACKGROUND_COLOR: Color = Color::WHITE;
const PRESSED_BUTTON_BACKGROUND_COLOR: Color = Color::BLACK;

#[derive(Component)]
#[require(
    Button,
    BackgroundColor(|| BackgroundColor(DEFAULT_BUTTON_BACKGROUND_COLOR)),
)]
pub struct CustomButton {
    transition: Option<Transition>,
}

struct Transition {
    start: Color,
    end: Color,
    start_time: std::time::SystemTime,
}

fn on_interaction(
    mut buttons: Query<(&mut CustomButton, &BackgroundColor, &Interaction), Changed<Interaction>>,
) {
    for (mut button, background_color, interaction) in buttons.iter_mut() {
        match interaction {
            Interaction::Pressed => {
                button.transition = Some(Transition {
                    start: background_color.0,
                    end: PRESSED_BUTTON_BACKGROUND_COLOR,
                    start_time: SystemTime::now(),
                });
            }
            Interaction::None => {
                button.transition = Some(Transition {
                    start: background_color.0,
                    end: DEFAULT_BUTTON_BACKGROUND_COLOR,
                    start_time: SystemTime::now(),
                });
            }
            _ => {}
        }
    }
}

fn update_background_color(
    mut buttons: Query<(&mut CustomButton, &mut BackgroundColor)>,
    mut redraw_request_events: EventWriter<RequestRedraw>,
) {
    for (mut button, mut background_color) in buttons.iter_mut() {
        if let Some(transition) = &button.transition {
            /// Unit: second
            const TRANSITION_DURATION: f32 = 0.2;
            let t = transition.start_time.elapsed().unwrap().as_secs_f32();

            info!("t = {t} s");
            if t > TRANSITION_DURATION {
                button.transition = None;
                info!("transition finished");
                continue;
            }
            let y = EasingCurve::new(0.0, 1.0, EaseFunction::SineIn)
                .sample(t / TRANSITION_DURATION)
                .unwrap();
            background_color.0 = transition.start.mix(&transition.end, y);
            redraw_request_events.send(RequestRedraw);
        }
    }
}

    commands.spawn(Camera2d);
    commands.spawn((
        CustomButton {
            transition: None,
        },
        Node {
            width: Val::Px(200.0),
            height: Val::Px(100.0),
            ..default()
        },
    ));
}

fn main() {
    App::new()
        .add_plugins(DefaultPlugins)
        .add_systems(Startup, setup)
        .add_systems(Update, (on_interaction, update_background_color))
        .run();
}

```
</details>

## What went wrong

It should be like this:

```
2024-12-14T09:16:13.352178Z  INFO bevy_demo: t = 0.0000044 s
2024-12-14T09:16:13.360099Z  INFO bevy_demo: t = 0.0079271 s
2024-12-14T09:16:13.360687Z  INFO bevy_demo: t = 0.0085149 s
2024-12-14T09:16:13.361251Z  INFO bevy_demo: t = 0.0090788 s
// ... 
2024-12-14T09:16:13.540491Z  INFO bevy_demo: t = 0.1883187 s
2024-12-14T09:16:13.544745Z  INFO bevy_demo: t = 0.19257231 s
2024-12-14T09:16:13.550594Z  INFO bevy_demo: t = 0.19842151 s
2024-12-14T09:16:13.556571Z  INFO bevy_demo: t = 0.2043975 s
2024-12-14T09:16:13.556679Z  INFO bevy_demo: transition finished
```

boondocklabs:
# Objective

Fixes #16817

## Solution

Currently `RequestRedraw` events can be missed in `about_to_wait()` as it is can call `run_app_update()` after reading the event cursor, thus the `redraw_requested` flag is stale.

This adds a subsequent read after the call to `run_app_update` to ensure `redraw_requested` is updated if any events were sent during the update.

## Testing

Tested in `Reactive` mode with animations using a forked `bevy_tweening` which supports 0.16-dev and sends `RequestRedraw` events while there are any `Animator` components active.

Prior to this change, the animation would only run a single update, and would stall for the reactive mode timeout before continuing the animation.

With this patch, the animation runs smooth in Reactive mode. Tested on MacOS and iOS on iPad and iPhone.


aloucks:
Does this change happen to fix https://github.com/bevyengine/bevy/issues/17488 ?

djeedai (commented):
I have no context for this change, I'd rather someone with winit knowledge to review @alice-i-cecile. Thanks

aloucks:
Do you still have the issue now that https://github.com/bevyengine/bevy/issues/17488 is merged?

boondocklabs:
> Do you still have the issue now that #17488 is merged?

No, the issue persists on the current main branch (64d57fad). I will create a small example that shows the behaviour.

boondocklabs:
I've created an example that reproduces this behaviour [here](https://github.com/boondocklabs/bevy-redraw-test).

The example uses `WinitSettings::desktop_app()` to enable reactive mode, spawns a cube mesh with mesh picking and an observer on `Pointer<Click>` that inserts or removes a marker component into the cube bundle. Left click enables animation, right click disables animation.

The `redraw` system on the Update schedule queries for the marker, and sends a `RequestRedraw` event when the marker is present.

The `update` system performs a basic rotation when `AnimationActive` marker is present.

Without this patch, clicking on the cube (being careful to not move the mouse) results in the animation not starting until either the reactive timeout occurs (5 second default from desktop_app()), or other interactions occur such as mouse movement.

With this patch, the animation runs as expected when the cube is clicked without delay.

@aloucks @alice-i-cecile

aloucks:
@boondocklabs 

- What platform are you on?

I wasn't able to reproduce your issue on the main branch. Clicking the cube started the animation with or without `commands.send_evet(RequestRedraw)` on line `47`. I was careful not to move the mouse.

I checked with Windows 11 on commits c14733d17717820d7cf4df7ec8250c0e7ac6e063 and 64d57fad083064ec73d3a6be013408df303065eb. 

I suspect that you're not on Windows, as the winit event handling is a little different. I'll be able to test on MacOS later next week.

In either case, with this change, it's now pulling from the event reader twice. Once before the the world update and once after the world update. Pulling the events from the reader _after_ the world update makes sense to me, but I don't think it should happen twice.

aloucks:
I was able to reproduce the issue on MacOS using your example. [Line 47](https://github.com/boondocklabs/bevy-redraw-test/blob/main/src/main.rs#L47) in your example didn't actually help for me though. 

The proposed fix works, but I think these lines should be removed so that it's only reading the `RequestRedraw` events once:
https://github.com/bevyengine/bevy/blob/906d78842077492a707a74e0d80af9878e5c6666/crates/bevy_winit/src/state.rs#L540-L544

boondocklabs:
From the looks of it, this is why it won't reproduce on Windows. Does reactive mode work at all? This appears to set `redraw_request` to true unconditionally. I can test on my windows machines when I have a minute.

https://github.com/bevyengine/bevy/blob/906d78842077492a707a74e0d80af9878e5c6666/crates/bevy_winit/src/state.rs#L452-L460


> I was able to reproduce the issue on MacOS using your example. [Line 47](https://github.com/boondocklabs/bevy-redraw-test/blob/main/src/main.rs#L47) in your example didn't actually help for me though.

Line 47 won't resolve the issue. I included that to show that it wasn't being caused by a frame delay/system ordering if the observer inserted the component after the system that checks for the marker and sends RedrawRequested.


> The proposed fix works, but I think these lines should be removed so that it's only reading the `RequestRedraw` events once:

I think you're right. I did consider that, but wasn't sure if that may have side effects as this patch calls `read()` conditionally on `should_update`.

I'll make this change, and do some testing on Mac/Linux/Windows.

aloucks:
> Does reactive mode work at all? 

Yes

aloucks:
> I was able to reproduce the issue on MacOS using your example. [Line 47](https://github.com/boondocklabs/bevy-redraw-test/blob/main/src/main.rs#L47) in your example didn't actually help for me though.
> 
> The proposed fix works, but I think these lines should be removed so that it's only reading the `RequestRedraw` events once:
> 
> https://github.com/bevyengine/bevy/blob/906d78842077492a707a74e0d80af9878e5c6666/crates/bevy_winit/src/state.rs#L540-L544

@boondocklabs Were you able to test your changes with the lines above removed on Linux?

aloucks:
@boondocklabs 

I rebased your branch, added your repro as a test-case example, and removed the extra read from the event reader.

https://github.com/bevyengine/bevy/compare/main...aloucks:bevy:fix-missed-request-redraw

I validated this works as expected on Window and MacOS. Note that the current main branch has buggy mesh picking and so clicking the cube right now sometimes doesn't fire the event. I opened a bug report for mesh picking here: #18495

boondocklabs:
@aloucks I'll try and get it tested on Linux today

PR #17994, opened 2025-02-23
Reimplement bindless storage buffers.

pcwalton:
Support for bindless storage buffers was temporarily removed with the bindless revamp. This commit restores that support.


crates/bevy_pbr/src/material_bind_groups.rs
```diff
@@ -699,7 +702,10 @@ where
                     bindless_buffer_descriptor.bindless_index,
                     render_device.create_buffer(&BufferDescriptor {
                         label: Some("bindless fallback buffer"),
-                        size: bindless_buffer_descriptor.size as u64,
+                        size: match bindless_buffer_descriptor.size {
+                            Some(size) => size as u64,
+                            None => DEFAULT_BINDLESS_FALLBACK_BUFFER_SIZE,
+                        },
```
superdump, line 708: What is the consequence of this?

pcwalton: We need some size for the fallback buffer, and we can't infer a size with material storage buffers because they're just untyped `Buffer`s (not `BufferVec<T>` or anything) in the material structure. So I just picked an arbitrary small size. I tested that it works even if the actual structures stored inside the buffer have different sizes.

pcwalton: I was considering allowing the developer to specify an explicit fallback size with `#[storage(10, fallback_size(32))]` or whatever, but I decided against supporting that because I didn't see any cases in which it'd be needed.

alice-i-cecile:
Added to the 0.16 milestone since this is a regression.

crates/bevy_pbr/src/material_bind_groups.rs
tychedelia: Might be worth just adding a comment that this was chosen with some testing but is arbitrary to the constant definition.

tychedelia (approved):
Tested on macOS.

crates/bevy_pbr/src/material_bind_groups.rs
pcwalton: Done.

crates/bevy_render/src/render_resource/bind_group.rs
```diff
@@ -196,15 +196,19 @@ impl Deref for BindGroup {
...
+///   a raw buffer, and the buffer will be bound as a storage buffer in shaders.
+///   In bindless mode, `binding_array()` argument that specifies the binding
+///   number of the resulting storage buffer binding array must be present.
+///
+/// | Arguments              | Values                                                                  | Default                |
+/// |------------------------|-------------------------------------------------------------------------|------------------------|
+/// | `visibility(...)`      | `all`, `none`, or a list-combination of `vertex`, `fragment`, `compute` | `vertex`, `fragment`   |
+/// | `read_only`            | if present then value is true, otherwise false                          | `false`                |
+/// | `buffer`               | if present then the field will be assumed to be a raw wgpu buffer       |                        |
+/// | `binding_array(...)`   | the binding number of the binding array, for bindless mode              | bindless mode disabled |
```
JMS55, line 211: I don't get what this means. Is the binding number like `@group(...) @binding(X) var<storage> binding_array<T>;` X here?

```suggestion
/// | `binding_array(...)`   | the binding number of the binding array, for bindless mode              | bindless mode disabled |
```

pcwalton: Yes, exactly. When you write something like `#[storage(10, binding_array(11))]`, the corresponding WGSL declaration looks like `@group(2) @binding(11) var<storage> binding_array<T>;`. In this case, the 10 means that the index *into* that binding array is given by the 10th *field* of the bindings structure (e.g. `StandardMaterialBindings`).

Thus there are two indices: the *bindless index*, which refers to a field of the structure that contains all the indices into the binding arrays for a single material, and the *binding number*, which refers to the actual `@binding` in the bind group layout.

IceSentry (approved):

PR #17996, opened 2025-02-23
Emphasize no structural changes in SystemParam::get_param

GlennFolker:
# Objective

Many systems like `Schedule` rely on the fact that every structural ECS changes are deferred until an exclusive system flushes the `World` itself. This gives us the benefits of being able to run systems in parallel without worrying about dangling references caused by memory (re)allocations, which will in turn lead to **Undefined Behavior**. However, this isn't explicitly documented in `SystemParam`; currently it only vaguely hints that in `init_state`, based on the fact that structural ECS changes require mutable access to the _whole_ `World`.

## Solution

Document this behavior explicitly in `SystemParam`'s type-level documentations.


alice-i-cecile (approved):

GlennFolker:
Should I document this behavior on `System::run_unsafe` and `SystemState::get_unchecked_manual` too? I just realized end-users of those might not stumble upon this documentation on `SystemParam`...

GlennFolker:
I've added the note to `WorldQuery` and `SystemState::get_unchecked_manual`... I'm now a little confused about what to do with `System::run_unsafe`. Requiring no structural ECS changes to `run_unsafe` would be a trait-level safety invariant, thus marking `System` an `unsafe trait`, but that'll probably be a breaking change and I'm not sure if that is in the scope of this PR.

GlennFolker:
Sorry for the overly long delay, I kinda forgot I made this PR at all 😅 Anyway, only `UnsafeWorldCell::world_mut` docs is changed now, which explicitly mentions `QueryParam` and `SystemState`. That should probably be enough to show the stop sign to users trying to invoke structural ECS changes there!

chescock (approved):

PR #18001, opened 2025-02-23
Update `Commands::entity` to return `Result`

JaySpruce:
## Objective

We upgraded commands to be able to handle Result, but decided to update Commands::entity in its own PR because adjusting the calls across the rest of the engine would make a big diff. This is that PR.

## Solution

- Changed `Commands::entity` to return `Result<EntityCommands, EntityDoesNotExistError>`.
- Deprecated `Commands::get_entity`.

## What to do with `get_entity`?

1. Don't check entity existence, let it return invalid `EntityCommands`
2. Check entity existence and panic (old `entity` behavior)
3. Check entity existence and do something else?
4. Deprecate `get_entity` (current draft)
5. Deprecate `get_entity`, add `entity_unchecked`

---

Alternative PR: #18043


alice-i-cecile (changes requested):
IMO we should just remove `get_entity` with this set of changes (maybe deprecate it first?). Its core purpose is already handled by returning a `Result` in `Commands::entity` in a much clearer and more resilient way.

alice-i-cecile:
CI is failing on doc tests, can you fix that up?

alice-i-cecile (approved):
This will be high impact and pushes users heavily towards the `?` pattern, but I think it's the right call. Runtime panics are a serious hazard in production Bevy code, and these sorts of very hard to predict panics when working with commands are extremely frustrating.

alice-i-cecile:
The other alternative here is that we can move `Commands::entity` to "Don't check entity existence, let it return invalid EntityCommands", and make `Commands::get_entity` return a `Result`. On reflection, I think I prefer that because of the lower breakage and boilerplate. Errors with entity commands are almost entirely due to the entity existing initially but being removed before the command is executed, which this PR won't help with.

JaySpruce:
I can make another PR for that since it'll be a lot smaller, then we can just pick one at will

crates/bevy_ecs/src/system/commands/mod.rs
```diff
@@ -423,42 +423,29 @@ impl<'w, 's> Commands<'w, 's> {
...
-    ///     commands.entity(entity)
+    ///     commands.entity(entity)?
     ///         // adds a new component bundle to the entity
     ///         .insert((Strength(1), Agility(2)))
     ///         // adds a single component to the entity
     ///         .insert(Label("hello world"));
+    ///     Ok(())
     /// }
     /// # bevy_ecs::system::assert_is_system(example_system);
     /// ```
```
aloucks, line 438: I'm confused by the objective of this PR. To me `commands.entity(entity)` is like indexing a slice when you know the index is valid. e.g.

```rust
for entity in query.iter() {
  commands.entity(entity)
}
```

> This will be high impact and pushes users heavily towards the `?` pattern, but I think it's the right call. Runtime panics are a serious hazard in production Bevy code, and these sorts of very hard to predict panics when working with commands are extremely frustrating.

Many of the changes here simply do an `unwrap`. The other changes to early return with `?` make it harder to find where a panic *should* have happened when the entity should have already existed.

alice-i-cecile:
Strong preference for #18043; closing this out.

PR #11245, opened 2024-01-07
mobile and webgpu: trigger redraw request when needed and improve window creation

mockersf:
# Objective

- Since #11227, Bevy doesn't work on mobile anymore. Windows are not created.

## Solution

- Create initial window on mobile after the initial `Resume` event. macOS is included because it's excluded from the other initial window creation and I didn't want it to feel alone. Also, it makes sense. this is needed for Android
https://github.com/bevyengine/bevy/blob/cfcb6885e3b475a93ec0fe7e88023ac0f354bbbf/crates/bevy_winit/src/lib.rs#L152
- request redraw during plugin initialisation (needed for WebGPU)
- request redraw when receiving `AboutToWait` instead of at the end of the event handler. request to redraw during a `RedrawRequested` event are ignored on iOS

PR #18004, opened 2025-02-23
Fix window freezing when dragged or resized on Windows

### LINKED ISSUE #17488: WakeUp doesn't unfreeze systems when dragging

aloucks:
# Objective

Fixes #17488

## Solution

The world update logic happened in the the `about_to_wait` winit window callback, but this is is not correct as (1) the winit documentation states that the callback should not be used for that purpose and (2) the callback is not fired when the window is resized or being dragged. However, that callback was used in #11245 to fix an iOS bug (which caused the regression). The solution presented here is a workaround until the event loop code can be re-written. 

## Testing

I confirmed that the `eased_motion` example continued to be animated when dragging or resizing the window. 

https://github.com/user-attachments/assets/ffaf0abf-4cd7-479b-83e9-e1850aaf3513

### DIFF
--- a/crates/bevy_winit/src/state.rs
+++ b/crates/bevy_winit/src/state.rs
@@ -431,6 +431,13 @@ impl<T: Event> ApplicationHandler<T> for WinitAppRunnerState<T> {
             }
             WindowEvent::RedrawRequested => {
                 self.ran_update_since_last_redraw = false;
+
+                // https://github.com/bevyengine/bevy/issues/17488
+                #[cfg(target_os = "windows")]
+                {
+                    self.redraw_requested = true;
+                    self.redraw_requested(_event_loop);
+                }
             }
             _ => {}
         }
@@ -468,6 +475,27 @@ impl<T: Event> ApplicationHandler<T> for WinitAppRunnerState<T> {
         create_windows(event_loop, create_window.get_mut(self.world_mut()));
         create_window.apply(self.world_mut());
 
+        // TODO: This is a workaround for https://github.com/bevyengine/bevy/issues/17488
+        //       while preserving the iOS fix in https://github.com/bevyengine/bevy/pull/11245
+        //       The monitor sync logic likely belongs in monitor event handlers and not here.
+        #[cfg(not(target_os = "windows"))]
+        self.redraw_requested(event_loop);
+    }
+
+    fn suspended(&mut self, _event_loop: &ActiveEventLoop) {
+        // Mark the state as `WillSuspend`. This will let the schedule run one last time
+        // before actually suspending to let the application react
+        self.lifecycle = AppLifecycle::WillSuspend;
+    }
+
+    fn exiting(&mut self, _event_loop: &ActiveEventLoop) {
+        let world = self.world_mut();
+        world.clear_all();
+    }
+}
+
+impl<T: Event> WinitAppRunnerState<T> {
+    fn redraw_requested(&mut self, event_loop: &ActiveEventLoop) {
         let mut redraw_event_reader = EventCursor::<RequestRedraw>::default();
 
         let mut focused_windows_state: SystemState<(Res<WinitSettings>, Query<(Entity, &Window)>)> =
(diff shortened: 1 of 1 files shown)


IceSentry (approved):
I confirm this works on my windows 11 machine too.

mockersf (approved):
I don't have a windows to test it works, but the fix looks like what I expected

aevyrie:
As discussed with @alice-i-cecile it would be nice to get this fix in a patch.

aevyrie (approved):

PR #18009, opened 2025-02-24
`TaskPool`: Prefer task completion over executing new tasks

aevyrie:
# Objective

- Systems that use the task pool, either explicitly or implicitly using parallel queries, will often end up executing tasks from different systems.
- This can cause random tasks to block the main or render schedule at random, adding frame variance and increasing frame times when CPU bound.
- This profile is a common occurrence on `main`. `propagate_parent_transforms` takes more than twice as long as it should, blocking the main schedule for that time, because it uses `task pool.scope`, which has decided to execute tasks from the render schedule on the main schedule.
image

## Solution

- In task pool scope execution, prefer to check if the current task is complete instead of ticking the executor to find new work.

## Testing

- Ran the scene viewer with tracy to look for images like the one in the objective section.
- Things look much, much better, and I could not find any occurrences: 
image
image

### DIFF
--- a/crates/bevy_tasks/src/task_pool.rs
+++ b/crates/bevy_tasks/src/task_pool.rs
@@ -478,7 +478,7 @@ impl TaskPool {
                     .is_ok();
             }
         };
-        execute_forever.or(get_results).await
+        get_results.or(execute_forever).await
     }
 
     #[inline]
@@ -497,7 +497,7 @@ impl TaskPool {
                 let _result = AssertUnwindSafe(tick_forever).catch_unwind().await.is_ok();
             }
         };
-        execute_forever.or(get_results).await
+        get_results.or(execute_forever).await
     }
 
     #[inline]
@@ -519,7 +519,7 @@ impl TaskPool {
                     .is_ok();
             }
         };
-        execute_forever.or(get_results).await
+        get_results.or(execute_forever).await
     }
 
     #[inline]
@@ -537,7 +537,7 @@ impl TaskPool {
                 let _result = AssertUnwindSafe(tick_forever).catch_unwind().await.is_ok();
             }
         };
-        execute_forever.or(get_results).await
+        get_results.or(execute_forever).await
     }
 
     /// Spawns a static future onto the thread pool. The returned [`Task`] is a


hymm:
So the change in this pr changes the order of events from:
1. Run scope function
1. Poll for tasks
2. Check if all the tasks are done.
3. Sleep thread if task done check is false.

To:
1. Run scope function
1. Check if all tasks are done.
2. Poll for tasks
3. Sleep thread if task done check is false.

So on main branch, we avoid a sleep if the "poll for tasks" completes all the tasks before the check happens. And in this pr we avoid a sleep and polling tasks if all the tasks are done by the time the scope function is done running. So depending on the workload this could be slower or faster.

In my case this pr ends up pretty neutral on `many_foxes`.  `propagate_parent_transforms` ends up a little faster, but some other schedules end up slower. I imagine that on other workloads it'll end up being better. As shown in the above testing. My testing in the past didn't show too much of a difference between the two either, but I went with the current as doing work and then checking for the work completion made more sense at the time. But with the testing that aevyrie has done, we can see it does help with the new transform propagation algorithm. I think this change is worth it, since the existing uses in the schedule and query par iter are mostly agnostic to the change.

hymm (approved):

chescock:
> So on main branch, we avoid a sleep if the "poll for tasks" completes all the tasks before the check happens. And in this pr we avoid a sleep and polling tasks if all the tasks are done by the time the scope function is done running. So depending on the workload this could be slower or faster.

Just to check my own understanding: Isn't the issue that 2-4 happen in a loop?  So, on main, if there are *any* tasks available in the task pool, we run those instead of checking if the scope is complete.  Which means the scope can't complete until the task pool is completely out of tasks.

aevyrie:
> if there are any tasks available in the task pool, we run those instead of checking if the scope is complete. Which means the scope can't complete until the task pool is completely out of tasks.

This is my understanding of the issue as well. I don't understand why that behavior is desired at all, or how we prevent it. It seems as if any uses of the task pool for *internal* (intra-system) parallelism end up being treated the same as *external* (inter-system) parallelism. Or, as you said, a system running a parallel query can end up picking up the task of running a whole other system  (or multiple systems) inside the scope of the first system. This seems like an outright bug.

chescock:
> This PR doesn't fully resolve this issue, it just seems to reduce the frequency. I have to wonder if we should be using `or` at all here? Ideally if you are in the scope of a system, you shouldn't be picking up the work of the scheduler.

It's still necessary because the tasks spawned to the scope just go in the general FIFO pool of work.  We do want the thread running the scope to be able to run its *own* tasks.  Otherwise we could get stuck where all threads have started a scope and pushed tasks for it, and there are no threads left to run them!  I think that's the deadlock that @hymm mentioned on Discord.  

If we can switch to a task pool with LIFO local queues, then we could implement something where the scope runs its own tasks but sleeps instead of stealing.  Something like that could prevent the issue completely.  

> This seems like an outright bug.

I agree!  Whenever I read this code I assumed it worked the way it does now.  It seemed more likely that `or` had confusing backwards semantics than that this worked the way it did :).

Issue #18010, opened 2025-02-24
Need some way to nest labeled assets


pcwalton:
## Bevy version

0.16-dev after https://github.com/bevyengine/bevy/pull/15481

## Overview

If you have an asset that loads nested glTFs, for example my `bevy-glxf-loader`, then if you have more than one glTF in that asset the names like `Scene0`, `Mesh0`, etc. will collide. What I'd like is to be able to namespace those loaded assets with a prefix of some kind.

andriyDev:
I took a stab at fixing this but looks like it's not so trivial. Or rather, there's a deeper bug lurking. Turns out `add_loaded_labeled_asset` is just kinda broken. When you do an immediate load of an asset, it assigns the path of the asset you are nested-loading. So if you are loading asset "a.txt", and within you load "b.ron", subassets of the second asset will have a path of "b.ron#subasset1", rather than "a.txt#subasset1". This means the path "a.txt#subasset1" perpetually is in the `NotLoaded` state.

I think we need to propagate the path into these subassets. We treat the "root" subasset of "b.ron" as a regular subasset of "a.txt", so we need to somehow set the path of its subassets to be "a.txt".

andriyDev:
Ok I've figured out more of the problem. Here's a diagram I came up with:

```
root loader tries to load "root#nested/subasset"
v
root -> nested -> subasset
          ^
nested loader finds handle for "nested#subasset" (or generates it) and stores it in nested

Once root calls add_loaded_labeled_asset, it registers all "nested" subassets as subassets of root - but now with the path "root#nested/subasset" (which is not what is stored in nested)
```

In essence, the handles we store in the "nested" asset (in the nested loader) correspond to a different path than "root". That means when we look up the handles of assets in "nested", those assets will never get loaded, since they were loaded as subassets of "root".

I don't understand yet how #15481 broke this. It seems to me like this problem should have been just the same in that situation...

andriyDev:
Now I think I understand the full scope of the issue. Nested asset loading is weird...

(Also to be clear, this is all addressing that "deeper bug" I mentioned above).

The problem is that before #15481, `LoadedAsset` had two functions:

1. It was the type you get from `LoadContext::finish` for a subasset in the current loader.
2. It was the type you get from `NestedLoader<StaticTyped, Immediate>::load`.

This means `LoadedAsset::labeled_assets` could store two kinds of assets:

1. Subassets of subassets.
2. Subassets of a dependency that you did a nested load for.

[...]

PR #18013, opened 2025-02-24
Make adding a subasset label return a result for if there is a duplicate label.

andriyDev:
# Objective

- Makes #18010 more easily debuggable. This doesn't solve that issue, but protects us from it in the future.

## Solution

- Make `LoadContext::add_labeled_asset` and friends return an error if it finds a duplicate asset.

## Testing

- Added a test - it fails before the fix.

---

## Migration Guide

- `AssetLoader`s must now handle the case of a duplicate subasset label when using `LoadContext::add_labeled_asset` and its variants. If you know your subasset labels are unique by construction (e.g., they include an index number), you can simply unwrap this result.


pcwalton (approved):
LGTM. Of course we should have some kind of solution for the problem, but this will at least prevent others from suffering the same difficulty debugging the issue.

alice-i-cecile (approved):

PR #18016, opened 2025-02-24
Automatically close weekly ci issue if weekly ci succeeds

Brezak:
# Objective

It's not always obvious when the issue created by the weekly CI run is safe to close.

## Solution

Automatically close it when a weekly CI run succeeds so it doesn't get forgotten.

## Testing

Hope.

### DIFF
--- a/.github/workflows/weekly.yml
+++ b/.github/workflows/weekly.yml
@@ -12,6 +12,7 @@ env:
   CARGO_INCREMENTAL: 0
   CARGO_PROFILE_TEST_DEBUG: 0
   CARGO_PROFILE_DEV_DEBUG: 0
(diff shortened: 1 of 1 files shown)


alice-i-cecile (approved):
I think it might make sense to merge this logic into the `open-issue` workflow below and just branch inside of it, but I like this well enough as is.

BD103 (changes requested):
Hi! Glad to see you back in action (though maybe you have been for a while, I haven't been reviewing too many PRs recently).

I definitely like this PR, I just caught one issue with the `$TITLE` variable!

BD103 (approved):
LGTM :)

PR #18017, opened 2025-02-24
allow `Call` and `Closure` expressions in hook macro attributes

RobWalt:
# Objective

This PR adds:

- function call hook attributes `#[component(on_add = func(42))]`
  - main feature of this commit
- closure hook attributes `#[component(on_add = |w, ctx| { /* ... */ })]`
  - maybe too verbose
  - but was easy to add
  - was suggested on discord

This allows to reuse common functionality without replicating a lot of boilerplate. A small example is a hook which just adds different default sprites. The sprite loading code would be the same for every component. Unfortunately we can't use the required components feature, since we need at least an `AssetServer` or other `Resource`s or `Component`s to load the sprite.

```rs
fn load_sprite(path: &str) -> impl Fn(DeferredWorld, HookContext) {
  |mut world, ctx| {
    // ... use world to load sprite
  }
}

#[derive(Component)]
#[component(on_add = load_sprite("knight.png"))]
struct Knight;

#[derive(Component)]
#[component(on_add = load_sprite("monster.png"))]
struct Monster;
```

---

The commit also reorders the logic of the derive macro a bit. It's probably a bit less lazy now, but the functionality shouldn't be performance critical and is executed at compile time anyways.

## Solution

- Introduce `HookKind` enum in the component proc macro module
- extend parsing to allow more cases of expressions

## Testing

I have some code laying around. I'm not sure where to put it yet though. Also is there a way to check compilation failures? Anyways, here it is:

```rs
use bevy::prelude::*;

#[derive(Component)]
#[component(
    on_add = fooing_and_baring,
    on_insert = fooing_and_baring,
    on_replace = fooing_and_baring,
    on_despawn = fooing_and_baring,
    on_remove = fooing_and_baring
)]
pub struct FooPath;

fn fooing_and_baring(
    world: bevy::ecs::world::DeferredWorld,
    ctx: bevy::ecs::component::HookContext,
) {
}

#[derive(Component)]
#[component(
    on_add = baring_and_bazzing("foo"),
    on_insert = baring_and_bazzing("foo"),
    on_replace = baring_and_bazzing("foo"),
    on_despawn = baring_and_bazzing("foo"),
    on_remove = baring_and_bazzing("foo")
)]
pub struct FooCall;

fn baring_and_bazzing(
    path: &str,
) -> impl Fn(bevy::ecs::world::DeferredWorld, bevy::ecs::component::HookContext) {
    |world, ctx| {}
}

#[derive(Component)]
#[component(
    on_add = |w,ctx| {},
    on_insert = |w,ctx| {},
    on_replace = |w,ctx| {},
    on_despawn = |w,ctx| {},
    on_remove = |w,ctx| {}
)]
pub struct FooClosure;

#[derive(Component, Debug)]
#[relationship(relationship_target = FooTargets)]
#[component(
    on_add = baring_and_bazzing("foo"),
    // on_insert = baring_and_bazzing("foo"),
    // on_replace = baring_and_bazzing("foo"),
    on_despawn = baring_and_bazzing("foo"),
    on_remove = baring_and_bazzing("foo")
)]
pub struct FooTargetOf(Entity);

#[derive(Component, Debug)]
#[relationship_target(relationship = FooTargetOf)]
#[component(
    on_add = |w,ctx| {},
    on_insert = |w,ctx| {},
    // on_replace = |w,ctx| {},
    // on_despawn = |w,ctx| {},
    on_remove = |w,ctx| {}
)]
pub struct FooTargets(Vec<Entity>);

// MSG:  mismatched types  expected fn pointer `for<'w> fn(bevy::bevy_ecs::world::DeferredWorld<'w>, bevy::bevy_ecs::component::HookContext)`    found struct `Bar`
//
// pub struct Bar;
// #[derive(Component)]
// #[component(
//     on_add = Bar,
// )]
// pub struct FooWrongPath;

// MSG: this function takes 1 argument but 2 arguements were supplied
//
// #[derive(Component)]
// #[component(
//     on_add = wrong_bazzing("foo"),
// )]
// pub struct FooWrongCall;
//
// fn wrong_bazzing(path: &str) -> impl Fn(bevy::ecs::world::DeferredWorld) {
//     |world| {}
// }

// MSG: expected 1 argument, found 2
//
// #[derive(Component)]
// #[component(
//     on_add = |w| {},
// )]
// pub struct FooWrongCall;
```

---

## Showcase

I'll try to continue to work on this to have a small section in the release notes.

### DIFF
--- a/crates/bevy_ecs/compile_fail/tests/ui/component_hook_relationship.rs
+++ b/crates/bevy_ecs/compile_fail/tests/ui/component_hook_relationship.rs
@@ -1,0 +1,27 @@
+use bevy_ecs::prelude::*;
+
+#[derive(Component, Debug)]
+#[relationship(relationship_target = FooTargets)]
+//~^ ERROR: Custom on_insert hooks are not supported as Relationships already define an on_insert hook
+#[component(on_insert = foo_hook)]
+pub struct FooTargetOf(Entity);
+
+#[derive(Component, Debug)]
+#[relationship_target(relationship = FooTargetOf)]
+//~^ ERROR: Custom on_replace hooks are not supported as RelationshipTarget already defines an on_replace hook
+#[component(on_replace = foo_hook)]
+pub struct FooTargets(Vec<Entity>);
+
+#[derive(Component, Debug)]
+#[relationship(relationship_target = BarTargets)]
+//~^ ERROR: Custom on_replace hooks are not supported as Relationships already define an on_replace hook
+#[component(on_replace = foo_hook)]
+pub struct BarTargetOf(Entity);
+
+#[derive(Component, Debug)]
+#[relationship_target(relationship = BarTargetOf)]
+//~^ ERROR: Custom on_despawn hooks are not supported as this RelationshipTarget already defines an on_despawn hook, via the 'linked_spawn' attribute
+#[component(on_despawn = foo_hook)]
+pub struct BarTargets(Vec<Entity>);
+
+fn foo_hook(world: bevy_ecs::world::DeferredWorld, ctx: bevy_ecs::component::HookContext) {}
--- a/crates/bevy_ecs/compile_fail/tests/ui/component_hook_relationship.rs
+++ b/crates/bevy_ecs/compile_fail/tests/ui/component_hook_relationship.rs
@@ -1,27 +1,63 @@
 use bevy_ecs::prelude::*;
 
-#[derive(Component, Debug)]
-#[relationship(relationship_target = FooTargets)]
-//~^ ERROR: Custom on_insert hooks are not supported as Relationships already define an on_insert hook
-#[component(on_insert = foo_hook)]
-pub struct FooTargetOf(Entity);
+mod case1 {
+    use super::*;
 
-#[derive(Component, Debug)]
-#[relationship_target(relationship = FooTargetOf)]
-//~^ ERROR: Custom on_replace hooks are not supported as RelationshipTarget already defines an on_replace hook
-#[component(on_replace = foo_hook)]
-pub struct FooTargets(Vec<Entity>);
+    #[derive(Component, Debug)]
+    #[component(on_insert = foo_hook)]
+    //~^ ERROR: Custom on_insert hooks are not supported as relationships already define an on_insert hook
+    #[relationship(relationship_target = FooTargets)]
+    pub struct FooTargetOfFail(Entity);
 
-#[derive(Component, Debug)]
-#[relationship(relationship_target = BarTargets)]
-//~^ ERROR: Custom on_replace hooks are not supported as Relationships already define an on_replace hook
-#[component(on_replace = foo_hook)]
-pub struct BarTargetOf(Entity);
+    #[derive(Component, Debug)]
+    #[relationship_target(relationship = FooTargetOfFail)]
+    //~^ E0277
+    pub struct FooTargets(Vec<Entity>);
+}
 
-#[derive(Component, Debug)]
-#[relationship_target(relationship = BarTargetOf)]
-//~^ ERROR: Custom on_despawn hooks are not supported as this RelationshipTarget already defines an on_despawn hook, via the 'linked_spawn' attribute
-#[component(on_despawn = foo_hook)]
-pub struct BarTargets(Vec<Entity>);
+mod case2 {
+    use super::*;
 
-fn foo_hook(world: bevy_ecs::world::DeferredWorld, ctx: bevy_ecs::component::HookContext) {}
+    #[derive(Component, Debug)]
+    #[component(on_replace = foo_hook)]
+    //~^ ERROR: Custom on_replace hooks are not supported as RelationshipTarget already defines an on_replace hook
+    #[relationship_target(relationship = FooTargetOf)]
+    pub struct FooTargetsFail(Vec<Entity>);
+
+    #[derive(Component, Debug)]
+    #[relationship(relationship_target = FooTargetsFail)]
+    //~^ E0277
+    pub struct FooTargetOf(Entity);
+}
+
+mod case3 {
+    use super::*;
+
+    #[derive(Component, Debug)]
+    #[component(on_replace = foo_hook)]
+    //~^ ERROR: Custom on_replace hooks are not supported as Relationships already define an on_replace hook
+    #[relationship(relationship_target = BarTargets)]
+    pub struct BarTargetOfFail(Entity);
+
+    #[derive(Component, Debug)]
+    #[relationship_target(relationship = BarTargetOfFail)]
+    //~^ E0277
+    pub struct BarTargets(Vec<Entity>);
+}
+
+mod case4 {
+    use super::*;
+
+    #[derive(Component, Debug)]
+    #[component(on_despawn = foo_hook)]
+    //~^ ERROR: Custom on_despawn hooks are not supported as this RelationshipTarget already defines an on_despawn hook, via the 'linked_spawn' attribute
+    #[relationship_target(relationship = BarTargetOf, linked_spawn)]
+    pub struct BarTargetsFail(Vec<Entity>);
+
+    #[derive(Component, Debug)]
+    #[relationship(relationship_target = BarTargetsFail)]
+    //~^ E0277
+    pub struct BarTargetOf(Entity);
+}
+
+fn foo_hook(_world: bevy_ecs::world::DeferredWorld, _ctx: bevy_ecs::component::HookContext) {}
--- a/crates/bevy_ecs/macros/src/component.rs
+++ b/crates/bevy_ecs/macros/src/component.rs
@@ -473,7 +473,6 @@ impl HookAttributeKind {
                 [
                     "Not supported in this position, please use one of the following:",
                     "- path to function",
-                    "- closure",
                     "- call to function yielding closure",
                 ]
                 .join("\n"),
--- a/crates/bevy_ecs/macros/src/component.rs
+++ b/crates/bevy_ecs/macros/src/component.rs
@@ -9,8 +9,8 @@ use syn::{
     punctuated::Punctuated,
     spanned::Spanned,
     token::{Comma, Paren},
-    Data, DataStruct, DeriveInput, ExprClosure, ExprPath, Field, Fields, Ident, Index, LitStr,
-    Member, Path, Result, Token, Type, Visibility,
+    Data, DataStruct, DeriveInput, Expr, ExprCall, ExprClosure, ExprPath, Field, Fields, Ident,
+    Index, LitStr, Member, Path, Result, Token, Type, Visibility,
 };
 
 pub fn derive_event(input: TokenStream) -> TokenStream {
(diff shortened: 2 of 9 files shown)


crates/bevy_ecs/macros/src/component.rs
```diff
@@ -507,19 +572,44 @@ fn parse_component_attr(ast: &DeriveInput) -> Result<Attrs> {
...
+                            .value()?
+                            .parse::<Expr>()
+                            .and_then(HookKind::from_expr)?,
+                    );
                     Ok(())
                 } else if nested.path.is_ident(ON_REPLACE) {
-                    attrs.on_replace = Some(nested.value()?.parse::<ExprPath>()?);
+                    attrs.on_replace = Some(
+                        nested
+                            .value()?
```
Bleachfuel, line 593: It would be better to implement syn::Parse for Hookkind, then you don't have to do and_then, just parse::<HookKind>

crates/bevy_ecs/macros/src/component.rs
```diff
@@ -435,14 +470,44 @@ pub const ON_DESPAWN: &str = "on_despawn";
...
+                    "- path to function",
+                    "- closure",
+                    "- call to function yielding closure",
+                ]
+                .join("\n"),
+            )),
+        }
+    }
+}
+
```
Bleachfuel, line 502:
```suggestion

impl HookKind {
    fn from_expr(value: Expr) -> Result<Self> {
        match value {
            Expr::Path(path) => Ok(HookKind::Path(path)),
            Expr::Call(call) => Ok(HookKind::Call(call)),
            Expr::Closure(closure) => Ok(HookKind::Closure(closure)),
            _ => Err(syn::Error::new(
                value.span(),
                [
                    "Not supported in this position, please use one of the following:",
                    "- path to function",
                    "- closure",
                    "- call to function yielding closure",
                ]
                .join("\n"),
            )),
        }
    }
}

impl Parse for HookKind {
    fn parse(input: syn::parse::ParseStream) -> Result<Self> {
        input.parse::<Expr>().and_then(Self::from_expr)
    }
}

```

RobWalt: addressed ✅

RobWalt:
@Bleachfuel Thanks a lot for the review, I addressed your comments :)

crates/bevy_ecs/macros/src/component.rs
```diff
@@ -75,25 +75,14 @@ pub fn derive_component(input: TokenStream) -> TokenStream {
...
+    if relationship.is_some() {
         if attrs.on_insert.is_some() {
             return syn::Error::new(
-                ast.span(),
-                "Custom on_insert hooks are not supported as relationships already define an on_insert hook",
-            )
-            .into_compile_error()
-            .into();
+                    ast.span(),
+                    "Custom on_insert hooks are not supported as relationships already define an on_insert hook",
```
alice-i-cecile, line 82: Future work: we can probably automatically concatenate these function calls 🤔

alice-i-cecile (changes requested):
This needs doc tests, both to ensure that this works and teach users that it does. Code looks great, and the feature is 🔥 though.

alice-i-cecile:
@cart, this steps on the toes of your `Construct` plans a bit, so I wanted to pull in your review to help you form a broader picture. Regardless of what you decide to do there, I think that this is a really nice intuitive addition to hooks.

crates/bevy_ecs/macros/src/component.rs
```diff
@@ -507,19 +573,19 @@ fn parse_component_attr(ast: &DeriveInput) -> Result<Attrs> {
                     };
                     Ok(())
                 } else if nested.path.is_ident(ON_ADD) {
-                    attrs.on_add = Some(nested.value()?.parse::<ExprPath>()?);
+                    attrs.on_add = Some(nested.value()?.parse::<HookAttributeKind>()?);
                     Ok(())
                 } else if nested.path.is_ident(ON_INSERT) {
-                    attrs.on_insert = Some(nested.value()?.parse::<ExprPath>()?);
+                    attrs.on_insert = Some(nested.value()?.parse::<HookAttributeKind>()?);
                     Ok(())
```
Bleachfuel, line 580: Is it not nicer to match on the ident with `get_ident` instead of all these else ifs.

Bleachfuel:
``` rust
let ident = nested.path.get_ident().ok_or_else(|| {
    syn::Error::new_spanned(nested, "Expected an identifier")
})?;

match ident {
    ON_ADD => {
        attrs.on_add = Some(nested.value()?.parse::<HookAttributeKind>()?);
        Ok(())
    }
    ON_INSERT => {
        attrs.on_insert = Some(nested.value()?.parse::<HookAttributeKind>()?);
        Ok(())
    }
    _ => Err(syn::Error::new_spanned(ident, "Unexpected attribute")),
}
```

RobWalt: addressed, unfortunately you can't match variables on constants, so this still includes an if

crates/bevy_ecs/src/component.rs
```diff
@@ -320,6 +320,37 @@ pub use bevy_ecs_macros::require;
 /// }
 /// ```
 ///
+/// This doesn't only support static functions but also
```
alice-i-cecile, line 323:
```suggestion
/// This  also supports
```

alice-i-cecile (approved):
Great! I have a small wording suggestion but those doc tests should be sufficient to validate that this keeps working.

I think you could probably be a bit more clear in your compile fail tests about *why* the code shouldn't compile but I won't block on it.

crates/bevy_ecs/macros/src/component.rs
Bleachfuel: oh that's unfortunate :(

crates/bevy_ecs/compile_fail/tests/ui/component_hook_relationship.rs
```diff
@@ -0,0 +1,33 @@
...
+#[component(
+    on_insert = |w,ctx| {},
+)]
+pub struct FooTargetOf(Entity);
+
+#[derive(Component, Debug)]
+#[relationship_target(relationship = FooTargetOf)]
+//~^ ERROR: Custom on_replace hooks are not supported as RelationshipTarget already defines an on_replace hook
+#[component(
+    on_replace = |w,ctx| {},
```
cart, line 15: I'm not sure encouraging inline function impls here is doing developers any favors. This (largely) breaks autocomplete in Rust analyzer, it breaks syntax highlighting, it breaks inline type hints. I'm assuming this would also break debuggers. Theres also just a general "weirdness factor" here. The UX improvement of being able to define it inline is fully offset by those UX costs in my book.  And once that code is written, it incurs a constant maintenance cost when others are trying to read / change the code.

crates/bevy_ecs/compile_fail/tests/ui/component_hook_closure_type_mismatch.rs
```diff
@@ -0,0 +1,23 @@
+use bevy_ecs::prelude::*;
+
+// this should fail since the function is required to have the signature
+// (DeferredWorld, HookContext) -> ()
+#[derive(Component)]
+//~^ E0057
+#[component(
+    on_add = wrong_bazzing("foo"),
```
cart, line 8: Do we have any real-world scenarios for this use case? I kind of like the simplicity of requiring an explicit function pointer, and the "pass arbitrary static data into a function" seems pretty limited in practice, to the point that I'm struggling to envision use cases here.

crates/bevy_ecs/compile_fail/tests/ui/component_hook_relationship.rs
RobWalt: Yeah I fully agree. I mainly implemented this since it was suggested when I asked over on discord. I wasn't sure myself if this feels right the whole time and probably should've made my own concern a little more clear.

I'll gladly remove this kind of functionality from the PR (we can always revert, if others really need it)

crates/bevy_ecs/compile_fail/tests/ui/component_hook_closure_type_mismatch.rs
RobWalt: I guess my main use case is outlined in the PR description. I was just quickly scaffolding a 2D game based on sprites and wanted to give entities some default sprites on spawn. I wanted to use `asset_server` loading, so the `require` feature didn't really provide what I needed there. The next logical step was using hooks which works, but now I have to replicate the whole function for every kind of thing that get's its default sprite `on_add` which felt really boilerplate-y.

There are probably some other solutions to the problem:

- pre-allocating handle ids or something (dunno how, but I think this is probably supported)
- maybe some kind of extra macro which hides the boilerplate and generates a static function

Not sure, maybe this feature is actually superfluous.

Other use cases might be adding audio assets `on_add` (or other hooks). I was also imagining that in a follow up PR we could support composing reusable hooks. Just let me phantasize something of the shape:

```rust
#[component(on_add = (
  add_sprite("monster.png"),
  add_audio("grumbling.png"),
  // ...
))]
```

cart: Ah yeah that makes sense. That is totally my fault for not reading the whole description. I now see why @alice-i-cecile called out how this overlaps with my [Construct](https://github.com/bevyengine/bevy/discussions/14437) plans a bit. And I do see the value now. This feels like a (very) limited solution to a bigger problem, but the scope is small enough that I'm down to include it. I don't see much of a downside, other than the fact that people will use it in these scenarios and pretty quickly hit the limitations.

> we could support composing reusable hooks.

I think I would probably draw the line here, as trying to build these apis pretty quickly exposes the limitations of such a system. Ex: if someone wants to set the sprite color too, how would they do that? We could add additional function variants to include parameters, but then suddenly we're adding a bunch of new "on_add hook spawn APIs", which isn't really a world I want to live in. I'd prefer to leave that door closed (at least, in upstream bevy) given that BSN will be here soon and given the tradeoffs of this approach (archetype moves, limited flexibility for inputs, erasure of the actual component types being inserted, requires a brand new set of APIs potentially with many variants, etc).

crates/bevy_ecs/compile_fail/tests/ui/component_hook_relationship.rs
cart: Cool yeah lets remove this.

alice-i-cecile:
@RobWalt I've removed the 0.16 milestone from this since I don't think it's super pressing, but this work is free to proceed. Do bother folks for more reviews once you've removed the in-line function syntax.

crates/bevy_ecs/compile_fail/tests/ui/component_hook_relationship.rs
RobWalt: Done, removed ✅

cart (approved):

alice-i-cecile:
Thank you to everyone involved with the authoring or reviewing of this PR! This work is relatively important and needs release notes! Head over to https://github.com/bevyengine/bevy-website/issues/1999 if you'd like to help out.

PR #18023, opened 2025-02-24
Make asset watcher work when path contains "../"

### LINKED ISSUE #18022: Asset watcher fails to update assets when path contains "../"
## Bevy version
Version 0.15.2

## What you did

My bevy app is a folder in a workspace, and the asset folder is in the workspace root. I initialize the `AssetPlugin` like this:
```rust
app.add_plugins(AssetPlugin {
    file_path: "../assets".to_string(),
    ..default()
});
```

Then while the app is running, I altered an image in the asset directory.
## What went wrong

Whenever an asset updates, a panic occurs: (edited)
```
thread 'notify-rs debouncer loop' panicked at <home-folder>/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/bevy_asset-0.15.2/src/io/file/file_watcher.rs:51:9:
FileWatcher::get_asset_path() failed to strip prefix from absolute path: absolute_path="<path-to-workspace>/my-app/../assets/gui/buttons.png", root="<path-to-workspace>/assets"
```
In a project with a different folder layout, the asset updates instead.

## Additional information

I cloned bevy and changed the `FileEventHandler::get_path` function to this:
```
    fn get_path(&self, absolute_path: &Path) -> Option<(PathBuf, bool)> {
        let absolute_path = absolute_path.canonicalize().unwrap();
        Some(get_asset_path(&self.root, &absolute_path))
    }
```
This is on line 264 of `bevy_assets/src/io/file/file_watcher.rs`. This fixed the issue

I also tested several of bevy's examples with the file_watcher feature and it worked as expected in all cases, so I believe this change is a complete solution to the issue.

Threadzless:
# Objective

Fixes #18022

## Solution

Canonicalize asset paths

## Testing

I ran the examples `sprite`, `desk_toy` and `game_menu` with the feature `file_watcher` enabled. All correctly updated an asset when the source file was altered.


alice-i-cecile (approved):

andriyDev (approved):

mcobzarenco:
I was wondering if this patch allows one to escape assets/ and access arbitrary paths by writing ../../.. etc.

viridia:
Different threat model. Use of std::fs implies adversarial Rust code, which is not something we can prevent. However, adversarial file paths could be introduced by a mod or script, not controlled by the original author.

Threadzless:
Its plausible a mod/script could hijack the AssetServer to load any file on the system. I'll look into it tomorrow, and open a new issue if I find a way to do it.

alice-i-cecile:
@viridia @andriyDev @cart, do y'all have opinions on whether this should be supported? I'm a little leery about the threat model raised above, but I also don't like the current fairly silent breakage.

andriyDev:
My ideal would be to allow "../" in paths for asset sources but not for asset paths. So you should be able to create an asset source like `AssetSourceBuilder::platform_default("../my_shared_assets", None)`, But that wouldn't let you do `asset_server.load("../freedom_muahaha.txt")`.

We can probably just achieve this by canonicalizing the asset path when building the source for that though.

viridia:
As far as I know, the Bevy project hasn't taken an official stance on security matters; if you read the page on the Bevy website that talks about the core principles of the engine, security isn't mentioned as a project goal.

However, security is very much a concern of the Rust language, and given that part of the reason for choosing Bevy is to choose Rust, I think both we and our users would sleep better at night if we did try to limit potential vulnerabilities in the code. While many game developers may not care about this issue, there are some who will (especially in the context of multiplayer).

When we talk about securing relative paths, we have to first ask what those paths are relative to. An asset root should be a sandbox from which relative paths cannot escape. However, not all paths are relative to asset roots (such as the location of the user preferences directory) and thus won't be sandboxed.

For mods and scripts, we would want to restrict embedded asset paths to only be relative to some asset root.

cart:
I don't think this would be the end of the world. But I'll also note that the asset system at large (and especially the asset processor) requires resolving paths to a canonical absolute path, to ensure everything is updated appropriately when an asset changes. A relative path that escapes the mounted root does not really have a canonical absolute path

I do think that treating an asset root as "inescapable" is nice from both a security standpoint and an addressability standpoint and this is probably the path forward. I agree that being able to define asset sources with a relative path should be valid.

mockersf:
It's already possible to escape, and it works well 👍 
This PR only fixes it when using file watching to hot reload something with "../"

cart:
> It's already possible to escape, and it works well 👍

Yup I'm aware. I also think that should be reigned in, but thats definitely not something to be resolved here. Given that it is my inclination to be stricter about asset paths, I was content with keeping it locked down here. But I do see the argument that we should make the existing state of affairs work consistently throughout Bevy. I won't stand in the way of merging here. But at some point in the future I'll probably reopen this conversation with a more global scope.

alice-i-cecile:
I think we should make the "can you escape the assets folder" a more global setting: some applications (and some uses within those appliocations) will want this behavior, but others will not. Since this particular PR fixes a subset of valid behavior (looking up within the assets root folder), without opening up a *new* hole, I think we should go ahead and merge this as a bug fix and keep talking about the big picture in the linked issue.

PR #18026, opened 2025-02-25
Load each glTF skin at most once.

pcwalton:
Currently, we reload a glTF skin each time we encounter a node that references it. By checking for duplicates, PR #18013 turned this into a fatal error. But this was always wasteful. This commit fixes the issue by caching each skin by its index as we load it.

The Maya babylon.js export plugin likes to emit glTFs with multiple nodes that reference the same skin, so this effectively unbreaks Maya rigs.

### DIFF
--- a/crates/bevy_gltf/src/loader.rs
+++ b/crates/bevy_gltf/src/loader.rs
@@ -801,7 +801,7 @@ async fn load_gltf<'a, 'b, 'c>(
 
     let mut nodes = HashMap::<usize, Handle<GltfNode>>::default();
     let mut named_nodes = <HashMap<_, _>>::default();
-    let mut skins = vec![];
+    let mut skins = <HashMap<_, _>>::default();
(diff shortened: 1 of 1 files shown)


crates/bevy_gltf/src/loader.rs
```diff
@@ -801,7 +801,7 @@ async fn load_gltf<'a, 'b, 'c>(
 
     let mut nodes = HashMap::<usize, Handle<GltfNode>>::default();
     let mut named_nodes = <HashMap<_, _>>::default();
-    let mut skins = vec![];
+    let mut skins = <HashMap<_, _>>::default();
```
andriyDev, line 804: This syntax is crazy. We can't just do `HashMap::default()` here?

andriyDev (approved):

alice-i-cecile (approved):

PR #18028, opened 2025-02-25
Use register_dynamic for merging

ElliottjPierce:
# Objective

I found a bug while working on #17871. When required components are registered, ones that are more specific (smaller inheritance depth) are preferred to others. So, if a ComponentA is already required, but it is registered as required again, it will be updated if and only if the new requirement has a smaller inheritance depth (is more specific).  However, this logic was not reflected in merging `RequriedComponents`s together. Hence, for complicated requirement trees, the wrong initializer could be used.

## Solution

Re-write merging to work by extending the collection via `require_dynamic` instead of blindly combining the inner storage.

## Testing

I created this test to ensure this bug doesn't creep back in. This test fails on main, but passes on this branch.

```rs
    #[test]
    fn required_components_inheritance_depth_bias() {
        #[derive(Component, PartialEq, Eq, Clone, Copy, Debug)]
        struct MyRequired(bool);

        #[derive(Component, Default)]
        #[require(MyRequired(|| MyRequired(false)))]
        struct MiddleMan;

        #[derive(Component, Default)]
        #[require(MiddleMan)]
        struct ConflictingRequire;

        #[derive(Component, Default)]
        #[require(MyRequired(|| MyRequired(true)))]
        struct MyComponent;

        let mut world = World::new();
        let order_a = world
            .spawn((ConflictingRequire, MyComponent))
            .get::<MyRequired>()
            .cloned();
        let order_b = world
            .spawn((MyComponent, ConflictingRequire))
            .get::<MyRequired>()
            .cloned();

        assert_eq!(order_a, Some(MyRequired(true)));
        assert_eq!(order_b, Some(MyRequired(true)));
    }
```
Note that when the inheritance depth is 0 (Like if there were no middle man above), the order of the components in the bundle still matters.

## Migration Guide

`RequiredComponents::register_dynamic` has been changed to `RequiredComponents::register_dynamic_with`.

Old:
```rust
required_components.register_dynamic(
      component_id,
      component_constructor.clone(),
      requirement_inheritance_depth,
);
```

New:
```rust
required_components.register_dynamic_with(
      component_id,
      requirement_inheritance_depth,
      || component_constructor.clone(),
);
```

This can prevent unnecessary cloning.


alice-i-cecile:
No migration guide for bug fixes in general :)

crates/bevy_ecs/src/component.rs
```diff
@@ -2131,8 +2131,21 @@ impl RequiredComponents {
...
+        for (
+            component_id,
+            RequiredComponent {
+                constructor,
+                inheritance_depth,
+            },
+        ) in required_components
+            .0
+            .iter()
+            .map(|(id, req)| (*id, req.clone()))
```
alice-i-cecile, line 2143: I'd prefer to avoid cloning here, but I don't see a way to do that without making register_dynamic's signature worse.

alice-i-cecile (approved):
Good test: that clearly demonstrated the problem to me.

crates/bevy_ecs/src/component.rs
ElliottjPierce: I agree. I will need to extend functionality here a tiny bit to work better with `RequiredComponentsMut` from #17871. I think I can avoid cloning on that branch, but over here, I was just focused on fixing the bug. If preferred, I can go ahead and add that extra functionality here, but the extra complexity won't really be used until #17871.

alice-i-cecile: Yeah, don't worry about it for now.

ElliottjPierce:
@alice-i-cecile The additional cloning during merging was annoying us. I was going to have to fix it anyway for #17871, but since we aren't going in that direction anymore, I went ahead and fixed it here. I added `register_dynamic_with` and changed to using that. That saved I think 4 cloning paths, which we now only do if the registration actually goes through. 

I kept `register_dynamic` even though it's not used anymore. It's public, and I didn't want this to be a breaking change. Happy to remove it though, if you say the word.

alice-i-cecile (changes requested):
Let's axe `register_dynamic` (just have a little migration guide showing the snippet you currently have). It really doesn't seem to add much value here. Once that's done, this LGTM.

alice-i-cecile (approved):

crates/bevy_ecs/src/component.rs
```diff
@@ -2037,67 +2042,69 @@ impl RequiredComponents {
...
         // `component_id` matches the type initialized by the `erased` constructor above.
         // `erased` initializes a component for `component_id` in such a way that
         // matches the storage type of the component. It only uses the given `table_row` or `Entity` to
         // initialize the storage corresponding to the given entity.
-        unsafe { self.register_dynamic(component_id, erased, inheritance_depth) };
+        unsafe { self.register_dynamic_with(component_id, inheritance_depth, erased) };
     }
 
-    /// Iterates the ids of all required components. This includes recursive required components.
+    /// Iterates the ids of all required components' ids. This includes recursive required components.
```
Jondolf, line 2107: Seems to be a typo?
```suggestion
    /// Iterates the ids of all required components. This includes recursive required components.
```

cart (approved):
This makes sense to me. Nice catch!

crates/bevy_ecs/src/component.rs
```diff
@@ -2115,8 +2122,20 @@ impl RequiredComponents {
     // Merges `required_components` into this collection. This only inserts a required component
     // if it _did not already exist_.
     pub(crate) fn merge(&mut self, required_components: &RequiredComponents) {
```
Jondolf, line 2124: Nit: This comment could now be updated to clarify that if a required component *does* already exist, its constructor may still get updated *if* the new requirement is more specific (meaning that the inheritance depth is smaller).

Also this should probably be a doc comment (`///`) instead of a regular comment.

Jondolf (approved):
Changes make sense, looks good!

PR #18031, opened 2025-02-25
Light Textures

robtfm:
# Objective

add support for light textures (also known as light cookies, light functions, and light projectors)

image

## Solution

- add components:

```rs
/// Add to a [`PointLight`] to add a light texture effect.
/// A texture mask is applied to the light source to modulate its intensity,  
/// simulating patterns like window shadows, gobo/cookie effects, or soft falloffs.
pub struct PointLightTexture {
    /// The texture image. Only the R channel is read.
    pub image: Handle<Image>,
    /// The cubemap layout. The image should be a packed cubemap in one of the formats described by the [`CubemapLayout`] enum.
    pub cubemap_layout: CubemapLayout,
}

/// Add to a [`SpotLight`] to add a light texture effect.
/// A texture mask is applied to the light source to modulate its intensity,  
/// simulating patterns like window shadows, gobo/cookie effects, or soft falloffs.
pub struct SpotLightTexture {
    /// The texture image. Only the R channel is read.
    /// Note the border of the image should be entirely black to avoid leaking light.
    pub image: Handle<Image>,
}

/// Add to a [`DirectionalLight`] to add a light texture effect.
/// A texture mask is applied to the light source to modulate its intensity,  
/// simulating patterns like window shadows, gobo/cookie effects, or soft falloffs.
pub struct DirectionalLightTexture {
    /// The texture image. Only the R channel is read.
    pub image: Handle<Image>,
    /// Whether to tile the image infinitely, or use only a single tile centered at the light's translation
    pub tiled: bool,
}
```

- store images to the `RenderClusteredDecals` buffer
- read the image and modulate the lights
- add `light_textures` example to showcase the new features

## Testing

see light_textures example


crates/bevy_pbr/src/decal/clustered.rs
```diff
@@ -97,6 +98,67 @@ pub struct ClusteredDecal {
     pub tag: u32,
 }
 
+/// Cubemap layout defines the order of images in a packed cubemap image.
+#[derive(Default, Reflect, Debug, Clone, Copy)]
+pub enum CubemapLayout {
```
robtfm, line 103: i added support for 4 standard cubemap layouts. unity supports these plus some others, which we could add in future. 

i didn't add cubemap array image support, as i wanted to reuse the existing decal texture array for simplicity.

it's maybe not ideal to have this in the shader, but preprocessing would be significantly harder.

pcwalton:
Let me know if you'd like me to review. This is exciting stuff :)

One thing I was thinking is that it might be better to have a separate example for light cookies instead of stuffing it into decals. You can just copy and paste a lot of the decal code if you like.

robtfm:
> Let me know if you'd like me to review. This is exciting stuff :)
> 

yes please! it's quite straightforward because it's using all your clustered decal infrastructure, so you are definitely the key reviewer.

> One thing I was thinking is that it might be better to have a separate example for light cookies instead of stuffing it into decals. You can just copy and paste a lot of the decal code if you like.

i set all the light cookie stuff in the demo to off by default so it's not so overwhelming as it looks in the screenshot, but i am happy to do that if you think it'd be better.

robtfm:
also pls note i need to replace the spotlight and directional light images with something uncopyrighted (the point light cookie was created by me so that one's ok)

alice-i-cecile (dismissed):
We should do a better job explaining what a light cookie is to users. I know it's a common and simple rendering term, but for many users this will be their first introduction to the concept.

I think that a sentence or so on the Component types will probably do the trick, but it might be worth saying more in the example too.

crates/bevy_pbr/src/render/pbr_lighting.wgsl
```diff
@@ -539,8 +622,26 @@ fn point_light(
...
+    var cookie_sample = 1f;
+
+#ifdef CLUSTERED_DECALS_ARE_USABLE
+    if enable_cookie && (*light).decal_index != 0xFFFFFFFFu {
+        let relative_position = (view_bindings::clustered_decals.decals[(*light).decal_index].local_from_world * vec4(P, 1.0)).xyz;
+        let cubemap_type = view_bindings::clustered_decals.decals[(*light).decal_index].tag;
+        let decal_uv = cubemap_uv(relative_position, cubemap_type);
+        let image_index = view_bindings::clustered_decals.decals[(*light).decal_index].image_index;
+
+        cookie_sample = textureSampleLevel(
```
robtfm, line 634: i used both textureSample and textureSampleLevel(0) in different places, i've switched them all to textureSampleLevel(0).

i think textureSample would be ok, we could use mips if available, but i understand unreal uses mips to blend from harder to softer shadows over distance, if we do something similar we would want to choose the level explicitly here.

alice-i-cecile (commented):
Good docs, thanks :) Revoking my "request changes" review, but I don't have the expertise to properly approve this.

fjkorf (approved):
This code is clear and direct. The comments are explanatory and helpful. I was able to run the clustered_decal example locally and it looks great. It panicked on my mac as expected. This is a really cool PR.

pcwalton:
I think I'd prefer to separate out the example, though I won't block the PR on this.

crates/bevy_pbr/src/decal/clustered.rs
pcwalton: I'm not a fan of this TBH -- I'd rather just use cubemap arrays -- but it's OK for now, as long as we put the code behind an `#ifdef` in the shader to avoid regressing our VGPR usage.

crates/bevy_pbr/src/decal/clustered.rs
```diff
@@ -97,6 +98,72 @@ pub struct ClusteredDecal {
...
+/// simulating patterns like window shadows, gobo effects, or soft falloffs.
+#[derive(Clone, Component, Debug, Reflect)]
+#[reflect(Component, Debug)]
+#[require(DirectionalLight)]
+pub struct DirectionalLightCookie {
+    /// The cookie image. Only the R channel is read.
+    pub image: Handle<Image>,
+    /// Whether to tile the image infinitely, or use only a single tile centered at the light's translation
+    pub tiled: bool,
+}
```
pcwalton, line 165: I'm kind of surprised that directional light cookies exist given that I think of directional lights as infinitely large, but cool that you added this!

crates/bevy_pbr/src/render/pbr_lighting.wgsl
```diff
@@ -625,5 +746,31 @@ fn directional_light(
...
+                0.0
+            ).r;                    
+        } else {
+            cookie_sample = 0f;
+        }
+    }
+#endif
+
+    return color * (*light).color.rgb * cookie_sample;
 }
```
pcwalton, line 776: Let's put light cookie support behind an ifdef which is only activated when light cookies are present. As it is, this is probably going to bloat our VGPR usage considerably, which is much too high already.

crates/bevy_pbr/src/render/pbr_lighting.wgsl
```diff
@@ -440,10 +440,93 @@ fn perceptualRoughnessToRoughness(perceptualRoughness: f32) -> f32 {
...
+
+    switch face_index {
+        case X_PLUS:  { face_uv = vec2<f32>(direction.z, -direction.y) / direction.x; }
+        case X_MINUS: { face_uv = vec2<f32>(-direction.z, -direction.y) / -direction.x; }
+        case Y_PLUS:  { face_uv = vec2<f32>(direction.x,  -direction.z) / direction.y; }
+        case Y_MINUS: { face_uv = vec2<f32>(direction.x, direction.z) / -direction.y; }
+        case Z_PLUS:  { face_uv = vec2<f32>(direction.x, direction.y) / direction.z; }
+        case Z_MINUS: { face_uv = vec2<f32>(direction.x, -direction.y) / -direction.z; }
+        default: {}
+    }
```
pcwalton, line 482: Could avoid some divergence if you put the numerator and denominator in separate variables, and then have that funnel down to a single division operation outside the switch.

crates/bevy_pbr/src/render/pbr_lighting.wgsl
```diff
@@ -440,10 +440,93 @@ fn perceptualRoughnessToRoughness(perceptualRoughness: f32) -> f32 {
...
+        case CUBEMAP_TYPE_CROSS_HORIZONTAL: { 
+            face_size = vec2(1.0/4.0, 1.0/3.0); 
+            switch face_index {
+                case X_PLUS:  { corner_uv = vec2(2, 1); }
+                case X_MINUS: { corner_uv = vec2(0, 1); }
+                case Y_PLUS:  { corner_uv = vec2(1, 0); }
+                case Y_MINUS: { corner_uv = vec2(1, 2); }
+                case Z_PLUS:  { corner_uv = vec2(3, 1); }
+                case Z_MINUS: { corner_uv = vec2(1, 1); }
+                default: {}
```
pcwalton, line 508: Could you factor this out to reduce shader code duplication? You could do `if cubemap_type == CUBEMAP_TYPE_CROSS_VERTICAL || cubemap_type == CUBEMAP_TYPE_CROSS_HORIZONTAL { switch face_index { ... } }` above the switch. You could also replace the entire switch with an index into a constant array, avoiding all the divergence.

crates/bevy_pbr/src/render/pbr_lighting.wgsl
pcwalton: I agree with using `textureSampleLevel(0)` for now, though we probably will want to select a mip level at some point.

crates/bevy_pbr/src/decal/clustered.rs
```diff
@@ -97,6 +98,72 @@ pub struct ClusteredDecal {
...
+    ///   +x
+    ///   -y
+    ///   +y
+    ///   -y
+    ///   -z
+    ///   +z
+    SequenceVertical = 2,
+    /// layout in a horizontal sequence
+    /// +x -y +y -y -z +z
+    SequenceHorizontal = 3,
```
pcwalton, line 126: Might want to indent your +x, -y, +y, etc. lists or surround them with \`\`\`text so that they'll render well in rustdoc.

pcwalton (commented):
This is fine and a needed feature, thanks for implementing it :) I do think that we should probably switch to cubemap arrays at some point; the hardware just makes everything so much easier. And the `#ifdef` thing may be important (cc @Elabajaba who has been measuring VGPR usage).

crates/bevy_pbr/src/decal/clustered.rs
```diff
@@ -97,6 +98,72 @@ pub struct ClusteredDecal {
...
+    pub cubemap_layout: CubemapLayout,
+}
+
+/// Add to a [`SpotLight`] to add a light cookie effect.
+/// A light cookie is a texture mask applied to a light source to modulate its intensity,  
+/// simulating patterns like window shadows, gobo effects, or soft falloffs.
+#[derive(Clone, Component, Debug, Reflect)]
+#[reflect(Component, Debug)]
+#[require(SpotLight)]
+pub struct SpotLightCookie {
```
cart, line 148: 3D packages don't agree on the term for this concept:

* Unreal: light functions
* Unity: light cookie
* Godot: light projector
* Blender: setting a texture as the light color

Of all of these terms, I find "light cookie" the least helpful. I like "light texture" or "light projector". "light function" would be too generic for the feature as implemented.

alice-i-cecile: light texture is the clearest to me as a rendering noob. We should doc alias the synonyms of course.

robtfm: It is called a cookie in theatre too apparently, from [cucoloris](https://en.m.wikipedia.org/wiki/Cucoloris). Light texture seems unambiguous though.

crates/bevy_pbr/src/decal/clustered.rs
robtfm: There’s an option to tile or use a single centered instance of the texture; this was basically copied from the unity api

crates/bevy_pbr/src/render/pbr_lighting.wgsl
robtfm: Last I knew array indexing had to be by constants, but I’ll double check

pcwalton: Oh yes, you're right. But you could do something like `corner_uv = (vec2(0x111102, 0x132011) >> (4 * face_index)) & 0xf`. Up to you, I know that's ugly, but it'd be fast :)

crates/bevy_pbr/src/render/pbr_lighting.wgsl
robtfm: i added a feature flag (`pbr_light_textures`) as doing it any other way was difficult. i also added `pbr_clustered_decals` at the same time (i guess vgpr is similar).

crates/bevy_pbr/src/render/pbr_lighting.wgsl
robtfm: done ... not totally convinced it helps since i don't know how good gpus are at scoping (we have a longer-lived divisor variable, vs mutually exclusive switch-scoped variables).

crates/bevy_pbr/src/render/pbr_lighting.wgsl
robtfm: that's horrifying. done.

crates/bevy_pbr/src/decal/clustered.rs
robtfm: done

crates/bevy_pbr/src/decal/clustered.rs
robtfm: renamed to `PointLightTexture` etc

robtfm:
i separated the example out into `light_textures`

crates/bevy_pbr/src/decal/clustered.rs
cart: Good call. Thanks!

fjkorf:
I pulled and ran the latest code without issue. I agree that Light Textures is more clear than Light Cookies. Also, I appreciate the added feature flags.

crates/bevy_pbr/src/decal/clustered.rs
Novakasa: I highly appreciate this addition, since it will probably make cloud shadows possible! Hopefully in the future this could be made to somehow work with a shader instead of just a static texture to add dynamism

robtfm: It should work already - you just need a camera rendering to the texture you use on the light.

Novakasa: That's true, but I guess I was imagining something like what this [unity tutorial](https://youtu.be/JLXLUA_DIQc?si=dh7SVTYZ0c9Un4lu&t=497) does (timestamp relevant) where you can just put a full on material into the "light cookie" slot. But now that I think about this I am not sure how this would be done in bevy. Maybe Unity internally also just renders the material to the texture to facilitate this.
[On discord](https://discord.com/channels/691052431525675048/866787577687310356/1349784636829733049), @NthTensor  implied that with mode modular shader logic in bevy in the future, this could be possible, which would be exciting.

IceSentry (approved):
Code looks good, would be nice if we didn't need to add more assets to make the example look good, but I don't really have any suggestions on what to do instead.

Assuming conflicts are fixed LGTM

alice-i-cecile:
Ran into this in my merge train today, but merge conflicts aren't fixed yet. Ping me when it's resolved!

robtfm:
@alice-i-cecile hope I caught the next train!

alice-i-cecile:
Just in time!

Issue #15704, opened 2024-10-07
Deprecate and remove insert_or_spawn_batch


alice-i-cecile:
Just like in #15459 / #15652 for `get_or_spawn`, `insert_or_spawn_batch` is an anti-pattern: it uses entity IDs in a non-opaque way, attempting to assign specific meaning to certain entities. With the retained rendering world in place, we should remove it to discourage this pattern and enforce the ECS invariants more strongly.

#15702 should be merged first, to ensure that the migration path for the non-footgunny part of this API is smooth.

alice-i-cecile:
https://github.com/bevyengine/bevy/issues/18054 showcases a serious performance footgun involved with these methods.

dlight:
The [deprecation notice](https://docs.rs/bevy/latest/bevy/prelude/struct.Commands.html#method.insert_or_spawn_batch) could also point to this issue (maybe with a markdown link so people can click the issue number and come to the Github issue).

JaySpruce:
Completed in #18148.

PR #18035, opened 2025-02-25
Replace internal uses of `insert_or_spawn_batch`

JaySpruce:
## Objective
`insert_or_spawn_batch` is due to be deprecated eventually (#15704), and removing uses internally will make that easier.

## Solution

Replaced internal uses of `insert_or_spawn_batch` with `try_insert_batch` (non-panicking variant because `insert_or_spawn_batch` didn't panic).

All of the internal uses are in rendering code. Since retained rendering was meant to get rid non-opaque entity IDs, I assume the code was just using `insert_or_spawn_batch` because `insert_batch` didn't exist and not because it actually wanted to spawn something. However, I am *not* confident in my ability to judge rendering code.


alice-i-cecile (changes requested):
Can you split apart the "replace internal usages" from the "deprecate"? The former is uncontroversial, but we're definitely going to get users screaming if we deprecate this without a clear answer on what to do for "allocate this specific entity ID".

JaySpruce:
I had thought not supporting it was the intention, do you know what the official way to do that should look like?

alice-i-cecile:
> I had thought not supporting it was the intention, do you know what the official way to do that should look like?

It is! But I've gotten serious pushback from users before when attempting to just remove it: there's clearly a disconnect there. I'm not sure what the right path forward is there, but we can/should definitely stop using this internally for now.

alice-i-cecile:
Adding to the 0.16 milestone; since this is showing up in real games as a bizarre performance issue I'd like to get the internals moved over ASAP.

ElliottjPierce:
Running through some examples on this PR compared to main:

- parenting example gives error: `ERROR bevy_core_pipeline::core_3d::main_opaque_pass_3d_node: Error encountered while rendering the opaque phase InvalidViewQuery` This is not present on main. The example still works, so this may be superficial but is worth investegating.
- alien_cake_addict example has exactly the same issue on launch.
- mesh_picking example has exactly the same issue on launch.
- testbed_3d example has exactly the same issue on many of the test scenes.
- loading_screen example has the same error when transitioning from scene 1 to scene 1, and gives the additinal error `ERROR bevy_core_pipeline::core_3d::main_transparent_pass_3d_node: Error encountered while rendering the transparent phase InvalidViewQuery` when going from scene 2 to scene 2. Going from scene 1 to 2 or visa versa works fine. It is functional aside from the error message.

These errors are not present on the main branch. I am really unexperienced in all of the rendering side of things, so I don't think I'm the right person to track down the cause here. But I'm happy to test again if anyone has ideas.

I did not notice any visual differences compared to main, and I didn't see entities disappear unexpectedly.

I only spot checked examples that looked suspect to me, so there may be errors I'm missing. 

If you think it's worth it, I can add an additional example that just spawns and despawns stuff a bunch to see if that triggers any red flags.

JaySpruce:
> These errors are not present on the main branch.

I get those errors on main; I bisected it to 7f1458149 (#16782, merged on Feb 24), and it turns out there's already an issue for it: #18094

ElliottjPierce:
> I get those errors on main

It's strange that I didn't but ok. In that case, I think we proceed. I couldn't find anything else fishy. @alice-i-cecile your thoughts when you get the chance?

alice-i-cecile (approved):
I'm comfortable with this for now: let's get this in and see if other weird rendering bugs arise.

PR #18038, opened 2025-02-25
Add methods to add single entity relationships

### LINKED ISSUE #17572: Method for adding a single entity relationship

When attempting to use relationships there is no method to allow for adding a single relationship to an entity.

A method similar to the one used by children, perhaps `add_relationship`? 

Using the current `add_related` method, though it feels a bit clunky for just wanting to add one entity.

Brezak:
# Objective

Closes #17572

## Solution

Add the `add_one_related` methods to `EntityCommands` and `EntityWorldMut`.

## Testing

Clippy

---

## Showcase

The `EntityWorldMut` and `FilteredResourcesMut` now include the `add_one_related` method if you just want to relate 2 entities.

### DIFF
--- a/crates/bevy_ecs/src/relationship/related_methods.rs
+++ b/crates/bevy_ecs/src/relationship/related_methods.rs
@@ -21,7 +21,9 @@ impl<'w> EntityWorldMut<'w> {
         self
     }
 
-    /// Relates the given entities to this entity with the relation `R`
+    /// Relates the given entities to this entity with the relation `R`.
+    ///
+    /// See [`add_one_related`](Self::add_one_related) if you want relate only one entity.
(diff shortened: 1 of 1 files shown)


crates/bevy_ecs/src/relationship/related_methods.rs
```diff
@@ -32,6 +32,11 @@ impl<'w> EntityWorldMut<'w> {
         self
     }
 
+    /// Relates the given entity to this with the relation `R`
```
alice-i-cecile, line 35: These methods should be crosslinked with the add_related methods.

alice-i-cecile (approved):
I don't like the `add_related` name in the first place, but this method should definitely exist :)

viridia (approved):

PR #18042, opened 2025-02-26
moved Debug from derive to impl_ptr in bevy_ptr

### LINKED ISSUE #17988: Change #[derive(Debug)] to manual impl for certain bevy_ptr types

`#[derive(Debug)]` places incorrect bounds on its debug implementation in certain cases involving generics. In particular, it requires that all generic parameters implement Debug.

This means that the Debug implementations for `Ptr`, `PtrMut` and `OwningPtr` do not function because they require `A: Debug`, and as neither `Aligned` nor `Unaligned` implement Debug, it is impossible to satisfy the trait bounds.

The outcome of this is that the following code:
```rust
#[derive(Debug)]
struct PtrWrapper<'a> {
    ptr: bevy_ptr::Ptr<'a>,
}
```
results in the following error
```
error[E0277]: `Aligned` doesn't implement `Debug`
 --> src/lib.rs:3:5
  |
1 | #[derive(Debug)]
  |          ----- in this derive macro expansion
2 | struct PtrWrapper<'a> {
3 |     ptr: bevy_ptr::Ptr<'a>,
  |     ^^^^^^^^^^^^^^^^^^^^^^ `Aligned` cannot be formatted using `{:?}` because it doesn't implement `Debug`
  |
  = help: the trait `Debug` is not implemented for `Aligned`
  = help: the trait `Debug` is implemented for `Ptr<'_, A>`
  = note: this error originates in the derive macro `Debug` (in Nightly builds, run with -Z macro-backtrace for more info)

For more information about this error, try `rustc --explain E0277`.
error: could not compile `playground` (lib) due to 1 previous error

```

Implement Debug for the affected types manually by deferring to the Debug implementation on `NonNull`:

```rust
impl core::fmt::Debug for Ptr<'_> {
    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> Result<(), core::fmt::Error> {
        core::fmt::Debug::fmt(self.0, f)
    }
}
```

- Users of the library can work around the broken implementation, by manually implementing Debug for their types. (not preferred because it just ignores the problem)
- Use [`derive_more::Debug`](https://docs.rs/derive_more/latest/derive_more/derive.Debug.html), which does not use the incorrect trait bounds. (not preferred because it adds a dependency)

mysteriouslyseeing:
# Objective

Fixes #17988

## Solution

Added two Debug impls to the `impl_ptr` macro - one for Aligned and one for Unaligned.

## Testing

No tests have been added. Would a test guaranteeing debug layout be appropriate?

---

## Showcase

The debug representation of a `Ptr<'_, Aligned>` follows. `PtrMut` and `OwningPtr` are similar.

`Ptr<Aligned>(0x0123456789ab)`

### DIFF
--- a/crates/bevy_ptr/src/lib.rs
+++ b/crates/bevy_ptr/src/lib.rs
@@ -9,7 +9,7 @@
 
 use core::{
     cell::UnsafeCell,
-    fmt::{self, Formatter, Pointer},
+    fmt::{self, Debug, Formatter, Pointer},
     marker::PhantomData,
     mem::ManuallyDrop,
     num::NonZeroUsize,
@@ -159,7 +159,7 @@ impl<'a, T: ?Sized> From<&'a mut T> for ConstNonNull<T> {
 ///
 /// It may be helpful to think of this type as similar to `&'a dyn Any` but without
 /// the metadata and able to point to data that does not correspond to a Rust type.
-#[derive(Copy, Clone, Debug)]
+#[derive(Copy, Clone)]
 #[repr(transparent)]
 pub struct Ptr<'a, A: IsAligned = Aligned>(NonNull<u8>, PhantomData<(&'a u8, A)>);
 
@@ -174,7 +174,6 @@ pub struct Ptr<'a, A: IsAligned = Aligned>(NonNull<u8>, PhantomData<(&'a u8, A)>
 ///
 /// It may be helpful to think of this type as similar to `&'a mut dyn Any` but without
 /// the metadata and able to point to data that does not correspond to a Rust type.
-#[derive(Debug)]
 #[repr(transparent)]
 pub struct PtrMut<'a, A: IsAligned = Aligned>(NonNull<u8>, PhantomData<(&'a mut u8, A)>);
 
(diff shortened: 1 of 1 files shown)


crates/bevy_ptr/src/lib.rs
```diff
@@ -265,6 +263,19 @@ macro_rules! impl_ptr {
                 Pointer::fmt(&self.0, f)
             }
         }
+
+        impl Debug for $ptr<'_, Aligned> {
```
MrGVSV, line 267: I’m in favor of the manual implementation, but I feel like we should also derive `Debug` on `Aligned` and `Unaligned`.

This will make it easy for users who define their own generic wrappers to derive `Debug` rather than also needing a manual implementation.

alice-i-cecile (approved):

crates/bevy_ptr/src/lib.rs
mysteriouslyseeing: Would it make sense to add most/all of the derives, then? The same issue with derive bounds exists for all std derive macros, IIRC

MrGVSV (approved):

PR #18043, opened 2025-02-26
Change `Commands::get_entity` to return `Result` and remove panic from `Commands::entity`

JaySpruce:
## Objective

Alternative to #18001.

- Now that systems can handle the `?` operator, `get_entity` returning `Result` would be more useful than `Option`.
- With `get_entity` being more flexible, combined with entity commands now checking the entity's existence automatically, the panic in `entity` isn't really necessary.

## Solution

- Changed `Commands::get_entity` to return `Result<EntityCommands, EntityDoesNotExistError>`.
- Removed panic from `Commands::entity`.

### DIFF
--- a/crates/bevy_ecs/src/relationship/mod.rs
+++ b/crates/bevy_ecs/src/relationship/mod.rs
@@ -116,7 +116,7 @@ pub trait Relationship: Component + Sized {
             {
                 relationship_target.collection_mut_risky().remove(entity);
                 if relationship_target.len() == 0 {
-                    if let Some(mut entity) = world.commands().get_entity(target_entity) {
+                    if let Ok(mut entity) = world.commands().get_entity(target_entity) {
                         // this "remove" operation must check emptiness because in the event that an identical
                         // relationship is inserted on top, this despawn would result in the removal of that identical
                         // relationship ... not what we want!
--- a/crates/bevy_ecs/src/system/commands/mod.rs
+++ b/crates/bevy_ecs/src/system/commands/mod.rs
@@ -467,7 +467,7 @@ impl<'w, 's> Commands<'w, 's> {
     ///     // Get the entity if it still exists, which it will in this case.
     ///     // If it didn't, the `?` operator would propagate the returned error
     ///     // to the system, and the system would pass it to an error handler.
-    ///     entity_commands = commands.get_entity(entity)?;
+    ///     let mut entity_commands = commands.get_entity(entity)?;
     ///
     ///     // Add a single component to the entity.
     ///     entity_commands.insert(Label("hello world"));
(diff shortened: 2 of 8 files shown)


alice-i-cecile (approved):
This is so much better. We should have done this years ago. I much prefer this solution: it's a straightforward improvement with minimal breaking changes.

Like I said on Discord, commands being queued for an entity that doesn't exist is rare (since you won't see them in queries), but commands being evaluated for an entity that doesn't exist is much more common. Simplifying the logic here to remove the pointless panic is a clear win IMO, and moving `get_entity` to return a Result is much more idiomatic.

aloucks:
What happens now that the panic is removed? Is the entity automatically created or does it silently ignore and discard it?

alice-i-cecile:
> What happens now that the panic is removed? Is the entity automatically created or does it silently ignore and discard it?

It will be treated as if the entity had been despawned after the command was sent. The exact details of how to respond are up to the specific entitycommand.

aloucks:
> > What happens now that the panic is removed? Is the entity automatically created or does it silently ignore and discard it?
> 
> It will be treated as if the entity had been despawned after the command was sent. The exact details of how to respond are up to the specific entitycommand.

that sounds very unintuitive and surprising. I would expect that if i do something like `commands.entity(id).add_children(&children_ids)`, it would spawn those children.

I feel like the existing API (panic and Option) is actually the correct way to go. It aligns with how `slice` behaves. A slice index operation panics if the index is out of bounds (i.e `commands.entity()`) and you can use `slice.get(index)` if you are unsure that it exists (i.e. `commands.get_entity(id)`)

chescock (approved):

alice-i-cecile:
> I use commands.entity(id) when I know it exists, which is when I've just created it or the entity is coming out of a query iterator. It doesn't make sense to me to make it return Result (as is in the other related PR).

Agreed completely here!

> That sounds very unintuitive and surprising. I would expect that if i do something like commands.entity(id).add_children(&children_ids), it would spawn those children. Having it fail silently sounds like a big footgun.

We deliberately restrict the ability to spawn entities with a specific ID. Spawning entities implicitly is not a good choice.

This PR is effectively equivalent to the current behavior: the panic / failure is simply moved later. This a) allows commands to respond as needed and b) ensures that this failure can be integrated into the broader error handling mechanisms added in 0.16.

aloucks:
> This PR is effectively equivalent to the current behavior: the panic / failure is simply moved later. This a) allows commands to respond as needed and b) ensures that this failure can be integrated into the broader error handling mechanisms added in 0.16.

Cool, this now makes much more sense given this context. Thank you!

hymm:
Should this be considered a breaking change? i.e. if you want to know it the entity is invalid at commands creation you need to switch to use get_entity now.

cart (approved):

cart:
> Should this be considered a breaking change? i.e. if you want to know it the entity is invalid at commands creation you need to switch to use get_entity now instead of entity.

Yeah in theory panicking is part of the API, but I also suspect that the breaking impact of this change will be non-existent or close to non-existent, as it will generally  just change _where_ a command panics, not _if_ it panics. If someone was using catch_unwind to recover in the system, this would definitely be breaking, but I doubt anyone is doing that.

But yeah from a strict "is this a breaking change" perspective the answer is yes. From a "is thinking about this worth our users' time" perspective, I'm not so sure.

Issue #16679, opened 2024-12-06
Forward rendering seems broken


JMS55:
## Bevy version
72f096c91ea1056c41f2da821d310765c4482c00

## Relevant system information

```
2024-12-06T04:20:17.216608Z  INFO bevy_diagnostic::system_information_diagnostics_plugin::internal: SystemInfo { os: "Windows 11 Home", kernel: "22631", cpu: "AMD Ryzen 7 5800X 8-Core Processor", core_count: "8", memory: "15.9 GiB" }
2024-12-06T04:20:17.720519Z  INFO bevy_render::renderer: AdapterInfo { name: "NVIDIA GeForce RTX 3080", vendor: 4318, device: 8710, device_type: DiscreteGpu, driver: "NVIDIA", driver_info: "560.94", backend: Vulkan }
```

## What you did

Run the deferred_rendering example

## What went wrong

Forward rendering seems broken (forward + prepass is fine)

* POM seems broken on the cube
* Normal maps seem to be missing from the flight helmet

pcwalton:
Bisecting this. The lack of POM on the cube repros even as far back as Sep 13. So that at least isn't a recent regression.

DGriffin91:
Lack of parallax on the cube was also in the initial creation of this scene. Forward + prepass has never supported parallax mapping.

JMS55:
Yes, but it's also broken in pure forward (there's some weird ghosting 🙃)

pcwalton:
The culprit for the loss of normal mapping is the retained render world, #15320.

Note that deferred was broken at the time that commit landed (due to PCSS) and wasn't fixed until #15656, which is ddd4b4daf8bb3df1dac0fd5f453269ff4bd4f99a. So you'll want to `git cherry-pick ddd4b4daf8bb3df1dac0fd5f453269ff4bd4f99a` if you check out that commit.

hukasu:
the issue being described here is the shadow being retained when changing from Deferred or Forward + Prepass into Forward, right?

![Image](https://github.com/user-attachments/assets/08c4461b-2018-4f37-96ce-c08e2eed4cd5)

taken using 85eceb022da0326b47ac2b0d9202c9c9f01835bb

hukasu:
things got worse on 7aeb1c51a6e6e8ec6d0ff659f90a38e848884b94

Deferred  
![Image](https://github.com/user-attachments/assets/e54e969d-8cf7-410a-9e4d-42a3af6fd875)

[...]

PR #18046, opened 2025-02-26
Add scenes similar to `deferred_rendering` to 3d testbed

hukasu:
# Objective

The `deferred_rendering` example is frequently broken, adding it to the testbed will warn PRs that break it.

## Solution

Add scenes similar to `deferred_rendering` to 3d testbed

## Testing

`cargo run --example testbed_3d --features="bevy_ci_testing"`

## Showcase

Screenshots for the new scenes
Forward
screenshot-Forward

ForwardPrepass
screenshot-ForwardPrepass

Deferred
screenshot-Deferred

 ## TODO
- [x] Merge conflicts
- [ ] Limit CI


hukasu:
simplyfied the test and added one more test to try and catch this issue with parallax https://github.com/bevyengine/bevy/issues/16679#issuecomment-2605160381

screenshot-RemovePrepass

hukasu:
Screenshot of the `RemoveForwardPrepass` scene on 5cc3f4727e356f66c8931c4c64fbd672cfc90b0e

image

alice-i-cecile:
Strongly in favor of this general direction. I'll probably defer to @mockersf on the exact testing details and the rendering folks on what should be shown though.

hukasu:
@mockersf  and @IceSentry will instruct me on how to filter this so that it is not run on EVERY PR

examples/testbed/3d.rs
```diff
@@ -15,22 +15,62 @@ fn main() {
...
+                deferred::remove_prepass_timer_init,
+            )
+                .chain(),
+        )
+        .add_systems(Update, switch_scene)
+        .add_systems(
+            Update,
+            (
+                deferred::remove_prepass.run_if(resource_removed::<deferred::RemovePrepassTimer>),
+                deferred::remove_prepass_timer_tick
```
mockersf, line 50: those systems should only run in their target scene. this is handled by the resource scope, but having the scene name in their condition would make it clearer

hukasu: just pushed a change on this, is this enough?

mockersf:
This feels a bit too much on the different passes when we're not doing a lot of other things in 3d tests, and it increases the testbed_3d execution time from 20 seconds to 1m10.

Maybe it's a better candidate to when we're able to select tests based on PR label

mockersf:
> @mockersf and @IceSentry will instruct me on how to filter this so that it is not run on EVERY PR

not possible at the moment

hukasu:
so it is `X-Blocked` until then?

BenjaminBrienen:
triage: Has merge conflicts, still blocked

PR #18047, opened 2025-02-26
Update `rand`, `glam` and `encase` to latest versions

Bluefinger:
# Objective

New `rand` version, which means updating `glam` and `encase` to support the newer ecosystem update. Does mean that this changes how WASM builds need to be done in order to configure `getrandom` correctly, but this can be remedied with updated docs.

## Solution

Updating all needs deps to their compatible versions. This PR is currently blocked by `encase`, which is waiting on [this PR](https://github.com/teoxoy/encase/pull/88) to be merged and then a new version published.

## Testing

- Must pass CI for all checks, tests, not introduce breaking changes.

---

## Migration Guide

TBA

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -497,8 +497,8 @@ tracing = { version = "0.1", default-features = false, optional = true }
 bevy_dylib = { path = "crates/bevy_dylib", version = "0.16.0-dev", default-features = false, optional = true }
 
 [dev-dependencies]
-rand = "0.8.0"
-rand_chacha = "0.3.1"
+rand = "0.9.0"
+rand_chacha = "0.9.0"
 ron = "0.8.0"
 flate2 = "1.0"
 serde = { version = "1", features = ["derive"] }
--- a/benches/Cargo.toml
+++ b/benches/Cargo.toml
@@ -29,9 +29,9 @@ bevy_platform_support = { path = "../crates/bevy_platform_support", default-feat
 ] }
 
 # Other crates
-glam = "0.29"
-rand = "0.8"
-rand_chacha = "0.3"
+glam = { version = "0.30" }
+rand = "0.9"
+rand_chacha = "0.9"
 
 # Make `bevy_render` compile on Linux with x11 windowing. x11 vs. Wayland does not matter here
 # because the benches do not actually open any windows.
--- a/crates/bevy_math/Cargo.toml
+++ b/crates/bevy_math/Cargo.toml
@@ -10,7 +10,7 @@ keywords = ["bevy"]
 rust-version = "1.85.0"
 
 [dependencies]
-glam = { version = "0.29", default-features = false, features = ["bytemuck"] }
+glam = { version = "0.30", default-features = false, features = ["bytemuck"] }
 thiserror = { version = "2", default-features = false }
 derive_more = { version = "1", default-features = false, features = [
   "from",
@@ -22,8 +22,8 @@ serde = { version = "1", default-features = false, features = [
 ], optional = true }
 libm = { version = "0.2", optional = true }
 approx = { version = "0.5", default-features = false, optional = true }
-rand = { version = "0.8", default-features = false, optional = true }
-rand_distr = { version = "0.4.3", optional = true }
+rand = { version = "0.9", default-features = false, optional = true }
+rand_distr = { version = "0.5", optional = true }
 smallvec = { version = "1.11" }
 bevy_reflect = { path = "../bevy_reflect", version = "0.16.0-dev", default-features = false, features = [
   "glam",
@@ -33,11 +33,11 @@ variadics_please = "1.1"
 [dev-dependencies]
 approx = "0.5"
 # Supply rngs for examples and tests
-rand = "0.8"
-rand_chacha = "0.3"
+rand = "0.9"
+rand_chacha = "0.9"
 # Enable the approx feature when testing.
 bevy_math = { path = ".", default-features = false, features = ["approx"] }
-glam = { version = "0.29", default-features = false, features = ["approx"] }
+glam = { version = "0.30", default-features = false, features = ["approx"] }
 
 [features]
 default = ["std", "rand", "curve"]
--- a/benches/benches/bevy_ecs/observers/propagation.rs
+++ b/benches/benches/bevy_ecs/observers/propagation.rs
@@ -70,7 +70,7 @@ impl<const N: usize> Event for TestEvent<N> {
 }
 
 fn send_events<const N: usize, const N_EVENTS: usize>(world: &mut World, leaves: &[Entity]) {
-    let target = leaves.iter().choose(&mut rand::thread_rng()).unwrap();
+    let target = leaves.iter().choose(&mut rand::rng()).unwrap();
 
     (0..N_EVENTS).for_each(|_| {
         world.trigger_targets(TestEvent::<N> {}, *target);
@@ -111,7 +111,7 @@ fn add_listeners_to_hierarchy<const DENSITY: usize, const N: usize>(
     }
     let mut rng = deterministic_rand();
     for e in nodes.iter() {
-        if rng.gen_bool(DENSITY as f64 / 100.0) {
+        if rng.random_bool(DENSITY as f64 / 100.0) {
             world.entity_mut(*e).observe(empty_listener::<N>);
         }
     }
--- a/benches/benches/bevy_ecs/world/entity_hash.rs
+++ b/benches/benches/bevy_ecs/world/entity_hash.rs
@@ -11,9 +11,9 @@ fn make_entity(rng: &mut impl Rng, size: usize) -> Entity {
     // * For ids, half are in [0, size), half are unboundedly larger.
     // * For generations, half are in [1, 3), half are unboundedly larger.
 
-    let x: f64 = rng.r#gen();
+    let x: f64 = rng.random();
     let id = -(1.0 - x).log2() * (size as f64);
-    let x: f64 = rng.r#gen();
+    let x: f64 = rng.random();
     let generation = 1.0 + -(1.0 - x).log2() * 2.0;
 
     // this is not reliable, but we're internal so a hack is ok
--- a/benches/benches/bevy_reflect/path.rs
+++ b/benches/benches/bevy_reflect/path.rs
@@ -3,7 +3,7 @@ use core::{fmt::Write, hint::black_box, str, time::Duration};
 use benches::bench;
 use bevy_reflect::ParsedPath;
 use criterion::{criterion_group, BatchSize, BenchmarkId, Criterion, Throughput};
-use rand::{distributions::Uniform, Rng, SeedableRng};
+use rand::{distr::Uniform, Rng, SeedableRng};
 use rand_chacha::ChaCha8Rng;
 
 criterion_group!(benches, parse_reflect_path);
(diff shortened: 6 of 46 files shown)


mrchantey (approved):
Looks good, lots of files changed but as mentioned in the pr its a 1:1 mapping to the updated function naming convention. I'm sure the wasm rustflags will be a gotcha for many a rust newbie, it'll be helful for us to explain this as much as possible in docs and guides.

janhohenheim (approved):

mcobzarenco:
Could this be added to 16.1 milestone?

Bluefinger:
> Could this be added to 16.1 milestone?

I think this would cause breaking changes that might be painful to deal with for 0.16 projects, which would not really fit nicely with semver. `getrandom` and WASM in particular will require some massaging and at least a blog post entry to notify people that they'll have to do some things a bit differently to get it to work.

I understand wanting to have this *now* though... shame it took like three months to get all the deps ready for this upgrade.

alice-i-cecile:
> I think this would cause breaking changes that might be painful to deal with for 0.16 projects, which would not really fit nicely with semver. getrandom and WASM in particular will require some massaging and at least a blog post entry to notify people that they'll have to do some things a bit differently to get it to work.

Correct: 0.17 only unfortunately :(

@Bluefinger, we've changed how migration guides work as of the 0.17 cycle. Could you copy your excellent migration guide into a dedicated file quick? Once that's done I'll merge this!

Bluefinger:
@alice-i-cecile Done. Feel free to double check the content or adjust wording.

alice-i-cecile:
> getrandom is also updated, and will require additional configuration when building Bevy for WASM/Web (if also using rand). The full details of how to do this is in the getrandom docs

FYI, @mockersf has warned me that the changes needed here for Web for `getrandom` are quite painful. Reading the docs, I agree with that initial assessment. I want to be careful to evaluate the impact there before merging this, as reverting this will not be fun.

I don't intend to leave us on the old versions forever, but it's important to note that this is much higher stakes than normal dependency bumps.

Bluefinger:
> > getrandom is also updated, and will require additional configuration when building Bevy for WASM/Web (if also using rand). The full details of how to do this is in the getrandom docs
> 
> FYI, @mockersf has warned me that the changes needed here for Web for `getrandom` are quite painful. Reading the docs, I agree with that initial assessment. I want to be careful to evaluate the impact there before merging this, as reverting this will not be fun.
> 
> I don't intend to leave us on the old versions forever, but it's important to note that this is much higher stakes than normal dependency bumps.

If we don't go forward with the upgrade, *please let's signal such to the `rand`/`getrandom` folks the reasoning for this*. Though there's also some [exploratory work](https://github.com/bushrat011899/getrandom/tree/wip_split_getrandom) being done by @bushrat011899 to see if this can be improved without all the ergonomic hits `getrandom` v0.3 has while still maintaining their desired security threshold. But I want to ensure there's motivation on both ends here to resolve what is a pretty... awful UX with the `getrandom` crate.

newpavlov:
>awful UX with the getrandom crate.

If you want a good out-of-box support, then you should advocate for a proper Web WASM target. For example, `getrandom` has absolutely no issues with the WASI targets. The "awfulness" is caused by the wide and long standing abuse of the `wasm32-unknown-unknown` target in the Rust ecosystem. Alternatively, you could push for adding entropy source to `std`/`core` (e.g. see https://github.com/rust-lang/rust/issues/130703) with an ability to overwrite it similarly to `#[global_allocator]`.

As a `getrandom` maintainer I would like to declare that we do not consider any potential changes to the `wasm32-unknown-unknown` handling in `getrandom` v0.3. We _may_ change it in a future breaking release (which is unlikely to happen in several years), but we do not have better ideas right now how to handle this problem with the existing Rust/Cargo capabilities (e.g. [global features](https://internals.rust-lang.org/t/19618) could help).

alice-i-cecile:
> If you want a good out-of-box support, then you should advocate for a proper Web WASM target.

Yep, this is high on our wishlist too :)

> Alternatively, you could push for adding entropy source to std/core  with an ability to overwrite it similarly to #[global_allocator].

Makes sense: no objections to this design either.

> As a getrandom maintainer I would like to declare that we do not consider any potential changes to the wasm32-unknown-unknown handling in getrandom v0.3. We may change it in a future breaking release (which is unlikely to happen in several years)

This is useful to know. Moderately frustrating, as I would like to be unblocked on updating our dependencies well before that, but I understand that as a foundational crate you don't want to release breaking changes too frequently. If we discover a good solution, we may temporarily fork to get unblocked sooner, although that will require coordination with our other dependencies.

The other option is for us to build out and push tooling to help manage this pain for users on the web, e.g. `bevy_cli`. That's not quite ready yet, but it is a solution that's entirely within our control, which is desirable.

josephlr:
Hey everyone, I'm the other maintainer of `getrandom`, and I'd like to know if there's anything we can improve here (subject to our constraints).

For context as @newpavlov mentioned, [`wasm32-unknown-unknown`](https://doc.rust-lang.org/nightly/rustc/platform-support/wasm32-unknown-unknown.html) is one of the most annoying targets for us to deal with. Unlike other fully-supported Web targets ([`wasm32-unknown-emscripten`](https://doc.rust-lang.org/nightly/rustc/platform-support/wasm32-unknown-emscripten.html),[`wasm32-wasip1`](https://doc.rust-lang.org/nightly/rustc/platform-support/wasm32-wasip1.html),[`wasm32-wasip2`](https://doc.rust-lang.org/nightly/rustc/platform-support/wasm32-wasip2.html)), this target assumes _nothing_ about the external environment, so we cannot even assume that JavaScript exists when building for this target. This means we need to have some control (be it Cargo flags, `--cfg` flags, some macro invocation, etc...) for a user to say "hey I'm compiling for the web".

Further complicating things, overriding the source of randomness is a very security-sensitive thing to do, so we don't want to allow any crate in the dependency graph to override it, making it challenging to only use Cargo features (which can be set by any crate on any other crate). So we use `RUSTFLAGS` to override it via `--cfg getrandom_backend=some_custom_backend`. Ideally, we would have something similar to [`#[global_allocator]`](https://doc.rust-lang.org/std/alloc/index.html#the-global_allocator-attribute), instead of `RUSTFLAGS` (which are not ideal), but https://github.com/rust-lang/rust/issues/130703 is required for that to work (we tried doing something like @bushrat011899's [exploration](https://github.com/bushrat011899/getrandom/tree/wip_split_getrandom), but we couldn't get it working without that feature).

The other problem is that the `wasm-bindgen` crates have a very large number of transitive dependencies, so we need to have an off-by-default Cargo feature to avoid issues like https://github.com/rust-random/getrandom/issues/433#issuecomment-2557411006. This leads to the situation where you have both `--cfg getrandom_backend="wasm_js"` and `features = ["wasm_js"]`.

Without releasing a breaking change (which would be quite disruptive), we could:
  - Improve the documentation, error messages, etc...
  - Make it so `features = ["wasm_js"]` activates `getrandom_backend="wasm_js"` by default (potentially with a warning saying `--cfg getrandom_backend="wasm_js"` is missing). 

> As a `getrandom` maintainer I would like to declare that we do not consider any potential changes to the `wasm32-unknown-unknown` handling in `getrandom` v0.3.

I would take a slightly more relaxed stance, that we cannot have any _breaking_ changes to our handling of the target, but if there's something we could do that's not a breaking change, I would consider it.

newpavlov:
>Make it so features = ["wasm_js"] activates getrandom_backend="wasm_js" by default

Personally, I am against it. The unfortunate reality is that people unconditionally enable such features in library crates for "convenience" despite strongly worded warnings against it. This breaks builds for users who target non-Web WASM. Even worse, just having `wasm-bindgen` in project's dependency tree can be problematic in such cases (we had one such user, unfortunately I can not find his comment right now).

Overall, we discussed this problem long enough before settling on the current solution with the full understanding of drawbacks. I think that the ecosystem should just bite the bullet here. Hopefully, it will add a bit of motivation for improving the unsatisfactory Web WASM situation. In my opinion, introducing new exceptions and workarounds at this stage would do more harm than good.

bushrat011899:
> we tried doing something like @bushrat011899's exploration, but we couldn't get it working without feature(random) or feature(extern_item_impls)

@josephlr If I completed that exploration branch, would it be a design the `getrandom` project would be interested in adopting? I don't believe I need either of those experimental features, as I'm basing my design off of the already stable (and widely used) `critical-section` crate.

Obviously no one person speaks for the project, so "interest" is good enough for me to be motivated to complete the prototype. I have no desire to fragment the Rust ecosystem for an ergonomics change (that's the _worst_ ergonomics!).

Bluefinger:
@alice-i-cecile @mockersf Just as a heads up, I don't think we can avoid the effect of RUSTFLAGS/`getrandom` v0.3 since it is already within our dependencies: `uuid` via `bevy_reflect`. So if we are worried about the ergonomics of this on our endusers, we either have to revert to an older `uuid` version in the mean time, or go ahead with the upgrade here (so to reduce the amount of duplicate deps a bit).

Bluefinger:
@josephlr just to clarify, the issue with the `getrandom` v0.3 dep is actually in our `v0.16` version, because `uuid` made a minor version update to update to `getrandom` v0.3 and `rand` v0.9, which is a semver breakage. So we will have to pin `uuid` version for `v0.16.1` as a fix for the mean time, and then decide whether to go ahead with this PR or not for v0.17.

tysen:
For what it's worth, I don't think the end-user requirements for using `getrandom` v0.3 are _that_ much worse than the additional load an end-user takes on by targeting the web in the first place. Nontrivial web projects need to use `web-sys`, `js-sys`, `wasm-bindgen` (both the crate as a library and the cli) at least, and probably also tools like `wasm-pack` and/or `wasm-opt`. Setting a `RUSTFLAGS` is a comparatively minor papercut, imo, and I think pinning bevy on an old version to avoid that will do more harm than good.

mnmaita:
@Bluefinger could you please update `encase_derive_impl` in this PR too? See https://github.com/bevyengine/bevy/pull/19669.

crates/bevy_math/Cargo.toml
```diff
@@ -33,11 +33,14 @@ variadics_please = "1.1"
...
-rand_chacha = "0.3"
+rand = "0.9"
+rand_chacha = "0.9"
 # Enable the approx feature when testing.
 bevy_math = { path = ".", default-features = false, features = ["approx"] }
-glam = { version = "0.29.3", default-features = false, features = ["approx"] }
+glam = { version = "0.30.1", default-features = false, features = ["approx"] }
+
+[target.'cfg(target_arch = "wasm32")'.dev-dependencies]
+getrandom = { version = "0.3", features = ["wasm_js"] }
```
mnmaita: This could also have `default-features = false` right?

benches/Cargo.toml
```diff
@@ -29,9 +29,9 @@ bevy_platform = { path = "../crates/bevy_platform", default-features = false, fe
 ] }
 
 # Other crates
-glam = "0.29"
-rand = "0.8"
-rand_chacha = "0.3"
+glam = { version = "0.30.1" }
```
mnmaita:
```suggestion
glam = "0.30.1"
```
Just to keep it as it was 😄

Cargo.toml
```diff
@@ -617,6 +617,7 @@ smol-hyper = "0.1"
 ureq = { version = "3.0.8", features = ["json"] }
 
 [target.'cfg(target_arch = "wasm32")'.dev-dependencies]
+getrandom = { version = "0.3", features = ["wasm_js"] }
```
mnmaita: Same question here regarding `default-features = false`

mnmaita (approved):
Minor nits but nothing blocking I think, would be nice to take this to the finish line so you don't have to wrangle conflicts for too long. 😅 Thanks a lot!

crates/bevy_math/Cargo.toml
Bluefinger: `getrandom` does not have any default features enabled, so it would be redundant.

alice-i-cecile: They might in the future though, so this would be mildly nice as a bit of defensive programming.

mnmaita: Both good points. I was thinking about the extra future proofing that this brings but it's not a big deal.

Cargo.toml
Bluefinger: As above, `getrandom` does not have any default features, so this would be redundant.

crates/bevy_math/Cargo.toml
Bluefinger: I would point out that features are (meant to be) additive so at least it wouldn't be a breaking change, especially going from `no_std` to `std` would be a breaking change that would cripple the rust ecosystem should that happen. However, given we don't want potentially extra deps being included in the future (somehow), it's probably wiser to ensure that this doesn't happen, however unlikely.

mockersf (approved):

mockersf:
I think this should be merged for the 0.17.

atlv24:
i think so too

james7132 (approved):
Having read through the discussion above and the additional considerations raised by @tysen, I think it's best for us to merge this in for the 0.17 release, even with the impact to web UX. Perhaps we should provide a utility or script for building for web should these "papercuts" continue to accumulate.

alice-i-cecile:
The [`bevy_cli`](https://github.com/TheBevyFlock/bevy_cli/) working group has assembled platform-specific build commands for exactly this reason :)

PR #18049, opened 2025-02-26
BRP resource methods

Leinnan:
# Objective

BRP has no way to interact with Resources, this is a first step to change that.

## Solution

- Add methods to:
  - list resources
  - get content of the resource
  - mutate a resource

## Testing

Tested the commands by running the BRP sample( `cargo run --example server --features="bevy_remote"`) and with these curl commands:
```
curl -X POST -d '{ "jsonrpc": "2.0", "id": 1, "method": "bevy/resources/list"}' 127.0.0.1:15702 | jq .
```
```
curl -X POST -d '{ "jsonrpc": "2.0", "id": 1, "method": "bevy/resources/get", "params": "bevy_picking::PickingPlugin" }' 127.0.0.1:15702 | jq .
```

```
curl -X POST -d '{ "jsonrpc": "2.0", "id": 1, "method": "bevy/resources/mutate", "params": {"resource": "bevy_picking::PickingPlugin", "path":"is_enabled", "value": false} }' 127.0.0.1:15702 | jq .
```

### DIFF
--- a/crates/bevy_remote/src/builtin_methods.rs
+++ b/crates/bevy_remote/src/builtin_methods.rs
@@ -986,7 +986,6 @@ pub fn process_remote_resources_list_request(
     serde_json::to_value(response).map_err(BrpError::internal)
 }
 
-#[allow(unsafe_code)]
 /// Handles a `bevy/resources/get` request (get resource) coming from a client.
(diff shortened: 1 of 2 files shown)


Leinnan:
Close since there is already one #17423 accepted and ready to merge

PR #18056, opened 2025-02-26
chore: update compile fail test stderr files

RobWalt:
I noticed this while working on #18017 . Some of the `stderr` compile_fail tests were updated while I generated the output for the new tests I introduced in the mentioned PR.

I'm on rust 1.85.0

### DIFF
--- a/crates/bevy_ecs/compile_fail/tests/ui/query_exact_sized_iterator_safety.stderr
+++ b/crates/bevy_ecs/compile_fail/tests/ui/query_exact_sized_iterator_safety.stderr
@@ -1,20 +1,22 @@
-error[E0277]: the trait bound `bevy_ecs::query::Changed<Foo>: ArchetypeFilter` is not satisfied
+error[E0277]: `bevy_ecs::query::Changed<Foo>` is not a valid `Query` filter based on archetype information
   --> tests/ui/query_exact_sized_iterator_safety.rs:7:28
    |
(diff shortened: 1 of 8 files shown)


alice-i-cecile (approved):
Not sure why CI wasn't catching these...

mockersf:
if CI didn't fail on that before, it means CI wasn't running them. Could you manage to get CI to fail before fixing them?

RobWalt:
> if CI didn't fail on that before, it means CI wasn't running them. Could you manage to get CI to fail before fixing them?

Nope doesn't seem like it was failing although it was clearly broken prior to https://github.com/bevyengine/bevy/pull/18017/commits/b452c928f9888d40a9ddb27fa8e9fc6fd838f6cc

Brezak:
> if CI didn't fail on that before, it means CI wasn't running them. Could you manage to get CI to fail before fixing them?

[We ignore] differences between the stderr output of compile fail tests. When testing derive macros the backtrace would leak the location of the developers `.cargo` directory. [I tried to filter the paths out](https://github.com/bevyengine/bevy/blob/2f384643c3cde3fb677828b836dc9e965be4ebed/tools/compile_fail_utils/src/lib.rs#L67-L70) but the filter would break randomly.

I seem to remember reading somewhere that rust now strips the path to `.cargo` in errors. I'll check tomorrow and if yes we can enable comparing checking the stderr output again.

Issue #18054, opened 2025-02-26
Improve Performance of Entities::alloc_at


ElliottjPierce:
`Entities::alloc_at` and `alloc_at_without_replacement` currently have performance issues. See [this conversation](https://discord.com/channels/691052431525675048/1342561489734602853/1344086844572438528) for the full background.

# When and why is it slow?

Both these functions currently have a linear search through the `pending` list to try to find and remove the requested entity. When there are lots of `pending` entities, this is very slow.

See also the [source](https://github.com/bevyengine/bevy/blob/11db71727d25bdbe86f709574b531cab686d0e0a/crates/bevy_ecs/src/entity/mod.rs#L700).

# Practical Impact

In the conversation mentioned above, a user was spawning waves of sprite projectiles. Each wave was ~90,000 entities and was spawned and desawned over 8 seconds, repeating for each wave. This lead to measurable frame drops because, on the second wave, the render world was using `alloc_at_without_replacement` with 90,000 entities in the pending list. For massive, wave oriented games, this is a big deal.

# Solution

Add a field to `EntityMeta` that stores the index of the entity in the `pending` list if it is fee. This will remove the linear search and fix the problem.

Downside: `EntityMeta` will be bigger, so less of them can fit in cache. This is relevant to lots of *very* hot paths.

# Alternatives

We could try to keep the `pending` list small by not storing tail entities. For example, if there are 16 entries in `meta` and entities of index 8-15 are in the pending list, we could remove them since their pending nature could be implied by shortening `meta`. But this gives extra overhead and can be ruined by keeping just one high-index entity around.

We could also try to pack the index of the entity in the `pending` list into some other fields of `EntityMeta`. Maybe `table_row` or something, but that looses/invalidates other information about the entity. If anyone has a way of doing this without dropping some information, this would be an ideal solution.

Finally, we could sort the pending list after `Entities::flush`. This could let us change the linear search to a binary search, but keeping the sorted invariance may have additional overhead.

alice-i-cecile:
We should start by adding a benchmark specifically stressing this. The exact pattern from the user of "spawn 100k entities, despawn most of them, try spawning more" is really suitable for creating a microbenchmark that we can use to measure regressions against this.

ElliottjPierce:
> We should start by adding a benchmark specifically stressing this. 

I'm on it!

alice-i-cecile:
[...]

PR #18057, opened 2025-02-26
Benches for wave spawning

ElliottjPierce:
# Objective

Establish benchmarks for waves of entity spawning as discussed in #18054. This does not include (for now) wave insertions, as mere insertions don't affect entity allocation.

## Solution

Add benchmarks for permutations of entity count, wave count, component size, and batch spawning function.

The benchmarks do not show much difference in the component size variance, so I would be happy to remove that and decrease complexity if desired.

## Showcase

Yikes! Here are the results on M2 MAX:

Screenshot 2025-02-26 at 4 11 41 PM
(Ignore the regressions and improvements. That's from testing the benches.)

Here, I only ran the `EmptyComponent` benches.

With only one wave, `spawn_batch` and `insert_or_spawn_batch` both run in roughly 3ms because there are no shenanigans in `Entities` regarding allocation.

So if one wave takes 3ms, 2 waves should take 6ms, right? Nope! `spawn_batch` on 2 waves takes 6ms, but `insert_or_spawn_batch` with the infamous `alloc_at_without_replacement` takes **760ms!!**. That's a little more than 6.

Running 16 waves was taking too long, but I'll leave it running for a bit so we can get a number. I would estimate ~11 seconds (700ms per wave on average) for `insert_or_spawn_batch`. This implies that the relationship isn't quite linear since the above 2 waves was also 700ms. But this could be memory allocator shenanigans too.

**EDIT:** It did end up being ~11.5 seconds.

### DIFF
--- a/benches/benches/bevy_ecs/world/mod.rs
+++ b/benches/benches/bevy_ecs/world/mod.rs
@@ -3,6 +3,7 @@ mod despawn;
 mod despawn_recursive;
 mod entity_hash;
 mod spawn;
+mod waves;
 mod world_get;
 
 use commands::*;
--- a/benches/benches/bevy_ecs/world/mod.rs
+++ b/benches/benches/bevy_ecs/world/mod.rs
@@ -36,4 +36,5 @@ criterion_group!(
     query_get_many::<5>,
     query_get_many::<10>,
     entity_set_build_and_lookup,
+    waves::world_wave_spawn,
 );
--- a/benches/benches/bevy_ecs/world/waves.rs
+++ b/benches/benches/bevy_ecs/world/waves.rs
@@ -1,0 +1,54 @@
+use bevy_ecs::prelude::*;
+use criterion::Criterion;
+
+#[derive(Component, Default)]
+struct EmptyComponent;
+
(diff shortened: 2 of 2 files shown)


alice-i-cecile:
Nice stuff! Now that we've isolated the problem in question to an API that we want to cut anyways, I'm a bit iffy on adding these benchmarks. This specific problem feels very bizarrely idiosyncratic. While this PR was 100% worth the time to demonstrate the problem, I'm not sure that this set of benches is going to catch any new problems.

> The benchmarks do not show much difference in the component size variance, so I would be happy to remove that and decrease complexity if desired.

If we decide to keep this let's cut that.

ElliottjPierce:
> I'm a bit iffy on adding these benchmarks.

Totally agree. If the solution is to remove the think we're benchmarking, what's really the point? Still glad I got the numbers though.

If you do choose to keep this, just ping me, and I'll clean it up. Otherwise feel free to close the pr.

Also can't help but adding that the most recent label change makes I think 100% of my prs and issues contentious or controversial. Rightfully so, but it is funny.

PR #18058, opened 2025-02-26
Add methods to bulk replace relationships on a entity

### LINKED ISSUE #18041: Bulk operations on entity relationships

Before relationships, we had the `replace_children` method, which wasn't particularly efficient (it removed all existing children, even the ones that were carried over, and then replaced them with the new list), but it was useful. There's no method in the new API which provides the same level of functionality.

For implementing reactions (as illustrated in cart's discussion post as well as my various libraries) it is a very common operation to want to replace all children of an entity with a new list of children, where the two lists may have elements in common. Currently this requires a hacky workaround: removing the `Children` component from the parent, and then adding the new children via `add_related`.

Ideally what we want to have happen is:

* Remove the `ChildOf` component from any existing children which are not in the new list.
* Add `ChildOf` components to any new children which are not in the old list.
* For elements which are in both the old and new lists, do nothing - don't modify the `ChildOf` components.
* Update the `Children` component to contain the new list with the correct ordering of entities.

Note that in many cases the first step can be omitted, since the old children are about to be despawned anyway.

Now, it might sound that in order to do this correctly, you would need to allocate memory - effectively what's described in the preceding steps are set operations (all A not in B) which would require an EntitySet. In other words, you'd start by initializing the set to the existing children, and remove from the set any children that are being carried over. What's left is the children that need to be unlinked. Unfortunately, set operations allocate memory.

However, we can avoid this by making the caller do the work. The reactive framework has in all likelyhood already done this calculation (determining which children are new and which are to be removed), so there's no need for the low-level relationship API to do this computation again. What the caller cannot (and should not) do, however, is directly modify the contents of Children and ChildOf components. So we need an API which allows the caller to provide the results of this calculation.

What I am envisioning is a method on `EntityWorldMut` which accepts three arguments (plus self), all of which are entity slices:

* The replacement entity list, which will become the new contents of `Children`.
* A list of entities which are being added. A `ChildOf` component will be added to each entity in this list.
* A list of entities which are being removed. The `ChildOf` component will be removed from each entity in this list.

It's the caller's responsibility to ensure that the arguments are consistent: the list of newly added entities should be a subset of the list of replacements, and the list of newly removed entities should be a subset of the exiting children.

In cases where the removed entities are being immediately despawned, we can get by with passing in an empty slice as the third argument - this means that the work of removing the `ChildOf` components can be skipped.

We may also want to have a variant of this which replaces just a sub-range of children, specified by starting index and length. So there would be 4 arguments: the three given above, plus a range argument.

It's possible to do this one entity at a time with current APIs, but it's not very efficient. The current hacky workaround also works, but does needless archetype moves (deleting the `ChildOf` component and then immediately adding it back again).

Assuming we can agree on the design of the API, the implementation should be relatively easy.

## Additional context

See the discussion on in-place reactions #17917

@cart @alice-i-cecile

Brezak:
# Objective

Add a way to efficiently replace a set of specifically related entities with a new set.
Closes #18041 

## Solution

Add new `replace_(related/children)` to `EntityWorldMut` and friends.

## Testing

Added a new test to `hierarchy.rs` that specifically check if `replace_children` actually correctly replaces the children on a entity while keeping the original one.

---

## Showcase

`EntityWorldMut` and `EntityCommands` can now be used to efficiently replace the entities a entity is related to.

```rust
/// `parent` has 2 children. `entity_a` and `entity_b`.
assert_eq!([entity_a, entity_b], world.entity(parent).get::<Children>());

/// Replace `parent`s children with `entity_a` and `entity_c`
world.entity_mut(parent).replace_related(&[entity_a, entity_c]);

/// `parent` now has 2 children. `entity_a` and `entity_c`.
///
/// `replace_children` has saved time by not removing and reading
/// the relationship between `entity_a` and `parent`
assert_eq!([entity_a, entity_c], world.entity(parent).get::<Children>());

### CODE REVIEW

File: crates/bevy_ecs/src/relationship/related_methods.rs [hunk 1]
```diff
@@ -34,6 +34,39 @@ impl<'w> EntityWorldMut<'w> {
         self
     }
 
+    /// Replaces all the related entities with the given set of new related entities.
+    pub fn replace_related<R: Relationship>(&mut self, related: &[Entity]) -> &mut Self {
+        if related.is_empty() {
```
viridia, line 39: This seems incorrect: replacing the children with an empty list should remove all existing children.

crates/bevy_ecs/src/relationship/related_methods.rs
```diff
@@ -34,6 +34,35 @@ impl<'w> EntityWorldMut<'w> {
...
+        let id = self.id();
+        self.world_scope(|world| {
+            for related in relations_to_remove {
+                world.entity_mut(related).remove::<R>();
+            }
+
+            for related in potential_relations {
+                world.entity_mut(related).insert(R::from(id));
+            }
+        });
```
villor, line 61: A problem with this approach is that the resulting target collection may not result in the same order as the passed in `related`.

As a user of `replace_related` I would assume I can pass in a new order of the same set of entities to reorder them (for relationships requiring a stable ordering, e.g. `Children`).

I'm also curious about the performance benefits of this approach (allocating hashsets) over just removing the target collection and calling `add_children` (archetype moves?).

villor (changes requested):

crates/bevy_ecs/src/relationship/related_methods.rs
viridia: In the original issue I posted (#18041), I proposed a method to do this without using hash sets, however it entails a more complex API in which the caller does the work of determining which entities are added and removed. This puts the responsibility on the caller to ensure that the arguments are consistent.

In my various reactive experiments, I need to do this computation anyway (for example `foreach` computes the diff as part of its operation), so it's relatively easy for me to pass that information along to the `replace_related` API. Having `replace_related` do the calculation internally means that the set operations get done twice.

That being said, I may be overly/prematurely optimizing here. Maybe using hash sets is OK? I can't really say one way or the other.

villor: I'm not too fond of requiring the user to pass correct lists. Seems like this can easily be misunderstood/lead to bugs.

For more specialized scenarios (like tree rendering mechanisms) where the caller already knows the “diff”, there could be an alternative function taking the `new_related` and `deleted` with a name that communicates the “riskiness”.

Or those use cases can update the relationship components directly, and use the hook mode to avoid archetype moves.

Brezak: >A problem with this approach is that the resulting target collection may not result in the same order as the passed in related.

We could directly modify the target collection to change the ordering but that would conflict with users overriding relationship `on_insert` hooks. My opinion is that a user will break any assumptions we make if given the chance. Imagine for example a `on_insert` hook that filters what entities can be inserted into a relationship. If we then manually insert entities into this collection we would be bypassing this filtering.

> For more specialized scenarios (like tree rendering mechanisms) where the caller already knows the “diff”, there could be an alternative function taking the new_related and deleted with a name that communicates the “riskiness”.

I'd be willing to add this method too, but such a method requires #17858. Otherwise the method would be left with just calculating a difference between the `insert` and `remove` entity sets and inserting that (if we add any existing relationships the on_add hook will cause a double insert). Exactly what the current implementation does (and calculating a difference either requires a allocated set or the ability to sort the passed in slices. i.e. either allocating or breaking ordering).

alice-i-cecile: #17858 has been merged now :)

viridia (approved):
Looks good. This will be a fundamental building block for entity templating systems.

Brezak:
This is mostly done. The documentation needs some editing and the naming could use some bikeshedding. Sadly I'm spent for today. See you tomorrow.

crates/bevy_ecs/src/relationship/related_methods.rs
```diff
@@ -34,6 +36,149 @@ impl<'w> EntityWorldMut<'w> {
...
+                    <R::RelationshipTarget as RelationshipTarget>::Collection::with_capacity(
+                        entities_to_relate.len(),
+                    );
+                empty.extend_from_iter(entities_to_relate.iter().copied());
+
+                // SAFETY: We've just initialized this collection
+                self.insert_with_relationship_insert_hook_mode(
+                    R::RelationshipTarget::from_collection_risky(empty),
+                    RelationshipInsertHookMode::Skip,
+                );
```
villor, line 175: I'm not sure you _have_ to use the hook mode here, as there is no `on_insert` hook for the `RelationshipTarget`.

crates/bevy_ecs/src/relationship/related_methods.rs
```diff
@@ -34,6 +36,149 @@ impl<'w> EntityWorldMut<'w> {
...
+        let Some(existing_relations) = self.get::<R::RelationshipTarget>() else {
+            return self.add_related::<R>(related);
+        };
+
+        let mut potential_relations = EntityHashSet::from_iter(related.iter().copied());
+        let mut relations_to_remove = EntityHashSet::with_capacity(related.len());
+
+        for related in existing_relations.iter() {
+            if !potential_relations.remove(related) {
+                relations_to_remove.insert(related);
```
villor, line 50: Maybe you could remove the component here, and avoid the extra hash set/double iteration.

crates/bevy_ecs/src/relationship/related_methods.rs
```diff
@@ -34,6 +36,149 @@ impl<'w> EntityWorldMut<'w> {
...
+        }
+
+        let id = self.id();
+        self.world_scope(|world| {
+            for related in relations_to_remove {
+                world.entity_mut(related).remove::<R>();
+            }
+
+            for related in potential_relations {
+                world.entity_mut(related).insert(R::from(id));
```
villor, line 61: I think you could also optimize this a bit with a similar approach as in `replace_related_with_difference`, using the hook mode to avoid unnecessary archetype moves.

villor (changes requested):
Nice job on the optimized version! 👍  Really liking the `debug_assertions` too.

Still have some concerns about the ordering for `replace_related`. And some minor perf comments etc.

crates/bevy_ecs/src/relationship/related_methods.rs
```diff
@@ -34,6 +36,149 @@ impl<'w> EntityWorldMut<'w> {
...
+            }
+        });
+
+        self
+    }
```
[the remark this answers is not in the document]
Brezak: I totally forgot about this method while implementing `replace_related_with_difference`. I've fixed it and included you test too.

crates/bevy_ecs/src/relationship/related_methods.rs
Brezak: Good idea.

Brezak:
This should be done now. There's still some performance left on the table (see #18300) but otherwise everything works and tests pass.

One thing we should do is write down all the assumptions we make about relationship hooks. They're currently publicly accessible and a motivated user could override them and make them do some wild stuff. Alternatively we could add `#[doc(hidden)]` to them.

villor (approved):
LGTM 👍

villor:
> One thing we should do is write down all the assumptions we make about relationship hooks. They're currently publicly accessible and a motivated user could override them and make them do some wild stuff. Alternatively we could add `#[doc(hidden)]` to them.

I don't think hooks are overridable. And the derive macro [stops you](https://github.com/bevyengine/bevy/blob/adbb53b87f146b8750cb932ca4deb4f875d3e6b6/crates/bevy_ecs/macros/src/component.rs#L115) from adding hooks that are already used for relationship components.

Brezak:
> I don't think hooks are overridable. And the derive macro [stops you](https://github.com/bevyengine/bevy/blob/adbb53b87f146b8750cb932ca4deb4f875d3e6b6/crates/bevy_ecs/macros/src/component.rs#L115) from adding hooks that are already used for relationship components.

We allow users to impl `Component` without the macro. We even have an [example](https://github.com/bevyengine/bevy/blob/adbb53b87f146b8750cb932ca4deb4f875d3e6b6/examples/ecs/component_hooks.rs#L31) doing just that. And if a user is doing that they would also need to manually implement the `Relationship` trait at which point they can override whatever they want.

crates/bevy_ecs/src/hierarchy.rs
```diff
@@ -179,6 +179,34 @@ impl<'w> EntityWorldMut<'w> {
...
+    /// Replaces all the related children with a new set of children.
+    ///
+    /// # Warning
+    ///
+    /// Failing to maintain the functions invariants may lead to erratic engine behavior including random crashes.
+    /// Refer to [`Self::replace_related_with_difference`] for a list of these invariants.
+    ///
+    /// # Panics
+    ///
+    /// Panics when debug assertions are enable, an invariant is are broken and the command is executed.
```
alice-i-cecile, line 196:
```suggestion
    /// Panics when debug assertions are enabled if an invariant is is broken and the command is executed.
```

crates/bevy_ecs/src/hierarchy.rs
```diff
@@ -230,6 +258,34 @@ impl<'a> EntityCommands<'a> {
...
+    /// Replaces all the related entities with a new set of entities.
+    ///
+    /// # Warning
+    ///
+    /// Failing to maintain the functions invariants may lead to erratic engine behavior including random crashes.
+    /// Refer to [`EntityWorldMut::replace_related_with_difference`] for a list of these invariants.
+    ///
+    /// # Panics
+    ///
+    /// Panics when debug assertions are enable, an invariant is are broken and the command is executed.
```
alice-i-cecile, line 275:
```suggestion
    /// Panics when debug assertions are enabled if an invariant is is broken and the command is executed.
```

alice-i-cecile (approved):
Sensible and well-tested. I've fixed a grammar issue, but now I'm comfortable merging this.

PR #18059, opened 2025-02-26
Create `bevy_label` crate

### LINKED ISSUE #17713: Pull `bevy_ecs::label` out into its own separate crate

The label infrastructure that `bevy_ecs` provides is not actually dependent on anything ECS-related, and could find use as a separate crate from `bevy_ecs`.

Introduce a `bevy_label` crate that holds everything currently defined in `bevy_ecs::label`.

Leave it as-is.

hukasu:
# Objective

Closes #17713 

## Solution

Create a crate `bevy_label` and update re-exports of `DynEq` on `bevy_ecs`, `bevy_app`, and `bevy_render`.

## Testing

`cargo run -p ci`
`cargo check -p bevy_ecs --no-default-features --features edge_executor,critical-section,bevy_debug_stepping,bevy_reflect --target x86_64-unknown-none`

## Migration Guide

Moved `bevy_ecs::label` to `bevy_label` and `bevy_ecs::intern` to `bevy_label::intern`

- [x] Merge conflicts


mockersf:
#17713 lacks motivation apart from "it can be done". What advantages to you expect from this?

crates/bevy_label/src/lib.rs
```diff
@@ -1,5 +1,14 @@
 //! Traits used by label implementations
```
bushrat011899, line 1: If we're going to pull this out, this documentation should be far more fleshed out with some examples as to _how_ people are supposed to use this crate.

crates/bevy_label/src/lib.rs
```diff
@@ -1,5 +1,14 @@
...
 
+#![no_std]
+
+pub mod intern;
+
+#[cfg(feature = "std")]
+extern crate std;
```
bushrat011899, line 8: This crate actually has zero use for `std`, so the `std` feature and this `extern crate std;` statement can be entirely removed.

bushrat011899 (changes requested):
Documentation and use of `std` need to be changed IMO.

Regarding whether we _should_ do this: the most popular internment crate, [`internment`](https://crates.io/crates/internment), currently has zero support for `no_std`, but is otherwise mature and widely used in the Rust ecosystem already. Either:

1. We should polish this crate further, potentially trying to surpass `internment` and make `bevy_label` a fully standalone crate (e.g., `disqualified`). Or;
2. We should abandon Bevy's interning solution and improve `internment` based on our needs (e.g., `no_std` support).

I _do_ think there is merit in trying to remove the labelling system from `bevy_ecs`; it's an already _very_ large crate with a lot of responsibilities across a wide range of domains. Anything which _can_ be split out without sacrificing DX should be pursued by virtue of making `bevy_ecs` itself easier to maintain. The sheer quantity of items in `bevy_ecs` which are `pub(crate)` alone makes it extremely daunting to make breaking changes unless you're well versed in how the whole crate works.

alice-i-cecile:
I think my preference here is to spin out the functionality into its own Bevy-org crate, or improve an existing one enough that we can pull a dependency on it. This is useful to the broader ecosystem, loosely coupled and pretty stable.

hukasu:
> https://github.com/bevyengine/bevy/issues/17713 lacks motivation apart from "it can be done". What advantages to you expect from this?

@ItsDoot can you answer that?

hukasu:
Also, discursion moves faster as a PR that as an issue, the issue existed for 3 weeks with no discursion whatsoever

crates/bevy_label/src/lib.rs
hukasu: done

ItsDoot:
> I _do_ think there is merit in trying to remove the labelling system from `bevy_ecs`; it's an already _very_ large crate with a lot of responsibilities across a wide range of domains. Anything which _can_ be split out without sacrificing DX should be pursued by virtue of making `bevy_ecs` itself easier to maintain. The sheer quantity of items in `bevy_ecs` which are `pub(crate)` alone makes it extremely daunting to make breaking changes unless you're well versed in how the whole crate works.

FWIW, this^ is the motivation I was thinking of when making the issue. Whether that is worthy-enough motivation to split it is up to someone else 👍

alice-i-cecile:
>  Whether that is worthy-enough motivation to split it is up to someone else 👍

Someone else here: that's worthy enough motivation to me ;)

bushrat011899:
As an FYI, [`internment` has accepted my `no_std` PR](https://github.com/droundy/internment/pull/61), which opens up the possibility of outsourcing that aspect of labels entirely.

janhohenheim:
Triage: has merge conflicts
@hukasu  do you want to update this PR or should I tag it `S-Needs-Adoption`? :)

hukasu:
I can update

mockersf:
too many conflicts, not enough interest

PR #18060, opened 2025-02-26
Improve `bevy_reflect` `no_std` support

### LINKED ISSUE #18051: bevy_reflect: no `no_std` support when the `functions` feature is enabled
## Bevy version

The release number or commit hash of the version you're using.

Bevy main `11db717`

## What you did

I'm trying to use `bevy_reflect` in a project and this project requires `no_std` support. the latest `bevy_reflect` release `v0.15.3` has `no_std` support, but not really, since the `#[derive(Reflect)]` uses `::std::boxed::Box<..>` in the trait implementation.

After some search and looking for the code on github, I found that it actually now supports `no_std`, However, enabling the `functions` feature makes it not compiling without `std` feature enabled as it still uses `std` here: https://github.com/bevyengine/bevy/blob/11db71727d25bdbe86f709574b531cab686d0e0a/crates/bevy_reflect/src/func/registry.rs#L5

## What went wrong

1. `bevy_reflect` has `no_std` support.
2. However, the `func` module does not fully work when `std` feature is disabled.

bushrat011899:
# Objective

- Fixes #18051

## Solution

- Switched function registry to use `bevy_platform_support::sync` instead of `std::sync`
- Also documented that `debug_stack` (and transitively `debug`) require `std`
- Also removed `std` from `wgpu-types` since that crate is now `no_std` compatible

## Testing

- `cargo check -p bevy_reflect --no-default-features --features critical-section,documentation,glam,glam/libm,smallvec,wgpu-types,functions --target thumbv6m-none-eabi`

---

## Notes

With these changes, `bevy_reflect`'s feature support looks like this:

| Feature | `no_std` (Before) | `no_std` (After) | Notes |
| - | - | - | - |
| `default` | ❌ | ❌ | |
| `std` | ❌ | ❌ | |
| `documentation` | ✔️ | ✔️ | |
| `functions` | ❌ | ✔️ | |
| `critical-section` | ✔️ | ✔️ | |
| `bevy` | ✔️* | ✔️* | Partial due to `smol_str` |
| `debug` | ❌ | ❌ | |
| `debug_stack` | ❌ | ❌ | |
| `petgraph` | ❌ | ❌ | |
| `glam` | ✔️* | ✔️* | Requires enabling `glam/libm` |
| `smallvec` | ✔️ | ✔️ | |
| `uuid` | ✔️* | ✔️* | Requires setting up `getrandom` backend |
| `wgpu-types ` | ❌ | ✔️ | |
| `smol_str` | ✔️* | ✔️* | Partial due to `smol_str` not supporting `portable-atomic` |

### DIFF
--- a/crates/bevy_reflect/Cargo.toml
+++ b/crates/bevy_reflect/Cargo.toml
@@ -29,7 +29,7 @@ bevy = ["smallvec", "smol_str"]
 debug = ["debug_stack"]
 
 ## When enabled, keeps track of the current serialization/deserialization context for better error messages
-debug_stack = []
+debug_stack = ["std"]
 
 # Integrations
 
@@ -46,7 +46,7 @@ smallvec = ["dep:smallvec"]
 uuid = ["dep:uuid"]
 
 ## Adds reflection support to `wgpu-types` types.
-wgpu-types = ["dep:wgpu-types", "std"]
+wgpu-types = ["dep:wgpu-types"]
 
 # Platform Compatibility
 
@@ -62,6 +62,7 @@ std = [
   "smol_str?/std",
   "uuid?/std",
   "bevy_platform_support/std",
+  "wgpu-types?/std",
 ]
 
 ## `critical-section` provides the building blocks for synchronization primitives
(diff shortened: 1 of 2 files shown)


shekohex (approved):
It works! Thank you for the quick fix!

BenjaminBrienen (approved):

MrGVSV (approved):

PR #18063, opened 2025-02-26
Add per-pixel iterators to `Image`

### LINKED ISSUE #17941: Implement the Iterator trait for Image

If a texture has it's transparent areas encoded as a mock color. It'd be nice to iterate through the pixels to replace the mock color with transparency

bevy::prelude Struct Image should implement the Iterator trait

Right now to iterate through all the pixels of a texture you have to make a for loop with 2 indexes. Iterator doesn't have any overhead and is easier to read and write.

## Additional context

Imagine the green is the mock color and is supposed to be replaced with a 0 alpha value (absolute transparency):
Image

grind086:
# Objective

At the moment, the only way to get or set the value of a pixel in an `Image` by `Color` is via `get_color_at`/`set_color_at` using the pixel coordinates. This PR makes those functions available via iterators.

Fixes #17941

## Solution

Create the iterators `Pixels` and `PixelsMut` (created with `Image::pixels` and `Image::pixels_mut` respectively) which are thin wrappers around `core::slice::ChunksExact`/`Mut` and `TextureFormat`. These iterators return `Pixel` and `PixelMut` elements, which provide access to the raw bytes of the pixel as well as utility functions for working with the pixel value as a `Color`.

## Testing

Added a test, and some examples that run doctests.

---

## Showcase

```rust
fn chroma_key(image: &mut Image, key_bytes: &[u8], replace_bytes: &[u8]) {
    for mut px in image.pixels_mut() {
        if px.bytes() == key_bytes {
            px.copy_from_bytes(replace_bytes);
        }
    }
} 

fn lighten_image(image: &mut Image, amount: f32) -> Result<(), TextureAccessError> {
    for mut px in image.pixels_mut() {
        px.try_edit_color(|c| c.lighter(0.1))?;
    }
    Ok(())
}

```


crates/bevy_image/src/image.rs
```diff
@@ -1054,6 +1054,60 @@ impl Image {
         offset.map(|start| &mut data[start..(start + len)])
     }
 
+    /// Returns an iterator over the image's pixels.
```
mockersf, line 1057: column-row or row-column? starting from which corner?

grind086: How does "The iteration order is x = 0 to width then y = 0 to height then z = 0 to depth." sound?

IceSentry (approved):

PR #18067, opened 2025-02-27
Add `IntoSystem::with_input` and `::with_input_from` system wrappers

ItsDoot:
# Objective

Originally [provided as a solution to a user's problem in Discord](https://discord.com/channels/691052431525675048/1247654592838111302/1344431131277394042), library authors might find the need to present user-registered systems with system-specific data. Typically `Local<T>` is used for this type of thing, but its not generally feasible or possible to configure/set the underlying `T` data for locals. Alternatively, we can use `SystemInput` to pass the data.

## Solution

- Added `IntoSystem::with_input`: Allows system-specific data to be passed in explicitly.
- Added `IntoSystem::with_input_from`: Allows system-specific data to be created at initialization time via `FromWorld`.

## Testing

Added two new tests, testing each of `with_input` and `with_input_from`.

### DIFF
--- a/crates/bevy_ecs/src/system/schedule_system.rs
+++ b/crates/bevy_ecs/src/system/schedule_system.rs
@@ -5,8 +5,8 @@ use crate::{
     component::{ComponentId, Tick},
     error::Result,
     query::{Access, FilteredAccessSet},
-    system::{input::SystemIn, BoxedSystem, System},
-    world::{unsafe_world_cell::UnsafeWorldCell, DeferredWorld, World},
+    system::{input::SystemIn, BoxedSystem, System, SystemInput},
+    world::{unsafe_world_cell::UnsafeWorldCell, DeferredWorld, FromWorld, World},
 };
 
 use super::{IntoSystem, SystemParamValidationError};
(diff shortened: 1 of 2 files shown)


crates/bevy_ecs/src/system/schedule_system.rs
```diff
@@ -116,5 +116,224 @@ impl<S: System<In = (), Out = ()>> System for InfallibleSystemWrapper<S> {
...
+{
+    /// Wraps the given system with the given input value.
+    pub fn new<M>(system: impl IntoSystem<S::In, S::Out, M, System = S>, value: T) -> Self {
+        Self {
+            system: IntoSystem::into_system(system),
+            value,
+        }
+    }
+
+    /// Returns a reference to the input value.
```
chescock, line 142: Why expose `value` but not `system`?

chescock (approved):
Yeah, this looks handy!  I've also used closure captures for this sort of thing, but that has other limitations.

MalekiRe (approved):

crates/bevy_ecs/src/system/schedule_system.rs
ItsDoot: Incidentally; I needed access to the value for unit testing

alice-i-cecile:
@ItsDoot I'm on board with this after peeking at @MalekiRe's motivation. Can you resolve merge conflicts for me?

Issue #16744, opened 2024-12-10
OpenRPC support in Bevy Remote Protocol


Leinnan:
Provide a `rpc.discover` endpoint in Bevy Remote Protocol that will return the OpenRPC schema with available methods. https://spec.open-rpc.org/

It will provide two main benefits:
- discoverabbility of existing methods
- API that allows easy generation of methods into language of choice without need for handcrafted code, not limited to Rust: https://github.com/open-rpc/generator

Leinnan:
I did attempt to make it but due to personal reasons I am not sure if I will have time and energy to do it in coming months. 
My initial idea was to make requirement for both BRP methods arguments and results to be reflect types but that does not work well with `BrpInsertParams` which has hashmap of serde_json values. On one hand I don't want to reinvent the things that could be already provided by reflection system and on the other I don't have enough experience with reflection system in Bevy ATM. 

If anyone else finds this idea interesting enough and wants to grab it and make it I will be more than happy to provide some feedback if wanted.

PR #18068, opened 2025-02-27
RPC Discover endpoint with basic informations

Leinnan:
# Objective

It does not resolves issue for full support for OpenRPC for `bevy_remote`, but it is a first step in that direction. Connected to the #16744 issue.

## Solution

- Adds `rpc.discover` endpoint to the bevy_remote which follows https://spec.open-rpc.org/#openrpc-document For now in methods array only the name, which is the endpoint address is populated. 
- Moves json_schema structs into new module inside `bevy_remote`. 

## Testing

Tested the commands by running the BRP sample( cargo run --example server --features="bevy_remote") and with these curl command:

```sh
curl -X POST -d '{ "jsonrpc": "2.0", "id": 1, "method": "rpc.discover"}' 127.0.0.1:15702 | jq .
```
The output is: 
```json
{
  "jsonrpc": "2.0",
  "id": 1,
  "result": {
    "info": {
      "title": "Bevy Remote Protocol",
      "version": "0.16.0-dev"
    },
    "methods": [
      {
        "name": "bevy/mutate_component",
        "params": []
      },
      {
        "name": "bevy/insert",
        "params": []
      },
      {
        "name": "bevy/get",
        "params": []
      },
      {
        "name": "bevy/spawn",
        "params": []
      },
      {
        "name": "bevy/get+watch",
        "params": []
      },
      {
        "name": "bevy/destroy",
        "params": []
      },
      {
        "name": "bevy/list",
        "params": []
      },
      {
        "name": "bevy/mutate_resource",
        "params": []
      },
      {
        "name": "bevy/reparent",
        "params": []
      },
      {
        "name": "bevy/registry/schema",
        "params": []
      },
      {
        "name": "bevy/get_resource",
        "params": []
      },
      {
        "name": "bevy/query",
        "params": []
      },
      {
        "name": "bevy/remove_resource",
        "params": []
      },
      {
        "name": "rpc.discover",
        "params": []
      },
      {
        "name": "bevy/insert_resource",
        "params": []
      },
      {
        "name": "bevy/list_resources",
        "params": []
      },
      {
        "name": "bevy/remove",
        "params": []
      },
      {
        "name": "bevy/list+watch",
        "params": []
      }
    ],
    "openrpc": "1.3.2",
    "servers": [
      {
        "name": "Server",
        "url": "127.0.0.1:15702"
      }
    ]
  }
}
```


crates/bevy_remote/src/builtin_methods.rs
```diff
@@ -806,6 +815,38 @@ pub fn process_remote_spawn_request(In(params): In<Option<Value>>, world: &mut W
...
+/// Handles a `rpc.discover` request coming from a client.
+pub fn process_remote_list_methods_request(
+    In(_params): In<Option<Value>>,
+    world: &mut World,
+) -> BrpResult {
+    let remote_methods = world.resource::<crate::RemoteMethods>();
+    let servers = match (
+        world.get_resource::<crate::http::HostAddress>(),
+        world.get_resource::<crate::http::HostPort>(),
+    ) {
```
mweatherley, line 827: This is a very minor nit, but I guess it bothers me slightly that this isn't transparent to the transport used for BRP; I think ideally in the future (e.g. when we eventually include more detailed method information in the OpenRPC document) we'd have the server info be separately registered by the transport plugin in some kind of resource in the internals. 

Just a thought!

mweatherley (approved):

Leinnan:
@alice-i-cecile Something missing or is it ready for merge?

alice-i-cecile:
This needs a second review: I would, but I've never worked with this flavor of thing before.

crates/bevy_remote/src/schemas/open_rpc.rs
```diff
@@ -0,0 +1,100 @@
...
+    pub description: Option<String>,
+    /// A collection of custom extension fields.
+    #[serde(flatten)]
+    pub extensions: HashMap<String, serde_json::Value>,
+}
+
+impl Default for InfoObject {
+    fn default() -> Self {
+        Self {
+            title: "Bevy OpenRPC".to_owned(),
```
villor, line 42:
```suggestion
            title: "Bevy Remote Protocol".to_owned(),
```
I think this makes more sense as a title for the API itself.

crates/bevy_remote/src/builtin_methods.rs
```diff
@@ -806,6 +815,38 @@ pub fn process_remote_spawn_request(In(params): In<Option<Value>>, world: &mut W
...
+            name: "Server".to_owned(),
+            url: url.0.to_string(),
+            ..default()
+        }]),
+        _ => None,
+    };
+    let doc = OpenRpcDocument {
+        info: Default::default(),
+        methods: remote_methods.into(),
+        openrpc: "1.0.0-rc1".to_owned(),
```
villor, line 843: How come we are using `1.0.0-rc1` spec instead of the latest [`1.3.2`](https://spec.open-rpc.org/)?

crates/bevy_remote/src/schemas/open_rpc.rs
```diff
@@ -0,0 +1,100 @@
...
+    pub summary: Option<String>,
+    /// An optional detailed description of the method.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub description: Option<String>,
+    // /// Parameters for the RPC method
+    // #[serde(default)]
+    // pub params: Vec<Parameter>,
+    // /// The expected result of the method
+    // #[serde(skip_serializing_if = "Option::is_none")]
+    // pub result: Option<Parameter>,
```
villor, line 83: Any particular reason these were left out?
`params` is required in OpenRPC `1.3.2`.
`result`is required in OpenRPC `1.0.0-rc1`.

Maybe these could be included as `$ref`s using some of the `json_schema` code? 🤔

villor (changes requested):
Overall happy with this kind of functionality! 🥳  Helps both with discoverability and things like generating API clients.

Curious about the use of `1.0.0-rc1` spec instead of the latest spec.

The information included for the methods is pretty sparse currently. It would be nice (maybe in a follow up) to extend it with more stuff, possibly by extending `RemoteMethods` to hold more metadata like:
- Summary
- Description
- etc

crates/bevy_remote/src/schemas/open_rpc.rs
Leinnan: Unfortunately it wasn't that easy: https://github.com/bevyengine/bevy/issues/16744#issuecomment-2572730182

villor: Right, yeah it's a bit tricky that the methods are serde-based rather than bevy_reflect, but it also makes sense for usecases with non-reflectable data 🤔 

There is https://github.com/GREsau/schemars which can generate JSON schemas for serde types. Not sure if we'd want to bring in another dep though.

I say include a (possibly empty) `params` either way though, just to be spec-compliant. There might be clients that rely on it being present.

villor: Another route could be to add manual schemas for the BRP types, which we already kind of do in doc comments: https://dev-docs.bevyengine.org/src/bevy_remote/lib.rs.html#99

Ideally those would not be duplicated though, so if there was a way to write them in one place and have them available _both_ through OpenRPC and in bevy docs that would be nice.

Leinnan: Argument about including `params` makes sense, done :)

Leinnan: To have a full RPC Discover support the stuff you've described would be great. For the basic one I am aiming on atm what this PR provides is enough IMHO. But I would love to see it with proper `params` support.

villor: That's fair, this is a great start! Approving now 😄

villor (approved):

alice-i-cecile:
@Leinnan ping me when this is passing CI and I'll get this merged for you.

Leinnan:
@alice-i-cecile ping ;)

PR #18076, opened 2025-02-27
Recursive run_system

### LINKED ISSUE #18030: One shot system cannot recursively call itself
## Bevy version

0.15.3

## What you did

I was creating a effect system that run one shot system, and can take effect as parameters to run them from that one shot system (which mean i can call the same one shot system from the same system, but it's not rly recursive)

## What went wrong
I discovered that you cannot call X one shot system from the same system

## Isolated minimal example

```rust
use bevy::{ecs::system::SystemId, prelude::*};

fn main() {
    let mut app = App::new();
    let id = app.register_system(one_shot_system);

    app.add_plugins(DefaultPlugins);
    
    app.insert_resource(TestRes {
        id: id,
        counter: 5
    });

    app.add_systems(Startup, test_startup);
    app.run();
}

pub fn test_startup(mut commands: Commands, test: Res<TestRes>) {
    commands.run_system(test.id);
}

#[derive(Resource)]
pub struct TestRes {
    id: SystemId,
    counter: u32
}

pub fn one_shot_system(mut commands: Commands, mut test: ResMut<TestRes>) {
    println!("{:?}", test.counter);
    test.counter = test.counter-1;

    if test.counter > 0 {
        commands.run_system(test.id);
    }
}
```

This only call one_shot_sytem once and the commands.run_system inside it do nothing.

Of course this is not rly my use case but it's the same bug in the end.

Apparently i can just go with observers since i was told they can recursivly call themself, which i tried to avoid since in my use case it's mean 500 entities, one for each effect), but in the end it's probably nothing and have 0 perf issue.

But i wanted to know if this was intended or not cause it's documented nowhere (or i'm blind, it's a posibility i just have reading issues).

I think this happens cause the system may be removed from the world while it's executed

SpecificProtagonist:
# Objective

Fixes #18030

## Solution

When running a one-shot system, requeue the system's command queue onto the world's command queue, then execute the later.

If running the entire command queue of the world is undesired, I could add a new method to `RawCommandQueue` to only apply part of it.

## Testing

See the new test.

---

## Showcase

```rust
#[derive(Resource)]
pub struct Test {
    id: SystemId,
    counter: u32,
}

let mut world = World::new();
let id = world.register_system(|mut commands: Commands, mut test: ResMut<Test>| {
    print!("{:?} ", test.counter);
    test.counter -= 1;
    if test.counter > 0 {
        commands.run_system(test.id);
    }
});
world.insert_resource(Test { id, counter: 5 });
world.run_system(id).unwrap();
```

```
5 4 3 2 1 
```

### DIFF
--- a/crates/bevy_ecs/src/system/system_registry.rs
+++ b/crates/bevy_ecs/src/system/system_registry.rs
@@ -199,11 +199,9 @@ impl World {
     /// This is different from [`RunSystemOnce::run_system_once`](crate::system::RunSystemOnce::run_system_once),
     /// because it keeps local state between calls and change detection works correctly.
     ///
+    /// Also runs any queued-up commands.
+    ///
     /// In order to run a chained system with an input, use [`World::run_system_with`] instead.
-    ///
-    /// # Limitations
-    ///
-    ///  - Stored systems cannot be recursive, they cannot call themselves through [`Commands::run_system`](crate::system::Commands).
     ///
     /// # Examples
     ///
@@ -291,9 +289,7 @@ impl World {
     /// Before running a system, it must first be registered.
     /// The method [`World::register_system`] stores a given system and returns a [`SystemId`].
     ///
-    /// # Limitations
-    ///
-    ///  - Stored systems cannot be recursive, they cannot call themselves through [`Commands::run_system`](crate::system::Commands).
+    /// Also runs any queued-up commands.
     ///
     /// # Examples
     ///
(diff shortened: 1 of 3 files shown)


alice-i-cecile:
Oh hey, I remember being annoyed by this and not being able to figure out how to fix it when I was working on the MVP one-shot systems. Definitely a bug.

chescock (approved):

crates/bevy_ecs/src/system/system_registry.rs
```diff
@@ -322,38 +318,45 @@ impl World {
...
+        // Return ownership of system trait object (if entity still exists)
         if let Ok(mut entity) = self.get_entity_mut(id.entity) {
             entity.insert::<RegisteredSystem<I, O>>(RegisteredSystem {
                 initialized,
                 system,
             });
         }
+
+        // Run any commands enqueued by the system
+        self.flush();
```
Carter0, line 359: I thought commands were only run at sync points? Not after any system has been run?

Carter0 (approved):
Minor question, but I think its good

crates/bevy_ecs/src/system/system_registry.rs
Carter0: All good I mostly asked for my own understanding.

PR #18077, opened 2025-02-27
Simplify Component derive macro

Bleachfuel:
# Objective

simplify several things in the component macro


crates/bevy_ecs/src/hierarchy.rs
```diff
@@ -94,12 +94,15 @@ use log::warn;
 )]
 #[relationship(relationship_target = Children)]
 #[doc(alias = "IsChild", alias = "Parent")]
-pub struct ChildOf(pub Entity);
+pub struct ChildOf {
```
alice-i-cecile, line 97: This shouldn't be appearing in the diff 🤔

alice-i-cecile:
Ah, looks like you need to rebase: that explains the weird diff.

Bleachfuel:
New commit #18083 is up

PR #18080, opened 2025-02-27
Remove ChildOf::get and Deref impl

cart:
# Objective

There are currently three ways to access the parent stored on a ChildOf relationship:

1. `child_of.parent` (field accessor)
2. `child_of.get()` (get function)
3. `**child_of` (Deref impl)

I will assert that we should only have one (the field accessor), and that the existence of the other implementations causes confusion and legibility issues. The deref approach is heinous, and `child_of.get()` is significantly less clear than `child_of.parent`.

## Solution

Remove `impl Deref for ChildOf` and `ChildOf::get`.

The one "downside" I'm seeing is that:

```rust
entity.get::<ChildOf>().map(ChildOf::get)
```
Becomes this:

```rust
entity.get::<ChildOf>().map(|c| c.parent)
```

I strongly believe that this is worth the increased clarity and consistency. I'm also not really a huge fan of the "pass function pointer to map" syntax. I think most people don't think this way about maps. They think in terms of a function that takes the item in the Option and returns the result of some action on it.

## Migration Guide

```rust
// Before
**child_of
// After
child_of.parent

// Before
child_of.get()
// After
child_of.parent

// Before
entity.get::<ChildOf>().map(ChildOf::get)
// After
entity.get::<ChildOf>().map(|c| c.parent)
```

### DIFF
--- a/crates/bevy_ecs/src/hierarchy.rs
+++ b/crates/bevy_ecs/src/hierarchy.rs
@@ -99,22 +99,6 @@ pub struct ChildOf {
     pub parent: Entity,
 }
 
-impl ChildOf {
-    /// Returns the parent entity, which is the "target" of this relationship.
-    pub fn get(&self) -> Entity {
-        self.parent
-    }
-}
-
-impl Deref for ChildOf {
-    type Target = Entity;
-
-    #[inline]
-    fn deref(&self) -> &Self::Target {
-        &self.parent
-    }
-}
-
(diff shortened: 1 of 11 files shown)


TimJentzsch (approved):

alice-i-cecile (approved):
You missed an accessor: [`Relationship::get`](https://dev-docs.bevyengine.org/bevy/ecs/relationship/trait.Relationship.html) :)

alice-i-cecile:
Feel free to merge once CI is green :)

PR #18082, opened 2025-02-27
Make Query::single (and friends) return a Result

alice-i-cecile:
# Objective

As discussed in #14275, Bevy is currently too prone to panic, and makes the easy / beginner-friendly way to do a large number of operations just to panic on failure.

This is seriously frustrating in library code, but also slows down development, as many of the `Query::single` panics can actually safely be an early return (these panics are often due to a small ordering issue or a change in game state.

More critically, in most "finished" products, panics are unacceptable: any unexpected failures should be handled elsewhere. That's where the new

With the advent of good system error handling, we can now remove this.

Note: I was instrumental in a) introducing this idea in the first place and b) pushing to make the panicking variant the default. The introduction of both `let else` statements in Rust and the fancy system error handling work in 0.16 have changed my mind on the right balance here.

## Solution

1. Make `Query::single` and `Query::single_mut` (and other random related methods) return a `Result`.
2. Handle all of Bevy's internal usage of these APIs.
3. Deprecate `Query::get_single` and friends, since we've moved their functionality to the nice names. 
4. Add detailed advice on how to best handle these errors.

Generally I like the diff here, although `get_single().unwrap()` in tests is a bit of a downgrade.

## Testing

I've done a global search for `.single` to track down any missed deprecated usages.

As to whether or not all the migrations were successful, that's what CI is for :)

## Future work

~~Rename `Query::get_single` and friends to `Query::single`!~~

~~I've opted not to do this in this PR, and smear it across two releases in order to ease the migration. Successive deprecations are much easier to manage than the semantics and types shifting under your feet.~~ 

Cart has convinced me to change my mind on this; see https://github.com/bevyengine/bevy/pull/18082#discussion_r1974536085.

## Migration guide

`Query::single`, `Query::single_mut` and their `QueryState` equivalents now return a `Result`. Generally, you'll want to:

1. Use Bevy 0.16's system error handling to return a `Result` using the `?` operator.
2. Use a `let else Ok(data)` block to early return if it's an expected failure.
3. Use `unwrap()` or `Ok` destructuring inside of tests.

The old `Query::get_single` (etc) methods which did this have been deprecated.

### DIFF
--- a/crates/bevy_ecs/src/query/builder.rs
+++ b/crates/bevy_ecs/src/query/builder.rs
@@ -33,7 +33,7 @@ use super::{FilteredAccess, QueryData, QueryFilter};
 ///     .build();
 ///
 /// // Consume the QueryState
-/// let (entity, b) = query.single(&world);
+/// let (entity, b) = query.get_single(&world).unwrap();
 /// ```
 pub struct QueryBuilder<'w, D: QueryData = (), F: QueryFilter = ()> {
     access: FilteredAccess<ComponentId>,
--- a/crates/bevy_ecs/src/query/builder.rs
+++ b/crates/bevy_ecs/src/query/builder.rs
@@ -33,7 +33,7 @@ use super::{FilteredAccess, QueryData, QueryFilter};
 ///     .build();
 ///
 /// // Consume the QueryState
-/// let (entity, b) = query.get_single(&world).unwrap();
+/// let (entity, b) = query.single(&world).unwrap();
 /// ```
 pub struct QueryBuilder<'w, D: QueryData = (), F: QueryFilter = ()> {
     access: FilteredAccess<ComponentId>,
@@ -297,13 +297,13 @@ mod tests {
             .with::<A>()
             .without::<C>()
             .build();
-        assert_eq!(entity_a, query_a.get_single(&world).unwrap());
+        assert_eq!(entity_a, query_a.single(&world).unwrap());
 
         let mut query_b = QueryBuilder::<Entity>::new(&mut world)
             .with::<A>()
             .without::<B>()
             .build();
-        assert_eq!(entity_b, query_b.get_single(&world).unwrap());
+        assert_eq!(entity_b, query_b.single(&world).unwrap());
     }
 
     #[test]
@@ -319,13 +319,13 @@ mod tests {
             .with_id(component_id_a)
             .without_id(component_id_c)
             .build();
-        assert_eq!(entity_a, query_a.get_single(&world).unwrap());
+        assert_eq!(entity_a, query_a.single(&world).unwrap());
 
         let mut query_b = QueryBuilder::<Entity>::new(&mut world)
             .with_id(component_id_a)
             .without_id(component_id_b)
             .build();
-        assert_eq!(entity_b, query_b.get_single(&world).unwrap());
+        assert_eq!(entity_b, query_b.single(&world).unwrap());
     }
 
     #[test]
@@ -385,7 +385,7 @@ mod tests {
             .data::<&B>()
             .build();
 
-        let entity_ref = query.get_single(&world).unwrap();
+        let entity_ref = query.single(&world).unwrap();
 
         assert_eq!(entity, entity_ref.id());
 
@@ -408,7 +408,7 @@ mod tests {
             .ref_id(component_id_b)
             .build();
 
-        let entity_ref = query.get_single(&world).unwrap();
+        let entity_ref = query.single(&world).unwrap();
 
         assert_eq!(entity, entity_ref.id());
 
--- a/crates/bevy_ecs/src/query/builder.rs
+++ b/crates/bevy_ecs/src/query/builder.rs
@@ -297,13 +297,13 @@ mod tests {
             .with::<A>()
             .without::<C>()
             .build();
-        assert_eq!(entity_a, query_a.single(&world));
+        assert_eq!(entity_a, query_a.get_single(&world).unwrap());
 
         let mut query_b = QueryBuilder::<Entity>::new(&mut world)
             .with::<A>()
             .without::<B>()
             .build();
-        assert_eq!(entity_b, query_b.single(&world));
+        assert_eq!(entity_b, query_b.get_single(&world).unwrap());
     }
 
     #[test]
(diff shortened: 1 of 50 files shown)


examples/stress_tests/many_text2d.rs
```diff
@@ -155,7 +155,9 @@ fn setup(mut commands: Commands, font: Res<FontHandle>, args: Res<Args>) {
 
 // System for rotating and translating the camera
 fn move_camera(time: Res<Time>, mut camera_query: Query<&mut Transform, With<Camera>>) {
-    let mut camera_transform = camera_query.single_mut();
+    let Ok(mut camera_transform) = camera_query.get_single_mut() else {
+        return;
+    };
```
bushrat011899, line 160: Could the signature for `move_camera` not be changed to return `Result`, allowing the use of `?` here too?

crates/bevy_ecs/src/query/state.rs
```diff
@@ -1603,8 +1603,12 @@ impl<D: QueryData, F: QueryFilter> QueryState<D, F> {
     /// [`get_single`](Self::get_single) to return a `Result` instead of panicking.
     #[track_caller]
     #[inline]
+    #[deprecated(
+        since = "0.16.0",
+        note = "Please use `get_single` instead, which returns a `Result` instead of panicking."
+    )]
     pub fn single<'w>(&mut self, world: &'w World) -> ROQueryItem<'w, D> {
```
bushrat011899, line 1610: Once these are removed, I think we should rename all the various `get_single` methods to take their `panic!` variant's names and then deprecate the old `get_X` methods as well.

crates/bevy_ecs/src/query/builder.rs
```diff
@@ -297,13 +297,13 @@ mod tests {
             .with::<A>()
             .without::<C>()
             .build();
-        assert_eq!(entity_a, query_a.single(&world));
+        assert_eq!(entity_a, query_a.get_single(&world).unwrap());
```
bushrat011899, line 300: Nit: I prefer to wrap the reference value in `Ok(...)` rather an `unwrap` in an `assert_eq`, as that will print a nicer error message. Has the same end-result of failing the test so it's not _that_ big of a deal though.

bushrat011899 (approved):
Good direction for the engine. I look forward to reclaiming those names so `get_single` can just be `single` in 0.17.

examples/stress_tests/many_text2d.rs
alice-i-cecile: Yes, but this is an example of a case where silently failing and just returning is always correct. You never want to panic here.

crates/bevy_ecs/src/query/state.rs
alice-i-cecile: 100% agree.

crates/bevy_ecs/src/query/builder.rs
alice-i-cecile: Yeah, mildly agree.

crates/bevy_ecs/src/system/query.rs
```diff
@@ -947,7 +947,7 @@ impl<'w, 's, D: QueryData, F: QueryFilter> Query<'w, 's, D, F> {
     ///     friends_query: Query<&Friends>,
```
cart, line 947: > I've opted not to do this in this PR, and smear it across two releases in order to ease the migration. Successive deprecations are much easier to manage than the semantics and types shifting under your feet.

Is this actually better? Feels to me like we're maximizing pain and confusion:

1. Everyone using `query.single()` is effectively forced to move to `query.get_single()`. They have to move to something that is longer and uglier, will still contending with the fact that it returns an error, either by manually unwrapping or adapting code, or by adjusting their code to use rust error handling. There will likely be negative sentiment because we've replaced something that used to feel nice and crisp with something that is not. 
3. Force people (previous `single()` users and existing `get_single()` users)  to rename `query.get_single()` to `query.single()`.

The alternative:

* Everyone using `query.single()` is broken (and sees a compiler error) and is immediately forced to contend with the fact that it returns a Result.  Everyone using `query.get_single()` gets a nice deprecation warning and moves to `single()`.


From a marketing / "public perception" standpoint, that allows Bevy 0.16 to be the "error handling update". You pay the cost of adapting to everything returning results, but in return you get vastly improved error handling _and_ the nice short APIs. By deferring, we're denying people one of the most important motivators for the switch (an ergonomic Results-everywhere system), and for many people they might not even know the "shorter method name" carrot is coming (reducing their happiness and trust in our decision making ... "why am I feeling this pain when its feels worse?").

alice-i-cecile: Hmm. I think that you've convinced me. This way allows users to put off fixing their problems for a cycle, but given that the migration is pretty straightforward I don't think that improves matters.

cart: I think the biggest downside of the "break immediately" approach is that we lose the chance to inform people about the new paradigm in the deprecation warning. I think after the first actual breakage encountered (where they identify that a result is returned), the rest of the migration is easy / on autopilot (ex: just add `unwrap()` or `?` whenever the error is encountered).

I think a combination of:
1. Migration guide with a nice front-and-center cohesive outline of the new paradigm 
2. Blog post hitting the same notes as (1)
3. Examples being updated to use error handling.
4. Fielding and managing confusion in our communities

Will largely mitigate the migration pain.

crates/bevy_ecs/src/query/state.rs
```diff
@@ -1597,56 +1597,114 @@ impl<D: QueryData, F: QueryFilter> QueryState<D, F> {
...
+    /// use bevy_ecs::prelude::*;
+    /// use bevy_ecs::query::QuerySingleError;
+    ///
+    /// #[derive(Component)]
+    /// struct A(usize);
+    ///
+    /// fn my_system(query: Query<&A>, mut commands: Commands) {
+    ///     match query.single() {
+    ///         Ok(a) => (), // Do something with `a`
+    ///         Err(err) => match err {
```
Carter0, line 1618: I like these docs!

Carter0 (approved):
Im a fan!

alice-i-cecile:
I still need to fix CI, but once that's done I'm going to call this good to merge.

Issue #14336, opened 2024-07-15
Allow specifying Event bubbling info as proc macro attributes


ItsDoot:
## What problem does this solve or what need does it fill?

Event bubbling info currently requires manual implementation:

```rust
#[derive(Component)]
struct MyEvent;

impl Event for MyEvent {
    type Traverse = Parent;
    const AUTO_PROPAGATE: bool = true;
}
```

## What solution would you like?

Allow the user to specify this information as proc macro attributes:

```rust
#[derive(Event)]
#[event(traverse = Parent, auto_propagate = true)]
struct MyEvent;
```

PR #18083, opened 2025-02-27
Improve derive(Event) and simplify macro code

Bleachfuel:
# Objective

simplify some code and improve Event macro

Closes https://github.com/bevyengine/bevy/issues/14336,

# Showcase

you can now write derive Events like so
```rust
#[derive(event)]
#[event(auto_propagate, traversal = MyType)]
struct MyEvent;
```


Bleachfuel:
Test that is failing to compile passes on my machine, but on CI it has a weird error :(

Bleachfuel:
Ready for review!

Bleachfuel:
According to Dtolnay, this is not desired behaviour.
```
#[relationship(relationship_target = Attachments<T>)]
```
Instead it should be 
```
#[relationship(relationship_target = Attachments::<T>)]
```

I could make it this way to make the relationship parsing nicer, but it makes it harder to implement it for users as the error is add `<T>` and not `::<T>`.

alice-i-cecile (approved):

fjkorf (approved):
This PR looks good. I'm not an expert in writing macros, but I was able to work through the changes and they make sense. The reformatted example is helpful and worked as expected.

alice-i-cecile:
@Bleachfuel I'm happy to merge this once merge conflicts are resolved.

ItsDoot:
This closes #14336, if you'd like to note that in the PR description.

Bleachfuel:
@alice-i-cecile The merge conflict caused visit entities to be implemtend for relationship_targets? is this correct behaviour? Otherwise I will remove it. Merge conflicts are done btw :)!

alice-i-cecile (approved):

PR #18084, opened 2025-02-28
Deprecate panicking `.resource` methods

alice-i-cecile:
# Objective

This PR is a sibling to #18082, and almost all of the motivation there applies here. These methods are much more commonly used in commands, tests and exclusive systems though, and deserve their own judgement call.

Ultimately part of #14275.

## Solution

1. Deprecate all of the panicking resource methods I can find.

I'm less confident about this PR, and there's 796 warnings in my IDE, so I want to get consensus before burning time on migrating all of these.

TODO:

- [ ] fix deprecation warnings
- [ ] add docs?
- [] make `World::resource` and friends return a `Result` for nicer `?` ergonomics?

## Future Work

Move the non-panicking versions to the better names during the 0.17 cycle, just like in #18082.

## Migration Guide

`World::resource` and all related methods are now deprecated. Use `World::get_resource` and so on and handle the returned `Option`.

### DIFF
--- a/crates/bevy_ecs/src/world/deferred_world.rs
+++ b/crates/bevy_ecs/src/world/deferred_world.rs
@@ -361,6 +361,10 @@ impl<'w> DeferredWorld<'w> {
     /// Use [`get_resource_mut`](DeferredWorld::get_resource_mut) instead if you want to handle this case.
     #[inline]
     #[track_caller]
+    #[deprecated(
+        since = "0.16.0",
+        note = "Please use `get_resource_mut` instead, which returns an `Option` instead of panicking."
+    )]
     pub fn resource_mut<R: Resource>(&mut self) -> Mut<'_, R> {
         match self.get_resource_mut() {
             Some(x) => x,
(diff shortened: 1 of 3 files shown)


alice-i-cecile:
We could choose to use a lint for these methods and keep them around (see https://github.com/TheBevyFlock/bevy_cli/issues/269) and keep them around, but I prefer the solution here because:

1. Our defaults should be the right choice for finished games.
2. Less API = More Good.
3. Relying on external unshipped tooling sucks.

JMS55:
This is going to make a lot of rendering code worse :/ (check how often `.resource::<` is called). Why not keep both methods around? If we want to make the engine itself more resilient, then I think targeted audits to find places we shouldn't be panicking is a better idea.

alice-i-cecile:
Yeah, I think that's reasonable. Given that we're not running into many panics here (and Bevy already panics when you're missing a resource in a normal system) I think this is net-negative. Closing out.

PR #18086, opened 2025-02-28
changed marker struct to const generic parameter

### LINKED ISSUE #18085: Replace IsAligned marker struct with const generic in bevy_ptr

The `Ptr`, `PtrMut` and `OwningPtr` types use a marker struct to indicate whether the pointer is aligned. It may be more appropriate to use a const generic parameter instead.

Replace the `Ptr`, `PtrMut` and `OwningPtr` declarations with variations on the following:

```rust
pub struct Ptr<'a, const IS_ALIGNED: bool = true>(NonNull<u8>, PhantomData<&'a u8>);
```

Perhaps also replace use of the boolean literals true and false with `ALIGNED` and `UNALIGNED` for readability (`Ptr::<ALIGNED>` rather than `Ptr::<true>`)

```rust
pub const ALIGNED: bool = true;
pub const UNALIGNED: bool = false;
```

## Additional context

Supercedes #17988 and thus #18042

mysteriouslyseeing:
# Objective

Fixes #18085.

## Solution

Implementation as described in issue:

- Replaced `Aligned` and `Unaligned` structs with `ALIGNED` and `UNALIGNED` constants
- Changed `A: IsAligned = Aligned` to `const IS_ALIGNED: bool = ALIGNED` in `Ptr`, `PtrMut` and `OwningPtr`

## Testing

- Changes have not been tested, aside from ensuring that `cargo check --workspace --all-targets` does not error

---

## Migration Guide

- Replace all uses of `bevy_ptr::Aligned` and `bevy_ptr::Unaligned` in the context of `Ptr`, `PtrMut` and `OwningPtr` with `ALIGNED` and `UNALIGNED`, respectively.
- Replace generic type parameters using the now removed `IsAligned` trait with a const generic parameter of type bool.

### DIFF
--- a/crates/bevy_ecs/src/world/command_queue.rs
+++ b/crates/bevy_ecs/src/world/command_queue.rs
@@ -3,7 +3,7 @@ use crate::{
     world::{DeferredWorld, World},
 };
 use alloc::{boxed::Box, vec::Vec};
-use bevy_ptr::{OwningPtr, Unaligned};
+use bevy_ptr::{OwningPtr, UNALIGNED};
 use core::{
     fmt::Debug,
     mem::{size_of, MaybeUninit},
@@ -20,7 +20,7 @@ struct CommandMeta {
     ///
     /// Advances `cursor` by the size of `T` in bytes.
     consume_command_and_get_size:
-        unsafe fn(value: OwningPtr<Unaligned>, world: Option<NonNull<World>>, cursor: &mut usize),
+        unsafe fn(value: OwningPtr<UNALIGNED>, world: Option<NonNull<World>>, cursor: &mut usize),
 }
 
 /// Densely and efficiently stores a queue of heterogenous types implementing [`Command`].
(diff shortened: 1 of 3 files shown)


alice-i-cecile:
I like that the two options are exhaustively enumerated, but I don't like the const ALIGNED = true stuff. `Pointer<true>` is particularly egregious, and that's the obvious way to use these types. I don't think this is a net improvement after seeing the ultimate effect, but thanks for poking at this.

mysteriouslyseeing:
> Maybe with const generic enums and `Pointer<Aligned::Yes>` or something this would be better, but even that reads pretty poorly.

I think you would be able to do `pub use Aligned::Yes;` like how the standard library reexports `Some` and `None`. Then it would be `Pointer<Yes>` - so if you name things carefully (something like `IsAligned::Aligned`) you could keep the usage almost exactly the same as it is now.

mysteriouslyseeing:
> I don't think this is a net improvement after seeing the ultimate effect, but thanks for poking at this.

Makes sense! Good to close it then?

PR #18087, opened 2025-02-28
One-to-One Relationships

dmyyy:
# Objective

Minimal implementation of directed one-to-one relationships via implementing `RelationshipSourceCollection` for `Entity`.

Now you can do

```rust
#[derive(Component)]
#[relationship(relationship_target = Below)]
pub struct Above(Entity);

#[derive(Component)]
#[relationship_target(relationship = Above)]
pub struct Below(Entity);
```

## Future Work

It would be nice if the relationships could be fully symmetrical in the future - in the example above, since `Above` is the source of truth you can't add `Below` to an entity and have `Above` added automatically.

## Testing

Wrote unit tests for new relationship sources and  and verified adding/removing relationships maintains connection as expected.

### DIFF
--- a/crates/bevy_ecs/src/relationship/relationship_source_collection.rs
+++ b/crates/bevy_ecs/src/relationship/relationship_source_collection.rs
@@ -234,4 +234,31 @@ mod tests {
         let collection = rel_target.collection();
         assert_eq!(collection, &a);
     }
+
+    #[test]
+    fn one_to_one_relationships() {
+        #[derive(Component)]
+        #[relationship(relationship_target = Below)]
+        struct Above(Entity);
+
+        #[derive(Component)]
+        #[relationship_target(relationship = Above)]
+        struct Below(Entity);
+
+        let mut world = World::new();
+        let a = world.spawn_empty().id();
+        let b = world.spawn_empty().id();
+
+        world.entity_mut(a).insert(Above(b));
+        assert_eq!(a, world.get::<Below>(b).unwrap().0);
+
+        // remove below and above should be removed
+        world.entity_mut(b).remove::<Below>();
+        assert!(world.get::<Above>(a).is_none());
+
+        // remove above and below should be removed
+        world.entity_mut(a).insert(Above(b));
+        world.entity_mut(a).remove::<Above>();
+        assert!(world.get::<Below>(b).is_none());
+    }
 }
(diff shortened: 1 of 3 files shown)


crates/bevy_ecs/src/relationship/relationship_source_collection.rs
```diff
@@ -116,6 +116,32 @@ impl<const N: usize> RelationshipSourceCollection for SmallVec<[Entity; N]> {
...
+    fn with_capacity(_capacity: usize) -> Self {
+        Entity::PLACEHOLDER
+    }
+
+    fn add(&mut self, entity: Entity) {
+        *self = entity;
+    }
+
+    fn remove(&mut self, entity: Entity) {
+        if *self == entity {
```
alice-i-cecile, line 131: This definitely feels sus. What if we impl'ed this for `Option<Entity>` instead?

dmyyy: I did it that way at first. It felt weird to me that it's valid for the `Relationship Target` to be set to `None` while the Relationship still points to that entity.

Carter0 (approved):
I like it! Though I like the Option<entity> idea more personally

ItsDoot (approved):
LGTM. FWIW, I prefer the `Entity` approach because it better matches how its interacted with; assuming all invariants are upheld, users will always have access to a single valid entity.

dmyyy:
The `Option<Entity>` approach feels like a leaky abstraction - if I'm trying to make a one-to-one relationship I wouldn't be able to guess that I need to write it like this without looking through the source code which is a bit of a red flag IMO.

```rust
#[derive(Component)]
#[relationship(relationship_target = Below)]
pub struct Above(Entity);

#[derive(Component)]
#[relationship_target(relationship = Above)]
pub struct Below(Option<Entity>);
```

@alice-i-cecile Is it possible for this to be considered for 0.16?

Carter0:
> 

Yeah I see what y'all mean. There should always be a 1-1 relationship between the two entities as an invariant. 

Not something to consider for this PR perhaps, but I wonder if this is a code smell then. It sorta weird to use the word "collection" in the code when really it's just one item.

alice-i-cecile:
Yep, I'll make the call today in my merge train :) The 0.16 milestone isn't exhaustive: it just captures "stuff that I would consider blocking the release for".

alice-i-cecile:
> The `Option<Entity>` approach feels like a leaky abstraction - if I'm trying to make a one-to-one relationship I wouldn't be able to guess that I need to write it like this without looking through the source code which is a bit of a red flag IMO.
> 
> ```rust
> #[derive(Component)]
> #[relationship(relationship_target = Below)]
> pub struct Above(Entity);
> 
> #[derive(Component)]
> #[relationship_target(relationship = Above)]
> pub struct Below(Option<Entity>);
> ```

This comment convinced me. Merging.

Issue #18073, opened 2025-02-27
Bevy exposes access to the entire file system to scripts and/or mods.


Threadzless:
In PR #18023 it was noticed that bevy's `AssetServer` places no restrictions on which files the executable can load, meaning any scripting system or loaded mod could potentially harvest arbitrary amounts of user data.

This Issue is concerned with how best to address the security vulnerability in which mods or scripts could harvest arbitrary user data using bevy's built in file reader systems, such as, but not limited to, the `AssetPlugin`.

## Proposed solution

In line with the discussion on #18023, I propose the `AssetServer::load` have limits placed on which files it will load. These limits are that the file must meet one of the following criteria:
 - Is located inside the directory `AssetPlugin::file_path` AND asset processing is disabled
 - Is located inside the directory `AssetPlugin::processed_file_path` AND asset processing is enabled
 - Is located in a different `AssetSource`, either the default file path or the processed file path, depending on if asset processing is enabled.

In total, this should prevent the asset system from loading any files not in one of the folders designated as an asset source by the developers.

## Additional Work

The rest of the bevy code base should be searched for other instances of filesystem access. Where found, these plugins should either have comparable safety measures imposed, or have warnings added to the documentations of Events, Resources, States, SystemParams, etc which can result in any file read or write.

Additionally, I recommend adding warnings and safety recommendations to the bevy book around the topics of mod support, scripting systems, and asset loading.

alice-i-cecile:
I would prefer an "secure-by-default" approach to filesystem access. Docs alone are not adequate, and the default stance should be secure. Users who need to break these rules locally should encounter a nice error that says "you can't do this for security reasons! Go change this setting to enable it!"

Threadzless:
That makes sense.

By my searching of the code base, there are only a few places that the file system is accessed outside of the `bevy_asset` crate:
- `BevyManifest` (bevy_macro_utils/src/bevy_manifest.rs) which is only used by macro crates.
- The `headless_render` example.
- The `build-templated-pages` tool.

My search was done by looking for the text "std::fs", "std::path", and "async_fs", finding all uses of structs and functions those crates expose, then tracing their usage outwards until I hit something public facing.

Unless there's something I missed, only the `AssetPlugin` needs to know how to handle out-of-bounds directories, so a bool in the plugin struct should be sufficient to control that.

alice-i-cecile:
[...]

PR #18088, opened 2025-02-28
AssetServer out of bounds protection

Threadzless:
Addresses Issue #18073


viridia (commented):
I like the overall approach, but I think that the tests could be more comprehensive.

viridia (approved):

Threadzless:
I added more tests. Now an unrelated test fails on github, but succeeds on my machine. Idk what to do from here

crates/bevy_asset/src/lib.rs
```diff
@@ -258,6 +258,30 @@ pub struct AssetPlugin {
...
+}
+
+/// Determines how asset paths that go out of bounds are handled.
+///
+/// An out of bounds asset path is one that attempts to load a file that
+/// is not inside the default asset folder or another asset source. This usually
+/// means it is either an absolute address or it contains enough "../"s to
+/// exit the designated asset folder.
+///
+/// The bevy team strongly discourages using [`Allow`](OutOfBoundsMode::Allow) if your
```
kristoff3r, line 272: It feels weird to refer to "the bevy team", you can just say that it's discouraged.

crates/bevy_asset/src/lib.rs
```diff
@@ -258,6 +258,30 @@ pub struct AssetPlugin {
...
+/// app will include scripts or modding support, as it could allow allow arbitrary file
+/// access for malitious code.
```
kristoff3r, line 274:
```suggestion
/// access for malicious code.
```

crates/bevy_asset/src/lib.rs
```diff
@@ -258,6 +258,30 @@ pub struct AssetPlugin {
...
+#[derive(Clone, Default)]
+pub enum OutOfBoundsMode {
```
kristoff3r, line 276: While something being "out of bounds" is used for memory access, afaik it isn't standard terminology for this. I'd call it something like AssetSandboxMode. This would require the docs to be rewritten as they use the term a lot. The standard term for the problem is "path traversal vulnerability", which we might want to refer to as well.

crates/bevy_asset/src/path.rs
```diff
@@ -478,6 +478,25 @@ impl<'a> AssetPath<'a> {
             }
         })
     }
+
+    pub fn is_out_of_bounds(&self) -> bool {
```
kristoff3r, line 482: This could use a doc comment describing what it checks for / guarantees.

crates/bevy_asset/src/server/mod.rs
```diff
@@ -365,16 +421,56 @@ impl AssetServer {
...
+        'a,
+        A: Asset,
+        S: Settings,
+        G: Send + Sync + 'static,
+    >(
+        &self,
+        path: impl Into<AssetPath<'a>>,
+        settings: impl Fn(&mut S) + Send + Sync + 'static,
+        guard: G,
+    ) -> Handle<A> {
```
kristoff3r, line 447: I don't like the duplication of all of these functions. In almost all cases you'd set the mode globally and never need the other variant I think? And if we really want a way to opt out of the sandbox can't we just make `load_with_meta_transform` public instead of creating every combination of how it can be called as separate methods?

crates/bevy_asset/src/lib.rs
```diff
@@ -258,6 +258,30 @@ pub struct AssetPlugin {
...
     pub meta_check: AssetMetaCheck,
+
+    pub out_of_bounds_mode: OutOfBoundsMode,
+}
+
+/// Determines how asset paths that go out of bounds are handled.
+///
+/// An out of bounds asset path is one that attempts to load a file that
+/// is not inside the default asset folder or another asset source. This usually
+/// means it is either an absolute address or it contains enough "../"s to
```
kristoff3r, line 269:
```suggestion
/// means it is either an absolute path or it contains enough "../"s to
```

crates/bevy_asset/src/lib.rs
Threadzless: Changed to "Out of bounds" to Unapproved Path

crates/bevy_asset/src/server/mod.rs
Threadzless: If you have a scripting system, you would give that system access to AssetServer::load, and then use AssetServer::load_override. If UnapprovedPathMode (previously OutOfBoundsMode) is Deny, this would let the developers load whatever files, while preventing mods from doing the same.

We could get rid of all _override functions and expose load_with_meta_transform, but that would make using this security feature more confusing

crates/bevy_asset/src/lib.rs
```diff
@@ -1856,4 +1887,94 @@ mod tests {
...
+    text: "a",
+    dependencies: [],
+    embedded_dependencies: [],
+    sub_texts: [],
+)"#;
+
+        dir.insert_asset_text(Path::new(a_path), a_ron);
+
+        let mut app = App::new();
+        let (gated_memory_reader, gate_opener) = GatedReader::new(MemoryAssetReader { root: dir });
```
andriyDev, line 1905: None of these tests use the gate. We can just remove the GatedReader and just pass along the MemoryAssetReader directly.

crates/bevy_asset/src/lib.rs
```diff
@@ -258,6 +258,33 @@ pub struct AssetPlugin {
...
+#[derive(Clone, Default)]
+pub enum UnapprovedPathMode {
+    /// Unapproved asset loading is allowed. This is strongly discouraged.
+    Allow,
+    /// Panics if any asset load is unapproved, unless an override method is used, like
+    /// [`AssetServer::load_override`]
+    Deny,
+    /// Panics if any asset load is unapproved.
+    #[default]
+    Forbid,
```
andriyDev, line 287: Panicking is quite scary behaviour. I don't want to go too deep into the security aspect, but it would suck if some mod could just intentionally crash your game.

Can we instead just make these error logs and then just don't load the asset? It could lead to a broken game, but I think that's better than a crashing game.

crates/bevy_asset/src/server/mod.rs
```diff
@@ -310,7 +321,16 @@ impl AssetServer {
     /// The asset load will fail and an error will be printed to the logs if the asset stored at `path` is not of type `A`.
     #[must_use = "not using the returned strong handle may result in the unexpected release of the asset"]
     pub fn load<'a, A: Asset>(&self, path: impl Into<AssetPath<'a>>) -> Handle<A> {
-        self.load_with_meta_transform(path, None, ())
+        self.load_with_meta_transform(path, None, (), false)
+    }
+
+    /// Same as [`load`](AssetServer::load), but you can load assets from unaproved paths
```
andriyDev, line 327: `unaproved` -> `unapproved`

crates/bevy_asset/src/server/mod.rs
```diff
@@ -334,7 +354,20 @@ impl AssetServer {
         path: impl Into<AssetPath<'a>>,
         guard: G,
     ) -> Handle<A> {
-        self.load_with_meta_transform(path, None, guard)
+        self.load_with_meta_transform(path, None, guard, false)
+    }
+
+    /// Same as [`load`](AssetServer::load_acquire), but you can load assets from unaproved paths
```
andriyDev, line 360: `unaproved` -> `unapproved`

crates/bevy_asset/src/server/mod.rs
```diff
@@ -346,7 +379,30 @@ impl AssetServer {
...
-        self.load_with_meta_transform(path, Some(loader_settings_meta_transform(settings)), ())
+        self.load_with_meta_transform(
+            path,
+            Some(loader_settings_meta_transform(settings)),
+            (),
+            false,
+        )
+    }
+
+    /// Same as [`load`](AssetServer::load_with_settings), but you can load assets from unaproved paths
```
andriyDev, line 390: `unaproved` -> `unapproved`

andriyDev (changes requested):

crates/bevy_asset/src/lib.rs
Threadzless: Good idea

crates/bevy_asset/src/server/mod.rs
Threadzless: one of these days I'll learn to speel

Threadzless:
Now denied asset loads return `Handle::default()` and print an error. Documentation has been updated.

crates/bevy_asset/src/lib.rs
```diff
@@ -1902,7 +1902,7 @@ mod tests {
         dir.insert_asset_text(Path::new(a_path), a_ron);
 
         let mut app = App::new();
-        let (gated_memory_reader, gate_opener) = GatedReader::new(MemoryAssetReader { root: dir });
+        let (gated_memory_reader, _gate_opener) = GatedReader::new(MemoryAssetReader { root: dir });
```
andriyDev, line 1905: This is still not quite right. Just remove the `GatedReader::new` entirely, and I think it should work.

andriyDev (changes requested):

andriyDev (approved):

kristoff3r (approved):

alice-i-cecile:
Good design: I particularly appreciate the secure-by-default approach. Thanks to everyone involved.

PR #18091, opened 2025-02-28
testbed for UI

mockersf:
# Objective

- have a testbed for UI

## Solution

- move previous `ui` example to `full_ui`
- add a testbed ui with several scenes
- `ui_layout_rounding` is one of those scenes, so remove it as a standalone example

the previous `ui` / new `full_ui` is I think still useful as it has some things like scroll, debug ui that are not shown anywhere else


alice-i-cecile (approved):
Reasonable enough. It's good to see this progressing :)

examples/testbed/ui.rs
```diff
@@ -1,436 +1,533 @@
...
+                }
+            });
     }
+}
 
-    if input.just_pressed(KeyCode::KeyS) {
-        // Toggle debug outlines for nodes with `ViewVisibility` set to false.
-        debug_options.show_hidden = !debug_options.show_hidden;
-    }
+mod slice {
```
ickshonpe, line 448: This slice scene would be more useful if it used an asymetrical border image.

ickshonpe (approved):
This is good, much better than just trying to stuff everything into the single scene full UI example.

examples/testbed/ui.rs
mockersf: Do you know if there's already such a texture in the repo?

ickshonpe: There's `numbered_slices.png` which is used in the `ui_texture_slice_flip_and_tile` example.

PR #18093, opened 2025-02-28
Transform Propagation Optimization: Static Subtree Marking

aevyrie:
# Objective

- Optimize static scene performance by marking unchanged subtrees.

## Solution

- Mark hierarchy subtrees with dirty bits to avoid transform propagation where not needed
- This causes a performance regression when spawning many entities, or when the scene is entirely dynamic.
- This results in massive speedups for largely static scenes.
- In the future we could allow the user to change this behavior, or add some threshold based on how dynamic the scene is?

## Testing

- Caldera Hotel scene


crates/bevy_transform/src/systems.rs
```diff
@@ -1,6 +1,6 @@
-use crate::components::{GlobalTransform, Transform};
+use crate::components::{GlobalTransform, Transform, TransformTreeChanged};
 use bevy_ecs::prelude::*;
-
+use derive_more::From;
```
OleStrohm, line 3: It doesn't look like this is used anywhere in this file, can it be removed?

crates/bevy_transform/src/components/transform.rs
```diff
@@ -641,3 +645,20 @@ impl Mul<Vec3> for Transform {
...
+#[cfg_attr(
+    feature = "bevy_reflect",
+    derive(Reflect),
+    reflect(Component, Default, PartialEq, Debug)
+)]
+#[cfg_attr(
+    all(feature = "bevy_reflect", feature = "serialize"),
+    reflect(Serialize, Deserialize)
+)]
+pub struct TransformTreeChanged;
```
pcwalton, line 664: Bikeshed: maybe `TransformSubtreeChanged`? It's not the entire tree, it's just the subtree rooted at this node.

crates/bevy_transform/src/systems.rs
```diff
@@ -9,7 +8,7 @@ pub use serial::propagate_parent_transforms;
 /// Update [`GlobalTransform`] component of entities that aren't in the hierarchy
 ///
 /// Third party plugins should ensure that this is used in concert with
-/// [`propagate_parent_transforms`] and [`compute_transform_leaves`].
+/// [`propagate_parent_transforms`] and [`mark_dirty_trees`].
```
pcwalton, line 11: Likewise, maybe `mark_dirty_subtrees`?

pcwalton (approved):
Looks great, just had a bikeshed comment. Thanks for doing this work!

mockersf:
could you make the PR title clearer about what it is doing?

aevyrie:
I have comments and a description in the PR post, what are you looking for?

alice-i-cecile (approved):
Surprisingly simple! Really nice. We should test out doing the same with visibility at some point.

alice-i-cecile:
Thank you to everyone involved with the authoring or reviewing of this PR! This work is relatively important and needs release notes! Head over to https://github.com/bevyengine/bevy-website/issues/2003 if you'd like to help out.

PR #18099, opened 2025-03-01
Decoupled naga for wgpu testing

JMS55:
# Objective
- Allow bevy and wgpu developers to test newer versions of wgpu without having to update naga_oil.

## Solution

- Currently bevy feeds wgsl through naga_oil to get a naga::Module that it passes to wgpu.
- Added a way to pass wgsl through naga_oil, and then serialize the naga::Module back into a wgsl string to feed to wgpu, allowing wgpu to parse it using it's internal version of naga (and not the version of naga bevy_render/naga_oil is using).

## Testing
1. Run 3d_scene (it works)
2. Run 3d_scene with `--features bevy_render/decoupled_naga` (it still works)
3. Add the following patch to bevy/Cargo.toml, run cargo update, and compile again (it will fail)
```toml
[patch.crates-io]
wgpu = { git = "https://github.com/gfx-rs/wgpu", rev = "2764e7a39920e23928d300e8856a672f1952da63" }
wgpu-core = { git = "https://github.com/gfx-rs/wgpu", rev = "2764e7a39920e23928d300e8856a672f1952da63" }
wgpu-hal = { git = "https://github.com/gfx-rs/wgpu", rev = "2764e7a39920e23928d300e8856a672f1952da63" }
wgpu-types = { git = "https://github.com/gfx-rs/wgpu", rev = "2764e7a39920e23928d300e8856a672f1952da63" }
```
4. Fix errors and compile again (it will work, and you didn't have to touch naga_oil)

### DIFF
--- a/crates/bevy_render/Cargo.toml
+++ b/crates/bevy_render/Cargo.toml
@@ -9,20 +9,17 @@ license = "MIT OR Apache-2.0"
 keywords = ["bevy"]
 
 [features]
-default = ["coupled_naga"]
-
-# Bevy users should _always_ leave this feature on.
+# Bevy users should _never_ turn this feature on.
 #
-# Bevy/wgpu developers can turn this feature off (see `default` above) to test a newer version of wgpu
-# without needing to also update naga_oil. Make sure to update both wgpu and wgpu-types.
+# Bevy/wgpu developers can turn this feature on to test a newer version of wgpu without needing to also update naga_oil.
 #
-# When turning this feature off, you can add the following to bevy/Cargo.toml, and then run `cargo update`:
+# When turning this feature on, you can add the following to bevy/Cargo.toml (not this file), and then run `cargo update`:
 # [patch.crates-io]
 # wgpu = { git = "https://github.com/gfx-rs/wgpu", rev = "..." }
 # wgpu-core = { git = "https://github.com/gfx-rs/wgpu", rev = "..." }
 # wgpu-hal = { git = "https://github.com/gfx-rs/wgpu", rev = "..." }
 # wgpu-types = { git = "https://github.com/gfx-rs/wgpu", rev = "..." }
-coupled_naga = ["wgpu/naga-ir"]
+decoupled_naga = []
 
 # Texture formats (require more than just image support)
 basis-universal = ["bevy_image/basis-universal"]
@@ -92,6 +89,7 @@ wgpu = { version = "24", default-features = false, features = [
   "wgsl",
   "dx12",
   "metal",
+  "naga-ir",
   "fragile-send-sync-non-atomic-wasm",
 ] }
 naga = { version = "24", features = ["wgsl-in"] }
--- a/crates/bevy_render/Cargo.toml
+++ b/crates/bevy_render/Cargo.toml
@@ -13,8 +13,15 @@ default = ["coupled_naga"]
 
 # Bevy users should _always_ leave this feature on.
 #
-# Bevy/wgpu developers can turn this feature off to test a newer version of wgpu
+# Bevy/wgpu developers can turn this feature off (see `default` above) to test a newer version of wgpu
 # without needing to also update naga_oil. Make sure to update both wgpu and wgpu-types.
+#
+# When turning this feature off, you can add the following to bevy/Cargo.toml:
+# [patch.crates-io]
+# wgpu = { git = "https://github.com/gfx-rs/wgpu", rev = "..." }
+# wgpu-core = { git = "https://github.com/gfx-rs/wgpu", rev = "..." }
+# wgpu-hal = { git = "https://github.com/gfx-rs/wgpu", rev = "..." }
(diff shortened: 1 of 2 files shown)


crates/bevy_render/src/render_resource/pipeline_cache.rs
```diff
@@ -306,7 +306,28 @@ impl ShaderCache {
...
-                        ShaderSource::Naga(Cow::Owned(naga))
+                        #[cfg(feature = "coupled_naga")]
+                        {
+                            ShaderSource::Naga(Cow::Owned(naga))
+                        }
+
+                        #[cfg(not(feature = "coupled_naga"))]
+                        {
+                            let mut validator = naga::valid::Validator::new(
+                                naga::valid::ValidationFlags::empty(),
```
cwfitzgerald, line 317: The validation should likely be on, just to ensure the generated wgsl isn't bunk

crates/bevy_render/Cargo.toml
```diff
@@ -9,6 +9,21 @@ license = "MIT OR Apache-2.0"
...
 
 [features]
+default = ["coupled_naga"]
+
+# Bevy users should _always_ leave this feature on.
+#
+# Bevy/wgpu developers can turn this feature off (see `default` above) to test a newer version of wgpu
+# without needing to also update naga_oil. Make sure to update both wgpu and wgpu-types.
+#
+# When turning this feature off, you can add the following to bevy/Cargo.toml:
```
JMS55, line 19:
```suggestion
# When turning this feature off, you can add the following to bevy/Cargo.toml, and then `cargo update`:
```

crates/bevy_render/Cargo.toml
```diff
@@ -9,6 +9,21 @@ license = "MIT OR Apache-2.0"
...
 
 [features]
+default = ["coupled_naga"]
+
+# Bevy users should _always_ leave this feature on.
+#
+# Bevy/wgpu developers can turn this feature off (see `default` above) to test a newer version of wgpu
+# without needing to also update naga_oil. Make sure to update both wgpu and wgpu-types.
+#
+# When turning this feature off, you can add the following to bevy/Cargo.toml, and then `cargo update`:
```
JMS55, line 19:
```suggestion
# When turning this feature off, you can add the following to bevy/Cargo.toml, and then run `cargo update`:
```

cwfitzgerald:
Would it make sense to invert the flag - so by default is coupled, and when you enable the decouple flag you can enable wgsl-out, and do the decoupling. This would allow it to be enabled by any crate in the stack too.

JMS55:
I couldn't figure out how to do it. The naga-ir feature in wgpu is not needed when in decoupled mode, but features can't _remove_ other features. I suppose I could leave it on though, and just have the feature continue to gate the actual code returning ShaderSource, but that still leaves the question of how to enable it... I don't really want it to be a bevy feature, so either way devs will have to go into bevy_render's Cargo.toml and add it to the default feature list, or remove it.

cwfitzgerald:
> The naga-ir feature in wgpu is not needed when in decoupled mode

The naga-ir feature in wgpu just turns on naga support, it doesn't actually add anything in naga, so that's fine.

> but that still leaves the question of how to enable it... 

You can enable it through feature unification. As a user, you add `bevy_render = { version = "blah", feature = ["decoupled-nagagpu"] }` or whatever, and then bevy_render for all users will have that feature enabled. This is how `profiling` works, for example.

alice-i-cecile (approved):
I'm fine with this as an intermediate step before moving to WESL and removing naga_oil. If we can get a wgpu maintainer to sign off on this (as a sign that it would be useful to them!) I'm happy to merge.

cwfitzgerald:
I just would like to see the feature flag inverted, then this should be good to go.

JMS55:
I'm not really convinced by inverting the feature flag. Either way developers will be going into cargo.toml to edit it, as I don't really want to expose this feature to the top level bevy crate.

fjkorf (approved):
This looks good. We're adding the ability to decouple naga. I don't have a strong opinion about calling it `coupled_naga` and defaulting to true, or calling it `decoupled_naga` and defaulting to false. I think the current implementation solves the problem.

IceSentry:
I think I agree with @cwfitzgerald that inverting the feature flag would make more sense? So by default it's coupled and to use decoupled-naga you just need to add a feature. Disabling a single feature is not supported by cargo so it seems way easier to me to have it so we need to add a feature instead of disabling one.

alice-i-cecile (approved):

cwfitzgerald (approved):

JMS55:
@alice-i-cecile can we override dependency checks?

mockersf:
dependency check is not required so it's not blocking to merge a pr

JMS55:
Oh it failed wasm build and macOS screenshots. Hmm.

IceSentry (approved):

PR #18105, opened 2025-03-01
Fix invisible window creation on Windows

### LINKED ISSUE #18027: Application started with a hidden window never actually runs
## Bevy version

v0.15.3

Windows 11

## What you did

Setting the primary window to `visible = false` when setting up the `WindowPlugin`.

## What went wrong

The application never actually starts running.

## Additional information

The  [`window_setting`](https://github.com/bevyengine/bevy/blob/main/examples/window/window_settings.rs) example does this to delay window visibility to "avoid the white window that shows up before the GPU is ready to render the app".

This bug was not present before v0.15.3 so I assume it comes from https://github.com/bevyengine/bevy/pull/18004.

aloucks:
# Objective

Fixes #18027 

## Solution

Run `redraw_requested` logic in `about_to_wait` on Windows during initial application startup and when in headless mode

## Testing

- Ran `cargo run --example window_settings` to demonstrate invisible window creation worked again and fixes #18027
- Ran `cargo run --example eased_motion` to demonstrate no regression with the fix for #17488 

Ran all additional `window` examples. 

Notes:

- The `transparent_window` was not transparent but this appears to have been broken prior to #18004.  See: #7544

### DIFF
--- a/crates/bevy_winit/src/state.rs
+++ b/crates/bevy_winit/src/state.rs
@@ -435,8 +435,12 @@ impl<T: Event> ApplicationHandler<T> for WinitAppRunnerState<T> {
                 // https://github.com/bevyengine/bevy/issues/17488
                 #[cfg(target_os = "windows")]
                 {
-                    self.redraw_requested = true;
-                    self.redraw_requested(_event_loop);
+                    // Have the startup behavior run in about_to_wait, which prevents issues with
+                    // invisible window creation. https://github.com/bevyengine/bevy/issues/18027
+                    if self.startup_forced_updates == 0 {
+                        self.redraw_requested = true;
+                        self.redraw_requested(_event_loop);
+                    }
                 }
             }
             _ => {}
@@ -480,6 +484,15 @@ impl<T: Event> ApplicationHandler<T> for WinitAppRunnerState<T> {
         //       The monitor sync logic likely belongs in monitor event handlers and not here.
         #[cfg(not(target_os = "windows"))]
         self.redraw_requested(event_loop);
+
+        // Have the startup behavior run in about_to_wait, which prevents issues with
+        // invisible window creation. https://github.com/bevyengine/bevy/issues/18027
+        #[cfg(target_os = "windows")]
+        {
+            if self.startup_forced_updates > 0 {
+                self.redraw_requested(event_loop);
+            }
+        }
     }
 
     fn suspended(&mut self, _event_loop: &ActiveEventLoop) {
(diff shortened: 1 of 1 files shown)


mockersf (approved):

greeble-dev:
I tested the `window_settings` example and confirmed it's broken on main and fixed with the PR.

However, I was a bit uncertain if the window was really hidden long enough, so I tried increasing the threshold in `window_settings.rs/make_visible` from 3 to 10 just to see. This broke the example again :[ - the window never appears.

After [adding some debug prints](https://github.com/greeble-dev/bevy/commit/628e1135ac6e28549da6739a8f3eebd2c2c349c6), my guess is that it breaks if the app waits for more frames than the default `startup_forced_updates` of 5.

Testing the PR unchanged:

```
INFO bevy_winit::system: Creating new window I am a window! (0v1)
about_to_wait self.startup_forced_updates = 5, headless = false
make_visible frames = 0
about_to_wait self.startup_forced_updates = 4, headless = false
make_visible frames = 1
about_to_wait self.startup_forced_updates = 3, headless = false
make_visible frames = 2
about_to_wait self.startup_forced_updates = 2, headless = false
make_visible frames = 3
make_visible setting window.visible = true
WindowEvent::RedrawRequested self.startup_forced_updates = 1
about_to_wait self.startup_forced_updates = 1, headless = false
make_visible frames = 4
WindowEvent::RedrawRequested self.startup_forced_updates = 0
make_visible frames = 5
```

Aftering increasing `make_visible` threshold to 10:

```
INFO bevy_winit::system: Creating new window I am a window! (0v1)
about_to_wait self.startup_forced_updates = 5, headless = false
make_visible frames = 0
about_to_wait self.startup_forced_updates = 4, headless = false
make_visible frames = 1
about_to_wait self.startup_forced_updates = 3, headless = false
make_visible frames = 2
about_to_wait self.startup_forced_updates = 2, headless = false
make_visible frames = 3
about_to_wait self.startup_forced_updates = 1, headless = false
make_visible frames = 4
about_to_wait self.startup_forced_updates = 0, headless = false
about_to_wait self.startup_forced_updates = 0, headless = false
about_to_wait self.startup_forced_updates = 0, headless = false
...repeats forever...
```

If I then increase `startup_forced_updates` so it's well above the threshold then the example works again.

`SystemInfo { os: "Windows 10 Pro", kernel: "19045" }`

aloucks:
I've added another check to explicitly look for when all windows are invisible. There is some other logic that is driving the event loop on the visibility status and it has a bunch `cfg` guards around it. The winnit docs suggest not to drive logic from that status (which is noted in comment in the cfg guards). I didn't want to to attempt to make changes in that area in fear that it might cause some other issue.

greeble-dev (approved):
Latest version fixes the issues I had. Tested with various delays in `window_settings`.

PR #18112, opened 2025-03-02
Add `on_insert` hook to `Node` to warn if parent has `Text`

### LINKED ISSUE #17551: Adding children to UI Text makes it disappear
## Bevy version

main (1c765c9ae761)

## What you did

When you add a `Node` as a child of a `Text`, the text is not rendered at all.

```rust
use bevy::prelude::*;

fn main() {
    App::new()
        .add_plugins(DefaultPlugins)
        .add_systems(Startup, setup)
        .run();
}

    commands.spawn(Camera2d);
    commands
        .spawn((
            Text::new("hello\nbevy!"),
            TextFont {
                font_size: 67.0,
                ..default()
            },
            Node {
                left: Val::Px(20.),
                ..default()
            },
        ))
        // Adding this makes the above text disappear
        .with_child((
            Node {
                width: Val::Px(40.),
                height: Val::Px(40.),
                ..default()
            },
            BackgroundColor(Color::WHITE.with_alpha(0.5)),
        ));
}
```

## What went wrong

I would expect the text to remain visible.

The entity with `Text` still seems to contribute to the layout, but it doesn't show up. It seems that `Text` is somehow hardcoded to only allow `TextSpan` children and breaks with anything else.

I ran into this while trying to make pseudo outlines by drawing multiple texts on top of each other in the hierarchy, but the parent text just doesn't show up.

### LINKED ISSUE #18015: UI Text nodes overlap when they have children
## Bevy version

0.15.1

`SystemInfo { os: "Windows 10 Pro", kernel: "19045", cpu: "AMD Ryzen 7 5800X 8-Core Processor", core_count: "4", memory: "15.9 GiB" }`

`cargo 1.86.0-nightly (0e3d73849 2025-02-01)`

## What you did

Create what's essentially an HBox node and add two Text nodes as children to it. Everything is fine, both nodes are visible and have their position calculated correctly.

```rs
fn spawn_text_input_child(parent: &mut ChildBuilder, name: &'static str, text: &str) {
    parent
        .spawn((Name::new(name), Text::new(text)));
        // .with_child((
        //     Name::new("cursor"),
        //     Node {
        //         min_height: Val::Px(20.0),
        //         min_width: Val::Px(1.0),
        //         ..default()
        //     },
        //     BackgroundColor(bevy::color::Color::Srgba(WHITE)),
        // ));
}
```

Image

Now add a child `Node::default()` to each of Text node. Both Text nodes suddenly disappear and their ComputedNode components have `.size == Vec2::ZERO`.

```rs
fn spawn_text_input_child(parent: &mut ChildBuilder, name: &'static str, text: &str) {
    parent
        .spawn((Name::new(name), Text::new(text)))
        .with_child((
            Name::new("cursor"),
            Node {
                // min_height: Val::Px(20.0),
                // min_width: Val::Px(1.0),
                ..default()
            },
            // BackgroundColor(bevy::color::Color::Srgba(WHITE)),
        ));
}
```

Image

Now add a BackgroundColor component to these newly-added default nodes and give them some size so that they're visible. Text nodes are now visible again alongside their children, but are heavily distorted and squeezed into one space.

```rs
fn spawn_text_input_child(parent: &mut ChildBuilder, name: &'static str, text: &str) {
    parent
        .spawn((Name::new(name), Text::new(text)))
        .with_child((
            Name::new("cursor"),
            Node {
                min_height: Val::Px(20.0),
                min_width: Val::Px(1.0),
                ..default()
            },
            BackgroundColor(bevy::color::Color::Srgba(WHITE)),
        ));
}
```

Image

## What went wrong

I expected children of Text nodes to simply be drawn over them. Instead, even simply adding an empty Node to a Text caused it to completely break apart

## Additional information

Other information that can be used to further reproduce or isolate the problem.
This commonly includes:

- screenshots: provided above
- logs: irrelevant
- theories about what might be going wrong: none
- workarounds that you used: couldn't find any
- links to related bugs, PRs or discussions: couldn't find anything similar

MRP: [zerosizebug.zip](https://github.com/user-attachments/files/18948828/zerosizebug.zip) (ignore the name, i've encountered this bug while trying to pinpoint another one)

bytemunch:
# Objective

Add a warning when adding a `Node` as a child of an entity with a `Text` component, and some information in `Text` docs.

Closes #18015
Would close #17551 if it weren't already

## Solution

Use an `on_insert` component hook on `Node` to check the node's parent and emit a warning.

## Testing

<details>
  <summary>Click to view test code</summary>

```rust
use bevy::{input::common_conditions::input_just_pressed, prelude::*};

fn main() {
    App::new()
        .add_plugins(DefaultPlugins)
        .add_systems(Startup, (setup, setup_insert))
        .add_systems(
            Update,
            insert_test.run_if(input_just_pressed(MouseButton::Left)),
        )
        .run();
}

    commands.spawn(Camera2d);
    commands
        .spawn(Text::new("Hello, bevy!"))
        .with_children(|cb| {
            cb.spawn(Node::default()); // Warning triggers here.
        });
}

fn setup_insert(mut commands: Commands) {
    commands
        .spawn(Text::new("Hello, bevy!"))
        .with_children(|cb| {
            cb.spawn_empty();
        });
}

fn insert_test(mut commands: Commands, q: Query<Entity, With<ChildOf>>) {
    for e in &q {
        commands.entity(e).insert(Node::default()); // Warning triggers here.
    }
}

```

</details>

---

### DIFF
--- a/crates/bevy_ui/Cargo.toml
+++ b/crates/bevy_ui/Cargo.toml
@@ -20,6 +20,7 @@ bevy_ecs = { path = "../bevy_ecs", version = "0.16.0-dev" }
 bevy_image = { path = "../bevy_image", version = "0.16.0-dev" }
 bevy_input = { path = "../bevy_input", version = "0.16.0-dev" }
 bevy_math = { path = "../bevy_math", version = "0.16.0-dev" }
+bevy_log = { path = "../bevy_log", version = "0.16.0-dev" }
 bevy_reflect = { path = "../bevy_reflect", version = "0.16.0-dev", features = [
   "bevy",
 ] }
--- a/crates/bevy_ui/src/ui_node.rs
+++ b/crates/bevy_ui/src/ui_node.rs
@@ -1,7 +1,8 @@
-use crate::{FocusPolicy, UiRect, Val};
+use crate::{widget::Text, FocusPolicy, UiRect, Val};
 use bevy_color::Color;
 use bevy_derive::{Deref, DerefMut};
-use bevy_ecs::{prelude::*, system::SystemParam};
+use bevy_ecs::{component::HookContext, prelude::*, system::SystemParam, world::DeferredWorld};
+use bevy_log::error_once;
 use bevy_math::{vec4, Rect, UVec2, Vec2, Vec4Swizzles};
 use bevy_reflect::prelude::*;
 use bevy_render::{
@@ -340,6 +341,7 @@ impl From<Vec2> for ScrollPosition {
     derive(serde::Serialize, serde::Deserialize),
     reflect(Serialize, Deserialize)
 )]
+#[component(on_insert=on_insert_node)]
 pub struct Node {
     /// Which layout algorithm to use when laying out this node's contents:
     ///   - [`Display::Flex`]: Use the Flexbox layout algorithm
@@ -620,6 +622,16 @@ pub struct Node {
     ///
     /// <https://developer.mozilla.org/en-US/docs/Web/CSS/grid-column>
     pub grid_column: GridPlacement,
+}
+
+fn on_insert_node(world: DeferredWorld, ctx: HookContext) {
+    let Some(childof) = world.get::<ChildOf>(ctx.entity) else {
+        return;
+    };
+
+    if let Some(_parent) = world.get::<Text>(childof.parent) {
+        error_once!("Found `Node` with `Text` as a component of parent entity. This is unsupported behaviour and will lead to unexpected results.");
+    }
 }
 
 impl Node {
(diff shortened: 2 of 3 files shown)


alice-i-cecile:
I'm a little nervous about this pattern (warning due to non-local invariant violation), but I don't have a strong argument for either why it's dangerous or what we should do instead.

bytemunch:
I've also noticed that inserting a `ChildOf` doesn't trigger the warning, so this code wouldn't be covered.

```rs
    let text = commands.spawn(Text::new("Hello, bevy!")).id();

    commands
        .spawn(Node::default())
        .insert(ChildOf { parent: text });
```

Which would point to running the check on the `ChildOf` component, but that feels like overreach for such a specific case. Maybe a system looking for `Query<&Node, Added<ChildOf>>`, but that could have performance impacts on dense UI scenes.

The pattern was inspired by `bevy_ecs::hierarchy::validate_parent_has_component`, which outputs a similar warning when the parent is expected to have a certain component.

Would a better approach perhaps be to define a `LeafNode` marker component that refuses to have children, and warn/forbid when using that specifically?

Issue #17869, opened 2025-02-15
`WGPU_SETTINGS_PRIO=webgl2` causes panics


greeble-dev:
## TLDR

Some rendering plugins may panic after trying to use wgpu features that have been disabled by the `WGPU_SETTINGS_PRIO` env var. The solution might be to add more checks.

## Bevy version

0.16-dev 15b795d7d6a97fd4006dd2c1233050662327dcce

## System info

```
SystemInfo { os: "Windows 10 Pro", kernel: "19045" }
AdapterInfo { name: "NVIDIA GeForce GT 1030", vendor: 4318, device: 7425, device_type: DiscreteGpu, driver: "NVIDIA", driver_info: "560.94", backend: Vulkan }
```

## What you did

I wanted to test what happens if the renderer's skinning pipeline uses uniform buffers. I was too lazy to make a real wasm build, so after a discord suggestion I tried the `WGPU_SETTINGS_PRIO` env var. This is a Bevy feature - see `settings_priority_from_env()`.

```sh
WGPU_SETTINGS_PRIO=webgl2
# Or for windows powershell: $env:WGPU_SETTINGS_PRIO = "webgl2"

cargo run --example animated_mesh
```

(Side note: shouldn't a Bevy env var start with `BEVY_`?)

## What went wrong

```
thread 'main' panicked at wgpu-24.0.1\src\backend\wgpu_core.rs:1079:26:
wgpu error: Validation Error

Caused by:
  In Device::create_bind_group_layout, label = 'oit_resolve_bind_group_layout'
    Too many bindings of type StorageBuffers in Stage ShaderStages(FRAGMENT), limit is 0, count was 2. Check the limit `max_storage_buffers_per_shader_stage` passed to `Adapter::request_device`

stack backtrace:
    ...
   8: bevy_render::renderer::render_device::RenderDevice::create_bind_group_layout<ref$<str$> >
             at .\crates\bevy_render\src\renderer\render_device.rs:146
   9: bevy_core_pipeline::oit::resolve::impl$1::from_world
             at .\crates\bevy_core_pipeline\src\oit\resolve\mod.rs:93    
```

## So what's going on?

`WGPU_SETTINGS_PRIO=webgl2` causes `WgpuSettings::limits` to be initialized to `wgpu::Limits::downlevel_webgl2_defaults()`, which `initialize_renderer()` then passes into `wgpu::DeviceDescriptor`.

[...]

PR #18113, opened 2025-03-02
Partially fix panics when setting `WGPU_SETTINGS_PRIO=webgl2`

greeble-dev:
# Overview

Fixes https://github.com/bevyengine/bevy/issues/17869.

# Summary

`WGPU_SETTINGS_PRIO` changes various limits on `RenderDevice`. This is useful to simulate platforms with lower limits.

However, some plugins only check the limits on `RenderAdapter` (the actual GPU) - these limits are not affected by `WGPU_SETTINGS_PRIO`. So the plugins try to use features that are unavailable and wgpu says "oh no". See https://github.com/bevyengine/bevy/issues/17869 for details.

The PR adds various checks on `RenderDevice` limits. This is enough to get most examples working, but some are not fixed (see below).

# Testing

- Tested native, with and without "WGPU_SETTINGS=webgl2". Win10/Vulkan/Nvidia".
- Also WebGL. Win10/Chrome/Nvidia.

```
$env:WGPU_SETTINGS_PRIO = "webgl2"
cargo run --example testbed_3d
cargo run --example testbed_2d
cargo run --example testbed_ui
cargo run --example deferred_rendering
cargo run --example many_lights
cargo run --example order_independent_transparency # Still broken, see below.
cargo run --example occlusion_culling # Still broken, see below.
```

# Not Fixed

While testing I found a few other cases of limits being broken. 

"Compatibility" settings (WebGPU minimums) breaks native in various examples.

```
$env:WGPU_SETTINGS_PRIO = "compatibility"
cargo run --example testbed_3d

  In Device::create_bind_group_layout, label = 'build mesh uniforms GPU early occlusion culling bind group layout'
    Too many bindings of type StorageBuffers in Stage ShaderStages(COMPUTE), limit is 8, count was 9. Check the limit `max_storage_buffers_per_shader_stage` passed to `Adapter::request_device`
```

`occlusion_culling` breaks fake webgl.
```
$env:WGPU_SETTINGS_PRIO = "webgl2"
cargo run --example occlusion_culling

thread '<unnamed>' panicked at C:\Projects\bevy\crates\bevy_render\src\render_resource\pipeline_cache.rs:555:28:
index out of bounds: the len is 0 but the index is 2
Encountered a panic in system `bevy_render::renderer::render_system`!
```

`occlusion_culling` breaks real webgl.

```
cargo run --example occlusion_culling --target wasm32-unknown-unknown

ERROR app: panicked at C:\Users\T\.cargo\registry\src\index.crates.io-1949cf8c6b5b557f\glow-0.16.0\src\web_sys.rs:4223:9:
Tex storage 2D multisample is not supported
```

OIT breaks fake webgl.

```
$env:WGPU_SETTINGS_PRIO = "webgl2"
cargo run --example order_independent_transparency

  In Device::create_bind_group, label = 'mesh_view_bind_group'
    Number of bindings in bind group descriptor (28) does not match the number of bindings defined in the bind group layout (25)

```

OIT breaks real webgl

```
cargo run --example order_independent_transparency --target wasm32-unknown-unknown

  In Device::create_render_pipeline, label = 'pbr_oit_mesh_pipeline'
    Error matching ShaderStages(FRAGMENT) shader requirements against the pipeline
      Shader global ResourceBinding { group: 0, binding: 34 } is not available in the pipeline layout
        Binding is missing from the pipeline layout
```

### DIFF
--- a/crates/bevy_core_pipeline/src/experimental/mip_generation/mod.rs
+++ b/crates/bevy_core_pipeline/src/experimental/mip_generation/mod.rs
@@ -350,6 +350,7 @@ fn create_downsample_depth_pipelines(
         .get_downlevel_capabilities()
         .flags
         .contains(DownlevelFlags::COMPUTE_SHADERS)
+        || (render_device.limits().max_compute_workgroup_storage_size == 0)
     {
         return;
     }
--- a/crates/bevy_core_pipeline/src/experimental/mip_generation/mod.rs
+++ b/crates/bevy_core_pipeline/src/experimental/mip_generation/mod.rs
@@ -325,6 +325,18 @@ pub struct DownsampleDepthPipelines {
     sampler: Sampler,
 }
 
+fn supports_compute_shaders(device: &RenderDevice, adapter: &RenderAdapter) -> bool {
+    adapter
+        .get_downlevel_capabilities()
+        .flags
+        .contains(DownlevelFlags::COMPUTE_SHADERS)
+    // Even if the adapter supports compute, we might be simulating a lack of
+    // compute via device limits (see `WgpuSettingsPriority::WebGL2` and
+    // `wgpu::Limits::downlevel_webgl2_defaults()`). This will have set all the
+    // `max_compute_*` limits to zero, so we arbitrarily pick one as a canary.
+    && (device.limits().max_compute_workgroup_storage_size != 0)
+}
+
 /// Creates the [`DownsampleDepthPipelines`] if downsampling is supported on the
 /// current platform.
 fn create_downsample_depth_pipelines(
@@ -346,12 +358,7 @@ fn create_downsample_depth_pipelines(
 
     // If we don't have compute shaders, we can't invoke the downsample depth
     // compute shader.
-    if !render_adapter
-        .get_downlevel_capabilities()
-        .flags
-        .contains(DownlevelFlags::COMPUTE_SHADERS)
-        || (render_device.limits().max_compute_workgroup_storage_size == 0)
-    {
+    if !supports_compute_shaders(&render_device, &render_adapter) {
         return;
     }
 
--- a/crates/bevy_core_pipeline/src/oit/resolve/mod.rs
+++ b/crates/bevy_core_pipeline/src/oit/resolve/mod.rs
@@ -53,29 +53,11 @@ impl Plugin for OitResolvePlugin {
             return;
         };
 
-        if !render_app
-            .world()
-            .resource::<RenderAdapter>()
-            .get_downlevel_capabilities()
-            .flags
-            .contains(DownlevelFlags::FRAGMENT_WRITABLE_STORAGE)
-        {
-            warn!("OrderIndependentTransparencyPlugin not loaded. GPU lacks support: DownlevelFlags::FRAGMENT_WRITABLE_STORAGE.");
-            return;
-        }
-
-        let max_storage_buffers_per_shader_stage = render_app
-            .world()
-            .resource::<RenderDevice>()
-            .limits()
-            .max_storage_buffers_per_shader_stage;
-
-        if max_storage_buffers_per_shader_stage < OIT_REQUIRED_STORAGE_BUFFERS {
-            warn!(
-                max_storage_buffers_per_shader_stage,
-                OIT_REQUIRED_STORAGE_BUFFERS,
-                "OrderIndependentTransparencyPlugin not loaded. RenderDevice lacks support: max_storage_buffers_per_shader_stage < OIT_REQUIRED_STORAGE_BUFFERS."
-            );
+        if !is_oit_supported(
+            render_app.world().resource::<RenderAdapter>(),
+            render_app.world().resource::<RenderDevice>(),
+            true,
+        ) {
             return;
         }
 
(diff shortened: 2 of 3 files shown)


crates/bevy_core_pipeline/src/experimental/mip_generation/mod.rs
```diff
@@ -350,6 +350,7 @@ fn create_downsample_depth_pipelines(
         .get_downlevel_capabilities()
         .flags
         .contains(DownlevelFlags::COMPUTE_SHADERS)
+        || (render_device.limits().max_compute_workgroup_storage_size == 0)
```
greeble-dev, line 353: This check is a little awkward - there's no straightforward "is compute available" on `wgpu::Limits`.

JMS55:
If we can get this setup in CI, I would love automated tests for this. It's not feasible to manually test this kind of stuff myself, but it would be really nice to have CI let me know when we're going above WebGL2 or WebGPU minspec.

crates/bevy_core_pipeline/src/experimental/mip_generation/mod.rs
JMS55: This is bizarre. The default for WebGPU is fairly high https://gpuweb.github.io/gpuweb/#dom-supported-limits-maxcomputeworkgroupstoragesize. I didn't even know some platforms support compute shaders, but not workgroup-shared data.

@cwfitzgerald, can we get some input from wgpu/gpuweb people? Is this valid?

JMS55 (approved):

crates/bevy_core_pipeline/src/experimental/mip_generation/mod.rs
greeble-dev: Clarified in https://github.com/bevyengine/bevy/pull/18113/commits/d89d39ce692497d1eef7741afb814a9199e034ab. Still kinda icky though.

I also considered working out some real limits for the mip generation, e.g. a minimum for `max_compute_workgroup_size_x/y/z`. But I'm not confident I can get that right.

cwfitzgerald: > I didn't even know some platforms support compute shaders, but not workgroup-shared data.

They don't.

> downlevel_webgl2_defaults()

You should only need to check the `COMPUTE_SHADERS` downlevel flag to know if compute shaders are supported.

greeble-dev: As far as I can tell, checking `COMPUTE_SHADERS` on the adapter is not sufficient when the device limits have been overridden via `DeviceDescriptor::limits` (with the intent of simulating webgl2 limits on another platform).

If I rely only on `COMPUTE_SHADERS`:

```
$env:WGPU_SETTINGS_PRIO = "webgl2"
cargo run --example occlusion_culling

  In Device::create_bind_group_layout, label = 'downsample depth bind group layout'
    Too many bindings of type StorageTextures in Stage ShaderStages(COMPUTE), limit is 0, count was 12. Check the limit `max_storage_textures_per_shader_stage` passed to `Adapter::request_device`
```

I guess there's a deeper question on whether setting device limits is the right way to simulate another platform's limits?

crates/bevy_core_pipeline/src/oit/resolve/mod.rs
```diff
@@ -61,6 +64,17 @@ impl Plugin for OitResolvePlugin {
...
         }
 
+        if render_app
+            .world()
+            .resource::<RenderDevice>()
+            .limits()
+            .max_storage_buffers_per_shader_stage
+            < OIT_REQUIRED_STORAGE_BUFFERS
+        {
+            warn!("OrderIndependentTransparencyPlugin not loaded. RenderDevice lacks support: max_storage_buffers_per_shader_stage < OIT_REQUIRED_STORAGE_BUFFERS.");
```
IceSentry, line 74:
```suggestion
            warn!("OrderIndependentTransparencyPlugin not loaded. RenderDevice lacks support: max_storage_buffers_per_shader_stage < {OIT_REQUIRED_STORAGE_BUFFERS}.");
```

I think it would be better to just write the value. Or maybe add it after in parentheses?

crates/bevy_pbr/src/render/mesh_view_bindings.rs
```diff
@@ -388,6 +390,8 @@ fn layout_entries(
             .get_downlevel_capabilities()
             .flags
             .contains(DownlevelFlags::FRAGMENT_WRITABLE_STORAGE)
+            && (render_device.limits().max_storage_buffers_per_shader_stage
```
IceSentry, line 393: Not super important for now, but it might be good to abstract this to a `is_oit_supported(adapter, device)` function.

crates/bevy_core_pipeline/src/oit/resolve/mod.rs
```diff
@@ -61,6 +64,17 @@ impl Plugin for OitResolvePlugin {
...
+            return;
```
IceSentry, line 75: It might be nice to return after all compatibility checks are printed, but this is good enough as is

IceSentry (approved):
Thanks! I added the OIT checks to my internal implementation at work and I should be more careful about this in the future 😅.

crates/bevy_core_pipeline/src/experimental/mip_generation/mod.rs
JMS55: > I guess there's a deeper question on whether setting device limits is the right way to simulate another platform's limits?

It is, yeah. Not quite sure I understand which part is the issue, but sounds like we're overriding the downlevel flags or limits incorrectly.

cwfitzgerald: The problem is that currently you cannot override downlevel features - you can only override the limits.

crates/bevy_core_pipeline/src/oit/resolve/mod.rs
greeble-dev: Trying a fancier version in c19f0f2855c94c6a9b1114f94e6f47d503d22c68. Maybe a little too wordy now.

```
OrderIndependentTransparencyPlugin not loaded. RenderDevice lacks support: max_storage_buffers_per_shader_stage < OIT_REQUIRED_STORAGE_BUFFERS. max_storage_buffers_per_shader_stage=0 OIT_REQUIRED_STORAGE_BUFFERS=2
```

crates/bevy_pbr/src/render/mesh_view_bindings.rs
greeble-dev: Tried this in 12aa7b1951fa597045abfe0777bbd6c20586c04a.

crates/bevy_core_pipeline/src/oit/resolve/mod.rs
```diff
@@ -73,6 +73,34 @@ impl Plugin for OitResolvePlugin {
     }
 }
 
+pub fn is_oit_supported(adapter: &RenderAdapter, device: &RenderDevice, warn: bool) -> bool {
```
greeble-dev, line 76: The `warn: bool` is a bit smelly, but wasn't sure if an enum was justified for two calls.

Also considered putting the warning behind a `OnceLock`, but that's technically kinda wrong as the warning is specifically about the plugin not loading.

greeble-dev:
Updated to address a few concerns - no functional changes, just refactoring and tweaking warnings.

PR #18118, opened 2025-03-02
Fix label name for Deliberate-Rendering-Change CI

alice-i-cecile:
# Objective

The label added in #18064 is the wrong color! It should be an `M-` (meta) label, not an `S-` (status) label!

## Solution

- Rename the label in the comment that gets left to `M-Deliberate-Rendering-Change`
- Also make the job check for the right label 🤦🏽‍♀️ 

Once this is approved and ready to merge, I'll rename the label on Github to match.

### DIFF
--- a/.github/workflows/example-run-report.yml
+++ b/.github/workflows/example-run-report.yml
@@ -90,7 +90,7 @@ jobs:
           PR: ${{ needs.make-macos-screenshots-available.outputs.pr-number }}
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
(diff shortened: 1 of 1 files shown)


mockersf (approved):
Not sure I understand what you mean by wrong colour? And I think it's more a state than meta?

But then you're the label boss!

alice-i-cecile:
> Not sure I understand what you mean by wrong colour?

This was a joke about the category being wrong: we generally only ever want one purple label per PR.

> And I think it's more a state than meta?

State should be for a single PR status, Meta is "metadata for our assorted workflows" :)

alice-i-cecile:
image

Issue #17590, opened 2025-01-28
Multiple Fog Volumes Panic


the-real-ale:
## Bevy version

release-0.15.1, tag: v0.15.1

## \[Optional\] Relevant system information

cargo 1.86.0-nightly (cecde95c1 2025-01-24)

```
SystemInfo { os: "Windows 10 Home", kernel: "19045", cpu: "11th Gen Intel(R) Core(TM) i7-11700K @ 3.60GHz", core_count: "8", memory: "63.9 GiB" }

AdapterInfo { name: "NVIDIA GeForce RTX 3060 Ti", vendor: 4318, device: 9353, device_type: DiscreteGpu, driver: "NVIDIA", driver_info: "566.36", backend: Vulkan }
```

## What you did

In `volumetric_fog` example, duplicate the FogVolume spawn.
Possibly unrelated, but the only way I could get volumetric fog to run on Windows 10 was to change the gpu backend to VULKAN.

```
    App::new()
        .add_plugins(DefaultPlugins.set(RenderPlugin {
            render_creation: RenderCreation::Automatic(WgpuSettings {
                backends: Some(Backends::VULKAN),
                ..default()
            }),
            ..default()
        }))
	...
```
Other than these two changes, the example code is identical to release 0.15.1.

## What went wrong

With the two fog volumes spawning, attempt to run and encounter panic:
```
thread 'Compute Task Pool (6)' panicked at bevy\crates\bevy_render\src\render_resource\uniform_buffer.rs:368:34:
called `Result::unwrap()` on an `Err` value: BufferTooSmall { expected: 288, found: 0 }
```

PR #18119, opened 2025-03-02
Fix panic with multiple fog volumes

greeble-dev:
# Objective

Fixes https://github.com/bevyengine/bevy/issues/17590.

## Solution

`prepare_volumetric_fog_uniforms` adds a uniform for each combination of fog volume and view. But it only allocated enough uniforms for one fog volume per view.

## Testing

Ran the `volumetric_fog` example with 1/2/3/4 fog volumes. Also checked the `fog_volumes` and `scrolling_fog` examples (without multiple volumes). Win10/Vulkan/Nvidia. 

To test multiple views I tried adding fog volumes to the `split_screen` example. This doesn't quite work - the fog should be centred on the fox, but instead it's centred on the window. Same result with and without the PR, so I'm assuming it's a separate bug.

image


pcwalton (approved):
Looks good!

JMS55 (approved):

mockersf (approved):

PR #1263, opened 2021-01-19
[Merged by Bors] - Query::get_unique

TheRawMeatball:
Adds `get_unique` and `get_unique_mut` to extend the query api and cover a common use case. Also establishes a second impl block where non-core APIs that don't access the internal fields of queries can live.

PR #18120, opened 2025-03-02
Deprecate `Query::get_many` (etc) and make `many` return a Result

alice-i-cecile:
# Objective

This PR is the less contentious sibling to #18082. See that PR for detailed motivation!

This PR affects APIs that are called much less often: as a result, handling said `Result` is much less onerous.

## Solution

1. Make the short `Query::many` methods return a `Result`.
2. Deprecate the `Query::get_many` methods now that they've been renamed.
3. Fix up all the warnings.

## Testing

No more compiler errors left! We'll see what CI finds.

## Migration Guide

As part of a broad shift to reduce panics in Bevy, `Query::many`, `Query::many_mut` and the same methods on `QueryState` now return a `Result`, which should be handled by the caller. The `get_many` and `get_many_mut` methods that previously did this operation while returning a `Result` have been deprecated, as they have effectively been renamed.


alice-i-cecile:
@Carter0 can I have your review here too? Double check that I didn't screw anything up: this was an annoying change set due to all the duplicated APIs.

crates/bevy_ecs/src/query/state.rs
```diff
@@ -952,12 +965,12 @@ impl<D: QueryData, F: QueryFilter> QueryState<D, F> {
     /// assert_eq!(match query_state.get_many(&mut world, [wrong_entity]).unwrap_err() {QueryEntityError::EntityDoesNotExist(error) => error.entity, _ => panic!()}, wrong_entity);
     /// ```
     #[inline]
-    pub fn get_many<'w, const N: usize>(
```
Carter0, line 955: Should you change the docs above this function too. So like 

```
let component_values = query_state.get_many(&world, entities).unwrap(); -> let component_values = query_state.many(&world, entities).unwrap();
```

crates/bevy_ecs/src/query/state.rs
```diff
@@ -1011,12 +1037,12 @@ impl<D: QueryData, F: QueryFilter> QueryState<D, F> {
     /// assert_eq!(query_state.get_many_mut(&mut world, [entities[0], entities[0]]).unwrap_err(), QueryEntityError::AliasedMutability(entities[0]));
     /// ```
     #[inline]
-    pub fn get_many_mut<'w, const N: usize>(
+    pub fn many_mut<'w, const N: usize>(
```
Carter0, line 1040: Same thing as above with the doc comments

```
query_state.get_many_mut -> query_state.many_mut 
```

crates/bevy_ecs/src/system/query.rs
```diff
@@ -1546,11 +1522,34 @@ impl<'w, 's, D: QueryData, F: QueryFilter> Query<'w, 's, D, F> {
...
+    ///
+    /// Deprecated alias for [`many_inner`](Self::many_inner).
+    #[inline]
+    #[deprecated(note = "Use `many_inner` instead, which now returns a Result.")]
+    pub fn get_many_inner<const N: usize>(
+        self,
+        entities: [Entity; N],
+    ) -> Result<[D::Item<'w>; N], QueryEntityError<'w>> {
+        // Verify that all entities are unique
+        for i in 0..N {
```
Carter0, line 1543: Could you do something like this if you wanted to? It might be easier to understand. Totally a nit though

```
if len(x) > len(set(x)):
```

crates/bevy_ecs/src/system/query.rs
```diff
@@ -1629,63 +1645,6 @@ impl<'w, 's, D: QueryData, F: QueryFilter> Query<'w, 's, D, F> {
...
-    ///     y: f32,
-    /// }
-    ///
-    /// #[derive(Component)]
-    /// struct Force {
-    ///     x: f32,
-    ///     y: f32,
-    /// }
-    ///
-    /// fn spring_forces(spring_query: Query<&Spring>, mut mass_query: Query<(&Position, &mut Force)>){
```
Carter0, line 1661: Is it possible to move these docs to the result version of many_mut with some changes. Its seems like such a nice example!

Carter0 (approved):
I like it overall! I'm down with returning results! 

Though I actually don't really like the 'many' function name, which is a different thing than this PR is trying to deal with. I actually sorta like the name get_many more since it's an action, like we are "getting" components. But whatever haha

Oh also, doc comment stuff

chescock (approved):
Yay!

crates/bevy_ecs/src/query/state.rs
```diff
@@ -952,12 +965,12 @@ impl<D: QueryData, F: QueryFilter> QueryState<D, F> {
     /// assert_eq!(match query_state.get_many(&mut world, [wrong_entity]).unwrap_err() {QueryEntityError::EntityDoesNotExist(error) => error.entity, _ => panic!()}, wrong_entity);
     /// ```
     #[inline]
-    pub fn get_many<'w, const N: usize>(
+    pub fn many<'w, const N: usize>(
```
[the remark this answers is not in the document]
alice-i-cecile: I think it's unlikely that that change makes it into 0.16.

crates/bevy_ecs/src/system/query.rs
```diff
@@ -1546,11 +1522,34 @@ impl<'w, 's, D: QueryData, F: QueryFilter> Query<'w, 's, D, F> {
...
+    /// This consumes the [`Query`] to return results with the actual "inner" world lifetime.
+    ///
+    /// Deprecated alias for [`many_inner`](Self::many_inner).
+    #[inline]
+    #[deprecated(note = "Use `many_inner` instead, which now returns a Result.")]
+    pub fn get_many_inner<const N: usize>(
+        self,
+        entities: [Entity; N],
+    ) -> Result<[D::Item<'w>; N], QueryEntityError<'w>> {
+        // Verify that all entities are unique
```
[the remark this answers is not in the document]
alice-i-cecile: Probably, but I've opted to not change the logic in this PR. Despite the diff, this was just a pair of renames.

crates/bevy_ecs/src/system/query.rs
alice-i-cecile: Probably, but I've opted to not change the logic in this PR. Despite the diff, this was just a pair of renames.

crates/bevy_ecs/src/system/query.rs
alice-i-cecile: I like these too, and I've double-checked: we already have nice doc tests for all of the retained result versions of these methods :)

alice-i-cecile:
Man, the diff is terrible on Github. Git really does not like the renames. Strongly recommend checking out the code locally if you end up wanting to review further.

crates/bevy_ecs/src/query/state.rs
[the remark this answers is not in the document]
chescock: Oh, I know!  I wasn't even going to attempt it for 0.16 :).  

But if we expect to do it in 0.17, then it might be nice to advise users of `QueryState::get_many` to switch to `query` directly instead of switching to ``QueryState::many`` in 0.16 and then having to update that code *again* in the next release.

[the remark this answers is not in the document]
alice-i-cecile: I think that the set of users calling QueryState::get_many is small enough that I'm not concerned about that :)

Victoronz:
I think it'd be nice to have doc aliases of the old names on the new ones! Even once the old ones disappear, users might still search for the equivalent of the `get_many` that exists in std.

(I personally prefer `get_many` over `many`, but I don't feel particularly strongly about it)

Victoronz:
Hmm, thinking on it a bit more I am actually unsure about this change.

It seems the reasoning for this is the same as for `single`, because both `get_single` and `get_many` had panicking `single` and `many` counterparts, but I don't think their situations are that similar.

With `single`, there is one unambiguous behavior, and it [mirrors APIs like `first()` and `next()`](https://github.com/bevyengine/bevy/pull/1263#issuecomment-792053200), so it makes sense to be the default.
It is nicely consistent with the `Single` system parameter.

However with `get_many`, it isn't really a default, nor is there a `Many` system parameter. It is but one way of "getting many values".
The other, more preferable way would be `iter_many`. Having both `many` and `iter_many` together leaves their relationship ambiguous, as it is not clear in what way `iter_many` "builds" on `many`. 
To me, shorter names are better only as long as they don't introduce ambiguity/increase mental load!

Moreover, `get_many`'s behavior is to eagerly retrieve values for each key, which matches `get_`-style behavior.

In addition, `std` has recently decided to rename its `get_many` to `get_disjoint`, exactly because they think it **isn't** designed with "many" in mind, which our `get_many` isn't either.
Does that mean we want to rename `many` into `disjoint` later? The meaning of that without the `get` would be unclear imo.

Further, unlike usual `get` methods, the std `get_many`/`get_disjoint` method does actually return `Result`, not `Option`.


I'd personally say we remove the panicking `many` and rename `get_many` into `get_disjoint`!

But that's my few cents, the decision is still left up to you!

alice-i-cecile:
> I'd personally say we remove the panicking many and rename get_many into get_disjoint!

I'd like to avoid more renames as part of this PR: we can always tweak that later. How would you feel about just deprecating the panicking `many` methods? Trivial migration, and hits the main goals of this PR (deduplication, fewer panics).

Victoronz:
> > I'd personally say we remove the panicking many and rename get_many into get_disjoint!
> 
> I'd like to avoid more renames as part of this PR: we can always tweak that later. How would you feel about just deprecating the panicking `many` methods? Trivial migration, and hits the main goals of this PR (deduplication, fewer panics).

That'd be completely fine! They always seemed somewhat weird to me. The same criticisms already apply to the current panicking `many` methods!

crates/bevy_ecs/src/system/query.rs
```diff
@@ -182,7 +182,9 @@ use core::{
 /// # use bevy_ecs::prelude::*;
 /// # #[derive(Component)]
 /// # struct Health;
-/// # #[derive(Component)]
+/// # #[derive(Component)]>>>>>>> main
+312
+
```
bushrat011899, line 187: Looks like a leftover from a conflict resolution?

bushrat011899 (approved):
Same reasons I approved the sibling PR. Just noted there appears to be a leftover hunk from a merge conflict.

crates/bevy_ecs/src/system/query.rs
```diff
@@ -182,7 +182,9 @@ use core::{
 /// # use bevy_ecs::prelude::*;
 /// # #[derive(Component)]
 /// # struct Health;
-/// # #[derive(Component)]
+/// # #[derive(Component)]>>>>>>> main
+312
```
alice-i-cecile, line 186:
```suggestion
/// # #[derive(Component)]
```

alice-i-cecile:
On reflection, I prefer the simple deprecation approach here. Closing.

PR #18134, opened 2025-03-03
wrap EntityIndexMap/Set slices as well

Victoronz:
# Objective

Continuation of #17449.

#17449 implemented the wrapper types around `IndexMap`/`Set` and co., however punted on the slice types.
They are needed to support creating `EntitySetIterator`s from their slices, not just the base maps and sets.

## Solution

Add the wrappers, in the same vein as #17449 and #17589 before.

The `Index`/`IndexMut` implementations take up a lot of space, however they cannot be merged because we'd then get overlaps.

They are simply named `Slice` to match the `indexmap` naming scheme, but this means they cannot be differentiated properly until their modules are made public, which is already a follow-up mentioned in #17954.

### DIFF
--- a/crates/bevy_ecs/src/entity/index_map.rs
+++ b/crates/bevy_ecs/src/entity/index_map.rs
@@ -375,7 +375,7 @@ impl<V> Slice<V> {
     /// # Safety
     ///
     /// `slice` must stem from an [`IndexMap`] using [`EntityHash`].
-    /// 
+    ///
     /// [`entity::index_map::Slice`]: `crate::entity::index_map::Slice`
     pub const unsafe fn from_slice_unchecked(slice: &map::Slice<Entity, V>) -> &Self {
         // SAFETY: Slice is a transparent wrapper around indexmap::map::Slice.
@@ -387,7 +387,7 @@ impl<V> Slice<V> {
     /// # Safety
     ///
     /// `slice` must stem from an [`IndexMap`] using [`EntityHash`].
-    /// 
+    ///
     /// [`entity::index_map::Slice`]: `crate::entity::index_map::Slice`
     pub const unsafe fn from_slice_unchecked_mut(slice: &mut map::Slice<Entity, V>) -> &mut Self {
         // SAFETY: Slice is a transparent wrapper around indexmap::map::Slice.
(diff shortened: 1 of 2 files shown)


crates/bevy_ecs/src/entity/index_map.rs
```diff
@@ -197,6 +351,471 @@ where
 
 impl<V: Eq> Eq for EntityIndexMap<V> {}
 
+pub struct Slice<V, S = EntityHash>(PhantomData<S>, map::Slice<Entity, V>);
```
chescock, line 354: Why do you need the `S` parameter here (and in `index_set::Slice`)?  I don't see any `impl`s that use anything other than the default, so I don't think these can be created or used with any other type parameter.

crates/bevy_ecs/src/entity/index_map.rs
```diff
@@ -197,6 +351,471 @@ where
...
+impl<V> IntoIterator for Box<Slice<V>> {
+    type Item = (Entity, V);
+    type IntoIter = IntoIter<V>;
+
+    fn into_iter(self) -> Self::IntoIter {
+        IntoIter(self.into_boxed_inner().into_iter(), PhantomData)
+    }
+}
+
+impl<V: PartialOrd> PartialOrd for Slice<V> {
```
chescock, line 681: Can you `#[derive(PartialOrd, Ord, PartialEq, Eq, Hash)]` for these?  (And similarly for `index_set::Slice`)

crates/bevy_ecs/src/entity/index_map.rs
chescock, line 354: You need `#[repr(transparent)]` here for some of the casts to be valid.  You have it on `index_set::Slice`, just not `index_map::Slice`.

chescock (approved):

crates/bevy_ecs/src/entity/index_map.rs
Victoronz: It is the same as with the iterator types: Technically not needed yet, but without it, the type alone doesn't convey why it exists!

Later on, when more trusted combinations of `Hasher` and `Hash` are introduced (`MainEntity` + `EntityHash`), this generic can carry proper semantic meaning.

alice-i-cecile (approved):

PR #18139, opened 2025-03-03
UI Node Gradients

ickshonpe:
# Objective

Allowing drawing of UI nodes with a gradient instead of a flat color.

## Solution

The are three gradient structs corresponding to the three types of gradients supported: `LinearGradient`, `ConicGradient` and `RadialGradient`. These are then wrapped in a `Gradient` enum discriminator which has `Linear`, `Conic` and `Radial` variants. 

Each gradient type consists of the geometric properties for that gradient and a list of color stops.
Color stops consist of a color, a position or angle and an optional hint.  If no position is specified for a stop, it's evenly spaced between the previous and following stops. Color stop positions are absolute, if you specify a list of stops:
```vec![vec![ColorStop::new(RED, Val::Percent(90.), ColorStop::new(Color::GREEN, Val::Percent(10.))```
the colors will be reordered and the gradient will transition from green at 10% to red at 90%. 

Colors are interpolated between the stops in SRGB space. The hint is a normalized value that can be used to shift the mid-point where the colors are mixed 50-50.  between the stop with the hint and the following stop.

For sharp stops with no interpolated transition, place two stops at the same position.

`ConicGradient`s and RadialGradient`s have a center which is set using the new `Position` type. `Position` consists of a normalized (relative to the UI node) `Vec2` anchor point and a responsive x, y offset.

To draw a UI node with a gradient you insert the components `BackgroundGradient` and `BorderGradient`, which both newtype a vector of `Gradient`s. If you set a background color, the background color is drawn first and the gradient(s) are drawn on top.

The implementation is deliberately simple and self contained. The shader draws the gradient in multiple passes which is quite inefficient for gradients with a very large number of color stops. It's simple though and there won't be any compatibility issues. We could make gradients a specialization for `UiPipeline` but I used a separate pipeline plugin for now to ensure that these changes don't break anything. 

#### Not supported in this PR
* Interpolation in other color spaces besides SRGB. 
* Images and text: This would need some breaking changes like a `UiColor` enum type with `Color` and `Gradient` variants, to enable `BorderColor`, `TextColor`, `BackgroundColor` and `ImageNode::color` to take either a `Color` or a gradient.
* Repeating gradients

## Testing

Includes three examples that can be used for testing:
```
cargo run --example linear_gradients
cargo run --example stacked_gradients
cargo run --example radial_gradients
```

Most of the code except the components API is contained within the `bevy_ui/src/render/linear_gradients` module.
There are no changes to any existing systems or plugins except for the addition of the gradients rendering systems to the render world schedule and the `Val` changes from #18164 . 

## Showcase

gradients
stacked
rad

Conic gradients can be used to draw simple pie charts like in CSS:
PIE

### CODE REVIEW

**Comment by ickshonpe** [comment]
> # Objective
> Draw UI nodes with a gradient instead of a flat colour.
> 
> ## Solution
> New components `LinearGradient` and `LinearGradientBorder`, `LinearGradientBorder` just newtypes `LinearGradient`.
> 
> `LinearGradient` takes an angle in radians and a list of color stops. These are interpreted following CSS rules (more or less) https://developer.mozilla.org/en-US/docs/Web/CSS/gradient/linear-gradient.
> 
> The implementation is kind of rough and inefficient. The shader draws the gradient in multiple passes. Using an SSBO or gradient texture might be better, but this works for now. We could make gradients a specialization for `UiPipeline` but I used a separate pipeline plugin for now to ensure that these changes don't break anything.
> 
> #### Not supported:
> * Hints: Left out to keep this initial PR simple.
> * Radial gradients: Rendering is trivial but radial gradients need a much more complicated API to handle positioning and ellipses.
> * Images and text: This would need some very breaking changes like a `UiColor` enum type with `Color` and `Gradient` variants, to enable `BorderColor`, `TextColor`, `BackgroundColor` and `ImageNode::color` to take either a `Color` or a gradient.
> 
> ## Testing
> ```
> cargo run --example linear_gradients
> ```
> 
> ## Showcase
> srgb-interp

**Comment by alice-i-cecile** [comment]
> Images and text: This would need some very breaking changes like a UiColor enum type with Color and Gradient variants, to enable BorderColor, TextColor, BackgroundColor and ImageNode::color to take either a Color or a gradient.

If we want to do this in the future, I think we can probably keep breakage pretty low by keeping the `Color::into()` impl around to generate a flat color. A lot of users are already doing this, and it's an intuitive terse API.

**Comment by ickshonpe** [comment]
> > Images and text: This would need some very breaking changes like a UiColor enum type with Color and Gradient variants, to enable BorderColor, TextColor, BackgroundColor and ImageNode::color to take either a Color or a gradient.
> 
> If we want to do this in the future, I think we can probably keep breakage pretty low by keeping the `Color::into()` impl around to generate a flat color. A lot of users are already doing this, and it's an intuitive terse API.

Yeah that's true, I can't see any practical alternatives to an enum and it wouldn't add much friction when `RED.into()`, `GREEN_100.into()` etc are already used everywhere with the current colors API

File: crates/bevy_ui/src/lib.rs [hunk 1]
```diff
@@ -170,6 +170,9 @@ impl Plugin for UiPlugin {
             .register_type::<BoxShadowSamples>()
             .register_type::<UiAntiAlias>()
             .register_type::<TextShadow>()
+            // .register_type::<LinearGradient>()
+            // .register_type::<ColorStops>()
+            // .register_type::<ComputedColorStops>()
```
fjkorf, line 175: maybe remove these lines, or comment on why they are helpful to leave

fjkorf (approved):
Very cool PR! Your stated goals are clear and your approach makes sense. `LinearGradient` and `LinearGradientBorder` are strong additions to the UI repertoire. I was able to read your code and understand most of what you're up to here.  I appreciate the extensive comments in the .wsgl file . I was able to run the new example locally without issue. Good stuff

ickshonpe:
The only other option I thought about is adding a `Gradient` variant to the `Color` enum itself. That doesn't seem good though as it wouldn't be compatible with the non-UI rendering APIs.

mockersf:
messages above due to https://github.com/bevyengine/bevy/pull/18153

ickshonpe:
Added radial and conic gradients support, some parts of the API aren't implemented fully yet like ellipitical geometry and angular color stop angles.

fjkorf:
I pulled the latest and checked it out. The radial and conical additions look great. The bit where you use `Val::Percent` to remove the blending is interesting. I had to dig around to figure out what you were up to, so it might be worth adding a few comments around that section to explain what's up. Impressive stuff

hukasu:
would this close #18152? or would there still be work after this pr to close the issue?

crates/bevy_ui/src/render/gradient.rs
```diff
@@ -0,0 +1,984 @@
...
+    pipeline_cache: Res<PipelineCache>,
+    draw_functions: Res<DrawFunctions<TransparentUi>>,
+) {
+    let draw_function = draw_functions.read().id::<DrawGradientFns>();
+    for (index, gradient) in extracted_gradients.items.iter().enumerate() {
+        let Ok((default_camera_view, ui_anti_alias)) =
+            render_views.get_mut(gradient.extracted_camera_entity)
+        else {
+            continue;
+        };
```
atlv24, line 660: i think fallible systems let you bubble these now with `?`

crates/bevy_ui/src/render/gradient.rs
```diff
@@ -0,0 +1,984 @@
...
+                                points[if angle < TAU / 4. {
+                                    3
+                                } else if angle < TAU / 2. {
+                                    0
+                                } else if angle < 3. * TAU / 4. {
+                                    1
+                                } else {
+                                    2
+                                }]
+                                .into(),
```
atlv24, line 837: can probably be extracted to a function

crates/bevy_ui/src/render/gradient.wgsl
```diff
@@ -0,0 +1,280 @@
...
+// is outside, and zero is exactly on the boundary.
+//
+// Arguments: 
+//  - `point`        -> The function will return the distance from this point to the closest point on 
+//                    the boundary.
+//  - `size`         -> The maximum width and height of the box.
+//  - `corner_radii` -> The radius of each rounded corner. Ordered counter clockwise starting 
+//                    top left:
+//                      x: top left, y: top right, z: bottom right, w: bottom left.
+fn sd_rounded_box(point: vec2<f32>, size: vec2<f32>, corner_radii: vec4<f32>) -> f32 {
```
atlv24, line 90: is this not already present in other ui code? can it be shader imported?

crates/bevy_ui/src/render/gradient.wgsl
```diff
@@ -0,0 +1,280 @@
...
+    // Vector from the center of the radius circle to the point.
+    let q = corner_to_point + radius;
+    // Length from center of the radius circle to the point, zeros a component if the point is not 
+    // within the quadrant of the radius circle that is part of the curved corner.
+    let l = length(max(q, vec2(0.0)));
+    let m = min(max(q.x, q.y), 0.0);
+    return l + m - radius;
+}
+
+fn sd_inset_rounded_box(point: vec2<f32>, size: vec2<f32>, radius: vec4<f32>, inset: vec4<f32>) -> f32 {
```
atlv24, line 108: likewise

crates/bevy_ui/src/render/gradient.wgsl
```diff
@@ -0,0 +1,280 @@
...
+    );
+
+    if enabled(in.flags, BORDER) {
+        return draw(in, gradient_color);
+    } else {
+        return draw_background(in, gradient_color);
+    }
+}
+
+fn srgb_mix(a: vec4<f32>, b: vec4<f32>, t: f32) -> vec4<f32> {
```
atlv24, line 215: this should be in a util file (and might already exist somewhere)

crates/bevy_ui/src/ui_node.rs
```diff
@@ -2827,3 +2792,346 @@ impl Default for TextShadow {
...
+
+/// A color stop for a gradient
+#[derive(Debug, Copy, Clone, PartialEq, Reflect)]
+#[reflect(Default, PartialEq, Debug)]
+#[cfg_attr(
+    feature = "serialize",
+    derive(serde::Serialize, serde::Deserialize),
+    reflect(Serialize, Deserialize)
+)]
+pub struct ColorStop {
```
atlv24, line 2804: is this file supposed to have every single possible ui node component? might make sense to split it up a bit at some point cus 3k lines is pretty beefy

crates/bevy_ui/src/render/gradient.rs
```diff
@@ -0,0 +1,984 @@
...
+                        if stops_out[0].1.is_nan() {
+                            stops_out[0].1 = 0.;
+                        }
+                        if stops_out.last().unwrap().1.is_nan() {
+                            stops_out.last_mut().unwrap().1 = TAU;
+                        }
+
+                        let mut i = 1;
+
+                        while i < stops_out.len() - 1 {
```
atlv24, line 596: is this loop the same as the other two above? maybe extract into a function and share or highlight the differences

atlv24:
I think id want a simple gradient mode to be available, which uses vertex colors and vertex interpolation. It would be more limited for sure though

ickshonpe:
> I think id want a simple gradient mode to be available, which uses vertex colors and vertex interpolation. It would be more limited for sure though

Yeah, maybe using `UiMaterial`s.

crates/bevy_ui/src/render/gradient.rs
ickshonpe: Yep, initial implementation they did have differences but I simplified things down once I understood what I was doing and they are the same now more or less and should be extracted.

crates/bevy_ui/src/ui_node.rs
ickshonpe: Yep it's huge. Maybe a natural split would be to have all the required-for-every-node components in the `uinode` module and then make a second module called `components` (or something) for the optional components. I didn't add an extra module with this PR as it's huge enough already.

crates/bevy_ui/src/render/gradient.rs
ickshonpe: Oh this looks stupid, it doesn't need a branch here, just need to divide by four and round and shift the result

crates/bevy_ui/src/render/gradient.wgsl
ickshonpe: I wasn't able to find anything. I could move it into `color_operations.wgsl` in `bevy_render` but I'm not sure as this 
implemenation might not be ideal. Maybe the colors should be passed into the shader in srgb space? I want to add support for interpolation in other color spaces eventually so didn't want to overthink things.

crates/bevy_ui/src/render/gradient.wgsl
ickshonpe: Ah I thought that wgsl didn't allow importing functions, I didn't know about #define_import_path

crates/bevy_ui/src/render/gradient.rs
ickshonpe: `?` would return from the system and not just continue though wouldn't it? Or am I misunderstanding

atlv24: true my bad

crates/bevy_ui/src/geometry.rs
```diff
@@ -688,6 +692,100 @@ impl Default for UiRect {
...
 
+#[derive(Debug, Clone, Copy, PartialEq, Reflect)]
+#[reflect(Default, Debug, PartialEq)]
+#[cfg_attr(
+    feature = "serialize",
+    derive(serde::Serialize, serde::Deserialize),
+    reflect(Serialize, Deserialize)
+)]
+/// Responsive position relative to a UI node.
+pub enum Position {
```
atlv24, line 703: its interesting that this didnt already exist for anchoring purposes - is there a reason it doesn't exist? maybe there's another preferred solution? i have a feeling this can be represented by a Vec4, but not sure if thats more complicated or less in line with the rest of bevy ui.

ickshonpe: Yeah, this position type I've been struggling to make my mind up on. 
CSS has different position APIs depending on the type of element. 

For regular `Node` positions we don't use `Vec4`s because UI positions needs to support responsive coordinates but the position field of `Node` has four fields:
```
    pub left: Val,
    pub right: Val,
    pub top: Val,
    pub bottom: Val,
```
they can all be set and they are interpreted according to CSS rules (same for the vertical axis except left=top and right=bottom):
1. if you set left and unset right (Val::Auto) or set right and leave left unset then the node is displaced from the set edge towards the center by the set value.
2. if you set all of left, right and width, the node is displaced from the left edge by the left value and the right value is ignored.
3. if you set both edges but leave width unset, the node is placed between the two positions with its width set to the difference between them.

For the positions of gradients though CSS uses just x, y coordinates relative to a given anchor value (or top-left if not defined). If no anchor or coordinates is set, then the gradient starts at the center.

There are three options (assuming we want to match the CSS API) that make sense to me:
#### 1. Seperate anchor enum
```
pub enum NodeAnchor {
    TopLeft,
    Left,
    //.. rest of variants
    BottomRight,
}

pub struct Position {
    pub anchor: NodeAnchor,
    pub x: Val,
    pub y: Val,
}
```

It could use `Anchor` from `bevy_sprite` but `Anchor` has that `Custom` variant and sprite specific docs.
It's less pretty but you don't need deconstruct the enum to set the inner values.

#### 2. enum per axis 
```
pub enum RelativeVal {
    Start(Val),
    Center(Val),
    End(Val),
}

pub struct Position {
   x: RelativeVal
   y: RelativeVal
}
```

I like this one the most asthetically, but I thought maybe users might find it too abstract and confusing.

#### 3 Single enum with nine variants
The one this PR uses, it has the clumsiest implementation but it doesn't need a seperate anchor type and its simple to understand.

atlv24: Oh man, I just checked sprite's Anchor implementation and I disagree with it too. Here's what i would consider ideal:

Anchor is not an enum, but a Vec2 newtype. it is accompanied by 9 associated constants, the topleft/bottomright etc values we all know and love. There is no resolve function, the values themselves are already resolved. Custom anchors are constructed by new instead of associated const.

Gradient uses Anchor + Vec2 relative_position. Again there is no enum and no resolve, stuff is initialized from associated const.

My disagreement with enums here is that Custom represents all variants, which renders them redundant. It's like having an enum Number with variants One, Two, Three, and TheRest(i64). Doesn't make much sense.

atlv24 (approved):
Looks great, thanks!

extrawurst:
@alice-i-cecile is there any chance to get this still in for 0.16? Would love to see that - what can I do to help?

alice-i-cecile:
This has two approvals so it's ready to merge but I'm reluctant to ship a feature after the initial release candidate has been cut.

extrawurst:
FWIW I tested the examples, works great for me.

NthTensor (changes requested):
This PR was created before we implemented the new release notes process, and so hasn't gotten the right pings by the bot. I'm re-adding the tags so that you'll get the new instructions.

NthTensor (approved):
Release content looks good.

Minor nit that I was not aware of before reading the new content:

> Colors are interpolated between the stops in SRGB space.

Srgb is almost certainly the *wrong* space to do any kind of color gradient or interpolation in, except in the case where you are trying to match another program (which also does it wrong). This should ideally be (a) configurable and (b) default to OkLab.

ickshonpe:
> Release content looks good.
> 
> Minor nit that I was not aware of before reading the new content:
> 
> > Colors are interpolated between the stops in SRGB space.
> 
> Srgb is almost certainly the _wrong_ space to do any kind of color gradient or interpolation in, except in the case where you are trying to match another program (which also does it wrong). This should ideally be (a) configurable and (b) default to OkLab.
> 
> I consider this feedback non-blocking, but I did want it to be registered.

I went with SRGB, despite the transitions looking so dull and muddy, as it's the default for CSS which the api here is roughly based on. I'd definitely be on board with making Bevy default to OkLab though. I'm kind of fighting myself not to make more changes here and add multiple color space support etc but I've got so many PRs in review atm need to get some of them merged hehe.

NthTensor:
Yeah, totally fair let's look into this as follow up.

PR #18140, opened 2025-03-03
Relationship(…Target) html trait tag

### LINKED ISSUE #18117: Relationship / RelationshipTarget traits should get their own HTML tag

This was introduced in https://github.com/bevyengine/bevy/pull/17758: `Relationship` and `RelationshipTarget` should be included too.

SpecificProtagonist:
# Objective

Fixes #18117

These are component subtraits, but unlike for `Event` (before that bound got removed) this still shows it as a component because it's actually used as such.

## Testing

```sh
 RUSTDOCFLAGS="--html-after-content docs-rs/trait-tags.html --cfg docsrs_dep" RUSTFLAGS="--cfg docsrs_dep" cargo doc --no-deps --package bevy_ecs
```

---

## Showcase

Screenshot from 2025-03-03 17-31-24

### DIFF
--- a/docs-rs/trait-tags.html
+++ b/docs-rs/trait-tags.html
@@ -16,6 +16,8 @@
         'ScheduleLabel',
         'SystemSet',
         'SystemParam',
+        'Relationship',
+        'RelationshipTarget'
     ];
 
     // Find all traits that are implemented by the current type.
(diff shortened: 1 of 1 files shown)


alice-i-cecile (approved):
I love how easy this was. Really nice stuff!

fjkorf (approved):
This makes sense. Thanks!

PR #18147, opened 2025-03-03
Deprecate `insert_or_spawn` function family

ElliottjPierce:
# Objective

Based on #18054, this PR builds on #18035 to deprecate:

- `Commands::insert_or_spawn_batch`
- `Entities::alloc_at_without_replacement`
- `Entities::alloc_at`
- `World::insert_or_spawn_batch`
- `World::insert_or_spawn_batch_with_caller`

## Testing

Just deprecation, so no new tests. Note that as of writing #18035 is still under testing and review.

## Open Questions

- [x] Should `entity::AllocAtWithoutReplacement` be deprecated? It is internal and only used in `Entities::alloc_at_without_replacement`.  **EDIT:** Now deprecated.

## Migration Guide

The following functions have been deprecated:

- `Commands::insert_or_spawn_batch`
- `World::insert_or_spawn_batch`
- `World::insert_or_spawn_batch_with_caller`

These functions, when used incorrectly, can cause major performance problems and are generally viewed as anti-patterns and foot guns. These are planned to be removed altogether in 0.17.

Instead of these functions consider doing one of the following:

Option A) Instead of despawing entities and re-spawning them at a particular id, insert the new `Disabled` component without despawning the entity, and use `try_insert_batch` or `insert_batch` and remove `Disabled` instead of re-spawning it.

Option B) Instead of giving special meaning to an entity id, simply use `spawn_batch` and ensure entity references are valid when despawning.


alice-i-cecile:
>  Should entity::AllocAtWithoutReplacement be deprecated? It is internal and only used in Entities::alloc_at_without_replacement. Currently not deprecated.

Mark this as deprecated so it's easier to find and remove later please :)

alice-i-cecile (changes requested):
Can you link the issue number in the deprecation notice please? I'd like to funnel affected users to it, and it's a much more compelling explanation than a bare assertion that this is slow.

ElliottjPierce:
> Can you link the issue number in the deprecation notice please? I'd like to funnel affected users to it, and it's a much more compelling explanation than a bare assertion that this is slow.

Great idea! I've added the issue number. If you like, I'd be happy to change it to a full link too.

alice-i-cecile (approved):

Carter0 (approved):
LGTM

alice-i-cecile:
Merging the prerequisite PR, then this.

PR #18148, opened 2025-03-03
Remove `insert_or_spawn` function family

### LINKED ISSUE #18054: Improve Performance of Entities::alloc_at

ElliottjPierce:
# Objective

Based on and closes #18054, this PR builds on #18035 and #18147 to remove:

- `Commands::insert_or_spawn_batch`
- `Entities::alloc_at_without_replacement`
- `Entities::alloc_at`
- `entity::AllocAtWithoutReplacement`
- `World::insert_or_spawn_batch`
- `World::insert_or_spawn_batch_with_caller`

## Testing

Just removing unused, deprecated code, so no new tests. Note that as of writing, #18035 is still under testing and review.

## Future Work

Per [this](https://github.com/bevyengine/bevy/issues/18054#issuecomment-2689088899) comment on #18054, there may be additional performance improvements possible to the entity allocator now that `alloc_at` no longer is supported. At a glance, I don't see anything obvious to improve, but it may be worth further investigation in the future.


ElliottjPierce:
removed `blocked` since #18147 is merged.

atlv24 (approved):

Henauxg (approved):
There was a comment from `Cart` in #18054 saying

> If we plan on removing support for this, we should also look into some of the "invalid checks" in Entities. Iirc we added additional checking to support this scenario, and we might be able to simplify / optimize some things. Worth digging up the PRs that introduced those changes.

Did someone have a look, if not we could maybe just track this somewhere ?

ElliottjPierce:
> Did someone have a look, if not we could maybe just track this somewhere ?

Yup! I've rewritten most of the `Entities` allocator 9 times now lol. I'm picturing building from #18670 (which this is blocking). If we don't do v9, I'll take another look at what we end up with.

Also marking as ready for final since it has two approvals. (I think that's how that's supposed to work.)

NthTensor:
Please commit your migration guide as a file following the instructions above.

ElliottjPierce:
> Please commit your migration guide as a file following the instructions above.

Oops. Thanks for pointing that out. Migration guide has been migrated lol.

PR #18153, opened 2025-03-04
rendering regression: mention in pr comment that the solution could be to update the pr

mockersf:
# Objective

- When a PR gets merged that modifies the rendering screenshots, the main reference will be updated
- Every in-flight PR will then "fail" rendering change detection as they come from an outdated main branch

## Solution

- Suggest updating the PR to the latest main branch


mnmaita (approved):

alice-i-cecile (approved):
Perfect, that's a useful suggestion.

PR #18156, opened 2025-03-04
Add `LeafNode` to forbid adding `Node` children to certain `Node`s

### LINKED ISSUE #18015: UI Text nodes overlap when they have children

bytemunch:
# Objective

Forbid adding children with `Node` components to certain entities, namely `Text` components.

Closes #18015

Related #17551
Prior conversation #18112

## Solution

Add a new marker component and accompanying system that prevents the addition of children with a `Node` component to any entity with the `LeafNode` component.

I've opted for removal/forbid over a warning as this is the same behavior as self-referencing relationships.

The validation system runs in `PreUpdate` but should be a couple no-op loops most of the time, as it removes the relationship it queries for when run. 

## Testing

<details>
  <summary>Click to view test code</summary>

```rust
use bevy::{input::common_conditions::input_just_pressed, prelude::*};

fn main() {
    App::new()
        .add_plugins(DefaultPlugins)
        .add_systems(Startup, (setup, setup_insert))
        .add_systems(
            Update,
            insert_test.run_if(input_just_pressed(MouseButton::Left)),
        )
        .run();
}

    commands.spawn(Camera2d);
    let text = commands.spawn(Text::new("Hello, bevy!")).id();

    commands
        .spawn(Node::default())
        .insert(ChildOf { parent: text }); // Removal triggers here.

    commands
        .spawn(Text::new("Hello, bevy!"))
        .with_children(|cb| {
            cb.spawn(Node::default()); // Removal triggers here.
        });
}

fn setup_insert(mut commands: Commands) {
    commands
        .spawn(Text::new("Hello, bevy!"))
        .with_children(|cb| {
            cb.spawn_empty().with_children(|cb| {
                cb.spawn(Node::default()); // Removal triggers on grandchild.
            });
        });
}

fn insert_test(mut commands: Commands, q: Query<Entity, With<ChildOf>>) {
    for e in &q {
        commands.entity(e).insert(Node::default()); // Removal triggers here.
    }
}
```

</details>

Local CI + typos

---

## Migration Guide

Relationships where entities with `Node` components have a `LeafNode` / `Text` in their ancestry will be broken with this change. Ensure that any currently used `Text` components do not have `Node` children.

### DIFF
--- a/crates/bevy_ui/src/ui_node.rs
+++ b/crates/bevy_ui/src/ui_node.rs
@@ -1,7 +1,7 @@
 use crate::{FocusPolicy, UiRect, Val};
 use bevy_color::Color;
 use bevy_derive::{Deref, DerefMut};
-use bevy_ecs::{prelude::*, system::SystemParam};
+use bevy_ecs::{component::HookContext, prelude::*, system::SystemParam, world::DeferredWorld};
 use bevy_math::{vec4, Rect, UVec2, Vec2, Vec4Swizzles};
 use bevy_reflect::prelude::*;
 use bevy_render::{
@@ -672,12 +672,19 @@ impl Default for Node {
     }
 }
 
-/// Marker struct for `Node`s that should not have `Node`s as children.
+/// Marker struct for `Node`s that can't have `Node`s as children.
+/// If you need to have a `Node` with a relationship to a `LeafNode`, look into using a parent
+/// entity and having the two `Node`s as siblings.
 #[derive(Component, Debug, Default, Clone, Reflect)]
 #[reflect(Component, Default, Debug)]
+#[component(on_remove=on_remove_leaf_node)]
 pub struct LeafNode;
 
-/// Keeps `LeafNode` entities from having `Node` entities as children.
+fn on_remove_leaf_node(_world: DeferredWorld, ctx: HookContext) {
+    warn!("Removing `LeafNode` from {}. This will allow you to add child `Node`s but will result in unexpected behavior.", ctx.entity);
+}
+
+/// Prevents `LeafNode` entities from having `Node` entities as children.
 pub fn validate_leaf_nodes(
     mut commands: Commands,
     children: Query<&Children>,
--- a/crates/bevy_ui/src/ui_node.rs
+++ b/crates/bevy_ui/src/ui_node.rs
@@ -672,6 +672,28 @@ impl Default for Node {
     }
 }
 
+/// Marker struct for `Node`s that should not have `Node`s as children.
+#[derive(Component, Debug, Default, Clone, Reflect)]
+#[reflect(Component, Default, Debug)]
+pub struct LeafNode;
+
+/// Keeps `LeafNode` entities from having `Node` entities as children.
+pub fn validate_leaf_nodes(
+    mut commands: Commands,
+    children: Query<&Children>,
+    leaf_nodes: Query<Entity, With<LeafNode>>,
+    nodes: Query<Entity, Added<Node>>,
+) {
+    for leaf_node in &leaf_nodes {
+        for child in children.iter_descendants(leaf_node) {
+            if let Ok(node) = nodes.get(child) {
+                warn!("Entity {leaf_node} is marked as a `LeafNode` but has entity {node} containing `Node` as a descendant. The invalid relationship has been removed.");
+                commands.entity(node).remove::<ChildOf>();
+            }
+        }
+    }
+}
+
 /// Used to control how each individual item is aligned by default within the space they're given.
 /// - For Flexbox containers, sets default cross axis alignment of the child items.
 /// - For CSS Grid containers, controls block (vertical) axis alignment of children of this grid container within their grid areas.
--- a/crates/bevy_ui/src/lib.rs
+++ b/crates/bevy_ui/src/lib.rs
@@ -198,6 +198,7 @@ impl Plugin for UiPlugin {
         app.add_systems(
             PostUpdate,
             (
+                validate_leaf_nodes.before(TransformSystem::TransformPropagate),
                 update_ui_context_system.in_set(UiSystem::Prepare),
                 ui_layout_system_config,
                 ui_stack_system
--- a/crates/bevy_ui/src/widget/text.rs
+++ b/crates/bevy_ui/src/widget/text.rs
@@ -1,6 +1,6 @@
 use crate::{
-    ComputedNode, ComputedNodeTarget, ContentSize, FixedMeasure, Measure, MeasureArgs, Node,
-    NodeMeasure,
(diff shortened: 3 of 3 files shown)


crates/bevy_ui/src/ui_node.rs
```diff
@@ -672,6 +672,35 @@ impl Default for Node {
...
+    mut commands: Commands,
+    children: Query<&Children>,
+    leaf_nodes: Query<Entity, With<LeafNode>>,
+    nodes: Query<Entity, Added<Node>>,
+) {
+    for leaf_node in &leaf_nodes {
+        for child in children.iter_descendants(leaf_node) {
+            if let Ok(node) = nodes.get(child) {
+                warn!("Entity {leaf_node} is marked as a `LeafNode` but has entity {node} containing `Node` as a descendant. The invalid relationship has been removed.");
+                commands.entity(node).remove::<ChildOf>();
```
ickshonpe, line 698: I think the marker component is the definitely the right approach here but I'm not sure about removing `Childof`. Each removed child then becomes a root UI entity which might be messy. Maybe we could deactivate the nodes somehow, perhaps forcing their corresponding taffy node to `Display::None` could work?

bytemunch: Oh my bad, default is auto, and default nodes seem not to affect the text display.

bytemunch: I'm seeing what you mean with the headache part now, I've just tried quickly reading `ui_layout_system`, I might need to give myself more time to digest how that's all working! I'm pretty sure I'll be able to block the recursion if the node has the leaf marker, just gotta understand the system first :)

bytemunch: Everything I've tried in layout is having no effect on the text display being blank with `Node` children. Probably because I'm not well versed enough in this area of the codebase, but is it possible this is a rendering thing? That would at least provide reason for `Display::None` not clearing up the issue.

Sorry to pick your brain @ickshonpe, but [previously](https://github.com/bevyengine/bevy/issues/17551#issuecomment-2618950685) you said
>If a node with a measure func has a child then IIRC the measure func is just ignored, which with text could result in not enough space being assigned to fit any of the words or with images it might stretch the image violating its aspect ratio.

Do you know if this still stands, and if so where I can find/start hunting for the code path that does this?

ickshonpe: Sorry been a bit distracted the last couple of days. Measure funcs are injected into the Taffy layout tree and called during `compute_layout` to determine the space to allocate for content. 

Maybe we could mark entities descended from a `LeafNode` with an `IgnoreNode` component. Then in `ui_layout_system` remove any taffy nodes associated to an entity with an `IgnoreNode` component.

ickshonpe: Maybe we don't need an `IgnoreNode` component even, maybe just orphaning the associated child taffy nodes of `LeafNode` entities would be enough.

ickshonpe: Not tested it but I think this is an approach that makes sense:
1. Find any child UI nodes of the `LeafNode`. 
2. Find the associated Taffy nodes of the child nodes and unparent them.
3. In `update_uinode_geometry_recursive` add `Has<LeafNode>` to `node_transform_query`. If a node has `LeafNode`, then switch to a second function that recursively walks its descendents setting their `ComputedNode` to zero size if they are UI nodes. 

We'd also need to make sure if the `LeafNode` is removed the Taffy nodes are reparented. 

Unparenting the taffy nodes means those nodes will be ignored completely when we recompute the layout until they are are reparented. Unparenting is much better than removal which would require some complicated synchronisation to handle UI nodes without an associated Taffy node.

bytemunch:
This is a bit more than I can handle at the moment, @ickshonpe could you put a help wanted label on please, or would it be better to close this for now?

Freyja-moth:
This isn't needed for now, but it might be a good idea to make this extensible for all relationships.

Something like this.

```rust
#[derive(Component)]
#[relationship(relationship_target = RelationshipTarget, leaf_node = <ComponentName>)]
pub struct Relationship(Entity);

#[derive(Component)]
#[relationship_targer(relationship = Relationship)]
pub struct RelationshipTarget(Vec<Entity>);
```

Where the macro adds a on_insert component hook that checks if the entity already has the relationship_target.

PR #18157, opened 2025-03-05
Remove lifetime from `QueryEntityError`

bushrat011899:
# Objective

- Allow `Query` methods such as `Query::get` to have their error short-circuited using `?` in systems using Bevy's `Error` type

## Solution

- Removed `UnsafeWorldCell<'w>` from `QueryEntityError` and instead store `ArchetypeId` (the information the error formatter was extracting anyway).
- Replaced trait implementations with derives now that the type is plain-old-data.

## Testing

- CI

---

## Migration Guide

- `QueryEntityError::QueryDoesNotMatch.1` is of type `ArchetypeId` instead of `UnsafeWorldCell`. It is up to the caller to obtain an `UnsafeWorldCell` now.
- `QueryEntityError` no longer has a lifetime parameter, remove it from type signatures where required.

## Notes

This was discussed on Discord and accepted by Cart as the desirable path forward in [this message](https://discord.com/channels/691052431525675048/749335865876021248/1346611950527713310). Scroll up from this point for context.

### DIFF
--- a/crates/bevy_ecs/src/query/error.rs
+++ b/crates/bevy_ecs/src/query/error.rs
@@ -33,10 +33,7 @@ impl core::fmt::Display for QueryEntityError {
     fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
         match *self {
             Self::QueryDoesNotMatch(entity, archetype_id) => {
-                write!(
-                    f,
-                    "The query does not match entity {entity}, which is a member of archetype {archetype_id:?}"
-                )
+                write!(f, "The query does not match entity {entity}")
             }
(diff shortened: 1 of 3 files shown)


bushrat011899:
Added to 0.16 under the assumption that this is desired for better ergonomics with the new result-systems functionality. Also marking as `X-Blessed` based on the linked Discord messages with Cart.

bushrat011899:
> I don't think printing the archetype id is useful.
Fair point! As stated, users have the power to inspect the archetype themselves as required.

kristoff3r (approved):

alice-i-cecile:
Much nicer; this has always bugged me.

PR #12090, opened 2024-02-24
Consolidate together Bevy's TaskPools

james7132:
# Objective
Fixes #1907. Spiritual successor to #4740.

Bevy currently creates a 50%/25%/25% split between the Compute, AsyncCompute, and IO task pools, meaning that any given operation can only be scheduled onto that subset of threads. This is suboptimal in the cases where apps are not using any IO or Async Compute, or vice versa, where available parallelism would be under utilized due to the split not reflecting the actual use case. This PR aims to fix that under utilization.

## Solution

 * Do away with the IO and AsyncCompute task pools and allocate all of the threads to the ComputeTaskPool.
 * Move all of the non-blocking IO tasks to the Compute task pool.
 * Add `TaskPool::spawn_blocking` as an alternative for blocking IO operations and any task that would previously be spawned on the AsyncCompute task pool. This will be backed by `blocking`'s dynamically scaled thread pool, which will spin up and down threads depending on the need.

This allows ECS systems and parallel iterators to be scheduled onto all CPU cores instead of artificially constraining them to half of the logical cores available, as well as for typical async IO tasks to interweave onto any thread.

The use of `spawn_blocking` to perform CPU-bound operations will have to rely on the OS's preemptive scheduler to properly schedule the threads when the main task pool's threads are sitting idle. This comes with potential context switching costs, but is generally preferable to choking out the entire app due the task not cooperatively yielding available threads or artificially limiting available parallelism.

Note: We're already using `blocking` through `async-fs` for loading assets from disk. This shift primarily moves all of the other blocking IO tasks, and the async compute tasks into `blocking`'s thread pool.

---

## Changelog
Added: `TaskPool::spawn_blocking`
Added: `TaskPool::spawn_blocking_async`
Removed: `IoTaskPool`
Removed: `AsyncComputeTaskPool`
Changed: `ComputeTaskPool` by default now spawns a thread for every available logical CPU core.

## Migration Guide
`IoTaskPool` and `AsyncComputeTaskPool` have been removed and merged with `ComputeTaskPool`. Replace `IoTaskPool::get` and `AsyncComputeTaskPool::get`.

If you were spawning futures that are reliant on non-blocking IO (as in the task spends most of it's time yielding via `await`), just spawn the task onto the `ComputeTaskPool`:

```rust
// in 0.13
IoTaskPool::get().spawn(async move {
     while let Ok(item) = stream.next().await {
         // process the item here
     }
});
// in 0.14
ComputeTaskPool::get().spawn(async move {
[...]

PR #4740, opened 2022-05-13
Internalize task distinction into TaskPool

james7132:
# Objective
Fixes #1907. IO and Async Compute threads are sitting idle even when the compute threads are at capacity. This results in poor thread utilization.

## Solution
Move the distinction between compute, async compute, and IO threads out of userspace, and internalize the distinction into `TaskPool`. This creates three tiers of threads and tasks: one for each group. Higher priority tasks can run on lower priority threads, if and only if they're currently idle, but the lower priority threads will prioritize their specific workload over higher priority ones. Priority goes `compute > io > async compute`.

For example, heavy per-frame compute workload, async compute and IO otherwise sitting idle: compute tasks will be scheduled onto async compute and IO threads whenever idle. Any IO task that is awoken will take precedence over the compute tasks.

In another case, a heavy IO workload will schedule onto idle async compute threads, but will never use a compute thread. This prevents lower priority tasks from starving out higher priority tasks.

The priority scheme chosen assumes well-behaved scheduling that adheres to the following principles:

 - Compute tasks are latency sensitive and generally do not hold the thread indefinitely or at the very minimum for time longer than the course of a frame.
 - IO tasks are not CPU intensive and will do minimal processing. Generally will yield readily at any incoming or outbound communication, and thus will only generally be under consistent heavy load only in specific circumstances.
 - Async compute tasks may run for a very long time, across multiple frames, and will generally not yield at any point during a given task's execution. Only under heavy load when a developer explicitly needs it, and can otherwise be used for IO or compute in the general case.

TODO:
 - [x] Add two additional executors to `TaskPool`
 - [x] Set up TaskPool threads to run tasks from all 3 with some prioritization
 - [x] Add user-facing APIs for choosing which priority to spawn tasks under.
 - [x] Remove the usages newtypes. Fix up dependent code.
 - [x] Properly set up individual thread groups and their individual prioritization schemes.  
 - [x] Update docs.

---

## Changelog
Changed: Newtypes of `TaskPool` (`ComputeTaskPool`, `AsyncComputeTaskPool`, and `IoTaskPool`) have been replaced with `TaskPool::spawn_as` with the appropriate `TaskGroup`. Thread utilization across these segments has been generally been increased via internal task prioritization. For convenience, `TaskPool::spawn` will always spawn on the compute task group.

## Migration Guide
[...]

PR #18163, opened 2025-03-05
Consolidate together Bevy's TaskPools [adopted]

### LINKED ISSUE #1907: Using ComputeTaskPool for a par_for_each query only uses half of available logical cores
## Bevy version

0.5

## Operating system & version

Linux

## What you did

I have a CPU heavy task and running it as part of a query `par_for_each` loop in a `ComputeTaskPool` and only half of my cores are being utilised. I originally used an AsyncTaskPool but this has even fewer threads assigned.

```rust
use bevy::tasks::ComputeTaskPool;
pub fn how_many_threads(pool: Res<ComputeTaskPool>, query: Query<(&DummyComponent)>) {
    info!("TaskPool threads: {}", pool.thread_num());
    info!("Available cores: {}", num_cpus::get());

    query.par_for_each_mut(&pool, 1, |_dummy| {
        // Will only run on
    });
}
```
> TaskPool threads: 6
> Available cores: 12

I know why this happens but it was unexpected:
https://github.com/bevyengine/bevy/blob/75ae20dc4a11f04fb0aad62e2a92ffe2b974ea89/crates/bevy_core/src/task_pool_options.rs#L54

## What you expected to happen

When not using the `IoTaskPool` or `AsyncComputeTaskPool` I would expect the `ComputeTaskPool` to use all logical cores. Likewise I would expect the other pools to expand to optimally use the CPU available.

Alternatively, documentation should be updated to make clear what the current behaviour is and steps to create a new task pool.

## What actually happened

Only half of the available core were used.

hymm:
> This is an adoption of #12090

## Changes from #12090

* Added a `spawn_blocking` and `spawn_blocking_async` to scope. This allows the gltf plugin to not block the compute task pool when running tasks.

## Todo

* [ ] figure out what to do with setting the environment variable for limiting the number of blocking threads. `set_var` is now unsafe and the api in the old pr could potentially set the environment variable from multiple threads. I may just expose setting the environment variable in the docs.
* [ ] test loading a large gltf
* [ ] Think a little bit more about the transmute in `Scope::spawn_blocking` and make sure that it is sound.
* [ ] Review the migration guide and check for any necessary change from the merge with main.

# Objective
 Fixes #1907. Spiritual successor to #4740.
 
 Bevy currently creates a 50%/25%/25% split between the Compute, AsyncCompute, and IO task pools, meaning that any given operation can only be scheduled onto that subset of threads. This is suboptimal in the cases where apps are not using any IO or Async Compute, or vice versa, where available parallelism would be under utilized due to the split not reflecting the actual use case. This PR aims to fix that under utilization.
 
 ## Solution
 * Do away with the IO and AsyncCompute task pools and allocate all of the threads to the ComputeTaskPool.
 * Move all of the non-blocking IO tasks to the Compute task pool.
 * Add `TaskPool::spawn_blocking` as an alternative for blocking IO operations and any task that would previously be spawned on the AsyncCompute task pool. This will be backed by `blocking`'s dynamically scaled thread pool, which will spin up and down threads depending on the need.
 
 This allows ECS systems and parallel iterators to be scheduled onto all CPU cores instead of artificially constraining them to half of the logical cores available, as well as for typical async IO tasks to interweave onto any thread.
 
 The use of `spawn_blocking` to perform CPU-bound operations will have to rely on the OS's preemptive scheduler to properly schedule the threads when the main task pool's threads are sitting idle. This comes with potential context switching costs, but is generally preferable to choking out the entire app due the task not cooperatively yielding available threads or artificially limiting available parallelism.
 
 Note: We're already using `blocking` through `async-fs` for loading assets from disk. This shift primarily moves all of the other blocking IO tasks, and the async compute tasks into `blocking`'s thread pool.
 
 ## Changelog
 Added: `TaskPool::spawn_blocking` Added: `TaskPool::spawn_blocking_async` Removed: `IoTaskPool` Removed: `AsyncComputeTaskPool` Changed: `ComputeTaskPool` by default now spawns a thread for every available logical CPU core.
 
 ## Migration Guide
 `IoTaskPool` and `AsyncComputeTaskPool` have been removed and merged with `ComputeTaskPool`. Replace `IoTaskPool::get` and `AsyncComputeTaskPool::get`.
 
 If you were spawning futures that are reliant on non-blocking IO (as in the task spends most of it's time yielding via `await`), just spawn the task onto the `ComputeTaskPool`:
 
 ```rust
 // in 0.13
 IoTaskPool::get().spawn(async move {
      while let Ok(item) = stream.next().await {
          // process the item here
      }
 });
 // in 0.14
 ComputeTaskPool::get().spawn(async move {
      while let Ok(item) = stream.next().await {
          // process the item here
      }
 });
 ```
 
 If you were spawning futures that are reliant on **blocking** IO (i.e. `std::fs::File::read_to_string`), use `ComputeTaskPool::spawn_blocking` instead. This will spawn and/or reuse one of the dynamically scaled threads explicitly made for blocking operations.
 
 ```rust
 // in 0.13
 IoTaskPool::get().spawn(async move {
      let contents = File::open(path).unwrap().read_to_string().unwrap();
      // process the file contents here
 });
 // in 0.14
 ComputeTaskPool::get().spawn_blocking(move || {
      let contents = File::open(path).unwrap().read_to_string().unwrap();
      // process the file contents here
});
```
 
If you were spawning futures for async compute, use `ComputeTaskPool::spawn_blocking` instead. This will spawn and/or reuse one of the dynamically scaled threads explicitly made for blocking operations.
 
If you were spawning futures that are reliant on **blocking** IO (i.e. `std::fs::File::read_to_string`), use `ComputeTaskPool::spawn_blocking` instead. This will spawn and/or reuse one of the dynamically scaled threads explicitly made for blocking operations.
 
```rust
// in 0.13
AsyncComputeTaskPool::get().spawn(async move {
     solve_traveling_salesman();
});
// in 0.14
ComputeTaskPool::get().spawn_blocking(move || {
     solve_traveling_salesman();
});
```

If you were spawning futures for long running tasks that used both blocking and non-blocking work, use `ComputeTaskPool::spawn_blocking_async` instead.
 
```rust
// in 0.13
AsyncComputeTaskPool::get().spawn(async move {
     let contents = async_fs::File::open("assets/traveling_salesman.json")
         .await
         .unwrap()
         .read_to_string()
         .await
         .unwrap();
     solve_traveling_salesman(contents);
});
// in 0.14
ComputeTaskPool::get().spawn_blocking_async(async move {
     let contents = async_fs::File::open("assets/traveling_salesman.json")
         .await
         .unwrap()
         .read_to_string()
         .await
         .unwrap();
     solve_traveling_salesman(contents);
});
> ```


hymm:
So I looked into the regression with loading assets and it seems like the issue is that the ImageLoader is doing longer running operations in `load`. Specifically the call to `Image::from_buffer` takes a decent amount of time. Besides that loading a large gltf queues up a large amount of image loading tasks, so we end up starving the schedules of any threads to run tasks.

There are a few possible solutions here:

1. Change all asset loaders to use blocking threads. Probably not a good idea as the blocking threads don't have a proper cooperative async executor. Might be worth trying and seeing what the extra overhead is. We'd be spawning a new thread for every task, up to the max threads limit, so could be expensive memory wise too.

2. Keep the IoTaskPool, but change the TaskPoolBuilder, so that it'll allow overprovisioning the threads. We would set the ComputeTaskPool to provision threads for all the logical processors, and set the IoTaskPool to be some percentage of that.

3. Just fix the image loader to use blocking threads for it's longer running tasks. This requires creating an async scope as we need to free up the thread that `load` is called on. This would put the burden on the asset loader authors for making their asset loaders behave properly.

My inclination here is to go with (2), and rework this pr to remove the AsyncComputeTaskPool. Long running async tasks should be using separate threads from the task pool. (3) probably isn't a good idea to assume that the loaders are always be written to behave. (1) seems like a bad idea for the reasons listed above.

There might be a (4) with some type of task priority system, but that requires more extensive executor changes. I'll probably explore overprovisioning threads and put up a pr for it if it looks ok.

Elabajaba:
image

Trace of Bistro with compressed textures for this PR, for context on the asset loading regression.

cart:
Yeah the "separate pools" model was solving a real problem that I don't think we can just fully ignore. I like (2).

cart:
Although I see no reason why IO is any different than "async compute". Seems like plenty of "async compute" cases could also benefit from a pooled model?

mockersf:
adopted in #20699

PR #18164, opened 2025-03-05
Extra `Val` test cases

ickshonpe:
# Objective

Just adds some extra test cases for `Val` now, as suggested by greeble-dev.

### DIFF
--- a/release-content/migration-guides/val_resolve_scale_factor_support.md
+++ b/release-content/migration-guides/val_resolve_scale_factor_support.md
@@ -1,6 +1,0 @@
----
-title: `Val::resolve` scale factor support
-pull_requests: [18164]
----
-
-`Val::resolve` now has a scale factor parameter. To resolve a `Val` to a logical value use a scale factor of `1.`.


crates/bevy_ui/src/geometry.rs
```diff
@@ -255,15 +255,19 @@ pub enum ValArithmeticError {
...
-    /// Returns a [`ValArithmeticError::NonEvaluable`] if the [`Val`] is impossible to resolve into a concrete value.
+    /// Resolves this [`Val`] to a value in physical pixels from the given `scale_factor`, `parent_size`,
+    /// and `viewport_size`.
     ///
-    /// **Note:** If a [`Val::Px`] is resolved, its inner value is returned unchanged.
-    pub fn resolve(self, parent_size: f32, viewport_size: Vec2) -> Result<f32, ValArithmeticError> {
+    /// Returns a [`ValArithmeticError::NonEvaluable`] if the [`Val`] is impossible to resolve into a concrete value.
+    pub fn resolve(
+        self,
+        scale_factor: f32,
```
greeble-dev, line 264: I'm not familiar with UI code and it took me a second to work out what `scale_factor` did. Maybe worth a rename and perhaps a comment mentioning it's related to physical/logical.

```suggestion
        pixel_scale_factor: f32,
```

greeble-dev (approved):
I'm not a UI expert, but I'm clicking approve as I can't see anything wrong - just two optional suggestions.

crates/bevy_ui/src/geometry.rs
ickshonpe: Yep scale factor has always been a bit difficult to deal with. 
There are multiple scale factors: the scale factor(s) of the render targets and also Bevy's global UI scale factor which is set using the `UiScale` resource. So if you create a UI node with `width: Val::Px(x)` then (assuming there's no scaling by the transform and the layout doesn't place any other constraints on the width) the final width in physical pixels is calculated as `x * render_target_scale_factor * ui_scale`.

viridia:
It looks like some of these changes might already have been merged into main?

Looking at the diffs for this PR, all I see is a couple of examples and a release note.

ickshonpe:
> It looks like some of these changes might already have been merged into main?
> 
> Looking at the diffs for this PR, all I see is a couple of examples and a release note.

Yep, the functionality was worked into some other PR that needed it. Left it open for the extra test cases.

mockersf (approved):

PR #18165, opened 2025-03-05
Fix Component require() IDE integration

cart:
# Objective

Component `require()` IDE integration is fully broken, as of #16575.

## Solution

This reverts us back to the previous "put the docs on Component trait" impl. This _does_ reduce the accessibility of the required components in rust docs, but the complete erasure of "required component IDE experience" is not worth the price of slightly increased prominence of requires in docs.

Additionally, Rust Analyzer has recently started including derive attributes in suggestions, so we aren't losing that benefit of the proc_macro attribute impl.

### DIFF
--- a/crates/bevy_core_pipeline/src/motion_blur/mod.rs
+++ b/crates/bevy_core_pipeline/src/motion_blur/mod.rs
@@ -9,10 +9,7 @@ use crate::{
 use bevy_app::{App, Plugin};
 use bevy_asset::{load_internal_asset, weak_handle, Handle};
 use bevy_ecs::{
-    component::{require, Component},
-    query::With,
-    reflect::ReflectComponent,
-    schedule::IntoSystemConfigs,
+    component::Component, query::With, reflect::ReflectComponent, schedule::IntoSystemConfigs,
 };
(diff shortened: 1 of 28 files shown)


alice-i-cecile (approved):

bushrat011899 (approved):

aloucks:
I don't see any IDE integration of required components after this change. For example, hovering over a `Mesh3d`:

Screenshot 2025-03-25 at 5 51 18 PM

Is it displayed somewhere else?

Why not add the generated docs **both** the struct and the `Component` impl?

cart:
> Why not add the generated docs both the struct and the Component impl?

The crux of it is that adding it to the struct required a "hack" where we layered an additional attribute proc maco on top of the normal derive macro. This broke all Rust Analyzer integration (autocomplete, go-to-definition, doc hover on required types, etc). That is the only known method of adding docs directly to the struct. We had to pick one, and we chose Rust Analyzer integration. This is obviously a tradeoff, but we generally agree that this is the right call.

If you want to see requires, you can just press F12 on the type you are interested in and it will navigate you to the definition, which will have the `require`  list. From there (thanks to Rust Analyzer working), you can hover the name of each require to see its docs.

PR #18169, opened 2025-03-06
Document required components with a marker trait

bushrat011899:
# Objective

- Builds on #18165
- Alternative to #18171

## Solution

- Added a new trait, `Require<C: Component>: Component`, which is automatically implemented in the `Component` derive macro.

## Testing

- CI

---

## Notes

- This trait is purely for documentation purposes and has no impact on the _functionality_ of required components.
- I recommend using `cargo doc --open` with this PR and inspecting the `Button` and `Node` components as examples for the effects of this PR.
- ~~I have based this PR off of #18165 and will rebase once it is merged.~~

## Examples

_`Button` required components are listed logically in *Trait Implementations*_
image

_The method used to insert a required component can also be documented_
image

_`Node` can document components it requires *and* components that require it_
image

### Cons

- Can only place information within the trait implementation section, whereas the extension can place it anywhere, including at the top of the page as-if a user had done so themselves.
- Information is more verbose, as it includes `where` bounds, etc.

### Pros

- This PR works for all ecosystem and user crates automatically, whereas the extension requires opt-in configuration during doc generation.
- List format is better suited for items with many required components.
- Can easily show additional information, such as the method used to insert a required component if it is not present.
- Automatically shows documentation on the required and requiring components (e.g., `Button` and `Node` both get documentation for `Button` requiring `Node`)

### DIFF
--- a/crates/bevy_ecs/macros/src/component.rs
+++ b/crates/bevy_ecs/macros/src/component.rs
@@ -220,7 +220,7 @@ pub fn derive_component(input: TokenStream) -> TokenStream {
                 quote! {
                     /// If not already present, the required component will be inserted using
                     #[doc = #insertion_info]
-                    unsafe impl #impl_generics #bevy_ecs_path::component::document_required_components::Require<#path> for #struct_name #type_generics #where_clause {}
+                    impl #impl_generics #bevy_ecs_path::component::document_required_components::Require<#path> for #struct_name #type_generics #where_clause {}
                 }
             });
 
--- a/crates/bevy_ecs/macros/src/component.rs
+++ b/crates/bevy_ecs/macros/src/component.rs
@@ -207,6 +207,28 @@ pub fn derive_component(input: TokenStream) -> TokenStream {
     let struct_name = &ast.ident;
     let (impl_generics, type_generics, where_clause) = &ast.generics.split_for_impl();
 
+    let required_component_impls = attrs.requires.as_ref().map(|r| {
+        let impls = r
+            .iter()
+            .map(|r| {
+                let path = &r.path;
+                let insertion_info = match &r.func {
+                    Some(RequireFunc::Closure(closure)) => format!("`{}`", closure.body.to_token_stream()),
+                    Some(RequireFunc::Path(path)) => format!("[`{}`].", path.to_token_stream()),
+                    None => "the [default](Default::default) value.".to_string(),
+                };
+                quote! {
+                    /// If not already present, the required component will be inserted using
+                    #[doc = #insertion_info]
+                    impl #impl_generics #bevy_ecs_path::component::Require<#path> for #struct_name #type_generics #where_clause {}
+                }
+            });
+
+        quote! {
+            #(#impls)*
+        }
+    });
+
     let required_component_docs = attrs.requires.map(|r| {
         let paths = r
             .iter()
@@ -270,6 +292,8 @@ pub fn derive_component(input: TokenStream) -> TokenStream {
         #relationship
 
         #relationship_target
+
+        #required_component_impls
     })
 }
 
--- a/crates/bevy_ecs/macros/src/component.rs
+++ b/crates/bevy_ecs/macros/src/component.rs
@@ -220,7 +220,7 @@ pub fn derive_component(input: TokenStream) -> TokenStream {
                 quote! {
                     /// If not already present, the required component will be inserted using
                     #[doc = #insertion_info]
-                    impl #impl_generics #bevy_ecs_path::component::Require<#path> for #struct_name #type_generics #where_clause {}
+                    unsafe impl #impl_generics #bevy_ecs_path::component::Require<#path> for #struct_name #type_generics #where_clause {}
                 }
             });
 
--- a/crates/bevy_ecs/macros/src/component.rs
+++ b/crates/bevy_ecs/macros/src/component.rs
@@ -220,7 +220,7 @@ pub fn derive_component(input: TokenStream) -> TokenStream {
                 quote! {
                     /// If not already present, the required component will be inserted using
                     #[doc = #insertion_info]
-                    unsafe impl #impl_generics #bevy_ecs_path::component::Require<#path> for #struct_name #type_generics #where_clause {}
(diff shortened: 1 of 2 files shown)


crates/bevy_ecs/macros/src/component.rs
```diff
@@ -202,6 +202,40 @@ pub fn derive_component(input: TokenStream) -> TokenStream {
     let struct_name = &ast.ident;
     let (impl_generics, type_generics, where_clause) = &ast.generics.split_for_impl();
 
+    let required_component_impls = attrs.requires.as_ref().map(|r| {
```
JMS55, line 205: Can you add [#[cfg(doc)]](https://doc.rust-lang.org/rustdoc/advanced-features.html#cfgdoc-documenting-platform-specific-or-feature-specific-information) to the generated impl/trait?

bushrat011899: Sadly not, I tried but `cargo doc` doesn't propagate `doc` to dependencies or macros in a way that worked reliably.

jnhyatt:
I'm not exactly a new user, but my knee-jerk reaction to seeing `impl Require<FocusPolicy> for Button` was "oh, I guess I can impl `Require<X>` for my component to require `X`". I'd be wary of newer users (especially new to Rust) having this misconception.

bushrat011899:
> I'm not exactly a new user, but my knee-jerk reaction to seeing `impl Require<FocusPolicy> for Button` was "oh, I guess I can impl `Require<X>` for my component to require `X`". I'd be wary of newer users (especially new to Rust) having this misconception.

Agreed. I'm going to mark the trait as `unsafe` and document that you _must_ sync implementations of this trait with the implementation of `Component` itself.

BD103 (approved):

alice-i-cecile (changes requested):
I'm on board with this plan: the backlinks are incredibly useful. That said, this trait needs to be way more explicit about a) it just being a marker for docs and b) the fact that users should effectively never ever implement this because it's automatically generated.

bushrat011899:
I've moved the trait into a `doc(hidden)` module named `document_required_components`, which ensures `Require` is never suggested by an IDE, and clearly communicates to users importing the trait that it's a documentation marker. I've also amended the documentation for `Require` to make it clear that it's up to the implementer to ensure `Component` _actually_ requires the components as marked by `Require`.

crates/bevy_ecs/src/component.rs
```diff
@@ -455,6 +455,23 @@ pub trait Component: Send + Sync + 'static {
...
+    use super::Component;
+
+    /// Indicates this [`Component`] requires another [`Component`] `C`.
+    ///
+    /// **This trait does not register required components on its own**
+    ///
+    /// This trait is similar to [`Eq`] in the sense that it is up to the implementer to ensure `C` is
+    /// appropriately registered as a required component.
+    ///
+    /// It is the implementor's responsibility to ensure the implementation of [`Component::register_required_components`]
```
alice-i-cecile, line 470:
```suggestion
    /// You should never manually implement this trait, but it is the implementor's responsibility to ensure the implementation of [`Component::register_required_components`]
```

alice-i-cecile (approved):

crates/bevy_ecs/macros/src/component.rs
```diff
@@ -207,6 +207,28 @@ pub fn derive_component(input: TokenStream) -> TokenStream {
...
+                let path = &r.path;
+                let insertion_info = match &r.func {
+                    Some(RequireFunc::Closure(closure)) => format!("`{}`", closure.body.to_token_stream()),
+                    Some(RequireFunc::Path(path)) => format!("[`{}`].", path.to_token_stream()),
+                    None => "the [default](Default::default) value.".to_string(),
+                };
+                quote! {
+                    /// If not already present, the required component will be inserted using
+                    #[doc = #insertion_info]
+                    impl #impl_generics #bevy_ecs_path::component::document_required_components::Require<#path> for #struct_name #type_generics #where_clause {}
```
cart, line 223: I'm not convinced that this is meaningfully "better" than the current "required list in Component docs":

1. Because traits are listed alphabetically, users will hit the `Component` trait before they hit the `Require<X>` traits.
2. The require list in `Component` is significantly less "noisy". You can consume the entire list at a glance, rather than needing to parse out multiple lines of docs for each required component (and needing to look "around" generics / formatting / etc)
3. The Require trait impl _is_ confusing / introduces a weirdness factor (even if we manage to hide it). This is non-standard trait usage (traits are generally about "rust behaviors"). People reading the docs need to deal with the fact that 
4. This introduces additional codegen, which is something we should be wary of, given that we are already "heavy" on the codegen side.

From my perspective the only real win here is that the require list shows up in the sidebar, but that is _also_ a downside, as a long list takes up significant space and obscures other "real" trait impls. And it doesn't necessarily meaningfully improve the prominence of the require list. The sidebar is often littered with other things (ex: I have to scroll to see the list for Node).

cart: Additionally, our docs are just generally already very "noisy" and this adds even more to the pile.

crates/bevy_ecs/src/component.rs
```diff
@@ -455,6 +455,24 @@ pub trait Component: Send + Sync + 'static {
     fn visit_entities_mut(_this: &mut Self, _f: impl FnMut(&mut Entity)) {}
 }
 
+// doc(hidden) module makes it clear that usage of this trait outside of `bevy_ecs` is unsupported.
+#[doc(hidden)]
```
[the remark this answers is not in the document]
bushrat011899: Ah whoops, should be tested first! I assumed keeping the trait public would keep it in the documentation.

crates/bevy_ecs/macros/src/component.rs
bushrat011899: I think these are valid concerns, and I don't think this technique can be amended to satisfy them. Thanks for providing the feedback! I'll close this out

bushrat011899:
Closed based on feedback as not being the desired path forward. Thanks for the reviews and feedback! Good to explore and experiment even if it doesn't get merged.

Issue #1615, opened 2021-03-11
Do We Want to Use CFG Aliases?


zicklag:
Just saw this TODO:

https://github.com/bevyengine/bevy/blob/b17f8a4bce5551b418654fffb1fe97ff0f9852f0/crates/bevy_winit/src/lib.rs#L174-L176

If it's something we want to do we can use the [`cfg_aliases`](https://github.com/katharostech/cfg_aliases) crate to make convenient aliases for cfg checks. It's an extra dependency, but not really because it's already used in GFX and WGPU and it's just a macro_rules macro that doesn't have any dependencies itself.

I could open a PR if that helps.

james7132:
This could be pretty useful given the explosion in platform-specific cfg attributes (i.e. webgl and webgpu). Could you open a PR for this?

zicklag:
I've been crazy busy lately so I probably won't have time to open a PR. 

It's super easy to setup, though, and [here](https://github.com/gfx-rs/wgpu/blob/8e15707631f1db5612824dea122ea8b6a4e9b271/wgpu/build.rs#L1-L18)'s an example of how it's used on `wgpu`, if anybody else wants to take a hack at it, it'd be a good beginner issue.

PR #18170, opened 2025-03-06
Create cfg aliases for `bevy_app`

hukasu:
# Objective

Implements partially feature requested in #1615 

## Solution

Create cfg aliases for some of the cfg used in `bevy_app`

## Testing

Ran `cargo run -p ci` but needs more testing

## Showcase

Before
```rust
#[cfg(all(any(unix, windows), feature = "std"))]
pub use terminal_ctrl_c_handler;
// ...
#[cfg(all(not(target_arch = "wasm32"), feature = "bevy_tasks"))]
bevy_tasks::tick_global_task_pools_on_main_thread();
```

After
```rust
#[cfg(std_windows_or_unix)]
pub use terminal_ctrl_c_handler::*;
// ...
#[cfg(can_run_tasks)]
bevy_tasks::tick_global_task_pools_on_main_thread();
```

### DIFF
--- a/crates/bevy_app/build.rs
+++ b/crates/bevy_app/build.rs
@@ -1,0 +1,44 @@
+//! Build script containing cfg aliases
+
+fn main() {
+    // Collecting information that will be used to decide which cfg aliases
+    // will be exported.
+    // Using `cfg!(target_arch = "wasm32")` and such does not work because
+    // `build.rs` uses the host archtecture and other information, not the target's.
+    // Cargo then defines environment variables for the information of the target
+    let windows = std::env::var("CARGO_CFG_WINDOWS").is_ok();
+    let unix = std::env::var("CARGO_CFG_UNIX").is_ok();
+
+    let is_wasm32 = target_archtecture("wasm32");
+
+    let feature_bevy_tasks = feature_present("bevy_tasks");
+    let feature_std = feature_present("std");
+
+    // Prevent warnings on uses of `#[cfg]`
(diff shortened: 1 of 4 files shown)


crates/bevy_app/build.rs
```diff
@@ -0,0 +1,44 @@
+//! Build script containing cfg aliases
```
bushrat011899, line 1: The inclusion of a `build.rs` is definitely going to be the controversial component of this PR. Between security and build time concerns, I see this single-handedly being a show-stopper.

Personally, I'm not against it, since we have many dependencies that use [`cfg_aliases`](https://docs.rs/cfg_aliases/latest/cfg_aliases/) already (which is effectively what you're doing here by hand). But I think this will be very hard to land.

hukasu: oh, `cfg_aliases` is already being used? i wanted to decrease the contentiousness by a little by not adding a new dependency, but if this is already present, maybe there won't be pushback on that point

bushrat011899: It's already brought in as a `build-dependency` by `winit` and `wgpu`, so I wouldn't say it's a controversial inclusion in of itself. It's purely the use of a `build.rs` that I see as being controversial.

alice-i-cecile:
build.rs is a complete non-starter for me. That way lies madness!

NthTensor:
Closing, the alternative (https://github.com/bevyengine/bevy/pull/18822) has been blessed by cart and seems on track to merge.

PR #18171, opened 2025-03-06
Rustdoc: Restore visibility of required components

SpecificProtagonist:
# Objective

Follow up on #18165 to make the list of a component's required components easily accessible again.

Alternative to #18169

This PR is stacked on top of #18165 and #17857, so review by looking at the changes in the last commit here if those aren't merged yet.

## Solution

When using the rustdoc extensions (which also add trait tags), add the list of required components, if there are any, as a new section directly after the main doc comment. Fall back to the current behavior if there's an incompatible change to rustdoc's html format.

This could be extended to also show what components require the documented component.

## Testing

Run
```sh
cargo build --package rustdoc-wrapper --release
cargo doc --config "build.rustdoc = \"target/release/rustdoc-wrapper\"" --workspace
```

---

## Showcase

bevy required components


alice-i-cecile:
I prefer #18169 because I don't particularly like #17857, and because I *really* like the backlinks there. Nominating to close, but please feel free to argue.

mockersf:
Still against #17857 and this PR on my side. We want to reduce custom work by upstreaming what we can to rustdoc, not pile on more things that would make it so custom it would be of no interest to the Rust community at large.

As far as I know they are still waiting on a proposal from us describing what we want from tags

NthTensor:
I'm a bit disappointed because the current doc format for required components is absolutely not discoverable and presents a real problem to new users. But I understand the maintainability concern here. Seems like we need prioritize working with the docs.rs team on this.

Issue #18155, opened 2025-03-04
Tracking: Atomically Staged Components


ElliottjPierce:
## What problem does this solve or what need does it fill?

Lots of planned and wishlist features require (or would benefit from) mutably sharing components between threads. These notably include assets as entities (#11266), multi-world bevy, readonly queries and other readonly system parameters, pre-registering components, and scripting.

## Roadmap

- [x] Prototype. See #17569.
- [x] Decouple registration from storage. See #17671.
- [ ] Fix required components merging. See #18028.
- [ ] Implement generalized atomic staging. See #17701.
- [ ] Implement atomic staging for components. See #17871.
- [ ] Use staged components in `World`.
- [ ] Change signatures of internal use of `Components` to use staged components.
- [ ] Allow readonly queries and system parameters to be initialized with only `&World`.
- [ ] Allow creating and running read-only systems with only `&World`.
- [ ] Remove `try_query` API (no longer needed).
- [ ] Expose components to users via a system parameter. Maybe.
- [ ] After final interfaces are decided, do a final pass for code quality.
- [ ] Related: Make `ComponentInfo` fully immutable in an arc or something. This mainly benefits code quality.

ElliottjPierce:
Noting for the record [this](https://discord.com/channels/691052431525675048/692572690833473578/1346499196655505534) conversation on discord. It discusses alternative design, that, at the time of writing is better IMO.

alice-i-cecile:
I think we should close this and open a tracking issue for that design :)

ElliottjPierce:
> I think we should close this and open a tracking issue for that design :)

Absolutely.

ElliottjPierce:
See: #18184

PR #18173, opened 2025-03-06
Queued component registration

ElliottjPierce:
# Objective

This is an alternative to #17871 and #17701 for tracking issue #18155. This thanks to @maniwani for help with this design.

The goal is to enable component ids to be reserved from multiple threads concurrently and with only `&World`. This contributes to assets as entities, read-only query and system parameter initialization, etc.

## What's wrong with #17871 ?

In #17871, I used my proposed staging utilities to allow *fully* registering components from any thread concurrently with only `&Components`. However, if we want to pursue components as entities (which is desirable for a great many reasons. See [here](https://discord.com/channels/691052431525675048/692572690833473578/1346499196655505534) on discord), this staging isn't going to work. After all, if registering a component requires spawning an entity, and spawning an entity requires `&mut World`, it is impossible to register a component fully with only `&World`.

## Solution

But what if we don't have to register it all the way? What if it's enough to just know the `ComponentId` it will have once it is registered and to queue it to be registered at a later time? Spoiler alert: That is all we need for these features.

Here's the basic design:

Queue a registration:

1. Check if it has already been registered.
2. Check if it has already been queued.
3. Reserve a `ComponentId`.
4. Queue the registration at that id.

Direct (normal) registration:

1. Check if this registration has been queued.
2. If it has, use the queued registration instead.
3. Otherwise, proceed like normal.

Appllying the queue:

1. Pop queued items off one by one.
2. Register them directly.

One other change:

The whole point of this design over #17871 is to facilitate coupling component registration with the World. To ensure that this would fully work with that, I went ahead and moved the `ComponentId` generator onto the world itself. That stemmed a couple of minor organizational changes (see migration guide). As we do components as entities, we will replace this generator with `Entities`, which lives on `World` too. Doing this move early let me verify the design and will reduce migration headaches in the future. If components as entities is as close as I think it is, I don't think splitting this up into different PRs is worth it. If it is not as close as it is, it might make sense to still do #17871 in the meantime (see the risks section). I'll leave it up to y'all what we end up doing though.

## Risks and Testing

The biggest downside of this compared to #17871 is that now we have to deal with correct but invalid `ComponentId`s. They are invalid because the component still isn't registered, but they are correct because, once registered, the component will have exactly that id.

However, the only time this becomes a problem is if some code violates safety rules by queuing a registration and using the returned id as if it was valid. As this is a new feature though, nothing in Bevy does this, so no new tests were added for it. When we do use it, I left detailed docs to help mitigate issues here, and we can test those usages. Ex: we will want some tests on using queries initialized from queued registrations.

## Migration Guide

Component registration can now be queued with only `&World`. To facilitate this, a few APIs needed to be moved around.

The following functions have moved from `Components` to `ComponentsRegistrator`:

- `register_component`
- `register_component_with_descriptor`
- `register_resource_with_descriptor`
- `register_non_send`
- `register_resource`
- `register_required_components_manual`

Accordingly, functions in `Bundle` and `Component` now take `ComponentsRegistrator` instead of `Components`.
You can obtain `ComponentsRegistrator` from the new `World::components_registrator`.
You can obtain `ComponentsQueuedRegistrator` from the new `World::components_queue`, and use it to stage component registration if desired.

# Open Question

Can we verify that it is enough to queue registration with `&World`? I don't think it would be too difficult to package this up into a `Arc<MyComponentsManager>` type thing if we need to, but keeping this on `&World` certainly simplifies things. If we do need the `Arc`, we'll need to look into partitioning `Entities` for components as entities, so we can keep most of the allocation fast on `World` and only keep a smaller partition in the `Arc`. I'd love an SME on assets as entities to shed some light on this.

### DIFF
--- a/crates/bevy_ecs/src/component.rs
+++ b/crates/bevy_ecs/src/component.rs
@@ -1539,6 +1539,7 @@ impl<'w> ComponentsRegistrator<'w> {
         // components
         while let Some(registrator) = {
             let queued = self
+                .components
                 .queued
                 .get_mut()
                 .unwrap_or_else(PoisonError::into_inner);
@@ -1553,6 +1554,7 @@ impl<'w> ComponentsRegistrator<'w> {
         // resources
         while let Some(registrator) = {
             let queued = self
+                .components
                 .queued
                 .get_mut()
                 .unwrap_or_else(PoisonError::into_inner);
@@ -1566,6 +1568,7 @@ impl<'w> ComponentsRegistrator<'w> {
 
         // dynamic
         let queued = &mut self
+            .components
             .queued
             .get_mut()
             .unwrap_or_else(PoisonError::into_inner);
@@ -1601,6 +1604,7 @@ impl<'w> ComponentsRegistrator<'w> {
         }
 
         if let Some(registrator) = self
+            .components
             .queued
             .get_mut()
             .unwrap_or_else(PoisonError::into_inner)
@@ -1777,6 +1781,7 @@ impl<'w> ComponentsRegistrator<'w> {
         }
 
         if let Some(registrator) = self
+            .components
             .queued
             .get_mut()
             .unwrap_or_else(PoisonError::into_inner)
--- a/crates/bevy_ecs/src/component.rs
+++ b/crates/bevy_ecs/src/component.rs
@@ -15,8 +15,11 @@ use crate::{
 use alloc::boxed::Box;
 use alloc::{borrow::Cow, format, vec::Vec};
 pub use bevy_ecs_macros::Component;
-use bevy_platform_support::collections::{HashMap, HashSet};
 use bevy_platform_support::sync::Arc;
+use bevy_platform_support::{
+    collections::{HashMap, HashSet},
+    sync::PoisonError,
+};
 use bevy_ptr::{OwningPtr, UnsafeCellDeref};
 #[cfg(feature = "bevy_reflect")]
 use bevy_reflect::Reflect;
@@ -1141,7 +1144,7 @@ impl ComponentCloneBehavior {
 
 /// Coordinates a registration that is reserved, but not registered.
 trait QueuedComponentRegistration {
-    fn register(&mut self, registrator: ComponentsRegistrator);
+    fn register(&mut self, registrator: &mut ComponentsRegistrator, this_id: ComponentId);
 }
 
 /// Allows queueing components to be registered.
@@ -1284,7 +1287,7 @@ impl<'w> ComponentsQueuedRegistrator<'w> {
     }
 }
 
-/// A type that enables registering in [`Components`].
+/// A [`Components`] wrapper that enables additional features, like registration.
 pub struct ComponentsRegistrator<'w> {
     components: &'w mut Components,
     ids: &'w mut ComponentIds,
@@ -1325,33 +1328,46 @@ impl<'w> ComponentsRegistrator<'w> {
     /// * [`Components::register_component_with_descriptor()`]
     #[inline]
     pub fn register_component<T: Component>(&mut self) -> ComponentId {
-        self.component_id::<T>().unwrap_or_else(|| {
-            let id = self.ids.next_mut();
-            // SAFETY: The component is not currently registered, and the id is fresh.
-            unsafe { self.register_component_internal::<T>(&mut Vec::new(), id) }
-            id
-        })
+        self.register_component_checked::<T>(&mut Vec::new())
     }
 
-    /// Same as [`register_component`] but keeps a `recursion_check_stack`.
+    /// Same as [`Self::register_component_internal`] but keeps a checks for safety.
     #[inline]
-    fn register_component_checked_internal<T: Component>(
+    fn register_component_checked<T: Component>(
         &mut self,
         recursion_check_stack: &mut Vec<ComponentId>,
     ) -> ComponentId {
-        self.component_id::<T>().unwrap_or_else(|| {
-            let id = self.ids.next_mut();
-            // SAFETY: The component is not currently registered, and the id is fresh.
-            unsafe { self.register_component_internal::<T>(recursion_check_stack, id) }
-            id
-        })
+        let type_id = TypeId::of::<T>();
+        if let Some(id) = self.indices.get(&type_id) {
+            return *id;
+        }
+
+        if let Some((id, mut registrator)) = self
+            .queued
+            .get_mut()
+            .unwrap_or_else(PoisonError::into_inner)
+            .components
+            .remove(&type_id)
+        {
+            // If we are trying to register something that has already been queued, we respect the queue.
+            // Just like if we are trying to register something that already is, we respect the first registration.
+            registrator.register(self, id);
+            return id;
+        }
+
+        let id = self.ids.next_mut();
+        // SAFETY: The component is not currently registered, and the id is fresh.
+        unsafe {
+            self.register_component_unchecked::<T>(recursion_check_stack, id);
+        }
+        id
     }
 
     /// # Safety
     ///
-    /// Neither this component, nor it's id may be registered. This must be a new registration.
+    /// Neither this component, nor it's id may be registered or queued. This must be a new registration.
     #[inline]
-    unsafe fn register_component_internal<T: Component>(
+    unsafe fn register_component_unchecked<T: Component>(
         &mut self,
         recursion_check_stack: &mut Vec<ComponentId>,
         id: ComponentId,
@@ -1442,8 +1458,8 @@ impl<'w> ComponentsRegistrator<'w> {
         inheritance_depth: u16,
         recursion_check_stack: &mut Vec<ComponentId>,
     ) {
-        let requiree = self.register_component_checked_internal::<T>(recursion_check_stack);
-        let required = self.register_component_checked_internal::<R>(recursion_check_stack);
+        let requiree = self.register_component_checked::<T>(recursion_check_stack);
+        let required = self.register_component_checked::<R>(recursion_check_stack);
 
         // SAFETY: We just created the components.
         unsafe {
@@ -1467,16 +1483,12 @@ impl<'w> ComponentsRegistrator<'w> {
     /// * [`Components::register_resource_with_descriptor()`]
     #[inline]
     pub fn register_resource<T: Resource>(&mut self) -> ComponentId {
-        self.resource_id::<T>().unwrap_or_else(|| {
-            let id = self.ids.next_mut();
-            // SAFETY: The resource is not currently registered, the id is fresh, and the [`ComponentDescriptor`] matches the [`TypeId`]
-            unsafe {
-                self.register_resource_with(TypeId::of::<T>(), id, || {
-                    ComponentDescriptor::new_resource::<T>()
-                });
-            }
-            id
-        })
+        // SAFETY: The [`ComponentDescriptor`] matches the [`TypeId`]
+        unsafe {
+            self.register_resource_with(TypeId::of::<T>(), || {
+                ComponentDescriptor::new_resource::<T>()
+            })
+        }
     }
 
     /// Registers a [non-send resource](crate::system::NonSend) of type `T` with this instance.
@@ -1484,20 +1496,48 @@ impl<'w> ComponentsRegistrator<'w> {
     /// the ID of the pre-existing resource.
     #[inline]
     pub fn register_non_send<T: Any>(&mut self) -> ComponentId {
-        let type_id = TypeId::of::<T>();
-        self.resource_indices
-            .get(&type_id)
-            .copied()
-            .unwrap_or_else(|| {
-                let id = self.ids.next_mut();
-                // SAFETY: The resource is not currently registered, the id is fresh, and the [`ComponentDescriptor`] matches the [`TypeId`]
-                unsafe {
-                    self.register_resource_with(type_id, id, || {
-                        ComponentDescriptor::new_non_send::<T>(StorageType::default())
-                    });
-                }
-                id
+        // SAFETY: The [`ComponentDescriptor`] matches the [`TypeId`]
+        unsafe {
+            self.register_resource_with(TypeId::of::<T>(), || {
+                ComponentDescriptor::new_non_send::<T>(StorageType::default())
             })
+        }
+    }
+
+    /// Same as [`Components::register_resource_unchecked_with`] but handles safety.
+    ///
+    /// # Safety
+    ///
+    /// The [`ComponentDescriptor`] must match the [`TypeId`].
+    #[inline]
+    unsafe fn register_resource_with(
+        &mut self,
+        type_id: TypeId,
+        descriptor: impl FnOnce() -> ComponentDescriptor,
+    ) -> ComponentId {
+        if let Some(id) = self.resource_indices.get(&type_id) {
+            return *id;
+        }
+
+        if let Some((id, mut registrator)) = self
+            .queued
+            .get_mut()
+            .unwrap_or_else(PoisonError::into_inner)
+            .resources
+            .remove(&type_id)
+        {
+            // If we are trying to register something that has already been queued, we respect the queue.
+            // Just like if we are trying to register something that already is, we respect the first registration.
+            registrator.register(self, id);
+            return id;
+        }
+
+        let id = self.ids.next_mut();
+        // SAFETY: The resource is not currently registered, the id is fresh, and the [`ComponentDescriptor`] matches the [`TypeId`]
+        unsafe {
+            self.register_resource_unchecked_with(type_id, id, descriptor);
+        }
+        id
     }
 
     /// Registers a [`Resource`] described by `descriptor`.
@@ -1531,6 +1571,7 @@ pub struct Components {
     components: HashMap<ComponentId, ComponentInfo>,
     indices: TypeIdMap<ComponentId>,
     resource_indices: TypeIdMap<ComponentId>,
+    // This is kept internal and local to verify that no deadlocks can occor.
     queued: bevy_platform_support::sync::RwLock<QueuedComponents>,
 }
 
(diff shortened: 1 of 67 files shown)


maniwani:
I've looked over this and it seems good. atm I can't think of anything to add on top of what we discussed on Discord.

The general strategy of "claim the ID now and its registration will be completed before we actually try to use it" isn't too different from how systems can specify commands to `insert` components on entities that don't exist yet, so I think it'll carry over to assets as well.

Thanks for the really thorough yet concise description and comments!

alice-i-cecile:
I'm in favor of this direction. Marking as Blessed, but I don't have the energy right this moment for a review. Pester me if I don't get to it within a week :)

crates/bevy_ecs/src/world/mod.rs
```diff
@@ -2177,12 +2196,14 @@ impl World {
         B: Bundle<Effect: NoBundleEffect>,
     {
         self.flush();
-
         let change_tick = self.change_tick();
 
+        // SAFETY: These come from the same world.
+        let mut registrator =
+            unsafe { ComponentsRegistrator::new(&mut self.components, &mut self.component_ids) };
```
andriyDev, line 2203: Not using `.component_registrator()` was confusing here. Can we add a comment to all of these cases (there are several) to explain "We can't use `.component_registrator()` since we want to mutably borrow other fields in the world".

crates/bevy_ecs/src/world/mod.rs
```diff
@@ -2814,12 +2841,22 @@ impl World {
         }
     }
 
+    /// Applies any queued component registration.
+    /// For spawning vanilla rust component types and resources, this is not strictly necessary.
+    /// However, flushing components can make information available more quickly,and can have performance benefits.
```
andriyDev, line 2846: Very nit: ",and" -> ", and"

crates/bevy_ecs/src/component.rs
```diff
@@ -1135,73 +1138,538 @@ impl ComponentCloneBehavior {
...
 }
 
-/// Stores metadata associated with each kind of [`Component`] in a given [`World`].
+/// Coordinates a registration that is reserved, but not registered.
+trait QueuedComponentRegistration {
+    /// Performs the registration.
+    ///
+    /// # Safery
+    ///
+    /// This must only ever be called once with a unique id.
```
andriyDev, line 1147: This wording was a little unintuitive to me - I initially thought it meant you can call register as long as each call provides a unique id. How about "This must only ever be called once, and must be called with a unique id." Splitting the line up makes it clear these are two separate requirements.

crates/bevy_ecs/src/component.rs
```diff
@@ -1135,73 +1138,538 @@ impl ComponentCloneBehavior {
...
+            queued.components.keys().next().copied().map(|type_id| {
+                // SAFETY: the id just came from a valid iterator.
+                unsafe { queued.components.remove(&type_id).debug_checked_unwrap() }
+            })
+        } {
+            // SAFETY: we own this value and it is being dropped, and the id came from the unique queue.
+            unsafe {
+                registrator.register(self, id);
+            }
+        }
```
andriyDev, line 1546: Wow this is a very complicated drain implementation :P Is the intent here to limit how long the lock is taken for? I wonder if this is worth it, since we're now locking multiple times rather than just locking once, registering all the components, and moving on.

In other words, should we just use `.drain()` instead? As long as we explicitly drop the lock after this loop, we should be good to go (at the minor cost of limiting how quickly components can be registered).

crates/bevy_ecs/src/component.rs
```diff
@@ -1135,73 +1138,538 @@ impl ComponentCloneBehavior {
     }
 }
 
-/// Stores metadata associated with each kind of [`Component`] in a given [`World`].
+/// Coordinates a registration that is reserved, but not registered.
+trait QueuedComponentRegistration {
```
andriyDev, line 1142: Why does this trait exist? It seems like its only purpose is to call `register` once. Could we not just replace it with `Box<dyn FnOnce(&mut ComponentsRegistrator, ComponentId)>`?

crates/bevy_ecs/src/component.rs
```diff
@@ -1135,73 +1138,538 @@ impl ComponentCloneBehavior {
...
+    /// enabling retrieving [`ComponentInfo`], etc.
+    pub fn apply_queued_registrations(&mut self) {
+        if !self.any_queued_mut() {
+            return;
+        }
+
+        // components
+        while let Some((id, mut registrator)) = {
+            let queued = self
+                .queued
```
andriyDev, line 1534: I got so dizzy looking for the `queued` field :P. Can we just do `self.components.queued`?

crates/bevy_ecs/src/component.rs
```diff
@@ -1135,73 +1138,538 @@ impl ComponentCloneBehavior {
...
+        // SAFETY: The component is not currently registered, and the id is fresh.
+        unsafe {
+            self.register_component_unchecked::<T>(recursion_check_stack, id);
         }
         id
     }
 
+    /// # Safety
+    ///
+    /// Neither this component, nor it's id may be registered or queued. This must be a new registration.
```
andriyDev, line 1631: "it's" -> "its"

andriyDev (changes requested):

andriyDev:
For my context: for components-as-entities, the idea is that you would reserve an Entity (the same way commands does today), but that entity would not be initialized until we apply the queue here?

crates/bevy_ecs/src/component.rs
ElliottjPierce: Good thoughts. I should add a comment to make this more clear.

We need to go one at a time for non dynamic registration. The dynamic registration is completely anonymous, so we can just drain it. (We actually need to use mem::take to preserve lifetimes. This could be side-stepped by moving around who owns `QueuedComponents`, but I don't think it's worth it.)

Anyway, for non-dynamic registration, it has a specific label attached to it, currently the `TypeId`. This can be a problem for recursive registration. Because once we do the drain (or in this case, the mem::take), that leaves `Components` with missing information.

For example, say `ComponentA` and `ComponentB` are queued, and `ComponentA` requires `ComponentB`. If we drain directly, and `ComponentA` was the first to be registered, then, when `ComponentA` registers `ComponentB` in `Component::register_required_components`, `Components` will not know that `ComponentB` was queued (since it will have been drained from the queue.) If that happened, `Components` would assign a new `ComponentId` to `ComponentB` which would be *different* than the id it was assigned in the queue. Then, when the drain iterator gets to `ComponentB`, it would be unsafely registering `ComponentB`, which is already registered.

That could produce some crazy undefined behavior. So, we have to pop the queue exactly one at a time, otherwise `Components` is left with incomplete or missing information when registering the queue.

crates/bevy_ecs/src/component.rs
ElliottjPierce: This is a crazy drain. Here's the root of the issue. `HashMap` doesn't provide a `pop` method because it would not be particularly fast, and the rust team didn't want to include that performance foot gun. But this is very much not a hot path, and it's not that bad anyway. See my other comment for why we have to `pop` instead of `drain`.

If this were a hot path, I would probably use `Vec` and enforce `TypeId` uniqueness manually. This would be faster because it has a `pop`, *and* it has better cache locality, so lookups would be faster (assuming the standard case of having a small or empty queue).

For now, we have to iterate the keys, grab only one key, and remove the corresponding value. That requires using `queued` twice, mutably. I put this in a local variable so we only need to this once: 

```rust
self
    .queued
    .get_mut()
    .unwrap_or_else(PoisonError::into_inner);
```

I think this is shorter and cleaner, and it saves an `unwrap_or_else` call. I would do this at the start of the function, but when I asked the borrow checker about it, we had a fight, and I lost.

ElliottjPierce:
> For my context: for components-as-entities, the idea is that you would reserve an Entity (the same way commands does today), but that entity would not be initialized until we apply the queue here?

AFAIK, that is correct. Of course, if we have `&mut World` anyway, we can just spawn and initialize the entity directly, but we'll cross that bridge when we get there.

The goal of this PR isn't necessarily to advance components as entities (though it technically does). The goal is be able to know the id of a component/resource before it is registered *and* to facilitate that without blocking `components as entities`.

crates/bevy_ecs/src/component.rs
andriyDev: I think you misunderstood. For this comment, I specifically meant we access `self.queued`. However ComponentsRegistrator does not have such a field. This is implicitly using the deref implementation to access the `queued` field on the `components` field. I just wanted to simplify that.

ElliottjPierce: OOOHHHH. Yes. Wow. I didn't even realize I'd done that lol. Great idea! It's fixed now.

crates/bevy_ecs/src/component.rs
andriyDev: Ahhhh this was a very tricky issue. Thank you for explaining and the accompanying comment!

crates/bevy_ecs/src/component.rs
```diff
@@ -1135,73 +1138,536 @@ impl ComponentCloneBehavior {
...
+            // SAFETY: Id uniqueness handled by caller.
+            unsafe {
+                registrator.register_component_inner(id, descriptor);
+            }
+        })
+    }
+}
+
+/// A [`Components`] wrapper that enables additional features, like registration.
+pub struct ComponentsRegistrator<'w> {
```
chescock, line 1476: Is this only necessary in order to borrow `components` and `component_ids` simultaneously?  Would it be possible to put `component_ids` inside `Components` instead of `World`, and then just use `&mut Components` and `&Components` again?  I think that would remove most of the changes outside of `World` and `Components`.  

Going a step farther, it looks like we only ever call `ids.next()` immediately before `queued.write()`.  Should we just make `next` be an ordinary `usize` inside `QueuedComponents`?

crates/bevy_ecs/src/component.rs
```diff
@@ -1135,73 +1138,536 @@ impl ComponentCloneBehavior {
     }
 }
 
-/// Stores metadata associated with each kind of [`Component`] in a given [`World`].
+/// A queued component registration.
+struct QueuedRegistration {
+    registrator: Box<dyn FnOnce(&mut ComponentsRegistrator, ComponentId)>,
```
chescock, line 1143: I think it might be possible to avoid allocating a `Box` here.  

The component and resource versions could pass plain `fn` pointers, since they don't need to capture anything.  (The resource one seems to capture `type_id::of::<T>()`, but you could evaluate that again in the function.)  

The dynamic component one has a `ComponentDescriptor` value, but always runs the same code.  So I think that could be a `Vec<(ComponentId, ComponentDescriptor)>` and not even have to do virtual dispatch.

crates/bevy_ecs/src/component.rs
```diff
@@ -1135,73 +1138,536 @@ impl ComponentCloneBehavior {
...
+    /// This ensures that every valid [`ComponentId`] is registered,
+    /// enabling retrieving [`ComponentInfo`], etc.
+    pub fn apply_queued_registrations(&mut self) {
+        if !self.any_queued_mut() {
+            return;
+        }
+
+        // Note:
+        //
+        // This is not just draining the queue. We need to empty the queue without removing the information from `Components`.
```
chescock, line 1524: This might be more clear if you stored the type-to-id mapping separately from the queue of things to register.  Something like `component_ids: TypeIdMap<ComponentId>` and `queued_components: Vec<QueuedRegistration>`.  For that matter, I think you could share a single `Vec` for resources, components, and dynamic components (or maybe dynamic components are separate if you store the `ComponentDescriptor`s directly.)  

Then you could just drain the `Vec`, since the ID mapping will still be available.  And you could either remove the IDs from the mapping as the components get registered, or clear it all at once at the end of `apply_queued_registrations`.

crates/bevy_ecs/src/component.rs
```diff
@@ -1539,14 +2161,21 @@ impl Components {
...
-            .get_mut(id.0)
+            .get_mut(&id)
             .map(|info| &mut info.required_by)
     }
 
     /// Type-erased equivalent of [`Components::component_id()`].
     #[inline]
     pub fn get_id(&self, type_id: TypeId) -> Option<ComponentId> {
-        self.indices.get(&type_id).copied()
+        self.indices.get(&type_id).copied().or_else(|| {
```
chescock, line 2171: This method can now return "correct but invalid" IDs, which the calling code may not be prepared for.  I'm not sure which behavior should be the default, but we may want to provide a non-blocking version that only ever returns fully-registered components, and we should note the behavior in the documentation.

chescock (commented):
Oh, yay, this looks much simpler!  

I left a few initial thoughts, but I've only skimmed the code so I may just be misunderstanding things that should be obvious.  I can try to do a more complete review tomorrow if it doesn't get merged first.

crates/bevy_ecs/src/component.rs
ElliottjPierce: Excellent point. I'm happy to do either behavior here as the default, but it's less work to keep it the same. If anyone expresses strong concern here, I'll change the default behavior. I'll add that alternative function too in a bit.

crates/bevy_ecs/src/component.rs
ElliottjPierce: All great points, and I'm happy to do that if that's what is decided. I mentioned this a little in my description, but the reason I did this is that it prepares us better for components as entities. For that, we will need access to `&Entities` from the world during registration. By keeping the generator on the world, it will save some headache latter, and it helps enforce that that coupling continues to be available until we do components as entities. (Which I feel is getting pretty close).

crates/bevy_ecs/src/component.rs
alice-i-cecile: Yeah, we can wait on this.

crates/bevy_ecs/src/component.rs
alice-i-cecile: > ? Would it be possible to put component_ids inside Components instead of World, and then just use &mut Components and &Components again

I'd prefer this!

> Going a step farther, it looks like we only ever call ids.next() immediately before queued.write(). Should we just make next be an ordinary usize inside QueuedComponents?

Yes please to this too :) Anything we can do to simplify these internals and guide authors into the correct patterns is appreciated here.

crates/bevy_ecs/src/component.rs
alice-i-cecile: Let's keep it the same for now.

crates/bevy_ecs/src/component.rs
ElliottjPierce: This would probably be safe. I think. I did consider that originally, but my concern was that it makes a lot of assumptions about how `Components` will work. That would probably be fine for today's implementation, but for the sake of future changes, I think keeping `Components` as the full source of truth is important. Hopefully that makes sense. If anyone feels strongly otherwise, I can try it out. Otherwise, I think this is another think to reconsider after the dust settles.

crates/bevy_ecs/src/component.rs
ElliottjPierce: @alice-i-cecile 100% not opposed to this, but I'm not sure my reply was visible when you wrote this. Is this still the direction you'd like to take? (I promise I'm not trying to be difficult. Just down want to have to undo the undo.)

alice-i-cecile: Hmm. Okay fine, we can leave it as is for now :)

chescock: Sounds good!  

(Although I wonder whether we'll want to put the `Entity` allocator behind an `Arc` anyway, so that we don't have to manage references to multiple structs like this for each thing that we everything-as-entities.)

crates/bevy_ecs/src/component.rs
chescock: > If you feel strongly about it

Nope, this is just Rust tempting me to micro-optimize unimportant things :).

crates/bevy_ecs/src/component.rs
chescock: > I think keeping `Components` as the full source of truth is important.

Sure, but `Components` *already* needs to account for the `TypeIdMap`s in `QueuedComponents`.  That's what this comment is explaining, right?  

Hmm, if I think about it in terms of wanting to prevent `Components` from needing to access `QueuedComponents`, then maybe what you want is an API on components to push in the `TypeIdMap<ComponentId>` as reserved ids.  If you reserve the IDs with `Components` first, then you could iterate over the queued registrations at your leisure.  

But what you have seems fine to start with!  I imagine it will have to be reworked for components-as-entities anyway.  I do hope we eventually get to something that's clear enough not to require quite so many lines of comments, though :).

crates/bevy_ecs/src/component.rs
ElliottjPierce: > (Although I wonder whether we'll want to put the Entity allocator behind an Arc anyway, so that we don't have to manage references to multiple structs like this for each thing that we everything-as-entities.)

If I understand components as entities correctly, we'll be putting the data of `ComponentsInfo` onto the entity itself. That lest us, for example, make requirements run on relations. If that understanding is correct, registration will need `&mut Storages` anyway. But, if, that's not the case, and we would put the entity allocator in an `Arc` anyway, then I agree.

> Hmm. Okay fine, we can leave it as is for now :)

Awesome! Do let me know if you change your mind. There's good reasons to go either way.

crates/bevy_ecs/src/component.rs
```diff
@@ -1212,46 +1678,245 @@ impl Components {
...
+        }
+        id
     }
+}
 
-    /// Returns the number of components registered with this instance.
+/// Stores metadata associated with each kind of [`Component`] in a given [`World`].
+#[derive(Debug, Default)]
+pub struct Components {
+    components: HashMap<ComponentId, ComponentInfo>,
```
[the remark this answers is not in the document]
ElliottjPierce: If I understand correctly, components as entities will put `ComponentInfo` on the entity itself in `Storages`. Se we won't have this at all. `Components` will just help look up `ComponentId`s.

But you're right. The `Vec` is much nicer. I don't see gaps happening that often, so I'll go ahead and convert to this. Great idea!

crates/bevy_ecs/src/component.rs
ElliottjPierce: > But what you have seems fine to start with! I imagine it will have to be reworked for components-as-entities anyway. I do hope we eventually get to something that's clear enough not to require quite so many lines of comments, though :).

100% Agree. Your comments here are valid, I just think compressing the safety to right here is less complicated than having to think about "What if this registration was queued in this particular way?" though out `Components`, even though that does make this particular function very non-trivial. But that's just my opinion.

IIUC, this part of Bevy is going to change a lot over 0.17. So I think we can revisit making this faster with `Vec` and `drain` later. It's certainly not a regression since queueing is new. If this is merged, I'll make an issue to track cleaning this up a bit once the dust settles.

ElliottjPierce:
I know y'all don't have the time to follow every detail here, so here's a status update:

Reviews:

There are a few ways code quality can be improved, but doing so would make assumptions for how this can evolve in the future. Consensus at the moment seems to be to revisit these after the dust settles. But keep the constructive criticism coming as you see issues! It only makes the code better.

@alice-i-cecile:

The issue is marked as waiting on author. Is there anything I'm missing that I still need to do? (I don't mean to bug you; I know you're busy, but if there's something I should be working on, I'd like to get to it!)

Working assumptions:

The current state of the PR is based on two assumptions. First, I am assuming that registration with components as entities will require access to `Storages`. If that is not the case, we can simplify even more, but my impression is that it is. Second, I am assuming that we only need to support queueing with `&World`; we don't need to provide something like `Arc<RemoteComponentQueueing>`. If that assumption is wrong, I can 100% still add that. But it might be best done in a follow up PR.

I'd love SME clarification on these assumptions before merging. Thanks!

crates/bevy_ecs/src/component.rs
maniwani: > If I understand components as entities correctly, we'll be putting the data of `ComponentsInfo` onto the entity itself. That lest us, for example, make requirements run on relations.

For safety reasons, there needs to be some component info that user code cannot touch. Like, we don't want people swapping the `drop` functions of components `A` and `B`.

So I was simply thinking to duplicate the information onto the components' entities for user code to consume, while still keeping a private copy for the world's internal operations.

ElliottjPierce: > For safety reasons, there needs to be some component info that user code cannot touch. Like, we don't want people swapping the `drop` functions of components `A` and `B`.

That would be scary. I would guess though, this would be better enforced with immutable components, if that's possible. Or we could keep the component private and expose a readonly query item that just forwards to `&the_component`. 

> So I was simply thinking to duplicate the information onto the components' entities for user code to consume, while still keeping a private copy for the world's internal operations.

Makes sense, but then why not #17871? Unless we want more data to be inserted on the component too.

maniwani: > Makes sense, but then why not https://github.com/bevyengine/bevy/pull/17871?

I don't get how that's related?

The change to get "components as entities" is to just surgically replace `ComponentId` with `Entity`. That shouldn't imply any big changes to how the world currently performs its internal operations (i.e. where it stores `drop` functions, etc.)

Each component will just so happen to have a representative entity that users can do stuff with. The world doesn't need to rely on anything about these entities besides them staying alive. (Maybe it can later, but not a priority.)

PR #18174, opened 2025-03-06
bevy_reflect: bump petgraph dependency

flokli:
# Objective

petgraph 0.7 has been released a while ago, but bevy still pulls in petgraph 0.6

## Solution

- Update the version in bevy_reflect's `Cargo.toml`.

## Testing

- I ran `cargo test --all-features`.

### DIFF
--- a/crates/bevy_animation/Cargo.toml
+++ b/crates/bevy_animation/Cargo.toml
@@ -31,7 +31,7 @@ bevy_platform_support = { path = "../bevy_platform_support", version = "0.16.0-d
 ] }
 
 # other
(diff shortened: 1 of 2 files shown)


bushrat011899:
`petgraph` is also used in `bevy_animation`, so that will need to be updated too.

flokli:
> `petgraph` is also used in `bevy_animation`, so that will need to be updated too.

Right! Updated.

fjkorf (approved):
looks good. pulled and ran some examples with the code successfully

bushrat011899 (approved):

mnmaita (approved):

PR #18175, opened 2025-03-06
Prevent an additional world update after all windows have closed on Windows

aloucks:
# Objective

The fix in #18105 includes a check for running headless, but this allows for an extra world update during shutdown.

This commit checks if the `AppExit` event has been recorded and prevents the additional world update.

### Before

```
2025-03-06T03:11:59.999679Z  INFO bevy_window::system: No windows are open, exiting    
2025-03-06T03:12:00.001942Z  INFO bevy_winit::system: Closing window 0v1
2025-03-06T03:12:00.012691Z  INFO bevy_window::system: No windows are open, exiting
```
### After

```
2025-03-06T03:18:45.552243Z  INFO bevy_window::system: No windows are open, exiting    
2025-03-06T03:18:45.554119Z  INFO bevy_winit::system: Closing window 0v1
```
## Testing

Ran `window` examples

- `monitor_info` continues to run after all windows are closed (it has `ExitCondition::DontExit`)
- `window_settings` invisible window creation works as expected
- `multiple_windows` exists after both windows are closed with a single exit message


greeble-dev (approved):
Tested on Win10/Nvidia. Tried the mentioned examples and a few others.

Side note: Is there a reason the `monitor_info` example is set to not close when all windows are closed? I assumed this was a bug at first. I tried removing the `ExitCondition::DontExit` and it worked fine, but I don't have multiple monitors.

aloucks:
> Side note: Is there a reason the `monitor_info` example is set to not close when all windows are closed? I assumed this was a bug at first. I tried removing the `ExitCondition::DontExit` and it worked fine, but I don't have multiple monitors.

That confused me on a previous PR. I also thought it was a bug at first. I think there needs to be an example that exercises that behavior, but I'm not sure that `monitor_info` is the right place. I didn't want to change the examples in this PR though.

PR #18178, opened 2025-03-06
Add `TextureAtlasSources::uv_rect` to get `Rect` in UV coords

copygirl:
# Objective

I'm building a bloxel game in which I (currently) use a texture atlas to render the blocks the world is made of. While I was coding it, I was using the `TextureAtlas...` types to build the terrain's texture atlas at runtime as shown in the [example](https://github.com/bevyengine/bevy/blob/latest/examples/2d/texture_atlas.rs). But when I was using it to build a 3D mesh out of the blocks, I found that there was no easy way get the texture rect in UV coordinates, only in pixels via `texture_rect()`. I had to resort to writing code like this:

```rs
let size = layout.size.as_vec2();
if let Some(rect) = sources.texture_rect(layout, texture) {
    let rect = rect.as_rect();
    let uvs = Rect::from_corners(rect.min / size, rect.max / size);
    // use the UVs here, such as to build vertex buffer
}
```

That is, until I wrote a helper function that's practically identical to the one in this PR.

## Solution

Add a `uv_rect` function to `TextureAtlasSources` that will return a `Rect` with coordinates in the range of 0.0 to 1.0 – that is, UV coordinates – which can then be used directly to build `Vec2` UV values to put into a buffer and send to the GPU.

I'm a little unsure about the wording of the `texture_rect` documentation but I kept it intact and based mine on it. If you think this could be improved and have some advice, I'd love to include that in this PR.

## Testing

I've not done any testing with the updated bevy branch, other than seeing that the original helper function (identical in functionality) worked in my currently very small project, and making sure `cargo build` doesn't error, but I'm new to making changes to Bevy so unsure if this is sufficient.

## Showcase

image

### DIFF
--- a/crates/bevy_image/src/texture_atlas.rs
+++ b/crates/bevy_image/src/texture_atlas.rs
@@ -61,6 +61,7 @@ impl TextureAtlasSources {
     }
 
     /// Retrieves the texture *section* rectangle of the given `texture` handle in UV coordinates.
+    /// These are within the range [0..1], as a fraction of the entire texture atlas's size.
     pub fn uv_rect(
         &self,
         layout: &TextureAtlasLayout,
--- a/crates/bevy_image/src/texture_atlas.rs
+++ b/crates/bevy_image/src/texture_atlas.rs
@@ -61,7 +61,7 @@ impl TextureAtlasSources {
     }
 
     /// Retrieves the texture *section* rectangle of the given `texture` handle in UV coordinates.
-    /// These are within the range [0..1], as a fraction of the entire texture atlas's size.
+    /// These are within the range [0..1], as a fraction of the entire texture atlas' size.
     pub fn uv_rect(
         &self,
         layout: &TextureAtlasLayout,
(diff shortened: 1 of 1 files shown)


crates/bevy_image/src/texture_atlas.rs
```diff
@@ -51,14 +51,27 @@ impl TextureAtlasSources {
...
+    /// Retrieves the texture *section* rectangle of the given `texture` handle in pixels.
     pub fn texture_rect(
         &self,
         layout: &TextureAtlasLayout,
         texture: impl Into<AssetId<Image>>,
     ) -> Option<URect> {
         layout.textures.get(self.texture_index(texture)?).cloned()
     }
+
+    /// Retrieves the texture *section* rectangle of the given `texture` handle in UV coordinates.
```
alice-i-cecile, line 63: For folks who are new to game dev, another line about what UV coordinates means would be helpful.

alice-i-cecile (approved):
I think the docs could be improved, but this is a clear improvement over the status quo and a nice little helper.

crates/bevy_image/src/texture_atlas.rs
copygirl: Not a native English speaker. Does this read okay?
> Retrieves the texture *section* rectangle of the given `texture` handle in UV coordinates.
> These are within the range [0..1], as a fraction of the entire texture atlas's size.

alice-i-cecile: Yeah, that's much better!

mnmaita (approved):

Issue #18111, opened 2025-03-02
Component trait docs should link to Relationship


ItsDoot:
## How can Bevy's documentation be improved?

Given relationships are defined via the `Component` derive macro, `Relationship` and `RelationshipTarget` should at minimum be linked to on the `Component` trait, since we have yet to document the proc macro itself (😔).

PR #18179, opened 2025-03-06
Update Component docs to point to Relationship trait

krunchington:
also updates Relationship docs terminology

# Objective

- Contributes to #18111 

## Solution

Updates Component docs with a new section linking to Relationship.  Also updates some Relationship terminology as I understand it.

## Testing

  - opened Docs, verified link
  - I don't think so
  - run `cargo doc --open` and check out Component and Relationship docs, verify their correctness
  - I think this is n/a but I ran the doc command on Ubuntu 24.04.2 LTS

---

## Showcase

image

## Migration Guide

n/a

### DIFF
--- a/crates/bevy_ecs/src/relationship/mod.rs
+++ b/crates/bevy_ecs/src/relationship/mod.rs
@@ -50,7 +50,7 @@ use log::warn;
 /// #[relationship(relationship_target = Children)]
 /// pub struct ChildOf {
 ///     #[relationship]
-///     pub child: Entity,
+///     pub parent: Entity,
 ///     internal: u8,
 /// };
 ///
--- a/crates/bevy_ecs/src/component.rs
+++ b/crates/bevy_ecs/src/component.rs
@@ -282,6 +282,11 @@ use thiserror::Error;
 /// Note that requirements must currently be registered before the requiring component is inserted
 /// into the world for the first time. Registering requirements after this will lead to a panic.
 ///
+/// # Relationships between Components
+///
+/// Sometimes it is useful to define relationships between entities.  Bevy allows you to do this by
+/// implementing the [`Relationship`](`crate::relationship::Relationship`) trait.
+///
 /// # Adding component's hooks
 ///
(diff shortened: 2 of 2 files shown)


crates/bevy_ecs/src/relationship/mod.rs
```diff
@@ -50,7 +50,7 @@ use log::warn;
 /// #[relationship(relationship_target = Children)]
 /// pub struct ChildOf {
 ///     #[relationship]
-///     pub child: Entity,
+///     pub parent: Entity,
```
krunchington, line 53: I updated this after checking out the ChildOf trait docs which refers to this as parent.  I'm pretty sure this is correct but open to changes.

alice-i-cecile: This is definitely correct.

alice-i-cecile (changes requested):
Hmm. I don't love the section on relations on the `Component` trait. I'd prefer to cut it, or talk more about how this relates to components.

That said, the little fix to the `ChildOf` docs is uncontroversially correct and I would merge that change immediately if it was alone.

krunchington:
> Hmm. I don't love the section on relations on the `Component` trait. I'd prefer to cut it, or talk more about how this relates to components.
> 
> That said, the little fix to the `ChildOf` docs is uncontroversially correct and I would merge that change immediately if it was alone.

Good callouts.  I'll take a pass at revising the docs for more clarity/relevance to Components.  If the second pass still needs work, I'll split out the `ChildOf` change separately to get that in and then we can continue the chat about how best to get discoverability of the `Relationship` trait into the docs separately.

alice-i-cecile:
Much better, thanks!

krunchington:
Just realized I didn't double check the generated links before committing, doing that now

krunchington:
Updated screenshot, links look good
image

ElliottjPierce (approved):
LGTM. This is really going to help new users. If I came across this for the first time, I would want to see an example of using it, and you link to those examples. 

Congrats on your first Bevy PR!

PR #18180, opened 2025-03-06
Improve component registration performance

ElliottjPierce:
# Objective

Make component registration faster. This is a tinny, almost petty PR, but it led to roughly 10% faster registration, according to my benchmarks in #17871.

Up to y'all if we do this or not. It is completely unnecessary, but its such low hanging fruit that I wanted to put it out there.

## Solution

Instead of cloning a `HashSet`, collect it into a `SmallVec`. Since this is empty for many components, this saves a lot of allocation and hashing work.


chescock (approved):

ElliottjPierce:
I have no idea why that one merge queue check keeps failing. At a glance, it seems totally unrelated to this. Maybe a dependency was updated? If other merges are still going through, I'm really not sure what's going on here.

alice-i-cecile:
Yeah I'm similarly confused. Let me find something else to merge.

mnmaita:
Some WASM test timed out. GitHub is acting up for me today FWIW so maybe there's some slow runners out there making some jobs fail 😬

alice-i-cecile:
🥺

ElliottjPierce:
This feels like GitHub shenanigans to me. Maybe we're hitting a cache or something? If you think it would help, I'd be happy to open another PR with the same changes. Since other PRs are merging fine now, that might help, but IDK.

PR #18183, opened 2025-03-06
Deprecated Query::many and many_mut

alice-i-cecile:
# Objective

Alternative to and closes #18120.

Sibling to #18082, see that PR for broader reasoning.

Folks weren't sold on the name `many` (get_many is clearer, and this is rare), and that PR is much more complex.

## Solution

- Simply deprecate `Query::many` and `Query::many_mut`
- Clean up internal usages

Mentions of this in the docs can wait until it's fully removed in the 0.17 cycle IMO: it's much easier to catch the problems when doing that.

## Testing

CI!

## Migration Guide

`Query::many` and `Query::many_mut` have been deprecated to reduce panics and API duplication. Use `Query::get_many` and `Query::get_many_mut` instead, and handle the `Result`.

### DIFF
--- a/crates/bevy_ecs/src/query/mod.rs
+++ b/crates/bevy_ecs/src/query/mod.rs
@@ -777,7 +777,7 @@ mod tests {
         let _: Option<&Foo> = q.get(e).ok();
         let _: Option<[&Foo; 1]> = q.get_many([e]).ok();
         let _: Option<&Foo> = q.single().ok();
-        let _: [&Foo; 1] = q.many([e]);
+        let _: [&Foo; 1] = q.get_many([e]).unwrap();
         let _: &Foo = q.single().unwrap();
     }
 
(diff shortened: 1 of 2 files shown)


crates/bevy_ecs/src/query/mod.rs
```diff
@@ -777,7 +777,7 @@ mod tests {
         let _: Option<&Foo> = q.get(e).ok();
         let _: Option<[&Foo; 1]> = q.get_many([e]).ok();
         let _: Option<&Foo> = q.single().ok();
-        let _: [&Foo; 1] = q.many([e]);
+        let _: [&Foo; 1] = q.get_many([e]).unwrap();
```
chescock, line 780: This just tests the return type, and we already check `get_many`.

```suggestion
```

chescock (approved):
Looks good!  

And the only uses I see in docs are on the `many` methods themselves, so I think you may have updated all the docs after all :).

Issue #15271, opened 2024-09-17
I am getting a stack overflow error from loading too many animations at once


victor7543:
## Bevy version
Bevy 0.14.2

## \[Optional\] Relevant system information
using rust 1.81.0, Windows 11

AdapterInfo { name: "NVIDIA GeForce GTX 1070", vendor: 4318, device: 7041, device_type: DiscreteGpu, driver: "NVIDIA", driver_info: "561.09", backend: Vulkan }

## What you did

I downloaded a model from mixamo with 50 animations and I am trying to load them with the usual method I saw in bevy's examples, but I am getting an error: thread 'IO Task Pool (0)' has overflowed its stack. Here's the part of the code that is causing the problem:

```
let mut graph = AnimationGraph::new();
    let animations = graph
        .add_clips(
            [
                GltfAssetLabel::Animation(0).from_asset("Paladin1.glb"),
                GltfAssetLabel::Animation(1).from_asset("Paladin1.glb"),
                ...
                GltfAssetLabel::Animation(49).from_asset("Paladin1.glb"),
            ]
            .into_iter()
            .map(|path| assets.load(path)),    // this seems to be causing the stack overflow, but only when I load too many animations 
            1.0,
            graph.root,
        )
        .collect();

    // Insert a resource with the current scene information
    let graph = graphs.add(graph);
    commands.insert_resource(Animations {
        animations,
        graph: graph.clone(),
    });
```

## What went wrong

I am getting an error:

Running `target\debug\bevy_test.exe`
2024-09-17T17:43:47.913011Z  INFO bevy_diagnostic::system_information_diagnostics_plugin::internal: SystemInfo { os: "Windows 11 Pro", kernel: "22631", cpu: "Intel(R) Core(TM) i5-9600KF CPU @ 3.70GHz", core_count: "6", memory: "31.9 GiB" }
2024-09-17T17:43:48.052848Z ERROR wgpu_hal::vulkan::instance: GENERAL [Loader Message (0x0)]
        Failed to open dynamic library "C:\ProgramData\ReShade\.\ReShade64.dll" with error 1114
2024-09-17T17:43:48.053205Z ERROR wgpu_hal::vulkan::instance:   objects: (type: INSTANCE, hndl: 0x18d4bbefc00, name: ?)
[...]

Issue #4078, opened 2022-03-02
WebAssembly multithreading tracking issue


colepoirier:
### Motivation

Currently, Bevy can only run single threaded on WebAssembly. Bevy's architecture was carefully designed to enable maximal parallelism so that it can utilize all cores available on a system. As of about six months ago the stable versions of all browsers have released the web platform features needed to accomplish this (`SharedArrayBuffer` and related CORS security functionality). I think now is a good time to attempt to make Bevy run natively in the browser like it does on the desktop: fully multithreaded.

There are three distinct tasks that will enable the accomplishment of this goal:

1. Create a modified versions of `task_pool::{Scope, TaskPool, TaskPoolBuilder}` that run on wasm called `wasm_task_pool::{Scope, TaskPool, TaskPoolBuilder}`, and use those instead of the `single_threaded_task_pool::{Scope, TaskPool, TaskPoolBuilder}` (TODO: create issue and link here)
2. Modify bevy_audio so that it runs mutlithreaded in the background (not on the main thread) on wasm (TODO: create issue and link here)
  a. Contribute the functionality needed to to do background multithreaded audio using WebAssembly via the web platform's [AudioWorklet API](https://developer.mozilla.org/en-US/docs/Web/API/Web_Audio_API/Using_AudioWorklet) to the upstream dependencies of bevy_audio  (TODO: create issue and link here):
    i. cpal (TODO: create issue and link here)
    ii. rodio (TODO: create issue and link here)
    iii. NOTE: as outlined below in the "Insights provided by developers who have tried to make things that run multithreaded on wasm" section, it may be necessary to PR wasm-bindgen to solve this issue <https://github.com/rustwasm/wasm-bindgen/issues/2367>
  b. Make necessary changes to bevy_audio to make use of this added upstream functionality  (TODO: create issue and link here)
3. Modify bevy_ecs, bevy_render, and the rest of bevy to run multithreaded on wasm (TODO: create issue and link here)
  a. Insights from @alice-i-cecile [below](https://github.com/bevyengine/bevy/issues/4078#issuecomment-1074418691). "For the ECS, there are two relevant bits:"
    i. Our [multi-threaded executor](https://github.com/bevyengine/bevy/blob/54fbaf4b4cc497503ec1af2e40488b582586ba60/crates/bevy_ecs/src/schedule/executor_parallel.rs#L119)
[...]

PR #18186, opened 2025-03-06
Temporary fix to prevent panics on loading Gltf assets in parallel due to stack overlows

john-sharratt:
# Objective

Temporary fix for
https://github.com/bevyengine/bevy/issues/15271

## Solution

Disables the inline invocation of the task engine after loading a Gltf as the blocking call is causing task engine re-entrance and stack overflow

## Testing

Tested on my local branch which is a fork, not directly tested on the issue raiser.
Further testing would be required before accepting this however this does fix my issue

### DIFF
--- a/crates/bevy_gltf/src/loader.rs
+++ b/crates/bevy_gltf/src/loader.rs
@@ -33,8 +33,8 @@ use bevy_render::{
     },
 };
 use bevy_scene::Scene;
-#[cfg(not(target_arch = "wasm32"))]
-use bevy_tasks::IoTaskPool;
+//#[cfg(not(target_arch = "wasm32"))]
+//use bevy_tasks::IoTaskPool;
 use bevy_transform::components::Transform;
 use bevy_utils::tracing::{error, info_span, warn};
 use bevy_utils::{HashMap, HashSet};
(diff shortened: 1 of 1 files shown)


alice-i-cecile (commented):
Can you say more about why this fixes the issue observed?

Note that we won't merge commented out code, but let's work on getting the fix nice before working on polish.

john-sharratt:
Yeah no worries, this is mainly just to help out victor if he can't wait for a proper patch.

Basically the bug is because blocking code is running within async code which is generally a bit of an anti-pattern, which `tokio` and others do is shift blocking code to a dedicated blocking thread pool so edge cases like this don't happen.

Ultimately you could fix this and make it work but its not really deterministic or easy to rationalize. Better would be to make sure all your blocking code and async code run on different thread pools.

john-sharratt:
Here's some background information you can read up on blocking inside async contexts

https://docs.rs/tokio/latest/tokio/task/fn.block_in_place.html
https://tokio.rs/tokio/topics/bridging
https://ryhl.io/blog/async-what-is-blocking/

john-sharratt:
B.t.w. Now we are talking a bit more in actively and given we're on the topic of task management - I've solved the multi-threading problem in WASM running the the main browsers if you want some tips

alice-i-cecile:
@john-sharratt notes over in #4078 would be hugely appreciated!

ChristopherBiscardi:
Since this has served its purpose as an informational source for the other issues, I'm going to close it. If it is still relevant to put forward a fix for 0.19+ we can do that in a new PR.

PR #11228, opened 2024-01-05
Plugin dependency graph v2

NthTensor:
This PR builds heavily off the initial work done by https://github.com/bevyengine/bevy/pull/7839. Thanks @maniwani <3.

# Objective
+ Implement https://github.com/bevyengine/bevy/issues/69.
+ Take a small step towards https://github.com/bevyengine/bevy/issues/1255.
+ Provide a better way for one plugin to spoof/substitute another than with names.
+ Continue to allow multiple instances of the same plugin.
+ Retain as much of the current plugin API as possible.

## Solution
The TL;DR is:
+ Plugins are identified by `TypeId`.
+ Plugins have a manifest which describes their relationship to other plugins.
+ Plugin builds are now differed until `App::run`, and are sorted according to the dependency graph.
+ Plugins can no longer add other plugins to the app in `Plugin::build`.
+ The plugin collections infrastructure is now based around a growable set of plugins.

This PR contains a few features which are too closely related for me to split into seperate PRs. I recommend reviewers start with the changes in `app.rs`, then move on to read `PluginSet::order` in `plugin.rs` before taking a look at the rest of `plugin.rs`. 

### Manifests
Plugins may now implement `Plugin::configure` to specify relationships with other plugins.
```rust
impl Plugin for MyPlugin {
    fn build(&self, app: &mut App) {
        // Normal plugin stuff...
    }

    fn configure(&self, manifest: &mut PluginManifest) {
        manifest.add_dependency::<RequiredPlugin>(true);
        manifest.add_dependency::<OptionalPlugin>(false);
        manifest.add_substitution::<ReplacedPlugin>();
    }
}
```

`PluginManifest::add_dependency` tells the app to build another plugin before this one. If the dependency is marked required, the app will panic if the target plugin is disabled or not present.

`PluginManifest::add_substitution` disables the target plugin, if it is present, and makes all of it's dependencies depend on this plugin instead. It can be used for debugging, for overriding default engine behavior, or for plugin aliasing. A plugin cannot be overridden by multiple other plugins at the same time. 

When dealing with manifests, we pretend like all instances of a non-unique plugin are the same. A dependency on a non-unique plugin means all instances of that plugin will be built first. All instances of a plugin are passed a reference to the same `PluginManfiest` in `Plugin::configure`. The same dependency can be specified multiple times, but the app will panic if, for example, one instance depends on a given plugin and another instance substitutes for it.

[...]

PR #8607, opened 2023-05-12
Enable async setup of an application

mockersf:
# Objective

- Async is cool. We need more async.
- Enable async initialisation of an application

## Solution

- Add async variants of methods to add a plugin or a plugin group. They transform the `build`/`ready`/`finish` steps of the plugin init to a future, and await that future
- Inside the same app, users can mix as they want async and sync plugins, depending on what they want to do. This can be nicer for plugins that actually needs the renderer (currently the only plugin with mandatory "asyncness") to do useful work
- the `#[bevy_main]` macro can handle an async `main` function

### Current state

Plugins are built when called `add_plugins`, then the application starts and the event loop wait for plugins to be ready then finish their setup before actually starting the application
```rust
fn main() {
    App::new()
        .add_plugins(DefaultPlugins)
        .run();
}
```

### New async :sparkles:

Plugins are built, waiting for them to be ready, then finished, and then the application can start directly
```rust
#[bevy_main]
async fn main() {
    App::new()
        .add_plugins_async(DefaultPlugins)
        .await
        .run();
}
```

Issue #1255, opened 2021-01-17
Order independence of AppBuilder


alice-i-cecile:
**What problem does this solve or what need does it fill?**

As with our previous discussion on implicit ordering of systems (see #1021), insertion order mattering is frustrating to debug at scale and can make refactoring more painful. 

In addition, this avoids a common beginner pain point, where they'll insert a resource after it's used and then get an error about the resource not existing.

**What solution would you like?**

Collect all of the AppBuilder methods, then process them more holistically (such that order doesn't matter) when `AppBuilder.run()` is called.

With system insertion order no longer mattering once #1021 is stabilized, there are only a [few places](https://docs.rs/bevy/0.4.0/bevy/app/struct.AppBuilder.html) that rely on ordering (as best as I can tell).

1. Resource and state insertion. These will typically be able to be added in parallel before any other steps. Special `FromResources` resources are the only wrinkle that comes to mind. We can also catch duplicate insertion of the same type, which is a nice correctness-checking feature.
2. Adding stages. Stage ordering is typically defined relative to other stages. We'll need a little tree resolution mechanism to make sure that dependencies are added first, but solving this will be a major win for refactorability (since you won't need to think about this manually).
3. Working with State.  These will want to be handled after resources and stages are added.
4. `set_runner`. We should probably just panic if two of these are set at once, rather than overwriting.

**What alternative(s) have you considered?**

Through clean code architecture, you can often define resources in the corresponding plugin, before any other systems are added. This reduces the pain substantially, but relies on consistent code hygiene, and won't work well for resources that are used by multiple plugins.

**Additional context**

Down the road, this approach opens us up to more sophisticated optimizations and analyses, as we'll have access to all of the parts of the `AppBuilder` before they're processed, rather than only getting a backwards-looking view at best.

Ratysz:
You probably meant #1144.

tigregalis:
I don't know if this is the right place for this, but I've been reading the discussion (and trying to follow), but is it accurate to say that mostly what a Stage does is just provide a point at which the queued commands get executed? If so, would it be feasible to have a stage-less API that allows someone to simply insert a "run_commands" system and depend on it?

Ratysz:
That's not all a stage does, and is not its most important function: a stage encapsulates a group of systhems and guarantees that they will run before or after another group of systems (encapsulated by a different stage). Everything else just follows from that definition - for example, applying commands issued during the stage at the end of the stage becomes an obvious idea.

[...]

PR #18187, opened 2025-03-07
Async Plugin Construction / Plugin Dependencies

### LINKED ISSUE #69: Plugin Dependencies
Some plugins require other plugins to work. We should allow plugins to define dependencies that _must_ be registered before them. This would also help ensure that plugins are registered in the correct order (where that matters).

SpecificProtagonist:
# Objective

- Fixes #69
- alternative to #11228

Since the issue was made, Bevy's plugins have been split into three stages that are (typically) run before app update: Build, finish and cleanup. They also have a `ready` method to delay the switch from the build to the finish phase.

This allows plugins to delay up to twice while waiting on other plugins, but the specific dependency is implicit even when reading the code and the ordering of plugins matters. Solve this by making plugin building concurrent.

Also, I've just now seen #8607, which also makes plugins asynchronous. I'll need to take a closer look, but afaik it has the goal of running on `bevy_tasks` and doesn't address dependencies / plugin ordering (as it build plugins one at a time).

## Solution

Adds `Plugin::build_async`, which runs concurrently for all plugins. The plugin can wait for the world to reach whatever state it needs to continue. While you can define a custom condition, there's also a helper for waiting on a resource; please suggest other ones.

This is intended to replace the old plugin lifecycle machinery, which can be deprecated (and eventually removed) in a later PR, which will make things much simpler. As noted in #1255, we'll want to defer app construction actions (like `add_systems`); with this PR, the collected actions can be run by turning them into an async plugin.

As a followup, we should be able to use a waker and achieve the goals of #8607.

## Testing

(TODO: Add plenty of tests if we decide to do this)

## Showcase

```rust
#[derive(Resource)]
struct A(i32);

struct MyPlugin;

impl Plugin for MyPlugin {
    async fn build_async(self: Box<Self>, mut ctx: PluginContext<'_>) {
        let a = ctx.resource::<A>().await.unwrap();
        println!("got A: {}", a.0);
    }
}

struct OtherPlugin;

impl Plugin for OtherPlugin {
    async fn build_async(self: Box<Self>, mut ctx: PluginContext<'_>) {
        ctx.app().world_mut().insert_resource(A(1));
    }
}
```

## Migration Guide

(TODO depending on changes)


JMS55:
Waiting on a specific plugin to finish would be nice since sometimes you have:
* Plugin A sometimes inserts a resource, sometimes doesn't
* Plugin B wants to wait for Plugin A to finish, and then check if the resource exists or not

SpecificProtagonist:
Closing for now as this should probably be done as part of a goal / needs consensus on whether async plugins should be used for plugin ordering or only for their own sake.

PR #18188, opened 2025-03-07
Warn about uses of `RenderLayers` and `UiTargetCamera`

### LINKED ISSUE #17402: Setting a `TargetCamera` on a non-UI entity, or `RenderLayers` on a UI entity should be an error

It's easy for users to assume that `TargetCamera` or `RenderLayers` work with all entities.  The only reason this hasn't come up much is because there aren't many Bevy apps using multiple windows.

Panic if a UI entity has a `RenderLayers` component or a non-UI entity has a `TargetCamera` component. 

We could remove `TargetCamera` and make the UI respect `RenderLayers` but that would cause a lot of problems with layout, text and scale factors that would need to be resolved so it isn't really feasible at the moment.

hukasu:
# Objective

Closes #17402 .
Users mix the uses of `RenderLayers` and `UiTargetCamera`:
* `UiTargetCamera` is to be used on UI roots
* `RenderLayers` is to be used on non-UI entites

## Solution

Add observers that log a warning when using `RenderLayers` on UI nodes or `UiTargetCamera` on non-UI entities.

## Testing

Modified some example to trigger the observers

### DIFF
--- a/crates/bevy_ui/src/ui_node.rs
+++ b/crates/bevy_ui/src/ui_node.rs
@@ -2825,3 +2825,79 @@ impl Default for TextShadow {
         }
     }
 }
+
+pub(super) mod observers {
+    use bevy_app::Plugin;
+    use bevy_ecs::{
+        observer::Trigger,
+        system::Query,
+        world::{OnInsert, OnRemove},
+    };
+    use bevy_render::view::RenderLayers;
+
+    use super::{Node, UiTargetCamera};
+
+    const TARGET_CAMERA_WARNING: &str =
+        "`UiTargetCamera` has no effect on entities that are not Ui roots.";
+    const RENDER_LAYERS_WARNING: &str =
+        "`RenderLayers` has no effect on entities that are Ui nodes.";
+
+    pub struct NodeObserversPlugin;
+
+    impl Plugin for NodeObserversPlugin {
+        fn build(&self, app: &mut bevy_app::App) {
+            app.add_observer(verify_node_inserted_to_render_layers)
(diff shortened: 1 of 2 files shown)


crates/bevy_ui/src/ui_node.rs
```diff
@@ -2825,3 +2825,79 @@ impl Default for TextShadow {
...
+    ) {
+        if !nodes.contains(trigger.target()) {
+            tracing::warn!(
+                "`UiTargetCamera` was inserted to entity without `Node`, {}",
+                TARGET_CAMERA_WARNING
+            );
+        }
+    }
+
+    fn verify_node_removed_from_target_camera(
```
cart, line 2868: I don't _love_ putting observers on high traffic components / triggers like this for this. We should consider alternatives, such as using hooks or observers with UiTargetCamera. Additionally, this is the kind of work we don't want to do in deployed app.s This is the type of thing that should only run during development.

hukasu: so should i have it behind a feature flag?

hukasu: also, my initial idea was to have hooks, but `bevy_render` does not have `bevy_ui` as a dependency, so i can't query `Node` to check on it

hukasu: using `cfg(debug_assertions)` and `--release` does not show the logs

IceSentry:
I think the potential performance impacts of having observers on so many core types are way too high compared to the benefits of the warnings. I believe this will need another approach.

Issue #18185, opened 2025-03-06
Bevy's default App::update() is limiting for subapp workflows.


moonheart08:
## What problem does this solve or what need does it fill?
Currently you're restricted to Bevy's default update and extract workflow, which is limiting for more complex workflows using subapps.

## What solution would you like?
Currently, the simple solution seems like the best solution, which is to allow the user to replace App::update() with their own function. 

## What alternative(s) have you considered?

I honestly couldn't think of anything that couldn't just be implemented in terms of replacing App::update()

An explicit extract flow graph for example (to allow the user to specify which apps extract from where) could just be done with the above solution instead.

Willing to pick this up and implement it myself to meet the 0.16.x release cycle.

JMS55:
Do you have an example of what you want to use this for? Hard to judge if it's a good idea or not without specific use cases.

moonheart08:
> Do you have an example of what you want to use this for? Hard to judge if it's a good idea or not without specific use cases.

The PR I ultimately made includes a convo that has an example.

Copying here:
<details>
[15:34]moony (Kaylie!): Is there some recommended way to customize the subapp update order? I'm doing my own shenanigans with extract and runtime addition of subapps and would rather not have the defaults.<br> 
[15:35]moony (Kaylie!): atm the blocker seems to be making whatever i do work with the existing winit code <br> 
[15:37]moony (Kaylie!): (I'm already targetting latest instead of stable due to needing .sub_apps() so if it's specific to latest that's still good to know)<br> 
[16:26]moony (Kaylie!): i have worked out a scrungly solution for the meantime:
App in my App.<br> 
[16:43]Alice 🌹: Not really, this is kinda half-baked still<br> 
[16:43]moony (Kaylie!): just being able to replace update() with my own func would be sufficient tbh and likely covers most uses<br> 
[16:44]moony (Kaylie!): filing an issue about that approach atm<br> 
[16:45]Alice 🌹: Okay sweet; that's appreciated. We haven't done a ton of digging into the exact ways that users might want to extend / modify this design, so experienced-based opinions are really useful<br> 
[16:45]Alice 🌹: Mostly my reaction to subapps and windowing is :pain2: due to the difficulty testing and all of the platforms, so changes there are really slow currently<br> 
[16:46]moony (Kaylie!): my exact constraint is I'm adding some arbitrary N set of scene subapps for worlds the player could be in, and allowing the user to swap between them<br> 
[...]

PR #18189, opened 2025-03-07
Add the ability for SubApps to have a custom update function

moonheart08:
# Objective

Implement the solution proposed for #18185

## Solution

Added a new field to SubApps, `update_fn`, which contains the function to invoke for running `SubApps::update()`. `SubApps::update()` was preserved to allow invariant update behavior to be introduced in the future if necessary, and for ease of use. Broke out the Default implementation for `SubApps` so that the update function is set by default.

```rs
  let app = ...;
  // App update is unsuitable, let's do something else.
  app.set_update_fn(|sub_apps| {
    println!("Hello! I did a thing! Probably!");
    ...
  });
  
  // Now invokes our update function internally, so custom runners like the Winit runner will now run our update code!
  app.update();
```

## Testing

PR opened to run bevy test suite (my poor laptop would get mad), API fits the constraints I described in https://canary.discord.com/channels/691052431525675048/742569353878437978/1347321570603044999, full convo transcribed:

<details>
[15:34]moony (Kaylie!): Is there some recommended way to customize the subapp update order? I'm doing my own shenanigans with extract and runtime addition of subapps and would rather not have the defaults.<br> 
[15:35]moony (Kaylie!): atm the blocker seems to be making whatever i do work with the existing winit code <br> 
[15:37]moony (Kaylie!): (I'm already targetting latest instead of stable due to needing .sub_apps() so if it's specific to latest that's still good to know)<br> 
[16:26]moony (Kaylie!): i have worked out a scrungly solution for the meantime:
App in my App.<br> 
[16:43]Alice 🌹: Not really, this is kinda half-baked still<br> 
[16:43]moony (Kaylie!): just being able to replace update() with my own func would be sufficient tbh and likely covers most uses<br> 
[16:44]moony (Kaylie!): filing an issue about that approach atm<br> 
[16:45]Alice 🌹: Okay sweet; that's appreciated. We haven't done a ton of digging into the exact ways that users might want to extend / modify this design, so experienced-based opinions are really useful<br> 
[16:45]Alice 🌹: Mostly my reaction to subapps and windowing is :pain2: due to the difficulty testing and all of the platforms, so changes there are really slow currently<br> 
[16:46]moony (Kaylie!): my exact constraint is I'm adding some arbitrary N set of scene subapps for worlds the player could be in, and allowing the user to swap between them<br> 
this is very easy to do if i can just control which subapp render is running extract on (and otherwise control the extract flow myself)<br> 
[16:46]moony (Kaylie!): in other words a single "main" subapp isn't the right approach for my experiment<br> 
[16:47]Alice 🌹: Interesting. What's the ultimate motivation for that? For the most part my opinion is "don't use subapps if at all possible", and I'd generally use something like states and scenes or something for that design<br> 
[16:47]moony (Kaylie!): isolating those scenes, because UGC<br> 
[16:48]moony (Kaylie!): if one scene gets completely mucked up for whatever reason the user should be able to just close or leave it and be fine<br> 
[16:48]Alice 🌹: Aha, interesting! Okay, that makes more sense to me<br> 
</details>

### DIFF
--- a/crates/bevy_app/src/app.rs
+++ b/crates/bevy_app/src/app.rs
@@ -141,7 +141,7 @@ impl App {
         Self {
             sub_apps: SubApps {
                 main: SubApp::new(),
-                sub_apps: HashMap::default(),
+                ..Default::default()
             },
             runner: Box::new(run_once),
         }
--- a/crates/bevy_app/src/app.rs
+++ b/crates/bevy_app/src/app.rs
@@ -157,7 +157,7 @@ impl App {
         self.sub_apps.update();
     }
 
-    /// Sets the update function to use for [`update`](Self::update) on the [SubApps] collection to the provided function.
+    /// Sets the update function to use for [`update`](Self::update) on the [`SubApps`] collection to the provided function.
     pub fn set_update_fn(&mut self, func: impl Fn(&mut SubApps) + 'static) {
         self.sub_apps.update_fn = Some(Box::new(func));
     }
@@ -1441,8 +1441,8 @@ impl Termination for AppExit {
 
 #[cfg(test)]
 mod tests {
-    use core::marker::PhantomData;
-    use std::sync::Mutex;
+    use core::{marker::PhantomData, sync::atomic::AtomicBool};
+    use std::{boxed::Box, sync::Mutex};
 
     use bevy_ecs::{
         change_detection::{DetectChanges, ResMut},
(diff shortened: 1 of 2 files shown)


ChristopherBiscardi:
Generally speaking, we don't want to encourage usage of subapp workflows, which this is intended to make easier. So it makes more sense to close this.

PR #18190, opened 2025-03-07
Address `Entities::len` inconsistency 

ElliottjPierce:
# Objective

I was recently exploreing `Entities` and stumbled on something strange. `Entities::len` (the field) has the comment `Stores the number of free entities for [`len`](Entities::len)`, refering to the method. But that method says `The count of currently allocated entities.` Looking at the code, the field's comment is wrong, and the public `len()` is correct. Phew!

## Solution

So, I was just going to fix the comment, so it didn't confuse anyone else, but as it turns out, we can just remove the field entirely. As a bonus, this saves some book keeping work too. We can just calculate it on the fly.

Also, add additional length methods and documentation for completeness. These new length methods might be useful debug tools in the future.

### DIFF
--- a/crates/bevy_ecs/src/entity/mod.rs
+++ b/crates/bevy_ecs/src/entity/mod.rs
@@ -982,6 +982,14 @@ impl Entities {
         (self.meta.len() as isize - self.free_cursor.load(Ordering::Relaxed) as isize) as usize
     }
 
+    /// The count of all entities in the [`World`] that have ever been allocated or reserved, including those that are freed.
+    ///
+    /// [`World`]: crate::world::World
+    #[inline]
+    pub fn total_prospective_count(&self) -> usize {
+        self.total_count() + (-self.free_cursor.load(Ordering::Relaxed)).min(0) as usize
+    }
+
     /// The count of currently allocated entities.
     #[inline]
     pub fn len(&self) -> u32 {
--- a/crates/bevy_ecs/src/entity/mod.rs
+++ b/crates/bevy_ecs/src/entity/mod.rs
@@ -582,8 +582,6 @@ pub struct Entities {
     /// [`flush`]: Entities::flush
     pending: Vec<u32>,
     free_cursor: AtomicIdCursor,
-    /// Stores the number of free entities for [`len`](Entities::len)
-    len: u32,
 }
 
 impl Entities {
@@ -592,7 +590,6 @@ impl Entities {
             meta: Vec::new(),
             pending: Vec::new(),
             free_cursor: AtomicIdCursor::new(0),
-            len: 0,
         }
     }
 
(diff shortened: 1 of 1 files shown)


crates/bevy_ecs/src/entity/mod.rs
```diff
@@ -985,16 +973,34 @@ impl Entities {
         self.meta.len()
     }
 
+    /// The count of all entities in the [`World`] that are used,
+    /// indluding both those allocated and those reserved, but not those freed.
```
andriyDev, line 977: including

andriyDev (approved):

crates/bevy_ecs/src/entity/mod.rs
ElliottjPierce: Thanks! I'll never fight this btw, so you can always use GitHub's suggest changes feature if that's easier. Either way works for me!

crates/bevy_ecs/src/entity/mod.rs
```diff
@@ -985,16 +973,34 @@ impl Entities {
...
+    /// The count of all entities in the [`World`] that are used,
+    /// including both those allocated and those reserved, but not those freed.
+    ///
+    /// [`World`]: crate::world::World
+    #[inline]
+    pub fn used_count(&self) -> usize {
+        (self.meta.len() as isize - self.free_cursor.load(Ordering::Relaxed) as isize) as usize
+    }
+
+    /// The count of all entities in the [`World`] that have ever been allocated or reserved, including those that are freed.
```
chescock, line 985: If I understand this calculation correctly, then this is the value that `total_count()` would return if `flush()` were called right now.  Is that right?  It might be worth saying something along those lines in the doc comment.  

```suggestion
    /// The count of all entities in the [`World`] that have ever been allocated or reserved, including those that are freed.
    /// This is the value that [`Self::total_count()`] would return if [`Self::flush()`] were called right now.
```

crates/bevy_ecs/src/entity/mod.rs
```diff
@@ -985,16 +973,34 @@ impl Entities {
...
+    ///
+    /// [`World`]: crate::world::World
+    #[inline]
+    pub fn total_prospective_count(&self) -> usize {
+        self.total_count() + (-self.free_cursor.load(Ordering::Relaxed)).min(0) as usize
```
chescock, line 990: Very much a nit, but: You use `self.meta.len()` in `used_count` and `self.total_count()` here, and it took me a minute to realize they were the same value.  It might be nice to write it the same way in both places.  

```suggestion
        self.meta.len() + (-self.free_cursor.load(Ordering::Relaxed)).min(0) as usize
```

chescock (approved):

PR #18191, opened 2025-03-07
Split example file docblock and code when generating web examples markdown

doup:
# Objective

Separate example explanation (file docblock) and the code so they can be layout differently in the website and we can give a higher importance to the explanation on the [website search tool](https://github.com/bevyengine/bevy-website/pull/1935). This would also allow us to improve the examples so they become even more like a cookbook.

## Solution

Update the `example-showcase` tool to extract the example file docblock and write it as the example markdown content. This allows us to access the explanation via `page.content` in Zola.

## Testing

I've checked that the output is correct after running the tool and it doesn't throw any error. I've also validated that the approach will work on the website.

## Showcase

This is a quick and dirty example of what we could do in the web examples after the change. When we implement the real thing we can put the explanation on a sidebar or explore other layout options.

image

### DIFF
--- a/tools/example-showcase/src/main.rs
+++ b/tools/example-showcase/src/main.rs
@@ -541,7 +541,9 @@ weight = {}
                 let _ = fs::create_dir_all(&example_path);
 
                 let code_path = example_path.join(Path::new(&to_show.path).file_name().unwrap());
-                let _ = fs::copy(&to_show.path, &code_path);
+                let code = fs::read_to_string(&to_show.path).unwrap();
+                let (docblock, code) = split_docblock_and_code(&code);
+                let _ = fs::write(&code_path, code);
 
                 let mut example_index = File::create(example_path.join("index.md")).unwrap();
                 example_index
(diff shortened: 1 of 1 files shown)


alice-i-cecile (approved):

kristoff3r (approved):

krunchington (approved):
The rust specification does allow for inner doc comments anywhere in the file, but that seems like a case that's not super worth putting a ton of time into handling.  This does appear to be a clear improvement over the alternative.  Personally I don't mind reading the inner doc comments in the code block, but see little risk in this approach.

mockersf (approved):

mockersf:
> The rust specification does allow for inner doc comments anywhere in the file, but that seems like a case that's not super worth putting a ton of time into handling. This does appear to be a clear improvement over the alternative. Personally I don't mind reading the inner doc comments in the code block, but see little risk in this approach.

Here we have control of where the doc comment we want is, so not a problem 🙂

PR #11700, opened 2024-02-04
Expose query accesses

JoJoJet:
# Objective

It would be useful to be able to inspect a `QueryState`'s accesses so we can detect when the data it accesses changes without having to iterate it. However there are two things preventing this:

* These accesses are unnecessarily encapsulated.
* `Has<T>` indirectly accesses `T`, but does not register it.

## Solution

* Expose accesses and matches used by `QueryState`.
* Add the notion of "archetypal" accesses, which are not accessed directly, but whose presence in an archetype affects a query result.

PR #18192, opened 2025-03-07
Add Allows filter to bypass DefaultQueryFilters

### LINKED ISSUE #17803: Add a `Includes` query filter for working with disabling components
> IMO this should include an example of a `Query` that matches entities that are both enabled and disabled. The ones shown match only entities that are disabled.

> Has matches both!

> Ah right. It still feels kinda hacky though.

_Originally posted by @SkiFire13 in https://github.com/bevyengine/bevy/pull/17768#discussion_r1950383782_

When working with default query filters, it's reasonably common to want to be agnostic to the presence of a given disabling component.

Currently, users can stick a `Has` or `Option` query data term in there, and get the desired behavior, but they often don't want the information contained in that term.

We should provide them with a query filter equivalent, which I think should be called `Includes<C>`. This filter would be absolutely useless for normal components, as it includes entities with or without C. But if we set the access up correctly, this should trick DQF into showing us the entities we want, regardless of whether or not they have C.

NiseVoid:
# Objective

Fixes #17803 

## Solution

- Add an `Allows<T>` `QueryFilter` that adds archetypal access for `T`
- Fix access merging to include archetypal from both sides

## Testing

- Added a case to the unit test for the application of `DefaultQueryFilters`

### DIFF
--- a/crates/bevy_ecs/src/query/state.rs
+++ b/crates/bevy_ecs/src/query/state.rs
@@ -2232,6 +2232,10 @@ mod tests {
         let mut query = QueryState::<Has<C>>::new(&mut world);
         assert_eq!(3, query.iter(&world).count());
 
+        // Allows should bypass the filter entirely
+        let mut query = QueryState::<(), Allows<C>>::new(&mut world);
+        assert_eq!(3, query.iter(&world).count());
+
         // Other filters should still be respected
         let mut query = QueryState::<Has<C>, Without<B>>::new(&mut world);
         assert_eq!(1, query.iter(&world).count());
--- a/crates/bevy_ecs/src/lib.rs
+++ b/crates/bevy_ecs/src/lib.rs
@@ -79,7 +79,7 @@ pub mod prelude {
         hierarchy::{ChildOf, ChildSpawner, ChildSpawnerCommands, Children},
         name::{Name, NameOrEntity},
         observer::{Observer, Trigger},
-        query::{Added, AnyOf, Changed, Has, Or, QueryBuilder, QueryState, With, Without},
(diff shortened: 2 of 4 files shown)


crates/bevy_ecs/src/query/state.rs
```diff
@@ -2232,6 +2232,10 @@ mod tests {
         let mut query = QueryState::<Has<C>>::new(&mut world);
         assert_eq!(3, query.iter(&world).count());
 
+        // Allows should bypass the filter entirely
```
Carter0, line 2235: Is it possible to write or update a test or example explaining why allows is different than has?

Carter0 (approved):
Could use an example, but I think the implementation is good!

chescock:
I don't think this use of "archetypal" access matches the original intent.  (Although I don't know whether anyone is actually using it for its original purpose!)  

From #11700, 

> It would be useful to be able to inspect a QueryState's accesses so we can detect when the data it accesses changes without having to iterate it. 
>
> Add the notion of "archetypal" accesses, which are not accessed directly, but whose presence in an archetype affects a query result.

`Allows<T>` is sort of the opposite, as it means `T` no longer affects the query result, even though it does by default on other queries.  

I'd gently vote for making this a new kind of access.  If we do want to use "archetypal" for this, then we should probably update the doc comment to say something other than "whose presence in an archetype may affect query results".

alice-i-cecile (approved):

alice-i-cecile:
> I'd gently vote for making this a new kind of access. If we do want to use "archetypal" for this, then we should probably update the doc comment to say something other than "whose presence in an archetype may affect query results".

I'm sympathetic to this, but I'm reluctant to block this particular PR over it, or to add more complexity for a very niche feature currently.

PR #17992, opened 2025-02-23
Fix mesh tangent attribute matching in mesh transform operations

aloucks:
Fixes #17170

# Objective

Tangents are not currently transformed as described in #17170. I came across this while working on #17989 and it seemed like low hanging fruit.

PR #18194, opened 2025-03-07
Add test for mesh transformations

greeble-dev:
## Objective

Add a test that will repro issues with mesh transformations, like https://github.com/bevyengine/bevy/pull/17992.

## Solution

The PR adds a `test_transform_mesh` example that transforms a mesh in three different ways - each transformation should have the same visual result.

image

The screenshot shows the test reproducing #17992 - the second mesh from the left should have matched the two meshes on the right.

### Concerns

- Right now the test is not suitable for screenshot comparison - the light is animated.
- The mesh uses an anisotropic material so that mesh tangents and normals are significant. Anisotropy is arguably an imprecise way to verify tangents and normals. But it's convenient and does everything in a single screenshot.

## Testing

Tested on Win10/Vulkan/Nvidia.

```sh
cargo run --example test_transform_mesh --features "pbr_anisotropy_texture bevy_dev_tools"
```


greeble-dev:
CI failure appears unrelated to the PR.

https://github.com/bevyengine/bevy/actions/runs/13727867039/job/38398331235?pr=18194#logs

```
Run cargo deny check advisories
error[unmaintained]: paste - no longer maintained
    ┌─ /home/runner/work/bevy/bevy/Cargo.lock:369:1
    │
369 │ paste 1.0.15 registry+https://github.com/rust-lang/crates.io-index
    │ ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ unmaintained advisory detected
    │
    ├ ID: RUSTSEC-2024-0436
    ├ Advisory: https://rustsec.org/advisories/RUSTSEC-2024-0436
    ├ The creator of the crate `paste` has stated in the [`README.md`](https://github.com/dtolnay/paste/blob/master/README.md) 
      that this project is not longer maintained as well as archived the repository
    ├ Announcement: https://github.com/dtolnay/paste
    ├ Solution: No safe upgrade is available!
    ├ paste v1.0.15
      ├── accesskit_windows v0.24.1
      │   └── accesskit_winit v0.23.1
      │       └── bevy_winit v0.16.0-dev
      │           └── bevy_internal v0.16.0-dev
      │               ├── bevy v0.16.0-dev
      │               └── bevy_dylib v0.16.0-dev
      │                   └── bevy v0.16.0-dev (*)
      ├── macro_rules_attribute v0.2.0
      │   └── (dev) bevy v0.16.0-dev (*)
      ├── metal v0.31.0
      │   └── wgpu-hal v24.0.2
      │       ├── wgpu v24.0.1
```

mockersf:
rather than using anisotropy, didn't we use to have a debug view to show tangents/normals?

greeble-dev:
The `shader_prepass` example shows normals, but I couldn't find anything that shows tangents. Although a `TangentPrepass` might be a valid (if somewhat niche) feature?

I also considered drawing a per-vertex gizmo. But that doesn't cover the third test case which validates transforms in the shader.

So my guess is that anisotropy is the least worst choice right now.

Cargo.toml
```diff
@@ -4127,6 +4127,15 @@ description = "Demonstrates specular tints and maps"
 category = "3D Rendering"
 wasm = true
 
+[[example]]
+name = "transform_mesh"
+path = "tests/3d/transform_mesh.rs"
+doc-scrape-examples = true
+required-features = ["pbr_anisotropy_texture"]
```
greeble-dev, line 4134: I think that requiring this feature is either a bug in the shader or the documentation - the example doesn't actually use anisotropy textures, but if I disable them then anisotropy doesn't work at all.

I'll file that as a bug if this PR goes through.

IceSentry:
We have the RenderDebugOverlay now that could be used here. Conflicts also need to be resolved.

greeble-dev:
Added render debug overlay. Although this can only show normals - not tangents - so I think using anisotropy is still the least worst solution for catching bugs.

Also added supported for CI testing, but didn't actually hook it up to CI - I'm not sure about the criteria for adding new CI tests.

PR #18195, opened 2025-03-07
Remote entity reservation

### LINKED ISSUE #18003: Reserve entities from async

## Risks

ElliottjPierce:
# Objective

fixes #18003

For assets as entities and components as entities, it is convinient to reserve an entity from an `Arc`, fully outside of the ECS. See [here](https://discord.com/channels/691052431525675048/1347575124068405288/1347602981100589218) on discord for why.

It is probably possible to do both assets and components as entities without this, but we would have to eat extra complexity for that, and so would users making their own custom assets.

## Solution

This design comes from [here](https://discord.com/channels/691052431525675048/749335865876021248/1347624804668276787) on discord.

Before this PR, `EntityMeta` is stored in a single `Vec` called `meta`. A `pending` list stores free entities, and an atomic cursor manages which `pending` entities are reserved and what new entities need to be created on `meta`.

All of that remains the same, but after this PR, an additional `Vec<EntityMeta>` called `remote_entities` exists. These indices are reversed from `meta`, so the index in `remote_entities` = `u32::MAX - entity.index` and visa versa. We only extend `remote_entities` based on an `Arc<AtomicInt>`. That counter represents `remote_entities`'s length after the next flush. When we free one of these entities, we just add it to the pending list like normal, and the existing atomic cursor manages it.

## Problems

I kinda had to tiptoe around `Entity::PLACEHOLDER`. The docs for that say it *may* exist, but with this, it *will* exist the first time the remote allocator is used. For now, I just start the generations at 2 instead of 1 to avoid accidental conflicts. This is easily changed if there are other ideas.

## Alternatives

It is possible to create a paging/chunking system for Entities that would additionally allow entity partitioning. This design will simplify that if bevy chooses to add it in the future. In the meantime, this PR is much simpler than that.

## Testing

We are not using the feature anywhere, so if current tests pass, we shouldn't have an issue.

However, it makes sense to add tests for this. If anyone has specific ideas for situations to test, I'd be more than happy to write them.

### DIFF
--- a/crates/bevy_ecs/src/entity/mod.rs
+++ b/crates/bevy_ecs/src/entity/mod.rs
@@ -611,6 +611,12 @@ impl RemoteEntityReserver {
         unsafe { NonZero::new_unchecked(2) }
     };
 
+    const REMOTE_FIRST_META: EntityMeta = const {
+        let mut default = EntityMeta::EMPTY;
+        default.generation = RemoteEntityReserver::REMOTE_FIRST_GENERATION;
+        default
+    };
+
     /// Reserves an entity that will be allocated in [`Entities::flush`].
     #[expect(
         clippy::allow_attributes,
@@ -1017,28 +1023,50 @@ impl Entities {
     pub unsafe fn flush(&mut self, mut init: impl FnMut(Entity, &mut EntityLocation)) {
         let free_cursor = self.free_cursor.get_mut();
         let current_free_cursor = *free_cursor;
+        let new_free_cursor = current_free_cursor.max(0);
 
-        let new_free_cursor = if current_free_cursor >= 0 {
-            current_free_cursor as usize
-        } else {
-            let old_meta_len = self.meta.len();
-            let new_meta_len = old_meta_len + -current_free_cursor as usize;
-            self.meta.resize(new_meta_len, EntityMeta::EMPTY);
-            for (index, meta) in self.meta.iter_mut().enumerate().skip(old_meta_len) {
-                init(
-                    Entity::from_raw_and_generation(index as u32, meta.generation),
-                    &mut meta.location,
-                );
-            }
-
-            *free_cursor = 0;
-            0
-        };
-
-        for index in self.pending.drain(new_free_cursor..) {
-            let meta = &mut self.meta[index as usize];
+        // pending
+        for index in self.pending.drain((new_free_cursor as usize)..) {
+            // SAFETY: it was in pending, so it exists
+            let meta = unsafe {
+                // We can't do get_mut_by_id_unchecked because of lifetimes
+                if index as usize > self.meta.len() {
+                    self.remote_entities
+                        .get_unchecked_mut((u32::MAX - index) as usize)
+                } else {
+                    self.meta.get_unchecked_mut(index as usize)
+                }
+            };
             init(
                 Entity::from_raw_and_generation(index, meta.generation),
+                &mut meta.location,
+            );
+        }
+
+        // reserved meta
+        let old_meta_len = self.meta.len();
+        let new_meta_len = old_meta_len + -(current_free_cursor.min(0)) as usize;
+        self.meta.resize(new_meta_len, EntityMeta::EMPTY);
+        for index in old_meta_len..new_meta_len {
+            // SAFETY: we just resized to make the id valid
+            let meta = unsafe { self.meta.get_unchecked_mut(index) };
+            init(
+                Entity::from_raw_and_generation(index as u32, meta.generation),
+                &mut meta.location,
+            );
+        }
+
+        // remote reserved
+        let new_remote_len = self.remote_reservations.0.load(Ordering::Relaxed) as usize;
+        let current_remote_len = self.remote_entities.len();
+        self.verify_room_for(new_remote_len - current_remote_len);
+        self.remote_entities
+            .resize(new_remote_len, RemoteEntityReserver::REMOTE_FIRST_META);
+        for index_in_list in current_remote_len..new_remote_len {
+            // SAFETY: we just resized to make the id valid
+            let meta = unsafe { self.remote_entities.get_unchecked_mut(index_in_list) };
+            init(
+                Entity::from_raw_and_generation(u32::MAX - index_in_list as u32, meta.generation),
                 &mut meta.location,
             );
         }
--- a/crates/bevy_ecs/src/entity/mod.rs
+++ b/crates/bevy_ecs/src/entity/mod.rs
@@ -982,6 +982,14 @@ impl Entities {
         (self.meta.len() as isize - self.free_cursor.load(Ordering::Relaxed) as isize) as usize
     }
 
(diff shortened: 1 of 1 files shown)


andriyDev:
Can we add #18003 to this PR (as it should fix it)

ElliottjPierce:
Also, I started this from #18190 since it is close to merging. Should merge that before this ideally.

crates/bevy_ecs/src/entity/mod.rs
```diff
@@ -582,20 +631,82 @@ pub struct Entities {
     /// [`flush`]: Entities::flush
     pending: Vec<u32>,
     free_cursor: AtomicIdCursor,
-    /// Stores the number of free entities for [`len`](Entities::len)
-    len: u32,
+
+    /// This functions exactly like [`Self::meta`], only it counts backwards instead of forwards.
+    remote_entities: Vec<EntityMeta>,
```
cart, line 636: This is a reasonably workable solution, but it does have downsides:

1. Increases the cost of entity lookups (the fallback logic, less cache friendly as we're now hopping around two lists, etc).
2. Parallel entity reserving does not pull from the recycled entity pool. For parallel-heavy workloads (ex: assets that are continually loaded and unloaded), this could meaningfully eat into our id space.
3. It introduces a "meet in the middle" error case (which is not currently handled and solving this would introduce overhead). And given that the "no parallel recycling behavior", the odds of hitting this go up.

If we're willing to forgo recycling (like we do in the current impl), we _might_ be able to do away with the "two sided allocation" behavior to solve (1) and (3).

To do recycling, we'd probably need to introduce parallel queues (ex: what I do with our current Asset allocators) or introduce locking of some kind. Neither of which feels particularly good.

crates/bevy_ecs/src/entity/mod.rs
```diff
@@ -582,20 +631,82 @@ pub struct Entities {
...
+    /// The counter for [`Self::remote_entities`]. This will equal it's length after flushing.
```
andriyDev, line 637:
```suggestion
    /// The counter for [`Self::remote_entities`]. This will equal [`Self::remote_entities`] length after flushing.
```

crates/bevy_ecs/src/entity/mod.rs
```diff
@@ -582,20 +631,82 @@ pub struct Entities {
...
+/// Provides remote reservation of an entity from anywhere.
+#[derive(Debug, Clone)]
+pub struct RemoteEntityReserver(Arc<RemoteIdCursor>);
+
+impl RemoteEntityReserver {
+    /// We do this because [`Entity::PLACEHOLDER`] has index of `u32::MAX` and generation of 1.
+    /// By starting at 2, we ensure we don't conflict with that, and, we allow similar Entity constants to be made in the future.
+    const REMOTE_FIRST_GENERATION: NonZero<u32> = const {
+        // SAFETY: We pass 2, which is greater than 0.
+        unsafe { NonZero::new_unchecked(2) }
```
andriyDev, line 650: If we want to get rid of this unsafe, I think just doing `NonZero::new(2).unwrap()` should work? Both `new` and `unwrap` are const.

crates/bevy_ecs/src/entity/mod.rs
```diff
@@ -582,20 +631,82 @@ pub struct Entities {
...
+    };
+
+    const REMOTE_FIRST_META: EntityMeta = const {
+        let mut default = EntityMeta::EMPTY;
+        default.generation = Self::REMOTE_FIRST_GENERATION;
+        default
+    };
```
andriyDev, line 657:
```suggestion
    const REMOTE_FIRST_META: EntityMeta = EntityMeta {
      generation: Self::REMOTE_FIRST_GENERATION,
      ..EntityMeta::EMPTY
    };
```

crates/bevy_ecs/src/entity/mod.rs
```diff
@@ -582,20 +631,82 @@ pub struct Entities {
...
             meta: Vec::new(),
             pending: Vec::new(),
             free_cursor: AtomicIdCursor::new(0),
-            len: 0,
+            remote_entities: Vec::new(),
+            remote_reservations: RemoteEntityReserver(Arc::new(RemoteIdCursor::new(0))),
         }
     }
 
+    /// Constructs a new [`RemoteEntityReserver`] for this [`Entities`].
```
andriyDev, line 703:
```suggestion
    /// Get a [`RemoteEntityReserver`] for this [`Entities`].
```

The previous description sounded like it's creating like a new allocator or something to me.

crates/bevy_ecs/src/entity/mod.rs
```diff
@@ -841,25 +960,21 @@ impl Entities {
     /// Clears all [`Entity`] from the World.
     pub fn clear(&mut self) {
         self.meta.clear();
-        self.pending.clear();
         *self.free_cursor.get_mut() = 0;
-        self.len = 0;
+        self.pending.clear();
+        self.remote_entities.clear();
+        self.remote_reservations.0.fetch_and(0, Ordering::Relaxed);
```
andriyDev, line 966:
```suggestion
        self.remote_reservations.0.store(0, Ordering::Relaxed);
```

crates/bevy_ecs/src/entity/mod.rs
```diff
@@ -106,6 +107,19 @@ use bevy_platform_support::sync::atomic::AtomicIsize as AtomicIdCursor;
...
+type RemoteInner = u64;
+
+/// Most modern platforms support 64-bit atomics, but some less-common platforms
+/// do not. This fallback allows compilation using a 32-bit cursor instead, with
+/// the caveat that some conversions may fail (and panic) at runtime.
+#[cfg(not(target_has_atomic = "64"))]
+use bevy_platform_support::sync::atomic::AtomicUsize as RemoteIdCursor;
+#[cfg(not(target_has_atomic = "64"))]
+type RemoteInner = usize;
+
```
andriyDev, line 122: I don't think these need to be u64? From what I understand, the normal cursor needs to be i64 because 1) we need to be able to handle 2^32 entities in pending, 2) we need to be able to handle 2^32 entities to be allocated that requires 33 bits to go +2^32 and -2^32. However, remote reservation only ever increments the cursor - it never goes down. So we only need to handle the +2^32, so we can stick to u32.

andriyDev (changes requested):

crates/bevy_ecs/src/entity/mod.rs
andriyDev: In regards to 2: I think we should just recreate the whole structure of pending + free_cursor for the parallel stuff. Whenever we flush, we can transfer the freed "parallel" entities to the `RemoteEntityReservation`. This will require us to lock on flush, and lock whenever reserving entities using `RemoteEntityReservation`. Personally I don't think this is a big issue - I think it's fine if remote entity reservation is slow. What is most important is non-remote entity reservation should be as fast as we can make it.

ElliottjPierce:
Closing in favor of v9: #18670

PR #18196, opened 2025-03-07
action shouldn't comment on a pr if last comment is from action

mockersf:
# Objective

- While a PR is in work, if several commits fail and trigger a comment from the action, this can create a lot of comments from actions with low value
- Noticed on https://github.com/bevyengine/bevy/pull/18139

## Solution

- If last comment on a PR is from an action, don't add a new comment

### DIFF
--- a/.github/workflows/example-run-report.yml
+++ b/.github/workflows/example-run-report.yml
@@ -96,8 +96,20 @@ jobs:
           else
             echo "result=false" >> $GITHUB_OUTPUT
           fi
+      - name: "Check if last comment is already from actions"
+        id: check-last-comment
+        env:
+          PR: ${{ needs.make-macos-screenshots-available.outputs.pr-number }}
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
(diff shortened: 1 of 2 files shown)


kristoff3r (approved):

.github/workflows/example-run-report.yml
```diff
@@ -96,8 +96,20 @@ jobs:
           else
             echo "result=false" >> $GITHUB_OUTPUT
           fi
+      - name: "Check if last comment is already from actions"
+        id: check-last-comment
+        env:
+          PR: ${{ needs.make-macos-screenshots-available.outputs.pr-number }}
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          if [[ `gh api --jq '.[-1].user.login' /repos/bevyengine/bevy/issues/18139/comments` == 'github-actions[bot' ]]
```
mnmaita, line 105:
```suggestion
          if [[ `gh api --jq '.[-1].user.login' /repos/bevyengine/bevy/issues/$PR/comments` == 'github-actions[bot' ]]
```

Shouldn't this use the PR env var too?

It might be good to put this into an /actions folder for better reusability.

mockersf: oh 🤦 a copy paste I missed to fix. thanks!

mnmaita (approved):

PR #18198, opened 2025-03-08
Introduce public `World::register_dynamic_bundle` method

urben1680:
Registering dynamic bundles was not possible for the user yet.

It is alone not very useful though as there are no methods to clone, move or remove components via a `BundleId`. This could be a follow-up work if this PR is approved and such a third (besides `T: Bundle` and `ComponentId`(s)) API for structural operation is desired. I certainly would use a hypothetical `EntityClonerBuilder::allow_by_bundle_id`.

I personally still would like this register method because I have a `Bundles`-like custom data structure and I would like to not reinvent the wheel. Then instead of having boxed `ComponentId` slices in my collection I could look up explicit and required components there.

For reference scroll up to the typed version right above the new one.


urben1680:
If this is too little to add, I can also extend `EntityClonerBuilder` and add `remove_by_bundle_id` to `EntityCommands`/`EntityWorldMut`.

crates/bevy_ecs/src/world/mod.rs
```diff
@@ -3117,6 +3117,20 @@ impl World {
         // SAFETY: We just initialized the bundle so its id should definitely be valid.
         unsafe { self.bundles.get(id).debug_checked_unwrap() }
     }
+
+    /// Registers the given component ids as a dynamic bundle and returns both the required component
+    /// ids and the bundle id.
+    ///
+    /// # Panics
+    /// This function will panic if any of the provided [`ComponentId`]s do not exist in this [`World`].
```
alice-i-cecile, line 3125: This warning is quite confusing to me. Isn't the *point* to register these componentids?

urben1680: The point is to register existing component ids are a bundle. If you wanted to register the components themselves, you'd need to also provide `ComponentDescriptor` for each id.

That would also make this less versatile because then you can only bundle components fully unknown to the world. You could not bundle dynamic + typed components.

alice-i-cecile: Okay, that makes sense to me and I can see the motivation now. Add this sort of explanation to the docs and then I'll approve and merge :)

alice-i-cecile (approved):

PR #18199, opened 2025-03-08
Cleaner macros utilizing syn::Member

Bleachfuel:
# Objective

Some macros were handling Idents and indexes as seperate things, we can use syn::Member to make this more readable and nicer.

## Testing

ran all revelant crate tests.


crates/bevy_macro_utils/src/member.rs
```diff
@@ -0,0 +1,45 @@
...
+///     .map(|(i, field)| { as_member(&field.ident, i) });
+///
+/// // it won't matter now if it's a named field or a unnamed field. e.g self.field or self.0
+/// quote!(
+///     #(self.#field_members.do_something();)*
+///     );
+///
+/// ```
+///
+pub fn as_member(ident: &Option<Ident>, index: usize) -> Member {
```
MrGVSV, line 43: Nit: I believe `Option<&Ident>` is more idiomatic. Any reason we shouldn't prefer that here?

crates/bevy_macro_utils/src/member.rs
```diff
@@ -0,0 +1,45 @@
...
+    ident.clone().map_or(Member::from(index), Member::Named)
```
MrGVSV, line 44: This will always call `Member::from`, which probably isn't a huge deal but could be avoided if we use `map_or_else`
```suggestion
    ident.clone().map_or_else(|| Member::from(index), Member::Named)
```

.DS_Store
MrGVSV, line 1: This should be removed.

As a Mac user myself, this happens too often 😭

crates/bevy_macro_utils/src/member.rs
Bleachfuel: No preference, ill add it

.DS_Store
Bleachfuel: How?

Bleachfuel: nvm

Bleachfuel:
#[derive(QueryData)]
#[query_data(mutable(foo))]
//~^ ERROR: `mutable` does not take any arguments
struct MutableInvalidAttributeParameters {
    a: &'static mut Foo,
}should we really give a error for this? seems so unnessesary and extra code. When u do this it already says "expected `,`"

chescock (approved):
Yeah, this seems cleaner!  

It's a little unfortunate that `syn::Member` doesn't have the rest of the `Field` value, forcing most of the uses here to do `fields.iter().enumerate().filter().map()` instead of just `field.members()`.  Would it ever make sense to add our own wrapper type around a `&Field` and its index, along with a function to iterate them from a `Fields`, so that we could share a little more code here?

Bleachfuel:
Yeah we could do that, but would be better as a follow up pr

james7132 (approved):
This is so much cleaner! LGTM.

alice-i-cecile:
@Bleachfuel, I would rather like to merge this, but the merge conflicts are non-trivial. Could you take a look?

alice-i-cecile:
CI is still failing unfortunately :(

alice-i-cecile:
Closing as adopted.

PR #18202, opened 2025-03-08
Add Saturation trait to bevy_color

boondocklabs:
# Objective

- Allow for convenient access and mutation of color saturation providing following the `Hue`, `Luminance` traits.

- `with_saturation()` builder method
- `saturation()` to get the saturation of a `Color`
- `set_saturation()` to set the saturation of a mutable `Color`

## Solution

- Defined `Saturation` trait in `color_ops.rs`
- Implemented `Saturation` on `Hsla` and `Hsva`
- Implemented `Saturation` on `Color` which proxies to other color space impls.
- In the case of colorspaces which don't have native saturation components
  the color is converted to 'Hsla` internally

## Testing

- Empirically tested
---

## Showcase

```rust
fn next_golden(&mut self) -> Color {
    self.current
        .rotate_hue((1.0 / GOLDEN_RATIO) * 360.0)
        .with_saturation(self.rand_saturation())
        .with_luminance(self.rand_luminance())
        .into()
}
```

### DIFF
--- a/crates/bevy_color/src/color_ops.rs
+++ b/crates/bevy_color/src/color_ops.rs
@@ -99,7 +99,6 @@ pub trait Hue: Sized {
 ///
 /// When working with color spaces that do not have native saturation components
 /// the operations are performed in ['Hsla`].
-
 pub trait Saturation: Sized {
     /// Return a new version of this color with the saturation channel set to the given value.
     fn with_saturation(&self, saturation: f32) -> Self;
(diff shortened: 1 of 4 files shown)


Bleachfuel (approved):

alice-i-cecile (approved):

crates/bevy_color/src/color_ops.rs
```diff
@@ -95,6 +95,18 @@ pub trait Hue: Sized {
     }
 }
 
+/// Trait for manipulating the saturation of a color.
```
alice-i-cecile, line 98:
```suggestion
/// Trait for manipulating the saturation of a color.
///
/// When working with color spaces that do not have native saturation components
/// the operations are performed in ['Hsla`].

```

PR #18206, opened 2025-03-09
Improve `Segment2d`/`Segment3d` API and docs

Jondolf:
# Objective

#17404 reworked the `Segment2d` and `Segment3d` types to be defined by two endpoints rather than a direction and half-length. However, the API is still very minimal and limited, and documentation is inconsistent and outdated.

## Solution

Add the following helper methods for `Segment2d` and `Segment3d`:

- `from_scaled_direction`
- `from_ray_and_length`
- `length_squared`
- `direction`
- `try_direction`
- `scaled_direction`
- `transformed`
- `reversed`

`Segment2d` has a few 2D-specific methods:

- `left_normal`
- `try_left_normal`
- `scaled_left_normal`
- `right_normal`
- `try_right_normal`
- `scaled_right_normal`

There are now also `From` implementations for converting `[Vec2; 2]` and `(Vec2, Vec2)` to a `Segment2d`, and similarly for 3D.

I have also updated documentation to be more accurate and consistent, and simplified a few methods.

---

## Prior Art

Parry's [`Segment`](https://docs.rs/parry2d/latest/parry2d/shape/struct.Segment.html) type has a lot of similar methods, though my implementation is a bit more comprehensive. A lot of these methods can be useful for various geometry algorithms.

### DIFF
--- a/crates/bevy_math/src/primitives/dim2.rs
+++ b/crates/bevy_math/src/primitives/dim2.rs
@@ -1242,7 +1242,7 @@ impl Segment2d {
         (Self::new(point1, point2), (point1 + point2) / 2.)
     }
 
-    /// Create a new `Segment2d` at the origin with the given direction and length.
+    /// Create a new `Segment2d` centered at the origin with the given direction and length.
     ///
     /// The endpoints will be at `-direction * length / 2.0` and `direction * length / 2.0`.
     #[inline(always)]
--- a/crates/bevy_math/src/primitives/dim2.rs
+++ b/crates/bevy_math/src/primitives/dim2.rs
@@ -1456,9 +1456,8 @@ impl Segment2d {
 
     /// Swaps the two endpoints of the line segment.
     #[inline(always)]
-    pub fn swap(&mut self) {
-        let [a, b] = &mut self.vertices;
-        core::mem::swap(a, b);
+    pub fn reversed(&self) -> Self {
+        Segment2d::new(self.point2(), self.point1())
     }
 }
 
--- a/crates/bevy_math/src/primitives/dim2.rs
+++ b/crates/bevy_math/src/primitives/dim2.rs
@@ -1454,10 +1454,19 @@ impl Segment2d {
         segment.translated(offset_from_origin)
     }
 
-    /// Swaps the two endpoints of the line segment.
+    /// Reverses the direction of the line segment by swapping the endpoints.
     #[inline(always)]
-    pub fn reversed(&self) -> Self {
-        Segment2d::new(self.point2(), self.point1())
+    pub fn reverse(&mut self) {
+        let [point1, point2] = &mut self.vertices;
+        core::mem::swap(point1, point2);
+    }
+
+    /// Returns the line segment with its direction reversed by swapping the endpoints.
+    #[inline(always)]
+    #[must_use]
+    pub fn reversed(mut self) -> Self {
+        self.reverse();
+        self
     }
 }
 
--- a/crates/bevy_math/src/primitives/dim2.rs
+++ b/crates/bevy_math/src/primitives/dim2.rs
@@ -1288,6 +1288,13 @@ impl Segment2d {
         self.vertices[1]
     }
 
+    /// Compute the midpoint between the two endpoints of the line segment.
+    #[inline(always)]
+    #[doc(alias = "midpoint")]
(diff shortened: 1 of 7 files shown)


crates/bevy_math/src/primitives/dim2.rs
```diff
@@ -1227,70 +1227,202 @@ pub struct Segment2d {
...
+        }
     }
 
-    /// Get the position of the first point on the line segment
+    /// Create a new `Segment2d` at the origin from a vector representing
+    /// the direction and length of the line segment.
+    ///
+    /// The endpoints will be at `-scaled_direction / 2.0` and `scaled_direction / 2.0`.
+    #[inline(always)]
+    pub fn from_scaled_direction(scaled_direction: Dir2) -> Self {
```
IQuick143, line 1261: I'm not sure I understand what scaled direction means here.
The Dir2 is normalised anyway?

Jondolf: Whoops, this was meant to be just a vector 🙈

IQuick143 (approved):

crates/bevy_math/src/primitives/dim2.rs
```diff
@@ -1299,28 +1431,49 @@ impl Segment2d {
...
         let segment = Segment2d::new(centered.point1() * ratio, centered.point2() * ratio);
         segment.translated(offset_from_origin)
     }
+
+    /// Swaps the two endpoints of the line segment.
+    #[inline(always)]
+    pub fn swap(&mut self) {
+        let [a, b] = &mut self.vertices;
+        core::mem::swap(a, b);
+    }
```
greeble-dev, line 1462: I found the `&mut self` and the name a bit surprising. Could consider this?

```suggestion
    pub fn reversed(&self) -> Self {
        Segment2d::new(self.point2(), self.point1())
    }
```

crates/bevy_math/src/primitives/dim2.rs
```diff
@@ -1227,70 +1227,202 @@ pub struct Segment2d {
...
-    /// Panics if `point1 == point2`
+    /// Create a new `Segment2d` from its endpoints and compute its geometric center.
     #[inline(always)]
     #[deprecated(since = "0.16.0", note = "Use the `new` constructor instead")]
     pub fn from_points(point1: Vec2, point2: Vec2) -> (Self, Vec2) {
         (Self::new(point1, point2), (point1 + point2) / 2.)
     }
 
-    /// Create a new `Segment2d` at the origin from a `direction` and `length`
+    /// Create a new `Segment2d` at the origin with the given direction and length.
```
greeble-dev, line 1245:
```suggestion
    /// Create a new `Segment2d` centered at the origin with the given direction and length.
```

"at the origin" feels a bit ambiguous - could mean "starting from the origin". Also considered "through the origin".

greeble-dev (commented):
Suggested some small changes.

crates/bevy_math/src/primitives/dim2.rs
IQuick143: "centered at the origin" seems best to me.

crates/bevy_math/src/primitives/dim2.rs
IQuick143: You could also consider `flip(&mut self)`.

alice-i-cecile (approved):

Jondolf:
@alice-i-cecile Those only changed the 2D methods so 3D is now inconsistent. I can fix it in a sec though

alice-i-cecile:
Thank you very much @Jondolf :)

Jondolf:
Also I would prefer if the reversing method memswapped like what I originally had, since that's what I personally need for my use cases (e.g. I have a segment clipping algo where I need to sort features, which includes swapping things around).

We could have both `reverse` (in-place mutation) and `reversed` (returns a new segment) to match the existing API of `Triangle2d`/`Triangle3d`. I can do that either here or in a follow-up, either is fine for me

alice-i-cecile:
@Jondolf let's add both here.

Jondolf:
Done! Also added `#[must_use]` to `reversed` to be more explicit about it returning a new segment instead of mutating the original segment

alice-i-cecile:
Much appreciated: you always do such excellent work.

Issue #12280, opened 2024-03-03
What is the relationship between `AmbientColor::brightness` and Blender's background color strength?


janhohenheim:
During Bevy 0.13's development, lots of attention went to `PointLight`, `DirectionalLight` and `Spotlight`. However, I failed to find any significant discussion of `AmbientLight`, neither on GitHub nor Discord. I'd be glad to be proved wrong if anyone's got a link though :)

The reason I care about this because the general consensus has been "let's try to have some kind of conformance with Blender", using Blender's EV100 and render outputs as direct references for how Bevy should behave. This worked out (mostly, there's still talk about Blender itself being wrong in places) for the lights mentioned, but `AmbientLight` kind of got left behind.

In Blender, under the tab "World", you can find the background light (aka environment light). The meaning behind its `strength` param is a bit elusive to me, and [the docs are not helping](https://docs.blender.org/manual/en/latest/render/shader_nodes/shader/background.html).

This is how Blender renders a scene with no lights and only a background light of strength 1:

![image](https://github.com/bevyengine/bevy/assets/9047632/00ca90de-016f-4ff7-8fbe-3b2f04d03fb0)

This is how Bevy renders the same scene with no lights and an `AmbientLight` of the same color with brightness 1:
![image](https://github.com/bevyengine/bevy/assets/9047632/eb0830e1-826d-44e1-bf66-745be6549c70)

As you can see, Bevy is *much* darker. (Ignore the procedural sky)

Let's bump the brightness to the [slightly magic constant 683](https://github.com/search?q=repo%3AKhronosGroup%2FglTF-Blender-IO%20PBR_WATTS_TO_LUMENS%20&type=code):
![image](https://github.com/bevyengine/bevy/assets/9047632/3687820d-de71-461c-87a4-a1ad8b76acb5)

Better, but very blue.

https://github.com/google/filament/issues/1667 says that using Blender's exposure settings (which are Bevy's standard settings), the background strength should be the same as their IBL intensity, which I believe corresponds to Bevy's [`EnvironmentMapLight::intensity`](https://docs.rs/bevy/latest/bevy/pbr/prelude/struct.EnvironmentMapLight.html#structfield.intensity). I would love to test this, but I fail to create a valid environment map from a simple `Color` using `Image::new` since I keep getting an error like "Texture binding 13 expects dimension = Cube, but given a view with dimension = D(1, 2 or 3)" no matter what I do. I am not an expert in the `Image` API, so this is probably possible in some way.

[...]

PR #11868, opened 2024-02-15
New Exposure and Lighting Defaults (and calibrate examples)

cart:
# Objective

After adding configurable exposure, we set the default ev100 value to `7` (indoor). This brought us out of sync with Blender's configuration and defaults. This PR changes the default to `9.7` (bright indoor or very overcast outdoors), as I calibrated in #11577. This feels like a very reasonable default.

The other changes generally center around tweaking Bevy's lighting defaults and examples to play nicely with this number, alongside a few other tweaks and improvements. 

Note that for artistic reasons I have reverted some examples, which changed to directional lights in #11581, back to point lights.
 
Fixes #11577 

---

## Changelog

- Changed `Exposure::ev100` from `7` to `9.7` to better match Blender
- Renamed `ExposureSettings` to `Exposure`
- `Camera3dBundle` now includes `Exposure` for discoverability
- Bumped `FULL_DAYLIGHT ` and `DIRECT_SUNLIGHT` to represent the middle-to-top of those ranges instead of near the bottom
- Added new `AMBIENT_DAYLIGHT` constant and set that as the new `DirectionalLight` default illuminance. 
- `PointLight` and `SpotLight` now have a default `intensity` of 1,000,000 lumens. This makes them actually useful in the context of the new "semi-outdoor" exposure and puts them in the "cinema lighting" category instead of the "common household light" category. They are also reasonably close to the Blender default.
- `AmbientLight` default has been bumped from `20` to `80`.

## Migration Guide

- The increased `Exposure::ev100` means that all existing 3D lighting will need to be adjusted to match (DirectionalLights, PointLights,  SpotLights, EnvironmentMapLights, etc). Or alternatively, you can adjust the `Exposure::ev100` on your cameras to work nicely with your current lighting values. If you are currently relying on default intensity values, you might need to change the intensity to achieve the same effect. Note that in Bevy 0.12, point/spot lights had a different hard coded ev100 value than directional lights. In Bevy 0.13, they use the same ev100, so if you have both in your scene, the _scale_ between these light types has changed and you will likely need to adjust one or both of them.

PR #18207, opened 2025-03-09
Re-introduce "Replace Ambient Lights with Environment Map Lights (#17482)"

### LINKED ISSUE #17367: Remove ambient light as a resource
I'd be in favor of marking the resource as deprecated, but we can do that later.

_Originally posted by @pcwalton in https://github.com/bevyengine/bevy/pull/17343#pullrequestreview-2547299084_
                        
I agree with this: there should be one clear way to do this.

SparkyPotato:
#17482 was reverted due to unintentional changes in ambient light rendering. 
However, after further inspection, that was due to environment maps being more physically-correct, so the changes are positive.

Closes #17367.


alice-i-cecile:
> However, after further inspection, that was due to environment maps being more physically-correct, so the changes are positive.

Can you say more about this? Has this been discussed elsewhere?

I'd really like to do this, but we need clear rationale for changing how user scenes are rendered.

mockersf:
`AmbientLight` is still a resource that is initialised by default with this PR, but also marked as deprecrated.

Does that mean we'll change again how every scene renders in the next version?

examples/3d/auto_exposure.rs
```diff
@@ -101,11 +101,11 @@ fn setup(
...
-    commands.insert_resource(AmbientLight {
-        color: Color::WHITE,
-        brightness: 0.0,
-        ..default()
-    });
+    #[expect(
+        deprecated,
+        reason = "Once AmbientLight is removed, the resource can be removed"
+    )]
+    commands.insert_resource(AmbientLight::NONE);
```
mockersf, line 108: Example code shouldn't use deprecated things, they need to be updated to use the new component.

mockersf (changes requested):
Example code shouldn't use deprecated things, they need to be updated to use the new component.

mockersf:
this would also close https://github.com/bevyengine/bevy/issues/17367

aloucks:
Would it make sense to rename the `EnvironmentMapLight` struct to `AmbientLight`? 

Conceptually, it _is_ ambient light, right? Isn't it just being implemented with an environment map?

IceSentry:
> Can you say more about this? Has this been discussed elsewhere?

It was discussed on discord.

> Does that mean we'll change again how every scene renders in the next version?

There's no plan that I'm aware of to do other changes related to this. The main goal is still to yeet the non physical ambient light but have an easy way to still emulate an ambient light with less overall shader code to maintain.

> Would it make sense to rename the EnvironmentMapLight struct to AmbientLight?

No, because the `EnvironmentMapLight` is still using an environment map, it can just be used with a small map to emulate an ambient light, but it's more advanced than just a flat ambient light.

mockersf:
is there an explanation why some scenes are darker (https://pixel-eagle.com/project/b25a040a-a980-4602-b90c-d480ab84076d/run/9021/compare/9009?screenshot=3D+Rendering/ssao.png or https://pixel-eagle.com/project/b25a040a-a980-4602-b90c-d480ab84076d/run/9021/compare/9009?screenshot=3D+Rendering/pbr.png) and some are lighter (https://pixel-eagle.com/project/b25a040a-a980-4602-b90c-d480ab84076d/run/9021/compare/9009?screenshot=3D+Rendering/spotlight.png or https://pixel-eagle.com/project/b25a040a-a980-4602-b90c-d480ab84076d/run/9021/compare/9009?screenshot=3D+Rendering/ssao.png)?

on https://pixel-eagle.com/project/b25a040a-a980-4602-b90c-d480ab84076d/run/9021/compare/9009?screenshot=3D+Rendering/anisotropy.png some details completely disappear. for example the dark bar between the light filaments that should be there when looking at https://github.com/KhronosGroup/glTF-Sample-Assets/tree/main/Models/AnisotropyBarnLamp

rparrett:
I feel like we should make some effort to preserve the scenes of our users. If the new light is brighter because it is more physically correct then it seems like the default intensity should be lower. Or we at least need to tell users what happened and how to fix their scenes.

Just linking for reference, the PR that set the current default ambient intensity: [11868](https://github.com/bevyengine/bevy/pull/11868).

mockersf:
> Examples that modified the ambient light have been updated to use environment map lights instead

There are still examples that use ambient light, they should be fixed now

mockersf:
I also think that if we deprecate `AmbientLight`, it should be not kept as a resource initialised by default.

SparkyPotato:
> I also think that if we deprecate `AmbientLight`, it should be not kept as a resource initialised by default.

This makes sense, but I'm not sure how to replace that with a default environment map, because the new constructors require an `&mut Assets<Image>` to create the 1x1 cubemaps. 

While I do think that having no lights by default is a good idea, as it forces users to be deliberate with their lighting setup, it does hurt quick prototyping.

hukasu:
can't you create a default image on startup to use for the default environment map?

hukasu:
bevy already has this 1x1 default images

https://github.com/bevyengine/bevy/blob/3b9e2e640f41fcc7c83dc477161e1b6b96764585/crates/bevy_render/src/texture/mod.rs#L24-L30

other locations with images created on setup

https://github.com/bevyengine/bevy/blob/3b9e2e640f41fcc7c83dc477161e1b6b96764585/crates/bevy_core_pipeline/src/post_process/mod.rs#L56-L61

https://github.com/bevyengine/bevy/blob/3b9e2e640f41fcc7c83dc477161e1b6b96764585/crates/bevy_core_pipeline/src/smaa/mod.rs#L86-L90

mockersf:
yup you can static embed small assets, a 1x1 cubemap shouldn't use a lot of memory

SparkyPotato:
I've swapped out the default `AmbientLight` resource to a required component on `Camera3d` that is added at runtime only if the user enables `PbrPlugin::default_environment_map_light`. I've also reduced the default intensity so that scenes are a similar brightness to before.

This should mean that users will only be using the `AmbientLight` fallback if they've explicitly added or changed the resource, in which case they'll also get the deprecated warning. On that note, do we really want to keep `AmbientLight` around for a version or should it just be removed altogether?

alice-i-cecile:
> On that note, do we really want to keep AmbientLight around for a version or should it just be removed altogether?

I think that the migration will be easier if we do a deprecation notice. This will point the users to the right spot.

inodentry:
I missed the original discussion and PR, but I'd like to ask something about the whole idea of removing ambient light.

Isn't the whole point of ambient light that it is extremely cheap? AFAIK, it's just another term added to the lighting equations computed in the shader, no? Just some extra arithmetic.

Whereas environment map lighting requires texture sampling operations (much slower operation for the GPU) in addition to computation.

I thought the reason why ambient light exists is to have an ultra-cheap fallback for very low-end GPUs. That alone makes it a useful feature. Yes, it looks ugly and is not physically correct, but that's kinda the point.

Using environment map lighting looks better, yes, but also costs performance, and that kinda defeats the point.

Emulating the look of ambient light (ugly) with environment maps (slower perf) just sounds like the worst of both worlds.

IceSentry:
If using an envmap is already too slow for you, then you should probably not be using bevy's default PBR shader and should be using a custom material that is more appropriate for that use case. And if bevy's PBR is good enough, then the envmap is likely not the bottleneck and it's more accurate. This PR also has the nice side effect of making scenes with only an envmap faster.

alice-i-cecile:
@SparkyPotato I'm going to cut this from the milestone as it's not ready for a final decision. I'd still very much like to do this, but it looks like there's a few things for you to clean up before I summon the SMEs.

SparkyPotato:
> @SparkyPotato I'm going to cut this from the milestone as it's not ready for a final decision. I'd still very much like to do this, but it looks like there's a few things for you to clean up before I summon the SMEs.

Pretty sure everything has been addressed already.

crates/bevy_light/src/probe.rs
```diff
@@ -96,6 +100,88 @@ pub struct EnvironmentMapLight {
...
+
+    /// An environment map with a uniform color, useful for uniform ambient lighting.
+    pub fn solid_color(assets: &mut Assets<Image>, color: Color) -> Self {
+        let handle = assets.add(Self::solid_color_image(color));
+        Self {
+            diffuse_map: handle.clone(),
+            specular_map: handle,
+            ..Default::default()
+        }
+    }
```
atlv24:
```suggestion
    /// An environment map with a uniform color, useful for uniform ambient lighting.
    pub fn solid_color(assets: &mut Assets<Image>, color: Color) -> Self {
        hemispherical_gradient(assets, color, color)
    }
```

crates/bevy_light/src/probe.rs
```diff
@@ -96,6 +100,88 @@ pub struct EnvironmentMapLight {
...
+
+    /// An environment map with a hemispherical gradient, fading between the sky and ground colors
+    /// at the horizon. Useful as a very simple 'sky'.
+    pub fn hemispherical_gradient(
+        assets: &mut Assets<Image>,
+        top_color: Color,
```
atlv24, line 113: consider making mid_color a parameter too

crates/bevy_pbr/src/render/pbr_ambient.wgsl
```diff
@@ -1,29 +0,0 @@
...
-    perceptual_roughness: f32,
-    occlusion: vec3<f32>,
-) -> vec3<f32> {
-    let diffuse_ambient = EnvBRDFApprox(diffuse_color, F_AB(1.0, NdotV));
-    let specular_ambient = EnvBRDFApprox(specular_color, F_AB(perceptual_roughness, NdotV));
-
-    // No real world material has specular values under 0.02, so we use this range as a
-    // "pre-baked specular occlusion" that extinguishes the fresnel term, for artistic control.
-    // See: https://google.github.io/filament/Filament.html#specularocclusion
-    let specular_occlusion = saturate(dot(specular_color, vec3(50.0 * 0.33)));
```
atlv24, line 26: do env-maps have a similar treatment for this?

examples/3d/irradiance_volumes.rs
```diff
@@ -414,7 +422,7 @@ fn toggle_irradiance_volumes(
     light_probe_query: Query<Entity, With<LightProbe>>,
     mut app_status: ResMut<AppStatus>,
     assets: Res<ExampleAssets>,
-    mut ambient_light: ResMut<AmbientLight>,
+    mut ambient_light: Query<&mut EnvironmentMapLight>,
```
atlv24, line 416: probably rename this to envmap_light or similar

examples/2d/custom_gltf_vertex_attribute.rs
```diff
@@ -63,7 +59,13 @@ fn setup(
         Transform::from_scale(150.0 * Vec3::ONE),
     ));
 
-    commands.spawn(Camera2d);
+    commands.spawn((
+        Camera2d,
+        EnvironmentMapLight {
+            intensity: 1.0 / 5.0,
+            ..EnvironmentMapLight::solid_color(&mut images, Color::WHITE)
```
atlv24: Shouldn't we just be falling back to the default `DEFAULT_ENVIRONMENT_MAP_TEXTURE_HANDLE` for all the white examples?
```suggestion
            ..default()
```

examples/3d/irradiance_volumes.rs
```diff
@@ -240,6 +244,10 @@ fn spawn_camera(commands: &mut Commands, assets: &ExampleAssets) {
             brightness: 150.0,
             ..default()
         },
+        EnvironmentMapLight {
+            intensity: 0.0,
+            ..EnvironmentMapLight::solid_color(&mut images, Color::WHITE)
```
atlv24:
```suggestion
            ..default()
```

examples/3d/motion_blur.rs
```diff
@@ -30,6 +30,10 @@ fn setup_camera(mut commands: Commands) {
         // MSAA and Motion Blur together are not compatible on WebGL
         #[cfg(all(feature = "webgl2", target_arch = "wasm32", not(feature = "webgpu")))]
         Msaa::Off,
+        EnvironmentMapLight {
+            intensity: 300.0,
+            ..EnvironmentMapLight::solid_color(&mut images, Color::WHITE)
```
atlv24:
```suggestion
            ..default()
```

examples/3d/spherical_area_lights.rs
```diff
@@ -16,12 +12,17 @@ fn main() {
...
+    mut images: ResMut<Assets<Image>>,
     mut materials: ResMut<Assets<StandardMaterial>>,
 ) {
     // camera
     commands.spawn((
         Camera3d::default(),
         Transform::from_xyz(0.2, 1.5, 2.5).looking_at(Vec3::ZERO, Vec3::Y),
+        EnvironmentMapLight {
+            intensity: 60.0,
+            ..EnvironmentMapLight::solid_color(&mut images, Color::WHITE)
```
atlv24:
```suggestion
            ..default()
```

examples/3d/spotlight.rs
```diff
@@ -122,6 +119,10 @@ fn setup(
         Camera3d::default(),
         Hdr,
         Transform::from_xyz(-4.0, 5.0, 10.0).looking_at(Vec3::ZERO, Vec3::Y),
+        EnvironmentMapLight {
+            intensity: 20.0,
+            ..EnvironmentMapLight::solid_color(&mut images, Color::WHITE)
```
atlv24:
```suggestion
            ..default()
```

examples/3d/ssao.rs
```diff
@@ -33,6 +30,10 @@ fn setup(
         Msaa::Off,
         ScreenSpaceAmbientOcclusion::default(),
         TemporalAntiAliasing::default(),
+        EnvironmentMapLight {
+            intensity: 1000.0,
+            ..EnvironmentMapLight::solid_color(&mut images, Color::WHITE)
```
atlv24:
```suggestion
            ..default()
```

atlv24:
im just gonna make a PR to this

atlv24 (approved):

crates/bevy_light/src/lib.rs
```diff
@@ -193,6 +219,20 @@ impl Plugin for LightPlugin {
...
                 ),
             );
+
+        if self.default_environment_map_light {
+            app.world_mut()
+                .register_required_components::<Camera3d, EnvironmentMapLight>();
+        }
+
+        app.world_mut().resource_mut::<Assets<Image>>().insert(
+            &DEFAULT_ENVIRONMENT_MAP_TEXTURE_HANDLE,
```
mate-h, line 240: Should the default environment light be pure white like this? I think a better default would be a sky-like gradient, or maybe even the atmosphere env light. probably out of scope for this PR though.

IceSentry: I think for this PR should keep it white to keep it similar to the current default of ambient light, but I agree that in the future we should consider a nicer default

IceSentry (approved):

crates/bevy_light/src/lib.rs
atlv24: i agree, but thats for a different PR

janhohenheim (approved):
Resolves https://github.com/bevyengine/bevy/issues/12280 too
As you can see from that issue, I've struggled with the non-physicality of `AmbientLight` for a *long* time as a user. Happy to see it go!

crates/bevy_light/src/probe.rs
```diff
@@ -96,11 +100,79 @@ pub struct EnvironmentMapLight {
     pub affects_lightmapped_mesh_diffuse: bool,
 }
 
+impl EnvironmentMapLight {
+    /// An environment map with a uniform color, useful for uniform ambient lighting.
+    pub fn solid_color(assets: &mut Assets<Image>, color: Color) -> Self {
```
tychedelia, line 105: Nit: I don't have an obvious alternative pattern to use here but for some reason constructors that accept asset collections feels awkward to me. part of me feels like a `impl From<Color> for Image` would be better:

```rust
EnvironmentLightMap::from_image(images.add(Color::WHITE));
```

but that wouldn't work for the gradient version. Although maybe would if there was some kind of gradient type in `bevy_color`.

crates/bevy_light/src/probe.rs
```diff
@@ -96,11 +100,79 @@ pub struct EnvironmentMapLight {
     pub affects_lightmapped_mesh_diffuse: bool,
 }
 
+impl EnvironmentMapLight {
```
tychedelia, line 103: Can't comment on the struct definition above, but I think we need a `#[doc(alias = "ambient")]` or something similar to point users towards this in our docs.

crates/bevy_light/src/probe.rs
```diff
@@ -96,11 +100,79 @@ pub struct EnvironmentMapLight {
...
+    uuid_handle!("99e3f21e-9c08-4924-9895-fa8599416316");
+
 impl Default for EnvironmentMapLight {
     fn default() -> Self {
         EnvironmentMapLight {
-            diffuse_map: Handle::default(),
-            specular_map: Handle::default(),
+            diffuse_map: DEFAULT_ENVIRONMENT_MAP_TEXTURE_HANDLE,
+            specular_map: DEFAULT_ENVIRONMENT_MAP_TEXTURE_HANDLE,
             intensity: 0.0,
```
atlv24, line 176: this needs to be made non-zero

atlv24: nah

crates/bevy_light/src/lib.rs
```diff
@@ -193,6 +219,20 @@ impl Plugin for LightPlugin {
                         .after(clear_directional_light_cascades),
                 ),
             );
+
+        if self.default_environment_map_light {
+            app.world_mut()
+                .register_required_components::<Camera3d, EnvironmentMapLight>();
```
atlv24:
```suggestion
                .register_required_components_with::<Camera3d, EnvironmentMapLight>(|| {
                    EnvironmentMapLight {
                        intensity: 50.0,
                        ..Default::default()
                    }
                });
```

rparrett:
The example runner is showing several failing examples in the previous two runs.

`specular_tint`, `transmission`, `fog_volumes`, `auto_exposure`, `fog` fail in this PR and not on main

```
thread 'main' panicked at /Users/me/src/bevy/crates/bevy_app/src/plugin_group.rs:315:13:
bevy_light::LightPlugin does not exist in this PluginGroup
```

rparrett:
I'm not sure what's going on with `lightmaps` and `mixed_lighting`. edit: also `depth_of_field`

Those have been reliable on https://bevyengine.github.io/bevy-example-runner/ for a while, so it seems like either new intermittent behavior caused by this PR or another very recent one. Maybe a timing issue with the screenshots specific to this PR?

I can't reproduce locally.

atlv24:
it affected DOF too

tychedelia (commented):
So I'm generally down with the approach here but I'm wondering whether we actually need to deprecate and remove AmbientLight? Rather than just change the defaults.

My concern here is a little bit about performance, but also about non-pbr 3d scenes. While they're mostly fungible, I'm thinking about stylized rendering where environment light may not make sense, or if you wanted to animate the ambient light color where creating a new texture every frame feels a bit overkill. The later is particularly relevant for more creative use cases.

I'd be curious about other engines here. The engine I use at work supports both.

atlv24:
I think the right way forward here is to have an optional tint color on EnvMapLights that multiplies over the image. This is because AmbientLights are currently not PBR-accurate and maintaining a second code path for them with different weird units, and also the fact that they dont affect reflections, and conflictingly interact with envmaps, is weird and bad. A tint color on EnvMapLight would let one animate it and do simple changes like orange-tint a cloudy envmap to fake sunset type of thing, and is just generally a useful artistic control. Then a white envmap + tint would fill the same role as AmbientLight, but better.

I do believe AmbientLight should be yeeted.

atlv24:
Actually, on further thought, we "just" have to significantly lower the barrier to adding custom light-like shading terms. Id like to see a bevy api where the clustering etc was reusable and it was possible to have user-land implementations of directional lights, spotlights, point lights, etc. which used the pbr infrastructure and worked in both forward and deferred.

tychedelia:
> A tint color on EnvMapLight would let one animate it and do simple changes like orange-tint a cloudy envmap to fake sunset type of thing, and is just generally a useful artistic control. Then a white envmap + tint would fill the same role as AmbientLight, but better.

Okay, I'm relatively convinced by this.

I still don't like the constructors here, they feel unidiomatic. But I'll approve for now.

tychedelia (approved):

examples/3d/fog.rs
```diff
@@ -29,8 +29,10 @@ use bevy::{
 
 fn main() {
     App::new()
-        .insert_resource(AmbientLight::NONE)
-        .add_plugins(DefaultPlugins)
+        .add_plugins(DefaultPlugins.set(bevy::pbr::PbrPlugin {
```
cart, line 32: I dislike that we need to reach in to plugin config to override a "camera default". Semantically in most cases I think people should be saying "I dont want _this_ camera to have the default". We are composing _scenes_ here, and Plugin settings are not scene properties. This design choice was almost certainly a product of required component behavior (which you currently can't _disable_ for a specific spawn ... and I'm not convinced we ever want that). I would almost prefer `camera.insert(EnvironmentMapLight { intensity: 0, ..default()})` for the common case. That would mean we'd be rendering environment maps when we don't need them though. Making `EnvironmentMapLight` an enum probably makes sense. See my next comment for more motivation.

examples/3d/lighting.rs
```diff
@@ -247,6 +240,11 @@ fn setup(
         Camera3d::default(),
         Transform::from_xyz(-2.0, 2.5, 5.0).looking_at(Vec3::ZERO, Vec3::Y),
         Exposure::from_physical_camera(**parameters),
+        // environment lights' brightnesses are measured in candela per meter square, calculable as (color * intensity)
+        EnvironmentMapLight {
+            intensity: 80.0,
+            ..EnvironmentMapLight::solid_color(&mut images, ORANGE_RED.into())
```
cart, line 246: This unnecessarily regresses the UX of the (very) common case "I want colored ambient light". As others have mentioned, it also makes doing things like animating the color more challenging. 

In light of this (and my comment above, and the desired `EnvironmentLight` rename), I think we should make `EnvironmentLight` an enum:

```rust
enum EnvironmentLight {
    CubeMap {
        diffuse_map: Handle<Image>,
        intensity: f32,
        /* other fields */
    },
    Color {
        color: Color,
        intensity: f32,
    },
    Disabled,
}
```

I think this should be done before 0.17 to make this a "pleasant" change, ease migration, and avoid double breakage. We can then also remove all of the `bevy::pbr::PbrPlugin { default_environment_map_light: false }` nastiness in favor of `camera.insert(EnvironmentLight::Disabled)`

cart:
I'm cool with merging this now, provided we block 0.17 on the changes above.

atlv24:
punting this to 0.18, will write up an issue on the path forward

cart:
(We had a [Discord conversation here](https://discord.com/channels/691052431525675048/743663924229963868/1402746153128165539) with some context)

Igor-dvr:
Iiuc we should probably have environment/dome light type that encapsulates an enum: Color/AmbientLight, EnvironmetTexture, SkyTexture, something else. You can configure this light with only one option but can have multiple such lights with different options in a scene. Related:
- Blender PR Environment/Dome light type https://projects.blender.org/blender/blender/pulls/146193  and a comment about current Blender solution https://devtalk.blender.org/t/dome-envlight-instead-of-world-environment/13940
- OpenUSD Dome light type https://openusd.org/dev/user_guides/schemas/usdLux/DomeLight.html , 
https://openusd.org/dev/api/class_usd_lux_dome_light.html#details

crates/bevy_light/src/probe.rs
```diff
@@ -96,14 +100,82 @@ pub struct EnvironmentMapLight {
     pub affects_lightmapped_mesh_diffuse: bool,
 }
 
+impl EnvironmentMapLight {
+    /// An environment map with a uniform color, useful for uniform ambient lighting.
+    pub fn solid_color(assets: &mut Assets<Image>, color: Color) -> Self {
```
janhohenheim, line 105:
```suggestion
    pub fn solid_color(assets: &mut Assets<Image>, color: impl Into<Color>) -> Self {
```

so that e.g. `tailwind::AMBER_100` works :)

crates/bevy_light/src/probe.rs
```diff
@@ -96,14 +100,82 @@ pub struct EnvironmentMapLight {
...
+        Self::hemispherical_gradient(assets, color, color, color)
+    }
+
+    /// An environment map with a hemispherical gradient, fading between the sky and ground colors
+    /// at the horizon. Useful as a very simple 'sky'.
+    pub fn hemispherical_gradient(
+        assets: &mut Assets<Image>,
+        top_color: Color,
+        mid_color: Color,
+        bottom_color: Color,
```
janhohenheim, line 115: plz turn these into `impl Into<Color>` too

janhohenheim (changes requested):

may-bell:
allowing users to select the color of ambient light is an extremely common use case that can run on all hardware. ancient, battle-tested technology that runs on everything should not be abandoned.

beicause:
I'm also concerned that this would be bad for low-end devices.

After #23982 if you don't use `EnvironmentMapLight` you can save 2 texture bindings, which is important on web or low-end devices because default `max_sampled_textures_per_shader_stage` limit is only 16.

PR #7317, opened 2023-01-21
reflect: improve container attributes

soqb:
# Objective

- Remove confusion around `#[reflect(Debug, PartialEq, Hash, Default)]`.
- Allow specifying paths like `#[reflect(my_crate::foo::MyTrait)]`.

## Solution

- Remove the [special casing](https://github.com/bevyengine/bevy/blob/cb4e8c832c0ee33ea0b28a6f79971955643055fb/crates/bevy_reflect/bevy_reflect_derive/src/from_reflect.rs#L95) around `#[reflect(Default)]`.
- Change the syntax to `#[reflect(debug, partial_eq, hash)]` to make obvious the distinction between normal and "special" idents.
  - `#[reflect(Hash(my_hash_fn))]` is now `#[reflect(hash = "my_hash_fn")]`.
- Implement parsing for paths in `#[reflect]` attributes.

---

## Changelog

- Replaced `#[reflect(Debug, PartialEq, Hash, Default)]` syntax with `#[reflect(debug, partial_eq, hash)]` syntax.

## Migration Guide

- Replaced `#[reflect(Debug, PartialEq, Hash, Default)]` syntax with `#[reflect(debug, partial_eq, hash)]` syntax.
- Derived `FromReflect::from_reflect` implementations no longer can create a value from a partial set of fields if it implements default. All fields are required.

PR #18211, opened 2025-03-09
bevy_reflect: Add `ReflectClone` type data

MrGVSV:
# Objective

This PR is a partial followup to #13432, but is not based on that branch as it also works as a completely standalone change.

The goal of this PR is to add reflection type data for the `Clone` trait.

For #13432, this is desired because it addresses inconsistencies mentioned in this [comment](https://github.com/bevyengine/bevy/pull/7317#issuecomment-2143075852) (which are further addressed in #13723). Namely, we should try to reduce the split between special trait type data registration and regular trait type data registration.

Apart from that PR, this PR should allow users to create concrete clones within reflection much easier.

For reference, here are some ways cloning is typically handled:

```rust
// If the value is an `Opaque` type (e.g. primitives):
let clone: Box<dyn PartialReflect> = value.clone_value();

// If the value is any other type, we can use `ReflectFromReflect`:
let clone: Box<dyn PartialReflect> = {
  let info = value.get_represented_type_info().unwrap();
  let rfr = registry.get_type_data::<ReflectFromReflect>(info.type_id())).unwrap();
  rfr.from_reflect(&*value).unwrap()
};

// Or `ReflectDefault`:
let clone: Box<dyn PartialReflect> = {
  let info = value.get_represented_type_info().unwrap();
  let rd = registry.get_type_data::<ReflectDefault>(info.type_id())).unwrap();
  let default = rd.default();
  default.apply(&*value);
  default
};
```

Those solutions are also not ideal since some types do extra things when cloned (e.g. increasing a ref-count). It would be better if we could call that `Clone` impl directly via reflection.

## Solution

This PR simply introduces a `ReflectClone` type data struct that can be registered like any other type data (assuming the type implements `Clone`).

This simplifies our above approach:

```rust
let clone: Box<dyn PartialReflect> = {
  let info = value.get_represented_type_info().unwrap();
  let rc = registry.get_type_data::<ReflectClone>(info.type_id()).unwrap();
  rc.clone(&*value)
};
```

Note that with #13432 this is even simpler, but this provides a solid alternative if needing to use type data (or if wanting to specially handle types that implement `Clone`).

## Testing

You can test locally by running:

```
cargo test --doc "std_traits::ReflectClone::try_clone" --package bevy_reflect
```

---

## Showcase

Reflected types that implement `Clone` can now register the `ReflectClone` type data, allowing for reflected values to be directly cloned.

```rust
#[derive(Reflect, Clone)]
#[reflect(Clone)]
struct MyStruct(String);

let mut registry = TypeRegistry::new();
registry.register::<MyStruct>();

let value: Box<dyn Reflect> = Box::new(MyStruct(String::from("Hello!")));

let reflect_clone = registry.get_type_data::<ReflectClone>(TypeId::of::<MyStruct>()).unwrap();

let clone: Box<dyn Reflect> = reflect_clone.clone(&*value);

assert!(value.reflect_partial_eq(&*clone).unwrap());
```

### DIFF
--- a/crates/bevy_reflect/src/std_traits.rs
+++ b/crates/bevy_reflect/src/std_traits.rs
@@ -1,4 +1,4 @@
-use crate::{FromType, Reflect};
+use crate::{FromType, PartialReflect, Reflect};
 use alloc::boxed::Box;
 
 /// A struct used to provide the default value of a type.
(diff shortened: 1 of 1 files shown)


alice-i-cecile (approved):

MrGVSV:
Note that with #18307 this will probably need to be rebased since it will cause a lot of errors where `ReflectClone` has not been imported.

alice-i-cecile:
Cutting from the milestone, but please feel free to explain why this is vital / part of a larger whole.

alice-i-cecile:
Okay that makes sense to me :) We'll get this in eventually, but not critical to block on.

alice-i-cecile:
Okay that makes sense to me :) We'll get this in eventually, but not critical to block on.

alice-i-cecile:
Okay that makes sense to me :) We'll get this in eventually, but not critical to block on.

PR #18213, opened 2025-03-09
Return an error when direct-nested-loading a subasset

### LINKED ISSUE #18010: Need some way to nest labeled assets

andriyDev:
# Objective

- Prevents #18291.
- Previously, attempting to direct-nested-load a subasset would return the root of the nested-loaded asset. This is most problematic when doing direct-nested-**untyped**-loads of subassets, where you may not even realize you're dealing with the entirely wrong asset (at least with typed loads, *most of the time* the root asset has a different type from the subasset, and so at least you'd get an error that the types don't match).

## Solution

- We now return an error when doing these kinds of loads.

Note an alternative would be to "solve" this problem, by just looking up the appropriate subasset after doing the nested load. However there's two problems: 1) we don't know which subassets of the root asset are necessary for the subasset we are looking up (so any handles in that subasset may never get registered), 2) a solution will likely hamper attempts to resolve #18010. AFAICT, no one has complained about this issue, so it doesn't seem critical to fix this for now.

## Testing

- Added a test to ensure this returns an error. I also checked that the error before this was just a mismatched type error, meaning it was trying to pass off the root asset type `CoolText` as the subasset type `SubText` (which would have worked had I not been using typed loads).


alice-i-cecile (approved):
I really like pushing asset loading towards clear, error-returning behavior. It's way too hard to reason about anything else IMO.

andriyDev:
(forgot to rebase before pushing, sorry)

crates/bevy_asset/src/loader.rs
```diff
@@ -359,10 +359,14 @@ impl<A: Asset> From<CompleteLoadedAsset<A>> for CompleteErasedLoadedAsset {
 /// [`NestedLoader::load`]: crate::NestedLoader::load
 /// [immediately]: crate::Immediate
 #[derive(Error, Debug)]
-#[error("Failed to load dependency {dependency:?} {error}")]
-pub struct LoadDirectError {
-    pub dependency: AssetPath<'static>,
-    pub error: AssetLoadError,
+pub enum LoadDirectError {
+    #[error("Requested to load an asset path ({0:?}) with a subasset, but this is unsupported")]
```
kristoff3r, line 363: I think this error message, despite trying its best, will be confusing to people that aren't pretty familiar with bevy's asset system. But maybe that's the target audience for now anyway? Maybe we should link one of the related issue(s)?

kristoff3r (approved):

crates/bevy_asset/src/loader.rs
andriyDev: Updated to link to a new issue.

PR #18215, opened 2025-03-09
Provide a safe abstraction for split access to entities and commands

ItsDoot:
# Objective

Currently experimenting with manually implementing `Relationship`/`RelationshipTarget` to support associated edge data, which means I need to replace the default hook implementations provided by those traits. However, copying them over for editing revealed that `UnsafeWorldCell::get_raw_command_queue` is `pub(crate)`, and I would like to not have to clone the source collection, like the default impl. So instead, I've taken to providing a safe abstraction for being able to access entities and queue commands simultaneously.

## Solution

Added `World::entities_and_commands` and `DeferredWorld::entities_and_commands`, which can be used like so:

```rust
let eid: Entity = /* ... */;
let (mut fetcher, mut commands) = world.entities_and_commands();
let emut = fetcher.get_mut(eid).unwrap();
commands.entity(eid).despawn();
```

## Testing

- Added a new test for each of the added functions.

### DIFF
--- a/crates/bevy_ecs/src/relationship/mod.rs
+++ b/crates/bevy_ecs/src/relationship/mod.rs
@@ -207,29 +207,23 @@ pub trait RelationshipTarget: Component<Mutability = Mutable> + Sized {
     /// The `on_replace` component hook that maintains the [`Relationship`] / [`RelationshipTarget`] connection.
     // note: think of this as "on_drop"
     fn on_replace(mut world: DeferredWorld, HookContext { entity, caller, .. }: HookContext) {
-        // NOTE: this unsafe code is an optimization. We could make this safe, but it would require
-        // copying the RelationshipTarget collection
-        // SAFETY: This only reads the Self component and queues Remove commands
-        unsafe {
-            let world = world.as_unsafe_world_cell();
-            let relationship_target = world.get_entity(entity).unwrap().get::<Self>().unwrap();
-            let mut commands = world.get_raw_command_queue();
-            for source_entity in relationship_target.iter() {
-                if world.get_entity(source_entity).is_ok() {
-                    commands.push(
-                        entity_command::remove::<Self::Relationship>()
-                            .with_entity(source_entity)
-                            .handle_error_with(error_handler::silent()),
-                    );
-                } else {
-                    warn!(
-                        "{}Tried to despawn non-existent entity {}",
-                        caller
-                            .map(|location| format!("{location}: "))
-                            .unwrap_or_default(),
-                        source_entity
-                    );
-                }
+        let (entities, mut commands) = world.entities_and_commands();
+        let relationship_target = entities.get(entity).unwrap().get::<Self>().unwrap();
+        for source_entity in relationship_target.iter() {
+            if entities.get(source_entity).is_ok() {
+                commands.queue(
+                    entity_command::remove::<Self::Relationship>()
+                        .with_entity(source_entity)
+                        .handle_error_with(error_handler::silent()),
+                );
+            } else {
+                warn!(
+                    "{}Tried to despawn non-existent entity {}",
+                    caller
+                        .map(|location| format!("{location}: "))
+                        .unwrap_or_default(),
+                    source_entity
+                );
             }
         }
     }
@@ -238,29 +232,23 @@ pub trait RelationshipTarget: Component<Mutability = Mutable> + Sized {
     /// that entity is despawned.
     // note: think of this as "on_drop"
     fn on_despawn(mut world: DeferredWorld, HookContext { entity, caller, .. }: HookContext) {
-        // NOTE: this unsafe code is an optimization. We could make this safe, but it would require
-        // copying the RelationshipTarget collection
-        // SAFETY: This only reads the Self component and queues despawn commands
-        unsafe {
-            let world = world.as_unsafe_world_cell();
-            let relationship_target = world.get_entity(entity).unwrap().get::<Self>().unwrap();
-            let mut commands = world.get_raw_command_queue();
-            for source_entity in relationship_target.iter() {
-                if world.get_entity(source_entity).is_ok() {
-                    commands.push(
-                        entity_command::despawn()
-                            .with_entity(source_entity)
-                            .handle_error_with(error_handler::silent()),
-                    );
-                } else {
-                    warn!(
-                        "{}Tried to despawn non-existent entity {}",
-                        caller
-                            .map(|location| format!("{location}: "))
-                            .unwrap_or_default(),
-                        source_entity
-                    );
-                }
+        let (entities, mut commands) = world.entities_and_commands();
+        let relationship_target = entities.get(entity).unwrap().get::<Self>().unwrap();
+        for source_entity in relationship_target.iter() {
+            if entities.get(source_entity).is_ok() {
+                commands.queue(
+                    entity_command::despawn()
+                        .with_entity(source_entity)
+                        .handle_error_with(error_handler::silent()),
+                );
+            } else {
+                warn!(
+                    "{}Tried to despawn non-existent entity {}",
+                    caller
+                        .map(|location| format!("{location}: "))
+                        .unwrap_or_default(),
+                    source_entity
+                );
             }
         }
     }
--- a/crates/bevy_ecs/src/world/deferred_world.rs
+++ b/crates/bevy_ecs/src/world/deferred_world.rs
@@ -341,7 +341,7 @@ impl<'w> DeferredWorld<'w> {
         self.get_entity_mut(entities).unwrap()
     }
 
-    /// Provides split access to fetching entities and to queuing commands.
+    /// Simultaneously provides access to entity data and a command queue, which will be applied when the [`World`] is next flushed.
     pub fn entities_and_commands(&mut self) -> (EntityFetcher, Commands) {
         let cell = self.as_unsafe_world_cell();
         // SAFETY: `&mut self` gives mutable access to the entire world, and prevents simultaneous access.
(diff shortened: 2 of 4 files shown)


crates/bevy_ecs/src/world/deferred_world.rs
```diff
@@ -341,6 +341,21 @@ impl<'w> DeferredWorld<'w> {
         self.get_entity_mut(entities).unwrap()
     }
 
+    /// Provides split access to fetching entities and to queuing commands.
```
alice-i-cecile, line 344: I'd like a bit of explanation in this doc comments (ideally on a new line following a line break) about why this might ever by desirable.

crates/bevy_ecs/src/world/entity_fetch.rs
```diff
@@ -3,12 +3,91 @@ use core::mem::MaybeUninit;
...
 use crate::{
     entity::{hash_map::EntityHashMap, hash_set::EntityHashSet, Entity, EntityDoesNotExistError},
+    error::Result,
     world::{
         error::EntityMutableFetchError, unsafe_world_cell::UnsafeWorldCell, EntityMut, EntityRef,
         EntityWorldMut,
     },
 };
 
+/// Provides a safe interface for accessing *just* the entities in a [`World`].
```
alice-i-cecile, line 13:
```suggestion
/// Provides a safe interface for non-structural access to the entities in a [`World`].
///
/// This cannot add or remove components, or spawn or despawn entities,
/// making it relatively safe to access in concert with other ECS data.
/// This type can be constructed via [`World::entities_and_commands`],
/// or the equivalent method on [`DeferredWorld`].
```

crates/bevy_ecs/src/world/deferred_world.rs
alice-i-cecile, line 344:
```suggestion
    /// Simultaneously provides access to entity data and a command queue, which will be applied when dropped.
```

alice-i-cecile:
I like the core idea and implementation, but we can do better on docs :)

ItsDoot:
Should I change `EntityFetcher` to mirror `World`'s entity accessors?

```diff
impl EntityFetcher {
-   pub fn get<F: WorldEntityFetch>(&self, entities: F) -> Result<F::Ref<'_>, EntityDoesNotExistError>;
-   pub fn get_mut<F: WorldEntityFetch>(&mut self, entities: F) -> Result<F::DeferredMut<'_>, EntityDoesNotExistError>;
+   pub fn get_entity<F: WorldEntityFetch>(&self, entities: F) -> Result<F::Ref<'_>, EntityDoesNotExistError>;
+   pub fn get_entity_mut<F: WorldEntityFetch>(&mut self, entities: F) -> Result<F::DeferredMut<'_>, EntityDoesNotExistError>;
+   pub fn entity<F: WorldEntityFetch>(&self, entities: F) -> F::Ref<'_>;
+   pub fn entity_mut<F: WorldEntityFetch>(&mut self, entities: F) -> F::DeferredMut<'_>;
+   pub fn get<T: Component>(&self, entity: Entity) -> Option<&T>;
+   pub fn get_mut<T: Component>(&mut self, entity: Entity) -> Option<&mut T>;
}
```

crates/bevy_ecs/src/relationship/mod.rs
```diff
@@ -238,29 +232,23 @@ pub trait RelationshipTarget: Component<Mutability = Mutable> + Sized {
...
-                            .unwrap_or_default(),
-                        source_entity
-                    );
-                }
+        let (entities, mut commands) = world.entities_and_commands();
+        let relationship_target = entities.get(entity).unwrap().get::<Self>().unwrap();
+        for source_entity in relationship_target.iter() {
+            if entities.get(source_entity).is_ok() {
+                commands.queue(
+                    entity_command::despawn()
```
Carter0, line 240: What's the difference between despawning vs removing the relationship? This code is very similar to the above code

crates/bevy_ecs/src/world/deferred_world.rs
```diff
@@ -341,6 +341,53 @@ impl<'w> DeferredWorld<'w> {
...
+    /// # use bevy_ecs::{prelude::*, world::DeferredWorld};
+    /// #[derive(Component)]
+    /// struct Targets(Vec<Entity>);
+    /// #[derive(Component)]
+    /// struct TargetedBy(Entity);
+    ///
+    /// # let mut _world = World::new();
+    /// # let e1 = _world.spawn_empty().id();
+    /// # let e2 = _world.spawn_empty().id();
+    /// # let eid = _world.spawn(Targets(vec![e1, e2])).id();
```
Carter0, line 364: minorist of nits in the entire world, but can you can make this `e_id` lol. Like e underscore id haha.

crates/bevy_ecs/src/world/entity_fetch.rs
```diff
@@ -3,12 +3,98 @@ use core::mem::MaybeUninit;
...
+///
+/// This cannot add or remove components, or spawn or despawn entities,
+/// making it relatively safe to access in concert with other ECS data.
+/// This type can be constructed via [`World::entities_and_commands`],
+/// or [`DeferredWorld::entities_and_commands`].
+///
+/// [`World`]: crate::world::World
+/// [`World::entities_and_commands`]: crate::world::World::entities_and_commands
+/// [`DeferredWorld::entities_and_commands`]: crate::world::DeferredWorld::entities_and_commands
+pub struct EntityFetcher<'w> {
```
Carter0, line 23: I could be misremembering, but don't we already have a ton of entity like structs in our codebase? Not saying we don't need this one, but just something to consider.

crates/bevy_ecs/src/world/entity_fetch.rs
```diff
@@ -3,12 +3,98 @@ use core::mem::MaybeUninit;
...
 use crate::{
     entity::{hash_map::EntityHashMap, hash_set::EntityHashSet, Entity, EntityDoesNotExistError},
+    error::Result,
     world::{
         error::EntityMutableFetchError, unsafe_world_cell::UnsafeWorldCell, EntityMut, EntityRef,
         EntityWorldMut,
     },
 };
 
+/// Provides a safe interface for non-structural access to the entities in a [`World`].
```
Carter0, line 13: Reading the docs for [DeferredWorld](https://docs.rs/bevy/latest/bevy/ecs/world/struct.DeferredWorld.html), it seems like it only allows non-structural changes.  It also seems like EntityFetcher also allows only non structural changes. So wouldn't allowing this for just the world be unneeded?

ItsDoot: Disallowing structural changes is a matter of upholding Rust's safety guarantees. Handing out `EntityWorldMut` would break rust's mutable aliasing requirements because we would have 2 `&mut World` references.

Carter0 (approved):
Not an expert on this by any means, but the code checks out!

crates/bevy_ecs/src/world/entity_fetch.rs
ItsDoot: This isn't an entity-like struct, it's a struct for fetching entity-likes. Not sure what you mean.

alice-i-cecile:
> Should I change EntityFetcher to mirror World's entity accessors?

No: I think we should eventually move in this direction for `World` instead.

alice-i-cecile (approved):

crates/bevy_ecs/src/world/entity_fetch.rs
Carter0: I think what you are saying is that EntityFetcher needs to disallow mutable changes because otherwise the returned Commands would cause safety conflicts because one could theoretically use both the EntityFetcher and the Commands to alter the world, thereby causing memory issues. 

So using commands in a world instead of a deferred world would allow structural changes, but only with the commands. Is that right?

ItsDoot: No structural changes are allowed for the duration of the returned `EntityFetcher` and `Commands` pair for either `World` or `DeferredWorld`; commands are only applied the next time `world.flush()` is called.

Carter0: If structural changes are allowed for a deferred world with (`EntityFetcher`, `Commands`) , then does that break deferred world? Or are the deferred world docs wrong? 

Thanks for responding btw! I appreciate it

`
/// A [`World`] reference that disallows structural ECS changes.
/// This includes initializing resources, registering components or spawning entities.
pub struct DeferredWorld<'w> {
    // SAFETY: Implementors must not use this reference to make structural changes
    world: UnsafeWorldCell<'w>,
}
`

ItsDoot: No, because commands are not applied until after the `DeferredWorld` goes out of scope (`DeferredWorld` is just a wrapper around a `&mut World`). You can queue commands all you want which perform structural changes, they just don't get applied until later.

Carter0: Ahhh okay, so then it sounds like with a `World` you can do something like `e = world.spawn()` whereas with a `DeferredWorld` you would need to do something like `commands.spawn()` because you can't just mutate state in the `DeferredWorld`  directly, you have to go through the command queue.

chescock:
> > Should I change EntityFetcher to mirror World's entity accessors?
> 
> No: I think we should eventually move in this direction for `World` instead.

I agree that we shouldn't introduce the panicking versions, but I think calling the methods `get_entity` instead of `get` to match the naming on `World` would be valuable.  

I imagine a common refactoring will be surrounding some code that calls `world.get_entity` with a call to `entities_and_commands` so that you can use `commands` concurrently, and that refactoring will be easier if the method name is the same for both types.

chescock (approved):
This seems good!  

(I have a nagging feeling that there is duplication between `World` and `DeferredWorld` that we ought to be able to consolidate, and this is adding a third place with the same API. 
 Like, `EntityFetcher` ought to be able to access resources and run `Query`s.  I wonder if there is some way to make that smoother in the future, maybe something like `World: DerefMut<Target = DeferredWorld>` or `type DeferredWorld = World<Access = NonStructural>`.)

PR #16273, opened 2024-11-07
Fix typos, grammar, and spelling mistakes

BenjaminBrienen:
# Objective

Code and documentation should use words that exist and are spelled correctly.

Also, I noticed a markdownlint warning to do with the table at the bottom having left and right bars (inconsistent style) so I fixed the template file.

## Solution

I used the cSpell VS Code extension and literally opened every single `.rs` file to see if it popped up any `cSpell` "Problems". Most of them were added as "words" in `settings.json`. See here:

```json
"cSpell.words": [
    "0xffffu",
    "Aabb",
    "aabbs",
    "Aaltonen's",
    "aarch",
    "Abney",
    "accesskit",
    "acosf",
    "acosh",
    "acoshf",
    "adjwgt",
    "Adreno",
    "ahash",
    "allocvec",
    "animatable",
    "animgraph",
    "antialiasing",
    "anymap",
    "anytuple",
    "aortiz",
    "APPCOMMAND",
    "arctan",
    "AREATEX",
    "argb",
    "argh",
    "arity",
    "arrayvec",
    "artifacting",
    "asinf",
    "asinh",
    "asinhf",
    "astc",
    "atanf",
    "atanh",
    "atanhf",
    "atomicow",
    "autovectorization",
    "AZERTY",
    "backface",
    "Backfaces",
    "Backquote",
    "backshift",
    "barebones",
    "Barré",
    "barycentrics",
    "batchable",
    "batchables",
    "Bdot",
    "Belen",
    "bevyengine",
    "bevymark",
    "Bezier",
    "Bézier",
    "Beziers",
    "Bezold",
    "Bgra",
    "bilinearly",
    "bincode",
    "bindgen",
    "bindgroup",
    "bindgroups",
    "Bindless",
    "bindpose",
    "bindposes",
    "bitand",
    "bitangent",
    "bitcast",
    "bitfield",
    "Bitfields",
    "bitflags",
    "bitflip",
    "Bitmask",
    "bitsets",
    "bitshifted",
    "bitslice",
    "bitstream",
    "bitvec",
    "BLIT",
    "blobby",
    "bokeh",
    "Borderless",
    "BRDF",
    "Brisebois",
    "Brücke",
    "BSDF",
    "bsparse",
    "Bufferable",
    "bufread",
    "buildable",
    "BURLYWOOD",
    "bytemuck",
    "bytewise",
    "callables",
    "callstack",
    "canonicalize",
    "castable",
    "Castano",
    "Castaño",
    "Casteljau",
[...]

PR #18219, opened 2025-03-10
Fix non-crate typos

BenjaminBrienen:
# Objective

Correct spelling

## Solution

Fix typos, specifically ones that I found in folders other than /crates

## Testing

CI

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1835,7 +1835,7 @@ path = "examples/asset/multi_asset_sync.rs"
 doc-scrape-examples = true
 
 [package.metadata.example.multi_asset_sync]
-name = "Mult-asset synchronization"
+name = "Multi-asset synchronization"
 description = "Demonstrates how to wait for multiple assets to be loaded."
 category = "Assets"
 wasm = true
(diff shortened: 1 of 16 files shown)


bushrat011899 (approved):

examples/games/contributors.rs
```diff
@@ -246,7 +246,7 @@ fn gravity(time: Res<Time>, mut velocity_query: Query<&mut Velocity>) {
     }
 }
 
-/// Checks for collisions of contributor-birbs.
```
rparrett, line 249: :(

mnmaita:
I'd just add "birb/s" as an exception 😅

BenjaminBrienen:
> I'd just add "birb/s" as an exception 😅

It is a lot easier to fix the typo than to fix all the places where they are called "bird" such as the file name and other code.

rparrett:
> It is a lot easier to fix the typo than to fix all the places where they are called "bird" such as the file name and other code.

I don't want to spend a whole lot of energy debating this because it's silly, but this is not a typographical error, and bird is also a word so there would be no need to "fix" every instance of "bird."

> I'd just add "birb/s" as an exception 😅

There is nothing to add an exception to, because Bevy doesn't use cSpell.

FWIW there was also some resistance to this change in the PR that gave birth to this one: https://github.com/bevyengine/bevy/pull/16273#pullrequestreview-2440843191

alice-i-cecile (approved):

PR #18220, opened 2025-03-10
Sync up the Derive of DragEntry to match the other events

Peepo-Juice:
# Objective
Add `#[derive(Clone, PartialEq, Debug, Reflect)]` to DragEntry so it matches the other picking events.

## Solution
Copy/paste (RIP Larry Tesler)

## Testing
Just ran cargo check. I don't believe this should break anything because I did not remove any derives it had before.

---

### DIFF
--- a/crates/bevy_picking/src/events.rs
+++ b/crates/bevy_picking/src/events.rs
@@ -278,7 +278,7 @@ pub struct DragDrop {
 }
 
 /// Dragging state.
-#[derive(Debug, Clone)]
+#[derive(Clone, PartialEq, Debug, Reflect)]
 pub struct DragEntry {
     /// The position of the pointer at drag start.
     pub start_pos: Vec2,


crates/bevy_picking/src/events.rs
```diff
@@ -278,7 +278,7 @@ pub struct DragDrop {
 }
 
 /// Dragging state.
-#[derive(Debug, Clone)]
+#[derive(Clone, PartialEq, Debug, Reflect)]
 pub struct DragEntry {
```
jnhyatt, line 282: Hmmm I don't believe `DragEntry` is a pointer event. Doesn't mean we don't want these impls for it but I *am* curious what you're doing that wanted them

jnhyatt (approved):

Olle-Lukowski (approved):

PR #18233, opened 2025-03-10
Add `print_stdout` and `print_stderr` lints (#17446)

### LINKED ISSUE #17446: Exactly, I wonder if Bevy should actually prohibit `eprintln` and `println` since we always have access to `log` and/or `tracing`...
Exactly, I wonder if Bevy should actually prohibit `eprintln` and `println` since we always have access to `log` and/or `tracing`...

_Originally posted by @bushrat011899 in https://github.com/bevyengine/bevy/pull/17442#discussion_r1921650500_

sirius94:
# Objective

- Prevent usage of `println!`, `eprintln!` and the like because they require `std`
- Fixes #17446

## Solution

- Enable the `print_stdout` and `print_stderr` clippy lints
- Replace all `println!` and `eprintln!` occurrences with `log::*` where applicable or alternatively ignore the warnings

## Testing

- Run `cargo clippy --workspace` to ensure that there are no warnings relating to printing to `stdout` or `stderr`

### DIFF
--- a/crates/bevy_ecs/examples/change_detection.rs
+++ b/crates/bevy_ecs/examples/change_detection.rs
@@ -8,6 +8,7 @@
 
 #![expect(
     clippy::std_instead_of_core,
+    clippy::print_stdout,
     reason = "Examples should not follow this lint"
 )]
 
--- a/crates/bevy_ecs/examples/events.rs
+++ b/crates/bevy_ecs/examples/events.rs
@@ -1,6 +1,8 @@
 //! In this example a system sends a custom event with a 50/50 chance during any frame.
 //! If an event was send, it will be printed by the console in a receiving system.
 
+#![expect(clippy::print_stdout, reason = "Allowed in examples.")]
+
 use bevy_ecs::{event::EventRegistry, prelude::*};
 
 fn main() {
--- a/crates/bevy_ecs/examples/resources.rs
+++ b/crates/bevy_ecs/examples/resources.rs
@@ -3,6 +3,7 @@
 
 #![expect(
     clippy::std_instead_of_core,
+    clippy::print_stdout,
     reason = "Examples should not follow this lint"
 )]
 
--- a/crates/bevy_ecs/src/schedule/executor/multi_threaded.rs
+++ b/crates/bevy_ecs/src/schedule/executor/multi_threaded.rs
@@ -5,10 +5,9 @@ use bevy_utils::{default, syncunsafecell::SyncUnsafeCell};
 use concurrent_queue::ConcurrentQueue;
 use core::{any::Any, panic::AssertUnwindSafe};
 use fixedbitset::FixedBitSet;
-use std::{
-    eprintln,
-    sync::{Mutex, MutexGuard},
-};
+#[cfg(feature = "std")]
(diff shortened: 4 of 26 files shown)


crates/bevy_ecs/examples/change_detection.rs
```diff
@@ -12,6 +12,7 @@
 )]
 
 use bevy_ecs::prelude::*;
+use log::info;
```
alice-i-cecile, line 15: In examples we should be using bevy's re-exports of this.

alice-i-cecile (approved):
Really thorough work. Thank you!

sirius94:
I've missed some things. I'll change this back to 'ready for review' when all the automated checks pass.

bushrat011899 (approved):
Nice! Standardising the code-base (and explicitly annotating exceptions with `expect`) is always worth it, even ignoring the `no_std` benefits.

mockersf (changes requested):
you changed a few tests where they should continue to use `println`. loggers are often not set for tests

crates/bevy_ecs/examples/change_detection.rs
```diff
@@ -37,7 +38,7 @@ fn main() {
 
     // Simulate 10 frames in our world
     for iteration in 1..=10 {
-        println!("Simulating frame {iteration}/10");
+        info!("Simulating frame {iteration}/10");
```
mockersf, line 41: logger is also not set here so this should still be a println

crates/bevy_ecs/examples/events.rs
```diff
@@ -2,6 +2,7 @@
 //! If an event was send, it will be printed by the console in a receiving system.
 
 use bevy_ecs::{event::EventRegistry, prelude::*};
+use log::info;
```
mockersf, line 5: same here, this should keep using println

crates/bevy_ecs/examples/resources.rs
```diff
@@ -7,6 +7,7 @@
 )]
 
 use bevy_ecs::prelude::*;
+use log::info;
```
mockersf, line 10: same here, this should keep using println

crates/bevy_ecs/src/schedule/executor/multi_threaded.rs
```diff
@@ -283,7 +281,7 @@ impl<'scope, 'env: 'scope, 'sys> Context<'scope, 'env, 'sys> {
             .push(SystemResult { system_index })
             .unwrap_or_else(|error| unreachable!("{}", error));
         if let Err(payload) = res {
-            eprintln!("Encountered a panic in system `{}`!", &*system.name());
+            log::error!("Encountered a panic in system `{}`!", &*system.name());
```
mockersf, line 284: there's no guarantee a logger is set here and this error is too important to risk to hide, it should keep using eprintln

mockersf:
very few of the existing `println` should be replaced for this, they should be `expect`. most of them have already been checked

mockersf:
this is still changing too many println to logs. those changes should be reverted

alice-i-cecile (approved):

PR #18235, opened 2025-03-10
Backtrace: std and threadsafe bevy_error_panic_hook

### LINKED ISSUE #18231: `bevy_ecs` fails to compile if `std` feature enabled without `backtrace` feature due to panic hook
## Bevy version

Main branch: https://github.com/bevyengine/bevy#3b9e2e64

Bug appeared on CI for Linux/MacOS/Windows machines.

Example run: https://github.com/Bluefinger/bevy_rand/actions/runs/13768017477/job/38499328922?pr=44

## What you did

Created new PR targeting Bevy main branch crates, ran tests that do not enable `bevy_ecs` default features and specifically do not enable `backtrace` but enable `std`. The CI then pulled the latest version of `bevy` from github.

## What went wrong

Happened on a compilation error, where `bevy_error_panic_hook` errors because it is trying to reference a `static` that does not exist.

## Additional information

It appears that the [hook](https://github.com/bevyengine/bevy/blame/main/crates/bevy_ecs/src/error/bevy_error.rs#L139) is incorrectly feature flagged, and needs *both* `std` AND `backtrace` to be enabled. I'd suggest that the `backtrace` feature requires `std` feature and to have the hook feature flagged behind `backtrace` instead of `std`.

SpecificProtagonist:
# Objective

Make `bevy_error_panic_hook` threadsafe. As it relies on a global variable, it fails when multiple threads panic.

## Solution

Switch from a global variable for storing whether an error message was printed to a thread-local one.

`thread_local` is in `std`; the `backtrace` already relies on `std` APIs. It didn't depend on the `std` feature though, so I've added that. I've also put `bevy_error_panic_hook` behind the `backtrace` feature, since it relies on the thread local variable, which fixes #18231.

## Testing

The following now loops instead of crashing:

```rust
std::thread::scope(|s| {
    use bevy_ecs::error::*;

    #[derive(Debug)]
    struct E;
    impl std::fmt::Display for E {
        fn fmt(&self, _: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
            todo!()
        }
    }
    impl std::error::Error for E {}

    std::panic::set_hook(Box::new(bevy_error_panic_hook(|_| {
        unreachable!();
    })));
    for _ in 0..2 {
        s.spawn(|| {
            loop {
                let _ = std::panic::catch_unwind(|| {
                    panic!("{:?}", BevyError::from(E));
                });
            }
        });
    }
});
```

### DIFF
--- a/crates/bevy_ecs/Cargo.toml
+++ b/crates/bevy_ecs/Cargo.toml
@@ -37,7 +37,7 @@ reflect_functions = ["bevy_reflect", "bevy_reflect/functions"]
 configurable_error_handler = []
 
 ## Enables automatic backtrace capturing in BevyError
-backtrace = []
+backtrace = ["std"]
 
 # Debugging Features
 
--- a/crates/bevy_ecs/src/error/bevy_error.rs
+++ b/crates/bevy_ecs/src/error/bevy_error.rs
@@ -134,7 +134,7 @@ const FILTER_MESSAGE: &str = "note: Some \"noisy\" backtrace lines have been fil
 #[cfg(feature = "backtrace")]
 std::thread_local! {
     static SKIP_NORMAL_BACKTRACE: core::cell::Cell<bool> =
-        core::cell::Cell::new(false);
+        const { core::cell::Cell::new(false) };
 }
 
 /// When called, this will skip the currently configured panic hook when a [`BevyError`] backtrace has already been printed.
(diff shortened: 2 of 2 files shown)


cart (approved):

cart:
Good call!

mockersf (approved):

PR #18236, opened 2025-03-10
Let `FilteredEntity(Ref|Mut)` receive access when nested.

### LINKED ISSUE #14349: `FilteredEntityRef`/`FilteredEntityMut`'s access is empty when used in composite queries
## Bevy version

Both 0.14.0 and https://github.com/bevyengine/bevy/commit/af865e76a323fc87dd16157e7f4d7593d02c2b95

## What you did

```rs
#[derive(Component)]
struct Foo;

let mut world = World::new();

let foo_id = world.init_component::<Foo>();

world.spawn(Foo);

// With just `FilteredEntityRef` it works fine.
let mut query = QueryBuilder::<FilteredEntityRef>::new(&mut world)
    .ref_id(foo_id)
    .build();
let entity_ref = query.single(&world);
assert!(entity_ref.access().has_read(foo_id));
assert!(entity_ref.get::<Foo>().is_some());

// Inside any tuple it does not work, even 1-element tuples.
let mut query = QueryBuilder::<(FilteredEntityRef,)>::new(&mut world)
    .ref_id(foo_id)
    .build();
let (entity_ref,) = query.single(&world);
assert!(entity_ref.access().has_read(foo_id));
assert!(entity_ref.get::<Foo>().is_some());
```

## What went wrong

It seems that when `FilteredEntityRef`/`FilteredEntityMut` are nested inside a tuple the access is not properly set. Indeed [the implementation of `WorldQuery` for tuples](https://github.com/bevyengine/bevy/blob/af865e76a323fc87dd16157e7f4d7593d02c2b95/crates/bevy_ecs/src/query/world_query.rs#L146-L230) does not call the [`set_access`](https://github.com/bevyengine/bevy/blob/af865e76a323fc87dd16157e7f4d7593d02c2b95/crates/bevy_ecs/src/query/world_query.rs#L105) method used by `FilteredEntityRef`/`FilteredEntityMut`. However the fix doesn't seem as simple as just calling it, since may ignore any potential conflict between the `FilteredEntity*` and the other `QueryData`s in the query.

chescock:
# Objective

Let `FilteredEntityRef` and `FilteredEntityMut` receive access when nested inside tuples or `#[derive(QueryData)]` types.  Make sure to exclude any access that would conflict with other subqueries!  

Fixes #14349

## Solution

Replace `WorldQuery::set_access(state, access)` with a new method, `QueryData::provide_extra_access(state, access, available_access)`, that passes both the total available access and the currently used access.  This is called after `WorldQuery::update_component_access()`, so any access used by ordinary subqueries will be known.  `FilteredEntityRef` and `FilteredEntityMut` can use the combination to determine how much access they can safely take, while tuples can safely pass those parameters directly to their subqueries.  

This requires a new `Access::remove_conflicting_access()` method that can be used to remove any access that would conflict with existing access.  Implementing this method was easier by first factoring some common set manipulation code out of `Access::extend`.  I can extract that refactoring to a separate PR if desired.  

Have `FilteredEntity(Ref|Mut)` store `Access` instead of `FilteredAccess` because they do not need to keep track of the filter.  This was necessary in an early draft but no longer is.  I left it in because it's small and I'm touching that code anyway, but I can extract it to a separate PR if desired.

### DIFF
--- a/crates/bevy_ecs/src/query/access.rs
+++ b/crates/bevy_ecs/src/query/access.rs
@@ -430,75 +430,55 @@ impl<T: SparseSetIndex> Access<T> {
 
     /// Adds all access from `other`.
     pub fn extend(&mut self, other: &Access<T>) {
-        let component_read_and_writes_inverted =
-            self.component_read_and_writes_inverted || other.component_read_and_writes_inverted;
-        let component_writes_inverted =
-            self.component_writes_inverted || other.component_writes_inverted;
-
-        match (
-            self.component_read_and_writes_inverted,
+        invertible_union_with(
+            &mut self.component_read_and_writes,
+            &mut self.component_read_and_writes_inverted,
+            &other.component_read_and_writes,
             other.component_read_and_writes_inverted,
-        ) {
-            (true, true) => {
-                self.component_read_and_writes
-                    .intersect_with(&other.component_read_and_writes);
-            }
-            (true, false) => {
-                self.component_read_and_writes
-                    .difference_with(&other.component_read_and_writes);
-            }
-            (false, true) => {
-                // We have to grow here because the new bits are going to get flipped to 1.
-                self.component_read_and_writes.grow(
-                    self.component_read_and_writes
-                        .len()
-                        .max(other.component_read_and_writes.len()),
-                );
-                self.component_read_and_writes.toggle_range(..);
-                self.component_read_and_writes
-                    .intersect_with(&other.component_read_and_writes);
-            }
-            (false, false) => {
-                self.component_read_and_writes
-                    .union_with(&other.component_read_and_writes);
-            }
-        }
-
-        match (
-            self.component_writes_inverted,
+        );
+        invertible_union_with(
+            &mut self.component_writes,
+            &mut self.component_writes_inverted,
+            &other.component_writes,
             other.component_writes_inverted,
-        ) {
-            (true, true) => {
-                self.component_writes
-                    .intersect_with(&other.component_writes);
-            }
-            (true, false) => {
-                self.component_writes
-                    .difference_with(&other.component_writes);
-            }
-            (false, true) => {
-                // We have to grow here because the new bits are going to get flipped to 1.
-                self.component_writes.grow(
-                    self.component_writes
-                        .len()
-                        .max(other.component_writes.len()),
-                );
-                self.component_writes.toggle_range(..);
-                self.component_writes
-                    .intersect_with(&other.component_writes);
-            }
-            (false, false) => {
-                self.component_writes.union_with(&other.component_writes);
-            }
-        }
+        );
 
         self.reads_all_resources = self.reads_all_resources || other.reads_all_resources;
         self.writes_all_resources = self.writes_all_resources || other.writes_all_resources;
-        self.component_read_and_writes_inverted = component_read_and_writes_inverted;
-        self.component_writes_inverted = component_writes_inverted;
         self.resource_read_and_writes
             .union_with(&other.resource_read_and_writes);
         self.resource_writes.union_with(&other.resource_writes);
+    }
+
+    /// Removes any access from `self` that would conflict with `other`.
+    /// This removes any reads and writes for any component written by `other`,
+    /// and removes any writes for any component read by `other`.
+    pub fn remove_conflicting_access(&mut self, other: &Access<T>) {
+        invertible_difference_with(
+            &mut self.component_read_and_writes,
+            &mut self.component_read_and_writes_inverted,
+            &other.component_writes,
+            other.component_writes_inverted,
+        );
+        invertible_difference_with(
+            &mut self.component_writes,
+            &mut self.component_writes_inverted,
+            &other.component_read_and_writes,
+            other.component_read_and_writes_inverted,
+        );
+
+        if other.reads_all_resources {
+            self.writes_all_resources = false;
+            self.resource_writes.clear();
+        }
+        if other.writes_all_resources {
+            self.reads_all_resources = false;
+            self.resource_read_and_writes.clear();
+        }
+        self.resource_read_and_writes
+            .difference_with(&other.resource_writes);
+        self.resource_writes
+            .difference_with(&other.resource_read_and_writes);
     }
 
     /// Returns `true` if the access and `other` can be active at the same time,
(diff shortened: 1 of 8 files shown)


crates/bevy_ecs/src/query/state.rs
```diff
@@ -2202,9 +2211,9 @@ mod tests {
 
         let query_1 = QueryState::<&mut A>::new(&mut world);
         let query_2 = QueryState::<&mut B>::new(&mut world);
-        let mut new_query: QueryState<FilteredEntityMut> = query_1.join(&world, &query_2);
+        let mut new_query: QueryState<(Entity, FilteredEntityMut)> = query_1.join(&world, &query_2);
 
-        let mut entity = new_query.single_mut(&mut world).unwrap();
+        let (_entity, mut entity) = new_query.single_mut(&mut world).unwrap();
```
alice-i-cecile, line 2216: This variable naming is very confusing.

crates/bevy_ecs/src/query/fetch.rs
```diff
@@ -712,6 +722,18 @@ unsafe impl<'a> QueryData for FilteredEntityRef<'a> {
         item
     }
 
+    #[inline]
+    fn provide_extra_access(
+        state: &mut Self::State,
+        access: &mut Access<ComponentId>,
+        available_access: &Access<ComponentId>,
+    ) {
+        state.clone_from(available_access);
```
alice-i-cecile, line 731: Comment here please: this is pretty tricky to understand when skimming out of context.

crates/bevy_ecs/src/query/access.rs
```diff
@@ -838,6 +818,40 @@ impl<T: SparseSetIndex> Access<T> {
     }
 }
 
+/// Performs an in-place union of `other` into `self`, where either set may be inverted.
```
alice-i-cecile, line 821: I'd like more docs here, explaining what the right logic is. The union terminology is very precise, but doesn't help people understand what that means in practice.

crates/bevy_ecs/src/query/access.rs
```diff
@@ -838,6 +818,40 @@ impl<T: SparseSetIndex> Access<T> {
...
+            // We have to grow here because the new bits are going to get flipped to 1.
+            self_set.grow(other_set.len());
+            self_set.toggle_range(..);
+            self_set.intersect_with(other_set);
+        }
+        (false, false) => self_set.union_with(other_set),
+    }
+}
+
+/// Performs an in-place set difference of `other` from `self`, where either set may be inverted.
```
alice-i-cecile, line 842: Same request for more docs here.

alice-i-cecile (changes requested):
As far as I can tell this is correct, and mildly useful. I'd like to move forward with this, but this makes an already confusing bit of the code base more confusing. I've left suggestion for more docs, but I'd also like to see some unit tests for the invertable union / difference methods.

Those are pretty rough to mess up, and seem very testable!

chescock:
Added some more comments and tests!  

While I'm adding docs to the set manipulation methods: A way to make this even more clear would be to create a new type that wraps a `FixedBitSet` and its `inverted` flag.  I avoided that initially because it was a larger change than necessary, and because the layout change would make the `Access` struct a little bigger (from 128 bytes to 144).  But it should be a lot easier to understand than the separate `inverted` flags.  

Would it be worth doing that, either before, after, or as part of this PR?

alice-i-cecile:
I'd prefer to do that separately, but I do think it's a good idea. As long as we're not getting perf regressions I think some clarity there would be really helpful.

crates/bevy_ecs/src/query/fetch.rs
```diff
@@ -712,6 +722,25 @@ unsafe impl<'a> QueryData for FilteredEntityRef<'a> {
...
+    ) {
+        // Claim any extra access that doesn't conflict with other subqueries
+        // This is used when constructing a `QueryLens` or creating a query from a `QueryBuilder`
+        // Start with the entire available access, since that is the most we can possibly access
+        state.clone_from(available_access);
+        // Prevent all writes, since `FilteredEntityRef` only performs read access
+        state.clear_writes();
+        // Prevent any access that would conflict with other accesses in the current query
+        state.remove_conflicting_access(access);
+        // Finally, add the resulting access to the query access
```
cBournhonesque, line 739: This is interesting; because of how the macro is written, each term of a tuple inside a query like `Query<(&A, &B)>` sequentially updates the access by comparing with the access computed from the first `n` previous elements of the tuple?

crates/bevy_ecs/src/query/builder.rs
```diff
@@ -422,6 +423,89 @@ mod tests {
         }
     }
 
+    #[test]
+    fn builder_provide_access() {
```
cBournhonesque, line 427: greater tests

crates/bevy_ecs/src/query/state.rs
```diff
@@ -1979,12 +1988,12 @@ mod tests {
     fn can_transmute_filtered_entity() {
         let mut world = World::new();
         let entity = world.spawn((A(0), B(1))).id();
-        let query =
-            QueryState::<(Entity, &A, &B)>::new(&mut world).transmute::<FilteredEntityRef>(&world);
+        let query = QueryState::<(Entity, &A, &B)>::new(&mut world)
+            .transmute::<(Entity, FilteredEntityRef)>(&world);
```
cBournhonesque, line 1992: How come the tests needs an extra 'Entity' in the signature?

cBournhonesque (approved):
This is great; the current limitations of FilteredEntity(Ref|Mut) having to be the only QueryParam of a query was a bit artificial and confusing.

The unit tests are great

cBournhonesque (approved):
This is great; the current limitations of FilteredEntity(Ref|Mut) having to be the only QueryParam of a query was a bit artificial and confusing.

The unit tests are great

crates/bevy_ecs/src/query/builder.rs
```diff
@@ -422,6 +423,89 @@ mod tests {
...
+        let mut world = World::new();
+        world.spawn((A(0), B(1)));
+
+        let mut query =
+            QueryBuilder::<(Entity, FilteredEntityRef, FilteredEntityMut)>::new(&mut world)
+                .data::<&mut A>()
+                .data::<&B>()
+                .build();
+
+        // The `FilteredEntityRef` only has read access, so the `FilteredEntityMut` can have read access without conflicts
```
cBournhonesque, line 437: The access is provided from left to right, so if the tuple had been (FilteredEntityMut,  FilteredEntityRef), the Ref wouldn't have read access to the mutable component?

It is not super obvious that the order matters, maybe this should be documented slightly more in the FilteredEntityRef/Mut docstrings?

crates/bevy_ecs/src/query/fetch.rs
chescock: I think so?  It does two passes.  We first call `update_component_access` to determine the static access for things like `&A` and `&mut B`, and then call `provide_extra_access` to offer any leftover access.  So you can do either `(&mut A, FilteredEntityMut)` or `(FilteredEntityMut, &mut A)`, and the `&mut A` will claim the `A` access before the `FilteredEntityMut` sees it.  

But within `provide_extra_access`, it offers access to each term in order and looks at the access that previous terms took.  That causes the behavior you asked about below where `(FilteredEntityRef, FilteredEntityMut)` can be different from `(FilteredEntityMut, FilteredEntityRef)`.

crates/bevy_ecs/src/query/state.rs
chescock: That was just to make the tests start failing before I made the other changes :).

crates/bevy_ecs/src/query/builder.rs
chescock: > The access is provided from left to right, so if the tuple had been (FilteredEntityMut, FilteredEntityRef), the Ref wouldn't have read access to the mutable component?

Yes, that's right!  

> It is not super obvious that the order matters, maybe this should be documented slightly more in the FilteredEntityRef/Mut docstrings?

I wanted to ensure that it was *sound* to use more than one `FilteredEntity` thing in a query, but it's not really *useful*, and I don't expect anyone to actually want to do it.  So I'd be inclined to leave the exact semantics out of the documentation so as not to confuse normal users with the details of unimportant edge cases.

cBournhonesque: That's reasonable

alice-i-cecile (approved):
Awesome, thank you!

PR #18237, opened 2025-03-10
docs:  Fix incorrect cycle_modes function documentation

Emmet-v15:
Update the documentation comment to correctly state that the function responds to spacebar presses rather than mouse clicks. This aligns the documentation with the actual implementation, which was already using keyboard input.

# Objective

- Fix incorrect documentation comment for the cycle_modes function that inaccurately described using mouse input when the implementation actually uses spacebar.
- This ensures documentation accuracy and prevents confusion for new learners.

## Solution

- Updated the documentation comment to correctly state that the function responds to spacebar presses rather than mouse clicks.
- The code functionality remains unchanged.

### DIFF
--- a/examples/window/low_power.rs
+++ b/examples/window/low_power.rs
@@ -112,7 +112,7 @@ pub(crate) mod test_setup {
         window::RequestRedraw,
     };
 
-    /// Switch between update modes when the mouse is clicked.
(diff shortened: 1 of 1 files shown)


alice-i-cecile (approved):

alice-i-cecile:
Good fix, thanks for taking the time to make a PR for this :)

PR #18239, opened 2025-03-10
Upgrade to cosmic-text 0.13

rparrett:
# Objective

Upgrade to `cosmic-text` 0.13

https://github.com/pop-os/cosmic-text/releases

This should include some performance improvements for layout and system font loading.

## Solution

Bump version, fix the one changed API.

## Testing

Tested some examples locally, will invoke the example runner.

## Layout Perf

||main fps|cosmic-13 fps|
|-|-|-|
|many_buttons --recompute-text --no-borders|6.79|9.42 🟩 +38.7%|
|many_text2d --no-frustum-culling --recompute|3.19|4.28 🟩 +34.0%|
|many_glyphs --recompute-text|7.09|11.17 🟩 +57.6%|
|text_pipeline |140.15|139.90 ⬜ -0.2%|

## System Font Loading Perf

I tested on macOS somewhat lazily by adding the following system to the `system_fonts` example from #16365.

<details>
<summary>Expand code</summary>

```rust
fn exit_on_load(
    mut reader: EventReader<bevy::text::SystemFontsAvailable>,
    mut writer: EventWriter<AppExit>,
) {
    for _evt in reader.read() {
        writer.write(AppExit::Success);
    }
}
```
</details>

And running `hyperfine 'cargo run --release --example system_fonts --features=system_font'`.

The results were nearly identical with and without this PR cherry-picked there.


rparrett:
I think the minor differences can be attributed to swash `0.2.0`, but it's difficult to follow the changes to that library and its dependencies.

rparrett:
I did initiate a full example run just in case, still pending.

rparrett:
Not seeing anything concerning in there, just the same minor differences noticed by the testbed.

However, it seems that `cosmic-text` is failing to build for android.

mnmaita:
https://github.com/pop-os/cosmic-text/issues/349 looks like someone had the same issue with the previous version too?

rparrett:
Reverting https://github.com/pop-os/cosmic-text/commit/829a59bd9567b1ce1dd38c2fd527cf9ff838625b seems to fix it. I guess android targets used to take the "no-op" path in `other.rs` and that's not happening anymore.

I have a branch that works but I'm not 100% confident in. Will open a PR for discussion.

rparrett:
https://github.com/pop-os/cosmic-text/pull/365

alice-i-cecile:
Bumping from 0.17; we shouldn't wait on a fix IMO. This is not essential to ship immediately.

rparrett:
`cosmic-text` `0.13.2` was released which should fix the Android issue.

Carter0 (approved):
Thank you!

PR #18241, opened 2025-03-10
Expand the `RelationshipSourceCollection` to return more information

Brezak:
# Objective

While redoing #18058 I needed `RelationshipSourceCollection` (henceforth referred to as the **Trait**) to implement `clear` so I added it.

## Solution

Add the `clear` method to the **Trait**.
Also make `add` and `remove` report if they succeeded.

## Testing

Eyeballs

---

## Showcase

The `RelationshipSourceCollection` trait now reports if adding or removing an entity from it was successful.
It also not contains the `clear` method so you can easily clear the collection in generic contexts.

## Changes

EDITED by Alice: We should get this into 0.16, so no migration guide needed.

The `RelationshipSourceCollection` methods `add` and `remove` now need to return a boolean indicating if they were successful (adding a entity to a set that already contains it counts as failure). Additionally the `clear` method has been added to support clearing the collection in generic contexts.

### DIFF
--- a/crates/bevy_ecs/src/relationship/relationship_source_collection.rs
+++ b/crates/bevy_ecs/src/relationship/relationship_source_collection.rs
@@ -20,16 +20,26 @@ pub trait RelationshipSourceCollection {
     fn with_capacity(capacity: usize) -> Self;
 
     /// Adds the given `entity` to the collection.
-    fn add(&mut self, entity: Entity);
+    ///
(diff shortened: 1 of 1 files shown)


Brezak:
I've also considered adding the `shrink_to_fit` method to the **Trait** but I don't know how useful it would be.

alice-i-cecile (approved):

dmyyy (approved):

PR #18242, opened 2025-03-10
Incorporate viewport position into clamping in `camera_system`

mweatherley:
# Objective

In its existing form, the clamping that's done in `camera_system` doesn't work well when the `physical_position` of the associated viewport is nonzero. In such cases, it may produce invalid viewport rectangles (i.e. not lying inside the render target), which may result in crashes during the render pass. 

The goal of this PR is to eliminate this possibility by making the clamping behavior always result in a valid viewport rectangle when possible.

## Solution

Incorporate the `physical_position` information into the clamping behavior. In particular, always cut off enough so that it's contained in the render target rather than clamping it to the same dimensions as the target itself. In weirder situations, still try to produce a valid viewport rectangle to avoid crashes.

## Testing

Tested these changes on my work branch where I encountered the crash.

### DIFF
--- a/crates/bevy_render/src/camera/camera.rs
+++ b/crates/bevy_render/src/camera/camera.rs
@@ -75,6 +75,42 @@ impl Default for Viewport {
     }
 }
 
+impl Viewport {
+    /// Cut the viewport rectangle so that it lies inside a rectangle of the
+    /// given size.
+    ///
+    /// If either of the viewport's position coordinates lies outside the given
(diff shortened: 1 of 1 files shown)


tychedelia (approved):
I think this behavior makes sense and presumably users will be able to tell that their viewport is being clamped visually. Part of me feels like it should be a a more descriptive error and they should compute the clamp themselves. But this is probably less surprising overall.

JMS55 (approved):

PR #18246, opened 2025-03-10
Fix surface validation errors in WSL2 environments during window resizing

tyler274:
This is a known issue with Vulkan surface compatibility in WSL2 environments during window resizing operations.

## Solution

I've implemented a solution that gracefully handles these WSL2-specific surface errors:

1. Added a function to detect if the application is running in a WSL2 environment by checking OS-specific identifiers in `/proc/sys/kernel/osrelease`.
2. Created a helper function that safely captures and handles surface configuration panics in WSL2 environments, allowing applications to continue running instead of crashing.
3. Modified the surface configuration logic to use this safer approach when running in WSL.
4. Added specific error handling for WSL-related surface errors that provides informative warning messages.

The implementation is non-intrusive and only activates the special error handling in WSL environments, with no impact on other platforms.

## Testing

I've tested these changes by:
1. Running the test suite, which passes with all tests succeeding

I need to test: 
2. Testing window resizing in a WSL2 environment that previously crashed
3. Verifying that there's no impact on non-WSL environments

This PR doesn't introduce any breaking changes and only adds graceful error handling to improve the user experience in WSL2 environments.

## Showcase

Before this change, attempting to resize a window in WSL2 would cause the entire application to crash with the error message shown above. After this change, the application continues running with a warning message, and while there may be temporary visual glitches during resizing, the application remains stable and functional.

This is particularly helpful for developers who use WSL2 for their Bevy development workflow.

### DIFF
--- a/crates/bevy_render/src/view/window/mod.rs
+++ b/crates/bevy_render/src/view/window/mod.rs
@@ -225,7 +225,7 @@ fn safely_configure_surface(
 ) {
     if is_wsl {
         // Use a closure to avoid polluting the backtrace with catch_unwind machinery
-        let result = core::panic::catch_unwind(core::panic::AssertUnwindSafe(|| {
+        let result = core::intrinsics::catch_unwind(core::panic::AssertUnwindSafe(|| {
             render_device.configure_surface(surface, config);
         }));
 
--- a/crates/bevy_render/src/view/window/mod.rs
+++ b/crates/bevy_render/src/view/window/mod.rs
@@ -225,7 +225,7 @@ fn safely_configure_surface(
 ) {
     if is_wsl {
         // Use a closure to avoid polluting the backtrace with catch_unwind machinery
-        let result = std::panic::catch_unwind(std::panic::AssertUnwindSafe(|| {
+        let result = core::panic::catch_unwind(core::panic::AssertUnwindSafe(|| {
             render_device.configure_surface(surface, config);
         }));
 
--- a/crates/bevy_render/src/view/window/mod.rs
+++ b/crates/bevy_render/src/view/window/mod.rs
@@ -19,6 +19,9 @@ use wgpu::{
     SurfaceConfiguration, SurfaceTargetUnsafe, TextureFormat, TextureUsages, TextureViewDescriptor,
 };
 
+#[cfg(target_os = "linux")]
+use std::io::Read;
+
 pub mod screenshot;
 
 use screenshot::{ScreenshotPlugin, ScreenshotToScreenPipeline};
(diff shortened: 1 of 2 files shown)


aloucks:
Are you using llvmpipe or the experimental Dozen mesa d3d12 backend? Or something else?

When using llvmpipe, I'll get validation errors but it doesn't crash.
```
2025-03-11T22:49:08.908121Z  INFO bevy_diagnostic::system_information_diagnostics_plugin::internal: SystemInfo { os: "Linux (Ubuntu 22.04)", kernel: "5.15.167.4-microsoft-standard-WSL2", cpu: "Intel(R) Core(TM) i9-9900K CPU @ 3.60GHz", core_count: "8", memory: "15.5 GiB" }
2025-03-11T22:49:12.964722Z  INFO bevy_render::renderer: AdapterInfo { name: "llvmpipe (LLVM 15.0.7, 256 bits)", vendor: 65541, device: 0, device_type: Cpu, driver: "llvmpipe", driver_info: "Mesa 23.2.1-1ubuntu3.1~22.04.3 (LLVM 15.0.7)", backend: Vulkan }
```

If I use Dozen, everything just works. I don't even get validation errors when resizing. I had to compile mesa from scratch though to get this working. 

```
2025-03-11T22:51:33.064508Z  INFO bevy_diagnostic::system_information_diagnostics_plugin::internal: SystemInfo { os: "Linux (Ubuntu 22.04)", kernel: "5.15.167.4-microsoft-standard-WSL2", cpu: "Intel(R) Core(TM) i9-9900K CPU @ 3.60GHz", core_count: "8", memory: "15.5 GiB" }
2025-03-11T22:51:33.072363Z  WARN winit::platform_impl::linux::x11::xdisplay: error setting XSETTINGS; Xft options won't reload automatically
WARNING: dzn is not a conformant Vulkan implementation, testing use only.
WARNING: dzn is not a conformant Vulkan implementation, testing use only.
2025-03-11T22:51:36.078912Z  INFO bevy_render::renderer: AdapterInfo { name: "Microsoft Direct3D12 (NVIDIA GeForce RTX 2060)", vendor: 4318, device: 7944, device_type: DiscreteGpu, driver: "Dozen", driver_info: "Mesa 25.1.0-devel (git-551770ccf8)", backend: Vulkan }
```

I have to set the allow noncomplient adapter for Dozen. It doesn't report the API version correctly for me.

```
.add_plugins(DefaultPlugins.set(RenderPlugin {
    render_creation: RenderCreation::Automatic(WgpuSettings {
        instance_flags: InstanceFlags::ALLOW_UNDERLYING_NONCOMPLIANT_ADAPTER,
        ..default()
    }),
    ..default()
}))
```

tyler274:
Bevy reports llvmpipe and I didn't know Dozen existed

crates/bevy_render/src/view/window/mod.rs
```diff
@@ -193,6 +196,61 @@ impl WindowSurfaces {
     }
 }
 
+/// Determines if the application is running in a Windows Subsystem for Linux (WSL) environment
+#[cfg(target_os = "linux")]
```
krunchington, line 200: Suggestion: you may want to put this behind a feature flag which is disabled by default for non dev environments.  Seems like a small amount of stuff that may not be necessary in production builds, unless someone very specifically wants to compile their binary in a way to be run inside a wsl environment?

IceSentry:
Why not catch errors on all platforms? Also, what happens when the configure surface fails? AFAIK it should still panic because we can't keep going with an invalid surface.

rzhb:
I got WSL GPU rendering basically working, except bevy_pbr. https://github.com/bevyengine/bevy/discussions/22851#discussioncomment-17675996

After above, an empty window works.  
However, resizing or maximizing the window causes a crash unless a `Camera2d` is spawned. This may be related to this issue.

PR #18247, opened 2025-03-11
Generalize `DiGraph`/`UnGraph` 📊

ItsDoot:
# Objective

As the first step towards more modular schedule graph structures, we first need to generalize the foundational pieces: the `Graph` data structure and related algorithms.

One might ask: "what are 'more modular schedule graph structures'?" I've come across quite a few use cases that can *almost* be fulfilled with a `Schedule`, but have different-enough requirements that render `Schedule` incompatible as-is:
- My WIP utility AI library needs a graph of read-only systems while linking system output values as inputs to other systems.
- `bevy_render`'s `RenderGraph` is already somewhat `Schedule`-like, but requires all nodes to be read-only and currently only allows a single query to be performed.
- `bevy_render`'s `Extract` schedule could be statically enforced to be read-only, which would allow its executor to skip access conflict checks.
- Occasionally users in the Discord will show interest in statically-dispatched schedules, where a fork-join sequence can be modeled at compile time.

Therefore, by providing modular and generic building blocks for building `Schedule`-shaped things, we can satisfy each of the above and more. And since most of these are likely to make use of `DiGraph` or `UnGraph`, generalizing them is a good start.

One might ask: "what does this give us over `petgraph`?" A couple things:
- `petgraph` is not currently `no_std`, which is a prerequisite to pull it in now that `bevy_ecs` is `no_std`.
    - https://github.com/petgraph/petgraph/pull/747 has been recently opened to fix this.
- The associated types on `GraphNodeId` allows us to use memory-efficient forms for edge and neighbor information, which currently saves us 8 bytes for each edge pair and each neighbor+direction.

## Solution

Replaces the hardcoded usage of `NodeId` with an implementable `GraphNodeId` trait in `DiGraph`/`UnGraph` and associated algorithms. Algorithms implemented as part of `ScheduleGraph` will be tackled in a future PR in order to keep this one small.

## Testing

Current tests are being reused.

---

## Migration Guide

- Bare `DiGraph` or `UnGraph` usage must be changed to `DiGraph<NodeId>` or `UnGraph<NodeId>`, respectively.

### DIFF
--- a/crates/bevy_ecs/src/schedule/graph/node.rs
+++ b/crates/bevy_ecs/src/schedule/graph/node.rs
@@ -1,4 +1,9 @@
-use core::fmt::Debug;
+use core::{
+    fmt::{self, Debug},
+    hash::Hash,
+};
+
+use crate::schedule::graph::Direction;
 
 /// Unique identifier for a system or system set stored in a [`ScheduleGraph`].
 ///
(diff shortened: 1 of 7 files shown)


alice-i-cecile:
Can you say more about why this will be useful for your long-term plans? IIRC we did the opposite thing when vendoring petgraph 😂

crates/bevy_ecs/src/schedule/graph/node.rs
```diff
@@ -45,3 +50,131 @@ impl NodeId {
...
+}
+
+/// A node in a [`DiGraph`] or [`UnGraph`].
+///
+/// [`DiGraph`]: crate::schedule::graph::DiGraph
+/// [`UnGraph`]: crate::schedule::graph::UnGraph
+pub trait GraphNodeId: Copy + Eq + Ord + Debug + Hash {
+    /// A pair of [`GraphNodeId`]s for storing edge information. Typically
+    /// stored in a memory-efficient format.
+    type Pair: GraphNodeIdPair<Self>;
```
ItsDoot, line 136: Thoughts on `type Edge: GraphNodeEdge<Self>`; instead?

crates/bevy_ecs/src/schedule/graph/node.rs
```diff
@@ -45,3 +50,131 @@ impl NodeId {
...
+    /// A pair of [`GraphNodeId`] and [`Direction`] for storing neighbor
+    /// information. Typically stored in a memory-efficient format.
+    type Directed: DirectedGraphNodeId<Self>;
```
ItsDoot, line 139: Thoughts on `type Neighbor: GraphNodeNeighbor<Self>;` instead?

ItsDoot:
> Can you say more about why this will be useful for your long-term plans? IIRC we did the opposite thing when vendoring petgraph 😂

Context added.

PR #18248, opened 2025-03-11
make various entity wrapper type modules public

Victoronz:
# Objective

Part of the #16547 series.

The entity wrapper types often have some associated types an aliases with them that cannot be re-exported into an outer module together.
Some helper types are best used with part of their path: `bevy::ecs::entity::index_set::Slice` as `index_set::Slice`.
This has already been done for `entity::hash_set` and `entity::hash_map`.

## Solution

Publicize the `index_set`, `index_map`, `unique_vec`, `unique_slice`, and `unique_array` modules.

## Migration Guide

Any mention or import of types in the affected modules have to add the respective module name to the import path.
F.e.:
`bevy::ecs::entity::EntityIndexSet` -> `bevy::ecs::entity::index_set::EntityIndexSet`

### DIFF
--- a/crates/bevy_ecs/src/entity/entity_set.rs
+++ b/crates/bevy_ecs/src/entity/entity_set.rs
@@ -13,7 +13,7 @@ use core::{
     option, result,
 };
 
-use super::{Entity, UniqueEntitySlice};
+use super::{unique_slice::UniqueEntitySlice, Entity};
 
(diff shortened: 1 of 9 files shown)


bushrat011899 (approved):
Makes sense!

alice-i-cecile (approved):

PR #18258, opened 2025-03-11
public function activeanimation update

### LINKED ISSUE #18197: why ActiveAnimation update function didn't public?
We don't need clone ActiveAnimation struct repackage to myself's struct

A description of why this particular feature should be added.
If update function public,we can easy to do once animation or myself's animation function

Let update function public, and add a system update all ActiveAnimation.
But I not sure what question happend after public function.

Actually,this change will make ActiveAnimation didn't only work by BevyAnimationSystem but also allow we do complex system whatevr is not animation

Other solutions to solve and/or work around the problem presented.
If ActiveAnimation update function cause unsafe or another din't able public, We can repackage ActiveAnimation for update function still private.

## Additional context
I make a easy OnceAnimationSystem,it will work one or more I can't sure.
If animation into resource,it will be waster memory.

MushineLament:
# Objective
Fixes #18197 

## Solution
 ActiveAnimation struct add 'clip_duration : Option<f32>',in order it update function only by time.
 if 'clip_duration : Option<f32>' is None type,it will update function din't work.
 ActiveAnimation impl new function 'set_clip_duration',to setup clip_duration.

## Testing

- not,I have not't test anything.
- no,this just little fix.I am not find anything could read AnimationPlayer  to get ActiveAnimation.
- think about ActiveAnimation update function public will make what unsafe,I am think not.
- find this change update didn't work or to be bad.
- 
---

## Showcase
about ActiveAnimation struct:

```
 pub struct ActiveAnimation {
+   /// `None` is mean animation was not work.
+   clip_duration : Option<f32>
      ...
 }
```

about impl ActiveAnimation:
add:
```
    pub fn set_clip_duration(&mut self,clip_duration: f32)->&mut self{
        self.clip_duration = clip_duration;
    }
```

change update function:
```
    pub fn update(&mut self, delta: f32, ) {
        if self.clip_duration.is_none(){
            return;
        }
    ...
}
```
about pub fn advance_animations:
```
pub fn advance_animations(
...
){
...
                if let Some(active_animation) = active_animations.get_mut(&node_index) {
                    // Tick the animation if necessary.
                    if !active_animation.paused {
                        if let AnimationNodeType::Clip(ref clip_handle) = node.node_type {
                            if let Some(clip) = animation_clips.get(clip_handle) {
+                              active_animation.set_clip_duration(clip.duration);
                                active_animation.update(delta_seconds);
                            }
                        }
                    }
                }
...
}
```

## Other opinion
maybe this change will name 'ActiveAnimation' isn't suitable,because is like in the curve point get a sample by animation.


crates/bevy_animation/src/lib.rs
```diff
@@ -482,6 +482,8 @@ pub struct ActiveAnimation {
     completions: u32,
     /// `true` if the animation was completed at least once this tick.
     just_completed: bool,
+    /// `None` is mean animation was never completed forever.
```
alice-i-cecile, line 485:
```suggestion
    /// `None`  means the animation was will never complete.
```

I'm not sure if this is correct, but I couldn't understand your original comment.

crates/bevy_animation/src/lib.rs
```diff
@@ -561,6 +567,12 @@ impl ActiveAnimation {
         self.weight
     }
 
+    /// this will the duration of the clip being played
```
alice-i-cecile, line 570:
```suggestion
    /// Sets the duration of the clip being played
```

alice-i-cecile:
I'm struggling a bit to understand the motivation here, and the docs could be clearer. We also need to test this before we can merge.

MushineLament:
> 我真的不明白这要完成什么，而且我也对在没有测试😅的情况下更改动画内部持谨慎态度

Is sorry my expressive ability,this is my first pull requests.
so I don't konw how best testing it,but I am have try to find what will this bad result.I didn't find other use update function and look like nobody could get ActiveAnimation by AnimationPlayer.
this function public,could to do more better once animation.my game project is copy ActiveAnimation,and is work pettry good .

MushineLament:
I added an instance to achieve custom curve animation using the public update function.
It can also create one-time animations for complex components.
(computer translation)

crates/bevy_animation/src/lib.rs
```diff
@@ -482,6 +482,8 @@ pub struct ActiveAnimation {
     completions: u32,
     /// `true` if the animation was completed at least once this tick.
     just_completed: bool,
+    /// `None`  means the animation was will never complete.
```
lomirus, line 485: "was will"?

lomirus, line 485:
```suggestion
    /// `None` means the animation will never complete.
```

lomirus (changes requested):

examples/animation/animated_curve.rs
```diff
@@ -0,0 +1,265 @@
...
+        )
+        .add_systems(PostUpdate, myanimation_system)
+        .run();
+}
+
+///update text
+#[derive(Component)]
+pub struct AnimationControlTextMark;
+
+/// Provide numerical values for animation.(computer translation)
```
lomirus, line 26: `computer translation` should not be kept in code.

examples/animation/animated_curve.rs
```diff
@@ -0,0 +1,265 @@
...
+            speed: 100.0,
+            distance: 1.0,
+        }
+    }
+}
+
+#[derive(Component)]
+struct MyAnimationCurve {
+    active: ActiveAnimation,
+    // actually,I don't want use dyn
```
lomirus, line 45: It is not recommended to begin sentences with the first person in the code as it's too subjective.

examples/animation/animated_curve.rs
```diff
@@ -0,0 +1,265 @@
...
+                pickup_sprite.run_if(input_just_pressed(MouseButton::Left)),
+                drop_sprite.run_if(input_just_pressed(MouseButton::Right)),
+                update_text.run_if(resource_changed::<AnimationControl>),
+            ),
+        )
+        .add_systems(PostUpdate, myanimation_system)
+        .run();
+}
+
+///update text
```
lomirus, line 22: Doc comment should start with a capital letter, and has a space between it and `///`.

lomirus (changes requested):

lomirus:
I did not review the specific code logic. Just pointed out a few grammar and format issues.

mcobzarenco:
In addition to the poor English in the doc strings that others noted (it'd also be nice to update the PR title to e.g. "Expose ActiveAnimation::update` as that's the commit message), I don't understand the motivation for this PR, it would be cool to have a sentence saying what this let's you achieve.

MushineLament:
After re studying some of the code, I realized that my behavior was foolish and hasty. Such functional improvements can cause many complex problems, but the scalability obtained is minimal. Therefore, I believe that such functionality should be left to users to copy and improve the relevant code or implement it themselves. I apologize for wasting some maintainers' time. I will close this request.

alice-i-cecile:
No worries, we appreciate your explorations here and contributions in general!

PR #18259, opened 2025-03-11
Implement SpawnableList for Vec<Bundle>

krunchington:
# Objective

In updating examples to use the Improved Spawning API I got tripped up on being able to spawn children with a Vec.  I eventually figured out that I could use `Children::spawn(SpawnIter(my_vec.into_iter()))` but thought there might be a more ergonomic way to approach it.  After tinkering with it for a while I came up with the implementation here.  It allows authors to use `Children::spawn(my_vec)` as an equivalent implementation.

## Solution

- Implements `<R: Relationship, B: Bundle SpawnableList<R> for Vec<B>`
- uses `alloc::vec::Vec` for compatibility with `no_std` (`std::Vec` also inherits implementations against the `alloc::vec::Vec` because std is a re-export of the alloc struct, thanks @bushrat011899 for the info on this!)

## Testing

  - see above

## Showcase

n/a

## Migration Guide

- Optional: you may use the new API to spawn `Vec`s of `Bundle` instead of using the `SpawnIter` approach.


crates/bevy_ecs/src/spawn.rs
```diff
@@ -7,6 +7,7 @@ use crate::{
     relationship::{RelatedSpawner, Relationship, RelationshipTarget},
     world::{EntityWorldMut, World},
 };
+use alloc::vec::Vec;
```
krunchington, line 10: Using this for `no_std` support

crates/bevy_ecs/src/spawn.rs
```diff
@@ -45,6 +46,18 @@ pub trait SpawnableList<R> {
     fn size_hint(&self) -> usize;
 }
 
+impl<R: Relationship, B: Bundle> SpawnableList<R> for Vec<B> {
+    fn spawn(self, world: &mut World, entity: Entity) {
+        for bundle in self {
```
alice-i-cecile, line 51: This should use spawn_batch: it should be substantially faster.

alice-i-cecile (changes requested):
Oh that's very nice. I like the net effect and this is a sensible extension.  spawn_batch should be a straightforward improvement though, so I'd like to swap to that before merging.

crates/bevy_ecs/src/spawn.rs
krunchington: Great suggestion!  I had to add a bundleeffect constraint to get that to work- I'm very unfamiliar with what that affects but welcome feedback for improvement

alice-i-cecile (approved):

PR #18260, opened 2025-03-11
Implement `RelationshipSourceCollection` for `Option<Entity>`

Brezak:
# Objective

`Option` is technically a collection. 

## Solution

Implement `RelationShipSourceCollection` for `Option<Entity>`

## Testing
Add

---

## Showcase

<details>

```rust
```

</details>

## Migration Guide


ItsDoot:
#18087 Decided to go with `Entity` instead of `Option<Entity>` for one-to-one relationships, I don't think we need to have two ways to do the same thing.

Brezak:
I've mostly decided to implement `RelationshipSourceCollection` for every collection that can take it and `Option` was one of them.

Looking at #18087. I'm mainly concerned about the inevitable situation of a user spawning a entity with the `Entity::PLACEHOLDER` id and throwing it into a relationship. The `RelationShipTarget` will end up reporting that it's empty even if it isn't and I don't know what bugs that will cause. The rarity of this situation is going to turn these bugs into the worst kind of heisenbugs. Ones that require the application to be running for an hour before they manifest.

alice-i-cecile:
> Looking at https://github.com/bevyengine/bevy/pull/18087. I'm mainly concerned about the inevitable situation of a user spawning a entity with the Entity::PLACEHOLDER id and throwing it into a relationship

We're restricting the ability to spawn entities with specific IDs and that ID will never be handed out otherwise.

That said, I do sympathize with the desire to implement this trait for every sensible collection type 🤔

Issue #17114, opened 2025-01-03
Character Attachments


viridia:
## What problem does this solve or what need does it fill?

It's very common in games to want to assemble animated characters out of separate parts. This includes things like clothing and armor styles, weapons, hair and jewelry, and so on. Often the parts will be multiple skins which all share the same armature, however in the case of weapons these might be separate meshes which don't have an armature of their own but which are simply attached to a "weapon bone".

A complete solution involves not only the game engine but the authoring process as well: characters need to be exported in a way that allows individual skins to be assembled dynamically at runtime.

While it is possible to accomplish this in Bevy, the process is not easy. Note: I have looked at the "modular characters" video, but honestly most users do not want to sit through a three-hour video tutorial, especially given that the resulting code is both complex and depends on internal details of skeleton structure that might break in future Bevy versions.

## What solution would you like?

A complete solution would have the following requirements:

* First, there would be documentation about the process for creating compatible .glb files.
* If there's any preprocessing needed for the .glb files (such as splitting attachments into separate scenes) tools should be provided for this. (I have written tools like this in JavaScript using gltf-transform).
* There would be an API that allows spawning a character, starting with an armature asset path. The API would provide a way to add skins and attachments to the armature. These would be specified as asset paths.
* It should work regardless of whether the skins and the armature are in the same GLTF file or in different files (GLTF allows this).
* The solution should avoid spawning multiple copies of the armature for a single character.
* If the various attachments are stored together in a single GLTF file, the solution should avoid loading the file multiple times. In fact, it's probably best if there's a way to hold on to the root asset in the asset cache so that multiple characters can be spawned without having to re-parse the GLTF file.
  * I bring this up because currently when you load a labeled asset, the labeled asset is cached but it does not retain a strong handle to the root asset. This means that if you later load a different labeled asset from the same file, it will re-load and re-parse the asset from scratch.
* The solution should allow post-processing of the character materials, such as color remapping, morphing and so on. While these kinds of things are out of scope, it should be possible to write an observer that traverses the entity hierarchy and makes the appropriate substitutions.
* Once the character and their attachments are ready, we can animate the character by assigning animations to the armature in the standard way.

## What alternative(s) have you considered?

Well, the two alternatives I can think of are (a) rolling my own implementation and (b) relying on a third-party extension to Bevy. I don't actually know of a library which does this, although it's possibly I am simply lacking the correct search term. But given how many games need this kind of functionality, it seems like something that should be built in.

[...]

PR #18262, opened 2025-03-11
:test_tube: Bone attachments

hukasu:
# Objective

Starts work on #17114

## Solution

Create a new crate `bevy_bone_attachments` and implement a extension trait on `EntityCommands` to allow attaching a new `Scene` to an `Entity`. The command then collects all `AnimationTarget`s of the entity and copies it to the new `Scene`.

This is very early and only includes the above method, the new crate can also be included on an existing crate.

## Testing

* New example

## Showcase

Bone attachments allow you to attach a mesh to another, think a character holding a weapon, or customizable clothing.

Experimental: The current state only allows attaching new scenes into models, and requires that the parent entity is already fully loaded (if it is also a scene, after its `SceneInstance` is ready) and have matching `AnimationTargetId`s. 

[Screencast_20250312_094221.webm](https://github.com/user-attachments/assets/b4e9867a-0ad8-44ce-8997-615bcbddc0cc)

## Use Guide

```rust
use bevy_bone_attachments::scene::SceneAttachmentExt;

// This requires that the parent is already fully loaded, and there is only a warning if it is not
commands
    .entity(parent)
    .attach_scene(
        asset_server.load_with_settings(
            GltfAssetLabel::Scene(0).from_asset("attachment.glb"),
            |settings: &mut GltfLoaderSettings| { settings.include_animation_target_ids = true; }
    ));
```


hukasu:
@alice-i-cecile any template on how to make the release note?

alice-i-cecile:
> @alice-i-cecile any template on how to make the release note?

We're kind of in flux right now, but for now, putting a `# Showcase` section in the PR would be really helpful. The key points to hit are:

- what new feature was introduced?
- why is it important?
- very briefly, how do you use it?

examples/README.md
```diff
@@ -209,6 +209,7 @@ Example | Description
 [Animation Events](../examples/animation/animation_events.rs) | Demonstrate how to use animation events
 [Animation Graph](../examples/animation/animation_graph.rs) | Blends multiple animations together with a graph
 [Animation Masks](../examples/animation/animation_masks.rs) | Demonstrates animation masks
+[Bone Attachments](../examples/animation/bone_attachments.rs) | Shows attachting a model to another
```
viridia, line 212:
```suggestion
[Bone Attachments](../examples/animation/bone_attachments.rs) | Shows attaching a model to another
```

viridia (commented):
Looks good generally. Note to @alice-i-cecile : this PR solves one of the two use cases mentioned in the original issue, so unless we want to split it into two separate tickets, it should not close the issue once merged.

hukasu:
> Looks good generally. Note to @alice-i-cecile : this PR solves one of the two use cases mentioned in the original issue, so unless we want to split it into two separate tickets, it should not close the issue once merged.

that is why i wrote `Starts work on #17114` instead of `Closes`

examples/README.md
hukasu: how do i add something to `typos`?

hukasu: https://github.com/crate-ci/typos/issues/1248#issuecomment-2715818672

hukasu:
@alice-i-cecile given how early this is I think it is best to add `Needs design`

pcwalton:
Can someone help me understand why we can't just use BSN overrides for this? BSN should have the ability to say "I want to load scene X, but add my own custom entity as a child of object Y". That's core to the concept of what overrides are. Once we do that, this feature should naturally fall out, no?

hukasu:
@pcwalton for skinned meshes there is another things needed for it to work, not just parenting, i'm preparing another model to actually show that this works for skinned meshes

Gaeric:
@alice-i-cecile 
Hello! I'd like to check if there are any outstanding tasks or pending reviews for this PR. Could you please share the expected timeline for merging?

This PR is important for me as it implements the ability for multiple skins to share a single armature scene.

alice-i-cecile:
This needs merge conflict resolution, and reviews by animation experts. I am not one, but I will socialize your PR and try to get opinions :)

It also needs an updated release note: we have a nice little folder for them now!

viridia:
> Can someone help me understand why we can't just use BSN overrides for this? BSN should have the ability to say "I want to load scene X, but add my own custom entity as a child of object Y". That's core to the concept of what overrides are. Once we do that, this feature should naturally fall out, no?

Well, the solution in this PR is based on GLTF scenes, not BSN. So that feature of BSN isn't going to be applicable, unless we are saying that BSN is going to be the one and only fully-supported scene format going forward.

greeble-dev:
My two cents as someone who's worked with similar features in Unreal:

There's really two parts to this change - adding the option to make the glTF loader calculate `AnimationTargetId`s for all nodes, and adding joint rebinding.

The glTF loader changes could be useful for other features. One example is loading a glTF with a skinned mesh but no animations, then binding the mesh to animations loaded from a second glTF. There's even an argument that the loader should do this by default for convenience.

I do think the code could be simplified if `AnimationTarget` was refactored a bit ([proof of concept](https://github.com/bevyengine/bevy/compare/main...greeble-dev:bevy:refactor-animation-target)). I'm open to filing that as a separate PR - it would make this PR less controversial.

The second part is binding the joints of one skinned mesh to another mesh's animation player. Note that this is not just a rigid attachment, so as far I can tell it's not doable through simple parent/child relationships or BSN patches - there needs to be logic to find and match up joint entities. There's various other ways to do similar things but with different trade-offs. I think it's fine for this PR to start with one option, even if it might not fit all users.

The feature is clunky as it has to wait for one scene to spawn first, and then the other scene has to be spawned and tweaked. I don't think there's a better way to do this within the current Bevy asset/scene pipeline. As the pipeline evolves I'd guess this feature would move into a more robust transform step within the pipeline.

So the options as I see them are:

1. Accept this PR. It's a bit clunky now, but it can evolve alongside the asset/scene pipeline.
2. Accept just the glTF changes, possibly with tweaks (I can file a PR for this). The joint rebinding can be a third party crate, or it could be redone as an engine example - something that's easy for users to copy and paste. Once the pipeline has evolved it can be redone as an engine feature.

hukasu:
i'm in favor of doing the `AnimationTarget` refactor first, my solution in this PR is hella clunky and even i kinda dislike it, that is why i marked it as a experiment with the 🧪

greeble-dev:
Ok, I'll take a crack at a PR with the `AnimationTarget` refactor and glTf importer options. It's also fine if this PR lands first and I can do the refactor afterwards.

alice-i-cecile:
I'd be very happy to see that refactor in 0.18 :)

greeble-dev:
I've submitted two PRs for the refactor I proposed in my [previous comment](https://github.com/bevyengine/bevy/pull/18262#issuecomment-3214047856).

- #20774 (already merged).
- #22046.

PR #18263, opened 2025-03-11
Internalize BevyManifest logic. Switch to RwLock

### LINKED ISSUE #18103: Compile / typecheck perf regression in 0.16 from proc macro changes
## Bevy version

* Original: 669d139c13f6b44652f38131a5c9d20ca54e024d
* Current: 1b7db895b7ee71a2c2cfb8bd14eeb37c56ff9b44

```
cargo --version
cargo 1.85.0 (d73d2caf9 2024-12-31)
uname -a
Linux h3-arch 6.12.10-arch1-1 #1 SMP PREEMPT_DYNAMIC Sat, 18 Jan 2025 02:26:57 +0000 x86_64 GNU/Linux
```
## What's performing poorly?

Howdy! I've upgraded Tiny Glade's Bevy deps to `0.16.0-dev` to play around with auto-registration of reflection info, and I've noticed Rust Analyzer being sluggish. The Rust Analyzer slowdown is also reflected in running build/check/clippy. I've bisected it down to https://github.com/bevyengine/bevy/pull/17330

The test I'm doing is touching/modifying a single .rs file in our chonkiest crate The numbers are best out of a few runs.

### Using bevy at commit 669d139c13f6b44652f38131a5c9d20ca54e024d:
* Rust Analyzer build (running `clippy`) after a ctrl+s: 6.53s
* `cargo clippy`: 5.36s
* `cargo check`: 1.98s
* `cargo build` (touch one .rs file, no change): 3.03s
* `cargo build`: (modify one function) 4.39s

### Using bevy at commit 1b7db895b7ee71a2c2cfb8bd14eeb37c56ff9b44:
* Rust Analyzer build (running `clippy`) after a ctrl+s: 12.40s
* `cargo clippy`: 11.19s
* `cargo check`: 7.83s
* `cargo build` (touch one .rs file, no change): 8.78s
* `cargo build` (modify one function): 10.18s

## Before and After Traces

Traces for touching a single .rs file and running `cargo build --timings`

* Before: https://jsfiddle.net/fgyhduwn/
* After: https://jsfiddle.net/Lv5sb6hr/

## Additional information

Here's our Bevy dependencies:

```
bevy_app = "0.16.0-dev"
bevy_ecs = { version = "0.16.0-dev", features = ["multi_threaded", "serialize"] }
bevy_input = { version = "0.16.0-dev", features = ["serialize"] }
bevy_log = "0.16.0-dev"
bevy_reflect = { version = "0.16.0-dev" }
bevy_state = { version = "0.16.0-dev", default-features = false, features = ["bevy_app", "bevy_reflect"] }
bevy_tasks = "0.16.0-dev"
bevy_time = { version = "0.16.0-dev", default-features = false }
```

cart:
# Objective

Fixes #18103

#17330 introduced a significant compile time performance regression (affects normal builds, clippy, and Rust Analyzer).  While it did fix the type-resolution bug (and the general approach there is still our best known solution to the problem that doesn't involve [significant maintenance overhead](https://github.com/bevyengine/bevy/issues/18103#issuecomment-2702724676)), the changes had a couple of issues:

1. It used a Mutex, which poses a significant threat to parallelization.
2. It externalized existing, relatively simple, performance critical Bevy code to a crate outside of our control. I am not comfortable doing that for cases like this. Going forward @bevyengine/maintainer-team should be much stricter about this.
3. There were a number of other areas that introduced complexity and overhead that I consider unnecessary for our use case. On a case by case basis, if we encounter a need for more capabilities we can add them (and weigh them against the cost of doing so).

## Solution

1. I moved us back to our original code as a baseline
2. I selectively ported over the minimal changes required to fix the type resolution bug
3. I swapped `Mutex<BTreeMap<PathBuf, &'static Mutex<CargoManifest>>>` for `RwLock<BTreeMap<PathBuf, CargoManifest>>`. Note that I used the `parking_lot` RwLock because it has a mapping API that enables us to return mapped guards.


alice-i-cecile (approved):

atlv24 (approved):

atlv24:
Its kind of scary that a crate dep was added that easily. I dont think it was malicious here but it shows this is a very possible attack vector. Sneak an official-sounding crate you made into macro path, then do a patch release that gets automatically picked up by cargo which has the macros insert malicious code invisibly.

cart:
My thoughts exactly. Dependencies should be a "last resort" sort of thing, and when we take them, they should be in a trusted (known actor / high traffic / high visibility / high review) context.

cart:
I have removed the bevy crate remapping test as it is no longer supported in this impl. I did write code to re-add support, but supporting that means scanning and string comparing every dependency (and every dev dependency) on all macro executions that don't match the `bevy` or `bevy_x` branches. We'll hit that for every macro execution in all built-in crates, in addition to the "crate alias". I would rather drop support than pay this compile time cost, given that remapping is super uncommon (github global search revealed zero remapped macro executions), and given that workarounds exist (see the added code comment).

PR #18265, opened 2025-03-11
FilteredResource returns a Result instead of a simple Option

### LINKED ISSUE #17480: Return `Result` instead of `Option` from `FilteredResources::get` and similar

Resource might not exists or the access might be missing. It would be nice to reflect that in the return value.

Return `Result` instead of `Option` with an error.

andristarr:
# Objective
FilteredResource::get should return a Result instead of Option

Fixes #17480 

---

## Migration Guide

Users will need to handle the different return type on FilteredResource::get, FilteredResource::get_id, FilteredResource::get_mut as it is now a Result not an Option.

### DIFF
--- a/crates/bevy_ecs/src/world/error.rs
+++ b/crates/bevy_ecs/src/world/error.rs
@@ -59,12 +59,12 @@ pub enum EntityMutableFetchError {
 /// An error that occurs when getting a resource of a given type in a world.
 #[derive(thiserror::Error, Debug, Clone, Copy, PartialEq, Eq)]
 pub enum ResourceFetchError {
-    /// The resource does not exist in the world.
-    #[error("The resource does not exist in the world.")]
-    MissingResource,
-    /// The resource with the given [`ComponentId`] is not initialized in the world.
-    #[error("The resource with ID {0:?} is not initialized in the world.")]
-    NotInitialized(ComponentId),
+    /// The resource has never been initialized or registered with the world.
+    #[error("The resource has never been initialized or registered with the world. Did you forget to add it using `app.insert_resource` / `app.init_resource`?")]
+    NotRegistered,
+    /// The resource with the given [`ComponentId`] does not currently exist in the world.
+    #[error("The resource with ID {0:?} does not currently exist in the world.")]
+    DoesNotExist(ComponentId),
     /// Cannot get access to the resource with the given [`ComponentId`] in the world as it conflicts with an on going operation.
     #[error("Cannot get access to the resource with ID {0:?} in the world as it conflicts with an on going operation.")]
     NoResourceAccess(ComponentId),
--- a/crates/bevy_ecs/src/world/error.rs
+++ b/crates/bevy_ecs/src/world/error.rs
@@ -55,3 +55,14 @@ pub enum EntityMutableFetchError {
     #[error("The entity with ID {0} was requested mutably more than once")]
     AliasedMutability(Entity),
 }
+
+/// An error that occurs when getting a resource of a given type in a world.
+#[derive(thiserror::Error, Debug, Clone, Copy, PartialEq, Eq)]
+pub enum ResourceFetchError {
+    /// The resource does not exist in the world.
+    #[error("The resource does not exist in the world.")]
+    MissingResource,
+    /// No access to the resource with the given [`ComponentId`] in the world.
+    #[error("No access to the resource with ID {0:?} in the world.")]
+    NoResourceAccess(ComponentId),
+}
--- a/crates/bevy_ecs/src/world/error.rs
+++ b/crates/bevy_ecs/src/world/error.rs
@@ -62,6 +62,9 @@ pub enum ResourceFetchError {
     /// The resource does not exist in the world.
     #[error("The resource does not exist in the world.")]
     MissingResource,
+    /// The resource with the given [`ComponentId`] is not initialized in the world.
+    #[error("The resource with ID {0:?} is not initialized in the world.")]
+    NotInitialized(ComponentId),
     /// Cannot get access to the resource with the given [`ComponentId`] in the world as it conflicts with an on going operation.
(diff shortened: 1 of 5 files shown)


crates/bevy_ecs/src/world/error.rs
```diff
@@ -55,3 +55,14 @@ pub enum EntityMutableFetchError {
...
     AliasedMutability(Entity),
 }
+
+/// An error that occurs when getting a resource of a given type in a world.
+#[derive(thiserror::Error, Debug, Clone, Copy, PartialEq, Eq)]
+pub enum ResourceFetchError {
+    /// The resource does not exist in the world.
+    #[error("The resource does not exist in the world.")]
+    MissingResource,
+    /// No access to the resource with the given [`ComponentId`] in the world.
```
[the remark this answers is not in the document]
andristarr: Hmmm, how about "Cannot get access to resource as it conflicts with an on going operation"?

Something along this line, I am not a native speaker so feel free to suggest something that is better!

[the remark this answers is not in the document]
alice-i-cecile: Something like that is better, but I think the key thing to mention is that the conflicting access is with respect to the world :)

crates/bevy_ecs/src/world/filtered_resource.rs
```diff
@@ -151,21 +153,27 @@ impl<'w, 's> FilteredResources<'w, 's> {
...
+            .ok_or(ResourceFetchError::MissingResource)?;
+
+        Ok(Ref {
+            // SAFETY: `component_id` was obtained from the type ID of `R`.
+            value: unsafe { value.deref() },
+            // SAFETY: We have read access to the resource, so no mutable reference can exist.
+            ticks: unsafe { Ticks::from_tick_cells(ticks, self.last_run, self.this_run) },
+            // SAFETY: We have read access to the resource, so no mutable reference can exist.
+            changed_by: unsafe { caller.map(|caller| caller.deref()) },
         })
```
krunchington, line 177: very idiomatic, I like it.

crates/bevy_ecs/src/world/error.rs
```diff
@@ -55,3 +55,14 @@ pub enum EntityMutableFetchError {
     #[error("The entity with ID {0} was requested mutably more than once")]
     AliasedMutability(Entity),
 }
+
+/// An error that occurs when getting a resource of a given type in a world.
+#[derive(thiserror::Error, Debug, Clone, Copy, PartialEq, Eq)]
+pub enum ResourceFetchError {
+    /// The resource does not exist in the world.
+    #[error("The resource does not exist in the world.")]
+    MissingResource,
```
chescock, line 64: You included `ComponentId` in the `NoResourceAccess` variant, but not in `MissingResource`.  Would it be useful to include it there, as well?  I guess it's not available in the case where the resource isn't even initialized, but you could do `Option<ComponentId>`, or even separate variants for "not initialized" and "not present".

chescock (commented):
Oh, yay, `FilteredResources` is getting used!  

Should we update `FilteredResources::get_by_id` and the various `FilteredResourcesMut::get_mut` methods, as well?  The extra information seems just as valuable for those, and it would be odd for them to be different.

crates/bevy_ecs/src/world/error.rs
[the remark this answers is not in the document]
chescock: > What sort of access are we talking about here?

The main way to use `FilteredResources` is with a `SystemParamBuilder`, so the answer is usually that you needed to request it in the `FilteredResourcesParamBuilder`.  

(The other way to get one is to convert from `&World` or `&mut World`, but those will have access to all resources and shouldn't ever return this variant.)

crates/bevy_ecs/src/world/error.rs
andristarr: I do not see how can I derive if the resource is initialized or not in the world at all.  Is there some built in method I am missing here?

chescock (approved):

crates/bevy_ecs/src/world/error.rs
andristarr: Gotcha!

crates/bevy_ecs/src/world/error.rs
```diff
@@ -55,3 +55,17 @@ pub enum EntityMutableFetchError {
...
+    /// The resource with the given [`ComponentId`] is not initialized in the world.
+    #[error("The resource with ID {0:?} is not initialized in the world.")]
+    NotInitialized(ComponentId),
```
chescock, line 67: The names here feel a little backwards.  I might suggest `NotRegistered` to match the term from `World::register_resource` and `DoesNotExist` to match `EntityDoesNotExistError`.  

```suggestion
    /// The resource has never been initialized or registered with the world.
    #[error("The resource has never been initialized or registered with the world.")]
    NotRegistered,
    /// The resource with the given [`ComponentId`] does not currently exist in the world.
    #[error("The resource with ID {0:?} does not currently exist in the world.")]
    DoesNotExist(ComponentId),
```

We may even want the "Did you forget...?" line from the panic message in `World::resource` to be part of the display string for `NotRegistered`: https://github.com/bevyengine/bevy/blob/81c0900cf23218732c013656669dfdeece2937a8/crates/bevy_ecs/src/world/mod.rs#L1881-L1884

andristarr: Updated it, please check if this is what you had in mind :)

chescock: > Updated it, please check if this is what you had in mind :)

Yup, that looks good!!

chescock (approved):

alice-i-cecile (approved):

alice-i-cecile:
Technically needed a second approval. But I can do that for you :)

PR #18266, opened 2025-03-11
Remote entity reservation v2

### LINKED ISSUE #18003: Reserve entities from async

ElliottjPierce:
fixes #18003

# Objective

Long story short: we need to be able to reserve entities from outside `&World`. 

This is an alternative to #18195.

## Solution

Thanks to everyone who helped brainstorm on discord. There's still ideas being tossed around, but here's how this PR works:

All entity meta is kept in a single unlocked `Vec<EntityMeta>`. We keep a pending list that can be used remotely to reserve entities, etc.

When we allocate an entity directly, we pop from the new `owned: Vec<Entity>`, and if it's empty, we reserve some more entities in bulk. These entities are not flushed.

To ensure we don't loose entities anywhere, flushing rebalances `pending` and `owned`. We can tweak the proportions later, but currently, it splits the pending entities between the two evenly.

## Practical Change

It used to be that a brand new `Entities` would yield sequential indices via `alloc`. This was never a guarantee, but many tests relied on this ordering. The ordering is still deterministic but is no longer sequential or aligned. In other words, if you allocate 5 entities. You might get back indices `[0, 4, 2, 6, 7]`. Generally entities like this are opaque and this should not affect users. However, tests and benchmarks that depend on this previous, not guaranteed behavior had to be reworked.

One practical effect this may have is that systems that run through `Entity`s sequentially, will have a different order. This may affect query orders, observer ordering, etc. None of these things were guarantees before, but this may expose some existing bugs in those systems.

I suspect this is why some rendering systems broke.

## Future Work

Flushing is now completely unnecessary for many of the interactions with [`Entities`]. We still need to do it before applying command queues, etc,  but we don't need to flush before allocations or freeing. If we go in this direction, we should explore relaxing those requirements for performance.

## Testing

Functionality hasn't changed, so no new tests were added. A few old tests were revised because of the practical changes.

However, it makes sense to add tests for remote registration, so if anyone has specific concerns, I'd be happy to write a test for it.


crates/bevy_ecs/src/archetype.rs
```diff
@@ -50,6 +50,9 @@ impl ArchetypeRow {
     /// Index indicating an invalid archetype row.
     /// This is meant to be used as a placeholder.
     pub const INVALID: ArchetypeRow = ArchetypeRow(u32::MAX);
+    /// This is the same as [`Self::INVALID`], but it signals to [`Entities`](crate::entity::Entities)
```
alice-i-cecile, line 53: I'd like to see this expanded to explain why it exists.

crates/bevy_ecs/src/entity/mod.rs
```diff
@@ -528,197 +531,401 @@ impl<'a> core::iter::FusedIterator for ReserveEntitiesIterator<'a> {}
...
+impl core::ops::Deref for RemoteEntities {
+    type Target = EntityReservations;
+
+    #[inline]
+    fn deref(&self) -> &Self::Target {
+        self.0.deref()
+    }
+}
+
+/// This allows entity reservation.
```
alice-i-cecile, line 549: Remember to beef these up a bit once we have consensus on the basic mechanism.

crates/bevy_ecs/src/entity/mod.rs
```diff
@@ -528,197 +531,401 @@ impl<'a> core::iter::FusedIterator for ReserveEntitiesIterator<'a> {}
...
+    /// # Safety
     ///
-    /// [`alloc`]: Entities::alloc
-    /// [`reserve_entity`]: Entities::reserve_entity
-    /// [`reserve_entities`]: Entities::reserve_entities
-    /// [`flush`]: Entities::flush
-    pending: Vec<u32>,
-    free_cursor: AtomicIdCursor,
-}
+    /// This **MUST** not be called concurrently.
```
alice-i-cecile, line 584: More detail here about what this means, and a brief note about *why* this safety requirement exists would be helpful.

crates/bevy_ecs/src/entity/mod.rs
ElliottjPierce: Absolutely. Just don't want to write full docs until we settle on a design. 🥲

crates/bevy_ecs/src/archetype.rs
ElliottjPierce: Basically, when we reserve `Entity`s from `EntityReservations` for `Entities::owned`, we don't want to flush them like normal. They haven't been requested yet, so the flush doesn't make sense, and when they are requested by `Entities::alloc`, the caller will be the one to `Entities::set` the `EntityLocation`.

In principal, we could just have a `HashSet` of `Entity`s that have been in `Entities::owned` since the last flush, but that lookup takes time and memory.

Instead, by keeping a special `EntityLocation::INVALID_BUT_DONT_FLUSH`, we can determine whether or not to flush based only on if `meta.location == EntityLocation::INVALID` or not.

Hopefully that makes sense. This is complicated.

crates/bevy_ecs/src/entity/mod.rs
ElliottjPierce: Just added it. Good call.

crates/bevy_ecs/src/entity/mod.rs
```diff
@@ -492,14 +496,15 @@ impl SparseSetIndex for Entity {
     }
 }
 
-/// An [`Iterator`] returning a sequence of [`Entity`] values from
+/// An [`Iterator`] returning a sequence of [`Entity`] values from [`EntityReservations`].
+///
+/// # Dropping
+///
+/// All reserved entities will continue to be reserved after dropping, *including those that were not iterated*.
+/// Dropping this without finishing the iterator is effectively leaking an entity.
```
Carter0, line 504: How bad is "leaking an entity"? Leaking memory is pretty bad but maybe leaking an index is less so?

ElliottjPierce: It's not that bad at all, and it isn't done anywhere in Bevy that I know of. The docs here were just incomplete.

Carter0: ah okay I gotcha, thank you!

ElliottjPierce:
This seems to have a breaking change with testbed ui's text wrap. See [here](https://github.com/bevyengine/bevy/blob/e21dfe81cee3f2901e8424dea061d230f0c7f82b/examples/testbed/ui.rs#L342).

I tested it locally and the text seems to not be rendering (aside from its background) unless the mouse is hovering over it. When the mouse isn't hovering over a text item, only the most recently hovered item is rendered. There's no meaningful logs either.

I have no knowledge of the internals of the text system, and I have no idea why this is broken with this PR. It seems like this is actually an existing bug that this PR just uncovered, but IDK. AFAIK the text system doesn't depend on observer ordering or particularly ordered queries, so I can't think of a reason this would break.

ElliottjPierce:
@alice-i-cecile I apologize for the misdirection, but I actually don't think this needs a migration guide. I was almost certain it did before, but now that the dust has settled, I don't see any breaking changes. Let me know if I'm missing something.

ickshonpe:
> This seems to have a breaking change with testbed ui's text wrap. See [here](https://github.com/bevyengine/bevy/blob/e21dfe81cee3f2901e8424dea061d230f0c7f82b/examples/testbed/ui.rs#L342).
> 
> I tested it locally and the text seems to not be rendering (aside from its background) unless the mouse is hovering over it. When the mouse isn't hovering over a text item, only the most recently hovered item is rendered. There's no meaningful logs either.
> 
> I have no knowledge of the internals of the text system, and I have no idea why this is broken with this PR. It seems like this is actually an existing bug that this PR just uncovered, but IDK. AFAIK the text system doesn't depend on observer ordering or particularly ordered queries, so I can't think of a reason this would break.

Yep revealed bug, the sort key for the transparent UI phase uses a pair of the stack index and the render entities index. I guess the render entity index was intended to break ties but it's not needed as the sort is stable.

crates/bevy_ecs/src/entity/mod.rs
```diff
@@ -517,197 +520,417 @@ impl<'a> core::iter::FusedIterator for ReserveEntitiesIterator<'a> {}
...
+        let next_pending_index = self.next_pending_index.swap(-1, Ordering::Relaxed);
+        // SAFETY: we just told all the remote entities that the next index is -1, so nothing will access this while modifying.
+        let pending = unsafe { &mut *(self.pending().load(Ordering::Relaxed)) };
+
+        if next_pending_index > 0 {
+            let drain = pending.drain(..next_pending_index as usize);
+            into(drain);
         }
+
+        self.pending().store(pending, Ordering::Relaxed);
```
Brezak, line 647: What is this supposed to do? From my understanding of the semantics in this situations `AtomicPtr::store` stores an address into the pointer. So `self.pending` will place an atomic pointer on the stack. This store will then atomically overwrite the pointer address on the stack.

crates/bevy_ecs/src/entity/mod.rs
```diff
@@ -517,197 +520,417 @@ impl<'a> core::iter::FusedIterator for ReserveEntitiesIterator<'a> {}
...
+    /// only instead of capping at 0, it just gets increasingly negative.
+    /// We use this instead because atomicaly doing [`u32::saturating_sub`]
+    /// is more work than doing [`AtomicIdCursor::fetch_sub`].
+    next_pending_index: AtomicIdCursor,
+}
+
+impl EntityReservations {
+    fn pending(&self) -> AtomicPtr<Vec<Entity>> {
+        AtomicPtr::new(self.pending.get())
+    }
```
Brezak, line 565: Looking at the rest of the code. Either you or me have misunderstood how `AtomicPtr` works. My understanding is that the atomic loads and stores affect the address (and provenance) within the pointer. Not the data it's pointing to. What this function does is place an atomic pointer on the callers stack that points to `self.pending`. Any atomic loads and stores will affect the address on the stack not the `self.pending`.

ElliottjPierce: You're absolutely right. (This is my first time using atomic). I fixed it now, with some very unsafe, but so far correct pointer magic with Vec. Marking resolved for now.

Thanks so much for catching this, and please do a follow up review if possible since you know more about this stuff than I do.

crates/bevy_ecs/src/entity/mod.rs
```diff
@@ -511,203 +512,577 @@ impl<'a> Iterator for ReserveEntitiesIterator<'a> {
...
+        meta: &mut [EntityMeta],
+        owned: &mut Vec<Entity>,
+        mut needs_init: impl FnMut(Entity, &EntityMeta) -> bool,
+        mut init: impl FnMut(Entity, &mut EntityLocation),
+    ) {
+        // SAFETY: We return right after flushing if the bool is true
+        self.pending_scope_mut(|pending, next_pending_index, in_use_elsewhere| {
+            // flush
+            let new_pending_len = (next_pending_index + 1).max(0) as usize;
+            for reused in pending.drain(new_pending_len..) {
```
chescock, line 777: This touches `pending` before checking `in_use_elsewhere`.  Won't this be UB because `drain` takes `&mut` access to the contents of the slice while another thread could have `&` access?

crates/bevy_ecs/src/entity/mod.rs
```diff
@@ -511,203 +512,577 @@ impl<'a> Iterator for ReserveEntitiesIterator<'a> {
...
-        } else {
-            // Grab a new ID, outside the range of `meta.len()`. `flush()` must
-            // eventually be called to make it valid.
-            //
-            // As `self.free_cursor` goes more and more negative, we return IDs farther
-            // and farther beyond `meta.len()`.
-            Entity::from_raw(
-                u32::try_from(self.meta.len() as IdCursor - n).expect("too many entities"),
-            )
+        self.pending_scope(1, |reserved| {
```
chescock, line 951: If I understand correctly, then when a flush is running, `pending_scope_mut` ensures that calls to `pending_scope` return `None`.  And when that happens, we fall back to allocating new entity indexes.  Is that right?  

Won't that mean that `meta_len` continues to grow as the app runs, even if the total number of live entities doesn't?  Even if there are always enough recycled `Entity` values to use, we'll occasionally lose the race for the lock and allocate more memory in `entity_meta`.

crates/bevy_ecs/src/entity/mod.rs
ElliottjPierce: I went ahead and added it. Figured it would be nice for review anyway.

crates/bevy_ecs/src/archetype.rs
ElliottjPierce: This is now recorded in comments, so I'm marking as resolved.

chescock (commented):
Is the idea here roughly

```rust
struct EntityReservations {
    pending: RwLock<Vec<Entity>>,
    next_pending_index: AtomicIdCursor,
}
```

except that you're combining the atomic for the lock with the atomic for `next_pending_index` so that instead of three atomic operations (lock, decrement, unlock), you only have to do two (lock-decrement, unlock-decrement)?  

I think that could be made to work!  It might be easier to reason about if you can encapsulate the lock-plus-counter in a generic lock type, and present a safe API that `EntityReservations` can use.  

But there are still a lot of ways to introduce subtle race conditions, so I'd be wary of building all of this just to go from 3 atomic ops to 2.

ElliottjPierce:
> Is the idea here roughly
> 
> ```rust
> struct EntityReservations {
>     pending: RwLock<Vec<Entity>>,
>     next_pending_index: AtomicIdCursor,
> }
> ```
> 
> except that you're combining the atomic for the lock with the atomic for `next_pending_index` so that instead of three atomic operations (lock, decrement, unlock), you only have to do two (lock-decrement, unlock-decrement)?

That's pretty much exactly the idea, yes. There's some extra data in there like `meta_len`, but that's the core concept.

> I think that could be made to work! It might be easier to reason about if you can encapsulate the lock-plus-counter in a generic lock type, and present a safe API that `EntityReservations` can use.

I could do that. I don't think we would reuse it elsewhere, but it may make it easier to understand. Then again, its also very connected to the rest of `EntityReservations`, so it might become limiting. I think I'll come back to this after some more review.

> But there are still a lot of ways to introduce subtle race conditions, so I'd be wary of building all of this just to go from 3 atomic ops to 2.

This is hard, and it needs lots of review, but we don't just save one atomic op for this. Here's the inner RwLock's read functions:

```rust
    #[inline]
    pub fn try_read(&self) -> bool {
        self.state.fetch_update(Acquire, Relaxed, read_lock).is_ok()
    }

    #[inline]
    pub fn read(&self) {
        if !self.try_read() {
            self.lock_contended(false)
        }
    }
```
The function `read_lock` is some bit magic.

Anyway, `fetch_update` is implemented with another `load` and `store`, 2 atomic ops. Unlocking also uses 2. See [here](https://llvm.org/docs/Atomics.html#atomic-instructions) for the LLVM this lowers to. Note that `fetch_update` isn't available.

So this trick actually goes from 5 atomic ops with some additional bit work, to 2 atomic ops that just add/subtract (which does have direct hardware support), so this actually saves quite a lot. We are leveraging the fact that these ops are commutative and the ordering doesn't matter to get the seed up. The `RwLock` is only needed if ordering matters *and* the work is not commutative.

I know this takes careful review to get right, and I don't pretend to have caught all the race conditions here myself, but *if* we can sort out those details, I think this custom solution is worth the effort.

crates/bevy_ecs/src/entity/mod.rs
```diff
@@ -511,203 +512,577 @@ impl<'a> Iterator for ReserveEntitiesIterator<'a> {
...
+                let slice = core::slice::from_raw_parts::<'a>(starting, len as usize);
+                func(slice)
+            };
+
+            // successful, so we need to update parity
+            self.next_pending_index_truth
+                .fetch_sub(num, Ordering::Relaxed);
+            Some(result)
+        } else {
+            // unsuccessful, so we need to undo the change `next_pending_index_intention` to keep parity.
```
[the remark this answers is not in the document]
ElliottjPierce: Maybe I'm missing something. The first `fetch_sub(20)` isn't added back, it's subtracted from `next_pending_index_truth`, signifying a successful op. So that first one would get indices [0..10] exclusively. This leaves `next_pending_index_intention` at -10.

All the other ops just subtract `num`, see that it's negative, and add it back.
I don't think there's an issue here. Unless I'm not seeing some ordering problem??

crates/bevy_ecs/src/entity/mod.rs
ElliottjPierce: The other threads can only access the contents at/before `next_pending_index`, so the drain doesn't affect them. That is what determines the "length" in terms of what is still pending, though it's not the length in terms of `Vec::len`. Hopefully that made sense :).

Drain doesn't cause reallocations or changes to capacity, so it satisfies safety.  All it does is bring `Vec::len` back into agreement with `next_pending_index`. We have to return before the rebalancing since that can resize the vec or access contents before `next_pending_index`.

It's also worth mentioning that there's almost never a conflict here. A remote registration would have to be in progress when `flush` is called.

crates/bevy_ecs/src/entity/mod.rs
ElliottjPierce: That's exactly right. Keep in mind that this *only* happens if someone is remotely reserving an entity at precisely the same time as `flush_pending`. And even that reserved entity can be reused. We have 4 billion entities. This might eat four or five of those, maybe. 

If you can think of a better solution, I'm all for it, but I think this is a reasonable price to pay IMO.

crates/bevy_ecs/src/entity/mod.rs
```diff
@@ -511,203 +512,577 @@ impl<'a> Iterator for ReserveEntitiesIterator<'a> {
...
-    pub(crate) const fn new() -> Self {
-        Entities {
-            meta: Vec::new(),
-            pending: Vec::new(),
-            free_cursor: AtomicIdCursor::new(0),
+        let prev_pending_ptr = pending.as_ptr();
+        let prev_pending_cap = pending.capacity();
+
+        if next_pending_index == next_pending_index_intention {
+            // SAFETY: we just told all the remote entities that the next index is negative,
```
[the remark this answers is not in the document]
ElliottjPierce: That makes sense. I may not get the chance to fix this until Monday, but I do see the problem and I will fix it. 

This is exactly the kind of thing I'll need help catching, so thanks!

crates/bevy_ecs/src/entity/mod.rs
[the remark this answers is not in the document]
chescock: Oh, sorry, I thought the `<= 0` check was on the new value, not the old value.  

I think there is still a problem if the `next_pending_index_intention` goes from negative to positive in between the `fetch_sub` and `fetch_add`, though.  It looks like that can only happen from `pending_scope_mut`.  So, I think the sequence is 

- `pending_scope_mut` does a fetch_sub
- op for 1 does fetch_sub(1) and fails
- `pending_scope_mut` finishes and does a fetch_add
- op for 5 does fetch_sub(5) and succeeds
- op for 1 rolls back and does fetch_add(1)

And the op for 5 saw a value that was too small because it started before the op for 1 finished rolling back.

crates/bevy_ecs/src/entity/mod.rs
chescock: Oh, I see, the implementation of `drain()` never actually touches the rest of the slice, so that's not UB.  

Making an `&mut Vec` when we don't have exclusive access to the entire slice still makes me really nervous!  I think it would be a little safer to create only a `&[T]` when the slice was partially locked.  You could combine the `pending` and `in_use_elsewhere` parameters into a `Result`, or even pass two closures for the two cases.

crates/bevy_ecs/src/entity/mod.rs
chescock: My concern is that even a small memory leak will make Bevy unusable for really long-running applications.  

I suppose it's mitigated here by the fact that `flush` is only called with `&mut World` access, so if you're running in a regular system then `commands.spawn()` is guaranteed to be able to acquire a read lock.  But if we start using these for assets, then an application that cycles through different assets will leak memory.

crates/bevy_ecs/src/entity/mod.rs
```diff
@@ -511,203 +512,650 @@ impl<'a> Iterator for ReserveEntitiesIterator<'a> {
...
-    /// the end of `meta.len()`.
+    /// # Safety
     ///
-    /// This formulation allows us to reserve any number of IDs first from the freelist
-    /// and then from the new IDs, using only a single atomic subtract.
+    /// `func` may change the vec, but it *must* not resize it if the passed bool is true.
+    /// If the bool is true, it is actively being read from.
     ///
-    /// Once [`flush`] is done, `free_cursor` will equal `pending.len()`.
+    /// `func` must also not free the vec. In other words, it's capacity must be > 0 when it finishes.
```
Brezak, line 667: If I'm reading this section right were handing out a mutable reference to the `Vec` while other threads may be reading from it's buffer. If so then mutating the `Vec`'s contents will trigger UB since we'd be aliasing a mutable reference with immutable ones.

crates/bevy_ecs/src/entity/mod.rs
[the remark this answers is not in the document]
ElliottjPierce: Ok, I'm going to walk though this to help me think through it. Let's start from both the intention and truth values being equal to some medium positive value x, lets say 4.

- `pending_scope_mut` does a fetch_sub. Intention: x -u32::MAX, truth: x
- reserve for 1 starts. Intention x - u32::MAX - 1, truth: x
- `pending_scope_mut` finishes with exclusive access: Intention: x - 1 + y, truth: x + y (where y is the `diff` for resizes/rebalancing).
- reserve for 5 starts. Intention: x + y - 1 - 5, truth x + y.
- reserve for 1 fails (it was negative). Intention: x + y - 5, truth: x + y. (The reserve then returns a new index. One may have been available, but it was not safe to access.)
- reserve for 5 succeeds. It uses the slice ending on x + y - 1.  Intention:  x + y - 5, truth: x + y - 5. Oh no! We skipped the pending entity at x + y!

Yikes! Good catch!

When we base the slice on the intention, and the intention fails and is added to, reservations that happen in the middle skip whatever slice the failed reservation would have used.

I'm going to think of a fix here, but I wanted to go ahead and reply.

crates/bevy_ecs/src/entity/mod.rs
ElliottjPierce: > Making an `&mut Vec` when we don't have exclusive access to the entire slice still makes me really nervous! 

Agreed. I'm making better safety comments to help combat this.

> I think it would be a little safer to create only a `&[T]` when the slice was partially locked. You could combine the `pending` and `in_use_elsewhere` parameters into a `Result`, or even pass two closures for the two cases.

I tried this originally. I agree this would be better in an ideal world. But even if something is accessing the pending part of the vec, we still need to be able to change the vec's length to signify that the flush happened. I don't see a good solution. Instead, I'm throwing safety comments everywhere. It's very unsafe code, but it's also only a tiny function to check the safety for.

crates/bevy_ecs/src/entity/mod.rs
ElliottjPierce: To clarify, this isn't a memory leak. The entities that are reserved in this (very rare) case are not lost, and they are recycled as well. 

This only ever happens when you have something like allocating from async during an `&mut World` flush. *very* rare. Lets say this happens 10_000 times, which is completely absurd. That's wasted memory of a grand total of 200_000 bytes, not even a megabyte. 

And it's only wasted if an the app doesn't end up using them. Having extra pending entities can help balance `pending` vs `owned`. At the very least, these extra entities would speed up `alloc`. But again, there would be very few of these.

I really think this is a non issue. But maybe I'm missing something. Let me know if you think of a situation that could cause a problem here.

crates/bevy_ecs/src/entity/mod.rs
ElliottjPierce: This is actually safe, but it *is* really confusing. See [this](https://github.com/bevyengine/bevy/pull/18266#discussion_r1992605416) comment thread. I'm improving docs here to make this more clear.

I'm marking resolved so we don't have parallel threads, but definitely comment if you see anything fishy.

crates/bevy_ecs/src/entity/mod.rs
ElliottjPierce: Ok. I see. I still think this is such a small leak that it's a non issue, but you're right, the RwLock will fix it anyway, so let's not worry about it.

Just to keep you in the loop, I've had another idea for how to do this. Should be even faster, but we'll see. I'm going to experiment with that and then come back to the `RwLock` if it doesn't work. I've been inspired by some of your ideas and the left_right crate. I think I can cook up something that is fully lock free with minimal atomics. But I'll know more soon.

ElliottjPierce:
I'm going to close this in favor of v4. At time of writing, the final choice is up in the air, but it's pretty clear that v4 is better than this in every way. Thanks for all the review on this!

Issue #18238, opened 2025-03-10
Migrate examples to the improved Spawning API (to `children! from `with_children`)


krunchington:
The Improved Spawning API, https://github.com/bevyengine/bevy/pull/17521, provides ergonomics to spawning entity hierarchies.  Many of the examples leverage `with_children` and can be moved to the new API, both in order to share and popularize the approach as well as (hopefully!) to simplify the examples' implementation.

alice-i-cecile:
When doing this, feel free to do this one / several examples at once to ease review and reduce merge conflicts.

PR #18269, opened 2025-03-12
Add child! macro for single child

krunchington:
# Objective

Contributes to #18238 

While updating the `gamepad_viewer` example, I came across a place where I was using the `children!` macro to spawn a single child, but it had a tuple of components, e.g.
```rs
commands.spawn(
    ...,
    children![(
        Transform::from_xyz(0., 0., 1.),
        Text(format!("{:.3}", 0.)),
        TextFont {
            font_size: 13.,
            ..default()
        },
        TextWithButtonValue(button_type),
    )],
);
```
and I thought it was difficult to tell the difference between that and one that didn't wrap the single child in a tuple at a glance, e.g.
```rs
commands.spawn(
    ...,
    children![
        Transform::from_xyz(0., 0., 1.),
        Text(format!("{:.3}", 0.)),
        TextFont {
            font_size: 13.,
            ..default()
        },
        TextWithButtonValue(button_type),
    ],
);
```
Since this latter example spawns 4 children, each with different Components, and the first spawns a single child with all 4 Components, I thought it might be useful to have a `child!` macro in addition to `children!`, as an option for authors to write code that's easier to grok at a glance:
```rs
commands.spawn(
    ...,
    child!((
        Transform::from_xyz(0., 0., 1.),
        Text(format!("{:.3}", 0.)),
        TextFont {
            font_size: 13.,
            ..default()
        },
        TextWithButtonValue(button_type),
    )),
);
```

## Solution

Adds a new `child!` macro that is basically the same as the `children!` macro but without the repetition.  I'm pretty sure it's right because it worked in my testing but my macro-fu is VERY WEAK so extra eyes are very much appreciated here.

## Testing

  - see above

---

## Showcase

n/a

## Migration Guide

Optionally use the `child!` macro in places where you spawn a single child.  It can replace `children!` macros that take a single bundle, or instances of `with_child`.

### DIFF
--- a/crates/bevy_ecs/src/hierarchy.rs
+++ b/crates/bevy_ecs/src/hierarchy.rs
@@ -324,11 +324,12 @@ macro_rules! children {
     };
 }
 
-/// Returns a [`Spawn`](crate::spawn::Spawn) that will insert the [`Children`] component, spawn an
-/// entity with the given bundle that relates to the [`Children`] entity via the [`ChildOf`] component, and reserve space
-/// in the [`Children`] for the spawned entity
+/// Returns a [`SpawnRelatedBundle`] that will insert the [`Children`] component, spawn a [`SpawnableList`] of
+/// a single entity with the given bundle, which relates to the [`Children`] entity via the [`ChildOf`] component,
+/// and reserve space in the [`Children`] for the singularly spawned entity.
 ///
-/// Also see [`children`] for a version of this that works with many bundles.
+/// Also see [`children`] for a related macro which spawns multiple children with potentially
+/// different bundles.
 ///
 /// ```
 /// # use bevy_ecs::hierarchy::Children;
--- a/crates/bevy_ecs/src/hierarchy.rs
+++ b/crates/bevy_ecs/src/hierarchy.rs
@@ -293,6 +293,12 @@ pub fn validate_parent_has_component<C: Component>(
 ///
 /// Any additional arguments will be interpreted as bundles to be spawned.
 ///
+/// Also see [`child`] for a similar macro which spawns a single child.
+/// While the [`children`] macro can also be used to spawn a single child,
+/// `children![Foo, Bar]` spawns two children, one with Foo, and one with Bar, and `children![(Foo,
+/// Bar)]` spawns one child with both Foo and Bar. `child!((Foo,Bar))` is the alternative provided
+/// for authors seeking the ability to differentiate at a glance.
+///
 /// Also see [`related`](crate::related) for a version of this that works with any [`RelationshipTarget`] type.
 ///
 /// ```
@@ -329,7 +335,12 @@ macro_rules! children {
 /// and reserve space in the [`Children`] for the singularly spawned entity.
 ///
 /// Also see [`children`] for a related macro which spawns multiple children with potentially
-/// different bundles.
+/// different bundles. While the [`children`] macro can also be used to spawn a single child,
+/// `children![Foo, Bar]` spawns two children, one with Foo, and one with Bar, and `children![(Foo,
+/// Bar)]` spawns one child with both Foo and Bar. `child!((Foo,Bar))` is the alternative provided
+/// for authors seeking the ability to differentiate at a glance.
+///
+/// Also see [`related`](crate::related) for a macro similar to [`children`] but for use with any [`RelationshipTarget`] type.
 ///
 /// ```
 /// # use bevy_ecs::hierarchy::Children;
--- a/crates/bevy_ecs/src/hierarchy.rs
+++ b/crates/bevy_ecs/src/hierarchy.rs
@@ -323,6 +323,7 @@ pub fn validate_parent_has_component<C: Component>(
 /// [`RelationshipTarget`]: crate::relationship::RelationshipTarget
 /// [`SpawnRelatedBundle`]: crate::spawn::SpawnRelatedBundle
 /// [`SpawnableList`]: crate::spawn::SpawnableList
+/// [`child`]: crate::hierarchy::child
 #[macro_export]
 macro_rules! children {
     [$($child:expr),*$(,)?] => {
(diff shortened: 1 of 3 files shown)


krunchington:
Looking back at the example, it might make sense to make the expression repeatable for multiple components/bundles, instead of requiring that to be passed in a tuple, `child!((Foo,Bar))` -> `child!(Foo, Bar)`

LMK if anyone has opinions about that.

krunchington:
Oh duh I should probably add the new macro to the prelude.  Was wondering why the children wasn't imported...

examples/tools/gamepad_viewer.rs
```diff
@@ -374,18 +358,17 @@ fn setup_triggers(
...
+    commands.spawn((
+        Text::new("Connected Gamepads:\n"),
+        Node {
+            position_type: PositionType::Absolute,
+            top: Val::Px(12.),
+            left: Val::Px(12.),
+            ..default()
+        },
+        ConnectedGamepadsText,
+        child!(TextSpan::new("None")),
```
alice-i-cecile, line 370: I think this is the only example change that makes sense in this PR.

alice-i-cecile (commented):
Can you just call the `children!` macro with a single element? That seems like the cleanest design to me.

krunchington:
Hey @alice-i-cecile thanks for the feedback!  I'll preface this by saying I'm totally fine if we don't want to take this approach, it was just something I came up with while tinkering with the examples.  I appreciate the consideration and conversation on it!

I think the example in the PR description as to potential value of the new macro is the most illustrative- in a more condensed format:

- `children![Foo, Bar]` spawns 2 children, one with Foo, one with Bar
- `children![(Foo, Bar)]` spawns 1 child, with both Foo and Bar

The difference is really subtle and hard to see when quickly scanning code (especially when the tuple is split over multiple lines as in the full examples in the PR description), so my thought was that as an author having a way to use

- `child!((Foo,Bar))` spawns 1 child, with both Foo and Bar

is easier to see in code review, and help me maintain my codebase.  I'll leave an inline comment on the exact lines that inspired this for me in the PR as well.

crates/bevy_ecs/src/hierarchy.rs
```diff
@@ -324,6 +324,36 @@ macro_rules! children {
     };
 }
 
+/// Returns a [`SpawnRelatedBundle`] that will insert the [`Children`] component, spawn a [`SpawnableList`] of
+/// a single entity with the given bundle, which relates to the [`Children`] entity via the [`ChildOf`] component,
+/// and reserve space in the [`Children`] for the singularly spawned entity.
+///
+/// Also see [`children`] for a related macro which spawns multiple children with potentially
+/// different bundles.
```
alice-i-cecile, line 332:
```suggestion
/// different bundles. While this macro can be used to spawn a single child,
/// `children![Foo, Bar]` spawns two children, while `children![(foo, bar)]` only spawns one,
/// which can result in hard-to-detect bugs.
```

examples/tools/gamepad_viewer.rs
```diff
@@ -347,25 +332,24 @@ fn setup_triggers(
...
+            ),
+            child!((
+                Transform::from_xyz(0., 0., 1.),
+                Text(format!("{:.3}", 0.)),
+                TextFont {
+                    font_size: 13.,
+                    ..default()
+                },
+                TextWithButtonValue(button_type),
+            )),
```
krunchington, line 351: @alice-i-cecile This was the single child that inspired my approach.  Without the child! macro it was
```rs
children![(
    Transform::from_xyz(0., 0., 1.),
    Text(format!("{:.3}", 0.)),
    TextFont {
        font_size: 13.,
        ..default()
    },
    TextWithButtonValue(button_type),
)]
```
when I typed that up I was like "huh, this is hard to tell if it's one child or multiple children, I could see that being hard to grok in my own game code"

alice-i-cecile (approved):
Okay, that motivation makes a lot of sense to me. I've left a suggestion for how to incorporate it into the docs. Your example was extremely helpful to me, and I think we should share it with users.

crates/bevy_ecs/src/hierarchy.rs
krunchington: Updated with a little extra as well.  I added some to the `children` macro docs to increase the visibility for people scanning their options.

examples/tools/gamepad_viewer.rs
```diff
@@ -347,25 +332,24 @@ fn setup_triggers(
...
-                    Text(format!("{:.3}", 0.)),
-                    TextFont {
-                        font_size: 13.,
-                        ..default()
-                    },
-                    TextWithButtonValue(button_type),
-                ));
-            });
+            ),
+            child!((
```
cart, line 343: There are two reasons why I didn't include this:

(1) People will be tempted to do this when they want two children:

```rust
worlds.spawn((
  Thing,
  child![NestedThing],
  child![NestedThing],
));
```

But functionally this would produce two `Children` components on the root, each with one entity. This would panic because you can't have duplicate components in a bundle.

(2)  Adding a second child would therefore mean you need to refactor to `children![]` after the first child. This is "annoying" for people

(3) Having one way to do things that always works the same way makes life more blissful. Never needing to think "do I want to use `child!` or `children!` here" and never needing to refactor between them, never needing to learn both, etc.

examples/tools/gamepad_viewer.rs
cart: I agree that this disambiguates, but I don't think the wins here outweigh the "price" that we pay (see my comment below).

examples/tools/gamepad_viewer.rs
krunchington: Yeah I think those are reasonable counterarguments.  Personally I think the refactor is "annoying" but worth the pain.  The deviation from having one way to do things is maybe not worth the potential value.

I just thought of this- If we don't merge this feature I think I'd be inclined, in my own code, to approach this with 
```rs
Children::spawn(
    Spawn((Foo,Bar))
    ),
```
which requires no additional macro but still makes it more visible at a glance even if it's more verbose.

cart:
Closing for now as I'm pretty against this for the reasons stated above. But feel free to continue discussing and we can reopen as necessary.

krunchington:
Thanks cart!  The reasoning makes sense- if I feel like it deserves reconsideration I won't be shy about bringing it back up!

PR #18272, opened 2025-03-12
Switch to `ImDocument` in `BevyManifest`

bushrat011899:
# Objective

When reviewing #18263, I noticed that `BevyManifest` internally stores a `DocumentMut`, a mutable TOML document, instead of an `ImDocument`, an immutable one. The process of creating a `DocumentMut` first involves creating a `ImDocument` and then cloning all the referenced spans of text into their own allocations (internally referred to as `despan` in `toml_edit`). As such, using a `DocumentMut` without mutation is strictly additional overhead.

In addition, I noticed that the filesystem operations associated with reading a manifest and parsing it were written to be completed _while_ a write-lock was held on `MANIFESTS`. This likely doesn't translate into a performance or deadlock issue as the manifest files are generally small and can be read quickly, but it is generally considered a bad practice.

## Solution

- Switched to `ImDocument<Box<str>>` instead of `DocumentMut`
- Re-ordered operations in `BevyManifest::shared` to minimise time spent holding open the write-lock on `MANIFESTS`

## Testing

- CI

---

## Notes

I wasn't able to measure a meaningful performance difference with this PR, so this is purely a code quality change and not one for performance.

### DIFF
--- a/crates/bevy_macro_utils/src/bevy_manifest.rs
+++ b/crates/bevy_macro_utils/src/bevy_manifest.rs
@@ -35,12 +35,12 @@ impl BevyManifest {
             }
         }
 
-        let key = manifest_path.clone();
         let manifest = BevyManifest {
             manifest: Self::read_manifest(&manifest_path),
             modified_time,
         };
-
+        
+        let key = manifest_path.clone();
         let mut manifests = MANIFESTS.write();
         manifests.insert(key, manifest);
 
--- a/crates/bevy_macro_utils/src/bevy_manifest.rs
+++ b/crates/bevy_macro_utils/src/bevy_manifest.rs
@@ -8,12 +8,12 @@ use std::{
     path::{Path, PathBuf},
     time::SystemTime,
 };
-use toml_edit::{DocumentMut, Item};
+use toml_edit::{ImDocument, Item};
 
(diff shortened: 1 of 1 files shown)


TimJentzsch (approved):

crates/bevy_macro_utils/src/bevy_manifest.rs
```diff
@@ -34,14 +34,15 @@ impl BevyManifest {
                 return manifest;
             }
         }
+
+        let key = manifest_path.clone();
```
MrGVSV, line 38: Nit: is there a reason the definition of `key` is up here and not by its usage?

MrGVSV (approved):

alice-i-cecile:
Leaving this for @cart to merge for now, mostly because I think it's good for him to see it :)

crates/bevy_macro_utils/src/bevy_manifest.rs
bushrat011899: Purely because it's a boxed string (so cloning requires allocation), and we take a write lock. Similar to the filesystem operations, allocating involves syscalls and _can_ block, so in general you want to do as little within the locked scope as possible.

cart (approved):
Good calls all around!

PR #10078, opened 2023-10-10
Fix UI borders

rparrett:
# Objective

Fixes #10069

## Solution

Extracted UI nodes were previously stored in a `SparseSet` and had a predictable iteration order. UI borders and outlines relied on this. Now they are stored in a HashMap and that is no longer true.

This adds `entity.index()` to the sort key for `TransparentUi` so that the iteration order is predictable and the "border entities" that get spawned during extraction are guaranteed to get drawn after their respective container nodes again.

I **think** that everything still works for overlapping ui nodes etc, because the z value / primary sort is still controlled by the "ui stack."

Text above is just my current understanding. A rendering expert should check this out.

I will do some more testing when I can.

PR #18273, opened 2025-03-12
Remove the entity index from the UI phase's sort key

ickshonpe:
# Objective

The sort key for the transparent UI phase is a (float32, u32) pair consisting of the stack index and the render entity's index. 
I guess the render entity index was intended to break ties but it's not needed as the sort is stable.  It also assumes the indices of the render entities are generated sequentially, which isn't guaranteed.

Fixes the issues with the text wrap example seen in #18266

## Solution

Change the sort key to just use the stack index alone.


kristoff3r (approved):

rparrett:
This was done in #10078, but should no longer be necessary after #17618. 👍

rparrett:
Though I really don't understand the issue in #18266. The entity index should only come into play when the `z` values / `UiStack` index are the same, right?

Are you saying that the text and the background end up with the same `z` value? I don't see how that could happen in that testbed.

rparrett:
Oh I see, the individual `Text` nodes also have background colors. It seems like there may be a deeper underlying issue that the text has no z offset from its background?

rparrett:
As a followup if/when this lands, we should check to see if `radsort` is faster for `TransparentUi` sorting.

ickshonpe:
The background and text glyphs both have the same stack index but the glyphs are extracted after the backgrounds and that ordering is preserved by the stable sort.

rparrett (approved):

PR #18279, opened 2025-03-12
Add process cpu/memory usage to SystemInformationDiagnosticsPlugin

### LINKED ISSUE #18135: Add current process cpu/memory usage to SystemInformationDiagnosticsPlugin

System cpu/memory is useful but current process cpu/memory is even more useful. 

Adding 2 extra metrics:
* process/cpu_usage in %core, eg 200% if 2 cores in use
* process/mem_usage in MB

Extending the plugin in my own app

## Additional context

Following code would need to be added to fetch the info.

```rust
                let pid = sysinfo::get_current_pid().expect("Failed to get current process ID");
                sys.refresh_processes(ProcessesToUpdate::Some(&[pid]), true);

                let process_mem_usage = sys
                    .process(pid)
                    .map(|p| p.memory() as f64 / 1024.0 / 1024.0) // Convert to MB
                    .unwrap_or(0.0);

                let process_cpu_usage = sys
                    .process(pid)
                    .map(|p| p.cpu_usage() as f64)
                    .unwrap_or(0.0);
```

If there is interest in this I'm willing to create a PR

tjlaxs:
# Objective

- Adding process specific cpu/mem usages to SystemInformationDiagnosticsPlugin
- Fixes #18135

## Solution

- Adding two new stats by using code provided in #18135

## Testing

- Tested by adding SystemInformationDiagnosticsPlugin into enabling_disabling_diagnostics example
- Tested only on Linux (Ubuntu 24.04.2 LTS)

---

## Showcase

Example output:
> 2025-03-12T18:20:45.355206Z  INFO bevy diagnostic: fps              :  144.139984   (avg 143.968838)    
> 2025-03-12T18:20:45.355229Z  INFO bevy diagnostic: system/cpu_usage :   17.299578%  (avg 16.410863%)    
> 2025-03-12T18:20:45.355235Z  INFO bevy diagnostic: frame_time       :    6.939720ms (avg 6.953508ms)    
> 2025-03-12T18:20:45.355239Z  INFO bevy diagnostic: frame_count      : 1271.000000    
> 2025-03-12T18:20:45.355243Z  INFO bevy diagnostic: process/cpu_usage:  172.151901%  (avg 165.337555%)    
> 2025-03-12T18:20:45.355247Z  INFO bevy diagnostic: process/mem_usage:  400.472656%  (avg 400.478516%)    
> 2025-03-12T18:20:45.355250Z  INFO bevy diagnostic: system/mem_usage :   34.244571%  (avg 34.356289%)

### DIFF
--- a/crates/bevy_diagnostic/src/system_information_diagnostics_plugin.rs
+++ b/crates/bevy_diagnostic/src/system_information_diagnostics_plugin.rs
@@ -102,7 +102,8 @@ pub mod internal {
             Diagnostic::new(SystemInformationDiagnosticsPlugin::PROCESS_CPU_USAGE).with_suffix("%"),
         );
         diagnostics.add(
-            Diagnostic::new(SystemInformationDiagnosticsPlugin::PROCESS_MEM_USAGE).with_suffix("%"),
+            Diagnostic::new(SystemInformationDiagnosticsPlugin::PROCESS_MEM_USAGE)
+                .with_suffix("MB"),
         );
     }
 
@@ -165,7 +166,7 @@ pub mod internal {
 
                 let process_cpu_usage = sys
                     .process(pid)
-                    .map(|p| p.cpu_usage() as f64)
+                    .map(|p| p.cpu_usage() as f64 / sys.cpus().len() as f64)
                     .unwrap_or(0.0);
 
                 SysinfoRefreshData {
--- a/crates/bevy_diagnostic/src/system_information_diagnostics_plugin.rs
+++ b/crates/bevy_diagnostic/src/system_information_diagnostics_plugin.rs
@@ -105,7 +105,7 @@ pub mod internal {
         );
         diagnostics.add(
             Diagnostic::new(SystemInformationDiagnosticsPlugin::PROCESS_MEM_USAGE)
-                .with_suffix("MB"),
+                .with_suffix("GiB"),
         );
     }
 
@@ -156,14 +156,13 @@ pub mod internal {
                 sys.refresh_cpu_specifics(CpuRefreshKind::nothing().with_cpu_usage());
                 sys.refresh_memory();
                 let system_cpu_usage = sys.global_cpu_usage().into();
-                // `memory()` fns return a value in bytes
-                let total_mem = sys.total_memory() as f64 / BYTES_TO_GIB;
-                let used_mem = sys.used_memory() as f64 / BYTES_TO_GIB;
+                let total_mem = sys.total_memory() as f64;
+                let used_mem = sys.used_memory() as f64;
                 let system_mem_usage = used_mem / total_mem * 100.0;
 
                 let process_mem_usage = sys
                     .process(pid)
-                    .map(|p| p.memory() as f64 / 1024.0 / 1024.0) // Convert to MB
+                    .map(|p| p.memory() as f64 * BYTES_TO_GIB)
                     .unwrap_or(0.0);
 
                 let process_cpu_usage = sys
--- a/crates/bevy_diagnostic/src/system_information_diagnostics_plugin.rs
+++ b/crates/bevy_diagnostic/src/system_information_diagnostics_plugin.rs
@@ -29,9 +29,9 @@ impl Plugin for SystemInformationDiagnosticsPlugin {
 
 impl SystemInformationDiagnosticsPlugin {
     /// Total system cpu usage in %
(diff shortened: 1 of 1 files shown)


crates/bevy_diagnostic/src/system_information_diagnostics_plugin.rs
```diff
@@ -144,9 +158,21 @@ pub mod internal {
...
                 let current_used_mem = used_mem / total_mem * 100.0;
 
+                let process_mem_usage = sys
+                    .process(pid)
+                    .map(|p| p.memory() as f64 / 1024.0 / 1024.0) // Convert to MB
+                    .unwrap_or(0.0);
+
+                let process_cpu_usage = sys
+                    .process(pid)
+                    .map(|p| p.cpu_usage() as f64)
```
IceSentry, line 168: The doc mentions that it should be divided by the number of core to get a more representative value.

crates/bevy_diagnostic/src/system_information_diagnostics_plugin.rs
```diff
@@ -32,6 +32,10 @@ impl SystemInformationDiagnosticsPlugin {
     pub const CPU_USAGE: DiagnosticPath = DiagnosticPath::const_new("system/cpu_usage");
     /// Total system memory usage in %
     pub const MEM_USAGE: DiagnosticPath = DiagnosticPath::const_new("system/mem_usage");
+    /// Process cpu usage in %
+    pub const PROCESS_CPU_USAGE: DiagnosticPath = DiagnosticPath::const_new("process/cpu_usage");
```
IceSentry, line 36: I feel like the other ones should probably be renamed to `GLOBAL_CPU_USAGE` or `SYSTEM_CPU_USAGE` to make the difference clearer. (same for mem of course)

IceSentry (changes requested):
This is neat, I didn't realize sysinfo was able to do that. Just a few nitpicks.

I tested this on windows 11 and it works. Although, the values seem a bit off? It gives me values higher than what task manager reports.

crates/bevy_diagnostic/src/system_information_diagnostics_plugin.rs
```diff
@@ -144,9 +158,21 @@ pub mod internal {
                 let used_mem = sys.used_memory() as f64 / BYTES_TO_GIB;
                 let current_used_mem = used_mem / total_mem * 100.0;
 
+                let process_mem_usage = sys
+                    .process(pid)
+                    .map(|p| p.memory() as f64 / 1024.0 / 1024.0) // Convert to MB
```
IceSentry, line 163: This should use the same conversion as the other memory related diagnostics and convert it to GiB. The suffix should also be adapted for this change.

crates/bevy_diagnostic/src/system_information_diagnostics_plugin.rs
```diff
@@ -94,11 +98,19 @@ pub mod internal {
             .add(Diagnostic::new(SystemInformationDiagnosticsPlugin::CPU_USAGE).with_suffix("%"));
         diagnostics
             .add(Diagnostic::new(SystemInformationDiagnosticsPlugin::MEM_USAGE).with_suffix("%"));
+        diagnostics.add(
+            Diagnostic::new(SystemInformationDiagnosticsPlugin::PROCESS_CPU_USAGE).with_suffix("%"),
+        );
+        diagnostics.add(
+            Diagnostic::new(SystemInformationDiagnosticsPlugin::PROCESS_MEM_USAGE).with_suffix("%"),
```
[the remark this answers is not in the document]
tjlaxs: Yes, now that I use my brain... 400% mem usage sounds bad. But also cannot be sure what the value actually is. My system monitor says that the app uses 90MB and calculates to 400MB.

crates/bevy_diagnostic/src/system_information_diagnostics_plugin.rs
tjlaxs: The earlier BYTES_TO_GIB were useless as they will disappear in the division on line 159. (Sidenote: they also did the conversion to wrong direction)

IceSentry (approved):

IceSentry:
@tjlaxs can you resolve the conversations so other reviewers don't think there are comments that need to be addressed

alice-i-cecile (approved):

Issue #17272, opened 2025-01-09
Unify system and command error handling


alice-i-cecile:
> But it's weird that this is not unified with the system result handler

Hmm yeah we should think about how to unify these. I think applying the global handler to the system result handler by default makes some sense. I'd prefer that we tackle unification in a separate PR though.

> I had some idea that we could be returning these errors to the scheduler through the sync-point system, idk if that's a viable.

Worth considering all of our options. Imo the direct handler approach is nice because it gives us both granularity and nice performance (no need to queue results up, look up which handler to use, etc). I think we should start from there with essentially optimal performance and then compare against the fancier options later.

> I don't really think this will be a niche feature though. Don't we anticipate the editor using custom error handling to populate an error console? But that's nothing to do with the implementation

Hmmm yeah thats true. Although thats an implementation we'd own ourselves, and editor integration would likely be behind a cargo feature (so we could abstract that out from a user perspective). I think the majority of users shouldn't need to explicitly think about custom global error handlers.

_Originally posted by @cart in https://github.com/bevyengine/bevy/issues/17215#issuecomment-2576481039_

alice-i-cecile:
System error handling:

- found in `bevy_ecs::error` module
- globally configured via `DefaultSystemErrorHandler` resource, which has App and Subapp helper methods
- locally configured via system piping
- handler is of the form `fn(BevyError, SystemErrorContext)`

Command error handling:

- found in `bevy_ecs::system::commands` module
- globally configured via the `GLOBAL_ERROR_HANDLER` `OnceLock`, which needs the `configurable_error_handler` feature to be enabled
- locally configured via `EntityCommands::queue_handled`
- handler is of the form `fn(&mut World, BevyError)`

Proposal:

- centralized in the `bevy_ecs::error` module
-  globally configured by setting the `DefaultErrorHandler` resource, with no helper methods (one way to do this please)
- locally configured via system piping or `queue_handled`
- handler is of the form `fn(&mut World, BevyError, EcsErrorContext)`

cart:
One consideration here is Observer error handling (which also uses `fn(BevyError, SystemErrorContext)`), which is executed with only DeferredWorld access.

NthTensor:
It wouldn't be too terrible to make all error handling use deferred access. I doubt making structural changes to the world after an error will be common practice, and commands would be fine for that.

alice-i-cecile:
Poking at this, the hard bits appear to be:

[...]

PR #18280, opened 2025-03-12
Small cleanup for ECS error handling

alice-i-cecile:
# Objective

While poking at https://github.com/bevyengine/bevy/issues/17272, I noticed a few small things to clean up.

## Solution

- Improve the docs
- ~~move `SystemErrorContext` out of the `handler.rs` module: it's not an error handler~~

### DIFF
--- a/crates/bevy_ecs/src/error/handler.rs
+++ b/crates/bevy_ecs/src/error/handler.rs
@@ -1,14 +1,6 @@
-use crate::{component::Tick, error::BevyError, resource::Resource};
-use alloc::borrow::Cow;
+use crate::{error::BevyError, resource::Resource};
 
-/// Additional context for a failed system run.
-pub struct SystemErrorContext {
-    /// The name of the system that failed.
-    pub name: Cow<'static, str>,
-
-    /// The last tick that the system was run.
-    pub last_run: Tick,
-}
+use super::SystemErrorContext;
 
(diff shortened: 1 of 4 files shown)


crates/bevy_ecs/src/error/handler.rs
```diff
@@ -1,14 +1,6 @@
-use crate::{component::Tick, error::BevyError, resource::Resource};
-use alloc::borrow::Cow;
+use crate::{error::BevyError, resource::Resource};
 
-/// Additional context for a failed system run.
-pub struct SystemErrorContext {
```
cart, line 5: I believe the rationale here was that SystemErrorContext is a "dependency of an error handler". It exists for and is defined by the "error handler context", so it belongs either in the error handler module, or as a submodule (and by homing it next to the handlers, it gives readers of that code the knowledge of what is being passed in).

cart: (I didn't write that code, but that was my read of the situation)

alice-i-cecile: That makes sense to me. My rationale here was "make handler.rs ignorable", by making it a simple collection of premade error handlers. I'll just revert this though.

mnmaita (approved):

cart (approved):

PR #18287, opened 2025-03-13
Update getrandom requirement from 0.2.0 to 0.3.1 

mnmaita:
# Objective

- Fixes #18222

## Solution

- Updated `getrandom` version requirement from 0.2.0 to 0.3.1.
- Replaced the deprecated "js" feature with "wasm_js".

## Testing

- Ran CI locally.


mnmaita:
We might need to update this alongside `rand` and maybe other crates too.

PR #18288, opened 2025-03-13
Add `examples/helpers/*` as library examples

bushrat011899:
# Objective

Some of Bevy's examples contain boilerplate which is split out into the `helpers` folder. This allows examples to have access to common functionality without building into Bevy directly. However, these helpers are themselves quite high-quality code, and we do intend for users to read them and even use them. But, we don't list them in the examples document, and they aren't explicitly checked in CI, only transitively through examples which import them.

## Solution

- Added `camera_controller` and `widgets` as library examples.

## Testing

- CI

---

## Notes

- Library examples are identical to any other example, just with `crate-type = ["lib"]` in the `Cargo.toml`. Since they are marked as libraries, they don't require a `main` function but do require public items to be documented.
- Library examples opens the possibility of creating examples which don't need to be actual runnable applications. This may be more appropriate for certain ECS examples, and allows for adding helpers which (currently) don't have an example that needs them without them going stale.
- I learned about this as a concept during research for `no_std` examples, but believe it has value for Bevy outside that specific niche.


Cargo.toml
```diff
@@ -4263,3 +4263,27 @@ name = "Occlusion Culling"
...
+
+[[example]]
+name = "camera_controller"
+path = "examples/helpers/camera_controller.rs"
+doc-scrape-examples = true
+crate-type = ["lib"]
+
+[package.metadata.example.camera_controller]
+name = "Camera Controller"
+description = "Demonstration of Occlusion Culling"
```
mgi388, line 4275: copy paste error

bushrat011899: Whoops! Good catch!

examples/helpers/camera_controller.rs
```diff
@@ -10,6 +10,7 @@ use bevy::{
 };
 use std::{f32::consts::*, fmt};
 
+/// A freecam-style camera controller plugin.
```
Carter0, line 13: Maybe worth adding a comment about how this example is used by other examples and that's why it doesn't have an app.run() function call. I imagine many users would just assume this is another normal example and maybe try to pull it down and run it.

Carter0 (approved):
I like code visibility! These examples seem broadly useful I agree

examples/helpers/camera_controller.rs
bushrat011899: Good point, I've added some documentation to both helper examples!

BenjaminBrienen (approved):

alice-i-cecile (approved):
I'm down with that for now.

PR #18296, opened 2025-03-13
Shrink from_reflect_with_fallback

SpecificProtagonist:
# Objective

Reduce duplicate code generated by `from_reflect_with_fallback`.

## Solution

Polymorphize part of the function.

In a binary with `#[reflect(Component)]` for 100 types (and nothing else), this reduces the size of the release binary by 3.8% in my testing.

The impact on a normal Bevy app will be almost nothing, but I'm hoping to make more PRs like this.

### DIFF
--- a/crates/bevy_ecs/src/reflect/mod.rs
+++ b/crates/bevy_ecs/src/reflect/mod.rs
@@ -94,6 +94,7 @@ pub fn from_reflect_with_fallback<T: Reflect + TypePath>(
     world: &mut World,
     registry: &TypeRegistry,
 ) -> T {
+    #[inline(never)]
     fn type_erased(
         reflected: &dyn PartialReflect,
         world: &mut World,


crates/bevy_ecs/src/reflect/mod.rs
```diff
@@ -94,53 +94,54 @@ pub fn from_reflect_with_fallback<T: Reflect + TypePath>(
...
-            T::type_path(),
-        );
-    }
-
-    // First, try `FromReflect`. This is handled differently from the others because
-    // it doesn't need a subsequent `apply` and may fail.
-    if let Some(reflect_from_reflect) =
-        registry.get_type_data::<ReflectFromReflect>(TypeId::of::<T>())
-    {
+    fn type_erased(
```
james7132, line 98:
```suggestion
    #[inline(never)]
    fn type_erased(
```

alice-i-cecile (approved):

PR #18297, opened 2025-03-13
Simplified system error handling

alice-i-cecile:
# Objective

While working towards #17272, I found a few things to clean up for system error handling. Since I intend to *mostly* use the system error handling approach for both systems and commands, I wanted to clean it up and reduce complexity before porting commands over to it to.

This is a good, simple stopping point for review, so here's the second PR in the series :)

## Solution

- rename the `fallible_systems` example to the less mysterious and more general `error_handling`
- rename `DefaultSystemErrorHandler` to the more general and more informative `FallbackErrorHandler`
- remove the helper methods on `App` and `SubApp` for configuring this: all they do is set a value of a resource
- remove the per-schedule error handlers: this is a fair bit of added complexity and inconsistency that can result in accidentally missing logs or shipping panics and no clear motivating user story exists

## Testing

- `error_handling` example works as intended


crates/bevy_ecs/src/schedule/schedule.rs
```diff
@@ -449,7 +440,7 @@ impl Schedule {
         self.initialize(world)
             .unwrap_or_else(|e| panic!("Error when initializing schedule {:?}: {e}", self.label));
 
-        let error_handler = self.error_handler.expect("schedule initialized");
+        let error_handler = world.get_resource_or_init::<FallbackErrorHandler>().0;
```
krunchington, line 443: The FallbackErrorHandler is mentioned as the thing that bevy will use if nothing else is provided- I assume this PR is an initial step and then this code will be updated later to grab the appropriate configured error handler for use at runtime?

krunchington:
many deleted lines make grug brain dev happy.  The resource initialization pattern without providing the redundant helper methods makes a ton of sense.

crates/bevy_ecs/src/schedule/schedule.rs
alice-i-cecile: This particular line of code is in its final form. There's other localized mechanisms of error handling, like system piping, which is what the docs on `FallbackErrorHandling` refer to. When those are used, this error handler is never called at all.

crates/bevy_ecs/src/error/handler.rs
```diff
@@ -10,12 +10,19 @@ pub struct SystemErrorContext {
     pub last_run: Tick,
 }
 
-/// The default systems error handler stored as a resource in the [`World`](crate::world::World).
-pub struct DefaultSystemErrorHandler(pub fn(BevyError, SystemErrorContext));
+/// The error handler of last resort used for [`bevy_ecs::error::Result`]s returned by systems, commands and observers,
```
cart, line 13: This is not currently used by commands (largely for performance reasons).

crates/bevy_ecs/src/error/handler.rs
```diff
@@ -10,12 +10,19 @@ pub struct SystemErrorContext {
...
+/// when an error is not otherwise handled.
+///
+/// This is stored as a resource in the [`World`](crate::world::World),
+/// and defaults to panicking if not set.
+///
+/// See [`bevy_ecs::error`] for more information on error handling,
+/// and [`bevy_ecs::error::handler`] for an assortment of built-in error handlers.
+pub struct FallbackErrorHandler(pub fn(BevyError, SystemErrorContext));
```
cart, line 21: I see the appeal of removing `System`, but as a user if I was going to change this, I would ask myself "how do I set the default bevy error handler". I also think it "fits". People tend to consider defaults "overridable".

krunchington:
> The generated `examples/README.md` is out of sync with the example metadata in `Cargo.toml` or the example readme template. Please run `cargo run -p build-templated-pages -- update examples` to update it, and commit the file change.

@alice-i-cecile I was going to pr another change to run this, but it looks like the table formatting may have changed in an update to the generation script- if this looks intentional I can send a PR for it?
image

alice-i-cecile:
Closing out; going to try moving to the command-based design instead.

Issue #17682, opened 2025-02-04
🐢 Make `!Send` resources `thread_local!`


joshua-holmes:
This is a sub-issue of #17667. See that issue for more high-level details.

## What problem does this solve or what need does it fill?

We are using `!Send` resources to store mostly global data, such as handles, `winit` windows, etc. This blocks the resources-as-components effort, which seeks to eliminate the usage of resources.

## What solution would you like?

In order to unblock resources-as-components, we need to remove any `!Send` components and replace them with the use of `thread_local!`. The benefit of using `thread_local!` in this context is that the `!Send` resources we are working with are (at least sometimes) `!Sync`. Wrapping a var in `thread_local!` ensures safety without requiring the wrapped type to be `Sync` or `Send`.

## What alternative(s) have you considered?

We have talked about:
* Using `static` vars without `thread_local!`, but that requires the var be `Sync`.
* Creating `!Send` `World`s to store this data. This is the long term plan (see parent issue), but right now we just need a short-term solution so resources-as-components is unblocked

PR #18301, opened 2025-03-13
Create new `NonSendMarker`

joshua-holmes:
# Objective

Create new `NonSendMarker` that does not depend on `NonSend`.

Required, in order to accomplish #17682. In that issue, we are trying to replace `!Send` resources with `thread_local!` in order to unblock the resources-as-components effort. However, when we remove all the `!Send` resources from a system, that allows the system to run on a thread other than the main thread, which is against the design of the system. So this marker gives us the control to require a system to run on the main thread without depending on `!Send` resources.

## Solution

Create a new `NonSendMarker` to replace the existing one that does not depend on `NonSend`.

## Testing

Other than running tests, I ran a few examples:
- `window_resizing`
- `wireframe`
- `volumetric_fog` (looks so cool)
- `rotation`
- `button`

There is a Mac/iOS-specific change and I do not have a Mac or iOS device to test it. I am doubtful that it would cause any problems for 2 reasons:
1. The change is the same as the non-wasm change which I did test
2. The Pixel Eagle tests run Mac tests

But it wouldn't hurt if someone wanted to spin up an example that utilizes the `bevy_render` crate, which is where the Mac/iSO change was.

## Migration Guide

If `NonSendMarker` is being used from `bevy_app::prelude::*`, replace it with `bevy_ecs::system::NonSendMarker` or use it from `bevy_ecs::prelude::*`. In addition to that, `NonSendMarker` does not need to be wrapped like so:
```rust
fn my_system(_non_send_marker: Option<NonSend<NonSendMarker>>) {
    ...
}
```

Instead, it can be used without any wrappers:
```rust
fn my_system(_non_send_marker: NonSendMarker) {
    ...
}
```

### DIFF
--- a/crates/bevy_ecs/src/system/system_param.rs
+++ b/crates/bevy_ecs/src/system/system_param.rs
@@ -1421,35 +1421,26 @@ unsafe impl<T: SystemBuffer> SystemParam for Deferred<'_, T> {
 }
 
 /// A dummy type that is [`!Send`](Send), to force systems to run on the main thread.
-pub struct NonSendMarker(PhantomData<*mut ()>);
+pub struct NonSendMarker();
 
-// SAFETY: Implementing SystemParam for a shell type that doesn't read or write any data. No
-// safety issue.
+// SAFETY: No world access.
 unsafe impl SystemParam for NonSendMarker {
-    type State = PhantomData<ComponentId>;
+    type State = ();
     type Item<'w, 's> = Self;
 
     #[inline]
-    fn init_state(_world: &mut World, _system_meta: &mut SystemMeta) -> Self::State {
-        PhantomData
+    fn init_state(_world: &mut World, system_meta: &mut SystemMeta) -> Self::State {
+        system_meta.set_non_send();
     }
 
     #[inline]
-    unsafe fn validate_param(
-        _state: &Self::State,
-        _system_meta: &SystemMeta,
-        _world: UnsafeWorldCell,
-    ) -> bool {
-        true
-    }
-
     unsafe fn get_param<'world, 'state>(
         _state: &'state mut Self::State,
         _system_meta: &SystemMeta,
         _world: UnsafeWorldCell<'world>,
         _change_tick: Tick,
     ) -> Self::Item<'world, 'state> {
-        Self(PhantomData)
+        Self()
     }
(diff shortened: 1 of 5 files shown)


crates/bevy_ecs/src/system/system_param.rs
```diff
@@ -1420,6 +1420,30 @@ unsafe impl<T: SystemBuffer> SystemParam for Deferred<'_, T> {
     }
 }
 
+/// A dummy type that is [`!Send`](Send), to force systems to run on the main thread.
+pub struct NonSendMarker();
```
chescock, line 1424: This could be a little simpler as a unit struct instead of a tuple struct:
```suggestion
pub struct NonSendMarker;
```

crates/bevy_ecs/src/system/system_param.rs
```diff
@@ -1420,6 +1420,30 @@ unsafe impl<T: SystemBuffer> SystemParam for Deferred<'_, T> {
...
+    }
+
+    #[inline]
+    unsafe fn get_param<'world, 'state>(
+        _state: &'state mut Self::State,
+        _system_meta: &SystemMeta,
+        _world: UnsafeWorldCell<'world>,
+        _change_tick: Tick,
+    ) -> Self::Item<'world, 'state> {
+        Self()
```
chescock, line 1443: (and here, if you take the suggestion above)

```suggestion
        Self
```

crates/bevy_ecs/src/system/system_param.rs
```diff
@@ -1420,6 +1420,30 @@ unsafe impl<T: SystemBuffer> SystemParam for Deferred<'_, T> {
...
+    }
+}
+
```
chescock, line 1446: Might as well `impl ReadOnlySystemParam`, too: 

```suggestion

// SAFETY: Does not read any world state
unsafe impl ReadOnlySystemParam for NonSendMarker {}

```

chescock (approved):
Looks good!

crates/bevy_ecs/src/system/system_param.rs
joshua-holmes: done!

maniwani (approved):

PR #18307, opened 2025-03-14
bevy_reflect: Add clone registrations project-wide

MrGVSV:
# Objective

Now that #13432 has been merged, it's important we update our reflected types to properly opt into this feature. If we do not, then this could cause issues for users downstream who want to make use of reflection-based cloning.

## Solution

This PR is broken into 4 commits:

1. Add `#[reflect(Clone)]` on all types marked `#[reflect(opaque)]` that are also `Clone`. This is mandatory as these types would otherwise cause the cloning operation to fail for any type that contains it at any depth.
2. Update the reflection example to suggest adding `#[reflect(Clone)]` on opaque types.
3. Add `#[reflect(clone)]` attributes on all fields marked `#[reflect(ignore)]` that are also `Clone`. This prevents the ignored field from causing the cloning operation to fail.
   
   Note that some of the types that contain these fields are also `Clone`, and thus can be marked `#[reflect(Clone)]`. This makes the `#[reflect(clone)]` attribute redundant. However, I think it's safer to keep it marked in the case that the `Clone` impl/derive is ever removed. I'm open to removing them, though, if people disagree.
4. Finally, I added `#[reflect(Clone)]` on all types that are also `Clone`. While not strictly necessary, it enables us to reduce the generated output since we can just call `Clone::clone` directly instead of calling `PartialReflect::reflect_clone` on each variant/field. It also means we benefit from any optimizations or customizations made in the `Clone` impl, including directly dereferencing `Copy` values and increasing reference counters.

   Along with that change I also took the liberty of adding any missing registrations that I saw could be applied to the type as well, such as `Default`, `PartialEq`, and `Hash`. There were hundreds of these to edit, though, so it's possible I missed quite a few.

That last commit is **_massive_**. There were nearly 700 types to update. So it's recommended to review the first three before moving onto that last one.

Additionally, I can break the last commit off into its own PR or into smaller PRs, but I figured this would be the easiest way of doing it (and in a timely manner since I unfortunately don't have as much time as I used to for code contributions).

## Testing

You can test locally with a `cargo check`:

```
cargo check --workspace --all-features
```

### DIFF
--- a/crates/bevy_a11y/src/lib.rs
+++ b/crates/bevy_a11y/src/lib.rs
@@ -54,7 +54,11 @@ pub struct ActionRequest(pub accesskit::ActionRequest);
 /// Useful if a third-party plugin needs to conditionally integrate with
 /// `AccessKit`
 #[derive(Resource, Default, Clone, Debug, Deref, DerefMut)]
-#[cfg_attr(feature = "bevy_reflect", derive(Reflect), reflect(Default, Resource))]
+#[cfg_attr(
+    feature = "bevy_reflect",
+    derive(Reflect),
+    reflect(Default, Clone, Resource)
+)]
 pub struct AccessibilityRequested(Arc<AtomicBool>);
 
 impl AccessibilityRequested {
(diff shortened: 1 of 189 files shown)


bushrat011899 (approved):
Good bit of maintenance. A Bevy CLI tool to automatically lint against missing reflected traits can't come soon enough!

MrGVSV:
> Good bit of maintenance. A Bevy CLI tool to automatically lint against missing reflected traits can't come soon enough!

Yep I was wanting to work on one. We'll see if I can do that before 0.17 is out 😅

Issue #18094, opened 2025-02-28
Error logged when running 3d examples


rparrett:
## Bevy version

main, bisected to #16782

## Relevant system information

SystemInfo { os: "macOS 15.3.1 Sequoia", kernel: "24.3.0", cpu: "Apple M4 Max", core_count: "16", memory: "64.0 GiB" }    
AdapterInfo { name: "Apple M4 Max", vendor: 0, device: 0, device_type: IntegratedGpu, driver: "", driver_info: "", backend: Metal }

## What you did

`cargo run --example 3d_scene`
`cargo run --example transmission`

## What went wrong

Example appears to work, but emits the following errors once:

```
ERROR bevy_core_pipeline::core_3d::main_transmissive_pass_3d_node: Error encountered while rendering the transmissive phase InvalidViewQuery
ERROR bevy_core_pipeline::core_3d::main_opaque_pass_3d_node: Error encountered while rendering the opaque phase InvalidViewQuery
```

JMS55:
Also happens with 3d_scene for me on my igpu.

hukasu:
on `deferred_rendering`
```
2025-03-01T12:43:41.959961Z  INFO bevy_diagnostic::system_information_diagnostics_plugin::internal: SystemInfo { os: "Linux (NixOS 24.11)", kernel: "6.6.69", cpu: "Intel(R) Core(TM) i7-14700KF", core_count: "20", memory: "31.2 GiB" }    
2025-03-01T12:43:42.139316Z  INFO bevy_render::renderer: AdapterInfo { name: "NVIDIA GeForce RTX 4060 Ti", vendor: 4318, device: 10243, device_type: DiscreteGpu, driver: "NVIDIA", driver_info: "565.77", backend: Vulkan }
2025-03-01T12:43:42.707485Z  WARN bevy_pbr::render::mesh_view_bindings: Too many textures in mesh pipeline view layout, this might cause us to hit `wgpu::Limits::max_sampled_textures_per_shader_stage` in some environments.
2025-03-01T12:43:43.592486Z ERROR bevy_core_pipeline::core_3d::main_opaque_pass_3d_node: Error encountered while rendering the opaque phase InvalidViewQuery
```

hukasu:
???

https://github.com/user-attachments/assets/54ec41b7-fc8a-4274-92e2-58c99442355b

rparrett:
[...]

PR #18308, opened 2025-03-14
Add missing system ordering constraint to prepare_lights

JMS55:
Fix https://github.com/bevyengine/bevy/issues/18094.

### DIFF
--- a/crates/bevy_pbr/src/lib.rs
+++ b/crates/bevy_pbr/src/lib.rs
@@ -133,11 +133,9 @@ use bevy_render::{
     camera::{sort_cameras, CameraUpdateSystem, Projection},
     extract_component::ExtractComponentPlugin,
     extract_resource::ExtractResourcePlugin,
-    render_asset::prepare_assets,
     render_graph::RenderGraph,
     render_resource::Shader,
     sync_component::SyncComponentPlugin,
-    texture::GpuImage,
     view::VisibilitySystems,
     ExtractSchedule, Render, RenderApp, RenderDebugFlags, RenderSet,
 };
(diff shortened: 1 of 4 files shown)


hukasu (approved):

JMS55:
Need to also remove the prepare_assets constraint, no longer needed.

mockersf (approved):

crates/bevy_pbr/src/lib.rs
```diff
@@ -130,14 +130,12 @@ use bevy_ecs::prelude::*;
...
     alpha::AlphaMode,
-    camera::{CameraUpdateSystem, Projection},
+    camera::{sort_cameras, CameraUpdateSystem, Projection},
     extract_component::ExtractComponentPlugin,
     extract_resource::ExtractResourcePlugin,
-    render_asset::prepare_assets,
     render_graph::RenderGraph,
     render_resource::Shader,
     sync_component::SyncComponentPlugin,
-    texture::GpuImage,
```
mockersf, line 140: you should keep those two import

JMS55: Why? They're not used in the file anywhere...

JMS55:
I have literately _no clue_ what's going on with the imports, but I added them to the files missing them and it compiles now.

crates/bevy_pbr/src/lib.rs
mockersf: they were imported in the lib.rs file which was putting it in scope for other files

IceSentry (approved):

hukasu:
confirmed that the errors are no longer logged

PR #18309, opened 2025-03-14
Required components accept const values (#16720)

### LINKED ISSUE #16720: Required components should accept constant and enum values of the appropriate type

When defining a component, we can specify that certain other components are required. Ordinarily, this inserts them with their default value.

```rust
#[derive(Component)]
#[require(
    LockedAxes(||LockedAxes::ROTATION_LOCKED),
)]
pub struct CharacterController;
```

If we want to override that, we can pass any function which takes no argument, and returns the appropriate component type.

```rust
#[derive(Component)]
#[require(
    LockedAxes(||LockedAxes::ROTATION_LOCKED),
)]
pub struct CharacterController;
```

However, this is verbose / frustrating when working with constants.

Users should be able to pass in a constant, like so:

```rust
#[derive(Component)]
#[require(
  LockedAxes::ROTATION_LOCKED,
)]
pub struct CharacterController;
```

I suspect that we'll need to generate a closure that constructs the constant, but I haven't looked closely at the code here. We will need to require `Clone` unless we're storing a closure.

If, for macro reasons, we absolutely need to pass in the type before the constructor, we could do:

```rust
#[derive(Component)]
#[require(
    LockedAxes(LockedAxes::ROTATION_LOCKED),
)]
pub struct CharacterController;
```

## Additional context

We might also want to provide helpers for doing this for dynamically registered components.

Wuketuke:
# Objective

Const values should be more ergonomic to insert, since this is too verbose
``` rust
#[derive(Component)]
#[require(
    LockedAxes(||LockedAxes::ROTATION_LOCKED),
)]
pub struct CharacterController;
```
instead, users can now abbreviate that nonsense like this
``` rust
#[derive(Component)]
#[require(
    LockedAxes = ROTATION_LOCKED),
)]
pub struct CharacterController;
```
it also works for enum labels.
I chose to omit the type, since were trying to reduce typing here. The alternative would have been this:
```rust
#[require(
    LockedAxes = LockedAxes::ROTATION_LOCKED),
)]
```
This of course has its disadvantages, since the const has to be associated, but the old closure method is still possible, so I dont think its a problem.
- Fixes #16720

## Testing

I added one new test in the docs, which also explain the new change. I also saw that the docs for the required components on line 165 was missing an assertion, so I added it back in

### DIFF
--- a/crates/bevy_ecs/macros/src/component.rs
+++ b/crates/bevy_ecs/macros/src/component.rs
@@ -222,6 +222,16 @@ pub fn derive_component(input: TokenStream) -> TokenStream {
                         );
                     });
                 }
+                Some(RequireFunc::Enum(label)) => {
+                    register_required.push(quote! {
+                        components.register_required_components_manual::<Self, #ident>(
+                            required_components,
+                            || #ident::#label,
+                            inheritance_depth,
+                            recursion_check_stack
+                        );
+                    });
+                }
                 None => {
                     register_required.push(quote! {
                         components.register_required_components_manual::<Self, #ident>(
@@ -505,6 +515,7 @@ struct Require {
 enum RequireFunc {
     Path(Path),
     Closure(ExprClosure),
+    Enum(Ident),
 }
 
 struct Relationship {
@@ -611,6 +622,10 @@ impl Parse for Require {
                 let func = content.parse::<Path>()?;
                 Some(RequireFunc::Path(func))
             }
+        } else if input.peek(Token![=]) {
+            let _t: Token![=] = input.parse()?;
+            let label = input.parse::<Ident>()?;
+            Some(RequireFunc::Enum(label))
         } else {
             None
         };
--- a/crates/bevy_ecs/macros/src/component.rs
+++ b/crates/bevy_ecs/macros/src/component.rs
@@ -222,16 +222,6 @@ pub fn derive_component(input: TokenStream) -> TokenStream {
                         );
                     });
                 }
-                Some(RequireFunc::Enum(label)) => {
-                    register_required.push(quote! {
-                        components.register_required_components_manual::<Self, #ident>(
-                            required_components,
-                            || #ident::#label,
-                            inheritance_depth,
-                            recursion_check_stack
-                        );
-                    });
-                }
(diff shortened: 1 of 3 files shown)


crates/bevy_ecs/macros/src/component.rs
```diff
@@ -611,6 +611,12 @@ impl Parse for Require {
                 let func = content.parse::<Path>()?;
                 Some(RequireFunc::Path(func))
             }
+        } else if input.peek(Token![=]) {
+            let _t: syn::Token![=] = input.parse()?;
+            let label: Ident = input.parse()?;
```
Brezak, line 616: We could parse out an expression instead and support more ways to set the value.

Wuketuke: yes, but you would somehow need to extract the type from the expression, so it will be messy

alice-i-cecile (approved):
A bit implicit but I think it's worth it. Using constants here should be dead easy.

cart:
I would have preferred the less implicit:

```rust
#[derive(Component)]
#[require(
    LockedAxes = LockedAxes::ROTATION_LOCKED),
)]
pub struct CharacterController;
```

That would also support this pattern:

```rust
const CHARACTER_CONTROLLER_LOCK_AXES: LockedAxes = LockedAxes::from_bits(...);

#[derive(Component)]
#[require(
    LockedAxes = CHARACTER_CONTROLLER_LOCKED_AXES,
)]
pub struct CharacterController;
```

The current implementation is unintuitive and will be confusingly broken when people try to do something like the above example.

cart:
Actually I think we should go even further and do this:

```rust
require(
   Transform, // becomes Transform::default
   LockedAxes::ROTATION_LOCKED, // becomes || LockedAxes::ROTATION_LOCKED
   Health(10), // becomes || Health(10)
   Defense = || Defense(8), // becomes  || Defense(8)
   Speed = my_speed, // becomes my_speed
   Position { x: 10, y: 20 }, // becomes || Position { x: 10, y: 20 }
)
```

This would be a breaking change, but I think its what pretty much everyone actually wants. I'll take a stab at this

PR #18311, opened 2025-03-14
Derive Macro for MapEntites (#17611)

### LINKED ISSUE #17611: Derive Macro for MapEntites

Would remove boilerplate like
```rust
impl MapEntities for Foo {
    fn map_entities<M: EntityMapper>(&mut self, entity_mapper: &mut M) {
        self.a.map_entities(entity_mapper);
        self.b.map_entities(entity_mapper);
    }
}
```

A derive macro which can implement MapEntities for any struct where all its attributes implement MapEntities (like Debug and Default do).

Just implementing it.

Wuketuke:
# Objective
instead of writing boilerplate like this
```rust
impl MapEntities for Foo {
    fn map_entities<M: EntityMapper>(&mut self, entity_mapper: &mut M) {
        self.a.map_entities(entity_mapper);
        self.b.map_entities(entity_mapper);
    }
}
```
users can now use the derive macro `MapEntities`.
If any fields dont implement that trait, they can be skipped with a `#[skip_mapping]` attribute.
```rust
#[derive(MapEntities)]
struct Foo {
    a: A,
    b: B,
    #[skip_mapping]
    c: C, // C doesnt implement MapEntities, so we need to skip it
}
```
Fixes #17611

## Solution

My implementation is a heavily simplified version of @Liam19's version, as I didnt want to add too much complexity at once, and I am not familiar with that trait. So feedback is appreciated.
Since its possible that people want to derive enums as well, I have put the derive function in its own file, so that it has space to grow

## Testing

Theres now an example in the documentation

### DIFF
--- a/crates/bevy_ecs/macros/src/map_entities.rs
+++ b/crates/bevy_ecs/macros/src/map_entities.rs
@@ -17,28 +17,28 @@ fn map_struct(name: syn::Ident, fields: &Fields) -> proc_macro2::TokenStream {
     let mut map_entities = vec![];
 
     'fields: for (i, field) in fields.iter().enumerate() {
-        // skip that field
+        // check for skipping fields
         for attr in &field.attrs {
             if attr.path().is_ident("skip_mapping") {
                 match &attr.meta {
+                    // skip that fields
                     syn::Meta::Path(_) => continue 'fields,
+                    // the attribute has been used incorrectly
                     _ => panic!("just use the bare `#[skip_mapping]`"),
                 }
             }
         }
 
         let ty = &field.ty;
-        let map_field = if let Some(field_name) = &field.ident {
-            // Named field (struct)
-            quote! {
-                <#ty as bevy_ecs::entity::MapEntities>::map_entities(&mut self.#field_name, entity_mapper);
-            }
-        } else {
+        let member = match &field.ident {
+            // a named field (struct)
+            Some(field_name) => syn::Member::Named(field_name.clone()),
             // Unnamed field (tuple-like struct)
-            let idx = syn::Index::from(i);
-            quote! {
-                <#ty as bevy_ecs::entity::MapEntities>::map_entities(&mut self.#idx, entity_mapper);
-            }
+            None => syn::Member::Unnamed(syn::Index::from(i)),
+        };
+
+        let map_field = quote! {
+            <#ty as bevy_ecs::entity::MapEntities>::map_entities(&mut self.#member, entity_mapper);
         };
 
         map_entities.push(map_field);
--- a/crates/bevy_ecs/macros/src/map_entities.rs
+++ b/crates/bevy_ecs/macros/src/map_entities.rs
@@ -8,7 +8,8 @@ pub(super) fn derive_map_entities(input: TokenStream) -> TokenStream {
 
     match input.data {
         Data::Struct(data_struct) => map_struct(name, &data_struct.fields),
-        _ => unimplemented!(),
+        Data::Enum(data_enum) => map_enum(name, &data_enum.variants),
+        Data::Union(_) => panic!("MapEntities is only valid on structs or enums"),
     }
     .into()
 }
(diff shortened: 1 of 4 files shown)


crates/bevy_ecs/src/entity/map_entities.rs
```diff
@@ -44,6 +44,37 @@ use super::{hash_map::EntityHashMap, VisitEntitiesMut};
...
+///
+/// #[derive(MapEntities)]
+/// struct MyStruct {
+///     a: A,
+///     #[skip_mapping]
+///     b: B,
+/// }
+///
+/// # #[derive(MapEntities)]
+/// # struct TestTuple(A);
```
waynemwashuma, line 65: An extra `#` here

Wuketuke: to eliminate that double newline?

crates/bevy_ecs/macros/src/map_entities.rs
```diff
@@ -0,0 +1,54 @@
...
+    let mut map_entities = vec![];
+
+    'fields: for (i, field) in fields.iter().enumerate() {
+        // skip that field
+        for attr in &field.attrs {
+            if attr.path().is_ident("skip_mapping") {
+                match &attr.meta {
+                    syn::Meta::Path(_) => continue 'fields,
+                    _ => panic!("just use the bare `#[skip_mapping]`"),
+                }
```
Bleachfuel, line 26: just continue without checking! This is unnessesary code

crates/bevy_ecs/macros/src/map_entities.rs
```diff
@@ -0,0 +1,54 @@
...
+        Data::Struct(data_struct) => map_struct(name, &data_struct.fields),
+        _ => unimplemented!(),
+    }
+    .into()
+}
+
+fn map_struct(name: syn::Ident, fields: &Fields) -> proc_macro2::TokenStream {
+    let mut map_entities = vec![];
+
+    'fields: for (i, field) in fields.iter().enumerate() {
```
Bleachfuel, line 19:
```suggestion
    for (i, field) in fields.iter().enumerate() {
```

crates/bevy_ecs/macros/src/map_entities.rs
Bleachfuel, line 26:
```suggestion
continue
```

crates/bevy_ecs/macros/src/map_entities.rs
```diff
@@ -0,0 +1,54 @@
...
+            }
+        } else {
+            // Unnamed field (tuple-like struct)
+            let idx = syn::Index::from(i);
+            quote! {
+                <#ty as bevy_ecs::entity::MapEntities>::map_entities(&mut self.#idx, entity_mapper);
+            }
+        };
+
+        map_entities.push(map_field);
```
Bleachfuel, line 44:
```suggestion
    let member = as_member(&field.ident, i);

    map_entities.push(quote! {
        <#ty as bevy_ecs::entity::MapEntities>::map_entities(&mut self.#member, entity_mapper);
    });

```

Bleachfuel (changes requested):

crates/bevy_ecs/macros/src/map_entities.rs
Wuketuke: the label is needed here, because later, we iterate over all attributes to find a "skip_mapping".
we could use an `attr.iter().find(|attr| attr.path().is_ident("skip_mapping")` to get rid of the label, but that hasnt made the code any easier to follow

crates/bevy_ecs/macros/src/map_entities.rs
Wuketuke: here i check that the user has really put a `#[skip_mapping]` and not `#[skip_mapping(something random)]`
it might not be necessary, but i prefer this tool to tell the user exactly how it wants to be used

Wuketuke:
i revisited this pr, and when merging from main, i noticed that this pr has already been made obsolete by #18432

PR #18313, opened 2025-03-14
Prefer `core::time::Duration` over `std::time::Duration`

mnmaita:
# Objective

- While working on a bevy 0.15 project and then checking the current state of the `main` bevy branch to see why my autocomplete wasn't picking up the `bevy::utils::Duration` export I noticed that I couldn't find it anymore. As far as I can see https://github.com/bevyengine/bevy/pull/17158 removed this re-export, yet some parts of the code still use `std::time::Duration`. To standardize this I'm replacing all instances of `std::time::Duration` in favor of `core::time::Duration`. These are mostly example code and comments for what is worth.

## Solution

- Replaced all instances of `std::time::Duration` in favor of `core::time::Duration`.

## Testing

- I haven't tested this changeset thoroughly as I think its impact should be negligible. Might be good to perform an example run on CI though.

### DIFF
--- a/benches/benches/bevy_picking/ray_mesh_intersection.rs
+++ b/benches/benches/bevy_picking/ray_mesh_intersection.rs
@@ -1,5 +1,5 @@
 use core::hint::black_box;
-use std::time::Duration;
+use core::time::Duration;
 
 use benches::bench;
 use bevy_math::{Dir3, Mat4, Ray3d, Vec3};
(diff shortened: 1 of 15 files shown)


alice-i-cecile:
Generally we don't worry about no_std in examples / test / benchmarks, which is why these are still around.

bushrat011899 (approved):
Personally, I think there's no harm here and it is more consistent with how the Bevy crates are written now. _However_, I did propose this before and @mockersf raised a good point that this is not how the majority of Bevy _users_ will write code, so it negatively affects the utility of the examples.

To be clear, _I_ support this change, but this is at the very least controversial.

PR #18315, opened 2025-03-14
implement get_many_unique

Victoronz:
# Objective

Continuation to #16547 and #17954.

The `get_many` family are the last methods on `Query`/`QueryState` for which we're still missing a `unique` version.

## Solution

Offer `get_many_unique`/`get_many_unique_mut` and `get_many_unique_inner`!

Their implementation is the same as `get_many`, the difference lies in their guaranteed-to-be unique inputs, meaning we never do any aliasing checks.

To reduce confusion, we also rename `get_many_readonly` into `get_many_inner` and the current `get_many_inner` into `get_many_mut_inner` to clarify their purposes.

## Testing

Doc examples.

## Migration Guide

`get_many_inner` is now called `get_many_mut_inner`.
`get_many_readonly` is now called `get_many_inner`.

### DIFF
--- a/crates/bevy_ecs/src/query/state.rs
+++ b/crates/bevy_ecs/src/query/state.rs
@@ -1,7 +1,7 @@
 use crate::{
     archetype::{Archetype, ArchetypeComponentId, ArchetypeGeneration, ArchetypeId},
     component::{ComponentId, Tick},
-    entity::{Entity, EntityBorrow, EntitySet},
+    entity::{unique_array::UniqueEntityArray, Entity, EntityBorrow, EntitySet},
     entity_disabling::DefaultQueryFilters,
     prelude::FromWorld,
     query::{Access, FilteredAccess, QueryCombinationIter, QueryIter, QueryParIter, WorldQuery},
@@ -997,6 +997,44 @@ impl<D: QueryData, F: QueryFilter> QueryState<D, F> {
         self.query(world).get_many_inner(entities)
     }
 
+    /// Returns the read-only query results for the given [`UniqueEntityArray`].
+    ///
+    /// In case of a nonexisting entity or mismatched component, a [`QueryEntityError`] is
+    /// returned instead.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use bevy_ecs::{prelude::*, query::QueryEntityError, entity::{EntitySetIterator, unique_array::UniqueEntityArray, unique_vec::UniqueEntityVec}};
+    ///
+    /// #[derive(Component, PartialEq, Debug)]
+    /// struct A(usize);
+    ///
+    /// let mut world = World::new();
+    /// let entity_vec: UniqueEntityVec<Entity> = world.spawn_batch((0..3).map(A)).collect_set();
+    /// let entities: UniqueEntityArray<Entity, 3> = entity_vec.try_into().unwrap();
+    ///
+    /// world.spawn(A(73));
+    ///
+    /// let mut query_state = world.query::<&A>();
+    ///
+    /// let component_values = query_state.get_many_unique(&world, entities).unwrap();
+    ///
+    /// assert_eq!(component_values, [&A(0), &A(1), &A(2)]);
+    ///
+    /// let wrong_entity = Entity::from_raw(365);
+    ///
+    /// assert_eq!(match query_state.get_many_unique(&mut world, UniqueEntityArray::from([wrong_entity])).unwrap_err() {QueryEntityError::EntityDoesNotExist(error) => error.entity, _ => panic!()}, wrong_entity);
+    /// ```
+    #[inline]
+    pub fn get_many_unique<'w, const N: usize>(
+        &mut self,
+        world: &'w World,
+        entities: UniqueEntityArray<Entity, N>,
+    ) -> Result<[ROQueryItem<'w, D>; N], QueryEntityError> {
+        self.query(world).get_many_unique_inner(entities)
+    }
+
     /// Gets the query result for the given [`World`] and [`Entity`].
     ///
     /// This is always guaranteed to run in `O(1)` time.
@@ -1053,7 +1091,52 @@ impl<D: QueryData, F: QueryFilter> QueryState<D, F> {
         world: &'w mut World,
         entities: [Entity; N],
     ) -> Result<[D::Item<'w>; N], QueryEntityError> {
-        self.query_mut(world).get_many_inner(entities)
+        self.query_mut(world).get_many_mut_inner(entities)
+    }
+
+    /// Returns the query results for the given [`UniqueEntityArray`].
+    ///
+    /// In case of a nonexisting entity or mismatched component, a [`QueryEntityError`] is
+    /// returned instead.
+    ///
+    /// ```
+    /// use bevy_ecs::{prelude::*, query::QueryEntityError, entity::{EntitySetIterator, unique_array::UniqueEntityArray, unique_vec::UniqueEntityVec}};
+    ///
+    /// #[derive(Component, PartialEq, Debug)]
+    /// struct A(usize);
+    ///
+    /// let mut world = World::new();
+    ///
+    /// let entities: UniqueEntityVec<Entity> = world.spawn_batch((0..3).map(A)).collect_set();
+    /// let entities: UniqueEntityArray<Entity, 3> = entities.try_into().unwrap();
+    ///
+    /// world.spawn(A(73));
+    ///
+    /// let mut query_state = world.query::<&mut A>();
+    ///
+    /// let mut mutable_component_values = query_state.get_many_unique_mut(&mut world, entities).unwrap();
+    ///
+    /// for mut a in &mut mutable_component_values {
+    ///     a.0 += 5;
+    /// }
+    ///
+    /// let component_values = query_state.get_many_unique(&world, entities).unwrap();
+    ///
+    /// assert_eq!(component_values, [&A(5), &A(6), &A(7)]);
+    ///
+    /// let wrong_entity = Entity::from_raw(57);
+    /// let invalid_entity = world.spawn_empty().id();
+    ///
+    /// assert_eq!(match query_state.get_many_unique(&mut world, UniqueEntityArray::from([wrong_entity])).unwrap_err() {QueryEntityError::EntityDoesNotExist(error) => error.entity, _ => panic!()}, wrong_entity);
+    /// assert_eq!(match query_state.get_many_unique_mut(&mut world, UniqueEntityArray::from([invalid_entity])).unwrap_err() {QueryEntityError::QueryDoesNotMatch(entity, _) => entity, _ => panic!()}, invalid_entity);
+    /// ```
+    #[inline]
+    pub fn get_many_unique_mut<'w, const N: usize>(
+        &mut self,
+        world: &'w mut World,
+        entities: UniqueEntityArray<Entity, N>,
+    ) -> Result<[D::Item<'w>; N], QueryEntityError> {
+        self.query_mut(world).get_many_unique_inner(entities)
     }
 
     /// Gets the query result for the given [`World`] and [`Entity`].
--- a/crates/bevy_ecs/src/system/query.rs
+++ b/crates/bevy_ecs/src/system/query.rs
@@ -1,7 +1,9 @@
 use crate::{
     batching::BatchingStrategy,
     component::Tick,
-    entity::{Entity, EntityBorrow, EntityDoesNotExistError, EntitySet},
+    entity::{
+        unique_array::UniqueEntityArray, Entity, EntityBorrow, EntityDoesNotExistError, EntitySet,
+    },
     query::{
         DebugCheckedUnwrap, NopWorldQuery, QueryCombinationIter, QueryData, QueryEntityError,
         QueryFilter, QueryIter, QueryManyIter, QueryManyUniqueIter, QueryParIter, QueryParManyIter,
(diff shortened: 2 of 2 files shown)


crates/bevy_ecs/src/query/state.rs
```diff
@@ -997,6 +997,44 @@ impl<D: QueryData, F: QueryFilter> QueryState<D, F> {
...
+    /// # Examples
+    ///
+    /// ```
+    /// use bevy_ecs::{prelude::*, query::QueryEntityError, entity::{EntitySetIterator, unique_array::UniqueEntityArray, unique_vec::UniqueEntityVec}};
+    ///
+    /// #[derive(Component, PartialEq, Debug)]
+    /// struct A(usize);
+    ///
+    /// let mut world = World::new();
+    /// let entity_vec: UniqueEntityVec<Entity> = world.spawn_batch((0..3).map(A)).collect_set();
```
Carter0, line 1014: Two quick questions if you have the time

1. Isn't a `UniqueEntityVec<Entity>` kinda redundant? What else would go between the <>? 
2. Isn't a `UniqueEntityVec` just a Set? Why not just call it a set?

Victoronz: Do you mean the questions in general? Or just in this specific case?
If its just these lines:
1. You're right here, this can probably just be inferred.
2. The variable name is more descriptive as `entity_set`, yeah!

crates/bevy_ecs/src/query/state.rs
```diff
@@ -997,6 +997,44 @@ impl<D: QueryData, F: QueryFilter> QueryState<D, F> {
...
+    /// let component_values = query_state.get_many_unique(&world, entities).unwrap();
+    ///
+    /// assert_eq!(component_values, [&A(0), &A(1), &A(2)]);
+    ///
+    /// let wrong_entity = Entity::from_raw(365);
+    ///
+    /// assert_eq!(match query_state.get_many_unique(&mut world, UniqueEntityArray::from([wrong_entity])).unwrap_err() {QueryEntityError::EntityDoesNotExist(error) => error.entity, _ => panic!()}, wrong_entity);
+    /// ```
+    #[inline]
+    pub fn get_many_unique<'w, const N: usize>(
```
chescock, line 1030: I'm hoping to deprecate the methods like this on `QueryState`, so I'd mildly prefer not to add these in the first place.  The functionality would still be available as `.query(world).get_many_unique_inner(entities)`.

crates/bevy_ecs/src/query/state.rs
```diff
@@ -1053,7 +1091,52 @@ impl<D: QueryData, F: QueryFilter> QueryState<D, F> {
...
+    ///
+    /// #[derive(Component, PartialEq, Debug)]
+    /// struct A(usize);
+    ///
+    /// let mut world = World::new();
+    ///
+    /// let entities: UniqueEntityVec<Entity> = world.spawn_batch((0..3).map(A)).collect_set();
+    /// let entities: UniqueEntityArray<Entity, 3> = entities.try_into().unwrap();
+    ///
+    /// world.spawn(A(73));
```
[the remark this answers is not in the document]
Victoronz: I have copied and adjusted the existing doc tests for `get_many` and `get_many_mut` for this, so I am not the original writer, but that does seem to be the reason for that spawn.

crates/bevy_ecs/src/query/state.rs
```diff
@@ -1053,7 +1091,52 @@ impl<D: QueryData, F: QueryFilter> QueryState<D, F> {
...
+    ///
+    /// let component_values = query_state.get_many_unique(&world, entities).unwrap();
+    ///
+    /// assert_eq!(component_values, [&A(5), &A(6), &A(7)]);
+    ///
+    /// let wrong_entity = Entity::from_raw(57);
+    /// let invalid_entity = world.spawn_empty().id();
+    ///
+    /// assert_eq!(match query_state.get_many_unique(&mut world, UniqueEntityArray::from([wrong_entity])).unwrap_err() {QueryEntityError::EntityDoesNotExist(error) => error.entity, _ => panic!()}, wrong_entity);
+    /// assert_eq!(match query_state.get_many_unique_mut(&mut world, UniqueEntityArray::from([invalid_entity])).unwrap_err() {QueryEntityError::QueryDoesNotMatch(entity, _) => entity, _ => panic!()}, invalid_entity);
```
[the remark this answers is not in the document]
Victoronz: These asserts test that the error cases are produced properly.
If an entity does not match the `D` in QueryData, that should return an error!
Same with an entity that does not exist in the first place.

crates/bevy_ecs/src/query/state.rs
Victoronz: Nevermind, seems like the `Entity` parameter in 1. doesn't infer!
While `Entity` could be the type default for `UniqueEntityVec`, same with `UniqueEntitySlice`, this isn't easily the case for `UniqueEntityArray`.
Because type defaults have to be trailing, this would turn `UniqueEntityArray<Entity, N>` into `UniqueEntityArray<N, Entity>`, which is quite an ugly inconsistency with `[Entity; N]`.
Not sure what should be done here

Carter0 (approved):
I like it! I would love an example for how to use unique entity lists and why it might be useful!

crates/bevy_ecs/src/system/query.rs
```diff
@@ -1616,6 +1734,25 @@ impl<'w, 's, D: QueryData, F: QueryFilter> Query<'w, 's, D, F> {
...
+    /// In case of a nonexisting entity, duplicate entities or mismatched component, a [`QueryEntityError`] is returned instead.
+    ///
+    /// # See also
+    ///
+    /// - [`get_many_unique`](Self::get_many_unique) to get read-only query items without checking for duplicate entities.
+    /// - [`get_many_unique_mut`](Self::get_many_unique_mut) to get items using a mutable reference.
+    #[inline]
+    pub fn get_many_unique_inner<const N: usize>(
+        self,
+        entities: UniqueEntityArray<Entity, N>,
```
chescock, line 1750: Would it make sense to generalize this from `Entity` to `T: TrustedEntityBorrow`?  (And I suppose the read-only one could be generalized to `T: EntityBorrow`.)

crates/bevy_ecs/src/query/state.rs
Carter0: Thanks for responding! 

Yeah, this is just a naming nit really. Maybe when the 'unique' functionality is all done it could be polished up in the future.

I think the type `UniqueEntityVec<Entity>` could maybe be turned into `Set<Entity>` or even just `EntitySet` at some point. But I would mostly asking just because I was curious about if there was a reason to call it UniqueEntityVec rather than anything else.

Victoronz:
> I like it! I would love an example for how to use unique entity lists and why it might be useful!

Once I am done with the feature work I'll have a look throughout the engine for where they can improve things :)
For now, one example can be found in `propagate_transforms`! (Done in #17840)

crates/bevy_ecs/src/query/state.rs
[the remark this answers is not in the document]
Carter0: Cool beans! I think that an example that showcases this functionality would be good then cause you could comment why `world.spawn(A(73));` was there. 

Totally not blocking tho!

chescock (approved):
Looks good!  

I think `get_many_inner` and `get_many_readonly` are new in 0.16, so if we can get this in before the release is cut then we won't need the migration guide.

crates/bevy_ecs/src/query/state.rs
Carter0: Yeah I was a little confused why this function needed to be implemented for both the state module and query module. It seems like it would only really be needed in the query module

crates/bevy_ecs/src/query/state.rs
chescock: > While `Entity` could be the type default for `UniqueEntityVec`, same with `UniqueEntitySlice`, this isn't easily the case for `UniqueEntityArray`.
> Because type defaults have to be trailing, this would turn `UniqueEntityArray<Entity, N>` into `UniqueEntityArray<N, Entity>`, which is quite an ugly inconsistency with `[Entity; N]`.
> Not sure what should be done here.

Oh, adding defaults there is a good idea!  I expect `Entity` to be the most common type by far.  The inconsistency with `UniqueEntityArray<N, T>` and `[T; N]` is unfortunate, but if the normal case is `UniqueEntityArray<N>` vs `[Entity; N]` then it doesn't seem so bad!

crates/bevy_ecs/src/query/state.rs
chescock: > Yeah I was a little confused why this function needed to be implemented for both the state module and query module.

For context: Until recently, the actual implementation of all of these methods was on `QueryState`, and the `Query` methods just called those.  I added the `QueryState::query` methods and moved the implementations over, with the eventual goal of getting rid of most of the other methods on `QueryState`.

Victoronz: I am aware that we're trying to deprecate them!
However in the spirit of being incremental, I find it more confusing to not add them here when the other methods do not yet have the #[deprecated] moniker. Once we do deprecate them before 0.16, then feel free to remove these again!

Given that sometimes PRs can miss a cut-off, or contributors don't have time before an RC, leaving changes unfinished seems like an uneccessary risk, when both adding and removing is simple implementation-wise.

crates/bevy_ecs/src/system/query.rs
Victoronz: Yeah, we could generalize them!
However, this also goes for a lot more places than just these methods.
The general decision of "We want to accept `EntityBorrow` instead of `Entity` in our APIs" is contentious, so I'd like to leave that for when the `EntitySet`-family of features is done.

crates/bevy_ecs/src/query/state.rs
Victoronz: > > While `Entity` could be the type default for `UniqueEntityVec`, same with `UniqueEntitySlice`, this isn't easily the case for `UniqueEntityArray`.
> > Because type defaults have to be trailing, this would turn `UniqueEntityArray<Entity, N>` into `UniqueEntityArray<N, Entity>`, which is quite an ugly inconsistency with `[Entity; N]`.
> > Not sure what should be done here.
> 
> Oh, adding defaults there is a good idea! I expect `Entity` to be the most common type by far. The inconsistency with `UniqueEntityArray<N, T>` and `[T; N]` is unfortunate, but if the normal case is `UniqueEntityArray<N>` vs `[Entity; N]` then it doesn't seem so bad!

Hmm, one way to potentially remedy the inconsistency is by introducing a different inconsistency, which is also a type default for `N`, f.e. `0`, I haven't thought a lot about that yet, but would you think it makes sense?

crates/bevy_ecs/src/system/query.rs
chescock: > However, this also goes for a lot more places than just these methods.
> The general decision of "We want to accept `EntityBorrow` instead of `Entity` in our APIs" is contentious, so I'd like to leave that for when the `EntitySet`-family of features is done.

Oh, I remember waffling about this when writing #18234.  Checking now, it looks like I left a few as `T: TrustedEntityBorrow`.  Should I switch those back to `Entity`, too?

crates/bevy_ecs/src/query/state.rs
Victoronz: > > in the spirit of being incremental, I find it more confusing to not add them here
> 
> Well, in that case, you forgot `get_many_unique_manual(&self, &World, entities)` and `get_many_unique_unchecked(&mut self, UnsafeWorldCell, entities)` :P

These already don't exist for `get_many`, so I haven't added them here.
These inconsistencies are not nice, but we can at least not make them worse in the interim :P

crates/bevy_ecs/src/system/query.rs
Victoronz: Did you? I don't see them!

The current status quo is:
`iter_many`: takes `IntoIterator<Item: EntityBorrow>`
`iter_many_unique` takes `EntitySet`

The `iter_many` signatures are currently the only signatures where we use `EntityBorrow`, and wherever we use `EntitySet` we have a hidden `TrustedEntityBorrow` bound.
Otherwise, the `EntityBorrow` and `TrustedEntityBorrow` traits shouldn't appear in non-`EntitySet` related methods for now I think.

crates/bevy_ecs/src/query/state.rs
```diff
@@ -997,6 +997,44 @@ impl<D: QueryData, F: QueryFilter> QueryState<D, F> {
...
+    ///
+    /// let wrong_entity = Entity::from_raw(365);
+    ///
+    /// assert_eq!(match query_state.get_many_unique(&mut world, UniqueEntityArray::from([wrong_entity])).unwrap_err() {QueryEntityError::EntityDoesNotExist(error) => error.entity, _ => panic!()}, wrong_entity);
+    /// ```
+    #[inline]
+    pub fn get_many_unique<'w, const N: usize>(
+        &mut self,
+        world: &'w World,
+        entities: UniqueEntityArray<N, Entity>,
```
chescock, line 1033: You left `Entity` written out in two places: 

```suggestion
        entities: UniqueEntityArray<N>,
```

crates/bevy_ecs/src/query/state.rs
```diff
@@ -1053,7 +1091,52 @@ impl<D: QueryData, F: QueryFilter> QueryState<D, F> {
...
+    /// let invalid_entity = world.spawn_empty().id();
+    ///
+    /// assert_eq!(match query_state.get_many_unique(&mut world, UniqueEntityArray::from([wrong_entity])).unwrap_err() {QueryEntityError::EntityDoesNotExist(error) => error.entity, _ => panic!()}, wrong_entity);
+    /// assert_eq!(match query_state.get_many_unique_mut(&mut world, UniqueEntityArray::from([invalid_entity])).unwrap_err() {QueryEntityError::QueryDoesNotMatch(entity, _) => entity, _ => panic!()}, invalid_entity);
+    /// ```
+    #[inline]
+    pub fn get_many_unique_mut<'w, const N: usize>(
+        &mut self,
+        world: &'w mut World,
+        entities: UniqueEntityArray<N, Entity>,
```
chescock, line 1137:
```suggestion
        entities: UniqueEntityArray<N>,
```

chescock (approved):

PR #18317, opened 2025-03-14
Update text2d example to use children macro

krunchington:
# Objective

Contributes to #18238 
Updates the `text2d`, example to use the `children!` macro.

I'm not sure I love the SpawnIter usage here, as I feel the `move` keyword in this case is subtle and error prone for those who lose fights with the borrow checker frequently (like me).  Feedback very much welcome.

## Solution

Updates examples to use the Improved Spawning API merged in https://github.com/bevyengine/bevy/pull/17521

## Testing

  - see above

---

## Showcase

n/a

## Migration Guide

n/a

### DIFF
--- a/examples/2d/text2d.rs
+++ b/examples/2d/text2d.rs
@@ -7,7 +7,6 @@
 
 use bevy::{
     color::palettes::css::*,
-    ecs::spawn::SpawnIter,
     math::ops,
     prelude::*,
     sprite::Anchor,
--- a/examples/2d/text2d.rs
+++ b/examples/2d/text2d.rs
@@ -12,6 +12,7 @@ use bevy::{
     sprite::Anchor,
     text::{FontSmoothing, LineBreak, TextBounds},
 };
+use bevy_ecs::spawn::SpawnIter;
 
 fn main() {
     App::new()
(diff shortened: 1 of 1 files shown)


alice-i-cecile (approved):

examples/2d/text2d.rs
```diff
@@ -118,40 +113,42 @@ fn setup(mut commands: Commands, asset_server: Res<AssetServer>) {
...
-        .with_children(|commands| {
-            for (text_anchor, color) in [
+    commands.spawn((
+        Sprite {
+            color: Color::Srgba(LIGHT_CYAN),
+            custom_size: Some(Vec2::new(10., 10.)),
+            ..Default::default()
+        },
+        Transform::from_translation(250. * Vec3::Y),
+        Children::spawn(SpawnIter(
```
alice-i-cecile, line 123: Yeah, let's just revert this changes here.

alice-i-cecile (changes requested):

examples/2d/text2d.rs
krunchington: Went to rebase and all I had to do was accept from HEAD and it worked great :sweat_smile:

alice-i-cecile (approved):

PR #18319, opened 2025-03-14
add Entity default to the entity set wrappers

Victoronz:
# Objective

Installment of the #16547 series.

The vast majority of uses for these types will be the `Entity` case, so it makes sense for it to be the default.

## Solution

`UniqueEntityVec`, `UniqueEntitySlice`, `UniqueEntityArray` and their helper iterator aliases now have `Entity` as a default.

Unfortunately, this means the the `T` parameter for `UniqueEntityArray` now has to be ordered after the `N` constant, which breaks the consistency to `[T; N]`. 
Same with about a dozen iterator aliases that take some `P`/`F` predicate/function parameter.
This should however be an ergonomic improvement in most cases, so we'll just have to live with this inconsistency.

## Migration Guide

Switch type parameter order for the relevant wrapper types/aliases.

### DIFF
--- a/crates/bevy_ecs/src/entity/unique_array.rs
+++ b/crates/bevy_ecs/src/entity/unique_array.rs
@@ -21,8 +21,7 @@ use alloc::{
 use bevy_platform_support::sync::Arc;
 
 use super::{
-    unique_slice::{self, UniqueEntitySlice},
-    TrustedEntityBorrow, UniqueEntityIter,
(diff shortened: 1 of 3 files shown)


alice-i-cecile (approved):

chescock (approved):

Carter0 (approved):
Thank you!

PR #8471, opened 2023-04-23
CI: Switch to DavidAnson/markdownlint-cli2-action for markdown linting

nicoburns:
# Objective

Speed up markdown linting on CI. 

## Solution

- Switch to `DavidAnson/markdownlint-cli2-action`. This action is markdownlint-specific rather than using a 1.4GB docker image to includes linters for several languages. It's also just plain faster.
- Lint every markdown file rather than only changed files

PR #18320, opened 2025-03-14
update superlinter

mockersf:
# Objective

- super-linter uses an old version in CI and wasn't updated for the last two years

## Solution

- Update super-linter


mnmaita (approved):
@mockersf do you think it would be worthwhile to adopt and update [this](https://github.com/bevyengine/bevy/pull/8471/files) as a follow-up?

PR #18321, opened 2025-03-15
Remove system set indirection on `TransformPlugin`

hukasu:
# Objective

Remove system set indirection from `TransformPlugin`

## Solution

Change uses of `PropagateTransformSet` with `TransformSystem::TransformPropagate`.

## Reasoning

This indirection does not seem to be needed anymore, on an older commit there were systems that were added to `PropagateTransformSet` and others to `TransformSystem::TransformPropagete`, and using `ambiguoes_with` to prevent the 2 sets from running in parallel, but now all systems are chained, removing the need to mark the systems as ambiguous with each other. 

## Testing

`cargo run -p ci` and `many_foxes` example

### DIFF
--- a/crates/bevy_transform/src/plugins.rs
+++ b/crates/bevy_transform/src/plugins.rs
@@ -15,45 +15,34 @@ pub struct TransformPlugin;
 
 impl Plugin for TransformPlugin {
     fn build(&self, app: &mut App) {
-        // A set for `propagate_transforms` to mark it as ambiguous with `sync_simple_transforms`.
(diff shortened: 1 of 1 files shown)


alice-i-cecile (approved):

hukasu:
welp, commit from after my commit has changed to use the internal system set again

kristoff3r:
> welp, #18363 has changed to use the internal system set again

That PR will probably be revived with a small bug fix according to https://github.com/bevyengine/bevy/issues/18365#issuecomment-2730563079

hukasu:
it is best if they do this change while remaking the fix

PR #18326, opened 2025-03-15
Use 4-byte LightmapSlabIndex for batching instead of 16-byte AssetId<Image>

superdump:
Less data accessed and compared gives better batching performance.

# Objective

- Use a smaller id to represent the lightmap in batch data to enable a faster implementation of draw streams.
- Improve batching performance for 3D sorted render phases.

## Solution

- 3D batching can use `LightmapSlabIndex` (a `NonMaxU32` which is 4 bytes) instead of the lightmap `AssetId<Image>` (an enum where the largest variant is a 16-byte UUID) to discern the ability to batch.

## Testing

Tested main (yellow) vs this PR (red) on an M4 Max using the `many_cubes` example with `WGPU_SETTINGS_PRIO=webgl2` to avoid GPU-preprocessing, and modifying the materials in `many_cubes` to have `AlphaMode::Blend` so that they would rely on the less efficient sorted render phase batching.
Screenshot_2025-03-15_at_12 17 21
A 44.75us or 7.5% reduction in median execution time of the batch and prepare sorted render phase system for the `Transparent3d` phase (handling 160k cubes).

---

## Migration Guide

- Changed: `RenderLightmap::new()` no longer takes an `AssetId<Image>` argument for the asset id of the lightmap image.

### DIFF
--- a/crates/bevy_pbr/src/lightmap/mod.rs
+++ b/crates/bevy_pbr/src/lightmap/mod.rs
@@ -116,9 +116,6 @@ pub struct Lightmap {
 /// There is one of these per visible lightmapped mesh instance.
 #[derive(Debug)]
 pub(crate) struct RenderLightmap {
-    /// The ID of the lightmap texture.
-    pub(crate) image: AssetId<Image>,
-
     /// The rectangle within the lightmap texture that the UVs are relative to.
     ///
     /// The top left coordinate is the `min` part of the rect, and the bottom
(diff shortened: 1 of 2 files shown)


pcwalton (approved):
LGTM! BTW I think `AssetId` is actually larger than 16 bytes because it doesn't just contain a UUID, it also contains an enum tag.

mockersf (approved):

PR #18329, opened 2025-03-15
Update testbed_ui to use Improved Spawning API

krunchington:
# Objective

Contributes to #18238 
Updates the `text2d`, example to use the `children!` macro.

~~The SpawnIter usage in this example is maybe not the best.  Very open to opinions.  I even left one `with_children` that I thought was just much better than any alternative.~~

## Solution

Updates examples to use the Improved Spawning API merged in https://github.com/bevyengine/bevy/pull/17521

## Testing

  - see above

---

## Showcase

n/a

## Migration Guide

n/a

### DIFF
--- a/examples/testbed/ui.rs
+++ b/examples/testbed/ui.rs
@@ -105,26 +105,24 @@ mod grid {
     pub fn setup(mut commands: Commands) {
         commands.spawn((Camera2d, StateScoped(super::Scene::Grid)));
         // Top-level grid (app frame)
-        commands
-            .spawn((
-                Node {
-                    display: Display::Grid,
-                    width: Val::Percent(100.0),
-                    height: Val::Percent(100.0),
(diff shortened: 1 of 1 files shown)


examples/testbed/ui.rs
```diff
@@ -313,10 +306,10 @@ mod box_shadow {
                         10.,
                         BorderRadius::MAX,
                     ),
-                ];
-
-                for (size, offset, spread, blur, border_radius) in example_nodes {
-                    commands.spawn((
+                ]
+                .into_iter()
```
alice-i-cecile, line 310: Pretty neutral on this change.

alice-i-cecile (approved):
I think I might revert the SpawnIter changes back to .with_children, but I like the straightforward children! macro usage. Overall, I think this is a mild improvment.

krunchington:
SpawnIter changes reverted, leaving only the one (nested) children change which felt decent.

mnmaita (approved):

PR #18334, opened 2025-03-16
`bevy_ecs` Debloating for `Res`, `ResMut`, and `BundleInserter`

bushrat011899:
# Objective

Compile times in Bevy have consistently degraded over the last several versions. Part of this is due to the extensive use of generic code, which typically leads to duplicated code during monomorphisation. Using [`cargo-llvm-lines`](https://github.com/dtolnay/cargo-llvm-lines) on the crate `bevy_pbr`, I identified several low-hanging-fruit items for refactoring:

* `bevy_ecs::bundle::BundleInserter::insert`: 41214 / 2.2%
* `<bevy_ecs::change_detection::ResMut<T> as bevy_ecs::system::system_param::SystemParam>::init_state` : 19470 / 1.0%
* `<bevy_ecs::change_detection::Res<T> as bevy_ecs::system::system_param::SystemParam>::init_state`: 9999 / 0.5%
* `<bevy_ecs::change_detection::Res<T> as bevy_ecs::system::system_param::SystemParam>::get_param`: 5049 / 0.3%

## Solution

Refactored the above functions to isolate type parameters from the common logic as much as possible. This negatively affects the legibility of the code. But for such fundamental code it may be a necessary evil for improved compile times.

~~Before this change, a clean build of `bevy` took 242 seconds on my laptop. After it took 231 seconds, a reduction of 3.5%.~~

Upon further benchmarking I was unable to reproduce the original results I opened this PR with consistently. I had hoped 11 seconds would be enough of a delta to indicate a definitive improvement, however that appears to be the upper range on the noise I can observe on my machine.

## Testing

- CI

---

## Notes

- Diff looks worse than it is because the indentation level increased.
- Refer to individual commits for function-by-function changes.

### DIFF
--- a/crates/bevy_ecs/src/bundle.rs
+++ b/crates/bevy_ecs/src/bundle.rs
@@ -1106,249 +1106,320 @@ impl<'w> BundleInserter<'w> {
         caller: MaybeLocation,
         relationship_insert_hook_mode: RelationshipInsertHookMode,
     ) -> (EntityLocation, T::Effect) {
-        let bundle_info = self.bundle_info.as_ref();
-        let archetype_after_insert = self.archetype_after_insert.as_ref();
-        let archetype = self.archetype.as_ref();
+        unsafe fn before_write<'w, 'a>(
+            this: &'a mut BundleInserter<'w>,
+            entity: Entity,
+            location: EntityLocation,
+            insert_mode: InsertMode,
+            caller: MaybeLocation,
+        ) -> (
+            &'a mut Archetype,
+            EntityLocation,
+            &'a mut Table,
+            &'a mut SparseSets,
+            &'a ArchetypeAfterBundleInsert,
+            Entity,
+            TableRow,
+            Tick,
+            InsertMode,
+            MaybeLocation,
+        ) {
+            let archetype_after_insert = this.archetype_after_insert.as_ref();
+            let archetype = this.archetype.as_ref();
 
-        // SAFETY: All components in the bundle are guaranteed to exist in the World
-        // as they must be initialized before creating the BundleInfo.
-        unsafe {
-            // SAFETY: Mutable references do not alias and will be dropped after this block
-            let mut deferred_world = self.world.into_deferred();
+            // SAFETY: All components in the bundle are guaranteed to exist in the World
+            // as they must be initialized before creating the BundleInfo.
+            unsafe {
+                // SAFETY: Mutable references do not alias and will be dropped after this block
+                let mut deferred_world = this.world.into_deferred();
 
-            if insert_mode == InsertMode::Replace {
-                if archetype.has_replace_observer() {
-                    deferred_world.trigger_observers(
-                        ON_REPLACE,
+                if insert_mode == InsertMode::Replace {
+                    if archetype.has_replace_observer() {
+                        deferred_world.trigger_observers(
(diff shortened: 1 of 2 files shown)


crates/bevy_ecs/src/bundle.rs
```diff
@@ -1106,250 +1106,321 @@ impl<'w> BundleInserter<'w> {
...
-        let bundle_info = self.bundle_info.as_ref();
-        let archetype_after_insert = self.archetype_after_insert.as_ref();
-        let archetype = self.archetype.as_ref();
-
-        // SAFETY: All components in the bundle are guaranteed to exist in the World
-        // as they must be initialized before creating the BundleInfo.
-        unsafe {
-            // SAFETY: Mutable references do not alias and will be dropped after this block
-            let mut deferred_world = self.world.into_deferred();
+        unsafe fn before_write<'w, 'a>(
```
chescock, line 1109: It might help to leave a comment on each method explaining why it's broken up like this, so that some future dev doesn't try to put it back to improve legibility.

crates/bevy_ecs/src/system/system_param.rs
```diff
@@ -862,25 +877,51 @@ unsafe impl<'a, T: Resource> SystemParam for Res<'a, T> {
...
-                    panic!(
-                        "Resource requested by {} does not exist: {}",
-                        system_meta.name,
-                        core::any::type_name::<T>()
-                    )
-                });
-        Res {
-            value: ptr.deref(),
-            ticks: Ticks {
+        unsafe fn get_param<'w>(
```
chescock, line 880: Not really related to this PR, but I was surprised to see that you updated `Res::get_param` but not `ResMut::get_param`.  It looks like that's because the common code there was already extracted to a `get_resource_mut_by_id` method.  I agree this is the right thing for now, but I wonder if we'll want to add a `struct RefUntyped` someday so that we can have a similar method for `Ref`.

chescock (approved):

IceSentry:
Not super scientific, but based on cargo timings, bevy_render, bevy_core_pipeline and bevy_pbr are all about 2s faster with this PR. It's _possible_ this is noise, but it did seem to make a difference.

IceSentry:
I may have been too excited, I looked at numbers from yesterday and now comparing to main the changes are a lot less visible in compile times. It's annoying how noisy benchmarking compile time is ☹️.

IceSentry:
So, just for bevy_render it's already almost 30k lines less according to cargo-llvm-lines. It went from 800k to 770k. So it's probably still worth it. For bevy_pbr it's a 60k line difference.

IceSentry (approved):
I agree with the suggestion to add a comment explaining why things are broken up, but other than that, LGTM

mockersf:
I'm not seeing an improvement on compile time with this PR, it was actually a bot longer... but I'm on a laptop, so results are not that useful.

This probably need benchmarking too

bushrat011899:
Unfortunately I believe I opened this PR too soon, as I now concur with @mockersf and @IceSentry that there is no consistent compilation time win here. Since that is the primary motivator for such a change, I am closing this PR. Reducing LLVM output _may_ be worthwhile on its own, but without a clear _reason_ to do so, I'm not comfortable advocating for uglier code without tangible benefits.

PR #18345, opened 2025-03-16
Canonicalize the root path to match the canonicalized asset path.

### LINKED ISSUE #18342: Hot reloading on Windows is broken.
## Bevy version

- 137451b

- Windows 10

## What you did

Ran `cargo run --example hot_asset_reloading --features file_watcher`. Modify the `models/torus/torus.gltf` to trigger hot reloading.

## What went wrong

Panics:

```
thread 'notify-rs debouncer loop' panicked at crates\bevy_asset\src\io\file\file_watcher.rs:56:9:
FileWatcher::get_asset_path() failed to strip prefix from absolute path: absolute_path=\\?\E:\Content\bevy\assets\models\torus\torus.glb, root=E:\Content\bevy\assets
```

## Additional information

This appears to be a regression after #18023. Digging in a little, the path before the canonicalize is correct, but after the canonicalize, it gets mangled to include a prefix of "\\?\".

andriyDev:
# Objective

- Fixes #18342.

## Solution

- Canonicalize the root path so that when we try to strip the prefix, it matches the canonicalized asset path.
- This is basically just a followup to #18023.

## Testing

- Ran the hot_asset_reloading example and it no longer panics.

### DIFF
--- a/crates/bevy_asset/src/io/file/file_watcher.rs
+++ b/crates/bevy_asset/src/io/file/file_watcher.rs
@@ -35,7 +35,7 @@ impl FileWatcher {
         sender: Sender<AssetSourceEvent>,
         debounce_wait_time: Duration,
     ) -> Result<Self, notify::Error> {
-        let root = normalize_path(&path);
+        let root = normalize_path(&path).canonicalize().unwrap();
         let watcher = new_asset_event_debouncer(
             path.clone(),
             debounce_wait_time,


hukasu (approved):

alice-i-cecile (approved):
Brutal. We really need better automated tests for bevy_assets.

PR #18346, opened 2025-03-16
bevy_winit: Create the window initially invisible as required by AccessKit

mwcampbell:
The initial `with_visible` call was intended to do this, but that was undone by a later `with_visible` call.

### DIFF
--- a/crates/bevy_winit/src/winit_windows.rs
+++ b/crates/bevy_winit/src/winit_windows.rs
@@ -110,6 +110,9 @@ impl WinitWindows {
             }
         };
 
+        // It's crucial to avoid setting the window's final visibility here;
+        // as explained above, the window must be invisible until the AccessKit
+        // adapter is created.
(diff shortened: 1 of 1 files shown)


alice-i-cecile (changes requested):
I trust your judgement on the importance here, but can you add a comment to the set_visible line explaining why this is needed? It's not immediately apparent, and I'm worried about accidentally reverting this if there's no comment there :)

mwcampbell:
OK, now there are a total of three comments (the first one was already there).

alice-i-cecile (approved):
Perfect, thank you so much.

mwcampbell:
I don't seem to have permission to merge this. Or are we waiting for another review?

alice-i-cecile:
Waiting for a second approval, then I'll merge it. This will make it into 0.16 though: it's in the milestone.

mockersf (approved):

PR #18349, opened 2025-03-17
Make TAA non-experimental, fixes

JMS55:
The first 4 commits are designed to be reviewed independently.

- Mark TAA non-experimental now that motion vectors are written for skinned and morphed meshes, along with skyboxes, and add it to DefaultPlugins
- Adjust halton sequence to match what DLSS is going to use, doesn't really affect anything, but may as well
- Make MipBias a required component on TAA instead of inserting it in the render world
- Remove MipBias, TemporalJitter, RenderLayers, etc from the render world if they're removed from the main world (fixes a retained render world bug)
- Remove TAA components from the render world properly if TemporalAntiAliasing is removed from the main world (fixes a retained render world bug)
  - extract_taa_settings() now has to query over `Option<&mut TemporalAntiAliasing>`, which will match every single camera, in order to cover cameras that had TemporalAntiAliasing removed this frame. This kind of sucks, but I can't think of anything better.
  - We probably have the same bug with every other rendering feature component we have.

## Migration Guide
- TAA is no longer experimental
  - TemporalAntiAliasing no longer needs to be added to your app to use TAA. It is now part of DefaultPlugins/CorePipelinePlugin.
  - TemporalAntiAliasing now uses MipBias as a required component in the main world, instead of overriding it manually in the render world.

### DIFF
--- a/crates/bevy_core_pipeline/src/taa/mod.rs
+++ b/crates/bevy_core_pipeline/src/taa/mod.rs
@@ -115,7 +115,6 @@ impl Plugin for TemporalAntiAliasPlugin {
 ///
 /// # Usage Notes
 ///
-/// The [`TemporalAntiAliasPlugin`] must be added to your app.
 /// Any camera with this component must also disable [`Msaa`] by setting it to [`Msaa::Off`].
 ///
 /// [Currently](https://github.com/bevyengine/bevy/issues/8423), TAA cannot be used with [`bevy_render::camera::OrthographicProjection`].
@@ -153,7 +152,7 @@ impl Default for TemporalAntiAliasing {
 
 /// Render [`bevy_render::render_graph::Node`] used by temporal anti-aliasing.
 #[derive(Default)]
-pub struct TemporalAntiAliasNode;
+struct TemporalAntiAliasNode;
 
 impl ViewNode for TemporalAntiAliasNode {
     type ViewQuery = (
@@ -408,7 +407,7 @@ fn prepare_taa_jitter_and_mip_bias(
 }
 
 #[derive(Component)]
-pub struct TemporalAntiAliasHistoryTextures {
+struct TemporalAntiAliasHistoryTextures {
     write: CachedTexture,
     read: CachedTexture,
 }
@@ -465,7 +464,7 @@ fn prepare_taa_history_textures(
 }
 
 #[derive(Component)]
-pub struct TemporalAntiAliasPipelineId(CachedRenderPipelineId);
+struct TemporalAntiAliasPipelineId(CachedRenderPipelineId);
 
 fn prepare_taa_pipelines(
     mut commands: Commands,
--- a/crates/bevy_core_pipeline/src/taa/mod.rs
+++ b/crates/bevy_core_pipeline/src/taa/mod.rs
@@ -342,16 +342,11 @@ impl SpecializedRenderPipeline for TaaPipeline {
 }
 
 fn extract_taa_settings(mut commands: Commands, mut main_world: ResMut<MainWorld>) {
-    let mut cameras_3d = main_world.query_filtered::<(
+    let mut cameras_3d = main_world.query::<(
         RenderEntity,
         &Camera,
         &Projection,
-        &mut TemporalAntiAliasing,
-    ), (
-        With<Camera3d>,
-        With<TemporalJitter>,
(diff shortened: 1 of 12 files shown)


crates/bevy_render/src/camera/camera.rs
```diff
@@ -1187,14 +1190,26 @@ pub fn extract_cameras(
...
             if let Some(render_layers) = render_layers {
                 commands.insert(render_layers.clone());
+            } else {
+                commands.remove::<RenderLayers>();
             }
 
             if let Some(perspective) = projection {
                 commands.insert(perspective.clone());
+            } else {
+                commands.remove::<Projection>();
```
tychedelia, line 1212: Oof, thanks for catching these. Still somewhat distressed about how ad hoc our patterns are here but at least we won't encounter these particular  bugs waiting to happen in the future.

crates/bevy_core_pipeline/src/taa/mod.rs
```diff
@@ -364,51 +356,53 @@ fn extract_taa_settings(mut commands: Commands, mut main_world: ResMut<MainWorld
         let mut entity_commands = commands
             .get_entity(entity)
             .expect("Camera entity wasn't synced.");
-        if camera.is_active && has_perspective_projection {
-            entity_commands.insert(taa_settings.clone());
-            taa_settings.reset = false;
+        if taa_settings.is_some() && camera.is_active && has_perspective_projection {
```
tychedelia, line 359: I bet this same bug is lurking a few other places.

tychedelia (approved):
This is a nice small cleanup + glad to see this move out of experimental.

alice-i-cecile:
Marked as contentious because this is taking a feature out of experimental, but I have no objections to doing so.

crates/bevy_core_pipeline/src/taa/mod.rs
JMS55: I looked at how ExtractComponentPlugin works and I think all of those will have the same issue too, since it's doing `type QueryData = &'static Self;`. Right? Double check me on this please.

tychedelia: Yes, that's immediately what came to mind since it only runs if the component is actually present and just removes components if it doesn't get extracted (e.g. because a camera is disabled). I don't love the idea of essentially running all extracts unconditionally all the time (especially if a component like taa has never even been added) but that's definitely the fix given the current patterns.

tychedelia: Yes, I agree, some kind of observer/event based system would be great for this kind of data that can change but doesn't need to be extracted every frame.

JMS55:
Oh, there's also some thorny problems around removing TAA jitter/mipbias/prepasses (in the main world) when toggling TAA off in examples. I made sure the anti_aliasing example does it correctly, but I'm not too confident in the rest.

janhohenheim:
Maybe consider https://github.com/bevyengine/bevy/issues/18903 before merging.
Not sure if this is blocking, but at least some documentation pointing users in the right direction would be nice :)

crates/bevy_anti_aliasing/src/taa/mod.rs
```diff
@@ -151,7 +148,7 @@ impl Default for TemporalAntiAliasing {
 
 /// Render [`bevy_render::render_graph::Node`] used by temporal anti-aliasing.
 #[derive(Default)]
-pub struct TemporalAntiAliasNode;
+struct TemporalAntiAliasNode;
```
IceSentry: Why did you remove the `pub` here?

alice-i-cecile:
@JMS55, lemme know if that pub removal was intentional or accidental? Once that's done, I'm happy to merge this for you ASAP.

crates/bevy_anti_aliasing/src/taa/mod.rs
JMS55: Might've been an accident, it looks like it's currently public.

JMS55:
```suggestion
pub struct TemporalAntiAliasNode;
```

JMS55:
@alice-i-cecile good to merge now.

alice-i-cecile:
Simple CI failure now :)

PR #18350, opened 2025-03-17
Update docs to explain default Hasher issue

krunchington:
# Objective

I experienced an issue where `HashMap::new` was not returning a value typed appropriately for a `HashMap<K,V>` declaration that omitted the Hasher- e.g. the Default Hasher for the type is different than what the `new` method produces.

After discussion on discord, this appears to be an issue in `hashbrown`, and working around it would be very nontrivial, requiring a newtype on top of the `hashbrown` implementation.  Rather than doing that, it was suggested that we add docs to make the issue more visible and provide a clear workaround.

## Solution

Updated the docs for `bevy_platform_support::collections`.  I couldn't update Struct docs because they're re-exports, so I had to settle for the module.

Note that the `[HashMap::new]` link wasn't generating properly- I'm not sure why.  I see the method in the docs.rs site, https://docs.rs/hashbrown/0.15.1/hashbrown/struct.HashMap.html#method.new, but not on the generated internal documentation.  I wonder if `hashbrown` isn't actually implementing the new or something? 

## Testing

n/a although I did generate and open the docs on my Ubuntu machine.

---

## Showcase

before:
image

after:
image

### DIFF
--- a/crates/bevy_platform_support/src/collections.rs
+++ b/crates/bevy_platform_support/src/collections.rs
@@ -7,7 +7,9 @@
 //! To bypass this issue, use [`HashMap::default`] instead.
 //!
 //! ```
-//! use bevy::{prelude::*, platform_support::collections::HashMap};
+//! # use bevy_ecs::component::Component;
+//! # use bevy_ecs::system::Commands;
+//! # use bevy_platform_support::collections::HashMap;
 //!
 //! #[derive(Component)]
 //! struct MyComponent {
--- a/crates/bevy_platform_support/src/collections.rs
+++ b/crates/bevy_platform_support/src/collections.rs
@@ -1,6 +1,25 @@
-//! Provides [`HashMap`] and [`HashSet`] from [`hashbrown`] with some customized defaults.\
+//! Provides [`HashMap`] and [`HashSet`] from [`hashbrown`] with some customized defaults.
 //!
 //! Also provides the [`HashTable`] type, which is specific to [`hashbrown`].
+//!
+//! Note that due to the implementation details of [`hashbrown`], [`HashMap::new`] produces a
+//! `HashMap<K,V,bevy::bevy_platform_support::hash::RandomState>` (uses a `RandomState` hasher)
+//! rather than the expected `FixedHasher` if you declare a `HashMap<K,V>` (without a hasher
+//! specified).  To bypass this issue, use [`HashMap::default`] instead.
+//! ```
+//! use bevy::{prelude::*, bevy_platform_support::collections::HashMap};
+//!
+//! #[derive(Component)]
+//! struct MyComponent {
+//!     map: HashMap<String, String>
+//! }
+//!
+//! fn my_system(mut commands: Commands) {
+//!     commands.spawn(MyComponent {
+//!         map: HashMap::default(),
+//!     });
+//! }
+//! ```
 
 pub use hash_map::HashMap;
 pub use hash_set::HashSet;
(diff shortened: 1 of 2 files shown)


alice-i-cecile (approved):

crates/bevy_platform_support/src/collections.rs
```diff
@@ -1,6 +1,25 @@
...
+//! Provides [`HashMap`] and [`HashSet`] from [`hashbrown`] with some customized defaults.
 //!
 //! Also provides the [`HashTable`] type, which is specific to [`hashbrown`].
+//!
+//! Note that due to the implementation details of [`hashbrown`], [`HashMap::new`] produces a
+//! `HashMap<K,V,bevy::bevy_platform_support::hash::RandomState>` (uses a `RandomState` hasher)
+//! rather than the expected `FixedHasher` if you declare a `HashMap<K,V>` (without a hasher
+//! specified).  To bypass this issue, use [`HashMap::default`] instead.
+//! ```
+//! use bevy::{prelude::*, bevy_platform_support::collections::HashMap};
```
bushrat011899, line 10:
```suggestion
//! use bevy::{prelude::*, platform_support::collections::HashMap};
```

crates/bevy_platform_support/src/collections.rs
```diff
@@ -1,6 +1,25 @@
-//! Provides [`HashMap`] and [`HashSet`] from [`hashbrown`] with some customized defaults.\
+//! Provides [`HashMap`] and [`HashSet`] from [`hashbrown`] with some customized defaults.
 //!
 //! Also provides the [`HashTable`] type, which is specific to [`hashbrown`].
+//!
+//! Note that due to the implementation details of [`hashbrown`], [`HashMap::new`] produces a
+//! `HashMap<K,V,bevy::bevy_platform_support::hash::RandomState>` (uses a `RandomState` hasher)
+//! rather than the expected `FixedHasher` if you declare a `HashMap<K,V>` (without a hasher
+//! specified).  To bypass this issue, use [`HashMap::default`] instead.
```
bushrat011899, line 8:
```suggestion
//! Note that due to the implementation details of [`hashbrown`], [`HashMap::new`] is only implemented for `HashMap<K, V, RandomState>`.
//! Whereas, Bevy exports `HashMap<K, V, FixedHasher>` as its default [`HashMap`] type, meaning [`HashMap::new`] will typically fail.
//! To bypass this issue, use [`HashMap::default`] instead.
//!
```

crates/bevy_platform_support/src/collections.rs
```diff
@@ -1,6 +1,25 @@
-//! Provides [`HashMap`] and [`HashSet`] from [`hashbrown`] with some customized defaults.\
+//! Provides [`HashMap`] and [`HashSet`] from [`hashbrown`] with some customized defaults.
```
bushrat011899, line 1: Good catch!

bushrat011899 (changes requested):
Just some slight changes to wording and formatting. Unless Bevy wants to new-type `HashMap`/`HashSet` I agree this is the best we can do to document the issue.

crates/bevy_platform_support/src/collections.rs
krunchington: thanks!  Just happened to be there so it was easy for me to spot :sweat_smile:

krunchington:
I committed the suggested changes but I suspect the docs CI will fail on the `HashMap::new` link again.  I can unlink that if it does.

krunchington:
NVM it was the example with imports.  Should be fixed now though I had to put a dev dependency on bevy_ecs

bushrat011899 (approved):
Nice work!

alice-i-cecile:
CI is upset at you:

> error[wildcard]: found 1 wildcard dependency for crate 'bevy_platform_support'

This needs to be fixed before we can merge.

PR #18351, opened 2025-03-17
Unify and simplify command and system error handling

### LINKED ISSUE #17272: Unify system and command error handling

alice-i-cecile:
# Objective

- ECS error handling is a lovely flagship feature for Bevy 0.16, all in the name of reducing panics and encouraging better error handling (#14275).
- Currently though, command and system error handling are completely disjoint and use different mechanisms.
- Additionally, there's a number of distinct ways to set the default/fallback/global error handler that have limited value. As far as I can tell, this will be cfg flagged to toggle between dev and production builds in 99.9% of cases, with no real value in more granular settings or helpers.
- Fixes #17272

## Solution

- Standardize error handling on the OnceLock global error mechanisms ironed out in https://github.com/bevyengine/bevy/pull/17215
   - As discussed there, there are serious performance concerns there, especially for commands
   - I also think this is a better fit for the use cases, as it's truly global
- Move from `SystemErrorContext` to a more general purpose `ErrorContext`, which can handle observers and commands more clearly
- Cut the superfluous setter methods on `App` and `SubApp`
- Rename the limited (and unhelpful) `fallible_systems` example to `error_handling`, and add an example of command error handling

## Testing

Ran the `error_handling` example.

## Notes for reviewers

- Do you see a clear way to allow commands to retain &mut World access in the per-command custom error handlers? IMO that's a key feature here (allowing the ad-hoc creation of custom commands), but I'm not sure how to get there without exploding complexity.
- I've removed the feature gate on the default_error_handler: contrary to @cart's opinion in #17215 I think that virtually all apps will want to use this. Can you think of a category of app that a) is extremely performance sensitive b) is fine with shipping to production with the panic error handler? If so, I can try to gather performance numbers and/or reintroduce the feature flag. UPDATE: see benches at the end of this message.
- ~~`OnceLock` is in `std`: @bushrat011899 what should we do here?~~
- Do you have ideas for more automated tests for this collection of features?

## Benchmarks

I checked the impact of the feature flag introduced: benchmarks might show regressions. This bears more investigation. I'm still skeptical that there are users who are well-served by a fast always panicking approach, but I'm going to re-add the feature flag here to avoid stalling this out.

image


bushrat011899:
Re. `OnceLock`: I've already got a suitable alternative implemented in `bevy_platform_support::sync::OnceLock`! Should be a drop-in replacement that'll defer to `std` when possible and use a spinlock when it can't.

crates/bevy_ecs/src/error/handler.rs
```diff
@@ -1,75 +1,150 @@
-use crate::{component::Tick, error::BevyError, resource::Resource};
-use alloc::borrow::Cow;
+use core::fmt::Display;
+use std::sync::OnceLock;
```
bushrat011899, line 2:
```suggestion
use bevy_platform_support::sync::OnceLock;
```
That _should_ just work.

crates/bevy_ecs/src/error/handler.rs
```diff
@@ -1,75 +1,150 @@
...
+/// ```
+/// # use bevy_ecs::system::error_handler::{GLOBAL_ERROR_HANDLER, warn};
+/// GLOBAL_ERROR_HANDLER.set(warn).expect("The error handler can only be set once, globally.");
+/// // initialize Bevy App here
+/// ```
+///
+/// # Warning
+///
+/// As this can *never* be overwritten, library code should never set this value.
+pub static GLOBAL_ERROR_HANDLER: OnceLock<fn(BevyError, ErrorContext)> = OnceLock::new();
```
bushrat011899, line 85: I think the fact that this is a `OnceLock` should be a hidden implementation detail users don't interact with, allowing us to experiment with the implementation without breaking user code needlessly. Perhaps exposing a simple `set_error_handler(...)` function to go with the `default_error_handler()` function that is already added?

crates/bevy_ecs/src/error/handler.rs
```diff
@@ -1,75 +1,150 @@
...
-    /// The name of the system that failed.
-    pub name: Cow<'static, str>,
+use crate::{component::Tick, error::BevyError};
+use alloc::borrow::Cow;
 
-    /// The last tick that the system was run.
-    pub last_run: Tick,
+/// Context for a [`BevyError`] to aid in debugging.
+#[derive(Debug, PartialEq, Eq, Clone)]
+pub enum ErrorContext {
```
bushrat011899, line 9: Future PR: Experimenting with this context type will be very helpful in the future. For example, systems and observers could include the `SystemId`, allowing a more rich error handler to have differing behaviour based on the system that failed, but entirely up to the user.

examples/ecs/error_handling.rs
```diff
@@ -10,6 +13,16 @@ use rand::SeedableRng;
 use rand_chacha::ChaCha8Rng;
 
 fn main() {
+    // By default, fallible systems that return an error will panic.
+    //
+    // We can change this by setting a custom error handler, which applies globally.
+    //Here we set the global error handler using one of the built-in
```
bushrat011899, line 19:
```suggestion
    // Here we set the global error handler using one of the built-in
```

bushrat011899 (approved):
I think this is a good direction to move error handling towards. There's definitely room for iteration, but this is the most flexible and most approachable form of error handling thus far. For example, it may be nice to have fallible plugins (e.g., `Plugin::build` could return a `Result`), etc. in the future.

Some minor typos and a suggestion to use `bevy_platform_support::sync::OnceLock` for `no_std` compatibility reasons. I do also raise the idea of making the `OnceLock` itself private and just having a function to initialize the handler. While I think that's a nicer API, I don't see it as a blocker.

crates/bevy_ecs/src/error/handler.rs
alice-i-cecile: The other direction we can go is to remove default_error_handler completely. I don't think it adds a ton without the feature flag.

alice-i-cecile: I decided to cut default_error_handler: it doesn't make this more misuse resistant, and end users generally won't be creating custom error handling code so the convenience argument doesn't hold water for me either.

crates/bevy_ecs/src/error/handler.rs
alice-i-cecile: FYI I think that sticking complex app-specific error handling in the global error handler is likely to be a bit of a mess. Instead, users should be trying to handle errors as close to where they're generated as possible, and we should provide them the tools required to do that.

NthTensor: I agree with Alice here complex conditional logic in the default handler will probably be an anti-pattern. We can expect users to want to have complex error handling for *their code* but much less precision is possible for engine code or ecosystem code. For code you write yourself, there's probably better ways to handle errors (like piping). What the default handler gives is consistency: you don't have to worry about one ecosystem developer choosing to panic and another choosing to silently fail without logging.

That said, we should still probably let users do it. Providing the system ID can't hurt.

crates/bevy_ecs/src/error/command_handling.rs
```diff
@@ -0,0 +1,110 @@
...
+    error::panic,
+    system::{entity_command::EntityCommandError, Command, EntityCommand},
+    world::{error::EntityMutableFetchError, World},
+};
+
+use super::{BevyError, ErrorContext, GLOBAL_ERROR_HANDLER};
+
+/// Takes a [`Command`] that returns a Result and uses a given error handler function to convert it into
+/// a [`Command`] that internally handles an error if it occurs and returns `()`.
+pub trait HandleError<Out = ()> {
```
NthTensor, line 14: I'm sort of confused why this is a generic with an empty default. The docs don't make it clear to me either. Can you help me out here?

alice-i-cecile: This was moved unchanged in this PR. It was introduced in 7f622096463e35e3ef38c2c59a68fa4c5e1f0f367632eaf1e9c4779fcc72d74eR69.

crates/bevy_ecs/src/error/handler.rs
alice-i-cecile: 100% agreed on source location tracking, and no objection to exposing system id etc. Seems handy for tooling! Just slap on some "please don't omni-error handling" advice in the docs :)

crates/bevy_ecs/src/error/handler.rs
```diff
@@ -1,75 +1,154 @@
...
+///    // Libraries that you ship should follow this convention to avoid surprising users!
+///    let error_handler = *GLOBAL_ERROR_HANDLER.get_or_init(|| panic);
+///    error_handler(error, ctx);        
+/// }
+/// ```
+///
+/// # Warning
+///
+/// As this can *never* be overwritten, library code should never set this value.
+pub static GLOBAL_ERROR_HANDLER: OnceLock<fn(BevyError, ErrorContext)> = OnceLock::new();
```
NthTensor, line 96: Thoughts on hiding this from users and having a `App::set_error_handler` fn which can be called multiple times, and then which actually sets this in `App::start`? We could even make this private and expose it only through a `get_error_handler()` function.

I think this might be closer to what bevy users expect from our API, and would eliminate the problems caused by calling it conflicting handlers or trying to change it while the app is running. Or IDK maybe we want to be able to swap it out at runtime for the editor, in which case a OnceLock isn't the right primitive either.

crates/bevy_ecs/src/error/command_handling.rs
NthTensor: After reading through the rest of the implementation it makes sense.

crates/bevy_ecs/src/error/handler.rs
NthTensor: I think this was the right call.

NthTensor (changes requested):
First code review from me in a while, sorry if I'm a bit rusty. I like seeing all the red, and I think what you've ended up on is really simple and clear. I don't really have any faults with the implementation, but I do still have some questions.

> Do you see a clear way to allow commands to retain &mut World access...

The issue is not that we don't have world-access when handling errors in commands or systems, right? The issue is observers. Is there a reason we can't go with something like the following:
```rust
fn error_handler(error: BevyError, ctx: ErrorContext, world: DeferredWorld)
```
Commands and systems appear to have `&mut World` access when they call `GLOBAL_ERROR_HANDLER.get_or_init(|| panic)`, which means it should be safe to transmute into `UnsafeWorldCell` and `DeferredWorld`, no? And observers already have a deferred world they can use. I'm guess I am probably missing something, but it seems like this would be a way to preserve access to commands.

> I've removed the feature gate on the default_error_handler...

I think this is probably the right call, but we should benchmark as it could have perf implications. Unfortunately I do think we should wait to merge this until we get either positive benchmarks or hear Cart's stance, which is the only reason I'm marking this as changes requested.

> OnceLock is in std...

The `once_cell` crate has `no_std` support via `critical_section`.

> Do you have ideas for more automated tests for this collection of features?

Honestly, I think we've landed on a pretty simple implementation. There doesn't seem to be much room for logic errors. Performance is the major concern.

crates/bevy_ecs/src/error/command_handling.rs
NthTensor: This makes more sense to me now that I've read the rest of the code.

crates/bevy_ecs/src/error/handler.rs
alice-i-cecile: That won't work for standalone bevy_ecs users: they don't have access to App. We also *really* want this to be truly global, not set per SubApp.

That said, I am down to make this a bit more opaque to give us flexibility and a nicer API. Let me see about newtyping it.

alice-i-cecile:
> The issue is not that we don't have world-access when handling errors in commands or systems, right? The issue is observers. Is there a reason we can't go with something like the following:

Originally yes. I'm increasingly convinced that allowing &mut World access in the global handler is a bit of a trap though. Without a really good motivating use case I'm worried that providing that will encourage omni-handlers. However, I'd *really* like it in `Commands::queue_handled`, so you can respond to failed commands right at the call site nice and easily.

That *can* easily wait for followup though.

> Honestly, I think we've landed on a pretty simple implementation. There doesn't seem to be much room for logic errors. Performance is the major concern.

Valid. I can try and spend some time benching this.

crates/bevy_ecs/src/error/handler.rs
NthTensor: In that case I think the current is probably fine TBH. The API works, it was just a thought.

alice-i-cecile: It does work, but I don't like it! 😂

NthTensor: We could always provide a wrapper as part of `App` that hides the less nice ECS implementation from most users.

alice-i-cecile: IMO there should be one way to set this: no need to complicate things.

alice-i-cecile:
@NthTensor I'm seeing in the ballpark of 10% perf regressions on some of the command benchmarks with the configurable error handling enabled. I'm not 100% convinced that's real, but that's enough to sell me on feature-flagging this again for this PR.

crates/bevy_ecs/src/error/handler.rs
```diff
@@ -1,75 +1,171 @@
...
+/// # Warning
+///
+/// As this can *never* be overwritten, library code should never set this value.
+#[cfg(feature = "configurable_error_handler")]
+pub static GLOBAL_ERROR_HANDLER: OnceLock<fn(BevyError, ErrorContext)> = OnceLock::new();
+
+/// The default error handler. This defaults to [`panic()`],
+/// but if set, the [`GLOBAL_ERROR_HANDLER`] will be used instead, enabling error handler customization.
+/// The `configurable_error_handler` feature must be enabled to change this from the panicking default behavior,
+/// as there may be runtime overhead.
```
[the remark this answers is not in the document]
alice-i-cecile: That was Cart's original design 🤷🏽‍♀️ I think that's reasonable, but I don't feel like running more benches here right now.

PR #18353, opened 2025-03-17
Added `io` Module to `bevy_platform_support`

bushrat011899:
# Objective

Continuing `no_std` proliferation in Bevy is currently blocked on `bevy_asset` which makes extensive use of the standard library in two main areas:
- `std::io`
- `std::path`
This is entirely sensible, since `bevy_asset`'s primary job is to load assets from a _path_ using filesystem _IO_.

However, assets can be used outside of strict IO operations. For example, users can manually add assets to an appropriate `Assets<T>` resource. Further, while Bevy _supports_ `std::path::Path`, we have strong rules around what's considered canonical for an `AssetPath`; namely that platform specific behaviour (backslashes and drive letters on Windows for example) is _not_ supported.

Since `no_std` support is likely to be highly disruptive in `bevy_asset`, this PR attempts to tackle a subset of the problem: independence from `std::io::Result`.

## Solution

- Created a new `io` module in `bevy_platform_support` which provides `Error`, `Result`, and `ErrorKind`.
- When the standard library is not available, `bevy_platform_support::io` will fallback to an API compatible alternative.
- Switch `bevy_asset` to use `bevy_platform_support::io::Result`. Since it _is_ `std::io::Result` with the `std` feature enabled this is purely a path change and has no impact on API or functionality.

## Testing

- CI

---

## Notes

- It may be desirable to actually make a more breaking change to `bevy_asset` and instead roll its own `AssetIoError` type. However, this is likely more controversial so I have opted for this as a more gradual step.

### DIFF
--- a/crates/bevy_asset/src/io/file/file_asset.rs
+++ b/crates/bevy_asset/src/io/file/file_asset.rs
@@ -17,14 +17,14 @@ impl AsyncSeekForward for File {
         mut self: Pin<&mut Self>,
         cx: &mut task::Context<'_>,
         offset: u64,
-    ) -> Poll<futures_io::Result<u64>> {
+    ) -> Poll<bevy_platform_support::io::Result<u64>> {
         let offset: Result<i64, _> = offset.try_into();
 
         if let Ok(offset) = offset {
             Pin::new(&mut self).poll_seek(cx, futures_io::SeekFrom::Current(offset))
         } else {
-            Poll::Ready(Err(std::io::Error::new(
-                std::io::ErrorKind::InvalidInput,
+            Poll::Ready(Err(bevy_platform_support::io::Error::new(
+                bevy_platform_support::io::ErrorKind::InvalidInput,
                 "seek position is out of range",
             )))
         }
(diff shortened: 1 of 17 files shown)


bushrat011899:
Putting on the 0.17 milestone just to highlight I don't intend for this to land in 0.16. It _could_, but I'd rather not make that milestone harder to hit than it already is!

alice-i-cecile:
> It may be desirable to actually make a more breaking change to bevy_asset and instead roll its own AssetIoError type. However, this is likely more controversial so I have opted for this as a more gradual step.

I actually prefer this solution. I generally don't think that relying on the standard IO error type is super important, and I prefer to minimize the amount of duplicated code from the standard library.

andriyDev (approved):
Sorry for the delay in review! This all looks great!

andriyDev:
In regards to rolling our own `AssetIoError`, I don't have any particular tie to the std IO error. Of course we'd need users to convert the std IO error to our error type (since the default asset sources/readers will use std IO). So I'd be up for it!

PR #14623, opened 2024-08-05
Meshlet software raster + start of cleanup

JMS55:
# Objective
- Faster meshlet rasterization path for small triangles
- Avoid having to allocate and write out a triangle buffer
- Refactor gpu_scene.rs

## Solution
- Replace the 32bit visbuffer texture with a 64bit visbuffer buffer, where the left 32 bits encode depth, and the right 32 bits encode the existing cluster + triangle IDs. Can't use 64bit textures, wgpu/naga doesn't support atomic ops on textures yet.
- Instead of writing out a buffer of packed cluster + triangle IDs (per triangle) to raster, the culling pass now writes out a buffer of just cluster IDs (per cluster, so less memory allocated, cheaper to write out). 
  - Clusters for software raster are allocated from the left side
  - Clusters for hardware raster are allocated in the same buffer, from the right side
  - The buffer size is fixed at MeshletPlugin build time, and should be set to a reasonable value for your scene (no warning on overflow, and no good way to determine what value you need outside of renderdoc - I plan to fix this in a future PR adding a meshlet stats overlay)
  - Currently I don't have a heuristic for software vs hardware raster selection for each cluster. The existing code is just a placeholder. I need to profile on a release scene and come up with a heuristic, probably in a future PR.
  - The culling shader is getting pretty hard to follow at this point, but I don't want to spend time improving it as the entire shader/pass is getting rewritten/replaced in the near future.
- Software raster is a compute workgroup per-cluster. Each workgroup loads and transforms the <=64 vertices of the cluster, and then rasterizes the <=64 triangles of the cluster.
  - Two variants are implemented: Scanline for clusters with any larger triangles (still smaller than hardware is good at), and brute-force for very very tiny triangles
  - Once the shader determines that a pixel should be filled in, it does an atomicMax() on the visbuffer to store the results, copying how Nanite works
  - On devices with a low max workgroups per dispatch limit, an extra compute pass is inserted before software raster to convert from a 1d to 2d dispatch (I don't think 3d would ever be necessary).
  - I haven't implemented the top-left rule or subpixel precision yet, I'm leaving that for a future PR since I get usable results without it for now
  - Resources used: https://kristoffer-dyrkorn.github.io/triangle-rasterizer and chapters 6-8 of https://fgiesen.wordpress.com/2013/02/17/optimizing-sw-occlusion-culling-index
- Hardware raster now spawns 64*3 vertex invocations per meshlet, instead of the actual meshlet vertex count. Extra invocations just early-exit. 
[...]

PR #18356, opened 2025-03-17
Provide `abort` and `exit` on all platforms

bushrat011899:
Panic within a panic is defined by the Rust language to result in an abort. We can use that to provide definitions for `abort` and `exit` on `no_std`.

# Objective

There is a somewhat archaic trick within Rust known as a double-panic abort. It is defined within the language that _if_ a `panic!` occurs during another `panic!`, then the process _must_ abort. This is used by many `no_std` libraries, such as [`async-lock`](https://github.com/smol-rs/async-lock/blob/cd0937d70a46f50cc64428660e249c8464e70088/src/lib.rs#L148C1-L168C2) to provide a `no_std` compatible mechanism for aborting.

While Bevy itself only uses `exit` in a handful of locations (and in ways which could be potentially removed, see notes below), this is a simple enough mechanism to provide.

## Solution

- Added a `process` module to `bevy_platform_support`
- Added an `abort` implementation based on the double-panic abort.
- Added an `exit` implementation which simply aborts without reporting the error code.
- Replaced all usage of `std::process::exit` with the new `bevy_platform_support::process::exit`, which uses `std::process::exit` when possible and a double-panic abort otherwise.

## Testing

- CI

---

## Notes

- The only usage of `exit` in Bevy is within the `MeshletPlugin` as a last-resort handler of invalid state. It may be more appropriate to replace its usage with a more result-ful alternative. For example, perhaps `Plugin::build` could be fallible to allow detecting these invalid states and allowing the user to decide how to proceed (e.g., recreate their `App` without the offending state, etc.)


chescock (approved):
The implementation of `abort` looks good!  

Do we really want to be encouraging the use of `abort` or `exit` over `panic`, though?  Even if we can't go all the way to `Result`, users may want to embed Bevy inside larger applications that can recover from panics.  

It looks like the `exit` calls in `meshlet` were changed *from* `panic`s in #14623, so I assume it's intentional, but I couldn't find any discussion of why.

alice-i-cecile (approved):

crates/bevy_pbr/src/meshlet/mod.rs
```diff
@@ -149,7 +149,7 @@ impl Plugin for MeshletPlugin {
 
         if self.cluster_buffer_slots > 2_u32.pow(25) {
             error!("MeshletPlugin::cluster_buffer_slots must not be greater than 2^25.");
-            std::process::exit(1);
+            bevy_platform_support::process::exit(1);
```
alice-i-cecile, line 152: The notion of a no_std platform running meshlets is very funny to me. Maybe on modern consoles though!

alice-i-cecile:
I don't think this is particularly controversial. @bushrat011899 can you explain that label to me?

bushrat011899:
I chose _Controversial_ because I thought this code would look pretty strange, and as @chescock points out, I don't think we want to encourage the use of `abort` or `exit` within Bevy.

But as they _also_ point out it appears to be very deliberately chosen over `panic`, so it's likely not trivial to swap off of. I suspect the reason for its use over panic is to ensure `catch_unwind` doesn't allow the app to run after this invalid state is detected.

crates/bevy_platform_support/src/process.rs
```diff
@@ -0,0 +1,35 @@
+//! Provides [`exit`] and [`abort`] and all platforms.
```
alice-i-cecile, line 1: We should add comments here about how these should rarely be used.

crates/bevy_platform_support/src/process.rs
```diff
@@ -0,0 +1,35 @@
...
+            }
+
+            /// Terminates the current process with the specified exit code.
+            #[cold]
+            pub fn exit(_code: i32) -> ! {
+                abort();
+            }
+        }
+
+        use fallback as process;
```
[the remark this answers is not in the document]
bushrat011899: Yeah exactly, just for consistency.

crates/bevy_pbr/src/meshlet/mod.rs
bushrat011899: Yeah it's not really in my sights any time soon! Mostly just my zealous removal of `std` from the Bevy codebase.

crates/bevy_platform_support/src/process.rs
bushrat011899: Agreed. I've added a little note here calling out `panic` and `Result` as the preferred mechanisms for communicating failure.

Issue #18357, opened 2025-03-17
`insert_with_relationship_insert_hook_mode` still runs the relationship `on_replace` implementation.


andriyDev:
## Bevy version

- 79655269f583ad2141c6b14e6a0697c67f49a2dc

## What you did

```rust
let app = App::new();
let world = app.world_mut();
let parent = world.spawn_empty().id();
let child = world
    .spawn(ChildOf(parent))
    .insert_with_relationship_insert_hook_mode(ChildOf(parent), RelationshipInsertHookMode::Skip);
assert!(world.entity(parent1).get::<Children>().is_some());
```

We made child a ChildOf parent, then we reinserted the ChildOf without its insert hooks. We'd expect the parent to still have the Children component from the initial hook.

## What went wrong

The parent is missing the Children component since the on_replace hook ran!

#17858 made it so that relationship hooks do not run when spawning scenes. However, it did not prevent on_replace hooks. This can result in a relationship being removed, but the corresponding relationship not being inserted back.

This only applies when the entity already has the component being inserted. In many cases this is not a problem - for entity cloning, the entity doesn't have any relationships, so nothing breaks. For spawning scenes from scratch, it is also not a problem. However for scene hot-reloading, this is a problem.

alice-i-cecile:
FYI @cart

andriyDev:
#18358 is how I've been toying around with this. Since the hot reloading now reuses the entities, the on_replace hook hurts us.

cart:
I'll look into this!

PR #18358, opened 2025-03-17
Fix Scene hot reloading.

### LINKED ISSUE #14698: Hot asset reloading example does not update when saving torus.gltf
## Bevy version

0.14.1

Tested on:
- Windows 10
- Linux (Ubuntu)

## What you did

Ran the [hot_asset_reloading](https://github.com/bevyengine/bevy/blob/release-0.14.1/examples/asset/hot_asset_reloading.rs) example on my Window 10 desktop, and Ubuntu laptop.

CLI args:
```bash
cargo run --example hot_asset_reloading --features file_watcher
```

## What went wrong

Updating the [torus](https://github.com/bevyengine/bevy/tree/release-0.14.1/assets/models/torus) gltf file does not update the model while running. Specifically, I updated the rotation quaternion in the file. Only restarting the app shows me the change.

## Additional information

Editing the base color of the [monkey model](https://github.com/bevyengine/bevy/blob/release-0.11.1/assets/models/monkey/Monkey.gltf) (same example) in version 0.11 seems to work,
though if I add a rotation quaternion, the monkey's rotation only updates upon restarting the app.

On a personal project, I'm listening for image events `AssetEvent<Image>`:

```rust
pub fn read_image_events(
    mut events: EventReader<AssetEvent<Image>>,
) {
    for event in events.read() {
        println!("Got event: {event:?}");
    }
}
```

This system runs in the Update schedule.
When starting up my game, my console logs the following:
```
Got event: Added { id: AssetId<bevy_render::texture::image::Image>{ index: 3, generation: 0} }
Got event: Added { id: AssetId<bevy_render::texture::image::Image>{ index: 5, generation: 0} }
Got event: Added { id: AssetId<bevy_render::texture::image::Image>{ index: 4, generation: 0} }
```

After updating / saving an image file image a couple of times, the console log looks like this:

```
Got event: Added { id: AssetId<bevy_render::texture::image::Image>{ index: 3, generation: 0} }  // Old
Got event: Added { id: AssetId<bevy_render::texture::image::Image>{ index: 5, generation: 0} }  // Old
Got event: Added { id: AssetId<bevy_render::texture::image::Image>{ index: 4, generation: 0} }  // Old
2024-08-10T20:33:26.687117Z  INFO bevy_asset::server: Reloading maps\tilesets\spring_forest.png because it has changed // New
2024-08-10T20:33:26.769548Z  INFO bevy_asset::server: Reloading maps\tilesets\spring_forest.png because it has changed // New
2024-08-10T20:33:39.859572Z  INFO bevy_asset::server: Reloading maps\tilesets\spring_forest.png because it has changed // New
2024-08-10T20:33:39.951233Z  INFO bevy_asset::server: Reloading maps\tilesets\spring_forest.png because it has changed // New
```

The four new lines at the end are bevy's internal logging.
Clearly, bevy is detecting the change, but for some reason, it does not fire `AssetEvent::<Image>::Modified` events.

The PBR models that I'm spawning do not seem to update either.
They do update if I restart the app, or do something that causes them to reload manually.

andriyDev:
```
This is a draft until #18357 is resolved - otherwise, hot-reloading scenes would leave them completely mangled.
```

# Objective

- Fixes #14698

## Solution

- Refactor all the `SceneSpawner` functions to include a `_dynamic` suffix where appropriate.
- Add non-DynamicScene versions of functions on `SceneSpawner`.
- Listen for `Scene` `AssetEvent`s and update the corresponding spawned scenes.

## Testing

- Ran the `hot_asset_reloading` example and mutated the GLTF to see it be updated
- TODO: Make sure this works properly after #18357 is fixed.

---

## Migration Guide

- Some of the methods on `SceneSpawner` have been renamed:
    - `despawn` -> `despawn_dynamic`
    - `despawn_sync` -> `despawn_dynamic_sync`
    - `update_spawned_scenes` -> `update_spawned_dynamic_scenes`
- In their place, we've also added `despawn`, `despawn_sync`, and `update_spawned_scenes` which all act on `Scene`s (as opposed to `DynamicScene`s).


andriyDev:
Looks like this PR isn't quite ready. I've got two meshes. Adding an empty node and saving after that causes a panic:

```
thread 'Compute Task Pool (0)' panicked at crates\bevy_transform\src\systems.rs:466:21:
assertion `left == right` failed
  left: 11v1#4294967307
 right: 12v1#4294967308
```

This also causes the entire app to freeze and blast my CPU usage to 50% (IIRC half the threads are for systems and half for assets, so I guess all the system threads are blasting).

Initial investigation shows the hierarchy is not set up correctly, but I'm not sure why yet.

schrottkatze:
Hi, uh, are you planning on continuing to work on this?
(if you forgot about this, i guess this comment could serve as a ping like "yes people do still need this" ^^', don't mean to be like passive-aggressive, i've just spent like an hour thinking i've been doing something wrong just to realize it's broken in bevy, actually and the only pr to fix it seems stalled >~<)

andriyDev:
Hi, so sorry I never actually went and updated the status here. 

This is somewhat blocked on #18418. I say somewhat because it's not clear to me that entities in a scene should be reused at all (which is what that issue is about).

Entities in a Scene do not have any semantic information. That means when you change the scene, the entities before and after have no correspondence. Imagine you have the following Scene (which comes from loading a GLTF)

```
Entity 1v0: Cube
Entity 2v0: Sphere
```

If you go and delete the cube, you get the Scene: 

```
Entity 1v0: Sphere
```

Note that the sphere now uses the cube's Entity. That means hot reloading would try to turn the cube entity into the sphere. Any components the cube had that the sphere doesn't will **not** be replaced (for example, if the cube had an IAmAnItemPickup component, now the sphere does. What if the sphere was actually the floor of your world!). This whole problem is exacerbated by relations. 

I think one approach is to just delete all the entities and spawn new ones. It makes hot reloading less powerful, but makes it behave more reasonably IMO. I can try drafting that up soon. The alternative is this being blocked on #18418 haha.

andriyDev:
Actually, maybe even that alternative is a non-starter since if you had something parented to your scene, it would be despawned when you despawned all your entities... So this PR might just be blocked...

andriyDev:
Alright, after some discussion in Discord, I've been convinced that deleting the scene and respawning it is better than doing nothing. I also added a test for scenes and dynamic scenes to make sure they continue to hot-reload.

Note: I've fixed the rendering change, so we don't need the label!

crates/bevy_scene/src/scene_spawner.rs
```diff
@@ -286,7 +310,29 @@ impl SceneSpawner {
...
+    ) -> Result<(), SceneSpawnError> {
+        for id in scene_ids {
+            if let Some(spawned_instances) = self.spawned_scenes.get(id) {
+                for instance_id in spawned_instances {
+                    if let Some(instance_info) = self.spawned_instances.get_mut(instance_id) {
+                        Self::spawn_sync_internal(world, *id, &mut instance_info.entity_map)?;
+                    }
+                }
+            }
+        }
```
janhohenheim: Nothing blocking, but this could be simplified a lot by using `.filter_map` for the spawned scenes, followed by `.flat_map` for the spawned instances, followed again by `filter_map` for the instance info ;)

janhohenheim (approved):
Looked through all commits and everything looks correct to me :)
Ping @Noxmore

crates/bevy_scene/src/scene_spawner.rs
andriyDev: I would prefer not to do this to keep it consistent with the dynamic one, and I'd prefer not to change that one to keep the diff small. bevy_scene could use a style pass IMO

ChristopherBiscardi:
seems to work for blender/skein.

https://github.com/user-attachments/assets/7ea9fa3c-5412-4554-b4e6-fa543513579d

One thing I noticed is that the entity list seems to grow with every reload. The code here loops over all `Query<&Character>` in `SceneInstanceReady` (Character is on the Cube object) and in normal code would only see one entity/component. I haven't dug too deep yet on it so it may be a trivial thing.

screenshot-2025-07-17-at-15 39 24@2x

edit: notably the Character Component here is added via basically `On<Add, GltfExtras>`

andriyDev:
@ChristopherBiscardi, could I see your observer code? Is it possible you're spawning an entity instead of adding a component? Hotreloading literally despawns every entity associated with the scene, so I don't know what entities your component could be living off of?

bevy_inspector_egui might also shed some light on this. I mean if the entire hierarchy is duplicated after hot reload, then we'll know this PR isn't working.

ChristopherBiscardi:
> @ChristopherBiscardi, could I see your observer code? Is it possible you're spawning an entity instead of adding a component? Hotreloading literally despawns every entity associated with the scene, so I don't know what entities your component could be living off of?

The observer code isn't afaik unless something changed significantly between 0.16 and main. The observer code is: https://github.com/rust-adventure/skein/blob/3d5949b7c21fc1b2e54f8414c3fb99e5c4885759/src/lib.rs#L272-L274 which hasn't substantially changed other than Trigger -> On

> bevy_inspector_egui might also shed some light on this. I mean if the entire hierarchy is duplicated after hot reload, then we'll know this PR isn't working.

Is there a version of bevy_inspector_egui that works on main?

---

I think the issue is that I have a scene_spawner in AssetEvent::LoadedWithDependencies, the LoadedWithDependencies is called every time the reload happens, which then creates another instance. So not an issue with the PR afaict.

on-load -> spawn is a shortcut I use for some examples that I'll just have to take a different approach for.

ChristopherBiscardi:
yeah that was it, I confirmed with a Local that prevented the spawn if already had been spawned. reloading itself is working fine.

alice-i-cecile:
I'm comfortable enough with this fix to merge it now. There's nothing crazy going on here, and the regression test is really helpful.

PR #18363, opened 2025-03-17
Revert "Transform Propagation Optimization: Static Subtree Marking (#18094)"

### LINKED ISSUE #18255: Transform propagation ignoring earlier branches
## Bevy version

main

## What you did
```
cargo run --example overflow
```

Looks fine:

Image

Now resize the window horizontally:

Image

## What went wrong

The global transforms for the text nodes aren't updated.

## Additional Info

Bisected to #18093 Transform Propagation Optimization: Static Subtree Marking

The descendants of the earlier branches are being skipped during transform propagation when there are multiple branches.

mockersf:
# Objective

- Fixes #18255
- Transform propagation is broken in some cases

## Solution

- Revert #18093


alice-i-cecile:
We should investigate redoing #18093: it's fundamentally a sensible strategy to optimize transform propagation, and I'm not convinced that all of its performance gains are due to this bug.

Issue #18366, opened 2025-03-17
Clear color sometimes seems to be applied to wrong RenderTarget::Image when using multiple cameras


fatho:
## Bevy version

0.15.3

## \[Optional\] Relevant system information

Arch Linux
KDE Plasma 6.3.3
Mesa 25.0.1

If your bug is rendering-related, copy the adapter info that appears when you run Bevy.

```ignore
`AdapterInfo { name: "AMD Radeon RX 7900 XTX", vendor: 4098, device: 29772, device_type: DiscreteGpu, driver: "AMD open-source driver", driver_info: "2025.Q1.3 (LLPC)", backend: Vulkan }`
```

## What you did

Unfortunately, I don't quite know when it started, since I'm mostly running the debug build, but the problem only manifests itself in the release binary.

I have the following rendering setup for my game (distilled into a single file below that also reproduces the problem):

1. a **light** camera to which I render all lights to an offscreen texture. Using a light gray as clear color to simulate the ambient light.
2. a foreground **stage** camera that renders all entities to an offscreen texture that should be lighted. This one is cleared with `LinearRgba::new(0.0, 0.0, 0.0, 0.0)` (i.e. all black with zero alpha, since it shouldn't cover any background in the **scene** below)
3. a **scene** camera that renders entities that should not be lighted to an offscreen texture, as well as the lighted stage with a custom shader that simply does a point-wise multiplication of the **stage** pixels and the corresponding **light** pixels. This one is cleared with `Color::BLACK`.
4. the final **screen** to which I render the scene (the idea was to have parts of the postprocessing here) with a `ColorMaterial`, and on top of that any other entities I might need. Since I expect the **scene** output to fully cover this camera, I clear this one with a bright magenta so that it's easy to see where the rendering goes wrong. This one renders to the window.

Or, as a picture (rects are cameras, rounded rects are rendered entities with materials)

```mermaid
graph TD;
  Light["Light (clear: gray, target: texture)"] --> LightedStage(["Light * Stage material"]);
  Stage["Stage (clear: transparent, target: texture)"] --> LightedStage;
  LightedStage --> Scene["Scene (clear: black, target: texture)"];
  Scene --> SceneMat(["ColorMaterial"]);
  SceneMat --> Screen["Screen (clear: magenta, target: window)"];
```

## What went wrong

Every couple seconds, there's a frame or two where it appears as though the clear colors from the above cameras are not applied to the right render targets.
[...]

Issue #16185, opened 2024-10-31
Shadows in 3D scene are flickering


lonegamedev:
## Bevy version

0.14.2

## \[Optional\] Relevant system information

Arch Linux

AdapterInfo { name: "AMD Radeon RX 7900 XTX", vendor: 4098, device: 29772, device_type: DiscreteGpu, driver: "AMD open-source driver", driver_info: "2024.Q3.3 (LLPC)", backend: Vulkan }

## What you did

Just run plain 3D scene example.

## What went wrong

Shadows are flickering.

## Additional information

![Image](https://github.com/user-attachments/assets/df84a0ea-1c19-46c1-b7b9-95020029a1c1)

jpteb:
I experience the same issue under NixOS and `bevy v0.15.0-dev`.

```
SystemInfo { os: "Linux 24.11 NixOS", kernel: "6.11.5", cpu: "AMD Ryzen 7 5800X 8-Core Processor", core_count: "8", memory: "31.3 GiB" }
AdapterInfo { name: "AMD Radeon RX 7800 XT", vendor: 4098, device: 29822, device_type: DiscreteGpu, driver: "AMD open-source driver", driver_info: "2024.Q3.2 (LLPC)", backend: Vulkan }
```

paholg:
I am also seeing this after updating from Bevy 0.13 to 0.14. I am also on NixOS.

Using `git bisect` in the bevy repo and repeatedly running `cargo run --example 3d_shapes`, I have found that this commit introduced the flicker: f4dab8a.

Strangely, that commit is included in the 0.13.0 release, yet in my project the shadows don't flicker until the 0.14 update. If desired, you can test it: https://github.com/paholg/gam/tree/216cd246c8c8984cf913b5ffb30c67a269e09571, run with `cargo run --bin client`.

Finally, if I compile my project for windows and run it with wine, the shadows do not flicker. However, if I take a version compiled on Ubuntu and run it with steam-run, there is flickering.

I would be happy to investigate further, but I am afraid that I am miles out of my league when it comes to anything related to rendering.

```
AdapterInfo { name: "AMD Radeon RX 7900 XTX", vendor: 4098, device: 29772, device_type: DiscreteGpu, driver: "AMD open-source driver", driver_info: "2024.Q3.3 (LLPC)", backend: Vulkan }
```

jpteb:
[...]

PR #18368, opened 2025-03-17
Force serial command encoding on Linux/amdvlk

fatho:
# Objective

Fixes https://github.com/bevyengine/bevy/issues/18366 which seems to have a similar underlying cause than the already closed (but not fixed) https://github.com/bevyengine/bevy/issues/16185.

## Solution

For Windows with the AMD vulkan driver, there was already a hack to force serial command encoding, which prevented these issues. The Linux version of the AMD vulkan driver seems to have similar issues than its Windows counterpart, so I extended the hack to also cover AMD on Linux.

I also removed the mention of `wgpu` since it was already outdated, and doesn't seem to be relevant to the core issue (the AMD driver being buggy).

## Testing

  - I ran the `3d_scene` example, which on `main` produced the flickering shadows on Linux with the amdvlk driver, while it no longer does with the workaround applied.
  - Not sure.
  - Requires a Linux system with an AMD card and the AMDVLK driver.
  - My change should only affect Linux, where I did test it.

### DIFF
--- a/crates/bevy_render/src/renderer/mod.rs
+++ b/crates/bevy_render/src/renderer/mod.rs
@@ -391,12 +391,13 @@ impl<'w> RenderContext<'w> {
         adapter_info: AdapterInfo,
         diagnostics_recorder: Option<DiagnosticsRecorder>,
     ) -> Self {
-        // HACK: Parallel command encoding is currently bugged on AMD + Windows + Vulkan with wgpu 0.19.1
-        #[cfg(target_os = "windows")]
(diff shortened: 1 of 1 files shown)


alice-i-cecile (approved):
Seems sensible. I take it you've tested this locally?

Elabajaba (approved):
Tested and confirmed that this doesn't `force_serial` on RADV (the more popular AMD linux vulkan driver) where multithreaded command encoding isn't broken.

I didn't bother setting up an amdvlk setup to test, as I assumed you already did.

PR #18370, opened 2025-03-17
Add tooltip with required components for trait tags on docs.rs

weihnaxbaum:
# Objective

To see a component's required components in the docs, you first have to find the `Component` impl which can be annoying.

## Solution

Add a tooltip to the `Component` and `Immutable Component` [trait tags](https://github.com/bevyengine/bevy/pull/17758) which lists the required components. The components in that list link to their own doc page.

## Testing

Build the docs locally:
```bash
RUSTDOCFLAGS="--html-after-content docs-rs/trait-tags.html --cfg docsrs_dep" RUSTFLAGS="--cfg docsrs_dep" cargo doc --no-deps -p <some_bevy_package>
```

---

## Showcase

Screenshot 2025-03-17 215044


mockersf:
I don't think we want to go in that direction.

The plan is to upstream the tags to rustdoc, and I don't see this tooltip as having enough general value for other crates.
This is also not very discoverable

Weihnachtsbaum:
> The plan is to upstream the tags to rustdoc

I didn't know that. I guess the tooltips could still be used for the bevy docs with the custom html, even if the trait tags get upstreamed, but I see that it could complicate things.
> This is also not very discoverable

If we want people to know about it, we can just mention it in the release notes. Also I think that actually a lot of users would just find it out by accidentally hovering over the tags. And even if that isn't the case, I still think that something that's useful should be kept, even if not everyone knows about it.

But of course cost of maintaining vs. value has to be considered, which I can't assess.

PR #18373, opened 2025-03-17
Fixes for WESL on Windows

tychedelia:
# Objective

WESL was broken on windows.

## Solution

- Upgrade to `wesl_rs` 1.2.
- Fix path handling on windows.
- Improve example for khronos demo this week.

### DIFF
--- a/crates/bevy_render/src/render_resource/shader.rs
+++ b/crates/bevy_render/src/render_resource/shader.rs
@@ -164,23 +164,34 @@ impl Shader {
         match import_path {
             ShaderImport::AssetPath(asset_path) => {
                 let asset_path = std::path::PathBuf::from(&asset_path);
-                // Resolve and normalize the path
-                let asset_path = asset_path.canonicalize().unwrap_or(asset_path);
-                // Strip the asset root
+
+                // Get the base path and canonicalize it to match the format of the asset path
                 let mut base_path = bevy_asset::io::file::FileAssetReader::get_base_path();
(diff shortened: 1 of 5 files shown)


crates/bevy_render/src/render_resource/shader.rs
```diff
@@ -164,23 +164,34 @@ impl Shader {
...
+                        match canonical_asset_path.strip_prefix(&base_path) {
+                            Ok(rel_path) => rel_path.to_path_buf(),
+                            Err(_) => canonical_asset_path,
+                        }
+                    }
+                    Err(_) => asset_path,
+                };
+
+                // Create the shader import path - always starting with "/"
+                let shader_path = std::path::Path::new("/").join(&relative_path);
```
superdump, line 185: This looks a bit odd... why starting with `/` if you're going to join it with a _relative_ path?

superdump: I just saw: `// Wesl paths are provided as absolute relative to the asset root` that was removed above. That was useful information I guess. Interesting.

superdump (approved):

mockersf (approved):
it was also broken on macOS, this fixes it

crates/bevy_render/src/render_resource/shader.rs
tychedelia: Oops, yeah, I shouldn't have removed that. This in general is kinda messy path mangling atm, it needs to be improved more comprehensively as I mentioned on discord.

PR #18375, opened 2025-03-17
Fix Formatting of Optimisation Table

bushrat011899:
# Objective

New markdown linter doesn't like this table.

## Solution

Fixed it.

## Testing

CI


alice-i-cecile (approved):
Okay CI, sure.

alice-i-cecile:
Ugh, hold on. I think you need to change the file that *generates* this table to avoid the secondary failure.

bushrat011899:
Ah yeah of course, thankfully that's just a nice template file.

alice-i-cecile (approved):

PR #18376, opened 2025-03-17
add AlignedRawBufferVec

IceSentry:
# Objective

- The ShaderType derive is pretty expansive so we need alternatives. Using RawBufferVec works well in many cases, but not when buffers have alignment rules like uniform buffers.

## Solution

- Add `AlignedRawBufferVec` which is almost identical to `RawBufferVec` but attempts to respect alignment rules.
- Uses encase `AlignmentValue` to handle alignment. This isn't super necessary and could be trivially upstreamed but as long as encase is in tree it doesn't matter that much.

## Testing

- Tested in a fork of the volumetric fog that uses this new buffer

### DIFF
--- a/crates/bevy_render/src/render_resource/buffer_vec.rs
+++ b/crates/bevy_render/src/render_resource/buffer_vec.rs
@@ -7,7 +7,7 @@ use crate::{
 };
 use bytemuck::{must_cast_slice, NoUninit};
 use encase::{
-    internal::{WriteInto, Writer},
+    internal::{AlignmentValue, WriteInto, Writer},
     ShaderType,
 };
 use wgpu::{BindingResource, BufferAddress, BufferBinding, BufferUsages};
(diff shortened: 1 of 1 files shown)


crates/bevy_render/src/render_resource/buffer_vec.rs
```diff
@@ -511,3 +511,181 @@ where
...
+///
+/// The contained data is stored in system RAM. Calling [`reserve`](RawBufferVec::reserve)
+/// allocates VRAM from the [`RenderDevice`].
+/// [`write_buffer`](RawBufferVec::write_buffer) queues copying of the data
+/// from system RAM to VRAM.
+///
+/// Unlike [`RawBufferVec`] this will add padding to respect alignment rules if necessary when
+/// adding data to it. This is useful when working with buffers that have alignment rules
+/// like uniform buffers.
+pub struct AlignedRawBufferVec<T: NoUninit> {
```
superdump, line 530: Maybe call it DynamicRawBufferVec to follow our conventions? Although I never really liked them, DynamicOffsetRawBufferVec is a bit long. :)

crates/bevy_render/src/render_resource/buffer_vec.rs
```diff
@@ -511,3 +511,181 @@ where
...
+
+    /// Adds a new value and returns its offset in bytes.
+    ///
+    /// This will convert the value to bytes and add any padding it needs to respect alignment
+    pub fn push(&mut self, value: T) -> usize {
+        let index = self.values.len();
+        let bytes: &[u8] = bytemuck::bytes_of(&value);
+        self.values.extend(bytes);
+        if self.required_padding > 0 {
+            for _ in 0..self.required_padding {
```
superdump, line 623: You could use Vec’s resize to len + required padding.

crates/bevy_render/src/render_resource/buffer_vec.rs
IceSentry: I don't feel like it would represent what it is correctly. It's very much a `RawBufferVec` but with padding for offset alignment rules. Dynamic doesn't really imply that to me. I never liked the `Dynamic*` types because it never made sense to me why they would be more dynamic than the non dynamic types.

Also, with the eventual goal to remove/replace encase, these Dynamic* types would most likely disappear since they only exist to wrap encase stuff.

crates/bevy_render/src/render_resource/buffer_vec.rs
IceSentry: Good idea, done.

IceSentry:
I added a few more methods from `RawBufferVec` so it can be a bit more useful. I didn't implement `.pop()` because dealing with the alignment and conversion was annoying. We can add it whenever it's actually needed.

IceSentry:
I still have plans in the general area of what this PR is doing, but it will be in a different direction so I'll close this and start fresh in another PR.

For the record, I did test using this to replace many instances of DynamicUniformBuffer and it reduces the overall compile time by 1s.

PR #18378, opened 2025-03-18
Support skipping Relationship on_replace hooks

### LINKED ISSUE #18357: `insert_with_relationship_insert_hook_mode` still runs the relationship `on_replace` implementation.

cart:
# Objective

Fixes #18357

## Solution

Generalize `RelationshipInsertHookMode` to `RelationshipHookMode`, wire it up to on_replace execution, and use it in the `Relationship::on_replace` hook.

### DIFF
--- a/crates/bevy_ecs/src/bundle.rs
+++ b/crates/bevy_ecs/src/bundle.rs
@@ -18,7 +18,7 @@ use crate::{
     observer::Observers,
     prelude::World,
     query::DebugCheckedUnwrap,
-    relationship::RelationshipInsertHookMode,
+    relationship::RelationshipHookMode,
     storage::{SparseSetIndex, SparseSets, Storages, Table, TableRow},
     world::{unsafe_world_cell::UnsafeWorldCell, EntityWorldMut, ON_ADD, ON_INSERT, ON_REPLACE},
 };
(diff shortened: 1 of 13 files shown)


alice-i-cecile (approved):
The fix makes sense to me: just an inconsistency. I'd prefer if we had a more representative test in here (that would have caught this bug in the first place), but I won't block on it.

andriyDev (approved):
I also tested this with #18358, and this fixes the hierarchy part at least!

PR #18380, opened 2025-03-18
Remote entity reservation v4

### LINKED ISSUE #18003: Reserve entities from async

ElliottjPierce:
fixes #18003

# Objective

Long story short: we need to be able to reserve entities from outside `&World`. This benefits both assets as entities and components as entities. 

This is an alternative to #18195 and #18266, v1 and v2 respectively. (If you're wondering where v3 went, it was some failed experimenting. I skipped the number to prevent my own confusion.)

## Solution

Thanks to everyone who helped brainstorm on discord.

All entity meta is kept in a single unlocked `Vec<EntityMeta>`. We split pending entities into multiple `PendingEntitiesChunk`s. One chunk is constantly pulled from while another is extended. When the pulling one is empty, we check if its worth swapping, and if so, the chunks shift, opening up more pending entities for reservation. We don't check for swapping every time; the frequency is user configurable. Checking to swap takes an additional 2 atomic operations when the swap is not needed, but can take over 10 atomic operations when it is needed (which is less than once per flush). When we don't check swapping or when we reserve directly on `&Entities` (since that never needs to check for swapping) it now takes two atomic operations instead of one.

When we allocate an entity directly, we pop from the new `owned: Vec<Entity>`, and if it's empty, we reserve some more entities in bulk. These entities are not flushed. The number of entities we reserve at once is `ideal_owned` and this is user configurable.

To ensure we don't loose entities anywhere, flushing drains any `owned` entities beyond the `ideal_owned` amount into the pending chunk.

It is possible for entity reservations to extend the `meta` vec instead of pulling from available pending entities. This happens, for example, when the pending entities are split between two chunks and we never check to swap chunks. In remote reservations, this behavior would be opt-in, and in `&Entities` reservations, this is guaranteed to be fixed on the next flush after the current pending list is exhausted. If we do some thread_local magic, we could eliminate this issue for `&Entities`, but that may introduce more overhead than it's worth.

## Small Fix

`Entity::index` is a u32, but on main, the final index can never be reserved. This is because `ReserveEntitiesIterator::new_indices` was a `Range<u32>`. Which doesn't include its end. So index `u32::MAX` could never be reserved. This is fixed in this PR, as it was a prerequisite fix that I discovered here. If we don't merge this, whatever solution we go with, I will make sure to follow up with a PR to fix this. (Or we can make `Entities::index` be `NonMax<u32>`, freeing up another niche.)

## Not implemented

`alloc_at` and `alloc_at_without_replacement` are not supported in this PR. This was one of the pain points and limiting factors in my previous attempts, but since it is being removed anyway in 0.17 (#18148), I chose not to support it.

Additionally, many of the various `len` methods were harder to implement since the `PendingEntitiesChunks` can be scattered to the four winds via Arcs. As a result, they are much slower, but these are really a diagnostic tool anyway, so it's not that big of a performance problem. 

## Practical Change

It used to be that a brand new `Entities` would yield sequential indices via `alloc`. This was never a guarantee, but many tests relied on this ordering. The ordering is still deterministic but is no longer sequential or aligned. In other words, if you allocate 5 entities. You might get back indices `[0, 4, 2, 6, 7]`. Generally entities like this are opaque and this should not affect users. However, tests and benchmarks that depend on this previous, not guaranteed behavior had to be reworked.

One practical effect this may have is that systems that run through `Entity`s sequentially, will have a different order. This may affect query orders, observer ordering, etc. None of these things were guarantees before, but this may expose some existing bugs in those systems.

## Future Work

Flushing is now completely unnecessary for many of the interactions with [`Entities`] (ex: free, alloc, etc). We still need to do it before applying command queues, etc,  but we don't need to flush before allocations or freeing. If we go in this direction, we should explore relaxing those requirements for performance.

## Testing

Functionality hasn't changed, so no new tests were added. A few old tests were revised because of the practical changes.

However, it makes sense to add tests for remote registration, so if anyone has specific concerns, I'd be happy to write a test for it.

---

## Showcase

Here's a tiny taste:

```rust
let mut wold = World::new();

// This `RemoteEntities` is 'static thanks to Arcs.
let entities: RemoteEntities = world.entities().get_remote_entities();

// This `EntityReserver` is also 'static thanks to Arcs.
let mut reserver: EntityReserver = entities.reserver();

let reserved = reserver.reserve_entity();
world.flush();
```

I would imagine, for example, the asset system hangs onto a `RemoteEntities` and passes out a `&mut EntityReserver` as part of asset loading contexts. I'm no expert in that area, but hopefully that give you an idea.

## Costs

If we are going to keep the `meta` vec all together (which we need to), I think this is about as fast as we can get. We need the extra pointer indirection because, for this to work, we need the `Arc` at some point to keep things `'static`. We need two atomic operations instead of the one `cursor.fetch_sub(num)` because we can't do `-cursor + meta.len()` without `meta` being in the `Arc` too (which would be either very slow since we'd need atomics for setting `EntityLocation`, or a massive safety headache). I may be able to speed up some of the details here, but I don't think a faster design is logically possible at this point. If this is too slow, we should reconsider other designs besides remote reservation. (Maybe make every `Asset` be `MapEntities`??). We're going to have to eat this performance cost somewhere.

### Benchmarks

```txt
group                                                                                                     main_baseline                            remote_entities_v4_baseline
-----                                                                                                     -------------                            ---------------------------
add_remove/sparse_set                                                                                     1.00    548.7±9.81µs        ? ?/sec      2.03  1113.6±59.79µs        ? ?/sec
add_remove/table                                                                                          1.00   763.8±22.88µs        ? ?/sec      1.70  1298.8±37.59µs        ? ?/sec
add_remove_big/sparse_set                                                                                 1.00   578.8±11.76µs        ? ?/sec      1.88  1089.7±19.90µs        ? ?/sec
add_remove_big/table                                                                                      1.00      2.7±0.05ms        ? ?/sec      1.23      3.3±0.30ms        ? ?/sec
add_remove_very_big/table                                                                                 1.00     33.4±1.36ms        ? ?/sec      1.07     35.8±1.68ms        ? ?/sec
added_archetypes/archetype_count/10000                                                                    1.22      5.3±0.43ms        ? ?/sec      1.00      4.3±0.05ms        ? ?/sec
added_archetypes/archetype_count/500                                                                      1.06   334.4±33.57µs        ? ?/sec      1.00   316.3±17.90µs        ? ?/sec
added_archetypes/archetype_count/5000                                                                     1.20      2.4±0.21ms        ? ?/sec      1.00  1966.2±203.19µs        ? ?/sec
all_changed_detection/50000_entities_ecs::change_detection::Table                                         1.31     60.1±1.97µs        ? ?/sec      1.00     45.7±1.25µs        ? ?/sec
all_changed_detection/5000_entities_ecs::change_detection::Table                                          1.22      5.5±0.67µs        ? ?/sec      1.00      4.5±0.13µs        ? ?/sec
build_schedule/1000_schedule                                                                              1.00   949.5±20.88ms        ? ?/sec      1.06  1004.4±32.29ms        ? ?/sec
build_schedule/1000_schedule_no_constraints                                                               1.10     20.6±0.98ms        ? ?/sec      1.00     18.8±1.04ms        ? ?/sec
build_schedule/100_schedule                                                                               1.00      5.8±0.10ms        ? ?/sec      1.50      8.8±3.18ms        ? ?/sec
build_schedule/500_schedule                                                                               1.00   173.8±14.13ms        ? ?/sec      1.06   183.8±15.92ms        ? ?/sec
busy_systems/02x_entities_15_systems                                                                      1.00    184.2±3.10µs        ? ?/sec      1.07    196.6±5.23µs        ? ?/sec
busy_systems/04x_entities_15_systems                                                                      1.00    321.6±2.29µs        ? ?/sec      1.05    338.9±9.42µs        ? ?/sec
contrived/01x_entities_03_systems                                                                         1.00     17.5±0.93µs        ? ?/sec      1.23     21.5±4.21µs        ? ?/sec
contrived/01x_entities_06_systems                                                                         1.00     33.0±1.52µs        ? ?/sec      1.07     35.1±1.68µs        ? ?/sec
contrived/01x_entities_09_systems                                                                         1.00     44.3±1.15µs        ? ?/sec      1.07     47.5±2.35µs        ? ?/sec
contrived/01x_entities_12_systems                                                                         1.00     57.0±2.06µs        ? ?/sec      1.08     61.8±2.45µs        ? ?/sec
contrived/01x_entities_15_systems                                                                         1.00     70.0±1.71µs        ? ?/sec      1.06     74.4±2.44µs        ? ?/sec
contrived/03x_entities_03_systems                                                                         1.00     32.9±1.75µs        ? ?/sec      1.07     35.1±2.09µs        ? ?/sec
contrived/03x_entities_12_systems                                                                         1.00     99.2±4.36µs        ? ?/sec      1.06    105.3±4.62µs        ? ?/sec
contrived/04x_entities_12_systems                                                                         1.00    122.3±6.43µs        ? ?/sec      1.08    131.8±6.23µs        ? ?/sec
contrived/04x_entities_15_systems                                                                         1.00    152.9±8.84µs        ? ?/sec      1.09    166.2±9.05µs        ? ?/sec
contrived/05x_entities_03_systems                                                                         1.00     46.1±1.62µs        ? ?/sec      1.06     49.0±2.82µs        ? ?/sec
contrived/05x_entities_09_systems                                                                         1.00    111.5±4.18µs        ? ?/sec      1.13   125.6±13.63µs        ? ?/sec
contrived/05x_entities_12_systems                                                                         1.00    146.7±2.63µs        ? ?/sec      1.07    157.5±9.93µs        ? ?/sec
despawn_world/10000_entities                                                                              1.00     55.1±1.91µs        ? ?/sec      2.76   152.0±10.81µs        ? ?/sec
despawn_world/1000_entities                                                                               1.00      5.6±0.47µs        ? ?/sec      2.68     15.0±0.20µs        ? ?/sec
despawn_world/100_entities                                                                                1.00    547.3±6.37ns        ? ?/sec      1.64   896.8±37.36ns        ? ?/sec
despawn_world/10_entities                                                                                 1.00     59.4±1.53ns        ? ?/sec      1.57     93.3±1.17ns        ? ?/sec
despawn_world/1_entities                                                                                  1.00      6.1±0.14ns        ? ?/sec      1.54      9.5±0.11ns        ? ?/sec
ecs::entity_cloning::hierarchy_many/clone                                                                 1.00   227.1±28.05µs 1565.2 KElem/sec    1.13   256.7±90.32µs 1385.0 KElem/sec
ecs::entity_cloning::hierarchy_many/reflect                                                               1.00   412.2±25.90µs 862.3 KElem/sec     1.57   645.7±22.05µs 550.5 KElem/sec
ecs::entity_cloning::hierarchy_tall/clone                                                                 1.00     11.9±0.47µs  4.1 MElem/sec      1.27     15.1±2.56µs  3.2 MElem/sec
ecs::entity_cloning::hierarchy_tall/reflect                                                               1.00     14.4±0.51µs  3.4 MElem/sec      1.41     20.2±0.41µs  2.4 MElem/sec
ecs::entity_cloning::hierarchy_wide/clone                                                                 1.00      9.6±0.72µs  5.1 MElem/sec      1.31     12.6±0.39µs  3.9 MElem/sec
ecs::entity_cloning::hierarchy_wide/reflect                                                               1.00     12.7±1.21µs  3.8 MElem/sec      1.46     18.5±2.47µs  2.6 MElem/sec
ecs::entity_cloning::single/clone                                                                         1.00   622.4±96.89ns 1569.0 KElem/sec    1.16   722.8±73.83ns 1351.2 KElem/sec
ecs::entity_cloning::single/reflect                                                                       1.00  1088.1±77.21ns 897.5 KElem/sec     1.66  1806.5±81.77ns 540.6 KElem/sec
empty_archetypes/iter/100                                                                                 1.07      8.9±1.08µs        ? ?/sec      1.00      8.4±0.55µs        ? ?/sec
empty_archetypes/iter/500                                                                                 1.00      8.6±0.76µs        ? ?/sec      1.06      9.1±0.94µs        ? ?/sec
empty_archetypes/par_for_each/1000                                                                        1.00     11.4±0.72µs        ? ?/sec      1.06     12.1±0.66µs        ? ?/sec
empty_commands/0_entities                                                                                 1.00      4.0±0.12ns        ? ?/sec      2.06      8.2±0.18ns        ? ?/sec
empty_systems/000_systems                                                                                 1.00      4.3±0.04ns        ? ?/sec      1.07      4.6±0.10ns        ? ?/sec
empty_systems/003_systems                                                                                 1.00      8.6±0.82µs        ? ?/sec      1.06      9.1±1.00µs        ? ?/sec
empty_systems/015_systems                                                                                 1.07     15.2±1.84µs        ? ?/sec      1.00     14.2±1.10µs        ? ?/sec
event_propagation/four_event_types                                                                        1.00    537.8±8.58µs        ? ?/sec      1.08   582.8±19.36µs        ? ?/sec
event_propagation/single_event_type                                                                       1.00   759.4±28.43µs        ? ?/sec      1.10   832.8±19.57µs        ? ?/sec
event_propagation/single_event_type_no_listeners                                                          1.12   318.7±19.31µs        ? ?/sec      1.00    285.0±9.58µs        ? ?/sec
events_send/size_16_events_100                                                                            1.07   146.4±27.91ns        ? ?/sec      1.00   136.8±23.07ns        ? ?/sec
fake_commands/2000_commands                                                                               1.00     13.0±1.47µs        ? ?/sec      1.49     19.4±0.26µs        ? ?/sec
fake_commands/4000_commands                                                                               1.00     24.9±0.34µs        ? ?/sec      1.56     38.9±0.65µs        ? ?/sec
fake_commands/6000_commands                                                                               1.00     38.4±1.50µs        ? ?/sec      1.52     58.3±1.34µs        ? ?/sec
fake_commands/8000_commands                                                                               1.00    54.2±12.12µs        ? ?/sec      1.43     77.5±1.02µs        ? ?/sec
few_changed_detection/5000_entities_ecs::change_detection::Sparse                                         1.09      4.3±0.76µs        ? ?/sec      1.00      4.0±0.29µs        ? ?/sec
few_changed_detection/5000_entities_ecs::change_detection::Table                                          1.07      4.9±0.25µs        ? ?/sec      1.00      4.6±0.28µs        ? ?/sec
insert_commands/insert                                                                                    1.00   397.8±20.41µs        ? ?/sec      1.21   482.5±34.93µs        ? ?/sec
insert_simple/base                                                                                        1.00   555.1±24.53µs        ? ?/sec      1.07   592.8±35.95µs        ? ?/sec
insert_simple/unbatched                                                                                   1.00  1010.0±148.12µs        ? ?/sec     1.35  1362.8±129.86µs        ? ?/sec
iter_fragmented_sparse/base                                                                               1.00      6.5±0.06ns        ? ?/sec      1.05      6.8±0.09ns        ? ?/sec
iter_fragmented_sparse/foreach_wide                                                                       1.00     26.2±0.53ns        ? ?/sec      1.39     36.4±0.89ns        ? ?/sec
iter_fragmented_sparse/wide                                                                               1.00     32.6±1.71ns        ? ?/sec      1.22     39.9±0.48ns        ? ?/sec
iter_simple/wide_sparse_set                                                                               1.10    95.2±22.47µs        ? ?/sec      1.00     86.8±0.77µs        ? ?/sec
multiple_archetypes_none_changed_detection/100_archetypes_10000_entities_ecs::change_detection::Sparse    1.00   999.3±72.74µs        ? ?/sec      1.15  1153.4±99.15µs        ? ?/sec
multiple_archetypes_none_changed_detection/100_archetypes_1000_entities_ecs::change_detection::Sparse     1.00     75.6±5.95µs        ? ?/sec      1.07     80.9±4.91µs        ? ?/sec
multiple_archetypes_none_changed_detection/20_archetypes_1000_entities_ecs::change_detection::Sparse      1.00     14.0±0.43µs        ? ?/sec      1.07     14.9±0.65µs        ? ?/sec
multiple_archetypes_none_changed_detection/5_archetypes_100_entities_ecs::change_detection::Sparse        1.15   387.6±58.56ns        ? ?/sec      1.00   335.8±26.86ns        ? ?/sec
no_archetypes/system_count/100                                                                            1.05  1148.5±15.50ns        ? ?/sec      1.00  1090.4±15.86ns        ? ?/sec
no_archetypes/system_count/20                                                                             1.06    224.6±3.19ns        ? ?/sec      1.00    212.4±1.05ns        ? ?/sec
no_archetypes/system_count/60                                                                             1.06    691.4±5.62ns        ? ?/sec      1.00    653.4±6.05ns        ? ?/sec
observe/trigger_simple                                                                                    1.12    472.1±8.46µs        ? ?/sec      1.00    420.8±1.86µs        ? ?/sec
par_iter_simple/hybrid                                                                                    1.11    62.6±12.98µs        ? ?/sec      1.00     56.2±0.77µs        ? ?/sec
run_condition/yes/016_systems                                                                             1.00     14.5±1.14µs        ? ?/sec      1.08     15.7±2.53µs        ? ?/sec
run_condition/yes_using_query/011_systems                                                                 1.06     12.9±1.34µs        ? ?/sec      1.00     12.2±0.72µs        ? ?/sec
run_condition/yes_using_query/016_systems                                                                 1.05     15.7±1.02µs        ? ?/sec      1.00     14.9±0.92µs        ? ?/sec
run_condition/yes_using_query/061_systems                                                                 1.05     61.1±3.64µs        ? ?/sec      1.00     57.9±1.95µs        ? ?/sec
run_condition/yes_using_resource/046_systems                                                              1.06     42.3±2.24µs        ? ?/sec      1.00     39.8±1.01µs        ? ?/sec
run_condition/yes_using_resource/061_systems                                                              1.06     56.7±2.81µs        ? ?/sec      1.00     53.5±1.28µs        ? ?/sec
run_condition/yes_using_resource/076_systems                                                              1.06     74.6±2.42µs        ? ?/sec      1.00     70.5±3.04µs        ? ?/sec
run_condition/yes_using_resource/081_systems                                                              1.06     81.6±2.54µs        ? ?/sec      1.00     77.0±1.85µs        ? ?/sec
run_empty_schedule/MultiThreaded                                                                          1.00      4.3±0.08ns        ? ?/sec      1.06      4.5±0.08ns        ? ?/sec
schedule/base                                                                                             1.00     22.1±0.64µs        ? ?/sec      1.08     23.8±1.38µs        ? ?/sec
sized_commands_0_bytes/2000_commands                                                                      1.00     10.8±0.66µs        ? ?/sec      1.67     18.1±0.28µs        ? ?/sec
sized_commands_0_bytes/4000_commands                                                                      1.00     21.3±0.83µs        ? ?/sec      1.70     36.2±0.56µs        ? ?/sec
sized_commands_0_bytes/6000_commands                                                                      1.00     31.8±1.11µs        ? ?/sec      1.71     54.3±1.56µs        ? ?/sec
sized_commands_0_bytes/8000_commands                                                                      1.00     42.1±0.38µs        ? ?/sec      1.71     72.1±0.87µs        ? ?/sec
sized_commands_12_bytes/2000_commands                                                                     1.00     11.7±1.71µs        ? ?/sec      1.62     18.9±0.26µs        ? ?/sec
sized_commands_12_bytes/4000_commands                                                                     1.00     22.9±0.73µs        ? ?/sec      1.65     37.8±0.49µs        ? ?/sec
sized_commands_12_bytes/6000_commands                                                                     1.00     34.5±0.31µs        ? ?/sec      1.65     57.0±1.40µs        ? ?/sec
sized_commands_12_bytes/8000_commands                                                                     1.00     49.1±0.58µs        ? ?/sec      1.65     80.8±1.98µs        ? ?/sec
sized_commands_512_bytes/2000_commands                                                                    1.00     46.8±2.74µs        ? ?/sec      1.21     56.4±1.60µs        ? ?/sec
sized_commands_512_bytes/4000_commands                                                                    1.00     95.3±4.00µs        ? ?/sec      1.18    112.6±2.80µs        ? ?/sec
sized_commands_512_bytes/6000_commands                                                                    1.00    143.7±8.77µs        ? ?/sec      1.19    170.7±7.69µs        ? ?/sec
sized_commands_512_bytes/8000_commands                                                                    1.00    191.8±9.68µs        ? ?/sec      1.20   229.3±39.90µs        ? ?/sec
spawn_commands/2000_entities                                                                              1.00    136.2±4.51µs        ? ?/sec      1.50    204.6±8.88µs        ? ?/sec
spawn_commands/4000_entities                                                                              1.00    262.0±5.92µs        ? ?/sec      1.67   436.3±20.16µs        ? ?/sec
spawn_commands/6000_entities                                                                              1.00   419.6±23.31µs        ? ?/sec      1.64   687.9±26.49µs        ? ?/sec
spawn_commands/8000_entities                                                                              1.00   568.0±29.60µs        ? ?/sec      1.67  950.9±176.23µs        ? ?/sec
spawn_world/10000_entities                                                                                1.00   425.1±48.82µs        ? ?/sec      1.16   492.2±43.99µs        ? ?/sec
spawn_world/1000_entities                                                                                 1.00     44.6±4.48µs        ? ?/sec      1.11     49.5±4.74µs        ? ?/sec
spawn_world/100_entities                                                                                  1.00      4.3±0.52µs        ? ?/sec      1.16      5.0±0.58µs        ? ?/sec
spawn_world/1_entities                                                                                    1.00     45.6±5.03ns        ? ?/sec      1.09     49.5±4.42ns        ? ?/sec
world_query_iter/50000_entities_table                                                                     1.00     23.9±0.39µs        ? ?/sec      1.24     29.5±0.24µs        ? ?/sec
```
These are from running:
- empty_archetypes::benches
- iteration::benches
- world::benches
on main vs here.

`cargo bench --bench ecs -p benches -- --noplot --save-baseline remote_entities_v4_baseline` (I have other groups commented out.)

I've filtered out benchmarks with less than 5% difference.

### Interpretation

In general, most of these benchmarks fall somewhere between "ok-ish" to"uh-oh". But there's some good news.

The good news: This branch beats main (usually by 7-12%) on a little over a third of the benchmarks. Of course, it then looses to main by usually 10-15% on the other two thirds. That sounds about right to me. But then there are the "uh-oh" benchmarks that fall short of main by anywhere from 50% to 150%! Let's look at that.

First, I don't consider the sparse set benchmarks to be entirely correct here. The ordering of entities on main is mostly sequential, which plays to the sparse set's advantage but does not reflect realistic use. The same goes for tables but to a lesser extent. (Keep in mind that I'm not too familiar with these internals; this is my best understanding.)

But for the other "uh-oh" benchmarks, what's the problem? Let's look at the "sized_commands" and "fake_commands" benchmarks. On those benchmarks, this PR looses to main by roughly 80%. **(Even though the commands do literally nothing.)** So the performance problem is actually coming from `Entities::flush`. That makes sense; on this branch, there's a *lot* more work to do with flushing, including some atomics.

That's not the only problem: 'alloc', 'free', 'reserve_entities', and 'reserve_entity' are all marginally more expensive here. That's why benchmarks like "despawn" are roughly 150% slower here (Keep in mind that a lot of that is also due to a `flush` call every despawn.) Also, 'alloc' and 'free'  have to support doing so when `flush` hasn't happened, and 'alloc',  'reserve_entities', and 'reserve_entity' need to make sure they don't step on remote reservation's toes. This extra work in combination with the excessive `flush` calls is the problem.

### Solutions

First, we can ditch a ton of unneeded `flush` calls because `free` and `alloc` no longer need it. I only dropped one on this branch, it it caused a third of the benchmarks to beat main. It follows that by getting rid of these, we could improve performance a lot here.

Second, we can create specialized `alloc` and `free` that require `flush` to have just been called. We can also add a `alloc_batch` to speed things up too.

Between those changes, I would guess we could get all the benchmarks within 5% of main (and many of them much faster). 

But those changes will take time and reorganization. For example, hooks and observers for "despawn" will need to use the world's entities and `alloc` and `free` instead of commands. That saves another flush, but it's non-trivial and could be a breaking change. So, for now, I'd like some review and feedback on the idea before working on implementing these things.

### Is it worth it?

Honestly, I don't know if remote reservations are even worth it. Making every asset `MapEntities` might even end up being faster. I'd love a clear justification for why we need this specifically. Unless of course, we can cut out so many `flush` calls that this is faster than `main` in enough places to be worth merging regardless of its impact on assets as entities and the like. I'll leave this up to SMEs to determine.


ElliottjPierce:
Closing in favor of v9: #18670

PR #18384, opened 2025-03-18
Small Docs PR for Deferred Worlds

Carter0:
# Objective

I was looking over a PR yesterday, and got confused by the docs on deferred world. I thought I would add a little more detail to the struct in order to clarify it a little. 

## Solution

Document some more about deferred worlds.

### DIFF
--- a/crates/bevy_ecs/src/world/deferred_world.rs
+++ b/crates/bevy_ecs/src/world/deferred_world.rs
@@ -20,6 +20,8 @@ use super::{unsafe_world_cell::UnsafeWorldCell, Mut, World, ON_INSERT, ON_REPLAC
 
 /// A [`World`] reference that disallows structural ECS changes.
 /// This includes initializing resources, registering components or spawning entities.
+///
+/// This means that in order to add entities, for example, you will need to use commands instead of the world directly.
 pub struct DeferredWorld<'w> {
     // SAFETY: Implementors must not use this reference to make structural changes
     world: UnsafeWorldCell<'w>,


crates/bevy_ecs/src/world/deferred_world.rs
```diff
@@ -20,6 +20,8 @@ use super::{unsafe_world_cell::UnsafeWorldCell, Mut, World, ON_INSERT, ON_REPLAC
 
 /// A [`World`] reference that disallows structural ECS changes.
 /// This includes initializing resources, registering components or spawning entities.
+///
+/// This means that in order to add entities, for example, you will need to use commands instead of the world directly.
 pub struct DeferredWorld<'w> {
```
Carter0, line 25: I would also love to add a comment about why DeferredWorld exists and why it's sometimes a bad idea to, for example, spawn an entity directly from the world. If anyone knows please respond to this comment!

alice-i-cecile: My understanding is that there are some cases where it's not sound to hand out a `&mut World`. Instead, we hand out this struct, which has the same sort of API, and then *later* apply all of the things we have planned to it.

chescock (approved):

alice-i-cecile (approved):

Carter0:
Y’all crack me up I was gonna add more haha. I’ll try to add more detail in another PR. I just wanted to get this up to get eyes on it, maybe I shoulda done something else

PR #18389, opened 2025-03-18
remove circular dependency between bevy_ecs and bevy_platform_support

mockersf:
# Objective

- https://github.com/bevyengine/bevy/pull/18350 introduced a circular dependency between bevy_ecs and bevy_platform_support which block publishing

## Solution

- Revert it


bushrat011899:
Ah I missed this when I reviewed it. May I open a PR to just remove the dependency and the example that used it, leaving the rest of the documentation intact?

PR #18393, opened 2025-03-18
Fix `Anchor` component inconsistancies

### LINKED ISSUE #18367: With Text2d anchors are stored in a component, but for sprites they are stored in a field.
## Bevy version

main

## What you did

Added an `Anchor` component to a sprite entity.

## What went wrong

The `Anchor` component is ignored. A sprite's anchor is stored in the `anchor` field on `Sprite`.

This definitely seems like a really bad API choice, particularly since there's no warning if you add the `Anchor` component to a sprite entity. `Anchor` should either be a component or not. Suggestions:
1. Remove the `anchor` field from `Sprite` and require `Anchor` instead.
2. Remove the component derive from `Anchor` and add an `anchor` field to an existing text component
3. Remove the component derive from `Anchor` and add a `TextAnchor` component newtyping `Anchor`.

I favour option 1 since there aren't really any Text2d-specific components you can add the `Anchor` field to without messing up change detection. Option 3 is fine but feels redundant.

ickshonpe:
## Objective

Fix the misleading 2d anchor API where `Anchor` is a component and required by `Text2d` but is stored on a field for sprites. 

Fixes #18367

## Solution

Remove the `anchor` field from `Sprite` and require `Anchor` instead.

## Migration Guide

The `anchor` field has been removed from `Sprite`. Instead the `Anchor` component is now a required component on `Sprite`.


kristoff3r (approved):
Makes sense to me, I tested the changed examples and they still work

alice-i-cecile (approved):

mgi388 (commented):
I thought cart generally seemed to favor fields on the “defining” component (Sprite in this case). I’m thinking back to texture atlas becoming a field on Sprite and ImageNode rather than a component. 

I don’t know what the answer is myself for this PR, and I don’t mean to unnecessarily block it, but I wanted to raise that it’s a bit unclear to me which way is the “right” way.

alice-i-cecile:
> I don’t know what the answer is myself for this PR, and I don’t mean to unnecessarily block it, but I wanted to raise that it’s a bit unclear to me which way is the “right” way.

Yeah, I don't have strong feelings here either, other than "it should be consistent". Fields are easier to discover, but separate components are better for change detection and improve cache locality in some cases by reducing struct size.

kristoff3r:
> I thought cart generally seemed to favor fields on the “defining” component (Sprite in this case). I’m thinking back to texture atlas becoming a field on Sprite and ImageNode rather than a component.
> 
> I don’t know what the answer is myself for this PR, and I don’t mean to unnecessarily block it, but I wanted to raise that it’s a bit unclear to me which way is the “right” way.

I think this is uncontroversial in the sense that `Anchor` is already a component, and having it be both a component and a field on a different component feels extremely error prone.

The alternative to this PR would be to inline the anchor logic on `Text2d` instead to mirror the `Sprite` component, but currently that component is very small with a bunch of required components. This makes it easy to spawn simple text while giving you the option to customize the text by adding these extra components when needed. Having `Text2d` contain just the string and the anchor would feel weird, but adding all of the other components to `Text2d` would be a big change for questionable gain, and might regress the ergonomics of spawning text. It would also make `Text` and `Text2d` diverge unless both are changed simultaneously.

In my opinion it's reasonable to merge this now, and then in the future if someone feels the other approach is better they can prototype it and see how it pans out.

kristoff3r:
I did notice 2 kinda weird things while looking at this:
* The `Anchor` component specifically mentions `Sprite` in its documentation, it should probably be generalized
* `Text` has a lot of the same required components as `Text2d`, but not `Anchor` (as it is a UI element that has it's own systems for this). I can see how that might be confusing, but should be solvable with good docs.

ickshonpe:
> I thought cart generally seemed to favor fields on the “defining” component (Sprite in this case). I’m thinking back to texture atlas becoming a field on Sprite and ImageNode rather than a component.
> 
> I don’t know what the answer is myself for this PR, and I don’t mean to unnecessarily block it, but I wanted to raise that it’s a bit unclear to me which way is the “right” way.

I think the way to look at it is that the fields on the defining component should contain properties that are intrinsic to the thing that the component represents. And `anchor`, unlike the rest of `Sprite`'s fields, is an extrinsic property and therefore probably doesn't belong.

mgi388:
> > I thought cart generally seemed to favor fields on the “defining” component (Sprite in this case). I’m thinking back to texture atlas becoming a field on Sprite and ImageNode rather than a component.
> > I don’t know what the answer is myself for this PR, and I don’t mean to unnecessarily block it, but I wanted to raise that it’s a bit unclear to me which way is the “right” way.
> 
> My intuition about this is that the fields on the defining component should contain properties that are intrinsic to the thing that the component represents. And `anchor`, unlike the rest of `Sprite`'s fields, is an extrinsic property and therefore probably doesn't belong.

I myself find it hard to say one way or another if it's intrinsic or extrinsic. By that logic, `flip_x` and `flip_y` also don't really seem to be intrinsic to me—the optionally flip the sprite in the same way as the anchor optionally moves the sprite on the screen. In both cases the sprite is still going to draw somewhere on the screen and look "right".

In some ways anchor _does_ feel intrinsic to me because if you, say, export sprite sheets from a program that makes sprite sheets / animations, and the sheet has been made with various anchors for each frame, then without anchor in Bevy the sprites will be drawn in the wrong place on the screen—therefore, following that the anchor is intrinsic.

Referencing the sprite sheet game files that I have, and am using for my game, each sprite sheet's frame has anchor data—the sheet was made of frames of a certain width/height, and each frame also carries explicit information about its anchor, e.g.,

```ron
(
    frames: [
        (
            width: 32,
            height: 32,
            anchor: (-0.03125, 0.46875),
        ),
        (
            width: 16,
            height: 24,
            anchor: (-0.0625, 0.45833334),
        ),
    ],
)
```

----

Just to play devil's advocate again, what if consistency for consistency's sake isn't required here, and the solution is to do nothing? I agree that adding an anchor field to `Text2d` does feel and look a bit strange. But I also feel that removing it from `Sprite` also feels a bit strange, for the reason above that _I feel maybe_ it is closer to intrinsic than extrinsic for a sprite.

That said, I'm sympathetic to the original issue though, and if that footgun is important enough to solve, then the solution proposed here is probably the right solve.

ickshonpe:
Oh yeah I forgot about completely about sprite offsets like that, in that situation the anchor is intrinsic. I've been working on texture slicing and text2d lately where anchors are just used for alignment. Another option would be to store the sprite anchor points in image/texture atlas asset metadata instead. That seems like it might be a better solution than any of the others to me, but it's a lot more effort to implement.

NthTensor:
Re-added migration guide tag so you get the new instructions. You should move your migration guide from the pr description to a new file.

NthTensor (approved):
Migration guide looks good. Sorry to block you over something so small.

PR #18395, opened 2025-03-18
Fix dynamic scene resources not being entity mapped

ElliottjPierce:
# Objective

The resources were converted via `clone_reflect_value` and the cloned value was mapped. But the value that is inserted is the source of the clone, which was not mapped.

I ran into this issue while working on #18380. Having non consecutive entity allocations has caught a lot of bugs.

## Solution

Use the cloned value for insertion if it exists.

### DIFF
--- a/crates/bevy_scene/src/dynamic_scene.rs
+++ b/crates/bevy_scene/src/dynamic_scene.rs
@@ -149,16 +149,22 @@ impl DynamicScene {
 
             // If this component references entities in the scene, update
             // them to the entities in the world.
-            if let Some(map_entities) = registration.data::<ReflectMapEntities>() {
-                let mut resource = clone_reflect_value(resource.as_partial_reflect(), registration);
+            let mut cloned_resource;
+            let partial_reflect_resource = if let Some(map_entities) =
+                registration.data::<ReflectMapEntities>()
+            {
+                cloned_resource = clone_reflect_value(resource.as_partial_reflect(), registration);
                 SceneEntityMapper::world_scope(entity_map, world, |_, mapper| {
-                    map_entities.map_entities(resource.as_partial_reflect_mut(), mapper);
+                    map_entities.map_entities(cloned_resource.as_partial_reflect_mut(), mapper);
                 });
-            }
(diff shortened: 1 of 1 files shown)


crates/bevy_scene/src/dynamic_scene.rs
```diff
@@ -149,16 +149,22 @@ impl DynamicScene {
 
             // If this component references entities in the scene, update
             // them to the entities in the world.
-            if let Some(map_entities) = registration.data::<ReflectMapEntities>() {
-                let mut resource = clone_reflect_value(resource.as_partial_reflect(), registration);
+            let mut cloned_resource;
+            let partial_reflect_resource = if let Some(map_entities) =
+                registration.data::<ReflectMapEntities>()
+            {
+                cloned_resource = clone_reflect_value(resource.as_partial_reflect(), registration);
```
kristoff3r, line 156:
```suggestion
            let partial_reflect_resource = if let Some(map_entities) =
                registration.data::<ReflectMapEntities>()
            {
                let mut cloned_resource = clone_reflect_value(resource.as_partial_reflect(), registration);
```

kristoff3r (approved):
Makes sense to me, but it would be nice to have a test for it

crates/bevy_scene/src/dynamic_scene.rs
ElliottjPierce: Unfortunately, `cloned_resource` needs to live outside the if scope. Otherwise it gets dropped at the end of the if, leaving the `partial_reflect_resource` with a dangling pointer. I put it outside to defeat the borrow checker lol.

kristoff3r: ah lol, fair enough

ElliottjPierce:
> Makes sense to me, but it would be nice to have a test for it

Absolutely. #18380 basically makes all entity allocation/reservation order deterministic but unreliable. So it kinda makes a test for all this stuff. (I found the bug because test cases were failing).

But if #18380 or an equivalent is not merged, I'll revisit this.

alice-i-cecile (approved):

PR #18404, opened 2025-03-19
Clean up unused dependencies

IceSentry:
# Objective

- A lot of crates define dependencies that aren't used. These dependencies are generally already in the tree so that's why CI isn't complaining and it will likely not result in better compile times but I think it's worth it to keep this in check

## Solution

- Run cargo machete and only remove dependencies I'm sure can be removed

## Testing

- Ran a few examples locally on windows and ran the ci tool locally

## Notes

In a local test, I did end up with 1 less dependencies being compiled. It went from 455 to 454 according to cargo's output. It won't make an impact when compiling the entirety of bevy but it might help users depending on sub crates.

### DIFF
--- a/crates/bevy_core_pipeline/Cargo.toml
+++ b/crates/bevy_core_pipeline/Cargo.toml
@@ -47,7 +47,6 @@ nonmax = "0.5"
 smallvec = "1"
 thiserror = { version = "2", default-features = false }
 tracing = { version = "0.1", default-features = false, features = ["std"] }
-bytemuck = { version = "1" }
 
 [lints]
 workspace = true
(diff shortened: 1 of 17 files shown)


JMS55 (approved):
I'm a little hesitant to do this so close to 0.16, but hopefully the RC will catch any regressions if there are any.

IceSentry:
It doesn't have to be done now, but some of those unused dependencies are unused because of changes in this dev cycle so I feel it makes sense to release something with a more accurate dep tree.

crates/bevy_core_pipeline/Cargo.toml
```diff
@@ -47,7 +47,6 @@ nonmax = "0.5"
 smallvec = "1"
 thiserror = { version = "2", default-features = false }
 tracing = { version = "0.1", default-features = false, features = ["std"] }
-bytemuck = { version = "1" }
```
mockersf, line 50: byte muck used to be needed to be in scope but unused for some macro. is that not the case anymore?

PR #18405, opened 2025-03-19
enable std when building bevy_dylib

mockersf:
# Objective

- bevy_dylib fails to build:
```
   Compiling bevy_dylib v0.16.0-rc.1 (/bevy/crates/bevy_dylib)
error: linking with `cc` failed: exit status: 1
  |
  = note: some arguments are omitted. use `--verbose` to show all linker arguments
  = note: Undefined symbols for architecture arm64:
            "__critical_section_1_0_acquire", referenced from:
                critical_section::with::h00cfbe529dea9dc9 in libbevy_tasks-53c9db6a3865f250.rlib[58](bevy_tasks-53c9db6a3865f250.evom2xwveqp508omiiqb25xig.rcgu.o)
            "__critical_section_1_0_release", referenced from:
                core::ptr::drop_in_place$LT$critical_section..with..Guard$GT$::hfa034e0208e1a49d in libbevy_tasks-53c9db6a3865f250.rlib[48](bevy_tasks-53c9db6a3865f250.d9dwgpd0156zfn2h5z5ff94zn.rcgu.o)
          ld: symbol(s) not found for architecture arm64
          clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

## Solution

- enable `std` when building bevy_dylib

### DIFF
--- a/crates/bevy_dylib/Cargo.toml
+++ b/crates/bevy_dylib/Cargo.toml
@@ -13,7 +13,7 @@ crate-type = ["dylib"]
 
 [dependencies]
 bevy_internal = { path = "../bevy_internal", version = "0.16.0-dev", default-features = false, features = [
-    "std",
+  "std",
 ] }
 
 [lints]
--- a/crates/bevy_dylib/Cargo.toml
+++ b/crates/bevy_dylib/Cargo.toml
@@ -12,6 +12,8 @@ keywords = ["bevy"]
 crate-type = ["dylib"]
 
 [dependencies]
+# feature std is needed to avoid an issue when linking critical_section
+# bevy_dylib is not expected to work in no_std
 bevy_internal = { path = "../bevy_internal", version = "0.16.0-dev", default-features = false, features = [
   "std",
 ] }
--- a/crates/bevy_dylib/Cargo.toml
+++ b/crates/bevy_dylib/Cargo.toml
@@ -12,7 +12,9 @@ keywords = ["bevy"]
 crate-type = ["dylib"]
 
 [dependencies]
-bevy_internal = { path = "../bevy_internal", version = "0.16.0-dev", default-features = false }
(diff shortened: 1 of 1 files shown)


bushrat011899 (approved):
`bevy_dylib` wasn't supposed to have `no_std` support, this was an accident on my part once I added the `std` feature to `bevy_internal`. Thanks for fixing it!

alice-i-cecile (approved):

crates/bevy_dylib/Cargo.toml
```diff
@@ -12,7 +12,9 @@ keywords = ["bevy"]
 crate-type = ["dylib"]
 
 [dependencies]
-bevy_internal = { path = "../bevy_internal", version = "0.16.0-dev", default-features = false }
+bevy_internal = { path = "../bevy_internal", version = "0.16.0-dev", default-features = false, features = [
+    "std",
```
cart, line 16: It is worth documenting why we're requiring `std` here. I doubt future people will be able to intuit that this fixes an obscure linking bug in a specific crate.

cart (approved):

bushrat011899:
Just for posterity, the reason `critical-section` failed to link is "deliberate". `critical-section` requires exactly one implementation to be provided in order for it function (see [_Undefined reference errors_](https://github.com/rust-embedded/critical-section?tab=readme-ov-file#undefined-reference-errors)). To create a dynamic library requires linking in that implementation, which is supposed to be provided by the specific platform you're developing on.

`critical-section` provides a built-in implementation when the `std` feature is enabled, so we just need to have that feature enabled in order for `bevy_dylib` to have all symbols resolved at link-time. That does mean `bevy_dylib` is currently not `no_std`, but that was entirely expected (it doesn't even have `#![no_std]` so it was "never" compatible without the standard library). Perhaps in the future that could change, but I believe the overlap in users who need `no_std` support and users who also need dynamic linking is _very_ small.

PR #11823, opened 2024-02-11
Immediately apply deferred system params in System::run

james7132:
# Objective
Fixes #11821. 

## Solution
* Run `System::apply_deferred` in `System::run` after executing the system.
* Switch to using `System::run_unsafe` in `SingleThreadedExecutor` to preserve the current behavior.
* Remove the `System::apply_deferred` in `SimpleExecutor` as it's now redundant.
* Remove the `System::apply_deferred` when running one-shot systems, as it's now redundant. 

---

## Changelog
Changed: `System::run` will now immediately apply deferred system params after running the system.

## Migration Guide
`System::run` will now always run `System::apply_deferred` immediately after running the system now. If you were running systems and then applying their deferred buffers at a later point in time, you can eliminate the latter.

```rust
// in 0.13
system.run(world);
// .. sometime later ...
system.apply_deferred(world);

// in 0.14
system.run(world);
```

PR #6698, opened 2022-11-20
[Merged by Bors] - Fix PipeSystem panicking with exclusive systems

inodentry:
Without this fix, piped systems containing exclusive systems fail to run, giving a runtime panic.
With this PR, running piped systems that contain exclusive systems now works.

## Explanation of the bug

This is because, unless overridden, the default implementation of `run` from the `System` trait simply calls `run_unsafe`. That is not valid for exclusive systems. They must always be called via `run`, as `run_unsafe` takes `&World` instead of `&mut World`.

Trivial reproduction example:
```rust
fn main() {
    App::new()
        .add_plugins(DefaultPlugins)
        .add_system(exclusive.pipe(another))
        .run();
}

fn exclusive(_world: &mut World) {}
fn another() {}
```
If you run this, you will get a panic 'Cannot run exclusive systems with a shared World reference' and the backtrace shows how bevy (correctly) tries to call the `run` method (because the system is exclusive), but it is the implementation from the `System` trait (because `PipeSystem` does not have its own), which calls `run_unsafe` (incorrect):
 - 3: <bevy_ecs::system::system_piping::PipeSystem<SystemA,SystemB> as bevy_ecs::system::system::System>::run_unsafe
 - 4: bevy_ecs::system::system::System::run

PR #6083, opened 2022-09-24
[Merged by Bors] - Exclusive Systems Now Implement `System`. Flexible Exclusive System Params

cart:
# Objective

The [Stageless RFC](https://github.com/bevyengine/rfcs/pull/45) involves allowing exclusive systems to be referenced and ordered relative to parallel systems. We've agreed that unifying systems under `System` is the right move.

This is an alternative to #4166 (see rationale in the comments I left there). Note that this builds on the learnings established there (and borrows some patterns).

## Solution

This unifies parallel and exclusive systems under the shared `System` trait, removing the old `ExclusiveSystem` trait / impls. This is accomplished by adding a new `ExclusiveFunctionSystem` impl similar to `FunctionSystem`. It is backed by `ExclusiveSystemParam`, which is similar to `SystemParam`. There is a new flattened out SystemContainer api (which cuts out a lot of trait and type complexity). 

This means you can remove all cases of `exclusive_system()`:

```rust
// before
commands.add_system(some_system.exclusive_system());
// after
commands.add_system(some_system);
```

I've also implemented `ExclusiveSystemParam` for `&mut QueryState` and `&mut SystemState`, which makes this possible in exclusive systems:

```rust
fn some_exclusive_system(
    world: &mut World,
    transforms: &mut QueryState<&Transform>,
    state: &mut SystemState<(Res<Time>, Query<&Player>)>,
) {
    for transform in transforms.iter(world) {
        println!("{transform:?}");
    }
    let (time, players) = state.get(world);
    for player in players.iter() {
        println!("{player:?}");
    }
}
```

Note that "exclusive function systems" assume `&mut World` is present (and the first param). I think this is a fair assumption, given that the presence of `&mut World` is what defines the need for an exclusive system.

I added some targeted SystemParam `static` constraints, which removed the need for this:
``` rust
fn some_exclusive_system(state: &mut SystemState<(Res<'static, Time>, Query<&'static Player>)>) {}
```

## Related

- #2923
- #3001
- #3946

## Changelog

- `ExclusiveSystem` trait (and implementations) has been removed in favor of sharing the `System` trait.
- `ExclusiveFunctionSystem` and `ExclusiveSystemParam` were added, enabling flexible exclusive function systems
- `&mut SystemState` and `&mut QueryState` now implement `ExclusiveSystemParam`
- Exclusive and parallel System configuration is now done via a unified `SystemDescriptor`, `IntoSystemDescriptor`, and `SystemContainer` api.

## Migration Guide

[...]

PR #18406, opened 2025-03-19
Fix `run_system` for adapter systems wrapping exclusive systems

chescock:
# Objective

Fix panic in `run_system` when running an exclusive system wrapped in a `PipeSystem` or `AdapterSystem`.  

#18076 introduced a `System::run_without_applying_deferred` method.  It normally calls `System::run_unsafe`, but `ExclusiveFunctionSystem::run_unsafe` panics, so it was overridden for that type.  Unfortunately, `PipeSystem::run_without_applying_deferred` still calls `PipeSystem::run_unsafe`, which can then call `ExclusiveFunctionSystem::run_unsafe` and panic.  

## Solution

Make `ExclusiveFunctionSystem::run_unsafe` work instead of panicking.  Clarify the safety requirements that make this sound.  

The alternative is to override `run_without_applying_deferred` in `PipeSystem`, `CombinatorSystem`, `AdapterSystem`, `InfallibleSystemWrapper`, and `InfallibleObserverWrapper`.  That seems like a lot of extra code just to preserve a confusing special case!  

Remove some implementations of `System::run` that are no longer necessary with this change.  This slightly changes the behavior of `PipeSystem` and `CombinatorSystem`: Currently `run` will call `apply_deferred` on the first system before running the second, but after this change it will only call it after *both* systems have run.  The new behavior is consistent with `run_unsafe` and `run_without_applying_deferred`, and restores the behavior prior to #11823.

The panic was originally necessary because [`run_unsafe` took `&World`](https://github.com/bevyengine/bevy/pull/6083/files#diff-708dfc60ec5eef432b20a6f471357a7ea9bfb254dc2f918d5ed4a66deb0e85baR90).  Now that it takes `UnsafeWorldCell`, it is possible to make it work.  See also Cart's concerns at https://github.com/bevyengine/bevy/pull/4166#discussion_r979140356, although those also predate `UnsafeWorldCell`.  

And see #6698 for a previous bug caused by this panic.

### DIFF
--- a/crates/bevy_ecs/src/schedule/executor/multi_threaded.rs
+++ b/crates/bevy_ecs/src/schedule/executor/multi_threaded.rs
@@ -452,6 +452,8 @@ impl ExecutorState {
 
                 // SAFETY:
                 // - Caller ensured no other reference to this system exists.
+                // - `system_task_metadata[system_index].is_exclusive` is `false`,
+                //   so `System::is_exclusive` returned `false` when we called it.
                 // - `can_run` has been called, which calls `update_archetype_component_access` with this system.
                 // - `can_run` returned true, so no systems with conflicting world access are running.
                 unsafe {
@@ -592,6 +594,7 @@ impl ExecutorState {
 
     /// # Safety
     /// - Caller must not alias systems that are running.
+    /// - `is_exclusive` must have returned `false` for the specified system.
     /// - `world` must have permission to access the world data
     ///   used by the specified system.
     /// - `update_archetype_component_access` must have been called with `world`
--- a/crates/bevy_ecs/src/schedule/executor/multi_threaded.rs
+++ b/crates/bevy_ecs/src/schedule/executor/multi_threaded.rs
@@ -609,6 +609,7 @@ impl ExecutorState {
                 // SAFETY:
                 // - The caller ensures that we have permission to
                 // access the world data used by the system.
+                // - `is_exclusive` returned false
                 // - `update_archetype_component_access` has been called.
                 unsafe {
                     if let Err(err) = __rust_begin_short_backtrace::run_unsafe(
--- a/crates/bevy_ecs/src/system/exclusive_function_system.rs
+++ b/crates/bevy_ecs/src/system/exclusive_function_system.rs
@@ -114,8 +114,7 @@ where
         input: SystemIn<'_, Self>,
         world: UnsafeWorldCell,
     ) -> Self::Out {
-        // SAFETY: `Self::is_exclusive` returns true, so the caller is
-        // required to ensure that the it's valid to call `world_mut()`
+        // SAFETY: The safety is upheld by the caller.
         let world = unsafe { world.world_mut() };
         world.last_change_tick_scope(self.system_meta.last_run, |world| {
             #[cfg(feature = "trace")]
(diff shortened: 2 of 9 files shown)


alice-i-cecile (approved):

crates/bevy_ecs/src/system/exclusive_function_system.rs
```diff
@@ -110,18 +110,12 @@ where
...
-        panic!("Cannot run exclusive systems with a shared World reference");
-    }
-
-    fn run_without_applying_deferred(
         &mut self,
         input: SystemIn<'_, Self>,
-        world: &mut World,
+        world: UnsafeWorldCell,
     ) -> Self::Out {
+        // SAFETY: `is_exclusive` returned true, so the caller ensured we had `&mut World` access
```
hymm, line 117: This safety comment doesn't make sense to me. We never check is_exclusive in this method. My understanding here is that you've passed the safety requirements for calling `world_mut` to the caller as it can't be verified here. https://docs.rs/bevy/latest/bevy/ecs/world/unsafe_world_cell/struct.UnsafeWorldCell.html#safety-1

And so need to update the safety comment on `run_unsafe`.

chescock: Right, I did update the safety comment on `run_unsafe`.  It now has [the additional requirement](https://github.com/bevyengine/bevy/pull/18406/files/e9dc5f56843472171c867754e4c8512fe950d0a2#diff-75be02e2bfe4197aaed28e4d0e5dd3e3bfe3c3f464817ea896bae6b3b128425cR72-R73) that 

```rust
    /// - If [`System::is_exclusive`] returns `true`, then it must be valid to call
    ///   [`UnsafeWorldCell::world_mut`] on `world`.
```

What I wanted to communicate here is that because `ExclusiveFunctionSystem::is_exclusive` returns `true`, that clause triggered and the caller had to ensure that `world_mut` was valid.  

If you have ideas on how to make the wording more clear, I'm happy to update it!

chescock:
Adding to 0.16 because I'm pretty sure this worked in 0.15 and will be a regression if not fixed.

crates/bevy_ecs/src/schedule/executor/multi_threaded.rs
```diff
@@ -609,6 +609,7 @@ impl ExecutorState {
                 // SAFETY:
                 // - The caller ensures that we have permission to
                 // access the world data used by the system.
+                // - `is_exclusive` returned false
```
hymm, line 612: is exclusive isn't checked in this function. You need to add another safety bullet to `spawn_system_task` and add a bullet the unsafe in `spawn_system_tasks`

crates/bevy_ecs/src/system/exclusive_function_system.rs
```diff
@@ -110,18 +110,13 @@ where
...
-        panic!("Cannot run exclusive systems with a shared World reference");
-    }
-
-    fn run_without_applying_deferred(
         &mut self,
         input: SystemIn<'_, Self>,
-        world: &mut World,
+        world: UnsafeWorldCell,
     ) -> Self::Out {
+        // SAFETY: `Self::is_exclusive` returns true, so the caller is
```
hymm, line 117: I would just change this to `The safety is upheld by the caller.`

hymm (approved):
Just some nits, but the safety comments in multi_threaded should probably be fixed.

alice-i-cecile:
Ping me when the safety comments are improved and I'll merge this in.

chescock:
It should be ready now!  

(Sorry, I had pushed up the fixes before your message, but I always forget that I'm allowed to click "resolve conversation" myself.)

PR #18408, opened 2025-03-19
Allow entity wrappers in entity set and map types

Victoronz:
# Objective

Related to  #16547.

Currently we wrap `HashSet`/`HashMap`/`IndexSet`/`IndexMap` over the `Entity` + `EntityHash` combination. This allows them to be used with `EntitySet` functionality.
However, when `Entity` is newtyped, these wrappers can suddenly no longer be used!
This causes types like `MainEntityHashSet`/`MainEntityHashMap` to fall back to the `hashbrown/indexmap` versions again.

## Solution

We allow keying our map and set wrappers with `K: TrustedEntityBorrow` types, not just `Entity`.

This requires a new trait to be sound: `TrustedBuildHasher<H: Hash>: BuildHasher`.
Because both a `BuildHasher` and the hashed `Hash` type could introduce nondeterminism into the final hash, we have to guarantee that such combinations produce valid results. Collisions are fine, set and map types can deal with them, only nondeterministic hashes present an issue.

On account of our wrappers only using `EntityHash`, we need only bound them on `EntityHash: TrustedBuilderHasher<T>`. Not all `TrustedEntityBorrow` types make sense to be hashed with `EntityHash` however, so the only `EntityHash: TrustedBuildHasher` implementations for now are with `Entity`, entity newtypes, and the `EntityRef`-family.

Adding a generic in place of `Entity` presents an issue for `HashMap` and `IndexMap`:
If we want to retain the main case of `HashMap<Entity, V>`, one would think to make `Entity` the default for K.
But if we were to do that, we'd need to swap the positions for the `K` and `V` generics, as defaulted type parameters need to trail.

Because map types with swapped `K` and `V` would be very misleading, we abstain from generics, and use a type alias for the `EntityHashMap`/`EntityIndexMap` case instead. 
A type alias and the aliased type need to have separate names, we name these new, broader versions `EntityEquivalent*` instead of just `Entity*`.
To keep the names consistent, we do the same for the set versions.

The name `EntityEquivalent` is more accurate to the actual meaning of the `TrustedEntityBorrow` trait, so it may be worth renaming it in another PR alongside the currently not so used `EntityBorrow`. (`EntityBorrow` could have a separate name, so we do not need a `Trusted` prefix for `EntityEquivalent`)

While the additions/deletions seem quite ominous, what this PR does is conceptually not that complex:
In `hash.rs`
 - Add `TrustedBuildHasher` trait and its impls for `Entity`/`EntityRef`-family
 In `hash_set.rs`/`hash_map.rs`/`index_set.rs`/`index_map.rs`:
 - Rename set/map type from `Entity*` to `EntityEquivalent*`
 - Make aliases with the old `Entity*` names
 - Replace `Entity` key with `K` generic
 - Bound `K` on `TrustedEntityBorrow + Hash` plus whatever specific impls require
 - Add `where EntityHash: TrustedBuildHasher<K>` bounds to type definitions and impl blocks
 In `sync_world.rs`:
 - Add `TrustedBuildHasher` impls with `MainEntity`/`RenderEntity`
 - Make `MainEntityHashSet`/`Map` aliases of `EntityEquivalentHashSet`/`Map` respectively.
 The rest:
 - Fix up some `HashMap::default` to `MainEntityHashMap::default`
 - Replace map/set imports
 - Adjust a mention of `hash_set::Iter`
 
It is painful to have wrapper types that take up this much space and should really not be necessary, but on the other hand, it clarifies/increases our justification to request the required changes upstream (`hashbrown`, then `indexmap`)!

With future upstream changes, we'd ultimately only need to retain the `TrustedBuildHasher` trait, some aliases and the added bounds in some impls.
If we'd then like to keep the wrappers around, we still could as well, but they could be made a lot lighter.

## Migration Guide

`MainEntityHashSet`/`MainEntityHashMap` are no longer aliases of the `hashbrown` types, so some associated functions have to use the proper names or aliases to be used in place of `HashMap`.
Example: `HashMap::default` -> `MainEntityHashMap::default`

Types associated with any of `EntityHashSet`, `EntityHashMap`, `EntityIndexSet`, `EntityIndexMap` now have an additional `K` generic. To maintain the previous meaning, use `Entity` for `K`.


chescock:
> This requires a new trait to be sound: `TrustedBuildHasher<H: Hash>: BuildHasher`. Because both a `BuildHasher` and the hashed `Hash` type could introduce nondeterminism into the final hash, we have to guarantee that such combinations produce valid results. Collisions are fine, set and map types can deal with them, only nondeterministic hashes present an issue.
> 
> On account of our wrappers only using `EntityHash`, we need only bound them on `EntityHash: TrustedBuilderHasher<T>`. Not all `TrustedEntityBorrow` types make sense to be hashed with `EntityHash` however, so the only `EntityHash: TrustedBuildHasher` implementations for now are with `Entity`, entity newtypes, and the `EntityRef`-family.

I don't understand the attack here.  `TrustedEntityBorrow` requires that `Hash` "must be equivalent for `Self` and its underlying entity".  Since hashing `Entity` with `EntityHash` is deterministic, wouldn't hashing any valid `TrustedEntityBorrow` with `EntityHash` *also* be deterministic?  If we didn't have `TrustedBuildHasher`, how could a malicious user cause nondeterminism without violating the existing safety contract of `TrustedEntityBorrow`?

Victoronz:
> > This requires a new trait to be sound: `TrustedBuildHasher<H: Hash>: BuildHasher`. Because both a `BuildHasher` and the hashed `Hash` type could introduce nondeterminism into the final hash, we have to guarantee that such combinations produce valid results. Collisions are fine, set and map types can deal with them, only nondeterministic hashes present an issue.
> > On account of our wrappers only using `EntityHash`, we need only bound them on `EntityHash: TrustedBuilderHasher<T>`. Not all `TrustedEntityBorrow` types make sense to be hashed with `EntityHash` however, so the only `EntityHash: TrustedBuildHasher` implementations for now are with `Entity`, entity newtypes, and the `EntityRef`-family.
> 
> I don't understand the attack here. `TrustedEntityBorrow` requires that `Hash` "must be equivalent for `Self` and its underlying entity". Since hashing `Entity` with `EntityHash` is deterministic, wouldn't hashing any valid `TrustedEntityBorrow` with `EntityHash` _also_ be deterministic? If we didn't have `TrustedBuildHasher`, how could a malicious user cause nondeterminism without violating the existing safety contract of `TrustedEntityBorrow`?

You're right, I didn't use the correct language here. For the current implementation, because `Entity` hashes deterministically, all `TrustedEntityBorrow` types must too. And because `EntityHash` does not introduce nondeterminism, the implementation is sound.

To be more accurate with my intent: "To ensure this implementation remains sound under future extensions/refactors, we introduce a new trait...".

It is feasible to want to extend the implementation to more hashers, whether that is to be able to use `TrustedEntityBorrow` types that are not `Entity` values themselves (any pointer to `Entity` f.e.), or if we want to support with focuses other than performance, like cryptographic hashers.

Even aside from the soundness, we want to have a trait that limits `Hasher/Hash` combinations like this, because we only want to use `TrustedEntityBorrow` types that `EntityHash` is optimized to hash for.

chescock:
> To ensure this implementation remains sound under future extensions/refactors, we introduce a new trait...

Do you have concrete plans for those extensions?  Otherwise it feels like premature generalization.  If we don't need it yet, then there's no reason to build it yet!  

Either way, I still don't see why `TrustedBuildHasher` needs to be parameterized by `H: Hash`.  Any realistic `BuildHasher`s is deterministic for *any* sequence of bytes.  What sort of `BuildHasher` would be deterministic for some (deterministic) `H: Hash` but not another?  

> Even aside from the soundness, we want to have a trait that limits `BuildHasher/Hash` combinations like this, because we only want to use `TrustedEntityBorrow` types that `EntityHash` is optimized to hash for.

Why would we want to limit the combinations?  As I said above, `TrustedEntityBorrow` requires that `Hash` "must be equivalent for `Self` and its underlying entity".  How can there be a valid `TrustedEntityBorrow` that doesn't work with `EntityHash`?

Victoronz:
> > To ensure this implementation remains sound under future extensions/refactors, we introduce a new trait...
> 
> Do you have concrete plans for those extensions? Otherwise it feels like premature generalization. If we don't need it yet, then there's no reason to build it yet!

That might be the case, I did conflate the purposes of "must be a sound `BuildHasher`/`Hash` combination" with "should be a designed-for `BuildHasher`/`Hash` combination".

> Either way, I still don't see why `TrustedBuildHasher` needs to be parameterized by `H: Hash`. Any realistic `BuildHasher`s is deterministic for _any_ sequence of bytes. What sort of `BuildHasher` would be deterministic for some (deterministic) `H: Hash` but not another?

The problem is that both the `Hasher` and `Hash` are in control of parts of the hashing process. Whether a `BuildHasher` is deterministic for any sequence of bytes does not matter if the same bytes for the same value aren't being fed in the first place.
A `Hash` impl could even `mem::swap` a `Hasher` out entirely if it wanted to.

> > Even aside from the soundness, we want to have a trait that limits `BuildHasher/Hash` combinations like this, because we only want to use `TrustedEntityBorrow` types that `EntityHash` is optimized to hash for.
> 
> Why would we want to limit the combinations? As I said above, `TrustedEntityBorrow` requires that `Hash` "must be equivalent for `Self` and its underlying entity". How can there be a valid `TrustedEntityBorrow` that doesn't work with `EntityHash`?

`TrustedEntityBorrow` does not limit `Hash` impls to be **equal** to the `Entity` impl, but to form the same total order as `Entity`!
The very purpose of `EntityHash` is to be performance optimized, and if we begin hashing pointers instead of literal `Entity`, we defeat the purpose.
This is one of the reasons why having more `BuildHasher`s might make sense, so other `TrustedEntityBorrow` types could use their own dedicated hashers.

It seems we might want to expand on the `TrustedEntityBorrow` safety comment somewhat.
The steps toward concluding that `EntityHash` is sound with any `TrustedEntityBorrow` type are quite nebulous with the current documentation.

chescock:
> The intent is not for `TrustedEntityBorrow` to limit `Hash` impls to be **equal** to the `Entity` impl, but for the hashes to form the same total order as `Entity`!

Ah, this must be the miscommunication!  I've been assuming it works the same was as `Borrow`, where they have to be equal so that you can get the hash from a `&str` to key a `HashMap<String, T>`.  

I'd be in favor of changing that, if we can.  I think it would just be a doc change, since all the existing impls do follow that.  And those of us who don't read docs very carefully are likely to assume that `TrustedEntityBorrow` works the same as `Borrow<Entity>`, no matter what we write :).  

> The problem is that both the `Hasher` and `Hash` are in control of parts of the hashing process. Whether a `BuildHasher` is deterministic for any sequence of bytes does not matter if the same bytes for the same value aren't being fed in the first place.

Yup, I agree we need an `unsafe trait` for both parts!  I just think we can make the requirements independent, so that we don't need to verify each combination.  (And then, since this PR only uses a single `BuildHasher` type, we don't need `TrustedBuildHasher` just yet.)

Victoronz:
> > The intent is not for `TrustedEntityBorrow` to limit `Hash` impls to be **equal** to the `Entity` impl, but for the hashes to form the same total order as `Entity`!
> 
> Ah, this must be the miscommunication! I've been assuming it works the same was as `Borrow`, where they have to be equal so that you can get the hash from a `&str` to key a `HashMap<String, T>`.
> 
> I'd be in favor of changing that, if we can. I think it would just be a doc change, since all the existing impls do follow that. And those of us who don't read docs very carefully are likely to assume that `TrustedEntityBorrow` works the same as `Borrow<Entity>`, no matter what we write :).

I made a rename/doc PR concerning this (and other doc/name issues) in #18470!

Victoronz:
> > The problem is that both the `Hasher` and `Hash` are in control of parts of the hashing process. Whether a `BuildHasher` is deterministic for any sequence of bytes does not matter if the same bytes for the same value aren't being fed in the first place.
> 
> Yup, I agree we need an `unsafe trait` for both parts! I just think we can make the requirements independent, so that we don't need to verify each combination. (And then, since this PR only uses a single `BuildHasher` type, we don't need `TrustedBuildHasher` just yet.)

If we split the trait, we'd end up with this structure:
- unsafe trait `DeterministicBuildHasher`
- unsafe trait `DeterministicHash`
- trait `CorrectBuildHasher<H: DeterministicHash>: DeterministicBuildHasher`

The `BuildHasher`/`Hasher`/`Hash` traits are quite a strange bit of design. While the `Hasher` API can handle just plain bytes, real `BuildHasher`/`Hash` combinations are not built to work in all cases. Some `BuildHasher`s will panic when encountering any but their designed-for type, and some `Hash` impls will do their own thing, and still be deterministic.
Either side can have its own notion of "this is what hashing is for", which can cause pathological logic errors when mixed.

The only way to prevent this is to have a layer that accounts for the full picture, which means an impl on one side that carries a generic of the other. (Instead of requiring that a `Hash` type be aware of various `BuildHasher`s that can be used to hash it, it makes more sense to require a `BuildHasher` to determine what types it should be used to hash for. A `BuildHasher` usually does not interact with non-hashing API, reducing the pain of newtyping by a lot.)

In the case of this PR, you are right that we can punt on is `DeterministicBuildHasher` because we currently only use `EntityHash`, and `DeterministicHash` because `TrustedEntityBorrow` already ensures determinism. But that doesn't enable what you thought it would!

We still need to limit inputs to `EntityHasher` to only u64 fields, otherwise there will be no just be hashes, but panics as well! 
This is an aspect of the `Hash` impl that `TrustedEntityBorrow` does not, and should not promise.

Meaning that while we can punt on the safety parts, the combination checking and consequent bounds on our set/map types will need to remain, just under a different name. :P

Victoronz:
> In the case of this PR, you are right that we can punt on is `DeterministicBuildHasher` because we currently only use `EntityHash`, and `DeterministicHash` because `TrustedEntityBorrow` already ensures determinism. But that doesn't enable what you thought it would!

I said here `TrustedEntityBorrow` ensures determinism, but I just realized that it does not do that, because the current wording doesn't take into account that the `Hash` impl alone cannot guarantee correct behavior, ugh.

While we might not need the traits themselves, we should still describe this situation in the safety comment of `TrustedEntityBorrow`, because we are otherwise requiring a promise that an implementer cannot ensure on their own.

chescock:
> We still need to limit inputs to `EntityHasher` to only u64 fields, otherwise there will be no just be hashes, but panics as well! This is an aspect of the `Hash` impl that `TrustedEntityBorrow` does not, and should not promise.

Why do we need to limit the combinations?  `std::collections::HashMap` doesn't.  

The only *safety* requirement we have is that the hashing be deterministic.  It's fine for `hash()` to panic, as long as it panics deterministically, since there's no way to get duplicates of a value that fails to hash.  

> I said here `TrustedEntityBorrow` ensures determinism, but I just realized that it does not do that, because the current wording doesn't take into account that the `Hash` impl alone cannot guarantee correct behavior, ugh.

Sure, it does.  A deterministic `Hash` is one that makes a deterministic sequence of calls to the provided `Hasher`.  The fact that the `Hasher` can do non-deterministic things doesn't change the fact that the `Hash` was deterministic.  

`Entity` has a deterministic `Hash`, and `TrustedEntityBorrow` is required to have the same `Hash` behavior as `Entity`, so `TrustedEntityBorrow` is required to have a deterministic `Hash`.

Victoronz:
This is a really subtle discussion! It has design considerations on multiple levels, which I find/can make it difficult to hone in on where the different aspects come in/from, I hope the back and forth isn't getting frustrating!
That being said, I'll try and be more thorough:

> > We still need to limit inputs to `EntityHasher` to only u64 fields, otherwise there will be no just be hashes, but panics as well! This is an aspect of the `Hash` impl that `TrustedEntityBorrow` does not, and should not promise.
> 
> Why do we need to limit the combinations? `std::collections::HashMap` doesn't.

Safety-wise, panicking is completely fine!
This part is more about "what should a EntityEquivalentHashMap" be, that is the issue. To be a useful and compatible design, I think we should restrict it to use cases that make sense. Why does `std` not limit combinations? Because their `DefaultHasher` is intentionally as compatible as feasible!

First, let me paint a situation from a standpoint of compatibility: 
We add an `EntityEquivalentHashMap` that has no restriction on `T: TrustedEntityBorrow`.
Some library/code comes along, and makes use of `EntityEquivalentHashSet<&Entity>`.
This code compiles perfectly fine! However, once somebody on a 32-bit platform comes along, their bevy program suddenly explodes.
That is because on 64-bit platforms, where pointers are `u64`s, `EntityHash` will not panic. On 32-bit platforms, where pointers are `u32`s, `EntityHash` is suddenly being fed something other than `u64` and panics.
Note that this can occur with any type that has platform/feature-dependent fields (which most pointer types do).
If we allow all combinations, we will still have to list the types that should not be used, and have users manage this concern manually.

Let's have a look from a performance standpoint:
`EntityHash` is designed to hash `Entity`. Designing hash functions is complicated, and the more they can assume about their input, the better. As consequence, if we then pass data that no longer follows those assumptions, we might run into some pathologically bad performance! An example is the recently created [`foldhash`](https://github.com/orlp/foldhash). 
I remember some people running into issues with a large performance regression when `foldhash` was introduced as the default. IIRC, this turned out to be because they were hashing pointers, and in their case, those pointers assumed a distribution that the hash function could not hash perfomantly for, i.e. There were a lot of collisions, regressing performance to a large degree.

The problems that are created by incompatible `Hasher`/`Hash` combinations are difficult to identify, and are made up of a lot of not-easily-found details. How is a user/library author supposed to know/predict when a combination fails? 
In the first situation I described, I talked about how a `u64` type can succeed with `EntityHash`, and a `u32` type doesn't.
However, you cannot actually tell from the type size whether `EntityHash` will panic. A `Hash` impl itself can make decisions about how it should be hashed, and pass less or more data than what the type actually consists of.

> The only _safety_ requirement we have is that the hashing be deterministic. It's fine for `hash()` to panic, as long as it panics deterministically, since there's no way to get duplicates of a value that fails to hash.

Yeah! Though I think even panicking non-deterministically is fine, "being deterministic" is only a property required when an execution does not diverge.

> > I said here `TrustedEntityBorrow` ensures determinism, but I just realized that it does not do that, because the current wording doesn't take into account that the `Hash` impl alone cannot guarantee correct behavior, ugh.
> 
> Sure, it does. A deterministic `Hash` is one that makes a deterministic sequence of calls to the provided `Hasher`. The fact that the `Hasher` can do non-deterministic things doesn't change the fact that the `Hash` was deterministic.
> 
> `Entity` has a deterministic `Hash`, and `TrustedEntityBorrow` is required to have the same `Hash` behavior as `Entity`, so `TrustedEntityBorrow` is required to have a deterministic `Hash`.

Calling `Hash::hash` on `Entity` is not deterministic for all `Hasher`s. Even if `Hash` calls `Hasher` deterministically, what happens inside of `Hash` is reliant on different trait code, that of `Hasher`. And unless we bind the code of `Hasher` to a safety contract, the execution of a `Hash` impl cannot be relied upon as deterministic by unsafe code.

And because `Hash` of `Entity` is not deterministic, the `Hash` of any `TrustedEntityBorrow` type doesn't have to be deterministic either!
Were I to paraphrase what the current wording in the safety contract of `TrustedEntityBorrow` is requiring:
"Ensure that all foreign `Hasher` impls are deterministic"
This is impossible to fulfill, which makes for an invalid safety contract, or at least one with no valid implementors.

Note, we can simply say "matches in `Hash` behavior when called with**EntityHasher**/a deterministic `Hasher`s".
We have then limited the scope of `Hash` impls we want guarantees about.

All of this is extremely subtle, to the point where I believe, that by this technicality, the official Rust docs for [`Hash`](https://doc.rust-lang.org/stable/std/hash/trait.Hash.html#hash-and-eq) are incorrect: 
`#[derive(PartialEq, Eq, Hash)]` does not prevent logic errors in all cases! The `H: Hasher` parameter would need a `DeterministicHasher` bound for this to be ensured.

Does this mean that those official docs have to change? Probably not, because most people don't actually have to worry about this. It is unsafe code (and thus `TrustedEntityBorrow`) that does!

Ultimately, I believe restricting combinations can avoid pushing the entire headache toward the user side:
Do you want a map over `Entity`? Use `EntityHashMap`. 
Do you want a map over an `Entity` newtype/EntityRef-like type? Use `EntityEquivalentHashMap`.
Do you want a map not covered by the above? Use `HashMap`.

The first two benefit from performance and `EntitySet`-compatibility, with the third, the user is free to manage these aspects themselves.

chescock:
> This is a really subtle discussion! It has design considerations on multiple levels, which I find/can make it difficult to hone in on where the different aspects come in/from, I hope the back and forth isn't getting frustrating!

Nope!  I'm worried we'll scare off other reviewers, though :).  

> Safety-wise, panicking is completely fine! This part is more about "what should an EntityEquivalentHashMap" be, that is the issue. To be a useful and compatible design, I think we should restrict it to use cases that make sense. Why does `std` not limit combinations? Because their `DefaultHasher` is intentionally as compatible as feasible!

This may be the fundamental disagreement.  I think we should follow a "simple things should be simple, complex things should be possible" approach.  So I agree that we want good *defaults* here: `EntityHashMap` and `EntityEquivalentHashMap` should default to `EntityHash`.  But I don't think we should spend a lot of effort *limiting* what users can do.  

> First, let me paint a situation from a standpoint of compatibility: We add an `EntityEquivalentHashMap` that has no restriction on `T: TrustedEntityBorrow`. Some library/code comes along, and makes use of `EntityEquivalentHashSet<&Entity>`. This code compiles perfectly fine! However, once somebody on a 32-bit platform comes along, their bevy program suddenly explodes. That is because on 64-bit platforms, where pointers are `u64`s, `EntityHash` will not panic. On 32-bit platforms, where pointers are `u32`s, `EntityHash` is suddenly being fed something other than `u64` and panics. Note that this can occur with any type that has platform/feature-dependent fields (which most pointer types do). If we allow all combinations, we will still have to list the types that should not be used, and have users manage this concern manually.

I mean, [`&Entity` *does* `impl TrustedEntityBorrow`](https://github.com/bevyengine/bevy/blob/c3f72ba4ad4a005b4560029ad65bb4efc9ad74d5/crates/bevy_ecs/src/entity/entity_set.rs#L69), which is fine because [`<&T as Hash>::hash()` just delegates to `T::Hash`](https://github.com/rust-lang/rust/blob/60a3084f64607e86dd7715d72f11764cd500c364/library/core/src/hash/mod.rs#L947).  If you meant `*const Entity`, that [calls `write_usize`](https://github.com/rust-lang/rust/blob/60a3084f64607e86dd7715d72f11764cd500c364/library/core/src/hash/mod.rs#L964) and would fail consistently on all platforms.  

But, why would we lift the `T: TrustedEntityBorrow` restriction?  Wouldn't that just wind up being an ordinary `HashSet`?  There's nothing we can do to stop `HashSet<*const Entity, EntityHash>`.  

This may be the other fundamental disagreement: I think we might agree that we don't need `DeterministicBuildHasher` *yet*, but you see a future use that I don't have the context for.  So I'd like to just punt the whole thing for now, and when you make a future change to loosen one of these bounds, I (and other reviewers) will be able to see the trait alongside the context that makes it necessary.  

> Yeah! Though I think even panicking non-deterministically is fine, "being deterministic" is a property only required when function execution does not diverge.

(Yeah, it's fine for `Hash` to panic non-deterministically, but it's not fine for `Hasher` to, because `Hash` could `catch_unwind` and turn it into regular non-determinism.  But panicking non-deterministically isn't *useful*, and the rules are simpler if we prohibit it.)  

> > Sure, it does. A deterministic `Hash` is one that makes a deterministic sequence of calls to the provided `Hasher`. The fact that the `Hasher` can do non-deterministic things doesn't change the fact that the `Hash` was deterministic.
> > `Entity` has a deterministic `Hash`, and `TrustedEntityBorrow` is required to have the same `Hash` behavior as `Entity`, so `TrustedEntityBorrow` is required to have a deterministic `Hash`.
> 
> Calling `Hash::hash` on `Entity` is not deterministic for all `Hasher`s. Even if `Hash` calls `Hasher` deterministically, what happens inside of `Hash` is reliant on different trait code, that of `Hasher`. And unless we bind the code of `Hasher` to a safety contract, the execution of a `Hash` impl cannot be relied upon as deterministic by unsafe code.

This is arguing semantics, but I don't think it's useful to consider a `Hash` nondeterministic just because you call it with a nondeterministic `Hasher`.  It seems a bit like saying `+` is nondeterministic because I can write `rng.random() + 1`.  

Either way, I think "makes the same sequence of calls to the provided `Hasher` that `self.entity().hash(state)` would" is the behavior we need from `TrustedEntityBorrow`.

Bluefinger (approved):
This now looks mostly mechanical changes to transition to `EntityEquivalent`, so there isn't much to worry about here, as long as the existing invariants are maintained around the trait.

Trashtalk217 (approved):
A really nice simple solution, looking forward to using this.

alice-i-cecile (changes requested):
Well-made and I like the idea. I couldn't find any new tests verifying that this works though :) Can you please add a unit test or three then I'll merge?

Victoronz:
Added tests, courtesy of @Trashtalk217.

alice-i-cecile (approved):

alice-i-cecile:
@Victoronz CI failures look real.

Victoronz:
Should just be some failed imports, on it.

PR #18409, opened 2025-03-19
Finish #17558, re-adding `insert_children`

ElliottjPierce:
# Objective

- Complete #17558.
- the `insert_children` method was previously removed, and as #17478 points out, needs to be added back.

## Solution

- Add a `OrderedRelationshipSourceCollection`, which allows sorting, ordering, rearranging, etc of a `RelationshipSourceCollection`.
- Implement `insert_related`
- Implement `insert_children`
- Tidy up some docs while I'm here.

## Testing

@bjoernp116 set up a unit test, and I added a doc test to `OrderedRelationshipSourceCollection`.

### DIFF
--- a/crates/bevy_ecs/src/relationship/mod.rs
+++ b/crates/bevy_ecs/src/relationship/mod.rs
@@ -158,12 +158,14 @@ pub trait RelationshipTarget: Component<Mutability = Mutable> + Sized {
     ///
     /// # Warning
     /// This should generally not be called by user code, as modifying the internal collection could invalidate the relationship.
+    /// The collection should not ever contain duplicates.
     fn collection_mut_risky(&mut self) -> &mut Self::Collection;
 
     /// Creates a new [`RelationshipTarget`] from the given [`RelationshipTarget::Collection`].
     ///
     /// # Warning
     /// This should generally not be called by user code, as constructing the internal collection could invalidate the relationship.
+    /// The collection should not ever contain duplicates.
     fn from_collection_risky(collection: Self::Collection) -> Self;
 
     /// The `on_replace` component hook that maintains the [`Relationship`] / [`RelationshipTarget`] connection.
--- a/crates/bevy_ecs/src/relationship/related_methods.rs
+++ b/crates/bevy_ecs/src/relationship/related_methods.rs
@@ -1,12 +1,15 @@
 use crate::{
     bundle::Bundle,
     entity::Entity,
+    prelude::DetectChangesMut,
     relationship::{Relationship, RelationshipTarget},
     system::{Commands, EntityCommands},
     world::{EntityWorldMut, World},
 };
 use alloc::vec::Vec;
 use core::marker::PhantomData;
+
+use super::OrderedRelationshipSourceCollection;
 
 impl<'w> EntityWorldMut<'w> {
     /// Spawns entities related to this entity (with the `R` relationship) by taking a function that operates on a [`RelatedSpawner`].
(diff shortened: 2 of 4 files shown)


crates/bevy_ecs/src/relationship/mod.rs
```diff
@@ -212,12 +212,14 @@ pub trait RelationshipTarget: Component<Mutability = Mutable> + Sized {
     ///
     /// # Warning
     /// This should generally not be called by user code, as modifying the internal collection could invalidate the relationship.
+    /// The collection should not never contain duplicates.
```
dmyyy, line 215:
```suggestion
    /// The collection should not contain duplicates.
```

crates/bevy_ecs/src/relationship/mod.rs
```diff
@@ -212,12 +212,14 @@ pub trait RelationshipTarget: Component<Mutability = Mutable> + Sized {
...
     fn collection_mut_risky(&mut self) -> &mut Self::Collection;
 
     /// Creates a new [`RelationshipTarget`] from the given [`RelationshipTarget::Collection`].
     ///
     /// # Warning
     /// This should generally not be called by user code, as constructing the internal collection could invalidate the relationship.
+    /// The collection should not never contain duplicates.
```
dmyyy, line 222:
```suggestion
    /// The collection should not contain duplicates.
```

crates/bevy_ecs/src/relationship/related_methods.rs
```diff
@@ -34,6 +37,25 @@ impl<'w> EntityWorldMut<'w> {
...
 
+    /// Relates the given entities to this entity with the relation `R`, starting at this particular index.
+    /// Note that this first adds the relation and then rearanges it.
+    /// See [`OrderedRelationshipSourceCollection::place`] for details on behavior.
+    pub fn insert_related<R: Relationship>(&mut self, index: usize, related: &[Entity]) -> &mut Self
+    where
+        <R::RelationshipTarget as RelationshipTarget>::Collection:
+            OrderedRelationshipSourceCollection,
+    {
+        self.add_related::<R>(related);
```
cart, line 48: In theory we can improve the performance here by taking advantage of the fact that we can know if the entities in `related` are in the RelationshipSourceCollection based on whether or not they have `R` (if do not have it, they aren't in it, if they do have `R` and it points to the target they are in it, and if they do have `R` and it does not point to the target they are not in it).

We could skip the insert hook (using RelationshipHookMode) and manually do the insert ourselves.

That would allow us to skip the `O(nm)` runtime that this currently has (for the remove op), in cases where the scan is not necessary (which should be most cases).

In cases where there are no reorders, it would also in theory allow us to do an optimized "resize collection, shift everything after index over, and insert everything in `related`", which would be much faster than inserting one-by one (which requires a "list shift" on every insert).

ElliottjPierce: I'm still relatively new to relationships, but I think I've made the requested changes. Let me know if I missed something here though.

It now skips the re-inserting if it already exists, preventing hooks. If we do need to add it, it no longer has to search for it.

> In cases where there are no reorders, it would also in theory allow us to do an optimized "resize collection, shift everything after index over, and insert everything in related", which would be much faster than inserting one-by one (which requires a "list shift" on every insert).

100% agree here, but I'm unsure how to practically do this. My best guess would be to remove all the duplicates from the new slice (which would allocate, maybe with bumpalo), then insert the relation component without hooks, then remove the entities in the passed slice from the target collection, and finally insert the slice into the vector directly. But between the allocation and removals IDK if that would be worth the performance. Maybe we could make it unsafe and presume that the entities have no duplicates and are not currently related?

crates/bevy_ecs/src/relationship/relationship_source_collection.rs
```diff
@@ -59,6 +59,58 @@ pub trait RelationshipSourceCollection {
     }
 }
 
+/// This trait signals that a [`RelationshipSourceCollection`] is ordered.
+pub trait OrderedRelationshipSourceCollection: RelationshipSourceCollection {
```
alice-i-cecile, line 63: Oh good I was hoping that something like this would be used.

alice-i-cecile (approved):

crates/bevy_ecs/src/relationship/related_methods.rs
```diff
@@ -36,6 +38,64 @@ impl<'w> EntityWorldMut<'w> {
         self
     }
 
+    /// Relates the given entities to this entity with the relation `R`, starting at this particular index.
+    ///
+    /// If the `related` has duplicates, a related entity will take the index of its last occorance in `related`.
```
viridia, line 43: occurrence

crates/bevy_ecs/src/relationship/related_methods.rs
```diff
@@ -36,6 +38,64 @@ impl<'w> EntityWorldMut<'w> {
...
+    /// If the indices go out of bounds, they will be "squeezed" into bounds.
```
viridia, line 44: The term of art here is "clamped" or "constrained".

bushrat011899 (approved):
Approved pending the documentation corrections already pointed out, nice work!

crates/bevy_ecs/src/relationship/related_methods.rs
ElliottjPierce: That's better, thanks!

PR #18410, opened 2025-03-19
Fix unecessary specialization checks for apps with many materials

### LINKED ISSUE #18388: Pipeline specialization tick not being cached, causing unnecessary re-specialization
For some reason, we aren't caching the pipeline specialization tick, which is causing the `needs_specialization` check in `specialize_material_meshes` to always return true, since there is no last specialized tick to compare the entity and view ticks to.

Additionally, we seem to be early returning frequently due to the material not being found, which is strange. Unclear if this is causing issues.

In the below logs from `3d_scene`, entity tick and view tick are stable, but the lack of a last specialized tick causes us to always try to specialize:
```shell
2025-03-18T06:38:11.273185Z  INFO bevy_pbr::material: Entity tick: Tick { tick: 316 }
2025-03-18T06:38:11.273190Z  INFO bevy_pbr::material: View tick: Tick { tick: 95 }
2025-03-18T06:38:11.273194Z  INFO bevy_pbr::material: Last specialized tick: None
2025-03-18T06:38:11.273212Z  INFO bevy_pbr::material: Material asset ID not found
```

Likely related to https://github.com/bevyengine/bevy/pull/17915.

tychedelia:
# Objective

For materials that aren't being used or a visible entity doesn't have an instance of, we were unnecessarily constantly checking whether they needed specialization, saying yes (because the material had never been specialized for that entity), and failing to look up the material instance.

## Solution

If an entity doesn't have an instance of the material, it can't possibly need specialization, so exit early before spending time doing the check.

Fixes #18388.

### DIFF
--- a/crates/bevy_pbr/src/material.rs
+++ b/crates/bevy_pbr/src/material.rs
@@ -863,6 +863,9 @@ pub fn specialize_material_meshes<M: Material>(
             .or_default();
 
         for (_, visible_entity) in visible_entities.iter::<Mesh3d>() {
+            let Some(material_asset_id) = render_material_instances.get(visible_entity) else {
+                continue;
+            };
             let entity_tick = entity_specialization_ticks.get(visible_entity).unwrap();
             let last_specialized_tick = view_specialized_material_pipeline_cache
                 .get(visible_entity)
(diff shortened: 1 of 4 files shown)


JMS55 (approved):
Lgtm.

We never did cold specialization for meshlet meshes, right?

tychedelia:
> Lgtm.
> 
> We never did cold specialization for meshlet meshes, right?

Yup, I promised to do so and then it slipped my mind. My 0.16 list is a little long right now but totally still happy to do it if there's time to get it in or right after.

JMS55:
No worries. Lets do it for 0.17.

superdump (approved):

Issue #18332, opened 2025-03-15
1-frame rendering flashes when changing the scene


JMS55:
## Bevy version

0.16-dev since 5e569af2d0efdac71beaa3e107c3a999717d8119.

## What you did

Run cargo run --example anti_aliasing, and toggle back and forth between the different AA methods (give the first few times a few seconds to compile the pipelines).

## What went wrong

Switching the AA method causes a 1-frame flash where the mesh doesn't render properly.

hukasu:
maybe the same thing that happens on `bloom_2d`?

https://github.com/user-attachments/assets/14e454d3-246f-4e12-b40a-83e87604eb01

hukasu:
`deferred_rendering` also has loading on the first time changing modes, but i don't know if both the examples that i brought are precisely the issue you are referring to

hukasu:
messing with the `anit_aliasing` example i saw that the flash is something else from the other 2 examples i mentioned

but i noticed this, the flash happens when changing out of `FXAA`, `SMAA`, and `TAA`

|To \ From | No AA | MSAA | FXAA | SMAA | TAA |
| -- | -- | -- | -- |-- | -- |
| No AA | | X | O | O | O |
| MSAA | X | | O | O | O |
| FXAA | X | X | | O | O |
| SMAA | X | X | O | | O |
| TAA | X | X | O | O | |

superdump:
Does X mean the flash happens and O means it doesn't?

hukasu:
@superdump the opposite, X no flash, O flash

JMS55:
I did a little more debugging.

The issue is specifically with shadows not rendering for 1(?) frame when something with the view changes (not sure exactly what yet).

If I comment out this line, the bug goes away.

https://github.com/bevyengine/bevy/blob/205ae6441fd200f923673d5f802d7132e519feb1/crates/bevy_pbr/src/render/light.rs#L1900

JMS55:
Even more debugging: it happens for exactly 1 frame, and it's due to this query:

https://github.com/bevyengine/bevy/blob/205ae6441fd200f923673d5f802d7132e519feb1/crates/bevy_pbr/src/render/light.rs#L1739

matching 0 entities.

This causes all_shadow_views to be empty, so the pipeline cache gets entirely emptied out.

PR #18412, opened 2025-03-19
Fix specialize_shadows system ordering

JMS55:
# Objective
- Fixes https://github.com/bevyengine/bevy/issues/18332

## Solution

- Move specialize_shadows to ManageViews so that it can run after prepare_lights, so that shadow views exist for specialization.
- Unfortunately this means that specialize_shadows is no longer in PrepareMeshes like the rest of the specialization systems.

## Testing
- Ran anti_aliasing example, switched between the different AA options, observed no glitches.

### DIFF
--- a/crates/bevy_pbr/src/material.rs
+++ b/crates/bevy_pbr/src/material.rs
@@ -351,9 +351,11 @@ where
                         Render,
                         (
                             check_views_lights_need_specialization.in_set(RenderSet::PrepareAssets),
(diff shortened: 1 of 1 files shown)


JMS55:
I'm still not really certain _why_ toggling the anti aliasing methods deletes the render world shadow views, but regardless the code is more correct after this PR (and fixes the bug).

tychedelia (approved):
Yup, totally makes sense. Thanks for tracking this down. These kinds of system set orderings can be really sneaky with retained causing these kinds of one frame weirdness.

alice-i-cecile (approved):

PR #18415, opened 2025-03-19
Introduce Displacement Map Example

craigfay:
## Summary of Changes

- Adds a new shader example that uses [MaterialPlugin](https://docs.rs/bevy/latest/bevy/prelude/struct.MaterialPlugin.html) to create a water effect using a scrolling noise texture.
- Introduces two low-resolution textures
  - toroidal grayscale noise (32 x 32)
  - a stone wall texture, based loosely on [cobble_stone.png](https://github.com/Render96/Render96Wiki/wiki/cobble_stone-(The-Texture-of-your-Childhood)) (32 x 32)

## Testing

Test using `cargo run --example displacement_map`

## WASM Support

Note: This example appears not to build successfully to `wasm32-unknown-unknown` using the [typical setup](https://bevy-cheatbook.github.io/platforms/wasm.html).

```
target/wasm32-unknown-unknown/debug/examples/displacement_map.wasm: target/wasm32-unknown-unknown/debug/examples/displacement_map.wasm: cannot execute binary file
```

---

## Showcase

| Description | Image |
| -- | -- |
| Screen recording of the example running (slowed down) | displacement_map |


alice-i-cecile:
Neat! I'm down to add more game-relevant rendering exmaples.

IceSentry:
I mentioned in discord that I prefer not having example where the important logic is only in the shader, but if we do decide to have that kind of examples they should live somewhere else that isn't the example/shader folder. This one is to show bevy apis related to writing shaders. The shaders themselves are rarely the important part of those examples.

We'd need something like a game_rendering or game_shader folder or something along those lines.

I won't push back any further on adding that kind of example as long as they aren't mixed with examples in `shader/`.

alice-i-cecile:
I'm going to close this out for now: we need a comprehensive plan for how to tackle these flavor of examples :)

craigfay:
That's totally fine 👍 Thanks for considering. Maybe worth saying: It's hard to overstate how ubiquitous and versatile displacement maps are in game development. I suspect I'm not the first or last who will be surprised not to find readily available examples for Bevy, given the relative size of `./examples`.

craigfay:
> I mentioned in discord that I prefer not having example where the important logic is only in the shader

@IceSentry good point. As a non-Bevy user, the shader logic in this example is completely uninteresting to me. What's non-obvious and relevant, is how to piece together Bevy primitives so that you're in a position to have one texture modifying another.

Issue #18423, opened 2025-03-19
Restructure rendering crates


superdump:
Discussions have been happening on Discord about some restructuring the rendering crates and there is some alignment across multiple independent thinkers so I wanted to capture it here. This is open for discussion. Comment and we can update this initial description as we change our minds / gain clarity.

# Current state

* bevy_mesh
* bevy_render
* bevy_core_pipeline
* bevy_sprite
* bevy_pbr

# Proposed states

Names can be discussed but should be illustrative of the kind of groupings we have in mind.

## Rob's suggestion

* bevy_mesh - mesh stuff not dependent on bevy_render
* bevy_render - agnostic rendering infrastructure like low level resources, render graph infrastructure, etc
* bevy_camera - camera stuff
* bevy_material - contains the one and only Material API
* bevy_post_processing - post-processing effects like bloom, motion blur, chromatic aberration, etc, etc
* bevy_antialiasing - all the AA implementations, FXAA, TAA, SMAA, etc
* bevy_2d - contains the code specific to 2D mesh material rendering, maybe also the default/base 2d render graph
* bevy_3d - contains the code specific to 3D mesh material rendering, including building blocks like clustering, lights, shadow mapping, etc, maybe also the default/base 3d render graph
* bevy_standard_material - the implementation of our standard PBR material using the other crates as necessary
* bevy_color_material - the implementation of a texture + color tint material that could be used also to support sprites, and the sprite abstraction on top of that

Things like lights, clustering, shadow mapping, should all be reusable for custom materials without having to pull in bevy's default materials (standard material, color material).

I think whether to use the 2D or 3D render graph should depend on the camera entity, and that the render phases should be functional for either. A phase is just a collection of drawable entities. The entities may have mesh data, and do have materials (sprites will be converted to color material using a shared and batchable quad mesh, or whatever mesh data performs best). 2D is then basically 3D in practice (the GPU thinks so) but perhaps you don't use 3D lighting infrastructure most of the time, but you could.

# Emerson's suggestion

* bevy material
* bevy_3d (contains infra for mesh rendering, light clustering, general 3d algos)
* bevy_2d (similar but probably cut down to 2d mesh rendering)
* bevy_pbr (library crate for pbr-specific stuff, probably mostly shaders and lights, maybe physical camera settings)
* (all of our effects crates like bevy_post_processing, bevy_aa, etc. Hopefully made graph agnostic at some point)
[...]

PR #18425, opened 2025-03-19
Add bevy_post_processing

IceSentry:
# Objective

- Follow up to #18323 and aligns with the points discussed in #18423 
- bevy_core_pipeline is still big and contains things that are not core pipelines. For example all the post processing effects.

## Solution

- Add bevy_post_processing crate
- The only missing post processing effect is tonemapping because it's used in the core3d module. This could be resolved by doing what is mentioned in this issue https://github.com/bevyengine/bevy/issues/16721

## Testing

- Ran various post processing effects examples and things still worked

---

## Notes

This one is probably a bit more controversial than anti_aliasing because it's harder to move every post processing effect to this new crate. Also, the line between what is and isn't post processing isn't as clear. We might want to simply use a subfolder instead of a new crate.


JMS55:
This gets tricky. 

Auto exposure shouldn't be here I don't think. DLSS is going to need to depend on it likely (haven't written that part yet). Which means that the anti aliasing crate would have to depend on the post processing crate.

Other potential issue, although I think you've handled this(?), is things like DOF and motion blur want to run before TAA/DLSS.

IceSentry:
The main limitation is the render graph being limited in how and when you can add nodes more than the crate itself. Right now things work because the PostProcessingPlugin gets insert bevy the AntiAliasing plugin. But it will start to break when they depend on each other. I'm fine with waiting on a better render graph. But I had the code written already so I opened the PR since for 0.16 it would help a bit with compile time.

robtfm (approved):
i think this is worthwhile, even if other things end up depending on both crates it'll still be faster overall (since both compile faster than a single unit with everything in).

i haven't reviewed in detail since it's mostly copy paste. the node ordering looks the same and the examples seem fine.

approved pending ci obviously

james7132:
Closing this out as #20778 was merged for 0.17.

PR #18432, opened 2025-03-20
Replace VisitEntities with MapEntities

cart:
# Objective

There are currently too many disparate ways to handle entity mapping, especially after #17687. We now have MapEntities, VisitEntities, VisitEntitiesMut, Component::visit_entities, Component::visit_entities_mut.

Our only known use case at the moment for these is entity mapping. This means we have significant consolidation potential.

Additionally, VisitEntitiesMut cannot be implemented for map-style collections like HashSets, as you cant "just" mutate a `&mut Entity`. Our current approach to Component mapping requires VisitEntitiesMut, meaning this category of entity collection isn't mappable. `MapEntities` is more generally applicable. Additionally, the _existence_ of the blanket From impl on VisitEntitiesMut blocks us from implementing MapEntities for HashSets (or any types we don't own), because the owner could always add a conflicting impl in the future.

## Solution

Use `MapEntities` everywhere and remove all "visit entities" usages.

* Add `Component::map_entities`
* Remove `Component::visit_entities`, `Component::visit_entities_mut`, `VisitEntities`, and `VisitEntitiesMut`
* Support deriving `Component::map_entities` in `#[derive(Coomponent)]`
* Add `#[derive(MapEntities)]`, and share logic with the `Component::map_entities` derive.
* Add `ComponentCloneCtx::queue_deferred`, which is command-like logic that runs immediately after normal clones. Reframe `FromWorld` fallback logic in the "reflect clone" impl to use it. This cuts out a lot of unnecessary work and I think justifies the existence of a pseudo-command interface (given how niche, yet performance sensitive this is).

Note that we no longer auto-impl entity mapping for ` IntoIterator<Item = &'a Entity>` types, as this would block our ability to implement cases like `HashMap`. This means the onus is on us (or type authors) to add explicit support for types that should be mappable.

Also note that the Component-related changes do not require a migration guide as there hasn't been a release with them yet.

## Migration Guide

If you were previously implementing `VisitEntities` or `VisitEntitiesMut` (likely via a derive), instead use `MapEntities`. Those were almost certainly used in the context of Bevy Scenes or reflection via `ReflectMapEntities`. If you have a case that uses `VisitEntities` or `VisitEntitiesMut` directly, where `MapEntities` is not a viable replacement, please let us know!

```rust
// before
#[derive(VisitEntities, VisitEntitiesMut)]
struct Inventory {
  items: Vec<Entity>,
  #[visit_entities(ignore)]
  label: String,
}

// after
#[derive(MapEntities)]
struct Inventory {
  #[entities]
  items: Vec<Entity>,
  label: String,
}
```


crates/bevy_ecs/src/component.rs
```diff
@@ -485,14 +485,21 @@ pub trait Component: Send + Sync + 'static {
...
-    /// Returns pointers to every entity stored on the component. This will be used to remap entity references when this entity
-    /// is cloned.
+    /// Maps the entities on this component using the given [`EntityMapper`]. This is used to remap entities in contexts like scenes and entity cloning.
+    /// When deriving [`Component`], this is populated by annotating fields containing entities with `#[entities]`
+    ///
+    /// ```
+    /// # use bevy_ecs::{component::Component, entity::Entity};
+    /// #[derive(Component)]
+    /// struct Inventory {
+    ///     #[entities]
```
cBournhonesque, line 495: Is it still possible to impl `MapEntities` manually in cases where the derive macro is not enough?

crates/bevy_ecs/src/component.rs
```diff
@@ -2986,35 +2989,20 @@ pub fn component_clone_via_reflect(
...
         let source_component_cloned = source_component_reflect.to_dynamic();
         let component_layout = component_info.layout();
         let target = ctx.target();
         let component_id = ctx.component_id();
-        for entity in mapped_entities.iter_mut() {
-            *entity = ctx.entity_mapper().get_mapped(*entity);
-        }
         drop(registry);
-        commands.queue(move |world: &mut World| {
+        ctx.queue_deferred(move |world: &mut World, mapper: &mut dyn EntityMapper| {
```
cBournhonesque, line 2997: Maybe add a comment here explaining that we cannot use `commands.queue` because we want to have access to EntityMapper inside the Command.

cBournhonesque (approved):

crates/bevy_ecs/src/component.rs
cart: Yup!

crates/bevy_ecs/src/entity/map_entities.rs
```diff
@@ -49,26 +49,62 @@ pub trait MapEntities {
     ///
     /// Implementors should look up any and all [`Entity`] values stored within `self` and
     /// update them to the mapped values via `entity_mapper`.
-    fn map_entities<M: EntityMapper>(&mut self, entity_mapper: &mut M);
+    fn map_entities<E: EntityMapper>(&mut self, entity_mapper: &mut E);
+}
+
+impl MapEntities for Entity {
+    fn map_entities<E: EntityMapper>(&mut self, entity_mapper: &mut E) {
+        *self = entity_mapper.get_mapped(*self);
```
Brezak, line 57: Does it make sense to map the entity when it's `Entity::PLACEHOLDER`. At least for one-to-one relationships we use `Enttiy::PLACEHOLDER` to mean empty.

crates/bevy_ecs/src/component.rs
Bleachfuel: I was thinking about making a derive MapEntities but we would need to have it be able to be recursive and have entities be directly used. What would you think would be a good name for the attribute?

cBournhonesque: You could have non-standard data structures containing entities

Brezak:
I've opened up a [pull request](https://github.com/cart/bevy/pull/33) on this pull request to implement  `MapEntities` on `EntityHashSet` so we can finally use it for relationships.

crates/bevy_ecs/src/component.rs
Bleachfuel: It would be possible if the data structures also derived MapEntities, we could have a #[mapentities], which would tell the derive to map entities a level deeper

Shatur (approved):
I'm excited about this change!

crates/bevy_ecs/src/entity/map_entities.rs
cart: That should be up to the EntityMapper impl I think.

crates/bevy_ecs/src/component.rs
```diff
@@ -2921,7 +2928,7 @@ pub fn component_clone_via_clone<C: Clone + Component>(
 /// [`PartialReflect::reflect_clone`]: bevy_reflect::PartialReflect::reflect_clone
 #[cfg(feature = "bevy_reflect")]
 pub fn component_clone_via_reflect(
-    commands: &mut Commands,
+    _commands: &mut Commands,
```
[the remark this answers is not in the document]
cart: Yeah I'm down to remove it until someone yells.

crates/bevy_ecs/src/entity/map_entities.rs
Brezak: Makes sense.

crates/bevy_ecs/src/component.rs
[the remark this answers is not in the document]
cart: Removed!

cart:
@Brezak 

> I've opened up a https://github.com/cart/bevy/pull/33 on this pull request to implement MapEntities on EntityHashSet so we can finally use it for relationships.

I'm going to hold off on merging this, as I think I'd like to walk back the `EntityHashSet` wrapper type and use a type alias instead, which would allow us to use the general hashset impl instead.

PR #13804, opened 2024-06-11
Revert "constrain WorldQuery::init_state argument to ComponentInitial…

hymm:
…izer (#13442)"

This reverts commit 5cfb063d4ae4b6abb40c3e7d2c7752a68ab5eb3d.

- This PR broke bevy-trait-query, which needs to be able to write a resource in init_state. See #13798 for more details.
- Note this doesn't fix everything as transmutes for bevy-trait-query will still be broken,. But the current usage in that crate is UB, so we need to find another solution.

PR #13442, opened 2024-05-20
constrain WorldQuery::init_state argument to ComponentInitializer

Victoronz:
# Objective

In #13343, `WorldQuery::get_state` was constrained from `&World` as the argument to `&Components`, but `WorldQuery::init_state` hasn't yet been changed from `&mut World` to match.

Fixes #13358

## Solution

Create a wrapper around `&mut Components` and `&mut Storages` that can be obtained from `&mut World` with a `component_initializer` method. 
This new `ComponentInitializer` re-exposes the API on `&mut Components` minus the `&mut Storages` parameter where it was present. For the `&Components` API, it simply derefs to its `components` field. 

### Added
The `World::component_initializer` method.
The `ComponentInitializer` struct that re-exposes `Components` API.
### Changed
`WorldQuery::init_state` now takes `&mut ComponentInitializer` instead of `&mut World`.

## Migration Guide
Instead of passing `&mut World` to `WorldQuery::init_state` directly, pass in a mutable reference to the struct returned from `World::component_initializer`.

PR #18434, opened 2025-03-20
Replace `try_query`

ElliottjPierce:
# Objective

Readonly queries should not need to mutate a world, not even to initialize it. Currently the `try_query` api exists to remedy this, but it requires all component ids, etc to be created ahead of time. This is not idea.

## Solution

Make all `WorldQuery` just take `&World` to initialize. Now instead of needing `try_query`, users can just call `query`.

## Testing

No new tests were added, though the existing ones did catch some bugs.

## Migration Guide

`WorldQuery::init_state` now takes `&World` instead of `&mut World`. Additionally, it is now explicitly unsafe to rely on "magic" values in `WorldQuery::init_state`. For example, queries can no longer access resources or entities during initialization. This was possible before but was an anti-pattern since it tied the query results to when the query was *created* instead of when it is *queried*.

Additionally, `Bundle::get_component_ids` now takes `ComponentsQueuedRegistrator` instead of `&Components` and outputs `ComponentId`s instead of `Option<ComponentId>`.

### DIFF
--- a/benches/benches/bevy_ecs/world/world_get.rs
+++ b/benches/benches/bevy_ecs/world/world_get.rs
@@ -101,7 +101,7 @@ pub fn world_query_get(criterion: &mut Criterion) {
 
     for entity_count in RANGE.map(|i| i * 10_000) {
         group.bench_function(format!("{}_entities_table", entity_count), |bencher| {
-            let mut world = setup::<Table>(entity_count);
+            let world = setup::<Table>(entity_count);
             let mut query = world.query::<&Table>();
 
             bencher.iter(|| {
@@ -112,7 +112,7 @@ pub fn world_query_get(criterion: &mut Criterion) {
             });
         });
         group.bench_function(format!("{}_entities_table_wide", entity_count), |bencher| {
-            let mut world = setup_wide::<(
+            let world = setup_wide::<(
                 WideTable<0>,
                 WideTable<1>,
                 WideTable<2>,
@@ -137,7 +137,7 @@ pub fn world_query_get(criterion: &mut Criterion) {
             });
         });
         group.bench_function(format!("{}_entities_sparse", entity_count), |bencher| {
-            let mut world = setup::<Sparse>(entity_count);
+            let world = setup::<Sparse>(entity_count);
             let mut query = world.query::<&Sparse>();
 
             bencher.iter(|| {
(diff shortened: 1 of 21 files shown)


Victoronz:
I agree that that `&mut World` parameter is bad, but how does this PR avoid the same issue #13442 ran into?
This will break `bevy-trait-query`, which was deemed bad enough for that aforementioned change to be reverted in #13804.
This was tracked in issue #13358 (not sure why this was closed).

ElliottjPierce:
> I agree that that `&mut World` parameter is bad, but how does this PR avoid the same issue #13442 ran into? This will break `bevy-trait-query`, which was deemed bad enough for that aforementioned change to be reverted in #13804. This was tracked in issue #13358 (not sure why this was closed).

I didn't realize that. Yikes. Looking at the source, I think `bevy-trait-query` could still work here with some statics. The path where it is initialized is already marked as cold, so I doubt moving from a resource to a static would hurt performance.
But I just glanced at it. @RobWalt (`bevy-trait-query` author), do you think you can still make your crate work with this change? If not, we should definitely hold off here.

If we can't make this work with `bevy-trait-query`, we'll need to delay until components as entities and in house trait queries are implemented.

Victoronz:
Note that `bevy-trait-query` is no longer the only user.
#16810 has adopted this pattern too, and #16843 enabled registering resource access to allow that to be sound.

I think removing the `&mut World` parameter will still be possible and desirable, but has to wait on Resources-as-Entities, was the conclusion when this matter was last discussed AFAIK.

ElliottjPierce:
> Note that `bevy-trait-query` is no longer the only user. #16810 has adopted this pattern too, and #16843 enabled registering resource access to allow that to be sound.

Yeah, using resources in `fetch` is fine. We just don't insert them on `init_state`. All this really does IMO is loose some ergonomics. Users have to add the resource manually, but then everything works fine.

> I think removing the `&mut World` parameter will still be possible and desirable, but has to wait on proper Resources-as-Entities, was the conclusion when this matter was last discussed AFAIK.

Ok. I don't mind waiting for that, and components as entities, etc. This was supper easy to do, so revisiting it shouldn't be a problem. 

I'm still convinced we can do this now, with pretty much no cost, but I understand the argument for holding off. I'll leave the timing here up to y'all.

ElliottjPierce:
Closing in favor of the more incremental #22670.

Issue #18416, opened 2025-03-19
`bevy_diagnostic` adding RenderDiagnosticsPlugin makes app panic


shwwwa:
## Bevy version

0.15.3

## \[Optional\] Relevant system information

If you cannot get Bevy to build or run on your machine, please include:
SystemInfo { os: "Linux  Void", kernel: "6.12.19_1", cpu: "Intel(R) Core(TM) i9-9900K CPU @ 3.60GHz", core_count: "8", memory: "15.6 GiB" }
rust 1.87.0-nightly
If your bug is rendering-related, copy the adapter info that appears when you run Bevy.

2025-03-19T06:33:47.898504Z  INFO bevy_render::renderer: AdapterInfo { name: "AMD Radeon RX 6700 XT (RADV NAVI22)", vendor: 4098, device: 29663, device_type: DiscreteGpu, driver: "radv", driver_info: "Mesa 24.3.4", backend: Vulkan }

You should also consider testing the examples of our upstream dependencies to help isolate any setup-specific issue:

- [`wgpu`](https://github.com/gfx-rs/wgpu) for rendering problems
- [`winit`](https://github.com/rust-windowing/winit) for input and window management
- [`gilrs`](https://docs.rs/gilrs/latest/gilrs/) for gamepad inputs

## What you did

 .add_plugins((FrameTimeDiagnosticsPlugin, SystemInformationDiagnosticsPlugin, RenderDiagnosticsPlugin::default()))
added renderdiagnosticsplugin

## What went wrong

thread 'main' panicked at /home/caffidev/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/bevy_render-0.15.3/src/diagnostic/mod.rs:66:41:
Requested resource bevy_render::renderer::render_device::RenderDevice does not exist in the `World`.
                Did you forget to add it using `app.insert_resource` / `app.init_resource`?
                Resources are also implicitly added via `app.add_event`,
                and can be added by plugins.
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

If it's not clear, break this out into:

- what were you expecting?
- what actually happened?

## Additional information

Other information that can be used to further reproduce or isolate the problem.
This commonly includes:

- screenshots
- logs
- theories about what might be going wrong
- workarounds that you used
- links to related bugs, PRs or discussions

shwwwa:
please reply if something else is needed
it's not critical for me because this plugin does not provide what i want to get (adapterinfo)

[...]

PR #18435, opened 2025-03-20
info: update log_diagnostics.rs example

shwwwa:
https://github.com/bevyengine/bevy/issues/18416#issuecomment-2735964348

### DIFF
--- a/examples/diagnostics/log_diagnostics.rs
+++ b/examples/diagnostics/log_diagnostics.rs
@@ -8,7 +8,7 @@ use bevy::{
 fn main() {
     App::new()
         .add_plugins((
-            // When using RenderDiagnosticsPlugin, DefaultPlugins must be initialized first in order to prevent panic.
+            // The diagnostics plugins need to be added after DefaultPlugins as they use e.g. the time plugin for timestamps.
             DefaultPlugins,
             // Adds frame time diagnostics
             FrameTimeDiagnosticsPlugin::default(),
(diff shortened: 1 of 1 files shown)


examples/diagnostics/log_diagnostics.rs
```diff
@@ -8,6 +8,7 @@ use bevy::{
 fn main() {
     App::new()
         .add_plugins((
+            // When using RenderDiagnosticsPlugin, DefaultPlugins must be initialized first in order to prevent panic.
```
kristoff3r, line 11:
```suggestion
            // The diagnostics plugins need to be added after DefaultPlugins as they use e.g. the time plugin for timestamps.
```

kristoff3r (commented):
This looks like a good thing to point out. More of the plugins have the same limitation so I suggest changing the language slightly.

Other things I noticed in this example that would be nice to fix now that we're updating it:
* The SystemInformationDiagnosticsPlugin line needs a trailing comma
* AssetCountDiagnosticsPlugin doesn't seem to exist anymore, but I can't find the PR where it was removed? Not sure why but for now that line should probably be removed.
* I feel like these should not be commented out, the log output isn't that bad with everything enabled and it prevents errors

examples/diagnostics/log_diagnostics.rs
shwwwa: Do as you want guys

kristoff3r: > Do as you want guys ( I allowed edits by maintainers )

I'm not a maintainer, but I can adopt the PR if you don't feel like doing changes?

shwwwa: sure

kristoff3r:
Was fixed in #18652

PR #18437, opened 2025-03-20
Reduce dependencies on `bevy_render` by preferring `bevy_mesh` imports

greeble-dev:
## Objective

Reduce dependencies on `bevy_render` by preferring `bevy_mesh` imports over `bevy_render` re-exports. 

```diff
- use bevy_render::mesh::Mesh;
+ use bevy_mesh::Mesh;
```

This is intended to help with #18423 (render crate restructure). Affects `bevy_gltf`, `bevy_animation` and `bevy_picking`.

## But Why?

As part of #18423, I'm assuming there'll be a push to make crates less dependent on the big render crates. This PR seemed like a small and safe step along that path - it only changes imports and makes the `bevy_mesh` crate dependency explicit in `Cargo.toml`. Any remaining dependencies on `bevy_render` are true dependencies.

## Testing

```
cargo run --example testbed_3d
cargo run --example mesh_picking
```


mgi388 (approved):
Code LGTM, CI passes, and being just an import change all makes sense to me.

JMS55 (approved):

PR #18439, opened 2025-03-20
Newtype `Anchor`

### LINKED ISSUE #18459: `Anchor`s doc comment's are sprite specific but it is also used with `Text2d`

`Anchor`s doc comments are sprite specific:
``` How a sprite is positioned relative to its [`Transform`]```
but `Anchor` is also used with `Text2d` which the documentation should acknowledge.

### LINKED ISSUE #18460: `Anchor` doesn't need to be an enum

`Anchor` doesn't need to an enum, it's variants are all mapped to `Vec2`s.

There's no reason to hide the underlying `Vec2` representation. An enum is less efficient as it requires extra space for it's discriminator.

`Anchor` newtypes `Vec2`, its variants are replaced with `Vec2` constants.

### PR DESCRIPTION
**Pull request by ickshonpe**
# Objective

The `Anchor` component doesn't need to be a enum. The variants are just mapped to `Vec2`s so it could be changed to a newtype with associated const values, saving the space needed for the discriminator by the enum.

Also there was no benefit I think in hiding the underlying `Vec2` representation of `Anchor`s.

Suggested by @atlv24.

Fixes #18459
Fixes #18460

## Solution

Change `Anchor` to a struct newtyping a `Vec2`, and its variants into associated constants.

## Migration Guide

The anchor component has been changed from an enum to a struct newtyping a `Vec2`. The `Custom` variant has been removed, instead to construct a custom `Anchor` use its tuple constructor:
```rust
Sprite {
     anchor: Anchor(Vec2::new(0.25, 0.4)),
     ..default()
}
```
The other enum variants have been replaced with corresponding constants:
* `Anchor::BottomLeft` to `Anchor::BOTTOM_LEFT`
* `Anchor::Center` to `Anchor::CENTER`
* `Anchor::TopRight` to `Anchor::TOP_RIGHT`
* .. and so on for the remaining variants


mnmaita:
I'm impartial to this change. Just wanted to note that this requires a migration guide, not only because the naming changes due to constants but also on how to make a custom Anchor. The custom variant was removed from the example AFAICS but I assume this can still be achieved by using `From<Vec2>` right?

crates/bevy_sprite/src/sprite.rs
```diff
@@ -240,41 +241,31 @@ pub enum ScalingMode {
     FitEnd,
 }
 
-/// How a sprite is positioned relative to its [`Transform`].
-/// It defaults to `Anchor::Center`.
-#[derive(Component, Debug, Clone, Copy, PartialEq, Default, Reflect)]
+/// Normalized (relative to its size) offset of a 2d renderable entity from its [`Transform`].
+#[derive(Component, Debug, Clone, Copy, PartialEq, Deref, DerefMut, Default, Reflect)]
```
mnmaita, line 245: What do you think about doing an explicit Default impl which returns `CENTER` instead of relying on the `Vec2` default value? I know they are the same but it feels less obscure to me.

ickshonpe: Yeah that's a good idea I think, it's better to be explicit in this case.

ickshonpe:
> I'm impartial to this change. Just wanted to note that this requires a migration guide, not only because of the naming changes due to constants but also on how to make a custom Anchor. The custom variant was removed from the example AFAICS but I assume this can still be achieved by using `From<Vec2>` right?

Yes and since it newtypes the value just `Anchor(Vec2::new(0.25, 0.4))` works, so the custom function is redundant.

Another advantage of this implementation is that it allows composition so you if you want an achor relative to the top left corner of the sprite you can do:
```
Sprite {
    image,
    anchor: (*Self::TOP_RIGHT - Vec2::new(0.25, 0.0)).into(),
    ..default()
}
```
You could do the same with I guess:
```
    anchor: Anchor::custom(Anchor::TopRight.as_vec2() - Vec2::new(0.25, 0.0)), 
```
but it's a bit more cumbersome and I don't think there's any benefit to hiding the underlying representation Vec2 representation of anchors.

atornity (approved):

MrGVSV (approved):

mockersf (approved):

PR #18451, opened 2025-03-20
Get names of queued components

ElliottjPierce:
# Objective

#18173 allows components to be queued without being fully registered. But much of bevy's debug logging contained `components.get_name(id).unwrap()`. However, this panics when the id is queued. This PR fixes this, allowing names to be retrieved for debugging purposes, etc, even while they're still queued.

## Solution

We change `ComponentInfo::descriptor` to be `Arc<ComponentDescriptor>` instead of not arc'd. This lets us pass the descriptor around (as a name or otherwise) as needed. The alternative would require some form of `MappedRwLockReadGuard`, which is unstable, and would be terribly blocking. Putting it in an arc also signifies that it doesn't change, which is a nice signal to users. This does mean there's an extra pointer dereference, but I don't think that's an issue here, as almost all paths that use this are for debugging purposes or one-time set ups.

## Testing

Existing tests.

## Migration Guide

`Components::get_name` now returns `Option<Cow<'_, str>` instead of `Option<&str>`. This is because it now returns results for queued components. If that behavior is not desired, or you know the component is not queued, you can use `components.get_info().map(ComponentInfo::name)` instead.

Similarly, `ScheduleGraph::conflicts_to_string` now returns `impl Iterator<Item = (String, String, Vec<Cow<str>>)>` instead of `impl Iterator<Item = (String, String, Vec<&str>)>`. Because `Cow<str>` derefs to `&str`, most use cases can remain unchanged.

### DIFF
--- a/crates/bevy_ecs/src/component.rs
+++ b/crates/bevy_ecs/src/component.rs
@@ -2005,7 +2005,7 @@ impl Components {
             })
     }
 
-    /// Gets the [`ComponentName`] of the component with this [`ComponentId`] if it is present.
+    /// Gets the name of the component with this [`ComponentId`] if it is present.
     /// This will return `None` only if the id is neither regisered nor queued to be registered.
     ///
     /// This will return an incorrect result if `id` did not come from the same world as `self`. It may return `None` or a garbage value.
--- a/crates/bevy_ecs/src/component.rs
+++ b/crates/bevy_ecs/src/component.rs
@@ -13,8 +13,8 @@ use crate::{
     world::{DeferredWorld, FromWorld, World},
 };
 use alloc::boxed::Box;
+use alloc::string::ToString;
 use alloc::{borrow::Cow, format, vec::Vec};
-use atomicow::CowArc;
 pub use bevy_ecs_macros::Component;
 use bevy_platform_support::sync::Arc;
 use bevy_platform_support::{
@@ -981,36 +981,6 @@ impl SparseSetIndex for ComponentId {
     #[inline]
     fn get_sparse_set_index(value: usize) -> Self {
         Self(value)
-    }
-}
-
-/// Represents the name of a component.
-#[derive(Clone)]
-pub struct ComponentName<'a>(pub CowArc<'a, ComponentDescriptor>);
-
-impl Deref for ComponentName<'_> {
-    type Target = str;
-
-    fn deref(&self) -> &Self::Target {
-        self.0.name()
-    }
-}
-
-impl Debug for ComponentName<'_> {
-    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
-        write!(f, "{}", self.0.name())
-    }
-}
-
-impl core::fmt::Display for ComponentName<'_> {
-    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
-        write!(f, "{}", self.0.name())
-    }
-}
-
-impl From<ComponentName<'_>> for alloc::string::String {
-    fn from(value: ComponentName) -> Self {
-        value.0.name().into()
     }
 }
 
@@ -2018,13 +1988,10 @@ impl Components {
     ///
     /// This will return an incorrect result if `id` did not come from the same world as `self`. It may return `None` or a garbage value.
     #[inline]
-    pub fn get_descriptor<'a>(
-        &'a self,
-        id: ComponentId,
-    ) -> Option<CowArc<'a, ComponentDescriptor>> {
+    pub fn get_descriptor<'a>(&'a self, id: ComponentId) -> Option<Cow<'a, ComponentDescriptor>> {
         self.components
             .get(id.0)
-            .and_then(|info| info.as_ref().map(|info| CowArc::Borrowed(&info.descriptor)))
+            .and_then(|info| info.as_ref().map(|info| Cow::Borrowed(&info.descriptor)))
             .or_else(|| {
                 let queued = self.queued.read().unwrap_or_else(PoisonError::into_inner);
                 // first check components, then resources, then dynamic
@@ -2034,7 +2001,7 @@ impl Components {
                     .chain(queued.resources.values())
                     .chain(queued.dynamic_registrations.iter())
                     .find(|queued| queued.id == id)
-                    .map(|queued| CowArc::Owned(Arc::new(queued.descriptor.clone())))
+                    .map(|queued| Cow::Owned(queued.descriptor.clone()))
             })
     }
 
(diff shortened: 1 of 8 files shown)


crates/bevy_ecs/src/component.rs
```diff
@@ -1959,13 +2020,49 @@ impl Components {
...
+        self.components
+            .get(id.0)
+            .and_then(|info| info.as_ref().map(|info| info.descriptor.clone()))
+            .or_else(|| {
+                let queued = self.queued.read().unwrap_or_else(PoisonError::into_inner);
+                // first check components
+                queued
+                    .components
+                    .values()
+                    .find(|queued| queued.id == id)
```
chescock, line 2038: The linear search here looks expensive.  Instead of putting the descriptors in the individual queues, would it make sense to have a separate `HashMap<ComponentId, Arc<ComponentDescriptor>>`?  For that matter, the `ComponentId`s are dense, so I think it could be `Vec<Arc<ComponentDescriptor>>`.

crates/bevy_ecs/src/component.rs
```diff
@@ -813,7 +813,7 @@ impl ComponentHooks {
 #[derive(Debug, Clone)]
 pub struct ComponentInfo {
     id: ComponentId,
-    descriptor: ComponentDescriptor,
+    descriptor: Arc<ComponentDescriptor>,
```
chescock, line 816: It's unfortunate to have to wrap ordinary `ComponentDescriptor`s in an `Arc` even though most of them aren't behind a lock.  That means we have to do a `clone()` on every lookup!  

Would it work to have `get_descriptor` return an `atomicow::CowArc<ComponentDescriptor>`?  That would let you return either a borrowed value or an owned `Arc`, and it would deref to `ComponentDescriptor`.

crates/bevy_ecs/src/component.rs
ElliottjPierce: Yeah, the linear search is not ideal, but this is a cold path. Ids are almost always correct unless this is in a inter-world context, and queued components aren't queued for long. I genuinely think it would be slower to cache it's location somewhere or change the backing data structure because that slows everything down by a tiny bit, where as this slows almost nothing down (but by a lot).

That's why I did it this way (even though it hurts my eyes too). But if the consensus is different, I can absolutely do otherwise.

crates/bevy_ecs/src/component.rs
cart, line 816: This adds additional indirection to something that is in the hot path. Doing this to improve debuggability feels like a hard pill to swallow.

ElliottjPierce: I just looked at the usages, and you're right (as usual). I need the atomicow crate to have no_std support in order to fix this. It has it on main, but not on latest release. Once that's out, I can add this.

Or I can source it directly from github, but if this is going in a release anytime soon, I'd rather not depend on atomicow's main branch.

alice-i-cecile: Just cut an atomicow 1.1 release for you :)

ElliottjPierce: And, it's updated. Thanks!

Should we re-export this crate too? It's part of the public API now but not all users will have atomicow as a dependency. I can create a wrapper for it so the `CowArc` is not public, we could add it to the `bevy` crate's prelude, or we could just let users add it if they want to use the relevant apis. What do y'all think?

chescock: > It's part of the public API now but not all users will have atomicow as a dependency. I can create a wrapper for it so the `CowArc` is not public, we could add it to the `bevy` crate's prelude, or we could just let users add it if they want to use the relevant apis. What do y'all think?

Would it work to have the return type be `impl Deref<Target = ComponentDescriptor>`?

alice-i-cecile: I would prefer to avoid re-exporting except where it's absolutely essential. Adding a dependency isn't hard.

ElliottjPierce: > Would it work to have the return type be `impl Deref<Target = ComponentDescriptor>`?

I would think more like `impl Deref<Target = ComponentDescriptor> + Clone`. That could work, but I'm generally not a huge fan of returning concrete types in `impl Trait` format, except iterators of course.

> I would prefer to avoid re-exporting except where it's absolutely essential. Adding a dependency isn't hard.

That's true and it has to be built anyway. Plus, this gives visibility to the crate. It's an excellent utility that I didn't know existed. I was thinking about making almost exactly an `CowArc` crate myself to enforce immutability until 
@chescock mentioned it. So I think letting users add the dependency is the best option.

I'm happy to do otherwise though if a different consensus is reached.

alice-i-cecile (approved):

crates/bevy_ecs/src/component.rs
```diff
@@ -1959,13 +2011,54 @@ impl Components {
...
+            .get(id.0)
+            .and_then(|info| info.as_ref().map(|info| CowArc::Borrowed(&info.descriptor)))
+            .or_else(|| {
+                let queued = self.queued.read().unwrap_or_else(PoisonError::into_inner);
+                // first check components
+                queued
+                    .components
+                    .values()
+                    .find(|queued| queued.id == id)
+                    .map(|queued| CowArc::Owned(Arc::new(queued.descriptor.clone())))
```
chescock, line 2035: Did you mean to make `queued.descriptor` be an `Arc`, so that this just clones the `Arc`?  Right now, you're cloning the underlying descriptor and allocating a new `Arc` each time.  

If you *do* want to clone the descriptor, then there's no need for an `Arc` and you can return an ordinary `Cow<ComponentDescriptor>` here.  

For that matter, if you're willing to just have `get_name` and not a general `get_descriptor`, you could return `Cow<str>` and just have to clone `name`.  It's a `Cow::Borrowed` in all cases except dynamic components, so that will usually be a cheap reference copy.

crates/bevy_ecs/src/component.rs
```diff
@@ -1959,13 +2011,54 @@ impl Components {
...
+                            .map(|queued| CowArc::Owned(Arc::new(queued.descriptor.clone())))
+                    })
+                    .or_else(|| {
+                        // otherwise check dynamic
+                        queued
+                            .dynamic_registrations
+                            .iter()
+                            .find(|queued| queued.id == id)
+                            .map(|queued| CowArc::Owned(Arc::new(queued.descriptor.clone())))
+                    })
```
chescock, line 2051: Since you do the same thing with each `Iterator`, it might be more clear to combine them with `chain()` and do `find()`/`map()` once: 

```suggestion
                // first check components, then resources, then dynamic
                queued
                    .components
                    .values()
                    .chain(queued.resources.values())
                    .chain(queued.dynamic_registrations.iter())
                    .find(|queued| queued.id == id)
                    .map(|queued| CowArc::Owned(Arc::new(queued.descriptor.clone())))
```

chescock (approved):
I left some suggestions, but neither of them should be blocking.  

The linear scan still makes me nervous!  I'd feel better if we instead added a `Vec<ComponentDescriptor>` to `QueuedRegistration`.  We know that it's a dense range of indices, so we can index it by offsetting `ComponentId`.  

But if we're okay with the linear scan, then this is a solid implementation of that approach and we should merge it!

crates/bevy_ecs/src/component.rs
ElliottjPierce: Yeah this was a mess. 

Originally it was an arc, which is why we needed a new release of atomicow.
But it being an arc, slowed down other operations because of the extra deref, so we removed it.
So this is basically an artifact. Nice catch.

Now it's removed and we can (ironically) just do normal `Cow`s.

ElliottjPierce:
> The linear scan still makes me nervous! I'd feel better if we instead added a `Vec<ComponentDescriptor>` to `QueuedRegistration`. We know that it's a dense range of indices, so we can index it by offsetting `ComponentId`.

It hurts me to see something unoptimized too, but this is a *really* slow path, and I think introducing an invariant (even though it's incidentally upheld today), is not worth it.

---

The migration guide has also been updated BTW.

crates/bevy_ecs/src/component.rs
```diff
@@ -1959,13 +1981,53 @@ impl Components {
...
+            .or_else(|| {
+                let queued = self.queued.read().unwrap_or_else(PoisonError::into_inner);
+                // first check components, then resources, then dynamic
+                queued
+                    .components
+                    .values()
+                    .chain(queued.resources.values())
+                    .chain(queued.dynamic_registrations.iter())
+                    .find(|queued| queued.id == id)
+                    .map(|queued| Cow::Owned(queued.descriptor.name().to_string()))
```
chescock, line 2029: I think you can access the field directly from here and `clone()` the `Cow` itself so that you don't need to allocate a `String` if it's already a `Cow::Borrowed`.  

```suggestion
                    .map(|queued| queued.descriptor.name.clone())
```

PR #18454, opened 2025-03-20
Make system param validation rely on the unified ECS error handling via the GLOBAL_ERROR_HANDLER

alice-i-cecile:
# Objective

There are two related problems here:

1. Users should be able to change the fallback behavior of *all* ECS-based errors in their application by setting the `GLOBAL_ERROR_HANDLER`. See #18351 for earlier work in this vein.
2. The existing solution (#15500) for customizing this behavior is high on boilerplate, not global and adds a great deal of complexity.

The consensus is that the default behavior when a parameter fails validation should be set based on the kind of system parameter in question: `Single` / `Populated` should silently skip the system, but `Res` should panic. Setting this behavior at the system level is a bandaid that makes getting to that ideal behavior more painful, and can mask real failures (if a resource is missing but you've ignored a system to make the Single stop panicking you're going to have a bad day).

## Solution

I've removed the existing `ParamWarnPolicy`-based configuration, and wired up the `GLOBAL_ERROR_HANDLER`/`default_error_handler` to the various schedule executors to properly plumb through errors .

Additionally, I've done a small cleanup pass on the corresponding example.

## Testing

I've run the `fallible_params` example, with both the default and a custom global error handler. The former panics (as expected), and the latter spams the error console with warnings 🥲 

## Questions for reviewers

1. Currently, failed system param validation will result in endless console spam. Do you want me to implement a solution for warn_once-style debouncing somehow?
2. Currently, the error reporting for failed system param validation is very limited: all we get is that a system param failed validation and the name of the system. Do you want me to implement improved error reporting by bubbling up errors in this PR?
3. There is broad consensus that the default behavior for failed system param validation should be set on a per-system param basis. Would you like me to implement that in this PR?

My gut instinct is that we absolutely want to solve 2 and 3, but it will be much easier to do that work (and review it) if we split the PRs apart.

## Migration Guide

`ParamWarnPolicy` and the `WithParamWarnPolicy` have been removed completely. Failures during system param validation are now handled via the `GLOBAL_ERROR_HANDLER`: please see the `bevy_ecs::error` module docs for more information.


crates/bevy_ecs/src/system/system_param.rs
```diff
@@ -2592,6 +2578,23 @@ unsafe impl SystemParam for FilteredResourcesMut<'_, '_> {
...
+    #[error("System failed to validate parameter")]
+    System,
+    /// A run condition failed to validate a parameter.
+    #[error("Run condition failed to validate parameter")]
+    RunCondition,
+    /// An observer failed to validate a parameter.
+    #[error("Observer failed to validate parameter")]
+    Observer,
+}
+
```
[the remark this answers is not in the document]
MiniaczQ: This should also contain system identifier if available

crates/bevy_ecs/src/schedule/executor/multi_threaded.rs
```diff
@@ -582,6 +583,14 @@ impl ExecutorState {
             // - `update_archetype_component_access` has been called for system.
             let valid_params = unsafe { system.validate_param_unsafe(world) };
             if !valid_params {
+                error_handler(
+                    SystemParamValidationError::System.into(),
+                    ErrorContext::System {
+                        name: system.name(),
+                        last_run: system.get_last_run(),
+                    },
+                );
```
MiniaczQ, line 592: It seems `ErrorContext` is enough to tell whether we're working with a system/observer, etc. do we need to duplicate this in `SystemParamValidationError`?

MiniaczQ:
A2:
I'm not sure I understand what "bubbling" would imply, is this for cases where a system runs another system?

A3:
I'm strongly opposed to per-system configuration, it's too granular and not granular enough at the same time.
A mix of global handling + per param handling would be ideal, and I agree that it's a feature in itself and belongs in a separate PR.

NthTensor:
Re: Warn once. I agreed that a stateful error handler would be smart. But currently the default error handler is just panicking, and if users want more complex behavior they have to write it themselves. I'd like to see what people do with the feature before we add other pre-made handlers. I I think that's well out of scope for this pr.

crates/bevy_ecs/src/system/system_param.rs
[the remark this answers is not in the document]
NthTensor: This should probably get piped through to `RegisteredSystemError`. That's a nit, as the `run_system` api probably needs to be refactored a tad now that we can return results from systems. I think right now it will give you a `Result<Result<O, E>, RegisteredSystemError>` when it really should be more monadic.

Let's not worry about that now.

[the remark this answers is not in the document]
alice-i-cecile: I would really like to add the name of the param here, but doing so will be substantially more complex. So, follow-up ideally :)

alice-i-cecile:
Alright, thanks for the feedback! I'm going to do minimal cleanup on this on Sunday, get it merged, and then see if I can't get per `SystemParam` panicking done without delaying 0.16 too bad. I would *really* like to get this to a good state so folks can happily use `Single` et al again.

MiniaczQ:
A possible strategy for per-param configuration are generic wrappers which change the error type to something that handlers are less sensitive to.
So `Single<Foo>` may return `InvalidSystemParam`, but `Cond<Single<Foo>>` will map `InvalidSystemParam -> PreventSystemExecution`.
I'm sure something like that already came up in a discord discussion, I'm more concerned about error-handler composition, so it's easier to standardize certain errors without rewriting handling of everything.

examples/ecs/fallible_params.rs
```diff
@@ -20,20 +31,13 @@ fn main() {
     App::new()
         .add_plugins(DefaultPlugins)
         .add_systems(Startup, setup)
-        // Default system policy is to panic if parameters fail to be fetched.
-        // We overwrite that configuration, to either warn us once or never.
+        //
+        // However, we can overwrite that configuration, to either warn us once or never.
         // This is good for catching unexpected behavior without crashing the app,
         // but can lead to spam.
```
MiniaczQ, line 37: We overwrite the error handling earlier, this comment is out of place

```suggestion
```

MiniaczQ (approved):
Looks good :)

alice-i-cecile:
I don't understand what's going on with https://github.com/bevyengine/bevy/actions/runs/14024499144/job/39260883882?pr=18454, which seems to be failing due to #18301. I also don't know why this is only failing on this PR.

It seems to be Mac-specific, so I can't reproduce it locally.

NthTensor:
Cloned and ran `cargo run -p ci -- test` on my Mac, could not reproduce compile error.

crates/bevy_ecs/src/schedule/executor/simple.rs
```diff
@@ -88,6 +89,16 @@ impl SystemExecutor for SimpleExecutor {
             let system = &mut schedule.systems[system_index];
             if should_run {
                 let valid_params = system.validate_param(world);
+                if !valid_params {
+                    error_handler(
+                        SystemParamValidationError.into(),
+                        ErrorContext::System {
```
NthTensor, line 95: Should this be `ErrorContext::RunCondition`?

NthTensor: Ah no, never mind.

NthTensor (approved):
Did a more thorough review; This is really nice and clean. A lot more to do with run conditions than I would have expected.

To me, this finally feels like the canonically right approach.

PR #18456, opened 2025-03-21
unify cursor lock behaviour

robtfm:
# Objective

unify the behaviour of `CursorGrabMode::Locked` across windows and macos/linux.

currently when the grab mode is changed to `Locked`, on windows only, the cursor is just constrained.
when the cursor is both locked and invisible, it randomly moves to the middle of the window. this changes the `Interaction` results and `cursor_position()` behaviour.

this pr changes windows behaviour to be inline with macos/linux on a best-effort basis
- reset the hardware cursor position each frame (results in jittery appearance but better than nothing)
- leaves the internal cursor at its original location for `Interaction`s and `cursor_position()`s
- avoids the random jump to the middle when locked and invisible

## Solution

- in `window_event`, when a `CursorMoved` event is received and the cursor is `Locked`, don't update the internal cursor's physical position. instead, update only the backend position. this effectively disregards the move to the middle of the screen that winit sends when `CursorGrabMode::Locked` and `!visible` are set
- in `changed_windows`, 
  - update the grab mode and visibliity before attempting to update position (required for the position update to be accepted by the backend)
  - if the internal physical position doesn't match the backend position, update the backend with the internal physical position

## Testing

left click to lock, right click to hide
```rs
use bevy::{prelude::*, window::CursorGrabMode};

fn main() {
    App::new()
        .add_plugins(DefaultPlugins)
        .add_systems(Startup, setup)
        .add_systems(Update, update)
        .run();
}

    commands.spawn((
        Node {
            min_width: Val::Px(40.0),
            min_height: Val::Px(40.0),
            ..Default::default()
        },
        BackgroundColor(Color::linear_rgb(1.0, 0.0, 0.0)),
    ));
    
    commands.spawn((Text2d::new(""), TextColor(Color::WHITE)));

    commands.spawn(Camera2d::default());
}

fn update(
    b: Res<ButtonInput<MouseButton>>,
    mut w: Query<&mut Window>,
    mut n: Query<&mut Node>,
    mut t: Query<&mut Text2d>
) {
    let mut w = w.single_mut().unwrap();
    let mut text = "\n\n\n\n".to_owned();

    if b.pressed(MouseButton::Left) {
        if w.cursor_options.grab_mode == CursorGrabMode::None {
            w.cursor_options.grab_mode = CursorGrabMode::Locked;
        }
        text = format!("{text}L");
    } else {
        if w.cursor_options.grab_mode == CursorGrabMode::Locked {
            w.cursor_options.grab_mode = CursorGrabMode::None;
        }
    }

    if b.pressed(MouseButton::Right) {
        if w.cursor_options.visible {
            w.cursor_options.visible = false;
        }
        text = format!("{text}R");
    } else {
        if !w.cursor_options.visible {
            w.cursor_options.visible = true;
        }
    }

    if let Some(cursor_position) = w.cursor_position() {
        let mut node = n.single_mut().unwrap();
        node.left = Val::Px(cursor_position.x - 10.0);
        node.top = Val::Px(cursor_position.y - 10.0);
    }
    t.single_mut().unwrap().0 = text;
}
```
before:

https://github.com/user-attachments/assets/76822a91-e9cb-4b54-aeb6-57861a0f65dd

after: 

https://github.com/user-attachments/assets/d2d0a13d-d8d6-439c-a601-678950afcba0

### DIFF
--- a/crates/bevy_window/src/window.rs
+++ b/crates/bevy_window/src/window.rs
@@ -614,7 +614,7 @@ impl Window {
 
     /// Retrieves the cursor position as reported by the back end.
     ///
-    /// This can be different to the phyical cursor position on
+    /// This can be different to the physical cursor position on
     /// the windows platform when `CursorGrabMode::Locked` is set.
     pub fn backend_cursor_position(&self) -> Option<Vec2> {
         self.internal
--- a/crates/bevy_window/src/window.rs
+++ b/crates/bevy_window/src/window.rs
@@ -588,6 +588,18 @@ impl Window {
             .map(|position| (position.as_dvec2() / self.scale_factor() as f64).as_vec2())
     }
 
+    fn constrain_physical_position(&self, position: DVec2) -> Option<Vec2> {
+        if position.x >= 0.
+            && position.y >= 0.
+            && position.x < self.physical_width() as f64
+            && position.y < self.physical_height() as f64
+        {
+            Some(position.as_vec2())
+        } else {
+            None
+        }
+    }
+
     /// The cursor position in this window in physical pixels.
     ///
     /// Returns `None` if the cursor is outside the window area.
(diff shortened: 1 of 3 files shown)


robtfm:
the jump to the middle still occurs with `Confined && !visible`, not sure there's anything we can do about that. this is probably better fixed in winit directly. 

that's kind of incidental to the `Locked` behaviour though, which was the main point of the pr.

alice-i-cecile:
I'm very much in favor of making this more consistent: I constantly see users frustrated by this. Is this something that should be upstreamed to winit at some point?

robtfm:
yes, definitely better handled in winit, we can remove the jitter completely.

apparently this solution doesn't work well on macos anyway, so i'm closing this regardless.

robtfm:
https://github.com/rust-windowing/winit/pull/4176

PR #18458, opened 2025-03-21
`Projection` Improvements

aevyrie:
# Objective

- Remove a component impl footgun
- Make projection code slightly nicer, and remove the need to import the projection trait when using the methods on `Projection`.

## Solution

- Do the things.


crates/bevy_render/src/camera/projection.rs
```diff
@@ -201,6 +201,36 @@ impl CustomProjection {
     }
 }
 
+// TODO: remove when trait upcasting is stabilized.
```
alice-i-cecile, line 204: Trait upcasting is merged into Rust, but not *quite* released. See https://github.com/rust-lang/rust/issues/65991/. Does it make sense to wait until April 3rd? https://releases.rs/docs/1.86.0/

atlv24 (approved):

alice-i-cecile (approved):

PR #18465, opened 2025-03-21
Restructure morph target pipeline to reduce crate dependencies

greeble-dev:
## Objective

- Restructure the morph target runtime pipeline to avoid copying weights between entities.
- The main goal is to make `bevy_animation` less dependent on `bevy_render`.
- It's also a small memory and CPU optimisation, and makes the pipeline more flexible.
- Will require migration guide.

## Why a Draft?

The PR is incomplete. The code works but I haven't updated documentation comments or done a final test and polish. I'm hoping to surface any design disagreements before I finish up. But I can also finish it now if preferred.

## Summary

I've been looking at making `bevy_animation` independent of `bevy_render`. I'm assuming this is good for compile times and alternative renderers.

If #18437 lands then there's only one remaining dependency - a system ordering that makes sure morph target weights are written by `bevy_animation` before the renderer reads them.

```rust
animate_targets.before(bevy_render::mesh::inherit_weights)
```

This PR removes the dependency in a slightly drastic way, by restructuring the morph target pipeline to eliminate `inherit_weights`. As well as removing the `bevy_render` dependency, it's a small optimisation and makes the pipeline more flexible.

The downside is that some users might need to update their code if they were manipulating the morph target pipeline directly. Users who were just importing glTF scenes are unaffected.

The PR also adds a `many_morph_targets` stress test - I can make it a separate PR if preferred.

If this PR and #18437 land then I'll make another PR to fully remove the `bevy_animation` dependency on `bevy_render`.

## Changes

The current pipeline as it's typically used is:

- Some entity has a `MorphWeights` component, which is mostly a `Vec<f32>` of weights.
- `bevy_animation` or user code sets weights on the `MorphWeights` component.
- One or more child entities have `Mesh3d` and `MeshMorphWeights` components. `MeshMorphWeights` is a `Vec<f32>` of weights.
- `bevy_render::inherit_weights` iterates over all `MorphWeights` and copies the weights to any `MeshMorphWeights` in child entities.
- `bevy_render::extract_morphs` iterates over all `Mesh3d` with `MeshMorphWeights` and copies the weights to internal buffers.

The changes are:

- `MeshMorphWeights` is no longer a `Vec<f32>`. It's now a reference to the entity that contains `MorphWeights`.
- `inherit_weights` disappears as there's nothing to copy.
- `extract_morphs` now iterates over `MeshMorphWeights`, uses it to find `MorphWeights`, then copies the weights.

The new pipeline is more flexible, as a mesh can reference `MorphWeights` on any entity. The current pipeline requires `MorphWeights` to be on a parent of the mesh entity.

## Breakage

**Unaffected**: Users who import a glTF scene and drive the `MorphWeights` component themselves or via `bevy_animation`.

**Possibly Affected**: Users who set up their own pipeline or modify the glTF pipeline. In particular, if they're manipulating `MeshMorphWeights` directly then they'll have to restructure to use `MorphWeights`.

I did a github code search and found only [one case that would break](https://github.com/ricky26/bevy-idol/blob/7ca60767a7415b6c59de2d0949429ac4f669d91d/crates/bevy_idol/src/main.rs#L514). Their fix should be fairly simple - just rearranging the components.

There are a couple of options to soften the blow, but they all involve some complexity/performance trade-offs. I can write these options up if desired.

# Performance

Tested on `many_morph_targets`. Win10, AMD Zen4.

- `extract_morphs` goes from 42.51us to 48.86us (+6.35us), probably due to the extra entity lookup.
- `inherit_weights` goes from 30.28us to zero.
- Throughput of the two systems goes from 14 to 21 meshes per us (+50%).

I expect a small memory saving due to one less copy of the weights per mesh.

There's room for further optimisation. The render buffers have a copy of the weights for each mesh, but multiple meshes could be sharing weights (which corresponds to multiple `MeshMorphWeights` referencing a single `MorphWeights`).

# Testing

```
cargo run --example morph_targets
cargo run --example many_morph_targets
cargo run --example scene_viewer -- assets/models/animated/MorphStressTest.gltf
```

Also tested a few other glTFs, and hacked the glTF importer to simulate some valid and invalid combinations (missing weights, mismatched arrays).

# Migration Guide

TODO

### DIFF
--- a/crates/bevy_gltf/src/loader/mod.rs
+++ b/crates/bevy_gltf/src/loader/mod.rs
@@ -1387,8 +1387,6 @@ fn load_node(
     // Map node index to entity
     node_index_to_entity_map.insert(gltf_node.index(), node.id());
 
-    let mut morph_weights = None;
-
     node.with_children(|parent| {
         // Only include meshes in the output if they're set to be retained in the MAIN_WORLD and/or RENDER_WORLD by the load_meshes flag
         if !settings.load_meshes.is_empty() {
@@ -1413,6 +1411,7 @@ fn load_node(
                         primitive: primitive.index(),
                     };
                     let bounds = primitive.bounding_box();
+                    let parent_entity = parent.target_entity();
 
                     let mut mesh_entity = parent.spawn((
                         // TODO: handle missing label handle errors here?
@@ -1424,22 +1423,13 @@ fn load_node(
 
                     let target_count = primitive.morph_targets().len();
                     if target_count != 0 {
-                        let weights = match mesh.weights() {
-                            Some(weights) => weights.to_vec(),
-                            None => vec![0.0; target_count],
-                        };
-
-                        if morph_weights.is_none() {
-                            morph_weights = Some(weights.clone());
-                        }
-
-                        // unwrap: the parent's call to `MeshMorphWeights::new`
-                        // means this code doesn't run if it returns an `Err`.
-                        // According to https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#morph-targets
-                        // they should all have the same length.
+                        // TODO: Should this check that `target_count` equals
+                        // `mesh.weights.len()`? Only necessary to catch malformed assets.
+                        //
+                        // https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#morph-targets
                         // > All morph target accessors MUST have the same count as
                         // > the accessors of the original primitive.
-                        mesh_entity.insert(MeshMorphWeights::new(weights).unwrap());
+                        mesh_entity.insert(MeshMorphWeights(parent_entity));
                     }
                     mesh_entity.insert(Aabb::from_min_max(
                         Vec3::from_slice(&bounds.min),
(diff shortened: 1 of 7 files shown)


crates/bevy_mesh/src/morph.rs
```diff
@@ -142,8 +144,16 @@ impl MorphWeights {
     pub fn weights_mut(&mut self) -> &mut [f32] {
         &mut self.weights
     }
+    pub fn clear_weights(&mut self) {
+        self.weights.clear();
+    }
+    pub fn extend_weights(&mut self, weights: &[f32]) {
+        self.weights.extend(weights);
+    }
```
greeble-dev, line 152: These were added so that `MorphWeights` supports everything that `MeshMorphWeights` did.

Personally I'd rather make `MorphWeights` members public and avoid the need for this interface.

crates/bevy_gltf/src/loader/mod.rs
```diff
@@ -1572,14 +1560,32 @@ fn load_node(
...
-                primitive_label.map(|label| load_context.get_label_handle(label.to_string()));
-            node.insert(MorphWeights::new(weights, first_mesh)?);
+        if let Some(mesh) = gltf_node.mesh() {
+            // Create `MorphWeights`. The weights will be copied from `mesh.weights()`
+            // if present. If not then the weights are zero.
+            //
+            // The glTF spec says that all primitives must have the same number
+            // of morph targets, and `mesh.weights()` should be equal to that
+            // number if present. We're more forgiving and take whichever is
+            // biggest, leaving any unspecified weights at zero.
```
greeble-dev, line 1570: I played it safe here, to the point of accepting malformed glTFs. I wasn't sure if the importer is intended to be strict or permissive.

greeble-dev:
Changed from draft to ready for review. Documentation has been updated, and it now fully removes the `bevy_animation` dependency on `bevy_render`.

atlv24 (approved):

atlv24:
Removed Contentious because I dont see why it would be contentious

greeble-dev:
PR has been updated to fix conflicts with #20472.

greeble-dev:
Putting this back to draft status while I merge and simplify some things.

greeble-dev:
Ready for review again. Instead of making `MeshMorphWeights` a reference, it's now an enum that's either a value or a reference. This adds a little bit of complexity, but makes migration simpler and could be an optimisation in some cases.

Pixel Eagle is complaining, but I think that's a false alarm as it only seems to affect a text example that doesn't have any morph targets. Might be related to the parley migration.

PR #18470, opened 2025-03-21
Rename EntityBorrow/TrustedEntityBorrow to ContainsEntity/EntityEquivalent

### LINKED ISSUE #9367: `AsEntity` Trait

When working with Bevy, it is often desirable to store references to entities in types that represent those entities.
For example:
```rust
struct Parent(Entity);

struct Player {
    entity: Entity,
    ...
}
```

Currently, when trying to use these types for queries and other similar functions which take an `Entity` as input, the user must dereference this type as the entity:
```rust
let player_component = player_query.get(player.entity);
```
Often times, such as with the standard `Parent(Entity)` type, this syntax can be made less verbose by implement `Deref<Target = Entity>`:
```rust
let parent_component = parent_query.get(**parent);
```
Notice how this requires `**`, because usually the parent is given as a `&Parent`.

This can often create redundantly verbose code.
I propose that this alternative would be much cleaner and more readable:
```rust
let player_component = player_query.get(player);
let parent_component = parent_query.get(parent);
```

To solve this issue, we could introduce a new trait:
```rust
trait AsEntity {
   fn as_entity(&self) -> Entity;
}
```
Then we can implement this trait for types such as `Entity`, `Parent`, etc.
Additionally, the user can implement this trait for custom user data.

To make this work, all the top level functions that take entities as input would need to be refactored to take an `impl AsEntity` instead of `Entity`. If `AsEntity` is implemented for `Entity`, this should have no visible effect to the end user.
This would be functions like `query.get()` or `world.entity()`.

I'm not sure if a better name for this would be `IntoEntity` (considering how we have `IntoSystem`).
To me `AsEntity` makes more sense since we're usually trying to reference a thing "as an entity".

## Additional context

N/A

Victoronz:
# Objective

Fixes #9367.

Yet another follow-up to #16547.

These traits were initially based on `Borrow<Entity>` because that trait was what they were replacing, and it felt close enough in meaning.
However, they ultimately don't quite match: `borrow` always returns references, whereas `EntityBorrow` always returns a plain `Entity`. 
Additionally, `EntityBorrow` can imply that we are borrowing an `Entity` from the ECS, which is not what it does.

Due to its safety contract, `TrustedEntityBorrow` is important an important and widely used trait for `EntitySet` functionality.
In contrast, the safe `EntityBorrow` does not see much use, because even outside of `EntitySet`-related functionality, it is a better idea to accept `TrustedEntityBorrow` over `EntityBorrow`.

Furthermore, as #9367 points out, abstracting over returning `Entity` from pointers/structs that contain it can skip some ergonomic friction.

On top of that, there are aspects of #18319 and #18408 that are relevant to naming: 
We've run into the issue that relying on a type default can switch generic order. This is livable in some contexts, but unacceptable in others.

To remedy that, we'd need to switch to a type alias approach: 
The "defaulted" `Entity` case becomes a `UniqueEntity*`/`Entity*Map`/`Entity*Set` alias, and the base type receives a more general name. `TrustedEntityBorrow` does not mesh clearly with sensible base type names.

## Solution
Replace any `EntityBorrow` bounds with `TrustedEntityBorrow`.
+
Rename them as such:
`EntityBorrow` -> `ContainsEntity`
`TrustedEntityBorrow` -> `EntityEquivalent`

For `EntityBorrow` we produce a change in meaning; We designate it for types that aren't necessarily strict wrappers around `Entity` or some pointer to `Entity`, but rather any of the myriad of types that contain a single associated `Entity`. 
This pattern can already be seen in the common `entity`/`id` methods across the engine.
We do not mean for `ContainsEntity` to be a trait that abstracts input API (like how `AsRef<T>` is often used, f.e.), because eliding `entity()` would be too implicit in the general case. 

We prefix "Contains" to match the intuition of a struct with an `Entity` field, like some contain a `length` or `capacity`.
It gives the impression of structure, which avoids the implication of a relationship to the `ECS`.
`HasEntity` f.e. could be interpreted as "a currently live entity", 

As an input trait for APIs like #9367 envisioned, `TrustedEntityBorrow` is a better fit, because it *does* restrict itself to strict wrappers and pointers. Which is why we replace any `EntityBorrow`/`ContainsEntity` bounds with `TrustedEntityBorrow`/`EntityEquivalent`.

Here, the name `EntityEquivalent` is a lot closer to its actual meaning, which is "A type that is both equivalent to an `Entity`, and forms the same total order when compared".
Prior art for this is the [`Equivalent`](https://docs.rs/hashbrown/latest/hashbrown/trait.Equivalent.html) trait in `hashbrown`, which utilizes both `Borrow` and `Eq` for its one blanket impl!

Given that we lose the `Borrow` moniker, and `Equivalent` can carry various meanings, we expand on the safety comment of `EntityEquivalent` somewhat. That should help prevent the confusion we saw in [#18408](https://github.com/bevyengine/bevy/pull/18408#issuecomment-2742094176).

The new name meshes a lot better with the type aliasing approach in #18408, by aligning with the base name `EntityEquivalentHashMap`.
For a consistent scheme among all set types, we can use this scheme for the `UniqueEntity*` wrapper types as well!
This allows us to undo the switched generic order that was introduced to `UniqueEntityArray` by its `Entity` default. 

Even without the type aliases, I think these renames are worth doing!

## Migration Guide

Any use of `EntityBorrow` becomes `ContainsEntity`.
Any use of `TrustedEntityBorrow` becomes `EntityEquivalent`.


crates/bevy_ecs/src/entity/entity_set.rs
```diff
@@ -17,46 +17,56 @@ use super::{unique_slice::UniqueEntitySlice, Entity};
...
+///
+/// It should only be implemented when:
+/// - Retrieving the [`Entity`] is a simple operation.
+/// - The [`Entity`] contained by the type is unambiguous.
+pub trait ContainsEntity {
+    /// Returns the contained entity.
     fn entity(&self) -> Entity;
 }
 
 /// A trait for [`Entity`] borrows with trustworthy comparison behavior.
```
Guvante, line 32: It seems that this phrasing should be updated since `EntityBorrow` had its nearly identical line changed.

crates/bevy_ecs/src/entity/entity_set.rs
```diff
@@ -17,46 +17,56 @@ use super::{unique_slice::UniqueEntitySlice, Entity};
...
-/// Comparison trait behavior between a [`TrustedEntityBorrow`] type and its underlying entity will match.
+/// Comparison trait behavior between an [`EntityEquivalent`] type and its underlying entity will match.
 /// This property includes [`PartialEq`], [`Eq`], [`PartialOrd`], [`Ord`] and [`Hash`],
 /// and remains even after [`Clone`] and/or [`Borrow`] calls.
 ///
 /// # Safety
-/// Any [`PartialEq`], [`Eq`], [`PartialOrd`], [`Ord`], and [`Hash`] impls must be
-/// equivalent for `Self` and its underlying entity:
+/// Any [`PartialEq`], [`Eq`], [`PartialOrd`], [`Ord`], and [`Hash`] impls must form a total order
+/// that matches that of `Entity`.
```
Guvante, line 40: I think should should be adjusted to preserve the original "underlying entity" or similar phrasing. Although the following line clarifies so maybe this is fine.

crates/bevy_ecs/src/query/iter.rs
```diff
@@ -1117,7 +1117,8 @@ impl<'w, 's, D: QueryData, F: QueryFilter, I: Iterator<Item = Entity>> Debug
 /// Entities that don't match the query are skipped.
 ///
 /// This struct is created by the [`Query::iter_many`](crate::system::Query::iter_many) and [`Query::iter_many_mut`](crate::system::Query::iter_many_mut) methods.
-pub struct QueryManyIter<'w, 's, D: QueryData, F: QueryFilter, I: Iterator<Item: EntityBorrow>> {
+pub struct QueryManyIter<'w, 's, D: QueryData, F: QueryFilter, I: Iterator<Item: EntityEquivalent>>
```
Guvante, line 1120: I didn't see this called out in the description (hard to tell from this if the tightening here is important)

Guvante (approved):

crates/bevy_ecs/src/entity/entity_set.rs
Victoronz: You're right, I'd missed this!

crates/bevy_ecs/src/entity/entity_set.rs
Victoronz: Hmm, I tried to convey two things:
That an instance of the type should match an instance of `Entity`, and that the types as a whole should form a total order.

But I think I can make it clearer by switching the two around and adjusting the wording!

Victoronz: Thinking on it a bit more, I think the issue here was that `PartialEq/`Eq`/`PartialOrd`/`Ord` are binary operations, while `Hash` is a unary operation!
We should be able to simplify this section again by moving `Hash` done to the other unary operations.

crates/bevy_ecs/src/query/iter.rs
Victoronz: I touched on it with "We should only accept `EntityEquivalent` in our APIs".

This PR represent a loosening of `EntityBorrow`/`ContainsEntity`, which means it is now too implicit to accept in our API.
By just accepting `EntityEquivalent`, we have a strong guarantee that only types close to `Entity` get passed to our code, and that we do not see strange behavior in `entity()` calls.

chescock (approved):
Looks good!  `EntityEquivalent` does seem more clear than `TrustedEntityBorrow`, especially as there is no longer any borrowing involved!

chescock:
What's the purpose of the `ContainsEntity` trait?  If we don't want to abstract over it in APIs, then it doesn't seem to offer any advantage over a convention of having an `entity()` method.  

If we aren't going to use it as a bound in `iter_many` then we might as well get rid of the separate trait and move the `entity()` method down to `EntityEquivalent`.  

> For a consistent scheme among all set types, we can use this scheme for the `UniqueEntity*` wrapper types as well!
> This allows us to undo the switched generic order that was introduced to `UniqueEntityArray` by its `Entity` default.

Your plan here is `struct UniqueEntityEquivalentArray<T, N>` and `type UniqueEntityArray<N> = UniqueEntityEquivalentArray<Entity, N>`, then?

crates/bevy_ecs/src/query/iter.rs
chescock: Yeah, I wasn't expecting that change, either.  It seems reasonable!  It's just hard to notice amid the similar non-functional changes.  It might help to mention explicitly in the first few lines of the PR description that you're changing the bound on `Query::iter_many` and friends.

Victoronz: Fair enough, I'll add this to the PR description!

Victoronz:
> What's the purpose of the `ContainsEntity` trait? If we don't want to abstract over it in APIs, then it doesn't seem to offer any advantage over a convention of having an `entity()` method.
> 
> If we aren't going to use it as a bound in `iter_many` then we might as well get rid of the separate trait and move the `entity()` method down to `EntityEquivalent`.

Not being a bound doesn't mean that it isn't useful! The intent is that an `entity()` call should only be implicit if a type is close enough to `Entity`. If it isn't, the intent is for the user to then call `entity()` themselves.

It is a common pattern for types that have an `Entity` associated with them to implement an `entity` or `id` method.
But currently, there is no standard or consistent meaning for such methods, and is sometimes even omitted. What this trait does is provide a standard abstraction over that pattern!
When I myself was learning Bevy, I struggled a lot with knowing whether you could get an `Entity` from a struct, and wished there was a way of navigating to those that had such API.
In that sense `ContainsEntity` is a searchable, standardized marker, and can be made derivable later on! 

> > For a consistent scheme among all set types, we can use this scheme for the `UniqueEntity*` wrapper types as well!
> > This allows us to undo the switched generic order that was introduced to `UniqueEntityArray` by its `Entity` default.
> 
> Your plan here is `struct UniqueEntityEquivalentArray<T, N>` and `type UniqueEntityArray<N> = UniqueEntityEquivalentArray<Entity, N>`, then?

Yeah!

alice-i-cecile:
Please bother me when this is finished and I'll merge it for you :)

crates/bevy_ecs/src/entity/entity_set.rs
```diff
@@ -17,46 +17,56 @@ use super::{unique_slice::UniqueEntitySlice, Entity};
...
-/// equivalent for `Self` and its underlying entity:
+/// Any [`PartialEq`], [`Eq`], [`PartialOrd`], and [`Ord`] impls must form a total order
+/// matching that of the underlying `Entity`:
 /// `x.entity() == y.entity()` should give the same result as `x == y`.
-/// The above equivalence must also hold through and between calls to any [`Clone`]
-/// and [`Borrow`]/[`BorrowMut`] impls in place of [`entity()`].
+///
+/// The above equivalence must also hold through and between calls to any [`Clone`] and
+/// [`Borrow`]/[`BorrowMut`] impls in place of [`entity()`]. This is also required of any [`Hash`] impl,
+/// but only when [`Hash::hash`] is called with a deterministic [`Hasher`].
```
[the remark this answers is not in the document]
Victoronz: I changed the requirement to "must delegate `Hash`". And added an explanation concerning its use!

Victoronz:
@alice-i-cecile Finished!

PR #18471, opened 2025-03-21
Implement `RelationshipSourceCollection` for `IndexSet`

Brezak:
# Objective

`IndexSet` doesn't implement `RelationshipSourceCollection`

## Solution

Implement `MapEntities` for `IndexSet`
Implement `RelationshipSourceCollection` for `IndexSet`

## Testing

`cargo clippy`


chescock (approved):

crates/bevy_ecs/src/relationship/relationship_source_collection.rs
```diff
@@ -277,6 +280,97 @@ impl RelationshipSourceCollection for Entity {
...
+
+impl RelationshipSourceCollection for EntityIndexSet {
+    type SourceIter<'a> = core::iter::Copied<crate::entity::index_set::Iter<'a>>;
+
+    fn new() -> Self {
+        EntityIndexSet::new()
+    }
+
+    fn reserve(&mut self, additional: usize) {
+        self.0.reserve(additional);
```
Victoronz, line 338: This trait impl should not rely on crate-public fields, but use an explicit `deref_mut()` call instead!

crates/bevy_ecs/src/entity/map_entities.rs
```diff
@@ -48,7 +51,7 @@ use smallvec::SmallVec;
 pub trait MapEntities {
     /// Updates all [`Entity`] references stored inside using `entity_mapper`.
     ///
-    /// Implementors should look up any and all [`Entity`] values stored within `self` and
```
Victoronz, line 51: I wonder, what exactly lints this correction? It is not a typo, both are correct, and the "o" version is more common in this context I believe

Victoronz (approved):

Victoronz:
small nit, the various set types are re-exported in the `entity` module, so no need to specify their specific submodule in their import!

crates/bevy_ecs/src/entity/map_entities.rs
Brezak: I'm using this extensions https://marketplace.visualstudio.com/items/?itemName=tekumara.typos-vscode. Which supposedly uses https://github.com/crate-ci/typos.

Victoronz: ah I see, then it seems that the vscode extension hasn't updated its `crate-ci` version yet, the newer versions of which no longer lint "implementor".

crates/bevy_ecs/src/relationship/relationship_source_collection.rs
```diff
@@ -557,7 +688,7 @@ mod tests {
         #[derive(Component)]
         #[relationship_target(relationship = Above)]
         struct Below(Entity);
-
+
```
mockersf, line 691:
```suggestion

```
messed up during conflict resolution

mockersf (approved):

crates/bevy_ecs/src/relationship/relationship_source_collection.rs
```diff
@@ -548,6 +645,40 @@ mod tests {
...
+        world.entity_mut(c).despawn();
+
+        let rel_target = world.get::<RelTarget>(a).unwrap();
+        let collection = rel_target.collection();
+
+        // Removals should maintain ordering
+        assert!(collection.iter().eq(&[b, d]));
+    }
+
+    #[test]
```
mockersf, line 681:
```suggestion
    #[test]
```
more conflict resolution issues...

crates/bevy_ecs/src/relationship/relationship_source_collection.rs
```diff
@@ -1,7 +1,12 @@
+use core::{
+    hash::BuildHasher,
+    ops::{Deref, DerefMut},
+};
 use alloc::collections::{btree_set, BTreeSet};
```
mockersf, line 5:
```suggestion
use alloc::collections::{btree_set, BTreeSet};
use core::{
    hash::BuildHasher,
    ops::{Deref, DerefMut},
};
```

PR #18474, opened 2025-03-22
Update to Glam 0.30.1 and adopt `nostd-libm` feature

### LINKED ISSUE #18397: `bevy_input` is now optional, but there's no way to enable it
## Bevy version

main, since #17955

## What you did

Attempted to migrate a crate dealing with input and using `default-features = false` to Bevy `0.16-dev`

## What went wrong

```
error[E0433]: failed to resolve: could not find `input` in `bevy`
  --> src/lib.rs:28:5
   |
28 |     input::keyboard::{Key, KeyboardInput},
   |     ^^^^^ could not find `input` in `bevy`
```

And there doesn't appear to be any way of rectifying this. It seems that we need a new `bevy_input` feature, but it's also possible that `bevy_input` becoming optional was a mistake. I haven't been following the `no_std` effort closely, so I'm not sure.

A bad workaround is to add the `std` feature, which includes `bevy_input`. (This may be a separate bug. Seems like with `bevy_input` optional, `bevy_internal/std` should be using `bevy_input?/std`).

bushrat011899:
# Objective

- Fixes #18397
- Simplify feature-gating in `bevy`

## Solution

- Update to Glam 0.30.1 which includes a new [`nostd-libm`](https://github.com/bitshifter/glam-rs/pull/617) feature
- Expose and offer a similar feature in `bevy_math`
- Enable `bevy_math/nostd-libm` in `bevy_internal`
- Remove feature gates on `bevy_math`, `bevy_transform`, and `bevy_input` in `bevy_internal` (reverting to pre-0.16 behaviour)

## Testing

- CI

---

## Notes

This causes `bevy` to always include `libm` as a dependency, but it is a very small and high quality one from the Rust language team themselves. Crucially, the `libm` and `std` features can both override the use of `nostd-libm`, still giving users total control over which implementation they wish to use in both Glam and `bevy_math`. I believe the singular forced dependency is worth it for the increased functionality that `default-features = false` Bevy can provide.

For users who cannot accept that dependency, you can depend directly on Bevy's sub-crates instead, which is generally recommended for enthusiast users anyway.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -546,7 +546,7 @@ tracing = { version = "0.1", default-features = false, optional = true }
 bevy_dylib = { path = "crates/bevy_dylib", version = "0.16.0-dev", default-features = false, optional = true }
 
 [dev-dependencies]
-rand = "0.8.0"
+rand = "0.9.0"
 rand_chacha = "0.3.1"
 ron = "0.8.0"
 flate2 = "1.0"
--- a/benches/Cargo.toml
+++ b/benches/Cargo.toml
@@ -30,7 +30,7 @@ bevy_platform_support = { path = "../crates/bevy_platform_support", default-feat
 
 # Other crates
 glam = "0.30.1"
-rand = "0.8"
+rand = "0.9"
 rand_chacha = "0.3"
 
 # Make `bevy_render` compile on Linux with x11 windowing. x11 vs. Wayland does not matter here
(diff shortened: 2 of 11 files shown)


andriyDev (changes requested):
Looks good, but CI is dead...

bushrat011899:
This has significant overlap with #18047 (I should've done more research before opening this PR!) I'll coordinate on that PR first and see if the possible blocker on encase can be resolved. If not, this may need to wait for 0.17

mockersf:
probably not for the 0.16, as updating glam comes with other crates to update that are not ready

bushrat011899:
Since there are two alternate approaches (either upgrade to 0.30 and _then_ add this feature, or get Glam to release a 0.29.3 with the feature backported) I'm going to close this PR while we see which option presents itself!

PR #18476, opened 2025-03-22
Regression fix: Reintroduce sorting/reordering methods on Children

inodentry:
# Objective

Bevy 0.15 used to have methods on `Children` for sorting and reordering them. This is very important, because in certain situations, the order of children matters. For example, in the context of UI nodes.

These methods are missing/omitted/forgotten in the current version, after the Relationships rework.

Without them, it is impossible for me to upgrade `iyes_perf_ui` to Bevy 0.16.

## Solution

Reintroduce the methods. This PR simply copy-pastes them from Bevy 0.15.

### DIFF
--- a/crates/bevy_ecs/src/hierarchy.rs
+++ b/crates/bevy_ecs/src/hierarchy.rs
@@ -139,6 +139,96 @@ impl FromWorld for ChildOf {
 #[doc(alias = "IsParent")]
 pub struct Children(Vec<Entity>);
 
+impl Children {
+    /// Swaps the child at `a_index` with the child at `b_index`.
+    #[inline]
+    pub fn swap(&mut self, a_index: usize, b_index: usize) {
+        self.0.swap(a_index, b_index);
+    }
+
+    /// Sorts children [stably](https://en.wikipedia.org/wiki/Sorting_algorithm#Stability)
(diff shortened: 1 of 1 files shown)


Victoronz:
What about `sort`/`sort_unstable`?
Sidenote: Sorting `Entity` directly does not benefit from stability, so `sort` could redirect to `sort_unstable`.

inodentry:
Those methods did not exist in the old version either. And it makes sense to not have them, because they are meaningless.

The exact values stored inside `Entity` are internal implementation details of the ECS and are not meaningful to the user. There is no guarantee what they might be. `Entity` is semantically an opaque ID.  Sorting it is nonsensical.

chescock (approved):

Victoronz:
> Those methods did not exist in the old version either. And it makes sense to not have them, because they are meaningless.
> 
> The exact values stored inside `Entity` are internal implementation details of the ECS and are not meaningful to the user. There is no guarantee what they might be. `Entity` is semantically an opaque ID. Sorting it is nonsensical.

An order does not need to be meaningful to be useful! One may just want to ensure consistency in some way.

But that is not the responsibility of this PR, it is just a regression fix.

Victoronz (approved):

Issue #18104, opened 2025-03-01
Black materials on `clearcoat` example


hukasu:
## Bevy version

Since #17898

## \[Optional\] Relevant system information

```
2025-03-01T13:29:03.153201Z  INFO bevy_diagnostic::system_information_diagnostics_plugin::internal: SystemInfo { os: "Linux (NixOS 24.11)", kernel: "6.6.69", cpu: "Intel(R) Core(TM) i7-14700KF", core_count: "20", memory: "31.2 GiB" }    
2025-03-01T13:29:03.321799Z  INFO bevy_render::renderer: AdapterInfo { name: "NVIDIA GeForce RTX 4060 Ti", vendor: 4318, device: 10243, device_type: DiscreteGpu, driver: "NVIDIA", driver_info: "565.77", backend: Vulkan }
```

## What you did

Ran `clearcoat` example

## What went wrong

3 of the sphere are completely black

## Additional information

![Image](https://github.com/user-attachments/assets/99fcda43-f8fa-4ad2-af6f-927303f12031)

pcwalton:
Hmm, I can't reproduce this.

hukasu:
some of the prs i opened seems to be linux exclusive

https://thebevyflock.github.io/bevy-example-runner/

hukasu:
full log for `cargo run --example clearcoat --features="pbr_multi_layer_material_textures wayland"`

```
2025-03-03T01:07:00.978711Z  INFO bevy_diagnostic::system_information_diagnostics_plugin::internal: SystemInfo { os: "Linux (NixOS 24.11)", kernel: "6.6.69", cpu: "Intel(R) Core(TM) i7-14700KF", core_count: "20", memory: "31.2 GiB" }    
2025-03-03T01:07:01.103290Z ERROR wgpu_hal::gles::egl: EGL 'eglCreateSyncKHR' code 0x3004: EGL_BAD_ATTRIBUTE error: In eglCreateSyncKHR: EGL_SYNC_NATIVE_FENCE_FD_ANDROID specified valid fd butEGL_SYNC_STATUS is also being set
    
2025-03-03T01:07:01.174318Z  INFO bevy_render::renderer: AdapterInfo { name: "NVIDIA GeForce RTX 4060 Ti", vendor: 4318, device: 10243, device_type: DiscreteGpu, driver: "NVIDIA", driver_info: "565.77", backend: Vulkan }
[...]

PR #18489, opened 2025-03-22
Move the construction of `LightingInput` in `apply_pbr_lighting` into its own function.

13ros27:
# Objective
Reduce duplication in the `bevy_pbr` shaders by using a function to create `LightingInput` from `PbrInput` which can then be shared between `apply_pbr_lighting` and `ssr::fragment`.

## Solution
I created a function `pbr_input_to_lighting_input` in `pbr_lighting` to do this (lighting seemed the most logical place for it as that is where `LightingInput` lives and in a higher level language this would be a constructor). This also required moving `calculate_F0` and `calculate_diffuse_color` into `pbr_lighting` to prevent a circular dependency (infinite load during shader compilation), this makes sense for `calculate_F0` as this is now the only thing that uses it but `calculate_diffuse_color` does still have another user file.

## Questions for reviewers

- Should `pbr_input_to_lighting_input` instead accept a `ptr` to `PbrInput`, I don't feel like I have a sufficient grasp of shader optimisers to know whether this is better (or if we can trust this should always be inlined)?
- Should it instead live in `pbr_functions` which has the advantage of making this a non-breaking change (because no functions would have to move) but separates a constructor from its struct definition?

## Testing

I ran `ssr` and `anisotropy` and checked them for visual issues/crashes and ran as much CI as I could before running out of disk space. I tried to test `clearcoat` but that example is broken for me (#18104)

## Migration Guide

Shader function changes:
  - `calculate_diffuse_color` and `calculate_F0` have been moved from `bevy_pbr::pbr_functions` to `bevy_pbr::lighting`.

### DIFF
--- a/crates/bevy_pbr/src/deferred/pbr_deferred_functions.wgsl
+++ b/crates/bevy_pbr/src/deferred/pbr_deferred_functions.wgsl
@@ -4,6 +4,7 @@
     pbr_types::{PbrInput, pbr_input_new, STANDARD_MATERIAL_FLAGS_UNLIT_BIT},
     pbr_deferred_types as deferred_types,
     pbr_functions,
+    lighting,
     rgb9e5,
     mesh_view_bindings::view,
     utils::{octahedral_encode, octahedral_decode},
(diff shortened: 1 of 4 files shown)


IceSentry:
I'm generally in favor of this change, but have you ran it through a gpu profiler? Using more functions can increase register pressure and lead to bad performance.

JMS55:
(Modern drivers tend to use LLVM and will inline functions, but not always, and this is very perf-sensitive code)

PR #18490, opened 2025-03-22
Tracy GPU support

JMS55:
# Objective

- Add tracy GPU support

## Solution

- Build on top of the existing render diagnostics recording to also upload gpu timestamps to tracy
- Copy code from https://github.com/Wumpf/wgpu-profiler

## Showcase
image

### DIFF
--- a/crates/bevy_pbr/src/render/light.rs
+++ b/crates/bevy_pbr/src/render/light.rs
@@ -2238,18 +2238,12 @@ impl ShadowPassNode {
         world: &'w World,
         is_late: bool,
     ) -> Result<(), NodeRunError> {
-        let diagnostics = render_context.diagnostic_recorder();
-
-        let view_entity = graph.view_entity();
-
         let Some(shadow_render_phases) = world.get_resource::<ViewBinnedRenderPhases<Shadow>>()
         else {
             return Ok(());
         };
 
-        let time_span = diagnostics.time_span(render_context.command_encoder(), "shadows");
-
-        if let Ok(view_lights) = self.main_view_query.get_manual(world, view_entity) {
+        if let Ok(view_lights) = self.main_view_query.get_manual(world, graph.view_entity()) {
             for view_light_entity in view_lights.lights.iter().copied() {
                 let Ok((view_light, extracted_light_view, occlusion_culling)) =
                     self.view_light_query.get_manual(world, view_light_entity)
@@ -2305,8 +2299,6 @@ impl ShadowPassNode {
                 });
             }
         }
-
-        time_span.end(render_context.command_encoder());
 
         Ok(())
     }
--- a/crates/bevy_render/src/diagnostic/internal.rs
+++ b/crates/bevy_render/src/diagnostic/internal.rs
@@ -53,7 +53,7 @@ impl DiagnosticsRecorder {
         #[cfg(feature = "tracing-tracy")]
         let tracy_gpu_context =
             super::tracy_gpu::new_tracy_gpu_context(adapter_info, device, queue);
-        let _ = adapter_info;
+        let _ = adapter_info; // Prevent unused variable warnings when tracing-tracy is not enabled
 
         DiagnosticsRecorder(WgpuWrapper::new(DiagnosticsRecorderInternal {
             timestamp_period_ns: queue.get_timestamp_period(),
--- a/crates/bevy_render/src/diagnostic/internal.rs
+++ b/crates/bevy_render/src/diagnostic/internal.rs
@@ -57,7 +57,12 @@ impl DiagnosticsRecorder {
         DiagnosticsRecorder(WgpuWrapper::new(DiagnosticsRecorderInternal {
             timestamp_period_ns: queue.get_timestamp_period(),
             features,
-            current_frame: Mutex::new(FrameData::new(device, features, tracy_gpu_context.clone())),
+            current_frame: Mutex::new(FrameData::new(
+                device,
+                features,
+                #[cfg(feature = "tracing-tracy")]
+                tracy_gpu_context.clone(),
+            )),
             submitted_frames: Vec::new(),
             finished_frames: Vec::new(),
             #[cfg(feature = "tracing-tracy")]
@@ -356,6 +361,7 @@ impl FrameData {
             begin_instant: None,
             end_instant: None,
             pipeline_statistics_index: None,
+            #[cfg(feature = "tracing-tracy")]
             tracy_gpu_span,
         });
 
@@ -372,6 +378,7 @@ impl FrameData {
             .next_back()
             .unwrap();
 
+        #[allow(unused_mut)]
         let mut span = self.open_spans.swap_remove(index);
 
         #[cfg(feature = "tracing-tracy")]
--- a/crates/bevy_render/src/diagnostic/internal.rs
+++ b/crates/bevy_render/src/diagnostic/internal.rs
@@ -12,10 +12,10 @@ use bevy_platform_support::time::Instant;
 use std::sync::Mutex;
 use wgpu::{
     Buffer, BufferDescriptor, BufferUsages, CommandEncoder, ComputePass, Features, MapMode,
-    PipelineStatisticsTypes, QuerySet, QuerySetDescriptor, QueryType, Queue, RenderPass,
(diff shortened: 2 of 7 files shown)


JMS55:
This is frustrating. I _do_ want to use allow here, not expect. Clippy is being dumb :/

JMS55:
I'm also not convinced that this is working. Idk why only the shadow/bloom spans show up, MainOpaquePass3d should as well.

Also it seems like it's showing a previous frame's GPU spans? Maybe because of buffer mapping delay? Not exactly sure how you're supposed to correlate things, it might be wrong.

JMS55:
I think this is fine actually. The transparent pass shows up. As for the delay, that's just because we don't submit until the end of the frame, so the data won't be available until next frame at least.

JMS55:
I think some of the weirdness might also be because of how we open the diagnostic time span before the command buffer generation task. It should really be a part of the closure...

JMS55:
Ok I found out where the main opaque pass went 
image

JMS55:
Ok so the problem is that the global ordering of open/close spans needs to be correct across multiple threads.

You can't do:
1. open (t1)
2. open (t2)
3. close (t1) // this actually closes t2's span, not t1's

So if multi_threading is disabled in bevy, this PR works fine. Otherwise it's completely broken.

JMS55:
Ok, we're all good now! It's now actually ready to review,

bushrat011899 (approved):
Excellent to get more debugging tools! I had some trouble getting this to work initially due to at first not adding the `RenderDiagnosticsPlugin`, and then due to odd zooming rules in Tracy itself. As pointed out by @JMS55, scrolling to the space between frames helps with "finding" the `RenderQueue` track.

crates/bevy_render/src/diagnostic/internal.rs
```diff
@@ -41,21 +43,31 @@ pub struct DiagnosticsRecorder(WgpuWrapper<DiagnosticsRecorderInternal>);
...
 
-        let timestamp_period_ns = if features.contains(Features::TIMESTAMP_QUERY) {
-            queue.get_timestamp_period()
-        } else {
-            0.0
-        };
+        #[cfg(feature = "tracing-tracy")]
+        let tracy_gpu_context =
+            super::tracy_gpu::new_tracy_gpu_context(adapter_info, device, queue);
+        let _ = adapter_info;
```
superdump, line 56: What is this line here for?

crates/bevy_render/src/diagnostic/tracy_gpu.rs
```diff
@@ -0,0 +1,67 @@
...
+pub fn new_tracy_gpu_context(
+    adapter_info: &RenderAdapterInfo,
+    device: &RenderDevice,
+    queue: &RenderQueue,
+) -> GpuContext {
+    let tracy_gpu_backend = match adapter_info.backend {
+        Backend::Vulkan => GpuContextType::Vulkan,
+        Backend::Dx12 => GpuContextType::Direct3D12,
+        Backend::Gl => GpuContextType::OpenGL,
+        Backend::Metal | Backend::BrowserWebGpu | Backend::Empty => GpuContextType::Invalid,
```
superdump, line 17: Noooooo it doesn’t work on metal? Whyyyyy :(

docs/profiling.md
```diff
@@ -135,15 +147,18 @@ For profiling GPU work, you should use the tool corresponding to your GPU's vend
...
+While it doesn't provide as much detail as vendor-specific tooling, Tracy can also be used to coarsely measure GPU performance.
 
-Specifically, you record a list of tasks (commands) for the GPU to perform into a CommandBuffer, and then submit that on a Queue to the GPU. At some point in the future, the GPU will receive the commands and execute them.
+Add the `RenderDiagnosticsPlugin` to your app, and then compile with Bevy's `trace_tracy` feature. GPU spans will show up in a separate row at the top of Tracy, labeled as `RenderQueue`.
 
-In terms of where your app is spending time doing graphics work, it might manifest as a CPU bottleneck (extracting to the render world, wgpu resource tracking, recording commands to a CommandBuffer, or GPU driver code), or it might manifest as a GPU bottleneck (the GPU actually running your commands).
+> [!NOTE]
+> Due to dynamic clock speeds, GPU timings will have large frame-to-frame variance, unless you use an external tool to lock your GPU clocks to base speeds. When measuring GPU performance via Tracy, only look at the MTPC column of Tracy's statistics panel, and not individual frame data.
+<!-- markdownlint-disable MD028 -->
```
superdump, line 158: I usually use medians rather than means. Medians are not as sensitive to outliers. Plus looking at the Tracy comparison histograms (the red vs yellow) gives a good feel. I do also look at the standard deviation but that’s usually visually apparent from the graph anyway.

superdump (commented):
I don’t have any blocking comments, but comments nonetheless.

crates/bevy_render/src/diagnostic/internal.rs
JMS55: The adapter info? Getting around unused variable warnings when the tracy feature is disabled. The allow attribute doesn't work because we have a global lint to only allow the expect attribute, and expect doesn't work because when the feature _is_ enabled, the variable won't actually be unused 🙃.

crates/bevy_render/src/diagnostic/tracy_gpu.rs
JMS55: I think it will still work, just tracy will show an unknown backend type in the UI. Don't think it really affects anything.

docs/profiling.md
JMS55: I don't actually know what MTPC is - is it the mean? 

Either way, agree, but I don't think tracy has any good way to check the median besides manually estimating from the histogram.

Regardless tools like nsight/rgp will be much better for stable and detailed measurements, this is meant to just be a quick aid to help people when reviewing PRs.

JMS55:
We might want to consider adding RenderDiagnosticPlugins to DefaultPlugins when the tracy feature is enabled, I think people are going to forget to add it when reviewing PRs. But I'll leave that to a possible followup.

bushrat011899:
> We might want to consider adding RenderDiagnosticPlugins to DefaultPlugins when the tracy feature is enabled, I think people are going to forget to add it when reviewing PRs. But I'll leave that to a possible followup.

Definitely agree with this (I was exactly that person). It's not like you'd want to run Tracy _without_ that information.

crates/bevy_pbr/src/render/light.rs
IceSentry, line 1: Why was the time span removed here?

IceSentry: Oh, I opened everything and I see there's already spans for everything

IceSentry (approved):
I'm in favor of adding the plugin by default when the feature is enabled. Other than that LGTM

crates/bevy_render/src/diagnostic/tracy_gpu.rs
```diff
@@ -0,0 +1,67 @@
...
+    };
+
+    let tracy_client = Client::running().unwrap();
```
IceSentry, line 20: Would be nice to avoid the unwraps and at least use expect or bubble up the error, but it's not a blocker.

crates/bevy_pbr/src/render/light.rs
JMS55: The top level shadow time span were setup weirdly, since each shadow cascade gets sent to its own thread and its own command encoder, it never really worked like it was supposed to.

We still have individual shadow spans.

bushrat011899 (approved):

IceSentry (approved):

tychedelia (approved):
Awesome, thanks for doing this!

crates/bevy_render/src/diagnostic/tracy_gpu.rs
superdump: Oh phew. It is just about UI labels then that’s very minor. :)

docs/profiling.md
superdump: I’m now not sure if it’s only in the comparison view with the histogram graphs or if it can also be seen elsewhere. But in the comparison view it calculates the median, mean, stddev and shows them so if they’d like standard spans they’ll be there :)

MTPC is mean time per call. Anyway, not blocking.

superdump (approved):

PR #18493, opened 2025-03-22
Fix UpdateMode::Reactive behavior on Windows

aloucks:
# Objective

The fix in #17488 forced Windows to always behave as if it were in `UpdateMode::Continuous`.

CC https://github.com/bevyengine/bevy/pull/17991

## Solution

Removed the unconditional `redraw_requested = true` and added a check for `Reactive` in `about_to_wait`.

## Testing

- Verified that the `low_power` example worked as expected with all `UpdateMode` options.
- Verified that animation continued in both `eased_motion ` and `low_power` examples when in `Continuous` update mode while:
  - Resizing the Window
  - Moving the window via clicking and dragging the title bar
- Verified that `window_settings` example still worked as expected.
- Verified that `monitor_info` example still worked as expected.

### DIFF
--- a/crates/bevy_winit/src/state.rs
+++ b/crates/bevy_winit/src/state.rs
@@ -454,7 +454,6 @@ impl<T: Event> ApplicationHandler<T> for WinitAppRunnerState<T> {
                     // Have the startup behavior run in about_to_wait, which prevents issues with
                     // invisible window creation. https://github.com/bevyengine/bevy/issues/18027
                     if self.startup_forced_updates == 0 {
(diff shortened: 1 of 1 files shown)


greeble-dev (approved):
I've run all the tests mentioned in the PR, and confirmed the `low_power` example was broken in main. Also confirmed the code is windows specific. Tested on Win10.

I'm not very familiar with the code or winit so I'm slightly nervous about this, but clicking approve as the test coverage seems reasonable.

Issue #18453, opened 2025-03-20
Increase structure for schedule executors to reduce boilerplate and ensure consistency


alice-i-cecile:
## Problem

When implementing a new system executor, users (and engine devs) need to ensure a wide array of functionality and behavior is followed. These requirements are scattered across bevy_ecs, leading to "copy-paste an existing executor" as the only sensible strategy.

Similarly, when attempting to make changes to this code, engine devs need to go and carefully manually implement these changes across multiple call sites, being sure to maintain consistency.

## Proposed solution

More behavior should be moved out of `SystemExecutor::run`, and into new methods on the `SystemExecutor` trait. Ideally, `SystemExecutor::run`'s default implementation can call all of these submethods in order to achieve the correct behavior.

Additionally, `evaluate_and_fold_conditions` should be moved out into a public helper and reused across executors.

ElliottjPierce:
After aquatinting myself roughly with how this all works, I'm kinda stumped as to where to draw the line on this abstraction. `SimpleExecutor` and `SingleThreadedExecutor` are nearly identical and `MultiThreadedExecutor` is completely different. It further complicates things that we'd kinda need a data level abstraction too between `SimpleExecutor` and `SingleThreadedExecutor`.

If we still want to do this, I'll probably end up just making some utility functions like `evaluate_and_fold_conditions` that are used mostly between `SimpleExecutor` and `SingleThreadedExecutor`.

As an alternative, why not get rid of `SimpleExecutor`? It looks like `SingleThreadedExecutor` basically fills the same purpose; it just does it better/more efficiently. The only use case I can think of for `SimpleExecutor` is a ton of systems that all need to be applied deferred. But then, why not an exclusive system? If users are concerned about performance, they won't have that use case, and if they're not, they won't mind switching to `SingleThreadedExecutor`. If we went that route, there would be very little duplication, and what little would be left, I'd argue would be more clear not abstracted.

Up to you, but what do you think about that?

alice-i-cecile:
> As an alternative, why not get rid of SimpleExecutor?

This was created as part of an effort to make scheduling more approachable for users who didn't care about performance. I'm not convinced it's carrying its weight after the inferred sync points though.

ElliottjPierce:
> > As an alternative, why not get rid of SimpleExecutor?
> 
> This was created as part of an effort to make scheduling more approachable for users who didn't care about performance. I'm not convinced it's carrying its weight after the inferred sync points though.

[...]

PR #18504, opened 2025-03-23
Define system param validation on a per-system parameter basis

alice-i-cecile:
# Objective

When introduced, `Single` was intended to simply be silently skipped, allowing for graceful and efficient handling of systems during invalid game states (such as when the player is dead).

However, this also caused missing resources to *also* be silently skipped, leading to confusing and very hard to debug failures. In 0.15.1, this behavior was reverted to a panic, making missing resources easier to debug, but largely making `Single` (and `Populated`) worthless, as they would panic during expected game states.

Ultimately, the consensus is that this behavior should differ on  a per-system-param basis. However, there was no sensible way to *do* that before this PR.

## Solution

Swap `SystemParam::validate_param` from a `bool` to:

```rust
/// The outcome of system / system param validation,
/// used by system executors to determine what to do with a system.
pub enum ValidationOutcome {
    /// All system parameters were validated successfully and the system can be run.
    Valid,
    /// At least one system parameter failed validation, and an error must be handled.
    /// By default, this will result in1 a panic. See [crate::error] for more information.
    ///
    /// This is the default behavior, and is suitable for system params that should *always* be valid,
    /// either because sensible fallback behavior exists (like [`Query`] or because
    /// failures in validation should be considered a bug in the user's logic that must be immediately addressed (like [`Res`]).
    Invalid,
    /// At least one system parameter failed validation, but the system should be skipped due to [`ValidationBehavior::Skip`].
    /// This is suitable for system params that are intended to only operate in certain application states, such as [`Single`].
    Skipped,
}
```
Then, inside of each `SystemParam` implementation, return either Valid, Invalid or Skipped.

Currently, only `Single`, `Option<Single>` and `Populated` use the `Skipped` behavior. Other params (like resources) retain their current failing

## Testing

Messed around with the fallible_params example. Added a pair of tests: one for panicking when resources are missing, and another for properly skipping `Single` and `Populated` system params.

## To do

- [x] get https://github.com/bevyengine/bevy/pull/18454 merged
- [x] fix the todo!() in the macro-powered tuple implementation (please help 🥺)
- [x] test
- [x] write a migration guide
- [x] update the example comments

## Migration Guide

Various system and system parameter validation methods (`SystemParam::validate_param`, `System::validate_param` and `System::validate_param_unsafe`) now return and accept a `ValidationOutcome` enum, rather than a `bool`.  The previous `true` values map to `ValidationOutcome::Valid`, while `false` maps to `ValidationOutcome::Invalid`.

However, if you wrote a custom schedule executor, you should now respect the new `ValidationOutcome::Skipped` parameter, skipping any systems whose validation was skipped. By contrast, `ValidationOutcome::Invalid` systems should also be skipped, but you should call the `default_error_handler` on them first, which by default will result in a panic.

If you are implementing a custom `SystemParam`, you should consider whether failing system param validation is an error or an expected state, and choose between `Invalid` and `Skipped` accordingly. In Bevy itself, `Single` and `Populated` now once again skip the system when their conditions are not met. This is the 0.15.0 behavior, but stands in contrast to the 0.15.1 behavior, where they would panic.


crates/bevy_ecs/src/system/system_param.rs
```diff
@@ -2065,9 +2107,10 @@ macro_rules! impl_system_param_tuple {
                 state: &Self::State,
                 system_meta: &SystemMeta,
                 world: UnsafeWorldCell,
-            ) -> bool {
+            ) -> ValidationOutcome {
                 let ($($param,)*) = state;
-                $($param::validate_param($param, system_meta, world)&&)* true
+                //$($param::validate_param($param, system_meta, world)&&)* true
+                todo!("How do I translate this to use ValidationOutcome::combine?");
```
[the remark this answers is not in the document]
MiniaczQ: Generating code for N-wise binary operation can add up, so since we define the function, I'd suggest adding a class method that accepts an array and use that here.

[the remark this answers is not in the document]
MiniaczQ: If you wanna go with the code-gen pair-wise operations, here's a reference solution
https://github.com/MiniaczQ/bevy-state-v3/blob/6257c58f228ca2113b2c25127d232825a118b940/bevy_state_v3/src/state_set.rs#L80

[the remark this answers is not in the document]
chescock: You want it to generate something like

```rust
ValidationOutcome::Valid
    .combine(P0::validate_param(P0, system_meta, world))
    .combine(P1::validate_param(P1, system_meta, world))
```

right?  

Then I think you can do it like

```rust
ValidationOutcome::Valid
    $(.combine($param::validate_param($param, system_meta, world)))*
```

(Unless you want it to short-circuit?)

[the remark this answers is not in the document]
MiniaczQ: Didn't think of that one

NthTensor:
Starting a review, but before I do I want to note that this is actually how I thought `Single` was implemented in the first place. Somehow I hadn't connected up the changes to resources with the `Single<T>` feature. Just from the description, skipping behavior seems like a really useful thing to be able to express.

alice-i-cecile:
> This is kinda hard to review with the commits from both PRs mixed together. I'm going to take a rain check until the other one is merged.

Agreed! This will be way cleaner after.

chescock (approved):

MiniaczQ:
> If we _do_ want a generic wrapper like that, then it might be more consistent if that wrapper were the _only_ thing that returned `Skipped`, in which case `Single` should return `Invalid`. That would make `Single` a little less convenient, but it would simplify the story for parameters-as-run-conditions, since the _only_ way to skip a system would be with that wrapper type.

It's worth the boilerplate, especially if we bikeshed a short name.

NthTensor:
I really like the optional system param thing, I was wondering about that myself. But I'm less sold on having a single skipping wrapper; however if we did decide to go with it my vote would be for `When<T>` (eg. `When<Res<MyResource>>` or `When<Single<Window>>`.

crates/bevy_ecs/src/schedule/executor/multi_threaded.rs
```diff
@@ -580,10 +581,21 @@ impl ExecutorState {
             // - The caller ensures that `world` has permission to read any data
             //   required by the system.
             // - `update_archetype_component_access` has been called for system.
-            let valid_params = unsafe { system.validate_param_unsafe(world) };
-            if !valid_params {
-                self.skipped_systems.insert(system_index);
-            }
+            let valid_params = match unsafe { system.validate_param_unsafe(world) } {
```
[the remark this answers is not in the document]
alice-i-cecile: This is part of the problem outlined in https://github.com/bevyengine/bevy/issues/18453, but yeah, maybe we can extract it.

crates/bevy_ecs/src/system/system_param.rs
```diff
@@ -2619,6 +2670,13 @@ unsafe impl SystemParam for FilteredResourcesMut<'_, '_> {
     }
 }
 
+/// An error that occurs when a system parameter is not valid.
+///
+/// Generated when [`SystemParam::validate_param`] returns `false`,
+/// and handled using the unified error handling mechanisms defined in [`bevy_ecs::error`].
+#[derive(Debug, PartialEq, Eq, Clone, Display, Error)]
+pub struct SystemParamValidationError;
```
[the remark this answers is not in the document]
alice-i-cecile: 100%! This would be really helpful.

NthTensor:
> Option<Single> is valid with 0 or 1, but not N!

Right, so this would mean changing the semantics of `Single`. Hmm, it's an interesting idea still. Want to know what @alice-i-cecile thinks.

NthTensor:
Well exactly, so does `Option<When<T>>` return `None` or does it skip? We could decide to give the outer wrapper precedence, but we could also give `Option` prescience.

MiniaczQ:
> Well exactly, so does `Option<When<T>>` return `None` or does it skip? We could decide to give the outer wrapper precedence, but we could also give `Option` prescience.

My first thought was to use type system, but that'd add too much complexity. We can consider such combinations to be invalid at runtime, much like overlapping Query access.

chescock:
> One slightly awkward result is the behavior of `Option<When<T>>` and `When<Option<T>>`. Should the former skip?

I would expect those combinations to only appear in generic code.  So you'd have a system like `fn generic_system<P>(p: Option<P>)` and then someone does `generic_system::<When<T>>`.  

If `Option<When<T>>` skips, then it's identical to `When<T>` (but it wraps the param in `Some`), and if it produces `None`, then it's identical to `Option<T>`.  Which means the skipping behavior is sometimes useful, since you can cause `generic_system` to skip, while the failing behavior is never useful, since you can already do `generic_system::<T>`.  

So I think `Option<When<T>>` should skip if `T` fails, and never produce `None`.  

... although that has the unfortunate property that you need to call `validate_param` twice, once in `validate_param` and once in `get_param`.  So if the answer is that nobody will ever actually use `Option<When<T>>`, then it should return `None` to get the less expensive implementation.

crates/bevy_ecs/src/system/system_param.rs
```diff
@@ -47,10 +47,7 @@ use variadics_please::{all_tuples, all_tuples_enumerated};
 ///
 /// Derived `SystemParam` structs may have two lifetimes: `'w` for data stored in the [`World`],
 /// and `'s` for data stored in the parameter's state.
-///
-/// The following list shows the most common [`SystemParam`]s and which lifetime they require
-///
-/// ```
+///validate
```
alice-i-cecile, line 50:
```suggestion
///
/// The following list shows the most common [`SystemParam`]s and which lifetime they require
///
/// ```
```

crates/bevy_ecs/src/system/system_param.rs
```diff
@@ -309,6 +307,38 @@ unsafe impl<'w, 's, D: ReadOnlyQueryData + 'static, F: QueryFilter + 'static> Re
...
+    /// Combines two [`ValidationOutcome`]s, returning the most severe one.
+    ///
+    /// If either outcome is [`ValidationOutcome::Invalid`], the result will be [`ValidationOutcome::Invalid`].
+    /// Otherwise, if either outcome is [`ValidationOutcome::Skipped`], the result will be [`ValidationOutcome::Skipped`].
+    /// Finally, if both outcomes are [`ValidationOutcome::Valid`], the result will be [`ValidationOutcome::Valid`].
+    pub const fn combine(self, other: Self) -> Self {
+        match (self, other) {
+            (Self::Invalid, _) | (_, Self::Invalid) => Self::Invalid,
+            (Self::Skipped, _) | (_, Self::Skipped) => Self::Skipped,
+            (Self::Valid, Self::Valid) => Self::Valid,
```
NthTensor, line 337: Note to the other reviewers, this section defines the precedence of how `Option` and `When` would interact.

crates/bevy_ecs/src/system/system_param.rs
```diff
@@ -309,6 +307,38 @@ unsafe impl<'w, 's, D: ReadOnlyQueryData + 'static, F: QueryFilter + 'static> Re
 {
 }
 
+/// The outcome of system / system param validation,
+/// used by system executors to determine what to do with a system.
+pub enum ValidationOutcome {
+    /// All system parameters were validated successfully and the system can be run.
+    Valid,
+    /// At least one system parameter failed validation, and an error must be handled.
+    /// By default, this will result in1 a panic. See [crate::error] for more information.
```
dmyyy, line 319:
```suggestion
    /// By default, this will result in a panic. See [crate::error] for more information.
```

NthTensor (approved):
I really really like this, the new `ValidationOutcome` enum is great. I think we should probably fork the `Option/When` discussion to a new issue or pr.

crates/bevy_ecs/src/schedule/executor/multi_threaded.rs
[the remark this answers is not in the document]
alice-i-cecile: I tried extracting this out, but I'm not pleased with how the straightforward approach requires eagerly constructing an `ErrorContext` in the happy path. IMO this is better suited to a proper overhaul of our executor stack.

alice-i-cecile:
Spun out the `When` discussion into #18516 (I really like it, but it's off-topic here). I'm going to add more tests and clean up the example then get this merged.

crates/bevy_ecs/src/schedule/executor/multi_threaded.rs
```diff
@@ -581,18 +581,22 @@ impl ExecutorState {
...
+                        ErrorContext::System {
+                            name: system.name(),
+                            last_run: system.get_last_run(),
+                        },
+                    );
+                    false
+                }
+                ValidationOutcome::Skipped => false,
+            };
+            self.skipped_systems.insert(system_index);
```
chescock, line 598: You still need this in an `if`, though, right?  

```suggestion
            if !valid_params {
                self.skipped_systems.insert(system_index);
            }
```

crates/bevy_ecs/src/system/system_param.rs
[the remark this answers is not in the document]
alice-i-cecile: https://github.com/bevyengine/bevy/issues/18515

crates/bevy_ecs/src/system/system_param.rs
```diff
@@ -503,8 +540,10 @@ unsafe impl<'a, D: QueryData + 'static, F: QueryFilter + 'static> SystemParam
                 world.change_tick(),
             )
         };
-        let result = query.single_inner();
-        !matches!(result, Err(QuerySingleError::MultipleEntities(_)))
+        match query.single_inner() {
+            Ok(_) | Err(QuerySingleError::NoEntities(_)) => ValidationOutcome::Valid,
+            Err(QuerySingleError::MultipleEntities(_)) => ValidationOutcome::Skipped,
+        }
```
mockersf, line 546: this is the same as before, but why is it like that?
0 would give None
1 would give Some
n would give Skipped?

crates/bevy_ecs/src/system/system_param.rs
```diff
@@ -309,6 +310,40 @@ unsafe impl<'w, 's, D: ReadOnlyQueryData + 'static, F: QueryFilter + 'static> Re
...
+    Skipped,
+}
+
+impl ValidationOutcome {
+    /// Combines two [`ValidationOutcome`]s, returning the most severe one.
+    ///
+    /// If either outcome is [`ValidationOutcome::Invalid`], the result will be [`ValidationOutcome::Invalid`].
+    /// Otherwise, if either outcome is [`ValidationOutcome::Skipped`], the result will be [`ValidationOutcome::Skipped`].
+    /// Finally, if both outcomes are [`ValidationOutcome::Valid`], the result will be [`ValidationOutcome::Valid`].
+    pub const fn combine(self, other: Self) -> Self {
```
mockersf, line 338: this is not the same as the `&&` that was used before to combine bools: `&&` allows us to stop evaluating parameters as soon as an invalid one is found, this will need to evaluate them all

alice-i-cecile: I don't think we should be short-circuiting here: it will substantially worsen our error messages once #18515 is added.

crates/bevy_ecs/src/system/system_param.rs
alice-i-cecile: Yeah, that was the intended design. I'm not sold on it being particularly useful.

crates/bevy_ecs/src/system/system_param.rs
mockersf: this has the potential to be a lot slower, but then it doesn't matter because right now we panic so we would be slower to report the error 😄 

I think we should still short circuit. We would still be able to report the first failure encountered which would be OK for #18515. accumulating failures would need more changes and probably a way bigger hit on perfs as you would need something with a dynamic size to accumulate errors

alice-i-cecile: Okay fine, the dynamically sized error accumulation would suck. I've mostly fixed short-circuiting, but left a PERF note on the macro.

mockersf (approved):

crates/bevy_ecs/src/system/system_param.rs
NthTensor: I recommend we continue discussion about improving the semantics of the existing system either in https://github.com/bevyengine/bevy/issues/18516 or a new issue.

PR #18506, opened 2025-03-24
Switch from `OnceCell` to `LazyLock` in `bevy_tasks`

bushrat011899:
# Objective

Remove `critical-section` from required dependencies, allowing linking without any features.

## Solution

- Switched from `OnceCell` to `LazyLock`
- Removed `std` feature from `bevy_dylib` (proof that it works)

## Testing

- CI

### DIFF
--- a/crates/bevy_tasks/Cargo.toml
+++ b/crates/bevy_tasks/Cargo.toml
@@ -26,11 +26,7 @@ async_executor = ["std", "dep:async-executor"]
 ## Allows access to the `std` crate. Enabling this feature will prevent compilation
 ## on `no_std` targets, but provides access to certain additional features on
 ## supported platforms.
-std = [
-  "futures-lite/std",
-  "async-task/std",
-  "bevy_platform_support/std",
-]
+std = ["futures-lite/std", "async-task/std", "bevy_platform_support/std"]
 
 ## `critical-section` provides the building blocks for synchronization primitives
 ## on all platforms, including `no_std`.
--- a/crates/bevy_tasks/Cargo.toml
+++ b/crates/bevy_tasks/Cargo.toml
@@ -35,7 +35,7 @@ std = [
 
 ## `critical-section` provides the building blocks for synchronization primitives
 ## on all platforms, including `no_std`.
-critical-section = ["bevy_platform_support/critical-section"]
+critical-section = ["bevy_platform_support/critical-section", "once_cell/critical-section"]
 
 ## Enables use of browser APIs.
 ## Note this is currently only applicable on `wasm32` architectures.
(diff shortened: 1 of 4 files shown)


crates/bevy_tasks/src/executor.rs
```diff
@@ -19,8 +19,8 @@ cfg_if::cfg_if! {
         type ExecutorInner<'a> = async_executor::Executor<'a>;
         type LocalExecutorInner<'a> = async_executor::LocalExecutor<'a>;
     } else {
-        type ExecutorInner<'a> = crate::edge_executor::Executor<'a, 64>;
-        type LocalExecutorInner<'a> = crate::edge_executor::LocalExecutor<'a, 64>;
+        type ExecutorInner<'a> = crate::edge_executor::Executor<'a, 128>;
+        type LocalExecutorInner<'a> = crate::edge_executor::LocalExecutor<'a, 128>;
```
bushrat011899, line 23: Noticed that some tests in `bevy_tasks` would fail at 64, but pass at 128.

crates/bevy_tasks/Cargo.toml
```diff
@@ -65,9 +64,6 @@ async-channel = { version = "2.3.0", optional = true }
 async-io = { version = "2.0.0", optional = true }
 concurrent-queue = { version = "2.0.0", optional = true }
 atomic-waker = { version = "1", default-features = false }
-once_cell = { version = "1.18", default-features = false, features = [
-  "critical-section",
-] }
```
bushrat011899, line 70: Don't need this anymore!

alice-i-cecile (approved):

crates/bevy_tasks/src/executor.rs
mockersf: Was it already true before this PR?
The default value from the crate is 64, do we have an explanation why we need more?

bushrat011899: Yeah I checked to be sure, the tests failed on main as well. It was specifically the `Slice` parallel chunking tests. My guess is they up-front create too many tasks and blow the pre-allocated queue.

mockersf: I don't like forcing everyone to use a larger value because of our tests.
Does that means also that if someone needs a larger queue, they can't? is it possible to make it user customisable?

those questions are not blocking for this PR has it already fails on main

bushrat011899:
I've removed the contentious change to executor buffer size, since it's strictly tangential to the goal of this PR. It still should be addressed, but it doesn't have to be here.

mockersf (approved):

PR #18513, opened 2025-03-24
add 2d_on_ui example

mirsella:
hello ! im not english native, don't hesitate to correct my code comments.

# Objective

a simple example for new users. This is a question asked a lot, and i struggled to do it at first too.
see https://github.com/bevyengine/bevy/discussions/11223

## Showcase

image


alice-i-cecile (changes requested):
I like the general idea, but I think we should try and pick something a bit more representative (and less confusing) for the content. Ideally the UI is buttons or text or something, and the 2D element is a sprite?

Maybe a simple paper doll mockup or something?

mirsella:
sorry i dont see what to do with a paper doll mockup to demonstrate this, but would be happy to do it !
or maybe a simple counter button with the count text, and a sprite inside the button then ?I

mirsella:
hm actually is it really good ? because a imagenode could have been used in button too, and would be better to keep it UI

mirsella:
would be happy to finish this pr, but im having trouble finding an idea of what to implement. 
> Maybe a simple paper doll mockup or something?
can you elaborate please ? like putting a random Sprite on top of the ui node background ?

alice-i-cecile:
Sorry, I missed getting back to you :) https://di.diablowiki.net/Paperdoll was what I had in mind. You don't need to implement a complete inventory: just show the rendering side of things.

mirsella:
[Screencast_20250522_231930.webm](https://github.com/user-attachments/assets/9aba06fa-0958-41a5-b7aa-01c0dadd40ed)

alice-i-cecile (approved):
Thanks for the feedback, and thanks for being so receptive making changes :) I'm happy with this now.

PR #18517, opened 2025-03-24
Render events

ecoskey:
# Objective

- Sometimes users want to send events from the render world to the main world. While not often advised, some advanced use-cases like readback use a similar pattern, which often requires setting up a channel and relaying events manually.

## Solution

- Add a simple abstraction (`App::init_render_event`) to set up the channel and relay events to the main world, retaining caller information.

- Note: this pr makes `Events::send_with_caller` public, but does *not* add a similar public method on `EventWriter`

## Testing

- Ran example

### DIFF
--- a/crates/bevy_render/src/render_event.rs
+++ b/crates/bevy_render/src/render_event.rs
@@ -1,0 +1,121 @@
+use core::marker::PhantomData;
+
+use async_channel::{Receiver, Sender};
+use bevy_app::{App, Plugin, PreUpdate};
+use bevy_ecs::{
+    change_detection::MaybeLocation,
+    event::{Event, Events},
+    resource::Resource,
+    system::{Res, ResMut, SystemParam},
+};
+
+use crate::RenderApp;
+
+pub trait RenderEventApp {
+    fn add_render_event<E: Event>(&mut self) -> &mut Self;
+}
+
+impl RenderEventApp for App {
+    fn add_render_event<E: Event>(&mut self) -> &mut Self {
+        self.add_plugins(RenderEventPlugin::<E>::default())
+    }
+}
+
+struct RenderEventPlugin<E: Event>(PhantomData<E>);
+
+impl<E: Event> Default for RenderEventPlugin<E> {
+    fn default() -> Self {
+        Self(PhantomData)
+    }
+}
+
+impl<E: Event> Plugin for RenderEventPlugin<E> {
+    fn build(&self, app: &mut App) {
+        app.add_event::<E>()
+            .add_systems(PreUpdate, relay_render_events::<E>);
+    }
+
+    fn finish(&self, app: &mut App) {
+        let (sender, receiver) = async_channel::unbounded::<(E, MaybeLocation)>();
+
+        app.insert_resource(RenderEventReceiver(receiver));
+
+        let Some(render_app) = app.get_sub_app_mut(RenderApp) else {
+            return;
+        };
+
+        render_app.insert_resource(RenderEventSender(sender));
+    }
+}
+
+#[derive(Resource)]
+struct RenderEventReceiver<E: Event>(Receiver<(E, MaybeLocation)>);
+#[derive(Resource)]
+struct RenderEventSender<E: Event>(Sender<(E, MaybeLocation)>);
+
+fn relay_render_events<E: Event>(
+    mut events: ResMut<Events<E>>,
+    receiver: Res<RenderEventReceiver<E>>,
+) {
+    while let Ok((event, caller)) = receiver.0.try_recv() {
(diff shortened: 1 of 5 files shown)


examples/shader/render_event.rs
```diff
@@ -0,0 +1,82 @@
...
+        )
+            .build_state(render_app.world_mut())
+            .build_system(send_render_event);
+
+        render_app.add_systems(Render, send_render_event);
+    }
+}
+
+#[derive(Resource, Default, Clone, Copy, ExtractResource)]
+struct FrameIndex(u32);
```
IceSentry, line 51: Very minor nitpick, but maybe rename this to something like Counter or anything else? Or maybe add a comment explaining this is for a render world frame index. I just don't want new users to think we don't have this data already. Most users should use `Res<FrameCount>` if they want this number. Tbf, `FrameCount` is actually about main world frames not render world frames so I guess there might be a valid reason to want this.

ecoskey: I just never noticed we had this built-in lol. As long as it's extracted I'll just use that.

IceSentry (approved):

examples/shader/render_event.rs
IceSentry: Yeah, I know many people aren't aware of it which is why I mentioned it 😅

IceSentry:
They are events from the render world. I think it's fair to call them render event. Presumably, users will need this to send events in the context of rendering. Otherwise they wouldn't need events sent from the render world. Maybe specify render world event just to remove any possible ambiguity, but I'm not sure what kind of ambiguity there could be.

crates/bevy_render/src/render_event.rs
```diff
@@ -0,0 +1,121 @@
...
+use bevy_ecs::{
+    change_detection::MaybeLocation,
+    event::{Event, Events},
+    resource::Resource,
+    system::{Res, ResMut, SystemParam},
+};
+
+use crate::RenderApp;
+
+pub trait RenderEventApp {
```
tychedelia, line 14: Not sure what our convention is here -- should this be in the prelude so it doesn't have to be imported?

examples/shader/render_event.rs
```diff
@@ -0,0 +1,90 @@
...
+        render_app.init_resource::<FrameCount>();
+
+        let send_render_event = (
+            ParamBuilder,
+            ParamBuilder,
+            ParamBuilder,
+            LocalBuilder(Timer::from_seconds(2.0, TimerMode::Repeating)),
+        )
+            .build_state(render_app.world_mut())
+            .build_system(send_render_event);
```
IceSentry, line 45: This feels a bit unnecessarily complex. I'd suggest using a resource instead just because that's a more common ECS api.

crates/bevy_render/src/render_event.rs
```diff
@@ -0,0 +1,121 @@
...
+    while let Ok((event, caller)) = receiver.0.try_recv() {
+        events.send_with_caller(event, caller);
+    }
+}
+
+/// An event writer for sending events from the render world to the main world.
+///
+/// Internally, this struct writes to a channel, the contents of which are relayed
+/// to the main world's [`Events`] stream during [`PreUpdate`]. Note that because
+/// the render world is pipelined, the events may not arrive before the next frame begins.
```
tychedelia, line 69: They almost certainly will not, but they also could. I think we might want to be clearer about how this works. In practice, you'll receive events from rendering frame N-2 with respect the the main world, but that isn't guaranteed. Users should supply their own temporal correlation id if necessary.

examples/shader/render_event.rs
```diff
@@ -0,0 +1,90 @@
+//! Simple example demonstrating the use of [`App::init_render_event`] to send events from the
```
tychedelia, line 1: This should be expanded to say, in effect, "you probably don't want to do this."

tychedelia (approved):
Think formalizing this makes sense versus ad hoc. 👍

alice-i-cecile (changes requested):
Broadly on-board with this pattern and it existing, but I think we need to clean this up before merging.

Key things:

- the system builder in the example is too complicated and distracts from the point of the example
- the example description needs to be expanded to give more context
- this needs a non "Event" naming scheming to avoid confusion. Maybe something with "channel" in it?

Long-term, this should be baked into a multi-world ECS design, and not need channels, but formalizing this pattern in *some* form is a good way to gather requirements and ease refactoring.

alice-i-cecile:
Cut from the milestone: this would be nice to have but shouldn't block Bevy 0.17's release.

ecoskey:
closing, stale.

PR #18525, opened 2025-03-25
Remote entity reservation v6

### LINKED ISSUE #18003: Reserve entities from async

ElliottjPierce:
# Objective

Like the other *5* versions, the goal is to allow entity reservation from any thread asynchronously. This is primarily important for assets as entities. fixes #18003.

## Solution

This differs from #18440 (v5) by reserving one by one instead of in bulk. This will probably have worse performance, but it's definitely passible, especially given that async reservation is a relatively slow path. In fact, with a high batch size, this may even be faster than v5. See v5 benchmarks for the impact on the rest of the ecs, as the changes there are almost identical.

## Testing

The test from v5 has been ported over.

## Questions for review

If we are ok with not exposing the batch size for remote, we could make this use a bounded queue, which would improve performance of remote reservation.

### DIFF
--- a/crates/bevy_ecs/src/entity/mod.rs
+++ b/crates/bevy_ecs/src/entity/mod.rs
@@ -45,6 +45,7 @@ use bevy_reflect::Reflect;
 use bevy_reflect::{ReflectDeserialize, ReflectSerialize};
 
 pub use clone_entities::*;
+use concurrent_queue::ConcurrentQueue;
 pub use entity_set::*;
 pub use map_entities::*;
 
@@ -73,8 +74,18 @@ use crate::{
     storage::{SparseSetIndex, TableId, TableRow},
 };
 use alloc::vec::Vec;
-use bevy_platform_support::sync::atomic::Ordering;
-use core::{fmt, hash::Hash, mem, num::NonZero, panic::Location};
+use bevy_platform_support::sync::{
+    atomic::{AtomicU32, Ordering},
+    Arc,
+};
+use core::{
+    fmt,
+    hash::Hash,
+    mem,
+    num::NonZero,
+    panic::Location,
+    task::{Poll, Waker},
+};
 use log::warn;
 
 #[cfg(feature = "serialize")]
@@ -515,6 +526,159 @@ impl<'a> core::iter::FusedIterator for ReserveEntitiesIterator<'a> {}
 // SAFETY: Newly reserved entity values are unique.
 unsafe impl EntitySetIterator for ReserveEntitiesIterator<'_> {}
 
+#[derive(Debug)]
+struct RemoteEntityRequest(Waker);
+
+#[derive(Debug)]
+struct RemoteEntitiesInner {
+    /// This is purely an optimization.
+    /// When this is greater than 0, we are waiting on request fulfillment.
+    is_waiting: AtomicU32,
+    /// The number of times requests have been fulfilled.
+    generation: AtomicU32,
+    /// Since anyone can make requests at any time, this must be unbounded.
+    requests: ConcurrentQueue<RemoteEntityRequest>,
+    reserved: ConcurrentQueue<Entity>,
+}
+
+/// An error that occurs when an [`Entity`] can not be reserved remotely.
+/// See also [`RemoteEntities`].
+#[derive(thiserror::Error, Debug, Clone, Copy, PartialEq, Eq)]
+pub enum RemoteReservationError {
+    /// This happens when [`Entities`] are closed, dropped, etc while a [`RemoteEntitiesReserver`] is trying to reserve from it.
+    #[error("A remote entity reserver tried to reserve an entity from a closed `Entities`.")]
+    Closed,
+}
+
+struct RemoteReservedEntities {
+    source: Arc<RemoteEntitiesInner>,
+    requested_on_generation: Option<u32>,
+}
+
+impl Future for RemoteReservedEntities {
+    type Output = Result<Entity, RemoteReservationError>;
+
+    fn poll(
+        mut self: core::pin::Pin<&mut Self>,
+        cx: &mut core::task::Context<'_>,
+    ) -> Poll<Self::Output> {
+        match self.source.reserved.pop() {
+            Ok(found) => Poll::Ready(Ok(found)),
+            Err(concurrent_queue::PopError::Closed) => {
+                Poll::Ready(Err(RemoteReservationError::Closed))
+            }
+            Err(concurrent_queue::PopError::Empty) => {
+                let current_generation = self.source.generation.load(Ordering::Acquire);
+                if self
+                    .requested_on_generation
+                    .is_none_or(|generation| generation != current_generation)
+                {
+                    self.requested_on_generation = self
+                        .source
+                        .requests
+                        .push(RemoteEntityRequest(cx.waker().clone()))
+                        .is_ok()
+                        .then_some(current_generation);
+                }
+
+                // We can use relaxed ordering here. Even though it might result in extra waiting, who cares?
+                self.source.is_waiting.fetch_add(1, Ordering::Relaxed);
+                // Now we are waiting for a flush
+                Poll::Pending
+            }
+        }
+    }
+}
+
+impl RemoteEntitiesInner {
+    fn fulfill(&self, entities: &Entities, batch_size: NonZero<u32>) {
+        self.is_waiting.store(0, Ordering::Relaxed);
+        for reserved in entities.reserve_entities(batch_size.get()) {
+            self.reserved
+                .push(reserved)
+                .expect("This never closes, and it can't run out of room.");
+        }
+
+        self.generation.fetch_add(1, Ordering::AcqRel);
+        for request in self.requests.try_iter() {
+            request.0.wake();
+        }
+    }
+
+    #[inline]
+    fn try_fulfill(&self, entities: &Entities, batch_size: NonZero<u32>) {
+        // we do this to hint to the compiler that the if branch is unlinkely to be taken.
+        #[cold]
+        fn do_fulfill(this: &RemoteEntitiesInner, entities: &Entities, batch_size: NonZero<u32>) {
+            this.fulfill(entities, batch_size);
+        }
+
+        if self.is_waiting.load(Ordering::Relaxed) > 0 {
+            // TODO: add core::intrinsics::unlikely once stable
+            do_fulfill(self, entities, batch_size);
+        }
+    }
+
+    fn new() -> Self {
+        Self {
+            is_waiting: AtomicU32::new(0),
+            generation: AtomicU32::new(0),
+            requests: ConcurrentQueue::unbounded(),
+            reserved: ConcurrentQueue::unbounded(),
+        }
+    }
+
+    fn close(&self) {
+        self.reserved.close();
+        self.requests.close();
+    }
+}
+
+/// Manages access to [`Entities`] from any thread and async.
+pub struct RemoteEntities {
+    inner: Arc<RemoteEntitiesInner>,
+}
+
+impl RemoteEntities {
+    /// Constructs a [`RemoteEntitiesReserver`] that can be shared between threads.
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// use bevy_ecs::prelude::*;
+    ///
+    /// let mut world = World::new();
+    /// let remote = world.entities().get_remote();
+    ///
+    /// // The reserve is async so we need it to be on a separate thread.
+    /// let thread = std::thread::spawn(move || {
+    ///     let future = async {
+    ///         for _ in 0..100 {
+    ///             reserver.reserve_entity().await.unwrap();
+    ///         }
+    ///     };
+    ///     bevy_tasks::block_on(future);
+    /// });
+    ///
+    /// // We need to flush the entities as needed or the remote entities will get stuck.
+    /// while !thread.is_finished() {
+    ///     world.flush();
+    /// }
+    /// ```
+    pub async fn reserve_entity(&self) -> Result<Entity, RemoteReservationError> {
+        RemoteReservedEntities {
+            source: self.inner.clone(),
+            requested_on_generation: None,
+        }
+        .await
+    }
+
+    /// Returns true only if the [`Entities`] has discontinued this remote access.
+    pub fn is_closed(&self) -> bool {
+        self.inner.reserved.is_closed()
+    }
+}
+
 /// A [`World`]'s internal metadata store on all of its entities.
 ///
 /// Contains metadata on:
@@ -525,6 +689,10 @@ unsafe impl EntitySetIterator for ReserveEntitiesIterator<'_> {}
 /// [`World`]: crate::world::World
 #[derive(Debug)]
 pub struct Entities {
+    /// This is the number of entities we reserve in bulk for [`RemoteEntities`].
+    /// A value too high can cause excess memory to be used, but a value too low can cause additional waiting.
+    pub remote_batch_size: NonZero<u32>,
+    remote: Arc<RemoteEntitiesInner>,
     meta: Vec<EntityMeta>,
 
     /// The `pending` and `free_cursor` fields describe three sets of Entity IDs
@@ -572,11 +740,23 @@ pub struct Entities {
 }
 
 impl Entities {
-    pub(crate) const fn new() -> Self {
+    /// The default value of [`remote_batch_size`](Self::remote_batch_size).
+    pub const DEFAULT_REMOTE_BATCH_SIZE: NonZero<u32> = NonZero::new(1024).unwrap();
+
+    pub(crate) fn new() -> Self {
         Entities {
             meta: Vec::new(),
             pending: Vec::new(),
             free_cursor: AtomicIdCursor::new(0),
+            remote: Arc::new(RemoteEntitiesInner::new()),
+            remote_batch_size: Self::DEFAULT_REMOTE_BATCH_SIZE,
+        }
+    }
+
+    /// Constructs a new [`RemoteEntities`] for this instance.
+    pub fn get_remote(&self) -> RemoteEntities {
+        RemoteEntities {
+            inner: self.remote.clone(),
         }
     }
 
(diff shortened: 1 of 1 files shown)


crates/bevy_ecs/src/entity/mod.rs
```diff
@@ -515,6 +526,157 @@ impl<'a> core::iter::FusedIterator for ReserveEntitiesIterator<'a> {}
...
+                    }
+                }
+            }
+        }
+    }
+}
+
+impl RemoteEntitiesInner {
+    fn fulfill(&self, entities: &Entities, batch_size: NonZero<u32>) {
+        self.is_waiting.store(0, Ordering::Relaxed);
```
chescock, line 593: This reserves exactly `batch_size` entities per `flush`, regardless of how many requests there were.  If we make enough async requests, it could take a while for them all to get fulfilled!  

Rather than counting waiters, it might make sense to have a `request_count`, increment it *before* popping, and do `reserve_entities(request_count.swap(0, Relaxed))`.  If you pre-fill the queue with `batch_size` then that will refill the original buffer every time.  

(My proposal above has exactly the same problem, so maybe we can't use `async_channel::bounded`.)

crates/bevy_ecs/src/entity/mod.rs
```diff
@@ -515,6 +526,157 @@ impl<'a> core::iter::FusedIterator for ReserveEntitiesIterator<'a> {}
...
+        if self.reserved.is_empty() {
+            for reserved in entities.reserve_entities(batch_size.get()) {
```
chescock, line 595: Since you have `&mut Entities` when calling `try_fulfill()`, you can use `alloc()` instead of `reserve_entities()` to avoid all the atomic operations.  (You'd have to move the `try_fulfill()` call to the *end* of `flush()`, of course.)

crates/bevy_ecs/src/entity/mod.rs
```diff
@@ -515,6 +526,157 @@ impl<'a> core::iter::FusedIterator for ReserveEntitiesIterator<'a> {}
...
+    /// This happens when [`Entities`] are closed, dropped, etc while a [`RemoteEntities`] is trying to reserve from it.
+    #[error("A remote entity reserver tried to reserve an entity from a closed `Entities`.")]
+    Closed,
+}
+
+struct RemoteReservedEntities {
+    source: Arc<RemoteEntitiesInner>,
+}
+
+impl Future for RemoteReservedEntities {
```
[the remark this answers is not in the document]
ElliottjPierce: > If I'm reading this right, the manual `Future` impl and the `ConcurrentQueue<Waker>` are here to wake up readers when new entities are pushed the the queue. Is that right?

Yup. That's right.

> I think that's essentially what a Channel does over a Queue, and you can simplify the code if you switch to using a Channel. `async_channel` seems to already be in our dependency tree and `no_std` compatible.

Yeah, I wanted to use a channel, but I could't find one that was no_std. `async_channel` looks like it will work. I'll try that out in a bit.

> If you're able to lock down `remote_batch_size` at creation time, then I think you can handle both `requests` and `reserved` with a single `async_channel::bounded(remote_batch_size)` by having `fn fulfill()` keep calling `try_send()` until it fails because the channel is full. 

That is my intention but I'm not sure "If you're able to lock down `remote_batch_size` at creation time". But if that's a possibility, we're on the same page.

> You could even use `Receiver<Entity>` as the public type and skip making a `RemoteEntities` wrapper.

Technically. But I see the backend for this changing potentially. So I'd like to keep the wrapper for now.

> (`try_send()` looks like it just does some `Relaxed` and `Acquire` loads on failure, so it should be almost free. `is_full()` and `len()` seem to do `SeqCst` loads, which I believe do a full fence and are expensive, so you wouldn't want to use those.)

Yeah. This is a good end game. I didn't do that because `reserve_entities` needs a definite amount ahead of time. As you point out, we could do `alloc` for now, but that would prevent it from being flushed. Plus, IDK if we'll always have `&mut` access to `Entities`. I mean, probably, but we may need to `try_fulfill` more often than we `flush`. Maybe. I'll try some stuff out. We can always make a `&Entities` version later.

crates/bevy_ecs/src/entity/mod.rs
ElliottjPierce: > This reserves exactly `batch_size` entities per `flush`, regardless of how many requests there were. If we make enough async requests, it could take a while for them all to get fulfilled!

Yeah. It's not ideal. I don't anticipate this being a real issue though (with a good `batch_size`), but it's bugging me too.

> Rather than counting waiters, it might make sense to have a `request_count`, increment it _before_ popping, and do `reserve_entities(request_count.swap(0, Relaxed))`. If you pre-fill the queue with `batch_size` then that will refill the original buffer every time.

That's an excellent idea! Brilliant actually. It would keep the number reserved stable, and no reserve would be waiting for more than a single flush. I'll need to split `reserve_entities(request_count.swap(0, Relaxed))` up to help with branch prediction, but I really like this idea.

> (My proposal above has exactly the same problem, so maybe we can't use `async_channel::bounded`.)

Pros and cons. Might be possible to do both. A fast bounded channel for standard operations. But if it runs out, we increment `request_count` and fulfill on a boundless. That might be slightly over engineered, but it would work. I'll hold off for now, but I'll keep this in mind for a follow up.

crates/bevy_ecs/src/entity/mod.rs
ElliottjPierce: `alloc` prevents them from being flushed. Maybe we just let the asset system make them valid, but I'd kinda like to hand the asset server a blank entity rather than an invalid one. I have an idea here though. I'll get back to you.

ElliottjPierce (commented):
Thanks for the review @chescock. Brilliant as always!

crates/bevy_ecs/src/entity/mod.rs
```diff
@@ -515,6 +518,136 @@ impl<'a> core::iter::FusedIterator for ReserveEntitiesIterator<'a> {}
 // SAFETY: Newly reserved entity values are unique.
 unsafe impl EntitySetIterator for ReserveEntitiesIterator<'_> {}
 
+#[derive(Debug)]
+struct RemoteEntitiesInner {
+    recent_requests: AtomicU32,
```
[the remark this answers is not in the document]
ElliottjPierce: Done. Good call

crates/bevy_ecs/src/entity/mod.rs
```diff
@@ -515,6 +518,136 @@ impl<'a> core::iter::FusedIterator for ReserveEntitiesIterator<'a> {}
...
+    /// This happens when [`Entities`] are closed, dropped, etc while a [`RemoteEntities`] is trying to reserve from it.
+    #[error("A remote entity reserver tried to reserve an entity from a closed `Entities`.")]
+    Closed,
+}
+
+impl RemoteEntitiesInner {
+    #[inline]
+    fn fulfill(
+        entities: &mut Entities,
+        mut reserve_allocated: impl FnMut(Entity, &mut EntityLocation),
```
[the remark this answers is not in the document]
ElliottjPierce: Yup. That's a better name.

crates/bevy_ecs/src/entity/mod.rs
```diff
@@ -1217,6 +1377,66 @@ mod tests {
...
+        assert_eq!(
+            entities.remote.keep_hot.load(Ordering::Relaxed),
+            entities.remote.reserved.len() as u32
+        );
+    }
+
+    #[test]
+    #[cfg(feature = "std")]
+    fn remote_reservation() {
+        let mut entities = Entities::new();
```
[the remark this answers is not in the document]
ElliottjPierce: I re-worked the tests per your suggestions. I'm trying to mash up the state as much as possible. (testing async is hard). I generally trust `async-channel`'s implementation, but we might not be using it long term, so crazy tests are nice here. 

Marking as resolved for now, but point out any issues in the new tests!

crates/bevy_ecs/src/entity/mod.rs
```diff
@@ -515,6 +518,136 @@ impl<'a> core::iter::FusedIterator for ReserveEntitiesIterator<'a> {}
...
+
+impl RemoteEntitiesInner {
+    #[inline]
+    fn fulfill(
+        entities: &mut Entities,
+        mut reserve_allocated: impl FnMut(Entity, &mut EntityLocation),
+        keep_hot: u32,
+    ) {
+        let to_fulfill = entities.remote.recent_requests.swap(0, Ordering::Relaxed);
+        let current_hot = entities.remote.keep_hot.load(Ordering::Relaxed);
```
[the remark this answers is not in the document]
ElliottjPierce: > I don't understand why `entities.remote.keep_hot` needs to be an Atomic. Could we just move it into `Entities` and then handle it with just regular `u32`s?

For now, it doesn't. However, the extra 2 atomic ops are nothing compared to the like 10 atomic ops *per* entity to push onto the queue. (This is something I think a custom queue implementation can do much better. I did this for v4).

Given that it's not a performance concern, I like to keep the state about remote entities on the type itself rather than somewhere on `Entities`. We can change this later, but this could still come in handy if we want to do fulfillment with only `&Entities` in the future (maybe).

> Similarly is `keep_hot` the correct name for this? Maybe we should call it like `in_channel` or something? (they may have already been claimed by `recent_request` but I think it's close enough?)

That's a more precise name, so I changed the internals to use that. But the field we expose to users I've kept as "hot" just because the channel detail might change later.

crates/bevy_ecs/src/entity/mod.rs
```diff
@@ -515,6 +518,144 @@ impl<'a> core::iter::FusedIterator for ReserveEntitiesIterator<'a> {}
...
+    #[inline]
+    fn fulfill(
+        entities: &mut Entities,
+        mut init_allocated: impl FnMut(Entity, &mut EntityLocation),
+        in_channel: u32,
+    ) {
+        let to_fulfill = entities.remote.recent_requests.swap(0, Ordering::Relaxed);
+        let current_in_channel = entities.remote.in_channel.load(Ordering::Relaxed);
+        let should_reserve = (to_fulfill + in_channel).saturating_sub(current_in_channel); // should_reserve = to_fulfill + (in_channel - current_in_channel)
+        let new_in_channel = (current_in_channel + should_reserve).saturating_sub(to_fulfill); // new_in_channel = current_in_channel + (should_reserve - to_fulfill).
```
chescock, line 553: I *think* the second subtraction can never saturate, and could be an ordinary `-`.  `new_in_channel` will be more than `in_channel` if the first subtraction saturates, and equal to `in_channel` if it doesn't, but it can never be less than `in_channel`, so it can never be less than zero.  

(It's possible that this would be more clear if `should_reserve` were calculated using `checked_sub` and an `if`, especially since we never need to allocate entities if it overflows.)

chescock (approved):
Looks good!  I really like how small this wound up being!  

I left a few more nits, but nothing blocking (assuming `async_channel` gets updated, which seems promising).

crates/bevy_ecs/src/entity/mod.rs
```diff
@@ -515,6 +518,136 @@ impl<'a> core::iter::FusedIterator for ReserveEntitiesIterator<'a> {}
...
+pub enum RemoteReservationError {
+    /// This happens when [`Entities`] are closed, dropped, etc while a [`RemoteEntities`] is trying to reserve from it.
+    #[error("A remote entity reserver tried to reserve an entity from a closed `Entities`.")]
+    Closed,
+}
+
+impl RemoteEntitiesInner {
+    #[inline]
+    fn fulfill(
+        entities: &mut Entities,
```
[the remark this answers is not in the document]
ElliottjPierce: I had decided against that, but looking at it again, I don't see why not.  Done.

crates/bevy_ecs/src/entity/mod.rs
```diff
@@ -821,6 +970,8 @@ impl Entities {
         self.meta.clear();
         self.pending.clear();
         *self.free_cursor.get_mut() = 0;
+        self.remote.close();
```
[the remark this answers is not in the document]
ElliottjPierce: I prefer this behavior. What if a clear happens while an asset is loading? Now we have to make sure *every* s included entity is valid instead of just making sure one is. Right now, the minute it clears, all current remote reservers fail, which I think a more transparent way of error handling.

(But like you said, it doesn't really mater).

crates/bevy_ecs/src/entity/mod.rs
ElliottjPierce: > I _think_ the second subtraction can never saturate, and could be an ordinary `-`. `new_in_channel` will be more than `in_channel` if the first subtraction saturates, and equal to `in_channel` if it doesn't, but it can never be less than `in_channel`, so it can never be less than zero.

I agree. I thought this was the case when I made it but I was too lazy to do the proof lol. But I also like the redundancy.

> (It's possible that this would be more clear if `should_reserve` were calculated using `checked_sub` and an `if`, especially since we never need to allocate entities if it overflows.)

I agreed with this when I read it, but when I tried it, it felt more confusing.

ElliottjPierce:
Screenshot 2025-03-27 at 1 07 12 PM

I'm not sure why this keeps happening, where GitHub doesn't let me reply.

>Another way to approach this is to have separate structs for each side, like Sender and Receiver on channels. That keeps the remote reservation logic together, but still keeps the provider and client data separate. That would also remove the need to close() the channel explicitly, since it automatically closes when the last Sender is dropped.

I can see the appeal here, but I opted not to do this for a few reasons. 1) We might move away from the channel in the near future. 2) We don't necessarily want that closing behavior. and 3) We may want to be able to push to the channel apart from entities. (Ex: I reserved more entities than I need, so I'll push them instead of flushing and freeing.)

>I think the issue with using an atomic here is that the ordinary load followed by a store looks like a race condition, while if you had &mut then it would be obvious that nothing could change it in between. Performance shouldn't be an issue; relaxed load and store calls like this are no more expensive than non-atomic ones.

Agreed. Hopefully this is more clear now that fulfillment is in `&mut Entities`.

crates/bevy_ecs/src/entity/mod.rs
```diff
@@ -934,6 +1040,40 @@ impl Entities {
...
+        }
+    }
+
+    // we do this to hint to the compiler that the if branch is unlinkely to be taken.
+    #[cold]
+    fn force_remote_fulfill(&mut self, init_allocated: impl FnMut(Entity, &mut EntityLocation)) {
+        let in_channel = self.entities_hot_for_remote;
+        let mut init_allocated = init_allocated;
+        let to_fulfill = self.remote.recent_requests.swap(0, Ordering::Relaxed);
+        let current_in_channel = self.remote.in_channel.load(Ordering::Relaxed);
```
chescock, line 1056: Another option here is to store the (signed) difference between `in_channel` and `recent_requests`.  Then you only have one value to update!  You can use the existing `IdCursor` type for signed numbers.  This would reduce to something like:

```rust
if self.remote.net_in_channel.load(Ordering::Relaxed) < self.entities_hot_for_remote {
// ...
let old_net_in_channel = self.remote.net_in_channel
    .fetch_max(self.entities_hot_for_remote, Ordering::Relaxed);
let to_fulfill = self.entities_hot_for_remote - old_net_in_channel;
if to_fulfill > 0 {
```

crates/bevy_ecs/src/entity/mod.rs
```diff
@@ -934,6 +1040,40 @@ impl Entities {
...
+        let should_reserve = (to_fulfill + in_channel).saturating_sub(current_in_channel); // should_reserve = to_fulfill + (in_channel - current_in_channel)
+        let new_in_channel = current_in_channel + should_reserve - to_fulfill; // new_in_channel = current_in_channel + (should_reserve - to_fulfill).
+        self.remote
+            .in_channel
+            .store(new_in_channel, Ordering::Relaxed);
+
+        for _ in 0..should_reserve {
```
chescock, line 1063: Is it worth calling `self.reserve(should_reserve)`?

crates/bevy_ecs/src/entity/mod.rs
ElliottjPierce: I suppose we could, but I I'm not sure what the motivation would be. IIUC, it's the same number of atomic ops and the same amount of data stored. Actually its 8 bytes instead of 4, but still. I'm not opposed to it though, if there's a reason to do this that I'm just missing.

crates/bevy_ecs/src/entity/mod.rs
ElliottjPierce: Probably. That only does something 32 times total per `Entities`, so it's not really needed, but if we want to go forward here, I'll add it. Cuz why not?

crates/bevy_ecs/src/entity/mod.rs
```diff
@@ -1403,38 +1390,58 @@ mod tests {
...
         let mut entities = Entities::new();
         // Lower batch size so more waiting is tested.
         entities.entities_hot_for_remote = 16;
-        test_remote_reservation(&mut entities);
+        test_remote_reservation::<false>(&mut entities);
+    }
+
+    #[test]
+    #[cfg(feature = "std")]
+    fn remote_reservation_volitol_hot() {
```
andriyDev, line 1429: I'm not sure, but `volitol` -> `volatile`? Maybe another word could be `frequently_changed`?

andriyDev (approved):
At this point I'm fine with the code itself. Whether we take this route is another discussion. IOW, if we decide this is the PR we want to get merged, I think my review is valid, but that doesn't mean this is a "vote" for this PR :P My ideal (and I think yours as well) is fully parallelized entity reservation. But this is a good second.

crates/bevy_ecs/src/entity/mod.rs
chescock: Oh, I was just trying to simplify the `saturating_sub` arithmetic by having one counter to modify instead of two.

ElliottjPierce:
Heads up that the async channel issue is fixed now. See [here](https://github.com/smol-rs/async-channel/pull/106#event-17438251773).  IDK when the next release will be, but I can fix CI here when it's out. (Unless we do v9, which I tend to prefer.)

ElliottjPierce:
Closing in favor of #18670.

PR #18526, opened 2025-03-25
Fix `bevy_math/transform/input` Improper Inclusion

bushrat011899:
# Objective

Enabling `serialize`, `critical-section`, or `async-executor` would improperly enable `bevy_math`, `bevy_input`, and/or `bevy_transform`. This was caused by those crates previously being required but are now optional (gated behind `std` and/or `libm`).

## Solution

- Added `?` to features not intended to enable those crates

## Testing

- CI

### DIFF
--- a/crates/bevy_internal/Cargo.toml
+++ b/crates/bevy_internal/Cargo.toml
@@ -97,11 +97,11 @@ serialize = [
   "bevy_color?/serialize",
   "bevy_ecs/serialize",
   "bevy_image?/serialize",
-  "bevy_input/serialize",
-  "bevy_math/serialize",
+  "bevy_input?/serialize",
+  "bevy_math?/serialize",
   "bevy_scene?/serialize",
   "bevy_time/serialize",
-  "bevy_transform/serialize",
+  "bevy_transform?/serialize",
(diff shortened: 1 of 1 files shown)


IQuick143:
How does the `?` work? I'm not that familiar with Cargo's workings.

bushrat011899:
> How does the `?` work? I'm not that familiar with Cargo's workings.

It activates a feature on a dependency _if_ it's already included. Consider the below:

```toml
serialize = ["dep:serde", "foo/serialize"]
foo = ["dep:foo"]
```

You would assume that `serialize` enables the `serialize` feature on `foo`, and that `foo` enables the optional `foo` dependency. But actually, `foo/serialize` enables the `serialize` feature on `foo` _and_ enables the dependency `foo`. In contrast:

```toml
serialize = ["dep:serde", "foo?/serialize"]
foo = ["dep:foo"]
```

Now, `serialize` brings in the `serde` dependency, but it only enables `serialize` on `foo` _if_ `foo` is already included by some other feature, such as `foo`.

IQuick143 (approved):

mockersf (approved):

PR #18527, opened 2025-03-25
Fixes #13466, implementing resource_scope_by_id

### LINKED ISSUE #13466: Add `World::resource_scope_by_id`

I working on https://github.com/projectharmonia/bevy_replicon and we do a lot of type erasure internally for perfomance and a nice API.

Sometimes I need to access a resource mutably by ID and other resources just as reference. Since we don't have something like `World::resource_scope_by_id`, I do the inverse:
```rust
world.resource_scope(|world, registry: Mut<AppTypeRegistry>| {
    world.resource_scope(|world, entity_map: Mut<ServerEntityMap>| {
        world.resource_scope(|world, server_events: Mut<ServerEvents>| {
        });
    });
});
```

But it would be nicer to do it like this:
```rust
world.resource_scope_by_id(id, |world, resource| {
    let mut registry = world.resource::<AppTypeRegistry>();
    let mut entity_map = world.resource::<ServerEntityMap>();
    let mut server_events = world.resource::<ServerEvents>();
    // ...
});
```

Kugel-Blitz:
# Objective

- Fixes #13466, implementing resource_scope_by_id

## Testing

- Tested in the same way resource_scope was tested. Used the old precedent to test our new function by pulling the component id first and passing it in.

---

## Showcase

Allows mutable retrieval of resources using component id in a way they are then inserted back into the world. Similar to resource_scope.

```rust

// Insert the resources into the world with initial values
world.insert_resource(MyResource { value: 42 });
world.insert_resource(MyOtherResource { value: 100 });

// Get the component ID for MyResource
let component_id = world
    .components()
    .get_resource_id(TypeId::of::<MyResource>())
    .expect("Failed to get component ID for MyResource");

println!("component_id of MyResource =  {}", component_id.index());

// Use resource_scope_by_id to modify MyResource safely
world.resource_scope_by_id(component_id, |world: &mut World, mut res: Mut<MyResource>| {
    println!("Resource before: {}", res.value);
    res.value += 10;  // Modify the res
    println!("Resource after modification: {}", res.value);

    // Now, let's access another resource inside the closure
    // Get the component ID for MyOtherResource
    let other_component_id = world
           .components()
           .get_resource_id(TypeId::of::<MyOtherResource>())
           .expect("Failed to get component ID for MyOtherResource");

    // Access MyOtherResource using resource_scope_by_id
    world.resource_scope_by_id(other_component_id, |_world: &mut World, mut other_res: Mut<MyResource>| {
        println!("Second Resource before: {}", other_res.value);
        other_res.value += 20;  // Modify the second resource
        println!("Second Resource after modification: {}", other_res.value);
    });
});
```

### DIFF
--- a/crates/bevy_ecs/src/world/mod.rs
+++ b/crates/bevy_ecs/src/world/mod.rs
@@ -2761,6 +2761,7 @@ impl World {
     ///  /// # Example
     /// ```
     /// use bevy_ecs::prelude::*;
+    /// use std::any::TypeId;
     /// #[derive(Resource)]
     /// struct A(u32);
     /// #[derive(Component)]
@@ -2768,7 +2769,10 @@ impl World {
     /// let mut world = World::new();
     /// world.insert_resource(A(1));
     /// let entity = world.spawn(B(1)).id();
-    /// let component_id = A.id();
+    /// let component_id = world
+    /// .components()
+    /// .get_resource_id(TypeId::of::<A>())
+    /// .expect("Failed to get component ID for MyResource");
     /// world.resource_scope_by_id(component_id, |world: &mut World, mut a: Mut<A>| {
     ///     let b = world.get_mut::<B>(entity).unwrap();
     ///     a.0 += b.0;
--- a/crates/bevy_ecs/src/world/mod.rs
+++ b/crates/bevy_ecs/src/world/mod.rs
@@ -2758,14 +2758,16 @@ impl World {
     ///
     /// This enables safe simultaneous mutable access to both a resource and the rest of the [`World`].
     /// For more complex access patterns, consider using [`SystemState`](crate::system::SystemState).
-    ///  /// # Example
+    /// # Example
     /// ```
     /// use bevy_ecs::prelude::*;
     /// use std::any::TypeId;
+    /// 
     /// #[derive(Resource)]
     /// struct A(u32);
     /// #[derive(Component)]
     /// struct B(u32);
+    /// 
     /// let mut world = World::new();
     /// world.insert_resource(A(1));
     /// let entity = world.spawn(B(1)).id();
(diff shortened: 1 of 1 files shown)


crates/bevy_ecs/src/world/mod.rs
```diff
@@ -2753,6 +2753,95 @@ impl World {
         Some(result)
     }
 
+    /// Temporarily removes the requested resource from this [`World`] if it exists, runs custom user code,
+    /// then re-adds the resource before returning. Returns the result of the closure.
+    ///
+    /// This enables safe simultaneous mutable access to both a resource and the rest of the [`World`].
+    /// For more complex access patterns, consider using [`SystemState`](crate::system::SystemState).
+    ///  /// # Example
```
alice-i-cecile, line 2761:
```suggestion
    ///
    /// # Example
```
Two points: you had an extra `///`, and you were missing a newline before the heading. I suspect these are related!

crates/bevy_ecs/src/world/mod.rs
```diff
@@ -2753,6 +2753,95 @@ impl World {
...
+    /// ```
+    /// use bevy_ecs::prelude::*;
+    /// use std::any::TypeId;
+    /// #[derive(Resource)]
+    /// struct A(u32);
+    /// #[derive(Component)]
+    /// struct B(u32);
```
alice-i-cecile, line 2768:
```suggestion
    /// struct B(u32);
    ///
```

crates/bevy_ecs/src/world/mod.rs
```diff
@@ -2753,6 +2753,95 @@ impl World {
...
+    /// Temporarily removes the requested resource from this [`World`] if it exists, runs custom user code,
+    /// then re-adds the resource before returning. Returns the result of the closure.
+    ///
+    /// This enables safe simultaneous mutable access to both a resource and the rest of the [`World`].
+    /// For more complex access patterns, consider using [`SystemState`](crate::system::SystemState).
+    ///  /// # Example
+    /// ```
+    /// use bevy_ecs::prelude::*;
+    /// use std::any::TypeId;
+    /// #[derive(Resource)]
```
alice-i-cecile, line 2765:
```suggestion
    ///
    /// #[derive(Resource)]
```

crates/bevy_ecs/src/world/mod.rs
```diff
@@ -2753,6 +2753,95 @@ impl World {
...
+    /// struct B(u32);
+    /// let mut world = World::new();
+    /// world.insert_resource(A(1));
+    /// let entity = world.spawn(B(1)).id();
+    /// let component_id = world
+    /// .components()
+    /// .get_resource_id(TypeId::of::<A>())
+    /// .expect("Failed to get component ID for MyResource");
+    /// world.resource_scope_by_id(component_id, |world: &mut World, mut a: Mut<A>| {
+    ///     let b = world.get_mut::<B>(entity).unwrap();
```
alice-i-cecile, line 2777: You should assert that you can't access A inside of this scope :)

crates/bevy_ecs/src/world/mod.rs
```diff
@@ -2753,6 +2753,95 @@ impl World {
...
+    ///     a.0 += b.0;
+    /// });
+    /// assert_eq!(world.get_resource::<A>().unwrap().0, 2);
+    /// ```
+    ///
+    /// See also [`try_resource_scope_by_id`](Self::try_resource_scope_by_id).
```
alice-i-cecile, line 2783:
```suggestion
    /// See also [`try_resource_scope_by_id`](Self::try_resource_scope_by_id) for a non-panicking variant.
```

alice-i-cecile, line 2783: These docs would benefit from a `# Panics` section :)

alice-i-cecile (changes requested):
In general, we're trying to move away from implicit panics, and towards explicit error handling. I think that we should collapse this down into one method, which returns an Option here. That added safety + hassle + lower API surface is appropriate for a niche tool like this.

crates/bevy_ecs/src/world/mod.rs
```diff
@@ -2753,6 +2753,95 @@ impl World {
...
+    #[track_caller]
+    pub fn resource_scope_by_id<R: Resource, U>(
+        &mut self,
+        id: ComponentId,
+        f: impl FnOnce(&mut World, Mut<R>) -> U,
```
notmd, line 2788: The callback should accept PtrMut instead. If you already know the concrete type why would you still use dynamic API?

mhsalem36: Hi @notmd, if we use MutUntyped. How could we read the ptr's value? 
let mut value = unsafe { ptr.read::<R>() };

notmd: Hi sorry I totally forgot this. MutUntyped can construct directly with PtrMut, you don't need to read to concrete type. See https://dev-docs.bevyengine.org/bevy/ecs/change_detection/struct.MutUntyped.html

Issue #16896, opened 2024-12-19
Materials not displaying correctly in `wireframe` example


rparrett:
## Bevy version

main, bisected to #16799

## Relevant system information

SystemInfo { os: "Windows 11 Home", kernel: "22631", cpu: "13th Gen Intel(R) Core(TM) i7-13700KF", core_count: "16", memory: "31.8 GiB" }
AdapterInfo { name: "NVIDIA GeForce RTX 4080 SUPER", vendor: 4318, device: 9986, device_type: DiscreteGpu, driver: "NVIDIA", driver_info: "560.94", backend: Vulkan }

## What you did

`cargo run --example wireframe`

## What went wrong

Cubes and plane are all red. They should be different colors.

## Additional information

This seems to be vulkan specific or perhaps intermittent. The issue isn't present with `WGPU_BACKEND=dx12` or on my m1 mac.

First spotted by [bevy-example-runner](https://thebevyflock.github.io/bevy-example-runner/) in [this run](https://pixel-eagle.com/project/b25a040a-a980-4602-b90c-d480ab84076d/run/6004/compare/6000?screenshot=3D%20Rendering/wireframe.png).

pcwalton:
I investigated this and found that this is because of some strange semantics we were relying on. If you have multiple `Material` types on a mesh, then it will be drawn multiple times, one for each material type. The `Wireframe` component exploits this by adding a wireframe material to a mesh, regardless of other materials it might have already had. This breaks the assumption, used throughout rendering, that each main entity maps to at most one mesh instance. The failure we noticed is just the first time it broke, but `Wireframe` has been fundamentally broken for some time.

There are two main ways I can see to solve this:

1. Change everywhere we key off main entities in rendering to key off a (main entity, material type ID) pair instead.

2. Specify that including `Wireframe` at the same time as a material is undefined behavior. If you want to fill *and* draw a wireframe, apps must duplicate the entity, once for the fill and once for the wireframe.

I prefer (2), because (a) I don't want to slow down and bloat rendering just for this one use case; (b) fill-and-wireframe mode is usually used for debugging and so a little pain on the user's side is preferable to slowing down rendering; (c) the current semantics is incompatible with the long-term direction we want to go with for materials, where different shaders map to the same *dynamic material* type instead of having a separate type for each material. But opinions are welcomed here.

hukasu:
video on a modified `anisotropy` example

[...]

PR #18528, opened 2025-03-25
Update wireframe example to spawn materials without wireframes

krunchington:
# Objective

Addresses the .16 scope of #16896 

## Solution

- Duplicates the spawning of all entities without wireframes when they have a material, and without materials when they have a wireframe.

## Testing

- Ran the example using --features wayland on Ubuntu 24.04.2 LTS

---

## Showcase

image

## Migration Guide

N/A

### DIFF
--- a/examples/3d/wireframe.rs
+++ b/examples/3d/wireframe.rs
@@ -50,34 +50,35 @@ fn main() {
 
 /// set up a simple 3D scene
 /// Note that entities cannot be spawned with both a material and a wireframe, so for cubes we want
-/// to show both for, we spawn one with the Material and NoWireframe, the other we spawn with a
+/// to show both for, we spawn one with the Material and `NoWireframe`, the other we spawn with a
 /// Wireframe (or global Wireframe config), and no Material.
 fn setup(
     mut commands: Commands,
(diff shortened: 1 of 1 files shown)


examples/3d/wireframe.rs
```diff
@@ -66,22 +69,36 @@ fn setup(
...
         Transform::from_xyz(0.0, 0.5, 0.0),
+        NoWireframe,
+    ));
+    commands.spawn((
+        Mesh3d(meshes.add(Cuboid::default())),
+        Transform::from_xyz(0.0, 0.5, 0.0),
     ));
     // Green cube: Always renders a wireframe
     commands.spawn((
         Mesh3d(meshes.add(Cuboid::default())),
```
JMS55, line 80: We should reuse the same mesh handle between objects.

krunchington: Good catch!  I updated to use a single shared cube mesh for all instances since they all used the same default call before.

krunchington:
cc @IceSentry - does the fix for the 2d wireframes also cover this case?

IceSentry:
Yes, the fix will apply to both 2d and 3d so this PR also won't be necessary.

PR #18529, opened 2025-03-25
Update wireframe 2d example to fix material issue

### LINKED ISSUE #17737: Wireframe on `2d_shapes` example is just an outline
## Bevy version

6ed3c3274f9439a5c577c2685d7dbc2a91509b03

```
2025-02-08T01:20:23.636141Z  INFO bevy_diagnostic::system_information_diagnostics_plugin::internal: SystemInfo { os: "Linux (NixOS 24.11)", kernel: "6.6.69", cpu: "Intel(R) Core(TM) i7-14700KF", core_count: "20", memory: "31.2 GiB" }    
2025-02-08T01:20:23.872382Z  INFO bevy_render::renderer: AdapterInfo { name: "NVIDIA GeForce RTX 4060 Ti", vendor: 4318, device: 10243, device_type: DiscreteGpu, driver: "NVIDIA", driver_info: "565.77", backend: Vulkan }
```

## What you did

Ran `2d_shapes` example and toggled wireframes on.

## What went wrong

Wireframe was just an outline instead of showing the triangles.

## Additional information

6ed3c3274f9439a5c577c2685d7dbc2a91509b03
Image

83b725f41ef0c29fbe0a6412b8ff76d3c4994f55
Image

I bisected 3 times and got 3 different results, but it is known that wireframes are generally broken

> @pcwalton 
> wireframes are known broken because of the fact that they break 1:1 entity:mesh
https://discord.com/channels/691052431525675048/743663924229963868/1337598834330832988

also #16896

krunchington:
# Objective

Addresses the .16 scope of #16896 
Fixes #17737 

## Solution

- Duplicates the spawning of all entities without wireframes when they have a material, and without materials when they have a wireframe.

## Testing

- Ran the example using --features wayland on Ubuntu 24.04.2 LTS

---

## Showcase

image

## Migration Guide

N/A

### DIFF
--- a/examples/2d/wireframe_2d.rs
+++ b/examples/2d/wireframe_2d.rs
@@ -50,7 +50,7 @@ fn main() {
 
 /// set up a simple 2D scene
 /// Note that entities cannot be spawned with both a material and a wireframe, so for objects we want
-/// to show both for, we spawn one with the Material and NoWireframe2d, the other we spawn with a
+/// to show both for, we spawn one with the Material and `NoWireframe2d`, the other we spawn with a
 /// Wireframe (or global Wireframe config), and no Material.
 fn setup(
     mut commands: Commands,
--- a/examples/2d/wireframe_2d.rs
+++ b/examples/2d/wireframe_2d.rs
@@ -48,7 +48,10 @@ fn main() {
         .run();
 }
 
-/// Set up a simple 3D scene
+/// set up a simple 2D scene
+/// Note that entities cannot be spawned with both a material and a wireframe, so for objects we want
+/// to show both for, we spawn one with the Material and NoWireframe2d, the other we spawn with a
+/// Wireframe (or global Wireframe config), and no Material.
 fn setup(
     mut commands: Commands,
     mut meshes: ResMut<Assets<Mesh>>,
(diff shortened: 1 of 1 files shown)


examples/2d/wireframe_2d.rs
```diff
@@ -66,16 +69,25 @@ fn setup(
...
     // Rectangle: Follows global wireframe setting
+    let rectangle_mesh = meshes.add(Rectangle::new(100.0, 100.0));
     commands.spawn((
-        Mesh2d(meshes.add(Rectangle::new(100.0, 100.0))),
+        Mesh2d(rectangle_mesh.clone()),
         MeshMaterial2d(materials.add(Color::BLACK)),
         Transform::from_xyz(0.0, 0.0, 0.0),
+        NoWireframe2d,
     ));
+    commands.spawn((Mesh2d(rectangle_mesh), Transform::from_xyz(0.0, 0.0, 0.0)));
```
kristoff3r, line 79: I think it would be slightly clearer if these were spawned as children, then if this code is copy pasted the wireframe will move along with the original.

kristoff3r (commented):
This fixes the problem, but it's not ideal how global mode works now. Unless you literally only need wireframes in your scene you'll have to modify a bunch of entities anyway, and then you might as well just insert the wireframe component instead?

If this is what we end up going with for 0.16 we should document the behavior on the global setting both here and in the docs.

examples/2d/wireframe_2d.rs
kristoff3r: Also Transform is a required component for Mesh2d so it doesn't need to be added (and if it did you could use Transform::default() instead :P)

krunchington:
> Hey, so we already have a planned fix for this that is being worked on. It's not ready yet, but the goal is to have a separate render phase for wireframes that runs after the main pass.
> 
> Your PR fixes the example temporarily, but it doesn't really fix the underlying issue and it doesn't align with what we want as the end user api. We will have a fix for the 0.16 release.

Great, love to hear this!  Should we also close the other one for the 3d wireframe example? or does the fix not encompass that for the scope of .16?
https://github.com/bevyengine/bevy/pull/18528

PR #18530, opened 2025-03-25
Fix relationship macro for multiple named members fields

### LINKED ISSUE #18466: Relationship annotation does not allow more than additional field
## Bevy version

0.16.0-rc.1

## What you did

Added more than one non-relationship annotated field.

```rust
//child entity
#[derive(Component, Debug)] // <-- error here
#[relationship(relationship_target = ModifierCollectionRefs)]
pub struct ModifierInstance {
    #[relationship]
    pub collection_owning_entity: Entity,
    pub tag: ValueTag,
    pub source_context: ModifierContext,
    pub target_context: Option<ModifierContext>,
    pub value: ModifierValue
}

// parent entity
#[derive(Component, Debug, Default)]
#[relationship_target(relationship = ModifierInstance)]
pub struct ModifierCollectionRefs {
    modifiers: Vec<Entity>
}
```

## What went wrong

```
error: expected identifier, found `,`
  --> src/modifiers.rs:32:10
   |
32 | #[derive(Component, Debug)]
   |          ^^^^^^^^^
   |          |
   |          expected identifier
   |          while parsing this struct
```

I was expecting to be able to add more than one other non-relationship field on a relational component, however I cannot compile.

## Additional information

Workarounds:
- extract out other fields into another component. However, since I am not sure how to implement Default for relational components, in which case it is not obvious what the default parent should be, I have to couple the relational component to the data (that implements default), which may actually be the correct way to do it? Idk.

Assumed issue:
Relationship macro or relationship annotated field does not support more than one field in the related struct.

krunchington:
# Objective

Fixes #18466 

## Solution

Updated the macro generation pattern to place the comma in the correct place in the pattern.

## Testing

- Tried named and unnamed fields in combination, and used rust expand macro tooling to see the generated code and verify its correctness (see screenshots in example below)

---

## Showcase

Screenshot showing expanded macro with multiple named fields
image

Screenshot showing expanded macro with single unnamed field
image

## Migration Guide

n/a

### DIFF
--- a/crates/bevy_ecs/macros/Cargo.toml
+++ b/crates/bevy_ecs/macros/Cargo.toml
@@ -14,7 +14,6 @@ bevy_macro_utils = { path = "../../bevy_macro_utils", version = "0.16.0-dev" }
 syn = { version = "2.0.99", features = ["full", "extra-traits"] }
 quote = "1.0"
 proc-macro2 = "1.0"
-
 [lints]
 workspace = true
 
--- a/crates/bevy_ecs/macros/Cargo.toml
+++ b/crates/bevy_ecs/macros/Cargo.toml
@@ -15,9 +15,6 @@ syn = { version = "2.0.99", features = ["full", "extra-traits"] }
 quote = "1.0"
 proc-macro2 = "1.0"
 
-[dev-dependencies]
-bevy_ecs = { path = "../", version = "0.16.0-dev" }
-
 [lints]
 workspace = true
 
--- a/crates/bevy_ecs/macros/Cargo.toml
+++ b/crates/bevy_ecs/macros/Cargo.toml
@@ -14,6 +14,10 @@ bevy_macro_utils = { path = "../../bevy_macro_utils", version = "0.16.0-dev" }
 syn = { version = "2.0.99", features = ["full", "extra-traits"] }
 quote = "1.0"
 proc-macro2 = "1.0"
+
+[dev-dependencies]
+bevy_ecs = { path = "../", version = "0.16.0-dev" }
+
 [lints]
 workspace = true
 
--- a/crates/bevy_ecs/macros/src/component.rs
+++ b/crates/bevy_ecs/macros/src/component.rs
@@ -746,7 +746,7 @@ fn derive_relationship(
             #[inline]
             fn from(entity: #bevy_ecs_path::entity::Entity) -> Self {
                 Self {
-                    #(#members: core::default::Default::default(),),*
+                    #(#members: core::default::Default::default(),)*
                     #relationship_member: entity
                 }
             }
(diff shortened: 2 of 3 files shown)


alice-i-cecile (changes requested):
Good fix, but needs a regression test please :)

krunchington:
> Good fix, but needs a regression test please :)

Gotcha I'm not familiar with testing macros but I'll poke around in this tmrw night if I can get time!

alice-i-cecile:
For a test, just add some dead code in a struct that checks that this compiles. Or use a doc test, which can be a bit nicer.

krunchington:
> For a test, just add some dead code in a struct that checks that this compiles. Or use a doc test, which can be a bit nicer.

Got it, sounds straightforward!

CHATALOT1 (commented):
Having tested with my own similar error case, I believe the same fix is necessary on line 781 also to avoid the same bug occurring in the relationship_target macro.

Bleachfuel:
woops sorry, messed up when implementing it for enum's

Bleachfuel (approved):
Also for 781

krunchington:
> Having tested with my own similar error case, I believe the same fix is necessary on line 781 also to avoid the same bug occurring in the relationship_target macro.

Good catch!

krunchington:
Extra case handled in `derive_relationship_target`, and I added doc tests.  I didn't put much extra doc content in them though because they're private and so don't show on the public crates docsite I believe?  I also had to add `bevy_ecs` as a dev dependency to handle this case for the imports.

krunchington:
Hm, these CI failures don't look related to my change.  Did I inherit them by rebasing main?

crates/bevy_ecs/macros/Cargo.toml
```diff
@@ -14,6 +14,10 @@ bevy_macro_utils = { path = "../../bevy_macro_utils", version = "0.16.0-dev" }
 syn = { version = "2.0.99", features = ["full", "extra-traits"] }
 quote = "1.0"
 proc-macro2 = "1.0"
+
+[dev-dependencies]
+bevy_ecs = { path = "../", version = "0.16.0-dev" }
```
mockersf, line 19: you can't have cyclic dependencies

mockersf (changes requested):

mockersf:
> Hm, these CI failures don't look related to my change. Did I inherit them by rebasing against main?
> 
> They're all in the bevy_reflect crate which I didn't touch

I think it's because you're changing a macro crate which are compiled differently. And you can't depend on bevy_ecs in it. so you should probably move your examples to bevy_ecs

krunchington:
> I think it's because you're changing a macro crate which are compiled differently. And you can't depend on bevy_ecs in it. so you should probably move your examples to bevy_ecs

Thanks.  I opted for actual test content rather than adding a bunch of cases to doc tests that might just be noise.  Hoping this covers the cases appropriately but happy to adjust if folks have suggestions for alternate approaches!

alice-i-cecile (approved):

Issue #16586, opened 2024-12-01
Mesh Over-Allocates memory, causing program to crash


Asempere123123:
## Bevy version

0.15.0

## Relevant system information

```ignore
AdapterInfo { name: "NVIDIA GeForce RTX 2060", vendor: 4318, device: 7944, device_type: DiscreteGpu, driver: "NVIDIA", driver_info: "566.14", backend: Vulkan }
```

## What you did

Upgraded to Bevy 0.15. The code is available here: https://github.com/Asempere123123/space_game. The version for Bevy 0.14 is on the main branch, while the updated code for 0.15 is on the migration branch.

## What went wrong

There appears to be a memory leak causing Bevy to over-allocate. Running the game with Bevy 0.14 results in memory usage around 400 MB to 1 GB. However, with Bevy 0.15, memory usage continually increases until it consumes all available RAM, causing the program to crash.

## Additional information

I think the bug might be mesh-related. I have a procedural mesh generator that creates the mesh of a planet procedurally (making it more precise the closer you are). If you get close to the planet (really close, on the surface), the bug starts to occur. I don't think it's my mesher's fault, since the same code works well in 0.14.

### 0.14 (Expected):
![Image](https://github.com/user-attachments/assets/e61b5402-4936-4c5c-b64c-62559fcee689)

### 0.15:
![Image](https://github.com/user-attachments/assets/5daaa85d-a9c4-44ac-9211-8a0eaddd61c1)
(Memory usage keeps increasing until the program crashes.)

### Code Changes:
Here are the code changes: https://github.com/Asempere123123/space_game/compare/migration?expand=1

### Controls (for testing localy)
If anybody wants to try the game to test it, the controls are currently a bit weird:
At startup, there is an orbiting camera. You can orbit by right-clicking and zoom by scrolling.

Pressing 'T' switches to a free camera. However, the math is currently incorrect, so it only works near the equator.
I recommend zooming in orbit mode at startup and then switching to the free camera. This way, you end up close to the cube shown in the screenshots.

Since the planets are large, in free mode:
* Pressing 'O' doubles the movement speed.
* Pressing 'P' halves the movement speed.

hukasu:
you say that the problem is over allocation of the meshes, but i did not find that, i used this system to print the length and capacity of position and indices
```rust

fn check_mesh_size(
    mut local: Local<HashMap<Entity, (usize, usize, usize, usize)>>,
    primitives: Query<(Entity, &Mesh3d)>,
[...]

PR #18531, opened 2025-03-25
Have the mesh allocator handle modified meshes

JMS55:
# Objective
Fixes https://github.com/bevyengine/bevy/issues/16586.

## Solution
- Free meshes before allocating new ones (so hopefully the existing allocation is used, but it's not guaranteed since it might end up getting used by a smaller mesh first).
- Keep track of modified render assets, and have the mesh allocator free their allocations.
- Cleaned up some render asset code to make it more understandable, since it took me several minutes to reverse engineer/remember how it was supposed to work.

Long term we'll probably want to explicitly reusing allocations for modified meshes that haven't grown in size, or do delta uploads using a compute shader or something, but this is an easy fix for the near term.

## Testing
Ran the example provided in the issue. No crash after a few minutes, and memory usage remains steady.

### DIFF
--- a/crates/bevy_render/src/mesh/allocator.rs
+++ b/crates/bevy_render/src/mesh/allocator.rs
@@ -361,7 +361,7 @@ pub fn allocate_and_free_meshes(
     // Process removed meshes.
     mesh_allocator.free_meshes(&extracted_meshes);
 
-    // Process newly-added meshes.
+    // Process newly-added or modified meshes.
     mesh_allocator.allocate_meshes(
         &mesh_allocator_settings,
         &extracted_meshes,
--- a/crates/bevy_render/src/mesh/allocator.rs
+++ b/crates/bevy_render/src/mesh/allocator.rs
@@ -358,7 +358,7 @@ pub fn allocate_and_free_meshes(
     render_device: Res<RenderDevice>,
     render_queue: Res<RenderQueue>,
 ) {
-    // Process removed meshes.
+    // Process removed or modified meshes.
     mesh_allocator.free_meshes(&extracted_meshes);
 
     // Process newly-added or modified meshes.
@@ -607,6 +607,7 @@ impl MeshAllocator {
         }
     }
 
+    /// Frees allocations for meshes that were removed or modified this frame.
     fn free_meshes(&mut self, extracted_meshes: &ExtractedAssets<RenderMesh>) {
         let mut empty_slabs = <HashSet<_>>::default();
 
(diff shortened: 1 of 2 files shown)


crates/bevy_render/src/render_asset.rs
```diff
@@ -235,21 +241,30 @@ pub(crate) fn extract_render_asset<A: RenderAsset>(
...
                 )]
                 match event {
-                    AssetEvent::Added { id } | AssetEvent::Modified { id } => {
-                        changed_assets.insert(*id);
+                    AssetEvent::Added { id } => {
+                        needs_extracting.insert(*id);
+                    }
+                    AssetEvent::Modified { id } => {
+                        needs_extracting.insert(*id);
+                        modified.insert(*id);
```
mockersf, line 259: could we not push modified meshes to `needs_extracting` and chain it with `modified` when its used?

crates/bevy_render/src/mesh/allocator.rs
```diff
@@ -358,17 +358,17 @@ pub fn allocate_and_free_meshes(
     render_device: Res<RenderDevice>,
     render_queue: Res<RenderQueue>,
 ) {
-    // Process newly-added meshes.
+    // Process removed meshes.
```
IceSentry, line 361: Should this be modified to say it also removes modified meshes?

crates/bevy_render/src/mesh/allocator.rs
```diff
@@ -609,7 +609,14 @@ impl MeshAllocator {
 
     fn free_meshes(&mut self, extracted_meshes: &ExtractedAssets<RenderMesh>) {
```
IceSentry, line 610: A doc comment saying that it removes modified meshes would be nice

IceSentry (approved):
Minor nitpicks on docs and the suggestion by @mockersf is interesting but the overall code works and LGTM

crates/bevy_render/src/render_asset.rs
IceSentry: In this code block, just do `modified.insert(*id)`, then in the loop over `needs_extracting` later do `needs_extracting.chain(modified)`

IceSentry: Thinking about this more, it might not be possible because of the drain part?

JMS55: Oh, yeah won't be possible.

mockersf (approved):

HackerFoo (approved):

pcwalton (approved):
Seems fine. I agree that we probably want to reuse allocations at some point, but Seb's allocator is so fast that it probably doesn't actually matter much!

PR #18532, opened 2025-03-25
Add methods to work with dynamic immutable components

### LINKED ISSUE #16861: Add dynamic methods for working with immutable components
Would it make sense to have a `with_component_by_id` method that works with dynamic components?  

... oh, there's no `get_mut_assume_mutable_by_id`.  Should we add that?

_Originally posted by @chescock in https://github.com/bevyengine/bevy/pull/16668#discussion_r1873519978_

bushrat011899:
# Objective

- Fixes #16861

## Solution

- Added: 
  - `UnsafeEntityCell::get_mut_assume_mutable_by_id`
  - `EntityMut::get_mut_assume_mutable_by_id`
  - `EntityMut::get_mut_assume_mutable_by_id_unchecked`
  - `EntityWorldMut::into_mut_assume_mutable_by_id`
  - `EntityWorldMut::into_mut_assume_mutable`
  - `EntityWorldMut::get_mut_assume_mutable_by_id`
  - `EntityWorldMut::into_mut_assume_mutable_by_id`
  - `EntityWorldMut::modify_component_by_id`
  - `World::modify_component_by_id`
  - `DeferredWorld::modify_component_by_id`
- Added `fetch_mut_assume_mutable` to `DynamicComponentFetch` trait (this is a breaking change)

## Testing

- CI

---

## Migration Guide

If you had previously implemented `DynamicComponentFetch` you must now include a definition for `fetch_mut_assume_mutable`. In general this will be identical to `fetch_mut` using the relevant alternatives for actually getting a component.

---

## Notes

All of the added methods are minor variations on existing functions and should therefore be of low risk for inclusion during the RC process.

### DIFF
--- a/crates/bevy_ecs/src/world/entity_ref.rs
+++ b/crates/bevy_ecs/src/world/entity_ref.rs
@@ -824,6 +824,39 @@ impl<'w> EntityMut<'w> {
         unsafe { component_ids.fetch_mut(self.cell) }
     }
 
+    /// Returns [untyped mutable reference(s)](MutUntyped) to component(s) for
+    /// the current entity, based on the given [`ComponentId`]s.
+    /// Assumes the given [`ComponentId`]s refer to mutable components.
+    ///
+    /// **You should prefer to use the typed API [`EntityMut::get_mut_assume_mutable`] where
+    /// possible and only use this in cases where the actual component types
+    /// are not known at compile time.**
+    ///
+    /// Unlike [`EntityMut::get_mut_assume_mutable`], this returns untyped reference(s) to
+    /// component(s), and it's the job of the caller to ensure the correct
+    /// type(s) are dereferenced (if necessary).
+    ///
+    /// # Errors
+    ///
+    /// - Returns [`EntityComponentError::MissingComponent`] if the entity does
+    ///   not have a component.
+    /// - Returns [`EntityComponentError::AliasedMutability`] if a component
+    ///   is requested multiple times.
+    ///
+    /// # Safety
+    /// It is the callers responsibility to ensure that
+    /// - the provided [`ComponentId`]s must refer to mutable components.
+    #[inline]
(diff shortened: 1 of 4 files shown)


bushrat011899:
I've added this PR to the 0.16 milestone because the associated issue is. Happy to have the PR moved to 0.17, but I would suggest doing the same with the issue too.

alice-i-cecile (approved):
Deduplication plz 🥲  Thanks for tackling this.

MalekiRe (approved):
LGTM

bushrat011899:
Yeah I thought about trying to refactor the existing methods so there was more shared behaviour, but since it's so late into 0.16 I'd rather take the safe (and stupid) route and let 0.17 have an easy win.

crates/bevy_ecs/src/world/entity_ref.rs
```diff
@@ -1326,6 +1455,23 @@ impl<'w> EntityWorldMut<'w> {
         unsafe { self.into_unsafe_entity_cell().get_mut() }
     }
 
+    /// Consumes `self` and gets mutable access to the component of type `T`
+    /// with the world `'w` lifetime for the current entity.
+    /// Returns `None` if the entity does not have a component of type `T`.
```
makspll, line 1460: :nit: Not educated enough to know why this needs self, guessing losing self is not a big deal since `EntityRef` is `Copy`.  But if it's possible a `Result<Mut<'w,T>, Self>` return type would be more ergonomic

bushrat011899: There are two methods, `into_mut_assume_mutable` and `get_mut_assume_mutable`. `get` takes `&mut self` while into takes `self`. IMO, I do think the API should be changed to `Result` now that fallible systems are a thing, but the `Err` variant can't borrow data and still be compatible with `BevyError`, so it still couldn't support this suggestion.

Issue #18495, opened 2025-03-22
mesh_picking behaves erratically in `0.16-dev`


aloucks:
## Bevy version

main 0244a841b78ea9a995880fea6733da21a6c0fea2

## Relevant system information

I checked on both macOS and Windows.

## What you did

```
cargo run --example mesh_picking
```

## What went wrong

Picking doesn't fire depending on the angle of intersection or the normal of the intersection is computed incorrectly. It could also possibly be intersecting on the back face. 

## Additional information

https://github.com/user-attachments/assets/a0f27e48-f79f-4290-8d21-6cc2f8020c04

alice-i-cecile:
I suspect this is a bug introduced via https://github.com/bevyengine/bevy/pull/18232. We should consider reverting at this stage if we can't get a prompt fix.

aloucks:
confirmed fixed

PR #18232, opened 2025-03-10
(Adoped) Remove panics and optimise mesh picking

BD103:
_Note from BD103: this PR was adopted from #16148. The majority of this PR's description is copied from the original._

# Objective

Adds tests to cover various mesh picking cases and removes sources of panics.

It should prevent users being able to trigger panics in `bevy_picking` code via bad mesh data such as #15891, and is a follow up to my comments in [#15800 (review)](https://github.com/bevyengine/bevy/pull/15800#pullrequestreview-2361694213).

This is motivated by #15979

## Testing

Adds 8 new tests to cover `ray_mesh_intersection` code.

## Changes from original PR

I reverted the changes to the benchmarks, since that was the largest factor blocking it merging. I'll open a follow-up issue so that those benchmark changes can be implemented.

PR #18533, opened 2025-03-25
Fix `mesh_picking` not working due to mixing vertex and triangle indices.

IQuick143:
# Objective

- #18495 

## Solution

- The code in the PR #18232 accidentally used a vertex index as a triangle index, causing the wrong triangle to be used for normal computation and if the triangle went out of bounds, it would skip the ray-hit.
- Don't do that.

## Testing

- Run `cargo run --example mesh_picking`

### DIFF
--- a/crates/bevy_picking/src/mesh_picking/ray_cast/intersections.rs
+++ b/crates/bevy_picking/src/mesh_picking/ray_cast/intersections.rs
@@ -83,9 +83,10 @@ pub fn ray_mesh_intersection<I: TryInto<usize> + Clone + Copy>(
 
         indices
             .chunks_exact(3)
+            .enumerate()
             .fold(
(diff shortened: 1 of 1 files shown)


kristoff3r (approved):
I tested the example and this fixes the problem. 

It would be nice to have a test that triggers it, this code already has a bunch of tests but apparently none of them caught this issue.

IQuick143:
Yeah I agree, but personally idk what the test should be.
Maybe something that tests the normals on a sphere? (Those are easy to calculate and were wrong before I fixed them.)

tbillington (approved):
Nice catch! I did add more tests, obviously not enough though 😅

ElDiddi:
This fix also worked for me on Linux. (openSUSE Tumbleweed, wayland)

PR #18535, opened 2025-03-25
don't flip sprites twice

mockersf:
# Objective

- After #17041, sprite flipping doesn't work

## Solution

- Sprite flipping is applied twice:
https://github.com/bevyengine/bevy/blob/b6ccc2a2a0fac4d1a8d717920166b32f39b40cd9/crates/bevy_sprite/src/render/mod.rs#L766-L773
https://github.com/bevyengine/bevy/blob/b6ccc2a2a0fac4d1a8d717920166b32f39b40cd9/crates/bevy_sprite/src/render/mod.rs#L792-L799
- Keep one

### DIFF
--- a/crates/bevy_sprite/src/render/mod.rs
+++ b/crates/bevy_sprite/src/render/mod.rs
@@ -789,15 +789,6 @@ pub fn prepare_sprite_image_bind_groups(
                         );
                     }
 
-                    if extracted_sprite.flip_x {
-                        uv_offset_scale.x += uv_offset_scale.z;
-                        uv_offset_scale.z *= -1.0;
-                    }
-                    if extracted_sprite.flip_y {
-                        uv_offset_scale.y += uv_offset_scale.w;
-                        uv_offset_scale.w *= -1.0;
-                    }
-
                     let transform = extracted_sprite.transform.affine()
                         * Affine3A::from_scale_rotation_translation(
                             quad_size.extend(1.0),


bushrat011899 (approved):

bushrat011899:
Out of curiosity, the linked PR didn't flag as causing a rendering change. Do we know why? Might be good to diagnose so we can rely on that (really helpful!) test more.

kristoff3r (approved):

ickshonpe:
> Out of curiosity, the linked PR didn't flag as causing a rendering change. Do we know why? Might be good to diagnose so we can rely on that (really helpful!) test more.

Testbed_2d just doesn't have any sprite flipping. Just needs the sprite from it's sprite example module copied another three times flipped in each axis and both.

IceSentry (approved):

mockersf:
> Out of curiosity, the linked PR didn't flag as causing a rendering change. Do we know why? Might be good to diagnose so we can rely on that (really helpful!) test more.

I put in the testbed a few cases that I had in mind, but they're definitely not covering everything. We have to balance duration and coverage. We get a better coverage by running all the examples (https://thebevyflock.github.io/bevy-example-runner/index.html) but that's not on every PR ci as it takes 1.5 hour.

That said, adding sprite flipping shouldn't add any duration, so it's a good idea to do 👍

PR #18536, opened 2025-03-25
Add `many_morph_targets` stress test

greeble-dev:
## Objective

Add a stress test for morph targets, similar to `many_cubes` and `many_foxes`. This is also a useful smoke test as no other example has morph targets on multiple meshes.

## Solution

425571366-b043c16c-6e6a-491e-a0bd-5ece630d7bf8

Note that #18465 adds the same stress test. I'll remove it from that PR if this one lands.

## Testing

```sh
cargo run --example many_morph_targets

# Test different mesh counts. Default is 1024.
cargo run --example many_morph_targets -- --count 42
```

Tested on Win10/Vulkan/Nvidia, Wasm/WebGl/Chrome/Win10/Nvidia.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -3012,6 +3012,17 @@ category = "Stress Tests"
 wasm = true
 
 [[example]]
+name = "many_morph_targets"
+path = "examples/stress_tests/many_morph_targets.rs"
+doc-scrape-examples = true
+
+[package.metadata.example.many_morph_targets]
+name = "Many Morph Targets"
+description = "Simple benchmark to test rendering many meshes with animated morph targets."
+category = "Stress Tests"
+wasm = true
+
+[[example]]
 name = "many_glyphs"
 path = "examples/stress_tests/many_glyphs.rs"
 doc-scrape-examples = true
--- a/examples/README.md
+++ b/examples/README.md
@@ -502,6 +502,7 @@ Example | Description
 [Many Gizmos](../examples/stress_tests/many_gizmos.rs) | Test rendering of many gizmos
 [Many Glyphs](../examples/stress_tests/many_glyphs.rs) | Simple benchmark to test text rendering.
 [Many Lights](../examples/stress_tests/many_lights.rs) | Simple benchmark to test rendering many point lights. Run with `WGPU_SETTINGS_PRIO=webgl2` to restrict to uniform buffers and max 256 lights
+[Many Morph Targets](../examples/stress_tests/many_morph_targets.rs) | Simple benchmark to test rendering many meshes with animated morph targets.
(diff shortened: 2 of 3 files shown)


IceSentry (commented):
Did you measure if the performance of this scene is actually bottlenecked on the morph targets part and not just rendering?

greeble-dev:
I'm not sure I understand? The intent isn't to reproduce a particular bottleneck. I just wanted a benchmark for #18465 and figured it would have some general use. On my machine it's embarrassingly GPU bound, but that's fine as the CPU numbers are still meaningful.

IceSentry:
I mean, if we want to say it's a morph targets stress test we need to be sure that's the part that's actually stessing the system when you increase the amount of things on screen. If it just gets harder to run because of rendering then it's not as useful without relying on other tools to measure perf.

greeble-dev:
Aren't morph targets part of rendering since some of the work is in the vertex shader? I guess I could try to cut down other GPU work but I'm not sure that's there's much else in the test... maybe turning off any post-processing? Or put the meshes off-screen and disable frustum culling, but that makes it less useful as a visual test?

Also happy to run GPU profiles and report results. But I'll need advice on which profiler is best - I'm limited to Win10/Nvidia.

IceSentry:
Well, I mean, it could be the fragment shader or whatever taking up most of the time. I just want to make sure using this actually measures the right thing if we just look at the frametime. It's possible it's already the case to be clear, I just don't know. I just want to make sure the overhead of everything else isn't higher than the one of morph targets. As in, if you render all the meshes without any morph targets would it be faster. If the overhead is higher then it's just benchmarking rendering many meshes and the morph targets part becomes irrelevant.

greeble-dev:
I added more options and tested the frame times with various combinations. 

`--camera far` = camera zoomed out to minimise pixel shader costs - each mesh is about one pixel.
`--weights zero` = all weights zero, morph targets kinda disabled (pipeline the same, but vertex shader does no per-morph work except checking for zero weight).
`--weights tiny` = all weights tiny, so vertex shader does same work as `one` but pixel shader has roughly same cost as `zero`.
`--weights one` = all weights one.
`--weights animated` = mix of weights, some zero.

| | `--camera default` | `--camera far` |
|-|-|-|
| `--weights zero` | 7.5ms |  4.9ms |
| `--weights tiny` | 18.3ms | 13.4ms |
| `--weights one` | 24.1ms | 13.5ms |
| `--weights animated` | 12.8ms | 8.1ms |

The main thing is that `zero` and `tiny` show a clear difference when the only major difference is morph target specific vertex shader work.

Without a GPU profile I can't prove that the default setting (`animated`, default camera) is meaningful since I can't isolate the vertex shader work.

All tests were GPU bound. But I'm using a potato (Nvidia 1030) with a modern CPU, so someone with a real GPU might not be.

IceSentry (approved):

PR #18537, opened 2025-03-25
Add sprite flipping to `testbed_2d`'s sprite scene

ickshonpe:
# Objective

Add sprite flipping to `testbed_2d`'s sprite scene

## Solution

Draw the sprite flipped in each axis and both axes.

Changed the sprite to the rectangular bevy banner with text and made the images different colors.

## Testing
```
cargo run --example testbed_2d
```

image

### DIFF
--- a/crates/bevy_anti_aliasing/Cargo.toml
+++ b/crates/bevy_anti_aliasing/Cargo.toml
@@ -12,7 +12,6 @@ keywords = ["bevy"]
 trace = []
 webgl = []
 webgpu = []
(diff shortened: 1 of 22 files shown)


IceSentry (approved):

mockersf:
I was thinking adding at least 1 with a color too, what do you think?

alice-i-cecile (approved):
I agree that we should color one of them but I won't block on it :)

mockersf (approved):

ickshonpe:
> I was thinking adding at least 1 with a color too, what do you think?

Yep that's good and changed the image too, much easier now to identify which is flipped which way.

Issue #17891, opened 2025-02-17
`Image::from_buffer`'s public signature changes if debug assertions are enabled


rparrett:
Stumbled on this review comment that came a bit too late after a merge and seems.

> This is a public API, so this will break bevy plugins using it if the "wrong" build configuration is enabled by the user.
>
> `tracing` allows attaching this information to a `Span` instead, which is a better way to do this (the `warn_once!` will include the span's fields).
>
> _Originally posted by @SludgePhD in https://github.com/bevyengine/bevy/pull/11767#discussion_r1592399454_

PR #18411, opened 2025-03-19
[DRAFT] KTX2 Texture Support Updates

brianreavis:
# Objective

This PR expands Bevy’s supported KTX2 texture formats and transcoding capabilities (and .basis capabilities, as well). It also features some optimizations to improve speed and reduce memory pressure. 

### ETC1s/BasisLZ + (U)ASTC HDR Formats 

I forked and updated basis-universal-rs [here](https://github.com/naturalatlas/basis-universal-rs/commit/4d58d6e2b2ed76fdc5aa5f914d62d910b527c7bd):

* Synced to [basis_universal/2da6664](https://github.com/BinomialLLC/basis_universal/commit/2da6664eab159fc116100dce760db76cb34cfb47)
* Exposed new KTX2 decoder/transcoder functionality
* Added distinct feature flags for “encoding” and “transcoding” (for lighter builds)

### Faster Zstd Decompression

For decompression of Zstd supercompression, users can now pick between [ruzstd](https://crates.io/crates/ruzstd) using the `"zstd_rust"` feature (_safer_ and what bevy has been using) and [zstd](https://crates.io/crates/zstd) bindings via the `"zstd_native"` feature (_faster_). 

### Reduced Memory Pressure

When reading a KTX, we were allocating new buffers for a number of intermediate steps (super decompression, per-level transcoding, etc). This PR reduces the number of allocations.

## For Maintainers / Reviewers

> [!IMPORTANT]
> Before merging, unless/until the basis-universal updates are merged, I suggest forking [my fork of basis-universal-rs](https://github.com/naturalatlas/basis-universal-rs) pointing to it. Also, fork the submodule within it. This PR currently points to my fork.

> [!IMPORTANT] 
> We also need to get the [ktx2](https://github.com/BVE-Reborn/ktx2) folks (@cwfitzgerald?) to release what’s in trunk for https://github.com/BVE-Reborn/ktx2/pull/20. It has unreleased ASTC HDR constants that are used here.

## To Do

- [ ] Add a texture format example that effectively tests/shows a grid of all supported texture types

## Migration Guide

* If you have optional dependencies disabled and use functionality that requires zstd decompression (`"tonemapping_luts"` or `"ktx2"`), specify the zstd implementation you want to use: `"zstd_native"` (faster) or `"zstd_rust"` (safer)
* If using `CompressedImageSaver`, you’ll need to enable the feature `"basis-universal-compressor"` (this is no longer enabled by default for all builds)
[...]

PR #18538, opened 2025-03-25
Remove Image::from_buffer `name` argument (only present in debug "dds" builds)

brianreavis:
# Objective

- Fixes https://github.com/bevyengine/bevy/issues/17891
- Cherry-picked from https://github.com/bevyengine/bevy/pull/18411

## Solution

The `name` argument could either be made permanent (by removing the `#[cfg(...)]` condition) or eliminated entirely. I opted to remove it, as debugging a specific DDS texture edge case in GLTF files doesn't seem necessary, and there isn't any other foreseeable need to have it.

## Migration Guide

- `Image::from_buffer()` no longer has a `name` argument that's only present in debug builds when the `"dds"` feature is enabled. If you happen to pass a name, remove it.

### DIFF
--- a/crates/bevy_anti_aliasing/Cargo.toml
+++ b/crates/bevy_anti_aliasing/Cargo.toml
@@ -12,7 +12,6 @@ keywords = ["bevy"]
 trace = []
 webgl = []
 webgpu = []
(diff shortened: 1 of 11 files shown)


IceSentry (approved):
Assuming CI is fixed LGTM

alice-i-cecile (approved):

Issue #18515, opened 2025-03-24
Return the failed system parameter(s) in SystemParamValidationError


alice-i-cecile:
> It might be an interesting follow-up to try to pipe through some way to identify the failing parameter, like `ShortName::of::<P>`.

_Originally posted by @chescock in https://github.com/bevyengine/bevy/pull/18504#discussion_r2010243549_

PR #18541, opened 2025-03-25
Replace ValidationOutcome with Result

chescock:
# Objective

Make it easier to short-circuit system parameter validation.  

Simplify the API surface by combining `ValidationOutcome` with `SystemParamValidationError`.  

## Solution

Replace `ValidationOutcome` with `Result<(), SystemParamValidationError>`.  Move the docs from `ValidationOutcome` to `SystemParamValidationError`.  

Add a `skipped` field to `SystemParamValidationError` to distinguish the `Skipped` and `Invalid` variants.  

Use the `?` operator to short-circuit validation in tuples of system params.

### DIFF
--- a/crates/bevy_ecs/src/schedule/executor/single_threaded.rs
+++ b/crates/bevy_ecs/src/schedule/executor/single_threaded.rs
@@ -96,7 +96,7 @@ impl SystemExecutor for SingleThreadedExecutor {
                 let valid_params = match system.validate_param(world) {
                     Ok(()) => true,
                     Err(e) => {
-                        if !e.is_skipped() {
+                        if !e.skipped {
                             error_handler(
                                 e.into(),
                                 ErrorContext::System {
@@ -223,7 +223,7 @@ fn evaluate_and_fold_conditions(conditions: &mut [BoxedCondition], world: &mut W
             match condition.validate_param(world) {
                 Ok(()) => (),
                 Err(e) => {
-                    if !e.is_skipped() {
+                    if !e.skipped {
                         error_handler(
                             e.into(),
                             ErrorContext::System {
--- a/crates/bevy_ecs/src/schedule/executor/single_threaded.rs
+++ b/crates/bevy_ecs/src/schedule/executor/single_threaded.rs
@@ -10,7 +10,6 @@ use std::eprintln;
 use crate::{
     error::{default_error_handler, BevyError, ErrorContext},
     schedule::{is_apply_deferred, BoxedCondition, ExecutorKind, SystemExecutor, SystemSchedule},
-    system::{SystemParamValidationError, ValidationOutcome},
     world::World,
 };
 
@@ -95,18 +94,19 @@ impl SystemExecutor for SingleThreadedExecutor {
             let system = &mut schedule.systems[system_index];
             if should_run {
                 let valid_params = match system.validate_param(world) {
-                    ValidationOutcome::Valid => true,
-                    ValidationOutcome::Invalid => {
-                        error_handler(
-                            SystemParamValidationError.into(),
-                            ErrorContext::System {
-                                name: system.name(),
-                                last_run: system.get_last_run(),
-                            },
-                        );
+                    Ok(()) => true,
+                    Err(e) => {
+                        if !e.is_skipped() {
+                            error_handler(
+                                e.into(),
+                                ErrorContext::System {
+                                    name: system.name(),
+                                    last_run: system.get_last_run(),
+                                },
+                            );
+                        }
                         false
                     }
-                    ValidationOutcome::Skipped => false,
                 };
 
                 should_run &= valid_params;
@@ -221,18 +221,19 @@ fn evaluate_and_fold_conditions(conditions: &mut [BoxedCondition], world: &mut W
         .iter_mut()
         .map(|condition| {
             match condition.validate_param(world) {
-                ValidationOutcome::Valid => (),
-                ValidationOutcome::Invalid => {
-                    error_handler(
-                        SystemParamValidationError.into(),
-                        ErrorContext::System {
-                            name: condition.name(),
-                            last_run: condition.get_last_run(),
-                        },
-                    );
+                Ok(()) => (),
+                Err(e) => {
+                    if !e.is_skipped() {
+                        error_handler(
+                            e.into(),
+                            ErrorContext::System {
+                                name: condition.name(),
+                                last_run: condition.get_last_run(),
+                            },
+                        );
+                    }
                     return false;
                 }
-                ValidationOutcome::Skipped => return false,
             }
             __rust_begin_short_backtrace::readonly_run(&mut **condition, world)
         })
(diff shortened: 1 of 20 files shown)


crates/bevy_ecs/src/system/system_param.rs
```diff
@@ -2127,12 +2080,12 @@ macro_rules! impl_system_param_tuple {
...
                 world: UnsafeWorldCell,
-            ) -> ValidationOutcome {
-                // PERF: short-circuit to avoid wasted validation work
+            ) -> Result<(), SystemParamValidationError> {
                 let ($($param,)*) = state;
-                // Run validation on each parameter in the tuple,
-                // combining the results into a single `ValidationOutcome`.
-                ValidationOutcome::Valid$(.combine($param::validate_param($param, system_meta, world)))*
+                $(
+                    $param::validate_param($param, system_meta, world)?;
```
alice-i-cecile, line 2086: This is the tricky macro code that I was struggling to make short-circuit before.

alice-i-cecile (approved):
Semantically this makes sense to me: parma-validation is very Result-like. I also like cutting down on the code needed, and you managed to get the macro short-circuiting properly!

Good cleanup; thanks for chewing on this :)

crates/bevy_ecs/src/system/system_param.rs
```diff
@@ -2692,12 +2645,43 @@ unsafe impl SystemParam for FilteredResourcesMut<'_, '_> {
...
+    /// failures in validation should be considered a bug in the user's logic that must be immediately addressed (like [`Res`]).
+    ///
+    /// If `true`, the system should be skipped.
+    /// This is suitable for system params that are intended to only operate in certain application states, such as [`Single`].
+    pub fn is_skipped(&self) -> bool {
+        self.skipped
+    }
+
+    /// Constructs a `SystemParamValidationError` that skips the system.
+    pub fn skipped() -> Self {
```
ElliottjPierce, line 2676: This is a small thing, but could we make these constants instead? Or make them const functions with inline?(Though they're probably inlined already)

crates/bevy_ecs/src/system/system_param.rs
```diff
@@ -2692,12 +2645,43 @@ unsafe impl SystemParam for FilteredResourcesMut<'_, '_> {
...
+/// used by system executors to determine what to do with a system.
 ///
-/// Generated when [`SystemParam::validate_param`] returns `false`,
+/// Returned as an error from [`SystemParam::validate_param`],
 /// and handled using the unified error handling mechanisms defined in [`bevy_ecs::error`].
 #[derive(Debug, PartialEq, Eq, Clone, Display, Error)]
-pub struct SystemParamValidationError;
+pub struct SystemParamValidationError {
+    /// Whether the system should be skipped.
+    pub skipped: bool,
```
ElliottjPierce, line 2656: Do we need this to be public? If we're already exposing `is_skipped`, etc, I don't see the need for this. Making it private would make this easier to change in the future, and it forces users to use the better documented functions.
```suggestion
    skipped: bool,
```

crates/bevy_ecs/src/system/system_param.rs
```diff
@@ -2692,12 +2645,43 @@ unsafe impl SystemParam for FilteredResourcesMut<'_, '_> {
...
-/// An error that occurs when a system parameter is not valid.
+/// An error that occurs when a system parameter is not valid,
+/// used by system executors to determine what to do with a system.
 ///
-/// Generated when [`SystemParam::validate_param`] returns `false`,
+/// Returned as an error from [`SystemParam::validate_param`],
 /// and handled using the unified error handling mechanisms defined in [`bevy_ecs::error`].
 #[derive(Debug, PartialEq, Eq, Clone, Display, Error)]
-pub struct SystemParamValidationError;
+pub struct SystemParamValidationError {
```
ElliottjPierce, line 2654: This is a thiserror derive, but I don't see the `#[error(msg)]` attribute. Instead `Display` is derived directly. Nothing wrong with that, but is this intentional?

crates/bevy_ecs/src/schedule/executor/single_threaded.rs
```diff
@@ -221,18 +221,19 @@ fn evaluate_and_fold_conditions(conditions: &mut [BoxedCondition], world: &mut W
...
-                    error_handler(
-                        SystemParamValidationError.into(),
-                        ErrorContext::System {
-                            name: condition.name(),
-                            last_run: condition.get_last_run(),
-                        },
-                    );
+                Ok(()) => (),
+                Err(e) => {
+                    if !e.is_skipped() {
```
ElliottjPierce, line 226: This is repeated in a bunch of places. It was before too, but it might be worth pulling out into an associated function on the error type. Definitely not a problem, but worth mentioning.

ElliottjPierce (approved):
A few suggestions but nothing blocking approval.

There's other ways to do the short circuiting in the macro, but I agree that `Result` is semantically valuable here. Might be worth a type alias to provide documentation for what the result means more directly, but that's not a huge deal.

crates/bevy_ecs/src/system/system_param.rs
alice-i-cecile: I'd prefer to keep this pub and kill the helper method if we're deduplicating.

alice-i-cecile:
I'll give you a couple days to respond to feedback here, but I'm going to merge this before the next release candidate regardless <3

crates/bevy_ecs/src/system/system_param.rs
chescock: I don't want to make them actual constants because I'm anticipating we'll want to add parameters (#18515) and it should be a smoother transition if they're already functions.  

Making them `const fn` is a good idea, though!  I'll do that.

crates/bevy_ecs/src/system/system_param.rs
chescock: Whoops, I had started with it `pub` and failed to remove it after I added the method.  Good catch!  

But it sounds like we want it `pub`, so I'll go remove the helper methods instead.  (I guess I still have some old C# habits.)

crates/bevy_ecs/src/system/system_param.rs
chescock: No clue, sorry.  It was like that when I got here, and I'm not familliar with `thiserror`.

crates/bevy_ecs/src/schedule/executor/single_threaded.rs
chescock: Yup, I said the same thing on the original PR :) https://github.com/bevyengine/bevy/pull/18504#discussion_r2010212849  It sounded like it was tricky to solve at the moment, so I didn't try.

crates/bevy_ecs/src/system/system_param.rs
ElliottjPierce: Works for me!

crates/bevy_ecs/src/system/system_param.rs
ElliottjPierce: That's fine. We can revisit this later if we need to.

crates/bevy_ecs/src/schedule/executor/single_threaded.rs
ElliottjPierce: Well, it's only a little duplication, and it may be subject to change anyway. So we can come back to it if it's final form still has duplication and it becomes boiler plate.

PR #3404, opened 2021-12-20
Make the default window title more descriptive

DJMcNab:
# Objective

- Default bevy windows don't have a very useful name:
image
- We can include the binary name in the default title:
image

## Solution

- Get the current binary title when using a `WindowDescriptor` which doesn't have a name set.
- When I timed the function which gets this, in debug mode it takes `26.8µs` (i.e. `0.0268ms`)

Note: We already transitively depend on `once_cell` in `bevy_window`, via `ahash`.

PR #18553, opened 2025-03-26
Use current_exe for default window title

IceSentry:
# Objective

- We currently default to "App" for the window title, it would be nice if examples had more descriptive names

## Solution

- Use `std::env::current_exe` to try to figure out a default title. If it's not present. Use "App".

## Testing

- I tested that examples that set a custom title still use the custom title and that examples without a custom title use the example name

---

### Showcase

Here's the 3d_scene example:

image

### Notes

Here's a previous attempt at this from a few years ago https://github.com/bevyengine/bevy/pull/3404

There's some relevant discussion in there, but cart's decision was to default to "App" when no name was found.

### DIFF
--- a/crates/bevy_window/src/window.rs
+++ b/crates/bevy_window/src/window.rs
@@ -25,10 +25,18 @@ use crate::VideoMode;
 ///
 /// It will try to use the name of the current exe if possible, otherwise it defaults to "App"
 static DEFAULT_WINDOW_TITLE: LazyLock<String> = LazyLock::new(|| {
-    std::env::current_exe()
-        .ok()
-        .and_then(|current_exe| Some(format!("{}", current_exe.file_stem()?.to_string_lossy())))
-        .unwrap_or_else(|| "App".to_string())
+    #[cfg(feature = "std")]
+    {
+        std::env::current_exe()
+            .ok()
+            .and_then(|current_exe| Some(format!("{}", current_exe.file_stem()?.to_string_lossy())))
+            .unwrap_or_else(|| "App".to_string())
+    }
+    
+    #[cfg(not(feature = "std"))]
+    {
+        "App".to_string()
+    }
 });
 
 /// Marker [`Component`] for the window considered the primary window.
--- a/crates/bevy_window/src/window.rs
+++ b/crates/bevy_window/src/window.rs
@@ -1,7 +1,7 @@
 use alloc::string::ToString;
 use alloc::{format, string::String};
 use core::num::NonZero;
-use std::sync::LazyLock;
+use bevy_platform_support::sync::LazyLock;
 
 use bevy_ecs::{
     entity::{Entity, EntityBorrow},
--- a/crates/bevy_window/src/window.rs
+++ b/crates/bevy_window/src/window.rs
@@ -1,6 +1,5 @@
 use alloc::string::ToString;
 use alloc::{format, string::String};
-use bevy_platform_support::sync::LazyLock;
 use core::num::NonZero;
 
 use bevy_ecs::{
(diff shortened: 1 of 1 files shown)


crates/bevy_window/src/window.rs
```diff
@@ -1,5 +1,7 @@
-use alloc::{borrow::ToOwned, string::String};
+use alloc::string::ToString;
+use alloc::{format, string::String};
 use core::num::NonZero;
+use std::sync::LazyLock;
```
bushrat011899, line 4:
```suggestion
use bevy_platform_support::sync::LazyLock;
```

crates/bevy_window/src/window.rs
```diff
@@ -19,6 +21,16 @@ use bevy_reflect::{ReflectDeserialize, ReflectSerialize};
...
 use crate::VideoMode;
 
+/// Default string used for the window title.
+///
+/// It will try to use the name of the current exe if possible, otherwise it defaults to "App"
+static DEFAULT_WINDOW_TITLE: LazyLock<String> = LazyLock::new(|| {
+    std::env::current_exe()
+        .ok()
+        .and_then(|current_exe| Some(format!("{}", current_exe.file_stem()?.to_string_lossy())))
+        .unwrap_or_else(|| "App".to_string())
```
bushrat011899, line 31:
```suggestion
    #[cfg(feature = "std")]
    {
        std::env::current_exe()
            .ok()
            .and_then(|current_exe| Some(format!("{}", current_exe.file_stem()?.to_string_lossy())))
            .unwrap_or_else(|| "App".to_string())
    }
    
    #[cfg(feature = "std")]
    {
        "App".to_string()
    }
```

bushrat011899 (changes requested):
Just a pair of minor changes for `no_std` support!

IceSentry:
Thank you! I didn't realize this was already in platform_support. That was easier than expected.

bushrat011899:
All good! Probably need to run `cargo fmt` but that _should_ work now

IceSentry:
(I only requested a review to remove the request for change. No need to actually review if you don't want to.)

bushrat011899 (approved):
Nice change! Simple and direct improvement IMO.

alice-i-cecile (approved):

mockersf (approved):

IceSentry:
@mockersf any suggestions how to fix the failing CI check for the example showcase patches? I'm not sure where/how to update them.

mockersf:
huh didn't saw it on mobile.

You replaced the `to_owned()` by `to_string()` and changed the import. You should do either the same change in the patch https://github.com/bevyengine/bevy/blob/7a8dfdaec288a7c964a1cb6062a67f2f0bb3e76d/tools/example-showcase/window-settings-wasm.patch#L13
or continue using `to_owned()`

crates/bevy_window/src/window.rs
```diff
@@ -19,6 +20,24 @@ use bevy_reflect::{ReflectDeserialize, ReflectSerialize};
...
+/// It will try to use the name of the current exe if possible, otherwise it defaults to "App"
+static DEFAULT_WINDOW_TITLE: LazyLock<String> = LazyLock::new(|| {
+    #[cfg(feature = "std")]
+    {
+        std::env::current_exe()
+            .ok()
+            .and_then(|current_exe| Some(format!("{}", current_exe.file_stem()?.to_string_lossy())))
+            .unwrap_or_else(|| "App".to_owned())
+    }
+
```
aloucks, line 34: You could also use the `file` macro to get the source filename. That might be better in both cases and then you wouldn't need special handling:

https://doc.rust-lang.org/core/macro.file.html

IceSentry: It will just give the name of window.rs file, not the name of the example. 

Also, even if it was the name of the top level file, this won't work for users with a cargo project that uses main.rs.

PR #18555, opened 2025-03-26
Improved Require Syntax

cart:
# Objective

Requires are currently more verbose than they need to be. People would like to define inline component values. Additionally, the current `#[require(Foo(custom_constructor))]` and `#[require(Foo(|| Foo(10))]` syntax doesn't really make sense within the context of the Rust type system. #18309 was an attempt to improve ergonomics for some cases, but it came at the cost of even more weirdness / unintuitive behavior. Our approach as a whole needs a rethink.

## Solution

Rework the `#[require()]` syntax to make more sense. This is a breaking change, but I think it will make the system easier to learn, while also improving ergonomics substantially:

```rust
#[derive(Component)]
#[require(
    A, // this will use A::default()
    B(1), // inline tuple-struct value
    C { value: 1 }, // inline named-struct value
    D::Variant, // inline enum variant
    E::SOME_CONST, // inline associated const
    F::new(1), // inline constructor
    G = returns_g(), // an expression that returns G
    H = SomethingElse::new(), // expression returns SomethingElse, where SomethingElse: Into<H> 
)]
struct Foo;
```

## Migration Guide

Custom-constructor requires should use the new expression-style syntax:

```rust
// before
#[derive(Component)]
#[require(A(returns_a))]
struct Foo;

// after
#[derive(Component)]
#[require(A = returns_a())]
struct Foo;
```

Inline-closure-constructor requires should use the inline value syntax where possible:

```rust
// before
#[derive(Component)]
#[require(A(|| A(10))]
struct Foo;

// after
#[derive(Component)]
#[require(A(10)]
struct Foo;
```

In cases where that is not possible, use the expression-style syntax:

```rust
// before
#[derive(Component)]
#[require(A(|| A(10))]
struct Foo;

// after
#[derive(Component)]
#[require(A = A(10)]
struct Foo;
```

### DIFF
--- a/crates/bevy_ecs/macros/src/component.rs
+++ b/crates/bevy_ecs/macros/src/component.rs
@@ -568,6 +568,9 @@ impl Parse for Require {
         let mut path = input.parse::<Path>()?;
         let mut last_segment_is_lower = false;
         let mut is_constructor_call = false;
+        // Use the case of the type name to check if it's an enum
+        // This doesn't match everything that can be an enum according to the rust spec
+        // but it matches what clippy is OK with
         let is_enum = {
             let mut first_chars = path
                 .segments
--- a/crates/bevy_ecs/macros/src/component.rs
+++ b/crates/bevy_ecs/macros/src/component.rs
@@ -3,13 +3,13 @@ use proc_macro2::{Span, TokenStream as TokenStream2};
 use quote::{format_ident, quote, ToTokens};
 use std::collections::HashSet;
 use syn::{
-    parenthesized,
+    braced, parenthesized,
     parse::Parse,
     parse_macro_input, parse_quote,
     punctuated::Punctuated,
     spanned::Spanned,
-    token::{Comma, Paren},
-    Data, DataEnum, DataStruct, DeriveInput, Expr, ExprCall, ExprClosure, ExprPath, Field, Fields,
+    token::{Brace, Comma, Paren},
+    Data, DataEnum, DataStruct, DeriveInput, Expr, ExprCall, ExprPath, Field, FieldValue, Fields,
     Ident, LitStr, Member, Path, Result, Token, Type, Visibility,
 };
 
@@ -207,17 +207,7 @@ pub fn derive_component(input: TokenStream) -> TokenStream {
                 );
             });
             match &require.func {
-                Some(RequireFunc::Path(func)) => {
-                    register_required.push(quote! {
-                        components.register_required_components_manual::<Self, #ident>(
-                            required_components,
-                            || { let x: #ident = #func().into(); x },
-                            inheritance_depth,
-                            recursion_check_stack
-                        );
-                    });
-                }
-                Some(RequireFunc::Closure(func)) => {
+                Some(func) => {
                     register_required.push(quote! {
                         components.register_required_components_manual::<Self, #ident>(
                             required_components,
@@ -478,12 +468,7 @@ enum StorageTy {
 
 struct Require {
     path: Path,
-    func: Option<RequireFunc>,
-}
-
-enum RequireFunc {
-    Path(Path),
-    Closure(ExprClosure),
+    func: Option<TokenStream2>,
 }
 
 struct Relationship {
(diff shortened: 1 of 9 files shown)


cart:
Note that I have included this in the 0.16 milestone (despite it being pretty late in the game) because #18309 introduces behaviors and syntax that I don't think should land in 0.16 (both in light of these incoming changes, which are incompatible, and because I think those behaviors are too confusing).

krunchington:
> G = returns_g(), // an expression that returns F

I think this is a typo, it returns G?

NthTensor:
This would be the biggest breaking change to be included in the RC after start, correct? To my knowledge the only other that was semi-breaking was the system-parameter validation fixes, which did sort of need to go in to fix the small mess left by 0.15. 

Is there a reason why we shouldn't revert https://github.com/bevyengine/bevy/pull/18309 instead and moving forward with this in the 0.17 cycle?

crates/bevy_ecs/macros/src/component.rs
```diff
@@ -580,25 +565,68 @@ fn parse_component_attr(ast: &DeriveInput) -> Result<Attrs> {
...
+        let mut last_segment_is_lower = false;
+        let mut is_constructor_call = false;
+        let is_enum = {
+            let mut first_chars = path
+                .segments
+                .iter()
+                .rev()
+                .filter_map(|s| s.ident.to_string().chars().next());
+            if let Some(last) = first_chars.next() {
+                if last.is_uppercase() {
```
krunchington, line 578: By convention, EnumItems should start with an upper case character, but technically in the language spec they are IDENTIFIERs which don't have to.  Clippy gets mad, but you can make an enum variant with a lowercase starting character.

cart: Yup I'm aware. Sadly this is the best we can do in proc macros to disambiguate between a type path and an enum path, and I personally think it is good enough. If someone actually needs a lowercase enum, they can just use expression syntax (but I suspect the population of people that needs this will be of size zero).

alice-i-cecile:
@Wuketuke can I get your review here please?

cart:
@NthTensor 
> Is there a reason why we shouldn't revert https://github.com/bevyengine/bevy/pull/18309 instead and moving forward with this in the 0.17 cycle?

That is also a reasonable path forward. Only reason not to is making this available faster, which when weighed against the cost of breaking people across RCs, _probably_ doesn't win out. Although I think its close. The "cost paid" here is libraries using closure or custom-function require syntax, that have already done the work to update, will need to do the work in the next RC, rather than in 0.17. The scope there feels small enough that it might be worth it. For "programming pattern" things, frontloading breakage is valuable, as it means more people learn the new pattern / more learning materials use it / etc.

crates/bevy_ecs/macros/src/component.rs
alice-i-cecile: Can you add a quick comment to that effect here? I agree with your analysis (no one uses lowercase enums), but it's a helpful note.

crates/bevy_ecs/src/component.rs
```diff
@@ -160,62 +160,87 @@ use thiserror::Error;
...
 /// # use bevy_ecs::prelude::*;
 /// #[derive(Component)]
-/// #[require(C(init_c))]
+/// #[require(
+///     B(1), // tuple structs
+///     C { value: 1 }, // named-field structs
+///     D::One, // enum variants
+///     E::ONE, // associated consts
+///     F::new(1) // constructors
+/// )]
```
alice-i-cecile, line 173: Very nice little demo; I like the comments.

alice-i-cecile (approved):
Code is clear and well-commented. I much prefer the new syntax. I share Nth's reluctance to ship this sort of thing post RC, but I agree with your analysis on why it's worth it here (and made the same call myself for param validation error handling).

alice-i-cecile:
I would really prefer not to have to run the migration guide / release note generator again: when this is merged can you make a manual PR to the bevy-website repo for this? Or just remind me to do it :)

crates/bevy_ecs/macros/src/component.rs
krunchington: Agreed, makes sense especially if there's no clear alternative. As Alice pointed out, docs would be valuable for that edge case.

NthTensor:
Your reasoning makes sense, and the bug surface area here seems reasonably small.   Hopefully there won't be too many more of these this cycle.

crates/bevy_ecs/macros/src/component.rs
```diff
@@ -580,25 +565,68 @@ fn parse_component_attr(ast: &DeriveInput) -> Result<Attrs> {
...
-        let path = input.parse::<Path>()?;
-        let func = if input.peek(Paren) {
-            let content;
-            parenthesized!(content in input);
-            if let Ok(func) = content.parse::<ExprClosure>() {
-                Some(RequireFunc::Closure(func))
+        let mut path = input.parse::<Path>()?;
+        let mut last_segment_is_lower = false;
+        let mut is_constructor_call = false;
+        let is_enum = {
```
mockersf, line 571:
```suggestion
        // Use the case of the type name to check if it's an enum
        // This doesn't match everything that can be an enum according to the rust spec
        // but it matches what clippy is OK with
        let is_enum = {
```

cart:
> I would really prefer not to have to run the migration guide / release note generator again (it's really awful to do after going through and deleting unneeded entries): when this is merged can you make a manual PR to the bevy-website repo for this? Or just remind me to do it :)

Sure thing!

Shatur:
Ah, never mind, I have to use `=`:
```rust
#[require(
    Node = Node {
        margin: UiRect::all(Val::Px(BUTTON_MARGIN)),
        ..Default::default()
    }
)]
```

cart:
Found an easy fix. I'll put it up!

PR #18563, opened 2025-03-26
don't wait during publishing

mockersf:
# Objective

- Publishing takes a long time
- There's a 20 second wait between crates to not hit the rate limit on crates.io

## Solution

- Our rate limit has been increased by the crates.io team, don't wait anymore!


alice-i-cecile (approved):
YOLO let's see if this works! Thanks crates.io.

PR #18567, opened 2025-03-27
Reverts

### LINKED ISSUE #18010: Need some way to nest labeled assets

andriyDev:
# Objective

- Fixes #18010.

## Solution

- Revert the offending PRs! These are #15481 and #18013. We now no longer get an error if there are duplicate subassets.
- In theory we could untangle #18013 from #15481, but that may be tricky, and may still introduce regressions. To avoid this worry (since we're already in RC mode), I am just reverting both.

## Testing

- This is just a revert.

---

## Migration Guide

<Remove the migration guides for #15481 and #18013>

I will make a PR after this is merged.


andriyDev:
I should mention that there is more context for why we are doing a revert https://github.com/bevyengine/bevy/pull/18295#issuecomment-2756024250

cart (approved):
Thanks for being understanding here. Excited to iterate on this with you!

mockersf (approved):

PR #18572, opened 2025-03-27
Fix animation transitions affecting other entities

### LINKED ISSUE #18557: Animation transitions don't apply weights correctly
https://github.com/bevyengine/bevy/blob/7a8dfdaec288a7c964a1cb6062a67f2f0bb3e76d/crates/bevy_animation/src/transition.rs#L121 

This line causes all entities to share the same remainder, distorting all animations when a single one is transitioning. It should be brought into the loop so it only affects the entity in transition.

greeble-dev:
## Objective

 Fix #18557.

## Solution

As described in the bug, `remaining_weight` should have been inside the loop.

## Testing

Locally changed the `animated_mesh_control` example to spawn multiple meshes and play different transitions.

### DIFF
--- a/crates/bevy_animation/src/transition.rs
+++ b/crates/bevy_animation/src/transition.rs
@@ -118,8 +118,9 @@ pub fn advance_transitions(
     // is divided between all the other layers, eventually culminating in the
     // currently-playing animation receiving whatever's left. This results in a
     // nicely normalized weight.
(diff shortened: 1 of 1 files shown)


kristoff3r (approved):
It would be great with a regression test as well

mockersf (approved):

PR #18576, opened 2025-03-27
Use desired language in migration guide template

alice-i-cecile:
# Objective

Both reading and writing migration guides is easier when the language is standardized.

However, migration guide authors don't have clear guidelines for the tone and phrasing to use.

## Solution

Communicate this information to authors by creating stub text with a clear and polite standard style.

We could instead write a style guide, but turning style guides into a writing style is slower and much harder than simply filling in the blanks. While style guides are a good fit for more free-form writing, they don't work well for the very mechanical and dry migration guides.


release-content/migration_guides_template.md
```diff
@@ -17,10 +17,19 @@ Aim to communicate:
 
 Keep it short and sweet:
 
-- Use bullet points and make sure it's searchable.
+- What, then why, then how to migrate.
+- Some helpful standardized phrases:
+  - `OldType` is now `NewType`. Replace all references and imports.
+  - The `Struct::method()` method now requires an additional `magnitude: f32` argument
+  - `Enum` has a new variant, `Enum::NewVariant`, which must be handled during `match` statements
```
NthTensor, line 24:
```suggestion
  - The `Struct::method()` method now requires an additional `magnitude: f32` argument.
  - `Enum` has a new variant, `Enum::NewVariant`, which must be handled during `match` statements.
```

doup (approved):

release-content/migration_guides_template.md
```diff
@@ -17,10 +17,19 @@ Aim to communicate:
...
+  - `OldType` is now `NewType`. Replace all references and imports.
+  - The `Struct::method()` method now requires an additional `magnitude: f32` argument.
+  - `Enum` has a new variant, `Enum::NewVariant`, which must be handled during `match` statements.
+  - The `Type::method` method has been removed. Use `Type::other_method` instead.
+  - The `crate::old_module` module is now `crate::new_module`. Update your imports.
+  - `function` now returns `Option<String>`, instead of `String`.
+- Make sure it's searchable by directly naming the types and methods involved.
+- Use backticks for types, methods and modules (e.g. `Vec<T>` or `core::mem::swap`).
+- Use bullet points to explain complex changes.
 - Avoid headings. If you must, use only level-two headings.
```
doup, line 31: It's unrelated to this PR, but I though @NthTensor said this would be handled with tooling. So we don't impose any heading level, the user start using H1, then H2, etc. and the export/import tool will convert to the proper level. The only limit we should impose is "don't use more than X levels" (I think it was worded like that at some point).

NthTensor: The markdown linter wants us to use level-2 because of the title in the front matter. No idea how to change that.

Pnoenix (approved):
These changes look fine. Maybe add another file or a section in here, that showcases what a complete migration guide file looks like?

alice-i-cecile:
> These changes look fine. Maybe add another file or a section in here, that showcases what a complete migration guide file looks like?

In most cases, looking at sibling migration guides will be best for this :)

PR #18578, opened 2025-03-27
Required Components: pass through all tokens in {} and () syntax

cart:
# Objective

#18555 added improved require syntax, but inline structs didn't support `..Default::default()` syntax (for technical reasons we can't parse the struct directly, so there is manual logic that missed this case).

## Solution

When a `{}` or `()` section is encountered for a required component, rather than trying to parse the fields directly, just pass _all_ of the tokens through. This ensures no tokens are dropped, protects us against any future syntax changes, and optimizes our parsing logic (as we're dropping the field parsing logic entirely).

### DIFF
--- a/crates/bevy_ecs/macros/src/component.rs
+++ b/crates/bevy_ecs/macros/src/component.rs
@@ -9,8 +9,8 @@ use syn::{
     punctuated::Punctuated,
     spanned::Spanned,
     token::{Brace, Comma, Paren},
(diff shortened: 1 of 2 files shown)


mockersf:
could you add a new case using the syntax that is now allowed in the doc examples or in a test?

alice-i-cecile (approved):

mockersf (approved):

PR #18579, opened 2025-03-27
Fix misleading documentation of Main schedule

### LINKED ISSUE #18562: Misleading information in the documentation of the Main schedule
The [documentation](https://docs.rs/bevy/latest/bevy/app/struct.Main.html) of the Main schedule says that on its first run it wil run `PreStartup`, `Startup` and `PostStartup`. Actually, it starts with `StateTransition` as can be seen [here](https://github.com/bevyengine/bevy/blob/main/crates/bevy_state/src/app.rs#L311).

inact1v1ty:
# Objective

Fixes #18562.

## Solution

- Specified that `StateTransition` is actually run before `PreStartup`.
- Specified consequences of this and how to actually run systems before any game logic regardless of state.
- Updated docs of `StateTransition` to reflect that it is run before `PreStartup` in addition to being run after `PreUpdate`.

## Testing

- `cargo doc`
- `cargo test --doc`

### DIFF
--- a/crates/bevy_app/src/main_schedule.rs
+++ b/crates/bevy_app/src/main_schedule.rs
@@ -15,9 +15,10 @@ use bevy_ecs::{
 /// By default, it will run the following schedules in the given order:
 ///
 /// On the first run of the schedule (and only on the first run), it will run:
-/// * [`StateTransition`]
-///      * This means that [`OnEnter(ExampleState)`] will be called *before* [`PreStartup`]
-///        if `ExampleState` is the starting state of the app.
+/// * [`StateTransition`] [^1]
+///      * This means that [`OnEnter(MyState::Foo)`] will be called *before* [`PreStartup`]
+///        if `MyState` was added to the app with `MyState::Foo` as the initial state,
+///        as well as [`OnEnter(MyComputedState)`] if it `compute`s to `Some(Self)` in `MyState::Foo`.
 ///      * If you want to run systems before any state transitions, regardless of which state is the starting state,
 ///        for example, for registering required components, you can add your own custom startup schedule
 ///        before [`StateTransition`]. See [`MainScheduleOrder::insert_startup_before`] for more details.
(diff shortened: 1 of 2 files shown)


alice-i-cecile (approved):

crates/bevy_app/src/main_schedule.rs
```diff
@@ -15,6 +15,12 @@ use bevy_ecs::{
 /// By default, it will run the following schedules in the given order:
 ///
 /// On the first run of the schedule (and only on the first run), it will run:
+/// * [`StateTransition`]
+///      * This means that [`OnEnter(ExampleState)`] will be called *before* [`PreStartup`]
+///        if `ExampleState` is the starting state of the app.
```
benfrankel, line 20:
```suggestion
///        if `ExampleState` was added to the app with that initial state.
```
"starting state of the app" is confusing because apps are not associated with one global state type.

benfrankel (changes requested):
I'll also note that the `StateTransition` mentions in the docs for the `Main` schedule are technically not entirely correct, because a user could use `bevy_app` without `bevy_state`. However, they were already incorrect before this PR so it's fine.

crates/bevy_app/src/main_schedule.rs
inact1v1ty: Hmm. I like your suggestion, but I don't think this wording is correct, too, because of computed states: if a computed state happens to be derived from the initial enum variant of a state, its `OnEnter` would run here, too. Do you have something on your mind that will both capture this case and make the sentence not confusing?

inact1v1ty: I think this is important to highlight here, as it can be a point of confusion. In our game, for example, we make use of `SubStates` and `ComputedStates` extremely heavily, I assume dozens of `ComputedStates` by the moment all features will be implemented - so most of the times it is `OnEnter(SomeComputedState)`.

benfrankel: Sure, maybe change `ExampleState` to `MyState::Foo` and then say "if `MyState::Foo` is the initial value of `MyState`" or similar.

inact1v1ty:
@benfrankel I have reworded according to your suggestion, but adding info about `ComputedStates`, and also left a footnote telling that `StateTransition` comes only with `bevy_state` 🙂

benfrankel (approved):

PR #18582, opened 2025-03-27
Web support for atmosphere

mate-h:
# Objective

Add web support to atmosphere by gating dual source blending and using a macro to determine the target platform. 
The main objective of this PR is to ensure that users of Bevy's atmosphere feature can also run it in a web-based context where WebGPU support is enabled. 

## Solution

- Make use of the `#[cfg(not(target_arch = "wasm32"))]` macro to gate the dual source blending, as this is not (yet) supported in web browsers.
- Rename the function `sample_sun_illuminance` to `sample_sun_radiance` and move calls out of conditionals to ensure the shader compiles and runs in both native and web-based contexts.
- Moved the multiplication of the transmittance out when calculating the sun color, because calling the `sample_sun_illuminance` function was causing issues in web. Overall this results in cleaner code and more readable.

## Testing

- Tested by building a wasm target and loading it in a web page with Vite dev server using `mate-h/bevy-webgpu` repo template.
- Tested the native build with `cargo run --example atmosphere` to ensure it still works with dual source blending.

---

## Showcase

Screenshots show the atmosphere example running in two different contexts:

atmosphere-web-showcase

### DIFF
--- a/crates/bevy_pbr/src/atmosphere/functions.wgsl
+++ b/crates/bevy_pbr/src/atmosphere/functions.wgsl
@@ -281,7 +281,7 @@ fn sample_sun_radiance(ray_dir_ws: vec3<f32>) -> vec3<f32> {
     let r = view_radius();
     let mu_view = ray_dir_ws.y;
     let shadow_factor = f32(!ray_intersects_ground(r, mu_view));
-    var sun_illuminance = vec3(0.0);
+    var sun_radiance = vec3(0.0);
     for (var light_i: u32 = 0u; light_i < lights.n_directional_lights; light_i++) {
         let light = &lights.directional_lights[light_i];
         let neg_LdotV = dot((*light).direction_to_light, ray_dir_ws);
@@ -289,9 +289,9 @@ fn sample_sun_radiance(ray_dir_ws: vec3<f32>) -> vec3<f32> {
         let pixel_size = fwidth(angle_to_sun);
         let factor = smoothstep(0.0, -pixel_size * ROOT_2, angle_to_sun - SUN_ANGULAR_SIZE * 0.5);
         let sun_solid_angle = (SUN_ANGULAR_SIZE * SUN_ANGULAR_SIZE) * 4.0 * FRAC_PI;
-        sun_illuminance += ((*light).color.rgb / sun_solid_angle) * factor * shadow_factor;
+        sun_radiance += ((*light).color.rgb / sun_solid_angle) * factor * shadow_factor;
     }
-    return sun_illuminance * view.exposure;
+    return sun_radiance;
 }
 
 // TRANSFORM UTILITIES
--- a/crates/bevy_pbr/src/atmosphere/functions.wgsl
+++ b/crates/bevy_pbr/src/atmosphere/functions.wgsl
@@ -277,7 +277,7 @@ fn sample_local_inscattering(local_atmosphere: AtmosphereSample, ray_dir: vec3<f
 
 const SUN_ANGULAR_SIZE: f32 = 0.0174533; // angular diameter of sun in radians
 
-fn sample_sun_illuminance(ray_dir_ws: vec3<f32>, transmittance: vec3<f32>) -> vec3<f32> {
+fn sample_sun_radiance(ray_dir_ws: vec3<f32>) -> vec3<f32> {
     let r = view_radius();
     let mu_view = ray_dir_ws.y;
     let shadow_factor = f32(!ray_intersects_ground(r, mu_view));
@@ -291,7 +291,7 @@ fn sample_sun_illuminance(ray_dir_ws: vec3<f32>, transmittance: vec3<f32>) -> ve
         let sun_solid_angle = (SUN_ANGULAR_SIZE * SUN_ANGULAR_SIZE) * 4.0 * FRAC_PI;
         sun_illuminance += ((*light).color.rgb / sun_solid_angle) * factor * shadow_factor;
     }
-    return sun_illuminance * transmittance * view.exposure;
+    return sun_illuminance * view.exposure;
 }
 
 // TRANSFORM UTILITIES
--- a/crates/bevy_pbr/src/atmosphere/mod.rs
+++ b/crates/bevy_pbr/src/atmosphere/mod.rs
@@ -46,8 +46,6 @@ use bevy_reflect::{std_traits::ReflectDefault, Reflect};
 use bevy_render::{
     extract_component::UniformComponentPlugin,
     render_resource::{DownlevelFlags, ShaderType, SpecializedRenderPipelines},
-    renderer::RenderDevice,
-    settings::WgpuFeatures,
 };
 use bevy_render::{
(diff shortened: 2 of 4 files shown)


ecoskey (changes requested):
Looks good! few nits though

crates/bevy_pbr/src/atmosphere/render_sky.wgsl
```diff
@@ -33,15 +35,24 @@ fn main(in: FullscreenVertexOutput) -> RenderSkyOutput {
...
         transmittance = sample_transmittance_lut(r, mu);
         inscattering += sample_sky_view_lut(r, ray_dir_as);
-        inscattering += sample_sun_illuminance(ray_dir_ws.xyz, transmittance);
+        inscattering += sun_radiance * transmittance;
     } else {
         let t = ndc_to_camera_dist(vec3(uv_to_ndc(in.uv), depth));
         inscattering = sample_aerial_view_lut(in.uv, t);
         transmittance = sample_transmittance_lut_segment(r, mu, t);
     }
+    #ifdef DUAL_SOURCE_BLENDING
```
JMS55, line 51: Btw formatting convention for ifdefs is that they go as far left as possible, with no identation (but their contents should be intended as if the ifdefs weren't there).

crates/bevy_pbr/src/atmosphere/resources.rs
```diff
@@ -367,7 +372,11 @@ impl SpecializedRenderPipeline for RenderSkyBindGroupLayouts {
                     blend: Some(BlendState {
                         color: BlendComponent {
                             src_factor: BlendFactor::One,
+                            #[cfg(not(target_arch = "wasm32"))]
                             dst_factor: BlendFactor::Src1,
+                            // Dual source blending is not supported on Web
```
JMS55, line 377: Just WebGL2, no? Afaik WebGPU does support dual source blending.

mate-h: You might be right about that, I saw this article: https://developer.chrome.com/blog/new-in-webgpu-130#dual_source_blending

I did get an error last time I tested with Chrome, maybe it needs that special directive at the top? 
```
enable dual_source_blending;
```

mate-h: This is the error I get otherwise:
```
Features(DUAL_SOURCE_BLENDING) is not enabled for this backend
```

mate-h: https://github.com/gfx-rs/wgpu/blob/700a2e8ced7966d5a89c0635734e8c325397bbb6/CHANGELOG.md#webgpu-compliant-dual-source-blending-feature

crates/bevy_pbr/src/atmosphere/functions.wgsl
```diff
@@ -277,7 +277,7 @@ fn sample_local_inscattering(local_atmosphere: AtmosphereSample, ray_dir: vec3<f
 
 const SUN_ANGULAR_SIZE: f32 = 0.0174533; // angular diameter of sun in radians
 
-fn sample_sun_illuminance(ray_dir_ws: vec3<f32>, transmittance: vec3<f32>) -> vec3<f32> {
+fn sample_sun_radiance(ray_dir_ws: vec3<f32>) -> vec3<f32> {
```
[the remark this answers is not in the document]
mate-h: We should keep "radiance" here because this shader calculates raw light transport before any perceptual/photometric transformations. While Bevy's PBR lights use photometric units (lux/lumens), the sky radiance calculation happens in linear HDR space, with perceptual considerations only applied later during the tonemapping post-process. Luminance would imply we're already working with human-eye weighted values, which isn't the case at this stage.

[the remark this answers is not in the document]
ecoskey: hmm ok. My thinking was this code seems to take `(*light).color.rgb` (in lux) and divide it by the sun's solid angle, which would give luminance in nits 🤔. I'll trust your judgement though, I may have to read the pbr book again lol

mate-h:
Looks like CI is passing now, it's ready for a final review!

JMS55 (changes requested):
Code looks fine (although I don't know if the math is correct as I'm not involved in any of the atmosphere stuff).

Could you please add docs indicating that it will look different on web due to the lack of dual source blending?

ecoskey (approved):

crates/bevy_pbr/src/atmosphere/mod.rs
```diff
@@ -25,6 +25,12 @@
 //! at once is untested, and might not be physically accurate. These may be
 //! integrated into a single module in the future.
 //!
+//! This plugin works in web browsers too! While the visuals might look slightly
+//! different from native platforms, that's because web browsers handle some advanced
+//! graphics features differently. Specifically, when calculating how light travels
+//! through the atmosphere, we use a simpler averaging technique instead of the more
+//! complex blending operations. This difference will be resolved in a future release.
+//!
```
JMS55, line 33:
```suggestion
//! On web platforms, atmosphere rendering will look slightly different. Specifically, when calculating how light travels
//! through the atmosphere, we use a simpler averaging technique instead of the more
//! complex blending operations. This difference will be resolved for WebGPU in a future release.
//!
```

Imo something like this.

BenjaminBrienen (approved):
image
```
2025-04-06T19:41:04.502979Z  INFO bevy_diagnostic::system_information_diagnostics_plugin::internal: SystemInfo { os: "Linux (Pop!_OS 22.04)", kernel: "6.12.10-76061203-generic", cpu: "AMD Ryzen 9 7950X 16-Core Processor", core_count: "16", memory: "61.9 GiB" }    
2025-04-06T19:41:04.654616Z  INFO bevy_render::renderer: AdapterInfo { name: "NVIDIA GeForce RTX 3070", vendor: 4318, device: 9348, device_type: DiscreteGpu, driver: "NVIDIA", driver_info: "565.77", backend: Vulkan }
2025-04-06T19:41:05.200937Z  INFO bevy_winit::system: Creating new window atmosphere (0v1)
2025-04-06T19:41:05.201035Z  INFO winit::platform_impl::linux::x11::window: Guessed window scale factor: 2
2025-04-06T19:41:05.443827Z  WARN bevy_render::view::window: Couldn't get swap chain texture after configuring. Cause: 'The underlying surface has changed, and therefore the swap chain must be updated'
2025-04-06T19:41:05.468849Z  WARN bevy_render::view::window: Couldn't get swap chain texture after configuring. Cause: 'The underlying surface has changed, and therefore the swap chain must be updated'
2025-04-06T19:41:07.824744Z  WARN bevy_render::view::window: Couldn't get swap chain texture after configuring. Cause: 'The underlying surface has changed, and therefore the swap chain must be updated'
2025-04-06T19:41:24.846299Z  INFO bevy_window::system: No windows are open, exiting    
2025-04-06T19:41:24.847742Z  INFO bevy_winit::system: Closing window 0v1
```

superdump (approved):

PR #18585, opened 2025-03-28
Create chainable `run_if_dyn`

hukasu:
# Objective

Most methods from `IntoScheduleConfigs` are chainable, bun not `run_if_dyn`

## Solution

Make `run_if_dyn` chainable

## Testing

`cargo run -p ci`

### DIFF
--- a/crates/bevy_ecs/src/schedule/config.rs
+++ b/crates/bevy_ecs/src/schedule/config.rs
@@ -233,7 +233,7 @@ impl<T: Schedulable<Metadata = GraphInfo, GroupMetadata = Chain>> ScheduleConfig
     ///
     /// This is useful if you have a run condition whose concrete type is unknown.
     /// Prefer `run_if` for run conditions whose type is known at compile time.
-    pub fn run_if_dyn(&mut self, condition: BoxedCondition) {
+    pub fn run_if_dyn_inner(&mut self, condition: BoxedCondition) {
         match self {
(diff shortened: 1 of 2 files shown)


crates/bevy_ecs/src/schedule/config.rs
```diff
@@ -233,7 +233,7 @@ impl<T: Schedulable<Metadata = GraphInfo, GroupMetadata = Chain>> ScheduleConfig
     ///
     /// This is useful if you have a run condition whose concrete type is unknown.
     /// Prefer `run_if` for run conditions whose type is known at compile time.
-    pub fn run_if_dyn(&mut self, condition: BoxedCondition) {
+    pub fn run_if_dyn_inner(&mut self, condition: BoxedCondition) {
```
alice-i-cecile, line 236: Should this be pub?

alice-i-cecile, line 236: I don't understand why this needs to exist. Can't you just discard the unused ScheduleConfigs?

hukasu:
don't know if this is really a `Bug`, more `Usability`

crates/bevy_ecs/src/schedule/config.rs
urben1680: I don't use this but `in_set_inner` and I prefer more than less of these methods being public because they enable easier complex configuration building where you don't want to play tennis with owned values.

PR #18589, opened 2025-03-28
Transform Propagation Optimization: Static Subtree Marking

### LINKED ISSUE #18255: Transform propagation ignoring earlier branches

### LINKED ISSUE #18365: Investigate static tree marking for optimizing transform propagation
> We should investigate redoing #18093: it's fundamentally a sensible strategy to optimize transform propagation, and I'm not convinced that all of its performance gains are due to this bug.

_Originally posted by @alice-i-cecile in https://github.com/bevyengine/bevy/issues/18363#issuecomment-2730406404_

aevyrie:
# Objective

- Optimize static scene performance by marking unchanged subtrees.
- [bef0209](https://github.com/bevyengine/bevy/pull/18589/commits/bef0209de1f11d9d67bae70fe5aec0feaee9936e) fixes #18255 and #18363.
- Closes #18365 
- Includes change from #18321

## Solution

- Mark hierarchy subtrees with dirty bits to avoid transform propagation where not needed
- This causes a performance regression when spawning many entities, or when the scene is entirely dynamic.
- This results in massive speedups for largely static scenes.
- In the future we could allow the user to change this behavior, or add some threshold based on how dynamic the scene is?

## Testing

- Caldera Hotel scene


Vrixyz:
🤩

- There should be a follow up to make a unit test (echoing comment from https://github.com/bevyengine/bevy/issues/18255#issuecomment-2730543075).

mockersf (approved):

alice-i-cecile (approved):
Thanks for tracking down the bug and getting the fix in <3 I wasn't pleased to have to revert this!

PR #18591, opened 2025-03-28
Fix wesl in wasm and webgl2

mockersf:
# Objective

- feature `shader_format_wesl` doesn't compile in Wasm
- once fixed, example `shader_material_wesl` doesn't work in WebGL2

## Solution

- remove special path handling when loading shaders. this seems like a way to escape the asset folder which we don't want to allow, and can't compile on android or wasm, and can't work on iOS (filesystem is rooted there)
- pad material so that it's 16 bits. I couldn't get conditional compilation to work in wesl for type declaration, it fails to parse
- the shader renders the color `(0.0, 0.0, 0.0, 0.0)` when it's not a polka dot. this renders as black on WebGPU/metal/..., and white on WebGL2. change it to `(0.0, 0.0, 0.0, 1.0)` so that it's black everywhere

### DIFF
--- a/assets/shaders/custom_material.wesl
+++ b/assets/shaders/custom_material.wesl
@@ -6,7 +6,8 @@ struct VertexOutput {
 }
 
 struct CustomMaterial {
-    time: f32,
+    // Needed for 16-bit alignment on WebGL2
+    time: vec4<f32>,
 }
 
 @group(2) @binding(0) var<uniform> material: CustomMaterial;
@@ -15,5 +16,5 @@ struct CustomMaterial {
 fn fragment(
     mesh: VertexOutput,
 ) -> @location(0) vec4<f32> {
-    return make_polka_dots(mesh.uv, material.time);
+    return make_polka_dots(mesh.uv, material.time.x);
 }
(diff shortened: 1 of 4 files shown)


IceSentry (approved):

assets/shaders/custom_material.wesl
```diff
@@ -6,7 +6,8 @@ struct VertexOutput {
 }
 
 struct CustomMaterial {
-    time: f32,
+    // Needed for 16-bit alignment on WebGL2
```
Nahor, line 9: I believe it's "16-**byte** alignment"

examples/shader/shader_material_wesl.rs
```diff
@@ -95,8 +95,9 @@ fn update(
 #[derive(Asset, TypePath, AsBindGroup, Clone)]
 #[bind_group_data(CustomMaterialKey)]
 struct CustomMaterial {
+    // Needed for 16 bit alignment in WebGL2
```
Nahor, line 98: Same as above: 16-**byte** alignment

PR #18593, opened 2025-03-28
Improve error message for missing resources

### LINKED ISSUE #18515: Return the failed system parameter(s) in SystemParamValidationError

chescock:
# Objective

Fixes #18515 

After the recent changes to system param validation, the panic message for a missing resource is currently: 

```
Encountered an error in system `missing_resource_error::res_system`: SystemParamValidationError { skipped: false }
```

Add the parameter type name and a descriptive message, improving the panic message to:

```
Encountered an error in system `missing_resource_error::res_system`: SystemParamValidationError { skipped: false, message: "Resource does not exist", param: "bevy_ecs::change_detection::Res<missing_resource_error::MissingResource>" }
```

## Solution

Add fields to `SystemParamValidationError` for error context.  Include the `type_name` of the param and a message.  

Store them as `Cow<'static, str>` and only format them into a friendly string in the `Display` impl.  This lets us create errors using a `&'static str` with no allocation or formatting, while still supporting runtime `String` values if necessary.  

Add a unit test that verifies the panic message.

## Future Work

If we change the default error handling to use `Display` instead of `Debug`, and to use `ShortName` for the system name, the panic message could be further improved to:

```
Encountered an error in system `res_system`: Parameter `Res<MissingResource>` failed validation: Resource does not exist
```

However, `BevyError` currently includes the backtrace in `Debug` but not `Display`, and I didn't want to try to change that in this PR.

### DIFF
--- a/crates/bevy_ecs/src/system/system_param.rs
+++ b/crates/bevy_ecs/src/system/system_param.rs
@@ -17,18 +17,21 @@ use crate::{
         FromWorld, World,
     },
 };
-use alloc::{borrow::ToOwned, boxed::Box, vec::Vec};
+use alloc::{
+    borrow::{Cow, ToOwned},
+    boxed::Box,
+    vec::Vec,
+};
 pub use bevy_ecs_macros::SystemParam;
 use bevy_ptr::UnsafeCellDeref;
 use bevy_utils::synccell::SyncCell;
 use core::{
     any::Any,
-    fmt::Debug,
+    fmt::{Debug, Display},
     marker::PhantomData,
     ops::{Deref, DerefMut},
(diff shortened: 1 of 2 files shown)


JaySpruce (approved):

alice-i-cecile (approved):
Thanks for polishing this up: this is important follow-up that I deliberately punted on :) I think I agree that it would be irresponsible to ship without this.

PR #18601, opened 2025-03-29
Added `StableInterpolate` implementations for linear colors.

mintlu8:
# Objective

Colors currently do not implement `StableInterpolate`, which makes them ineligible for functions like `smooth_nudge` and make some generic APIs awkward.

## Solution

Implemented `StableInterpolate` for linear color types that should be uncontroversial. Non-linear types like `Hsl` are not implemented in this PR.

## Testing

Added a test that checks implementations are correct.

### DIFF
--- a/crates/bevy_color/src/interpolate.rs
+++ b/crates/bevy_color/src/interpolate.rs
@@ -1,24 +1,20 @@
-use bevy_math::StableInterpolate;
+use bevy_math::{StableInterpolate, VectorSpace};
 
 use crate::{Laba, LinearRgba, Oklaba, Srgba, Xyza};
 
 macro_rules! impl_stable_interpolate_linear {
-    ($name: ident $(, $field: ident)*) => {
-        impl StableInterpolate for $name {
+    ($($name: ident),*) => {
+        $(impl StableInterpolate for $name {
             fn interpolate_stable(&self, other: &Self, t: f32) -> Self {
-                $name {
-                    $($field: self.$field.interpolate_stable(&other.$field, t),)*
-                }
+                $name::lerp(*self, *other, t)
             }
-        }
+        })*
     };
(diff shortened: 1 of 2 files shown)


greeble-dev (commented):
Could consider reusing `impl_componentwise_vector_space`? https://github.com/bevyengine/bevy/commit/b8d7c352f1f966662c892c5722c1eb1900cda59c

greeble-dev (approved):

crates/bevy_color/src/interpolate.rs
```diff
@@ -0,0 +1,37 @@
...
+#[cfg(test)]
+mod test {
+    use bevy_math::StableInterpolate;
+
+    use crate::{Gray, Laba, LinearRgba, Oklaba, Srgba, Xyza};
+
+    #[test]
+    pub fn test_color_stable_interpolate() {
+        let b = Srgba::BLACK;
+        let w = Srgba::WHITE;
```
IQuick143, line 12: :thinking: Is Srgba actually a linear space?

IQuick143 (approved):
Seems good, although I'm not personally certain about the impls other than `LinearRgba`.

crates/bevy_color/src/interpolate.rs
mintlu8: I could be wrong but I think the point of color spaces is to use them in a linear space.

NthTensor:
This isn't the place to get into it I guess, but I'm increasingly of the feeling that the color types shouldn't implement `VectorSpace` and if you want to do color math, use `to_vec4` and `from_vec4`.

That said, all of the colors can still actually implement `StableInterpolate` I think. So I guess this is fine. I'm working on something that will make some more significant changes here.

PR #18605, opened 2025-03-29
Return triangle index instead of vertex index (Fixes #18081)

### LINKED ISSUE #18081: RayMeshHit.triangle_index contains the first vertex index instead of the triangle index
**Bevy Version: 0.15**

**Issue Summary**  
`RayMeshHit.triangle_index` currently stores the index of the first vertex of the hit triangle rather than the triangle’s actual index. This makes it impossible to determine the other two vertices when the mesh uses indexed geometry (especially when vertices are shared among multiple triangles or aren’t strictly grouped in triples). As a result, getting UVs or texture colors for the exact hit point is blocked.
This prevents use-cases like texture painting and texture based decals from working.

**Background**  
This problem was previously identified in the [`bevy_mod_raycast` crate](https://github.com/aevyrie/bevy_mod_raycast/pull/103) and an unmerged PR [#117](https://github.com/aevyrie/bevy_mod_raycast/pull/117) proposed a fix. The same issue appears to apply in the core Bevy code where the relevant code assigns `Some(a)` for the triangle index instead of using the iteration (triangle) index:

```rust
for triangle in indices.chunks_exact(3) {
    let [a, b, c] = [
        triangle[0].try_into().ok()?,
        triangle[1].try_into().ok()?,
        triangle[2].try_into().ok()?,
    ];
    let triangle_index = Some(a);
    // ...
}
```

Switching to:

```rust
for (i, triangle) in indices.chunks_exact(3).enumerate() {
    let [a, b, c] = [
        triangle[0].try_into().ok()?,
        triangle[1].try_into().ok()?,
        triangle[2].try_into().ok()?,
    ];
    let triangle_index = Some(i);
    // ...
}
```

would correctly store the triangle’s index instead of the first vertex index, resolving the issue.

**Request**  
Please let me know if there’s any oversight or if I should open a PR to address this.

brookman:
# Objective

- Fixes #18081
- Enable use-cases like getting UVs or texture colors for the hit point (which are currently not possible due to this bug).

## Solution

- Return the triangle index instead of the first vertex index of the triangle.

## Testing

Tested successfully with my project which does a raycast to get the UV coordinates of the hit. My code:
```rust
fn get_uv(
    mesh: &Mesh,
    attribute: &MeshVertexAttribute,
    hit: &RayMeshHit,
    _gizmos: &mut Gizmos,
) -> Result<Vec2> {
    let (a, b, c) = get_indices(mesh, hit)?;

    let attrs = mesh
        .attribute(*attribute)
        .ok_or_eyre(format!("Attribute {:?} not found", &attribute))?;
    let all_uvs: &Vec<[f32; 2]> = match &attrs {
        VertexAttributeValues::Float32x2(positions) => positions,
        _ => bail!("Unexpected types in {:?}", Mesh::ATTRIBUTE_UV_0),
    };

    let bary = hit.barycentric_coords;

    Ok(Vec2::from_array(all_uvs[a]) * bary.x
        + Vec2::from_array(all_uvs[b]) * bary.y
        + Vec2::from_array(all_uvs[c]) * bary.z)
}

fn get_indices(mesh: &Mesh, hit: &RayMeshHit) -> Result<(usize, usize, usize)> {
    let i = hit
        .triangle_index
        .ok_or_eyre("Intersection Index not found")?;

    Ok(mesh.indices().map_or_else(
        || (i, i + 1, i + 2),
        |indices| match indices {
            Indices::U16(indices) => (
                indices[i * 3] as usize,
                indices[i * 3 + 1] as usize,
                indices[i * 3 + 2] as usize,
            ),
            Indices::U32(indices) => (
                indices[i * 3] as usize,
                indices[i * 3 + 1] as usize,
                indices[i * 3 + 2] as usize,
            ),
        },
    ))
}
```

Edit: I also noticed that `triangle` and `triangle_index` can never be "None" and hence removed the Options.
Please let me know if this change is reasonable in the same PR.


crates/bevy_picking/src/mesh_picking/ray_cast/intersections.rs
```diff
@@ -17,9 +17,9 @@ pub struct RayMeshHit {
     /// The distance from the ray origin to the intersection point.
     pub distance: f32,
     /// The vertices of the triangle that was hit.
```
aloucks, line 19: I'm curious if these were defined as `Option` so that the API wouldn't have to change if/when support for `PrimitiveTopology` other than `TriangleList` was added.

brookman: Good point. But wouldn't `barycentric_coords: Vec3` also have to be an `Option` in case on non-triangles?
Anyway, let's keep this change separate. I will revert this change and create separate PR

brookman: @aloucks: Do you mind if I close this conversation?

aloucks: go for it 👍

brookman:
Closed this PR because it was targeting the wrong branch. [new PR](https://github.com/bevyengine/bevy/pull/18647)

PR #18610, opened 2025-03-29
Fix diffuse transmission for anisotropic materials

JMS55:
Expand the diff, this was obviously just a copy paste bug at some point.


Feilkin (approved):
Looks good to me. I didn't test the change as I'm not quite sure what anisotropic transmission is supposed to look like.

IceSentry (approved):

PR #18613, opened 2025-03-29
Expose symbols needed to replicate SetMeshBindGroup in ecosystem crates.

komadori:
# Objective

My ecosystem crate, bevy_mod_outline, currently uses `SetMeshBindGroup` as part of its custom rendering pipeline. I would like to allow for possibility that, due to changes in 0.16, I need to customise the behaviour of `SetMeshBindGroup` in order to make it work. However, not all of the symbol needed to implement this render command are public outside of Bevy.

## Solution

- Include `MorphIndices` in re-export list. I feel this is morally equivalent to `SkinUniforms` already being exported.
- Change `MorphIndex::index` field to be public. I feel this is morally equivalent to the `SkinByteOffset::byte_offset` field already being public.
- Change `RenderMeshIntances::mesh_asset_id()` to be public (although since all the fields of `RenderMeshInstances` are public it's possible to work around this one by reimplementing).

These changes exclude:
- Making any change to the `RenderLightmaps` type as I don't need to bind the light-maps for my use-case and I wanted to keep these changes minimal. It has a private field which would need to be public or have access methods.
- The changes already included in #18612.

## Testing

Confirmed that a copy of `SetMeshBindGroup` can be compiled outside of Bevy with these changes, provided that the light-map code is removed.

### DIFF
--- a/crates/bevy_pbr/src/render/mesh.rs
+++ b/crates/bevy_pbr/src/render/mesh.rs
@@ -938,7 +938,7 @@ impl RenderMeshInstances {
     }
 
     /// Returns the ID of the mesh asset attached to the given entity, if any.
-    pub(crate) fn mesh_asset_id(&self, entity: MainEntity) -> Option<AssetId<Mesh>> {
+    pub fn mesh_asset_id(&self, entity: MainEntity) -> Option<AssetId<Mesh>> {
         match *self {
             RenderMeshInstances::CpuBuilding(ref instances) => instances.mesh_asset_id(entity),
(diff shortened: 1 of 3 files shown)


alice-i-cecile (approved):

tychedelia (approved):

JMS55 (approved):

IceSentry (approved):

alice-i-cecile:
Resolved merge conflicts and fixed a re-export bug. Merging :)

PR #18614, opened 2025-03-29
Expose some fields of Stepping 

### LINKED ISSUE #18100: Expose some fields of `Stepping`

I'd like to build a simple system stepping tool, but the fields of `Stepping` are private so I can't update the tool's UI without manually tracking all changes to `Stepping` and replicating its internal data structures (and even then, I have to assume that `Stepping` is never mutated anywhere else).

`Stepping` only exposes the current cursor position via `cursor()`, the schedule order via `schedules()`, and whether it's enabled via `is_enabled()`. I can't check if a system is set to always run, or never run, or has a breakpoint, for example.

The simplest solution is to expose the fields of `Stepping` and `ScheduleState` as `pub`. Since `Stepping` is a debugging tool, I think it makes sense to expose the internals.

Add more getter methods like `cursor()` and `schedules()` until all the relevant info is accessible.

gustav-horn:
# Objective

- Fixes #18100

## Solution

- Create get methods for Stepping. The reason I decided to add get methods instead of making the fields public is to prevent the possibility of the user accidentally mutating the struct.

## Testing

### DIFF
--- a/crates/bevy_ecs/src/schedule/stepping.rs
+++ b/crates/bevy_ecs/src/schedule/stepping.rs
@@ -220,6 +220,7 @@ impl Stepping {
         self.action != Action::RunAll
     }
 
+    // Returns a reference to the Action to be executed during the current frame.
     pub fn get_action(&self) -> &Action {
         &self.action
     }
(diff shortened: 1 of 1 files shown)


alice-i-cecile:
I think that's fine to leave out for now.

PR #18618, opened 2025-03-30
Avoid performance pitfalls in update_previous_global_transform system

alice-i-cecile:
# Objective

@aevyrie saw this system showing up in his `big_space` traces.

image

Upon investigation, this system is using commands to unconditionally mutate a component. That's slow!

Apparently this fixes https://github.com/bevyengine/bevy/issues/14681.

## Solution

1. Don't use commands.
2. Move the should_run guard statement into an actual run condition.
3. Throw together some simple docs.
4. Make it parallel.
5. Use required components to reduce archetype moves (and bugs).

~~I'm going to avoid parallelizing for now; let's take the free wins here first.~~

>  but good news, it is down to 3ms from 115ms on main

Casual 38x perf improvement.

## Requests for reviewers

1. I am not sure where this is used: what should I be testing?
~~2. @aevyrie I'd be interested in a fresh bench on this, but this is so clearly better that I don't care if we skip it.~~
~~3. Should I swap to using required components here and avoid ever inserting?~~

### DIFF
--- a/crates/bevy_pbr/src/prepass/mod.rs
+++ b/crates/bevy_pbr/src/prepass/mod.rs
@@ -270,7 +270,7 @@ pub fn any_camera_active(views: Query<&Camera, Or<(With<Camera3d>, With<ShadowVi
     views.iter().any(|camera| camera.is_active)
 }
 
-/// Updates the [`PreviousGlobalTransform`] of all meshes that have a [`PreviousMeshFilter`] component,
+/// Updates the [`PreviousGlobalTransform`] of all meshes that match the [`PreviousMeshFilter`],
 /// inserting it if needed.
 pub fn update_mesh_previous_global_transforms(
     mut commands: Commands,
--- a/crates/bevy_pbr/src/prepass/mod.rs
+++ b/crates/bevy_pbr/src/prepass/mod.rs
@@ -187,7 +187,13 @@ where
                     BinnedRenderPhasePlugin::<AlphaMask3dPrepass, MeshPipeline>::new(
                         self.debug_flags,
                     ),
-                ));
+                ))
+                // This can't be an attribute on the mesh types due to the shape of our dependency graph
+                .register_required_components::<Mesh3d, PreviousGlobalTransform>();
+
+            //
+            #[cfg(feature = "meshlet")]
+            app.register_required_components::<MeshletMesh3d, PreviousGlobalTransform>();
         }
 
         let Some(render_app) = app.get_sub_app_mut(RenderApp) else {
@@ -257,6 +263,11 @@ pub fn update_previous_view_data(
     }
 }
 
+/// Tracks the previous [`GlobalTransform`] of a mesh,
+/// assuming that the mesh's transform is affine.
+///
+/// This is added as a required component for [`Mesh3d`] and the meshlet equivalent in
+/// the [`PrepassPlugin`].
 #[derive(Component, PartialEq, Default)]
 pub struct PreviousGlobalTransform(pub Affine3A);
 
@@ -270,24 +281,13 @@ pub fn any_camera_active(views: Query<&Camera, Or<(With<Camera3d>, With<ShadowVi
     views.iter().any(|camera| camera.is_active)
 }
 
-/// Updates the [`PreviousGlobalTransform`] of all meshes that match the [`PreviousMeshFilter`],
-/// inserting it if needed.
+/// Updates the [`PreviousGlobalTransform`] of all meshes that match the [`PreviousMeshFilter`].
 pub fn update_mesh_previous_global_transforms(
-    mut commands: Commands,
     mut meshes: Query<(&GlobalTransform, &mut PreviousGlobalTransform), PreviousMeshFilter>,
-    new_meshes: Query<
-        (Entity, &GlobalTransform),
-        (PreviousMeshFilter, Without<PreviousGlobalTransform>),
-    >,
 ) {
-    for (entity, transform) in &new_meshes {
-        commands
-            .entity(entity)
-            .try_insert(PreviousGlobalTransform(transform.affine()));
-    }
-    for (transform, mut previous) in &mut meshes {
+    meshes.par_iter_mut().for_each(|(transform, mut previous)| {
         previous.set_if_neq(PreviousGlobalTransform(transform.affine()));
-    }
+    });
 }
 
 #[derive(Resource)]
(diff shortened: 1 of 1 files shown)


crates/bevy_pbr/src/prepass/mod.rs
```diff
@@ -265,21 +265,35 @@ type PreviousMeshFilter = With<Mesh3d>;
...
 type PreviousMeshFilter = Or<(With<Mesh3d>, With<MeshletMesh3d>)>;
 
+/// A run condition for [`update_mesh_previous_global_transforms`] that checks if any camera is active.
+pub fn any_camera_active(views: Query<&Camera, Or<(With<Camera3d>, With<ShadowView>)>>) -> bool {
+    views.iter().any(|camera| camera.is_active)
+}
+
+/// Updates the [`PreviousGlobalTransform`] of all meshes that have a [`PreviousMeshFilter`] component,
+/// inserting it if needed.
 pub fn update_mesh_previous_global_transforms(
```
aevyrie, line 275: you can avoid branching with
```rs
pub fn update_mesh_previous_global_transforms(
    mut commands: Commands,
    views: Query<&Camera, Or<(With<Camera3d>, With<ShadowView>)>>,
    new_meshes: Query<
        (Entity, &GlobalTransform),
        (PreviousMeshFilter, Without<PreviousGlobalTransform>),
    >,
    mut meshes: Query<(&GlobalTransform, &mut PreviousGlobalTransform), PreviousMeshFilter>,
) {
    let should_run = views.iter().any(|camera| camera.is_active);

    if should_run {
        for (entity, transform) in &new_meshes {
            let new_previous_transform = PreviousGlobalTransform(transform.affine());
            commands.entity(entity).try_insert(new_previous_transform);
        }
        for (transform, mut previous) in &mut meshes {
            previous.set_if_neq(PreviousGlobalTransform(transform.affine()));
        }
    }
}
```

aevyrie (approved):

crates/bevy_pbr/src/prepass/mod.rs
aevyrie: Oh, and parallel
```rs
meshes.par_iter_mut().for_each(|(transform, mut previous)| {
    previous.set_if_neq(PreviousGlobalTransform(transform.affine()));
});
```

crates/bevy_pbr/src/prepass/mod.rs
```diff
@@ -187,7 +187,13 @@ where
                     BinnedRenderPhasePlugin::<AlphaMask3dPrepass, MeshPipeline>::new(
                         self.debug_flags,
                     ),
-                ));
+                ))
+                // This can't be an attribute on the mesh types due to the shape of our dependency graph
+                .register_required_components::<Mesh3d, PreviousGlobalTransform>();
+
+            //
```
mockersf, line 194:
```suggestion
```

mockersf, line 194: or
```suggestion
            // This can't be an attribute on the mesh types due to the shape of our dependency graph
```

crates/bevy_pbr/src/prepass/mod.rs
```diff
@@ -187,7 +187,13 @@ where
                     BinnedRenderPhasePlugin::<AlphaMask3dPrepass, MeshPipeline>::new(
                         self.debug_flags,
                     ),
-                ));
+                ))
+                // This can't be an attribute on the mesh types due to the shape of our dependency graph
+                .register_required_components::<Mesh3d, PreviousGlobalTransform>();
```
JMS55, line 192: I'm not sure that this is something we want.

PreviousGlobalTransform is intentionally _not_ required by Mesh3d, and it's not due to dependencies (I've talked about wanting to move PreviousGlobalTransform into bevy_transform before).

If you're making a game without temporal effects (TAA/DLSS/RT/motion blur/etc), then there's no need for PreviousGlobalTransform. It's only required for Mesh3d if at least one camera has MotionVectorPrepass. 

I think we're better off:
1. Moving PreviousGlobalTransform to bevy_transform
2. Having a system in bevy_transform to update PreviousGlobalTransform, no filters based on mesh or whatever
3. Have a new system in bevy_pbr/bevy_core_pipeline that when there's at least one camera with a MotionVectorPrepass, query for Mesh3d entities without a PreviousGlobalTransform component and insert a dummy one. Order this before the system from 2.

crates/bevy_pbr/src/prepass/mod.rs
```diff
@@ -187,7 +187,13 @@ where
...
+
+            //
+            #[cfg(feature = "meshlet")]
+            app.register_required_components::<MeshletMesh3d, PreviousGlobalTransform>();
```
JMS55, line 196: This is not needed, MeshletMesh3d already requires PreviousGlobalTransform.

JMS55 (changes requested):

crates/bevy_pbr/src/prepass/mod.rs
JMS55: We could also do something like add a cargo feature for this, and say using MotionVectorPrepass requires turning the flag on, and then use cfg_attr to require PreviousGlobalTransform.

crates/bevy_pbr/src/prepass/mod.rs
```diff
@@ -265,24 +275,18 @@ type PreviousMeshFilter = With<Mesh3d>;
 #[cfg(feature = "meshlet")]
 type PreviousMeshFilter = Or<(With<Mesh3d>, With<MeshletMesh3d>)>;
 
+/// A run condition for [`update_mesh_previous_global_transforms`] that checks if any camera is active.
+pub fn any_camera_active(views: Query<&Camera, Or<(With<Camera3d>, With<ShadowView>)>>) -> bool {
```
IceSentry, line 279: Do we have somewhere to have common run conditions? I feel like this could be useful for more than just the prepass module

aevyrie:
This is an alternative/sibling PR that only addresses perf within functions, so might be less controversial. https://github.com/bevyengine/bevy/pull/18632

alice-i-cecile:
Moving this out of the milestone in favor of #18632. I'll clean this up once that's merged: there's some smaller quality issues to address still.

crates/bevy_pbr/src/prepass/mod.rs
IceSentry: This doesn't feel like something that needs a cargo feature, but I agree with everything else.

PR #18621, opened 2025-03-30
Code quality cleanup pass for #[require]

Bleachfuel:
#18555 improved syntax for required components.

However some code was a bit redundant after the new parsing and struct initializing would not give proper errors. 
This PR fixes that.

### DIFF
--- a/crates/bevy_ecs/macros/src/component.rs
+++ b/crates/bevy_ecs/macros/src/component.rs
@@ -599,6 +599,12 @@ impl Parse for Require {
             input.parse::<Token![=]>()?;
             let expr: Expr = input.parse()?;
             Some(quote!(|| #expr ))
+        } else if input.peek(Brace) {
+            // This is a "value style" named-struct-like require
+            let content;
+            braced!(content in input);
+            let content = content.parse::<TokenStream2>()?;
+            Some(quote!(|| #path { #content }))
         } else if input.peek(Paren) {
             // This is a "value style" tuple-struct-like require
             let content;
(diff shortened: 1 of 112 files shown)


crates/bevy_ecs/macros/src/component.rs
```diff
@@ -595,40 +596,32 @@ impl Parse for Require {
 
         let func = if input.peek(Token![=]) {
             // If there is an '=', then this is a "function style" require
-            let _t: syn::Token![=] = input.parse()?;
+            input.parse::<Token![=]>()?;
             let expr: Expr = input.parse()?;
-            let tokens: TokenStream = quote::quote! (|| #expr).into();
-            Some(TokenStream2::from(tokens))
+            Some(quote!(|| { #expr }))
```
Bleachfuel, line 601:
```suggestion
            Some(quote!(||  #expr ))
```

HugoPeters1024:
Could you give a specific example of where the errors where not clear before? It's not that obvious from the code what has changed and why :)

alice-i-cecile (approved):

Bleachfuel:
I've removed the extra struct parsing. Now it should only be code quality things.

crates/bevy_ecs/macros/src/component.rs
```diff
@@ -595,40 +596,32 @@ impl Parse for Require {
...
             let content = content.parse::<TokenStream2>()?;
-            let tokens: TokenStream = quote::quote! (|| #path { #content }).into();
-            Some(TokenStream2::from(tokens))
+            Some(quote!(|| #path { #content }))
         } else if input.peek(Paren) {
             // This is a "value style" tuple-struct-like require
             let content;
             parenthesized!(content in input);
-            let content = content.parse::<TokenStream2>()?;
+            let content = content.parse_terminated(Expr::parse, Token![,])?;
```
cart, line 612: I'd prefer to use the "passthrough" approach here too for consistency and performance.

cart (approved):

cart:
(I put this in 0.16 because it is non-breaking, unrisky, and could improve compile times)

PR #18622, opened 2025-03-30
Improve `Query`'s top-level documentation

BD103:
# Objective

- There's been several changes to `Query` for this release cycle, and `Query`'s top-level documentation has gotten slightly out-of-date.
- Alternative to #18615.

## Solution

- Edit `Query`'s docs for consistency, clarity, and correctness.
- Make sure to group `get()` and `get_many()` together instead of `single()` and `get_many()`, to enforce the distinction from https://github.com/bevyengine/bevy/pull/18615#issuecomment-2764355672.
- Reformat doc tests so they would be readable if extracted into their own file. (Which mainly involves adding more spacing.)
- Move link definitions to be nearer where they are used.
- Fix the tables so they are up-to-date and correctly escape square brackets `\[ \]`.

## Testing

I ran `cargo doc -p bevy_ecs --no-deps` to view the docs and `cargo test -p bevy_ecs --doc` to test the doc comments.

## Reviewing

The diff is difficult to read, so I don't recommend _just_ looking at that. Instead, run `cargo doc -p bevy_ecs --no-deps` locally and read through the new version. It should theoretically read smoother with less super-technical jargon. :)

## Follow-up

I want to go through some of `Query`'s methods, such as `single()`, `get()`, and `get_many()`, but I'll leave that for another PR.

### DIFF
--- a/crates/bevy_ecs/src/system/query.rs
+++ b/crates/bevy_ecs/src/system/query.rs
@@ -15,33 +15,40 @@ use core::{
     ops::{Deref, DerefMut},
 };
 
-/// [System parameter] that provides selective access to the [`Component`] data stored in a [`World`].
+/// A [system parameter] that provides selective access to the [`Component`] data stored in a [`World`].
 ///
-/// Enables access to [entity identifiers] and [components] from a system, without the need to directly access the world.
-/// Its iterators and getter methods return *query items*.
-/// Each query item is a type containing data relative to an entity.
+/// Queries enable systems to access [entity identifiers] and [components] without requiring direct access to the [`World`].
+/// Its iterators and getter methods return *query items*, which are types containing data relative to an entity.
 ///
 /// `Query` is a generic data structure that accepts two type parameters:
 ///
-/// - **`D` (query data).**
-///   The type of data contained in the query item.
+/// - **`D` (query data)**:
+///   The type of data fetched by the query, which will be returned in the query item.
 ///   Only entities that match the requested data will generate an item.
 ///   Must implement the [`QueryData`] trait.
-/// - **`F` (query filter).**
-///   A set of conditions that determines whether query items should be kept or discarded.
+/// - **`F` (query filter)**:
+///   An optional set of conditions that determine whether query items should be kept or discarded.
+///   This defaults to [`unit`], which means no additional filters will be applied.
 ///   Must implement the [`QueryFilter`] trait.
-///   This type parameter is optional.
 ///
+/// [system parameter]: crate::system::SystemParam
+/// [`Component`]: crate::component::Component
 /// [`World`]: crate::world::World
+/// [entity identifiers]: Entity
+/// [components]: crate::component::Component
 ///
 /// # Similar parameters
 ///
-/// [`Query`] has few sibling [`SystemParam`](crate::system::system_param::SystemParam)s, which perform additional validation:
-/// - [`Single`] - Exactly one matching query item.
+/// `Query` has few sibling [`SystemParam`]s, which perform additional validation:
+///
+/// - [`Single`](crate::system::Single) - Exactly one matching query item.
 /// - [`Option<Single>`] - Zero or one matching query item.
 /// - [`Populated`] - At least one matching query item.
 ///
-/// Those parameters will prevent systems from running if their requirements aren't met.
+/// These parameters will prevent systems from running if their requirements are not met.
+///
+/// [`SystemParam`]: crate::system::system_param::SystemParam
+/// [`Option<Single>`]: Single
 ///
 /// # System parameter declaration
 ///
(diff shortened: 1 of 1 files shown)


crates/bevy_ecs/src/system/query.rs
```diff
@@ -50,330 +57,424 @@ use core::{
...
-/// |[`iter_many`]\[[`_mut`][`iter_many_mut`]]|O(k)|
-/// |[`iter_combinations`]\[[`_mut`][`iter_combinations_mut`]]|O(<sub>n</sub>C<sub>r</sub>)|
-/// |[`get`]\[[`_mut`][`get_mut`]]|O(1)|
-/// |([`get_`][`get_many`])[`many`]|O(k)|
-/// |([`get_`][`get_many_mut`])[`many_mut`]|O(k<sup>2</sup>)|
-/// |[`single`]\[[`_mut`][`single_mut`]],<br>[`single`]\[[`_mut`][`single_mut`]]|O(a)|
-/// |Archetype based filtering ([`With`], [`Without`], [`Or`])|O(a)|
+/// |-|-|
+/// |[`iter`]\[[`_mut`][`iter_mut`]\]|O(n)|
+/// |[`iter().for_each()`][`for_each`],<br />[`iter_mut().for_each()`][`for_each`],<br/>[`par_iter`]\[[`_mut`][`par_iter_mut`]\]|O(n)|
```
chescock, line 431: It might make sense to do the same `[_mut]` syntax for the `for_each` row.  

```suggestion
/// |[`iter\[_mut\]().for_each()`][`for_each`],<br/>[`par_iter`]\[[`_mut`][`par_iter_mut`]\]|O(n)|
```

chescock (approved):
Looks good!

crates/bevy_ecs/src/system/query.rs
```diff
@@ -15,33 +15,40 @@ use core::{
...
 };
 
-/// [System parameter] that provides selective access to the [`Component`] data stored in a [`World`].
+/// A [system parameter] that provides selective access to the [`Component`] data stored in a [`World`].
 ///
-/// Enables access to [entity identifiers] and [components] from a system, without the need to directly access the world.
-/// Its iterators and getter methods return *query items*.
-/// Each query item is a type containing data relative to an entity.
+/// Queries enable systems to access [entity identifiers] and [components] without requiring direct access to the [`World`].
+/// Its iterators and getter methods return *query items*, which are types containing data relative to an entity.
```
Carter0, line 21: nit: relative to sounds weird cause its makes me think its brothers with the entity or something. 

Maybe just say "related to" or just "on" an entity

crates/bevy_ecs/src/system/query.rs
```diff
@@ -50,330 +57,424 @@ use core::{
...
+/// ) {
+///     // ...
+/// }
+/// #
+/// # bevy_ecs::system::assert_is_system(complex_query);
+/// ```
 ///
-/// In the following example, two components are accessed simultaneously, and the query items are filtered on two conditions.
+/// Note that this currently only works on tuples with 15 or fewer items. You may nest tuples,
+/// however, to get around this limit:
```
Carter0, line 152: Nit: That however sounds odd to me. Maybe just get rid of it. 

"You can nest tuples to get around this limit if needed" or something

crates/bevy_ecs/src/system/query.rs
```diff
@@ -50,330 +57,424 @@ use core::{
...
+/// [`ParamSet`]: crate::system::ParamSet
 ///
 /// ## Whole Entity Access
 ///
-/// [`EntityRef`]s can be fetched from a query. This will give read-only access to any component on the entity,
-/// and can be used to dynamically fetch any component without baking it into the query type. Due to this global
-/// access to the entity, this will block any other system from parallelizing with it. As such these queries
-/// should be sparingly used.
+/// [`EntityRef`] can be used in a query to gain read-only access to all components of an entity.
+/// This is useful when dynamically fetching components instead of baking them into the query type.
```
Carter0, line 294: Is this just not true anymore? 

`Due to this global access to the entity, this will block any other system from parallelizing with it. As such these queries should be sparingly used.`

crates/bevy_ecs/src/system/query.rs
```diff
@@ -50,330 +57,424 @@ use core::{
...
+/// [`get_mut`]: Self::get_mut
+/// [`get_many`]: Self::get_many
+/// [`get_many_mut`]: Self::get_many_mut
+///
 /// # Performance
 ///
-/// Creating a `Query` is a low-cost constant operation.
-/// Iterating it, on the other hand, fetches data from the world and generates items, which can have a significant computational cost.
+/// Creating a `Query` is a low-cost constant operation. Iterating it, on the other hand, fetches
+/// data from the world and generates items, which can have a significant computational cost.
```
Carter0, line 400: Can a 'query' query multiple worlds? Does it do that by default?

Carter0 (approved):
Minor thoughts and some questions, but it looks good!

crates/bevy_ecs/src/system/query.rs
```diff
@@ -50,330 +57,424 @@ use core::{
...
+/// |-|-|
+/// |[`iter`]\[[`_mut`][`iter_mut`]\]|Returns an iterator over all query items.|
+/// |[`iter().for_each()`][`for_each`],<br />[`iter_mut().for_each()`][`for_each`],<br />[`par_iter`]\[[`_mut`][`par_iter_mut`]\]|Runs a specified function for each query item.|
+/// |[`iter_many`]\[[`_mut`][`iter_many_mut`]\]|Iterates over query items that match a list of entities.|
+/// |[`iter_combinations`]\[[`_mut`][`iter_combinations_mut`]\]|Iterates over all combinations of query items.|
+/// |[`single`](Self::single)\[[`_mut`][`single_mut`]\]|Returns a single query item if only one exists.|
+/// |[`get`]\[[`_mut`][`get_mut`]\]|Returns the query item for a specified entity.|
+/// |[`get_many`]\[[`_mut`][`get_many_mut`]\]|Returns all query items that match a list of entities.|
 ///
 /// There are two methods for each type of query operation: immutable and mutable (ending with `_mut`).
```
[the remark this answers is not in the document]
BD103: Unless someone objects, I think I'll avoid mentioning it here, since it seems like a _very_ specific thing that not many will need to use.

crates/bevy_ecs/src/system/query.rs
```diff
@@ -50,330 +57,424 @@ use core::{
...
+/// |-|-|
+/// |[`iter`]\[[`_mut`][`iter_mut`]\]|O(n)|
+/// |[`iter().for_each()`][`for_each`],<br />[`iter_mut().for_each()`][`for_each`],<br/>[`par_iter`]\[[`_mut`][`par_iter_mut`]\]|O(n)|
+/// |[`iter_many`]\[[`_mut`][`iter_many_mut`]\]|O(k)|
+/// |[`iter_combinations`]\[[`_mut`][`iter_combinations_mut`]\]|O(<sub>n</sub>C<sub>r</sub>)|
+/// |[`single`](Self::single)\[[`_mut`][`single_mut`]\]|O(a)|
+/// |[`get`]\[[`_mut`][`get_mut`]\]|O(1)|
+/// |[`get_many`]|O(k)|
+/// |[`get_many_mut`]|O(k<sup>2</sup>)|
+/// |Archetype-based filtering ([`With`], [`Without`], [`Or`])|O(a)|
```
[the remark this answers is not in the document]
BD103: How would you change it to make it clearer? Would it be $O(1)$ with a footnote about this?

crates/bevy_ecs/src/system/query.rs
BD103: I added this back to the Performance section!

crates/bevy_ecs/src/system/query.rs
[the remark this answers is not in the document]
BD103: I'm going to leave this to a follow-up issue / PR, since I'm not sure either. (My knowledge of the ECS is limited to the user-facing side, I don't know much about the internals!)

crates/bevy_ecs/src/system/query.rs
Carter0: cool, thanks for getting back to me!

PR #18625, opened 2025-03-30
Implement `System` for `Box<dyn System>`

### LINKED ISSUE #18624: Implement `System` for `Box<dyn System>`

Boxed systems cannot be used with `IntoSystem`. This means that they can't use `IntoSystem::pipe` or any of the other related functions. This is especially useful for downstream crates relying on user-provided type erased systems.

Implement `System` for `Box<dyn System>` and both `System` and `ReadOnlySystem` for `Box<dyn ReadOnlySystem>`

N/A

hankjordan:
# Objective

- Implement `System` for `Box<dyn System>`
- Implement `System` and `ReadOnlySystem` for `Box<dyn System>`
- Add `BoxedReadOnlySystem` type alias
- Closes #18624 

## Solution

- See above

## Testing

- Added module test
- Should the implementation of `type_id` return the `type_id` of the Box or the underlying `type_id` impl?

### DIFF
--- a/crates/bevy_ecs/src/system/system.rs
+++ b/crates/bevy_ecs/src/system/system.rs
@@ -216,6 +216,172 @@ pub unsafe trait ReadOnlySystem: System {
 /// A convenience type alias for a boxed [`System`] trait object.
 pub type BoxedSystem<In = (), Out = ()> = Box<dyn System<In = In, Out = Out>>;
 
+/// A convenience type alias for a boxed [`ReadOnlySystem`] trait object.
+pub type BoxedReadOnlySystem<In = (), Out = ()> = Box<dyn ReadOnlySystem<In = In, Out = Out>>;
+
+impl<In, Out> System for Box<dyn System<In = In, Out = Out>>
+where
+    In: SystemInput + 'static,
+    Out: 'static,
+{
+    type In = In;
+    type Out = Out;
+
+    fn name(&self) -> Cow<'static, str> {
+        (**self).name()
+    }
+
+    fn component_access(&self) -> &Access<ComponentId> {
+        (**self).component_access()
+    }
+
+    fn archetype_component_access(&self) -> &Access<ArchetypeComponentId> {
+        (**self).archetype_component_access()
+    }
+
+    fn is_send(&self) -> bool {
+        (**self).is_send()
+    }
+
+    fn is_exclusive(&self) -> bool {
+        (**self).is_exclusive()
+    }
+
+    fn has_deferred(&self) -> bool {
+        (**self).has_deferred()
+    }
+
+    unsafe fn run_unsafe(
+        &mut self,
+        input: SystemIn<'_, Self>,
+        world: UnsafeWorldCell,
+    ) -> Self::Out {
+        (**self).run_unsafe(input, world)
+    }
+
+    fn apply_deferred(&mut self, world: &mut World) {
+        (**self).apply_deferred(world);
+    }
+
+    fn queue_deferred(&mut self, world: DeferredWorld) {
+        (**self).queue_deferred(world);
+    }
+
+    unsafe fn validate_param_unsafe(
+        &mut self,
+        world: UnsafeWorldCell,
+    ) -> Result<(), SystemParamValidationError> {
+        (**self).validate_param_unsafe(world)
+    }
+
+    fn initialize(&mut self, world: &mut World) {
+        (**self).initialize(world);
+    }
+
+    fn update_archetype_component_access(&mut self, world: UnsafeWorldCell) {
+        (**self).update_archetype_component_access(world);
+    }
+
+    fn check_change_tick(&mut self, change_tick: Tick) {
+        (**self).check_change_tick(change_tick);
+    }
+
+    fn get_last_run(&self) -> Tick {
+        (**self).get_last_run()
+    }
+
+    fn set_last_run(&mut self, last_run: Tick) {
+        (**self).set_last_run(last_run);
+    }
+}
+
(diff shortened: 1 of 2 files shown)


alice-i-cecile:
We've resisted similar things in the past (`impl Reflect for Box<dyn Reflect>`) out of concern for confusing double-boxing. I don't have a strong opinion on whether we should follow that precedent here, but I wanted to raise it.

crates/bevy_ecs/src/system/system.rs
```diff
@@ -216,6 +216,220 @@ pub unsafe trait ReadOnlySystem: System {
...
+/// A convenience type alias for a boxed [`ReadOnlySystem`] trait object.
+pub type BoxedReadOnlySystem<In = (), Out = ()> = Box<dyn ReadOnlySystem<In = In, Out = Out>>;
+
+impl<In, Out> System for Box<dyn System<In = In, Out = Out>>
+where
+    In: SystemInput + 'static,
+    Out: 'static,
+{
+    type In = In;
+    type Out = Out;
```
chescock, line 228: Looking at impls for `Box` in `std`, it seems more common to make them generic in the contents of the box, like

```suggestion
impl<S: System + ?Sized> System for Box<S> {
    type In = S::In;
    type Out = S::Out;
````

Then, since `dyn System: System`, that impl covers `Box<dyn System>`.

**Review by chescock — APPROVED** [review]
> We've resisted similar things in the past (`impl Reflect for Box<dyn Reflect>`) out of concern for confusing double-boxing. I don't have a strong opinion on whether we should follow that precedent here, but I wanted to raise it.

I don't have a strong opinion here, either, but I can see how being able to use combinators like `pipe` with already-boxed systems could be useful.  

> * Should the implementation of `type_id` return the `type_id` of the Box or the underlying `type_id` impl?

I'd vote for using the underlying impl, as you're doing now.  After all, the whole point of that method is to get the actual type of a `dyn System`!

**Comment by schrottkatze** [comment]
Adding my one cent here: I need something like this currently, and while I might find another solution eventually, this would've been much more intuitive and easy to use. 

Edit: yup, only possible to do elegantly and performantly (I have to resort to exclusive systems multiple times currently) with this merged. Hope that happens some day :p 

edit edit: clarification for why I need this (i was advised to post this here, as a motivation on why this should get merged)

I'm working on a mass-loading plugin for bevy, that allows you to take systems as inputs for:
- generating `AssetPath`s on each run (a system with `AssetPath<'static>` as output)
- running an action (just another system) when an asset is done loading (so you can, for example, spawn the scene you just loaded or tell the rest of the program it's done loading and it can start doing whatever else)
- running a system when every asset is done, done here meaning failed or loaded, that doesn't matter

Currently, I frequently have to [resort to exclusive systems and weird pipelineing](https://codeberg.org/schrottkatze/84/src/commit/f5012c03573a98a3cd3c283f0a1e1443981c9762/src/level/staged.rs#L79-L94) (this code is likely outdated by the time this is read, but that's how it looks currently).

My preferred way would be to just be able to schedule the boxed systems like any other and for example just have an abstraction over the paths updating a resource, or internally doing some fancy stuff with events, so I don't have to resort to exclusive systems. (IMO that would also make it a lot more readable for people who aren't familiar with bevy and a lot more maintainable, since it allows me to treat systems as just any other type of data)

[This](https://codeberg.org/schrottkatze/84/src/commit/f5012c03573a98a3cd3c283f0a1e1443981c9762/src/level/stage1.rs#L15-L25) would be the intended outcome, vaguely.

It's not perfect yet and doesn't work yet but I wouldn't be surprised if the code being simpler due to not having so much code duplication etc would've avoided me writing the logic bug that currently breaks it.

Currently I also have to track variants of my path type (among others) for [plugin creation time](https://codeberg.org/schrottkatze/84/src/commit/f5012c03573a98a3cd3c283f0a1e1443981c9762/src/level/staged.rs#L297-L300) and [runtime](https://codeberg.org/schrottkatze/84/src/commit/f5012c03573a98a3cd3c283f0a1e1443981c9762/src/level/staged.rs#L305-L308) which would not be necessary if this was merged (the internal wiring would look very different in general, but I designed this in the assumption that something like this would work because it felt kinda obvious to me).

Hope this helps with design work and decision making :p

**Comment by ItsDoot** [comment]
Might there be a way to detect triple+ boxing when a system gets added to a schedule or registered as a one-shot? That would probably be helpful to users.

**Comment by chescock** [comment]
> Might there be a way to detect triple+ boxing when a system gets added to a schedule or registered as a one-shot? That would probably be helpful to users.

If you add a `System: Any` bound, then I think you could do `(&**boxed_system as &dyn Any).is::<BoxedSystem>()` to check for a boxed boxed system.  

Though if we generalize it to `impl<S: System + ?Sized> System for Box<S>` as I suggested above, then you'd also need to handle cases like `Box<dyn ReadOnlySystem>`.  

Another approach is to add a trait method like:

```rust
trait System {
    fn is_boxed(&self) -> bool { false }
}

impl System for Box<dyn System> {
    fn is_boxed(&self) -> bool { true }
}
```

and then call it on the inner `dyn System` of a `Box<dyn System>`.

ItsDoot:
At the risk of stack overflow, I think you could potentially do:

```rust
trait System {
    fn indirection_count(&self) -> usize {
        0
    }
}

impl<S: System + ?Sized> System for Box<S> {
    fn indirection_count(&self) -> usize {
        1 + (*self).indirection_count()
    }
} 
```

And then have some `warn_on_indirection_above: usize` configurable property that logs a warning if a system is registered with higher indirection than it.

Issue #18570, opened 2025-03-27
Flexible Reflection Paths


makspll:
## What problem does this solve or what need does it fill?
Let's say you are given a reflection path, and you don't know the type to expect on the other side, or the intention of this reflection path.

Let's assume the type:
```rust
struct Stucty {
      inner_tuple: Tuply,
      list: Vec<String>
}

struct Tuply(pub String);
```

I want these paths to be able to accees anything in the Structy type. With that in mind let's assume we're just yoinking the paths through `ParsedPath::parse`.

In this case whoever is providing the paths, to access the tuple string, I'd need to receive:
```rust
inner_tuple.0
```
to access the list I'd need:
```rust
list[0]
```

Now let's say these paths are coming from your bog standard scripting language, the user can only specify their intentions in the following ways:
- `[<string>]`
- `[<int>]`
- `.<string>`

How would these paths look like with these limitations? 
- `inner_tuple.0` would be: `["inner_tuple"][0]` right?
- `list[0]` would be: `["list"][0]` and so on

But hold on a minute, assuming I am doing these accesses one at a time (for the sake of brevity, I think it works out the same anyway):
- ["inner_tuple"] -> gets parsed to field access
- [0] -> gets parsed to tuple access

So then both paths cannot be valid! The tuple struct requires a tuple access whereas the list requires a list access!

So then as the mediator I have to "differentiate" between these types of accesses somehow. In the case of `bevy_mod_scripting` I currently do this by lettting users use `_1` to mean tuple access, and `[1]` to mean list access.

Ideally I don't want the user to have to know anything about the internals of reflection.

## What solution would you like?

- Reflection paths which do not assume what the target "kind" is

## Additional 
BMS examples:
```lua
assert(retrieved._1 == "Initial Value", "Resource should have default value but got: " .. retrieved._1)
assert(retrieved["1"] == "Initial Value", "Resource should have default value but got: " .. retrieved["1"])
```

makspll:
The problem would be made worse if we added `map` reflection paths, since then I'd have to differentiate another kind of thing!

alice-i-cecile:
> Reflection paths which do not assume what the target "kind" is at the time of parsing

Do you have any ideas about how this could be implemented? I don't immediately see how this would work, and it would be nice to have a design to hand to contributors :)

makspll:
> Do you have any ideas about how this could be implemented? I don't immediately see how this would work, and it would be nice to have a design to hand to contributors :)

[...]

PR #18626, opened 2025-03-30
feat: initial attempt at making reflection more flexible

makspll:
# Objective

Tries to address #18570 

## Solution

Changes the way reflection paths are traversed. Instead of looking at the parsed access path, and then deciding what we need to expect with a mostly 1-1 mapping, tries to be as flexible as possible, and give the user "the benefit of the doubt".

## Testing

Added basic tests, but not very comprehensive at the moment


makspll:
Closing as I don't think this is going to go anywhere

PR #18627, opened 2025-03-30
Remove entities from specialization caches when despawned.

### LINKED ISSUE #17872: Cold Specialization Memory Leaks
## Bevy version

- Main since https://github.com/bevyengine/bevy/pull/17567

### Pipeline Caches

Cold specialization added [`SpecializedMaterialPipelineCache<M>`](https://github.com/bevyengine/bevy/blob/3c9e696faa961ac01cec5933dcb4d16e653a5ecf/crates/bevy_pbr/src/material.rs#L739) and [`SpecializedMaterial2dPipelineCache<M>`](https://github.com/bevyengine/bevy/blob/3c9e696faa961ac01cec5933dcb4d16e653a5ecf/crates/bevy_sprite/src/mesh2d/material.rs#L594) resources, which are hashmaps mapping `(view_entity, entity)` to a tick and pipeline id. There's also [`SpecializedShadowMaterialPipelineCache<M>`](https://github.com/bevyengine/bevy/blob/3c9e696faa961ac01cec5933dcb4d16e653a5ecf/crates/bevy_pbr/src/render/light.rs#L1615) keyed by `(retained_view_entity, entity)`.

The specialization systems insert into these maps but they're never cleaned out. If one spawns and despawns lots of content, they continuously grow.

### Entity Specialization Ticks

There's a similar problem with `EntitySpecializationTicks` ([2d](https://github.com/bevyengine/bevy/blob/3c9e696faa961ac01cec5933dcb4d16e653a5ecf/crates/bevy_sprite/src/mesh2d/material.rs#L548-L559) and [3d](https://github.com/bevyengine/bevy/blob/3c9e696faa961ac01cec5933dcb4d16e653a5ecf/crates/bevy_pbr/src/material.rs#L693-L704)). It's inserted into, but never cleaned out.

cc @tychedelia

tychedelia:
# Objective

Fixes #17872 

## Solution

This should have basically no impact on static scenes. We can optimize more later if anything comes up. Needing to iterate the two level bin is a bit unfortunate but shouldn't matter for apps that use a single camera.

### DIFF
--- a/crates/bevy_pbr/src/material.rs
+++ b/crates/bevy_pbr/src/material.rs
@@ -708,7 +708,7 @@ fn extract_mesh_materials<M: Material>(
 pub fn extract_entities_needs_specialization<M>(
     entities_needing_specialization: Extract<Res<EntitiesNeedingSpecialization<M>>>,
     mut entity_specialization_ticks: ResMut<EntitySpecializationTicks<M>>,
-    mut removed_view_visibility_components: Extract<RemovedComponents<ViewVisibility>>,
+    mut removed_mesh_material_components: Extract<RemovedComponents<MeshMaterial3d<M>>>,
     mut specialized_material_pipeline_cache: ResMut<SpecializedMaterialPipelineCache<M>>,
     mut specialized_prepass_material_pipeline_cache: Option<
         ResMut<SpecializedPrepassMaterialPipelineCache<M>>,
@@ -726,7 +726,7 @@ pub fn extract_entities_needs_specialization<M>(
         entity_specialization_ticks.insert((*entity).into(), ticks.this_run());
     }
     // Clean up any despawned entities
-    for entity in removed_view_visibility_components.read() {
+    for entity in removed_mesh_material_components.read() {
         entity_specialization_ticks.remove(&MainEntity::from(entity));
         for view in views {
             if let Some(cache) =
--- a/crates/bevy_pbr/src/material.rs
+++ b/crates/bevy_pbr/src/material.rs
@@ -34,6 +34,7 @@ use bevy_platform_support::collections::{HashMap, HashSet};
 use bevy_platform_support::hash::FixedHasher;
 use bevy_reflect::std_traits::ReflectDefault;
 use bevy_reflect::Reflect;
+use bevy_render::camera::extract_cameras;
 use bevy_render::mesh::mark_3d_meshes_as_changed_if_their_assets_changed;
 use bevy_render::render_asset::prepare_assets;
 use bevy_render::renderer::RenderQueue;
@@ -315,7 +316,7 @@ where
                     ExtractSchedule,
                     (
                         extract_mesh_materials::<M>.before(ExtractMeshesSet),
-                        extract_entities_needs_specialization::<M>,
+                        extract_entities_needs_specialization::<M>.after(extract_cameras),
                     ),
                 )
                 .add_systems(
@@ -709,11 +710,11 @@ pub fn extract_entities_needs_specialization<M>(
     mut entity_specialization_ticks: ResMut<EntitySpecializationTicks<M>>,
     mut removed_view_visibility_components: Extract<RemovedComponents<ViewVisibility>>,
     mut specialized_material_pipeline_cache: ResMut<SpecializedMaterialPipelineCache<M>>,
-    mut specialized_prepass_material_pipeline_cache: ResMut<
-        SpecializedPrepassMaterialPipelineCache<M>,
+    mut specialized_prepass_material_pipeline_cache: Option<
+        ResMut<SpecializedPrepassMaterialPipelineCache<M>>,
     >,
-    mut specialized_shadow_material_pipeline_cache: ResMut<
-        SpecializedShadowMaterialPipelineCache<M>,
+    mut specialized_shadow_material_pipeline_cache: Option<
+        ResMut<SpecializedShadowMaterialPipelineCache<M>>,
     >,
     views: Query<&ExtractedView>,
     ticks: SystemChangeTick,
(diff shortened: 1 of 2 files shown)


crates/bevy_pbr/src/material.rs
```diff
@@ -715,6 +724,27 @@ pub fn extract_entities_needs_specialization<M>(
         // Update the entity's specialization tick with this run's tick
         entity_specialization_ticks.insert((*entity).into(), ticks.this_run());
     }
+    // Clean up any despawned entities
+    for entity in removed_view_visibility_components.read() {
```
IceSentry, line 728: Does removing `ViewVisibility` counts as despawning? I guess from the point of view of the render world if an entity is not visible then it can be removed.

IceSentry (approved):
I'm not entirely if it's correct to say that removing ViewVisibility implies a despawned entity, but that seems good enough.

crates/bevy_pbr/src/material.rs
tychedelia: Yeah this is probably inaccurate to say that's it's only a result of being despawned but the user removing the component is both something they shouldn't do in practice and will result in it not being rendered either way.

crates/bevy_pbr/src/material.rs
```diff
@@ -707,6 +708,15 @@ fn extract_mesh_materials<M: Material>(
 pub fn extract_entities_needs_specialization<M>(
     entities_needing_specialization: Extract<Res<EntitiesNeedingSpecialization<M>>>,
     mut entity_specialization_ticks: ResMut<EntitySpecializationTicks<M>>,
+    mut removed_view_visibility_components: Extract<RemovedComponents<ViewVisibility>>,
```
brianreavis, line 711: I'd consider changing this to `Extract<RemovedComponents<MeshMaterial3d<M>>>` to prevent despawns of other materials / other entities affecting the performance of this method.

crates/bevy_pbr/src/material.rs
```diff
@@ -715,6 +725,29 @@ pub fn extract_entities_needs_specialization<M>(
         // Update the entity's specialization tick with this run's tick
         entity_specialization_ticks.insert((*entity).into(), ticks.this_run());
     }
+    // Clean up any despawned entities
+    for entity in removed_view_visibility_components.read() {
+        entity_specialization_ticks.remove(&MainEntity::from(entity));
+        for view in views {
```
brianreavis, line 731: Iterating and looking up each view per entity may not be as efficient as it could be, but it's probably fine for now. The only alternatives I see currently:

1. Reverse the loop order: buffer the removed entities into a Vec and then loop through them for each view. Downside: buffering the entities.
2. Resolve the view caches ahead of time, and iterate those per entity like below. Downside: even more squirmy, given the unsafe usage to circumvent the borrow checker... but no buffering.

```rust
// 2

let mut view_caches: SmallVec<
    [(
        Option<*mut SpecializedMaterialViewPipelineCache<M>>,
        Option<*mut SpecializedPrepassMaterialViewPipelineCache<M>>,
        Option<*mut SpecializedShadowMaterialViewPipelineCache<M>>,
    ); 8],
> = SmallVec::new();

for view in views.iter() {
    view_caches.push((
        specialized_material_pipeline_cache
            .get_mut(&view.retained_view_entity)
            .map(|cache| cache as *mut _),
        specialized_prepass_material_pipeline_cache
            .as_mut()
            .and_then(|c| {
                c.get_mut(&view.retained_view_entity)
                    .map(|cache| cache as *mut _)
            }),
        specialized_shadow_material_pipeline_cache
            .as_mut()
            .and_then(|c| {
                c.get_mut(&view.retained_view_entity)
                    .map(|cache| cache as *mut _)
            }),
    ));
}

for entity in removed_view_visibility_components.read() {
    let main_entity = MainEntity::from(entity);
    entity_specialization_ticks.remove(&main_entity);

    for (mat_cache, prepass_cache, shadow_cache) in &view_caches {
        unsafe {
            if let Some(cache) = mat_cache {
                (*cache).remove(&main_entity);
            }
            if let Some(cache) = prepass_cache {
                (*cache).remove(&main_entity);
            }
            if let Some(cache) = shadow_cache {
                (*cache).remove(&main_entity);
            }
        }
    }
}
```

Curious what others' thoughts are.

crates/bevy_sprite/src/mesh2d/material.rs
```diff
@@ -555,6 +556,9 @@ pub const fn tonemapping_pipeline_key(tonemapping: Tonemapping) -> Mesh2dPipelin
 pub fn extract_entities_needs_specialization<M>(
     entities_needing_specialization: Extract<Res<EntitiesNeedingSpecialization<M>>>,
     mut entity_specialization_ticks: ResMut<EntitySpecializationTicks<M>>,
+    mut removed_view_visibility_components: Extract<RemovedComponents<ViewVisibility>>,
```
brianreavis, line 559: ```Extract<RemovedComponents<MeshMaterial2d<M>>>```

brianreavis (commented):
Thanks for addressing this!

crates/bevy_pbr/src/material.rs
tychedelia: Yeah, that's a good call.

alice-i-cecile (approved):
Agreed that this could probably be faster, but I'm mostly concerned about trying to improve the core architecture to make this sort of bug much harder to accidentally create.

That list of "now remove it from *this* cache" makes me nervous!

tychedelia:
> That list of "now remove it from this cache" makes me nervous!

Yeah we are *super* aware that this kind of ad hoc pattern is very difficult to maintain. The cost of bad abstractions in rendering code is just also unfortunately very high. But it's currently a really high priority of us to reduce the amount of brittle, duplicated code we have in the renderer. We're on it!

PR #18628, opened 2025-03-30
Fix no indirect drawing

tychedelia:
# Objective

The `NoIndirectDrawing` wasn't working and was causing the scene not to be rendered.

## Solution

Check the configured preprocessing mode when adding new batch sets and mark them as batchable instead of muli-drawable if indirect rendering has been disabled. 

## Testing

`cargo run --example many_cubes -- --no-indirect-drawing`

### DIFF
--- a/crates/bevy_render/src/render_phase/mod.rs
+++ b/crates/bevy_render/src/render_phase/mod.rs
@@ -375,8 +375,8 @@ pub enum BinnedRenderPhaseType {
     /// can be batched with other meshes of the same type.
     MultidrawableMesh,
 
-    /// The item is a mesh that's eligible for single-draw indirect rendering
-    /// and can be batched with other meshes of the same type.
+    /// The item is a mesh that can be batched with other meshes of the same type and
+    /// drawn in a single draw call.
     BatchableMesh,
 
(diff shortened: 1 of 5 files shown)


IceSentry (approved):

bushrat011899 (approved):

crates/bevy_render/src/render_phase/mod.rs
```diff
@@ -462,9 +465,17 @@ where
         bin_key: BPI::BinKey,
         (entity, main_entity): (Entity, MainEntity),
         input_uniform_index: InputUniformIndex,
-        phase_type: BinnedRenderPhaseType,
+        mut phase_type: BinnedRenderPhaseType,
         change_tick: Tick,
     ) {
+        // If the user has overridden indirect drawing for this view, we need to
+        // force the phase type to be batchable instead.
+        if let GpuPreprocessingMode::PreprocessingOnly = self.gpu_preprocessing_mode {
```
superdump, line 473: Why are these if let? Why not just compare the variant with == or use matches!()?

alice-i-cecile: Mild preference to just use == but I don't particularly mind this style.

PR #18629, opened 2025-03-30
Use Display instead of Debug in the default error handler

chescock:
# Objective

Improve error messages for missing resources.  

The default error handler currently prints the `Debug` representation of the error type instead of `Display`.  Most error types use `#[derive(Debug)]`, resulting in a dump of the structure, but will have a user-friendly message for `Display`.  

Follow-up to #18593

## Solution

Change the default error handler to use `Display` instead of `Debug`.  

Change `BevyError` to include the backtrace in the `Display` format in addition to `Debug` so that it is still included.  

## Showcase

Before: 

```
Encountered an error in system `system_name`: SystemParamValidationError { skipped: false, message: "Resource does not exist", param: "bevy_ecs::change_detection::Res<app_name::ResourceType>" }

Encountered an error in system `other_system_name`: "String message with\nmultiple lines."
```

After

```
Encountered an error in system `system_name`: Parameter `Res<ResourceType>` failed validation: Resource does not exist

Encountered an error in system `other_system_name`: String message with
multiple lines.
```

### DIFF
--- a/examples/ecs/error_handling.rs
+++ b/examples/ecs/error_handling.rs
@@ -55,7 +55,7 @@ fn main() {
 
     // If we run the app, we'll see the following output at startup:
     //
-    //  WARN Encountered an error in system `fallible_systems::failing_system`: "Resource not initialized"
+    //  WARN Encountered an error in system `fallible_systems::failing_system`: Resource not initialized
     // ERROR fallible_systems::failing_system failed: Resource not initialized
     //  INFO captured error: Resource not initialized
     app.run();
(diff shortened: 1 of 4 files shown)


examples/ecs/error_handling.rs
```diff
@@ -55,7 +55,7 @@ fn main() {
 
     // If we run the app, we'll see the following output at startup:
     //
-    //  WARN Encountered an error in system `fallible_systems::failing_system`: "Resource not initialized"
+    //  WARN Encountered an error in system `fallible_systems::failing_system`: Resource not initialized
```
chescock, line 58: I changed this line of documentation because we'll no longer print the quotes, but the example doesn't actually output this error!  The sole use of `failing_system` is used with `pipe` to handle the result.

ItsDoot (approved):

JaySpruce (approved):

bushrat011899 (approved):

